PR #11510, opened 2023-04-25
Windows buildenv: Fix 7-Zip quoting

fwcd:
Due to quoting, the `7z` invocation ended up being expanded as

```
""7z.exe"" x -o"C:\Users\<user>\git\mixxx\buildenv" "C:\Users\<user>\git\mixxx\buildenv\mixxx-deps-2.4-x64-windows-db02b82.zip"
```

which led to 'command not found' errors while running the script. This fixes the quoting by stripping one level of quotes around 7z.exe and moving -o into quotes too (otherwise [PowerShell will complain](https://stackoverflow.com/questions/1673967/how-to-run-an-exe-file-in-powershell-with-parameters-with-spaces-and-quotes))

### DIFF
--- a/tools/windows_buildenv.bat
+++ b/tools/windows_buildenv.bat
@@ -114,7 +114,7 @@ EXIT /B 0
 
 
 :DETECT_SEVENZIP
-    SET SEVENZIP_PATH=7z.exe
+    SET SEVENZIP_PATH="7z.exe"
     !SEVENZIP_PATH! --help >NUL 2>NUL
     IF errorlevel 1 (
         SET SEVENZIP_PATH="c:\Program Files\7-Zip\7z.exe"
@@ -123,11 +123,11 @@ EXIT /B 0
             SET SEVENZIP_PATH="c:\Program Files (x86)\7-Zip\7z.exe"
             !SEVENZIP_PATH! --help >NUL 2>NUL
             if errorlevel 1 (
-                SET SEVENZIP_PATH=
+                SET SEVENZIP_PATH=""
             )
         )
     )
-    SET RETVAL="!SEVENZIP_PATH!"
+    SET RETVAL=!SEVENZIP_PATH!
     GOTO :EOF
 
 
--- a/tools/windows_buildenv.bat
+++ b/tools/windows_buildenv.bat
@@ -72,7 +72,7 @@ EXIT /B 0
             CALL :UNZIP_POWERSHELL "!BUILDENV_PATH!.zip" "!BUILDENV_BASEPATH!"
         ) ELSE (
             ECHO ^Unpacking "!BUILDENV_PATH!.zip" using 7z...
-            CALL :UNZIP_SEVENZIP "!RETVAL!" "!BUILDENV_PATH!.zip" "!BUILDENV_BASEPATH!"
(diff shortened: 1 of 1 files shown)


Swiftb0y:
The CI seems to pass, so no regressions there I guess, but I don't understand under which circumstances this breaks since it didn't break on CI afaik.

fwcd:
Perhaps CI doesn't use the 7-zip extractor, so it never went down that code path?

JoergAtGithub:
This needs to be tested with cmd.exe and Powershell. And especially with "x64 Native Tools Command Prompt for VS 2019" which is required by our Readme.md.

fwcd:
I agree, though this is arguably unrelated to the Visual Studio tools since this is just about the 7-Zip invocation. Since it's a Batchfile, it shouldn't make a difference since the script will execute in `cmd.exe` anyway (at least that is my understanding).

daschuer:
How about simplify the script by using: 
```
cmake -E tar xzf ...
```
This can handle zip files and is mandatory anyway. So we don't need to probe for any other unzip tool.

fwcd:
Done

JoergAtGithub:
The different ZIP tools are probed for performance reasons. There are several minutes difference in unzipping with Windows buildin ZIP tool and 7Zip. The Powershell version is somewhere inbetween, but not available in any Powershell version. Removing this requires benchmarking the CMake version against 7Zip.

fwcd:
Perhaps we should then fix the quoting issue first, so people can use the 7z extractor again, and move the CMake version to a separate PR where it can be tested more extensively?

JoergAtGithub:
This does not work for me:
grafik

fwcd:
Since we probably want to take our time to get the CMake extraction right, especially to make sure that we don't regress on performance, I've moved it to #11524 for further discussion.

I have also retargeted this PR to 2.4 and reset it to the original fix, so users can at least use the 7-Zip extractor again (which is broken in its current state).

JoergAtGithub (changes requested):
Unfortunately, this doesn't work for me. The default path for 7z `C:\Program Files\7-Zip\7z.exe` contains a space, which is only considered, if the path is in quotes:
grafik

JoergAtGithub:
Could you give us more details, with which setup the original issue occured?

fwcd:
Ah I think the issue is that we get different levels of quoting depending on whether `7z` is on the user's PATH:

https://github.com/mixxxdj/mixxx/blob/1dfd1dc85d78d9d7ee53a09d91a3b526db5076f4/tools/windows_buildenv.bat#L117-L124

On my dev machine that's the case, which is likely why we got different results. Batch files aren't 'smart' about their quotes (like in Unix shell scripts), making this easy to overlook.

fwcd:
I've updated `DETECT_SEVENZIP` to use consistent quoting. @JoergAtGithub, could you check whether that fixes the issue for you?

JoergAtGithub:
This works for me now. Tested with:
-7z started by double click from Explorer
-Powershell by double click from Explorer
-7z started from "x64 Native Tools Command Prompt for VS 2022" 

The later failed during bitsadmin download, but this is unrelated to this change.
Also unrelated is the Clazy fail in CI.

Thank you!

PR #11511, opened 2023-04-25
Allow playing tracks longer than 6 h

### LINKED ISSUE #11504: Mixxx doesn't play very long track (9h)
### Bug Description

I am having a very long track in my collection that does not play in Mixxx.
It's 9h05 long. After mixxx analysed the track it doesn't show any waveform and refuse to play it.
Even it says it's ready. 
Also it shows the wrong tracklength in the desk. Library length shows correctly in opus or m4a format.
Same track in AAC shows wrong filelength in the library too.

My Mixxx version is 2.3.3. (HEAD)
Kubuntu 22.04

mixxx-long tracks

### Version

2.3.3.

### OS

Kubuntu 22.04

daschuer:
After this PR is merged, we can play 13h31:35.8 length @ 44100 Hz on 32 bit targets and 3186904 years on 64 bit.

This fixes #11504


daschuer:
This can be considered as bugfix. Merge to 2.3.5?

daschuer:
Conflicts again resolved

daschuer:
Has one interest to test and review this or should we move this to 2.3.6?

src/engine/cachingreader/cachingreader.h
```diff
@@ -111,7 +111,7 @@ class CachingReader : public QObject {
   signals:
     // Emitted once a new track is loaded and ready to be read from.
     void trackLoading();
-    void trackLoaded(TrackPointer pTrack, int iSampleRate, int iNumSamples);
+    void trackLoaded(TrackPointer pTrack, int trackSampleRate, double trackNumSamples);
```
Swiftb0y, line 114: why do you prefer double of `int64_t`/`long long int` in this case?

Swiftb0y (changes requested):

src/analyzer/analyzerbeats.h
```diff
@@ -46,7 +48,6 @@ class AnalyzerBeats : public Analyzer {
     bool m_bPreferencesFastAnalysis;
 
     mixxx::audio::SampleRate m_sampleRate;
-    SINT m_totalSamples;
-    int m_iMaxSamplesToProcess;
-    int m_iCurrentSample;
+    SINT m_maxFramesToProcess;
+    SINT m_currentFrame;
```
Holzhaus, line 52: Can't we use FramePos here?

src/analyzer/analyzerebur128.h
```diff
@@ -14,8 +14,10 @@ class AnalyzerEbur128 : public Analyzer {
         return rgSettings.isAnalyzerEnabled(2);
     }
 
-    bool initialize(TrackPointer tio, int sampleRate, int totalSamples) override;
-    bool processSamples(const CSAMPLE* pIn, const int iLen) override;
+    bool initialize(TrackPointer tio,
```
Holzhaus, line 17:
```suggestion
    bool initialize(TrackPointer pTrack,
```

src/sources/audiosource.cpp
```diff
@@ -27,6 +27,7 @@ AudioSource::AudioSource(
           m_signalInfo(signalInfo),
           m_bitrate(inner.m_bitrate),
           m_frameIndexRange(inner.m_frameIndexRange) {
+    DEBUG_ASSERT(m_frameIndexRange.length() >= 0);
```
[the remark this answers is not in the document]
daschuer: What do you mean with "technically UB"? 

This basically asserts the trivial condition here: 
https://github.com/mixxxdj/mixxx/blob/4a51c393eb8820a5728f4b2b53dce8ef203d2b8c/src/util/indexrange.h#L68
My guess is that the I originally mean to assert that the range is not backward like in `AudioSource::initFrameIndexRangeOnce()`
So I think it is reasonable to replace it with a check that range is not backward.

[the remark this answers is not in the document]
Swiftb0y: > What do you mean with "technically UB"?

No, I mean integer over-/underflow is only defined behavior on unsigned integral types. So there is no reliable way of detecting them just by inspecting the value (strictly theoretically speaking). The only way would be to check after every arithmetic operation (feasible by using an abstraction such as [SafeInt](https://github.com/dcleblanc/SafeInt)). 

But that's just my theoretical two cents on the matter and not a change I propose for this PR. 

> My guess is that the I originally mean to assert that the range is not backward like in AudioSource::initFrameIndexRangeOnce()

I apologize, the location of the comment was chosen suboptimally. The concern rather applies to the changes in the `Analyzer*` classes such as here:

https://github.com/mixxxdj/mixxx/pull/11511/commits/ae94ca7d8a4fcfecfa2fb1f6fbaeb9f4f1fd1bcc#diff-d76237758c9ef4a039bb906e468962e074a9464f40a0219a273d4263523ed88fR47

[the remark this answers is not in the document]
daschuer: The link cites `AnalyzerBeats::initialize()`
```
    if (frameLength <= 0) {
        return false;
    }
```
There is noting wrong with checking the the input `frameLength` here. 

> So there is no reliable way of detecting them just by inspecting the value (strictly theoretically speaking). 

This is the case for signed AND unsigned values, because the overflowed value can be a valid value in an expected range in both cases. So we are here on the same page IMHO.  

The undefined signed int overflow is fortunately only true for C++ itself. The int overflow behavior is well defined by our current target architectures. I know there are some exotic target architectures out there, where this is not the case, but they will never be a Mixxx build target. I am also in doubt that Qt will run on them. 
Can we put a statement in our style guide or implement a unit test for in overflow. Than we can cite it next time.

src/engine/cachingreader/cachingreader.h
daschuer: This is used as double in the receiver EngineBuffer::slotTrackLoaded() and was double in CachingReaderWorker::loadTrack() the change is a fix.

src/analyzer/analyzerbeats.h
daschuer: FramePos is not available in the 2.3 branch.

src/engine/cachingreader/cachingreader.h
Swiftb0y: yes but why use `double` when we only want/expect precise integer values?

src/test/analyserwaveformtest.cpp
```diff
@@ -57,7 +58,9 @@ class AnalyzerWaveformTest : public MixxxTest {
 
 // Basic test to make sure we don't alter the input buffer and don't step out of bounds.
 TEST_F(AnalyzerWaveformTest, canary) {
-    m_aw.initialize(m_pTrack, m_pTrack->getSampleRate(), kBigBufSize);
+    m_aw.initialize(m_pTrack,
+            mixxx::audio::SampleRate(44100),
+            kBigBufSize / kChannelCount);
```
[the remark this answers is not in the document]
daschuer: This is a test using the artificial canary buffer.
It can be done with any sample rate and will not break. Now with kCanarySampleRate the info is matches the other definitions regarding that buffer. 

Do you have a different solution in mind?

src/engine/cachingreader/cachingreader.h
daschuer: At one point we need to go from the Audio Source SINT to double. The original 2.3 code does it in the CachingReaderWorker. Using int for passing this value was the bug. double Int double.
Now it is double double double.

src/test/analyserwaveformtest.cpp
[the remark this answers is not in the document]
Swiftb0y: > Do you have a different solution in mind?

Just revert to using `pTrack->getSampleRate()` again? There is no reason to decouple these as far as I can tell.

src/engine/cachingreader/cachingreader.h
Swiftb0y: Right that makes sense, now why is the engine using `double`?

src/sources/audiosource.cpp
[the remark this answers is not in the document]
Swiftb0y: > There is noting wrong with checking the the input frameLength here.

Yes, in this particular case its fine since we're checking the length of something small, but we absolutely can't be doing that for detecting overflow.

> This is the case for signed AND unsigned values, because the overflowed value can be a valid value in an expected range in both cases. So we are here on the same page IMHO.

Well since this is a count of samples, we expect negative values to be unexpected right? and that's why we're checking for `<= 0`. 

> The undefined signed int overflow is fortunately only true for C++ itself. The int overflow behavior is well defined by our current target architectures. I know there are some exotic target architectures out there, where this is not the case, but they will never be a Mixxx build target. I am also in doubt that Qt will run on them.

That does not matter. Its UB in C++ and thus compilers will make use of that for optimization. Here are two examples where the compiler optimizes based on the non-existence of overflow: https://godbolt.org/z/zezfcxWxE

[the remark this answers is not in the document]
Swiftb0y: I realize that I'm complaining about an issue that does not apply to this PR, so we can stop here now if you prefer to. Just to summarize, checking whether integer overflow occurred by inspecting a value (as it looks like on first sight in https://github.com/mixxxdj/mixxx/pull/11511/commits/a5ddb48a572baf5fc353425fdcc24fc6ae951d89 judging by the commit message) will not work. Can we agree on that?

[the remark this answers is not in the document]
Swiftb0y: > Can we put a statement in our style guide or implement a unit test for in overflow. 

Whats the statement supposed to be? "Don't rely on undefined behavior"?

src/engine/cachingreader/cachingreader.h
daschuer: The engine transport is controlled by the output after resampling in whole frames. Fractional rates are causing fractional input positions from the sound source. That's why the engine buffer is using double for positions.

Swiftb0y: right, but that's for positions, why do we have fractional lengths? There are no sample buffers with fractional sizes, right?

daschuer: That is right. The lengths is also the end position. Sure, we can change that to SINT and compare it to the double position on the fly. But that will be a refactoring PR and not a bug fix. I also don't see a benefit to change that.

Swiftb0y: Well if we changed that to SINT, wouldn't we suffer the overflow issue again? The thing I'm worried about is that the code inside those methods will get more difficult since now all of a sudden we might get fractional lengths passed. That will make the code more difficult to reason about. I think a comment and/or a DEBUG_ASSERT is the least we can do document that thats not the case.

src/sources/audiosource.cpp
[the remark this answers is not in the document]
daschuer: > Just to summarize, checking whether integer overflow occurred by inspecting a value (as it looks like on first sight in https://github.com/mixxxdj/mixxx/commit/a5ddb48a572baf5fc353425fdcc24fc6ae951d89 judging by the commit message) will not work. Can we agree on that?

Yes.

> Whats the statement supposed to be? "Don't rely on undefined behavior"?

Something like that: 

Note: Even though the encoding of signed integer is not defined by the C++ standard, the two's complement representation of signed numbers is enforced by the Qt library: https://github.com/qt/qtbase/blame/6.5/src/corelib/global/qtypes.cpp#L362  
This means you can rely that for instance a a int8_t of 127 will become -128 after increment by one. There is no need to consider other singed encodings.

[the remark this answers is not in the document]
Swiftb0y: > Note: Even though the encoding of signed integer is not defined by the C++ standard, the two's complement representation of signed numbers is enforced by the Qt library: https://github.com/qt/qtbase/blame/6.5/src/corelib/global/qtypes.cpp#L362

The linked static assert does not check whether integer overflow is valid. It just checks the boundaries of `int` and infers that they're encoded in two's complement. It says nothing about overflowing behavior!!

src/test/analyserwaveformtest.cpp
```diff
@@ -103,7 +102,7 @@ class AnalyzerWaveformTest : public MixxxTest {
 // Basic test to make sure we don't alter the input buffer and don't step out of bounds.
 TEST_F(AnalyzerWaveformTest, canary) {
     m_aw.initialize(m_pTrack,
-            kCanarySampleRate,
+            m_pTrack->getSampleRate(),
```
Swiftb0y, line 105: please do the same for analyzersilence and amend.

Swiftb0y (approved):
LGTM. We can finish the unresolved discussion points later as they're not strictly relevant to this PR anymore.

src/sources/audiosource.cpp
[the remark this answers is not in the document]
daschuer: Ah right. Here is the original commit https://github.com/qt/qtbase/commit/34ae919f23b7d89844358afea23c73dc6469faf3
The overflow behavior is still undefined but the possible options are narrowed. Since we are at C++20 now and the undefined option is the fastest. There is noting to do.

PR #4027, opened 2021-06-24
Denon Prime 4 - MIDI Mapping

ultra-vi-olet:
This PR intends to allow the usage of the Denon Prime 4 to control Mixxx while in computer mode. The end goal of this PR is for the Prime 4 unit to have as much functionality with Mixxx as it would standalone, except for the screens as these require more than just MIDI scripting.

NOTE: I created a PR for this mapping a few days ago, but I based it off the wrong branch, and rebasing was giving me a lot of errors that I didn't know how to fix. So I closed that PR and re-opened this one on the right branch from the get-go. Apologies for the clutter and mess. I'm still quite new to this process but would like to help wherever I can.

PR #11512, opened 2023-04-26
components hotcue press highlight

Swiftb0y:
This feature request came up while reviewing #4027

I'll take care of the docs after merging.

### DIFF
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -351,9 +351,12 @@
         },
         connect: function() {
             Button.prototype.connect.call(this); // call parent connect
-            if (undefined !== this.group && this.colorKey !== undefined) {
+            if (this.group === undefined) {
+                return;
+            }
+            if (this.colorKey !== undefined) {
                 this.connections[1] = engine.makeConnection(this.group, this.colorKey, function(color) {
-                    if (engine.getValue(this.group, this.outKey)) {
+                    if (this.outGetValue()) {
                         this.outputColor(color);
                     }
                 });
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -301,6 +301,9 @@
         this.outKey = "hotcue_" + this.number + "_enabled";
         Button.call(this, options);
     };
+    HotcueButton.isValidColor = function(colorCode) {
+        return _.isInteger(colorCode) && colorCode >= 0 && colorCode <= 0xFFFFFF;
+    };
     HotcueButton.prototype = new Button({
         unshift: function() {
             this.inKey = "hotcue_" + this.number + "_activate";
@@ -323,7 +326,7 @@
         outputColor: function(colorCode) {
             // Sends the color from the colorCode to the controller. This
             // method will not be called if no colorKey has been specified.
-            if (colorCode === undefined || colorCode < 0 || colorCode > 0xFFFFFF) {
+            if (!HotcueButton.isValidColor(colorCode)) {
                 print("Ignoring invalid color code '" + colorCode + "' in outputColor()");
                 return;
             }
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -361,6 +361,13 @@
                     }
                 });
             }
+            // Users can specify their own highlight(active) function
+            // so the mapping can respond when hotcues are currently active.
+            // The User should not call `trigger` in their highlight function
+            // to avoid crashing from endless recursion.
+            if (typeof this.highlight === "function") {
+                this.connections[2] = engine.makeConnection(this.group, "hotcue_" + this.number + "_activate", this.highlight);
+            }
         },
     });
     var SamplerButton = function(options) {


JoergAtGithub:
The code itself LGTM, but the name "highlight" is not selfexplaining to me. Why not using the same wording as in the CO name, e.g. hotcueActive or hotcueActiveCallback ?

**Out of Scope:**
Regarding the mentioned risk of an endless loop by triggering a connection from inside it's callback. Could we add a flag to the connection class, which inhibits triggering a connection during execution of it's callback: https://github.com/mixxxdj/mixxx/blob/f36cd4f29086aa5370ece5a10d198f690a3619da/src/control/controlobjectscript.cpp#L130-L139

https://github.com/mixxxdj/mixxx/blob/f36cd4f29086aa5370ece5a10d198f690a3619da/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp#L287-L300

Swiftb0y:
> The code itself LGTM, but the name "highlight" is not selfexplaining to me. 

I'm open to the naming, though I chose highlight because that's usually the name of the functionality on lighting consoles where you can select some lights and they light up according to the selection only while they're selected, so it seemed fitting. But I acknowledge that this term might be a little confusing for people unfamiliar with lighting consoles. Nevertheless, I still like the name because its clear enough but generic enough to be also used for other functionality (eg a hotcue could get highlighted when its selected in the mixxx GUI, in case we ever get that feature). Calling it the same name as the CO its connected to seems like its leaking internals.

> Regarding the mentioned risk of an endless loop by triggering a connection from inside it's callback. Could we add a flag to the connection class, which inhibits triggering a connection during execution of it's callback: 

Good idea, we could look into that in the future.

daschuer:
@JoergAtGithub Is this than ready for merge?

JoergAtGithub:
@daschuer Is the term "highlight" self-explaining for you? Maybe it's just me, who doesn't understand what it does from the name.

daschuer:
I understand the term highlight as a hint for the user which is not necessary active state indicator. 

In a common GUI I can imagine a help button that "highlights" all buttons that are related to a feature. 

In the original request it is proposed to "highlight" the jogwheel in scratch mode. That is a fit. 
We maybe also highlight all fee cue buttons when creating a new cue or such ...

Does that match to this PR? Probably not really for my understanding.

Swiftb0y:
> In the original request it is proposed to "highlight" the jogwheel in scratch mode. That is a fit.
> We maybe also highlight all fee cue buttons when creating a new cue or such ...

> Does that match to this PR? Probably not really for my understanding.

Not really, the scope here is much more reduced. All this change really does is (optionally) mapping an output function to the `hotcue_X_active` CO. This CO is usually 1 when the hotcue button is held down and 0 when released. This has only been implemented for the hotcue buttons since by default, plain `components.Button`s mapped to `_activate` COs (`beatloop(roll)_activate`) already have this highlighting behavior.

res/controllers/midi-components-0.0.js
```diff
@@ -348,13 +351,23 @@
...
                         this.outputColor(color);
                     }
                 });
             }
+            // Users can specify their own highlight(active) function
+            // so the mapping can respond when hotcues are currently active.
+            // The User should not call `trigger` in their highlight function
+            // to avoid crashing from endless recursion.
+            if (typeof this.highlight === "function") {
+                this.connections[2] = engine.makeConnection(this.group, "hotcue_" + this.number + "_activate", this.highlight);
```
daschuer, line 369: Did you also consider the shift feature? 

The "_activate" control is the control that is normally bound to the pressed state of the cue button. Not original designed for feedback. 
We already have also the "enabled" alias "status" CO, that has: 
https://github.com/mixxxdj/mixxx/blob/d993ec5bab09533dd124dbbdd039ddb131fafee0/src/engine/controls/cuecontrol.h#L61-L72

Calling the callback "highlight" puts implementation details of the callback into the name. I would prefer to call it according to the event when it is called. like "onActiveChanged" or such. 
In this case "onStatusChanged"?

daschuer (changes requested):

res/controllers/midi-components-0.0.js
Swiftb0y: This is not really what its doing though. It's only highlighting the cue while its being pressed. So yeah its not designed for feedback, but it does tell us when the hotcue button is being pressed and when its being released. The `enabled` CO does not carry the information intended.  

> Did you also consider the shift feature?

Good point, I didn't, I'll have to test how that behaves, but I don't think it makes sense to have this in the shift state because then, there is no hotcue thats actually being activated.

daschuer: Ok, so the issue is that the name of the callback does not reveal which state is highlighted. It can be called "pressed" 
But that also reveals Implementation details, because you can also use the cue activate control without releasing. 

It feels wrong to to me to introduce here a convenient callback without having one for the "status" Co used in skins. We should encourage the mapping authors to use that instead. For this, IMHO corner case, the mapping authors can connect directly to the activate control. 

Can you please outline in a table for hot and cure loops why we need to highlight the pressed state? Maybe we can implement it by extending the status Co even more and that way also skins can benefit from it. 

@ronso0 what is you opinion?

Swiftb0y: We don't _need_ to highlight the pressed state. It's simply a cosmetic feature requested in #4027 that I thought is better solved upstream instead of the mapping itself.
I'll look into what it takes to use `hotcue_X_status` instead. It could simplify the hotcuebutton component in general.

Swiftb0y: So because the user can overwrite `outValueScale`, we can't really change the `outKey` nor easily simplify the button. ComponentsJS needs a refactoring (with breaking changes) for 2.4 anyways, so I'll postpone this change until then.

Swiftb0y: fyi, I wrote down a bunch of thoughts in the wiki if you're interested. I don't plan on acting on them though in the near future. Getting 2.4 stable has priority. https://github.com/mixxxdj/mixxx/wiki/Component-JS-ES7-Brainstorming

Swiftb0y:
Closing for now. This will be implemented properly in a newer componentsJS version.

PR #11513, opened 2023-04-27
CMakeSettings.json: start with a copy of a UTF-8-BOM file. 

daschuer:
Starting with a UTF-8-BOM file removes the non ASCII Characters form the ASCII + code page encoded batch file.
These turns out to get lost regularly.


JoergAtGithub:
@JosepMaJAZ Could you've a look. If I remind correct, the original code was added by you.

daschuer:
If the resulting CMakeSettings.json is not changed by this patch, it should be save to merge.

JosepMaJAZ:
It's an interesting solution. I guess this way it should not get damaged.

daschuer:
CI is working now

daschuer:
@JoergAtGithub can we merge this now?

JoergAtGithub:
I tested this and it works for me. Thank you!

PR #11304, opened 2023-02-24
toggle the menubar with single Alt key press

ronso0:
I decided to make another attempt to hide the menubar, this time by filtering a single Alt keypress + the essentials from #3184, but without the ControlObject and the skin button.
This is not available on macOS since there's a permanent global menu, also in fullcreen mode

I decided to require a single Alt press so Alt bindings in custom keyboard mappings don't toggle the menubar.

**TODO**
- [x] single-press Alt to show or hide the menubar (press duration doesn't matter, only that there's no other key pressed between Alt press and release)
- [x] hide the menu when going fullscreen [^1]
- [x] unhide the menubar if a menu hotkey (like Alt+F) is pressed while the menu is hidden [^2]
- [x] add a label / dummy action to View menu "Toggle menu bar with Alt key"
that action is not clickable (or does nothing respectively), so users have to press **Alt** themselves
- [x] show a hint when going fullscreen
image
I wonder if that should be omitted on macOS altogether?
- [ ] style hint for all skins
- [x] test
- [x] test
- [ ] test

**TESTING** (only done on *ubuntu so far)
I added debug output for the keystroke filtering, so in case something is not working as expected please run from the command line with `--developer` and share the relevant output snippets so we can debug it.

Closed #8902 

[^1]: for me this didn't work when I used the fullscreen **hotkey** (at least in debug builds I tested so far) because the window manager (xfwm4) hotkey and the Mixxx hotkey were identical, and the keystroke was consumed by the window manager.
I have a workaround 167c5696a2d, though that's not applicable to linux distros with a global menu that require the current menubar hack, but for those hiding the menubar isn't available anyway :shrug: Maybe if my workaround is considered mergable I'll try to polish it so it replaces the hook to slotViewFullScreen on demand.
[^2]: I tested with Ubuntu Studio 20.04 (xfce desktop) (menus open immediately) and Ubuntu 22.04 (Gnome) (a second hotkey press is required to unroll the menu)

PR #11526, opened 2023-05-01
toggle the menubar with single Alt key press (auto hide)

### LINKED ISSUE #8902: hide menu in full screen mode
Reported by: **[bronxio](https://launchpad.net/~bronxio)**
Date: 2017-07-12T08:31:56Z
Status: In Progress
Importance: Wishlist
Launchpad Issue: [lp1703777](https://bugs.launchpad.net/bugs/1703777)

---

It would be awesome to be able to hide the menu bar in full screen mode, like in Firefox, for example (menu bar can be shown if you press Alt + some letter for some specific menu, or some Fx key just to show/hide it)

ronso0:
preferred alternative to #11304 which was already reported to work on Windows and Ubuntu.

The menubar is hidden by default and can be toggled with a single Alt key press.
This is available only on Linux and Windows. On macOS there's a permanent global menu, also in fullscreen mode.

This works only with a _single_ Alt press, Alt bindings in custom keyboard mappings don't toggle the menubar.

Users can opt-out via the popup or in Preferences > Interface.

* same mechanics as #11304 (already tested on Linux and Windows):
`setMinimumHeight()` to show, `setFixedHeight(0)` to hide
* show when a menu hotkey is pressed (e.g. Alt +F)
* auto-hide when
    * all menus are closed (click action or press Esc)
    * another widget gets focused
    * fullscreen is toggled
* show a popup after startup presenting this new option (unless the OS has a global menu **and** Mixxx is not fullscreen, then the dialog is shown when switching to fullscreen mode)
* add an option in Preferences > Interface
* add "Auto-hide menu bar" action to the View menu

image
image

The popup introduces this new feature but also allows to change the behaviour. This is supposed to be a safeguard to prevent locking out touch screen users who updated/installed Mixxx but may not have a keyboard attached when starting this new Mixxx version for the first time.

Please test!


ronso0:
I finally fixed all macOS ifndefs, this is ready for review.

daschuer (changes requested):
My original review comments went to the wrong PR
https://github.com/mixxxdj/mixxx/pull/11566

The following findings are pending: 
* Unnecessary popup window on targets with a global menu bar
* glitch when closing menus after opening the menu bar by a direct shortcut Alt+V 
* Issue opening sub-sub menu when navigating with the right cursor key.

ronso0:
> * Issue opening sub-sub menu when navigating with the right cursor key

This is unrelated to this PR, and expected behaviour when the top item is a submenu.
How should that be fixed anyway?

daschuer:
We may just not open the menu by default like in Firefox or move the menus with sub-menus down. 

Directly opening the menu, is IMHO even more usefull if we remember the previous menu. This will remove the need for moving to a diffent menu in most cases.

daschuer:
How about adding "Show menu bar" in the view menu?

ronso0:
> There is an other issue, when pressing Alt-V to open the View menu and than alt gain to hide the menu bar, the menu bar disappears and reapers shot after.

Keypress handling depends on the window manager / desktop environment. Some pass the keypress to the app (e.g. Gnome), some don't (e.g. xfwm4, reason for #11566).
With xfce4, open menus don't act on Alt press.
Same with Alt press: Gnome (Ubuntu 22.04) opens the first menu, xfwm4 does nothing.
In Firefox (doesn't use Qt) in xfwm4 `File` is highlighted, but not opened.

I can try to work around the glitch you noticed with an eventFilter() in WMainMenuBar.

Btw #11572 broke this branch. I'll rebase onto main and fix it.

daschuer:
The glitch with Alt+V and Alt is IMHO a minor issue that does not block this PR. 

If the showing the menu bar without menu is hard to do, can we than show the last menu instead of "File"? 

Did you consider to move the Viny control menu down to not open it when moving right to "Help"?

ronso0:
> If the showing the menu bar without menu is hard to do, can we than show the last menu instead of "File"?

I'm pretty sure this behaviour is system dependent, if a menu is highlighted at all, if that is even opened... thus I'm not inlined to touch it.
You tested with Gnome, right? so please check what Alt press does in other apps.

ronso0:
> Did you consider to move the Viny control menu down to not open it when moving right to "Help"?

This can happen in another PR. I'd prefer to merge the hide feature first.

daschuer:
Did you try to re-implement QMenuBar focus events?   

I just added this with the menu bar always on and it works as if the menu bar where always there:  

```
 void focusInEvent(QFocusEvent *pEv) override {
    	setMinimumHeight(sizeHint().height());
    	qDebug() << "focusInEvent()";
    	QMenuBar::focusInEvent(pEv);
    }

    void focusOutEvent(QFocusEvent *pEv) override {
    	qDebug() << "focusOutEvent()";
    	QMenuBar::focusOutEvent(pEv);
    	setFixedHeight(0);
    }
```

This solution required a lot less code and be haves natural. 
Does it work for you as well?

ronso0:
This is a nice idea, but it doesn't work at all with xfce4. I tried some variations but it doesn't even print the debug messages.

daschuer:
OK than it looks like this is a bit hacky because of xfce4. Can you put a warning that we must only touch the code again when testing with xfce4? 
We may also consider to use the open menu workaround only for xfce4. This will fix the glitch when closing the menu with Alt. But I don't want to put to much extra work on you.

ronso0:
Correct me if I'm wrong, we tested only with Gnome and xfwm4 so far (and Windows 10), but there are _a few more window managers out there_ ;) so I guess testing with xfwm4 revealed issues that may occur anywhere else.
That being said
> We may also consider to use the open menu workaround only for xfce4.

I think we may _not_ apply the open window trick on Gnome.

Anyhow, I'll test some other distros/WMs, for example KDE Plasma, then we can decide how to proceed.

daschuer:
Sounds good. At a bare minimum, we need some comments about the issues and workaround with xfce4 to protect it from removal.

daschuer:
This LGTM, Can you add a comment about the xfwm4 behavior to the code?

ronso0:
While I'd appreciate if we get this into 2.4 this is a last minute feature.
We should definitely testing again on Windows (10 and 7 if possible) and on Linux distros other than Ubuntu & Xubuntu.
@JoergAtGithub @holzhaus do you have time for this?

Jonas-fr:
I know many people find this feature silly or even useless, but for DJs with old PC and low-definition displays (1366x768 or less) each pixel is a premium so removing a menu that is useless when mixing is more than a nice to have, it could actually enable one to display more FX or tracks or even a deck ; )

IMO the WinAMP example is a good one, the main menu was hidden in a small icon (like 8x8px (or even smaller) at the time IIRC), such icon could be easily embedded into any skins out there and this way we skip the whole keyboard shortcut headache with WM.

ronso0:
Yes, the long term solution will probably be a menu button which is an established concept.
FWIW there is a stale PR implementing this, but IIRC this also has usability regressions (screenreader access?)

This PR is just a small step to increase the usable screen area. The menubar itself is not changed and all menu hotkeys work as before. I'm confident the initial hint will also make screenreader users aware of the new feature.

Russe:
I'm interested in this feature, can I do some help with testing?
Currently I'm using Ubuntu 22.04 with XFCE xfwm4 and most recent Mixxx beta.

ronso0:
Sure, feedback is welcome, everything should work as described in the first post.
Though, I don't expect any surprises since your also using xfce.

ronso0:
I've been testing this for months now, no issues whatsoever (inkl. dev-mod, i.e. skin reload without --developer).

Is anyone interested to get this into 2.5?

daschuer:
> Is anyone interested to get this into 2.5?

Yes, unfortunately a conflict has developed.

daschuer:
Thank you. LGTM

ronso0:
Unfortunately it's not quite working anymore as designed:
the auto-hide confirm dialog also pops up when I leave fullscreen (and still have "Ask me again" checked).

I'll take a look.

PR #11271, opened 2023-02-13
main EQ: properly restore parameters + also save 'no effect' to config

ronso0:
* the empty Main EQ `---` wasn't saved to config, so the previous valid effect was restored on next start unnoticed
* Main EQ parameters were not restored because the loaded parameters were overwritten with default values during initial GUI setup, regression introduced by #4884

PR #11272, opened 2023-02-13
Preferences: join EQ and crossfader settings into new Mixer page

ronso0:
Now all mixer settings are in one place.
Plus some small tweaks, incl. the Main EQ fix from #11271 

Closes #9401

PR #4667, opened 2022-02-07
Pref EQ: apply & save settings only in slotApply()

ronso0:
before, "Use the same EQ for all decks" was reset to `true`, overriding the config value.

Noticed here https://mixxx.discourse.group/t/mixx-strongly-began-to-heavily-load-the-system-linux/24068/29

PR #11527, opened 2023-05-01
Pref > Mixer: apply & save settings only in slotApply(), fix bugs, improve UX

ronso0:
This is like #4667 but for the entire new Mixer page.

* update (read from config) only in slotUpdate()
* apply settings only in slotApply()
* ~~add checkboxes to apply EQ shelves and Main EQ instantly (default: off) to allow the previous (unexpected) behaviour~~ applied instantly but saved to config in slotApply() so that Cancel (> slotUpdate() resets to saved values

+ UX tweaks, small bug fixes, review proposals from #11272 

~~This makes quick fix #11271 obsolete but I'll leave that open for now.~~
The last commits adds debug output that allows easy tracing. To be removed after review.

There's still a round-trip issue with the EQ shelves' `getEqFreq()` and `getSliderPosition()`. This is noticable when resetting to defaults (250, 2500), the resulting values are 246 and 2484 :man_shrugging: Maybe someone has an idea...


ronso0:
> There's still a round-trip issue with the EQ shelves' `getEqFreq()` and `getSliderPosition()`. This is noticable when resetting to defaults (250, 2500), the resulting values are 246 and 2484 man_shrugging Maybe someone has an idea...

Fixed with much larger slider ranges and std::round instead of casting to int.

src/preferences/dialog/dlgprefmixerdlg.ui
```diff
@@ -570,6 +581,18 @@
...
      </property>
      <layout class="QVBoxLayout" name="verticalLayout_3">
+
+      <item>
+       <widget class="QCheckBox" name="CheckBoxInstantMainEQ">
+        <property name="toolTip">
+         <string>EQ parameters are applied instantly without clicking Apply</string>
+        </property>
+        <property name="text">
+         <string>Apply EQ parameters instantly</string>
```
daschuer, line 591: We are in beta-state, and there are only 5 weeks left. This is too short to get this translated to a significant degree. 

Is this feature important enough for this? For me it was OK to have that instantly applied. Who will have benefit from Applying it after "Apply" only? Is there a real use case?

daschuer:
I can confirm that the instant applied EQ settings can be surprising. On the other hand I can't remember that a user has complained about it. It itches only because it is a non consistent GUI.  
How about Enable the "Auto Apply" box by default?

daschuer:
This PR is working good.

daschuer:
The Deck 1 label is gone. Is this intended? I think it should be there to fill the empty space. 
In the One for all case, the grayed out boxes are self explaining.

daschuer:
With this PR I see a DEBUG_ASSERT at shut down: 
```
warning [Main] The following 2 controls were leaked:
warning [Main] "[Mixer Profile]" "LoEQFrequency" QObject(0x0)
warning [Main] "[Mixer Profile]" "HiEQFrequency" QObject(0x0)
```

ronso0:
> I can confirm that the instant applied EQ settings can be surprising. On the other hand I can't remember that a user has complained about it. It itches only because it is a non consistent GUI. How about Enable the "Auto Apply" box by default?

I'll revert that and apply the shelves and main EQ parameters instantly.
The motivation was to have a consistent 'apply only when pressing Apply', and Cancel does restore the previous values, but indeed this is impractical for those controls.

So let's apply immediately and write to config in slotApply(), update from config in slotUpdate().

ronso0:
> The Deck 1 label is gone. Is this intended? I think it should be there to fill the empty space. In the One for all case, the grayed out boxes are self explaining.

whoopsy, not intended. I'll fix it.

> With this PR I see a DEBUG_ASSERT at shut down:

Didn't notice that, I'll take a look.

ronso0:
The shelve EQ values are now PollingControlProxies.

Both the EQ shleves and the main EQ are applied instantly, but stored to config only in slotApply().
This allow immediate testing but Cancel can reset to the previous values.

ronso0:
Set to draft because I want to double-check the recent changes / UX

ronso0:
Ready, config is saved and restored correctly and I'm happy now with UX.

@daschuer now just some noop optimization commits. If you don't have time to review those (or think now is not the time for them) let me know and I'll stash them.

daschuer:
We have a conflict now.

ronso0:
I picked a helper commit from #11695 and removed some VERIFY_OR_DEBUG_ASSERT that would only occur when the effects manager code would be changed.

Done, ready for final review.

ronso0:
I decided to change the EQ shelves tooltip now so translators for missing languages can take care of all 'Mixer' occurences at once.

src/effects/defs.h
```diff
@@ -40,6 +40,10 @@ constexpr int kNumEffectsPerUnit = 4;
 
 const QString kNoEffectString = QStringLiteral("---");
 
+const QString kMixerProfile = QStringLiteral("[Mixer Profile]");
+const QString kHighEqFrequency = QStringLiteral("HiEQFrequency");
+const QString kLowEqFrequency = QStringLiteral("LoEQFrequency");
```
daschuer, line 45: The current solution is perfectly OK, but has a performance implications. 

Putting a const QString into a header (internal namespace) will construct (malloc) the QString again in every cpp file that includes the header. 
The memory allocation "may" take place when the QString is used the first time. 

This is in our case when moved to ConfigObject 
https://github.com/mixxxdj/mixxx/blob/dfde28dbe44375706abcf8846e7fd099c1c66a87/src/preferences/configobject.h#L20

It would be possible to share the strings across the translation units by moving this to the global (external) namespace by adding this to a defs.cpp as 
`extern const QString kMixerProfile = QStringLiteral("[Mixer Profile]");`

and make this: 
`extern const QString kMixerProfile;`

This way the memory allocation is only done a single time and than shared in whole mixxx binary

Maybe we can sourround it with a defs namespace or such.

We have more of these shared const QString in Mixxx, so I am not sure if it worth to target the issue here?

ronso0: Okay, thanks for the info. So, like in library prefs.h/cpp (where we could enclose the defs in namespaces `mixxx`, `library` & `prefs` to have `const ConfigKey kSyncTrackMetadataConfigKey` instead of
`const ConfigKey mixxx::library::prefs::kSyncTrackMetadataConfigKey`)

Out of curiosity, how does this compare to `#define` like in [src/defs_urls.h](https://github.com/mixxxdj/mixxx/blob/dfde28dbe44375706abcf8846e7fd099c1c66a87/src/defs_urls.h#L24)?

I wouldn't do it for these 3 strings here, and I didn't find noteworthy occurences of more than one `const QString k*` in header files. File a reminder if you know of more places where this could be improved?

ronso0: I tried
defs.h
```
extern const QString kEmptyEffect;
```
new defs.ccp
```
const QString kEmptyEffect = QStringLiteral("---");
```
but get linker errors.

When omitting `extern` it builds and works well.

ronso0:
_friendly ping_
I'm aware this is not a very thrilling PR though I'd appreciate if we could finally get this merged.
**TODO** find a solution for the shared QStrings (see above)

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -736,25 +733,21 @@ void DlgPrefMixer::drawXfaderDisplay() {
...
-    if (m_pxfScene) {
-        delete m_pxfScene;
-        m_pxfScene = nullptr;
-    }
-    m_pxfScene = new QGraphicsScene();
-    m_pxfScene->setSceneRect(0, 0, sizeX, sizeY);
-    m_pxfScene->setBackgroundBrush(Qt::black);
+    QGraphicsScene* pXfScene = new QGraphicsScene();
+    pXfScene->setSceneRect(0, 0, sizeX, sizeY);
+    pXfScene->setBackgroundBrush(Qt::black);
```
Swiftb0y, line 792: this will leak, as `QGraphicsView->setScene(QGraphicsScene)` will not take ownership of the `QGraphicsScene`. Please add the member back but make it a `paranted_ptr<QGraphicsScene>`. Thank you

Swiftb0y (commented):
didn't get farther than the first 4 commits...

src/preferences/dialog/dlgprefmixer.cpp
ronso0: so I use `m_pxfScene = make_parented...`, but do I need to delete the object being pointed to or is this been taken care of by parented_ptr??

Swiftb0y: Ideally, this wouldn't need memory allocation at all (nor the multistep initialization) but that's too much of a refactoring right now.

ronso0: Okay, thanks. I knew about the parent/object auto-deleton. So parented_ptr is really just a wrapper, no magic involved.

So what would be the difference to `new QGraphicsScene(this);` (current implementation + child of this)?
Except that the pointer is deleted as soon as the scope is closed. In both cases the object remains until the parent is deleted.

ronso0: Furthermore I noticed the loaded scene can be `clear()`ed hence there is actually no need to create a new one every time the curve is updated.

Swiftb0y: > So parented_ptr is really just a wrapper

Yeah, essentially a marker as it fully compiles away.

> So what would be the difference to new QGraphicsScene(this); (current implementation + child of this)?
Except that the pointer is deleted as soon as the scope is closed. In both cases the object remains until the parent is deleted.

I'm actually not sure. I mean its sorta leaking because the pointer has been lost to the class, except the parent QObject knows about it. Also, its probably a use after free and maybe a double delete if it wasn't heap allocated. I don't fully know, either way the situation is too nuanced and bug prone for my liking, so I'd prefer if we used the standard pattern with the `paranted_ptr` (also follows the CppCoreGuidelines better https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es60-avoid-new-and-delete-outside-resource-management-functions).

ronso0: Okay, I sticked to your initial suggestion (restore member, parented_ptr)

ronso0:
While fiddling with the xfader graph I couldn't resist to polish it (make it look like 2023, not 1990 ;)

this PR / before
Screenshot_2023-09-14_12-30-09

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -39,6 +39,15 @@ const ConfigKey kLowEqFreqKey = ConfigKey(kMixerProfile, kLowEqFrequency);
...
 
+const ConfigKey kXfaderModeKey = ConfigKey(EngineXfader::kXfaderConfigKey,
+        QStringLiteral("xFaderMode"));
+const ConfigKey kXfaderCurveKey = ConfigKey(EngineXfader::kXfaderConfigKey,
+        QStringLiteral("xFaderCurve"));
+const ConfigKey kXfaderCalibrationKey = ConfigKey(EngineXfader::kXfaderConfigKey,
+        QStringLiteral("xFaderCalibration"));
+const ConfigKey kXfaderReverseKey = ConfigKey(EngineXfader::kXfaderConfigKey,
+        QStringLiteral("xFaderReverse"));
+
```
Swiftb0y, line 50: this depends on the constants in EngineXfader to be initialized first, right? That order is undefined, so I'm afraid this break / is UB. @daschuer you probably know more about that.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -779,72 +779,84 @@ void DlgPrefMixer::slotUpdate() {
...
-        pointTotalPrev = pointTotal;
-        point1Prev = point1;
-        point2Prev = point2;
+        polylineTotal.append(QPointF(i + 1, (1. - gainTotal) * (sceneH)-3));
+        polylineL.append(QPointF(i + 1, (1. - gainL) * (sceneH)-3));
+        polylineR.append(QPointF(i + 1, (1. - gainR) * (sceneH)-3));
     }
 
-    graphicsViewXfader->setScene(pXfScene);
+    QPainterPath pathTotal;
```
Swiftb0y, line 850: missing include...

Swiftb0y:
> While fiddling with the xfader graph I couldn't resist to polish it (make it look like 2023, not 1990 ;)

Looks great :+1:. Now make it compile please :stuck_out_tongue_winking_eye:

src/preferences/dialog/dlgprefmixer.cpp
ronso0: builds fine on Ubu 20.04 locally and on macOS, so only  22.04 and Win need the include?
"Verrry, verrry strange!"

Swiftb0y: Well, probably just a transient include from somewhere else that got removed later... thats why you always include what you use even if some of the previous includes _may_ already have done so.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -779,77 +780,84 @@ void DlgPrefMixer::slotUpdate() {
...
+
+    const int sceneW = static_cast<int>(m_pxfScene->width());
+    const int sceneH = static_cast<int>(m_pxfScene->height());
 
     // Draw grid
-    for (int i = 1; i < kGrindXLines; i++) {
-        m_pxfScene->addLine(
-                QLineF(0, i * (sizeY / kGrindXLines), sizeX, i * (sizeY / kGrindXLines)), gridPen);
+    const int kGridHLines = 3;
+    const int kGridVLines = 5;
```
Swiftb0y, line 813: should this be constexpr static and at the top of the file to follow the usual pattern?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -779,77 +780,84 @@ void DlgPrefMixer::slotUpdate() {
...
-    double xfadeStep = 2. / (pointCount - 1);
-    for (int i = 0; i < pointCount; i++) {
-        CSAMPLE_GAIN gain1, gain2;
+    // Draw gain curves
+    QPolygonF polylineTotal;
+    QPolygonF polylineL;
+    QPolygonF polylineR;
+    // Required to have symmetrical curves not overdraw at the right
+    int pointCount = sceneW - 2;
+    double xfadeStep = 2. / (pointCount);
```
Swiftb0y, line 829: what do these constants symbolize?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -779,77 +780,84 @@ void DlgPrefMixer::slotUpdate() {
...
-        point2 = QPointF(i + 1, (1. - gain2) * (sizeY)-3);
-
-        if (i > 0) {
-            m_pxfScene->addLine(QLineF(pointTotal, pointTotalPrev), QPen(Qt::red));
-            m_pxfScene->addLine(QLineF(point1, point1Prev), graphLinePen);
-            m_pxfScene->addLine(QLineF(point2, point2Prev), graphLinePen);
-        }
+        gainL *= 0.71f;
+        gainR *= 0.71f;
+        gainTotal *= 0.71f;
```
Swiftb0y, line 844: the purpose of this to make the lines align with the grid, right? So it would be better if the scale was derived from those gridlines then?
```suggestion
        align with first (top) horizontal gridline.
        constexpr static double scaleFactorToAlignWithGrid = (kGridHLines + 1) / kGridHLines
        gainL *= scaleFactorToAlignWithGrid;
        gainR *= scaleFactorToAlignWithGrid;
        gainTotal *= scaleFactorToAlignWithGrid;
```

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -779,77 +780,84 @@ void DlgPrefMixer::slotUpdate() {
...
 
-        // Save old values
-        pointTotalPrev = pointTotal;
-        point1Prev = point1;
-        point2Prev = point2;
+        polylineTotal.append(QPointF(i + 1, (1. - gainTotal) * (sceneH)-3));
```
Swiftb0y, line 846: why the -3?

src/preferences/dialog/dlgprefmixer.cpp
ronso0: * pixel steps for the graph calculation
* steps the xfader has to go to cover the CO (potmeter) range -1..0..1

src/preferences/dialog/dlgprefmixer.cpp
ronso0: Tto shift the curve up, otherwise it'd stick to the bottom. And appearantly, in conjunction with the v-scaling, to align the curves with the grid.

src/preferences/dialog/dlgprefmixer.cpp
ronso0: Yes, that could work (it'd be `kGridHLines / (kGridHLines + 1)`).

Since I switched to QPainterPath we may simply transform the path later on. I'll do a brief test, actually I didn't want get lost in the details, but I touched it so let's go.

ronso0: > Since I switched to QPainterPath we may simply transform the path later on.

Nope, only the painter can be scaled and translated, but that would scale the line width, too.

I picked your scaler proposal and extended that to consider the v-offset added later on.

ronso0:
The xfader graph looks lke this now:
Screenshot_2023-09-15_16-26-30

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -220,245 +253,179 @@ void DlgPrefMixer::slotNumDecksChanged(double numDecks) {
...
-
+void DlgPrefMixer::setDefaultShelves() {
     SliderHiEQ->setValue(
-            getSliderPosition(highEqFreq,
+            getSliderPosition(2500,
                     SliderHiEQ->minimum(),
                     SliderHiEQ->maximum()));
     SliderLoEQ->setValue(
-            getSliderPosition(lowEqFreq,
+            getSliderPosition(250,
```
Swiftb0y, line 426: why are these magic constants all of a sudden?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -908,83 +991,85 @@ void DlgPrefMixer::slotMainEqEffectChanged(int effectIndex) {
...
-                ConfigValue(pManifest->uniqueId()));
+        int i = param->index();
+
+        QLabel* pCenterFreqLabel = new QLabel(this);
+        QString labelText = param->name();
+        pCenterFreqLabel->setText(labelText);
+        m_mainEQLabels.append(pCenterFreqLabel);
+        slidersGridLayout->addWidget(pCenterFreqLabel, 0, i + 1, Qt::AlignCenter);
+
+        QSlider* pSlider = new QSlider(this);
```
Swiftb0y, line 1018: these are being leaked right now from what I can tell. seems like a not-insignificant memory leak to me.

Swiftb0y (commented):
minor comments, still didn't get through everything, but I feel like thats just the nature of QWidgets...

src/preferences/dialog/dlgprefmixer.cpp
Swiftb0y: mhmm okay, don't really understand it but as long it looks fine, it should be fine.

src/preferences/dialog/dlgprefmixer.cpp
ronso0: They have been before.
There's an offset in the diff: before points to `loadSettings()`, after to `setDefaultShelves()`.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -220,245 +253,179 @@ void DlgPrefMixer::slotNumDecksChanged(double numDecks) {
...
-                        box->itemData(box->currentIndex()).toString());
-
-        box->clear();
-        int currentIndex = -1; // Nothing selected
-
-        // Add empty item at the top: no effect
-        box->addItem(kNoEffectString);
-        int i = 1;
+        // Get currently loaded EQ effect
+        int deck = m_deckEqEffectSelectors.indexOf(pBox);
```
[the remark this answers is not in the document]
ronso0: okay, so I'll rework all 5 loops in this file.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -220,245 +250,179 @@ void DlgPrefMixer::slotNumDecksChanged(double numDecks) {
...
-    m_inSlotPopulateDeckEffectSelectors = false;
-}
-
-void DlgPrefMixer::populateDeckEqBoxList(
-        const QList<QComboBox*>& boxList,
-        EffectManifestFilterFnc filterFunc) {
-    const QList<EffectManifestPointer> pManifestList = getFilteredManifests(filterFunc);
-    for (QComboBox* box : boxList) {
+    const QList<EffectManifestPointer> pManifestList = getDeckEqManifests();
+    for (QComboBox* pBox : qAsConst(m_deckEqEffectSelectors)) {
```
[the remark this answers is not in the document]
ronso0: Doesn't apply anymore when I change the loops as suggested below.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -48,167 +70,178 @@ DlgPrefMixer::DlgPrefMixer(
...
-    int oldDecks = m_deckEqEffectSelectors.size();
     while (m_deckEqEffectSelectors.size() < static_cast<int>(numDecks)) {
         int deckNo = m_deckEqEffectSelectors.size() + 1;
+        QString deckGroup = PlayerManager::groupForDeck(deckNo - 1);
 
         QLabel* label = new QLabel(QObject::tr("Deck %1").arg(deckNo), this);
 
-        // Create the drop down list for deck EQs
+        // Create the EQ selector //////////////////////////////////////////////
         QComboBox* pEqComboBox = new QComboBox(this);
```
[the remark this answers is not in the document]
Swiftb0y: The lambda is capturing `pQuickEffectComboBox` by value (which is fine for a raw pointer, but not fine for a smart-ish-pointer such as `parented_ptr`). Thus the compiler wants to create a copy which it can't because the copy-ctor is deleted (intentional and correct for now). 
This is solved most easily by capturing by reference: `[this, &pQuickEffectComboBox]`

@daschauer do we want to rethink the `parented_ptr` semantics? Eg enabling moves and and copies since the `parented_ptr` is not really owning anyways.

src/preferences/dialog/dlgprefmixer.cpp
ronso0: :+1: using parented_ptr for all spontaneously created widgets now, and for the numDecks ControlProxy.
no `new` anymore in this file.

src/preferences/dialog/dlgprefmixer.cpp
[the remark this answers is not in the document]
ronso0: no, there were stil occurences of this.

src/preferences/dialog/dlgprefmixer.cpp
Swiftb0y: Thank you :+1:

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -220,245 +253,179 @@ void DlgPrefMixer::slotNumDecksChanged(double numDecks) {
...
-            box->addItem(pManifest->name(), QVariant(pManifest->uniqueId()));
-            if (pCurrentlySelectedManifest &&
-                    pCurrentlySelectedManifest.data() == pManifest.data()) {
+            pBox->addItem(pManifest->name(), QVariant(pManifest->uniqueId()));
+            int i = pBox->count() - 1;
+            // <b> makes the effect name bold. Also, like <span> it serves as hack
+            // to get Qt to treat the string as rich text so it automatically wraps long lines.
+            pBox->setItemData(i,
+                    QVariant(QStringLiteral("<b>") + pManifest->name() +
+                            QStringLiteral("</b><br/>") + pManifest->description()),
```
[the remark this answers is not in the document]
Swiftb0y: I'm afraid you missed some occurrences, this is duplicated three times across the file.

src/preferences/dialog/dlgprefmixer.cpp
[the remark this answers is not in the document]
Swiftb0y: Thanks for changing them too.

src/preferences/dialog/dlgprefmixer.cpp
ronso0: but I messed it up when using it for the Main EQ widgets,crashes with qDeleteAll and/or QList<Widged pointers> in `slotMainEqEffectChanged(int)` here https://github.com/mixxxdj/mixxx/pull/11527/commits/5f43d7e786fdce9c2474895f929ccf3a2d54ee24#diff-c4e587c957428bd6afd10136eb4dad00d554179984d800b2252947ef2c27e666L984

I'll take a look at this later on.
Or we stick to `new` and re-add the destructor to delete those widgets.

src/preferences/dialog/dlgprefmixer.cpp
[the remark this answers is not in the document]
ronso0: :+1: fixed

src/preferences/dialog/dlgprefmixer.cpp
Swiftb0y: I'll try to look into it later as well. You can probably just get rid of the `qDeleteAll` call.

src/effects/defs.h
daschuer: @ronso0 My guess is it does not work, because defs.h is not included in defs.cpp. 

defs.h 
`extern const QString kEmptyEffect;` 
extern is here is a prototype definition in the global namespace it means 
* link to a kEmptyEffect from the global namespace. 
* Don't initialize or call constructor

defs.cpp 
`extern const QString kEmptyEffect;` via `#include "defs.h"`
`const QString kEmptyEffect = QStringLiteral("---");`
means instantiate and initialize kEmptyEffect. 
Use the publish it as kEmptyEffect in the global namespace. 

You may also join extern and the initialization 
`extern const QString kEmptyEffect = QStringLiteral("---");`

I think we should prefer that to indicate that it is extern visible. 

The allocation is done before main and leaked finally. 

Advantage: we have only a single allocation. 
Disadvantage cutter the global namespace (can be fixed by a named namespace).

daschuer: > I thought the entire point of the QStringLiteral macro was that there was no allocation taking place?!?

Oh right, my bad. We have no heap allocation here. 

The uinicode string along with 20 bytes of control data is used directly from the text segment and of every compilation unit that includes and uses the string. This is almost like a constexpr with reference counting disabled. 
An addition there is a pointer to that data in the bss initalized with that location.    

So there is no performance benefit, but a negligible memory impact. 

Sorry for he noise.

src/preferences/dialog/dlgprefmixer.cpp
ronso0: Looks like this is about some other widgets I made parented_ptr.

src/effects/defs.h
Swiftb0y: Right, moreover, what about initialization order? The initialization order of TU's is undefined. If you declare something as extern const, you need to take care that no other global variable is being initialized based on it. This reeks of a potential source of bugs. And even if the string is `QStringLiteral`, the constructor for that variable still needs to run in order for the variable to be usable, even if the constructor doesn't do much, right?

daschuer: > If you declare something as extern const, you need to take care that no other global variable is being initialized based on it.

Yes. In this case it is no issue, because the constructor boils down to a pointer initializer. No
Other dependencys.

src/preferences/dialog/dlgprefmixer.cpp
Swiftb0y: Okay, I double checked, simply removing the `qDeleteAll`s is the correct choice. The compiler should've warned you if you were to call `qDeleteAll` on a `QList<parented_ptr<T>>`, but it apparently didn't. The resulting code still won't be leaky since all objects in the QList will have a parent (as `DEBUG_ASSERT`ed when calling `.clear()` on the container).

The thing I'm rather confused by is that apparently `QList::append`ing a `parented_ptr<T>` works, even though it shouldn't because copy and move have been deleted...

Swiftb0y: Then its probably from [this lambda](https://github.com/mixxxdj/mixxx/pull/11527/files#diff-c4e587c957428bd6afd10136eb4dad00d554179984d800b2252947ef2c27e666R226-R235). It's probably that the lambda (the connection its bound to) outlives the `pQuickEffectComboBox` reference, which lives shorter because of some `m_deckQuickEffectSelectors.clear()` call. But I wonder why that bug wasn't there previously.

ronso0: It's not the Main EQ (and therefore: false alarm, qDeleteAll doesn't cause any issues), and there is no `m_deckQuickEffectSelectors.clear()`.

But it's indeed that lambda and that `pQuickEffectComboBox` reference, though the crash happens during startup and the presetChanged() signal is not emitted.

I worked around it by capturing just the deck group string and the preset name and update the index in a separate slot (just to have a compact lambda body).

src/effects/defs.h
ronso0: > @ronso0 My guess is it does not work, because defs.h is not included in defs.cpp.

Aaahm, nope. I also tried namespace(s) to no avail.
The pattern you suggest works well with ConfigKey (like in library_prefs.h/cpp) but appearantly not with QString.
Btw I stopped investigating what's wrong.

daschuer: Yes, let's move on, not worth to discuss it in the scope of an unrelated PR.

src/preferences/dialog/dlgprefmixer.cpp
ronso0: ...and I added a note about the crash and the workaround.

src/preferences/dialog/dlgprefmixer.cpp
Swiftb0y: ping @daschauer
Can you lend us your insight here? I'm blocking this PR over my concern here right now.

daschuer: This will work, because the initialization order of variables is undefined, but order of the regions is defined. 

1. The text segment, this is "initalized" at compile time and copied to the RAM via the OS binary loader. That are constexpr and static const.    
2. Global variables in the bss. They are nomaly memcpyed() by from their initalization values or just zeroed out.  
3. Global Object constructors in undefined order.  
4. Main()  

Since EngineXfader::kXfaderConfigKey is 1. 

I have recently verifies that a QStringLiteral() initalized variable is a combination from 1. 2.  So it means it would work as well. But it is a little bit brittle, because it is not a constexpr.

Swiftb0y: Okay if you say so, I'll trust you to have more expertise here. Can still you link me a doc for this? I'd like to read more detail on it.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -220,245 +249,182 @@ void DlgPrefMixer::slotNumDecksChanged(double numDecks) {
...
+            pBox->setItemData(currentIndex,
+                    QVariant(QStringLiteral("<b>%1</b><br/>%2")
+                                     .arg(pLoadedManifest->name(),
+                                             pLoadedManifest->description())),
+                    Qt::ToolTipRole);
+            // Deactivate item to hopefully clarify the item is not an EQ
+            const QStandardItemModel* pModel =
+                    qobject_cast<QStandardItemModel*>(pBox->model());
+            DEBUG_ASSERT(pModel);
+            auto pItem = pModel->item(currentIndex);
```
Swiftb0y, line 311: clang-tidy. also affects `slotSingleEqToggled`
```suggestion
            auto* pItem = pModel->item(currentIndex);
```

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -220,245 +253,179 @@ void DlgPrefMixer::slotNumDecksChanged(double numDecks) {
...
+                    QVariant(QStringLiteral("<b>") + pLoadedManifest->name() +
+                            QStringLiteral("</b><br/>") + pLoadedManifest->description()),
+                    Qt::ToolTipRole);
+            // Deactivate item to hopefully clarify the item is not an EQ
+            const QStandardItemModel* pModel =
+                    qobject_cast<QStandardItemModel*>(pBox->model());
+            DEBUG_ASSERT(pModel);
+            auto pItem = pModel->item(currentIndex);
+            DEBUG_ASSERT(pItem);
+            pItem->setFlags(pItem->flags() & ~Qt::ItemIsEnabled);
```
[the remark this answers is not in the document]
Swiftb0y: still missing

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -1017,58 +1119,74 @@ void DlgPrefMixer::validate_levels() {
...
+    for (const auto& pManifest : allManifests) {
+        if (isMixingEQ(pManifest.data())) {
+            eqs.append(pManifest);
+        } else if (!m_eqEffectsOnly) {
+            nonEqs.append(pManifest);
+        }
     }
+    // Append non-EQs. No-op if 'EQs only' is unchecked.
+    eqs.append(nonEqs);
+    return eqs;
```
Swiftb0y, line 1179: same thing, just faster and easier to read IMO. if you need the order to stay the same as in `allManifests` (which I don't think you do tbh), use `stable_partition` instead. 
```suggestion
    auto nonEqsStartIt = std::partition(allManifests.begin(), allManifests.end(), [](const auto& pManifest) {
        return isMixingEQ(pManifest.data());
    });
    if (m_eqEffectsOnly) {
        allManifests.erase(nonEqsStartIt, allManifests.end());
    }
    return allManifests;
```

ronso0: Thanks for another std magic trick!

Had to remove `const` from `allManifests`, and yes, the [lexigraphically sorted list](https://github.com/mixxxdj/mixxx/blob/5b32026fa7cea38291b577367907887e8ad6af90/src/effects/backends/effectmanifest.cpp#L17-L29) was fine so I used `stable_partition`.

I was wondering why Balance was listed as Stereo Balance so I also changed the combobox items to use the `displayName`. And I added a separator between the EQs and non-EQs.

daschuer (approved):
LGTM and works. Thank you.

ronso0:
Finally off my table.. Thanks for your reviews!

ronso0:
Whoops, had to fix the 'Bypass EQ procesing' default, it didn't consider empty string (no cfg value).
https://github.com/mixxxdj/mixxx/commit/e39db38375e4be6372dffd6c09ab0be7fd05daf0

PR #11528, opened 2023-05-02
Fix too strict debug assertion

daschuer:
It seems to be not correct to use the error value after a libmad function succeeds. 
This has slipped through, because the debian package of this library uses a patch that fixes the behavior. 
I was able to reproduce the issue with a unpatched local build of libmad.

### DIFF
--- a/src/sources/soundsourcemp3.cpp
+++ b/src/sources/soundsourcemp3.cpp
@@ -152,7 +152,9 @@ bool decodeFrameHeader(
         logFrameHeader(kLogger.info(), *pMadHeader);
         return false;
     }
-    DEBUG_ASSERT(pMadStream->error == MAD_ERROR_NONE);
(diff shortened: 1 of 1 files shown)


daschuer:
The patch itself fixes: 
This bug: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=287519

Reported upstream here: 
https://lists.mars.org/hyperkitty/list/mad-dev@lists.mars.org/thread/EJI3ZNSQMU2JPYDFCBUAQ3BZPUZ4O7K6/

Also other distros like Fedora is using this patch since libmad-0.15.1b-26

Swiftb0y (approved):
code LGTM. Waiting for @JoergAtGithub to confirm that this fixes the issue.

JoergAtGithub:
This solves the problem for me! Thank you!
But I still don't understand, why the CI builds were not affected - and therefore CI CTests passed.

PR #11533, opened 2023-05-03
Add `Kokkos::mdspan`

Swiftb0y:
mdspan will be in C++23 and could greatly help with making the audio engine more flexible. It allows to interpret a contiguous chunk of memory as a multidimensional `std::span`. The result is that the iteration logic can be separate from the memory access logic (and thus the memory layout), allowing us to write generic audio algorithms which can easily adapt to different channel layouts (consider surround audio and STEMs).
This PR contains a demo commit to confirm the changes to the build actually work, no meaningful work has yet been done in any mixxx-related audio code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2363,6 +2363,11 @@ target_link_libraries(mixxx-lib PUBLIC mixxx-proto)
 # https://github.com/rigtorp/SPSCQueue
 target_include_directories(mixxx-lib SYSTEM PUBLIC lib/rigtorp/SPSCQueue/include)
 
+# Kokkos mdspan
+# https://github.com/kokkos/mdspan
+# TODO: obsolete after adopting C++23
+target_include_directories(mixxx-lib SYSTEM PUBLIC lib/mdspan/include)
+
 # Qt
 if(QT6)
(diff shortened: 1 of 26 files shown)


JoergAtGithub:
Sounds good to me!

Swiftb0y:
> It would be interesting if auto vectorization is still done when we for instance use it for:

Yes, I'm quite certain though it probably depends on the complexity of layout and the accessor implementation. Unfortunately there is not much accessible documentation for mdspan yet. These are probably the best: 
* https://github.com/kokkos/mdspan/wiki/A-Gentle-Introduction-to-mdspan 
* https://www.youtube.com/watch?v=lvylnaqB96U

> Unfortunatly MSVC 2019 is not able to compile it:

They seem to be aware but not doing particularly much about it. https://github.com/kokkos/mdspan/issues/26

We could conditionally use the vcpkg port on MSVC, though I don't really want to look into this right now.

Swiftb0y:
It builds. I'll remove the demo commit and remove draft now.

JoergAtGithub (approved):

daschuer (approved):
LGTM. 

However let's keep it as draft to not merge an unused library to our lib folder. 
If one starts to refactor our code using mdspan they can base on this PR.

Swiftb0y:
Fine for me.

PR #11541, opened 2023-05-07
Avoid ambiguous != operator with QByteRef and QByteRef

daschuer:
This fixes the clazy issue in the 2.4 branch: 
https://github.com/mixxxdj/mixxx/actions/runs/4907974024/jobs/8763430851

### DIFF
--- a/src/test/analyserwaveformtest.cpp
+++ b/src/test/analyserwaveformtest.cpp
@@ -55,15 +55,14 @@ class AnalyzerWaveformTest : public MixxxTest {
         bool pass = true;
         // If the file is not there, we will fail and write out the .actual
         // reference file.
-        QByteArray actual = pWaveform->toByteArray();
-        QByteArray reference;
+        const QByteArray actual = pWaveform->toByteArray();
 
         ASSERT_TRUE(f.open(QFile::ReadOnly));
-        reference = f.readAll();
+        const QByteArray reference = f.readAll();
 
         if (actual.size() == reference.size()) {
             for (int i = 0; i < actual.size(); ++i) {
-                if (actual[i] != reference[i]) {
+                if (actual.at(i) != reference.at(i)) {
                     qDebug() << "#" << i << QString::number(actual[i], 16)
                              << QString::number(reference[i], 16);
                     pass = false;


src/test/analyserwaveformtest.cpp
```diff
@@ -63,7 +63,12 @@ class AnalyzerWaveformTest : public MixxxTest {
 
         if (actual.size() == reference.size()) {
             for (int i = 0; i < actual.size(); ++i) {
-                if (actual[i] != reference[i]) {
+                // oprerator[] returns a QByteRef holding the index and a
+                // reference to the original array. The != operator is not
+                // explicit defined and considered as ambiguous
+                // [-Werror,-Wambiguous-reversed-operator]
+                // The operator!=(char c) is defined.
+                if (actual[i] != static_cast<char>(reference[i])) {
```
Swiftb0y, line 71: not sure I'm comfortable with the cast here to be honest. If you refactor the the declaration of `reference` to be const, `operator[]` will return a char instead of a `QByteRef`. QByteRef is a bit weird anyways IMO since there is no documentation for it. Would probably be better to just make both `QByteArrays` const. 

```cpp
        ASSERT_TRUE(f.open(QFile::ReadOnly));
        QByteArray const reference = f.readAll();
```

daschuer: Oh yes, good idea. I have also used at() to make clear that I want only the char. This way we can also remove the comment. 

Regarding the const position, we follow the Google style recommendation in Mixxx that recommends putting const first: https://google.github.io/styleguide/cppguide.html#Use_of_const

fwcd:
Since the issue seems to be upstream in the way Qt 5 interacts with C++20, another solution would be to just set `-Wno-error=ambiguous-reversed-operator` for now, until we update to a patched version of Qt.

daschuer:
Looking at this case, the warning has helped to improve the code. So I think it is a good idea to keep it in. 
If it bites us over and over again, we can reconsider that.

src/test/analyserwaveformtest.cpp
Swiftb0y: > Oh yes, good idea. I have also used at() to make clear that I want only the char. This way we can also remove the comment.

:+1: 

> Regarding the const position, we follow the Google style recommendation in Mixxx that recommends putting const first: https://google.github.io/styleguide/cppguide.html#Use_of_const

The style guide only recommends to add const, not particularly where IMO. I think we should prefer east const in the future. https://isocpp.org/wiki/faq/const-correctness#const-ref-alt

Swiftb0y (approved):
LGTM. though `.at` is kinda overkill since we really don't need bounds checking (since we already check the size beforehand).

Issue #10690, opened 2022-08-23
Controller mappings removed from Mixxx 2.4 ?


mixxxbot:
Reported by: **[nprevail](https://launchpad.net/~nprevail)**
Date: 2022-03-25T18:20:04Z
Status: New
Importance: Undecided
Launchpad Issue: [lp1966507](https://bugs.launchpad.net/bugs/1966507)

---

Git Version: 2.4-alpha-1131-g02150fa2b6 (HEAD)
Linux 5.16.15-201

Not sure if this was intentional, or perhaps it's due to testing, but I was comparing the most current and stable release, Mixxx 2.3.2, to 2.4.0 Alpha-pre, and noticed that a handful of midi controllers are missing from the Preferences>Controllers library in 2.4.0 Alpha-pre.

I have a Novation Twitch, and a midi map existed in the default Controller library in 2.3.2. I was hoping to do more MIDI controller tests with 2.4.0. Is there any reason why Novation Twitch (and other controllers) are missing?

Just wanted to bring attention to this, in case someone forgot to include midi maps that were in the previous version of Mixxx.

PR #2682, opened 2020-04-19
migrate legacy controller system to QJSEngine

Be-ing:
This builds on @ferranpujolcamins's work to migrate to QJSEngine for ES7 support (#1795). This PR ports the Components library to use ES6 classes.

Note: Mixco mappings for Korg NanoKontrol2 and MAudio Xponent are removed, unmaintainable

PR #11542, opened 2023-05-07
Changelog for 2.4.0

daschuer:
That was quite a but of work. We have now > 900 enties in the 2.4 changelog. 
For a release that will be too long, but for the beta it is IMHO ok. 

Not sure how much we need to meld it down, because we also have the blog post for a condensed version. 

Has one interest to prepare the release version of his change log by  removing the issues and PRs that are not interesting from a user perspective.


res/linux/org.mixxx.Mixxx.metainfo.xml
```diff
@@ -569,7 +674,30 @@
...
+   #11418
+  </li>
+  <li>
+   Verify the "first sound" of as an analysis sanity check
+   #4773
+  </li>
+  <li>
+   Fix a possible crash when ejecting a track #11334](
+   https://github.com/mixxxdj/mixxx/pull/11334
+   )
```
ronso0, line 694:
```suggestion
   Fix a possible crash when ejecting a track
   #11334
```

res/linux/org.mixxx.Mixxx.metainfo.xml
```diff
@@ -1261,6 +1525,17 @@
...
+   #10927
+   #11063
+   #11153
+   #11180
+   #11167
+   #11181
+   #11266
+   #11312
+   [#11389](
+   https://github.com/mixxxdj/mixxx/pull/11389
```
ronso0, line 1538:
```suggestion
   #11389
```

ronso0:
> Has one interest to prepare the release version of his change log by removing the issues and PRs that are not interesting from a user perspective.

Sure, can do that. We can group a lot of UX fixes for example, and some minor bugfixes in different categories.

ronso0 (changes requested):
Pre-commit complains about plain-text urls

.codespellignore
```diff
@@ -6,3 +6,4 @@ sord
 doubleClick
 sur
 jus
+euclidian
```
fwcd, line 9: It's spelled "euclidean", thus probably shouldn't be excluded

Swiftb0y: Yep, sorry my mistake. We should correct the title and the changelog entry of the corresponding PR instead.

Swiftb0y: Yep, sorry my mistake. We should correct the title and the changelog entry of the corresponding PR instead.

daschuer:
rebase on 2.4 to fix the faulty commits and the failing workflow run.

JoergAtGithub:
We should mention, that the mappings for

- Novation Twitch
- Korg Nanokontrol 2
- MAudio Xponent

were removed in this release (see https://github.com/mixxxdj/mixxx/pull/2682) due to incompatibility of the old code with the new mapping system.

ronso0:
Great you remembered that.
See also #10690

daschuer:
Done

ronso0:
Thank you!

Issue #11260, opened 2023-02-08
Hotcue default color will be ignored


Russe:
### Bug Description

When setting a hotcue it will always get the yellow/orange color.
I changed the default hotcue color to a different one at Mixxx settings, but this will not solve the issue.
Every time when setting a new hotcue it will get this yellow/orange color.

For testing I removed folder .mixxx with all settings, but this did not solve the issue.
Ass well I tried to remove and install again most recent alpha version, but this did not solve the issue.
As well I tried with different skins, but this did not solve the issue.

### Version

2.4.0-alpha-pre (2.4-alpha-1403... (main))

### OS

Xubuntu 22.04

Swiftb0y:
Can you post a screenshot of your Color Preferences?

Russe:
As you can see, default color at preferences is green, but when adding a hotcue it will become orange color.

![mixxx-hotcue-colors](https://user-images.githubusercontent.com/1218375/217764837-de4b1569-7cd4-449c-8e8d-d60aafc7d909.png)

Swiftb0y:
Confirmed, I can also reproduce the issue on main. I'll try to find some time to investigate.

Swiftb0y:
Quick note. I can't reproduce the issue with a fresh settings dir. This is a little more complicated than I thought.

Russe:
Currently I'm using this version where I can reproduce the issue: 2.4.0-alpha-pre (2.4-alpha-1403-gcce89d09f3 (main))
It's also reproducible with a new .mixxx folder and new library.

Swiftb0y:
Thank you, I'll try my best to find the cause.

Swiftb0y:
Ah I see why I couldn't reproduce it anymore. Coloring "By hotcue number" works, but not when choosing some other static color.

daschuer:
I have found the issue. It looks like a wrong merge conflict resolve. I will prepare a PR.

Swiftb0y:
Fixed in 2.4 newer. @Russe please test these new versions and reopen this issue if it still occurs.

Russe:
Thanks for fixing this. As I'm on the road during next weeks I can test new versions later.

Russe:
Sorry for late answering. I updated to 2.4 beta and hotcue default is fixed now.
Thanks a lot for your work.

PR #11554, opened 2023-05-14
Fix setting the wrong default cue color.

daschuer:
This PR fixes also the issue that cues are shown with wrong default for a split of a second.

Fixes https://github.com/mixxxdj/mixxx/issues/11260


Swiftb0y (changes requested):
put my review suggestions into a PR: https://github.com/daschuer/mixxx/pull/85

Swiftb0y (approved):
LGTM. I appreciate the team work. Should we merge or do we need a 3rd pair of eyes?

daschuer:
I think we it is OK to just merge in cases like this.

daschuer:
Thank you :-)

PR #11558, opened 2023-05-15
Don't allow to set loop in/out at the silence padding after the track.

### LINKED ISSUE #11557: Loop out at track end is not reachable
### Bug Description

This is because the track is padded with silence to fill the last engine buffer entirely. 
If a loop out is set at this position, it may not be reached in every loop iteration because the silence padding may have a shorter length. 

### Version

2.3.5

### OS

Linux

daschuer:
This fixes https://github.com/mixxxdj/mixxx/issues/11557

### DIFF
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -110,7 +110,6 @@ LoopingControl::LoopingControl(const QString& group,
             Qt::DirectConnection);
 
     m_pQuantizeEnabled = ControlObject::getControl(ConfigKey(group, "quantize"));
-    m_pTrackSamples = ControlObject::getControl(ConfigKey(group, "track_samples"));
     m_pSlipEnabled = ControlObject::getControl(ConfigKey(group, "slip_enabled"));
 
     // DEPRECATED: Use beatloop_size and beatloop_set instead.
@@ -244,11 +243,11 @@ void LoopingControl::slotLoopScale(double scaleFactor) {
         return;
     }
     const double loopLength = (loopSamples.end - loopSamples.start) * scaleFactor;
-    const int trackSamples = static_cast<int>(m_pTrackSamples->get());
+    SampleOfTrack sampleOfTrack = getSampleOfTrack();
 
     // Abandon loops that are too short of extend beyond the end of the file.
     if (loopLength < MINIMUM_AUDIBLE_LOOP_SIZE ||
-            loopSamples.start + loopLength > trackSamples) {
+            loopSamples.start + loopLength > sampleOfTrack.total) {
         return;
     }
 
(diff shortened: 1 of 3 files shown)


ronso0 (changes requested):
.

daschuer:
Done.

ronso0:
Thanks, but unfortunately it's still not clamping to track end.
I tried with the same track many times, and often it failed when dragging the loop_out, but it definitely fails when I let the track play to the end then hit Loop_out.

Finally I figured the issue is that `QuantizeControl::lookupBeatPositions` uses the current position (which may already be / likely _IS_ after the end), so --with a bit of luck-- even `m_pPreviousBeat` is after the end, too : \
With that in mind, maybe it makes sense to go back to square one here, instead of incremental fixes, idk.

Btw, if you decide to keep the current commits, could you please reword the last commit?
"Make sure the loop in and loop out is **not** placed on an extended beat behind the track"

src/engine/controls/loopingcontrol.cpp
```diff
@@ -513,17 +513,26 @@ void LoopingControl::setLoopInToCurrentPosition() {
...
-            if (closestBeat == m_currentSample.getValue()) {
+            if (closestBeat == pos) {
                 quantizedBeat = closestBeat;
             } else {
                 quantizedBeat = m_pPreviousBeat->get();
             }
         } else {
-            quantizedBeat = m_pClosestBeat->get();
+            if (closestBeat > sampleOfTrack.total) {
+                quantizedBeat = m_pPreviousBeat->get();
```
ronso0, line 532: even the previous beat can be after the end, depending on how far the playpos moved past the end.
[QuantizeControl::lookupBeatPositions(double dCurrentSample)](https://github.com/mixxxdj/mixxx/blob/5f5cd28b3f8b471f5f2ca271e1ecacf8f2b1a659/src/engine/controls/quantizecontrol.cpp#L79)

daschuer:
I have fixed the issue along with some corner cases with holding the loop in or out button. Instead of stopping the track or take an inconsistent state it now loops to loop in instead and keeps the track running. 
I consider this much better than the silence of a stopped track.

ronso0:
While that last change isn't expected behaviour IMO it may still be a welcome helper.
And it seems to work around a new (yet unknown) issue in the engine buffer (or ReadAheadManager or LoopingControl::nextTrigger, idk):
* I reverted that last commit
* set the loop_out to the latest possible position
* set something like an 8-beat loop, play
* click and hold loop_in
* when it comes very close to loop_out, the smallest loop size possible is used
* that small loop plays a few times but suddenly the deck stops (seems the play pos crossed the track end)

So, LGTM, I'll do some more testing soonish.

ronso0 (approved):

ronso0:
LGTM, this is much beter than before!

However, there still seems to be a chance to stop the deck unexpectedly.
I set loop_out at track end, let the loop run, and click loop_in repeatedly.
At some pont this should produce the shortest possible loop. Though I managed to stop the deck with that (once, didn't try more often, and I don't see how that could happen :man_shrugging: ).

Another issue I spotted (unrelated to this PR) is that the **waveform** play pos definitely goes beyond the track end with that very short loop.

PR #11561, opened 2023-05-16
AnalyzerWaveform: Fix commented out code

uklotzde:
The commented out code didn't compile.

TODO: The code does not contain any references how the frequency bands have been chosen nor why selected filters are used.

### DIFF
--- a/src/analyzer/analyzerwaveform.cpp
+++ b/src/analyzer/analyzerwaveform.cpp
@@ -3,7 +3,7 @@
 #include "analyzer/analyzertrack.h"
 #include "engine/engineobject.h"
 #include "engine/filters/enginefilterbessel4.h"
-#include "engine/filters/enginefilterbutterworth8.h"
+//#include "engine/filters/enginefilterbutterworth8.h"
 #include "track/track.h"
 #include "util/logger.h"
 #include "waveform/waveformfactory.h"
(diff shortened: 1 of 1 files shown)


src/analyzer/analyzerwaveform.cpp
```diff
@@ -12,6 +12,10 @@ namespace {
 
 mixxx::Logger kLogger("AnalyzerWaveform");
 
+constexpr double LOW_MID_FREQ_HZ = 600.0;
+
+constexpr double MID_HIGH_FREQ_HZ = 4000.0;
```
Swiftb0y, line 17:
```suggestion
constexpr double kLowMidFreqHz = 600.0;

constexpr double kMidHighFreqHz = 4000.0;
```

Swiftb0y (approved):
LGTM, but I'd appreciate if you'd use the naming scheme for the sake of consistency.

src/analyzer/analyzerwaveform.cpp
uklotzde: I was just procrastinating while doing Rust development in parallel :rofl:

Swiftb0y (approved):
Thank you. I'd still like to wait with merging until CI passes. Unfortunately that seems to be broken right now because azure...

daschuer (approved):
All green now. LGTM. Thank you.

PR #11565, opened 2023-05-17
WStarRating: don't update track directly, use signals/slots

### LINKED ISSUE #11540: Updating Star rating in track properties clears other unsaved properties
### Bug Description

Clicking on any of 5 stars in file/track properties inside Mixxx, will revert unsaved changes in all other fields.
I think this bug lasted for long, but never reported it since.

To reproduce:
In library right mouse click on any track > properties >  type anything in any input field/s > click on star class. (doesn't matter the initial star value) > unsaved changes reverted.

Let me know if backtrace log is needed

### Version

2.4-alpha-1467-g425ad1e47a

### OS

Windows 10 Pro 21H2

ronso0:
fixes #11540 

changes to the star rating are now handled via change signals, like with the QLineEdits

**edit:** instead of working around the symptom, as per @uklotzde's suggestion, this PR now adds slots and signals for track rating. Now it's consistent with other track properties and WStarRating  now much smaller.

**Note:** updating the rating (or any other track property) elsewhere emits the `trackCganged` signal and will therefore still wipe pending changes in the properties dialog. This is not a bug.

### DIFF
--- a/src/mixer/basetrackplayer.h
+++ b/src/mixer/basetrackplayer.h
@@ -44,6 +44,7 @@ class BaseTrackPlayer : public BasePlayer {
     virtual void slotCloneFromGroup(const QString& group) = 0;
     virtual void slotCloneDeck() = 0;
     virtual void slotEjectTrack(double) = 0;
+    virtual void slotSetTrackRating(int rating) = 0;
 
   signals:
     void newTrackLoaded(TrackPointer pLoadedTrack);
@@ -51,6 +52,7 @@ class BaseTrackPlayer : public BasePlayer {
     void loadingTrack(TrackPointer pNewTrack, TrackPointer pOldTrack);
     void playerEmpty();
     void noVinylControlInputConfigured();
+    void trackRatingChanged(int rating);
 };
 
 class BaseTrackPlayerImpl : public BaseTrackPlayer {
@@ -90,6 +92,7 @@ class BaseTrackPlayerImpl : public BaseTrackPlayer {
     // to compensate so there is no audible change in volume.
     void slotAdjustReplayGain(mixxx::ReplayGain replayGain);
     void slotSetTrackColor(const mixxx::RgbColor::optional_t& color);
+    void slotSetTrackRating(int rating) final;
     void slotPlayToggled(double);
 
   private slots:
--- a/src/track/track.h
+++ b/src/track/track.h
@@ -436,6 +436,7 @@ class Track : public QObject {
     // adjusted in the opposite direction to compensate (no audible change).
     void replayGainAdjusted(const mixxx::ReplayGain&);
     void colorUpdated(const mixxx::RgbColor::optional_t& color);
+    void ratingUpdated(int rating);
     void cuesUpdated();
     void loopRemove();
     void analyzed();
--- a/src/widget/wstarrating.cpp
+++ b/src/widget/wstarrating.cpp
@@ -10,8 +10,8 @@
 
 WStarRating::WStarRating(const QString& group, QWidget* pParent)
         : WWidget(pParent),
-          m_starRating(0, 5),
-          m_currRating(0) {
+          m_starCount(0),
+          m_visualStarRating(m_starCount, 5) {
     // Controls to change the star rating with controllers.
     // Note that 'group' maybe NULLPTR, e.g. when called from DlgTrackInfo,
     // so only create rate change COs if there's a group passed when creating deck widgets.
@@ -36,23 +36,27 @@ void WStarRating::setup(const QDomNode& node, const SkinContext& context) {
 QSize WStarRating::sizeHint() const {
     QStyleOption option;
     option.initFrom(this);
-    QSize widgetSize = style()->sizeFromContents(QStyle::CT_PushButton, &option,
-                                                 m_starRating.sizeHint(), this);
+    QSize widgetSize = style()->sizeFromContents(QStyle::CT_PushButton,
+            &option,
+            m_visualStarRating.sizeHint(),
+            this);
 
     m_contentRect.setRect(
-        (widgetSize.width() - m_starRating.sizeHint().width()) / 2,
-        (widgetSize.height() - m_starRating.sizeHint().height()) / 2,
-        m_starRating.sizeHint().width(),
-        m_starRating.sizeHint().height()
-    );
+            (widgetSize.width() - m_visualStarRating.sizeHint().width()) / 2,
+            (widgetSize.height() - m_visualStarRating.sizeHint().height()) / 2,
+            m_visualStarRating.sizeHint().width(),
+            m_visualStarRating.sizeHint().height());
 
     return widgetSize;
 }
 
-void WStarRating::slotSetRating(int rating) {
-    m_starRating.setStarCount(rating);
-    m_currRating = rating;
-    update();
+void WStarRating::slotSetRating(int starCount) {
+    if (starCount == m_starCount) {
+        // Unchanged
+        return;
+    }
+    m_starCount = starCount;
+    updateVisualRating(starCount);
 }
 
 void WStarRating::paintEvent(QPaintEvent * /*unused*/) {
@@ -63,7 +67,7 @@ void WStarRating::paintEvent(QPaintEvent * /*unused*/) {
     painter.setBrush(option.palette.text());
     painter.drawPrimitive(QStyle::PE_Widget, option);
 
-    m_starRating.paint(&painter, m_contentRect);
+    m_visualStarRating.paint(&painter, m_contentRect);
 }
 
 void WStarRating::mouseMoveEvent(QMouseEvent *event) {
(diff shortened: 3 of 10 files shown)


src/widget/wstarrating.cpp
```diff
@@ -166,7 +174,11 @@ void WStarRating::mouseReleaseEvent(QMouseEvent* /*unused*/) {
         return;
     }
 
-    m_pCurrentTrack->setRating(m_starRating.starCount());
+    m_currRating = m_starRating.starCount();
+    if (m_updateTrack) {
```
uklotzde, line 178: Thanks for addressing these issues. QML will hopefully help to prevent those mistakes in the (far away) future.

The widget should never directly update the track state on its own. Instead the receiver of the signal could do that if desired or otherwise reject the proposed new value by resetting the widget -> unidirectional data flow.

Offering both options is confusing and not maintainable. But I am afraid that trying to avoid this mixed behavior could open another _Pandora's Box_.

uklotzde: Ideally the widget would only offer a slot for setting the rating (input, write-only) and emit a signal when the rating changed (output, read-only). Maybe a getter for reading the current rating on demand. No track pointer, no implicit context management.

ronso0: I understand what you're saying, and I agree that WStarRating becomes more complex and bit harder to maintain. Looks like this can be changed without too much hazzle, with a few changes to Track and BaseTrackPlayer(Impl) (and mediators like LegacySkinParser, and DlgTrackinfo, of course), and WStarRating would be quite clean then.

I think I'll take a stab at this for 2.4 beta, though it's out of scope for this quick bug fix.

ronso0:
@uklotzde That was easier than I thought, though tedious.
I went back to square one and reverted the previous commit.
Please take a look.

src/widget/wstarrating.h
```diff
@@ -22,16 +22,13 @@ class WStarRating : public WWidget {
...
-    ///
-    /// The value must be consistent with the current track if connected.
-    void setRating(int rating);
-
   public slots:
-    void slotTrackLoaded(TrackPointer pTrack = TrackPointer());
+    void slotSetRating(int rating);
+
+  signals:
+    void ratingChangeRequest(int rating);
```
uklotzde, line 29: Events are named using past tense, i.e. `ratingChanged()`. For requests (method or command) you use an imperative verb.

src/mixer/basetrackplayer.h
```diff
@@ -44,13 +44,15 @@ class BaseTrackPlayer : public BasePlayer {
     virtual void slotCloneFromGroup(const QString& group) = 0;
     virtual void slotCloneDeck() = 0;
     virtual void slotEjectTrack(double) = 0;
+    virtual void slotRatingChangeRequest(int rating) = 0;
```
uklotzde, line 47: `slotSetTrackRating()`

src/widget/wstarrating.cpp
```diff
@@ -118,32 +82,24 @@ void WStarRating::mouseMoveEvent(QMouseEvent *event) {
...
 void WStarRating::slotStarsDown(double v) {
-    if (!m_pCurrentTrack) {
-        return;
-    }
     if (v > 0 && m_starRating.starCount() > 0) {
         int star = m_starRating.starCount() - 1;
-        m_starRating.setStarCount(star);
-        update();
-        m_pCurrentTrack->setRating(star);
+        emit ratingChangeRequest(star);
```
uklotzde, line 94: The signal should only be emitted if the value has actually changed to prevent cycles.

src/widget/wstarrating.cpp
```diff
@@ -162,22 +118,16 @@ int WStarRating::starAtPosition(int x) {
...
-    m_pCurrentTrack->setRating(m_starRating.starCount());
+    emit ratingChangeRequest(m_starRating.starCount());
 }
 
 void WStarRating::fillDebugTooltip(QStringList* debug) {
     WWidget::fillDebugTooltip(debug);
 
-    QString currentRating = "-";
+    QString currentRating;
+    currentRating.setNum(m_currRating);
```
uklotzde, line 128: What's the difference between m_currRating and m_starRating? The widget should not store any redundant state that could get inconsistent.

src/track/track.h
```diff
@@ -436,6 +436,7 @@ class Track : public QObject {
     // adjusted in the opposite direction to compensate (no audible change).
     void replayGainAdjusted(const mixxx::ReplayGain&);
     void colorUpdated(const mixxx::RgbColor::optional_t& color);
+    void ratingUpdated(int rating);
```
uklotzde, line 439: I would prefer the neutral and agnostic _changed_ terminology. That would be inconsistent with the existing signals, but we could add a TODO to those later.

uklotzde (changes requested):
Glad to hear that this was easier than expected. The refactoring avoids to pile up more legacy code.

src/widget/wstarrating.cpp
ronso0: `m_starRating.starCount()` is the current value of StarRating, i.e. what you see (for example while hovering).
`m_currRating` is the confirmed value of the track / pending track record value in DlgTrackInfo, after LeaveEvent.

src/mixer/basetrackplayer.h
ronso0: :+1:

src/widget/wstarrating.h
ronso0: :+1:

src/widget/wstarrating.cpp
uklotzde: Then `m_currRating` is redundant. The widget should maintain only a single value and don't care what it's clients or subscribers are assuming.

src/widget/wstarrating.cpp
ronso0: Sure. This just handles the `group,stars_up` CO and changes the current value. The cycle prevention is in [`Track::setRating`](https://github.com/mixxxdj/mixxx/blob/2.4/src/track/track.cpp#L1363-L1368).

However, I changed this method to work with `m_currRating`, i.e. the stored value delivered by `Track::ratingUpdated`, not the visual state (which avoids confusion in the unlikely case a user hovers the rating **and** uses the CO).

ronso0:
pushed some fixups which I'd like to squash if they're okay.

src/widget/wstarrating.cpp
```diff
@@ -162,22 +118,16 @@ int WStarRating::starAtPosition(int x) {
...
     QString maximumRating = QString::number(m_starRating.maxStarCount());
 
-    if (m_pCurrentTrack) {
-        currentRating.setNum(m_pCurrentTrack->getRating());
-    }
```
ronso0, line 180: showing `- / 5` in the debug tooltip with no track would be the only reason to track the "track loaded" state. not worth the effort IMO

src/widget/wstarrating.cpp
ronso0: > Then `m_currRating` is redundant.

Redundant with what?
The current value of `StarRating` (starCount()) just reflects the count of currently painted stars, be it final (clicked, not necessarily accepted by Track or DlgTrackInfo) or temporary (hovering, mouse moving).

Only `m_currRating` is storing the applied / real value and we need that for the COs and to restore the real value when leaving the widget (after mouse move, not clicked).

ronso0:
> pushed some fixups which I'd like to squash if they're okay.

setting this to draft just to avoid accidental merge.

uklotzde:
> pushed some fixups which I'd like to squash if they're okay.

Sure. I will take a final look when done.

src/widget/wstarrating.cpp
uklotzde: I'll check it later. Only viewed the diff, not the whole code.

uklotzde: Nevertheless my comments reveal that meaning and purpose of those member variables is unclear on first sight.

ronso0:
Go ahead, I'm done, I just addressed your review comments and packed the changes in small chunks for easy review.
Added a missing setrating to DlgTrackInof. All performs as expected.

uklotzde:
Stopping my review here. I wasn't able to fully understand what the code is supposed to do, already wasted too much time. If it works, then feel free to merge it. This has to be rewritten for QML anyway.

ronso0:
@uklotzde Thanks for your time and https://github.com/ronso0/mixxx/pull/43.
Your renaming and cleanup commits are valuable so I'll pick them, I'll just revert (ammend) that one small thing I spotted.

It's working fine now: rating is synced if updated anywhere (library, decks, DlgTrackInfo), as it's not possible to set the rating in an empty deck (or empty DlgTrackInfo, if that could be displayed).

daschuer:
@ronso0 do you still plan to rebase this or does it put to much work on us?

daschuer:
@ronso0 do you still plan to rebase this or does it put to much work on us?

ronso0:
Thanks for the reminder!
Yes, I can rebase onto 2.3

daschuer (approved):
LGTM and works good, Thank you.

ronso0:
Thanks for reviewing!

src/widget/wstarrating.cpp
```diff
@@ -37,55 +36,27 @@ void WStarRating::setup(const QDomNode& node, const SkinContext& context) {
...
+        // Unchanged
+        return;
     }
-}
-
-void WStarRating::slotTrackChanged(TrackId trackId) {
-    Q_UNUSED(trackId);
-    updateRatingFromTrack();
+    m_starCount = starCount;
+    updateVisualRating(starCount);
```
ronso0, line 59: had to add `emit ratingChanged(m_starCount)` in order to make the _up/_down controls to also update the track value
pushed a fix 0550e07eddcb7d45256038c67fac13f40d52328d

PR #11566, opened 2023-05-17
Fullscreen toggle rework

ronso0:
### Motivation
While working on #11304 I noticed that my window manager (xfwm4 on Ubuntu Studio) consumes the fullscreen hotkey F11 and `MixxxMainWindow::slotViewFullScreen()` never gets called, except when I click the menu checkbox.
This happens because the fullscreen hotkey in Mixxx and in my  window manager are identical.
(so most of my fullscreen testing was kinda worthless... : \ )

### Changes
* `MixxxMainWindow::slotViewFullScreen()` does now just **toggle** fullscreen
* `MixxxMainWindow::eventFilter()` catches ALL fullscreen changes (`QWindowStateChangeEvent`), no matter if triggered by OS or Mixxx, and also emits a signal to recreate & connect the menu bar (only on Linux desktops that use a global menu)

This works perfectly fine for me, and I'm curious how it performs on Windows and macOS. Please test!

### TODO
- [x] ensure smooth startup (fullscreen or not), i.e. no flickering, static launch screen

### Testing
* clicking `View` > `Fullscreen` as well as any of the hotkeys printed next to it should toggle fullscreen without any unusual flickering 
* Mixxx should start in fullscreen when using the fullscreen command line arguments `--f` or `--full-screen`, regardless the preferences option "Start in fullscreen"
* fullscreen main window should not flicker if any dialog is openened (About dialog, key wheel window, preferences window etc.)

### DIFF
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -190,10 +190,12 @@ void MixxxMainWindow::initialize() {
     // Turn on fullscreen mode
     // if we were told to start in fullscreen mode on the command-line
     // or if the user chose to always start in fullscreen mode.
-    // Remember to refresh the Fullscreen menu item after connectMenuBar()
+    // The Fullscreen menu item is refreshed in connectMenuBar()
     bool fullscreenPref = m_pCoreServices->getSettings()->getValue<bool>(
             ConfigKey("[Config]", "StartInFullscreen"));
-    if (CmdlineArgs::Instance().getStartInFullscreen() || fullscreenPref) {
+    if ((CmdlineArgs::Instance().getStartInFullscreen() || fullscreenPref) &&
+            // could be we're fullscreen already after setGeomtery(previousGeometry)
+            !isFullScreen()) {
         showFullScreen();
     }
 
@@ -373,6 +375,11 @@ void MixxxMainWindow::initialize() {
     // Show the menubar after the launch image is replaced by the skin widget,
     // otherwise it would shift the launch image shortly before the skin is visible.
     m_pMenuBar->show();
+#ifndef __APPLE__
+    // If we went fullscreen earlier and the desktop features a global menu
+    // this will signal the menu to move into the window and hide itself
+    emit fullScreenChanged(isFullScreen());
+#endif
 
     // The launch image widget is automatically disposed, but we still have a
     // pointer to it.
@@ -498,7 +505,11 @@ void MixxxMainWindow::initializeWindow() {
     Pal.setColor(QPalette::Window, MenuBarBackground);
     m_pMenuBar->setPalette(Pal);
 
-    // Restore the current window state (position, maximized, etc)
+    // Restore the current window state (position, maximized, etc).
+    // This will also restore fullscreen and thereby create a seamless
+    // start if we did shut down while in fullscreen mode and with
+    // [Config],StartInFullscreen = 1
+    // (slotViewFullScreen(true) in  initialize() is a no-op then)
     restoreGeometry(QByteArray::fromBase64(
             m_pCoreServices->getSettings()
                     ->getValueString(ConfigKey("[MainWindow]", "geometry"))
@@ -728,7 +739,8 @@ void MixxxMainWindow::connectMenuBar() {
     connect(m_pMenuBar,
             &WMainMenuBar::showKeywheel,
             this,
-            &MixxxMainWindow::slotShowKeywheel);
+            &MixxxMainWindow::slotShowKeywheel,
+            Qt::UniqueConnection);
 
     // Fullscreen
     connect(m_pMenuBar,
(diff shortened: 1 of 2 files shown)


ronso0:
Note: I consider this a bugfix (though we don't know who's affected), and if others agree I can rebase onto 2.4

daschuer:
A conflict has been developed. 

Can we add the a "Show Menu bar" entry to the view menu as well? Firefox has it as well. 

A difference compared to Firefox is that the file menu opens on Alt. This is convenient on one hand, but surprising.
An issue is that navigating to other menus opens the "Viny Control" submenu before it moves ti the help menu.
This coud be fixed by reordering the sub-menu. 
  
Since I never use the file menu, we may consider to either not open a menu or store the last opened menu and open that. 

There is an other issue, when pressing Alt-V to open the View menu and than alt gain to hide the menu bar, the menu bar disappears and reapers shot after. Is there the visible state not stored at the beginning?

daschuer:
I works the same with Qt6

daschuer (changes requested):
It works on Unity as well. There is an issue that Mixxx initially does pops up the auto hide warning which does not apply on such devices with a Global Menu bar. Can we skip it in this case? 

There is a conflict with ALT key which is used in Unity used to open the global search. I think we can ignore this issue.

ronso0:
You refer to the changes of #11304, while this PR is about a consistent internal fullscreen state, and both can be merged independently.

Anyway, moving comments is cumbersome, so let's discuss it here.
> Can we add the a "Show Menu bar" entry to the view menu as well? Firefox has it as well.

Sure, though when unticking the checkbox the menubar dialog shold be shown. And in this case 
> There is an issue that Mixxx initially does pops up the auto hide warning which does not apply on such devices with a Global Menu bar. Can we skip it in this case?

It should be delayed until Mixxx goes fullscreen.

ronso0:
> There is an other issue, when pressing Alt-V to open the View menu and than alt gain to hide the menu bar, the menu bar disappears and reapers shot after. Is there the visible state not stored at the beginning?

I'll take a look.

daschuer:
You mean the comments are for https://github.com/mixxxdj/mixxx/pull/11526
Yes. Let's stick to the original merge plan and merge that first.

ronso0:
I rebased onto 2.4 so it can be merged independent from #11526, just in case any issues arise there.

src/mixxxmainwindow.cpp
```diff
@@ -169,6 +172,7 @@ void MixxxMainWindow::initializeQOpenGL() {
 #endif
 
 void MixxxMainWindow::initialize() {
+    qWarning() << "     $ initialize";
```
daschuer, line 175: Debug leftover? There are more warning that should be removed or become debug messages if it adds any value to the mixxx.log.

src/mixxxmainwindow.cpp
```diff
@@ -372,6 +379,11 @@ void MixxxMainWindow::initialize() {
     // Show the menubar after the launch image is replaced by the skin widget,
     // otherwise it would shift the launch image shortly before the skin is visible.
     m_pMenuBar->show();
+#ifndef __APPLE__
+    // If we went fullscreen earlier and the desktop features a global menu
+    // this will signal the menu to move into the window and hide itself
```
daschuer, line 384: I don't understand that comment can you rephrase it?

daschuer (changes requested):

src/mixxxmainwindow.cpp
ronso0: This is the last commit with debug output which should help debugging in case something doesn't work as expected.
I'll remove that in case the other commits are considered okay.

src/mixxxmainwindow.cpp
ronso0: I could but I notice it's not required and doesn't belong here. It was required in #11526 at some point for hiding the menubar  after starting in fullscreen mode.

Will remove it.

ronso0:
I rebased onto `main` to clean up the 2.4 milestone.
This needs broad testing, so let's target 2.5

daschuer (approved):
LGTM

What is the status here? Can we just merge it to 2.5 to be part of the upcoming 2.5 beta?

ronso0:
Sure, this is for main/2.5. I'd even argue it's an improvement (fixing a bug in window manager(s)) and could go to 2.4, but I won't fight for that.

We need testing on Windows and macOS @mixxxdj/developers 
> ### Testing
> * clicking `View` > `Fullscreen` as well as any of the hotkeys printed next to it should toggle fullscreen without any unusual flickering 
> * Mixxx should start in fullscreen mode when using the fullscreen command line arguments `--f` or `--full-screen`, regardless the preferences option "Start in fullscreen"
> * fullscreen main window should not flicker if any dialog is openened (About dialog, key wheel window, preferences window etc.)

daschuer:
OK Then lets merge. Thank you.

ronso0:
Cool.
Might also fix one part of #12640
> One issue ...; When I press F11 recently nothing happens.

ronso0:
The DEBUG messages are still in, though I'd say it's a good thing this time : )
Like, if someone runs into issues relevant info is already logged.

PR #3411, opened 2020-12-07
Slimdown moc_compilation.cpp

poelzi:
The moc_compilation file became unbearable large, easily taking 10 minutes
to compile whenever a Q_OBJECT changes.
AUTOMOC scans for <basename>_moc.cpp includes and does not add the file
to the moc_compilation whenever it is included in the propper cpp file.

After this patch, moc_compilation.cpp is empty and the mocs are compiled
with the approptial cpp, increasing recompile speeds by a factor of 10 when
just a header file changed.

PR #11567, opened 2023-05-18
Manual include #include "moc_*.cpp" to speed up build

daschuer:
Most moc files have already been included this includes the recent additions as well.


Swiftb0y:
Can you also fix the remaining mocs for Qt6?

daschuer:
All green finally.

daschuer:
Conflicts resolved

Swiftb0y (commented):
Can you elaborate on how this speeds up the build?

daschuer:
Without including them where required, cmake builds a giant cpp file moc_compile.cpp that is build whenever you touch an QObject based class. Now this file is empty and consumes only the minimum build time.

Swiftb0y:
Ok, so the inclusion forces those mocs to be split into separate translation units, increasing concurrency opportunities in the build process?

daschuer:
Yes, probably as well. The main benefit is however for incremental builds, when only one QObject based class has been touched.

src/util/moc_included_test.cpp
```diff
@@ -0,0 +1,7 @@
+#include "../mocs_compilation.cpp"
+
+// QT_VERSION will be defined by any moc_<header_base>.cpp file included from mocs_compilation.cpp
+// It is empty in case all moc files are included
+#ifdef QT_VERSION
+#error "Not all moc_<header_base>.cpp files included from their cpp sources. This is a Mixxx requirement to speed up the build process. See https://cmake.org/cmake/help/latest/prop_tgt/AUTOMOC.html for details."
```
Swiftb0y, line 6: The link doesn't really explain why this makes sense. It just explains the CMake `moc` integration. Thats why I asked the other questions previously since the benefit just isn't apparent by reading the code and comments.

daschuer: The link in the errror message was ment to help resolving the error when it occurs. 
Do you wish to improve it?

JoergAtGithub:
I did a short test (only one pass) and measured the following build times:
2.4 branch: 5'53''
This PR: 5'48''

daschuer:
Measuring this PR against 2.4 does not reveal the whole picture, because we started a while ago to include the moc files. This PR moves only a few moc files of classes that have been introduced since https://github.com/mixxxdj/mixxx/pull/3411

src/util/moc_included_test.cpp
Swiftb0y: yes, I think someone encountering this error doesn't really what to do to make it go away. Neither do so I struggle to find a better message. I think adding something like `please ensure to add #include "moc_<header_base>.cpp" in your all files with QObject-based classes`. is that correct? I'd then move the "This is a Mixxx requirement to speed up the build process. See https://cmake.org/cmake/help/latest/prop_tgt/AUTOMOC.html for details." into a plain comment.

src/util/moc_included_test.cpp
```diff
@@ -0,0 +1,8 @@
+#include "../mocs_compilation.cpp"
+
+// QT_VERSION will be defined by any moc_<header_base>.cpp file included from mocs_compilation.cpp
+// It is empty in case all moc files are included, a requirement to speed up incremental builds.
+// See https://cmake.org/cmake/help/latest/prop_tgt/AUTOMOC.html for details.
+#ifdef QT_VERSION
+#error "mocs_compilation.cpp not empty. Move all #include "moc_<header_base>.cpp" lines from mocs_compilation.cpp to the cpp files of the related classes."
```
Swiftb0y, line 7:
```suggestion
#error mocs_compilation.cpp not empty. Move all #include "moc_<header_base>.cpp" lines from mocs_compilation.cpp to the cpp files of the related classes.
```

Swiftb0y (commented):
LGTM otherwise

Swiftb0y (approved):
LGTM. waiting for CI

daschuer:
@Swiftb0y all green, merge?

JoergAtGithub:
Worked for me locally with a slight build performance benefit. Thank you!

Issue #8948, opened 2022-08-23
Add option to mark history as "already played"


mixxxbot:
Reported by: **[LindyBalboa](https://github.com/LindyBalboa)**
Date: 2017-10-07T07:29:08Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1721929](https://bugs.launchpad.net/bugs/1721929)
Tags: easy, library

---

For dance DJs, it is a common occurence to play to the same crowd 2 or 3 days in a row. It would be useful to be able to mark all the tracks from the previous days set as "played". 

An alternative would be to make sessions explicit, saved, and loadable. However I think that would require more work to implement and the option to mark arbitrary tracks/playlists as already played is more atomic.

ronso0:
Fixed by #11574

PR #11574, opened 2023-05-21
Lib sidebar: right click selection fixes

ronso0:
follow-up for #11208 with small UX fixes

### DIFF
--- a/src/library/trackset/setlogfeature.cpp
+++ b/src/library/trackset/setlogfeature.cpp
@@ -591,7 +591,7 @@ void SetlogFeature::slotPlaylistTableChanged(int playlistId) {
     bool rootWasSelected = false;
     if (isChildIndexSelectedInSidebar(m_lastClickedIndex)) {
         // a child index was selected (actual playlist or YEAR item)
-        int lastClickedPlaylistId = m_pPlaylistTableModel->getPlaylist();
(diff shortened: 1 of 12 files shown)


ronso0:
Woo, done. I did a lot of testing and now the sidebar selection is kept if possible, and the sidebar selection and track view are always in sync.

Furthermore I added a "Mark tracks played" sidebar action to fix #8948 (for this and 'Join with previous' I added a new method to track.cpp).
I can also move the respective commits to a new branch if this is considered too much of a new feature for 2.4 (since we somehow agreed on feature freeze). I don't mind.

src/library/trackset/setlogfeature.cpp
```diff
@@ -66,6 +64,12 @@ SetlogFeature::SetlogFeature(
             this,
             &SetlogFeature::slotJoinWithPrevious);
 
+    m_pMarkTracksPlayedAction = new QAction(tr("Mark all tracks played)"), this);
```
daschuer, line 67:
```suggestion
    m_pMarkTracksPlayedAction = new QAction(tr("Mark all tracks played"), this);
```

daschuer (changes requested):
One extra ")" else LGTM. 
And works!

PR #11583, opened 2023-05-22
Remove from disk: shrink dialog so buttons are (hopefully) on screen

ronso0:
Make the "Remove track files from disk" dialog smaller. Track file list widget is now max 75% of screen.

That should suffice as quick fix. A more robust solution ([like for the preferences](https://github.com/mixxxdj/mixxx/blob/5f5cd28b3f8b471f5f2ca271e1ecacf8f2b1a659/src/preferences/dialog/dlgpreferences.cpp#L348)) is significantly more elaborate, because IIRC the dialog needs to be shown before we can query its size and window decoration.


uklotzde (approved):

daschuer (approved):
LGTM, thank you.

PR #11588, opened 2023-05-23
Pref Interface: update scale factor member with config value

### LINKED ISSUE #11586: False positive restart dialog after chaning unrelated settings with screen scale 125%
### Bug Description

This dialog should appear once, when the screen scaling has been reconfigured. 
Currently it is shown after each restart and changing the Main output in my case. 

### Version

2.4

### OS

Linux

ronso0:
fixes #11586


daschuer (approved):
It works. Thank you.

PR #11591, opened 2023-05-25
History: restore menu action "Remove from playlist"

### LINKED ISSUE #10974: Re-add "Remove" to History track menu
### Feature Description

I think the "Remove" action is really useful to clean up set lists, for example to remove
* tracks from the soundcheck after forgetting about 'Start new' (split history)
* played tracks that didn't make it into the main mix (recording) in a back-to-back session
* duplicate tracks that were not filtered

And it would comply with disallowing rewriting the history. 'Forgetting' events is different from creating new ones : )

ronso0:
The "Remove" action is really useful to clean up set lists, for example to remove

* tracks from the soundcheck after forgetting about 'Start new' (split history)
* played tracks that didn't make it into the main mix (recording) in a back-to-back session
* duplicate tracks that were not filtered

And it would comply with disallowing rewriting the history. 'Forgetting' events is different from creating new ones : )

Fixes #10974

### DIFF
--- a/src/library/playlisttablemodel.cpp
+++ b/src/library/playlisttablemodel.cpp
@@ -354,12 +354,11 @@ TrackModel::Capabilities PlaylistTableModel::getCapabilities() const {
     } else {
         caps |= Capability::Remove;
     }
-    if (PlaylistDAO::PLHT_SET_LOG ==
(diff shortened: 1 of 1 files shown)


daschuer (approved):
Looks and works good. Thank you.

Issue #11589, opened 2023-05-23
Segfault after quitting Mixxx in the "no sound device" selection dialog


daschuer:
### Bug Description

This has append during testing https://github.com/mixxxdj/mixxx/pull/10989

Here is the backtrace:
```
Thread 32 "BrowseThread" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fff9a68f700 (LWP 18421)]
0x00007ffff3992e5b in std::__atomic_base<QMutexData*>::compare_exchange_strong
    (__m2=std::memory_order_acquire, __m1=std::memory_order_acquire, __p2=0x1, 
    __p1=<synthetischer Zeiger>: 0x0, this=0x20)
    at /usr/include/c++/9/bits/atomic_base.h:769
769	      compare_exchange_strong(__pointer_type& __p1, __pointer_type __p2,
(gdb) 
(gdb) 
(gdb) bt
#0  0x00007ffff3992e5b in std::__atomic_base<QMutexData*>::compare_exchange_strong(QMutexData*&, QMutexData*, std::memory_order, std::memory_order)
    (__m2=std::memory_order_acquire, __m1=std::memory_order_acquire, __p2=0x1, __p1=<synthetischer Zeiger>: 0x0, this=0x20)
    at /usr/include/c++/9/bits/atomic_base.h:769
#1  std::atomic<QMutexData*>::compare_exchange_strong(QMutexData*&, QMutexData*, std::memory_order, std::memory_order)
    (__m2=std::memory_order_acquire, __m1=std::memory_order_acquire, __p2=0x1, __p1=<synthetischer Zeiger>: 0x0, this=0x20) at /usr/include/c++/9/atomic:565
#2  QAtomicOps<QMutexData*>::testAndSetAcquire<QMutexData*>(std::atomic<QMutexData*>&, QMutexData*, QMutexData*, QMutexData**)
    (currentValue=<synthetischer Zeiger>, newValue=0x1, expectedValue=0x0, _q_value=...) at ../../include/QtCore/../../src/corelib/thread/qatomic_cxx11.h:290
#3  QBasicAtomicPointer<QMutexData>::testAndSetAcquire(QMutexData*, QMutexData*, QMutexData*&)
    (currentValue=<synthetischer Zeiger>: <optimized out>, newValue=0x1, expectedValue=0x0, this=0x20)
    at ../../include/QtCore/../../src/corelib/thread/qbasicatomic.h:263
[...]

PR #11593, opened 2023-05-26
Stop browse thread before deleting the library

daschuer:
This will fix https://github.com/mixxxdj/mixxx/issues/11589

### DIFF
--- a/src/library/browse/browsetablemodel.cpp
+++ b/src/library/browse/browsetablemodel.cpp
@@ -474,8 +474,8 @@ QAbstractItemDelegate* BrowseTableModel::delegateForColumn(const int i, QObject*
     return nullptr;
 }
 
-void BrowseTableModel::stopBrowseThread() {
-    // The shared browse thread is actually stopped in the desctuctor
+void BrowseTableModel::releaseBrowseThread() {
+    // The shared browse thread is stopped in the destructor
     // if this is the last reference. All references must be reset before
     // the library is destructed.
     m_pBrowseThread.reset();
--- a/src/library/browse/browsetablemodel.cpp
+++ b/src/library/browse/browsetablemodel.cpp
@@ -475,5 +475,8 @@ QAbstractItemDelegate* BrowseTableModel::delegateForColumn(const int i, QObject*
 }
 
 void BrowseTableModel::stopBrowseThread() {
+    // The shared browse thread is actually stopped in the desctuctor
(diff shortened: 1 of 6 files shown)


ronso0:
IIUC this isn't stopping the BrowseThread, it just resets the BrowseThreadPointer to prevent further calls _from_ and callbacks _to_ the BrowseFeature.

Is real stopping required? (didn't understand the actual bug tbh)
If yes, that could be achieved via `m_bStopThread` like in the BrowseThread destructor.
If no, the new method should probably be renamed.

daschuer:
Yes, the code is not obvious. I have added a comment. 

The issue was that if the browse thread is using the library and when you quit Mixxx in the mean time from the main thread. It is using dangling pointers and crashes Mixxx. 
The fix is to stop these threads and wait until they are really terminated and then continue destructing the library. 
This is done here implicit by releasing all references to the browse thread (The last reset() calls immediately the destructor). 

Currently the recording feature and the browse feature are using this thread. The recording feature is destructed earlier because it is places in the widget tree of the skin. This asymmetry is not obvious and bites me during development, but a fix would involve a bigger refactoring. Probably required when moving to QML, but not in the stable 2.3 branch.

ronso0 (approved):
LGTM. Thanks for the QSharedPointer explanation.

src/library/browse/browsetablemodel.cpp
```diff
@@ -471,3 +473,10 @@ QAbstractItemDelegate* BrowseTableModel::delegateForColumn(const int i, QObject*
     }
     return nullptr;
 }
+
+void BrowseTableModel::stopBrowseThread() {
+    // The shared browse thread is actually stopped in the desctuctor
```
uklotzde, line 478: Typo

src/library/browse/browsetablemodel.cpp
```diff
@@ -471,3 +473,10 @@ QAbstractItemDelegate* BrowseTableModel::delegateForColumn(const int i, QObject*
     }
     return nullptr;
 }
+
+void BrowseTableModel::stopBrowseThread() {
```
uklotzde, line 477: The name of this function is inappropriate as the comment tells. It doesn't stop the browse thread, it just drops one of multiple references.

-> `resetBrowseThread()` or `releaseBrowseThread()`?

daschuer:
Done.

ronso0:
Thank you!

Issue #11600, opened 2023-05-29
Dark border around VU-Meter GLGS widgets in 200 % scale   


daschuer:
### Bug Description

Since https://github.com/mixxxdj/mixxx/pull/10989 the VU-Meter widgets have a black border. 

![grafik](https://github.com/mixxxdj/mixxx/assets/1777442/701b4647-183c-4759-b6cf-be9cce8cb9ea)

![grafik](https://github.com/mixxxdj/mixxx/assets/1777442/2bcf60a3-7599-4861-b83c-8c785de88f1d)

![grafik](https://github.com/mixxxdj/mixxx/assets/1777442/b0e85f08-9252-4861-b5c7-81264bef52bb)
![grafik](https://github.com/mixxxdj/mixxx/assets/1777442/1ad2e462-3b26-43e5-946a-83f36ca20250)

### Version

2.4-alpha

### OS

Ubuntu Focal

daschuer:
@m0dB do you have an idea? Maybe it is related to the issue that the widget thinks it is painted on black and we see antialiasing.

m0dB:
@daschuer it was something silly, in src/util/texture.cpp I was using 
```
pTexture->setWrapMode(QOpenGLTexture::ClampToBorder);
```
instead of
```
pTexture->setWrapMode(QOpenGLTexture::ClampToEdge);
```

What is the correct procedure to create a PR for a specific issue? Do I give the branch a specific name? Do I just mention the issue in the description?

ronso0:
> Do I just mention the issue in the description?

:+1: 
> Do I give the branch a specific name?

Whatever works best for you, though I find that handy while collaborating on a certain branch, it's just easier to remember and easier to spot in a `git branch` list.

daschuer:
I can confirm that the suggested change fixes the issue. Thank you.

daschuer:
I have added the fix here: https://github.com/mixxxdj/mixxx/pull/11601

JoergAtGithub:
Did https://github.com/mixxxdj/mixxx/pull/11722 fix this?

daschuer:
Yes, I can confirm it is fixed: 
![grafik](https://github.com/mixxxdj/mixxx/assets/1777442/554e5249-6c79-4cd0-8d92-e805c661e182)

daschuer:
This one is also back on Ubuntu Jammy :-(

[...]

PR #11598, opened 2023-05-29
Shade VU-Meter fix

daschuer:
See commits.

PR #11601, opened 2023-05-29
Shade: Audio Latency meter fix 

daschuer:
This fixes an issue with the GLSL Shader when the widget has only on pixel height. 
Simply by using two widgets. 

I have moved a +1 to various places in our code, but I was not able to make it work. Only if the Widget itself has a height of two pixels, it works. 

@m0dB do you have an idea how to fix it? 
Can you confirm the issue with macOS?


m0dB:
I confirm that on macOS the vumeter doesn't show either when the height is 1. The problem is not in the shader code, I tried by just simply making the window red with glClearColor and returning. With 2 pixels there is no problem (note btw that the size is determined by the background pixmap). Looks like a Qt bug to me.

I could implement a workaround: when the height is 1, make it two and set the glViewport to 1 pixel height.

Alternatively, I think it would be much better of the overload formed part of the vumeter, instead of using a separate statuslight for this. That would also solve the problem.

daschuer:
Thank you for confirming that it is nothing wrong in your code. Maybe one with a more recent QT version can confirm if it has been fixed with QT 5.15. 

A new widget that combined both indicators would be nice. However it affects the skins. Not suitable for a beta. It would be welcome for main though.  

Is the glViewport workaround easy enough or should we just merge this workaround here?

daschuer:
The tooltip is already fixed here: https://github.com/mixxxdj/mixxx/pull/11598

I had no luck with a transparent line in the background image, because black is shining through not gray. 
I will verify it again and also add the skin warning.

daschuer:
With the border halve transparent and 200 % it looks like this: 
grafik

daschuer:
With QOPENGL=OFF 
grafik

daschuer:
And finally with --disable-vumetergl
grafik

daschuer:
The issue is that in shade the background is a png, not a brush. The dark gray color is the backround color used before the background png is rendered. I don't think there is a big demand to fix that, we just need to look for another workaround here.

daschuer:
Using the background color instead of transparent works: 
grafik

daschuer:
Now it looks exactly like before. 
I have the fix for https://github.com/mixxxdj/mixxx/issues/11600 in my stash as well and added it here.

daschuer:
Superseded by https://github.com/mixxxdj/mixxx/pull/11722

daschuer:
I can confirm that this fixes the Shade issues on Ubuntu Jammy as well. 
@m0dB should we merge this and carry on?

m0dB:
LGTM! Merged. Thanks for taking care of this!

Issue #11094, opened 2022-11-27
Some SoundSourceProxyTests fail on Win11 with VS2022


JoergAtGithub:
### Bug Description

I build the same versions of Main ([9965d55](https://github.com/mixxxdj/mixxx/commit/9965d55f6e00cd1982f78cf09ac6196887869487)) and PR #11074 on two Windows laptops. On the older one with Win7 and VS2019 all tests passed.
But on the newer one with Win11 and VS2022 four tests failed:
```
SoundSourceProxyTest.seekForwardBackward 
SoundSourceProxyTest.seekBoundaries
SoundSourceProxyTest.readBeyondEnd
SoundSourceProxyTest.regressionTestCachingReaderChunkJumpForward
```

The test SoundSourceProxyTest.seekForwardBackward generated ~2GByte of similar log messages, here is shortened version of the log file:
[LastTest_soundsourceproxytestsfailed_newbuilenv.log](https://github.com/mixxxdj/mixxx/files/10098214/LastTest_soundsourceproxytestsfailed_newbuilenv.log)

### Version

Main

### OS

Windows11 + VS2022

uklotzde:
`SoundSourceMediaFoundation` is hardly working due to lack of Windows developers/testers and might fail when Microsoft decides to either change the API or behavior.

The API itself is not really suitable for sample-accurate decoding.

uklotzde:
https://www.reddit.com/r/DJs/comments/z9f7yf/comment/iygulke/?utm_source=share&utm_medium=web2x&context=3

JoergAtGithub:
I just updated to the latest Windows11 update - these tests still fail.

daschuer:
Without rebuilding?

JoergAtGithub:
I now copied all .exe and .dll files from the build directory of my Windows7 laptop to the build directory of my Windows11 laptopand ran CTest.
On Windows7 all tests with these executables were pass, on Windows11 the 4 SoundSourceProxyTests failed.

Conclusion: It's an unfixed OS bug in Windows11 and not a Mixxx bug.

JoergAtGithub:
These tests still fail, using latest Win11.

daschuer:
@JoergAtGithub: Does Windows 11 find the Sound Start Cue created in Windows < 11? Just play one or more m4a file you had in your Windows before the OS Update. 

and look for `First sound has been moved!`

Than please Build Mixxx with ffmpeg 
[...]

PR #4054, opened 2021-07-03
use hidraw backend of hidapi instead of libusb backend

Be-ing:
This has two advantages:
1. Bluetooth HID devices are supported. These are relatively rare, but we have had a request to support these before: https://mixxx.discourse.group/t/bluetooth-remote/16195/8
2. https://github.com/libusb/hidapi/issues/274 does not affect the hidraw backend, so the hacks around this bug can be removed for the Kontrol S4 Mk2 & Mk3 mappings.

The disadvantage is that hid_get_input_report for explicitly requesting a specific input report is [not yet implemented](https://github.com/libusb/hidapi/issues/259) for the hidraw backend. I will work on this upstream. That is used by #3038 but not in any mappings currently in the main branch.

PR #11605, opened 2023-05-29
2.4.0 changelog updates 

daschuer:
This should be merged before starting 2.4-beta

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -374,8 +374,8 @@
 ### Controller Backend
 
 * Never raise a fatal error if a controller mapping tries access a non-existent control object [#2947](https://github.com/mixxxdj/mixxx/pull/2947)
-* Add generic USB HID "Set Reports (Feature)" functionality [#3051](https://github.com/mixxxdj/mixxx/pull/3051)
-* Add support for reading the status of an HID controller (like MIDI SYSEX) [#3317](https://github.com/mixxxdj/mixxx/pull/3317)
+* Add support to access HID FeatureReports [#3051](https://github.com/mixxxdj/mixxx/pull/3051)
+* Add function to request HID InputReports, to determine controller state at startup [#3317](https://github.com/mixxxdj/mixxx/pull/3317)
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
Shouldn't we mention, the change of the hidapi backend on Linux? I could imagine that some Linux users might have issues with the udev file changes.

JoergAtGithub:
Could you please change
```
* Add generic USB HID "Set Reports (Feature)" functionality [#3051](https://github.com/mixxxdj/mixxx/pull/3051)
* Add support for reading the status of an HID controller (like MIDI SYSEX) [#3317](https://github.com/mixxxdj/mixxx/pull/3317)
```
to
```
* Add support to access HID FeatureReports [#3051](https://github.com/mixxxdj/mixxx/pull/3051)
* Add function to request HID InputReports, to determine controller state at startup [#3317](https://github.com/mixxxdj/mixxx/pull/3317)
```

daschuer:
> Use hidapi's hidraw backend instead of libusb on Linux https://github.com/mixxxdj/mixxx/pull/4054

We have that already. 

We have anyway the task to beautify the list before releasing 2.4.0. Any help is welcome.

JoergAtGithub:
What is about https://github.com/mixxxdj/mixxx/issues/11094 - even if the bug is introduced by a change in Windows, it's a major restriction. Do we need something like a known bugs list for the Beta?

daschuer:
I have added the suggested changes. 

Regarding https://github.com/mixxxdj/mixxx/issues/11094 we have no list of common issues other than our bug tracker. 
Here we can use the regression tag to make them discoverable. I am also in doubt it is reasonable to track an extra list. 

If we are not able to find a fix for issue #11094, we may enable FFMPEG instead as suggested. Do you have interest to take care. First I would take a look if we are doing something wrong when seeking. Do we handle all error situations correctly.

daschuer:
Can we merge this one now and advance to 2.4-beta?

JoergAtGithub:
LGTM! Thank you for takeing care!

PR #11607, opened 2023-05-29
Fix Windows warnings and enable "Treat all compiler warnings as errors"

JoergAtGithub:
This PR contains fixes of warnings, which occured on Windows, either with MSVC or ClangCL or both.


CMakeLists.txt
```diff
@@ -2201,6 +2202,7 @@ add_library(FpClassify STATIC EXCLUDE_FROM_ALL src/util/fpclassify.cpp)
 
 if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_SIMULATE_ID MATCHES "MSVC")
   target_compile_options(FpClassify PRIVATE /fp:precise)
+  target_compile_options(mixxx-lib PRIVATE -Wno-microsoft-cast)
```
daschuer, line 2205: What does "-Wno-microsoft-cast" do? Normally it would be better to fix the warnings.

CMakeLists.txt
```diff
@@ -2233,7 +2235,7 @@ add_library(MP3GuessEnc STATIC EXCLUDE_FROM_ALL
   lib/mp3guessenc-0.27.4/bit_utils.c
 )
 if(WIN32)
-  target_compile_definitions(MP3GuessEnc PRIVATE __WINDOWS__)
+  target_compile_definitions(MP3GuessEnc PRIVATE __WINDOWS__ _CRT_SECURE_NO_WARNINGS)
```
daschuer, line 2238: The same here. Is it possible to disable the warning conditional near the effected code?

lib/benchmark/src/benchmark.cc
```diff
@@ -390,8 +390,6 @@ std::unique_ptr<BenchmarkReporter> CreateReporter(
     return PtrType(new ConsoleReporter(output_opts));
   } else if (name == "json") {
     return PtrType(new JSONReporter);
-  } else if (name == "csv") {
-    return PtrType(new CSVReporter);
```
daschuer, line 394: I don't think we have yet patched benchmark.cc. Did you check if this is still an issue in v1.8.0?

daschuer (changes requested):
Thank you for taking care. I have added some comments.

CMakeLists.txt
JoergAtGithub: MSVC an GCC compiler allows it to cast an object pointer to a function pointer. This is not described in the C++ standard.
Clang warns about this non standard compliant cast, but executes it as the other compilers. This warning is inhibited hereby.

CMakeLists.txt
JoergAtGithub: We use this already for our own code. It inhibits the MSVC security warnings when you use string functions without the _s suffix. A C++ standard extension which only Microsoft implemented, because all the other compiler developer  do not agree, that these functions improve the code security.
With _CRT_SECURE_NO_WARNINGS defined, MSVC behaves like gcc and clang.

lib/benchmark/src/benchmark.cc
JoergAtGithub: No, but as war as I can see, we doesen't use this deprecated output format. Do you see a problem to remove this ununsed code lines?

CMakeLists.txt
daschuer: Do we use that conversion?

CMakeLists.txt
daschuer: OK, thank you.

lib/benchmark/src/benchmark.cc
daschuer: No, but it is a maintenance burden whenever we upgrade the library. Did you consider to just remove the warning in this external library?

CMakeLists.txt
JoergAtGithub: Yes, that's why we got the warnings: https://github.com/mixxxdj/mixxx/blob/c3b808cf90e124d25f324c784c8887c77f2bd8fd/src/encoder/encoderfdkaac.cpp#L134-L139

daschuer: Ah ok, in this case we have a function pointer. I think we can avoid the warning by a proper explicit cast to void*, or a reinterpret cast. 

At this point we are also not interested in the actual address. We want to see which of the pointers is NULL. 
So we may change the debugging to explicit report that.

CMakeLists.txt
```diff
@@ -1293,7 +1293,8 @@ if(MSVC)
   if(WARNINGS_PEDANTIC)
     target_compile_options(mixxx-lib PUBLIC /W4)
   else()
-    target_compile_options(mixxx-lib PUBLIC /W3)
+    # Warning Level 3 (production quality) / Treat Warnings as Errors
+    target_compile_options(mixxx-lib PUBLIC /W3 /WX)
```
daschuer, line 1297: We already have the /WX flag, enabled by the WARNINGS_FATAL switch. The reason is that this allows to try out things without being forced to fix all warnings during development. The Flag is set in case of Ubuntu in the CI that prevents that any warning slips into the code. This is very convenient. 
So instead adding the fag here, please enable WARNINGS_FATAL in .github/workflow/build.yml

daschuer (changes requested):

CMakeLists.txt
JoergAtGithub: But than you don't notice it in local builds. It's very inconvinient if the local build passes and CI fails.

daschuer: Remember that warnings can also happen after a compiler update or if you use a different compiler than the CI.  
My IDE shows me the warnings inline in the code. It is hard to miss that. But you can anyway configure WARNINGS_FATAL as you like. I thing we should stick with that approach for all platforms.

daschuer (approved):
LGTM, thank you.

PR #11608, opened 2023-05-29
Build with Qt6 and optionally with QML

m0dB:
With this PR mixxx with its current skin-based UI can be built with Qt 6. Building with Qt 6 + QML is still possible, but the cmake option QT6 now builds the skin-based UI with Qt6, while the combination of cmake options QT6 and QML build the QML-based UI.

I have been using the official Qt 6.5 binary installer:

```
cmake .. -DCMAKE_PREFIX_PATH=~/Qt6/6.5.0/macos/ -DCMAKE_BUILD_TYPE=Release -DQT6=ON -DQTKEYCHAIN=OFF
```

As far as I see, after some smoke testing on macOS, everything works as expected except for the following:

- The splitter below the waveforms doesn't work. I have done some debugging and found that the size policy of the child widgets changes somewhere between when the skin is read in the legacy skin parser and when I try to move the splitter. If I force the size policy back to it's original value it works.

- The Latenight knobs look slightly different (more pronounced outline?), but nothing major.

- I get a lot of the following warning: warning [Main] skipping QEventPoint(id=1 ts=0 pos=0,0 scn=599.994,259.461 gbl=599.994,259.461 Released ellipse=(1x1 ∡ 0) vel=0,0 press=-599.994,-259.461 last=-599.994,-259.461 Δ 599.994,259.461) : no target window

Also during compilation there are still a bunch of "deprecated" warnings for the Qt API. These should be easy to fix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,10 +109,19 @@ if(NOT CMAKE_CONFIGURATION_TYPES)
 endif()
 
 option(QT6 "Build with Qt6" OFF)
-
+option(QML "Build with QML" OFF)
 option(QOPENGL "Use QOpenGLWindow based widget instead of QGLWidget" ON)
+
+if(QML AND NOT QT6)
+  message(FATAL_ERROR "Building with option QML=ON requires QT6=ON")
+endif()
+
 if(QOPENGL)
   add_compile_definitions(MIXXX_USE_QOPENGL)
+endif()
+
+if(QML)
+  add_compile_definitions(MIXXX_USE_QML)
 endif()
 
 if(APPLE)
@@ -120,7 +129,7 @@ if(APPLE)
     # Minimum macOS version supported by Qt 6
     set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum macOS version the build will be able to run on")
     if(NOT VCPKG_TARGET_TRIPLET)
-      set(VCPKG_TARGET_TRIPLET "x64-osx-min10.15")
+      set(VCPKG_TARGET_TRIPLET "x64-osx-min1012")
     endif()
   else()
     if(VCPKG_TARGET_TRIPLET STREQUAL "arm64-osx-min1100")
@@ -1103,7 +1112,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/widget/wwidgetgroup.cpp
   src/widget/wwidgetstack.cpp
 )
-if(QT6)
+if(QML)
   target_sources(mixxx-lib PRIVATE
     src/qml/asyncimageprovider.cpp
     src/qml/qmlapplication.cpp
@@ -2038,6 +2047,8 @@ if(ENGINEPRIME)
     message(STATUS "Using existing system installation of libdjinterop")
     target_include_directories(mixxx-lib PUBLIC ${DjInterop_INCLUDE_DIRS})
     target_link_libraries(mixxx-lib PRIVATE DjInterop::DjInterop)
+    find_package(ZLIB 1.2.8 REQUIRED)
+    target_link_libraries(mixxx-lib PRIVATE ${ZLIB_LIBRARIES})
   else()
     # Fetch djinterop sources from GitHub and build them statically.
 
@@ -2318,7 +2329,9 @@ if(QT6)
   foreach(COMPONENT ${QT6_NEW_COMPONENTS})
     target_link_libraries(mixxx-lib PUBLIC Qt6::${COMPONENT})
   endforeach()
+endif()
 
+if(QML)
   set(QT_QML_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/qml)
   set_target_properties(mixxx-lib PROPERTIES AUTOMOC ON)
   qt_add_qml_module(mixxx-lib
@@ -2416,7 +2429,7 @@ endif()
 
 
 if(APPLE)
-  if(Qt_IS_STATIC)
+  if(Qt_IS_STATIC OR QT6)
     target_link_libraries(mixxx-lib PRIVATE
         "-weak_framework Accelerate"
         "-weak_framework AppKit"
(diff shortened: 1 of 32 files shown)


res/linux/org.mixxx.Mixxx.metainfo.xml
```diff
@@ -1907,7 +1907,7 @@
  </ul>
 </description>
     </release>
-    <release version="2.3.6" type="development" date="2023-05-29" timestamp="1685366841">
+    <release version="2.3.6" type="development" date="2023-05-19" timestamp="1684533240">
```
JoergAtGithub, line 1910: The changes in this file are clearly unrelated

m0dB: Yeah, pre-commit made this change.

daschuer: This was a failed merge conflict resolve. 
In this case it is reasonable to remove this change via rebase.

m0dB: Got it. Will do this tomorrow.

daschuer: Thank you.

CMakeLists.txt
```diff
@@ -2038,6 +2047,8 @@ if(ENGINEPRIME)
     message(STATUS "Using existing system installation of libdjinterop")
     target_include_directories(mixxx-lib PUBLIC ${DjInterop_INCLUDE_DIRS})
     target_link_libraries(mixxx-lib PRIVATE DjInterop::DjInterop)
+    find_package(ZLIB 1.2.8 REQUIRED)
+    target_link_libraries(mixxx-lib PRIVATE ${ZLIB_LIBRARIES})
```
daschuer, line 2051: Does this change relate to Qt6 or is it something we need in the 2.4 branch anyway?

CMakeLists.txt
```diff
@@ -109,18 +109,27 @@ if(NOT CMAKE_CONFIGURATION_TYPES)
...
+  add_compile_definitions(MIXXX_USE_QML)
+endif()
+
 if(APPLE)
   if(QT6)
     # Minimum macOS version supported by Qt 6
     set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum macOS version the build will be able to run on")
     if(NOT VCPKG_TARGET_TRIPLET)
-      set(VCPKG_TARGET_TRIPLET "x64-osx-min10.15")
+      set(VCPKG_TARGET_TRIPLET "x64-osx-min1012")
```
daschuer, line 132: How does this change relate to the comment above?

CMakeLists.txt
m0dB: Qt6 only, and possibly because I am using the official Qt6 binary install. I suppose we should use a vcpkg buildenv with Qt6 instead?

daschuer: Yes. I will take a look to provide such a package.

CMakeLists.txt
daschuer: Ah ok. This is probanly a leftover anyway because we have removed the dot some time ago. I have no objections to merge this as it is and adjust it later on.

CMakeLists.txt
m0dB: Btw, with the current buildenv I get this warning for all Qt libraries when linking: warning: dylib (/Users/m0dB/Qt6/6.5.0/macos/lib/QtPrintSupport.framework/Versions/A/QtPrintSupport) was built for newer macOS version (11.0) than being linked (10.15)

res/linux/org.mixxx.Mixxx.metainfo.xml
m0dB: Done

m0dB:
I have rebased on main to remove the irrelevant commit and I have also found and fixed the splitter issue. This was in fact a bug in the Latenight skin but apparently Qt 5 deals with conflicting size policies differently than Qt 6.

ywwg:
https://github.com/m0dB/mixxx/pull/5 ?

m0dB:
Sorry, I didn't see your PR. Thanks. Merged!

ywwg:
the ubuntu 22.04 build error just looks like complaints about float/double conversion. it should be easy to add a static_cast to fix that

ywwg:
silencing the conversion warnings / errors: https://github.com/m0dB/mixxx/pull/6

daschuer:
Here is a QT6 build of the vcpkg environment: 
https://github.com/daschuer/vcpkg/actions/runs/5125452456

daschuer:
I can confirm that this runs flawlessly even on my ages Ubuntu Focal using Qt 6.2.3 from: 
https://launchpad.net/~daschuer/+archive/ubuntu/qt6-backports

CMakeLists.txt
daschuer: You use the upstream distribution of QT, right? Since Qt 6.5 they build by default with a minimum of macOS 11.00
without a technical reason https://github.com/qt/qtbase/commit/f2415df24de6860048bd0dd4c3e3ec40f867ee53, while hey still allow to build for 10.15 
https://github.com/qt/qtbase/blob/38c8eb8564495b0fe345eaca700966889f1f8aa3/cmake/QtBuild.cmake#L235

If we we do not use 11.0 only API,  I think we should not drop 10.15 in our own build artificially.

daschuer: Oh, I just noticed that it the minimum requirement is macOS 10.14 
https://github.com/qt/qtbase/commit/4933a5f89231afd64b1604d47220e1ff34d2c448

My build confirms it: 
https://github.com/daschuer/vcpkg/actions/runs/5125452456

m0dB: I agree and it's a good reason to use Qt6 from our own buildenv.

CMakeLists.txt
```diff
@@ -2520,7 +2533,11 @@ endif()
 if(APPLE OR WIN32)
   # qt_de.qm is just one arbitrary file in the directory that needs to be located;
   # there is no particular reason to look for this file versus any other one in the directory.
-  find_file(QT_TRANSLATION_FILE qt_de.qm PATHS "${Qt5_DIR}/../../../translations" "${Qt5_DIR}/../../qt5/translations" REQUIRED NO_DEFAULT_PATH)
+  if(QT6)
+    find_file(QT_TRANSLATION_FILE qt_de.qm PATHS "${Qt6_DIR}/../../../translations" "${Qt6_DIR}/../../qt5/translations" REQUIRED NO_DEFAULT_PATH)
```
daschuer, line 2537: This works for me: 
```suggestion
    find_file(QT_TRANSLATION_FILE qt_de.qm PATHS "${Qt6_DIR}/../../translations/Qt6" REQUIRED NO_DEFAULT_PATH)
```

daschuer (changes requested):

uklotzde:
It would be beneficial for early build testing if we could at least compile 2.4 with Qt6.

Is it mandatory to base these changes on `main`? Since the changes are hidden behind a build flag rebasing them should not cause any issues.

daschuer:
@uklotzde Please use the main branch for Qt 6 testing, this will be soon our new 2.5-beta after August 08, 2023 the 2.4 release date and we need a good amount of developer testing until than. For now it sound to me that 2.5 will be our first Qt 6 release. Let's make it real.  
I have no interest that distros starting to enable Qt6 for 2.4 creating basically untested builds, as it has happened when switching to Qt 5.

ywwg:
finding a couple more build issues, updating my pr

m0dB:
Thanks, @ywwg , LGTM but before adding your PR to mine, I'd like to ask @daschuer if he prefers one larger PR to merge to main that contains all these changes or to do it in smaller steps.

daschuer:
Just merge https://github.com/m0dB/mixxx/pull/6 and than I will merge this one.

ywwg:
all builds pass so my PR is ready to merge into this one

ywwg:
@daschuer #6 is merged

daschuer (approved):

daschuer:
Thank you. LGTM. 
The next one is merging this: https://github.com/mixxxdj/vcpkg/pull/71 before we can finally switch the 2.5 workflows to Qt6. Who can have a look?

PR #11610, opened 2023-05-29
chore pin actions/labeler to v4.0.4

Swiftb0y:
Closes #11609

Should prevent future frequent nagging from dependabot.


daschuer (approved):
Thank you. LGTM

PR #11619, opened 2023-06-03
Improve class definition layout of new classes 

daschuer:
This adjust the layout of the recently introduced class definitions with the rest of Mixxx and our style guide.

### DIFF
--- a/src/widget/tooltipqopengl.cpp
+++ b/src/widget/tooltipqopengl.cpp
@@ -1,7 +1,5 @@
 #include "widget/tooltipqopengl.h"
 
-#include <widget/wglwidget.h>
-
 #include <QStyle>
 #include <QTimer>
 #include <QToolTip>
--- a/src/widget/tooltipqopengl.cpp
+++ b/src/widget/tooltipqopengl.cpp
@@ -7,14 +7,18 @@
 #include <QToolTip>
 #include <memory>
 
-ToolTipQOpenGL::ToolTipQOpenGL() {
+#include "widget/wglwidget.h"
+
+ToolTipQOpenGL::ToolTipQOpenGL()
+        : m_active(true),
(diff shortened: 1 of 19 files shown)


src/widget/tooltipqopengl.cpp
```diff
@@ -7,14 +7,18 @@
 #include <QToolTip>
 #include <memory>
 
-ToolTipQOpenGL::ToolTipQOpenGL() {
+#include "widget/wglwidget.h"
```
Swiftb0y, line 10: this is already included in the form of `<widget/wglwidget.h>` on line 3... I guess remove that one then?

src/widget/tooltipqopengl.cpp
```diff
@@ -7,14 +7,18 @@
...
+
+ToolTipQOpenGL::ToolTipQOpenGL()
+        : m_active(true),
+          m_pWidget(nullptr) {
```
Swiftb0y, line 14: the tooltip is active without a widget? That doesn't seem right.

src/widget/tooltipqopengl.cpp
daschuer: Removed.

src/widget/tooltipqopengl.cpp
daschuer: The actual value does not matter because it is adjusted during MixxxMainWindow::initialize();

daschuer:
Done.

src/widget/tooltipqopengl.cpp
Swiftb0y: Ah I misinterpreted the meaning of the variable nevermind.

Swiftb0y (approved):
LGTM.

PR #11622, opened 2023-06-04
common controller scripts es7 silence

Swiftb0y:
#11284 is suffering from a bunch of semi-bogus eslint issues. This somewhat fixes them. The most important change in the first commit, where we reduce the scope of the typescript-eslint to only apply to typescript definition files. The second commit then runs the automatic changes for eslint and adds the appropriate `eslint-disable no-var` where appropriate.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,7 +1,6 @@
 {
   "extends": [ "eslint:recommended",
-               "plugin:jsdoc/recommended",
-               "plugin:@typescript-eslint/recommended" ],
+               "plugin:jsdoc/recommended"],
 
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
I just tried an alternative for the no-unused-vars:
```
 "no-unused-vars": "off",
    "@typescript-eslint/no-unused-vars": [
      "error",
      {
        "argsIgnorePattern": "^_"
      }
    ],
```
which for me only printed the really unused vars.
see: https://typescript-eslint.io/rules/no-unused-vars/

Not sure what is the better solution.

Swiftb0y:
that sorta works too, but that still leaves the eslint-typescript with linting our ES3 files. Moreover, the `eslint-typescript:recommended` config disables some of the eslint rules (such as no-unused-vars), because `tsc` already lints them by default. Unfortunately, the config for those doesn't carry over, which is why it creates bogus warnings again.

I think only applying the eslint-typescript rules to our `d.ts` files is the best solution for until someone figures out how to do this better.

JoergAtGithub:
LGTM! Thank you!

Swiftb0y:
Thank you

PR #11623, opened 2023-06-04
controller_mapping_validation_test: Check for manual link instead of wiki link

JoergAtGithub:
Controller documentation is moved

### DIFF
--- a/src/controllers/legacycontrollermapping.h
+++ b/src/controllers/legacycontrollermapping.h
@@ -138,12 +138,12 @@ class LegacyControllerMapping {
         return MIXXX_MANUAL_CONTROLLERMANUAL_PREFIX + page + MIXXX_MANUAL_CONTROLLERMANUAL_SUFFIX;
     }
 
-    inline void setWikiLink(const QString& wikilink) {
-        m_manuallink = wikilink;
+    inline void setManualLink(const QString& manuallink) {
+        m_manuallink = manuallink;
         setDirty(true);
     }
 
-    inline QString wikilink() const {
+    inline QString manuallink() const {
         return m_manuallink;
     }
 
--- a/src/controllers/legacycontrollermapping.h
+++ b/src/controllers/legacycontrollermapping.h
@@ -139,12 +139,12 @@ class LegacyControllerMapping {
     }
 
     inline void setWikiLink(const QString& wikilink) {
-        m_wikilink = wikilink;
+        m_manuallink = wikilink;
         setDirty(true);
     }
 
     inline QString wikilink() const {
-        return m_wikilink;
+        return m_manuallink;
     }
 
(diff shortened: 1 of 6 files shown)


src/controllers/legacycontrollermapping.h
```diff
@@ -139,12 +139,12 @@ class LegacyControllerMapping {
...
     inline void setWikiLink(const QString& wikilink) {
-        m_wikilink = wikilink;
+        m_manuallink = wikilink;
         setDirty(true);
     }
 
     inline QString wikilink() const {
-        return m_wikilink;
+        return m_manuallink;
     }
```
Swiftb0y, line 149: the XML element is still called `wiki` though... so this is essentially an API break and confusing at the same time. I think this refactoring needs to a little deeper. Can you also mass reformat all the controller XMLs to have a manual link instead of a wiki link?

daschuer: If we update the controller API this becomes a main PR. 

Here is also a missing bit: 
https://github.com/mixxxdj/mixxx/blob/b52b6d476ff7925b85e79d732a90bbb6dc599cb8/src/controllers/dlgprefcontroller.cpp#L341

Swiftb0y: sure... its only linting anyways. And the lint is a little questionable as well since nobody actually checks it.

daschuer (changes requested):
Is this a main PR?

JoergAtGithub:
The mappings itself were mass converted from wiki to manual already in 2.3. Now warnings appear for every single mapping.

src/controllers/legacycontrollermapping.h
JoergAtGithub: Done!

src/test/controller_mapping_validation_test.cpp
```diff
@@ -93,10 +93,10 @@ bool lintMappingInfo(const MappingInfo& mapping) {
...
     }
 
-    if (mapping.getWikiLink().trimmed().isEmpty()) {
-        qWarning() << "LINT:" << mapping.getPath() << "has no wiki link.";
-    } else if (!checkUrl(mapping.getWikiLink())) {
-        qWarning() << "LINT:" << mapping.getPath() << "has invalid wiki link";
+    if (mapping.getManualLink().trimmed().isEmpty()) {
+        qWarning() << "LINT:" << mapping.getPath() << "has no manual link.";
+    } else if (!checkUrl(mapping.getManualLink())) {
+        qWarning() << "LINT:" << mapping.getPath() << "has invalid manual link";
```
Swiftb0y, line 99: checking a couple mapping XMLs, the tag in question doesn't actually contain a link, just a name for a page. Am I missing something?

PR #11625, opened 2023-06-04
common-controller-scripts.js: Convert "script" to freezed object, and remove global "controller", "deck", "control", "button" definitions

JoergAtGithub:
- Silenced eslint warnings vor unused vars
- Use `console.log` instead of deprecated `print` function
- Made global functions const.


daschuer (approved):
LGTM, However I am unable to test this. 
Can we risk to merge it and see what happens during the remaining 5 Weeks beta?

JoergAtGithub:
I think it's unecessary risk for 2.4 now.

daschuer:
Oh sorry this one slipped through. Can you fix the conflicts and than we can merge?

res/controllers/common-controller-scripts.js
```diff
@@ -158,511 +487,205 @@ script.LIBRARY_COLUMNS = Object.freeze({
...
             engine.setValue(group, this.mappedFunction, outputValue);
         }
-    } else {
-        engine.setValue(group, this.mappedFunction, outputValue);
     }
-};
+}
+// Add class Control to the Global JavaScript object
+// @ts-ignore Same identifier for class and instance needed for backward compatibility
+this.Control = Control;
```
Swiftb0y, line 672: I don't know if `this` in the global context works. I think `globalThis` is standardized.

JoergAtGithub: `this` works here and refers the `globalThis`.

res/controllers/common-controller-scripts.js
```diff
@@ -109,24 +108,354 @@ var msecondstominutes = function(msecs) {
...
     };
 };
 
-/* eslint @typescript-eslint/no-empty-function: "off" */
-var script = function() {
-};
+
+
+// @ts-ignore Same identifier for class and instance needed for backward compatibility
+class script {
```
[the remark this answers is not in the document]
JoergAtGithub: It groups the static members into their namespace and makes the code much easier readable, while preventing warnings. I think such Utility-Classes are a common programming pattern.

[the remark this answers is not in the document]
Swiftb0y: right, but why doesn't the less verbose object syntax fullfil that same requirement?
eg
```
const script = {
    samplerRegEx: ...,
    debug: function () {...}
}

File: res/controllers/common-controller-scripts.js (hunk 1 above)
Comment (Swiftb0y) [reply res/controllers/common-controller-scripts.js:671]: good to know, that assumption probably won't hold for ES modules though. but we can revisit that topic when esm-based mappings are a thing.

File: res/controllers/common-controller-scripts.js (hunk 2 above)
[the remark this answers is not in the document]
Comment (JoergAtGithub) [reply res/controllers/common-controller-scripts.js:129]: This introduces another code style, where we could use the same as we use for non-static classes. It's of cause a matter of taste, but I prefer to use as less different styles as possible.
And, in my opinion, the class syntax is easier understandable.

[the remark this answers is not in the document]
Comment (Swiftb0y) [reply res/controllers/common-controller-scripts.js:129]: right. I think it there is value in separating the two styles however. Things that should be instantiated should be implemented as classes, while things that should not be instantiated (such as namespaces) should be declared as objects. Using different styles consistently would make it easier to understand how to use it. 
Currently, when I read the `script`-class I need to scan all the way through the class that it only contains static methods and no constructor. If I saw that `script` was just an object, I would immediately understand that its just a collection of variables (which more closely resembles the definition of a namespace). Does that make sense?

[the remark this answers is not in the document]
Comment (Swiftb0y) [reply res/controllers/common-controller-scripts.js:129]: For that same reason, I don't think that this style where a class only contains static members is very popular. If I knew this was an established style, I would probably accept it, but I have never seen this previously, which is where my skepticism comes from.

**Comment by daschuer** [comment]
Is this still a 2.5-beata PR?

**Comment by JoergAtGithub** [comment]
Don't wait for it! I will try.

File: res/controllers/common-controller-scripts.js [hunk 3]
```diff
@@ -109,560 +108,594 @@ var msecondstominutes = function(msecs) {
...
-    if (bpm.tap.length > 8) { bpm.tap.shift(); }
-    let sum = 0;
-    for (let i=0; i<bpm.tap.length; i++) {
-        sum += bpm.tap[i];
-    }
-    const average = sum / bpm.tap.length;
+}
+// Add class bpm to the Global JavaScript object
+// @ts-ignore Same identifier for class and instance needed for backward compatibility
+this.bpm = bpm;
```
Swiftb0y, line 571: since `bpm` was effectively a singleton previously and now its a class with state, shouldn't this call the constructor of that class instead of just aliasing it?

Swiftb0y (commented):
thank you for changing the script namespace. No complaints apart from the single comment

JoergAtGithub:
There seems to be a naming collision between Deck here and Deck in midi-components-0.0.js . But this occurs even if I revert the Deck code here to the original implementation.

Swiftb0y:
I see. Well I don't think Deck, Controller and Button from common-controller-scripts are used very much (also undocumented on the wiki). Can we identify which mappings use them and namespace them now? I know its an API break, but I think its fairly low impact. Unless you have a better idea on how to fix it though. Do you have a stacktrace or something that gives more insight into the underlying issue?

JoergAtGithub:
I've not fully understood it, but I agree that these are useless stubs.

Swiftb0y:
Interesting that they actually were there from the beginning. https://github.com/mixxxdj/mixxx/commit/8d047cdcbc9

Not sure if they ever saw much use. Seems like the American-Audio VMS2 & VMS4, Hercules RMX scripts use them. We may consider simply inlining these stubs there are removing them from common-controller-scripts?

JoergAtGithub:
The main issue seems to be, that Reloop-Jockey-3-ME-scripts.js:25 redefines script.crossfaderCurve which is now constant. This is why the last change broke the test.

The Deck naming errors were always there, and do not break the test.

Swiftb0y:
good catch. Anything left to do then before we can merge?

JoergAtGithub:
> Interesting that they actually were there from the beginning. [8d047cdcbc9](https://github.com/mixxxdj/mixxx/commit/8d047cdcbc9)
> 
> Not sure if they ever saw much use. Seems like the American-Audio VMS2 & VMS4, Hercules RMX scripts use them. We may consider simply inlining these stubs there are removing them from common-controller-scripts?

The following mapping tests fail if I remove them:
grafik

Swiftb0y:
Thats weird, I can't tell why.

res/controllers/common-controller-scripts.js
```diff
@@ -109,561 +102,464 @@ var msecondstominutes = function(msecs) {
...
+        fRateScale = (fRateScale - 1.) / engine.getValue(group, "rateRange");
+
+        engine.setValue(
+            group, "rate",
+            fRateScale * engine.getValue(group, "rate_dir"));
+    }
+}
+// Add class bpm to the Global JavaScript object
+// @ts-ignore Same identifier for class and instance needed for backward compatibility
+this.bpm = new bpm();
```
Swiftb0y, line 565: wdyt of naming the class differently so it can be used in newer code? Currently mappings are forced to use the "static" API because it shadows the class.

res/controllers/common-controller-scripts.js
```diff
@@ -109,561 +102,464 @@ var msecondstominutes = function(msecs) {
...
-    if (group.substring(2, 8) === "hannel") {
-        // Extract deck number from the group text
-        deck = group.substring(8, group.length - 1);
-    }
-    /*
-        else if (group.substring(2,8)=="ampler") {
-            // Extract sampler number from the group text
-            deck = group.substring(8,group.length-1);
+const script = {
+    // ----------------- Mapping script constants ---------------------
```
Swiftb0y, line 121: should we also `Object.freeze` the entire namespace?

res/controllers/common-controller-scripts.js
```diff
@@ -109,561 +102,463 @@ var msecondstominutes = function(msecs) {
...
+    spinback: function(channel, control, value, status, group, factor, rate) {
+        // if brake is called without defined factor and rate, reset to defaults
+        if (factor === undefined) {
+            factor = 1;
+        }
+        // if brake is called without defined rate, reset to default
+        if (rate === undefined) {
+            rate = -10;
+        }
+        // disable on note-off or zero value note/cc
```
Swiftb0y, line 447: does our version of js support default parameters?

```suggestion
    spinback: function(channel, control, value, status, group, factor = 1, rate = -10) {
        // disable on note-off or zero value note/cc
```

res/controllers/common-controller-scripts.js
```diff
@@ -109,561 +102,463 @@ var msecondstominutes = function(msecs) {
...
+                from xml for a brake effect
+                e.g: <key>script.brake</key>
+       Input:   channel, control, value, status, group, factor (optional)
+       Output:  none
+       -------- ------------------------------------------------------ */
+    brake: function(channel, control, value, status, group, factor) {
+        // if brake is called without factor defined, reset to default
+        if (factor === undefined) {
+            factor = 1;
+        }
```
Swiftb0y, line 465:
```suggestion
    brake: function(channel, control, value, status, group, factor = 1) {
```

res/controllers/common-controller-scripts.js
```diff
@@ -109,561 +102,463 @@ var msecondstominutes = function(msecs) {
...
+                defined by pitch slider, can also interrupt engine.brake()
+                e.g: <key>script.softStart</key>
+       Input:   channel, control, value, status, group, acceleration factor (optional)
+       Output:  none
+       -------- ------------------------------------------------------ */
+    softStart: function(channel, control, value, status, group, factor) {
+        // if softStart is called without factor defined, reset to default
+        if (factor === undefined) {
+            factor = 1;
+        }
```
Swiftb0y, line 485:
```suggestion
    softStart: function(channel, control, value, status, group, factor = 1) {
```

Swiftb0y (approved):
LGTM. just a couple cleanup opportunities in case you have the nerve for them. Also please rebase in order to pull in #12779. Thank you

JoergAtGithub:
I struggle with merging the changes from #12779. The validation tests of mappings with midi-components now fail with:
`Uncaught exception: file:///D:/mixxx/res/controllers/midi-components-0.0.js:664: ReferenceError: script is not defined\nBacktrace: applyLayer@file:///D:/mixxx/res/controllers/midi-components-0.0.js:664\nComponentContainer@file:///D:/mixxx/res/controllers/midi-components-0.0.js:523\n@file:///D:/mixxx/res/controllers/midi-components-0.0.js:698\n%entry@file:///D:/mixxx/res/controllers/midi-components-0.0.js:29`

JoergAtGithub:
> not sure tbh. Maybe try doing reverting back to `var script...` again?

Doesn't help.

res/controllers/common-controller-scripts.js
```diff
@@ -121,616 +102,513 @@ var msecondstominutes = function(msecs) {
...
+                if (
+                    Array.isArray(target[key]) && Array.isArray(source[key]) ||
+                    this.isSimpleObject(target[key]) && this.isSimpleObject(source[key])
+                ) {
+                    this.deepMerge(target[key], source[key]);
+                } else if (source[key] !== undefined && source[key] !== null) {
+                    Object.assign(target, {[key]: source[key]});
+                }
+            });
+        }
```
Swiftb0y, line 205: as this is not a class, `this` is not `script` but instead the function object. You'll have to call `script.deepMerge/isSimpleObject` explicitly.

Swiftb0y (commented):
not sure if thats the culprit of the CI failure though.

res/controllers/common-controller-scripts.js
JoergAtGithub: I tried both, no difference in the error message.

Swiftb0y: interesting. Well either way, `this` is wrong.

daschuer:
CI fails. I like to move this either to 2.5.0 or 2.6-alpha milestone. Does this have an impact on mapping interface regarding documentation?

JoergAtGithub:
No impact on documentation. I currently don't know how to fix this.

christophehenry:
Let's wait for the end of the CI but I believe [ac4ab33](https://github.com/mixxxdj/mixxx/pull/13477/commits/ac4ab335b5831f7d1e2a885c7d014b04e3e8c2bd) solves this CI error. At least it solves it locally here. I think the error comes from that [*const declarations do not create properties on [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) when declared at the top level of a script.*](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const) so `const script` won't dump `script` into the global namespace. I eclosed the whole code into a `(function(global) { … }(this));` to solve this. I also change `script`'s function definitions to ES6 methods synctatic sugar.

I don't know why reverting to `var` didn't solve the problem, though.

Swiftb0y:
Great. Thanks for investigating.

JoergAtGithub:
Unfortunately the `(function(global) { … }(this));` breaks the IDE support for me. Not a single symbol from common-controller-scripts.js is recognized anymore and I get spamed with errors.

JoergAtGithub:
Now everything is Pass!
The reason was that variable objects in JavaScript exist from the start - as undefined, while const objects do not exist before completion of their initialization.
The function script.deepMerge contains recursive calls of itself. While initializing script.deepMerge it fails, because the recursive called function script.deepMerge does not exist yet.

Swiftb0y:
great. the only remaining comments are the ones from above suggesting the use of default parameters. Have you looked into that?

JoergAtGithub:
Done!

res/controllers/common-controller-scripts.js
```diff
@@ -9,728 +9,618 @@
...
+        fRateScale = (fRateScale - 1.) / engine.getValue(group, "rateRange");
+
+        engine.setValue(
+            group, "rate",
+            fRateScale * engine.getValue(group, "rate_dir")
+        );
+    }
+}
+// Add instance bpm of bpmClass to the Global JavaScript object
+this.bpm = new bpmClass();
```
christophehenry, line 626: Why didn't you keep the IIFE in my fix?

Swiftb0y: iiuc because it broke editor support for Joerg.

res/controllers/common-controller-scripts.js
```diff
@@ -9,728 +9,618 @@
...
+/**
+ * Converts an object with "red", "green", and "blue" properties (value range 0-255) into an RGB color code (e.g. 0xFF0000).
+ * @param {Object} color An object with "red", "green", and "blue" properties, each a number from 0 to 255.
+ * @returns {number} Single integer representing the RGB color code corresponding to the input color components.
+ */
 // eslint-disable-next-line no-unused-vars
-var colorCodeFromObject = function(color) {
+const colorCodeFromObject = function(color) {
     return ((color.red & 0xFF) << 16 | (color.green & 0xFF) << 8 | (color.blue & 0xFF));
 };
```
Swiftb0y, line 133: did you double check the `const` here didn't break anything?

res/controllers/common-controller-scripts.js
christophehenry: What about using [JSDoc's `@global`](https://jsdoc.app/tags-global)?

JoergAtGithub: I just tried it again, with the proposed code the symbol bpm becomes invisible outside of the function statement. Neither code completion, pop-up help nor type checking work anymore. While with the code of this PR everything works.
`this.bpm = new bpmClass();` is not a new code pattern, it's already used in the hid-parser of 2.4 and seems to work for everybody.

christophehenry: Weird. But ok then.

res/controllers/common-controller-scripts.js
```diff
@@ -9,728 +9,618 @@
...
-                * (this.midInput - this.minInput);
-        } else {
-            currentRelative = this.midInput
-                + ((currentValue - this.midOutput) / (this.maxOutput - this.midOutput))
-                * (this.maxInput - this.midInput);
+/**
+ * Used for tapping the desired BPM for a deck
+ */
+// @ts-ignore Same identifier for class and instance needed for backward compatibility
+class bpmClass {
```
Swiftb0y, line 562:
```suggestion
class bpmClass {
```

I also just noticed that we now have `[ChannelN],tempo_tap` which essentially deprecates `script.bpm`, right? cc @ronso0

daschuer:
What is the state here, is this still a 2.5.0 candidate?

daschuer:
@JoergAtGithub This one has developed conflicts

Issue #11621, opened 2023-06-04
Bug in Autodj Transition


ikr83:
### Bug Description

Hello,

Since the update to version 2.3.5, there has been a regression in the use of the autodj function. Instead of aligning the transition of the "intro start marker" of track 2 with the "outro START marker" of track 1, Autodj sometimes aligns the transition of the "intro start marker" of track 2 with the "outro END marker" of track 1. The bug seems to occur more frequently on one of the two main decks than the other. It appears that the bug is related to certain "intro start markers." On the affected tracks, deleting the marker and repositioning it slightly seems to resolve the issue. However, without knowing which tracks are affected, it's difficult to fix everything.

In the attached video, there are two transitions (from the same track), and the bug occurs on the first transition but not on the second one.

https://github.com/mixxxdj/mixxx/assets/135498842/e806eeeb-5f88-452d-9a44-5fd1446b0823

Thanx!

:)

### Version

2.3.5

### OS

Window 11

daschuer:
The "Full Intro + Outro mode" aligns either Outro End with the Intro End or Outro Start with Intro Start in a way that both is fully played: 

```   
        // In the diagrams below,
        // - is part of a track outside the outro/intro,
        // o is part of the outro
        // i is part of the intro
        // | marks the boundaries of the transition
        //
        // When outro > intro:
        // ------ooo|ooo|
        //          |iii|------
        //
        // When outro < intro:
        // ------|ooo|
        //       |iii|iii-----
```

If no intro is set, it should use the outro time. It looks like Mixxx assumes a zero lenght intro in the first case which causes the crossfader jump. I will investigate how this can happen.

daschuer:
I can confirm the issue. It happens when you seek the next track before the intro start. This can be a micro seek but this is obviously a bug.

ikr83:
Thank you! I have installed Mixxx using the downloadable installer from the website. How do I proceed to correct Mixxx with the autodjprocessor.cpp file that you modified? Should I wait for the next update? Or should I reinstall Mixxx from the source code with the updated autodjprocessor.cpp file located at src/library/autodj/autodjprocessor.cpp?

daschuer:
I have just added some improvement that limit the intro length when it is in the middle of the track an you have re-cued the track at the beginning. Please check if it feels natural. 
You can install the GitHub artefact that will appear in a few minutes at the bottom of: 
https://github.com/mixxxdj/mixxx/actions/runs/5175300378?pr=11629

[...]

PR #11629, opened 2023-06-04
Fix sharp cut transition after cueing a track without a defined intro

daschuer:
This fixed https://github.com/mixxxdj/mixxx/issues/11621

It also adds a plausibility check for the intro length. Reject intros that are extended by more than the original length or by the transition time.

### DIFF
--- a/src/library/autodj/autodjprocessor.cpp
+++ b/src/library/autodj/autodjprocessor.cpp
@@ -1281,13 +1281,13 @@ void AutoDJProcessor::calculateTransition(DeckAttributes* pFromDeck,
 
     // introEnd is equal introStart in case it has not yet been set
     if (toDeckStartSeconds < introEnd && introStart < introEnd) {
-        // Limit the intro length adjustments to the original length x 2
-        // or original length + transition time, whichever is longer.
-        double seekBack = introStart - toDeckStartSeconds;
-        if (seekBack <= 0 ||
-                seekBack <= m_transitionTime ||
-                seekBack <= introEnd - introStart) {
-            introLength = introEnd - toDeckStartSeconds;
+        // Limit the intro length that results from a revers seek
+        // to a reasonable values. If the seek was too big, ignore it.
+        introLength = introEnd - toDeckStartSeconds;
+        if (introLength > (introEnd - introStart) * 2 &&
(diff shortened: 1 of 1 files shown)


daschuer:
I need when I find time.

daschuer:
I could not reproduce the issue using the PC only, also toggling quantize does not trigger it, but a micro movement on the jog wheel can reproduce the issue.

ikr83:
I just tried on Windows and it's perfect !! Thanx!!

src/library/autodj/autodjprocessor.cpp
```diff
@@ -1281,13 +1281,13 @@ void AutoDJProcessor::calculateTransition(DeckAttributes* pFromDeck,
...
-                seekBack <= m_transitionTime ||
-                seekBack <= introEnd - introStart) {
-            introLength = introEnd - toDeckStartSeconds;
+        // Limit the intro length that results from a revers seek
+        // to a reasonable values. If the seek was too big, ignore it.
+        introLength = introEnd - toDeckStartSeconds;
+        if (introLength > (introEnd - introStart) * 2 &&
+                introLength > (introEnd - introStart) + m_transitionTime &&
+                introLength > outroLength) {
+            introLength = 0;
```
Swiftb0y, line 1290: why this limitation? IIUC if the user seeks far before the intro, then this will just skip the fade, does this make sense? Why would this be desirable?

daschuer: Imagine a live recording with some introduction words. 
The user probably has marked an intro when the music starts. Without this change and with a previous track without an outro the loooong intro time is used for fading if it is cued at 0:00. This is likely not desired. Now the transition time from the spin box is used in this case.

The other use case is the part of the bug fix. If the track is cued at intro start and the user touches the jog for a small seek. The intro is still used.

Swiftb0y: Makes sense I guess. I just worry about the complexity this introduces. AutoDJ is already quite unpredictable IMO and this doesn't make it any better.

Swiftb0y: It's also a drop in the bucket, so who cares...

Swiftb0y (approved):

Issue #9369, opened 2022-08-23
improve alignment of strings in library fields


mixxxbot:
Reported by: **[ronso0](https://github.com/ronso0)**
Date: 2018-07-02T22:40:04Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1779775](https://bugs.launchpad.net/bugs/1779775)
Tags: library, usability

---

for example
Duration: omit leading zeros
BPM: align decimal points

This would ease finding long tracks and tracks with high/low BPM tracks in large track tables.
Also this could clean up the deck view.

veeroohre:
I'd prefer to have an alignment of the "filepath" field to the left because I'm more interested in the beginning of the path rather than the filename (e. g. for searching the file in the file system).

daschuer:
In the right side is elided this allows to see the file name. The tooltip reveals the entire path. The best way to open the track in the file browser is using the right click menu. 
Does this work for you? If not, please explain why.

ronso0:
I assume this is about spotting tracks from certain directories, not about opening containing directories.

ronso0:
Fix for the BPM and duration column #11634

veeroohre:
You're right. Opening the correct directory is easy manageable via the context menu. My use case was more about cleaning my music library in the file system. So checking if a file is not in the correct directory would be much easier if I could see the beginning of the path in the track list. Maybe the alignment coud be made changeable for the user (left/right)?

daschuer:
@veeroohre Please file a new bug for your issue. 
For me it is still unclear which folder level you like to see and how the user may want to adjust the view. 
Please describe your idea. Do we need a new "Folder" column or such?

ronso0:
Reopening, because after #11634 adjacent left-aligned columns are to close.

Maybe this can be fixed by adding some horizontal padding to all cells.
For my personal branch I tried AlignHCenter for the date columns (date added [to library | to playlist], last played).

What do you think?

daschuer:
A unconditionally padding is not desired in case of small screens where users wants to make the best of their vertical space. 

A easy fix is to reorder the columns. Can we make use of a different styling? 

But I am afraid the DecimalDelegate https://github.com/mixxxdj/mixxx/pull/11634#issuecomment-1582608066 is back on the cutting table. ??

ronso0:
> A unconditionally padding is not desired in case of small screens where users wants to make the best of their vertical space.

true.

[...]

PR #11634, opened 2023-06-06
library: right align BPM, duration & bitrate values

ronso0:
Right align BPM, duration & bitrate so big/small values can easily be spotted by length (number of digits).

Also adds 'BPM precision' to the Library preferences to set the number of decimals for the BPM column.

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -426,7 +426,8 @@ QVariant BaseTrackTableModel::data(
             role != Qt::EditRole &&
             role != Qt::CheckStateRole &&
             role != Qt::ToolTipRole &&
-            role != kDataExportRole) {
+            role != kDataExportRole &&
+            role != Qt::TextAlignmentRole) {
         return QVariant();
     }
 
@@ -571,6 +572,21 @@ QVariant BaseTrackTableModel::roleValue(
             break;
         }
         M_FALLTHROUGH_INTENDED;
+    // Right align BPM, duraation and bitrate so big/small values can easily be
+    // spotted by length (number of digits)
+    case Qt::TextAlignmentRole: {
+        switch (field) {
+        case ColumnCache::COLUMN_LIBRARYTABLE_BPM:
+        case ColumnCache::COLUMN_LIBRARYTABLE_DURATION:
+        case ColumnCache::COLUMN_LIBRARYTABLE_BITRATE: {
+            // We need to cast to int due to a bug similar to
+            // https://bugreports.qt.io/browse/QTBUG-67582
+            return static_cast<int>(Qt::AlignVCenter | Qt::AlignRight);
+        }
+        default:
+            return QVariant(); // default AlignLeft for all other columns
+        }
+    }
     case Qt::DisplayRole:
         switch (field) {
         case ColumnCache::COLUMN_LIBRARYTABLE_DURATION: {
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -92,6 +92,24 @@ QSqlDatabase cloneDatabase(
 
 } // anonymous namespace
 
+// static
+constexpr int BaseTrackTableModel::kBpmColumnPrecisionDefault;
+constexpr int BaseTrackTableModel::kBpmColumnPrecisionMinimum;
+constexpr int BaseTrackTableModel::kBpmColumnPrecisionMaximum;
+
+int BaseTrackTableModel::s_bpmColumnPrecision =
+        kBpmColumnPrecisionDefault;
+
+// static
+void BaseTrackTableModel::setBpmColumnPrecision(int precision) {
+    VERIFY_OR_DEBUG_ASSERT(precision >= BaseTrackTableModel::kBpmColumnPrecisionMinimum) {
+        precision = BaseTrackTableModel::kBpmColumnPrecisionMinimum;
+    }
+    VERIFY_OR_DEBUG_ASSERT(precision <= BaseTrackTableModel::kBpmColumnPrecisionMaximum) {
+        precision = BaseTrackTableModel::kBpmColumnPrecisionMaximum;
+    }
+    s_bpmColumnPrecision = precision;
+}
 //static
 QStringList BaseTrackTableModel::defaultTableColumns() {
     return kDefaultTableColumns;
@@ -704,7 +722,8 @@ QVariant BaseTrackTableModel::roleValue(
                 if (role == Qt::ToolTipRole || role == kDataExportRole) {
                     return QString::number(bpm.value(), 'f', 4);
                 } else {
-                    return QString::number(bpm.value(), 'f', 1);
+                    // custom precision, set in DlgPrefLibrary
+                    return QString::number(bpm.value(), 'f', s_bpmColumnPrecision);
                 }
             } else {
                 return QChar('-');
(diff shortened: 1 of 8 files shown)


ronso0:
[@uklotzde's point](https://github.com/mixxxdj/mixxx/issues/9369#issuecomment-1225425026) is still valid, though I think this should be solved otherwise, not by shrinking the column
> I'm still unsure though, because for the BPM column I prefer to cut off the less significant fractional digits to the right.

daschuer:
This is clear improvident when enough space is available. 

Unfortunately it feels like a regression when the eliding with "..." kicks in on the right, because this still assumes left alignment. 

I have checked how it is done with Excel. It does not use ... eliding. It cuts off the decimal places without any hint and replaces whole values with ### if they does no longer fit. This feels like a good solution for us as well. 

The issue is that we have no info about the available space when composing the string for display. So a similar solution is probably quite hard to implement. 

Maybe we can force real decimal point alignment by making all strings equal length via padding it with a spaces of number and punctuation width 

1234567890
     11111 <- with 5 leading EN QUAD

https://www.fileformat.info/info/unicode/char/2000/index.htm

1:12:44
   1:34 <- EN QUAD + PUNCTUATION SPACE+ EN QUAD

https://www.fileformat.info/info/unicode/char/2008/index.htm

What do you think?

ronso0:
I tried exactly that but didn't find a suitable whitespace char, yet. I'll test with en quad.

ronso0:
> 1234567890
>      11111 <- with 5 leading EN QUAD

these are not aligned, so either num chars are not equal in width for var. width fonts, or en quad is not the right choice.
Maybe these are reverted to regular whitespace here in the web GUI. I'll test it.

Thing is, since users can select the library font (incl. monospace) it has to work for any font.

ronso0:
IMO we should avoid on-the-fly font meteics calculation in every BPMDelegate, I fear this has a performance impact when scrilling the library.
Maybe some pre-calculation can happen in setFont()

daschuer:
> IMO we should avoid on-the-fly font meteics calculation in every BPMDelegate, I fear this has a performance impact when scrilling the library.

There must be a metric already that generates the ... conditionally. 

I am more concerned about our personal time we waste on that. 

How would elide left work in your eyes? 

Regarding the fonts, there is even no guarantee that the numbers have equal width. So IMHO a solution that follows the normal font principals should be fine.

daschuer:
FIGURE SPACE is our friend https://www.fileformat.info/info/unicode/char/2007/index.htm
> this is equivalent to the digit width of fonts with fixed-width digits

1234567890
     11111 <- with 5 leading FIGURE SPACE
     11111 <- with 5 leading EN QUAD
     11111 <- with 5 leading EN SPACE

ronso0:
> FIGURE SPACE is our friend

okay, I'll try that.



> How would elide left work in your eyes?

It wouldn't, because the purpose of right elide is to shrink the column in order to hide the decimal(s).

daschuer:
> It wouldn't, because the purpose of right elide is to shrink the column in order to hide the decimal(s).

That was an idea during the discussion, but in the 2.4 branch we have only one decimal point and when the ... are coming they eat up more then one decimal. 

I am afraid we can't have both, a dynamic precision + right alignment.  For sake of consistency, I can imagine these solutions: 
* Right align, elide left.
* Left align with padding spaces, elide right.
* Left align like In Excel (real dynamic precision) 

As far as I remember reducing the BPM precision to 1 was a compromise, so everyone may effort the space for it. The ... elide indicates when the column is unusable small and the tooltips can be used.

daschuer:
Since FIGURE SPACE is part of the font, it should be exactly the width of a digit. Unfortunately some font designer forget to adjust it and that's why it only works for well established fonts. Does it not work for the fonts normally used in Mixxx? 

> IMO the discussion about avaliable space and elide is moot.

Yes, that is my conclusion as well. The elide ... Is almost like the Excel ### because the value can no longer be interpreted reliable. 

Since the Excel like dynamic precision solution seems to be hard to implement, I am OK with the preferences option. 
But do we really need it? Can't we live with one decimal place? That works for me. I see no use case for more digits. 

Can we than use the elide on the left than?

daschuer:
This is by the way the original QItemDelegate::drawDisplay function we need to reimplement for dynamic precision: 
https://github.com/qt/qtbase/blob/f7c8ff511c30dc4310a72b3da4b4a345efe1fba0/src/widgets/itemviews/qitemdelegate.cpp#L579
A challenge :-)

daschuer:
Ich habe mal aus interesse ChatGPT gefragt. Ich bin verblüfft: 
```
#include <QItemDelegate>
#include <QPainter>
#include <QStyleOptionViewItem>
#include <QLocale>

class DecimalDelegate : public QItemDelegate
{
public:
    void drawDisplay(QPainter* painter, const QStyleOptionViewItem& option, const QRect& rect, const QString& text) const override
    {
        // Retrieve the decimal separator for the current locale
        QLocale locale;
        QString decimalSeparator = locale.decimalPoint();

        // Calculate the number of decimal places based on the available width
        QFontMetrics fontMetrics(option.font);
        QString displayText = text;

        int textWidth = fontMetrics.horizontalAdvance(displayText);
        int maxWidth = rect.width();

        if (textWidth > maxWidth) {
            int ellipsisWidth = fontMetrics.horizontalAdvance("...");
            int numChars = displayText.length();
            int visibleWidth = maxWidth - ellipsisWidth;

            for (int i = numChars - 1; i >= 0; --i) {
                if (fontMetrics.horizontalAdvance(displayText.left(i) + "...") <= visibleWidth) {
                    if (displayText[i] == decimalSeparator) {
                        // Only decimal places are truncated, remove the ellipsis
                        displayText = displayText.left(i);
                    } else {
                        // Truncate the text and append ellipsis
                        displayText = displayText.left(i) + "...";
                    }
                    break;
                }
            }
        }

        // Draw the display text with the calculated decimal places
        QStyleOptionViewItem newOption = option;
        newOption.displayAlignment = Qt::AlignRight | Qt::AlignVCenter;
        newOption.text = displayText;

        QItemDelegate::drawDisplay(painter, newOption, rect, displayText);
    }
};

```

To use this delegate, you can set it on your QAbstractItemView using the setItemDelegate function:

```
QAbstractItemView* view = ...; // Your QAbstractItemView instance
DecimalDelegate* delegate = new DecimalDelegate;
view->setItemDelegate(delegate);
```

ronso0:
The current solution works fine for me.
* adjust the decimals in the preferences
* on next paint, the table is updated
* double-click the right-hand BPM column splitter to auto-adjust
* done.

daschuer:
Ok, this is a good step forward. 

Just the preferences are not aligned with a stretched window: 

grafik

ronso0:
Yes, I need to fix that. For the other spinbox labels as well, and the spinboxes won't have more than 4 digits, so maybe the size policy / alignment can be improved.

ronso0:
Alright, this is finished.

Please double-check the v-alignment of the right aligned columns is 'center.
The pref page should be fine now, incl. the scroll fix I already applied to comboboxes in DlgPrefSound.

daschuer (approved):
LGTM, Thank you.

m0dB:
After pulling 2.4 I get
```
AutoUic: /Users/m0dB/mixxx/src/preferences/dialog/dlgpreflibrarydlg.ui: Warning: Tab-stop assignment: 'bpmPrecisionSpinBox' is not a valid widget.
```
Could that be because of this PR?

ronso0:
jup, seems I forgot to rename that.
I'll push a fix to 2.4 this night

Issue #11596, opened 2023-05-28
Cannot open include file: 'pthread.h'


daprose-dotcom:
### Bug Description

While compiling, I get this error message. How do I solve this problem?

fatal error C1083: Cannot open include file: 'pthread.h': No such file or directory

### Version

2.3.5

### OS

Windows 10, Visual Studio Community 2019

ronso0:
Not a Mixxx issue.

daschuer:
@ikr83 has reported the same issue here: https://github.com/mixxxdj/mixxx/issues/11621#issuecomment-1576487778 Even if we know that it works if all is "correctly" configured it seems to be hard to find out what is correct. 
Maybe it is possible to fail with a better message or make is work instantly.

daschuer:
Related: 
https://github.com/mixxxdj/mixxx/issues/11597

ikr83:
I followed instructions here ...

https://github.com/mixxxdj/mixxx/wiki/Compiling%20on%20Windows
https://github.com/mixxxdj/mixxx/issues/11597

... and it works even if I didn't understand everything.

daschuer:
Fine, now it is working. Can you reproduce your original issue? My idea is to make it initially work for new contributors or at least give less obscure error messages.

ikr83:
The biggest difference is that I used CMake directly from VS2019 instead of using the command line method.

To be more specific:

At the beginning,

Download Visual Studio Community
https://aka.ms/vs/16/release/vs_community.exe

![1-Visual Studio Install Detail](https://github.com/mixxxdj/mixxx/assets/135498842/5d3bd376-6d23-4105-9cc3-64c4bb7ebd8c)
I didn't start VS2019 after installation.

Download Rust + Install
https://static.rust-lang.org/rustup/dist/x86_64-pc-windows-msvc/rustup-init.exe

Installation sccache
CMD + cargo install --git https://github.com/Be-ing/sccache.git --branch fix_msvc_fp

Download Mixxx source code
GitHub mixxxdj/mixxx

I tried with this code first:
https://github.com/mixxxdj/mixxx/releases/tag/2.3.5

Then I tried with this code:
https://github.com/mixxxdj/mixxx/archive/refs/heads/main.zip

Same results at the end...

Download + UNZIP code

Download buildenv
x64 Native Tools Command Prompt for VS 2019

    mixxx/tools/windows_buildenv.bat

Visual Studio

Open a local folder "mixxx folder"

[...]

PR #11636, opened 2023-06-08
Use ISO C fixed length types uint32_t not u_int32_t

daschuer:
The underscore types are defined  in sys/types.h in my case with an obscure comment:

```
/* These were defined by ISO C without the first `_'.  */
typedef __uint8_t u_int8_t;
typedef __uint16_t u_int16_t;
typedef __uint32_t u_int32_t;
typedef __uint64_t u_int64_t;
```  

In Mixxx there was anyway a conversion from the libraries `uint32_t`  so lets stick with that. 

This fixes a build issue with MSVC reported here: https://github.com/mixxxdj/mixxx/issues/11596#issuecomment-1581244468
by @ikr83

### DIFF
--- a/src/controllers/bulk/bulkcontroller.cpp
+++ b/src/controllers/bulk/bulkcontroller.cpp
@@ -54,7 +54,7 @@ void BulkReader::run() {
     qDebug() << "Stopped Reader";
 }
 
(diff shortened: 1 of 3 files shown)


JoergAtGithub:
Works for me if I enable BULK locally. Thank you!

PR #11638, opened 2023-06-09
script to create inverted colors skins, and a LateDay skin

michelealbano:
I developed a python script that visits the folders containing a skin, and re-creates in the current folder the same folder structure, copies locally the png / psd files, and for everything else it copies the files but inverts the colors.

This PR could solve issue #9217

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/labeler@v4.2.0
+      - uses: actions/labeler@v4.3.0
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           sync-labels: false


ronso0:
Interesting! (I didn't look into the details, yet. )

In general the easiest way to fix #9217 would be another color scheme for an existing skin (for LateNight preferrably). Creating yet another skin increases the maintenance burden, especially if we just duplicate an existing skin.

Could you look into making the script work with the existing LateNight architecture and use the PaleMoon scheme?
Each scheme has a
* `Scheme` section in skin.xml https://github.com/mixxxdj/mixxx/blob/2.4/res/skins/LateNight/skin.xml#L125-L234
* style subdirectory https://github.com/mixxxdj/mixxx/tree/2.4/res/skins/LateNight/palemoon
* a stylesheet that complements the basic stylesheet style.qss https://github.com/mixxxdj/mixxx/blob/2.4/res/skins/LateNight/style_palemoon.qss

michelealbano:
As suggested, I have no new skin anymore, but a color scheme for LateNight. Solving #9217 possibly

michelealbano:
Should I checkout main, or 2.4?
Michele

daschuer:
Skin changes can go to 2.4-beta. Release is scheduled at 2023-08-08  so no stress, if you miss that date we will merge it to 2.5 Wich will be released 6 month later. 

Just last weekend it was itching me that the Shade skin is good for sunlight, but  the library is not. 

I did not had a look to the script, is it similar to the color shema implementation of Shade? In that case, the colors from the PNGs are replaced on the fly. However recalculating the colors from the library css is missing. 

Can you script help here?

ronso0:
@michelealbano When you rebase could you please separate the changes into
* add script
* add PaleSun scheme

The color inversion seems to work but the theme itself is not usable IMO
* most button icons are hard to recognise, the dark outlines that should increase local contrast now make the icons appear muddy
* text/background contrast is as low (or even lower?) as in PaleMoon, but for daylight we need high contrast
* knobs are hard to read

image
image

Bottom line: simply inverting the colors doesn't produce a daylight skin. I think the issue is that the script handles all colors the same way (or misses some SVG colors?) and can't really replace manual labor.
I chose distinct colors in most places (e.g. `#000001` instead of `#000`) so they can be bulk-changed with simple find & replace scripts, for example the icon shadows, the button borders etc.

ronso0:
Yeah, this may be due to the scour SVG optimization.

michelealbano:
Hi, I found a way to improve my script + the skin. Please check it out.
Best,
Michele

JoergAtGithub:
Link to new PR: https://github.com/mixxxdj/mixxx/pull/11747

PR #11640, opened 2023-06-09
use ClampToEdge instead of ClampToBorder

m0dB:
@daschuer you reported artifacts at the vu meter edges, and I fixed it but apparently the fix didn't make it into 2.4

From the Qt docs:

QOpenGLTexture::ClampToEdge | 0x812F | Clamps the texture coordinates to [0,1]
QOpenGLTexture::ClampToBorder | 0x812D | As for ClampToEdge but also blends samples at 0 and 1 with a fixed border color

so ClampToEdge is what we want


m0dB:
Ah nevermind, I see you have it prepared in https://github.com/mixxxdj/mixxx/pull/11601/commits/33be86ba1fb1b37dd38b80ff8d7a13ad720a10e3
Sorry for the noise!

PR #11643, opened 2023-06-10
Remove usage of ControlObject::getControl

ferranlala:
This is a small PR removing one usage of `ControlObject::getControl`, which we want to eventually remove AFAIU.


daschuer (approved):
Thank you. LGTM

PR #11644, opened 2023-06-11
fix #11603 table draw delaying the waveformwidgetfactory render and swap

m0dB:
Drawing the table can be slow (profiling points to the text rendering, so there is no easy way to improve this). As a result the paintEvent holds up the event loop and the WaveformWidgetFactory may not receive its render and swap signal meta events in time. See https://github.com/mixxxdj/mixxx/issues/11603

This PR fixes this by forcing the processing of such time sensitive events during the drawing of the table cells (in the paint call of each cell) to fix this.

### DIFF
--- a/src/mixxxapplication.h
+++ b/src/mixxxapplication.h
@@ -12,6 +12,8 @@ class MixxxApplication : public QApplication {
 
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     bool notify(QObject*, QEvent*) override;
+    bool hasTimeSensitiveEvents() const;
+    void processTimeSensitiveEvents();
 #endif
 
   private:
--- a/src/mixxxapplication.h
+++ b/src/mixxxapplication.h
@@ -12,9 +12,9 @@ class MixxxApplication : public QApplication {
 
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     bool notify(QObject*, QEvent*) override;
+#endif
     bool hasTimeSensitiveEvents() const;
     void processTimeSensitiveEvents();
-#endif
 
   private:
     bool touchIsRightButton();
--- a/src/waveform/vsyncthread.cpp
+++ b/src/waveform/vsyncthread.cpp
@@ -53,6 +53,7 @@ void VSyncThread::run() {
             usleep(1000);
         } else { // if (m_vSyncMode == ST_TIMER) {
             emit vsyncRender(); // renders the new waveform.
+            setHasPendingRender(true);
 
             // wait until rendering was scheduled. It might be delayed due a
             // pending swap (depends one driver vSync settings)
@@ -68,6 +69,7 @@ void VSyncThread::run() {
 
             // swaps the new waveform to front in case of gl-wf
             emit vsyncSwap();
+            setHasPendingSwap(true);
 
             // wait until swap occurred. It might be delayed due to driver vSync
             // settings.
--- a/src/waveform/waveformwidgetfactory.cpp
+++ b/src/waveform/waveformwidgetfactory.cpp
@@ -708,6 +708,7 @@ void WaveformWidgetFactory::notifyZoomChange(WWaveformViewer* viewer) {
 }
 
 void WaveformWidgetFactory::render() {
+    m_vsyncThread->setHasPendingRender(false);
     ScopedTimer t("WaveformWidgetFactory::render() %1waveforms",
             static_cast<int>(m_waveformWidgetHolders.size()));
 
@@ -783,6 +784,7 @@ void WaveformWidgetFactory::render() {
 }
 
 void WaveformWidgetFactory::swap() {
+    m_vsyncThread->setHasPendingSwap(false);
     ScopedTimer t("WaveformWidgetFactory::swap() %1waveforms",
             static_cast<int>(m_waveformWidgetHolders.size()));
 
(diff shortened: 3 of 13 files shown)


m0dB:
Maybe there are other time sensitive events that could make use of the same mechanism? If I am not mistaken, a lot of mixxx's control flow goes through the event loop, but I don't know that part of the code enough.

daschuer:
This is a good idea for a workaround. 

Unfortunately it is a kind of cooperate multi threading that fights a high CPU load with even more CPU load. This basically delays scrolling in favor of a smooth waveform scrolling. Is this a good trade off, because the users attention is on the library and not on the waveform?  On the other hand, it is fair to give away the CPU, when the scrolling is likely stuttering anyway to allow other parts of Mixxx behave normally.  

It would be really interesting why Linux is not affected. Do we have an issue with text rendering? Maybe some misconfiguration of QT? The best solution would be one solving the root cause. 

If we continue this PR, we need to make sure that the extra CPU is limited to a minimum. We need to make sure scrolling still behaves normal on low power devices like Raspberries and friends. 

Unfortunately `QCoreApplication::sendPostedEvents` is not a nop() when there is no event pending. It locks the event cue and cycles through it to find the signals for given receiver object.  Doing it in every delegate seems a bit too much.

daschuer:
I can confirm this works without a regression on Ubuntu Focal.

m0dB:
> Unfortunately QCoreApplication::sendPostedEvents is not a nop() when there is no event pending. It locks the event cue and cycles through it to find the signals for given receiver object. Doing it in every delegate seems a bit too much.

Yeah, I see what you mean. This could be avoided very easily with an atomic bool, set by the vsync thread.

daschuer:
Yes I think with such a bool in place we can safely commit this.

src/mixxxapplication.h
```diff
@@ -12,6 +12,7 @@ class MixxxApplication : public QApplication {
 
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     bool notify(QObject*, QEvent*) override;
+    static void processTimeSenstiveEvents();
```
ronso0, line 15: typo
```suggestion
    static void processTimeSensitiveEvents();
```

ronso0 (changes requested):

src/waveform/vsyncthread.cpp
```diff
@@ -63,7 +63,9 @@ void VSyncThread::run() {
                 m_timer.elapsed().toIntegerMicros());
             // waiting for interval by sleep
             if (remainingForSwap > 100) {
-                usleep(remainingForSwap);
+                // Leave a bit of margin for event delivery. In the worst case we will
+                // call swap a bit too early, so it will block, but not more than 100us.
+                usleep(remainingForSwap - 100);
```
daschuer, line 68: The reason for > 100 was that usleep() was not able to sleep shorter on my test device at that time. 
Not sure how the value is now. 

The -100 here should have no effect, because if we are to early, the timer will loaded with +100 to compensate this adjustment. I have just tested with 3000 also without any effect.  Does it have an effect on your system?

daschuer (changes requested):
It fails with Qt6

src/waveform/vsyncthread.cpp
m0dB: I don't understand why you say subtracting 100 from the sleep time will have no effect. If I understand correctly, remainingForSwap is the time until the next swap is supposed to happen, so, if we call swap at that exact moment, the swap will take place immediately. If we would call swap a bit earlier, it will block until the swap takes place. My reason to sleep a bit less, and call swap a bit earlier, is that we are using a queued signal/slot to make the call. If the queued signal/slot takes to long, we will miss the swap and the call to swap will significantly longer. By sleeping a bit less, we reduce that risk. 

But the truth is, this is speculation. I will do some logging to check if this actually occurs.

m0dB:
Unfortunately, after rebasing this on 2.4 the table isn't drawn correctly anymore and I get a bunch of warnings:
```
warning [Main] QPainter::begin: A paint device can only be painted by one painter at a time.
warning [Main] QPainter::setCompositionMode: Painter not active
warning [Main] QBackingStore::endPaint() called with active painter; did you forget to destroy it or call QPainter::end() on it?
```
and even crashes. I will put this on hold until I have time to debug this.

m0dB:
I have revived this branch. It works very well on macOS, please review and test on Linux and Windows.

JoergAtGithub:
This does seems not influence the performance on Windows11 - neither positive nor negative.

JoergAtGithub:
Did you tried how #12009 influence the library performance on macOS? There I noticed a performance gain on Windows.

m0dB:
No, I haven't. I will try that tomorrow.

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -1224,3 +1226,7 @@ QSurfaceFormat WaveformWidgetFactory::getSurfaceFormat() {
 #endif
     return format;
 }
+
+bool WaveformWidgetFactory::hasPendingTimeSensitiveEvents() const {
+    return m_vsyncThread->hasPendingSwap() || m_vsyncThread->hasPendingRender();
```
JoergAtGithub, line 1231:
```suggestion
    return m_vsyncThread && (m_vsyncThread->hasPendingSwap() || m_vsyncThread->hasPendingRender());
```

Issue #11626, opened 2023-06-04
Zooming in waveform using scrollwheel is broken


aareschluchtje:
### Bug Description

When using 2.3.5 it was possible to use the scrollwheel to zoom the waveform in and out. I'm unable to do the same in 2.4.0 (the scrollwheel simply doesn't do anything when the mouse is positioned over the waveform). When reinstalling 2.3.5 it was working fine again.

### Version

2.4.0

### OS

Windows 10

daschuer:
I cannot confirm this with Ubuntu Focal. Is this a windows only thing?

MarcPlace:
Can confirm this issue with 2.4.0 (592c0c728e538fb57a1fbcf617fed2c0dc39085d) using Gentoo Linux. Scrollwheel doesn't zoom in or out when mouse pointer is positioned inside the waveform.

atskler:
The issue is also there on openSUSE Tumbleweed.

Swiftb0y:
Can also confirm on Fedora 38. bringing @m0dB in the loop. Do you have an idea? Possibly just something you forgot to implement and that none of us spotted while testing #10989?

ywwg:
agree that seems a likely candidate

Swiftb0y:
No worries, take your time. Thank you for taking care of it.

m0dB:
On macOS it works fine (tried both with two-fingers on the trackpad and with an external mouse). And looking at the code, I don't see anything that explains this. I forward all events from the openglwindow to the widget...

Maybe I can figure this out if someone with a system where it doesn't work can add some logging and send me the results.

In src/widget/openglwindow.cpp add the following:

At the top of the file:

#include <QDebug>

and as first line of bool OpenGLWindow::event(QEvent* pEv) {

    qDebug() << pEv;

and run with ./mixxx --logLevel debug

m0dB:
Ah, and in src/mixxxmainwindow.cpp:

1131 bool MixxxMainWindow::eventFilter(QObject* obj, QEvent* event) {
1132     if (event->type() == QEvent::Wheel) {
1133         qDebug() << "MixxxMainWindow::eventFilter" << obj << event;
1134     }
1135

Swiftb0y:
<details><summary>Log output with your patch</summary>
<p>


```
[...]

PR #11645, opened 2023-06-11
Fix waveform zooming regression

ywwg:
Fixes https://github.com/mixxxdj/mixxx/issues/11626


ywwg:
this is not the right place to fix this

ywwg:
ok moved things around

PR #11647, opened 2023-06-11
fix macOS slider styling in preferences dialog

### LINKED ISSUE #11604: Mixxx Preferences: Sliders have transparency
### Bug Description

Open `Mixxx Preferences > Waveforms`and look at sliders while moving preferences window around. This effects all panels using a sliders so `Vinyl Control`, `Mixer` and other panels are affected just the same.

**Currently**

The slider dots all have opacity.

**Expected**

Seems unexpected. Preferences should not apply opacity.

### Version

Mixxx 2.4-alpha-1508-gb763121761 (HEAD)

### OS

macOS 13.4

slider opacity

In this example you can see the waveform through the first slider knob.

m0dB:
This PR fixes several issues with sliders in the preferences dialog:

- the handle is semi-transparent
- the groove is not correctly centered vertically in relation to the widgets left and right of it
- the handle is cut off at the top (the last two issues seem to be a consequence of the slider being higher than the space reserved for it in the layout)
- the style of the handle seems based on older versions of macOS.

The fix simply uses a different style sheet for these sliders, that fixes all these issues (the look is based on the sliders as can be found in the macOS system settings).

The PR also includes obj-c code to check if macOS is set up to use dark or light appearance, and a different style sheet is used accordingly.

This is a very non-invasive fix so I don't see any risk of regression.

### DIFF
--- a/src/preferences/dialog/dlgpreferences.cpp
+++ b/src/preferences/dialog/dlgpreferences.cpp
@@ -48,6 +48,10 @@
 #include "util/color/color.h"
 #include "util/widgethelper.h"
 
+#ifdef __APPLE__
+#include "util/darkappearance.h"
+#endif
+
 DlgPreferences::DlgPreferences(
         std::shared_ptr<mixxx::ScreensaverManager> pScreensaverManager,
         std::shared_ptr<mixxx::skin::SkinLoader> pSkinLoader,
@@ -61,6 +65,7 @@ DlgPreferences::DlgPreferences(
           m_pConfig(pSettingsManager->settings()),
           m_pageSizeHint(QSize(0, 0)) {
     setupUi(this);
+    fixSliderStyle();
     contentsTreeWidget->setHeaderHidden(true);
 
     // Add '&' to default button labels to always have Alt shortcuts, indpependent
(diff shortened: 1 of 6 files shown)


src/preferences/dialog/dlgpreferences.cpp
```diff
@@ -61,6 +65,50 @@ DlgPreferences::DlgPreferences(
           m_pConfig(pSettingsManager->settings()),
           m_pageSizeHint(QSize(0, 0)) {
     setupUi(this);
+
```
ferranpujolcamins, line 68: I'd like to have this extracted into a helper function. When reading the `DlgPreferences` constructot I just need to know that we `fixSlidersAppareance();`. I don't need to know the details.

ferranlala (changes requested):
Thank you for taking this bug @m0dB.

I've just added one change request, otherwise looks good.

src/preferences/dialog/dlgpreferences.cpp
m0dB: Gràcies, Ferran, good point. I have moved the code to a private method as suggested.

foss-:
Looking great in 2.4-beta-30-gbd3a6f584a
Thanks so much for your work  🚀

Screenshot 2023-06-24

m0dB:
Thanks. My pleasure!

PR #11661, opened 2023-06-14
Import from Rhythmbox playlist fix

daschuer:
This fixes a regression since https://github.com/mixxxdj/mixxx/pull/11230 and Mixxx 2.3.4
that bulk importing tracks from playlists was broken.

### DIFF
--- a/src/library/rhythmbox/rhythmboxfeature.cpp
+++ b/src/library/rhythmbox/rhythmboxfeature.cpp
@@ -91,13 +91,14 @@ RhythmboxFeature::~RhythmboxFeature() {
     delete m_pRhythmboxPlaylistModel;
 }
 
-BaseSqlTableModel* RhythmboxFeature::getPlaylistModelForPlaylist(const QString& playlist) {
-    BaseExternalPlaylistModel* pModel = new BaseExternalPlaylistModel(
-                                            this, m_pLibrary->trackCollections(),
-                                            "mixxx.db.model.rhythmbox_playlist",
-                                            "rhythmbox_playlists",
-                                            "rhythmbox_playlist_tracks",
-                                            m_trackSource);
+std::unique_ptr<BaseSqlTableModel>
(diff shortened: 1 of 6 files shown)


src/library/rhythmbox/rhythmboxfeature.cpp
```diff
@@ -91,13 +91,14 @@ RhythmboxFeature::~RhythmboxFeature() {
...
 
-BaseSqlTableModel* RhythmboxFeature::getPlaylistModelForPlaylist(const QString& playlist) {
-    BaseExternalPlaylistModel* pModel = new BaseExternalPlaylistModel(
-                                            this, m_pLibrary->trackCollections(),
-                                            "mixxx.db.model.rhythmbox_playlist",
-                                            "rhythmbox_playlists",
-                                            "rhythmbox_playlist_tracks",
-                                            m_trackSource);
+std::unique_ptr<BaseSqlTableModel>
+RhythmboxFeature::createPlaylistModelForPlaylist(const QString& playlist) {
```
Swiftb0y, line 95: just to make sure I understand the fix. The issue was basically that an oversight during refactoring resulted in the wrong member function being called?

Swiftb0y (commented):
can you spend some more time making clazy happy? I think you'll need to add the appropriate `override`s to all member functions of the class.

src/library/rhythmbox/rhythmboxfeature.cpp
daschuer: Yes, the getPlaylistModelForPlaylist() is no longer called, instead the default base class implementation is used returning null.

Swiftb0y (approved):
code LGTM. I don't use rythmbox so I can't really verify this fix easily

daschuer:
I can. And the coding mistake was obvious.
Merge?

Swiftb0y:
Sure...

PR #11662, opened 2023-06-16
controller_mapping_validation_test: Support all 2.4 controller mapping API functions

JoergAtGithub:
- Not all HID API functions were allowed in controller_mapping_validation_test
- controller_mapping_validation_test wouldn't have detect, if the HID function sendOutputReport would have been used in MIDI or BULK mappings
- Now there is a hint printed, if somebody uses send for HID or MIDI were newer more specific alternatives exist


Swiftb0y (approved):
LGTM. thank you

PR #2941, opened 2020-07-14
Preferences > Sound Hardware: prevent changing combobox or spinbox values ...

ronso0:
... when scrolling the preferences page to avoid accidentally changing important sound settings and applying them unnoticed.
https://bugs.launchpad.net/mixxx/+bug/1885628

Boxes still accept scroll events when they're focused.

PR #11663, opened 2023-06-17
Preferences: use helper for input widget scroll safeguard

ronso0:
This moves the existing safeguards (I started with #2941) to helpers in DlgPreferencePage and uses it in the other pages as well. 

Btw, am I the only who's annoyed by this hover scroll issue? or is just that we have too many big pref pages that require scrolling even on large screens so the issue is relevant at all?

The helper iterates over all dialog children and install the safeguard for input widgets.
There's also a per-widget helper if a page only has a few widgets.

Unfortunately, this doesn't work for QGroupBoxes.
But before I found a way to deal with this, I already started reworking `DlgPrefVinyl` to store the input widgets in lists (and simplified the entire page code, and fixed an init bug¹, and added proper tabstops)... so this page is pretty much rewritten with widget lists. There are now quite some lines to review in that page. I didn't want to throw away that work so I finished it. If no one wants to review that I can move that commit to a separate PR (and maybe isolate the init bugfix, too).

¹on show, the lead-in time for non existing config values was set to 0 instead of the default time that is picked when selecting a VC type.

### DIFF
--- a/src/preferences/dialog/dlgprefbeats.cpp
+++ b/src/preferences/dialog/dlgprefbeats.cpp
@@ -47,6 +47,8 @@ DlgPrefBeats::DlgPrefBeats(QWidget* parent, UserSettingsPointer pConfig)
             &QCheckBox::stateChanged,
             this,
             &DlgPrefBeats::slotReanalyzeImportedChanged);
+
+    setScrollSafeGuard(comboBoxBeatPlugin);
 }
 
 DlgPrefBeats::~DlgPrefBeats() {
--- a/src/preferences/dialog/dlgprefcolors.cpp
+++ b/src/preferences/dialog/dlgprefcolors.cpp
@@ -69,6 +69,7 @@ DlgPrefColors::DlgPrefColors(
             &DlgPrefColors::slotReplaceCueColorClicked);
 
     loadSettings();
+    setScrollSafeGuardForAllInputWidgets(this);
 }
 
 DlgPrefColors::~DlgPrefColors() {
--- a/src/preferences/dialog/dlgprefcolors.cpp
+++ b/src/preferences/dialog/dlgprefcolors.cpp
@@ -68,18 +68,15 @@ DlgPrefColors::DlgPrefColors(
             this,
             &DlgPrefColors::slotReplaceCueColorClicked);
 
-    loadSettings();
     setScrollSafeGuardForAllInputWidgets(this);
+
+    slotUpdate();
 }
 
 DlgPrefColors::~DlgPrefColors() {
 }
 
 void DlgPrefColors::slotUpdate() {
-    loadSettings();
-}
-
-void DlgPrefColors::loadSettings() {
     comboBoxHotcueColors->clear();
     comboBoxTrackColors->clear();
     for (const auto& palette : qAsConst(mixxx::PredefinedColorPalettes::kPalettes)) {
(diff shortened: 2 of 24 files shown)


src/preferences/dialog/dlgprefcolors.cpp
```diff
@@ -68,17 +68,15 @@ DlgPrefColors::DlgPrefColors(
             this,
             &DlgPrefColors::slotReplaceCueColorClicked);
 
-    loadSettings();
+    setScrollSafeGuardForAllInputWidgets(this);
```
daschuer, line 71: since it is a member function, it already has the "this" pointer.

src/preferences/dialog/dlgpreferencepage.cpp
```diff
@@ -13,3 +21,63 @@ DlgPreferencePage::~DlgPreferencePage() {
...
+    for (auto* ch : obj->children()) {
+        // children() does not descend into QGroupBox,
+        // so we need to do it manually
+        QGroupBox* gBox = qobject_cast<QGroupBox*>(ch);
+        if (gBox) {
+            setScrollSafeGuardForAllInputWidgets(gBox);
+            continue;
+        }
+
+        QComboBox* combo = qobject_cast<QComboBox*>(ch);
```
daschuer, line 37: We use a p prefix for pointers.

daschuer (changes requested):
Thank you. I have added some comments

src/preferences/dialog/dlgprefbeats.cpp
```diff
@@ -47,6 +47,8 @@ DlgPrefBeats::DlgPrefBeats(QWidget* parent, UserSettingsPointer pConfig)
             &QCheckBox::stateChanged,
             this,
             &DlgPrefBeats::slotReanalyzeImportedChanged);
+
+    setScrollSafeGuard(comboBoxBeatPlugin);
```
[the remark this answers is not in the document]
ronso0: I assume setting the scroll guard per widget is faster than using `setScrollSafeGuardForAllInputWidgets` which iterates over **all** dialog children, do an object cast etc.
To keep preferences construction (startup) fast I chose the per-widget function if there are only a few affected input widgets.
I didn't measure the impact though, maybe the difference is negligible.

For this page it probably doesn't make a big difference since there are only a few widgets.

ronso0:
I'll add the p prefix and remove obsolete `this` arguments, and pack that as fixup commit.

Let me know when you intend to start reviewing DlgPrefVinyl, so I can squash that beforehand.

src/preferences/dialog/dlgpreferencepage.cpp
ronso0: :+1:

src/preferences/dialog/dlgprefcolors.cpp
ronso0: :+1:

src/preferences/dialog/dlgprefbeats.cpp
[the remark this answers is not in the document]
daschuer: I don't have a strong opinion here. It can stay like this if you like.

ronso0:
> We use a p prefix for pointers.

fixed and rebased.

daschuer (approved):
LGTM and works good, Thank you.

PR #11326, opened 2023-03-04
FIFO mode option for HID OutputReports

JoergAtGithub:
While #4585 makes the Mixxx HID subsystem fast and responsive in overload situations for any device, that use reports in the sense of the USB HID Class specification, these optimizations do not work for the Traktor Kontrol S4 Mk3, which uses HID reports to transport the data of it's own protocol.
First attemp to address this S4 specific protocol was https://github.com/mixxxdj/mixxx/pull/4692 (Option for resending reports with identical data), but this was not enough. It came out #10828, that the S4 protocol also depend on sequences of reports with the same ReportID.

This PR replaces the Resend-Option from #4692 with an Option to use a FIFO for reports, which must not be skipped.

Closes: #10828

PR #11664, opened 2023-06-18
hid-controller-api.d.ts: Adapted the description from hidcontroller.h for parameter useNonSkippingFIFO

JoergAtGithub:
The formatting should be readable with editors that the support MarkDown in JSDoc and editors that print the text unformatted
Behavior of the parameter was changed in #11326

### DIFF
--- a/res/controllers/hid-controller-api.d.ts
+++ b/res/controllers/hid-controller-api.d.ts
@@ -18,18 +18,54 @@ declare namespace controller {
      * @param dataList Data to send as list of bytes
      * @param length Unused but mandatory argument for backwards compatibility
      * @param reportID 1...255 for HID devices that uses ReportIDs - or 0 for devices, which don't use ReportIDs
-     * @param resendUnchangedReport If set, the report will also be send, if the data are unchanged since last sending [default = false]
+     * @param useNonSkippingFIFO If set, the report will send in FIFO mode
+     *
+     *   FALSE (default):
+     *     - Reports with identical data will be sent only once.
+     *     - If reports were superseded by newer data before they could be sent,
+     *       the oudated data will be skipped.
+     *     - This mode works for all USB HID class compatible reports,
+     *       in these each field represents the state of a control (e.g. an LED).
+     *     - This mode works best in overload situations, where more reports
+     *       are to be sent, than can be processed.
+     *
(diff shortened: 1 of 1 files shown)


res/controllers/hid-controller-api.d.ts
```diff
@@ -18,18 +18,54 @@ declare namespace controller {
      * @param dataList Data to send as list of bytes
      * @param length Unused but mandatory argument for backwards compatibility
      * @param reportID 1...255 for HID devices that uses ReportIDs - or 0 for devices, which don't use ReportIDs
-     * @param resendUnchangedReport If set, the report will also be send, if the data are unchanged since last sending [default = false]
+     * @param useNonSkippingFIFO If set, the report will send in FIFO mode
+     *
+     *   FALSE (default):
```
Swiftb0y, line 23: lets use the javascript literal name (I guess optionally enclosing it in ``?). This is not COBOL ;)
```suggestion
     *   `false` (default):
```

res/controllers/hid-controller-api.d.ts
```diff
@@ -18,18 +18,54 @@ declare namespace controller {
...
+     *       are to be sent, than can be processed.
+     *
+     *  TRUE:
+     *    - The report will not be skipped under any circumstances,
+     *      except FIFO memory overflow.
+     *    - All reports with useNonSkippingFIFO set True will be send before
+     *      any cached report with useNonSkippingFIFO set False.
+     *    - All reports with useNonSkippingFIFO set True will be send in
+     *      strict First In / First Out (FIFO) order.
+     *    - Limit the use of this mode to the places, where it is really necessary.
```
Swiftb0y, line 39: and neither is this python
```suggestion
     *    - All reports with useNonSkippingFIFO set `true` will be send before
     *      any cached report with useNonSkippingFIFO set `false`.
     *    - All reports with useNonSkippingFIFO set `true` will be send in
     *      strict First In / First Out (FIFO) order.
     *    - Limit the use of this mode to the places, where it is really necessary.
```

Swiftb0y (changes requested):
lgtm otherwise

Swiftb0y (approved):
LGTM thank you

PR #11666, opened 2023-06-18
ITunesXMLImporterTest: Fix sporadic crash due to a not initalized reference.

daschuer:
Use a function call with a null check instead.

The failing test has happened on Launchpad: 
https://launchpadlibrarian.net/672913729/buildlog_ubuntu-focal-amd64.mixxx_2.4~beta~22~ge7bb02f7a5-1~focal_BUILDING.txt.gz

### DIFF
--- a/src/library/itunes/itunesfeature.h
+++ b/src/library/itunes/itunesfeature.h
@@ -31,6 +31,11 @@ class ITunesFeature : public BaseExternalLibraryFeature {
 
     TreeItemModel* sidebarModel() const override;
 
+    // This is called form the ITunesXMLImporter thread
+    bool isImportCanceled() {
+        return m_cancelImport.load();
+    }
+
   public slots:
     void activate() override;
     void activate(bool forceReload);
--- a/src/library/itunes/itunesfeature.h
+++ b/src/library/itunes/itunesfeature.h
@@ -31,8 +31,8 @@ class ITunesFeature : public BaseExternalLibraryFeature {
 
     TreeItemModel* sidebarModel() const override;
 
-    // This is called from the ITunesXMLImporter thread
-    bool isImportCanceled() {
+    // This is called from the ITunesXMLImporter thread and generally threadsafe
+    bool isImportCanceled() const {
         return m_cancelImport.load();
     }
 
--- a/src/library/itunes/itunesfeature.h
+++ b/src/library/itunes/itunesfeature.h
@@ -31,7 +31,7 @@ class ITunesFeature : public BaseExternalLibraryFeature {
 
     TreeItemModel* sidebarModel() const override;
 
-    // This is called form the ITunesXMLImporter thread
+    // This is called from the ITunesXMLImporter thread
     bool isImportCanceled() {
         return m_cancelImport.load();
     }
--- a/src/library/itunes/itunesimporter.cpp
+++ b/src/library/itunes/itunesimporter.cpp
@@ -1,0 +1,14 @@
+#include "library/itunes/itunesfeature.h"
+#include "library/itunes/itunesxmlimporter.h"
+
+ITunesImporter::ITunesImporter(ITunesFeature* pParentFeature)
+        : m_pParentFeature(pParentFeature) {
+}
+
+bool ITunesImporter::canceled() {
+    // The parent feature may be null during testing
+    if (m_pParentFeature) {
+        return m_pParentFeature->isImportCanceled();
+    }
+    return false;
+}
--- a/src/library/itunes/itunesimporter.cpp
+++ b/src/library/itunes/itunesimporter.cpp
@@ -5,7 +5,7 @@ ITunesImporter::ITunesImporter(ITunesFeature* pParentFeature)
         : m_pParentFeature(pParentFeature) {
 }
 
-bool ITunesImporter::canceled() {
+bool ITunesImporter::canceled() const {
     // The parent feature may be null during testing
     if (m_pParentFeature) {
         return m_pParentFeature->isImportCanceled();
--- a/src/library/itunes/itunesimporter.h
+++ b/src/library/itunes/itunesimporter.h
@@ -2,7 +2,8 @@
 
 #include <memory>
 
-#include "library/treeitem.h"
+class TreeItem;
+class ITunesFeature;
 
 struct ITunesImport {
     std::unique_ptr<TreeItem> playlistRoot;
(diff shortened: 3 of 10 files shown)


daschuer:
Finally re based and all green. Sorry for the noise. 
We should merge this soon, because our beta ppa keeps failing.

ronso0:
> because our beta ppa keeps failing

Just the Focal amd64 build fails, so to have a package ASAP for this, too, how about disabling the test just for this build? (if that's possible) To not merge this in a rush..

Swiftb0y:
humbly requesting @fwcd opinion on this refactoring.

src/library/itunes/itunesimporter.h
```diff
@@ -2,22 +2,23 @@
...
+class TreeItem;
+class ITunesFeature;
 
 struct ITunesImport {
     std::unique_ptr<TreeItem> playlistRoot;
 };
 
 class ITunesImporter {
   public:
+    ITunesImporter(ITunesFeature* pParentFeature);
```
fwcd, line 14: Hm, not sure how I feel about the cyclic dependency here (and in the DAO). Perhaps we could at least turn this into an interface for "cancellables" to avoid the forward declarations?

src/library/itunes/itunesfeature.h
```diff
@@ -31,6 +31,11 @@ class ITunesFeature : public BaseExternalLibraryFeature {
 
     TreeItemModel* sidebarModel() const override;
 
+    // This is called form the ITunesXMLImporter thread
```
fwcd, line 34:
```suggestion
    // This is called from the ITunesXMLImporter thread
```

src/library/itunes/itunesfeature.h
```diff
@@ -31,6 +31,11 @@ class ITunesFeature : public BaseExternalLibraryFeature {
...
+    bool isImportCanceled() {
+        return m_cancelImport.load();
```
fwcd, line 36: Even though this is a trivial implementation, should we move this to the .cpp file?

fwcd (commented):
Some quick notes from glancing over the changes, I'll see if I can do a more thorough review over the next few days

fwcd:
I haven't dug into the test log, so I am not sure if I understand the problem this fixes yet, but if there is some unsafe reference in there, could we perhaps fix the bug in isolation before refactoring it?

src/library/itunes/itunesimporter.h
daschuer: Both inherited classes had an ITunesFeature* member anyway. Moving it into the base class is just a change for code deduplication. 

Defining an interface could be an option. I am just not the right person to do any more refactoring, because I have no access to a macOS build device.

fwcd: Does it need to be in the base class though? Since `ITunesImporter` is interface-ish, we generally don't instantiate it directly and I don't see much benefit in moving the member up there (we wouldn't really save on boilerplate and the fact that the interface is unaware of the parent feature was a more or less intentional choice to precisely avoid this cyclic dependency involving forward declarations etc)

daschuer:
> if there is some unsafe reference in there, could we perhaps fix the bug in isolation before refactoring it?

OK, due to disabling the test the pressure is gone and we may go for a refactoring. 

The original root cause for the crash was that the `const std::atomic<bool>& m_cancelImport;` had random values. 
During the fix I have read the suggestion for refactoring and tried to follow it make m_cancelImport a value type in ITunesImport. This however did not work good, because it is used in ImporterImpl keeping the cyclic dependency. 
In addition the cyclic dependency is used anyway in `TreeItem::newRoot(m_pParentFeature);`

IMHO this solution is now on a pragmatic middle ground with a function call utilizing the already existing pointer after a null check.

fwcd:
> the `const std::atomic<bool>& m_cancelImport;` had random values.

How does this happen? Do members in C++ not initialize (deterministically) before methods can be called or does the test framework do some magic there that breaks this?

https://github.com/mixxxdj/mixxx/blob/a5ae80ac68005dc34f9dcc8030d80035b8053238/src/test/itunesxmlimportertest.cpp#L30

My understanding was that this would (safely) initialize `cancelImport` to `false`.

daschuer:
I have also no explanation why it is not initialized during tests.

src/library/itunes/itunesfeature.h
```diff
@@ -31,6 +31,11 @@ class ITunesFeature : public BaseExternalLibraryFeature {
 
     TreeItemModel* sidebarModel() const override;
 
+    // This is called from the ITunesXMLImporter thread
+    bool isImportCanceled() {
```
Swiftb0y, line 35:
```suggestion
    // This is called from the ITunesXMLImporter thread and generally threadsafe
    bool isImportCanceled() const {
```

src/library/itunes/itunesimporter.cpp
```diff
@@ -0,0 +1,14 @@
+#include "library/itunes/itunesfeature.h"
+#include "library/itunes/itunesxmlimporter.h"
+
+ITunesImporter::ITunesImporter(ITunesFeature* pParentFeature)
+        : m_pParentFeature(pParentFeature) {
+}
+
+bool ITunesImporter::canceled() {
```
Swiftb0y, line 8:
```suggestion
bool ITunesImporter::cancelled() const {
```

src/library/itunes/itunesimporter.h
Swiftb0y: whats the ownership relationship like now? We should either document that or use an explicit alias such as [`observer_ptr`](https://en.cppreference.com/w/cpp/experimental/observer_ptr).

Swiftb0y (changes requested):

Swiftb0y:
> I have also no explanation why it is not initialized during tests.

Apparently thats legal, even though many compilers still initialized it. https://stackoverflow.com/questions/36320008/whats-the-default-value-for-a-stdatomic

src/library/itunes/itunesfeature.h
Swiftb0y: I prefer in the header, its so simple and being able to quickly check whether the method is actually threadsafe has value too.

daschuer:
Done

Swiftb0y (commented):
Thank you. Should be good now after the build issues are fixed.

Swiftb0y (approved):

Issue #11657, opened 2023-06-13
Export Playlist: .CSV format is blank


bunkbedsss:
### Bug Description

When exporting a playlist in the .csv file format, much of the information is missing.  See below for example.
[npr 081322.csv](https://github.com/mixxxdj/mixxx/files/11736040/npr.081322.csv)

### Version

2.4.0

### OS

macOS ARM64

PR #11668, opened 2023-06-19
Export CSV: fixing null values in export

burnsak:
Improper location of right-alignment role caused fallthrough of export fields into the TextAlignment role instead of the DisplayRole leading null values in export.

Fixes: #11657 

Seems to be an unintended consequence of edits from #11634

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -590,21 +590,9 @@ QVariant BaseTrackTableModel::roleValue(
             break;
         }
         M_FALLTHROUGH_INTENDED;
-    // Right align BPM, duraation and bitrate so big/small values can easily be
-    // spotted by length (number of digits)
-    case Qt::TextAlignmentRole: {
-        switch (field) {
-        case ColumnCache::COLUMN_LIBRARYTABLE_BPM:
-        case ColumnCache::COLUMN_LIBRARYTABLE_DURATION:
-        case ColumnCache::COLUMN_LIBRARYTABLE_BITRATE: {
-            // We need to cast to int due to a bug similar to
-            // https://bugreports.qt.io/browse/QTBUG-67582
-            return static_cast<int>(Qt::AlignVCenter | Qt::AlignRight);
-        }
-        default:
-            return QVariant(); // default AlignLeft for all other columns
-        }
-    }
+    // NOTE: for export we need to fall through to Qt::DisplayRole,
+    // so do not add any other role cases here, or the export
+    // will be empty
     case Qt::DisplayRole:
         switch (field) {
         case ColumnCache::COLUMN_LIBRARYTABLE_DURATION: {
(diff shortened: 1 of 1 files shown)


ronso0:
Whoopsy...
Thanks for the quick fix. I'll take a look asap.

src/library/basetracktablemodel.cpp
```diff
@@ -873,6 +859,21 @@ QVariant BaseTrackTableModel::roleValue(
             return Qt::PartiallyChecked;
         }
     }
+    // Right align BPM, duraation and bitrate so big/small values can easily be
```
uklotzde, line 862: Please fix the typo while editing this code.

daschuer:
Thank you for this contribution. 
Before merge, we need your permission to distribute your changes. 
Please sign: https://docs.google.com/a/mixxx.org/spreadsheet/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ
and comment here when done.

burnsak:
I've signed the permission doc.

src/library/basetracktablemodel.cpp
```diff
@@ -872,6 +858,23 @@ QVariant BaseTrackTableModel::roleValue(
...
+        case ColumnCache::COLUMN_LIBRARYTABLE_DURATION:
+        case ColumnCache::COLUMN_LIBRARYTABLE_BITRATE: {
+            // We need to cast to int due to a bug similar to
+            // https://bugreports.qt.io/browse/QTBUG-67582
+            return static_cast<int>(Qt::AlignVCenter | Qt::AlignRight);
+        }
+        default:
+            return QVariant(); // default AlignLeft for all other columns
+        }
+        break;
```
ronso0, line 877: why break here?

ronso0 (changes requested):

src/library/basetracktablemodel.cpp
```diff
@@ -590,21 +590,7 @@ QVariant BaseTrackTableModel::roleValue(
...
-        case ColumnCache::COLUMN_LIBRARYTABLE_BITRATE: {
-            // We need to cast to int due to a bug similar to
-            // https://bugreports.qt.io/browse/QTBUG-67582
-            return static_cast<int>(Qt::AlignVCenter | Qt::AlignRight);
-        }
-        default:
-            return QVariant(); // default AlignLeft for all other columns
-        }
-    }
+        // NOTE: do not add any code here or the export will be empty
```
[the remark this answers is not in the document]
burnsak: I agree that is a much better comment than mine.

src/library/basetracktablemodel.cpp
burnsak: I added the break just for consistency with the other role case statements, and as a personal preference I always explicitly add a break at the end of a case in a switch statement. I agree it will never reach that break since there are returns prior to it.

ronso0: Yes, other role cases' default case would break to return the raw role value.
Though, IMO the break is obsolete for `Qt::CheckStateRole` and `Qt::TextAlignmentRole`.

burnsak: I agree the breaks are unnecessary for the two roles you mention. I'm fine reverting my additions of the breaks.

ronso0: Great. This is just a small change, so please squash the commits into one and force-push.

burnsak: Sorry, I don't know how to do that.

Swiftb0y: https://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

afterwards `git push --force` otherwise github will refuse the new commits since you rewrote the history.

ronso0:
Thanks for this fix, LGTM!

burnsak:
Thanks for the feedback and help with this.

PR #11673, opened 2023-06-22
fix OpenGL version detection

Be-ing:
This allows OpenGL waveforms to be used with the Wayland Qt Platform Abstraction on KDE and Weston. Though GL_MAJOR_VERSION and GL_MINOR_VERSION reported much higher versions (4.6 when tested on Intel UHD Graphics 770 (ADL-S GT1)), QSurfaceFormat reports 2.0 on KDE Wayland and Weston (somehow it reports 2.1 on GNOME Wayland though) and 2.1 with the Xcb QPA. QSurfaceFormat does not report higher versions if they are requested in the OpenGLWindow constructor.

This has only been tested on Linux. @m0dB can you test on macOS? @JoergAtGithub could you test on Windows?

### DIFF
--- a/src/waveform/waveformwidgetfactory.cpp
+++ b/src/waveform/waveformwidgetfactory.cpp
@@ -10,6 +10,7 @@
 
 #include <QGuiApplication>
 #include <QOpenGLFunctions>
+#include <QRegularExpression>
 #include <QStringList>
 #include <QTime>
 #include <QWidget>
@@ -71,6 +72,8 @@ bool shouldRenderWaveform(WaveformWidgetAbstract* pWaveformWidget) {
 
     return glw->shouldRender();
 }
+
+const QRegularExpression openGLVersionRegex(QStringLiteral("^(\\d+)\\.(\\d+).*$"));
 }  // anonymous namespace
 
 ///////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


Be-ing:
Finally with Wayland high DPI scaling Just Works automatically without needing to set the environment variable `QT_AUTO_SCREEN_SCALE_FACTOR=1`.

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -142,13 +142,29 @@ WaveformWidgetFactory::WaveformWidgetFactory()
...
+            qDebug().noquote() << QLatin1String(
+                    "OpenGL driver version string \"%1\", vendor \"%2\", "
+                    "renderer \"%3\"")
+                                          .arg(versionString, vendorString, rendererString);
+
+            GLint majorVersion = -1;
+            glFunctions->glGetIntegerv(GL_MAJOR_VERSION, &majorVersion);
+            GLint minorVersion = -1;
+            glFunctions->glGetIntegerv(GL_MINOR_VERSION, &minorVersion);
+            if (majorVersion == -1 || minorVersion == -1) {
```
Be-ing, line 154: If glGetIntegerv fails somehow I don't actually know what the value would be. I'm guessing it doesn't modify the passed in pointer, but I have no way to verify that.

JoergAtGithub:
On my Windows11 laptop I get:
```
debug [Main] OpenGL driver version string "4.6.0 Compatibility Profile Context 22.20.28.15.230324", vendor "ATI Technologies Inc.", renderer "AMD Radeon(TM) Graphics"
debug [Main] Supported OpenGL version: 4.6
```

src/waveform/waveformwidgetfactory.cpp
Swiftb0y: I guess the proper thing to do would be to check for `GL_INVALID_ENUM` errors? https://registry.khronos.org/OpenGL-Refpages/gl4/html/glGet.xhtml#errors
Also apparently you can rely on the `GL_VERSION` string to extract the version `GL_MAJOR/MINOR_VERSION` is not available...
https://www.khronos.org/opengl/wiki/OpenGL_Context#Context_information_queries

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -142,13 +142,29 @@ WaveformWidgetFactory::WaveformWidgetFactory()
...
                     reinterpret_cast<const char*>(glFunctions->glGetString(GL_RENDERER))));
-            qDebug() << versionString << vendorString << rendererString;
-
-            // note: the requested version has been set in WGLWidget's OpenGLWindow constructor
-            const int majorVersion = context->surface()->format().majorVersion();
-            const int minorVersion = context->surface()->format().minorVersion();
+            qDebug().noquote() << QStringLiteral(
+                    "OpenGL driver version string \"%1\", vendor \"%2\", "
+                    "renderer \"%3\"")
+                                          .arg(versionString, vendorString, rendererString);
```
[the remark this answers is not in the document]
Be-ing: That's what it was like before but the formatting was less than ideal:
`OpenGL driver version string "VERSION-STRING" vendor "VENDOR-STRING" renderer "RENDER-STRING"`

This way allows inserting the the commas between the fields:
`OpenGL driver version string "VERSION-STRING", vendor "VENDOR-STRING", renderer "RENDER-STRING"`

Using `<<` would insert awkward spaces before the commas:

``OpenGL driver version string "VERSION-STRING" , vendor "VENDOR-STRING" , renderer "RENDER-STRING"`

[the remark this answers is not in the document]
Swiftb0y: Good point. don't really like that the resulting code is harder to read, but not worth arguing over.

src/waveform/waveformwidgetfactory.cpp
Be-ing: done

Swiftb0y (approved):
LGTM. can confirm this to also fix the issue on Fedora 38 (Gnome running XWayland). There are other issues on Wayland though I'll file in separate gh issues.

This probably doesn't make any difference on MacOS, but lets wait with merging until someone confirms it.

m0dB:
I'll test on macOS this weekend.

m0dB:
It seems to work fine on macOS. However, I wanted to check what the preferences dialogs shows... but the dialog doesn't open. Not caused by this PR, it also happens with 2.4 (after pulling and installing the new buildenv) . Is anybody else experiencing this? Any idea what may cause this?

JoergAtGithub:
LGTM! Thank you!

PR #11676, opened 2023-06-23
Do not write Debug logs unconditionally into file

uklotzde:
Mixxx produces huge debug log files and there is currently no way to prevent this. Instead logging should adhere to the given options.

### DIFF
--- a/src/util/logging.cpp
+++ b/src/util/logging.cpp
@@ -266,10 +266,6 @@ namespace mixxx {
 
 namespace {
 
-bool isControllerLoggingCategory(const QString& categoryName) {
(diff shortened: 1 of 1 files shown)


daschuer (changes requested):
Why is that actually a problem? How much disk do you like to save? This file has already safed our back in case of sudden errors. So it is IMHO a good idea to keep it enabled by default.

Swiftb0y:
Frequent debug logs combined with long sessions can become a serious diskspace problem... Though I recognize the value in having the debug log. Still its only useful in a tiny fraction of cases, so I support to only opt in to the debug log when we actually need it.

PR #11679, opened 2023-06-23
fix the preferences menu and opening the settings directory on macOS

m0dB:
- fix the Preferences menu by avoiding text-heuristics treating "Settings directory" as Preferences (macOS), by change the default menu role QAction::TextHeuristicsRole to QAction::NoRole
- fix opening the folder by using QUrl::fromLocalFile (this required changing the argument of slotVisitUrl from QString to QUrl, which makes more sense anyway)

### DIFF
--- a/src/widget/wmainmenubar.cpp
+++ b/src/widget/wmainmenubar.cpp
@@ -578,8 +578,9 @@ void WMainMenuBar::initialize() {
     auto* pHelpSupport = new QAction(supportTitle, this);
     pHelpSupport->setStatusTip(supportText);
     pHelpSupport->setWhatsThis(buildWhatsThis(supportTitle, supportText));
-    connect(pHelpSupport, &QAction::triggered,
-            this, [this] { slotVisitUrl(MIXXX_SUPPORT_URL); });
+    connect(pHelpSupport, &QAction::triggered, this, [this] {
+        slotVisitUrl(QUrl(MIXXX_SUPPORT_URL));
+    });
     pHelpMenu->addAction(pHelpSupport);
 
     // User Manual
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
I confirm, that this PR works on Windows as 2.4.
This is a string change which affects translation - but it seems we need to change the string to fix this bug. Are we in string freeze phase?
Regarding the string itself, I would suggest to use "Mixxx settings directory" instead of "User settings directory", since we point to a Mixxx specific subdirectory of the (Windows) user settings directory. Other tools I know, use terms like "Open file location" for similar operations. Maybe "Open Mixxx settings directory" is better to recognize therefore?

m0dB:
I agree that Mixxx settings directory sounds better.

m0dB:
I don't know about string freeze, but mind you that this was added only days ago.

m0dB:
Let me see if I can avoid the Settings / Preferences mix-up without changing the string at all.

m0dB:
Ok, fixed without having to change the string (aka fixed correctly)

```pHelpSettingsDir->setMenuRole(QAction::NoRole);```
did the trick.

By default the menuRole is QAction::TextHeuristicRole which means that "This action should be put in the application menu based on the action's text as described in the QMenuBar documentation." 

Which lead to "&Settings directory" being treated as Preferences.

JoergAtGithub:
LGTM! Works on Windows11 as expected. Waiting for CI. Thank you!

ronso0:
@m0dB Thanks for debugging this! No regression on Linux afaict. LGTM

Lesson learned: any PR touching the menubar should be tested manually on macOS @mixxxdj/developers 

@JoergAtGithub Feel free to rename to "Mixxx settings directory". Better sooner than later, before translators start to work on 2.4-beta.

m0dB:
Now that this has been fixed without renaming, I don't think renaming to "Mixxx settings ..." is necessary and it actually seems a bit redundant. It should be pretty obvious what settings it refers too!

JoergAtGithub:
"Settings directory" and "Mixxx settings directory" work both for me. Just "User settings directory" sounds wrong to me.

Swiftb0y:
Sorry introducing the regression. Many thanks for fixing it this quickly.  

> Lesson learned: any PR touching the menubar should be tested manually on macOS https://github.com/orgs/mixxxdj/teams/developers

Agree. I'll be more careful next time.

Issue #11617, opened 2023-06-02
waveform scrolling


robi970:
### Bug Description

I installed Mixxx 2.4 beta and started it.
I quickly realized that the waveforms, spinnys and elapsed time display are displayed as if "4-5 frames per second" is set.

on the same PC I keep Mixxx 2.2.4 installed, which obviously still works perfectly

### Version

2.4 beta

### OS

Win10 pro 64bit - AMD ryzen 5 3600; 16GB ram; AMD radeon R9 200 series

JoergAtGithub:
- Could you please share a screenshot of the Waveforms tab in the Mixxx preferences!
- How much CPU and GPU load does the Windows Task Manger show?
- Did you also tried Mixxx 2.3 builds, or only 2.2 and 2.4 beta?

JoergAtGithub:
Could you please try all waveform types available in the top selector. Does any waveform type shows different frame rates?

robi970:
with all waveforms the value is always around 6.67, except with HSV (FPS: 7.50)

same thing also changing the values ​​of the second selector

(OT question: in this editor is it possible to write in your own language, or do I have to continue with google translate?)

JoergAtGithub:
Please write in English!

JoergAtGithub:
@m0dB Could you have a look

JoergAtGithub:
Does it help if you start with `.\mixxx.exe --disableVuMeterGL`

robi970:
switched to 15 fps

m0dB:
I have no idea, this is very strange and the first time something like this has been reported.

To be sure: 2.3.5 runs as 60 fps?

I will create a build with some logging, but it might take a while, sorry.

m0dB:
Are you having no fps issues with other applications (games)?

Question for @daschuer : are we configuring Qt on windows with 
configure -opengl dynamic ?

JoergAtGithub:
It's dynamic and ANGLE is used when the OPENGL driver is missing or blacklisted by Qt. If ANGLE is used, you see it clearly in the driver information in the waveform preferences.

m0dB:
Thanks, @JoergAtGithub ! So I understand that in this case ANGLE isn't used because a proper OpenGL driver is detected. It might be interesting to see if forcing ANGLE makes a difference, right? Have you ever tried that yourself?

[...]

PR #11681, opened 2023-06-25
Use set interval setting to fix performance degration for AMD graphics adapters

JoergAtGithub:
Use set interval setting to fix performance degration for AMD graphics adapters on Windows with "Wait for vertical refresh" not set to "Allways off" in the driver settings.
see: https://github.com/mixxxdj/mixxx/issues/11617#issuecomment-1606106257

Closes: #11617

### DIFF
--- a/src/widget/openglwindow.cpp
+++ b/src/widget/openglwindow.cpp
@@ -14,6 +14,29 @@ OpenGLWindow::OpenGLWindow(WGLWidget* pWidget)
     QSurfaceFormat format;
     format.setVersion(2, 1);
     format.setProfile(QSurfaceFormat::CoreProfile);
+
+    // setSwapInterval sets the application preferred swap interval
+    // in minimum number of video frames that are displayed before a buffer swap occurs
+    // - 0 will turn the vertical refresh syncing off
+    // - 1 (default) means swapping after drawig a video frame to the buffer
+    // - n means swapping after drawing n video frames to the buffer
+    //
+    // The vertical sync setting requested by the OpenGL application, can be overwritten
+    // if a user changes the "Wait for vertical refresh" setting in AMD graphic drivers
+    // for Windows.
+
+#if defined(__APPLE__)
+    // On OS X, syncing to vsync has good performance FPS-wise and
+    // eliminates tearing. (This is an comment from pre QOpenGLWindow times)
+    format.setSwapInterval(1);
+#else
+    // It seems that on Windows (at least for some AMD drivers), the setting 1 is not
+    // not properly handled. We saw frame rates divided by exact integers, like it should
+    // be with values >1 (see https://github.com/mixxxdj/mixxx/issues/11617)
+    // Reported as https://bugreports.qt.io/browse/QTBUG-114882
+    // On Linux, horrible FPS were seen with "VSync off" before switching to QOpenGLWindow too
+    format.setSwapInterval(0);
+#endif
     setFormat(format);
 }
 


m0dB:
Great! Thanks for all your time and persistence getting to the bottom of this!

daschuer:
I have not tested it yet. But I could remember that this was a mayor issue when introducing the vsync code for waveforms.
The debug code for this is also still in place. 
m_vSyncMode == ST_FREE
With this you should see the maximum possible frame rate. If this is limited to 60, we know that the swap is artificialy delayed by the system settings, which should not happen.

JoergAtGithub:
It's already confirmed by one of the two affected users, that this is the cause of the issue. It's easy to test, because the AMD driver allows to overwrite the setting in the application.
While we know the correct setting for Windows now, testing for the correct setting on MacOS and Linux is needed. Without this explicit setting in the application, we depend on the default of the graphics adapter driver.
This PR contains the same settings per OS, which we have in 2.3.

Swiftb0y:
Can't spot any regressions on Fedora 38 under `4.6 (Compatibility Profile) Mesa 23.1.2 (Mesa Intel(R) UHD Graphics (CML GT2))`. Both native wayland and xwayland work equally well (though the xwayland layer causes about 10% extra gpu load).

src/widget/openglwindow.cpp
```diff
@@ -14,6 +14,14 @@ OpenGLWindow::OpenGLWindow(WGLWidget* pWidget)
     QSurfaceFormat format;
     format.setVersion(2, 1);
     format.setProfile(QSurfaceFormat::CoreProfile);
+#if defined(__APPLE__)
+    // On OS X, syncing to vsync has good performance FPS-wise and
+    // eliminates tearing.
+    format.setSwapInterval(1);
+#else // Otherwise, turn VSync off because it could cause horrible FPS on Linux
+      // and Windows depending on the graphics driver settings
```
Be-ing, line 22: Please add a link to the bug report.

JoergAtGithub: Done!

src/widget/openglwindow.cpp
```diff
@@ -14,12 +14,29 @@ OpenGLWindow::OpenGLWindow(WGLWidget* pWidget)
...
-#else // Otherwise, turn VSync off because it could cause horrible FPS on Linux
-      // and Windows depending on the graphics driver settings
+#else
+    // It seems that on Windows (at least for some AMD drivers), the setting 1 is not
+    // not properly handled. We saw frame rates divided by exact integers, like it should
+    // be with values >1 (see https://github.com/mixxxdj/mixxx/issues/11617)
+    // Reported as https://bugreports.qt.io/browse/QTBUG-114882
+    // Otherwise, turn  because it could cause horrible FPS on Linux
+    // and Windows depending on the graphics driver settings.
+    // On Linux, horrible FPS were seen with "VSync off" before switching to QOpenGLWindow too
```
Swiftb0y, line 39: ?

JoergAtGithub: Line 37/38 were the old comment - missed to delete them

Swiftb0y (approved):
LGTM. do you want to squash before merging?

JoergAtGithub:
Ok, will do!

Swiftb0y:
Do we still need a manual test on macOS?

fwcd:
Performance seems fine to me from a quick smoke test on macOS ARM.

Swiftb0y:
Thanks for the confirmation.

PR #11684, opened 2023-06-27
Fix clang-tidy warnings for generated code 

daschuer:
The original attempt to exclude the kaitai struct file via CMake and CXX_CLANG_TIDY did not work, because this is a target property. A fix would be to create a target for them which clutters the build process and adds maintenance effort. 

The slim alternative is to move the files to an extra folder and add a .clang-format config file to that folder that disables all checks. This has been implemented here.


JoergAtGithub:
LGTM! Thank you!

PR #11689, opened 2023-06-28
Disable false positive clang-tidy warning

daschuer:
The proposed fix does not compile with MSVC and seems to be confusing anyway:
`'const auto result' can be declared as 'const auto *const result'`

The use of auto is however required here, because the type of "all_versions" is `std::array<engine_version, 19>` which may be a subject to change. 
https://github.com/xsco/libdjinterop/blob/8a32c822c1422843f521e2cdd16e36e7530e1d93/include/djinterop/engine/engine.hpp#L141


daschuer:
This one is obsolete since we have merged https://github.com/mixxxdj/mixxx/pull/11720

PR #11696, opened 2023-06-30
on macos, set preferences dialog title to the selected page title

m0dB:
On macOS Ventura / SDK 13, the "Preferences..." application menu item has been renamed to "Settings...". This leads to an inconsistency with Mixxx preferences dialog title, which is still "Preferences".

According to Apple's Human Interface Guidelines, however, the title of the dialog should actually be neither Preferences nor Settings. See https://developer.apple.com/design/human-interface-guidelines/patterns/settings/ which reads:

> Update the window’s title to reflect the currently visible pane. If your settings window doesn’t have multiple panes, use the title App Name Settings.

As we do have multiple panes in the dialog (called pages, in the mixxx source code), where the current page is selected from the list (tree item) on the left, it makes sense to use that. This way we kill two birds with one stone: we don't have to worry about a inconsistency between the menu item and the dialog title, and we integrate more correctly in the host OS.

This PR sets the dialog title to the text (pageTitle) from the selected item from the tree. E.g. "Sound hardware", "Library", "Controllers". In the case of "Controllers", when a specific controller is selected from the subtree, this is concatenated, e.g. "Controllers - DJControl Inpulse 200". The dialog title is only changed on macOS, Linux and Windows behave like before.

### DIFF
--- a/src/controllers/dlgprefcontrollers.cpp
+++ b/src/controllers/dlgprefcontrollers.cpp
@@ -109,13 +109,16 @@ bool DlgPrefControllers::handleTreeItemClick(QTreeWidgetItem* clickedItem) {
     if (controllerIndex >= 0) {
         DlgPrefController* pControllerDlg = m_controllerPages.value(controllerIndex);
         if (pControllerDlg) {
-            m_pDlgPreferences->switchToPage(pControllerDlg);
+            const QString pageTitle = m_pControllersRootItem->text(0) + " - " +
+                    clickedItem->text(0);
+            m_pDlgPreferences->switchToPage(pageTitle, pControllerDlg);
         }
         return true;
     } else if (clickedItem == m_pControllersRootItem) {
         // Switch to the root page and expand the controllers tree item.
         m_pDlgPreferences->expandTreeItem(clickedItem);
-        m_pDlgPreferences->switchToPage(this);
+        const QString pageTitle = clickedItem->text(0);
+        m_pDlgPreferences->switchToPage(pageTitle, this);
         return true;
     }
     return false;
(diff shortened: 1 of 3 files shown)


daschuer:
The Gnome Control Center has the same behavior.
Should we enable it for all targets?

daschuer:
Works and looks good. 
Libre-Office has a kind of breadcrumb "Optionen - Libre Office - Drucken"

m0dB:
> Should we enable it for all targets?

That makes sense to me, and certainly if there is a precedent like Gnome Control Center on Linux.

ronso0:
my 2 cents:
I understand that it's an Apple HIG and I'm okay with adding this to macOS for the sake of consistency.
It also makes sense for apps like the Gnome Control Center which lack a persistent toolbar.

The Mixxx preferences though do have a toolbar so personally I don't see the benefit. I prefer distinct window titles when Alt-Tab'ing through apps. For mulitple web browser windows the current page title is helpful since all browser windows are equal.
For Mixxx however it makes more sense to me to have `Artist - Title Remix | Mixxx` and `Preferences` than `Library` or `Controllers - Reloop TerminalMix 2` for example.

m0dB:
I see your point, @ronso0, and I am not saying this is ideal, or that the HIG is without issues. But it definitely has resulted in a much more consistent user experience on macOS than the UI jungle of Linux and Windows. And I think that in this case, following the HIG on macOS makes sense. So how about we restrict this to macOS for now?

(It is common practice, even with toolbars. See for example Safari's Preferences dialog. Even though it has a toolbar, the window title duplicates the selected item in the toolbar.)

Screenshot 2023-07-01 at 14 21 01

In the future, we can expand on this feature. On Linux and/or Windows we could for example add a prefix, e.g. "Preferences - Sound hardware" or "Mixxx Preferences - Sound hardware", if that would be more of your liking. Or just leave it as is.

ronso0:
> So how about we restrict this to macOS for now?

Sure, I don't object.

> See for example Safari's Preferences dialog

macOS, I hate floating toolbars :p

> "Mixxx Preferences - Sound hardware"

Yep, that would work for me.

ronso0:
LGTM, thanks.

@daschuer Any more comments?

daschuer (approved):
No more comments. Thank you.

PR #11698, opened 2023-07-01
use QOpenGLWindow::update to trigger the WVuMeter rendering

m0dB:
Instead of rendering and swapping the opengl-based WVuMeter directly from WaveformWidgetFactory, call updateGL (through QOpenGLWindow::update()), similar to the mechanism used for WVuMeterLegacy.

### DIFF
--- a/src/waveform/waveformwidgetfactory.cpp
+++ b/src/waveform/waveformwidgetfactory.cpp
@@ -423,22 +423,18 @@ void WaveformWidgetFactory::addVuMeter(WVuMeterLegacy* pVuMeter) {
     // We don't activate update() or repaint() directly so listener widgets
     // can decide whether to paint or not.
     connect(this,
-            &WaveformWidgetFactory::waveformUpdateTick,
+            &WaveformWidgetFactory::maybeUpdateVuMeters,
             pVuMeter,
             &WVuMeterLegacy::maybeUpdate,
             Qt::DirectConnection);
 }
 
 void WaveformWidgetFactory::addVuMeter(WVuMeterBase* pVuMeter) {
-    // WVuMeterGLs to be rendered and swapped from the vsync thread
     connect(this,
-            &WaveformWidgetFactory::renderVuMeters,
+            &WaveformWidgetFactory::maybeUpdateVuMeters,
             pVuMeter,
-            &WVuMeterBase::render);
-    connect(this,
-            &WaveformWidgetFactory::swapVuMeters,
-            pVuMeter,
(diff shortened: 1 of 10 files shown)


daschuer:
I can confirm that this branch performs still good with Ubuntu Focal and 
"4.6 (Compatibility Profile) Mesa 21.2.6 (Mesa Intel(R) UHD Graphics (CML GT2))"

m0dB:
Thanks for confirming, @daschuer . Let's see if @JoergAtGithub and/or @NotYourAverageAl can test this on Windows.

(I have some additional tweaks, but I want to go one step at a time)

m0dB:
fixed formatting issue

JoergAtGithub:
Sounds promising. I will test it later this week!

NotYourAverageAl:
@m0dB  You wont believe this but I can't reproduce this bug on 2.4. I updated my chipset driver and got a Windows update [kb5027292](https://support.microsoft.com/en-us/topic/june-28-2023-kb5027292-os-build-22000-2124-preview-11074490-ad96-4660-b988-1b507ccab80c) and cleaned out my mixxx settings folder. Apologies for all the noise, possibly just a system issue.

To note this PR works perfectly and I don't have any lag.

JoergAtGithub:
I tested this PR on my Windows 11 system and still see the lag in the UI. It might be, that this PR is slightly more responsive than the 2.4 branch, but with `--disable-vumeterGL` both branches are significiantly more responsive.

m0dB:
Ok, that's too bad. And strange that you are both seeing something else.

I found another solution: on macOS, --disable-vumetergl behaves badly. But: I believe this is basically caused by the use of QStylePainter. If I replace: 
```
    QStyleOption option;
    option.initFrom(this);
    QStylePainter p(this);
    p.drawPrimitive(QStyle::PE_Widget, option);
```
with a simple
```
    QPainter p(this);
```
it runs smooth.

I believe the only thing the QStylePainter is used for is to set the background color. I already am determining the background color differently for the GL VuMeter so we could do the same here.

I will create a PR with this change. If this works well everywhere, I would prefer this because it would allow us to have the same implementation on all platforms.

m0dB:
Done in https://github.com/mixxxdj/mixxx/pull/11722 . This disables the GL based VuMeter by default and enables an improved legacy VuMeter (no QStylePainter).

JoergAtGithub:
I think this PR can be closed?

m0dB:
yes! thanks.

PR #11104, opened 2022-12-02
Traktor Kontrol S2MK3: Use FX select buttons to set quick effect presets

echozio:
This repurposes the "FX SELECT" buttons on the controller to change the loaded quick effect preset rather than toggle effect chains. This allows them to be used in conjunction with the filter knobs on either side.

Pressing a button will load the preset at that index of the quick effect preset list (pressing button 1 will load the first preset, 2 the second, etc.). Holding down shift while pressing a button will only change the preset for the deck whose shift button was used. Buttons will light up accordingly, and if two different presets are active on one deck each, both of the buttons will light up.

One issue now is that changing the effect in the GUI won't change the LED status of the buttons, but I'm not sure how feasible this would even be to implement.

PR #11198, opened 2023-01-13
Add super knob adoption option for quick effect chain presets

robbert-vdh:
This is similar to the existing option for meta knobs. This option is needed for DJ controllers that allow the user to apply different quick effect chain presets to tracks, like the Mixer FX section on the Native Instruments Traktor Kontrol line of controllers. The idea with those controllers is that you can pick from a plain filter, or that same filter plus an effect (they have a delay, reverb, white noise, and trance gate), and assign that to a channel. You'd usually select the effect while the filter knob is still in its neutral 12 o' clock position, but if it is not then the filter frequency settings should be preserved and the other effect should merely be added on top of that. I have a Traktor Kontrol S3 script improvement ready that adds this exact feature to Mixxx. That also avoids the potential issue where an effect chain's default super knob position isn't quite at 0.5 and you need to fiddle around with the filter knob to get the soft takeover to work.

This PR doesn't change the default behavior, so if you don't enable the option nothing changes. This change also uncovered another issue with Mixxx's effect handling where changing these chain presets will not immediately switch between presets (when the new one is done loading)i. Instead it removes the old one, and it then takes one or more buffers before the new chain takes effect. Fixing this will require more investigation in the future.

I've added the preference next to the existing meta knob preference using the same style. By the way, shouldn't this be 'meta knob' and 'super knob' instead of 'metaknob' and 'superknob'?
image

PR #10859, opened 2022-09-02
Effects: allow clearing chains with empty '---' preset

ronso0:
Alternative to #4892 
This implementation works ~~without hacks~~ only minor adjustments in the GUI counterparts (EQ / effects preferences, skins).

An empty preset named  `---` is created by `EffectChainPresetManager` on each start.
* it is added at the top of the common chain preset list and the QuickEffect preset list
* it is shown in the chain preset selectors in Preferences > EQ & `WEffectChainPresetSelector`
* it is **hidden** from the preset lists in Preferences > Effects to avoid manipulation
when those lists are changed they are pushed back to `EffectChainPrestManager` to update the lists in effect widgets, `---` is missing and needs to be re-added to the standard and QuickEffect preset lists
* it is **not** saved to `[settings_dir]/effects/chains`
* `---` is not accepted as preset name when saving or importing chain presets to avoid conflicts
* `---` can be selected with `[fx chain], chain_preset_selector` and `.., next/prev_chain_preset`
* `loaded_chain_preset` called with `0` will now clear the chain

**Update:**
`---` is only added to the QuickEffect chain presets list to avoid a few incosistencies with standard chains.

~~When picked in **standard chains**' preset selectors, [EffectChain::loadChainPreset](https://github.com/mixxxdj/mixxx/pull/10859/files#diff-fa28becfc988b55ba9c54ec2aee6524f22af4a27efb9a4d18a2a9fe121ddab35R219-R230) will clear the name (more details in https://github.com/mixxxdj/mixxx/pull/10859#discussion_r962314889), so~~
* ~~`chainPresetChanged()` signal will clear the selector combobox (index = -1) and the selection in the chain preset menu~~
* ~~the empty preset can not be overwritten via 'Update Preset' in the `WEffectChainPresetButton` menu~~
* ~~it is **not** stored as preset name for standard effect chains in `effects.xml`~~

When picked in **QuickEffect chains**' preset selectors or in Preferences > Equalizer, the name `---` persists
* `---` remains selected in preset selectors to signal the current chain is empty (and will remain empty because there's no (skin) interface to load effects into chain slots)
* it is used as preset identifier for QuickEffect chains in `effects.xml`:
    ```
     <QuickEffectChains>
      <ChainPresetName group="[Channel1]">Echo</ChainPresetName>
      <ChainPresetName group="[Channel2]">---</ChainPresetName>
      <ChainPresetName group="[Channel3]">Filter</ChainPresetName>
[...]

PR #11702, opened 2023-07-01
Traktor Kontrol S2MK3: Use FX select buttons to set quick effect presets

olivier-mauras:
Rework of https://github.com/mixxxdj/mixxx/pull/11104 on 2.4 branch

### DIFF
--- a/res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
+++ b/res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
@@ -1,6 +1,11 @@
 ///////////////////////////////////////////////////////////////////////////////////
 // JSHint configuration                                                          //
 ///////////////////////////////////////////////////////////////////////////////////
+/* global engine                                                                 */
+/* global script                                                                 */
+/* global HIDDebug                                                               */
+/* global HIDPacket                                                              */
+/* global HIDController                                                          */
 /* jshint -W016                                                                  */
 ///////////////////////////////////////////////////////////////////////////////////
 /*                                                                               */
@@ -674,15 +679,41 @@ TraktorS2MK3.fxHandler = function(field) {
         return;
     }
 
-    const fxNumber = parseInt(field.id[field.id.length - 1]);
-    const group = "[EffectRack1_EffectUnit" + fxNumber + "]";
+    const availableGroups = [ "[Channel1]", "[Channel2]" ];
+    let targetGroups = [];
+    let fxNumber = parseInt(field.id[field.id.length - 1]);
 
-    // Toggle effect unit
-    TraktorS2MK3.fxButtonState[fxNumber] = !TraktorS2MK3.fxButtonState[fxNumber];
+    // Target decks whose shift button is pressed.
+    for (let group of availableGroups) {
+        if (TraktorS2MK3.shiftPressed[group]) targetGroups.push(group);
+    }
 
-    engine.setValue(group, "group_[Channel1]_enable", TraktorS2MK3.fxButtonState[fxNumber]);
-    engine.setValue(group, "group_[Channel2]_enable", TraktorS2MK3.fxButtonState[fxNumber]);
-    TraktorS2MK3.outputHandler(TraktorS2MK3.fxButtonState[fxNumber], field.group, "fxButton" + fxNumber);
+    // If no shift button was pressed fall back to target all decks.
+    if (targetGroups.length === 0) targetGroups = availableGroups;
+
+    for (let group of targetGroups) {
+        engine.setValue(`[QuickEffectRack1_${group}]`, "loaded_chain_preset", fxNumber);
+    }
+};
+
+TraktorS2MK3.fxOutputHandler = function() {
+    const fxButtonCount = 4;
+    const availableGroups = [ "[Channel1]", "[Channel2]" ];
+
+    let activeFx = {};
+    for (let group of availableGroups) {
+        activeFx[group] = engine.getValue(`[QuickEffectRack1_${group}]`, "loaded_chain_preset");
+    }
+
+    /* There is no way on the controller to indicate which deck the effect applies *
+     * to, but keeping both lit indicates that different effects are in use.       */
+    for (let fxButton = 1; fxButton <= fxButtonCount; ++fxButton) {
+        let active = false;
+        for (let group of availableGroups) {
+            active = active || activeFx[group] === fxButton;
+        }
+        TraktorS2MK3.outputHandler(active, "[ChannelX]", "fxButton" + fxButton);
+    }
 };
 
 TraktorS2MK3.reverseHandler = function(field) {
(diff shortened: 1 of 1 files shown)


ronso0:
Thank you!
Did also take a look at the reviews of #11104 ?

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
```diff
@@ -670,19 +670,92 @@ TraktorS2MK3.wheelDeltas = function(deckNumber, value) {
...
+    if (targetGroups.length === 0) {
+        targetGroups = availableGroups;
+        noShift = true;
+    }
+
+    const fxToApply = {};
+    const activeFx = {};
+    // Detect which fx should be enabled
+    for (const group of targetGroups) {
+        activeFx[group] = engine.getValue(`[QuickEffectRack1_${group}]`, "loaded_chain_preset");
```
echozio, line 702:
```suggestion
        activeFx[group] = engine.getValue(`[QuickEffectRack1_${group}]`, "loaded_chain_preset") - 1;
```

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
```diff
@@ -670,19 +670,92 @@ TraktorS2MK3.wheelDeltas = function(deckNumber, value) {
...
+            fxToApply[deck2] = fxToApplyAll;
+        }
+    }
 
-    engine.setValue(group, "group_[Channel1]_enable", TraktorS2MK3.fxButtonState[fxNumber]);
-    engine.setValue(group, "group_[Channel2]_enable", TraktorS2MK3.fxButtonState[fxNumber]);
-    TraktorS2MK3.outputHandler(TraktorS2MK3.fxButtonState[fxNumber], field.group, "fxButton" + fxNumber);
+    // Now apply the new fx value
+    for (const group of targetGroups) {
+        engine.setValue(`[QuickEffectRack1_${group}]`, "loaded_chain_preset", fxToApply[group]);
```
echozio, line 735:
```suggestion
        engine.setValue(`[QuickEffectRack1_${group}]`, "loaded_chain_preset", fxToApply[group]) + 1;
```

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
```diff
@@ -670,19 +670,92 @@ TraktorS2MK3.wheelDeltas = function(deckNumber, value) {
...
+    }
+};
+
+TraktorS2MK3.fxOutputHandler = function() {
+    const fxButtonCount = 4;
+    const availableGroups = ["[Channel1]", "[Channel2]"];
+
+    const activeFx = {};
+    for (const group of availableGroups) {
+        activeFx[group] = engine.getValue(`[QuickEffectRack1_${group}]`, "loaded_chain_preset");
```
echozio, line 745:
```suggestion
        activeFx[group] = engine.getValue(`[QuickEffectRack1_${group}]`, "loaded_chain_preset") - 1;
```

echozio (changes requested):
I tested this out with the controller, and the 8 effect logic works intuitively, however pressing 1 sets the effect at index 0, which is no effect (not visible when reordering in the settings).
This change should fix that.

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
olivier-mauras: Actually I didn't want to sort it out this way, because I believe there's probably a bug in the quick effects settings page.
in 2.4-beta - at least for me: 

- the list order is never saved upon Mixxx exit and needs to be changed after restart
- List index does not start with the first effect in the list but with "---" - If you get to the effect settings, hit "Restore Defaults", then the first index becomes the first effect

I believe that working around this behavior in the mapping will thus break next time the effect list behavior is sorted out

ronso0:
Yes, I added the empty preset in #10859 so the list is 1-based.
Some (all?) new effect controls are not documented , yet, see https://github.com/mixxxdj/manual/issues/501

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
ronso0: > the list order is never saved upon Mixxx exit and needs to be changed after restart

Can't confirm. The list is saved correctly to effects.xml when Mixxx exits. Are you sure Mixxx is shutting correctly?

echozio: > List index does not start with the first effect in the list but with "---" - If you get to the effect settings, hit "Restore Defaults", then the first index becomes the first effect
I would imagine this is a bug in the restore logic ("---" is not re-added when restoring)?

olivier-mauras: Mmmmh I must have badly tested, you're right list is properly saved, but "---" disappear on "Restore Defaults" which would break the "workaround" mapping though.
Which way should we fix it first? :)

olivier-mauras:
Ok so I tested @echozio fix and added a new commit.
Works like a charm, just don't expect the behavior to be correct right after reloading defaults :)

echozio (approved):

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
ronso0: I'll take a look.

ronso0:
Fix for the missing empty preset / list offset is here #11705

ronso0:
If mappings are created/updated the respective manual PR is a prerequisite for merging.
https://manual.mixxx.org/2.3/en/hardware/controllers/native_instruments_traktor_kontrol_s2_mk3.html
@olivier-mauras are you familiar with the manual repo?

olivier-mauras:
@ronso0 https://github.com/mixxxdj/manual/pull/568

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
```diff
@@ -670,19 +670,92 @@ TraktorS2MK3.wheelDeltas = function(deckNumber, value) {
...
+    // Target decks whose shift button is pressed.
+    for (const group of availableGroups) {
+        if (TraktorS2MK3.shiftPressed[group]) { targetGroups.push(group); }
+    }
+
+    // If no shift button was pressed fall back to target all decks.
+    if (targetGroups.length === 0) {
+        targetGroups = availableGroups;
+        noShift = true;
+    }
```
JoergAtGithub, line 696: Wouldn't it make sense to check here that the effect knob is in zero position (or close to it)? Otherwise we could reassign the effect chain preset, while it's audible.

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
ronso0: This was fixed by #11705

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
ronso0: IMO this is up to the user to decide, and the learing curve should be pretty steep ;)
Related: #11198

JoergAtGithub:
@ronso0 Than I think we can merge this. But in https://github.com/mixxxdj/manual/pull/568 you asked to have a final look, when the mapping PR is ready.

ronso0:
Indeed, I'll do so soonish.

res/controllers/Traktor-Kontrol-S2-MK3-hid-scripts.js
echozio: I'm not sure how this works currently, but when I used the code from my old PR knobs would always reset when switching effect, whether that be to the center position or the zero position and the knob would only control it again when moved back to the initial point.

JoergAtGithub:
LGTM! Thank you!

daschuer:
Hi @olivier-mauras thank you for your contribution.
It looks like we have missed to ask for your permission to distribute your code before merge. Can you catch that up, sign https://docs.google.com/a/mixxx.org/spreadsheet/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ and comment here when done? I will then put your name to the Mixxx 2.4 contributor list in the Mixxx about box.

PR #11705, opened 2023-07-02
Effects: re-add empty chain preset when resetting to defaults

ronso0:
re-add the empty preset after clicking "Reset to defaults" in Pref > Effects

### DIFF
--- a/src/effects/presets/effectchainpresetmanager.cpp
+++ b/src/effects/presets/effectchainpresetmanager.cpp
@@ -36,6 +36,18 @@ EffectChainPresetPointer loadPresetFromFile(const QString& filePath) {
     return pEffectChainPreset;
 }
 
+EffectChainPresetPointer createEmptyChainPreset() {
+    EffectManifestPointer pEmptyManifest(new EffectManifest());
+    pEmptyManifest->setName(kNoEffectString);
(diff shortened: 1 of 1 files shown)


src/effects/presets/effectchainpresetmanager.cpp
```diff
@@ -593,10 +593,27 @@ void EffectChainPresetManager::resetToDefaults() {
...
+    // Re-add the empty chain preset
+    EffectChainPresetPointer pEmptyChainPreset = createEmptyChainPreset();
+    m_effectChainPresets.insert(pEmptyChainPreset->name(), pEmptyChainPreset);
+    m_quickEffectChainPresetsSorted.prepend(pEmptyChainPreset);
+
     emit effectChainPresetListUpdated();
     emit quickEffectChainPresetListUpdated();
 }
 
+EffectChainPresetPointer EffectChainPresetManager::createEmptyChainPreset() {
```
Swiftb0y, line 605: doesn't need to be a member function. Can you make this a free function and move it to an anonymous namespace at the top of this file?

ronso0: sure.

daschuer (approved):
LGTM, Thank you.

PR #11706, opened 2023-07-02
Skins, library: draw focus border around BPM delegate

ronso0:
forgot that in #3172

### DIFF
--- a/src/library/bpmdelegate.cpp
+++ b/src/library/bpmdelegate.cpp
@@ -87,4 +87,24 @@ void BPMDelegate::paintItem(QPainter* painter,const QStyleOptionViewItem &option
                                m_pCheckBox);
         }
     }
+
+    // Draw a border if the color cell has focus
+    if (option.state & QStyle::State_HasFocus) {
+        // This uses a color from the stylesheet:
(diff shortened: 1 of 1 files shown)


src/library/bpmdelegate.cpp
```diff
@@ -87,4 +87,24 @@ void BPMDelegate::paintItem(QPainter* painter,const QStyleOptionViewItem &option
                                m_pCheckBox);
         }
     }
+
+    // Draw a border if the color cell has focus
```
daschuer, line 91:
```suggestion
    // Draw a border around the lock icon if the bpm cell has focus
```

daschuer (changes requested):
Now it looks like this: 
grafik
Before the extra box around the lock icon was missing. 

Is this really better? 
I had expected that the bar between icon and value is not drawn. 

But I don't have a strong opinion here.

ronso0:
> Now it looks like this: grafik Before the extra box around the lock icon wa

Uff, so this is just an PaleMoon issue. I'll take a look.

Btw the Played item is decorated the same way (indicator not included), and I don't thinkwe should spend time 'fixing' this.

PR #11708, opened 2023-07-03
Library sidebar: expand items to full width to maximize click-responsive area

ronso0:
A small UX helper. Works fine without glitches.

### Issue
Currently, the sidebar header expands to the width of the widest item.
If the sidebar is wider than that, there's some space right next to items that is not responsive to clicks. This is somewhat frustration as it is perceived inconsistent with the state when e.g. Playlist are expanded and the entire 'Tracks' row responds to clicks.

Desired:
* full-width items
* full item text (no elide)
* show scrollbars as needed

Unfortunately, there's no combination of 
`header()->setStretchLastSection(true);`
`header()->setSectionResizeMode(QHeaderView::ResizeMode);`
to achieve that* 
*<sub>at least not when set in the constructor. If set later on, e.g. on `QResizeEvent`, it works as desired, though only with collapsed rows. go figure...</sub>

Probably a Qt issue, or just a use case they didn't know/care about.
(tested with Qt5. I didn't test again with Qt6, but I don't expect a fix tbh, given this is low prio and Qt is very slow with fixing such QWidget bugs)
___
### Fix
The header, and thereby the rows, are (maybe) adjusted
* on every `Resize`, `LayoutRequest`, `FontChange` & `Polish` event
* when expanding/collapsing items and
* when item text is changed (eg. playlist renamed, track count changed)

we then do
* `setStretchLastSection(false)` to show horizontal scrollbars, ie. no ellipsis in item labels
* `setStretchLastSection(true)` if sidebar is wider than the widest item

To avoid unnecessary and redundant adjustments [^1], adjust requests are queued and delayed by 50 ms [^2] so that only the last queued request is processed.
This is implemented along the drag'n'drop timer events with a lightweight QBasicTimer, so I think this is a compact solution.

[^1]: this applies in particular to skin creation/first show and manual resizing, eg. while dragging the sidebar splitter, where many events may be sent in quick succession.
[^2]: 50 ms seem to be a good compromise between fast GUI update and long enough wait time for potential follow-up events to reduce adjust requests.
Update can be seen only when expanding from width with h-scrollbars to one without.

### DIFF
--- a/src/library/library.cpp
+++ b/src/library/library.cpp
@@ -363,6 +363,10 @@ void Library::bindSidebarWidget(WLibrarySidebar* pSidebarWidget) {
             &WLibrarySidebar::deleteItem,
             m_pSidebarModel,
             &SidebarModel::deleteItem);
+    connect(m_pSidebarModel,
+            &SidebarModel::dataChanged,
+            pSidebarWidget,
+            &WLibrarySidebar::queueHeaderAdjustRequest);
 
     connect(pSidebarWidget,
             &WLibrarySidebar::setLibraryFocus,
--- a/src/widget/wlibrarysidebar.cpp
+++ b/src/widget/wlibrarysidebar.cpp
@@ -8,8 +8,12 @@
 #include "moc_wlibrarysidebar.cpp"
 #include "util/defs.h"
 #include "util/dnd.h"
+#include "util/duration.h"
 
 constexpr int expand_time = 250;
+// Delay for header adjust requests. 50 ms seems to be a good compromise between
+// fast GUI update and long enough wait time for potential follow-up events.
+constexpr int resize_header_delay = 50;
 
 WLibrarySidebar::WLibrarySidebar(QWidget* parent)
         : QTreeView(parent),
@@ -28,6 +32,15 @@ WLibrarySidebar::WLibrarySidebar(QWidget* parent)
     header()->setStretchLastSection(false);
     header()->setSectionResizeMode(QHeaderView::ResizeToContents);
     header()->setHorizontalScrollMode(QAbstractItemView::ScrollPerPixel);
+    // Adjust header when an item's expand state has changed
+    connect(this,
+            &QTreeView::expanded,
+            this,
+            &WLibrarySidebar::adjustHeaderStretch);
+    connect(this,
+            &QTreeView::collapsed,
+            this,
+            &WLibrarySidebar::adjustHeaderStretch);
 }
 
 void WLibrarySidebar::contextMenuEvent(QContextMenuEvent *event) {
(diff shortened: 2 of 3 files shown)


ronso0:
Unfortunately clang-tidy isn't run because coverage fails repeatedly when running `HotcueControlTest.SavedLoopMove`

```
362/785 Test #362: HotcueControlTest.SavedLoopMove ............................................................***Exception: Interrupt  0.18 sec
QML debugging is enabled. Only use this in a safe environment.
Note: Google Test filter = HotcueControlTest.SavedLoopMove
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from HotcueControlTest
[ RUN      ] HotcueControlTest.SavedLoopMove
... (list of supported decoders)
critical [Engine] DEBUG ASSERT: "(startSample % CachingReaderChunk::kChannels == 0) && (numSamples % CachingReaderChunk::kChannels == 0) && (numSamples >= 0)" in function virtual CachingReader::ReadResult CachingReader::read(SINT, SINT, bool, CSAMPLE*) at /home/runner/work/mixxx/mixxx/src/engine/cachingreader/cachingreader.cpp:296
```
In all other jobs all test pass :shrug:

ronso0:
> There's just one small, negligible glitch that sometimes the horizontal scrollbar shows up briefly while shrinking even though it's not required

Looks like the viewport already shrunk further while the header was just resized to fit the old width.
This PR is ready as it is, but I'll try to work around this glitch with a timer (resize header n msec after the last resize event).

ronso0:
The first attempt was simple but the resize function was called way too often.
I changed it to use a timer and handle only the last queued event (and stop the timer afterwards).
Works well (except when renaming items but that's not an issue for me).

I consider it helpful but after all it's a hack to compensate for a Qt issue IMO, so I will not fight for getting this merged.
Though, if someone considers this is an okayish way to fix a small usabilty issue, or has better, less-hacky ideas to fix it (e.g. overriding a QTreeView function) let's discuss it.

Check the first commit only, the 2nd one adds lots of debug output to catch events etc.

ronso0:
Btw I wasted some hours just to discover a QTimer quirk: already queued `callOnTimeout` signals are notr removed from the event queue when the timer is stopped or even when it's deleted, that has to be done manually.
Affects the knob wheel timer (show parameter value #11077).

I'll provide a fix soonish.

ronso0:
If polished this and rebased onto main.
Works fine for all sorts of layout or item changes.

daschuer:
I can confirm the issue and that this PR fixes it. After testing it it feels like a bug fix for 2.5. 
What do you think?

daschuer:
The solution looks a bit laborious. Can't we adjust the header size to the sidebar size at least?

ronso0:
Umm, it's been two years this fix is around (and just two people noticed/cared, me + @nPrevail IIRC).
This is just a workaround a Qt flaw and should actually be fixed in QHeaderView.

> The solution looks a bit laborious.

Please elaborate which part you're worried about.

> Can't we adjust the header size to the sidebar size at least?

Believe me, I've tried a lot before I settled with this solution.
Sure we could just set `[QHeaderView::stretchLastSection(true)`](https://doc.qt.io/qt-6/qheaderview.html#stretchLastSection-prop). We'd get full-width items, but ellipsis if they're wider than the view -- and no horizontal scrollbars.

___
Actually, taking another look after this long time, I figured an easier fix (which took me 5 minutes :laughing:)
It's here #15330

ronso0:
Superseded by #15330

PR #11712, opened 2023-07-05
Library sidebar: focus selected item for all keypress events

ronso0:
incl. those handled by QTreeView (Right for example)

fixes a small regression introduced by #11208
if a unselected item was right-clicked (to rename or whatever), then switched to another window and back to Mixxx, navigation keypresses not handled by Mixxx would reset the cursor to Tracks.


daschuer (approved):
Thank you. I can confirm the fix.

PR #11713, opened 2023-07-06
DjInterop: require version

daschuer:
Before, a DIinterop installation was used if the version was matching or no version was found. 
The later happens in case of vcpkg environments where the required DjInteropCinfig.cmake is installed at the wrong location. 

After merging the fallback is used and the library is downloaded as source code. 

I have also noticed that our GitHub Ubuntu builds do also use the fallback, because the package is not part of the debian_buildenv.sh. On Launchpad the package is used.


daschuer:
It looks like cross compiling for the ARM64 macOS hardware form source is broken. 
```
 ld: warning: ignoring file lib/libdjinterop-install/lib/libdjinterop.a, building for macOS-arm64 but attempting to link with file built for macOS-x86_64
Undefined symbols for architecture arm64:
```

daschuer:
Superseded by https://github.com/mixxxdj/mixxx/pull/11714

PR #11714, opened 2023-07-06
Always build djinterop from source and static link

mr-smidge:
Always build libdjinterop from source and statically link, because the library does not yet have a stable ABI.

This situation can be reviewed when libdjinterop reaches version 1.x.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2063,76 +2063,72 @@ endif()
 # Denon Engine Prime library export support (using libdjinterop)
 option(ENGINEPRIME "Support for library export to Denon Engine Prime" ON)
 if(ENGINEPRIME)
+  # libdjinterop does not currently have a stable ABI, so we fetch sources for a specific tag, build here, and link
+  # statically.  This situation should be reviewed once libdjinterop hits version 1.x.
   set(LIBDJINTEROP_VERSION 0.16.1)
-  # Look for an existing installation of libdjinterop and use that if available.
-  # Otherwise, download and build from GitHub.
-  # Note: Version 0.17.0 is not yet compatible
-  find_package(DjInterop ${LIBDJINTEROP_VERSION} EXACT)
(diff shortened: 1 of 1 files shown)


uklotzde:
> does this play nice with build systems that are offline during the build process, such as fedora? And what about platforms that disallow static linking, also such as fedora? hopefully @uklotzde knows.

Should work as it does now. We already copy the source archive manually into the download directory.

uklotzde:
:+1:  Building libdjinterop as an integral part of Mixxx is fine. No need to depend on an external version or a shared library.

Swiftb0y (commented):
LGTM, but I'm not a cmake expert...

mr-smidge:
The macOS arm64 failure was also seen on [another PR](https://github.com/mixxxdj/mixxx/pull/11713) recently... I don't have access to Mac hardware, but will look into this.  Any advice from Mac experts is much appreciated!

daschuer:
That was not the solution :-(

daschuer:
Please try to forward these flags as well: 

CMAKE_OSX_DEPLOYMENT_TARGET
CMAKE_OSX_ARCHITECTURES
CMAKE_SYSTEM_PROCESSOR
CMAKE_SYSTEM_NAME

Can you add the same for the keyfinder library?

mr-smidge:
> Please try to forward these flags as well:
> 
> CMAKE_OSX_DEPLOYMENT_TARGET
> CMAKE_OSX_ARCHITECTURES
> CMAKE_SYSTEM_PROCESSOR
> CMAKE_SYSTEM_NAME
> 
> Can you add the same for the keyfinder library?

This did the trick, thank you @daschuer !

I'm assuming that the failed coverage check and cancelled clang-tidy runs are not blockers for this PR, given they are also broken on the main 2.4 branch right now?

daschuer (approved):
LGTM, Thank you.

PR #11704, opened 2023-07-02
Adjust the crossfading duration when looping at track boundaries

daschuer:
This code got lost somehow when merging https://github.com/ronso0/mixxx/pull/42

The missing bit for real gapless repeat of loop samples:
grafik

PR #11717, opened 2023-07-07
ReadAheadManager: fix loop wraparound reader condition

ronso0:
should fix a regression caused by #11704 / #11532
noticed because HotcueControlTest.SavedLoopMove failed in Coverage check.

Though requested samples shouldn't be negative in the first place, so maybe we overlooked something.

### DIFF
--- a/src/engine/readaheadmanager.cpp
+++ b/src/engine/readaheadmanager.cpp
@@ -7,8 +7,6 @@
 #include "util/math.h"
 #include "util/sample.h"
 
-static constexpr int kNumChannels = 2;
-
 ReadAheadManager::ReadAheadManager()
         : m_pLoopingControl(nullptr),
           m_pRateControl(nullptr),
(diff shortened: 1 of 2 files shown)


ronso0:
@daschuer please take a look.

daschuer:
Thank you. The changes are looking reasonable. 
Is kLogger.critical() crashing the release version? Maybe we should make this a warning. And keep the party going.

src/engine/readaheadmanager.cpp
```diff
@@ -160,7 +160,7 @@ SINT ReadAheadManager::getNextSamples(double dRate, CSAMPLE* pOutput,
             }
         }
 
-        if (crossFadeSamples) {
+        if (crossFadeSamples >= 0) {
```
ronso0, line 163: still trying to figure when `crossFadeSamples < 0` occurs.
if I get it right this can happen only if the overshoot (fractional frames?) corrects the curr. play pos AND we're looping over the track start, so that `loop_read_position > crossFadeStart`?
so we actually thow away the overshoot, no?

ronso0:
> Is kLogger.critical() crashing the release version? Maybe we should make this a warning. And keep the party going.

`qFatal()` does IIUC, and `qCritical()` does, too "if the environment variable QT_FATAL_CRITICALS is not empty"
https://doc.qt.io/qt-6/qtlogging.html#qCritical
Don't know why this would be enabled. Do you? Sure we can make it a warning.

daschuer:
> qFatal() does IIUC, and qCritical() does, too "if the environment variable QT_FATAL_CRITICALS is not empty"

Oh I was confused because our assertions are Critical and elevated Fatal for a crash. 
So all fine.

daschuer (changes requested):
I want still investigate why the value is negative. But that can be done after merge.

ronso0:
so this one's ready

daschuer (approved):
It turns out that the value has to be negative, because the loop cue is shifted into the pre-roll.
So this is OK.

Unfortunately this issue is also a 2.3 issue. I will add only the fix to the 2.3 branch without the refactoring. 

LGTM

PR #11719, opened 2023-07-07
Mixtrack pro fx update

andybeetz:
Adding a switch to allow the FX encoder to control the meta knobs of Effect 1 on FX 1 and FX 2 (when shifted).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -143,7 +143,7 @@ repos:
         language: python
         additional_dependencies:
           - beautifulsoup4==4.11.1
-          - lxml==4.9.1
+          - lxml==4.9.2
           - Markdown==3.4.1
         types: [text]
         files: ^(CHANGELOG\.md|res/linux/org\.mixxx\.Mixxx\.metainfo.xml)$
--- a/res/controllers/Numark-Mixtrack-Pro-FX-scripts.js
+++ b/res/controllers/Numark-Mixtrack-Pro-FX-scripts.js
@@ -11,7 +11,7 @@ MixtrackProFX.pitchRanges = [0.08, 0.16, 1];
 MixtrackProFX.waveformsSynced = true;
 
 // FX encoder controls meta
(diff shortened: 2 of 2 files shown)


JoergAtGithub:
Why do you use `meta `and not `super1`?

andybeetz:
I don't know what super1 is or does.

JoergAtGithub:
grafik
Blue is the super knob and yellow the meta knob. The super knob controls all meta knobs together and is also visible in colapsed view of the FX unit.
Note: The screenshot is from Mixxx 2.4, it might be different on Mixxx 2.3.

ronso0:
> The knob doesn't exist in my installed version

GUi Super knobs are hidden by default. You can enable them in the Skin Settings (Ctrl+1).
https://manual.mixxx.org/2.3/en/chapters/appendix/mixxx_controls.html#control-[EffectRack1_EffectUnitN]-super1

andybeetz:
> > The knob doesn't exist in my installed version
> 
> GUi Super knobs are hidden by default. You can enable them in the Skin Settings (Ctrl+1). [https://manual.mixxx.org/2.3/en/chapters/appendix/mixxx_controls.html#control-[EffectRack1_EffectUnitN]-super1](https://manual.mixxx.org/2.3/en/chapters/appendix/mixxx_controls.html#control-%5BEffectRack1_EffectUnitN%5D-super1)

I will pick this as my reason for using meta

ronso0:
Hmm, this is no good reason IMO.
@JoergAtGithub is right, Super is a more versatile control as it allows users to work with all 3 effect slots. If just one effect is desired they can leave the unused effects disabled (default state).
Even though the Super knob maybe hidden (1) the mapping documentation can link to https://manual.mixxx.org/2.3/en/chapters/appendix/mixxx_controls.html#control-%5BEffectRack1_EffectUnitN%5D-super1 and (2) users will see the Meta knobs move anyway.

PR #11721, opened 2023-07-07
Use std:move instead of copy by value

JoergAtGithub:
This fixes most of the performance issues reported by Coverity Scan of type COPY_INSTEAD_OF_MOVE.

On some places I also changed the name of the moved variable, to comply with our naming conventions.


src/controllers/controlpickermenu.cpp
```diff
@@ -881,7 +881,7 @@ ControlPickerMenu::ControlPickerMenu(QWidget* pParent)
 
         QString assignMenuTitle = tr("Effect Unit Assignment");
         QString assignString = tr("Assign ");
-        QMenu* effectUnitGroups = addSubmenu(assignMenuTitle,
+        QMenu* effectUnitGroups = addSubmenu(std::move(assignMenuTitle),
```
daschuer, line 884: This QString is passed finally to `addMenu(const QString &text)`. 
So this warning should be solved by changing addSubmenu to take `const QString&` as well-

src/broadcast/broadcastmanager.cpp
```diff
@@ -128,12 +128,12 @@ void BroadcastManager::slotControlEnabled(double v) {
...
-void BroadcastManager::slotProfileAdded(BroadcastProfilePtr profile) {
-    addConnection(profile);
+void BroadcastManager::slotProfileAdded(BroadcastProfilePtr pProfile) {
+    addConnection(std::move(pProfile));
 }
 
-void BroadcastManager::slotProfileRemoved(BroadcastProfilePtr profile) {
-    removeConnection(profile);
+void BroadcastManager::slotProfileRemoved(BroadcastProfilePtr pProfile) {
+    removeConnection(std::move(pProfile));
```
daschuer, line 136: Since removeConnection only borrows the pointer, it should become a const & or a raw pointer parameter. 
The later requires a bit more refactoring.

src/controllers/dlgcontrollerlearning.cpp
```diff
@@ -484,7 +484,7 @@ void DlgControllerLearning::loadControl(const ConfigKey& key,
 void DlgControllerLearning::controlPicked(const ConfigKey& control) {
     QString title = m_pControlPickerMenu->controlTitleForConfigKey(control);
     QString description = m_pControlPickerMenu->descriptionForConfigKey(control);
-    loadControl(control, title, description);
+    loadControl(control, title, std::move(description));
```
daschuer, line 487: DlgControllerLearning::loadControl() can make use of a refactoring to use const & 
std::move dos not really help her, because the string is not kept in a member variable.

src/controllers/dlgprefcontroller.cpp
```diff
@@ -57,7 +57,7 @@ DlgPrefController::DlgPrefController(
     initTableView(m_ui.m_pOutputMappingTableView);
 
     std::shared_ptr<LegacyControllerMapping> pMapping = m_pController->cloneMapping();
-    slotShowMapping(pMapping);
+    slotShowMapping(std::move(pMapping));
```
daschuer, line 60: You may consider to use m_pMapping = std::move(pMapping); in the very beginning of slotShowMapping() and use this afterwards.

src/controllers/midi/midicontroller.cpp
```diff
@@ -279,7 +279,7 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
                 status,
                 mapping.control.group,
         };
-        if (!pEngine->executeFunction(function, args)) {
+        if (!pEngine->executeFunction(std::move(function), args)) {
```
daschuer, line 282: This is conflicting with my change in https://github.com/mixxxdj/mixxx/pull/11701 making it a pointer to fix clang-tidy Qt6 warnings.

src/controllers/midi/portmidienumerator.cpp
```diff
@@ -201,7 +201,7 @@ QList<Controller*> PortMidiEnumerator::queryDevices() {
         qDebug() << " Found output device"
                  << "#" << i << pDeviceInfo->name;
         QString deviceName = pDeviceInfo->name;
-        unassignedOutputDevices[i] = deviceName;
+        unassignedOutputDevices[i] = std::move(deviceName);
```
daschuer, line 204: Is this a false positive? Operator[] takes a const &. 
Maybe call insert() as a workaround?

src/controllers/scripting/controllerscriptmoduleengine.cpp
```diff
@@ -32,7 +32,7 @@ bool ControllerScriptModuleEngine::initialize() {
     }
 
     QJSValue initFunction = mod.property("init");
-    if (!executeFunction(initFunction)) {
+    if (!executeFunction(std::move(initFunction))) {
```
daschuer, line 35: Pass by pointer like https://github.com/mixxxdj/mixxx/pull/11701

src/effects/effectslot.cpp
```diff
@@ -249,12 +249,12 @@ void EffectSlot::loadEffectFromPreset(const EffectPresetPointer pPreset) {
...
     EffectManifestPointer pManifest = m_pBackendManager->getManifest(
             pPreset->id(), pPreset->backendType());
-    loadEffectInner(pManifest, pPreset, true);
+    loadEffectInner(std::move(pManifest), pPreset, true);
 }
 
 void EffectSlot::loadEffectWithDefaults(const EffectManifestPointer pManifest) {
     EffectPresetPointer pPreset = m_pPresetManager->getDefaultPreset(pManifest);
-    loadEffectInner(pManifest, pPreset, false);
+    loadEffectInner(pManifest, std::move(pPreset), false);
```
daschuer, line 257: above you move pManifest ... ? 
I think pPreset should be passed as raw pointer of const reference, because it is only borrowed.

src/effects/presets/effectpresetmanager.cpp
```diff
@@ -62,8 +62,8 @@ void EffectPresetManager::loadDefaultEffectPresets() {
 }
 
 void EffectPresetManager::saveDefaultForEffect(EffectSlotPointer pEffectSlot) {
-    EffectPresetPointer pPreset(new EffectPreset(pEffectSlot));
-    saveDefaultForEffect(pPreset);
+    EffectPresetPointer pPreset(new EffectPreset(std::move(pEffectSlot)));
```
daschuer, line 65: pEffectSlot is only borrowed use const & or raw pointer.

src/effects/presets/effectpresetmanager.cpp
```diff
@@ -62,8 +62,8 @@ void EffectPresetManager::loadDefaultEffectPresets() {
...
+    saveDefaultForEffect(std::move(pPreset));
```
daschuer, line 66: the same here.

daschuer (changes requested):
I have reviewed ~1/10 of this PR. 
Most of this these issues are that the original pass by value is sub-optimal anyway. 
the std:move() is a band aid but does not fix the original issue. 
Can you rework this PR accordingly? 

Check it the copied object is copied into a member again? 
In this case the compiler is able to combine all copies to just one. 
If that is not the case, please pass the value as const reference or as plain pointer. 
This is even faster than moving to a value parameter. 
 
It would be also OK to create a quick takaway PR tahat only used std::move when copying to a member like `m_value = std::move(value);`

PR #11723, opened 2023-07-08
Feature: added a flag to prevent the risk of an endless loop.

### LINKED ISSUE #11525: Inhibt CO connection trigger while executing the callback of this connection
### Feature Description

To prevent the risk of an endless loop by triggering a CO connection from inside the connections callback. Could we add a flag to the connection class, which inhibits triggering a connection during execution of it's callback: https://github.com/mixxxdj/mixxx/blob/f36cd4f29086aa5370ece5a10d198f690a3619da/src/control/controlobjectscript.cpp#L130-L139

https://github.com/mixxxdj/mixxx/blob/f36cd4f29086aa5370ece5a10d198f690a3619da/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp#L287-L300

prikshitsingh24:
To prevent the risk of an endless loop by triggering a CO connection from inside the connections callback added a flag to the connection class, which inhibits triggering a connection during execution of it's callback.

### DIFF
--- a/src/control/controlobjectscript.cpp
+++ b/src/control/controlobjectscript.cpp
@@ -149,3 +149,11 @@ void ControlObjectScript::slotValueChanged(double value, QObject*) {
         }
     }
 }
+
+void ControlObjectScript::trigger(double value, QObject*) {
+    if (this->m_callbackExecuting) {
+        std::cerr << "Warning: trigger() called inside the callback. ";
+        return;
+    }
+    emit trigger(value, object);
+}
--- a/src/control/controlobjectscript.cpp
+++ b/src/control/controlobjectscript.cpp
@@ -132,13 +132,12 @@ void ControlObjectScript::slotValueChanged(double value, QObject*) {
     // This allows a script to disconnect a callback from inside the
      const QVector<ScriptConnection> connections = m_scriptConnections;
     // Flag to inhibit triggering connections during callback execution
-     bool callbackExecuting = false;
 
     for (auto&& conn: connections) {
-        if (!callbackExecuting) {
-            callbackExecuting = true;
+        if (!this->m_callbackExecuting) {
+            this->m_callbackExecuting = true;
             conn.executeCallback(value);
-            callbackExecuting = false;
+            this->m_callbackExecuting = false;
         }
     }
 }
--- a/src/control/controlobjectscript.cpp
+++ b/src/control/controlobjectscript.cpp
@@ -130,10 +130,15 @@ void ControlObjectScript::disconnectAllConnectionsToFunction(const QJSValue& fun
 void ControlObjectScript::slotValueChanged(double value, QObject*) {
     // Make a local copy of m_connectedScriptFunctions first.
     // This allows a script to disconnect a callback from inside the
-    // the callback. Otherwise the this may crash since the disconnect call
-    // happens during conn.function.call() in the middle of the loop below.
-    const QVector<ScriptConnection> connections = m_scriptConnections;
+     const QVector<ScriptConnection> connections = m_scriptConnections;
+    // Flag to inhibit triggering connections during callback execution
+    bool callbackExecuting = false;
+
     for (auto&& conn: connections) {
-        conn.executeCallback(value);
+        if (!callbackExecuting) {
+            callbackExecuting = true;
+            conn.executeCallback(value);
+            callbackExecuting = false;
+        }
     }
 }
--- a/src/control/controlobjectscript.cpp
+++ b/src/control/controlobjectscript.cpp
@@ -130,8 +130,9 @@ void ControlObjectScript::disconnectAllConnectionsToFunction(const QJSValue& fun
 void ControlObjectScript::slotValueChanged(double value, QObject*) {
     // Make a local copy of m_connectedScriptFunctions first.
     // This allows a script to disconnect a callback from inside the
-     const QVector<ScriptConnection> connections = m_scriptConnections;
-    // Flag to inhibit triggering connections during callback execution
+    // the callback. Otherwise the this may crash since the disconnect call
+    // happens during conn.function.call() in the middle of the loop below.
+    const QVector<ScriptConnection> connections = m_scriptConnections;
 
     for (auto&& conn: connections) {
           conn.executeCallback(value);
--- a/src/control/controlobjectscript.cpp
+++ b/src/control/controlobjectscript.cpp
@@ -146,14 +146,8 @@ void ControlObjectScript::slotValueChanged(double value, QObject*) {
             this->m_callbackExecuting = true;
             conn.executeCallback(value);
             this->m_callbackExecuting = false;
+        } else {
+            qCWarning(m_logger) << "Warning: trigger() called inside the callback.";
         }
     }
-}
-
-void ControlObjectScript::trigger(double value, QObject*) {
-    if (this->m_callbackExecuting) {
-        std::cerr << "Warning: trigger() called inside the callback. ";
-        return;
-    }
-    emit trigger(value, object);
 }
--- a/src/control/controlobjectscript.cpp
+++ b/src/control/controlobjectscript.cpp
@@ -132,7 +132,7 @@ void ControlObjectScript::slotValueChanged(double value, QObject*) {
     // This allows a script to disconnect a callback from inside the
      const QVector<ScriptConnection> connections = m_scriptConnections;
     // Flag to inhibit triggering connections during callback execution
-    bool callbackExecuting = false;
+     bool callbackExecuting = false;
 
     for (auto&& conn: connections) {
         if (!callbackExecuting) {
--- a/src/control/controlobjectscript.cpp
+++ b/src/control/controlobjectscript.cpp
@@ -151,21 +151,17 @@ void ControlObjectScript::slotValueChanged(double value, QObject*) {
      const QVector<ScriptConnection> connections = m_scriptConnections;
     // Flag to inhibit triggering connections during callback execution
 
-
-    
-
     for (auto&& conn: connections) {
-         if (!m_callbackExecuting && !m_preventRecursiveCalls) {
-             m_callbackExecuting = true;
-             conn.executeCallback(value);
-             m_callbackExecuting = false;
-         } else if (m_preventRecursiveCalls) {
-             // Handle preventing recursive calls without a warning
-             // Add any necessary logic here
-         } else {
-             qCCritical(m_logger) << "Critical Error: Triggering a connection inside the callback is not allowed.";
-         }  
+        if (!m_callbackExecuting && !m_preventRecursiveCalls) {
+            m_callbackExecuting = true;
+            conn.executeCallback(value);
+            m_callbackExecuting = false;
+        } else if (m_preventRecursiveCalls) {
+            // Handle preventing recursive calls without a warning
+            // Add any necessary logic here
+        } else {
+            qCCritical(m_logger) << "Critical Error: Triggering a connection "
+                                    "inside the callback is not allowed.";
+        }
     }
-
-    
 }
--- a/src/control/controlobjectscript.cpp
+++ b/src/control/controlobjectscript.cpp
@@ -7,9 +7,7 @@ ControlObjectScript::ControlObjectScript(
         : ControlProxy(key, pParent, ControlFlag::AllowMissingOrInvalid),
           m_logger(logger),
           m_proxy(key, logger, this),
-          m_skipSuperseded(false),
-          m_callbackExecuting(false),
-          m_preventRecursiveCalls(false) {
+          m_skipSuperseded(false) {
 }
 
 bool ControlObjectScript::addScriptConnection(const ScriptConnection& conn) {
@@ -129,22 +127,6 @@ void ControlObjectScript::disconnectAllConnectionsToFunction(const QJSValue& fun
     }
 }
 
-bool ControlObjectScript::isCallbackExecuting() const {
-    return m_callbackExecuting;
-}
-
-void ControlObjectScript::setCallbackExecuting(bool executing) {
-    m_callbackExecuting = executing;
-}
-
-bool ControlObjectScript::isPreventRecursiveCalls() const {
-    return m_preventRecursiveCalls;
-}
-
-void ControlObjectScript::setPreventRecursiveCalls(bool recursive) {
-    m_preventRecursiveCalls = recursive;
-}
-
 void ControlObjectScript::slotValueChanged(double value, QObject*) {
     // Make a local copy of m_connectedScriptFunctions first.
     // This allows a script to disconnect a callback from inside the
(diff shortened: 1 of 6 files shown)


src/control/controlobjectscript.cpp
```diff
@@ -130,10 +130,15 @@ void ControlObjectScript::disconnectAllConnectionsToFunction(const QJSValue& fun
 void ControlObjectScript::slotValueChanged(double value, QObject*) {
     // Make a local copy of m_connectedScriptFunctions first.
     // This allows a script to disconnect a callback from inside the
-    // the callback. Otherwise the this may crash since the disconnect call
-    // happens during conn.function.call() in the middle of the loop below.
-    const QVector<ScriptConnection> connections = m_scriptConnections;
+     const QVector<ScriptConnection> connections = m_scriptConnections;
+    // Flag to inhibit triggering connections during callback execution
+     bool callbackExecuting = false;
```
JoergAtGithub, line 135: callbackExecuting is a local variable, which is notvisible outside of this function. Therefore it can't work.

JoergAtGithub (changes requested):

JoergAtGithub:
Welcome at Mixxx!

src/control/controlobjectscript.cpp
PrikshitSingh24: what if I specify this variable as a data member of class ControlObjectScript

JoergAtGithub: This could work! 
In general, you need to create a test case, which conntains a trigger() call of itself in it's callback. Without, you can't test if your code does, what it should.

JoergAtGithub: You still have two independent variables, it can only work if you write/read into the same membe variable

PrikshitSingh24: can i make the ControllerScriptInterfaceLegacy as a friend class of ControlObjectScript then i will be able to create a single private data member inside ControlObjectScript.

PrikshitSingh24: or i could make the m_callbackExecuting a public data member

src/control/controlobjectscript.h
```diff
@@ -44,5 +47,6 @@ class ControlObjectScript : public ControlProxy {
     QVector<ScriptConnection> m_scriptConnections;
     const RuntimeLoggingCategory m_logger;
     CompressingProxy m_proxy;
-    bool m_skipSuperseded; // This flag is combined for all connections of this Control Object
+    bool m_skipSuperseded;
```
JoergAtGithub, line 50: Why did you delete this comment?

PrikshitSingh24: i am sorry might have been deleted by mistake

JoergAtGithub:
Could you add a warning that is printed when trigger() is called in the callback.

JoergAtGithub:
To fix the pre-commit fails automatically, please set up automatic code checking: https://github.com/mixxxdj/mixxx/wiki/Using-Git#set-up-automatic-code-checking

src/control/controlobjectscript.cpp
```diff
@@ -127,13 +127,27 @@ void ControlObjectScript::disconnectAllConnectionsToFunction(const QJSValue& fun
...
     // This allows a script to disconnect a callback from inside the
-    // the callback. Otherwise the this may crash since the disconnect call
-    // happens during conn.function.call() in the middle of the loop below.
-    const QVector<ScriptConnection> connections = m_scriptConnections;
+     const QVector<ScriptConnection> connections = m_scriptConnections;
+    // Flag to inhibit triggering connections during callback execution
+
     for (auto&& conn: connections) {
-        conn.executeCallback(value);
+        if (!this->m_callbackExecuting) {
```
daschuer, line 145:
```suggestion
        if (!m_callbackExecuting) {
```
this-> prefix is redundant

src/control/controlobjectscript.cpp
```diff
@@ -127,13 +127,27 @@ void ControlObjectScript::disconnectAllConnectionsToFunction(const QJSValue& fun
...
+            this->m_callbackExecuting = true;
+            conn.executeCallback(value);
+            this->m_callbackExecuting = false;
+        } else {
+            qCWarning(m_logger) << "Warning: trigger() called inside the callback.";
```
daschuer, line 150: This is a mapping script error and shall not happen in a release version of the script. 
Can we make this a critical error and rephrase the string accordingly?

src/control/controlobjectscript.h
```diff
@@ -19,6 +19,9 @@ class ControlObjectScript : public ControlProxy {
 
     bool removeScriptConnection(const ScriptConnection& conn);
 
+    bool isCallbackExecuting() const;
+    void setCallbackExecuting(bool executing);
```
daschuer, line 23: IMHO this function should be removed.
Calling it can only mess up the ne feature.

src/control/controlobjectscript.h
```diff
@@ -45,4 +48,5 @@ class ControlObjectScript : public ControlProxy {
     const RuntimeLoggingCategory m_logger;
     CompressingProxy m_proxy;
     bool m_skipSuperseded; // This flag is combined for all connections of this Control Object
+    bool m_callbackExecuting = false;
```
daschuer, line 51: Please initialize the variable in the constructor.
Initialization in the header will create a hidden static inline constructor which we want to avoid.

daschuer (changes requested):
Thank you. I have added some comments

src/control/controlobjectscript.h
PrikshitSingh24: Then how will the coScript object access the variable in controllerscriptinterfacelegacy.cpp . Should I make the variable public to solve that?

src/control/controlobjectscript.cpp
PrikshitSingh24: Could I use std::cerr for which I have to include iostream??

JoergAtGithub: No, just qCCritical instead of qCWarning.

src/control/controlobjectscript.h
daschuer: The getter can remain, I am any concerned about the setter. Do we have a use case for it?

PrikshitSingh24: Using it in controllerscriptinterfacelegacy.cpp file 
if (!coScript->isCallbackExecuting()) {
            coScript->setCallbackExecuting(true);
            connection.executeCallback(coScript->get());
            coScript->setCallbackExecuting(false);
    }
}

JoergAtGithub: From my point of view, this is correct implemented and used. I don't get this point.

src/control/controlobjectscript.cpp
JoergAtGithub: The error should be reported, in the trigger function - not here.

src/control/controlobjectscript.h
PrikshitSingh24: so, should I keep the functions

JoergAtGithub: First ensure, that it works, than we see which functions are needed therefore.

PrikshitSingh24: I have implemented all the necessary changes that were informed now how to check that it works?

daschuer: It looks like your code example reveals a conceptual issue. Currently the m_callbackExecuting is a private member of ControlObjectScript guarding slotValueChanged(). It will detect if you change the value again from the callback and start an endless recursive loop, which is always a serious script issue. 

Your code issue is different. It is a cooperative api (the script programmer can use it if he knows it) to assert that the value triggering the callback again does not happen. This is not necessary a programming issue. It becomes one if the value is triggered again, but this already in place even if the user does not use the coScript->setCallbackExecuting(true);

I can Imagine that there is a demand to intentionally block slotValueChanged() from executing the callback. This is another desired feature which must not issue any warning. In Qt we have already QObject::blockSignals(bool block) for this. Not sure if we can use it in this context but if here is a similar demand from the script we should expose/re-implement that in addition.

src/control/controlobjectscript.cpp
```diff
@@ -127,13 +127,27 @@ void ControlObjectScript::disconnectAllConnectionsToFunction(const QJSValue& fun
...
-    // the callback. Otherwise the this may crash since the disconnect call
-    // happens during conn.function.call() in the middle of the loop below.
-    const QVector<ScriptConnection> connections = m_scriptConnections;
+     const QVector<ScriptConnection> connections = m_scriptConnections;
+    // Flag to inhibit triggering connections during callback execution
+
     for (auto&& conn: connections) {
-        conn.executeCallback(value);
+        if (!this->m_callbackExecuting) {
+            this->m_callbackExecuting = true;
```
daschuer, line 146: Can we move this outside the loop, in a way that the check is only performed once if there are n callbacks connected?

src/control/controlobjectscript.cpp
daschuer: Do you mean in triggerScriptConnection()? 

I am afraid we want two things with a single flag. Currently the flag is a ControlObjectScript member guarding recursive valueChange calls. If you want to guard recursive connection calls, we need to move the flag to ControllerScriptInterfaceLegacy.

daschuer (changes requested):

src/control/controlobjectscript.h
PrikshitSingh24: So should I Remove the m_callbackExecuting flag from ControlObjectScript. It will no longer be responsible for guarding against recursive callback execution.
Introduce a new flag, let's call it m_blockCallbackExecution, in ControllerScriptInterfaceLegacy to handle cooperative blocking of callback execution.

daschuer: It depends. What is the goal we like to archive? 

It looks like guarding individual connections sounds reasonable. This should implicit guard against recursive valueChange calls. 

I see clearly the use-case of the automatic stack overflow protraction that should be mandatory. 
This should be at least read only from public or no interface at all. 

Do we have a use case to intentionally prevents recursive calls without a warning? 
For this we need a second flag. 
It should be easy to maintain such a flag in the script and just not call trigger() or the co update again. This should avoid reaching the internal last resort flag. 

Is the a use-case to bypass the internal flag? We will learn that after merging this PR.

daschuer:
Did you consider to move the logic to ScriptConnection::executeCallback()?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -298,7 +298,8 @@ void ControllerScriptInterfaceLegacy::triggerScriptConnection(
         return;
     }
 
-    connection.executeCallback(coScript->get());
+        connection.executeCallback(coScript->get());
```
daschuer, line 301: Unrelated change. 
There are a few other changes like this.
Maybe it is treasonable to remove them and then squash all commits to one or two.

src/controllers/scripting/legacy/scriptconnection.cpp
```diff
@@ -5,19 +5,36 @@
...
-    QJSValue func = callback; // copy function because QJSValue::call is not const
-    QJSValue result = func.call(args);
-    if (result.isError()) {
-        if (controllerEngine != nullptr) {
-            controllerEngine->showScriptExceptionDialog(result);
+
+    if (!m_callbackExecuting && !m_preventRecursiveCalls) {
+        m_callbackExecuting = true;
+        m_preventRecursiveCalls = true;
+
```
daschuer, line 12: Setting that flag here leads to never having the critical warning. I think it has to be only checked to inhibit the warning below.

daschuer (changes requested):

src/control/controlobjectscript.cpp
```diff
@@ -130,10 +130,10 @@ void ControlObjectScript::disconnectAllConnectionsToFunction(const QJSValue& fun
 void ControlObjectScript::slotValueChanged(double value, QObject*) {
     // Make a local copy of m_connectedScriptFunctions first.
     // This allows a script to disconnect a callback from inside the
-    // the callback. Otherwise the this may crash since the disconnect call
-    // happens during conn.function.call() in the middle of the loop below.
-    const QVector<ScriptConnection> connections = m_scriptConnections;
```
daschuer, line 135: Was this comment removed accidentally?

src/controllers/scripting/legacy/scriptconnection.cpp
```diff
@@ -5,19 +5,32 @@
...
-            value,
-            key.group,
-            key.item,
-    };
-    QJSValue func = callback; // copy function because QJSValue::call is not const
-    QJSValue result = func.call(args);
-    if (result.isError()) {
-        if (controllerEngine != nullptr) {
-            controllerEngine->showScriptExceptionDialog(result);
+    if (!m_preventRecursiveCalls) {
```
daschuer, line 8: Now you can prevent the callback entirely. 
Was this intended? Does it need a public interface?

src/controllers/scripting/legacy/scriptconnection.cpp
```diff
@@ -5,19 +5,32 @@
...
+        if (!m_callbackExecuting) {
+            m_callbackExecuting = true;
```
daschuer, line 10: m_callbackExecuting initialization is missing in the constructor of ScriptConnection.

daschuer (changes requested):

src/controllers/scripting/legacy/scriptconnection.cpp
PrikshitSingh24: No, the code is not preventing the callback entirely. The execution of the callback is still happening when the conditions are met.
The code checks if m_preventRecursiveCalls is false and m_callbackExecuting is false before executing the callback. If these conditions are met, the callback function is executed. However, if m_callbackExecuting is true, it means that a connection is being triggered inside the callback, and a critical error message is logged.

src/controllers/scripting/legacy/scriptconnection.cpp
PrikshitSingh24: I have implemented both m_preventRecursiveCalls and m_callbackExecuting now , I will send you a clean PR which will be understandable.

daschuer: Thank you.

src/controllers/scripting/legacy/scriptconnection.h
```diff
@@ -33,4 +33,6 @@ class ScriptConnection {
     inline bool operator!=(const ScriptConnection& other) const {
         return !(*this == other);
     }
+    private:
+    const RuntimeLoggingCategory m_logger;
```
PrikshitSingh24, line 37: How to access and use ``qCCritical(m_logger) << "Critical Error: Triggering a connection "
                                    "inside the callback is not allowed.";`` inside the ScriptConnection.

ronso0:
What's the stats here @daschuer ?
Is there a quick fix in sight? (i.e. could you take over and finish this?)

acolombier:
Marking this as draft for now - feel free to mark as ready if you get a chance to complete the requested changes, or close in case you don't have interest in finishing this anymore

ronso0:
One year without response...
Hope it's okay to close this now @daschuer @JoergAtGithub 
____
I'll close this abandoned PR now to clean up the list of open PRs.
If the author or maintainers think this should remain open please reopen.
If anyone thinks this is worth salvaging and is motivated to pick it up, resolve the conflicts and move it forward, feel free to open a new PR.

Issue #11725, opened 2023-07-08
Ressource leak in autodjfeature.cpp#L307


JoergAtGithub:
### Bug Description

https://github.com/mixxxdj/mixxx/blob/0f59bf86a1c33eb9acd101917a06864e535f6eb4/src/library/autodj/autodjfeature.cpp#L307-L308

Results in a ressource leak, because https://doc.qt.io/qt-5/qwidget.html#addAction is documented as "The ownership of action is not transferred to this QWidget"

### Version

2.4

### OS

Windows11

daschuer:
This is no resource leak, because the action is parented.
We have here a double ownership issue that should ne solved. A parented_ptr would indicate that.

PR #11726, opened 2023-07-08
Fixed unnecessary transfer of the ownership before release which returns the pointer itself

JoergAtGithub:
These were reported as resoource leak by Coverity Scan, not real issues, but also an unnecessary commands. So this is the cleanest way to silence the warning.
While going through these warnings I noticed also #11725 which looks similar, but is a real ressource leak.

### DIFF
--- a/src/track/taglib/trackmetadata_id3v2.cpp
+++ b/src/track/taglib/trackmetadata_id3v2.cpp
@@ -382,10 +382,10 @@ void writeTextIdentificationFrame(
         auto pFrame =
                 std::make_unique<TagLib::ID3v2::TextIdentificationFrame>(id, stringType);
         pFrame->setText(toTString(text));
-        pTag->addFrame(pFrame.get());
-        // Now that the plain pointer in pFrame is owned and managed by
-        // pTag we need to release the ownership to avoid double deletion!
-        pFrame.release();
+
+        // Release returns a pointer to the managed object and releases the ownership,
+        // the plain pointer in pFrame is owned and managed by pTag
+        pTag->addFrame(pFrame.release());
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/track/taglib/trackmetadata_id3v2.cpp
```diff
@@ -382,10 +382,10 @@ void writeTextIdentificationFrame(
         auto pFrame =
                 std::make_unique<TagLib::ID3v2::TextIdentificationFrame>(id, stringType);
         pFrame->setText(toTString(text));
-        pTag->addFrame(pFrame.get());
-        // Now that the plain pointer in pFrame is owned and managed by
-        // pTag we need to release the ownership to avoid double deletion!
-        pFrame.release();
+
+        // Release returns a pointer to the managed object and releases the ownership,
```
uklotzde, line 386: The comment is obsolete after dropping the 2-step get-release pattern. Repeating the std definition  of release() is pointless. Also applies to other locations in the code.

The 2-step pattern is required for exception safe C++ programming. But since any exception thrown by addFrame() is expected to be fatal a resource leak doesn't matter then.

uklotzde (changes requested):

src/track/taglib/trackmetadata_id3v2.cpp
uklotzde: This code has been written with purpose and the comment explained why. If you decide to take a simpler approach then delete the obsolete comment. I don't care, does not really matter. But please change it in a consistent way.

uklotzde: The resource leak reported by Coverity is obviously a false positive.

src/track/taglib/trackmetadata_id3v2.cpp
```diff
@@ -382,10 +382,10 @@ void writeTextIdentificationFrame(
...
+        // the plain pointer in pFrame is owned and managed by pTag
```
daschuer, line 387: Something like this should be sufficient 
```suggestion
        // pTag takes the ownership of pFrame 
```

daschuer (approved):
LGTM, Thank you

PR #11720, opened 2023-07-07
Update to libdjinterop 0.19

mr-smidge:
This PR updates Mixxx to use version 0.19 of libdjinterop, which supports Denon firmware up to version 3.0.0.

There are no functional changes as part of this PR - the changes are limited to those necessary to adhere to the evolving public API of libdjinterop whilst it is in its early 0.x state.

PR #11729, opened 2023-07-09
DjInterop: Fix Beta PPA and use VCPG static build

daschuer:
In https://github.com/mixxxdj/mixxx/pull/11714 we have decided to download DjInterop always and do a static build. 

Accordingly I have prepared a vcpkg environment that has a pre-build STATIC version of DjInterop that will save us some build time. Since the cmake.config issue was fixed upstream it is now found reliable (see https://github.com/mixxxdj/vcpkg/pull/74)
So we can here re-enable the package_find() stage. 

The PPA build is not able to reach the GitHub tar ball. @mr-smidge has uploaded it to Launchpad as well so that we can use the Launchpad mirror that should be reach-able. 
This should fix our currently broken beta PPA builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2066,69 +2066,81 @@ if(ENGINEPRIME)
   # libdjinterop does not currently have a stable ABI, so we fetch sources for a specific tag, build here, and link
   # statically.  This situation should be reviewed once libdjinterop hits version 1.x.
   set(LIBDJINTEROP_VERSION 0.16.1)
-
-  # On MacOS, Mixxx does not use system SQLite, so we will use libdjinterop's
-  # embedded SQLite in such a case.
-  if (APPLE AND NOT SQLite3_IS_STATIC)
-    message(STATUS "Building libdjinterop sources (with embedded SQLite) fetched from GitHub")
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
Sounds good, but to test this, we need first to merge #11720 and than use the new VCPKG buildenv. Because the fix from https://github.com/microsoft/vcpkg/pull/32312 is only in the 0.19.1port of the updated buildenv.

daschuer:
This is actually independent form https://github.com/mixxxdj/mixxx/pull/11720 because it will no longer pick the vcpkg version without finding the version. 
The windows build download the source. See CI 
```
2023-07-09T17:15:16.9635647Z -- Could NOT find DjInterop (missing: DjInterop_DIR)
2023-07-09T17:15:17.0005471Z -- Could NOT find DjInterop (missing: DjInterop_VERSION) (Required is exact version "0.16.1")
2023-07-09T17:15:17.0006195Z -- Building libdjinterop sources (with system SQLite) fetched from GitHub
```
Once we have moved to the new VCPKG the first check will succeed.

JoergAtGithub:
I tested this on Windows with 0.16.1, 0.20.1 and 0.19.1 and it worked as intended. Thank you!

Issue #11732, opened 2023-07-09
waveform randomly disappears if shifting a loop backwards in pre-roll


ronso0:
https://github.com/mixxxdj/mixxx/assets/5934199/09811685-9095-483f-bf1a-fac1c3005c2e

wavefrom is RGB (GLSL)

daschuer:
I cannot confirm it on Ubuntu Focal using the latest 2.4

ronso0:
Weird, can't reproduce it anymore.
I'll reopen this if I notice it again.

PR #11735, opened 2023-07-10
Looping: fix asserts for loop move

ronso0:
`adjustedPosition <= newLoopEndPosition` is wrong since `newLoopEndPosition` is considered outside the loop and jumping there would disable the loop.

I hit those assertions while trying to reproduce #11732, while moving a loop (set exactly at track start 0.00) back and forth crossing track start sample (where rounding issues occur.. ?)


daschuer (approved):
LGTM, Thank you.

Issue #11734, opened 2023-07-10
Vinyl Control: Show Signal Quality On Skin, not working


NotYourAverageAl:
### Bug Description

The Vinyl Control option "Show Signal Quality On Skin", does not work on the latest 2.4 beta. The signal quality widget no longer shows on the spinnies when set in the preferences.

Thought it might be related to @m0dB 's GLSL work on the spinnies and I can confirm that.

Snapshot 2.4-alpha-1512-gbf1102212f works, but after [10989](https://github.com/mixxxdj/mixxx/pull/10989) it does not.

### Version

2.4-beta-48-g2cb54d9281

### OS

Windows 11

m0dB:
Thanks, I will take a look later this week.

ronso0:
Might as well be a regression of #11272, I'll take a look.

ronso0:
> Might as well be a regression of ~#11272~ #11663 , I'll take a look.

Nope, it's not, VC manager is still updating from config when clicking Apply

ronso0:
Another aspect of this bug seems to be that the vinyl signal quality is drawn in the preferences only if "Show quality in skins" is checked (+ Apply ofc). Should be drawn in the preferences regardless (as before 8aef3199f436046d7670008ca332230ab0cd0679)

Btw I was curious and started to git bisect but that's impossible due to unbuildable commits  :|
for example
```
/src/waveform/renderers/glwaveformrenderer.h:13:15: error: class ‘GLWaveformRenderer’ does not have any field named ‘WaveformRendererSignalBase’
   13 |             : WaveformRendererSignalBase(waveformWidgetRenderer)
      |               ^~~~~~~~~~~~~~~~~~~~~~~~~~

```

m0dB:
I have this under control for the spinnies, but I need to figure out the most efficient way to do this. I could commit a temporary fix in the meantime. However, I would like to really test that it works, but I don't have a vinyl controller myself. Is there a way to emulate this? Is that just a matter of feeding an audio input with the right signal? Where do I get such a signal (all this is new to me).

I haven't investigated what is drawn (or not) in the preferences yet. Not sure that is related to my changes.

JoergAtGithub:
You could download the Serato Control Signal and burn it to a CD: https://serato.com/controlcd/downloads?manualsanddownloads#manuals-and-download

ronso0:
[...]

PR #11743, opened 2023-07-14
VinylSignalQuality in WSpinnyGLSL

m0dB:
transfer the VinylSignalQualityReport scope to a texture with glTexSubImage2D and use a special shader to draw the VinylSignalQuality overlay on WSpinnyGLSL

Fixes https://github.com/mixxxdj/mixxx/issues/11734

### DIFF
--- a/src/widget/wspinnybase.cpp
+++ b/src/widget/wspinnybase.cpp
@@ -220,11 +220,13 @@ void WSpinnyBase::setup(const QDomNode& node,
 
     m_pVinylControlEnabled = new ControlProxy(
             m_group, "vinylcontrol_enabled", this, ControlFlag::NoAssertIfMissing);
+    updateVinylControlEnabled(m_pVinylControlEnabled->get());
     m_pVinylControlEnabled->connectValueChanged(this,
             &WSpinnyBase::updateVinylControlEnabled);
 
     m_pSignalEnabled = new ControlProxy(
             m_group, "vinylcontrol_signal_enabled", this, ControlFlag::NoAssertIfMissing);
+    updateVinylControlSignalEnabled(m_pSignalEnabled->get());
     m_pSignalEnabled->connectValueChanged(this,
             &WSpinnyBase::updateVinylControlSignalEnabled);
 
--- a/src/widget/wspinnybase.cpp
+++ b/src/widget/wspinnybase.cpp
@@ -48,6 +48,7 @@ WSpinnyBase::WSpinnyBase(
           m_iVinylInput(-1),
           m_bVinylActive(false),
           m_bSignalActive(true),
+          m_bDrawVinylSignalQuality(false),
           m_iVinylScopeSize(0),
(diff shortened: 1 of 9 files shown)


NotYourAverageAl:
I can confirm this fixes the issue. Thanks @m0dB

m0dB:
Great! Thanks for confirming.

JoergAtGithub:
I can confirm this too. I noticed one bug (which also exist in 2.3), if I switch the skin the VinylSignalQuality donut disappears until I toggle the enable VinylControl button.

m0dB:
I will have a look to see if that's an easy fix. While at it...

src/widget/wspinnybase.cpp
```diff
@@ -106,6 +107,10 @@ WSpinnyBase::~WSpinnyBase() {
 #endif
 }
 
+bool WSpinnyBase::shouldDrawVinylQuality() const {
+    return m_bVinylActive && m_bSignalActive && m_bDrawVinylSignalQuality;
+}
```
JoergAtGithub, line 112:
```suggestion
bool WSpinnyBase::shouldDrawVinylQuality() const {
#if defined __VINYLCONTROL__
    return m_bVinylActive && m_bSignalActive && m_bDrawVinylSignalQuality;
#else
    return false;
#endif
}
```
This could simplify the code which calls `shouldDrawVinylQuality()`

ronso0:
> if I switch the skin the VinylSignalQuality donut disappears until I toggle the enable VinylControl button.

I think this is a feature: IIRC the VC processor is idle as long as VC is off, even with input devices configured, so there's no signal to analyze.
I set up VC inputs just for passthrough and I'd like to keep the CPU load low.

JoergAtGithub:
The VinylControl button stays in ON state when I switch the skin. Just the donut disappears until I switched it first to OFF and than to ON again. This is clearly a bug an not a feature - but not an important bug of cause.

src/widget/wspinnybase.cpp
m0dB: Are you suggesting to removed the ifdefs there? Makes sense. Actually, this will never become true if __VINYLCONTROL__ is not defined, but I guess it's more clear like this.

m0dB:
> The VinylControl button stays in ON state when I switch the skin. Just the donut disappears until I switched it first to OFF and than to ON again. This is clearly a bug an not a feature - but not an important bug of cause.

I agree it's a bug, I think @ronso0 misread your comment. So I went ahead and fixed it. Not an important bug as you say but the fix was very simple.

With that, can this PR be approved and merged so we don't forget about it?

ronso0:
yes, I misread that, sorry for the noise. (should be more thorough when I do review on mobile B)

JoergAtGithub:
I just tested this and it worked flawless. I could switch between all waveformtypes and the signal quality donut always appeared when it should. Thank you!

Issue #11709, opened 2023-07-03
crash during mix disabling quantize


ywwg:
### Bug Description

Twice now (in many hours of DJing) I have had a crash when disabling quantize on deck 1. It is very, very hard to reproduce. 

```
#0  0x0000561aa8d934cc in void std::__insertion_sort<QList<QSharedPointer<WaveformMark> >::iterator, __gnu_cxx::__ops::_Iter_less_iter>(QList<QSharedPointer<WaveformMark> >::iterator, QList<QSharedPointer<WaveformMark> >::iterator, __gnu_cxx::__ops::_Iter_less_iter) [clone .isra.0] ()
[Current thread is 1 (Thread 0x7f42e99d8380 (LWP 117578))]
(gdb) bt
#0  0x0000561aa8d934cc in void std::__insertion_sort<QList<QSharedPointer<WaveformMark> >::iterator, __gnu_cxx::__ops::_Iter_less_iter>(QList<QSharedPointer<WaveformMark> >::iterator, QList<QSharedPointer<WaveformMark> >::iterator, __gnu_cxx::__ops::_Iter_less_iter) [clone .isra.0] ()
#1  0x0000561aa8d97004 in WOverview::updateCues(QList<CuePointer> const&) ()
#2  0x0000561aa8d9799c in WOverview::onMarkChanged(double) ()
#3  0x00007f42f24f1793 in  () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#4  0x0000561aa8790064 in ControlProxy::valueChanged(double) ()
#5  0x00007f42f24e741e in QObject::event(QEvent*) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#6  0x00007f42f556c713 in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /lib/x86_64-linux-gnu/libQt5Widgets.so.5
#7  0x00007f42f24b9e3a in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#8  0x00007f42f24bcf27 in QCoreApplicationPrivate::sendPostedEvents(QObject*, int, QThreadData*) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#9  0x00007f42f2513a67 in  () at /lib/x86_64-linux-gnu/libQt5Core.so.5
[...]

PR #11744, opened 2023-07-15
Fix crash disabling quantize

daschuer:
Make m_marksToRender a std::map to have a constant sort key.

This fixes the crash due to changing sort keys reported in https://github.com/mixxxdj/mixxx/issues/11709

### DIFF
--- a/src/waveform/renderers/waveformmark.h
+++ b/src/waveform/renderers/waveformmark.h
@@ -1,5 +1,4 @@
 #pragma once
-
 #include <QDomNode>
 #include <QImage>
 
@@ -92,6 +91,11 @@ class WaveformMark {
 
 typedef QSharedPointer<WaveformMark> WaveformMarkPointer;
 
+// This class provides an immutable sortkey for the WaveformMark using sample
+// position and hotcue number. IMPORTANT: The Mark's position may be changed after
+// a key's creation, and those updates will not be reflected in these sortkeys.
+// Currently they are used to render marks on the Overview, a situation where
+// temporarily incorrect sort order is acceptable.
 class WaveformMarkSortKey {
   public:
     WaveformMarkSortKey(double samplePosition, int hotcue)
--- a/src/waveform/renderers/waveformmark.h
+++ b/src/waveform/renderers/waveformmark.h
@@ -92,21 +92,29 @@ class WaveformMark {
 
 typedef QSharedPointer<WaveformMark> WaveformMarkPointer;
 
-inline bool operator<(const WaveformMarkPointer& lhs, const WaveformMarkPointer& rhs) {
-    double leftPosition = lhs->getSamplePosition();
-    int leftHotcue = lhs->getHotCue();
-    double rightPosition = rhs->getSamplePosition();
-    int rightHotcue = rhs->getHotCue();
-    if (leftPosition == rightPosition) {
-        // Sort WaveformMarks without hotcues before those with hotcues;
-        // if both have hotcues, sort numerically by hotcue number.
-        if (leftHotcue == Cue::kNoHotCue && rightHotcue != Cue::kNoHotCue) {
-            return true;
-        } else if (leftHotcue != Cue::kNoHotCue && rightHotcue == Cue::kNoHotCue) {
-            return false;
-        } else {
-            return leftHotcue < rightHotcue;
-        }
+class WaveformMarkSortKey {
+  public:
+    WaveformMarkSortKey(double samplePosition, int hotcue)
+            : m_samplePosition(samplePosition),
+              m_hotcue(hotcue) {
     }
-    return leftPosition < rightPosition;
-}
+
+    bool operator<(const WaveformMarkSortKey& other) const {
+        if (m_samplePosition == other.m_samplePosition) {
+            // Sort WaveformMarks without hotcues before those with hotcues;
+            // if both have hotcues, sort numerically by hotcue number.
+            if (m_hotcue == Cue::kNoHotCue && other.m_hotcue != Cue::kNoHotCue) {
+                return true;
+            } else if (m_hotcue != Cue::kNoHotCue && other.m_hotcue == Cue::kNoHotCue) {
+                return false;
+            } else {
+                return m_hotcue < other.m_hotcue;
+            }
+        }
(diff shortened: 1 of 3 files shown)


ywwg:
I'll set this up with my js spammer just to confirm the fix :)

src/waveform/renderers/waveformmark.h
```diff
@@ -122,21 +122,29 @@ class WaveformMark {
...
-        // if both have hotcues, sort numerically by hotcue number.
-        if (leftHotcue == Cue::kNoHotCue && rightHotcue != Cue::kNoHotCue) {
-            return true;
-        } else if (leftHotcue != Cue::kNoHotCue && rightHotcue == Cue::kNoHotCue) {
-            return false;
-        } else {
-            return leftHotcue < rightHotcue;
+class WaveformMarkSortKey {
+  public:
+    WaveformMarkSortKey(double samplePosition, int hotcue)
```
uklotzde, line 127: Should be constructed from a WaveformMark.

src/waveform/renderers/waveformmark.h
```diff
@@ -122,21 +122,29 @@ class WaveformMark {
...
-    if (leftPosition == rightPosition) {
-        // Sort WaveformMarks without hotcues before those with hotcues;
-        // if both have hotcues, sort numerically by hotcue number.
-        if (leftHotcue == Cue::kNoHotCue && rightHotcue != Cue::kNoHotCue) {
-            return true;
-        } else if (leftHotcue != Cue::kNoHotCue && rightHotcue == Cue::kNoHotCue) {
-            return false;
-        } else {
-            return leftHotcue < rightHotcue;
+class WaveformMarkSortKey {
```
uklotzde, line 125: In the future no one will understand why this class has been introduced for storing seemingly redundant information.

uklotzde (changes requested):
Using a std::map here might be convenient but is the wrong choice for this use case.

src/waveform/renderers/waveformmark.h
daschuer: Why? This is IMHO a mater of taste.

src/waveform/renderers/waveformmark.h
daschuer: I will add a comment.

daschuer:
@uklotzde, your review is pretty useless if you don't explain you opinion and don't add a proposal for improvements.

uklotzde:
A map is uses for dynamically adding and removing elements. In this case the keys are stable, but become inconsistent at the time when they are created. You will not be able to remove an element reliably once by its key once it has been inserted. This only gets worse while time passes.

Using a map is deceiving. A developer will easily create a key from a WaveformMarkPointer and try to use it for accessing the map. Using a map makes it easy to make this and other mistakes.

uklotzde:
I am done with my "useless" review.

ywwg:
I see the point, that usually the key in a map is an accessor, and that the sample position is not really a useful way to access the items in this map.  However this code is isolated and this hack is pretty self-contained.  It is unlikely that this choice will cause large problems down the line, and if this turns out to be an inconvenient solution we can leave enough comments to explain why this choice was made.  I think this is fine for now to fix the severe party-crashing bug.

We are always open to specific alternative approaches to fixing issues. I think ideally the solution would involve a dirty flag and snapshotting the cues at render time.  Most likely all this code will disappear when the QML rewrite is done, so I'm not sure it's worth doing a 100%-correct-and-pretty solution here.

ywwg (approved):

daschuer:
Thank you for review. I have rebased this to 2.3. because the 2.3 branch is also affected form the crasher.

uklotzde:
> so I'm not sure it's worth doing a 100%-correct-and-pretty solution here.

Please note that I deliberately did not request "a 100%-correct-and-pretty solution"! I only pointed out that the current design and code (which includes comments) is prone to introducing new bugs.

JoergAtGithub:
The text of the commit message seems to be accidentally here

daschuer:
Oh, yes. I must have messed up the rebase to 2.3. Now the old message is back.

ywwg:
> Please note that I deliberately did not request "a 100%-correct-and-pretty solution"! I only pointed out that the current design and code (which includes comments) is prone to introducing new bugs.

I appreciate the review, truly. I think this solution is at least incrementally better than the current proven-dangerous code.  Would you be opposed to merging as-is, or is it sufficient that we have raised the issue and are aware of the possible downsides?

uklotzde:
It is the duty of the PR owner to choose a suitable option to resolve the questions and concerns that have been expressed. I didn't exclude any of them.

daschuer:
This this is still a regular map and I cannot confirm your concerns. The issue for me is that your complains are kind of common, not specific to this use case of this private container. I have considered different solutions and this is IMHO the best best we can pick in terms of performance and architecture. But maybe I am wrong, you can prove it by providing a different solution.
For now I will tweak the comments a bit and than I am done.

src/waveform/renderers/waveformmark.h
```diff
@@ -92,21 +91,32 @@ class WaveformMark {
...
-        // if both have hotcues, sort numerically by hotcue number.
-        if (leftHotcue == Cue::kNoHotCue && rightHotcue != Cue::kNoHotCue) {
-            return true;
-        } else if (leftHotcue != Cue::kNoHotCue && rightHotcue == Cue::kNoHotCue) {
-            return false;
-        } else {
-            return leftHotcue < rightHotcue;
+// This class stores the data used as a sorting key immutable.
+// We cannot use the the mark directly, because it can be
+// altered at any time.
```
ywwg, line 96:
```suggestion
// This class provides an immutable sortkey for the WaveformMark using sample position and hotcue enum. IMPORTANT: the Mark's position may be changed after a key's creation, and those updates will not be reflected in these sortkeys. Therefore these keys should not be retained for long periods of time. Currently they are used to render marks on the Overview, a situation where temporarily incorrect mark position is acceptable. 
```

ywwg: (Feel free to copy and paste to fix newlines, etc)

daschuer:
Done.

PR #11746, opened 2023-07-15
shortkeys to cut/copy/paste/remove tracks

m0dB:
(Below where I say Ctrl, this is Cmd on macOS)

This PR implements short-keys for Cut, Copy, Paste (keysequences as in https://doc.qt.io/qt-6/qkeysequence.html, typically Ctrl-x Ctrl-c Ctrl-v) and Remove (Delete, Backspace) of tracks in the library table view (Tracks, Playlists, Crates and the Auto DJ queue).

The focus has to be on the table view for the short-keys to work.

- Copy copies the selected track(s) to the clipboard. (What is actually copied are the urls, so you can paste the clipboard in a text editor as well)

- Cut is as copy, but removes the tracks from their original location.

- Remove just removes, without places the tracks on the clipboard.

- Paste inserts the track(s) from the clipboard at the position of the currently selected track in the case of Playlists and the Auto DJ queue. In the case of Crates or when no track is selected, the track(s) from the clipboard are appended at the end.

In addition to this:

- Esc will clear the selection

- Ctrl Left and Ctrl Right move the focus between the sidebar panel and the library panel. This could already be done with Tab and Shift-Tab, but I find this much more intuitive, and much easier if you already are with one hand on Ctrl for Ctrl X/C/V and the other hand on the cursors for navigating the table view.

Notes:

- I have had my doubts about the insertion point for paste. My only reference is Spotify, which inserts below the currently selected track (or below the selected track lowest in the list, when multiple tracks or selected). But this means that you cannot paste at the first position, so to place a track at the first position you will have to do multiple operations. I am using the currently selected row is the insert point. This solves this problem and it is straightforwards: my current row is where I insert.

But I am open to reconsidering this. Does this work for you? Can you compare with other applications? I can even imagine that this should be a preference (a radio button "Paste below selected row(s)" / "Paste inserts at currently selected row"

- As mentioned, the short-keys only work when the focus is on the table view. I have found myself trying to paste immediately after selecting a Playlist, a Crate or the Auto DJ queue, without first changing the focus to the table view and wondering why nothing was happening.

It would make sense to implement paste in this situation (appending at the end of the selected track set). But it does open a pandora’s box. What to do with the other short-keys when the focus is on the sidebar? In any case, I prefer to go one step at a time and I think this PR is valid without this addition.

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -369,13 +369,13 @@ void BaseTrackTableModel::cutTracks(const QModelIndexList& indices) {
 }
 
 void BaseTrackTableModel::copyTracks(const QModelIndexList& indices) const {
-    Clipboard::begin();
+    Clipboard::start();
     for (const QModelIndex& index : indices) {
         if (index.isValid()) {
             Clipboard::add(QUrl::fromLocalFile(getTrackLocation(index)));
         }
     }
-    Clipboard::end();
+    Clipboard::finish();
 }
 
 int BaseTrackTableModel::pasteTracks(const QModelIndex& insertionIndex) {
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -19,6 +19,7 @@
 #include "moc_basetracktablemodel.cpp"
 #include "track/track.h"
 #include "util/assert.h"
+#include "util/clipboard.h"
 #include "util/datetime.h"
 #include "util/db/sqlite.h"
 #include "util/logger.h"
@@ -360,6 +361,30 @@ int BaseTrackTableModel::columnCount(const QModelIndex& parent) const {
         return 0;
     }
     return countValidColumnHeaders();
+}
+
+void BaseTrackTableModel::cutTracks(const QModelIndexList& indices) {
+    copyTracks(indices);
+    removeTracks(indices);
+}
+
+void BaseTrackTableModel::copyTracks(const QModelIndexList& indices) const {
+    Clipboard::begin();
+    for (const QModelIndex& index : indices) {
+        if (index.isValid()) {
+            Clipboard::add(QUrl::fromLocalFile(getTrackLocation(index)));
+        }
+    }
+    Clipboard::end();
+}
+
+int BaseTrackTableModel::pasteTracks(const QModelIndex& insertionIndex) {
+    const QList<QUrl> urls = Clipboard::urls();
+    const QList<TrackId> trackIds = m_pTrackCollectionManager->resolveTrackIdsFromUrls(urls, false);
+    if (trackIds.isEmpty()) {
+        return 0;
+    }
+    return addTracksWithTrackIds(insertionIndex, trackIds);
 }
 
 bool BaseTrackTableModel::isColumnHiddenByDefault(
--- a/src/library/sidebarmodel.cpp
+++ b/src/library/sidebarmodel.cpp
@@ -477,7 +477,7 @@ QModelIndex SidebarModel::translateIndex(
 }
 
 void SidebarModel::slotDataChanged(const QModelIndex& topLeft, const QModelIndex& bottomRight) {
-    //qDebug() << "slotDataChanged topLeft:" << topLeft << "bottomRight:" << bottomRight;
+    // qDebug() << "slotDataChanged topLeft:" << topLeft << "bottomRight:" << bottomRight;
     QModelIndex topLeftTranslated = translateSourceIndex(topLeft);
     QModelIndex bottomRightTranslated = translateSourceIndex(bottomRight);
     emit dataChanged(topLeftTranslated, bottomRightTranslated);
@@ -501,8 +501,8 @@ void SidebarModel::slotRowsInserted(const QModelIndex& parent, int start, int en
     Q_UNUSED(parent);
     Q_UNUSED(start);
     Q_UNUSED(end);
-    //qDebug() << "slotRowsInserted" << parent << start << end;
-    //QModelIndex newParent = translateSourceIndex(parent);
+    // qDebug() << "slotRowsInserted" << parent << start << end;
+    // QModelIndex newParent = translateSourceIndex(parent);
     endInsertRows();
 }
 
--- a/src/util/clipboard.cpp
+++ b/src/util/clipboard.cpp
@@ -3,21 +3,16 @@
 #include <QApplication>
 #include <QClipboard>
 
-QString& Clipboard::text() {
-    static QString s_text;
-    return s_text;
-}
-
-void Clipboard::begin() {
-    text() = "";
+void Clipboard::start() {
+    instance()->m_text.clear();
 }
 
 void Clipboard::add(const QUrl& url) {
-    text().append(url.toString() + "\n");
+    instance()->m_text.append(url.toString() + "\n");
 }
 
-void Clipboard::end() {
-    QApplication::clipboard()->setText(text());
+void Clipboard::finish() {
+    QApplication::clipboard()->setText(instance()->m_text);
 }
 
 QList<QUrl> Clipboard::urls() {
@@ -28,7 +23,7 @@ QList<QUrl> Clipboard::urls() {
     QStringList strings = text.split("\n", QString::SkipEmptyParts);
 #endif
     QList<QUrl> result;
-    for (QString string : strings) {
+    for (const QString& string : strings) {
         result.append(QUrl(string));
     }
     return result;
--- a/src/util/clipboard.cpp
+++ b/src/util/clipboard.cpp
@@ -1,0 +1,35 @@
+#include "clipboard.h"
+
+#include <QApplication>
+#include <QClipboard>
+
+QString& Clipboard::text() {
+    static QString s_text;
+    return s_text;
+}
+
+void Clipboard::begin() {
+    text() = "";
+}
+
+void Clipboard::add(const QUrl& url) {
+    text().append(url.toString() + "\n");
+}
+
+void Clipboard::end() {
+    QApplication::clipboard()->setText(text());
+}
+
+QList<QUrl> Clipboard::urls() {
+    const QString text = QApplication::clipboard()->text();
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+    QStringList strings = text.split("\n", Qt::SkipEmptyParts);
+#else
+    QStringList strings = text.split("\n", QString::SkipEmptyParts);
+#endif
+    QList<QUrl> result;
+    for (QString string : strings) {
+        result.append(QUrl(string));
+    }
+    return result;
+}
--- a/src/util/clipboard.cpp
+++ b/src/util/clipboard.cpp
@@ -2,29 +2,23 @@
 
 #include <QApplication>
 #include <QClipboard>
+#include <QMimeData>
 
 void Clipboard::start() {
-    instance()->m_text.clear();
+    instance()->m_urls.clear();
 }
 
 void Clipboard::add(const QUrl& url) {
-    instance()->m_text.append(url.toString() + "\n");
+    instance()->m_urls.append(url);
 }
 
 void Clipboard::finish() {
-    QApplication::clipboard()->setText(instance()->m_text);
+    QMimeData* data = new QMimeData;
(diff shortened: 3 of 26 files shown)


daschuer:
Thank you.

There is a clang-tidy error:
```
Error: /home/runner/work/mixxx/mixxx/src/util/clipboard.cpp:31:10: error: Missing reference in range-for with non trivial type (QString) [-Wclazy-range-loop-reference]
    for (QString string : strings) {
         ^
         const   &
1 error generated.
```

ronso0:
Just did a quick test with Tracks and playlists, and it works good as expected, thanks!
Ctrl+L/R makes absolute sense, and it doesn't interfere with cell editing where it triggers jumping by words.

Though, in Tracks, why does Ctrl+X deselect the current index (no matter if it's single- or multi-selection) while Ctrl+C doesn't change the selection?

And indeed Ctrl+V in the sidebar would be handy for appending tracks. Guess that's quite some work though..

ronso0:
@m0dB would you mind splitting this up into isolated commits (before any code review is started)?

If yes, please make sure each commit builds.

src/library/basetracktablemodel.cpp
```diff
@@ -362,6 +363,30 @@ int BaseTrackTableModel::columnCount(const QModelIndex& parent) const {
     return countValidColumnHeaders();
 }
 
+void BaseTrackTableModel::cutTracks(const QModelIndexList& indices) {
+    copyTracks(indices);
+    removeTracks(indices);
```
daschuer, line 368: Just a thought: 
My expectation was that the removeTracks is delayed until the past was successful. 

On the other hand this is no file manager and when doing it in a text editor the text is immediately removed. 
But there is a undo feature Ctrl-Z ..

Did you consider this?

ronso0: I agree this doesn't match the usual file manager behaviour.
However, tracks are removed on Cut, not just greyed out as in file managers, hence I think the current solution provides predictable UX.

FWIW cut & paste works reliably as Undo, i.e. tracks positions in playlists are properly restored regardless the sorting.

m0dB: Implementing a proper undo should be feasible though and pretty simple if it's just one step deep.

m0dB:
> would you mind splitting this up into isolated commits (before any code review is started)?

Not sure how to split this one up though and not sure why. It's not that big and it's all related.

m0dB:
> Just did a quick test with Tracks and playlists, and it works good as expected, thanks! Ctrl+L/R makes absolute sense, and it doesn't interfere with cell editing where it triggers jumping by words.

Great :-)
 
> Though, in Tracks, why does Ctrl+X deselect the current index (no matter if it's single- or multi-selection) while Ctrl+C doesn't change the selection?

As you cut the row that is selected, it seemed to me that it made sense to unselect after cut. But I can change this, I see what you mean.

> And indeed Ctrl+V in the sidebar would be handy for appending tracks. Guess that's quite some work though..

Not really, I already implemented this, but I didn't want to add too much in one PR.

I do see a bit of a problem that this might create the expectation that Ctrl-X and Ctrl-C also work in the sidebar... But maybe that's okay.

src/util/clipboard.cpp
```diff
@@ -0,0 +1,35 @@
...
+void Clipboard::begin() {
+    text() = "";
+}
+
+void Clipboard::add(const QUrl& url) {
+    text().append(url.toString() + "\n");
+}
+
+void Clipboard::end() {
+    QApplication::clipboard()->setText(text());
```
daschuer, line 20: setText() only a handy wrapper for text. 
See https://github.com/qt/qtbase/blob/d6dacfecb1add4bfd3e4d0aeea5047dd95206588/src/gui/kernel/qclipboard.cpp#L297

Our data is a list of tracks instead. 

They can be copied to the systems Chipboard by 
QMimeData::setUrls()

This has been successfully implemented fro drag and drop here: 
https://github.com/mixxxdj/mixxx/blob/3de29630a80513127f8d3c0e5eb006922d240d08/src/util/dnd.cpp#L13

I think we should do the same for fully interoperability with other applications.

daschuer: This gives some insight for Linux:
https://www.medo64.com/2019/12/copy-to-clipboard-in-qt/
Not sure if this is still relevant.

src/widget/wlibrary.cpp
```diff
@@ -142,3 +142,9 @@ bool WLibrary::event(QEvent* pEvent) {
     }
     return QStackedWidget::event(pEvent);
 }
+
+void WLibrary::keyPressEvent(QKeyEvent* event) {
+    if (event->key() == Qt::Key_Left && event->modifiers() & Qt::ControlModifier) {
+        emit setLibraryFocus(FocusWidget::Sidebar);
+    }
```
daschuer, line 149: Do we need to chain call the base class implementation here?

daschuer (changes requested):
Thank you for this nice feature. 

I have left some comments.

src/library/sidebarmodel.cpp
```diff
@@ -501,8 +501,8 @@ void SidebarModel::slotRowsInserted(const QModelIndex& parent, int start, int en
     Q_UNUSED(parent);
     Q_UNUSED(start);
     Q_UNUSED(end);
-    //qDebug() << "slotRowsInserted" << parent << start << end;
-    //QModelIndex newParent = translateSourceIndex(parent);
+    // qDebug() << "slotRowsInserted" << parent << start << end;
+    // QModelIndex newParent = translateSourceIndex(parent);
```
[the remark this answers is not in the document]
ronso0: pre-commit demands this extra space

ronso0:
> Not sure how to split this one up though and not sure why. It's not that big and it's all related.

Never mind, it's just that I like small chunks as they are easier to review, i.e. more encouraging to start a review.

src/util/clipboard.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include <QApplication>
+#include <QClipboard>
+
+QString& Clipboard::text() {
+    static QString s_text;
+    return s_text;
+}
+
+void Clipboard::begin() {
+    text() = "";
```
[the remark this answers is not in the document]
m0dB: I have changed the functions to start() and finish() and I have turned the Clipboard class into a proper Singleton (which might be a bit overkill, but it doesn't hurt)

src/library/sidebarmodel.cpp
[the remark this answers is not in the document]
m0dB: My editor doesn't format, this was indeed pre-commit. Pre-commit wouldn't normally touch this, but I made changes here which I later reverted, but this pre-commit change made it through the rebase. Anyway, no worries, @daschuer, I do have pre-commit set up correctly :-)

src/util/clipboard.cpp
m0dB: Hm, indeed it seems to make more sense to use QClipboard::setMimeData directly. Not sure why I didn't do that. I will give it a try. Thanks for the suggestion.

src/widget/wlibrary.cpp
m0dB: I guess it's cleaner to do so, but I don't think QStackedWidget does any key handling. Anyway, I will make the change.

m0dB: Done

src/util/clipboard.cpp
m0dB: Done

m0dB:
I believe I have addressed all comments.

I can implement Ctrl+V in the sidebar in this same PR or in a different one. I would prefer the later; I think that without Ctrl+V in the sidebar the functionality so far is already very useable and I'd like to keep the life of the PRs short.

daschuer:
Thank you. It works now better, but is not fully compatible to Nemo File manager. 
Via Drag and Drop I can pull files from and to Mixxx. 
I have noticed a difference when pulling and Ctrl+C/V form Nemo File manager to Gedit Text editor. In the Ctrl+C case the path is pasted (without file://) and in the pull case it opens the file.

daschuer:
I have added this code for debugging: 
```
    for(QString& formatName: QApplication::clipboard()->mimeData()->formats()) {
        QByteArray arr = QApplication::clipboard()->mimeData()->data(formatName);
        qDebug() << "name=" << formatName << arr.size();
        qDebug() << arr;
    }
```

The output when Ctrl+V in Mixxx is: 
```
Debug [Main] keyboard press:  "Ctrl+V"
Debug [Main] name= "text/uri-list" 94
Debug [Main] "file:///home/daniel/Musik/Abba%20-%20Best%20Of%20Abba/02%20Abba%20-%20Andante,%20Andante.m4a\r\n"
```

Doing the same form Nemo file manager to Mixxx is: 
```
Debug [Main] keyboard press:  "Ctrl+V"
Debug [Main] name= "TIMESTAMP" 4
Debug [Main] "\xC4L\x87\x02"
Debug [Main] name= "TARGETS" 44
Debug [Main] "\xC0\x01\x00\x00\xBE\x01\x00\x00\xBF\x01\x00\x00=\x04\x00\x00\\\x02\x00\x00X\x01\x00\x00\x1A\x02\x00\x00\x1B\x02\x00\x00\x1F\x00\x00\x00\x1C\x02\x00\x00\x1D\x02\x00\x00"
Debug [Main] name= "MULTIPLE" 0
Debug [Main] ""
Debug [Main] name= "x-special/gnome-copied-files" 43
Debug [Main] "copy\nfile:///home/sperry/Musik/noise.2_.mp3"
Debug [Main] name= "text/uri-list" 40
Debug [Main] "file:///home/daniel/Musik/noise.2_.mp3\r\n"
Debug [Main] name= "text/plain" 31
Debug [Main] "/home/sperry/Musik/noise.2_.mp3"
Debug [Main] name= "COMPOUND_TEXT" 31
Debug [Main] "/home/sperry/Musik/noise.2_.mp3"
Debug [Main] name= "text/plain;charset=utf-8" 31
Debug [Main] "/home/sperry/Musik/noise.2_.mp3"

This is also mentioned in:
https://doc.qt.io/qt-6/qclipboard.html#notes-for-x11-users

**Review by daschuer — APPROVED** [review]
LGTM. Pasting into the Nemo File manager does not work. 
I am not sure if there is a strong demand. So we should not bock this PR over it. 

Is this than ready to merge?

File: src/widget/wlibrarytableview.cpp [hunk 6]
```diff
@@ -267,12 +267,15 @@ void WLibraryTableView::focusInEvent(QFocusEvent* event) {
     QTableView::focusInEvent(event);
 
     if (event->reason() == Qt::TabFocusReason ||
-            event->reason() == Qt::BacktabFocusReason) {
+            event->reason() == Qt::BacktabFocusReason ||
+            event->reason() == Qt::OtherFocusReason) {
```
ronso0, line 271: I was wondering why Return in the searchbar would jump to the table but the selection trick wasn't working anymore lately (2.4).
I'll pick this for 2.4

ronso0: https://github.com/mixxxdj/mixxx/commit/19a1bdc8639c7bf2dd5d2019bcbe925949e29d55

ronso0:
> Though, in Tracks, why does Ctrl+X deselect the current index (no matter if it's single- or multi-selection) while Ctrl+C doesn't change the selection?

_ping_

m0dB:
Hi @ronso0 , @daschuer ,

Sorry for having been unresponsive. I am still working on this: I found a problem with the selection of the pasted tracks in playlists when the sort order is not position. I have implemented a solution, but it still needs a bit of cleanup. (The library code certainly isn't the easiest code to deal with and it's been quite a bit of trial and error...)

I also have the short-keys working when the focus is on the sidebar, but this also needs a bit of cleanup. As I said before, I will do that in a follow-up PR, but I would like to have that PR ready immediately, because I do feel that without this addition the user experience is not ideal -  I have been trying to paste immediately after selecting a playlist in the sidebar way too often!

I will be on holidays for the next two weeks, and I will be travelling and have irregular internet access. But it is my intention to find the time during these two weeks and finish this.

ronso0:
This branch targets main, so no hurry, don't worry

m0dB:
Replaced by https://github.com/mixxxdj/mixxx/pull/12020

PR #11756, opened 2023-07-23
Pref Controller: avoid scrollbars in I/O tabs if Info tab exceeds page height

ronso0:
Scroll only Info tab if required while I/O tabs still fit the page. Otherwise the I/O tabs expand equally and Add, Remove, Clear buttons are hidden.

(backported 2.4 fix, didn't check this locally yet)

### DIFF
--- a/src/controllers/dlgprefcontrollerdlg.ui
+++ b/src/controllers/dlgprefcontrollerdlg.ui
@@ -29,7 +29,13 @@
       <attribute name="title">
        <string>Controller Setup</string>
       </attribute>
-      <layout class="QGridLayout" name="gridLayout_2">
+      <property name="sizePolicy">
+       <sizepolicy hsizetype="Ignored" vsizetype="Ignored">
+        <horstretch>0</horstretch>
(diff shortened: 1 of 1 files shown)


daschuer:
I have just tested it on top of the 2.3 branch,  but I cannot spot the difference. I have probably not yet understand what the issue is. At least I see no regression.

ronso0:
The Info tab correctly expands if the content doesn't fit the page.
Screenshot_2023-07-25_18-16-13
Though that forces the I/O tabs o expand as well (needlessly as the table can shrink) and buttons are hidden, see double scrollbars
Screenshot_2023-07-25_18-20-04

This PR fixes that: Info may expand, In/Out are fit:
Screenshot_2023-07-25_18-21-50

daschuer:
Strange, I cannot confirm the issue here on Ubuntu Focal. But since there is an evidence that it is fixed on your side and I do not see a regression, it should be save to merge.

daschuer (approved):
LGTM

Issue #11545, opened 2023-05-09
Warning: Overlapping range


daschuer:
### Bug Description

I have see this warning when shutting down Mixxx with an m4a file playing: 
```
warning [CachingReaderWorker 1] ReadAheadFrameBuffer - Overlapping range [960 -> 1024) in output buffer [1024 -> 8192) with input buffer [960 -> 1984)
```

### Version

2.4-alpha

### OS

Ubunti 20.4 Focal

daschuer:
This might be related to a bugfix for ffmpeg 4.4 introduced in 2.3.1 here: 
https://github.com/mixxxdj/mixxx/issues/10464
Ubuntu Focal uses ffmpeg 2.2

daschuer:
I have seen also warnings about a shifted sound start cue along with this issue.

ywwg:
what is the effect of the warning? why is this a blocker for 2.3.6?

daschuer:
The effect is that a single chunk is shifted. The overlapping samples are dropped and the gap is filled with silence. 

I don't consider it a strict blocker because only older distros are affected, but it is part of the list of regressions, (Introduced in 2.3.1) and we have already a PR for it.

Swiftb0y:
closed by #11760

PR #11760, opened 2023-07-24
Fix overlapping buffers when decoding m4a files using ffmpeg

daschuer:
This fixes: https://github.com/mixxxdj/mixxx/issues/11545

The actual fix is in https://github.com/mixxxdj/mixxx/commit/07a90995c30ea1c0a4067afdfb7065c76ec30a3a

The rest removed unused code and improves const. 

To make it work with  shared libraries, we now read the ffmpge version at runtime and Apply the patch form https://github.com/mixxxdj/mixxx/issues/10464 only if required. 
I have tested this on Ubuntu Focal using FFmpeg 4.2 and 4.4 via a PPA.

### DIFF
--- a/src/sources/readaheadframebuffer.h
+++ b/src/sources/readaheadframebuffer.h
@@ -119,9 +119,6 @@ class ReadAheadFrameBuffer final {
     /// All discontinuities in the output stream, i.e. both gaps and overlapping
     /// regions are unexpected and will trigger a debug assertion.
     ///
-    /// The output sample buffer may be null. In this case the consumed
-    /// samples are dropped instead of copied.
-    ///
     /// Returns the remaining portion that could not be filled from
     /// the buffer.
     WritableSampleFrames consumeAndFillBuffer(ReadableSampleFrames inputBuffer,
--- a/src/sources/readaheadframebuffer.h
+++ b/src/sources/readaheadframebuffer.h
@@ -121,7 +121,8 @@ class ReadAheadFrameBuffer final {
     ///
     /// Returns the remaining portion that could not be filled from
     /// the buffer.
-    WritableSampleFrames consumeAndFillBuffer(ReadableSampleFrames inputBuffer,
+    WritableSampleFrames consumeAndFillBuffer(
+            const ReadableSampleFrames& inputBuffer,
             const WritableSampleFrames& outputBuffer,
             FrameIndex minOutputIndex);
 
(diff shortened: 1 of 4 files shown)


src/sources/readaheadframebuffer.h
```diff
@@ -92,17 +92,6 @@ class ReadAheadFrameBuffer final {
...
-    enum class DiscontinuityOverlapMode {
-        Ignore,
-        Rewind,
-        Default = Rewind, // recommended default
-    };
-    enum class DiscontinuityGapMode {
-        Skip,
-        FillWithSilence,
-        Default = FillWithSilence, // recommended default
-    };
```
Swiftb0y, line 104: I don't know why this was introduced in the first place but I assume there was a good reason for it. Is there a better reason for removing it or why did you choose to do so?

src/sources/soundsourceffmpeg.cpp
```diff
@@ -1159,24 +1159,27 @@ ReadableSampleFrames SoundSourceFFmpeg::readSampleFramesClamped(
...
+                        kLogger.debug()
+                                << "Skipping"
+                                << inaudibleFrameCountUntilStartOfStream
+                                << "inaudible sample frames before the start of the stream";
 #endif
-                    streamFrameIndex = kMinFrameIndex;
+                        streamFrameIndex = kMinFrameIndex;
+                    }
                 }
-                DEBUG_ASSERT(streamFrameIndex >= kMinFrameIndex);
```
Swiftb0y, line 1179: still don't understand why you removed this. Judging by naming this makes sense, why would `streamFrameIndex < kMinFrameIndex` be valid if `kMinFrameIndex` is supposed to denote a minimum?

src/sources/readaheadframebuffer.h
daschuer: I have removed it, because it was no longer used. This helped to have a better understanding how the code works and about the corner cases to consider.

src/sources/soundsourceffmpeg.cpp
daschuer: ffmpeg add priming samples before the requested samples start this is the reason why streamFrameIndex can be negative. 

So it turns out that kMinFrameIndex = 0 is badly named. It was also confusing to me. 
I think we should replace it with just 0 for a better readability.

daschuer: Since the frame count is correct before ffmpeg 4.4 the code below deals with that case. The hack of setting the start frame to zero is only required for newer ffmpeg versions. 
Do you have a suggestion to improve the code or comments here?

Swiftb0y: I don't think inlining the constant here improves readability. 
The question I rather have is whether its valid to have `streamFrameIndex < 0` before ffmpeg 4.4?

src/sources/readaheadframebuffer.h
Swiftb0y: Okay going through the code again, I also don't see much value in keeping the `Skip` and `Ignore` modes.

src/sources/soundsourceffmpeg.cpp
daschuer: Yes, the assertion will kick in here before 4.4

Swiftb0y: Okay, thanks for clarifying.

Swiftb0y (approved):
LGTM

PR #11764, opened 2023-07-25
WTrackProperty: double-clicked property field is focused in DlgTrackInfo

ronso0:
A little helper for tag editing:
* double-click [property] label to open Track Info
* if there is a line edit for [property] it is focused, start typing

### DIFF
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -62,6 +62,22 @@ void DlgTrackInfo::init() {
     setupUi(this);
     setWindowIcon(QIcon(MIXXX_ICON_PATH));
 
+    // Store tag edit widget pointers to allow focusing a specific widgets when
+    // this is opened by double-clicking a WTrackProperty label.
+    // Associate with property strings taken from library/dao/trackdao.h
+    m_propertyWidgets.insert("artist", txtArtist);
+    m_propertyWidgets.insert("title", txtTrackName);
+    m_propertyWidgets.insert("album", txtAlbum);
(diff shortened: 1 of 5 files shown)


src/library/dlgtrackinfo.cpp
```diff
@@ -484,6 +500,16 @@ void DlgTrackInfo::loadTrack(const QModelIndex& index) {
     }
 }
 
+void DlgTrackInfo::focusField(const QString& property) {
+    if (property.isNull() || property.isEmpty()) {
```
daschuer, line 504: isEmpty() includes isNull()
```suggestion
    if (property.isEmpty()) {
```

src/library/dlgtrackinfo.h
```diff
@@ -121,6 +123,8 @@ class DlgTrackInfo : public QDialog, public Ui::DlgTrackInfo {
     TapFilter m_tapFilter;
     mixxx::Bpm m_lastTapedBpm;
 
+    QMultiHash<QString, QWidget*> m_propertyWidgets;
```
daschuer, line 126: > A multi-valued hash is a hash that allows multiple values with the same key.
Is this the case here? 

```suggestion
    QHash<QString, QWidget*> m_propertyWidgets;
```

daschuer (changes requested):
The trick works here, left some comments.

src/library/dlgtrackinfo.h
ronso0: Nope, just dumb copy/pasting.
fixed

daschuer (approved):
Than you.

ronso0:
**TODO**
document double-click in the manual

PR #11766, opened 2023-07-25
Fix building Mixxx with SDK 11

daschuer:
Our minimum is SDK 10.12

### DIFF
--- a/src/library/itunes/itunesmacosimporter.mm
+++ b/src/library/itunes/itunesmacosimporter.mm
@@ -96,9 +96,12 @@ class ImporterImpl {
         // Filter out the primary playlist (since we already show the library
         // under the main iTunes node)
         bool isPrimary;
+#if (MAC_OS_X_VERSION_MAX_ALLOWED >= 120000)
         if (@available(macOS 12.0, *)) {
             isPrimary = [playlist isPrimary];
-        } else {
+        } else
+#endif
+        {
             isPrimary = [playlist isMaster];
         }
         if (isPrimary) {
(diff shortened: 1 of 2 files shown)


src/library/itunes/itunesmacosimporter.mm
```diff
@@ -96,9 +96,12 @@ bool isPlaylistShown(ITLibPlaylist* playlist) {
         // Filter out the primary playlist (since we already show the library
         // under the main iTunes node)
         bool isPrimary;
+#if (MAC_OS_X_VERSION_MAX_ALLOWED >= 120000)
         if (@available(macOS 12.0, *)) {
             isPrimary = [playlist isPrimary];
-        } else {
+        } else
+#endif
```
Swiftb0y, line 103: whats the error here? I figured if `@available` is asking for a version lower than the current SDK it would just return `false`. the official docs don't specify that unfortunately. https://developer.apple.com/documentation/swift/marking-api-availability-in-objective-c

daschuer: The error was "isPrimary() is not a member of ..." 

This is because you need the maximum SDK at build time to have all used definitions in the headers. 

Your assumption is correct for the runtime check of `@available`

Swiftb0y: Ah that makes sense. Thanks for the explanation. Isn't the `@available` check redundant in that case?

daschuer: You need both in case of a new SDK build executed on an old macOS. Than the `#if` is true and `@available` false.

Swiftb0y: Ah right, so one check is compile-time one is runtime...

Swiftb0y (approved):
LGTM

PR #11755, opened 2023-07-22
WTrackProperty: show simple editor on delayed double-click

ronso0:
show a simple QLineEdit after selected-click (delayed double click) onto track property labels

a coffee break project to help with my tag editing workflow (quick in-deck tag editing), i.e. create a fast alternative to double-click -> Track Properties -> click desired property line edit -> edit  -> hit Apply:

Gonna share this early, but Draft until it's polished.
track-label-edit-2

- [x] styles for line edit
- [x] ~~get real double-click interval, use x3 for selected click~~ 2s timeout are fine
- [x] use timer, reset 'selected' state automatically
- [x] visual feedback for 'selected' state (while widget expects second click):
new `selected` widget property, `WTrackProperty[selected="true"] { background: red; };`
- [x] add selected-click info to tooltips

May reuse #11752 for comments, but QLineEdit is fine for now.

PR #11767, opened 2023-07-26
Refocus library

ronso0:
Another little GUI helper:

When clicking into any  beatsize spinbox or selecting an effect/effect chain via the GUI selectors these widgets get keyboad focus. This interrupts the current library workflow (and keyboard shortcuts wouldn't work anymore) since you'd need to Tab, Esc or click to get back to the desired library widget.
Currently the focus is set to the Tracks table since that is the most likely place to be when DJing, but obviously not always.

With this PR the library widget that was focused previously is stored and can be foused again any time by any widget/object that has controlobject.h included via
```cpp
ControlObject::set(ConfigKey("[Library]", "refocus_prev_widget"), 1);
```
(any value except 0 will do, it's just triggering a valueChangeRequest)

Especially helpful for #11755 :)

### DIFF
--- a/src/library/librarycontrol.cpp
+++ b/src/library/librarycontrol.cpp
@@ -57,7 +57,8 @@ void LoadToGroupController::slotLoadToGroupAndPlay(double v) {
 LibraryControl::LibraryControl(Library* pLibrary)
         : QObject(pLibrary),
           m_pLibrary(pLibrary),
-          m_pFocusedWidget(FocusWidget::None),
+          m_focusedWidget(FocusWidget::None),
+          m_prevFocusedWidget(FocusWidget::None),
           m_pLibraryWidget(nullptr),
           m_pSidebarWidget(nullptr),
           m_pSearchbox(nullptr),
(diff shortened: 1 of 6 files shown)


src/library/librarycontrol.cpp
```diff
@@ -172,6 +173,16 @@ LibraryControl::LibraryControl(Library* pLibrary)
             });
 #endif
 
+    // Pure trigger control. Alternative for signal/slot since widgets that want
+    // to call refocusPrevLibraryWidget() are cumbersome to connect to.
+    // This CO is never actually set or read so the value just needs to be not 0
+    m_pRefocusPrevWidgetCO = std::make_unique<ControlPushButton>(
+            ConfigKey("[Library]", "refocus_prev_widget"));
```
daschuer, line 180: This should work when connecting to keyboard or controllers. Probably not a use case, but this would make it complete. 
```suggestion
            ConfigKey("[Library]", "refocus_prev_widget"));
    m_pRefocusPrevWidgetCO->setButtonMode(ControlPushButton::TRIGGER);
```

src/library/librarycontrol.cpp
```diff
@@ -172,6 +173,16 @@ LibraryControl::LibraryControl(Library* pLibrary)
...
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
```
daschuer, line 181: What is the reason to not connect it in Qt 6?

daschuer (changes requested):

src/library/librarycontrol.cpp
ronso0: QML. No legacy library widgets, so it's guarded like the other functions that work with these widgets.

src/library/librarycontrol.cpp
ronso0: Interesting! Will try.

ronso0: Jup, works. Thanks.

daschuer (approved):
LGTM, Thank you.

PR #11768, opened 2023-07-26
silence some 'unused parameter' warnings

ronso0:


src/controllers/hid/hidiooutputreport.cpp
```diff
@@ -31,6 +31,8 @@ void HidIoOutputReport::updateCachedData(const QByteArray& data,
         const RuntimeLoggingCategory& logOutput,
         HidIoGlobalOutputReportFifo* pGlobalOutputReportFifo,
         bool useNonSkippingFIFO) {
+    Q_UNUSED(deviceInfo);
+    Q_UNUSED(pGlobalOutputReportFifo);
```
JoergAtGithub, line 35: Please remove the unused parameters from this function, instead of silencing the warnings.

src/controllers/hid/hidiooutputreport.cpp
```diff
@@ -78,6 +80,7 @@ bool HidIoOutputReport::sendCachedData(QMutex* pHidDeviceAndPollMutex,
         hid_device* pHidDevice,
         const mixxx::hid::DeviceInfo& deviceInfo,
         const RuntimeLoggingCategory& logOutput) {
+    Q_UNUSED(deviceInfo);
```
JoergAtGithub, line 83: Please remove the unused parameter from this function, instead of silencing the warning.

src/musicbrainz/tagfetcher.cpp
```diff
@@ -351,6 +351,8 @@ void TagFetcher::slotCoverArtArchiveLinksTaskNetworkError(
         QNetworkReply::NetworkError errorCode,
         const QString& errorString,
         const mixxx::network::WebResponseWithContent& responseWithContent) {
+    Q_UNUSED(errorCode);
+    Q_UNUSED(errorString);
```
JoergAtGithub, line 355: There is an open TODO some lines below - I'm not sure if we should silence the warning in this case.

JoergAtGithub (changes requested):

src/controllers/hid/hidiooutputreport.cpp
ronso0: Tbh I didn't take a closer look if this are just base implementations that may be overriden.
If not, sure I can remove the arguments. (why are they there in the first place? no one cared about the warnings?)

JoergAtGithub: They are a leftover of a PR review. This information was used to print information in console messages. These message printing was removed during the review.
I didn't saw ans warnings. Shouldn't warnings be handled as errors?

ronso0: Alright, I removed the arguments.

> Shouldn't warnings be handled as errors?

Not that I'm aware of, I just assumed we try to get rid of warnings so they can easily be spotted in the build log when adding new code.

src/musicbrainz/tagfetcher.cpp
ronso0: I don't understand. We should keep the warning as a reminder for a TODO?

Holzhaus: I'd rather have an open GitHub issue for this instead of leaving warnings in the code to make it more likely to find this `TODO`.

Holzhaus (approved):
LGTM, thank you.

src/musicbrainz/tagfetcher.cpp
JoergAtGithub: At least there should be a comment, that describes that this silences a real warning, and not a false positive.

ronso0: I'll move the TODO up below Q_UNUSED

ronso0: and filed #11903

JoergAtGithub (approved):
LGTM! Waiting for CI

JoergAtGithub:
Thank you!

PR #11771, opened 2023-07-28
Handle setting loop_in and loop_out at the same position

### LINKED ISSUE #10600: loop_in & loop_out at same postion should not create a loop
Date: 2021-11-09T22:57:44Z
Status: Confirmed
Importance: Undecided
Launchpad Issue: [lp1950392](https://bugs.launchpad.net/bugs/1950392)
Tags: looping

---

I discovered this by chance:

* set loop_in
(don't change play position)
* set loop_out
= loop is created and activated
* hit Play
= waveform 'vibrates'

IMO when setting the loop_out in the same position Mixxx should refuse to create a loop.

    check for loop_in < loop_out here
    https://github.com/mixxxdj/mixxx/blob/441181a1338bc00be45cde8bf772bca844bbbd95/src/engine/controls/loopingcontrol.cpp#L844

ronso0:
* loop_in @ loop_out clears loop, sets loop_in
* loop_out @ loop_in is ignored

I also added two tests.

Closes #10600

### DIFF
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -583,7 +583,7 @@ void LoopingControl::setLoopInToCurrentPosition() {
     // Reset the loop out position if it is before the loop in so that loops
     // cannot be inverted.
     if (loopSamples.end != kNoTrigger &&
-            loopSamples.end < pos) {
+            loopSamples.end <= pos) {
         loopSamples.end = kNoTrigger;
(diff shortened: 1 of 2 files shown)


Swiftb0y (approved):
I find it a little weird that the behavior is not "symmetrical" but if that's easiest to implement then lets go for it.

daschuer (approved):
I have just tested it and it works good. IMHO the different behavior is desired because we do not allow to set a loop-out before the loop-in. Hitting exactly loop-in is not special in that regards.

PR #11775, opened 2023-07-28
Broadcast: Improve error message in case of timeout 

daschuer:
In case of timeout was the error string "No error" which is confusing. 
See: https://github.com/mixxxdj/mixxx/issues/11740#issuecomment-1650105772
This PR replaces the string with "Connection establishment time-out"

This fits IMHO to other error messages like from libshout: 
""TLS connection can not be established because of bad certificate"


Swiftb0y (approved):
LGTM.

PR #11781, opened 2023-08-02
fix(fx): don't reset fxunit "super" and "mix" knob on startup

### LINKED ISSUE #11773: FX dry/wet value on startup
### Bug Description

Hi,
it seems v2.4 always resets dry/wet FX value at startup.

I'm not 100% sure but I think 2.3.x used to store this value and apply it again at startup.

Thanks !

### Version

2.4

### OS

DietPiOS (debian bullseye based)

Swiftb0y:
fixes #11773

I assume this was an oversight from the original PR?


daschuer (approved):
LGTM. That was definitely wrong.

Swiftb0y:
Thank you

PR #11441, opened 2023-04-04
Fix Serato Metadata checkbox behaviour

tech-bash:
Fix: #11226

Its working perfectly as wanted.

gsoc2

PR #11782, opened 2023-08-02
Revive #11441

### LINKED ISSUE #11226: Write Serato Metadata checkbox is ignored when writing track metadata is disabled
### Bug Description

I think it would be reasonable to disable and gray out the checkbox in this case. 

### Version

2.3.3

### OS

_No response_

Swiftb0y:
Fixes #11226, Supersedes #11441

I added another little feature where it disables the serato metadata export when the general metadata export is also disabled. I'm not sure if enforcing that in the UI is the best solution, suggestions welcome.

### DIFF
--- a/src/preferences/dialog/dlgpreflibrary.cpp
+++ b/src/preferences/dialog/dlgpreflibrary.cpp
@@ -195,6 +195,7 @@ void DlgPrefLibrary::slotUpdate() {
             ConfigKey("[Library]","SyncTrackMetadataExport"), false));
     checkBox_SeratoMetadataExport->setChecked(m_pConfig->getValue(
             ConfigKey("[Library]", "SeratoMetadataExport"), false));
+    setSeratoMetadataEnabled(checkBox_SyncTrackMetadataExport->isChecked());
     checkBox_use_relative_path->setChecked(m_pConfig->getValue(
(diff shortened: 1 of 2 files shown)


Swiftb0y:
> In addition, this PR resets the value of "SeratoMetadataExport". So that after cycling "SyncTrackMetadataExport", "SeratoMetadataExport" is off.

Right, I wanted to show that disabling metadata export will disable serato export as well. Just making the checkbox unclickable was not enough of an indication for me since it still looked like it was enabled, just not changeable by the user.

> Bug or feature?

Not sure either, I'd consider this "minor visual polishing". I don't care what branch it goes into, I just wanted to close the abandoned PR this is based on. What branch would you like this to be targeted at?

daschuer:
I would prefer to merge this to 2.3. if this is not to much hassle for you. 

Just graying out the checkbox does not work. It need to be grayed out and unchecked. 
The question is if we want to restore the old state after enabling metadata export. 
I think we have the option "checked"/"unchecked"/"old state". 

From UX perspective, it should be the "old state" to decouple dependencies, but if we have good reasons for a difference choice it is also fine.

Swiftb0y:
done, but expect merge conflicts when merging to 2.4.

daschuer (approved):
LGTM, Thank you.

Swiftb0y:
Thank you

Issue #11778, opened 2023-08-01
Could Not Create audio file for recording


DJSlaz:
### Bug Description

<img width="488" alt="Screenshot 2023-07-31 at 21 13 37" src="https://github.com/mixxxdj/mixxx/assets/141093017/83c4b09e-0125-4995-a0f6-afce5b6253f5">

I was attempting to record a show while broadcasting this evening.   When I hit record, I received this error.  I shut down and restarted the application and still received this error. This makes ZERO sense. I have 750 gig free on my local HD.  This only started to happen after the last MacOS security updates about a week ago. 

Even if I change the directory to point to a new location I still receive this error.  

### Version

2.3.5

### OS

Mac Os Ventura 13.5

daschuer:
What is the new location? 

Does mixxx.log contain any information? 
https://github.com/mixxxdj/mixxx/wiki/Finding-the-mixxx.log-file

DJSlaz:
Here are additional log files.  This morning, I created a new folder in a new location and tried to record, but got the same error.  I attached all log files.  The log files do not say much - I found this in <debug> section: 

Debug [Main]: Failed to create folder "/Users/russlazarus/Documents/Gutsy/EarRational/NewRecordedshowsAug21" for recordings
Debug [Main]: Recordings folder set to "/Users/russlazarus/Documents/Gutsy/EarRational/NewRecordedshowsAug21”

[mixxx.log](https://github.com/mixxxdj/mixxx/files/12241180/mixxx.log)

Attached are the relevant log files. 

[mixxx.log2August.txt](https://github.com/mixxxdj/mixxx/files/12241267/mixxx.log2August.txt)
[mixxx.log.1_2August.txt](https://github.com/mixxxdj/mixxx/files/12241268/mixxx.log.1_2August.txt)
[mixxx.log.2_31July 2123.txt](https://github.com/mixxxdj/mixxx/files/12241269/mixxx.log.2_31July.2123.txt)
[mixxx.log.3_31July 2110.txt](https://github.com/mixxxdj/mixxx/files/12241270/mixxx.log.3_31July.2110.txt)
[mixxx.log.4_31July 2104.txt](https://github.com/mixxxdj/mixxx/files/12241271/mixxx.log.4_31July.2104.txt)
[mixxx.log.5_31July 1900.txt](https://github.com/mixxxdj/mixxx/files/12241272/mixxx.log.5_31July.1900.txt)

DJSlaz:
[...]

PR #11783, opened 2023-08-02
Improve debug output for some file system operations

daschuer:
This aims to help finding the root cause of issue https://github.com/mixxxdj/mixxx/issues/11778

### DIFF
--- a/src/library/parsercsv.cpp
+++ b/src/library/parsercsv.cpp
@@ -5,6 +5,7 @@
 #include <QTextStream>
 #include <QtDebug>
 
+#include "errordialoghandler.h"
 #include "moc_parsercsv.cpp"
 
 namespace {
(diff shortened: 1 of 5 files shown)


src/library/parsercsv.cpp
```diff
@@ -129,7 +133,7 @@ bool ParserCsv::writeCSVFile(const QString &file_str, BaseSqlTableModel* pPlayli
     if (!file.open(QIODevice::WriteOnly | QIODevice::Text)) {
         QMessageBox::warning(nullptr,
                 tr("Playlist Export Failed"),
-                tr("Could not create file") + " " + file_str);
+                tr("Could not create file") + " " + file_str + "\n" + file.errorString());
```
Holzhaus, line 136: This is possibly not localized and may look weird to inexperienced users. I think we should use the property-based API here and use the [detailed text](https://doc.qt.io/qt-5/qmessagebox.html#detailedText-prop) for this.

src/recording/recordingmanager.cpp
```diff
@@ -189,11 +189,14 @@ void RecordingManager::setRecordingDir() {
         ConfigKey(RECORDING_PREF_KEY, "Directory")));
     // Note: the default ConfigKey for recordDir is set in DlgPrefRecord::DlgPrefRecord.
 
-    if (!recordDir.exists()) {
-        if (recordDir.mkpath(recordDir.absolutePath())) {
+    if (recordDir.exists()) {
```
Holzhaus, line 192: where did the `!` go?

src/recording/recordingmanager.cpp
```diff
@@ -189,11 +189,14 @@ void RecordingManager::setRecordingDir() {
...
+        if (!recordDir.mkpath(recordDir.absolutePath())) {
```
Holzhaus, line 193: I think the exclamation mark is wrong here.

daschuer:
Done.

Holzhaus (approved):
LGTM, thank you.

PR #11792, opened 2023-08-06
Denon MC6000MK2: Delete mapping for master gain

uklotzde:
The master knob also controls the hardware and must not be mapped.

No one ever noticed :shrug: I forgot about this fix in my patched version that became necessary due to a hardware defect.


daschuer (approved):
LGTM, thank you.

PR #11793, opened 2023-08-06
translations: Remove redundant `cmn` locale in favor of `zh`

### LINKED ISSUE #11791: Unsupported locale "cmn"
### Bug Description

I see this in the terminal every time I start Mixxx:

    warning [Main] Unsupported locale "cmn"

It looks like we use ISO 639-1 language codes (some with ISO 3166-1 alpha-2 country codes) in `res/translations`, except for `cmn` which is an ISO 639-3 language code for Mandarin Chinese. Not sure if this is redundant with `zh`/`zh_CN`/`zh_HK`/`zh_TW` which are all codes for Chinese (but may be another chinese dialect).

In any case, the error message indicates that the ISO 639-3 name might be incompatible with the way we handle locales at the moment (or the error message is wrong).

### Version

4a51c393eb8820a5728f4b2b53dce8ef203d2b8c

### OS

Arch Linux

Holzhaus:
Resolves #11791.


Holzhaus:
This targets `main`, let me know if I should rebase on `2.3` or `2.4`.

Swiftb0y:
Does this also need administrative action on transifex so it doesn't get readded on the next translations pull? @daschuer

Holzhaus:
> Does this also need administrative action on transifex so it doesn't get readded on the next translations pull? @daschuer 

Probably. I don't think I have the permissions to do that and I wanted to discuss this solution here first anyway.

daschuer:
Thank you. Please rebase this on 2.4. The 2.3 branch has no cmn language.

daschuer (approved):
LGTM. I have also removed cnm form Transifex. 
Thank you.

PR #11794, opened 2023-08-06
iTunes: Permit duplicate playlist names by identifying playlists by id (rather than name)

fwcd:
This patches the iTunes feature to identify playlists by id rather than their name and updates the created `TreeItem`s to also use the original, non-suffixed names (without `... #n`). Given that the corresponding database column has a `UNIQUE` constraint, the suffixed name will still be used there for backwards compatibility/to avoid a schema change, but no longer presented in the UI.

Thus if the user e.g. has two playlists named `A`, they will now both be presented as `A` in the sidebar, rather than `A` and `A #2`, which generally looks cleaner since it mirrors the "original" naming of the playlists.

For comparison:

| Before | After |
| - | - |
| Screenshot 2023-08-06 at 23 53 08 | Screenshot 2023-08-06 at 23 51 33 |

### DIFF
--- a/src/library/baseexternalplaylistmodel.cpp
+++ b/src/library/baseexternalplaylistmodel.cpp
@@ -111,6 +111,10 @@ void BaseExternalPlaylistModel::setPlaylist(const QString& playlist_path) {
         return;
     }
 
+    setPlaylistById(playlistId);
+}
+
+void BaseExternalPlaylistModel::setPlaylistById(int playlistId) {
     // Store search text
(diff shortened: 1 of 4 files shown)


fwcd:
Side note: I am targeting 2.4 for this since it's a rather minor addition to the "new" iTunes feature in 2.4, feel free to comment on whether that's fine or if I should retarget to `main`.

daschuer (approved):
LGTM, Thank you for this nice addition.

PR #11790, opened 2023-08-05
Prefer mold or lld as linker to speed up linking

daschuer:
This uses mold or lld as linker if available. 

It also installs it on our Ubuntu CI. The release builds on Launchpad are still on ld. 
So the 2.4 merge target should be fine. 

We may also consider to use lld for macOS, but that would effect our release builds. 
Not sure about the consequences. It seems to be used by Google for Chrome. 
`mold` is not available on macOS requires the commercial clone `sold` 

Discussed here: 
https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/Reducing.20linking.20times

PR #11796, opened 2023-08-07
cmake: use -B option to select mold linker for GCC < 12.1.0

ronso0:
Follow-up for #11790 to avoid
```c++
[100%] Linking CXX executable mixxx
c++: error: unrecognized command-line option ‘-fuse-ld=mold’; did you mean ‘-fuse-ld=gold’?
```

Disregard this commit if all officially suported build platforms are expected to provide (if configured) GCC 12.1.0+

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -547,7 +547,25 @@ if(NOT MSVC)
   if(MOLD_SUPPORT)
     if(MOLD_EXECUTABLE)
       message(STATUS "Selecting mold as linker")
-      add_link_options("-fuse-ld=mold")
+      set(SELECT_MOLD_WITH_FUSE true)
+      # With GCC, mold is a valid '-fuse' argument only with version 12.1.0 and newer.
+      # Earlier versions must use the -B option.
+      if (GNU_GCC)
+        execute_process(
+          COMMAND gcc -dumpfullversion
+          ERROR_QUIET
+          OUTPUT_VARIABLE GCC_VERSION_STRING
+          OUTPUT_STRIP_TRAILING_WHITESPACE
+        )
+        if (GCC_VERSION_STRING VERSION_LESS 12.1.0)
+          set(SELECT_MOLD_WITH_FUSE false)
+        endif()
+      endif()
+      if (SELECT_MOLD_WITH_FUSE)
+        add_link_options("-fuse-ld=mold")
+      else()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -547,7 +547,25 @@ if(NOT MSVC)
   if(MOLD_SUPPORT)
     if(MOLD_EXECUTABLE)
       message(STATUS "Selecting mold as linker")
-      add_link_options("-fuse-ld=mold")
+      set(SELECT_MOLD_WITH_FUSE true)
```
ronso0, line 550: I'm not a cmake expert, maybe there's a smarter way to achieve the -fuse/-B switch without this bool.

daschuer:
Yes, that can be improved, but is good for now. I will file a follow up. 
Thank you for taking care.

daschuer (approved):

daschuer:
Interestingly our CI does not fail with `-fuse-ld=mold` 
Because the Ubuntu version supports it from gcc 11.2:
http://changelogs.ubuntu.com/changelogs/pool/main/g/gcc-11/gcc-11_11.4.0-1ubuntu1~22.04/changelog

I think we should query the mold version as we do with lid and use that as indicator.

daschuer:
Oh, I have just realized this does not work, because it puts 
-B/usr/bin/mold instead of the folder with the symlink named ld to mold. Is this symlink part of the dev package? 
`usr/libexec/mold/ld`

ronso0:
This symlink is just an addition to the executable installed to the default/custom bin dir, which should ofc be in $PATH which cmake uses to lookup programs.
https://github.com/rui314/mold/blob/f631477a681d92f721b45ed049c9731f951546d2/CMakeLists.txt#L461-L464

ronso0:
> ```c++
> [100%] Linking CXX executable mixxx
> c++: error: unrecognized command-line option ‘-fuse-ld=mold’; did you mean ‘-fuse-ld=gold’?
> ```

FWIW I didn't see that warning when I configured manually with `-fuse-ld=mold`, but it popped up during the linking phase when I _built_  the first time after  pulling that change.

daschuer:
The issue with  default_option() is that the default is kept tin the CMakeCache.txt and not evaluated again. So you need to clear the MOLD_SUPPORT setting from the file after any change regarding it.

PR #11800, opened 2023-08-09
Improve mold linker configurations for some corner cases

daschuer:
This checks now if gcc is patched for mold like in Ubuntu Jammy and finds the mold symlink in various places for the -B option as fall back.


ronso0:
@daschuer I'm still having trouble with the cmake logic, I'm not the right one to review this atm.
@JoergAtGithub or @Holzhaus maybe?

CMakeLists.txt
```diff
@@ -542,58 +542,59 @@ else()
 endif()
 
 if(NOT MSVC)
-  find_program(MOLD_EXECUTABLE mold)
-  default_option(MOLD_SUPPORT "Use 'mold' for linking" "MOLD_EXECUTABLE")
+  # Try to configure mold as linker via -fuse-ld=mold, support from gcc 12.1, gcc 11.2.0-16 or clang
+  execute_process(
```
Holzhaus, line 546: Maybe use https://cmake.org/cmake/help/latest/module/CheckCompilerFlag.html?

CMakeLists.txt
```diff
@@ -542,58 +542,59 @@ else()
...
+  if(NOT MOLD_FUSE_VERSION_STRING)
+    # check if the symlink ld is in the mold folder for older compiler
+    find_program(MOLD_SYMLINK "${CMAKE_INSTALL_LIBEXECDIR}/mold/ld")
+    get_filename_component(MOLD_SYMLINK_DIRECTORY ${MOLD_SYMLINK} DIRECTORY)
+    set(CMAKE_FIND_DEBUG_MODE FALSE)
+  endif()
+  default_option(MOLD_SUPPORT "Use 'mold' for linking" "MOLD_FUSE_VERSION_STRING OR MOLD_SYMLINK")
   if(MOLD_SUPPORT)
-    if(MOLD_EXECUTABLE)
+    if(MOLD_FUSE_VERSION_STRING)
```
Holzhaus, line 560: I recommend to use a name that communicates what you are interested in. Something like `MOLD_FUSE_FOUND` maybe?

Holzhaus (commented):
I can check the cmake code, but I have no idea about the different variants of selection the proper linker. I just have `mold` executable in `/usr/bin` and that's it.

CMakeLists.txt
daschuer: The advantage of querying the version via the compiler is that the whole chain is checked and not only the compiler flag. 
check_compiler_flag() does only check the flag.

daschuer:
Done

daschuer:
@Holzhaus Can this be merged now?

daschuer:
ping @Holzhaus

daschuer:
This is ready for merge IMHO

Holzhaus:
Sorry for the long delay, but I intentionally used the comment feature instead of "Request Changes" to indicate that my comments are not a blocker and someone else may merge without a second review loop by me. 

I currently cannot build Mixxx due to #12505. The issue also occurs with this branch. Cmake configuration still works so it's likely okay, but someone else who is able to actually build Mixxx should confirm and merge.

ronso0:
Alright, thanks.
I'll try to find time to remove my mold symlinks in order to test this.

ronso0:
I can built and verify mold was used with
/usr/local/bin/ld.mold symlink mold
`MOLD_SUPPORT` is set TRUE
I renamed the mold binary and updated the symlink, so I assume cmake is using ld.mold

Anything else I should test?

daschuer:
No, thank you for testing. I think we can just merge it and will notice quickly if something falls apart.

PR #11804, opened 2023-08-10
(fix) WTrackProperty: focus double-clicked property field in DlgTrackInfo

### LINKED ISSUE #11802: Double Click on the track title label in decks crashes Mixxx
### Bug Description

Double click on the title should bring the properties window - the window appears, but stays black and after 2 seconds Mixxx crashes. Hope fast fix because i use Mixxx on a wedding party on friday. 

### Version

2.4 beta 79

### OS

Linux Mint 21.2 Cinnamon

ronso0:
Embarrassing that I didn't test the title label and assumed QHash::end() would return QWidget nullptr.

Closes #11802

### DIFF
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -67,6 +67,7 @@ void DlgTrackInfo::init() {
     // Associate with property strings taken from library/dao/trackdao.h
     m_propertyWidgets.insert("artist", txtArtist);
     m_propertyWidgets.insert("title", txtTrackName);
+    m_propertyWidgets.insert("titleInfo", txtTrackName);
     m_propertyWidgets.insert("album", txtAlbum);
     m_propertyWidgets.insert("album_artist", txtAlbumArtist);
     m_propertyWidgets.insert("composer", txtComposer);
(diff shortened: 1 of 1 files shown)


src/library/dlgtrackinfo.cpp
```diff
@@ -504,9 +505,9 @@ void DlgTrackInfo::focusField(const QString& property) {
     if (property.isEmpty()) {
         return;
     }
-    QWidget* w = m_propertyWidgets.find(property).value();
-    if (w) {
-        w->setFocus();
+    auto it = m_propertyWidgets.constFind(property);
+    if (it == m_propertyWidgets.constEnd()) {
```
Swiftb0y, line 509: major typo flaw. Did you forget to test this?
```suggestion
    if (it != m_propertyWidgets.constEnd()) {
```

Swiftb0y (changes requested):

src/library/dlgtrackinfo.cpp
ronso0: I reverted early return just before pushing without testing again..
fixed.

Swiftb0y (approved):
LGTM

stefan-franz:
Hmmm.....i'm right that it is fixed? (because of closed the the issues...?) I'm a bit confused.....
If it is fixed, there is still no update available for Linux Mint (over ppa)

ronso0:
I thought you're around for long and know that a fix lands in the build after the respective PR was merged.
"fixed by #[PR number]" means the PR adds the code to fix a certain issue. Until that is merged, built, and the build being pushed to the ppa you can test with the CI builds as explained here https://github.com/mixxxdj/mixxx/wiki/Testing

Building takes a while, just check the ppa in an hour or so.

stefan-franz:
Thanks - i didn't know that. But now it works. Update was found by Linux Mint and all works. 
Thank you very much to all involved for that fast fix.

Issue #11576, opened 2023-05-22
GNOME 44 (Wayland): Context menu on tree items in side bar closes immediately


uklotzde:
### Bug Description

Right clicking on any tree item in the side bar has no effect. **Update:** It actually appears for a split second, but disappears immediately without further interaction.

Opening the context menu with a right click works for tracks in the table view.

Qt 5.15.9 on a single screen.

### Version

2.3.5, 2.4, main

### OS

Fedora 38 / GNOME 44 on Wayland

### Workaround

Start _GNOME on Xorg_ instead of using Wayland as the display server. This bug only seems to affect the XWayland backend.

uklotzde:
The right click itself is received as the following debug log (custom code) proves:

```
debug [Main] aoide LibraryFeature - onRightClickChild QModelIndex(0,0,0x226e7250,SidebarModel(0x470fff0))
```

Swiftb0y:
I can partly confirm the issue. The context menu does open sometimes if you're lucky.

ronso0:
One difference between [`WLibrarySidebar::contextMenuEvent`](https://github.com/mixxxdj/mixxx/blob/7f79194df0a53987816b4e47e34067e3eb12361d/src/widget/wlibrarysidebar.cpp#L35-L45) (doesn't work) and [`WTrackTableView::contextMenuEvent`](https://github.com/mixxxdj/mixxx/blob/7f79194df0a53987816b4e47e34067e3eb12361d/src/widget/wtracktableview.cpp#L500-L514) (works)is that in the latter the event is accepted.
Can you test if accpeting the event in the sidebar makes a difference?

uklotzde:
> One difference between [`WLibrarySidebar::contextMenuEvent`](https://github.com/mixxxdj/mixxx/blob/7f79194df0a53987816b4e47e34067e3eb12361d/src/widget/wlibrarysidebar.cpp#L35-L45) (doesn't work) and [`WTrackTableView::contextMenuEvent`](https://github.com/mixxxdj/mixxx/blob/7f79194df0a53987816b4e47e34067e3eb12361d/src/widget/wtracktableview.cpp#L500-L514) (works)is that in the latter the event is accepted. Can you test if accpeting the event in the sidebar makes a difference?

Unfortunately, that doesn't make a difference :disappointed: But the accept() calls are nevertheless missing, also for the column headers that seem to react only when clicking twice sometimes.

ronso0:
[...]

PR #11805, opened 2023-08-10
Delete obsolete -platform xcb args

uklotzde:
Enforcing Mixxx to use XWayland on Wayland now causes bugs like #11576.

I also don't experience any issue with the legacy GL/GLSL waveform shaders.


uklotzde:
Otherwise I would need to manually patch this out.

Swiftb0y (approved):
Thank you. I forgot we decided to do this.

Swiftb0y:
just fyi there is a small visual glitch with the (GL) waveform variants. My guess is that some blend setting or similar is wrong resulting in out-of-range values because the resulting waveforms look blown out onscreen but the screenshot produced by gnome looks entirely different. The difference probably comes from the GPU just clamping the values while the compositor (or whatever is responsible for screenshots) just ignores the integer part. 
RGB (GL) waveforms seemingly wrapped

In either case, nothing blocking this PR but we'll probably get reports of people being on the GL waveforms from 2.3 even though they should switch to the (all-shaders) anyways. I guess would be a valid reason to remove the GL waveforms altogether?

daschuer (approved):
LGTM, Thank you.

PR #11807, opened 2023-08-11
Refactor timers

Swiftb0y:
Initially wanted to investigate something performance-related but then got distracted by the questionable state of `util/performancetimer.h` and `util/timer.h` so I decided to clean it up a little.


src/util/battery/battery.cpp
```diff
@@ -16,16 +16,16 @@
 #include "util/math.h"
 
 // interval (in ms) of the timer which calls update()
-constexpr int kiUpdateInterval = 5000;
+constexpr int kBatteryUpdateIntervalMS = 5000;
```
Holzhaus, line 19:
```suggestion
constexpr int kBatteryUpdateIntervalMillis = 5000;
```
That is the usual suffix in the codebase.

src/util/performancetimer.h
```diff
@@ -1,75 +1,38 @@
...
-**
-****************************************************************************/
-
 #pragma once
+#include <chrono>
 
-//
-// This is a fork of QPerformanceTimer just without the Q prefix
-// To fix interface changes issues in different QT versions
-// Added restart() function.
```
Holzhaus, line 47: Have you checked if this class is still necessary? Maybe we can use plain `QPerformanceTimer` nowadays.

Swiftb0y: QPerformanceTimer doesn't exist anymore as far as I can tell. Closest thing would be QElapsedTimer. Didn't investigate whether that satisfies the same constraints. I will check.

src/util/timer.h
```diff
@@ -33,103 +35,58 @@ class Timer {
...
+            : m_state(std::nullopt) {
+        if (!CmdlineArgs::Instance().getDeveloper()) {
+            return;
         }
+#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
+        QString strKey = arg.isEmpty() ? key.toString() : key.arg(arg);
+#else
+        QString strKey = arg.isEmpty() ? key.toString() : key.toString().arg(arg);
+#endif
+        m_state = std::make_optional<TimerData>(TimerData{Timer(strKey, compute), false});
```
daschuer, line 61:
```suggestion
        m_state = std::make_optional<TimerData>(TimerData{Timer(std::move(strKey), compute), false});
```

src/util/timer.h
```diff
@@ -33,103 +35,58 @@ class Timer {
...
+    ScopedTimer& operator=(const ScopedTimer&) = delete;
+
+    ScopedTimer(ScopedTimer&&) = default;
+    ScopedTimer& operator=(ScopedTimer&&) = default;
 
     void cancel() {
-        m_cancel = true;
+        if (m_state) {
+            m_state->cancelled = true;
+        }
```
daschuer, line 82: The cancelled state can be replaced by m_state == std::nullopt;
 ```suggestion
        m_state.reset();
```
Because the timers is no longer required and can be destructed.

File: src/util/battery/battery.cpp (hunk 1 above)
Comment (JoergAtGithub) [reply src/util/battery/battery.cpp:19]: Why not using the proper type `std::chrono::milliseconds` instead of adding this information to the name?

File: src/util/performancetimer.h (hunk 2 above)
Comment (daschuer) [reply src/util/performancetimer.h:47]: QPerformanceTimer  was forked from QElapsedTimer and later reintegrated: 
https://github.com/qt/qt/commit/88966515993773e4d7d6276bce6fccfcc13ab6f4
I think we can adopt that commit to our code and drop our PerformanceTimer entirely. 
I recommend that over the std::chrono::steady_clock solution because it is only weak defined by the C++ standard and may change by any update or different standard library.

**Review by daschuer — CHANGES_REQUESTED** [review]

File: src/util/timer.h (hunk 4 above)
Comment (Swiftb0y) [reply src/util/timer.h:82]: I'm not super familiar with C++ destructors but I thought this would still happen automatically for RAII types even when using a user-provided constructor. Or is there a way to call a default-generated destructor within the user-provided destructor?

File: src/util/battery/battery.cpp (hunk 1 above)
Comment (Swiftb0y) [reply src/util/battery/battery.cpp:19]: Thank you, much better solution

File: src/util/performancetimer.h (hunk 2 above)
Comment (Swiftb0y) [reply src/util/performancetimer.h:47]: Unfortunately the API differs a little bit because qt has no duration type. `PerformanceTimer` is also spread over more than 40 classes in our codebase. IMO we should keep PerformanceTimer as a facade over QElapsedTimer that makes use of `mixxx::Duration` (which we should also replace with std::chrono::duration` IMO but thats even harder than replacing PerformanceTimer verbatim).

Comment (JoergAtGithub) [reply src/util/performancetimer.h:47]: Should we add a task to our Semantic Type Refactoring project https://github.com/mixxxdj/mixxx/projects/5 ?

File: src/util/timer.h (hunk 4 above)
Comment (Swiftb0y) [reply src/util/timer.h:82]: > The cancelled state can be replaced by m_state == std::nullopt;

That implies the removal of the `cancel` method, but its only used in one location and thats before a return, so basically useless anyway.

File: src/util/timer.h [hunk 5]
```diff
@@ -58,16 +65,15 @@ class ScopedTimer {
...
     ~ScopedTimer() noexcept {
-        if (m_state) {
-            if (!m_state->cancelled) {
-                m_state->timer.elapsed(true);
-            }
+        if (m_maybeTimer) {
+            m_maybeTimer->elapsed(true);
         }
+        m_maybeTimer.reset();
     }
```
Swiftb0y, line 77: The reset triggers a `-Wmaybe-uninitialized` warning locally for me and I'm not sure why...

```cpp

In destructor ‘QString::~QString()’,
    inlined from ‘Timer::~Timer()’ at /home/swiftb0y/Sourcerepositories/mixxx/src/util/timer.h:20:7,
    inlined from ‘constexpr void std::_Optional_payload_base<_Tp>::_M_destroy() [with _Tp = Timer]’ at /usr/include/c++/13/optional:287:35,
    inlined from ‘constexpr void std::_Optional_payload_base<_Tp>::_M_reset() [with _Tp = Timer]’ at /usr/include/c++/13/optional:318:14,
    inlined from ‘constexpr std::_Optional_payload<_Tp, false, _Copy, _Move>::~_Optional_payload() [with _Tp = Timer; bool _Copy = false; bool _Move = false]’ at /usr/include/c++/13/optional:439:65,
    inlined from ‘constexpr std::_Optional_base<Timer, false, false>::~_Optional_base()’ at /usr/include/c++/13/optional:510:12,
    inlined from ‘constexpr std::optional<Timer>::~optional()’ at /usr/include/c++/13/optional:705:11,
    inlined from ‘ScopedTimer::~ScopedTimer()’ at /home/swiftb0y/Sourcerepositories/mixxx/src/util/timer.h:71:5,
    inlined from ‘virtual void VinylControlXwax::analyzeSamples(CSAMPLE*, size_t)’ at /home/swiftb0y/Sourcerepositories/mixxx/src/vinylcontrol/vinylcontrolxwax.cpp:648:1:
/usr/include/qt5/QtCore/qstring.h:1311:35: warning: ‘*(QString*)((char*)&t + offsetof(ScopedTimer, ScopedTimer::m_maybeTimer.std::optional<Timer>::<unnamed>.std::_Optional_base<Timer, false, false>::<unnamed>)).QString::d’ may be used uninitialized [-Wmaybe-uninitialized]
 1311 | inline QString::~QString() { if (!d->ref.deref()) Data::deallocate(d); }

```

src/util/performancetimer.h
Swiftb0y: Maybe. We should at least think about replacing the `mixxx::Duration` internals with `std::chrono::duration` to allow for gradual transition. But `mixxx::Duration` is already a semantic type, so I don't see a pressing need. The only major advantage `std::chrono::duration` has is its `duration_cast` as far as I can tell.

src/util/timer.h
```diff
@@ -33,103 +35,58 @@ class Timer {
...
-
-    ScopedTimer(const char* key, const char *arg = NULL,
-                Stat::ComputeFlags compute = kDefaultComputeFlags)
-            : m_pTimer(NULL),
-              m_cancel(false) {
-        if (CmdlineArgs::Instance().getDeveloper()) {
-            initialize(QString(key), arg ? QString(arg) : QString(), compute);
-        }
+    ScopedTimer(QStringView key, int i, Stat::ComputeFlags compute = kDefaultComputeFlags)
+            : ScopedTimer(key, QString::number(i), compute) {
```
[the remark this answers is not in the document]
Swiftb0y: fixed, now results in a duplicate `getDeveloper()` call, but I'm pretty sure that gets optimized away.

src/sources/soundsourcemodplug.cpp
```diff
@@ -133,8 +133,6 @@ SoundSource::OpenResult SoundSourceModPlug::tryOpen(
     m_pModFile = ModPlug::ModPlug_Load(m_fileBuf.constData(),
             m_fileBuf.length());
     if (m_pModFile == nullptr) {
-        // an error occurred
-        t.cancel();
```
daschuer, line 137: This was used to not have a log entry of the "SoundSourceModPlug::open()" timer. 
Was it intended to have now always a timer entry? 

My original suggestion was to keep cancel() but reset the optional during cancel. This has the same effect.

Swiftb0y: Ah yeah, I misunderstood the code. I'll revert.

Swiftb0y:
The QElapsedTimer::clockType() == PerformanceTimer debug assert fails on some platforms... How do we go about that? relax the restrictions and just insist its monotonic?

JoergAtGithub:
According to the documentation QElapsedTimer::PerformanceCounter is a Windows implementation:
 https://doc.qt.io/qt-5/qelapsedtimer.html#ClockType-enum

src/util/timer.h
```diff
@@ -33,103 +35,59 @@ class Timer {
...
-        if (m_pTimer) {
-            if (!m_cancel) {
-                m_pTimer->elapsed(true);
-            }
-            m_pTimer->~Timer();
+    ~ScopedTimer() noexcept {
+        if (m_maybeTimer) {
+            m_maybeTimer->elapsed(true);
         }
+        m_maybeTimer.reset();
```
daschuer, line 76: This will happen automatically.

Swiftb0y: Are you sure? Thats what I was asking in this comment: https://github.com/mixxxdj/mixxx/pull/11807#discussion_r1292249067

src/util/timer.h
daschuer: The Uninitialized QString seems to be Timer::m_key. The warning happens after the user destructor in the auto generated code. 
It is uninitialized if m_maybeTimer is std::nullopt. But in that case the ~Timer() not be run.

src/util/timer.h
daschuer: Yes, the back trace with the warning [pr](https://github.com/mixxxdj/mixxx/pull/11807#discussion_r1292351607) proves it. 
But there seems to be a related issue causing the warnig.

daschuer: I cannot confirm that warning with valgrind, It seems to be a false positive.

daschuer:
QElapsedTimer::clockType() seems to be not a nop() in some cases. Can we move the check to an extra function and call it only once? 

In my case "QElapsedTimer::ClockType::MonotonicClock" is returned. QElapsedTimer::ClockType::PerformanceCounter" is nerver returned.

JoergAtGithub:
Checking clock time make not much sense. The documentation says: "QElapsedTimer will always use the same clock type in a particular machine, so this value will not change". It just tells you which implementation is used on the particular system (result will differ on operating system).
What makes sense to check in an assertion is https://doc.qt.io/qt-5/qelapsedtimer.html#isMonotonic Also this needs to be checked only once on a particular machine.

daschuer:
I agree. Since Qt uses the timer internternally for the same propose we, may rely on Qt that they will use the optimal timer. 

Since system calls are involve, a subject to change with OS versions, we may even consider to check the monotonic nature it in release builds. So the check is in place for every "particular machine".

JoergAtGithub:
On any modern desktop computer with any OS officially supported by Mixxx, you will get a monotonic clock.
I guess thiss is more relevant for the embedded and mobile platforms, which Qt also supports.

src/util/performancetimer.h
```diff
@@ -1,75 +1,43 @@
...
+class PerformanceTimer {
+  public:
+    PerformanceTimer()
+            : m_elapsedTimer() {
+    }
+
+    // call this once at startup to ensure that the QElapsedTimer is monotonic
+    static void debugEnsureClockIsMonotonic() {
+        // TODO: turn this into a static_assert and inline this method
+        // once Qt enables it.
```
daschuer, line 15: This will probably never be a static assertion, because it depends on the OS executing Mixxx. 
```suggestion
```

daschuer (changes requested):
Just a final nit pick and than LGTM

src/util/performancetimer.h
Swiftb0y: yes, but the OS / target is usually known at compile time anyways. That's why the stdlib can get a away with a `constexpr static is_steady`. https://en.cppreference.com/w/cpp/chrono/system_clock

Swiftb0y:
I'm unsure whether the clock is monotonic in all cases. I experienced some test failures. https://github.com/Swiftb0y/mixxx/actions/runs/5913234283/job/16037422961

daschuer (approved):
LGTM, thank you.

Swiftb0y:
you just merged dozens of fixup commits.... how do we proceed?

uklotzde:
> you just merged dozens of fixup commits.... how do we proceed?

@Swiftb0y Switch the repo to "Allow squash merging" exclusively for preventing this in the future? Avoids tons of low-value commits in the repo and ensures that `git bisect` works for every commit.

I have seen this strategy in a customer project recently and adopted it immediately after experiencing the benefits. Saves you from a lot of headaches.

PR #11815, opened 2023-08-12
Fix calc of first beatgrid bar in Engine export

mr-smidge:


src/library/export/engineprimeexportjob.cpp
```diff
@@ -143,7 +143,7 @@ bool tryGetBeatgrid(BeatsPointer pBeats,
 
     int numBeatsToCue = pBeats->numBeatsInRange(firstBeatPlayPos, cueBeatPlayPos);
     const auto firstBarAlignedBeatPlayPos = pBeats->findNBeatsFromPosition(
-            cueBeatPlayPos, numBeatsToCue & ~0x3);
+            cueBeatPlayPos, -1 * (numBeatsToCue & ~0x3));
```
mr-smidge, line 146: What a blunder... this code has been unchanged for a long time, but it mistakenly trying to find the first bar in the track by scanning forwards instead of backwards.

This resulted in the beatgrid being calculated more and more innaccurately the further forward the main cue point was in the track.

daschuer (approved):
LGTM, Thank you.

Issue #11786, opened 2023-08-03
GLSL Waveforms disappear on resize


tgucio:
### Bug Description

When resizing the window on a large 4k display on a MacBook Pro M1, beyond certain size the waveforms disappear:

![mixxx-waveform](https://github.com/mixxxdj/mixxx/assets/72562119/dbb514ba-4b13-4df4-b67a-591c2ba6dd81)

HW: MacBook Pro 14", 2021 (Apple M1 silicon), macOS Ventura 13.4.1 (c) (22F770820d)
Mixxx version: 2.4-beta-70-g2e3d72f479 

### Version

2.4-beta-70-g2e3d72f479 (Apple silicon build)

### OS

macOS Ventura 13.4.1 (c) (22F770820d) (Apple M1 silicon)

JoergAtGithub:
@m0dB Could you've a look

m0dB:
Hi, I am on holidays and not near a system where I can try to reproduce this. @tgucio are you building mixxx yourself and able / willing try out some things? I would like to know which of the layers (aka renderers) that make up the waveform display cause this issue. But first of all: does this happen with all waveform types?

tgucio:
@m0dB The problem with building is that I don't have an Intel silicon Mac to build on - I take it building on M1/M2 is still not supported officially (I'm getting some failures building [fwcd/m1xxx](https://github.com/fwcd/m1xxx) now).

I've experimented a bit with waveform types and here's what I see:

NOK:
- Filtered (GLSL)
- RGB (GLSL)
- RGB Stacked (GLSL)

OK (not reproduced):
- Filtered (GL)
- Filtered (all-shaders) (GLSL)
- RGB (GL)
- RGB (all-shaders) (GLSL)
- RGB L/R (all-shaders) (GLSL)
- Simple (GL)
- Simple (all-shaders) (GLSL)

foss-:
Another data point: Can't reproduce on
macOS 13.5 (Intel mac)
2.4-beta-75-gc3df3da55d
Tried all three NOK waveforms and resizing on an external 4k monitor.

m0dB:
Seeing that the "all-shaders" waveforms behave fine, I don't see a reason to investigate this further. Instead, we should just recommend (and default to) the all-shaders waveform types, at least on macOS.

daschuer:
We will remove the legacy GLSL Waveforms soon.

Issue #11399, opened 2023-03-23
Waveform jerking with Windows 11 and GeForce GTX770 


daschuer:
### Bug Description

RGB (GLSL) runs only with 44 f/s (of 60) 
RGB (GL) -> 50 f/s
RGB  -> 55 f/s 

RGB wuns with 30 f/s if 30 is selected, but the jerking does not stop and the droped frames are keeping counting. 

My guess is that there is a new anti tearing code active. 

Using 2.4-alpa or @m0dB https://github.com/mixxxdj/mixxx/pull/10989 does not make a difference. 
Where the later gives extra smoothness on my Ubuntu Focal with intel GPU.

### Version

2.3.4

### OS

Windows 11

daschuer:
After setting the v-sync settings in the NVIDIA control panel to various values and back to NVIDIA Default "For 3D applications"
the major jerking was gone. All waveforms are now running without dropped frames and at 60 f/s. 

However there is every 5 seconds a phase of 1 s with a micro jitter. All v-sync settings are effected. 
I have also updated the driver form 456 to 474 but without any effect. 

This issue is the same for all tested Mixxx versions including https://github.com/mixxxdj/mixxx/pull/10989

m0dB:
Can you check if this still is the case?

m0dB:
And can you check if this branch makes a difference: https://github.com/m0dB/mixxx/tree/vsyncthreadless

(This is an experimental branch that does the vsync timing directly in the main thread instead of using a separate thread signaling the main thread, thus avoiding the overhead and impression of the signaling and the hassle of the semaphore)

JoergAtGithub:
This branch doesn't start up:

```
DEBUG ASSERT: "addShaderFromSourceCode( GLShader::Vertex, vertexShaderCode)" in function void __cdecl mixxx::Shader::load(const class QString &,const class QString &)

Qt5Core.dll!00007ffe92d305c8()
Qt5Core.dll!00007ffe92d2e83d()
mixxx.exe!mixxx::`anonymous namespace'::handleMessage(QtMsgType type=QtCriticalMsg, const QMessageLogContext & context={...}, const QString & input={...}) Line 355
	at C:\Users\Joerg.WORLDWARTWEB\source\repos\JoergAtGithub\mixxx\src\util\logging.cpp(355)
[External Code]
mixxx.exe!mixxx::Shader::load(const QString & vertexShaderCode, const QString & fragmentShaderCode={...}) Line 31
[...]

Issue #9589, opened 2022-08-23
Waveforms jerk but not when another window is focused on KDE


mixxxbot:
Reported by: **[mm-mbs](https://launchpad.net/~mm-mbs)**
Date: 2019-02-10T11:13:11Z
Status: Confirmed
Importance: High
Launchpad Issue: [lp1815341](https://bugs.launchpad.net/bugs/1815341)
Tags: waveform
Attachments: [mixxx-2.2.1-rc-RGB-glsl.mp4](https://bugs.launchpad.net/bugs/1815341/+attachment/5237404/+files/mixxx-2.2.1-rc-RGB-glsl.mp4), [mixxx-2.2.1-rc-RGB.mp4](https://bugs.launchpad.net/bugs/1815341/+attachment/5237405/+files/mixxx-2.2.1-rc-RGB.mp4), [Mixxx-qopenglwidget.png](https://bugs.launchpad.net/bugs/1815341/+attachment/5237522/+files/Mixxx-qopenglwidget.png)

---

When I run Mixxx 2.2.1 (Git hash 5c7697662b9aa964344580f9d7e5fe6f75d69a6d) in foreground, then the waveform dither little bit. When I focus another program, then the waveform run smother. When I run Mixxx 2.1.7, then the waveform is always smother.
I take a video where you can see the effect. First I start the song and the waveform dither little bit. Then I change to Firefox and the waveform run smother. Change to Mixxx and the waveform dither again and so on. On the second video I change the waveform type from RGB (GLSL) to RGB and there the same result.

* I compile Mixxx on Debian Stretch.
* There is an Laptop with Intel Core i5 (4 x 1.7GHz) with integrated Intel GPU.

PS: The master branch (8aeda6ff17ba7c135d6a373b3ffa6077c42f5b37) has the same problem.

JoergAtGithub:
@mm-mbs Could you try the following development build of PR10989: https://github.com/mixxxdj/mixxx/suites/12873892391/artifacts/694473745

Does this build still show the waveform jerking under KDE?

daschuer:
This is likely fixed with Mixxx 2.4-alpha. Can you confirm?

PR #11822, opened 2023-08-13
use WaveformWidgetType::AllShaderRGBWaveform as autoChooseWidgetType

m0dB:


daschuer (approved):
LGTM

However this makes me think what is the best way to get rid of the old GLSL types that are known buggy https://github.com/mixxxdj/mixxx/issues/11786 https://github.com/mixxxdj/mixxx/issues/11399 https://github.com/mixxxdj/mixxx/issues/9589 with MIXXX_USE_QOPENGL defined. 

How about remove the enum value "WaveformWidgetType::AllShaderRGBWaveform = 17" and use the old "WaveformWidgetType::GLSLRGBWaveform = 12" value for both conditionally via the MIXXX_USE_QOPENGL macro? We may do this for all GLSL types. 

The drawback is that the 2-4-alpha use that have selected a different "All Shaders" type will end up with RGB. 

What do you think?  

A related issue is https://github.com/mixxxdj/mixxx/issues/8639 btw.

Swiftb0y:
So this changes the default, but how do we make sure that existing installations also get changed on update?

ronso0:
You mean map the legacy types to the respective new type?
Either src/preferences/upgrade.cpp or some upgrade logic in WaveformWidgetFactory::setConfig.
I'd prefer the latter to have the config operations in one place.

daschuer:
How about my idea to just remove the enums for the All Shader types and instead replace the GLSL type with them according to the MIXXX_USE_QOPENGL macro.

This allows a smooth transition forward and backward between 2.3 and 2.4 and users without accelerated types are not bothered.

Swiftb0y:
> How about my idea to just remove the enums for the All Shader types and instead replace the GLSL type with them according to the MIXXX_USE_QOPENGL macro.

That still leaves people on the GL variants [which are also broken](https://github.com/mixxxdj/mixxx/pull/11805#issuecomment-1673584911).

daschuer:
> That still leaves people on the GL variants https://github.com/mixxxdj/mixxx/pull/11805#issuecomment-1673584911.

For my understanding the Shader variants have higher GPU requirements than the GL variants. So I consider the GL users use them for some reasons. At least when they where introduced they did not run everywhere. What is the situation now?

daschuer:
At least we check `QOpenGLShaderProgram::hasOpenGLShaderPrograms()`. Maybe this is also an issue with openGLES on Rasberry Pi and friends.

But anyway, the question of migrating users from GL to All-Shaders can be decided separate form the question to use one settings for GLSL and All Shaders depending on the Mixxx version/build flags.

Swiftb0y:
> For my understanding the Shader variants have higher GPU requirements than the GL variants. So I consider the GL users use them for some reasons. At least when they where introduced they did not run everywhere. What is the situation now?

not sure. I think its pretty safe now. The shaders all only require OpenGL 2.1 and might also be compatible with OpenGL ES. But I have no experince with GLES so idk.

m0dB:
I expect the all-shaders waveform types to be faster on all platforms and they should work on OpenGL ES (I haven't tested this but I don't see why not).

I would advocate for forcing all users that upgrade to 2.4 to the default (RGB all-shaders) either by using the upgrade mechanism @ronso0 mentioned (not in front of a computer so not sure about this) or by bumping the value of the old enums to a higher value so the switch will fail on fall back to the default (but they will still be available for selection).

m0dB:
With all the discussion here and in zulip, and looking at the code, it's still not clear to me what it is we want to do. So let me resume what I think it the objective:

1. The allshaders waveform types are generally preferred
2. When possible we maintain for now the other waveform types 
3. New users should start with a good default (I would say "RGB (all-shaders)")
4. Old users with waveform types that don't exist anymore should either
4.a. start with the same as 3
4.b. start with an allshader waveform type that matches the old config as good as possible
5. Old users with waveform types that still exist should be proactively moved to an allshader waveform type but have the possibility to go back. The waveform type they are moved to is either
5.a. the same as 3
5.b. an allshader waveform type that matches the old config as good as possible

Is this correct?

If so, I don't think it's worth it bothering with 4.b. and 5.b.; we could simply move all users to "RGB (all-shaders)", users that changed the waveform type in the past will know how to change it again.

Swiftb0y:
> Is this correct?

In my opinion, this is completely correct, though I don't know whether every team member agrees with me here. 

> If so, I don't think it's worth it bothering with 4.b. and 5.b.; we could simply move all users to "RGB (all-shaders)", users that changed the waveform type in the past will know how to change it again.

Right, its not super necessary, but IMO trivial enough to implement (just a giant switch case mapping in a separate function somewhere) that we should still do it. Though I'm willing to change my opinion if it turns out that it does require more work than the 20 SLOC I estimate for that switch-case.

Swiftb0y:
As ronso0 pointed out already (thank you). The config migration is likely trivial to implement here:
https://github.com/mixxxdj/mixxx/blob/28ad94197030bc82a9ff4d17607a15d5b234d523/src/preferences/upgrade.cpp#L453

daschuer:
The most work for me.

Did you consider the idea to hard replace the GLSL types with the all Shader types? This way we don't even need to migrate any config. 

We can do the same for the GL types if we are confident they are no longer needed and buggy anyway. This is all around the question if there is hardware that has openGL but no OpenGLShaderProgram::hasOpenGLShaderPrograms()

From the history of Mixxx I remebers some issues with badly configured GPU drivers. There is also a blacklist in Qt. But things may have changed. My old EPC was able to use the GL waveforms with Qt4 but not with Qt5. And how about devices using Angle? 

The 60 Hz default is a good idea in any case.

Swiftb0y:
> Did you consider the idea to hard replace the GLSL types with the all Shader types? This way we don't even need to migrate any config.

Yes I did consider that but I have thre issues with that. In ascending importance:
1. I don't like the idea of switching out these implementations by changing the contents of an enum (something that probably shouldn't have been exposed in the first place). 
2. I have with that is that considering the timing of the all-shader introduction we should preserve as much of the previous waveforms since I'm not confident they will replace the legacy waveforms in 100% of all cases. 
3. In order to confidently retire the legacy waveforms (GL & GLSL) we need to be confident they can actually serve as a replacement. In order to test that, we should do a test rollout (this would usually only be done to a fraction of users, but in this case I'd argue we could do it for everyone). In order to confidently remove the GL waveforms in future versions (which, let me remind you, will be required for Qt6!) we need to ensure the all-shaders can actually replace them.

> From the history of Mixxx I remebers some issues with badly configured GPU drivers. There is also a blacklist in Qt. But things may have changed. My old EPC was able to use the GL waveforms with Qt4 but not with Qt5. And how about devices using Angle?

Right, that's one of the reasons for 3. If people experience issues, they can reported them and we can instruct them to manually switch back.

m0dB:
Ok, those are valid points, @Swiftb0y . But in that case I do think it would be good to not only actively move users to the all-shaders waveform types (in upgrade.cpp), but also to mark the other waveform types as "legacy". This way old and new users will understand that the all-shaders ones are the ones to use by default. And it allows me to get rid of the "all-shaders" label ;-)

Would that work for you?

Swiftb0y:
Yes, lets mark them as legacy.

m0dB:
Okay, I'll add these changes to this PR.

m0dB:
I have made the following changes, which I hope match what we discussed and agreed upon:

- Changed the default framerate to 60 fps and the default waveform type to AllShaderRGBWaveform
- When upgrading from pre-2.4.0 to 2.4.0:
 - map the WaveformType as set in the config to the most similar allshader waveform widget type and change the config
 - set the framerate to 60 fps
- In the preference dialog, the "(all-shader)"  postfix has been removed, and instead the old waveform widget types are now marked "(legacy)"
- I did this by changing the static bool developerOnly() to static WaveformWidgetCategory category(). Based on the category a different string is added to the name in the preferences panel. So developerOnly is now WaveformWidgetCategory::DeveloperOnly.

While doing the mapping I realized I never added the all-shader implementation of the HSV waveform type (which got lost in some branch). I have added this.

To test if the upgrade mechanism works, you will either have to edit your mixxx.cfg file and set the Version to 2.3.9, or you will have to set(MIXXX_VERSION_PRERELEASE "") in the CmakeLists.txt

src/preferences/upgrade.cpp
```diff
@@ -29,6 +30,61 @@ Upgrade::Upgrade()
...
+namespace {
+// mapping to proactively move users to the new all-shader waveform types
+WaveformWidgetType::Type upgradeToAllShaders(WaveformWidgetType::Type waveformType) {
+    switch (waveformType) {
+    case WaveformWidgetType::EmptyWaveform:
+        return waveformType;
+    case WaveformWidgetType::SoftwareSimpleWaveform:
+        return WaveformWidgetType::AllShaderSimpleWaveform;
+    case WaveformWidgetType::SoftwareWaveform:
+        return WaveformWidgetType::AllShaderRGBWaveform;
```
Swiftb0y, line 42: I'm not sure if we want to upgrade to an accelerated waveform since those exist purely for the cases where accelerated waveforms are not available. If acceleration is not available, mixxx should still automatically fall back to the corresponding CPU-waveform right?

src/waveform/renderers/allshader/waveformrendererhsv.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+using namespace allshader;
+
+WaveformRendererHSV::WaveformRendererHSV(
+        WaveformWidgetRenderer* waveformWidget)
+        : WaveformRendererSignalBase(waveformWidget) {
+}
+
+WaveformRendererHSV::~WaveformRendererHSV() {
+}
```
Swiftb0y, line 21:
```suggestion
```

src/waveform/renderers/allshader/waveformrendererhsv.cpp
```diff
@@ -0,0 +1,213 @@
...
+        QColor color;
+        color.setHsvF(h, 1.0f - hi, 1.0f - lo);
+
+        // lines are thin rectangles
+        // maxAll[0] is for left channel, maxAll[1] is for right channel
+        m_vertices.addRectangle(fpos - 0.5f,
+                halfBreadth - heightFactor * maxAll[0],
+                fpos + 0.5f,
+                halfBreadth + heightFactor * maxAll[1]);
+        m_colors.addForRectangle(color.redF(), color.greenF(), color.blueF());
```
Swiftb0y, line 183: needs static_casts for Qt5...
```suggestion
        m_colors.addForRectangle(static_cast<float>(color.redF()), static_cast<float>(color.greenF()), static_cast<float>(color.blueF()));
```

src/waveform/renderers/allshader/waveformrendererhsv.h
```diff
@@ -0,0 +1,30 @@
...
+#include "waveform/renderers/allshader/waveformrenderersignalbase.h"
+
+namespace allshader {
+class WaveformRendererHSV;
+}
+
+class allshader::WaveformRendererHSV final : public allshader::WaveformRendererSignalBase {
+  public:
+    explicit WaveformRendererHSV(WaveformWidgetRenderer* waveformWidget);
+    ~WaveformRendererHSV() override;
```
Swiftb0y, line 16:
```suggestion
```

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -1028,14 +1033,29 @@ void WaveformWidgetFactory::evaluateWidgets() {
...
+            category = allshader::HSVWaveformWidget::category();
             break;
 #endif
         default:
             DEBUG_ASSERT(!"Unexpected WaveformWidgetType");
             continue;
         }
 
+        if (category == WaveformWidgetCategory::Legacy) {
+            widgetName += " " + tr("(legacy)");
```
Swiftb0y, line 1056: lets ensure the translators don't swallow the parentheses.
```suggestion
            widgetName += QStringLiteral(" (%1)").arg(QObject::tr("legacy"));
```

src/waveform/widgets/allshader/hsvwaveformwidget.h
```diff
@@ -0,0 +1,46 @@
...
+class WaveformWidgetFactory;
+
+namespace allshader {
+class HSVWaveformWidget;
+}
+
+class allshader::HSVWaveformWidget final : public allshader::WaveformWidget {
+    Q_OBJECT
+  public:
+    ~HSVWaveformWidget() override;
```
Swiftb0y, line 15: no destructor necessary as far as I can tell because the WaveformWidget one is already non-virtual, so its not called polymorphically, right?
```suggestion
```

Swiftb0y (changes requested):
Thank you very much for your continued time-consuming effort. I have a couple minor comments.

src/preferences/upgrade.cpp
m0dB: Yes, you are right. I will changes these to return waveformType.

src/waveform/renderers/allshader/waveformrendererhsv.cpp
m0dB: Yeah, I copy/pasted this from WaveformRendererRGB which also has this (and maybe others too). I will remove all these unnecessary destructors.

src/waveform/renderers/allshader/waveformrendererhsv.cpp
m0dB: Thanks!

src/waveform/widgets/allshader/hsvwaveformwidget.h
m0dB: Removed

src/waveform/renderers/allshader/waveformrendererhsv.h
m0dB: Removed

src/preferences/upgrade.cpp
```diff
@@ -329,6 +385,29 @@ UserSettingsPointer Upgrade::versionUpgrade(const QString& settingsPath) {
         configVersion = "1.9.0";
         config->set(ConfigKey("[Config]","Version"), ConfigValue("1.9.0"));
     }
+
```
Swiftb0y, line 388: this is quite early in the function. Meaning that the rest of the function body seems to be sorted in ascending version order and this new section is not sorted that way. Can somebody with the upgrade procedure check whether this is correct here?

Swiftb0y (commented):
LGTM otherwise. Also please go back and rewrite the last couple of commits to apply pre-commit on the failing commit. Thank you

src/preferences/upgrade.cpp
m0dB: Yes, this wasn't very clear to me either. I double-checked and I see that I have to move it a bit lower.

But it has to go before the block 
```
if (configVersion.startsWith("1.11")) {
```
because at the end of that block the version in the config is set to the current version.
```
config->set(ConfigKey("[Config]", "Version"), ConfigValue(VersionStore::version()));
```

I will leave a comment explaining this.

m0dB: Ok, I would say it's correct now. Thanks @Swiftb0y for noticing that it looked a bit fishy, after studying the code a bit more I now think I understand the intent and I added a comment. But it would indeed be good if someone who knows this code better can confirm.

Swiftb0y (approved):
Thank you. waiting for CI and someone more knowledgeable to clear up my unresolved review comment.

src/preferences/upgrade.cpp
Swiftb0y: Yeah, judging by the blame, none of the actual changes have been made by people that are still active...

ronso0: I could at least take a look but the I'll be afk for a few more days

m0dB:
@Swiftb0y your comment has been resolved, right? The upgrades are now all in ascending version order. I have tripple checked and I am sure the code is correct. Besides, it's all a bit academic; apart from the upgrade I added, the remaining upgrade code is for versions that are very old.

I would like to see this PR approved and merged as soon as possible. It is unfortunate to have to tell people all the time "Try 2.4, but please make sure you change the waveform type first" 😅

JoergAtGithub:
If I see this right, @ronso0 wants to have a final look here. But he's away from keyboard for some days.

ronso0:
Not quite, since I'm also not super familiar with upgrade.cpp I'd just be that second pair of eyes. I doubt I find time before sunday, so if we're in a hurry someone else needs to take a look.

m0dB:
Ping!

src/preferences/upgrade.cpp
```diff
@@ -29,6 +30,61 @@ Upgrade::Upgrade()
...
+        return waveformType;
+    case WaveformWidgetType::AllShaderSimpleWaveform:
+        return waveformType;
+    case WaveformWidgetType::AllShaderHSVWaveform:
+        return waveformType;
+    case WaveformWidgetType::Count_WaveformwidgetType:
+        return waveformType;
+    }
+    return WaveformWidgetType::AllShaderRGBWaveform;
+}
```
Swiftb0y, line 85: I prefer this because its easier to see what types return a certain value. 
```suggestion
WaveformWidgetType::Type upgradeToAllShaders(WaveformWidgetType::Type waveformType) {
    // TODO: convert `WaveformWidgetType::Type` to an enum class then shorten more `using enum ...`
    using WWT = WaveformWidgetType;
    switch (waveformType) {
    case WWT::EmptyWaveform:
    case WWT::SoftwareSimpleWaveform:
    case WWT::SoftwareWaveform:
    case WWT::GLVSyncTest:
    case WWT::QtVSyncTest:
    case WWT::AllShaderRGBWaveform:
    case WWT::AllShaderLRRGBWaveform:
    case WWT::AllShaderFilteredWaveform:
    case WWT::AllShaderSimpleWaveform:
    case WWT::AllShaderHSVWaveform:
    case WWT::Count_WaveformwidgetType:
        return waveformType;
    case WWT::QtSimpleWaveform:
    case WWT::GLSimpleWaveform:
        return WaveformWidgetType::AllShaderSimpleWaveform;
    case WWT::GLFilteredWaveform:
    case WWT::GLSLFilteredWaveform:
        return WaveformWidgetType::AllShaderFilteredWaveform;
    case WWT::QtWaveform:
    case WWT::HSVWaveform:
    case WWT::RGBWaveform:
    case WWT::GLRGBWaveform:
    case WWT::GLSLRGBWaveform:
    case WWT::QtRGBWaveform:
    case WWT::GLSLRGBStackedWaveform:
        return WaveformWidgetType::AllShaderHSVWaveform;
    case WWT::QtHSVWaveform:
        return WaveformWidgetType::AllShaderHSVWaveform;
    }
    return WaveformWidgetType::AllShaderRGBWaveform;
}
```

Swiftb0y (commented):
ronso0 seems busy so I took another review lap. Found this minor nitpick but LGTM otherwise.

m0dB:
Thanks, @Swiftb0y . I applied your suggested change.

src/preferences/upgrade.cpp
```diff
@@ -29,6 +30,45 @@ Upgrade::Upgrade()
...
+    case WWT::QtWaveform:
+    case WWT::HSVWaveform:
+    case WWT::RGBWaveform:
+    case WWT::GLRGBWaveform:
+    case WWT::GLSLRGBWaveform:
+    case WWT::QtRGBWaveform:
+    case WWT::GLSLRGBStackedWaveform:
+        return WaveformWidgetType::AllShaderHSVWaveform;
+    case WWT::QtHSVWaveform:
+        return WaveformWidgetType::AllShaderHSVWaveform;
```
ronso0, line 66:
```suggestion
    case WWT::GLSLRGBStackedWaveform:
    case WWT::QtHSVWaveform:
        return WaveformWidgetType::AllShaderHSVWaveform;
```

ronso0:
> ronso0 seems busy so I took another review lap. Found this minor nitpick but LGTM otherwise.

Yes, I was busy and afk, and this slipped off my ToDo list tbh, sorry for that.
I commented on the upgrad.cpp changes.

src/preferences/upgrade.cpp
ronso0: > Good catch, but your suggestion looks wrong. Do this instead:

Yes, I didn't check the upgrade behaviour in detail, just the proposed code, so yeah: migrating RGB -> RGB makes sense, of course.

src/preferences/upgrade.cpp
```diff
@@ -29,6 +30,45 @@ Upgrade::Upgrade()
...
+        return WaveformWidgetType::AllShaderFilteredWaveform;
+    case WWT::QtWaveform:
+    case WWT::HSVWaveform:
+    case WWT::RGBWaveform:
+    case WWT::GLRGBWaveform:
+    case WWT::GLSLRGBWaveform:
+    case WWT::QtRGBWaveform:
+    case WWT::GLSLRGBStackedWaveform:
+        return WaveformWidgetType::AllShaderHSVWaveform;
+    case WWT::QtHSVWaveform:
```
Swiftb0y, line 65: sorry, I was a little sloppy with my previous suggestion. I double checked and now it should be correct. I also forgot to apply the `WWT` alias, so do these for the others as well. Please amend and force push. Thank you.
```suggestion
    case WWT::QtWaveform:
    case WWT::RGBWaveform:
    case WWT::GLRGBWaveform:
    case WWT::GLSLRGBWaveform:
    case WWT::QtRGBWaveform:
    case WWT::GLSLRGBStackedWaveform:
        return WWT::AllShaderRGBWaveform;
    case WWT::HSVWaveform:
    case WWT::QtHSVWaveform:
```

Swiftb0y (changes requested):

src/preferences/upgrade.cpp
m0dB: I will make the suggestion changes this weekend and with that hopefully this can be merged.

Swiftb0y: I'm sorry. This could've been merged a week ago if I didn't add the previous suggestion which had introduced a bug. My fault.

m0dB:
I rebased the last 3 commits (grouping per case, use of WWT) into 1 commit, not sure if I should have done that.
Anyway, I think it's all good now and ready to be merged! Thanks for you help.

Swiftb0y (approved):
> I rebased the last 3 commits (grouping per case, use of WWT) into 1 commit, not sure if I should have done that.

Thank you, that's perfect. LGTM

daschuer:
Now the box looks more overcrowded than ever. I think it needs some love.  
I have just filed: https://github.com/mixxxdj/mixxx/issues/11915

Issue #10667, opened 2022-08-23
Traktor-Kontrol-S2-MK2-hid-scripts.js uses no longer existing CO "load_preset"


mixxxbot:
Reported by: **[JoergAtGithub](https://github.com/JoergAtGithub)**
Date: 2022-02-17T18:27:11Z
Status: New
Importance: Undecided
Launchpad Issue: [lp1961331](https://bugs.launchpad.net/bugs/1961331)
Tags: controllers, effects

---

This CO is existing until 2.3, but no longer in Main. I guess that "loaded_chain_preset" can somehow replace "load_preset", but I've no Traktor-Kontrol-S2-MK2 to test this.

--
In the meantime we have also a mapping for the Traktor Kontrol S2 MK1, where this code section got copied.

PR #11237, opened 2023-02-01
Traktor S2 MK1: Add calibration and refactor

leifhelm:
As discussed in #3905 with @JoergAtGithub this PR add the capability to read the feature report of the controller and apply the calibration stored in it.
I cleaned up the messy code and replaced it with nice OOP code.

~~I guess this has to wait until #3905 gets into 2.4.~~

PR #11823, opened 2023-08-13
Traktor Kontrol S2 Mk2 fix loaded chain preset CO

JoergAtGithub:
Closes: #10667

I don't have this controller and therefore did only this minimal change without running pre-commit. Therefore pre-commit might fail on CI.

The change was verified to work on the very similar S2 Mk1 (only the jog wheels differ between these controllers) in https://github.com/mixxxdj/mixxx/pull/11237/commits/add0f497bf1176b95486ccc0e68f8f776ec58727


Swiftb0y (approved):
LGTM. thank you. We still need to document these new COs in the manual.

ronso0:
> LGTM. thank you. We still need to document these new COs in the manual.

https://github.com/mixxxdj/manual/issues/501 it is, I have this tab open for quite some time now and will tackle that after the release.

PR #11827, opened 2023-08-13
Fix for randomly not loaded CoverArt in Spinnies

JoergAtGithub:
Im not sure if this fix is the appropriated solution, but it fixes the issue #11131 for me. 
The problem is, that the requestor of cached CoverArt is not always WSpinnyBase - it can also be WCoverArt.

### DIFF
--- a/src/widget/wspinnybase.cpp
+++ b/src/widget/wspinnybase.cpp
@@ -297,8 +297,7 @@ void WSpinnyBase::slotCoverFound(
         bool coverInfoUpdated) {
     Q_UNUSED(requestedCacheKey);
     Q_UNUSED(coverInfoUpdated); // CoverArtCache has taken care, updating the Track.
-    if (pRequestor == this &&
-            m_pLoadedTrack &&
+    if (m_pLoadedTrack &&
             m_pLoadedTrack->getLocation() == coverInfo.trackLocation) {
         setLoadedCover(pixmap);
         coverChanged();


daschuer (changes requested):
Every widget is running it's own request, because of size constrains. 
There must be a race-codition that the cover is not requested of the responds is lost. 
Is WSpinnyBase::slotTrackCoverArtUpdated always called?

JoergAtGithub:
I set a breakpoint in `WSpinnyBase::slotTrackCoverArtUpdated()` and it's was never called.

ronso0:
I agree with daschuer, now the spinny will receive signals meant for other widgets, so it will receive more unneeded updates.

@JoergAtGithub You found out https://github.com/mixxxdj/mixxx/issues/11131#issuecomment-1676401662
> The following function is executed three times, when a track is loaded into a deck. The first two times pixmap is alway NULL and it returns early. The third call differs in failure case, when the Coverart is not loaded, pixmap is NULL as well.
> 
> https://github.com/mixxxdj/mixxx/blob/4a51c393eb8820a5728f4b2b53dce8ef203d2b8c/src/widget/wspinnybase.cpp#L377-L386

I wonder why it's called 3 times in the first place.

JoergAtGithub:
Unfortunately this bug is not fixed by #11840

ronso0:
Sorry, I confused it with the slip indicator bug.

JoergAtGithub:
Closing this PR - I agree, that this is not the correct fix for this bug.

PR #11035, opened 2022-11-03
Sync Lock: End of track checking is not needed

ywwg:
It's not necessary to unset leader status at the end of the track based on the playposition heck.  It will change automatically anyway.

PR #11831, opened 2023-08-14
After disabling Explicit leader, request always follower mode 

### LINKED ISSUE #11829: Reseting explicit sync leader picks the wrong implicit leader
### Bug Description

Current behavior: The previous explicit leader is selected even if the leader is no longer playing. 
Desired behavior:  Pick a playing deck.

### Version

2.3-beta

### OS

_No response_

daschuer:
… and let Mixxx pick a new Soft leader.

This fixes the bug that a stopped or silenced deck can become a Soft leader. #11829

### DIFF
--- a/src/engine/sync/enginesync.cpp
+++ b/src/engine/sync/enginesync.cpp
@@ -60,11 +60,18 @@ void EngineSync::requestSyncMode(Syncable* pSyncable, SyncMode mode) {
     }
     case SyncMode::Follower: {
         // A request for follower mode may be converted into an enabling of soft
-        // leader mode.
-        activateFollower(pSyncable);
+        // leader mode if this still be best choice
+        if (m_pLeaderSyncable == pSyncable) {
+            // Avoid that pickLeader() returns pSyncable with SyncMode::LeaderExplicit
+            m_pLeaderSyncable = nullptr;
+        }
         Syncable* newLeader = pickLeader(pSyncable);
+        if (newLeader != pSyncable) {
+            activateFollower(pSyncable);
+        }
         if (newLeader && newLeader != m_pLeaderSyncable) {
             // if the leader has changed, activate it (this updates m_pLeaderSyncable)
+            // But don't make an an explicit leader soft.
             activateLeader(newLeader, SyncMode::LeaderSoft);
         }
         break;
--- a/src/engine/sync/enginesync.cpp
+++ b/src/engine/sync/enginesync.cpp
@@ -59,20 +59,22 @@ void EngineSync::requestSyncMode(Syncable* pSyncable, SyncMode mode) {
         break;
     }
     case SyncMode::Follower: {
-        // A request for follower mode may be converted into an enabling of soft
-        // leader mode if this still be best choice
-        if (m_pLeaderSyncable == pSyncable) {
-            // Avoid that pickLeader() returns pSyncable with SyncMode::LeaderExplicit
-            m_pLeaderSyncable = nullptr;
-        }
-        Syncable* newLeader = pickLeader(pSyncable);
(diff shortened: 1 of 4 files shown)


Holzhaus:
Test fail on windows

JoergAtGithub:
I don't think this is the correct behavior. IMHO we had the handling of track end until https://github.com/mixxxdj/mixxx/pull/11035
which removed the code instead of fixing the bug in it.

daschuer:
This has nothing to do with end of track and explicit leader. 

The scenario is that you have picked an explicit leader and you are done with the track. In current 2.4 you have the option to pick a new explicit leader or make the original done track a soft leader. The later is the bug, because there is no usecase for this. This PR picks the best deck as soft leader, which should be the other deck currently playing.

daschuer:
@JoergAtGithub Can you have another look?

ywwg:
@daschuer I think this is the correct fix -- once the explicit leader is done playing the DJ will want something else useful to be the new leader, and that will probably be some other playing deck.  Selecting Follower will trigger the implicit leader selection, and I think that's good behavior

src/engine/sync/enginesync.cpp
```diff
@@ -60,11 +60,18 @@ void EngineSync::requestSyncMode(Syncable* pSyncable, SyncMode mode) {
     }
     case SyncMode::Follower: {
         // A request for follower mode may be converted into an enabling of soft
-        // leader mode.
-        activateFollower(pSyncable);
+        // leader mode if this still be best choice
+        if (m_pLeaderSyncable == pSyncable) {
+            // Avoid that pickLeader() returns pSyncable with SyncMode::LeaderExplicit
```
ywwg, line 65: this code feels like a workaround for a bug.  Am I right that just calling activateFollower(pSyncable) is not working because that syncable is already explicit leader?  I think this code should remain the same and the edge case should be fixed elsewhere

src/engine/sync/synccontrol.cpp
```diff
@@ -519,8 +519,6 @@ void SyncControl::slotSyncLeaderEnabledChangeRequest(double state) {
         // Turning off leader goes back to follower mode.
         switch (mode) {
         case SyncMode::LeaderExplicit:
-            m_pChannel->getEngineBuffer()->requestSyncMode(SyncMode::LeaderSoft);
```
ywwg, line 522: yes this looks right

Holzhaus: Wasn't this intentional? I think the explicit leader also enabled scratching both decks in sync. In case you don't want this but you're unhappy with Mixxx automatic selection, the trick was to make a deck explicit leader, then disable the explicit leader to make it soft leader.

ywwg: hm good question

ywwg: yeah that could be confusing to users, it might pick a different  deck.  However in practice, when implicit leader is active there's no functional difference between decks.  Adjusting rate on any deck will adjust rates on all other decks.

I think I agree, we want to keep this the way it was and just focus on the specific bug

daschuer:
Should work now as demanded.

daschuer:
It is now working. It was more tricky than expected. Sorry for the rebase.

JoergAtGithub (approved):
I built this PR and played a bit with the different state of the decks (audible/not audible, playing/not-playing, synced/not-synced, track end). I needed a bit to get used to it, but I think the behaviour is correct now.
Code LGTM!

daschuer:
Because this is a bug-fix I have changed the target to 2.4 
Merge?

JoergAtGithub:
Thank you!

PR #11834, opened 2023-08-15
Update to vcpkg environment with benchmark and libdjinterop 0.19.2

daschuer:


JoergAtGithub:
Tested this on Win11 and works as expected! Thank you!

Issue #11836, opened 2023-08-17
Waveform playmarker position shown outside loop, while track is inside loop


JoergAtGithub:
### Bug Description

In the screenshot below, you see the blue actual playmarker position outside the active green loop: 
![grafik](https://github.com/mixxxdj/mixxx/assets/64457745/86cb1a93-7f59-4653-8c10-79f7e634e687)
Above the loop size was 15ms and the refresh interval 60fps (16,6ms).

The problem increases with lower framerate (here 20fps - 50ms):
![grafik](https://github.com/mixxxdj/mixxx/assets/64457745/61c1174c-8be3-46d2-a0b9-70430949a5f1)

### Version

2.4.0beta

### OS

Windows11

m0dB:
I think this can be done easily by adding `loopInfo.startPosition` and `loopInfo.endPosition` from the `EngineBuffer`'s `m_pLoopingControl` to the arguments to the `m_visualPlayPos->set(...)` call in `EngineBuffer`.

In `VisualPlayPosition::getAtNextVSync` and `VisualPlayPosition::getPlaySlipAtNextVSync`, we can than adjust the predicted `playPosition` and `slipPosition` accordingly.

The code to do the adjustment is in `LoopControl::adjustedPositionInsideAdjustedLoop` which doesn't seem to use any member variables to it could be moved to a static method.

ronso0:
Yeah, might work. That's what I meant by "copying engine code to waveforms" 😉

m0dB:
Right, but I mean that this can be done without any actual code duplication. There are other ways to do this (e.g. replicating the entire loop control at the UI side which would act on the some control parameters) that would certainly involve a lot more "copying engine code to waveforms" 😅 

@JoergAtGithub maybe you can give it a try with these suggestions? Otherwise, I can have a go at it. But I assume this is not a blocker for 2.4, so no hurry, right?

ronso0:
Yep, no blocker, it's 'always' been like this.

Great if it's as easy as you say 👍

daschuer:
@JoergAtGithub: Can you confirm you have taken the screenshot while playing and that is does not happen when paused? 
Than it is the mentioned side effect of quantization of loop cycle, audio buffer and refresh rate. 

I am not sure if the mentioned solution by m0dB is worth the work, because this will introduce other issues like waveform moving backwards or stopping. Just like wheels in movies turning backwards in certain speeds.

JoergAtGithub:
[...]

PR #11435, opened 2023-04-03
Slip mode: consider active (regular) loop

ronso0:
I was curious why this hasn't been implemented earlier, and couldn't really believe "this is super complicated though so it will be a lot of effort to fix" (2013, #6993).

So, this is POC and it performs nicely with active loops (no change for rolling loops).
I  didn't run the tests, yet, only tested manually: play a track, set up a loop, clone that deck, then engage slip mode, mess around a bit with the waveform. Disabled slip mode and both decks were in sync.

I don't intend to finish this (unless just polishing is required for merging), I just want to share my experiment and give a starting point for others, so feel free to pick this up.
I guess the code can be improved and the tests should probably to be extended.

PR #11840, opened 2023-08-19
Fix wrong visual play position when inside loop

JoergAtGithub:
Closes: #11836

The loop is now considered in VisualPlayPosition::getAtNextVSync which determines the visual play position used fo the waveform rendering. The play position indicator is no always inside an active loop. Forward and reverse playing direction is considered.

It wasn't neccessary to change VisualPlayPosition::getPlaySlipAtNextVSync because this is only used for the spinnies, which do not visualize loops. Therefore no special handling for slip position was necesarry.

Furthermore I unified some odd sympol naming in the VisualPlayPosition code.

### DIFF
--- a/src/engine/controls/loopingcontrol.h
+++ b/src/engine/controls/loopingcontrol.h
@@ -58,12 +58,23 @@ class LoopingControl : public EngineControl {
     void setLoop(mixxx::audio::FramePos startPosition,
             mixxx::audio::FramePos endPosition,
             bool enabled);
-    mixxx::audio::FramePos getLoopStartPosition() {
-        return m_loopInfo.getValue().startPosition;
+
+    enum class LoopSeekMode {
+        Changed, // force the playposition to be inside the loop after adjusting it.
+        MovedOut,
+        None,
+    };
+
+    struct LoopInfo {
+        mixxx::audio::FramePos startPosition;
+        mixxx::audio::FramePos endPosition;
+        LoopSeekMode seekMode;
+    };
+
+    LoopInfo getLoopInfo() {
+        return m_loopInfo.getValue();
     }
-    mixxx::audio::FramePos getLoopEndPosition() {
-        return m_loopInfo.getValue().endPosition;
-    }
+
     void setRateControl(RateControl* rateControl);
 
     bool isLoopingEnabled() {
@@ -128,18 +139,6 @@ class LoopingControl : public EngineControl {
     void slotLoopEnabledValueChangeRequest(double enabled);
 
   private:
-    enum class LoopSeekMode {
-        Changed, // force the playposition to be inside the loop after adjusting it.
-        MovedOut,
-        None,
-    };
-
-    struct LoopInfo {
-        mixxx::audio::FramePos startPosition;
-        mixxx::audio::FramePos endPosition;
-        LoopSeekMode seekMode;
-    };
-
     void setLoopingEnabled(bool enabled);
     void setLoopInToCurrentPosition();
     void setLoopOutToCurrentPosition();
--- a/src/engine/controls/loopingcontrol.h
+++ b/src/engine/controls/loopingcontrol.h
@@ -58,6 +58,12 @@ class LoopingControl : public EngineControl {
     void setLoop(mixxx::audio::FramePos startPosition,
             mixxx::audio::FramePos endPosition,
             bool enabled);
+    mixxx::audio::FramePos getLoopStartPosition() {
+        return m_loopInfo.getValue().startPosition;
+    }
+    mixxx::audio::FramePos getLoopEndPosition() {
+        return m_loopInfo.getValue().endPosition;
+    }
     void setRateControl(RateControl* rateControl);
     bool isLoopingEnabled();
     bool isLoopRollActive();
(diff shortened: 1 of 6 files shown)


src/engine/enginebuffer.cpp
```diff
@@ -1387,7 +1387,17 @@ void EngineBuffer::updateIndicators(double speed, int iBufferSize) {
     m_iSamplesSinceLastIndicatorUpdate += iBufferSize;
 
     const double fFractionalPlaypos = fractionalPlayposFromAbsolute(m_playPosition);
```
m0dB, line 1389: if you rename m_slipPosition to m_slipPos , also rename m_playPosition to m_playPos

m0dB:
Nice to see it was indeed straightforward and could be done with a few changes!

daschuer:
>It wasn't neccessary to change VisualPlayPosition::getPlaySlipAtNextVSync because this is only ...

The slip position also follows loops. Even though there are no loop markers on the spinners, it would be correct to consider loops as well.

JoergAtGithub:
> 
> The slip position also follows loops

This seems to be another bug. According to our manual, not to follow loops is explicitly the purpose of the Slip mode:

`When active, the playback continues muted in the background during a loop, reverse, scratch, etc.`

ronso0:
I think @daschuer refers to the situation where a loop is enabled before slip mode is enabled, see #11435

ronso0:
FWIW I noticed the offset with quantized full-track loops aka repeat where the playpos would be before track start for a glimpse, and then jump to the actual engine playpos.
Hence I was thinking about using `LoopingControl::nextTrigger` (which considers both loops and repeat) when I gave my pessimistic view on a quick fix.

JoergAtGithub:
> > It wasn't neccessary to change VisualPlayPosition::getPlaySlipAtNextVSync because this is only ...
> 
> The slip position also follows loops. Even though there are no loop markers on the spinners, it would be correct to consider loops as well.

In general I agree, but this is out of scope of this waveform PR!
I tried to create a follow-up PR for the spinnies, but I struggled with the behaviour of the Slip Mode in general. It just not does what I expect from a Slip mode, and the documentation doesn't helped me to get a better understanding.

daschuer (changes requested):
This does not work with catching loops in font of the play position. In this case the waveform shows already the loop.

src/waveform/visualplayposition.cpp
```diff
@@ -77,7 +83,27 @@ double VisualPlayPosition::getAtNextVSync(VSyncThread* pVSyncThread) {
     if (m_valid) {
         const VisualPlayPositionData data = m_data.getValue();
         const double offset = calcOffsetAtNextVSync(pVSyncThread, data);
-        return data.m_enginePlayPos + offset * data.m_rate;
+
+        double interpolatedPlayPos = data.m_playPos + offset * data.m_playRate;
+
+        if (data.m_loopEnabled) {
```
daschuer, line 89: I think this must only applied when the `data.m_playPos` is on one side of the loop marker and the  `interpolatedPlayPos` is on the other side.

daschuer:
You can use exactly the same code in getPlaySlipAtNextVSync() for the spinnies for both positions slip and normal. 
Both will follow the loop.

src/waveform/visualplayposition.cpp
JoergAtGithub: I've commited a fix.

JoergAtGithub:
> You can use exactly the same code in getPlaySlipAtNextVSync() for the spinnies for both positions slip and normal. Both will follow the loop.

I don't think it's that easy for the slip position, because it depends on the order of activation of loop and slip mode. I worked already on a follow-up PR for the spinnies, but than found: #11844

JoergAtGithub:
Based on #11848 I could easily add the code for the slip indicator in the spinnies too.
The new code also fixes the visual bug, that the slip indicator appeared for a moment at 12 o'clock position when enabling it.

src/waveform/visualplayposition.h
```diff
@@ -14,6 +14,12 @@ typedef void VSyncThread;
 class VSyncThread;
 #endif
 
+enum class SlipModeStates {
```
ronso0, line 17:
```suggestion
enum class SlipModeState {
```
I think singular is used in other places

JoergAtGithub:
Ping

src/engine/controls/loopingcontrol.h
```diff
@@ -58,6 +58,12 @@ class LoopingControl : public EngineControl {
     void setLoop(mixxx::audio::FramePos startPosition,
             mixxx::audio::FramePos endPosition,
             bool enabled);
+    mixxx::audio::FramePos getLoopStartPosition() {
```
daschuer, line 61: Loop Start and End should be only read atomically. This avoids that the position can slip out during moving the loop. 
In this case LoopInfo should be returned, or both values in a single call.

src/engine/slipmodestate.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+enum class SlipModeState {
+    Disabled = 0,
+    Armed = 1,
```
daschuer, line 5: Can you add a brief comment about armed?

daschuer (changes requested):
Thank you. Two remaining comments.

JoergAtGithub:
Requested changes are done!

daschuer (approved):
LGTM, Thank you

PR #11842, opened 2023-08-20
WTrackmenu: use moveToTrash() with Qt >= 5.15

ronso0:


src/widget/wtrackmenu.cpp
```diff
@@ -2090,19 +2131,35 @@ void WTrackMenu::slotRemoveFromDisk() {
...
             msgTitle = tr("Track Files Deleted");
+#else
+            msgTitle = tr("Track Files Moved To Trash");
+#endif
             msgText =
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+                    tr("%1 track files were moved to trash and purged "
+                       "from the Mixxx database.")
+#else
                     tr("%1 track files were deleted from disk and purged "
```
Swiftb0y, line 2144: shouldn't this also follow the dynamic plural forms?

```suggestion
                    tr("%n track file(s) were moved to trash and purged "
                       "from the Mixxx database.")
#else
                    tr("%n track file(s) were deleted from disk and purged "
```

Swiftb0y (commented):
LGTM otherwise.

src/widget/wtrackmenu.cpp
ronso0: I just didn't want to touch existing tr strings to not lose the translations short before the release. Considering we drop support for Ubuntu 20.04 and can also switch to Qt 5.15 on macOS for Mixxx 2.5, I think the tr inconsistencies are okay.

Swiftb0y: fair.

Swiftb0y (approved):
LGTM

Swiftb0y:
Thank you

PR #11848, opened 2023-08-21
Slip mode: consider loop for background position only if it was enabled  before slip

### LINKED ISSUE #11844: Slip mode handles loops not correct
### Bug Description

- play a track
- engage slip mode
- set and enable a new loop 
- stop the loop after some iterations
- disable slip mode

=> Play position should jump to the position where the track would be without the loop - this is the case with 2.3 but not with 2.4

see discussion at:
https://mixxx.zulipchat.com/#narrow/stream/109122-general/topic/Slip.20mode/near/386262097

### Version

2.4beta

### OS

Windows 11

ronso0:
This restores 2.3 behaviour (broken by c5127eb):
* enable slip mode (no active loop)
* active loop (NOT rolling loop since that would quit slip mode when disabling it)
* let it do a few interations so the slip position and the background position can be clearly distinguished
* disable loop
* quit slip mode = position is where it would have been without slip mode

Closes #11844 

@JoergAtGithub please test!

### DIFF
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -49,6 +49,7 @@ LoopingControl::LoopingControl(const QString& group,
         : EngineControl(group, pConfig),
           m_bLoopingEnabled(false),
           m_bLoopRollActive(false),
+          m_bLoopWasEnabledBeforeSlipEnable(false),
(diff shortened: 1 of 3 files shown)


ronso0:
We could also add a test for this situation.

JoergAtGithub:
This fixes the bug! Thank you!

PR #10999, opened 2022-10-24
Full Ableton Link support

JoergAtGithub:
Based on the old PR #3783 from @spensbot, I created a new PR with Ableton Link support for Main.
grafik
The LINK-Button enables/disables the Link session. The number right to it shows the number of connected Link peers.
It syncronizes Tempo and Beat-Position (Phase).

In Ableton Link all peers are equal, there is no leader. Therefore syncronization is always bidirectional.

In Mixxx, it syncs the Sync-Leader (the deck with SYNC and Crown-Button active) with the other peers.

The most important part of this PR is the hosttime filter implemetation, which provides a jitter free timestamp for each audio buffer start - this timestamp is in host time (computer clock not soundcard clock) which is the needed input for the Ableton Link library.

What does not always work automatical, is to determine the latency of the Sound IO. This is because portaudio does not report reliable latency information for all sound apis. On my system it works well for ASIO API, but not so well for the others. Therefore I added a manual latency compensation setting in the preferences, to allow the user to correct the latency:
grafik

I tested this with the official LinkHut-App of the Ableton Link package and with the Android App "G-Stomper Rythm", which is a free Drummachine, but with advertizing. A very cool Ableton Link software is also the music visualisation and lighting software from spensbot: https://captivatesynth.com/

**Test Plan**
Any Ableton Link software, should pass the following test plan:
https://github.com/Ableton/link/blob/master/TEST-PLAN.md

To install the official Linkhut reference tools used in the test plan, you can use VCPKG:  `./vcpkg install ableton[hut,hutsilent]`

Closes #8650

PR #11850, opened 2023-08-21
Refactor timers (#11807) 2nd try

Swiftb0y:
So, this is #11807 reopened after I reset it out of the branch. Unfortunately this PR does suffer a significant issue: Due to Qt API limitations, `PerformanceTimer::difference` can not provide the same resolution as previously. 

Fortunately, the only real consumer of that function is `VSyncThread::fromTimerToNextSyncMicros`. Unfortunately, the resolution provided by Qt (Milliseconds) does not suffice in the context of frame timings:
https://github.com/mixxxdj/mixxx/blob/dfde28dbe44375706abcf8846e7fd099c1c66a87/src/waveform/vsyncthread.cpp#L120-L124

I'd prefer to just get rid of `PerformanceTimer::difference` since it doesn't really have much to do with Timers anyways and just use `std::chrono::time_point` in the `VSyncThread` with the appropriate `std::chrono` clock. The alternative would be to drop the QElapsedTimer commit and just use the `std::chrono`-based implementation of `PerformanceTimer`. wdyt? @daschuer 

Marking PR as draft until we discussed a solution and the PR is ready to be merged.


JoergAtGithub:
I've somewhere a branch where I changed VSyncThread::fromTimerToNextSyncMicros to an absolute time reference. This absolute time reference was needed in the sounddevice implementation anyway, because Abeton-Link needs a jitter-free absolute clock reference. And the filtering to get this jitter-free clock requires an absolute monotonic clock.

Swiftb0y:
great. could you try to dig that up? Then we can merge it independently as a per-requisite.

JoergAtGithub:
I found it: https://github.com/JoergAtGithub/mixxx/tree/visualPlaypositionAbsoluteTime
The clock filtering code itself is more refined in #10999.

Swiftb0y:
Mhmm okay, so what's your preferred plan of action? Just implement a more basic `std::chrono`-based mechanism in the vsync branch independent of your visualPlaypositionAbsoluteTime branch?

JoergAtGithub:
The Ableton Link code uses the following platform specific monotonic clock implementations:
https://github.com/Ableton/link/blob/master/include/ableton/platforms/darwin/Clock.hpp
https://github.com/Ableton/link/blob/master/include/ableton/platforms/linux/Clock.hpp
https://github.com/Ableton/link/blob/master/include/ableton/platforms/windows/Clock.hpp
My visualPlaypositionAbsoluteTime branch used these as clock reference. In general this code should work with any monotonic clock reference, which retruns `std::chrono::microseconds`.

Swiftb0y:
right, so lets just trust the stdlib implementors and use `std::chrono::steady_clock`. I don't like the interface of the ableton clock implementations.

daschuer:
I am confused. From https://doc.qt.io/qt-6/qelapsedtimer.html#nsecsElapsed It looks like that QElapsedTimer gives us a nanoseconds resolution if possible. 

Before this commit, Qt uses the desired low level API as before:
https://github.com/qt/qtbase/commit/c9f4c0d45e85c7cd7c6f7c18c3f3391857135ce2#diff-8523988969292961dc389bb18a73fbe5e2328ee0fe49798c14b61850ed270c97

Does this involve a regression?

daschuer:
I think Yes :-( it uses the "std::chrono::steady_clock" which is know to may have a low resolution: https://stackoverflow.com/questions/29551126/does-steady-clock-have-only-10ms-resolution-on-windows-cygwin
std::chrono::high_resolution_clock seems to be the better choice. But we don't have a guaranteed monotonic and overflow behavior.  

Such high level APIs break at any time just like the QElapsedTimer in QT 6.6 because we rely on undefined behavior. 

What was the issue with our original Implementation? Can we just revert to the original low level API call with guaranteed behaviors and fix the issues there? 

Alternative we may copy the QT < 6.6 implementation to our source tree and extend it by a "nsecsTo" or "durationTo()"

Swiftb0y:
> I am confused. From https://doc.qt.io/qt-6/qelapsedtimer.html#nsecsElapsed It looks like that QElapsedTimer gives us a nanoseconds resolution if possible.

Yes, but it only gives us the elapsed time at the time of calling. In other words, it only gives us the duration between when the timer was started and at the time of calling `nsecsElapsed`. `PerformanceTimer::difference(rhs)` instead gives us the difference between when `this` was started and `rhs` was started. We can not implement that ourselves because `QElapsedTimer` does not let us access the start time. Instead `QElapsedTimer::msecsTo` implements that, but only with millisecond precision for some reason. This gives us two/three options:
1. Use our own timer implementation, not bound by the API limitations of `QElapsedTimer`.
2. remove `PerformanceTimer::difference`
3. implement `QElapsedTimer::nsecsTo` upstream. Useless to us right now because it would only be picked up by Qt6 anyways.

Right now, I'm currently opting for 2. which entails a `std::chrono` based rewrite of vsyncthread...

Swiftb0y:
> What was the issue with our original Implementation? Can we just revert to the original low level API call with guaranteed behaviors and fix the issues there?

Nothing much, I just wanted to clean it up because nobody has looked at it since the last 10 years (and probably won't for another 10).

> I think Yes :-( it uses the "std::chrono::steady_clock" which is know to may have a low resolution: https://stackoverflow.com/questions/29551126/does-steady-clock-have-only-10ms-resolution-on-windows-cygwin

thats unfortunate, but in proper implementation this should be checkable statically: `static_assert(std::ratio_greater_equal_v<std::chrono::steady::clock::period, std::micro>);

> Such high level APIs break at any time just like the QElapsedTimer in QT 6.6 because we rely on undefined behavior.

Can you elaborate on the undefined behavior you are eluding to?

daschuer:
> This gives us two/three options:

1. I am in favor for our own timer using the low level API, because this way we have a guaranteed behavior at build time. 
2. We need `difference()` aka `durationTo()` because all other ns precision functions requires 2 x now()
3. Fix QTs 6.6 QElapsedTimer would be a nice project, to make Mixxx needs visible. There are some precision issues to fix, but since they pick std::chrono::steady_clock the use case is probably not the high precision we want. I don't see an issue for waiting, because we can copy the files to our source tree like before. 

> Can you elaborate on the undefined behavior you are eluding to? 

The standard doe not define a resolution nor an overflow behavior. Yes, it provides a API to peek it like you proposed, but that need to be done at runtime, because it depends on the OS and std implementation. 
I am worrying about std::chrono clocks, because there is no `high_resolution_clock` that is also guaranteed monotonic. 

Given that we have a reliable low level API code, it does not make much sense to give away control to glue layers and than take the efforts to investigate if the implementation suits our needs. This is too much work IMHO but can of cause be done.

So I would either tidy our own established PerformanceTimer or copy the QElapsedTimer before QT6.6 instead and add the missing function.

Swiftb0y:
> The standard doe not define a resolution nor an overflow behavior. Yes, it provides a API to peek it like you proposed, but that need to be done at runtime, because it depends on the OS and std implementation.

Why can't we do this at compile time?

Swiftb0y:
> We need difference() aka durationTo() because all other ns precision functions requires 2 x now()

not really. could even be kept just as concise with a small helper function. 

>  I am worrying about std::chrono clocks, because there is no high_resolution_clock that is also guaranteed monotonic.

Right, but `high_resolution_clock` is monotonic on windows and `steady_clock` is high resolution enough on macOS and linux (which can be asserted at compile time).

Swiftb0y:
> We need difference() aka durationTo() because all other ns precision functions requires 2 x now()

daschuer:
> Why can't we do this at compile time? 

Because we use the same binary in different OS versions. 
The used functions are not constexpr. Sure, we can look up implementatiin detail and assume behavior and than assert them at runtime. It's just that when have an established solution already using the low level API. 

> steady_clock is high resolution enough 

What is the resolution on Linux. What was the resolution of our low level implementation. I don't think it is reasonable to pick a timer with a lower resolution and a glue layer that shifts the overflow artificialy. We should aim for the fastest solution with maximum precision. 

I still don't like "difference()'  Qt 6.6 has durationTo() which fulfils the same use case. Maybe we can introduce that. 

PerformanceTimer is nothing else than a timestamp. It can be passed around as you like. durationTo() is just a renamed operator-() implementation.

Swiftb0y:
Okay, so lets come to a compromise. We keep the previous low-level implementation but implement the high-level interface using `std::chrono` primitives. 

> PerformanceTimer is nothing else than a timestamp. It can be passed around as you like. durationTo() is just a renamed operator-() implementation.

Under the hood its a timestamp yes, but it does not have the value semantics you would expect from a timestamp. As such, we should avoid passing it around as such!

> What is the resolution on Linux. What was the resolution of our low level implementation. 

Let me investigate, I doubt there is any difference with libstdc++'s timer implementation.

Swiftb0y:
This is my crude test, suggesting that the difference between `steady_clock` and `high_resolution_clock` is none under linux (I assume thats what gcc runs on on compiler explorer). https://compiler-explorer.com/z/ex6o3srTY

Swiftb0y:
> I think we see that we have ns resolution in both cases.

Right... so, using chrono clocks instead of the lowlevel ones is an option after all?

>  That's why std::chrono::time_point exists passing that around is equivalent to our original solution with difference() or the Qts durationTo() So I am afraid we do not get around that or similar.

Right, my point is that we should not compare and pass around Timers but instead `std::chrono::time_point`s as those have explicit value semantics. The solutions are equivalent in practice, but abusing Timers as `time_stamp`s is more brittle (since timers can be invalid when they haven't been started yet).

Swiftb0y:
fyi, digging deeper in the low-level APIs, there is no guarantee over the precision of timers on macOS either as far as I can tell... I still think we just should use `std::chrono::steady_clock`. If we need to ensure some precision at runtime, lets just port the previous demo code to a unit test... Yes the precision is undefined, but its not the same UB where the compiler will break our code if we rely on it. Platforms which have a monotonic clock whose resolution is too low, we simply can't support (as already the case with the current PerformanceTimer implementation, evident by the dummy implementation). https://github.com/mixxxdj/mixxx/blob/dfde28dbe44375706abcf8846e7fd099c1c66a87/src/util/performancetimer.cpp#L272-L292

daschuer:
> Right... so, using chrono clocks instead of the lowlevel ones is an option after all?

What are your reasons to use it?

> The solutions are equivalent in practice, but abusing Timers as time_stamps is more brittle (since timers can be invalid when they haven't been started yet).

Since our PerformanceTimer is basically a time stamp I do not see an abuse here. But of cause we can warp the timestamp part of PerformanceTime into a PerformanceTimeStamp class or such and pass that around instead.  

> std::chrono::time_points as those have explicit value semantics.

I don't get that point. You need always build the difference to get a human readable value. That is exactly the same with
PerformanceTimer::difference().  

We have good experience with the low level timers and no reported issues with it. So we need some good reasons to replace that with a additional abstraction layer that takes more cpu time, developer and review resources and introduces additional dependencies we can't control.

Swiftb0y:
> What are your reasons to use it?

Less code, less maintenance. Makes it easier to benefit from the modern abstractions. 

> Since our PerformanceTimer is basically a time stamp I do not see an abuse here. But of cause we can warp the timestamp part of PerformanceTime into a PerformanceTimeStamp class or such and pass that around instead.

Yes, I'd be more okay with a method that extracts a `chrono::time_stamp` from a timer which can then be passed around.

daschuer:
>> What are your reasons to use it?

> Less code, less maintenance. Makes it easier to benefit from the modern abstractions.

Don't touching the code without an issue is the minimum of maintenance.  

We don't make use of the modern abstraction or have any benefit or demands.

I think creating `chrono::time_stamp` from get_time() is exactly what std::chrono::steady_clock doe under the hood. So there is no reason to code it. Unfortunately this does not comes without any cost. But in our case there is no reason for a common abstraction layer, so lets not spend this cost. 

If you like to use the modern abstraction in places where precision and performance is not a curlpit, you can already use plain QElapsedTimer or the Chrono functions. There is no reason to wrap them. I consider using a Mixxx function more dificult, than a QT or Std API function 

Did you consider to spit out PerformanceTimeStamp form PerformanceTimer? For my understanding that targets you concern the best. I fully agree that passing something around that is called "Timer" looks odd. 

Something like this: 

````
class PerformanceTimeStamp
{
public:
    PerformanceTimerStamp() {
      t1 = 0;
#if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
      t2 = 0;
#endif
    };

    mixxx::Duration PerformanceTimeStamp::operator-(const PerformanceTimeStamp& time_stamp) const {
#if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
        return mixxx::Duration::fromNanos((t1 - time_stamp.t1) * Q_INT64_C(1000000000) + t2 - time_stamp.t2);
#else
        return mixxx::Duration::fromNanos(absoluteToNSecs(t1 - time_stamp.t1));
#endif
    }

private:
    qint64 t1;
#if defined(Q_OS_UNIX) && !defined(Q_OS_MAC)
    qint64 t2;
#endif
};

For the rest we can use QElapsedTimer directly.

**Comment by Swiftb0y** [comment]
> I think creating chrono::time_stamp from get_time() is exactly what std::chrono::steady_clock doe under the hood. So there is no reason to code it. Unfortunately this does not comes without any cost.

What cost?

> Did you consider to spit out PerformanceTimeStamp form PerformanceTimer?

Why not use `std::chrono::time_stamp` instead of reinventing the wheel?

**Comment by Swiftb0y** [comment]
Please have a look at the most recent commit. It implements uses the high res clock of the system if its known to be monotonic and if not a custom clock using the low level old low-level primitives. The result is much less and easier to understand code than previously.

**Comment by Swiftb0y** [comment]
Why does the linux timer need two qint64's anyways? $2^64$ nanoseconds is more than 500 years!

**Comment by daschuer** [comment]
The costs for the abstraction can be seen in my compiler explorer:
https://compiler-explorer.com/z/1PnaTqWGE

I think you implementation is head-up with the libstdc++:
 
https://github.com/gcc-mirror/gcc/blob/92f2ec417c57e980b92b8966226fc2bfbf042af8/libstdc%2B%2B-v3/src/c%2B%2B11/chrono.cc#L79

IMHO, it is obvious that the cpu penalty come form the conversion to std::chrono::duration instead of keeping the native representation as time stamp, which the std::chrono::time_point template allows.   

I think if we decide to keep the low level implementation of the timestamps, to be independent form weak definitions and compiler libraries we should also take the benefit to keep the low level Duration for differences. 

Does that make sense?

**Comment by Swiftb0y** [comment]
> IMHO, it is obvious that the cpu penalty come form the conversion to std::chrono::duration instead of keeping the native representation as time stamp, which the std::chrono::time_point template allows.

How would that work with time_point? Its only constructible from a duration.

Also, I think the overhead from the duration conversion is negligible. If the stdlib implementation has it, its either negligible or unavoidable.

**Comment by Swiftb0y** [comment]
> I think if we decide to keep the low level implementation of the timestamps, to be independent form weak definitions and compiler libraries we should also take the benefit to keep the low level Duration for differences.

Please elaborate. what weak definitions? what low-level duration?

**Comment by daschuer** [comment]
Ok I see, my idea does not work and will pollute the std:: namespace. So the own time_stamp is back on the table. If we give it the same interface than `std::chrono::time_stamp` it will work. The prototype is outlined above as PerformanceTimeStamp. 

> what weak definitions?

I mean the degree of details about the chrono timer definitions in the c++ standard as already discussed above. 

> what low-level duration?

Sore what we get from the low level API, as shown in PerformanceTimeStamp. 

To convert a `std::chrono::time_stamp` to a duration, you always need to make a difference of two of them. 
That's why I consider it not important which type or namespace it has.

It can even have a mixxx::Duration and std::chrono::Duration result if you like.

**Comment by Swiftb0y** [comment]
> I mean the degree of details about the chrono timer definitions in the c++ standard as already discussed above.

Right, so the idea implemented in the latest commit is that we sidestep that by implementing our own chrono-timer-like timer, but with the low-level APIs which give us the guarantees we want. 

> Store what we get from the low level API, as shown in PerformanceTimeStamp.

But why is that necessary? Why can't we just store a `std::chrono::time_stamp`?

> To convert a std::chrono::time_stamp to a duration, you always need to make a difference of two of them.

Not really, a `std::chrono::time_stamp` is just a duration since the clocks epoch. Thats why you can construct a `time_stamp` from a `duration` and also retrieve the duration again with `time_stamp::time_since_epoch`.

**Comment by daschuer** [comment]
> But why is that necessary? Why can't we just store a std::chrono::time_stamp?

To have the fastest measurement possible where nanoseconds are important and also to have no regression compared to the PerformanceTimer. 

The issue is that I am not convinced that it is worth the efforts, compared to a vendored QElapsetTimer < Qt6.6.
But since you seems to be eager to do that task, there is no reason to stop you, if the new solution performs the same or better.  

It seams you are looking for a solution for all. For me, it would be fine to use the low level API for sub-seconds timer and the plain QElapsetTimer for the rest. This would reduce the interface requirement. 
If you like to keep the joint solution, you can easily add a conversion from the low level time stamp to std::chrono::time_stamp.

**Comment by Swiftb0y** [comment]
> To have the fastest measurement possible where nanoseconds are important and also to have no regression compared to the PerformanceTimer.

Yes, that is understandable, but I don't understand why a `std::chrono::time_stamp` based implementation would perform worse than a custom implementation by a significant margin. The performance-impact of `absoluteToNSecs` (macOS) and multiply-add (linux) is probably negligible. I will write some benchmarks if that will convince you.

**Comment by Swiftb0y** [comment]
> The issue is that I am not convinced that it is worth the efforts, compared to a vendored QElapsetTimer < Qt6.6.

I'd be fine with that too, but I struggle to find the sources for all platforms tbh.

**Comment by JoergAtGithub** [comment]
I see a big potential in global timestamps for future development. Various standard audio computing solutions like MIDI or Ableton Link rely on a single application-wide monotonic high-resolution clock for a reason.
For example we could use it to get MIDI-Crosfader movement and Vinyl-Control movement during scratching in sync (delay to the speaker will of cause depend on the audio buffer size - but sync is more important than delay).

**Comment by daschuer** [comment]
> I'd be fine with that too, but I struggle to find the sources for all platforms tbh.

Thank you. 

The removed PerformanceTimer contains a solution for all. You may compare it with the removed low level timers in this commit: 
https://github.com/qt/qtbase/commit/c9f4c0d45e85c7cd7c6f7c18c3f3391857135ce2#diff-8523988969292961dc389bb18a73fbe5e2328ee0fe49798c14b61850ed270c97

**Comment by Swiftb0y** [comment]
>> I'd be fine with that too, but I struggle to find the sources for all platforms tbh.
> 
> Thank you.

I've taken a look at the source and I think we have a misunderstanding here. I thought you suggested to vendor the Qt <6.6 ElapsedTimer and replace it with the 6.6 Timer once Qt6.6 is released and we're using it, but thats not what you meant, right?

**Comment by daschuer** [comment]
Our discussion has evolved. 

For my understanding the latest version is this: 

You offered to write a low level API based clock, that uses a custom time stamp object that just carries what received form the low level API. The operator-() shall result in a either a mixxx::Duration or std::chrono::duration (overloaded). The low level implementation for the Mixxx targets can be copied from our old PerformaceTimer or form QElapsedTimer before it has been ripped of in 6.6 

This way we have a fast and precise timer for measuring sub-second durations and due to the std::chrono::duration interface it can be used for everything else as well. 

This does not limit us and plays well with using the other chrono clocks where it is desired an appropriated.  

The alternative we have discussed but dismissed, because of passing the timer around, is to vendor the Qt <6.6 QElapsedTimer as low level ElapsedTimer, add [durationTo()](https://github.com/qt/qtbase/blob/c9f4c0d45e85c7cd7c6f7c18c3f3391857135ce2/src/corelib/kernel/qelapsedtimer.h#L45) and use it right away.

**Comment by Swiftb0y** [comment]
> You offered to write a low level API based clock, that uses a custom time stamp object that just carries what received form the low level API. 

Not quite. I wanted to write a low-level (as high level as possible while maintaining the high-resolution and monotonic requirements) clock (as in `std::chrono::is_clock`). Making the time_stamp be something other than `std::chrono::time_stamp` will violate that. So that's why the custom timestamp class is out of the question. That `chrono`-compatible clock (currently called `HighResolutionMonotonicClock`) would then be used to reimplement PerformanceTimer. 

The only point of contention I currently see is that you the std::chrono::time_stamp` is out of the question because that requires two instructions of integer math on every call. I'm of the opinion that these are negligible while these seem to be a dealbreaker to you. In order to settle this debate, I'd be willing to write both implementations and benchmark them in order to prove that the "overhead" is negligible. 

My overall goal with this PR is to minimize the SLOC required for PerformanceTimer as much possible while staying C++-standards-compliant. Both of these should result in maximal long-term maintainability.

**Comment by daschuer** [comment]
> Making the time_stamp be something other than std::chrono::time_stamp will violate that. 

I disagree here, because that is the source of the cpu penalty. For me the actual type of time_stamp is not relevant as long it support the interface to enter the chrono world. At least I can not imagine a use case for this. 

You don't need to code both solution to compare the difference. Just look into my compiler explorer links above. I am not going to discuss what is neglecteible and what not, because you are currently proposing to replace a good established and fast solution, only to follow a standard approach. The standard implementation is already there in the libstdc++ and matches you code shown above. 

The question that is interesting for me, how will a timer with Chrono interface fit into Mixxx. Can you outline this plan? Maybe you can use this to demonstrate where the standard chrone based timer will help us.

**Comment by Swiftb0y** [comment]
So, I wrote some basic benchmarks, the result is that the `std::chrono::steady_clock` wins easily... See https://github.com/Swiftb0y/mixxx/tree/refactor/timers-benchmark

```
------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
------------------------------------------------------------------------------------
BM_PerformanceTimerLegacy                       23,5 ns         20,8 ns     35437710
BM_PerformanceTimerChronoStdSteady              15,7 ns         15,6 ns     43555154
BM_PerformanceTimerChronoStdHighRes             15,5 ns         15,5 ns     43911787
BM_PerformanceTimerChronoHighResMonotonic       19,0 ns         18,9 ns     36253810
```

**Comment by daschuer** [comment]
Great results. The higher CPU load is sourced from the additional test that is done on each value read. I think that can be improved. But it is interesting to see that the BM_PerformanceTimerChronoHighResMonotonic outperforms the BM_PerformanceTimerLegacy where both have the checks. Well done! 

So my demand of no regression has been proved.  

I took my performance assumption not from the assembler but from the from the measured times in the output, the first column is the duration of a `std::chrono::high_resolution_clock` lookup and the second column the native call: 

52ns 33
36ns 29
25ns 22
22ns 21
26ns 26
23ns 22
24ns 20
25ns 24
21ns 22
31ns 29

On a second look I see that the comparison is unfair, because the high_resolution_clock variant already has a ns value while the
native call has to deal with 128 bits which is not covered in the measurement.

**Comment by Swiftb0y** [comment]
> The higher CPU load is sourced from the additional test that is done on each value read.

What check, the `unixCheckClockType()`?

So whats your conclusion from the benchmark? My conclusion (without looking at the results on other platforms) is to just use `std::chrono::steady_clock`.

**Comment by Swiftb0y** [comment]
fyi, I `unixCheckClockType` outright (see latest commit on `refactor/timers-benchmark`) and Legacy still underperforms as far as I can tell.

```
BM_PerformanceTimerLegacy                       19,5 ns         19,4 ns     36805899
BM_PerformanceTimerChronoStdSteady              15,7 ns         15,6 ns     44246633
BM_PerformanceTimerChronoStdHighRes             16,0 ns         15,9 ns     40089067
BM_PerformanceTimerChronoHighResMonotonic       16,4 ns         16,4 ns     42771744
```

**Comment by daschuer** [comment]
Can you also remove the extra function call and test again?
It would be nice to have this test always on our CI. So that we have a chance to spot regressions later. 

> My conclusion (without looking at the results on other platforms) is to just use std::chrono::steady_clock.

We have still no guarantee that this clock is also HighRes. In order to check that you, have to investigate one time if the std::chrono::steady_clock has ns resolution. 

We may also assert for: 
`using high_resolution_clock = steady_clock;`
https://github.com/microsoft/STL/blob/c33874c3777f1596f4cecce6c00bdda41a4fc1b0/stl/inc/chrono#L716

They use a smart caching: https://github.com/microsoft/STL/blob/60f18856c56389001c38a9929ef37bd5c01c4e47/stl/src/xtime.cpp#L93

I have tried to investigate the mingw issue but I was not able to find a trace to the reported precision issue. 

It seems that mingw implements its clocks it via the bulky clock_gettime() interface which is a waste of CPU: 
https://github.com/mingw-w64/mingw-w64/blob/d3ab8f62b95653d4875d0191785ccaada4c9e635/mingw-w64-libraries/winpthreads/src/clock.c#L169

So this is the final issue:
The c++ standard gives no guarantee about the performance of the chrono clocks by definition. 
We may either check them or implement them ourself low level. The check is difficult because in case of mingw for example I am afraid the duration's period is ns, but it is advancing in bigger steps. Fortunately we do not use mingw (yet).

What is your idea to target this issue?

**Comment by Swiftb0y** [comment]
> Can you also remove the extra function call and test again?

Which extra call? `do_gettime`? I'm certain thats being inlined but I can test. 

> We have still no guarantee that this clock is also HighRes. In order to check that you, have to investigate one time if the std::chrono::steady_clock has ns resolution.

I'm pretty sure it does on most platforms, but we can add a Unit test that will let us verify that easily in case we do have the suspicion that that's not the case. 

> They use a smart caching: https://github.com/microsoft/STL/blob/60f18856c56389001c38a9929ef37bd5c01c4e47/stl/src/xtime.cpp#L93

I assume you're talking about `freq_cached`? We already do that on windows as well (without the atomic which is admittedly dangerous).

> It seems that mingw implements its clocks it via the bulky clock_gettime() interface which is a waste of CPU:

Right... Low quality implementation, not sure what we can do about that. If its an issue, we could implement a workaround I guess. 

> What is your idea to target this issue?

Use `std::chrono::steady_clock` by default, possibly use the fallback `HighResMonotonic` clock if we experience issues on some platforms. Implement a unit test that easily lets us verify the precision of the `chrono::steady_clock`. wdyt?

**Comment by Swiftb0y** [comment]
> We may also assert for:
> `using high_resolution_clock = steady_clock;`

Whats the need for that? Lets just use `high_resolution_clock` if its steady, otherwise use `steady_clock` and assume it has high-enough resolution. If we notice that that's not the case, we can change the fallback to a platform specific implementation.

**Comment by Swiftb0y** [comment]
>> Can you also remove the extra function call and test again?
>
> Which extra call? do_gettime? I'm certain thats being inlined but I can test.

Interestingly, HighResMonotonic is now up to par with the rest. 
https://github.com/mixxxdj/mixxx/commit/e7cdb7f13312bbf608f3226df5609a18bd09a0aa
```
------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
------------------------------------------------------------------------------------
BM_PerformanceTimerLegacy                       18,5 ns         18,5 ns     36250948
BM_PerformanceTimerChronoStdSteady              16,0 ns         16,0 ns     44175894
BM_PerformanceTimerChronoStdHighRes             15,3 ns         15,2 ns     44951517
BM_PerformanceTimerChronoHighResMonotonic       15,6 ns         15,5 ns     45604557
```

**Comment by Swiftb0y** [comment]
Doesn't change much when applied `PerformanceTimerLegacy`

```
------------------------------------------------------------------------------------
Benchmark                                          Time             CPU   Iterations
------------------------------------------------------------------------------------
BM_PerformanceTimerLegacy                       18,7 ns         18,6 ns     36761869
BM_PerformanceTimerChronoStdSteady              15,3 ns         15,3 ns     41677391
BM_PerformanceTimerChronoStdHighRes             15,5 ns         15,5 ns     44478064
BM_PerformanceTimerChronoHighResMonotonic       15,3 ns         15,2 ns     45262423

```

**Comment by Swiftb0y** [comment]
Also, the chrono timers perform horribly on macOS for some reason... Even `HighResMonotonic`??

```
BM_PerformanceTimerLegacy                       17.9 ns         17.9 ns     39108986
BM_PerformanceTimerChronoStdSteady              43.7 ns         43.6 ns     15890023
BM_PerformanceTimerChronoStdHighRes             43.7 ns         43.6 ns     16024247
BM_PerformanceTimerChronoHighResMonotonic       44.2 ns         44.2 ns     15906163
```

**Comment by daschuer** [comment]
OK how about that 
Linux: ChronoStdSteady
Windows MSVC: ChronoStdSteady
MinGW: break sine unused or reimplement the the MSVC solution
MacOS: Low Lewel, Copy from PerfomanceTimer.

Rational: 
The steady nature is more important
I trust Linux and msvc
Avoid Posix wrapper

**Comment by Swiftb0y** [comment]
>  The steady nature is more important

I agree, I don't think an overhead of 30ns is super important as long as its consistent.

**Comment by Swiftb0y** [comment]
@daschuer, how do we want to proceed? I'm in favor of just using plain `chrono:steady_clock`

**Comment by daschuer** [comment]
How about introducing a chrono:steady_clock for Linux and MacOS here inclusive the benchmarks in our workflows. 
And re-implement the MVC solution in our source to work around issues with other compilers. Than we improve the situation on macOS independent in a follow up PR if we have ideas.

**Comment by Swiftb0y** [comment]
> How about introducing a chrono:steady_clock for Linux and MacOS here inclusive the benchmarks in our workflows.

The goal was that have to maintain less code, adding all the code written for the benchmarks has the opposite effect. What timers/benchmarks in particular would you like to see included?

> And re-implement the MVC solution in our source to work around issues with other compilers. 

Why re-implement the MSVC solution if we can just use it?

> Than we improve the situation on macOS independent in a follow up PR if we have ideas.

I don't think there is much to improve there, otherwise it would've already been done so in the established standard library implementations.

**Comment by daschuer** [comment]
> The goal was that have to maintain less code, adding all the code written for the benchmarks has the opposite effect. What timers/benchmarks in particular would you like to see included?

My idea was to test our vendored solution compared to the chrno timers. This way we can easily spot regressions or compare old compiler libraries to new ones. This just because you took the efforts to write it and it was useful. 

By the way, which timer does benchmark use? Does the longer time come from benchmark? Two now() calls and check for the difference eliminates this dependency. 

> Why re-implement the MSVC solution if we can just use it?

The idea was to have a solution that works with non MVC builds as well.  We are not even sure in which MSV version it is used. The benchmark would help to detect it. 

The alternative would be to check for it. But what is the solution for the else branch? If we have a vendored solution anyway, we can use it for all windows targets without tests. 

> I don't think there is much to improve there, otherwise it would've already been done so in the established standard library implementations.

We will see. If the POSIX wrapper is used like in mingw, there is something to improve.

**Comment by JoergAtGithub** [comment]
> > 
> 
> We will see. If the POSIX wrapper is used like in mingw, there is something to improve.

This is not a supported platform. If somebody want to add mingw support, he has to implement it - as many other things too. 

BTW: MinGW is a really ugly piece of code, neither POSIX nor Windows. It's a maintenance nightmare, that introduces countless workarounds into any codebase that uses it.

**Comment by Swiftb0y** [comment]
> We will see. If the POSIX wrapper is used like in mingw, there is something to improve.

It isn't using that though. It uses the exact same code as we do with `mach_absolute_time`, which is why it had the same performance as HighResMonotonicClock. Looking for the implementation in libstdc++ and libcxx confirms this. Moreover, I also tried the posix wrapper, which results in the same performance (https://github.com/mixxxdj/mixxx/commit/e0d809c98fb2481f0d83349642b60b8f6912470f https://github.com/Swiftb0y/mixxx/actions/runs/5978076166/job/16219428130)

> My idea was to test our vendored solution compared to the chrno timers. This way we can easily spot regressions or compare old compiler libraries to new ones. This just because you took the efforts to write it and it was useful.

Yes, I took the time to write it but not because I wanted to include it but because I wanted to prove the point that we should stop having these discussions and trust the standard library.

> By the way, which timer does benchmark use? Does the longer time come from benchmark? Two now() calls and check for the difference eliminates this dependency.

idk which timer it uses. It does use `mach_absolute_time` for counting cpu cycles though. I'm pretty sure the longer time doesn't come from benchmark. It would be hideous if a microbenchmarking library would be implemented in a way where it skews the results in these micro scales, that's what gbenchmark is designed for. 

> The idea was to have a solution that works with non MVC builds as well. We are not even sure in which MSV version it is used. The benchmark would help to detect it.

Thats exactly what we get by using `std::chrono`. Any environment we want to run on will have to ship a C++20 compiler. Any C++11 and up compiler will have an implementation of `steady_clock`. If we really need something custom for a different platform, we can implement that when needed. Using the chrono clock API makes that trivial (as demonstrated with the Demo `HighResMonotonicClock` implementation).

File: src/util/duration.h [hunk 1]
```diff
@@ -154,6 +154,15 @@ class Duration : public DurationBase {
         return Duration(nanos);
     }
 
+    static constexpr Duration fromStdDuration(
+            std::chrono::duration<std::int64_t, std::nano> duration) {
+        return Duration::fromNanos(duration.count());
+    }
```
daschuer, line 160: I think this should become a template, to allow any durations with any Period

src/util/performancetimer.h
```diff
@@ -1,75 +1,39 @@
...
-#include <QtCore/qglobal.h>
+class PerformanceTimer {
+  public:
+    // TODO: make this configurable via a template parameter?
+    // note that the resolution of std::chrono::steady_clock is not guaranteed
+    // to be high resolution, but it is guaranteed to be monotonic.
+    // However, on all major platforms, it is high resolution enough.
+    using ClockT = std::chrono::steady_clock;
+    PerformanceTimer()
+            : m_startTime(ClockT::duration(0)){};
```
daschuer, line 14: Redundant
```suggestion
```

src/util/performancetimer.h
```diff
@@ -1,75 +1,39 @@
...
-// To fix interface changes issues in different QT versions
-// Added restart() function.
-// Returns time in nanosecond resolution.
-//
+#include "util/duration.h"
 
-#include <QtCore/qglobal.h>
+class PerformanceTimer {
+  public:
+    // TODO: make this configurable via a template parameter?
```
daschuer, line 8: Don't know if that is a good idea. At least not a todo.

src/util/performancetimer.h
```diff
@@ -1,75 +1,39 @@
...
+        return dur;
     };
 
-    void start();
-    mixxx::Duration elapsed() const;
-    mixxx::Duration restart();
-    mixxx::Duration difference(const PerformanceTimer& timer) const;
+    mixxx::Duration difference(const PerformanceTimer& timer) const {
+        return mixxx::Duration::fromStdDuration(m_startTime - timer.m_startTime);
+    };
```
daschuer, line 32: What is your plan with this function? I think getting rid of it was the source of our long discussion. Now it is still there?  

Right now `std::chrono::time_point()` is a private implementation detail which can be easily replaced with the native faster representation.

daschuer (changes requested):

src/util/duration.h
Swiftb0y: That would only really result in the duration_cast being hidden from the caller. In this case I chose the simplest solution that would result in the simplest codegen. So yeah, making this a template would not be hard, but it would essentially result in an implicit conversion based on the period, which I thought we might prefer if it was explicit. The code in the caller wouldn't be very complicated either when the duration_cast is explicit thanks to function template argument deduction...

So I won't object to making the cast implicit, I think its just better if it isn't since it keeps the code a little simpler and possibly more performant (since the casts always result in a multiplication or even worse, division).

src/util/performancetimer.h
Swiftb0y: Yes, technically redundant, but I wanted to make it explicit as well since we're relying on the fact thats its 0 in `PerformanceTimer::running`. Do you insist in the removal?

src/util/performancetimer.h
Swiftb0y: Well, it would be trivial and if we ever need to swap out the clock implementation using ifdefs, it'd prefer if it was done by `#ifdef`ing `using PerformanceTimer = PerformanceTimerGeneric</* clock impl */>` instead of in the class body. That's purely a style choice however. So lets KISS and just remove the comment?

src/util/performancetimer.h
Swiftb0y: > What is your plan with this function? I think getting rid of it was the source of our long discussion. Now it is still there?

Yes, I want to get rid of and I have a commit flying around somewhere that does that but I didn't want to blow up the PR more than it already is. Step 1 was to simplify the internals, step 2 (separate PR) would be to change the API from `mixxx::Duration` to `chrono::duration`, as well as getting rid of `PerformanceTimer::difference`.

Swiftb0y:
> The remaining point is how to protect against weak implementations out there. Since you have dismissed the idea of benchmarks and vendoring the mvc solution we need IMHO an alternative against the known issues from mingw and friends. My idea is to detect that early.

Let me clarify, I dismissed the inclusion of benchmarks of the legacy timer (since I want that vendored legacy code out). I have nothing against a small benchmark that just checks for the speed of `ClockT::now` and the duration resolution which we can run in case we suspect those are at fault. I don't want to keep the vendored legacy timer just for the sake of benchmarking it against chrono clocks, but just benchmarking chrono clocks (as long as the benchmarks stay simple enough) is fine for me.

Swiftb0y:
> BM_PerformanceTimerLegacy is the original implementation where the time is stored in the OS provided format.
I am not sure if it actually works, but I think that can be achieved by creating inherit from chono::duration to create a custom time with period that can be changes at runtime (boot-up) like TickDuration and than use it in.
std::chrono::time_point as class Duration = typename TickDuration

That... could work, but I'm not sure, afaik inheriting from stdlib types is not a good idea generally, even if not used polymorphically. Another solution could be a runtime std::ratio, that templates the `duration` and `time_point`, but I'm not sure if that's possible. Another solution I also want to experiment with is whether the division is really required on MacOS to get the time, as I've read somewhere that the returned ratio is usually 1 on most CPUs anyways. Also see https://developer.apple.com/documentation/kernel/1462446-mach_absolute_time#discussion

src/util/performancetimer.h
daschuer: OK

daschuer:
Yes, Interesting, but anyway a topic for a follow up PR. 
Do you want to keep the PerformanceTimerChronoHighResMonotonic as a playground for this?  

Which tests do we need to sort out mingw like issues? A unitTest that measures two now() calls like in our original Godolt?

Something like 
```
t1 = now()
t2 = now()
loop of < 5 ms 
t3 = now() 
```
The time difference between t1 and t2 and t2 and t3 version must not >= 10 ms to sort out timers with a 10 ms or bigger steps.

Swiftb0y:
>  Another solution I also want to experiment with is whether the division is really required on MacOS to get the time, as I've read somewhere that the returned ratio is usually 1 on most CPUs anyways.

Intel CPU's seem to use a ratio of 1, but ARM Macs seem to have a different ratio, so it seems it's not possible to avoid the division. https://eclecticlight.co/2020/11/27/inside-m1-macs-time-and-logs/
Also, `clock_gettime_nsec_np(CLOCK_UPTIME_RAW)` is slightly faster than the default `std::chrono::steady_clock`, so it might be worth experimenting with in the future. It's probably not used by most standard libraries because its only barely faster and probably less broadly supported. 

```
BM_PerformanceTimerChronoStdSteady              49.0 ns         49.0 ns     14892255
BM_PerformanceTimerChronoStdHighRes             52.2 ns         52.1 ns     10000000
BM_PerformanceTimerChronoHighResMonotonic       46.4 ns         46.3 ns     15350776
```

Swiftb0y:
> Do you want to keep the PerformanceTimerChronoHighResMonotonic as a playground for this?

Not really, I'll keep it in the experimental `refactor/timers-benchmark` for now but lets not include it in this PR (that's why I rebased those commits out yesterday). 

> Which tests do we need to sort out mingw like issues? A unitTest that measures two now() calls like in our original Godolt?

Not sure, lets just keep a benchmark for manual verification or a non-fatal test. Otherwise I'm afraid the test will get flaky really fast. I already had a problem with the fact that the timer seems to be much lower resolution (or just slower in general) in the unit tests run by coveralls...

Swiftb0y:
> The time difference between t1 and t2 and t2 and t3 version must not >= 10 ns to sort out timers with a 10 ms or bigger steps.

I don't quite understand the suggestion (first of all because I'm not sure the units are correct and second because I don't understand the logic). 
My approach would be to spin a couple times in a loop, collecting the durations of consecutive `chrono::clock::now` calls and then making sure we have no durations of zero length (to rule out aliasing) and making sure the smallest duration is larger than our cutoff (lets say 1 millisecond?). The smallest duration is chosen to avoid the test getting flaky due to outliers and because we want to assert the minimum bound of the clock resolution.

src/util/duration.h
Swiftb0y: fyi, the duration cast is not even necessary, and thus the template. I wish it wasn't but it is, so you get your easy of use and I get the simple function signature. https://compiler-explorer.com/z/bbEb16zvs

daschuer:
My suggestion is almost the same, but with a CPU burning loop in between from a known short duration. This loop must be longer than our minimum duration (idea 5 ms). This way we have 
t2 - t1 = 20 ns (just the minimum we are Abe t measure) 
and
t4 - t3 = 25 ns (the minimum + the loop) 

If this is also 20 ns we, we can fail the test, because our timer is not able to measure the loop duration.
If this is 40 ns we can also fail the test because this is an indicator for a 20 ns resolution which is too coarse.

daschuer:
The actual duration of the for() loop might be different on our target hardwares as well as the timer durations. 
We will learn quickly which boundaries suits to us. 

Maybe we need one dummy now() call in advance to populate the cached multipliers and divisors.

Swiftb0y:
> My suggestion is almost the same, but with a CPU burning loop in between from a known short duration. This loop must be longer than our minimum duration (idea 5 ms). This way we have t2 - t1 = 20 ns (just the minimum we are Abe t measure) and t4 - t3 = 25 ns (the minimum + the loop)
> 
> If this is also 20 ns we, we can fail the test, because our timer is not able to measure the loop duration. If this is 40 ns we can also fail the test because this is an indicator for a 20 ns resolution which is too coarse.

I still don't get it. What duration should be loop be? 5ms or 5ns? What resolution do you expect of the timer? It doesn't make sense of expect a resolution in the single digit nanosecond range if even getting the timestamp takes >20ns.

daschuer:
Just as info: 
found the probably fastest timer: 
https://github.com/petersenna/rdtscbench/blob/d5b99e7d70045361b870e8634eef20a48c68bb77/rdtscbench/rdtscbench.c#L65
https://en.wikipedia.org/wiki/Time_Stamp_Counter
Unfortunately depending on he frequency stepping of the CPU.

Swiftb0y:
How would ensure that the spinloop spins for a certain duration in a portable way?

daschuer:
> How would ensure that the spinloop spins for a certain duration in a portable way?

There is no guarantee just experience. My idea was to pick a loop that has a of > 1 ns duration, but < 10 ns on all platforms.

JoergAtGithub:
I doubt that this will work on modern CPUs with power management - they vary the clock speed depending on CPU load and core temperature.

daschuer:
It is only a question of the picked boundaries.

Swiftb0y:
> I doubt that this will work on modern CPUs with power management - they vary the clock speed depending on CPU load and core temperature.

I agree. That test will get brittle really fast. I've already observed in the timers-benchmark branch that the [median duration of subsequent `now()` calls is 200 nanoseconds!!!](https://github.com/Swiftb0y/mixxx/actions/runs/6014180513/job/16313306457) Thats why I set the minimum resolution the initial benchmark to such a high bound. 

That said, I did spend/waste some time on my only attempt at implement the benchmark you suggested in a robust manner. I'll push the PoC later.

JoergAtGithub (approved):

daschuer (approved):
Looks good now. 
Before merge you nee to remove the Draft state.

Swiftb0y:
ofcourse, ancient AppleClang has no C++20 ranges support...

Swiftb0y:
all green, no draft. lets go

daschuer (approved):
Thank you. LGTM.

Swiftb0y:
Thank you. And lets please reduce the arguing over unnecessary features.

Swiftb0y:
I will also take the liberty to remove the steady_clock test if it turns out to be brittle.

PR #11851, opened 2023-08-21
fix(library): fix `PlaylistDAO::removeHiddenTracks` performance

### LINKED ISSUE #11724: UI lags a long time when loading a rather big playlist
### Bug Description

Loading a 3k items playlist takes roughly a minute.
Meanwhile, UI is stuck and various controls like bas/mid/tre knobs do not respond as long as the UI is stuck.
2.3.x used to load these playlists within a second.
Did a tail -f ~/.mixxx/mixxx.log while loading a playlist and nothing came out, and no excessive cpu usage observed.

Setup used :
- rpi4 (4GB) + 15.6 hdmi touch screen
- denon mc4000
- 3k items playlist (1k is enough to reproduce)

### Version

2.4-beta

### OS

DietPiOS (debian bullseye based)

Swiftb0y:
fixes #11724

The query consists of an outer query and inner subquery. In the previous form, the inner query had a dependency on `p1.playlist_id` of the outer query. Which could change for each row of the outer select statement. This made the inner query a `CORRELATED LIST SUBQUERY`, which caused the subquery to be re-evaluated for each row of the outer query. This dependency however is unnecessary because of the `AND p1.playlist_id=:id` restriction in the outer query. So in theory, `p1.playlist_id` never changes. Unfortunately, the SQLite was not able to make that deduction which lead to the dependency. Manually specifying that the `p1.playlist_id` is constant for the entire subquery, removes the dependecy to the outer query, resulting in a plain `LIST SUBQUERY` which can be evaluated once and then reused for all rows of the outer subquery. This results in reduced algorithmic complexity and thus drastically improved performance for large tables.

### DIFF
--- a/src/library/dao/playlistdao.cpp
+++ b/src/library/dao/playlistdao.cpp
@@ -562,7 +562,7 @@ void PlaylistDAO::removeHiddenTracks(const int playlistId) {
             "WHERE p1.id NOT IN ("
             "SELECT p2.id FROM PlaylistTracks AS p2 "
             "INNER JOIN library ON library.id=p2.track_id "
-            "WHERE p2.playlist_id=p1.playlist_id "
+            "WHERE p2.playlist_id=:id "
             "AND library.mixxx_deleted=0) "
             "AND p1.playlist_id=:id"));
     query.bindValue(":id", playlistId);


ronso0:
Just to get an idea of the impact of this tiny fix:
opening a playlists with 100 and 3000 tracks takes
|branch|100|3000|
|------|--------|--------|
|2.4| 60 ms|13384 ms|
|this| 12 ms|30 ms|

:tada: 

As mentioned in #11724 I can't tell whether that query actually does what it should since the playlists seem to be cleaned up during startup already (and i don't rescan the collection on startup).

ronso0 (approved):

ronso0:
LGTM, thanks a lot!

Swiftb0y:
> Just to get an idea of the impact of this tiny fix:
opening a playlists with 100 and 3000 tracks takes

The effect of algorithmic complexity in action ;)


> LGTM, thanks a lot!

Thank even more for testing and figuring out what line and commit caused the regression. That cut the time required ten fold.

PR #11852, opened 2023-08-21
BaseExternalLibraryFeature: Add ability to import external playlists as crates

fwcd:
This is a small extension to the existing import-as-playlist action found in `BaseExternalLibraryFeature` that enables users to import external playlists as crates too:

image

Crates have the unique property that their names must be unique, so the import can fail if the name already exists. Currently this is dealt with by presenting an error dialog, but we may consider prompting the user e.g. to specify a custom name instead or to merge the tracks into the existing crate. Ideas welcome.

Something else to consider is that the import is currently synchronous and can be quite slow for large external playlists (thousands of songs). This applies to the existing playlist import too though, therefore I would consider any optimizations to be out-of-scope for this PR.

### DIFF
--- a/src/library/baseexternallibraryfeature.cpp
+++ b/src/library/baseexternallibraryfeature.cpp
@@ -6,6 +6,7 @@
 #include "library/library.h"
 #include "library/trackcollection.h"
 #include "library/trackcollectionmanager.h"
+#include "library/trackset/crate/crate.h"
 #include "moc_baseexternallibraryfeature.cpp"
 #include "util/logger.h"
 #include "widget/wlibrarysidebar.h"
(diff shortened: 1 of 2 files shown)


src/library/baseexternallibraryfeature.cpp
```diff
@@ -125,6 +133,30 @@ void BaseExternalLibraryFeature::slotImportAsMixxxPlaylist() {
...
+    crate.setName(playlist);
+
+    CrateId crateId;
+
+    if (m_pTrackCollection->insertCrate(crate, &crateId)) {
+        m_pTrackCollection->addCrateTracks(crateId, trackIds);
+    } else {
+        QMessageBox::warning(nullptr,
+                tr("Crate Creation Failed"),
+                tr("Could not create crate, it most likely already exists: ") + playlist);
```
Swiftb0y, line 156: would be nice to have a better error message here but I guess thats not currently possible with the `insertCrate` API, right?

Swiftb0y (commented):
LGTM otherwise tbh. but I'd prefer if somebody with library experience (@ronso0 maybe) could express their opinion as well.

Swiftb0y (approved):
welp, LGTM. lets merge.

PR #11853, opened 2023-08-21
Pref Interface: avoid unneeded skin reload, clean up

ronso0:
Previously, on Ok/Apply the skin would be reloaded even though the skin or color scheme configuration didn't actually change, for example if users just checked out the previews of other skins or schemes.

plus a separator in the locale list (below 'System'), improved locale logging and some clean up


Swiftb0y (approved):
LGTM. thank you

PR #11859, opened 2023-08-23
Pref Sound: auto select free device channels

### LINKED ISSUE #10163: Sound Hardware preferences should auto-select next free channel
Reported by: **[Holzhaus](https://github.com/Holzhaus)**
Date: 2020-10-13T22:39:15Z
Status: Confirmed
Importance: Low
Launchpad Issue: [lp1899703](https://bugs.launchpad.net/bugs/1899703)
Tags: easy, preferences, usability

---

When configuring sound hardware in the preferences, Mixxx currently always assigns the first channel of a soundcard to every input or output instead of using the next free available channel.

For example, if you want to configure the main and headphones output, the workflow is the following

1. Select soundcard A in the main output's soundcard combobox. The channel combobox automatically picks channels 1-2.
2. Select soundcard A in the headphone output's soundcard combobox. The channel combobox automatically picks channels 1-2. (the configuration is now invalid because both outputs use the same channels)
3. Select channels 3-4 in the headphone output's channel combobox.

Mixxx should automatically pick channels 3-4 in step 2, because channels 1-2 are already in use. That would make step 3 unnecessary.

ronso0:
I started another coffee break project to add little helper:
when you need to configure multiple sound outputs using _one_ device, like channels 1-2 for Main & 3-4 for headphones as with many controller soundcards, and you configured the first output it is very pleasant if the next free channel pair is selected automatically for the second output.
1. Main: select device XY -> channels 1-2 selected (default from populating the channels box)
2. Headphones: select device XY -> channels 3-4 are selected automatically

That saves one click (yeaiiy!!), but more importantly: it avoids mis(re-)configuration in the Heat of the Night:tm: (assign same channels twice, Apply, error, meeeh, select correct channels, Apply).

**Note:** this does just combobox index matching, i.e. does not check channel overlap like
Mic1: channel 1, Aux1: channels 1-2.

Closes #10163 

This is still WIP but it's working nicely already.
I'll do more testing and polish it.
- [x] remove DEBUG commit

### DIFF
--- a/src/preferences/dialog/dlgprefsound.cpp
+++ b/src/preferences/dialog/dlgprefsound.cpp
@@ -378,16 +378,24 @@ void DlgPrefSound::addPath(const AudioInput& input) {
             &DlgPrefSound::refreshInputDevices,
             pSoundItem,
             &DlgPrefSoundItem::refreshDevices);
-    insertItem(pSoundItem, inputVLayout);
     connectSoundItem(pSoundItem);
+    insertItem(pSoundItem, inputVLayout);
 
     setScrollSafeGuardForAllInputWidgets(pSoundItem);
 }
 
 void DlgPrefSound::connectSoundItem(DlgPrefSoundItem* pItem) {
-    connect(pItem, &DlgPrefSoundItem::settingChanged, this, &DlgPrefSound::deviceSettingChanged);
+    connect(pItem,
+            &DlgPrefSoundItem::selectedDeviceChanged,
+            this,
+            &DlgPrefSound::deviceChanged);
+    connect(pItem,
+            &DlgPrefSoundItem::selectedChannelsChanged,
+            this,
+            &DlgPrefSound::deviceChannelsChanged);
     connect(this, &DlgPrefSound::loadPaths, pItem, &DlgPrefSoundItem::loadPath);
     connect(this, &DlgPrefSound::writePaths, pItem, &DlgPrefSoundItem::writePath);
+    connect(this, &DlgPrefSound::refreshInputDevices, pItem, &DlgPrefSoundItem::refreshDevices);
     connect(this, &DlgPrefSound::updatingAPI, pItem, &DlgPrefSoundItem::save);
     connect(this, &DlgPrefSound::updatedAPI, pItem, &DlgPrefSoundItem::reload);
 }
@@ -397,7 +405,7 @@ void DlgPrefSound::insertItem(DlgPrefSoundItem *pItem, QVBoxLayout *pLayout) {
     for (pos = 0; pos < pLayout->count() - 1; ++pos) {
         DlgPrefSoundItem *pOther(qobject_cast<DlgPrefSoundItem*>(
             pLayout->itemAt(pos)->widget()));
-        if (!pOther) {
+        if (!pOther) { // not a sound item, skip
             continue;
         }
         if (pItem->type() < pOther->type()) {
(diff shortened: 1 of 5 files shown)


fwcd:
I like this, it makes the common case very convenient, though I think it could also feel a bit "magical", especially if the user has a more esoteric device/channel configuration.

ronso0:
I rarely use more than one IO device so real-world examples that are can be improved are welcome!

Note that the occupied channel indices are stored per device so it won't suggest weird channels for formerly unconfigured devices.
If for example a user first selected channels 3/4 and selects the same device again it'll select 1/2 as before, like for all formerly unused devices. Can't think of an example case where the auto-select would be disturbing, or more inconvenient than the current behaviour.
For scenarios where muliple mono channels are configured for a device, auto-select pick 1/2 but this is also the current behaviour. Making auto-select pick another mono channel is doable but I'm not sure it's desireable in all cases.

ronso0:
It's working nicely, didn't encounter any hickups, incl. Query Devices.

Ready for review.

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -655,38 +666,96 @@ void DlgPrefSound::refreshDevices() {
...
+        return;
+    }
+    QHash<DlgPrefSoundItem*, QPair<SoundDeviceId, int>>* channels;
+    if (pItem->isInput()) {
+        channels = &m_selectedInputChannelIndices;
+    } else {
+        channels = &m_selectedOutputChannelIndices;
+    }
+    auto id = pItem->getDeviceId();
+    qWarning() << "     dlg: dev Changed" << id.name;
```
Holzhaus, line 695: Shouldn't this be `qDebug()` instead? Samoe goes for all the other log messages below.

ronso0: whoopsy, re-added the debug commit while merging 2.4 into the local branch.
I'll remove it again.

ronso0:
_bump_
Is there anything left to do?

ronso0:
_ba-bump_

JoergAtGithub:
Just tested this and it works very neat with my Traktor Z2 (2xOut Ch + 4x In Ch in unusual order) accross all the Windows Audio driver APIs.
Please fix the conflict.

src/coreservices.cpp
```diff
@@ -304,6 +304,7 @@ void CoreServices::initialize(QApplication* pApp) {
     }
 
     m_pPlayerManager->addConfiguredDecks();
+    // addConfiguredSamplers() ?
```
JoergAtGithub, line 307: This should be removed

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -405,7 +403,7 @@ void DlgPrefSound::insertItem(DlgPrefSoundItem *pItem, QVBoxLayout *pLayout) {
     for (pos = 0; pos < pLayout->count() - 1; ++pos) {
         DlgPrefSoundItem *pOther(qobject_cast<DlgPrefSoundItem*>(
             pLayout->itemAt(pos)->widget()));
-        if (!pOther) {
+        if (!pOther) { // don't replace existing. DEBUG_ASSERT ?
```
JoergAtGithub, line 406: Please decide, if this should be an assert. If not remove the comment.

src/coreservices.cpp
ronso0: :+1: slipped in, and has already been added

src/preferences/dialog/dlgprefsound.cpp
ronso0: :+1:

ronso0: didn't look closer earlier, this could theoratically happen if we added a spacer or separator, which is not the case currently

ronso0:
Thanks for your review @JoergAtGithub !
Ready for merge.

JoergAtGithub:
LGTM! Thank you!

PR #11862, opened 2023-08-24
DlgAbout: Add Qt version to the dialog

fwcd:
A small QoL improvement that adds the Qt version to the about dialog:

Screenshot 2023-08-25 at 00 38 17

### DIFF
--- a/src/dialog/dlgaboutdlg.ui
+++ b/src/dialog/dlgaboutdlg.ui
@@ -110,12 +110,12 @@
         <item row="1" column="0">
          <widget class="QLabel" name="label_2">
           <property name="text">
-           <string>Date:</string>
+           <string>Qt Version:</string>
           </property>
          </widget>
         </item>
         <item row="1" column="1">
-         <widget class="QLabel" name="date_label">
+         <widget class="QLabel" name="qt_version_label">
           <property name="text">
            <string>Unknown</string>
           </property>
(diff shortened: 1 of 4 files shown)


src/dialog/dlgaboutdlg.ui
```diff
@@ -78,8 +78,27 @@
       </item>
       <item>
        <layout class="QGridLayout" name="gridLayout">
-        <item row="2" column="1">
-         <widget class="QLabel" name="platform_label">
+        <item row="0" column="0">
```
fwcd, line 81: I apologize for the slightly noisy diff: The markup was quite hard to read since rows/columns were in no recognizable order, therefore I've sorted them first. Reviewing commit-by-commit might be the easiest here.

ronso0: 👍 
good ol' QtCreator used to randomize item order when items were added/moved in the GUI editor and diffs were unreviewable. Though, out if laziness I tried it lately and this seems to be fixed.

src/util/versionstore.cpp
```diff
@@ -1,5 +1,6 @@
 #include "util/versionstore.h"
 
+#include <qglobal.h>
```
Swiftb0y, line 3: The Qt docs refer to this as header by `<QtGlobal>` which already included below. https://doc.qt.io/qt-5/qtglobal.html

```suggestion
```

Swiftb0y (commented):
LGTM Otherwise.

src/util/versionstore.cpp
fwcd: Whoops, clangd is doing funny stuff again, thanks for catching this

Swiftb0y (approved):
LGTM, waiting for CI

Holzhaus (approved):
LGTM, thank you!

PR #11863, opened 2023-08-24
Qt6 prepare

daschuer:
It turns out that this is required to build main with the new Qt6 vcpkg environment. 
I like to merge this first to verify the Qt5 compatibility which I like to maintain during 2.5 beta to be able to compare certain behaviors.

* Sleef is a FFT library used on macOS from Rubberband 3.1
* The logic around WIN32/WIN64 definitions was inverted which proves that it was wrong. The WIN32 define is correct even on a 64 bit system, because it has still the so called WIN32 API.

### DIFF
--- a/cmake/modules/Findrubberband.cmake
+++ b/cmake/modules/Findrubberband.cmake
@@ -85,14 +85,32 @@ if(rubberband_FOUND)
         INTERFACE_COMPILE_OPTIONS "${PC_rubberband_CFLAGS_OTHER}"
         INTERFACE_INCLUDE_DIRECTORIES "${rubberband_INCLUDE_DIR}"
     )
-    is_static_library(rubberband_IS_STATIC Chromaprint::Chromaprint)
+    is_static_library(rubberband_IS_STATIC rubberband::rubberband)
     if(rubberband_IS_STATIC)
-      find_package(FFTW REQUIRED)
       find_library(SAMPLERATE_LIBRARY samplerate REQUIRED)
       set_property(TARGET rubberband::rubberband APPEND PROPERTY INTERFACE_LINK_LIBRARIES
-        FFTW::FFTW
         ${SAMPLERATE_LIBRARY}
       )
+      find_package(FFTW)
+      if (FFTW_FOUND)
+        set_property(TARGET rubberband::rubberband APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+          FFTW::FFTW
+        )
+      endif()
+      find_package(FFTW)
+      if (FFTW_FOUND)
+        set_property(TARGET rubberband::rubberband APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+          FFTW::FFTW
+        )
+      endif()
+      find_package(Sleef)
+      if (Sleef_FOUND)
+        find_library(sleefdft_path sleefdft REQUIRED)
+        set_property(TARGET rubberband::rubberband APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+          Sleef::sleef
+          ${sleefdft_path}
+        )
+      endif()
     endif()
   endif()
 endif()
--- a/cmake/modules/Findrubberband.cmake
+++ b/cmake/modules/Findrubberband.cmake
@@ -97,12 +97,6 @@ if(rubberband_FOUND)
           FFTW::FFTW
         )
       endif()
-      find_package(FFTW)
-      if (FFTW_FOUND)
-        set_property(TARGET rubberband::rubberband APPEND PROPERTY INTERFACE_LINK_LIBRARIES
-          FFTW::FFTW
-        )
-      endif()
       find_package(Sleef)
       if (Sleef_FOUND)
         find_library(sleefdft_path sleefdft REQUIRED)
--- a/src/engine/sidechain/shoutconnection.cpp
+++ b/src/engine/sidechain/shoutconnection.cpp
@@ -6,14 +6,7 @@
 #include <unistd.h>
 #endif
 
-// shout.h checks for WIN32 to see if we are on Windows.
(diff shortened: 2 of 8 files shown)


cmake/modules/Findrubberband.cmake
```diff
@@ -85,14 +85,32 @@ if(rubberband_FOUND)
...
+        set_property(TARGET rubberband::rubberband APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+          FFTW::FFTW
+        )
+      endif()
+      find_package(FFTW)
+      if (FFTW_FOUND)
+        set_property(TARGET rubberband::rubberband APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+          FFTW::FFTW
+        )
+      endif()
```
Swiftb0y, line 105: copy-paste mistake?
```suggestion
```

cmake/modules/Findrubberband.cmake
```diff
@@ -85,14 +85,32 @@ if(rubberband_FOUND)
...
+      find_package(Sleef)
+      if (Sleef_FOUND)
+        find_library(sleefdft_path sleefdft REQUIRED)
+        set_property(TARGET rubberband::rubberband APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+          Sleef::sleef
+          ${sleefdft_path}
+        )
+      endif()
```
Swiftb0y, line 113: whats the error like if neither sleef nor fftw is found?

src/library/dao/trackdao.cpp
```diff
@@ -1381,7 +1381,7 @@ TrackPointer TrackDAO::getTrackById(TrackId trackId) const {
         QString columnsStr;
         int columnsSize = 0;
         for (int i = 0; i < columnsCount; ++i) {
-            columnsSize += qstrlen(columns[i].name) + 1;
+            columnsSize += static_cast<int>(qstrlen(columns[i].name)) + 1;
         }
         columnsStr.reserve(columnsSize);
```
Swiftb0y, line 1386: I would've preferred changing `columnsSize` to `std::size_t`/`qsizetype` instead.

src/util/versionstore.cpp
```diff
@@ -8,22 +8,12 @@
...
-#ifdef WIN64
-#undef WIN32
-#endif
-
 #include <FLAC/format.h>
 #include <chromaprint.h>
 #include <lame/lame.h>
 #include <portaudio.h>
 #include <rubberband/RubberBandStretcher.h>
+#include <shoutidjc/shout.h>
```
Swiftb0y, line 16: what happened to the `#ifdef __BROADCAST__`?

src/engine/sidechain/shoutconnection.cpp
```diff
@@ -611,12 +604,6 @@ bool ShoutConnection::processConnect() {
             kLogger.warning()
                     << "processConnect() socket error."
                     << "Is socket already in use?";
-        } else if (timeout >= kConnectRetries) {
-            // Not translated, because shout_get_error() returns also English only
-            m_lastErrorStr = QStringLiteral("Connection establishment time-out");
-            kLogger.warning()
-                    << "processConnect() error:"
-                    << m_iShoutStatus << m_lastErrorStr;
```
Swiftb0y, line 619: why the removal?

src/preferences/dialog/dlgprefbroadcast.cpp
```diff
@@ -158,8 +151,6 @@ DlgPrefBroadcast::DlgPrefBroadcast(QWidget *parent,
              &QCheckBox::stateChanged,
              this,
              &DlgPrefBroadcast::enableCustomMetadataChanged);
-
-     setScrollSafeGuardForAllInputWidgets(this);
```
Swiftb0y, line 162: why the removal?

Swiftb0y (changes requested):

src/engine/sidechain/shoutconnection.cpp
daschuer: Oh that was a faulty rebase. Forced pushed the last commit.

src/library/dao/trackdao.cpp
daschuer: Yes that world be nice. However QT provides no common ground with qsizetype for QString.reserve(). So using `int` is the compatible solution.

cmake/modules/Findrubberband.cmake
daschuer: Rubberband is linked one of it, which is the choice of the package provider. If the library is not found build will fail with a significant linker error. 
I have no idea how to investigate which library is used to fail early. 
We need to wait for the Rubberband CONFIG mode.

src/library/dao/trackdao.cpp
Swiftb0y: What do you mean, "no common ground"? `QString.reserve()` takes a `qsizetype` on Qt6...

cmake/modules/Findrubberband.cmake
Swiftb0y: Couldn't we just detect if `!fftw_FOUND && !Sleef_FOUND` and error out early then? Sorry for the dumb questions, I'm not familiar enough with cmake. I'm just trying to wrap my head around it.

src/library/dao/trackdao.cpp
JoergAtGithub: Shouldn't `auto` use `int` on Qt5 and `size_t` on Qt6?

daschuer: We need a cast anyway, because Qt6 uses std::ptrdifsize for reserve()

Swiftb0y (approved):
LGTM, lets keep the ball rolling

PR #11880, opened 2023-08-28
CMakeLists: Fix `QT_TRANSLATION_FILE` path for Qt 6

fwcd:
This fixes the (currently broken) Qt 6 configure for me. It is effectively an adaptation of 06aec1dfe482c073f759e43cf47f7e17c2600487 from the [`qt6_switch`](https://github.com/daschuer/mixxx/tree/qt6_switch) branch, which, however, still uses the `Qt5_DIR` variable instead of `Qt6_DIR` (not sure why that worked in the first place).

cc @daschuer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2646,7 +2646,7 @@ if(APPLE OR WIN32)
   # qt_de.qm is just one arbitrary file in the directory that needs to be located;
   # there is no particular reason to look for this file versus any other one in the directory.
   if(QT6)
-    find_file(QT_TRANSLATION_FILE qt_de.qm PATHS "${Qt6_DIR}/../../../translations" "${Qt6_DIR}/../../qt5/translations" REQUIRED NO_DEFAULT_PATH)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -2646,7 +2646,7 @@ if(APPLE OR WIN32)
   # qt_de.qm is just one arbitrary file in the directory that needs to be located;
   # there is no particular reason to look for this file versus any other one in the directory.
   if(QT6)
-    find_file(QT_TRANSLATION_FILE qt_de.qm PATHS "${Qt6_DIR}/../../../translations" "${Qt6_DIR}/../../qt5/translations" REQUIRED NO_DEFAULT_PATH)
```
fwcd, line 2649: Side note: `QT_DIR` seems to work too, as well as simply passing `/translations/Qt6` (which seems to be implicity resolved against `CMAKE_PREFIX_PATH`, i.e. the vcpkg install root, even if `NO_DEFAULT_PATH` is passed?!).

For consistency I went with `Qt6_DIR` though.

daschuer (approved):
LGTM, thank you

PR #11891, opened 2023-08-29
Update buildenvs for macOS and Windows

Holzhaus:
See the commit messages.

I had to use the `2.5-rel` instead of `2.5` buildenv for arm64-osx, because the `2.5` buildenv is out-of-date. I also had to switch the macOS builds to Qt6 because these do not contain Qt5 anymore. Since 2.5 should be Qt6 anyway, this may be not that much of an issue because switching sooner means we'll have more time to find issues.

 At least the CI shouild be fixed for now (hopefully).


daschuer:
I have a similar PR in progress, unfortunately the windows test is failing. 
It is interesting that the test is here not failing. 

I have just created a PR https://github.com/mixxxdj/mixxx/pull/11892 
let's see how it works.

PR #11901, opened 2023-08-30
(fix) Library track menu: show Hide action also in Playlist & Crates

ronso0:
Adds 'Hide from library' to Playlists (except History) and Crates features.

Test:
`Del` hotkey should be shown in 'Remove ..' actions and should trigger only these actions.
If there's no 'Remove ..' actions `Del`is used for 'Hide from library'

Closes #9049

### DIFF
--- a/src/library/playlisttablemodel.cpp
+++ b/src/library/playlisttablemodel.cpp
@@ -368,6 +368,7 @@ TrackModel::Capabilities PlaylistTableModel::getCapabilities() const {
             Capability::LoadToPreviewDeck |
             Capability::ResetPlayed |
             Capability::RemoveFromDisk |
+            Capability::Hide |
             Capability::Analyze;
 
     if (m_iPlaylistId !=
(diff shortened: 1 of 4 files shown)


ronso0:
Meeh, back to draft:
the crates & playlist views are updated correctly, but in History views the action does nothing.

ronso0:
> Meeh, back to draft: the crates & playlist views are updated correctly, but in History views the action does nothing.

Okay, that's because History shows **all** played tracks, incl. now hidden tracks, hence the view won't change after hiding tracks.
We may show a hint to avoid confusion:
> History shows all tracks that have been played, including tracks that are marked hidden.

This should go to the History root page. Showing a note after clicking Hide is cumbersome and requires setting another 'showHidefromHistoryHint' flag.

ronso0:
Not sure how to proceed:
1. show Hide also in History (current implementation, no hint)
2. show Hide also in History, show hint, set dontShowAgain flag
3. don't show /disable Hide in History

I prefer 1 but that may cause some confusion.
> This should go to the History root page.

History doens't have a root view anymore :facepalm: 

2 and 3. require more changes, e.g. subclass `PlaylistTableModel` -> `HistoryPlaylistTableModel`, or alter `BaseSqlTableModel::hideTracks`, to show the hint only for the setlog feature.

daschuer:
This one has a conflict now. Sorry.

ronso0:
Conflicts resolved.

daschuer (approved):
LGTM, Thank you.

PR #11921, opened 2023-09-04
Preferences > Sound: ensure sound item tab order matches visual order

ronso0:
Another UX helper like #11859:
Currently the sound items¹ are inserted into the I/O layouts in random order (though being sorted by type and index on insertion) which results in a random Tab order.

This PR sorts them the same way, just _before_ insertion.
Does apply only to the audio paths added during initialization, hence sound items added later on may theoretically be inserted in between existing items.  (in which situation? I didn't investigate)

¹Audio path label ('Main' etc.) + device combobox + channel combobox

### DIFF
--- a/src/preferences/dialog/dlgprefsound.cpp
+++ b/src/preferences/dialog/dlgprefsound.cpp
@@ -312,12 +312,51 @@ QUrl DlgPrefSound::helpUrl() const {
  * if necessary.
  */
 void DlgPrefSound::initializePaths() {
-    foreach (AudioOutput out, m_pSoundManager->registeredOutputs()) {
-        if (!out.isHidden()) {
-            addPath(out);
+    // Pre-sort paths so they're added in the order they'll appear later on
+    // so Tab key order matches order in layout:
+    // * by AudioPathType
+    // * identical types by index
+    QList<AudioOutput> outputsSorted;
+    for (const auto& outNew : m_pSoundManager->registeredOutputs()) {
+        if (outNew.isHidden()) { // Don't add Record/Broadcast and possibly others
+            continue;
         }
+        int pos;
+        for (pos = 0; pos < outputsSorted.count(); ++pos) {
+            const auto& outOld = outputsSorted.at(pos);
+            if (outNew.getType() < outOld.getType()) {
+                break;
+            } else if (outOld.getType() == outNew.getType() &&
+                    outOld.isIndexed(outOld.getType()) &&
+                    outNew.isIndexed(outNew.getType()) &&
+                    outNew.getIndex() < outOld.getIndex()) {
+                break;
+            }
+        }
+        outputsSorted.insert(pos, outNew);
     }
-    foreach (AudioInput in, m_pSoundManager->registeredInputs()) {
+    for (const auto& out : qAsConst(outputsSorted)) {
+        addPath(out);
+    }
+
+    // Inputs
+    QList<AudioInput> inputsSorted;
+    for (const auto& inNew : m_pSoundManager->registeredInputs()) {
+        int pos;
+        for (pos = 0; pos < inputsSorted.count(); ++pos) {
+            const auto& inOld = inputsSorted.at(pos);
+            if (inNew.getType() < inOld.getType()) {
+                break;
+            } else if (inOld.getType() == inNew.getType() &&
+                    inOld.isIndexed(inOld.getType()) &&
+                    inNew.isIndexed(inNew.getType()) &&
+                    inNew.getIndex() < inOld.getIndex()) {
+                break;
+            }
+        }
+        inputsSorted.insert(pos, inNew);
+    }
+    for (const auto& in : qAsConst(inputsSorted)) {
         addPath(in);
     }
 }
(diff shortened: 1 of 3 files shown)


src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -312,12 +312,51 @@ QUrl DlgPrefSound::helpUrl() const {
...
+                break;
+            } else if (inOld.getType() == inNew.getType() &&
+                    inOld.isIndexed(inOld.getType()) &&
+                    inNew.isIndexed(inNew.getType()) &&
+                    inNew.getIndex() < inOld.getIndex()) {
+                break;
+            }
         }
+        inputsSorted.insert(pos, inNew);
     }
```
Swiftb0y, line 358: Should do the same, but probably results in horrible compiler errors because I just typed this out freehand. If you don't want to dig through those I can work them out for you and send you a PR.

I highly prefer this solution because it's more concise and more importantly it's actually algorithmically faster because it's not using the $O(n^2)$ insertion sort.

We should also consider to shift the sorting logic to the path itself instead of implementing it here. and `addPath` could probably be sped-up with an rvalue-ref overload. 

```suggestion
    auto sortFilterAdd = [this]<template T>(QList<T> l) {
        std::ranges::sort(l, [](const T& lhs, const T& rhs){
            return (lhs.getType() < rhs.getType()) || (lhs.getType() == rhs.getType() && lhs.isIndexed(lhs.getType()) && rhs.isIndexed(rhs.getType()) && lhs.getIndex() < rhs.getIndex())
        });
        for (const T& path: l | std::views::filter([](const T& path){ return path.isHidden(); }))) {
            addPath(path);
        };
    };
    sortFilterAdd(m_pSoundManager->registeredOutputs());
    sortFilterAdd(m_pSoundManager->registeredInputs());
```

Swiftb0y (changes requested):
as I said, I'd be happy to implement the changes myself and just file you a PR.

src/preferences/dialog/dlgprefsound.cpp
ronso0: oh, nice I was looking into templates briefly but gave up, and didn't consider std::ranges, so...
> I can work them out for you and send you a PR

please do so!
Though, this pretty much reverts my changes so you might as well open a separate PR. I don't care..

ronso0: > We should also consider to shift the sorting logic to the path itself instead of implementing it here.

Are you talking about moving the sort logic to `SoundManager::registerOutput` so it returns a sorted list in the first place?

Swiftb0y: > Are you talking about moving the sort logic to SoundManager::registerOutput so it returns a sorted list in the first place?

No sorry, that wasn't expressed properly. I meant to add the sorting criteria to the type itself (implementing `operator<=>` on AudioPath) since there are probably other parts of the code that could also benefit from being able to compare these paths.

ronso0: Okay, looks interesting. As you know I'm not into c++ basics so go ahead and I'm happy to help with testing and other low-level stuff like the GUI implementation.

> since there are probably other parts of the code that could also benefit from being able to compare these paths.

Currently it's only the sound preferences that requests the registered outputs and connects to the `outputRegistered` signal.

FYI I adopted the sorting logic from [`DlgPrefSound::insertItem`](https://github.com/mixxxdj/mixxx/blob/4c82ffa01a49b6c6e617e96cb6dcc01028c72fe9/src/preferences/dialog/dlgprefsound.cpp#L401-L418)

daschuer: Can we make this a normal member function? This will require less c++ knowledge to read. 

```
        if (outNew.isHidden()) { // Don't add Record/Broadcast and possibly others
            continue;
        }
```
Is already quite good readable. Can we stick with it? 

sortFilterAdd implicit creates a deep copy of QList first. It would be better to use an empty container and reserve() the required memory and than add the elements in a sorted way. 
Maybe we can avoid the temporary container at all by adding the sorting logic to addPath()?

Swiftb0y: > Can we make this a normal member function? This will require less c++ knowledge to read.

What exactly should be a normal member function?

> Is already quite good readable. Can we stick with it?

If it makes sense yes, though in the meantime I arrived with a version that's does not suite this behavior.

> sortFilterAdd implicit creates a deep copy of QList first. It would be better to use an empty container and reserve() the required memory and than add the elements in a sorted way.
Maybe we can avoid the temporary container at all by adding the sorting logic to addPath()?

> reserve() the required memory and than add the elements in a sorted way.

Thats not trivial without verbose code and without being $O(n^2)$. Better option would be to have a temporary `vector<reference_wrapper<AudioPath>>` that is being used for sorting. 

>  Maybe we can avoid the temporary container at all by adding the sorting logic to addPath()?

That would result in the same inefficient insertionsort behavior as `DlgPrefSound::insertItem`



I'm currently still working on this but I'm stuck on some compile-errors related to `operator<=>` which will take some time to dig through.

ronso0:
I'm closing this is favour of #11926.
Thanks @Swiftb0y for taking over!

PR #11925, opened 2023-09-05
Small fixes for common-hid-packet-parser.js

JoergAtGithub:
-TypedArray is no public JS keyword, we need to list them all, not only for createDataView, but also for it's callers
-Last 3 arguments of addControl are optional (but were not defined this way in JSDoc)
-length argument of parsePacket is only needed for a warning message -> make it optional

The other changes are automatic reformatting by pre-commit(including the removal of default values for optional arguments)


daschuer (approved):
LGTM, Thank you

PR #11927, opened 2023-09-06
chore(ci): bump runners to macos-12 to benefit from newer compiler

Swiftb0y:
this should not change our minimum supported MacOS version, right?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,8 +58,8 @@ jobs:
             artifacts_path: build/*.deb
             artifacts_slug: ubuntu-jammy
             qt_qpa_platform: offscreen
-          - name: macOS 11 x64
-            os: macos-11
+          - name: macOS 12 x64
+            os: macos-12
             cmake_args: >-
               -DBULK=ON
               -DCOREAUDIO=ON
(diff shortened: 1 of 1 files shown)


daschuer:
What is the complier feature you try to benefit from? 
Normally we use the oldest CI runner to make sure everything still works with that.

Holzhaus:
Yeah, both approaches are have their advantages. A newer compiler may have support for better optimization/vectorization/etc. On the other hand it makes sense to ensure that old compilers are still working unless we have a reason to drop support.

Maybe some macOS users can test this branch. If there are noticeable performance improvements, I'd be in favor of merging this. But I doubt that the compiler makes that much of a difference to be honest.

Swiftb0y:
> What is the complier feature you try to benefit from?

C++20 library features. macos-11 only supports xcode 13 but a couple quite useful library features (`std::bit_cast`, constexpr `std::vector` and `std::string`, c++ ranges, see https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_library_features) are only available in Xcode 14.

daschuer:
These features are also not supported on Ubuntu Focal 20.4 gcc 9 (Mixxx 2.4) and Ubuntu Jammy 22.4 gcc 11 (Mixxx 2.5) 

We currently use Xcode 13.2.1, the maximum we can use on MacOS 11. 
Once we use Xcode 14 only features we also deprecate MacOS 11 as build environment for our contributors and as testing environment on our CI.

Swiftb0y:
Okay, In reality I really just want to not have to deal with the MacOS CI failure on #11926. Not sure whats going on there, my guess is that apple is shipping a weird version of libc++ as `operator<=>` has only been implemented on tuples [somewhat recently](https://reviews.llvm.org/D108250). I can work around this but encountering these sorts of bogus issues is straining.

PR #11933, opened 2023-09-07
CMakeLists: Match arbitrary `arm64-osx` triplets

fwcd:
Compiling natively on an arm64 macOS machine will, by default, use `arm64-osx` as a vcpkg host triplet, which currently isn't recognized by the CMake configuration. Since macOS 11 is the first version that supports arm64, we'd want to use a deployment target of 11.0 on _all_ Apple Silicon machines instead of defaulting to 10.15. This branch fixes this by matching all triplets beginning with `arm64-osx`.


JoergAtGithub:
LGTM! Thank you!

PR #11892, opened 2023-08-30
Qt6 switch

daschuer:
This switches the main branch to Qt6.

PR #11934, opened 2023-09-07
CMakeLists: Default `QT6` to `ON`

fwcd:
Given that

- #11892 

was recently merged and that the 2.5 vcpkg environments now ship Qt 6, we should set `-DQT6=ON` by default, both for consistency and to eliminate another configuration gotcha. This should also remove the need to set this explicitly in CI (though I didn't remove those here to keep the PR focused and since the workflows might be intentionally explicit).

Of course, developers can still set `-DQT6=OFF` to build with Qt 5.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,7 @@ if(NOT CMAKE_CONFIGURATION_TYPES)
   endif()
 endif()
 
-option(QT6 "Build with Qt6" OFF)
+option(QT6 "Build with Qt6" ON)
 option(QML "Build with QML" OFF)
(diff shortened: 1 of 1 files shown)


Swiftb0y (approved):
LGTM. we should do that anyways otherwise people on macos and windows will run into issues because the environment doesn't ship Qt5 anymore. Objections?

Holzhaus:
> LGTM. we should do that anyways otherwise people on macos and windows will run into issues because the environment doesn't ship Qt5 anymore. Objections?

No.

Holzhaus (approved):

PR #11936, opened 2023-09-07
File Access fix

daschuer:
Due to refactoring the thread safe access of the Tracks File access object is no longer thread safe. 
https://github.com/mixxxdj/mixxx/blob/2.4/src/track/track.h#L82

I have removed this function and pass the thread safe TrackPointer instead. 
This has the big advantage, that we can use it directly with the sound source for extracting the cover and don't need to use the location of the track outside the track locking it up again from the Track cache later. 

In https://github.com/mixxxdj/mixxx/issues/11131 we have seen that the independent track location was empty in the concurrent case. 

I don't think that getFileAccess() was the cause or even a real Problem, because it is almost all the time read only, but hopefully the additional ASSERTS and the streamlined control flow helps to find the real root cause.

(I am still not able to produce the issue) 

@JoergAtGithub can you check this out with debug asserts fatal?

### DIFF
--- a/src/library/coverart.cpp
+++ b/src/library/coverart.cpp
@@ -106,8 +106,7 @@ QDebug operator<<(QDebug dbg, const CoverInfoRelative& info) {
             << '}';
 }
 
-CoverInfo::LoadedImage CoverInfo::loadImage(
-        const SecurityTokenPointer& pTrackLocationToken) const {
+CoverInfo::LoadedImage CoverInfo::loadImage(TrackPointer pTrack) const {
     LoadedImage loadedImage(LoadedImage::Result::ErrorUnknown);
     if (type == CoverInfo::METADATA) {
         VERIFY_OR_DEBUG_ASSERT(!trackLocation.isEmpty()) {
@@ -115,10 +114,12 @@ CoverInfo::LoadedImage CoverInfo::loadImage(
             return loadedImage;
         }
         loadedImage.location = trackLocation;
-        loadedImage.image = CoverArtUtils::extractEmbeddedCover(
-                mixxx::FileAccess(
-                        mixxx::FileInfo(trackLocation),
-                        pTrackLocationToken));
+        if (pTrack) {
+            loadedImage.image = CoverArtUtils::extractEmbeddedCover(pTrack);
+        } else {
+            loadedImage.image = CoverArtUtils::extractEmbeddedCover(
+                    mixxx::FileAccess(mixxx::FileInfo(trackLocation)));
+        }
         if (loadedImage.image.isNull()) {
             // TODO: extractEmbeddedCover() should indicate if no image
             // is available or if loading the embedded image failed.
@@ -171,8 +172,7 @@ CoverInfo::LoadedImage CoverInfo::loadImage(
 }
 
 bool CoverInfo::refreshImageDigest(
-        const QImage& loadedImage,
-        const SecurityTokenPointer& pTrackLocationToken) {
+        const QImage& loadedImage) {
     if (!imageDigest().isEmpty()) {
         // Assume that a non-empty digest has been calculated from
         // the corresponding image. Otherwise we would refresh all
@@ -183,7 +183,7 @@ bool CoverInfo::refreshImageDigest(
     }
     QImage image = loadedImage;
     if (image.isNull()) {
-        image = loadImage(pTrackLocationToken).image;
+        image = loadImage(TrackPointer()).image;
     }
     if (image.isNull() && type != CoverInfo::NONE) {
         kLogger.warning()
--- a/src/library/coverart.cpp
+++ b/src/library/coverart.cpp
@@ -115,6 +115,7 @@ CoverInfo::LoadedImage CoverInfo::loadImage(TrackPointer pTrack) const {
         }
         loadedImage.location = trackLocation;
         if (pTrack) {
+            DEBUG_ASSERT(trackLocation == pTrack->getLocation());
             loadedImage.image = CoverArtUtils::extractEmbeddedCover(pTrack);
         } else {
             loadedImage.image = CoverArtUtils::extractEmbeddedCover(
--- a/src/library/coverartutils.cpp
+++ b/src/library/coverartutils.cpp
@@ -209,17 +209,17 @@ CoverInfoRelative CoverInfoGuesser::guessCoverInfo(
 
 CoverInfoRelative CoverInfoGuesser::guessCoverInfoForTrack(
         TrackPointer pTrack) {
-    const auto fileAccess = pTrack->getFileAccess();
+    const auto fileInfo = pTrack->getFileInfo();
     if (kLogger.debugEnabled()) {
         kLogger.debug()
                 << "Guessing cover art for track"
-                << fileAccess.info();
+                << fileInfo;
     }
 
     QImage embeddedCover = CoverArtUtils::extractEmbeddedCover(pTrack);
 
     return guessCoverInfo(
-            fileAccess.info(),
+            fileInfo,
             pTrack->getAlbum(),
             embeddedCover);
 }
--- a/src/library/coverartutils.cpp
+++ b/src/library/coverartutils.cpp
@@ -56,6 +56,18 @@ QImage CoverArtUtils::extractEmbeddedCover(
     return image;
 }
 
+QImage CoverArtUtils::extractEmbeddedCover(
+        TrackPointer pTrack) {
+    QImage image;
+    // Both resetMissingTagMetadata = false/true have the same effect
+    constexpr auto resetMissingTagMetadata = false;
+    SoundSourceProxy(pTrack).importTrackMetadataAndCoverImage(
+            nullptr,
+            &image,
+            resetMissingTagMetadata);
+    return image;
+}
+
 //static
 QList<QFileInfo> CoverArtUtils::findPossibleCoversInFolder(const QString& folder) {
     // Search for image files in the track directory.
@@ -204,7 +216,7 @@ CoverInfoRelative CoverInfoGuesser::guessCoverInfoForTrack(
                 << fileAccess.info();
     }
 
-    QImage embeddedCover = CoverArtUtils::extractEmbeddedCover(fileAccess);
+    QImage embeddedCover = CoverArtUtils::extractEmbeddedCover(pTrack);
 
     return guessCoverInfo(
             fileAccess.info(),
--- a/src/library/coverartutils.cpp
+++ b/src/library/coverartutils.cpp
@@ -46,8 +46,9 @@ QString CoverArtUtils::supportedCoverArtExtensionsRegex() {
 QImage CoverArtUtils::extractEmbeddedCover(
         mixxx::FileAccess trackFileAccess) {
     QImage image;
-    // Both resetMissingTagMetadata = false/true have the same effect
-    constexpr auto resetMissingTagMetadata = false;
+    // Since pTrackMetadataBoth is null, resetMissingTagMetadata
+    // has no effect
+    const auto resetMissingTagMetadata = false;
     SoundSourceProxy::importTrackMetadataAndCoverImageFromFile(
             std::move(trackFileAccess),
             nullptr,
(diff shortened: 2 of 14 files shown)


JoergAtGithub:
I did a short test. It worked without assert. I noticed, that the spinny coverart is now loaded only very seldom (once in 20 track loads), otherwise everything feels as before.

daschuer:
That is good news. 
We know now that the issue is fixed with a lock and got worse without any lock. 

I have already tried to make a unit-test, but it was always good. Maybe I can make it fail now. 

I suggest to merge this PR, because it makes the issue more likely, so it will be easier to verify a potential fix of the real root cause.

m0dB:
@daschuer , I am not familiar enough with this code to have a useful opinion about this PR, and it is not entirely clear to me what the objective is, but I will can try to run it on macOS with thread sanitizer to see if flags up some issue. Would that help?

daschuer:
I could drill that down into Taglib near https://github.com/taglib/taglib/blob/master/taglib/toolkit/tfilestream.cpp#L57. But I am not able to explain it. Everything looks good. 

The symptoms are that the FileName std::wstring which is copied around many times, is loosing its value becomming null. 
LastError is Null as well. Maybe an outer class is reinitialized or such. 

It happens whenever the same file is accessed at the same time from different threads.

The poor mans solution would be to guard that region with a mutex. 

There is on the other hand no reason to access the same file from different threads at the same time. So we should in any case think about improving the coverart cache to sort this out. I have something in progress. 

But I am really curious to find the root cause to not have another symptom popping up elsewhere later.

Debugging on Linux and MacOS does not make any sense, because that code is windows only.

daschuer:
A first step would be to merge this which makes the issue more likely. Maybe a windows contributor can than take-over. 
I will continue in the meantime to improve the cover-art cache.

m0dB:
Sounds good! If there is anything I can do on macOS, let me know. In the meantime, I did try to make a thread sanitizer enabled build: Independent of this bug, we definitely should run with thread sanitizer. Unfortunately, I immediately ran into false positives with QMutex. I will take the discussion about that to zulip.

src/track/globaltrackcache.cpp
```diff
@@ -341,15 +341,13 @@ void GlobalTrackCache::relocateTracks(
...
-            if (fileAccess.info() != relocatedFileAccess.info()) {
+            auto relocatedFileAccess = pRelocator->relocateCachedTrack(trackRef.getId());
+            if (relocatedFileAccess.info().hasLocation() &&
+                    fileAccess.info() != relocatedFileAccess.info()) {
                 plainPtr->relocate(relocatedFileAccess);
                 trackRef = TrackRef::fromFileInfo(
                         relocatedFileAccess.info(),
                         trackRef.getId());
-                fileAccess = std::move(relocatedFileAccess);
             }
```
Swiftb0y, line 349: why the removal? this line looks crucial to me.

src/library/coverartutils.cpp
```diff
@@ -196,46 +196,46 @@ CoverInfoRelative CoverInfoGuesser::guessCoverInfo(
 }
 
 CoverInfoRelative CoverInfoGuesser::guessCoverInfoForTrack(
-        const Track& track) {
-    const auto fileAccess = track.getFileAccess();
+        TrackPointer pTrack) {
+    const auto fileAccess = pTrack->getFileAccess();
```
Swiftb0y: how does this avoid a TrackCache roundtrip?

src/library/coverartutils.h
```diff
@@ -30,6 +30,8 @@ class CoverArtUtils {
     // Extracts the first cover art image embedded within the file.
     static QImage extractEmbeddedCover(
             mixxx::FileAccess trackFileAccess);
+    static QImage extractEmbeddedCover(
+            TrackPointer pTrack);
```
Swiftb0y, line 34: whats the advantage of this overload over the `FileAccess` version?

src/library/coverartutils.cpp
```diff
@@ -56,6 +56,18 @@ QImage CoverArtUtils::extractEmbeddedCover(
     return image;
 }
 
+QImage CoverArtUtils::extractEmbeddedCover(
+        TrackPointer pTrack) {
+    QImage image;
+    // Both resetMissingTagMetadata = false/true have the same effect
+    constexpr auto resetMissingTagMetadata = false;
```
Swiftb0y, line 63: that sounds like a bug?

src/track/track.h
```diff
@@ -78,14 +78,16 @@ class Track : public QObject {
...
         // i.e. no locking needed.
-        return m_fileAccess;
+        return m_fileAccess.token();
     }
+
     mixxx::FileInfo getFileInfo() const {
-        // Copying QFileInfo is thread-safe due to implicit sharing,
+        // Copying mixxx::FileInfo based on QFileInfo is thread-safe due to implicit sharing,
         // i.e. no locking needed.
+        static_assert(sizeof(QSharedDataPointer<int>) == sizeof(mixxx::FileInfo));
```
Swiftb0y, line 84: while I appreciate the usage of the `static_assert`, I'm not sure the way its implemented is a good idea. It's very leaky. 
I think it would be better if there was a `constexpr static` member on `mixxx::FileInfo` that indicated whether it only consisted of implicitly shared types. Ideally Qt had some machinery for this, but I don't think it does.

src/track/track.h
```diff
@@ -78,14 +78,16 @@ class Track : public QObject {
     Q_PROPERTY(QString titleInfo READ getTitleInfo STORED false NOTIFY infoChanged)
     Q_PROPERTY(QDateTime sourceSynchronizedAt READ getSourceSynchronizedAt STORED false)
 
-    mixxx::FileAccess getFileAccess() const {
-        // Copying QFileInfo is thread-safe due to implicit sharing,
+    SecurityTokenPointer getFileAccessToken() const {
+        // Copying a QSharedPointer is thread-safe
         // i.e. no locking needed.
```
Swiftb0y: why was this unsafe previously? `FileAccess` consists of a `FileInfo` and a `QSharedPointer`, both are thread-safe to copy.

src/library/coverart.cpp
```diff
@@ -106,19 +106,20 @@ QDebug operator<<(QDebug dbg, const CoverInfoRelative& info) {
...
     if (type == CoverInfo::METADATA) {
         VERIFY_OR_DEBUG_ASSERT(!trackLocation.isEmpty()) {
             loadedImage.result = LoadedImage::Result::ErrorMetadataWithEmptyTrackLocation;
             return loadedImage;
         }
         loadedImage.location = trackLocation;
-        loadedImage.image = CoverArtUtils::extractEmbeddedCover(
-                mixxx::FileAccess(
-                        mixxx::FileInfo(trackLocation),
-                        pTrackLocationToken));
```
Swiftb0y, line 121: won't this cause problems on environments with stricter security such as MacOS since we don't have a token anymore?

src/track/globaltrackcache.cpp
daschuer: `fileAccess` has been removed completely from this function.  The new file access is already owned by the track after  `plainPtr->relocate()`. So there is no use for this code line anymore.

src/track/track.h
daschuer: I have put it here because it proves the comment just in the next line.  If one will add members to FileInfo it will fail here where the mutex lock is required.

daschuer:
It turns out that the real bug is a weakness of the "CreateFileW()" the first thread is opening the file which seems to lock it. 
The second thread tries to access the same file R/W which fails than it tries again in RO mode which also fails and gives up finally.

src/track/globaltrackcache.cpp
Swiftb0y: Ahh sorry, I was only looking at this line in the context of the commit which still contained the fileAccess variable above.

src/track/track.h
Swiftb0y: Right, but it will not fail if the implementation changes but it happens to have the same size. That's why I proposed to have the evaluation of the property in the fileaccess class, but leave the static_assert here. It's just shifting code around a bit in order to not break encapsulation.

daschuer: Now we have a static isQFileInfo(). I was attempted to call it isThreadSafeCopyable() but that would be very confusing in a Info class describing file actions.

src/track/track.h
daschuer: Yes, but the context switch may happen in between so that the you have the old FileInfo with a new access token.

src/library/coverart.cpp
daschuer: We till have the token. In case we have a Track, the track owns the access token. 
In the other case a new access token is requested like before where pTrackLocationToken was null.

src/library/coverartutils.cpp
daschuer: I did not investigate that. The comment is a literal copy form the previous version.

src/library/coverartutils.h
daschuer: If we have a track, we can us it directly and bypass the track lookup logic.

src/util/fileinfo.h
```diff
@@ -255,6 +257,13 @@ class FileInfo final {
         return m_fileInfo.size();
     }
 
+    // This can be used to assert that the object is thread-safe copyable like QFileInfo
+    static constexpr bool isQFileInfo {
+        // Additional member variables will violate this assumption
+        return (sizeof(QFileInfo) == sizeof(mixxx::FileInfo) &&
+                std::is_same_v<decltype(m_fileInfo), QFileInfo>)
+    }
```
Swiftb0y:
```suggestion
    // This only works because Qt's implicit sharing allows copies to be threadsafe and
    // this class only consists of a single QFileInfo member.
    // Adding additional member variables will likely violate this assumption.
    static constexpr bool canBeCopiedThreadSafe = sizeof(QFileInfo) == sizeof(mixxx::FileInfo) && std::is_same_v<decltype(m_fileInfo), QFileInfo>
    }
```

src/track/track.h
```diff
@@ -81,7 +81,7 @@ class Track : public QObject {
     mixxx::FileInfo getFileInfo() const {
         // Copying mixxx::FileInfo based on QFileInfo is thread-safe due to implicit sharing,
         // i.e. no locking needed.
-        static_assert(sizeof(QSharedDataPointer<int>) == sizeof(mixxx::FileInfo));
+        static_assert(mixxx::FileInfo::isQFileInfo());
```
Swiftb0y, line 84:
```suggestion
        static_assert(mixxx::FileInfo::canBeCopiedThreadSafe);
```

Swiftb0y (changes requested):

src/track/track.h
Swiftb0y: >  I was attempted to call it isThreadSafeCopyable() but that would be very confusing in a Info class describing file actions.

Yeah, but its a naming issue. The entire point of the static is not to expose the internal QFileInfo imo. I think it sufficies to use a better name or doc comment.

src/library/coverartutils.h
Swiftb0y: What do we gain from that?

src/library/coverart.cpp
Swiftb0y: Okay, that second behavior was not obvious. Makes sense.

src/track/track.h
daschuer: As I have described above, I had this original but picked isQFileInfo because than it is obvious that it is not related to the file itself. It is intended to expose the QFileInfo nature here, because than the developer can read through the QT docs.
It is a difference to expose something to the developer by a name, compared to exposing implementation details to the compiler. The later is now fixed and the first desired.

src/library/coverartutils.cpp
uklotzde: If you don't read any other metadata by passing `nullptr` the option has no effect.

Swiftb0y (approved):
Since this fixes a bug, I won't hold this PR up over naming and codestyle discussions. LGTM.

daschuer:
Thank you.

PR #11937, opened 2023-09-07
refactor(util/timer): cleanup includes

Swiftb0y:
I noticed an AutoMoc warning during the build and got annoyed by it. Turns out there was an unnecessary `#include "moc_timer.cpp"`. While I was at it, I also cleaned up the rest of the headers.


JoergAtGithub:
Thank you! LGTM!

PR #11112, opened 2022-12-05
Cover fullsize fixes

ronso0:
These are some UX improvements for the full-size cover art windows:

* full-size cover in Track Properties dialog: changes are applied only when clicking Apply in Track Properties
* always update visible full-size covers when parent cover label changed
* in decks / library: keep size and position of full-size cover if a new track (with cover image) is loaded / selected

This is #11109 rebased onto 2.3 with 09cca5ed32a23c5627626693bdc36a2077a4eba3 and deb95b39f3218055312672bcf212da9f1f3a3403 from main.

I'll take care of merging to main, and I can also take care of rebasing #4851.

Commits are as small as possible, so review should be easy.

___
**Notes for testing:**

Full-size cover in decks (cover & spinny cover), library and Track Properties dialog:
* ~~image is updated in place (fit into the existing window) when a new track is loaded or selected via Next/Previous, or when a new image is assigned via the cover menu (from full-size, WCoverArt and WCoverArtLabel widgets or from any track menu)~~ post-ponde
* (new/fix) cover widget next to tracks table is updated if that track's cover is changed in Track Properties
* when full-size cover  is updated it is not 'activated', i.e. does not get keyboard focus, that is still on the previously focused widget
* full-size is closed if new track has no cover or if the cover is cleared via any related cover menu 

In Track Properties dialog:
* cleared/reset/new cover selected via cover menu (full-size or label in dialog) is applied only when pressing Apply
* full-size cover: if cover is cleared/reset/changed via the full-size menu, both that and the cover in Track Properties should be updated (track cover, e.g. in decks) should not change until Apply is clicked

In Tag Fetcher dialog:
* full-size is closed when a new MusicBrainz result is selected

**TODO**
- [ ] remove debug commit

Closes #10815
Closes #10816

PR #11938, opened 2023-09-08
Cover Art Fetcher: allow applying just the found cover

### LINKED ISSUE #11086: Tag Fetcher: allow applying tags and new cover art separately
### Feature Description

#4851 
With the current state of #4851 cover art is looked up based on each tag result, though best-matching tags may not coincide with best-matching/preferred cover art.

Let's split the dialog so that we have separate "Apply tgas" and "Apply cover art" controls.

ronso0:
... + some layouts tweaks

before:
image

this PR:
image

Closes #11086

### DIFF
--- a/src/library/dlgtagfetcher.cpp
+++ b/src/library/dlgtagfetcher.cpp
@@ -142,7 +142,8 @@ void DlgTagFetcher::init() {
         btnPrev->hide();
     }
 
-    connect(btnApply, &QPushButton::clicked, this, &DlgTagFetcher::apply);
+    connect(btnApply, &QPushButton::clicked, this, &DlgTagFetcher::applyTagsAndCover);
+    connect(btnApplyCover, &QPushButton::clicked, this, &DlgTagFetcher::applyCover);
     connect(btnQuit, &QPushButton::clicked, this, &DlgTagFetcher::quit);
     connect(btnRetry, &QPushButton::clicked, this, &DlgTagFetcher::retry);
     connect(tags, &QTreeWidget::currentItemChanged, this, &DlgTagFetcher::tagSelected);
@@ -234,6 +235,7 @@ void DlgTagFetcher::loadTrack(const TrackPointer& pTrack) {
 
     btnRetry->setDisabled(true);
     btnApply->setDisabled(true);
+    btnApplyCover->setDisabled(true);
     statusMessage->setVisible(false);
     loadingProgressBar->setVisible(true);
     loadingProgressBar->setValue(kMinimumValueOfQProgressBar);
@@ -261,7 +263,7 @@ void DlgTagFetcher::slotTrackChanged(TrackId trackId) {
     }
 }
 
-void DlgTagFetcher::apply() {
+void DlgTagFetcher::applyTagsAndCover() {
     int tagIndex = m_data.m_selectedTag;
     if (tagIndex < 0) {
         return;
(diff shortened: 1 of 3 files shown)


Swiftb0y:
The status message seems clunky, couldn't that be conveyed better by greying out the apply button when its not possible to apply and then updating the "current cover art" preview if it was successfully applied?

ronso0:
> greying out the apply button when its not possible to apply and then updating the "current cover art" preview if it was successfully applied?

Thank you, that slipped through. IIRC I already fixed that in https://github.com/mixxxdj/mixxx/pull/11112, I'll look into either backporting the fix or do it manually in DlgTagFetcher.

ronso0:
Actually, the 'Ccurrent Cover' should be updated after it was written to the track file succesfully (need to click Okay to overwrite the track file).
The status message is displayed already when the file overwrite dialog pops up.

ronso0:
Furthermore, there are some inconsistencies in the file update process: the tags are written directly whereas the cover art update is initiated but results (dialog/success/failure) are processed via signals/slots, so there's no guarantee the messages are displayed in the correct order.
* "Selected metadata applied" message may be displayed after the cover was written
* "Metadata & Cover Art applied" is displayed by `slotWorkerCoverArtUpdated` even though that slot is responsible only for the cover art (until now, it was assumed that both tags and cover are updated simultaneously).

I'll check whether that can be fixed while keeping the tr strings in place.

src/library/dlgtagfetcher.cpp
```diff
@@ -325,6 +328,19 @@ void DlgTagFetcher::apply() {
     }
 #endif // __EXTRA_METADATA__
 
+    applyCover();
+
+    statusMessage->setText(tr("Selected metadata applied"));
+
+    m_pTrack->replaceMetadataFromSource(
```
Swiftb0y, line 335: small nitpick, but shouldn't these be ordered the other way around?

src/library/dlgtagfetcher.cpp
```diff
@@ -582,13 +594,15 @@ void DlgTagFetcher::slotLoadBytesToLabel(const QByteArray& data) {
...
-    statusMessage->setText(tr("Cover art is ready to be applied"));
+    statusMessage->clear();
     statusMessage->setVisible(true);
 
     m_fetchedCoverArtByteArrays = data;
     m_pWFetchedCoverArtLabel->loadData(
             m_fetchedCoverArtByteArrays); // This data loaded because for full size.
     m_pWFetchedCoverArtLabel->setCoverArt(coverInfo, fetchedCoverArtPixmap);
+
+    btnApplyCover->setEnabled(!data.isNull());
```
Swiftb0y, line 605: nitpick, but easier to read.
```suggestion
    btnApplyCover->setDisabled(data.isNull());
```

src/library/dlgtagfetcher.cpp
```diff
@@ -371,19 +386,13 @@ void DlgTagFetcher::apply() {
...
-    statusMessage->setText(tr("Selected metadata applied"));
-
-    m_pTrack->replaceMetadataFromSource(
-            std::move(trackMetadata),
-            // Prevent re-import of outdated metadata from file tags
-            // by explicitly setting the synchronization time stamp
-            // to the current time.
-            QDateTime::currentDateTimeUtc());
+    statusMessage->setText(tr("Selected cover art applied"));
 }
```
Swiftb0y, line 391: IIUC, since the metadata is not replaced anymore, this will cause the metadata to not be applied if people press "Apply Cover" and then exit the dialog, right?

ronso0: Yup, correct.

Swiftb0y: Right, that seems unexpected to me. If I pressed the Apply button and exited the dialog, I'd expect the changes to be applied.

Swiftb0y: Ah, nevermind, I found the issue. I was misunderstood the code.

Swiftb0y (approved):
LGTM. Thank you.

ronso0:
Thanks for your review!

> Furthermore, there are some inconsistencies in the file update process: the tags are written directly whereas the cover art update is initiated but results (dialog/success/failure) are processed via signals/slots, so there's no guarantee the messages are displayed in the correct order.
> 
>     * "Selected metadata applied" message may be displayed after the cover was written
> 
>     * "Metadata & Cover Art applied" is displayed by `slotWorkerCoverArtUpdated` even though that slot is responsible only for the cover art (until now, it was assumed that both tags and cover are updated simultaneously).

Follow-up is #12041

PR #11941, opened 2023-09-08
Use `mixxx::audio::ChannelCount` type instead of `int`/`unsigned char`/etc.

Holzhaus:


Swiftb0y (approved):
LGTM.
I fear in the future `ChannelCount` won't be a powerful enough abstraction anymore, for stems we probably need a ChannelLayout that can manage an arbitrary amount of channels and give them meaning (surround sound, stems, etc?).

Holzhaus:
> I fear in the future `ChannelCount` won't be a powerful enough abstraction anymore, for stems we probably need a ChannelLayout that can manage an arbitrary amount of channels and give them meaning (surround sound, stems, etc?).

We already have https://github.com/mixxxdj/mixxx/blob/1fce230e3ddb5818271d33cc9aa8d25b8095ee3e/src/audio/types.h#L26-L31. We could start replacing `ChannelCount` instances with it (where applicable).

Swiftb0y:
Ah didn't know about that. Yeah makes sense.

Issue #11931, opened 2023-09-07
Removal of offensive terms from the codebase


Holzhaus:
In June 2020, [we committed to remove the term *master* from our codebase for the 2.4 release](https://mixxx.org/news/2020-06-29-black-lives-matter/). Unfortunately, we did not follow through. Although we did remove some occurrences (e.g., the leader sync rename), we did not really fulfill that promise.

We need to discuss how we want to proceed now.
One one hand, our announcement might be seen as performative if we don't actually remove these terms for 2.4.
On the other hand I think it's a bit late now to refactor so much code before 2.4 beta, so we might consider postponing it to 2.5.

For the actual renaming, we need to discuss what we want as a replacement. Here's my initial proposal:

<details><summary>Click here to show proposal. <i>(Content Warning: Contains offensive terms associated with slavery)</i></summary>

| Term used | Course of Action
| ---------- | -------------
| `[ChannelN],master` CO | Add alias `[ChannelN],enabled`
| `[Master]` CO group | Add group alias `[Main]`
| `EngineMaster` object | Rename to `EngineMixer`
| `MicMonitorMode::MASTER` enum member | Rename to `MicMonitorMode::Main`
| `MicMonitorMode::MASTER_AND_BOOTH` enum member | Rename to `MicMonitorMode::MainAndBooth`

</details>

Swiftb0y:
Proposal LGTM, only issue I see is that creating a group alias might be difficult.

Also, monitoring the DJ community as a whole, it doesn't seem like the term "Master" fell out of fashion (which is admittedly sad). I still think we should pull through with the renaming, but I'm concerned of how much we loose by deciding against using the established domain language.

daschuer:
Right, .. another 2.4 task. The advantage to do it in 2.4 is that we wont have merge conflicts if we merge 2.4 to 2.5 regularly.  
But we will have a lot initially due to the "EngineMaster" -> "EngineMixer" This applies to the change at any time. 

The other issue is that we have an annoying long list of 2.4 regressions: 
https://github.com/mixxxdj/mixxx/issues?q=is%3Aopen+is%3Aissue+label%3Aregression+milestone%3A2.4.0
And we are past due by one month. 

We need to decide at one point to release anyway.

ronso0:
> but I'm concerned of how much we loose by deciding against using the established domain language.

What do you mean exactly?

Holzhaus:
[...]

PR #11942, opened 2023-09-08
Remove offensive language (Pt. 1)

Holzhaus:
This is a partial fix for #11931. What is missing is the CO group rename of `[Master]` and the `EngineMaster` rename, which would affect way more files.


src/engine/enginemaster.cpp
```diff
@@ -67,11 +67,11 @@ EngineMaster::EngineMaster(
     m_pWorkerScheduler->start(QThread::HighPriority);
 
     // Master sample rate
```
Swiftb0y, line 69:
```suggestion
```

src/engine/enginemaster.cpp
```diff
@@ -93,20 +93,20 @@ EngineMaster::EngineMaster(
     m_pBalance = new ControlPotmeter(ConfigKey(group, "balance"), -1., 1.);
 
     // Master gain
```
Swiftb0y, line 95:
```suggestion
```

Swiftb0y (changes requested):
`AudioPath::AudioPathType` enum members (in `soundmanagerutil.h`) is not yet converted. Didn't do a thorough exhaustiveness check apart from that.

Holzhaus:
> `AudioPath::AudioPathType` enum members (in `soundmanagerutil.h`) is not yet converted. Didn't do a thorough exhaustiveness check apart from that.

I added this in https://github.com/mixxxdj/mixxx/pull/11942/commits/43535a7c41c32363919b2ba41a8cb7a35930cd12. I think the PR is big enum now, let's do further changes in subsequent PRs to keep stuff reviewable.

Holzhaus:
Ah, and please don't merge this before 2.4 has been merged into main (if necessary). This will make it easier to resolve conflicts, because I guess there will be a bunch due to the channelcount/samplerate refactoring.

Swiftb0y:
> please merge this 2.4 into main before merging this PR (if necessary)

I assume this is not necessary anymore since you just did that an hour ago, right?

src/soundio/soundmanagerutil.h
```diff
@@ -56,46 +56,46 @@ class AudioPath {
...
-    /// Returns the minimum number of channels needed on a sound device for an
-    /// AudioPathType.
-    static unsigned char minChannelsForType(AudioPathType type);
-
-    // Returns the maximum number of channels needed on a sound device for an
-    // AudioPathType.
-    static unsigned char maxChannelsForType(AudioPathType type);
-
-    uint hashValue() const {
+  enum class AudioPathType : int {
```
Swiftb0y, line 59: any particular reason for specifying the underlying type explicitly?

Swiftb0y (approved):
LGTM. any objections?

src/soundio/soundmanagerutil.h
Holzhaus: > any particular reason for specifying the underlying type explicitly?

There is code that uses the enum's `int` value and converts between the types. Although the underlying type of a scoped `enum class` always `int` if not otherwise specified, I though that explicit is better than implicit and shows that we actually want to use `int` here.

Holzhaus:
I have a follow-up PR in progress where we rename to `EngineMixer` and a few other variable names and comments are fixed.

src/soundio/soundmanagerutil.h
Swiftb0y: I think its only important to specify the underlying type when doing a bit_cast/reinterpret_cast on a struct that contains this. Otherwise the type is only specified to be as many bits wide as required to hold all enum values... 

Just because we static_cast from and to `int` for this enum doesn't really mean that it needs to be represented in memory as an `int`... Anyways, I don't really see the usage warranted here, but idc. Just my two cents.

Swiftb0y:
Welp, here goes nothing. Lets see how many conflicts this creates.

PR #11945, opened 2023-09-08
Use maintained ilammy/msvc-dev-cmd@v1

daschuer:
This way we get rid of the ::set-config warning in our workflows.


Swiftb0y (commented):
still builds, so LGTM. :shrug:

Swiftb0y (approved):

PR #11948, opened 2023-09-08
iTunes: Re-enable test and add `composer`, `playCount`, `lastPlayedAt` and `dateAdded` to model

fwcd:
Now that #11666 has been merged, this re-enables the iTunes test. Additionally it adds a few new fields to `ITunesTrack` (along with the corresponding XML and macOS import implementations) that map directly to fields that Mixxx uses for its library tracks:

- `QString composer`
- `int playCount`
- `QDateTime lastPlayedAt`
- `QDateTime dateAdded`

The test is also updated to expect these new fields.

Note, however, that this is only the first piece of the puzzle, **this PR won't make the fields user-visible yet**. For that, the database schema has to be updated (albeit in a backwards-compatible-way). Since that's a more invasive change, my plan was to do that in a separate PR.

I'm targeting 2.4 since these changes should be pretty safe to merge and only affect testing. Should the follow-up PR (changing the schema to add the respective new columns to the db) also target 2.4 or is that too large of a change? Feel free to leave some notes.


fwcd:
Interesting, GCC seems to be rather strict with omitted fields in C++20's designated initializers, which I thought were explicitly intended to allow default initialization of their omitted members...

Swiftb0y (approved):
LGTM. any1 else?

Swiftb0y:
Thank you

PR #11951, opened 2023-09-09
DlgPrefSound: Fix fetching of sample rate

### LINKED ISSUE #11949: Regression: Audio buffer size UI shows `inf ms` on macOS
### Bug Description

Mixxx reports bogus audio latencies with the Core Audio backend (though it still seems to work fine):

image

According to my regression finder, this was introduced between 8d0ded74be (good) and e4f2e050db (bad), see [this diff](https://github.com/mixxxdj/mixxx/compare/8d0ded74be240f1249500b8b9484c00031e22f3e...e4f2e050dbf793fb195902328413ced50d1a46aa). Looking at the commits it seems as if it was introduced by
- #11904

cc @Holzhaus 

### Version

2.5-alpha-98-gf4692aa5ea (Qt 6.5.2)

### OS

macOS 13.5

fwcd:
Fixes #11949, i.e. causes audio buffer sizes to be displayed correctly again:

image

The issue was that the sample rate combo box was still queried for a double, while underlying data already used the `SampleRate` type, causing it to presumably default to 0 and thus result in divisions by zero. This was likely an oversight in #11904.

### DIFF
--- a/src/preferences/dialog/dlgprefsound.cpp
+++ b/src/preferences/dialog/dlgprefsound.cpp
@@ -608,7 +608,11 @@ void DlgPrefSound::updateAudioBufferSizes(int sampleRateIndex) {
                 static_cast<unsigned int>(SoundManagerConfig::
                                 JackAudioBufferSizeIndex::Size4096fpp));
     } else {
-        double sampleRate = sampleRateComboBox->itemData(sampleRateIndex).toDouble();
+        DEBUG_ASSERT(sampleRateComboBox->itemData(sampleRateIndex)
+                             .canConvert<mixxx::audio::SampleRate>());
+        double sampleRate = sampleRateComboBox->itemData(sampleRateIndex)
+                                    .value<mixxx::audio::SampleRate>()
+                                    .toDouble();
         unsigned int framesPerBuffer = 1; // start this at 0 and inf loop happens
         // we don't want to display any sub-1ms buffer sizes (well maybe we do but I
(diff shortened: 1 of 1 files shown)


src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -608,7 +608,8 @@ void DlgPrefSound::updateAudioBufferSizes(int sampleRateIndex) {
                 static_cast<unsigned int>(SoundManagerConfig::
                                 JackAudioBufferSizeIndex::Size4096fpp));
     } else {
-        double sampleRate = sampleRateComboBox->itemData(sampleRateIndex).toDouble();
+        double sampleRate = sampleRateComboBox->itemData(sampleRateIndex)
+                                    .value<mixxx::audio::SampleRate>();
```
Swiftb0y, line 612: lets add a debug assert to avoid the issue with the default construction in the future.
```suggestion
        DEBUG_ASSERT(sampleRateComboBox->itemData(sampleRateIndex).canConvert<mixxx::audio::SampleRate>());
        double sampleRate = sampleRateComboBox->itemData(sampleRateIndex)
                                    .value<mixxx::audio::SampleRate>().toDouble();
```

Swiftb0y (changes requested):

Swiftb0y:
thank you, please run pre-commit and squash.

Holzhaus (approved):
LGTM, thank you!

Swiftb0y (approved):
LGTM thank you

PR #11871, opened 2023-08-26
CMakeLists: Add `QT_RELEASE` flag to force linking a release build of Qt

fwcd:
Qt 6 debug builds can be quite slow, especially during startup (~20 seconds for debug builds, ~3 seconds for release builds of Qt). To get around this issue during development, this introduces a new flag, `QT_RELEASE` that lets the user build Mixxx in debug mode while still linking release binaries of Qt.

The approach is based on

- https://stackoverflow.com/questions/58402517/how-to-link-to-release-build-of-qt-in-cmake-with-debug-build
- https://stackoverflow.com/questions/57223210/cmake-link-to-only-release-config-of-target-in-both-debug-and-release/57223840#57223840

> Side note: While mixing release and debug binaries seems to be discouraged on Windows, it seems to be fine and even encouraged on Linux/macOS. See [this thread](https://stackoverflow.com/questions/11658915/mixing-debug-and-release-library-binary-bad-practice) for details. In any case, the flag is opt-in, so it wouldn't break anything about the current build.

PR #11952, opened 2023-09-09
CMakeLists: Always default `CMAKE_BUILD_TYPE` to `RelWithDebInfo`

fwcd:
As discussed in https://github.com/mixxxdj/mixxx/pull/11871#issuecomment-1694760590, we should default to release builds on all platforms since debug builds of Qt (especially Qt 6) are too slow even for development. Currently this is only done on Windows.

cc @daschuer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,11 +110,6 @@ set(X_VCPKG_APPLOCAL_DEPS_INSTALL ON CACHE BOOL "Automatically copy dependencies
 # Set a default build type if none was specified
 # See https://blog.kitware.com/cmake-and-the-default-build-type/ for details.
 set(default_build_type "RelWithDebInfo")
-if(EXISTS "${CMAKE_SOURCE_DIR}/.git" AND NOT WIN32)
-  # On Windows, Debug builds are linked to unoptimized libs
-  # generating unusable slow Mixxx builds.
-  set(default_build_type "Debug")
-endif()
 
 if(NOT CMAKE_CONFIGURATION_TYPES)
   if(NOT CMAKE_BUILD_TYPE)


Swiftb0y:
From what I can tell, this only sets build mode of mixxx to release, the build mode of Qt is decided by the packager (which is us in the case of vcpkg, but not on linux)...

fwcd:
True, though at least for the platforms where vcpkg is used, it would be convenient to have a reasonable default. IMO #11871 would be nice-to-have too, but that might be a rather specialized use-case (mixing debug Mixxx with release Qt)

Swiftb0y:
I'm not an expert, but IMO using a release version of a library (especially if its big and slow in debug as Qt) with a debug version of the application is not that uncommon...

fwcd:
Yeah, that's why I advocated for #11871 (and still do). This PR was mostly due to @daschuer's suggestion that we should default to release builds for everything. `RelWithDebInfo` should cover the debug symbols too.

daschuer (changes requested):
What is the goal for this change? On Linux it works good because the system libraries are optimized and Mixxx is optimized in Debug as well. This means that contributors have debug assertions and so one enabled by default. 

Is there a risk that package maintainer has a .git folder?  

We may adjust our concept a bit, but we need a clear picture what we want to achieve. 

What is puzzling me current is also build time. 
https://github.com/mixxxdj/mixxx/issues/12081

With the default configuration I am now suffering almost full build time after switching branches. I have a workaround, but casual contributors wont even know that there is an option to reduce build time after branch switching by a magnitude.

fwcd:
I will close this for now. Building Mixxx itself in debug mode is not really the issue, using debug builds of Qt is. Since we plan on shipping release-only build environments (see #12370) that would make it easy to build debug Mixxx with release Qt, there should not really be a need for building everything in release mode.

Issue #11932, opened 2023-09-07
macOS ARM: application crash after warning "passed a string to `engine.beginTimer`, please pass a function instead!"


todx0:
### Bug Description

This happens only on ARM version
Works fine on intel version that I installed from homebrew (2.3.6)

When I press next/previous track button on a controller DJ Control MP3 application crashes.

I checked with --developer and it seems that it crashes after lots of warnings:

 controller.herculesdjcontrolmp3wt10:warning [Controller] "passed a string to `engine.beginTimer`, please pass a function instead!"

### Version

2.4.0-beta

### OS

macOS Ventura 13.5.1

Swiftb0y:
Thank you for reporting your issue. Please understand that in order for us to be able to properly debug this, we need a backtrace. Have a look at the wiki for finding out how to create one: https://github.com/mixxxdj/mixxx/wiki/Creating%20Backtraces

todx0:
@Swiftb0y 
[backtrace.log](https://github.com/mixxxdj/mixxx/files/12550034/backtrace.log)

todx0:
Oops sorry for closing.

@Swiftb0y 
[backtrace.log](https://github.com/mixxxdj/mixxx/files/12550047/backtrace.log)

Swiftb0y:
Thank you, seems like there is a recursive call to `engine.beginTimer` in that mapping, causing a stack overflow... Which mapping are you using exactly?

Swiftb0y:
I assumed you were using the `Hercules-DJ-Control-MP3-scripts.js` mapping. Can you please try this updated version and upload another log and backtrace if it crashes? https://github.com/Swiftb0y/mixxx/blob/debug/gh11932/res/controllers/Hercules-DJ-Control-MP3-scripts.js

todx0:
No I'm using this one:

<img width="350" alt="image" src="https://github.com/mixxxdj/mixxx/assets/47313333/5d608c98-7576-4e13-8dba-a0374a36ab43">


/Applications/mixxx.app/Contents/Resources/controllers/Hercules-DJ-Control-MP3-hid-scripts.js

The file you provided didn't work but I seem to figure it out.



Uncomment string in line 341 in **_Hercules-DJ-Control-MP3-hid-scripts.js_** 

```js
HerculesMP3Hid.scroll_timer = engine.beginTimer(150, 'HerculesMP3Hid.scroll_tracks("[Playlist]","' + e + '",' + v + ')');
```
to
[...]

PR #11953, opened 2023-09-09
fix(controllers): `engine.beginTimer(ms, string)` would evaluate the code immediately instead of after `ms`

Swiftb0y:
From what I can tell, this must've been broken for years in 2.4 (075664d792 to be exact)...
The resulting behavior until now was that the string was evaluated immediately instead of after the timeout and when the timer was recurring, it was never evaluated after (since the stored result value was likely not callable).

The first commit fixes the mapping, the second commit fixes the API call, the third commit is some refactoring required for the fourth commit which will cause a `logOrThrowError` when the callback was not callable.

### DIFF
--- a/res/controllers/Hercules-DJ-Control-MP3-hid-scripts.js
+++ b/res/controllers/Hercules-DJ-Control-MP3-hid-scripts.js
@@ -142,7 +142,7 @@ HerculesMP3Hid.init = function() {
 
             if (v < 30) v = 30;
             if (HerculesMP3Hid.track_timer) engine.stopTimer(HerculesMP3Hid.track_timer);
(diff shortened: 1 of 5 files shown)


Swiftb0y:
fyi the pre-commit issue is from eslint because I touched a legacy mapping. Lets just ignore the eslint failure.

daschuer:
This aims to fix https://github.com/mixxxdj/mixxx/issues/11932 Let's wait for @todx0 for confirmation.

Swiftb0y:
they already did for the actual fix (second commit). the fourth is just to spot issues like this sooner in the future.

daschuer (approved):
OK, Than it should be safe to merge. 
Thank you for the fix.

Swiftb0y:
Thank you.

PR #11955, opened 2023-09-10
SearchQuery: Add ability to filter tracks in Mixxx library by iTunes playlist

fwcd:
This adds a new filter node to the search query machinery, namely `ExternalPlaylistFilterNode` that matches the name of an external playlist, along with a 'reference implementation' `ITunesFilterNode` that uses the iTunes tables (though the mechanism is explicitly intended to be generic, so support for Serato, Traktor etc. could easily be added in the future).

Thus, if the user has an iTunes playlist named "House", queries like `itunes:house` or `itunes:house daft punk` will now explicitly search for tracks in that playlist in a similar way to the `crate` operator.

External tracks are matched with library tracks via their location (since external libraries generally use a different track indexing scheme). Consequentially, this only affects the case where the external tracks are in the user's Mixxx library too (not necessarily the case, but probably common enough to make this a worthwhile use case).

### To do

- [x] Perform fuzzy matching on the playlist name (instead of requiring the exact name)
- [x] Investigate whether this is performant enough to warrant the inclusion of the untagged `ITunesFilterNode` (if not, we might want to consider gating this behind the explicit `itunes` operator)
  - We'll require the `itunes:` operator for now to avoid any potential performance regressions in regular searches, but leave this open for revisiting it in the future (See 1a2fca195df0d7b085bcfdec293ea7c4b06a4196)
- [x] Add unit test

### Future Directions

- Add Serato, Traktor etc. support as mentioned above (likely pretty easy, would only require subclassing `ExternalPlaylistFilterNode` and adding the corresponding operator to `SearchQueryParser`)
- Filter by non-uniquified iTunes playlist name (see `ITunesDAO::uniquifyPlaylistName`, we might want to store the non-unique playlist name in a separate column, that would require a schema change though and also require us to abstract over the playlist name column in `ExternalPlaylistFilterNode` since other external playlist integrations may not necessarily have this machinery).

### DIFF
--- a/src/library/itunes/itunesdao.cpp
+++ b/src/library/itunes/itunesdao.cpp
@@ -64,12 +64,12 @@ void ITunesDAO::initialize(const QSqlDatabase& database) {
     m_insertPlaylistTrackQuery.prepare(QStringLiteral(
             "INSERT INTO %1 (playlist_id, track_id, "
             "position) VALUES (:playlist_id, :track_id, :position)")
-                                               .arg(ITUNES_PLAYLIST_TRACKS_TABLE));
+                    .arg(ITUNES_PLAYLIST_TRACKS_TABLE));
 
     m_applyPathMappingQuery.prepare(QStringLiteral(
             "UPDATE %1 SET location = replace( location, "
             ":itunes_path, :mixxx_path )")
-                                            .arg(ITUNES_LIBRARY_TABLE));
+                    .arg(ITUNES_LIBRARY_TABLE));
 
     m_isDatabaseInitialized = true;
 }
--- a/src/library/itunes/itunesdao.cpp
+++ b/src/library/itunes/itunesdao.cpp
@@ -2,10 +2,12 @@
 
 #include <QObject>
 #include <QSqlQuery>
+#include <QStringLiteral>
 #include <gsl/pointers>
 
 #include "library/itunes/ituneslocalhosttoken.h"
 #include "library/itunes/itunespathmapping.h"
+#include "library/itunes/itunesschema.h"
 #include "library/queryutil.h"
 #include "library/treeitem.h"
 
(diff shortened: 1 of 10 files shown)


src/library/itunes/itunesschema.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <QString>
+
+#define ITUNES_LIBRARY_TABLE "itunes_library"
+#define ITUNES_PLAYLISTS_TABLE "itunes_playlists"
+#define ITUNES_PLAYLIST_TRACKS_TABLE "itunes_playlist_tracks"
```
fwcd, line 7: Implementation note: I've also extracted the names of the iTunes tables to avoid hardcoding them in different places (though admittedly not the column names since using those as constants would get excessively verbose e.g. in `ITunesDAO` and drastically reduce readability of the admittedly already quite long queries).

src/library/itunes/itunesdao.cpp
```diff
@@ -46,22 +48,27 @@ void ITunesDAO::initialize(const QSqlDatabase& database) {
...
-    m_applyPathMappingQuery.prepare(
-            "UPDATE itunes_library SET location = replace( location, "
-            ":itunes_path, :mixxx_path )");
+            QStringLiteral("INSERT INTO %1 (id, artist, title, album, "
+                           "album_artist, genre, grouping, year, duration, "
+                           "location, rating, comment, tracknumber, bpm, bitrate) "
+                           "VALUES (:id, :artist, :title, :album, :album_artist, "
+                           ":genre, :grouping, :year, :duration, :location, "
+                           ":rating, :comment, :tracknumber, :bpm, :bitrate)")
+                    .arg(ITUNES_LIBRARY_TABLE));
```
Swiftb0y, line 57: Even though this is not user input, please lets try to avoid string manipulations on queries whenever possible regardless. 

```suggestion
            QStringLiteral("INSERT INTO :table (id, artist, title, album, "
                           "album_artist, genre, grouping, year, duration, "
                           "location, rating, comment, tracknumber, bpm, bitrate) "
                           "VALUES (:id, :artist, :title, :album, :album_artist, "
                           ":genre, :grouping, :year, :duration, :location, "
                           ":rating, :comment, :tracknumber, :bpm, :bitrate)"));
               m_insertTrackQuery.bindValue("table", ITUNES_LIBRARY_TABLE);
```

src/library/itunes/itunesschema.h
Swiftb0y: can you make them `constexpr static`? or QStrings? Or is there a good reason for choosing a preprocessor define here?

Swiftb0y (commented):
small comments, full review to come some other time

src/library/itunes/itunesschema.h
fwcd: No particular reason aside from consistency with the other tables (e.g. as listed in `crateschema.h`)

src/library/itunes/itunesdao.cpp
fwcd: Looks like that's not really possible with Qt: https://stackoverflow.com/questions/15902859/qsqlquery-with-prepare-and-bindvalue-for-column-name-sqlite

Swiftb0y: Thank you. I didn't know that. Then disregard my suggestion.

Swiftb0y (commented):
this review was pending from ages ago but I forgot to finish it. Better comment late and unfinished than never.

src/library/itunes/itunesdao.cpp
```diff
@@ -46,22 +48,27 @@ void ITunesDAO::initialize(const QSqlDatabase& database) {
...
-
-    m_insertPlaylistTrackQuery.prepare(
-            "INSERT INTO itunes_playlist_tracks (playlist_id, track_id, "
-            "position) VALUES (:playlist_id, :track_id, :position)");
-
-    m_applyPathMappingQuery.prepare(
-            "UPDATE itunes_library SET location = replace( location, "
-            ":itunes_path, :mixxx_path )");
+            QStringLiteral("INSERT INTO %1 (id, artist, title, album, "
+                           "album_artist, genre, grouping, year, duration, "
```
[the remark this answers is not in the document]
Swiftb0y: Right, the comment was from before we agreed to change it. I mean, if we didn't use QString it would've worked but we don't so yeah...

acolombier:
Do you have interest to complete this @fwcd ? I'll mark as draft for now, but feel free to revive it once you've completed your todo list and want another review!

fwcd:
Yeah, I'm still personally interested in this feature and will take a look at getting this PR up to speed once I find some time

PR #11962, opened 2023-09-11
Use tags instead of branches for GitHub actions

Holzhaus:
At least pin a tag. We may consider switching to full commit SHA for better security in the future.


Swiftb0y (approved):
LGTM. thank you. Waiting for CI, just to be sure.

PR #11963, opened 2023-09-11
Fix crash when no GL context is available

### LINKED ISSUE #11929: Mixxx 2.4-beta crash on Windows 10 with No GL
### Bug Description

Tested on Virtual Box 7. 

EXCEPTION_ACCESS_VIOATION 

At the beginning ther is a popup about missing OpenGL 

### Version

2.4-beta

### OS

Windows 10

daschuer:
This fixes #11929

A next iteration would be to get rid of the initial GL widget. Maybe we can already use the context without the widget. 
I am just afraid that this requires a lot of testing. Nothing suitable just before a release.

### DIFF
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -1,22 +1,25 @@
 #include "mixxxmainwindow.h"
 
+#include <QDebug>
 #include <QDesktopServices>
 #include <QFileDialog>
+#include <QOpenGLContext>
+#include <QUrl>
+
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
+#include <QGLFormat>
+#endif
+
+#ifdef __LINUX__
+#include <QtDBus>
+#endif
 
 #include "widget/wglwidget.h"
 
 #ifdef MIXXX_USE_QOPENGL
 #include "widget/tooltipqopengl.h"
 #include "widget/winitialglwidget.h"
-#else
-#include <QGLFormat>
 #endif
-
-#include <QUrl>
-#ifdef __LINUX__
-#include <QtDBus>
-#endif
-#include <QtDebug>
 
 #include "dialog/dlgabout.h"
 #include "dialog/dlgdevelopertools.h"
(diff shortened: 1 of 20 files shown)


JoergAtGithub:
I guess this should be targeted to 2.4?

src/mixxxmainwindow.cpp
```diff
@@ -149,22 +152,36 @@ MixxxMainWindow::MixxxMainWindow(std::shared_ptr<mixxx::CoreServices> pCoreServi
...
+            // but allows us to take advantage of the existing classes.
+            WInitialGLWidget* widget = new WInitialGLWidget(this);
+            widget->setGeometry(QRect(0, 0, 3, 3));
+            SharedGLContext::setWidget(widget);
+            // When the widget's QOpenGLWindow has been initialized, we continue
+            // with the actual initialization
+            connect(widget, &WInitialGLWidget::onInitialized, this, &MixxxMainWindow::initialize);
+            widget->show();
+            // note: the format is set in the WGLWidget's OpenGLWindow constructor
+            return;
```
Swiftb0y, line 180: this comment no longer applies since the context has been created earlier in the function

src/mixxxmainwindow.cpp
```diff
@@ -149,22 +152,36 @@ MixxxMainWindow::MixxxMainWindow(std::shared_ptr<mixxx::CoreServices> pCoreServi
...
-        // When the widget's QOpenGLWindow has been initialized, we continue
-        // with the actual initialization
-        connect(widget, &WInitialGLWidget::onInitialized, this, &MixxxMainWindow::initialize);
-        widget->show();
-        // note: the format is set in the WGLWidget's OpenGLWindow constructor
-    } else {
-        initialize();
+#endif
+        QSurfaceFormat format;
+        format.setVersion(2, 1);
```
m0dB, line 163: Without this, we had a single place where QSurfaceFormat was set (in OpenGLWindow constructor). Also, I liked going through an actual QOpenGLWindow creation to do the check. But I am okay with this change, especially because I agree with the eventual removal of SharedGLContext.

daschuer: We may consider to set the default format in a single place instead. But I was not sure about the implications.

Swiftb0y: should be easy enough to solve by just putting the QSurfaceFormat creation into a function that is called from both places, right?

Swiftb0y (commented):
LGTM. please fix pre-commit before merge

daschuer:
Done.

Swiftb0y (approved):

Issue #11964, opened 2023-09-11
No Filename in Taglib Error Message 


daschuer:
### Bug Description

During debugging of https://github.com/mixxxdj/mixxx/issues/11131 this error hides the real issue of the Windows API I guess. 

### Version

2.4-beta

### OS

Windows 10 (Vbox)

Swiftb0y:
I don't understand the bug here. What taglib error message?

daschuer:
This is the message: 
`warning [Thread (pooled)] TagLib - Cannot read audio properties from inaccessible/unreadable/invalid file:`
The bug is that there is nothing after the colon.

Swiftb0y:
Thank you. I was missing a reference to the line in the code.

Swiftb0y:
fixed by #11965

PR #11965, opened 2023-09-11
Don't use a refrence into an rvalue object

daschuer:
This should fix the empty filename in the taglib error message
https://github.com/mixxxdj/mixxx/issues/11964

Maybe one can share some C++ knowledge, was 
```
const std::wstring& filename = file.name().wstr();
```
not allowed or is it a compiler bug? 
 
with `const std::wstring &wstr() const;`
and file name() returns by value.

### DIFF
--- a/src/track/taglib/trackmetadata_file.cpp
+++ b/src/track/taglib/trackmetadata_file.cpp
@@ -120,7 +120,8 @@ bool readAudioPropertiesFromFile(
     // we must pick one explicit,
     // to prevent "use of overloaded operator '<<' is ambiguous" error
     // on clang-cl builds.
-    const std::wstring& filename = file.name().wstr();
+    TagLib::FileName filename_owning = file.name();
+    const std::wstring& filename = filename_owning.wstr();
 #else
     const char* filename = file.name();
 #endif
--- a/src/track/taglib/trackmetadata_file.cpp
+++ b/src/track/taglib/trackmetadata_file.cpp
@@ -120,8 +120,8 @@ bool readAudioPropertiesFromFile(
     // we must pick one explicit,
     // to prevent "use of overloaded operator '<<' is ambiguous" error
     // on clang-cl builds.
-    // We need to save the filename here because if it was chained 
-    // (`file.name().wstr()`) the wstr() result would be dangling. 
+    // We need to save the filename here because if it was chained
+    // (`file.name().wstr()`) the wstr() result would be dangling.
     TagLib::FileName filename_owning = file.name();
     const std::wstring& filename = filename_owning.wstr();
 #else
--- a/src/track/taglib/trackmetadata_file.cpp
+++ b/src/track/taglib/trackmetadata_file.cpp
@@ -120,6 +120,8 @@ bool readAudioPropertiesFromFile(
     // we must pick one explicit,
     // to prevent "use of overloaded operator '<<' is ambiguous" error
     // on clang-cl builds.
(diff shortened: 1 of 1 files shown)


src/track/taglib/trackmetadata_file.cpp
```diff
@@ -114,20 +114,24 @@ bool hasMP4Tag(TagLib::MP4::File& file) {
...
-    // For _WIN32 there are two types std::string and std::wstring
-    // we must pick one explicit,
-    // to prevent "use of overloaded operator '<<' is ambiguous" error
-    // on clang-cl builds.
-    const std::wstring& filename = file.name().wstr();
+        // For _WIN32 there are two types std::string and std::wstring
+        // we must pick one explicit,
+        // to prevent "use of overloaded operator '<<' is ambiguous" error
+        // on clang-cl builds.
+        TagLib::FileName filename = file.name();
```
Swiftb0y, line 124: I think the best way to fix this would be something like this:
```cpp
#ifdef _WIN32
TagLib::FileName filename_owning = file.name();
auto filename = std::basic_string_view(filename_owning.wstr());
#else 
auto filename = std::basic_string_view(file.name());
#endif
```
this is polymorphic over the CharT thanks to CTAD and does not involve conversion. Could be even simpler if we use just convert to `char*` on windows directly, or was there a reason to use the wide variant?

Swiftb0y (changes requested):
please fix the compile errors and amend.

src/track/taglib/trackmetadata_file.cpp
Swiftb0y: I wonder what implications the unfortunate API design has for ownership? Eg is `file.name()` just leaking the string if not on windows?

src/track/taglib/trackmetadata_file.cpp
```diff
@@ -120,7 +120,8 @@ bool readAudioPropertiesFromFile(
     // we must pick one explicit,
     // to prevent "use of overloaded operator '<<' is ambiguous" error
     // on clang-cl builds.
-    const std::wstring& filename = file.name().wstr();
+    TagLib::FileName filename_owning = file.name();
+    const std::wstring& filename = filename_owning.wstr();
```
Swiftb0y, line 124: Please add a comment. IMO this is a common enough pitfall that we might also want to consider raising a bug upstream.
```suggestion
    // We need to save the filename here because if it was chained 
    // (`file.name().wstr()`) the wstr() result would be dangling. 
    TagLib::FileName filename_owning = file.name();
    const std::wstring& filename = filename_owning.wstr();
```

Swiftb0y (changes requested):

src/track/taglib/trackmetadata_file.cpp
daschuer: file.name() is just a non owning pointer to the original string on Linux/MacOS
I have kept the native types for filename to have this issue more visible.

Swiftb0y: okay, so no leaks, just extra memory allocation on windows it seems.

Swiftb0y (approved):
LGTM. waiting for CI

Swiftb0y:
I should start caring about the whitespace in my suggestions. These pre-commit trailing whitespace failures are annoying. I'm sorry. Can you please fix and amend? Thank you.

Swiftb0y (approved):
LGTM thank you.

PR #11968, opened 2023-09-11
Remove offensive language (Pt. 4)

Holzhaus:
This is part of https://github.com/mixxxdj/mixxx/issues/11931 and contains mostly tooltip and skin changes, as well as replacing some remaining occurrences.


Holzhaus:
Do we need an alias system for tooltip IDs for backwards-compat with third-party skins? I hope not.

ronso0:
I don't think so. Would be nice to have, but the effort is not worth it IMO. There's a skin warning logged in case of 'invalid' tooltip ids (which needlessly contains the entire file path of legacyskinparser.cpp).
What might be helpful for 3rd-party skin users/developers would be a link to the respective tooltip migration commit (in the changelog maybe?), because, for tooltip ids, we probably won't have something like the 'deprecated, use XX instead' hint in manual > Mixxx controls, it's too much effort and tbh I don't know where to put that so it's actually more discoverable than a link in the changelog.

Whta do you think?

ronso0:
Apart from that the changes LGTM (as good as they can when looking at a diff of 112 files ;), and I'm no checking the skins for obvious regressions.

src/track/serato/color.cpp
```diff
@@ -27,7 +27,7 @@ namespace mixxx {
 /// libraries, we need to convert between the two values.
 ///
 /// See this for details:
-/// https://github.com/Holzhaus/serato-tags/blob/master/docs/colors.md#track-colors
+/// https://github.com/Holzhaus/serato-tags/blob/main/docs/colors.md#track-colors
```
ronso0, line 30: Btw, wouldn't it be better to move this to a Mixxx repo/branch?

ronso0 (approved):
Thank you, LGTM

PR #11980, opened 2023-09-13
CO Renaming (Pt. 1)

Holzhaus:
*This PR is part of the ongoing effort to remove offensive language from the Mixxx codebase (#11931).*

Here is the first bunch of renamings discussed in https://github.com/mixxxdj/mixxx/pull/11960#issuecomment-1714591738. I removed the `num_mics_configured` CO since it's used nowhere outside the `EngineMixer` class (not even mappings or skins) and can be replaced with a simple member variable.

| Old Name                                                | New Name                                              | Alias?
| ------------------------------------------------------- | ----------------------------------------------------- | ----------
| `[Master],guiTick50ms`                                  | `[App],gui_tick_50ms_period_s`                             | :heavy_check_mark: (although undocumented)
| `[Master],guiTickTime`                                  | `[App],gui_tick_full_period_s`                                 | :heavy_check_mark: (although undocumented)
| `[Master],indicator_250millis`                          | `[App],indicator_250ms`                           | :heavy_check_mark:
| `[Master],indicator_500millis`                          | `[App],indicator_500ms`                           | :heavy_check_mark:
| `[Master],keylock_engine`                               | `[App],keylock_engine`                                | :x:
| `[Master],num_auxiliaries`                              | `[App],num_auxiliaries`                               | :heavy_check_mark: (although undocumented)
| `[Master],num_decks`                                    | `[App],num_decks`                                     | :heavy_check_mark:
| `[Master],num_microphones`                              | `[App],num_microphones`                               | :heavy_check_mark: (although undocumented)
| `[Master],num_mics_configured`                          | **Removed**                                           | n/A
| `[Master],num_preview_decks`                            | `[App],num_preview_decks`                             | :heavy_check_mark:
| `[Master],num_samplers`                                 | `[App],num_samplers`                                  | :heavy_check_mark:
| `[Master],samplerate`                                   | `[App],samplerate`                                    | :heavy_check_mark:


Swiftb0y:
While were at it, I'd also like to propose to retire the `millis` suffix.  We had that as a `suffix` for constants and member variables for some time, but this will get obsolete with `<chrono>` types and literals fast. So I'd propose to stick with the `ms` actually since that mirrors `std::chrono_literals`. Wdyt?

src/engine/enginebuffer.cpp
```diff
@@ -260,7 +260,7 @@ EngineBuffer::EngineBuffer(const QString& group,
                                                m_pLoopingControl);
     m_pReadAheadManager->addRateControl(m_pRateControl);
 
-    m_pKeylockEngine = new ControlProxy("[Master]", "keylock_engine", this);
+    m_pKeylockEngine = new ControlProxy(kAppGroup, QStringLiteral("keylock_engine"), this);
```
Swiftb0y, line 263: unrelated to the change, but we might want to consider switching `ConfigKey` from `QString` to `QLatin1String` as those can be significantly faster apparently and I don't think we need/want Unicode support in COs, do we?

src/engine/enginemixer.h
```diff
@@ -336,6 +336,7 @@ class EngineMixer : public QObject, public AudioSource {
     CSAMPLE_GAIN m_headphoneGainOld;
     CSAMPLE_GAIN m_balleftOld;
     CSAMPLE_GAIN m_balrightOld;
+    std::atomic<unsigned int> m_numMicsConfigured;
```
Swiftb0y, line 339: Did you check that this is needed due to concurrent access or was this just a safe guess?

src/test/autodjprocessor_test.cpp
```diff
@@ -124,10 +127,10 @@ class FakeDeck : public BaseTrackPlayer {
...
   public:
     MockPlayerManager()
-            : numDecks(ConfigKey("[Master]", "num_decks"), true),
-              numSamplers(ConfigKey("[Master]", "num_samplers"), true),
-              numPreviewDecks(ConfigKey("[Master]", "num_preview_decks"),
-                              true) {
+            : numDecks(ConfigKey(kAppGroup, QStringLiteral("num_decks")), true),
+              numSamplers(ConfigKey(kAppGroup, QStringLiteral("num_samplers")), true),
+              numPreviewDecks(ConfigKey(kAppGroup, QStringLiteral("num_preview_decks")),
+                      true) {
```
JoergAtGithub, line 133: Please only add a test for the new CO name and leave the existing one as it is

JoergAtGithub (changes requested):

src/engine/enginemixer.h
Holzhaus: `EngineMixer::process` reads the CO in the engine thread and I think the `onInputConnected`/`onInputDisconnected` methods that modify the value are likely triggered from the GUI thread (although I didn't check).

src/engine/enginebuffer.cpp
Holzhaus: This makes sense, but I'd rather not do this here, because it would mean we need to change *a lot* of COs. This can go into `main` IMHO.

Holzhaus:
> While were at it, I'd also like to propose to retire the `millis` suffix.  We had that as a `suffix` for constants and member variables for some time, but this will get obsolete with `<chrono>` types and literals fast. So I'd propose to stick with the `ms` actually since that mirrors `std::chrono_literals`. Wdyt?

Okay for me. By the way, should I also rename `gui_tick_time` to `gui_tick_s` then?

src/engine/enginebuffer.cpp
Swiftb0y: Yes, I agree. Just wanted to raise the point so I don't forget about it.

Swiftb0y:
> Okay for me. By the way, should I also rename gui_tick_time to gui_tick_s then? I think it's in seconds if I remember correctly.

Sure. though I didn't check either, never heard of that CO before tbh...

src/test/autodjprocessor_test.cpp
Holzhaus: The plan was to remove the word from the codebase as much as possible which includes tests. Also, these tests are supposed to check if the AutoDJ processor works correctly, not if old CO names still work. I agree that it makes sense to test this, too, but that should be a different test IMHO. 

I was planning to add a new test that checks if the aliases exist. This would ideally go into `coreservicestest.cpp` since most other tests do not create all COs. Unfortunately this test is broken at the moment. I'm currently looking into it.

Swiftb0y: this would / could result in warnings raised by tests though if we decided to implement the warnings on deprecated CO access.

JoergAtGithub: We need to ensure the backward compatibility of our legacy API. I'm open for any test implementation, that ensures this.

daschuer:
'guiTickTime' and 'guiTick50ms' have are both in seconds in terms of std::chrono. 
The difference is that the guiTickTime is advanced in refresh rate steps and guiTick50ms is advanced in 50ms steps. The diffrence is the period in terms of chono. 

So we may think of
'gui_full_period_tick_s'
'gui_50ms_period_tick_s'

daschuer:
Derived form that are
'indicator_250millis'
'indicator_500millis'

They just cycle 0.0 ad 1.0. 
Rename as well?

src/widget/wtracktableview.cpp
```diff
@@ -59,7 +59,8 @@ WTrackTableView::WTrackTableView(QWidget* parent,
     // Connect slots and signals to make the world go 'round.
     connect(this, &WTrackTableView::doubleClicked, this, &WTrackTableView::slotMouseDoubleClicked);
 
-    m_pCOTGuiTick = new ControlProxy("[Master]", "guiTick50ms", this);
+    m_pCOTGuiTick = new ControlProxy(
+            QStringLiteral("[App]"), QStringLiteral("gui_tick_50ms_period_s"), this);
     m_pCOTGuiTick->connectValueChanged(this, &WTrackTableView::slotGuiTick50ms);
```
Swiftb0y, line 64: this dependency is... interesting... worrisome? 
Anyways, obsolete with QML.

res/controllers/Traktor-Kontrol-S4-MK3.js
```diff
@@ -2885,7 +2885,7 @@ class S4MK3 {
 
         const that = this;
         /* eslint no-unused-vars: "off" */
-        const meterConnection = engine.makeConnection("[Master]", "guiTick50ms", function(_value) {
+        const meterConnection = engine.makeConnection("[App]", "gui_tick_50ms_period_s", function(_value) {
```
Swiftb0y, line 2888: we should probably track this in an issue on the QML project board, otherwise this will silently break under QML. putting this on an `engine.beginTimer` instead should be sufficient.

Swiftb0y (approved):
LGTM

Swiftb0y (approved):
LGTM. last chance for anyone to object.

Holzhaus:
I was planning to add tests for the legacy aliases as @JoergAtGithub requested, but at the `CoreServicesTest` does not work on CI. I'll try to find another way.

JoergAtGithub:
> I was planning to add tests for the legacy aliases as @JoergAtGithub requested, but at the `CoreServicesTest` (which would be the ideal location for such a test because it creates all COs) does not work on CI. I'll try to find another way.

I appreciate your effort! I think it's safer to have this test case in place before merging this. Otherwise this is LGTM!

src/test/controlobjectaliastest.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+namespace {
+
+const QString kAppGroup = QStringLiteral("[App]");
+const QString kLegacyGroup = QStringLiteral("[Master]");
+
+class ControlObjectAliasTest : public MixxxTest {
+};
+
+#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
```
Holzhaus, line 24: I think this needs to be removed when merging into main, right?

Holzhaus:
> > I was planning to add tests for the legacy aliases as @JoergAtGithub requested, but at the `CoreServicesTest` (which would be the ideal location for such a test because it creates all COs) does not work on CI. I'll try to find another way.
> 
> I appreciate your effort! I think it's safer to have this test case in place before merging this. Otherwise this is LGTM!

Tests are in place now.

JoergAtGithub (approved):
LGTM! Waiting for CI

JoergAtGithub:
Thank you!

Swiftb0y:
Thank you as well.

Holzhaus:
https://github.com/mixxxdj/manual/pull/579

PR #4139, opened 2021-07-24
Link time asserts

daschuer:
Finding a way to improve debug asserts in const expressions I have developed this PR as a drive by. 

This extends our debug asserts to fail linking if the likely good branch is optimized away and not reachable. 

This utilizes the __builtin_constant_p() build in function available on gcc and clang.

Unfortunately it does not work in case the DEBUG_ASSERT() statement is duplicated into two branches during optimization. 
For these case I have added the old version as  RUNTIME_DEBUG_ASSERT() 

Things like DEBUG_ASSERT(false) do no longer work, because they have no good branch and will fail. 
For this purpose I have added DEBUG_ASSERT_UNREACHABLE(false)

The good thing is that this works without all constexpr keywords and uses the feature the compiler uses anyway in O3. 
The bad thing is that it relies on the compiler optimization.  

What do you think about it?

PR #11981, opened 2023-09-14
Refactor(parented_ptr): make trivially destructible in release mode, delete move operations

Swiftb0y:
not ready yet, just did this for fun and an experiment. wdyt about this? It's obviously not super necessary but it may take of even more burden of a reviewer when they see a `make_parented`, knowing that it will only work if the object is actually capable of participating in the Object Tree memory management and not just taking a `QObject` by chance. 
Thoughts and reviews welcome even though this is just a draft.

### DIFF
--- a/src/util/parented_ptr.h
+++ b/src/util/parented_ptr.h
@@ -36,8 +36,11 @@ class parented_ptr final {
     // explicitly generate trivial destructor (since decltype(m_ptr) is not a class type)
     ~parented_ptr() noexcept = default;
 #endif
+    // Rule of 5
     parented_ptr(const parented_ptr<T>&) = delete;
     parented_ptr& operator=(const parented_ptr<T>&) = delete;
+    parented_ptr(const parented_ptr<T>&& other) = delete;
+    parented_ptr& operator=(const parented_ptr<T>&& other) = delete;
 
     // If U* is convertible to T* then parented_ptr<U> is convertible to parented_ptr<T>
     template<
--- a/src/util/parented_ptr.h
+++ b/src/util/parented_ptr.h
@@ -26,12 +26,16 @@ class parented_ptr final {
     explicit parented_ptr(T* t) noexcept
             : m_ptr{t} {
     }
-
+// Only generate destructor if not empty, otherwise its empty but will
+// cause the parented_ptr to be not trivially destructible even though it could be.
+#ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
     ~parented_ptr() noexcept {
         DEBUG_ASSERT(!m_ptr || static_cast<const QObject*>(m_ptr)->parent());
     }
-
-    // Delete copy constructor and copy assignment operator
+#else
+    // explicitly generate trivial destructor (since decltype(m_ptr) is not a class type)
+    ~parented_ptr() noexcept = default;
+#endif
     parented_ptr(const parented_ptr<T>&) = delete;
     parented_ptr& operator=(const parented_ptr<T>&) = delete;
 
--- a/src/widget/findonwebmenufactory.cpp
+++ b/src/widget/findonwebmenufactory.cpp
@@ -11,14 +11,9 @@ namespace mixxx {
 namespace library {
 
 void createFindOnWebSubmenus(QMenu* pFindOnWebMenu, const Track& track) {
-    auto pFindOnWebMenuSoundcloud = make_parented<QMenu>(
-            new FindOnWebMenuSoundcloud(pFindOnWebMenu, track));
-
-    auto pFindOnWebMenuDiscogs = make_parented<QMenu>(
-            new FindOnWebMenuDiscogs(pFindOnWebMenu, track));
-
-    auto pFindOnWebMenuLastfm = make_parented<QMenu>(
-            new FindOnWebMenuLastfm(pFindOnWebMenu, track));
+    make_parented<QMenu>(new FindOnWebMenuSoundcloud(pFindOnWebMenu, track));
+    make_parented<QMenu>(new FindOnWebMenuDiscogs(pFindOnWebMenu, track));
+    make_parented<QMenu>(new FindOnWebMenuLastfm(pFindOnWebMenu, track));
 }
 
 } // namespace library


daschuer:
I did not understand the `std::is_invocable_r_v` constrains.

Swiftb0y:
They currently don't work either, but the idea was to confirm that there exists a constructor for the given args. Ideally those args could also be checked to contain at least one QObject (the parent argument), but I didn't figure out how to check that in a reasonable and performant fashion. I wanted to avoid the case where people would use `make_parented` on a class that supports it, but then don't actually pass a parent object, which would create a leak.

daschuer:
The parented pointer already checks if the object has a parent. Constructing the object without a parent already fails.

Swiftb0y:
yes, it does in a debug assert, but for one, thats based on a ducktyping which might result in false negatives (can be solved via static_assert with `std::derived_from`) and one the other hand it's at runtime even though a large subset of cases could be detected at compile time. Also, the runtime assertion in the destructor also doesn't protect against the problem when [qobjects are destructed out-of-order](https://doc.qt.io/qt-6/objecttrees.html#construction-destruction-order-of-qobjects).

daschuer:
Do you want to replace the runtime assertion?

Swiftb0y:
No the runtime assertion is fine (mostly, commit is in the works). But the runtime assertion only happens in debug builds on shutdown and is probably only noticed with `—DDEBUG_ASSERTIONS_FATAL`. I want to turn it into a compile time assertion so its easier to spot.

daschuer:
How to detect that the parent pointer is not null at compile time? A not_null pointer wrapped does not help, because it is also a runtime check. Maybe with link time assets #4139 or https://github.com/daschuer/mixxx/tree/not_null_link_assert

Swiftb0y:
> How to detect that the parent pointer is not null at compile time?

you can't reliably. Again, I don't want to remove the runtime time check, I just want to catch a subset of cases earlier. 

I have reservations against link-time asserts so I don't want to go down that route now.

Swiftb0y:
got sick of all the compile errors on incomplete compilers. I'll try again in a year or two when we hopefully have clang 15 on macos... 

So this PR just degenerated to the "misc improvements" part of the title...

src/widget/findonwebmenufactory.cpp
```diff
@@ -11,14 +11,9 @@ namespace mixxx {
...
-            new FindOnWebMenuSoundcloud(pFindOnWebMenu, track));
-
-    auto pFindOnWebMenuDiscogs = make_parented<QMenu>(
-            new FindOnWebMenuDiscogs(pFindOnWebMenu, track));
-
-    auto pFindOnWebMenuLastfm = make_parented<QMenu>(
-            new FindOnWebMenuLastfm(pFindOnWebMenu, track));
+    make_parented<FindOnWebMenuSoundcloud>(pFindOnWebMenu, track);
+    make_parented<FindOnWebMenuDiscogs>(pFindOnWebMenu, track);
+    make_parented<FindOnWebMenuLastfm>(pFindOnWebMenu, track);
```
Swiftb0y, line 16: this removes one QMenu indirection, this should still be valid right?

Swiftb0y: Nevermind, this triggers the parented_ptr debug assert, I'll investigate.

Swiftb0y:
friendly ping, requesting review.

src/util/parented_ptr.h
```diff
@@ -26,14 +26,21 @@ class parented_ptr final {
...
+#ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
     ~parented_ptr() noexcept {
         DEBUG_ASSERT(!m_ptr || static_cast<const QObject*>(m_ptr)->parent());
     }
-
-    // Delete copy constructor and copy assignment operator
+#else
+    // explicitly generate trivial destructor (since decltype(m_ptr) is not a class type)
+    ~parented_ptr() noexcept = default;
+#endif
```
daschuer, line 38: for my understanding '~parented_ptr() = default' and `~parented_ptr() {}` is the same. So I can imagine that this is not needed.

src/util/parented_ptr.h
```diff
@@ -26,14 +26,21 @@ class parented_ptr final {
...
+    // Rule of 5
     parented_ptr(const parented_ptr<T>&) = delete;
     parented_ptr& operator=(const parented_ptr<T>&) = delete;
+    parented_ptr(const parented_ptr<T>&& other) = delete;
+    parented_ptr& operator=(const parented_ptr<T>&& other) = delete;
```
daschuer, line 43: I have never understood why it should be not lowed to copy or move a parented_ptr() 
A copy of the parented pointer can be safely moved around since it is not owning the pointer. It is just telling the programmer: "This is a borrowed pointer owned by the parent object".

daschuer:
> can be solved via static_assert with `std::derived_from`

this was a good idea IMHO, why did you discard it?

src/util/parented_ptr.h
daschuer: Ok, It makes sense to have only one instance checking the ownership. The deleted copy constructor enforces that. 
Unfortunately it prevents also to copy the object that contains it. 
A move will move the responsibility for checking the parent along with the pointer. There is nothing wrong with it. Is it?
What is your opinion here?

src/util/parented_ptr.h
Swiftb0y: No its not equivalent. One if an empty definition which means that its still user provided even if it does nothing, the other one is the default declared one which does not count as user-provided. 

See
https://en.cppreference.com/w/cpp/language/destructor#Trivial_destructor
https://compiler-explorer.com/z/cbG99jbzr

Swiftb0y:
Because its not supported on the outdated apple-clang we're stuck with and working around that resulted in too much of a hassle and IIRC also resulted in false positives (?)

src/util/parented_ptr.h
Swiftb0y: I didn't really check to be honest. I just saw that we didn't follow the rule of five and thus I chose the safe default. We can lift the restrictions if we can be sure that its safe.

Swiftb0y: I have reservations about moving. If we move an object that is owned by its parent, its only safe if the parent knows that it has been moved, do QObjects take care of telling their parent?

daschuer: Yes, it is probably OK, we can change it whenever we find a valid use case for it.

daschuer (approved):
LGTM, Thank you.

Swiftb0y:
Thank you

Issue #11930, opened 2023-09-07
Spinnies black on Windows 10 with Angle and Gallium


daschuer:
### Bug Description

The spinnies are black. All Waveform Types are working. Legacy GL types are loosing frames. 

OpenGL-Version: 3.0 Mesa 12.0.0-rc2 (Gallium 0.4 on llvmpipe (LLVM 3.6, 256 bits))

With --enable-vumetergl, the VU-Meters are black as well. 

In Virtual Box 7.0 

### Version

2.5.0-alpha

### OS

Windows 10

Swiftb0y:
> With --enable-vumetergl, the VU-Meters are back as well.

back or black?

daschuer:
Upsi, black.

JoergAtGithub:
This is not a general problem, on my Windows 11 system with AMD GPU the Qt6 spinnies look good and work smooth:
![grafik](https://github.com/mixxxdj/mixxx/assets/64457745/16e70a8a-306c-4278-b41e-05ea849afa4d)

Swiftb0y:
> This is not a general problem, on my Windows 11 system with AMD GPU the Qt6 spinnies look good and work smooth:

Thanks for confirming. I suspected that the virtualization was at fault here. I guess the next step would to be try gallium on linux and see if the issue persists.

daschuer:
When I replace the GLSL Spinnies here by the non GL Types, it works: 
https://github.com/daschuer/mixxx/commit/1c5f6f0e484cd1d624b2bbd9efad067f27bd8b55

Swiftb0y:
difficult to debug why it fails... especially when the glsl waveforms still work... nothing suspicious in the log?

daschuer:
I have updated to 
4.5 (Compatibility Profile) Mesa 23.0.1 (llvmpipe (LLVM 15.0.7, 256 bits))
The latest opengl32sw.dll + liglapi.dll + liggallium_wgl.dll from VCPKG and it works everything with no regressions. 

The only issue is that this extends the 2.5-rel package by 2 GB and exceeds the GitHub build time limits of 6 hours by far. 

Since it is a plug-In DLL, we don't need anything else from the 2 GB than the three DLLs of together 44.1 MB (packed 16,7 MB) 
Our windows installer is 83 MB 

I can imagine to set up an extra VCPKG build, that produces these DLLs and just add them to the windows installer. 
[...]

PR #11982, opened 2023-09-14
EGL

### LINKED ISSUE #11641: 2.4 doesn't start with QT_OPENGL=angle on Windows
### Bug Description

@m0dB :
I tried to start Mixxx with the environment variable QT_OPENGL set to angle. It fails to start:
```
debug [Main] Set root GL Context widget valid: WInitialGLWidget(0x19544c9bb10) false
critical [Main] Failed to load libEGLd (Das angegebene Modul wurde nicht gefunden.)
warning [Main] QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
debug [Main] Displaying main window
debug [Main] Running Mixxx
critical [Main] Failed to load libEGLd (Das angegebene Modul wurde nicht gefunden.)
warning [Main] QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
warning [Main] QOpenGLWindow::beginPaint: Failed to create context
warning [Main] QOpenGLWindow::beginPaint: Failed to make context current
```

### Version

2.4.0

### OS

Windows11

daschuer:
This brings back the GLES support required for Windows drivers that are ignored by QT and on Single PCB devices like Raspberry PI. 

The spinnies are still black. https://github.com/mixxxdj/mixxx/issues/11930

Fixes https://github.com/mixxxdj/mixxx/issues/11641

### DIFF
--- a/src/shaders/endoftrackshader.cpp
+++ b/src/shaders/endoftrackshader.cpp
@@ -4,9 +4,9 @@ using namespace mixxx;
 
 void EndOfTrackShader::init() {
     QString vertexShaderCode = QStringLiteral(R"--(
-attribute vec4 position;
-attribute float gradient;
-varying float vgradient;
+attribute highp vec4 position;
+attribute highp float gradient;
+varying highp float vgradient;
 void main()
 {
     vgradient = gradient;
@@ -15,12 +15,12 @@ void main()
 )--");
 
     QString fragmentShaderCode = QStringLiteral(R"--(
-uniform vec4 color;
-varying float vgradient;
+uniform highp vec4 color;
+varying highp float vgradient;
 void main()
 {
-    float minAlpha = 0.5 * color.w;
-    float maxAlpha = 0.83 * color.w;
+    float minAlpha = 0.5f * color.w;
+    float maxAlpha = 0.83f * color.w;
     gl_FragColor = vec4(color.xyz, mix(minAlpha, maxAlpha, max(0.,vgradient)));
 }
 )--");
--- a/src/waveform/waveformwidgetfactory.cpp
+++ b/src/waveform/waveformwidgetfactory.cpp
@@ -173,10 +173,13 @@ WaveformWidgetFactory::WaveformWidgetFactory()
             m_openGLVersion = pContext->isOpenGLES() ? "ES " : "";
             m_openGLVersion += majorVersion == 0 ? QString("None") : versionString;
 
-            if (majorVersion * 100 + minorVersion >= 201) {
-                if (pContext->isOpenGLES()) {
+            // Qt5 requires at least OpenGL 2.1 or OpenGL ES 2.0
+            if (pContext->isOpenGLES()) {
+                if (majorVersion * 100 + minorVersion >= 200) {
                     m_openGlesAvailable = true;
-                } else {
+                }
+            } else {
+                if (majorVersion * 100 + minorVersion >= 201) {
                     m_openGlAvailable = true;
                 }
             }
@@ -1188,11 +1191,11 @@ QString WaveformWidgetFactory::buildWidgetDisplayName() const {
 QSurfaceFormat WaveformWidgetFactory::getSurfaceFormat() {
     QSurfaceFormat format;
     // Qt5 requires at least OpenGL 2.1 or OpenGL ES 2.0, default is 2.0
-    format.setVersion(2, 1);
+    // format.setVersion(2, 1);
     // Core and Compatibility contexts have been introduced in openGL 3.2
     // From 3.0 to 3.1 we have implicit the Core profile and Before 3.0 we have the
     // Compatibility profile
-    format.setProfile(QSurfaceFormat::CoreProfile);
+    // format.setProfile(QSurfaceFormat::CoreProfile);
 
     // setSwapInterval sets the application preferred swap interval
     // in minimum number of video frames that are displayed before a buffer swap occurs
--- a/src/waveform/waveformwidgetfactory.cpp
+++ b/src/waveform/waveformwidgetfactory.cpp
@@ -1170,14 +1170,14 @@ QString WaveformWidgetFactory::buildWidgetDisplayName() const {
     if (isLegacy) {
         extras.push_back(tr("legacy"));
     }
-    if (isOpenGlAvailable() || isOpenGlesAvailable()) {
-        if (WaveformT::useOpenGles()) {
-            if (WaveformT::useOpenGLShaders()) {
-                extras.push_back(QStringLiteral("GLSL ES"));
-            } else {
-                extras.push_back(QStringLiteral("GLES"));
-            }
-        } else if (WaveformT::useOpenGLShaders()) {
+    if (isOpenGlesAvailable()) {
+        if (WaveformT::useOpenGLShaders()) {
+            extras.push_back(QStringLiteral("GLSL ES"));
+        } else if (WaveformT::useOpenGles()) {
+            extras.push_back(QStringLiteral("GLES"));
+        }
+    } else if (isOpenGlAvailable()) {
+        if (WaveformT::useOpenGLShaders()) {
             extras.push_back(QStringLiteral("GLSL"));
         } else if (WaveformT::useOpenGl()) {
             extras.push_back(QStringLiteral("GL"));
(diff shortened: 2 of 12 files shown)


src/waveform/waveformwidgetfactory.cpp
```diff
@@ -1168,14 +1171,17 @@ QString WaveformWidgetFactory::buildWidgetDisplayName() const {
         extras.push_back(tr("legacy"));
     }
     if (isOpenGlAvailable() || isOpenGlesAvailable()) {
-        if (WaveformT::useOpenGLShaders()) {
+        if (WaveformT::useOpenGles()) {
+            if (WaveformT::useOpenGLShaders()) {
+                extras.push_back(QStringLiteral("GLSL ES"));
+            } else {
+                extras.push_back(QStringLiteral("GLES"));
```
Swiftb0y, line 1178:
```suggestion
                extras.push_back(QStringLiteral("GL ES"));
```

src/shaders/endoftrackshader.cpp
```diff
@@ -15,12 +15,12 @@ void main()
 )--");
 
     QString fragmentShaderCode = QStringLiteral(R"--(
-uniform vec4 color;
-varying float vgradient;
+uniform highp vec4 color;
+varying highp float vgradient;
```
Swiftb0y, line 19: `highp` support is optional in the fragment shader. `lowp` is sufficient for color afaik.
```suggestion
uniform lowp vec4 color;
varying mediump float vgradient;
```

src/waveform/waveformwidgetfactory.cpp
Swiftb0y, line 1178: we now have a bunch of shaders being compatible with plain openGL and OpenGL ES, what "backend" does it choose by default?

Swiftb0y (changes requested):
Thank you. I'll do a manual test soon. also, iiuc this has nothing to do with EGL and is only concerning GL ES support/compatibility...

src/waveform/waveformwidgetfactory.cpp
daschuer: This is currently not visible anyway but this is taken from: [OpenGL ES or GLES](https://en.wikipedia.org/wiki/OpenGL_ES)

src/shaders/endoftrackshader.cpp
daschuer: I have read that the hardware takes anyway the ALU it has which should be 32 bit for all floats. So I think there is no point in reducing it and risk regression compared to the OpenGL version that is 32 bit anyway. 

If we have a ES hardware that has only 24 bit floats among our targets, we need to reduce all float to mediump.

src/waveform/waveformwidgetfactory.cpp
Swiftb0y: Right, that strategy makes sense, but it does not reflect how the name is chosen here where ES is prioritized just based on what the shader supports.

src/shaders/endoftrackshader.cpp
daschuer: I have no interest maintain two precisions for no reason. Do you have an example device, that is not supported right now? If not, let's learn from the user once rendering fails and save our developer resources here.

src/waveform/waveformwidgetfactory.cpp
daschuer: The name is dynamically changed according to the context that is used.  You will see 
"RGB (GLSL ES)" on a Raberry Pie and "RGB (GLES)" on your Laptop.

src/shaders/endoftrackshader.cpp
Swiftb0y: there is no reason to maintain two precisions. We don't need `highp` for many of these fragment shader uniforms, so lets just choose the safer default.

src/waveform/waveformwidgetfactory.cpp
daschuer: Oh I see that is currently broken. I will fix it.

src/shaders/endoftrackshader.cpp
daschuer: Yes that's why I picked highp. It works like OpenGL and we don't need to test two versions.

daschuer: We can change it at any time, no objections. It just requires tests, I don't want to do right now.

Swiftb0y: Right. My argument is that we change it now because in the case where the GLES waveforms are broken because of `highp` people will just fallback to the cpu waveforms assuming their hardware is incapable, even though it technically would be.

Swiftb0y: The alternative would to be `ifdef` it in the shader, that would likely be most robust, but creates a bunch of duplicated code across the shaders.

daschuer: The ifdef is already there in the Qt shader wrapper. 
`#define highp` 
Removing highp for the GL case.

Swiftb0y: No I meant
```glsl
#ifndef GL_FRAGMENT_PRECISION_HIGH
#define highp mediump
#endif
```
in the fragment shader. Then the shader will use highp if available and fallback if otherwise.

daschuer: It turns out that it is already there behind the scenes: 
https://github.com/qt/qtbase/blob/737dd9595058a457fdd8f336ee8630cb1e2c57bb/src/opengl/qopenglshaderprogram.cpp#L445

Not sure if it is in action though.

Swiftb0y: Ok weird, well I'll take the gamble then and hope it is in action on the platforms where it makes sense.

Swiftb0y (approved):
No regressions on fedora as far as I can tell. LGTM

daschuer:
Thank you.

JoergAtGithub:
I just build 2.4 and now Mixxx does no longer start up under Windows 11, when started from install directory of `x64_legacy` with `$env:QT_OPENGL = 'angle'`
```
warning [Main] QWindowsEGLContext: Failed to create context, eglError: 3004, this: 0x25b91296890
warning [Main] QOpenGLWindow::beginPaint: Failed to create context
warning [Main] QOpenGLWindow::beginPaint: Failed to make context current
```
With `$env:QT_OPENGL = 'desktop'` or  `$env:QT_OPENGL = 'software'` everything works still fine using the AMD OpenGL driver.

Swiftb0y:
Have you considered removing the `QSurfaceFormat::setProfile` call? The qt docs say that it'll be ignored when not supported but maybe that's broken?

JoergAtGithub:
It seem my previous test result was not correct. I cleaned the install directory and installed again, and now get:
```
debug [Main] OpenGL driver version string "OpenGL ES 3.0.0 (ANGLE 2.1.5414 git hash: aa63ea230e0c)", vendor "Google Inc. (AMD)", renderer "ANGLE (AMD, AMD Radeon(TM) Graphics Direct3D11 vs_5_0 ps_5_0, D3D11-31.0.12044.50003)"
debug [Main] Supported OpenGL version: 3.0
```
but later it fails with:
```
warning [Main] QOpenGLShader::compile(Fragment): ERROR: 0:7: '0.75f' : Floating-point suffix unsupported prior to GLSL ES 3.00
ERROR: 0:7: '0.75f' : syntax error

warning [Main] *** Problematic Fragment shader source code ***
#ifdef GL_KHR_blend_equation_advanced
#extension GL_ARB_fragment_coord_conventions : enable
#extension GL_KHR_blend_equation_advanced : enable
#endif
#line 1

uniform sampler2D sampler;
uniform highp vec4 color;
varying highp vec3 vTexcoor;
void main()
{
    gl_FragColor = vec4(color.xyz, texture2D(sampler, vTexcoor.xy) * 0.75f);
}

***
DEBUG ASSERT: "addShaderFromSourceCode( GLShader::Fragment, fragmentShaderCode)" in function void __cdecl mixxx::Shader::load(const class QString &,const class QString &) at C:\Users\Joerg.WORLDWARTWEB\source\repos\JoergAtGithub\mixxx\src\shaders\shader.cpp:24
```

Swiftb0y:
right, thats a clearer error now at least. Can you fix the shaders locally and try again?

PR #11985, opened 2023-09-15
Fixup: Removed unsupported floating-point suffix prior to GLSL ES 3.00

daschuer:
This fixes the issue reported here: 
https://github.com/mixxxdj/mixxx/pull/11982

```
warning [Main] QOpenGLShader::compile(Fragment): ERROR: 0:7: '0.75f' : Floating-point suffix unsupported prior to GLSL ES 3.00
ERROR: 0:7: '0.75f' : syntax error
```

### DIFF
--- a/src/shaders/endoftrackshader.cpp
+++ b/src/shaders/endoftrackshader.cpp
@@ -19,8 +19,8 @@ uniform highp vec4 color;
 varying highp float vgradient;
 void main()
 {
-    float minAlpha = 0.5f * color.w;
-    float maxAlpha = 0.83f * color.w;
+    float minAlpha = 0.5 * color.w;
+    float maxAlpha = 0.83 * color.w;
     gl_FragColor = vec4(color.xyz, mix(minAlpha, maxAlpha, max(0.,vgradient)));
 }
 )--");
--- a/src/shaders/vinylqualityshader.cpp
+++ b/src/shaders/vinylqualityshader.cpp
@@ -21,7 +21,7 @@ uniform highp vec4 color;
 varying highp vec3 vTexcoor;
 void main()
 {
-    gl_FragColor = vec4(color.xyz, texture2D(sampler, vTexcoor.xy) * 0.75f);
+    gl_FragColor = vec4(color.xyz, texture2D(sampler, vTexcoor.xy) * 0.75);
 }
 )--");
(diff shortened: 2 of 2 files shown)


Swiftb0y (approved):
LGTM.

JoergAtGithub:
Now Angle works (with the black spinnies reported in #11930 - this means this is a 2.4 bug and not a Qt6 regression). But now the normal OpenGL operation is broken. Mixxx asserts before the main window is shown:
```
warning [Main] QOpenGLShader::compile(Fragment): ERROR: 0:5: '' :  syntax error, unexpected IDENTIFIER
ERROR: 1 compilation errors.  No code generated.


warning [Main] *** Problematic Fragment shader source code ***
#ifdef GL_KHR_blend_equation_advanced
#extension GL_ARB_fragment_coord_conventions : enable
#extension GL_KHR_blend_equation_advanced : enable
#endif
#define lowp
#define mediump
#define highp
#line 1

uniform highp vec4 color;
varying highp float vgradient;
precision highp float;
void main()
{
    float minAlpha = 0.5 * color.w;
    float maxAlpha = 0.83 * color.w;
    gl_FragColor = vec4(color.xyz, mix(minAlpha, maxAlpha, max(0.,vgradient)));
}

***
DEBUG ASSERT: "addShaderFromSourceCode( GLShader::Fragment, fragmentShaderCode)" in function void __cdecl mixxx::Shader::load(const class QString &,const class QString &) at C:\Users\Joerg.WORLDWARTWEB\source\repos\JoergAtGithub\mixxx\src\shaders\shader.cpp:24
```

JoergAtGithub:
I found out, that we need to specify 
`precision highp float;`
when using OpenGL ES, but not when using normal OpenGL.

daschuer:
Yes, that's clear. But In my tests the Qt boilerplate Code defined highp away. Why does this fail in your case?

daschuer:
Ah I see it is the added:
precision highp float;
This does not work along with 
#define highp 
It will become 
precision float;
and fails.

Swiftb0y:
yeah, so I guess we can't specify the precision via the default `precision precision-qualifier​ type​;` but instead need to specify it manually for every variable so it gets `#define`d away properly?

PR #11987, opened 2023-09-15
Tidy and modernize SampleBuffer

abique:


JoergAtGithub:
Welcome at Mixxx!

abique:
Thank you!
I've signed it.
I'll update the PR with the style fix later tonight.

JoergAtGithub:
Regarding the code style issues, we recommend to install pre-commit locally: https://github.com/mixxxdj/mixxx/wiki/Using-Git#set-up-automatic-code-checking
Alternatively, you could download the patch file from the GitHub action and apply it manually.

abique:
Unfortunately archlinux doesn't have yet the required version of pre-commit, so I'll use the patch for now.
I hope that clang-format does it right, so I'll use that in the mean time.

daschuer:
Which pre-commit version do you have? 

Did you try to Install pre-commit via pip? 
https://github.com/mixxxdj/mixxx/wiki/Using-Git#set-up-automatic-code-checkinng.

If you configure you IDE for clang-fotmat, make sure to reformat only touched lines!

src/util/assert.h
```diff
@@ -56,14 +56,16 @@ inline void mixxx_release_assert(const char* assertion, const char* file, int li
...
+        if (!static_cast<bool>(cond)) [[unlikely]] {                        \
             mixxx_debug_assert(#cond, __FILE__, __LINE__, ASSERT_FUNCTION); \
         }                                                                   \
     while (0)
+#define MIXXX_RELEASE_NOEXCEPT
 #else
 #define DEBUG_ASSERT(cond) \
     do {                   \
     } while (0)
+#define MIXXX_RELEASE_NOEXCEPT noexcept
```
daschuer, line 68: What is the use for this? Since we don't use exception handling in Mixxx, does it at any value?

src/util/sample.h
```diff
@@ -83,53 +83,52 @@ class SampleUtil {
     }
 
     // Limits a CSAMPLE value to the valid range [-CSAMPLE_PEAK, CSAMPLE_PEAK]
-    inline static CSAMPLE clampSample(CSAMPLE in) {
+    [[nodiscard]] inline static CSAMPLE clampSample(CSAMPLE in) {
```
daschuer, line 86: Please remove here [[nodiscard]] 

src/util/sample.h
```diff
@@ -27,7 +27,7 @@ class SampleUtil {
 
     // Allocated a buffer of CSAMPLE's with length size. Ensures that the buffer
     // is 16-byte aligned for SSE enhancement.
-    static CSAMPLE* alloc(SINT size);
+    [[nodiscard]] static CSAMPLE* alloc(SINT size);
```
daschuer, line 30: This is a welcome place for [[nodiscard]] btw.

src/util/samplebuffer.cpp
```diff
@@ -12,6 +11,17 @@ SampleBuffer::SampleBuffer(SINT size)
 
 SampleBuffer::~SampleBuffer() {
     SampleUtil::free(m_data);
+
+    /* cleanup the memory, just in case of use after free */
+    m_data = nullptr;
+    m_size = 0;
```
daschuer, line 17: Using it after destruct will crash anyway. Please remove that unnecessary code.

src/util/samplebuffer.cpp
```diff
@@ -12,6 +11,17 @@ SampleBuffer::SampleBuffer(SINT size)
...
+}
+
+SampleBuffer& SampleBuffer::operator=(SampleBuffer&& that) {
```
daschuer, line 20: What is the use case for this? Do we also want a move constructor?

src/util/samplebuffer.h
```diff
@@ -164,33 +163,40 @@ class SampleBuffer final {
...
-        CSAMPLE& operator[](SINT index) const {
+        [[nodiscard]] CSAMPLE& operator[](SINT index) const MIXXX_RELEASE_NOEXCEPT {
             return *data(index);
         }
+
       private:
-        CSAMPLE* m_data;
-        SINT m_length;
+        CSAMPLE* m_data = nullptr;
+        SINT m_length = 0;
```
daschuer, line 194: We prefer the initialization within the constructor: 
https://github.com/mixxxdj/mixxx/wiki/Coding-Guidelines#non-static-data-member-initializers

src/util/samplebuffer.h
```diff
@@ -199,7 +205,7 @@ namespace std {
 
 // Template specialization of std::swap() for SampleBuffer
 template<>
-inline void swap(::mixxx::SampleBuffer& lhs, ::mixxx::SampleBuffer& rhs) {
+inline void swap(::mixxx::SampleBuffer& lhs, ::mixxx::SampleBuffer& rhs) /* noexcept */ {
```
daschuer, line 208: Why is noexcept commented out?

daschuer (changes requested):

src/util/samplebuffer.h
abique: Doesn't compile on mac, but probably will in the future.

src/util/samplebuffer.cpp
abique: No, it isn't true.
If you use it after destruct, it might just point to the heap and corrupt something else which will make it even harder to debug.

src/util/samplebuffer.h
daschuer: What is the error message. 
Please add an explanation as comment.

src/util/sample.h
abique: Indeed.
Having `[[nodiscard]]` has an habit will likely save you more than it'll cost you.

src/util/assert.h
abique: The DEBUG_ASSERT macro, uses lots of external C++ code (Qt), which may throw, std::bad_alloc for example or maybe on I/O error when printing to the log I suppose. Unless you trust you assert macro to absolutely never throw?

src/util/samplebuffer.cpp
daschuer: Ok, than please add a comment to the code explaining the reason.

src/util/assert.h
abique: Indeed Qt doesn't throw exceptions so it was unnecessary. I remove it.

src/util/sample.h
abique: I've removed them.

src/util/samplebuffer.cpp
abique: I think there is one already, but I'll elaborate further more.

abique: I've added a longer explanation, if you disagree I won't discuss it further and simply remove the pointer reset.

src/util/samplebuffer.cpp
abique: A simple use case:

```C++
class MyProcessor
{
   SampleBuffer m_b;
   
   void activate(double samplerate, int blockSize)
   {
      m_b = SampleBuffer(2 * blockSize);
   }
};

Comment (abique) [reply src/util/samplebuffer.cpp:23]: There is already a move ctor.

File: src/util/sample.h (hunk 2 above)
Comment (abique) [reply src/util/sample.h]: Done.

File: src/util/samplebuffer.h (hunk 7 above)
Comment (abique) [reply src/util/samplebuffer.h]: Done.

File: src/util/samplebuffer.h (hunk 6 above)
Comment (abique) [reply src/util/samplebuffer.h]: done

File: src/util/samplebuffer.cpp (hunk 4 above)
Comment (Swiftb0y) [reply src/util/samplebuffer.cpp:20]: Still an unnecessary performance cost. This only makes sense from a practical standpoint, but accessing the buffer after the destructor is UB anyways so its still questionable whether this makes much sense from a theoretical perspective.

File: src/util/samplebuffer.h [hunk 8]
```diff
@@ -34,64 +34,65 @@ namespace mixxx {
...
     }
-    virtual ~SampleBuffer() final;
+    ~SampleBuffer();
 
     SampleBuffer& operator=(SampleBuffer& that) = delete;
-    SampleBuffer& operator=(SampleBuffer&& that) {
-        swap(that);
-        return *this;
-    }
+    SampleBuffer& operator=(SampleBuffer&& that);
```
Swiftb0y, line 50: is a particular reason you chose to put the move ctor in the header but not the move-assignment?

src/util/samplebuffer.h
```diff
@@ -34,64 +34,65 @@ namespace mixxx {
...
         return m_size;
     }
 
-    CSAMPLE* data(SINT offset = 0) {
+    constexpr CSAMPLE* data() noexcept {
+        return m_data;
+    }
+    constexpr const CSAMPLE* data() const noexcept {
+        return m_data;
+    }
```
Swiftb0y, line 61: in the future we should mark these `[[deprecated]]` because the span API is just better, but that is out-of-scope for this PR.

src/util/samplebuffer.cpp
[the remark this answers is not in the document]
abique: OK, done.

src/util/samplebuffer.h
abique: I move it to the .cpp

Swiftb0y: Well I would've done it the other way around. Seems like the compiler could benefit from knowing the implementation (being able to inline, etc). I'm just curious why you preferred it in the .cpp?

src/util/samplebuffer.h
abique: Actually the span isn't a good option too.

I suggest a custom implemented span, which also keeps the following info:
- samplerate of the buffer
- channel count
- frame count (instead of simply size)

This isn't essential, but very handy in many situations, if you want to add assertions or you're debugging the app and want to understand the program while in the debugger.

abique:
> Which pre-commit version do you have?
> 
> Did you try to Install pre-commit via pip? https://github.com/mixxxdj/mixxx/wiki/Using-Git#set-up-automatic-code-checkinng.
> 
> If you configure you IDE for clang-fotmat, make sure to reformat only touched lines!

Archlinux has the x.20.x instead of the x.21.x and I've tried pip, but it wanted a system wide install, etc... so I didn't want to go that far.

src/util/samplebuffer.h
Swiftb0y: Yes, that will come in the future. We want an abstraction based on `mdspan` that keeps track of all that, but the current API does not offer anything better than `span` so we might as well use that.

Swiftb0y:
have you tried installing pre-commit via pip in a venv?

src/util/samplebuffer.h
abique: I didn't think about it, I just wrote it were my cursor was.

abique: At the end of the day, if you build with IPO then it is more or less the same right?

abique:
> have you tried installing pre-commit via pip in a venv?

No, I'm short on disk space and I don't want to start hacking my system.
I'll wait for Archlinux to pickup the update.

src/util/samplebuffer.h
Swiftb0y: Probably, but we currently don't build with LTO afaik, I think we tried but it caused some issues. Especially on a codebase as large as mixxx. We already switched `mold` because linking times are quite long out-of-the-box, using LTO wouldn't help.

abique: Then, all the SampleBuffer.cpp could go into the header.

abique: If you don't mind, I'll leave it as it is?

Swiftb0y:
Thats fine too, as long as you're fine with applying the pre-commit patches from CI for now. I'm confused why its not newer though. pre-commit 2.20.0 has been released more than a year ago and arch is a rolling distro, right? so it should've been picked up by now if there is not technical reason why it couldn't.

abique:
> Thats fine too, as long as you're fine with applying the pre-commit patches from CI for now. I'm confused why its not newer though. pre-commit 2.20.0 has been released more than a year ago and arch is a rolling distro, right? so it should've been picked up by now if there is not technical reason why it couldn't.

It is because the maintainer is probably lacking time to update the package.
I think that arch is working on allowing PR for updating packages, which could help maybe let's see.
In general it is quite up to date, but sometimes arch is behind.

src/util/samplebuffer.h
abique: I doubt that any of the SampleBuffer.cpp is used from a hotspot except fill() and clear() maybe.

Swiftb0y: > Then, all the SampleBuffer.cpp could go into the header.

Frankly, yes...

> If you don't mind, I'll leave it as it is?

I don't really mind tbh, but that kinda feels like its been half done. I think samplebuffer is in a much better state already thanks to your efforts, and this would be just be the last step to make it consistent.

abique: I've done it.
Waiting for my computer to build it... taking at least 10 minutes it seems.

abique:
@daschuer @Swiftb0y thank you very much for the review.
I think we're good now.

src/util/samplebuffer.h
abique: Just use `SampleBuffer::ReadSlice`, etc... so that way when you add more info to the sample buffer they'll be naturally present in the read slice.

Looking at `mdspan` I don't see it better than ReadSlice or WriteSlice.

Swiftb0y:
Thank you. Can you please ensure pre-commit passes now? If you don't want / can't install it, you can find a patch file generated by CI in the pre-commit job. In this case https://github.com/mixxxdj/mixxx/actions/runs/6208187984

abique:
> Thank you. Can you please ensure pre-commit passes now? If you don't want / can't install it, you can find a patch file generated by CI in the pre-commit job. In this case https://github.com/mixxxdj/mixxx/actions/runs/6208187984

Yes I did apply this patch. However, it feels weird how much it indents the clobber list, feels like it has 1 extra level of indentation.

Swiftb0y:
I assume you mean `ConstructorInitializerIndentWidth: 8`? Yeah that choice is weird but as far as I can tell its based on the style-guide from long before we used clang-format (or even before the majority of currently active devs even worked on mixxx...)

Swiftb0y (approved):
LGTM. Thank you. Your contributions are highly appreciated. waiting for LGTM from @daschuer

src/util/samplebuffer.h
```diff
@@ -34,79 +35,94 @@ namespace mixxx {
...
     SampleBuffer& operator=(SampleBuffer&& that) {
-        swap(that);
+        SampleUtil::free(m_data);
+        m_data = std::exchange(that.m_data, nullptr);
+        m_size = std::exchange(that.m_size, 0);
         return *this;
     }
 
-    constexpr SINT size() const {
+    constexpr SINT size() const noexcept {
```
daschuer, line 63: I am confused about the use of noexept in this class. IMHO it has a little value in this case since the function will be inlined anyway. 
On the other hand you did not add noexcept to the move constructors and move assignment operator as explicit recommended in the "C++ Core Guidelines" and other places. In these cases it may help std:: containers to optimize the code.

We do not use noexcept everywhere is Mixxx rightnow, because Mixxx does not use exceptions. So this is a new topic for us, which is ortogonal to to the main topic of this PR. Introducing noexcept to Mixxx can be considered as unnecessary boilerplate on one hand that drew developer resources on the other hand it may help under certain conditions.  

In practice, many codebases do not use noexcept on trivial getter functions, and it's not considered a best practice to do so unless there are specific reasons. The Mixxx style guide references the Google style guild here which is also kept weak: "Specify noexcept when it is useful and correct." 
  
What do you think, is it "useful" here?

daschuer (changes requested):

src/util/samplebuffer.h
Holzhaus: I also made a remark on that above. I would not have added it, but it's not wrong here.

The main culprit is that it's not really related to the PR and therefore increases the diff size which is a non-issue here because @Swiftb0y already reviewed it.

You "requested changes" in your review. Do you think it's absolutely necessary to remove these before merge? Otherwise I'd like to spare @abique the additional work and just keep it.

daschuer: I have requested a change, because it looks wrong to add noexcept to all functions in the class but not to the move constructors. 

I have no experience with except but I have read a bit and discovered three established styles: 
* Use noexcept everywhere 
 ** when using exceptions, can be errorphrone, extra boilerplate to maintain. 
* Use noexcept where useful
 ** unclear what useful is, google style 
* Use noexcept only on move assignment operators an constructors
** to allow optimazion in the std lib 
* don't use noexcept if no exceptions are used. 

I like to know at least what is the reasoning for the current changes. @abique

abique: The sample buffer class is meant to be used in real-time context.
Having the noexcept keywords means both that it should not throw exception (for the dev writting code), and that it won't throw exception (for the compiler).

Indeed I didn't annotate the move ctor and assignment, it is an overlook.
Though the assignment one, calls a free function which wasn't annotated itself, so I wasn't 100% sure about that one, but yes it unlikely that the free() will be the one to throw. And yes, having those two annotated will be useful if you have vectors of audio buffers which you resize.

abique: I don't claim to be a killer C++ dev, who knows everything.
But if the standard C++ library use those, there is certainly a good reason.

daschuer: > But if the standard C++ library use those, there is certainly a good reason.

We are no library, so different reasons apply here. 

After reading ab it about noexecpt it should be a good middle ground to use noexcept for all move ctor and assignments and not clutter the code with the noexcept everywhere. 

In this case I don't mind about he extra noexcept in this PR, but please add it to move ctors and assignments

abique: Done.

daschuer: I have started a topic in Zulip to fix the decision in our style guide: 
https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/noexcept

src/util/samplebuffer.h
```diff
@@ -34,79 +35,94 @@ namespace mixxx {
 //
 class SampleBuffer final {
   public:
-    SampleBuffer()
-        : m_data(nullptr),
-          m_size(0) {
+    constexpr SampleBuffer()
```
Swiftb0y, line 38:
```suggestion
    constexpr SampleBuffer() noexcept
```

src/util/samplebuffer.h
```diff
@@ -34,79 +35,94 @@ namespace mixxx {
...
-    SampleBuffer(SampleBuffer&& that)
-        : m_data(that.m_data),
-          m_size(that.m_size) {
-        that.m_data = nullptr;
-        that.m_size = 0;
+    SampleBuffer(SampleBuffer&& that) noexcept
+            : m_data(std::exchange(that.m_data, nullptr)),
+              m_size(std::exchange(that.m_size, 0)) {
+    }
+    ~SampleBuffer() {
```
Swiftb0y, line 51: TIL destructors are `noexcept` by default since C++11, so even if we don't add it here, the destructor would `std::terminate` be default if `SampleUtil::free` would panic.

daschuer (approved):
We have now refined our rules regarding noexcept in https://github.com/mixxxdj/mixxx/wiki/Coding-Guidelines#specifier-nexecept. 
So here we have some extra noexecpt now. I will merge this anyway, because this PR has been started before. We can fix the style later. 
Thank you @abique for the work!

abique:
Hi,

Thanks for merging!

Sorry for all the noise with the annotations.

Cheers,
Alex.

Holzhaus:
> > have you tried installing pre-commit via pip in a venv?
> 
> No, I'm short on disk space and I don't want to start hacking my system.
> I'll wait for Archlinux to pickup the update.

Just wanted to let you know that I reached out to the package maintainer. The package was updated now.

abique:
> Just wanted to let you know that I reached out to the package maintainer. The package was updated now.

Yes I've seen that, thanks!

PR #11988, opened 2023-09-16
reduce sample buffer memory usage

abique:
This set of changes reduce the amount of memory used by the application for storing audio samples.

Here two memory dump (SVG), before and after so you can compare.
[pprof-after](https://github.com/mixxxdj/mixxx/assets/31200/2dc6df77-2fe7-417d-9341-833618c8333e)
[pprof-before](https://github.com/mixxxdj/mixxx/assets/31200/bfb9143a-9989-422b-8581-9f17e5b8b287)

I had a look over BalanceGroupState, and this one is safe to shrink.


src/engine/cachingreader/cachingreaderchunk.h
```diff
@@ -15,68 +15,76 @@
...
+  // on the properties of the AudioSource as the remarks above suggest!
+  static const constexpr mixxx::audio::ChannelCount kChannels = mixxx::kEngineChannelCount;
+  static const constexpr SINT kFrames = 8192; // ~ 170 ms at 48 kHz
+  static const constexpr SINT kSamples = kFrames * kChannels;
+
+  // Converts frames to samples
+  [[nodiscard]] static constexpr SINT frames2samples(SINT frames) noexcept {
+      return frames * kChannels;
+  }
+  [[nodiscard]] static constexpr double dFrames2samples(SINT frames) noexcept {
```
daschuer, line 34: We have decided not to maintain [[nodiscard]] in our code. 

abique:
I'll take a look at the compiler errors later today, this is strange my build didn't pick it up.

src/engine/cachingreader/cachingreaderchunk.h
abique: > Use it when not checking the return value is a program error like a memory leak. Don't use it everywhere, like in simple getter functions where the compiler will discard the call if the return value is not used.

How to know if the compiler will discard the call? This feels like a condition that no one can predict, especially when the project is built with many different compilers.

abique: Also, the implementation might be inline and simple at day 1, but maybe day 2, it isn't inline anymore, complicated and not optimized out by the compiler when the return value is discarded.

abique: I don't mean to start trolling your coding convention, if you insist I'll change it so it is conform to what you want. Though I fail to see a disadvantage to a function that is annotated with `[[nodiscard]]` when the return value shouldn't be discarded or the function shouldn't be called.

abique:
I've run the tests and some failed:

```
[  FAILED  ] 7 tests, listed below:
[  FAILED  ] AnalyzerWaveformTest.canary
[  FAILED  ] EngineMixerTest.SingleChannelOutputWorks
[  FAILED  ] EngineMixerTest.SingleChannelPFLOutputWorks
[  FAILED  ] EngineMixerTest.TwoChannelOutputWorks
[  FAILED  ] EngineMixerTest.TwoChannelPFLOutputWorks
[  FAILED  ] EngineMixerTest.ThreeChannelOutputWorks
[  FAILED  ] EngineMixerTest.ThreeChannelPFLOutputWorks
```

Let's see if it is a regression.

abique:
The regression is caused by: `constexpr unsigned int MAX_BUFFER_LEN = 32 * 1024;`

src/engine/cachingreader/cachingreaderchunk.h
daschuer: We have put the rule in our code style after a discussion to avoid discussing it again in reviews.
Your arguments valid to support your opinion. The disadvantage is the effort to maintain `[[nodiscard]]` everywhere to follow a uniform code style.

src/util/defs.h
```diff
@@ -4,7 +4,7 @@
 
 // Maximum buffer length to each EngineObject::process call.
 //TODO: Replace this with mixxx::AudioParameters::bufferSize()
-constexpr unsigned int MAX_BUFFER_LEN = 160000;
+constexpr unsigned int MAX_BUFFER_LEN = 32 * 1024;
```
daschuer, line 7: Please add a comment that describes the considerations for the new value.

daschuer (changes requested):

src/util/defs.h
abique: The thing is that this value is arbitrary.
In zulip, I've read that mix doesn't support processing blocks larger than 4K.
So I'd actually even shrink it down to MAX_FRAMER * MAX_CHANNELS (4096 * 2).

abique:
Actually the problem found by the test suite is in the test itself.
It doesn't correctly compare the output and the reference.

src/engine/cachingreader/cachingreaderchunk.h
abique: Can be added step by step, it doesn't have to be done all at once.

Anyway, I'm not here to make you change your mind. I've reverted that change.

src/util/defs.h
abique: I've added a comment.
I think in the long run this constant should be gone.

src/engine/cachingreader/cachingreaderchunk.h
```diff
@@ -139,8 +147,8 @@ class CachingReaderChunkForOwner: public CachingReaderChunk {
             CachingReaderChunkForOwner** ppTail);
 
 private:
-    State m_state;
+  State m_state = FREE;
 
-    CachingReaderChunkForOwner* m_pPrev; // previous item in double-linked list
-    CachingReaderChunkForOwner* m_pNext; // next item in double-linked list
+  CachingReaderChunkForOwner* m_pPrev = nullptr; // previous item in double-linked list
+  CachingReaderChunkForOwner* m_pNext = nullptr; // next item in double-linked list
```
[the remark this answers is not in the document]
abique: I've reverted it.

Swiftb0y (commented):
sorry for the pedantic comments in regards to our codestyle, but we really want to maintain some consistency here. If there are good reasons to change something we'd love to discuss possible improvements.

src/util/defs.h
```diff
@@ -3,26 +3,35 @@
...
-constexpr unsigned int MAX_BUFFER_LEN = 160000;
+// Note: MAX_BUFFER_LEN shouldn't be use for audio processing,
+// because the buffers should be sized according to the processing
+// block in frames and then multiplied by the channel count.
+// Then audio buffers used in I/O and non realtime operation
+// should be sized according to their task with a balance in mind
+// between CPU efficiency and memory usage.
+//
+// TODO: Replace this with mixxx::AudioParameters::bufferSize()
+static constexpr const unsigned int MAX_PROCESS_FRAMES = 4096;
```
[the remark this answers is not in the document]
abique: I'm not sure that `constexpr` implies `const` in all situation.
I remember reading something about it.

src/util/defs.h
```diff
@@ -3,26 +3,35 @@
...
+static constexpr const unsigned int MAX_PROCESS_CHANNELS = 2;
+static constexpr const unsigned int MAX_BUFFER_LEN = MAX_PROCESS_CHANNELS * MAX_PROCESS_FRAMES;
```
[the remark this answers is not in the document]
abique: OK, but I won't change `MAX_BUFFER_LEN` because it'll trigger too many changes.

src/util/defs.h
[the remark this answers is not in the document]
Swiftb0y: pretty sure it does but I'd gladly to be convinced otherwise. Can you do me a favor and lookup the source?

src/util/defs.h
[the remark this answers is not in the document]
Swiftb0y: Yes thats fine. Maybe add an alias to allow for gradual adoption?

src/util/defs.h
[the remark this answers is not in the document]
abique: I've searched a bit and I think you are correct.
I change it.

src/util/defs.h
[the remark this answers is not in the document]
abique: Done.

abique:
> sorry for the pedantic comments in regards to our codestyle, but we really want to maintain some consistency here. If there are good reasons to change something we'd love to discuss possible improvements.

It is your house, keep it clean. :+1:

abique:
Regarding the max buffer size, a few thing.

1. it can't be too big, otherwise the GUI will feel bad (vu meters don't refresh often for example)
2. latency is bad, especially for live performances; I personally think that 256@44.1 is a safe default, 1024@44.1 already feels like shit.
3. I don't know how often you get the MIDI info, and if you run them ASAP or if you shift them within the block to avoid jitters, but anyway, at large blocks you'll probably feel more "jittery"

I don't think there is a point in working in higher sample rate than 48K except if your HW can't do it.

Then in soundmanagerconfig.h:
```c++
    // Size1xms presents the first buffer size of 2^X
    // that results in a buffer time above 1 ms
    // It is 1.45 ms @ 44.1 kHz
    // The other values are representing the following 2^X sizes.
    enum class AudioBufferSizeIndex {
        Size1xms = 1,
        Size2xms = 2,
        Size5xms = 3,
        Size10xms = 4,
        Size20xms = 5,
        Size40xms = 6,
        Size80xms = 7,
    };

    // Represents the sample rate independent frame/period
    // index values in case of Jack
    enum class JackAudioBufferSizeIndex {
        SizeAuto = 5,
        Size2048fpp = 6,
        Size4096fpp = 7,
    };

static constexpr auto kMaxAudioBufferSizeIndex =
            static_cast<unsigned int>(AudioBufferSizeIndex::Size80xms);
static constexpr auto kDefaultAudioBufferSizeIndex =
            static_cast<unsigned int>(AudioBufferSizeIndex::Size20xms);
```

So at 44.1, min is 64 samples, max 4096.

Now the sample rate:
```C++
    static constexpr value_t kValueMin = 8000;   // lower bound (inclusive, = minimum MP3 sample rate)
    static constexpr value_t kValueMax = 192000; // upper bound (inclusive)
```

You've decided to support up to 192kHz.
According to dlgprefsound.cpp, then you'll let the user choose up to 16384.

Here I suggest two things:
1. Make the max buffer size at 44.1kHz to 1024 frames or at most 2048. Really 1024 already feels like shit, so I don't think that there's anything good in letting the user put himself in a bad experience. I think pipewire's default is 512 at 44.1?
2. Limit the max buffer size to 4K, even if the user is at 192kHz. If you want to run at 192kHz, then you probably can afford a low latency (or just use a lower sampling rate...).

I really think that offering 2K at 44.1kHz is not a service to the user.

Swiftb0y:
I'm not familiar with many parts of the audio engine, but I think we already do support low latency's via low periods, the thing is just that sometimes we just allocate `MAX_BUFFER_SIZE` upfront but only use the fraction defined by the user. Thats why there is high memory size, but still reasonable latency.

> I don't think there is a point in working in higher sample rate than 48K except if your HW can't do it.

That might be true for the majority of users, but some users still want that higher samplerate. Same applies for latency (better to have high latency and no underruns instead of constant underruns). Unless this forces us to make tradeoffs that worsens the experience for the majority, I don't see why we shouldn't support those buffer sizes & samplerates. 

> I really think that offering 2K at 44.1kHz is not a service to the user.

I think its reasonable to expect users to know what settings they're choosing. If people want to use those settings because they might be forced to its better to let them instead of patronizing them in that regard.

abique:
> I think its reasonable to expect users to know what settings they're choosing. If people want to use those settings because they might be forced to its better to let them instead of patronizing them in that regard.

Not necessarily, it could be that someone just wants to try mixing, doesn't know about buffer size and its implications and ends up with a setting leading to a bad user experience.

Having more options doesn't means a better beginner experience.

Let's keep the current settings then, but maybe consider advertising 256@44.1kHz as a good default (or even 128). Currently the drop down menu doesn't give any hint, maybe add an `auto` entry which will pick a good default?

Swiftb0y:
Yes, we should definitely try to choose a good default, but we shouldn't take options away from power users or people that know what they're doing just so beginners don't accidentally choose the wrong setting.

src/engine/cachingreader/cachingreaderchunk.h
```diff
@@ -15,15 +15,23 @@
...
+    // 8192 frames contain about 170 ms of audio at 48 kHz, which
+    // is well above (hopefully) the latencies people are seeing.
+    // At 10 ms latency one chunk is enough for 17 callbacks.
+    // Additionally the chunk size should be a power of 2 for
+    // easier memory alignment.
+    // TODO(XXX): The optimum value of the "constant" kFrames depends
+    // on the properties of the AudioSource as the remarks above suggest!
+    static const constexpr mixxx::audio::ChannelCount kChannels = mixxx::kEngineChannelCount;
+    static const constexpr SINT kFrames = 8192; // ~ 170 ms at 48 kHz
+    static const constexpr SINT kSamples = kFrames * kChannels;
```
Holzhaus, line 28:
```suggestion
    static constexpr mixxx::audio::ChannelCount kChannels = mixxx::kEngineChannelCount;
    static constexpr SINT kFrames = 8192; // ~ 170 ms at 48 kHz
    static constexpr SINT kSamples = kFrames * kChannels;
```

src/engine/cachingreader/cachingreaderchunk.h
```diff
@@ -45,7 +53,7 @@ class CachingReaderChunk {
     CachingReaderChunk(const CachingReaderChunk&) = delete;
     CachingReaderChunk(CachingReaderChunk&&) = delete;
 
-    SINT getIndex() const {
+    SINT getIndex() const noexcept {
```
Holzhaus, line 56: We don't use `noexcept`anywhere. We might change this, but why introduce it in this unrelated PR?

abique: Are you asking me to remove all `noexcepts`?

abique: I didn't read all the coding style document.
When I write code I just write what I consider good C++ with the correct annotations, in order to gives semantic and improves compiler checks and code analysis.
I've a different approach to development than the mixxx team, so after this PR I'm out.

Holzhaus: Not necessarily, but please try to avoid such changes in the future unless you touch the line anyway. That would make the diff easier to review.

Holzhaus: > I didn't read all the coding style document.
> When I write code I just write what I consider good C++ with the correct annotations, in order to gives semantic and improves compiler checks and code analysis.

Sure, and that is fine. It's just easier to grasp the essence of the PR if the diff is small. 

> I've a different approach to development than the mixxx team, so after this PR I'm out.

I'm sorry to hear that, that was not my intention.

src/engine/cachingreader/cachingreaderchunk.h
abique: Thank you for picking up this one.

src/engine/cachingreader/cachingreaderchunk.h
abique: It isn't about what you said specifically, but I didn't enjoy making this change in general. So I don't see myself going into a full refactoring of the audio buffers usage over all the app.

Holzhaus: > It isn't about what you said specifically, but I didn't enjoy making this change in general. 

Would you be interested in discussing your experience on Zulip? Maybe we can improve our DX, our workflows and our policies. Hopefully we can make contributing a more pleasant experience.

src/util/defs.h
```diff
@@ -3,26 +3,39 @@
 #include <Qt>
 
 // Maximum buffer length to each EngineObject::process call.
-//TODO: Replace this with mixxx::AudioParameters::bufferSize()
-constexpr unsigned int MAX_BUFFER_LEN = 160000;
```
daschuer, line 7: This buffer is also used for the scalers. I am a bit concerned if the reduction of buffer size has bad effects there. 
So I think it would be more safe to introduce another macro and replace every usage one by one. 
Maybe we can finally get rid of this fixed number at all.

src/util/defs.h
```diff
@@ -3,26 +3,39 @@
...
+// block in frames and then multiplied by the channel count.
+// Then audio buffers used in I/O and non realtime operation
+// should be sized according to their task with a balance in mind
+// between CPU efficiency and memory usage.
+//
+// Max frames at 44.1kHz:     4096
+// Max frames at  192kHz: 4 * 4096
+//
+// TODO: Replace this with mixxx::AudioParameters::bufferSize()
+static constexpr unsigned int kMaxProcessFrames = 4 * 4096;
```
daschuer, line 17: Where is the 4 coming from? 
Mixxx supports audio buffers up to 80 ms. This is 8192 Frames @ 192kHz

src/util/defs.h
```diff
@@ -3,26 +3,39 @@
...
+static constexpr unsigned int kMaxProcessChannels = 2;
+static constexpr unsigned int kMaxProcessSamples = kMaxProcessChannels * kMaxProcessFrames;
+static constexpr unsigned int MAX_BUFFER_LEN = kMaxProcessSamples;
 
-constexpr int kMaxNumberOfDecks = 4;
+static constexpr int kMaxNumberOfDecks = 4;
```
daschuer, line 22: We use anonymous namepsaces instead of the almost deprecated static keyword for this purpose.

src/test/reference_buffers/SingleChannelOutputWorks-main
daschuer, line 1: What has been changed here? For my understanding the other changes in this commit should not change the test results.

daschuer (changes requested):

src/util/defs.h
uklotzde: Anonymous namespaces in `.cpp` files serve a different purpose and are not applicable here. The `static` keyword serves many different purposes.

Though I guess `static` is redundant here, because the evaluation of the constant could safely be inlined and the value does not need to be stored at a fixed memory location.

Swiftb0y: good point, though at the same time, the compiler might decide not to actually inline nor evaluate a `constexpr` variable at compile time if you omit the `static`, though this would be incredibly unlikely in the case of an `int`.

daschuer: > Anonymous namespaces in .cpp files serve a different purpose and are not applicable here.

Why do you think that? The compiler has no idea of headers. It is the predecessor that builds a giant Cpp file with all headers literally included. So it is zero differences if the anonymous namespace is defined in the header or the cpp file.

src/engine/cachingreader/cachingreaderchunk.h
```diff
@@ -74,9 +82,9 @@ class CachingReaderChunk {
     void init(SINT index);
 
 private:
-    SINT frameIndexOffset() const {
+  SINT frameIndexOffset() const noexcept {
         return m_index * kFrames;
-    }
+  }
```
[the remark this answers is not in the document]
abique: The pre-commit didn't complain, I'm not sure I have to change this.

src/util/defs.h
abique: I've run the app and the test with ASAN, no issues so far.

src/util/defs.h
abique: The 4 is coming from 192 = 4 * 48;
If you say so we can shrink the buffer size, when I've looked at the code 16K was the safe upper bound I've calculated, it is possible that I've overlooked something.

src/util/defs.h
abique: I've added static because I wasn't sure if you don't have it if the compiler will have to export a symbol.

src/test/reference_buffers/SingleChannelOutputWorks-main
abique: It is related to how the test is implemented.
It produces `MAX_BUFFER_LEN` and compares it to the reference, so the ref has to have exactly the same size.

src/util/defs.h
daschuer: Did you use a 192 kHz audio file with extreme low pitch, rubberband and soutdtouch and keylock disabled enabled? 
As far as I know ASAN is able to detect some out of bound accesses but now ASAN findings is no proof that there are no.
An additional theoretic consideration is required IMHO.

abique: You're describing an extreme scenario.
I think what the mixxx team should do is:
- take the PR
- add bounds checking assertions in the place you're specially worried about, and even if possible make the buffer allocation use the exact amount or required space
- add unit tests to trigger the worst situation
- run the unit tests with assertions and asan

But if you don't do that then you have no idea anyway.

daschuer: Yes that is the sore point here and the reason why I have suggested to keep the high buffer in places where we have "no idea".
Than we can barrow them one by one with the method you suggested. 

Keeping this PR open as it is will likely lead to one more stalled PR in our list.

abique: The real question is will someone continue this work of tidying the audio buffers?
If yes then you can merge and continue on this, there is time until 2.5.
Otherwise we can simply close and not merge.

daschuer: >The real question is will someone continue this work of tidying the audio buffers?

We can't commit on that, because we are all volunteers. 

Closing this would be a waste of resource (developer and review), because this PR contains some very welcome changes. 

In Mixxx we normally do not commit work in progress changes. The main branch should still be usable without extra risk. 
So I like to merge all parts that do not involve a risk.

abique: I'll do one last change later so it'll be in a state you're more comfortable with.

src/util/defs.h
daschuer: I still don't understand. How about this: 
```suggestion
// Mixxx engine uses audio buffers up to ~80 ms. This is 8192 Frames @ 96 kHz
// see class SoundManagerConfig::AudioBufferSizeIndex
// 96 kHz * 80 ms = 7680 -> 8192 (2^13)
static constexpr unsigned int kMaxEngineFrames = 8192;
```

Than we need to have a close look into the scalers that have t o deal with 192kHz
Maybe in separate PR? 
Keep the original here:  
`constexpr unsigned int MAX_BUFFER_LEN = 160000;`
where we are unsure?

src/util/defs.h
daschuer: Here is a good summary of our discussion here:  
https://releases.llvm.org/16.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/misc/use-anonymous-namespace.html

It turns out that `static constexpr` is redundant. Please remove the static keyword.

src/util/defs.h
abique: I'm missing something.
It seems that your suggestions is for 96kHz, or it seems to me that the apps can run at up to 192kHz. So it'd double up the max number of frames per process call right?

abique:
Something you could consider, is to always run the engine at 44.1kHz or 48kHz, and then down/up sample before reading/writing to the audio interface.

abique:
I've made `MAX_BUFFER_LEN` back to were it initially was and tried to use `kMaxEngineSamples` or `kMaxEngineFrames` when applicable.

src/util/defs.h
abique: Done.

abique:
BTW, I've run the tests with ASAN.
I've also run the application with ASAN, played a few tracks, used some effects, pitch up/down, stretch... I didn't test everything but I've done basic testing and it was working.

abique:
Hi,
Did anyone had a chance to review the changes?
Are you happy now?

daschuer:
Yes, I had a brief review and it looks good now. I just want to do some test. Hope I find time soon.

daschuer (approved):
All is working good. Thank you very much.

PR #11991, opened 2023-09-16
Add `/.cache` to .gitignore

abique:
`.cache` is populated by tools like clangd.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,9 @@ compile_commands.json
 /build*
 /install
 
+# clangd's index, and other analysis tools
+/.cache
+
 # Exclude unneeded files from GoogleTest library when updating our bundled
 # GoogleTest library in the lib/ directory
 lib/googletest/docs


Holzhaus:
This is not really related to Mixxx and therefore belongs into your global `.gitignore` file, not the project-specific one. See https://david-brown.dev/posts/global-gitignore/ for details.

abique:
Ah thanks, I didn't know about git's global ignore!

Swiftb0y:
> Thanks, but this directory is not really related to Mixxx and not created by the normal build toolchain. Therefore, it belongs into your global .gitignore file, not the project-specific one. See https://david-brown.dev/posts/global-gitignore/ for details.

We should add this disclaimer to our gitignore file for people to notice before going through the effort of opening a PR.

Holzhaus:
> > Thanks, but this directory is not really related to Mixxx and not created by the normal build toolchain. Therefore, it belongs into your global .gitignore file, not the project-specific one. See https://david-brown.dev/posts/global-gitignore/ for details.
> 
> We should add this disclaimer to our gitignore file for people to notice before going through the effort of opening a PR.

We already have that: https://github.com/mixxxdj/mixxx/blob/main/.gitignore#L1-L4

Swiftb0y:
Oh yeah, well we might also want to add a link to a longer explanatory post then with hint to the global gitignore.

PR #11998, opened 2023-09-17
CO Renaming (Pt. 3): Add `ControlPotMeter` aliasing and move `audio_latency_*` COs

Holzhaus:
*This PR is part of the ongoing effort to remove offensive language from the Mixxx codebase (#11931).*

I had to add proper support for adding aliases to `ControlPotMeter` due to `audio_latency_usage` and `audio_latency_overload`. However, I'm wondering why these controls are `ControlPotMeter` objects in the first place. These should be read-only IIRC, so there is no need for all these setter-COs.

| Old Name                                                | New Name                                              | Alias?
| ------------------------------------------------------- | ----------------------------------------------------- | ----------
| `[Master],audio_latency_usage`                          | `[App],audio_latency_usage`                           | :heavy_check_mark:
| `[Master],audio_latency_overload`                       | `[App],audio_latency_overload`                        | :heavy_check_mark:
| `[Master],audio_latency_overload_count`                 | `[App],audio_latency_overload_count`                  | :heavy_check_mark:

### DIFF
--- a/src/control/controlpotmeter.cpp
+++ b/src/control/controlpotmeter.cpp
@@ -5,6 +5,14 @@
 #include "control/controlpushbutton.h"
 #include "moc_controlpotmeter.cpp"
 
+namespace {
+
+ConfigKey configKeyFromBaseKey(const ConfigKey& key, const QString& suffix) {
+    return ConfigKey(key.group, QString(key.item) + suffix);
+}
+
+} // namespace
+
 ControlPotmeter::ControlPotmeter(const ConfigKey& key,
         double dMinValue,
         double dMaxValue,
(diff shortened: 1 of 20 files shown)


src/control/controlpotmeter.cpp
```diff
@@ -146,76 +117,95 @@ PotmeterControls::PotmeterControls(const ConfigKey& key)
...
+    m_controlSetDefault.addAlias(ConfigKey(
+            key.group, QString(key.item) + QStringLiteral("_set_default")));
+    m_controlSetZero.addAlias(ConfigKey(
+            key.group, QString(key.item) + QStringLiteral("_set_zero")));
+    m_controlSetOne.addAlias(ConfigKey(key.group, QString(key.item) + QStringLiteral("_set_one")));
+    m_controlSetMinusOne.addAlias(ConfigKey(
+            key.group, QString(key.item) + QStringLiteral("_set_minus_one")));
+    m_controlToggle.addAlias(ConfigKey(key.group, QString(key.item) + QStringLiteral("_toggle")));
+    m_controlMinusToggle.addAlias(ConfigKey(
+            key.group, QString(key.item) + QStringLiteral("_minus_toggle")));
```
Swiftb0y, line 136: this is duplicated across the constructor, I'm worried that we might forget to update both places in case we add another suffix, would this be possible to deduplicate?

src/engine/enginemixer.cpp
```diff
@@ -81,9 +82,24 @@ EngineMixer::EngineMixer(
...
+            ConfigKey(
+                    kAppGroup, QStringLiteral("audio_latency_overload_count")),
+            true,
+            true);
+    m_pAudioLatencyOverloadCount->addAlias(ConfigKey(
+            kLegacyGroup, QStringLiteral("audio_latency_overload_count")));
+    m_pAudioLatencyUsage = new ControlObject(
+            ConfigKey(kAppGroup, QStringLiteral("audio_latency_usage")),
+            0.0,
+            0.25);
```
Swiftb0y, line 95:
```suggestion
            ConfigKey(kAppGroup, QStringLiteral("audio_latency_usage")));
```

ronso0:
> I couldn't find any use of the additional `ControlPotmeter` COs, and I think they don't make sense. I converted `audio_latency_usage` and `audio_latency_overload` into regular COs.

Great! So we don't need Potmeter aliasing at all? If yes, why keep that commit?

Holzhaus:
> > I couldn't find any use of the additional `ControlPotmeter` COs, and I think they don't make sense. I converted `audio_latency_usage` and `audio_latency_overload` into regular COs.
> 
> Great! So we don't need Potmeter aliasing at all? If yes, why keep that commit?

There are other COs that might need it. We didn't rename all of them.

And I thinks it's good if `addAlias` "just works". Without that commit, if we later add an alias to an ControlPotmeter, it would work but just alias the main control, not the other ones and we might not notice it.

src/control/controlpotmeter.cpp
Holzhaus: I don't really know how to do that in an elegant way. I pushed a little helper function to make the ConfigKey creation less verbose though.

Swiftb0y: Well, I came up with this: https://compiler-explorer.com/z/xxE83Gf8r
I find it very elegant, just don't know how easy it is to understand.
Frankly, now that I've written it, I don't think we should use it here. Its very seldom that we would add another suffix for `ControlPotMeter`.

Swiftb0y: Next best thing would just be a comment in both lists.

Holzhaus: https://github.com/mixxxdj/mixxx/pull/11998/commits/a3ca5eae6c50384650c5f6248f74e10efc6f9958

src/engine/enginemixer.cpp
```diff
@@ -81,9 +82,19 @@ EngineMixer::EngineMixer(
...
+            ConfigKey(kAppGroup, QStringLiteral("audio_latency_overload_count")));
+    m_pAudioLatencyOverloadCount->addAlias(ConfigKey(
+            kLegacyGroup, QStringLiteral("audio_latency_overload_count")));
+    m_pAudioLatencyUsage = new ControlObject(
+            ConfigKey(kAppGroup, QStringLiteral("audio_latency_usage")));
+    m_pAudioLatencyUsage->addAlias(ConfigKey(kLegacyGroup, QStringLiteral("audio_latency_usage")));
+    m_pAudioLatencyOverload = new ControlObject(
+            ConfigKey(kAppGroup, QStringLiteral("audio_latency_overload")),
+            0.0,
+            1.0);
```
Swiftb0y, line 95: these must go too, right?
```suggestion
            ConfigKey(kAppGroup, QStringLiteral("audio_latency_overload")));
```

Swiftb0y (commented):
lgtm otherwise, thank you

Swiftb0y (approved):
LGTM, thank you.

Holzhaus:
Merge?

Holzhaus:
Thanks.

PR #12002, opened 2023-09-19
feat: allow sanitizers on GCC as well

Swiftb0y:
`-fsanitize=address/undefined/thread` are all available on GCC too so I see no reason for the restriction to clang.


Holzhaus:
Thanks.

Swiftb0y:
Thank you.

CMakeLists.txt
```diff
@@ -2773,26 +2773,26 @@ if(CLANG_COLORDIAG)
...
+if(SANITIZE_UNDEFINED)
+  list(APPEND SANITIZERS "undefined")
+endif()
+option(SANITIZE_THREAD "Clang Thread Sanitizer" OFF)
+if(SANITIZE_THREAD)
+  list(APPEND SANITIZERS "thread")
+endif()
+if(NOT SANITIZERS STREQUAL "")
+  if(NOT (LLVM_CLANG OR GNU_GCC))
+    message(FATAL_ERROR "Sanitizers are only available on Clang or GCC")
```
JoergAtGithub, line 2791: ASAN is also available in MSVC: https://learn.microsoft.com/en-us/cpp/sanitizers/asan

Swiftb0y: yes, but the commandline syntax is a little different and I didn't want to go KISS (also I would have had no way to test the MSVC code path). Feel free to implement yourself in a follow-up PR though.

PR #12008, opened 2023-09-21
Enable more clazy checks by default

Holzhaus:


src/test/replaygaintest.cpp
```diff
@@ -5,6 +5,10 @@
 #include "test/mockedenginebackendtest.h"
 #include "track/replaygain.h"
 
+// The gtest `ASSERT_*` macros return a void expression and thus may lead to a
+// lot of clazy warnings.
+// clazy:excludeall=returning-void-expression,missing-qobject-macro
```
Swiftb0y, line 10: why `missing-qobject-macro`?

src/mixxxapplication.cpp
```diff
@@ -154,7 +155,8 @@ bool MixxxApplication::notify(QObject* target, QEvent* event) {
         break;
     }
     case QEvent::MouseButtonRelease: {
-        QMouseEventEditable* mouseEvent = static_cast<QMouseEventEditable*>(event);
+        QMouseEventEditable* mouseEvent =
+                static_cast<QMouseEventEditable*>(event); // clazy:exclude=wrong-qevent-cast
```
Swiftb0y, line 159: I just looked at this recently, this actually triggers UBSAN, so we should find an alternative to `QMouseEventEditable` and this cast.

.github/workflows/build-checks.yml
```diff
@@ -117,7 +117,7 @@ jobs:
         run: cmake --build . -j $(nproc) -- --keep-going
         working-directory: build
         env:
-          CLAZY_CHECKS: level2,no-rule-of-two-soft,no-non-pod-global-static,no-qproperty-without-notify,no-wrong-qevent-cast,no-qstring-allocations,no-function-args-by-value,no-copyable-polymorphic,no-ctor-missing-parent-argument,no-missing-qobject-macro,no-rule-of-three,no-returning-void-expression,no-missing-typeinfo,no-base-class-event
+          CLAZY_CHECKS: level2,no-rule-of-two-soft,no-non-pod-global-static,no-qproperty-without-notify,no-qstring-allocations,no-function-args-by-value,no-copyable-polymorphic,no-ctor-missing-parent-argument,no-rule-of-three
```
Swiftb0y, line 120: I don't think we should use `no-returning-void-expression`. It results in too much noise in the tests and the bug it protects against (at least in their example) is very unlikely to occur in mixxx due to our code formatting rules.

Swiftb0y (changes requested):
Thank you, but this needs a little more work and discussion. IMO all `clazy:exclude`s need an explanation why its valid.

src/test/replaygaintest.cpp
Holzhaus: We have some test classes that inherit from it. But adding Q_OBJECT won't help, because gtest generates new classes from it via macro magic, and we can't add Q_OBJECT to these anyway.

src/mixxxapplication.cpp
Holzhaus: I agree. Fortunately this file seems to be the only occurrence so we should enable that clazy rule to avoid adding it elsewhere.

daschuer (changes requested):

src/mixxxapplication.cpp
daschuer: Qt seems to have a similar issue here: https://bugreports.qt.io/browse/QTBUG-99563

daschuer: Normally a downcast to a inherited object is undefined, because the object might be bigger and we might access memory out of the boundary of the original object. I think the only way around the undefined behavior is to make a copy, modify the value and than copy the result back. That is a CPU cost to fix an non-issue   ... :-/

Swiftb0y: I don't think these are particularly expensive... I'd take the hit for avoiding UB. It would be very annoying to use ubsan with "false positives" like this.

.github/workflows/build-checks.yml
Swiftb0y: I didn't spot and true positives for `no-returning-void-expression` in the diff. Did you?

src/engine/filters/enginefilteriir.h
```diff
@@ -25,13 +25,11 @@ enum IIRPass {
     IIR_HP2,
 };
 
-
-class EngineFilterIIRBase : public EngineObjectConstIn {
+class EngineFilterIIRBase : public EngineObjectConstIn { // clazy:exclude=missing-qobject-macro
```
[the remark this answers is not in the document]
Holzhaus: I'm questioning the purpose of this class. We could just merge it into `EngineFilterIIR`:

```
$ git grep EngineFilterIIRBase
src/analyzer/analyzerwaveform.h:class EngineFilterIIRBase;
src/analyzer/analyzerwaveform.h:    EngineFilterIIRBase* m_filter[FilterCount];
src/engine/filters/enginefilteriir.h:class EngineFilterIIRBase : public EngineObjectConstIn {
src/engine/filters/enginefilteriir.h:class EngineFilterIIR : public EngineFilterIIRBase {
```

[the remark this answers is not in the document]
Holzhaus: Ah, it's because EngineFilterIIR is a template class.

src/test/replaygaintest.cpp
Swiftb0y: makes sense, this should also go in the comment imo.

src/engine/filters/enginefilteriir.h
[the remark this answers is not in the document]
Holzhaus: Anyway, we can't use it here because there is no CPP file for it and leads to linker errors:

```
mold: error: undefined symbol: EngineFilterIIRBase::qt_metacast(char const*)
>>> referenced by enginefilterlinkwitzriley4.cpp
>>>               libmixxx-lib.a(enginefilterlinkwitzriley4.cpp.o):(EngineFilterLinkwitzRiley4Low::qt_metacast(char const*))>>> referenced by enginefilterbessel4.cpp
>>>               libmixxx-lib.a(enginefilterbessel4.cpp.o):(EngineFilterBessel4Low::qt_metacast(char const*))>>> referenced by enginefilterbessel4.cpp
>>>               libmixxx-lib.a(enginefilterbessel4.cpp.o):(EngineFilterBessel4Band::qt_metacast(char const*))>>> referenced 13 more times

mold: error: undefined symbol: typeinfo for EngineFilterIIRBase
>>> referenced by enginefilterbessel4.cpp
>>>               libmixxx-lib.a(enginefilterbessel4.cpp.o)>>> referenced by enginefilterbessel4.cpp
>>>               libmixxx-lib.a(enginefilterbessel4.cpp.o)>>> referenced by enginefilterbessel4.cpp
>>>               libmixxx-lib.a(enginefilterbessel4.cpp.o)>>> referenced 7 more times

mold: error: undefined symbol: EngineFilterIIRBase::staticMetaObject
>>> referenced by enginefilterlinkwitzriley4.cpp
>>>               libmixxx-lib.a(enginefilterlinkwitzriley4.cpp.o)>>> referenced by enginefilterlinkwitzriley4.cpp
>>>               libmixxx-lib.a(enginefilterlinkwitzriley4.cpp.o)>>> referenced by enginefilterbessel4.cpp
>>>               libmixxx-lib.a(enginefilterbessel4.cpp.o)>>> referenced 13 more times

mold: error: undefined symbol: EngineFilterIIRBase::qt_metacall(QMetaObject::Call, int, void**)
>>> referenced by enginefilterbessel4.cpp
>>>               libmixxx-lib.a(enginefilterbessel4.cpp.o):(EngineFilterBessel4Low::qt_metacall(QMetaObject::Call, int, void**))>>> referenced by enginefilterlinkwitzriley4.cpp
>>>               libmixxx-lib.a(enginefilterlinkwitzriley4.cpp.o):(EngineFilterLinkwitzRiley4Low::qt_metacall(QMetaObject::Call, int, void**))>>> referenced by enginefilterbessel4.cpp
>>>               libmixxx-lib.a(enginefilterbessel4.cpp.o):(EngineFilterBessel4Band::qt_metacall(QMetaObject::Call, int, void**))>>> referenced 13 more times

clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
```

src/library/multilineeditdelegate.cpp
```diff
@@ -67,7 +67,7 @@ bool MultiLineEditor::eventFilter(QObject* obj, QEvent* event) {
         if ((ke->key() == Qt::Key_Return || ke->key() == Qt::Key_Enter) &&
                 ke->modifiers().testFlag(Qt::NoModifier)) {
             emit editingFinished();
-            return false;
+            return false; // clazy:exclude=base-class-event
```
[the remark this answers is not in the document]
Holzhaus: I wasn't sure, it emits `editingFinished()` before. I can remove the `return false` but I was unsure if there are any sideeffects.

src/engine/filters/enginefilteriir.h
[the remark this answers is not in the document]
Swiftb0y: good point, QObjects can't be templated...

.github/workflows/build-checks.yml
daschuer: https://github.com/mixxxdj/mixxx/blob/main/src/engine/controls/enginecontrol.cpp#L74

Swiftb0y: unnecessary, sure, but not a bug.

src/library/multilineeditdelegate.cpp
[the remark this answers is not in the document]
daschuer: Just checked the Qt source and it should not make a difference. The base class returns false as well. 
So I would prefer to remove return false.

Holzhaus:
I removed the noisy `returning-void-expression` and `missing-qobject-cast` checks for now. This can happen later. But I found some useful checks from the "manual" level that make sense to enable.

src/library/multilineeditdelegate.cpp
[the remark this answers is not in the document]
Holzhaus: https://github.com/mixxxdj/mixxx/pull/12008/commits/93ccf8279586b49df807d57b1f7135cd634123ac

Swiftb0y:
[`container-anti-pattern`](https://github.com/KDE/clazy/blob/master/docs/checks/README-container-anti-pattern.md#container-anti-pattern) is also worth to consider in a follow up PR.

src/mixxxapplication.cpp
Holzhaus: I opened an issue for this: https://github.com/mixxxdj/mixxx/issues/12013

Swiftb0y (approved):
lgtm thank you. waiting for @daschuer

daschuer (changes requested):
Just one remaining comment. 
Thank you.

src/effects/backends/effectsbackend.cpp
```diff
@@ -29,11 +29,15 @@ QString EffectsBackend::backendTypeToString(EffectBackendType backendType) {
 QString EffectsBackend::translatedBackendName(EffectBackendType backendType) {
     switch (backendType) {
     case EffectBackendType::BuiltIn:
-        //: Used for effects that are built into Mixxx
```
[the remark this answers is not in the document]
Holzhaus: Done.

daschuer (approved):
LGTM, Thank you.

Holzhaus:
Thanks!

PR #12009, opened 2023-09-22
CoverArtCache refactoring + Fix scrolling lag after updating Mixxx 

### LINKED ISSUE #11131: Deck/Spinny Coverart is randomly not read
### Bug Description

If I load a track into a deck, the CoveArt is sometimes displayed, and sometimes not. This doesn't depend on the track, it can be reproduced for any track after some retries.
This doesn't happen with 2.3, only with Main.

In the --developer output I see in cases without loaded coverart:
```
info [Thread (pooled)] MetadataSourceTagLib - No track metadata or cover art found in file "C:/Users/Joerg.WORLDWARTWEB/Music/Der Dritte Raum - Hale Bopp (original mix).mp3" with type 3
debug [Thread (pooled)] SoundSourceProxy - SoundSourceProvider "MAD: MPEG Audio Decoder" created a SoundSource for file "file:///C:/Users/Joerg.WORLDWARTWEB/Music/Der Dritte Raum - Hale Bopp (original mix).mp3" of type "mp3"
debug [AnalyzerThread 0 #1] AnalyzerThread - Dequeued next track 140
debug [Thread (pooled)] SoundSourceProxy - SoundSourceProvider "MAD: MPEG Audio Decoder" created a SoundSource for file "file:///C:/Users/Joerg.WORLDWARTWEB/Music/Der Dritte Raum - Hale Bopp (original mix).mp3" of type "mp3"
warning [Thread (pooled)] TagLib - Cannot read audio properties from inaccessible/unreadable/invalid file:
debug [AnalyzerThread 0 #1] AnalyzerThread - Analyzing QFileInfo(C:\Users\Joerg.WORLDWARTWEB\Music\Der Dritte Raum - Hale Bopp (original mix).mp3)
info [Thread (pooled)] MetadataSourceTagLib - No track metadata or cover art found in file "C:/Users/Joerg.WORLDWARTWEB/Music/Der Dritte Raum - Hale Bopp (original mix).mp3" with type 3
warning [Thread (pooled)] TagLib - Cannot read audio properties from inaccessible/unreadable/invalid file:
info [Thread (pooled)] MetadataSourceTagLib - No track metadata or cover art found in file "C:/Users/Joerg.WORLDWARTWEB/Music/Der Dritte Raum - Hale Bopp (original mix).mp3" with type 3
```

### Version

2.4

### OS

Windows7 & Windows11

daschuer:
This is a prerequisite to fix #11131 

It consolidates a grown interface with a hard to use `enum class Loading` that was passed in most cases redundant to the different inputs to a function. To use it right you needed to pass the correct loading value with the correct set of parameters. 

This has been now replaces with single purpose functions which should be self-explainable. 
``` 
static void requestCover() 
static void requestTrackCover()
static QPixmap getCachedCover();
static  void requestUncachedCover();
```
The original universal interface has been mad private. 

To make this work, a refactoring of the migration from legacy hash was required. 
This is now always done in the worker thread, to get around the hard to understand extra information of `requestedCacheKey` and `coverInfoUpdated` in the `CoverFound()` signal. This should improve the library scrolling after updating from a legacy Mixxx version. 

I have tried hard to do this PR in single testable commits, to allow a good review. 

I have two follow up in my stash: 
* Fix the cover art tooltip if not in cache
* Fix the #11131


src/track/track.cpp
```diff
@@ -852,7 +852,7 @@ void Track::setBitrate(int iBitrate) {
 }
 
 TrackId Track::getId() const {
-    const auto locked = lockMutex(&m_qMutex);
+    static_assert(sizeof(TrackId) <= sizeof(void*)); // no locking required
     return m_record.getId();
```
Swiftb0y, line 856: why? are you sure the same assumption is true for ARM?

daschuer: Yes. There is theoretical the issue for a possible odd aliment we may assert as well, but you need to try hard to trick the compiler to achieve that.

daschuer:
Is s class always integer alligned?

src/track/track.cpp
daschuer: In this case, a `alignas(int) int m_value;` is the best solution without loosing pipelining because of the memory barriers in an a `std::atomc_int`. I have also remove the generic value_type which gives us the guarantee that we have always int and allows fully in-lining of the DbId class.

JoergAtGithub:
This doesn't compile here:
`src\util\db\dbid.h(27): error C3615: constexpr function 'DbId::DbId' cannot result in a constant expression`

JoergAtGithub:
Now I can build locally. Everything works as before, except the Coverart-Tooltip in the library, which does no longer appear. I guess this is intended for this intermediate step.

daschuer:
It should appear after the second try. The first try fills the cache. This should however the same in 2.4.
Can you confirm?

JoergAtGithub:
No, this was the behavior before. With this PR a tooltip for the CoverArt does not appear at all.

src/track/track.cpp
Swiftb0y: I'm still not comfortable with bugprone optimizations like this without evidence that they are needed. I also doubt that a properly used `std::atomic` would be any slower than this.

daschuer: I was originally started to implement a std::atomic based solution here. With a 'memory_order_relaxed' you can also achieve that the processor pipeline is not cleared. But than I have noticed that the load a and store calls are (probably intentional) not inline, they are real calls. std::atomic are used for atomic operations like read and increments and memory briers, all this not needed here. We would need a lot of boilerplate code to achieve something with std::atomic that is reliable out of the box. 

Here we really just want to read a single integer as a whole that is in addition almost constant. This is already a thing that cannot be interrupted = atomic. The compiler only uses unaligned integers when it is forced to. See also https://stackoverflow.com/a/70967180
 
I have found a one header implementation of an atomic int that also uses aligneas(int): https://github.com/wingfiring/xirang/blob/eda84647cca476a29e8d785b8fd8ceda78bb952f/trunk/xirang/backward/atomic.h 
so we are in a good companion. 

But anyway now with alingeas(int) we have sorted out this case as well. So why do you still think this is bugprone. What is the scenario that may fail?

daschuer:
Oh, there was missing ! introduced in the last minute. Now everything should be like before, just faster.

src/track/track.cpp
Swiftb0y: > I was originally started to implement a std::atomic based solution here. With a 'memory_order_relaxed' you can also achieve that the processor pipeline is not cleared. But than I have noticed that the load a and store calls are (probably intentional) not inline, they are real calls.

Are you sure, they're fully inlined on `-O1` on all compilers. https://compiler-explorer.com/z/n3MbKqr3q
There is no reason why they couldn't be inlined unless atomics where using dynamic dispatch (which they absolutely aren't, considering how performance-sensitive they are). 

> We would need a lot of boilerplate code to achieve something with std::atomic that is reliable out of the box.

Why would that be? Just make the member atomic and then `.load()` and `.store()` to it. 

> This is already a thing that cannot be interrupted = atomic.

I'm worried the assumption might break, silently resulting in a race-condition. I'd rather have the fallback to slower behavior, than undefined behavior. 

> I have found a one header implementation of an atomic int that also uses aligneas(int): https://github.com/wingfiring/xirang/blob/eda84647cca476a29e8d785b8fd8ceda78bb952f/trunk/xirang/backward/atomic.h
so we are in a good companion.

That project has been dead for 6 years, I doubt they did any work to validate whether that actually works on every platform they support, which are probably even fewer platform than we support. 

> But anyway now with alingeas(int) we have sorted out this case as well. So why do you still think this is bugprone. What is the scenario that may fail?

Because on paper `alignas` just guarantees alignment not atomicity. The conclusion you've drawn that aligned access means atomic access is implicit but not strong enough to not silently break. We either need to be able to assert this (which we can't afaik), or have it documented rigidly for all platforms we support (and might support, I don't want mixxx be unusable on RISC-V because we relied on a bunch of ARM and x86 specific behavior).

Swiftb0y: If you want to convince me that `alignas` guarantees atomicity, just link me the section in the standard (or even cppreference would be fine) that supports your assumption. But I don't think that assumption is true in all cases and thus its also not defined in the standard. As such, we can't use it blindly.
If its only defined for some CPU architectures, fine, that works too, but then we need to assert that we only run on those CPU architectures, etc. Do you see the pattern? just linking to some code and stackoverflow articles describing what might happen on some systems doesn't let us draw conclusions over all systems!

daschuer: Ok I have committed the std::atomic_int version.

uklotzde:
I need to leave a comment here.

`DbId` is supposed to be a simple, type-safe _newtype_ wrapper around a native type. It is and should not be aware of multiple theads. It is supposed to be used in single-threaded contexts only. A simple POD.

Now it has become an opinionated `AtomicDbId` that fits a single use case. But everyone else using this generic type has to pay the price.

src/util/db/dbid.h
```diff
@@ -22,35 +22,36 @@
 // not add any additional state (= member variables). Inheritance is
 // only needed for type-safety.
 class DbId {
-protected:
-public:
-    // Alias for the corresponding native type. It keeps the
```
uklotzde, line 27: The comment explained this intended purpose. But is has been and deleted.

JoergAtGithub:
I tested this, everything works now as before, but the CoverArt loading while scrolling through the library feels to be faster now.

src/track/track.cpp
Swiftb0y: Thank you. I will try to take a look, soon-ish.

Swiftb0y (changes requested):
blocking this PR for now as I don't agree with the changes. I'll elaborate asap.

daschuer:
I wanted to use the TrackId to check if it is a temporary track or not. However at that point the temporary track used for lookup the metadata is out of scope. So it means the original demand of optimize the locking is not there. 

This means we can revert the introduction of an atomic here.

daschuer:
OK the mutex is back and getID() is no longer called from the coverArt thread.

Swiftb0y (dismissed):
> OK the mutex is back and getID() is no longer called from the coverArt thread.

To be clear, I was fine with the atomic, my main issue was just the contrived conclusion that accessing an `int` was safe if it was aligned a certain way (I'm still interested in how exactly you arrived at that that conclusion, btw). Now my only issue is the unnecessary changes to `DbId`.

daschuer:
> I'm still interested in how exactly you arrived at that that conclusion, btw. 

All CPU architectures we use have a data buss width of 32 or 64 bit and have a single instruction to access this memory. Since or value is on a single address, it is electrical impossible to load the value in two steps or interrupt the access half way. However C++ also supports controllers with a smaller data bus than int. This is probably the reason the language model does not guarantee atomic int read and writes. 

On example is the Cx51. It has a 8 bit architecture and 16 bit int and requires therefore two memory accesses. 

The RISC-V (or MIPS) controller you mentions above is a RISC architecture with only 32 bit instructions only. Due to the 32/64 bit data bus it has also a single instruction for load and write to memory. The special issue here is that it is not able to load a 32 bit values > 0x0000FFFF form the text segment with a single instruction because instruction is only 32 bit. It has two instructions for this that carry 16 bit of data: LUI/ORI. They are used for constexpr values and addresses of values in RAM. So it may happen that a half way written int goes to the stack memory due to a task switch. This is not an issue here, because the stack is private to each task. All memory access in the data segment are still guaranteed atomic. 

During our discussion I was looking for a compiler flag or build in function that I can query to assert for that. But I have not found it. so I gave up.

src/util/db/dbid.h
```diff
@@ -107,21 +101,30 @@ class DbId {
...
+        int value = variant.toInt(&ok);
+        if (ok && isValidValue(value)) {
+            return value;
+        }
+        qCritical() << "Invalid database identifier value:"
+                    << variant;
+        return kInvalidValue;
+  }
+
+  int m_value;
```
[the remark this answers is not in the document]
daschuer: There was dynamic type code used even though the value is not dynamic and will never be. Even if we change that one day, the value is still a fixed type. So in the unlikely even we will ever change the data type we just need too re-implement the QVariant roundtrip. The benefit is, that this save as us from extra checks and fully inlining now. 

By the way, it also helps reading the code because you immediately see that value() returns an int.

daschuer:
Done.

JoergAtGithub:
> Due to the 32/64 bit data bus it has also a single instruction for load and write to memory. 

I don't think it's that simple anymore. Each CPU core has it's own L1 cache which runs with full CPU clock, while the main memory is external and runs with with lower clock rate.

daschuer:
> I don't think it's that simple anymore. Each CPU core has it's own L1 cache which runs with full CPU clock, while the main memory is external and runs with with lower clock rate.

The same rules apply also for the caching buffers. They have all at least 32 bit data bus that grantees that we never have partially written aligned integers. 

The cache is an issue when you consider synchronising. The required memory fences, part of the std::atomic take care that the cache is flushed (along with the penalty).

daschuer:
> That's precisely the reason why insisted on using atomics instead of the loose assumptions because unsynchronized access on the DbId is possible and could lead to data races. 

We do not use the DbId for synchronisation at all! 

You need to distinguish between two scenarios. The first is using atomics for synchronizing access to other no atomic data. Here the synchronization is relevant. 

The atomic value has to reach the ram before a task switch. If it is kept in cache an other CPU core will read the old value, which may cause a data race in other objects > 32 bit. An other issue is the pipeline with recalculated results that may base on an old cache value. All this has to be flushed for a proper synchronization, which involves a performance penalty.

In our case here we need to read only one 32 bit value from memory. The only thing that can happen in the concurrent case is that we read the old value from cache. This is still a valid value. Perfectly OK, because we would have read the same value if the task has been scheduled a spit of a second earlier. The 32 bit data bus grantees that we never read a corrupt value. 

> This also affects the entire cacheline since several members of the same class share the same cacheline so even if some code is accessing different members, of the class, the data races could still occur.

Please explain, how can cause a data race. 

Conclusion: There is nothing "bugprone" by this optimization, it is just that we have no grantee from the C++ programming model. The used architectures (all with >=32 bit databus) gives the guarantee. We had the same discussion around the Two's complement.

daschuer:
We are talking here just about a the replacement of: 
```
int a = m_atomic_int.load(std::memory_order_relaxed); 
int a = m_aligned_int; 
```
This has the same guards against race conditions, but the plain version is faster, because it does not require a function call and it is more handy because `m_plain_int` does keeps the containing object trivial copyable. The compiler cannot do anything else then assert the alignment and apply memory fences with std::memory_order_relaxed request not to apply barriers.  

I have found this in the GCC source itself: 
https://github.com/gcc-mirror/gcc/blob/master/libgcc/config/pa/linux-atomic.c#L120
I have not found the point where the memory fences are applied yet.
But it is clear that atomic load is implemented by 
__atomic_thread_fence() and atomic_load_n(), nothing else. 

> Yes, concurrent access to an object needs to be synchronized! There is no way around that! 

That is not correct. Using std::memory_order_relaxed is fine when you don't use the atomic for thread synchronisation. 

I agree to the rest. We either need to use the C++ programming model or take care that the hardware assumptions are correct.
In linux_atomi.c the later is done.

daschuer:
> I still don't believe the aligned version is safe. the atomic version has well-defined semantics for concurrent access in the C++ memory model (iiuc), aligned variables do not.

It is save for all 32 bit platforms I know of. I have tried hard to find any aditinal magic apart of alignment and memory barriers, but I didn't found anything. 

Do you have any hint that there is something else? Can you prove you assumption that extra call is optimized away? Theoreticslly both versions should than boil down to the same assembler instructions.

daschuer:
By the way: merge?

Swiftb0y:
> It is save for all 32 bit platforms I know of.

Right, but that's only safe for the underlying hardware, not the abstract C++ machine we're actually programming against.

> Do you have any hint that there is something else?

Can you clarify?

> Can you prove you assumption that extra call is optimized away? 

I don't know what qualifies as sufficient proof, but here's a godbolt with gcc, clang and msvc on x86 and arm64 respectively, all at `-O1` producing virtually the same instructions on their respective target hardware. Does that suffice? https://compiler-explorer.com/z/Ghdn7znq4

> By the way: merge?

if you remove the `value_t` -> `int` refactoring of DbId, I'm probably happy.

daschuer:
> if you remove the value_t -> int refactoring of DbId, I'm probably happy.

This value_t  typedef is only clutter. I am not happy to remove it. The whole class is already the semantic type, the is no reason to hid the real type even more.

Swiftb0y:
The "clutter" is minimal and allows for very easy change down the road. If the type was fully opaque, I wouldn't complain, but there is `value()` and a converting constructor. Since you're not a proponent of "almost-always-auto", there is value in the calling code being able to use `DbId::value_t` instead of just `int` which looses all context. I wouldn't care much if we used AAA, but we don't.

daschuer:
There is not a single point where `DbId::value_type` has been replaced by in in the calling code. So your demand is virtual. I like to go with the easy int based QVariant api instead.

Swiftb0y:
> There is not a single point where DbId::value_type has been replaced by in in the calling code. So your demand is virtual.

True. purely theoretical indeed, but I'd argue thats because the calling code shouldn't have assumed `int` (or is already using `auto` I guess). 

> I like to go with the easy int based QVariant api instead. I also don't see a reason for a typedef at all.

QVariant seems quite expensive to me. So it shouldn't be the primary API, but I haven't really looked too closely at the code to judge whether my concern is valid. 

> I also don't see a reason for a typedef at all.

daschuer:
It is nice to see that the load() and store() calls vanish with -01. 
Here is the one that confirms my assumption:  
```
int a = m_atomic_int.load(std::memory_order_relaxed); 
int a = m_aligned_int;
``` 
https://compiler-explorer.com/z/rbsK4z9E7

daschuer:
> It's the type equivalent of using a local variable instead of a magic constant.

I am in the opinion that this is exactly the opposite. A magic number is just a number without any info. With using it via a constant you have a name with info. Hiding a well known type behind an anonymous type like  DbId::value_type does the opposite. We already have DbId for the semantic. In case of DbId::value() the returned type int is relevant when hovering the code.

Swiftb0y:
> https://compiler-explorer.com/z/rbsK4z9E7

Yes, seems to apply for the 32-bit platforms. still using `int` instead of the atomic int is likely UB. Also, the assumption breaks for most platforms that aren't arm of x86. 

> I am in the opinion that this is exactly the opposite. A magic number is just a number without any info. With using it via a constant you have a name with info. Hiding a well known type behind an anonymous type like DbId::value_type does the opposite. We already have DbId for the semantic. In case of DbId::value() the returned type int is relevant when hovering the code.

The same applies here, using `int` carries no info or context, using `DbId::value_type` makes it recognizable that the value carries meaning and thus it needs to be handled with care (ideally we would have no way to convert from and to the underlying type at all). Besides, there is no reason to know `DbId::value_type`. Do you want to do bit-twiddling on the ID or why do you need to know that its an `int`?! You're making the abstraction leakier than it already is.
Your argument is analogous to hardcoding a constant everywhere in the code because "how else would I know what value I'm passing". Applying your argument, we should abolish all typedefs in mixxx, SINT, CSAMPLE, etc.

This is ridiculous and refuse to argue about non-issues like this any longer. This is not a useful refactoring, this is not a constructive discussion. This is a waste of time for the both of us and the one winning the argument is simply the one willing to waste more time arguing. 

I know this is ridiculous, but I will call for a vote over this so we can stop arguing.

daschuer:
It looks like with the std::atomic issue we are already on the same page. This PR is in a good condition for a while and we have discussed, implementation details, because you have asked for that. 

For the `DbId::value_type` issue, I have changed to int, to allow us to use the easy int based API which has some performance implications. 

> This is ridiculous and refuse to argue about non-issues like this any longer. 

All this discussion is about to understand the reasons from each other. Reading something like this feels bad and does not contribute the the problem. It is not about to overrule someone. It is about a common understanding.

I will have a look look into the issue and propose a change that covers the performance benefit and hiding of implementation details.

daschuer:
@Swiftb0y Ok, the int is gone. Please have a look.

src/util/db/dbid.h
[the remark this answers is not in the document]
JoergAtGithub: @Swiftb0y Do you agree with Daniels answer here, or do you request to change this?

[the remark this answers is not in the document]
Swiftb0y: I don't agree, but I don't care enough to continue fighting over it.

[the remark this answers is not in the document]
daschuer: My comment is outted. Now the int type is no longer part of the public interfacebofvthe class. This should fully resolve your concerns.

src/util/db/dbid.h
```diff
@@ -24,33 +24,19 @@
...
-    }
-    explicit DbId(QVariant variant)
-        : DbId(valueOf(std::move(variant))) {
-    }
+  constexpr DbId()
+          : m_value(kInvalidValue) {
+  }
+  explicit DbId(const QVariant& variant)
+          : m_value(valueOf(variant)) {
+  }
```
Swiftb0y, line 32: since you eliminated the `int` from the interface, how can we ensure to avoid the QVariant construction overhead in the case where we are getting passed an int?

Swiftb0y (commented):
can't comment on `coverartcache`, `coverartdelegate` and `basetracktablemodel`, but my concerns over DbIb have been mostly resolved.

src/util/db/dbid.h
daschuer: We always get a QVariant.

JoergAtGithub: What is the conclusion here?

Swiftb0y: well, I'd either like to see a constructor taking `int` (at which point the `int` is part of the interface again), or a deleted ctor taking `int` to warn us of the fact that we're passing an int directly... Anyways, I'm probably overcomplicating the issue as always. resolved.

daschuer: Deleting `DbId(int)` was a good idea. It reveals some int roundtrips. I have removed them from the Mixxx code. 
Unfortunately we have some constants that are derived from `int` the testing code. I have fixed them in the latest commit.

daschuer:
The clazy failure is already fixed in 2.4.  
Shall I rebase it?

JoergAtGithub:
@daschuer Please fix the build failures - by rebase, or whatever is needed

daschuer:
All green again.

Swiftb0y:
@JoergAtGithub if you're willing to sign off on the coverartcache, coverartdelegate and basetracktablemodel changes, feel free to merge this PR.

JoergAtGithub (changes requested):
I reviewed the code and found only some minor nitpicks. I also build it on Windows 11 and it works as expected.

src/library/coverart.h
```diff
@@ -70,20 +70,25 @@ class CoverInfoRelative {
...
         m_legacyHash = legacyHash;
     }
 
+    bool hasCacheKey() const {
+        return !m_imageDigest.isEmpty() || m_legacyHash != defaultLegacyHash();
+    }
+
     bool hasImage() const {
-        return cacheKey() != CoverImageUtils::defaultCacheKey();
+        return type != NONE;
```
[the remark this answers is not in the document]
daschuer:
```
Source source;
Type type; 
```
Are global members that are used in many places like `coverInfo.type == ...`
We may consider to rename them to private m_ types an and add getters and setters, but that is out of scope of this PR.

src/library/coverart.h
```diff
@@ -70,20 +70,25 @@ class CoverInfoRelative {
...
     }
 
-    const QByteArray imageDigest() const {
+    const QByteArray& imageDigest() const {
```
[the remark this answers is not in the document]
daschuer: CoverInfoRelative has the style that the getter functions don't use the get prefix. 
If we change one function we might want to change all, but this is a bigger refactoring out of scope of this PR.

src/library/coverart.h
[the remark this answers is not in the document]
JoergAtGithub: Ok

src/library/coverartcache.cpp
```diff
@@ -67,71 +72,97 @@ void CoverArtCache::requestTrackCover(
...
+void CoverArtCache::requestUncachedCoverPrivate(
+        const QObject* pRequester,
+        const TrackPointer& pTrack,
+        const CoverInfo& coverInfo,
+        int desiredWidth) {
+    CoverArtCache* pCache = CoverArtCache::instance();
+    VERIFY_OR_DEBUG_ASSERT(pCache) {
+        return;
+    }
+    pCache->tryLoadCover(
```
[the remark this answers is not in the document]
daschuer: Good idea. Done.

JoergAtGithub (approved):
LGTM! Waiting for CI

daschuer:
All green again

JoergAtGithub:
What did you push here?

daschuer:
I have amended the last commit, because there was a code style issue.

daschuer:
Oh no, I see. I have messed with the follow up PTRs.

daschuer:
Restored the original state from reflog.

JoergAtGithub:
Thank you!

Issue #10626, opened 2022-08-23
DEBUG ASSERT Beats::iteratorFrom 


mixxxbot:
Reported by: **[daschuer](https://github.com/daschuer)**
Date: 2021-12-28T22:42:28Z
Status: New
Importance: Critical
Launchpad Issue: [lp1955929](https://bugs.launchpad.net/bugs/1955929)

---

I have just experienced Debug Assert violation in current main branch. 

    critical Engine DEBUG ASSERT: "it == cbegin() || it == cend() || *it - position < std::prev(it).beatLengthFrames()" in function mixxx::Beats::ConstIterator mixxx::Beats::iteratorFrom(mixxx::audio::FramePos) const at /home/daniel/workspace/2.3/src/track/beats.cpp:447

This happens after pressing play with a loaded Track, called from 
Beats::findPrevNextBeats

ronso0:
Did this happen again?
Can't reproduce with setting loop_in before existing loop (#11665)

Swiftb0y:
I also tried reproducing this some time ago and I couldn't but judging from the initial posts the issue is sporadic anyways.

daschuer:
It has just happen again with the current 2.4 branch: 
```
critical [Engine] DEBUG ASSERT: "it == cbegin() || it == cend() || (*it >= position && *std::prev(it) < position)" in function mixxx::Beats::ConstIterator mixxx::Beats::iteratorFrom(mixxx::audio::FramePos) const at /home/sperry/workspace/mixxx2/src/track/beats.cpp:455
```

m0dB:
Since the failing assert
```it == cbegin() || it == cend() || (*it >= position && *std::prev(it) < position```
is preceeded by this assert
 ```it == cbegin() || it == cend() || *it >= position```
we know that it >= position, so we could also write the failing assert as
```it == cbegin() || it == cend() || *it > *std::prev(it)```

I have been looking at the code to try to understand how it could ever occur that ```*it > *std::prev(it)``` doesn't hold true, but I don't see it. We could add some more asserts in the ConstIterator += and -= operators, that might help to trace it down.

Now, I have never been able to reproduce this myself...

Also I wonder, is this really a "party stopper"? What happens if you continue despite this condition not holding true?

m0dB:
@daschuer do you have a way to reproduce this? Can you try with the additional asserts? In any case, those additional asserts make sense so I propose to merge the PR anyway.

daschuer:
[...]

PR #12016, opened 2023-09-24
added beats asserts

m0dB:
added asserts in Beats::ConstIterator operator -= and += to make sure that the resulting ConstIterator has indeed a earlier or later position respectively, with the intention of finding the cause of https://github.com/mixxxdj/mixxx/issues/10626

### DIFF
--- a/src/track/beats.cpp
+++ b/src/track/beats.cpp
@@ -61,6 +61,8 @@ Beats::ConstIterator Beats::ConstIterator::operator+=(Beats::ConstIterator::diff
 
     DEBUG_ASSERT(n > 0);
     const int beatOffset = m_beatOffset + n;
+    const auto origValue = m_value;
+    (void)origValue;
 
     // Detect integer overflow
     if (beatOffset < m_beatOffset) {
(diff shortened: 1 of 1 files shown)


src/track/beats.cpp
```diff
@@ -61,6 +61,8 @@ Beats::ConstIterator Beats::ConstIterator::operator+=(Beats::ConstIterator::diff
 
     DEBUG_ASSERT(n > 0);
     const int beatOffset = m_beatOffset + n;
+    const auto origValue = m_value;
+    (void)origValue;
```
daschuer, line 65:
```suggestion
#ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
    const auto origValue = m_value;
#endif
```

daschuer (changes requested):
just a minor improvement suggested

src/track/beats.cpp
m0dB: Changed.

daschuer (approved):

daschuer:
Thank you.

PR #3572, opened 2021-01-18
Better file export

poelzi:
Greatly enhance the track export feature.

* Added template functionality for formatting filenames
* Better handling of errors while exporting
* Remove now obsolete wizzard
* Does not crash anymore when tracklist contains a nullptr or other problems

### Done
* add custom filters for formatting key and building folders for ranges (bpm)
* fill CrateSummary in template context when exporting crate
* fill PlaylistSummary in template context when exporting playlist (after #3406 ). Also playlist posittion as variable
* Export tracks directly out of the library

### TODO
* Document all available properties

### Future
* Incorperate #3102 as an option here
* Add transcoding support into different file formats

Screenshot_20210118_024227

PR #12019, opened 2023-09-24
Better file export (rebased #3572 with Windows build fixes)

JoergAtGithub:
This PR is a version of #3572 from @poelzi rebased to latest Main with Qt6. A PR which where the review was allready approved, but that was never merged, because the neccessary dependency Grantlee wasn't available in VCPKG.

I added the Grantlle library to VCPKG in https://github.com/microsoft/vcpkg/pull/28809 therefore.

This PR contains the neccessary fixes to run on Windows now.

But unfortunatly it doesn't build on Linux anymore, because the Grantlee package in Ubuntu 22 is still build against Qt5.
There is also a build issue with macOS, but this should be just a minor CMake issue, because it compiles and links, but it doesn't find the file `libGrantlee_Templates.5.dylib` during packaging and test.

I would appreciate PRs to my branch from devs on macOS and Linux, to get these two platforms building too.


poelzi:
thanks for picking this up 👍🏽

daschuer:
It looks like https://github.com/KDE/ktexttemplate is the Qt6 successor of grantlee. Unlike grantlee, it is actively developed and has become part of Linux distros: https://repology.org/project/ktexttemplate/versions

libkf6texttemplate6 is started to be part of Ubuntu Questing 25.10.

PR #12020, opened 2023-09-24
shortkeys for track list management

m0dB:
This PR implements shortkeys for track list management (between library, playlists, crates and autodj)

Ctrl-X
- with focus on tracklist: remove selected tracks from tracklist and place them on the clipboard

Ctrl-C
- with focus on tracklist: copy selected tracks from tracklist to the clipboard

Ctrl-V 
- paste tracks from the clipboard to the current tracklist
  - with focus on tracklist: 
    - in playlists and autodj cue: inserting them at the current index position (the last selected row) or appending at the bottom when no track is selected
    - in crates: inserting them according to the current sorting
  - with focus on the sidebar
    - appending at the bottom of the track list selected in the sidebar

Shift-Up/Down
- Select multiple tracks

Delete/Backspace
- with focus on tracklist: remove selected tracks from tracklist but without placing them on the clipboard

Ctrl+A
- with focus on tracklist: select all tracks

Escape
- with focus on tracklist: clear the selection

Ctrl-Left
- move focus to the sidebar (similar to using tab, but without going through the search bar)
 
Ctrl-Right
- move focus to the track list (similar to using tab, but without going through the search bar)

Please keep the discussion here focussed on implementation, and discuss the actual UI interaction at https://mixxx.zulipchat.com/#narrow/stream/247619-UI-.26-UX-design/topic/short-keys.20for.20tracklist.20management.20RFC 

(Note, this replaces an earlier PR)

### DIFF
--- a/src/library/browse/browsetablemodel.cpp
+++ b/src/library/browse/browsetablemodel.cpp
@@ -323,6 +323,10 @@ void BrowseTableModel::copyTracks(const QModelIndexList& indices) const {
         }
     }
     Clipboard::finish();
+
+    // TODO Investigate if we can also implement cut and paste (via QFile
+    // operations) so mixxx could manage files in the filesystem, rather than
+    // having to go switch between mixxx and the system file browser.
 }
 
 void BrowseTableModel::removeTracks(const QModelIndexList&) {
--- a/src/library/browse/browsetablemodel.cpp
+++ b/src/library/browse/browsetablemodel.cpp
@@ -14,6 +14,7 @@
 #include "moc_browsetablemodel.cpp"
 #include "recording/recordingmanager.h"
 #include "track/track.h"
+#include "util/clipboard.h"
 #include "widget/wlibrarytableview.h"
 
 namespace {
@@ -312,6 +313,16 @@ bool BrowseTableModel::isColumnHiddenByDefault(int column) {
 }
 
 void BrowseTableModel::moveTrack(const QModelIndex&, const QModelIndex&) {
+}
+
+void BrowseTableModel::copyTracks(const QModelIndexList& indices) const {
+    Clipboard::start();
+    for (const QModelIndex& index : indices) {
+        if (index.isValid()) {
+            Clipboard::add(QUrl::fromLocalFile(getTrackLocation(index)));
+        }
+    }
+    Clipboard::finish();
 }
 
 void BrowseTableModel::removeTracks(const QModelIndexList&) {
--- a/src/library/library.cpp
+++ b/src/library/library.cpp
@@ -472,7 +472,7 @@ void Library::addFeature(LibraryFeature* feature) {
     connect(feature,
             &LibraryFeature::pasteFromSidebar,
             this,
-            &Library::slotPasteFromSidebar);
+            &Library::pasteFromSidebar);
     connect(feature,
             &LibraryFeature::showTrackModel,
             this,
@@ -526,10 +526,6 @@ void Library::onPlayerManagerTrackAnalyzerIdle() {
     if (m_pAnalysisFeature) {
         m_pAnalysisFeature->resumeAnalysis();
     }
-}
-
-void Library::slotPasteFromSidebar() {
-    emit pasteFromSidebar();
 }
 
 void Library::slotShowTrackModel(QAbstractItemModel* model) {
--- a/src/library/library.cpp
+++ b/src/library/library.cpp
@@ -389,6 +389,10 @@ void Library::bindLibraryWidget(
             &Library::showTrackModel,
             pTrackTableView,
             &WTrackTableView::loadTrackModel);
+    connect(this,
+            &Library::pasteFromSidebar,
+            m_pLibraryWidget,
+            &WLibrary::pasteFromSidebar);
     connect(pTrackTableView,
             &WTrackTableView::loadTrack,
             this,
@@ -399,6 +403,10 @@ void Library::bindLibraryWidget(
             &Library::slotLoadTrackToPlayer);
     m_pLibraryWidget->registerView(m_sTrackViewName, pTrackTableView);
 
+    connect(m_pLibraryWidget,
+            &WLibrary::setLibraryFocus,
+            m_pLibraryControl,
+            &LibraryControl::setLibraryFocus);
     connect(this,
             &Library::switchToView,
             m_pLibraryWidget,
@@ -462,6 +470,10 @@ void Library::addFeature(LibraryFeature* feature) {
     m_features.push_back(feature);
     m_pSidebarModel->addLibraryFeature(feature);
     connect(feature,
+            &LibraryFeature::pasteFromSidebar,
+            this,
+            &Library::slotPasteFromSidebar);
+    connect(feature,
             &LibraryFeature::showTrackModel,
             this,
             &Library::slotShowTrackModel);
@@ -516,8 +528,12 @@ void Library::onPlayerManagerTrackAnalyzerIdle() {
     }
 }
 
+void Library::slotPasteFromSidebar() {
+    emit pasteFromSidebar();
+}
+
 void Library::slotShowTrackModel(QAbstractItemModel* model) {
-    //qDebug() << "Library::slotShowTrackModel" << model;
+    // qDebug() << "Library::slotShowTrackModel" << model;
     TrackModel* trackModel = dynamic_cast<TrackModel*>(model);
     VERIFY_OR_DEBUG_ASSERT(trackModel) {
         return;
@@ -528,7 +544,7 @@ void Library::slotShowTrackModel(QAbstractItemModel* model) {
 }
 
 void Library::slotSwitchToView(const QString& view) {
-    //qDebug() << "Library::slotSwitchToView" << view;
+    // qDebug() << "Library::slotSwitchToView" << view;
     emit switchToView(view);
 }
 
--- a/src/library/playlisttablemodel.cpp
+++ b/src/library/playlisttablemodel.cpp
@@ -189,29 +189,31 @@ void PlaylistTableModel::selectPlaylist(int playlistId) {
     setSort(defaultSortColumn(), defaultSortOrder());
 }
 
-int PlaylistTableModel::addTracks(const QModelIndex& index,
-        const QList<QString>& locations) {
-    if (locations.isEmpty()) {
+int PlaylistTableModel::addTracksWithTrackIds(const QModelIndex& insertionIndex,
+        const QList<TrackId>& trackIds,
+        int* pOutInsertionPos) {
+    if (trackIds.isEmpty()) {
         return 0;
     }
 
-    QList<TrackId> trackIds = m_pTrackCollectionManager->resolveTrackIdsFromLocations(
-            locations);
-
     const int positionColumn = fieldIndex(ColumnCache::COLUMN_PLAYLISTTRACKSTABLE_POSITION);
-    int position = index.sibling(index.row(), positionColumn).data().toInt();
+    int position = index(insertionIndex.row(), positionColumn).data().toInt();
 
     // Handle weird cases like a drag and drop to an invalid index
     if (position <= 0) {
         position = rowCount() + 1;
     }
 
+    if (pOutInsertionPos) {
+        *pOutInsertionPos = position;
+    }
+
     int tracksAdded = m_pTrackCollectionManager->internalCollection()->getPlaylistDAO().insertTracksIntoPlaylist(
             trackIds, m_iPlaylistId, position);
 
-    if (locations.size() - tracksAdded > 0) {
+    if (trackIds.size() - tracksAdded > 0) {
         qDebug() << "PlaylistTableModel::addTracks could not add"
-                 << locations.size() - tracksAdded
+                 << trackIds.size() - tracksAdded
                  << "to playlist" << m_iPlaylistId;
     }
     return tracksAdded;
--- a/src/library/playlisttablemodel.cpp
+++ b/src/library/playlisttablemodel.cpp
@@ -212,9 +212,12 @@ int PlaylistTableModel::addTracksWithTrackIds(const QModelIndex& insertionIndex,
             trackIds, m_iPlaylistId, position);
 
     if (trackIds.size() - tracksAdded > 0) {
+        QString playlistName = m_pTrackCollectionManager->internalCollection()
+                                       ->getPlaylistDAO()
+                                       .getPlaylistName(m_iPlaylistId);
         qDebug() << "PlaylistTableModel::addTracks could not add"
                  << trackIds.size() - tracksAdded
-                 << "to playlist" << m_iPlaylistId;
+                 << "to playlist id" << m_iPlaylistId << "name" << playlistName;
     }
     return tracksAdded;
 }
--- a/src/library/sidebarmodel.cpp
+++ b/src/library/sidebarmodel.cpp
@@ -152,9 +152,9 @@ void SidebarModel::clear(const QModelIndex& index) {
     if (index.internalPointer() == this) {
         m_sFeatures[index.row()]->clear();
     } else {
-        TreeItem* tree_item = (TreeItem*)index.internalPointer();
-        if (tree_item) {
-            LibraryFeature* feature = tree_item->feature();
+        TreeItem* pTreeItem = (TreeItem*)index.internalPointer();
+        if (pTreeItem) {
+            LibraryFeature* feature = pTreeItem->feature();
             feature->clearChild(index);
         }
     }
@@ -164,9 +164,9 @@ void SidebarModel::cut(const QModelIndex& index) {
     if (index.internalPointer() == this) {
         m_sFeatures[index.row()]->cut();
     } else {
-        TreeItem* tree_item = (TreeItem*)index.internalPointer();
-        if (tree_item) {
-            LibraryFeature* feature = tree_item->feature();
+        TreeItem* pTreeItem = (TreeItem*)index.internalPointer();
+        if (pTreeItem) {
+            LibraryFeature* feature = pTreeItem->feature();
             feature->cutChild(index);
         }
     }
@@ -176,9 +176,9 @@ void SidebarModel::copy(const QModelIndex& index) const {
     if (index.internalPointer() == this) {
         m_sFeatures[index.row()]->copy();
     } else {
-        TreeItem* tree_item = (TreeItem*)index.internalPointer();
-        if (tree_item) {
-            LibraryFeature* feature = tree_item->feature();
+        TreeItem* pTreeItem = (TreeItem*)index.internalPointer();
+        if (pTreeItem) {
+            LibraryFeature* feature = pTreeItem->feature();
             feature->copyChild(index);
         }
     }
@@ -188,9 +188,9 @@ void SidebarModel::paste(const QModelIndex& index) {
     if (index.internalPointer() == this) {
         m_sFeatures[index.row()]->paste();
     } else {
-        TreeItem* tree_item = (TreeItem*)index.internalPointer();
-        if (tree_item) {
-            LibraryFeature* feature = tree_item->feature();
+        TreeItem* pTreeItem = (TreeItem*)index.internalPointer();
+        if (pTreeItem) {
+            LibraryFeature* feature = pTreeItem->feature();
             feature->pasteChild(index);
         }
     }
@@ -200,9 +200,9 @@ void SidebarModel::selectAll(const QModelIndex& index) {
     if (index.internalPointer() == this) {
         m_sFeatures[index.row()]->selectAll();
     } else {
-        TreeItem* tree_item = (TreeItem*)index.internalPointer();
-        if (tree_item) {
-            LibraryFeature* feature = tree_item->feature();
+        TreeItem* pTreeItem = (TreeItem*)index.internalPointer();
+        if (pTreeItem) {
+            LibraryFeature* feature = pTreeItem->feature();
             feature->selectAllChild(index);
         }
     }
--- a/src/library/sidebarmodel.cpp
+++ b/src/library/sidebarmodel.cpp
@@ -148,6 +148,66 @@ QModelIndex SidebarModel::getFeatureRootIndex(LibraryFeature* pFeature) {
     return ind;
 }
 
+void SidebarModel::clear(const QModelIndex& index) {
+    if (index.internalPointer() == this) {
+        m_sFeatures[index.row()]->clear();
+    } else {
+        TreeItem* tree_item = (TreeItem*)index.internalPointer();
+        if (tree_item) {
+            LibraryFeature* feature = tree_item->feature();
+            feature->clearChild(index);
+        }
+    }
+}
+
+void SidebarModel::cut(const QModelIndex& index) {
+    if (index.internalPointer() == this) {
+        m_sFeatures[index.row()]->cut();
+    } else {
+        TreeItem* tree_item = (TreeItem*)index.internalPointer();
+        if (tree_item) {
+            LibraryFeature* feature = tree_item->feature();
+            feature->cutChild(index);
+        }
+    }
+}
+
+void SidebarModel::copy(const QModelIndex& index) const {
+    if (index.internalPointer() == this) {
+        m_sFeatures[index.row()]->copy();
+    } else {
+        TreeItem* tree_item = (TreeItem*)index.internalPointer();
+        if (tree_item) {
+            LibraryFeature* feature = tree_item->feature();
+            feature->copyChild(index);
+        }
+    }
+}
+
+void SidebarModel::paste(const QModelIndex& index) {
+    if (index.internalPointer() == this) {
+        m_sFeatures[index.row()]->paste();
+    } else {
+        TreeItem* tree_item = (TreeItem*)index.internalPointer();
+        if (tree_item) {
+            LibraryFeature* feature = tree_item->feature();
+            feature->pasteChild(index);
+        }
+    }
+}
+
+void SidebarModel::selectAll(const QModelIndex& index) {
+    if (index.internalPointer() == this) {
+        m_sFeatures[index.row()]->selectAll();
+    } else {
+        TreeItem* tree_item = (TreeItem*)index.internalPointer();
+        if (tree_item) {
+            LibraryFeature* feature = tree_item->feature();
+            feature->selectAllChild(index);
+        }
+    }
+}
+
 QModelIndex SidebarModel::parent(const QModelIndex& index) const {
     //qDebug() << "SidebarModel::parent index=" << index.getData();
     if (index.isValid()) {
@@ -475,7 +535,7 @@ QModelIndex SidebarModel::translateIndex(
 }
 
 void SidebarModel::slotDataChanged(const QModelIndex& topLeft, const QModelIndex& bottomRight) {
-    //qDebug() << "slotDataChanged topLeft:" << topLeft << "bottomRight:" << bottomRight;
+    // qDebug() << "slotDataChanged topLeft:" << topLeft << "bottomRight:" << bottomRight;
     QModelIndex topLeftTranslated = translateSourceIndex(topLeft);
     QModelIndex bottomRightTranslated = translateSourceIndex(bottomRight);
     emit dataChanged(topLeftTranslated, bottomRightTranslated);
@@ -499,8 +559,8 @@ void SidebarModel::slotRowsInserted(const QModelIndex& parent, int start, int en
     Q_UNUSED(parent);
     Q_UNUSED(start);
     Q_UNUSED(end);
-    //qDebug() << "slotRowsInserted" << parent << start << end;
-    //QModelIndex newParent = translateSourceIndex(parent);
+    // qDebug() << "slotRowsInserted" << parent << start << end;
+    // QModelIndex newParent = translateSourceIndex(parent);
     endInsertRows();
 }
 
--- a/src/library/sidebarmodel.cpp
+++ b/src/library/sidebarmodel.cpp
@@ -151,36 +151,6 @@ QModelIndex SidebarModel::getFeatureRootIndex(LibraryFeature* pFeature) {
 void SidebarModel::clear(const QModelIndex& index) {
     if (index.internalPointer() == this) {
         m_sFeatures[index.row()]->clear();
-    } else {
-        TreeItem* pTreeItem = (TreeItem*)index.internalPointer();
-        if (pTreeItem) {
-            LibraryFeature* feature = pTreeItem->feature();
-            feature->clearChild(index);
-        }
-    }
-}
-
-void SidebarModel::cut(const QModelIndex& index) {
-    if (index.internalPointer() == this) {
-        m_sFeatures[index.row()]->cut();
-    } else {
-        TreeItem* pTreeItem = (TreeItem*)index.internalPointer();
-        if (pTreeItem) {
-            LibraryFeature* feature = pTreeItem->feature();
-            feature->cutChild(index);
-        }
-    }
-}
-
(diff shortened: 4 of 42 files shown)


ronso0:
FYI I started reviewing in small chunks. Will publish asap.

Re UX, it's not possible anymore to drag track files from file browser to Crates for example :|
Also I expected being able to copy/paste URLs from file browser.

I had some trouble reviewing this tbh.
Do you mind splitting up the first commit in order to separate the `addTracks`/`addTracksWithTrackIds` change set from the hotkey implementation?

m0dB:
> FYI I started reviewing in small chunks. Will publish asap.

Thanks! Much appreciated!
 
> Re UX, it's not possible anymore to drag track files from file browser to Crates for example :|

Hm, that is surprising, but indeed it's something I didn't try (because I never do that)

> Also I expected being able to copy/paste URLs from file browser.

Right. Also something I never do, though I do think it did work at some point: initially I implemented the clipboard handling with plain text, but switched to mimeData() / setMimeData() instead, on suggestion of @daschuer . I will see what I get on the clipboard if I copy/paste something from the Finder (and hopefully this will be the same across platforms...)

Anyway, I expect this will be easy to support.

> I had some trouble reviewing this tbh. Do you mind splitting up the first commit in order to separate the `addTracks`/`addTracksWithTrackIds` change set from the hotkey implementation?

I will see if I can do that, from the top of my head I am not sure about the ramifications.

Now, I am not sure how to do this git-wise.

ronso0:
Thanks.

> how to do this git-wise

Reset to [first commit]^1. Then try to `git add -p` and commit isolated chunks. If that's not possible, never mind.
FWIW I have a git alias `gr1` which does `git reset HEAD^1`. Very helpful in case I messed up a commit, like when unrelated changes or debug stuff slipped in.

m0dB:
> Re UX, it's not possible anymore to drag track files from file browser to Crates for example :| Also I expected being able to copy/paste URLs from file browser.

I just tested this and both work fine from the macOS Finder :-/
 
Can you see what you have on the clipboard when copy/paste from the file browser?

```
      qDebug() << "CLIPBOARD MIMEDATA->URLS:" << data->urls() <<" TEXT" <<  QA    pplication::clipboard()->text();
```

ronso0:
FYI I also experienced a segfault (only with this branch)
- Mixxx window not visible
- select files in browser, start dragging
- Alt+Tab to raise Mixxx
- as soon as I hover Mixxx (tracks table I think, Crate selected) Mixxx crashes

m0dB:
Or when you say file browser, you mean the "file browser" tree under computer in the side bar? (From there, drag&drop works for me, copy/paste indeed doesn't work)

m0dB:
> Reset to [first commit]^1. Then try to `git add -p` and commit isolated chunks. If that's not possible, never mind.

Done. There was only one file that had changes for both the addTracks/addTracksWithTrackIds change and the cut/copy/paste change so I edited that for the first commit. Hope this helps!

m0dB:
> FYI I also experienced a segfault (only with this branch)
> 
> * Mixxx window not visible
> * select files in browser, start dragging
> * Alt+Tab to raise Mixxx
> * as soon as I hover Mixxx (tracks table I think, Crate selected) Mixxx crashes
> 
> I'll try to catch a backtrace soonish.

I can't reproduce this crash on macOS

ronso0:
> Or when you say file browser, you mean the "file browser" tree under computer in the side bar? (From there, drag&drop works for me, copy/paste indeed doesn't work)

I referred to the OS' file browser, Thunar in my case.

ronso0:
I can reproduce the segfault when attempting drop files (no Alt+tab involved):
```
debug [Main] dropEventFiles() true (QUrl("file:///some/file/path/El Payo - Night Crawler xXx.flac"), QUrl("file:///some/file/path/El Payo - The Battle Against Giants xXx.flac"))
debug [Main] mimeData.hasText() true ""
debug [Main] SqlTransaction - Started new SQL database transaction on "MIXXX-1"
debug [Main] SqlTransaction - Committed SQL database transaction on "MIXXX-1"
debug [Main] BaseTrackCache(0x558824948b30) updateIndexWithQuery took 0 ms
<------ roughly 10 seconds passed ------>
[1]    326895 segmentation fault (core dumped)  ./mixxx --developer --settingsPath /***
```

m0dB:
Can you provide a call stack? Only happens with crates and not with playlists? And are you sure this works ok without the PR?

ronso0:
This happens only with this branch, and only with crates.
* select a crate
* select a few files in OS file browser
* drag files onto track table

```
CrateTableModel::addTracksWithTrackIds (this=0x555565d48c58, index=..., trackIds=..., pOutInsertionPos=0x0) at /src/library/trackset/crate/cratetablemodel.cpp:160
160	        *pOutInsertionPos = 0;
```

Full backtrace (why is it so huge??)
[segfault_shortkeys_drag-from-file-browser-to-crate.txt](https://github.com/mixxxdj/mixxx/files/12910607/segfault_shortkeys_drag-from-file-browser-to-crate.txt)

ronso0:
Idea for extending this feature: (certainly out of scope for this PR):
if Cut/Paste would also work in the Computer feature (i.e. do real file operations) that would be awesome since that would allow basic track file management from within Mixxx, i.e. without having to switch back and forth between Mixxx and file browser.

ronso0 (changes requested):
Okay, so I managed to finish my first, rather shallow review.

One thing I noticed: from my file manager I'm used to Cut only finally cutting when I paste the selection elsewhere. I.e. until Paste, Cut is rather a Cut preview which is aborted as soon as I cut/copy something else. Not sure how this was on Windows (long time ago, luckily), and maybe this behaviour is simply specific to Ubuntu or XFCE?

I'm not requesting this 'feature' here, just want to note that, in the Cut/Copy/Paste documentation, we should explicitly mention the Cut (remove) behaviour.

m0dB:
Thanks for taking the time to review this, @ronso0 . I agree with all your comments and will address them soon.

m0dB:
The crash should be solved. Can you please confirm, @ronso0 ?

src/library/browse/browsetablemodel.cpp
```diff
@@ -316,6 +317,16 @@ bool BrowseTableModel::isColumnHiddenByDefault(int column) {
...
+void BrowseTableModel::copyTracks(const QModelIndexList& indices) const {
+    Clipboard::start();
+    for (const QModelIndex& index : indices) {
+        if (index.isValid()) {
+            Clipboard::add(QUrl::fromLocalFile(getTrackLocation(index)));
+        }
+    }
+    Clipboard::finish();
+}
+
```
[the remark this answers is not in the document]
m0dB: done

src/library/playlisttablemodel.cpp
```diff
@@ -189,29 +189,31 @@ void PlaylistTableModel::selectPlaylist(int playlistId) {
...
+    }
+
     int tracksAdded = m_pTrackCollectionManager->internalCollection()->getPlaylistDAO().insertTracksIntoPlaylist(
             trackIds, m_iPlaylistId, position);
 
-    if (locations.size() - tracksAdded > 0) {
+    if (trackIds.size() - tracksAdded > 0) {
         qDebug() << "PlaylistTableModel::addTracks could not add"
-                 << locations.size() - tracksAdded
+                 << trackIds.size() - tracksAdded
```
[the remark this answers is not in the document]
m0dB: done

src/library/sidebarmodel.cpp
```diff
@@ -150,6 +150,66 @@ QModelIndex SidebarModel::getFeatureRootIndex(LibraryFeature* pFeature) {
...
 }
 
+void SidebarModel::clear(const QModelIndex& index) {
+    if (index.internalPointer() == this) {
+        m_sFeatures[index.row()]->clear();
+    } else {
+        TreeItem* tree_item = (TreeItem*)index.internalPointer();
+        if (tree_item) {
+            LibraryFeature* feature = tree_item->feature();
+            feature->clearChild(index);
```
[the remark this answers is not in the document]
m0dB: done

src/library/trackset/basetracksetfeature.cpp
```diff
@@ -12,6 +12,11 @@ BaseTrackSetFeature::BaseTrackSetFeature(
           m_pSidebarModel(make_parented<TreeItemModel>(this)) {
 }
 
+void BaseTrackSetFeature::pasteChild(
+        const QModelIndex&) {
```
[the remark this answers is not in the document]
m0dB: yeah, don't know why clang-format didn't correct this... done.

src/widget/wlibrary.cpp
```diff
@@ -76,6 +76,14 @@ void WLibrary::switchToView(const QString& name) {
     }
 }
 
+void WLibrary::pasteFromSidebar() {
+    QWidget* current = currentWidget();
+    LibraryView* view = dynamic_cast<LibraryView*>(current);
+    if (view) {
+        view->pasteFromSidebar();
+    }
```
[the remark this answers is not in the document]
m0dB: fixed them all

src/library/library.cpp
```diff
@@ -466,6 +474,10 @@ void Library::addFeature(LibraryFeature* feature) {
     }
     m_features.push_back(feature);
     m_pSidebarModel->addLibraryFeature(feature);
+    connect(feature,
+            &LibraryFeature::pasteFromSidebar,
+            this,
+            &Library::slotPasteFromSidebar);
```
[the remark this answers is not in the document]
m0dB: done

m0dB:
@ronso0 I have made all requested changes. please re-review and resolve the conversations accordingly. thanks!

m0dB:
> One thing I noticed: from my file manager I'm used to Cut only finally cutting when I paste the selection elsewhere. I.e. until Paste, Cut is rather a Cut preview which is aborted as soon as I cut/copy something else. Not sure how this was on Windows (long time ago, luckily), and maybe this behaviour is simply specific to Ubuntu or XFCE?

I believe on Windows it's the same. On macOS (Finder) Cmd+X does nothing, but you can do Cmd+C , Cmd+Option+V to move files (I had to look this up, I always use the terminal for file management).

FWIW, in Spotify, shortkey cut behaves like what I have implemented in mixxx (paste however, behaves slightly differently, for good reasons). I prefer it like this and the implementation sure is simpler.

ronso0:
> @ronso0 I have made all requested changes. please re-review and resolve the conversations accordingly. thanks!

Thanks, done.
I'll take another look soonish. Having some trouble to review from the diff only, will need to view this in my IDE.

src/widget/wlibrary.cpp
```diff
@@ -142,3 +150,10 @@ bool WLibrary::event(QEvent* pEvent) {
     }
     return QStackedWidget::event(pEvent);
 }
+
+void WLibrary::keyPressEvent(QKeyEvent* pEvent) {
+    if (pEvent->key() == Qt::Key_Left && pEvent->modifiers() & Qt::ControlModifier) {
+        emit setLibraryFocus(FocusWidget::Sidebar);
```
ronso0, line 156: This is not how I expected the feature to work, it conflicts/overlaps with `WLibraryTableView::moveCursor`:
Ctrl+Left moves the cursor until it hits the left-most cell, **then** focus is moved to the sidebar, which kind of defeats the purpose of having alternative to Tab / Shift+Tab.

Tbh I don't see the benefit of `WLibraryTableView::moveCursor` handling Ctrl+Left/Right. If it's about (implicit) horzontal scrolling we could map that to Alt+Left/Right (didn't investigate what the standard behaviour is), and ignore Ctrl+Left/Right.

What do you think?

m0dB: I just rebased the code, and I don't see what you mean. Without this key handling in wlibrary.cpp ctrl-left/right does nothing.

ronso0: Right, our `moveCursor` does nothing, it just hands over to QTableView::moveCursor.
Ctrl+Left/Right allows moving the focus (table cell) while preserving the (multi-)selection.

With this branch I can move the focus around as before, and **only if I'm in the first column** Ctrl+left would move focus to the sidebar.
Is this how you intended this to work?

The other way around, Ctrl+Right moves focus from the sidebar to the tracks table, but there it restores the previous focus cell (expected, might be any column).
Therefore it's not possible to use Ctrl+Left/Right exactly as Tab / Shift+Tab.

m0dB: Ah, that is not how it is behaving on macOS. Here with ctrl (cmd on macOS) right / left switch between sidebar and library just like tab / shift+tab, independent of the column in which the focus is.

When moving with left/right **without** modifier key the full row stays selected anyway. Isn't that the case on Linux? If I select multiple rows (e.g. with shift up/down, or which ctrl-mouseclick), left/right **without** modifier key results in only the current row being selected. Shift left/right doesn't seem to do anything (except triggering a redraw of the waveforms for some reason)

Still, I don't really understand what "moving the focus (table cell) while preserving the (multi-)selection." implies. Is that referring to a multi-row selection? If so, I don't see the value of that at all. Do you? I have a hard time imagining a situation in which I would like to navigate the focus to a different cell while I am doing something with a multi row selection.

ronso0: > When moving with left/right without modifier key the full row stays selected anyway.

> I select multiple rows (e.g. with shift up/down, or which ctrl-mouseclick), left/right without modifier key results in only the current row being selected.

Same on Linux in both cases.

> till, I don't really understand what "moving the focus (table cell) while preserving the (multi-)selection." implies. Is that referring to a multi-row selection? If so, I don't see the value of that at all.

Yes, that's what I mean.
Well, the only thing coming to mind is using the currently focused column for sorting via CO (`[Library]sort_column_toggle` value `0`), though I can't tell if anyone would ever use it in such a scenario, probably not.

ronso0: IIUC the only way forward to enforce the intended focus back/forth behaviour would be block the native Ctrl+Left/Right cursor event and do the focus move, right?

m0dB:
Can you check if this works for you?

ronso0:
I guess that will work as intended (not for me since I usually have two hands on the leyboard and am used to Tab/Shift+Tab that I won't switch to Ctrl+Left/Right in this context).

I'll take a look at the actual hotkey feature ; ) after 2.4

manuelbertelli (approved):
:shipit:

daschuer:
It looks like it is KDE Thunar has also adopted this type: https://indigo.re/posts/2021-12-21-clipboard-data.html

daschuer:
The code is here: https://github.com/mixxxdj/mixxx/pull/12888

src/widget/wlibrarysidebar.cpp
```diff
@@ -239,12 +239,36 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
     // TODO(XXX) Should first keyEvent ensure previous item has focus? I.e. if the selected
     // item is not focused, require second press to perform the desired action.
 
-    // make the selected item the navigation starting point
+    SidebarModel* sidebarModel = qobject_cast<SidebarModel*>(model());
+    QModelIndexList selectedIndices = selectionModel()->selectedRows();
+    if (sidebarModel && !selectedIndices.isEmpty()) {
+        QModelIndex index = selectedIndices.at(0);
+        if (event->matches(QKeySequence::Delete) || event->key() == Qt::Key_Backspace) {
```
daschuer, line 246: What does this do? I cannot see anything.

src/widget/wlibrarysidebar.cpp
```diff
@@ -239,12 +239,36 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
...
+            sidebarModel->clear(index);
+            return;
+        }
+        if (event->matches(QKeySequence::Cut)) {
+            sidebarModel->cut(index);
```
daschuer, line 251: It feels wrong that the clipboard is not cleared, when trying to copy the whole playlist.

src/widget/wlibrarysidebar.cpp
```diff
@@ -239,12 +239,36 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
...
+            return;
+        }
+        if (event->matches(QKeySequence::Copy)) {
```
daschuer, line 254: It feels wrong that this copies a track from the list where you have no control, because the focus is still on the SideBar

src/widget/wlibrarysidebar.cpp
```diff
@@ -239,12 +239,36 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
...
+            sidebarModel->copy(index);
+            return;
+        }
+        if (event->matches(QKeySequence::Paste)) {
```
daschuer, line 258: I think we should remove all these shortcut from the side-bar. The only reasonable shortcut is this. 
Select a playlist and drop a track to it.

src/widget/wlibrarysidebar.cpp
```diff
@@ -239,12 +239,36 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
...
+            sidebarModel->paste(index);
+            return;
+        }
+        if (event->matches(QKeySequence::SelectAll)) {
+            sidebarModel->selectAll(index);
```
daschuer, line 263: This does nothing

src/widget/wlibrarysidebar.cpp
```diff
@@ -268,6 +292,14 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
         emit pressed(selIndex);
         return;
     }
+    case Qt::Key_Right: {
+        if (event->modifiers() & Qt::ControlModifier) {
+            emit setLibraryFocus(FocusWidget::TracksTable);
```
daschuer, line 297: My expectation was to have this feature without a modifier. 
An alternative would be to use "Enter" this seems to be unused when selecting a playlist.

src/widget/wtracktableview.cpp
```diff
@@ -840,16 +905,41 @@ void WTrackTableView::keyPressEvent(QKeyEvent* event) {
...
-    } break;
+        return;
+    }
     default:
-        QTableView::keyPressEvent(event);
+        break;
+    }
+    TrackModel* trackModel = getTrackModel();
+    if (trackModel && !trackModel->isLocked()) {
+        if (event->matches(QKeySequence::Delete) || event->key() == Qt::Key_Backspace) {
```
daschuer, line 921: Backspace is used to navigate back in NEMO. It is sup-prizing that the track is deleted without a warning. 
Can we remove it here?

daschuer (changes requested):
Some more comments

src/widget/wlibrarysidebar.cpp
m0dB: No, it doesn't. I will remove it.

src/widget/wlibrarysidebar.cpp
m0dB: Yes, I left all these methods as virtuals without implementations, during development, to decide what made sense to implement on the go. It might be better to remove them now altogether, and add them back if needed.

m0dB: Would you prefer that?

src/widget/wlibrarysidebar.cpp
m0dB: This is intentional as it allows for fast navigation through both sidebar and tracklist using only combination of cursor keys + modifier.

m0dB: And it is symetrical with left+modifier to go back.

That said, I see no reason not to use Enter as well.

src/widget/wtracktableview.cpp
m0dB: Yes. Also I see that ctrl->backspace also deleted the track (with confirmation popup). It surprised me I missed that.

src/widget/wlibrarysidebar.cpp
m0dB: I don;t think this does anything

src/widget/wlibrarysidebar.cpp
m0dB: I don;t think this does anything

m0dB:
@daschuer I have looked at your review, and answered some things and asked you some questions in return. I will try to address everything this weekend (and how will depend on your answers). Overall, I think the changes will be minor, mostly removing some code that leads to empty functions.

And thanks for taking care of the linux file manager compatibility!

src/widget/wlibrarysidebar.cpp
daschuer: I have no strong opinion here.

src/widget/wlibrarysidebar.cpp
daschuer: Yes. We should not allow to "remotely" copy or delete a track when the focus is on the sidebar. The only thing that we may allow is Ctrl+V analog to doping a track to the playlist in the sidebar.

src/widget/wlibrarysidebar.cpp
daschuer: I may have commented the wrong place. This is also covered in the comment below about remote copy from the sidebar.

src/widget/wlibrarysidebar.cpp
daschuer: My original expectation was that we copy the whole playlist when Crl+C a playlist in the sidebar. 
But that would be a new feature. For now the behaviour is OK after retesting and we can resole it.

src/widget/wlibrarysidebar.cpp
daschuer: Uh, that is dangerous. I think we should remove that feature.

src/widget/wlibrarysidebar.cpp
daschuer: I cant reproduce the behaviour I might have seen. So it was probably a wrong observation.

src/widget/wlibrarysidebar.cpp
ronso0: Let's please not complicate the navigation.
Right should allow descending into features (directory tree) and stop at leafes. One more (accidental) press making it move to the tracks table is unexected and thereby confusing. Also, Left in the tracks table does not move back to the sidebar.

src/widget/wlibrarysidebar.cpp
m0dB: Ok, but still, I will remove the code that only can lead to empty methods.

m0dB:
Hm, I did a git pull --rebase, locally... Not sure what happened here.

src/widget/wtracktableview.cpp
```diff
@@ -1279,7 +1376,7 @@ bool WTrackTableView::hasFocus() const {
 }
 
 void WTrackTableView::setFocus() {
-    QWidget::setFocus();
+    QWidget::setFocus(Qt::OtherFocusReason);
```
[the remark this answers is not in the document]
m0dB: can I resolve this conversation, @ronso0  ?

src/widget/wtracktableview.cpp
m0dB: Removed. Please resolve this conversation, @daschuer

src/widget/wlibrarysidebar.cpp
m0dB: All methods that are empty or only called empty methods have been removed. Please resolve this conversation, @daschuer

m0dB:
I believe I addressed all issues and also solved a minor bug where after pasting multiple tracks 1 track wasn't selected.

daschuer:
The history looks wrong. I have rebated it again into my repro: 
https://github.com/daschuer/mixxx/tree/tracklist_shortkeys

Maybe you can reset to that and then rebate again to have your as committer. 
You may also try to sqash fixup commits using `rebase upstream/main -i`

m0dB:
I have rebased on main, and diffed against your branch after rebase that one on main as well to make sure there were no differences, so it should be fine now. (btw, you have both tracklist_shortkeys and tracklists_shortkeys, which is a bit confusing)

(I know rebasing is not ideal, but as all reviews have been addressed I hope it's no problem)

I think this can be merged now.

ronso0:
I did another quick test (not that 'quick' when I copied my entire 16k library :roll_eyes: ) and I noticed that it's possible to paste into History playlists (not the root item though). We disallow dragging tracks into those so for consistency we should also disallow pasting tracks.

While this override
`void pasteChild(const QModelIndex&) override {};`
added to setlogfeature.h seems to be sufficient for the sidebar, we need something else for the tracks view.
Maybe reuse `trackModel->getCapabilities` (seen in [WTrackTableView](https://github.com/mixxxdj/mixxx/blob/5524ca032860478c67d1f04dbe59ba0912a91681/src/widget/wtracktableview.cpp#L302-L307)) and [`Capability::ReceiveDrops`](https://github.com/mixxxdj/mixxx/blob/5524ca032860478c67d1f04dbe59ba0912a91681/src/library/playlisttablemodel.cpp#L372-L377)?

ronso0:
...and we need to disallow pasting into **locked** sets, probably with `Trackmodel::isLocked()`.

m0dB:
Well spotted, @ronso0 ! I'll take care of this (though PRs are welcome too)

ronso0:
Sure.
I also discovered a few bugs/flaws with the crate model and removing tracks.
#12926

src/widget/wtracktableview.cpp
```diff
@@ -819,6 +824,68 @@ TrackModel* WTrackTableView::getTrackModel() const {
...
+    // in this way.
+    if (rows.back() > lastVisibleRow) {
+        selectRow(rows.back());
+    } else {
+        selectRow(rows.front());
+    }
+
+    // Select all the rows that we pasted
+    for (const auto row : rows) {
+        selectionModel()->select(model()->index(row, 0),
```
ronso0, line 884: In other places in this file we assert there is a selection model.
I'm not sure whether this test is actually required (don't know how QTableView could _not_ have a selection model), but let's check here, too,  for consistency.

src/util/clipboard.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+namespace {
+
+QByteArray urlsToUtf8(const QList<QUrl>& urls) {
+    QByteArray result;
+    for (const QUrl& url : urls) {
+        result += url.toEncoded() + '\n';
+    }
+    if (!result.isEmpty())
+        result.chop(1);
```
daschuer, line 15:
```suggestion
    if (!result.isEmpty()) {
        result.chop(1);
    }
```

daschuer (changes requested):

src/widget/wtracktableview.cpp
m0dB: Done, by added a helper function getSelectedRows() . With that, all selectionModel() access is protected. Please Resolve conversation if you agree with this solution, @ronso0

ronso0:
Since this is a mixxx feature branch I took the liberty to merge main and resolve the conflicts.
IMO this is ready to roll.

@daschuer do you agree?

ronso0:
Just one thing I noticed (not a blocker):
find tracks in a file browser that are not in the Mixxx library
* paste (sidebar or tracks table):
doesn't work `TrackDAO::getTrackIds(): Found only 0 of 1 tracks in library`
* find the same track in Computer tree, copy, paste into crate (sidebar or tracks table):
works
* drag those to playlists/crates:
works

My file browser is Thunar and pasting tracks that are in the library works.

daschuer (approved):
LGTM, Thank you very much. 
I have fixed dropping new track in the last commit.

daschuer:
@ronso0 merge?

ronso0:
Thank you, it works as desired now!

Only issue is that I'm running into the build warning below.
I'm using Qt 6.2 from your backport ppa.

<details>
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:1878,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qhashfunctions.h:44,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:47,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/QList:1,
                 from /mixxx_source/src/master/src/util/clipboard.h:3,
                 from /mixxx_source/src/master/src/util/clipboard.cpp:1:
In static member function ‘static void QConcatenable<QByteArray>::appendTo(const QByteArray&, char*&)’,
    inlined from ‘static void QConcatenable<QStringBuilder<A, B> >::appendTo(const type&, T*&) [with T = char; A = QByteArray; B = char]’ at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstringbuilder.h:428:35,
    inlined from ‘QByteArray& QtStringBuilder::appendToByteArray(QByteArray&, const QStringBuilder<A, B>&, char) [with A = QByteArray; B = char]’ at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstringbuilder.h:459:52,
    inlined from ‘QByteArray& operator+=(QByteArray&, const QStringBuilder<A, B>&) [with A = QByteArray; B = char]’ at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstringbuilder.h:476:46,
    inlined from ‘QByteArray {anonymous}::urlsToUtf8(const QList<QUrl>&)’ at /mixxx_source/src/master/src/util/clipboard.cpp:12:37,
    inlined from ‘static void Clipboard::finish()’ at /mixxx_source/src/master/src/util/clipboard.cpp:35:72:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qstringbuilder.h:411:20: warning: writing 1 byte into a region of size 0 [-Wstringop-overflow=]
  411 |             *out++ = *a++;
      |             ~~~~~~~^~~~~~
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:46:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qarraydata.h: In static member function ‘static void Clipboard::finish()’:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qarraydata.h:129:8: note: at offset [8, 24] into destination object ‘QTypedArrayData<char>::<anonymous>’ of size 24
  129 | struct QTypedArrayData
      |        ^~~~~~~~~~~~~~~
</details>

ronso0:
Thank you, the warning is gone.

One last issue:
when I'm in a playlist, have the last track selected and want to append tracks (= paste tracks after the last track), they get inserted before the last position. Moving the to the end requires a drag.

If no track is selected paste appends tracks as expected. With playlists filling the track table widget there is no empty space to click (deselect tracks).

I understand that this is the desired implementation to insert tracks at a certain position, but how to achieve append without having to drag?
* add Alt+Up/Down hotkeys to move the selection? just a workaround, not discoverable
* add a special rule for 'last track selected'? might be unexpected..

daschuer:
Due to the wrap-around, all hacks using the DOWN key do not work. 

I can think of adding an empty dummy row to every table. This would be discoverable but is probably not easy to implement without messing around. 
Or past below if the rightmost column is selected of the bottom track. This is however also not discoverable. 

I think I would keep this as a feature request and use the workaround of pasting to the side bar.

m0dB:
> I understand that this is the desired implementation to insert tracks at a certain position, but how to achieve append without having to drag?

Yes. The problem will occur either at the bottom or at the top (in spotify for example the problem is at the top). Pasting in the sidebar is indeed the solution.

ronso0:
Alright then, who's gonna press Merge : )

ronso0:
oh, CI still running.

daschuer:
@ronso0 Your Review is still red. So it is looks like it is up to you to approve and merge.

ronso0 (approved):

ronso0:
Thank you @m0dB !

m0dB:
Thanks for all the feedback and improvements!

PR #12021, opened 2023-09-24
CO Renaming (Pt. 4): VU Meter Controls

Holzhaus:
Part of #11931.


daschuer:
We have an issue that our VU-Meter is no VU-Meter 
> The current Mixxx VU meter is somewhere between a paek meter and a VU meter, offering a reasonable one fits all type, unfortunately not comparable to other devices or any standard scale.
See: https://github.com/mixxxdj/mixxx/issues/7888
Can we use a better name?

Holzhaus:
> Can we use a better name?

Do you have a suggestion?

Though I think this is not really an issue and we may add support for different modes later on that can be changed in the settings, so its rather future-proof :)

daschuer:
https://en.m.wikipedia.org/wiki/VU_meter 

For my feeling the "VU Meter" for volume units is kind of deprecated. But the term is widely used. 
If we decide to make it a more recent LUFS meter, the term would be wrong, also if we make it a peak indicator of the digital scale. 

I have looked around how DJ controller named it and found "audio meter" and  "level meter".
Not sure if this is better to understand.

Holzhaus:
Well, `level_indicator` or `level_meter` is okay and pretty generic. On the other hand, it may be confusing because it's not really google-able. The wiki article on "Sound Level Meter" mostly refers to microphones: https://en.wikipedia.org/wiki/Sound_level_meter

JoergAtGithub:
Let's stay with standard terms

daschuer:
Ok, then let's stick with the old.

JoergAtGithub:
LGTM! Thank you!

PR #11121, opened 2022-12-08
JACK buffer size fix.

daschuer:
This PR disposes all stored buffer size values and uses consequently the value passed as callback parameter.
This is required because the JACK buffer size is externally controlled and can change at any time.
Current main uses the crazed out buffer size form hardware preferences. 

Because if that the timing compensation with two soundcards or the network clock fails and is now grayed out in the hardware preferences. 

Note: Multi soundcard setup is working, but because if the joint stream in JACK and the split-up in Portaudio. It adds an extra buffer latency for the non main device. It is also known that the clock synchronisation in native JACK is broken and the situation is not clear in case in pipewire. 

So if you want a minimum latency setup, use ALSA directly. 

This PR is on top of https://github.com/mixxxdj/mixxx/pull/11092

Many thanks to @robbert-vdh for finding the original issues and for the initial PR

PR #12022, opened 2023-09-24
CO Renaming (Pt. 5): Latency Control

Holzhaus:
*Part of https://github.com/mixxxdj/mixxx/issues/11931.*

Moves the `latency` control into the `[App]` group.

### DIFF
--- a/src/engine/enginemixer.cpp
+++ b/src/engine/enginemixer.cpp
@@ -80,9 +80,10 @@ EngineMixer::EngineMixer(
     m_pMainSampleRate->set(44100.);
 
     // Latency control
-    m_pMainLatency = new ControlObject(ConfigKey(group, "latency"),
+    m_pMainLatency = new ControlObject(ConfigKey(kAppGroup, QStringLiteral("output_latency_ms")),
             true,
             true); // reported latency (sometimes correct)
+    m_pMainLatency->addAlias(ConfigKey(kLegacyGroup, QStringLiteral("latency")));
     m_pAudioLatencyOverloadCount = new ControlObject(
             ConfigKey(kAppGroup, QStringLiteral("audio_latency_overload_count")));
     m_pAudioLatencyOverloadCount->addAlias(ConfigKey(
--- a/src/engine/enginemixer.cpp
+++ b/src/engine/enginemixer.cpp
@@ -74,16 +74,17 @@ EngineMixer::EngineMixer(
     m_pWorkerScheduler->start(QThread::HighPriority);
 
     // Main sample rate
-    m_pMainSampleRate = new ControlObject(
+    m_pSampleRate = new ControlObject(
             ConfigKey(kAppGroup, QStringLiteral("samplerate")), true, true);
-    m_pMainSampleRate->addAlias(ConfigKey(group, QStringLiteral("samplerate")));
-    m_pMainSampleRate->set(44100.);
+    m_pSampleRate->addAlias(ConfigKey(group, QStringLiteral("samplerate")));
+    m_pSampleRate->set(44100.);
 
     // Latency control
-    m_pMainLatency = new ControlObject(ConfigKey(kAppGroup, QStringLiteral("output_latency_ms")),
+    m_pOutputLatencyMs = new ControlObject(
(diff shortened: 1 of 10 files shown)


JoergAtGithub:
It would be future proof to call it `output_latency` instead of just `latency`. We currently do not use the input-latency value that portaudio reports, but for more complex multi device sync scenarios it might be needed too.

Holzhaus:
> It would be future proof to call it `output_latency` instead of just `latency`. We currently do not use the input-latency value that portaudio reports, but for more complex multi device sync scenarios it might be needed too.

@daschuer @Swiftb0y opinions?

src/engine/enginemixer.cpp
```diff
@@ -79,9 +79,10 @@ EngineMixer::EngineMixer(
     m_pMainSampleRate->set(44100.);
 
     // Latency control
-    m_pMainLatency = new ControlObject(ConfigKey(group, "latency"),
+    m_pMainLatency = new ControlObject(ConfigKey(kAppGroup, QStringLiteral("latency")),
```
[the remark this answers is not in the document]
daschuer: Or the new name you decide for ...

daschuer (changes requested):

Holzhaus:
I opted for `output_latency_ms`.

src/test/co_dumps/co_dump_inital.csv
```diff
@@ -29,7 +29,7 @@
 [EffectRack1_EffectUnit4_Effect4],parameter4_set_default,0
 [EffectRack1_EffectUnit2_Effect2],parameter2_link_type,0
 [QuickEffectRack1_[Channel3]_Effect1],parameter5_set_zero,0
-[Master],latency,5.80499
+[App],output_latency_ms,5.80499
```
daschuer, line 32: My dump button is broken. Does yours work? I will file a bug.

daschuer (changes requested):
I think the Akai mapping is still broken. The reported output_latency can be long and is unrelated to the internal processing. 
I have found out where the "audio_buffer_size" CO is gone that is probably the correct use here: 
https://github.com/mixxxdj/mixxx/pull/11121/files

But does the script really need that compensation? Why do we not see the use case in other mappings? 

Did you consider how to continue?

src/test/co_dumps/co_dump_inital.csv
Holzhaus: > My dump button is broken. Does yours work? I will file a bug. 

Last time it check it didit creates a file in `~/.mixxx`. It does not create a save file dialog.

daschuer: Oh, right it is working. I just needed to look in my testing .mixxx folder 🤦

Holzhaus:
> I think the Akai mapping is still broken. The reported output_latency can be long and is unrelated to the internal processing. 
> I have found out where the "audio_buffer_size" CO is gone that is probably the correct use here: 
> https://github.com/mixxxdj/mixxx/pull/11121/files
> 
> But does the script really need that compensation? Why do we not see the use case in other mappings? 
> 
> Did you consider how to continue? 

Yes, I considered it and decided that I won't mess with the mapping in this PR. I updated the control, but I don't plan other changes. The bug was not caused by this PR and I don't own this controller, so I cannot properly test changes.

daschuer:
OK, what puzzles me is the question if we need to revive the "audio_buffer_size" for such use cases. The bug in the Akai mapping itself is there or not, probably hardly notable. Do you have there a good idea for a bug report regarding the that?

Holzhaus:
> OK, what puzzles me is the question if we need to revive the "audio_buffer_size" for such use cases.

The use case is "implementing backjumps in JavaScript instead of using the proper Mixxx COs" ?

daschuer:
> The use case is "implementing beatjumps in JavaScript instead of using the proper Mixxx COs" ?

Ok, that is clearly not a valid use case. So there is nothing else to do.  
Thank you for this PR.

daschuer (approved):

daschuer:
We can also ignore the Lint issues. Someone really needs to put some love on this mapping.

PR #12024, opened 2023-09-25
Precompiled headers

Holzhaus:
*Based on #12025.*

No profiling done, I just selected a bunch of headers from the `util` subdirectory plus some other ones that I though might be widely used and don't change that often.

Ideally we'd also add this:

```cmake
target_precompile_headers(mixxx-test REUSE_FROM mixxx-lib)
```

However, this does not work because I think the `libshout-idjc` build messes with the compile options, so the precompiled headers are not reusable and need to be recompiled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1093,7 +1093,6 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/util/logger.cpp
   src/util/logging.cpp
   src/util/mac.cpp
-  src/util/moc_included_test.cpp
   src/util/movinginterquartilemean.cpp
   src/util/performancetimer.cpp
   src/util/rangelist.cpp
@@ -1932,7 +1931,6 @@ add_executable(mixxx-test
   src/test/wbatterytest.cpp
   src/test/wpushbutton_test.cpp
   src/test/wwidgetstack_test.cpp
-  src/util/moc_included_test.cpp
 )
 find_package(GTest CONFIG REQUIRED)
 set_target_properties(mixxx-test PROPERTIES AUTOMOC ON)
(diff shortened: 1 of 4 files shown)


daschuer:
> I think the libshout-idjc build messes with the compile options

Would it be than reasonable to move libshout-idjc to a separate library?

daschuer:
Not sure why the https://github.com/mixxxdj/mixxx/pull/12025 commits are still listed.
Can you rebase it to get rid of them?

CMakeLists.txt
```diff
@@ -1093,7 +1093,6 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/util/logger.cpp
   src/util/logging.cpp
   src/util/mac.cpp
-  src/util/moc_included_test.cpp
```
daschuer, line 1096: Why does this no longer work? Does it error with a false positive?

Holzhaus: It always fails on my machine. If you want to try yourself, just checkout this branch and remove the commit to see if it works for you.

Holzhaus:
> Not sure why the https://github.com/mixxxdj/mixxx/pull/12025 commits are still listed.
> Can you rebase it to get rid of them?  

Done.

JoergAtGithub:
The Windows build failure has it's origin in this name collision:
grafik

Renaming the variable interface fixed the issue. The build of Mixxx takes 1:50 instead of 6 minutes now. And more than 30 seconds of this time is for libdjinterop.

JoergAtGithub:
Tested various build targets locally all work flawless with a massive improvement in build time.
Full build of x64_off is down to 1'20
Full build of x64_legacy is down to 1'55
Code LGTM and CI passes on all platforms. Thank you!

PR #12033, opened 2023-09-26
fix(coreservices): Remove obsolete `getLV2Backend()` method

Holzhaus:
This method is obsolete. It's never used in production code and always returns `nullptr` because the `m_pLV2Backend` member of `CoreServices` is never set.


Holzhaus:
This is pulled from https://github.com/mixxxdj/mixxx/pull/11984 and may be separately merged.

JoergAtGithub:
LGTM! CI passes without it. Thank you!

PR #12037, opened 2023-09-27
Precompiled headers fix

daschuer:
This fixes building on Ubuntu Focal and reintroduces the moc_included_test.cpp


JoergAtGithub:
LGTM! Thank you!
I did not notice, that this test was removed in the original PR.

PR #12041, opened 2023-09-28
Tag Fetcher: fix apply status messages

ronso0:
This removes the message added in #11938 and shifts the cover/cover+tags message to the correct spots.
I.e. no new translation strings except "Selected cover art applied" from #11938


Swiftb0y (approved):
whoops, sorry. Forgot we had string freeze at the beta beginning.

PR #12047, opened 2023-09-30
Denon Prime GO - Mixxx Mapping

ultra-vi-olet:
This PR is being made to allow the Denon Prime GO to be used as a controller for Mixxx when in Computer Mode.

### DIFF
--- a/res/controllers/Denon-Prime-GO.midi.xml
+++ b/res/controllers/Denon-Prime-GO.midi.xml
@@ -967,7 +967,7 @@
             </control>
             <control>
                 <group></group>
-                <key></key>
+                <key>Prime4.leftDeck.padGrid.padModeButtonPressed</key>
                 <description>Deck 1 - Hot Cue Pad On</description>
                 <status>0x92</status>
                 <midino>0x0B</midino>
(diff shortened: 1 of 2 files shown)


dnk8n:
I would be interested in reviving it. Do you think it needs to be specific to Prime Go? Or could the mapping apply to:

- Prime 2/4/4+
- SC Live 4 (what I have to work with)
- Prime Go

Swiftb0y:
The hardware probably differs in subtle ways, but you can take this (or #4027) as a basis and adapt the mapping values in order to fit your specific hardware.

JoergAtGithub:
> I would be interested in reviving this PR. Do you think it needs to be specific to Prime Go?

While it's in general good, to prevent code duplication, this is not the case for controller mappings. This is because of maintenance. If somebody in future change the mapping for e.g. the Prime Go, it's very unlikely, that this person also own the other Prime controllers. And without hardware, we can't verify that the change will not breaks something for the other controller types.

But it would be great to develop all these mappings together and split them, when everything is finished.

dnk8n:
Yeah definitely. Develop them separately. Write common functions that each
type can call with different parameters.

Parameter changes and specific function changes don't need the other
devices to verify.

Only if changing common utils/functions.

Issue #12042, opened 2023-09-29
midi4lights only working on deck 1 


sesoho:
### Bug Description

midi4lights script is only working on deck 1. 

im having isssues with not resvieing any data from deck any other deck than deck one. also the samplers dont send out any data either. 

i didnt realize this bug untill i started going deeper into my set and realized i wasnt sending out any data from any fo the other decks. I would love to know if this is actually a bug or if this is the way the script is supposed to work. in that case i would love to see MIXXX sending out data from all dekcs. and samplers aswell. 

I love mixxx and would lvoe to contribute in any way i can.

KR
Seb

### Version

Mixxx 2.5.0-alpha

### OS

Windows 11 Home

geraldog:
Hi @sesoho do you see anything unusual in the logs?

Remember that when using MIDI 4 Lights you run Mixxx on developer mode so you get the benefit of extra logging on a real shell like PowerShell if you use Windows for example.

In #12048 there was a specific warning I would get when losing beat information, perhaps our problems are related.

geraldog:
Also, specifically, you weren't using Auto DJ and let it automatically fade to the next song, were you? I have more testing to do but with those specific steps, i.e. AutoDJ ON and never touching the crossfader but instead relying on AutoDJ to crossfade, I was able to reproduce your bug.

PR #12048, opened 2023-10-01
MIDI4lights: Give beginTimer callbacks the anonymous function expression form

geraldog:
Merged PR #11953 "fix(controllers): engine.beginTimer(ms, string) would evaluate the code immediately instead of after ms" requires anonymous functions expressions instead of string literals.

This should solve #12042

### DIFF
--- a/res/controllers/Midi_for_light-scripts.js
+++ b/res/controllers/Midi_for_light-scripts.js
@@ -84,14 +84,14 @@ midi_for_light.init = function(id) { // called when the MIDI device is opened & 
 
     engine.connectControl("[Master]", "crossfader", "midi_for_light.crossfaderChange");
 
-    if (enable_vu_meter_global === true) midi_for_light.vu_meter_timer = engine.beginTimer(40, "midi_for_light.vuMeter()");
+    if (enable_vu_meter_global === true) midi_for_light.vu_meter_timer = engine.beginTimer(40, midi_for_light.vuMeter);
 
     // Check midi_channel if value valid. Possible range is 1 to 16.
     if (midi_channel > 16) midi_channel = 16;
     if (midi_channel < 1) midi_channel = 1;
 
     for (var i = 0; i <= 3; i++) {
-        deck_beat_watchdog_timer[i] = engine.beginTimer(beat_watchdog_time, "midi_for_light.deckBeatWatchdog(" + i + ")");
+        deck_beat_watchdog_timer[i] = engine.beginTimer(beat_watchdog_time, function() { midi_for_light.deckBeatWatchdog(i); });
         engine.connectControl("[Channel" + (i + 1) + "]", "beat_active", "midi_for_light.deckBeatOutputToMidi");
         engine.connectControl("[Channel" + (i + 1) + "]", "volume", "midi_for_light.deckVolumeChange");
         engine.connectControl("[Channel" + (i + 1) + "]", "play", "midi_for_light.deckButtonPlay");
@@ -118,7 +118,7 @@ midi_for_light.deckButtonPlay = function(value, group, control) { // called when
     if (value == 1) { // deck play on
         engine.stopTimer(deck_beat_watchdog_timer[deck]);
         beat_watchdog[deck] = false;
-        deck_beat_watchdog_timer[deck] = engine.beginTimer(beat_watchdog_time, "midi_for_light.deckBeatWatchdog(" + deck + ")");
+        deck_beat_watchdog_timer[deck] = engine.beginTimer(beat_watchdog_time, function() { midi_for_light.deckBeatWatchdog(deck); });
     } else { // deck play stop
         engine.stopTimer(deck_beat_watchdog_timer[deck]);
         beat_watchdog[deck] = true;
@@ -416,7 +416,7 @@ midi_for_light.deckVolumeChange = function(value, group, control) { // deck volu
         midi_for_light.deck_current = deckneu;
         midi.sendShortMsg(0x8F + midi_channel, 0x30, 0x64 + deckneu); // Note C on with 64 and add deck
         midi_for_light.volumeBeatBlockStatus = true;
-        midi_for_light.volumeBeatBlock_timer = engine.beginTimer(1000, "midi_for_light.volumeBeatBlock()");
+        midi_for_light.volumeBeatBlock_timer = engine.beginTimer(1000, midi_for_light.volumeBeatBlock);
     }
 };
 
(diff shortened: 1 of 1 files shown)


Swiftb0y:
https://github.com/mixxxdj/mixxx/pull/11953 should've fixed this exact issue already though. If #11953 works correctly, this PR would just be noop. Do you think #11953 is broken?

geraldog:
Hi @Swiftb0y and sorry for bringing you out of your time off.

I tested again and again with real fixture to make sure we understand better the problem and save some face. Still, it seems I need to test even more but let me explain.

I use a modified midi_for_light.vuMeter to interact with AudioPathType::Stem I have created in my fourstems_padj_2023 branch.

Without the proposed fix, here's the warning I get when I engage my Programmable AutoDJ with Stems capabilities:

` controller.midithroughport0:warning [Controller] "passed a string to `engine.beginTimer`, please pass a function instead!"`

And it doesn't work properly. I specifically lose the on-beat information!

geraldog:
@Swiftb0y as not to waste your precious time I reproduced #12042 with Vanilla Mixxx (no weird customizations). I had a hunch the author of #12042 was using AutoDJ too like me, and lo and behold if you don't touch the crossfader but instead rely on AutoDJ to do crossfading to the next song for you, the bug happens and I get the same warning as with the strange customizations.

Swiftb0y:
I appreciate it highly btw that you are so understanding of my personal resource constraints. Thank you.

geraldog:
> I have no doubt that you are experiencing the issue you're describing (I have never looked into the midi4lights script myself so I'm not the right person for this anyways). The thing I don't understand is how this PR is supposed to fix #12042. Since as I said, the behavior should I identical (assuming #11953 works as intended).
> 
> So, are you confident that this PR actually fixes #12042? I'm unfortunately not able to confirm for myself.

No, I'm not confident at all. We need the author of #12042 to test the proposed fix, see if it solves his bug too.

res/controllers/Midi_for_light-scripts.js
```diff
@@ -416,7 +416,7 @@ midi_for_light.deckVolumeChange = function(value, group, control) { // deck volu
         midi_for_light.deck_current = deckneu;
         midi.sendShortMsg(0x8F + midi_channel, 0x30, 0x64 + deckneu); // Note C on with 64 and add deck
         midi_for_light.volumeBeatBlockStatus = true;
-        midi_for_light.volumeBeatBlock_timer = engine.beginTimer(1000, "midi_for_light.volumeBeatBlock()");
+        midi_for_light.volumeBeatBlock_timer = engine.beginTimer(1000, function() { midi_for_light.volumeBeatBlock(); });
```
Holzhaus, line 419: Does this work, too? 

```suggestion
        midi_for_light.volumeBeatBlock_timer = engine.beginTimer(1000, midi_for_light.volumeBeatBlock);
```

res/controllers/Midi_for_light-scripts.js
```diff
@@ -441,7 +441,7 @@ midi_for_light.crossfaderChange = function() { // crossfader chenge, check deck
     engine.stopTimer(midi_for_light.volumebeat_on_delay_timer);
     if (engine.getValue("[Master]", "crossfader") > -0.25) { // crossfader more than 25% left;
         if (engine.getValue("[Master]", "crossfader") < 0.25) { // crossfader more then 25% right;
-            midi_for_light.volumebeat_on_delay_timer = engine.beginTimer(3000, "midi_for_light.volumeBeatOnDelay()");
+            midi_for_light.volumebeat_on_delay_timer = engine.beginTimer(3000, function() { midi_for_light.volumeBeatOnDelay(); });
```
Holzhaus, line 444:
```suggestion
            midi_for_light.volumebeat_on_delay_timer = engine.beginTimer(3000, midi_for_light.volumeBeatOnDelay);
```
?

res/controllers/Midi_for_light-scripts.js
```diff
@@ -463,7 +463,7 @@ midi_for_light.crossfaderChange = function() { // crossfader chenge, check deck
         midi_for_light.deck_current = deck;
         midi.sendShortMsg(0x8F + midi_channel, 0x30, 0x64 + deck); // note C on with value 64 + deck
         midi_for_light.crossfader_block = true;
-        midi_for_light.crossfader_change_block_timer = engine.beginTimer(1000, "midi_for_light.crossfaderChangeBlock()");
+        midi_for_light.crossfader_change_block_timer = engine.beginTimer(1000, function() { midi_for_light.crossfaderChangeBlock(); });
```
Holzhaus, line 466:
```suggestion
        midi_for_light.crossfader_change_block_timer = engine.beginTimer(1000, midi_for_light.crossfaderChangeBlock);
```

Holzhaus (commented):
Thanks. Please check if we can pass the function directly instead of wrapping it in an anonymous function.

geraldog:
Yes Jan it works normally by passing the function directly, except for the functions which need arguments of course, which I kept as anonymous function expressions.

Holzhaus:
> Yes Jan it works normally by passing the function directly, except for the functions which need arguments of course, which I kept as anonymous function expressions.

Good :+1:

I think you didn't push your changes yet though.

Holzhaus (approved):
Code changes look good to make and make sense, even if they don't fix #12042 (the changes should make no difference, but I didn't test).

@geraldog it would be great if you could apply the proposed fixes from eslint to fix the pre-commit CI failure, too.

geraldog:
Hi Jan, thanks for the review by the way, and thank you @ronso0 for updating with the correct title, my bad.

> Code changes look good to make and make sense, even if they don't fix #12042 (the changes should make no difference, but I didn't test).

The changes do make a difference Jan, they don't go through that code path that Niko fixed inside src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp in #11953. Namely:

`int ControllerScriptInterfaceLegacy::beginTimer(
        int intervalMillis, QJSValue timerCallback, bool oneShot) {
    if (timerCallback.isString()) {
        m_pScriptEngineLegacy->logOrThrowError(
                QStringLiteral("passed a string to 'engine.beginTimer', please "
                               "pass a function instead!"));
        // wrap the code in a function to make the evaluation lazy.
        // otherwise the code would be evaluated immediately instead of after
        // the timer which is obviously undesired and could also cause
        // issues when used recursively.
        timerCallback = m_pScriptEngineLegacy->jsEngine()->evaluate(
                QStringLiteral("()=>%1").arg(timerCallback.toString()));
    } else if (!timerCallback.isCallable()) {`

> @geraldog it would be great if you could apply the proposed fixes from eslint to fix the pre-commit CI failure, too.

I will look into it Jan, and I don't want to be rude that's why I held back on this comment I'm about to make.

Niko said the script needs more love, but I consider this an understatement. Quite frankly that script looks like one of my hacks, amazing but kind-of constructed in a rush, if you know what I mean.

Swiftb0y (approved):
welp, lgtm, eslint failure is expected. Still very likely _won't_ solve https://github.com/mixxxdj/mixxx/issues/12042

geraldog:
Thanks @Swiftb0y

PR #12051, opened 2023-10-01
WKey: Fix regression in key notation change

### LINKED ISSUE #12044: Inconsistent/wrong musical keys in the UI
### Bug Description

Switching the key notation mode (e.g. Traditional <-> Lancelot/Traditional) causes the key to change, additionally jumping down/up a semitone doesn't always land on the starting key, especially after changing the notation mode. See the video for a short demo:

https://github.com/mixxxdj/mixxx/assets/30873659/e0c77d6d-9b9c-456d-b5f6-2078a916bb84

Anyone's got an idea what the issue could be?

### Version

2.5-alpha-114-g75ac186cfa

### OS

macOS 14.0

fwcd:
### Fixes #12044 

This fixes an issue where the _key notation index_ would incorrectly be interpreted as the _value_ of the key itself. This would often result in an incorrect key being displayed after switching the key notation:

https://github.com/mixxxdj/mixxx/assets/30873659/60bab9f1-0174-4435-8de1-73c6c9a0817b

See the issue for more details. The fix was to simply revert bb6db98fcff0293e927f7338f0d9324c7909bd04. I've also added a short explanatory comment to avoid confusion in the future.

cc @daschuer

### DIFF
--- a/src/widget/wkey.cpp
+++ b/src/widget/wkey.cpp
@@ -62,6 +62,8 @@ void WKey::setCents() {
     setValue(m_dOldValue);
 }
 
-void WKey::keyNotationChanged(double dValue) {
+void WKey::keyNotationChanged(double dKeyNotationValue) {
+    // NOTE: dKeyNotationValue is the index of the key notation type, NOT the
+    // key itself, so we intentionally set the old value again to update the UI.
     setValue(m_dOldValue);
 }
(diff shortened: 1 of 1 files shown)


daschuer (approved):
Thank you for the fix. LGTM.

PR #12058, opened 2023-10-04
Reuse from precompiled headers for mixxx-test 

daschuer:
Now it is working. We need to change some definition to PRIVATE or PUBLIC. 
In case of libshout-idjc the dependent headers are no longer part of the public interface. I guess that must have changed in a version. 

To use the same headers in mixxx-test like in mixxx-lib all libraries of mixxx-lib needs to become public that all definition are inherited by mixxx-test and it uses that way the same definitions. This has the benefit that we are sure the mixxx-test is compiled with the same flags as the mixxx-lib. 

An alternative solution would be to maintain public and private headers in mixxx-lib. A maintenance burden I would like to avoid.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1516,7 +1516,7 @@ endif()
 
 # QML Debugging
 if(CMAKE_BUILD_TYPE STREQUAL "Debug")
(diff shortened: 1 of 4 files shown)


JoergAtGithub:
Works locally on Windows 11. CI passed. Code LGTM! Thank you!

PR #12060, opened 2023-10-04
privat generated ui headers

daschuer:
This makes the ui_dlgpreferencesdlg.h private by moving the instantiation of the DlgPreferences to CoreServices.

This fixes one issue from https://github.com/mixxxdj/mixxx/pull/11407

### DIFF
--- a/src/coreservices.cpp
+++ b/src/coreservices.cpp
@@ -22,6 +22,7 @@
 #include "mixer/playerinfo.h"
 #include "mixer/playermanager.h"
 #include "moc_coreservices.cpp"
+#include "preferences/dialog/dlgpreferences.h"
 #include "preferences/settingsmanager.h"
 #ifdef __MODPLUG__
 #include "preferences/dialog/dlgprefmodplug.h"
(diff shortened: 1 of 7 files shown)


daschuer:
@acolombier Does this work for you? 
Not sure why the code coverage drops. I think we can ignore that.

acolombier:
I'm currently away from the computer but I will test it as soon as I get back to it. Alternatively, I have managed to also remove that work around by using the forward declaration in case this change would have repercussions on other things.

acolombier:
It looks like that did the trick, thanks!

src/qml/qmlapplication.cpp
```diff
@@ -5,6 +5,7 @@
 #include "control/controlsortfiltermodel.h"
 #include "controllers/controllermanager.h"
 #include "moc_qmlapplication.cpp"
+#include "preferences/dialog/dlgpreferences.h"
```
acolombier, line 8: This line seems to cause issue since it means that the QML library (in my feature, I have splitted `mixxx-lib` in two) will also need `preferences/dialog/ui_dlgpreferencesdlg.h`. I belive however that I could take your change in, plus the one I've made to initialize the QML signletons as part of the CoreService instead of as part of the QML application, which then removes the need for `DlgPreferences` in this file

acolombier: I believe this line can actually be removed.

daschuer:
Ready for merge, the Coveralls failure seems to be unrelated.

acolombier (approved):

src/coreservices.cpp
```diff
@@ -527,6 +528,19 @@ bool CoreServices::initializeDatabase() {
     return MixxxDb::initDatabaseSchema(dbConnection);
 }
 
+std::shared_ptr<QDialog> CoreServices::makeDlgPreferences() const {
+    std::shared_ptr<DlgPreferences> pDlgPreferences = std::make_shared<DlgPreferences>(
```
Swiftb0y, line 532: whats the reason for the type-erasure? I think this needs a comment.

daschuer:
Done

JoergAtGithub:
Code LGTM! All review comments are addressed and I tested it as part of #12139 on Win11. Thank you!

PR #12061, opened 2023-10-05
SearchQueryParser: Add "OR" search operator

### LINKED ISSUE #8881: make a way to combine search query nodes with an OR relationship
Date: 2017-06-14T19:07:14Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1697983](https://bugs.launchpad.net/bugs/1697983)

---

Currently, every search term typed into the library query field is combined with AND relationships. It would be nice if there was a way to combine nodes with an OR relationship too. For example:
`(artist:foo | artist:bar)`
would match tracks that have an artist field matching either "foo" or "bar".

fwcd:
### Fixes #8881

This adds `|` and (uppercase) `OR` as syntax for the OR operator in search queries. Thus queries such as `house | disco` or equivalently `house OR disco` will now yield the combined results from querying `house` and `disco`.

Both `|` and `OR` have strong precedent, e.g. with Google using these operators[^1]. Especially `|` is terse, easy to read/write and familiar to technical users.

[^1]: https://ahrefs.com/blog/google-advanced-search-operators

### DIFF
--- a/src/library/searchqueryparser.cpp
+++ b/src/library/searchqueryparser.cpp
@@ -266,7 +266,7 @@ void SearchQueryParser::parseTokens(QStringList tokens,
 }
 
 std::unique_ptr<AndNode> SearchQueryParser::parseAndNode(const QString& query) const {
-    auto pQuery(std::make_unique<AndNode>());
+    auto pQuery = std::make_unique<AndNode>();
 
     QStringList tokens = query.split(" ");
     parseTokens(std::move(tokens), pQuery.get());
@@ -275,7 +275,7 @@ std::unique_ptr<AndNode> SearchQueryParser::parseAndNode(const QString& query) c
 }
 
 std::unique_ptr<OrNode> SearchQueryParser::parseOrNode(const QString& query) const {
-    auto pQuery(std::make_unique<OrNode>());
+    auto pQuery = std::make_unique<OrNode>();
 
     QStringList rawAndNodes = query.split("|");
     for (const QString& rawAndNode : rawAndNodes) {
--- a/src/library/searchqueryparser.cpp
+++ b/src/library/searchqueryparser.cpp
@@ -8,10 +8,15 @@
 
 constexpr char kNegatePrefix[] = "-";
 constexpr char kFuzzyPrefix[] = "~";
-constexpr char kOrOperator[] = "|";
+
 // see https://stackoverflow.com/questions/1310473/regex-matching-spaces-but-not-in-strings
+#define STRING_QUOTE_SUFFIX "(?=[^\"]*(\"[^\"]*\"[^\"]*)*$)"
+
 const QRegularExpression kSplitIntoWordsRegexp = QRegularExpression(
-        QStringLiteral(" (?=[^\"]*(\"[^\"]*\"[^\"]*)*$)"));
+        QStringLiteral(" " STRING_QUOTE_SUFFIX));
+
+const QRegularExpression kSplitOnOrOperatorRegexp = QRegularExpression(
+        QStringLiteral("\\|" STRING_QUOTE_SUFFIX));
 
 SearchQueryParser::SearchQueryParser(TrackCollection* pTrackCollection, QStringList searchColumns)
         : m_pTrackCollection(pTrackCollection),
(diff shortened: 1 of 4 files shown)


src/library/searchqueryparser.cpp
```diff
@@ -263,6 +265,28 @@ void SearchQueryParser::parseTokens(QStringList tokens,
     }
 }
 
+std::unique_ptr<AndNode> SearchQueryParser::parseAndNode(const QString& query) const {
+    auto pQuery(std::make_unique<AndNode>());
```
daschuer, line 269: we prefer here the = initializer
```suggestion
    auto pQuery = std::make_unique<AndNode>();
```

src/library/searchqueryparser.cpp
```diff
@@ -263,6 +265,28 @@ void SearchQueryParser::parseTokens(QStringList tokens,
...
+
+    QStringList tokens = query.split(" ");
+    parseTokens(std::move(tokens), pQuery.get());
+
+    return pQuery;
+}
+
+std::unique_ptr<OrNode> SearchQueryParser::parseOrNode(const QString& query) const {
+    auto pQuery(std::make_unique<OrNode>());
```
daschuer, line 278:
```suggestion
    auto pQuery = std::make_unique<OrNode>();
```

daschuer (changes requested):
Thank you. Can you also ass a test for this?

fwcd:
I've updated the parser to handle quotes correctly (i.e. `title:"some | title" | house`, for example, will now behave as expected) and added some tests to verify that this works correctly.

fwcd:
I don't understand why the tests are failing in CI, they work on my machine™. Off all the tests, `SearchQueryParserTest.EmptyOrOperator` seems to time out, whose implementation is [rather trivial](https://github.com/mixxxdj/mixxx/pull/12061/files#diff-8d27f20b0180e546defc6e60fc0e80190d0457845c1390be6f92983dcae9d71cR1041-R1047). Any ideas?

fwcd:
Couldn't reproduce the test failure in an Arch Linux Docker container either (though admittedly on arm64 too).

fwcd:
Ah, I see, CI ran into a debug assert which I didn't get since I was building in `RelWithDebInfo` mode:

```
critical [0x55a1d4186d30] DEBUG ASSERT: "!m_nodes.empty()" in function virtual bool OrNode::match(const TrackPointer&) const at /home/runner/work/mixxx/mixxx/src/library/searchquery.cpp:127
```

https://github.com/mixxxdj/mixxx/blob/9e411ff323d96829de9d586859823c8b902e052d/src/library/searchquery.cpp#L125-L132

Might be an interesting question how we should handle this. Apart from the top-level `AndNode`, empty queries will simply not produce a node at all and queries such as `|` produce an (empty) `OrNode`. IMO an empty `OrNode`, even if it's a bit odd, shouldn't be a semantic error and simply not match anything. Could we update the SQL generation to reflect that?

daschuer:
I think we should consider the intermediate results when typing. I can imagine that showing nothing because of a semantic error feels bad. Is this related here? I have not tested. What happen if you type "ABBA |" ?

The other issue is that we should only assert impossible conditions. Sincea user can trigger it, it must not be an assertion.

fwcd:
> What happen if you type "ABBA |"?

Empty children are currently filtered out, so this would be equivalent to simply typing `ABBA`:

https://github.com/fwcd/mixxx/blob/93b96bcc593543337f05b9921072c89d1f1f53a8/src/library/searchqueryparser.cpp#L292-L296

The only real edge case where the user would end up with an empty `OrNode` would e.g. be `|`, `| |` etc. Probably a contrived example, but we wouldn't want to hit this debug assert at all, ideally.

I'd be fine with specifying the empty OR as meaning "match everything", even if it deviates from the common convention to define it as "match nothing" if that makes SQL generation more convenient and/or lines up with what the user expects (i.e. still showing all tracks, when the query is `|` and thus empty-ish), but we should remove this debug assert either way.

Wdyt?

JoergAtGithub:
> we could even support both `|` and `OR`, which is what Google does

I think we should support both!

fwcd:
@JoergAtGithub I've had a go at that, but parsing it seems to be less trivial than I would've hoped for, particularly the tests around lowercase `or` still fail: 3e702054a6

fwcd:
Both `|` and `OR` are now supported and I've added a bunch of tests to make sure that especially the latter is handled correctly with all of its edge cases (e.g. different casing or adjacent letters shouldn't cause it to be parsed as an operator).

Additionally, I have implemented the empty-OR-as-false interpretation, which essentially amounted to removing the debug assert and generating `FALSE` as SQL code if there are no child nodes. As explained in https://github.com/mixxxdj/mixxx/pull/12061#issuecomment-1749352944 this is the logically sound way to handle this and, in practice, only affects edge cases anyway (so unless the user e.g. searches `|`, i.e. the empty OR operator, the results table isn't necessarily empty).

daschuer (approved):
This works like a charm now and the code LGTM as well. 
Special thanks for the extra efforts in testing.

Be-ing:
Thanks for implementing my 6 year old feature request.

ronso0:
I'm wondering: i don't spot any changes to user-facing strings (tr), what was the reason to target 2.5 instead of 2.4?
Do you think it would be worth backporting it to 2.4?

fwcd:
My general philosophy was to target `main` for any non-trivial new features so we can focus on improving stability in 2.4 while giving these changes some time for testing. Since this isn't a _huge_ change, I don't have strong feelings here, but it does change the parsing logic of search queries a bit.

PR #12063, opened 2023-10-05
Track search: use := and quotes to find exact matches

### LINKED ISSUE #10699: Search: allow to find exact matches, like title:="Nice"
Reported by: **[geozubuntu](https://launchpad.net/~geozubuntu)**
Date: 2022-04-03T12:14:12Z
Status: Confirmed
Importance: Low
Launchpad Issue: [lp1967667](https://bugs.launchpad.net/bugs/1967667)

---

OS Linux Mint 20.3 Una (based on Focal Fossa) x64
Mixxx 2.4-alpha-1132-g6de81c8994 (main) with a very big library.

Try to locate a single word track title in library using the search function and will be disappointed.

I would like to find all versions of a song with the title Evil in my library. Not every track containing the word evil in its title but only those which their title is EVIL. I found out that it is almost impossible with the current implementation of search function of Mixxx. 

I tried to write in search field the following:

𝐄𝐯𝐢𝐥  returned thousands of titles containing the 𝐬𝐭𝐫𝐢𝐧𝐠 ..evil.. anywhere i.e Evil's or Devil or Neville or Sevilla, and the 𝐰𝐨𝐫𝐝 Evil like ..the Evil prince.. which is normal and expected. 

"𝐄𝐯𝐢𝐥"  returned the same as above (even if I am not 100% sure since there isn't a counter to know how many results each search returned, (I will submit another (bug) wish for this) but also thousands of titles containing the 𝐬𝐭𝐫𝐢𝐧𝐠 ..evil.. anywhere (I suppose it isn't normal since I used quotes and mixxx should at least exclude words as Devil or Sevilla returning ..Evil eye.. or .. Evil woman .. or any title containing the 𝐰𝐨𝐫𝐝 Evil)

" 𝐄𝐯𝐢𝐥" (leading space), "𝐄𝐯𝐢𝐥 " (trailing space), " 𝐄𝐯𝐢𝐥 " (leading and trailing spaces) much less results but again hundreds, if not thousands of tracks containing this string anywhere. I can't tell if there is a difference (no counter to say how many).

𝐭𝐢𝐭𝐥𝐞:𝐄𝐯𝐢𝐥 and 𝐭𝐢𝐭𝐥𝐞:"𝐄𝐯𝐢𝐥" returned much less results since it looks only in title column but also everything containing the 𝐬𝐭𝐫𝐢𝐧𝐠 ..evil.. i.e Devil, evil woman, evil ways, Sevilla etc. etc.  hundreds of tracks, no counter to say which one returned more. 

𝐭𝐢𝐭𝐥𝐞:"𝐄𝐯𝐢𝐥 and 𝐭𝐢𝐭𝐥𝐞:𝐄𝐯𝐢𝐥" returned nothing at all which was expected (probably invalid filter since quotes wasn't closed in both cases)

I couldn't find any combination to have only the tracks with the title "Evil" in the library view.🙁

So I was unable to find the tracks with the title "Evil". I figured out that I have 8 tracks with this single word title Evil after sorting the title by clicking on the title and then scroll down to letter E so they were all sorted together. But it is a workaround in my opinion. 

So a revision of the search function should be done to address this situation. 
Think that you like to find all covers of a song for a show or a podcast... 

Thanks in advance

ronso0:
A quick check that has proven to work:
`[text property]:="string"` finds only tracks where the property is exactly `string`. 
`title:="Blob"` finds all track titles that exactly "Blob", not "What a Blob" etc.

May be extended to include spaces and special chars before and after to find ` Blob` and `..Blob!`

Bug fix or feature? Doesn't change the ui, hence I targeted 2.4
But it reuires a manual update.

Closes #10699

### DIFF
--- a/src/library/searchquery.cpp
+++ b/src/library/searchquery.cpp
@@ -168,10 +168,12 @@ QString NotNode::toSql() const {
 
 TextFilterNode::TextFilterNode(const QSqlDatabase& database,
         const QStringList& sqlColumns,
-        const QString& argument)
+        const QString& argument,
+        const StringMatch matchMode)
         : m_database(database),
           m_sqlColumns(sqlColumns),
-          m_argument(argument) {
+          m_argument(argument),
+          m_matchMode(matchMode) {
     mixxx::DbConnection::makeStringLatinLow(&m_argument);
 }
 
@@ -184,8 +186,14 @@ bool TextFilterNode::match(const TrackPointer& pTrack) const {
 
         QString strValue = value.toString();
         mixxx::DbConnection::makeStringLatinLow(&strValue);
-        if (strValue.contains(m_argument)) {
-            return true;
+        if (m_matchMode == StringMatch::Equals) {
+            if (strValue == m_argument) {
+                return true;
+            }
+        } else {
+            if (strValue.contains(m_argument)) {
+                return true;
+            }
         }
     }
     return false;
@@ -201,8 +209,17 @@ QString TextFilterNode::toSql() const {
             argument.append('_');
         }
     }
-    QString escapedArgument = escaper.escapeString(
-            kSqlLikeMatchAll + argument + kSqlLikeMatchAll);
+    QString escapedArgument;
+    // Using a switch-case without default case to get a compile-time -Wswitch warning
+    switch (m_matchMode) {
+    case StringMatch::Contains:
+        escapedArgument = escaper.escapeString(
+                kSqlLikeMatchAll + argument + kSqlLikeMatchAll);
+        break;
+    case StringMatch::Equals:
+        escapedArgument = escaper.escapeString(argument);
+        break;
+    }
     QStringList searchClauses;
     for (const auto& sqlColumn : m_sqlColumns) {
         searchClauses << QString("%1 LIKE %2").arg(sqlColumn, escapedArgument);
--- a/src/library/searchqueryparser.cpp
+++ b/src/library/searchqueryparser.cpp
@@ -1,8 +1,57 @@
 #include "library/searchqueryparser.h"
 
 #include <QRegularExpression>
+#include <utility>
 
+#include "library/searchquery.h"
 #include "track/keyutils.h"
+#include "util/assert.h"
+
+namespace {
+
+enum class Quoted : bool {
+    Incomplete,
+    Complete,
+};
+
+std::pair<QString, Quoted> consumeQuotedArgument(QString argument,
+        QStringList* tokens) {
+    DEBUG_ASSERT(argument.startsWith("\""));
+
+    argument = argument.mid(1);
+
+    int quote_index = argument.indexOf("\"");
+    while (quote_index == -1 && tokens->length() > 0) {
+        argument += " " + tokens->takeFirst();
+        quote_index = argument.indexOf("\"");
+    }
+
+    if (quote_index == -1) {
+        // No ending quote found. Since we think they are going to close the
+        // quote eventually, treat the entire token list as the argument for
+        // now.
+        return {argument, Quoted::Incomplete};
+    }
+
+    // Stuff the rest of the argument after the quote back into tokens.
+    QString remaining = argument.mid(quote_index + 1).trimmed();
+    if (remaining.size() != 0) {
+        tokens->push_front(remaining);
+    }
+
+    if (quote_index == 0) {
+        // We have found an explicit empty string ""
+        // return it as "" to distinguish it from an unfinished empty string
+        argument = kMissingFieldSearchTerm;
+    } else {
+        // Found a closing quote.
+        // Slice off the quote and everything after.
+        argument = argument.left(quote_index);
+    }
+    return {argument, Quoted::Complete};
+}
+
+} // anonymous namespace
 
 constexpr char kNegatePrefix[] = "-";
 constexpr char kFuzzyPrefix[] = "~";
@@ -84,12 +133,8 @@ void SearchQueryParser::setSearchColumns(QStringList searchColumns) {
     }
 }
 
-QString SearchQueryParser::getTextArgument(QString argument,
-        QStringList* tokens,
-        StringMatch* matchMode) const {
-    if (matchMode != nullptr) {
-        *matchMode = StringMatch::Contains;
-    }
+SearchQueryParser::TextArgumentResult SearchQueryParser::getTextArgument(QString argument,
+        QStringList* tokens) const {
     // If the argument is empty, assume the user placed a space after an
     // advanced search command. Consume another token and treat that as the
     // argument.
@@ -99,51 +144,20 @@ QString SearchQueryParser::getTextArgument(QString argument,
             argument = tokens->takeFirst();
         }
     }
-
-    bool shouldMatchExactly = false;
+    StringMatch mode = StringMatch::Contains;
     if (argument.startsWith("=")) {
+        // strip the '=' from the argument
         argument = argument.mid(1);
-        shouldMatchExactly = true;
+        mode = StringMatch::Equals;
     }
-    // Deal with quoted arguments. If this token started with a quote, then
-    // search for the closing quote.
     if (argument.startsWith("\"")) {
-        argument = argument.mid(1);
-
-        int quote_index = argument.indexOf("\"");
-        while (quote_index == -1 && tokens->length() > 0) {
-            argument += " " + tokens->takeFirst();
-            quote_index = argument.indexOf("\"");
-        }
-
-        if (quote_index == -1) {
-            // No ending quote found. Since we think they are going to close the
-            // quote eventually, treat the entire token list as the argument for
-            // now.
-            return argument;
-        }
-
-        // Stuff the rest of the argument after the quote back into tokens.
-        QString remaining = argument.mid(quote_index+1).trimmed();
-        if (remaining.size() != 0) {
-            tokens->push_front(remaining);
-        }
-
-        if (quote_index == 0) {
-            // We have found an explicit empty string ""
-            // return it as "" to distinguish it from an unfinished empty string
-            argument = kMissingFieldSearchTerm;
-        } else {
-            // Found a closing quote.
-            // Slice off the quote and everything after.
-            argument = argument.left(quote_index);
-            if (matchMode != nullptr && shouldMatchExactly) {
-                *matchMode = StringMatch::Equals;
-            }
-        }
+        Quoted quoted;
+        std::tie(argument, quoted) = consumeQuotedArgument(argument, tokens);
+        mode = quoted == Quoted::Complete && mode == StringMatch::Equals
+                ? StringMatch::Equals
+                : StringMatch::Contains;
     }
-
-    return argument;
+    return {argument, mode};
 }
 
 void SearchQueryParser::parseTokens(QStringList tokens,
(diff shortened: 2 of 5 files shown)


src/library/searchquery.cpp
```diff
@@ -168,10 +168,12 @@ QString NotNode::toSql() const {
...
 TextFilterNode::TextFilterNode(const QSqlDatabase& database,
         const QStringList& sqlColumns,
-        const QString& argument)
+        const QString& argument,
+        bool exactMatch)
         : m_database(database),
           m_sqlColumns(sqlColumns),
-          m_argument(argument) {
+          m_argument(argument),
+          m_exactMatch(exactMatch) {
```
uklotzde, line 176: `bool` is a poor type choice. An `enum class StringMatch` with `Equals` and `Contains` variants would be much more concise and maintainable.

ronso0: Thanks for the hint.
Though this is on hold for now, someone else may adopt it.

ronso0: Done.
And I implemented it in the existing text filter matcher, so the changes are much smaller, no code duplication anymore.

src/test/searchqueryparsertest.cpp
```diff
@@ -178,6 +178,32 @@ TEST_F(SearchQueryParserTest, TextFilter) {
...
+
+    TrackPointer pTrack(Track::newTemporary());
+    pTrack->setComment("test ASDF test");
+    EXPECT_FALSE(pQuery->match(pTrack));
+    pTrack->setComment("ASDF");
+    EXPECT_TRUE(pQuery->match(pTrack));
+
+    EXPECT_STREQ(
+            qPrintable(QString("comment LIKE 'asdf'")),
+            qPrintable(pQuery->toSql()));
```
Swiftb0y, line 193: this tests implementation details IMO and does not belong in a test IMO. wdyt?

src/library/searchquery.cpp
```diff
@@ -168,10 +168,13 @@ QString NotNode::toSql() const {
...
         const QStringList& sqlColumns,
-        const QString& argument)
+        const QString& argument,
+        const StringMatch matchMode)
         : m_database(database),
           m_sqlColumns(sqlColumns),
-          m_argument(argument) {
+          m_argument(argument),
+          m_matchMode(matchMode) {
+    qRegisterMetaType<StringMatch>("StringMatch");
```
Swiftb0y, line 177: `StringMatch` doesn't need to be a metatype afai can tell.
```suggestion
```
same unnecessary registration in `SearchQueryParser` ctor.

src/library/searchqueryparser.cpp
```diff
@@ -88,7 +90,8 @@ void SearchQueryParser::setSearchColumns(QStringList searchColumns) {
 }
 
 QString SearchQueryParser::getTextArgument(QString argument,
-                                           QStringList* tokens) const {
+        QStringList* tokens,
+        StringMatch* matchMode) const {
```
Swiftb0y, line 94: I dislike the implicitness of the out parameter. Eg its not fully transparent if the pointer is valid but `shouldMatchExactly` is `false`. then it'll just not set anything. A caller might expect it to always set it. The result is that if used incorrectly (which is not hard in this case), the caller could be left with an initialized variable. 

Imo the better design would be not use an out-parameter in this case and instead just return a struct with both the matchmode and the argument. 

```suggestion
struct TextArgumentResult {
    QString argumentValue;
    StringMatch matchmode;
};
TextArgumentResult SearchQueryParser::getTextArgument(QString argument,
                                           QStringList* tokens) const {
```

src/library/searchquery.cpp
```diff
@@ -201,8 +210,13 @@ QString TextFilterNode::toSql() const {
...
     }
-    QString escapedArgument = escaper.escapeString(
-            kSqlLikeMatchAll + argument + kSqlLikeMatchAll);
+    QString escapedArgument;
+    if (m_matchMode == StringMatch::Equals) {
+        escapedArgument = escaper.escapeString(argument);
+    } else {
+        escapedArgument = escaper.escapeString(
+                kSqlLikeMatchAll + argument + kSqlLikeMatchAll);
+    }
```
Swiftb0y, line 219: Less brittle and more expressive (imo) solution:
```suggestion
    QString escapedArgument;
    // use a switch-case without default case so get a compile-time -Wswitch warning
    switch (m_matchMode) {
    case StringMatch::Contains:
        escapedArgument = escaper.escapeString(
                kSqlLikeMatchAll + argument + kSqlLikeMatchAll);
    case StringMatch::Equals:
        escapedArgument = escaper.escapeString(argument);
    }
```

Swiftb0y (commented):
couldn't resist "meinen senf abzugeben" ;)

Thank you.

src/test/searchqueryparsertest.cpp
ronso0: These are crucial implementation details, sure they need to be tested.
Please elaborate.

src/library/searchquery.cpp
ronso0: Yes, I can build just fine without registration on Linux, but I remember this was(?) necessary for some OS/compiler when I added `enum class FocusWidget` (CI failed to built without it).

src/library/searchqueryparser.cpp
ronso0: > Eg its not fully transparent if the pointer is valid but shouldMatchExactly is false. then it'll just not set anything.

For now, I fixed that by setting matchMode at the beginning.

A struct would for me, too, but it requires more refactoring.

src/test/searchqueryparsertest.cpp
Swiftb0y: well it conflates two different levels of abstraction. The first part of the test checks for the behavior. That should be very stable and even if we changed some details about our database these won't change. On the other hand, all the tests that involve `toSql` don't only rely on the fact that we use SQL (something that would not be true if we were to transition to aoide) but even how exactly we do the fuzzy matching (using `LIKE`) even though there are other implementations as well. Do you see how this introduces coupling? The test should be ensuring the behavior, not the way it works internally. 
A little (admittedly childish and simplified) analogy: If I need to put a nail into a wall, I can just use a hammer, I don't care if the hammerhead is made of iron or steel, both do the job (I think, I'm not a carpenter :sweat_smile:).
Sure if you need to test specific properties of a hammer, that is legitimate to test, but it doesn't belong in the same test as "it puts a nail into a wall".

I'm aware that this is quite theoretical and likely a little controversial the code in question just follows the same pattern as the rest of the file. I'm just saying that the current design is not great and we should question it.

src/library/searchquery.cpp
Swiftb0y: well, in this case, I'm pretty sure it shouldn't

src/library/searchqueryparser.cpp
Swiftb0y: not so much imo. PR underway

Swiftb0y: https://github.com/ronso0/mixxx/pull/47

src/test/searchqueryparsertest.cpp
ronso0: got it. of course, ideally, we'd just test the parser to return certain track ids of a sample library.

src/library/searchquery.cpp
ronso0: I'll push a test commit. If it's not needed (anymore?), great.

ronso0:
Looking good now.
@Swiftb0y Please let me know if you're happy with the fixup.

src/library/searchqueryparser.cpp
```diff
@@ -148,15 +148,20 @@ SearchQueryParser::TextArgumentResult SearchQueryParser::getTextArgument(QString
         }
     }
     StringMatch mode = StringMatch::Contains;
+    bool equalRequested = false;
     if (argument.startsWith("=")) {
         // strip the '=' from the argument
         argument = argument.mid(1);
+        // TODO(ronso0) should 'tag:"string"' really be equal to 'tag:="string"?
```
ronso0, line 155: edit
`tag="string"` and `tag:="string"`

src/library/searchqueryparser.cpp
```diff
@@ -98,42 +144,23 @@ QString SearchQueryParser::getTextArgument(QString argument,
...
+        equalRequested = true;
     }
-
-    return argument;
+    if (argument.startsWith("\"")) {
+        Quoted quoted;
+        std::tie(argument, quoted) = consumeQuotedArgument(argument, tokens);
+        mode = equalRequested && quoted == Quoted::Complete
+                ? StringMatch::Equals
+                : StringMatch::Contains;
```
Swiftb0y, line 161: `equalRequested` is kind off unnecessary...
```suggestion
        mode = mode == StringMatch::Equals && quoted == Quoted::Complete
                ? StringMatch::Equals
                : StringMatch::Contains;
```

src/library/searchqueryparser.cpp
```diff
@@ -98,42 +144,23 @@ QString SearchQueryParser::getTextArgument(QString argument,
...
-
-        if (quote_index == 0) {
-            // We have found an explicit empty string ""
-            // return it as "" to distinguish it from an unfinished empty string
-            argument = kMissingFieldSearchTerm;
-        } else {
-            // Slice off the quote and everything after.
-            argument = argument.left(quote_index);
-        }
+        // TODO(ronso0) should 'tag:=string' really be equal to 'tag:="string"?
```
Swiftb0y, line 152: I'd argue yes, because `string` is just a single word. so the quotes are unnecessary and thus these should be equal imo.

Swiftb0y:
lgtm otherwise. thank you.

src/library/searchqueryparser.cpp
ronso0: True, makes sense. And it's consistent with `bpm:=...` which takes also a single value.

src/library/searchqueryparser.cpp
```diff
@@ -98,42 +144,20 @@ QString SearchQueryParser::getTextArgument(QString argument,
...
-            argument = argument.left(quote_index);
-        }
+        mode = StringMatch::Equals;
     }
-
-    return argument;
+    if (argument.startsWith("\"")) {
+        Quoted quoted;
+        std::tie(argument, quoted) = consumeQuotedArgument(argument, tokens);
+        mode = quoted == Quoted::Complete
```
Swiftb0y, line 156: this should fix the tests again, right?
```suggestion
        mode = quoted == Quoted::Complete && mode == StringMatch::Equals
```

Swiftb0y:
fyi the parser could like be simplified even more if the tokenizer recognized the quotes and just consumed the entire thing as one token. That would make `consumeQuotedArgument` largely unnecessary...

src/library/searchqueryparser.cpp
ronso0: yes :roll_eyes:

ronso0:
> fyi the parser could like be simplified even more if the tokenizer recognized the quotes and just consumed the entire thing as one token. That would make `consumeQuotedArgument` largely unnecessary...

Tbh I'm not motivated to do more optimization, this has already grown way beyond my initial commit (though I very much appreciate your input!).

Swiftb0y:
> Tbh I'm not motivated to do more optimization, this has already grown way beyond my initial commit (though I very much appreciate your input!).

Yes, I'm sorry for blowing up the scope. The previous comment was just a remark and not a request for more changes.

Swiftb0y (approved):
LGTM, Thank you very much.

Swiftb0y:
remove draft and merge?

ronso0:
Ready!

**TODO**
update the manual

Swiftb0y:
Thank you

ronso0:
I'd say this deserves a changelog entry to make the change a bit more discoverable for upgraders (if they already read the search documentation they probably won't notice the change..).
Wdyt?

Issue #12065, opened 2023-10-05
RTE when engine.setValue("[Library]", "MoveVertical", delta) is called during startup


JoergAtGithub:
### Bug Description

The line `engine.setValue("[Library]", "MoveVertical", delta);` in my Traktor Z2 mapping results in a runtime exception in recent versions of Mixxx 2.4 - the unmodified mapping worked some time ago:
![MixxxStartUpExecption](https://github.com/mixxxdj/mixxx/assets/64457745/ca626933-8e4b-4187-829a-7f0c3a4913d4)
The same issue was also reported here from a Ubuntu user: https://github.com/mixxxdj/mixxx/pull/3038#issuecomment-1712587674

Reloading the mapping after startup does not result in the RTE

### Version

2.4.0

### OS

Windows 11 & Ubuntu

JoergAtGithub:
This is a regression introduced by:
https://github.com/mixxxdj/mixxx/commit/19a1bdc8639c7bf2dd5d2019bcbe925949e29d55
Reverting this change prevents the RTE.

ronso0:
Could you please share a link to your Z2 script?

JoergAtGithub:
It's this line, which triggers the RTE, if I comment it out, Mixxx starts up without RTE:
https://github.com/mixxxdj/mixxx/blob/34662d1aad43a0d002dfad0dc543fcf21e8b2a1a/res/controllers/Traktor-Kontrol-Z2-hid-scripts.js#L1036

JoergAtGithub:
The RTE happens always, without touching the knob. The value of the knob encoder is read during initialisation, this triggers an event. I will clean up this in the mapping as well.

PR #12066, opened 2023-10-05
Fix library focus RTE

JoergAtGithub:
Closes:  #12065

### DIFF
--- a/src/widget/wlibrarytableview.cpp
+++ b/src/widget/wlibrarytableview.cpp
@@ -274,7 +274,7 @@ void WLibraryTableView::focusInEvent(QFocusEvent* event) {
         // This is especially helpful if the table has only one track, which can
         // not be selected with up/down buttons, either physical or emulated via
         // [Library],MoveVertical controls. See #9548
-        if (model()->rowCount() > 0) {
+        if (model() && model()->rowCount() > 0) {
             if (selectionModel()->hasSelection()) {
                 DEBUG_ASSERT(!selectionModel()->selectedIndexes().isEmpty());
                 if (!currentIndex().isValid() ||


Holzhaus (approved):
Thanks, code looks good to me, but I'd appreciate confirmation from @ronso0.

PS: This should really use the early exit technique instead of nesting 4 `if` statements. But this is just a bugfix, so it's out of scope here.

src/widget/wlibrarytableview.cpp
```diff
@@ -274,7 +274,7 @@ void WLibraryTableView::focusInEvent(QFocusEvent* event) {
         // This is especially helpful if the table has only one track, which can
         // not be selected with up/down buttons, either physical or emulated via
         // [Library],MoveVertical controls. See #9548
-        if (model()->rowCount() > 0) {
+        if (model() != nullptr && model()->rowCount() > 0) {
```
ronso0, line 277: Why not just `if (model() && model()->rowCount() > 0) {`?

ronso0:
Thanks!
I suggest to squash into one commit and rephrase the message to clarify the fix (not which symptom it avoids), something along the lines of
"WLibraryTableView: assert model when selecting row"

uklotzde:
> Thanks! I suggest to squash into one commit and rephrase the message to clarify the fix...

My recommendation for Mixxx:

image

One commit on upstream per PR with a customizable commit message on merge. Clean history with meaningful commits under your control. Example: https://github.com/tokio-rs/tokio/commits/master

daschuer:
We have discussed the sqash merging earlier and decided against. The reason is that we loose the meaningful commit messages form the intermediate commits. We want to have them in our repository for bisect and blame. We don't want to rely on the history kept by GitHub that is not under our control.

ronso0:
While this argmuent is true for most cases, squash & merge would be helpful in (admittedly rare) cases like this where we're talking about tiny commits and fixups that can(should) be squashed safely without loosing meaningful info.

ronso0:
**PS**

IMO this would accelerate the workflow.
This PR would already have been merged by now, but we need the contributor to do the work locally, push, then merge.

IMO we should enable Squash & merge **but** define the conditions precisely. E.g. one-commit PRs with additional fixups, text-only changes (new blog post, adding manual chapters), new controller mappings.

Let's discuss that on Zulip later on, maybe when 2.4 is out the door?

--enough off-topic--

This is ready. @JoergAtGithub would you mind to squash & rename?

JoergAtGithub:
I did already

ronso0:
I'm sorry I overlooked that, Notifications send me to the comment after.

Thank you, LGTM and thanks for quickly fixing my failure, of course!

PR #12070, opened 2023-10-06
Search: also search for half & double BPM

ronso0:
`bpm:N` now also finds N*2 and N/2, the latter with a range of prev - next integer in case N is odd, considering electronic music with mostly integer BPM.

In case the behaviour is unexpected and disturbing for users we may add an option to Preferences > Library > Search (default on).

This may be extended / rewritten for range operators `n-m` if desired (fuzzy BPM match), and in that case probably better moved to new FilterNode `BpmFilterNode`.

- [ ] add a test

If wanted this for years and since it's working I'll use it, but I can't tell when I'll finish it (tests, polishing). Feel free to adopt/edit if there's a demand.

### DIFF
--- a/src/library/searchqueryparser.cpp
+++ b/src/library/searchqueryparser.cpp
@@ -186,9 +186,36 @@ void SearchQueryParser::parseTokens(QStringList tokens,
                                        .trimmed();
 
             if (!argument.isEmpty()) {
+                bool isSingleNumber = false;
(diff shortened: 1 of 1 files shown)


ronso0:
Meeh, seems I broke too many use cases.
I'll go with `BpmFilterNode`, that allow to get both half/double search and fuzzy search.

ronso0:
new attempt in #12072

PR #12072, opened 2023-10-07
Search: add special BPM filters

### LINKED ISSUE #8191: fuzzy BPM search
Date: 2015-08-04T16:23:09Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1481411](https://bugs.launchpad.net/bugs/1481411)

---

It would be convenient to have a fuzzy BPM search operation rather than manually typing in a BPM range. For example "~bpm:115" could search for tracks with a BPM within 5 (or whatever number is decided upon) BPM of 115.

ronso0:
rework of #12070
Closes #8191 

This adds `BpmFilterNode` which allows the following special filtering:
* `~bpm:N` finds `N` +- n%
n is % of the current rate slider range, can be configured in Pref > Library, default is 75%
* `bpm:N` finds N¹ as well as N/2 and N*2 (in case N is odd N/2 is replaced with a range of prev - next integer, considering electronic music with mostly integer BPM)
can be toggled in Pref > Library, default ON

¹N is turned into a range to cover the rounded values (in the tracks table or BPM widgets).
For example 102.5 becomes [102.45 - 102.55]

Btw, this works beautifully with the new `OR` operator: `~bpm:80 | ~bpm:160` : )

**Results:**
* `~bpm:100`: 94-106 **NEW**
* `bpm:125.2`: [125.15 - 125.25] and [250 - 251] and [52 - 53] **NEW**
* `-bpm:100`: anything except exactly 100.0
* `bpm:=100`: 125.15 - 125.25

**Preferences > Library**
I re-grouped the options a bit, though we may discuss details like Metadata /Playlist title and the search note label
- moved Misc options to Metadata & Track Table groups
- added the Search group
  - fuzzy range selector (percentage of pitch slider)

library-search-options_

**TODO**
- [x] adjust NumericFilterTests to use something else than `bpm`
- [x] add BpmFilterNode tests
- [x] adjust 'fuzzy range' in Preferences > Library > Search
- [x] use fuzzy range also for 'Search related tracks' > BPM?
- [x] half/double search can be overriden with `=` operator (we just need to document it)

**Manual**
- [x] documentation PR: https://github.com/mixxxdj/manual/pull/693

### DIFF
--- a/src/library/library_prefs.cpp
+++ b/src/library/library_prefs.cpp
@@ -48,6 +48,11 @@ const ConfigKey mixxx::library::prefs::kSearchDebouncingTimeoutMillisConfigKey =
                 mixxx::library::prefs::kConfigGroup,
                 QStringLiteral("SearchDebouncingTimeoutMillis")};
 
+const ConfigKey mixxx::library::prefs::kSearchBpmFuzzyRangeConfigKey =
+        ConfigKey{
+                mixxx::library::prefs::kConfigGroup,
+                QStringLiteral("search_bpm_fuzzy_range")};
+
 const ConfigKey mixxx::library::prefs::kEnableSearchCompletionsConfigKey =
         ConfigKey{
                 mixxx::library::prefs::kConfigGroup,
--- a/src/library/searchquery.cpp
+++ b/src/library/searchquery.cpp
@@ -471,6 +471,86 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
     return 60 * m + s;
 }
 
+BpmFilterNode::BpmFilterNode(QString& argument)
+        : m_isNullQuery(false),
+          m_isOperatorQuery(false),
+          m_isRangeQuery(false),
+          m_operator("="),
+          m_bpm(0.0),
+          m_rangeLower(0.0),
+          m_rangeUpper(0.0) {
+    if (argument == kMissingFieldSearchTerm) {
+        m_isNullQuery = true;
+        return;
+    }
+
+    QRegularExpressionMatch opMatch = kNumericOperatorRegex.match(argument);
+    if (opMatch.hasMatch()) {
+        m_operator = opMatch.captured(1);
+        argument = opMatch.captured(2);
+    }
+
+    bool isDouble = false;
+    const double bpm = argument.toDouble(&isDouble);
+    if (isDouble) {
+        m_bpm = bpm;
+        m_isOperatorQuery = true;
+    }
+    // else test if this is a valid range query
+    QStringList rangeArgs = argument.split("-");
+    if (rangeArgs.length() == 2) {
+        bool lowOk = false;
+        m_rangeLower = rangeArgs[0].toDouble(&lowOk);
+        bool highOk = false;
+        m_rangeUpper = rangeArgs[1].toDouble(&highOk);
+
+        if (lowOk && highOk && m_rangeLower <= m_rangeUpper) {
+            m_isRangeQuery = true;
+        }
+    }
+}
+
+bool BpmFilterNode::match(const TrackPointer& pTrack) const {
+    double value = pTrack->getBpm();
+    if (m_isNullQuery && value == mixxx::Bpm::kValueUndefined) {
+        return true;
+    }
+
+    if (m_isOperatorQuery) {
+        if ((m_operator == "=" && value == m_bpm) ||
+                (m_operator == "<" && value < m_bpm) ||
+                (m_operator == ">" && value > m_bpm) ||
+                (m_operator == "<=" && value <= m_bpm) ||
+                (m_operator == ">=" && value >= m_bpm)) {
+            return true;
+        }
+    } else if (m_isRangeQuery && value >= m_rangeLower &&
+            value <= m_rangeUpper) {
+        return true;
+    } else if (value == m_bpm) {
+        return true;
+    }
+    return false;
+}
+
+QString BpmFilterNode::toSql() const {
+    if (m_isNullQuery) {
+        return QString("bpm IS NULL");
+    }
+
+    if (m_isOperatorQuery) {
+        return QString("bpm %1 %2").arg(m_operator, QString::number(m_bpm));
+    }
+
+    if (m_isRangeQuery) {
+        return QString(QStringLiteral("bpm BETWEEN %1 AND %2"))
+                .arg(QString::number(m_rangeLower),
+                        QString::number(m_rangeUpper));
+    }
+
+    return QString();
+}
+
 KeyFilterNode::KeyFilterNode(mixxx::track::io::key::ChromaticKey key,
         bool fuzzy) {
     if (fuzzy) {
--- a/src/library/searchquery.cpp
+++ b/src/library/searchquery.cpp
@@ -486,13 +486,41 @@ void BpmFilterNode::setBpmRelativeRange(double range) {
     s_relativeRange = range;
 }
 
+namespace {
+
+inline QString rangeSqlString(double lower, double upper) {
+    return QStringLiteral("bpm BETWEEN %1 AND %2")
+            .arg(QString::number(lower),
+                    QString::number(upper));
+}
+
+inline std::pair<double, double> rangeFromTrailingDecimal(double bpm) {
+    // Set up a range if we have decimals. This will include matches
+    // for which we show rounded values in the library. For example
+    // 124  finds 119.95 - 124.05
+    // 124.1  finds 124.05 - 124.15
+    // 124.92 finds 124.915 - 124.925
+    if (bpm == 0.0) {
+        return std::pair<double, double>(bpm, bpm);
+    }
+    int numDecimals = 1;
+    double intPart;
+    double fractPart = std::modf(bpm, &intPart);
+    if (fractPart != 0.0) {
+        QString decimals = QString::number(fractPart);
+        numDecimals = decimals.split('.').at(1).length();
+    }
+    double roundRange = 5 / pow(10, numDecimals + 1);
+    // Don't search for negative BPM
+    double lower = std::max(0.0, bpm - roundRange);
+    double upper = bpm + roundRange;
+    return std::pair<double, double>(lower, upper);
+}
+
+} // namespace
+
 BpmFilterNode::BpmFilterNode(QString& argument, bool fuzzy, bool negate)
-        : m_fuzzy(fuzzy),
-          m_negate(negate),
-          m_isNullQuery(false),
-          m_isOperatorQuery(false),
-          m_isRangeQuery(false),
-          m_isHalfDoubleQuery(false),
+        : m_matchMode(MatchMode::Invalid),
           m_operator("="),
           m_bpm(0.0),
           m_rangeLower(0.0),
@@ -502,14 +530,15 @@ BpmFilterNode::BpmFilterNode(QString& argument, bool fuzzy, bool negate)
           m_bpmDoubleLower(0.0),
           m_bpmDoubleUpper(0.0) {
     if (argument == kMissingFieldSearchTerm) {
-        m_isNullQuery = true;
+        m_matchMode = MatchMode::Null;
         return;
     }
 
     QRegularExpressionMatch opMatch = kNumericOperatorRegex.match(argument);
     if (opMatch.hasMatch()) {
-        if (m_fuzzy) {
+        if (fuzzy) {
             // fuzzy can't be combined with operators
+            // m_matchMode is already Invalid.
             return;
         }
         m_operator = opMatch.captured(1);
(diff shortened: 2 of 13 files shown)


ronso0:
This works as intended now.

We may discuss the default behaviour and the fuzzy options:

- adjust 'fuzzy range'?
  - dynamically: full or part of pitch slider range?
  - in Preferences > Library > Search?
    * % range spinbox?
    * % range spinbox clamped to current pitch slider range?
    * also apply range to 'Search related tracks' > BPM?
- toggle half/double search with checkbox in Preferences > Library > Search ?

ronso0:
This works beautifully now, please test. Manual PR coming soon.
This has grown way beyond what I had in mind initially (story of my life..) so I'll take a break now. No hurry with reviews and so on.

Swiftb0y (commented):
only looked at the first commit so far. The rest got too complicated too quickly for me. I think that we can merge without N/2 & N*2 as well as the configurable config option since those two changes incur the most complexity.

ronso0:
> I think that we can merge without N/2 & N*2

Half/was the reason I started this. Tbh I'm not motivated to split that up, first I'll wait for a real Mixxx session and test it in real life.

src/library/searchquery.cpp
```diff
@@ -471,6 +473,62 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
...
+BpmFilterNode::BpmFilterNode(const QString& argument, bool fuzzy)
+        : NumericFilterNode(QStringList(QString("bpm"))),
+          m_fuzzy(fuzzy) {
+    // init() has to be called from this class directly to invoke
+    // the implementation of this and not that of the base class!
+    init(argument);
+}
+
+void BpmFilterNode::init(QString argument) {
+    if (argument == kMissingFieldSearchTerm) {
```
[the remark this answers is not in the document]
ronso0: yup, good catch. Leftover from c/p'ing from NumericFilter.

Swiftb0y:
I'm sorry, I'm not able to review this further then right now.

ronso0:
Never mind, didn't expect anyone soon. Thanks for taking a look.

src/library/searchquery.cpp
```diff
@@ -471,6 +473,62 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
...
+        return;
+    }
+
+    QRegularExpressionMatch opMatch = kNumericOperatorRegex.match(argument);
+    if (opMatch.hasMatch()) {
+        m_operator = opMatch.captured(1);
+        argument = opMatch.captured(2);
+    }
+
+    bool parsed = false;
```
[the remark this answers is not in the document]
ronso0: now inlined in `BpmFilterNode` ctor, named `isDouble`

ronso0:
I've split this up into a bit, and I removed the half/double config option since that can be overwritten easily with the `=` operator. Hope this is a bit easier to review now (but don't feel pushed, I'm aware this has low prio atm).

ywwg:
There was a tiktok going around about doing bpm matches at x/0.75 to go from house (129ish) to drum and bass (172) using a 3/4 beatloop.  Would it be possible to extend this feature (not necessarily in this PR) to calculate multiples like that?

ronso0:
It's been a while since I last worked on this..
Looking at ebd39a8dfc68010cf3fe1ad03deaefd7bb0df88e, this is certainly possible. We could add a factor spinbox to the preferences and use that.

The current half/double could also be extended to cover a certain range (not sure what the current range is, but Ithink it's exact half/double).

ronso0:
I have simplified the half/double query generation and rebased onto main.

I think the half/double search is too strict (searches _exact_ half/double BPM), can probably be a (small) range, too.
But we some feedback from real mixxxing first.

src/library/library_prefs.cpp
```diff
@@ -48,6 +48,11 @@ const ConfigKey mixxx::library::prefs::kSearchDebouncingTimeoutMillisConfigKey =
                 mixxx::library::prefs::kConfigGroup,
                 QStringLiteral("SearchDebouncingTimeoutMillis")};
 
+const ConfigKey mixxx::library::prefs::kSearchBpmFuzzyRangeConfigKey =
+        ConfigKey{
+                mixxx::library::prefs::kConfigGroup,
+                QStringLiteral("search_bpm_fuzzy_range")};
```
daschuer, line 54: We have "SearchDebouncingTimeoutMillis" and "search_bpm_fuzzy_range" 
Do we want to follow the first in this case?

src/library/searchquery.h
```diff
@@ -184,6 +184,52 @@ class DurationFilterNode : public NumericFilterNode {
     double parse(const QString& arg, bool* ok) override;
 };
 
+// BPM filter that supports fuzzy matching via ~ prefix.
+// If no operator is provided (bpm:123) it also finds half & double BPM matches.
+// Half/double values aren't integers, int ranges are used. E.g. bpm:123.1 finds
+// 61-61, 123.1 and 246-247 BPM
```
daschuer, line 190:
```suggestion
// 61-62, 123.1 and 246-247 BPM
```

daschuer: Why it finds only a full match in the 1x case but a 1 BPM rang in other cases? I don't think that the underlying use case is different. 
Did you consider to always match integer rounded values? 

Or distinguish 123.0 for only 123.0 but 123 for 122.5 .. 123.5

daschuer (changes requested):
Unfortunately searching for 98,7 BPM is broken on a German locale. Also using a decimal dot does not fix the issue.

src/library/library_prefs.cpp
ronso0: I thought we want to migrate to snake case for ConfigKeys?

ronso0:
> Unfortunately searching for 98,7 BPM is broken on a German locale. Also using a decimal dot does not fix the issue.

Please share the debug output of "98,7" and "98.7".

src/library/searchquery.h
ronso0: It's only a 1 BPM for half and double if the search term has decimals, because it that case it seems unlikely the user has/wants exact half/double matches.

ronso0:
> Unfortunately searching for 98,7 BPM is broken on a German locale. Also using a decimal dot does not fix the issue.

This happens also with `main`.
Trying to find 124.92:
`bpm:124,9`: no filter, all tracks shown
`bpm:124.9`: no results
`bpm:124.`: only exact matches of 124.0

I added replacement of `,` to `.` which fixes the issue.
Also I turned the basic query into a range, rounding up/down the last decimal. This allows finding tracks for which we see rounded BPM in the tracks table.

ronso0:
I'll update this PR soonish.

src/library/library_prefs.cpp
daschuer: I don't mind. I just wanted to mention it.

src/library/searchquery.h
daschuer: Yes I agree but why do the user want exact 1x matches but not exact 2x and 0.5x matched. 
You can easily have the same track annotate as 123 or 62,5. So it will be found if it is in the 0.5x range but not in the 1x range. That doe not feel reasonable. 

Did you consider to use this: 
* 123 find 122.5 .. 123.5 | 62 .. 63 | 245,5 .. 246,5 
* 123.3 find 123.25 .. 123.35 | 62,1 .. 63,2 | 246,55 .. 246,65
That means that the omitted decimal places are matching to the rounded value of the searched tracks.

ywwg:
oooo I have been wanting EXACTLY this feature

src/library/searchquery.cpp
```diff
@@ -471,6 +471,221 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
...
+          m_rangeUpper(0.0),
+          m_bpmHalfLower(0.0),
+          m_bpmHalfUpper(0.0),
+          m_bpmDoubleLower(0.0),
+          m_bpmDoubleUpper(0.0) {
+    if (argument == kMissingFieldSearchTerm) {
+        m_isNullQuery = true;
+        return;
+    }
+    qWarning() << "     .BpmFilter:" << argument;
```
ywwg, line 508: As written, this will create a lot of warning spam.  In other modules we have a localized debug option for debugging that specific feature by namespace (see EngineBuffer).

src/library/searchquery.cpp
```diff
@@ -471,6 +471,221 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
...
+            m_isRangeQuery = true;
+        } else if (!opMatch.hasMatch() && !m_negate) {
+            qWarning() << "     .halfDouble";
+            // Simple 'bpm:NNN' search.
+            // Also searches for half/double matches.
+            // If decimals are provided, extend the core range, else search for
+            // exact matches.
+            ifDecimalsSetRange(argument, bpm);
+            // Include half/double BPM (rounded to int)
+            m_bpmHalfLower = floor(bpm / 2);
```
ywwg, line 542: it's a little awkward how some ranges are updated in the function as a side effect and others are updated manually.   And then, there is a lot of logic that switches depending on the mode of the query, which I believe is actually constant once this object is created, right? This creates a bit of difficulty following the logic with a lot of conditional branching.  

Based on the call pattern in this node, I think it might make more sense to use a different programming pattern: write a separate function for each search type that takes the constant parameters as currying arguments

So the constructor is basically:

```
switch (mode) {
 case operator_mode:
   m_matchFunc = match_operator_fn(op, bpm)
  case fuzzy_mode:
    m_matchFunc = match_fuzzy_fn(bpm)
  ...
}

bool BpmFilterNode::match(const TrackPointer& pTrack) const {
  // get the track bpm...
  return m_matchFunc(track_bpm);
}
```

and then 

```
my_matcher_fn_type match_operator_fn(string op, bpm compare_bpm) {
	// precalculate range and whatever

	// curry necessary arguments and return a customized function for this search
	return [=](float track_bpm) {
      // do stuff with range as compared to track_bpm
	};
}
```

This way the individual match functions can be easily tested in isolation and the ranges are still precalculated but don't require separate member variables to do so.

src/library/searchquery.cpp
ronso0: yes, the last commit is for developing and I'll remove that commit
(usually I mention this in the PR TODO, recently I've been abusing "fixup!" to block accidental merge)

ronso0:
> The matchMode enum is certainly a good idea to make the code more readable, I'll try that.

Alrighty, this is much cleaner now.

The commit with debug output is bd83d54f76b0b395832f1d23bb5f92c1691f5968

ronso0:
I think the match lambdas are actually no longer necessary after reafactoring, but it was an interesting exercise.
Can as well move the match functions to match() and add the switch() there.

src/library/searchquery.cpp
```diff
@@ -471,6 +471,217 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
...
+    }
+    case MatchMode::Operator: {
+        return (m_operator == "=" && value == m_bpm) ||
+                (m_operator == "<" && value < m_bpm) ||
+                (m_operator == ">" && value > m_bpm) ||
+                (m_operator == "<=" && value <= m_bpm) ||
+                (m_operator == ">=" && value >= m_bpm);
+    }
+    default: // e.g. MatchMode::Invalid
+        // Show no results to indicate the query is invalid.
```
ywwg, line 653: this can be a debug assert -- in a release build we can just silently fail, but during development we'd want to know it was happening

src/library/searchquery.cpp
```diff
@@ -471,6 +471,217 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
...
+        }
+        m_operator = opMatch.captured(1);
+        argument = opMatch.captured(2);
+    }
+
+    // Replace the locale's decimal separator with .
+    // This is handy if numbers are typed with the numpad.
+    argument.replace(',', '.');
+    bool isDouble = false;
+    double bpm = argument.toDouble(&isDouble);
```
ywwg, line 552: bpm is only used after this mode conditional branch, can you move it down to, say, line 600 or so? That way it's closer to where it's being used

ywwg (approved):
this is much cleaner, thank you

src/library/searchquery.cpp
ronso0: Well, we need to know if the argument is a double here already, and doing so with toDouble() gives us the bpm.
We may assign m_bpm = bpm right away at the top of the isDouble branch, regardless whether it'll be used or not (that would save the ::Operator case at the end of this function).

src/library/searchquery.cpp
ronso0: MatchMode::Invalid is not for programmming errors but for malformed **typed** queries (user error), for example
fuzzy range ~bpm:123-125
fuzzy operator ~bpm:<=95

ywwg:
@daschuer what do you think?

daschuer:
I have just tested with the decimal points and it does not feel consistent: 

bpm:114 does find 114.0 only 
bpm:114.8 finds 114,8 only 

This is kind of OK, but it would be convenient to have this match. 
bpm:114 = 113.5 .. 114.5
bpm:114.0 = 113.95 .. 114.05
bpm:114.8 = 114.75 .. 114.85

It is IMHO a bug that:
bpm:57 does find 114.0 only 
bpm:57,4 finds 114,8 114.0 and 115,0 

My expectation would be that 
bpm:57 does find 113.0 .. 115.0 (double the range as proposed above)  
bpm:57,0 does find 113.9 .. 114.1 
bpm:57,4 finds 114,7 ... 114.9

ronso0:
Indeed, UX with/without decimals needs to be improved.

However, your observation is not quite correct (it may be that you simply don't have tracks that would be found):
`bpm:114`: 113.95 - 114.05 or 57 or 228
`bpm:114.8`: 114.75 - 114.85 or 57 - 58 or 229 - 230

Actually I don't really agree with your expectation, 114 and 114.0 should be handled identically.
But if we'd make
`bpm:114`: 113.95 - 114.999
(should find everything that is shown as 114.**, but some hits may also be shown as 115.0 which might be confusing)
maybe users would then try
`bpm:114.0`: 113.95 - 114.05 (find everything that is shown as 114.0)
and it becomes clear how to use `bpm:`

Seems doable, just requires two more _MatchMode_s (I think).

ronso0:
new debug commit is f306ef1893c2a25bfc43aa30d4caf464a032e1fa

daschuer:
> Actually I don't really agree with your expectation, 114 and 114.0 should be handled identically.

There is no strong demand for it. I got the impression you had implemented that. So I was wrong. 

My expectations is just the reverse, our library rounding does. The underlying use case is: If the library displays a rounded value, the same number should be able to reveal it. 

The point I think that should be changes in any case is the different rule for 2x match. IMHO it should reveal the same boundaries 2x just as if you had divided the bpm by 2 before.

ronso0:
And I have to adjust the test to the new behaviour.

daschuer:
In my test `bpm:76` reveals 76.0 76.5 151.0 152.0 153.0 
finding 151.0 is surprising, because 151.0/2 is is 75.5 out of the original range. Do we have a use case for this?

The rest works good.

ronso0:
> In my test bpm:76 reveals 76.0 76.5 151.0 152.0 153.0
finding 151.0 is surprising, because 151.0/2 is is 75.5 out of the original range.

but it matches the expectation you mentioned above ; )

> My expectation would be that
bpm:57 does find 113.0 .. 115.0 (double the range as proposed above)

daschuer:
My original expectation was that it works like commercial rounding. The idea to round up to have a more text based matching works for me as well. 

In both cases my expectation is that the boundaries are just multipied to match also compatible tracks. That's easy to describe in the manual. 
Just like as if you have adjusted the beats with the multiplier in the context menu. 

I don't want to stress that point to much. If you think it is OK, go ahead. I was just unsure if it is intended or a leftover from earlier versions.

src/library/searchquery.cpp
```diff
@@ -471,6 +471,263 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
...
+    // 'BETWEEN' is inclusive, i.e. returns true if lower <= value <= upper
+    return QStringLiteral("bpm BETWEEN %1 AND %2")
+            .arg(QString::number(lower),
+                    QString::number(upper));
+}
+
+inline std::pair<double, double> rangeFromTrailingDecimal(double bpm) {
+    // Set up a range if we have decimals. This will include matches
+    // for which we show rounded values in the library. For example
+    // 124  finds 119.95 - 124.05
```
daschuer, line 501: Is this example still correct here?

src/library/searchquery.h
daschuer: The comment is now outdated.

src/library/searchquery.cpp
ronso0: nope.
I'm about to clean up and add some examples in comments to make the modes and ranges easier to grasp.

ronso0:
Done.
I'm sorry I had to force-push, I removed the fixup locally.

New DEBUG commit is ec9b5570ef

src/test/searchqueryparsertest.cpp
```diff
@@ -402,84 +400,143 @@ TEST_F(SearchQueryParserTest, NumericFilterEmpty) {
...
+    // HalveDouble
+    auto pQuery(m_parser.parseQuery("bpm:127", QString()));
     TrackPointer pTrack = newTestTrack();
+    EXPECT_FALSE(pQuery->match(pTrack));
+    pTrack->trySetBpm(127.13);
+    EXPECT_TRUE(pQuery->match(pTrack));
+
+    EXPECT_STREQ(
+            qPrintable(QString("(bpm >= 126.95 AND bpm < 128) OR "
+                               "(bpm BETWEEN 63 AND 64) OR "
```
daschuer, line 510: Reading this test, I think I got your point:

If we consider that we have a electronic music library with only *.0 BPMs the resulting 1/2 range of  >= 63,5 and < 64.0  
64 will produce no match. Even if matches they are only 1% off.

Is this your rational? Can you put that somewhere as comment?

src/test/searchqueryparsertest.cpp
```diff
@@ -402,84 +400,143 @@ TEST_F(SearchQueryParserTest, NumericFilterEmpty) {
...
+                               "(bpm BETWEEN 253 AND 255)")),
```
daschuer, line 511: 127 * 2 = 254 
128 * 2 = 256
following the rational above we will find *.0 tracks with just `bpm >= 254.95 AND bpm < 255`
The current implementation lead to a not fully overlapping range: 
```
original 
    127 [----[ 128 
x 2 (my expectation) 
    254 [--------[ 256
implementation
253 [--------] 255 
/2
126,5 [----] 127,5
```

Do we have a reason for this to match tracks, that would not be no longer match after editing the BPM /2? 
I have probably not understand your goal.

daschuer (changes requested):

ywwg:
@daschuer Is there specific behavior in the parser that you consider blocking your approval of this PR? Can you be specific about which query results you consider incorrect?

daschuer:
OK, I will give it a try.

daschuer:
You will find my proposal here: https://github.com/ronso0/mixxx/pull/67

daschuer (approved):
Is this ready for merge after CI is done?

Thank you very much for this nice enhancement!

ronso0:
Yup, IMO this is ready to go.

I'll update the manual ASAP.

daschuer:
After a final test I have discovered on remaining issue, that an even bpm value only reveals the single exact halve value. 
128 reveals only "64.000000" 

I have fixed it here: 
https://github.com/ronso0/mixxx/pull/68

src/library/searchquery.cpp
```diff
@@ -471,6 +473,277 @@ double DurationFilterNode::parse(const QString& arg, bool* ok) {
...
+            // m_matchMode is already Invalid.
+            return;
+        }
+        m_operator = opMatch.captured(1);
+        argument = opMatch.captured(2);
+    }
+
+    // Replace the locale's decimal separator with .
+    // This is handy if numbers are typed with the numpad.
+    argument.replace(',', '.');
```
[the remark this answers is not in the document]
daschuer: I considered your solution as quite smart, because we display the bpm with dot separator in the GUI and Library table but use a comma separator in editors. It is unlikely that one is using bpms above 999.

This issue is however only present in case the user has a locale with comma separators. In case of a dot separator local everything is consistent and the hack is probably not that smart. We should make it conditional then.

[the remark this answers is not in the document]
ronso0: So, ideally we'd have a regex using QLocale::groupSeparator() and QLocale::decimalPoint()

I'd prefer to merge this now to collect alpha/beat feedback and work on the regex in a follow-up.

@daschuer @ywwg Ready to roll?

ywwg (approved):

ywwg:
sorry for the hold-up.  lgtm!

PR #12074, opened 2023-10-07
make decks' xfader assignment persistent

### LINKED ISSUE #10122: Crossfader orientation isn't saved/restored
Reported by: **[mxmilkiib](https://github.com/mxmilkiib)**
Date: 2020-09-23T20:02:49Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1896832](https://bugs.launchpad.net/bugs/1896832)
Tags: easy, usability

---

If I set all channels to ignore the crossfader, then quit and restart, the options I had set have reset.

ronso0:
Closes #10122

I agree with Be's [comment](https://github.com/mixxxdj/mixxx/issues/10122#issuecomment-1225430189):
> I am leaning towards simply making the crossfader orientation switches persistent in all cases. This is how every 4 deck mixer and controller with crossfader assignment switches works. I think if a user wants to change these, they probably want that change to persist.

### DIFF
--- a/src/engine/channels/enginechannel.cpp
+++ b/src/engine/channels/enginechannel.cpp
@@ -23,10 +23,11 @@ EngineChannel::EngineChannel(const ChannelHandleAndGroup& handleGroup,
     m_pMainMix = new ControlPushButton(ConfigKey(getGroup(), "main_mix"));
     m_pMainMix->setButtonMode(ControlPushButton::POWERWINDOW);
     m_pMainMix->addAlias(ConfigKey(getGroup(), QStringLiteral("master")));
-    m_pOrientation = new ControlPushButton(ConfigKey(getGroup(), "orientation"));
+    // crossfader assignment is persistent
+    m_pOrientation = new ControlPushButton(
+            ConfigKey(getGroup(), "orientation"), true, defaultOrientation);
     m_pOrientation->setButtonMode(ControlPushButton::TOGGLE);
     m_pOrientation->setStates(3);
-    m_pOrientation->set(defaultOrientation);
     m_pOrientationLeft = new ControlPushButton(ConfigKey(getGroup(), "orientation_left"));
     connect(m_pOrientationLeft, &ControlObject::valueChanged,
             this, &EngineChannel::slotOrientationLeft, Qt::DirectConnection);


src/engine/channels/enginechannel.cpp
```diff
@@ -23,10 +23,10 @@ EngineChannel::EngineChannel(const ChannelHandleAndGroup& handleGroup,
     m_pMainMix = new ControlPushButton(ConfigKey(getGroup(), "main_mix"));
     m_pMainMix->setButtonMode(ControlPushButton::POWERWINDOW);
     m_pMainMix->addAlias(ConfigKey(getGroup(), QStringLiteral("master")));
-    m_pOrientation = new ControlPushButton(ConfigKey(getGroup(), "orientation"));
+    // crossfader assignment is persistent
+    m_pOrientation = new ControlPushButton(ConfigKey(getGroup(), "orientation"), true);
```
Swiftb0y, line 27:
```suggestion
    // crossfader assignment is persistent
    m_pOrientation = new ControlPushButton(ConfigKey(getGroup(), "orientation"), true, defaultOrientation);
```

ronso0: :+1: of course, picked the 1st wrong commit and didn't double-check

daschuer (approved):
LGTM, thank you. 
Can this become a 2.4 fix?

ronso0:
Personally I don't care.
It's a 3yo feature request, which in some settings may be considered a bug, or rather inconvenience.
For users who always want/need certain samplers to play on the left/right channel bus (VC setup with 2-channel mixer) it's am annoyance.
As Be stated correctly IMO, users who change the xfader assignment for the main decks do this either in skins (and the skin & skin config is persistent across restarts so the xfader toggles are visible) or via controller switches (which also persist).
And we didn't document that the xfader switches are reset on each start.
So, even though it's a behaviour change I don't consider it a regression or significantly break use cases.


If I didn't overlook any aspects may indeed target 2.4

Swiftb0y:
don't see any reason this couldn't go into 2.4.

One thought I had during the curse of the recent couple PRs that made more CO's persistent was that we should probably have a button somewhere in the preferences that reset _all_ CO's to their defaults.
Imagine this: A user messed with some controller or with the CO tools that changed the audio routing or just some characteristics ("orientation, "volume", "gain", etc.) Their Skin doesn't over the same amount of controls and just assumed some of these CO's to be at their default (simplified skins might make this assumption for example). The result could be that the user would end up with Mixxx in a state they can't explain and also not easily diagnose or change. Previously, a restart would've fixed the issue since the CO would've been reset from that. Now the only way for them to fix their install would be to delete the file where the CO values are saved (likely by just blatantly nuking their settings directory as a whole). The entire process is quite unpleasant so there should be a more user friendly and easier to discover way.

Just a thought, this is likely an incredibly unlikely case and very much contrived. Still something I thought of (I think I remember being a situation like this in my early days where I started messing with mixxx internals).

ronso0:
FWIW simplified skins should set inaccessible controls in the manifest, like Shade does with the samplers' effect routing switches:
https://github.com/mixxxdj/mixxx/blob/3db5755418a51b709da73bd5a1c07c8fa1da97b6/res/skins/Shade/skin.xml#L83-L94
We can't do anything about this if custom skins are used :man_shrugging: 

Though, a **Total Reset**:tm: button would be handy indeed.

daschuer (approved):
LGTM, Thank you

PR #12080, opened 2023-10-08
Warn if ccache is not configured for precompiled headers

daschuer:
See https://ccache.dev/manual/4.1.html#_precompiled_headers


JoergAtGithub:
Make sense to me - Recommendation in warning text matches the ccache documentation.  Thank you!

Issue #12084, opened 2023-10-09
Include right single quotation marks in searches


mminer237:
### Feature Description

I have a number of songs from Qobuz with `’` (`U+2019 : RIGHT SINGLE QUOTATION MARK`) in the title:

![’](https://github.com/mixxxdj/mixxx/assets/2881340/af2ee0ad-8597-414d-9686-f41a8f58f90c)

Currently it's very hard to search for these songs because if I am a normal person and type an apostrophe (`'`, `U+0027`) in its place it will not match any of them. I can't imagine anybody ever only wanting to match one character instead of the other. I think when one is searched for, Mixxx should automatically search for both variations.

Theoretically other normalizations would be good as well, but this seems like the most pressing and easier to fix.

acolombier:
Adding scope for other characters alias in here, such as accentuated letters (`é` for `e`)

PR #12085, opened 2023-10-09
Try apostrophes and right single quotation marks

mminer237:
Implements #12084


fwcd (commented):
Thanks for the PR! I agree that better support for fuzzy matching would be great, I've left some thoughts below.

In any case, please make sure to add some test cases [here](https://github.com/mixxxdj/mixxx/blob/main/src/test/searchqueryparsertest.cpp) too.

src/library/searchqueryparser.cpp
```diff
@@ -291,6 +291,18 @@ std::unique_ptr<OrNode> SearchQueryParser::parseOrNode(const QString& query) con
...
+            if (rawAndNode.contains(QChar('\''))) {
+                QString quotationMarkAndNode = rawAndNode;
+                quotationMarkAndNode.replace(QChar('\''), QChar(0x2019));
+                pQuery->addNode(parseAndNode(quotationMarkAndNode));
+            }
+            if (rawAndNode.contains(QChar(0x2019))) {
+                QString apostropheAndNode = rawAndNode;
+                apostropheAndNode.replace(QChar(0x2019), QChar('\''));
+                pQuery->addNode(parseAndNode(apostropheAndNode));
+            }
```
fwcd, line 304: We should be a bit careful about performing substitutions on the pre-tokenized input. With these characters it's likely to be fine, but e.g. `"` is used to quote strings.

I also feel like this could be generalized. Instead of hardcoding substitutions, perhaps there is a way to perform a fully symbol-insensitive search (unless e.g. the query is quoted)?

IIRC there were also aalready some earlier discussions on adding 'actual' fuzzy search, which would make these tricks largely irrelevant since most parts of the string would still match:

- #7470

mminer237: I was conscious of that, but it's much simpler to do it before than trying to copy and modify the already-created nodes. At least currently it wouldn't be an issue with `'`.

It could be done by modifying the SQL queries to `REPLACE()` certain symbols or regex could be used in the SQL to strip out symbols in general, but I was concerned by the performance implications and honestly couldn't think of other common symbols, although I'm sure I'm missing some. I guess words may be hyphenated or not.

Fuzzy search would be an alternative, but my understanding is that it will require prepending a tilde (e.g., `~that's`) to match. I doubt most people will know there are multiple types of apostrophes to know they may need to use fuzzy matching to find something. Plus, it's just more broad, and you don't always want "close" words just because you have apostrophes.

I will write some tests though. Another thought is that I think using the [ICU SQLite extension](https://www.sqlite.org/src/artifact?ci=trunk&filename=ext/icu/README.txt) for normalization would be the best way to treat equivalent symbols the same, and I think that would work without editing the search code since I think it just replaces `LIKE` under the hood. I'm just less familiar with that. I'll look into that as a more general alternative.

Thanks for your feedback!

acolombier:
Hi @mminer237, thanks for your contribution here! Improving searching by adding a level of fuzziness with char aliasing would be great! For example, you have the same case in french, where you may have title with `é` but you would usually want them to display when searching with a simple `e`.

Do you have interest to make your solution for generic and allow a map of aliases that we can extends?

ronso0:
No response by author for 2.5 years. Close?

mminer237:
It's still on my to-do list to try to see if I can do this in SQL more
generically, but I haven't found the time yet.

PR #12091, opened 2023-10-12
Update manifest even if one job fails

Holzhaus:


daschuer:
Can you describe the new behaviour and the old? 

For my understanding we want a manifest only if the upload succeeds as well.Does that still works? 
Do we perform an upload if signing fails now? 
Does this ends up in unsigned binaries without someone notice it?

.github/workflows/build.yml
```diff
@@ -370,6 +370,7 @@ jobs:
       - name: "[macOS] Sign, Notarize, and Staple Package"
         if: runner.os == 'macOS' && env.MACOS_CODESIGN_OPENSSL_PASSWORD != null && env.MACOS_CODESIGN_CERTIFICATE_PASSWORD != null && env.APPLE_APP_SPECIFIC_PASSWORD != null
         run: packaging/macos/sign_notarize_staple.sh build/*.dmg
+        timeout-minutes: 5
```
Swiftb0y, line 373: not sure, but that might a little conservative. Apple might take much longer. I guess we can keep it low for now and increase it if it turns out to be a problem. I think the better solution would be to terminate the script early if we received a status that is known to be final (such as "invalid").

tools/deploy.py
```diff
@@ -277,6 +272,13 @@ def generate_manifest(args):
         except IOError:
             pass
 
+    if args.update:
+        manifest_data = remote_manifest_data.update(manifest_data)
```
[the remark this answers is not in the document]
Holzhaus: oof that's embarrassing. you're right. In my defense, I wrote that code on my phone while sitting in the tram on my way to work :sweat_smile:

.github/workflows/build.yml
Holzhaus: I looked up how long it took last time it worked, it was ~3min. But yeah, it might take longer. Let's keep it that low for testing, and whoever actually fixes the notarization failure can set it to some reasonable value.

Holzhaus:
> Can you describe the new behaviour and the old?

Before: if one or multiple jobs of the build workflow fail, no manifest is uploaded. Instead, the old manifest stays as-is. Hence, the failing macOS job prevent the manifest from being updated and we present old downloads even for Windows although we actually have a more recent artifact for that OS.

New behavior:
We always run the manifest generation even if jobs in the build workflow fail. The resulting manifest initally only contains entries from jobs that succeeded. Then we download the current manifest from the server and copy over those entries that are missing in the new manifest.

> For my understanding we want a manifest only if the upload succeeds as well.Does that still works?

Yes. If the upload fails, the job fails. I did not touch this.

> Do we perform an upload if signing fails now? 

No..

> Does this ends up in unsigned binaries without someone notice it?

No.

tools/deploy.py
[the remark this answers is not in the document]
Swiftb0y: No worries, thats what code review is for :wink:

.github/workflows/build.yml
```diff
@@ -477,6 +481,7 @@ jobs:
     name: "Update manifest file on download server"
     runs-on: ubuntu-latest
     needs: build
+    if: always()
```
Swiftb0y, line 484: this needs a comment, why is this necessary?

tools/deploy.py
```diff
@@ -260,22 +257,27 @@ def generate_manifest(args):
...
-        except IOError:
-            pass
+            remote_manifest_data = url_download_json(manifest_url)
+        except IOError as e:
+            print("Fetching remote manifest failed: %s", e.msg)
+
+    if args.update:
+        for key, value in remote_manifest_data.items():
+            if key not in manifest_data:
+                manifest_data[key] = value
```
Swiftb0y, line 276: this assumes the remote_manifest contains data, can we reasonably expect this all the time? Why did you decide against `remote_manifest_data.update()`?

Holzhaus: If it doesn't it's an empty dict so that the entire `for` loop is a no-op.

.github/workflows/build.yml
Holzhaus: Because otherwise it will not be executed if any of the previous `build` jobs fail. See https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#example-not-requiring-successful-dependent-jobs.

Holzhaus: Comment added.

Swiftb0y: Thank you, I figured it was something like that. Thank you for the comment. It should prevent people removing the `if: always()` as part of a drive-by cleanup.

Holzhaus:
This is ready now.

Swiftb0y (approved):
LGTM. Thank you

tools/deploy.py
Swiftb0y: Ah right, misinterpreted a previous error. Of course an empty loop will do nothing.

PR #12093, opened 2023-10-12
[DRAFT] fix for tableview drawing performance

m0dB:
Fixes library tableview drawing performance issues which holds up the Qt event loop, causing frame drops in the waveforms. (https://github.com/mixxxdj/mixxx/issues/11603)

Profiling (on macOS) pointed to QStyledItemDelegate.

This PR derives TableItemDelegate from QItemDelegate instead of from QStyledItemDelegate. The text is manually with painter->drawText(..) (and without eliding, which makes a big difference). 

BPMDelegate is still TODO (not drawing checkbox)

To discuss: how important is styling for the tableview?

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -416,7 +416,9 @@ QAbstractItemDelegate* BaseTrackTableModel::delegateForColumn(
                 &BaseTrackTableModel::slotRefreshCoverRows);
         return pCoverArtDelegate;
     }
-    return nullptr;
+    // Explicitly return a TableItemDelegate, otherwise
+    // the default QStyleItemDelegate will be used.
+    return new TableItemDelegate(pTableView);
 }
 
 QVariant BaseTrackTableModel::data(
--- a/src/library/bpmdelegate.cpp
+++ b/src/library/bpmdelegate.cpp
@@ -81,14 +81,21 @@ void BPMDelegate::paintItem(QPainter* painter,const QStyleOptionViewItem &option
     // #LibraryBPMButton::indicator:unchecked {
     //  image: url(:/images/library/ic_library_unlocked.svg);
     // }
-    QStyleOptionViewItem opt = option;
-    initStyleOption(&opt, index);
 
-    if (m_pTableView != nullptr) {
-        QStyle* style = m_pTableView->style();
-        if (style != nullptr) {
-            style->drawControl(QStyle::CE_ItemViewItem, &opt, painter,
-                               m_pCheckBox);
-        }
-    }
+    // TODO: Draw the checkbox. Considering using the approach used with
+    // PreviewButtonDelegate and LibraryPreviewButton
+    //
+    // QStyleOptionViewItem opt = option;
+    // initStyleOption(&opt, index);
+    // if (m_pTableView != nullptr) {
+    //     QStyle* style = m_pTableView->style();
+    //     if (style != nullptr) {
+    //         style->drawControl(QStyle::CE_ItemViewItem, &opt, painter,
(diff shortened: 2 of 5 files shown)


ronso0:
Interesting...
So this breaks some general styles and the most of the custom delegates:
* bg color of unselected Preview button, location & Star delegate
* h-position of Star delegate
* 'selected' text color
* margin cuts off editor content (BPM spinbox, Stars)
* 'selected' cell frame is indented by margin

Still okay:
* comment delegate
* bg color of hovered Star delegate
* 'selected' Star color
* editor colors

image


> how important is styling for the tableview?

In general: very important I'd say. For example we can't use a static horizontal padding, else we'd break compact skins like Shade which are likely being used on low-res screens where every pixel counts (and restrict skin designers' options). It's currentl very convenient to do (almost) all styling with the skins' stylesheets.

m0dB:
Even more interesting: here it definitely doesn't look like this. The background colors are correct everywhere.

Ok, at least we know now where the problems come from. Not sure what is the way forward though. I personally would be more than happy with a style-less and optimized library view, but I fully understand that that's not an option (I wasn't aware skins also actively modified the library). Maybe it could be optional. I can try if I can still derive from QStyleItemDelegate without the performance penalty.

ronso0:
Ooouuuh stupid me...
For query testing purposes I set the **track color** to grey for ~98% of my tracks.
So, including track colors in my test, here are my

**Updated  findings:**
broken:
* track color bg default columns and BPM delegate
* h-position of Star delegate
* 'selected' text color of all comuns except Location
* margin cuts off editor content (BPM spinbox, Stars)
* 'selected' cell frame is indented by margin

still okay:
* bg color of hovered Star delegate (no track color)
* editor colors

image

src/library/basetracktablemodel.cpp
```diff
@@ -416,7 +416,9 @@ QAbstractItemDelegate* BaseTrackTableModel::delegateForColumn(
                 &BaseTrackTableModel::slotRefreshCoverRows);
         return pCoverArtDelegate;
     }
-    return nullptr;
+    // Explicitly return a TableItemDelegate, otherwise
+    // the default QStyleItemDelegate will be used.
```
daschuer, line 420: This comment does loose it's context without this PR. How about: 
```suggestion
    // Return the base class TableItemDelegate, for all other text based columns
```

src/library/basetracktablemodel.cpp
```diff
@@ -416,7 +416,9 @@ QAbstractItemDelegate* BaseTrackTableModel::delegateForColumn(
...
+    return new TableItemDelegate(pTableView);
```
daschuer, line 421: The ownership is not obvious here. No issue of this PR though. 
setItemDelegateForColumn() does not take the ownership: https://doc.qt.io/qt-6/qabstractitemview.html#setItemDelegateForColumn
Can we introduce a parented pointer?

src/library/basetracktablemodel.cpp
daschuer: Is it possible to get did of the downcast above? 
```
auto* pTableView = qobject_cast<WTrackTableView*>(pParent);
```

daschuer (commented):
Some code style comments. 

What is your plan with this PR?

m0dB:
> What is your plan with this PR?

From @ronso0 comments, it is clear that removing the styling is not an option. We could implement a browser that looks good (within the context of let's say Latenight) without any styling but I don't think that's worth the effort.

Replacing eliding with a fade-to-background overlay at the right of each column would lead to an improvement (not enough to avoid framedrops but should be noticable), but that should go in a different PR.

I am therefore closing the PR.

PR #12094, opened 2023-10-12
Add a debug message, which appears when event loop processing in Mixxx application takes very long

JoergAtGithub:


src/mixxxapplication.cpp
```diff
@@ -61,6 +61,8 @@ class QMouseEventEditable : public QMouseEvent {
 #endif
 };
 
+constexpr mixxx::Duration kEventNotifyExecTimeWarningThreshold = mixxx::Duration::fromMillis(20);
```
daschuer, line 64: What is the reasoning behind this 20 ms? Can you add a comment? A Waveform cycle is 1/60 s = 16.6 ms

src/mixxxapplication.cpp
```diff
@@ -168,7 +170,23 @@ bool MixxxApplication::notify(QObject* target, QEvent* event) {
...
     }
-    return QApplication::notify(target, event);
+
+    PerformanceTimer time;
+    time.start();
+
+    bool ret = QApplication::notify(target, event);
+
+    if (time.elapsed() > kEventNotifyExecTimeWarningThreshold) {
+        qDebug() << "Processing event type"
```
daschuer, line 180: This debug message makes this issue worse because of the extra time it consumes. 
Is it useful to have this active on the user machine, or should this be a compile option for debug builds only?

daschuer: Maybe with CmdlineArgs::Instance().getDeveloper() ?

JoergAtGithub: Done

src/mixxxapplication.cpp
JoergAtGithub: I adjusted the threshold to 5ms. This prints more messages on my system, but I think all of them are indicating visible performance issues.
After reducing the limit I see many warnings about slow processing of paint events in KnobComposed widgets. But these disapear, if I apply the fix from #12904 . So I think these are real and no false positives.

JoergAtGithub:
@daschuer A friendly Ping

src/mixxxapplication.cpp
```diff
@@ -181,10 +182,15 @@ bool MixxxApplication::notify(QObject* target, QEvent* event) {
 #endif
 
     PerformanceTimer time;
-    time.start();
+    bool isDeveloper = CmdlineArgs::Instance().getDeveloper();
```
ronso0, line 194: I'm curious: isquerying a bool for each event as fast as using a (const) bool member for this constant property?

JoergAtGithub: I cache this in a member variable now, which is initialized in the constructor.

Swiftb0y: > I'm curious: isquerying a bool for each event as fast as using a (const) bool member for this constant property?

If the entire thing is completely inlined, maybe. but caching makes sense here.

src/mixxxapplication.cpp
JoergAtGithub: I adjusted this again, to 10ms, as there are still many events in the 5ms range after fixing the pixmap cache.

daschuer (approved):
LGTM, Thank you.

daschuer:
Do you want to re-base it on 2.5 before merge? I have no strong demands.

JoergAtGithub:
It should be on 2.5 already

daschuer (approved):
LGTM, Thank you.

PR #12095, opened 2023-10-12
ci(packaging): Fail notarization if server returns "invalid" status

Holzhaus:
Otherwise it loops forever until the CI jobs runs into a timeout.


Swiftb0y (approved):
LGTM, thank you. the script doesn't run for this PR but it should be fine regardless.

Issue #12015, opened 2023-09-24
Shader Waveforms have filled lead in/out 


daschuer:
### Bug Description

Normally we have only the outlines of the triangles. With the GLSL types the triangles are filled. Not a real bug, but maybe it hides another issue: 
![grafik](https://github.com/mixxxdj/mixxx/assets/1777442/4631f0a6-c13e-4c19-b691-05f0752a9be6)
![grafik](https://github.com/mixxxdj/mixxx/assets/1777442/12d7783d-f7c0-41ae-ad6f-e904ac6924d4)

### Version

2.4-alpha

### OS

Ubuntu

daschuer:
@m0dB do you have an idea why this is the case?

JoergAtGithub:
This is an intended visual improvement. We discussed this in the PR review and decided to keep it this way.

m0dB:
To be fair, initially it was laziness, as drawing a filled triangle is much simpler than drawing only the outline. But I found I actually liked it much better, visually, so I kept it like this.

So, to put your mind at ease, @daschuer , it doesn't hide another issue.

daschuer:
Ok thank you for confirm.

Holzhaus:
IMHO waveforms should be consistent, so I'd like to keep this open until the other renders are adapted or the fill is removed.

m0dB:
Since @Holzhaus insisted, but also because I thought of an approach that is much nicer implementation-wise (a single triangle drawn into an texture which is then repeated pattern-wise in the shader instead of in c++), I have fixed this is https://github.com/mixxxdj/mixxx/pull/12100

m0dB:
<img width="340" alt="Screenshot 2023-10-13 at 17 27 28" src="https://github.com/mixxxdj/mixxx/assets/79429057/304334ce-5bc9-4a99-a51c-96e85aec80f9">

m0dB:
@Holzhaus after some discussion, I went with a slightly different look then legacy, using a semi-transparent fill. So, not 100% "consistent" as you requested, but given the 50% transparency of the fill, and 100% match of the outline, I would say at least 50% consistent ;-)

Check out the PR for some screenshots.

PR #12100, opened 2023-10-13
improve glsl pre-roll triangles

m0dB:
draw the glsl preroll and postroll markers with a patterned texture and some minor additional cleanups

fixes https://github.com/mixxxdj/mixxx/issues/12015 , initially bringing GLSL waveforms in line with the legacy waveforms, but after discussion going for a slightly different look (closer to the legacy waveform look but with a semi-transparent filling)

before:
Screenshot 2023-10-13 at 19 49 49

initial after
Screenshot 2023-10-13 at 19 48 36

final after (with this PR)
Screenshot 2023-10-15 at 17 27 56

### DIFF
--- a/src/shaders/patternshader.cpp
+++ b/src/shaders/patternshader.cpp
@@ -1,0 +1,35 @@
+#include "shaders/patternshader.h"
+
+using namespace mixxx;
+
+void PatternShader::init() {
+    QString vertexShaderCode = QStringLiteral(R"--(
+uniform highp mat4 matrix;
+attribute highp vec4 position;
+attribute highp vec2 texcoor;
+varying highp vec2 vTexcoor;
+void main()
+{
+    vTexcoor = texcoor;
+    gl_Position = matrix * position;
+}
+)--");
+
+    QString fragmentShaderCode = QStringLiteral(R"--(
+uniform sampler2D sampler;
+uniform vec2 repetitions;
+varying highp vec2 vTexcoor;
+void main()
+{
+    gl_FragColor = texture2D(sampler, fract(vTexcoor * repetitions));
+}
+)--");
+
+    load(vertexShaderCode, fragmentShaderCode);
+
+    m_matrixLocation = uniformLocation("matrix");
+    m_samplerLocation = uniformLocation("sampler");
+    m_repetitionsLocation = uniformLocation("repetitions");
+    m_positionLocation = attributeLocation("position");
+    m_texcoordLocation = attributeLocation("texcoor");
+}
--- a/src/shaders/patternshader.cpp
+++ b/src/shaders/patternshader.cpp
@@ -5,7 +5,7 @@ using namespace mixxx;
 void PatternShader::init() {
     QString vertexShaderCode = QStringLiteral(R"--(
 uniform highp mat4 matrix;
-attribute highp vec4 position;
+attribute highp vec4 position; // use vec4 here (will be padded) for matrix multiplication
 attribute highp vec2 texcoor;
 varying highp vec2 vTexcoor;
 void main()
--- a/src/shaders/patternshader.cpp
+++ b/src/shaders/patternshader.cpp
@@ -16,20 +16,20 @@ void main()
 )--");
 
     QString fragmentShaderCode = QStringLiteral(R"--(
-uniform sampler2D sampler;
+uniform sampler2D texture;
 uniform vec2 repetitions;
 varying highp vec2 vTexcoor;
 void main()
 {
-    gl_FragColor = texture2D(sampler, fract(vTexcoor * repetitions));
+    gl_FragColor = texture2D(texture, fract(vTexcoor * repetitions));
 }
 )--");
 
     load(vertexShaderCode, fragmentShaderCode);
 
     m_matrixLocation = uniformLocation("matrix");
-    m_samplerLocation = uniformLocation("sampler");
-    m_repetitionsLocation = uniformLocation("repetitions");
     m_positionLocation = attributeLocation("position");
     m_texcoordLocation = attributeLocation("texcoor");
+    m_textureLocation = uniformLocation("texture");
+    m_repetitionsLocation = uniformLocation("repetitions");
 }
--- a/src/shaders/vinylqualityshader.cpp
+++ b/src/shaders/vinylqualityshader.cpp
@@ -6,30 +6,30 @@ void VinylQualityShader::init() {
     QString vertexShaderCode = QStringLiteral(R"--(
 uniform mat4 matrix;
 attribute highp vec4 position;
-attribute highp vec3 texcoor;
-varying highp vec3 vTexcoor;
+attribute highp vec2 texcoord;
+varying highp vec2 vTexcoord;
 void main()
 {
-    vTexcoor = texcoor;
+    vTexcoord = texcoord;
     gl_Position = matrix * position;
 }
 )--");
 
     QString fragmentShaderCode = QStringLiteral(R"--(
-uniform sampler2D sampler;
-uniform highp vec4 color;
-varying highp vec3 vTexcoor;
+uniform sampler2D texture;
+uniform highp vec3 color;
+varying highp vec2 vTexcoord;
 void main()
 {
-    gl_FragColor = vec4(color.xyz, texture2D(sampler, vTexcoor.xy) * 0.75);
+    gl_FragColor = vec4(color, texture2D(texture, vTexcoord) * 0.75);
 }
 )--");
 
     load(vertexShaderCode, fragmentShaderCode);
 
     m_matrixLocation = uniformLocation("matrix");
-    m_samplerLocation = uniformLocation("sampler");
-    m_colorLocation = uniformLocation("color");
     m_positionLocation = attributeLocation("position");
-    m_texcoordLocation = attributeLocation("texcoor");
+    m_texcoordLocation = attributeLocation("texcoord");
+    m_textureLocation = uniformLocation("texture");
+    m_colorLocation = uniformLocation("color");
 }
--- a/src/shaders/vinylqualityshader.cpp
+++ b/src/shaders/vinylqualityshader.cpp
@@ -4,7 +4,7 @@ using namespace mixxx;
 
 void VinylQualityShader::init() {
     QString vertexShaderCode = QStringLiteral(R"--(
-uniform mat4 matrix;
+uniform highp mat4 matrix;
 attribute highp vec4 position; // use vec4 here (will be padded) for matrix multiplication
 attribute highp vec2 texcoord;
 varying highp vec2 vTexcoord;
@@ -17,11 +17,11 @@ void main()
 
     QString fragmentShaderCode = QStringLiteral(R"--(
 uniform sampler2D texture;
-uniform highp vec3 color;
+uniform highp vec4 color;
 varying highp vec2 vTexcoord;
 void main()
 {
-    gl_FragColor = vec4(color, texture2D(texture, vTexcoord) * 0.75);
+    gl_FragColor = vec4(color.xyz, texture2D(texture, vTexcoord) * color.w);
 }
 )--");
 
--- a/src/shaders/vinylqualityshader.cpp
+++ b/src/shaders/vinylqualityshader.cpp
@@ -5,7 +5,7 @@ using namespace mixxx;
 void VinylQualityShader::init() {
     QString vertexShaderCode = QStringLiteral(R"--(
 uniform mat4 matrix;
-attribute highp vec4 position;
+attribute highp vec4 position; // use vec4 here (will be padded) for matrix multiplication
 attribute highp vec2 texcoord;
 varying highp vec2 vTexcoord;
 void main()
--- a/src/util/texture.cpp
+++ b/src/util/texture.cpp
@@ -1,10 +1,10 @@
 #include "util/texture.h"
 
-QOpenGLTexture* createTexture(const QImage& image) {
+std::unique_ptr<QOpenGLTexture> createTexture(const QImage& image) {
     if (image.isNull()) {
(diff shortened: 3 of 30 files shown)


Swiftb0y (commented):
looks good so far, no complaints from me, except that this is just the shader code atm with the code for the pre-roll still missing.

m0dB:
oops! pushed!

ronso0:
could your please add before / after images to the description?

m0dB:
> could your please add before / after images to the description?

done!

(i personally like the filled rectangles better btw)

ronso0:
> (i personally like the filled rectangles better btw)

Me too. The translucent fill is a nice compromise :+1:

src/waveform/renderers/allshader/waveformrendererpreroll.cpp
```diff
@@ -101,45 +127,98 @@ void WaveformRendererPreroll::paintGL() {
...
+    const float imageW = static_cast<float>(imagePixelW) / devicePixelRatio;
+    const float imageH = static_cast<float>(imagePixelH) / devicePixelRatio;
+
+    QImage image(imagePixelW, imagePixelH, QImage::Format_ARGB32_Premultiplied);
+    image.setDevicePixelRatio(devicePixelRatio);
+
+    const float penWidth = 1.5f;
+    const float offset = penWidth / 2.f;
+
+    image.fill(QColor(0, 0, 0, 0).rgba());
```
Swiftb0y, line 194: Does this work? Should avoid a temporary QColor construction.
```suggestion
    image.fill(Qt::transparent);
```

Swiftb0y (changes requested):
Thank you. code looks in good shape, mostly minor suggestions. I really like the look of the translucent anti-aliased triangles. Reminds me of the Vital synth.

src/shaders/patternshader.cpp
```diff
@@ -0,0 +1,35 @@
...
+varying highp vec2 vTexcoor;
+void main()
+{
+    vTexcoor = texcoor;
+    gl_Position = matrix * position;
+}
+)--");
+
+    QString fragmentShaderCode = QStringLiteral(R"--(
+uniform sampler2D sampler;
```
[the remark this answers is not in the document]
m0dB: actually, I think I prefer ```sampler2D texture```, here and elsewhere.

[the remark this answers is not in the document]
Swiftb0y: also works. though in this case this is the pattern shader, so I'd argue it makes more sense. But anything is better than `sampler` imo.

src/shaders/patternshader.cpp
```diff
@@ -0,0 +1,35 @@
+#include "shaders/patternshader.h"
+
+using namespace mixxx;
+
+void PatternShader::init() {
+    QString vertexShaderCode = QStringLiteral(R"--(
+uniform highp mat4 matrix;
+attribute highp vec4 position;
```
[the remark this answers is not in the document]
m0dB: yes, because it is being multiplied with a mat4... which indeed could also be a mat2. i will do some additional cleanup of the shaders, while at it.

src/waveform/renderers/allshader/waveformrendererpreroll.cpp
```diff
@@ -101,45 +127,98 @@ void WaveformRendererPreroll::paintGL() {
...
 
     glDrawArrays(GL_TRIANGLES, 0, m_vertices.size());
+}
 
-    m_shader.disableAttributeArray(vertexLocation);
-    m_shader.release();
+void WaveformRendererPreroll::generateTexture(float markerLength, float markerBreadth) {
+    const float devicePixelRatio = m_waveformRenderer->getDevicePixelRatio();
+    m_markerLength = markerLength;
+    m_markerBreadth = markerBreadth;
```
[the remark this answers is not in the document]
m0dB: agreed, will change

src/shaders/patternshader.cpp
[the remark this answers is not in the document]
m0dB: of course the day we want waveforms projected on rotating 3d cubes will will have to revert this ;-)

src/waveform/renderers/allshader/waveformrendererpreroll.h
```diff
@@ -17,15 +19,23 @@ class WaveformRendererPreroll;
...
+    void drawPattern(float x1, float y1, float x2, float y2, float repetitions, bool flip);
+    void generateTexture(float markerLength, float markerBreadth);
+
+    mixxx::PatternShader m_shader;
     QColor m_color;
+    float m_markerBreadth{};
+    float m_markerLength{};
+    std::unique_ptr<QOpenGLTexture> m_pTexture;
     VertexData m_vertices;
+    VertexData m_texcoords;
```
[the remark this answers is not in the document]
m0dB: the only reason was to use the VertexData::addRectangle to populate the vertex values, and making them members it to reuse the memory (which is not particularly relevant here, but in line with the other renderers).

not very strong reasons, so I will change this.

src/shaders/patternshader.cpp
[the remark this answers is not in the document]
m0dB: i will do a bit of type cleanup though

src/waveform/renderers/allshader/waveformrendererpreroll.cpp
m0dB: that works

m0dB:
@Swiftb0y I have addressed all comments, please mark the conversations as resolved when re-reviewed.

src/shaders/patternshader.cpp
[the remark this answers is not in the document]
Swiftb0y: > there is no advantage in doing the conversion from vec2 to vec4 manually, and doing it like this keeps the shader more flexible (usuable for 3d, which is of course not our usecase, but since we get it for free, why cripple the shader?)

Sure, makes sense. the vec3 is probably padded to 4 anyways. This just needs a comment then.

src/waveform/renderers/allshader/waveformrendererpreroll.h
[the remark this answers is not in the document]
Swiftb0y: Being able to use `VertexData::addRectangle` is a valid reason. Reusing the memory is also valid but unnecessary if the data could just be stack-local.

src/widget/wspinnyglsl.cpp
```diff
@@ -97,17 +97,17 @@ void WSpinnyGLSL::paintGL() {
     m_textureShader.bind();
 
     int matrixLocation = m_textureShader.matrixLocation();
-    int samplerLocation = m_textureShader.samplerLocation();
+    int textureLocation = m_textureShader.textureLocation();
     int positionLocation = m_textureShader.positionLocation();
     int texcoordLocation = m_textureShader.texcoordLocation();
 
-    QMatrix4x4 matrix;
+    QMatrix matrix;
```
Swiftb0y, line 104: `QMatrix` is deprecated in Qt5 and removed in Qt6, why did you change to it?

src/waveform/renderers/allshader/waveformrendererpreroll.cpp
```diff
@@ -12,6 +12,53 @@
 #include "widget/wskincolor.h"
 #include "widget/wwidget.h"
 
+namespace {
+QOpenGLTexture* generateTexture(float markerLength,
```
Swiftb0y, line 16: I'm sorry for expanding the scope yet again, but this needs a little more supporting refactoring. The pointer returned here (and thus by createTexture) is owning. We should really avoid those in new code. Can you refactor this and `createTexture` to return by value or `unique_ptr` (preferably the former).

Swiftb0y:
> please mark the conversations as resolved when re-reviewed.

Did so and also reviewed the new code. Thank you.

src/widget/wspinnyglsl.cpp
m0dB: good question. i'll change it back.

src/waveform/renderers/allshader/waveformrendererpreroll.cpp
m0dB: sure. i will use a unique_ptr though

Swiftb0y: Sure... seems a little unnecessary, but I won't insist.

src/shaders/vinylqualityshader.cpp
```diff
@@ -5,7 +5,7 @@ using namespace mixxx;
 void VinylQualityShader::init() {
     QString vertexShaderCode = QStringLiteral(R"--(
 uniform mat4 matrix;
-attribute highp vec4 position;
+attribute highp vec4 position; // use vec4 here (will be padded) for matrix multiplication
 attribute highp vec2 texcoord;
 varying highp vec2 vTexcoord;
 void main()
```
Swiftb0y, line 11: we should really deduplicate some of these shaders, but thats out-of-scope.

src/util/texture.cpp
```diff
@@ -1,31 +1,31 @@
 #include "util/texture.h"
 
-QOpenGLTexture* createTexture(const QImage& image) {
+std::unique_ptr<QOpenGLTexture> createTexture(const QImage& image) {
     if (image.isNull()) {
-        return nullptr;
+        return std::unique_ptr<QOpenGLTexture>{};
```
Swiftb0y, line 5: why do you prefer this over just returning `nullptr`?

src/util/texture.cpp
```diff
@@ -1,31 +1,31 @@
...
     }
-    QOpenGLTexture* pTexture = new QOpenGLTexture(image);
+    std::unique_ptr<QOpenGLTexture> pTexture{new QOpenGLTexture(image)};
```
Swiftb0y, line 7: We use `make_unique` everywhere. https://stackoverflow.com/a/22571331
```suggestion
    auto pTexture = std::make_unique<QOpenGLTexture>(image);
```

Swiftb0y (commented):
Thank you, small nitpicks.

src/util/texture.cpp
m0dB: I like to be explicit, but I'm okay with changing that if you commonly return nullptr elsewhere.

src/util/texture.cpp
m0dB: Thought we were restricted to c++11

Swiftb0y: Fortunately not. 2.4 is restricted to gcc10, so C++17 is fully okay. On main, you can even use C++20.

src/util/texture.cpp
Swiftb0y: I'd prefer that.

m0dB: ok, changed back. i think with that I addressed all your comments.

Swiftb0y (approved):
Yes, LGTM. Thank you very much. Highly appreciated. The anti-aliasing is a huuge improvement. Any objections @Holzhaus?
CI failures is unrelated, IIUC @daschuer is looking into it.

src/widget/wspinnyglsl.cpp
```diff
@@ -138,15 +138,15 @@ void WSpinnyGLSL::paintGL() {
...
+        rotate.rotate(-m_fGhostAngle, 0, 0);
         m_textureShader.setUniformValue(matrixLocation, rotate);
 
         drawTexture(m_pGhostTextureScaled.get());
     }
 
     if (m_pFgTextureScaled) {
         QMatrix4x4 rotate;
-        rotate.rotate(m_fAngle, 0, 0, -1);
+        rotate.rotate(-m_fAngle, 0, 0);
```
JoergAtGithub, line 149:
```suggestion
        rotate.rotate(m_fAngle, 0, 0, -1);
```
Same here

JoergAtGithub (changes requested):

JoergAtGithub:
I just tested this PR with vinyl control signal and the spinnies show it without color (just black):
grafik

m0dB:
@JoergAtGithub Aargh, the only think I forgot to test myself! Thanks, I will fix this now.

m0dB:
@JoergAtGithub I have fixed the vinyl quality shader. If you confirm it works, this is ready for merge.

Holzhaus:
> Yes, LGTM. Thank you very much. Highly appreciated. The anti-aliasing is a huuge improvement. Any objections @Holzhaus?

Looks nice. My only objection regarding the previous look was that Software/GL/GLSL/Shader waveform rendering should be consistent. Is this the case now.

m0dB:
With 50% transparency, it is 50% more consistent ;-)

At least it is now clearly deliberate and can't be mistaken for a bug. So IMO it is "consistent" enough and sacrificing a bit of consistency for a nicer look makes sense. 

We could try to change the legacy waveforms to match this look, but I don't think that's worth the effort, particularly taking into account that there is already a lot of variation between the different waveform types.

JoergAtGithub:
The color of the Vinylcontrol signal is now back. 

I'm not sure if the following is a regression, or was always the case, but when I drop the needle, I see the signal outside of the round area:

https://github.com/mixxxdj/mixxx/assets/64457745/f7a308f9-c952-41e5-a869-a45cd5894365

Maybe it's just me, but I expected this to be masked as the Cover-Art.

m0dB:
> Maybe it's just me, but I expected this to be masked as the Cover-Art.

That would be as simple as swapping the draw order of the mask and the vinyl quality. But I took this order from the original spinny code, and changing this should not be part of this PR.

Merge?

JoergAtGithub:
Since it's not a regression, let's merge this as it is!

Swiftb0y:
Thank for this nice visual improvement @m0dB. Thank you for the thorough testing @JoergAtGithub.

PR #4774, opened 2022-05-30
macOS packaging: Perform ad-hoc signing of macOS bundle by default

fwcd:
Starting with arm64 macOS Apple will require ad-hoc signatures, completely unsigned app bundles will no longer launch:

Screen Shot 2022-05-29 at 17 06 37

This branch fixes the issue by setting an ad-hoc signature, which only includes a checksum, by default if no other code signing identity is provided. With this patch, Mixxx launches normally again:

Screen Shot 2022-05-30 at 14 16 29

While arm64 macOS isn't officially supported by Mixxx yet, setting an ad-hoc signature on x86 too probably cannot hurt and fixing discrepancies between the x86 and arm64 builds in in advance is IMO always a good idea.

Marked as a draft PR for now since I haven't tested it with the x86 build yet.

PR #12101, opened 2023-10-13
macOS packaging: Enable app sandbox in ad-hoc-packaged (i.e. non-notarized) bundles too

### LINKED ISSUE #12098: "Mixxx.app" would like to access data from other apps prompt on macOS Sonoma
### Bug Description

image

According to [this blog post](https://lapcatsoftware.com/articles/2023/6/1.html) and [this WWDC23 session](https://developer.apple.com/videos/play/wwdc2023/10053/) (around 17:30), this happens if applications try to access other apps' sandbox containers (`~/Library/Containers/...`). We should figure out where that happens since, if the user hasn't added that folder to their music library, there should be no need to access anything in `~/Library/Containers` outside of the Mixxx container.

Note that this only seems to occur in Ad-Hoc-signed builds, notarized builds are fine.

### Version

2.5-alpha-131-gffe0ddd6eb

### OS

macOS 14.0

fwcd:
In #4774 and specifically e9223294c5c5af9d633431b591ebad9b575ac54d we set up ad-hoc signing for (non-notarized) macOS builds in a way that neither uses the hardened runtime nor app sandbox.

Apparently, macOS Sonoma will prompt the user at every launch when a non-sandboxed app accesses a sandboxed path, even of the app itself (`~/Library/Containers/org.mixxx.mixxx/...`). See #12098 and [this blog post](https://lapcatsoftware.com/articles/2023/6/1.html) for details.

For this reason, this PR passes the entitlements, and thus enables app sandbox, in ad-hoc-signed builds too (note that we'll still leave the hardened runtime disabled for now), which should fix the issue.

### DIFF
--- a/cmake/modules/BundleInstall.cmake.in
+++ b/cmake/modules/BundleInstall.cmake.in
@@ -23,6 +23,7 @@ if(DEFINED APPLE_CODESIGN_IDENTITY AND DEFINED APPLE_CODESIGN_ENTITLEMENTS)
       message(STATUS "Ad-hoc signing bundle without hardened runtime")
       execute_process(COMMAND
           codesign --verbose=4 --deep --force
+          --entitlements "${APPLE_CODESIGN_ENTITLEMENTS}"
           --sign "${APPLE_CODESIGN_IDENTITY}"
           "${PATH_TO_SIGN}"
       )


fwcd:
Note that users will now get this warning instead:

Screenshot 2023-10-13 at 18 26 32

This warning, however, only shows up once (as opposed to the dialog discussed in #12098).

fwcd:
Note that the standard Gatekeeper workflow of having to double-click the app once to get

Screenshot 2023-10-13 at 18 32 12

and *then* having to right-click -> Open to get

Screenshot 2023-10-13 at 18 32 20

...still applies. Nothing special to see here, this is how macOS treats every non-notarized app.

daschuer (approved):
This looks reasonable. Thank you.

PR #12102, opened 2023-10-13
Add [Main] to the list of valid groups for HID.

ywwg:
Fixes issues with calls to resolveGroup returning undefined

### DIFF
--- a/res/controllers/common-hid-packet-parser.js
+++ b/res/controllers/common-hid-packet-parser.js
@@ -1240,6 +1240,7 @@ class HIDController {
          * @type {string[]}
          */
         this.valid_groups = [
+            "[App]",
             "[Channel1]",
             "[Channel2]",
             "[Channel3]",
--- a/res/controllers/common-hid-packet-parser.js
+++ b/res/controllers/common-hid-packet-parser.js
@@ -1252,6 +1252,7 @@ class HIDController {
             "[Sampler6]",
             "[Sampler7]",
             "[Sampler8]",
+            "[Main]",
             "[Master]",
             "[PreviewDeck1]",
             "[Effects]",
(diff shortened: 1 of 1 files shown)


res/controllers/common-hid-packet-parser.js
```diff
@@ -1252,6 +1252,7 @@ class HIDController {
             "[Sampler6]",
             "[Sampler7]",
             "[Sampler8]",
+            "[Main]",
```
Swiftb0y, line 1255: what group is this again? we also need to add `[App]` IIRC.

ywwg: There seem to have been recent changes that moved, say, vu_meter_l to the "[Main]" group. This list needs to be kept up to date or HID controllers stop working.

ywwg: Added [App]

JoergAtGithub:
I think we have also more valid control group names:

- [Skin]
- [Sampler9...64]

@Holzhaus Could you have a look, if this list is complete?

ywwg:
I wonder if we can get rid of this list completely.  it's only used in two places:

1: inside resolveGroup to decide if the group being resolved is already known.  But all this function does is resolve "deck1" and "deck2" to the appropriate channel.  As written, otherwise it returns the group as-is.  So we could just replace the whole function with a test for deck1 and deck2.  There is an indirect debugging effect that it returns undefined if the group is unknown, but in practice any typoed group name will be caught elsewhere with a missed engine connection or otherwise.

2: inside getActiveFieldGroup, where it is essentially doing the exact same thing -- calling resolvegroup.


So I am going to get rid of this list and see what happens

JoergAtGithub:
Code LGTM! Did vou check that no mapping uses this public list directly?

daschuer:
This seems not the case. LGTM.

daschuer (approved):

Issue #12086, opened 2023-10-11
Change Rate by editing the displayed BPM value


ghost:
### Feature Description

Hi

After this forum post, I had the idea of a feature. 

https://mixxx.discourse.group/t/set-bpm-counter-to-specific-value/28160

Set the BPM counter to specific value, while the track is loaded/playing.

The user could click on the BPM counter, enter a value for the BPM and Mixxx should adjust the pitch % to match the chosen BPM value.
The goal of this feature is to quickly set a track to a chosen BPM value, instead of using the pitch slider to do this.

![afbeelding](https://github.com/mixxxdj/mixxx/assets/147563220/1bbee0f6-f164-4359-960b-a3f4a6820e47)

m0dB:
I think this has a good implement-dificulty / usefulness ratio. @ronso0 what do you think? This could be as simple as creating a QDoubleSpinBox derived widget, called for example BpmSpinbox, with setButtonSymbols(ButtonSymbols::NoButtons), a range between 0 and 240 (?) and 2 decimals, and add it to the legacy skin parser. Change the type of the bpm value from Number to BpmSpinBox, and done, right? Or am I missing something?

deejay-girl:
Thanks for liking my feature!
I would prefer no limit of 240 because tracker music or hardcore/gabber can reach this limit. If you want to speed it up, you can certainly exceed it. Maybe it's possible to have no limit?

ronso0:
> I think this has a good implementation-dificulty / usefulness ratio. @ronso0 what do you think? This could be as simple as creating a QDoubleSpinBox derived widget, called for example BpmSpinbox, with setButtonSymbols(ButtonSymbols::NoButtons), a range between 0 and 240 (?) and 2 decimals, and add it to the legacy skin parser. Change the type of the bpm value from Number to BpmSpinBox, and done, right? Or am I missing something?

Sure. I'm sure we'll find a good implementation (code and UX) for a isolated BPM display/editor. What I'm worried about is that there is also the (transparent) Tap button on top that changes the (track) BPM, not the (playback) rate.
Admittedly, I never understood why this a track edit tool was placed in the rate control section in the first place.

Maybe this drives the discussion off topic, but it's definitely related:
How about making the BPM display a multi-task widget:
* on click: highlight border
* first click: show 'edit' icon in the top-right corner (pen icon)
  * click on edit icon: show BPM spinbox
* consecutive clicks:
  * (hide edit icon)
[...]

PR #3890, opened 2021-05-22
WScrollable: Add a scrollable widget type. 

ywwg:
This allows skins to add scrollbars to areas.

Not used in any current skins, but tested with my raspi branch. It would be simple
to add support for controlling whether horizontal and/or vertical scrollbars are used,
but for now I'd like to leave it on auto.  (https://martinfowler.com/bliki/Yagni.html).

PR #12104, opened 2023-10-14
add Rate Tap button

ronso0:
* add `tempo_tap` which works exactly like BPM tap, just it sets the tempo
* change BPM display overlay and buttons in skins to do
  * **tempo** tap on left click
  * BPM tap on right-click.

quoting myself from #12086
> Admittedly, I never understood why this a track edit tool [BPM tap button] was placed in the rate control section in the first place.

> double-clicking any track (text) label opens the Track Properties editor where the (track) BPM can be set (type or tap)
Use case

### Reasoning
BPM tap edits a track property (permanently) while the rate section is solely about controlling the playback speed (+, -, sync). Hence, IMO BPM tap doesn't belong there.
IIUC the intention for putting it there was to allow _setting_ the track BPM, i.e. set initial value or correct BPM analysis. I assume it was helpful in the early days of Mixxx but with the improved beats analyzer nowadays I suppose it's not needed (that much) anymore. Manual correction can be done in Track Properties > BPM tab.

### My use case
Infrequently I need to "Sync" to external players (vinyl via Aux, other DJ in b2b sessions), i.e. at least estimate the target BPM. Until now I was using a Python BPM counter [1], activated with a global hotkey, which gives very accurate results. However, it'd be much more convenient to do this within Mixxx.
Either I apply the guessed target BPM (e.g. 128) to loaded tracks, or I do a bpm search `bpm:125-130`.
Rate Tap is a good step forward.

Thinking this further, a BPM tap utility might be helpful, see #12105 

- 1 https://github.com/m13253/bpm-counter/blob/master/trunk/bpm-counter.py

### DIFF
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -319,6 +319,7 @@ void BpmControl::slotBpmTapFilter(double averageLength, int numSamples) {
     }
     pTrack->trySetBeats(*newBeats);
 }
+
 void BpmControl::slotTempoTap(double v) {
     if (v > 0) {
         m_tempoTapFilter.tap();
@@ -1005,16 +1006,17 @@ mixxx::audio::FrameDiff_t BpmControl::getPhaseOffset(mixxx::audio::FramePos this
     return position - thisPosition;
 }
 
-void BpmControl::slotUpdateEngineBpm(double value) {
+void BpmControl::slotUpdateEngineBpm(double rateRatio) {
     // Adjust playback bpm in response to a rate_ratio update
-
     if (kLogger.traceEnabled()) {
-        kLogger.trace() << getGroup() << "BpmControl::slotUpdateEngineBpm"
-                        << m_pLocalBpm->get() << value;
+        kLogger.trace() << getGroup() << "BpmControl::slotUpdateEngineBpm: local BPM:"
+                        << m_pLocalBpm->get()
+                        << "rate ratio:"
+                        << rateRatio;
         // This can be used to detect pitch shift issues with cloned decks
         // DEBUG_ASSERT(getGroup() != "[Channel1]" || m_pRateRatio->get(); == 1);
     }
-    m_pEngineBpm->set(m_pLocalBpm->get() * value);
+    m_pEngineBpm->set(m_pLocalBpm->get() * rateRatio);
 }
 
 void BpmControl::slotUpdateRateSlider(double value) {
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -51,9 +51,9 @@ BpmControl::BpmControl(const QString& group,
     m_dSyncTargetBeatDistance.setValue(0.0);
     m_dUserOffset.setValue(0.0);
 
-    m_pPlayButton = new ControlProxy(group, "play", this);
-    m_pReverseButton = new ControlProxy(group, "reverse", this);
-    m_pRateRatio = new ControlProxy(group, "rate_ratio", this);
+    m_pPlayButton = std::make_unique<ControlProxy>(group, "play", this);
+    m_pReverseButton = std::make_unique<ControlProxy>(group, "reverse", this);
+    m_pRateRatio = std::make_unique<ControlProxy>(group, "rate_ratio", this);
     m_pRateRatio->connectValueChanged(this, &BpmControl::slotUpdateEngineBpm,
                                       Qt::DirectConnection);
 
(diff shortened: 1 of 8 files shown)


deejay-girl:
Where can we download nightly build of Mixxx?
The 2.6 alpha development version on the website is never updated so for me it's impossible to test the latest changes :(
I can't compile Mixxx myself.

ronso0:
Getting alpha builds and testing Pull Requests is described on the wiki
https://github.com/mixxxdj/mixxx/wiki/Testing

ronso0:
Very weird: I built this locally and it works as expected.
Though, it was built with Qt5 because I forgot to set the Qt6 switch.
So with Qt6 [`slotUpdateRateSlider(double bpm)`](https://github.com/mixxxdj/mixxx/blob/0c2faf06a89e5aca14d1aa7df8e90f9c5b50c69c/src/engine/controls/bpmcontrol.cpp#L939) is not called when BPM are tapped, even though it should
https://github.com/mixxxdj/mixxx/blob/0c2faf06a89e5aca14d1aa7df8e90f9c5b50c69c/src/engine/controls/bpmcontrol.cpp#L101-L111

When I change `bpm` from dev tools it works, it just doesn't when `bpm` is set in `slotRateTapFilter` :shrug: 
It also works if I make `bpm` read-only and use `connectValueChangeRequest` and copy all required calls there.

Also, I have no idea why unrelated tests fail n all OS except ARM 64, for example https://github.com/mixxxdj/mixxx/actions/runs/6517384471/job/17707390983?pr=12104

I'd appreciate if someone could give me some hints what's wrong here.

ronso0:
I pushed a different (working) version with my findings added inline.

ronso0:
> Very weird: I built this locally and it works as expected. Though, it was built with Qt5 because I forgot to set the Qt6 switch. So with Qt6 [`slotUpdateRateSlider(double bpm)`](https://github.com/mixxxdj/mixxx/blob/0c2faf06a89e5aca14d1aa7df8e90f9c5b50c69c/src/engine/controls/bpmcontrol.cpp#L939) is not called when BPM are tapped, even though it should
...

Seems this is a known issue? Just stumbled upon https://github.com/mixxxdj/mixxx/blob/558253682acc5c07452f829209fefca91bd8ba8b/src/broadcast/broadcastmanager.cpp#L71-L78
Still wondering why the comment was written for Qt5 whereas it worked here with Qt5 and fails with Qt6 :shrug: 
I changed it to call the slot manually.

ronso0:
I changed all skin buttons to tap the tempo on left click and tap the BPM on right click.

ronso0:
I changed it to `tempo_tap` to make it more clear what it does.

Ready to roll!

src/engine/controls/bpmcontrol.h
```diff
@@ -123,51 +124,53 @@ class BpmControl : public EngineControl {
     friend class SyncControl;
 
     // ControlObjects that come from EngineBuffer
-    ControlProxy* m_pPlayButton;
+    std::unique_ptr<ControlProxy> m_pPlayButton;
     QAtomicInt m_oldPlayButton;
-    ControlProxy* m_pReverseButton;
-    ControlProxy* m_pRateRatio;
+    std::unique_ptr<ControlProxy> m_pReverseButton;
```
ronso0, line 129: note: can be PollingControlProxy

fwcd:
> I assume it was helpful in the early days of Mixxx but with the improved beats analyzer nowadays I suppose it's not needed (that much) anymore

I use the BPM tap pretty frequently on tracks where the analyzer still struggles (acapellas or tracks with complex polyrhythms, e.g. Salsa music). Would appreciate to at least keep the option to keep having BPM tap available in the UI, having to dig into the track properties makes this too inconvenient to really be useful IMHO.

ronso0:
> Moving the BPM tap to the beatgrid editing sections seems reasonable too, to untangle the BPM editing controls from rate changing.

Good idea!
Feasible in Tango and LateNight, not so nice in Deere but the BPM control section is already interfering anyway with the track dsiplays.

> perhaps a switch in the skin settings to toggle left/right-click behavior?

Hmm, the skin settings menu is growing and growing, in LateNight we already had to squeeze so it fits at minimum height.
At some point we need to wrap it into a WScrollable #3890 (still unused in official skins)

daschuer:
Can you move the first commit to a separate PR? 
This way we see here better the proposed functional changes.

ronso0:
Why not view by commit?

daschuer:
Now a conflict has developed.

ronso0:
> > Moving the BPM tap to the beatgrid editing sections seems reasonable too, to untangle the BPM editing controls from rate changing.

Just filed a feature request to track this.

> > perhaps a switch in the skin settings to toggle left/right-click behavior?
> 
> Hmm, the skin settings menu is growing and growing, in LateNight we already had to squeeze so it fits at minimum height. At some point we need to wrap it into a WScrollable #3890 (still unused in official skins)

I'll look into this during beta.

ronso0:
Conflicts are resolved.

daschuer:
Now the unit-test are failing.

src/engine/controls/bpmcontrol.cpp
```diff
@@ -99,61 +123,61 @@ BpmControl::BpmControl(const QString& group,
...
+    connect(&m_tempoTapFilter,
+            &TapFilter::tapped,
+            this,
+            &BpmControl::slotTempoTapFilter,
             Qt::DirectConnection);
 
-    m_pBpmLock = std::make_unique<ControlPushButton>(
-            ConfigKey(group, "bpmlock"), false);
-    m_pBpmLock->setButtonMode(ControlPushButton::TOGGLE);
-    m_pBpmLock->connectValueChangeRequest(
```
daschuer, line 136: because of this change m_pBpmLock is left null and crashes when used in trackBpmLockChanged

daschuer (changes requested):

daschuer:
Oh sorry, and new conflicts have developed.

ronso0:
> Oh sorry, and new conflicts have developed.

Yeah, a lot has happend in BpmControl since I started this.
I rebased and it looks & works good now, let's see what CI says.

ronso0:
small TODO:
highlight the tap button also on right-click. Not sure how to do that, though.

src/engine/controls/bpmcontrol.cpp
```diff
@@ -1007,15 +1007,14 @@ mixxx::audio::FrameDiff_t BpmControl::getPhaseOffset(mixxx::audio::FramePos this
 
 void BpmControl::slotUpdateEngineBpm(double value) {
     // Adjust playback bpm in response to a rate_ratio update
-    double dRate = m_pRateRatio->get();
```
Swiftb0y, line 1010: please update the name `value` to reflect the fact that its now the `rate`.

Swiftb0y (approved):
Only nitpick. rest LGTM. haven't tested though

daschuer (approved):

daschuer:
Thank you.

PR #12112, opened 2023-10-15
Effects: Add backend for Audio Unit (AU) plugins on macOS

fwcd:
This implements an effects backend for AU plugins on macOS. In other words, all of the stock AU plugins, along with any installed third-party Logic/GarageBand plugins can now conveniently be used directly from Mixxx's effects panel.

Screenshot 2023-10-15 at 03 39 06
Screenshot 2023-10-15 at 03 39 44

## Implementation Notes

This leverages the [AUv2 C API](https://developer.apple.com/documentation/audiotoolbox/audio_unit_v3_plug-ins/hosting_audio_unit_extensions_using_the_auv2_api?language=objc) for hosting Audio Units in a standalone fashion, without an `AUGraph`, `AVAudioEngine` or similar. The audio units are instantiated out-of-process (though this can easily be changed) and are passed audio data directly via the render callback mechanism ([documented in detail here](https://developer.apple.com/library/archive/documentation/MusicAudio/Conceptual/AudioUnitProgrammingGuide/AudioUnitDevelopmentFundamentals/AudioUnitDevelopmentFundamentals.html#//apple_ref/doc/uid/TP40003278-CH7-SW14)).

The rationale for not using the [AUv3 Objective-C API](https://developer.apple.com/documentation/audiotoolbox/audio_unit_v3_plug-ins/migrating_your_audio_unit_host_to_the_auv3_api?language=objc) is that the latter seems to be a bit more finicky to deal with in a standalone context (without `AUGraph`/`AVAudioEngine`), e.g. throwing `NoConnection` errors during rendering. Since the API is supposed to map 1-to-1 to AUv2, it should probably still be possible to express our use case with the AUv3 API, but since the AUv2 API is not deprecated (and also still widely used), we'll use that for now. Note that AUv3 plugins will work fine too, since the AUv2 and AUv3 APIs are bridged in both directions.

The (compiling, but non-functional) implementation using the AUv3 API can be found here:

- https://github.com/fwcd/mixxx/pull/5

## Link dump

Further reading and anything that may be of use to implementors:

- General:
  - https://developer.apple.com/library/archive/documentation/MusicAudio/Conceptual/AudioUnitHostingGuide_iOS/AudioUnitHostingFundamentals/AudioUnitHostingFundamentals.html
- AUv2:
  - https://developer.apple.com/documentation/audiotoolbox/audio_unit_v2_c_api?language=objc
  - https://developer.apple.com/documentation/audiotoolbox/audio_unit_v3_plug-ins/hosting_audio_unit_extensions_using_the_auv2_api?language=objc
  - https://stackoverflow.com/questions/56778707/does-augraph-deprecation-means-no-more-audio-render-callbacks
  - https://stackoverflow.com/questions/6735444/ios-audio-units-when-is-usage-of-augraphs-necessary
- AUv3:
  - https://developer.apple.com/documentation/audiotoolbox/audio_unit_v3_plug-ins
  - https://developer.apple.com/documentation/audiotoolbox/audio_unit_v3_plug-ins/migrating_your_audio_unit_host_to_the_auv3_api?language=objc
  - https://developer.apple.com/forums/thread/14458
  - https://github.com/hotpaw2/auv3test5

## Prior art

- Implementation of Audio Unit effects in Audacity (also uses the AUv2 API): https://github.com/audacity/audacity/blob/52708c770ffcd98cb86527cd7756413c4179f7fe/libraries/lib-audio-unit/AudioUnitInstance.cpp

## Caveats

The plugins currently have no UI. Might be a nice candidate for a future PR, but for now they have to be controlled via the parameter knobs.


deejay-girl:
I don't have a Mac but as I understand correctly on Windows those are called VSTs?
Would it be possible to have the same feature for VSTs?
I found online a Mac can run VSTs too.

fwcd:
Well, yes and no, AUs and VSTs are similar in that they are both audio plugins, but they're completely different implementations, so no, they aren't interchangeable.

In principle the effects framework provides a general interface that could be implemented for VSTs too ([see the wiki](https://github.com/mixxxdj/mixxx/wiki/Effects-Framework)). Since AUs use a completely different API, support for VSTs would have to be implemented "from scratch" (which is a fair bit of work too and definitely out-of-scope for this PR).

m0dB:
I will try to review this over the weekend.

src/effects/backends/audiounit/audiounitmanager.mm
```diff
@@ -0,0 +1,109 @@
...
+        AudioUnitUninitialize(m_audioUnit);
+        AudioComponentInstanceDispose(m_audioUnit);
+    }
+}
+
+AudioUnit _Nullable AudioUnitManager::getAudioUnit() {
+    // We need to load this atomic flag to ensure that we don't get a partial
+    // read of the audio unit pointer (probably extremely uncommon, but not
+    // impossible: https://belkadan.com/blog/2023/10/Implicity-Atomic)
+    if (!m_isInstantiated.load()) {
```
m0dB, line 44: Are you using an atomic instead of a mutex because this method is called from the realtime audio thread? If so, a comment mentioning that would be good.

m0dB:
This is really nice! The code looks good to me, but I would like to do a bit of testing before approving.

m0dB:
Unfortunately it isn't working here. I tried a bunch of the Apple default AudioUnits. The logs are outputting a stream of `warning [Engine] Rendering Audio Unit failed with status -10874` messages, and the audio output is stuck on a 1024 sample buffer.

Note that this is after rebasing, compilation of the branch as is failed in unrelated code.

fwcd:
I'll have a look, it's been a while since I touched this PR, might even have been before the Sonoma update

fwcd:
https://osstatus.com/search/results?platform=all&framework=all&search=-10874 says this is `kAudioUnitErr_TooManyFramesToProcess`. Maybe it needs a smaller buffer

src/effects/backends/audiounit/audiounitmanager.mm
fwcd: That was what I assumed, `getAudioUnit` is called from `processChannel` which operates directly on sample buffers, though it might be worth checking whether that is indeed the audio thread. Since that's a method on the hot path, we want to avoid blocking there.

fwcd:
The branch is now rebased onto the most recent `main`, so please reset any local clone. I can confirm that the Audio Unit effects currently only work if the audio buffer is <= 10.7 ms. We probably have to perform multiple rendering passes over fixed-sized chunks if the buffer is larger.

src/effects/backends/audiounit/audiounitmanager.mm
fwcd: Added a short note in 440d1cf4b4.

JoergAtGithub:
@m0dB Could you test this again?

m0dB:
I confirm it works fine with 11.6 ms @ 44,100 Hz .

I don't think it's a good idea to merge this before taking care of handling larger buffer sizes with multiple render passes. Did you look into implementing that, @fwcd? Or would you like me to give it a try?

fwcd:
IIRC I did implement that in https://github.com/mixxxdj/mixxx/pull/12112/commits/f9cfd6295c86a4fcd87c3b87e5ea75b8d587ef3c 😄

m0dB:
Will you add that to this PR then?

JoergAtGithub:
> Will you add that to this PR then?

This commit is already part of this PR!

m0dB:
Looks like I somehow tested without that commit last time. I retested and it indeed works fine!

m0dB (approved):

JoergAtGithub:
Code LGTM! Thank you!

fwcd:
Can/should we add this to the change log?

JoergAtGithub:
The label changelog is set for this. When the next changelog PR will be opened, at leat all PRs with changelog label should be added to the changelog. Feel free to take over this task.

PR #12118, opened 2023-10-15
Use cpack_add_component to hide applocal

### LINKED ISSUE #11992: Undocumented applocal in .msi installer (ANGLE DLLs)
### Bug Description

I noticed, that the `applocal` introduced in #11935 is not as hidden as intended. It became an optional installation component in the .msi installer, which the end user can deselect.
Therefore should give it a meaningful name/description like ANGLE OpenGL fallback driver. That the user understands, what he can select/deselect.

### Version

2.4.0

### OS

Windows

daschuer:
This fixed #11992

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3366,14 +3366,6 @@ if(WAVPACK)
   target_link_libraries(mixxx-lib PRIVATE WavPack::wavpack)
 endif()
 
-if (APPLOCAL_COMPONENT_DEFINED)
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
Tested the installer from this PR on Win7 and Win11: applocal is no longer visible as optional component, only PDB remains.
Also a local build on Win11 was startable from build and install directory.
Code LGTM! Thank you!

PR #12119, opened 2023-10-15
Notarization: Migrate script to `notarytool`

### LINKED ISSUE #12089: Migrate macOS notarization scripts to `notarytool`
As per [the Apple docs on notarization](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow#3087734):

> The `altool` command has been deprecated, and the Apple notary service no longer supports `altool` from November 1, 2023. For more information, see the WWDC22 session [What’s new in notarization for Mac apps](https://developer.apple.com/wwdc22/10109).

We should migrate our notarization scripts to the new `notarytool` ASAP, given that `altool` will stop being supported in a few weeks (though admittedly the notarization has been broken for other reasons already).

fwcd:
### Fixes #12089 

As per [the official migration guide](https://developer.apple.com/documentation/technotes/tn3147-migrating-to-the-latest-notarization-tool). In principle everything should work as-is, but I'll mark this as a draft for now for some more testing. Since notarization fails on 2.4 and `main`, we might have to generate new passwords anyway.

### To do

- [x] Migrate to the new credential flags
- [x] Migrate the polling mechanism to the `--wait` flag
- [x] Fix the plist parsing
- [ ] Print `notarytool log ...` if a submission fails

### DIFF
--- a/packaging/macos/sign_notarize_staple.sh
+++ b/packaging/macos/sign_notarize_staple.sh
@@ -20,16 +20,17 @@ xcrun notarytool submit \
     --wait \
     "${DMG_FILE}" > notarize_status.plist
 
+trap 'rm notarize_status.plist' EXIT
+cat notarize_status.plist
+
 # shellcheck disable=SC2181
 if [ "$?" != "0" ]; then
     echo "Notarization failed:"
-    cat notarize_status.plist
     curl "$(/usr/libexec/PlistBuddy -c 'Print notarization-info:LogFileURL' notarize_status.plist)"
     exit 1
 fi
 
 NOTARIZATION_STATUS="$(/usr/libexec/PlistBuddy -c 'Print notarization-info:Status' notarize_status.plist)"
-rm notarize_status.plist
 
 case "${NOTARIZATION_STATUS}" in
     success)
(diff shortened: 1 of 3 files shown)


packaging/macos/sign_notarize_staple.sh
```diff
@@ -2,57 +2,49 @@
...
+    "${DMG_FILE}" > notarize_status.plist
+
+trap 'rm notarize_status.plist' EXIT
+cat notarize_status.plist
+
+# shellcheck disable=SC2181
+if [ "$?" != "0" ]; then
+    echo "Notarization failed:"
+    curl "$(/usr/libexec/PlistBuddy -c 'Print notarization-info:LogFileURL' notarize_status.plist)"
+    exit 1
```
Swiftb0y, line 30: I don't think this makes sense nor am I sure if it ever worked. This will likely only run if `notarize_status.plist` does not exist, at which point the PlistBuddy will also fail, right?

Holzhaus: Do we really need error handling? We could just put `set -e` at the top of the script and either everything works or it doesn't.

fwcd:
Moved to #12123

PR #12124, opened 2023-10-16
GitHub Actions: Scope push trigger to main branches

fwcd:
This prevents CI from running the entire build matrix twice for PRs from upstream branches in `mixxxdj/mixxx` (triggering both `push` and `pull_request`), as per [this suggestion](https://github.com/orgs/community/discussions/26940#discussioncomment-5686753).

The side effect of this change would be that pushes no longer trigger CI on non-main, non-release branches in `mixxxdj/mixxx` outside of PRs. If that's not acceptable, we could [work around the issue using conditionals](https://github.com/orgs/community/discussions/26940#discussioncomment-6656489).

Thoughts?

### DIFF
--- a/.github/workflows/build-checks.yml
+++ b/.github/workflows/build-checks.yml
@@ -4,6 +4,9 @@ name: build-checks
 
 on:
   push:
+    branches:
+      - main
+      - "[0-9].[0-9]"
   pull_request:
 
 permissions:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,6 +4,9 @@ name: Build
 
 on:
   push:
+    branches:
+      - main
+      - "[0-9].[0-9]"
   pull_request:
 
 permissions:
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -4,6 +4,9 @@ name: pre-commit
 
 on:
   push:
+    branches:
+      - main
+      - "[0-9].[0-9]"
   pull_request:
 
 permissions:


daschuer:
I have just noticed that you don't have actions enabled for your personal branch. 
I have them enabled and noticed, that the same work is done twice after opening a pull request. On thine in the mixxxdj repo and one time in my repo. This is also a wast of resources. 
On the other hand a build on pushes is desired in my personal repo.

How does this relate here?

fwcd:
Well, forks usually don't need to enable GitHub Actions (and no longer do by default), since pushing to a PR already triggers the `pull_request` event for workflows. Since `mixxxdj/mixxx` currently runs builds on _all_ branch pushes, the workflows will run a second time under the `push` event.

Therefore my proposal is to:

- Limit `push` events to `main` and release branches (`2.4`, ...)
- Have the `pull_request`-Trigger take care of all PRs

The downside is that pushing to a feature branch that does not have an associated PR will no longer trigger CI. Is that an acceptable tradeoff?

fwcd:
Also note that `pull_request` builds run in the context of the targeted repo (i.e. PRs here will build in the context of this repo) whereas `push` builds will run in the repo that was pushed to (e.g. a fork).

> On the other hand a build on pushes is desired in my personal repo.

This is a tricky one. We could probably do that with the [conditional I've linked](https://github.com/orgs/community/discussions/26940#discussioncomment-6656489) in the PR description, but maybe there's a prettier solution too.

fwcd:
I'll close this since there doesn't seem to be that much interest in handling this edge case, given that most feature branches are in forks anyway. The only rather common case of creating pull requests from upstream branches would be PRs for release branch merges, but those would still cause the push trigger and duplicate CI runs, even with this PR.

TL;DR: The status quo seems to work well enough for now

PR #12135, opened 2023-10-18
Adjust position for FATAL_ERROR_MISSING_ENV()

daschuer:
It turns out that the warning is issued to later after the adjustments for QT6. 
Here is the fix. 
https://mixxx.zulipchat.com/#narrow/stream/247620-development-help/topic/help.20compiling.20mixxx


daschuer:
Committed as part of #12370

Issue #12137, opened 2023-10-18
Mixxx seeking access permission form iTunes Media folder on every launch


AlrightHector:
### Bug Description

Each time I launch Mixxx 2.4 beta it is seeking access permission to the iTunes Media folder.It's not retaining permissions.

### Version

2.4

### OS

Mac OS 10.14.6

fwcd:
Interesting, I didn't even know that we still supported macOS Mojave. Looks like 2.4 still supports macOS all the way back to Sierra (10.12).

That sounds a bit like #11552. Is "Full Disk Access" enabled for Mixxx in your settings?

fwcd:
By the way, the 2.4 snapshot builds from the website are quite outdated and IIRC do not include #11936 yet.

Could you try e.g. [this build](https://github.com/mixxxdj/mixxx/actions/runs/6411366768)? (scroll down and download `macOS Intel DMG`)

AlrightHector:
Thanks, the message doesn't appear for this build.

fwcd:
Yep, that build was unsandboxed and wouldn't have had the issue anyway, sorry...

Could you try the following build? https://github.com/mixxxdj/mixxx/actions/runs/6562071482

AlrightHector:
This build has the same issue, requesting permission. I've tried granting
full disc access, but the issue remains.

fwcd:
Hm, that's annoying. So I'll assume that #11552 isn't fixed after all...

fwcd:
I _think_ the issue was introduced by this PR a while back (cc @uklotzde):

- https://github.com/mixxxdj/mixxx/pull/3761

If you have some time to spare and would like to dig into this, you could try [building Mixxx](https://github.com/mixxxdj/mixxx/blob/main/README.md#building-mixxx) before:

```sh
git clone https://github.com/mixxxdj/mixxx.git
cd mixxx
git checkout 9e974cdb && git cherry-pick 59168913 e9223294 cf8e5a44
source tools/macos_buildenv.sh setup
cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DMACOS_BUNDLE=ON -B build
cmake --build build
cd build
cpack -G DragNDrop
```

and after that PR:

```sh
git clone https://github.com/mixxxdj/mixxx.git
cd mixxx
git checkout c9f52dc1 && git cherry-pick 59168913 e9223294 cf8e5a44
source tools/macos_buildenv.sh setup
[...]

PR #12138, opened 2023-10-18
macOS packaging: Enable app sandbox in unsigned 2.4 builds too

fwcd:
Cherry-pick of #12101 onto 2.4.

This would be great to have to move our test builds closer to the release builds and would really help in diagnosing issues such as #12137.


daschuer (approved):
Thank you. LGTM

PR #12194, opened 2023-10-19
Preferences: apply changes from all pages when pressing Apply (like when pressing Okay)

ronso0:
Currently it's very confusing that
* pressing **Apply** + **Cancel** (apply pending changes on the **current page**, close, discard other changes)
is not the same as
* pressing **Okay** (apply pending changes in **all** pages, close)

One step further to fixing #7440


daschuer (approved):
LGTM, Thank you.

PR #12199, opened 2023-10-20
Ability for controller to share data at runtime

acolombier:
This is a poor man solution, but it does the job with a minimal amount of effort.

TODO:
- [x] Support connection removal
- [x] Unit test


src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
```diff
@@ -24,6 +24,11 @@ class ControllerScriptInterfaceLegacy : public QObject {
 
     Q_INVOKABLE double getValue(const QString& group, const QString& name);
     Q_INVOKABLE void setValue(const QString& group, const QString& name, double newValue);
+
+    Q_INVOKABLE QJSValue getRuntimeData();
+    Q_INVOKABLE void setRuntimeData(const QJSValue& value);
+    Q_INVOKABLE QJSValue onRuntimeDataUpdate(const QJSValue& callback);
```
JoergAtGithub, line 30:
```suggestion
    Q_INVOKABLE QJSValue makeRuntimeDataConnection(const QJSValue& callback);
```
As far as I can see, this is very similar to `makeConnection` and should get a similar name for consistency across the API therefore.

src/controllers/controllerruntimedata.h
```diff
@@ -0,0 +1,34 @@
...
+#include <QVariant>
+
+#include "util/assert.h"
+
+/// ControllerRuntimeData is a wrapper that allows controllers script runtimes
+/// to share arbitrary data via a the JavaScript interface. It doesn't enforce
+/// any type consistency and it is the script responsibility to use this data in
+/// a considerate, non-destructive way (append to lists, extend to objects,
+/// ...), as well as expecting that others won't do so.
+class ControllerRuntimeData : public QObject {
```
JoergAtGithub, line 12: For me the term `ControllerRuntimeData` is not self-descriptive. I would suggest to call it `ControllerSharedData` instead.

JoergAtGithub (changes requested):
Please add declarations for the new functions/class to res/controllers/engine-api.d.ts

acolombier:
I have made the naming change as suggested. Would you suggest I completely remove the term `runtime` in the Engine API?

JoergAtGithub:
I think
```
getSharedData
setSharedData
makeSharedDataConnection
```
would be a consistent and self explaining naming. Do you agree?

Swiftb0y:
Two quick points (I haven't looked at the implementation too deeply):
1. I think I remember reading somewhere (couldn't find the source right now) that QJSValues are bound to the engine they have been created from and accessing them from a different engine is technically undefined behavior. QML's WorkerScript has a lot of code to work around that fact by serializing the objects each time.
2. Sharing controller data puts the "controller lock-in" issue (controller scripts assuming that another mapping is running at the same time, potentially breaking if that is not the case) back on the table. Do we have suggestions for avoiding that?

acolombier:
> would be a consistent and self explaining naming. Do you agree?

Yep sounds good, will update accordingly.

> I think I remember reading somewhere (couldn't find the source right now) that QJSValues are bound to the engine they have been created from and accessing them from a different engine is technically undefined behavior. QML's WorkerScript has a lot of code to work around that fact by serializing the objects each time.

Yep, that's right. This is why the `SharedData` is stored as a `QVariant` and a `QJSValue` is created just in time (e.g when calling via `getSharedData` or via the callback, with the related engine to get ownership of it). This will have a performance impact and definitely isn't intended for burst/low latency communication. IMHO, right approach for this use case (if it ever becomes a requirement) would be to have multiple devices as part of the same mapping/engine, but that would require a significantly bigger amount of work. 

> Sharing controller data puts the "controller lock-in" issue (controller scripts assuming that another mapping is running at the same time, potentially breaking if that is not the case) back on the table. Do we have suggestions for avoiding that?

The suggestion is likely to carefully review device that want to rely on this feature not to have any hard dependency. An example of this is the [Tracktor S4 MK3 PR](https://github.com/mixxxdj/mixxx/pull/12557) relying on this feature. As you can see in there, code is written in both way to handle arbitrary value for that `SharedData` (that is, case where the HID or Bulk part isn't being used). Realistically, I don't think many upstream mapping devices should build on top of this feature (S4 Mk3 is an exception as the controller exposes two devices), but this feature will allow user with some JS knowledge to make there home setup in sync (say you have two controllers and want to "pair" their behavior, this method offers you an easy way to build a bridge between both)

Swiftb0y:
> MHO, right approach for this use case (if it ever becomes a requirement) would be to have multiple devices as part of the same mapping/engine, but that would require a significantly bigger amount of work.

I haven't made up my mind up about this yet, but that has the problem that the currently frequently used technique of overwriting a libraries prototype would break mappings (since ES6 modules are singletons). So it requires a bit of thought. At least I'd be hesitant to allow arbitrary mappings to share the same engine because of the subtle incompatibilities they might have. This also roughly relates to the controller lock-in issue.

> The suggestion is likely to carefully review device that want to rely on this feature not to have any hard dependency.

That works for mappings that receive a review, but many mappings don't get upstreamed and thus never receive a review. so I'm hesitant to rely on that. 

> but this feature will allow user with some JS knowledge to make there home setup in sync (say you have two controllers and want to "pair" their behavior, this method offers you an easy way to build a bridge between both)

Yes, that is certainly a possibility of that approach, but I don't know whether that warrants the possibility for "controller lock-in". 

---

Going back to the usecase here: The problem is that a controller unit functioning as one is posing as two different devices. This currently has numerous problems such as subpar UX (the user needs to select the correct mapping for the different devices), the mappings can't communicate with each other (which is the reason for this PR) etc.

While I don't want to reject this PR outright or start increasing the scope enormously I do want to emphasize that this smells like it needs a different solution: Rather than having two completely different mappings for the different devices, it would make sense to have only one mapping that can talk to both devices. That mapping could then run in the same engine and none of the downsights I mentioned until now would apply (apart from the fact that it would require a huge refactoring)...

JoergAtGithub:
I think, that it's a big architectural benefit, that the different scripts run in own threads. The displaythread can need to do at lot at 60fps. Especially if we think about high-res display like the one of a Denon Prime 4. If such an image transfer thread could block the HID input from a jog or crossfader while scratching, the user would have a problem.
But any transfer between these controller threads must be also unblocking.

We had also requests to share data between seperated devices I remind of a global Shift button and of Deck-Switching for multiple CDJs.

acolombier:
@Swiftb0y I agree with everything you said, and join you on the fact that a larger work should be planned to enable a better integration with multi-device setup (whether these are multiple interface on the same hardware - eg S4Mk3 - or common setup - eg 2xCDJ+DJM).

The point of this PR/feature is to address limitation with a workaround (allowing mapping and users with basic script knowledge), and potentially deprecate and drop this in the future, where mapping can interface with multiple devices/interfaces. That being said, I've only been active contributor for the last year, but my experience tends to show that large refactors or works (QML, library, controller, ...) tend to take many years before they eventually yield to outcomes (which sometimes is just an abort). I appreciate I'm getting of topic, so happy to continue that discussion on Zulip

> Rather than having two completely different mappings for the different devices, it would make sense to have only one mapping that can talk to both devices. That mapping could then run in the same engine and none of the downsights I mentioned until now would apply

FYI, I started looking into this in my early days of "Getting screens for S4MK3" and either I'm not experienced/skilled enough, or you are underestimating the amount of refactor to "just" do that would require. But for the record, that was again what I meant with the "the same mapping/engine" ("engine" as in "Mixxx controller engine", not "JS single thread engine")

---

TL;DR; agreed it is suboptimal and only a workaround, happy to join efforts on working on larger refactor with a prior plan with consensus (in [proposal](https://github.com/mixxxdj/proposals)?)

JoergAtGithub:
Did you saw my comment above, that the `__win32__` to `__WINDOWS__`fix got lost in a later push?

acolombier:
> Did you saw my comment above, that the `__win32__` to `__WINDOWS__`fix got lost in a later push?

No I didn't. I don't see where there is Windows-specifc code on this PR. Did you mean to comment on #13008 ?

JoergAtGithub:
> > Did you saw my comment above, that the `__win32__` to `__WINDOWS__`fix got lost in a later push?
> 
> No I didn't. I don't see where there is Windows-specifc code on this PR. Did you mean to comment on #13008 ? I will fix 13008 as well

Indeed, my comment was for the other PR

Swiftb0y:
> @Swiftb0y I agree with everything you said

Thank you. I'm glad we're on the same page. 

> potentially deprecate and drop this in the future, where mapping can interface with multiple devices/interfaces.

I think that's also easier said than done. Deprecating and hopefully removing features that will cause user-generated content to break is bound to result in backlash. So I'm still unsure of that. 

> I appreciate I'm getting off topic, so happy to continue that discussion on Zulip

sure. 

> TL;DR; agreed it is suboptimal and only a workaround, happy to join efforts on working on larger refactor with a prior plan with consensus (in [proposal](https://github.com/mixxxdj/proposals)?)

I'm always short on time so I don't know how much I can help with actual implementing but I'm happy to collab on a proposal for now. 

There are many longstanding issues with the current engine reaching from small inconveniences all the way to architectural problems. So I'm very tempted to just write a separate engine entirely because the complications arising from trying to maintain compatibility with the current implementation doesn't seem worth the effort. 

> I've only been active contributor for the last year, but my experience tends to show that large refactors or works (QML, library, controller, ...) tend to take many years before they eventually yield to outcomes (which sometimes is just an abort).

Yes unfortunately that is the case. IMO its mostly a combination of refactors being too large, too breaking and/or people not reviewing.

JoergAtGithub:
For the single controller use case I can imagine two solution:
1.) Multiple mappings share one XML-File but the different controller scripts remain unmodified. Than we could declare the shared variables in the XML file.
2.) We add an identifier to the shared data. Than we could use namespacing to prevent collisions and if a single value is not set yet, the function could just return `undefined`.
Instead of:
```
engine.setSharedData({
            group: {
                "leftdeck": "[Channel1]",
                "rightdeck": "[Channel2]",
            },
            scrollingWavefom: {
                "[Channel1]": false,
                "[Channel2]": false,
                "[Channel3]": false,
                "[Channel4]": false,
            }
        });
```
we would have:
```
engine.setSharedData("S4Mk3", "group", {
                "leftdeck": "[Channel1]",
                "rightdeck": "[Channel2]",
            });
engine.setSharedData("S4Mk3", "scrollingWavefom", {
                "[Channel1]": false,
                "[Channel2]": false,
                "[Channel3]": false,
                "[Channel4]": false,
            });
```
Both approaches have the issue, that multiple controllers of the same type (CDJs) are difficult to group, but for a single controller with different interfaces (HID, MIDI, BULK) this should work.

Swiftb0y:
Yes, I think sharing data between scripts being declared in the same XML is completely fine because those scripts have to know about each other logically. This avoids the issue that two completely unrelated mapping get dependent on or incompatible with each other. The namespace solution (2.) only solves the incompatibility issue. 

I think for mappings declared in the same XML, its fine if the share data (its even fine if the run in the same engine for the same reason). I guess that just doesn't seem to work currently (#13015; though unconfirmed).

acolombier:
What is the consensus on required API change to get this feature over the line?

Swiftb0y:
> Maybe we can create the shared data from the Mixxx domain and define the businesses logic for it.

I don't think that makes sense. Requirements for controllers continually change. It's also difficult for us to actually define what kind of data needs to be shared. Mapping also may only look at whats there already and give up when the functionality needed is not there (instead of adding it to the C++ mixxx domain). Putting this functionality into the mixxx domain would also make it stable, making it impossible to remove once its not needed anymore. 

I prefer the current approach, full power to the mapping dev, but keep it scoped to the particular mapping their working on:

> What is the consensus on required API change to get this feature over the line?

I stand by https://github.com/mixxxdj/mixxx/pull/12199#issuecomment-2038027596

Any objections?

acolombier:
Declaring mapping in the same XML appears to be be a massive amount of work (AFAIK, that's what `ControllerScriptModuleEngine` was aiming for,  which is been work in progress for 4 years). It would also require getting a new API ready for XML (remember we are solving a 2 device with different protocol/2 mappings, part of a single gear of equipment with that PR).
Don't get me wrong, that suggestion sounds ideal if we had full time commitments, and a few devs ready to take on that work for the next 3 months, but I think we should be pragmatic on this issue.

Do you or the core team have other suggestions, which can be addressed in a circa 25h worth of work? If not, please feel free to close that PR.

Swiftb0y:
Can we scope runtime data to a namespace that would be defined in the XML? Then XMLs belonging to the same device would just need to declare the same namespace in the XML. I know its not as elegant, but it avoids the "mapping lock-in" issue I'm worried about.

acolombier:
That sounds reasonable. Are we happy with something like this?

```xml
<?xml version="1.0" encoding="utf-8"?>
<MixxxControllerPreset ...>
    <info>
        ...
    </info>
    <controller id="..." sharedDataNamspace="myNamespace">
       ...
    </controller>
</MixxxControllerPreset>
```

Would be nice to hear your thought @JoergAtGithub and @daschuer

Swiftb0y:
> That sounds reasonable. Are we happy with something like this?

Yes, thats pretty much what I was suggesting. I just dont know how feasible it is technically to implement these different per-engine namespace restrictions. You can probably judge that best.

daschuer:
I have collected here some arguments form above, describing the open issue: 

> Yes, I think sharing data between scripts being declared in the same XML is completely fine because those scripts have to know about each other logically. 

>> Maybe we can create the shared data from the Mixxx domain and define the businesses logic for it.

> I don't think that makes sense. Requirements for controllers continually change. It's also difficult for us to actually define what kind of data needs to be shared. Mapping also may only look at whats there already and give up when the functionality needed is not there (instead of adding it to the C++ mixxx domain). Putting this functionality into the mixxx domain would also make it stable, making it impossible to remove once its not needed anymore.

I can confirm that all these arguments are valid. 

If we go the easy route as suggested to allow defining any data form the mapping itself. It will only work reliable with two controllers of the same type. However, I consider it much more common that users have two different controllers. Once user are starting to build different mapping sharing data, we automatically inherit the "stable" problem. Mappings can no longer be changed independently, because we don't even know which custom mapping relies on them.

So let's think about the best way around this interoperability issue. 

How about distinguish two namespaces, one the unstable controller namespace, only suitable for exchanging data with the same controller, and one stable mixxx namespace for data exchange across mappings.

Swiftb0y:
> So let's think about the best way around this interoperability issue.

Yes, thats exactly what I'm trying to avoid. I'm glad we're on the same page now. 

I fully agree that having an open interchange point would inherit the stable issue, which is why I'm arguing to not support it at all right now. Just let each controller mapping have their own namespace (obviously different mappings belonging to the same controller share a namespace).

> How about distinguish two namespaces, one the unstable controller namespace, only suitable for exchanging data with the same controller, and one stable mixxx namespace for data exchange across mappings.

That's basically what I'm saying, except that we cancel the "stable mixxx namespace" completely because I don't want to open up that pandoras box right now. For that I completely agree that we should essentially have it implemented in mixxx with the appropriate businesslogic (in concrete terms, that would likely be in the form of new COs).

acolombier:
So just two summarise the two suggestion options:

## 1. Mapping define a namespace in manifest

This means that two mapping meant to work together (e.g S4Mk3 controller & screen) will define the same namespace, eg

```xml
<?xml version="1.0" encoding="utf-8"?>
<MixxxControllerPreset ...>
    <info>
        ...
    </info>
    <controller id="..." sharedDataNamspace="S4MK3">
       ...
    </controller>
</MixxxControllerPreset>
```

In the code, they will both use `getSharedData()/setSharedData(newValue: SharedDataValue)`, as defined currently

## 2. Use two types of namespaces

This means that mapping will be able to exchange data in a global namespace, unrestricted namespace using `getSharedData()/setSharedData(newValue: SharedDataValue)`, and will also have a second set of API (eg. `getLocalSharedData()/setLocalSharedData(newValue: SharedDataValue)`) which will be shared between controller using the same mapping  (absolute path, as I understand we fear user-modified mapping could create issues) 

In the case of the S4 Mk3, only the global namespace would be usable as both devices (screen and hid controller) uses different mapping

--- 

Both solution feels fairly easy to implement, personal preference would be for `1.` which feels simpler and less convoluted (e.g. someone that want to create, say `CDJ.hid.xml` and `CDJ_with_foobar.hid.xml` would have to update the first mapping and use the global namespace to get the two working, assuming it was previously using the local namespace, as all devices would use the same `CDJ.hid.xml`).

Just one comment I'd like to add, I think we should keep in mind that users modifying mappings should be assumed to know what they do to some extends - if they decide to share data across mapping, they understand that the data structure needs to match on both ends, and if they don't, it's fine to expect some misbehaviour, as it would already do if you modify certain Mixxx controller function without understanding why it does such a thing.

Swiftb0y:
> I think we should keep in mind that users modifying mappings should be assumed to know what they do to some extends

I very much disagree with that. I'd say the majority of mapping contributions we receive are from people with good intentions (thus the willingness to go through the PR process) but not much skill (usually their mapping is just copy-pasted together from 5 different other mappings). Thats one of the reasons I'm so obsessed about the quality of a mapping, because I know that if I don't flag an issue now, I'll have to flag it a bunch of times again when people eventually copy it. 

This is essentially the reason why I'm trying to keep the mapping engine fairly locked down, so people don't continuously footgun themselves.

Swiftb0y:
>  if they decide to share data across mapping, they understand that the data structure needs to match on both ends, and if they don't, it's fine to expect some misbehaviour, as it would already do if you modify certain Mixxx controller function without understanding why it does such a thing.

Thats easy to do if you're the author of both mappings. It's virtually impossible if it would have to match with any other mapping. That's why I'm advocating for the XML namespace approach (since it would be unlikely that people would try to use the exact same namespace).

acolombier:
> Thats easy to do if you're the author of both mappings. It's virtually impossible if it would have to match with any other mapping.

If you need to share some data at runtime between controllers, that means you are the author of the both (modified) mapping, so you may use some due diligence to make sure you do have clashing data. 

I'm also in favour of `1.` anyway, I was just making this point so we stay pragmatic about who this feature is for, and don't spend too much time red tapping everything.

JoergAtGithub:
I wonder if it wouldn't be better to decalare each shared variable instead of just the namespace. While more text to write, I think it would make it more clear how it works. And since this will remain a rarely used functionality, self explaining code might be more importat, than small code.

```
<?xml version="1.0" encoding="utf-8"?>
<MixxxControllerPreset ...>
    <info>
        ...
    </info>
    <sharedData namspace="S4MK3">
         <variable name = "globalShift" type = boolean>
         <variable name = "deckNo" type = number>
    </sharedData>
    <controller id="...">
       ...
    </controller>
</MixxxControllerPreset>


```

acolombier:
I have added namespace restriction. Hopefully this is heading toward the right direction.

> I think it would make it more clear how it works

While this provide more visibility of what data is available in certain namespace, this is likely to make the devx less efficient as the mapping are static file, which you usually don't use actively while developing new controller mappings.

res/controllers/engine-api.d.ts
```diff
@@ -25,6 +25,7 @@ declare interface ScriptConnection {
 
 declare namespace engine {
     type SettingValue = string | number | boolean;
+    type SharedDataValue = string | number | boolean | object | array | undefined;
```
Swiftb0y, line 28:
```suggestion
    type SharedDataValue = any;
```
at that point there is no need for the alias imo. It only worsens the editor UX

src/controllers/scripting/controllerscriptenginebase.cpp
```diff
@@ -218,3 +218,7 @@ void ControllerScriptEngineBase::errorDialogButton(
 void ControllerScriptEngineBase::throwJSError(const QString& message) {
     m_pJSEngine->throwError(message);
 }
+
+void ControllerScriptEngineBase::setSharedData(ControllerNamespacedSharedData* runtimeData) {
```
Swiftb0y, line 222:
```suggestion
/// Takes ownership of `runtimeData`
void ControllerScriptEngineBase::setSharedData(ControllerNamespacedSharedData* runtimeData) {
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -164,6 +172,88 @@ void ControllerScriptInterfaceLegacy::setValue(
...
+}
+
+void ControllerScriptInterfaceLegacy::onRuntimeDataUpdated(const QVariant& value) {
+    auto pJsEngine = m_pScriptEngineLegacy->jsEngine();
+    const auto args = QJSValueList{
+            pJsEngine->toScriptValue(value),
+    };
+
+    for (const auto& connection : m_runtimeDataConnections) {
+        QJSValue func = connection.callback; // copy function because QJSValue::call is not const
```
Swiftb0y, line 244: then just iterate mutably?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -335,6 +426,41 @@ void ControllerScriptInterfaceLegacy::triggerScriptConnection(
...
+bool ControllerScriptInterfaceLegacy::removeRuntimeDataConnection(
+        const ScriptConnection& connection) {
+    if (m_pScriptEngineLegacy->jsEngine() == nullptr ||
+            !m_runtimeDataConnections.contains(connection)) {
+        return false;
+    }
+
+    m_runtimeDataConnections.removeAll(connection);
+
+    return true;
```
Swiftb0y, line 438:
```suggestion
    VERIFY_OR_DEBUG_ASSERT(m_pScriptEngineLegacy->jsEngine() != nullptr) {
        return false;
    }
    return m_runtimeDataConnections.removeAll(connection) > 0;
```

res/controllers/engine-api.d.ts
acolombier: Why? Can you give me a specific example where it worsen the experience?

JoergAtGithub: 1. Do we want to share complex data types or only basic data types?
2. If we allow any data type, than we should use the keyword `any` instead of an alias.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -164,6 +172,87 @@ void ControllerScriptInterfaceLegacy::setValue(
     }
 }
 
+QJSValue ControllerScriptInterfaceLegacy::getSharedData() {
+    auto pJsEngine = m_pScriptEngineLegacy->jsEngine();
+    VERIFY_OR_DEBUG_ASSERT(pJsEngine) {
+        return QJSValue();
+    }
+    auto pRuntimeData = m_pScriptEngineLegacy->getSharedData();
```
JoergAtGithub, line 180: Please fix the clang-tidy warning

res/controllers/engine-api.d.ts
```diff
@@ -72,6 +73,28 @@ declare namespace engine {
      */
     function setParameter(group: string, name: string, newValue: number): void;
 
+    /**
+     * Gets the shared runtime data.
+     * @returns Runtime shared data value
+     */
+    function getSharedData(): SharedDataValue;
```
JoergAtGithub, line 80:
```suggestion
    /**
     * Gets the shared runtime data.
     * @returns Runtime shared data value - or undefined if not yet defined
     */
    function getSharedVariable(variablename:string): SharedDataValue;
```
We should not share one single complex blob. This makes it really difficult to understand for developers and we can't print meaningful error messages, when two mappings do no match together.
The later can especially happen, when you edit the variable type at runtime of Mixxx, because only one mapping is reloaded at a time.

JoergAtGithub (changes requested):

res/controllers/engine-api.d.ts
Swiftb0y: Well, trying this API out in VSCode, this is the intellisense I get: 
image
You can see how the type alias is opaque. 

Interestingly this is only the case if I reduce the number of types `SharedDataValue` could be. If I leave it with the current set, its automatically coerced to `any`...

image

At that point, I'd say we can replace the entire set with `any`. For some reason the type alias then becomes opaque. So I guess we can keep it if you insist.

acolombier: > Do we want to share complex data types or only basic data types?

Yes, I think we do. For example, one may want to share a `Uint8Array`, used as a HID packet, so it can be send to two or more devices

> If we allow any data type, than we should use the keyword any instead of an alias.

We don't - for example `callable` is not meant to be shared.
The reason why I disagree with using `any` is that it goes against TypeScript. Now if you think it just makes it simpler, I'm not going to argue.

JoergAtGithub: > Yes, I think we do. For example, one may want to share a `Uint8Array`, used as a HID packet, so it can be send to two or more devices

I don't see a use case for sharing such low level data between scripts. We should really limit this to high level syncronisation like a deck-switch or a global shift button. For everything else, we have existing APIs available.

> 
> > If we allow any data type, than we should use the keyword any instead of an alias.
> 
> We don't - for example `callable` is not meant to be shared. The reason why I disagree with using `any` is that it goes against TypeScript. Now if you think it just makes it simpler, I'm not going to argue.

If this is not equal `any`, than we should keep the type definition!

res/controllers/engine-api.d.ts
acolombier: > We should not share one single complex blob.

What if, in your namespace `foobar`, you want to share just a `TypedArray` or a string, like in the tests? I think adding variables don't add any values to the feature and just another level.

> we can't print meaningful error messages, when two mappings do no match together.

This isn't possible anyway. I thought we agreed that introducing namespace was a good enough option to prevent conflict. 
I also don't see how this would work anyway either since any controllers part of a namespace can get/edit values as well: this means that it is acceptable for `MappingA` to do `get("MappingBRunning")` (or `get().MappingBRunning` currently) and find no values.

JoergAtGithub: Namespacing is good enough to prevent unintended iterference of completly unrelated mappings.
But this does not address compatibility between the related scripts using the same namespace.

JoergAtGithub: I think this request is consistent with my original namespacing proposal above: https://github.com/mixxxdj/mixxx/pull/12199#issuecomment-2038000562

res/controllers/engine-api.d.ts
acolombier: > I don't see a use case for sharing such low level data between scripts. We should really limit this to high level syncronisation like a deck-switch or a global shift button. For everything else, we have existing APIs available.

I disagree. You should be able to share any memory representation you want, whether it is `class MySharedContext`, or a `TypedScript`. Since we are now focusing on future problems that could arise with this solution, we should also consider slightly unorthodox case and not introduce limitations IMO.

res/controllers/engine-api.d.ts
acolombier: How your solution will compatibility between the related scripts compare to the current one? Do you mean I should also implement the [API change mentioned here](https://github.com/mixxxdj/mixxx/pull/12199#issuecomment-2106768820)?

JoergAtGithub: This would be one approach. But also without an explicit declaration, it would be possible to print warnings like:
`Line 1234 - Shared variable "deckSwitch" changed type in call of engine.setSharedVariable from number to string`

res/controllers/engine-api.d.ts
Swiftb0y: > The reason why I disagree with using any is that it goes against TypeScript.

I think the issue is that technically according to the signature, sharing a callable would be possible because a callable is also an object, which is allowed. There is probably a reason why vscode coerced the current signature to `any` automatically. but take my words with a grain of salt, I don't really know typescript.

res/controllers/engine-api.d.ts
Swiftb0y: This is all javascript, there is no schema / static typing, nor is it really necessary. If there is no technical reason why variables could not change their type I don't see a good reason to introduce one (especially when it would only result in extra complexity).

acolombier: While I understand you are using type change as a way to detect a potential problem, this lefts out other usecase where a type change wouldn't occur but yet doesn't occur - eg wrong unit.
I think it all boils down to understanding what you are sharing, and making sure producers and consumers expect the same data AND don't conflict with each other. 
I think this type of detection is a nice to have, but I cannot really see how this is required. Realistically, only a hand full of official mappings will use this features, so I think it is a matter of making sure this is well document (e.g using shared context class for example).
If you really feel like there should be a key for data set/get, I will add it, but lets agree first this is the last significant change I'm being asked to do.

Swiftb0y: for the record, I don't think a schema a necessary. exchanging an untyped/dynamic object is completely fine.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -335,6 +425,37 @@ void ControllerScriptInterfaceLegacy::triggerScriptConnection(
...
+    };
+    QJSValue result = func.call(args);
+    if (result.isError()) {
+        if (m_pScriptEngineLegacy != nullptr) {
+            m_pScriptEngineLegacy->showScriptExceptionDialog(result);
+        }
+        qWarning() << "ControllerEngine: Invocation of runtime data connection "
+                   << connection.id.toString()
+                   << "connected to runtime data failed:" << result.toString();
+    }
```
Swiftb0y, line 456: how about you throw a JS exception instead?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -164,6 +172,87 @@ void ControllerScriptInterfaceLegacy::setValue(
...
+    const auto args = QJSValueList{
+            pJsEngine->toScriptValue(value),
+    };
+
+    for (auto& connection : m_runtimeDataConnections) {
+        QJSValue result = connection.callback.call(args);
+        if (result.isError()) {
+            if (m_pScriptEngineLegacy != nullptr) {
+                m_pScriptEngineLegacy->showScriptExceptionDialog(result);
+            }
```
Swiftb0y, line 248: you already assumped it was non-null above for `pJsEngine`,

Swiftb0y (commented):
full review. we're almost there IMO

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
acolombier: Updated this one and the one above

res/controllers/engine-api.d.ts
acolombier: @JoergAtGithub is that a hard requirement in your opinion, or are you happy to go with the untyped dynamic data as it is currently?

Swiftb0y (approved):
LGTM from my side tbh. Thank you very much for your patience @acolombier. waiting for @JoergAtGithub.

JoergAtGithub (changes requested):
I am really sorry, but I cannot agree with this technical approach. This is due to a number of concerns, in particular:

This is an extension of the JS mapping API. Once such an API is in in the field, we have to maintain it forever due to backward compatibility issues. Especially because we cannot test later adaptations of existing mappings, for which we have no hardware. Furthermore any duplicated API would mean a future debt for maintenance.

As a consequence, an extension of the mapping API must be thought through much more thoroughly or designed in such a way that it can be extended at a later date.

So far, I am only aware of two use cases for shared data between mappings:
- The synchronization of a shift button status
- The synchronization of a deck-select status

However, these two states can have a very different scope depending on the functional distribution across the devices. With a global namespace which is hardcoded in XML file, this is impossible to achieve.

I have created the following sketches to illustrate this:

grafik

grafik

grafik

grafik

grafik

My suggestion would be that we create a concept document at https://github.com/mixxxdj/proposals in which we draft an API design that covers all requirements.
Perhaps a subset can be derived from this, which is sufficient for the use case of the S4 Mk3. Everything else could then be extended at a later time.

Swiftb0y:
Thank you @JoergAtGithub for this thorough writeup. I have a couple comments:

Case2: This is a very valid usecase and to be honest, not one I had thought about before. The issue here is even more general in that sense that in that scenario, it would even be difficult to match which screen belongs to which controller (at least with PortMidi, since it doesn't expose the required information, right?).

Case 3, 4 and 5 are not supposed to be handled by this PR. There should not be (enforced by code review) a global namespace. I fully agree that the current solution does not work for "open-ended" interfaces. Those need to be centrally coordinated. 

> So far, I am only aware of two use cases for shared data between mappings

I also have a another one to add. the Traktor Kontrol S8 has buttons on the sides of the screens which likely communicate via the HID endpoint, but in practice only change the state of the QML screens. 

Usecase 5 is also interesting because its slightly ambigous in the sense that we should not assume that the user would always want these layers to be in sync. maybe they would like Deck 1/2 to be purely their "Stems deck" controlled by the S8 (having access to all the stems features) and deck 3/4 controlled by the turntables to have their turntable control (for accurate scratching). In that case, whatever mechanism manages the "focused layer" needs to be able to also be ignored. But this discussion should probably be continued in a proposal setting.

Swiftb0y:
TLDR; we need to address case 2 somehow, case 3,4,5 are and should not be solved in/via this PR.

acolombier:
> Once such an API is in in the field, we have to maintain it forever due to backward compatibility issues

This (and the following about maintaining an API) is only true for stable API. In this case, we could release this feature and mark this feature as unstable to unblock S4 Mk3 implementation (which as a reminder is meant to be [officially supported by Mixxx](https://opencollective.com/mixxx/expenses/189210)). This way you can freely remove those methods once there is a replacement.

I assume your point is also valid for #11407

I must say I find it disrespectful to play that card after 2 months of back and forth review processes. 

> So far, I am only aware of two use cases for shared data between mappings:

I guess you are taking Traktor/Native Instrument as reference. I suggest you look into #12557 or [this mod](https://community.native-instruments.com/discussion/26910/my-traktor-s4-mk3-screen-hack-is-back) for Traktor. 

TL;DR: The scope for mapping synchronisation is infinite.

> My suggestion would be that we create a concept document at https://github.com/mixxxdj/proposals in which we draft an API design that covers all requirements.

That sounds reasonable, except that https://github.com/mixxxdj/proposals doesn't seem of great interest to the Mixxx community so far (see [my stale PR](https://github.com/mixxxdj/proposals/pulls/1) trying to bring that repo to live). So till proven otherwise, I won't spend time draft a thorough proposal to see it being ignored.

Swiftb0y:
> This (and the following about maintaining an API) is only true for stable API. In this case, we could release this feature and mark this feature as unstable to unblock S4 Mk3 implementation (which as a reminder is meant to be [officially supported by Mixxx](https://opencollective.com/mixxx/expenses/189210)). This way you can freely remove those methods once there is a replacement.

I agree, that it would be okay to mark this feature private. especially since its essentially just a workaround for the fact that we can't run multiple controllers in the same engine anyways. ergo, this should be replaced in the mid-term future anyways.

> I must say I find it disrespectful to play that card after 2 months of back and forth review processes.

Let me assure you that @JoergAtGithub didn't mean for his veto to be disrespectful. His criticism is valid (especially considering the complications for case 2 which we should work through). 

> That sounds reasonable, except that https://github.com/mixxxdj/proposals doesn't seem of great interest to the Mixxx community so far (see https://github.com/mixxxdj/proposals/pull/1 trying to bring that repo to live). So till proven otherwise, I won't spend time draft a thorough proposal to see it being ignored.

I'm sorry that the PR has gone stale for so long. It's just easy to loose track of things. @daschuer would you mind continuing your ongoing review in https://github.com/mixxxdj/proposals/pull/1?

acolombier:
What is the status of this PR? Should I close it or is there something we can do to finish it?

Swiftb0y:
I think case 2 is valid. Do you have an idea on how to identify two different controllers and match with screen belongs to which controller? I think once we have identified that, attaching that info (invisibly) to the namespace handles is straightforward. But we need some way of ensure that the screen of one device doesn't get matched to the midi ports of another device.

acolombier:
I'm sorry but no, it does not make sense, and quite frankly I've lost the motivation to try and explain why. 
Since we refuse to work with iteration on that matter, I'm going to chose my battle on forfeit on this one.

For anyone reading this with a S4 Mk3 - note that I will keep  my dev branch (`traktor-s4-mk3`) with this feature, as long as I keep using this device.

FYI @ywwg please note that it won't be possible to complete #13004 now and will need someone else to take on that feature.

ywwg:
it does sound like a formal proposal is required for this to be resolved.  Hopefully the ultimate solution is not so far off from this implementation that we can use this code.

@acolombier I appreciate your contributions and understand your frustration.  With large, delicate new features such as the ones you are working on, it is often hard to do design in the PR process, which is why we need to make better use of the proposal feature.  That way the implementation can be worked out before you have done a bunch of work.

 In the past, when we have taken a step back from PR discussions and moved to a design discussion we have usually found that
1: people are not that far apart, and
2: the changes required to make everyone happy are not very much.

I am thinking of the hotcue color PR which was a huge issue back in the day.  We ended up mapping out the two designs and almost immediately found  a way through.

JoergAtGithub:
> Please elaborate.

If we would have the namespace as string in JavaScript like in

`engine.getSharedData("GlobalNamspace", "Shift");`

we could later add an API like

`engine.getUsbSerialNo();`

and could use it together with the unmodified sharing API like 

`engine.getSharedData("DeviceSpecificNamespace" + engine.getUsbSerialNo(), "DeckSwitch");`

Swiftb0y:
I see two fundamental issues with that:

1. It allows to specify a global namespace, something which as you rightfully pointed out doesn't work with the schema-less object being exchanged here. But I'm arguing that this PR is not supposed to address that usecase. 
2. The controllerscripts manually need to take care of avoiding naming collisions. A mapping author might not think of that instead expecting mixxx to take care of that. It's reasonable to expect this collision handling to be taken account for controllers that are made to be used in tandem (single-deck controller like the Traktor D2 for instance), but it may not be reasonable for the S4 to take care of that (even though its totally possible to use two S4's with a single mixxx instance). Moreover, it may not be feasible to get the USB-serial number, so we'd need to enable some way of manual grouping from the UI as a fallback anyways IMO.

daschuer:
"Viele Köche verderben den Brei"
This PR is really a can of worms, because everyone has a different idea. 

Sorry, I was also involved in that here: https://github.com/mixxxdj/mixxx/pull/12199#issuecomment-2105684738
Our conclusion at that time was to keep this as a mapping private solution and consider a C++/Control Object API for other interoperable features that are usable for all mappings. 

Thank you @JoergAtGithub to bringing all the cases up. 
I have probably not fully understand everything, but for me they look like candidates for a global solution for all mappings. The user needs probably a GUI to set it up as well. 

My hope is to strip that PR down to the original use case and just make sure we have not created a maintenance burden that will bite us in future as expressed. 

Let him cook? What do you think.

acolombier:
> "Viele Köche verderben den Brei" This PR is really a can of worms, because everyone has a different idea.

I had the same feeling. I appreciate we came to the same conclusion.
 
> The user needs probably a GUI to set it up as well.

I think this makes sense. This way I can keep the namespace scoping I had already implemented following the general consensus on the need to isolate shared data, but keep it to a "default" for now. 
In second phase, we could decide how we would want the mapping and GUI to look like regarding that need.

ywwg:
Per discussion, this work is not abandoned but on the back burner while we do a proper design for controller data sharing.

Issue #11745, opened 2023-07-15
Overlapping cues are rendered different in waveform and overview


daschuer:
### Bug Description

See:
![image](https://github.com/mixxxdj/mixxx/assets/1777442/dcd3fae2-fbe5-435b-9901-3620808aeba2)

Here we I have stacked hot cue 5, 6, 7 

The good think is that we see, where 5 and 7 is. 6 is invisible. 

My expectation was that waveform and overview show both 7.

### Version

2.3

PR #12203, opened 2023-10-21
improve rendering of waveform marks

m0dB:
This PR improves the rendering of the waveform marks, by auto-scaling and better aligning the symbolic marks (such as the intro/outro triangles and the loop open circle arrow), as discussed here: https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/look.20of.20loop.20out.20marker

Additionally it:
- removes code duplication between `waveformrendermark.cpp` and `allshaders/waveformrendermark.cpp` by moving the mark image rendering to `waveformmark.cpp` and doing a general cleanup of that code
- doesn't draw a small empty rectangle for marks without a label
- fixes drawing of marks in vertical orientation in legacy waveforms
- fixes a small bug with hovering over marks (propagate leave event)
- uses antialiasing to render the marks and the glsl play position mark image

Screenshot 2023-10-21 at 03 34 09

### DIFF
--- a/res/skins/LateNight/waveform.xml
+++ b/res/skins/LateNight/waveform.xml
@@ -62,6 +62,7 @@
             </MarkRange>
             <Mark>
               <Control>loop_start_position</Control>
+              <!--Text>&#8635;</Text-->
               <Icon>skin:../LateNight/classic/style/mark_loop.svg</Icon>
               <Align>top|left</Align>
               <Color><Variable name="LoopColor"/></Color>
@@ -86,7 +87,8 @@
             </MarkRange>
             <Mark>
               <Control>intro_start_position</Control>
-              <Text>&#9698;</Text>
+              <!--Text>&#9698;</Text-->
+              <Icon>skin:../LateNight/classic/style/mark_intro.svg</Icon>
               <VisibilityControl>[Skin],show_intro_outro_cues</VisibilityControl>
               <Align>top|right</Align>
               <Color><Variable name="IntroOutroColor"/></Color>
@@ -95,7 +97,8 @@
             <Mark>
               <Control>intro_end_position</Control>
               <VisibilityControl>[Skin],show_intro_outro_cues</VisibilityControl>
-              <Text>&#9698;</Text>
+              <!--Text>&#9698;</Text-->
+              <Icon>skin:../LateNight/classic/style/mark_intro.svg</Icon>
               <Align>top|left</Align>
               <Color><Variable name="IntroOutroColor"/></Color>
               <TextColor>#FFFFFF</TextColor>
@@ -113,14 +116,16 @@
             <Mark>
               <Control>outro_start_position</Control>
               <VisibilityControl>[Skin],show_intro_outro_cues</VisibilityControl>
-              <Text>&#9699;</Text>
+              <!--Text>&#9699;</Text-->
+              <Icon>skin:../LateNight/classic/style/mark_outro.svg</Icon>
               <Align>top|right</Align>
               <Color><Variable name="IntroOutroColor"/></Color>
               <TextColor>#FFFFFF</TextColor>
             </Mark>
             <Mark>
               <Control>outro_end_position</Control>
-              <Text>&#9699;</Text>
+              <!--Text>&#9699;</Text-->
+              <Icon>skin:../LateNight/classic/style/mark_outro.svg</Icon>
               <VisibilityControl>[Skin],show_intro_outro_cues</VisibilityControl>
               <Align>top|left</Align>
               <Color><Variable name="IntroOutroColor"/></Color>
--- a/res/skins/LateNight/waveform.xml
+++ b/res/skins/LateNight/waveform.xml
@@ -62,7 +62,7 @@
             </MarkRange>
             <Mark>
               <Control>loop_start_position</Control>
-              <Text>&#8635;</Text>
+              <Icon>skin:../LateNight/classic/style/mark_loop.svg</Icon>
               <Align>top|left</Align>
               <Color><Variable name="LoopColor"/></Color>
               <TextColor>#FFFFFF</TextColor>
--- a/res/skins/LateNight/waveform.xml
+++ b/res/skins/LateNight/waveform.xml
@@ -62,7 +62,7 @@
             </MarkRange>
             <Mark>
               <Control>loop_start_position</Control>
-              <!--Text>&#8635;</Text-->
+              <Text>&#8635;</Text>
               <Icon>skin:../LateNight/classic/style/mark_loop.svg</Icon>
               <Align>top|left</Align>
               <Color><Variable name="LoopColor"/></Color>
@@ -87,7 +87,7 @@
             </MarkRange>
             <Mark>
               <Control>intro_start_position</Control>
-              <!--Text>&#9698;</Text-->
+              <Text>&#9698;</Text>
               <Icon>skin:../LateNight/classic/style/mark_intro.svg</Icon>
               <VisibilityControl>[Skin],show_intro_outro_cues</VisibilityControl>
               <Align>top|right</Align>
@@ -97,7 +97,7 @@
             <Mark>
               <Control>intro_end_position</Control>
               <VisibilityControl>[Skin],show_intro_outro_cues</VisibilityControl>
-              <!--Text>&#9698;</Text-->
+              <Text>&#9698;</Text>
               <Icon>skin:../LateNight/classic/style/mark_intro.svg</Icon>
               <Align>top|left</Align>
               <Color><Variable name="IntroOutroColor"/></Color>
@@ -116,7 +116,7 @@
             <Mark>
               <Control>outro_start_position</Control>
               <VisibilityControl>[Skin],show_intro_outro_cues</VisibilityControl>
-              <!--Text>&#9699;</Text-->
+              <Text>&#9699;</Text>
               <Icon>skin:../LateNight/classic/style/mark_outro.svg</Icon>
               <Align>top|right</Align>
               <Color><Variable name="IntroOutroColor"/></Color>
@@ -124,7 +124,7 @@
             </Mark>
             <Mark>
               <Control>outro_end_position</Control>
-              <!--Text>&#9699;</Text-->
+              <Text>&#9699;</Text>
               <Icon>skin:../LateNight/classic/style/mark_outro.svg</Icon>
               <VisibilityControl>[Skin],show_intro_outro_cues</VisibilityControl>
               <Align>top|left</Align>
--- a/src/waveform/renderers/waveformmark.cpp
+++ b/src/waveform/renderers/waveformmark.cpp
@@ -261,13 +261,10 @@ QImage WaveformMark::generateImage(float breadth, float devicePixelRatio) {
         QString path = m_pixmapPath;
         // Use devicePixelRatio to properly scale the image
         QImage image = *WImageStore::getImage(path, devicePixelRatio);
-        // QImage image = QImage(path);
         //  If loading the image didn't fail, then we're done. Otherwise fall
         //  through and render a label.
         if (!image.isNull()) {
-            image =
-                    image.convertToFormat(QImage::Format_ARGB32_Premultiplied);
-            // WImageStore::correctImageColors(&m_image);
+            image = image.convertToFormat(QImage::Format_ARGB32_Premultiplied);
             //  Set the pixel/device ratio AFTER loading the image in order to get
             //  a truly scaled source image.
             //  See https://doc.qt.io/qt-5/qimage.html#setDevicePixelRatio
--- a/src/waveform/renderers/waveformmark.cpp
+++ b/src/waveform/renderers/waveformmark.cpp
@@ -178,43 +178,40 @@ struct MarkerGeometry {
 
         const qreal margin{3.f};
 
-        qreal capHeight;
-        {
-            QFontMetricsF metrics{m_font};
-            capHeight = metrics.capHeight();
+        QFontMetricsF metrics{m_font};
+        const qreal capHeight = metrics.capHeight();
 
-            if (m_isSymbol) {
-                // Symbols can be aligned and sized in an way that is not ideal.
-                // We auto-scale the font size so the symbol fits in capHeight
-                // (the height of a flat capital letter such as H) but without
-                // exceeded a width of capHeight * 1.1
-                const auto targetHeight = std::ceil(capHeight);
-                const auto targetWidth = std::ceil(capHeight * 1.1f);
+        if (m_isSymbol) {
+            // Symbols can be aligned and sized in an way that is not ideal.
+            // We auto-scale the font size so the symbol fits in capHeight
+            // (the height of a flat capital letter such as H) but without
+            // exceeded a width of capHeight * 1.1
+            const auto targetHeight = std::ceil(capHeight);
+            const auto targetWidth = std::ceil(capHeight * 1.1f);
 
-                // As a starting point we look at the tight bounding rect at a
-                // large font size
-                m_font.setPointSize(50.0);
-                metrics = QFontMetricsF(m_font);
-                m_contentRect = metrics.tightBoundingRect(label);
+            // As a starting point we look at the tight bounding rect at a
+            // large font size
+            m_font.setPointSize(50.0);
+            metrics = QFontMetricsF(m_font);
+            m_contentRect = metrics.tightBoundingRect(label);
 
-                // Now we calculate how much bigger this is than our target
-                // size.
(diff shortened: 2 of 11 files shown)


acolombier:
What happens when labels overlap? (e.g in the screenshot, label of hotcue 4 would be long enough to overlap label of 3)
Do you think it would be possible to add some logic to bring labels on the foreground? For example, the last activated on is always on foreground or if the user hover them on the waveform, or in the deck

ronso0:
@acolombier I think you're referring to #11745 

I suggest to keep this PR compact, then take care of the other issues (#11745 and dark/bright detection / hover effect).

I'll test this asap.

m0dB:
There is currently no such logic. But the case you describe (the previous marker being so long that it overlaps the next marker) could be very easily solved with very simple logic: draw the markers from right to left. 

But what annoys me most is when you have two markers at the same position (I actually use that quite a lot, e.g. a normal and a looping hotcue). I want to try placing the second one higher than the first in that case.

In any case, all that should be in a separate PR.

acolombier:
@ronso0 this isn't quite the same now - my question was if instead of `OK!` as a label, you have `Some long text`(up to 23 chars + the `...` suffix), how will that be interfering with the 3rd hotcue? I assume one will overlap the other, and I'm wondering if we should account for further work (no particularly in that PR) to improve the UX

@m0dB I also use that behavior and suffering from the same limitation and the change you are describing would be very appreciated :)

m0dB:
I think https://github.com/mixxxdj/mixxx/issues/11745 is more about the inconsistency between the scrolling waveform and the waveform overview.

src/waveform/renderers/waveformmark.h
```diff
@@ -94,23 +96,31 @@ class WaveformMark {
...
   private:
     std::unique_ptr<ControlProxy> m_pPositionCO;
     std::unique_ptr<ControlProxy> m_pEndPositionCO;
     std::unique_ptr<ControlProxy> m_pVisibleCO;
-    std::unique_ptr<QOpenGLTexture> m_pTexture; // used by allshader::WaveformRenderMark
+
     friend class allshader::WaveformRenderMark;
+
+    std::unique_ptr<QOpenGLTexture> m_pTexture; // Used by allshader::WaveformRenderMark
+    QImage m_image;                             // Used by WaveformRenderMark
```
Swiftb0y, line 119: so IIUC, `m_pTexture` is always `nullptr` when running with a legacy shader and QImage `isNull` the other way around, right? Wouldn't it be better instead then if these concepts lived in the shaders then instead? At least that sounds like thats a better separation of concern.

src/waveform/renderers/waveformmark.h
```diff
@@ -94,23 +96,31 @@ class WaveformMark {
...
+    bool m_needsUpdate{true};                   // To indicate either the texture or the image has
```
Swiftb0y, line 120: while I don't have any objection to this personally, the pre-established style was to always initialize all members only in the member initializer list of the constructor(s).

src/waveform/renderers/waveformmark.h
m0dB: In the renderer you mean? (WaveformMarkRenderer and allshaders::WaveformMarkRenderer). But that would require keeping track of which image or texture goes with which WaveformMark. No, this really needs to be part of WaveformMark. But maybe it can be done by subclassing WaveformMark... I will have a look, I agree this is not ideal.

src/waveform/renderers/waveformmark.h
m0dB: Ok, I wasn't aware of that. I was under the impression that the general push was to go for modern c++. Do you think this is open for discussion? Otherwise I will change it.

src/waveform/renderers/waveformmark.h
Swiftb0y: > But that would require keeping track of which image or texture goes with which WaveformMark.

The WaveformMarks in the renderer are just stored in a container, should just be sufficient to store a `std::pair<std::unique_ptr<QOpenGLTexture>, WaveformMark>` and then implement `generateMarkImage` in terms of that, right?

src/waveform/renderers/waveformmark.h
Swiftb0y: I'm in favor of it, but you'll have to discuss with "the keeper of the styleguide" @daschuer. Do you have any references that this style is preferred over ours in modern c++?

JoergAtGithub:
On my Win11 system the green loop marker looks different than in your screenshot:
grafik

m0dB:
> On my Win11 system the green loop marker looks different than in your screenshot:

My screenshot is HDPI so that will make a difference for sure, but indeed this looks worse than it should... Can you try changing, in waveformmark.cpp:

```
// And we scale the font size accordingly.
m_font.setPointSizeF(50.0 * ratio);
```

to

```
m_font.setPixelSize(19);
```

and see if that improves things? (This will make the intro/outro markers look weird, but that's expected)

m0dB:
and without this PR ?

JoergAtGithub:
Plain 2.4:
grafik

m0dB:
Thanks for testing and following up. Well, that's a relieve! That also looks bad!

What if you paste this character ↻ in a text editor with font Open Sans (bold) at different sizes?

And does it improve things if in `waveformmark.cpp` instead of `painter.drawText(pos, label);` you use `painter.drawText(static_cast<int>(pos.x()), static_cast<int>(pos.y()), label);` ?

JoergAtGithub:
This PR with `painter.drawText(static_cast<int>(pos.x()), static_cast<int>(pos.y()), label);`:
grafik

m0dB:
My conclusion is that Windows does a pretty bad job at rendering this symbol. Wondering what it looks like on Linux.

Maybe we should use a png instead?

ronso0:
On Ubuntu Studio 20.04 it looks okay @150%
image

@250%
image

JoergAtGithub:
> Maybe we should use a png instead?

If possible, we should use svg vector graphics, as in the skins.

src/waveform/renderers/waveformmark.h
m0dB: No, they are stored in a WaveformMarkSet. But I can derive a WaveformMarkImage and a WaveformMarkTexture from WaveformMark and templatize WaveformMarkSet.

m0dB: I started doing this, but it adds way too much complexity. I prefer having these two data members, of which either one is used depending on the context, instead of adding all this complexity.

m0dB:
I have added the possibility to specify an svg.

This is with an svg for the loop mark.

Screenshot 2023-10-28 at 12 24 03

Does this look okay on non-HDPI, @JoergAtGithub ? Maybe the svg needs a bit of tweaking.

src/waveform/renderers/waveformmark.h
Swiftb0y: okay. Have you considered putting the members in a variant to highlight the fact that they are mutual exclusive? If that introduces too much complexity again, then ignore it. I'm just trying to come up with solutions to make the code more self-documenting.

m0dB: yes, that makes sense! will do!

m0dB: done. hope this is in line with what you had in mind.

src/waveform/renderers/waveformmark.h
m0dB: not relevant anymore

JoergAtGithub:
Yes, the loop marker looks now fine on Windows too!
grafik
The Intro/Outro markers look not that bad as the loop markers looked before, but also there SVG should be used.

m0dB:
> Yes, the loop marker looks now fine on Windows too!

Great!

> The Intro/Outro markers look not that bad as the loop markers looked before, but also there SVG should be used. Which can be done in another PR.

No, I will add it to this PR, it's very little effort.

ronso0:
> No, I will add it to this PR, it's very little effort.

All skins? :grimacing: 
Reason for asking: once upon a time we decided to add the Sync Leader control to all skins (even Shade?), and just today I noticed I only managed to add it to LateNight.
It's an entirely different story (in terms of xml integration and design effort), though I wanted to add that note.

m0dB:
That's a good point, @ronso0 . I only added this to Latenight, but the text rendering is still there, so of course other skins can continue using that.

JoergAtGithub:
Now all the markers look fine on Windows11! Thank you!

src/waveform/renderers/waveformmark.h
Swiftb0y: Yes, almost. I was rather thinking of `std::variant` but this also works. The only thing I'm slighly concerned about is the blind static_cast from the parent to the derived type but I don't know any better solution either right now without introducing lots of boilerplate.

Swiftb0y: Thank you for going through the extra effort.

Swiftb0y (approved):
LGTM. thanks a lot. before merging, did someone test this on linux as well?

ywwg:
looks good on linux

src/waveform/renderers/waveformmark.cpp
```diff
@@ -127,14 +136,242 @@ bool WaveformMark::contains(QPoint point, Qt::Orientation orientation) const {
...
     }
-    bool lineHovered = m_linePosition >= position - lineHoverPadding &&
-            m_linePosition <= position + lineHoverPadding;
+    bool lineHovered = m_linePosition >= point.x() - lineHoverPadding &&
+            m_linePosition <= point.x() + lineHoverPadding;
 
     return m_label.area().contains(point) || lineHovered;
 }
+
+// Helper class to calculate the geometry and fontsize needed by generateImage
```
ywwg, line 148: technically this is a struct not a class, so let's call it that

src/waveform/renderers/waveformmark.cpp
```diff
@@ -127,14 +136,242 @@ bool WaveformMark::contains(QPoint point, Qt::Orientation orientation) const {
...
+        // party the resulting font size is hard to predict (affects all supported OS).
+
+        m_font.setPixelSize(13);
+        m_font.setWeight(75); // bold
+        m_font.setItalic(false);
+
+        const qreal margin{3.f};
+
+        qreal capHeight;
+        {
```
ywwg, line 182: why is this scoping needed?

src/waveform/renderers/waveformmark.cpp
```diff
@@ -127,14 +136,242 @@ bool WaveformMark::contains(QPoint point, Qt::Orientation orientation) const {
...
+    // Load the pixmap from file.
+    // If that succeeds loading the text and stroke is skipped.
+
+    m_breadth = static_cast<int>(breadth);
+
+    if (!m_pixmapPath.isEmpty()) {
+        QString path = m_pixmapPath;
+        // Use devicePixelRatio to properly scale the image
+        QImage image = *WImageStore::getImage(path, devicePixelRatio);
+        // QImage image = QImage(path);
```
ywwg, line 267: please remove commented-out lines like this throughout

src/waveform/renderers/waveformmark.cpp
```diff
@@ -127,14 +136,242 @@ bool WaveformMark::contains(QPoint point, Qt::Orientation orientation) const {
...
+        //  If loading the image didn't fail, then we're done. Otherwise fall
+        //  through and render a label.
+        if (!image.isNull()) {
+            image =
```
ywwg, line 271: unnecessary newline.  if possible, please run the code formatter on changed lines

src/waveform/renderers/waveformmark.cpp
ywwg: oh is this copied from the original code? then maybe that's fine 👍🏻

src/waveform/renderers/waveformmark.h
```diff
@@ -21,6 +21,14 @@ class WaveformRenderMark;
 
 class WaveformMark {
   public:
+    class Graphics {
+      public:
+        Graphics()
+                : m_obsolete{false} {
+        }
+        bool m_obsolete;
```
ywwg, line 29: I think this is equivalent and simpler, though we might have a style reason not to use this format? 
```suggestion
        bool m_obsolete = false;
```

src/waveform/renderers/waveformmark.h
```diff
@@ -94,23 +104,29 @@ class WaveformMark {
...
+    QImage generateImage(float breath, float devicePixelRatio);
+
     QColor m_textColor;
     QString m_text;
     Qt::Alignment m_align;
     QString m_pixmapPath;
+    QString m_iconPath;
 
-    float m_linePosition;
+    float m_linePosition{};
```
ywwg, line 115: I'm not sure we use this style of initialization elsewhere, probably clearer to do = 0; instead

ywwg (commented):
looks good! Please clean up stray commented-out lines before merging, and I've made some comments.  I am not blocking this PR on them though since I am famously bad at getting back to my reviews :)

res/skins/LateNight/waveform.xml
```diff
@@ -62,7 +62,8 @@
             </MarkRange>
             <Mark>
               <Control>loop_start_position</Control>
-              <Text>&#8635;</Text>
+              <!--Text>&#8635;</Text-->
```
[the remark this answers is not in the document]
ronso0: not sure about this (here). sometimes it's helpful to have the matching unicode char at hand.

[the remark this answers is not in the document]
m0dB: Yes, I agree, the Text should be fallback. I will revert the commenting.

src/waveform/renderers/waveformmark.h
m0dB: I would like to get developer concensus on this. @Swiftb0y (IIRC) recently told me to initialize in the initializer list...

But there are two issues at hand here: 1) using curly braces aka uniform initialization and 2) in-class initialization vs initializer lists.

As for 1, from what I understand, uniform initialization is the prefered way since c++11. The standard https://isocpp.org/wiki/faq/cpp11-language#uniform-init clearly says: "Prefer initializing using {}."

As for 2, I don't think there is such a clear preference. When you have multiple constructors, in-class initialization certainly makes things easier, otherwise I don't think it matters much.

src/waveform/renderers/waveformmark.h
m0dB: See my comment below on uniform initialization.

src/waveform/renderers/waveformmark.h
Swiftb0y: > I would like to get developer concensus on this. @Swiftb0y (IIRC) recently told me to initialize in the initializer list...

Let me be clear, I didn't mean to tell you to use initializer lists, I just said other people will complain and I thought I'd mention it so it doesn't seem like I didn't review your PR thoroughly. I also prefer this style and if other devs do so too, then lets use it. If others complain we can settle this in a simple vote. 

> I'm not sure we use this style of initialization elsewhere, probably clearer to do = 0; instead


I also prefer `= 0` but only when the type is clearly numerical. On the other hand the meaning of `= 0` is already overloaded in the context of pure virtual member functions. So my preference is probably wrong and we should prefer using `{}` as per the cpp core guidelines.

res/skins/LateNight/waveform.xml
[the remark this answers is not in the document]
m0dB: Done

src/waveform/renderers/waveformmark.cpp
m0dB: Done

src/waveform/renderers/waveformmark.cpp
m0dB: i removed it, as it allows capHeight to be const.

src/waveform/renderers/waveformmark.cpp
m0dB: Done

src/waveform/renderers/waveformmark.cpp
m0dB: I am using pre-commit , not sure why this wasn't formatted. Anyway, fixed.

src/waveform/renderers/waveformmark.h
m0dB: Ok, let's take this discussion to zulip

m0dB: removing in-class initialization

m0dB:
All comments have been addressed, so I think this can be merged.

JoergAtGithub:
Just tested this again, and it's still working. LGTM! Thank you!

Issue #12115, opened 2023-10-15
Why do waveform types have different visual gains?


ronso0:
![image](https://github.com/mixxxdj/mixxx/assets/5934199/1adc2651-8e6a-42a7-840f-95daa4c35ad2)
![image](https://github.com/mixxxdj/mixxx/assets/5934199/ffa11872-299d-420d-b20e-3add6afc94a5)
![image](https://github.com/mixxxdj/mixxx/assets/5934199/a451e57c-cf02-4dbe-8adb-dc6f0bc55930)
![image](https://github.com/mixxxdj/mixxx/assets/5934199/76f57cfa-a94d-411d-9b92-b9211931d271)

Is this to be expected?
Anyhow, if I want to test waveforms I'd also need to adjust the visual gain to make the comparable.

daschuer:
I think we can consider this a bug.

m0dB:
I agree. The cause must be in how the amplitude is calculated by combining the high/mid/low values vs the real amplitude. So there might be an inherent and unavoidable difference but it shouldn't be this big. I will have a look when I have a moment.

m0dB:
First of all, note that when I implemented this for the new waveform types, I just based it on the existing implementations, so I am not 100% sure what the original thought behind all this is.

The waveform data contains values for _low_, _mid_, _high_, and _all_ . The RGB waveform renderer combines the _low_ , _mid_ and _high_ values to calculate the total amplitude: _combined = sqrt(low^2+mid^2+high^2)_, apparently based on the assumption that this value will be close to the _all_ value. This is unfortunately not the case. I plotted _all_ (x axis) against _combined_ values (y axis):

<img width="708" alt="Screenshot 2023-10-21 at 22 00 08" src="https://github.com/mixxxdj/mixxx/assets/79429057/8de68222-4aff-444c-8880-ba749874364c">

In the best case, they are equal, but generally the combined value is lower. I don't know how the _low_, _mid_ and _high_ values are calculated but clearly not in a way that makes the above assumption hold.

So the obvious question is: why don't we use the _all_ value for the amplitude that we draw, and the _low_, _mid_ and _high_ only to calculate the color? Because with the current implementation, the amplitude adapts to the low, mid and high gains, which is really nice.

Now, I think I have a way to improve this, I will try something and let you know.

ronso0:
Thanks for looking into this!

m0dB:
I have generated another plot, based on more data.

[...]

PR #12205, opened 2023-10-21
make scaling of GLSL RGB and RGB L/R waveform amplitudes consistent with simple waveform

m0dB:
As pointed out by @ronso0 in https://github.com/mixxxdj/mixxx/issues/12115 , the amplitudes of the waveform rendering is inconsistent between waveform types. I discuss the cause there.

This PR fixes the scaling, so the waveform amplitudes are the same as the simple waveform (when gains are 0), while still maintaining the adaptation of the waveform amplitude to the low/mid/high gains, by scaling the amplitude as drawn with the ratio between the low+mid+high magnitude with gains applies and without gains applied.

(Note that the wrong scaling is still present in the legacy waveforms. I will not fix those)

See https://github.com/mixxxdj/mixxx/issues/12115 for some screenshots that show the issue before this fix. With this fix, compare Simple with RGB and RGB L/R.
simple
RGB
RGBLR

This is RGB and RGB L/R with some gain applied (high and low bands turned down)

RGBgained
RGBLRgained

### DIFF
--- a/src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
+++ b/src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
@@ -160,23 +160,25 @@ void WaveformRendererLRRGB::paintGL() {
             float maxHigh = static_cast<float>(u8maxHigh);
             float maxAll = static_cast<float>(u8maxAll);
 
-            // Calculate the magnitude of the maxLow, maxMid and maxHigh values
-            const float magnitude = std::sqrt(
-                    math_pow2(maxLow) + math_pow2(maxMid) + math_pow2(maxHigh));
+            // Calculate the squared magnitude of the maxLow, maxMid and maxHigh values.
+            // We take the square root to get the magnitude below.
+            const float sum = math_pow2(maxLow) + math_pow2(maxMid) + math_pow2(maxHigh);
 
             // Apply the gains
             maxLow *= lowGain;
             maxMid *= midGain;
             maxHigh *= highGain;
 
-            // Calculate the magnitude of the gained maxLow, maxMid and maxHigh values
-            const float magnitudeGained = std::sqrt(
-                    math_pow2(maxLow) + math_pow2(maxMid) + math_pow2(maxHigh));
+            // Calculate the squared magnitude of the gained maxLow, maxMid and maxHigh values
+            // We take the square root to get the magnitude below.
+            const float sumGained = math_pow2(maxLow) + math_pow2(maxMid) + math_pow2(maxHigh);
 
             // The maxAll value will be used to draw the amplitude. We scale them according to
             // magnitude of the gained maxLow, maxMid and maxHigh values
-            if (magnitude != 0.f) {
-                const float factor = magnitudeGained / magnitude;
+            if (sum != 0.f) {
+                // magnitude = sqrt(sum) and magnitudeGained = sqrt(sumGained), and
+                // factor = magnitudeGained / magnitude, but we can do with a single sqrt:
+                const float factor = std::sqrt(sumGained / sum);
                 maxAll *= factor;
             }
 
--- a/src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
+++ b/src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
@@ -74,7 +74,7 @@ void WaveformRendererLRRGB::paintGL() {
     const float breadth = static_cast<float>(m_waveformRenderer->getBreadth()) * devicePixelRatio;
     const float halfBreadth = breadth / 2.0f;
 
-    const float heightFactorAbs = allGain * halfBreadth / std::sqrt(3.f * 256.f * 256.f);
+    const float heightFactorAbs = allGain * halfBreadth / 255.f;
     const float heightFactor[2] = {-heightFactorAbs, heightFactorAbs};
 
     const float low_r = static_cast<float>(m_rgbLowColor_r);
@@ -92,13 +92,12 @@ void WaveformRendererLRRGB::paintGL() {
 
     const int numVerticesPerLine = 6; // 2 triangles
 
-    const int linesReserved = numVerticesPerLine * (length * 2 + 1);
-    const int colorsReserved = numVerticesPerLine * (length * 2 + 1);
+    const int reserved = numVerticesPerLine * (length * 2 + 1);
 
     m_vertices.clear();
-    m_vertices.reserve(linesReserved);
+    m_vertices.reserve(reserved);
     m_colors.clear();
-    m_colors.reserve(colorsReserved);
+    m_colors.reserve(reserved);
 
     m_vertices.addRectangle(0.f,
             halfBreadth - 0.5f * devicePixelRatio,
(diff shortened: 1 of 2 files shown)


m0dB:
It would be good to have this merged for 2.4, but it could wait for a 2.4.1 (assuming that is on the road map).

ywwg:
yesssss thank you

src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
```diff
@@ -140,48 +139,61 @@ void WaveformRendererLRRGB::paintGL() {
...
+                u8maxMid = math_max(u8maxMid, waveformData.filtered.mid);
+                u8maxHigh = math_max(u8maxHigh, waveformData.filtered.high);
+                u8maxAll = math_max(u8maxAll, waveformData.filtered.all);
+            }
 
-                maxLow = math_max(maxLow, filteredLow);
-                maxMid = math_max(maxMid, filteredMid);
-                maxHigh = math_max(maxHigh, filteredHigh);
+            // Cast to float
+            float maxLow = static_cast<float>(u8maxLow);
```
ywwg, line 158: Not sure I understand why maxLow is a different type than u8maxLow. are we casting from uchar to float because of math max?

src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
```diff
@@ -193,15 +205,15 @@ void WaveformRendererLRRGB::paintGL() {
...
             m_colors.addForRectangle(red, green, blue);
         }
 
         xVisualSampleIndex += visualIncrementPerPixel;
     }
 
-    DEBUG_ASSERT(linesReserved == m_vertices.size());
-    DEBUG_ASSERT(colorsReserved == m_colors.size());
+    DEBUG_ASSERT(reserved == m_vertices.size());
+    DEBUG_ASSERT(reserved == m_colors.size());
```
ywwg, line 216: so I guess m_colors and m_vertices are the same size? (do we need to check both these things? This seems like an odd place to enforce this restriction rather than when reserved is calculated)

src/waveform/renderers/allshader/waveformrendererrgb.cpp
```diff
@@ -137,63 +137,76 @@ void WaveformRendererRGB::paintGL() {
...
 
-        const float max = math_max3(red, green, blue);
-
-        // Normalize red, green, blue, using the maximum of the three
-
-        if (max == 0.f) {
-            // avoid division by 0
+        // Normalize the color components using the maximum of the three
+        const float maxComponent = math_max3(red, green, blue);
+        if (maxComponent == 0.f) {
```
ywwg, line 193: do we have to worry about near-zero values causing the normFactor to explode?

ywwg (commented):
thank you for this! I have a couple small questions. Leaving these as comments because I have a bad habit of not following up

ywwg:
With this change, the waveforms seem over-zoomed at 1.0:

image

Compare to 0.9 zoom:

image

And RGB (not marked legacy) is a lot smaller:

image

ywwg:
my best friend is still the softer / higher rez legacy glsl at 1.5:
image

src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
```diff
@@ -140,48 +139,61 @@ void WaveformRendererLRRGB::paintGL() {
...
             maxLow *= lowGain;
             maxMid *= midGain;
             maxHigh *= highGain;
 
+            // Calculate the magnitude of the gained maxLow, maxMid and maxHigh values
+            const float magnitudeGained = std::sqrt(
+                    math_pow2(maxLow) + math_pow2(maxMid) + math_pow2(maxHigh));
+
+            // The maxAll value will be used to draw the amplitude. We scale them according to
+            // magnitude of the gained maxLow, maxMid and maxHigh values
```
daschuer, line 177: A bit more background info is useful here. 
because ....

src/waveform/renderers/allshader/waveformrendererrgb.cpp
```diff
@@ -137,63 +137,76 @@ void WaveformRendererRGB::paintGL() {
...
+        const float magnitude = std::sqrt(
+                math_pow2(maxLow) + math_pow2(maxMid) + math_pow2(maxHigh));
+
+        // Apply the gains
         maxLow *= lowGain;
         maxMid *= midGain;
         maxHigh *= highGain;
 
+        // Calculate the magnitude of the gained maxLow, maxMid and maxHigh values
+        const float magnitudeGained = std::sqrt(
```
daschuer, line 175: You can use only one std::sqrt() if you build the quotient (factor below) before.

daschuer (commented):
Just two comments, not yet tested.

src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
m0dB: yes, this way we cast only once, rather than casting for every iteration in the for loop in line 148.

src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
m0dB: yes, they are the same size. we could remove these asserts, it's just to make sure i didn't make a mistake when filling these vectors. but they don't enforce anything, they merly check. and checking doesn't hurt.

src/waveform/renderers/allshader/waveformrendererrgb.cpp
m0dB: no, if you look at how red green and blue are calculated, there is no such risk.

m0dB:
> With this change, the waveforms seem over-zoomed at 1.0:

Yes, I noticed this too, but you will see that happens to the simple waveform as well, so I guess that's "correct"

ronso0:
Are you sure this is not an issue with the pre-gain?

src/waveform/renderers/allshader/waveformrendererrgb.cpp
m0dB: Sorry, I understand what you mean here

daschuer: You can make use of the rule:
factor = Sqrt(A)/Sqrt(B) = Sqrt(A/B) 
To get rid of one expensive Sqrt call.

m0dB: ah, well spotted! i'll make the change.

m0dB: Done

src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
ywwg: Got it.  Wouldn't it be better to check them back up where reserved is created, then?  If there's something wrong with that size, then we will encounter problems before we get to this check.

m0dB: What is being checked with this assert is that the amount of data that we push to the m_vertices / m_colors vectors is correct (in m_vertices.addRectangle and m_colors.addForRectangle), to make sure we reserve exactly the amount we are going to push.

ywwg (approved):
This look great! and the /2 fix makes the new waveforms look SO much better. thank you!

daschuer (approved):
LGTM, Thank you.

m0dB:
Thanks for merging @daschuer . I also have fixed the drawing resolution for HSV, Filtered and Simple. I wanted to include that in this PR but I will create a new one.

m0dB:
https://github.com/mixxxdj/mixxx/pull/12356

fwcd:
Quick ping here since this seems to be the cause for #12448.

> This PR fixes the scaling, so the waveform amplitudes are the same as the simple waveform [...] Note that the wrong scaling is still present in the legacy waveforms. I will not fix those

I think the new scaling makes the waveforms a bit harder to read, e.g. very bass heavy tracks are almost "rectangular" until I turn down the visual gain manually. The old visual gain struck a nice balance by providing some "breathing space" for the waveform IMO.

Perhaps we could do it the other way around and lower the simple waveforms instead?

PR #12208, opened 2023-10-22
common HID: always return the passed-in group in case of error

ywwg:
The previous controller fix would return undefined if activedeck was not set, but that was not consistent with previous behavior.

As an aside the activedeck logic appears to be broken, because it is per-controller instead of per-deck. Many controllers have two decks, each with their own activedeck selector, and that currently is not supported by the library. A future fix should address this issue but is out of scope to unbreak controller configs for 2.4

### DIFF
--- a/res/controllers/common-hid-packet-parser.js
+++ b/res/controllers/common-hid-packet-parser.js
@@ -1344,7 +1344,6 @@ class HIDController {
             if (this.activeDeck === 3 || this.activeDeck === 4) {
                 return "[Channel3]";
             }
-            return undefined;
         } else if (group === "deck2") {
             if (this.activeDeck === 1 || this.activeDeck === 2) {
                 return "[Channel2]";
@@ -1352,7 +1351,6 @@ class HIDController {
             if (this.activeDeck === 3 || this.activeDeck === 4) {
                 return "[Channel4]";
             }
-            return undefined;
         }
         return group;
     }


ywwg:
tested with my traktor s3, and it works again

JoergAtGithub (changes requested):
I compared this with the original implementation in the 2.3 branch, and this original implementation returned undefined by default.
And for the case of this.activeDeck and group equal "deck1" ord "deck2" it always used this default branch:
https://github.com/mixxxdj/mixxx/blob/691596c177a80d2420a29c2d9273364921d9a21e/res/controllers/common-hid-packet-parser.js#L825-L848

ywwg:
Yes but this was the problem with the original implementation -- it required an allowlist of groups in order to return the passed-in group name. Restoring that behavior would re-break the function. Instead, my proposal is: only return a resolved deck if the group is like "deck.*" AND activedeck is set.  Otherwise, return the passed-in group name as-is.  (As I explained in the other PR, if that is a problem, then there will be an error in the log like "unknown group deck1" -- I don't think it's this function's job to do correctness checking.  it should just do a best-effort to translate the name and give up if it can't)

ywwg:
The reason the mapping broke is because the original function had an allowlist for "known groups" and [Main] was not one of them. But then as we worked on it we realized there were more groups that also were not in that list.  As I looked at it, there was no reason to have that list at all and it could be safely removed.  In my opinion the only thing that function should do is:

* if the group is like regex "deck[12]?" and activeDeck is set, return the resolved group name
* return the passed-in group name.

JoergAtGithub:
Where do you see the sense, of returning "deck1" or "deck2" ? Isn't it the purpose of the function to resolve these two cases to proper channel groups? If a user passes a deck string to a CO it will fail anyway. But if it returns `undefined` as in the original Mixxx 2.3 implementation, the user has to check only for this one error value and not for three (deck1, deck2 and undefined).

JoergAtGithub:
Superseded by #12406

PR #12213, opened 2023-10-24
Search related Tracks: add checkboxes + Search button

### LINKED ISSUE #12211: Search related Tracks: allow to combine filters
Just an idea while reading #5655 and other 'find similar tracks' feature requests.

The 'Search related Tracks' menu could be 'combo items':
checkboxes with the existing actions

* Clicking the action (text) would instantly trigger a search with the respective query
* Checking one or more items would enable a 'Search' button at the bottom of the menu
* Clicking 'Search' would combine the queries and trigger a search

This would, for example, simplify searching for compatible keys **and** compatible BPM, without having to combine (copy/paste) queries manually.
The current workflow (clicking one specific action) would not be affected.

Implementation can be adopted from Crates menu (QWidgetAtcion with QCheckBox).

ronso0:
Closes #12211 
Implementation adopted from **Crates** menu.

Menu items are checkboxes now, so there are two ways to search for tracks:
* single filter (as before):
click an action (text), or press `Return` on a selected checkbox, to trigger its search
* combine filters:
   * check multiple actions (click `[ ]` box or press `Space` on a selected item
   * **Search selected** is enabled if at least one box is checked
   * click **Search selected** (or press `S`, or `Return` on the selected action) to search with all checked filters

image

While this may actually go into 2.4 I'm targeting `main` because this adds yet another tr string ("Search selected").


ronso0:
Now clazy is inhibited by clang-tidy failing on unrelated files :shrug:

uklotzde:
I only consider *key*, *bpm*, and *genre* as useful conjunctions. Everything else is probably too restrictive in the context of DJing. Even combining the *genre* filter with others is too restrictive if you encode multiple genres in the field as I do.

Remembering and recalling the last checkmarks is not helpful. I often switch between different filters, e.g. artist and directory/album. That would require lots of extra clicks.

Clicking at least twice for selecting a filter before actually starting the search is also inconvenient.

To me this extension seems more like a workaround for the absence of dynamic filters and smart playlists. Using the properties of the selected track instead of hard-coded values for the filter conditions of a smart playlist would be a new concept.

ronso0:
> Clicking at least twice for selecting a filter before actually starting the search is also inconvenient.

Why twice?
> * single filter (as before):
      * click an action (text), or press Return on a selected checkbox, to trigger its search
> 
> * combine filters:
       * check multiple actions (click [ ] box or press Space on a selected item
       * click Search (or press S, or Return on Search) to search with all checked filters

uklotzde:
If no additional clicks are required for searching with a single condition then this extension is reasonable.

uklotzde:
The last item should be renamed to "Search checked" and only be enabled if one or more check boxes are checked.

ronso0:
> only be enabled if one or more check boxes are checked.

That's how it's implemented, disabled by default.

"Search checked" is indeed better, thanks.

daschuer:
This works good. However with Qt5 I don't see the magnifier icon. 
The workflow is usable, but it is actually a pop up window with a button disguised as a menu. 

Did you consider to give the "Search selected" menu entry a button style.
This would allow to "test" checkboxes without to select them again after no result and reopening then menu. 

Now building with Qt6

daschuer:
Also no magnifier icon with Qt6. 
This is no blocker though.

ronso0:
> The workflow is usable, but it is actually a pop up window with a button disguised as a menu.

Indeed. Thinking this further, the ideal solution would be a 'Filter' popup attached to the searchbar with customizable rows
[tag selector] [contains|equal|not] [text]
with 'Pin search' and other features we already discussed.
In case of "Search related.." that would be populated with the selected checkboxes/filters.

ronso0:
> Did you consider to give the "Search selected" menu entry a button style.

I'm not sure if that really makes the situation more clear.

ronso0:
> This would allow to "test" checkboxes without to select them again after no result and reopening then menu.

Which use case do you intend to cover?
Either the selected filters yields valuable results or not, and if not users may refine the query manually.

ronso0:
> Also no magnifier icon with Qt6.

Oh, the screenshot is from a previous state.
I'll try to restore it.

src/widget/wsearchrelatedtracksmenu.cpp
```diff
@@ -332,4 +354,83 @@ void WSearchRelatedTracksMenu::addActionsForTrack(
...
+        // toggling the checkbox happens as usual.
+        QCheckBox* box = qobject_cast<QCheckBox*>(obj);
+        if (box) {
+            QMouseEvent* me = static_cast<QMouseEvent*>(e);
+            VERIFY_OR_DEBUG_ASSERT(me) {
+                return true;
+            }
+            QStyleOptionButton option;
+            option.initFrom(box);
+            auto pStyle = box->style();
```
acolombier, line 383: clang-tidy

```suggestion
            auto *pStyle = box->style();
```

acolombier (commented):
Just a small nit from clang-tidy 

Couple of feedback on UI - looks like the checkbox doesn't render when selected. ~Also, it be great if the menu would stay open till it loose the focus after the first click, so you can easily select key + BPM for example.~ (edit: realise I had to click the checkbox. Still would be great, but nice to have only
Edit 2: Realised that the `Search related` at the bottom is meant to be a submit button. Make sense now)

image

Note that I merged `main` as I needed TagLib 2.0 support, maybe why these issue didn't occur before? 

Love the feature btw! So many time I struggle to filter key + bpm and end up having to cheat with sorting... My last wish would be a configurable `kRelativeBpmRange` or the the ability to define the range :D

acolombier:
A bit of an issue with controller compatibility:
I often search related track using the controllers (`[Library],focused_widget`/`[Library],GoToItem`), it'd bee nice if that could be compatible with that. I guess we need to make `GoToItem` trigger the checkbox rather than the action?

acolombier:
> You may add an extra `FocusWidget` type, e.g. `SearchMenu`, [here](https://github.com/mixxxdj/mixxx/blob/153899a4762d6a8bafc3d437e6a26ce707d1401b/src/library/librarycontrol.cpp#L826) and use Qt::Key_space for that [here](https://github.com/mixxxdj/mixxx/blob/153899a4762d6a8bafc3d437e6a26ce707d1401b/src/library/librarycontrol.cpp#L996).

Sound good. I might give that a go one its merged so that feature is available mouse/click-free.

> In main you can set that range in Pref -> Library. It's used for the fuzzy BPM search and for the "Search related" menu.

Humm looks it might not work.. I'll create an issue for that.

Tested and search button now works fine on Qt6!

ronso0:
> > You may add an extra `FocusWidget` type, e.g. `SearchMenu`, [here](https://github.com/mixxxdj/mixxx/blob/153899a4762d6a8bafc3d437e6a26ce707d1401b/src/library/librarycontrol.cpp#L826) and use Qt::Key_space for that [here](https://github.com/mixxxdj/mixxx/blob/153899a4762d6a8bafc3d437e6a26ce707d1401b/src/library/librarycontrol.cpp#L996).
> 
> Sound good. I might give that a go one its merged so that feature is available mouse/click-free.

I just implemented it :laughing: 
Will open a separate PR for it.

ronso0:
> ~~Humm looks it might not work.. I'll create an issue for that~~ Works just fine!

Can you try to find out why it didn't work? Maybe with a clean config?

acolombier:
No, it's was just my eyes that were failing me :sweat_smile: 
Changing the setting was leading the criteria to move from `98~102` to `92~108` which for some reason I thought was the same value - that `2/8` swap tricked me!

daschuer:
> This would allow to "test" checkboxes without to select them again after no result and reopening then menu.

> Which use case do you intend to cover?
> Either the selected filters yields valuable results or not, and if not users may refine the query manually.

It is just the minor issue that the "refine" is more like a "redo", because after selecting "search related" the menu is closed and you need to repopulate the check boxes before manually on a second attempt.

(Just an idea ...)

ronso0:
Yes, the clearing can be a bit annoying, I already added a TODO inline
https://github.com/mixxxdj/mixxx/pull/12213/files#diff-848ff12a6bbb8c39f28e2a6be33ca2bbbe31f53e6ba550b2f6df7b8bdfaecf1fR227-R230

Saving the check states is a bit more eleaborate, at least if you refer to when opening the entire **track menu** again.

ronso0:
There is/was a slight delay after clicking search actions or the Search button (also a checkbox internally), I have the impression it's longer with this PR. Can someone confirm?

daschuer (approved):
LGTM, @acolombier is your issue also solved?

acolombier:
> @acolombier is your issue also solved

Yeah, the only issue I had was sitting between the chair and the keyboard :sweat_smile: 

Only thing is I think it would be great to keep state between context menu, but that can come as an iteration.

daschuer:
Fine. Thank you.

PR #12214, opened 2023-10-24
WTrackMenu: provide the same features in all deck track widgets

ronso0:
Same features for WTrackProperty, WTrackText & WTrackWidgetGroup.

Btw I'm missing 'Hide' in decks (+ Unhide, for completeness), but that requires quite some changes since it's currently done via the track model.


daschuer (approved):
LGTM, Thank you.

PR #12217, opened 2023-10-25
ParametricEQEffect: Fix typo in description

fwcd:
Just a minor typo I've come across.

### DIFF
--- a/src/effects/backends/builtin/parametriceqeffect.cpp
+++ b/src/effects/backends/builtin/parametriceqeffect.cpp
@@ -22,7 +22,7 @@ EffectManifestPointer ParametricEQEffect::getManifest() {
     pManifest->setAuthor("The Mixxx Team");
     pManifest->setVersion("1.0");
     pManifest->setDescription(QObject::tr(
-            "An gentle 2-band parametric equalizer based on biquad filters.\n"
+            "A gentle 2-band parametric equalizer based on biquad filters.\n"
             "It is designed as a complement to the steep mixing equalizers."));
     pManifest->setEffectRampsFromDry(true);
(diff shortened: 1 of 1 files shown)


ronso0:
I assume this didn't affect translations.
In order to keep existing translations, it is preferred to change the source string in transifex (note that I have no experience with this):
https://help.transifex.com/en/articles/6236820-edit-source-strings-online#h_7e60a9b609

ronso0:
tr wizard @daschuer what is your opinion?

daschuer:
This string is translated in some languages, 
Better a redundant "n" in one language than an untranslated string. 
Let's postpone this fix to 2.5.

daschuer:
... We cannot change the source text, because it is also the lookup key.

ronso0:
Sounds resonable, though did you take a look at the article I linked above?

fwcd:
Can we change the key for all languages?

ronso0:
I don't know, I'm not familiar with the Transifex/lupdate workflow.
Here's the intro to the new'Update source' function added in July this year. it's available via Transifex User Interface (UI), Transifex API and Transifex Go CLI
https://community.transifex.com/t/introducing-enhanced-resource-updating-in-transifex/3503

Here the documentation for the 'new' CLI client:
https://github.com/transifex/cli#pushing-files-to-transifex
> `--keep-transations`: If present, translations of source strings with the same key whose content changes will not be discarded. This can also be set on a per-resource level in the configuration file.

daschuer:
Ok, I will merge that now and fix the translations manually on Transifex.

ronso0:
Thank you very much!

daschuer:
Done.

PR #12219, opened 2023-10-25
Skins: reload default.qss when (re)loading a skin

ronso0:
This makes the workflow of tweaking `/res/skins/default.qss` a bit... faster.
Previously, a ~~restart~~ rebuild was required because deafult.qss is imported via /res/mixxx.qrc

### DIFF
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -315,13 +315,14 @@ void MixxxMainWindow::initialize() {
     QWidget* oldWidget = m_pCentralWidget;
 
     // Load default styles that can be overridden by skins
-    QFile file(":/skins/default.qss");
+    const QString resPath = m_pCoreServices->getSettings()->getResourcePath();
+    QFile file(resPath + "/skins/default.qss");
     if (file.open(QIODevice::ReadOnly)) {
         QByteArray fileBytes = file.readAll();
         QString style = QString::fromLocal8Bit(fileBytes);
         setStyleSheet(style);
     } else {
-        qWarning() << "Failed to load default skin styles!";
+        qWarning() << "Failed to load default skin styles /skins/default.qss!";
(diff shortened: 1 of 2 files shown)


src/mixxxmainwindow.cpp
```diff
@@ -1095,6 +1095,16 @@ void MixxxMainWindow::rebootMixxxView() {
     bool wasFullScreen = isFullScreen();
     slotViewFullScreen(false);
 
+    // Reload default styles that can be overridden by skins
+    QFile file(":/skins/default.qss");
+    if (file.open(QIODevice::ReadOnly)) {
+        QByteArray fileBytes = file.readAll();
+        QString style = QString::fromLocal8Bit(fileBytes);
```
daschuer, line 1102: I don't think that fromLocal8Bit is correct here, because default.qss has always the same encoding on all our targets. 
Is it Latin1 or Utf8?

daschuer (changes requested):
I have found an issue that is also present in: 
https://github.com/mixxxdj/mixxx/blob/2db33fd55d3b048e0254acd21c73c1856e30bad3/src/mixxxmainwindow.cpp#L323
Can you fix both?

src/mixxxmainwindow.cpp
ronso0: ParserM3u uses Utf8 :shrug: 
I moved the redundant parsing to an inline function. We could as well move it to a member function (incl. setStyleSheet()).

ronso0: > an inline function

that was pointless, moved it to a member function.

ronso0:
@daschuer I didn't test properly before. Now default.qss is truly reloaded without a rebuild or restart.
Please take a look.

daschuer (approved):
LGTM, thank you.

Issue #12109, opened 2023-10-14
allow pitch range per deck


ronso0:
Currently the selected rate range is applied to all decks, even though it can already be set per deck, for example in controller scripts.
So this is actually only about the Preferences:
* rate range selector for main decks 1-4
* rate range selector for samplers?

Swiftb0y:
I think this is related to the more general issue where we have settings/COs per Deck but people also want a global CO/setting. In this case its the case that people want more fine-grained control per deck easily. In the case of selecting the QuickEffect, its the other way around... I think in order to solve this elegantly, we need some abstraction that unifies both cases.

ronso0:
Could you explain that? I understand the situation, though what kind of abstraction are you referring to?
What other controls that come to mind?

Swiftb0y:
In this case I just mean a general solution that doesn't have to be reimplemented for every single CO. Something like `GroupCO(COKey groupCOname, double nonUniformVal, QList<COKey> groupMembers)`, where `groupCOname` is the name of the overarching CO, `nonUniformVal` is the value of the CO when COs of the `groupMembers` don't agree with each other, and `groupMembers` is a collection of COs within. 
The behavior of the resulting CO is that its whatever value any member of `groupMembers` is when they all agree with each other, and `nonUniformVal` if they differ.

COs that come to mind would be most `[QuickEffectRack1_[ChannelI]]`, for compatibility with the global QuickEffects concept of most controllers. 

This feature applied to `[ChannelN], rateRange` would simplify the implementation of the current global rateRange implementation in the preferences. Though it doesn't directly solve the original issue.

ronso0:
Ok, so like #11378. Got it, thx.

Swiftb0y:
yes, exactly, just as a general implementation. I don't know when/whether I'll finish #11378.

Issue #12108, opened 2023-10-14
feature request: RANGE button or similar feature to switch pitch range 


deejay-girl:
### Feature Description

Hi

A useful button in Serato Pro DJ:

![275264015-5eedb3b3-dde2-49f6-abb3-f4dd1a0cc124](https://github.com/mixxxdj/mixxx/assets/147563220/d7eb7f06-48db-417b-94b0-4bee5147816b)

Look at the button "RANGE" (left, bottom): it could be very useful to add this button in Mixxx. 

- If you're mixing and you see "oh no, my pitch range is too small to be able to adjust"
- Or the other way: the pitch range is set to high and you want to more "fine tune" the pitch for a specific track by switching to a lower pitch range

=> you don't want to go to settings to adjust the pitch range for each track you need to adjust the pitch range. 
 
The RANGE button switches between all available pitch ranges and in this way you can quickly choose another range (you see the pitch range immediately in the Spinny, at the right side)

Alternative: in order not to add a RANGE button for this, the pitch range switch could be implemented by a right click with the mouse on the pitch slider area.

deejay-girl:
An addition thought: is it possible in Mixxx to adjust pitch slider range for EACH deck seperately?
Because this feature request also means this. I doesn't make sense to adjust pitch range of deck 1 while all other decks are changed too.

ronso0:
Actually the pitch range can already be set per deck independently, for example in controller scripts, it's just that this setting isn't exposed in Decks preferences (additionally, when opened the range of deck 1 is expected to be the same for all decks, and when clicking Apply that is set for all decks again).
I filed #12109 

Related: #8167 

Btw, right-click on sliders or knobs (or double-click, Mixxx 2.4) is 'reset'.

daschuer:
I am still very much in favor of the rate ultra slider. 
For my understanding it fits that best to the use case of having on one hand a fine trim for the normal rate slider, but also have the power to do magnitude changes for effects or to match with doubled tempo or such.

An adjustable range would not work well, because you will loose the fine adjustment behavior if you have conditionally picked a 100% range to double a tempo. 

@deejay-girl: Can you confirm that? What exactly is you use case? Please describe your task step by step. How is the controller involved?

The two sliders can be found here: 
https://github.com/mixxxdj/mixxx/pull/1767

[...]

PR #12221, opened 2023-10-25
new rate range controls: rateRange_next / _prev / _select

### LINKED ISSUE #11068: `next_/prev_rateRange` control to cycle through `Preferences > Decks > Slider Range` options
### Feature Description

Often times controllers seem to have a way of displaying the current rateRange but are flexible in the ranges they allow. They often just copy the ranges defined in the Deck preferences and implement their own cycling logic, but that comes with usual disadvantages of unnecessarily duplicated code. So I'd like to propose a `next_rateRange` and `prev_rateRange` control (similar to `next_`/`prev_chain`). 

Mappings which currently implement the same raterange cycling logic are (non exhaustive list):
* https://github.com/mixxxdj/mixxx/blob/0cf07a192e93d7c02fa62273879d1b9f5c8bcde2/res/controllers/Denon-MC7000-scripts.js#L60
* https://github.com/mixxxdj/mixxx/blob/0cf07a192e93d7c02fa62273879d1b9f5c8bcde2/res/controllers/Numark-N4-scripts.js#L27
* https://github.com/mixxxdj/mixxx/blob/0cf07a192e93d7c02fa62273879d1b9f5c8bcde2/res/controllers/Numark-V7-scripts.js#L16
* https://github.com/mixxxdj/mixxx/blob/0cf07a192e93d7c02fa62273879d1b9f5c8bcde2/res/controllers/American-Audio-VMS2-scripts.js#L32

ronso0:
* make `rate_range` a pure **ControlObject**
(no ControlPotmeter with  _up & _down controls etc. which are useless for rate range)
* add `rate_range_next`, `_prev` and `_select` controls, allow to change the range **per deck** with with buttons (controller & GUI) 
* clear the combobox in Preferences -> Decks if users have set individual ranges

- [ ] update documentation

**Nice to have, but outof scope for this initial PR:**
* add per-deck comboboxes to the preferences
* control to set the range for all per-deck controls, see 'general wrapper' suggestion pitched by @Swiftb0y https://github.com/mixxxdj/mixxx/issues/12109#issuecomment-1765790312

Closes #11068 
Paves the way for fixing #12108

### DIFF
--- a/src/engine/controls/ratecontrol.cpp
+++ b/src/engine/controls/ratecontrol.cpp
@@ -32,8 +32,17 @@ const double RateControl::kPausedJogMultiplier = 18.0;
 
 // static
 constexpr double RateControl::kDefaultRateRange;
-constexpr double RateControl::kMinRateRange;
-constexpr double RateControl::kMaxRateRange;
+constexpr int RateControl::kMinRateRangePercent;
+constexpr int RateControl::kMaxRateRangePercent;
+QMap<int, QString> RateControl::s_rateRangesPercentAndTrStrings = {
+        {4, tr("4%")},
+        {6, tr("6% (semitone)")},
+        {8, tr("8% (Technics SL-1210)")},
+        {10, tr("10%")},
+        {16, tr("16%")},
+        {24, tr("24%")},
+        {50, tr("50%")},
+        {90, tr("90%")}};
 
 RateControl::RateControl(const QString& group,
         UserSettingsPointer pConfig)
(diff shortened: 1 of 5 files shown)


ronso0:
eslint fails since I touched 'some' controller mappings when mass-renaming `rateRange`.
Maybe I'll revert that.

ronso0:
Just a note regarding the preferences implementation:
Actually, I'd like to move some of the logic out of the preferences by simply making the `group, rate_range` CO persist, exactly like the Quick Effects (and EQs, if someone is actually doing that).
Though, IMO the situation / UX I expect for the per deck / general **rate ranges** differs a bit from per deck / general **Quick Effect** / **EQ**.

**Quick Effect:**
* effect is picked either in the skins or in the preferences
* effect should persist across restarts

**Rate ranges:**
* users pick a default value in the preferences
* individual deck range may be changed on the fly to mix songs of a wide BPM range
* after restart the previous tracks are irrelevant, hence the ranges should be **reset** to the default general value the user selected (as done by some controllermappings' init)

PR #12224, opened 2023-10-26
(fix) Looping: reset loop_end_pos on eject

### LINKED ISSUE #12223: "loop_end_position" does not update on eject
### Bug Description

Discovered while writing a mapping for the reloop mixage series

I'm using the `loop_start_position`  and `loop_end_position` controls to provide users feedback about the status of the loop for a given track. these leds are also updated when a new track is loaded or when a loaded track is ejected. The `loop_end_position` seems to be returning a incorrect value when a track is ejected.

**expected behaviour**
The `loop_end_position` return -1 just like the `loop_start_position` when a track has been ejected.

**Observed behaviour**
- the `loop_end_position` returns the position of the loop end point of the track that has just been ejected instead of -1)
- if a track new track is loaded (with no loop present or previously set) the `loop_end_position` returns the value of when it was last updated. This is usually the loop end point of a previously loaded track

**steps to reproduce**
- bind a simple logging function to the `eject` control, I used the script below
```
var test = {};

test.controlsToFunctions = {
    "eject": function(v) {
        if (v === 0) { 
            print("start " + engine.getValue("[Channel1]", "loop_start_position"));
            print("end " + engine.getValue("[Channel1]", "loop_end_position")); 
        }
    }
}

test.init = function(_id, _debugging) {
    script.bindConnections("[Channel1]", test.controlsToFunctions);
};

test.shutdown = function() {
    script.bindConnections("[Channel1]", test.controlsToFunctions, remove);
};

```
- load a either a new track or a track for which no loop points are present in the data base
  - eject the track, the logging function returns -1 for both `loop_start_position` and `loop_end_position`
- load a track with a loop set or set a new loop
  - eject the track, the function returns -1 for `loop_start_position` and the loop end point for `loop_end_position`
- load another track without any loop points present
  - eject the track, the function returns -1 for `loop_start_position` and the loop end point of the previous track for `loop_end_position`

I have tested it with both 2.3.6 and 2.4 and got the same result

### Version

2.4-beta-199-g8c925617e3

### OS

Fedora 38

ronso0:
Just a quick test if this fixes #12223
It works as expected, but let's run the tests.

In theory:
On eject, `loop_in_pos` was just reset (before `slotLoopEndPos` is called) so this eval would always be true, regardless the invalid requested `loop_end_pos`.
Fall through here and set `loop_end_pos` to -1 a few lines below.
This code hasn't changed for very long, but maybe the callers?

### DIFF
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -1107,8 +1107,8 @@ void LoopingControl::slotLoopEndPos(double positionSamples) {
 
     // Reject if the loop-in is not set, or if the new position is before the
     // start point (but not -1).
-    if (!loopInfo.startPosition.isValid() ||
-            (position.isValid() && position <= loopInfo.startPosition)) {
+    if (position.isValid() &&
+            (!loopInfo.startPosition.isValid() || position <= loopInfo.startPosition)) {
         m_pCOLoopEndPosition->set(loopInfo.endPosition.toEngineSamplePosMaybeInvalid());
         return;
     }
--- a/src/test/looping_control_test.cpp
+++ b/src/test/looping_control_test.cpp
@@ -678,6 +678,14 @@ TEST_F(LoopingControlTest, LoopResizeSeek) {
     EXPECT_FRAMEPOS_EQ(mixxx::audio::FramePos{250}, currentFramePos());
 }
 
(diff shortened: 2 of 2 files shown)


ronso0:
Existing tests passed.
I added another one for checking loop_in/_out are reset on eject.

daschuer (approved):
LGTM, thank you.

PR #12229, opened 2023-10-27
Improve Reverb includes, remove Qt dependency

daschuer:
This does not fix the original issue, but improves compile time a bit.

### DIFF
--- a/lib/reverb/basics.h
+++ b/lib/reverb/basics.h
@@ -43,8 +43,7 @@
 #include <assert.h>
 #include <stdio.h>
 
-#include "util/types.h"
-typedef CSAMPLE sample_t;
+typedef float sample_t;
 
 // NOTE(rryan): 3/2014 Added these for the MSVC build.
 #include <QtGlobal>
@@ -67,7 +66,7 @@ typedef unsigned int uint;
 typedef unsigned long ulong;
 
 /* prototype that takes a sample and yields a sample */
-typedef CSAMPLE (*clip_func_t) (CSAMPLE);
+typedef sample_t (*clip_func_t) (sample_t);
 
 #ifndef max
 template <class X, class Y> X min (X x, Y y) { return x < (X)y ? x : (X)y; }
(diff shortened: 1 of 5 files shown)


lib/reverb/basics.h
```diff
@@ -29,78 +31,16 @@
...
-
 #include <cmath>
-
-#include <stdlib.h>
-#include <string.h>
-
-#include <math.h>
-#include <float.h>
-
+#include <cstring>
```
Swiftb0y, line 35: seems unnecessary

lib/reverb/basics.h
```diff
@@ -29,78 +31,16 @@
 	02111-1307, USA or point your web browser to http://www.gnu.org.
 */
 
-#ifndef BASICS_H
-#define BASICS_H
-
 #include <cmath>
```
Swiftb0y, line 34: can also be removed I think, right?

daschuer:
I have added comments to clarify the reason for the include.

lib/reverb/basics.h
daschuer: No, we need it for M_PI

lib/reverb/basics.h
daschuer: required for memset()

daschuer:
Rebased to make clazy work.

Swiftb0y (commented):
Thank you, but the questions remain.

lib/reverb/basics.h
```diff
@@ -29,78 +31,16 @@
...
-typedef qint64 int64;
-typedef quint64 uint64;
-
-#define MIN_GAIN 1e-6 /* -120 dB */
-/* smallest non-denormal 32 bit IEEE float is 1.18e-38 */
-#define NOISE_FLOOR 1e-20 /* -400 dB */
-
-/* //////////////////////////////////////////////////////////////////////// */
+#include <cmath> // for M_PI
+#include <cstring> // for memset
```
Swiftb0y, line 35: Why is this in the header if neither `M_PI` nor `memset` are used in it?

daschuer: I have moved them closer to usage.

daschuer:
Done.

Swiftb0y (approved):
Thank you. LGTM.

PR #12231, opened 2023-10-27
Preferences waveforms: recall correct waveform type when selecting an overview type

### LINKED ISSUE #12226: Changing the waveform overview type after changing the main waveform type resets the main waveform type
### Bug Description

I set the main waveform type to "Simple (GLSL)" and Ok the preferences.

Then I go back in, and select "Filtered (GLSL)" for the main waveform type, and it changes to the newer non-semitransparent stacked RGB style. (I know "stacked" and "RGB" relate to other waveform styles, but I really don't know how else to simply describe the new style!)

But I then change the waveform overview type, and the main waveform type changes style from "Filtered (GLSL)" to something else.

I Ok the preferences, then I reopen the preferences and the option main waveform type has changed to "RBG (GLSL)"

This happened regardless of what the initial main waveform type is.

Arguably related to #9745 in some way

### Version

2.5-alpha-137-gc2f2f64891 (main)

### OS

Arch btw

ronso0:
Closes #12226 

The waveform type was written to config but the local variable `m_configType` holding this type for updating the widgets after skin (re)load was not updated accordingly.

Note that I consider this just a quick fix.

IMO we should not store the value in `m_configType` but read it from the config in `setWidgetTypeFromConfig` (move the logic from `setConfig()`. AFAICT this is required only in the constructor and in `slotSkinLoaded()`.
A commit that should do th trick is 80bb79c012, though I'm not motivated to verify this actually works under all conditions / on all OS. If someone wants to check that and do the testing, please go ahead.
Maybe also call `DlgPrefWaveforms::slotUpdate` when `MixxxMainWindow::rebootMixxxView` succeeded.

### DIFF
--- a/src/waveform/waveformwidgetfactory.cpp
+++ b/src/waveform/waveformwidgetfactory.cpp
@@ -567,6 +567,9 @@ bool WaveformWidgetFactory::setWidgetTypeFromConfig() {
     int empty = findHandleIndexFromType(WaveformWidgetType::EmptyWaveform);
     int desired = findHandleIndexFromType(m_configType);
     if (desired == -1) {
+        qDebug() << "WaveformWidgetFactory::setWidgetTypeFromConfig"
+                 << " - configured type" << static_cast<int>(m_configType)
+                 << "not found -- using 'EmptyWaveform'";
         desired = empty;
     }
     return setWidgetTypeFromHandle(desired, true);
@@ -574,7 +577,8 @@ bool WaveformWidgetFactory::setWidgetTypeFromConfig() {
 
 bool WaveformWidgetFactory::setWidgetTypeFromHandle(int handleIndex, bool force) {
     if (handleIndex < 0 || handleIndex >= m_waveformWidgetHandles.size()) {
-        qDebug() << "WaveformWidgetFactory::setWidgetType - invalid handle --> use of 'EmptyWaveform'";
+        qDebug() << "WaveformWidgetFactory::setWidgetTypeFromHandle"
+                    " - invalid handle --> using 'EmptyWaveform'";
         // fallback empty type
         setWidgetType(WaveformWidgetType::EmptyWaveform);
         return false;
@@ -582,7 +586,8 @@ bool WaveformWidgetFactory::setWidgetTypeFromHandle(int handleIndex, bool force)
 
     WaveformWidgetAbstractHandle& handle = m_waveformWidgetHandles[handleIndex];
     if (handle.m_type == m_type && !force) {
-        qDebug() << "WaveformWidgetFactory::setWidgetType - type already in use";
+        qDebug() << "WaveformWidgetFactory::setWidgetTypeFromHandle - type"
+                 << handle.getDisplayName() << "already in use";
         return true;
     }
 
@@ -590,7 +595,6 @@ bool WaveformWidgetFactory::setWidgetTypeFromHandle(int handleIndex, bool force)
     setWidgetType(handle.m_type);
 
     m_skipRender = true;
-    //qDebug() << "recreate start";
 
     //re-create/setup all waveform widgets
     for (auto& holder : m_waveformWidgetHolders) {
@@ -619,7 +623,6 @@ bool WaveformWidgetFactory::setWidgetTypeFromHandle(int handleIndex, bool force)
     }
 
     m_skipRender = false;
-    //qDebug() << "recreate done";
     return true;
 }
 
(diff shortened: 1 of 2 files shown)


src/waveform/waveformwidgetfactory.cpp
```diff
@@ -1163,6 +1174,16 @@ int WaveformWidgetFactory::findHandleIndexFromType(WaveformWidgetType::Type type
     return index;
 }
 
+QString WaveformWidgetFactory::getDisplayNameFromType(WaveformWidgetType::Type type) {
+    for (int i = 0; i < m_waveformWidgetHandles.size(); i++) {
```
daschuer, line 1178: This can become a range based loop.

ronso0: done.

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -537,6 +538,8 @@ bool WaveformWidgetFactory::setWidgetType(
         WaveformWidgetType::Type type,
         WaveformWidgetType::Type* pCurrentType) {
     if (type == *pCurrentType) {
+        qDebug() << "WaveformWidgetFactory::setWidgetType - type"
```
ywwg, line 541: this does not seem like a debug statement we need -- ok to delete it?

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -564,25 +569,33 @@ bool WaveformWidgetFactory::setWidgetType(
 }
 
 bool WaveformWidgetFactory::setWidgetTypeFromConfig() {
+    // qDebug() << "WaveformWidgetFactory::setWidgetTypeFromConfig";
     int empty = findHandleIndexFromType(WaveformWidgetType::EmptyWaveform);
     int desired = findHandleIndexFromType(m_configType);
     if (desired == -1) {
+        qDebug() << "WaveformWidgetFactory::setWidgetTypeFromConfig"
```
ywwg, line 576: (whereas, this one seems useful)

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -1153,14 +1166,23 @@ WaveformWidgetType::Type WaveformWidgetFactory::findTypeFromHandleIndex(int inde
 }
 
 int WaveformWidgetFactory::findHandleIndexFromType(WaveformWidgetType::Type type) {
-    int index = -1;
-    for (int i = 0; i < m_waveformWidgetHandles.size(); i++) {
-        WaveformWidgetAbstractHandle& handle = m_waveformWidgetHandles[i];
+    int index = 0;
+    for (const auto& handle : std::as_const(m_waveformWidgetHandles)) {
```
ywwg, line 1170: if you're going to maintain an index like this, you might as well keep the original style of loop (and use .get() to get a const ref instead of mutable)

ywwg (commented):
I made some style notes, but the basic code looks good! thanks for this fix

ronso0:
Thanks for your review. Feel free to edit this branch, I'm ill and probably won't fix it this year.

ywwg:
> Thanks for your review. Feel free to edit this branch, I'm ill and probably won't fix it this year.

<3 hope you feel better and happy new year

ronso0:
Thanks, doing better already.
Either you edit this as you desire, or I find time to do this, whatever happens earlier. After all I need some distraction from the usual xmas madness..

src/waveform/waveformwidgetfactory.cpp
ronso0: Good point, I reverted this.
Though this is a QVector, not sure what you mean by `get()`.

ronso0:
I addressed your comments with !fixups. Please check and give :+1: if it's okay, then I'll squash and force-push.

ywwg (approved):

ronso0:
Draft again to avoid merging the separate fixups.

ywwg:
Let me know how you want to treat this PR, not quite clear how you wanted to treat the fixups?

ronso0:
I want to squash them but give @daschuer a chance to review them separately.

ywwg:
Can you mark this as ready for review, and then add a note that it needs to be squashed before merge?  that way it gets on his radar

ronso0:
@daschuer I pushed to fixups and left them separate for now so you can easily review those.
Please give me a hint when you consider this okay, then I'll squash and force-push.
Thanks!

daschuer (approved):
I can confirm the issue is fixed and the Code looks good. Thank you. 
Please give a hint once I can press merge.

ronso0:
Ready!

daschuer:
Thank you.

PR #12252, opened 2023-11-03
make WHotcueButton learnable with the MIDI Wizard

### LINKED ISSUE #10842: allow point-and-click MIDI learning of Hotcue buttons
Hotcue buttons can not be mapped with the MIDI Learning Wizard with point-and-click, because the widget connections of [`WHotcueButton` are created by the class itself](https://github.com/mixxxdj/mixxx/blob/f12c6393a12a930d02cc38faaba7c38ea799bab1/src/widget/whotcuebutton.cpp#L61-L76), not by `LegacySkinParser::setupConnections`, so the controls are not available in the `ControllerLearningEventFilter` (actually only `activate` and `enabled` are controls, remove and the context menu are handled in `mousePressEvent`)

Note: this is not a big deal since the controls are available in the controls menu, though to accelerate mapping point-and-click would be nice.

ronso0:


JoergAtGithub:
Tested this (as far as I could without MIDI hardware) and it works as described! Thank you!

PR #12255, opened 2023-11-04
Fix Qt6/QML build

uklotzde:


uklotzde:
I wonder how these build errors could slip through unnoticed :roll_eyes:

JoergAtGithub:
These includes look correct to me! I also do not understand why CI and several local builds passed. Thank you!

Issue #12256, opened 2023-11-04
Removing a track from history list set all play counters to 0


MarcPlace:
### Bug Description

Yesterday I fell over a nasty bug in Mixxx 2.4.0 using e3e0f6e669457c7de78f1e4fa5bee695eee42fda, but it also happened a few weeks ago. Unfortunately I didn't know what was the cause for it so I did some investigations today.

Track list looks like this ("Played" counter is set to 1 or more):
![shot01](https://github.com/mixxxdj/mixxx/assets/71196229/273ecc75-c3fa-4ed3-af6d-7715e8ddef90)

Then I go to the "History" crate and remove a track from the list:
![shot02](https://github.com/mixxxdj/mixxx/assets/71196229/1e5423f0-4cf2-441d-8c1b-5ae18829e0cf)

After restarting Mixxx, all "Played" counters are now set to 0 (funny: the track I removed from the list seems to be OK and the counter has still 3 plays):
![shot03](https://github.com/mixxxdj/mixxx/assets/71196229/ac78fc9a-4a5f-4ef8-9f33-cc666a589549)

Luckily I had a backup of "mixxxdb.sqlite" file with me, so I could quickly restore all the settings again (keep this in mind if you try to reproduce this issue :-) )

Hope this helps for digging into it to confirm this issue.

Kind regards,
Marco

### Version

2.4.0

### OS

Gentoo Linux

uklotzde:
Good job and excellent bug report! You were able to reproduce #12046 which was reported earlier.

@Swiftb0y Mark #12064 as a duplicate and continue here?

MarcPlace:
Sorry, I've crawled the bug reports before of course, but haven't found anything similar to my issue. But #12046 looks indeed like something related to mine. Glad that I could help you.

uklotzde:
The query `updateNotPlayed` in `TrackDAO::updatePlayCounterFromPlayedHistory()` seems to be the cuplrit.

uklotzde:
I'll fix it. Looks like a closing brace in the SQL statement is misplaced.

uklotzde:
The bug only affects the code path for SQLite >= 3.33.0 that makes use of _UPDATE-FROM_.

Swiftb0y:
> The bug only affects the code path for SQLite >= 3.33.0 that makes use of UPDATE-FROM.

That would likely explain why the devs on ubuntu 20.04. did not experience this issue. 

[...]

PR #12258, opened 2023-11-04
Fix update of play count after removing tracks from history playlist

### LINKED ISSUE #12046: `library.timesplayed` lost for entire database after upgrade
### Bug Description

I'm not sure in what commit exactly (I don't have the time to bisect right now), but somewhere the `timesplayed` for all tracks in my 6K track library have been nuked. Not a huge issue for me as I have backups as well, but not something that can stay in 2.4 in the release. I have confirmed that the issue affects the DB and not the hardware (as `SELECT COUNT(*) FROM library where timesplayed > 0` is only the dozen or so played recently. I'm not sure what action caused, nor do I think I have the log from when this happened as I only realized it after the fact.
Can anybody reproduce?
I'll tag this as critical for now as this counts as data loss imo.

### Version

2.4-beta-163-g420b678c5e-modified (fix/gh11923)

### OS

Fedora Linux 38

uklotzde:
@Swiftb0y: Fixes #12046 and #12256.

### DIFF
--- a/tools/mixxxdb_cleanup.sql
+++ b/tools/mixxxdb_cleanup.sql
@@ -48,6 +48,27 @@ DELETE FROM PlaylistTracks WHERE track_id NOT IN (SELECT id FROM library);
 DELETE FROM track_analysis WHERE track_id NOT IN (SELECT id FROM track_locations);
 
 -----------------------------------------------------------------------
+-- Restore the play count columns in the library table from the      --
+-- history playlists (requires SQLite v3.33.0 or newer)              --
+-----------------------------------------------------------------------
+
+UPDATE library
+SET timesplayed=q.timesplayed,last_played_at=q.last_played_at
+FROM (
+    SELECT
+    PlaylistTracks.track_id as id,
+    COUNT(PlaylistTracks.track_id) as timesplayed,
+    MAX(PlaylistTracks.pl_datetime_added) as last_played_at
+    FROM PlaylistTracks
+    JOIN Playlists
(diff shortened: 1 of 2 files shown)


tools/mixxxdb_cleanup.sql
```diff
@@ -47,6 +47,26 @@ DELETE FROM PlaylistTracks WHERE track_id NOT IN (SELECT id FROM library);
...
+SET timesplayed=q.timesplayed,last_played_at=q.last_played_at
+FROM (
+    SELECT
+    PlaylistTracks.track_id as id,
+    COUNT(PlaylistTracks.track_id) as timesplayed,
+    MAX(PlaylistTracks.pl_datetime_added) as last_played_at
+    FROM PlaylistTracks
+    JOIN Playlists
+    ON PlaylistTracks.playlist_id=Playlists.id
+    WHERE Playlists.hidden=2
```
Swiftb0y, line 65:
```suggestion
    -- PlaylistDAO::PLHT_SET_LOG=2
    WHERE Playlists.hidden=2 
```

Swiftb0y (commented):
I'm not well versed in SQL so I would've never spotted this issue during code review. Looking very carefully at your changes, I see how the previous code was buggy and this is fixing it. Thank you. Almost LGTM

Swiftb0y (approved):
Thank you very much for this fix. I doubt this would've gotten fixed by anyone else. LGTM

Swiftb0y:
Waiting for CI, just to be sure.

PR #12263, opened 2023-11-05
SettingsManager: remove unused initializeDefaults()

ronso0:
noticed when I stumbled upon `[Config] Path` in mixxx.cfg which is supposed to store the last used resource path, though it's never read.


Swiftb0y (approved):
LGTM. Thank you.

PR #12265, opened 2023-11-05
feat(control): Do not show CO aliases in developer tools window

Holzhaus:
This hides aliases in the control model. Aliases are only for backwards compatibility, so this reduces the risk that mapping developers use these COs for new mappings.

This also removes a bunch of CO deprecation warnings when opening the developer tools window or starting the QML interface.

### DIFF
--- a/src/control/controlmodel.cpp
+++ b/src/control/controlmodel.cpp
@@ -15,8 +15,6 @@ ControlModel::ControlModel(QObject* pParent)
     // Add all controls to Model
     const QList<QSharedPointer<ControlDoublePrivate>> controlsList =
             ControlDoublePrivate::getAllInstances();
-    const QHash<ConfigKey, ConfigKey> controlAliases =
-            ControlDoublePrivate::getControlAliases();
 
     for (const QSharedPointer<ControlDoublePrivate>& pControl : controlsList) {
         if (!pControl) {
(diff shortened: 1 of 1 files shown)


Swiftb0y:
I'm concerned that this causes more obstruction than good. The developer tools window should be the ground truth for all COs. if we suddenly hide some that'll probably introduce confusion. For example if somebody was troubleshooting a legacy using the alias CO. Not being able to observe the value of some CO just because the mapping wasn't updated is sure to create confusion. 
I'm not familiar with Qt's Model-APIs but is it possible to make it optional to hide the aliases? Then the dev window could hide them by default but a simple checkbox could show them again in case its needed?

Reflecting on this, it makes me question how realistic this case is. How likely is it for this circumstance to occur? Since we did a mass-replacement of the aliases, probably not very high, right?

Holzhaus:
There are some old mappings that still use aliases from before the recent refactorings (some even try to use removed COs without replacement). I think removing aliases from the control model makes it easier to find COs in the table, this is why I proposed this change. The alternative would be to add a new `ControlFlag::NoAliasWarning` and use it for adding alias controls to the model to silence the warnings. But that would be a bit more complex and also makes it more likely to use deprecated COs. There is still a warning for the deprecated control in the logs, so if someone tries to debug an old mapping, it probably makes sense to fix this warning first anyway.

Swiftb0y (approved):
I agree with this rationalization. LGTM.
Waiting with merging for other to have a chance to object if they feel strongly enough about it.

ronso0 (approved):

Swiftb0y:
confirmed, and its not easy to fix either...

ronso0:
A temp list to detect duplicates in the ctor could work.

ronso0:
something like this #12707

PR #12268, opened 2023-11-05
Fix Qt 6.6 build

uklotzde:


uklotzde:
`QColor::setNamedColor()` is also deprecated and the build fails.

JoergAtGithub:
LGTM! Thank you!

PR #12282, opened 2023-11-09
Effects: read effect data for decks added after initial setup

### LINKED ISSUE #12277: no Quick Effect loaded for decks 3 & 4 by default
With a fresh install there are no Quick Effects loaded for decks 3 & 4.

* start Mixxx the first time (LateNight skin)
* Filter is loaded for decks 1/2
* toggle 4 decks: no effects for decks 3/4
* close, restart
* all decks have Filter loaded

I think the default should be `Filter` for all decks, without a need to configure anything.

The issue becomes obvious (no EQs on decks 3/4) with #11695 which also stores the EQ effects in effects.xml

I'm not sure whether that was always the case. I think it may have been the EQ Preferences which (unintentionally) have set the default for all decks before #11527.
___
`EffectsManager::readEffectsXml` is responsible for initialising/restoring Quick Effects, though it only cares about existing decks when it's called (once) via `CoreServices::initialize` -> `EffectsManager::setup`. The deck count is still the initial 2 then, the default count `PlayerManager`  is initialized with.
If any decks are added later on, e.g. when the default (4-deck) skin LateNight is loaded, effects.xml is not read again and the Quick Effect slots remain empty.
On next start it uses the `deck_count` attribute from `soundconfig.xml`.

IMO only `EffectsManager` should be responsible for effect defaults, not the Preferences.
Simply reading effects.xml again when a deck is added doesn't seem to work right away.
I'll check if it works with a slight refactoring of `EffectsManager`.

ronso0:
Closes #12277 

> `EffectsManager::readEffectsXml` is responsible for initialising/restoring Quick Effects, though it only cares about existing decks when it's called (once) via `CoreServices::initialize` -> `EffectsManager::setup`. The deck count is still the initial 2 then, the default count `PlayerManager`  is initialized with.
> If any decks are added later on, e.g. when the default (4-deck) skin LateNight is loaded, effects.xml is not read again and the Quick Effect slots remain empty.
> On next start it uses the `deck_count` attribute from `soundconfig.xml`.
> 
> IMO only `EffectsManager` should be responsible for effect defaults, not the Preferences.

Fixed by creating a separate, stripped down version of `readEffectsXml` that extracts only the Quick Effect for one deck, and call that each time a deck is added after the initial `setup()` call.

### DIFF
--- a/src/effects/effectsmanager.cpp
+++ b/src/effects/effectsmanager.cpp
@@ -27,7 +27,8 @@ EffectsManager::EffectsManager(
         : m_pConfig(pConfig),
           m_pChannelHandleFactory(pChannelHandleFactory),
           m_loEqFreq(ConfigKey(kMixerProfile, kLowEqFrequency), 0., 22040),
-          m_hiEqFreq(ConfigKey(kMixerProfile, kHighEqFrequency), 0., 22040) {
+          m_hiEqFreq(ConfigKey(kMixerProfile, kHighEqFrequency), 0., 22040),
+          m_initializedFromEffectsXml(false) {
     qRegisterMetaType<EffectChainMixMode>("EffectChainMixMode");
 
     m_pBackendManager = EffectsBackendManagerPointer(new EffectsBackendManager());
@@ -72,6 +73,12 @@ void EffectsManager::setup() {
     addStandardEffectChains();
     addOutputEffectChain();
     // EQ and QuickEffect chain slots are initialized when PlayerManager creates decks.
+    // Now read effects.xml to load defaults or restore previous states of standard
+    // effect chains and QuickEffect chain, though only for decks that have been
+    // created by now. For decks added later on see addDeck().
+    // Note: flip this bool now so any deck potentially being added while
+    // readEffectsXml() is running is also initialized.
+    m_initializedFromEffectsXml = true;
     readEffectsXml();
 }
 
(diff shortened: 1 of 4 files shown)


src/effects/effectsmanager.cpp
```diff
@@ -202,6 +215,25 @@ void EffectsManager::readEffectsXml() {
     m_pVisibleEffectsList->readEffectsXml(doc, m_pBackendManager);
 }
 
+void EffectsManager::readEffectsXmlSingleDeck(const QString& deckGroup) {
+    QDir settingsPath(m_pConfig->getSettingsPath());
```
daschuer, line 219: Normally we prefer initialization with = `QDir settingsPath = m_pConfig->getSettingsPath();`
To have a clear visible distinguishing from a function call. 

Recently we have discussed to prefer initialization lists like:  
`QDir settingsPath{m_pConfig->getSettingsPath()};`

This is just a hint, because that never made the way into our style guides.

src/effects/presets/effectchainpresetmanager.cpp
```diff
@@ -784,6 +789,36 @@ EffectsXmlData EffectChainPresetManager::readEffectsXml(
...
+    auto pQuickEffectChainPreset = getDefaultQuickEffectPreset();
+
+    // Read name of last loaded QuickEffect preset
+    QDomElement quickEffectPresetsElement =
+            XmlParse::selectElement(root, EffectXml::kQuickEffectChainPresets);
+    QDomNodeList quickEffectNodeList =
+            quickEffectPresetsElement.elementsByTagName(
+                    EffectXml::kChainPresetName);
+    for (int i = 0; i < quickEffectNodeList.count(); ++i) {
+        QDomElement presetNameElement = quickEffectNodeList.at(i).toElement();
```
daschuer, line 805: This can become a range based loop, to get rid of i.

daschuer (commented):
LGTM, just code style hints.

src/effects/effectsmanager.cpp
ronso0: ackknowldeged.
It's copy/paste from `readEffectsXml()`, and actually it's mixed in the vaious effect managers already:
QString xyPath(m_pConfig->getSettingsPath() + ...);
QDir xyDir(xyPath);
vs
QFile xyFile(m_pConfig->getSettingsPath() + ...);

src/effects/presets/effectchainpresetmanager.cpp
ronso0: There's QList<type>::iterator, but none for QDomeNodeList.

daschuer (approved):
Ok than let's merge. Thank you. LGTM.

PR #12284, opened 2023-11-10
MixxxApplication: Support linking Qt statically on Linux

fwcd:
Currently, Mixxx cannot be built at all using static Qt on Linux, since it will always error at compile time with a message indicating the a plugin for the current platform is missing (see e.g. [this build](https://github.com/fwcd/m1xxx/actions/runs/6825053309/job/18562147900)).

While using the system-packaged (and dynamically linked) Qt is generally the preferred way, there are use cases for statically linking Qt (e.g. portable or embedded builds), therefore it would be nice if we could at least support building Mixxx this way.

Unfortunately Linux does not have a single display API, instead there are several choices for platform plugins on Linux, as described in the Qt documentation:

- https://doc.qt.io/qt-6/qpa.html

This patch chooses to hardcode the X11/XCB plugin for static Qt builds, which is certainly not ideal, especially since Wayland seems to increasingly be the preferred backend, but seems to be the only plugin supported by vcpkg's `qtbase`:

- https://github.com/mixxxdj/vcpkg/blob/2.5-rel/ports/qtbase/portfile.cmake

In the spirit of having at least something that works, I would consider this an acceptable tradeoff for now. In the future we could e.g. add a configuration flag for selecting a plugin at build-time (including e.g. the Wayland plugin). Any thoughts?

### DIFF
--- a/src/mixxxapplication.cpp
+++ b/src/mixxxapplication.cpp
@@ -28,6 +28,8 @@ Q_IMPORT_PLUGIN(QWindowsVistaStylePlugin)
 #elif defined(Q_OS_MACOS)
 Q_IMPORT_PLUGIN(QCocoaIntegrationPlugin)
 Q_IMPORT_PLUGIN(QMacStylePlugin)
+#elif defined(Q_OS_LINUX)
+Q_IMPORT_PLUGIN(QXcbIntegrationPlugin)
 #else
 #error "Q_IMPORT_PLUGIN() for the current patform is missing"
 #endif


daschuer:
I have no objections.

fwcd:
I'll mark this as ready for review, since it's a pretty trivial PR, but would like to hold off merging before we've tested it with an actual build (which effectively means it's blocked on all the PRs listed in fwcd/m1xxx#48).

fwcd:
Can confirm that this works, tested the ["static" build](https://github.com/fwcd/m1xxx/pull/48) that only links essential and X11-related libraries in a VM:

image

daschuer (approved):
Good job, Thank you. LGTM

ronso0:
Lesson learned: we need to test on all affected platforms next time merging such a change.

ronso0:
@fwcd Do you have an idea how to fix it?

fwcd:
This doesn‘t make much sense to me, do the Ubuntu builds link Qt statically? (I‘d find that very surprising if it were the case) Otherwise we should investigate why `QT_STATIC` is set at build-time…

The condition that this adds would have produced a build error before (if `QT_STATIC` was set on Linux), so I don‘t really understand how that could have happened in the first place.

Still, these kinds of bugs can probably easily go unnoticed if many people still have the X libraries installed.

fwcd:
I don’t have much experience with Ubuntu packaging and Launchpad, so perhaps someone more familiar with that process has an idea?

ronso0:
Hmm, maybe this is not related to #12525 at all, it just seemed to this was the only relevant PR.

fwcd:
If it is a more general issue, it‘s probably something with the way Qt 6 is packaged on Ubuntu

PR #12288, opened 2023-11-11
Fix passing CMAKE_OSX_DEPLOYMENT_TARGET to subprojects

daschuer:
This fixes linker warnings like: 
```
ld: warning: object file (lib/libdjinterop-install/lib/libdjinterop.a(crate.cpp.o)) was built for newer macOS version (11.0) than being linked (10.12)
```


m0dB:
lgtm

PR #12290, opened 2023-11-12
BundleInstall: Enable hardened runtime in ad-hoc-signed macOS builds too

fwcd:
A small follow-up to #12101 and #12138.

Currently we have App Sandbox + Hardened Runtime in notarized builds, but only App Sandbox in unnotarized builds (e.g. in PR builds). In an effort to both streamline the signing logic and to have fewer different configurations floating around, this patch enables the hardened runtime in ad-hoc-signed macOS builds too.

This means any potential issues that arise in production builds due to the hardened runtime should now be reproducible in development/PR builds too.

### DIFF
--- a/cmake/modules/BundleInstall.cmake.in
+++ b/cmake/modules/BundleInstall.cmake.in
@@ -19,25 +19,14 @@ fixup_bundle("${BUNDLE_PATH}" "${BUNDLE_LIBS}" "${BUNDLE_DIRS}")
 
 if(DEFINED APPLE_CODESIGN_IDENTITY AND DEFINED APPLE_CODESIGN_ENTITLEMENTS)
   foreach(PATH_TO_SIGN IN LISTS BUNDLE_LIBS BUNDLE_PATH)
-    if(APPLE_CODESIGN_IDENTITY STREQUAL "-")
-      message(STATUS "Ad-hoc signing bundle without hardened runtime")
-      execute_process(COMMAND
-          codesign --verbose=4 --deep --force
-          --entitlements "${APPLE_CODESIGN_ENTITLEMENTS}"
-          --sign "${APPLE_CODESIGN_IDENTITY}"
-          "${PATH_TO_SIGN}"
-          RESULT_VARIABLE CODESIGN_EXIT_CODE
-      )
-    else()
-      message(STATUS "Signing bundle with hardened runtime and identity ${APPLE_CODESIGN_IDENTITY}")
-      execute_process(COMMAND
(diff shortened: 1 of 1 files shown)


fwcd:
Marked as a draft for now, since it seems as if the problems in https://github.com/mixxxdj/mixxx/pull/4774#issuecomment-1142158953 are still present.

The app crashes at launch with

```
Termination Reason:    Namespace DYLD, Code 1 Library missing
Library not loaded: @executable_path/../Frameworks/liblilv-0.0.dylib
Referenced from: <31B8A64E-9488-3DF2-8433-66A908CE7B95> /Volumes/*/mixxx.app/Contents/MacOS/mixxx
Reason: tried: '/Volumes/mixxx-2.4-beta-224-g51be918a9e/mixxx.app/Contents/Frameworks/liblilv-0.0.dylib' (code signature in <6D8DACD5-3EB7-3969-90CE-85910B228873> '/Volumes/mixxx-2.4-beta-224-g51be918a9e/mixxx.app/Contents/Frameworks/liblilv-0.0.dylib' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs)
(terminated at launch; ignore backtrace)
```

daschuer:
The crash seems to be related to dynamic linking. Up to now we use static linking, so is this related to this PR? 
For QML, we probably need dynamic lining though a future topic.

fwcd:
Yes, we link most libraries statically but not all of them:

image

The problem seems to be with the way how Library Validation (one of the features of the hardened runtime) interacts with ad-hoc signatures. Normally this would check that all libraries are signed with the same team id as the executable, with ad-hoc signatures, however, no team id is set:

```
$ codesign -dvv mixxx
Executable=/Volumes/mixxx-2.4-beta-224-g51be918a9e/mixxx.app/Contents/MacOS/mixxx
Identifier=org.mixxx.mixxx
Format=app bundle with Mach-O thin (arm64)
CodeDirectory v=20500 size=536400 flags=0x10002(adhoc,runtime) hashes=16752+7 location=embedded
Signature=adhoc
Info.plist entries=14
TeamIdentifier=not set   <----
Runtime Version=14.0.0
Sealed Resources version=2 rules=13 files=2759
Internal requirements count=0 size=12
```

So I am not entirely sure how this is intended to work, workarounds in the forum suggest disabling library validation entirely, but I am not sure if it's worth creating a separate `.entitlements` file just for supporting the hardened runtime + ad-hoc signing use case, since that would eventually just risk drifting out-of-sync again...

PR #10892, opened 2022-09-12
Reloop Mixage MIDI mappings

HorstBaerbel:
Redoing pull request. MIDI mappings for the Reloop Mixage DJ controller. Corresponding manual pull request is [#516](https://github.com/mixxxdj/manual/pull/516).

PR #592, opened 2015-05-24
[PREVIEW - DO NOT MERGE] TrackInfoObject: Locking + Caching + Tag Writing

uklotzde:
The interaction between TrackInfoCache and TrackDAO is tricky when track references expire. Currently it does not work as promised ;)

First I started to fixe some locking in other issues in TrackInfoObject:
- Avoid double locking in setDirty()
- Reduce locking in set-/getMetadata()
- Fix invalid locking when setting BPM/Beat Grid
- Some minor improvements and fixes

The next step to add a central weak-reference cache for TrackInfoObjects and re-enabling the tag writing options in Mixxx was inevitable ;) I also added a new context menu entry that allows users to write their existing information of selected tracks from the Mixxx library back into the corresponding files. This is necessary after tag writing has been disabled in Mixxx some time ago.

I use this version on a daily basis to manage my music library. I modify tags of tracks while they are playing without any issues so far. I have even enabled the new option that keeps the Mixxx library up-to-date with the tags of files, because I need external applications for mass-tagging that modify files already contained in the Mixxx library. Unfortunately rescanning the library will not detect modified files. With this option enabled the Mixxx library will be synchronized with the modified tags whenever such a file is selected, loaded, or edited in Mixxx.

[2015-08-23] I've fixed all comments so far on this branch and will replace it with an aggregated commit that reflects the remaining delta.

This PR will be broken down into smaller PRs:
- https://github.com/mixxxdj/mixxx/pull/684
- ...

PR #12296, opened 2023-11-12
Reloop Mixage MK1, MK2, Controller Edition mappings for Mixxx 2.4

HorstBaerbel:
Much improved and cleaned up version of PR #10892 by @gqzomer (Gersom Zomer) using Mixxx 2.4 functions like effect presets. Documentation in a new manual PR here [#592](https://github.com/mixxxdj/manual/pull/592).
Please consider this for inclusion in Mixxx 2.4. If you have improvements or need adjustments we're happy to improve the PR.


res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,784 @@
...
+        }
+
+        for (let effectUnit = 1; effectUnit <= Mixage.numEffectUnits; effectUnit++) {
+            // make connections for the fx on LEDs
+            const fxGroup = "group_"+channel+"_enable";
+            Mixage.fxOnConnection.push(engine.makeConnection("[EffectRack1_EffectUnit"+effectUnit+"]", fxGroup, function() { Mixage.toggleFxLED(channel); }));
+
+            // set soft takeovers for effectunit meta
+            engine.softTakeover("[EffectRack1_EffectUnit"+effectUnit+"]", "super1", true);
+            engine.setValue("[EffectRack1_EffectUnit"+effectUnit+"]", "show_focus", 1);
```
Swiftb0y, line 179: some renaming, deduplication, and 2.4 supports template literals I think.
```suggestion
            const fxControl = `group_${channel}_enable`;
            const fxGroup = "[EffectRack1_EffectUnit${effectUnit}]"
            Mixage.fxOnConnection.push(engine.makeConnection(fxGroup, fxControl, function() { Mixage.toggleFxLED(channel); }));

            // set soft takeovers for effectunit meta
            engine.softTakeover(fxGroup, "super1", true);
            engine.setValue(fxGroup, "show_focus", 1);
```

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,784 @@
...
+        }
+
+        // set soft takeover for filter effect
+        engine.softTakeover("[QuickEffectRack1_"+channel+"]", "super1", true);
+
+        // make connections for status LEDs
+        Mixage.vuMeterConnection.push(engine.makeConnection(channel, "vu_meter", function(val) { midi.sendShortMsg(0x90, Mixage.ledMap[channel].vu_meter, val * 7); }));
+        Mixage.loopConnection.push(engine.makeConnection(channel, "track_loaded", function() { Mixage.toggleReloopLED(channel); }));
```
Swiftb0y, line 187: what led is this controlling?

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,784 @@
...
+    }
+
+    if (enabledFxGroups.indexOf(1) !== -1) {
+        Mixage.toggleLED(ON, group, "fx_on");
+    } else {
+        Mixage.toggleLED(OFF, group, "fx_on");
+    }
+};
+
+// Flash the Reloop LED if a loop is set but currently not active
```
Swiftb0y, line 234: I think this needs discussion. LEDs should only really blink if they need the users attention, a loop that is set but not active is not uncommon and does not need urgent attention.

Swiftb0y (commented):
a couple comments, though I didn't get all the way through yet. I'm sorry for having neglected this mapping until now.

res/controllers/Reloop-Mixage.scripts.js
gqzomer: It's controlling the Reloop LED. when a new track is loaded that has a loop point set it starts the slowly blinking the reloop led (1s on 1s off). When a loop has not been set and thus reloop is not available it disables the blinking

res/controllers/Reloop-Mixage.scripts.js
gqzomer: Currently it blinks slowly (1 second on/off) to indicate that reloop is available for the deck. In my opinion the slow blinking rate doesn't draw immediate attention and is a helpfull indication that reloop is available. However this is open for debate. I chose this implementation so I could use the `loop/loop in` and `reloop/loop out` to communicate to a user if a loop is active (both leds on) or if only a loop in or loop out point has been set (although a loop out point cannot be set without a loop in point present)

res/controllers/Reloop-Mixage.scripts.js
gqzomer: Coming from a python background I was wondering if template literals where available in JS as it would greatly improve code readability. I wrote this code initially in es5 (mixxx 2.3) which as far as I could find didn't support this feature. With the move to es6 with 2.4 this should indeed now be possible. I will change the string concatination to template literals where applicable.

res/controllers/Reloop-Mixage.scripts.js
Swiftb0y: While your implementation is certainly valid and makes sense, I'd personally prefer if the behavior was simpler and would follow the mixxx UI more closely for the sake of consistency (the loop button should bind to `beatloop_activate` for input and `loop_active` for output and the reloop button bound to `reloop_toggle`). wdyt?

Swiftb0y: Indicating whether reloop is actually available makes a ton of sense, but it might be better implemented in C++ so all mappings and the GUI can benefit from it.

res/controllers/Reloop-Mixage.scripts.js
Swiftb0y: Ah yeah, should've read the mapping more carefully.

res/controllers/Reloop-Mixage.scripts.js
gqzomer: IMO binding the outputs as you suggested doesn't make optimal use of the LEDs on this particular controller. Effectively it means that the reloop LED is only lit when the reloop button is pressed communicating nothing about the state of the deck to the user. All LEDs on this controller (beside the -/+) button do communicate something about the state of the controller to the user. I even looked into using the -/+ LEDs to indicate the jog direction of a deck but I could not find a way to implement this in a clean manner. Buttons that only light up as long as they are being pressed is IMO a missed opportunity to communicate deck state. 

I do agree with you point that it might be better to implement this natively in C++ but I can always adjust this mapping accordingly when that happens.

res/controllers/Reloop-Mixage.scripts.js
gqzomer: I replaced all concatenations with string literals which should improve code readability

res/controllers/Reloop-Mixage.scripts.js
Swiftb0y: > I do agree with you point that it might be better to implement this natively in C++ but I can always adjust this mapping accordingly when that happens.

Would you be interested in implementing the C++ functionality?

Holzhaus:
Thanks. Please have a look at the failing check (Detecting code style issues). You can either use pre-commit locally or download the patch file from github actions and apply it.

HorstBaerbel:
I'm not sure if it's possible, but could the pipeline be changed to not build the whole of Mixxx when a file e.g. in the controllers folder changes? Or are the JS files read / needed in the build process?
In Azure pipelines it is possible to restrict triggers to certain folders and exclude others. Not sure if that options there on GitHub.

I have tried setting up eslint according to [this](https://github.com/mixxxdj/mixxx/wiki/Contributing%20mappings#coding-conventions-for-JavaScript), but I'm not sure the correct config is used? No problems show up in VS Code in the terminal.

res/controllers/Reloop-Mixage.scripts.js
gqzomer: >Would you be interested in implementing that feature in C++?

Maybe, although I have no prior experience with C++ or any compiling programming language for that matter. I might look into it after this mapping has been merged. I'm curious if there is any documentation available to get started with C++ in Mixxx

gqzomer:
>I have tried setting up eslint according to [this](https://github.com/mixxxdj/mixxx/wiki/Contributing%20mappings#coding-conventions-for-JavaScript), but I'm not sure the correct config is used? No problems show up in VS Code in the terminal.

you can manually load the eslint config for Mixxx by running in the root of the git repo
```sh
npx eslint -c .eslintrc.json
```

I sometimes still have to install eslint and the used plugins locally with npm before I can use the eslint extension in vscode. There is a node_module folder with the needed packages already present in the repo but it is for some reason not picked up correctly by npm.

res/controllers/Reloop-Mixage.scripts.js
gqzomer: >Indicating whether reloop is actually available makes a ton of sense

So do you agree with the current reloop blinking implementation in the mapping? Or should be adjusted or removed before the pull request can be accepted? I think I made my case as to why I chose this implementation. In any case it is not as egregious as the `load` button flashing on beat which the controller [originally ](https://youtu.be/iXk5s0fQPIQ?si=0PHyUuqRO47YBw6-&t=105)shipped with

res/controllers/Reloop-Mixage.scripts.js
HorstBaerbel: I'm resolving this. Feel free to comment if more thins pop up

res/controllers/Reloop-Mixage.scripts.js
HorstBaerbel: I'm proficient in C++ and I've taken a look at the source, but I have honestly no clue where to even start. Also I don't know what @Swiftb0y you're suggesting. Change Mixxx behaviour or create new JS variable binding?

I'd suggest we resolve this by either accepting the functionality as-is or changing the script to use standard `loop_active`.

HorstBaerbel:
@gqzomer I've played around with the mapping a lot in the last two weeks and it is working very well for me! Still like the layout and the functions make sense (to me at least). Two last thing we should look at the effects section:
- I've found the soft-takeover of the effects super knob to get stuck sometimes when I did not even change effects. It is not yet reproduceable for me.
- Also I've had the the effect enable button not toggle correctly. It only toggled a single effect enable even after changing effect presets or manually toggling that single enable button to the same state as the others. Hope you know what I mean.

And I would really like to close this PR, just for my peace of mind. This is a good mapping and it is definitely better than none at all. I would hate seeing this PR fade into oblivion again like the 2.3 one.

gqzomer:
>I've played around with the mapping a lot in the last two weeks and it is working very well for me! Still like the layout and the functions make sense (to me at least).

glad to hear you like the mapping.

>I've found the soft-takeover of the effects super knob to get stuck sometimes when I did not even change effects. It is not yet reproduceable for me.

Strange the soft takeover for the effect unit super control is currently being set once the user has cycled through the effect slots using the `fx_sel` button, or any time the `shift` button is released. I'm able to reproduce your issue by entering and leaving shift mode without changing the position of the `amount` knob.

I think the issue lies in the current soft take over implementation in Mixxx. I believe it currently works like this: Setting a soft take over ignores any input for a control until the control reads exactly the same value as when the soft take over was set. I think currently there is no tolerance allowed for a soft take over @Swiftb0y or @ronso0 might be able to confirm this suspicion. What I think is happening here is when shift is pressed while the position of a control is unchanged the first value when the control is changed reads higher or lower then when the soft takeover was set, thus the input is ignored.

edit: my suspicisons appear to be confirmed by the [midi-scripting page ](https://github.com/mixxxdj/mixxx/wiki/Midi-Scripting)

>If it is called unconditionally when switching to another layer and the user doesn't actually move the knob/fader before the next layer change, Mixxx will mistakenly initiate soft takeover if the user moves the knob/fader fast enough.

I believe this is not something I can fix in the mapping unless a tolerance setting is added to the soft-takeover implementation but I'm not sure if this has been suggested yet. I would suggest adding a tolerance parameter to the soft takeover like this
```ts
engine.softTakeover(group:string, key:string, enable:boolean, tolerance:number);
```
This would give contributors more flexibility  when developing mappings for Mixxx

>Also I've had the the effect enable button not toggle correctly. It only toggled a single effect enable even after changing effect presets or manually toggling that single enable button to the same state as the others. Hope you know what I mean.

Changing effect presets does not change anything about the effect toggle behavior. the state of the the effectslots is saved when all effect slots are disabled by pressing the `DRY/WET` button when no effectslot is selected. It's unclear to me what your issue is can you provide me with the steps to reproduce the problem?

Swiftb0y:
>  I believe it currently works like this: Setting a soft take over ignores any input for a control until the control reads exactly the same value as when the soft take over was set. I think currently there is no tolerance allowed for a soft take over @Swiftb0y or @ronso0 might be able to confirm this suspicion. 

That is not quite correct, there is a small epsilon (3/128th I think) at which point the values are considered close enough to take over. The direction of travel is also taken into account IIRC. 
If you enable soft-takeover and then move the knob fast enough so the first message after the soft-takeover is outside that range (and likely traveling away from it) mixxx will obviously think that the knob has been moved while in another layer. 

I don't know how a configurable threshold would solve this to be honest. A larger threshold would make it less likely that a sudden movement after softtakeover is out of that range, but that the same time, it would make softtakeover much less "soft".

gqzomer:
> If you enable soft-takeover and then move the knob fast enough so the first message after the soft-takeover is outside that range (and likely traveling away from it) mixxx will obviously think that the knob has been moved while in another layer.

Thanks for clearing that up @Swiftb0y I reworked the mapping to take this into account. The mapping now compares the position of the `filter/amount` knob with the corresponding engine value in Mixxx before setting enabling `softTakeoverIgnoreNextValue`. @HorstBaerbel Can you validate that the stuck issue is now fixed?

Also @Swiftb0y I was wondering if you had time to review the rest of the mapping, I can understand if you are busy but so far you've come with some great suggestions to improve the mapping.

Swiftb0y:
I'm sorry I currently don't have the time. but please ping me during mid of march, I'll have some time then.

res/controllers/Reloop-Mixage.scripts.js
ronso0: `loop_active` should light up the loop button(s) IMO.
With the current implementation it would blink as soon as a track with a temporary loop is loaded?
I agree that turning the LED on only when the button is pressed is not helpful.

Personally I use blinking when the loop is active, it's an important temporary state that deserves attention IMO.
So it could be
* on: reloop available
* blink: loop is active

gqzomer: In this mapping the LEDS for the `LOOP` / `LOOP IN` and `RELOOP` / `LOOP OUT`  buttons are controlled by a combination of the engine parameters `loop_enabled`,  `loop_start_position` and `loop_end_position` using the following logic

- If `loop_enabled` is true both the `LOOP` / `LOOP IN` and `RELOOP` / `LOOP OUT` light up
- If `loop_enabled` is false and `loop_start_position` and `loop_end_position` are set  the `RELOOP` / `LOOP OUT` button will flash slowly (1 second interval) indicating that reloop is available 
- If `loop_enabled` is false and `loop_start_position` and `loop_end_position` are not set  the ``LOOP` / `LOOP IN` and `RELOOP` / `LOOP OUT` will be turned off
- If `loop_enabled` is false and only the `loop_start_position` is set (occurs when manually setting a loop in point) the `LOOP` / `LOOP IN` will light up.
- If `loop_enabled` is false and only the `loop_end_position` is set (which is currently not possible in Mixxx) the `RELOOP` / `LOOP OUT` will light up

I'm not using `loop_active` as this will interfere with the LED feedback I have implemented for the loop adjustment mode. In this mode the user can adjust the loop in and out points with the jog wheel. The respective  `LOOP IN` or `LOOP OUT` button will flash rapidly (250ms interval) to indicate if the start, end or both are currently being adjusted with the jog wheel. I was inspired to add this feature after seeing this [youtube video](https://youtu.be/goyauO5rkv4?si=IItmDmU7hkeduu-z&t=202). 

I think the LED behaviour of the mapping is fine and follows the guidelines for Mixxx. I'm all for expanding the `reloop_toggle` output to indicate if a reloop is available. But until that feature is added I think it is fine to have it build into this mapping.

>Personally I use blinking when the loop is active, it's an important temporary state that deserves attention IMO. So it could be
>- on: reloop available
>- blink: loop is active

While I do see your point about it being a temporary state it is the exact opposite of the current Mixxx implementation. Where an enabled state is usually represented by a steady burning LED (the flashing Cue button on pause is a good example of this behavior). However I think either implementation could work to indicate reloop availability

ronso0: > If loop_enabled is false and loop_start_position and loop_end_position are set the RELOOP / LOOP OUT button will flash slowly (1 second interval) indicating that reloop is available

...which is true as soon as you load a track that had a loop set earlier (might even have been a short temporary loop_roll), which is most likely irrelevant in a new session.
Therefore I agree with @Swiftb0y (and the [mapping guideline](https://github.com/mixxxdj/mixxx/wiki/Contributing%20mappings#blinking-leds) btw) to **not** make it blink. This state does not deserve this much attention. You can just make it optional if you want that behavior for your own use case, but we shouldn't force it on anyone else.

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,812 @@
...
+Mixage.eject = function(group) {
+    if (engine.getValue(group, "play") === 0) {
+        if (Mixage.adjustLoop[group]) {
+            Mixage.stopLoopAdjust();
+        } else {
+            Mixage.blinkLED(Mixage.ledMap[group].reloop, group, 0);
+            Mixage.toggleLED(OFF, group, "loop");
+        }
+    }
+};
```
ronso0, line 283: Correct me if I'm wrong: the mapping does not provide eject, so this is called when `eject` is toggled in the GUI / via keyboard. Also it does not filter the value so it acts on press **and** release.

I suggest to merge this with the callback `Mixage.loopConnection.push(..)` (`track_loaded`) above.

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,812 @@
...
+
+// Removes any loop that is currently set on a track
+Mixage.clearLoop = function(_channel, _control, _value, _status, group) {
+    engine.setValue(group, "loop_end_position", -1);
+    engine.setValue(group, "loop_start_position", -1);
+};
```
ronso0, line 289: This is not necessary, `script.triggerControl(group, loop_remove, 100);` is easier, which can be used in `Mixage.handleLoopLengthPress` directly (so this function becomes obsolete).
 [[ChannelN],loop_remove](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-loop_remove)

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,812 @@
...
+            engine.scratchEnable(unitNr, Mixage.scratchTicksPerRevolution, 33.33, alpha, beta);
+        } else {
+            engine.scratchDisable(unitNr);
+        }
+    }
+};
+
+// The wheel that controls the scratching / jogging
+Mixage.wheelTurn = function(_channel, _control, value, _status, group) {
+    const deckNr = script.deckFromGroup(group);
```
ronso0, line 734: can be moved to the else branch where it is needed

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,812 @@
...
+    engine.setValue(group, "beatjump", beatjumpSize);
+};
+
+// clears a loop on a short press, set internal variable to true to adjust "beatjump_size"
+Mixage.handleLoopLengthPress = function(_channel, _control, value, _status, group) {
+    if (value === DOWN) {
+        Mixage.loopLengthPressed[group] = true;
+        Mixage.loopLengthPressTimer = engine.beginTimer(400, function() {
+            Mixage.loopLengthPressTimer = 0;
+        }, true);
```
ronso0, line 787: have a timer per group, like the longLengthPressed bool?

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,812 @@
...
+    } else if (value === DOWN) { // scratch toggle is not pressed
+        script.toggleControl(group, "play");
+    }
+};
+
+// Checks whether the Traxx button is double pressed
+Mixage.handleTraxPress = function(channel, control, value, status, group) {
+    if (value === DOWN) {
+        if (Mixage.traxxPressTimer === 0) { // first press
+            Mixage.traxxPressTimer = engine.beginTimer(400, function() {
```
ronso0, line 530: I suggest to increase the timeout a bit, just to improve accessibility for all usesr that are not able to press buttons quickly (like they can adjust the mouse double-click timeout on system level). 1000 is safe IMO

ronso0 (changes requested):
some findings and questions

res/controllers/Reloop-Mixage.scripts.js
gqzomer: Thanks for clearing that up @ronso0 I removed the reloop blinking behaviour from the mapping

res/controllers/Reloop-Mixage.scripts.js
gqzomer: This was only called when `eject` was toggled in the gui, I merged the `Mixage.stopLoopAdjust();`  callback with the `track_loaded` engine connection

res/controllers/Reloop-Mixage.scripts.js
gqzomer: didn't know that function was available, I changed it accordingly

res/controllers/Reloop-Mixage.scripts.js
gqzomer: I increased the timeout for the double press on the traxx button to 800 ms. I also added the `Mixage.shortPressTimeout` and `Mixage.doublePressTimeout` so these timeouts can be adjusted by users more easily

res/controllers/Reloop-Mixage.scripts.js
gqzomer: I didn't think this was necessary as under normal use it is unlikely a user would clear the loop on both decks at the same time for instance. I added a timer per group now just to be safe

res/controllers/Reloop-Mixage.scripts.js
gqzomer: good call I moved it to the else if branch

gqzomer:
@HorstBaerbel can you resolve the comments that are still open? I addressed them all with my last commit but cannot resolve reviews and comments since you submitted this pull request.

ronso0:
FYI, except for trivial fixes  it's up to the reviewer to resolve review conversations.

I'm not sure when I'll have time to take an in-depth look at this.
Please ping me if you have the feeling I forgot about it ; )

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,780 @@
...
+            Mixage.toggleLED(Mixage.scratchToggleState[group], group, "scratch_active");
+            if (Mixage.scrollToggleState[group]) {
+                Mixage.scrollToggleState[group] = !Mixage.scrollToggleState[group];
+                Mixage.toggleLED(Mixage.scrollToggleState[group], group, "scroll_active");
+            }
+        }
+    }
+};
+
+// The "loupe" button that enables/disables track scrolling
```
ronso0, line 666: :wolf: or :loop: ; )

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,780 @@
...
+    }
+
+    // find controls and make engine connections for each channel in Mixage.channels
+    // A predefined list with channels is used instead of a for loop to prevent engine connections to be overwritten
+    Mixage.channels.forEach(function(channel) {
+        const deck = script.deckFromGroup(channel);
+        Mixage.connectControlsToFunctions(channel);
+        const effectUnit = `[EffectRack1_EffectUnit${deck}]`;
+
+        // set soft takeovers for effectslot amount
```
ronso0, line 182:
```suggestion
        // set soft takeovers for effect slot Meta knobs
```

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,780 @@
...
+            engine.softTakeover(fxGroup, "super1", true);
+            engine.setValue(fxGroup, "show_focus", 1);
+        }
+
+        // set soft takeover for filter effect
+        engine.softTakeover(`[QuickEffectRack1_${channel}]`, "super1", true);
+
+        // make connections for status LEDs
+        Mixage.vuMeterConnection.push(engine.makeConnection(channel, "vu_meter", function(val) { midi.sendShortMsg(0x90, Mixage.ledMap[channel].vu_meter, val * 7); }));
+        Mixage.trackLoadedConnection.push(engine.makeConnection(channel, "track_loaded", function() { if (Mixage.adjustLoop[channel]) { Mixage.stopLoopAdjust(channel); } }));
```
ronso0, line 204: IIUC `makeUnbufferedConnection` is preferred for VU connections
example: https://github.com/mixxxdj/mixxx/blob/77b466e745dc9cf8bdf57286426cdca4b12f485f/res/controllers/Traktor-Kontrol-Z1-scripts.js#L113
Doc: https://github.com/mixxxdj/mixxx/blob/77b466e745dc9cf8bdf57286426cdca4b12f485f/res/controllers/engine-api.d.ts#L118-L129

ronso0 (dismissed):

ronso0:
@HorstBaerbel Did you alread sign the [Mixxx Contributor Agreement](https://docs.google.com/forms/d/e/1FAIpQLScC9QG327sjLL0eWftmfGUasxFFLxg6LCXJ2xHDYRzFIRqyiw/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ)?
If yes, which alias did you pick?

ronso0:
I just made another attempt to review this. I only did a brief check, still no js expert.
Would appreciate some help by @JoergAtGithub or @Swiftb0y

HorstBaerbel:
> @HorstBaerbel Did you alread sign the [Mixxx Contributor Agreement](https://docs.google.com/forms/d/e/1FAIpQLScC9QG327sjLL0eWftmfGUasxFFLxg6LCXJ2xHDYRzFIRqyiw/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ)? If yes, which alias did you pick?

Signed as HorstBaerbel :+1:

res/controllers/Reloop-Mixage.scripts.js
HorstBaerbel: https://en.wikipedia.org/wiki/Loupe
"Magnifying glass" is a tad bit on the long side. Other suggestions?

ronso0: aaah, okay, thought it's a typo and you mean loop.
In that case loupe hopefully works

res/controllers/Reloop-Mixage.scripts.js
HorstBaerbel: Can not test this currently. On holiday without the controller. Can check this in a couple of days

gqzomer:
@ronso0 Is it enough for the owner of the pull request to sign the contributor agreement? Or should I also sign the agreement as a contributor to this mapping?

ronso0:
This PR contains many commits of yours so it would be very welcome if sign it as well, thank you!

gqzomer:
>This PR contains many commits of yours so it would be very welcome if sign it as well, thank you!

Signed as gqzomer

res/controllers/Reloop-Mixage.scripts.js
HorstBaerbel: Using an unbuffered connection now

gqzomer:
The requested changes have been applied a while back. @ronso0 can you maybe resolve your request for changes so this pull request correctly reflects it is ready to be merged?

ronso0:
Thanks @gqzomer for the reminder.
I dismissed my review (won't approve though since I'm still no js expert)

I'll leave some more notes.

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,780 @@
...
+
+    // all button LEDs off
+    for (let i = 0; i < 255; i++) {
+        midi.sendShortMsg(0x90, i, 0);
+    }
+
+    // find controls and make engine connections for each channel in Mixage.channels
+    // A predefined list with channels is used instead of a for loop to prevent engine connections to be overwritten
+    Mixage.channels.forEach(function(channel) {
+        const deck = script.deckFromGroup(channel);
```
ronso0, line 178: Just a note:
you may replace the string operation with a simple lookup table where applicable
```js
const fruitColors = {
  [Channel1]: 1,
  [Channel2]: 2,
};
```
(don't remember if but channels may be 0-based)

res/controllers/Reloop-Mixage.scripts.js
```diff
@@ -0,0 +1,780 @@
...
+            const fxControl = `group_${channel}_enable`;
+            const fxGroup = `[EffectRack1_EffectUnit${effectUnit}]`;
+            Mixage.fxOnConnection.push(engine.makeConnection(fxGroup, fxControl, function() { Mixage.toggleFxLED(channel); }));
+
+            // set soft takeovers for effectunit meta
+            engine.softTakeover(fxGroup, "super1", true);
+            engine.setValue(fxGroup, "show_focus", 1);
+        }
+
+        // set soft takeover for filter effect
```
ronso0, line 199:
```suggestion
        // set soft takeover for Quick Effect
```

res/controllers/Reloop-Mixage.scripts.js
Swiftb0y: script.deckFromGroup is fine IMO (likely faster than a lookup tbh).
Also your suggestion probably doesn't work ootb, it would need to be
```js
const fruitColors = {
    "[Channel1]": 1,
    "[Channel2]": 2,
}
```
because what you posted are actually [`computed property names`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names)

Swiftb0y (approved):
I'm not a fan of this style of review anyways and since I don't have the controller I also can't really make any suggestion from a mapping user perspective. If this mapping leveraged [ComponentsJS](https://github.com/mixxxdj/mixxx/wiki/components%20JS) I could make better suggestions but I understand if thats too much to ask. I can make a bunch of code-style nitpicks but that wouldn't make it more readable either so I guess LGTM?

daschuer:
Merge?

gqzomer:
@Swiftb0y I have looked into ComponentsJS in the past for this controller. But as I understood it Component style mappings are mostly beneficial for controllers that use a different channel for each deck. Sadly this controller uses only one channel and opts  for different midi inputs for the left and right deck. Either way the starting point for this pull request was a non-ComponentsJS mapping submitted in an earlier pull request by @HorstBaerbel

Swiftb0y:
> But as I understood it Component style mappings are mostly beneficial for controllers that use a different channel for each deck.

No that doesn't particularly matter. ComponentJS is useful for encapsulating state which applies to almost every mapping (this one included as you can see by the many global state-tracking variables at the top). Anyways, doesn't particularly matter now.

res/controllers/Reloop-Mixage.scripts.js
HorstBaerbel: Resolving for rons0 (interpreting your thumbs-up)

daschuer:
@Swiftb0y merge?

Swiftb0y:
waiting for @ronso0 to give his LGTM on the manual PR.

ronso0:
Thanks to everyone involved!

PR #12300, opened 2023-11-14
(fix)Tag Fetcher: enable Apply button regardless the cover art

ronso0:
fixes a regression from #12041 or a later commit:
Tags couldn't be applied because the button remained diasbled.


daschuer (approved):
LGTM, Thank you.

PR #12301, opened 2023-11-14
Tagfetcher: cache fetched covers

### LINKED ISSUE #11084: Cover Art Fetcher: cache fetched images per track
### Feature Description

#4851
Selecting a TagFetcher result instantly fetches the cover image. Thus, selecting multiple results to find the best-matching cover (actual picture, size, quality) would needlessly re-download images that were already viewed.

Let's cache the images per track. We can either use a strict memory limit or just clear the cache once another track is loaded.

ronso0:
Closes #11084

A quick test to cache the fetched cover images with ~~the global `QPixmapCache`~~ a QMap.
Cache keys are ~~"DlgTagftecher" +~~ the MusicBrainz release id.
This greatly improves UX when comparing covers of MusicBrainz results.

~~Works good, though removing the covers from the cache is a bit dirty I think.~~
~~To discuss:~~
~~* really use global QPixmapCache which is also used by CoverArtCache?~~
~~* shall we increase the QPixmapCache cache limit when DlgTagFetcher is constructed to avoid clearing cached covers by adding large fetched covers?~~

I didn't check the sizes of fetched covers, yet, we need to estimate a reasonable limit.

### ToDo
- [x] clear cache when loading another track
- [ ] preferred image size vs. cached images:
currently the preference is read from config each time image urls are received. With the current implementation this may result in a mismatch of current size preference / cached images.
We could 
a) read from config **once** in the constructor to avoid the issue in the first place
b) read the preference each time urls are received and use the size preference (int) for the cache key
- [x] remove/disable debug output

based on #12300

### DIFF
--- a/src/library/dlgtagfetcher.h
+++ b/src/library/dlgtagfetcher.h
@@ -61,8 +61,10 @@ class DlgTagFetcher : public QDialog, public Ui::DlgTagFetcher {
             const QObject* pRequester,
             const CoverInfo& coverInfo,
             const QPixmap& pixmap);
-    void slotStartFetchCoverArt(const QList<QString>& allUrls);
-    void slotLoadBytesToLabel(const QByteArray& data);
+    void slotStartFetchCoverArt(const QUuid& albumReleaseId,
+            const QList<QString>& allUrls);
+    void slotLoadFetchedCoverArt(const QUuid& albumReleaseId,
+            const QByteArray& data);
     void slotCoverArtLinkNotFound();
     void slotWorkerStarted();
     void slotWorkerAskOverwrite(const QString& coverArtAbsolutePath,
@@ -74,10 +76,12 @@ class DlgTagFetcher : public QDialog, public Ui::DlgTagFetcher {
   private:
     // Called on population or changed via buttons Next&Prev.
     void loadTrackInternal(const TrackPointer& pTrack);
-    void getCoverArt(const QString& url);
+    void getCoverArt(const QUuid& albumReleaseId, const QString& url);
     void loadCurrentTrackCover();
 
     void saveCheckBoxState();
+
+    void loadPixmapToLabel(const QPixmap& pPixmap);
 
     UserSettingsPointer m_pConfig;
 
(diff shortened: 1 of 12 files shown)


ronso0:
Rebased, I adjusted & simplified related code a bit, e.g. loading a cover to DlgCoverArtFullsize.
It's working fine now.

However, (ab)using QPixmapCache is not the way to go: it's capped to 20 MB, and fetching some 1200px covers would flush all covers cached by the library table.

So a [QCache](https://doc.qt.io/qt-6/qcache.html) is unavoidable.

ronso0:
Now the covers are cached in a QMap. I didn't set a cache limit, yet. Let's figure out a reasonable size, i.e. for how many high-res covers the size would become a problem.

I couldn't get it to work with QCache, I always got segfaults when retrieving pixmaps. Must be something with QCache taking ownership of the objects, but I could figure it out. Not even when I stored the fetched QByteArray, it was always nulled for some reason. If someone wants to take a look, the experimental QCache commit is 2e231a2a9f

src/library/dlgtagfetcher.h
```diff
@@ -107,5 +111,7 @@ class DlgTagFetcher : public QDialog, public Ui::DlgTagFetcher {
 
     QByteArray m_fetchedCoverArtByteArrays;
 
+    QMap<QString, QPixmap> m_coverCache;
```
daschuer, line 114: Did you consider to extend use the exiting cover cache here? 
I have not looked into it, but I can imagine that this way we can save one cover access. 

An other option is to use QPixmapCache or the general purpose QCache container.

ronso0: In the comments above I explained why QPixmapCache is unsuitable, and that I didn't get it to work with QCache (even though QPixmapCache is derived from QCache). Any idea how to fix the QCache implementation is welcome, even though the only benefits are slightly faster lookup and automatic cleanup (limit).

daschuer:
It looks like this is mergable after you have rebased out the debug commit.

ronso0:
Oh great, will do.

ronso0:
Done.
For TODO #2 I filed #13099

daschuer (approved):
Looks and works good! Thank you.

PR #12305, opened 2023-11-15
Sampler DeckVisuals fix

### LINKED ISSUE #12304: sampler's "end_of_track" control is not created
Sampler's "end_of_track" control is not created if the sampler count in `samplers.xml` is lower than the samplers requested by a skin.
(and indeed the eot highlight is not drawn in the affected sample decks)
```
warning [Main] ControlDoublePrivate::getControl returning NULL for ( "[Sampler17]" , "end_of_track" )
```

I noticed these warnings in the log but couldn't reproduce immediatley, even though it's easy:
just delete samplers.xml (like with a fresh profile), or remove some lines, and you'll get the warnings.

While writing this, and taking a look at DeckVisuals / Visualsmanager (creator of the eot control), the cause is clear:
PlayerManager doesn't have a sampler equivalent for `emit numberOfDecksChanged(m_decks.count());`.
Fix should be easy it seems.

ronso0:
Closes #12304 
```
info [Main] Creating skin control object: "[Sampler16],visual_bpm"
warning [Main] ControlDoublePrivate::getControl returning NULL for ( "[Sampler9]" , "end_of_track" )
```

Symptons were
* no end-of-track highlight
* no BPM display
* no key display

for samplers requested by a skin that exceed the initial sampler count (4) or the number of samplers requested by `samplers.xml`

And some cleanup/tweaking en passant.

### DIFF
--- a/src/coreservices.cpp
+++ b/src/coreservices.cpp
@@ -59,6 +59,7 @@ namespace {
 const mixxx::Logger kLogger("CoreServices");
 constexpr int kMicrophoneCount = 4;
 constexpr int kAuxiliaryCount = 4;
+constexpr int kSamplerCount = 4;
 
 #define CLEAR_AND_CHECK_DELETED(x) clearHelper(x, #x);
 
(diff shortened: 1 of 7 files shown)


daschuer (approved):
LGTM, Thank you.

PR #12303, opened 2023-11-15
add missing include /util/time.h

ronso0:
noticed locally with pre-built headers OFF

I'm a bit worried that such things seem to slip through CI

PR #12308, opened 2023-11-17
Add missing includes

fwcd:
For some reason, the `main` build randomly started failing for me due to some missing includes.

Now I wonder why this even worked in the first place (`hss1394enumerator.h` doesn't even include `usersettings.h` transitively, so it probably relied on having downstream source files include that somewhere above `hss1394enumerator.h`?) Not sure if there was some update to Xcode that made the compiler stricter, but even that seems strange, since C++ compilers usually go to great lengths to ensure backwards compatibility.

Anyway, here's the patch.

### DIFF
--- a/src/controllers/midi/hss1394enumerator.h
+++ b/src/controllers/midi/hss1394enumerator.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "controllers/midi/midienumerator.h"
+#include "preferences/usersettings.h"
 
 // Handles discovery and enumeration of DJ controllers that appear under the
 // HSS1394 cross-platform API.
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -22,6 +22,7 @@
 #include "util/cmdlineargs.h"
 #include "util/console.h"
 #include "util/logging.h"
+#include "util/sandbox.h"
 #include "util/versionstore.h"
 
 namespace {
--- a/src/preferences/upgrade.cpp
+++ b/src/preferences/upgrade.cpp
@@ -14,6 +14,7 @@
 #include "library/trackcollection.h"
 #include "preferences/beatdetectionsettings.h"
 #include "preferences/usersettings.h"
+#include "util/cmdlineargs.h"
 #include "util/db/dbconnectionpooled.h"
 #include "util/db/dbconnectionpooler.h"
 #include "util/math.h"


daschuer (approved):
LGTM, Thank you.

ronso0:
Could it be pre-compiled headers masking missing includes until 'something' requires a header to be rebuilt? Same with #12303, the PR built fine but locall I noticed immediately with PCH disabled.

fwcd:
I've been building with PCH disabled for a while already. Also, shouldn't the CI builds fail in that case too?

daschuer:
This are remaining issues after my header clean up PR. 
I have tested it successful with PCH disabled, but only on Ubuntu Focal with one Qt 5 and one Qt 6 version.
Different QT and libstdcpp seems to include different files from their headers. Even before my PR we had sometimes missing headers in certain environments. 

On the CI we have PCH enabled where all headers are included everywhere, so it cannot detect these issues. 

I am sorry that this hits you when you want to do something else.

daschuer (approved):
LGTM, thank you.

fwcd:
Since CI performs clean builds anyway, would it make sense to disable PCHs there?

daschuer:
Unsure. The PCH saves some Build time = carbon footprint in case of clean builds. In case of builds at home on the other hand it forced a clean build once you touch a header. 

I am considering the following: Default to PCH builds only in Release builds, and do Debug builds without. 

An alternative is to write a script that collects all headers in angle brackets put them in a central header file at root and use only this for precompiling. 

These headers are from Qt and system library. When Qt or other libs are updated we do almost a clean build anyway. 

At least it will prevent a full rebuild after switching a branch. 

The header cleanup was a prerequisite to have a base line for all build time measures.

fwcd:
Maybe there's a way to lint for missing/superfluous header dependencies? I'm a bit worried that we might regress on this in the future again if PCHs are used everywhere...

daschuer:
There are ways of cause l. But they require a lot of build time  and need to consider each possible configuration. 
Compared to that, the current situation is not too bad IMHO. It has just happened in a high frequency lately because of my header clean up PR.

fwcd:
Hm, the shorter build time is of course nice, but is it really crucial that all CI jobs complete quickly? This seems to be a situation of maybe 20 vs. 30 minutes (so within the [same soft order of magnitude](https://belkadan.com/blog/2023/10/Soft-Orders-of-Magnitude/)), additionally this build time is highly variable anyway, due to differences in load across GitHub's runners, environments etc.

One idea was that we could add a third Ubuntu configuration (similar to how we build with different Qt versions) that runs in parallel and builds with PCH disabled. That would first and foremost ensure that all we don't introduce correctness issues accidentally (especially if contributors build with PCHs locally too, which I wouldn't consider unlikely given that it is now the default configuration).

daschuer:
I have build the latest branch on Ubuntu Focal without PCH enabled, and these issues slipped though anyway. So it would be only one additional configuration of many.

What do you think about the idea to disable PCH for Debug builds by default? Than we have a good test coverage on many different systems ...

fwcd:
Sounds reasonable!

fwcd:
@daschuer Could we get a 2.4 -> `main` merge? This is currently blocking me from building `main`...

daschuer:
Done

PR #12320, opened 2023-11-20
Effects: center Super knob when loading empty (QuickEffect) chain preset

ronso0:
picked from #11695 

Center Super knob in order to eliminate the colored ring when not at 12 o'clock. It's simply less visual noise this way.


JoergAtGithub:
Works for me! Clazy fail seems to be unrelated. Thank you!

PR #12321, opened 2023-11-20
Tracks table: avoid wrong pre-selection with no tracks selected previously

### LINKED ISSUE #12064: a single table cell is drawn as selected when selecting a previous search query
Noticed this a few times while testing #12063 (no changes to tracks table or search box, just searchqueryparser).

Can not reproduce but this is what I did IIRC
* type a query, hit Return to save it
* go to tracks table, select tracks, change selection, edit tags etc.
* clear search
* select a previous query
* the top-left table cell is drawn like if it was selected (coloured background)
* switching to the tracks table selects the entire row with the previously 'selected' cell being focused (white border)

This shouldn't be possible since we `setSelectionBehavior(QAbstractItemView::SelectRows);` in WLibraryTableView.
My Qt version didn't change either (still on 5.12.8) :man_shrugging:

ronso0:
by initializing with invalid row/column -1

Closes #12064

### DIFF
--- a/src/widget/wlibrarytableview.cpp
+++ b/src/widget/wlibrarytableview.cpp
@@ -21,8 +21,8 @@ constexpr int kModelCacheSize = 1000;
 WLibraryTableView::WLibraryTableView(QWidget* parent,
         UserSettingsPointer pConfig)
         : QTableView(parent),
-          m_prevRow(0),
-          m_prevColumn(0),
+          m_prevRow(-1),
+          m_prevColumn(-1),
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
I could reproduce the second case described in https://github.com/mixxxdj/mixxx/issues/12064 with 2.4 - this gone with this PR. Clazy fail is unrelated. Thank you!

ronso0:
Thanks for testing!

PR #12322, opened 2023-11-20
Korg nanoKontrol2: don't try to configure more than 4 main decks

### LINKED ISSUE #12317: debug assert "num <= kMaxNumberOfDecks" in PlayerManager::slotChangeNumDecks on changing nanoKontrol2 mapping from custom to built-in
### Bug Description

coredump; https://nxtcld.thingsandstuff.org/s/CdiMwZE3AprJWjG

I had both plugged in when I started Mixxx.

I go into the nanoKontrol2 controller config page, change from my "kn" preset to the official one in the drop-down, hit apply, there is several seconds pause then a crash.

### Version

r8934-1

### OS

Arch

ronso0:
This is just a quick fix, I'm not going to take care of the countless eslint errors (this mapping was added 2013).

Closes #12317

### DIFF
--- a/res/controllers/Korg-nanoKONTROL-2-scripts.js
+++ b/res/controllers/Korg-nanoKONTROL-2-scripts.js
@@ -1,7 +1,7 @@
 function NK2() {}
 NK2.debug=0;//set debug level
 NK2.LEDflasheson=true;//false disables led flash timers for debugging purposes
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
Looks like the correct thing to do in this case. Thank you!

PR #12324, opened 2023-11-20
Snap packaging yaml

jssotomdz:
This pull request adds the snapcraft yaml necessary to provide snap packaging and distribution.

Right now I have it published with the name mixxx-community, and I've been testing it for some time now and it works as the deb. Let me know if you're interested in publishing an official snap, I could help maintain it. Otherwise just let me know and I'll close/you can close the PR.

In case you're interested, the first step would be to ask for your permission to use the mixxx name instead of mixxx-community in the snap store.

I wanted to add the snap directory in the packaging directory, but unfortunately snapcraft only supports having the snap/snapcraft.yaml in the root directory. In case you prefer it that way, the file can exist on a separate repo since it builds from the ppa and not from source.

Lmk if you have any further comments or questions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,6 @@ src/mixxx.rc.include
 src/mixxx.res
 src/test/**/*.actual
 res/qrc_mixxx.cc
+
+# Snap packages built locally
+*.snap
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,185 @@
+name: mixxx
+base: core22
+adopt-info: mixxx
+grade: stable
+confinement: strict
+
+environment:
+  HOME: $SNAP_USER_COMMON
+  QT_PLUGIN_PATH: $SNAP/usr/lib/$CRAFT_ARCH_TRIPLET/qt5/plugins:$SNAP/usr/lib/$CRAFT_ARCH_TRIPLET:$SNAP_DESKTOP_RUNTIME/usr/lib/$CRAFT_ARCH_TRIPLET/qt5/plugins:$SNAP_DESKTOP_RUNTIME/usr/lib/$CRAFT_ARCH_TRIPLET${QT_PLUGIN_PATH:+:$QT_PLUGIN_PATH}
+  QML2_IMPORT_PATH: $SNAP/usr/lib/$CRAFT_ARCH_TRIPLET/qt5/qml:$SNAP/lib/$CRAFT_ARCH_TRIPLET:$SNAP_DESKTOP_RUNTIME/usr/lib/$CRAFT_ARCH_TRIPLET/qt5/qml:$SNAP_DESKTOP_RUNTIME/lib/$CRAFT_ARCH_TRIPLET${QML2_IMPORT_PATH:+:$QML2_IMPORT_PATH}
+  ALSA_CONFIG_PATH: $SNAP/usr/share/alsa/alsa.conf
+  DISABLE_WAYLAND: 1
+
+compression: lzo
+architectures:
+  - build-on: amd64
+
+package-repositories:
+ - type: apt
+   ppa: mixxx/mixxx
+
+parts:
+  mixxx:
+    plugin: nil
+    parse-info: [usr/share/metainfo/org.mixxx.Mixxx.metainfo.xml]
+    override-build: |
+      craftctl default
+      # Set the snap version from the .deb package version
+      craftctl set version=$(dpkg-parsechangelog -l $CRAFT_PART_INSTALL/usr/share/doc/mixxx/changelog.Debian.gz -S Version | sed -e s/2://g | cut -d- -f1)
+      # Copy over the icon
+      mkdir -p $CRAFT_PART_INSTALL/meta/gui
+      cp $CRAFT_PART_INSTALL/usr/share/icons/hicolor/scalable/apps/mixxx.svg $CRAFT_PART_INSTALL/meta/gui/icon.svg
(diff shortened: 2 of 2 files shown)


uklotzde:
Thanks for taking responsibility. Maintaining the different distribution channels is an underestimated effort.

These distribution-dependent artifacts do not need to be maintained in the central Mixxx repo. It is pointless to add them without a CI job for verification.

Example (Fedora / RPM Fusion):

- https://github.com/rpmfusion/mixxx
- https://koji.rpmfusion.org/koji/packageinfo?packageID=148

I decided to keep maintaining this package, although I am no longer a member of the Mixxx core team.

The Mixxx team might approve you to use the official name for the package. Get in touch with Snapcraft and ask them what they need to acquire this name. Adding a link to the install instructions on the Mixxx website would be helpful then.

daschuer:
Thank you for contribution the "Mixxx (community)" snap. I support the idea to make it official. 

Your current solution in https://github.com/jssotomdz/mixxx is a bit cumbersome, because users cannot see at a glance which version of Mixxx is build or if you have applied custom patches. 

For my understanding an established home of "unofficial" snaps is https://github.com/snapcrafters
Looking at the Gimp snap you see that they have just a reference for the gimp source itself: 
https://github.com/snapcrafters/gimp/blob/f43e10142a45bbd46ddc5aad7da5694e56cf67f6/snap/snapcraft.yaml#L641  

Ubuntu itself maintains the "official" snaps here https://github.com/ubuntu/thunderbird
 
VLC for instance maintains a snapcraft.yaml within its source trees to produce nightly builds: 
https://github.com/videolan/vlc/blob/f7bb59d9f51cc10b25ff86d34a3eff744e60c46e/extras/package/snap/snapcraft.yaml#L65

So we may discuss if we want to produce nightly snaps. 

For now I don't think that would be a big improvement, because we already have a nightly PPA. We can decide it at any time later.

I think the best solution for now is to modify the snapcraft.yaml according to the snapcrafters template and join them. 
This way we become a kind of "official unoffcial" snap from the Ubuntu perspective. And we benefit from the review process among the snapcrafters, that adds another level of confidence to the resulting snap. 

The only concerning part is in their description: 
> Having the upstream application developer maintain and publish a snap for their application is preferred but not always possible. If the developers are not yet interested in maintaining the snap, we fill the gap.

What do you think? Do you have interest to contribute the work towards snapcrafters? I think we should get in contact with them.

jssotomdz:
Hi! a few comments on your thoughts:

> Your current solution in https://github.com/jssotomdz/mixxx is a bit cumbersome, because users cannot see at a glance which version of Mixxx is build or if you have applied custom patches.

Yes, snaps tend to be a mystery on what they contain. The best way to solve this with a direct link to the repo containing the instructions to build it. Something like what flatpak does when you click the manifest link. In this case the yaml would be public and as you can see, it is grabbing the PPA version built for jammy and containerizing it, nothing more.

> For now I don't think that would be a big improvement, because we already have a nightly PPA. We can decide it at any time later.

One of the nice things about the snap store is that you can provide builds in different channels, for example a stable build, a release candidate and a bleeding edge (nightly). And you can switch to any of those with a single command. But yes, tbh if you're already maintaining a nightly and stable PPA for now it wouldn't be a huge improvement.

> What do you think? Do you have interest to contribute the work towards snapcrafters? I think we should get in contact with them.

II think they aren't accepting any more projects since keeping everything they own up to date has been a challenge, especially because most of them are community contributors. But we can try, I have direct contact with them.

Regarding the description, this is because snapcrafters exists kind of to provide an incubator for snaps. The long term idea is for every app developer to own their snap and maintain it (because growing the amount of stuff snapcrafters own will eventually fill their hands and lead to outdated software). Also it is much better as a user to see that the app is distributed by the official team rather than a random community member. Snapcrafters is a bit better than a community member but the developers themselves is much better.

From what I've learned in 10 months maintaining the mixxx-community snap, is that given the work you do with PPAs, maintaining the snap is only a matter of retriggering a build after you release a new version to the PPA. The only case where yaml changes are needed is when the dependencies are added or removed. Otherwise, bumps to dependencies or new releases only need a rebuild.

On second thought I think it would be better to have a separate repo for cleanliness and to avoid retriggering unwanted builds.

But let me know what you think, maybe for now the best course of action would be for me to request the mixxx name in the snap-store with your permission so that it can use the right name, and when you feel it's the right time to make it official, include it as a distribution channel and even place the repo in the org.

daschuer:
Looking at https://snapcraft.io/vlc they have a "verified account" icon, the snapcrafters are "only" "Star Developers"
So I see how this PR has actually some benefits.  
In such a scenario we need to put the credentials into a GitHub secret. 

Any thoughts?

jssotomdz:
I'm sorry, I'm not following. Credentials for the snap store account itself? If so, yes, maybe so that all of the owners have access to the account.
Also, FYI on how to get a verified account.: https://forum.snapcraft.io/t/verified-accounts/34002

daschuer:
No owner will have direct access to the account, because a GitHub workflow secret is write only. 
It "just" allows to upload the snap directly from a workflow, which is traceable and removed the human as an error factor from the process. 

Implicit all owners have access though, via a custom workflow.

acolombier:
@mixxxdj/developers what's the status on this? It feels to me that this is related to the discussion on Zulip about building flatpack in our build pipeline? 
I agree with @uklotzde 
> It is pointless to add them without a CI job for verification.
If we want to make that part of the repo, shall we add some more pipeline setup? (Nightly build, release build, ...)

PR #12340, opened 2023-11-22
Fix missing #include directive

uklotzde:


daschuer:
"util/make_const_iterator.h" is a main file. 
I have in the meantime already pushed a fix to main. 
Thank you anyway.

uklotzde:
Didn't check when changing the target from main to 2.4.

Nevertheless I expect that these issues are caught by the CI builds.

PR #4489, opened 2021-10-26
[POC] Beats: Add editing controls and show downbeats on scrolling waveforms

Holzhaus:
Based on #4411.

**To Do:**
- [ ] Change Beatmap import to ensure that `beatsTillNextMarker` is always a multiple of `beatsPerBar` to fix downbeats support in beatmaps 
- [ ] ...

PR #12343, opened 2023-11-23
Beats: Add editing controls and show downbeats on scrolling waveforms [v2]

fwcd:
This is a rebase of @Holzhaus's #4489 onto `main`, a proof-of-concept implementation of downbeat markers and buttons for setting downbeats in the grid:

image

### To do

- [ ] Make this an off-by-default, marked-as-experimental feature in the settings, so people can play with it, without burdening regular users with its limitations (e.g. incorrect downbeat markers on non-4/4 tracks)
- [ ] Address open feedback from #4489

### Deferred

- Support non-4/4 (and changing time signatures): Definitely planned, but out-of-scope to keep this PR simple and to avoid changing the data model in a backwards-incompatible way for now


src/track/beats.cpp
```diff
@@ -21,10 +21,48 @@ struct BeatGridV1Data {
...
+        return numBeats - numExcessBeats;
+    }
+
+    return numBeats + beatsPerBar - numExcessBeats;
+}
+
+int calculateBeatsTillNextMarker(const mixxx::Beats::ConstIterator& it,
+        mixxx::audio::FramePos nextMarkerPosition) {
+    const double numBeatsDouble = std::fabs(nextMarkerPosition - *it) / it.beatLengthFrames();
+    const int numBeatsInt = static_cast<int>(std::round(numBeatsDouble));
```
daschuer, line 39: We round here twice. That may lead to a wrong overall rounding

src/track/beats.cpp
```diff
@@ -21,10 +21,48 @@ struct BeatGridV1Data {
...
+int roundBeatCountToFullBar(int numBeats, int beatsPerBar) {
+    const int numExcessBeats = (numBeats % beatsPerBar);
+    if (numExcessBeats < (beatsPerBar / 2)) {
+        return numBeats - numExcessBeats;
+    }
+
+    return numBeats + beatsPerBar - numExcessBeats;
+}
+
+int calculateBeatsTillNextMarker(const mixxx::Beats::ConstIterator& it,
```
daschuer, line 36: The function name is misleading. How about: 
`int beatsTillBarNearPosition()`

daschuer: It looks like there is a need for "calculateBeatsTillNextMarker()" this should however be independent form a Bar.

daschuer (changes requested):

src/track/beats.h
```diff
@@ -81,8 +96,30 @@ class Beats : private std::enable_shared_from_this<Beats> {
...
 
+        /// Returns the number of beats per bar.
+        ///
+        /// Note: Currently always returns 4, as time signatures other than 4/4
+        /// are not supported yet.
+        int beatsPerBar() const;
+
+        /// Returns the current index of the beat in the bar.
+        int beatIndexInBar() const {
+            return m_beatOffset % beatsPerBar();
```
daschuer, line 110: This assumes that the markers always align with a bar. This is probably musically correct. 
However the model allows overlapping bars, and especially when we have dynamic beat-gits created in earlier version of Mixxx, we can have tempo changes in any place. We need to deal with it. 

From the GUI perceptive, I would be in favour to require a bar for a temp change.

ywwg: for cases where the tempo is increasing gradually, every beat may be a little faster or slower than the last.  We should allow tempo changes on any beat marker.

JoergAtGithub: Perhaps a second parameter would be useful to describe accelerando/ritardando in beats/s². On the other hand, the case might be too rare to justify this overhead.

src/track/beats.cpp
fwcd: Assuming every `int` can safely be represented as `double`, which is the case on the overwhelming majority of desktop platforms where `int` is 32-bit and `double` is an 64-bit IEEE-754 floating point and given that `std::round` is guaranteed to round to the _nearest_ integer, this is usually safe. See https://stackoverflow.com/a/47153373/19890279.

I agree, however, that it might be more elegant to go with `std::lround`. We would still have to cast the `long` to an `int` though.

src/track/beats.h
fwcd: I agree with @ywwg, tempo changes on every beat are common enough that we should support them. The problem I see is that it's probably not easy to retrofit into our current beatgrid model (that has no information about bars/measures) if we cannot interpret markers as downbeats. So we'd probably have to defer this for now, just like changing time signatures, which I'd very much be interested in having too.

fwcd:
By the way, another idea I had for making the implementation of downbeats + changing time signatures backwards-compatible would be to store downbeat/time-signature change markers in a separate map/db column, e.g. in terms of beat indices (as opposed to frames to avoid redundancy). That way we could safely iterate on that format without impacting older Mixxx versions until the 2.5 release. Thoughts?

src/track/beats.h
daschuer: I see a chance to integrate that. 
For my understanding we need to add a an bar offset along with the beats per bar value. 
This way we are still able to express the old free beat map with the new model

```
     ‖___|___|___|___‖___|__|__|__‖__|__|___|___‖___     
     M                  M              M  . 
     120bpm.            180 bpm        120 bpm   
     barAt 0.           barAt 3        barAt 2
     beatsPerBar 4      beatsPerBar 4  beatsPerBar 4
 ```

daschuer:
The downside is that this needs another lookup. 
After a seek, we need to look up the beats from the time/sample position. Than we need to look up the downbeats Brom the brat numbers.

fwcd:
That's a fair point, though that lookup should be fairly quick if we do a binary search. I don't have a good enough overview of the current implementation, but if this is something that only becomes relevant at seeks, my intuition would be that that should be performant enough to be worth trying?

I think maintaining a backwards-compatible data format would be a huge plus, since I worry that once we start changing the format, we'll (understandably) want to take our time to get it right to avoid messing up the databases of users/testers/developers.

src/track/beats.h
ywwg: it may well be that bpm-per-measure is "close enough" that it works with bpm speed-up/slowdowns or non-quantized tracks (disco etc). If we can test this situation early then we'll know if it's a problem or not.  I have a few tracks that are my go-tos for difficult beatgrids, I can send them to you @fwcd if that helps

fwcd:
Did a quick rebase onto `main` to resolve the conflicts. Note that the downbeat markers currently only render with the legacy waveforms.

> Maybe the new system could be used alongside the old one and we could have two beatgrids for one track. At track load time, the new code would first do a lookup in the new table, and if the beat grid is not there, look for it in the old table and then do an upgrade to the new format. Old code would ignore the new table and leave it alone.

I think that's a neat idea, though the redundancy in keeping both grids around might manifest itself in quite a bit of space in large libraries. Also, I'm a bit worried that if we bake this into our schema now, the decision to keep two grids around per track is going to be set in stone once we ship it (if we want to keep backwards compatibility). Another issue is that I am not sure how well this would "scale" to future changes of the beat grid scheme, since we would effectively have to add a new column every time (a table is probably overkill in any case).

Storing just the "new" information would solve this if we could figure out how to make the lookups performant enough.

JoergAtGithub:
I think we should import the old library at the first start of the new version. We can ask the user in a dialog, if he wants to delete the old library or delete it.

fwcd:
Yeah, the question would be if we could make it backwards-compatible to avoid introducing a "hard" migration. I could see this being convenient, especially for developers which like to hop between different versions. The fact that I can run 2.3 without switching databases has proven useful several times while diagnosing issues.

JoergAtGithub:
Or we add the read functionality for the new format in Mixxx 2.5, but the downbeat feature that writes this format one release later.

fwcd:
I am not quite following, how would we use downbeats if we can't write them anywhere? Use a fresh grid every time the user opens the track?

JoergAtGithub:
Than 2.5 would only show the beats, but read from the new library format, ignoring all new elements like downbeats.

fwcd:
Ah I see, we would still break backwards compatibility with all versions older than 2.5 in that case though. I agree that it would be an option, but I also think it would be nice if users already had something to play around with. The specifics of the format should ideally be able to evolve alongside the UI as we figure out more details.

ywwg:
I like the idea of allowing the user to choose whether to keep the old grids or delete them. For me, disk space is cheap, and even the 4.4G I have in my analysis directory is not a big deal. We could also give the user the option to delete the old beats later if they kept them during the upgrade.  And when 2.6 comes out we can say "hey if you never plan on using 2.4 again we can delete this data".

ywwg:
A two-version migration for the hard format change is a great idea, +1!

ywwg:
@fwcd do you think you can work on this during the dev phase?

acolombier:
@fwcd I'd like to use your work as foundation for #13308, do you have integrest in finishing this work or would you be okay with me trying to finish it?

fwcd:
Sure, feel free to take over this PR (which is mostly just a rebase of @Holzhaus, nothing really novel here). I'm a bit short on time right now, so if you're interested, go ahead

fwcd:
Closing in favor of #13330

0cwa:
I've been running a version of Mixxx with this enabled so I think I understand ok this PR.  Because this is the recommended branch to work off from what I can tell by the last comment on #13330, which this was originally closed for: the only other item besides a setting button is to address is "open feedback from https://github.com/mixxxdj/mixxx/pull/4489 "  - what is it that needs to be addressed?  Most of what I read in #4489 was about allowing alternative beatgrids and being able to edit where the downbeat would be, right?


I see there being 2 main features to address to add this? :

1. Editing downbeat location, which I would guess would simply consist of: 
    1. Being able to set a new downbeat location (in case of a skipped beat)  
    2. Moving the whole grid forward/back by one beat (at least until the next manually set downbeat) 
    
(Only setting/editing the downbeats' locations (in case of skipped beats) would seem to keep the data structure compatible with future editable time signatures.)

2. Saving the info: 
    1. Saving the beatgrid alignment to the song.
    2. Intentionally saving if downbeats have been set as (enabled/disabled) on a song-by-song basis. (and by default loading as disabled for now)

Perhaps the above would be a minimum viable implementation which would be compatible with future customization?

Sorry if I misunderstand what is holding up this feature but the many implementation and sheer amount of discussion on this feature request has kinda been a bit of a black hole of info to dive into.  I think some summation of conclusions from the discussion about what would be acceptable for a release (ideally in a checklist like the above) would help a lot towards creating an acceptable implementation.  🙏

ywwg:
Thank you for your interest! We had a bit of a breakdown in consensus over the design of the feature. I think in order for this feature to move ahead we need to create a proper [Proposal](https://github.com/mixxxdj/proposals/) for it so we can hash out the design questions before we write code.  We are also discussing this issue right now in our Zulip developer chat. Please join us there and we can talk more!

PR #12344, opened 2023-11-23
Debug: Add Rust-inspired `DBG` macro for quick print-debugging

fwcd:
During debugging, I often find myself logging a bunch of variables to retrace/understand some part of the codebase. While this is essentially a very crude, quick-and-dirty form of print debugging, it seems to be so common that the Rust standard library has a macro for this, [`dbg!`](https://doc.rust-lang.org/std/macro.dbg.html).

Therefore, to make it easier to quickly debug Mixxx, this PR adds a heavily Rust-inspired and quick-to-type `DBG` macro that writes an expression in both its quoted and evaluated form to `qDebug`, along with the file and line to make it easy to quickly locatable in the source code.

To avoid accidentally including `DBG` invocations in release builds, the macro is redefined to emit a message if debug assertions are disabled. Ideally this should be promoted to an error, but I couldn't figure out a way to do so (from within the macro definition where we can't just use `#error`), also the QML library which includes this header currently seems to be built without debug assertions even if `DEBUG_ASSERTIONS_FATAL` is set.

Thoughts?

### To do

- [x] Target 2.4
- [ ] Figure out how to shorten the file path format

### DIFF
--- a/src/util/debug.h
+++ b/src/util/debug.h
@@ -9,12 +9,10 @@
 #define DBG(value) \
     (qDebug().nospace() << #value << " = " << (value) << " [" << __FILE__ << ":" << __LINE__ << "]")
 #else
-// We have to stop clang-format from breaking up the string literal since that
-// breaks the pragma (at least when using Clang).
-// clang-format off
-#define DBG(value)                                                            \
-    _Pragma("message \"DBG(...) should not be used in a build without debug assertions, please remove it! This will be a no-op.\"")
-// clang-format on
+#define DBG(value)                                                             \
+    static_assert(false,                                                       \
+            "DBG() is not allowed in builds without debug assertions, please " \
+            "remove it!")
 #endif
 
 template <typename T>
--- a/src/util/debug.h
+++ b/src/util/debug.h
@@ -7,7 +7,7 @@
 
 #ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
 #define DBG(value) \
-    (qDebug().nospace() << #value << " = " << value << " [" << __FILE__ << ":" << __LINE__ << "]")
+    (qDebug().nospace() << #value << " = " << (value) << " [" << __FILE__ << ":" << __LINE__ << "]")
 #else
 // We have to stop clang-format from breaking up the string literal since that
 // breaks the pragma (at least when using Clang).
--- a/src/util/debug.h
+++ b/src/util/debug.h
@@ -6,9 +6,8 @@
 #include "errordialoghandler.h"
 
 #ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
-#define DBG(value)                                                       \
-    (qDebug().nospace() << __FILE__ << ":" << __LINE__ << ": " << #value \
-                        << " = " << value)
+#define DBG(value) \
+    (qDebug().nospace() << #value << " = " << value << " [" << __FILE__ << ":" << __LINE__ << "]")
 #else
 // We have to stop clang-format from breaking up the string literal since that
 // breaks the pragma (at least when using Clang).
--- a/src/util/debug.h
+++ b/src/util/debug.h
@@ -5,6 +5,19 @@
 
 #include "errordialoghandler.h"
 
+#ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
+#define DBG(value)                                                       \
+    (qDebug().nospace() << __FILE__ << ":" << __LINE__ << ": " << #value \
+                        << " = " << value)
+#else
+// We have to stop clang-format from breaking up the string literal since that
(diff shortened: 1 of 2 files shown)


src/util/debug.h
```diff
@@ -5,6 +5,19 @@
 
 #include "errordialoghandler.h"
 
+#ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
+#define DBG(value)                                                       \
+    (qDebug().nospace() << __FILE__ << ":" << __LINE__ << ": " << #value \
+                        << " = " << value)
```
fwcd, line 11: The format would also be up for debate, this can quickly become a bit hard-to-read if the repo is cloned to a path deep in the filesystem. Suggestions are welcome!

daschuer:
Nice, thank you. Yes that can go to 2.4. 

Did you consider to swap the value and the location like in rust? 
I think this way we get a nicer "table" in a loop. Is there something else to consider? 
Maybe place a comment. 

Currrently a DEBUG_ASSERT is printed like this: 

```
critical [CachingReaderWorker 13] DEBUG ASSERT: "m_value < origValue" in function mixxx::Beats::ConstIterator mixxx::Beats::ConstIterator::operator-=(mixxx::Beats::ConstIterator::difference_type) at /home/sperry/workspace/mixxx2/src/track/beats.cpp:123
```
We may also consider to align the output. 

I think you can use 
`std::static_assert(false, "DBG(...) should not be used in a build without debug assertions ....") `

Maybe we can rephrase it. `DBG() should not be remain in release builds ..`

src/util/debug.h
ronso0: Indeed, that bothers me with the current debug asserts as well.
A way to remove the root dir down to `src/` would be nice.

fwcd:
> Did you consider to swap the value and the location like in rust?

Yeah, that was my first idea (see the comment above), but quickly got unreadable due to long paths of varying length. Surprisingly, the variable-first approach was more readable, probably since many variables are of somewhat similar length.

I think the file path-first could work if we could shorten it to only include the relative path (the filename alone might not be unambiguous and in many cases doesn't let terminals like VSCode's format it as a link). GCC and Clang seem to offer `-fmacro-prefix-map` which changes the prefix for all `__FILE__`s: https://stackoverflow.com/a/53848526 (which could btw also be neat to avoid baking GitHub runner paths into the binaries).

fwcd:
> I think you can use `std::static_assert(false, "DBG(...) should not be used in a build without debug assertions ....")`

Neat, the static assert trick is really clever!

> Maybe we can rephrase it. `DBG() should not be remain in release builds ..`

I've avoided phrasing it "release builds", since release builds may still set `-DDEBUG_ASSERTIONS_FATAL=ON` (that's what I frequently do too, albeit with `RelWithDebInfo`). We could of course change the condition to be something else (wouldn't have to be debug assertions), but the error message IMO should clearly reflect the build setting that prevents this from compiling.

src/util/debug.h
fwcd: I would probably keep `src/`, that seems to play a bit better with IDEs when opening the `mixxx` repo there and also handles the case where we hit something in `lib/` code (which should become less likely as we move more dependencies to `vcpkg` though). Using `-fmacro-prefix-map` would relativize all `__FILE__` paths.

daschuer (approved):
Looks good, thank you.

src/util/debug.h
fwcd: - #12350

src/util/debug.h
```diff
@@ -5,6 +5,16 @@
 
 #include "errordialoghandler.h"
 
+#ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
+#define DBG(value) \
+    (qDebug().nospace() << #value << " = " << (value) << " [" << __FILE__ << ":" << __LINE__ << "]")
```
Swiftb0y, line 10: I think the real utility of rusts debug macro comes from the fact that it returns the value that has been passed in. That enables it to be inserted into any expression at will. The merged solution doesn't do that unfortunately, but it is achievable with a an IILE:
```suggestion
    ([](auto&& val) {
        qDebug().nospace() << #value << " = " << val << " [" << __FILE__ << ":" << __LINE__; << "]"; 
        return val;
    })(value)
```
Not sure how the types for this work out though.

ronso0:
Sounds useful, how is it supposed to be used? Like the SKIN_WARNING macro?

Swiftb0y:
No, more like quick and dirty printf-debugging. The power of my suggestion comes from the fact that you can use it within complicated expressions:
```c++
foo(DBG(DBG(a+b)*c/d);
```
See the rust doc for better examples:
https://doc.rust-lang.org/std/macro.dbg.html#further-examples

Issue #12336, opened 2023-11-22
Manual does not explain the various MIDI bind options available, like Spread64, HercJog, Rot64, Rot64Invert, Rot64Fast, etc.


mxmilkiib:
### Bug Description

https://github.com/mixxxdj/mixxx/blob/main/src/controllers/midi/midicontroller.cpp#L397

https://github.com/mixxxdj/mixxx/blob/main/src/controllers/midi/midimessage.h#L102

ronso0:
Info can be copied from https://github.com/mixxxdj/mixxx/wiki/Midi-Controller-Mapping-File-Format#input-options

ronso0:
Btw, manual issues go to https://github.com/mixxxdj/manual/issues

ronso0:
Ah, I see:
the manual only mentions the options available in the MIDI Learning wizard. There are more options available in the Input table, Options column.

So another issue is that not all options are available in the Learning dialog?

PR #12348, opened 2023-11-24
MIDI Input editor: allow selecting multiple Options

### LINKED ISSUE #7416: allow multiple MIDI option selection via GUI
Reported by: **[daschuer](https://github.com/daschuer)**
Date: 2014-04-11T06:38:53Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1306397](https://bugs.launchpad.net/bugs/1306397)
Tags: controllers, midi, polish

---

This is a copy of a discussion started at https://github.com/mixxxdj/mixxx/pull/214

daschuer:
Is there a way to edit the mapping options e.g. add softtakover to an existing mapping?

kain88-de: 
Can we add this as a checkbox-column to the QTable, so that I can also see for which controlls softtake over is activated?

rryan: 
That's the MIDI Options column in the MIDI mapping table. I couldn't get a multi-select delegate working so it's just one midi-option selection now using the table now.

daschuer: 
But what about an "edit" button, that just opens the midi wizard with the CO preselected focused on the table?

ronso0:
Multiple options can be set in the learning dialog (not all available options are shown there https://github.com/mixxxdj/mixxx/issues/12336#issuecomment-1825007161) and by editing the mapping XML, of course. Though, editing multi-select in the mapping table was not possible since selecting an option would deselect all others.

This PR allows to set multiple options in the list view of the Options combobox.
Either click / press Enter on a selected option to toggle it and close the list, or press Space on an option to toggle it and keep the list open.
image

It took me a while to figure that the 'Normal' option should not be in options list now that it allows multiple options. Due to how the options flags work here 'Normal' is exclusive, meaning it's the implicit result of unchecking all other options, so showing it checked is pointless (and since it's a dumb combobox it's not unchecked immediately if others are checked). Also, clicking it does not uncheck all other options.

The second commit removes 'Normal' entirely. IMO the mapping table is much clearer without it, individual mappings that have options set are now much easier to spot.

The third commit (experimental) adds an 'Unset all' action. Please test if you think if that is helpful in real life. I didn't manipulate mappings on that level for very long.

____

I'm targeting `main` since this doesn't work with Qt5 as is unfortunately. Checkboxes are not shown and the check state can't be changed. I tried with setting QAction icons in c++ but handling the checkstate was a mess.
An alternative approach with a QMenu is to complex for this small improvement.

Closes #7416

### DIFF
--- a/src/controllers/delegates/midioptionsdelegate.cpp
+++ b/src/controllers/delegates/midioptionsdelegate.cpp
@@ -61,6 +61,11 @@ QWidget* MidiOptionsDelegate::createEditor(QWidget* parent,
         pItem->setCheckable(true);
         pModel->appendRow(pItem);
     }
+    // Add a special item to uncheck all. See commitAndCloseEditor()
+    QStandardItem* pItem = new QStandardItem(tr("Unset all"));
+    pItem->setCheckable(false); // doesn't hurt to set this explicitly
+    pModel->appendRow(pItem);
+
     // Unsetting the index clears the display text which is visible when closing
     // the list view by clicking anywhere else. Default text is that of first
     // added item. It can be set to any combination of existing item texts, e.g.
@@ -112,6 +117,9 @@ void MidiOptionsDelegate::setEditorData(QWidget* editor,
     DEBUG_ASSERT(pModel);
     for (int row = 0; row < pModel->rowCount(); row++) {
         auto* pItem = pModel->item(row, 0);
+        if (!pItem->isCheckable()) {
+            continue;
+        }
         auto opt = static_cast<MidiOption>(pItem->data().toInt());
         pItem->setCheckState(options.testFlag(opt) ? Qt::Checked : Qt::Unchecked);
     }
@@ -151,7 +159,18 @@ void MidiOptionsDelegate::commitAndCloseEditor(int index) {
     DEBUG_ASSERT(pModel);
     auto* pItem = pModel->item(index);
     DEBUG_ASSERT(pItem);
-    pItem->setCheckState(pItem->checkState() == Qt::Checked ? Qt::Unchecked : Qt::Checked);
+    if (pItem->isCheckable()) {
+        pItem->setCheckState(pItem->checkState() == Qt::Checked ? Qt::Unchecked : Qt::Checked);
+    } else {
+        // Clear was selected. Uncheck all other items
+        for (int row = 0; row < pModel->rowCount() - 1; row++) {
+            if (row == index) { // Actually it's the last item, but this is safer.
+                continue;
+            }
+            pItem = pModel->item(row, 0);
+            pItem->setCheckState(Qt::Unchecked);
+        }
+    }
 
     emit commitData(pComboBox);
     emit closeEditor(pComboBox);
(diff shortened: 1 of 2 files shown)


ronso0:
I adjusted the description to reflect the current state. It works smoothly now.

There's a tiny regression compared to the default combobox: usually, when closing the list view by clicking elsewhere, the (closed) combobox would show the seleted item. With multi-select the display string is updated only after the editor is closed (combobox hidden, data applied to model) so it would not reflect the new state. I choose to clear the index (no name displayed) to avoid confusion. I tried closing the editor as soon as the list is closed but that is cumbersome (subclass QComboBox, handle events) and may not work 100% as expected in all situations.

daschuer:
We have windows CI error.

ronso0:
(force-pushed to trigger CI, logs have already been deleted)

ronso0:
...and now also Windows builds it just fine.

src/controllers/delegates/midioptionsdelegate.cpp
```diff
@@ -66,28 +106,72 @@ QString MidiOptionsDelegate::displayText(const QVariant& value,
...
-        if (MidiOptions(pComboBox->itemData(i).toInt()) & options) {
-            pComboBox->setCurrentIndex(i);
-            return;
+
+    // Update checked states
+    const MidiOptions options = index.data(Qt::EditRole).value<MidiOptions>();
+    const auto* pModel = static_cast<QStandardItemModel*>(pComboBox->model());
+    DEBUG_ASSERT(pModel);
+    for (int row = 0; row < pModel->rowCount(); row++) {
+        auto pItem = pModel->item(row, 0);
```
daschuer, line 119:
```suggestion
        auto *pItem = pModel->item(row, 0);
```

src/controllers/delegates/midioptionsdelegate.cpp
```diff
@@ -66,28 +106,72 @@ QString MidiOptionsDelegate::displayText(const QVariant& value,
...
+    DEBUG_ASSERT(pItem);
+    if (pItem->isCheckable()) {
+        pItem->setCheckState(pItem->checkState() == Qt::Checked ? Qt::Unchecked : Qt::Checked);
+    } else {
+        // Clear was selected. Uncheck all other items
+        for (int row = 0; row < pModel->rowCount() - 1; row++) {
+            if (row == index) { // Actually it's the last item, but this is safer.
+                continue;
+            }
+            auto pItem = pModel->item(row, 0);
```
daschuer, line 170:
```suggestion
            auto *pItem = pModel->item(row, 0);
```

daschuer (changes requested):

src/controllers/delegates/midioptionsdelegate.cpp
ronso0: actually just `pItem = pModel->item(row, 0);` because it's already declared a few lines above.

daschuer:
It works better than before. However now concurrent bits can be checked. This can be don a in a follow up PR during beta or even later when you like. Else, we can merge after cleanup

ronso0:
> However now concurrent bits can be checked.

Hmm, so pretty much like it is possible via XML, just much easier.
I've filed #13254 and added a comment.

daschuer (approved):
LGTM, Thank you.

PR #12352, opened 2023-11-25
ffmpeg and Qt include impovements

daschuer:
Another tidy up PR 
The ffmpeg headers are included elsewhere with  `extern "C"` so we should do it everywhere to not trick the linker.


JoergAtGithub:
Built it locally on Win11. All fine! Thank you!

Issue #12358, opened 2023-11-26
Windows: Event loop is blocked for 500ms, when mouse down on title bar of any Mixxx window


JoergAtGithub:
### Bug Description

In the video you see how the waveforms stop, when I left click on the title bar:
https://github.com/mixxxdj/mixxx/assets/64457745/bc7ea543-b9bc-4dbf-8f77-56ed827f9d96

The handling of this event by Windows blocks the whole Qt event loop for exact 500ms. This only happens if Windows is configured with the option "Show window contents while dragging" which is the standard.

This is a general WIN32 API behavior and not Qt related. It happens for every application when Windows handles the message [WM_NCLBUTTONDOWN](https://learn.microsoft.com/en-us/windows/win32/inputdev/wm-nclbuttondown)
It's not new as you can see in this discussion from 2009: https://www.gamedev.net/forums/topic/520860-sizemove-loop-and-delay-in-defwindowproc/4382282/

In the thread view, you can see how the Mixxx Main Thread blocks the other threads (you clearly see the 500ms blocks when I clicked on the title bar):
![grafik](https://github.com/mixxxdj/mixxx/assets/64457745/6d3d34fb-521a-419c-b4aa-4350c845601d)

### Version

2.4

### OS

Windows 11

PR #12359, opened 2023-11-26
Fix 500ms blocking of the whole event loop, when holding mouse down on title bar on Windows

JoergAtGithub:
Closes: #12358

This adapts a solution for this problem from https://github.com/rust-windowing/winit/pull/839 

Before:
https://github.com/mixxxdj/mixxx/assets/64457745/2886ab96-2dfa-4b5b-ad22-35b23a978a02

After:
https://github.com/mixxxdj/mixxx/assets/64457745/956f2a83-866c-46f6-9b62-74f44a36c5ec

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -617,16 +617,8 @@ endif()
 
 set(CLANG_TIDY "" CACHE STRING "CMAKE_CXX_CLANG_TIDY equivalent that only applies to mixxx sources, not bundled dependencies")
 
-if(WIN32)
-  set(NATIVE_SOURCE_FILES src/nativeeventhandlerwin.cpp)
-else()
-  set(NATIVE_SOURCE_FILES "")
-endif()
-
-
 # Mixxx itself
 add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
-  ${NATIVE_SOURCE_FILES}
   src/analyzer/analyzerbeats.cpp
   src/analyzer/analyzerebur128.cpp
   src/analyzer/analyzergain.cpp
@@ -966,6 +958,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/musicbrainz/web/coverartarchiveimagetask.cpp
   src/musicbrainz/web/coverartarchivelinkstask.cpp
   src/musicbrainz/web/musicbrainzrecordingstask.cpp
+  src/nativeeventhandlerwin.cpp
   src/network/jsonwebtask.cpp
   src/network/networktask.cpp
   src/network/webtask.cpp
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -617,8 +617,16 @@ endif()
 
 set(CLANG_TIDY "" CACHE STRING "CMAKE_CXX_CLANG_TIDY equivalent that only applies to mixxx sources, not bundled dependencies")
 
+if(WIN32)
+  set(NATIVE_SOURCE_FILES src/nativeeventhandlerwin.cpp)
```
daschuer, line 621: Until now, we have the concept of adding all files and #ifdef the content out. 
see util/console.cpp and util/mac.cpp 
I am not sure about the implications with different IDEs, but I think we should stick with one concept. 
In case of Eclipse the file is showed grayed out in the #ifdef region.

CMakeLists.txt
```diff
@@ -2738,6 +2746,9 @@ elseif(UNIX AND NOT APPLE)
     Qt${QT_VERSION_MAJOR}::DBus
   )
 elseif(WIN32)
+    target_link_libraries(mixxx-lib PRIVATE
+    Comctl32
```
daschuer, line 2750: this is in the Qt related block. 
No problem, but we have also 
`target_link_libraries(mixxx-lib PRIVATE shell32)` and than `shell32` later in the static build block. 
Can you join that?

src/nativeeventhandlerwin.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+WindowsEventHandler::WindowsEventHandler(MixxxApplication* pApp)
+        : m_pApp(pApp) {
+}
+
+bool WindowsEventHandler::nativeEventFilter(
+        const QByteArray& eventType, void* message, long* result) {
+    MSG* msg = (MSG*)message;
+    if (msg && msg->message == WM_NCLBUTTONDOWN) {
+        // Trigger the modal loop to prevent 500ms wait in Event Loop
```
daschuer, line 17: Do we also have a reference where the reason of the issue is explained?

src/nativeeventhandlerwin.cpp
```diff
@@ -0,0 +1,35 @@
...
+                RDW_INTERNALPAINT);
+
+        if (msg->wParam == HTCAPTION) {
+            PostMessageW(msg->hwnd, WM_MOUSEMOVE, 0, 0);
+        }
+
+        DefSubclassProc(msg->hwnd, msg->message, msg->wParam, msg->lParam);
+    }
+
+    return false;
```
daschuer, line 34: can you also add a comment about always returning false?

daschuer (changes requested):
Thank you for the nice fix. I have added some comments about improved code maintenance.

JoergAtGithub:
All suggestions implemented!

daschuer (approved):
LGTM, Thank you.

PR #12360, opened 2023-11-26
Add missing header to console.cpp

daschuer:
No idea why this has popped up during my tests. But it is anyway correct.


JoergAtGithub:
LGTM! Waiting for CI!

PR #12364, opened 2023-11-26
Fix svg rendering on macOS retina

### LINKED ISSUE #12361: many images not rendered at retina resolution on macOS
### Bug Description

On macOS retina, many images are not rendered at HDPI. I always noticed this with the rec button, the on air icon, the settings button, and the MIXXX logo, but on closer inspection, it affects many more icons.

This can be clearly observed in this zoomed screenshot composition: the pixels of the images are double the size of the pixels of the text, which is rendered correctly. Note that the images are all svgs.

Screenshot 2023-11-26 at 13 44 21

### Version

2.4.0

### OS

macOS

m0dB:
This fixes https://github.com/mixxxdj/mixxx/issues/12361 by upscaling the  btn__ svgs with a factor 2, by multiplying to document size of each btn__ svgs by two and inserting a 2x scale transform, for LateNight palemoon and classic. (This way this could be automated, rather than opening and resizing each svg manually)

It is surprising that this is necessary, but apparently Qt renders these svgs, when rendering with stylesheets, at the document size in pixels and only scales down, never up. I have been looking for a code-based solution but to no avail.

Apparently this is not necessary on windows / linux, so it looks like the HDPI handling is different then on macOS. I expect that this upscaled svgs have no negative effect on windows / linux, but please confirm.

Latenight palemoon:

Before:

before

With upscaled svgs:

after2

Latenight classic:

Before

classic-before

After

classic-after


Swiftb0y:
This is a workaround, isn't it? We should file a bug report with them. Does the issue persist in Qt6 builds?

JoergAtGithub:
Qt5 seems to be different here than Qt6: https://doc.qt.io/qt-5/highdpi.html

Did you had a look at https://github.com/mixxxdj/mixxx/blob/2c2e706b4419794e1b29f23f124ce65f49b88bc7/src/main.cpp#L145
Maybe these settings can influence the behavior.

m0dB:
> This is a workaround, isn't it?

Yes.

> We should file a bug report with them. Does the issue persist in Qt6 builds?

I still need to try this.

m0dB:
> Qt5 seems to be different here than Qt6: https://doc.qt.io/qt-5/highdpi.html
> 
> Did you had a look at
> 
> https://github.com/mixxxdj/mixxx/blob/2c2e706b4419794e1b29f23f124ce65f49b88bc7/src/main.cpp#L145
> 
> 
> Maybe these settings can influence the behavior.

IMO these flags are all set as they should and changing them didn't help.

ywwg:
this may be a workaround for a QT bug but I think it's a fine workaround and worth merging in to fix the issue

ywwg:
looks good in linux

ywwg (approved):

daschuer:
The rendering is done here: 
https://github.com/qt/qtbase/blob/da6e958319e95fe564d3b30c931492dd666bfaff/src/widgets/styles/qstylesheetstyle.cpp#L1402
It is done form a QIcon with a list of pre-rendered QPixmaps in different sizes. 
https://github.com/qt/qtbase/blob/c5ba3ac95cb7dc157a4d84ecc32cdadd7d45d387/src/gui/image/qicon.cpp#L268
Is can be populated from style sheets with the @ prefix for different scaling.  

This means we have no real benefit from the scaling of a svg. It is treated like png, just with the extra penalty of rendering from svg. Your current solution looks good for 2x but not for other scaling and comes with a performance penalty for smaller (non 2x) screens. 
 
I am not sure how the situation is with the "BackPath" skin property. Did you check that? 
Like here: 
https://github.com/mixxxdj/mixxx/blob/2c2e706b4419794e1b29f23f124ce65f49b88bc7/res/skins/Tango/mixer/mixer_main.xml#L25

Will it work? What do you think?

daschuer:
First of all, I have no strong demands which solution we go. I just want to make sure we have taken a decision with all informations on the table. 

@m0dB would you mind to test with one knob, if the unscaled svg is blurry removing it from the style and adding it to the skin? 
I think it is here: 
https://github.com/mixxxdj/mixxx/blob/2c2e706b4419794e1b29f23f124ce65f49b88bc7/res/skins/LateNight/style_classic.qss#L1732
https://github.com/mixxxdj/mixxx/blob/2c2e706b4419794e1b29f23f124ce65f49b88bc7/res/skins/LateNight/controls/button_1state_right.xml#L17

> We definitely do have benefit from rendering svgs, and they do scale properly! 

The issue is that they are rendered and scaled with their native size to a QPixmap in QIcon, then they are sacaled from that to the final size. This is the reason we see blurriness whenever an upscaling or an odd number scaling happens. 

I think with "BackPath", we render directly to the final size. So we safe CPU and memory for one rendering. 

> But this would require changing a lot of skin code, wouldn't it? I don't want to go there.

Yes, that is the issue. I can imagine that all the scaling done here was already tedious. 
So let's first check if the "BackPath" solution really solves the issue as well. Than we have a clear view.

m0dB:
With BackPath the svg renders at the correct resolution:

Screenshot 2023-11-27 at 12 54 05

ronso0:
> I am not sure how the situation is with the "BackPath" skin property. Did you check that?

This means we'd have to drop the convenient qss solution and set all icons in xml, including potential scheme suffix :|

> With BackPath the svg renders at the correct resolution:
> Screenshot 2023-11-27 at 12 54 05

The drawback of setting backgroundPathis obviously that it doesn't center the icon and allows no additional position correction like in qss.

@m0dB 
> instead of modifying the original files, I could add a step that modifies the files when the are copied to the resource, only on macOS. That way we don't "polute" the original svgs, and the workaround is restricted to macOS.

if that is feasible it very much appreciated and I'd definitely prefer that over the skin changes, either SVG or xml.
Could you give that a shot?

daschuer:
Don't we need the dynamic solution on the users machine? We don't know which scale factor the user has.

daschuer:
Is adding both with a @2x prefix an option?

m0dB:
> Don't we need the dynamic solution on the users machine? We don't know which scale factor the user has.

My proposal was to replace the svgs with the 2x versions only on macOS, where retina is pretty much the norm. 

But your comment made me think: we could install both always, and change the search path order based on the devicePixelRatio.

m0dB:
> Is adding both with a @2x prefix an option?

You think that will work based on the Qt code? If it does, that would be perfect.

daschuer:
I can confirm the same blurriness on windows and Linux as well at 200 % scaling: 
grafik

JoergAtGithub:
I tested on 400% and it looked good. Can you confirm this?

ronso0:
Here it's looking good @200% (though Qt icons in the preferences look blurry, checkboxes, radio buttons, up/down arrows etc.)

So we have
* broken
   * macOS (Qt version @m0dB ?)
   * Windows (version? Qt version?)
* good
   * Windows 11 (official vcpkg?)
   * Ubuntu 20.04, Qt 5.12.8

m0dB:
> I can confirm the same blurriness on windows and Linux as well at 200 % scaling:
> grafik

is that what or without this PR?

m0dB:
> Is adding both with a @2x prefix an option?

You mean postfix, right? I doesn't seem to open the @2x file at all.

ronso0:
I assume @daschuer refers to this, but that is never called (or is it called by Qt?). Even though I vaguely remember discussing this high-res image solution I never tested it IIRC. Maybe it's broken and no one noticed..
https://github.com/mixxxdj/mixxx/blob/8e41e53adcb025e009989add2f5b5129454c286c/src/widget/paintable.cpp#L278

On the Qt side there is https://doc.qt.io/qt-6/scalability.html mentioning @2x, though IIUC it refers to images in the qrc file.
And https://doc.qt.io/qt-5/qicon.html#addFile, not sure if that is called when setting images via stylesheets.

daschuer:
Ok, I think we have now discussed most aspects. 
This PR is definitely an improvement, but unfortunately still does not render the SVG directly to the desired size. 

So we have the choice to merge this and accept the remaining edge cases and the extra CPU on start up or replace it with the "BackPath" + center solution where possible. 

A macOS only upscaling on our build server seems to be not required. 

What do you think?

m0dB:
I got a bit lost with all the observations. So there is no visual degradation on non-retina setups?

In any case, I do see a simple solution to modify the path to the svgs on the fly: `LegacySkinParser::stylesheetAbsIconPaths`
We could have the 2x svgs in a separate directory and where required (aka on macOS), in this function we could change the path in the style automatically.

On the long run, moving the svgs out of the style seems to be best solution, but I find it hard to say how much work that would be on a qss/xml level, and if everything now done in the styles could be done with the xml tags, and if not, what would need to be added.

daschuer:
All targets share the same problem:
*  qss SVGs are first rendered in their original size and than scaled to their target size 
* Even on macOs we have fractional scaling like 125 % with the help of the spin box in the interface preferences. 

> On the long run, moving the svgs out of the style seems to be best solution

This gives the best visual result and performance. 
The costs are, that not every Image can be moved, we loose convenience with maintenance in one place, and it requires extra work and testing.

ronso0:
> In any case, I do see a simple solution to modify the path to the svgs on the fly: LegacySkinParser::stylesheetAbsIconPaths

:+1: 
This can be applied to external style sheets and `<Style>` attributes in xml.
Note that icons are also set in `default.qss`, picked from the qrc file, so we'd also need to consider
* other paths, e.g. `url(:/images/`
* other properties `qproperty-icon: url(:/..`

> I got a bit lost with all the observations.

I think we mixed up native app scaling (macOS retina, Windows 125% @daschuer mentioned) and Mixxx' internal scaling (which I tested, and @JoergAtGithub as well?)

> On the long run, moving the svgs out of the style seems to be best solution, but I find it hard to say how much work that would be on a qss/xml level, and if everything now done in the styles could be done with the xml tags, and if not, what would need to be added.

It is not possible to set all icons in xml, for example AFAIK scrollbars, the treeview decoration and the buttons in the library panes (AutoDJ, Recording, Analyze, Missing, Hidden) and menu checkboxes.
And, as I mentioned, there are drawbacks in terms of flexibility when setting images in xml.
I just can't believe Mixxx is the only app affected by this, hence there must be another fix available.

I'm reluctant to invest that much work into a skin system we plan to replace anyway.

m0dB:
I have put all 2x scaled svgs in a subdirectory `2x/` (at the level of the original file) and in `LegacySkinParser::stylesheetAbsIconPaths` I check for these files and use them when available. If we want to make the usage of these files conditional, we can do that easily now.

Does this work? Should this be macOS only or no need?

ronso0:
Thank you. I don't see any impact on startup time caused by the scaling SVG -> pixmap with my i7, still floating around 6s. Would be good to check how this affects low-spec machines. @daschuer could you test with your eeePC?

Some small issues: icons set on QWidgets need the size fixed in qss to be scaled correctly.
Without fixed size (current state):
Screenshot_2023-12-06_00-35-06


fixed size, one dimension is sufficient, aspect ratio is maintained, smaller dimension is dominant appearantly
https://github.com/ronso0/mixxx/commit/5b48d4e11b51af1ab80763d50af149fec73dd5a6

Still loking good: spinboxes in decks and library, all menus (main menu, decks, library), AutoDJ controls, tracks table in-line editors, effect selectors.
This is probably a last-minute fix so anyone testing this please check this thoroughly on your machine, too.

ronso0:
> AutoDJ controls

They're also too big, don't know why that looked okay earlier.
I'll provide another commit.

ronso0:
I set the scaling to 200% in Mixxx and that also draws oversized icons.
@m0dB Are the AutoDJ control icons rendered correctly in 2.4 for you?
If yes, they need to be excluded from this fix, and we should add a comment to the stylessheet (style-palemoon.qss because I consider it a reference, given that the actual skin documentation on the wiki is outdated, and doesn't cover such quirks anyway).

m0dB:
Thanks for dedicating some time to this, @ronso0 . I will merge your commit that fixes the size.

As for the AutoDJ icons: I see what you see: on 2.4 they are rendered without non-retina resolution, with the 2x scaled svgs they are rendered too big.

m0dB:
What is strange though: on the autodj buttons the icons are too big, but they are rendered using the retina dpi.

daschuer:
I cannot measure a significant difference in start-up time and memory usage. Maybe the original rendering is only a temporary.

So we may pick a size, that is a common denominator of all common scalings.     
I think on windows it is 
100 % 125 % 150 %  200 %

4x would work, need to check.  

We "just" need to make sure that the final rendering is exactly 1/4 of the source. Otherwise we will se blurry again at different scaling. This is unfortunately not yet the case everywhere.

m0dB:
Okay, I am finding something very interesting: these images have both width and height and viewport. I am experimenting with `2x/btn__autodj_enable_off.svg` : When I remove the <g transform="scale(2)"> , the image renders correctly. If I double the viewport width and height, the image becomes half the size. It might be that with the correct combination of width, height and viewport values we don't need the scaling at all?

m0dB:
Right! For example btn__broadcast_off.svg renders fine with:
```
<svg width="132" height="32" viewBox="0 0 66 16" version="1.1" xmlns="http://www.w3.org/2000/svg">
```
and no need to insert the `<g transform="scale(2)">`

originally:
```
<svg width="66" height="16" version="1.1" xmlns="http://www.w3.org/2000/svg">
```

This might work everywhere! I can't look into this right now, but this might also make the fixed sizes you added to the qss unnecessary.

daschuer:
issues at 125 % using this branch: 
grafik (gray bottom) 
grafik (starecase) 
Interestingly they are not visible at 250 % even if it is also not a integer fraction.  

Maybe we cab stick with 2x for all and just tweak the original svg for better double scaling.

m0dB:
replaced by https://github.com/mixxxdj/mixxx/pull/12407

PR #12349, opened 2023-11-24
Beats: Fix unsafe overflow checks and debug asserts

fwcd:
### Fixes #12071 (for me, testing would be appreciated!)

With this patch Mixxx no longer crashes after loading variable BPM tracks with debug assertions enabled.

As noted in https://github.com/mixxxdj/mixxx/issues/12071#issuecomment-1825743901, the overflow checks in `Beats::ConstIterator` were rather unsafe and probably even undefined behavior, therefore this patch replaces them with a safe check.

Additionally, I have updated the `DEBUG_ASSERT`s to accept equality too. This should make sense, since we can't move the iterator further once we're at the limit, as the warning should indicate:

```
warning [CachingReaderWorker 1] Beats: Iterator would go out of possible range, capping at earliest possible position.
```

It should be noted, however, that I still think there's an issue somewhere in the (de)serialization machinery of `BeatMap-1.0`, since this feels like an edge case we shouldn't be hitting, even for variable BPM maps. Some more debugging would probably be helpful, but for now this should hopefully at least fix the assertion.

Issue #12071, opened 2023-10-07
DEBUG ASSERT: "m_value < origValue" in function mixxx::Beats::ConstIterator


Holzhaus:
### Bug Description

Just encountered this. When I run the tests again, everything passes.

```
 91/808 Test #387: HotcueControlTest.SavedLoopActivate ........................................................***Exception: Interrupt  0.35 sec
QML debugging is enabled. Only use this in a safe environment.
info [0x55f4d12c7180] Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
Note: Google Test filter = HotcueControlTest.SavedLoopActivate
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from HotcueControlTest
[ RUN      ] HotcueControlTest.SavedLoopActivate
[...]

PR #12368, opened 2023-11-27
Fix beat iterator clamping 

daschuer:
This is the second part of the fix for https://github.com/mixxxdj/mixxx/issues/12071
It continues https://github.com/mixxxdj/mixxx/pull/12349

This PR avoids the warning about the iterator clamping and introduces a DEBUG assert for cases that must not have slipped through.

### DIFF
--- a/src/track/beats.cpp
+++ b/src/track/beats.cpp
@@ -67,8 +67,8 @@ Beats::ConstIterator Beats::ConstIterator::operator+=(Beats::ConstIterator::diff
     // Detect integer overflow in `m_beatOffset + n`
     const int maxBeatOffset = std::numeric_limits<Beats::ConstIterator::difference_type>::max();
     if (m_beatOffset > maxBeatOffset - n) {
-        qWarning() << "Beats: Iterator" << m_beatOffset << "+" << n
-                   << "would go out of possible range, capping at latest possible position.";
+        qDebug() << "Beats: Iterator" << m_beatOffset << "+" << n
+                 << "would go out of possible range, capping at latest possible position.";
         m_it = m_beats->m_markers.cend();
         m_beatOffset = maxBeatOffset;
         updateValue();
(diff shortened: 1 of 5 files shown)


src/track/beats.cpp
```diff
@@ -595,7 +597,12 @@ mixxx::Bpm Beats::getBpmAroundPosition(audio::FramePos position, int n) const {
 
     // To make sure we are always counting n beats, iterate backward to the
     // lower bound, then iterate forward from there to the upper bound.
-    it -= n;
+    for (int i = 0; i < n; ++i) {
+        if (it == cbegin()) {
+            break;
+        }
+        it--;
```
daschuer, line 604: I am unsure about this solution because it has performance implications. 
We go here step by step backward to not reach cbegin() and hit the assertion. 
The code is able to do it in one go. If we rely on this, we need to remove the warning and the assertion form the -= operator. 

What do you think?

daschuer: The same below, we first hit the assertion and than do the correction.

daschuer:
I think here we should go for performance first and remove the Warning. 
Can it happen in the engine thread? That would be one reason more for removal.

daschuer:
In favor of a good performance,  I have removed the loop and lowered the warning to debug message. This seems to be the most straight forward solution.

src/track/beats.cpp
```diff
@@ -115,8 +115,8 @@ Beats::ConstIterator Beats::ConstIterator::operator-=(Beats::ConstIterator::diff
     // Detect integer overflow
     const int minBeatOffset = std::numeric_limits<Beats::ConstIterator::difference_type>::lowest();
     if (m_beatOffset < minBeatOffset + n) {
-        qWarning() << "Beats: Iterator would go out of possible range, capping "
-                      "at earliest possible position.";
+        qDebug() << "Beats Iterator" << m_beatOffset << "-" << n
```
Holzhaus, line 118: What's the rationale for lowering the log level here?

ywwg (approved):

JoergAtGithub:
Test it and no warnings/info appeared. Code LGTM! Thank you!

daschuer:
I did not test it. But I cannot explain the issue with saving a beatgrid from this issue.

Issue #12316, opened 2023-11-19
Precompiled heades (PCH) should be opt-in and not opt-out


uklotzde:
### Bug Description

Pre-compiled headers seem to cause more issues than benefits.

Requiring to explicitly opt-out by specifying `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON` is not intuitive.

daschuer:
Please describe your issues. 
Is this a duplicate of https://github.com/mixxxdj/mixxx/issues/12081?

Issue #12081, opened 2023-10-08
ccache is bypassed after switching branches 


daschuer:
### Bug Description

We have originally introduced ccache to be able to switch between feature branches and build don't require a long build time if the branch has been build before or there are only a few files changed. This does not longer work with recompiled headers. Because the precompiled header is a giant file that is included in every source file. A single change in a header requires that all files are rebuild. Even after applying the sloppiness setting https://github.com/mixxxdj/mixxx/pull/12080 I was not able to get a reasonable number chache hits after switching a branch. There seems to be an issue that a new generated pre compiled header dos not produce the  same checksum like a previous one from the same headers. The new precompiled header probably flushes the cache, because every object using it is considered as new.  

Precompiled headers should benefit significantly clean builds, but this is not really the case with gcc. Here are my results using the 2.4 branch. 

Clean build with PCH:    user 43m43,093s
Clean build without PCH: user 58m50,340s
Clean build with PCH + ccache: use 46m26,845s cache hits = 63 
Clean build with PCH + ccache + sloppy: user 54m53,676s cache hits = 63
Clean build with PCH + ccache + sloppy: user   4m3,951s cache hits = 679

Conclusion, caching works with the sloppy settings, but is not helpful in my use case, because it requires a magnitude larger cache which I cannot effort. 

It looks like pch+ccache is working, but not reasonable. 
What should be our default? 30 % faster clean build, vs long wait time when testing different branches. 
What do you think? 

### Version

2.4

### OS

Ubunti 20.4 Focal

Holzhaus:
I just use `-DCMAKE_DISABLE_PRECOMPILE_HEADERS=ON`.

daschuer:
Yes that works. This bug is about the default configuration. I can imagine that even casual contributor suffer longer bui

Did you consider to put only system headers to the PCH?

ronso0:
I suffer from this as well.
I change certain build options or have to set them after wiping /build, e.g. QT6, Release for production builds, CMAKE_EXPORT_COMPILE_COMMANDS for Eclipse, OPTIMIZE:Native.
I use CMakeGUI for that, which also easily allows to select a certain g++ version in case I experience issues.
Though, `CMAKE_DISABLE_PRECOMPILE_HEADERS` is not listed so I need to configure manually via command line.

Setting CMAKE_DISABLE_PRECOMPILE_HEADERS in our Cmakelist would certainly help.
[...]

PR #12371, opened 2023-11-28
Library pch

daschuer:
This branch replaces the internal headers which are subject of changes with the headers in angle brackets, which are treated as read only. 

This was an experiment to check if we can make some benefits of such a pch configuration during development. 

Single thread: 
Clean build with PCH: user 40m33,092s
Clean build without PCH: user 51m34,550s
Clean build with PCH this branch: user 44m12,263s

So it still has some benefits, but not close the original solution. 
Compared to the maintenance effort not a good benefit.

I think we should stick with the shortest compile time for release and disable pch for debug. 

Related issues:
https://github.com/mixxxdj/mixxx/issues/12081
https://github.com/mixxxdj/mixxx/issues/12316


daschuer:
The times on GitHub confirm this regression on windows.

daschuer:
It turns out this has no benefit and we can close this PR.

PR #12372, opened 2023-11-29
add hotcue shift controls

### LINKED ISSUE #12367: Feature Request: Add a quick way to nudge a hotcue earlier or later / left and right
### Feature Description

While creating hotcues I frequently have to adjust the hotcue to come in a bit earlier or later than what I achieve when I tap one of the hotcue pads on a hardware controller or in the mix UI while playing.

To correct the "bad" hotcue I usually trigger the bad hotcue, then delete the bad hotcue, move the track a bit to the left or right and recreate the good hotcue.

I see 2 ways of correcting the hotcues more conveniently:

 * A very natural feeling for correcting it would be to press and hold a hotcue button, sliding the song pointer left and right to its desired position and when releasing the hotcue button the cuepoint is finally moved. To listen and decide where the hotcue should go you would basically "scratch" the point around the cue point until you hear where it should be. Pretty much the vinyl style of finding a cue point.

 * An other way of doing it would be to have "nudge buttons" that nudge the most recently used hotcue left and right. Listening is then done by repeatedly pushing the hotcue. This would be more the Digital-DJ way of finding a good point for the hotcue.

Is this something that should be implemented controller specific? Or should mixxx's interface also include these controls for mouse and keyboard use? Is there something I have missed that does this already?

(btw: I'm not sure if audio latency is often the cause for "too late hotcues" or this is simply me being too slow or too fast ;)

ronso0:
This PR adds
* `hotcue_N_shift_earlier` and `hotcue_N_shift_later`, incl. entry in the control picker menu
(not supposed to be used in skins, hence no tooltip)
* respective buttons to the hotcue menu popup
* ControlEncoder `hotcue_N_shift` for easy encoder mapping
* ControlEncoder `group,shift_focused_hotcue` to allow cramming it into dense mappings¹

The controls shift the cue by 10ms or, if `quantize` is enabled, to the previous / next beat. It is not checked whether the cue is already quantized.

image

**ToDo**
- [ ] style WCueMenuPopup controls in all skins

¹I considered creating a control that can address ANY cue (main cue, hotcues, intro, outro) by looking up the cue closest to the play position, but it requires more complex (debatable) code, and even though it may be useful I think it's not required since
1. the clear/set controls for the main cue, intro and outro are a bit more accessible than those for the hotcues (in the GUI)
2. I think there's more demand during performance being able to re-set hotcues than the intro/outro.

Fixes both #10812 (nudge) and #12367 (shift by one beat)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1172,6 +1172,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/widget/findonwebmenuservices/findonwebmenusoundcloud.cpp
   src/widget/hexspinbox.cpp
   src/widget/paintable.cpp
+  src/widget/rightclickpushbutton.cpp
   src/widget/wanalysislibrarytableview.cpp
   src/widget/wbasewidget.cpp
   src/widget/wbattery.cpp
--- a/res/skins/Deere/style.qss
+++ b/res/skins/Deere/style.qss
@@ -2320,20 +2320,37 @@ WEffectChainPresetSelector::indicator:unchecked:selected,
   selection-background-color: #d2d2d2;
 }
 
-#CueDeleteButton {
-/* tall button, about the same height as cue number + label edit box */
-  width: 28px;
-  height: 46px;
-  /* make the icon slightly larger than default 16px */
-  qproperty-iconSize: 20px;
-  qproperty-icon: url(skin:/../Deere/icon/ic_delete.svg);
+WCueMenuPopup #CueDeleteButton,
+WCueMenuPopup #CueShiftEarlier,
+WCueMenuPopup #CueShiftLater {
   background-color: #3B3B3B;
   border-radius: 2px;
   outline: none;
 }
-
-#CueDeleteButton:hover {
+WCueMenuPopup #CueDeleteButton:hover,
+WCueMenuPopup #CueShiftEarlier:hover,
+WCueMenuPopup #CueShiftLater:hover {
   background-color: #4B4B4B;
+}
+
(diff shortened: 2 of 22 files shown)


mxmilkiib:
This is only partly implemented, yes?

Latenight has the skin settings option, and that toggles the UI elements.

Deere (64 Samplers) has the skin settings option, but nothing happens on toggling it.

And Deere doesn't have that skin setting option.

Running 2.4-alpha-1578-gc47f67985d.

ronso0:
IIUC you refer to the buttons that shuft **all** cues at once. This PR add shift controls for individual hotcues.

I'll check what's wrong with LateNight64.

ronso0:
> Latenight has the skin settings option, and that toggles the UI elements.
> 
> Deere (64 Samplers) has the skin settings option, but nothing happens on toggling it.
> 
> And Deere doesn't have that skin setting option.

In the PR that introduced these controls they [were shown uncondiotionally](https://github.com/mixxxdj/mixxx/commit/aea7693f8f8397afe44737eb973e954633a4a3a1) (based on the assumption "The option is overcomplicated and would probably confuse more people who don't need the cue shift buttons than ignoring buttons they don't need."

Later on, during the polish / redesign processes, we decided to re-add the skin option (one reason to make it optional in Deere is the limited horizontal space). Obviously that fix was incomplete.
I just pushed the fix to 2.4

ronso0:
I'm not motivated to finish this, will close now.

We may address this in the QML redesign with a new hotcue menu that holds a cue position preview (waveform window) and WYSIWYG or dedicated shift controls.

Issue #12337, opened 2023-11-22
Add a beats_translate_move? Like beats_translate_earlier and beats_translate_later, but can be used with a single CC SelectKnob style bind


mxmilkiib:
### Feature Description

My Akai MPD218 rotary encoders give out "ctrl 0x0E, val 0x01" when turned clockwise and "ctrl 0x0E, val 0x7F" when turned anti-clockwise.

Having looked at [the code](https://github.com/mixxxdj/mixxx/blob/c1665e50e9045cfe8fce70308d78fb002c5e41e5/src/controllers/learningutils.cpp#L191), this is two's complement, and is that the SelectKnob option is for, AFAIU.

But to move the beatgrid, there is only beats_translate_earlier and beats_translate_later, and it is required that these are bound to two separate/different CCs, so only one direction of movement can assigned to a rotary encoder as it stands.

If there were a generic beats_translate_move that could act upon a SelectKnob style input, that would be very handy.

(This relates to #7557 in that the two's complement SelectKnob system is using different values of the same CC, so this is arguably somewhat a subset of that issue)

Holzhaus:
I think this is quite a specific case. Maybe this would better be mapped in JS?

mxmilkiib:
I think it's arbitrary to draw the line there, and adds a hurdle of complexity to the general user.

I mean, moving the grid left and right with a pot? That's imho really not that far out!!

daschuer:
Finally "source code wins". Since we already have a bunch of beats_translate_.... controls on more does not make a big difference. I would not object, if one contributes such a solution.

mxmilkiib:
Would it be something like [this](https://github.com/mxmilkiib/mixxx/blob/beats_translate_move/src/engine/controls/bpmcontrol.cpp#L238)?

ronso0:
Yes, https://github.com/mxmilkiib/mixxx/commit/d4f5e9048685fd2ad41a64fb5fa0dbd58ab78d0e would work.
But, to reduce code duplication in `slotTranslateBeatsEarlier` and `slotTranslateBeatsLater`, we can make them call `slotTranslateBeats` with the argument (+-) and do the sample / beats offset calculation there.

daschuer:
Do we have a control type for it? It looks a bit hack to have the midi key value in our engine code.

daschuer:
I think it is "ControlEncoder"

ronso0:
[...]

PR #12373, opened 2023-11-29
controllers: add helper to control "waveform_zoom" with potmeters

ronso0:
This came up in the forums https://mixxx.discourse.group/t/waveform-zoom-midi-control-on-a-cc-dial/28293

* first comit is just a cleanup after I failed to find "behaviour"
* second commit is the script helper: default is a continuous range, set 'round' to `true` to get discrete integer steps
* third commit fixes some mappings using the `waveform_zoom` with the ancient range 0..6

**TODO**
- [ ] add note to [[ChannelN],waveform_zoom](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-waveform_zoom), improve Feedback note
- [ ] remove 'zoom' from ControlPotmeter controls examples

### DIFF
--- a/res/controllers/KANE_QuNeo_scripts.js
+++ b/res/controllers/KANE_QuNeo_scripts.js
@@ -888,9 +888,8 @@ KANE_QuNeo.closeSliderMode = function () {
 KANE_QuNeo.deckZoom = function (deck, value) {
     var channel = deck - 1; // track channels start at 0 to properly reference arrays
     var channelName = KANE_QuNeo.getChannelName(deck)
-    var normalized = Math.ceil(6 * ((127 - value) / 127))
     // adjust zoom
-    engine.setValue(channelName, "waveform_zoom", normalized)
+    script.waveformZoom(group, value, true);
 }
 
 KANE_QuNeo.deckCursor = function (deck, value) {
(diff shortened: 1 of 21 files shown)


mxmilkiib:
Very nice, thank you, but, ultimately, why not push it down to the C++ level? That way it could be bound to a control via the GUI/wizard.

Maybe I'm wrong, but I'm starting to get the feeling there's a "standard library" of these kinds of level of functionality that aren't bindable via the UI? Which as I wordly put on #12337, is an abstraction that's beyond the scope of most users.

daschuer:
Yes, can't we make it a ControlPotmeter?

ronso0:
I've tried that, or rather I tried to make a helper pot.

The issue with `waveform_zoom` is that the user can set the default value at runtime via the waveform preferences, hence we can of course create a ControlPotmeter (and set a step size of 1 so the GUI zoom buttons still work as before) but as soon as the default is changed the _set_default CO would not be working as expected anymore.
I also looked into creating a ControlPotmeterSimple that does not create all sub-controls (only _up/_down) and keep the `_set_default` pushbutton separate, but that that was too much just for one control.

We could also connect the preferences combobox to a separate ControlObject `waveform_zoom_set_default_value` which then triggers a slot in BaseTrackPlayer to `zoomCO->setDefault(newDefaultvalue)`. I didn't check if that is even possible after the control is created and that also felt like a bit too much work.
I can check that again though.

One thing to note: if you want individual zoom levels per deck (to look ahead or tweak cues on a deck while having the overview on another) the current stepped behaviour is good since it is very easy to manually re-sync the zoom levels of multiple decks. With a continuous range that's almost impossible.

ronso0:
> We could also connect the preferences combobox to a separate ControlObject waveform_zoom_set_default_value which then triggers a slot in BaseTrackPlayer to zoomCO->setDefault(newDefaultvalue). I didn't check if that is even possible after the control is created and that also felt like a bit too much work.

Okay, that was easier than expected.

Replaced by #12387

PR #12375, opened 2023-11-30
OS macros

daschuer:
This makes them independent from Qt includes.

I have another branch in preparation where the build checks are build without pch. 
This failed due to a related issue.


JoergAtGithub:
LGTM! Thank you!

Issue #11630, opened 2023-06-05
bpm lock icon rendered incorrectly on QT6 (no QML)


ywwg:
### Bug Description

The bpm column does not look correct on QT6:

![image](https://github.com/mixxxdj/mixxx/assets/888940/9466c250-4e70-4cf8-89f6-a4aefc4205cc)

The custom lock icon we use is not being drawn.

### Version

HEAD

### OS

Ubuntu 22.04

ronso0:
Maybe just an error in the SVG? How does it look like in PaleMoon?

`/res/skins/LateNight/classic/buttons/btn__lib_bpm_locked_orange.svg` is not optimized / cleaned up. Can you try saving it as 'Optimized SVG' with default settings (which should remove unused tags and collapse groups)?

daschuer:
I can confirm the issue with all skins. However only if the row is not selected. In selected case everything looks fine.

ronso0:
Can you try the explicit `!selected` selector? LateNight Classic:
```
#LibraryBPMButton::indicator:!selected:checked,
#LibraryBPMButton::indicator:selected:checked {
  image: url(skin:/classic/buttons/btn__lib_bpm_locked_orange.svg);
}
#LibraryBPMButton::indicator:!selected:unchecked,
#LibraryBPMButton::indicator:selected:unchecked {
  image: url(skin:/classic/buttons/btn__lib_bpm_unlocked_grey.svg);
}
```

ywwg:
no, that does not work for me. weird.

ronso0:
So the `WTrackTableView::indicator` (played column) rules (applied to all QCheckBox indicators) are not overwritten by the `#LibraryBPMButton::indicator` rules.

https://github.com/mixxxdj/mixxx/blob/d763be606148a1047522b21a8840d50d882cf1c2/src/library/bpmdelegate.cpp#L63-L94

Can you try some other explicit selector, for example `WTrackTableView QCheckBox::indicator`?

ronso0:
Oh, I think I  got it (but it's surprising this worked before...):
the LibraryBPMButton rule
https://github.com/mixxxdj/mixxx/blob/d763be606148a1047522b21a8840d50d882cf1c2/res/skins/LateNight/style_classic.qss#L2588-L2595
is **after** the general indicator rule
[...]

PR #12380, opened 2023-11-30
Fix track color background with Qt6

### LINKED ISSUE #12355: regression in main, qt6: library track colors no longer appear
### Bug Description

Track colors have disappeared from the track listing in the main branch. I did a git bisect and the commit that broke it was the QT6 checkbox fix: 1ef4e4d6216fafab3b2393cf96942ec23559df60

@fwcd can you take a look?

### Version

HEAD

### OS

Ubuntu 22.04

ronso0:
Fixes #12355 for me, both with Qt5 and Qt6 (note I didn't see #11630).

So, the issue is that https://github.com/mixxxdj/mixxx/commit/1ef4e4d6216fafab3b2393cf96942ec23559df60
```css
/* Prevent OS-style checkbox from being rendered underneath the custom style. */
WTrackTableView::item,
#LibraryBPMButton::item {
  border: 0px;
}
```
clears the track color set via _BackgroundRole_.

Seems it sufficient to call [`TableItemDelegate::paintItemBackground`](https://github.com/mixxxdj/mixxx/blob/2c2e706b4419794e1b29f23f124ce65f49b88bc7/src/library/tableitemdelegate.cpp#L58) like it's done for the other delegates.

I don't know how to apply this to the Played checkbox (it's a native QTableView::indicator), so I added a delegate for that, too. Dumb c/p from BPM delegate, without the editor.

Please check if that works also on Windows and macOS @fwcd @JoergAtGithub 
Should look good with all skins.


fwcd:
Looks good to me

ronso0:
Can someone tell if the last commits makes sense?
I just feeled weird that every delegate includes QTableView etc.

daschuer:
The last commit has some implications regarding compile time. It includes unnecessarily headers into all cpp files that are including  "src/library/tableitemdelegate.h"
The tool https://github.com/include-what-you-use/include-what-you-use points that out suggesting to remove all headers that are not used in the including files itself.

ronso0:
Okay, thanks. I reverted that header commit.

Ready to roll if it's been confirmed to fix the issue on Windows, too.

JoergAtGithub:
I tested on Win11. The track color, checkbbox and padlock symbol appear as expected on all skins!

But there seems to be a missing refresh/update, after the color unassign operation. The update happens delayed, when I select another row:
https://github.com/mixxxdj/mixxx/assets/64457745/d4a3d7de-78ed-4153-844f-c71a509a366c

ronso0:
Thanks for testing.

The refresh issue occurs also in 2.4 IIRC, and it affects the entire row. This PR only touches the BPM delegate, and the new one, of course.

PR #12370, opened 2023-11-28
vcpkg update 2.4

daschuer:
This updates to the latest vcpkg environment. 
This uses the new release prefix for the release only triplets like upstream.

PR #12382, opened 2023-12-01
macOS buildenv: Improve error messages for CI-only wrappers

fwcd:
Since these scripts currently rely on being invoked and cannot be sourced (due to writing to `GITHUB_ENV`, see https://github.com/mixxxdj/mixxx/pull/12370#issuecomment-1834435406), this PR improves the error message by clarifying that the scripts are intended for CI use only. The message further directs developers interesting in setting up a local environment to `tools/macos_buildenv.sh`.

Previously, executing the wrappers in a non-CI-environment would result in an error message in `macos_buildenv` that the latter script has to be sourced, which may be a bit confusing to newcomers.

In the future we may investigate making these scripts "sourcable" too (though there are some subtleties involved, e.g. accidentally "leaking" `BUILDENV_RELEASE` and similar variables into the parent shell).

### DIFF
--- a/tools/macos_arm64-cross-release_buildenv.sh
+++ b/tools/macos_arm64-cross-release_buildenv.sh
@@ -1,3 +1,9 @@
 #!/bin/bash
+
+if [ -z "${GITHUB_ENV}" ]; then
+  echo "This script is currently intended for CI use only (source tools/macos_buildenv.sh instead)"
+  exit 1
+fi
+
 tools_path=$(dirname "$0")
 BUILDENV_ARM64=TRUE BUILDENV_RELEASE=TRUE "${tools_path}/macos_buildenv.sh" "$@"
--- a/tools/macos_arm64-cross_buildenv.sh
+++ b/tools/macos_arm64-cross_buildenv.sh
@@ -1,3 +1,9 @@
 #!/bin/bash
+
+if [ -z "${GITHUB_ENV}" ]; then
+  echo "This script is currently intended for CI use only (source tools/macos_buildenv.sh instead)"
+  exit 1
+fi
+
 tools_path=$(dirname "$0")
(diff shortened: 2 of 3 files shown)


daschuer:
I have not understand the issue around "sourcable". Can you explain what happens if you try to use these scripts?

fwcd:
Outside of CI, these scripts have to be invoked with `source <path to script>` to export any variables, the fact that we can run them in CI is solely thanks to the fact that we can export variables by appending them to `$GITHUB_ENV`. These scripts, however, are not designed to be sourced:

https://github.com/mixxxdj/mixxx/blob/0f0deb18160547bcd96e0a60d5fc7c7a9d4ba306/tools/macos_release_buildenv.sh#L3

Sourcing `macos_release_buildenv.sh` will still try to execute `macos_buildenv.sh` and thus error in

```
This script must be run by sourcing it:
source tools/macos_buildenv.sh
```

This is pretty confusing for anyone trying to set up a local development environment that isn't already familiar with these scripts. Therefore this PR proposes to emit a better error message:

```
This script is currently intended for CI use only (source tools/macos_buildenv.sh instead)
```

Of course, the best solution would be if the scripts could work across CI and local environments. This would currently entail making the scripts both sourcable and executable. I have tried to make this work in 9e65c7ab6bc6f01a93a683111ccb4fc0d0ee8152, but there are some drawbacks to this that I wouldn't be comfortable merging right now. For example, sourcing these wrapper scripts will currently leak the `BUILDENV_*` variables into the parent shell. This could then e.g. result in (unintentionally) cross-building if the `macos_arm64-cross_buildenv.sh` script is sourced before running another buildenv script.

fwcd:
Any blockers here? I think we should either add an error or make the scripts "sourcable", since the current situation is confusing to anyone wishing to develop Mixxx locally and stumbling upon these scripts.

daschuer (approved):
Thank you for explanations. I agree that it is better to fail early.

PR #12383, opened 2023-12-01
feat(debug): make DBG macro usable within expressions

Swiftb0y:
The DBG macro now accuraletely models the rust dbg! macro by returning the argument its been given verbatim, making it possible to embed in the middle of expressions.

Here's a godbolt if you want to play around with it outside of mixxx
https://compiler-explorer.com/z/Pjf3z3bEx

I also removed some unnecessary `<<` for strings that can be concatenated automatically at compile time.

### DIFF
--- a/src/util/debug.h
+++ b/src/util/debug.h
@@ -6,8 +6,11 @@
 #include "errordialoghandler.h"
 
 #ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
-#define DBG(value) \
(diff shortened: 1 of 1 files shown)


src/util/debug.h
```diff
@@ -5,17 +5,22 @@
 
 #include "errordialoghandler.h"
 
+// this might break when the deduced return type is not a value type,
+// but this is fine for now.
```
daschuer, line 9: I have no idea how this can happen. Can we remove the comment?

daschuer:
Thank you. This works nice.

daschuer (approved):
LGTM, thank you.

PR #12385, opened 2023-12-01
Track menu: enable Lock BPM action if any selected track BPM is unlocked

ronso0:
**BPM** > **Lock** was not enabled if any of the selected track is locked.


ronso0:
There's probably a smarter solution, but it does the job :shrug:

JoergAtGithub (changes requested):
Except the code duplication issue, it's fine for me.

JoergAtGithub (approved):
Code LGTM now! Works as expected on local Windows11 build. Thank you!

PR #12386, opened 2023-12-02
Tracks table: Ctrl+Space toggles BPM & Played checkboxes

ronso0:
I tried `Space` intuitively a few times but  that maxmizes the library, of course.

Usually `Space` toggles checkboxes in dialogs, and `Ctrl`+`Space` toggles combobox views, so IMO it make sense to use `Ctrl`+`Space` for checkboxes in the table view, somehow fits the `Ctrl` cursor move keys.

### DIFF
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -840,6 +840,20 @@ void WTrackTableView::keyPressEvent(QKeyEvent* event) {
             return;
         }
     } break;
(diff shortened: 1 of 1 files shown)


ronso0:
Now it's working without this branch :shrug:
Will checkif I confused it with the Qt6 build.

ronso0:
Yes, this now works out of the box.

Would be cool if Ctrl + Space would also toggle the Preview button.

PR #12388, opened 2023-12-02
Sync Lock: protect against syncing phase to a stopped leader

### LINKED ISSUE #11788: Pitch of deck 1 is changing when skipping on paused deck 2
### Bug Description

1. Load track 1 to deck 1
2. Play deck 1
3. Enable Sync for deck 1
4. Enable Quantize for deck 1
5. Load track 2 to deck 2
6. Enable Sync Leader for deck 2
7. Keep deck 2 paused
8. Click with mouse at deck 2 on waveform that paused track 2 will get a different position
9. Now playing deck 1 will get a sound pitch
10. Repeat point 8 and 9

This seems to be only when sync on both decks is enabled, sync leader is deck 2 and deck 1 is quantized

Please let me know when I should do some other tests.

### Version

2.4.0-beta

### OS

Ubuntu 22.04.2

ywwg:
Fixes #11788

Propagate playing/audible state to the bpmcontrol. If a Leader is not playing, we should not use it for sync adjustment or phase seeking, because that causes audible artifacts.

Moreover, we surrender an explicit leader if it is stopped. In the past we had thought of explicit leader as a persistent user choice, but in reality it is more likely to be a temporary setting which the user would like us to fix when it stops being relevant. There are lots of issues with having a stopped deck as leader, so this seems like the right choice.

### DIFF
--- a/src/engine/bufferscalers/enginebufferscale.h
+++ b/src/engine/bufferscalers/enginebufferscale.h
@@ -69,6 +69,7 @@ class EngineBufferScale : public QObject {
     virtual void onSampleRateChanged() = 0;
 
   protected:
+    FRIEND_TEST(EngineSyncTest, BpmAdjustFactorStoppedLeader);
     double m_dBaseRate;
     bool m_bSpeedAffectsPitch;
     double m_dTempoRatio;
--- a/src/engine/bufferscalers/enginebufferscale.h
+++ b/src/engine/bufferscalers/enginebufferscale.h
@@ -69,7 +69,6 @@ class EngineBufferScale : public QObject {
     virtual void onSampleRateChanged() = 0;
 
   protected:
-    FRIEND_TEST(EngineSyncTest, BpmAdjustFactorStoppedLeader);
     double m_dBaseRate;
     bool m_bSpeedAffectsPitch;
     double m_dTempoRatio;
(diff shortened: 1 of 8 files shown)


ywwg:
ok this *only* surrenders leader if an explicit leader stops or becomes inaudible.  This may be good enough to be the easy fix

daschuer:
This unfortunately fixes the issue only halfway. The explicit leader is removed form the deck after stopped which is OK but kind of surprising because it was "explicit". If you than make it explicit again, we are back with the old issue. 

Can we just not allow to make s stopped deck explicit master? 
But what is than the difference between explicit and implicit master? 

Do we still have a valid use case for an explicit master? 
Or can we dispose it altogether?

daschuer:
> There is still a use for explicit leader when two decks with non-const beatgrids are playing and the user would prefer one to be leader over the other.

Yes, a very valid one. This is however also covered with the implicit leader (two clicks) and comes with the extra safety of removing the leader automatically whenever reasonable. The crucial question is if there is still a use case for a sticky explicit leader. 

I have only the scratching feature in mind we have already disabled. 

As as side issue we have the missing option explicit select the internal clock. 

So how about this: 
* Click on flower crown of a playing deck: implicit leader 
* Click on implicit leader crown: follower, internal clock becomes leader 
* Click on flower crown stopped deck: ignore

This way the user has all control when two decks playing in sync and Mixxx takes over once only one is playing. 

The explicit leader can be reintroduced for scratching later.

What do you think?

ywwg:
I think we can continue to work on the UX for leader selection, but that it's too late for 2.4 to try out these options.  Instead I filed https://github.com/mixxxdj/mixxx/pull/12431 which simply removes Explicit Leader for now. Not much code changes and it will be trivial to revert the patch and try some of these other approaches.

I think double-deck-scratching is clever but I am very skeptical of its actual usefulness, and I bet if the beat grids are non-constant it will quickly fall apart. I do agree we need a way to explicitly engage internal clock. That too will have to wait I think.

PR #12387, opened 2023-12-02
turn `waveform_zoom` into ControlPotmeter

ronso0:
c++ alternative to  #12373 

* `waveform_zoom` is now a ControlPotmeter incl. `_up`, `_down`, `_set_default` and some rather useless controls
step size is 1, i.e. `_up`/`_down` act exactly as before
* ~~added `waveform_zoom_selector`, triggers `_up`/`_down` and thereby simplifies mapping rotary encoders with the `SelectKnob` MIDI option~~ already possible with the e.g. the `Rot64` MIDI option
* ~~fixed some hard-coded ranges in mappings~~ #12393 

**TODO**
- [x] update the manual https://github.com/mixxxdj/manual/pull/722

Note to myself:
pick https://github.com/mixxxdj/mixxx/pull/12373/commits/55418d00da24e01077e2bd8d7cedad561a672979 from #12373

PR #12393, opened 2023-12-04
controllers: fix `waveform_zoom` ranges

ronso0:
picked from #12387

updates some old mappings to use the current range 1..10


JoergAtGithub:
Ok, let's merge it as stop-gap fix.

PR #12397, opened 2023-12-05
CI test: type declaration

ronso0:
This is just a test.

One of the initial commit in #12376 made via Github web UI was missing the `ControlEncoder` type delaration (no header, no forward-declaration), yet CI didn't complain (it could be built on all targets). Locally it failed with gc 13 / mold.

Let's see if this is reproducible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The Mixxx team uses [Github Issues][issues] to manage Mixxx development.
 Have a bug or feature request? [File a bug on Github][fileabug].
 
 Want to get involved in Mixxx development? Assign yourself a bug from the [easy
-bug list][easybugs] and get started!   
+bug list][easybugs] and get started!
 Read [CONTRIBUTING](CONTRIBUTING.md) for more information.
 
 ## Building Mixxx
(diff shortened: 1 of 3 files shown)


ronso0:
Hmmkay, it succeeded.
Pushed a more complete commit with a slot conected to see if it even works.

Swiftb0y:
Weird. With `gcc -E` You can tell gcc to emit the file after preprocessing. For me, It contains no `ControlEncoder`, so I don't think its a transitive-dependency-issue (at least not without extra preprocessor defines, the switches on CI might make it pass).

ronso0:
Okay, so the complete example was built and, with the deb installed in a VM, I can also use `beat_translate_move` without issues.

"Verrry, verrry strange!"

daschuer:
We have precompiled headers active on our CI. This means almost all headers are included everywhere. 
I am still  in the process to clean that up. Once we have all test data, we can decide how to proceed with pch.

daschuer:
@ronso0: Is this one still relevant?

ronso0:
Well, the last Windows build succeeded.
I'll rebase to trigger builds.
If at least one build runner fails errors can be detected and we can close this IMO. I'm not into pch and I don't plan to work on it.

PR #12400, opened 2023-12-05
Feat: eslint plugin diff

Swiftb0y:
This will make it less scaring to make changes to outdated mappings that are not style-conform.
The motivation is basically the same as for the diff-based clang-format tool we use.


ronso0:
Very much appreciated, thank you!

Just merge and try, or does anyone object?

daschuer:
no objections, Thank you.

PR #12396, opened 2023-12-05
controllers: fix engine.beginTimer() syntax

ronso0:
Reason to start this was #12384 

I know this is tedious to review. 35 mappings scripts.
These are not all mappings that are broken. A few other mappings seem to have other syntax issues, see https://github.com/mixxxdj/mixxx/blob/9eda4c399ca2f3bddd3fe602703285aac8252eeb/res/controllers/Numark-DJ2Go-scripts.js#L133 and I I have no clue how to fix those with the find/replace pattern I used.

PR #12401, opened 2023-12-05
Fix/gh12369 beginTimer callback syntax

### LINKED ISSUE #12369: Numark MixTrack Pro 3 errors on startup
### Bug Description

Every time I start Mixxx, I get 2 errors:
```
Uncaught exception: file:///usr/share/mixxx/controllers/Numark-Mixtrack-3-scripts.js:375: TypeError: Property 'flashOnceOff' of object [object Object] is not a function
Backtrace: @file:///usr/share/mixxx/controllers/Numark-Mixtrack-3-scripts.js:375
```
```
Uncaught exception: file:///usr/share/mixxx/controllers/Numark-Mixtrack-3-scripts.js:316: TypeError: Property 'flashOnceOn' of object [object Object] is not a function
Backtrace: @file:///usr/share/mixxx/controllers/Numark-Mixtrack-3-scripts.js:316
```
It seems to mostly work fine if I just ignore these errors however.

### Version

2.4.0-beta

### OS

Fedora 40 (rawhide)

### LINKED ISSUE #12384: DDJ-SB3 not woring with Mixxx 2.4.0-beta
### Bug Description

After upgrading to the "2.4.0-beta" version (from "2.3.6 (HEAD)"), Mixxx raises an error regarding my DDJ-SB3 controller: 
```
The mapping for your controller "DDJ-SB3 MIDI 1" is not working properly.
The script code needs to be fixed.
The functionality provided by this controller mapping will be disabled until the issue has been resolved.

Details:
Uncaught exception: file:///usr/share/mixxx/controllers/Pioneer-DDJ-SB3-scripts.js:280: Error: Invalid timer callback provided to engine.beginTimer. Valid callbacks are strings and functions. Make sure that your code contains no syntax errors.
Backtrace: @file:///usr/share/mixxx/controllers/Pioneer-DDJ-SB3-scripts.js:280
```

The line in question:
```
    if (PioneerDDJSB3.twinkleVumeterAutodjOn) {
        PioneerDDJSB3.vuMeterTimer = engine.beginTimer(100, PioneerDDJSB3.vuMeterTwinkle());
    }
```

I've checked: this file did not change between "2.3.6 (HEAD)" and "2.4.0-beta"...

### Version

Mixxx 2.4.0-beta

### OS

Ubuntu 22.04.3 LTS

Swiftb0y:
Based on #12400. Alternative to #12396. Fixes #12384. Fixes #12369 apart from https://github.com/mixxxdj/mixxx/issues/12369#issuecomment-1835700452.

@ywwg can you take a quick look at the changes to the Reloop-Beatpad script? I did some refactorings that seemed sensible, but you probably know the code better than me.

I'd highly appreciate if anyone that still has access to any of these controllers can give these changes a quick test and see if they broke anything in their respective mappings.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,6 +1,7 @@
 {
   "extends": [ "eslint:recommended",
-               "plugin:jsdoc/recommended"],
+               "plugin:jsdoc/recommended",
+               "plugin:diff/diff"],
 
   "parser": "@typescript-eslint/parser",
   "parserOptions": {
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -79,6 +79,7 @@ repos:
           - eslint-plugin-jsdoc@^v46.5.1
           - "@typescript-eslint/eslint-plugin"
           - "@typescript-eslint/parser"
+          - eslint-plugin-diff@^2.0.3
   - repo: local
     hooks:
       - id: clang-format
--- a/res/controllers/Akai-LPD8-RK-scripts.js
+++ b/res/controllers/Akai-LPD8-RK-scripts.js
@@ -253,7 +253,7 @@ LPD8RK.init = function (id, debug) { // called when the device is opened & set u
     engine.softTakeover("[Sampler2]","rate",true);
 
     //set LED timer
-    LPD8RK.ledTimer = engine.beginTimer(LPD8RK.LEDinterval, "LPD8RK.setLeds()");
+    LPD8RK.ledTimer = engine.beginTimer(LPD8RK.LEDinterval, LPD8RK.setLeds);
 };
 
 LPD8RK.shutdown = function () {
@@ -263,7 +263,7 @@ LPD8RK.shutdown = function () {
 LPD8RK.resetLEDTimer = function () {
     engine.stopTimer(LPD8RK.ledTimer);
     LPD8RK.setLeds()
-    LPD8RK.ledTimer = engine.beginTimer(LPD8RK.LEDinterval, "LPD8RK.setLeds()");
+    LPD8RK.ledTimer = engine.beginTimer(LPD8RK.LEDinterval, LPD8RK.setLeds);
 };
 
 LPD8RK.setLeds = function () {
@@ -376,7 +376,7 @@ LPD8RK.reloopButton = function (channel, control, value, status, group) {
         engine.stopTimer(LPD8RK.reloopTimer);
         LPD8RK.loopbuttonDown=true;
         LPD8RK.doreloop=true;
-        LPD8RK.reloopTimer = engine.beginTimer(500, "LPD8RK.disablereloop()", true);
+        LPD8RK.reloopTimer = engine.beginTimer(500, LPD8RK.disablereloop, true);
         } else {//button was released
         LPD8RK.loopbuttonDown=false;
         if (LPD8RK.doreloop===true) {engine.setValue(group, "reloop_exit", 1);};
@@ -437,11 +437,11 @@ LPD8RK.hotcueBankDial = function (channel, control, value, status, group) {
     LPD8RK.oldHotcueBank=LPD8RK.hotcueBank;
     //set timer to clear old bank number after 500 msec, so bank indicator light will light up
     engine.stopTimer(LPD8RK.oldbanktimer);
-    LPD8RK.oldbanktimer = engine.beginTimer(500, "LPD8RK.resetOldBank()", true);
+    LPD8RK.oldbanktimer = engine.beginTimer(500, LPD8RK.resetOldBank, true);
 
     //set timer to restart LED updates in 500 msec
     engine.stopTimer(LPD8RK.LEDPauseTimer);
-    LPD8RK.LEDPauseTimer = engine.beginTimer(LPD8RK.LEDinterval, "LPD8RK.resetLEDTimer()", true);
+    LPD8RK.LEDPauseTimer = engine.beginTimer(LPD8RK.LEDinterval, LPD8RK.resetLEDTimer, true);
     };
 
 LPD8RK.looplenDial = function (channel, control, value, status, group) {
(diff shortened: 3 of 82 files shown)


Swiftb0y:
yikes, still suffering from eslint issues, though this time they're valid and only related to the actually changed lines. I think we should ignore those for now too since some of them just require too many changes (eg renaming an entire controller because its named `DeFOrmEd_SnakE_CAsE`)...

Swiftb0y:
I wouldn't mind but I don't have any. I went through it manually. My initial regex attempts failed, this was faster and gave me more confidence that I'm not accidentally breaking anything.

ywwg:
> @ywwg can you take a quick look at the changes to the Reloop-Beatpad script? I did some refactorings that seemed sensible, but you probably know the code better than me.

hmm I don't actually know that code.  I used to maintain the Hercules DJControls so I know those better and those look fine.

JoergAtGithub (changes requested):
The code change itself is as far as I can see correct ES7 code, that should work.
But unfortunatly, we are not only in the JavaScript domain here. We evaluate the code from the C++ domain and I doubt, that an arrow function can resolves 'this' correct.
I think we should not merge this code unless we have a working unit test for each of these pattern:
-beginTimer with arrow-function
-beginTimer with external function name
-beginTimer with function() {}.bind(this)
-beginTimer with function body in exclamation marks

The last might fail, but I think it is fixable, because we pass it as string in https://github.com/mixxxdj/mixxx/blob/dabca58d22a45e71500e1f141826c4017287ca7e/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp#L472-L473 anyway.

JoergAtGithub:
BTW: I wonder if this `this` bug is related: https://doc.qt.io/qt-6/qtqml-cppintegration-exposecppattributes.html#c-methods-and-the-this-object

Swiftb0y:
>  But unfortunatly, we are not only in the JavaScript domain here. We evaluate the code from the C++ domain and I doubt, that an arrow function can resolves 'this' correct.

Can you elaborate? I don't understand your concern. 

>  I think we should not merge this code unless we have a working unit test for each of these pattern:

I'll see how feasible it is to implement those. I don't know how testable our JS Api is. 

> BTW: I wonder if this this bug is related:

Yes, but its QML- and Qt6-only and requires an even bigger API change, so I ruled it out as a potential solution.

JoergAtGithub:
> > But unfortunatly, we are not only in the JavaScript domain here. We evaluate the code from the C++ domain and I doubt, that an arrow function can resolves 'this' correct.
> 
> Can you elaborate? I don't understand your concern.

I have some code that looks like this:
```
class MyClass {
    constructor() {
         this.myVar = 1;
    }
    buttonHandler(field) {
         this.shiftPressedTimer = engine.beginTimer(200, ()=> {
              this.myVar = 0;
        }, true);
    }
}
```
and the timer code never sets the member var of the class.

But what works is:
```
class MyClass {
    constructor() {
         this.myVar = 1;
    }
    buttonHandler(field) {
         this.shiftPressedTimer = engine.beginTimer(200, function() {
              this.myVar = 0;
        }.bind(this), true);
    }
}
```

This lead me think, that the arrow function doesn't get the relation to it's parent JS class.

But honestly I'm not sure about this.

Swiftb0y:
> The example I posted above is taken from a real mapping, that doesn't work with arrow function.

which one exactly? I'll look into it.

> Maybe I'm thinking to simple, but I thought removing exclamation marks around a code snippet to evaluate, should be an easy regex before 

Yes, of course the most robust solution would be to not pass as a string. That's why this PR eliminates that for all mappings we have. But we can't do anything about mappings that have not been contributed. The arrow function on the C++ side is there because otherwise the code was evaluated immediately instead of after the timer, which is obviously a bug.

JoergAtGithub:
> > The example I posted above is taken from a real mapping, that doesn't work with arrow function.
> 
> which one exactly? I'll look into it.

A version of my Z2 mapping under development: https://github.com/mixxxdj/mixxx/blob/8796b3cc5145f399daed76ae6870215bbcf0567f/res/controllers/Traktor-Kontrol-Z2-hid-scripts.js#L1396-L1405

Swiftb0y:
I don't have access to a Z2, so I can't test this in context of your mapping. I created a little QJSEngine playground outside of Qt, but I can't reproduce the issue there. I'll try reproducing it in Mixxx.

Swiftb0y:
Can't even reproduce within mixxx with a test script. Please double check your mapping then. I don't know what the culprit could be.

ywwg:
this is hard to test on the Traktor S3 because a bunch of other controlobjects are broken, some of which should be fixed by #12208 but others need to be fixed too.

ywwg:
(https://github.com/mixxxdj/mixxx/pull/12409

Swiftb0y:
Testing the mappings is not a requirement IMO. the only thing that worries me slightly is the issue reported by @JoergAtGithub previously. Speaking of which, if you can still reproduce the issue, what Qt version is this on? I'm compiling with 5.15.11, maybe yours is older?

Swiftb0y:
Yup, I remember that PR, but I think it would make more sense to try to integrate [Qt Quick test](https://doc.qt.io/qt-6/qtquicktest-index.html) into our current test runner instead (which would allow us to test Qt quick and by extension Qml and Js code on its own).

Swiftb0y:
I agree, your concerns are valid. I'm quite short on time due to university, so if you want to do me a favor and get this merged, I would really appreciate if you could take care of the unit test implementation.

ywwg:
This is not the first time we have to choose between "PR without test" or "No PR" -- if this were a merge to main then I'd feel better about just getting it in. 

Lacking a test, what I'd prefer to do is only fix those mappings that we know have problems, and save the mass search-and-replace for `main`. That way we have less risk of breaking the imminent release.

ronso0:
We could also only fix those mappings where the pattern matches the mappings that were reported to be broken (and fixed by this), namely the DDJ-SB3 and Mixtrack Pro 3, #12384 and #12369, and merge the rest to main.
Didn't check what would be left then x)

JoergAtGithub (approved):
I created testcases for engine.beginTimer myself in #12437 and all tested code pattern work as intended.
o.t.: I couldn't reproduce the issue, that I see with my Z2 mapping with these testcase. So there might still be a latent error in the code of the beginTimer implementation.

JoergAtGithub:
Thank you!

Swiftb0y:
Thank you.

PR #12406, opened 2023-12-07
common-hid-packet-parser.js: Fix regression for resolving decks in getActiveFieldGroup 

JoergAtGithub:
This restores the old behavior of getActiveFieldGroup without altering the behavior of resolveGroup
And adds warnings at various callers, for the case that resolveGroup returns undefined

This fix a regression introduced in: #12102 
Closes: #12208


ywwg (approved):

JoergAtGithub:
@ywwg It would be great, if you would merge this soon!

ywwg:
This works! thanks.

PR #12411, opened 2023-12-08
fix clamping of the index for drawing the waveform left of zero position

m0dB:
Left of the zero position was drawn was the value of the first position. I guess nobody noticed because the start of tracks tend to be low amplitude.

Before:
Screenshot 2023-12-08 at 11 15 32

After:
Screenshot 2023-12-08 at 11 49 00


ywwg (approved):

JoergAtGithub:
Tested on Windows11 - all fine. Code LGTM! Thank you!

PR #12414, opened 2023-12-08
use frameSwapped - update for continuously repaint instead of swapBuffers from vsyncthread

m0dB:
According to the Qt docs:

> Applications that wish to continuously repaint synchronized to the vertical refresh, should issue an [update](https://doc.qt.io/qt-6/qpaintdevicewindow.html#update-2)() upon the \[QOpenGLWindow::frameSwapped\] signal. This allows for a much smoother experience compared to the traditional usage of timers.
> This signal is emitted after the potentially blocking [buffer swap](https://doc.qt.io/qt-6/qopenglcontext.html#swapBuffers) has been done. 

This PR uses the mechanism, replacing the manual swapBuffers calls from the vsyncThread. The timing mechanism from the vsyncThread is still used, but now triggered by the frameSwapped signal.

Apart from being the recommended way, this also solves artifacts in the OpenGL widgets when resizing / moving the splitter.

I didn't want to fully replace the old code, so this is all conditional using the const bool USE_FRAME_SWAPPED.

### DIFF
--- a/src/waveform/useframeswapped.h
+++ b/src/waveform/useframeswapped.h
@@ -1,0 +1,1 @@
+constexpr bool USE_FRAME_SWAPPED = true;
--- a/src/waveform/vsyncthread.cpp
+++ b/src/waveform/vsyncthread.cpp
@@ -3,6 +3,7 @@
 #include "moc_vsyncthread.cpp"
 #include "util/math.h"
 #include "util/performancetimer.h"
+#include "waveform/useframeswapped.h"
 
 VSyncThread::VSyncThread(QObject* pParent)
         : QThread(pParent),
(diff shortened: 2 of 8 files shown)


m0dB:
Ok, the vu-meters are triggered independently, that is why they do show. Apparently the frameSwapped signal is not triggered at all on your system.

JoergAtGithub:
Setting format.setSwapInterval(1); in \src\waveform\waveformwidgetfactory.cpp let the waveform appear

JoergAtGithub:
But with format.setSwapInterval(1) I get only 25fps instead of 60fps.

m0dB:
> I did only work once with format.setSwapInterval(1); - not reproduceable anymore - might be something else

Ok. It is quite strange that this recommended mechanism isn't working.

m0dB:
But it is a very interesting discovery that the vumeters are actually being drawn without explicitly calling them.

JoergAtGithub:
The frameSwap function in VSyncThread is executed, the problem seems to be, that the timer is not initialized/started.
I added `m_timer.start();` to the constructor of VSyncThread and the wavforms appear always independent of format.setSwapInterval.

m0dB:
But I now realise everything is rendered twice: once by the continous repaint (the update from the frameSwapped), once explictly by the vsyncthread - waveformwidgetfactory... This needs a bit more thought.

m0dB:
> The frameSwap function in VSyncThread is executed, the problem seems to be, that the timer is not initialized/started. I added `m_timer.start();` to the constructor of VSyncThread and the wavforms appear always independent of format.setSwapInterval.

Okay, got it. I will make some changes. Thanks a lot for trying!

JoergAtGithub:
In general I think the single timer inteval is not a good base for fixed VSync interval rendering. I think it would be better to store several frame swap timestamps in an array and apply a filter on the intervals, to interpolate the time of the next VSync.

m0dB:
Hmm, it is really tricky to do this while maintaining the old code...

m0dB:
Needs more thought...

PR #12417, opened 2023-12-09
MSVC: Suppress possible loss of data warning 

daschuer:
a /w after /W3 seems to no longer work "overriding '/W3' with '/w'"

It puzzles me, why this happens only without PCH ...


JoergAtGithub:
I build this locally and see no warnings for these files. LGTM! Thank you!

PR #12418, opened 2023-12-09
Display the samplerate in the track info dialog.

daschuer:
During debug, I had no quick way to look up the sample rate. Here it is: 

grafik

I think it is also useful for the user to pick the soundcard sample-rate. 

This is on top of 2.4 and can IMHO slip in 2.4. But I don't mind to merge it to 2.5 otherwise.


JoergAtGithub:
Works flawless for me! Code LGTM! Waiting for CI!

JoergAtGithub:
Thank you!

PR #12421, opened 2023-12-10
Labeler: Migrate configuration to v5

fwcd:
This adapts our PR labeler configuration to the v5 syntax (which is broken as of #12394).

Note that we have to merge the PR for the updated configuration to take effect, even for this PR itself.


ronso0:
Well, that's also how I understand the documentation https://github.com/actions/labeler/releases/tag/v5.0.0-beta.1

Let's see how it works.

Thanks @fwcd

PR #2737, opened 2020-04-29
add ControlObjects for shifting cue points

Be-ing:
To correct for timing offsets when importing from other software.
This does not shift the beatgrid because Mixxx does not currently
import beatgrids.

PR #12423, opened 2023-12-11
Trackmenu: shift cues and beats

ronso0:
I was wondering why fixing track offsets after migration or decoder changes requires shifting beats and cues individually per track.

Since no automatic detection / solution is in sight...
Adds an action to the track menu that allows shifting both beats and cues by an abitrary amount.
The last used offset is remembered.

The idea is that users can collect affected tracks that appear being shifted by the same amount, check with one track how much the cues are off, and mass-fix them easily.
Can't tell which relevance this has (never had such issues lately), but #2737 has been implememented for a reason.

**Notes**
* the tooltips for `shift_cues_..` mention the offset (10ms) while `beats_translate_...` just says "a small amount"

### DIFF
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -212,16 +212,7 @@ void BpmControl::slotTranslateBeatsMove(double v) {
     if (!pTrack) {
         return;
     }
-    const mixxx::BeatsPointer pBeats = pTrack->getBeats();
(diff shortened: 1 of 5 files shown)


ronso0:
Same as #12493 
In order to clean up the backlog of open PRs (and due to lack of interest) I'll close this.
When I got the S4 I implemented it in the mapping (Shift + GRID + MOVE encoder)

PR #12431, opened 2023-12-12
Sync Lock: disallow selection of Explicit Leader

ywwg:
In response to the sync_leader CO, enable soft leader instead of explicit leader. This is due to the large number of side effects and bugs in the explicit leader code. Effectively this removes the Explicit Leader feature from mixxx 2.4.  The user can still select which deck they want to be soft leader if that selection is valid, but Mixxx will not allow the user to select an invalid leader -- specifically, a stopped deck while another deck is playing. If all decks are stopped, the first deck with a loaded track will always be leader.

Fixes: https://github.com/mixxxdj/mixxx/issues/11788.

We can reenable this later if we can fix those issues.

### DIFF
--- a/src/engine/sync/enginesync.cpp
+++ b/src/engine/sync/enginesync.cpp
@@ -56,9 +56,25 @@ void EngineSync::requestSyncMode(Syncable* pSyncable, SyncMode mode) {
         [[fallthrough]];
     }
     case SyncMode::LeaderSoft: {
-        if (pSyncable->getBaseBpm().isValid() &&
-                pSyncable->isPlaying() &&
-                pSyncable->isAudible()) {
+        if (pSyncable->getBaseBpm().isValid()) {
+            if (!pSyncable->isPlaying()) {
+                bool anyPlaying = false;
+                // Only allow a non-playing deck to be leader if no deck is playing
+                for (Syncable* pSyncable : m_syncables) {
+                    if (!pSyncable->isSynchronized()) {
+                        continue;
+                    }
+
+                    if (pSyncable->isPlaying()) {
+                        anyPlaying = true;
+                        break;
+                    }
+                }
+                if (!anyPlaying) {
+                    activateLeader(pSyncable, mode);
+                }
+                break;
+            }
             activateLeader(pSyncable, mode);
             break;
         }
--- a/src/engine/sync/enginesync.cpp
+++ b/src/engine/sync/enginesync.cpp
@@ -60,7 +60,7 @@ void EngineSync::requestSyncMode(Syncable* pSyncable, SyncMode mode) {
             if (!pSyncable->isPlaying()) {
                 bool anyPlaying = false;
                 // Only allow a non-playing deck to be leader if no deck is playing
-                for (Syncable* pSyncable : m_syncables) {
+                for (Syncable* pSyncable : std::as_const(m_syncables)) {
                     if (!pSyncable->isSynchronized()) {
                         continue;
                     }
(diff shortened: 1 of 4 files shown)


ronso0:
Please remember to update the manual as well. Thank you!

JoergAtGithub:
Please describe the intended behavior in more words. Otherwise it's impossible to decide if the code implements it correct.

daschuer:
```
Error: /home/runner/work/mixxx/mixxx/src/engine/sync/enginesync.cpp:63:17: error: c++11 range-loop might detach Qt container (QList) [-Wclazy-range-loop-detach]
                for (Syncable* pSyncable : m_syncables) {
                ^
                                           qAsConst(  )
```
Please use std::as_const()

src/engine/sync/enginesync.cpp
```diff
@@ -56,9 +56,25 @@ void EngineSync::requestSyncMode(Syncable* pSyncable, SyncMode mode) {
         [[fallthrough]];
     }
     case SyncMode::LeaderSoft: {
-        if (pSyncable->getBaseBpm().isValid() &&
-                pSyncable->isPlaying() &&
-                pSyncable->isAudible()) {
+        if (pSyncable->getBaseBpm().isValid()) {
+            if (!pSyncable->isPlaying()) {
+                bool anyPlaying = false;
+                // Only allow a non-playing deck to be leader if no deck is playing
```
daschuer, line 62: Do we have a use case for this? 
* Load two decks. 
* sync both. 
* select a soft-leader
* play the soft leader -> it is picked anyway 
* play the follower -> it becomes the new leader

So it looks like users choice is ignored anyway.

ywwg: > So it looks like users choice is ignored anyway.

I think I can fix that be preferring the current leader if there are multiple equivalent choices

ywwg:
https://github.com/mixxxdj/manual/pull/600

daschuer:
Thank you. What is the workflow where the user has a benefit if a preselected soft leader. Can you give a hint for workflow we can test?

ywwg:
> Thank you. What is the workflow where the user has a benefit if a preselected soft leader. Can you give a hint for workflow we can test?

Mostly it's to prevent user confusion when the decks are stopped, like this:

1. enable sync on both decks (both stopped)
2. by default, deck 1 is the sync leader
3. Try to select deck two as sync leader -- nothing happens! annoying!
4. play deck two -- now it activates as sync leader.

Alternatively we could always choose internal clock as sync leader when all decks are stopped.

daschuer:
With this branch (not tested) 
* Select deck 2 as implicit leader
* play deck 1, deck 1 becomes again implicit leader 
* users choice is not respected - annoying! 

So yes, hiding the leader crown on stopped deck seems to be a good workaround for the visual issue.

ywwg:
Oof unfortunately that behavior breaks a LOT of tests and assumptions, and it's not trivial to fix.  I am unable to make it work easily.  I think we can update the docs to say that "Mixxx will always prefer a playing deck for leader, overriding the user's preference" which is actually what we want, based on the original bug report.

daschuer:
Original we could also not change the soft leaders in stopped decks. So to make progress here, how about just remove the the added logic that allows to shifting a soft leader to an other stopped deck and merge this. 

During testing I have discovered that it is currently not allowed to select a soft leader. The feature would be nice use this to switch to the internal clock as leader.

ywwg:
ok done.

ywwg:
In the future I'd like to make it so that when all decks are stopped, the InternalClock is leader, but doing that causes too many test failures / side effects right now.  I think it can be solved with some logic fixes around setting the internal clock BPM but that's out of scope for 2.4

daschuer (approved):
LGTM and works good. Thank you for taking care.

Russe:
Excellent, thanks for resolving this issue. I tested the most recent beta, now the problem (#11788) should never happen again.

PR #12434, opened 2023-12-14
Bump libc6 downgrade version

daschuer:
libc6=2.35-0ubuntu3.4 is no longer available

PR #12442, opened 2023-12-17
Use imported target ZLIB::ZLIB

daschuer:
This should fix the build issue reported here: 
https://github.com/mixxxdj/mixxx/pull/12434#issuecomment-1859081823


JoergAtGithub:
Confirmed, I could build both x64_portable and x64_off again. Waitng for CI, to see that also the rel-only buildenv works.

JoergAtGithub:
Thank you for the fast fix!

PR #12450, opened 2023-12-18
Exclude Windows-only sources from Linux/macOS build

uklotzde:
Fixes a build warning caused by automoc.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -989,7 +989,6 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/musicbrainz/web/coverartarchiveimagetask.cpp
   src/musicbrainz/web/coverartarchivelinkstask.cpp
   src/musicbrainz/web/musicbrainzrecordingstask.cpp
-  src/nativeeventhandlerwin.cpp
   src/network/jsonwebtask.cpp
   src/network/networktask.cpp
   src/network/webtask.cpp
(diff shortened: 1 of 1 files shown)


daschuer (changes requested):
Thank you for taking care. 
In other classes we deal with it by adjusting the position of the #ifdefs.
This is also a 2.4 issue.

daschuer:
This discovers also an unrelated issue 
```
D:\a\mixxx\mixxx\src\test\enginebufferscalelineartest.cpp: warning C4651: '/DDJINTEROP_SOURCE' specified for precompiled header but not for current compile
```

uklotzde:
I don't care how the issues are fixed. But someone should do it.

uklotzde:
Btw, all these unrelated breakage is an awful developer experience. PCH seems to become the new boss in this respect. It does not only happen here but also in other PRs.

uklotzde:
@JoergAtGithub Cherry-picked your commit from #12452.

uklotzde:
Rebased and retargeted.

daschuer:
An example is `src/util/console.cpp`

uklotzde:
After further analysis I came to the conclusion that cleaning up all those platform-dependent sources and fixing inconsistencies is a rabbit hole. There is a lot of legacy code that needs to be redesigned for proper isolation between platforms without cluttering the code with `#ifdef`s.

Since I am not aware how the project plans to fix these I am closing this PR. Someone else might do the quick fix.

uklotzde:
My recommendation would be to extract platform specific code into separate classes/files/directories as much as possible, e.g. place it in folders `src/platform/(linux|macos|windows)`. This includes various `SoundSource`s, battery management, ...

In Rust I try to isolate platform-specific code in separate modules that are enabled conditionally at the outer level by a feature flag. Then you don't need to clutter the native code with conditional compilation instructions. Some thoughts are necessary to come up with designs that allow to hook platform-specific code into the generic code at a single extension point.

PR #12452, opened 2023-12-19
Link library DJINTEROP as PUBLIC, to make it visible in mixxx-test too

JoergAtGithub:
This fixes a build issue, when libdjinterop is used from the VCPKG buildenv. Currently we use the VCPKG version of libdjnterop only in the Main branch. Therefore the build of 2.4 did not fail, but the bug in the CMakelist is also in here.
That it works can be seen in this Main based branch:
https://github.com/JoergAtGithub/mixxx/actions/runs/7267409626/job/19801163505


uklotzde (approved):

daschuer (approved):
LGTM, thank you.

PR #12456, opened 2023-12-20
Pioneer DDJ-ERGO:  added mappings

jssotomdz:
Followed the original mapping as close as possible here.


jssotomdz:
Created manual merge request as well
https://github.com/mixxxdj/manual/pull/602

JoergAtGithub:
Welcome at Mixxx!

jssotomdz:

Done!

JoergAtGithub:
The best way to fix pre-commit issues is to install pre-commit locally on your system, as described here: https://github.com/mixxxdj/mixxx/wiki/Using-Git#set-up-automatic-code-checking
grafik
unzip it, and apply it using:
`git apply pre-commit.patch`

jssotomdz:
> The best way to fix pre-commit issues is to install pre-commit locally on your system, as described here: https://github.com/mixxxdj/mixxx/wiki/Using-Git#set-up-automatic-code-checking Than it will fix these issues automatically before every commit. Alternatively, you can download the pre-commit.patch file from the artifacts of this PR grafik unzip it, and apply it using: `git apply pre-commit.patch`

Thanks, applied them

JoergAtGithub:
The code stale is applied now, the remaining eslint errors need to be fixed manually

jssotomdz:
> The code stale is applied now, the remaining eslint errors need to be fixed manually

Unused parameters should be removed from function declaration? And the constants storing the callback connections should be dismissed and just call the methods without assigning them?

JoergAtGithub:
Unused parameters are allowed, when they are named starting with an underscore. But if there is no reason to keep them, better delete them, than adding an underscore.
Storing the connections, make only sense, if you later use them for .trigger() or .disconnect().

jssotomdz:
> Unused parameters are allowed, when they are named starting with an underscore. But if there is no reason to keep them, better delete them, than adding an underscore.
> Storing the connections, make only sense, if you later use them for .trigger() or .disconnect().

Fixed. It is kind of strange though. Sometimes status is marked as used even though it isn't referenced in the code. Is there some guidance over this? And for some reasons, Mixxx is complaining about expecting a `;` whenever I use `let` as a declaration. That is why I use `var` but then the action complains. How should I deal with those?

jssotomdz:
sorry, messed up the PR. Will rebase and reopen

Issue #11552, opened 2023-05-14
macOS sandboxing: User reprompted for folder permission at every launch


fwcd:
### Bug Description

In recent Mixxx snapshots, the user is asked for permission to access a folder every time Mixxx launches:

![image](https://github.com/mixxxdj/mixxx/assets/30873659/6bc35a6e-2bd4-49af-add9-c08189c56562)

Not sure if this is intentional, most apps on macOS persist this permission in some way, should Mixxx do something similar? This dialog shows up even if the user has granted "Full Disk Access" via the system settings:

![image](https://github.com/mixxxdj/mixxx/assets/30873659/de08a995-0b32-4140-89b6-ebc57923243c)

This issue seems to be present in all snapshots since 3f9d90c16e, snapshots before 9e974cdb44 didn't have it so it's most likely caused by a sandboxing change [between these commits](https://github.com/mixxxdj/mixxx/compare/9e974cdb44..3f9d90c16e).

### Version

2.5-alpha-33-g3e500b1abc

### OS

macOS 13.4 (arm64)

fwcd:
Maybe of interest: https://lapcatsoftware.com/articles/2023/6/1.html

fwcd:
I cannot reproduce this any longer. Maybe it was macOS Sonoma that fixed it or https://github.com/mixxxdj/mixxx/pull/11936?

If anyone can reproduce this, feel free to comment and I'll reopen this.

PR #3761, opened 2021-04-02
Remove TrackFile class

uklotzde:
~~### TODO~~
- ~~[ ] Check if the sandboxing commit could be split into smaller chunks. Although I don't expect that it is worth the effort.~~

### DONE
- Replaced the special case `TrackFile` with `mixxx::FileInfo` and `mixxx::FileAccess`. The latter includes a sandboxing token.
- `Sandbox`: Create a sandboxing token from either `mixxx::FileInfo` or `QDir`. I preferred not to remove the existing `QDir` member functions in favor of `mixxx::FileInfo` (that could represent both, i.e. files and directories).

PR #12457, opened 2023-12-22
Sandbox: Fix regression that caused `Sandbox::canAccess` to fail

### LINKED ISSUE #12137: Mixxx seeking access permission form iTunes Media folder on every launch

fwcd:
### Fixes #12137 (and #11552)

This fixes a regression with macOS sandboxing introduced in #3761 that caused `Sandbox::canAccess` to always fail and to wrongly present this dialog on every launch:

image

The issue was that the token is not assigned to a variable and thus immediately deinitialized (and invalidated):

```cpp
bool Sandbox::canAccess(mixxx::FileInfo* pFileInfo) {
    ...
    openSecurityToken(pFileInfo, true); // Token is invalidated immediately
    return pFileInfo->isReadable();     // therefore this may wrongly return false!
}
```

See https://github.com/mixxxdj/mixxx/pull/3761#issuecomment-1867131495 for details. To avoid regressing on this in the future, I have added a `[[nodiscard]]` to the `openSecurityToken` methods, which will generate a warning if used like this again.

### DIFF
--- a/src/util/sandbox.h
+++ b/src/util/sandbox.h
@@ -54,17 +54,17 @@ class Sandbox {
     }
 
 #if defined(__GNUC__) && __cplusplus < 201907L
-#define RATIONALE
+#define SECURITY_TOKEN_NODISCARD_RATIONALE
 #else
-#define RATIONALE                                                      \
+#define SECURITY_TOKEN_NODISCARD_RATIONALE                             \
     ("A new security token should be used, e.g. by assigning it to a " \
      "variable, otherwise it will be invalidated immediately.")
 #endif
 
-    [[nodiscard RATIONALE]] static SecurityTokenPointer openSecurityToken(
-            mixxx::FileInfo* pFileInfo, bool create);
-    [[nodiscard RATIONALE]] static SecurityTokenPointer openSecurityTokenForDir(
-            const QDir& dir, bool create);
+    [[nodiscard SECURITY_TOKEN_NODISCARD_RATIONALE]] static SecurityTokenPointer
+    openSecurityToken(mixxx::FileInfo* pFileInfo, bool create);
+    [[nodiscard SECURITY_TOKEN_NODISCARD_RATIONALE]] static SecurityTokenPointer
+    openSecurityTokenForDir(const QDir& dir, bool create);
 
   private:
     Sandbox() = delete;
@@ -72,7 +72,8 @@ class Sandbox {
     static ConfigKey keyForCanonicalPath(const QString& canonicalPath);
 
     // Must hold s_mutex to call this.
-    [[nodiscard RATIONALE]] static SecurityTokenPointer openTokenFromBookmark(
+    [[nodiscard SECURITY_TOKEN_NODISCARD_RATIONALE]] static SecurityTokenPointer
+    openTokenFromBookmark(
             const QString& canonicalPath, const QString& bookmarkBase64);
 
     // Creates a security token. s_mutex is not needed for this method.
--- a/src/util/sandbox.h
+++ b/src/util/sandbox.h
@@ -53,14 +53,18 @@ class Sandbox {
         return createSecurityToken(dir.canonicalPath(), true);
     }
 
-    [[nodiscard(
-            "A new security token should be assigned to a variable, otherwise "
-            "it will be invalidated immediately.")]] static SecurityTokenPointer
-    openSecurityToken(mixxx::FileInfo* pFileInfo, bool create);
-    [[nodiscard(
-            "A new security token should be assigned to a variable, otherwise "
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
Doesn't the compiler optimize this unused variable away?

Holzhaus:
> Doesn't the compiler optimize this unused variable away?

I think not. We also use RAII and there the unused variable is not optimized away either. But in any case we should probably mark the token with `Q_UNUSED` or something?

src/util/sandbox.h
```diff
@@ -53,8 +53,14 @@ class Sandbox {
...
-    static SecurityTokenPointer openSecurityToken(mixxx::FileInfo* pFileInfo, bool create);
-    static SecurityTokenPointer openSecurityTokenForDir(const QDir& dir, bool create);
+    [[nodiscard(
+            "A new security token should be assigned to a variable, otherwise "
+            "it will be invalidated immediately.")]] static SecurityTokenPointer
+    openSecurityToken(mixxx::FileInfo* pFileInfo, bool create);
+    [[nodiscard(
+            "A new security token should be assigned to a variable, otherwise "
+            "it will be invalidated immediately.")]] static SecurityTokenPointer
+    openSecurityTokenForDir(const QDir& dir, bool create);
```
daschuer, line 63: This fails unfortunately on Ubuntu Focal  
```
src/util/sandbox.h:59:62: error: wrong number of arguments specified for ‘nodiscard’ attribute
   59 |     openSecurityToken(mixxx::FileInfo* pFileInfo, bool create);
      |   
```

This works: 

```suggestion
#if defined(__GNUC__) && __cplusplus < 201907L
#define RATIONAL
#else
#define RATIONAL                                                            \
    ("A new security token should be assigned to a variable, otherwise it " \
     "will be invalidated immediately.")
#endif

    [[nodiscard RATIONAL]] static SecurityTokenPointer openSecurityToken(
            mixxx::FileInfo* pFileInfo, bool create);
    [[nodiscard RATIONAL]] static SecurityTokenPointer openSecurityTokenForDir(
            const QDir& dir, bool create);
```

daschuer (changes requested):

daschuer:
I think we have no issue with the variable being unused. It is a QSharedPointer<SandboxSecurityToken> which has a scope until the end of the function.

src/util/sandbox.h
JoergAtGithub: I don't think we should add workarounds for such old C++ language versions.
This is in line with our [Minimum requirements policy](https://github.com/mixxxdj/mixxx/wiki/Minimum-requirements-policy). The oldest supported Ubuntu version is currently 22.04.

daschuer: This is no reason to drop Ubuntu Focal support on the last minute before the 2.4 release. 
The alternative would be to make the rational a comment. It will anyway never appear, because the bug is now fixed.

JoergAtGithub: Than we should at least ensure, that this workaround get not merged into Main.

daschuer: I will take care.

fwcd: Nit: It should be "rationale"

fwcd: > Than we should at least ensure, that this workaround get not merged into Main.

We could simply add a commit on `main` removing this workaround after the next 2.4 -> `main` merge. That should make sure that everything not touching those lines would merge cleanly in the future.

daschuer: I have updated the suggestion.

Swiftb0y (approved):

src/util/sandbox.h
```diff
@@ -53,17 +53,27 @@ class Sandbox {
         return createSecurityToken(dir.canonicalPath(), true);
     }
 
-    static SecurityTokenPointer openSecurityToken(mixxx::FileInfo* pFileInfo, bool create);
-    static SecurityTokenPointer openSecurityTokenForDir(const QDir& dir, bool create);
+#if defined(__GNUC__) && __cplusplus < 201907L
+#define RATIONALE
```
uklotzde, line 57: Introducing a define with a generic name in a widely used header file is probably not a good idea.

fwcd: I agree, I forgot this was the header and not the implementation file when I saw this. I've renamed it to `SECURITY_TOKEN_NODISCARD_RATIONALE`.

daschuer (approved):
Thank you. 2.4 is coming into sight!

PR #12462, opened 2023-12-23
Library: fix track relocation query

ronso0:
After all it was missing `:` that made the entire relocation fail.
Seems this was broken for years.

I'm surprised there's no error raised for no-op `bindvalue()`.

mentioning **Relink** so I find this PR later on

### DIFF
--- a/src/library/dao/directorydao.cpp
+++ b/src/library/dao/directorydao.cpp
@@ -201,8 +201,9 @@ QList<RelocatedTrack> DirectoryDAO::relocateDirectory(
     const QString replacement = "UPDATE track_locations SET location=:newloc WHERE id=:id";
     query.prepare(replacement);
     for (int i = 0; i < loc_ids.size(); ++i) {
-        query.bindValue("newloc", relocatedTracks.at(i).updatedTrackRef().getLocation());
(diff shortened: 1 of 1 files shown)


ronso0:
@mixxxdj/developers This is quick fix and I'd appreciate if someone could take a look and press merge, for those millions of Mixxx users who will migrate to the new PCs they'll get for christmas ; )

fwcd (approved):
Looks reasonable to me, should this target 2.4 (base branch is `main` currently)?

ronso0:
jup, correct, switched to 2.4

ronso0:
Oh, oh... a bug slipped in here.
It unconditionally sets `directory` to the new **root** dir for all relocated tracks.

This will probably not be noticed. Afaict `directory` is not used, eg. "Search related.." -> Directory uses the `location` of the track, and a later relocate action will also work.
But there may be issues after purging tracks where the track file has not been deleted, see https://github.com/mixxxdj/mixxx/blob/43d6356a2fc4aa00d0399d1f12d0c3d0c04b66d4/src/library/dao/trackdao.cpp#L1116

And the `dir:` filter will not yield all results -- which partially broke #14683 for me :|

Fix is #16578

PR #12463, opened 2023-12-23
Skins: add `skins:` path alias

ronso0:
This fixes an inconvenience noticed by @mxmilkiib in https://github.com/mixxxdj/mixxx/issues/10123#issuecomment-1859008968

If you want to mod a skin that uses template urls like `skin:../Deere/some_template.xml` by copying it to the user skins directory and rename it so it's easy to identify in the skins combobox, this would break the template url scheme (requiring to alter all template urls to fix).

This PR adds the system skins directory as `skins` to the QDir searchpaths for all `src` attributes. For the plain `LaunchImageStyle` string `skins:[skin name]` is resolved to the actual path.
Now it's possible to locate templates via `src="skins:Deere/some_template.xml`, which in turn simplifies the process of creating simple skin mods:
* create user skin directory **LateNight (bright waveforms)**
* copy/paste only `skin.xml` and adjust the waveform color variables for example
* (loads all templates and stylesheets from original LateNight)
* (benefit from all LateNight updates, except those in skin.xml)

The other commits adjust Deere, LateNight and Tango to use `skins:`, and simplifies the 64-sampler templates a bit.


src/skin/legacy/legacyskinparser.cpp
```diff
@@ -398,6 +398,17 @@ QWidget* LegacySkinParser::parseSkin(const QString& skinPath, QWidget* pParent)
         }
     }
 
+    // This enables file paths like 'skins:Deere/some_template.xml',
+    // in addition to relative paths like 'skins:Deere/some_template.xml'
+    // Note: Here we assume this path exists. If it doesn't SkinLoader::getSkinSearchPaths()
+    // would have already triggered an error message.
+    // Note: we may also add the user skins path, in case there are custom skins
+    // that use the same template inheritance scheme like official skins, but we
+    // don't because unfirtunatey there is no reliable way to apply equivalent
```
daschuer, line 407:
```suggestion
    // don't because unfortunately there is no reliable way to apply equivalent
```

daschuer (changes requested):
Just a typo, the rest LGTM Thank you.

JoergAtGithub:
LGTM and works on my Windows11 system as before! Thank you!

ronso0:
Thanks for testing this!

Russe:
I don't get this to run with a modded skin. Is there any further description how to use this feature?
What do I need to do when I want to change LateNight toolbar.xml only?

These are my steps trying to get a modded skin to run:
- create user skin folder "LateNight (Test)" at .mixxx/skins/
- copy toolbar.xml from /usr/share/mixxx/skins/LateNight to .mixxx/skins/LateNight (Test)/
- change some content of this copied toolbar.xml
- restart Mixxx and choose skin "LateNight (Test)"

Now the "original" skin will be loaded instead my adapted toolbar.xml
When also copying skin.xml and adapt the path to my modded toolbar.xml the toolbar will not be displayed anymore.

ronso0:
If you want a skin mod you need to copy only skin.xml and the (top-level) files the files to be modded.
For templates in subdirectories you'd need that subdirectory and mod the respective skins.

So, in your case
* copy skin.xml and toolbar.xml
* change `<Template src="skins:LateNight/toolbar.xml"/>` to
`<Template src="skin:toolbar.xml"/>` so Mixxx looks for it in the parent skin directory
* change `toolbar.xml` to your liking
* load "LateNight (Test)"

I agree, all in all this is still no optimal solution but IMO it's better than before.

Russe:
This is what I did.
Today, I did an update of Mixxx to most recent version 2.4-beta-314-g71dceb27ba
Again, with the same steps of your description the toolbar will not be displayed any more.
Am I using the wrong version of Mixxx?

My System: Ubuntu 22.04.3 LTS

ronso0:
Hmm, works nicely for me.
Note the difference `skin:` vs. `skins:` (plural).
 I created `LateNight---toolbar` in the user skins dir, with just `skin.xml` and `toolbar.xml`, and in `skin.xml` I'm using
`<Template src="skin:toolbar.xml"/>` (which loads `LateNight---toolbar/toolbar.xml`) and I see the changes I made.

However, there seems to be a double `/` when adding the system skin path, I'll push a fix.

ronso0:
With more debug output I now see the QDir searchPaths persist, i.e. `skins` is added repeatedly when reloading a skin.
I'll fix that, too.

PR #12464, opened 2023-12-24
mp3 garbage detection fix

daschuer:
@acolombier has provided a file with some garbage at the beginning here https://github.com/mixxxdj/mixxx/issues/11923

Due to a missing sanity check, libmad fails to decode the file with an insane high bit rate. There is a similar check for Layer III but in this case Layer I was read from the garbage leads to bypassing the check.
I have added the check now in the Mixxx code and I can now play the file without any issues.   

The same check for layer III can be found here: 
https://github.com/markjeee/libmad/blob/c2f96fa4166446ac99449bdf6905f4218fb7d6b5/frame.c#L286


JoergAtGithub:
Could you extract this section of the mp3 file, to use it as test case for all sound sources, that support MP3?

JoergAtGithub:
Is main the right target for this bugfix?

JoergAtGithub:
LGTM! Thank you!

PR #12465, opened 2023-12-24
Azure code signing for windows

daschuer:
This uses my new Remote Code Signing Certificate on Azure. 

The PR has no Access to the Secrets, but my private build works:  
https://github.com/daschuer/mixxx/actions/runs/7311877284/job/19921895739

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -238,20 +238,6 @@ jobs:
           security set-keychain-settings Mixxx.keychain
           echo "CMAKE_ARGS_EXTRA=${CMAKE_ARGS_EXTRA} -DAPPLE_CODESIGN_IDENTITY=${APPLE_CODESIGN_IDENTITY}" >> "${GITHUB_ENV}"
 
-      - name: "[Windows] Set up Windows code signing"
-        env:
-          WINDOWS_CODESIGN_CERTIFICATE_PATH: ${{ github.workspace }}\packaging\certificates\windows_sectigo_codesign_certificate.pfx
-          WINDOWS_CODESIGN_CERTIFICATE_PASSWORD: ${{ secrets.WINDOWS_CODESIGN_CERTIFICATE_PASSWORD }}
-          WINDOWS_CODESIGN_SECURE_FILE_SALT: ${{ secrets.WINDOWS_CODESIGN_SECURE_FILE_SALT }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -345,6 +331,25 @@ jobs:
...
+        uses: azure/azure-code-signing-action@v0.2.22
+        with:
+          azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          azure-client-secret: ${{ secrets.AZURE_CLIENT_SECRET }}
+          endpoint: https://weu.codesigning.azure.net/
+          code-signing-account-name: mixxx
+          certificate-profile-name: mixxx
+          files-folder: build
+          files-folder-filter: dll,exe
```
JoergAtGithub, line 347: mixxx.exe is signed, but none of the DLLs (with exception of vcruntime140.dll, which is provided as signed binary by Microsoft).
I'm not sure, if this is neccessary, because my Windows Firewall settings are higly customized.

.github/workflows/build.yml
```diff
@@ -345,6 +331,25 @@ jobs:
           # Render analyzer waveform tests to an offscreen buffer
           QT_QPA_PLATFORM: ${{ matrix.qt_qpa_platform }}
 
+      - name: "[Windows] Sign binaries"
```
JoergAtGithub, line 334:
```suggestion
      - name: "[Windows] Sign executables"
```

JoergAtGithub: or "executable binaries"

daschuer:
I have added a code siging PR for the DLLs here: https://github.com/mixxxdj/vcpkg/pull/108

fwcd:
Just a side note, should this target 2.4?

daschuer:
Of cause ... Done

JoergAtGithub:
Should we merge this now, or should we wait for the buildenv PR with signed DLLs?

daschuer:
We can merge this now. There is nothing else to do than switch to the environment for the signed DLLs.
By the way in 2.3 the DLLs are also not signed.

JoergAtGithub:
Then let's merge this.
Then we should get Beta builds for Windows again. I'm curious what other user report - because my system is configured in a way, that it never claims about unsigned Mixxx builds.
Thank you!

ronso0:
Thanks @daschuer fo all this tedious, not very thrilling work on the build system lately!

Issue #12448, opened 2023-12-18
New waveforms have much higher visual gain than legacy ones


fwcd:
### Bug Description

https://github.com/mixxxdj/mixxx/assets/30873659/5ea174eb-0057-4844-8fe1-4ff5d0eef974

Is this intended?

cc @m0dB 

### Version

2.5-alpha-164-g6acfe63a39 (HEAD)

### OS

macOS 14.2

uklotzde:
Same here after merging the latest changes from 2.4/main:

![image](https://github.com/mixxxdj/mixxx/assets/1474154/e7caa1c9-e8ae-4453-8495-49ac34fdc308)

Doesn't look as before.

uklotzde:
For comparison the same waveform rendered with an older 2.4 beta version:

![image](https://github.com/mixxxdj/mixxx/assets/1474154/615657f4-4da8-469f-878b-89d679cefee4)

If this occurs for the current 2.4 version than it should be a release blocker!

fwcd:
Did a quick bisect over the `main` snapshots, the last good is 2c2e706b44, the first bad 0f0deb1816, [here's the diff](https://github.com/mixxxdj/mixxx/compare/2c2e706b4419794e1b29f23f124ce65f49b88bc7...0f0deb18160547bcd96e0a60d5fc7c7a9d4ba306). I assume it is this PR:

- https://github.com/mixxxdj/mixxx/pull/12205

m0dB:
Well, either this is the bug or https://github.com/mixxxdj/mixxx/issues/12115 is the bug. 

My changes were to have the colored waveforms have the same amplitude as the simple waveform. As reported in 12115 by @ronso0 , it was strange that that wasn't the case.

You will also find my analysis there that shows that the combined high-mid-low is not the same as all.

Anyway, I am fine with either approach but note that there are now two contradicting bug reports.

JoergAtGithub:
My expectation as user is, that 100% amplitude is the same as the threshold for the clipping LED in the VU-Meter - accross all waveform types.

m0dB:
I agree that that's the reasonable expectation. Now, why the waveform
data filtered.all exceeds 100%, I have no idea. I can look into that but not today. But maybe somebody knows?

In any case, if we want the same waveform amplitude across waveform types, which I think makes sense, the current glsl waveform types are correct.

fwcd:
[...]

Issue #11833, opened 2023-08-15
Waveforms like RekordBox: more difference between ups and downs along the song


ejgutierrez74:
### Feature Description

Same song in Mixxx or Rekordbox, the waveform is much clearer in Rekordbox, specially the difference between ups and downs...

One example: difference between ups and downs is not clear
![Captura desde 2023-08-15 18-34-44](https://github.com/mixxxdj/mixxx/assets/11474846/571ddac7-be53-4b9e-ae95-125f8282e962)

The same song in Rekordbox:
![imagen](https://github.com/mixxxdj/mixxx/assets/11474846/e19604ba-039f-4caa-a684-f6776ec391cb)

The difference between ups and downs are clearer.

Perhaps with some tweaks you can fix this.

hsn10:
I think both preview waves are pretty useless. It is just background for marks (which are important).

There should be plugin system for waveforms + builtin plugin manager - both small ones and scrolling zoomed one - because there will be _always_ discussions what wave is better.

daschuer:
Since Mixxx is open source there is no need for a Plug-In system. You are welcome to contribute a new waveform type directly to the Mixxx source.

ejgutierrez74:
I thought that i n preferences there is a tab for Waveform settings, perhaps changing or tweaking these options, you can make something similar to rekordbox...with more difference between ups and downs...and easier to put hot cues

Swiftb0y:
Whats the advantage of the style that rekordbox uses? Its just distorted, resulting in less and even downright misleading information IMO. Even when just comparing your screenshots, you can only really discern the break and verse from the rest. The fact that the different parts of the drops have different volume is completely lost. There is very little visual difference between the drop and intro&outro, information that is crucial for mixing. 

Also: A plugin-system for waveform does not only not make much sense when the entire application is completely open anyways, its also infeasible under the constraints of modern graphics API's. That is unless you're okay with abysmal performance or essentially needing to compile the application anyways, at which point it would be easier to implement directly, as already suggested.

ejgutierrez74:
First advantage is that Rekordbox is an standard music for djing, so this could help many people switching or changing from Rekordbox or Traktor to Mixxx.

Im a newbie trying to learn the basics ind djing and Waveform in Rekordbox, is easier to understand for me, with very little knowledge and experience. When i became something like bob Sinclair or David Guetta, perhaps id change my opinion, and would prefer Mixxx waveform ;)

[...]

PR #12466, opened 2023-12-25
Waveform tweaks

m0dB:
Adjust the scaling of the waveform data: removing a * 2 factor which doesn't seem justified and undo a scaling curve applied to the waveform.filtered.all data which also lacks an explanation.

See my analysis here https://github.com/mixxxdj/mixxx/issues/12448

### DIFF
--- a/src/waveform/renderers/waveformrenderersignalbase.cpp
+++ b/src/waveform/renderers/waveformrenderersignalbase.cpp
@@ -169,11 +169,14 @@ void WaveformRendererSignalBase::setup(const QDomNode& node,
     onSetup(node);
 }
 
-void WaveformRendererSignalBase::getGains(float* pAllGain, float* pLowGain,
-                                          float* pMidGain, float* pHighGain) {
+void WaveformRendererSignalBase::getGains(float* pAllGain,
+        bool applyCompensation,
+        float* pLowGain,
+        float* pMidGain,
+        float* pHighGain) {
     WaveformWidgetFactory* factory = WaveformWidgetFactory::instance();
     if (pAllGain) {
-        *pAllGain = static_cast<CSAMPLE_GAIN>(m_waveformRenderer->getGain()) *
+        *pAllGain = static_cast<CSAMPLE_GAIN>(m_waveformRenderer->getGain(applyCompensation)) *
                 static_cast<CSAMPLE_GAIN>(factory->getVisualGain(WaveformWidgetFactory::All));
         ;
     }
--- a/src/waveform/renderers/waveformrenderersignalbase.cpp
+++ b/src/waveform/renderers/waveformrenderersignalbase.cpp
@@ -223,3 +223,13 @@ void WaveformRendererSignalBase::getGains(float* pAllGain, float* pLowGain,
         }
     }
 }
+
+float* WaveformRendererSignalBase::unscaleTable() {
+    // Table to undo the scaling std::pow(invalue, 2.0f * 0.316f);
+    // done in scaleSignal in analyzerwaveform.h
+    static float result[256];
+    for (int i = 0; i < 256; i++) {
+        result[i] = 255.f * std::pow(static_cast<float>(i) / 255.f, 1.f / 0.632f);
+    }
+    return result;
+}
--- a/src/waveform/renderers/waveformrenderersignalbase.cpp
+++ b/src/waveform/renderers/waveformrenderersignalbase.cpp
@@ -227,10 +227,10 @@ void WaveformRendererSignalBase::getGains(float* pAllGain,
     }
 }
 
-float* WaveformRendererSignalBase::unscaleTable() {
+std::span<float, 256> WaveformRendererSignalBase::unscaleTable() {
     // Table to undo the scaling std::pow(invalue, 2.0f * 0.316f);
     // done in scaleSignal in analyzerwaveform.h
-    static float result[256];
+    static std::array<float, 256> result;
     for (int i = 0; i < 256; i++) {
         result[i] = 255.f * std::pow(static_cast<float>(i) / 255.f, 1.f / 0.632f);
     }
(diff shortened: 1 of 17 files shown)


JoergAtGithub:
The scaling for looks good now, but there seems to be something, that clamps the visual amplitude. See the following video of AllShader RGB R/L and notice the Peak Indicator of the VU-Meter:

https://github.com/mixxxdj/mixxx/assets/64457745/a5203c58-8a74-4dda-b74a-348f54e9eedc


For comparision, the next video is recorde with exactly the same setting but GLSL RGB stacked waveforms:

https://github.com/mixxxdj/mixxx/assets/64457745/4bb56544-e590-4b3f-ac3c-ac9c0f405cad

m0dB:
I see what you mean. Can you send me this track so I can have a look at the values?

m0dB:
@JoergAtGithub the scaling curve seems intentional. See the discussion @Swiftb0y dug up in https://github.com/mixxxdj/mixxx/issues/12448

m0dB:
What visual gain are you using? On what Replay Gain do you have on that track? I got -9.59306 dB. 

Here is looks like this: 
Screenshot 2023-12-28 at 01 45 12

Which seems correct, considering the ReplayGain. If I boost the visual gain with the replay gain (9.59306  dB =~ x3) I get this:

Screenshot 2023-12-28 at 01 49 24

m0dB:
I now made the gain compensation optional, so each renderer can decide whether to apply it or not. With this the amplitude is pretty consistent, only the "filtered" types are a bit different depending on the content, but I think for now this is as good as it gets. Should be merged to 2.4. Ping @ronso0 @JoergAtGithub

mxmilkiib:
Cool cool. (On my phone and busy for a while)

Btw, because it hasn't been stated; the fact that the test track looks clipped before any gain.. is that just accidental? Like, that's the waveform like Audacity or whatever shows? Because it's maybe not the best example to use when specifically looking into waveform gain/height :) e.g. if the render somehow gets clipped itself, there's no telling that apart from the true waveform.

src/waveform/renderers/waveformrenderersignalbase.cpp
```diff
@@ -223,3 +226,13 @@ void WaveformRendererSignalBase::getGains(float* pAllGain, float* pLowGain,
         }
     }
 }
+
+float* WaveformRendererSignalBase::unscaleTable() {
```
Swiftb0y, line 230:
```suggestion
std::span<float, 256> WaveformRendererSignalBase::unscaleTable() {
```

src/waveform/renderers/waveformrenderersignalbase.cpp
```diff
@@ -169,11 +169,14 @@ void WaveformRendererSignalBase::setup(const QDomNode& node,
     onSetup(node);
 }
 
-void WaveformRendererSignalBase::getGains(float* pAllGain, float* pLowGain,
-                                          float* pMidGain, float* pHighGain) {
+void WaveformRendererSignalBase::getGains(float* pAllGain,
+        bool applyCompensation,
```
Swiftb0y, line 173: maybe reorder and make it a default to `false`?

Swiftb0y (changes requested):
little modernization opportunities.

src/waveform/renderers/waveformrenderersignalbase.cpp
m0dB: or const std::array<float, 256>& maybe? what do you prefer?

src/waveform/renderers/waveformrenderersignalbase.cpp
m0dB: I prefer to keep it explicit and to keep it next to allGain.

And eventually, all this should be cleaned up and we should have a discussion about what waveforms we want to keep and how they should scale. This whole compensation business is quite weird.

m0dB:
@mxmilkiib  wrote:
> Btw, because it hasn't been stated; the fact that the test track just above looks clipped before any gain.. is that just accidental?

This is caused be the scaling curve that is being applied to the waveform and this is why it looks different in Audacity than in mixxx. I think it's a bad idea, but it's unclear to me if this is a bug or a feature, so I didn't want to change it.

Screenshot 2023-12-30 at 19 36 31

src/waveform/renderers/waveformrenderersignalbase.cpp
Swiftb0y: I'd prefer `std::span` tbh, at least until a `static constexpr` table is an option (which it is not until constexpr std::pow is available). A reference to a value type when there is also a type to represent that exact concept type available seems like the inferior solution.

src/waveform/renderers/waveformrenderersignalbase.cpp
Swiftb0y: I agree. I'm a fan of explicitness as well. The idea was just to use the default value to suggest what direction we'd like to go to in the future.

mxmilkiib:
Aah right, thank you, had kinda got the gist from the convo but that makes it explicit. I have always had this niggling feeling that Mixxx waveforms were ever so sightly.. "fatter".

Does feel like, if intentional, it would be for some take re UX. Maybe there's a convo somewhere from the date of the blame commit, if no-one remembers the rational.

It is kind of an imposition though, with no option to toggle it, or/tho is that what you've just added? I did wonder how this relates to the existing waveform scaling preferences..

I guess this in a way relates to https://github.com/mixxxdj/mixxx/issues/11833

(2p ramble; I've still not settled on a preference for scaling yet. Tbh, again, Pioneer is the club standard, they'll have had certified and paid UI people come up with those, certainly by the recent start of the era of 3000s, though obvs. their exact magic is private. Figuring that out using, like, a series pure tone samples, would be a fun for some but a super low priority side quest..)

ronso0:
@m0dB Thank you, this is much more consistent now!

src/waveform/renderers/waveformrenderersignalbase.cpp
m0dB: done

Swiftb0y (approved):
Code LGTM, though in the future, I'd really like to gain back some consistency by removing the scaling in the analyzer. Are you interested in looking into that in the future?

Issue #12460, opened 2023-12-22
Sync button does not work as expected on Numark Mixtrack Pro 3 Controller


ridicux:
### Bug Description

Hi,

Long-pressing the sync buttons on my Numark Mixtrack Pro 3 Controller is supposed to lock bpm on the respective deck. However, nothing happens instead.
If I do a short press on the sync buttons, the bpm sync happens, but at the same time, the respective deck starts playing, which is unexpected.
Sometimes, the following error is thrown when I press the sync button:
<img width="748" alt="Screenshot 2023-12-22 at 21 47 46" src="https://github.com/mixxxdj/mixxx/assets/154633625/ddc349eb-5c59-4650-8e95-a3aecfd6dd55">

### Version

2.4-beta-301-geeb6412862

### OS

MacOS 14.2.1

Swiftb0y:
confirmed previously in https://github.com/mixxxdj/mixxx/issues/12369#issuecomment-1835700452

Swiftb0y:
@ridicux can you try the mapping changes in made in #12468 (the two line changes made [here](https://github.com/mixxxdj/mixxx/pull/12468/files#diff-5d2c8b5f78f741322b10411269ca08a6cdc0417a8d3f58458f9f1e4fa7bfeb45)) and confirm whether these fix the issue for you?

shuda:
> @ridicux can you try the mapping changes in made in #12468 (the two line changes made [here](https://github.com/mixxxdj/mixxx/pull/12468/files#diff-5d2c8b5f78f741322b10411269ca08a6cdc0417a8d3f58458f9f1e4fa7bfeb45)) and confirm whether these fix the issue for you?

![image](https://github.com/mixxxdj/mixxx/assets/4147559/af8a26ee-2303-49e9-803e-210a0a30a756)

numark mixtrack pro 3 script missing 2 ".bind(this)" in line 579,583 but apart from this https://github.com/mixxxdj/mixxx/pull/12468 works well using numark mixtrack 3 pro

Swiftb0y:
Yup, I missed those. PR is updated. Thanks for testing.

ronso0:
fixed by #12468

PR #12437, opened 2023-12-16
Unit tests for engine.beginTimer

JoergAtGithub:
We've several reports about suspicious about the scripting API function `engine.beginTimer` in 2.4. To cover this functionality, I added testcase for this function.

PR #12468, opened 2023-12-27
fix(controllers): ensure callbacks have appropriate `this`

Swiftb0y:
This should fix most mappings that were previously broken, because they assumed when a function was passed as a callback to `engine.makeConnection` or `engine.beginTimer` that the `this` argument referred to the same `this` as when the function was defined (eg when defined as a "method" on a class). In reality, any function executed via [`QJSValue::call`](https://doc.qt.io/qt-5/qjsvalue.html#call) (which is the case with callbacks executed via
`ControllerScriptEngineBase::executeFunction`), will have its `this` argument set to the globalObject!

This commit fixes some of these issues in known cases (https://github.com/mixxxdj/mixxx/issues/12460), potential places introduced as part of (https://github.com/mixxxdj/mixxx/pull/12401) as well as some other potential spots that were easy to grep for.

### DIFF
--- a/res/controllers/Numark-Mixtrack-3-scripts.js
+++ b/res/controllers/Numark-Mixtrack-3-scripts.js
@@ -422,7 +422,7 @@ SingleDoubleBtn.prototype.buttonDown = function(channel, control, value, status,
     this.group = group;
 
     if (this.buttonTimer === 0) { // first press
-        this.buttonTimer = engine.beginTimer(this.doublePressTimeOut, this.buttonDecide, true);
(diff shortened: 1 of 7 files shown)


Swiftb0y:
@JoergAtGithub it probably makes sense to assert this behavior using a unit test in #12437

Swiftb0y:
friendly ping. This has been confirmed working for the mixtrack 3 in https://github.com/mixxxdj/mixxx/issues/12460#issuecomment-1873813821

Holzhaus (approved):
LGTM, thanks.

Swiftb0y:
Thank you.

PR #12475, opened 2023-12-29
store/restore regular loop when toggling rolling loops

### LINKED ISSUE #8947: rolling beatloop shouldn't alter beatloop size in spinbox
Date: 2017-10-06T21:32:50Z
Status: Confirmed
Importance: Low
Launchpad Issue: [lp1721881](https://bugs.launchpad.net/bugs/1721881)

---

This is a bug only if you use beatlooproll like I do:
I have 4 buttons on my controller to activate 4 rolling beatloops (1/4, 1/2, 1, 2 beats) and it's fun to use them as a stutter effect and for 'glitchy transitions'. Another button is mapped to beatloop_activate and uses beatloop_size.

Any rolling loop deletes/overwrites the last enabled loop, places loop markers on the waveform and alters beatloop_size shown by the spinbox.

Point is, I don't want to recall those spontaneous, tiny rolling loops later, more likely they would overwrite an intro loop that I'd like to save.
Would it make sense to ignore *rolling* beatloops that are shorter than -let's say 1 beat or ignore them generally? Manually set and normal loops by any size would still saved, of course.

ronso0:
kind of fixes #8947 
* have a regular loop set (maybe it's also active)
* activate a rolling beatloop
⇒ loop size spinbox is adopting the size of the rolling loop
⇒ rolling loop is shown in the overview & waveform
* deactivate the rolling loop
⇒ if there was a regular loop previously, restore that
⇒ loop size spinbox is updated, previous loop is shown in the overview & waveform


ronso0:
IMO this is a bugfix (well, I wrote the bug report ; ), if someone agrees I can rebase onto 2.4

Holzhaus:
I kind of agree (as rolling loops are used for different use cases as normal loops) and would like to see this in 2.4, but let's see what the others are saying 🙂

daschuer (changes requested):
There are conflicts now. 
I can also not confirm this is working. Right clicking the beatloop button is doing nothing right now.
How is slip mode considered in this new feature?

ronso0:
Conflicts resolved.

> I can also not confirm this is working. Right clicking the beatloop button is doing nothing right now.

Please describe what you did.
For me it works as expected with the looproll buttons in Tango and Loop right-click in other skins.
I've also been using this in my private branch for months without issues.

The only issue I noticed just now is that, if there was no loop when activating the rolling loop, the rolling loop is not removed (because there's no loop to store and therefore no loop to restore afterwards).
Fixed by storing the current beatloop size if there's no loop and explicitly clearing the loop controls if we only have the loop size for restore.

ronso0:
In addition to the test routine mentioned in https://github.com/mixxxdj/mixxx/pull/12475#issue-2060188944 you can now also test this:
* load a track that has no loop stored (neither temp nor saved loop)
* note the beatloop size in the spinbox
* play
* create a rolling loop (Tango: right-click loop size buttons, other skins: right-click Loop button)
* release button to deactivate rolling loop
* = prev. loop size is restored
* = no loop range in overview or waveform

daschuer:
I can confirm, it is working as desired.
I have only found an old issue that slip mode is left after releasing looproll. Is that desired?

daschuer (approved):

ronso0:
> I have only found an old issue that slip mode is left after releasing looproll. Is that desired?

Well, rolling loops are nothing more than enable slip mode and activate a loop. When all active rolling loops are disabled, slip modes is disabled, too, regardless the previous state.

PR #12476, opened 2023-12-29
Loop: `beatloop_activate` adopts rolling loop and quits slip mode

### LINKED ISSUE #9445: transform a rolling beatloop into a 'static' loop
Date: 2018-09-20T12:29:48Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1793520](https://bugs.launchpad.net/bugs/1793520)
Tags: gui, looping

---

When using rolling beatloops to fade/stutter out of a track and transition into the next, I sometimes catch a nice point in a track that would I'd like to grab and play with it.
For this it would be nice if a rolling beatloop could be transformed into a regular, static loop.

    This would allow to
    1) release the rolling beatloop button, obviously = one more finger to do stuff ;)
    2) resize and/or shift the loop

I can only think of a controller/keyboard implementation:
- 1/2 rollling beatloop is pressed
- press 'beatloop_activate' to make the loop static (or maybe 'reloop_toggle'?)
- release roll. beatloop button

ronso0:
While a rolling beatloop is active, press `beatloop_activate` to adopt this loop and quit slip mode.

This fixes #9445 for me.

* press and hold a looproll button for a transition, or just as filler
* notice I want to play with that loop, or simply need my looproll finger for other stuff, instead of holding the looproll button for the entire transition
* press `beatloop_activate`
=> loop becomes a regular loop, slip mode is deactivated (no seek!)
* release looproll button (nothin happens)
* do whatever I like with the loop

### DIFF
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -1309,6 +1309,15 @@ void LoopingControl::slotBeatLoopDeactivate(BeatLoopingControl* pBeatLoopControl
 
 void LoopingControl::slotBeatLoopDeactivateRoll(BeatLoopingControl* pBeatLoopControl) {
     pBeatLoopControl->deactivate();
+
+    if (!m_bLoopRollActive) {
+        // beatloop_activate was pressed while rolling and slotBeatLoopToggle()
+        // did already reset roll status (m_activeLoopRolls, m_bLoopRollActive)
+        // and EngineBuffer quit slip mode (but didn't seek).
+        // So nothing to do here, just leave the adopted loop active.
+        return;
+    }
+
     const double size = pBeatLoopControl->getSize();
     // clang-tidy wants auto to be auto* because QStack inherits from QVector
     // and QVector::iterator is a pointer type in Qt5, but QStack inherits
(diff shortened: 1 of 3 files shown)


ronso0:
This works well, however fo convenience with mappings that don't have `beatloop_activate` but explicit loop controls (eg. `beatloop_4_activate`) we could add this functionality to the explicit loop controls, too.
Need to double-check what happens currently when activating a loop in a rolling loop though so we don't break any (valid) use cases.

acolombier:
I think this change makes sense - this is also a lacking aspect I have missed in the past. Does this need a manual update? Feels like it would be good to document it.

ronso0:
Yup, we need to document it.
Will open a manual PR asap.

ronso0:
Hmm, not sure how I tested back then and/or what changed since then, but this wasn't working as expected.

Now it works.

ronso0:
Tested with keyboard mapping, controller test still pending.

The manual PR is https://github.com/mixxxdj/manual/pull/766

ronso0:
I almost went mad when I noticed that it works perfectly via GUI and keyboard but not with the Traktor Kontrol S4mk3 and my personal branch.

Then I realized that the S4 script does the same as #12475 :man_facepalming:

acolombier:
Yeah, I had implemented this feature before it was a thing in the engine... I'll try a give this a go within the next few day but fee free to `@` me if you haven't heard from my by the weekend!

acolombier (approved):
Tested and LGTM!

PR #12477, opened 2023-12-29
LateNight: fix xfader icons in samplers and aux units

ronso0:
The samplers xfader icns were only half the size, so doubling the size and setting the 1:1 viewbox attribute does not work for icons set in xml templates via 
```xml
<State>
  <Number>2</Number>
  <Pressed scalemode="STRETCH_ASPECT">skins:LateNight/<Variable name="BtnScheme"/>/buttons/btn__xfader_sampler_right.svg</Pressed>
  <Unpressed scalemode="STRETCH_ASPECT">skins:LateNight/<Variable name="BtnScheme"/>/buttons/btn__xfader_sampler_right.svg</Unpressed>
</State>
```
it has the contrary effect, exactly what **daschuer** noticed here https://github.com/mixxxdj/mixxx/pull/12364#issuecomment-1828761228
grafik

@m0dB please check if that looks okay on your machine.

Addionally, I fixed blurred edges caused by a pixel offset in double-size icons (Aux xfader icons).

### DIFF
--- a/res/skins/LateNight/classic/buttons/btn__xfader_sampler_left.svg
+++ b/res/skins/LateNight/classic/buttons/btn__xfader_sampler_left.svg
@@ -1,5 +1,5 @@
-<svg width="42" height="36" version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 21 18">
-	<g transform="translate(-2,-1034.4)">
-		<path d="m8 1039.4v9h8v-2h-6v-7h-2" fill="#f856e7"/>
-	</g>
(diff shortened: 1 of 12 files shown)


m0dB:
At a first glance it looks good, but some icons are now huge: 253px x 300px . Is this intentional?

ronso0:
Oh, I didn't look at the diff. The width and height attributes were removed (now both "0") as well as the closing `</svg>`.
I'll fix it.

ronso0:
Fixed. @m0dB please take anoter look.

m0dB:
LGTM, merged!

PR #12481, opened 2023-12-30
Pitch Shift effect: set Meta default to 0.5 (center)

ronso0:
matches the defaults of both linked knob parameters


Swiftb0y (approved):
yup, noticed that as well. LGTM. Thank you.

PR #12487, opened 2023-12-31
Additional paint and swap from resizeGL instead of from paintGL

m0dB:
Avoids flickering when resizing

### DIFF
--- a/src/widget/openglwindow.cpp
+++ b/src/widget/openglwindow.cpp
@@ -10,8 +10,7 @@
 #include "widget/wglwidget.h"
 
 OpenGLWindow::OpenGLWindow(WGLWidget* pWidget)
-        : m_pWidget(pWidget),
-          m_dirty(false) {
+        : m_pWidget(pWidget) {
     setFormat(WaveformWidgetFactory::getSurfaceFormat());
 }
 
@@ -26,12 +25,6 @@ void OpenGLWindow::initializeGL() {
 
 void OpenGLWindow::paintGL() {
     if (m_pWidget && isExposed()) {
-        if (m_dirty) {
-            // Extra render and swap to avoid flickering when resizing
-            m_pWidget->paintGL();
-            m_pWidget->swapBuffers();
-            m_dirty = false;
-        }
         m_pWidget->paintGL();
     }
 }
@@ -44,7 +37,10 @@ void OpenGLWindow::resizeGL(int w, int h) {
         // QGLWidget::resizeGL has devicePixelRatio applied, so we mimic the same behaviour
         m_pWidget->resizeGL(static_cast<int>(static_cast<float>(w) * devicePixelRatio()),
                 static_cast<int>(static_cast<float>(h) * devicePixelRatio()));
-        m_dirty = true;
+        // additional paint and swap to avoid flickering
+        m_pWidget->paintGL();
+        m_pWidget->swapBuffers();
+
         m_pWidget->doneCurrent();
     }
 }
--- a/src/widget/openglwindow.h
+++ b/src/widget/openglwindow.h
@@ -22,5 +22,4 @@ class OpenGLWindow : public QOpenGLWindow {
     bool event(QEvent* pEv) override;
 
     WGLWidget* m_pWidget;
-    bool m_dirty;
 };


src/waveform/vsyncthread.cpp
```diff
@@ -32,8 +36,8 @@ void VSyncThread::run() {
     m_timer.start();
 
     //qDebug() << "VSyncThread::run()";
-    while (m_bDoRendering) {
-        if (m_vSyncMode == ST_FREE) {
+    if (m_vSyncMode == ST_FREE) {
```
uklotzde, line 39: The code in the different if-else arms should be extracted into separate functions `render_free()`/`render_pll()`/`render_timer()`. Use a debug assertion in each function to verify that `m_vSyncMode` matches.

Using `switch(m_vSyncMode)` for the different cases would be more idiomatic.

m0dB: I agree but that is part of https://github.com/mixxxdj/mixxx/pull/12469 , I will change it there.

(There might be a better way to create a PR on top of other PR, to avoid this confusion)

m0dB: done

JoergAtGithub:
I tested this, the waveforms still move, but the scaling of all the OpenGLWindows is totally wrong with this PR:
grafik

m0dB:
Interesting, that definitely doesn't happen here. What happens if you make the waveforms smaller?

JoergAtGithub:
> What happens if you make the waveforms smaller?

https://github.com/mixxxdj/mixxx/assets/64457745/6a5b9051-fb83-4711-9a81-172fc57d04fa

Swiftb0y:
fyi this breaks the waveforms and spinnies completely for me on Fedora 39.

Swiftb0y:
I'm not very familiar with debugging graphics issues, but looking the problem in render doc, I can see that basically all `glDrawArrays` calls fail with `GL_INVALID_OPERATION in glGetVertexAttribfv(index==0)`. So my guess is that openGL is not initialized correctly?

m0dB:
Yes, looks like it. Argh, those platform differences. But both reports give me a good idea of what to change. I'll commit a potential fix later on today.

m0dB:
In the end I found a much simpler and cleaner approach to improve the flickering (I have renamed the PR and edited the description accordingly). Now note that this approach does not get rid of the additional update events during the resize, but at least on macOS this is much smoother.

src/widget/openglwindow.cpp
```diff
@@ -44,7 +37,10 @@ void OpenGLWindow::resizeGL(int w, int h) {
         // QGLWidget::resizeGL has devicePixelRatio applied, so we mimic the same behaviour
         m_pWidget->resizeGL(static_cast<int>(static_cast<float>(w) * devicePixelRatio()),
                 static_cast<int>(static_cast<float>(h) * devicePixelRatio()));
-        m_dirty = true;
+        // additional paint and swap to avoid flickering
+        m_pWidget->paintGL();
+        m_pWidget->swapBuffers();
```
JoergAtGithub, line 42: Doesn't trigger this a swap out of the VSync interval?
Shouldn't it be enough to paint to framebuffer with the new viewport size and wait for the next regular swap at VSync interval?

m0dB: Yes, it does trigger an additional swap for the widgets that are actually painting, but resizing messes with the swaps anyway. If you are talking about the case of `ST_PLL` it doesn't interfere with the syncing, because:

- The `frameSwapped` signal is only derived when Qt is handling the frame swap in response to an `update()` call, and not when the frame swap is done with an explicit `swapBuffers()` call` (at least, that is what I observe on macOS)
- The widget/OpenGLWindow that is driving the PLL is never resized.

> Shouldn't it be enough to paint to framebuffer with the new viewport size and wait for the next regular swap at VSync interval?

I can't get that to work without flickering.

m0dB: Now, maybe it is possible to skip the explicit paint & swap as triggered by vsync, if we detect that the resize already triggered it.

ywwg:
I think perhaps at this late date, it's too risky to include another waveform change, especially one that may cause strange behaviors.  Waveforms are extremely complex and we have to check them on 3 different platforms and as many video cards as we can find.  I think some flickering is acceptable and we can bump this change to 2.4.1

m0dB:
This is a very small change though and honestly I don't see any regression risk.

ywwg:
@JoergAtGithub is the issue you saw fixed with the change?

JoergAtGithub:
> @JoergAtGithub is the issue you saw fixed with the change?

Yes, the PR got completly rewritten since them. Now it's just a minor refacturing, which works flawless on my Windows11 system!

JoergAtGithub:
LGTM! Thank you!

PR #12488, opened 2023-12-31
(fix) library/qt6: workaround 'selected click' bug

ronso0:
Workaround for Qt6 bug https://bugreports.qt.io/browse/QTBUG-108595
> Ctrl-clicking on a selected item opens the editor instead of clearing the selection, while the selection mode is ExtendedSelection, the edit trigger is SelectedClicked, and the item has the ItemIsDragEnabled flag.

This bug has been open for a year without any activity. Please vote so that it gets attention and maybe gets fixed at some point :+1: 

Ctrl+Space, in conjunction with Ctrl+Up/Down, still works but is not practical for keyboard+mouse workflow.
___

The workaround seems to have no side effects. Ctrl+click after switching between library views with multi-/single-selection works as expected.

### DIFF
--- a/src/library/library.h
+++ b/src/library/library.h
@@ -86,6 +86,10 @@ class Library: public QObject {
         return m_trackTableFont;
     }
 
+    bool selectedClickEnabled() const {
+        return m_editMetadataSelectedClick;
+    }
+
     //static Library* buildDefaultLibrary();
 
     static const int kDefaultRowHeightPx;
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -103,6 +103,19 @@ void WTrackTableView::selectionChanged(
         const QItemSelection& selected, const QItemSelection& deselected) {
     m_selectionChangedSinceLastGuiTick = true;
     enableCachedOnly();
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    // Workaround for Qt6 bug https://bugreports.qt.io/browse/QTBUG-108595:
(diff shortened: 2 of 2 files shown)


acolombier:
I can't reproduce this bug without this fix. Could you provide detailed step to reproduce so I can help testing?

Here is what I tried:
- Click (select) track in library
-> Ctrl+click (again) unselect the track
----
- Switch between view 
- Ctrl+click and Ctrl+space on a few track
-> Ctrl+click  unselect the tracks
---
(Using Wayland)

ronso0:
> Could you provide detailed step to reproduce so I can help testing?

Sure:
* **enable "Edit metadata after clicking selected track" in Preferences -> Library**
* select a track, multiple tracks, continuous or  non-continuous, doesn't matter
* Ctrl + click to unselect a track
* = inline editor is opened, clicked track remains selected

Still happens here with Qt 6.23

acolombier (approved):
>     * **enable "Edit metadata after clicking selected track" in Preferences -> Library**

Ah I wasn't aware of that feature.

Managed to reproduce.

And the workaround fixes the issue!

ronso0:
Thanks for taking a look!

m0dB:
I can't reproduce the bug on macOS, so maybe this is OS dependent? That said, this branch works fine too, so I think it can be merged.

ronso0:
Maybe the bug is resolved (or simply not present in the first oalce) in Qt versions newer than 6.4.1?

daschuer:
@ronso0 Thank you for the fix and @acolombier for review and testing.

PR #12489, opened 2023-12-31
Default vsyncmode

m0dB:
- Make VSync mode 0 refer to the default mode; make ST_PLL the default on macOS, ST_TIMER otherwise

- Write the mode to the config
```
[Waveform]
VSync 0
```

- Users can overwrite the default in the config
```
[Waveform]
VSync <value>
```

where valid <values> are 0 (DEFAULT), 4 (ST_FREE), 5 (ST_PLL), 6 (ST_TIMER)

### DIFF
--- a/src/preferences/upgrade.cpp
+++ b/src/preferences/upgrade.cpp
@@ -19,6 +19,7 @@
 #include "util/db/dbconnectionpooler.h"
 #include "util/math.h"
 #include "util/versionstore.h"
+#include "waveform/vsyncthread.h"
 #include "waveform/widgets/waveformwidgettype.h"
 
 Upgrade::Upgrade()
@@ -65,6 +66,32 @@ WaveformWidgetType::Type upgradeToAllShaders(WaveformWidgetType::Type waveformTy
         return WWT::AllShaderHSVWaveform;
     }
     return WWT::AllShaderRGBWaveform;
+}
+
+VSyncThread::VSyncMode upgradeDeprecatedVSyncModes(int configVSyncMode) {
+    using VT = VSyncThread;
+    if (configVSyncMode >= 0 || configVSyncMode <= static_cast<int>(VT::ST_COUNT)) {
+        switch (static_cast<VSyncThread::VSyncMode>(configVSyncMode)) {
+        case VT::ST_DEFAULT:
+            return VT::ST_DEFAULT;
+        case VT::ST_MESA_VBLANK_MODE_1_DEPRECATED:
+            return VT::ST_DEFAULT;
+        case VT::ST_SGI_VIDEO_SYNC_DEPRECATED:
+            return VT::ST_DEFAULT;
+        case VT::ST_OML_SYNC_CONTROL_DEPRECATED:
+            return VT::ST_DEFAULT;
+        case VT::ST_FREE:
+            return VT::ST_FREE;
+        case VT::ST_TIMER:
+            return VT::ST_TIMER;
+        case VT::ST_PLL:
+            return VT::ST_PLL;
+        case VT::ST_COUNT:
+            return VT::ST_DEFAULT;
+        }
+    }
+
+    return VT::ST_DEFAULT;
 }
 } // namespace
 
@@ -272,6 +299,10 @@ UserSettingsPointer Upgrade::versionUpgrade(const QString& settingsPath) {
         }
 #endif
     }
+
+    config->set(ConfigKey("[Waveform]", "VSync"),
+            ConfigValue(upgradeDeprecatedVSyncModes(
+                    config->getValue(ConfigKey("[Waveform]", "VSync"), 0))));
 
     // If it's already current, stop here
     if (configVersion == VersionStore::version()) {
--- a/src/waveform/vsyncthread.cpp
+++ b/src/waveform/vsyncthread.cpp
@@ -4,13 +4,22 @@
 #include "util/math.h"
 #include "util/performancetimer.h"
 
-VSyncThread::VSyncThread(QObject* pParent)
+namespace {
+VSyncThread::VSyncMode defaultVSyncMode() {
+#ifdef __APPLE__
+    return VSyncThread::ST_PLL;
+#else
+    return VSyncThread::ST_TIMER;
+#endif
+}
+} // namespace
+
+VSyncThread::VSyncThread(QObject* pParent, VSyncThread::VSyncMode vSyncMode)
         : QThread(pParent),
           m_bDoRendering(true),
-          m_vSyncTypeChanged(false),
           m_syncIntervalTimeMicros(33333), // 30 FPS
           m_waitToSwapMicros(0),
-          m_vSyncMode(ST_TIMER),
+          m_vSyncMode(vSyncMode == VSyncThread::ST_DEFAULT ? defaultVSyncMode() : vSyncMode),
           m_syncOk(false),
           m_droppedFrames(0),
           m_swapWait(0),
@@ -217,17 +226,6 @@ void VSyncThread::setSyncIntervalTimeMicros(int syncTime) {
     }
 }
 
-void VSyncThread::setVSyncType(int type) {
-    // qDebug() << "setting vsync type" << type;
-
-    if (type >= (int)VSyncThread::ST_COUNT) {
-        type = VSyncThread::ST_TIMER;
-    }
-    m_vSyncMode = (enum VSyncMode)type;
-    m_droppedFrames = 0;
-    m_vSyncTypeChanged = true;
-}
-
 int VSyncThread::fromTimerToNextSyncMicros(const PerformanceTimer& timer) {
     int difference = static_cast<int>(m_timer.difference(timer).toIntegerMicros());
     // int math is fine here, because we do not expect times > 4.2 s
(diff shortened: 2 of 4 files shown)


JoergAtGithub:
@fwcd Could you test this on macOS?

ywwg:
We will need code as part of the config load to change the deprecated values to zero

m0dB:
> We will need code as part of the config load to change the deprecated values to zero

Not sure what you mean by that. The deprecrated values are interpreted as default.

m0dB:
@ywwg I have moved the upgrade of the deprecated values to preferences/upgrade.cpp. I left the decision of converting ST_DEFAULT to either ST_PLL or ST_TIMER to VSyncThread, but maybe you prefer to move it to WaveformWidgetFactory, where the config value is read (`WaveformWidgetFactory::startVSync`)

src/preferences/upgrade.cpp
```diff
@@ -66,6 +67,32 @@ WaveformWidgetType::Type upgradeToAllShaders(WaveformWidgetType::Type waveformTy
     }
     return WWT::AllShaderRGBWaveform;
 }
+
+VSyncThread::VSyncMode upgradeDeprecatedVSyncModes(int configVSyncMode) {
+    using VT = VSyncThread;
+    if (configVSyncMode >= 0 || configVSyncMode <= static_cast<int>(VT::ST_COUNT)) {
```
ywwg, line 73: this is perfect thanks

src/preferences/upgrade.cpp
```diff
@@ -66,6 +67,32 @@ WaveformWidgetType::Type upgradeToAllShaders(WaveformWidgetType::Type waveformTy
...
+            return VT::ST_DEFAULT;
+        case VT::ST_OML_SYNC_CONTROL_DEPRECATED:
+            return VT::ST_DEFAULT;
+        case VT::ST_FREE:
+            return VT::ST_FREE;
+        case VT::ST_TIMER:
+            return VT::ST_TIMER;
+        case VT::ST_PLL:
+            return VT::ST_PLL;
+        case VT::ST_COUNT:
```
ywwg, line 89: let's remove this case and then change the range check above to < instead of <=.  That's a little more clear.

src/waveform/vsyncthread.cpp
```diff
@@ -4,13 +4,22 @@
 #include "util/math.h"
 #include "util/performancetimer.h"
 
-VSyncThread::VSyncThread(QObject* pParent)
+namespace {
+VSyncThread::VSyncMode defaultVSyncMode() {
```
ywwg, line 8: 👍🏻

src/waveform/vsyncthread.h
```diff
@@ -13,24 +13,24 @@ class VSyncThread : public QThread {
...
     enum VSyncMode {
-        ST_TIMER = 0,
-        ST_MESA_VBLANK_MODE_1,
-        ST_SGI_VIDEO_SYNC,
-        ST_OML_SYNC_CONTROL,
-        ST_FREE,
-        ST_PLL,
-        ST_COUNT // Dummy Type at last, counting possible types
+        ST_DEFAULT = 0,
+        ST_MESA_VBLANK_MODE_1_DEPRECATED, // 1
```
ywwg, line 17: seems unnecessary to say what numbers the enums correspond to.

ywwg (approved):
Looks good, thank you! Yes this is what I imagined for how to handle the values. I made a few small notes.

fwcd:
Haven't dug into the details yet, just my observation from quickly test-driving this with the default configuration: While this mostly works well, I also get considerably more frame drops with this build (testing on an M1 Mac):

https://github.com/mixxxdj/mixxx/assets/30873659/8ebb3978-79bb-4330-888a-d90e05963fd3

For comparison, here's `main` (don't have `2.4` at hand, but I'd assume it is the same), with mostly consistent 60 fps:

https://github.com/mixxxdj/mixxx/assets/30873659/9e8699da-6537-4999-8c8f-278c7c7325cb

fwcd:
Perhaps it would make sense to defer this for now, i.e. remove it from the 2.4.0 milestone, to make sure that we have enough time to test this new mode on macOS.

m0dB:
I am not against that, as in keeping the default ST_TIMER and documenting ST_PLL, but I would like to understand why you are seeing different results than me on an M1.

For me with ST_TIMER having good or bad framerate varies per execution, with ST_PLL it is always stable once the PLL is (which is pretty fast).

Are you using fixed framerate or ProMotion in your macOS display settings? Are you using the internal (laptop) or an external monitor? If you turn on logging, do you see the PLL settling around 16667 ms?

ywwg:
I think we can give this another few rounds of testing, but if indeed the results are inconsistent, let's leave the default to ST_TIMER for all platforms.  Now that we have a default, we can always enable PLL on mac (or elsewhere) in a minor/patch release

daschuer:
I will move it to 2.4.1 for now, because it can be considered a bug fix.

fwcd:
> Are you using fixed framerate or ProMotion in your macOS display settings? Are you using the internal (laptop) or an external monitor?

This is using the built-in ProMotion laptop display with variable frame rate.

> If you turn on logging, do you see the PLL settling around 16667 ms?

I'll have to investigate that, you mean the regular debug logging, right?

m0dB:
> > Are you using fixed framerate or ProMotion in your macOS display settings? Are you using the internal (laptop) or an external monitor?
> 
> This is using the built-in ProMotion laptop display with variable frame rate.

Can you try with a fixed framerate?

> > If you turn on logging, do you see the PLL settling around 16667 ms?
> 
> I'll have to investigate that, you mean the regular debug logging, right?

Yes. It will periodically log the interval resulting from the PLL

m0dB:
It might be good though to merge this MR anyway, with the minor change of keeping the default ST_TIMER for all platforms. The main reason is that it will write the
```
[Waveform]
VSync 0
```
to the settings file. This will be easier and less error prone to instruct users to change the value 0 to 5 to try the PLL mode, rather than asking them to add the missing VSync key.

m0dB:
> Haven't dug into the details yet, just my observation from quickly test-driving this with the default configuration: While this mostly works well, I also get considerably more frame drops with this build (testing on an M1 Mac):

Hm, I just experienced lots of frame drops with this and logging the PLL showed it was not settling on 16666 microseconds. After changing the refresh rate in the macOS display settings between promotion and back to 60 Hz the problem the problem was fixed and rock solid again. Can you try if that works for you too, @fwcd ?

m0dB:
@fwcd I found the problem with ProMotion. 

If the user selected framerate (in mixxx settings) is lower than the actual (PLL) framerate, I run the swap at the nearest multiple interval. So, If the PLL is 60, the user selected frame rate is 30 (or in fact, between 25 and 40), I swap every 2 x 1/60 interval. The implementation to do this was wrong and this become noticeable at 120 fps.

I have a fix ready, and will create a separate PR.

daschuer:
Do we really need the frame rate slider in case of the PLL?

m0dB:
No, and I'd be happy to remove the logic to deal with the frame rate value. Better keep it simple.

In the meantime I have come to the conclusion that with ProMotion driving the continuous swapped->update iteration that triggers the PLL (at 120 fps) I keep getting more frame drops (at 60 fps) than when configuring the display at fixed frame rate. I will continue working on this but in the meantime the conclusion is: to use PLL, configure the display at fixed framerate. @fwcd can you test that?

fwcd:
I'll test-drive it once I find some time.

> in the meantime the conclusion is: to use PLL, configure the display at fixed framerate

In any case we should make sure that the default vsync mode works well, regardless of whether the user uses ProMotion or has set the display to a fixed framerate.

m0dB:
Yes, I have it working now for both. Just need to do some final cleanup.

m0dB:
Can you please check https://github.com/mixxxdj/mixxx/pull/12611 ?

daschuer:
Oh sorry, I have just realized that I have pushed your changes to 2.4. 
Shall we revert that commit? Probably not. What do you think?

m0dB:
No, I don't see a need to revert it.

m0dB:
@fwcd has confirmed that now that https://github.com/mixxxdj/mixxx/pull/12611 has been merged to 2.4, ST_PLL behaves smoothly (and better than ST_TIMER) on macOS, so I would like to make a last minute attempt to get this PR merged for the release.

JoergAtGithub:
I tested this on Windows7 and Windows11. The configuration behavior is as intended. Code looking good to me and was before approved by Owen.
If I overwrite the default setting on my Windows7 laptop with Intel GPU with `VSync 5` the framerate drops from 60fps to 5.5fps, while it works well on my Windows11 laptop with AMD GPU.
Since the dafault is only changed for macOS very the hardware variance is low, and two developers tested this successful on various macOS hardware, lets merge this!
Thank you!

m0dB:
Much appreciated! I will edit the wiki page and the blogpost accordingly (mentioning that ST_PLL is now the default on macOS, and how to select ST_TIMER or ST_PLL explicitly)

Out of curiosity, when you say "works very well on windows11", is that better than ST_TIMER or the same? (i.e. when playing, but also when scrolling the library).

PR #12493, opened 2024-01-01
Shifting cue points also shift beats

ronso0:
Recently I went through my old mp3 library and came across many tracks were I had to shift the cues and beats, pretty much what was requested in the PR introducing these controls https://github.com/mixxxdj/mixxx/pull/2737#issuecomment-630276761

For now, this PR simply makes the cue shift controls `shift_cues_earlier/_later`also shift the beats accordingly.
Based on dde5302b164d8b8281bfbc0cb6fbfeca7eed5dd1 from #12423.

Let's discuss whether this should be the default (CO name doesn't fit anymore), or if we need to make it optional:
**a)** dedicated buttons for cues and cues+beats?
**b)** a preferences option (Decks? Waveforms?) that sets a static bool in `BaseTrackPlayer`?

### DIFF
--- a/src/track/track.cpp
+++ b/src/track/track.cpp
@@ -950,12 +950,27 @@ void Track::shiftCuePositionsMillis(double milliseconds) {
     VERIFY_OR_DEBUG_ASSERT(m_record.getStreamInfoFromSource()) {
         return;
     }
-    double frames = m_record.getStreamInfoFromSource()->getSignalInfo().millis2frames(milliseconds);
+    const double frames =
+            m_record.getStreamInfoFromSource()->getSignalInfo().millis2frames(
+                    milliseconds);
     for (const CuePointer& pCue : std::as_const(m_cuePoints)) {
         pCue->shiftPositionFrames(frames);
     }
 
     markDirtyAndUnlock(&locked);
+}
+
+void Track::shiftBeatsMillis(double milliseconds) {
+    if (milliseconds == 0 || !m_pBeats) {
+        return;
+    }
+    const double frames =
+            m_record.getStreamInfoFromSource()->getSignalInfo().millis2frames(
+                    milliseconds);
+    const auto translatedBeats = m_pBeats->tryTranslate(frames);
+    if (translatedBeats) {
+        trySetBeats(*translatedBeats);
+    }
 }
 
 void Track::analysisFinished() {
--- a/src/track/track.h
+++ b/src/track/track.h
@@ -290,6 +290,9 @@ class Track : public QObject {
     void setMainCuePosition(mixxx::audio::FramePos position);
     /// Shift all cues by a constant offset
     void shiftCuePositionsMillis(mixxx::audio::FrameDiff_t milliseconds);
+    /// Shift beatgrid by a constant offset
+    void shiftBeatsMillis(double milliseconds);
+
     // Call when analysis is done.
     void analysisFinished();
 
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -212,16 +212,7 @@ void BpmControl::slotTranslateBeatsMove(double v) {
     if (!pTrack) {
         return;
     }
-    const mixxx::BeatsPointer pBeats = pTrack->getBeats();
-    if (pBeats) {
-        // TODO(rryan): Track::frameInfo is possibly inaccurate!
-        const double sampleOffset = frameInfo().sampleRate * v * 0.01;
-        const mixxx::audio::FrameDiff_t frameOffset = sampleOffset / mixxx::kEngineChannelCount;
-        const auto translatedBeats = pBeats->tryTranslate(frameOffset);
-        if (translatedBeats) {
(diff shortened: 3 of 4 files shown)


daschuer:
Thank you for bringing this up. Do you have an idea why the beats where wrong for your old tracks? I think for this feature we need to consider the whole picture of use cases including beat detector updates and audio decoder issues. 

When the audio itself or the decoder  has changed, everything need to be moved. 

In case we have reanalyzed the beats, the cues need to be shifted to the new beats. 

A cue, sitting exactly on a beat before have been likely added with quantize on and should stick with the beat. 
Unfortunately this information is lost. 

Maybe we need a mass quantize cues? 
The offset is likely different per cue if we change from constvto non const beats. 

Ideas?

src/track/track.cpp
```diff
@@ -950,14 +950,29 @@ void Track::shiftCuePositionsMillis(double milliseconds) {
...
+            m_record.getStreamInfoFromSource()->getSignalInfo().millis2frames(
+                    milliseconds);
     for (const CuePointer& pCue : std::as_const(m_cuePoints)) {
         pCue->shiftPositionFrames(frames);
     }
 
     markDirtyAndUnlock(&locked);
 }
 
+void Track::shiftBeatsMillis(double milliseconds) {
```
Holzhaus, line 963: Can we use std::chrono here?

src/track/track.h
```diff
@@ -290,6 +290,9 @@ class Track : public QObject {
     void setMainCuePosition(mixxx::audio::FramePos position);
     /// Shift all cues by a constant offset
     void shiftCuePositionsMillis(mixxx::audio::FrameDiff_t milliseconds);
+    /// Shift beatgrid by a constant offset
+    void shiftBeatsMillis(double milliseconds);
```
Holzhaus, line 294: Not trying to get greedy, but while you're at it, could you fix the signature of the sibling method above as well? Framediff for milliseconds looks definitely wrong.

Holzhaus (commented):
Thanks. Some minor nits, but nothing blocking IMHO. Usabilitywise I cannot judge this, but 5ms seems very fine.

How often do you need to trigger the CO to fix your offsets? If it's a lot, it may make sense to add something like beats_translate which moves cues (i. e., nearest beat it moved to the current playback position and cues positions are moved accordingly relative to the beatgrid). Not sure if that makes sense, just an idea.

src/track/track.cpp
ronso0: Sure, let's skip mixxx::Duration.

src/track/track.h
ronso0: Sure, I spotted that, too, but didn't want to do too much refactoring until we came to a conclusion regarding the beats+cues implementation.

ronso0:
> Do you have an idea why the beats where wrong for your old tracks?

I think back then I have set cues and beats visually XD, but the visuals were off by some millis. Sometimes I noticed it, but often enough I didn't. Gear was rubbish and latency was high, so I think I blamed that for being off when hitting cue pads.

Now, when these tracks are reanalyzed by 2.5-alpha, both cues and beats are off (too late) by the same amount.
IIRC all tracks were mp3.
I never use quantize when playing, so that is irrelevant in my case.

I stumble over these tracks infrequently, and mostly when playing live, so I didn't note how much exactly I had to shift cues. IIRC I have to press the shift_back button 4-7 times, so 20-35 ms offset I guess.

ronso0:
> When the audio itself or the decoder has changed, everything need to be moved.

Sounds reasonable.
Looks like that's the case with my tracks. Beats were not reanalyzed as far as I can tell.

ronso0:
> A cue, sitting exactly on a beat before have been likely added with quantize on and should stick with the beat.

True if the user relied on the beat analysis and didn't double check (like I relied on the waveform).

I understand it's a nice challenge to automate this, but I think there are too many variables (user) to predict the correct result.

My motivation was rather to allow some sort of batch-correction, see #12423.
The missing piece there is to record the last-applied offset and make it availble for batch-correction. Like, accumulate millis from all back/forth commands and make it avalable for some **redo** control, or auto-fill the millis spinbox that can be opened via the track menu.

daschuer:
https://github.com/mixxxdj/mixxx/pull/2737 mentions another use case, import from a third party software where the beat grid import fails. 

For now I have this conclusion: We need these options
1: Shift ALL annotations to compensate audio offset
  * Decoder issues 
  * re-enoding
  * Audio edits 
 2: Shift annotations except the beats. 
  * Change beat detection algorithm
  * manual beat edits 

Both should be possible manually. 

For 1. We may detect the offset automatically via the first sound cue. An a additional plausibility check is required though in case of re-enoding.

For 2. We may use a quantize algorithm. 

What do you think? 

The GUi is a bit tricky, because we need space to inform the user adequate. Fortunately this does not need to be visible all the time.

Ideas?

ronso0:
In order to clean up the backlog of open PRs (and due to lack of interest) I'll close this.
When I got the S4 I implemented it in the mapping (Shift + GRID + MOVE encoder)

PR #12494, opened 2024-01-01
fix/History: remove obsolete placeholder playlists

ronso0:
I just noticed I made a mistake in 60198d5e376a87c0bcfcabffd2442c7e0f85b3e6  (I had ~300 `historyPlaceholder#` playlists in my database):
the YEAR placeholders are not removed in the destructor because 1) the type is `PlaylistDAO::PLHT_UNKNOWN` which is not covered by `deleteAllUnlockedPlaylistsWithFewerTracks()`, and 2) if it was, they'd be skipped because they are locked.

Note: this affects 2.4-beta users only, so we may remove this at some point (2.5?). Otoh the cleanup almost no time after the first execution.

### DIFF
--- a/src/library/trackset/setlogfeature.cpp
+++ b/src/library/trackset/setlogfeature.cpp
@@ -43,8 +43,18 @@ SetlogFeature::SetlogFeature(
     // remove unneeded entries
     deleteAllUnlockedPlaylistsWithFewerTracks();
 
-    // Create empty placeholder playlist for YEAR items
     QString placeholderName = "historyPlaceholder";
+    // remove previously created placeholder playlists
+    const QList<QPair<int, QString>> pls = m_playlistDao.getPlaylists(PlaylistDAO::PLHT_UNKNOWN);
+    QStringList plsToDelete;
+    for (const QPair<int, QString>& pl : pls) {
+        if (pl.second.startsWith(placeholderName)) {
+            plsToDelete.append(QString::number(pl.first));
+        }
+    }
+    m_playlistDao.deletePlaylists(plsToDelete);
+
+    // Create empty placeholder playlist for YEAR items
     m_yearNodeId = m_playlistDao.createUniquePlaylist(&placeholderName,
             PlaylistDAO::PLHT_UNKNOWN);
     DEBUG_ASSERT(m_yearNodeId != kInvalidPlaylistId);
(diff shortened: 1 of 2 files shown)


src/library/trackset/playlistfeature.cpp
```diff
@@ -270,7 +270,10 @@ void PlaylistFeature::decorateChild(TreeItem* item, int playlistId) {
 
 void PlaylistFeature::slotPlaylistTableChanged(int playlistId) {
     //qDebug() << "slotPlaylistTableChanged() playlistId:" << playlistId;
-    enum PlaylistDAO::HiddenType type = m_playlistDao.getHiddenType(playlistId);
+    enum PlaylistDAO::HiddenType type = PlaylistDAO::PLHT_UNKNOWN;
+    if (playlistId != kInvalidPlaylistId) { // type is known, save a query
```
daschuer, line 274: Do we have a benefit elsewhere to put this check into getHiddenType()? 
If not, is an assertion for that sensible?

src/library/trackset/setlogfeature.cpp
```diff
@@ -43,8 +43,18 @@ SetlogFeature::SetlogFeature(
     // remove unneeded entries
     deleteAllUnlockedPlaylistsWithFewerTracks();
 
-    // Create empty placeholder playlist for YEAR items
     QString placeholderName = "historyPlaceholder";
+    // remove previously created placeholder playlists
+    const auto pls = m_playlistDao.getPlaylists(PlaylistDAO::PLHT_UNKNOWN);
+    QStringList plsToDelete;
+    for (const auto& pl : pls) {
+        if (pl.second.startsWith(placeholderName)) {
```
daschuer, line 51: auto above and second here is not well readable. 
Not your fault. If you have the mood to improve it go ahead. 
Else we can ignore it in this bufix PR

daschuer (changes requested):
Thank you for taking care.

src/library/trackset/playlistfeature.cpp
ronso0: Good point, let's put in `PlaylistDAO::getHiddenType`.

`kInvalidPlaylistId` occurs if multiple playlists were deleted for example, so no need to assert it's valid. Is that what you meant?

src/library/trackset/setlogfeature.cpp
ronso0: Yeah, auto was simply faster to type.
I'll change it.

daschuer (approved):
LGTM Thank you.

PR #12495, opened 2024-01-02
[DRAFT] use QOpenGLBuffer for vertex attributes

m0dB:
Use QOpenGLBuffer and QOpenGLShaderProgram::setAttributeBuffer instead of setAttributeArray to pass vertex data to the shader. So far only used in allshader::WaveformRendererRGB but it should be simple to apply everywhere.

From the OpenGL docs:
"Buffer objects are general purpose memory storage blocks allocated by OpenGL. They are intended to be used in a great many ways. To give the implementation great flexibility in exactly what a particular buffer object's data store will be, so as to better optimize performance, the user is required to give usage hints. These provide a general description as to how exactly the user will be using the buffer object."

### DIFF
--- a/src/waveform/renderers/allshader/vertexbuffer.h
+++ b/src/waveform/renderers/allshader/vertexbuffer.h
@@ -1,0 +1,186 @@
+#pragma once
+
+#include <QOpenGLBuffer>
+#include <cassert>
+
+namespace allshader {
+class VertexBuffer;
+class Vector2DVertexBuffer;
+class Vector2DRGBVertexBuffer;
+class Vector2DRGBAVertexBuffer;
+}
+
+class allshader::VertexBuffer : public QOpenGLBuffer {
+  public:
+    VertexBuffer(int tupleSize)
+            : QOpenGLBuffer(QOpenGLBuffer::VertexBuffer), m_tupleSize(tupleSize) {
+    }
+
+    void reserve(int count) {
+        int nBytes = count * stride() * sizeof(float);
+        if (nBytes > m_allocatedBytes) {
+            allocate(nBytes);
+            m_allocatedBytes = nBytes;
+        }
+    }
+
+    void mapForWrite() {
+        void* ptr = QOpenGLBuffer::map(QOpenGLBuffer::WriteOnly);
+        assert(ptr);
+        m_pMapped = static_cast<float*>(ptr);
+        m_valueIndex = 0;
+    }
+
+    void setIndex(int index) { // in amount of tuples
+        m_valueIndex = index * m_tupleSize;
+    }
+
+    int tupleSize() const { // in amount of floats
+        return m_tupleSize;
+    }
+
+    int size() const { // in tuples
+        return m_valueIndex / m_tupleSize;
+    }
+
+    int stride() const { // in bytes
+        return m_tupleSize * sizeof(float);
+    }
+
+  protected:
+    const int m_tupleSize;
+    int m_allocatedBytes{};
+    float* m_pMapped{};
+    int m_valueIndex{}; // index into m_pMapped, in amount of floats
+};
+
+class allshader::Vector2DVertexBuffer : public allshader::VertexBuffer {
+  public:
+    Vector2DVertexBuffer()
+            : VertexBuffer{2} {
+    }
+
+    constexpr int offset() const {
+        return 0;
+    }
+
+    void add(float x, float y) {
+        m_pMapped[m_valueIndex++] = x;
+        m_pMapped[m_valueIndex++] = y;
+    }
+
+    void addRectangle(float x1, float y1, float x2, float y2) {
+        add(x1, y1);
+        add(x2, y1);
+        add(x1, y2);
+
+        add(x1, y2);
+        add(x2, y2);
+        add(x2, y1);
+    }
+};
+
+class allshader::Vector2DRGBVertexBuffer : public VertexBuffer {
+  public:
+    Vector2DRGBVertexBuffer()
+            : VertexBuffer{positionTupleSize() + colorTupleSize()} {
+    }
+
+    constexpr int positionOffset() const {
+        return 0;
+    }
+
+    constexpr int positionTupleSize() const {
+        return 2;
+    }
+
+    constexpr int colorOffset() const {
+        return positionTupleSize() * sizeof(float);
+    }
+
+    constexpr int colorTupleSize() const {
+        return 3;
+    }
+
+    void add(float x, float y, float r, float g, float b) {
+        m_pMapped[m_valueIndex++] = x;
+        m_pMapped[m_valueIndex++] = y;
+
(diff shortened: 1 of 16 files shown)


src/waveform/renderers/allshader/dataaccess.h
```diff
@@ -0,0 +1,60 @@
+#pragma once
+
+#include <QVector3D>
+#include <QVector>
+
+namespace allshader {
+template<typename T>
+class DataAccess {
+    T* m_pData{};
```
daschuer, line 9: Normally we put private members at the bottom of a class

src/waveform/renderers/allshader/dataaccess.h
```diff
@@ -0,0 +1,60 @@
...
+        m_pData = pData;
+        m_pWrite = pData;
+    }
+
+    int size() const {
+        return static_cast<int>(m_pWrite - m_pData);
+    }
+
+    void pushBack(const T& elem) {
+        *m_pWrite++ = elem;
```
daschuer, line 29: How is the overall size of the reserved memory handled?

daschuer (changes requested):
I have not understand why we need two classes, one for the access and one for allocating the memory. Can you explain?

src/waveform/renderers/allshader/dataaccess.h
m0dB: Sure, sure, still not used to that! I will change that.

m0dB:
> I have not understand why we need two classes, one for the access and one for allocating the memory. Can you explain?

Because I needed to be able to populate both memory allocated by the CPU and memory allocated by the GPU driver (though QOpenGLBuffer::allocate) and I wanted to do so without having to touch all the code and without code duplication.

But please note the Draft status :-)

Swiftb0y:
I don't understand what advantage `setAttributeBuffer` over `setAttributeArray` has.

src/waveform/renderers/allshader/vertexbuffer.h
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    template<typename T>
+    T* data(int block) {
+        if (m_blocks[block].m_elemSize != sizeof(T)) {
+            return nullptr;
+        }
+        return reinterpret_cast<T*>(reinterpret_cast<char*>(m_pMapped) + offset(block));
+    }
+};
```
Swiftb0y, line 70: ⚠️

src/waveform/renderers/allshader/dataaccess.h
```diff
@@ -0,0 +1,60 @@
...
+    ~ReservedData() {
+        delete[] m_pData;
+    }
+    void reserve(int size) {
+        if (size > m_reserved) {
+            delete[] m_pData;
+            m_pData = new T[size];
+            m_reserved = size;
+        }
+    }
```
Swiftb0y, line 48: ⚠️

src/waveform/renderers/allshader/dataaccess.h
Swiftb0y, line 29: ⚠️

Swiftb0y (commented):
Leaving some ⚠️ to mark snippets I'm very concerned about and want to revisit once this PR leaves Draft status. If we want to go down this route, lets better make sure there is significant enough benefit to justify this additional complexity.

m0dB:
> I don't understand what advantage `setAttributeBuffer` over `setAttributeArray` has. The Qt docs for both member functions are virtually identical.

The Qt docs are very unclear.

Both methods eventually call glVertexAttribPointer, which can be used with client data, or with a VBO (glBuffer). From what I understand the former is basically for backwards compatibility, the recommended way is to use VBOs. (Depending on what documentation you read, the client data use isn't even mentioned). 

See: 

https://developer.apple.com/library/archive/documentation/3DDrawing/Conceptual/OpenGLES_ProgrammingGuide/TechniquesforWorkingwithVertexData/TechniquesforWorkingwithVertexData.html

particularly under "Use Vertex Buffer Objects to Manage Copying Vertex Data"

BTW, this document has some other recommendations, like interleaving the vector and color. Of course it is doubtful that any of this really matters here, these kind of optimisations are relevant for applications with much more data throughput like games. But from a code quality perspective, I'd like to do things correctly.

I share your ⚠️ concerns and they will be addressed before this leaves draft stage. But please note that none of this is for 2.4.0., the mean reason to publish this as a PR was to check if @JoergAtGithub saw any difference with the OpenGL garbage collection.

daschuer:
Luckily Qt is open source :-) 

setAttributeBuffer() and setAttributeArray() are using glVertexAttribPointer

Since we use GL_FLOAT, there is no difference. 

https://github.com/qt/qtbase/blob/c4bd6ac4e5a0782d8771af1485a3b78733180785/src/opengl/qopenglshaderprogram.cpp#L2048

https://github.com/qt/qtbase/blob/c4bd6ac4e5a0782d8771af1485a3b78733180785/src/opengl/qopenglshaderprogram.cpp#L1797

https://registry.khronos.org/OpenGL-Refpages/gl4/html/glVertexAttribPointer.xhtml

Swiftb0y:
Right. good to know. considering the added complexity, I think we should stick to the simple and relatively safe `setAttributeArray` implementation for now. Though you're obviously free to experiment as much as you'd like in this PR.

m0dB:
> Since we use GL_FLOAT, there is no difference.

The difference has nothing to do with using GL_FLOAT or not, but with how the data is transferred to the graphics hardware. And this is potentially more optimised when using glBuffer.

> Right. good to know. considering the added complexity, I think we should stick to the simple and relatively safe setAttributeArray implementation for now. Though you're obviously free to experiment as much as you'd like in this PR.

I think in the end the code will not be more complex, or hardly. But yes, for now we can stick with setAttributeArray, and certainly for 2.4.0.

daschuer:
> The difference has nothing to do with using GL_FLOAT or not, but with how the data is transferred to the graphics hardware. And this is potentially more optimised when using glBuffer.

Where did the optimization happen? The only difference i can spot in the Qt code is the normalization flag which should be not relevant for floats. 

```
d->glfuncs->glVertexAttribPointer(location, tupleSize, type, GL_TRUE, stride, reinterpret_cast<const void *>(qintptr(offset)));
```

m0dB:
Compare setAttributeArray
```
d->glfuncs->glVertexAttribPointer(
    location, tupleSize, GL_FLOAT, GL_FALSE, stride, 
    values);
```
which passes values which is a `float*` (client memory)

with setAttributeBuffer
```
d->glfuncs->glVertexAttribPointer(
   location, tupleSize, type, GL_TRUE, stride,
    reinterpret_cast<const void *>(qintptr(offset)));
```
which passes an int offset instead, which is the offset in the currently bound VBO (QOpenGLBuffer, this driver allocated memory, with usage hints, etc).

JoergAtGithub:
Is this still relevant? Should we close this?

PR #12496, opened 2024-01-02
Numark Mixtrack 3: add new softovertake method to script

shuda:
- added UseEngineSoftTakeOver variable (default: true), enables/disables softtakeover changes (eg.: false technically works as before)
- eqlow,eqmid,eqhigh,pregain,filter and effectunit knobs now able to use engine's softtakeover method instead of the custom method implemented via script

### DIFF
--- a/res/controllers/Numark-Mixtrack-3-scripts.js
+++ b/res/controllers/Numark-Mixtrack-3-scripts.js
@@ -25,7 +25,7 @@ var ShiftFilterFX4 = true;
 var PitchBendOnWheelOff = true;
 
 // use mixxx softtakeover implementation instead of the script implemented method
-// suggestion: true 
+// suggestion: true
 var UseEngineSoftTakeOver = true;
 /**************************
  *  scriptpause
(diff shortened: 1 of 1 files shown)


JoergAtGithub:

JoergAtGithub:
grafik
unzip it, and apply it using:
`git apply pre-commit.patch`

res/controllers/Numark-Mixtrack-3-scripts.js
```diff
@@ -24,6 +24,9 @@ var ShiftFilterFX4 = true;
 // allow pitch bend with wheel when wheel is not active
 var PitchBendOnWheelOff = true;
 
+// use mixxx softtakeover implementation instead of the script implemented method
+// suggestion: true
+var UseEngineSoftTakeOver = true;
```
JoergAtGithub, line 29: Why you think an option is needed here? Shouldn't we just replace the old soft takeover functionality by engine.softTakeover?

shuda: No real reason, my first changes so as a new guy on the playground i tried to not be a "hearthless jerk" and remove someone else's work entirely with my first or. :)

My honest self thought / thinks the same thing, old soft takeover makes no sense and it must go :)

shuda: Gonna do some other fixes, gonna fix the browse button and replace / update to use 2.4 features if needed(probably tomorrow)

Should I close this PR or can I add a few more commit?

ronso0:
@shuda Is there any progress on this? (pre-commit issues, squashing)

PR #12501, opened 2024-01-02
gradually "compact" the markers if the waveform height is reduced

m0dB:
To address this issue: https://github.com/mixxxdj/mixxx/pull/12273#issuecomment-1874402179

Screenshot 2024-01-03 at 00 11 34
Screenshot 2024-01-03 at 00 11 56
Screenshot 2024-01-03 at 00 12 03
Screenshot 2024-01-03 at 00 12 09

### DIFF
--- a/src/waveform/renderers/waveformmark.cpp
+++ b/src/waveform/renderers/waveformmark.cpp
@@ -254,7 +254,7 @@ struct MarkerGeometry {
         }
 
         const float increment = std::max<float>(0.f,
-                m_labelRect.height() + 2.f - std::max(0.f, 80.f - breadth));
+                static_cast<float>(m_labelRect.height()) + 2.f - std::max(0.f, 80.f - breadth));
 
         if (alignV == Qt::AlignVCenter) {
             m_labelRect.moveTop((m_imageSize.height() - m_labelRect.height()) / 2.f);
--- a/src/waveform/renderers/waveformmark.cpp
+++ b/src/waveform/renderers/waveformmark.cpp
@@ -253,13 +253,16 @@ struct MarkerGeometry {
             m_labelRect.moveLeft(0.5f);
         }
 
+        const float increment = std::max<float>(0.f,
+                m_labelRect.height() + 2.f - std::max(0.f, 80.f - breadth));
+
         if (alignV == Qt::AlignVCenter) {
             m_labelRect.moveTop((m_imageSize.height() - m_labelRect.height()) / 2.f);
         } else if (alignV == Qt::AlignBottom) {
             m_labelRect.moveBottom(m_imageSize.height() - 0.5f -
-                    level * (m_labelRect.height() + 2.f));
+                    level * increment);
         } else {
-            m_labelRect.moveTop(0.5f + level * (m_labelRect.height() + 2.f));
+            m_labelRect.moveTop(0.5f + level * increment);
         }
     }
     QSize getImageSize(float devicePixelRatio) const {
--- a/src/waveform/renderers/waveformmark.cpp
+++ b/src/waveform/renderers/waveformmark.cpp
@@ -56,6 +56,17 @@ Qt::Alignment decodeAlignmentFlags(const QString& alignString, Qt::Alignment def
 
     return hflags | vflags;
 }
+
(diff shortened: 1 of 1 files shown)


src/waveform/renderers/waveformmark.cpp
```diff
@@ -253,13 +253,16 @@ struct MarkerGeometry {
             m_labelRect.moveLeft(0.5f);
         }
 
+        const float increment = std::max<float>(0.f,
+                m_labelRect.height() + 2.f - std::max(0.f, 80.f - breadth));
```
m0dB, line 257: The threshold for this mechanism `(80.f)` to kick in is to be decided.

src/waveform/renderers/waveformmark.cpp
```diff
@@ -254,7 +254,7 @@ struct MarkerGeometry {
         }
 
         const float increment = std::max<float>(0.f,
-                m_labelRect.height() + 2.f - std::max(0.f, 80.f - breadth));
+                static_cast<float>(m_labelRect.height()) + 2.f - std::max(0.f, 80.f - breadth));
```
m0dB, line 258: The threshold for this mechanism (80.f) to kick in is to be decided.

JoergAtGithub:
In the last case you can't click on the marker anymore. A small section of should be always visible.

src/waveform/renderers/waveformmark.cpp
fwcd: Maybe we could make this a local variable to avoid the "magic number"?

```cpp
const float theshold = 80.f;
const float increment = std::max<float>(0.f,
        static_cast<float>(m_labelRect.height()) + 2.f - std::max(0.f, threshold - breadth));
```

fwcd:
> In the last case you can't click on the marker anymore. A small section of should be always visible.

Sounds reasonable, though from what I can tell, the markers in the "main" waveform aren't clickable, only the preview ones are (or am I missing something?)

JoergAtGithub:
> > In the last case you can't click on the marker anymore. A small section of should be always visible.
> 
> Sounds reasonable, though from what I can tell, the markers in the "main" waveform aren't clickable, only the preview ones are (or am I missing something?)

Look at this example, a track from a DJ download pool prepared with Serato Hotcues and Saved Loops, which you can't access anywhere else in the Mixxx UI, because Mixxx skins have only 8 Hotcue-Buttons:

https://github.com/mixxxdj/mixxx/assets/64457745/a5206c12-bb3f-4523-8631-f69fa5897df1

fwcd:
Ah nice, TIL, didn't know this worked differently for hotcues

src/waveform/renderers/waveformmark.cpp
m0dB: Sure. I will also change the first 0.f for a local variable minIncrement, in case we decide we want to always have a bit visible, as Joerg requests.

m0dB:
Great! Would you adjust the threshold or do you think it's fine link this? (I personally think it could kick in a little bit sooner)

I disagree with @JoergAtGithub's assessment that "a small section of should be always visible", and IIUC @fwcd also thinks this isn't necessary. Indeed, if you want make the waveforms small like this, you probably aren't particularly interested in the markers at that point. But my opinion on this is not a very strong one, so maybe after merging we can ask what people think and adjust it if a majority prefers to keep a small section visible.

JoergAtGithub:
Let's discuss this on Zulip, to get more opinions

m0dB:
I just tried it out with a minimum increment of 2 and actually I like it better! It's just 2 pixels but in exchange you know that there is more than one marker there.

fwcd:
I think leaving a small section visible is a fine compromise, it hints at the user that there's more content behind the marker even if it's not shown. If the user is interested, they can expand the view as needed.

(I just typed out the same thing, what a coincidence haha)

m0dB:
Done!

fwcd (approved):

fwcd:
Thanks! Since this is a rather small, but useful UI improvement over status quo, and the concerns by @JoergAtGithub have been addressed, I'll go ahead and merge it. Tweaking the thresholds or making them configurable could be done in a future PR, if desired, but this is a good default for now.

PR #12508, opened 2024-01-04
added tsdiffmatrix.pl script

m0dB:
Script to display a matrix with the amount of differences between translation files, as discussed here https://github.com/mixxxdj/mixxx/issues/10060


daschuer:
Thank you.

PR #1187, opened 2017-02-15
new loop and beatjump controls

Be-ing:
This PR introduces 3 new ControlObjects:
* beatloop_size: indicates the size of a beatloop to set
* beatloop_toggle: sets a beatloop of the size indicated by beatloop_size
* beatloop_enabled: whether a beatloop is enabled

This also introduces a new widget for displaying beatloop_size in skins, although it may be used for any CO that indicates a number of beats. I have added tests for the new COs as well as a test for the old "beatloop" CO that had no tests. The behavior of that old CO remains the same, but now it also updates beatloop_size.

This approach has several advantages to the current approach of displaying a huge grid of buttons for different loop sizes:
* #940 will be able to fit in its minimum width using default settings, allowing it to be used on netbook screens.
* Less cluttered GUI.
* Any beatloop size supported by Mixxx (1/32 to 64) can be shown without worrying about taking up way too much screen space with more buttons.
* A place to show the selected loop size on screen for controllers that have an encoder for selecting loop size but no LED digit readout or screen on the controller. This can also be used for controllers that only have pads for loop controls by mapping buttons to beatloop_toggle, loop_halve, and loop_double. I will write JS Components to make that easy.

I tried hacking something like this together in #940 using a [WidgetStack with loop_halve/loop_double as the prev/next controls](https://github.com/mixxxdj/mixxx/pull/940/commits/3f99e1dc18cff63a9a6f351cee4d4ec34ab7f3fb), but that had several issues:
* Using beatloop_X_toggle, which many controllers that use pads for setting beatloops do, updates beatloop_size and the display in skin. This was not possible with the WidgetStack hack.
* "1/64" as the text label for a WPushButton got very wide.
* loop_halve/loop_double have a lower and upper bound for the size of loops, but with a WidgetStack the displayed number wrapped around and got out of sync with the actual loop size.

fixing https://bugs.launchpad.net/mixxx/+bug/1429331

________________________________________________________
Known remaining issues, none of which I think should delay merging:
  - [x] WBeatSpinBox does not scale with skin
  - [x] If reloop_toggle is pressed when the loop is disabled, the play position is before the loop, and the track is paused, when the track is played, it will jump to the loop instead of playing normally then catching the loop.
  - [x] If reloop_toggle is pressed while the deck is playing and the play position is before a disabled loop, then the deck seeks to another position outside the loop, the deck will then seek inside the loop.

PR #12509, opened 2024-01-04
Looping: update loopsize spinbox when recalling loop with any size

ronso0:
fixes #10511 

The loop size is now correctly detected and updated when
* setting a loop with a size not in the preset list (0.03125, 0.0625, 0.125, 0.25, 0.5, 1, 2, 4, 8 ... 512)
* by setting the size with the beatloop size spinbox and when re-activating such a saved loop

Done with new `Beats::numFractionalBeatsInRange(start, end)`.

**Note:**
Unfortunately this does not work for loops set with IN/OUT buttons  (intentionally, see [my comment below](https://github.com/mixxxdj/mixxx/pull/12509#issuecomment-1879126230)).
**edit:** addressed in #12522

### DIFF
--- a/src/audio/frame.h
+++ b/src/audio/frame.h
@@ -121,6 +121,16 @@ class FramePos final {
         return util_isfinite(m_framePosition);
     }
 
+    // returns true if a is valid and is fairly close to target (within +/- 1 frame).
+    bool isNear(mixxx::audio::FramePos target) const {
+        VERIFY_OR_DEBUG_ASSERT(isValid()) {
+            return false;
+        }
+        return target.isValid() &&
+                value() > target.value() - 1.0 &&
+                value() < target.value() + 1.0;
+    };
+
     void setValue(value_t framePosition) {
         m_framePosition = framePosition;
     }
--- a/src/audio/frame.h
+++ b/src/audio/frame.h
@@ -120,6 +120,16 @@ class FramePos final {
         return util_isfinite(m_framePosition);
     }
 
+    // returns true if a is valid and is fairly close to target (within +/- 1 frame).
+    bool isNear(mixxx::audio::FramePos target) const {
+        VERIFY_OR_DEBUG_ASSERT(isValid()) {
+            return false;
+        }
+        return target.isValid() &&
+                value() > target.value() - 1.0 &&
+                value() < target.value() + 1.0;
+    };
+
     void setValue(value_t framePosition) {
         m_framePosition = framePosition;
     }
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -1417,7 +1417,9 @@ double LoopingControl::findBeatloopSizeForLoop(
             }
         }
     }
-    return -1;
+
+    // No hit. Calculate the fractional beat length
+    return pBeats->numFractionalBeatsInRange(startPosition, endPosition);
 }
 
 void LoopingControl::updateBeatLoopingControls() {
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -17,11 +17,6 @@
 
 namespace {
 constexpr mixxx::audio::FrameDiff_t kMinimumAudibleLoopSizeFrames = 150;
-
-// returns true if a is valid and is fairly close to target (within +/- 1 frame).
-bool positionNear(mixxx::audio::FramePos a, mixxx::audio::FramePos target) {
-    return a.isValid() && a > target - 1 && a < target + 1;
-}
 } // namespace
 
 double LoopingControl::s_dBeatSizes[] = { 0.03125, 0.0625, 0.125, 0.25, 0.5,
@@ -1400,7 +1395,7 @@ bool LoopingControl::currentLoopMatchesBeatloopSize(const LoopInfo& loopInfo) co
     const auto loopEndPosition = pBeats->findNBeatsFromPosition(
             loopInfo.startPosition, m_pCOBeatLoopSize->get());
 
-    return positionNear(loopInfo.endPosition, loopEndPosition);
+    return loopEndPosition.isNear(loopInfo.endPosition);
 }
 
 bool LoopingControl::quantizeEnabledAndHasTrueTrackBeats() const {
@@ -1646,8 +1641,8 @@ void LoopingControl::slotBeatLoop(double beats,
         // or if the endpoints are nearly the same, do not seek forward into the adjusted loop.
         if (!keepSetPoint ||
                 !(enable || m_bLoopingEnabled) ||
-                (positionNear(newloopInfo.startPosition, loopInfo.startPosition) &&
-                        positionNear(newloopInfo.endPosition, loopInfo.endPosition))) {
+                (newloopInfo.startPosition.isNear(loopInfo.startPosition) &&
+                        newloopInfo.endPosition.isNear(loopInfo.endPosition))) {
             newloopInfo.seekMode = LoopSeekMode::MovedOut;
         } else {
             newloopInfo.seekMode = LoopSeekMode::Changed;
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -4,7 +4,6 @@
 
 #include "control/controlobject.h"
 #include "control/controlpushbutton.h"
-#include "engine/controls/bpmcontrol.h"
 #include "engine/controls/enginecontrol.h"
 #include "engine/controls/ratecontrol.h"
 #include "engine/enginebuffer.h"
@@ -1801,7 +1800,7 @@ void LoopingControl::slotLoopMove(double beats) {
     }
 
     FrameInfo info = frameInfo();
-    if (BpmControl::getBeatContext(pBeats,
+    if (pBeats->getContext(
                 info.currentPosition,
                 nullptr,
                 nullptr,
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -1408,7 +1408,9 @@ double LoopingControl::findBeatloopSizeForLoop(
             }
         }
     }
-    return -1;
+
+    // No hit. Calculate the fractional beat length
+    return pBeats->numFractionalBeatsInRange(startPosition, endPosition);
 }
 
 void LoopingControl::updateBeatLoopingControls() {
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -4,7 +4,6 @@
 
 #include "control/controlobject.h"
 #include "control/controlpushbutton.h"
-#include "engine/controls/bpmcontrol.h"
 #include "engine/controls/enginecontrol.h"
 #include "engine/controls/ratecontrol.h"
 #include "engine/enginebuffer.h"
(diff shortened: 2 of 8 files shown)


src/engine/controls/loopingcontrol.cpp
```diff
@@ -1307,7 +1307,9 @@ double LoopingControl::findBeatloopSizeForLoop(
             }
         }
     }
-    return -1;
+
+    // No hit. Calculate the fractional beat length
+    return pBeats->numFractionalBeatsInRange(startPosition, endPosition);
```
daschuer, line 1312: I wonder why numBeatsInRange() does not returns factions in the first place. 
It could be also required here: 
https://github.com/mixxxdj/mixxx/blob/fe0c53030ca56563dc6b07bbff3923105d9ddb45/src/engine/controls/loopingcontrol.cpp#L768

src/track/beats.cpp
```diff
@@ -734,6 +734,28 @@ int Beats::numBeatsInRange(audio::FramePos startPosition, audio::FramePos endPos
...
 };
 
+double Beats::numFractionalBeatsInRange(audio::FramePos startPos,
+        audio::FramePos endPos) const {
+    // find first beat in range (pos >= startPos)
+    audio::FramePos firstBeatPos = findNextBeat(startPos);
+
+    // maybe shift range to start at first beat
+    audio::FrameDiff_t firstBeatOffset = firstBeatPos - startPos;
+    endPos += firstBeatOffset;
```
daschuer, line 744: This only works if the beats have an equal length. For details around a position you may use: 
`BpmControl::getBeatContext()`. This way you get the beat fraction from before and after the fist and last beat. 
We may move this function to this class?

daschuer (changes requested):
Thank you. I have left suggestions.

ronso0:
Thanks for you review!

This is appearantly not ready since
a) there is some issue with `loop_remove` when a track is loaded to multiple players
b) I ran into a debug assert in CueControl after seeking and reactivating a saved loop

src/engine/controls/loopingcontrol.cpp
ronso0: This branch is for when quantize is enabled, we don't need fractional lengths then.

src/track/beats.cpp
ronso0: That function is helpful, thanks!
Yes, that (and `BpmControl::getBeatContextNoLookup`) should be in the `Beats` class.

Anyhow, the current implementation works well with constant BPM.
For tracks with variable BPM, the fractional size is just an estimation. Though IMO that doesn't matter since with var. BPM tracks `loop_halve/_double` aren't working as expected anyway AFAICT: the loop is simply scaled based on loop_in position and number of beats, i.e. it doesn't consider shorter/longer beats inside or after the loop.

daschuer: Ok I understand. 

What could be a real live use case? 
A beat loop with an integer number moved by a half beat and a tempo change inside a loop. 

Which beat length is used for moving? Probably the first beat? After moving, the loop will start at the middle of the fist beat and end somewhere else in the last beat. 
I think the loop size should be still without a fraction, right? 

What happens if the loop is moved fully in the new tempo region? 

We need to test this.

ronso0:
> b) I ran into a debug assert in CueControl after seeking and reactivating a saved loop

This happens when re-activating a saved loop for example, and it's a round-trip issue that becomes visible with this fix.
It seems avoiding the round-trip offset is close to impossible without a major refactoring, but it can be dealt with by using `LoopingControl::positionNear(a, b)` in the DEBUG_ASSERT (like in other places) instead of exact `==`.

Puuh, I already feared I was poking into a rats nest with this small fix.

The `loop_remove` issue however is still unclear to me, I don't understand why my changes cause that.

ronso0:
Alright, I moved some helpers to more suitable classes and picked 1dafd349742e2af4422478b941ed677820d1f0de.

The actual loop size fix is on top of it and it's now much simpler and more accurate using `Beats::getContext()`. Thanks @daschuer for the hint!

"Good" news: the `loop_remove` issue exists also in 2.4 so that's not blocking this PR.
I'll look into it (I fear it never worked as intended in the first place..)

___
**Though**
this fix reveals an inconsistency:
* set a loop manually with the IN/OUT buttons
   * loop size spinbox is not updated
   * halve/double don't work
   * save that loop to a hotcue
   = halve/double **work**

and

* set a loop manually with the IN/OUT buttons
   * halve/double **don't** work
   * eject track, reload
   * loop is restored
   = halve/double **work**

I'm not sure why it was decided to **not** update the spinbox with the manual loop's length and thereby **not** make halve/double work, but there are even tests for that
[LoopDoubleButton_DoesNotResizeManualLoop](https://github.com/mixxxdj/mixxx/blob/ee1263b0db8e4c55acb8867608da292927ec09af/src/test/looping_control_test.cpp#L490-L503)
[LoopHalveButton_DoesNotResizeManualLoop](https://github.com/mixxxdj/mixxx/blob/ee1263b0db8e4c55acb8867608da292927ec09af/src/test/looping_control_test.cpp#L541-L554)

After skimming #1187 I assume this behaviour should preserve (standart pre-set) loop sizes so `beatloop_activate` can be used for new loops right away, without having to reset the loop size to some default 2^n size? Same when loading a new track.
One way around this (for new tracks) would be to reset the loop size to the default 4 beats if the track has no loop set. However, this seems complicated considering how loops and cues are read when loading a track.

From the top of my head I have two real-life use case where it would have been helpful:
**1)** loop constant BPM sections of tracks where you can't use integer/preset loops because beats are wrong, e.g. for long mixes with variable BPM but constant section.
**2)** Mix a part of a track with no/useless beats (vocals, noise ...) with track2 by pressing IN and OUT in sync with the beats of track2. Admittedly, this will be off beat soon but it's good enough for short transitions.

Forgive me opening this can of worms so soon before the release ; )

src/track/beats.cpp
ronso0: Loops that contain tempo changes are moved correctly: both loop_in and loop_out are moved to the correct fractional position. This works with all loop sizes. :+1:
I tested with 4 and 3.5 beats loop moved by 1/2 and 8 beats several times while the tempo changed from ~110 to ~140.

src/audio/frame.h
```diff
@@ -91,6 +91,16 @@ class FramePos final {
         return util_isfinite(m_framePosition);
     }
 
+    // returns true if a is valid and is fairly close to target (within +/- 1 frame).
+    bool isNear(mixxx::audio::FramePos target) const {
+        VERIFY_OR_DEBUG_ASSERT(isValid()) {
+            return false;
+        }
+        return target.isValid() &&
```
daschuer, line 99: Can this isValid() also an debug assert?

daschuer (commented):
Thank you for the update. 
I need time for testing.

src/audio/frame.h
ronso0: IMO the caller is responsible for testing the position. It may be possible this is called with an invalid FramePos and in that case we simply return `false`.

src/audio/frame.h
```diff
@@ -91,6 +91,16 @@ class FramePos final {
         return util_isfinite(m_framePosition);
     }
 
+    // returns true if a is valid and is fairly close to target (within +/- 1 frame).
```
ronso0, line 94:
```suggestion
    // returns true if this and target are valid and target is fairly close (within +/- 1 frame).
```

ronso0 (commented):
reminder

daschuer:
I have tested with a changing tempo track and have these issues: 
* the spin box jumps to rather random value like 4.01 or 5.07 after loading track, even there is no loop set. (Reloop does nothing)
* When I manual edit the value, and than press the beat loop button the old value is applied, because the beatloop still has the input focus. The Enter key works. 
* The same issue is applies also to the beat jump box. 
* When I change the loop length using loop-out the spin-box is not updated.  (intended?) 
* This manual edited loop changes length but not beat size when beat moved, OK. 
* Reloop cycle does not change the loop size. (intended?)  
* New loops with Loop-In and Out do not change the loop size box (intended?)

ronso0:
> the spin box jumps to rather random value like 4.01 or 5.07 after loading track, even there is no loop set. (Reloop does nothing)

This may be due to invalid loop cues. Can you verify it occurs with new tracks? (new in the db, for example a track copy)

ronso0:
> When I manual edit the value, and than press the beat loop button the old value is applied, because the beatloop still has the input focus. The Enter key works.
The same issue is applies also to the beat jump box.

It is required to press Return to confirm the new value.
I stumble over this repeatedly and find this not intuitive and quite annoying tbh. I took a look at getting the unconfirmed value from the deck's (loop) WBeatSpinbox via signals/slots, and while it seems possible it is quite complicated (or at least not time-safe since it's not syncronous).

ronso0:
> When I change the loop length using loop-out the spin-box is not updated. (intended?)

yes, see #12522.

ronso0:
> Reloop cycle does not change the loop size. (intended?)

Not sure what you mean.

daschuer:
Here my retest results: 
 
> * the spin box jumps to rather random value like 4.01 or 5.07 after loading track, even there is no loop set. (Reloop does nothing)

Not able to reproduce it exactly. Currently the spinbox does not change. If this is a fractional position from the previous deck, you have hard times to use the double/half buttons in a reasonable way. How about just resetting it to a reasonable default? 

 > * When I manual edit the value, and than press the beat loop button the old value is applied, because the beatloop still has the input focus. The Enter key works.
> * The same issue is applies also to the beat jump box. 

Known issues . do we have a bug for it?  
 
> * When I change the loop length using loop-out the spin-box is not updated.  (intended?)
> * Reloop cycle does not change the loop size. (intended?)   

Yes, both seems to be the desired behaviour. Unfortunately it is inconsistent with the "feature" of showing the loop size at start up. Did you consider to not adopt non integer beat fractions after loading? This will also solve the first bullet point would feel more consistent.  

> * New loops with Loop-In and Out do not change the loop size box (intended?) 

Yes. 

Discovered another issue: 
* Load a track with a stored loop in hot-cue, but no loop in marker set. 
* Reloop button starts from the very beginning.
* This was one time issue.  
Reproducible: 
* Remove a loop
* Seek behind a stored loop
* Press loop out -> nothing happens = OK
* Eject the track 
* Reload the track 
* Seek behind a stored loop
* Press loop out -> track is looping from the stored loop-in poition. 

Conclusion: There is something fishy with reloading a track after a loop has been removed before. Probably thecause of the glitches I have experienced earlier.

ronso0:
> If this is a fractional position from the previous deck, you have hard times to use the double/half buttons in a reasonable way. How about just resetting it to a reasonable default?

Yes, this is part of #12522 

It seems there are a few issues with loop reset that were noticeable preiously, for example 560609898bfd2ec2c3a9a4f376db3d71c586cd32

> Did you consider to not adopt non integer beat fractions after loading? This will also solve the first bullet point would feel more consistent.

I'd even vote for not adopting any loop size after track load. I don't understand why loop halve/double should immediatly affect the existing loop (temp or saved), that's unhandy if I just want to preprare a new loop.
We can keep the previous loop size if it is in the list of 2^n sizes, else reset to the default value (4).

daschuer:
Sounds good.

daschuer:
How is the state here? I will put it to draft. Please adjust if this is wrong.

ronso0:
Yeah I've put this on hold already anyway, currently I'm lacking the motivation to get this ready for 2.5
Also I have the impression (vaguely remember) that we did not yet come to a conclusion regarding the optimal UX, related: #12522

I'll pick this up at some point after 2.5 is out.

PR #12470, opened 2023-12-29
Clamp offset in VisualPlayPosition with extra time of two vsync interval

daschuer:
It turns out that this is required to not clamp in all good conditions. Clamping causes a jittery waveform and must not happen under normal conditions. 

This removes one cause of jittery waveforms, during my tests. All other tests where successful with reasonable values. So I guess that the remaining jitters are real vsync misses. 

The clamping has been introduced to not move the waveform with a stalled audio engine. This still works, but is done one vsync cycle later.

PR #12516, opened 2024-01-05
Improved loging in case of  clamping visual position

daschuer:
It turns out that the log spam discovered by @m0dB in https://github.com/mixxxdj/mixxx/pull/12470 is actually a bug. 

This PR improves the log output and the negative side that is not a transport issue but a vsync issue.

It happens if the rendere requests a position of an already missed vsync cycle. 

A solution is to detect that and advance to the next vsync cycle that has not yet passed.

### DIFF
--- a/src/waveform/visualplayposition.cpp
+++ b/src/waveform/visualplayposition.cpp
@@ -70,9 +70,9 @@ double VisualPlayPosition::calcOffsetAtNextVSync(
         const int maxOffset = static_cast<int>(
                 data.m_audioBufferMicroS + 2 * syncIntervalTimeMicros);
 
-        // The negative offset is limited to -data.m_callbackEntrytoDac to not
-        // More negative values indicated an outdated request, that is anyway no
-        // longer valid. Probably cause bay a vsync issue.
+        // The minimum offset is limited to -data.m_callbackEntrytoDac to avoid a more
+        // negative value indicating an outdated request that is no longer valid anyway.
+        // This is probably caused by a vsync problem.
         const int minOffset = -data.m_callbackEntrytoDac;
 
         // Calculate the offset in micros for the position of the sample that will be transferred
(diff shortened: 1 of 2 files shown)


daschuer:
The problematic function is 

https://github.com/daschuer/mixxx/blob/f610d58a25b6f97172b097fb2a7a79147d863f7d/src/waveform/vsyncthread.cpp#L188

Here we need to advance to the next interval if we have missed the vsync interval the values are prepared for. 

@m0dB 
What was the reasoning for the combined sync and render step in the PLL case. I have good results with the combined timer and pll version when I start the timers after swap. 

I think it depends on the drivers implementation of vsync. When I remember correctly, the rendering waits to not introduce tearing for two reasons: 
1. Swap is delayed just after the previous picture has been displayed. 
2. rendering waits, when both buffers are occupied. One is in use for display, the other has already a pending swap. 

Maybe macOS has a workaround for 1.  and constantly hits 2. 

I have seen both with different drivers. 

Idea: Maybe we can use the PLL everwhere and just make the sync moment an option? 

Next I will contribute a PR that moves the position request into the right interval in case of vsync misses.

daschuer:
Requesting the right position was easy, I have added it here. Even the free mode is now working properly with ~500 Hz.

src/waveform/visualplayposition.cpp
```diff
@@ -64,16 +64,22 @@ double VisualPlayPosition::calcOffsetAtNextVSync(
...
-        // The offset is limited to the audio buffer + 2 x waveform sync interval
+        // The positive offset is limited to the audio buffer + 2 x waveform sync interval
         // This should be sufficient to compensate jitter, but does not continue
         // in case of underflows.
         const int maxOffset = static_cast<int>(
                 data.m_audioBufferMicroS + 2 * syncIntervalTimeMicros);
+
+        // The negative offset is limited to -data.m_callbackEntrytoDac to not
+        // More negative values indicated an outdated request, that is anyway no
+        // longer valid. Probably cause bay a vsync issue.
```
ronso0, line 75: Is there something missing in the first line?
Please fix the typos.

daschuer:
Done.

ywwg (approved):
works fine on my machine

m0dB:
Works fine here too, with both ST_TIMER and ST_PLL.

JoergAtGithub:
Works for me too! Thank you!

PR #4802, opened 2022-06-14
WTrackMenu: enable Reset Loops action

ronso0:
Requested on the forums + https://bugs.launchpad.net/mixxx/+bug/1978507

"Reset loops" was already in place but deactivated (though included in "Reset All" :| )
In **2.3** this does nothing as the FIXME says, in **main** it clears the saved loop cues -- though the last active loop persists, no matter if it had a loop cue attached.

- [x] show loop reset action in track menu (remove loop cues),
extent to clear the last active loop (markers removed, `reloop` detached)
- [x] add `loop_remove` control (saved loops are not touched)

Should we distinguish between saved loops and the current active loop?
The latter is requested on the forums now & then, though users are not aware of saved loops in 2.4, and with that there's at least a workaround to clear the unsaved loop (save, then delete loop cue)

PR #12521, opened 2024-01-06
fix `loop_remove`

ronso0:
This fixes two bugs with the loop cue (temp loop not attached to a hotcue):

**1)**
Appearantly `loop_remove` #4802 did only work for the first loop that was ever created for a track.
If a track already has a loop cue that was not cleared.
(and it seems Track menu > Reset > Loops only clears the temp. loop because Track emits cuesUpdated())

~~I fixed this in `BaseTrackPlayerImpl::unloadTrack()` rather than in `LoopingControl::slotLoopRemove()`, simply because that function already reads the loop positions and the loop cue.~~
Nope, that's not sufficient, clear it in `slotRemoveLoop` so it is not restored when the unsaved track is be loaded to another player.

**2)**
Currently, when an existing temp. loop is invalidated (set loop_in after loop_out) the loop cue is not cleared, i.e. the previous temp loop is restored next time the track is loaded.

___

The third commit changes LateNight so this can be tested easily. You may put that on top of 2.4 to confirm the bugs:
**1)**
* load a track that has a loop set (or load a track without loop, set a loop, eject, reload)
* click the Slip button (or trigger `loop_remove` otherwise)
= loop is removed
* eject
* reload
= loop is restored

**2)**
* load a track that has a loop set (or load a track without loop, set a loop, eject, reload)
* set loop_in after loop_out
= loop is removed
* eject
* reload
= previous loop is restored
Draft until that test commit is removed.

### DIFF
--- a/src/engine/controls/loopingcontrol.cpp
+++ b/src/engine/controls/loopingcontrol.cpp
@@ -779,6 +779,21 @@ void LoopingControl::slotLoopRemove() {
     m_loopInfo.setValue(loopInfo);
     m_pCOLoopStartPosition->set(loopInfo.startPosition.toEngineSamplePosMaybeInvalid());
     m_pCOLoopEndPosition->set(loopInfo.endPosition.toEngineSamplePosMaybeInvalid());
+    // The loop cue is stored by BaseTrackPlayerImpl::unloadTrack()
+    // if the loop is valid, else it is removed.
+    // We remove it here right away so the loop is not restored
+    // when the track is loaded to another player in the meantime.
+    auto pLoadedTrack = getEngineBuffer()->getLoadedTrack();
+    if (!pLoadedTrack) {
+        return;
+    }
+    const QList<CuePointer> cuePoints = pLoadedTrack->getCuePoints();
+    for (const auto& pCue : cuePoints) {
+        if (pCue->getType() == mixxx::CueType::Loop && pCue->getHotCue() == Cue::kNoHotCue) {
+            pLoadedTrack->removeCue(pCue);
+            return;
+        }
+    }
 }
 
 void LoopingControl::slotLoopIn(double pressed) {
--- a/res/skins/LateNight/decks/deck_settings.xml
+++ b/res/skins/LateNight/decks/deck_settings.xml
@@ -31,10 +31,10 @@
             <Layout>horizontal</Layout>
             <Children>
 
-              <Template src="skins:LateNight/controls/button_2state.xml">
+              <Template src="skins:LateNight/controls/button_1state.xml">
                 <SetVariable name="TooltipId">slip_mode</SetVariable>
                 <SetVariable name="ObjectName">SlipmodeButton<Variable name="DeckGroup"/></SetVariable>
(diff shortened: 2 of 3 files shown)


ywwg:
thank you for this

src/engine/controls/loopingcontrol.cpp
```diff
@@ -779,6 +779,21 @@ void LoopingControl::slotLoopRemove() {
...
+    // if the loop is valid, else it is removed.
+    // We remove it here right away so the loop is not restored
+    // when the track is loaded to another player in the meantime.
+    auto pLoadedTrack = getEngineBuffer()->getLoadedTrack();
+    if (!pLoadedTrack) {
+        return;
+    }
+    const QList<CuePointer> cuePoints = pLoadedTrack->getCuePoints();
+    for (const auto& pCue : cuePoints) {
+        if (pCue->getType() == mixxx::CueType::Loop && pCue->getHotCue() == Cue::kNoHotCue) {
```
daschuer, line 792: The second condition is unclear for me. Is it required? Can it be an assertion? 

The same below.

daschuer (commented):
Just a first comment, the manual test is pending.

daschuer:
It works, thank you. 

The reloop button does not work, clear. I pressed it expecting an undo feature without thinking, because in case of a saved loop, there is not removing the in and out markers is barely visible.  
Is this an issue?

ronso0:
> I pressed it expecting an undo feature without thinking, because in case of a saved loop, there is not removing the in and out markers is barely visible.
Is this an issue?

I don't quite understand, can you elaborate please?

Thing is, `loop_remove` is only available via the control (there's no GUI button in the official skins¹), so I assume it'll barely be used anyway until we add that as an action to the track menu, so I doubt anyone expects an undo feature.

¹And tbh I don't know where to put it. Loop, reloop, In and Out already have a right-click connection.

daschuer:
OK, so let's merge. Thank you.

ronso0:
And the TEST commit slipped in again.
Now reverting it, and merging 2.4 into main.

What's the best strategy to avoid this in the future?
**a)** leave in draft state until fixups are squashed / TEST commits are removed, and ping reviewers repeatedly to get attention
**b)** set the 'changes requested' flag and hope it (the reason) is not overlooked before merging

What do you think?

daschuer:
Oh sorry. 
How about setting up a second Draft branch with the test commit? This will also release you from the work of touching the branch after review. 
The 'changes requested' will probably also work.

ronso0:
> Draft branch with the test commit? This will also release you from the work of touching the branch after review.

I see your point, but that requires rebasing the TEST branch in case the base branch receives fixes, e.g. applying requested changes, and reviewers can not be sure they the TEST branch is in sync.

Therefore I'd prefer 'Changes requested'. The 'Compare' button after force-push is our friend for this case: it's easy to verify the force-push only reverted skin changes / removed debug output / fixed typos etc.

daschuer:
It is finally up to you. I personally consider it more work to rebase after review compared to keeping two branches during development, just because the developer and the reviewer need to find an extra time slot after the successful review.

PR #12522, opened 2024-01-06
update beatloop size for loops set manually with IN/OUT

ronso0:
This updates the `beatloop_size` for loops set manually with IN/OUT, which also enables `loop_halve`/`_double`.
**This is on top of #12509, so check only the last two commits.**

That was discussed in #1187 and rejected for some reasons (or rather concerns and assumptions IMO).
I disabled the respective tests.

Fixing #10511 in #12509 made the previous inconsistency obvious (https://github.com/mixxxdj/mixxx/pull/12509#issuecomment-1879126230).

The second commit tries to address the concerns raised in #1187:
When a track with an unusual loop size is ejected, that loop size is kept (now not only for saved loops but also for temp. loops).
If a track with no loop is loaded, beatloop_size is reset to default (4 beats), or, if there's a saved loop, its size is adopted.

What do you think?
(I know it's a bit late for such a change)

### DIFF
--- a/src/audio/frame.h
+++ b/src/audio/frame.h
@@ -121,6 +121,16 @@ class FramePos final {
         return util_isfinite(m_framePosition);
     }
 
+    // returns true if a is valid and is fairly close to target (within +/- 1 frame).
+    bool isNear(mixxx::audio::FramePos target) const {
+        VERIFY_OR_DEBUG_ASSERT(isValid()) {
+            return false;
+        }
+        return target.isValid() &&
+                value() > target.value() - 1.0 &&
+                value() < target.value() + 1.0;
+    };
+
     void setValue(value_t framePosition) {
         m_framePosition = framePosition;
     }
--- a/src/audio/frame.h
+++ b/src/audio/frame.h
@@ -120,6 +120,16 @@ class FramePos final {
         return util_isfinite(m_framePosition);
     }
 
+    // returns true if a is valid and is fairly close to target (within +/- 1 frame).
+    bool isNear(mixxx::audio::FramePos target) const {
+        VERIFY_OR_DEBUG_ASSERT(isValid()) {
+            return false;
+        }
+        return target.isValid() &&
+                value() > target.value() - 1.0 &&
+                value() < target.value() + 1.0;
+    };
+
     void setValue(value_t framePosition) {
         m_framePosition = framePosition;
     }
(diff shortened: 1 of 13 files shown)


ywwg:
Yeah this is a tough one -- it feels like a lot of risky change so close to release.  I feel like our test coverage is pretty good for beatloop size stuff though?   In my opinion, given the bugs you are describing, they seem pretty minor and cosmetic.  There is some annoyance, but the current experience is not that bad.  Whereas, if there are bugs introduced here, we risk breaking loops which is a major feature.

My position would be to save this fix for the next patch release.

ronso0:
You're right, this should go to 2.4.1
I'm pretty confident though this doesn't break anything, actually it makes workflows consistent. I started with #12509 to fix a minor bug but this grew a bit beyond a small bugfix due to the consequneces / other bug it revealed.

ronso0:
It is not clear to me why [`LoopingControlTest, LoopInOutButtons_QuantizeEnabled`](https://github.com/mixxxdj/mixxx/blob/e3ba718b77b3f6ddfbfa734db1b33049fa0518dc/src/test/looping_control_test.cpp#L325-L359) fails with this change.

Furthermore, if I do the test steps manually **with 2.4** I don't get the expected result:
* enable quantize
* set playpos slightly after a beat
* set IN
= IN snaps to beat, playpos is unchanged :heavy_check_mark:  
* move forward by 4 beats
= playpos is slightly after the beat :heavy_check_mark: 
* set OUT
= OUT snaps to beat :heavy_check_mark: 
= playpos jumps to loop_in :x:  (expected: playpos remains after the beat)

ronso0:
> I feel like our test coverage is pretty good for beatloop size stuff though?

Yes. And no, see https://github.com/mixxxdj/mixxx/pull/12522#issuecomment-1879865194
I assume the test succeed because no visuals are involved?

Can anyone confirm my observation, or does it behave as expected?

ronso0:
> The second commit tries to address the concerns raised in https://github.com/mixxxdj/mixxx/pull/1187:
When a track with an unusual loop size¹ is ejected, that loop size is kept (now not only for saved loops but also for temp. loops).
If a track with no loop is loaded, beatloop_size is reset to default (4 beats), or, if there's a saved loop, its size is adopted.
¹a loop size that's not in the list of 2^n sizes

and quoting from #12509 
I'd even vote for _not adopting any loop_ size after track load. I don't understand why loop halve/double should immediatly affect the existing loop (temp or saved), that's unhandy if I just want to preprare a new loop.
We can keep the previous loop size if it is in the list of 2^n sizes, else reset to the default value (4).

ronso0:
Rebased onto main, will do manual test soonish.

Issue #12451, opened 2023-12-19
Can’t use microphone talkover mode with compressor effect


fonsargo:
### Bug Description

I’m trying to use compressor effect for music (like auto gain control), in order to have almost maximum music level for each song (almost 0 DB). But when I try to talk something over music, I can’t decrease music level. I’ve set _Microphone talkover ducking strength_ to low level and when I turn on my microphone it works only for a few seconds. After that my compressor plugin increase music level back to 0 DB and music becomes louder than my voice.

As I understood, microphone ducking is applied before any effects. I’ve tried to apply my effects for each Deck and for Main mixer, but the problem is the same. Every time my effect neutralizes microphone ducking (talkover mode). Applying it to only the main mix gives the same result. Moreover, when I chose in hardware settings _Microphone monitor mode_ to _Main output only_, my effect isn’t applied to microphone input (however when I apply some effect to the main mix I expect to apply it to all sound going through the main mix).

So it seems to me that microphone ducking (volume decrease) is applied BEFORE effects panel, but microphone input is mixed AFTER effects of the main mix, which looks quite strange. And I can’t find any settings related to it in order to change this behavior.

### Version

2.3.6, 2.4.0-beta

### OS

Windows 11

fonsargo:
Does anyone have any ideas how it could be fixed? I guess microphone ducking should be applied to the main output gain, not to the deck output. However I don't have a clue how the Mixxx is actually mixing decks and microphone together. Maybe someone has some links to the starting point in code?

ronso0:
The mixing is happening in `EngineMixer`, somewhere around here I think https://github.com/mixxxdj/mixxx/blob/1f85c9c4502e42e744a1797c85c2b1343f5d3609/src/engine/enginemixer.cpp#L535

Issue #6453, opened 2022-08-22
compressor effect


mixxxbot:
Reported by: **[Pegasus-RPG](https://github.com/Pegasus-RPG)**
Date: 2012-05-17T13:51:48Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1000759](https://bugs.launchpad.net/bugs/1000759)
Tags: effects, microphone

---

(Testing with a Shure SM58 through an SCS.1m firewire sound card via JACK on Debian Squeeze.)

At first glance, the microphone input seems to be too quiet, but because it's possible to max out the level in the software (though the clip indicator never lights) and clip the hardware input just by yelling into it, we should add a compressor to the input stage so that the voice can be substantially louder without the user having to turn the gain so high on the hardware.

fwcd:
Closing this since #12523 has been implemented and merged. More specific requests (say a multiband compressor[^1]) can be tracked in separate issues.

[^1]: This one seems to already be tracked in #5994

PR #12523, opened 2024-01-06
Compressor effect

fonsargo:
This is a simple implementation of general single-band compressor, which was requested in this feature request #6453 years ago.
I'm not an expert of compressors, so I've just used general information about it from the Internet, mainly from this readme: https://github.com/p-hlp/CTAGDRC
It's the simpliest solution with gain computer and ballistics, without parameters automation. However in most cases it will be enough, for example, for radio broadcasting.
The only automation I did is the Auto makeup. I haven't found some good example of it, so I tried to introduce some "homemade" algorithm.

I've been using custom build with this effect for radio broadcasting for 2 weeks and it looks good. However, due to this bug #12451 I have some limitations of using this compressor.

Also I'm not a C++ developer, I've been working with Java for years, so there may be some stupid mistakes which I apologize for in advance :)

### DIFF
--- a/src/effects/backends/builtin/compressoreffect.cpp
+++ b/src/effects/backends/builtin/compressoreffect.cpp
@@ -108,8 +108,8 @@ EffectManifestPointer CompressorEffect::getManifest() {
 CompressorGroupState::CompressorGroupState(
         const mixxx::EngineParameters& engineParameters)
         : EffectState(engineParameters),
-          previousMakeUpGain(0),
-          previousStateDB(0) {
+          previousStateDB(0),
+          previousMakeUpGain(0) {
 }
 
 void CompressorEffect::loadEngineEffectParameters(
--- a/src/effects/backends/builtin/compressoreffect.cpp
+++ b/src/effects/backends/builtin/compressoreffect.cpp
@@ -17,16 +17,6 @@ EffectManifestPointer CompressorEffect::getManifest() {
     pManifest->setEffectRampsFromDry(true);
     pManifest->setMetaknobDefault(0.0);
 
-    EffectManifestParameterPointer clipping = pManifest->addParameter();
-    clipping->setId("clipping");
-    clipping->setName(QObject::tr("Clipping"));
-    clipping->setShortName(QObject::tr("Clipping"));
-    clipping->setDescription(QObject::tr("Hard limiter to full scale"));
-    clipping->setValueScaler(EffectManifestParameter::ValueScaler::Toggle);
-    clipping->setRange(0, 1, 1);
-    clipping->appendStep(qMakePair(QObject::tr("Off"), static_cast<int>(Clipping::ClippingOff)));
-    clipping->appendStep(qMakePair(QObject::tr("On"), static_cast<int>(Clipping::ClippingOn)));
-
     EffectManifestParameterPointer autoMakeUp = pManifest->addParameter();
     autoMakeUp->setId("automakeup");
     autoMakeUp->setName(QObject::tr("Auto Makeup Gain"));
@@ -139,7 +129,6 @@ void CompressorEffect::loadEngineEffectParameters(
     m_pRelease = parameters.value("release");
     m_pGain = parameters.value("gain");
     m_pAutoMakeUp = parameters.value("automakeup");
-    m_pClipping = parameters.value("clipping");
 }
 
 void CompressorEffect::processChannel(
@@ -165,11 +154,6 @@ void CompressorEffect::processChannel(
     // Output gain
     CSAMPLE gain = static_cast<CSAMPLE>(db2ratio(m_pGain->value()));
     SampleUtil::applyGain(pOutput, gain, numSamples);
-
-    // Clipping
-    if (m_pClipping->toInt() == static_cast<int>(Clipping::ClippingOn)) {
-        SampleUtil::applyClamp(pOutput, numSamples);
-    }
 }
 
 void CompressorEffect::applyAutoMakeUp(CompressorGroupState* pState,
--- a/src/effects/backends/builtin/compressoreffect.cpp
+++ b/src/effects/backends/builtin/compressoreffect.cpp
@@ -1,0 +1,280 @@
+#include "effects/backends/builtin/compressoreffect.h"
+#include <Windows.h>
+
+// static
+QString CompressorEffect::getId() {
+    return "org.mixxx.effects.compressor";
+}
+
+//static
+EffectManifestPointer CompressorEffect::getManifest() {
+    EffectManifestPointer pManifest(new EffectManifest);
+    pManifest->setId(getId());
+    pManifest->setName(QObject::tr("Compressor"));
+    pManifest->setShortName(QObject::tr("Compressor"));
+    pManifest->setAuthor("The Mixxx Team");
+    pManifest->setVersion("1.0");
+    pManifest->setDescription(
+            "A single-band compressor effect");
+    pManifest->setEffectRampsFromDry(true);
+    pManifest->setMetaknobDefault(0.0);
+
+    EffectManifestParameterPointer clipping = pManifest->addParameter();
+    clipping->setId("clipping");
+    clipping->setName(QObject::tr("Clipping"));
+    clipping->setShortName(QObject::tr("Clipping"));
+    clipping->setDescription(QObject::tr("Hard limiter to 0 db"));
+    clipping->setValueScaler(EffectManifestParameter::ValueScaler::Toggle);
+    clipping->setRange(0, 1, 1);
+    clipping->appendStep(qMakePair(QObject::tr("Off"), Clipping::ClippingOff));
+    clipping->appendStep(qMakePair(QObject::tr("On"), Clipping::ClippingOn));
+
+    EffectManifestParameterPointer autoMakeUp = pManifest->addParameter();
+    autoMakeUp->setId("automakeup");
+    autoMakeUp->setName(QObject::tr("Auto Make Up Gain"));
+    autoMakeUp->setShortName(QObject::tr("Make Up"));
+    autoMakeUp->setDescription(QObject::tr(
+            "Auto make up gain to 0 db level"));
+    autoMakeUp->setValueScaler(EffectManifestParameter::ValueScaler::Toggle);
+    autoMakeUp->setRange(0, 1, 1);
+    autoMakeUp->appendStep(qMakePair(QObject::tr("Off"), AutoMakeUp::AutoMakeUpOff));
+    autoMakeUp->appendStep(qMakePair(QObject::tr("On"), AutoMakeUp::AutoMakeUpOn));
+
+
+    //TODO description
+    EffectManifestParameterPointer threshold = pManifest->addParameter();
+    threshold->setId("threshold");
+    threshold->setName(QObject::tr("Threshold (dB)"));
+    threshold->setShortName(QObject::tr("Threshold"));
+    threshold->setDescription(QObject::tr(
+            "The amount of amplification "
+            "applied to the audio signal. At higher levels the audio will be more distored."));
+    threshold->setValueScaler(EffectManifestParameter::ValueScaler::Linear);
+    //threshold->setDefaultLinkType(EffectManifestParameter::LinkType::Linked);
+    threshold->setUnitsHint(EffectManifestParameter::UnitsHint::Decibel);
+    threshold->setNeutralPointOnScale(0);
+    threshold->setRange(-50, -20, 0);
+
+    EffectManifestParameterPointer ratio = pManifest->addParameter();
+    ratio->setId("ratio");
+    ratio->setName(QObject::tr("Ratio (:1)"));
+    ratio->setShortName(QObject::tr("Ratio"));
+    ratio->setDescription(QObject::tr(
+            "The amount of ratio."));
+    ratio->setValueScaler(EffectManifestParameter::ValueScaler::Linear);
+    //ratio->setDefaultLinkType(EffectManifestParameter::LinkType::Linked);
+    ratio->setUnitsHint(EffectManifestParameter::UnitsHint::Coefficient);
+    ratio->setNeutralPointOnScale(0);
+    ratio->setRange(1.0, 4.0, 20);
+
+    EffectManifestParameterPointer knee = pManifest->addParameter();
+    knee->setId("knee");
+    knee->setName(QObject::tr("Knee (dB)"));
+    knee->setShortName(QObject::tr("Knee"));
+    knee->setDescription(QObject::tr(
+            "The amount of knee."));
+    knee->setValueScaler(EffectManifestParameter::ValueScaler::Linear);
+    //knee->setDefaultLinkType(EffectManifestParameter::LinkType::Linked);
+    knee->setUnitsHint(EffectManifestParameter::UnitsHint::Coefficient);
+    knee->setNeutralPointOnScale(0);
+    knee->setRange(0.0, 4.0, 24);
+
+    EffectManifestParameterPointer attack = pManifest->addParameter();
+    attack->setId("attack");
+    attack->setName(QObject::tr("Attack (ms)"));
+    attack->setShortName(QObject::tr("Attack"));
+    attack->setDescription(QObject::tr(
+            "Attack"));
+    attack->setValueScaler(EffectManifestParameter::ValueScaler::Integral);
+    attack->setUnitsHint(EffectManifestParameter::UnitsHint::Millisecond);
+    attack->setRange(0, 30, 250);
+
+    EffectManifestParameterPointer release = pManifest->addParameter();
+    release->setId("release");
+    release->setName(QObject::tr("Release (ms)"));
+    release->setShortName(QObject::tr("Release"));
+    release->setDescription(QObject::tr(
+            "Release"));
+    release->setValueScaler(EffectManifestParameter::ValueScaler::Integral);
+    release->setUnitsHint(EffectManifestParameter::UnitsHint::Millisecond);
+    release->setRange(0, 150, 2000);
+
+    EffectManifestParameterPointer gain = pManifest->addParameter();
+    gain->setId("gain");
+    gain->setName(QObject::tr("Make up gain"));
+    gain->setShortName(QObject::tr("Gain"));
+    gain->setDescription(QObject::tr(
+            "Gain"));
+    gain->setValueScaler(EffectManifestParameter::ValueScaler::Linear);
+    gain->setUnitsHint(EffectManifestParameter::UnitsHint::Decibel);
+    gain->setRange(-25, 0, 25);
+
+    return pManifest;
+}
+
+CompressorGroupState::CompressorGroupState(
+        const mixxx::EngineParameters& engineParameters)
+        : EffectState(engineParameters),
+          samplerate(engineParameters.sampleRate()),
+          previousMakeUpGain(-1),
+          previousStateDB(0) {
+}
+
+void CompressorEffect::loadEngineEffectParameters(
+        const QMap<QString, EngineEffectParameterPointer>& parameters) {
+    m_pThreshold = parameters.value("threshold");
+    m_pRatio = parameters.value("ratio");
+    m_pKnee = parameters.value("knee");
+    m_pAttack = parameters.value("attack");
+    m_pRelease = parameters.value("release");
+    m_pGain = parameters.value("gain");
+    m_pAutoMakeUp = parameters.value("automakeup");
+    m_pClipping = parameters.value("clipping");
+}
+
+void CompressorEffect::processChannel(
+        CompressorGroupState* pState,
+        const CSAMPLE* pInput,
+        CSAMPLE* pOutput,
+        const mixxx::EngineParameters& engineParameters,
+        const EffectEnableState enableState,
+        const GroupFeatureState& groupFeatures) {
+    Q_UNUSED(groupFeatures);
+    Q_UNUSED(enableState);
+
+    // TODO test: Marc Benjamin, Zana - Edge of Paradise; Cassette - Tell Me Why; Adele - Skyfall
+    // url: https://github.com/p-hlp/CTAGDRC
+
+    SINT numSamples = engineParameters.samplesPerBuffer();
+    int channelCount = engineParameters.channelCount();
+    SINT numFrames = engineParameters.framesPerBuffer();
+
+    //CSAMPLE max = SampleUtil::maxAbsAmplitude(pInput, numSamples);
+    //if (max == CSAMPLE_ZERO) {
+    //    SampleUtil::copy(pOutput, pInput, numSamples);
+    //    return;
+    //}
+
+    CSAMPLE thresholdParam = static_cast<CSAMPLE>(m_pThreshold->value());
+    CSAMPLE ratioParam = static_cast<CSAMPLE>(m_pRatio->value());
+    //CSAMPLE kneeParam = static_cast<CSAMPLE>(m_pKnee->value());
+    //CSAMPLE kneeHalf = kneeParam / 2.0f;
+    CSAMPLE gainParamDB = static_cast<CSAMPLE>(m_pGain->value());
+    CSAMPLE attackParam = m_pAttack->value();
+    CSAMPLE releaseParam = m_pRelease->value();
+    CSAMPLE attackCoeff = exp(-1000.0 / (attackParam * pState->samplerate));
+    CSAMPLE releaseCoeff = exp(-1000.0 / (releaseParam * pState->samplerate));
+    CSAMPLE makeUpCoeff = 0.03;
+
+    CSAMPLE stateDB = pState->previousStateDB;
+    CSAMPLE makeUpStateDB = pState->previousMakeUpGain;
+
+    std::vector<int> gains(numFrames);
+    CSAMPLE maxGain = thresholdParam;
+    for (SINT i = 0; i < numSamples; i += channelCount) {
+        CSAMPLE maxSample = std::max(fabs(pInput[i]), fabs(pInput[i + 1]));
+        if (maxSample == CSAMPLE_ZERO) {
+            gains[i / channelCount] = 0.0;
+            continue;
+        }
+
+        CSAMPLE maxSampleDB = ratio2db(maxSample);
+        CSAMPLE overDB = maxSampleDB - thresholdParam;
+        //if (overDB <= kneeHalf) {
+        //    overDB = 0.0f;
+        //} else if (overDB > -kneeHalf && overDB <= kneeHalf) {
+        //    overDB = 0.5f * (overDB + kneeHalf) * (overDB + kneeHalf) / kneeParam;
+        //}
+
+        // atack/release
+        if (overDB > stateDB) {
+            stateDB = overDB + attackCoeff * (stateDB - overDB);
+        } else {
+            stateDB = overDB + releaseCoeff * (stateDB - overDB);
+        }
+
+
+        overDB = stateDB;
+        CSAMPLE gainReductionDB = overDB * (1.0 / ratioParam - 1.0);
+        gains[i / channelCount] = gainReductionDB;
+        CSAMPLE totalGain = maxSampleDB + gainReductionDB;
+        if (totalGain > maxGain) {
+            maxGain = totalGain;
+        }
+    }
+
+    CSAMPLE autoMakeUp = 0;
+    if (m_pAutoMakeUp->toInt() == AutoMakeUpOn) { 
+        CSAMPLE minGainReductionDB = -maxGain - 3.0;
+        if (makeUpStateDB == -1) {
+            //std::string msg2 = std::string("makeUpStateDB: ") + std::to_string(makeUpStateDB) + std::string(" minGainReductionDB: ") + std::to_string(minGainReductionDB) + std::string("\n");
+            //OutputDebugStringA(msg2.c_str());
+            makeUpStateDB = minGainReductionDB;
+        }
+        makeUpStateDB = makeUpCoeff * minGainReductionDB + (1 - makeUpCoeff) * makeUpStateDB;
+        autoMakeUp = makeUpStateDB;
+    }
+
+    for (SINT i = 0; i < numSamples; i += channelCount) {
+        CSAMPLE gain = db2ratio(gains[i / channelCount] + gainParamDB + autoMakeUp);
+        pOutput[i] = pInput[i] * gain;
+        pOutput[i + 1] = pInput[i + 1] * gain;
+    }
+    
+    if (m_pClipping->toInt() == ClippingOn) {
+        SampleUtil::copyClampBuffer(pOutput, pOutput, numSamples);
+    }
+
+
+    pState->previousStateDB = stateDB;
+    pState->previousMakeUpGain = makeUpStateDB;
+    //std::string msg2 = std::string("makeUpStateDB: ") + std::to_string(makeUpStateDB) + std::string(" stateDB: ") + std::to_string(stateDB) + std::string("\n");
+    //OutputDebugStringA(msg2.c_str());
+
+    /*
+    if (m_pAutoMakeUp->toInt() == On) {
+         CSAMPLE sum = CSAMPLE_ZERO;
+        for (SINT i = 0; i < numSamples; i += channelCount) {
+            sum += std::max(fabs(pInput[i]), fabs(pInput[i + 1]));
+        }
+        CSAMPLE max = SampleUtil::maxAbsAmplitude(pOutput, numSamples);
+        if (max != CSAMPLE_ZERO) {
+            CSAMPLE averageDB = -ratio2db(max) - 2;
+            CSAMPLE makeUpSatetDB = pState->previousMakeUpGain;
+
+            makeUpSatetDB = averageDB + makeUpCoeff * (makeUpSatetDB - averageDB);
+            std::string msg = std::string("makeUpCoeff: ") + std::to_string(makeUpCoeff) + std::string(" averageDB: ") + std::to_string(averageDB) + std::string(" makeUpSatetDB: ") + std::to_string(makeUpSatetDB) + std::string("\n");
+            OutputDebugStringA(msg.c_str());
+
+            gainParamDB += averageDB;
+            pState->previousMakeUpGain = makeUpSatetDB;
+        }
+    }
+
+
+    SampleUtil::applyGain(pOutput, db2ratio(gainParamDB), numSamples);
+    */
+    
+
+   /* if (thresholdParam < 0.01) {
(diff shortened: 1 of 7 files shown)


JoergAtGithub:
Welcome at Mixxx!

fonsargo:

Done!

JoergAtGithub:
The Ubuntu builds and the pre-commit check are failing. All these CI checks needs to be pass.
The best way to fix pre-commit issues is to install pre-commit locally on your system, as described here: https://github.com/mixxxdj/mixxx/wiki/Using-Git#set-up-automatic-code-checking
grafik
unzip it, and apply it using:
`git apply pre-commit.patch`

src/effects/backends/builtin/compressoreffect.h
```diff
@@ -0,0 +1,72 @@
...
+#include "engine/effects/engineeffect.h"
+#include "engine/effects/engineeffectparameter.h"
+#include "util/class.h"
+#include "util/defs.h"
+#include "util/sample.h"
+#include "util/types.h"
+
+namespace {
+    constexpr double makeUpAttackCoeff = 0.03;
+    constexpr double makeUpTarget = -3.0;
```
Swiftb0y, line 13:
```suggestion
    constexpr CSAMPLE_GAIN kMakeUpAttackCoeff = 0.03;
    constexpr CSAMPLE_GAIN kMakeUpTarget = -3.0;
```

src/effects/backends/builtin/compressoreffect.h
```diff
@@ -0,0 +1,72 @@
...
+
+ private:
+
+    enum AutoMakeUp {
+        AutoMakeUpOff = 0,
+        AutoMakeUpOn = 1,
+    };
+
+    enum Clipping {
+        ClippingOff = 0,
```
Swiftb0y, line 50:
```suggestion
    enum class AutoMakeUp {
        AutoMakeUpOff = 0,
        AutoMakeUpOn = 1,
    };

    enum class Clipping {
        ClippingOff = 0,
```

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,215 @@
+#include "effects/backends/builtin/compressoreffect.h"
+
+// static
+QString CompressorEffect::getId() {
+    return "org.mixxx.effects.compressor";
+}
+
+//static
+EffectManifestPointer CompressorEffect::getManifest() {
+    EffectManifestPointer pManifest(new EffectManifest);
```
Swiftb0y, line 10:
```suggestion
    auto pManifest = EffectManifestPointer::create();
```

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,215 @@
...
+    gain->setRange(-25, 0, 25);
+
+    return pManifest;
+}
+
+CompressorGroupState::CompressorGroupState(
+        const mixxx::EngineParameters& engineParameters)
+        : EffectState(engineParameters),
+          previousMakeUpGain(0),
+          previousStateDB(0) {
```
Swiftb0y, line 112: reorder to fix compiler error
```suggestion,
          previousStateDB(0) 
          previousMakeUpGain(0) {
```

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    // Output gain
+    CSAMPLE gainParamDB = static_cast<CSAMPLE>(m_pGain->value());
+    SampleUtil::applyGain(pOutput, db2ratio(gainParamDB), numSamples);
+
+    // Clipping
+    if (m_pClipping->toInt() == ClippingOn) {
+        SampleUtil::copyClampBuffer(pOutput, pOutput, numSamples);
+    }
```
Swiftb0y, line 154: `SampleUtil::copyClampBuffer` can't be run in place (due to `-Wrestrict`). Feel free to create an inplace-clamp function to `SampleUtil`.

Swiftb0y (commented):
couple comments, thank you very much so far.

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,215 @@
...
+            "For a ratio of 4:1, one dB remains for every 4dB of input signal above the threshold. "
+            "At a ratio of 1:1 no compression is happening, as the input is exactly the output"));
+    ratio->setValueScaler(EffectManifestParameter::ValueScaler::Logarithmic);
+    ratio->setUnitsHint(EffectManifestParameter::UnitsHint::Coefficient);
+    ratio->setNeutralPointOnScale(0);
+    ratio->setRange(1.0, 4.0, 1000);
+
+    EffectManifestParameterPointer knee = pManifest->addParameter();
+    knee->setId("knee");
+    knee->setName(QObject::tr("Knee (dB)"));
```
daschuer, line 67: dB from what?

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,215 @@
...
+    pManifest->setVersion("1.0");
+    pManifest->setDescription("A single-band compressor effect");
+    pManifest->setEffectRampsFromDry(true);
+    pManifest->setMetaknobDefault(0.0);
+
+    EffectManifestParameterPointer clipping = pManifest->addParameter();
+    clipping->setId("clipping");
+    clipping->setName(QObject::tr("Clipping"));
+    clipping->setShortName(QObject::tr("Clipping"));
+    clipping->setDescription(QObject::tr("Hard limiter to 0 db"));
```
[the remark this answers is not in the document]
daschuer: Is that ever useful? A hard clipper will introduce additional frequencies.  Did you consider a tanh clipper, that is close to analogue circuit?

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,215 @@
...
+        SampleUtil::applyGain(pOutput, db2ratio(makeUpStateDB), numSamples);
+    }
+}
+
+void CompressorEffect::applyCompression(CompressorGroupState* pState, const mixxx::EngineParameters& engineParameters, const CSAMPLE* pInput, CSAMPLE* pOutput) {
+    CSAMPLE thresholdParam = static_cast<CSAMPLE>(m_pThreshold->value());
+    CSAMPLE ratioParam = static_cast<CSAMPLE>(m_pRatio->value());
+    CSAMPLE kneeParam = static_cast<CSAMPLE>(m_pKnee->value());
+    CSAMPLE kneeHalf = kneeParam / 2.0f;
+    CSAMPLE attackCoeff = exp(-1000.0 / (m_pAttack->value() * engineParameters.sampleRate()));
```
daschuer, line 180: I am afraid exp() is slow can we store and reuse the result?

daschuer (commented):
Thank you. I have left some comments. Not yet tested.

src/effects/backends/builtin/compressoreffect.cpp
Swiftb0y: alternatively, implement a good-enough approximation? see https://www.schraudolph.org/pubs/Schraudolph99.pdf

src/effects/backends/builtin/compressoreffect.cpp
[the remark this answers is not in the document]
fonsargo: Thank you, it's a good point. Actually I added it just to avoid sound going to "red" level, for example if there are some very fast peaks, which compressor didn't catch because of the slow attack time. And no, I haven't considered any other clippers yet.

Despite the fact that it's a very simple solution, I believe it's useful. We could introduce more advanced solution later, for example in this feature request: #6513.

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,215 @@
...
+    clipping->setRange(0, 1, 1);
+    clipping->appendStep(qMakePair(QObject::tr("Off"), Clipping::ClippingOff));
+    clipping->appendStep(qMakePair(QObject::tr("On"), Clipping::ClippingOn));
+
+    EffectManifestParameterPointer autoMakeUp = pManifest->addParameter();
+    autoMakeUp->setId("automakeup");
+    autoMakeUp->setName(QObject::tr("Auto Makeup Gain"));
+    autoMakeUp->setShortName(QObject::tr("Makeup"));
+    autoMakeUp->setDescription(QObject::tr(
+            "The AutoMakeup button enables automatic makeup gain to 0 db level"));
```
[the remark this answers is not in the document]
fonsargo: In fact this description isn't true, because this algorithm adjusts volume to 0 dBFS (actually around -3 dBFS) in anyways. So it doesn't matter, what was the original level, output will be almost 0 dBFS (even if original input was -30 dBFS for example).

src/effects/backends/builtin/compressoreffect.cpp
fonsargo: I think it's also dBFS

src/effects/backends/builtin/compressoreffect.cpp
fonsargo: Actually we can't store it, because user can change attack time in any time during playback.

I believe the compiler already has fast enough approximation, hasn't it? Because, for example, in Java world it isn't a problem, but I'm not sure about C++.

src/effects/backends/builtin/compressoreffect.cpp
fonsargo: Done

src/effects/backends/builtin/compressoreffect.cpp
daschuer: The compiler use the math exact method by default. 
You may check if the user values have changed. The comparison is fast.

src/effects/backends/builtin/compressoreffect.cpp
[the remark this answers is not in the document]
daschuer: OK, than you may add a true description?

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,236 @@
...
+    threshold->setRange(-50, -20, 0);
+
+    EffectManifestParameterPointer ratio = pManifest->addParameter();
+    ratio->setId("ratio");
+    ratio->setName(QObject::tr("Ratio (:1)"));
+    ratio->setShortName(QObject::tr("Ratio"));
+    ratio->setDescription(
+            QObject::tr("The Ratio knob determines how much the signal is "
+                        "attenuated above the chosen threshold. "
+                        "For a ratio of 4:1, one dB remains for every 4dB of "
```
daschuer, line 62:
```suggestion
                        "For a ratio of 4:1, one dB remains for every four dB of "
```

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,236 @@
...
+    CSAMPLE maxSample = SampleUtil::maxAbsAmplitude(pOutput, numSamples);
+    if (maxSample > 0) {
+        CSAMPLE maxSampleDB = ratio2db(maxSample);
+        CSAMPLE minGainReductionDB = -maxSampleDB + kMakeUpTarget;
+        makeUpStateDB = kMakeUpAttackCoeff * minGainReductionDB +
+                (1 - kMakeUpAttackCoeff) * makeUpStateDB;
+        CSAMPLE levelDB = makeUpStateDB + maxSampleDB;
+        // logarithmic smoothing
+        if (levelDB > -1.0) {
+            makeUpStateDB = log10(levelDB + 2.0f) - 1.0f - maxSampleDB;
```
daschuer, line 183: Please also take care of the warnings.

src/util/sample.cpp
```diff
@@ -480,6 +480,14 @@ void SampleUtil::copyClampBuffer(CSAMPLE* M_RESTRICT pDest,
     }
 }
 
+// static
+void SampleUtil::applyClamp(CSAMPLE* pBuffer, SINT iNumSamples) {
+    // note: LOOP VECTORIZED.
```
daschuer, line 485: did you verify that this loop is vectored? 
https://github.com/mixxxdj/mixxx/blob/59ca3a25b768c44ac0934ed4cb4504cac52bde54/src/util/sample.cpp#L16

daschuer (changes requested):

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,236 @@
...
+        applyAutoMakeUp(pState, pOutput, numSamples);
+    }
+
+    // Output gain
+    CSAMPLE gainParamDB = static_cast<CSAMPLE>(m_pGain->value());
+    SampleUtil::applyGain(pOutput, db2ratio(gainParamDB), numSamples);
+
+    // Clipping
+    if (m_pClipping->toInt() == static_cast<int>(Clipping::ClippingOn)) {
+        SampleUtil::applyClamp(pOutput, numSamples);
```
daschuer, line 166: What is the use case for this? If we consider a certain loudness like 16 LUFS this clipping is way to loud. It will artificially introduce extra noise. I can imagine that the hard clipping can be an effect that can be moved along with the threshold. 
A natural sounding tanh clipper is also useful.

src/effects/backends/builtin/compressoreffect.cpp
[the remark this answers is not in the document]
fonsargo: I think originally it was true, however your suggestion changes the meaning. If you don't like the original one, I can try to reformulate it. In fact it just adjusts volume to almost 0 dBFS level for any input.

src/util/sample.cpp
fonsargo: No, I have no clue what it is. It's just copy-paste of the original method. I'll remove this note.

src/effects/backends/builtin/compressoreffect.cpp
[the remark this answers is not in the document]
fonsargo: I was thinking about it and it's just come to my mind that maybe it's better to call it "AGC (Auto Gain Coontrol)"? What do you think?

src/effects/backends/builtin/compressoreffect.cpp
daschuer: I think it is piontless here as it is, because exactly that will happen when passed to the soundcard. 
So we may either remove it, or make it a useful effect. How about removing it now and postpone the work to a follow up PR?

fonsargo: Oh, I didn't know about that, I thought that without such clipping these peaks may cause problems for headphones, sound speakers and so on, especially when it broadcats to radio listeners (we never knows which speakers they use and etc.).

If so, I think it's a good idea to remove it now and do it as a separate effect, for example in #6513 .

src/effects/backends/builtin/compressoreffect.cpp
daschuer: These peaks still cause harm in your speakers. It is caused by the sub based introduced by the flatened sine waves.

src/effects/backends/builtin/compressoreffect.cpp
fonsargo: Indeed, you're right. Finally I did caching of these coeffs and compare it to user values.

src/effects/backends/builtin/compressoreffect.cpp
fonsargo: I think this message belongs to discussion of clipping? I've already removed clipping :)

Also I've changed float variables to doubles, so I believe it fixes warnings for Ubuntu.

daschuer: Oh yes, thank you.

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,238 @@
...
+    SampleUtil::applyGain(pOutput, gain, numSamples);
+}
+
+void CompressorEffect::applyAutoMakeUp(CompressorGroupState* pState,
+        CSAMPLE* pOutput,
+        const SINT& numSamples) {
+    double makeUpStateDB = pState->previousMakeUpGain;
+    CSAMPLE maxSample = SampleUtil::maxAbsAmplitude(pOutput, numSamples);
+    if (maxSample > 0) {
+        double maxSampleDB = ratio2db(maxSample);
```
daschuer, line 165: I have a warning here: 
```
/home/daniel/workspace/mixxx2/src/effects/backends/builtin/compressoreffect.cpp:165:48:   required from here
/home/daniel/workspace/mixxx2/src/util/math.h:78:21: warning: conversion from ‘double’ to ‘float’ may change value [-Wfloat-conversion]
   78 |     return log10(a) * 20;
      |            ~~~~~~~~~^~~~
```

src/effects/backends/builtin/compressoreffect.h
```diff
@@ -0,0 +1,77 @@
...
+#include "util/class.h"
+#include "util/defs.h"
+#include "util/sample.h"
+#include "util/types.h"
+
+namespace {
+constexpr CSAMPLE_GAIN kMakeUpAttackCoeff = 0.03f;
+constexpr CSAMPLE_GAIN kMakeUpTarget = -3.0f;
+
+double calculateBallistics(double paramMs, const mixxx::EngineParameters& engineParameters) {
```
daschuer, line 15: Unused?

daschuer:
I have tested this briefly and it works good. Thank you. 
Some findings: 
* The tootlips are long and in a single line. Can you add reasonable line breaks (@ronso0 or is there a better solution?) 
* The effect has not meta binding, not strictly necessary, but you may consider to add one with center at unity to be able to be used as a quick effect. 
* Many controllers have only access to the first three or four knobs. Verify that these are the knobs that users may touch during the performance. 
* Even though the Gain is well described as output gain in tool tips, I initially was confused I had the input gain in mind. On some stomp-boxes it is called level. Would that work for you as well?
* The default level is an issue when used as a deck effect, because it is way higher than the default -18 LUFS of replay gain and has zero head-room for mixing. To not destroy the gain level, the Output Gain should be ~-6 for modern pop music. 
* The Make-up switch is a tool to blow your ears/speakers, if you come form manual compensation. Did you consider to reset the gain knob to default while switching?  
* With the effect applied + Auto Make-Up, you are no longer able to mix, because the effect compensates your fade out. Do you have an idea how to fix that? Can we calculate the Make-Up gain form the other knobs alone?
* The effect introduces crackling. This is sourced form rectangular gain changes. It can be fixed by fade all gains for one buffer size.

src/effects/backends/builtin/compressoreffect.cpp
fonsargo: Actually I have no clue, why this warning is still here. I think I have convertion from float to double, but not double to float.

src/effects/backends/builtin/compressoreffect.h
fonsargo: No, I use it on these lines: 117,119,193,201 in compressoreffect.cpp

src/effects/backends/builtin/compressoreffect.cpp
daschuer: ratio2db is a template type. It uses the float implementation because of `CSAMPLE maxSample` 
But log10() is double. So it looks like a static cast to T is missing. 
`return static_cast<T>(log10(a)) * 20;` 
Not you fault, but please fix it. 

The other question is if we need double precision for maxSampleDB, nd can we avoid the DB round trip and instead use the ratio values?

src/effects/backends/builtin/compressoreffect.h
daschuer: Ah, the issue is that it is unused in very cpp file that includes the h file. Can we move the whole namespace to the CPP file?

src/effects/backends/builtin/compressoreffect.cpp
fonsargo: Yes, it's a good question. I managed to avoid this DB round trip, so now I don't use ratio2db() and db2ratio() for applying make up.

src/effects/backends/builtin/compressoreffect.h
fonsargo: Yes. sure. I've moved it to CPP file.

fonsargo:
Thank you for testing and for the good points.
I'll try to answer to it:

> * The tootlips are long and in a single line. Can you add reasonable line breaks

Yes, I've added line breaks.

> * The effect has not meta binding, not strictly necessary, but you may consider to add one with center at unity to be able to be used as a quick effect.
> * Many controllers have only access to the first three or four knobs. Verify that these are the knobs that users may touch during the performance.

Actually I'm not sure, that it could be useful. Compressor it's not a sound effect, it's more about sound tuning. Usually you find the best settings for you and use it all the time. So it will be strange if someone tries to increase and decrease attack or release knob for example dureing the perfomance, because it just mess up the sound.

> * Even though the Gain is well described as output gain in tool tips, I initially was confused I had the input gain in mind. On some stomp-boxes it is called level. Would that work for you as well?

Personally I think it should be "Gain", because it's the same in meaning as "Main Output Gain" knob in the main mixer for example. 

> * The default level is an issue when used as a deck effect, because it is way higher than the default -18 LUFS of replay gain and has zero head-room for mixing. To not destroy the gain level, the Output Gain should be ~-6 for modern pop music.

I guess you are talking about using "Auto Makeup Gain". So as it says in the description, it tries to adjust volume level to 0dBFS. So if someone wants to decrease it, they definetely can do it via gain knob.

> * The Make-up switch is a tool to blow your ears/speakers, if you come form manual compensation. Did you consider to reset the gain knob to default while switching?

I think it's a good idea, but I don't know how to introduce some "listener" to the Makeup switch button. Do you know how to do it?

> * With the effect applied + Auto Make-Up, you are no longer able to mix, because the effect compensates your fade out. Do you have an idea how to fix that? Can we calculate the Make-Up gain form the other knobs alone?

I believe it's because of the same bug as #12451. So if we apply compressor effect to the Deck 1 for example, then it should be applied _before_ fade out (volume control). But in reallity it is applied after volume control and so completely comprensate it.
However, I think it's better to use compressor as an output effect (in main mixer).

> * The effect introduces crackling. This is sourced form rectangular gain changes. It can be fixed by fade all gains for one buffer size.

I'm not sure, that I understand what you mean. However compressor is a powerful tool, it's always possible to mess everything  up with the wrong settings. So it's kind of "by design".

daschuer:
First of all, I am very happy with your work looking forward to integrate it into Mixxx. I am sorry for the delays in review. Sometimes a written review sounds hard and does not  transport the precognition your work deserves.

>> * The effect has not meta binding ...

> Actually I'm not sure, that it could be useful. Compressor it's not a sound effect, it's more about sound tuning. Usually you find the best settings for you and use it all the time. So it will be strange if someone tries to increase and decrease attack or release knob for example during the performance, because it just mess up the sound.

An issue is that we have pre-fader and postfader effects. Currently the compressor works only for as prefader, which is the Quick effect where only on knob is available. My use case is to use it to adjust 70th tracks or classic tracks with modern pop songs. 
Currently the workflow is a bit chunky. Place the compressor into the Effect rack, store find a good setting, safe it and put it into the quick effect. This cannot be done during a set. So we should find OK-ish defaults for the quick effect ans allow to adjust them. 

>> * Gain -> Level 

> Personally I think it should be "Gain", because it's the same in meaning as "Main Output Gain" knob in the main mixer for example. 

The "Main Output Gain" is the input Gain of you PA. So we should not judge from that. I have found more instances where it is called level without make-up the context. On the other hand it is usually called "make-up gain" which is quite long, when it controls the "make-up gain" stage which finally controls the "Level". You currently describe the knob as plain Output Gain which is the "Level". So I suggest to use the label "Level" then. But we may adjust this all altogether with the "makeup" button.      

>> * The default level is an issue when used as a deck effect, because it is way higher than the default -18 LUFS of replay gain and has zero head-room for mixing. To not destroy the gain level, the Output Gain should be ~-6 for modern pop music.

> I guess you are talking about using "Auto Makeup Gain". So as it says in the description, it tries to adjust volume level to 0dBFS. So if someone wants to decrease it, they definitely can do it via gain knob.

make-up is used to turn the output level of the compressed signal up to compensate the compressor effect. It should ideal match the input gain. The default Input gain is NOT 0 dBFs, as deck effect, because you don't have room for Mixing. In the main stage, you should also not compensate for 0 dBFs, because a digital signal of 0dBFs ma have high frequency parts between the samples which will than clip. 0 dBFs is not recommended for broadcast. This means we should not default to a problematic high value. Can we pick a default that matches the default level of Mixxx? 
  
>> * The Make-up switch is a tool to blow your ears/speakers, if you come form manual compensation. Did you consider to reset the gain knob to default while switching?

> I think it's a good idea, but I don't know how to introduce some "listener" to the Makeup switch button. Do you know how to do it?

Ah I see, this is an issue, we can do this easy. We may adjust the two gain knob scales that they match each other for the default Mixxx levels. This should lower the level jump we experience with the "Make-up" button. 

>> * With the effect applied + Auto Make-Up, you are no longer able to mix, because the effect compensates your fade out. Do you have an idea how to fix that? Can we calculate the Make-Up gain form the other knobs alone?

> I believe it's because of the same bug as #12451. So if we apply compressor effect to the Deck 1 for example, then it should be applied _before_ fade out (volume control). But in reallity it is applied after volume control and so completely comprensate it.
However, I think it's better to use compressor as an output effect (in main mixer).

We don't have on other solution to put the effect to the quick effect rack. Maybe the auto-make-up is a bit to fast? Can we slow it down or even make it adjustable? For my understanding it needs a time "learn" the input level. 

>> * The effect introduces crackling. This is sourced form rectangular gain changes. It can be fixed by fade all gains for one buffer size.

> I'm not sure, that I understand what you mean. However compressor is a powerful tool, it's always possible to mess everything  up with the wrong settings. So it's kind of "by design".

You hear it clearly if you put a reverb before the compressor. All gain changes must be ramped to not have the noise from the resulting rectangular wave.

fonsargo:
Thank you for the review!

> An issue is that we have pre-fader and postfader effects. Currently the compressor works only for as prefader, which is the Quick effect where only on knob is available. My use case is to use it to adjust 70th tracks or classic tracks with modern pop songs.
Currently the workflow is a bit chunky. Place the compressor into the Effect rack, store find a good setting, safe it and put it into the quick effect. This cannot be done during a set. So we should find OK-ish defaults for the quick effect ans allow to adjust them.

I'm not sure, that I completely understand you. But I agree that we need to find good default values. I thought that I found those, but I will try to do it better.

> So I suggest to use the label "Level" then. But we may adjust this all altogether with the "makeup" button.

Ok, I will rename it to "Level".

> make-up is used to turn the output level of the compressed signal up to compensate the compressor effect. It should ideal match the input gain. The default Input gain is NOT 0 dBFs, as deck effect, because you don't have room for Mixing. In the main stage, you should also not compensate for 0 dBFs, because a digital signal of 0dBFs ma have high frequency parts between the samples which will than clip. 0 dBFs is not recommended for broadcast. This means we should not default to a problematic high value. Can we pick a default that matches the default level of Mixxx?

Yes, I will decrease the default value. But I agree with you, that strictly speaking it's not a make-up, it's kind of AGC (Auto gain control).

>  We may adjust the two gain knob scales that they match each other for the default Mixxx levels. This should lower the level jump we experience with the "Make-up" button.

Yes, I will do it too.

> We don't have on other solution to put the effect to the quick effect rack. Maybe the auto-make-up is a bit to fast? Can we slow it down or even make it adjustable? For my understanding it needs a time "learn" the input level.

The problem is that if we even don't use the make-up, compressor will compensate fade out until the level becomes lower than compressor threshold. After that the level will drop dramatically, so it will looks like just stopping without fading out. Therefore I think we have to apply fader only after all effects, not before. So may be we need to create an issue for that? Because as I understand most of the other mixxing software do it in such order: firstly effects, secondly fader (or crossfade, microphone ducking and all other stuff).

> You hear it clearly if you put a reverb before the compressor. All gain changes must be ramped to not have the noise from the resulting rectangular wave.

Thank you, now I hear it. I think I need some time to investigate into it, maybe we need to use another algorithm for make-up or just delete this part.

fonsargo:
Eventually I decided to change the auto make up algorithm, so for now it tries to match the input loudness level (compensate decreasing gain effect from compressor) - as it usually works. Also I fixed crackling (I believe), and renamed "Gain" knob to "Level".

As I said before, the auto make up algorithm which I used before was a kind of AGC effect, so I decided to split it to a separate effect. But I can't find any existing issue about AGC. Does anyone know if it's ever been requested? Is it okay if I try to create separate PR with AGC effect later?

acolombier:
>  Does anyone know if it's ever been requested? 

AFAIK, the AGC compressor would be very helpful for the NI stem support feature as the DSP compressor setup embedded in the stem metadata. seem to be relying on such a compressor. SO it would be very useful to have this feature as well!

> Is it okay if I try to create separate PR with AGC effect later?

I think that makes sense yes

src/util/sample.cpp
acolombier: You might want to make sure it is  vectorized instead as it is [a significant optimisation](https://github.com/mixxxdj/mixxx/blob/ba1ac4545e821f3ab8bea602f002e416a26b8397/src/util/sample.h#L56). You can do so by using the `INFO_VECTORIZE` CMake option. In the build output, you should see an info/notice with the file and the line (`src/util/sample.cpp:483`) saying something along the lines of "vectorized with block of x bytes`

daschuer:
I just did a brief test, but I was not able to make it work again. Is something broken now?

daschuer (changes requested):

fonsargo:
> I just did a brief test, but I was not able to make it work again. Is something broken now?

No, it's not broken. It should work. As you can see, all checks have passed in GitHub. Maybe the problem is that it's already too old branch? Do I need to rebase it on new master? I see that the 2.4 version is already released, so maybe I need to rebase on 2.5?

src/effects/backends/builtin/compressoreffect.cpp
```diff
@@ -0,0 +1,248 @@
...
+    int channelCount = engineParameters.channelCount();
+    for (SINT i = 0; i < numSamples; i += channelCount) {
+        CSAMPLE maxSample = std::max(fabs(pInput[i]), fabs(pInput[i + 1]));
+        if (maxSample == CSAMPLE_ZERO) {
+            pOutput[i] = CSAMPLE_ZERO;
+            pOutput[i + 1] = CSAMPLE_ZERO;
+            continue;
+        }
+
+        double maxSampleDB = ratio2db(maxSample);
```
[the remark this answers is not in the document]
fonsargo: Do I have to do it? I understand your point, but it's not easy to transform such calculations. I'm sure that it's possible in theory, but the final code will be unreadable. All examples of compressor that I saw before were written in DB values.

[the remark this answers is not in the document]
daschuer: No, that was just a suggestion. You may add TODO comment instead, that someone else may pick it up.

daschuer:
OK, I will test again maybe there was something wrong in my setup.

acolombier:
> I see that the 2.4 version is already released, so maybe I need to rebase on 2.5?

Yes, it looks like this PR has been move to the `2.5` milestone, so I think you can rebase to `main`

fonsargo:
> Yes, it looks like this PR has been move to the `2.5` milestone, so I think you can rebase to `main`

Ok, I've rebased to `main`

daschuer (approved):
Thank you, works good now. 

In my first test I have tested with long attack and release times which was compensated by the auto markup gain. The kMakeUpAttackCoeff fits to the defaults. 
Did you consider to adjust it with the buffer size and the attack and the release times? 

It is good enough for beta I will file an issue for the final release. I would be happy if you could pick it up.

fwcd:
After playing around with this a bit: Pretty cool effect, I've found myself "compressing" decks manually which this effect can effectively automate. Though one note on practical use: I almost always end up turning off the "Auto Makeup" button and I don't really understand how it's supposed to work, even though it's on by default. The description says

```
The Auto Makeup button enables automatic gain adjustment to keep the input signal
and the processed output signal as close as possible in perceived loudness
```

but this seems to negate the whole point of compressing the deck, doesn't it? I don't want my level to depend on the input signal, that's why I'm compressing it, and most compressors I've seen in the wild apply auto-gain to some fixed value like -12 dB or 0 dB (which seems to have been the original idea in this PR? Haven't read the whole thread, so maybe there's a good reason for this).

Swiftb0y:
I haven't confirmed whether this applies to this implementation, but usually when compressing, you're reducing the peaks of the audio (as illustrated by the graph)

image

So while the dynamics get reduced, the overall sound gets quite a lot quieter too. The Auto-makeup should then apply uniform gain to the output to bring the compressed signal up to the previous loudness again. Some compressors simply compensate by applying a fixed gain on the output while the auto makeup calculates the RMS power of the input and boosts the output to match the RMS of the input. Does that make sense?

fwcd:
That makes sense for reducing the microdynamics of a signal (e.g. slightly extending the tail of a kick), but IMO makes the compressor curve a lot harder to "predict" mentally. Over which interval will the RMS be taken? Does the compression now depend on the buffer size, in which case a larger buffer would effectively result in a lower compression ratio since the auto makeup will automatically "decompress" the signal.

A specific example where reducing macrodynamics is helpful: I have a track with a relatively quiet intro and I'd like to compensate for this by bringing it to the same level as the "drop". The auto makeup would effectively negate the effect of the compressor here (unless the RMS is taken over a large enough window/the entire track, which would be hard to do given that effects operate on a live signal IIUC)

fonsargo:
I think compressor isn't only about different levels of loudness, it's also about reducing dynamic range. So with auto make up you still have quiet intro and loud chorus, but it's more compressed (like all instruments are clearly heard etc.) To be honest I also don't like this auto make up settings and usually I don't use it. I just use manual fixed gain, which I determine experimentally. Luckily you have this choice with this effect :)

Also to me it looks like you are looking for Auto Gain Control (AGC), which will keep loudness at the same level all the time. I have plans to implement it as well, so hopefully I will do it next year.

Regarding depending on buffer size, as far as I remember it shouldn't be the case. You can test it if you want, but it shouldn't depend on buffer size.

fwcd:
> Luckily you have this choice with this effect :)

Yeah, that's true, the only thing that bothers me a bit is that it resets to the default every time I select the effect again... I know I could probably save the chain as a preset, but I'm usually too lazy for that, especially when mixing and matching effects 😄

Maybe it would make sense to revisit the defaults though, if none of us uses them?

fonsargo:
As far as I remember these defaults were chosen because it doesn't change the sound rapidly. If auto make up is off by default, the music will be really quiet after turning on this effect. If we set some fixed gain by default, it could be not enough or too much, because it depends on the sound and other parameters. So the idea was that someone, who doesn't know anything about compressors, could use it "online" while mixing and doesn't mess up the sound.

PR #12530, opened 2024-01-07
AutoDJ: fix GUI freeze when updating duration for many selected tracks

### LINKED ISSUE #12520: Selecting many tracks in the AutoDJ queue may freeze the GUI
### Bug Description

Reported here https://mixxx.discourse.group/t/i-did-something-stupid/28750, and confirmed.

I suspect it's the function that updates the track count and play duration of the queue.
https://github.com/mixxxdj/mixxx/blob/f610d58a25b6f97172b097fb2a7a79147d863f7d/src/library/autodj/dlgautodj.cpp#L365-L375

### Version

_No response_

### OS

_No response_

ronso0:
fixes #12520 

* ~~move the duration collection to TrackDAO~~
* ~~query the database duration column instead of inspecting each Track object~~
____
* move the duration collection to PlaylistTableModel
(could also be in any of the classes that is derived from but... YAGNI?)
* query the duration column of each row instead of inspecting each Track object

Now updating the info for ~16k tracks takes ~5ms here.
With the 2.4 code, updating the info for 3 tracks already takes ~40ms :grimacing:

### DIFF
--- a/src/library/playlisttablemodel.cpp
+++ b/src/library/playlisttablemodel.cpp
@@ -319,6 +319,22 @@ void PlaylistTableModel::shuffleTracks(const QModelIndexList& shuffle, const QMo
     m_pTrackCollectionManager->internalCollection()->getPlaylistDAO().shuffleTracks(m_iPlaylistId, positions, allIds);
 }
 
+double PlaylistTableModel::getDurationOfRows(const QModelIndexList& indices) {
+    double durationTotal = 0.0;
+    if (indices.isEmpty()) {
+        return durationTotal;
+    }
+
+    const int durationColumnIndex = fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_DURATION);
+    for (const auto& index : indices) {
+        durationTotal += index.sibling(index.row(), durationColumnIndex)
+                                 .data(Qt::EditRole)
+                                 .toDouble();
+    }
+
+    return durationTotal;
+}
+
 bool PlaylistTableModel::isColumnInternal(int column) {
     return column == fieldIndex(ColumnCache::COLUMN_PLAYLISTTRACKSTABLE_TRACKID) ||
             TrackSetTableModel::isColumnInternal(column);
--- a/src/library/autodj/dlgautodj.cpp
+++ b/src/library/autodj/dlgautodj.cpp
@@ -367,12 +367,9 @@ void DlgAutoDJ::updateSelectionInfo() {
 
     QModelIndexList indices = m_pTrackTableView->selectionModel()->selectedRows();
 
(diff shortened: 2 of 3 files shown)


ronso0:
@uklotzde Do you have time to take a look?

ronso0:
I'll check if this should better be a helper function in PlaylistDAO, for example `getDuration(int playlistId, QList<int> positions)`.

ronso0:
> Hmm, no, this should better be a helper in PlaylistTableModel

Yes, and that's also much faster: ~5ms vs. 28ms for 16k tracks

src/library/playlisttablemodel.cpp
```diff
@@ -319,6 +319,22 @@ void PlaylistTableModel::shuffleTracks(const QModelIndexList& shuffle, const QMo
     m_pTrackCollectionManager->internalCollection()->getPlaylistDAO().shuffleTracks(m_iPlaylistId, positions, allIds);
 }
 
+double PlaylistTableModel::getDurationOfRows(const QModelIndexList& indices) {
```
ywwg, line 322: can you add a short explanatory comment, like:
// Derive total duration from the display widget instead of the database to speed up calculation.

ywwg (approved):

src/library/playlisttablemodel.cpp
ronso0: I'll add that in DlgAutoDJ where it is called, no need to mention that here IMO

src/library/playlisttablemodel.cpp
```diff
@@ -319,6 +319,22 @@ void PlaylistTableModel::shuffleTracks(const QModelIndexList& shuffle, const QMo
...
+    double durationTotal = 0.0;
+    if (indices.isEmpty()) {
+        return durationTotal;
+    }
+
+    const int durationColumnIndex = fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_DURATION);
+    for (const auto index : indices) {
```
daschuer, line 329:
```suggestion
    for (const auto& index : indices) {
```

daschuer (changes requested):
just on missing &

ronso0:
Done.

daschuer (approved):
LGTM, thank you

ronso0:
Oh, I just pushed the changes suggested by @uklotzde (it was a comment not a formal change request).
https://github.com/mixxxdj/mixxx/commit/4240e9056c48985ac91e747aeaead6ceadca125d?notification_referrer_id=NT_kwDOAFqMd7I5MDIwODc3MjE3OjU5MzQxOTk

The update is #12537

PR #12535, opened 2024-01-09
Deploy QML Module on Ubuntu

daschuer:
This deploys the QML Modules in version 6 instead of 5.

PR #12539, opened 2024-01-09
Ubuntu: Don't depend on not used QML/Quick libraries

daschuer:
Clean up PR ..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2607,7 +2607,7 @@ target_link_libraries(mixxx-lib PUBLIC
   Qt${QT_VERSION_MAJOR}::Xml)
 if(QT_EXTRA_COMPONENTS)
   foreach(COMPONENT ${QT_EXTRA_COMPONENTS})
-    target_link_libraries(mixxx-lib PUBLIC Qt6::${COMPONENT})
+    target_link_libraries(mixxx-lib PUBLIC Qt${QT_VERSION_MAJOR}::${COMPONENT})
   endforeach()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2568,12 +2568,11 @@ if(QT6)
   find_package(QT 6.2 NAMES Qt6 COMPONENTS Core REQUIRED)
   list(APPEND QT_EXTRA_COMPONENTS "SvgWidgets")
   list(APPEND QT_EXTRA_COMPONENTS "Core5Compat")
-else()
-  find_package(QT 5.12 NAMES Qt5 COMPONENTS Core REQUIRED)
-endif()
-if(QML)
   list(APPEND QT_EXTRA_COMPONENTS "Quick")
   list(APPEND QT_EXTRA_COMPONENTS "QuickControls2")
+  list(APPEND QT_EXTRA_COMPONENTS "QuickWidgets")
+else()
+  find_package(QT 5.12 NAMES Qt5 COMPONENTS Core REQUIRED)
 endif()
 find_package(Qt${QT_VERSION_MAJOR}
   COMPONENTS
@@ -2583,8 +2582,7 @@ find_package(Qt${QT_VERSION_MAJOR}
     Network
     OpenGL
     PrintSupport
-    Qml
-    QuickWidgets
+    Qml # for QJSEngine
     Sql
     Svg
     Test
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -2612,7 +2609,9 @@ if(QT_EXTRA_COMPONENTS)
   foreach(COMPONENT ${QT_EXTRA_COMPONENTS})
     target_link_libraries(mixxx-lib PUBLIC Qt6::${COMPONENT})
```
uklotzde, line 2610:
```suggestion
    target_link_libraries(mixxx-lib PUBLIC Qt${QT_VERSION_MAJOR}::${COMPONENT})
```

CMakeLists.txt
```diff
@@ -2602,17 +2600,18 @@ target_link_libraries(mixxx-lib PUBLIC
...
   Qt${QT_VERSION_MAJOR}::Widgets
   Qt${QT_VERSION_MAJOR}::Xml)
 if(QT_EXTRA_COMPONENTS)
   foreach(COMPONENT ${QT_EXTRA_COMPONENTS})
-    target_link_libraries(mixxx-lib PUBLIC Qt6::${COMPONENT})
+    target_link_libraries(mixxx-lib PUBLIC Qt${QT_VERSION_MAJOR}::${COMPONENT})
   endforeach()
+endif()
 
+if(QT6)
```
uklotzde, line 2614:
```suggestion
if(QML)
```

CMakeLists.txt
```diff
@@ -2670,7 +2669,7 @@ endif()
 option(DEBUG_ASSERTIONS_FATAL "Fail if debug become true assertions" OFF)
 if(DEBUG_ASSERTIONS_FATAL)
   target_compile_definitions(mixxx-lib PUBLIC MIXXX_DEBUG_ASSERTIONS_FATAL MIXXX_DEBUG_ASSERTIONS_ENABLED)
-  if(QML)
+  if(QT6)
```
uklotzde, line 2672:
```suggestion
  if(QML)
```

Why did you change this? QT6 and QML are different options.

uklotzde (changes requested):

CMakeLists.txt
daschuer: This is the 2.4 release where QT6 implies QML. The Qt6 widget GUI is in the 2.5 branch.

CMakeLists.txt
daschuer: Only in 2.5 / main

uklotzde:
I didn't follow the evolution of the `QT6`/`QML` options. What matters most is that the unneeded dependencies are finally fixed.

daschuer:
@ronso0 Can you verify and merge this, one of the last 2.4 PRs ...

ronso0:
IIUC, to verify I'd need to uninstall the QML/Quick packages?
I'll do this tonight/tomorrow.

daschuer:
This is probably not worth the effort. My Idea was to just do a code review and verify in the package dependencies id there no unnecessary extra. From the build Server we know that it builds and runs. This version has no QML, so there is nothing QML loaded at runtime, not included in the CI tests. 

With 2.5 and QML it is more difficult because we need to make sure all runtime loaded QML plug-Ins are installed. The best is to have a test that tries to start the QML GUI on the CI that we know nothing falls apparat. No Idea how to do this effective. But that is a future topic.

ywwg:
@daschuer I have a second machine I can test this on, is all I have to do a build/run?

daschuer:
The test will not reveal if we have still unnecessary dependencies. The other way around would work, but that is not the case here.  If you have a fresh installed Ubuntu you may test 2.5 and https://github.com/mixxxdj/mixxx/pull/12535 install the created deb from GitHub and check if `mixxx --qml` works.

tools/debian_buildenv_qt6.sh
```diff
@@ -100,10 +100,6 @@ case "$1" in
             protobuf-compiler \
             qtkeychain-qt6-dev \
             qt6-declarative-dev \
-            qml-module-qtquick-controls \
-            qml-module-qtquick-controls2 \
-            qml-module-qt-labs-qmlmodels \
-            qml-module-qtquick-shapes \
```
ronso0, line 106: I thought this is just about removing them from Qt5.
Why are these removed from the Qt6 buildenv.
Won't cmake miss them when building with QT6=ON?

Sorry, I didn't follow the Qt6 topic lately, did I miss the fact that QML was removed _entirely_ from 2.4?

daschuer: These are the Qt5 variants. We may need the Qt6 variants, like in https://github.com/mixxxdj/mixxx/pull/12535 which is targeted to 2.5. Should we use the 2.5 file here as well? I original considered as unnecessary because we have also artificially failed the Qt6 build in this branch when not on GitHub. What do you think?

ronso0: I don't think so.
If the CI Qt6 build passes we should not complicate things in order to allow a QML build of Mixxx 2.4, that's what 2.5 is for.

ronso0 (approved):
LGTM, thanks.

ronso0:
If anyone else more familiar with building/packaging wants to take a look, please do so @mixxxdj/developers 

If there are no objections, let's just merge this in 2 days.

PR #12541, opened 2024-01-09
fix(qml): Improve knobs by applying selective 4xMSAA on the Arc shape

### LINKED ISSUE #12536: QML UI shows aliasing on certaint components
### Bug Description

**Expected**

The QML UI should not have any aliasing.

**Actual**

The QML UI shows aliasing, especially on component using `Shape`

Screenshot from 2024-01-08 20-00-00

(zoom on the component)

image
image

**Workaround**

Following this [Stack Overflow thread](https://stackoverflow.com/a/49576756), this patch seems to resolve the problem:
```
From 057ca11131c89943feae3584d7e627d3d23b438b Mon Sep 17 00:00:00 2001
From: Antoine C <mixxx@acolombier.dev>
Date: Mon, 8 Jan 2024 20:01:40 +0000
Subject: [PATCH] QML: Use multi-sampling to prevent aliasing

---
 src/main.cpp | 6 ++++++
 1 file changed, 6 insertions(+)

diff --git a/src/main.cpp b/src/main.cpp
index c910926e51..5a8e154c4b 100644
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -5,6 +5,9 @@
 #include <QStringList>
 #include <QTextCodec>
 #include <QThread>
+#ifdef MIXXX_USE_QML
+#include <QSurfaceFormat>
+#endif
 #include <QtDebug>
 #include <cstdio>
 #include <stdexcept>
@@ -56,6 +59,9 @@ int runMixxx(MixxxApplication* pApp, const CmdlineArgs& args) {
     int exitCode;
 #ifdef MIXXX_USE_QML
     if (args.isQml()) {
+        QSurfaceFormat format;
+        format.setSamples(16);
+        QSurfaceFormat::setDefaultFormat(format);
         mixxx::qml::QmlApplication qmlApplication(pApp, pCoreServices);
         exitCode = pApp->exec();
     } else
-- 
2.34.1
```

Screenshot from 2024-01-08 20-00-39

(zoom on the component)

image
image

### Version

main branch

### OS

Pop OS 22.04 (with Wayland)

Holzhaus:
This is way less resource-consuming than doing MSAA on the entire GUI.

Fixes #12536.

### DIFF
--- a/res/qml/Mixxx/Controls/Knob.qml
+++ b/res/qml/Mixxx/Controls/Knob.qml
@@ -63,6 +63,12 @@ Item {
         anchors.fill: parent
         antialiasing: true
         visible: root.arc
+        // Enable smooth curves. For QtQuick Shapes, this currently only works
+        // by enabling multisampling, so we use 4xMSAA here.
+        //
+        // See https://www.qt.io/blog/2017/07/07/let-there-be-shapes for details.
+        layer.enabled: true
+        layer.samples: 4
 
         ShapePath {
             id: arcPath


Holzhaus:
@acolombier please have a look if that works for you, too

Swiftb0y:
lgtm, but enabling layers for an item that might be onscreen often may also have performance issues. https://doc.qt.io/qt-6/qml-qtquick-item.html#memory-and-performance

```quote

Also, an item using a layer can not be [batched](https://doc.qt.io/qt-6/qtquick-visualcanvas-scenegraph-renderer.html#batching) during rendering. This means that a scene with many layered items may have performance problems.

Layering can be convenient and useful for visual effects, but should in most cases be enabled for the duration of the effect and disabled afterwards.
```

Holzhaus:
Sure, but practically I didn't notice any performance issues on my laptop (in contrast to using 16xMSAA globally which made the QML completely unresponsive and unusable for me).

acolombier:
@Holzhaus it works but the issue is that other component also shows aliasing (see the waveform pre-roll triangle in the issue). Would we want to apply local multi sampling to each components subject to aliasing? Also, do we want to hardcode the multi-sampling level? Alternatively, from @Swiftb0y's comment, I've started some work to allow setting a custom multi-sampling level in the setting

Swiftb0y:
I'd probably say both. I'm guessing that as long as the offscreen framebuffers are smaller than the screen, doing AA per component is probably cheaper. 

> Would we want to apply local multi sampling to each components subject to aliasing?

It depends on the skin, I'd say that on overage 50% of screen content probably needs AA (for 2 decks, maybe 80% for 4 decks), but when the library is maximized that drops down to 10% as the library is pretty square anyways and text is already AA'd anyways. Doing AA for the entire application in that case is probably quite wasteful. So only doing it per-component would avoid the overhead when those components aren't shown (theoretically, not sure if Qt Quick is implemented like that).

@Holzhaus have you measured doing just 4xMSAA on the entire GUI? afaik that should only have 1/4th of the AA overhead while still providing nice results.

Anyways, I think its too soon to decide purely based on performance. The QML gui is already not super well optimized (lots of draw calls, suboptimal batching due to the abundance of transparent vector graphics) so I think we can postpone optimization work like that for later.

Holzhaus:
No, I didn't measure anything. I just tested this commit https://github.com/Holzhaus/mixxx/pull/24/commits/057ca11131c89943feae3584d7e627d3d23b438b and the qml screen remained blank for roughly 10 secs before it started to render, and even then it didn't react to changes and was basically too laggy to be usable.

And that was with 8x MSAA (I saw a warning message on the terminal that 16 is not supported and 8 is used).

With this patch the knobs look nicer and didn't notice a performance downgrade.

Swiftb0y:
I'm wondering whether the fallback to 8xMSAA is an indicator for a driver issue... I'd be surprised if 2020 hardware can't handle 16x.

Swiftb0y (approved):
Thank you. This should probably be lightweight enough for most systems. @acolombier volunteered to build upon this in https://github.com/mixxxdj/mixxx/pull/12546 to make the quality configurable.

ronso0:
:tada:

PR #12542, opened 2024-01-10
Tag fetcher: allow to apply cover and tags separately // var.2

### LINKED ISSUE #12532: Fetched cover art from "MusicBrainz" is auto applied
### Bug Description

Mixxx 2.4.0-beta
Git Version: 2.4-beta-314-g71dceb27ba
Platform: Linux x86_64
OS: Debian 11

**Preferences -> Library -> "Synchronize library track metadata from/to file tags" is checked (enabled).**

When fetching tags from "MusicBrainz", cover art is also fetched after selecting one of the suggestions.
Clicking on "Apply" auto applies the cover art along with the tags without requiring a click on "Apply Cover".

After manually clearing the auto applied cover art (which I do not want), the downloaded image is still in the filesystem and not removed.

Shouldn't "Apply" be renamed to "Apply Tags" and its functionality changed to only apply tags?
How can I only apply the tags without applying the suggested cover art?

### Version

_No response_

### OS

_No response_

ronso0:
Fixes #12532 

Now Apply only applies the tags.
(now-wrong status message is disabled)

The last commit renames the Apply button and adjusts the status message.
'Apply' does not necessarily need to be renamed (translations will be abandoned), so if we don  **not** want to change tr strings that late I'll revert that commit.

image

### DIFF
--- a/src/library/dlgtagfetcher.cpp
+++ b/src/library/dlgtagfetcher.cpp
@@ -112,6 +112,18 @@ void addTagRow(
     }
 }
 
+void addDivider(const QString& text, QTreeWidget* pParent) {
+    QTreeWidgetItem* pItem = new QTreeWidgetItem(pParent);
+    pItem->setFirstColumnSpanned(true);
+    pItem->setText(0, text);
+    pItem->setFlags(Qt::NoItemFlags);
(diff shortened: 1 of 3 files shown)


daschuer:
We have decided for this alternative: https://github.com/mixxxdj/mixxx/pull/12543

Issue #12525, opened 2024-01-06
2.5-alpha won't start on Uuntu 22.04 due to missing Qt platform plugins


ronso0:
### Bug Description

reported here https://mixxx.discourse.group/t/ubuntu-22-04-update-mixxx-not-working/28743

Similiar reports exist for other apps and it was fixed by installing some xcb packages.
Do we need to update dependencies?

### Version

2.5-alpha

### OS

Ubuntu 22.04

naught101:
Related? https://github.com/mixxxdj/mixxx/pull/12284

naught101:
A work around for me was removing the mixxx-nightlies repos, adding https://launchpad.net/~mixxx/+archive/ubuntu/mixxxbetas, and downgrading to 2.4.0-beta-whatever.

ronso0:
So, since
* similiar issues where reported for other Qt apps much earlier
* 2.5-alpha worked until recently for @stefan-franz @naught101 and other alpha testers

I assume it's not an Qt update / Qt packaging issue but indeed related to #12284 

Everyone affected: please test the CI build from #12545 
Instruction for testing and backing up your data are here https://github.com/mixxxdj/mixxx/wiki/Testing

daschuer:
This is the warning from the original report: 
```
azz@Dazz-VPCEH3N6E:~$ mixxx
Using preferences ScaleFactor 1
Warning: Ignoring WAYLAND_DISPLAY on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
qt.qpa.plugin: Could not find the Qt platform plugin “xcb” in “”
qt.qpa.plugin: Could not find the Qt platform plugin “wayland” in “”
```

daschuer:
Does starting via 
```
QT_DEBUG_PLUGINS=1 mixxx
``` 
Reveals an issue?

daschuer:
Normally 'QT_DEBUG_PLUGINS=1 mixxx' should produce extra log info like this: 

```
...
Got keys from plugin meta data ("ico", "cur")
QFactoryLoader::QFactoryLoader() looking at ...
Found metadata in lib ...
metadata=
{
    "IID": "org.qt-project.Qt.QImageIOHandlerFactoryInterface",
    "MetaData": {
        "Keys": [
            "jpg",
            "jpeg"
        ],
        "MimeTypes": [
            "image/jpeg",
            "image/jpeg"
        ]
    },
    "className": "QJpegPlugin",
[...]

PR #12545, opened 2024-01-10
Revert "MixxxApplication: Use XCB plugin for static Qt builds on Linux"

ronso0:
This reverts commit 6b943eb7732ee6362e08383c1f631dd3f95e120a.

Just a test if that commit causes #12525 
(I didn't look into the change, no idea if that can affect the ppa build at all)


daschuer (changes requested):
This code is guarded by `#ifdef QT_STATIC` which is not set in the PPA build case.

PR #12546, opened 2024-01-10
fix: add multi-sampling settings for QML

acolombier:
This extends the work made on #12541 to address #12536 with a customizable multi-sampling level.

### DIFF
--- a/res/qml/Mixxx/Controls/Knob.qml
+++ b/res/qml/Mixxx/Controls/Knob.qml
@@ -67,8 +67,9 @@ Item {
         // by enabling multisampling, so we use 4xMSAA here.
         //
         // See https://www.qt.io/blog/2017/07/07/let-there-be-shapes for details.
-        layer.enabled: true
-        layer.samples: 4
+        property int multiSamplingLevel: Mixxx.Config.getMultiSamplingLevel()
+        layer.enabled: multiSamplingLevel > 1
+        layer.samples: multiSamplingLevel
 
         ShapePath {
             id: arcPath
--- a/src/preferences/dialog/dlgprefinterface.cpp
+++ b/src/preferences/dialog/dlgprefinterface.cpp
@@ -194,7 +194,7 @@ DlgPrefInterface::DlgPrefInterface(
         mulitSamplingComboBox->addItem(tr("8x MSAA"), 8);
         mulitSamplingComboBox->addItem(tr("16x MSAA"), 16);
 
-        m_multiSampling = m_pConfig->getValue(ConfigKey(kPreferencesGroup, kMultiSamplingKey), 0);
+        m_multiSampling = m_pConfig->getValue(ConfigKey(kPreferencesGroup, kMultiSamplingKey), 4);
         int mulitSamplingIndex = mulitSamplingComboBox->findData(m_multiSampling);
         if (mulitSamplingIndex != -1) {
             mulitSamplingComboBox->setCurrentIndex(mulitSamplingIndex);
--- a/src/preferences/dialog/dlgprefinterface.cpp
+++ b/src/preferences/dialog/dlgprefinterface.cpp
@@ -26,12 +26,14 @@ namespace {
 
 const QString kConfigGroup = QStringLiteral("[Config]");
 const QString kControlsGroup = QStringLiteral("[Controls]");
+const QString kPreferencesGroup = QStringLiteral("[Preferences]");
 const QString kScaleFactorKey = QStringLiteral("ScaleFactor");
 const QString kStartInFullscreenKey = QStringLiteral("StartInFullscreen");
 const QString kSchemeKey = QStringLiteral("Scheme");
 const QString kResizableSkinKey = QStringLiteral("ResizableSkin");
 const QString kLocaleKey = QStringLiteral("Locale");
 const QString kTooltipsKey = QStringLiteral("Tooltips");
+const QString kMultiSamplingKey = QStringLiteral("multi_sampling");
 
 } // namespace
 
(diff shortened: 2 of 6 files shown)


acolombier:
@Swiftb0y are you happy with the current solution?

src/qml/qmlconfigproxy.h
```diff
@@ -18,8 +18,11 @@ class QmlConfigProxy : public QObject {
             UserSettingsPointer pConfig,
             QObject* parent = nullptr);
 
+    // We use method here instead of properties as there is no way to achieve property binding
+    // with UserSettings, since there is no synchronisation upon mutations.
```
acolombier, line 22: Added that comment as I was tempted to refactor this to `Q_PROPERTY` at first glance, but then I remember the static aspect of user settings at the moment, so thought I will put the thought here to save someone else the same temptations :)

src/qml/qmlconfigproxy.cpp
```diff
@@ -31,6 +35,11 @@ QVariantList QmlConfigProxy::getTrackColorPalette() {
     return paletteToQColorList(colorPaletteSettings.getTrackColorPalette());
 }
 
+int QmlConfigProxy::getMultiSamplingLevel() {
+    ColorPaletteSettings colorPaletteSettings(m_pConfig);
```
Swiftb0y, line 39: copy-paste error?
```suggestion
```

Swiftb0y (commented):
lgtm otherwise.

src/qml/qmlconfigproxy.cpp
acolombier: Spot on - fixed now

res/qml/Mixxx/Controls/Knob.qml
```diff
@@ -67,8 +67,9 @@ Item {
         // by enabling multisampling, so we use 4xMSAA here.
         //
         // See https://www.qt.io/blog/2017/07/07/let-there-be-shapes for details.
-        layer.enabled: true
-        layer.samples: 4
+        property int multiSamplingLevel: Mixxx.Config.getMultiSamplingLevel()
+        layer.enabled: multiSamplingLevel != 0
```
Swiftb0y, line 71: technically doesn't make a difference, but use strict comparision for good practice.
```suggestion
        layer.enabled: multiSamplingLevel !== 0
```

src/preferences/dialog/dlgprefinterface.cpp
```diff
@@ -25,12 +25,14 @@ namespace {
...
 const QString kConfigGroup = QStringLiteral("[Config]");
 const QString kControlsGroup = QStringLiteral("[Controls]");
+const QString kPreferencesGroup = QStringLiteral("[Preferences]");
 const QString kScaleFactorKey = QStringLiteral("ScaleFactor");
 const QString kStartInFullscreenKey = QStringLiteral("StartInFullscreen");
 const QString kSchemeKey = QStringLiteral("Scheme");
 const QString kResizableSkinKey = QStringLiteral("ResizableSkin");
 const QString kLocaleKey = QStringLiteral("Locale");
 const QString kTooltipsKey = QStringLiteral("Tooltips");
+const QString kMultiSamplingKey = QStringLiteral("multi_sampling");
```
Swiftb0y, line 35: consistency...
```suggestion
const QString kMultiSamplingKey = QStringLiteral("MultiSampling");
```

Swiftb0y (commented):
more thorough review

src/preferences/dialog/dlgprefinterface.cpp
acolombier: That's what I started with, and then I realized than other keys in `[Preference]` were using a lower snake case formatting. Is the capitalized pascal case now enforce for every new setting keys?

res/qml/Mixxx/Controls/Knob.qml
Holzhaus: Is 1x multisampling even a thing? Should this be `multisampling > 1`?

src/preferences/dialog/dlgprefinterface.cpp
Holzhaus: I think the name could be improved, this could also be a boolean value. How about something like `MultiSamplingFactor`?

acolombier: Sounds good to me. 
Do you concur that capitalized pascal case is now the new naming standard for new settings keys?

Swiftb0y: Sorry, I forgot to respond earlier. I only looked at the context of the file in regards to the naming scheme and assumed the rest was consistent. Lets not make this a discussion and keep it as nobody will look at this anyways.

Swiftb0y (approved):
LGTM. Thank you

src/preferences/dialog/dlgprefinterface.cpp
```diff
@@ -181,6 +183,31 @@ DlgPrefInterface::DlgPrefInterface(
...
+#ifdef MIXXX_USE_QML
+    if (CmdlineArgs::Instance().isQml()) {
+        mulitSamplingComboBox->clear();
+        mulitSamplingComboBox->addItem(tr("Disabled"), 0);
+        mulitSamplingComboBox->addItem(tr("2x MSAA"), 2);
+        mulitSamplingComboBox->addItem(tr("4x MSAA"), 4);
+        mulitSamplingComboBox->addItem(tr("8x MSAA"), 8);
+        mulitSamplingComboBox->addItem(tr("16x MSAA"), 16);
+
+        m_multiSampling = m_pConfig->getValue(ConfigKey(kPreferencesGroup, kMultiSamplingKey), 0);
```
Swiftb0y, line 196: actually thinking about it, shouldn't the default value be 4 instead? otherwise we're back to the AA-less default, resulting in poor out-of-the-box experience. wdyt?

Swiftb0y:
friendly ping @acolombier ;)

src/preferences/dialog/dlgprefinterface.cpp
acolombier: Yep, that would makes sense as default, but I'm wondering what would be the impact on lightweight setup. I've updated it anyway.

acolombier:
Sorry, I missed the notification on that PR @Swiftb0y !

src/preferences/dialog/dlgprefinterface.cpp
Swiftb0y: I think those are niche enough. On lightweight setups we can reduce it down (maybe autodetect the framerate and adjust defaults downward accordingly?) I'd prefer the OOTB experience to be more polished for the majority of users.

src/preferences/dialog/dlgprefinterface.cpp
```diff
@@ -182,6 +184,31 @@ DlgPrefInterface::DlgPrefInterface(
     int inhibitsettings = static_cast<int>(m_pScreensaverManager->status());
     comboBoxScreensaver->setCurrentIndex(comboBoxScreensaver->findData(inhibitsettings));
 
+    // Multi-Sampling
+#ifdef MIXXX_USE_QML
+    if (CmdlineArgs::Instance().isQml()) {
+        mulitSamplingComboBox->clear();
```
Holzhaus, line 190: Typo

Swiftb0y (approved):
LGTM. Thank you.

src/preferences/dialog/dlgprefinterface.cpp
```diff
@@ -26,12 +26,14 @@ namespace {
...
 const QString kConfigGroup = QStringLiteral("[Config]");
 const QString kControlsGroup = QStringLiteral("[Controls]");
+const QString kPreferencesGroup = QStringLiteral("[Preferences]");
 const QString kScaleFactorKey = QStringLiteral("ScaleFactor");
 const QString kStartInFullscreenKey = QStringLiteral("StartInFullscreen");
 const QString kSchemeKey = QStringLiteral("Scheme");
 const QString kResizableSkinKey = QStringLiteral("ResizableSkin");
 const QString kLocaleKey = QStringLiteral("Locale");
 const QString kTooltipsKey = QStringLiteral("Tooltips");
+const QString kMultiSamplingKey = QStringLiteral("multi_sampling");
```
Holzhaus, line 36: Is there a reason why we don't call it "multi_sampling" instead of "msaa" everywhere? The latter is much less ambiguous IMHO.

src/preferences/dialog/dlgprefinterface.cpp
Swiftb0y: damn... why did CI not catch this?

Swiftb0y:
sorry for the review race-condition...

src/preferences/dialog/dlgprefinterface.cpp
Holzhaus: Codespell uses a dictionary of common misspellings and does not support camelcase at the moment. The chance that "mulitsamplingcombobox" is in the dict is rather small :)

Holzhaus:
> sorry for the review race-condition...

my bad, my review came too late xD

PR #12547, opened 2024-01-10
allshader RGB: Show a ghosted version of the original waveform behind…

ywwg:
… the gain-adjusted version.

This was a feature of the legacy RGB waveform and allows the DJ to see frequencies that will be present if the EQ is changed.

this does *not* have to go in 2.4 -- strictly nice-to-have.

### DIFF
--- a/src/waveform/renderers/allshader/waveformrendererrgb.cpp
+++ b/src/waveform/renderers/allshader/waveformrendererrgb.cpp
@@ -26,6 +26,7 @@ void WaveformRendererRGB::onSetup(const QDomNode& node) {
 void WaveformRendererRGB::initializeGL() {
     WaveformRendererSignalBase::initializeGL();
     m_shader.init();
+    m_shader_ghost.init();
 }
 
 void WaveformRendererRGB::paintGL() {
@@ -94,6 +95,10 @@ void WaveformRendererRGB::paintGL() {
     m_vertices.reserve(reserved);
     m_colors.clear();
     m_colors.reserve(reserved);
+    m_vertices_ghost.clear();
+    m_vertices_ghost.reserve(reserved);
+    m_colors_ghost.clear();
+    m_colors_ghost.reserve(reserved);
 
     m_vertices.addRectangle(0.f,
             halfBreadth - 0.5f * devicePixelRatio,
@@ -140,6 +145,12 @@ void WaveformRendererRGB::paintGL() {
         float maxMid = static_cast<float>(u8maxMid);
         float maxHigh = static_cast<float>(u8maxHigh);
         float maxAllChn[2]{static_cast<float>(u8maxAllChn[0]), static_cast<float>(u8maxAllChn[1])};
+
+        float ghostLow = maxLow;
+        float ghostMid = maxMid;
+        float ghostHigh = maxHigh;
+        float maxAllChn_ghost[2]{static_cast<float>(u8maxAllChn[0]),
+                static_cast<float>(u8maxAllChn[1])};
         // Uncomment to undo scaling with pow(value, 2.0f * 0.316f) done in analyzerwaveform.h
         // float maxAllChn[2]{unscale(u8maxAllChn[0]), unscale(u8maxAllChn[1])};
 
@@ -147,7 +158,7 @@ void WaveformRendererRGB::paintGL() {
         // We take the square root to get the magnitude below.
         const float sum = math_pow2(maxLow) + math_pow2(maxMid) + math_pow2(maxHigh);
 
-        // Apply the gains
+        // Apply the gains (non-ghosted version only)
         maxLow *= lowGain;
         maxMid *= midGain;
         maxHigh *= highGain;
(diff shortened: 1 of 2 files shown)


Swiftb0y:
from a technical note, this might be a good use for geometry shaders to generate the waveform mesh copy on the GPU instead... IMO for this to get merged it also needs to be configurable, as I can imagine that there is a sizable portion of people (me included) that doesn't like the look.

ywwg:
I don't love adding settings for everything, I'd prefer to find visual solutions that are generally acceptably pleasant and useful.  In this case, I'm restoring a feature in the legacy display that we've had for years and is just missing from the new one.  What about the ghosts don't you like?

Swiftb0y:
> I don't love adding settings for everything, I'd prefer to find visual solutions that are generally acceptably pleasant and useful.

I fully agree. 

> I'm restoring a feature in the legacy display that we've had for years and is just missing from the new one. 

I didn't know about that. Good to know, though I doubt many actually people miss it (I haven't seen an issue about it yet). 

> What about the ghosts don't you like? 

I haven't tried this PR, but imagining how it will look, results in too much clutter. Moreover some people might simply not need the feature and would like to avoid the additional mental load induced by that extra visual noise. 

I also dislike that its a feature limited to only a single shader again, while with a little bit of clever rearranging it could be implemented generically for every shader.

m0dB:
Independent of this specific feature (I still need to check it out), and in line with @Swiftb0y's comment, in my opinion the current approach with selecting a specific waveform renderer from a list is not ideal and I think a more option-based approach would make more sense. E.g.

* Color mapping: Mono, RGB, HSV (and I would like to add my centroid+flatness hue/saturation, https://github.com/m0dB/mixxx/pull/7)
* Color per L/R channel: on/off
* Scaling: linear / boost low level

and sure:
* Ghosted unfiltered waveform:  on/off

This of course would mean removing all the legacy waveforms, but that was the plan anyway. Once that is done, this should be pretty straightforward and would remove a lot of code duplication.

> I'd prefer to find visual solutions that are generally acceptably pleasant and useful.

I kind of agree, and I do see that the flexibility and configurability of mixxx comes with it downsides (it slows us down, it increases the maintenance burden, it can cause confusion, it dilutes the developer effort), but I also think that it is what makes mixxx stand out in comparison with closed source applications, and it is an inherent aspect of an open source project. So IMO if we can come up with a default setting that is "generally acceptably pleasant and useful" while maintaining a degree of costumization, that would be ideal.

src/waveform/renderers/allshader/waveformrendererrgb.cpp
```diff
@@ -140,21 +145,29 @@ void WaveformRendererRGB::paintGL() {
...
-        // Apply the gains
+        // Apply the gains (non-ghosted version only)
         maxLow *= lowGain;
         maxMid *= midGain;
         maxHigh *= highGain;
 
         // Calculate the squared magnitude of the gained maxLow, maxMid and maxHigh values
         // We take the square root to get the magnitude below.
         const float sumGained = math_pow2(maxLow) + math_pow2(maxMid) + math_pow2(maxHigh);
+        const float sumGained_ghost = math_pow2(ghostLow) +
```
m0dB, line 169: You are overcomplicating things. You don't need to sum the low+mid+high components because ... (see below)

src/waveform/renderers/allshader/waveformrendererrgb.cpp
```diff
@@ -164,6 +177,10 @@ void WaveformRendererRGB::paintGL() {
             const float factor = std::sqrt(sumGained / sum);
             maxAllChn[0] *= factor;
             maxAllChn[1] *= factor;
+
+            const float factor_ghost = std::sqrt(sumGained_ghost / sum);
```
m0dB, line 181: ... sumGained_ghost  == sum, so factor_ghost is always 1.

src/waveform/renderers/allshader/waveformrendererrgb.cpp
```diff
@@ -185,13 +202,35 @@ void WaveformRendererRGB::paintGL() {
...
                 fpos + 0.5f,
                 halfBreadth + heightFactor * maxAllChn[1]);
         m_colors.addForRectangle(red, green, blue);
 
+        m_vertices_ghost.addRectangle(fpos - 0.5f,
+                halfBreadth - heightFactor * maxAllChn_ghost[0],
+                fpos + 0.5f,
+                halfBreadth + heightFactor * maxAllChn_ghost[1]);
+        m_colors_ghost.addForRectangle(red_ghost, green_ghost, blue_ghost, ghost_alpha);
+
```
m0dB, line 233: Since the alpha doesn't change it would be better to use a uniform for that. It would be a simple to add the uniform for the alpha to the RGBShader, and that way you could use the same shader for normal and ghost.

src/waveform/renderers/allshader/waveformrendererrgb.h
```diff
@@ -22,8 +24,13 @@ class allshader::WaveformRendererRGB final : public allshader::WaveformRendererS
 
   private:
     mixxx::RGBShader m_shader;
+    mixxx::RGBAShader m_shader_ghost;
     VertexData m_vertices;
     RGBData m_colors;
+    VertexData m_vertices_ghost;
+    RGBAData m_colors_ghost;
+
```
m0dB, line 32: The convention is to use camel case, not snake case.

m0dB:
@ywwg I think this should be targeting main. 

Are you planning to address the reviews? If you don't have time, I can do that too.

ywwg:
yup I can repoint to main.  I am back from travel so I will address the reviews this week

JoergAtGithub:
Is this supersedded by #13151, or is this additional?

ywwg:
we can close this one

PR #12548, opened 2024-01-11
add multi-track property editor / batch tag editor

### LINKED ISSUE #9023: batch tag editing
Date: 2017-12-28T08:54:34Z
Status: Confirmed
Importance: Low
Launchpad Issue: [lp1740375](https://bugs.launchpad.net/bugs/1740375)
Tags: library, metadata
Attachments: [amarok-batch-tag-editing.png](https://bugs.launchpad.net/bugs/1740375/+attachment/5028335/+files/amarok-batch-tag-editing.png)

---

Mixxx already has a detailed track properties dialog that can edit the tags of one track at a time. It would be nice if this could be used for batch tag editing. I like the way Amarok implements this. When multiple tracks are selected in Amarok and right clicked on to bring up the track properties dialog, any tags that are common to all selected tracks are directly editable. To edit tags of tracks that are different between tracks, a "Per track" checkbox has to be enabled, which lets the user select individual tracks to edit without having to repeatedly close and open the properties dialog.

ronso0:
I have been wanting this desperatly for ages for unifying track comments in Mixxx because the alternative workflows are a PITA, e.g. copy/paste per track, or use a tag editor and reload tags in Mixxx.

Intended to fix #9023

This is a modded/stripped clone of DlgTrackInfo, just without DlgTagFetcher and the BPM tab.
It's is not polished yet and I'll push fixes, so please don't take a too close look at the code, yet ; )

* for editable fields <various>
* Cover show, clear and set works (shown empty if covers differ, maybe we can show a special placeholder)
* Star rating works (is 0 if values differ)

### Currently supported properties
**Editable fields:**
- [x] title
- [x] artist
- [x] album title
- [x] album artist
- [x] genre
- [x] grouping (?)
- [x] comment
- [x] track color (rainbow button for multiple track colors)
- [x] musical key
- [x] cover art
- [x] star rating

**Static labels:**
- [x] BPM: show range of loaded track BPM, or nothing if all are 0 / invalid
- [x] bitrate: show common value or range
- [x] duration: show common value or range
- [x] directory: show common directory or _<various>_ (might be helpful to grab info like album title, catalog id etc.)
- [x] "Open in File Browser" button is enabled if all tracks are in the same dir

**Special fields, to be discussed:**
* BPM: I only see few valid uses cases for setting the BPM of multiple tracks:
  * BPM are all horribly off, and scaling via track menu is not applicable for some reason
  * I sometimes explicitly set BPM to `1` (and lock) in order to 'flag' tracks as 'beatless' (athmo, fields recordings and alike) when I'm too lazy to add this to the comments.

**Nice to have:**
* placeholder for [multiple covers] ?
* reset button for each field? (quite some work, probably not worth it, there's the 'Reload from File' button)


ronso0:
Wooh, that was easier than I expected when I took the first look a while back (actually half the time I spent on debugging stupid mistakes..)

This seems to work well for all fields, except cover reload.
(and except key and BPM, I pushed those back since I don't consider them useful here anyway)

fwcd:
Wow, very nice to see progress on this. Got to play around with the PR once I find some time.

acolombier:
@ronso0 is this ready for review? I can see couple of `TODO`  and commented code. Is there some clean-up to be expected before starting review/testing?

ronso0:
Thanks for taking a look!

I did remove commit with (cover) trace logging and temp. stuff. There are small spots I _may_ polish (turn DEBUG_ASSERT into VERIFY_OR_DEBUG_ASSERT) but that doesn't matter much. For further changes I'll push small commits and maybe squash them later on.

ronso0:
Currently rating and color were stored immediately (not saved to tracks of course).
I'm about to change that so all properties are handled consistently, i.e. apply all changes to track records only when saving.

ronso0:
Please let me know when someone starts reviewing.

acolombier (commented):
I just got a chance to test this feature. Overall, everything works fine, although there seems to remain some warning noise when editing fields.

Only got one issue with the current state: differing field separated by `;` are misleading as it get's unclear what happens if you change a "field" (text between two `;` (disclaimer, it updates all tracks field with the summary, now got 10 tracks with the title `My name #1;My name #2;My name #3;My name #4;My name #5;My name #6;My new name #7;My name #8;My name #9;My name #10` :sweat_smile: ). Perhaps, you could make this field read-only? Or make so when it gets edited, the summary disappear (like a placeholder), so it gets clear that all field will be overwritten with the new value. Finally, you could get the individual value. contained between each `;`, although probably overkill and counter-intuitive.

ronso0:
Thanks for testing!

> there seems to remain some warning noise when editing fields.

What are you referring to?

ronso0:
Re: multi-valued fields
I thought, as first implementation, the italic/regular font would be sufficient to signal the IS and NEW (edit) state.
I don't know which separator to choose that is _not_ contained in any tags in order to separate the edited multi-value content and assign it to the correct tracks. 

Ideally, differing values would be indicated and for each field (not directly editable) there would be a dropdown with all available values. (like in Kid3 for example)

We should keep in mind that the main purpose of this batch editor (IMHO) is to assign common value(s) to certain field(s), not edit individual fields per track (that can be done in the library ,or with the single-track editor).

ronso0:
> There was some log warning with just . in the line, not sure where this comes from but looks like that might not have been introduced with your PR so I guess this is out of scope.

Hmm, this might be debug output related to window switching, see #13030

ronso0:
> How about using kVarioustext + italic here too if there is multiple value? During my test, this field was very useful anyway as the value was too large to be displayed. Since you are using this value for RO attributes, perhaps that could be a way to keep things consistent.

Sounds reasonable. I'll try that.
_various_ + combobox with all options would be cool.

ronso0:
> _<various>_ + combobox with all options would be cool.

This required some fiddling with focus events, but it works :tada: 
I hope UX is good now. I tested all combinations of the belwo steps and didn't encounter issues.

**When testing please also:**
* clear _<various>_ (should clear field for all tracks)
* restore _<various>_ manually (should restore the italic, dim _<various>_ placeholder)
* Reset button
* Reload from file tags

ronso0:
Only unhandled field is now Key.
Will take a look.

ronso0:
I removed the hackish way for clearing multi-value boxes via events etc. There were some situations where restoring values sfrom files didn't work correctly.

This now simply uses  a dedicated  Clear item and everything works as desired.

Only TODO left is adding the key validation when users enter a key text manually.

acolombier:
Just had a quick look and it looks much better now, thank you!

Let me know when it is ready for testing.

ronso0:
It's already ready for testing, I only need to implement the key validation.

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,833 @@
...
+    trackColorDialogSetColorStyleButton(mixxx::RgbColor::nullopt());
+    m_pWCoverArtLabel->loadTrack(TrackPointer());
+    m_pWCoverArtLabel->setCoverArt(CoverInfo(), QPixmap());
+}
+
+void DlgTrackInfoMulti::connectTracksChanged() {
+    for (auto it = m_pLoadedTracks.constBegin();
+            it != m_pLoadedTracks.constEnd();
+            it++) {
+        const auto pTrack = it.value();
```
daschuer, line 602:
```suggestion
    for (const auto& pTrack : std::as_const(m_pLoadedTracks)) {
```

daschuer (changes requested):
Some first comments

daschuer:
Just tested and it works nice. Thank you very much. 
Does it make sense to gray out the combo box when there is no value to select?

daschuer:
This happens after picking a new cover for two tracks: 
```
critical [Main] DEBUG ASSERT: "!m_pLoadedTracks.isEmpty()" in function void DlgTrackInfoMulti::slotCoverFound(const QObject*, const CoverInfo&, const QPixmap&) at ./src/library/dlgtrackinfomulti.cpp:797
```

ronso0:
> Does it make sense to gray out the combo box when there is no value to select?

Not sure what you mean. If there is only one value the combobox is "disabled" in the sense that there is no drop-down.
If a field is empty for all tracks users may want to fill it.

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,833 @@
...
+    auto font = pEditor->font();
+    if (font.bold() == bold) {
+        return;
+    }
+    font.setBold(bold);
+    pEditor->setFont(font);
+}
+
+/// Check if the text has been edited, i.e. is not <various>
+std::optional<QString> validEditText(QComboBox* pBox) {
```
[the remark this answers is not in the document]
ronso0: Oh, thanks for the hint!
I vaguely remembered there issues with isNull() / isEmpty() in other code so I didn't consider it here.

daschuer:
It was just out of my disappointment to click on the combo box button and nothing happens. No strong demands. 

I have actually a bigger demand to have a combo box with all Genres in the library, to avoid issue like "Hip Hop" vs. "Hiphop" vs "Hip-Hop" or such. 
Or just offer this list:
https://en.m.wikipedia.org/wiki/List_of_ID3v1_genres

ronso0:
> It was just out of my disappointment to click on the combo box button and nothing happens.

I understand.
Unfortunately there is no (official) way to disable the drop-down arrow, only the entire combobox (no edit possible). I briefly looked into making the combobox appear as QLineEdit by removing the drop-down button via stylesheet (no icon, zero width) but that doesn't work.

Ideally we'd dynamically switch between QLineEdit (no value or common value) and QComboBox (multiple values).

For now the initially empty box vs. _<various>_  vs. _Text value_ is good enough IMO.

ronso0:
> This happens after picking a new cover for two tracks:
> 
> ```
> critical [Main] DEBUG ASSERT: "!m_pLoadedTracks.isEmpty()" in function void DlgTrackInfoMulti::slotCoverFound(const QObject*, const CoverInfo&, const QPixmap&) at ./src/library/dlgtrackinfomulti.cpp:797
> ```

Can't reproduce. How did you manage to get this? Select a new cover in the dialog, or in the table while the dialog is open?

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,833 @@
...
+    QStringList composers;
+    QStringList grouping;
+    QStringList years;
+    QStringList keys;
+    QStringList nums;
+    QStringList comments;
+    QList<double> bpms;
+    QList<uint32_t> bitrates;
+    QList<double> durations;
+    QStringList samplerates;
```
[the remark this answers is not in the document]
ronso0: Yeah, though that requires some refactoring / more complexity IIUC.
Someting like `convertToQStringConvertible` (actually exactly that) from #13236 would help me a lot I think.

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,833 @@
...
+const QString kVariousText = char('<') + QObject::tr("various") + char('>');
+const char* kOrigValProp = "origVal";
+const QString kClearItem = QStringLiteral("clearItem");
+
+/// If value differs from the current value, add it to the list.
+/// If new and current are identical, keep only one. Later on we can use the
+/// item count to maybe join the list and format the font accordingly.
+/// List can be a QStringList or a QList<double>
+template<typename T>
+void appendToTagList(QList<T>* pList, const T& value) {
```
ywwg, line 35: it is not clear from the function name that this does special management to keep the list unique. Maybe "appendUniqueToTagList" or "addToTagSet" or "includeInTagSet" would make that more clear

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,833 @@
...
+    }
+}
+
+} // namespace
+
+DlgTrackInfoMulti::DlgTrackInfoMulti(UserSettingsPointer pUserSettings)
+        // No parent because otherwise it inherits the style parent's
+        // style which can make it unreadable. Bug #673411
+        : QDialog(nullptr),
+          m_pUserSettings(std::move(pUserSettings)),
```
ywwg, line 108: std::move does not make sense here because this constructor is called with m_pConfig elsewhere, and that m_pConfig is still owned by the caller, so it cannot be moved into dlgtrackinfomulti.  in general, std::move is used in situations where the object being passed into the function call will *never again* be used by the caller.  This is not the case for the call pattern here, so I believe this std::move will have no effect, and it can be removed.  (UserSettingsPointer is just a shared pointer anyway, it is meant to be copied around and shared, not moved)

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,833 @@
...
+
+    connect(btnImportMetadataFromFile,
+            &QPushButton::clicked,
+            this,
+            &DlgTrackInfoMulti::slotImportMetadataFromFiles);
+
+    QList<QComboBox*> valueComboBoxes;
+    valueComboBoxes.append(txtArtist);
+    valueComboBoxes.append(txtTitle);
+    valueComboBoxes.append(txtTitle);
```
ywwg, line 158: title appended twice?

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,833 @@
...
+    // Block signals to not set the 'modified' flag.
+    m_pWStarRating->blockSignals(true);
+    m_pWStarRating->slotSetRating(commonRating);
+    m_starRatingModified = false;
+    m_pWStarRating->blockSignals(false);
+
+    // Same procedure for the track color
+    mixxx::RgbColor::optional_t commonColor = m_trackRecords.first().getColor();
+    bool multipleColors = false;
+    for (auto& rec : m_trackRecords) {
```
ywwg, line 307: probably want const auto& since you don't want to modify rec

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,833 @@
...
+    // Paints a rainbow gardient in case of multiple colors.
+    trackColorDialogSetColorStyleButton(commonColor, multipleColors);
+    m_colorChanged = false;
+    m_newColor = mixxx::RgbColor::nullopt();
+
+    // And the track directory
+    QString commonDir = m_pLoadedTracks.constBegin().value()->getFileInfo().canonicalLocationPath();
+    QStringList dirs;
+    dirs.append(commonDir);
+    for (auto dirIt = m_pLoadedTracks.constBegin();
```
ywwg, line 324: I think std::as_const accomplishes this?

ywwg (changes requested):
I am excited about this feature!  Do you still plan to add more fixes and address the TODOs for 2.5?

a few notes on C++ style to get started.

src/library/dlgtrackinfomulti.cpp
ywwg: I agree with daniel's comment that these should be QSets, and then use an adapter to convert them into Lists when needed. it feels brittle otherwise

ronso0: Jup, that's exactly what I did. Had to deal with some type difficulties but managed it with `convertToQStringConvertible` from #13236

ronso0: "addToTagSet" sounds good.

src/library/dlgtrackinfomulti.cpp
ronso0: I worked off DlgTrackInfo without questioning the basic structure, so I guess it can removed there as well.

src/library/dlgtrackinfomulti.cpp
ronso0: yup, as Daniel mentioned above.

daschuer:
> Can't reproduce. How did you manage to get this? Select a new cover in the dialog, or in the table while the dialog is open?

I have selected a new cover from the cover art context menu file picker in the new dialog. It was open for two files.

src/library/dlgtrackinfomulti.cpp
daschuer: This does not move the ownership, because it is a shared pointer. It moves the shared pointer to avoid reference counting. So it is a valid approach. 

This follows the pattern to copy objects (in this case the shared pointer) by value parameters and then move them to the final destination.

ywwg: so you're moving the SharedPointer object from the constructor function parameter to the object being created... doesn't the parameter to the constructor go out of scope anyway, so there's no reference counting issue?  it still seems like overkill for such a light object, but ok, we can leave it.

ywwg:
I'll also try actually building/using this.  I also do a lot of track maintenance and this will be super handy

src/library/dlgtrackinfomulti.cpp
daschuer: I agree that this is micro optimizing this case. 
But we have started to implement this pattern everywhere following the rule of thump.
* const ref for local read access
* value + move when kept as member 
* raw pointer, for write access

It would be interesting if the compiler is able to optimize the reference counting away without the move. I am afraid it cannot. 
The reference counting is not cheap, because memory fences are involved.

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,835 @@
...
+                mixxx::Duration::formatTime(durList.last()));
+    } else {
+        txtDuration->setText(mixxx::Duration::formatTime(*durations.constBegin()));
+    }
+}
+
+template<typename T>
+void DlgTrackInfoMulti::addValuesToComboBox(QComboBox* pBox, QSet<T>& values, bool sort) {
+    // Verify that T can be used for pBox->addItem()
+    DEBUG_ASSERT(!convertToQStringConvertible(*values.constBegin()).isNull());
```
ronso0, line 461: b**shit, will add a dedicated helper.

src/library/dlgtrackinfomulti.cpp
```diff
@@ -17,24 +17,24 @@
...
 /// item count to maybe join the list and format the font accordingly.
-/// List can be a QStringList or a QList<double>
 template<typename T>
-void appendToTagList(QList<T>* pList, const T& value) {
+void addToTagSet(QSet<T>* pList, const T& value) {
     if (pList->isEmpty() || !pList->contains(value)) {
-        pList->append(value);
+        pList->insert(value);
     }
 }
```
ywwg, line 40: Now that we are using QSet, you don't need this function at all ("inserts **if value isn't already in the set**, and returns an iterator pointing at the inserted item.") https://doc.qt.io/qt-6/qset.html#insert

src/library/dlgtrackinfomulti.cpp
```diff
@@ -170,22 +172,63 @@ void DlgTrackInfoMulti::init() {
...
+    // regular tags, the two buddies require a special setup:
+    // * txtCommentBox is not editable
+    // * if an item is selected in txtCommentBox, the text is shown in txtComment
+    // * for multiple values, we show the <various> placeholder also in txtCommentBox
+    // This also requires some special handling in saveTracks().
+    txtCommentBox->setInsertPolicy(QComboBox::NoInsert);
+    connect(txtCommentBox,
+            &QComboBox::currentIndexChanged,
+            this,
+            [this](int idx) {
```
ywwg, line 205: note the clazy warning -- idx not used (so I think you can do `int _` ? )

ywwg (commented):
this is cool! I was able to make some bulk changes and it makes sense.  I like "various" for explaining that different tracks already have values set.

One bug, I tried to set the Color of multiple tracks and it did not apply

src/library/dlgtrackinfomulti.cpp
ronso0: thanks, wasn't aware of that. just read that QSet may contain duplicates.

src/library/dlgtrackinfomulti.cpp
ronso0: debug leftover, will remove

src/library/dlgtrackinfomulti.cpp
```diff
@@ -0,0 +1,886 @@
...
+#include "library/library_prefs.h"
+#include "moc_dlgtrackinfomulti.cpp"
+#include "preferences/colorpalettesettings.h"
+#include "sources/soundsourceproxy.h"
+#include "track/beatutils.h"
+#include "track/track.h"
+#include "util/color/color.h"
+#include "util/datetime.h"
+#include "util/duration.h"
+#include "util/optional.h"
```
daschuer, line 19: This can be probably removed?

ronso0:
Just a note about an annoying QComboBox bug I noticed:
with long comment lines the popup¹ is much wider than necessary (word-wrap enabled by default).
The QListView and the items have the correct width (max width of the combobox = dialog width) but its _container_ is muuch wider.

It seems the popup width is derived from the box' width hint, i.e. the width of the longest item text, but apparently that doesn't take word-wrap into account. I.e. the multi-line text is displayed correctly but the popup covers the entire screen width. 
I didn't manage to fix it with various elide / sizeAdjust / size policies.

Not sure how to fix this (and I don't expect Qt to fix this soonish). IIUC an custom item delegate wouldn't fix it since the items are actually good. We could sub-class QComboBox and overwrite only the sizeHint / widthHint. I'll do a quick test. If that fails we'll just have to take it as it is.

ronso0:
I managed to work around this bug by setting the max width for the view container, too. It's updated in resizeEvent().
Also I moved the comment combobox up (next to the 'Comments:' label) so it's clear it holds the comments.

This is finally ready to roll I think.

src/library/dlgtrackinfomulti.cpp
ronso0: Yup, removed.

ronso0:
> One bug, I tried to set the Color of multiple tracks and it did not apply

Also fixed, the 'modified' flag was not set.

daschuer (approved):

daschuer:
I was able to reproduce the DEBUG_ASSERT 
Here is the fix: https://github.com/ronso0/mixxx/pull/73

ywwg (approved):
my notes are addressed, thank you!

ronso0:
> I was able to reproduce the DEBUG_ASSERT Here is the fix: [ronso0#73](https://github.com/ronso0/mixxx/pull/73)

Thanks for your investigation!
I could reproduce with Ok (tried only Apply earlier).

Please correct me if I'm wrong but I think the actual issue is `clear()` which clears the QHash with loaded tracks and the dialog, even though it is now hidden, can still receive the `CoverArtCache::coverFound` signal.

I adopted it from DlgTrackInfo but actually we don't need it at all. These are the relevant cases:
* load tracks from WTrackTableView (via WTrackMenu): the object hold by the unique ptr `m_pDlgTrackInfoMulti` is deleted if tracks are loaded while a DlgTrackInfo/~Multi is open
* closed with X or Cancel: no issue, dialog is deleted by WTrackMenu
https://github.com/mixxxdj/mixxx/pull/12548/files#diff-848ff12a6bbb8c39f28e2a6be33ca2bbbe31f53e6ba550b2f6df7b8bdfaecf1fR2427-R2434
* clicked Okay (with new cover): `CoverArtCache::coverFound` signal is received but who cares, dialog is deleted soon after being closed

daschuer:
I am afraid we need clear(), to release the reference and in turn write the changes to the track file. 
Since there can always be a stray CoverFound signal, we may just deal with it and remove the DEBUG_ASSERT()

ywwg:
build warning:

```
AutoUic: /home/owen/src/github/mixxx/src/library/dlgtrackinfomulti.ui: Warning: The name 'horizontalLayout_5' (QHBoxLayout) is already in use, defaulting to 'horizontalLayout_51'.
/home/owen/src/github/mixxx/src/library/dlgtrackinfomulti.ui: Warning: The name 'horizontalSpacer_6' (QSpacerItem) is already in use, defaulting to 'horizontalSpacer_61'.
```

ronso0:
> I am afraid we need clear(), to release the reference ...

Okay, we can do that in the destructor.

> ... and in turn write the changes to the track file.

I don't quite understand?

daschuer:
I think the dialog outlives longer, is only hidden after OK. We need to take care that m_pLoadedTrack becomes nullptr, this event will then trigger the write to  the file itself (if this is the last instance) 

Clear() in the destructor us redundant, because it is cleared anyway.

daschuer:
Probably a straight forward solution is to just remove the assertion.

ronso0:
> I think the dialog outlives longer, is only hidden after OK.

Maybe you overlooked it, I explained when & how the dialog is deleted here https://github.com/mixxxdj/mixxx/pull/12548#issuecomment-2119463777



> We need to take care that m_pLoadedTrack becomes nullptr, this event will then trigger the write to the file itself (if this is the last instance)

I don't understand. Tracks are updated in `saveTracks()` by replacing each TrackRecord. (there is no `TrackPointer m_pLoadedTrack` in `DlgTrackInfoMulti`). 
I though your comment was about explicitly clearing the QHash<TrackId, TrackPointer> (which also happens if the parent dialog is deleted, no?).

I appended a fixup that transforms the DEBUG_ASSERT (which I couldn't trigger aymore with the current state btw).
I also added a demo commit just to `VERIFY_OR_DEBUG_ASSERT(!m_pLoadedTracks.isEmpty())`.

daschuer:
Oh, you are right. Now everything works flawlessly. 
Can you rebase the fixup and than we can merge.

ronso0:
Alright.

I noticed some issues I need to fix, though:
* selecting an empty item in the key box does not clear the key
* dialog is still a QTabWidget, though removing the 'Summary' tab miraculously that creates some size policiy / strecth issues with some comboboxes (I love Qt)

ronso0:
Sooo, the key bug is fixed.

> removing the 'Summary' tab miraculously that creates some size policiy / strecth issues with some comboboxes (I love Qt...)

puuuh, hours of head scratching and fiddling with various SizePolicy combos...
Think I got it right now, at least it somewhat looks like the single-track Track Info dialog.
The "Clear" item is now only an icon, the (X) which is the default Clear icon for the search bar.

I'd appreciate another round of review if someone has the time to do so, and get this in to 2.5-alpha finally.
Thank you!

daschuer:
Thank you. Lets merge and test it during beta.

PR #12549, opened 2024-01-11
add `qt6-qpa-plugins` to dependencies

### LINKED ISSUE #12525: 2.5-alpha won't start on Uuntu 22.04 due to missing Qt platform plugins

ronso0:
Should fix #12525


daschuer (changes requested):
Thank you. This need to go also to the tools/Debian... And packaging/Debian/control.in 

Unfortunately it is conflicting with 
https://github.com/mixxxdj/mixxx/pull/12535
Targeting a similar issue. 

Should I take over, on top of that and you review and merge?

ronso0:
I thought about control.in first and noticed that it uses  CPACK_DEBIAN_PACKAGE_DEPENDS populated in cmakelist. It's not a build dependency IIUC.

Please take over if you're sure it's required.

daschuer:
Oh, yes, thank you for the head up. I think I need to adjust my PRs accordingly.

PR #12259, opened 2023-11-04
[Traktor S4 Mk3] Screen support

acolombier:
## TODO

- [ ] Update the documentation for the controller
- [ ] Add markers (loops, cues) on the scrolling waveform

---

This depends on #12199 and @m0dB's `qml_scrolling_waveform` branch

PR #3967, opened 2021-06-08
QML: Add scrolling waveforms

Holzhaus:

PR #12557, opened 2024-01-12
[Traktor S4 Mk3] Screen support

acolombier:
This is related to PR #12259.

New mapping feature:
- Record button can be use to toggle scrolling waveform
- Mute button can be use to minify widget and get full screen artwork
- New shortcut to set loop backward (the playposition is the endpoint instead of start point)
- Ability to change hotcue color with shortcut

Depends of:
- #3967 
- #11407 
- #12199 
- #12558
- #12559
- #12745

A branch with all dependencies can be used to test/use this feature while the above dependencies gets reviewed. Branch is `acolombier:traktor-s4-mk3-updated`

TODO:
- Documentation update


ywwg:
can you pull in the latest main changes to resolve the conflict?

ywwg:
I think my next request would be for overview explanatory comments on all the new objects / files -- mixxx project has a bad habit of not doing code documentation and I'd like to fix that.

ywwg:
(ah, I'm getting confused between branches, no c++ in this PR)

acolombier:
@ywwg Yes, this PR is only the mapping part (QML + JS), though I can take your comment for #11407 

That being said, I agree with you - enforcing code documentation would be a great thing, even for QML. I'm just not so sure how it would look like tho. But I'm keep to do some work on this, so the first QML mapping kinda set standards for future contribution

acolombier:
Cannot be completed due to rejection of #12199

Will be kept updated in `acolombier:traktor-s4-mk3-updated` for anyone interested

ywwg:
Per discussion, this work is not abandoned but on the back burner while we do a proper design for controller data sharing.

PR #11378, opened 2023-03-15
Implement CO to set all QuickEffectRack Effects to the same Chain.

Swiftb0y:
Fixes #11371 

Current Issues to discuss:
- [ ] `[QuickEffectRack1]` is a new CO group, not sure if thats desired or where else to put the CO.
- [x] `[QuickEffectRack1]`,`loaded_chain_preset` is zero-based while their `[QuickEffectRack1_[ChannelN]]` equivalent is 1-based. Considering the first real preset starts at 2 now (since 1 is the empty chain introduced in #10859). 
  - [x] When making the CO 0-based again ,`[EqualizerRack1]` and `[EffectRack1]` `loaded_chain_preset` now get initialized to weird, very large values for some reason.

PR #12561, opened 2024-01-13
unify `[effect chain],loaded_chain_preset` control (all 0-indexed now)

ronso0:
In #10859 I introduced an inconsistency, `[QuickEffectRack1_[ChannelN]]` loaded_chain_preset is offset by one compared to `[EffectRack1_EffectUnitN]` because of the prepended empty preset.

**Fix1**
The '---' preset is now also available for standard chains, e.g. via the effect chain preset button (⚙️ icon in effect units), which removes the offset. As before, selecting this preset clears the effect chain and `loaded_chain_preset` is set to `0` (works with **all** effect chains).
_(and `-1` for nameless items, the default state of standard effect units and the main output effect unit)_
_(fixes the issue discovered by @Swiftb0y in https://github.com/mixxxdj/mixxx/pull/11378#issue-1626390581)_
I didn't do this before to avoid the UX issue I tried to describe here https://github.com/mixxxdj/mixxx/pull/11378#issuecomment-1889794345
which is now fixed by

**Fix2**
When an effect is about to be loaded while the '---' preset is loaded (and slots are empty)
* an empty, nameless preset is loaded (`loaded_chain_preset` is set to `-1`)
* the effect is loaded
* '---' is again available for clearing the effect chain

#### Don't store/reload '---' preset
* clear an effect unit by selecting the '---' preset in the chain preset menu
* `loaded_chain_preset` is `0`
* restart Mixxx
* check
   * _no preset_ is selected in the chain preset menu
   * `loaded_chain_preset` is `-1`

#### Don't allow loading an effect into the '---' preset
* clear an effect unit by selecting the '---' preset in the chain preset menu
* trigger loading an effect (via effect selector or `[(unit)_Effect1],next_/prev_effect`/`effect_selector`
* check
   * desired effect is loaded
   * _no preset_ is selected in the chain preset menu
   * `loaded_chain_preset` is `-1`

### DIFF
--- a/src/effects/presets/effectchainpresetmanager.cpp
+++ b/src/effects/presets/effectchainpresetmanager.cpp
@@ -394,6 +394,20 @@ void EffectChainPresetManager::setPresetOrder(
                 m_effectChainPresets.value(chainPresetName));
     }
 
+    // After having changed the presets order in DlgPrefEffects, we received the
+    // new lists. The '---' was not displayed there, so it's not in the list.
+    // Make sure the empty preset '---' is the first item in the list.
+    const auto& pEmptyPreset = m_effectChainPresets.value(kNoEffectString);
+    VERIFY_OR_DEBUG_ASSERT(pEmptyPreset) {
+        return;
+    }
+    int index = m_effectChainPresetsSorted.indexOf(pEmptyPreset);
+    if (index == -1) { // not in list, re-add it
+        m_effectChainPresetsSorted.prepend(pEmptyPreset);
+    } else if (index != 0) { // not first item, move to top
+        m_effectChainPresetsSorted.move(index, 0);
+    }
+
     emit effectChainPresetListUpdated();
 }
 
@@ -409,7 +423,9 @@ void EffectChainPresetManager::setQuickEffectPresetOrder(
                 m_effectChainPresets.value(chainPresetName));
     }
 
-    // Ensure empty '---' preset is the first list item
+    // After having changed the presets order in DlgPrefEffects, we received the
+    // new lists. The '---' was not displayed there, so it's not in the list.
+    // Make sure the empty preset '---' is the first item in the list.
     const auto& pEmptyPreset = m_effectChainPresets.value(kNoEffectString);
     VERIFY_OR_DEBUG_ASSERT(pEmptyPreset) {
         return;
@@ -614,6 +630,7 @@ void EffectChainPresetManager::resetToDefaults() {
     // Re-add the empty chain preset
     EffectChainPresetPointer pEmptyChainPreset = createEmptyChainPreset();
     m_effectChainPresets.insert(pEmptyChainPreset->name(), pEmptyChainPreset);
+    m_effectChainPresetsSorted.prepend(pEmptyChainPreset);
     m_quickEffectChainPresetsSorted.prepend(pEmptyChainPreset);
 
     emit effectChainPresetListUpdated();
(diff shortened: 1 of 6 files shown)


ronso0:
IMO this is definitely a 2.4.0 candidate since it restores the API consistency.

I'm sorry for not taking care of it sooner, #10859 got too big / took too long, I had no motivation to fix it right away.
Now, with a bit of distance, the fix was pretty easy.

Swiftb0y:
Thank you. I'll try to take a look ASAP.

src/effects/presets/effectchainpresetmanager.cpp
```diff
@@ -394,6 +394,20 @@ void EffectChainPresetManager::setPresetOrder(
                 m_effectChainPresets.value(chainPresetName));
     }
 
+    // After having changed the presets order in DlgPrefEffects, we received the
+    // new lists. The '---' was not displayed there, so it's not in the list.
+    // Make sure the empty preset "---" is the first item in the list.
```
ywwg, line 399: please use consistent quote marks in the comments just for neatness

src/effects/presets/effectchainpresetmanager.cpp
```diff
@@ -896,6 +931,10 @@ void EffectChainPresetManager::saveEffectsXml(QDomDocument* pDoc, const EffectsX
     QDomElement chainsElement = pDoc->createElement(EffectXml::kChainsRoot);
     rackElement.appendChild(chainsElement);
     for (const auto& pPreset : std::as_const(data.standardEffectChainPresets)) {
+        // Don't store the name '---', see readEffectsXml() for explanation.
+        if (pPreset->name() == kNoEffectString) {
```
ywwg, line 935: if we don't want to store the noeffect, shouldn't we skip it (continue) instead of setting the name empty and adding it?

ywwg: (as you did below)

ywwg (commented):
if I understand correctly, the major issue is that we want effects to be 0-based, but we also want a nullable-style "no effect" option.  Is that right?  I don't love the hackiness of this solution but at this late date I don't have a better idea.  Something to look at later, perhaps.

ronso0:
> I don't love the hackiness of this solution but at this late date I don't have a better idea.

I understand. IMO the hackiness (filtering '---' during preset export/import,  Read/SaveEffectsXml and in the preferences) stems from the fact that we don't want to expose the '---' in the GUI except in the chain selectors and the chain preset menu.

IMO this fix can be considered a hack only because the preset switch happens _after_ the effect was loaded into the '---' preset.
Would be better to that when the effect change is requested.
Just now I realize `EffectSlot` knows its parent `EffectChain` and also has a pointer to the preset manager, so since only `EffectSlot::loadEffectInner` emits the `effectChanged()` signal we care about, the preset swap can very well be triggered from there.

I'll push fixups in response to your review, and another commit moving the preset swap to `EffectSlot`.

src/effects/presets/effectchainpresetmanager.cpp
ronso0: I added this only as safeguard to not dump loaded effects in the (unexpected) case that there are effects loaded despite the '---' preset name.
I can turn it into a VERIFY_OR_DEBUG_ASSERT.

src/effects/presets/effectchainpresetmanager.cpp
ronso0: :+1:

ronso0:
Okay, I'm happy with this now. Definitely less hacky. Works as desired.
I also added a **Tests** section to the PR description.

daschuer:
@ywwg Is this ready for merge form your perspective?

src/effects/presets/effectchainpresetmanager.cpp
ronso0: Fixed by the last commit https://github.com/mixxxdj/mixxx/pull/12561/commits/d9b104bee7a6a61909e81f618a5367ac281693d0

daschuer (approved):
The code Looks good and it works without any issues. Than you.

PR #12578, opened 2024-01-15
Keyboard sheet update

ronso0:
* add Return: preview load+play / preview stop / track info dialog
* fix 'rev' (is now back 1beat)
* update manual links


daschuer:
Is this a 2.4 PR?

ronso0:
Yes, good catch. Changed the base.

daschuer (approved):
Looks good. Thank you.

ronso0:
Thanks, that was quick!

Oh, and one thing to keep in mind @mixxxdj/developers:
when the mapping SVG was touched and saved as "Optimized SVG", double-check all layers are still there.
IIUC that may discard layers which makes the export workflow (for the manual)... much less fun.

Issue #12560, opened 2024-01-12
COs not being deleted -- DEBUG_ASSERT on skin reload


ronso0:
### Bug Description

I've hit this a few times, though never cared to investigate.

```
warning [Main] ControlObject "[Channel1]" "stars_up" already created
DEBUG ASSERT: "!"pCreatorCO != nullptr, ControlObject already created"" in function static QSharedPointer<ControlDoublePrivate> ControlDoublePrivate::getControl(const ConfigKey&, ControlFlags, ControlObject*, bool, bool, bool, double) at src/master/src/control/control.cpp:173
```

It happened with these two, both COs are created by widgets
`[Channel1],stars_up` from WStarRating [here](https://github.com/mixxxdj/mixxx/blob/ed6602666af40bf4fbeda32e5c9ae31b587b65de/src/widget/wstarrating.cpp#L20-L21)
`[Channel1],end_of_track` from VisualsManager [here](https://github.com/mixxxdj/mixxx/blob/ed6602666af40bf4fbeda32e5c9ae31b587b65de/src/waveform/visualsmanager.cpp#L16) owned by MixxxMainWindow

### Version

2.4 + 2.5

ronso0:
Fixed by #12591

PR #12591, opened 2024-01-19
avoid issue with `stars_up/_down` ControlObjects

ronso0:
Supposed to fix ~~one half of~~ #12560 

Moved the COs from `WStarRating` to `BaseTrackPlayerImpl`, and the skin parser now connects this extra signal.
The reason for #12560 is not clear to me, but this is safer because the only responsibility of the widgets is now to listen to change request signals.

The first commit is picked from #12582

**edit** @Swiftb0y mentioned a potential cause https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/help.20with.20unique_ptr.20.2F.20ControlObjects.20not.20being.20destroyed

### DIFF
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -45,7 +45,7 @@ DlgTrackInfo::DlgTrackInfo(
           m_tapFilter(this, kFilterLength, kMaxInterval),
           m_pWCoverArtMenu(make_parented<WCoverArtMenu>(this)),
           m_pWCoverArtLabel(make_parented<WCoverArtLabel>(this, m_pWCoverArtMenu)),
-          m_pWStarRating(make_parented<WStarRating>(nullptr, this)),
+          m_pWStarRating(make_parented<WStarRating>(this)),
           m_pColorPicker(make_parented<WColorPickerAction>(
                   WColorPicker::Option::AllowNoColor |
                           // TODO(xxx) remove this once the preferences are themed via QSS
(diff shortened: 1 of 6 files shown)


daschuer (commented):
This looks good. What is the reason not always set up the stars_up/down controls in the BaseTrackPlayerImpl() constructor? This way it is always available for controllers and keyboards.

ronso0:
You mean for all players, not just decks?
These controls are useful if there's a star widget that gives visual feedback. No widget, no controls needed. That was my intention.

daschuer:
> You mean for all players, not just decks? 

That was my first idea actually but I am unsure.  We can move alternative to the Deck() constructor. 

The other aspect is, that we normally instantiate the COs without GUI. Here the skin takes control which creates a asymmetry compared to the "track_color" CO for instance.

ronso0:
Verifying the track color is easier.
Otoh, setting the rating currently requires a StarRating instance (in WStarRating) to verify values. An instance only because only that knows the current max rating, though this flexibility is actually not required since WStarRating always sets it to 5.

So, yeah, the controls _could_ be created unconditionally but for that to be useful some of the rating logic has to moved to BaseTrackPlayer.

Tbh I'm not really up to that right now, but I'll definitey look into it after this was merged.

daschuer:
Ah OK that the issue, the star logic is anyway in the widget. Yes we need to change that on the long run to be prepared for QML. 
But for now its good enough. Thank you.

ronso0:
Okay, thanks for reviewing.

ronso0:
I took a look and it was less work than expected. #12648

PR #12593, opened 2024-01-19
move COs from DeckVisuals/VisualsManager to BaseTrackPlayerImpl

ronso0:
This is an attempt to fix a warning I encountered a few times, which was something like
```
[warning] Control 'end_of_track' does not exist
```
Probably triggered when WOverview tried to create the `end_of_track` ControlProxy with ControlFlag::NoAssertIfMissing flag, while the ControlObject has not been created, yet.
IIRC I noticed this only with my personal branch with #3534 merged (which I like very much btw).

### The root cause (probably)
* `LegacySkinParser::parseSkin` sets `[App],num_decks` from 2 (default) to 4 after reading the skin attributes
* change request for `[App],num_decks` by `LegacySkinParser::parseSkin` signals `PlayerManager` to create a new `BasetrackPlayer`
* after the player has been created, [`PlayerManager` signals `VisualsManager`](https://github.com/mixxxdj/mixxx/blob/dc22fd5f963a7efc54cb2cd64c7251c180368295/src/mixxxmainwindow.cpp#L206-L229) to create new `DeckVisual` 
* so, could this be caused by `PlayerManager` living in a different thread than `VisualsManager` (with [#3534](https://github.com/mixxxdj/mixxx/pull/3534)), and the connection being implicitely a `Qt::AutoConnection` ([Qt::QueuedConnection](https://doc.qt.io/qt-6/qt.html#ConnectionType-enum) in this case), and `WOverview` trying to create the proxy _before_ `end_of_track` has been created by DeckVisual?

### Fix
* first, create `end_of_track` and other controls in BaseTrackPlayer
* then the proxies in WOverview and DeckVisual
* use `Qt::DirectConnection` for `PlayerManager::numberOfDecksChanged` -> `VisualsManager` (is this helpful??)

### DIFF
--- a/src/mixer/basetrackplayer.cpp
+++ b/src/mixer/basetrackplayer.cpp
@@ -205,9 +205,15 @@ BaseTrackPlayerImpl::BaseTrackPlayerImpl(
             this,
             &BaseTrackPlayerImpl::slotShiftCuesMillis);
 
-    // BPM of the current song
+    // BPM and key of the current song
     m_pFileBPM = std::make_unique<ControlObject>(ConfigKey(getGroup(), "file_bpm"));
+    m_pVisualBpm = std::make_unique<ControlObject>(ConfigKey(getGroup(), "visual_bpm"));
     m_pKey = make_parented<ControlProxy>(getGroup(), "file_key", this);
+    m_pVisualKey = std::make_unique<ControlObject>(ConfigKey(getGroup(), "visual_key"));
+
+    m_pTimeElapsed = std::make_unique<ControlObject>(ConfigKey(getGroup(), "time_elapsed"));
+    m_pTimeRemaining = std::make_unique<ControlObject>(ConfigKey(getGroup(), "time_remaining"));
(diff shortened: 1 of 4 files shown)


daschuer:
> use Qt::DirectConnection for PlayerManager::numberOfDecksChanged -> VisualsManager (is this helpful??)

I don't think so. It bypasses the qt thread handling and executes the connected slot in the signal thread. All explicit Qt::DirectConnection have a risk of a race condition if the signal is from different thread, like the controller thread. 
I had a brief look, an this should not be the case and Qt will always pick a Qt::DirectConnection anyway. So I would prefer to remove that to not suffer surprises, if our assumption was wrong.

daschuer:
I can confirm the issue and also confirm this PR solves it. Interesting that we did not had no user complains before.

daschuer (changes requested):
Please remove the explicit direct connections and then this is ready to go.

daschuer (approved):
Thank you.

PR #12595, opened 2024-01-19
Remove "RUNTIME_DEPENDENCY_SET mixxx"

daschuer:
It is created, but not installed.


JoergAtGithub:
I tested the installer from this PR and Mixxx still starts with ANGLE. Thank you!

PR #12602, opened 2024-01-20
scale library star rating with the library font

ronso0:
Supersedes #11772. Rebased, simplified, I think I addressed all of @daschuer's review comments.

Works nicely!
For me stars are slightly larger now (16px instead of 15px) with my default font Open Sans Regular 11pt.

__
~~The thing is: most users will see a different size than before since star size now depends on the library font.
This may be the system's default font or the user-set font (Library preferences).
One idea to lint this would be:~~
* ~~get the font/stars ratio the user has seen before (get height() of default app font)~~
* ~~multiply that with the current font's capital height or some other font metric~~

Too much magic.
**edit** maybe not. has to be done only once

I'd say stars size should be something in between **ooooo** and **OOOOO** (compared to the library text) to be visible and accessible.
What do you think? Does the current state fullfill this requirement?

### DIFF
--- a/src/library/stardelegate.cpp
+++ b/src/library/stardelegate.cpp
@@ -27,6 +27,8 @@ void StarDelegate::paintItem(
     paintItemBackground(painter, option, index);
 
     StarRating starRating = index.data().value<StarRating>();
+    // Scale star size with the library's current font height.
+    starRating.setStarBreadth(StarEditor::starBreadthFromFont(option));
     starRating.paint(painter, option.rect);
 }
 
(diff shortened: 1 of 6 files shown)


daschuer:
Here it looks now like this: 
grafik
before: 
grafik

src/library/stardelegate.cpp
```diff
@@ -27,13 +27,16 @@ void StarDelegate::paintItem(
...
+    // Scale star size with the library's current font height.
+    starRating.setStarBreadth(StarEditor::starBreadthFromFont(option));
     starRating.paint(painter, option.rect);
 }
 
 QSize StarDelegate::sizeHint(const QStyleOptionViewItem& option,
                              const QModelIndex& index) const {
     Q_UNUSED(option);
     StarRating starRating = index.data().value<StarRating>();
+    starRating.setStarBreadth(static_cast<int>(option.fontMetrics.height() / 2));
```
daschuer, line 39: Why don't we have `(mf.capHeight() + mf.ascent()) / 2` here?

daschuer (changes requested):

ronso0:
Thanks for testing!
I'll put this on hold for now. I did a lot of testing (code, back and forth) and it looks like I messed up something.

I'll move this to 2.4.1

Issue #12614, opened 2024-01-22
Skin: theme system can't see "skins:BaseTheme/secondfile.xml" refered to from a custom "skin:firstfile.xml"


mxmilkiib:
### Bug Description

Can't refer in a single custom theme file back to the root theme by using "skin:".

If the user customises an .xml that references to another file that only the root has, that is not in the subtheme directory, then "skin:" fails.

Temp fix is to replace some of the "skin:" with "skins:/Basetheme".

```
warning [Main] LegacySkinParser::loadTemplate - setContent failed see "/home/milk/.mixxx/skins/Deere-RekboxHighwave/skin:left_right_1state_button.xml" line: 0 column: 0
warning [Main] LegacySkinParser::loadTemplate - message: "QDomDocument::setContent: Failed to open device."
warning [Main] Skin parsing failed at skins:Deere/deck_looping_controls.xml:56 <Template>: Template instantiation for template failed: skin:left_right_1state_button.xml | ./src/skin/legacy/legacyskinparser.cpp:1735
warning [Main] Could not open template file: "/home/milk/.mixxx/skins/Deere-RekboxHighwave/skin:left_right_1state_button.xml"
warning [Main] QDomDocument called with unopened QIODevice. This will not be supported in future Qt versions.
warning [Main] QDomDocument::setContent: Failed to open device.
warning [Main] LegacySkinParser::loadTemplate - setContent failed see "/home/milk/.mixxx/skins/Deere-RekboxHighwave/skin:left_right_1state_button.xml" line: 0 column: 0
warning [Main] LegacySkinParser::loadTemplate - message: "QDomDocument::setContent: Failed to open device."
warning [Main] Skin parsing failed at skins:Deere/deck_looping_controls.xml:72 <Template>: Template instantiation for template failed: skin:left_right_1state_button.xml | ./src/skin/legacy/legacyskinparser.cpp:1735
warning [Main] Could not open template file: "/home/milk/.mixxx/skins/Deere-RekboxHighwave/skin:left_right_2state_button.xml"
warning [Main] QDomDocument called with unopened QIODevice. This will not be supported in future Qt versions.
warning [Main] QDomDocument::setContent: Failed to open device.
```

### Version

2.5-alpha

### OS

Linux

ronso0:
Hmm, this reveals multiple issues:
1. mass-replace in #12463 missed a lot of `<Template src="skin:` occurences
2. stylesheets also need to use `skins:[skin name]`
[...]

PR #12621, opened 2024-01-22
Skins: replace remaining 'skin:' occurences with 'skins:[skin name]'

ronso0:
@mxmilkiib discovered a flaw/regression I introduced with #12463, #12614.
The intent of #12463 was to simplify skin modding by requiring only a skin dir with only the modded sub-level template.
However, this fails if that template itself requires templates from the base skin.

This PR replaces all remaining 'skin:' occurences with 'skins:[skin name]'.

The procedure to mod a single template is now not that simple anymore, unfortnately.
But it still is rather inexpensive considering you'd benefit from future updates to the base skin. Use `<Template src="skin:path/to/template.xml">`, or explicitly `<Template src="skins:[name of modded skin]/path/to/template.xml`

**Example:** modifiy `Deere/deck_controls_row.xml`
The include tree looks like this:
```
+ Deere
∟ skin.xml
  ∟ left_/right_gutter.xml
    ∟ deck.xml
       ∟ deck_controls_row.xml
```
 * skin.xml: use `skin:left_gutter.xml` and  `skin:right_gutter.xml`
 * left_/right_gutter.xml: use `skin:deck.xml`
 * deck.xml: use `skin:path/to/deck_controls_row.xml`

Thanks @mxmilkiib for battle testing Mixxx beta and alpha!

### DIFF
--- a/res/skins/Deere/auxiliary.xml
+++ b/res/skins/Deere/auxiliary.xml
@@ -46,7 +46,7 @@
                     <SizePolicy>min,min</SizePolicy>
                     <Layout>vertical</Layout>
                     <Children>
-                      <Template src="skin:knob.xml">
+                      <Template src="skins:Deere/knob.xml">
                         <SetVariable name="TooltipId">microphone_pregain</SetVariable>
                         <SetVariable name="control">pregain</SetVariable>
                         <SetVariable name="color">green</SetVariable>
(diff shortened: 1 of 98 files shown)


ronso0:
I need to give this a second thought.
FWIW I'm tempted to revert #12463, since this here would somehow throw us back to square one in case you'd want to work on a clone skin (all template and style paths `skins:SkinName/..` would need to be renamed.

daschuer:
What is the state here?

daschuer:
@ronso0 is this safe to be shifted to 2.4.1?

ronso0:
Sure, no hurry. Skin mods is nothing that should hold us up.
Maybe we come up with another solution.

mxmilkiib:
N.b. Getting a crash currently with an incorrect template filename; #12729

PR #12632, opened 2024-01-23
Skins: add loop cue icons

### LINKED ISSUE #12619: 2.4 update Hotcue buttons with 'Loop cue' icon in Deere, Tango and Shade
### Bug Description

Currently, only LateNight shows the loop icon when the cue is a loop cue.

### Version

_No response_

### OS

_No response_

ronso0:
**LateNight classic**
Not ideal but there'a clear distinction between hotcues and loop cues.
image image

**Deere**
image

**Tango**
image
Since bg icons are scaled (down) it looks non-ideal in samplers and squeezed decks, but better than nothing
image  image
Re squeezed decks: I'll open a PR to increase Tango's minimum size. With most configurations it'll still fit into 1024px screens but at least we don't get complaints it looks bad when we still claim to support 1024px.

**Shade**
I ave no patience to fiddle with the PNGs, I just added a WIP GiMP template.
If someone wants to finish that:
* move hotcue icons from `<State>` nodes in xml to style sheets
  `WHotCueButton[displayValue="0"]` for empty state
  `WHotCueButton[displayValue="1"], WHotCueButton[displayValue="2"]` for loaded and 'active' state, add hotcue icons per hotcue number
* add `WHotCueButton[displayValue="1"][type="loop], WHotCueButton[displayValue="2"][type="loop]` selectors, add loop cue icons per hotcue number


daschuer (approved):
I will take a look at Shade.

PR #12633, opened 2024-01-23
Effects preferences: allow preset Export & Delete of multi-selection

ronso0:
... simply by enabling/disabling the respective buttons accordingly.
Export/delete slots can already handle multi-selection.


daschuer (approved):
LGTM, Thank you.

PR #12638, opened 2024-01-24
add WEffectMetaKnob, draws arc from default meta position

### LINKED ISSUE #12634: origins of effect Meta knob arcs are not consistent relative to their default positions
### Bug Description

Mixxx 2.4.0-beta
Git Version: 2.4-beta-347-gc323d0cd54
Platform: Linux x86_64
OS: Debian 11

Some knobs do not draw the delta ring relative to their default position.

Screenshot from 2024-01-23 20-06-51
Screenshot from 2024-01-23 20-07-18

"Echo", "Moog Filter" and "Balance" in the above shots do not draw the ring relative to the default (minimal/no-effect) position.

### Version

_No response_

### OS

_No response_

ronso0:
When an effect is loaded, the knob gets the default meta parameter from the effect manifest and draws the knob arc from there to current angle.

Closes #12634

Don't let this distract you from the 2.4 release, this can very well wait for 2.4.1 or later.

I added this widget only where arcs are or may be used, i.e. all skins except Shade.

### DIFF
--- a/src/effects/effectknobparameterslot.cpp
+++ b/src/effects/effectknobparameterslot.cpp
@@ -239,10 +239,3 @@ void EffectKnobParameterSlot::syncSofttakeover() {
 double EffectKnobParameterSlot::getValueParameter() const {
     return m_pControlValue->getParameter();
 }
-
-std::optional<double> EffectKnobParameterSlot::neutralPointOnScale() const {
-    if (m_pManifestParameter) {
-        return m_pManifestParameter->neutralPointOnScale();
-    }
-    return std::nullopt;
-}
--- a/src/effects/effectknobparameterslot.cpp
+++ b/src/effects/effectknobparameterslot.cpp
@@ -239,3 +239,10 @@ void EffectKnobParameterSlot::syncSofttakeover() {
 double EffectKnobParameterSlot::getValueParameter() const {
     return m_pControlValue->getParameter();
 }
+
+double EffectKnobParameterSlot::neutralPointOnScale() const {
+    if (m_pManifestParameter) {
+        return m_pManifestParameter->neutralPointOnScale();
+    }
+    return 0.0;
+}
--- a/src/effects/effectknobparameterslot.cpp
+++ b/src/effects/effectknobparameterslot.cpp
@@ -240,9 +240,9 @@ double EffectKnobParameterSlot::getValueParameter() const {
     return m_pControlValue->getParameter();
 }
 
-double EffectKnobParameterSlot::neutralPointOnScale() const {
+std::optional<double> EffectKnobParameterSlot::neutralPointOnScale() const {
     if (m_pManifestParameter) {
         return m_pManifestParameter->neutralPointOnScale();
     }
-    return 0.0;
+    return std::nullopt;
 }
--- a/src/effects/effectparameterslotbase.h
+++ b/src/effects/effectparameterslotbase.h
@@ -50,6 +50,10 @@ class EffectParameterSlotBase : public QObject {
 
     virtual void setParameter(double value) = 0;
 
+    virtual double neutralPointOnScale() const {
+        return 0.0;
+    }
+
   signals:
     // Signal that indicates that the EffectParameterSlotBase has been updated.
     void updated();
--- a/src/effects/effectparameterslotbase.h
+++ b/src/effects/effectparameterslotbase.h
@@ -51,8 +51,11 @@ class EffectParameterSlotBase : public QObject {
 
     virtual void setParameter(double value) = 0;
 
-    virtual std::optional<double> neutralPointOnScale() const {
-        return std::nullopt;
+    std::optional<double> neutralPointOnScale() const {
+        if (m_pManifestParameter == nullptr) {
+            return std::nullopt;
+        }
+        return m_pManifestParameter->neutralPointOnScale();
     }
 
   signals:
--- a/src/effects/effectparameterslotbase.h
+++ b/src/effects/effectparameterslotbase.h
@@ -2,6 +2,7 @@
 
 #include <QObject>
 #include <QString>
+#include <optional>
 
 #include "effects/backends/effectmanifestparameter.h"
 #include "util/class.h"
@@ -50,8 +51,8 @@ class EffectParameterSlotBase : public QObject {
 
     virtual void setParameter(double value) = 0;
 
-    virtual double neutralPointOnScale() const {
-        return 0.0;
+    virtual std::optional<double> neutralPointOnScale() const {
+        return std::nullopt;
     }
 
   signals:
--- a/src/widget/weffectmetaknob.cpp
+++ b/src/widget/weffectmetaknob.cpp
@@ -1,0 +1,46 @@
+#include "widget/weffectmetaknob.h"
+
+#include "effects/effectslot.h"
+#include "moc_weffectmetaknob.cpp"
+#include "widget/effectwidgetutils.h"
+
+WEffectMetaKnob::WEffectMetaKnob(QWidget* pParent, EffectsManager* pEffectsManager)
+        : WKnobComposed(pParent),
+          m_pEffectsManager(pEffectsManager) {
+}
+
+void WEffectMetaKnob::setup(const QDomNode& node, const SkinContext& context) {
+    WKnobComposed::setup(node, context);
+    auto pChainSlot = EffectWidgetUtils::getEffectChainFromNode(
+            node, context, m_pEffectsManager);
+    m_pEffectSlot =
+            EffectWidgetUtils::getEffectSlotFromNode(node, context, pChainSlot);
+    VERIFY_OR_DEBUG_ASSERT(m_pEffectSlot) {
+        SKIN_WARNING(node, context, QStringLiteral("Could not find effect slot"));
+        return;
+    }
+    connect(m_pEffectSlot.data(),
+            &EffectSlot::effectChanged,
+            this,
+            &WEffectMetaKnob::effectChanged);
+    effectChanged();
+}
+
+void WEffectMetaKnob::effectChanged() {
+    EffectManifestPointer pManifest = nullptr;
+    if (m_pEffectSlot->isLoaded()) {
+        pManifest = m_pEffectSlot->getManifest();
+    }
+
+    if (pManifest) {
+        setDefaultAngleFromParameterOrReset(pManifest->metaknobDefault());
+        setBaseTooltip(QString("%1\n%2").arg(
+                pManifest->name(),
+                pManifest->description()));
+    } else {
+        // This makes drawArc() fall back to drawing from center (if Reverse is not set)
+        setDefaultAngleFromParameterOrReset(-1);
+        setBaseTooltip("");
+    }
+    update();
+}
--- a/src/widget/weffectmetaknob.cpp
+++ b/src/widget/weffectmetaknob.cpp
@@ -34,7 +34,7 @@ void WEffectMetaKnob::effectChanged() {
 
     if (pManifest) {
         setDefaultAngleFromParameterOrReset(pManifest->metaknobDefault());
-        setBaseTooltip(QString("%1\n%2").arg(
+        setBaseTooltip(QStringLiteral("%1\n%2").arg(
                 pManifest->name(),
                 pManifest->description()));
     } else {
--- a/src/widget/weffectmetaknob.cpp
+++ b/src/widget/weffectmetaknob.cpp
@@ -38,8 +38,7 @@ void WEffectMetaKnob::effectChanged() {
                 pManifest->name(),
                 pManifest->description()));
     } else {
-        // This makes drawArc() fall back to drawing from center (if Reverse is not set)
-        setDefaultAngleFromParameterOrReset(-1);
+        setDefaultAngleFromParameterOrReset(std::nullopt);
         setBaseTooltip("");
     }
     update();
--- a/src/widget/weffectmetaknob.h
+++ b/src/widget/weffectmetaknob.h
@@ -1,0 +1,23 @@
+#pragma once
+
+#include "effects/defs.h"
+#include "widget/wknobcomposed.h"
+
+class EffectsManager;
+
+// This is used for effect parameter knobs with dynamic
+// tooltips, if the knob value is displayed by rotating a single SVG..
+class WEffectMetaKnob : public WKnobComposed {
+    Q_OBJECT
+  public:
+    WEffectMetaKnob(QWidget* pParent, EffectsManager* pEffectsManager);
+
+    void setup(const QDomNode& node, const SkinContext& context) override;
+
+  private slots:
+    void effectChanged();
+
+  private:
+    EffectsManager* m_pEffectsManager;
+    EffectSlotPointer m_pEffectSlot;
+};
--- a/src/widget/weffectparameterknobcomposed.cpp
+++ b/src/widget/weffectparameterknobcomposed.cpp
@@ -29,6 +29,7 @@ void WEffectParameterKnobComposed::parameterUpdated() {
         setBaseTooltip(QString("%1\n%2").arg(
                 m_pEffectParameterSlot->name(),
                 m_pEffectParameterSlot->description()));
+        setDefaultAngleFromParameterOrReset(m_pEffectParameterSlot->neutralPointOnScale());
     } else {
         // The knob should be hidden by the skin when the parameterX_loaded ControlObject
         // indicates no parameter is loaded, so this tooltip should never be shown.
--- a/src/widget/weffectparameterknobcomposed.cpp
+++ b/src/widget/weffectparameterknobcomposed.cpp
@@ -26,7 +26,7 @@ void WEffectParameterKnobComposed::setup(const QDomNode& node, const SkinContext
 
 void WEffectParameterKnobComposed::parameterUpdated() {
     if (m_pEffectParameterSlot->isLoaded()) {
-        setBaseTooltip(QString("%1\n%2").arg(
+        setBaseTooltip(QStringLiteral("%1\n%2").arg(
                 m_pEffectParameterSlot->name(),
                 m_pEffectParameterSlot->description()));
         setDefaultAngleFromParameterOrReset(m_pEffectParameterSlot->neutralPointOnScale());
--- a/src/widget/wknobcomposed.cpp
+++ b/src/widget/wknobcomposed.cpp
@@ -124,6 +124,18 @@ void WKnobComposed::setDefaultAngleFromParameterOrReset(std::optional<double> pa
     }
 }
 
+void WKnobComposed::resizeEvent(QResizeEvent* re) {
+    Q_UNUSED(re);
+    // In order to always draw the arcs undistorted we set up
+    // a quadratic target rectangle regardless of the widget's
+    // width-to-height ratio.
+    qreal centerX = width() / 2.0 + m_dKnobCenterXOffset;
+    qreal centerY = height() / 2.0 + m_dKnobCenterYOffset;
+    QPointF topLeft = QPointF((centerX - m_dArcRadius), (centerY - m_dArcRadius));
+    QPointF bottomRight = QPointF((centerX + m_dArcRadius), (centerY + m_dArcRadius));
+    m_rect = QRectF(topLeft, bottomRight);
+}
+
 void WKnobComposed::paintEvent(QPaintEvent* e) {
     Q_UNUSED(e);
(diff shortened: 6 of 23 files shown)


daschuer:
This one looks now: 
grafik
The Meta knob has the correct ring, but the parameter knob is still wrong.

ronso0:
No prob, ~~fix~~ second new feature is here.

This revealed some misplaced 'neutral point on scale' values, e.g. high freq. of Graphic EQ.

ronso0:
Draft so I can clean up and force-push  without disrupting a review.

src/widget/wknobcomposed.cpp
```diff
@@ -116,6 +114,15 @@ void WKnobComposed::onConnectedControlChanged(double dParameter, double dValue)
     }
 }
 
+void WKnobComposed::setDefaultAngleFromParameterOrReset(double parameter) {
+    if (parameter < 0 || parameter > 1) {
+        m_defaultAngleIsValid = false;
+    } else {
+        m_defaultAngle = m_dMinAngle + (m_dMaxAngle - m_dMinAngle) * parameter;
```
Swiftb0y, line 121:
```suggestion
        m_defaultAngle = std::lerp(m_dMinAngle, m_dMaxAngle, parameter);
```

ronso0: nice one, thanks!

src/effects/effectknobparameterslot.cpp
```diff
@@ -239,3 +239,10 @@ void EffectKnobParameterSlot::syncSofttakeover() {
 double EffectKnobParameterSlot::getValueParameter() const {
     return m_pControlValue->getParameter();
 }
+
+double EffectKnobParameterSlot::neutralPointOnScale() const {
+    if (m_pManifestParameter) {
+        return m_pManifestParameter->neutralPointOnScale();
+    }
+    return 0.0;
```
Swiftb0y, line 247: this looks like a std::optional candidate. wdyt?

src/widget/wknobcomposed.h
```diff
@@ -28,6 +28,10 @@ class WKnobComposed : public WWidget {
     void mouseReleaseEvent(QMouseEvent *e) override;
     void paintEvent(QPaintEvent* /*unused*/) override;
 
+    bool m_defaultAngleIsValid;
+    double m_defaultAngle;
```
Swiftb0y, line 32: if you wrap this in an optional, its impossible to accidentally use it without checking.

Swiftb0y (commented):
just brainstorming here. not trying to drag this out.

src/effects/effectknobparameterslot.cpp
daschuer: In this case 0.0 is a good default and valid value for the unlike case that the is no parameter connected.

daschuer (approved):
Works perfect. Thank you.

src/effects/effectknobparameterslot.cpp
Swiftb0y: but 0.0 could also be a valid neutral point (for example in a high-pass filter), it's impossible for the caller to tell. I think the caller needs to explicitly handle the circumstance (for example by showing "error" as a value, not just 0).

daschuer: An error value for the neutral position, when we have also no value and no scale? 
This case should never happen, since the parameter knobs are hidden in that case. So coding around that is not reasonable. 

We have that state for the meta knob when no effect is loaded. Currently a 0 .. 1 knob with default is assumed and no error state. Since the value of the knob remains where is was when loading "---".  What kind of error state you have in mind?

Swiftb0y: > An error value for the neutral position, when we have also no value and no scale?

> What kind of error state you have in mind?

The error state is if `m_pManifestParameter == nullptr`. 

> We have that state for the meta knob when no effect is loaded. Currently a 0 .. 1 knob with default 0 is assumed and no special error state. The value of the knob remains where is was when loading "---".

That doesn't seem like particularly good design IMO because that assumes some other part of the code does the appropriate checking. This assumption could get violated easily in a future refactoring. If you just look at this public member function (thus part of the class's API), there is no way to verify whether the value is returned represents an error or real value. And even if there is, the caller would have to know about that implementation detail. The result is an API that is easy to misuse.

src/widget/wknobcomposed.h
ronso0: so `std::optional<double> m_defaultAngle;` would not require ..isValid anymore, we'd just check
`if (m_defaultAngle) { .. } else ..`

ronso0: No, I'm not getting it.
(your comment in _this_ place)

I'll try using std::optional<double> for neutralPointOnScale()

Swiftb0y: yes, though its often nicer to use `value_or` and the monadic operations added in C++23. Id also prefer `has_value` over `operator bool` in if conditions. [`std::optional<T>::operator->, std::optional<T>::operator*`](https://en.cppreference.com/w/cpp/utility/optional/operator*) should be avoided!

ronso0: ah okay, I was confused why I could use m_angle right away, std::optional<T> has `value()`

src/widget/weffectmetaknob.h
```diff
@@ -0,0 +1,23 @@
...
+    WEffectMetaKnob(QWidget* pParent, EffectsManager* pEffectsManager);
+
+    void setup(const QDomNode& node, const SkinContext& context) override;
+
+  private slots:
+    void effectChanged();
+
+  private:
+    EffectsManager* m_pEffectsManager;
+    EffectSlotPointer m_pEffectSlot;
```
Swiftb0y, line 22: encode constraints and ownership in the type system (`gls::not_null<EffectsManager*>` would blow the scope too much right now). implicit info should at least receive a comment.
```suggestion
    gls::not_null<std::weak_ptr<EffectsManager>> m_pEffectsManager;
    // m_pEffectSlot can be null, in which case the effect does not have an effect assigned.
    EffectSlotPointer m_pEffectSlot;
```

src/widget/weffectmetaknob.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+void WEffectMetaKnob::effectChanged() {
+    EffectManifestPointer pManifest = nullptr;
+    if (m_pEffectSlot->isLoaded()) {
+        pManifest = m_pEffectSlot->getManifest();
+    }
+
+    if (pManifest) {
+        setDefaultAngleFromParameterOrReset(pManifest->metaknobDefault());
+        setBaseTooltip(QString("%1\n%2").arg(
```
Swiftb0y, line 37:
```suggestion
        setBaseTooltip(QStringLiteral("%1\n%2").arg(
```

src/widget/weffectmetaknob.cpp
```diff
@@ -0,0 +1,46 @@
...
+                pManifest->name(),
+                pManifest->description()));
+    } else {
+        // This makes drawArc() fall back to drawing from center (if Reverse is not set)
+        setDefaultAngleFromParameterOrReset(-1);
```
Swiftb0y, line 42: seems better now that we take an optional.
```suggestion
        setDefaultAngleFromParameterOrReset(std::nullopt);
```

src/effects/effectknobparameterslot.cpp
Swiftb0y: > Are you satisfied with 1. ?

Yes. 

>  In this particular case it does not makes sense to consider a case that never happens.
We can probably ASSERT that, I am just afraid that will bite us later.

playing devils advocate here, this sounds nonsensical. If it never happens, then assert and encode that in the type-system (plain double, no special values indicating an error state). If it happens, lets encode it in the type system (`std::optional<double>`).

src/widget/wknobcomposed.cpp
```diff
@@ -116,6 +114,14 @@ void WKnobComposed::onConnectedControlChanged(double dParameter, double dValue)
     }
 }
 
+void WKnobComposed::setDefaultAngleFromParameterOrReset(std::optional<double> parameter) {
+    if (!parameter.has_value() || parameter.value() < 0 || parameter.value() > 1) {
+        m_defaultAngle = std::nullopt;
+    } else {
+        m_defaultAngle = std::lerp(m_dMinAngle, m_dMaxAngle, parameter.value());
+    }
```
Swiftb0y, line 122: not a suggestion, just a preview of what would be possible in C++23. (imo easier to read and also faster because it only does a single has_value check. 
```suggestion
    m_defaultAngle = parameter.transform([this](double value) {
        return (parameter < 0 || parameter > 1) ? 
            std::nullopt : 
            std::lerp(m_dMinAngle, m_dMaxAngle, parameter);
    });
```

Swiftb0y (commented):
looks like ur currently missing a couple `#include <optional>`s.

src/widget/weffectmetaknob.cpp
ronso0: :+1: thanks, I overlooked that

src/widget/weffectmetaknob.h
ronso0: Ahmm, and change it everywhere in src/effects ??

There is `VERIFY_OR_DEBUG_ASSERT(m_pEffectSlot)` in
`WEffectMetaKnob::setup` and
`WEffectParameterKnobComposed::setup`
and in case of null the respective slots where `m_pEffectSlot`/`m_pParameterSlot` are accessed are not connected.

src/widget/wknobcomposed.h
Swiftb0y: yeah. the difference is that one is checked the other is not. The problem with that API is that the check could now be very far away from the use and could get lost in a future refactoring. `value()` is safer (in the sense that it `std::terminates` when there is no value, instead of UB) but obviously checks each and every time. optional monadic operations `C++23` solve that.

src/widget/weffectparameterknobcomposed.cpp
```diff
@@ -25,13 +25,16 @@ void WEffectParameterKnobComposed::setup(const QDomNode& node, const SkinContext
 }
 
 void WEffectParameterKnobComposed::parameterUpdated() {
+    std::optional<double> neutralPoint = std::nullopt;
     if (m_pEffectParameterSlot->isLoaded()) {
-        setBaseTooltip(QString("%1\n%2").arg(
+        setBaseTooltip(QStringLiteral("%1\n%2").arg(
                 m_pEffectParameterSlot->name(),
                 m_pEffectParameterSlot->description()));
+        neutralPoint = m_pEffectParameterSlot->neutralPointOnScale();
```
daschuer, line 33: If we have here a std:nullopt we have a skin error. Likely the a button slot is connected to a knob. 
Should we issue a warning here?

src/widget/weffectmetaknob.cpp
```diff
@@ -0,0 +1,45 @@
...
+        pManifest = m_pEffectSlot->getManifest();
+    }
+
+    if (pManifest) {
+        setDefaultAngleFromParameterOrReset(pManifest->metaknobDefault());
+        setBaseTooltip(QStringLiteral("%1\n%2").arg(
+                pManifest->name(),
+                pManifest->description()));
+    } else {
+        setDefaultAngleFromParameterOrReset(std::nullopt);
```
daschuer, line 41: This means that the "ArcUnipolar" and "ArcReversed" take control which are not set from the skin. The old position remains. This is not consistent. I would prefer to also keep the old Arc. 
We may introduce a "not in use state" in a follow up PR as suggested form @Swiftb0y

src/widget/weffectmetaknob.h
Swiftb0y: > Ahmm, and change it everywhere in src/effects

In the long term, yes. Not in this PR.

> There is VERIFY_OR_DEBUG_ASSERT(m_pEffectSlot)

Yes, but only on setup. The raw pointer could become dangling, thus a `std::weak_ptr` to ensure that can't happen (and to prevent ref cycles, which would likely form if it was replaced by `std::shared_ptr`. 

Don't get me wrong, the current solution works, no doubts, it's just not very robust IMO.

Swiftb0y (approved):
LGTM. feel free to force-push the squashed fixups. Thanks for your patience.

src/widget/wknobcomposed.cpp
```diff
@@ -116,6 +114,14 @@ void WKnobComposed::onConnectedControlChanged(double dParameter, double dValue)
     }
 }
 
+void WKnobComposed::setDefaultAngleFromParameterOrReset(std::optional<double> parameter) {
+    if (!parameter.has_value() || parameter.value() < 0 || parameter.value() > 1) {
+        m_defaultAngle = std::nullopt;
```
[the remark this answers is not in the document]
Swiftb0y: whats the advantage here?

src/widget/weffectparameterknobcomposed.cpp
```diff
@@ -25,13 +25,16 @@ void WEffectParameterKnobComposed::setup(const QDomNode& node, const SkinContext
...
         setBaseTooltip(QString("%1\n%2").arg(
                 m_pEffectParameterSlot->name(),
                 m_pEffectParameterSlot->description()));
+        neutralPoint = m_pEffectParameterSlot->neutralPointOnScale();
     } else {
         // The knob should be hidden by the skin when the parameterX_loaded ControlObject
         // indicates no parameter is loaded, so this tooltip should never be shown.
         setBaseTooltip("");
     }
+    setDefaultAngleFromParameterOrReset(neutralPoint);
```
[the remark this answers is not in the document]
ronso0: The knob doesn't know if it's visible (sure, in official skins it's hidden if there's no parameter loaded)

src/widget/wknobcomposed.cpp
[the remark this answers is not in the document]
ronso0: Yes, we could have just one painter call, though at the cost of more temp vars.

And there's some more potential for optimization, sure.
For exampe the rect doesn't need to be calculated each time drawArc() is called, it'd be sufficient to do this in a resize event (which is called on initial show() and when being resized of course, if someone prefers to use variable sizes)

src/widget/weffectmetaknob.cpp
ronso0: If a skin designer unsets `<ArcUnipolar>` and/or sets `ArcReversed` for a Meta knob or effect parameter knob, well, then there is _some_ intention behind it :thinking: :man_shrugging:

src/widget/weffectparameterknobcomposed.cpp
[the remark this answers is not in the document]
daschuer: In the theoretical case it is not hidden, the same ideas like the meta knob applies. 
It does not make sense to change the scale to center. Just keeping it unchanged seems to be consistent. 

Or remove the arc entirely or such in a follow up PR.

src/widget/weffectparameterknobcomposed.cpp
daschuer: Ok, I have not tested that case. That is sufficient than.

src/widget/weffectmetaknob.cpp
daschuer: Since this PR, the default position is taking over. I consider adopting the old values for the "---" selections as a minor inconsistency.

daschuer:
By the way: Is this a last minute 2.4 PR? 
I don't mind. At least my latest comments are not intended blocking a merge.

ronso0:
Yes, that can slip into 2.4.1 last minute!
(need to double-check all skins for regressions! though I need to quit now, can take care of it on sunday earliest)

Please check the rect commit and the fixup with your suggestion.

daschuer:
> Please check the rect commit and the fixup with your suggestion.

Looks good. Thank you.

ronso0:
> Error: ../src/widget/wknobcomposed.cpp:118:45: error: 'value' is unavailable: introduced in macOS 10.13
    if (!parameter.has_value() || parameter.value() < 0 || parameter.value() > 1) {
                                            ^
/Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/c++/v1/optional:954:27: note: 'value' has been explicitly marked unavailable here
    constexpr value_type& value() &
                          ^

:disappointed: 
Any ideas how to work around that?

Otherwise we may as well move this to 2.5

src/widget/wknobcomposed.cpp
```diff
@@ -116,6 +114,26 @@ void WKnobComposed::onConnectedControlChanged(double dParameter, double dValue)
...
+    } else {
+        m_defaultAngle = std::lerp(m_dMinAngle, m_dMaxAngle, parameter.value());
```
daschuer, line 121: you may try 
```suggestion
    if (!parameter.has_value() || *parameter < 0 || *parameter > 1) {
        m_defaultAngle = std::nullopt;
    } else {
        m_defaultAngle = std::lerp(m_dMinAngle, m_dMaxAngle, *parameter);
```

Swiftb0y:
>  Any ideas how to work around that?

No idea. Just AppleClang being completely outdated as usual... 

> Otherwise we may as well move this to 2.5

Yeah. maybe leave a TODO there so the next time someone stumbles over this, they can do the trivial refactoring.

src/widget/wknobcomposed.cpp
ronso0: okay, great, that did the trick!
and I added a TODO as @Swiftb0y suggested.

ronso0:
Alright, CI all green.

Please take a look, then I'll squash the fixups (had to fixup since the commit introducing std::optional couldn't be built).

src/widget/wknobcomposed.cpp
Swiftb0y: > In general I would prefer to give m_defaultAngle a valid value here. It is better to have a variable that can be used unconditional in all cases.

That defeats the purpose of the optional. This snippet of code here can't accurately determine what value is appropriate.

src/widget/wknobcomposed.cpp
```diff
@@ -115,6 +115,8 @@ void WKnobComposed::onConnectedControlChanged(double dParameter, double dValue)
 }
 
 void WKnobComposed::setDefaultAngleFromParameterOrReset(std::optional<double> parameter) {
+    // TODO(xxx) Use m_defaultAngle->value() as soon as
+    // AppleClang 10.13+ is used.
```
Swiftb0y, line 119: ideally this just used the monadic operations from C++23 by then :wink:

ronso0:
Am I done here, all reviews addressed?

Swiftb0y (commented):
from my side, yes. almost

src/effects/effectparameterslotbase.h
```diff
@@ -50,6 +51,10 @@ class EffectParameterSlotBase : public QObject {
 
     virtual void setParameter(double value) = 0;
 
+    virtual std::optional<double> neutralPointOnScale() const {
+        return std::nullopt;
+    }
```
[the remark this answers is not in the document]
ronso0: Then we'd have to override it for `EffectButtonParameterSlot`, just for the _possible_ case that the neutral point is requested for a button parameter?
That would otherwise return the unchanged default 0. Though, in that case (controlling a button parameter with a knob widget) a lot else wouldn't work quite as desired.
I don't mind..

Wdyt?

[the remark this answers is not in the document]
Swiftb0y: > Then we'd have to override it for EffectButtonParameterSlot, just for the possible case that the neutral point is requested for a button parameter?

Well if that makes sense for a button parameter, we should make sure the value also makes sense. If it doesn't make sense (which is what you are suggesting, right?) then we can also let the return value be nonsensical. 

The advantage of implementing it in the base class is that we can avoid virtual dispatch. 

I think the primary issue here is that `neutralPointOnScale` only makes sense on knobs, but the existing code infrastructure requires it to be implemented on the base-class. Is that correct?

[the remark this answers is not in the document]
ronso0: Alright, implemented as you suggested. In the unlikely button case it'll return the default 0.

Swiftb0y (approved):
LGTM. sorry for having forgotten about this.

Swiftb0y:
Thank you

ronso0:
> LGTM. sorry for having forgotten about this.

Never mind, there was an unaddressed review and I've pushed this back until now.
Thanks for your review!

PR #1746, opened 2018-07-09
Keyboard controller

daschuer:
Successor of https://github.com/mixxxdj/mixxx/pull/966
Please see the original PR for comments.

PR #12639, opened 2024-01-24
Keyboard mapping: Return triggers doubleclick, move Preview functions to P / Shift+P

ronso0:
**Old**
`Return` key stops a playing preview deck.
`Alt`+`Return` loads a track and starts playing it

**New**
`P` loads a track and starts playing it
`Shift`+`P` stops it
`Return` triggers the configured doubleclick action in the tracks table
(`Ctrl`+`Return` opens the track properties editor, for one track only, for multi-selection see #12548)

`Return` for activating a selection feels natural, and I caught myself multiple times trying that.
Considering that, we may as well
* keep `Alt`+`Return` for load to preview deck
* `P` does start/pause
* `Shift`+`P` does go to start (or Cue?), and stop?

Keyboard mappings and the keyboard sheet are updated accordingly.
image
* is supposed to mark hard-coded hotkeys, need to check how to add that hint elegantly.

This is kind of a breaking change, it shouldn't go to a bugfix release.
Otoh I don't want to delay 2.4

I'll probably split this up into the bugfixes (2.4) and the feature part (2.5).
Anyway, take a look and let me know what you think.


daschuer (approved):
That looks good to me. I support the idea to merge it to the 2.4 branch.

ronso0:
I removed the intermediate PDF changes (binary file).
I'll create and export the PDF when this is ready to be merged, and update the manual.

daschuer:
What is missing to be merged? Do we want a quick merge to 2.4 this weekend or should it go to main?

ronso0:
Someone needs to review the greek and russian mappings.

And I need to update the PDF.

daschuer:
Some findings: 
* Greek: has `OptionsMenu_ReloadSkin Ctrl+Shift+ρ` which is not present in de_DE
* Greek: `cue_default ΄` is a dead key
* Greek: `cue_gotoandstop Shift+ ¨` is also a dead key

There are some other wrong keys assigned. I will issue a PR.

ronso0:
Oh, thanks for taking a close look!
I was just about to load those keyboard layouts and double-check the mappings. Not sure how meaningful such a test is.

daschuer:
I just did the same. I am currently facing the issue that Mixxx comes always up with the English layout.
It looks like  `QGuiApplication::inputMethod();` is broken. The same happens with Mixxx 2.3.6 using the same method.

daschuer:
We can't do much about the dead keys. One main issue is that Mixxx does not follows if the users switching the keyboard layout. This is what Greek users do when searching for Greek tracks. We  have code rotting around that issue all these issues.

daschuer:
https://github.com/mixxxdj/mixxx/pull/1746

ronso0:
I pushed once more because the Github diff of the SVG (rendering) looked wrong, which I blamed Inkscape for.
Anyhow, the SVG and PDF look as intended.

daschuer:
Cool, thank you for this last minute PR. 
LGTM.

ronso0:
Yeah, great this still slipped in!

Reminder for myself: update the manual

PR #12646, opened 2024-01-26
MacOsVersion: Import Foundation instead of Cocoa

fwcd:
Cocoa/AppKit is unavailable on iOS and since we only use Foundation APIs here anyway, we can just import it directly.


daschuer (approved):
Still green, Thank you. LGTM

PR #12649, opened 2024-01-27
LegacyControllerMappingFileHandler: Add missing cond-compile for building without HID

fwcd:
This is something I stumbled upon while building Mixxx without HID support. Everything in `controllers/hid` is only available when `-DHID=ON`, including `legacyhidcontrollermappingfilehandler.h`:

https://github.com/mixxxdj/mixxx/blob/59ca3a25b768c44ac0934ed4cb4504cac52bde54/CMakeLists.txt#L3354-L3408

LegacyControllerMappingFileHandler was missing a cond-compile around the HID handler, therefore this patch.


JoergAtGithub:
LGTM! Thank you!

PR #12650, opened 2024-01-27
ScreenSaver: Check `Q_OS_MACOS` instead of `Q_OS_MAC`

fwcd:
`Q_OS_MAC` is a deprecated synonym for `Q_OS_DARWIN`, referring to both macOS and iOS. Since screensavers aren't available on iOS, this replaces the cond-compile with `Q_OS_MACOS`.


JoergAtGithub:
LGTM! Thank you!

PR #12656, opened 2024-01-27
macOS packaging: Capitalize bundle and executable name when `MACOS_BUNDLE` is set (Mixxx.app)

fwcd:
The dominant convention on macOS is that both bundle and executable names are capitalized.

Currently our bundles are of the form `mixxx.app/Contents/MacOS/mixxx`, though Finder will display the name capitalized since we already set a capitalized display name. With this patch, however, the true name of the bundle will also be `Mixxx.app/Contents/MacOS/Mixxx`.

Additionally, this patch sets the bundle identifier in the CMakeLists, for consistency with the way we set the bundle name.


daschuer (approved):
Looks Good. Thank you for noticing this detail.

PR #12657, opened 2024-01-27
Samplers: dont create extra empty players during startup

ronso0:
Fixes #9115
If `samplers.xml` has more empty(!) elements than sampler players have been created by request of either skins or controller mappings, `SamplerBank` would trigger creation of all these unneeded samplers.
This happened after having shutdown with a 64-samplers skin, and starting with a skin with less samplers, even though none of the additional samplers has a track loaded.

Additionally, I  improved the track menu and the control picker menu (MIDI Learning Wizard):
in the "Load To" submenu samplers are now grouped into submenus with 16 samplers each.
image

### DIFF
--- a/src/controllers/controlpickermenu.cpp
+++ b/src/controllers/controlpickermenu.cpp
@@ -48,79 +48,79 @@ ControlPickerMenu::ControlPickerMenu(QWidget* pParent)
     m_otherGroupsTrMap.insert("Master", "Master");
 
     // Mixer Controls
-    QMenu* mixerMenu = addSubmenu(tr("Mixer"));
+    QMenu* pMixerMenu = addSubmenu(tr("Mixer"));
     // Crossfader / Orientation
-    QMenu* crossfaderMenu = addSubmenu(tr("Crossfader / Orientation"), mixerMenu);
+    QMenu* pCrossfaderMenu = addSubmenu(tr("Crossfader / Orientation"), pMixerMenu);
     addControl("[Master]",
             "crossfader",
             tr("Crossfader"),
             tr("Crossfader"),
-            crossfaderMenu,
+            pCrossfaderMenu,
             true);
     addDeckAndSamplerControl("orientation",
             tr("Orientation"),
             tr("Mix orientation (e.g. left, right, center)"),
-            crossfaderMenu);
+            pCrossfaderMenu);
     addDeckAndSamplerControl("orientation_left",
             tr("Orient Left"),
             tr("Set mix orientation to left"),
-            crossfaderMenu);
+            pCrossfaderMenu);
     addDeckAndSamplerControl("orientation_center",
             tr("Orient Center"),
             tr("Set mix orientation to center"),
-            crossfaderMenu);
+            pCrossfaderMenu);
     addDeckAndSamplerControl("orientation_right",
             tr("Orient Right"),
             tr("Set mix orientation to right"),
-            crossfaderMenu);
+            pCrossfaderMenu);
     // Main Output
-    QMenu* mainOutputMenu = addSubmenu(tr("Main Output"), mixerMenu);
+    QMenu* pMainOutputMenu = addSubmenu(tr("Main Output"), pMixerMenu);
     addControl("[Master]",
             "gain",
             tr("Main Output Gain"),
             tr("Main Output gain"),
-            mainOutputMenu,
+            pMainOutputMenu,
             true);
     addControl("[Master]",
             "balance",
             tr("Main Output Balance"),
             tr("Main Output balance"),
-            mainOutputMenu,
+            pMainOutputMenu,
             true);
     addControl("[Master]",
             "delay",
             tr("Main Output Delay"),
             tr("Main Output delay"),
-            mainOutputMenu,
+            pMainOutputMenu,
             true);
     // Headphone
-    QMenu* headphoneMenu = addSubmenu(tr("Headphone"), mixerMenu);
+    QMenu* pHeadphoneMenu = addSubmenu(tr("Headphone"), pMixerMenu);
     addControl("[Master]",
             "headGain",
             tr("Headphone Gain"),
             tr("Headphone gain"),
-            headphoneMenu,
+            pHeadphoneMenu,
             true);
     addControl("[Master]",
             "headMix",
             tr("Headphone Mix"),
             tr("Headphone mix (pre/main)"),
-            headphoneMenu,
+            pHeadphoneMenu,
             true);
     addControl("[Master]",
             "headSplit",
             tr("Headphone Split Cue"),
             tr("Toggle headphone split cueing"),
-            headphoneMenu);
+            pHeadphoneMenu);
     addControl("[Master]",
             "headDelay",
             tr("Headphone Delay"),
             tr("Headphone delay"),
-            headphoneMenu,
+            pHeadphoneMenu,
             true);
-    mixerMenu->addSeparator();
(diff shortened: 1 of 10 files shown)


src/mixer/samplerbank.cpp
```diff
@@ -195,33 +195,48 @@ bool SamplerBank::loadSamplerBankFromPath(const QString& samplerBankPath) {
...
     }
 
-    QDomNode n = root.firstChild();
-
-    while (!n.isNull()) {
+    const auto samplerNodes = root.childNodes();
+    if (samplerNodes.isEmpty()) {
+        return true;
+    }
+    for (int i = 0; i < samplerNodes.size(); i++) {
```
ronso0, line 202: I refactored this to avoid `n = n.nextSibling();` before each `continue`.
Unfortunately there is no iterator for `QDomNodeList`.

src/mixer/samplerbank.cpp
```diff
@@ -195,33 +195,48 @@ bool SamplerBank::loadSamplerBankFromPath(const QString& samplerBankPath) {
...
+        if (initializing && location.isEmpty()) {
+            continue;
+        }
+
+        // Later on, when the user loads a samplers file manually, we
+        // want to eject loaded tracks if the file's sample slot is empty.
+        // We also create new players even if they are not present in the
+        // GUI to not drop tracks loaded to invisible samplers when saving
+        // the samplers file.
+        if (m_pPlayerManager->numSamplers() < (unsigned)samplerNum) {
```
daschuer, line 231: This way around should be more safe since samplerNum might be read from a corrupt file. 
```suggestion
        if (static_cast<int>(m_pPlayerManager->numSamplers()) < samplerNum) {
```

daschuer: Here is by the way also suspicious casting: 
https://github.com/mixxxdj/mixxx/blob/a7af5725c72804622fc44b270b634ca5e5b10b7b/src/mixer/playermanager.cpp#L278
Maybe we should just use int everywhere?

src/mixer/samplerbank.h
```diff
@@ -19,7 +19,7 @@ class SamplerBank : public QObject {
             PlayerManager* pPlayerManager);
 
     bool saveSamplerBankToPath(const QString& samplerBankPath);
-    bool loadSamplerBankFromPath(const QString& samplerBankPath);
+    bool loadSamplerBankFromPath(const QString& samplerBankPath, bool initializing = false);
```
daschuer, line 22: How about: 
```suggestion
    bool loadSamplerBankFromPath(const QString& samplerBankPath, bool SkipCreateingHiddenSamplers = false);
```
To document what the function does not when it should be called.

src/mixer/playermanager.cpp
```diff
@@ -450,7 +450,7 @@ void PlayerManager::addDeckInner() {
 }
 
 void PlayerManager::loadSamplers() {
-    m_pSamplerBank->loadSamplerBankFromPath(getDefaultSamplerPath(m_pConfig));
+    m_pSamplerBank->loadSamplerBankFromPath(getDefaultSamplerPath(m_pConfig), true);
```
daschuer, line 453: You may work around the anonymous bool
```suggestion
    bool initializing = true; 
    m_pSamplerBank->loadSamplerBankFromPath(getDefaultSamplerPath(m_pConfig), initializing);
```

daschuer (changes requested):

JoergAtGithub:
Works for me on Windows11!

src/mixer/samplerbank.cpp
```diff
@@ -195,33 +195,48 @@ bool SamplerBank::loadSamplerBankFromPath(const QString& samplerBankPath) {
...
+        if (e.isNull() || e.tagName() != "sampler") {
+            continue;
+        }
+
+        const QString group = e.attribute("group", "");
+        const QString location = e.attribute("location", "");
+        int samplerNum;
+
+        if (group.isEmpty() || m_pPlayerManager->isSamplerGroup(group, &samplerNum)) {
+            continue;
```
ronso0, line 214: uff
```suggestion
        if (group.isEmpty() || !m_pPlayerManager->isSamplerGroup(group, &samplerNum)) {
            continue;
```

src/mixer/samplerbank.cpp
ronso0: You mean instead of `unsigned int`?

daschuer: Yes.

ronso0: Okay, I'll take a look and open another PR.

src/controllers/controlpickermenu.cpp
```diff
@@ -1479,14 +1479,23 @@ void ControlPickerMenu::addPlayerControl(const QString& control,
     }
 
     if (samplerControls) {
-        QMenu* samplerControlMenu = new QMenu(tr("Samplers"), controlMenu);
-        controlMenu->addMenu(samplerControlMenu);
+        QMenu* samplerControlMainMenu = addSubmenu(tr("Samplers"), controlMenu);
```
daschuer, line 1482: A p prefix would be nice.

src/controllers/controlpickermenu.cpp
```diff
@@ -1479,14 +1479,23 @@ void ControlPickerMenu::addPlayerControl(const QString& control,
...
+        QMenu* samplerResetControlMainMenu = nullptr;
         QMenu* samplerResetControlMenu = nullptr;
         if (resetControlMenu) {
-            samplerResetControlMenu = new QMenu(tr("Samplers"), resetControlMenu);
-            resetControlMenu->addMenu(samplerResetControlMenu);
+            samplerResetControlMainMenu = addSubmenu(tr("Samplers"), resetControlMenu);
+            samplerResetControlMenu = samplerResetControlMainMenu;
         }
+        const int maxSamplersPerMenu = 16;
+        int counter = 1;
```
daschuer, line 1491: Maybe 
```suggestion
        int samplerInMenu = 0;
```
I would also initialize it with zero, because else you exceed  maxSamplersPerMenu.

src/widget/wtrackmenu.cpp
```diff
@@ -818,19 +821,35 @@ void WTrackMenu::updateMenus() {
             }
         }
 
-        int iNumSamplers = static_cast<int>(m_pNumSamplers->get());
+        int iNumSamplers = static_cast<int>(m_pNumSamplers.get());
+        const int maxSamplersPerMenu = 16;
         if (iNumSamplers > 0) {
             m_pSamplerMenu->clear();
+            QMenu* pMenu = m_pSamplerMenu;
+            int counter = 1;
```
daschuer, line 829: The same here. `int samplerInMenu = 0;`

daschuer (changes requested):
Works already good. Hers some comments

src/controllers/controlpickermenu.cpp
ronso0: Done. For all menus, the diff is big but it's done.

src/controllers/controlpickermenu.cpp
ronso0: Done.
I applied the same renaming and the fix from your other suggestion also to WTrackMenu. Force-pushed but the changes were small to not cause review trouble I hope.

ronso0:
Now the menu is also updated if samplers or decks are added (skin change for example).

daschuer (approved):
LGTM, Thank you.

PR #12661, opened 2024-01-27
Engine: Add feature for building with/without Rubberband

fwcd:
This makes it possible to build Mixxx without Rubberband, using SoundTouch as the default keylock engine instead. This has proven pretty useful in building Mixxx for iOS, both for technical reasons (cross-compiling the library seems to be non-trivial) and potential licensing reasons (non-App-Store-clause-GPL).


daschuer (approved):
LGTM, Thank you.

PR #12663, opened 2024-01-27
macOS: Add more `Q_OS_MACOS` cond-compiles where appropriate

fwcd:
In preparation of iOS support, this turns a bunch of `__APPLE__` cond-compiles into `Q_OS_MACOS`, since e.g. dark title bars or the migration from `~/Library/Application Support/...` to `~/Library/Containers/...` only affects macOS.

### DIFF
--- a/src/coreservices.cpp
+++ b/src/coreservices.cpp
@@ -3,6 +3,7 @@
 #include <QApplication>
 #include <QFileDialog>
 #include <QStandardPaths>
+#include <QtGlobal>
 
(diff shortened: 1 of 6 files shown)


daschuer:
Q_OS_MACOS does only work if a Qt header has been included before. Is this always the case?

fwcd:
Yeah, this is why I added the `#include <QtGlobal>`, just to be on the safe side

daschuer (approved):
LGTM, Thank you.

PR #12664, opened 2024-01-27
GLWaveformRenderer: Move OpenGL 2.1 include into cond-compile

fwcd:
A minor fix that surfaced while compiling for iOS.


daschuer (approved):
LGTM, thank you.

PR #12665, opened 2024-01-28
VersionStore: Make iOS a recognized platform

fwcd:
A small "risk-free" patch towards iOS support

### DIFF
--- a/src/util/versionstore.cpp
+++ b/src/util/versionstore.cpp
@@ -72,8 +72,12 @@ QString VersionStore::applicationName() {
 
 // static
 QString VersionStore::platform() {
-#ifdef __APPLE__
+#ifdef Q_OS_IOS
+    QString base = QStringLiteral("iOS");
+#elif defined(Q_OS_MACOS)
     QString base = QStringLiteral("macOS");
+#elif defined(__APPLE__)
+    QString base = QStringLiteral("Apple");
 #elif defined(__LINUX__)
     QString base = QStringLiteral("Linux");
 #elif defined(__WINDOWS__)
--- a/src/util/versionstore.cpp
+++ b/src/util/versionstore.cpp
@@ -76,8 +76,6 @@ QString VersionStore::platform() {
     QString base = QStringLiteral("iOS");
 #elif defined(Q_OS_MACOS)
     QString base = QStringLiteral("macOS");
-#elif defined(__APPLE__)
-    QString base = QStringLiteral("Apple");
 #elif defined(__LINUX__)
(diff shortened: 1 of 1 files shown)


src/util/versionstore.cpp
```diff
@@ -72,8 +72,12 @@ QString VersionStore::applicationName() {
 
 // static
 QString VersionStore::platform() {
-#ifdef __APPLE__
+#ifdef Q_OS_IOS
+    QString base = QStringLiteral("iOS");
+#elif defined(Q_OS_MACOS)
     QString base = QStringLiteral("macOS");
+#elif defined(__APPLE__)
```
JoergAtGithub, line 79: Is this APPLE branch to catch other Apple OSes like iPadOS?

JoergAtGithub:
AFAIK, xwax which we use for vinyl control has no App Store Clause too.

src/util/versionstore.cpp
fwcd: Yes, though iPadOS will fall under iOS from what I understand. Just felt weird to leave let that branch fall through, so why not handle it

fwcd:
Good point, though the library stuff is coming later.

src/util/versionstore.cpp
JoergAtGithub: Than it's an unknown OS, which is already covered at the end.

JoergAtGithub (approved):
LGTM now! Waiting for CI.

daschuer:
By the way, did you use a special iOS vcpkg? Is it possible to crosscompile for the "arm64-ios" triplet on a x64 macOS?

daschuer (approved):
All Green. Thank you.

fwcd:
> By the way, did you use a special iOS vcpkg? Is it possible to crosscompile for the "arm64-ios" triplet on a x64 macOS?

Probably, I've crosscompiled from arm64-osx and there are still some packages I had to patch (most was minor stuff, like a GL ES 3.0 flag in `qtbase`, the largest one is probably a patch adding iOS support to CoreAudio, not sure what the upstream status of that is). Details are here: https://github.com/fwcd/m1xxx/issues/16, this is my vcpkg branch: https://github.com/fwcd/vcpkg/tree/mixxx/ios-qt6

JoergAtGithub:
The review and development cycles of https://github.com/PortAudio/portaudio/ are very long. I recommend to open a PR as early as possible, to ensure, that it's available, when we need it.

fwcd:
Done:

- https://github.com/PortAudio/portaudio/pull/881

PR #12672, opened 2024-01-28
CMakeLists: Support building for iOS

fwcd:
This PR implements the necessary changes to instruct CMake to build for iOS. This mainly boils down to:

- Setting `CMAKE_SYSTEM_NAME` to iOS
- Scoping macOS-specific stuff under a `NOT IOS` condition
- Linking iOS-specific frameworks (UIKit, OpenGLES)
- Linking the iOS Qt integration plugin
- Packaging (sold separately in #12676)

The rationale for targeting 2.4 is that, even if we don't build 2.4 for iOS, we avoid painful merge conflicts in the future by already having the relevant conditionals in place.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,22 +158,31 @@ if(APPLE)
     message(FATAL_ERROR "'xcode-select -v' failed with '${XCODE_SELECT_RESULT}'. You may need to install Xcode and run 'sudo xcode-select --install'.")
   endif()
 
-  if(VCPKG_TARGET_TRIPLET MATCHES "^arm64-osx")
-    # Minimum macOS version for arm64 Support
-    set(CMAKE_OSX_DEPLOYMENT_TARGET 11.0 CACHE STRING "Minimum macOS version the build will be able to run on")
-    set(CMAKE_OSX_ARCHITECTURES arm64 CACHE STRING "The target architecture")
-    set(CMAKE_SYSTEM_PROCESSOR arm64 CACHE STRING "The target system processor")
-    set(CMAKE_SYSTEM_NAME Darwin CACHE STRING "Setting this enables CMAKE_CROSSCOMPILE")
-  else()
-    if(QT6)
-      # Minimum macOS version supported by Qt 6
-      set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum macOS version the build will be able to run on")
+  if(VCPKG_TARGET_TRIPLET MATCHES "^[a-zA-Z0-9]+-osx")
+    message(STATUS "Targeting macOS (${VCPKG_TARGET_TRIPLET})")
+    set(CMAKE_SYSTEM_NAME Darwin CACHE STRING "Target macOS")
+    if(VCPKG_TARGET_TRIPLET MATCHES "^arm64-")
+      # Minimum macOS version for arm64 Support
+      set(CMAKE_OSX_DEPLOYMENT_TARGET 11.0 CACHE STRING "Minimum macOS version the build will be able to run on")
+      set(CMAKE_OSX_ARCHITECTURES arm64 CACHE STRING "The target architecture")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -160,28 +160,31 @@ if(APPLE)
...
-      add_compile_options(-fno-aligned-allocation)
     endif()
+  elseif(VCPKG_TARGET_TRIPLET MATCHES "^[a-zA-Z0-9]+-ios")
+    message(STATUS "Targeting iOS (${VCPKG_TARGET_TRIPLET})")
+    set(CMAKE_SYSTEM_NAME iOS CACHE STRING "Target iOS")
+    set(CMAKE_OSX_DEPLOYMENT_TARGET 14.0 CACHE STRING "Minimum iOS version to target")
+  else()
+    message(WARNING "Targeting an Apple platform, but VCPKG_TARGET_TRIPLET is not set. This is not a supported scenario!")
   endif()
 endif()
```
fwcd, line 189: Implementation note: This logic slightly refactors the macOS part too and is based on the current `main`. I believe the clear separation between macOS (arm64/x64) and iOS is worth it.

On the next 2.4 -> `main` merge this _will_ produce a conflict, in which case this entire incoming hunk should be picked, as it already includes the `main`'s changes (mostly just making the `arm64` regex more general).

JoergAtGithub:
Is 2.4 really the correct target for this PR? It should be for bugfixes only.

fwcd:
Since the majority of this logic doesn't apply to the desktop 2.4 builds anyway, I think not having to deal with merge conflicts would be worth it here (even minor things such as adding a source file could get annoying when 2.5 wraps everything into a `NOT IOS` conditional).

daschuer:
This change looks to risky for me at least in the current phase of 2.4.0 release candidate. 
I am in doubt that it is reasonable to introduce an iOS version based on Qt5 unless we have specific reasons for it.
How about moving this to 2.5 and Qt6?

daschuer (changes requested):
Let's fix our plan before merge.

fwcd:
Seems reasonable, I'll rebase this one and #12676.

Could you take a look at merging 2.4 to `main`? That would let me rebase my iOS branch to see what's still missing

daschuer:
I can do it tonight, it is also welcome to do it yourself. Don't know if there are difficult conflicts. 
I use SmartGit for a nice 3 way diff.

fwcd:
Marking as a draft for now, once #12685 is merged, I'll retarget this to `main` and push my local conflict resolutions (if I do it now, the GitHub labeler will run wild and the PR will blow up with commits again)

CMakeLists.txt
```diff
@@ -158,22 +158,31 @@ if(APPLE)
...
+        # Minimum macOS version supported by Qt 6
+        set(CMAKE_OSX_DEPLOYMENT_TARGET 10.15 CACHE STRING "Minimum macOS version the build will be able to run on")
+      else()
+        # Minimum macOS version supported by Qt 5.12
+        set(CMAKE_OSX_DEPLOYMENT_TARGET 10.12 CACHE STRING "Minimum macOS version the build will be able to run on")
+        # Needed for deployment target < 10.14
+        add_compile_options(-fno-aligned-allocation)
+      endif()
     endif()
+  elseif(VCPKG_TARGET_TRIPLET MATCHES "^[a-zA-Z0-9]+-ios")
```
daschuer, line 180: We have now a redundant way to select IOS, can we ditch the external IOS flag and handle everything automatically using the VCPKG_TARGET_TRIPLET?

daschuer (changes requested):
I think at least a check of IOS is missing when dealing with the triplets.

CMakeLists.txt
fwcd: The `IOS` flag is set by CMake, so it's pretty much what we do: https://cmake.org/cmake/help/latest/variable/IOS.html#ios

daschuer: Ah ok, I was not aware.

daschuer (approved):

PR #12673, opened 2024-01-28
ScreenSaverHelper: Implement screen lock suspension for iOS

fwcd:
This implements screen lock inhibition/suspension for iOS. Uses Objective-C++, though the implementation is rather trivial.

See https://stackoverflow.com/questions/32775828/how-to-prevent-screen-lock-ios-with-qt


daschuer (approved):
LGTM, Thank you.

PR #12674, opened 2024-01-28
DesktopHelper: Cond-compile out process creation on iOS

fwcd:
Forking is not supported on iOS, so we'll emit a warning instead.

### DIFF
--- a/src/util/desktophelper.cpp
+++ b/src/util/desktophelper.cpp
@@ -90,12 +90,16 @@ bool selectInXfce(const QString& path) {
 #endif
 
 void selectViaCommand(const QString& path) {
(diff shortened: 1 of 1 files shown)


daschuer (changes requested):
Will this feature ever of showing a file in "finder" ever be available? What is equivalent for the use case? Or can we remove the feature entirely on iOS?

fwcd:
I'm not sure what the precise API is, but I _think_ there might be a way to deep link into the iOS files app (the native file browser on iOS, which only presents a small view of the file system, namely the user's documents and app sandboxes).

daschuer (approved):
LGTM, Thank you.

PR #12675, opened 2024-01-28
Sandbox: Add basic iOS support

fwcd:
This updates the cond-compiles (e.g. the `Security` headers are unavailable on iOS) and then adds basic support for sandboxing on iOS.

I am not sure if the implementation entirely works as intended (the bookmark part is from [this Qt bug](https://bugreports.qt.io/browse/QTBUG-67522)), since the iOS sandbox is a fair bit more restrictive compared to macOS, especially concerning bookmark creation. IIUC, we would need some special logic to store URLs returned by a file picker and call `startAccessingSecurityScopedResource` etc., though even then we likely cannot persist this temporary permission like on macOS and may have to resort to copying the file into a sandbox directory or similar.

In any case, it shouldn't hurt to already have this logic in place to make it compile and, if needed, we can always update it later (the worst that would currently happen is spamming the user with too many sandbox popups since the OS may not give the right permission for anything outside the app's sandbox).

### DIFF
--- a/src/util/sandbox.cpp
+++ b/src/util/sandbox.cpp
@@ -4,12 +4,16 @@
 #include <QFileInfo>
 #include <QObject>
 #include <QtDebug>
+#include <QtGlobal>
 
 #include "util/mac.h"
 
 #ifdef __APPLE__
 #include <CoreFoundation/CoreFoundation.h>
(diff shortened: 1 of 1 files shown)


daschuer (approved):
LGTM, Thank you.

PR #4545, opened 2021-11-28
Add a Big Sur-style rounded rectangle icon on macOS

fwcd:
Fixes https://bugs.launchpad.net/mixxx/+bug/1952532.

This PR adds a macOS Big Sur-style icon for macOS, following the conventions outlined in Apple's [Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/macos/overview/whats-new-in-macos/):

mixxx_macos

With this icon, Mixxx looks more 'at home' with other macOS apps:

new_dock
old_dock

The new icon is provided both as an SVG and as corresponding ICNS files.

PR #12676, opened 2024-01-28
iOS packaging: Add Info.plist, launch screen and app icon

fwcd:
This adds packaging assets used by the iOS app:

- A borderless rectangular app icon, based on #4545 
  
  - Like for the macOS icon, a script is provided to regenerate the PNG (which is actually used) from the SVG
- An Info.plist that e.g. also provides the permission to view Mixxx's sandbox directories from the iOS files app
- A launch screen that displays the Mixxx logo (which is included as PDF, the format iOS expects) before any UI is loaded:


fwcd:
Yeah, the problem is that iOS needs a borderless icon (as opposed to macOS), therefore we need a separate one. Technically we could use the bordered one, but the App Store generally disallows transparent icons and if iOS changes the corner radius again (they did that back in iOS 7 IIRC), the icon would look strange again.

JoergAtGithub:
LGTM! Thank you!

PR #12678, opened 2024-01-29
Device pixel non GL waveforms

daschuer:
This daws the waveforms directly in device resolution instead of paint them fist and then upscale them. 

Before:
grafik

After: 
grafik


m0dB:
LGTM

PR #12687, opened 2024-01-29
ConfigObject: Find resource path on iOS

fwcd:
This updates the `qResourcePath` logic to handle iOS app bundles (which use a flat directory layout with resources at the top level, see [the Apple docs](https://developer.apple.com/go/?id=bundle-structure)).

### DIFF
--- a/src/preferences/configobject.cpp
+++ b/src/preferences/configobject.cpp
@@ -63,6 +63,11 @@ QString computeResourcePathImpl() {
         else {
             qResourcePath = QCoreApplication::applicationDirPath();
         }
+#elif defined(Q_OS_IOS)
+        // On iOS the bundle contains the resources directly.
+        else {
+            qResourcePath = QCoreApplication::applicationDirPath();
+        }
 #elif defined(__APPLE__)
         else if (mixxxDir.cd("../Resources")) {
             // Release configuration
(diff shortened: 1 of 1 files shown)


daschuer (approved):
LGTM, 

Just to clarify: this is against the main branch. 
Will you now put everything iOS related there? That is OK. Just asking, because we have put similar low risk PRs to 2.4.

fwcd:
No strong feelings here. But if 2.4 is fine, let's do that (this likely won't produce a conflict either way).

src/preferences/configobject.cpp
```diff
@@ -63,6 +63,11 @@ QString computeResourcePathImpl() {
         else {
             qResourcePath = QCoreApplication::applicationDirPath();
         }
+#elif defined(Q_OS_IOS)
+        // On iOS the bundle contains the resources directly.
+        else {
+            qResourcePath = QCoreApplication::applicationDirPath();
+        }
 #elif defined(__APPLE__)
```
JoergAtGithub, line 71:
```suggestion
#elif defined(Q_OS_MACOS)
```

JoergAtGithub (changes requested):

src/preferences/configobject.cpp
fwcd: That's fair, in these cases my defensive coding instincts seem to kick in, which is why I like to leave these kinds of fallbacks, but in this case it indeed only makes sense on macOS since I'm not aware of any other Apple OS that uses this bundle structure.

JoergAtGithub:
LGTM! Thank you!

Please target future iOS PRs to main, as 2.4 becomes a bugfix only branch soon.

PR #12688, opened 2024-01-29
CmdlineArgs: Move config directory to a user-accessible location on iOS

fwcd:
Currently, Mixxx uses the default config directory both on iOS and macOS, namely

```
<sandbox home>/Library/Application Support/Mixxx
```

On macOS this is a fine choice since the user can access the file system normally via Finder, on iOS, however, only `<sandbox home>/Documents` is accessible to the user, via the Files app. For Mixxx, making the config directory accessible to the user has a number of benefits, including

- The ability to back up the mixxxdb
- Transparency to the user how disk storage is used (and the ability to clean up e.g. analysis files themselves)
- Access to log files which could be useful for debugging
- The ability to manually change configs if needed

For this reason, this PR proposes to move the default config directory, on iOS only, to

```
<sandbox home>/Documents/Library/Application Support/Mixxx
```

...thus letting the user browse this directory:

For more information about the iOS file system conventions, see https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html.

### DIFF
--- a/src/util/cmdlineargs.cpp
+++ b/src/util/cmdlineargs.cpp
@@ -11,6 +11,7 @@
 #include <QCoreApplication>
 #include <QProcessEnvironment>
 #include <QStandardPaths>
+#include <QtGlobal>
 
 #include "config.h"
 #include "defs_urls.h"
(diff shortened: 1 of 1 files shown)


daschuer:
This is a kind of hack around system requirements. I am afraid some normal backup scripts or other things will be tricked. 

Do we have alternatives? Is there an established way for storing data in the Documents folder? 

Maybe we should start like default, but add a helper that import and exports to the Document Directory or such. 
This will help even on other targets where less technical user struggle with such tasks.

fwcd:
Another IMO underrated advantage of having the config directory in the user-accessible documents directory is that all files should up in the storage view:



The default directory would leave the user wondering where all their space went while the analysis data silently occupies it.

JoergAtGithub:
LGTM! Thank you!

PR #12689, opened 2024-01-29
DlgPrefInterface: Disable tooltips on iOS by default

fwcd:
These are usually more obtrusive than helpful on touchscreens, especially when covering other controls, therefore this PR disables them, by default, on iOS.

### DIFF
--- a/src/preferences/dialog/dlgprefinterface.cpp
+++ b/src/preferences/dialog/dlgprefinterface.cpp
@@ -4,6 +4,7 @@
 #include <QList>
 #include <QLocale>
 #include <QScreen>
+#include <QtGlobal>
 
 #include "control/controlobject.h"
 #include "control/controlproxy.h"
@@ -439,7 +440,11 @@ void DlgPrefInterface::slotApply() {
 void DlgPrefInterface::loadTooltipPreferenceFromConfig() {
     const auto tooltipMode = static_cast<mixxx::TooltipsPreference>(
             m_pConfig->getValue(ConfigKey(kControlsGroup, kTooltipsKey),
+#ifdef Q_OS_IOS
+                    static_cast<int>(mixxx::TooltipsPreference::TOOLTIPS_OFF)));
+#else
                     static_cast<int>(mixxx::TooltipsPreference::TOOLTIPS_ON)));
+#endif
     switch (tooltipMode) {
     case mixxx::TooltipsPreference::TOOLTIPS_OFF:
         radioButtonTooltipsOff->setChecked(true);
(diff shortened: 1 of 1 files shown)


src/preferences/dialog/dlgprefinterface.cpp
```diff
@@ -439,7 +440,11 @@ void DlgPrefInterface::slotApply() {
 void DlgPrefInterface::loadTooltipPreferenceFromConfig() {
     const auto tooltipMode = static_cast<mixxx::TooltipsPreference>(
             m_pConfig->getValue(ConfigKey(kControlsGroup, kTooltipsKey),
+#ifdef Q_OS_IOS
```
JoergAtGithub, line 443: Is the OS the apprpriated criteria here? Shouldn't it be screen size, or screen size + touch instead? This also applies to RPi etc.

fwcd: Screen size would be debatable IMO, touch seems reasonable, but remember that this is only a default value and setting it depending on whether there's a mouse currently connected doesn't make much sense IMO (or would lead to surprising behavior with missing tooltips once the user connects one).

If we'd want to go this route, we would likely have to dynamically update this option based on screen size/connected input devices etc. Such a mechanism would be a lot more complex to implement than this fix, which is safe to apply given that all iOS devices use a touchscreen as their primary input.

We could implement a more complex system in future PRs, but I'd consider that out of scope for this small UX improvement which is trivial to replace.

ronso0:
I'm wondering, how are tooltips triggered on touch devices?
Does the pointer remain on the last touch point?

Did you intend to disable tooltips altogether?
If so, there's also
https://github.com/mixxxdj/mixxx/blob/def8d1537f82e22aa0ca2109847ad6080980c82a/src/mixxxmainwindow.cpp#L96
and
https://github.com/mixxxdj/mixxx/blob/def8d1537f82e22aa0ca2109847ad6080980c82a/src/mixxxmainwindow.cpp#L168-L174
and
https://github.com/mixxxdj/mixxx/blob/def8d1537f82e22aa0ca2109847ad6080980c82a/src/mixxxmainwindow.cpp#L1132-L1155

fwcd:
I couldn't get tooltips to trigger reliably, usually they would pop up on the last tapped control after I've already tapped something else, but only every now and then (just often enough to become a minor annoyance, because they seem to block touches to the underlying UI).

> Did you intend to disable tooltips altogether?
> If so, there's also ...

Ah interesting, I didn't know about those! I think it's reasonable to leave it opt-in for now, that way adventurous users could still enable them manually in the settings (or users wishing to connect a mouse to their iOS device), even if the vast majority likely won't find them useful on iOS.

fwcd:
Can we move forward with this PR or is there a reason we should hold off on this fix?

ronso0:
LGTM, I think `slotResetToDefaults()` should be adjusted accordingly?

fwcd:
Updated, thanks

ronso0:
Thank you, looks good!

PR #12690, opened 2024-01-29
iTunes: Add iOS importer using the Media Player framework

fwcd:
### Based on #12672 

This adds a new iTunes importer for iOS, using the Media Player framework as the iTunes Library framework is macOS-only. Since the APIs are pretty similar, the implementation is largely similar to the existing macOS importer.

One major difference to macOS is that iOS does not expose the track locations directly, but instead gives us a special URL[^1]:

```
ipod-library://item/item.m4a?id=...
```

Since the assumption that every track has an on-disk location is present in many parts of the codebase, we use an [`AVAssetExportSession`](https://developer.apple.com/documentation/avfoundation/avassetexportsession) to export the audio files into the Mixxx sandbox whenever a track from iTunes is added to the Mixxx library[^2]. These tracks are saved to

```
<path to sandbox>/Documents/Music/Mixxx/iTunes Tracks
```

### Future Improvements

- We currently use a `DispatchSemaphore` to synchronously export the track, even though the underlying API is asynchronous. This is non-ideal, since Xcode rightfully warns about potentially hanging the UI thread and thus causing a priority inversion. In practice this does not seem to be so bad, since we only exports single tracks at a time as the user selects them/load them to a deck, but ideally we should make this asynchronous. Retrofitting asynchrony into the synchronous track loading seems to be a larger task, however. One option would be to simply accept the (potentially still non-existent) output path as the export happens asynchronously, but that would require being more liberal with non-existent paths (e.g. in `BaseExternalTrackModel` and `BaseExternalPlaylistModel`).
- If Mixxx eventually moves to storing track locations as URLs, we could revisit simply storing the `ipod-library` URLs in the db and loading the track files without having to copy them into the sandbox first, where they are stored redundantly+

[^1]: Sandboxing on iOS is a bit more restrictive than on macOS, see #12675
[^2]: This happens whenever the user selects a track or drags it to the deck

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -946,6 +946,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/library/itunes/itunesfeature.cpp
   src/library/itunes/itunesimporter.cpp
   src/library/itunes/itunesplaylistmodel.cpp
+  src/library/itunes/itunestrackmodel.cpp
   src/library/itunes/itunesxmlimporter.cpp
   src/library/library_prefs.cpp
   src/library/library.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1575,6 +1575,15 @@ if(APPLE)
       src/soundio/soundmanagerios.mm
       src/util/screensaverios.mm
     )
+
+    option(IOS_ITUNES_LIBRARY "Native iOS music library integration" ON)
+    if(IOS_ITUNES_LIBRARY)
+      target_sources(mixxx-lib PRIVATE
+        src/library/itunes/itunesiosimporter.mm
+      )
+      target_link_libraries(mixxx-lib PRIVATE "-weak_framework MediaPlayer")
+      target_compile_definitions(mixxx-lib PUBLIC __IOS_ITUNES_LIBRARY__)
+    endif()
   else()
     target_sources(mixxx-lib PRIVATE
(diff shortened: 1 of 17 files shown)


fwcd:
Marking as a draft for now, while the importer "works"  (i.e. finds all the tracks and playlists), it's not really useful until we figure out how to actually play those tracks. Since that might potentially require changes here too, let's wait with this for now...

fwcd:
Note: We may have to investigate whether we need to call [`MPMediaLibrary.requestAuthorization`](https://developer.apple.com/documentation/mediaplayer/mpmedialibrary/1621276-requestauthorization)

acolombier (approved):
Code LGTM, but I don't have a device to test it own. Could anyone give this a go? Might be able to get this running on Simulator in Hackintosh but likely better if someone has a bare metal...

fwcd:
This can currently only be tested from the branch linked in https://github.com/fwcd/m1xxx/issues/16 until all iOS patches have been upstreamed (most of them have been merged, but there's some minor stuff here and there still missing a PR, see the TBD section in that issue).

Since this importer implementation is pretty much entirely isolated from the remaining iTunes integration, I don't see any harm in merging this, since it shouldn't affect anything outside of the iOS build.

acolombier:
Sounds good then, thanks for your work!

acolombier:
Fixed in #14369 14369

PR #12698, opened 2024-01-30
DesktopHelper: Add openUrl abstraction to support iOS

fwcd:
As mentioned in https://github.com/mixxxdj/mixxx/pull/12674#issuecomment-1913793086, iOS supports opening files in the native Files app by replacing `file://` URLs with `shareddocuments://` URLs. To avoid duplication, this PR introduces a new abstraction, `DesktopHelper::openUrl` which serves as a drop-in replacement for `QDesktopServices::openUrl` that handles this schema mapping on iOS.

Since #12688 makes the config directory (including controller mappings etc.) user-accessible, links to the settings directory or user mappings "just work", even on iOS:

https://github.com/mixxxdj/mixxx/assets/30873659/3509eb2a-0df5-466c-b3e5-16a104255fb7

### DIFF
--- a/src/util/desktophelper.cpp
+++ b/src/util/desktophelper.cpp
@@ -19,7 +19,7 @@ const QString kSelectInXfce = "xf";
 QString sSelectInFileBrowserCommand;
 
 QString getSelectInFileBrowserCommand() {
-#if defined(Q_OS_MAC)
+#if defined(Q_OS_MACOS)
     return "open -R";
 #elif defined(Q_OS_WIN)
     return "explorer.exe /select,";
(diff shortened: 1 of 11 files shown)


src/util/desktophelper.cpp
```diff
@@ -165,10 +165,23 @@ void DesktopHelper::openInFileBrowser(const QStringList& paths) {
...
     }
 }
 
+bool DesktopHelper::openUrl(const QUrl& url) {
+#ifdef Q_OS_IOS
+    QUrl urlToOpen = url;
+    // Open files and folders in the iOS Files app
+    if (urlToOpen.scheme() == "file") {
+        urlToOpen.setScheme("shareddocuments");
+    }
```
Swiftb0y, line 180: this could use a link to the official docs or smth.

Swiftb0y (commented):
lgtm otherwise.

src/util/desktophelper.cpp
Swiftb0y: better than nothing I guess. though I'm very surprised that this is not official.

Swiftb0y (approved):
LGTM. Thank you

PR #12703, opened 2024-01-30
Tango: allow to toggle crossfader indepedently from mixer

### LINKED ISSUE #12654: Crossfader visible all time
### Feature Description

First thanks for all Mixxx developpers. 

I play Swing musics for couple dancers (that means no controller, no effect, no filter, no samples, to play music from beginning to end with a silent between tracks). Tango Skin.

I use only AutoDJ without using the Mixer itself, so I can reduce a lot the items of the screen. Reduction of items is even more interesting for beginner people who are afraid to have so many information that they do not need. It always a difficult step to explain Mixxx to people who  they discover Mixxx.

When I hide the the Mixer (button Toggle Mixer), I do not see anymore on which side the crossfader is. I need to see the crossfader side all time, also because when I stop AutoDJ I need to see that crossfader .

My suggest is to let crossfaser visible all time, even when Mixer is hidden.

ronso0:
As suggested in #12654 
This makes sense in Tango since the crossfader is separate from the mixer.

### DIFF
--- a/res/skins/Tango/skin_settings.xml
+++ b/res/skins/Tango/skin_settings.xml
@@ -506,15 +506,15 @@ Description:
           <WidgetGroup><!-- EQ knobs, EQ Kill buttons, Crossfader -->
             <ObjectName>SkinSettingsSubMenu</ObjectName>
             <Layout>stacked</Layout>
-            <MinimumSize>182,39</MinimumSize>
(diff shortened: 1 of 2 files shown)


daschuer (approved):
Works good. Thank you.

PR #12704, opened 2024-01-30
History: show only current year's playlists at the top level

ronso0:
I found it confusing that when I
* have a few History playists from 2024 (not yet enough to start a 2024 log group) and
* delete some of them, that
* playlists from 2023 are moved to the top level

IMO 'grouped by year' should be strict, unless the first group is from this year:
* we are in January 2025, I have like 8 playlists from 2025
* remove all but one (= remove entire 2025 group and remaining 2025 top-level playlists)
* = **one** 2025 playlist at the top level, playlists from 2024 remain in the **2024** group

### DIFF
--- a/src/library/trackset/setlogfeature.cpp
+++ b/src/library/trackset/setlogfeature.cpp
@@ -1,6 +1,5 @@
 #include "library/trackset/setlogfeature.h"
 
-#include <QDateTime>
 #include <QMenu>
 
 #include "library/library.h"
@@ -229,6 +228,7 @@ QModelIndex SetlogFeature::constructChildModel(int selectedId) {
     int nameColumn = record.indexOf("name");
     int idColumn = record.indexOf("id");
     int createdColumn = record.indexOf("date_created");
+    int currentYear = QDate::currentDate().year();
 
     // Nice to have: restore previous expanded/collapsed state of YEAR items
     clearChildModel();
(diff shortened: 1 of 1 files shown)


daschuer:
I can confirm that it is a surprising behaviour, when after deleting the 2024 folder disappears and playlist are popping up form the 2023 folder. 

On the other hand seeing the top 5 playlist is broken, when starting Mixxx on new year. Since saving clicks to reach the recent playlist is, also a higher value, I hesitate to merge this. 

Can we have both? Applying this change only when deleting Playlists but not after restart. 
I have in mind to not repopulate the 5 top playlist when deleting.

ronso0:
> Since saving clicks to reach the recent playlist is, also a higher value, I hesitate to merge this.

Yes, I understand.

> Can we have both? Applying this change only when deleting Playlists but not after restart.
I have in mind to not repopulate the 5 top playlist when deleting.

I'm not sure if that is feasible. The tree model is also rebuilt when a new playlist is started, so _not_ pulling up playlist from the next Year folder after deletion would create an inconsistent UX.

Therefore, I'd chose always or never here.
My motivation for this was to have 'year over, case closed' picture, and I understand that this might not be expected or appreciated by all users, so I'm okay with closing this and keeping it in my personal branch.

ronso0:
I'm using this in my production branch in case anyone wants to pick it https://github.com/ronso0/mixxx/tree/rop3210

PR #12707, opened 2024-01-31
ControlModel: avoid duplicates in developer tools window

ronso0:


src/control/controlmodel.cpp
```diff
@@ -16,11 +16,19 @@ ControlModel::ControlModel(QObject* pParent)
     const QList<QSharedPointer<ControlDoublePrivate>> controlsList =
             ControlDoublePrivate::getAllInstances();
 
+    QSet<ConfigKey> controlKeys;
```
Swiftb0y, line 19: ahead of time, predictable construction is probably preferable over doing it in the loop. wdyt?
```suggestion
    const auto aliases = ControlDoublePrivate::getControlAliases();
    QSet<ConfigKey> controlKeys(aliases.cbegin(), aliases.cend()); 
```

src/control/controlmodel.cpp
```diff
@@ -16,11 +16,19 @@ ControlModel::ControlModel(QObject* pParent)
...
     for (const QSharedPointer<ControlDoublePrivate>& pControl : controlsList) {
         if (!pControl) {
             continue;
         }
 
+        // Skip duplicates
+        if (controlKeys.contains(pControl->getKey())) {
+            continue;
+        }
+        controlKeys.insert(pControl->getKey());
```
Swiftb0y, line 30:
```suggestion
```

Swiftb0y (commented):
I'm afraid there is no solution that would be able to avoid the temporary container. I'd prefer a different implementation though.

src/control/controlmodel.cpp
```diff
@@ -16,11 +16,19 @@ ControlModel::ControlModel(QObject* pParent)
     const QList<QSharedPointer<ControlDoublePrivate>> controlsList =
             ControlDoublePrivate::getAllInstances();
 
+    const auto aliases = ControlDoublePrivate::getControlAliases();
+    QSet<ConfigKey> controlKeys(aliases.cbegin(), aliases.cend());
```
ronso0, line 20: We need the `key`s here, but `aliases` is `QHash<ConfigKey, Configkey>`.
And if we use `aliases` directly and do
```
if (aliases.contains(pControl->getKey())) {
    continue;
}
```
we'd skip both the original and the alias :|

Unless you have another idea I'd use my original implementation.

ronso0: Btw I'm surprised some CI runners build this.

Swiftb0y: > we'd skip both the original and the alias :|

Oh, right. Yeah, the problem with the "lazy" solution (in terms of evaluation, not the quality of your approach ;) ) is that it depends on the iteration order whether the alias control or the actual control is hidden. I'd like to avoid that because the iteration order is dependent on the iteration order of a QHash. In this concrete case, it may be irrelevant. 

> Btw I'm surprised some CI runners build this.

I'm rather surprised some don't. Are the MacOS runners still not on Qt5.14?

Swiftb0y: fyi the initial proposal works (at least it builds controlKeys) correctly, because the returned QHash has the original configkey as the hash key and the alias configkey as the value. Dereferencing a QHash::const_iterator instance (which is what the QSet ctor is doing) returns the value of that QHash entry. 
It's only a shame that we have to do it lazily...

Swiftb0y: All of this would be much easier if `ControlDoublePrivate::getAllInstances();` would not return the deprecated controls...

ronso0: > Are the MacOS runners still not on Qt5.14?

IIUC we're still uing 5.12 for the Ubuntu 20.04 ppa builds :|

ronso0: > it depends on the iteration order whether the alias control or the actual control is hidden.

Yeah, I didn't care about that fact here.
If we agree to revert I'll add a comment.

Swiftb0y: yes, pls revert to your previous solution. Mine is undoubtedly flawed.

ronso0: Alright, done.
Thanks for you review though, I learned something.

Swiftb0y (approved):
LGTM as a bandaid fix.

ronso0:
Shall I add a TODO, maybe someone wants to fix it another way?

Swiftb0y:
I'm not seeing any trivial way. May make sense to filter before constructing the return value of `ControlDoublePrivate::getAllInstances()` but that just moves the necessary temporary container. I don't see any way to avoid that easily.

PR #12713, opened 2024-02-02
DlgTrackInfo: remove redundant (explicit) update call

ronso0:
When writing the new Trackrecord `updateFromTrack` was called twice:
first by the Track::changed() signal
then explicitly

Doesn't cause issues, it's just not needed.


daschuer (approved):
It is still working. Thank you.

PR #12714, opened 2024-02-02
SoundManager: Set up `AVAudioSession` on iOS

fwcd:
### Based on #12672 

On iOS, audio apps should set up an [`AVAudioSession`](https://developer.apple.com/documentation/avfaudio/avaudiosession) to tailor the OS behavior towards their use case (e.g. whether the app should interrupt other audio etc.) As explained in https://github.com/PortAudio/portaudio/pull/881, this is currently not handled by PortAudio and likely doesn't fit into its API, since different apps will have different requirements that can't be handled by a "one-size-fits-all" initialization.

For Mixxx, I've set up the session in a way that it mixes with other apps[^1] and uses the standard playback mode.

[^1]: Not strictly necessary, but since we're mixing our own decks anyway, I see no downside in mixing with other apps too? Also, we wouldn't interrupt audio when launching Mixxx that way (an alternative would be to activate the audio session first upon playing a track, but that would likely require the `AVAudioSession` logic to be more deeply intertwined with other playback logic, which might not be desirable in order to keep platform-specific stuff isolated).

### DIFF
--- a/src/soundio/soundmanagerios.mm
+++ b/src/soundio/soundmanagerios.mm
@@ -1,0 +1,31 @@
+#include <QDebug>
+#include <QtGlobal>
+
+#import <AVFAudio/AVFAudio.h>
+
+namespace mixxx {
+
+void initializeAVAudioSession() {
+    AVAudioSession* session = AVAudioSession.sharedInstance;
+    AVAudioSessionCategory category = AVAudioSessionCategoryPlayback;
+    AVAudioSessionMode mode = AVAudioSessionModeDefault;
+    AVAudioSessionCategoryOptions options =
+            AVAudioSessionCategoryOptionMixWithOthers |
+            AVAudioSessionCategoryOptionAllowAirPlay |
+            AVAudioSessionCategoryOptionAllowBluetoothA2DP;
+
+    NSError* error = nil;
+    [session setCategory:category mode:mode options:options error:&error];
+    if (error != nil) {
+        qWarning() << "Could not initialize AVAudioSession:"
+                   << error.localizedDescription;
+    }
+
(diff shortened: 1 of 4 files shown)


src/soundio/soundmanagerios.mm
```diff
@@ -0,0 +1,31 @@
...
+void initializeAVAudioSession() {
+    AVAudioSession* session = AVAudioSession.sharedInstance;
+    AVAudioSessionCategory category = AVAudioSessionCategoryPlayback;
+    AVAudioSessionMode mode = AVAudioSessionModeDefault;
+    AVAudioSessionCategoryOptions options =
+            AVAudioSessionCategoryOptionMixWithOthers |
+            AVAudioSessionCategoryOptionAllowAirPlay |
+            AVAudioSessionCategoryOptionAllowBluetoothA2DP;
+
+    NSError* error = nil;
```
daschuer, line 17: What I have read is: nil is defined as (id)0. I think we should use NULL here in objective C.

daschuer: Or is the pointer asterisk wrong? Because of & below?

fwcd: By convention, `nil` is the preferred spelling for Objective-C objects. And the double pointer is intended, as every Objective-C object has to be a pointer, but the API additionally needs to reassign the value here (i.e. it is an output parameter)

daschuer: Thank you for explaining. I cross read a bit about the topic and I think I got it.

daschuer (approved):
LGTM thank you.

PR #12716, opened 2024-02-02
SoundManager: Use correct PortAudio backend on iOS

fwcd:
The iOS backend (https://github.com/PortAudio/portaudio/pull/881) is called `iOS Audio`, therefore we should pick the correct default here.


daschuer (approved):
LGTM, thank you.
I'm looking forward to your vcpkg PR.

PR #12733, opened 2024-02-05
(fix) hotcue_X_color control: color not stored in cue

ronso0:
Fixes the hotcue color issue reported in https://mixxx.zulipchat.com/#narrow/stream/113295-controller-mapping/topic/Hotcue.20setting.20script

For some reason the valueChanged() signal is not emitted, therefore slotHotcueColorChanged() not being called and the new color not stored in the cue. The color control is reset as soon as anothe cue is set (track emits cuesUpdated() adn cues are reloaded from track, default color).
Reproducible with Qt5 and Qt6.
without this fix:
https://github.com/mixxxdj/mixxx/assets/5934199/aad6ae0f-2ea3-4825-ae77-7da8ba95b22b

For testing, use this mapping with virtual MIDI device:
load a track to deck1, click Eject on deck2
https://gist.github.com/ronso0/471efe54370a32fa9eb26a571a818dd5

### DIFF
--- a/src/engine/controls/cuecontrol.cpp
+++ b/src/engine/controls/cuecontrol.cpp
@@ -2477,11 +2477,6 @@ HotcueControl::HotcueControl(const QString& group, int hotcueIndex)
             this,
             &HotcueControl::slotHotcueColorChangeRequest,
             Qt::DirectConnection);
-    connect(m_hotcueColor.get(),
-            &ControlObject::valueChanged,
-            this,
-            &HotcueControl::slotHotcueColorChanged,
-            Qt::DirectConnection);
 
     m_hotcueSet = std::make_unique<ControlPushButton>(keyForControl(QStringLiteral("set")));
     connect(m_hotcueSet.get(),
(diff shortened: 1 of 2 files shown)


src/engine/controls/cuecontrol.cpp
```diff
@@ -2656,6 +2656,10 @@ void HotcueControl::slotHotcueColorChangeRequest(double color) {
     }
     // qDebug() << "HotcueControl::slotHotcueColorChangeRequest" << color;
     m_hotcueColor->setAndConfirm(color);
+    // Note: for some reason the valueChanged() signal is not emitted,
```
daschuer, line 2659: This is blocked to avoid endless loops. I think due to the read-only nature, slotHotcueColorChanged() is never called and can be inlined. setAndConfirm() should then only be set at the end when all other checks also have been performed successful.

daschuer (changes requested):

src/engine/controls/cuecontrol.cpp
ronso0: Please elaborate, I'm not very familiar with the contrl internals.
I see 
https://github.com/mixxxdj/mixxx/blob/94ffc14168c6e044fb36e7d32dbfaaef0026332c/src/control/control.cpp#L289-L304
This emits `valueChanged`, not a value change request. Where is that blocked? I don't see how that would create an infinite loop.

`slotHotcueColorChanged()` is indeed never called, and we also have an unused signal `hotcueColorChanged`.
I'll adjust that.

daschuer: It is blocked here: 
https://github.com/mixxxdj/mixxx/blob/94ffc14168c6e044fb36e7d32dbfaaef0026332c/src/control/controlobject.cpp#L50

ronso0: Thanks!

daschuer (approved):
Thank you. LGTM

PR #12739, opened 2024-02-05
fix Loop_out not seeking back

ronso0:
This fixes seeking in the following situation I noticed in https://github.com/mixxxdj/mixxx/pull/12522#issuecomment-1879865194
* enable quantize
* move playhead slightly after a beat
* set `loop_in`
= snaps to previous beat
* jump forward by 4 beats
* set `loop_out`
= snaps to previous beat, playhead moves 4 beats back inside the loop
* deactivate loop
* jump forward by 4 beats
* set `loop_out`
(= previously: snaps to prev. beat, playhead doesn't move, loop will never be reached)
= **FIX:** snaps to previous beat, playhead moves 4 beats back inside the loop

### Note:
As before, (re)setting loop_out with playpos exactly at loop_out of the inactive loop, playpos should stay at loop_out and **not** jump to loop_in.

See discussion on Zulip https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/2.2E4.20behaviour.20.2F.20failing.20test.20of.20loop_in.2F_out.20with.20quantize

## TODO
- [x] Adjust `LoopInOutButtons_QuantizeEnabled`

### DIFF
--- a/src/test/looping_control_test.cpp
+++ b/src/test/looping_control_test.cpp
@@ -322,7 +322,8 @@ TEST_F(LoopingControlTest, LoopOutButton_AdjustLoopOutPointInsideLoop) {
     EXPECT_FRAMEPOS_EQ_CONTROL(mixxx::audio::FramePos{1500}, m_pLoopEndPoint);
 }
 
-TEST_F(LoopingControlTest, LoopInOutButtons_QuantizeEnabled) {
+// Disabled because seek is not queued as expected https://github.com/mixxxdj/mixxx/pull/12739
+TEST_F(LoopingControlTest, DISABLED_LoopInOutButtons_QuantizeEnabled) {
     const auto bpm = mixxx::Bpm{60};
     m_pTrack1->trySetBpm(bpm);
     m_pQuantizeEnabled->set(1);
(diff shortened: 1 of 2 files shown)


ronso0:
The test `LoopInOutButtons_QuantizeEnabled` should fail now.

ywwg (approved):
lgtm

ywwg:
extremely annoyingly, the first time I hand-tested it appeared the fix did not work, but every time I tried afterward it did.  Is there a chance that old_loop_info could be blank, but an out point is set?

ywwg:
Unless someone else can reproduce the failure, I think this is a Good Enough fix for the release.  If we find a further edge case later we can fix it in 2.4.1

ronso0:
Draft until the test is fixed.

ronso0:
I disabled the test and filed #12741

src/test/looping_control_test.cpp
```diff
@@ -322,31 +322,47 @@ TEST_F(LoopingControlTest, LoopOutButton_AdjustLoopOutPointInsideLoop) {
     EXPECT_FRAMEPOS_EQ_CONTROL(mixxx::audio::FramePos{1500}, m_pLoopEndPoint);
 }
 
-TEST_F(LoopingControlTest, LoopInOutButtons_QuantizeEnabled) {
+TEST_F(LoopingControlTest, DISABLED_LoopInOutButtons_QuantizeEnabled) {
```
ywwg, line 325: can you add a short comment that links to the issue?

ronso0: Yup, done.

ywwg (approved):

PR #12742, opened 2024-02-06
Fix LoopingControlTest.LoopInOutButtons_QuantizeEnabled

daschuer:
A second ProcessBuffer() was needed. 
As an alternative, we could  check for a stopped deck before scheduling the seek, but we are stopped anyway and can effort the extra cycle.


ronso0:
Hmm, I tried exactly that but the tests failed anyway https://github.com/mixxxdj/mixxx/compare/83d31a337f20e89677fbcb81eb6459d50d1eaad7..30f9bfda9d8ac05fa938da23c1f1dc1706117d9e
Waiting for CI...

ronso0:
Great, thank you very much!

Issue #9365, opened 2022-08-23
Deleting track from disk leaves track listed in playlists, even though track no longer exists


mixxxbot:
Reported by: **[h-pa5l-v](https://launchpad.net/~h-pa5l-v)**
Date: 2018-07-01T22:07:13Z
Status: Confirmed
Importance: Medium
Launchpad Issue: [lp1779584](https://bugs.launchpad.net/bugs/1779584)
Tags: library, playlist

---

I deleted a track from my hard drive. In Mixxx, I ran "rescan library." This, correctly, removed the track from appearing in search results in my list of TRACKS in the Mixxx Library. However, the track still appears as available in my PLAYLISTS. Trying to PLAY the track fails with "track not found." In the middle of a DJ set, this is BAD, BAD, BAD. I assumed that PLAYLISTS were just links to tracks in the Master Library and that PLAYLIST track listings would be updated when "rescan library" was run. Apparently, this is not so. Unplayable, nonexistent tracks seem to be left in PLAYLISTS. There is no "rescan library" option for PLAYLISTS. Running ANALYZE on PLAYLISTS does not remove the bad listings. I thought that I remembered there being a toggle for "display unplayable tracks", but I couldn't find any such option. I have tracks added to several of my PLAYLISTS which point to the same, now nonexistent, track in Master Library. It will be painful to manually edit them all out. Thanks!

ronso0:
Marking tracks is possible, and easier than I initially thought :tada: See #12744 
If both 'played' and 'missing' are marked we should probably use RED or some other warning color for the latter.

ronso0:
I think we can close this as fixed, considering that playlists may contain _temporarily_ unavailable tracks and that missing tracks are now highlighted (default: red text color).

Please reopen if you disagree and have a better solution in mind.

PR #12744, opened 2024-02-06
Tracks: custom text color for played tracks (qss)

### LINKED ISSUE #5911: Display text for already played tracks in a different color
Reported by: **[esbrandt](https://github.com/esbrandt)**
Date: 2011-05-31T14:40:40Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp790726](https://bugs.launchpad.net/bugs/790726)
Tags: easy, polish, weekend

---

While the checkbox in the "PLAYED" column is helpful, the distinction between played and not yet played tracks is not clear enough. This is especially true if looking from some distance , in a rush or with difficult light conditions.

I propose to introduce a new key for use in skin.xml which allows to define a separate color for tracks already played in that session. If a tracks "PLAYED" checkbox gets unchecked it should obtain the original color again.

ronso0:
Quick POC, fixing #5911 

Set the custom property `qproperty-playedInactiveColor` in qss for painting the text of played tracks.
```
WTrackTableView {
  qproperty-playedInactiveColor: #555;
}
```
`BaseTrackTableModel::data()` uses this to override Qt::ForegroundRole (QBrush).
Works as desired, only the BPM delegate doesn't want to play along. **edit:** fix with explicit stylesheet.
image

Can be tested, but is still WIP (PaleMoon only) and I'm cleaning it up a bit. Will rebase onto main.

Alternatively, this could be an opacity factor (like that for the track colors), but I don't see a way to get that in the SQL model. I think with this we could also modify the 'selected' color.
I did a quick test by returning a TableItemDelegate for _every column_ and do the painting there, but it's not as straight forward as the current implementation.
___
While testing it I noticed that the focus border added in #3172 was only working for Tracks and other top-level features -- in AutoDJ, Analysis etc. it was the default color (WTrackTableView nested in LibraryView :shrug: ).
Fixed by connecting the property changed signal and listening to that where required.

### DIFF
--- a/res/skins/Deere/style.qss
+++ b/res/skins/Deere/style.qss
@@ -247,11 +247,14 @@ WTrackTableView,
 WTrackTableView QCheckBox:focus {
   outline: 1px solid #D6D6D6;
 }
-/* This uses a custom qproperty to set the focus border
-  for Color and Cover Art cells, 1px solid, sharp corners.
-  See src/library/tableitemdelegate.cpp */
 WTrackTableView {
+  /* This uses a custom qproperty to set the focus border
+     for Color and Cover Art cells, 1px solid, sharp corners.
+     See src/library/tableitemdelegate.cpp */
   qproperty-focusBorderColor: #D6D6D6;
+  /* This is the color used to paint the text of played tracks.
+     BaseTrackTableModel::data() uses this to override Qt::ForegroundRole (QBrush). */
+  qproperty-playedInactiveColor: #555;
 }
 
 WTrackTableView:focus,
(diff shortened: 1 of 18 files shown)


ronso0:
I played a session with my personal branch (didn't rmemeber I already merged this PR) and was surprised how well it works.
We should consider making the Played checkmark less dominant.

daschuer (approved):
LGTM, Nice feature, thank you.

ronso0:
Great this is in now!

As stated in #9365 we may consider to paint missing tracks, too. (didn't look into the details, yet, might be a bit more elaborate than the Played state which is a column)

ronso0:
This deserves a changelog entry.

PR #12745, opened 2024-02-07
Add beatloop anchor to set and adjust loop from either start or end

### LINKED ISSUE #12737: Ability to set and activate a beatloop from the end marker
### Feature Description

The usecase is the following: imagine you have a track about to finish and you don't have an extro in point set/hotcue. A great option would be to have a way to set a loop backward, by setting the end point rather than start point. This way, you can quickly set a 16 or 32 beats loop.

In order to be usable, this would also require to allow adjusting the beatloop length via `beatloop_size` even if the loop would go beyond the end of the track. The [logic](https://github.com/mixxxdj/mixxx/blob/fd41a3ad3f1a5f4b5fd21b09f953e3e8f3af37e9/src/engine/controls/loopingcontrol.cpp#L1476) in `LoopingControl::slotBeatLoop` could take into the account if a loop currently enable.

(Please help me finding a better title, I couldn't put words on my thoughts...)

acolombier:
Add the ability to define what is the static point of the loop or which point gets inherited from the current position. 

Better wording suggestion would be greatly appreciated!

[Kooha-2024-02-07-10-30-13.webm](https://github.com/mixxxdj/mixxx/assets/7086688/5f8c0e77-c706-4454-a53b-7df04ca0021a)

Resolves #12737


daschuer:
Thank you for strating this topic. Can you take s screenshot of the new button in it skin area? 

The icon itself does not work for me well, it looks like a beat grid edit icon. Did you consider to reuse the icons of the loop-in/out button? 

I can think of a loop-out button with a lock overlay for the reverse case. 

Or the full loop icon with a pointer or lock overlay right or left.

daschuer:
The most puzzling question is how make this accessible from the controllers. Did you consider this? 

We are currently facing the issue that the loop out button alone does nothing or surprising things. Did you consider to make use of it?

src/controllers/controlpickermenu.cpp
```diff
@@ -628,6 +628,15 @@ ControlPickerMenu::ControlPickerMenu(QWidget* pParent)
     QString beatJumpBackwardDescription = tr("Jump backward by %1 beats, or if a loop is enabled, move the loop backward %1 beats");
     addDeckControl("beatjump_forward", tr("Beat Jump / Loop Move Forward Selected Beats"), tr("Jump forward by the selected number of beats, or if a loop is enabled, move the loop forward by the selected number of beats"), beatJumpMenu);
     addDeckControl("beatjump_backward", tr("Beat Jump / Loop Move Backward Selected Beats"), tr("Jump backward by the selected number of beats, or if a loop is enabled, move the loop backward by the selected number of beats"), beatJumpMenu);
+    addDeckControl("beatjump_anchor",
```
daschuer, line 631: How about calling it just "loop_anchor"? A beat jump can be done without a loop and will jump the entire loop without in and out"

src/engine/controls/loopingcontrol.cpp
```diff
@@ -136,6 +136,29 @@ LoopingControl::LoopingControl(const QString& group,
...
+
+    m_pCOBeatLoopAnchorToggle = new ControlObject(
+            ConfigKey(group, "beatloop_anchor_toggle"), false);
+    connect(
+            m_pCOBeatLoopAnchorToggle,
+            &ControlObject::valueChanged,
+            this,
+            [this](double value) {
+                if (value > 0) {
+                    slotBeatLoopAnchorChangeRequest(static_cast<double>(
```
daschuer, line 156: I think you can read the value of m_pCOBeatLoopAnchorToggle right away using `toBool()` without any connection.

acolombier:
> Thank you for strating this topic. Can you take s screenshot of the new button in it skin area?

I have put a little screen recording in the PR description. Here is a focus on the section as well

image


> The icon itself does not work well for. It looks like a beat grid edit icon. Did you consider to reuse the icons of the loop-in/out button?

> I can think of a loop-out button with a lock overlay for the reverse case.

Yeah, I've used the loop-in SVG as a starting point, but I must confess I have no clue what I'm doing in Inkscape. Some help on the icons would be greatly appreciated :) 

I will update change the `start` one to be `loop-in` + the lock and same for `end`.

> The most puzzling question is how make this accessible from the controllers. Did you consider this?

Yep, I have already implemented it on my controller, before doing the UI work in order to test the feature on a real setup.

Currently with my S4 MK3 mapping, if you keep the "Sample" pressed, the `loop_anchor` is set to `End`. Using `loop_anchor_toggle`, one may also use a toggle rather than a hold. 

> We are currently facing the issue that the loop out button alone does nothing or surprising things. Did you consider to make use of it?

Yes, it does now weird thing consistently in the tho modes. At first glance, there seems to be multiple issues, but I believe this will be out of scope for the current piece of work.

src/engine/controls/loopingcontrol.cpp
acolombier: I've used `toBool()` but not sure how I can make the `toggle` work without connection. Could yous how me a example?

acolombier:
I have rework the icon to follow your suggestion @daschuer 

[Kooha-2024-02-07-11-09-43.webm](https://github.com/mixxxdj/mixxx/assets/7086688/308d082c-ce65-4bfb-868b-d79e6609cba1)

[Kooha-2024-02-07-11-19-15.webm](https://github.com/mixxxdj/mixxx/assets/7086688/98349db9-9d34-4dda-8b49-5df6cbbeb8b3)

daschuer:
Thank you for the new Icon. I think this is good enough to see the intention. We can polish it at any time. No need to learn Inkscape ;-)

src/engine/controls/loopingcontrol.cpp
daschuer: You need to make the control a toggle pushbutton like here: https://github.com/mixxxdj/mixxx/blob/7b3e48544a5a5475553742678115951179c41526/src/skin/skinloader.cpp#L233

acolombier: Sorry, I'm not sure how you want me to remove the connection. This should already be a `ControlPushButton`. In the example you've linked, there also seem to be a `connect`

```cpp
m_pSelectBigSpinnyCover = std::make_unique<ControlPushButton>(
            ConfigKey("[Skin]", "select_big_spinny_or_cover"), true);
m_pSelectBigSpinnyCover->setButtonMode(ControlPushButton::TOGGLE);
connect(m_pSelectBigSpinnyCover.get(),
            &ControlObject::valueChanged,
            this,
            &SkinLoader::updateSpinnyCoverControls);
```

Just to make sure you haven't missed it, but this `connect` is for the `toggle` (`loop_anchor_toggle`)  control, which flip the state of the main CO (`loop_anchor`), later use in the loop logic. Do you think I can still remove the connect for the toggle? If yes, could you make a code suggestion?

daschuer: The toggling is handled inside the control, via ControlPushButton::TOGGLE 
You need only to connect the signal when you like to react on the event. But this is not the case here.

src/engine/controls/loopingcontrol.cpp
```diff
@@ -136,6 +136,27 @@ LoopingControl::LoopingControl(const QString& group,
...
+    connect(
+            m_pCOLoopAnchorToggle,
+            &ControlObject::valueChanged,
+            this,
+            [this](double value) {
+                if (value > 0) {
+                    slotLoopAnchorChangeRequest(static_cast<double>(!m_pCOLoopAnchor->toBool()));
+                }
+            },
+            Qt::DirectConnection);
```
daschuer, line 159:
```suggestion
    m_pCOLoopAnchor = std::make_unique<ControlPushButton>(ConfigKey(group, "loop_anchor"));
    m_pCOLoopAnchor->setButtonMode(ControlPushButton::TOGGLE);
```

src/engine/controls/loopingcontrol.cpp
acolombier: Why do I not want to react to the toggle event here?  Since the loop anchor state is stored in an other CO, do I not need to replicate it here? I'm confused

daschuer: I think you can do it with just with one CO. But maybe I miss something?

acolombier:
I have removed the redundant CO (`loop_anchor_toggle`) and fixed the last UI weirdness on the classic color scheme. I have also increased the beatloop size spinbox by couple of pixel to align it with the button

[Kooha-2024-02-08-10-22-24.webm](https://github.com/mixxxdj/mixxx/assets/7086688/bb02d367-6dab-4951-8eb1-06507d37dd7e)

[Kooha-2024-02-08-10-21-51.webm](https://github.com/mixxxdj/mixxx/assets/7086688/561ff148-1677-43c3-80f1-55eb79216a10)

ronso0:
I'm wondering how this solves the use case mentioned in #12737 where you want to _spontaneously_ create a loop backwards from the playhead. Wouldn't that require to always keep the loop anchor set to End, but how does this work with the 'regular' beatloop usecase where you want a forward loop?

daschuer:
For my understanding you can switch between these two modes by the new toggle button. 

I have not yet played with it, but I can imagine that will become my new default. 

Yo can kind of listen the first path of the planed loop before enable it. Before you need to remove the loop before the loop end if you notice something is messed up.

ronso0:
> I'm wondering how this solves the use case mentioned in #12737 where you want to _spontaneously_ create a loop backwards from the playhead. Wouldn't that require to always keep the loop anchor set to End, but how does this work with the 'regular' beatloop usecase where you want a forward loop?

I think we should extend this by `LoopAnchorPoint::Auto` so the toggle has three states.
With Auto we can check if the desired loop fits in between playpos and track end, if not use `LoopAnchorPoint::End` and create the loop backwards.
Admittedly that makes `slotBeatLoop()` even more complex but it would solve the origianl issue.

What do you think?

acolombier:
I'm afraid it may lead to unexpected behavior, since it could mean that one could "accidentally" set a loop on the last X beats rather than on the next X ones. Say you want to put a 64 beats loop on the outro that just started, but it turns out that the outro only has 32 beats, you unexpectedly end up setting the loop on the last 64 beats, leading to a jump back in the verse or chorus preceding the outro.

It's worth to point out that, this PR in its current state has already solved the issue #12737, by:
- Change the beatloop selector unconditionally of the playposition and whether or not the new size would fit within the track boundary
- Allowing the loop to be set from th end

ronso0:
Yeah, I was just referring to the timing issue (emergency loop literally just before the track ends), and admittedly with shorter loop (4, 8 or 16 beats) for which you still have the waveform in sight, and for (my) electronic dance music where there's no siginificant fading outro.

For other music and longer loops this may indeed produce surprises.

acolombier:
Apollogies @ronso0 I missed the notification of your reply

> Maybe I can map that (toggle anchor mode, set loop) to Shift+loopMove encoder

Yep, that's exactly what I did on my [aggregate of feature branch for S4MK3](https://github.com/acolombier/mixxx/blob/traktor-s4-mk3/res/controllers/Traktor-Kontrol-S4-MK3.js#L2320). This way I can use all the the `loop_*` CO just as the same (`loop_size`, `loop_half`, ...)

(Just a demo - you can see the loop size indicator going orange to indicate the loop anchor mode change)

https://github.com/mixxxdj/mixxx/assets/7086688/e9f4dd1e-52e6-4855-b4a1-b4e9cc4576ae

acolombier:
Do you guys think that something that could make it to 2.5 by any chance?

ronso0:
I think so, yes.

Sorry I did not yet take the time to test this properly.
I hope I'll find time to test this with my controller next week (my setup is the studio / club).

Yet some more ideas how to implement this in mappings without having to select 'loop_out' anchor mode:
<sub>_based on the assumption that the anchor mode os mainly relevant for beatloops > 1 beat
also assuming there's no controller button left to be mapped to anchor mode toggle
and I don't want to reach for the trackpad to toggle anchor mode in the GUI, or reach for the keyboard_</sub>

**1 lonpress**
On longpress release (> 1 sec?) BeatLoop  uses the Out point as anchor.
The crucial point is to still have immediate response on (short) press, i.e. avoid loop wraparound before the button is released.

**1 A**: consider longpress only for beatloop_size >= 4
* **activate** loop on press (default loop_in anchor)
* **deactivate** if the loop is >= 4 beats and button is still down after ~1sec
* **activate** with loop_out anchor on release

or 
**1 B**: even smarter, with longpress duration _kLongDur_
* press: check if _kLongDur_ is shorter than selected loop length * playback rate
  * A no: activate (loop_in)
  * B yes: do nothing, start timer
* button still pressed after _kLongDur_:
   * 
* release: activate (loop_out)

or:
**2 Loop_out + BeatLoop**
* press & hold loop_out (yes, that would potentially activate the previous loop and drag loop_out with the play position)
* press BeatLoop to activate a loop with loop_out anchor
* restore the loop that was modified while dragging loop_out?
not necessary IIUC, loop_out is set on **release** so we just need to make sure that the `m_bAdjustingLoopOut` flag is reset when a new loop is activated and do nothing on loop_out release if that flag is not set (anymore).

I think the **2** is my favourite. Predictable, no confusion, no hacks required.

acolombier:
Thanks for the feedback @ronso0 

Perhaps one thing to start on the assumption. I think the loop anchor mode change goes beyond just a unique use-case: as @daschuer mentioned, being able to listen to the section before deciding to put a loop is pretty handy and thus the ability to change the "snapping" mode at runtime has proven to be quite useful during the last month or so of testing

> assuming there's no controller button left to be mapped to anchor mode toggle

It is a legit concern, IMHO we would have to decide on a per-controller basis. Personally, I tend to dislike long press but for timed action (e.g when not using quantize) as it makes it very hard to sync due to the needed "pre-shot" of the action (for for both the "short press" and the "long press", since the short press action can only be triggered at release - except if you want to trigger the two actions and undo the first one, which would be suboptimal in this case)

Also for example **2** isn't possible on the S4 Mk3 since there is no `loop_out`, and also not possible in the UI because of the double press.

Finally, I would like to bring to your attention that the loop mode doesn't only change how to loop get set but also how it gets resized. Something I have been using extensively is the reducing the loop by moving the start cue instead of end point (extremely useful for loop one vocal sample - say you put a 4 beat loop on a vocal and want to eventually have a 1 beat loop on the last beat, this is very easy to do, and particularly smooth since you can resize the loop and allow the playback to be before the starting point , making it easier to mix with additional actions, says on another deck)

ronso0:
> as @daschuer mentioned, being able to listen to the section before deciding to put a loop is pretty handy and thus the ability to change the "snapping" mode at runtime has proven to be quite useful during the last month or so of testing

Thanks, I'm aware of that feature. The only reason for me to think about implementations is that I'd like to keeo the IN anchor as default: I'm used to hit the BEATLOOP button on time (just slightly before a drum kick or so), and having to count beat markers to set the OUT point in time (IN is were I want it) requires to much "thinking" IMO and it's a tricky moment because the loop will instantly wrap around, whereas with the IN marker the loop is engaged unnoticed and I have some time to fix it (shift slight back/forth by 1/8 of a beat).

So yeah, I don't question the evaluated use cases, they just don't fit for _me_ : ) I'll probably only use it on rare occassions at the track end, or missed loop IN for other reasons.

I'm aware of the longpress downsides. I did focus only the controller-only use case with the mentioend limitations where I'd accept small compromises in terms of timing.

That being said, thanks for this new feature! very much appreciated.
Hope I'll find some time for a live test this week.

daschuer:
I am also not sure that AUTO will work. The issue is that you can not rely on it. You need always check if the picked loop is usable or on any outro garbage. The Ideal outro loop is on some beats without lyrics or lead instruments. 
For that purpose the reverse loop is ideal because of the pre-listening for free. 

All the long press tricks will not work, because we need immediate responds for the end loop in the emergency use case and it will also immediately mess up the mix when doing wrong.  

2. Sounds reasonable. Maybe we can implement that in any case. 

Ideas: 

**3.**
We have currently  "beatloop_activate" and "beatlooproll_activate". Maybe we can remove "beatlooproll_activate" because you can do it with slip mode + beatloop_activate and have the shift + "beatloop_activate" for reverse loops? 
This will not work because of the transition form old to new. 

**4.**
Swap in/out, keep size:  
1 beatloop_activate at loop in. 
2 Push Loop_out to fold the loop reverse 

This works for the emergency case where you realize the forward loop will produce rubbish, but probably not on a regular basis. 

**5.**
A set of "beatloop_r4_activate" controls for putting them on a keypad.

ronso0:
> 4. Swap in/out, keep size:
1 beatloop_activate at loop in.
2 Push Loop_out to fold the loop reverse

So a bit like my loop_out idea? (hold loop_out, press beatloop for backwards loop)
* press & hold beatloop (loop activated on press
* press loop_out for backwards loop from release position

acolombier:
I like **3** since that would encourage more granularity on the mapping and describe behaviour more explicitly (e.g "while you press this button, you also enable slip" - as it is the case with `beatlooproll_activate`) - it will still be possible the set `slip` and unset on release as part of the script callback. Although I guess it will induce a (neglectable?) performance decrease.

**4** will not work in case of emergency since you aren't able to set a lop beyond the end of a track, so `4.1` won't set the loop at all

I like **5** to although I thing it is mutual exclusive with **3** - you can also set `loop_anchor=end`, `beatloop_4_activate` and `loop_anchor=start` on release on a keypad mapping callback

daschuer:
> I like 5 too although I thing it is mutual exclusive with 3 - you can also set loop_anchor=end, beatloop_4_activate and loop_anchor=start on release on a keypad mapping callback

Yes, Drawback: It will mess with the loop_anchor button, and is not mappable via XML (GUI)

acolombier:
> not mappable via XML (GUI)

Yeah, that sounds like a major drawback. Perhaps option **5** is the best then? Just add a `r` version of each `beatloop_X_activate` and `beatlooproll_X_activate`?

acolombier:
I have added the the `r` CO now for `beatloop` and `beatlooproll`. Let me know if there is something more to, otherwise will prepare the manual PR.

ronso0 (changes requested):

ronso0:
I'm not happy with the LateNight icons, the lock is too small and has cutouts whic makes it hard to read IMO.
Will check if I can prepare a better version sooish.

acolombier:
I have updated the LateMoon - hopefully it looks better now, although I'm sure it can always been improved

src/engine/controls/loopingcontrol.cpp
```diff
@@ -1967,3 +2058,42 @@ void BeatLoopingControl::slotToggle(double value) {
         emit activateBeatLoop(this);
     }
 }
+
+void BeatLoopingControl::slotReverseActivate(double value) {
+    bool isReverseAnchor = m_pCOLoopAnchor->toBool();
+    if (!isReverseAnchor) {
+        // If the reverse mode is not enabled, we do so temporarily
+        m_pCOLoopAnchor->set(1.0);
```
daschuer, line 2066: This looks a bit hacky, because m_pCOLoopAnchor may have the side effect that the GUI and controller buttons are updated. Can you pass it as bool parameter?

daschuer (changes requested):

src/skin/legacy/tooltips.cpp
```diff
@@ -778,6 +778,11 @@ void Tooltips::addStandardTooltips() {
             << QString("%1: %2").arg(rightClick, tr("Temporarily enable a rolling loop over the set number of beats."))
             << tr("Playback will resume where the track would have been if it had not entered the loop.");
 
+    add("loop_anchor")
+            << tr("Beatloop")
```
daschuer, line 782:
```suggestion
            << tr("Beatloop Anchor")
```

src/skin/legacy/tooltips.cpp
```diff
@@ -778,6 +778,11 @@ void Tooltips::addStandardTooltips() {
...
+            << tr("Define whether the loop is defined and adjusted from its "
```
daschuer, line 783:
```suggestion
            << tr("Define whether the loop is created and adjusted from its "
```

src/engine/controls/loopingcontrol.cpp
acolombier: Added the enum - does it work for you?

daschuer: That's even better. Thank you.

daschuer (approved):

daschuer:
@ronso0 Your review is pending. Are all your topics fixed?

ronso0:
I'm taking a look at the LateNight visuals and will send a commit with some fixes & tweaks soonish.

acolombier:
Are you happy to do it as a subsequent PR or do you think it should come as part of that one?

ronso0:
Tbh I'd like to finish it now, just to get it off the table ; )

Please check these commits:
81b339c416
91899b2ded

I desaturated the lock icons, because IMO they stood out too much in the loop section.

Looks like this now
image
image

ronso0:
> @ronso0 Your review is pending. Are all your topics fixed?

Yes, thanks.

FWIW I'd prefer if we stick to the rule that the reviewer closes his review convos if the issues/request have been addressed. That allows to quickly check for open issues.
If reviews are clsoed by other, and then references later on (like here) it's tedious to revisit convos closed by others just get back on track and check etc.

ronso0 (dismissed):

daschuer:
Thank you, I am curious if this will change my habit with beatliops.

PR #12760, opened 2024-02-08
Use Taglib1 package on Linux if found 

daschuer:
This allows to use the https://aur.archlinux.org/packages/taglib1 package on Linux Arch. 
This is a workaround for https://github.com/mixxxdj/mixxx/issues/12790

### DIFF
--- a/cmake/modules/FindTagLib.cmake
+++ b/cmake/modules/FindTagLib.cmake
@@ -47,6 +47,8 @@ include(IsStaticLibrary)
 
 find_package(PkgConfig QUIET)
 if(PkgConfig_FOUND)
+  # priorize the taglib1 package introduced in https://aur.archlinux.org/packages/taglib1
+  set(ENV{PKG_CONFIG_PATH} "/usr/lib/taglib1/lib/pkgconfig/")
   pkg_check_modules(PC_TagLib QUIET taglib)
 endif()
 
--- a/cmake/modules/FindTagLib.cmake
+++ b/cmake/modules/FindTagLib.cmake
@@ -52,14 +52,14 @@ endif()
 
 find_path(TagLib_INCLUDE_DIR
   NAMES tag.h
-  PATHS ${PC_TagLib_INCLUDE_DIRS}
+  HINTS ${PC_TagLib_INCLUDE_DIRS}
   PATH_SUFFIXES taglib
   DOC "TagLib include directory")
 mark_as_advanced(TagLib_INCLUDE_DIR)
 
 find_library(TagLib_LIBRARY
   NAMES tag
-  PATHS ${PC_TagLib_LIBRARY_DIRS}
+  HINTS ${PC_TagLib_LIBRARY_DIRS}
   DOC "TagLib library"
 )
 mark_as_advanced(TagLib_LIBRARY)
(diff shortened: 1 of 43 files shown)


mxmilkiib:
That builds 2.4 for me!

(Though I would prefer to be using main.)

vlada-dudr:
Wouldn't be better to parametric? Like `-DTAGLIB_PKGCONFIG_PATH=/wherever/it/is` or we don't care because it is temporary solution?

toszlanyi:
> That builds 2.4 for me!

also confirming this change builds nicely now on my EndeavourOS (Arch) with taglib2 and taglib1 installed side by side. Thank you @daschuer

Holzhaus:
I think should be dealt with by the package/AUR maintainer, not in our code. `taglib1` is not even an official Arch package, it's a user-maintained package from the AUR. I'd rather not add workaround for that.

cmake/modules/FindTagLib.cmake
```diff
@@ -47,19 +47,23 @@ include(IsStaticLibrary)
 
 find_package(PkgConfig QUIET)
 if(PkgConfig_FOUND)
+  if(UNIX AND NOT APPLE)
+    # priorize the taglib1 package introduced in https://aur.archlinux.org/packages/taglib1
+    set(ENV{PKG_CONFIG_PATH} "/usr/lib/taglib1/lib/pkgconfig/:$ENV{PKG_CONFIG_PATH}")
```
daschuer, line 52: @Holzhaus Only these two lines are ARCH specific. The rest is required to have two version of taglib installed. We may remove it but on the other hand it is convenient for our contributors and does not really hurt.

daschuer:
This is the situation: https://repology.org/project/taglib/versions

daschuer:
It looks like Debian will allow using a taglib1 and taglib2 package but the development files are conflicting right now.

daschuer:
This is a workaround for https://github.com/mixxxdj/mixxx/issues/12790

ywwg:
I am ok with this kind of hack in order to get a release out -- we can do a "proper" fix later if possible,

ywwg:
I do not know anything about this issue, @Holzhaus are you ok for us to merge this in or is your objection blocking?

Holzhaus:
No, I'm not blocking this but I still think this is unnecessary and the `taglib/` prefix removal from the includes makes the includes harder to understand what library it belongs to.

A patch like that can easily be added to the Mixxx AUR package (which I maintain).

For the `mixxx` package in the Arch package repository, the package maintainer could apply it as well if deemed necessary. However, the `mixxx` package for 2.4 on Arch Linux will definitely not depend on the `taglib1` AUR package. Either the package maintainer creates a new distro package for legacy TagLib and adds that to the distro repositories (and we don't know if these patches will even be necessary for that package), or `mixxx` is removed from the Arch repos because they don't want to add such a package.

That's why I don't think it makes sense to hack workarounds for specific distros here. That's something package maintainers are responsible for.

daschuer:
Thank you @Holzhaus 

The removal of the "taglib/"  include prefix is required because of this upstream PR :https://github.com/taglib/taglib/commit/bd4c9cbf9783733041d7a2d2e97a2c968018aa97
We don't know the include folder package maintainers will choose. Now, we just follow where pkgconf points us. Before we ended up in a mixture of headers of both trees.

ywwg:
merging mostly just to get the release moving.  We can come back to the issue later and maybe move per-distro patches to their respective packaging repos

Issue #12770, opened 2024-02-09
Strange text in log-files : "FUCK YOU MOTHERFUCKERS\nfuck you" , 1


Eve00000:
### Bug Description

Reading the logfiles I found following text

Warning [Main] QString::arg: Argument missing: "FUCK YOU MOTHERFUCKERS\nfuck you" , 1

on different timestamps, in total (10 logfiles together) +/- 3000 times.
Can Somebody explain please?

### Version

2.5-alpha-229-g7b3e48544a (main)

### OS

Win 10

JoergAtGithub:
It seems somebody spamed our translation files with this: https://github.com/mixxxdj/mixxx/blob/b8adc2181d241caaae5dac96eed85636a1c8fef9/res/translations/mixxx_nl.ts#L742

Eve00000:
Can I enter a translation, I am a native dutch speaker.

JoergAtGithub:
> Can I enter a translation, I am a native dutch speaker.

In general yes: https://github.com/mixxxdj/mixxx/wiki/Internationalization

But since this is just vandalsim overwriting the original translation strings, it would be better, if we find a way to just restore the original strings.

Eve00000:
When I enter the dutch translation, how do I have to save my work?

Eve00000:
> > Can I enter a translation, I am a native dutch speaker.
> 
> In general yes: https://github.com/mixxxdj/mixxx/wiki/Internationalization
> 
> But since this is just vandalsim overwriting the original translation strings, it would be better, if we find a way to just restore the original strings.

I imagine you have a tracelog so you can block the one who did this....

ywwg:
thank you! It looks like the spammer only changed strings that were "unfinished" in the original translation files so we don't have proper text to revert to.  We would greatly appreciate your contribution.  

https://github.com/mixxxdj/mixxx/wiki/Internationalization
https://explore.transifex.com/mixxx-dj-software/mixxxdj/

ywwg:
Please do the work in transifex, that's where the spam came from.  We'll lock down that system. Our github tree is secure.

Eve00000:
I followed the Internationalization - manual, I signed up to transifex. I joined the translation team.
Fo I have to wait for authorisation? I can't see where to enter a translation.

ronso0:
oh, smells like Deejaygirl trouble :/

mxmilkiib:
[...]

PR #12771, opened 2024-02-09
Fix translation hijacking swearwords, replace them with google transl…

luisfcorreia:
…ated equivalents

Fixes one file 

Related to #12770 12770


ywwg:
instead of google translate, can we do a simple revert of the files back to the original, even though those are unfinished?  the spam affects mixxx_nl_BE.ts, mixxx_nl_NL.ts, and mixxx_nl.ts. We have a native speaker offering to do human translations

luisfcorreia:
Sure,
a human is always a better choice for these :)

PR #12772, opened 2024-02-09
Revert translations from 060b0e95d888ff2364a211156282f94ed0c11415

luisfcorreia:
Hopefully, sanity is restored.

No manual verification was done, QM files were not changed.

### DIFF
--- a/res/translations/mixxx_nl.ts
+++ b/res/translations/mixxx_nl.ts
@@ -11,7 +11,7 @@
 <context>
     <name>AnalysisFeature</name>
     <message>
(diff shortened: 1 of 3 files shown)


ywwg (changes requested):
A few notes:
* let's please target the 2.4 branch since that's the more urgent place that needs fixing
* please don't include the revert of the previous change, better to squash down the commits so the patch is clean.  (In general we usually keep a full commit history, but there's no real gain in keeping a commit that is totally reverted)
* when pushing to your own repo, you'll probably want to push to a named branch that describes your change, or else it will get confusing.

PR #12779, opened 2024-02-10
Drop lodash dependency in ComponentJS

christophehenry:
This removes the need to source `lodash.mixxx.js` when sourcing `midi-components-0.0.js`

`Object.deepMerge()` should work lik this:

```js
target = {
    a: "a",
    c: "c1",
    d: [{a: "a"}, 1],
    e: {},
    f: "f",
    g: {h: "", i: {j: "j"}}
}

source = {
    b: "b",
    c: "c2",
    d: [{b: "b"}, 2],
    e: new Date(),
    f: undefined,
    g: {i: {k: "k"}}
}

Object.deepMerge(target, source);
// Should produce
{
    a: "a",
    b: "b",
    c: "c2",
    d: [{a: "a", b: "b"}, 2],
    e: new Date(),
    f: "f",
    g: {h: "", i: {j: "j", k: "k"}}
}
```

### DIFF
--- a/res/controllers/common-controller-scripts.js
+++ b/res/controllers/common-controller-scripts.js
@@ -48,19 +48,27 @@ var printObject = function(obj, maxdepth) {
 
 
 var stringifyObject = function(obj, maxdepth, checked, prefix) {
-    if (!maxdepth) { maxdepth = 2; }
+    if (!maxdepth) {
+        maxdepth = 2;
+    }
     try {
         return JSON.stringify(obj, null, maxdepth);
     } catch (e) {
-        if (!checked) { checked = []; }
-        if (!prefix) { prefix = ""; }
+        if (!checked) {
+            checked = [];
+        }
+        if (!prefix) {
+            prefix = "";
+        }
         if (maxdepth > 0 && typeof obj === "object" && obj !== null &&
             Object.getPrototypeOf(obj) !== "" && !arrayContains(checked, obj)) {
             checked.push(obj);
             let output = "{\n";
             for (const property in obj) {
                 const value = obj[property];
-                if (typeof value === "function") { continue; }
+                if (typeof value === "function") {
+                    continue;
+                }
                 output += prefix + property + ": "
                     + stringifyObject(value, maxdepth - 1, checked, prefix + "  ")
                     + "\n";
@@ -74,7 +82,9 @@ var stringifyObject = function(obj, maxdepth, checked, prefix) {
 
 var arrayContains = function(array, elem) {
     for (let i = 0; i < array.length; i++) {
-        if (array[i] === elem) { return true; }
+        if (array[i] === elem) {
+            return true;
+        }
     }
     return false;
 };
@@ -97,7 +107,9 @@ var msecondstominutes = function(msecs) {
     const secs = (msecs / 1000) | 0;
     msecs %= 1000;
     msecs = Math.round(msecs * 100 / 1000);
-    if (msecs === 100) { msecs = 99; }
+    if (msecs === 100) {
+        msecs = 99;
+    }
 
     return (m < 10 ? "0" + m : m)
         + ":"
@@ -127,6 +139,58 @@ var colorCodeToObject = function(colorCode) {
 /* eslint @typescript-eslint/no-empty-function: "off" */
 var script = function() {
 };
+
+/**
+ * Discriminates whether an object was created using the `{}` synthax.
+ *
+ * Returns true when was an object was created using the `{}` synthax.
+ * False if the object is an instance of a class like Date or Proxy or an Array.
+ *
+ * isSimpleObject({}) // true
+ * isSimpleObject(null) // false
+ * isSimpleObject(undefined) // false
+ * isSimpleObject(new Date) // false
+ * isSimpleObject(new (class {})()) // false
+ * @param {any} obj Object to test
+ * @returns {boolean} true if obj was created using the `{}` or `new Object()` synthax, false otherwise
+ */
+const isSimpleObject = function(obj) {
+    return obj !== null && typeof obj === "object" && obj.constructor.name === "Object";
+};
+
+script.isSimpleObject = isSimpleObject;
+
+/**
+ * Deeply merges 2 objects (Arrays and Objects only, not Map for instance).
+ * @param target {object | Array} Object to merge source into
+ * @param source {object | Array} Object to merge into source
+ */
+const deepMerge = function(target, source) {
+    if (target === source || target === undefined || target === null || source === undefined || source === null) {
+        return;
+    }
+
+    if (Array.isArray(target) && Array.isArray(source)) {
+        const objTarget = target.reduce((acc, val, idx) => Object.assign(acc, {[idx]: val}), {});
+        const objSource = source.reduce((acc, val, idx) => Object.assign(acc, {[idx]: val}), {});
+        deepMerge(objTarget, objSource);
+        target.length = 0;
+        target.push(...Object.values(objTarget));
+    } else if (isSimpleObject(target) && isSimpleObject(source)) {
+        Object.keys(source).forEach(key => {
+            if (
+                Array.isArray(target[key]) && Array.isArray(source[key]) ||
+              isSimpleObject(target[key]) && isSimpleObject(source[key])
+            ) {
+                deepMerge(target[key], source[key]);
+            } else if (source[key] !== undefined && source[key] !== null) {
+                Object.assign(target, {[key]: source[key]});
+            }
+        });
+    }
+};
+
+script.deepMerge = deepMerge;
 
 // ----------------- Mapping constants ---------------------
 
@@ -243,7 +307,7 @@ script.triggerControl = function(group, control, delay) {
         delay = 200;
     }
     engine.setValue(group, control, 1);
-    engine.beginTimer(delay, () => engine.setValue(group, control, 0) , true);
+    engine.beginTimer(delay, () => engine.setValue(group, control, 0), true);
 };
 
 /* -------- ------------------------------------------------------
@@ -399,7 +463,9 @@ script.posMod = function(a, m) {
    Output:  none
    -------- ------------------------------------------------------ */
 script.loopMove = function(group, direction, numberOfBeats) {
-    if (!numberOfBeats || numberOfBeats === 0) { numberOfBeats = 0.5; }
+    if (!numberOfBeats || numberOfBeats === 0) {
+        numberOfBeats = 0.5;
+    }
 
     if (direction < 0) {
         engine.setValue(group, "loop_move", -numberOfBeats);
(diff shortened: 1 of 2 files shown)


res/controllers/midi-components-0.0.js
```diff
@@ -656,7 +656,7 @@
                 });
             }
 
-            _.merge(this, newLayer);
+            Object.assign(this, newLayer);
```
christophehenry, line 659: I don't think there was ever really a need for a deep merge here, so I just replaced by `Object.assign()`.

JoergAtGithub:
Less dependencies are always good. And it makes the code easier readable and more typesafe.
@Swiftb0y What do you think about this change?

res/controllers/midi-components-0.0.js
Swiftb0y: I'm a little skeptical about this. It is a breaking change, so let's double check this is actually true for at least the mappings we include.

christophehenry: I already checked the current usages but a double check is always a good idea.

Swiftb0y: yeah, so I also double checked and while it shouldn't cause any issues in-tree now it I'm still not sure if its a good decision from an API design standpoint. ComponentJS is essentially based on the ideas of QML, so I'd like to retain the same behavior as when you create QML components. It's essentially described as [grouped properties](https://doc.qt.io/qt-6/qtqml-syntax-objectattributes.html#grouped-properties) and when you transfer the behavior to JS, it does a deep object merge rather than an `Object.assign`. So I'd like to retain the similarity there and keep using the behavior of `_.merge`. I'm still in favor of removing lodash, but then we'd need an implementation of our own object merge. wdyt?

christophehenry: Ok so I initially wrote a `deepAssign`implementation before checking if deep merge was necessary and deciding to drop it. I can bring back that implementation.

Swiftb0y: Yes, please do so.

res/controllers/common-controller-scripts.js
```diff
@@ -28,6 +28,44 @@ String.prototype.toInt = function() {
...
+        return;
+    }
+
+    Object.keys(source).forEach(k => {
+        if (Array.isArray(source[k]) && Array.isArray(target[k])) {
+            source[k].push(...target[k]);
+        } else if (Object.isSimpleObject(source[k]) && Object.isSimpleObject(target[k])) {
+            Object.deepMerge(source[k], target[k]);
+        }
+        Object.assign(source, target);
```
Swiftb0y, line 65: I'm not sure how critical it is, but `_.merge({a:1},{a:undefined})` and `Object.deepMerge({a:1},{a:undefined})` will differ in behavior due to the fact that `Object.assign` will overwrite with `undefined` and `_.merge` will not.

res/controllers/common-controller-scripts.js
```diff
@@ -28,6 +28,44 @@ String.prototype.toInt = function() {
...
+Object.isSimpleObject = function(obj) {
+    return obj !== null && typeof obj === "object" && obj.constructor.name === "Object";
+};
+
+/**
+ * Deeply merges 2 objects.
+ * @param source {object | Array} Object to merge target into
+ * @param target {object | Array} Object to merge into target
+ */
+Object.deepMerge = function(source, target) {
```
Swiftb0y, line 54: I think the naming should be switched around. The keys of the second argument are merged into the first one, not the other way around. the jsdoc comment is a little nonsensical too IMO. The usage below in the componentJS ctor is then also switched around.

res/controllers/common-controller-scripts.js
```diff
@@ -28,6 +28,44 @@ String.prototype.toInt = function() {
...
+ * Object.isSimpleObject(null) // false
+ * Object.isSimpleObject(undefined) // false
+ * Object.isSimpleObject(new Date) // false
+ * Object.isSimpleObject(new (class {})()) // false
+ * @param {any} obj Object to test
+ * @returns {boolean} true if obj was created using the `{}` synthax, false otherwise
+ */
+Object.isSimpleObject = function(obj) {
+    return obj !== null && typeof obj === "object" && obj.constructor.name === "Object";
+};
```
Swiftb0y, line 47: fyi `Object.isSimpleObject(new Object()) === true`, probably doesn't matter in terms of behavior, but the comment/naming is a little misleading.

res/controllers/common-controller-scripts.js
```diff
@@ -28,6 +28,44 @@ String.prototype.toInt = function() {
...
+
+/**
+ * Deeply merges 2 objects.
```
Swiftb0y, line 50:
```suggestion
 * Deeply merges 2 objects (Arrays and Objects only, not Map for instance).
```

res/controllers/midi-components-0.0.js
christophehenry: Done. So the code I write is far simpler that Lodash's but should suffice in most cases. Here is a test sample:

```js
target = {
    a: "a",
    b: [1, 2],
    c: {
        d: {
            e: [1, 2]
        },
    },
    g: new Date()

}

source = {
    b: [2, 3, new Date()],
    c: {
        d: {
            f: new Date()
        }
    },
    g: {test: function(){}},
    h: function test() {}
}

Object.deepMerge(target, source);
// Should produce:
 {
    a: "a",
    b: [1, 2, 2, 3, new Date()],
    c: {
        d: {
            e: [1, 2],
            f: new Date(),
        },
    },
    g: {test: function(){}},
    h: function test() {}
}
```

res/controllers/common-controller-scripts.js
christophehenry: Right. This is fixed now.

res/controllers/common-controller-scripts.js
christophehenry: True.

christophehenry:
Sorry, I first pushed a code that wasn't polished. It should be ok now.

Swiftb0y:
No worries ;)

res/controllers/common-controller-scripts.js
```diff
@@ -28,6 +28,52 @@ String.prototype.toInt = function() {
...
+ * @param target {object | Array} Object to merge source into
+ * @param source {object | Array} Object to merge into source
+ */
+Object.deepMerge = function(target, source) {
+    if (target === source || target === undefined) {
+        return;
+    }
+
+    if (Array.isArray(target) && Array.isArray(source)) {
+        target.push(...source);
```
Swiftb0y, line 60: this'll only concatenate the array, right? not merge them. I think you'd have to zip em together and then call deepMerge on each pair of elements.

christophehenry: This should be fixed with the last commit. Are there tests for JS anywhere?

Swiftb0y: not sure. some basics ones somewhere burrowed in the controller engine tests somewhere, but idk. I'd be nice to have [Qt Quick test](https://doc.qt.io/qt-6/qtquicktest-index.html) integration into our current test suite to do proper testing.

christophehenry:
Is there anything more to fix, here?

Swiftb0y:
not really. I just don't have the time for a proper review atm. Can you take over @JoergAtGithub?

res/controllers/common-controller-scripts.js
JoergAtGithub: > This should be fixed with the last commit. Are there tests for JS anywhere?

You can implement Mixxx unit test for JavaScript, like I did in https://github.com/mixxxdj/mixxx/pull/12437

christophehenry:
Hello there! Any updates here?

res/controllers/common-controller-scripts.js
```diff
@@ -28,6 +28,54 @@ String.prototype.toInt = function() {
...
+    if (target === source || target === undefined || target === null || source === undefined || source === null) {
+        return;
+    }
+
+    if (Array.isArray(target) && Array.isArray(source)) {
+        const objTarget = target.reduce((acc, val, idx) => Object.assign(acc, {[idx]: val}), {});
+        const objSource = source.reduce((acc, val, idx) => Object.assign(acc, {[idx]: val}), {});
+        Object.deepMerge(objTarget, objSource);
+        target.length = 0;
+        target.push(...Object.values(objTarget));
```
Swiftb0y, line 64: this will probably not retain indicies when its been given an partially empty array (eg `let arr = new Array(2); arr[1] = 1`. But it think its very much an edge case we can ignore.

Swiftb0y (commented):
minor nitpicks, sorry for forgetting about this.

res/controllers/common-controller-scripts.js
christophehenry: I don't see why. `new Array(2)` should fill the array with `undefined's.

res/controllers/common-controller-scripts.js
```diff
@@ -28,6 +28,54 @@ String.prototype.toInt = function() {
...
+ * Object.isSimpleObject(null) // false
+ * Object.isSimpleObject(undefined) // false
+ * Object.isSimpleObject(new Date) // false
+ * Object.isSimpleObject(new (class {})()) // false
+ * @param {any} obj Object to test
+ * @returns {boolean} true if obj was created using the `{}` or `new Object()` synthax, false otherwise
+ */
+Object.isSimpleObject = function(obj) {
+    return obj !== null && typeof obj === "object" && obj.constructor.name === "Object";
+};
```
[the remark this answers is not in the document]
christophehenry: `Object.deepMerge` too?

res/controllers/common-controller-scripts.js
christophehenry: Thank you. TIL.

christophehenry:
Done.

Swiftb0y:
Not sure. I'm don't really know why its failing either. can you try adding it to the `script` namespace? also can you rebase this PR on `main`? I'd like to give it some beta-time before it ships to all users.

christophehenry:
What's the branch to rebase to?

Swiftb0y (approved):
LGTM. Thank you. Sorry for having this neglected. Looking forward to your future PRs. @JoergAtGithub Knowing that this will conflict with #11625, how do you want to proceed?

res/controllers/common-controller-scripts.js
```diff
@@ -128,6 +140,56 @@ var colorCodeToObject = function(colorCode) {
...
+ * isSimpleObject(null) // false
+ * isSimpleObject(undefined) // false
+ * isSimpleObject(new Date) // false
+ * isSimpleObject(new (class {})()) // false
+ * @param {any} obj Object to test
+ * @returns {boolean} true if obj was created using the `{}` or `new Object()` synthax, false otherwise
+ */
+const isSimpleObject = function(obj) {
+    return obj !== null && typeof obj === "object" && obj.constructor.name === "Object";
+};
```
JoergAtGithub, line 159: Is there a reason, why this is not part of `script`?

christophehenry: No. Not really.

JoergAtGithub: #11625 make script a frozen object. Would like to merge this PR first and than unify it in #11625 .

christophehenry: Ok, I amended the commit to add it to `script`.

Swiftb0y (approved):
LGTM. waiting for @JoergAtGithub

JoergAtGithub:
Thank you!

Swiftb0y:
Thank you @christophehenry and @JoergAtGithub

PR #12780, opened 2024-02-10
Introduce a previousTimerId parameter to ControllerScriptInterfaceLegacy::beginTimer()

christophehenry:
This parameter is used to specifiy a timer ID to cancel before starting a new one. This is especially useful for repeating timers that are not assured to end there execution before the next occurence is executed.

### DIFF
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -167,10 +167,14 @@ declare namespace engine {
      *                   function() { print("Executed Timer") }
      * @param oneShot If true the function is only once,
      *                if false the function is executed repeatedly  [default = false]
+     * @param previousTimerId The ID a previous timer to cancel prior to start this one.
+     *                        If this is parameter is defined and a timer exists with
+     *                        this ID, it will be canceled before scheduling the new
+     *                        script code. [default = undefined]
      * @returns timerId which is needed to stop a timer.
      *          In case of an error, 0 is returned.
      */
-    function beginTimer(interval: number, scriptCode: () => any, oneShot?: boolean): TimerID;
+    function beginTimer(interval: number, scriptCode: () => any, oneShot?: boolean, previousTimerId?: number): TimerID;
 
     /**
      * Stops the specified timer
--- a/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
@@ -459,7 +459,11 @@ void ControllerScriptInterfaceLegacy::log(const QString& message) {
     qCDebug(m_logger) << message;
 }
 int ControllerScriptInterfaceLegacy::beginTimer(
-        int intervalMillis, QJSValue timerCallback, bool oneShot) {
+        int intervalMillis, QJSValue timerCallback, bool oneShot, const QJSValue& previousTimerId) {
+    if (previousTimerId.isNumber() && m_timers.contains(previousTimerId.toInt())) {
+        this->stopTimer(previousTimerId.toInt());
+    }
+
     if (timerCallback.isString()) {
         m_pScriptEngineLegacy->logOrThrowError(
                 QStringLiteral("passed a string to `engine.beginTimer`, please "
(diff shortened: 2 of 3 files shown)


JoergAtGithub:
Why not just calling stopTimer in your mapping script, before calling beginTimer? Where do you see the benefit?

christophehenry:
It saves havong to write a wrapper around `beginTimer()`. I agree it's not a killer feature. But it's convenient and easy to implement. Seemed like a nice place to start getting familiar with the code base.

Swiftb0y:
I think we should be careful when introducing new API surface such as this one... whatever we add, there is currently no way to soft-deprecate or remove features, so every feature added needs to be well thought out. I'm worried that if we need change beginTimer sometime in the future for a good reason, it will be much more difficult because of the larger API surface. As such I don't think we should add this feature when the only reason to have it is added convenience. Does that make sense? 

If you don't want to write a wrapper in each mapping (understandable, DRY-ness is very important), then lets add wrapper to common-controller-scripts.js and make it universally available. JS APIs don't need to be as stable as the C++ APIs, so I'm okay with adding a change there. Wdyt?

christophehenry:
Sound good to me. I just have one tiny issue with that solution: the current `stopTimer` implementation [prints a warning when the timer id is unknown] (https://github.com/mixxxdj/mixxx/blob/601bd562455215d941b309189d0dd23b689fe84a/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp#L512) and, to my knowledge, there is no way to check that a timer exists, which pollute the logs in this kind of use case. So I'd be in favor of dropping this line or turn it into a qCDebug.

Swiftb0y:
well it's an indicator for a bug. When you stop a timerID that is unknown something clearly happened with the timer that should not have happened (eg you creating an ID out of thin air, or scrambled on you had accidentally). Do you have an concrete example of when the log is wrong and undesirable but the code correct?

JoergAtGithub:
In my opinion 
```
if (previousTimerId) {
   engine.stopTimer(previousTimerID); 
}
newTimerID = engine.startTimer(<scriptcode>);
```
is easier understandable code, than
```
newTimerID = engine.startTimer(
   <scriptcode>,
    true,
    previousTimerID);
```
To understand what the later code does, a user would need to read the documentation.

Swiftb0y:
fair, but the kinda already also need to do that to understand the purpose of the trailing oneshot bool parameter...

JoergAtGithub:
> fair, but the kinda already also need to do that to understand the purpose of the trailing oneshot bool parameter...

Which is optional and does not occur in the first case (except you have the one-shot timer of cause).

christophehenry:
> Which is optional and does not occur in the first case (except you have the one-shot timer of cause).

The `previousTimerID` argument is also optionnal.

> When you stop a timerID that is unknown something clearly happened with the ID that should not have happened

Or maybe just the previous task had time to complete and is not listed anymore. In this case, try to stop the timer is not a bug. Anyway, I don't think there's a need to log this except during the development of the controller. Then `qCDebug` seem enough to me.

Swiftb0y:
> Or maybe just the previous task had time to complete and is not listed anymore. In this case, try to stop the timer is not a bug.

Well, then you're mapping is using a resource (in this case the timer ID) after it has been destroyed (timer stopped). That's a resource leak. Not as critical as if it was pointers to memory, but still a leak, thus still a bug and worth a warning (since your mapping is doing things it shouldn't, in your example case, unnecessarily calling stopTimer). 

I recognize that the timer API is bad (its based on JS's ancient setTimout/Interval API as well as on `QObject::startTimer`) and improvements are very welcome, but I think improvements would have to be more fundamental (getting rid of exposing a user-manipulable ID, instead opting for more of a QTimer-like approach).

christophehenry:
Hmm… Since Mixxx now supports ES6+, why not design a new API that returns a Promise instead?

Swiftb0y:
promises probably don't make much sense, they can't be easily used for reoccurring events  and more importantly their usage is incredibly clunky in the version of JS we offer because we do not have async/await support. I think a QTimer-like design would probably work best. Probably in a reduced form similar to the [Qml Timer](https://doc.qt.io/qt-6/qml-qtqml-timer.html)

christophehenry:
I'm closing this then. Would somebody like to write a spec for a new timer API?I may work on this.

PR #12767, opened 2024-02-09
Fix deprecated hotcue status control in midi-components.js

christophehenry:
This fixes the `ControlObject accessed via deprecated key "[Channel1]" "hotcue_x_enabled" - use "[Channel1]" "hotcue_x_status" instead` warning.

PR #2868, opened 2020-06-14
add support for loading controller scripts as JS modules

Be-ing:
This bypasses the old code path used for the `scriptfiles`/`file` XML elements. Instead, the root JS module file is specified by a new XML element:
```
<controller>
    <module>MyController.mjs</module>
</controller>
```

MyController.mjs can use the standard JS `import` keyword to import other modules.

The module must export 3 functions:
  * handleInput(data, timestamp)
  * init()
  * shutdown()

All incoming data is passed to `receiveData`. Functions defined in the JS module cannot be used in the legacy XML system and this will not be supported.

The standard JS "import" keyword only works if the script is loaded with QJSEngine::importModule. JS modules cannot be used with the legacy system of reading JS files into a QString and calling QJSEngine::evaluate.

This makes Qt 5.12 a strict requirement to build Mixxx.

TODO:
  * [x] ~~cleanup redundancies with legacy code~~ The legacy code is a mess. I will not touch it.
  * [x] automatic reloading scripts
  * [x] search for XML files in subdirectories and load JS module from path relative to XML file so controllers can have their own directories
  * [x] add MidiDispatcher module

PR #12781, opened 2024-02-10
Registering MIDI Input Handlers From Javascript

christophehenry:
This is a try at implementing the [*Registering MIDI Input Handlers From Javascript*](https://github.com/mixxxdj/mixxx/wiki/registering-midi-input-handlers-from-javascript).

My C++ knowledge is close to zero so beware of the memory managment problems ahead.

The implementation is still very rough as I am discovering the code. Nothing has been tested yet. I'm opening this PR to get review as early as possible. I still have to connect the `MidiInputHandlerController` to the QJSEngine.

I will appreciate any hint on this feature.

### DIFF
--- a/res/controllers/midi-controller-api.d.ts
+++ b/res/controllers/midi-controller-api.d.ts
@@ -1,5 +1,6 @@
-
-/** MidiControllerJSProxy */
+declare interface MidiInputHandlerController {
+    disconnect(): boolean;
+}
 
 declare namespace midi {
 
@@ -29,4 +30,16 @@ declare namespace midi {
      */
     function sendSysexMsg(dataList: number[], length?: number): void;
 
+    type InputCallback = (channel: string, control: string, value: number, status: number) => void
+
+    /**
+     * Calls the provided callback whenever Mixxx receives a MIDI signal with the first two bytes matching the
+     * provided status and midino argument.
+     * @param status
+     * @param midino
+     * @param callback
+     * @see https://github.com/mixxxdj/mixxx/wiki/midi%20scripting
+     * @see https://github.com/mixxxdj/mixxx/wiki/Midi-Crash-Course
+     */
+    function makeInputHandler(status: number, midino: number, callback: InputCallback): MidiInputHandlerController
 }
--- a/res/controllers/midi-controller-api.d.ts
+++ b/res/controllers/midi-controller-api.d.ts
@@ -30,7 +30,7 @@ declare namespace midi {
      */
     function sendSysexMsg(dataList: number[], length?: number): void;
 
-    type InputCallback = (channel: string, control: string, value: number, status: number) => void
+    type InputCallback = (channel: string, control: string, value: number, status: number, group: string) => void
 
     /**
      * Calls the provided callback whenever Mixxx receives a MIDI signal with the first two bytes matching the
--- a/res/controllers/midi-controller-api.d.ts
+++ b/res/controllers/midi-controller-api.d.ts
@@ -30,7 +30,7 @@ declare namespace midi {
      */
     function sendSysexMsg(dataList: number[], length?: number): void;
 
-    type InputCallback = (channel: string, control: string, value: number, status: number, group: string) => void
+    type InputCallback = (channel: string, control: string, value: number, status: number) => void
 
     /**
      * Calls the provided callback whenever Mixxx receives a MIDI signal with the first two bytes matching the
--- a/src/controllers/controllerinputmappingtablemodel.cpp
+++ b/src/controllers/controllerinputmappingtablemodel.cpp
@@ -197,6 +197,11 @@ QVariant ControllerInputMappingTableModel::data(const QModelIndex& index,
         }
 
         const MidiInputMapping& mapping = m_midiInputMappings.at(row);
+
+        if (std::holds_alternative<QJSValue>(mapping.control)) {
+            return QVariant();
+        }
+
         switch (column) {
             case MIDI_COLUMN_CHANNEL:
                 return MidiUtils::channelFromStatus(mapping.key.status);
@@ -213,9 +218,12 @@ QVariant ControllerInputMappingTableModel::data(const QModelIndex& index,
             case MIDI_COLUMN_ACTION:
                 if (role == Qt::UserRole) {
                     // TODO(rryan): somehow get the delegate display text?
-                    return QVariant(mapping.control.group + QStringLiteral(",") + mapping.control.item);
+                    return QVariant(
+                            std::get<ConfigKey>(mapping.control).group +
+                            QStringLiteral(",") +
+                            std::get<ConfigKey>(mapping.control).item);
                 }
-                return QVariant::fromValue(mapping.control);
+                return QVariant::fromStdVariant(mapping.control);
             case MIDI_COLUMN_COMMENT:
                 return mapping.description;
             default:
@@ -233,6 +241,10 @@ QString ControllerInputMappingTableModel::getDisplayString(const QModelIndex& in
     int row = index.row();
     int column = index.column();
     const MidiInputMapping& mapping = m_midiInputMappings.at(row);
+
+    if (!std::holds_alternative<ConfigKey>(mapping.control)) {
+        return QString();
+    }
 
     switch (column) {
     case MIDI_COLUMN_COMMENT: {
@@ -259,7 +271,7 @@ QString ControllerInputMappingTableModel::getDisplayString(const QModelIndex& in
         // untranslated script control
         return data(index, Qt::UserRole).toString() + QStringLiteral(" ") +
                 del->displayText(
-                        QVariant::fromValue(mapping.control), QLocale());
+                        QVariant::fromStdVariant(mapping.control), QLocale());
     }
     default:
         return QString();
--- a/src/controllers/controllerinputmappingtablemodel.cpp
+++ b/src/controllers/controllerinputmappingtablemodel.cpp
@@ -202,6 +202,12 @@ QVariant ControllerInputMappingTableModel::data(const QModelIndex& index,
             return QVariant();
         }
 
+        const auto* const control = std::get_if<ConfigKey>(&mapping.control);
+
+        VERIFY_OR_DEBUG_ASSERT(control != nullptr) {
+            return QVariant();
+        }
+
         switch (column) {
             case MIDI_COLUMN_CHANNEL:
                 return MidiUtils::channelFromStatus(mapping.key.status);
@@ -218,12 +224,9 @@ QVariant ControllerInputMappingTableModel::data(const QModelIndex& index,
             case MIDI_COLUMN_ACTION:
                 if (role == Qt::UserRole) {
                     // TODO(rryan): somehow get the delegate display text?
-                    return QVariant(
-                            std::get<ConfigKey>(mapping.control).group +
-                            QStringLiteral(",") +
-                            std::get<ConfigKey>(mapping.control).item);
+                    return QVariant(control->group + QStringLiteral(",") + control->item);
                 }
-                return QVariant::fromValue(&std::get<ConfigKey>(mapping.control));
+                return QVariant::fromValue(control);
             case MIDI_COLUMN_COMMENT:
                 return mapping.description;
             default:
--- a/src/controllers/controllerinputmappingtablemodel.cpp
+++ b/src/controllers/controllerinputmappingtablemodel.cpp
@@ -198,7 +198,7 @@ QVariant ControllerInputMappingTableModel::data(const QModelIndex& index,
 
         const MidiInputMapping& mapping = m_midiInputMappings.at(row);
 
-        if (std::holds_alternative<QJSValue>(mapping.control)) {
+        if (std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
             return QVariant();
         }
 
@@ -223,7 +223,7 @@ QVariant ControllerInputMappingTableModel::data(const QModelIndex& index,
                             QStringLiteral(",") +
                             std::get<ConfigKey>(mapping.control).item);
                 }
-                return QVariant::fromStdVariant(mapping.control);
+                return QVariant::fromValue(&std::get<ConfigKey>(mapping.control));
             case MIDI_COLUMN_COMMENT:
                 return mapping.description;
             default:
--- a/src/controllers/midi/legacymidicontrollermappingfilehandler.cpp
+++ b/src/controllers/midi/legacymidicontrollermappingfilehandler.cpp
@@ -271,10 +271,15 @@ QDomElement LegacyMidiControllerMappingFileHandler::makeTextElement(QDomDocument
 
 QDomElement LegacyMidiControllerMappingFileHandler::inputMappingToXML(
         QDomDocument* doc, const MidiInputMapping& mapping) const {
+    if (std::holds_alternative<QJSValue>(mapping.control)) {
+        return QDomElement();
+    }
+
     QDomElement controlNode = doc->createElement("control");
 
-    controlNode.appendChild(makeTextElement(doc, "group", mapping.control.group));
-    controlNode.appendChild(makeTextElement(doc, "key", mapping.control.item));
+    controlNode.appendChild(makeTextElement(
+            doc, "group", std::get<ConfigKey>(mapping.control).group));
+    controlNode.appendChild(makeTextElement(doc, "key", std::get<ConfigKey>(mapping.control).item));
     if (!mapping.description.isEmpty()) {
         controlNode.appendChild(
                 makeTextElement(doc, "description", mapping.description));
--- a/src/controllers/midi/legacymidicontrollermappingfilehandler.cpp
+++ b/src/controllers/midi/legacymidicontrollermappingfilehandler.cpp
@@ -271,7 +271,7 @@ QDomElement LegacyMidiControllerMappingFileHandler::makeTextElement(QDomDocument
 
 QDomElement LegacyMidiControllerMappingFileHandler::inputMappingToXML(
         QDomDocument* doc, const MidiInputMapping& mapping) const {
-    if (std::holds_alternative<QJSValue>(mapping.control)) {
+    if (std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
         return QDomElement();
     }
 
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -210,7 +210,7 @@ void MidiController::learnTemporaryInputMappings(const MidiInputMappings& mappin
                         },
                         [message](const std::shared_ptr<QJSValue>& control) {
                             Q_UNUSED(control);
-                            qDebug() << "Set mapping for" << message << "to lambda";
+                            qDebug() << "Set mapping for" << message << "to anonymous JS function";
                         }},
                 mapping.control);
     }
@@ -275,22 +275,22 @@ void MidiController::receivedShortMessage(unsigned char status,
     if (isLearning()) {
         emit messageReceived(status, control, value);
 
-        auto it = m_temporaryInputMappings.find(mappingKey.key);
-        if (it != m_temporaryInputMappings.end()) {
-            for (; it != m_temporaryInputMappings.end() && it.key() == mappingKey.key; ++it) {
+        auto it = m_temporaryInputMappings.constFind(mappingKey.key);
+        if (it != m_temporaryInputMappings.constEnd()) {
+            for (; it != m_temporaryInputMappings.constEnd() && it.key() == mappingKey.key; ++it) {
                 processInputMapping(it.value(), status, control, value, timestamp);
             }
             return;
         }
     }
 
-    auto it = m_pMapping->getInputMappings().find(mappingKey.key);
-    for (; it != m_pMapping->getInputMappings().end() && it.key() == mappingKey.key; ++it) {
+    auto it = m_pMapping->getInputMappings().constFind(mappingKey.key);
+    for (; it != m_pMapping->getInputMappings().constEnd() && it.key() == mappingKey.key; ++it) {
         processInputMapping(it.value(), status, control, value, timestamp);
     }
 }
 
-void MidiController::processInputMapping(MidiInputMapping& mapping,
+void MidiController::processInputMapping(const MidiInputMapping& mapping,
         unsigned char status,
         unsigned char control,
         unsigned char value,
@@ -305,28 +305,43 @@ void MidiController::processInputMapping(MidiInputMapping& mapping,
             return;
         }
 
-        if (std::holds_alternative<ConfigKey>(mapping.control)) {
-            // Lazily compute the QJSValue
-            QJSValue function = pEngine->wrapFunctionCode(
-                    std::get<ConfigKey>(mapping.control).item, 5);
-            mapping.control = std::make_shared<QJSValue>(function);
-        }
+        return std::visit(
+                MidiUtils::overloaded{
+                        [pEngine, this, channel, status, control, value](
+                                const ConfigKey& target) {
+                            QJSValue function = pEngine->wrapFunctionCode(
+                                    target.item, 5);
+                            const auto args = QJSValueList{
+                                    channel,
+                                    control,
+                                    value,
+                                    status,
+                                    target.group,
+                            };
 
-        VERIFY_OR_DEBUG_ASSERT(std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
-            return;
-        }
+                            if (!pEngine->executeFunction(&function, args)) {
+                                qCWarning(m_logBase) << "MidiController: Invalid script function"
+                                                     << target.item;
+                            }
+                        },
+                        [pEngine, this, channel, status, control, value](
+                                const std::shared_ptr<QJSValue>& target) {
+                            const auto args = QJSValueList{
+                                    channel,
+                                    control,
+                                    value,
+                                    status,
+                            };
 
-        const auto args = QJSValueList{
-                channel,
-                control,
-                value,
-                status,
-        };
-        const auto& fun = std::get<std::shared_ptr<QJSValue>>(mapping.control);
-        if (!pEngine->executeFunction(fun.get(), args)) {
-            qCWarning(m_logBase) << "MidiController: Invalid script function";
-        }
-        return;
+                            if (!pEngine->executeFunction(target.get(), args)) {
+                                qCWarning(m_logBase).nospace()
+                                        << "MidiController: Invalid script "
+                                           "anonymous function with args ["
+                                        << channel << ", " << control << ", "
+                                        << value << ", " << status << "]";
+                            }
+                        }},
+                mapping.control);
     }
 
     VERIFY_OR_DEBUG_ASSERT(std::holds_alternative<ConfigKey>(mapping.control)) {
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -14,9 +14,29 @@
 #include "util/make_const_iterator.h"
 #include "util/math.h"
 
+const QString kMakeInputHandlerError = QStringLiteral(
+        "Invalid timer callback provided to midi.makeInputHandler. "
+        "Please pass a function and make sure that your code contains no syntax errors.");
+
+MidiInputHandleJSProxy::MidiInputHandleJSProxy(
+        const std::shared_ptr<LegacyMidiControllerMapping> mapping,
+        const MidiInputMapping& inputMapping)
+        : m_mapping(mapping), m_inputMapping(inputMapping) {
+}
+
+bool MidiInputHandleJSProxy::disconnect() {
+    // We want to remove only this mapping when disconnecting
+    return m_mapping->removeInputMapping(m_inputMapping.key.key, m_inputMapping);
+}
+
 MidiController::MidiController(const QString& deviceName)
         : Controller(deviceName) {
     setDeviceCategory(tr("MIDI Controller"));
+}
+
+void MidiController::onBeforeEngineShutdown() {
+    Controller::onBeforeEngineShutdown();
+    m_pMapping->removeInputHandlerMappings();
 }
 
 MidiController::~MidiController() {
@@ -181,8 +201,18 @@ void MidiController::learnTemporaryInputMappings(const MidiInputMappings& mappin
                      .rightJustified(2,'0')) :
                 QString("0x%1")
                 .arg(QString::number(mapping.key.status, 16).toUpper());
-        qDebug() << "Set mapping for" << message << "to"
-                 << mapping.control.group << mapping.control.item;
+
+        std::visit(
+                MidiUtils::overloaded{
+                        [message](const ConfigKey& control) {
+                            qDebug() << "Set mapping for" << message << "to"
+                                     << control.group << control.item;
+                        },
+                        [message](const QJSValue& control) {
+                            Q_UNUSED(control);
+                            qDebug() << "Set mapping for" << message << "to lambda";
+                        }},
+                mapping.control);
     }
 }
 
@@ -265,7 +295,7 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
                                          unsigned char control,
                                          unsigned char value,
                                          mixxx::Duration timestamp) {
-    Q_UNUSED(timestamp);
+    Q_UNUSED(timestamp)
     unsigned char channel = MidiUtils::channelFromStatus(status);
     MidiOpCode opCode = MidiUtils::opCodeFromStatus(status);
 
@@ -275,23 +305,53 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
             return;
         }
 
-        QJSValue function = pEngine->wrapFunctionCode(mapping.control.item, 5);
-        const auto args = QJSValueList{
-                channel,
-                control,
-                value,
-                status,
-                mapping.control.group,
-        };
-        if (!pEngine->executeFunction(&function, args)) {
-            qCWarning(m_logBase) << "MidiController: Invalid script function"
-                                 << mapping.control.item;
-        }
+        return std::visit(
+                MidiUtils::overloaded{
+                        [this, pEngine, channel, control, value, status](
+                                const ConfigKey& configKey) {
+                            QJSValue function =
+                                    pEngine->wrapFunctionCode(
+                                            configKey.item, 5);
+                            const auto args = QJSValueList{
+                                    channel,
+                                    control,
+                                    value,
+                                    status,
+                                    configKey.group,
+                            };
+                            if (!pEngine->executeFunction(&function, args)) {
+                                qCWarning(m_logBase)
+                                        << "MidiController: Invalid "
+                                           "script function"
+                                        << configKey.item;
+                            }
+                            return;
+                        },
+                        [this, pEngine, channel, control, value, status](
+                                QJSValue function) {
+                            const auto args = QJSValueList{
+                                    channel,
+                                    control,
+                                    value,
+                                    status,
+                            };
+                            if (!pEngine->executeFunction(&function, args)) {
+                                qCWarning(m_logBase)
+                                        << "MidiController: Invalid "
+                                           "anonymous script function";
+                            }
+                            return;
+                        }},
+                mapping.control);
+    }
+
+    if (!std::holds_alternative<ConfigKey>(mapping.control)) {
         return;
     }
 
     // Only pass values on to valid ControlObjects.
-    ControlObject* pCO = ControlObject::getControl(mapping.control);
+    auto configKey = std::get<ConfigKey>(mapping.control);
+    ControlObject* pCO = ControlObject::getControl(configKey);
     if (pCO == nullptr) {
         return;
     }
@@ -314,7 +374,12 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
         for (auto it = m_fourteen_bit_queued_mappings.constBegin();
                 it != m_fourteen_bit_queued_mappings.constEnd();
                 ++it) {
-            if (it->first.control == mapping.control) {
+            const auto* const fourteen_bit_control = std::get_if<ConfigKey>(&it->first.control);
+            VERIFY_OR_DEBUG_ASSERT(fourteen_bit_control != nullptr) {
+                continue;
+            }
+
+            if (*fourteen_bit_control == configKey) {
                 if ((it->first.options & mapping.options) &
                         (MidiOption::FourteenBitLSB | MidiOption::FourteenBitMSB)) {
                     qCWarning(m_logBase)
@@ -538,4 +603,38 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
     }
     qCWarning(m_logBase) << "MidiController: No script function specified for"
                          << MidiUtils::formatSysexMessage(getName(), data, timestamp);
+}
+
+QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue& scriptCode) {
+    auto pJsEngine = getScriptEngine()->jsEngine();
+    VERIFY_OR_DEBUG_ASSERT(pJsEngine) {
+        return QJSValue();
+    }
+
+    if (!scriptCode.isCallable()) {
+        auto error = kMakeInputHandlerError;
+        if (scriptCode.isError()) {
+            error.append("\n" + scriptCode.toString());
+        }
+        getScriptEngine()->throwJSError(error);
+        return QJSValue();
+    }
+
+    if (status <= 0 || midino <= 0) {
+        auto mStatusError = QStringLiteral(
+                "Invalid status or midino passed to midi.makeInputHandler. "
+                "Please pass a strictly positive integer. status=%1,midino=%2")
+                                    .arg(status, midino);
+
+        getScriptEngine()->throwJSError(mStatusError);
+        return QJSValue();
+    }
+
+    MidiInputMapping inputMapping(
+            MidiKey(status, midino),
+            MidiOption::Script,
+            scriptCode);
+
+    m_pMapping->addInputMapping(inputMapping.key.key, inputMapping);
+    return pJsEngine->newQObject(new MidiInputHandleJSProxy(m_pMapping, inputMapping));
 }
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -208,7 +208,7 @@ void MidiController::learnTemporaryInputMappings(const MidiInputMappings& mappin
                             qDebug() << "Set mapping for" << message << "to"
                                      << control.group << control.item;
                         },
-                        [message](const QJSValue& control) {
+                        [message](const std::shared_ptr<QJSValue>& control) {
                             Q_UNUSED(control);
                             qDebug() << "Set mapping for" << message << "to lambda";
                         }},
@@ -328,14 +328,14 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
                             return;
                         },
                         [this, pEngine, channel, control, value, status](
-                                QJSValue function) {
+                                const std::shared_ptr<QJSValue>& function) {
                             const auto args = QJSValueList{
                                     channel,
                                     control,
                                     value,
                                     status,
                             };
-                            if (!pEngine->executeFunction(&function, args)) {
+                            if (!pEngine->executeFunction(function.get(), args)) {
                                 qCWarning(m_logBase)
                                         << "MidiController: Invalid "
                                            "anonymous script function";
@@ -633,7 +633,7 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
     MidiInputMapping inputMapping(
             MidiKey(status, midino),
             MidiOption::Script,
-            scriptCode);
+            std::make_shared<QJSValue>(scriptCode));
 
     m_pMapping->addInputMapping(inputMapping.key.key, inputMapping);
     return pJsEngine->newQObject(new MidiInputHandleJSProxy(m_pMapping, inputMapping));
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -34,8 +34,8 @@ MidiController::MidiController(const QString& deviceName)
     setDeviceCategory(tr("MIDI Controller"));
 }
 
-void MidiController::onBeforeEngineShutdown() {
-    Controller::onBeforeEngineShutdown();
+void MidiController::slotBeforeEngineShutdown() {
+    Controller::slotBeforeEngineShutdown();
     m_pMapping->removeInputHandlerMappings();
 }
 
@@ -275,26 +275,26 @@ void MidiController::receivedShortMessage(unsigned char status,
     if (isLearning()) {
         emit messageReceived(status, control, value);
 
-        auto it = m_temporaryInputMappings.constFind(mappingKey.key);
-        if (it != m_temporaryInputMappings.constEnd()) {
-            for (; it != m_temporaryInputMappings.constEnd() && it.key() == mappingKey.key; ++it) {
+        auto it = m_temporaryInputMappings.find(mappingKey.key);
+        if (it != m_temporaryInputMappings.end()) {
+            for (; it != m_temporaryInputMappings.end() && it.key() == mappingKey.key; ++it) {
                 processInputMapping(it.value(), status, control, value, timestamp);
             }
             return;
         }
     }
 
-    auto it = m_pMapping->getInputMappings().constFind(mappingKey.key);
-    for (; it != m_pMapping->getInputMappings().constEnd() && it.key() == mappingKey.key; ++it) {
+    auto it = m_pMapping->getInputMappings().find(mappingKey.key);
+    for (; it != m_pMapping->getInputMappings().end() && it.key() == mappingKey.key; ++it) {
         processInputMapping(it.value(), status, control, value, timestamp);
     }
 }
 
-void MidiController::processInputMapping(const MidiInputMapping& mapping,
-                                         unsigned char status,
-                                         unsigned char control,
-                                         unsigned char value,
-                                         mixxx::Duration timestamp) {
+void MidiController::processInputMapping(MidiInputMapping& mapping,
+        unsigned char status,
+        unsigned char control,
+        unsigned char value,
+        mixxx::Duration timestamp) {
     Q_UNUSED(timestamp)
     unsigned char channel = MidiUtils::channelFromStatus(status);
     MidiOpCode opCode = MidiUtils::opCodeFromStatus(status);
@@ -305,47 +305,31 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
             return;
         }
 
-        return std::visit(
-                MidiUtils::overloaded{
-                        [this, pEngine, channel, control, value, status](
-                                const ConfigKey& configKey) {
-                            QJSValue function =
-                                    pEngine->wrapFunctionCode(
-                                            configKey.item, 5);
-                            const auto args = QJSValueList{
-                                    channel,
-                                    control,
-                                    value,
-                                    status,
-                                    configKey.group,
-                            };
-                            if (!pEngine->executeFunction(&function, args)) {
-                                qCWarning(m_logBase)
-                                        << "MidiController: Invalid "
-                                           "script function"
-                                        << configKey.item;
-                            }
-                            return;
-                        },
-                        [this, pEngine, channel, control, value, status](
-                                const std::shared_ptr<QJSValue>& function) {
-                            const auto args = QJSValueList{
-                                    channel,
-                                    control,
-                                    value,
-                                    status,
-                            };
-                            if (!pEngine->executeFunction(function.get(), args)) {
-                                qCWarning(m_logBase)
-                                        << "MidiController: Invalid "
-                                           "anonymous script function";
-                            }
-                            return;
-                        }},
-                mapping.control);
+        if (std::holds_alternative<ConfigKey>(mapping.control)) {
+            // Lazily compute the QJSValue
+            QJSValue function = pEngine->wrapFunctionCode(
+                    std::get<ConfigKey>(mapping.control).item, 5);
+            mapping.control = std::make_shared<QJSValue>(function);
+        }
+
+        VERIFY_OR_DEBUG_ASSERT(std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
+            return;
+        }
+
+        const auto args = QJSValueList{
+                channel,
+                control,
+                value,
+                status,
+        };
+        const auto& fun = std::get<std::shared_ptr<QJSValue>>(mapping.control);
+        if (!pEngine->executeFunction(fun.get(), args)) {
+            qCWarning(m_logBase) << "MidiController: Invalid script function";
+        }
+        return;
     }
 
-    if (!std::holds_alternative<ConfigKey>(mapping.control)) {
+    VERIFY_OR_DEBUG_ASSERT(std::holds_alternative<ConfigKey>(mapping.control)) {
         return;
     }
 
--- a/src/controllers/midi/midicontroller.h
+++ b/src/controllers/midi/midicontroller.h
@@ -1,11 +1,27 @@
 #pragma once
 
+#include <QJSValue>
+#include <utility>
+
 #include "controllers/controller.h"
 #include "controllers/midi/legacymidicontrollermappingfilehandler.h"
 #include "controllers/midi/midimessage.h"
 #include "controllers/softtakeover.h"
 
 class MidiOutputHandler;
+
+class MidiInputHandleJSProxy final : public QObject {
+    Q_OBJECT
+  public:
+    MidiInputHandleJSProxy(
+            const std::shared_ptr<LegacyMidiControllerMapping> mapping,
+            const MidiInputMapping& inputMapping);
+    Q_INVOKABLE bool disconnect();
+
+  protected:
+    std::shared_ptr<LegacyMidiControllerMapping> m_mapping;
+    MidiInputMapping m_inputMapping;
+};
 
 /// MIDI Controller base class
 ///
@@ -51,6 +67,8 @@ class MidiController : public Controller {
         send(data);
     }
 
+    QJSValue makeInputHandler(int status, int midino, const QJSValue& scriptCode);
+
   protected slots:
     virtual void receivedShortMessage(
             unsigned char status,
@@ -60,6 +78,7 @@ class MidiController : public Controller {
     // For receiving System Exclusive messages
     void receive(const QByteArray& data, mixxx::Duration timestamp) override;
     int close() override;
+    void onBeforeEngineShutdown() override;
 
   private slots:
     bool applyMapping() override;
@@ -116,6 +135,10 @@ class MidiControllerJSProxy : public ControllerJSProxy {
 
     Q_INVOKABLE void sendSysexMsg(const QList<int>& data, unsigned int length = 0) {
         m_pMidiController->sendSysexMsg(data, length);
+    }
+
+    Q_INVOKABLE QJSValue makeInputHandler(int status, int midino, const QJSValue& scriptCode) {
+        return m_pMidiController->makeInputHandler(status, midino, scriptCode);
     }
 
   private:
--- a/src/controllers/midi/midicontroller.h
+++ b/src/controllers/midi/midicontroller.h
@@ -89,7 +89,7 @@ class MidiController : public Controller {
 
   private:
     void processInputMapping(
-            MidiInputMapping& mapping,
+            const MidiInputMapping& mapping,
             unsigned char status,
             unsigned char control,
             unsigned char value,
--- a/src/controllers/midi/midicontroller.h
+++ b/src/controllers/midi/midicontroller.h
@@ -78,7 +78,7 @@ class MidiController : public Controller {
     // For receiving System Exclusive messages
     void receive(const QByteArray& data, mixxx::Duration timestamp) override;
     int close() override;
-    void onBeforeEngineShutdown() override;
+    void slotBeforeEngineShutdown() override;
 
   private slots:
     bool applyMapping() override;
@@ -89,7 +89,7 @@ class MidiController : public Controller {
 
   private:
     void processInputMapping(
-            const MidiInputMapping& mapping,
+            MidiInputMapping& mapping,
             unsigned char status,
             unsigned char control,
             unsigned char value,
--- a/src/controllers/midi/midimessage.h
+++ b/src/controllers/midi/midimessage.h
@@ -149,15 +149,14 @@ struct MidiOutput {
 
     double min;
     double max;
-    union
-    {
-        uint32_t    message;
+    union {
+        uint32_t message;
         struct
         {
-            unsigned char    status  : 8;
-            unsigned char    control : 8;
-            unsigned char    on      : 8;
-            unsigned char    off     : 8;
+            unsigned char status : 8;
+            unsigned char control : 8;
+            unsigned char on : 8;
+            unsigned char off : 8;
         };
     };
 };
@@ -170,13 +169,12 @@ struct MidiKey {
         return key == other.key;
     }
 
-    union
-    {
-        uint16_t    key;
+    union {
+        uint16_t key;
         struct
         {
-            unsigned char    status  : 8;
-            unsigned char    control : 8;
+            unsigned char status : 8;
+            unsigned char control : 8;
         };
     };
 };
@@ -195,7 +193,7 @@ struct MidiInputMapping {
 
     MidiInputMapping(MidiKey key,
             MidiOptions options,
-            const std::variant<ConfigKey, QJSValue>& control)
+            const std::variant<ConfigKey, std::shared_ptr<QJSValue>>& control)
             : key(key),
               options(options),
               control(control) {
@@ -203,7 +201,7 @@ struct MidiInputMapping {
 
     MidiInputMapping(MidiKey key,
             MidiOptions options,
-            const std::variant<ConfigKey, QJSValue>& control,
+            const std::variant<ConfigKey, std::shared_ptr<QJSValue>>& control,
             const QString& description)
             : key(key),
               options(options),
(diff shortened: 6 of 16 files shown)


res/controllers/midi-controller-api.d.ts
```diff
@@ -29,4 +38,7 @@ declare namespace midi {
      */
     function sendSysexMsg(dataList: number[], length?: number): void;
 
+    type InputCallback = (channel: string, control: string, value: number, status: number, group: string) => void
```
Swiftb0y, line 41: I think you can get rid of the group string. That only really makes sense for handlers defined in XML. but the entire reason for registering handlers from JS is that you don't need the XML (at least not the handler definitions).

res/controllers/midi-controller-api.d.ts
```diff
@@ -1,5 +1,14 @@
 
 /** MidiControllerJSProxy */
+declare interface ControlDefinition {
+    status: number;
+    midino: number;
+    description?: string;
```
Swiftb0y, line 6: I don't think the description is really needed either, wdyt?

src/controllers/midi/midicontroller.cpp
```diff
@@ -13,6 +13,16 @@
...
 #include "util/math.h"
 
+MidiInputHandlerController::MidiInputHandlerController(
+        const LegacyMidiControllerMapping& mapping,
+        const MidiInputMapping& inputMapping)
+        : m_mapping(mapping), m_inputMapping(inputMapping) {
+}
+
+void MidiInputHandlerController::disconnect() {
+    this->m_mapping.removeInputMapping(this->m_inputMapping.key.key);
```
Swiftb0y, line 23: Not a style we use in mixxx. it's already handled by the `m_` prefix.
```suggestion
    m_mapping.removeInputMapping(m_inputMapping.key.key);
```

src/controllers/midi/midicontroller.cpp
```diff
@@ -537,3 +547,41 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
     qCWarning(m_logBase) << "MidiController: No script function specified for"
                          << MidiUtils::formatSysexMessage(getName(), data, timestamp);
 }
+
+void MidiController::makeInputHandler(const QJSValue& control, const QJSValue& scriptCode) {
+    if (!scriptCode.isCallable()) {
+        QString sErrorMessage(
+                "Invalid timer callback provided to midi.makeInputHandler. "
+                "Please pass a function and make sure that your code contains no syntax errors.");
```
Swiftb0y, line 555: make `const` and put into an anonymous namespace at the top of the file instead.

Swiftb0y (changes requested):
Thank you for this early PR, it is going into the right direction.

Swiftb0y:
Also please sign the [Mixxx Contributor Agreement](https://docs.google.com/forms/d/e/1FAIpQLScC9QG327sjLL0eWftmfGUasxFFLxg6LCXJ2xHDYRzFIRqyiw/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ) and comment here when you have done so. It gives us permission to distribute your contribution under the GPL v2 or later license and the Apple Mac App Store. It is also helpful for us to have contact information for contributors in case we may need it in the future.

res/controllers/midi-controller-api.d.ts
christophehenry: This is the signature of `components.Component.input`. I feel like it may be confusing for the user if this argument is undefined. Don't you think?

Swiftb0y: Good point. I'm not really sure either. It doesn't really make sense for componentJS either and just came from signature of the XML-bound handlers.
I imagine usage of this API to look like this:

```js
midi.makeInputHandler(..., (channel, control, value, status) => {...} );
```

If this was bound to an XML-originating handler, you could just simply supply the group if you need it right then and there. 

```js
midi.makeInputHandler(..., (channel, control, value, status) => {
    someComponent.input(channel,control,value,status, someGroupFromEnclosingScope
} );
```

If thats too much to type, HOF got you covered:


```js
// once somewhere
const makeLegacyWithGroup = (group, handler) => 
    (channel,control,value,status) => handler(channel,control,value,status,group);

// then use it
midi.makeInputHandler(..., makeLegacyWithGroup(someGroupFromEnclosingScope, someComponent.input));
```

As I see it, the only reason the group was introduce in the first place was that there was more information available in the manually created XML than in the corresponding JS handler. Now that all the information is already available in JS (since thats where the handler is created), that loophole is not needed anymore. Do you agree?

christophehenry:
I signed the contributor agreement after opening #12767.

res/controllers/midi-controller-api.d.ts
christophehenry: Sounds good to me. I'll just need to remember about documenting this behavior when this is merged.

Swiftb0y:
> I signed the contributor agreement after opening https://github.com/mixxxdj/mixxx/pull/12767.

Thank you, I just had assumed you hadn't because github still labels you as a first-time contributor. ;)

src/controllers/midi/midicontroller.cpp
```diff
@@ -537,3 +547,41 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
     qCWarning(m_logBase) << "MidiController: No script function specified for"
                          << MidiUtils::formatSysexMessage(getName(), data, timestamp);
 }
+
+void MidiController::makeInputHandler(const QJSValue& control, const QJSValue& scriptCode) {
```
Swiftb0y, line 551: I think I'd prefer to use the values directly instead of packing them in a map because the map only results in extra overhead and it would allow this class to be simpler as we don't need to do the typechecking ourselves. Wdyt?
```suggestion
void MidiController::makeInputHandler(int status, int midino, const QJSValue& scriptCode) {
```

christophehenry:
Oh? May that be because my email address is private on github?

Swiftb0y:
idk, possibly related to the fact that this PR is targetting a different branch to the previous one. Our CLI is not connected to github in any way. I doubled checked and you signed, so we're all good.

christophehenry:
I think I need some help on design, here. To be able to store arbitrary callback in `LegacyMidiControllerMapping`, I would need to change [`ConfigKey::item`](https://github.com/mixxxdj/mixxx/blob/main/src/preferences/configobject.h#L33) from `QString`to `QJSValue`. Calling `toString()` on a `QJSValue that represents a callable just converts to `"function() { [native code] }"` which creates an error. But this is a significant change. A big portion of the code is impacted.

Swiftb0y:
yeah. so the problem stems from the fact that if the midi option is a of type script, the `MidiInputMapping::control` is representing a CO (group and key). The JS engine then expects that key-string to be a "path" so to speak to the input handler (see any JS-based controller mapping for examples).

Changing `Configkey` is not the right solution, also that would make the `group` unnecessary (see also our previous discussion). Instead, you'll probably want to change `MidiInputMapping` to be polymorphic. Since its a value-type, changing `MidiInputMapping::control` from the `ConfigKey` it is currently to a `std::variant<ConfigKey, QJSValue>` and then change the code that uses `MidiInputMapping` to deal with the new type would be the best option. Does that make sense?

christophehenry:
Hmm… Yes. I think I can do that. Thank you for the help.

christophehenry:
So I tried the template solution. For now, I can't test it. A few compilation errors remain that I don't know how to solve with my low C++ knowledge. I would love some help here.

christophehenry:
I wonder if it wouldn't be simpler to just maintain a second hash of mappings.

src/preferences/configobject.h
```diff
@@ -2,24 +2,42 @@
...
+    }
+    QString getFunctionDesc() const {
+        return "";
+    }
+};
+
 // Class for the key for a specific configuration element. A key consists of a
 // group and an item.
-class ConfigKey final {
+class ConfigKey final : public InputHandlerExecutable {
```
Swiftb0y, line 36: `ConfigKey` is almost a vocabulary type used in mixxx. Its very primitive and used everywhere (in the config, the interaction with the audio engine, skins, UI, controllers, etc). In order for those different systems to stay separate, we can't mix anything (in this case InputHandlerExecutable) into it. This feature can/should be implemented entirely without touching `configobject.h`

src/controllers/midi/midimessage.h
```diff
@@ -178,42 +179,52 @@ struct MidiKey {
...
+
     bool operator==(const MidiInputMapping& other) const {
         return key == other.key && options == other.options &&
                 control == other.control && description == other.description;
     }
 
     MidiKey key;
     MidiOptions options;
-    ConfigKey control;
+    T control;
```
Swiftb0y, line 223: I'm afraid we had a misunderstanding with my suggestion here. All I was suggesting was to change this to 
```suggestion
    std::variant<ConfigKey, QJSValue> control;
```

And then dispatch accordingly using `std::visit`. Does that make sense?

src/controllers/midi/midicontroller.h
```diff
@@ -66,7 +67,7 @@ class MidiController : public Controller {
         send(data);
     }
 
-    void makeInputHandler(const QJSValue& control, const QJSValue& scriptCode);
+    void makeInputHandler(const QString& group, int status, int midino, const QJSValue& scriptCode);
```
Swiftb0y, line 70: I still think that the group is unnecessary for mappings declared from JS.

```suggestion
    void makeInputHandler(int status, int midino, const QJSValue& scriptCode);
```

Swiftb0y (commented):
> So I tried the template solution. For now, I can't test it. A few compilation errors remain that I don't know how to solve with my low C++ knowledge. I would love some help here.

Sure. I'm afraid your most recent commit to a bit of a wrong term and we should try again (see review comments). I'm sorry that my suggestions haven't been clear enough. Can you stash your most recent commit away and we'll try again? I think the `std::variant` based solution will be simpler. 

> I wonder if it wouldn't be simpler to just maintain a second hash of mappings.

Yes, certainly.

src/controllers/midi/midimessage.h
christophehenry: Quick question here: what's the difference between [this function](https://github.com/mixxxdj/mixxx/blob/main/src/controllers/midi/midicontroller.cpp#L263) and [this one](https://github.com/mixxxdj/mixxx/blob/main/src/controllers/midi/midicontroller.cpp#L527) ?

Swiftb0y: one is for the standard 3-byte midi, the other one (with the QByteArray) is intended for arbitrary binary data such as sysex.

christophehenry:
Ok, I think I got to something correct. I haven't tested yet but it compiles, which is a promising first step.

res/controllers/midi-controller-api.d.ts
```diff
@@ -29,4 +30,7 @@ declare namespace midi {
      */
     function sendSysexMsg(dataList: number[], length?: number): void;
 
+    type InputCallback = (channel: string, control: string, value: number, status: number) => void
+
+    function makeInputHandler(status: number, midino: number, callback: InputCallback): MidiInputHandlerController
```
christophehenry, line 35: Note to self: I need to document that.

src/controllers/controllerinputmappingtablemodel.cpp
```diff
@@ -197,6 +197,11 @@ QVariant ControllerInputMappingTableModel::data(const QModelIndex& index,
         }
 
         const MidiInputMapping& mapping = m_midiInputMappings.at(row);
+
+        if (holds_alternative<QJSValue>(mapping.control)) {
+            return "";
+        }
+
```
christophehenry, line 204: I don't really understand what that function does so I just opted to not execute it if the control is a lambda.

src/controllers/controllerinputmappingtablemodel.cpp
```diff
@@ -234,6 +242,10 @@ QString ControllerInputMappingTableModel::getDisplayString(const QModelIndex& in
     int column = index.column();
     const MidiInputMapping& mapping = m_midiInputMappings.at(row);
 
+    if (!holds_alternative<ConfigKey>(mapping.control)) {
+        return "";
+    }
```
christophehenry, line 247: Same as [r1486719352](https://github.com/mixxxdj/mixxx/pull/12781/files#r1486719352). Please tell me if I got something wrong.

src/controllers/midi/legacymidicontrollermappingfilehandler.cpp
```diff
@@ -239,8 +239,13 @@ void LegacyMidiControllerMappingFileHandler::addControlsToDocument(
...
                 it != mapping.getInputMappings().constEnd() && it.key() == key;
                 ++it) {
-            QDomElement controlNode = inputMappingToXML(doc, it.value());
-            controls.appendChild(controlNode);
+            std::visit(MidiUtils::overloaded{
+                               [&](ConfigKey control) {
+                                   QDomElement controlNode = inputMappingToXML(doc, it.value());
+                                   controls.appendChild(controlNode);
+                               },
+                               [](QJSValue control) { /* Do nothing */ }},
```
christophehenry, line 247: As I understand here, this function writes an XML file from manually registered control in the Mixxx interface, so it should do nothing on a anonymous function from JS.

Swiftb0y: Yup, sounds about right.

src/controllers/midi/legacymidicontrollermappingfilehandler.cpp
```diff
@@ -273,8 +278,11 @@ QDomElement LegacyMidiControllerMappingFileHandler::inputMappingToXML(
         QDomDocument* doc, const MidiInputMapping& mapping) const {
     QDomElement controlNode = doc->createElement("control");
 
-    controlNode.appendChild(makeTextElement(doc, "group", mapping.control.group));
-    controlNode.appendChild(makeTextElement(doc, "key", mapping.control.item));
+    assert((void("Can't produce XML from a Javascript registered handler"),
+            holds_alternative<QJSValue>(mapping.control)));
```
Swiftb0y, line 282: I think it would make sense to check this at the very start and then either DEBUG_ASSERT (or VERIFY_OR_DEBUG_ASSERT) (see `util/debug.h`) or just plain return without creating an control element (not sure if thats possible with the way this currently implemented).

src/controllers/controllerinputmappingtablemodel.cpp
Swiftb0y, line 247: I prefer the std prefix when applicable.
```suggestion
    if (!std::holds_alternative<ConfigKey>(mapping.control)) {
        return QVariant();
    }
```

src/controllers/midi/midicontroller.cpp
```diff
@@ -284,23 +292,47 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
...
+                                           << "MidiController: Invalid "
+                                              "anonymous script function";
+                               }
+                               return;
+                           }},
+                mapping.control);
+    }
+
+    assert((void("Only ConfigKey mappings allowed beyond this point"),
+            holds_alternative<ConfigKey>(mapping.control)));
```
Swiftb0y, line 331: this is expected to happen, so in that case just check and then return imo.

src/controllers/midi/midimessage.h
```diff
@@ -178,40 +181,69 @@ struct MidiKey {
...
 
-    MidiInputMapping(MidiKey key, MidiOptions options, const ConfigKey& control)
+    MidiInputMapping(MidiKey key,
+            MidiOptions options,
+            const std::variant<ConfigKey, QJSValue>& control)
             : key(key),
               options(options),
-              control(control) {
+              control(control),
+              uuid(QUuid::createUuid()) {
```
Swiftb0y, line 204: is the Uuid really required? Is relying on QJSValue::strictlyEquals too weak for comparision?

Swiftb0y (commented):
Thank you, This is is much more the direction that I initially suggested.

src/controllers/midi/midimessage.h
```diff
@@ -204,13 +212,32 @@ struct MidiInputMapping {
...
+            } else if constexpr (std::is_same_v<T, ConfigKey>) {
+                return key == other.key &&
+                        options == other.options &&
+                        std::get<ConfigKey>(control) == std::get<ConfigKey>(other.control) &&
+                        description == other.description;
+            } else if constexpr (std::is_same_v<T, QJSValue>) {
+                return key == other.key && options == other.options &&
+                        std::get<QJSValue>(control).strictlyEquals(
+                                std::get<QJSValue>(other.control)) &&
+                        description == other.description;
```
christophehenry, line 230: Does this prevent from inserting the same lamdba multiple times?

Swiftb0y: I think so. If it refers to the same lambda (obviously it won't work if different lambdas with the same body are declared, but that's fine IMO).
```javascript
> const a = () => {}
> const b = () => {}
> const c = a
> a === a
true
> a === b
false
> c === a
true
```

Swiftb0y:
fyi, In the future I'd prefer if I can resolve my own comments when I see them as resolved. Thank you.

src/controllers/midi/midicontroller.cpp
```diff
@@ -13,6 +13,16 @@
...
 #include "util/math.h"
 
+MidiInputHandlerController::MidiInputHandlerController(
+        const LegacyMidiControllerMapping& mapping,
+        const MidiInputMapping& inputMapping)
+        : m_mapping(mapping), m_inputMapping(inputMapping) {
+}
+
+void MidiInputHandlerController::disconnect() {
+    m_mapping.removeInputMapping(this->m_inputMapping.key.key);
```
Swiftb0y, line 23: In the long term, `removeInputMapping` probably needs a little more work because it just removes all mappings with the same key. I can imagine that it's probably not super uncommon that people will want to bind different handlers to the same key (midi status + control) and then only remove one of them. At least that's required to make a proper `MidiInputHandlerController.disconnect()` work.

src/controllers/midi/midicontroller.cpp
```diff
@@ -537,3 +585,36 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
...
+        this->getScriptEngine()->throwJSError(error);
+        return;
+    }
+
+    if (status <= 0 || midino <= 0) {
+        auto mStatusError = std::format(
+                "Invalid status or midino passed to midi.makeInputHandler. "
+                "Please pass a strictly positive integer. status={},midino={}",
+                status,
+                midino);
```
Swiftb0y, line 608: as much as I like std::format too, we still support compilers that don't have a built-in implementation. So you'll have to resort to Qt's formatting facilities for now.

```suggestion
        auto mStatusError = QStringLiteral(
                "Invalid status or midino passed to midi.makeInputHandler. "
                "Please pass a strictly positive integer. status=%1,midino=%2").arg(
                status,
                midino);
```

src/controllers/midi/midimessage.h
```diff
@@ -204,13 +212,32 @@ struct MidiInputMapping {
...
+            } else
+                static_assert(always_false_v<T>, "non-exhaustive visitor");
```
Swiftb0y, line 232: shouldn't this also suffice? I don't want to introduce unnecessary boilerplate.
```suggestion
                static_assert(false, "non-exhaustive visitor");
```

src/controllers/midi/midicontroller.cpp
```diff
@@ -312,7 +360,7 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
         bool found = false;
         for (auto it = m_fourteen_bit_queued_mappings.begin();
              it != m_fourteen_bit_queued_mappings.end(); ++it) {
-            if (it->first.control == mapping.control) {
+            if (std::get<ConfigKey>(it->first.control) == configKey) {
```
Swiftb0y, line 363: I think for exception safety, it would be better to check in every loop iteration.

```suggestion
            const ConfigKey* fourteen_bit_control = std::get_if<ConfigKey>(it->first.control);
            VERIFY_OR_DEBUG_ASSERT(fourteen_bit_control == nullptr) {
                continue;
            }
            if (*fourteen_bit_control == configKey) {
```

src/controllers/midi/midicontroller.cpp
```diff
@@ -537,3 +585,36 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
     qCWarning(m_logBase) << "MidiController: No script function specified for"
                          << MidiUtils::formatSysexMessage(getName(), data, timestamp);
 }
+
+const QString sErrorMessage(
+        "Invalid timer callback provided to midi.makeInputHandler. "
+        "Please pass a function and make sure that your code contains no syntax errors.");
```
Swiftb0y, line 591: these belong in an anonymous namespace (our pattern to achieve internal linkage) at the top of the file. Also prefix them `k` for constant instead of `s` for static and wrap the string in a QStringLiteral.
```suggestion
const QString kMakeInputHandlerError = QStringLiteral(
        "Invalid timer callback provided to midi.makeInputHandler. "
        "Please pass a function and make sure that your code contains no syntax errors.");
```

src/controllers/midi/midimessage.h
christophehenry: This line comes from [`std::visit`](https://en.cppreference.com/w/cpp/utility/variant/visit#Example) documentation. Coming from Python and Kotlin, I admit I'm having troubles to understand C++'s type system.

Swiftb0y: yeah I saw that it comes from there, I'm just confused whether its really required. 

> I admit I'm having troubles to understand C++'s type system.

Yeah... its complicated. I can recommend you the cppcon "back to basics" talks. Probably "Templates - Bob Steagall", maybe "Function and Class Templates - Dan Saks" as primer. Also look into overload resolution since that becomes important for generics (I think). I'm don't understand it good enough either, just barely good enough for basic everyday use.

christophehenry:
So I have the feature working here. Beyond documentation, I suppose I need to add a few tests?

res/controllers/midi-controller-api.d.ts
```diff
@@ -32,5 +32,5 @@ declare namespace midi {
 
     type InputCallback = (channel: string, control: string, value: number, status: number) => void
 
-    function makeInputHandler(status: number, midino: number, callback: InputCallback): MidiInputHandlerController
+    function makeInputHandler(status: number, midino: number, callback: InputCallback): MidiInputHandlerController | undefined
```
Swiftb0y, line 35: when would it return `undefined`? It either succeeds or throws, right?

src/controllers/midi/midicontroller.h
```diff
@@ -10,15 +10,16 @@
 
 class MidiOutputHandler;
 
-class MidiInputHandlerController : public QObject {
+class MidiInputHandlerController final : public QObject {
```
Swiftb0y, line 13: lets also call this `MidiInputHandleJSProxy` to follow the naming scheme of the other classes exposed to JS. In the long-term we should probably just put these in a namespace, but thats a refactoring for another PR.

Swiftb0y (commented):
great. this is taking shape.

Swiftb0y:
> So I have the feature working here. Beyond documentation, I suppose I need to add a few tests?

That would be wonderful, yes. Also we should try to get CI building now...

Swiftb0y:
Can you please rebase this branch on `main`? We're not accepting new features into 2.4 any more and the compiler features I suggested you use are only available on the targets used on `main`.

res/controllers/midi-controller-api.d.ts
christophehenry: Right. It's a bit counter-intuitif since the C++ implementation still returns an undefined value.

Swiftb0y: yeah... but I think that is simply swallowed essentially. So I think we can omit that detail from the API description. wdyt?

christophehenry:
> Can you please rebase this branch on main?

Sure. Can I squash the commits?

Swiftb0y:
Sure, just don't forget about the two comments from my last review.

christophehenry:
I think I messed up my rebase. I'l figure it out tomorrow.

Swiftb0y:
2.4 probably contains commit not yet in `main`. You could've done `git rebase --onto main 2.4` then it would've only taken the commits until `2.4` and put those onto `main`. But since you squashed anyways you can just force reset to `main` and then cherry-pick the single squashed commit.

src/controllers/midi/midicontroller.cpp
```diff
@@ -292,46 +292,52 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
...
-                               };
-                               if (!pEngine->executeFunction(function, args)) {
-                                   qCWarning(m_logBase)
-                                           << "MidiController: Invalid "
-                                              "anonymous script function";
-                               }
-                               return;
-                           }},
+        return std::visit(
+                MidiUtils::overloaded{
```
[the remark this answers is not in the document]
Swiftb0y: so now this is causing build failures on ubuntu... I guess try removing it again, maybe its not necessary anymore now that we target `main` instead of 2.4?

christophehenry:
Ok so I don't really know how to test this. Could you point me to relevent example tests in code and/or resources? Also, while testing the feature, I noticed Mixxx was often segfaulting after reloading changes from controller code. It may be related to that feature.

src/controllers/midi/midicontroller.h
```diff
@@ -1,12 +1,28 @@
...
+    Q_OBJECT
+  public:
+    MidiInputHandleJSProxy(
+            const std::shared_ptr<LegacyMidiControllerMapping> mapping,
+            const MidiInputMapping& inputMapping);
+    Q_INVOKABLE bool disconnect();
+
+  protected:
+    std::shared_ptr<LegacyMidiControllerMapping> m_mapping;
+    MidiInputMapping m_inputMapping;
```
christophehenry, line 23: Is there a risk of memory leak here? I'm not sure whether or not a `MidiInputHandleJSProxy` will ever be collected and in which conditions.

christophehenry, line 23: Is there a risk of memory leak here? I'm not sure whether or not a `MidiInputHandleJSProxy` will ever be collected and in which conditions.

christophehenry:
The macOS target fails with that `deduction guide` error. Maybe a macro would solve this?

Swiftb0y:
> Could you point me to relevent example tests in code and/or resources?

Sure, I think `src/test/midicontrollertest.cpp` is good example and a good place to add these tests to. 

> Also, while testing the feature, I noticed Mixxx was often segfaulting after reloading changes from controller code.

I'll try to investigate.

src/controllers/midi/midicontroller.h
Swiftb0y: So the ownership of the object is basically transferred to the javascript engine and its garbage collector when you call `QJSEngine::newQObject`. So `MidiInputHandleJSProxy` is destroyed once its unreachable from JS and the js-engines garbage collector destroys it. So no risk of memory leaks here. Also a memory leak wouldn't results in crash, just resource exhaustion in the long term.

Swiftb0y:
> The macOS target fails with that deduction guide error. Maybe a macro would solve this?

Possibly, yes. Maybe a combination of `__cplusplus` and `__APPLE__` ifdefs.

Swiftb0y:
Can you also take care of the clazy complaint? Thank you.

Swiftb0y:
> Also, while testing the feature, I noticed Mixxx was often segfaulting after reloading changes from controller code.

Can you put together a more concrete plan to reproduce the crash. I currently have trouble with that.

Swiftb0y:
Alternatively a backtrace would be nice for me to analyze. See https://github.com/mixxxdj/mixxx/wiki/Creating%20Backtraces

christophehenry:
[Here is a backtrace.](https://github.com/mixxxdj/mixxx/files/14283358/gdb.txt) Tell me if it's sufficient.

Swiftb0y:
> [Here is a backtrace.](https://github.com/mixxxdj/mixxx/files/14283358/gdb.txt) Tell me if it's sufficient.

Can you also post the JS code you used to get this? Does this also happen on startup or when you first start the mapping or only when you reload it?

> Also Clazy is still failing. I don't know why.

It's the same error as on MacOS. I guess the issue is clang-related and the preprocessor defines only catch apple. I guess you can replace the `__APPLE__` with a `__clang__` for now.

Swiftb0y:
Currently, my best guess is that we're still holding on to a QJSValue (in the `QMultiHash`) that was managed by an already destroyed QJSEngine (due to reloading).

christophehenry:
> Can you also post the JS code you used to get this? Does this also happen on startup or when you first start the mapping or only when you reload it?

Sure. I was able to reproduce with only this:

```js
DenonDNS3700 = new (class {
    init(id, debug){
        let handler = midi.makeInputHandler(0x90, 0x43, () => {
            console.error(arguments);
        });

        handler.disconnect();
    }
})();


DenonDNS3700.shutdown = function(){
}
```

```xml
<?xml version='1.0' encoding='utf-8'?>
<MixxxControllerPreset mixxxVersion="2.2.3" schemaVersion="1">
    <info>
        <name>Denon DN-S3700</name>
        <author>HoneyHazard, Wtrz, Augier</author>
        <description>Controller preset for Denon DN-S3700 turntable</description>
    </info>
    <controller id="DN-S3700">
        <scriptfiles>
            <file filename="lodash.mixxx.js"/>
            <file filename="midi-components-0.0.js"/>
            <file functionprefix="DenonDNS3700" filename="Denon-DN-S3700-hybrid-scripts.js"/>
        </scriptfiles>
        <outputs/>
    </controller>
</MixxxControllerPreset>
```

A simplement back and forth between removing and adding `handler.disconnect();` line reproduces the error after a couple of times.

Swiftb0y:
yeah, so without having confirmed it 100% myself. I think the problem is the lingering QJSValues in the mappings. Since due to the embedding of `QJSValue` into `MidiInputMapping`, the `MidiInputMapping` is not allowed to outlive the JSEngine. I think the most practical workaround is to just remove all the QJSValue-containing `MidiInputMappings` from `LegacyMidiControllerMapping::m_inputMappings` before the engine has shut down. Does that make sense?

src/controllers/midi/midicontroller.cpp
```diff
@@ -275,23 +300,53 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
...
-                mapping.control.group,
-        };
-        if (!pEngine->executeFunction(&function, args)) {
-            qCWarning(m_logBase) << "MidiController: Invalid script function"
-                                 << mapping.control.item;
-        }
+        return std::visit(
+                MidiUtils::overloaded{
+                        [this, pEngine, channel, control, value, status](
+                                ConfigKey configKey) {
```
Swiftb0y, line 306: make clazy happy.
```suggestion
                                const ConfigKey& configKey) {
```

christophehenry: I can't make it a `const`. `pEngine->executeFunction` expects a not const pointer. Is `ConfigKey& configKey` ok or there's another way?

Swiftb0y: I don't see how thats a problem. you can construct the non-const `QJSValue function` from a `const configKey.item`, or a I wrong? Do you get a compiler error?

christophehenry: Yup. [On that block](https://github.com/mixxxdj/mixxx/pull/12781/files/c2a31a0cafcab566dd21b4406a2fa689b923eadb#diff-ee87c49f05ff8aa897f732378f814f8e917cd4ec0b014b3021cda3c3e839dc7cR325-R340). I can have `[this, pEngine, channel, control, value, status](QJSValue& function) { ... }`, but not `const QJSValue& function`. Is this a problem?

Swiftb0y: Ah right, there it makes sense. to take by mutable reference (`QJSValue&`). but the snippet we're talking about is the ConfigKey which doesn't need to be taken by mutable ref. I think they can be different for both branches.

Swiftb0y: Oh, interesting I see why. We take the MidiInputMapping as const ref, so in any way to execute a function on the QJSValue, it needs to copy it. I think the constness goes quite deep here, so it probably doesn't make much sense to refactor it to avoid the copy. It's probably quite cheap anyways (at least that's what Qt usually assumes). In that case, you can just take the `const ConfigKey&` in one lambda and by-value (creating a copy in the process) with `QJSValue`. So `[this, pEngine, channel, control, value, status](QJSValue function) {`

christophehenry: How do I disable the clang-tidy warning?

Swiftb0y: I'm not sure what you mean, which clang-tidy warning? the clazy warning is fixed by taking the `const ConfigKey&`. The ugly `auto fun = function` is avoided by taking the `QJSValue` by value. You can do both at the same time.

christophehenry: Well, clang-tidy complains if I write:

```c++
[this, pEngine, channel, control, value, status](
        QJSValue function) {
    const auto args = QJSValueList{ /* snip **/  };
    if (!pEngine->executeFunction(&function, args)) {  /* snip **/    }
    return;
}}
```

and I get a compiler error if I write:

```c++
[this, pEngine, channel, control, value, status](
        &QJSValue function) {
    const auto args = QJSValueList{ /* snip **/  };
    if (!pEngine->executeFunction(function, args)) {  /* snip **/    }
    return;
}}
```

The only thing that pleases both is:

```c++
[this, pEngine, channel, control, value, status](
        const &QJSValue function) {
    const auto args = QJSValueList{ /* snip **/  };
    auto fun = function;
    if (!pEngine->executeFunction(function, args)) {  /* snip **/    }
    return;
}}
```

which is… meh.

christophehenry: I don't understant why `bool ControllerScriptEngineBase::executeFunction` was modified to pass `QJSValue* pFunctionObject` and not `const QJSValue& pFunctionObject`. I thought the second was safer than the first? :thinking:

Swiftb0y: > (Parodn me If I ask too many questions. C++ is still very obscure to me)

No worries. 

> I don't understant why bool ControllerScriptEngineBase::executeFunction was modified to pass QJSValue* pFunctionObject and not const QJSValue& pFunctionObject

Because `ControllerScriptEngineBase::executeFunction` will eventually call `QJSValue::call`, which is non-const. So `executeFunction` can't take a const reference. Otherwise it would have to copy the QJSValue every time before being able to call `call` on it.

> Well, clang-tidy complains if I write:

What clang-tidy error exactly? I don't get any error locally.

christophehenry:
Can you help me with the current build fail error? It compiles here. I don't understand what's hapening.

Swiftb0y:
> Can you help me with the current build fail error? It compiles here. I don't understand what's hapening.

https://bugreports.qt.io/browse/QTBUG-119461

QMultiHash no longer derives from QHash. So that's why it can't convert the QMultiHash to the QHash taken by QMutableHashIterator.
I assume it passes locally for you because you're compiling for Qt5?

use [`QMultiHash::removeIf`](https://doc.qt.io/qt-6/qmultihash.html#removeIf) instead. 

You can easily compile for Qt6 by passing `-DQT6=ON` to cmake.

Swiftb0y:
Thinking about this more, I think we need a different solution. I think the JS-ownership is only useful for objects that are exclusively managed by the QJSEngine. Cleaning out the QJSValues helps to avoid the crash after engine shutdown, but the QJSValues could get deallocated even sooner. I haven't tried to confirm myself, but something like this should probably crash eventually too (without shutting down the engine first).

```js
makeInputHandler(...); // ignore the returned handler, thereby immediately going out of scope 
gc() // from QJSEngine::GarbageCollectionExtension
```

This would probably cause the `MidiInputHandleJSProxy` to get deallocated too. Which in turn will get the stored QJSValue containing the callback function to get deallocated too. I'm not super sure how to solve this right now. Maybe create the `MidiInputHandleJSProxy` with `QJSEngine::CppOwnership`, or maybe put the callback in a shared_ptr? So, `std::variant<ConfigKey, std::shared_ptr<QJSValue>>`... I don't know right now.

christophehenry:
It is still useful to remove the JS handlers upon reload anyway.

src/controllers/midi/midicontroller.cpp
christophehenry: Hmm… You were right. clang-tidy doesn't complain.

christophehenry:
So, typically, how I would test this feature in Python is:
- [ ] test 1: I would call `makeInputHandler` from JS and verify that a new QJSValue mapping was created (I don't know how to access `MidiController::getScriptEngine` from the test; the method is protected, is it something I must achive with gMock?),
- [ ] test 2: trigger a midi signal and verify that the JS method was called (I don't know how to verify that; any ideas/relevant tests?)
- [ ] tests 3: trigger `ControllerScriptEngineBase::shutdown` and verify that the mapping was removed from the midi controller.

Is it the correct granularity, here? Any more suggestions?

Swiftb0y:
> Is it the correct granularity, here? 

I think so yes. 

I haven't done much testing myself to be fair (not in mixxx at least). I think its best to introduce a new test class that runs its own engine (see ControllerScriptEngineLegacyTest) and then mocks `LegacyMidiControllerMapping`. That way you  can verify that the add/removeInputMapping methods are called as expected from JS. 

>  don't know how to access MidiController::getScriptEngine from the test; the method is protected, is it something I must achive with gMock?),

While debatable whether its a good use of a feature, rather than trying to get access via the linker (which is what gMock does afaik), you can just mark the test class a friend.

https://en.cppreference.com/w/cpp/language/friend

>  test 2: trigger a midi signal and verify that the JS method was called (I don't know how to verify that; any ideas/relevant tests?)

Maybe look at `MidiControllerTest`? Not sure.

Swiftb0y:
> It is still useful to remove the JS handlers upon reload anyway.

I agree, though it would probably make more sense to refactor so all handlers are removed anyway since when the XML is reloaded the non-js handlers would have to get added back as well. I'm not even sure if the reloading currently respects the XML, it may not now that I think about it.

christophehenry:
So tests are written and passing locally. I don't really what's with that `aifffile.h` build error. Have you a solution for the ownership problem?

Swiftb0y:
> So tests are written and passing locally. I don't really what's with that aifffile.h build error. 

Errors in the QML build, has been fixed in main just now. 

>  Do you have a solution for the ownership problem?

Well, I don't currently know what the best solution is. Most of them require a bit of refactoring ideally. The primary problem I see is that `MidiInputMapping::control` is essentially shared between `MidiInputHandleJSProxy` and `LegacyMidiControllerMapping::m_inputMappings`. Previously that was not a problem because the member (of type `ConfigKey`) could easily be copied (just strings), while the QJSValue can not since it has reference semantics. I think we can try the `std::variant<ConfigKey, std::shared_ptr<QJSValue>>` first and see if that'll solve it.

christophehenry:
> I think we can try the `std::variant<ConfigKey, std::shared_ptr<QJSValue>>` first and see if that'll solve it.

If you're referring to the segfault I observed upon reload, cleaning the mapping when the engine shuts down seems to have solved it.

Swiftb0y:
> Errors in the QML build, has been fixed in main just now.

in case it wasn't clear, CI will work again once you rebase on or merge the most recent `main` commit.

Swiftb0y:
To avoid the issue, wrap the QJSValue of the callback function in a `std::shared_ptr`, so it'll stay around until the both `MidiInputHandleJSProxy` and `LegacyMidiControllerMapping` have got rid of it.

Swiftb0y:
I think it would be interesting to test whether [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) is available in Qt and whether it could be used to confirm whether the previously outlined issue would/could occur.

christophehenry:
So I made the `std::shared_ptr` change and I will test the FinalizationRegistry tomorrow, maybe start a mix to test the case.

src/controllers/midi/midimessage.h
```diff
@@ -223,10 +221,13 @@ struct MidiInputMapping {
                         options == other.options &&
                         std::get<ConfigKey>(control) == std::get<ConfigKey>(other.control) &&
                         description == other.description;
-            } else if constexpr (std::is_same_v<T, QJSValue>) {
+            } else if constexpr (std::is_same_v<T, std::shared_ptr<QJSValue>>) {
+                const std::shared_ptr<QJSValue> otherControl =
+                        std::get<std::shared_ptr<QJSValue>>(other.control);
+                const std::shared_ptr<QJSValue> thisControl =
+                        std::get<std::shared_ptr<QJSValue>>(control);
```
Swiftb0y, line 228: if you make these a reference, we don't encounter the reference-counting overhead. In this case its possible because we don't take ownership and the references don't outlive the shared_ptr they're obtained from.

christophehenry: You mean like this :

```suggestion
                const std::shared_ptr<QJSValue>& otherControl =
                        std::get<std::shared_ptr<QJSValue>>(other.control);
                const std::shared_ptr<QJSValue>& thisControl =
                        std::get<std::shared_ptr<QJSValue>>(control);
```

christophehenry: Or more like this?

```suggestion
                const auto& otherControl = *std::get<std::shared_ptr<QJSValue>>(other.control);
                const auto& thisControl = *std::get<std::shared_ptr<QJSValue>>(control);
```

src/controllers/midi/midicontroller.cpp
```diff
@@ -275,23 +305,53 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
...
-        if (!pEngine->executeFunction(&function, args)) {
-            qCWarning(m_logBase) << "MidiController: Invalid script function"
-                                 << mapping.control.item;
-        }
+        return std::visit(
+                MidiUtils::overloaded{
+                        [this, pEngine, channel, control, value, status](
+                                const ConfigKey& configKey) {
+                            QJSValue function =
+                                    pEngine->wrapFunctionCode(
```
daschuer, line 313: I think we should dispose the MidiOption::Script flag and use only the variant type instead. This function wrapper creation can be moved to LegacyMidiControllerMappingFileHandler::load().

daschuer (changes requested):
Thank you for picking this up. I have left some comments. 
It would be nice, If we still allow the mapping wizard to reassign a certain button. 
I think in order to make that work the makeInputHandler needs to check if a midi messag is already mapped.

src/controllers/midi/midimessage.h
```diff
@@ -204,13 +210,35 @@ struct MidiInputMapping {
...
+                static_assert(always_false_v<T>, "non-exhaustive visitor");
+        },
+                control,
+                other.control);
     }
 
     MidiKey key;
     MidiOptions options;
-    ConfigKey control;
+    std::variant<ConfigKey, std::shared_ptr<QJSValue>> control;
```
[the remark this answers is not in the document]
Swiftb0y: > We may consider a new name, "destination" or such...
> 
> Is it correct, that it can now carry a
> 
>     * ConfigKey form a Mixxx control object
> 
>     * Group + Name of a script binding comming from XML
> 
>     * The new function pointer.
> 
> 


Yes, though group+name of script binding was previously also stored in the same configkey. It's a bad design decision made way earlier. Refactoring this now is out-of-scope IMO.

> I think we should also get rid of this on the fly wrapping that is done for every incoming midi message:

yes, I agree. this PR would make that possible now. Though it can also be done later.

src/controllers/midi/midimessage.h
christophehenry: The second one compiles and run correctly. I prefer it tbh. So I added this code to my controller:

```js
for(let i = 0; i < 1000; i++){
    midi.makeInputHandler(0x90, 0x43, () => {
        console.error("Without variable");
    });
    gc();
    this.inputHandlers.push(
        midi.makeInputHandler(0x90, 0x43, () => {
            console.error("With variable");
        })
    );
}
```

None of the handlers were reclaimed. It's been running for an hour now. The handler still correctly react to the buttons press.

Swiftb0y: > The second one compiles and run correctly. I prefer it tbh

But it's incorrect... at least it doesn't avoid the unnecessary reference-count manipulation... I also find it a little weird because its quite unidiomatic. Out of curiosity, why do you prefer it?

Swiftb0y: > None of the handlers were reclaimed.

So the one without variable didn't either? I guess that's possible, it's up to the gc to decide when to reclaim... Still, it would probably crash once it does reclaim any of these handlers (without the shared_ptr we added).

Be-ing:
I don't think this is a good approach. It would be much simpler to dump all incoming MIDI to a designated function and handle the routing to callbacks in JavaScript.

src/controllers/midi/midimessage.h
christophehenry: > So the one without variable didn't either?

Nope. Everything worked perfectly until the end of the mix.

daschuer:
> dump all incoming MIDI to a designated function

Than we would have two dispatchers in a row which has a performance implications.

src/controllers/midi/midicontroller.cpp
daschuer: Maybe I don't have yet a good overview about all, so lets have a look into is a bit more and correct me If I am wrong. 

Here: https://github.com/mixxxdj/mixxx/blob/4e860da4821bfd7c4cd118c42d3d6ee0aac1ec31/src/controllers/midi/legacymidicontrollermappingfilehandler.cpp#L95
The stored xml file is read and the flag `MidiOption::Script` is set, indicating that the `item` element dos not contain a control object but a script function name. 

This PR now adds the feature to add fast direct pointer into the data structure, which should be blazing fast compared to the old method, that calls `wrapFunctionCode()` for every single incoming telegram. My idea was to call `wrapFunctionCode()` only once and than reuse the `QJSValue function` like you did here. 

Therefore wee need a `Controller::getScriptEngine()` access, which is not available in `load()`
Maybe we populate that in `MidiController::setMapping()` or just lazy during `processInputMapping()` itself. 

Does that make sense?

Swiftb0y:
> Than we would have two dispatchers in a row which has performance implications.

We can't really know until we measured. Having a single function to dispatch to could allow to fast-path the C++ code, potentially fast enough to avoid the resulting JS dispatch overhead.

res/controllers/midi-controller-api.d.ts
christophehenry: So if we don't distinguish anymore from anonymous and named function, I have to reintroduce this argument. At least for documentation consistency.

Swiftb0y: yes... you're right, to some extend. I still don't think its super useful though. Why would the handler take a group, when there is no way to specify one from the handler creation function. I still don't think you should be able to specify one because its just unnecessary overhead in the C++ domain.

Swiftb0y:
fyi I'm quite busy currently and then I'll be on holiday. In the worst case I won't get around to reviewing this until march. Thanks for understanding. In the meantime, someone else feel free to take over.

christophehenry:
Hoy, hoy! Are there something left to fix here or can we discuss about merging?

src/controllers/midi/midicontroller.cpp
```diff
@@ -245,27 +275,27 @@ void MidiController::receivedShortMessage(unsigned char status,
...
                 processInputMapping(it.value(), status, control, value, timestamp);
             }
             return;
         }
     }
 
-    auto it = m_pMapping->getInputMappings().constFind(mappingKey.key);
-    for (; it != m_pMapping->getInputMappings().constEnd() && it.key() == mappingKey.key; ++it) {
+    auto it = m_pMapping->getInputMappings().find(mappingKey.key);
+    for (; it != m_pMapping->getInputMappings().end() && it.key() == mappingKey.key; ++it) {
```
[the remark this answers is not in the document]
Swiftb0y: I'm also confused why this code even works in the first place. Could this be replaced by an `QMultiHash::equal_range`?

src/controllers/controllerinputmappingtablemodel.cpp
```diff
@@ -197,6 +197,17 @@ QVariant ControllerInputMappingTableModel::data(const QModelIndex& index,
...
+
+        if (std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
+            return QVariant();
+        }
+
+        const auto* const control = std::get_if<ConfigKey>(&mapping.control);
+
+        VERIFY_OR_DEBUG_ASSERT(control != nullptr) {
+            return QVariant();
+        }
```
Swiftb0y, line 209: this can be simplified, the first if can be removed.

Swiftb0y (commented):
I don't have time for a proper review really, but IMO it seems fairly low risk. What the other @mixxxdj/developers think?

src/controllers/midi/midicontroller.cpp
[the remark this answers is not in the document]
christophehenry: About the constness, that's because [`processInputMapping` now modifies the `InputMapping`](https://github.com/mixxxdj/mixxx/pull/12781/files/b5bfd1677cfd4820dbfd966d4595a3aca972b1be#diff-ee87c49f05ff8aa897f732378f814f8e917cd4ec0b014b3021cda3c3e839dc7cR312)

src/controllers/controllerinputmappingtablemodel.cpp
christophehenry: IU don't think this should `VERIFY_OR_DEBUG_ASSERT` in line 201. This is a valid case, not a code error. Originally, lines 207-209 weren't present but I  think I remeber I wa told to replace `std::get` by `std::get_if` and add `VERIFY_OR_DEBUG_ASSERT`.

But 201-209 can perfectly be replaced by:

```suggestion
        if (std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
            return QVariant();
        }

        const auto* const control = std::get<ConfigKey>(&mapping.control);
```

Swiftb0y: Yup, I'm sorry the current state is fine, I totally looked past the `VERIFY_OR_DEBUG_ASSERT`.

christophehenry:
I hate to be insistant but I'd really love this to get merged so I can progress on my DN-S3700 controller.

res/controllers/midi-controller-api.d.ts
Swiftb0y: lets keep the group for now, but please add some jsdoc comments and mark the group as deprecated, wdyt?

Swiftb0y (approved):
lgtm apart from that. @daschuer can you double check whether your comments have been addressed? 
The next step after this PR that would be nice would be a first class ComponentJS integration.

src/controllers/midi/midicontroller.cpp
```diff
@@ -275,23 +305,37 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
...
+        VERIFY_OR_DEBUG_ASSERT(std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
+            return;
+        }
+
         const auto args = QJSValueList{
                 channel,
                 control,
                 value,
                 status,
-                mapping.control.group,
```
christophehenry, line 284: Oh dear, I just noticed this: If I lazily compute the function, the `group` parameter is lost. This is a breaking change then.

christophehenry: Also, [there's already a function cache](https://github.com/mixxxdj/mixxx/blob/main/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp#L71) so is there really a need to lazily compute the function after all?

src/controllers/midi/midicontroller.cpp
```diff
@@ -14,11 +14,31 @@
 #include "util/make_const_iterator.h"
 #include "util/math.h"
 
+const QString kMakeInputHandlerError = QStringLiteral(
+        "Invalid timer callback provided to midi.makeInputHandler. "
+        "Please pass a function and make sure that your code contains no syntax errors.");
```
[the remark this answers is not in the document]
daschuer: OK.

src/controllers/midi/midimessage.h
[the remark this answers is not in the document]
daschuer: Before, It was a ConfigKey called `control`. Now this `control` is wrapped in a std::variant witch really requires a new name to emphasize that and not create confusion when reading it elsewhere. 

I am fine with going in small steps here. Than please add a TODO comment for adding a third type for more type safety in std::variant  and another TODO for getting rid of the on the fly wrapping. I am afraid otherwise this discussion will got into oblivion.

daschuer (changes requested):

Be-ing:
> I don't have time for a proper review really, but IMO it seems fairly low risk. What the other @mixxxdj/developers think?

I disagree that adding a major new public API is low risk. I voiced strong concerns before which were largely ignored. I wrote the original document cited as the inspiration for this PR many years ago, and in the intervening years realized it was a bad idea that is drastically overcomplicated. I'm frankly confused why that ancient wiki page was taken as a starting point for a PR without so much as asking whether it's still relevant or desired. I checked Zulip for recent discussion on this topic and found nothing. I am dismayed those old discussions seem to have been forgotten and that this is being seriously considered for merging without discussion of alternatives.

Swiftb0y:
> I disagree that adding a major new public API is low risk.

While true in theory, its not public until properly documented. We have not committed to stability yet. 

>  I voiced strong concerns before which were largely ignored

where?

> in the intervening years realized it was a bad idea that is drastically overcomplicated.

There is no mention of that in the wiki page, nor any arguments that brought you to that conclusion.

> I'm frankly confused why that ancient wiki page was taken as a starting point for a PR without so much as asking whether it's still relevant or desired. 

There is no indication on that wiki page that it was outdated. Not even I was aware of that. 

> I am dismayed those old discussions seem to have been forgotten and that this is being seriously considered for merging without discussion of alternatives.

Then please suggest some. Don't blame us for not being able to read whatever thoughts you never bothered to write down!

Be-ing:
> where?

https://github.com/mixxxdj/mixxx/pull/12781#issuecomment-1951880005

> There is no mention of that in the wiki page

Why would there be? The wiki has been barely used for years. That page hasn't been touched since the migration from DocuWiki to GitHub Wiki.

> Then please suggest some.

I did: https://github.com/mixxxdj/mixxx/pull/12781#issuecomment-1951880005

Swiftb0y:
>> where?
>
> https://github.com/mixxxdj/mixxx/pull/12781#issuecomment-1951880005

I'm sorry that comment got lost over time. You're very welcome to provide an alternative implementation so we can make an informed decision based on measurements. If not, we'll just have to stick to what we have (aka this PR).

Be-ing:
> You're very welcome to provide an alternative implementation so we can make an informed decision based on measurements. If not, we'll just have to stick to what we have (aka this PR).

This feels very unfair, flouting normal procedures, and I'm bewildered why this is happening. Why is it up to me to write the alternative? If you recall, I *did* do that years ago (https://github.com/mixxxdj/mixxx/pull/2868), others (IIRC, yourself included @Swiftb0y) asked that the scope be dramatically expanded, then the whole project got derailed by technical obstacles and sits unimplemented.

JoergAtGithub:
This PR is not about a new controller system, it's a small enhancement to the existing one. Let's keep the scope small and discuss only the implementation details here.

Be-ing:
The implementation details are beside the point when the whole approach is bad. This discussion needs to go back to the drawing board.

mxmilkiib:
It is a big change. Like, the ability to dump the XML files, right? That's big!

Also, complaining in certain ways can most often help achieve the opposite of its intention.

fwcd:
> Like, the ability to dump the XML files, right?

IIUC we still need them for the mapping wizard as pointed out by @daschuer. While it's nice to have the flexibility of implementing custom input handling in JS, reassigning controls mechanically is probably going to be easier with structured XML than unstructured/dynamic JS. Therefore I would be slightly hesitant to rewrite all of the existing mappings in JS.

Swiftb0y:
> Picking up the old ideas for a whole new controller system with a new metadata format is still possible, though not a task suited to someone new to C++, and there are alternatives.

I definitely agree. But I think I'm favoring this incremental improvement for now. Even if its just to gain experience with it. If we decide it was a mistake, we can still throw it out when the change to modules comes (which will likely be breaking anyways). 

> This feels very unfair, flouting normal procedures, and I'm bewildered why this is happening. Why is it up to me to write the alternative? 

It's not. But this is what we have now and if you think your solution is better it should compete against this one based on real code and data. Discussing hypothetical alternatives is not productive right now. Holding up Christophes work over some glorified yet non-existent alternative is not fair either IMO. 

Anyways, I appreciate the change of tone in the last two comments from you. The discussion is much more rewarding this way. Also, this PR doesn't rule out any other approaches right now. I'd be very happy to see #3641 and following PRs revived.

JoergAtGithub:
The  feature added by this PR is useful for the complex functionality, that requires scripting and can't be implemeted in XML.
But this feature must not be used for simple 1:1 mappings to not break the MIDI wizard experience for the users who want create simple mappings, or later customize the official mappings without having programming skills. We should clearly state this in the documentation!

mxmilkiib:
> It would be easier and more fun to write mappings if development could be done entirely in JavaScript

> But this feature must not be used for simple 1:1 mappings to not break the MIDI wizard experience for the users who want create simple mappings, or later customize the official mappings

I guess what the system for is ambiguous? I, at least, need to reread the article and this issue afresh!

Practically thinking, I guess both systems react in parallel to the same trigger?

I am frustrated that access to participating in editing the Mixxx wiki is set the way it is, because if one wants a reason as to why we might be not working from the same page, imho I think it's that.

Be-ing:
Because this is adding to the legacy controller mapping system, how it interacts with the old XML way and the mapping wizard should be clarified. I suggest to let them coexist in the same mapping in case someone starts with the wizard and then wants to rewrite it in JS. However, I suggest adding a policy for reviewing mappings, that if a mapping uses JS callbacks, it should not use the legacy `<controls>` XML, but stick with one or the other.

No arguments have been presented for why the implementation was done this way versus dumping all incoming MIDI to one function, as [I suggested before](https://github.com/mixxxdj/mixxx/pull/12781#issuecomment-1951880005). I do not think there is any compelling reason to implement the routing from MIDI input to callbacks in C++ rather than JavaScript. Doing it in C++ adds more complexity/long term maintenance burden to legacy code.

Dumping the raw bytes to JS then letting JS handle the routing gives maximum flexibility to the mapping developer of how they want to organize their JS code. Although MIDI is a standardized protocol, we know manufacturers sometimes do weird things with it sometimes. So, it's best for the interface between Mixxx and the JS code to be as minimal as it can be to give mapping developers maximum flexibility in the device-specific JS. This also mirrors how HID mappings already work, making it easier for someone who has written a mapping for one protocol to write one for another, and, perhaps, write code that could work with either MIDI or HID. A JS library could be provided with Mixxx for the callback routing, possibly based on [my old code](https://github.com/mixxxdj/mixxx/blob/3ef56defc53c6d2376cd25b91bdd428d922c4196/res/controllers/lib/mididispatcher.mjs). If we decide that API isn't great, no big deal, just keep the old JS library around while providing a new one. Old mappings would continue to work without needing to commit to maintaining a C++ API forever. And if someone ever does get around to resuming the JS module based controller system, porting a JS library to it should be fairly simple, whereas a C++ API likely would not be so straightforward to port to a new system.

christophehenry:
@Swiftb0y @daschuer I added a commit to revert the lazy wrapping of XML's control entries to prevent breaking change and added a todo on the `control` name as per @daschuer's comment.

daschuer (approved):
LGTM, and still works. Thank you. 

Do you have interest to create a follow up PR that interacts with learned mappings?

daschuer:
I think we need to do the following: 
* skip adding a JS callback when midi value is already mapped by XML 
* Show script input mappings in the Input Mapping Tab (form this PR) 
* Show script output mappings in the Input Mapping Tab (engine.connectControl)

A complete overview in the GUI Tabs would lift the requirement to do simple things still with XML IMHO @JoergAtGithub  Do you agree?

JoergAtGithub:
@daschuer Your proposal sounds good, but I think for the simple 1:1 relations the XML should be always the prefered solution, and we should continue to require this for official mappings, because:
- Better maintainability of sematic information (we can automatically convert mappings to any future mapping syntax like QML at any time, if we once lose the sematic information, we are in a dead end and must support this particular syntax forever)
- Users without programming skills are able to do custom mapping modifications of existing mappings using the wizard (e.g. reassign a button that they don't use to a function that they use)

daschuer:
That makes sense. Let's lint the second point by being cooperative and not install a mapping form JS if the midi value has been mapped via wizard.

christophehenry:
I'm sorry, what are learned mappings?

Swiftb0y:
Mappings that have been created by using the [controller wizard](https://manual.mixxx.org/2.4/en/chapters/advanced_topics#advanced-controller)

christophehenry:
I'm not sure I understand what I should do next.

Swiftb0y:
I think this is a good next small feature:

> skip adding a JS callback when midi value is already mapped by XML

mxmilkiib:
> I'm sorry, what are learned mappings?

🤦 I think you are owed a bit of apology; if you're asking that, at this point, whilst actively developing the controller system, then it means we have failed in our attempts to advise on what the point of this FR and now PR is, and the variety of ways people.. relate to controller mappings.

The "problem" [statement] this FR and now PR is solving is different in everyone's head.

It feels like a number of people think it would rock the boat too much to holistically point this out (and  actively reassess the drawing board) and instead piecemeal work through any implementation and only inform the person doing the work of the ramifications and requirements one at a time. If anything, I think this is a bit disrespectful to @christophehenry.

The difference or overlaps between entry level user mapping creator, advanced user mapping creator, and dev mapping creator is not clear in this conversation. Heck, the fact that mappings can be created in Vim before any use starts, or in Mixxx during use (this being what "learned mappings" refers to), hasn't been made clear.

Let us, everyone, *please*, turn down the bolshie level, and *be humble*, and slow down, and admit that there are questions that need to be actively and deeply listened to and mindfully approached for various points of consensus to be achieved, and that there *may* possibly need to be some element of backtracking, at least widening in conceptions.

There's no magic path to everyone grokking everything that needs to be grokked, and for solutions to become obviously apparent.

I don't think there should be any shame in pointing all this out (n it's not fair to try intimate there is), or in pivoting to reassess understanding and implementation.

If it is feared that reassessing will slow things down, well I think it will actually be slower to not because energy will be shed at a far higher rate and development (which includes defining problem statements) will stall.

christophehenry:
Is this feature planned for 2.5?

Swiftb0y:
Depends on when we branched `main` (I don't remember of the top of my head). If we branched 2.5 after this pr got merged, it'll be in `2.5`.

mxmilkiib:
Would this be a method for [re]connecting pad MIDI presses like in [this instance](https://mixxx.discourse.group/t/launchpad-pro-mk3-basic-4x16-hotcue-setup-troubles/29701/10?u=milkii)? I had assumed the ControllerJS input would work, but is it needing this feature? What kind of syntax would it be?

christophehenry:
This is an alternative to registering MIDI handler from XML. Itks a implementation of https://github.com/mixxxdj/mixxx/wiki/Registering-Midi-Input-Handlers-From-Javascript

mxmilkiib:
I've got [this](url) currently, but now the pads aren't even lighting up.

mxmilkiib:
I'm getting as far as getting this, but I can't find a way to fix the logic or encapsulation or whatever.
```
 controller.launchpadpromk3lppromk3midi.input:debug [Controller] "incoming: " "Launchpad Pro MK3 LPProMK3 MIDI: t:184167 ms status 0x90 (ch 1, opcode 0x9), ctrl 0x54, val 0x7F"
warning [Controller] ControlDoublePrivate::getControl returning NULL for ( "[Channel5]" , "hotcue_17_activate" )
```
Are there any code examples beyond the test? Cheers!

Swiftb0y:
I think you're getting quite far. the problem now is likely that there is no `[Channel5]` created so it can't access a hotcue there. What skin are you using? You can can use the developer tools to check if `[Channel5],hotcue_17_activate` actually exists.

mxmilkiib:
It doesn't, this is 4 decks, and whilst each deck has 32 or 36 or whatever hotcues, I'm only going up to 16.

I get an error when I try to use .bind to bind the scope of the counter variables to that if the hotcue handler instantiation and not the prototype (if that's the right way to phrase it)

Am I using the syntax wrong? Or does they new method, as it stands, not account for that requirement yet?

Swiftb0y:
right, but `[Channel5]` must be coming from somewhere, you probably have an off-by-one error somewhere. Having a quick look at the gist however, I can't tell you where exactly. Try adding some logmessages and see how that happens.

Swiftb0y:
I think the problem is the mutability of the `deck` in line 155. try copying that into the handler closure on line 177.

mxmilkiib:
Thanks for the feedback. I added the console output log, but indeed, that does sound about right. After the 4th loop, the loop counter has been ++ to 5, same with i to 17.

Ah yes, the deck and i loop counters should be made with `let` not `var`! Gosh, I've forgotten so much JS. This change appears to fix the main issue I was having. New hotcues are created, and existing ones get played.

The strange thing that still remains is that the deck UI play/pause button doesn't change after pressing a pad triggers a cue, needing two clicks to stop the playback.

Swiftb0y:
> Ah yes, the deck and i loop counters should be made with let not var! Gosh, I've forgotten so much JS. This change appears to fix the main issue I was having. New hotcues are created, and existing ones get played.

Yeah, just use `const` whenever you can, `let` whenever you must and `var` only in very very special edge cases (which you should probably consult with an expert on before).

> The strange thing that still remains is that the deck UI play/pause button doesn't change after pressing a pad triggers a cue, needing two clicks to stop the playback.

yeah, thats probably because you simply reject the hotcuepad release message, leading mixxx to believe that you're still holding down the hotcuebutton even though you released it. If you simply delete the entire `input` definition of the `HotcueButton` in your mapping, it should work. The default hotcuebutton already does everything you want it to do. you only need to keep the `sendRGB`, `group`, `midi` and `number` definitions.

mxmilkiib:
Aah, thanks for the feedback. I've [commented out](https://gist.github.com/mxmilkiib/2362403b62ef37699ccfd498b0b6065f#file-lpp3-js-L114) the `input` but I don't get any activity on pressing the lit pads now. I had only asked about it here and added that method because this was the case before. I had thus come to think the ComponentJS system was a bit weak in functionality, but if .HotcueButton is really meant to provide an input that, then I was wrong to think that, but I'm back to that square again, albeit mildly wiser.

Swiftb0y:
ah, this is conflating multiple things now. So currently `ComponentsJS` is not yet setup to automatically bind input handlers. Currently you don't get any input because you commented out the code where the input handling connection is made. Until ComponentJS has native support here, you'll want:
```js
this.hotcueButtons[i-1] = new components.HotcueButton({...});
this.hotcueButtons[i-1].inputHandlerHandle = midi.makeInputHandler(0x90, padAddress, someArray[i].input);
```
You can then use `this.hotcueButtons[i-1].inputHandlerHandle.disconnect()` IIRC if you ever want to unbind that handler again.

PR #12783, opened 2024-02-10
Add simple script for developing with Fedora 39

luisfcorreia:
Aiming at 2.4 branch dependencies

### DIFF
--- a/tools/fedora_buildenv.sh
+++ b/tools/fedora_buildenv.sh
@@ -1,0 +1,99 @@
+#!/bin/bash
+# This script works with Fedora Worstation 39 (and possibly other RPM based distros, not tested)
+# shellcheck disable=SC1091
+set -o pipefail
+
+case "$1" in
+    name)
+        echo "No build environment name required for Debian based distros." >&2
+        echo "This script installs the build dependencies via apt using the \"setup\" option." >&2
+        ;;
+
+    setup)
+        source /etc/os-release 2>/dev/null
+        case "${ID}" in
+            fedora) # Fedora 39
+                echo "Tested with Fedora 39 for mixxx 2.4 branch\n"
+                ;;
+            *) # untested
+                echo "Untested distro\n"
+                                
+        esac
+
+        echo "Enabling RPMFusion free and nonfree repos"
+		sudo dnf install -y \
+			https://mirrors.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm\
+			https://mirrors.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm
+
+		sudo dnf config-manager --enable fedora-cisco-openh264
+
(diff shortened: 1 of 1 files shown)


Swiftb0y:
whats the advantage of this script over the wiki section? https://github.com/mixxxdj/mixxx/wiki/Compiling%20on%20Linux#fedora

luisfcorreia:
if Ubuntu has a script, why can't Fedora have one as well?

daschuer:
The difference is that the Ubuntu script is used on the GitHub CI. However, we have also scripts to setup visual Studio, which not used in the CI. From this view we may integrate it, to lower the barrier for Fedora contributors.  
Does it have benefits for the Rpm Fusion build? 

@Swiftb0y What do you think?

Swiftb0y:
1. Ideally we'd have no packaging tools for linux distributions in our repo, that is bound to get unmaintainable considering the diversity linux distributions.
2. ubuntu (or rather apt(-get)) doesn't have a `builddep` command like `dnf` does nor does it have a `.spec` file in the packaging process as far as I know. So it makes sense to manage the list of dependencies in the repo (since there is no good place for it externally).
3. We currently do packaging for 3 platforms: Win, Mac and Ubuntu. The scripts for those are in the repo because they're required by our packaging CI (github Actions). Packaging for fedora is done on rpmfusion with the extra repository and thus packaging resources belong there (if they make sense, I doubt this script would make sense there currently). 

Does that make sense to you? Again, what is the advantage of the script over the wiki?

luisfcorreia:
> 1. Ideally we'd have no packaging tools for linux distributions in our repo, that is bound to get unmaintainable considering the diversity linux distributions.
> 
>     2. ubuntu (or rather apt(-get)) doesn't have a `builddep` command like `dnf` does nor does it have a `.spec` file in the packaging process as far as I know. So it makes sense to manage the list of dependencies in the repo (since there is no good place for it externally).
> 
>     3. We currently do packaging for 3 platforms: Win, Mac and Ubuntu. The scripts for those are in the repo because they're required by our packaging CI (github Actions). Packaging for fedora is done on rpmfusion with the extra repository and thus packaging resources belong there (if they make sense, I doubt this script would make sense there currently).
> 
> 
> Does that make sense to you? Again, what is the advantage of the script over the wiki?

But you do have packaging tools for Ubuntu and it isn't the only Linux distro.

I'm honestly baffled by these replies.

Copy paste errors can happen when people copy stuff from wiki, it's just that.

Swiftb0y:
> But you do have packaging tools for Ubuntu and it isn't the only Linux distro.

Yes, but its the exception. We only have packaging tools for the platforms we package for using github actions. 

> I'm honestly baffled by these replies.

I apalogize. I didn't intend for this discussion to give you a negative impression. But please understand our arguments that adding this script does little more apart from increasing the maintenance burden. Building Mixxx on fedora is already quite easy as you've probably noticed yourself (none of the hiccups you encountered and posted on zulip would've been solved by accepting this PR).

luisfcorreia:
it's fine, PR is closed

PR #12791, opened 2024-02-11
Tracks: select track row when clicking the preview button (only when starting preview)

### LINKED ISSUE #12789: Highlight the full line of the track, in the track list, when preview a track
### Feature Description

I is difficult to read the data of a track when preview, in the track list. Only Preview icon change to red color, but it does not help to read data in columns far away from right.

If the full line of preview is highlighted, it is a lot more easy, like when I click on another column in the track list.

ronso0:
Low prio, draft until we discussed if that is desired or conflicts with library use cases.

Selects the Preview row if Preview button is clicked to **start previewing**, no seletion change if preview is stopped.

The use case mentioned in #12789 is
* click Preview
* scroll right to read metadata in offscreen columns
* have some highlight to keep track of the row

Quoting myself:
> I can't imagine why one would like to keep the current selection when clicking the preview button. On the contrary, that would simplify loading that track to a deck (with Return key if double-click is configured accordingly in the Library preferences, Mixxx 2.4)

I think simply selecting the row is predictable (consistency with other columns).
Some other highlight is certainly feasible  but overkill IMHO.

### DIFF
--- a/src/library/previewbuttondelegate.cpp
+++ b/src/library/previewbuttondelegate.cpp
@@ -49,8 +49,7 @@ PreviewButtonDelegate::PreviewButtonDelegate(
           m_column(column),
           m_pPreviewDeckPlay(make_parented<ControlProxy>(
                   kPreviewDeckGroup, QStringLiteral("play"), this)),
-          m_pCueGotoAndPlay(make_parented<ControlProxy>(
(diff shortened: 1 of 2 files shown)


daschuer (approved):
LGTM, Thank you

PR #12798, opened 2024-02-12
Refactor/preferences enums

Swiftb0y:
Alternative to #12796 along with a little extra refactoring

### DIFF
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -27,7 +27,6 @@
 #include "dialog/dlgdevelopertools.h"
 #include "dialog/dlgkeywheel.h"
 #include "moc_mixxxmainwindow.cpp"
-#include "preferences/constants.h"
 #include "preferences/dialog/dlgpreferences.h"
 #ifdef __BROADCAST__
 #include "broadcast/broadcastmanager.h"
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -99,7 +99,7 @@ MixxxMainWindow::MixxxMainWindow(std::shared_ptr<mixxx::CoreServices> pCoreServi
 #endif
           m_pDeveloperToolsDlg(nullptr),
           m_pPrefDlg(nullptr),
-          m_toolTipsCfg(mixxx::preferences::constants::Tooltips::On) {
+          m_toolTipsCfg(mixxx::preferences::Tooltips::On) {
     DEBUG_ASSERT(pCoreServices);
     // These depend on the settings
 #ifdef __LINUX__
@@ -175,10 +175,10 @@ void MixxxMainWindow::initialize() {
     // Set the visibility of tooltips, default "1" = ON
     m_toolTipsCfg = pConfig->getValue(
             ConfigKey("[Controls]", "Tooltips"),
-            mixxx::preferences::constants::Tooltips::On);
+            mixxx::preferences::Tooltips::On);
 #ifdef MIXXX_USE_QOPENGL
     ToolTipQOpenGL::singleton().setActive(
-            m_toolTipsCfg == mixxx::preferences::constants::Tooltips::On);
+            m_toolTipsCfg == mixxx::preferences::Tooltips::On);
 #endif
 
 #ifdef __ENGINEPRIME__
@@ -1153,11 +1153,11 @@ void MixxxMainWindow::slotShowKeywheel(bool toggle) {
     }
 }
 
-void MixxxMainWindow::slotTooltipModeChanged(mixxx::preferences::constants::Tooltips tt) {
+void MixxxMainWindow::slotTooltipModeChanged(mixxx::preferences::Tooltips tt) {
     m_toolTipsCfg = tt;
 #ifdef MIXXX_USE_QOPENGL
     ToolTipQOpenGL::singleton().setActive(
-            m_toolTipsCfg == mixxx::preferences::constants::Tooltips::On);
+            m_toolTipsCfg == mixxx::preferences::Tooltips::On);
 #endif
 }
 
@@ -1260,14 +1260,14 @@ bool MixxxMainWindow::eventFilter(QObject* obj, QEvent* event) {
                         "DlgPreferences") {
             // return true for no tool tips
             switch (m_toolTipsCfg) {
-            case mixxx::preferences::constants::Tooltips::OnlyInLibrary:
+            case mixxx::preferences::Tooltips::OnlyInLibrary:
                 if (dynamic_cast<WBaseWidget*>(obj) != nullptr) {
                     return true;
                 }
                 break;
-            case mixxx::preferences::constants::Tooltips::On:
+            case mixxx::preferences::Tooltips::On:
                 break;
-            case mixxx::preferences::constants::Tooltips::Off:
+            case mixxx::preferences::Tooltips::Off:
                 return true;
             default:
                 DEBUG_ASSERT(!"m_toolTipsCfg value unknown");
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -100,7 +100,7 @@ MixxxMainWindow::MixxxMainWindow(std::shared_ptr<mixxx::CoreServices> pCoreServi
 #endif
           m_pDeveloperToolsDlg(nullptr),
           m_pPrefDlg(nullptr),
-          m_toolTipsCfg(mixxx::TooltipsPreference::TOOLTIPS_ON) {
+          m_toolTipsCfg(mixxx::preferences::constants::Tooltips::On) {
     DEBUG_ASSERT(pCoreServices);
     // These depend on the settings
 #ifdef __LINUX__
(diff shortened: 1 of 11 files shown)


Swiftb0y:
okay, enough force pushing now. I'd appreciate a review.

ronso0:
Looks good as far as I can tell, though I'd prefer to keep the constants compact therefore I removed the nested `constants` namespace (and tweaked some vars etc.)

Commit is b7713096df, or https://github.com/Swiftb0y/mixxx/pull/18, as you prefer.

Swiftb0y:
>  I'd prefer to keep the constants compact therefore I removed the nested constants namespace

I'd prefer to remove that too, but its required to work around QTBUG-68611 (as pointed out in the comment above). If we remove the constants namespace, all other members of the `preferences` namespace would have to be declared in that file as well.

ronso0:
Hmkay, which other members are you referring to? Future extension?
My PR is build without issues, or when would we notice issues?

src/preferences/dialog/dlgprefinterface.cpp
```diff
@@ -4,12 +4,14 @@
 #include <QList>
 #include <QLocale>
 #include <QScreen>
+#include <QVariant>
```
ronso0, line 7: not required

src/preferences/dialog/dlgprefinterface.cpp
```diff
@@ -4,12 +4,14 @@
...
 #include <QtGlobal>
 
 #include "control/controlobject.h"
 #include "control/controlproxy.h"
 #include "defs_urls.h"
 #include "moc_dlgprefinterface.cpp"
+#include "preferences/constants.h"
```
ronso0, line 14: can be moved to `preferences/dialog/dlgpreferencepage.h` to include it only once

ronso0: ... and be removed from `src/mixxxmainwindow.cpp`

src/preferences/dialog/dlgprefinterface.cpp
Swiftb0y: I used `QVariant::fromValue` further below, so I felt it was appropriate to include as per "include what you use".

ronso0: sure, but the DlgPef.. dialogs are QWidgets which already bring QVariant IINM

src/preferences/dialog/dlgprefinterface.cpp
Swiftb0y: are you sure about dlgpreferencepage.h? its not used in there anywhere.

ronso0: That's the baseclass of all pref pages, would it increase the include 'load' including it there?
If so, just leave as it is.

src/preferences/dialog/dlgprefinterface.cpp
Swiftb0y: yeah I understand that reasoning but that would mean that we'd rely on transitive dependencies, which is bad practice. In reality it doesn't make any difference, but I think leaving it there is better practice than removing (which is why I added it in the first place).

src/preferences/dialog/dlgprefinterface.cpp
Swiftb0y: not really, but its bad practice as well. It also wouldn't make any difference in terms of compilation time. In general its best to include what you use at the point of usage and nowhere else...

Swiftb0y: I'm okay with removing it from mixxxmainwindow.cpp, though it doesn't hurt to keep either. Doesn't really go against good practices either since it is used in that file...

ronso0: okay. it's already included in the header

Swiftb0y: yeah, so the cppcoreguidelines don't say much about that. If you insist, I'll remove the header from the `.cpp`.

ronso0: Yes, please remove it. It'd be removed anyway with the next include cleanup.

ronso0: Umm, this got a little mixed up, I was referring to the include in `src/mixxxmainwindow.cpp`.
Anyway, clean this up as much you like, I don't mind too much, just noticed it when looking for places where constants.h is included.

Swiftb0y: yup sorry. I don't think anymore cleanup is needed now.

Swiftb0y:
friendly ping @ronso0. If you're happy with the current state, I'll squash the fixup.

ronso0:
No one else reviewed this, and my c++/enum knowledge can't compete with yours, so... it LGTM, it builds, it works AFAICT.
Ready for merge IMO.
If there are no objections within the next 24h I'll press the button @mixxxdj/developers

ronso0:
Can you squash the fixups and push?

src/preferences/constants.h
```diff
@@ -1,20 +1,44 @@
...
 namespace mixxx {
 
+namespace preferences {
+
+namespace constants {
+Q_NAMESPACE
+
+// In order for this Q_NAMESPACE to work, all members of the namespace must
+// be declared here. see QTBUG-68611
+
```
Swiftb0y, line 15: I think for the other devs, this would be most controversial aspect. Putting a QObject header into this header just for the enums is not great in terms of compile-time overhead, but I think its a necessary evil anyways in order to make these enums usable from within QML.

ronso0:
@mixxxdj/developers Does someone have time to take a look at this?

src/mixxxmainwindow.cpp
```diff
@@ -1159,14 +1160,14 @@ bool MixxxMainWindow::eventFilter(QObject* obj, QEvent* event) {
                         "DlgPreferences") {
             // return true for no tool tips
             switch (m_toolTipsCfg) {
-            case mixxx::TooltipsPreference::TOOLTIPS_ONLY_IN_LIBRARY:
+            case mixxx::preferences::constants::Tooltips::OnlyInLibrary:
```
daschuer, line 1163: All the double :: look cluttered. Can we remove the "constants" namespace?

src/preferences/constants.h
daschuer: Q_NAMESPACE is defined in qobjectdefs.h

daschuer (changes requested):
Thank you. I have just two comments.

src/mixxxmainwindow.cpp
ronso0: we discussed that here https://github.com/mixxxdj/mixxx/pull/12798#issuecomment-1941300550

daschuer: Ah, I just read it. I can however not follow the reasoning. This PR introduces both "preferences::constants" there is no need to make it that long just to safe the "preferences" namespace for future use. We can introduce another independent namespace all the time. 

Btw: We have no policy of folder matching namespaces like in C#

src/preferences/constants.h
Swiftb0y: Tried that, but I got MOC errors.

Swiftb0y: Tried 5.15.12. Though I may be misremembering what header I tried.

Swiftb0y:
friendly ping @daschuer

src/mixxxmainwindow.cpp
daschuer: I have just checked how others do this: 
They define a namespace each type and than class less enums: 
https://github.com/zhuzichu520/FluentUI/blob/f830d5a9bfc0f6836e6e2aa7c80d7201b730013f/src/Def.h#L95
https://github.com/KDE/kdenlive/blob/7283b739900a01ef3caf86d458bedcf418ce1f60/src/definitions.h#L119

Lets follow them and not introduce  preferences and constants namespace. Just mixxx is sufficient IMHO.

Swiftb0y: I don't like the class-less enums though. There is no reason to open up this enum to implicit conversions to `int` etc. wbout
```c++
namespace mixxx {
namespace Tooltips {
Q_NAMESPACE
enum class TooltipsOptions {...}
Q_ENUM_NS(TooltipsOptions)
using enum TooltipsOptions;
}
}
```
would that work for you?

daschuer: The reason is to get rid of the enum class namespace when using the enum in QML and use the extra namespace instead. 

But it also works for me if we get rid of at least one of the extra namespaces as you like.

Swiftb0y: > The reason is to get rid of the enum class namespace when using the enum in QML and use the extra namespace instead.

Well, QML is a different discussion here. IIUC we can rename the type as we'd like. I'm primarily focused on the C++ ergonomics right now.

Swiftb0y:
I removed the namespace naming requirement by use of an `inline namespace`. wdyt @daschuer?

src/mixxxmainwindow.h
```diff
@@ -148,9 +148,9 @@ class MixxxMainWindow : public QMainWindow {
     std::unique_ptr<mixxx::LibraryExporter> m_pLibraryExporter;
 #endif
 
-    mixxx::TooltipsPreference m_toolTipsCfg;
+    mixxx::Tooltips m_toolTipsCfg;
 
-    mixxx::ScreenSaverPreference m_inhibitScreensaver;
+    mixxx::ScreenSaver m_inhibitScreensaver;
```
Swiftb0y, line 153: this looks a little short now, wdyt about reintroducing the preferences namespace again? Adding it to the enum name looks a little clunky.

ywwg: I agree, mixxx:screensaver would imply a subsystem about screensaver handling.  something like mixxx::preferences::screensaver make sense

daschuer:
> I removed the namespace naming requirement by use of an inline namespace. wdyt @daschuer?

Interesting. I didn't even know this feature exists. 

Works for me: 

```
            case mixxx::TooltipsPreference::TOOLTIPS_ONLY_IN_LIBRARY: // original 
            case mixxx::preferences::Tooltips::OnlyInLibrary: // proposal 
```

Swiftb0y:
@daschuer I think I addressed everything

daschuer (approved):
LGTM, Thank you for polishing this.

Issue #12802, opened 2024-02-12
Mixxx 2.4 AArch64 flatpak builds fail since PR #12664


djantti:
### Bug Description

Hi all! Mixxx 2.4 flatpak builds currently do not succeed on AArch64. On x86_64 everything builds just fine. The problem appeared with Mixxx PR #12664 from a couple of weeks ago and reverting the PR makes AArch64 flatpak builds work again.

The compilation fails with errors like this:

```
In file included from /run/build/mixxx/src/waveform/renderers/glwaveformrenderersignal.h:3,
                 from /run/build/mixxx/src/waveform/renderers/glvsynctestrenderer.h:3,
                 from /run/build/mixxx/src/waveform/renderers/glvsynctestrenderer.cpp:1:
/run/build/mixxx/src/waveform/renderers/glwaveformrenderer.h:7:59: error: expected class-name before ‘{’ token
    7 | class GLWaveformRenderer : protected QOpenGLFunctions_2_1 {
      |                                                           ^
/run/build/mixxx/src/waveform/renderers/glwaveformrenderer.h: In member function ‘virtual void GLWaveformRenderer::initializeGL()’:
/run/build/mixxx/src/waveform/renderers/glwaveformrenderer.h:10:9: error: ‘initializeOpenGLFunctions’ was not declared in this scope
   10 |         initializeOpenGLFunctions();
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~
```

This might be related to the fact that Flathub AArch64 org.kde.SDK is built using OpenGL ES2 for Qt Base (-opengl es2), but that's how far the investigation has progressed. Here's a link to the Flathub buildbot log and I've also attached a build log from my local AArch64 build.

[https://buildbot.flathub.org/#/builders/9/builds/9015/steps/7/logs/stdio](https://buildbot.flathub.org/#/builders/9/builds/9015/steps/7/logs/stdio)
[mixxx_flatpak_aarch64_build.txt](https://github.com/mixxxdj/mixxx/files/14248457/mixxx_flatpak_aarch64_build.txt)

### Version

2.4.0

### OS

Fedora 39

daschuer:
Pleas check if https://github.com/mixxxdj/mixxx/pull/12804 fixes the build.

Be-ing:
Fixed by #12804, though that doesn't mean #12803 shouldn't be considered too.

PR #12803, opened 2024-02-12
inherit unversioned QOpenGLFunctions instead of QOpenGLFunctions_2_1

Be-ing:
Currently the aarch64 KDE Flatpak runtime is built with `-opengl es` passed to Qt's configure script. It's questionable whether that's necessary, but regardless, Mixxx can support Qt built with that option.
https://invent.kde.org/packaging/flatpak-kde-runtime/-/issues/19
    
Fixes https://github.com/mixxxdj/mixxx/issues/12802

### DIFF
--- a/src/waveform/renderers/glwaveformrenderer.h
+++ b/src/waveform/renderers/glwaveformrenderer.h
@@ -2,15 +2,15 @@
 
 #include <QtGui/qtgui-config.h> // for QT_NO_OPENGL and QT_OPENGL_ES_2
 
-#if !defined(QT_NO_OPENGL) && !defined(QT_OPENGL_ES_2)
+#if !defined(QT_NO_OPENGL)
 
-#include <QOpenGLFunctions_2_1>
+#include <QOpenGLFunctions>
 
-class GLWaveformRenderer : protected QOpenGLFunctions_2_1 {
+class GLWaveformRenderer : protected QOpenGLFunctions {
   public:
     virtual void initializeGL() {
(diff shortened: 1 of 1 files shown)


daschuer:
We already have a release candidate and this is not a regression (except reverting the build error). So lets postpone this to 2.4.1 after some testing.

daschuer:
Did you consider to use QOpenGLFunctions_2_1 or QOpenGLFunctions_E2 conditionally depending on the QT_OPENGL_ES_2 flag? This way the change does not affect the other builds.

Be-ing:
> Did you consider to use QOpenGLFunctions_2_1 or QOpenGLFunctions_E2 conditionally depending on the QT_OPENGL_ES_2 flag? This way the change does not affect the other builds.

I think that would be overcomplicated. This simpler code works.

daschuer:
The difference is that this has an impact on all builds, which needs a good amount of test. The conditional option will only affect GLES only builds and will fail at compile time, not at runtime. I consider such a solution as a reasonable Quick 2.4.1 fix. 

By the way, this enables the LEGACY GL waveforms, that are scheduled for removal anyway. So we may also consider to close this without a merge.

Be-ing:
Now that the urgent build failure has been fixed with #12804, it is not urgent to merge this to the 2.4 branch, and I agree this would need thorough testing.

> By the way, this enables the LEGACY GL waveforms, that are scheduled for removal anyway. So we may also consider to close this without a merge.

That makes sense.

PR #12806, opened 2024-02-12
Implement cpu pinning and cgroup cpuset support

ronso0:
This @poelzi's #3534, rebased onto main and fixed.

### Requires `cpuset` tool
Linux:  official repos or https://github.com/lpechacek/cpuset/releases/tag/v1.6
maybe macOS 10.13.6 / 12.7.3 / 13.6.5 / 14.3.1 [[1]](https://man.freebsd.org/cgi/man.cgi?cpuset)

Quoting myself:
With all CPU governors set to `performance` I can run 3 decks with effects and pitch != 0 (RubberBand) with 5ms stable 🎉 (previously 10ms)
I've been running this for like a year without any issues on Ubuntu 20.04 with vanilla low-latency kernels.
Moved _mixxx-isolate.py_ to the build dir and ran mixxx with `./mixxx-isolate.py --mixxx ./mixxx` (will ask for root permission)

### Update:
I've changed the isolate script to switch to `performance` governor for all CPUs with `cpufreq-set -g`, and back to previous governor on exit. This was a step I was doing manually with cpupower-gui each time.

### _Disclaimer:_
_I have absolutely no clue (yet) how all the magic is done, I just fixed the conflicts and build errors.
So bare with me when you review this :p_
____
Original description from #3534:
Add command line argument to specify a cpuid and cpuset cgroup name for the engine.
The engine tries at startup to move itself to the cpuset provided or set the cpu affinity.
This will allow a wrapper program to free a core for the engine, minimizing latency jitter or CPU starvating.

## Instructions for script
Start with mixxx-isolate `tools/mixxx-isolate.py -- [mixxx args]`

The script will start itself as root in the background to do the necessary system changes. After that, it starts
mixxx which will itself move the engine into a special cpuset group.
After mixxx exists, the system is restored.
The script also takes the hyperthreading cpus offline that are siblings of the engine cpu.
This way, you get the most power out of your CPU but have none of the negative effects of hyperthreading.
You can customize the behaviour:

```
usage: mixxx-isolate.py
  [-h]
  [--cpuset_engine CPUSET_ENGINE]
  [--cpuset_system CPUSET_SYSTEM]
  [--cpuset_isolation CPUSET_ISOLATION]
  [--no-isolate-ht]
  [--no-offline]
  [--cpu-id CPU]
  [--cleanup]
  [--mixxx MIXXX] [--debug]
  [mixxx_args [mixxx_args ...]]

Runs mixxx with isolated engine on a dedicated CPU. You can pass mixxx arguments using ' -- '
after mixxx-isolate arguments. This program will seperate one CPU core from the system and
place the mixxx engine thread on a dedicated CPU.
Hyperthreading siblings will be isolated and taken offline if possible. The normal system state
is restored when mixxx is closed.

positional arguments:
  mixxx_args            Arguments passed to mixxx

optional arguments:
  -h, --help            show this help message and exit
  --cpuset_engine CPUSET_ENGINE
                        cpuset name for the mixxx engine
  --cpuset_system CPUSET_SYSTEM
                        cpuset name for the mixxx engine
  --cpuset_isolation CPUSET_ISOLATION
                        cpuset name for the mixxx engine
  --no-isolate-ht       don't isolate hyperthreading partner (not recommended)
  --no-offline          don't take hyperthreading cpus offline
  --cpu-id CPU          ID of the cpu to use for the engine. -1 for autoselect
  --cleanup             Cleanup old isolation settings
  --mixxx MIXXX         Mixxx executable
  --debug               Debug output
```

## Manual Instructions

You need to find out which cpu id's are one core by looking at /proc/cpuinfo and compare the core id.
In my case I use cpu 3 and it's hyperthreading partner 7

```sh
sudo cset shield --cpu 3,7 --kthread=on
echo 3 | sudo tee -a /sys/fs/cgroup/cpuset/user/cpuset.cpus
sudo chown $(id -u) /sys/fs/cgroup/cpuset/user/tasks
mixxx  --settingsPath ~/Projects/mixxx-devsettings/ --logLevel debug --developer --controllerDebug --engineCpuSet user
```

### Restore system
Stop the shield after you finished your usage:
```
sudo cset shield -r
```

### Find out which CPU to use

Look at the cpuinfo file
```sh
cat /proc/cpuinfo
```
The output looks like this:
```
processor       : 0
vendor_id       : GenuineIntel
cpu family      : 6
model           : 158
model name      : Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
stepping        : 9
microcode       : 0xde
cpu MHz         : 2839.943
cache size      : 8192 KB
physical id     : 0
siblings        : 8
core id         : 0
cpu cores       : 4
apicid          : 0
initial apicid  : 0
fpu             : yes
fpu_exception   : yes
cpuid level     : 22
wp              : yes
flags           : fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb invpcid_single pti ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2 erms invpcid rtm mpx rdseed adx smap clflushopt intel_pt xsaveopt xsavec xgetbv1 xsaves dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp md_clear flush_l1d
bugs            : cpu_meltdown spectre_v1 spectre_v2 spec_store_bypass l1tf mds swapgs taa itlb_multihit srbds
bogomips        : 5799.77
clflush size    : 64
cache_alignment : 64
address sizes   : 39 bits physical, 48 bits virtual
power management:

[...]
```
You can now choose any cpu, and find the matching entries which have the same *core id*. Use those in the instructions above.
Be aware that CPU Ids start at 0.


geraldog:
Very good @ronso0 thanks! That helps a lot!

ronso0:
@geraldog Very welcome!

If you're using this updated patch, could you share your experience?
How is overall stability?
Does it allow you to lower the buffer size?

geraldog:
@ronso0 for now I'm stuck at 2.4 with a big party coming up in mid-April.

I already apply a couple of patches to Mixxx (https://github.com/geraldog/mixxx/tree/fourstems_padj) and can't deal with the burden of applying even more right now.

After mid-April I plan on migrating to main, so I will be able to give you more detailed feedback.

ronso0:
@geraldog Did you have a chance to test this?

geraldog:
Hey @ronso0 no as I haven't got the chance to update to 2.5 yet.

ronso0:
I've changed the isolate script to switch to `performance` governor for all CPUs with `cpufreq-set -g`, and back to previous governor on exit. This was a step I was doing manually with cpupower-gui each time.

acolombier (commented):
Only did a surface review, but code LGTM. 

I'm curious to  hear if you've seen any significant performance improvement? Also wondering if we should allow other threads to be pinnable (I'm thinking of the controller which can go through some burst on chatty devices such as the S4 Mk3)

ronso0:
> I'm curious to hear if you've seen any significant performance improvement?

I don't think so : \ at least it didn't enable 5ms for me, but 10 ms is pretty stable now with 2 decks + pitch shift + effects and passthrough (almost no xruns).

ronso0:
> I'm curious to hear if you've seen any significant performance improvement?

I'll test again with current setup, I think some configs might have changed since I last did (kernel, rt-related configs).

geraldog:
I'm having a bit of trouble getting this PR to run properly with CPU isolation.

It seems my kernel / userspace has switched to cgroups v2 which doesn't work with cpuset cset.

This is gentoo Linux on OpenRC and I've switched rc.conf to use hybrid cgroups but ```mount```  (from util-linux) which is used by ```cset``` does not recognized the cpuset fs type, which is really weird, I double-checked the util-linux sources and they include cpuset support still.

@ronso0 I finally began taking the plunge to Mixxx main branch to use @acolombier Stems.

I have a full setup including a custom mapping for AKAI MIDIMix and it was my hope that this PR would allow me to use Stems with Rubberband R3.

geraldog:
Nevermind, I got it working after I realized there's a kernel option to activate legacy cpuset v1.

And hey, it works. There's still one xrun-causing latency spike when I press play on a deck, but otherwise I have Rubberband R3 running pretty stable with 2 decks and Stems.

acolombier:
Thanks for the feedback here! @ronso0 do you think there is a chance we can complete this for 2.6?

ronso0:
@geraldog Great you made it!
And good to hear it improves perfromance for you, too!

As stated in the disclaimer, I just picked this up and didn't deal with the cpuset commands or other major internals.
How exactly did you notice it is incompatible with cgroups v2?
What sort of script check / response could we implement to make this easy for users?

ronso0:
> Thanks for the feedback here! @ronso0 do you think there is a chance we can complete this for 2.6?

Sure, I'm available for fixups if someone does an in-depth review.
Though, thorough  testing is crucial here, and IME the little responses indicate the lack of interest amongst core devs, so idk if we get sufficient testing.
Or just publish the feature with the 'experimental' disclaimer.

geraldog:
> @geraldog Great you made it! And good to hear it improves perfromance for you, too!

Thanks!

> As stated in the disclaimer, I just picked this up and didn't deal with the cpuset commands or other major internals. How exactly did you notice it is incompatible with cgroups v2? What sort of script check / response could we implement to make this easy for users?

See https://github.com/SUSE/cpuset/issues/40

src/util/cpupinning.h
```diff
@@ -0,0 +1,13 @@
+#include <Qt>
+
+namespace mixxx {
+
+class CpuPinning {
+  public:
+    /// Pin the current thread to cpuid
+    static bool pinThreadToCpu(qint32 cpuid);
+    /// Moves the thread to a cpuset cgroup
+    static bool moveThreadToCpuset(const QString& cgroup);
```
acolombier, line 10: The `#ifdef __LINUX__` is missing here

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+            cpuToCore[currentCpu] = coreId
+            lastCore = max(lastCore, coreId)
+
+    useCore = -1
+    if userCpu == -1:
+        useCore = lastCore
+    else:
+        useCore = cpuToCore.get(userCpu, -1)
+        if useCore == -1:
+            print("Selected CPU does not exist. Using autodetect.")
```
acolombier, line 61: Should your use stderr here?

```suggestion
            print("Selected CPU does not exist. Using autodetect.", file=sys.stderr)
```

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+        else:
+            system.append(cpu)
+
+    return {"use": useCpu, "isolate": isolate, "system": system}
+
+
+def cpusetname(name):
+    if name and name[0] == "/":
+        return name
+    return "/%s" % name
```
acolombier, line 81: Python 3 syntax

```suggestion
    return f"/{name}"
```

tools/mixxx-isolate.py
acolombier: Actually, it maybe worth consolidating everything inside `log` module

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+    processes = 1
+    while processes:
+        try:
+            log.debug(" Move processes from cpuset %s %s", setInstance, name)
+            proc.move(setInstance, "/", None, False)
+            processes = setInstance.tasks
+            log.debug(" Processes left: %s", processes)
+            break
+        except BrokenPipeError:
+            continue
```
acolombier, line 99: I feel like this is a significant risk to create an infinite loop. Shall we error out and break/abort here? Or maybe introduce a retry count limit

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+        return True
+    # move processes until empty
+    processes = 1
+    while processes:
+        try:
+            log.debug(" Move processes from cpuset %s %s", setInstance, name)
+            proc.move(setInstance, "/", None, False)
+            processes = setInstance.tasks
+            log.debug(" Processes left: %s", processes)
+            break
```
acolombier, line 97: `processes` only seems to be used as the loop predicate, but there is also an explicit `break`. Could we consolidate in a single approach here?

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+        except BrokenPipeError:
+            continue
+        except Exception as e:
+            log.exception(e)
+            print(e)
```
acolombier, line 102: Any reason for that blanket catch? We should consolidate that in a single logging facility and fail

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+                fp.write("1")
+
+    log.info(" Deactivate cpusets")
+    deleteCpuset(args.cpuset_engine, args.debug)
+    deleteCpuset(args.cpuset_isolation, args.debug)
+    deleteCpuset(args.cpuset_system, args.debug)
+
+    if prev_gov is not None:
+        log.info(" Restore governor %s" % prev_gov)
+        system("sudo cpufreq-set -g %s" % prev_gov)
```
acolombier, line 138: This looks quite worrying from a security stand point:
- Can we make sure we don't hardcode a `sudo` here and either expect the user to run the script with `sudo` or provide the right permission setup instruction 
- Can we make sure we don't eval an arbitrary string inside a `system` command

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+
+    log.info(" Cleanup finished")
+    print("STATUS: CLEAN")
```
acolombier, line 141: Ditto: logging facility consolidation

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+
+
+def createcpuset(name, cpuspec, memspec, cx, mx):
+    try:
+        ccset.create(name, cpuspec, memspec, cx, mx)
+    except CpusetExists:
+        pass
+    except Exception as e:
+        raise e
```
acolombier, line 150: This is redundant

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+
+    log.info(" Cleanup finished")
+    print("STATUS: CLEAN")
+
+
+def createcpuset(name, cpuspec, memspec, cx, mx):
+    try:
+        ccset.create(name, cpuspec, memspec, cx, mx)
+    except CpusetExists:
+        pass
```
acolombier, line 148: Shall we add some debug or info logging here?

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+                memspec,
+                False,
+                False,
+            )
+    except Exception as instance:
+        teardown(args, conf["isolate"])
+        # unroll
+        log.exception(instance)
+        log.critical(" Failed to create shield, hint: do other cpusets exist?")
+        sys.exit(23)
```
acolombier, line 186:
```suggestion
    except:
        teardown(args, conf["isolate"])
        # unroll
        log.exception("Failed to create shield, hint: do other cpusets exist?")
        sys.exit(23)
```

[log.exception](https://docs.python.org/3/library/logging.html#logging.Logger.exception) should already log the current exception details/traceback, and the argument should be used to pass a custom message

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+    while True:
+        try:
+            proc.move("root", args.cpuset_system, root_tasks, args.debug)
+            break
+        except cset.CpusetException:
+            pass
+
+    if not args.no_offline:
+        for iso in conf["isolate"]:
+            with open("/sys/devices/system/cpu/cpu%d/online" % iso, "w") as fp:
```
acolombier, line 200:
```suggestion
            with open(f"/sys/devices/system/cpu/cpu{iso}/online", "w") as fp:
```

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+
+    # move all processes to the system cpuset
+    log.info(" Move processes into system group")
+    root_tasks = cset.unique_set("/").tasks
+    while True:
+        try:
+            proc.move("root", args.cpuset_system, root_tasks, args.debug)
+            break
+        except cset.CpusetException:
+            pass
```
acolombier, line 196: Risk of infinite loop - retry count?

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+        try:
+            s = input(">>>")
+            if s == "exit":
+                break
+            else:
+                print("unknown command")
+        except KeyboardInterrupt:
+            break
+        except EOFError:
+            break
```
acolombier, line 226: Do we need an interactive prompt? Could we simply wait for `sig` here and not require a TTY?

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+
+    teardown(args, conf["isolate"])
+    sys.exit(0)
+
+
+def selectCommand(args):
```
acolombier, line 232: Ditto `sudo` usage

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+                        + args.mixxx_args
+                    )
+                    mixxxProc.wait()
+                    log.info(" Mixxx finished")
+                    helper.stdin.write("exit\n".encode("ASCII"))
+                    try:
+                        # this might likely not work
+                        helper.send_signal(signal.SIGUSR1)
+                    except PermissionError:
+                        pass
```
acolombier, line 331: what is the value of silencing this exception?

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+            else:
+                print(line)
+
+        except KeyboardInterrupt:
+            log.info(" Received keyboard interrupt. Stopping")
+            if mixxxProc and mixxxProc.returncode is None:
+                mixxxProc.send_signal(signal.SIGINT)
+            break
+        except Exception as e:
+            log.exception(e)
```
acolombier, line 346:
```suggestion
        except:
            log.exception("An error occurred")
```

Also, there is a risk of infinite loop

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+    # trigger system cleanup
+    try:
+        # this might likely not work
+        helper.send_signal(signal.SIGUSR1)
+    except PermissionError:
+        pass
```
acolombier, line 352: what is the value of silencing this exception?

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+
+    helper.stdin.write("exit\n".encode("ASCII"))
+    helper.communicate()
+    log.info(" Exit successfully")
+    sys.exit(0)
+
+
+def isCpufreqSetAvailable():
+    from shutil import which
```
acolombier, line 361: Move the import to file header

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+
+        global prev_gov
+        if isCpufreqSetAvailable():
+            prev_gov = subprocess.run(
+                [
+                    "cat",
+                    "/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor",
+                ],
+                capture_output=True,
+            ).stdout.decode()
```
acolombier, line 479: Use `open` here instead of subprocess

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+            if prev_gov is not None:
+                log.info(" Current governor is '%s'" % prev_gov)
+                log.info(" Set 'performance' governor for all CPUs")
+                system("sudo cpufreq-set -g performance")
+        elif platform.system() == "Linux":
+            log.info(" Command 'cpufreq-set' not found.")
```
acolombier, line 485: `warn` or `error`?

tools/mixxx-isolate.py
```diff
@@ -0,0 +1,496 @@
...
+                [
+                    "cat",
+                    "/sys/devices/system/cpu/cpu0/cpufreq/scaling_governor",
+                ],
+                capture_output=True,
+            ).stdout.decode()
+            if prev_gov is not None:
+                log.info(" Current governor is '%s'" % prev_gov)
+                log.info(" Set 'performance' governor for all CPUs")
+                system("sudo cpufreq-set -g performance")
```
acolombier, line 483: Ditto `sudo`

acolombier (changes requested):
The Mixxx core changes looks okay, but the helper script still needs some care

ronso0:
Thanks for your script review!
I'm by no means a python programmer, please feel free to push fixes to my branch!

ronso0:
Turns out this still allows me to lower the buffer by one step.
(5ms with SignalSmith, without this I get occassional xruns when one deck plays and when beatjumping)

Happy to receive commits to fixing the script as per @acolombier suggestions.
Will resolve the conflicts soonish.

Issue #12808, opened 2024-02-13
Numark Mixtrack Pro 3: script tried to connect to ControlObject ([Sampler5], play) which is non-existent


eiglow:
### Bug Description

This error occurs on startup. It seems to cause my controller to not work, because along with the error, it says "The functionality provided by this controller mapping will be disabled until the issue has been resolved."

```
Uncaught exception: file:///usr/share/mixxx/controllers/Numark-Mixtrack-3-scripts.js:814: Error: script tried to connect to ControlObject ([Sampler5], play) which is non-existent.
Backtrace: @file:///usr/share/mixxx/controllers/Numark-Mixtrack-3-scripts.js:814
@file:///usr/share/mixxx/controllers/Numark-Mixtrack-3-scripts.js:894
```

The error only seems to have started occurring after my last update, so it may be from a recent change. I'll edit the ticket if I find out more info for a known good version.

### Version

2.4.0-beta (git d112e4c)

### OS

Fedora 40 (rawhide)

ronso0:
Thank you for reporting this bug!

The commit causing this regression is reverted in #12809

PR #12769, opened 2024-02-09
controllers: ensure required samplers are created

ronso0:
This PR simply ensures mappings request all required samplers being created, exactly like they set `[App],num_decks` if they need more than 2 decks.

With #12657 some mapping flaws produce mapping warnings, and users may not be able to resolve the issues.
This will occur with a fresh profile or when users don't have any tracks loaded in samplers with numbers higher than the deafult (4) or skin minimum (8, Shade skin).
After having merged #12657 this is just much more likely to happen.

To reproduce:
* ~~load Shade skin~~ doesn't matter, mappings are loaded before skins are created https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/2.2E4.20samplers.20regression
* close Mixxx, remove samplers.xml to ensure not more than the samplers required by the skin are created (Shade needs 8)
* start Mixxx (8 sample players are created)
* load Denon MC7000 or DDJ-FLX4 mapping (see #12747) or any of the mappings fixed in this PR
* mapping warning will popup

There may be more mappings that affected in a different way, I just did a quick check where `[Sampler` is used with
https://github.com/search?q=repo%3Amixxxdj%2Fmixxx+%22%5BSampler+path%3A%2F%5Eres%5C%2Fcontrollers%5C%2F%2F&type=code&p=2

PR #12809, opened 2024-02-13
Revert "Samplers: don't create unneeded empty samplers during startup"

ronso0:
This reverts the core commit of #12657 

Besides the Numark Mixtrack Pro3 #12808 other mappings slipped through in #12769.
I'm at it but I can't guarantuee to fix _all_ of them _in time_.
Also, user mappings may be broken, too.

Let's push #12657 to 2.4.1

To summarize: the mapping control error can still occur if
* _samplers.xml_ doesn't exist / is broken
* loading a mapping that needs N samplers
* before loading a skin that provides N samplers

### DIFF
--- a/src/mixer/playermanager.cpp
+++ b/src/mixer/playermanager.cpp
@@ -450,12 +450,7 @@ void PlayerManager::addDeckInner() {
 }
 
 void PlayerManager::loadSamplers() {
-    // This is only called by CoreServices during startup to restore
-    // samplers from the previous session.
-    // We don't want it to create more players than necessary.
-    bool dontCreateEmptySamplers = true;
-    m_pSamplerBank->loadSamplerBankFromPath(getDefaultSamplerPath(m_pConfig),
-            dontCreateEmptySamplers);
+    m_pSamplerBank->loadSamplerBankFromPath(getDefaultSamplerPath(m_pConfig));
 }
 
 void PlayerManager::addSampler() {
--- a/src/mixer/samplerbank.cpp
+++ b/src/mixer/samplerbank.cpp
@@ -165,8 +165,7 @@ void SamplerBank::slotLoadSamplerBank(double v) {
     }
 }
 
-bool SamplerBank::loadSamplerBankFromPath(const QString& samplerBankPath,
-        bool dontCreateEmptySamplers) {
+bool SamplerBank::loadSamplerBankFromPath(const QString& samplerBankPath) {
     // The user has picked a new directory via a file dialog. This means the
     // system sandboxer (if we are sandboxed) has granted us permission to this
     // folder. We don't need access to this file on a regular basis so we do not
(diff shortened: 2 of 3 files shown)


ronso0:
I think in order to not require users to create expected decks and samplers explicitly in script files this could be done where control existence is checked in `ControllerScriptInterfaceLegacy`:
* if control refers to a deck/sampler > current deck/sampler count
   * if control exists for deck/sampler **1**
      * create deck/sampler
   * else throw error
* else throw error

daschuer:
Oh, I am sorry this topic has put extra workload stress on you.

I am not sure if merging this is a good idea. Do we hunt one regression (now only in custom mappings) with another, the performance regression and the reason we have created dedicated sampler skins? 

What is the effect when a mapping requests missing samplers? I think using a sampler without a GUI is unlikely. The assertion about the missing control is easy to fix. Probably more easy than a latency increase. 

After merging this, we have hard time to identify the not updated mappings. 

In general it is probably also questionable if mappings should be allowed to control samplers without a GUI.

ronso0:
> Oh, I am sorry this topic has put extra workload stress on you.

Never mind, it was my mistake to not do a smoke test with my regular setup. I'd have noticed right away.

Tbh I didn't measure the performance implication of 64 unneeded samplers. Should be about more memory being used, not really CPU load (no sampler loaded --> no caching reader work, no engine buffer work IIUC).
And no one noticed or cared about for years, so I think this is no big deal for now.

> What is the effect when a mapping requests missing samplers? I think using a sampler without a GUI is unlikely. 

The GUI samplers are created _after_ the mapping has been loaded, that's the issue.
That's the sample deck reation order:
* `CoreServices::initialize` triggers creation of the default samplers (4 currently) [here, line 309](https://github.com/mixxxdj/mixxx/blob/170987331434fc33becaabe13d6c0446bc107023/src/coreservices.cpp#L309-L311)
* a bit later, same function, _samplers.xml_ is read [here, line 431](https://github.com/mixxxdj/mixxx/blob/170987331434fc33becaabe13d6c0446bc107023/src/coreservices.cpp#L431)
   for example: 64 samplers, tracks in samplers 1-12, = 12 samplers are created
* (MixxxMainWindow is initialized)
* main.cpp triggers controller setup [here](https://github.com/mixxxdj/mixxx/blob/170987331434fc33becaabe13d6c0446bc107023/src/main.cpp#L98), mappings are loaded 
* skin is loaded, skin manifest requests 16 samplers in [`LegacySkinParser::parseSkin`](https://github.com/mixxxdj/mixxx/blob/170987331434fc33becaabe13d6c0446bc107023/src/skin/legacy/legacyskinparser.cpp#L346)


> The assertion about the missing control is easy to fix.

Not sure what you have in mind. Something like I expressed above  https://github.com/mixxxdj/mixxx/pull/12809#issuecomment-1941111344

> In general it is probably also questionable if mappings should be allowed to control samplers without a GUI.

Hmm, not sure about this.
Also, that can not be detected reliably: we know if a mapping conatins a direct xml mapping for a deck/sampler control, or tries to make a connection to one, but such a control  also be in a mapping function and IIUC we would only know about this when the function is executed because that's when it's evaluated.

**Bottom line:**
I recommend to merge this and work on a fix for 2.4.1 or 2.5

daschuer (approved):
OK, than lets merge it. Thank you for explaining.

PR #12811, opened 2024-02-13
History: show track count and duration in sidebar

### LINKED ISSUE #12788: Number of tracks and global duration on History
### Feature Description

To have the same information than Playlists and Crates.
- number of tracks.
- Global duration of all crates. 

To know that today I must copy the historic into a crate (or playlist).

ronso0:
Exactly like in Playlists.

I thought it would be possible to avoid the code duplication ([PlaylistFeature::createPlaylistLabels()](https://github.com/mixxxdj/mixxx/blob/23e91ac0bf7c86c5b03dc2d7048396be6de2ad5c/src/library/trackset/playlistfeature.cpp#L135) and move this to a shared base method, but that wouldn't simplify the situation since History also needs the 'year created' in addition to an id/label pair.

Closes #12788

### DIFF
--- a/src/library/trackset/baseplaylistfeature.cpp
+++ b/src/library/trackset/baseplaylistfeature.cpp
@@ -802,6 +802,16 @@ void BasePlaylistFeature::markTreeItem(TreeItem* pTreeItem) {
     }
 }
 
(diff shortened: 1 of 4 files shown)


daschuer (approved):
LGTM, Thank you.

Issue #12824, opened 2024-02-16
Confusing Midi Sysex handling


Swiftb0y:
### Bug Description

So after struggling a bit, the documentation for handling incoming sysex from midi controllers as well as its actual implementation is a bit weird. 
In order to receive sysex, the mapping needs to specify a `<control>` element. While a `<key>` can be added (schema-wise) its value will be ignored, instead the engine will always expect there to be a `ControllerPrefix.incomingData` function to be available. That behavior itself is weird and counter intuitive, either it should not be necessary to specify a control, or an arbitrary handler function should be possible.

### Version

2.5-alpha-241-g73378b2853

### OS

Fedora Linux 39

git-developer:
This issue is probably a duplicate of #11536 ([details](https://github.com/mixxxdj/mixxx/issues/13133#issuecomment-2665440271)).

PR #12827, opened 2024-02-16
refactor(controllers): send sysex to all handlers

Swiftb0y:
Something I noticed while digging through #12824, probably a bug, but its masked by the unintuitive behavior described in #12824 anyways. Potentially proper fixes for #12824 to come.

### DIFF
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -516,9 +516,9 @@ void MidiController::receive(const QByteArray& data, mixxx::Duration timestamp) 
         }
     }
 
-    auto it = m_pMapping->getInputMappings().constFind(mappingKey.key);
-    for (; it != m_pMapping->getInputMappings().constEnd() && it.key() == mappingKey.key; ++it) {
-        processInputMapping(it.value(), data, timestamp);
+    const auto inputMappings = m_pMapping->getInputMappings().values(mappingKey.key);
+    for (const auto& inputMapping : inputMappings) {
+        processInputMapping(inputMapping, data, timestamp);
     }
 }
 


JoergAtGithub:
LGTM! Thank you!

Swiftb0y:
Thank you.

src/controllers/midi/midicontroller.cpp
```diff
@@ -516,9 +516,9 @@ void MidiController::receive(const QByteArray& data, mixxx::Duration timestamp)
         }
     }
 
-    auto it = m_pMapping->getInputMappings().constFind(mappingKey.key);
-    for (; it != m_pMapping->getInputMappings().constEnd() && it.key() == mappingKey.key; ++it) {
-        processInputMapping(it.value(), data, timestamp);
+    const auto inputMappings = m_pMapping->getInputMappings().values(mappingKey.key);
```
daschuer, line 519: values() returns a fresh allocated QList()
https://github.com/qt/qtbase/blob/adc0920c48d812fb1b3e2b7bfc76217a7126a41e/src/corelib/tools/qhash.h#L1088C38-L1088C63
This happens for every message and should be avoided.

daschuer:
@Swiftb0y can we find a solution that does not allocate memory?

src/controllers/midi/midicontroller.cpp
Swiftb0y: Damn, I tought the QMultiHash was using them internally and just returned an implicitly-shared copy. I should've checked. 

> @Swiftb0y can we find a solution that does not allocate memory?

I didn't at first, but I just found it now, It's https://doc.qt.io/qt-6/qmultihash.html#equal_range. PR underway.

daschuer: Thank you.

Swiftb0y:
https://github.com/mixxxdj/mixxx/pull/12843

PR #11245, opened 2023-02-03
Pioneer DDJ-FLX4: New mapping base on DDJ-400

Robert904:
Mapping for the Pioneer DDJ-FLX4 controller.

It has been based on the DDJ-400, adapting some part to the new layout and buttons.
Mostly the FX part has been adapted and the lights indication the performance pad mode.

Separate PR for the manual is also proposed

PR #12842, opened 2024-02-18
Pioneer DDJ-FLX4 mapping improvements

grizeldi:
2 improvements to the built in FLX4 mapping:

1. Controller now properly exits Bluetooth pairing mode while connected to Mixxx.
2. Removed the mapping for the software headphone gain knob, as the knob on the controller already affects hardware gain, which resulted in very sensitive gain knob (as discussed in the last couple messages of #11245).

The behavior of headphone cueing and mixing now seems to be consistent with how FLX4 works with Rekordbox.


JoergAtGithub:
Welcome at Mixxx!

grizeldi:
Done. Apple complicating things for devs again I see :sweat_smile:

res/controllers/Pioneer-DDJ-FLX4-script.js
```diff
@@ -169,6 +169,11 @@ PioneerDDJFLX4.lights = {
 // Store timer IDs
 PioneerDDJFLX4.timers = {};
 
+// Keep alive timer
+PioneerDDJFLX4.sendKeepAlive = function() {
+    midi.sendSysexMsg([0xF0, 0x00, 0x40, 0x05, 0x00, 0x00, 0x04, 0x05, 0x00, 0x50, 0x02, 0xf7], 12);
+};
```
Swiftb0y, line 175: just out of curiosity, how did you find this message? Why is that one the proper one to keep it out of bluetooth mode?

Swiftb0y (commented):
LGTM otherwise. Thank you.

res/controllers/Pioneer-DDJ-FLX4-script.js
grizeldi: In the official MIDI mapping documentation, no sysex messages are documented, even though the controller seems to employ quite a couple. A while ago I captured a bit of traffic between Rekordbox and my FLX4 with Wireshark for another project and this is the one sysex that had a rather clear purpose, as Rekordbox sends it every 200ms and if it isn't sent in time, the controller reverts into BT pairing mode.

Is that a problem? I rarely have to reverse engineer stuff, so I'm not sure how the legality of it works with international projects.

Swiftb0y: I have no idea about the legal aspects, but [usually reverse-engineering is the only solution for Pioneer hardware](https://swiftb0y.github.io/CDJHidProtocol/hid-analysis/) (I'm the author, so :shrug: :wink: ). I was rather just asking whether there was some documentation out there that I didn't know about. 

Can you just add a quick comment next to the function that the sysex was reverse-engineered from packet captures? That way these questions don't crop up in the future. Thank you.

grizeldi: To my knowledge, there's only [this](https://www.pioneerdj.com/-/media/pioneerdj/software-info/controller/ddj-flx4/ddj-flx4_midi_message_list_e1.pdf), which doesn't cover any sysex, sadly. Pioneer hardware/software seems to be a lot less extensible than professional grade software I'm used to from other industries.

I added a comment with the notice.

Swiftb0y (approved):
LGTM. Thank you.

Swiftb0y:
too quick yet again, this should've gone into 2.4... whats the procedure? cherry-pick to backport?

daschuer:
Yes.

Swiftb0y:
thank you. done

PR #12844, opened 2024-02-18
QML Auto-Reload Fixes

Holzhaus:
Some fixes for #12795.

### DIFF
--- a/src/qml/qmlautoreload.cpp
+++ b/src/qml/qmlautoreload.cpp
@@ -22,15 +22,14 @@ QUrl QmlAutoReload::intercept(const QUrl& url, QQmlAbstractUrlInterceptor::DataT
         return url;
     }
     m_fileWatcher.addPath(url.toLocalFile());
+    m_pathTriggeringReload.insert(url.toLocalFile());
     return url;
 }
 
 void QmlAutoReload::slotFileChanged(const QString& changedFile) {
     qDebug() << "File" << changedFile << "used in QML interface has been changed.";
-    // This is to prevent double-reload when a file is updated twice
-    // in a row as part of the normal saving process. See note in
-    // QFileSystemWatcher::fileChanged documentation.
-    if (m_fileWatcher.removePath(changedFile)) {
+    if (m_pathTriggeringReload.remove(changedFile)) {
+        m_fileWatcher.removePath(changedFile);
         emit triggered();
     }
 }
--- a/src/qml/qmlautoreload.cpp
+++ b/src/qml/qmlautoreload.cpp
@@ -36,7 +36,10 @@ void QmlAutoReload::slotFileChanged(const QString& changedFile) {
 }
 
 void QmlAutoReload::clear() {
-    m_fileWatcher.removePaths(m_fileWatcher.files());
+    const auto files = m_fileWatcher.files();
+    if (!files.isEmpty()) {
+        m_fileWatcher.removePaths(files);
+    }
 }
 
 } // namespace qml
--- a/src/qml/qmlapplication.h
+++ b/src/qml/qmlapplication.h
@@ -1,7 +1,6 @@
 #pragma once
 
 #include <QApplication>
-#include <QFileSystemWatcher>
 #include <QQmlApplicationEngine>
 
 #include "coreservices.h"
--- a/src/qml/qmlautoreload.h
+++ b/src/qml/qmlautoreload.h
@@ -26,6 +26,24 @@ class QmlAutoReload : public QObject, public QQmlAbstractUrlInterceptor {
 
   private:
     QFileSystemWatcher m_fileWatcher;
+    // We use a separate set to keep track of file allowed to trigger a reload.
+    // This is to prevent double-reload when a file is updated twice
+    // in a row as part of the normal saving process. See note in
+    // QFileSystemWatcher::fileChanged documentation.
+    // We've identified to saving policy so far:
+    // 1. Append/Truncate (VSCode)
+    //    The program first append the new content to the file (firing a signal,
+    //    while keeping the path in the watch list) then subsequent truncate to
+    //    move the previous file content (firing a second signal, and also
+    //    keeping the path in the watch list)
+    // 2. Delete/Rewrite (sed in-place, vim)
+    //    The program first delete the file (firing a signal, AND removing the
+    //    path from the watch list) then subsequent create/write to the same
+    //    file path (not firing a second signal, since it has been removed from
+    //    the watch path previously)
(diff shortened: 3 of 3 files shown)


src/qml/qmlautoreload.cpp
```diff
@@ -18,25 +18,40 @@ QmlAutoReload::QmlAutoReload() {
...
-    if (m_fileWatcher.removePath(changedFile)) {
-        emit triggered();
-    }
+    // in a row as part of the normal saving process. See note in the
+    // `QFileSystemWatcher::fileChanged()` documentation.
+    //
+    // Unfortunately we cannot be sure that `QFileSystemWatcher::removePath()`
+    // actually returns true, so we need to emit the signal anyway.
+    m_fileWatcher.removePath(changedFile);
+    emit triggered();
```
Swiftb0y, line 47: doesn't this unconditional trigger now result in the double-reload?

Holzhaus: Yes, if the editor writes the file twice.
The idea was to remove the path from the watcher, and once the path is added back (when QML was reloaded) the second write already happened. 

However, `removePath` may return false for many reasons and we cannot rely on it. On my system, if I edit a watched QML file in vim and then save it with `:w` then for some reason the method returns false and no reload is triggered. I don't know why, but this makes autoreload useless for me.

Is double reload really an issue? At worst it's a small optical disturbance and can be expected. On the other hand, the feature is currently broken for me. Does it work for you?

Swiftb0y: No not a problem. I was just questioning my understanding of the code. It no longer matches the comment where it says that it is supposed to prevent the double-reload, so can you remove it? If the double-reload is annoying, we could try to workaround it by implementing some debouncing logic I guess, but we can do that once we deem it to be an actual problem.

acolombier: `m_fileWatcher.removePath` should return `true` for the signal to be emitted. If it return false, it means the file isn't watched anymore, which happens when a signal has already `triggered` a reload. This is the suggested implementation as per the [Qt doc](https://doc.qt.io/qt-6/qfilesystemwatcher.html#fileChanged)

> Note: As a safety measure, many applications save an open file by writing a new file and then deleting the old one. In your slot function, you can check watcher.files().contains(path). If it returns false, check whether the file still exists and then call addPath() to continue watching it.

src/qml/qmlautoreload.cpp
```diff
@@ -18,25 +18,40 @@ QmlAutoReload::QmlAutoReload() {
...
-    if (!url.isLocalFile() || !QFileInfo(url.toLocalFile()).isFile()) {
+    const auto path = url.toLocalFile();
+
+    if (!url.isLocalFile() || !QFileInfo(path).isFile()) {
+        return url;
+    }
+
+    // Do not add path if it's already watched. See note in the
+    // `QFileSystemWatcher::fileChanged()` documentation.
+    if (m_fileWatcher.files().contains(path)) {
```
acolombier, line 29: This is redundant as `[addPath](https://doc.qt.io/qt-6/qfilesystemwatcher.html#addPaths)` we already perform this check.

> Paths are not added if they not exist, or if they are already being monitored by the file system watcher

acolombier:
Would you mind listing the issue you are trying to fix?

src/qml/qmlautoreload.cpp
Holzhaus: > `m_fileWatcher.removePath` should return `true` for the signal to be emitted. If it return false, it means the file isn't watched anymore, which happens when a signal has already `triggered` a reload.

As I explained above, this is not correct. If you look at the documentation of [`removePath`](https://doc.qt.io/qt-6/qfilesystemwatcher.html#removePath):

> Reasons for watch removal failing are generally system-dependent, but may be due to the path having already been deleted, for example.

If I add logging immediately before the `removePath` call, I can see that the slot is called *exactly once* and that it fails:

    warning [Main] m_fileWatcher.removePath "/home/jan/Projects/mixxx/res/qml/Knob.qml" - result false

So if my editor deleted the file when saving and wrote a new one, it's entirely correct that `removePath` returns false because the old file was deleted, even though triggered was never called before. And this is just an example, I have no idea how my editor is implemented and why `removePath` returns false.

src/qml/qmlautoreload.cpp
Holzhaus: Alright, removed.

Holzhaus:
> Would you mind listing the issue you are trying to fix?

Yes, the issue that autorload is not working at all and the QML interface is never reloaded. See here: https://github.com/mixxxdj/mixxx/pull/12844#discussion_r1495092737

src/qml/qmlautoreload.cpp
Holzhaus: > If I add logging immediately before the `removePath` call, I can see that the slot is called _exactly once_ and that `removePath` returns false

This behavior is 100% reproducible with neovim, gedit and `sed -i` which are all text editors that I have installed. Only `touch Knob.qml` will result in a `true` return value. However, with the fixes from this PR it makes no difference. In all cases the QML GUI is reloaded.

Can you give an example of an editor that benefits from checking the return value? What happens if we don't? Won't it still work, just auto-reload twice?

acolombier: > Is double reload an actual problem? Which editors update the file twice in row? If there are any, then it's a small optical disturbance at worst. On the other hand, the feature is currently broken for me. Does it work for you?

I know VS Code relies on this append-delete file edit policy for resilience (whereas vim relies on a swap file along side the edited file if I recall properly?). The disturbance isn't neglectable if you are using some i3-style windows manager. 

I have managed to reproduce it locally to using vim, so I'll see if I can find a way to get both saving policies to work nicely.

acolombier: I have worked on a fix that ensure single reload for both policies [here](https://github.com/mixxxdj/mixxx/compare/main...acolombier:mixxx:fix/qml-auto-reload), could you confirm it works for you?

If yes, perhaps, you can get it in your PR, since you've also added some code cleanup that makes sense to get in (`QmlAutoReload::clear`, var for `url.toLocalFile()`)

acolombier: Have you had a chance to look at the patch above @Holzhaus ?

Holzhaus: yes, works for me.

Holzhaus:
This is ready and can be merged IMHOm

Swiftb0y (approved):
LGTM

Swiftb0y:
Thank you

PR #12847, opened 2024-02-19
SoundSourceMP3: fix unused function warning

Be-ing:


src/sources/soundsourcemp3.cpp
```diff
@@ -109,7 +109,9 @@ inline bool isUnrecoverableError(mad_error error) {
     return (MAD_ERROR_NONE != error) && !MAD_RECOVERABLE(error);
 }
 
-inline bool hasUnrecoverableError(const mad_stream* pMadStream) {
+[[maybe_unused]] // only used in DEBUG_ASSERT
```
daschuer, line 112: Please hide this entirely from the compiler by using `#ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED`
This way we will later get the unused warning if the function becomes truly unused. 
It also self explaining and does not required a comment.

daschuer (changes requested):

daschuer (approved):
LGTM, thank you.

PR #12854, opened 2024-02-20
Handle new " / " from taglib 2.0 

### LINKED ISSUE #12790: Taglib 2.0: Mixxx joins multi value tags due to new " / " separator (loss of user data)
### Bug Description

Taglib 2.0 has the following behavioral change:
* The basic tag methods (e.g. genre()) separate multiple values with " / " instead of " ".

https://github.com/taglib/taglib/releases/tag/v2.0

After playing a track, Mixxx writes back changes to the file metadata. If a multi value tag has been imported with Taglib 1. It is stored in the Mixxx database with the old separator " " but will be returned from taglib 2 with " / ". Mixxx assumes a user change and will write the old " " separator to the file, joining both tags. This must not happen. 

Example: A track has two genres: "Rock" and "Pop". 
Taglib 1 return "Rock Pop" which is likely in the Mixxx database. After update to Taglib 2, it returns "Rock / Pop" and Mixxx corrects it to "Rock Pop" which will create a new "Rock Pop" genre tag instead of keeping two separate genres.     

### Version

_No response_

### OS

_No response_

daschuer:
This fixes #12790 by ignoring the new multi value separator when comparing tags. 
After installing taglib 2.0 this separator will populate the Mixxx database when reading a multi value tag.

After editing such tag, the multi values are deleted and the new value is stored a singe Tag containing the " / " separator. 

This is a workaround to allow to use Taglib 2. Writing full multi value tags is more work.

### DIFF
--- a/src/track/taglib/trackmetadata_common.cpp
+++ b/src/track/taglib/trackmetadata_common.cpp
@@ -274,20 +274,25 @@ void exportTrackMetadataIntoTag(
         WriteTagMask writeMask) {
     DEBUG_ASSERT(pTag); // already validated before
 
-    pTag->setTitle(toTString(trackMetadata.getTrackInfo().getTitle()));
-    pTag->setAlbum(toTString(trackMetadata.getAlbumInfo().getTitle()));
-
     // The mapping of multi-valued fields in TagLib is not bijective.
     // We don't want to overwrite existing values if the corresponding
-    // field has not been modified in Mixxx. This workaround only covers
-    // the most common multi-valued fields.
+    // field has not been modified in Mixxx.
     //
     // See also: <https://github.com/mixxxdj/mixxx/issues/12587>
-    const auto artist = toTString(trackMetadata.getTrackInfo().getArtist());
+
+    const TagLib::String title = toTString(trackMetadata.getTrackInfo().getTitle());
+    if (title != pTag->title()) {
+        pTag->setTitle(title);
+    }
+    const TagLib::String album = toTString(trackMetadata.getAlbumInfo().getTitle());
+    if (album != pTag->album()) {
+        pTag->setAlbum(album);
+    }
+    const TagLib::String artist = toTString(trackMetadata.getTrackInfo().getArtist());
     if (artist != pTag->artist()) {
         pTag->setArtist(artist);
     }
-    const auto genre = toTString(trackMetadata.getTrackInfo().getGenre());
+    const TagLib::String genre = toTString(trackMetadata.getTrackInfo().getGenre());
     if (genre != pTag->genre()) {
         pTag->setGenre(genre);
     }
@@ -297,7 +302,10 @@ void exportTrackMetadataIntoTag(
     // different purposes, e.g. ID3v2. In this case setting the
     // comment here should be omitted.
     if (0 == (writeMask & WriteTagFlag::OmitComment)) {
-        pTag->setComment(toTString(trackMetadata.getTrackInfo().getComment()));
+        const TagLib::String comment = toTString(trackMetadata.getTrackInfo().getComment());
+        if (comment != pTag->comment()) {
+            pTag->setComment(comment);
+        }
     }
 
     // Specialized write functions for tags derived from Taglib::Tag might
--- a/src/track/taglib/trackmetadata_common.cpp
+++ b/src/track/taglib/trackmetadata_common.cpp
@@ -268,6 +268,10 @@ void importTrackMetadataFromTag(
     }
 }
 
+bool isMultiValueTagEqual(const TagLib::String& taglibVal, const QString& mixxxVal) {
+    return toQString(taglibVal) == mixxxVal;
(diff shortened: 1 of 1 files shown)


daschuer:
This need some good testing. @Holzhaus can you have a look since you use Taglib 2.0 in you build environment?

daschuer:
Who has interest to look into this? Is one using Taglib 2.0?

daschuer:
This bugs me a bit. @Holzhaus do you have interest to test this? Who else has a Distribution with taglib 2 to test?

daschuer:
We meed proper testing here. Who as interest? For now I will remove it form the 2.4.2 milestone.

acolombier:
@daschuer what testing can we do on this? Any concrete usecase? If so happy to test, otherwise I guess we could get it in for the beta?

daschuer:
The issue is described here: 
https://github.com/mixxxdj/mixxx/issues/12790

Originally we fail building with Taglib 2 to avoid inconsistent changes to the database. With the 2.5 branch we do not fail anymore. This means without the fix and taglib 2, all multi value tags are rewritten with `/` separator. 
After downgrading to taglib 1 a single tag with the / separator is stored, breaking the multi value tag written by a third party tool. This should not happen (fixed in this PR)

JoergAtGithub:
Could you please add a test for this.

daschuer:
@PPN-SD

The todos here are:
Test this: 
* Create an MP3 with two genres in separate fields. 
* Import it with Mixxx and taglib 1
* Check how the genre is shown.
* Verify that the genre file metadata is not touched when changing the artist.
* Update to taglib 2. 
* Alter artist again 
* Verify that the genre file metadata is not touched 

Now repeate the same with another file but starting with taglib2.


It would be also nice if you can share such a file with two genres and only a sec of sound. This may be used in a unit test.

PPN-SD:
@daschuer I confirm the behaviour : 

Tested with mixxx-2.5.0

**Taglib1:** 
kid3-cli Nachtmusik.flac
set genre[0] Acoustic
set genre[1] Classical

Launch mixxx
Mixxx-library :
  Nachtmusik.flac
  genre shown : Acoustic Classical
  artist : W Mozart

**Taglib2:** 
kid3-cli Nachtmusik-tag2.flac
set genre[0] Acoustic
set genre[1] Classical

Launch mixxx
Mixxx-library :
  Nachtmusik.flac
  genre shown : Acoustic Classical
  artist set : W Mozart

  Nachtmusik_-tag2_.flac
  genre shown : Acoustic **/** Classical

But, if I launch Mixxx patched with a new config (rm -r ~/.mixxx), I have "Acoustic / Classical" for all files.

daschuer:
Does this mean this PR works? Can you confirm that Mixxx from this PR does not overwrite unnecessarily this file metadata, while the artist change is correctly stored? 

```
genre[0] "Acoustic"
genre[1] "Classical"
```

We don't want:  
```
genre[0] "Acoustic / Classical"
genre[1] "Classical"
```
Or 
```
genre[0] "Acoustic Classical"
genre[1] ""
```

PPN-SD:
Even without patch, I never see (does it mean I can't reproduce ? I'll double-check this point - and others too) :

```
genre[0] "Acoustic / Classical"
genre[1] "Classical"
```
or 
```
genre[0] "Acoustic Classical"
genre[1] ""

```

> Does this mean this PR works?

Depend of what you expext.

With a fresh config, mixxx/taglib2/patch-12854 :
I have "Acoustic / Classical" for all files.

Without wiping config :
"Acoustic Classical" for the file already imported before with mixxx/taglib1
"Acoustic / Classical" for the file imported after installation of mixxx/taglib2/patch-12854

BUT, thanks to your questions, I understand a little better now. I'll double-check again all this in the evening.

daschuer:
Can you provide a testing track to my private email? I may use it to set up a unit-test. 

To confirm the issue: 
* Create a copy of your testing track 
* Start Mixxx with taglib 1 and without this PR 
* Enable File tag syncronisation in the Library preferences
* Drop th track to the library 
* Genre will appear as "Acoustic Classical"
* Close Mixxx 
* Start Mixxx with Taglib 2 without this PR 
* Edit the Artist from the track. 
* Mixxx will store the new Artist AND the Genre as "Acoustic Classical" this is wrong and destroys the two genres = user data. This happens because Mixxx sees now "Acoustic / Classical" and considers an edit in Mixxx which has not happen.

acolombier:
Sounds like this is ready to go, thanks for the test! Should I press the merge button @daschuer ?

daschuer:
Yes, than it is part of 2.5.1.
I will do the the unit test and cake changes in a follow up.

JoergAtGithub (approved):
As the test proves, that it works, let's merge this!

PPN-SD:
@JoergAtGithub @daschuer Thanks both !
Could you remove the warning message before merging 2.5 in main ?

Issue #12858, opened 2024-02-21
relinking library directory should automatically rescan library


Be-ing:
### Bug Description

Relinking a library directory in the Library preferences requires manually rescanning the library for Mixxx to find the files. This should be done automatically when saving the changed preferences.

### Version

2.4.0

### OS

Fedora 39

manuelbertelli:
How do I assign this issue to me?

Be-ing:
There's no need to assign the issue, you can just go ahead and open a pull request when you have a fix implemented.

manuelbertelli:
Relinking = relocate dir?

Be-ing:
Yes, using the Relink button in the Library preferences.

ronso0:
Why is a rescan _required_?

IIUC, if the (inner) directory structure didn't change (dirs or tracks renamed / deleted) all tracks should be updated correctly.
or are you referring to the relink after migrating the Mixxx db to another OS / computer / ...? In that case encoders (availability / version) might have changed and rescan might be needed.

ronso0:
Note: relinking was actually broken before #12462 (dirs were update but not the tracks).

Be-ing:
> are you referring to the relink after migrating the Mixxx db to another OS / computer / ...?

I noticed this issue when testing migration into the Flatpak sandbox https://mixxx.org/news/2024-02-21-official-flatpak-now-available-on-flathub/

ronso0:
I saw that, but does that mean we need a full rescan in order to get access to each track?

ronso0:
Okay, I tried this myself:
* installed org.mixxx.Mixxx
* (~/.mixxx exists but is empty)
* copied _mixxx.cfg_ and _mixxxdb.sqlite_ to _~/.var/app/org.mixxx.Mixxx/.mixxx_
* start Mixxx with `flatpak run org.mixxx.Mixxx`
* asks me for a music directory (mixxxdb.sqlite and config are ignored?), I pick one
* close Mixxx
* now I suddenly have _~/.mixxx_
* okay, so I copy _mixxx.cfg_ and _mixxxdb.sqlite_ there
* start Mixxx, old music dirs are shown in Pref -> Library
* relink one that is marked missing: no permission, no advice how to grant it
* close Mixxx, research how to grant file permissions, do so with
`sudo flatpak override org.mixxx.Mixxx --filesystem=~/someMusicDir`
(I also used `Flatseal` for another directory)
* start Mixxx
[...]

PR #12864, opened 2024-02-23
Rescan library after relocating a music directory

manuelbertelli:
I don't know what to put here because the thing was very simple. I've signaled again to scan library when slotUpdate() is triggered.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,3 +66,5 @@ src/mixxx.rc.include
 src/mixxx.res
 src/test/**/*.actual
 res/qrc_mixxx.cc
+/out/build/x64-Debug
+/.vs
--- a/src/preferences/dialog/dlgpreflibrary.cpp
+++ b/src/preferences/dialog/dlgpreflibrary.cpp
@@ -142,10 +142,11 @@ DlgPrefLibrary::~DlgPrefLibrary() = default;
 
 void DlgPrefLibrary::slotShow() {
     m_bAddedDirectory = false;
+    m_bRelocatedDir = false;
 }
 
 void DlgPrefLibrary::slotHide() {
-    if (!m_bAddedDirectory) {
+    if (!m_bAddedDirectory && !m_bRelocatedDir) {
         return;
     }
 
@@ -325,6 +326,11 @@ void DlgPrefLibrary::slotUpdate() {
                     kBpmColumnPrecisionConfigKey,
                     BaseTrackTableModel::kBpmColumnPrecisionDefault);
     spinbox_bpm_precision->setValue(bpmColumnPrecision);
+
+    // Rescan Library if a directory is relocated
+    if (m_bRelocatedDir) {
+        emit scanLibrary();
+    }
 }
 
 void DlgPrefLibrary::slotCancel() {
@@ -420,6 +426,7 @@ void DlgPrefLibrary::slotRelocateDir() {
 
     if (!fd.isEmpty()) {
         emit requestRelocateDir(currentFd, fd);
+        m_bRelocatedDir = true;
         slotUpdate();
     }
 }
--- a/src/preferences/dialog/dlgpreflibrary.h
+++ b/src/preferences/dialog/dlgpreflibrary.h
@@ -75,6 +75,8 @@ class DlgPrefLibrary : public DlgPreferencePage, public Ui::DlgPrefLibraryDlg  {
     UserSettingsPointer m_pConfig;
     std::shared_ptr<Library> m_pLibrary;
     bool m_bAddedDirectory;
+    // Flag if a directory is relocated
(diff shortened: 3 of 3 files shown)


src/preferences/dialog/dlgpreflibrary.cpp
```diff
@@ -142,10 +142,11 @@ DlgPrefLibrary::~DlgPrefLibrary() = default;
 
 void DlgPrefLibrary::slotShow() {
     m_bAddedDirectory = false;
+    m_bRelocatedDir = false;
```
ronso0, line 145: Like `m_bAddedDirectory` this should also be initializedin the constructor, not only on show

.gitignore
```diff
@@ -66,3 +66,5 @@ src/mixxx.rc.include
 src/mixxx.res
 src/test/**/*.actual
 res/qrc_mixxx.cc
+/out/build/x64-Debug
+/.vs
```
ronso0, line 70: Please remove that commit.
Personal ignores belong into your personal gitignore, not in the Mixxx repo.
(we had this discussion too often already, at the end it was always 'No.' ; )

src/preferences/dialog/dlgpreflibrary.cpp
```diff
@@ -325,6 +326,11 @@ void DlgPrefLibrary::slotUpdate() {
                     kBpmColumnPrecisionConfigKey,
                     BaseTrackTableModel::kBpmColumnPrecisionDefault);
     spinbox_bpm_precision->setValue(bpmColumnPrecision);
+
+    // Rescan Library if a directory is relocated
+    if (m_bRelocatedDir) {
+        emit scanLibrary();
```
ronso0, line 332: I suggest to show a dialog like we do when a directory was added.
See slotHide, that needs to be structured a bit and the message text can be a variable which is selected based `m_bRelocatedDir` or `m_bAddedDirectory`.

ronso0 (changes requested):

src/preferences/dialog/dlgpreflibrary.cpp
Be-ing: What is the need for a dialog?

ronso0: The same reason there's a dialog when adding a new directory. The user has to do it at some point, but we don't want to enforce the potentially very long scan NOW.

ronso0:
@Be-ing Please explain why this is necessary in the first place.
The Flatplak migration issue
https://github.com/mixxxdj/mixxx/issues/12858#issuecomment-1961987282 should be solved after relinking, please explain why the full rescan is needed (this PR enforces it for everyone! immediately)

src/preferences/dialog/dlgpreflibrary.cpp
Be-ing: Ah, I get what you're saying. Yeah, a dialog saying that is required that lets the user choose when to rescan would also work. It could have a button to rescan now or close the dialog without rescanning. The problem currently is that it isn't obvious why merely using the relink button doesn't make the tracks appear in the library again.

ronso0: Exactly, and we already have that dialog for added dirs, so we just need to adjust the message text for each case.

I'm curious what's happening under the hood during the migration  (I can't test the Flatpak migration myself atm).
I don't understand why the tracks don't show up.
They should be shown (but inaccessible) initially, and IIUC the relink adjusts the paths/loacations in the db, so they _should_ be accessible then.

Be-ing: Relinking is needed when the pre-Flatpak database referred to files that are outside the Flatpak sandbox. Accessing them inside the sandbox requires going through the XDG File Chooser Portal. The path that Mixxx gets back from the portal is different from its real path on the underlying filesystem. Rather than `/home/user/dir1/dir2`, Mixxx gets back `/run/user/...`.

ronso0: Yes, I got that. So the rescan is required to get _per-track_ access via the sandbox?
I thought access right for the (each) parent directory is sufficient?

Be-ing: Yes, permission for the parent directory seems to be sufficient. But when I tested, the tracks that appear to have been relocated need a rescan to show up in the library.

ronso0:
@manuelbertelli Thanks for working on this.

However, the resan does not seem to be required in general as soon as Mixxx Flatpak is granted access to the dir in question.
See my test result here https://github.com/mixxxdj/mixxx/issues/12858#issuecomment-1963141880
Setting this to draft until the situatin is clarified. Then we should also remove rescan hint from the blog post.

PR #12873, opened 2024-02-26
[WIP] Rekordbox parser Rust impl

christophehenry:
This is an attempt at integrating Rust into Mixxx' toolchain by rewriting Rekordbox parser using [recordcrate](https://holzhaus.github.io/rekordcrate/rekordcrate/). It uses [cxx-qt](https://kdab.github.io/cxx-qt/book/) to integrate Rust code in Qt.

This is an early PR to get review as quick as possible to ensure I'm going in the right direction. Also, I have a few questions.


src/rust/src/cxxqt_object.rs
```diff
@@ -0,0 +1,42 @@
...
+    string: QString,
+}
+
+impl rekordbox::Rekordbox {
+    /// Increment the number Q_PROPERTY
+    pub fn increment_number(self: Pin<&mut Self>) {
+        let previous = *self.number();
+        self.set_number(previous + 1);
+    }
+}
```
christophehenry, line 42: This is currently dumb code for testing.

src/mixxxmainwindow.cpp
```diff
@@ -62,6 +62,8 @@
 #include "vinylcontrol/vinylcontrolmanager.h"
 #endif
 
+#include "cxx-qt-gen/rekordbox.cxxqt.h"
```
christophehenry, line 65: This import in this file doesn't compile (I get no file or directory with this name error) whereas [importing in `main.cpp`](https://github.com/mixxxdj/mixxx/pull/12873/files#diff-34d21af3c614ea3cee120df276c9c4ae95053830d7f1d3deaf009a4625409ad2R25) correctly compiles and execute. Any ideas why?

src/rust/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "mixxx-rustlib"
```
fwcd, line 2: My personal opinion is that it would be better to wrap rekordcrate in a C or C++ interface outside of Mixxx and then link that wrapped library in instead of adding Rust code to Mixxx itself. This would make it easy to gate the Rekordcrate support behind a feature flag too, which IMO is a good idea anyway (we've done so for a number of other libraries, e.g. Rubberband in #12661).

While Rust is a cool language and I'd love to see parts of Mixxx written in it, integrating it into Mixxx's core toolchain opens a huge can of worms:

- Adding a new language with a wildly different ecosystem will necessarily add complexity, both to the codebase and the build system
- Interaction between Rust/C++/Qt involves requires pulling in more layers of dependencies
- Building Mixxx, especially in a cross-compilation scenario (such as our GitHub Actions build targeting ARM macOS or a future iOS version) would now require not just a C++ toolchain, but also a Rust toolchain with the correct target toolchain. This will likely complicate the build setup quite significantly for anything outside of the standard host to host compilation scenario... I'm not sure how the interaction between CMake, vcpkg and Cargo handles this by default.

Keeping the Rust code encapsulated in an out-of-tree C/C++-wrapped library would contain this complexity in a way that's much easier to manage (e.g. cross-building a Rust library packaged up in a vcpkg port should be pretty easy, but hacking together the Mixxx build system to cross-build a mixed Rust/C++ codebase for a new target will always be a non-trivial endeavor.

christophehenry: Ok I get the argument. So I suppose the code should live in mixxxdj/mixxx-rekordboxlib or something similar?

JoergAtGithub: Let's first discuss this with the other devs on Zulip, before spending work on this.

Be-ing:
> Also, can you explain some background why you opted for cxx-qt here? Wouldn't cxx or plain c bindings be preferable for reusability?

Yes, there's no need for QObjects with this, so CXX would be more appropriate than CXX-Qt.

christophehenry:
Ok, then I go for the C bindings. @Holzhaus would you guide me through this? I'm still learning Rust.

Swiftb0y:
while I agree that for single dependencies, it would be preferred to not pull in all of the rust machinery, I think it would make sense in the long term so new functionality within mixxx (smaller than some external library, just little internal stuff) can be written using Rust instead of C++. I think what solution we opt for is a matter of the long-term direction we would want our codebase to go.

christophehenry:
Closing a per Zulip conversation.

PR #12875, opened 2024-02-26
(fix) Tooltips: update 'hotcue' with saved loop features

ronso0:
IIUC this will just insert an untranslated line into the existing tooltip, so can be merged anytime, right?


daschuer (approved):
Thank you. LGTM

Issue #12829, opened 2024-02-16
Traktor F1 mapping failing since update to 2.4


ZenKyma:
### Bug Description

While working correctly in Mixxx 2.3.6, the mapping for Traktor F1 throws an error in 2.4.0:
```
Uncaught exception: file:///usr/share/mixxx/controllers/common-hid-packet-parser.js:413: RangeError: index out of range
Backtrace: pack@file:///usr/share/mixxx/controllers/common-hid-packet-parser.js:413
@file:///usr/share/mixxx/controllers/common-hid-packet-parser.js:1061
fastForIn@file:///usr/share/mixxx/controllers/common-hid-packet-parser.js:2236
@file:///usr/share/mixxx/controllers/common-hid-packet-parser.js:1058
fastForIn@file:///usr/share/mixxx/controllers/common-hid-packet-parser.js:2236
send@file:///usr/share/mixxx/controllers/common-hid-packet-parser.js:1055
@file:///usr/share/mixxx/controllers/Traktor-Kontrol-F1-scripts.js:304
@file:///usr/share/mixxx/controllers/Traktor-Kontrol-F1-scripts.js:335
```
Reproducible in Windows 10 for Mixxx 2.4.0 but not for 2.3.6.

The respective lines from [`Traktor-Kontrol-F1-scripts.js`](https://github.com/mixxxdj/mixxx/blob/main/res/controllers/Traktor-Kontrol-F1-scripts.js) are the following:
```
008 function KontrolF1Controller() {
...
293     // reset all lights to off state
294     this.resetLEDs = function() {
295         var controller = this.controller;
296         var packet = controller.getOutputPacket("lights");
297         for (var group_name in packet.groups) {
298             var group = packet.groups[group_name];
299             for (var field_name in group) {
300                 var field = group[field_name];
301                 field.value = 0;
302             }
303         }
304         packet.send();
305     }
...
313 }
...
335     KontrolF1.resetLEDs();

### Version

2.4.0

### OS

Ubuntu 23.10 & Windows 10

**Comment by JoergAtGithub** [comment]
Does also crash, if you comment out line 335 ?

**Comment by ZenKyma** [comment]
Yes:
```
Uncaught exception: file:///usr/share/mixxx/controllers/Traktor-Kontrol-F1-scripts.js:379: TypeError: Property 'shutdownHardware' of object [object Object] is not a function
[...]

PR #12876, opened 2024-02-26
Fixes for hid-parser and Traktor F1 script

JoergAtGithub:
Closes: #12829
@ZenKyma confirmed the fix with his F1 hardware.

This PR fixes 4 single issues:
- The F1 mapping used addControl instead of addOutput to define the outputs
- The F1 mapping did not taken into account the offset due to the ReportID byte (seems only matter with addOutput)
- The hid-parser optimization with the FastForIn wrapper did not work in one case (this object was not set correct). I reverted this to the last version without the wrapper
- The hid-parser contained an equal condition with a bitmask, where the compared value was by one too high

### DIFF
--- a/res/controllers/common-hid-packet-parser.js
+++ b/res/controllers/common-hid-packet-parser.js
@@ -696,14 +696,14 @@ class HIDPacket {
         field.auto_repeat = undefined;
         field.auto_repeat_interval = undefined;
 
-        const packet_max_value = Math.pow(2, this.packSizes[field.pack] * 8);
+        const packet_max_value = Math.pow(2, this.packSizes[field.pack] * 8) - 1;
         const signed = this.signedPackFormats.includes(field.pack);
         if (signed) {
-            field.min = 0 - (packet_max_value / 2) + 1;
-            field.max = (packet_max_value / 2) - 1;
+            field.min = 0 - ((packet_max_value + 1) / 2) + 1;
+            field.max = ((packet_max_value + 1) / 2) - 1;
         } else {
             field.min = 0;
-            field.max = packet_max_value - 1;
+            field.max = packet_max_value;
         }
 
         if (bitmask === undefined || bitmask === packet_max_value) {
--- a/res/controllers/common-hid-packet-parser.js
+++ b/res/controllers/common-hid-packet-parser.js
@@ -1679,7 +1679,7 @@ class HIDController {
      */
     processIncomingPacket(packet, delta) {
 
-        HIDController.fastForIn(delta, (field_name) => {
+        HIDController.fastForIn(delta, (function(field_name) {
             // @ts-ignore ignoredControlChanges should be defined in the users mapping
             // see EKS-Otus.js for an example
             if (this.ignoredControlChanges !== undefined &&
@@ -1697,8 +1697,7 @@ class HIDController {
             } else {
                 console.warn(`HIDController.processIncomingPacket - Unknown field ${field.name} type ${field.type}`);
             }
-        }
-        );
+        }).bind(this)); // Qt < 6.2.4 : .bind(this) needed because of QTBUG-95677
     }
     /**
      * Get active group for this field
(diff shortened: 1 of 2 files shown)


ZenKyma:
Hardware test passed successfully

res/controllers/common-hid-packet-parser.js
```diff
@@ -1679,7 +1679,12 @@ class HIDController {
      */
     processIncomingPacket(packet, delta) {
 
-        HIDController.fastForIn(delta, (field_name) => {
+        // Fast loop implementation over delta object
+        const fieldKeyArr = Object.keys(delta);
+        let fieldKeyIdx = fieldKeyArr.length;
+        while (fieldKeyIdx--) {
+            const field_name = fieldKeyArr[fieldKeyIdx];
```
Swiftb0y, line 1686: > The hid-parser optimization with the FastForIn wrapper did not work in one case (this object was not set correct). I reverted this to the last version without the wrapper

I think this is related to some qt bug (I don't remember the number, nor could I find the discussion on zulip anymore). But using an function expression and calling `.bind(this)` on it worked as a workaround. Have you tried that?





```suggestion
        HIDController.fastForIn(delta, (function(field_name) {....}).bind(this));
```

Swiftb0y (commented):
I have zero experience with common-hid-packet-parser, but most of the changes seem very sensible.

res/controllers/common-hid-packet-parser.js
ZenKyma: I would gladly hardware test the proposed change but I'm not sure how to properly formulate the function expression here.

Swiftb0y: this would be the exact patch I'm proposing. Are you comfortable enough with git to know how to apply it?

```patch
diff --git a/res/controllers/common-hid-packet-parser.js b/res/controllers/common-hid-packet-parser.js
index 9d1b3cc9c8..ac9ad684ff 100644
--- a/res/controllers/common-hid-packet-parser.js
+++ b/res/controllers/common-hid-packet-parser.js
@@ -1679,11 +1679,7 @@ class HIDController {
      */
     processIncomingPacket(packet, delta) {
 
-        // Fast loop implementation over delta object
-        const fieldKeyArr = Object.keys(delta);
-        let fieldKeyIdx = fieldKeyArr.length;
-        while (fieldKeyIdx--) {
-            const field_name = fieldKeyArr[fieldKeyIdx];
+         HIDController.fastForIn(delta, (function(field_name) {
 
             // @ts-ignore ignoredControlChanges should be defined in the users mapping
             // see EKS-Otus.js for an example
@@ -1702,7 +1698,7 @@ class HIDController {
             } else {
                 console.warn(`HIDController.processIncomingPacket - Unknown field ${field.name} type ${field.type}`);
             }
-        }
+        }).bind(this));
     }
     /**
      * Get active group for this field

```

ZenKyma: Yeah sure, I can confirm that the following implementation works as expected with both the F1 and the X1mkII:
```js
    processIncomingPacket(packet, delta) {

        HIDController.fastForIn(delta, (function(field_name) {

            // @ts-ignore ignoredControlChanges should be defined in the users mapping
            // see EKS-Otus.js for an example
            if (this.ignoredControlChanges !== undefined &&
                    // @ts-ignore
                    this.ignoredControlChanges.indexOf(field_name) !== -1) {
                return; // continue loop - by returning to fastForIn
            }
            const field = delta[field_name];
            if (field.type === "button") {
                // Button/Boolean field
                this.processButton(field);
            } else if (field.type === "control") {
                // Numeric value field
                this.processControl(field);
            } else {
                console.warn(`HIDController.processIncomingPacket - Unknown field ${field.name} type ${field.type}`);
            }
        }).bind(this));
    }
```

Swiftb0y: great. @JoergAtGithub can you implement that (along with a comment referencing https://bugreports.qt.io/browse/QTBUG-95677)? Then we can merge.

Swiftb0y (approved):
LGTM. Thank you.

Issue #12715, opened 2024-02-02
Relink music Directory from Linux to Windows fails


CaMeLCaseError:
### Bug Description

Relinking Music Library paths from a Linux config on a Windows install doesn't work. The linux paths (containing slashes) seem to cause Problems when relinking - I guess it's because C:\ is prepended or the slashes aren't compatible with search and replace.
See attached screenshots for examples.

Details to reproduce: I tried to move my .mixxx settings directory and all my audio files to a windows machine. Both using 2.5 alpha versions:
* Renamed .mixxx to Mixxx and put it into the C:\Users\...\AppData\Local\Mixxx
* Installed Mixxx
* After start went into preferences to relink. (when clicking on "relink" simply nothing visibly happens).
* closed Mixxx
* restarted Mixxx (and retried to relink also)

![Screenshot 2024-02-02 211814](https://github.com/mixxxdj/mixxx/assets/622460/fbf49abb-2790-4037-86ad-f7907b230eb5)

Maybe I have accidentally touched the recording button or mixx just created the recording directory using the linux file structure:
![Screenshot 2024-02-02 212703](https://github.com/mixxxdj/mixxx/assets/622460/663cc156-8901-4edf-9d89-ca4aef44ca8b)

also visible here after relinking and a restart of mixxx the old path is still there. Interestingly C:\ is always only shown on file which are currently selected.
![Screenshot 2024-02-02 214946](https://github.com/mixxxdj/mixxx/assets/622460/1afe01df-3c1d-4262-9662-777f552c7224)

### Version

2.5-alpha-222-gae53865a26 (main)

### OS

Linux, Windows

ronso0:
Do you still have the log file from the first session on Windows when you did the relink?
(Mixxx stores 11 log files, mixxx.log.10 is the oldest)

If no, can you go back to square one (remove/rename the Mixxx config directory in Windows) and relink again?
Then please share the log file. Maybe it contains some hints.

Actually the relinking should have been fixed with #12462 but I'm pretty sure no one tested this cross-OS.

ronso0:
Additionally, (again: simulate a fresh install by removing the Mixxx config directory) could you add one dummy music directory with just some tracks, then move it and try to relink?

CaMeLCaseError:
i will just do it again and save the logs... didn't know they existed until now. and still don't know where...
 
so I'll:
* rename the Mixxx folder to Mixxx_bak
* extract the .mixxx folder from Linux 
[...]

PR #12436, opened 2023-12-15
(fix UX) Library: add feedback to directory operations (add, remove, relink)

ronso0:
This improves user feedback for library directory action in Library preferences (add, relink, remove) and when adding a directory in the Computer feature by showing appropriate message popups.

Fixes #10481

PR #12878, opened 2024-02-26
(fix) relink directory / migrate Linux/macOS <--> Windows

ronso0:
### The issue:
When trying to relink a music directory with a database created in another OS (actually only Windows <--> Linux/macOS),
directory relocation will fail. This is due to the fact that mixxx::FileInfo (QFileInfo), which is part of the roundtrip 
DirectoryDAO -> [Library preferences](https://github.com/mixxxdj/mixxx/blob/23e91ac0bf7c86c5b03dc2d7048396be6de2ad5c/src/preferences/dialog/dlgpreflibrary.cpp#L403-L425) -> ... -> DirectoryDAO, does unexpected things (no explicit documentation?) if the path string it's constructed from doesn't match the current OS' path schema:
* Windows -> Linux:
`C:\some\path` is considered relative and the the current working dir path is prepended
`/mixxx/build/dir/C:/some/path`
* Linux -> Windows
same issue, `/home/Music` is missing "[drive letter]:" prefix and becomes `C:/home/Music` (not found etc.)

For both directions it fails with this DEBUG_ASSERT in [DirectoryDAO::relocateDirectory](https://github.com/mixxxdj/mixxx/blob/474b82a282b50f0c7251e8ba223ae49119560b9f/src/library/dao/directorydao.cpp#L147) and relocation will fail silently.
See https://github.com/mixxxdj/mixxx/issues/12715

For both directions I also hit this [assertion in `mixxx::Fileinfo`](https://github.com/mixxxdj/mixxx/blob/23e91ac0bf7c86c5b03dc2d7048396be6de2ad5c/src/util/fileinfo.h#L108)), probably when trying to load cover arts.

(note #12436 which adds some feedback to directory operations)

___

### Simple fix:
avoid the string -> fileInfo -> string roundtrip and simply use QStrings for the entire display/relocation flow.
Just the 'oldDirectory' test needs to be fixed.
Or, does it? Actually it's now the same path returned by `DirectoryDAO::getRootDirStrings()`, so if it was changed on the roundtrip the query will fail :shrug: 

### Note for testing with a db from another OS:
You need to build with DEBUG_ASSERTIONS_FATAL=OFF, otherwise you'll hit them in fileinfo.h when the library tries to display cover art.
This also prevents building helpful migration tests.

### DIFF
--- a/src/library/dao/directorydao.cpp
+++ b/src/library/dao/directorydao.cpp
@@ -49,6 +49,30 @@ QList<mixxx::FileInfo> DirectoryDAO::loadAllDirectories(
     return allDirs;
 }
 
+QStringList DirectoryDAO::getRootDirStrings() const {
+    DEBUG_ASSERT(m_database.isOpen());
+    const auto statement =
+            QStringLiteral("SELECT %1 FROM %2")
+                    .arg(
+                            kLocationColumn,
+                            kTable);
+    FwdSqlQuery query(
+            m_database,
+            statement);
+    VERIFY_OR_DEBUG_ASSERT(query.execPrepared()) {
+        return {};
+    }
+
+    QStringList allDirs;
+    const auto locationIndex = query.fieldIndex(kLocationColumn);
+    while (query.next()) {
+        const auto locationValue =
+                query.fieldValue(locationIndex).toString();
+        allDirs.append(locationValue);
+    }
+    return allDirs;
+}
+
 DirectoryDAO::AddResult DirectoryDAO::addDirectory(
         const mixxx::FileInfo& newDir) const {
     DEBUG_ASSERT(m_database.isOpen());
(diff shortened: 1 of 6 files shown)


src/library/dao/directorydao.cpp
```diff
@@ -144,7 +168,7 @@ DirectoryDAO::RemoveResult DirectoryDAO::removeDirectory(
 QList<RelocatedTrack> DirectoryDAO::relocateDirectory(
         const QString& oldDirectory,
         const QString& newDirectory) const {
-    DEBUG_ASSERT(oldDirectory == mixxx::FileInfo(oldDirectory).location());
+    // DEBUG_ASSERT(oldDirectory == mixxx::FileInfo(oldDirectory).location());
```
ronso0, line 171: How can we accomlish the same validation, just without QFileInfo?
IIUC this guarantees that the string and location (string) are identical, if not we can check if it's only the prefix that differs and move on in that case.

Anyone has a smart QString trick handy for this?

ronso0: Actually, do we really need an extensive check here?
It's now the same path returned by `DirectoryDAO::getRootDirStrings()`, so if it was changed on the roundtrip the query will fail.

ronso0:
(updated the description with more findings)

src/library/dao/directorydao.cpp
ronso0: I have omitted the rather complex check (with no actual benefit except debug output) and added a comment instead.
Any ideas to improve the situation are welcome.

daschuer (approved):
LGTM, Good catch.

Issue #12877, opened 2024-02-26
Allow custom track select delay


Eve00000:
### Feature Description

Hi,
Is it possible to add a 'setting' to select the time before a selected track gets 'processed' when browsing in the library. 
Now it's 100 ms, better would be 300 ms ([src/widget/wtracktableview.cpp#L113] 
More explanation; <a href="https://mixxx.discourse.group/t/feature-or-something-that-slipt-developers-minds-or/29093">29093[)</a> [Feature or … something that slipt developers’ minds or]
thank you

ronso0:
Just for anyone attempting to work on this: the 100 ms wait is set here
https://github.com/mixxxdj/mixxx/blob/23e91ac0bf7c86c5b03dc2d7048396be6de2ad5c/src/widget/wtracktableview.cpp#L113

I can confirm that when using my controller's Trax knob for scrolling the GUI hangs occasionally. I'll test if it's really the wait period being too short. At least I can confirm that 100 ms is shorter than the pause between Down key presses (when I try to move the selection fast but precise, i.e. not press and hold to use key auto-repeat), and 300 ms works better. The slightly shorter delay until playlist labels and the cover art are updated will definitely not bother me in live situations.

SjxSubham:
Hi @ronso0 ,
I will Solve it... can i go for it?

Eve00000:
Hi SjxSubham,
Thank you for your time but I think you thought it was too easy.
It's not about just updating the 100 to 300 in the file, it's about making it a variable getting a default value (100 / 300) that can be changed in the settings.
So you need to define a var, use it in wtracktableview.cpp and add a menu-settings-"custom track delay"-option where users can select the prefered delay.
Sorry and thank you

ronso0:
Thanks @Eve00000, I have just proposed the same in https://github.com/mixxxdj/mixxx/pull/12881#issuecomment-1966078881
No need to rush this, let's evaluate who's affected and if / how to deal with it.

PR #12881, opened 2024-02-27
Increase track select delay

SjxSubham:
Issue #12877 
Custom track set delay:  updated time delay from `100ms` to `300ms`


ronso0:
Ummm, thanks for that proposal, but just changing the hard-coded value (for every user) is not 'custom' :upside_down_face: 
What I had in mind was rather making it 'customizable' for testing first, by reading the value from mixxx.cfg (i.e. hidden option). Not sure where this should be done, probably DlgPrefLibrary.
And then we usually discuss if / why / how such a GUI change should be implemented.

PR #12882, opened 2024-02-27
(fix) Library sidebar: fix History label update

### LINKED ISSUE #12880: History labels do not expand when renamed
### Bug Description

(I tried to search for this, but I didn't know the right words.)

1. Go into History, right click on one and pick Rename.
2. increase the length of the name (add 10 X's at the end), click Save.

Label shows up with ellipses ("2024-02-25 XX..."), even when the sidebar is wide enough. I don't know if it matters, but I'm on LateNight Classic.

### Version

2.4.0 (HEAD)

### OS

macOS arm64

ronso0:
Fixes #12880


daschuer (approved):
LGTM, Thank you.

PR #12884, opened 2024-02-27
Developer Tools: initially sort controls by group name, ascending

ronso0:
When debugging I find myself always clicking on the group column to sort and have Channel1 shown at the top.


Swiftb0y (approved):
LGTM, thank you. I agree with this sensible default.

PR #12888, opened 2024-02-27
Add support for "x-special/gnome-copied-files" on Linux

daschuer:
This allows to use Crtl+C and Ctrl+V with various Linux file managers

### DIFF
--- a/src/util/clipboard.cpp
+++ b/src/util/clipboard.cpp
@@ -4,6 +4,22 @@
 #include <QClipboard>
 #include <QMimeData>
 
+namespace {
+
+#ifdef __LINUX__
+QByteArray urlsToUtf8(const QList<QUrl>& urls) {
+    QByteArray result;
+    for (const QUrl& url : urls) {
+        result += url.toEncoded() + '\n';
+    }
+    if (!result.isEmpty())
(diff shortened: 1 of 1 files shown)


src/util/clipboard.cpp
```diff
@@ -15,6 +31,11 @@ void Clipboard::add(const QUrl& url) {
 void Clipboard::finish() {
     QMimeData* data = new QMimeData;
     data->setUrls(instance()->m_urls);
+#ifdef __LINUX__
```
1fabi0, line 34: The limitation on linux is a bit confusing here because it depends on the file manager not on the OS

daschuer: It would not break anything to make it available for all targets. But is there a real use case for it?

1fabi0: For me it's not about the real use case, it's more about the difference between the OS targets caused by a file-manager difference that could in theory exist on all platforms

m0dB:
Thanks for this, @daschuer !

daschuer:
Oh Sorry, I see I have pushed directly to your branch without noticing. I need more careful next time.

PR #3406, opened 2020-12-06
Infobar

poelzi:
This PR adds a Infobar into the library to show additional information about track selection and later notifications that are not
well presented on the table view. This is the successor of #1656 

Screenshot_20201206_011916

The infobar has currently 3 widgets:
- crates
- playlists
- history

Each widgets shows buttons for each entry. The widget itself has 3 visibility states:
- collapsed (only the button of the category in a grey symbol)
- visible (the widget has some medium size but grows if necessary)
- extended (widget takes as much space as possible)

## todo
- safe state
- context menu
- add to other skins
- finish styling

For later PRs:
- Add configuration panel to allow full configuration of the status bar
   - maybe DND of elements
- which widgets
- order of widgets
- Add more exotic but useful widgets
  - free space
- crate/playlist colors
- when notification infrastructure lands, show the last message for n seconds

PR #12890, opened 2024-02-28
View menu: add 'Show Auto DJ' action

ronso0:
requested here https://mixxx.zulipchat.com/#narrow/stream/109122-general/topic/AutoDJ.20as.20a.20View.20option

AutoDJ is a special feature that may need immediate attention, i.e. a way to activate it quickly, that's why I think a shortcut seems reasonable.
Screenshot_2024-02-28_13-21-01

#### Which keyboard shortcut?
There are already Ctrl+1..6
Ctrl+7 would be next, but maybe we want to extend the skin controls, and then it would be better to have consecutive shortcuts for toggles ("Show AutoDJ is a _trigger_).
What about Ctrl+9?

### DIFF
--- a/src/library/autodj/autodjfeature.cpp
+++ b/src/library/autodj/autodjfeature.cpp
@@ -23,12 +23,6 @@
 #include "widget/wlibrarysidebar.h"
 
 namespace {
-
-const QString kViewName = QStringLiteral("Auto DJ");
-
-} // namespace
-
-namespace {
 constexpr int kMaxRetrieveAttempts = 3;
 
     int findOrCrateAutoDjPlaylistId(PlaylistDAO& playlistDAO) {
@@ -55,6 +49,7 @@ AutoDJFeature::AutoDJFeature(Library* pLibrary,
           m_pAutoDJProcessor(nullptr),
           m_pSidebarModel(make_parented<TreeItemModel>(this)),
           m_pAutoDJView(nullptr),
+          m_viewName(Library::kAutoDJViewName),
           m_autoDjCratesDao(m_iAutoDJPlaylistId, pLibrary->trackCollectionManager(), m_pConfig) {
     qRegisterMetaType<AutoDJProcessor::AutoDJState>("AutoDJState");
     m_pAutoDJProcessor = new AutoDJProcessor(this,
@@ -152,7 +147,7 @@ void AutoDJFeature::bindLibraryWidget(
             m_pLibrary,
             m_pAutoDJProcessor,
             keyboard);
-    libraryWidget->registerView(kViewName, m_pAutoDJView);
+    libraryWidget->registerView(m_viewName, m_pAutoDJView);
     connect(m_pAutoDJView,
             &DlgAutoDJ::loadTrack,
             this,
@@ -196,7 +191,7 @@ TreeItemModel* AutoDJFeature::sidebarModel() const {
 
 void AutoDJFeature::activate() {
     //qDebug() << "AutoDJFeature::activate()";
-    emit switchToView(kViewName);
+    emit switchToView(m_viewName);
     emit disableSearch();
     emit enableCoverArtDisplay(true);
 }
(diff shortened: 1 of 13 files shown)


ronso0:
have to reconsider this:
IIUC the idea was to quickly jump to AutoDJ when you're way down in the sidebar.
Currently this action selects AutoDJ but also scrolls there.

However, since you (maybe) only want to activate AutoDJ and then continue working in the last view (computer, history, ...), it would be helpful if the position of the sidebar were fixed. Sure, the current view would be deselected, but it would still be easy to pick up where you left off.
With Tracks or other views it could be confusing, but the AutoDJ view is recognizable.

What do you think?

> Maybe adding the other "main" sections would make sense too, e.g. "Show Tracks"?

And we'd have #9912 (somehow) : )

src/library/autodj/autodjfeature.cpp
```diff
@@ -50,55 +44,56 @@ AutoDJFeature::AutoDJFeature(Library* pLibrary,
...
-            &AutoDJFeature::slotCrateChanged);
-
-    // Create context-menu items to allow crates to be added to, and removed
-    // from, the auto-DJ queue.
-    m_pRemoveCrateFromAutoDj = new QAction(tr("Remove Crate as Track Source"), this);
-    connect(m_pRemoveCrateFromAutoDj,
-            &QAction::triggered,
-            this,
-            &AutoDJFeature::slotRemoveCrateFromAutoDj);
+        qRegisterMetaType<AutoDJProcessor::AutoDJState>("AutoDJState");
```
daschuer, line 49: This extra indentation looks wrong. A clang-format bug? Can we workaround it?

ronso0: True. If I revert that clang-format insists on 8 spaces.
No idea how to work around it.

daschuer: Yes, I would have no issue to merge it under clang-format complains. (It should help us, not annoy us)

daschuer:
> However, since you (maybe) only want to activate AutoDJ and then continue working in the last view (computer, history, ...), it 

Correct. I played a bit with it and it reminds me to taped browsing. Search for something in the Library, verify the Auto-DJ playlist. 
Continue searching. 

We may emulate that by adding tabs to the sidebar ... Or a back button.

The ultimate solution "Open Auto DJ in a new Window" We have discussed the idea earlier to have a skin with two windows. 

Or even allow to have two or more tabs on the library. This will allow to switch between any two contexts without headaches.

ronso0:
Yes, these are nice ideas and part of the library redesign discussion.
I'm not going ot work on it : )

Here, I think we are good if we simply add something like `SidebarModel::selectFeatureNoScroll(LibraryFeature* pFeature)`. Though this will inevtably be mostly redundant with the existing `selectFeature`, hmm.

vespadj:
My suggestion
AutoDJ in Tray bar

ronso0:
Thank you, having a (additional*) topbar toggle is a good idea!
_*Shade skin has no top bar, so we should always provide the action in the View menu_

Note: moving widgets out of the top bar is out of scope here.
There is WIP regarding a global status bar though #3406

src/library/autodj/autodjfeature.cpp
ronso0: reverted

ronso0:
I think I will pick parts of this to fix #13512

src/mixxxmainwindow.cpp
```diff
@@ -450,6 +450,12 @@ void MixxxMainWindow::initialize() {
     if (CmdlineArgs::Instance().getStartAutoDJ()) {
         qDebug("Enabling Auto DJ from CLI flag.");
         ControlObject::set(ConfigKey("[AutoDJ]", "enabled"), 1.0);
+        // Switch to Auto DJ feature
+        auto pLibrary = m_pCoreServices->getLibrary().get();
```
daschuer:
```suggestion
        auto* pLibrary = m_pCoreServices->getLibrary().get();
```

daschuer (changes requested):
Works good. Thank you. One clazy hint.

daschuer:
Thank you. The CI Failure is unrelated.

daschuer (approved):
LGTM

PR #12893, opened 2024-02-28
(fix) Computer feature: update removable devices on Linux

### LINKED ISSUE #12891: 'Computer' sidebar is not updated correctly when removeable drives are un/mounted
### Bug Description

here and below https://mixxx.zulipchat.com/#narrow/stream/109122-general/topic/linux.3A.20mounting.20drive.20after.20starting.20mixxx/near/423779346

Seems to be a model update issue:
When I unmount the drive, the child is removed but the 'user' item (parent) still has the '>' icon indicating children.
Explicit repaint doesn't fix it

And some extra debug stuff tells me Mixxx _does_ see the newly mounted drive directory.
But it's currently not able to expand the item, or doesn't consider it expandable in the first place.

Issue occurs with 2.4

### Version

_No response_

### OS

_No response_

ronso0:
Re #12891 

### Issue 1
I didn't see my SD card in /media/user after I re-mounted it.
I have been focused only on `/media/[user]` because this is where my USB/SD drives are mounted. Devices changes (or removed and re-mounted devices) where not reflected in the tree view because these are **subdirs** of the device lookup path /media for which the respective **'hasChildren()` state has been cached**.

This is not the case for 1st-level children of the lookup paths, e.g. /run/media/[user]. In those we show devices correctly (newly added or removed) when (re-)expanding the 'Removable Devices' node.

**Fix:**
add /media/[user] to lookup paths, but don't show it when adding children of /media :heavy_check_mark: 
= all devices are now leaves of the Devices branch

### Issue 2
We won't see changes in the dir tree after initial expand due to `hasChildren()` caching.
While for static mounts this may be an okayish tradeoff for (assumed) faster repaint for static mounts, it is much more likely that the dir tree of a removable device has changed.

**Ideas and simple benchmarking:**
**a)** when expanding Devices, tell the FolderTreeitemModel to remove all child paths from the cache:
with 10.000 cached paths this takes about 2.5ms, and is only done once when [devices] is expanded
**b)** in `directoryHasChildren`, check if `path` starts with one of the device root paths, if yes: don't cache `hasChildren`
with 10.000 cached paths this takes about 7ms **per dir**, and causes some notable delay under some circumstances

**Fix:**
So the winner is a) :heavy_check_mark: 
To refresh, just collapse & expand 'Removable Devices' (click on branch icon or hit Return twice).

### Addon / Proposal
We could add a new Browse  item action 'Refresh directory tree', mapped to the new method `FolderTreeModel::removeChildDirsFromCache(QStringList)`.
This would allow to force-refresh the dir tree of that item, e.g. when you copied dirs to a previously empty QuickLink dir.
#12908 

**TODO**
make Refresh more discoverable (tooltip), mention it in the manual
___
Related meta discussion: https://mixxx.zulipchat.com/#narrow/stream/109122-general/topic/linux.3A.20mounting.20drive.20after.20starting.20mixxx

### DIFF
--- a/src/library/browse/browsefeature.cpp
+++ b/src/library/browse/browsefeature.cpp
@@ -277,10 +277,6 @@ void BrowseFeature::activateChild(const QModelIndex& index) {
 
 void BrowseFeature::onRightClickChild(const QPoint& globalPos, const QModelIndex& index) {
     TreeItem *item = static_cast<TreeItem*>(index.internalPointer());
-
-    // Make sure that this is reset when the related TreeItem is deleted.
-    m_pLastRightClickedItem = item;
-
     if (!item) {
         return;
     }
@@ -290,6 +286,9 @@ void BrowseFeature::onRightClickChild(const QPoint& globalPos, const QModelIndex
     if (path == QUICK_LINK_NODE || path == DEVICE_NODE) {
         return;
     }
+
+    // Make sure that this is reset when TreeItems are deleted in onLazyChildExpandation()
+    m_pLastRightClickedItem = item;
 
     QMenu menu(m_pSidebarWidget);
 
--- a/src/library/browse/browsefeature.cpp
+++ b/src/library/browse/browsefeature.cpp
@@ -414,6 +414,13 @@ void BrowseFeature::onLazyChildExpandation(const QModelIndex& index) {
 
     // If we are on the special device node
     if (path == DEVICE_NODE) {
+#if defined(__LINUX__)
+        // Tell the model to remove the cached 'hasChildren' states of all sub-
+        // directories when we expand the Device node.
+        // This ensures we show the real dir tree. This is relevant when devices
+        // were unmounted, changed and mounted again.
+        m_pSidebarModel->removeChildDirsFromCache(removableDriveRootPaths());
+#endif
         folders = createRemovableDevices();
     } else {
         // we assume that the path refers to a folder in the file system
--- a/src/library/browse/browsefeature.cpp
+++ b/src/library/browse/browsefeature.cpp
@@ -19,6 +19,8 @@
 
 namespace {
 
+const QString kViewName = QStringLiteral("BROWSEHOME");
+
 const QString kQuickLinksSeparator = QStringLiteral("-+-");
 
 } // anonymous namespace
@@ -79,7 +81,7 @@ BrowseFeature::BrowseFeature(
     // show drive letters
     QFileInfoList drives = QDir::drives();
     // show drive letters
-    foreach (QFileInfo drive, drives) {
+    for (const QFileInfo& drive : std::as_const(drives)) {
         // Using drive.filePath() to get path to display instead of drive.canonicalPath()
         // as it delay the startup too much if there is a network share mounted
         // (drive letter assigned) but unavailable
@@ -123,7 +125,7 @@ BrowseFeature::BrowseFeature(
 
     loadQuickLinks();
 
-    foreach (QString quickLinkPath, m_quickLinkList) {
+    for (const QString& quickLinkPath : std::as_const(m_quickLinkList)) {
         QString name = extractNameFromPath(quickLinkPath);
         qDebug() << "Appending Quick Link: " << name << "---" << quickLinkPath;
         m_pQuickLinkItem->appendChild(name, quickLinkPath);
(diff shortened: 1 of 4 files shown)


ronso0:
Whoops, not working yet, 'optimized' the actual fix away x)

src/library/browse/browsefeature.cpp
```diff
@@ -21,6 +21,13 @@ namespace {
 
 const QString kQuickLinksSeparator = QStringLiteral("-+-");
 
+const QStringList linuxRemovableDrivePaths() {
+    QStringList paths;
+    paths.append("/media");
+    paths.append(QStringLiteral("/run/media/") + QString::fromLocal8Bit(qgetenv("USER")));
```
acolombier, line 27: It should be the same for `/media ` on Ubuntu/Pop-Os - `/media/$USER`

ronso0: I know, it's like that on Ubuntu (which Pop!OS is based on).
Though TLDP https://tldp.org/LDP/Linux-Filesystem-Hierarchy/html/media.html says (can't tell how up-to-date that is, and which distros have their own rules, but /media seems to be the common base):
> The following directories, or symbolic links to directories, must be in /media, if the corresponding subsystem is installed:
>
>floppy     Floppy drive (optional)
cdrom      CD-ROM drive (optional)
cdrecorder CD writer (optional)
zip        Zip drive (optional)
>
>On systems where more than one device exists for mounting a certain type of
media, mount directories can be created by appending a digit to the name of
those available above starting with '0', but the unqualified name must also
exist.
>
> A compliant implementation with two CDROM drives might have /media/cdrom0
and /media/cdrom1 with /media/cdrom a symlink to either of these.

ronso0:
I also added `/Volumes/` for macSO (iOS?).
Someone on macOS needs to test this (and if currently the update issue exists there, too).

src/library/browse/foldertreemodel.cpp
```diff
@@ -48,26 +46,33 @@ bool FolderTreeModel::hasChildren(const QModelIndex& parent) const {
...
-    auto it = m_directoryCache.constFind(path);
-    if (it != m_directoryCache.constEnd()) {
-        return it.value();
+    bool ignore;
+    for (const auto& ignorePath : std::as_const(m_cacheIgnoreList)) {
+        if (path.startsWith(ignorePath)) {
+            ignore = true;
+            break;
+        }
+    }
```
ronso0, line 55: Apparently the cache has been  implemented to boost performance for dirs with many children QStringList and startsWith() _may_ not be the best choice here.

I'll look into [QStringMatcher](https://doc.qt.io/qt-6/qstringmatcher.html) and check for oher performant solutions.

src/library/browse/foldertreemodel.cpp
```diff
@@ -48,26 +46,33 @@ bool FolderTreeModel::hasChildren(const QModelIndex& parent) const {
 }
 
 bool FolderTreeModel::directoryHasChildren(const QString& path) const {
-    auto it = m_directoryCache.constFind(path);
-    if (it != m_directoryCache.constEnd()) {
-        return it.value();
+    bool ignore;
```
ronso0, line 49: not initialised local var...

ronso0:
I must admit that I didn't fully understand the Removable Drives node (or Volumes) and will revise this fix.
I've updated the description with my findings and proposals.

ronso0:
> Oh, one issue with moving [$USER] subdirs to the top level:
> even with no devices `directoryHasChildren` returns true for Removable Devices (▶ icon) because it finds empty [$USER] but BrowseFeature doesn't display this directory anymore, so expanding the branch gives no results.

Okay, this is not a regression, ignored for now.

acolombier:
Tested and working as expected! 

I have some custom mount points configured with udisk which aren't shown in Mixxx but this is expected due to the static path lists in `removableDriveRootPaths()`.

(Not for this PR but for a next iteration) - what do you think of using `/proc/mouns` to detect removable devices on Linux? Paired with a partition type filter + `QFileInfo::isReadable`, I think it could work a treat and support my usecase, as well as those distros that comes with different mounting path.

ronso0:
Sure, might work. `df` might also work.
Though I'll leave further tweaking to others ; )

acolombier (approved):

daschuer:
Thank you @ronso0 for the PR and @acolombier for testing!

ronso0:
@acolombier Wanna file an issue re: `/proc/mounts`/`df`?

PR #12895, opened 2024-02-29
Tracks: custom color for missing tracks

ronso0:
Fix (first commit) and follow-up for #12744 
Lints #9365 

Default color is red.
Pink in LateNight PaleMoon.

The text color might conflict with track colors. If both are similiar tracks would look like if they had empty tags.
The QBrush returned for the Qt::ForeGroundRole may as well be a QLinearGradient. Maybe too much magic?

### TODO
- [ ] set colors in all skins (or keep default `red`?)
- [x] set override for **Missing** feature in all skins

### DIFF
--- a/res/skins/Deere/style.qss
+++ b/res/skins/Deere/style.qss
@@ -225,6 +225,11 @@ WLibrarySidebar {
   /* background of Color delegate in selected row */
   selection-background-color: #006596;
 }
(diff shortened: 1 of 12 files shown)


ronso0:
We may allow picking a custo color with QColorDialog::getColor(), like used in the Mixxx color palette editor.

daschuer (approved):
LGTM and works. Thank you.

PR #12900, opened 2024-03-01
Test Nowplaying & DeckStatus

Eve00000:
This is my first PR test -> Test Nowplaying & DeckStatus (Artist / Title / TimeLoaded)
settingsdir uses the right var
AutoDJ Centers when being disabled


ronso0:
Your Mixxx fork runs the same CI hooks and will also produce builds this way, so if you don't intend to get a branch merged into Mixxx please use open PRs against your own repo and help keeping the list of open PRs clean.

Thank you!

src/mixer/basetrackplayer.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <QMessageBox>
 #include <QMetaMethod>
+#include <QString>
```
ronso0, line 5: this is not required, this class already uses QString in many places

src/mixer/basetrackplayer.cpp
```diff
@@ -626,7 +627,36 @@ void BaseTrackPlayerImpl::slotTrackLoaded(TrackPointer pNewTrack,
...
+    //  Different file for each Deck / Sampler
+        //QMessageBox::information(nullptr /*or parent*/, "Title", QString("EveStatusFileLocation2: %1").arg(EveStatusFileLocation2)); //
+    QString DeckStatusTxtLine1 = "var TrackDeck" + trackInfoDeck + " = { ";
+    QString DeckStatusTxtLine2 = "Artist : \"" + trackInfoArtist + "\",";
+    QString DeckStatusTxtLine3 = "Title : \"" + trackInfoTitle + "\",";
+    QString DeckStatusTxtLine4 = "};";
+       
+    QFile DeckStatusFile(DeckStatusFileLocation);
+    DeckStatusFile.remove();
+    DeckStatusFile.open(QIODevice::ReadWrite | QIODevice::Append);
```
ronso0, line 652: No need for `QIODevice::Append` IMO if you just removed the file, i.e. are creating a new file here.

In general, I think you're better off collecting the data stream first, then write in one step.
Example is writing effects.xml in `EffectsManager::saveEffectsXml()`.

Also, it would be good to clean up, for example when ejecting a track (clear the file, I think this is done somewhere above when loading a null track) and when the player is deleted (remove the file, BaseTrackPlayerImpl::~BaseTrackPlayerImpl). The latter would require to define the filepath in the constructor.

ronso0 (commented):
Just some hints

src/mixer/basetrackplayer.cpp
Eve00000: OK, thank you

src/mixer/basetrackplayer.cpp
Eve00000: Some explication:
-> The 4 "Status[Channelx].js" files are linked in the controller xml, when they are not found at startup Mixxx won't link them when they are filled.
            <file filename="Denon DN-HC4500 4Decks-Scrolling (20240316) H.js" functionprefix=""/>
                -> Header file + declaring the vars and tables to fill with the statusfiles
            <file filename="Status[Channel1].js" functionprefix=""/>
            <file filename="Status[Channel2].js" functionprefix=""/>
            <file filename="Status[Channel3].js" functionprefix=""/>
            <file filename="Status[Channel4].js" functionprefix=""/>
            <file filename="Denon DN-HC4500 4Decks-Scrolling (20240316).midi.js" functionprefix="DenonDNHC4500"/>
- the line  DeckStatusTxtLine4 = "Time : \"" + DeckStatusTime + "\",";
-> When a deck changes the "Status[Channelx].js" Mixxx reloads the whole script, with 4 decks this means that when deck 3/4 are loaded the controller jumps back to 1/2. In my javascript I check the time in the "Status[Channelx].js".
When time in "Status[Channel1].js" < time in "Status[Channel3].js" this means Deck 3 is last manipulatef (load/eject) so the controller has to 'toggle' the deck to control that deck. Else ejecting/loading a track is frustrating.

-> So removing the files is no good, the files have to stay, especialy for the time in it.

Eve00000: I have problems putting code here.... lines starting with "<" can't be displayed

src/mixer/basetrackplayer.cpp
Eve00000: I thought 'resolve' meant see whole conversation.....  sorry

src/mixer/basetrackplayer.cpp
ronso0: just use code formatting: 3 x ` at new line before and after code block

Eve00000: this is not the right repo, I don't know how this occured... it should be "Nowplaying+DeckStatus+AutoDJCenterXfader " https://github.com/mixxxdj/mixxx/commit/255cb774d9152f2a6f965b17c27294c1220fa0be

PR #12903, opened 2024-03-01
require a minimum movement before initiating the drag&drop of tracks

### LINKED ISSUE #12902: Clicking to select a track often initiates a drag&drop
### Bug Description

Often when I click on a track to select it, this triggers the initiation of a drag&drop. This happens if in the brief time between mouse press and mouse release the mouse is moved only minimally, which is very common, especially when using a track pad. Though nothing happens with the drag&drop, it is strange and confusing and could be easily avoided. I will create a PR with a (very simple) fix.

### Version

2.4, main

### OS

macOS

m0dB:
fixes #12902

### DIFF
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -577,13 +577,11 @@ void WTrackTableView::mouseMoveEvent(QMouseEvent* pEvent) {
     }
     //qDebug() << "MouseMoveEvent";
     // Iterate over selected rows and append each item's location url to a list.
-    QList<QString> locations;
-    const QModelIndexList indices = getSelectedRows();
 
     if (DragAndDropHelper::mouseMoveInitiatesDrag(pEvent)) {
         // Iterate over selected rows and append each item's location url to a list.
         QList<QString> locations;
-        const QModelIndexList indices = selectionModel()->selectedRows();
+        const QModelIndexList indices = getSelectedRows();
 
         for (const QModelIndex& index : indices) {
             if (!index.isValid()) {
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -554,6 +554,11 @@ void WTrackTableView::onSearch(const QString& text) {
 void WTrackTableView::onShow() {
 }
 
+void WTrackTableView::mousePressEvent(QMouseEvent* pEvent) {
+    DragAndDropHelper::mousePressed(pEvent);
+    WLibraryTableView::mousePressEvent(pEvent);
+}
+
 void WTrackTableView::mouseMoveEvent(QMouseEvent* pEvent) {
     // Only use this for drag and drop if the LeftButton is pressed we need to
     // check for this because mousetracking is activated and this function is
@@ -575,13 +580,19 @@ void WTrackTableView::mouseMoveEvent(QMouseEvent* pEvent) {
     QList<QString> locations;
     const QModelIndexList indices = getSelectedRows();
 
-    for (const QModelIndex& index : indices) {
-        if (!index.isValid()) {
-            continue;
+    if (DragAndDropHelper::mouseMoveInitiatesDrag(pEvent)) {
+        // Iterate over selected rows and append each item's location url to a list.
+        QList<QString> locations;
+        const QModelIndexList indices = selectionModel()->selectedRows();
+
+        for (const QModelIndex& index : indices) {
+            if (!index.isValid()) {
+                continue;
+            }
+            locations.append(trackModel->getTrackLocation(index));
         }
-        locations.append(trackModel->getTrackLocation(index));
+        DragAndDropHelper::dragTrackLocations(locations, this, "library");
     }
-    DragAndDropHelper::dragTrackLocations(locations, this, "library");
 }
 
 // Drag enter event, happens when a dragged item hovers over the track table view
--- a/src/util/dnd.cpp
+++ b/src/util/dnd.cpp
@@ -209,7 +209,9 @@ bool DragAndDropHelper::allowDeckCloneAttempt(
 
 // static
 void DragAndDropHelper::mousePressed(QMouseEvent* pEvent) {
-    mouseMoveInitiatesDragHelper(pEvent, true);
+    if (pEvent->button() == Qt::LeftButton) {
+        mouseMoveInitiatesDragHelper(pEvent, true);
+    }
 }
 
 // static
--- a/src/util/dnd.cpp
+++ b/src/util/dnd.cpp
@@ -100,6 +100,37 @@ bool allowLoadToPlayer(
     return allowLoadTrackIntoPlayingDeck;
 }
 
+// Helper function for DragAndDropHelper::mousePressed and DragAndDropHelper::mouseMoveInitiatesDrag
+bool mouseMoveInitiatesDragHelper(QMouseEvent* pEvent, bool isPress) {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+    const qreal x = pEvent->position().x();
+    const qreal y = pEvent->position().y();
+#else
+    const qreal x = pEvent->x();
+    const qreal y = pEvent->y();
+#endif
+
+    static qreal pressX{};
+    static qreal pressY{};
+
+    if (isPress) {
+        pressX = x;
+        pressY = y;
+        return false;
+    }
+
+    const qreal threshold = static_cast<qreal>(QApplication::startDragDistance());
+    // X and Y distance since press
+    const qreal dx = x - pressX;
+    const qreal dy = y - pressY;
+
+    // Check if the distance surpasses the threshold. Per Pythagoras Theorem:
+    // distance = sqrt(dx^2 + dy^2)
+    // We can avoid the sqrt by squaring the threshold.
+
+    return dx * dx + dy * dy >= threshold * threshold;
+}
+
 } // anonymous namespace
 
 //static
@@ -174,6 +205,16 @@ bool DragAndDropHelper::allowDeckCloneAttempt(
     }
 
     return true;
+}
+
+// static
+void DragAndDropHelper::mousePressed(QMouseEvent* pEvent) {
+    mouseMoveInitiatesDragHelper(pEvent, true);
+}
+
+// static
+bool DragAndDropHelper::mouseMoveInitiatesDrag(QMouseEvent* pEvent) {
+    return mouseMoveInitiatesDragHelper(pEvent, false);
 }
 
 //static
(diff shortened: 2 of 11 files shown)


src/widget/wtracktableview.cpp
```diff
@@ -565,6 +570,14 @@ void WTrackTableView::mouseMoveEvent(QMouseEvent* pEvent) {
     if (!trackModel) {
         return;
     }
+
+    const auto delta = m_mousePressPosition - pEvent->position();
```
daschuer, line 574: The type of auto is out of sight here. Can you make it explicit?

src/widget/wtracktableview.cpp
```diff
@@ -565,6 +570,14 @@ void WTrackTableView::mouseMoveEvent(QMouseEvent* pEvent) {
...
+    const qreal threshold = 5;
+    if (delta.x() * delta.x() + delta.y() * delta.y() < threshold * threshold) {
```
daschuer, line 576: It took me a moment to understand this formula. Can you add comment about  the used "Pythagorean Theorem"

daschuer (changes requested):
Thank you.

ronso0:
cross-posting from the issue:
Isn't this a system setting, "drag-and-drop threshold"?

Did you check if this also happens with track label widgets?

m0dB:
> Isn't this a system setting, "drag-and-drop threshold"?

No, or if there is such a setting, Qt doesn't use it. From the [Qt drag&drop documentation](https://doc.qt.io/qt-6/dnd.html):

> Dragging
>
> To start a drag, create a [QDrag](https://doc.qt.io/qt-6/qdrag.html) object, and call its exec() function. In most applications, it is a good idea to begin a drag and drop operation only after a mouse button has been pressed and the cursor has been moved a certain distance. However, the simplest way to enable dragging from a widget is to reimplement the widget's [mousePressEvent](https://doc.qt.io/qt-6/qwidget.html#mousePressEvent)() and start a drag and drop operation

So basically this PR does what the Qt documentation suggests.

> Did you check if this also happens with track label widgets?

Not sure which widgets this refers to.

src/widget/wtracktableview.cpp
m0dB: done

src/widget/wtracktableview.cpp
m0dB: I added an std::sqrt and an explicitly named variabled to make things more obvious.

ronso0:
> Not sure which widgets this refers to.

I refer to the Track labels in decks.

ronso0:
I'm just curious why no one has noticed this earlier..

ronso0 (changes requested):
I'm blocking this because I simply don't believe we have to resolve this on app level and shouldn't rush this because
a) there are ways to adjust the drag'n'drop threshold in the OS / DE / driver level
b) how come there were no issue filed in the past ten+ years

Let's please thoroughly investigate if this is indeed a Mixxx issue.

m0dB:
The Qt document clearly states how to implement this, and I implement it as stated, so I don't understand the pushback. Also, I see no drawback adding this. On macOS there is not system setting for this threshold.

I am sure nobody complained because it is merely a small visual nuisance, it doesn't do anything. But it has been annoying me a lot.

JoergAtGithub:
I just tested plain Mixxx 2.4 on Windows, and I could move the mouse pointer very slightly (~ 1 pixel on non-HiDPI display) before the pointer changed to the drag and drop icon.
Maybe this is paltform specific code in Qt and only macOS is affected, or the threshold is one pixel, which is too low for a Retina display?

m0dB:
No, it's not platform specific. It is just implemented this way in Mixxx: Drag&Drop starts on the very first mouse movement (so yes, the absolute minimal movement) with mouse button pressed. In other words, the most naive implementation, and the Qt documentation suggests to do something (marginally) more sophisticated, which is what this PR does.

ronso0:
Sorry, I don't mean to reject / block this in general (and my 'block' can be ignored / overruled if the situation is clarified) and I appreciate you take care of such issues.

FWIW in xfce and any *buntu distro / DE I have used I could configure the d'n'd threshold. In xfce the default is 8 px.
@m0dB @JoergAtGithub Can you verify _all_ widgets are affected, and that the issue is not specific to the tracks view?
If yes we should take of it on the app level (if that's possible).

m0dB:
There are four locations where the drag can be initiated:

```
widget/wtracktableview.cpp
widget/wcoverart.cpp
widget/wtrackproperty.cpp
widget/wtrackwidgetgroup.cpp
```

All of them use the most naive implementation (without threshold). While I think it is most critical for the table (where selecting tracks without dragging them is more common), I agree that it would be best to implement this centralised (in ```util/dnd.cpp``` I guess) and have all widgets use the same threshold mechanism.

m0dB:
Ah, and one more: ```./widget/weffectparameternamebase.cpp``` , which is even more basic: initiate drag on mouse press, doesn't even wait for a minimum move.

src/widget/wtracktableview.cpp
```diff
@@ -565,6 +570,15 @@ void WTrackTableView::mouseMoveEvent(QMouseEvent* pEvent) {
     if (!trackModel) {
         return;
     }
+
+    const QPointF delta = m_mousePressPosition - pEvent->position();
+    const qreal threshold = 5;
+    const qreal distanceSinceMousePress = std::sqrt(delta.x() * delta.x() + delta.y() * delta.y());
```
ronso0, line 576: there's a shorthand [QPointF::manhattanLength()](https://doc.qt.io/qt-6/qpointf.html#manhattanLength)

ronso0:
Ah okay, thank you for the explanation.

QGuiApplication::styleHints() can hold the dnd threshold.
https://doc.qt.io/qt-6/qstylehints.html#startDragDistance-prop
I added this to main.cpp right after the GUI application has been created and it reports `10` even though I have set `8` in my mouse/touchpad settings, so it seems this is also not able to pick the OS settings. But at least we can use it to set 'our' global value without having to use a config value.

ronso0:
@m0dB Could you check what `QApplication::startDragDistance()` reports on macOS?

IMO a higher threshold of 10px would improve the situation on touch (in case they don't have a overriding dnd threshold), I think that's about half the default library row height.

src/widget/wtracktableview.cpp
JoergAtGithub: According to the documentation, the manhattan length is the sum of x and y. Here we need the hypotenuse.

ronso0: Oh, sorry for the noise :roll_eyes:, I was confused by that snippet in https://doc.qt.io/qt-6/qpoint.html#manhattanLength
> This is a useful, and quick to calculate, approximation to the true length:
> 
> > double trueLength = std::sqrt(std::pow(x(), 2) + std::pow(y(), 2));

src/widget/wtracktableview.cpp
```diff
@@ -549,6 +549,11 @@ void WTrackTableView::onSearch(const QString& text) {
 void WTrackTableView::onShow() {
 }
 
+void WTrackTableView::mousePressEvent(QMouseEvent* pEvent) {
+    m_mousePressPosition = pEvent->position();
```
daschuer, line 553: position() is Qt6 only. For my understanding we like to keep 2.5 build-able to check for Qt6 regressions. 
localPos( )is the Qt5 equivalent

daschuer (changes requested):
I can confirm that the PR does what is should. I have lets two comments.

daschuer:
I have just read the https://doc.qt.io/qt-6/qapplication.html#startDragDistance-prop
and it looks like that the recommended solution is a different one compared to the proposed one here.

daschuer:
It is initialized from `gtk-dnd-drag-threshold` or 10 on other platforms 

See also: 
https://github.com/qt/qtbase/blob/269919fa1ea23513a8e36912d06f413e3baefa5d/src/corelib/tools/qpoint.h#L323

daschuer:
Conclusion: Let's follow the Qt's recommendation.

ronso0:
I think @m0dB and @JoergAtGithub are right to use the hypothenuse, also recommeded by Qt (except they use `pwr()`).
> double trueLength = std::sqrt(std::pow(x(), 2) + std::pow(y(), 2));

BUT `manhattanLength` is much faster and for small distance this doesn't make much of a difference. 
I don't care which way we pick, we just need to increase the threshold if ``manhattanLength`` is used.

m0dB:
Given the amount of sqrt calls used to draw the waveform, this single sqrt call in a mouse event handler is not going to be a bottle neck :-). I do not like using manhattanLength for this purpose at all. It might make sense in a situation where you are dragging either virtually or horizontally, but here you might be dragging to any directions. Calculation this distance is the way to go.

I am okay reverting to my original a^2+b^2 < c^2 implementation with a comment. It might even be faster than using manhattanLength (a single comparison, no abs)!

So my way forward, apart from this: have a single implementation (probably inside dnd.cpp) to check if the drag has to be started, for all 4 use cases, and compare the distance with the threshold as reported by QApplication::startDragDistance()

m0dB:
I have rebased on main (since the implementation changed anyway), moved the threshold testing to dnd.cpp (using QApplication::startDragDistance() for the threshold), and now use this for all 4 situations where a drag&drop can be initiated. Additionally a bit of cleanup, renaming event to pEvent.

I use the hypothenuse for the distance, but squaring the threshold to avoid the sqrt, with comment to explain. I use the same #if for qt 5 and 6 compatibility to use either pEvent->x() or pEvent->position()->x() as done at several other places in the code.

ronso0:
There's a conflict now after the hotkeys PR has been merged.

ronso0:
IIUC this now drags regardless which button is used, before it was only Qt::LeftButton.
Is this intended? If yes, why?

m0dB:
> IIUC this now drags regardless which button is used, before it was only Qt::LeftButton. Is this intended? If yes, why?

Not intended, I fixed it.

Also rebased on main and resolved the conflict. I think this is ready to be merged now.

ronso0:
Thank you, LGTM
Only WTrackWigetGroup needs to be adjusted, too.

m0dB:
Ah, thanks, not sure how I missed that one! Done. Someone please merge when the pipelines are green!

daschuer (approved):
LGTM, thank you

daschuer:
@ronso0 merge?

ronso0 (approved):

ronso0:
Yup, looks good, thank you! Nice move with the static x/y.

ronso0:
Hmm, was there any specific reason to base this on main? (if so I don't find it)

_Looks like_ this would also fix #13095

m0dB:
> Hmm, was there any specific reason to base this on main? (if so I don't find it)

The reason is that I first noticed the unintended drag&drop when working on the tracklist short-keys (as I was very actively selecting tracks to test the feature) and since that PR was based on main, this one as well.

But indeed, it's a good candidate for a backport to 2.4!

PR #12908, opened 2024-03-03
Computer feature: add sidebar action "Refresh directory tree"

ronso0:
This allows to show newly added directories when the parent had no directories when it was first added to the tree.

This works around the caching of the 'hasChildren' state of each tree item.

based on #12893

### DIFF
--- a/src/library/browse/browsefeature.cpp
+++ b/src/library/browse/browsefeature.cpp
@@ -327,29 +327,29 @@ void BrowseFeature::onRightClickChild(const QPoint& globalPos, const QModelIndex
 
     QMenu menu(m_pSidebarWidget);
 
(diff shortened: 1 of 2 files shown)


daschuer (changes requested):
I did the following: 
* Opened a folder with Mixxx
* Added a file using a file manager. 
* Select Refresh form the context menu. 
Expected: File is shown 
Actual: nothing happens.

daschuer:
I have also intuitive pressed F5, but that has a different meaning.

ronso0:
> The menu is missing for "Quick access" folders

I added that.

ronso0:
Yeah, F5 in the tracks view would also be cool.

daschuer (approved):
Thank you. Works good

PR #12909, opened 2024-03-03
Synchronize AutoDJ next deck with top track in queue

### LINKED ISSUE #8956: Auto DJ: Synchronizing queue's top item and next deck
Reported by: **[asfan.shirim](https://launchpad.net/~asfan.shirim)**
Date: 2017-10-26T14:40:47Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1727747](https://bugs.launchpad.net/bugs/1727747)
Tags: autodj

---

This is a feature request, although for me is fixing an unwanted AutoDJ behavior.
Continuing https://www.mixxx.org/forums/viewtopic.php?f=1&t=9409

I wish to have a feature that synchronizes the AutoDJ's top queue item with the next available deck, when AutoDJ is enabled.

What currently happens in Mixxx ?
1. AutoDJ is enabled and playing from the list.
2. One deck is playing while the second deck is queued with the top track in the AutoDJ queue.
3. If I move a different track up the queue to be top (or move the top track downward), the next deck still points to the track that was previously in queue.

I find this terribly confusing. It already happened several times that I changed the top track, but forgot to load it to the deck.

I think there should be an option to enable synchronizing the top queue slot with the next available deck, so when I change the one the other changes as well.
Additionally, if I double-click a track to load to the deck, it will ALSO be added to the top slot of the AutoDJ queue.
If I double-click on a track that is already in the AutoDJ list, it will moved to the top and loaded onto the deck.

I think my suggested feature is very intuitive and it is what I expected as the default behavior of AutoDJ.

Bacadam:
This fixes bug #8956, adding a signal in playlisttablemodel which is emitted when a change in track order involves the first track. The signal is connected to a slot in autodjprocessor which skips the track in next deck, replacing it with the new top track in queue.

### DIFF
--- a/src/library/autodj/autodjprocessor.cpp
+++ b/src/library/autodj/autodjprocessor.cpp
@@ -1671,7 +1671,22 @@ void AutoDJProcessor::playerRateChanged(DeckAttributes* pAttributes) {
 
 void AutoDJProcessor::playlistFirstTrackChanged() {
     qDebug() << this << "playlistFirstTrackChanged";
-    skipNext();
+    if (m_eState != ADJ_DISABLED) {
+        DeckAttributes* pLeftDeck = getLeftDeck();
+        DeckAttributes* pRightDeck = getRightDeck();
+        // if (!pLeftDeck || !pRightDeck) {
+        //     // User has changed the orientation, disable Auto DJ
+        //     toggleAutoDJ(false);
+        //     emit autoDJError(ADJ_NOT_TWO_DECKS);
+        //     return ADJ_NOT_TWO_DECKS;
+        // }
+
+        if (!pLeftDeck->isPlaying()) {
+            loadNextTrackFromQueue(*pLeftDeck);
+        } else if (!pRightDeck->isPlaying()) {
+            loadNextTrackFromQueue(*pRightDeck);
+        }
+    }
 }
 
 void AutoDJProcessor::setTransitionTime(int time) {
--- a/src/library/autodj/autodjprocessor.cpp
+++ b/src/library/autodj/autodjprocessor.cpp
@@ -1670,7 +1670,9 @@ void AutoDJProcessor::playerRateChanged(DeckAttributes* pAttributes) {
 }
 
 void AutoDJProcessor::playlistFirstTrackChanged() {
-    qDebug() << this << "playlistFirstTrackChanged";
+    if constexpr (sDebug) {
+        qDebug() << this << "playlistFirstTrackChanged";
+    }
     if (m_eState != ADJ_DISABLED) {
         DeckAttributes* pLeftDeck = getLeftDeck();
         DeckAttributes* pRightDeck = getRightDeck();
--- a/src/library/autodj/autodjprocessor.cpp
+++ b/src/library/autodj/autodjprocessor.cpp
@@ -534,6 +534,10 @@ AutoDJProcessor::AutoDJError AutoDJProcessor::toggleAutoDJ(bool enable) {
                 &DeckAttributes::rateChanged,
                 this,
                 &AutoDJProcessor::playerRateChanged);
+        connect(m_pAutoDJTableModel,
+                &PlaylistTableModel::firstTrackChanged,
+                this,
+                &AutoDJProcessor::playlistFirstTrackChanged);
 
         if (!leftDeckPlaying && !rightDeckPlaying) {
             // Both decks are stopped. Load a track into deck 1 and start it
@@ -1663,6 +1667,11 @@ void AutoDJProcessor::playerRateChanged(DeckAttributes* pAttributes) {
         return;
     }
     calculateTransition(fromDeck, getOtherDeck(fromDeck), false);
+}
+
+void AutoDJProcessor::playlistFirstTrackChanged() {
+    qDebug() << this << "playlistFirstTrackChanged";
+    skipNext();
 }
 
 void AutoDJProcessor::setTransitionTime(int time) {
--- a/src/library/autodj/autodjprocessor.cpp
+++ b/src/library/autodj/autodjprocessor.cpp
@@ -1674,12 +1674,6 @@ void AutoDJProcessor::playlistFirstTrackChanged() {
     if (m_eState != ADJ_DISABLED) {
         DeckAttributes* pLeftDeck = getLeftDeck();
         DeckAttributes* pRightDeck = getRightDeck();
-        // if (!pLeftDeck || !pRightDeck) {
-        //     // User has changed the orientation, disable Auto DJ
-        //     toggleAutoDJ(false);
-        //     emit autoDJError(ADJ_NOT_TWO_DECKS);
-        //     return ADJ_NOT_TWO_DECKS;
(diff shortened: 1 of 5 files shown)


JoergAtGithub:
Welcome at Mixxx!

ronso0:
I agree this is a bugfix, so this should go to the 2.4 branch.
@Bacadam please chnage the base branch here, and if there are unrelated commits shown also rebase your branch onto 2.4 and force-push.

Thank you!

src/library/playlisttablemodel.cpp
```diff
@@ -275,6 +279,11 @@ void PlaylistTableModel::moveTrack(const QModelIndex& sourceIndex,
     }
 
     m_pTrackCollectionManager->internalCollection()->getPlaylistDAO().moveTrack(m_iPlaylistId, oldPosition, newPosition);
+
+    if (oldPosition==1 || newPosition==1) {
+        qDebug() << "First track changed";
```
ronso0, line 284: IMO we don't need this debug message here, the one in AutoDJProcessor is sufficient.

ronso0 (changes requested):
Thank you for this fix, works as desired.

src/library/playlisttablemodel.cpp
daschuer: clang-format is complaining here.
You can avoid these complains by installing 
pre-commit 
https://github.com/mixxxdj/mixxx/wiki/Using-Git#set-up-automatic-code-checking

Bacadam:

Thanks! I've signed it under the name Bacadam.

daschuer (changes requested):
I have an issue: 
* Load the second track from the Auto-DJ playlist to the paused deck. 
* Now the we have an "inconsistent state"
* Fix it by moving the track up. 
* The track disappears from the playlist and the original first track is loaded
This should not happen.

Bacadam:
> Thank you for this fix, works as desired.

Thanks, I'll do the requested changes and the rebase tomorrow.

Bacadam:
Sorry, I force pushed before having change the base branch here and so Github bot added some irrelevant tags.

Bacadam:
I can't build the 2.4 branch anymore since I've installed Qt6 on my machine. Do you know how I can build using Qt5, when both versions are installed?

Bacadam:
Okay I've installed pre-commit and I think I solved all clang-format issues with last commit

ronso0:
I think you can squash all three commits. (and force-push)

Did you manage to fix the issue @daschuer discovered?

Bacadam:
The behavior @daschuer discovered is due to the subsequent calls  to `removeLoadedTrackFromTopOfQueue` and `loadNextTrackFromQueue` in `skipNext`.
I tried to copy all the `skipNext` code in the `playlistFirstTrackChanged()` slot, without the `removeLoadedTrackFromTopOfQueue` lines and it worked.

The question now is what are the checks in `skipNext` that are useful in our case and that must be done in the slot as well?

ronso0:
I suggest to first squash the current state, push, then append an experimental commit so we can discuss the code in place.

Bacadam:
I copied the skipNext code and kept what I think is useful. I tested it and it works.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -1665,6 +1669,26 @@ void AutoDJProcessor::playerRateChanged(DeckAttributes* pAttributes) {
...
+    qDebug() << this << "playlistFirstTrackChanged";
+    if (m_eState != ADJ_DISABLED) {
+        DeckAttributes* pLeftDeck = getLeftDeck();
+        DeckAttributes* pRightDeck = getRightDeck();
+        // if (!pLeftDeck || !pRightDeck) {
+        //     // User has changed the orientation, disable Auto DJ
+        //     toggleAutoDJ(false);
+        //     emit autoDJError(ADJ_NOT_TWO_DECKS);
+        //     return ADJ_NOT_TWO_DECKS;
+        // }
```
Bacadam, line 1682: I just don't know what to do with these commented lines as I don't know what deck orientation is, and how user can change it.

Bacadam:
I found another thing that we can consider as a bug now that this synchronization is implemented.
When you right-click on tracks or playlists and select `Add to AutoDJ (top)` or `Add to AutoDJ (replace)`, the track are added in second position, and so the next deck is not modified.

Bacadam:
> I found another thing that we can consider as a bug now that this synchronization is implemented. When you right-click on tracks or playlists and select `Add to AutoDJ (top)` or `Add to AutoDJ (replace)`, the track are added in second position, and so the next deck is not modified.

The code responsible for that is in `playlistdao.cpp`, and I proposed a working solution.
The problem is I need to call my `playlistFirstTrackChanged()`slot from `playlistdao.cpp`and I don't know what's the right way.
For my precedent commit to work, I changed `playlistFirstTrackChanged()` to public slot but that may not be a really good practice?

daschuer:
> When you right-click on tracks or playlists and select Add to AutoDJ (top) or Add to AutoDJ (replace), the track are added in second position, and so the next deck is not modified.

I think that is the desired behavior. The use case is that the users may have already cued the track in the second deck, and just wants to collect some other tracks while browsing the library.

If hey want to play the track after the next they can load the track directly into the deck. So we have both cases covered.

Bacadam:
Are we sure about that? If the users have selected the next track and browse more tracks, they probably want to add them to the end of the queue? Why would they especially want to add a track in third position?

If they use AutoDJ, they probably don't load directly in decks?

I might have a "music player" approach on that, this is how most music players work. But I can't tell for DJ softwares.

Anyway, what you prefer, I can revert my last commit if you want.

Bacadam:
Okay I reverted my work on Add to AutoDJ buttons, so this PR is in a clean state.

daschuer (approved):
LGTM, thank you

daschuer:
@ronso0: Ready for merge?

src/library/autodj/autodjprocessor.cpp
```diff
@@ -1665,6 +1669,20 @@ void AutoDJProcessor::playerRateChanged(DeckAttributes* pAttributes) {
     calculateTransition(fromDeck, getOtherDeck(fromDeck), false);
 }
 
+void AutoDJProcessor::playlistFirstTrackChanged() {
+    qDebug() << this << "playlistFirstTrackChanged";
```
ronso0, line 1673: please wrap this in `if constexpr (sDebug) {` like the other debug logs.

ronso0 (changes requested):

src/library/autodj/autodjprocessor.cpp
Bacadam: Done, thanks for your feedback.

daschuer:
@ronso0 Is this now merge-able?

ronso0:
Yes, sorry for the delay, this looks good.
Thank you!

ronso0:
@daschuer do yo already have a changelog update ready, or shall I do one for this PR?

PR #12910, opened 2024-03-03
CMakeLists: Emit better errors for exotic target platforms

fwcd:
Currently, users will get a slightly misleading error message e.g. when referring to the macOS environment when building for "exotic" platforms like iOS or Debian when targeting Emscripten (WebAssembly).

This fixes the issue by only suggesting the prebuilt environments when targeting the corresponding platforms and referring the user to vcpkg otherwise.


daschuer (approved):
LGTM, thank you.

PR #12911, opened 2024-03-03
Labeler: Add `qml` to labeler config

fwcd:
This adds the `qml` to the PR labeler, which might be useful to filter PRs working on QML.


daschuer (approved):
LGTM, thank you.

PR #12913, opened 2024-03-04
Build: Add `PORTMIDI` flag for compiling with(out) PortMidi

fwcd:
This is consistent with the other controller backends and will be useful both for targeting Emscripten/WebAssembly (since PortMidi doesn't support Web MIDI yet) and for a potential `libremidi` migration in the future.


daschuer (approved):
LGTM, thank you.

PR #12914, opened 2024-03-04
Build: Add `MUSICBRAINZ` flag for compiling with(out) MusicBrainz

fwcd:
This adds the flag `MUSICBRAINZ` which controls whether the tag fetcher dialog and the rest of the MusicBrainz integration, including acoustic fingerprinting, is built.

This is useful, because it makes it possible to build Mixxx without Chromaprint and thus ffmpeg, which is pretty tricky to get to build for WebAssembly (the port doesn't build as-is and integrating `ffmpeg.wasm` seems to be nontrivial).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -929,8 +929,6 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/library/dao/trackschema.cpp
   src/library/dlgcoverartfullsize.cpp
   src/library/dlgcoverartfullsize.ui
-  src/library/dlgtagfetcher.cpp
-  src/library/dlgtagfetcher.ui
   src/library/dlgtrackinfo.cpp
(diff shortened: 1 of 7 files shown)


daschuer:
Uh, this clutters the code a lot. Is this only a temporary workaround? If yes, we may keep the Musikbrainz stubban only fail the chroma ID generation.

daschuer:
Alternative would be to extract the Musikbrainz classes a bit more using a proxy class or such that would also allow to use alternative implementations. (Just a rough idea) What do you think?

fwcd:
Great idea, I admittedly didn't really anticipate that the MusicBrainz backend was mostly coupled to the tag fetcher, so this ended up being noisier than I thought.

Would be nice to cleanly separate out the MusicBrainz part from general tag fetching (then MusicBrainz could then be moved behind a flag without cluttering the code with lots of #ifdefs in other places).

PR #12915, opened 2024-03-04
MixxxApplication: Use `QWasmIntegrationPlugin` when targeting WebAssembly

fwcd:
This imports the WASM integration plugin when `Q_OS_WASM` is defined and compiles out the `QOffscreenIntegrationPlugin` and `QMinimalIntegrationPlugin`, which are unavailable when targeting WASM.


daschuer (approved):
LGTM, thank you.

PR #12916, opened 2024-03-04
DesktopHelper: Compile out process-spawning on WASM too

fwcd:
Like on iOS, `QProcess` is unavailable when targeting WebAssembly.


daschuer (approved):
LGTM, thank you.

PR #12920, opened 2024-03-04
[PreviewDeckN],LoadSelectedTrackAndPlay toggles play/pause if the track is already loaded

### LINKED ISSUE #9819: Pressing [Enter] does not pause but stops the preview
Reported by: **[mnikitin](https://launchpad.net/~mnikitin)**
Date: 2019-12-05T07:11:35Z
Status: Confirmed
Importance: Medium
Launchpad Issue: [lp1855239](https://bugs.launchpad.net/bugs/1855239)
Tags: keyboard, usability, windows

---

Pressing [Enter] key while a song is been played in the Preview deck does not pause but stops the playback. New press does not resume nor restart the playback, one needs to re-load the song by [Alt+Enter].

ronso0:
Fixes #9819 

Due to limited unused hotkeys (in the default mapping) and to allow controlling the preview deck with a few hotkeys only I propose to extend/override the behaviour of `[PreviewDeckN], LoadSelectedTrackAndPlay`:
if the track is already loaded, toggle play/pause.

This is mapped to `P`
`Shift`+`P` does `start_stop` (stop and seek to start) (could as well be `cue_gotoandstop`?)

**TODO**
- [x] update [manual](https://manual.mixxx.org/2.5/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-LoadSelectedTrackAndPlay)  https://github.com/mixxxdj/manual/pull/625
- [x] update the mapping sheet
image


daschuer:
Good idea.

src/mixer/playermanager.cpp
```diff
@@ -696,6 +696,20 @@ void PlayerManager::slotLoadTrackToPlayer(TrackPointer pTrack, const QString& gr
             // so clone another playing deck instead of loading the selected track
             clone = true;
         }
+    } else if (isPreviewDeckGroup(group) && play) {
+        // This extends/overrides the behaviour of [PreviewDeckN],LoadSelectedTrackAndPlay:
+        // if the track is already loaded, toggle play/pause.
+        auto pPlayer = getPlayer(group);
```
daschuer, line 702: We have pPlayer and the null check already above.

src/mixer/playermanager.cpp
```diff
@@ -696,6 +696,20 @@ void PlayerManager::slotLoadTrackToPlayer(TrackPointer pTrack, const QString& gr
...
+        VERIFY_OR_DEBUG_ASSERT(pPlayer) {
+            return;
+        }
+        if (pTrack == pPlayer->getLoadedTrack()) {
+            bool isPlaying =
+                    ControlObject::get(ConfigKey(group, QStringLiteral("play"))) > 0.0;
+            ControlObject::set(ConfigKey(group, QStringLiteral("play")),
+                    isPlaying ? 0.0 : 1.0);
```
daschuer, line 710: Please avoid to look up the control object twice.

daschuer (changes requested):
Thank you

ronso0:
The corresponding manual PR is https://github.com/mixxxdj/manual/pull/625

src/mixer/playermanager.cpp
```diff
@@ -696,6 +696,16 @@ void PlayerManager::slotLoadTrackToPlayer(TrackPointer pTrack, const QString& gr
             // so clone another playing deck instead of loading the selected track
             clone = true;
         }
+    } else if (isPreviewDeckGroup(group) && play) {
+        // This extends/overrides the behaviour of [PreviewDeckN],LoadSelectedTrackAndPlay:
+        // if the track is already loaded, toggle play/pause.
+        if (pTrack == pPlayer->getLoadedTrack()) {
+            auto* pPlay =
+                    ControlObject::getControl(ConfigKey(group, QStringLiteral("play")));
+            double newPlay = pPlay->get() > 0.0 ? 0.0 : 1.0;
```
daschuer, line 705:
```suggestion
            double newPlay = pPlay->toBool() ? 0.0 : 1.0;
```

daschuer (changes requested):
Just one minor.
Why is this still in n draft?

ronso0:
> Why is this still in n draft?

I noticed a glitch in the keyboard sheet preview (:heavy_check_mark:  seemst to be an issue with the GH preview) and I didn't check if this affects the Preview delegate (:heavy_check_mark: ).

So this is ready to roll.
Thanks for your prompt review!

ronso0:
I noticed an incosistency:
* the preview Play button right-click is mapped to `start` (and stops, so this is the same as `start_stop`?)
* the tooltip says `cue_gotoandplay` (LateNight only)
* `Shift`+`P` is mapped to `start_stop`
* the Preview delegate always does `cue_gotoandplay` and `play` (pause)

I think we should map the Play button right-click and `Shift`+`P` to `cue_gotoandplay`, too.
If the user wants to go back to the the start this can be done by dragging the playhead on the overview, while otoh jumping to Cue requires to click the C icon (which is rather small in the preview deck).
For unprepared tracks Cue = start, and for prepared tracks we may assume Cue has been set to a different position on purpose.

Not going to fix this here, let's discuss it in #12925

src/mixer/playermanager.cpp
ronso0: fixed

ronso0:
FWIW, with 2.4 I managed to get a mismatch of the preview deck's play button (not lit) and Preview delegate (lit). I continued clicking the delegate so I can't tell whether the deck was actually playing (shouldn't since the Play button CO is reliable IMO).

ronso0:
The manual PR has been merged already, let's merge this, too.
Any objections?

daschuer (approved):
LGTM, Thank you.

PR #12921, opened 2024-03-05
CMakeLists: Enable asyncify when targeting WASM

fwcd:
This fixes calls that yield to the event loop when targeting WebAssembly (e.g. `QDialog::exec` would previously crash), as per https://doc.qt.io/qt-6/wasm.html#asyncify.

It does add some overhead unfortunately (the WASM binary went from 256M -> 304M and linking became considerably slower), even with the recommended optimizations (`-Os`), but better to have something that runs than crashes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1585,6 +1585,13 @@ if(APPLE)
   endif()
 endif()
 
+if(EMSCRIPTEN)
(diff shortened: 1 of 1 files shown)


daschuer (approved):
LGTM, thank you.

PR #12922, opened 2024-03-05
Resources: Bundle resources for preloading when targeting Emscripten/WASM

fwcd:
This is the final piece in the puzzle to getting Mixxx to successfully launch in the browser[^1].

Emscripten provides a virtual file system as described in https://emscripten.org/docs/porting/files/packaging_files.html. Since our WASM binary is already pretty heavy (304M) I went with the preloading option for now, which bundles our `res` directory in a single `mixxx.data` (96M), located at `/res` within the virtual file system. `qResourcePath` is also updated to use this path when targeting Emscripten.

[^1]: Of course there is still a lot of work ahead to make this run well

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1811,6 +1811,14 @@ if (APPLE)
   endif()
 endif()
 
+if(EMSCRIPTEN)
+  # Package resources for the web using preloading.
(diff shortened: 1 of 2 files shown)


daschuer (approved):
LGTM , thank you.

PR #12928, opened 2024-03-06
(fix) Deere: make sampler rows persist

ronso0:
Currently the selection is not stored, not sure when that broke.


daschuer (approved):
LGTM, Thank you.

PR #12929, opened 2024-03-06
Show hint if resource path in CMakeCache.txt does not exist

cyberchrime:
If a `CMakeCache.txt` was located in the same directory as the `mixxx` binary, `mixxx` tried to read the resource path from the `CMakeCache.txt` file. However, if the resource path is not available, is simply prints an error and exits the application. Now a hint is displayed stating to use the command line option `--resource-path <path>`. 

Additionally, this error is suppressed when a user requests help on command line options with `--help`.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -53,9 +53,9 @@ const QString kScaleFactorKey = QStringLiteral("ScaleFactor");
 constexpr int kPixmapCacheLimitAt100PercentZoom = 32 * 1024; // 32 MByte
 
 int runMixxx(MixxxApplication* pApp, const CmdlineArgs& args) {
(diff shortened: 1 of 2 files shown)


acolombier:
While this change makes sense to improve developer experience, I'm afraid it could be be misleading if a final user is seeing this (e.g after accidentally moving/removing is `res` folder.

Perhaps a solution would be to add a CMake definition (ON by default OFF when building production artefacts) and give different message depending that definition. @mixxxdj/developer any objection?

daschuer:
Another solution would be to make the message conditional for the different sources of the res path.

daschuer (approved):
Looks and works good. Thank you.

PR #12930, opened 2024-03-07
ScreenSaverHelper: Add no-op implementation for WASM

fwcd:
Previously Mixxx would crash when debug asserts were enabled. There may be some workarounds to prevent sleep in a browser (I've seen people use `<video>` tags for that), but for now having an empty implementation seems like the best option.


JoergAtGithub:
LGTM! Thank you!

PR #12931, opened 2024-03-07
CMakeLists: Add `WASM_ASSERTIONS` option

fwcd:
This option will add additional checks, including stack overflow protection, which are disabled by default when building with optimizations (which we almost always do).

For more details, see https://emscripten.org/docs/porting/Debugging.html#compiler-settings

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2802,6 +2802,13 @@ if(DEBUG_ASSERTIONS_FATAL)
   endif()
 endif()
 
+if(EMSCRIPTEN)
+  cmake_dependent_option(WASM_ASSERTIONS "Enable additional checks when targeting Emscripten/WebAssembly" ON "DEBUG_ASSERTIONS_FATAL OR CMAKE_BUILD_TYPE STREQUAL Debug" OFF)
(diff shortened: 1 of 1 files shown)


fwcd:
Hm, with assertions enabled we unfortunately get a lot of errors of the form

```
Cannot have multiple async operations in flight at once
```

Sadly, it is pretty hard to debug where these calls are coming from, i.e. whether it's something in Qt or in Mixxx, since we do not get a stack trace... This results in Mixxx not being able to launch at all.

Marking this as a draft for now.

fwcd:
We'll disable this option for now since it prevents Mixxx from launching currently (see [this post for details](https://forum.qt.io/topic/155159/cannot-have-multiple-async-operations-in-flight-at-once-during-qapplication-initialization), which is likely related to https://github.com/emscripten-core/emscripten/issues/18412 and [QTBUG-102827](https://bugreports.qt.io/browse/QTBUG-102827)), but may be valuable for debugging in the future.

Marking as ready-for-review since it should be safe to merge (it doesn't change anything about the default config).

daschuer (approved):
LGTM, thank you.

PR #12937, opened 2024-03-08
(fix) Library: allow adding new dirs while we have missing watched dir(s)

### LINKED ISSUE #10481: improve UX when encountering missing/invalid library directory
Date: 2021-07-23T16:45:10Z
Status: New
Importance: Undecided
Launchpad Issue: [lp1937851](https://bugs.launchpad.net/bugs/1937851)
Tags: library, usability

---

Amongst others, my library contains a folder on an external drive.
That drive was not mounted when I wanted to add a subfolder of my home directory. Adding was refused, message says it's already in the library or the library is being scanned. Neither was true.
That was driving me crazy so I removed all watched folders to finally  figur it's failing due to the unmounted drive :\

the message
https://github.com/mixxxdj/mixxx/blob/3546054e299956a952c1fb1b7855e95e5f2be200/src/library/library.cpp#L504-L509

caused by
https://github.com/mixxxdj/mixxx/blob/3546054e299956a952c1fb1b7855e95e5f2be200/src/library/dao/directorydao.cpp#L68

Though the logged line is clear about the cause:
> Aborting to add QFileInfo(/home/abc/Music/mmumu) : Loaded directory QFileInfo(/mnt/yxo/audio) does not exist or is inaccessible

After closing the dialog and hitting Apply in the preferences Mixxx says "one or more directories were added... Rescan / Cancel"
Also wrong.

1) can `DirectoryDAO::AddResult` be passed through to the requester so that we can have explicit error messages?

2) Why is a missing directory hindering the addition of others in the first place? I'm aware the dir is missing, that was the reason to add another track source ;)

ronso0:
Fixes the original issue of  #10481

I don't understand how the "inconsistencies" mentioned in the comment can be so serious that we have to prevent adding a new directory.
After a rescan, everything should be fine.

### DIFF
--- a/src/library/dao/directorydao.cpp
+++ b/src/library/dao/directorydao.cpp
@@ -62,27 +62,25 @@ DirectoryDAO::AddResult DirectoryDAO::addDirectory(
     const auto newCanonicalLocation = newDir.canonicalLocation();
     DEBUG_ASSERT(!newCanonicalLocation.isEmpty());
     QList<mixxx::FileInfo> obsoleteChildDirs;
-    for (auto&& oldDir : loadAllDirectories()) {
-        if (!oldDir.exists() || !oldDir.isDir()) {
-            // Abort to prevent inconsistencies in the database
-            kLogger.warning()
-                    << "Aborting to add"
-                    << newDir
-                    << ": Loaded directory"
-                    << oldDir
-                    << "does not exist or is inaccessible";
-            return AddResult::InvalidOrMissingDirectory;
-        }
+    // Ignore invalid or missing directories in order to allow adding new dirs
+    // while the list contains e.g. currently unmounted removable drives.
+    // Worst that can happen is that we have orphan tracks in the database that
+    // would be moved to missing after the rescan (which is required anyway after
+    // having added a new dir).
+    for (auto&& oldDir : loadAllDirectories(true)) {
         const auto oldCanonicalLocation = oldDir.canonicalLocation();
         DEBUG_ASSERT(!oldCanonicalLocation.isEmpty());
         if (mixxx::FileInfo::isRootSubCanonicalLocation(
                     oldCanonicalLocation,
                     newCanonicalLocation)) {
+            // New dir is a child of an existing dir, return
             return AddResult::AlreadyWatching;
         }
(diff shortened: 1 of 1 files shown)


src/library/dao/directorydao.cpp
```diff
@@ -62,27 +62,25 @@ DirectoryDAO::AddResult DirectoryDAO::addDirectory(
...
-                    << oldDir
-                    << "does not exist or is inaccessible";
-            return AddResult::InvalidOrMissingDirectory;
-        }
+    // Ignore invalid or missing directories in order to aloow adding new dirs
+    // while the list contains e.g. currently unmounted removable drives.
+    // Worst that can happen is that we have orphan tracks in the database that
+    // would be moved to missing after the rescan (which is required anyway after
+    // having added a new dir).
+    for (auto&& oldDir : loadAllDirectories(true)) {
```
ronso0, line 70: Alternatively, use the current implemantation but don't return if the old dir is missing. Instead, do the QString equivalent of `mixxx::FileInfo::isRootSubCanonicalLocation(new, old)` in that case.
This requires refactirng here, i.e. use `DirectoryDAO::getRootDirStrings()` added in #12878 and construct the mixxx::FileInfo here.

src/library/dao/directorydao.cpp
```diff
@@ -62,27 +62,25 @@ DirectoryDAO::AddResult DirectoryDAO::addDirectory(
...
-            // Abort to prevent inconsistencies in the database
-            kLogger.warning()
-                    << "Aborting to add"
-                    << newDir
-                    << ": Loaded directory"
-                    << oldDir
-                    << "does not exist or is inaccessible";
-            return AddResult::InvalidOrMissingDirectory;
-        }
+    // Ignore invalid or missing directories in order to aloow adding new dirs
```
daschuer, line 65:
```suggestion
    // Ignore invalid or missing directories in order to allow adding new dirs
```

daschuer (changes requested):

src/library/dao/directorydao.cpp
ronso0: yup, fixed.

daschuer (approved):
LGTM, thank you

PR #12938, opened 2024-03-08
(fix) Tango waveform toggle, remove unneeded waveform Singleton

ronso0:
currently waveforms can't be toggled (state depends on config value 0/1)


daschuer:
How can I reproduce the issue?

ronso0:
Hmm, I'm not sure what I expereienced. Indeed this is a no-op, it just uses the official controls and removes the unneeded singleton.

daschuer (approved):
Ok, than let's merge. Thanks.

PR #12934, opened 2024-03-07
feat: add support for scaling BPM by different ratios

mxmilkiib:
This change introduces the ability to scale the BPM of a track by different ratios using ControlPushButton objects. Each button is connected to a lambda function that calls the slotScaleBpm method with the corresponding BPM scale factor.

PR #12941, opened 2024-03-08
feat: add additional BPM control options for more precise adjustments

mxmilkiib:
Depends on #12934

### DIFF
--- a/src/controllers/controlpickermenu.cpp
+++ b/src/controllers/controlpickermenu.cpp
@@ -243,6 +243,12 @@ ControlPickerMenu::ControlPickerMenu(QWidget* pParent)
     addDeckAndSamplerControl("bpm_up_small", tr("BPM +0.1"), tr("Increase BPM by 0.1"), pBpmMenu);
     addDeckAndSamplerControl("bpm_down_small", tr("BPM -0.1"), tr("Decrease BPM by 0.1"), pBpmMenu);
     addDeckAndSamplerControl("bpm_tap", tr("BPM Tap"), tr("BPM tap button"), pBpmMenu);
+    addDeckAndSamplerControl("beats_set_halve", tr("BPM Halve"), tr("BPM halve"), pBpmMenu);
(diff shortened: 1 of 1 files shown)


src/controllers/controlpickermenu.cpp
```diff
@@ -243,6 +243,12 @@ ControlPickerMenu::ControlPickerMenu(QWidget* pParent)
     addDeckAndSamplerControl("bpm_up_small", tr("BPM +0.1"), tr("Increase BPM by 0.1"), pBpmMenu);
     addDeckAndSamplerControl("bpm_down_small", tr("BPM -0.1"), tr("Decrease BPM by 0.1"), pBpmMenu);
     addDeckAndSamplerControl("bpm_tap", tr("BPM Tap"), tr("BPM tap button"), pBpmMenu);
+    addDeckAndSamplerControl("beats_set_halve", tr("BPM Halve"), tr("BPM halve"), pBpmMenu);
```
ronso0, line 246:
```suggestion
    addDeckAndSamplerControl("beats_set_halve", tr("BPM Halve"), tr("Multiply current BPM with 0.5"), pBpmMenu);
```
etc.

ronso0 (changes requested):
Please change the description as suggested, then amend the commit to #12934 in order to have all related changes in one place

src/controllers/controlpickermenu.cpp
ronso0: Taking a second look, I think it would be better to have the actions descried like the occur in the track menu
halve - 2/3 - 3/4 - 4/3 - 3/2 - double
https://github.com/mixxxdj/mixxx/blob/3aa73fd1cb0f39c3993bff2913f72c7d03c58c13/src/widget/wtrackmenu.cpp#L454

ronso0 (changes requested):

mxmilkiib:
Integrated into #12934

PR #12945, opened 2024-03-10
OpenGLWindow: Fix sizing on Wasm by setting `Qt::FramelessWindowHint`

fwcd:
The Wasm implementation of windows [imposes a minimum size of 100x100](https://github.com/qt/qtbase/blob/0804109d686e0a99ab0de0f1c70e3422183c6e98/src/plugins/platforms/wasm/qwasmwindow.cpp#L214-L215) when this flag is not set. This patch fixes it.

| Before | After |
| - | - |
|  |  |


JoergAtGithub:
Works for me! Thank you!

PR #12946, opened 2024-03-10
increase keylock disable limit to 250%, apply only for SoundTouch

ronso0:
see https://mixxx.zulipchat.com/#narrow/stream/109122-general/topic/keylock.20for.20more.20than.20.2B100.25.20speed.20change.20possible.3F

### We need to test for regressions with
* speeds up to 250%
* all keylock engines.
* scratching with Rubberband v2/v3
* seeking

Draft until we have confirmed no / minimal regressions.

### DIFF
--- a/src/engine/enginebuffer.cpp
+++ b/src/engine/enginebuffer.cpp
@@ -884,16 +884,14 @@ void EngineBuffer::processTrackLocked(
 
     bool useIndependentPitchAndTempoScaling = false;
 
-    // TODO(owen): Maybe change this so that rubberband doesn't disable
-    // keylock on scratch. (just check m_pScaleKeylock == m_pScaleST)
-    if (is_scratching || fabs(speed) > 1.9) {
-        // Scratching and high speeds with always disables keylock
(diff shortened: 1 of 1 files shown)


ronso0:
In case of doubts or issues, we may create a minimal fix (SoundTouch only, 100%) for 2.4

daschuer:
How is the state here? Do we have any test results? We may just merge this to 2.5 and collect test data during 2.5-beta.

ronso0:
Nope, I did only a brief test with rate +100%.

Targetting main now.

daschuer (changes requested):
Just did a test, Rubberband and Soundtouch are sounding terrible at 250. But it works. It also sounds terrible when you pass the threshold and it goes to chipmunk without a warning. 

So it is OK to raise the threshold. But how about removing the speed threshold entirely?

The other issue it that cueing by scratching for a first beat, sound now totally different with key-lock enabled. My first impulse was, that it is broken. After a bit testing I think one can get used to, but auto-disabling key lock during scratching is IMHO a good feature for that use case. 

Is there a real use case for scratching with key lock?

PR #12281, opened 2023-11-09
Fix for visual waveform position while scratching outside of an activated loop

JoergAtGithub:
Closes: #12274 
Add additional conditions to ensure, that loop boundary adjustment of the visual playposition is only done, when the engine playposition is inside the loop.

PR #12951, opened 2024-03-11
add `hotcue_X_indicator` control, is 1.0 if playposition is at hotcue

ronso0:
Just a little helper that allows having callbacks for when the playposition crosses a hotcue / loopcue.
For example:
* always trigger the fog machine on hotcue5 via MIDI_4_Lights
* auto-censor certain spots in _all_ tracks by making hotcue17 trigger a certain sample / engage revere-roll for 1 sec
* flash controller hotcue buttons if deck is stopped at a hotcue
* ...

### Description
Basically, it's something like `beat_active` and `cue_indicator`: it is activated (1.0) if
* the play position is on the hotcue
(when stopped or when the buffer's new position is coincidentally at the hotcue position)
* the play position crossed the hotcue since the last update
(while playing forward or backward and while scratching, **not** when seeking over a hotcue)

_`beat_active` is 1.0 if the playposition is in a certain **range** around a beat marker_
_`cue_indicator`_ is 1.0 (maybe flashing) if the play position is **exactly** at the Cue (unlikely when playing)_

### Improvement?
Integrating it into `hotcue_X_status` would allow GUI feedback via WHotcueButton, it'd be just two more states that could be styled rather easily. #9760 
For now this is separate because integration/extension seems a bit cumbersome.

### Nice to have
- [ ] tests that covers looping (hotcue slight before and at loop_out, slightly after loop_in)

### Proposal for testing
* load this mapping to the MIDI Through controller https://gist.github.com/ronso0/2dd322622ddac2590020ce9b2af5d695
* load a track to deck1, set hotcues 1-2-3 with a distance of ~10s
* load a track to deck2
* both decks stoped
* Test1:
   * play from start
   * jump to hotcue 1
   * jump to hotcue 3
   * == deck2 should **not** start
* Test2:
   * play from start
   * jump to hotcue 1
   * jump to hotcue 2
   * == deck2 should start
* Test1:
   * stopped
   * jump to hotcue 3
   * scratch backwards over hotcue 2
   * == deck2 should start

### DIFF
--- a/src/engine/controls/cuecontrol.cpp
+++ b/src/engine/controls/cuecontrol.cpp
@@ -500,6 +500,7 @@ void CueControl::trackLoaded(TrackPointer pNewTrack) {
         setHotcueFocusIndex(Cue::kNoHotCue);
         m_pLoadedTrack.reset();
         m_usedSeekOnLoadPosition.setValue(mixxx::audio::kStartFramePos);
+        m_prevPosition.setValue(mixxx::audio::kStartFramePos);
     }
 
     if (!pNewTrack) {
@@ -2168,11 +2169,44 @@ void CueControl::updateIndicators() {
             m_pCueIndicator->setBlinkValue(ControlIndicator::ON);
         }
     }
+
+    // Update hotcue indicators: activate the indicator if we are at the cue pos
+    // or if we passed it since the last update, else deactivate.
+    const auto prevPos = m_prevPosition.getValue();
+    const auto currPos = frameInfo().currentPosition;
+    VERIFY_OR_DEBUG_ASSERT(prevPos.isValid()) {
+        m_prevPosition.setValue(currPos);
+        return;
+    }
+
+    for (const auto& pControl : std::as_const(m_hotcueControls)) {
+        const auto cuePos = pControl->getPosition();
+        if (!cuePos.isValid()) {
+            continue;
+        }
+        if ((cuePos >= prevPos && cuePos <= currPos) ||     // forward
+                (cuePos <= prevPos && cuePos >= currPos)) { // reverse
+            pControl->setIndicator(1.0);
+        } else {
+            pControl->setIndicator(0.0);
+        }
+    }
+    m_prevPosition.setValue(currPos);
 }
 
 void CueControl::resetIndicators() {
     m_pCueIndicator->setBlinkValue(ControlIndicator::OFF);
     m_pPlayIndicator->setBlinkValue(ControlIndicator::OFF);
+    for (const auto& pControl : std::as_const(m_hotcueControls)) {
+        pControl->setIndicator(0.0);
+    }
+}
+
+/// Store the position so we can do a correct range check when updating the
+/// hotcue indicators. Otherwise m_prevPosition would be the position from before
+/// the seek and we'd falsely activate indicators.
+void CueControl::notifySeek(mixxx::audio::FramePos position) {
+    m_prevPosition.setValue(position);
 }
 
 CueControl::TrackAt CueControl::getTrackAt() const {
@@ -2461,9 +2495,12 @@ HotcueControl::HotcueControl(const QString& group, int hotcueIndex)
 
     m_pHotcueStatus = std::make_unique<ControlObject>(keyForControl(QStringLiteral("status")));
     m_pHotcueStatus->setReadOnly();
-
     // Add an alias for the legacy hotcue_X_enabled CO
     m_pHotcueStatus->addAlias(keyForControl(QStringLiteral("enabled")));
+
+    m_pHotcueIndicator = std::make_unique<ControlObject>(
+            keyForControl(QStringLiteral("indicator")));
+    m_pHotcueIndicator->setReadOnly();
 
     m_hotcueType = std::make_unique<ControlObject>(keyForControl(QStringLiteral("type")));
     m_hotcueType->setReadOnly();
@@ -2702,6 +2739,7 @@ void HotcueControl::resetCue() {
     setEndPosition(mixxx::audio::kInvalidFramePos);
     setType(mixxx::CueType::Invalid);
     setStatus(Status::Empty);
+    setIndicator(0.0);
 }
 
 void HotcueControl::setPosition(mixxx::audio::FramePos position) {
@@ -2718,6 +2756,12 @@ void HotcueControl::setType(mixxx::CueType type) {
 
 void HotcueControl::setStatus(HotcueControl::Status status) {
     m_pHotcueStatus->forceSet(static_cast<double>(status));
+}
+
+void HotcueControl::setIndicator(double on) {
+    m_pHotcueIndicator->forceSet(on);
+    // TODO: check if we can squeez this into hotcue_N_status without too much effort.
+    // IINM that would require reading e.g. m_pCurrentSavedLoopControl.
 }
 
 HotcueControl::Status HotcueControl::getStatus() const {
--- a/src/engine/controls/cuecontrol.cpp
+++ b/src/engine/controls/cuecontrol.cpp
@@ -202,6 +202,8 @@ void CueControl::createControls() {
     m_pVinylControlEnabled = std::make_unique<ControlProxy>(m_group, "vinylcontrol_enabled");
     m_pVinylControlMode = std::make_unique<ControlProxy>(m_group, "vinylcontrol_mode");
 
+    m_pReverse = std::make_unique<ControlProxy>(m_group, "reverse");
+
     m_pHotcueFocus = std::make_unique<ControlObject>(ConfigKey(m_group, "hotcue_focus"));
     setHotcueFocusIndex(Cue::kNoHotCue);
     m_pHotcueFocusColorPrev = std::make_unique<ControlObject>(
@@ -2175,11 +2177,33 @@ void CueControl::updateIndicators() {
 
     // Update hotcue indicators: activate the indicator if we are at the cue pos
     // or if we passed it since the last update, else deactivate.
-    const auto prevPos = m_prevPosition.getValue();
+    auto prevPos = m_prevPosition.getValue();
     const auto currPos = frameInfo().currentPosition;
     VERIFY_OR_DEBUG_ASSERT(prevPos.isValid()) {
         m_prevPosition.setValue(currPos);
         return;
+    }
+
+    // If we're looping check if we just wrapped around.
+    // Since explicit seeks are reported in notifiySeek(), this is safe
+    // to detect the true loop wrap-around.
+    LoopInfo loopInfo = m_loopInfo;
+    bool loopingWrapAround = false;
+    bool loopingWrapAroundReverse = false;
+    // This covers both `reverse` and scratching backwards
+    bool reverse = getEngineBuffer()->getSpeed() < 0.0;
+    if (m_pLoopEnabled->toBool() &&
+            loopInfo.startPosition.isValid() &&
+            loopInfo.endPosition.isValid() &&
+            // Are we inside the current loop?
+            posInsideLoop(prevPos, loopInfo) &&
+            posInsideLoop(currPos, loopInfo)) {
+        // Check if we wrapped around
+        if (!reverse && prevPos > currPos) {
+            loopingWrapAround = true;
+        } else if (reverse && prevPos < currPos) {
+            loopingWrapAroundReverse = true;
+        }
     }
 
     for (const auto& pControl : std::as_const(m_hotcueControls)) {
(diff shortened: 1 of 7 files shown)


daschuer (commented):
Good idea, the only issue we have that the code does not work in reverse. I think we have similar code for the beat indicator, you may re-use

ronso0:
Yes, for reverse I'll add a else branch.

beat_active is different though and the code is more complex because it is supposed to light up LEDs or widgets for X milliseconds while the track is playing, i.e. be ON when the playhead is in a certain +- range at/around the actual beat.

This is not required for hotcue_X_indicator which is only supposed to trigger while playing (don't miss it, duration doesn't matter) and when being stopped.

ronso0:
I adjusted it to
* consider both forward and reverse play
* ignore seeks

Scratching over a hotcue also activates the indicator.

ronso0:
For testing I created a controller mapping with a callback connection and added a testing proposal to the PR description.

ronso0:
:man_shrugging: 
> The following tests FAILED:
	287 - EngineSyncTest.LoadTrackInitializesLeader (Failed)
	292 - EngineSyncTest.EjectTrackSyncRemains (Failed)
	368 - HotcueControlTest.SavedLoopUnloadTrackWhileActive (Failed)
	546 - AdjustReplayGainTest.AdjustReplayGainUpdatesPregain (Failed)

ronso0:
I simplified it and the tests all pass. Dunno what went wrong and why only these tests failed, couldn't spot a common call that could cause the failure.

JoergAtGithub:
How should this CO behave in case of a saved loop?

JoergAtGithub:
You describe the 3 tests in the description. Why not implementing them as test case?

ronso0:
Because I was lazy and noticed there are no tests for the cue indicator either ; p

I don't think tests are  important as these controls don't affect performance, but of course I can write some.

JoergAtGithub:
> > How should this CO behave in case of a saved loop?
> 
> Exactly like with hotcues (saved loops are _just_ hotcues with an end point).

I wonder if this is what the user expects? I could imagine that he would expect the indicator active, as long as the saved loop is running. Not sure about the use case here.

ronso0:
What you describe is `hotcue_X_status`. The indicator added here is just an addition, as equivalent to `cue_indicator` which is also a _point_ control.

JoergAtGithub:
> What you describe is `hotcue_X_status`. The indicator added here is just an addition, as equivalent to `cue_indicator` which is also a _point_ control.

Doesn't `hotcue_X_status` just reports if the saved loop is enabled? I think it doesn't take into account, if and when the playposition reachs the loop.

ronso0:
What users may expect depends on the documentation.

`hotcue_X_status`
0 not set
1 set
2 loop active or previewing hotcue

If users want a loop indicator the can combine `hotcue_X_status` and `hotcue_X_indicator`.
Or we may extend `hotcue_X_status` with
3 inside active loop

Let's merge this first step, then extend it if requested.

src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2169,44 @@ void CueControl::updateIndicators() {
...
+        const auto cuePos = pControl->getPosition();
+        if (!cuePos.isValid()) {
+            continue;
+        }
+        if ((cuePos >= prevPos && cuePos <= currPos) ||     // forward
+                (cuePos <= prevPos && cuePos >= currPos)) { // reverse
+            pControl->setIndicator(1.0);
+        } else {
+            pControl->setIndicator(0.0);
+        }
```
JoergAtGithub, line 2192: In case of a saved loop and playing reverse, the duration of the loop needs to be added.

ronso0: Good catch, I did not test looping.
And admittedly that is a tricky one. There is a seek from [loop end - x] to [loop in + y] where x + y = buffer length, so we'd need to get/store loop start/end and check loop_enabled as well as whether the hotcue is inside the loop, and that for both play directions. Uff...

JoergAtGithub: Well - I've made some experience with this topic, in #11840 and #12281

ronso0: I think I managed to catch all hotcues correctly.

Manual tests passed (with the trace output and the test mapping).
I'll update the Mixxx tests accordingly.

src/engine/controls/loopingcontrol.h
```diff
@@ -98,8 +86,7 @@ class LoopingControl : public EngineControl {
 
   signals:
     void loopReset();
-    void loopEnabledChanged(bool enabled);
-    void loopUpdated(mixxx::audio::FramePos startPosition, mixxx::audio::FramePos endPosition);
+    void loopUpdated(LoopInfo loopInfo);
```
ronso0, line 89: Why does clazy insist this should be fully qualified when used in engine controls?
It has no issue with `FrameInfo`.

Also, the reason `clazy-fully-qualified-moc-types` warns is
> not using fully-qualified type names [..] will break old-style connects and interaction with QML.

No old style here anymore, and is QML supposed to interact with engine controls??

daschuer: No idea. Does the proposed change work?

ronso0: Well, it builds (didn't run clazy locally), but I wonder if the reason is relevant here.
I'm not aware of any QML <--> EngineControl interaction, QML uses proxies for the library, players etc.
Control interaction is all done through _EngineBuffer_ and the individual controls interacting with _each other_.

So I'd rather ignore it and go with a clazy exclude for now
`// clazy:exclude=clazy-fully-qualified-moc-types`

ronso0: Though it's annoying that it has to be added everywhere it is used.

src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2174,106 @@ void CueControl::updateIndicators() {
...
+                cuePos < loopInfo.endPosition &&
+                (cuePos > prevPos || cuePos <= currPos)) {
+            qWarning() << "     >> activate"
+                       << pControl->getHotcueIndex() + 1
+                       << "(looping wrap-around)";
+            active = 1.0;
+        } else if (loopingWrapAroundReverse &&
+                // Hotcue inside loop?
+                cuePos >= loopInfo.startPosition &&
+                cuePos < loopInfo.endPosition &&
```
daschuer, line 2233: Can we put this into a helper  `isInsideLoop()` ? to deduplicate this a bit?

src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2174,106 @@ void CueControl::updateIndicators() {
...
+        }
+        double active = 0.0;
+        // If we just wrapped around, check if the cue is inside the
+        // processed frame ranges.
+        // Note: loop_end is considered outside the loop.
+        if (loopingWrapAround &&
+                // Hotcue inside loop?
+                cuePos >= loopInfo.startPosition &&
+                cuePos < loopInfo.endPosition &&
+                (cuePos > prevPos || cuePos <= currPos)) {
```
daschuer, line 2225: `cuePos > prevPos` is clear, because the loop has wrapped around, we passed the cue. 
`cuePos <= currPos` This one is not clear for me, doesn't it apply to all cues is is the loop?

src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2174,106 @@ void CueControl::updateIndicators() {
...
+            qWarning() << "     >> activate"
+                       << pControl->getHotcueIndex() + 1
+                       << "(looping wrap-around)";
+            active = 1.0;
+        } else if (loopingWrapAroundReverse &&
+                // Hotcue inside loop?
+                cuePos >= loopInfo.startPosition &&
+                cuePos < loopInfo.endPosition &&
+                (cuePos < prevPos || cuePos >= currPos)) {
```
daschuer, line 2234: The same here: 
`cuePos >= currPos` What is this condition for?

src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2174,106 @@ void CueControl::updateIndicators() {
...
+    if (m_pLoopEnabled->toBool() &&
+            loopInfo.startPosition.isValid() &&
+            loopInfo.endPosition.isValid() &&
+            // Are we inside the current loop?
+            prevPos >= loopInfo.startPosition &&
+            currPos >= loopInfo.startPosition &&
+            prevPos < loopInfo.endPosition &&
+            currPos < loopInfo.endPosition) {
+        // Check if we wrapped around
+        if (reverse && prevPos < currPos) {
```
daschuer, line 2203: Doesn't the condition apply to any seek, independent of a loop? 

I am just considering if it will work more universal we just check if cue position is inside the range of prevPos + the buffer length? 

This will not work if the cue is just after the current loop though. 
In case of seeks the buffer before the seek is cossfaded with the buffer after the seek. Is it inside a loop? 
What is the use case for such a corner case?

daschuer (changes requested):

src/engine/controls/cuecontrol.cpp
ronso0: > Doesn't the condition apply to any seek, independent of a loop?

Seeks are registered in notifySeek() which sets prevPos(), i.e. the loop detection doesn't kick in.

> check if cue position is inside the range of prevPos + the buffer length?

Since the buffer is split into two parts if we wrap around, that's exactly what we do here. (besides, I also considered using LoopingControl's `nextTrigger()` like ReadAheadManager does, or get the trigger already in EngineBuffer so it can pass it to CueControl and RAManager, but that is overkill IMO)

Actually, the original implementation worked just fine (incl. scratching and reverse, NOT for loops),
the current implementation works also for loops, but fails for scratching backwards inside a loop.

Back to square one...

ronso0: okay, "speed" from EngineBuffer works: it is "effective" reverse, i.e. `reverse` or scratching backwards.
now everything works as expected.

src/engine/controls/cuecontrol.cpp
ronso0: since we already detected that we wrapped around, `cuePos <= currPos` covers all cues in between `loop_in` and `currPos`

src/engine/controls/cuecontrol.cpp
ronso0: same as above: when playing reverse, this covers all cues that are after/at `currPos` and before `loop_out`

ronso0:
I tested
* playing forwards & backwards, both with/without looping
* scratching forth and back (stopped and playing)
 
and everything works as expected:
* indicator is activated every time the playpos crosses a hotcue
* seeks are ignored
* hotcues at `loop_out` are correctly ignored (`loop_out` is considered outside the loop, i.e. jumping there would disable the loop)

Now it's just the question if there need to be tests for every case.

daschuer:
> Now it's just the question if there need to be tests for every case.

There is no pressing need for testing every single case. It depends more on you interest to contribute them. 

> Or how much trace output we want/need to troubleshoot issues.

There must be no debug output in the audio thread because it is locking. CueControl::updateIndicators()  is inside the audio thread.

ronso0:
> > Or how much trace output we want/need to troubleshoot issues.
> 
> There must be no debug output in the audio thread because it is locking. CueControl::updateIndicators() is inside the audio thread.

I know, I was more referring to output that can be enabled on demand for troubleshooting.
If this is considered ready for merge I'll remove the TRACE commit.

Re tests, I think the essence of the looping scenario can be covered rather easily.

ronso0:
For now I have removed looping test commit c73a8f1959f06bde2097db6638bf4529bfab621a
if someone can shed a light on the timing in tests, or has a fix handy for the failing (automatic) test we may re-add it.

ronso0:
I've redone this based on the [`wrapAround` notification from ReadAheadManager](https://github.com/mixxxdj/mixxx/pull/13007/commits/50e16c96face8890d7f912385a9ba59f26953ed2) I used for #13007.

The relevant commit is 0256bd4243, test PR is #13076

daschuer:
Yes, the new implementation looks more robust.

ronso0:
So I push to this branch?

daschuer:
Yes, I like it more.

ronso0:
yes, me too, the question was just if you agree to overwrite this one.

daschuer:
Or just close this, I don't mind.

ronso0:
I'll close this one and refine #13076 sometime

PR #12952, opened 2024-03-11
CMakeLists: Require WebGL 2.0 when building for Wasm

fwcd:
We need OpenGL ES 3.0 and by requiring a WebGL 2 browser, we can reduce the size of the binary.

We also need `-sFULL_ES2=1` to avoid warnings like

```
WebGL: INVALID_OPERATION: vertexAttribPointer: no ARRAY_BUFFER is bound and offset is non-zero
```

...and the link options have to be public to apply to the final binary.


daschuer (approved):
LGTM, Thank you.

PR #12954, opened 2024-03-12
Beats: allow undoing the last BPM/beats change

### LINKED ISSUE #12774: Have the "Reset BPM" in the "Adjust BPM" submenu only restore BPM to before any new user change, not fully clear the BPM and beatgrid?
### Feature Description

It would be nice to reset an altered BPM to the number that it was before (on load to the deck I guess), rather than clear the beatgrid, which then requires the extra step of Analysing the track again to get back to where things were.

There's already "BPM and beatgrid" in the "Reset" submenu.

A new phrasing to differentiate? Personally I think "Reset" would be turning it back to the prior number, where as something like "Clear" would be a clearer title for the "Reset" submenu, but if the meaning of "Reset" stays the same, then the "Adjust BPM" reset entry could be maybe "Restore BPM" instead?

ronso0:
Allow undoing the last BPM/beats change.

Up to 10 BeatsPointers are stored on each BPM/beats change.
_Apparently the beat analyzer set beats twice on initial scan, so the first result is added to the stack :man_shrugging:_
For changes done in quick succession (<800 ms currently) intermediate states are not stored.

* add `beats_undo_adjustment` ControlPushButton
* add "Undo last BPM/beats change" to BPM menu

Fixes #12774 (#10138)

#### TODO
- [ ] add Undo button to skins to make this more accessible, probably to beatgrid controls section next to the waveforms
(Tango + LateNight only? in Deere controls are already quite dense)
- [x] add a test?
- [x] remove debug output / transform into trace logs

#### Do we need 'Redo'?
I'd say no. Might be nice to have for comparison, but beats operations are easy to redo, so :man_shrugging: 
It would definitely complicate things.

### DIFF
--- a/src/track/track.cpp
+++ b/src/track/track.cpp
@@ -55,6 +55,8 @@ inline mixxx::Bpm getBeatsPointerBpm(
     return pBeats->getBpmInRange(mixxx::audio::kStartFramePos, trackEndPosition);
 }
 
+constexpr int kMaxBeatsUndoStack = 10;
+
 } // anonymous namespace
 
 // Don't change this string without an entry in the CHANGELOG!
@@ -83,7 +85,8 @@ Track::Track(
           m_fileAccess(std::move(fileAccess)),
           m_record(trackId),
           m_bDirty(false),
-          m_bMarkedForMetadataExport(false) {
+          m_bMarkedForMetadataExport(false),
+          m_undoingBeatsChange(false) {
     if (kLogStats && kLogger.debugEnabled()) {
         long numberOfInstancesBefore = s_numberOfInstances.fetch_add(1);
         kLogger.debug()
@@ -429,6 +432,15 @@ bool Track::setBeatsWhileLocked(mixxx::BeatsPointer pBeats) {
     if (m_pBeats == pBeats) {
         return false;
     }
+    // Don't add null beats to the undo stack. Happens when beats are deserialized,
+    // e.g. when opening the track menu.
+    // Don't add beats to stack which we're about to undo.
+    if (!m_undoingBeatsChange && m_pBeats != nullptr) {
+        if (m_pBeatsUndoStack.size() >= kMaxBeatsUndoStack) {
+            m_pBeatsUndoStack.removeFirst();
+        }
+        m_pBeatsUndoStack.push(m_pBeats);
+    }
     m_pBeats = std::move(pBeats);
     m_record.refMetadata().refTrackInfo().setBpm(getBeatsPointerBpm(m_pBeats, getDuration()));
     return true;
@@ -470,6 +482,18 @@ bool Track::trySetBeatsMarkDirtyAndUnlock(
 mixxx::BeatsPointer Track::getBeats() const {
     const auto locked = lockMutex(&m_qMutex);
     return m_pBeats;
+}
+
+void Track::undoBeatsChange() {
+    if (!canUndoBeatsChange()) {
+        return;
+    }
+
+    auto locked = lockMutex(&m_qMutex);
+    m_undoingBeatsChange = true;
+    const auto pPrevBeats = m_pBeatsUndoStack.pop();
+    trySetBeats(pPrevBeats);
+    m_undoingBeatsChange = false;
 }
 
 void Track::afterBeatsAndBpmUpdated(
--- a/src/track/track.cpp
+++ b/src/track/track.cpp
@@ -8,6 +8,7 @@
 #include "sources/metadatasource.h"
 #include "util/assert.h"
 #include "util/logger.h"
+#include "util/time.h"
 
 namespace {
 
@@ -56,6 +57,9 @@ inline mixxx::Bpm getBeatsPointerBpm(
 }
 
 constexpr int kMaxBeatsUndoStack = 10;
+// The minimum time that has to pass between beat changes to consider them 'separate'.
+// Used to filter actions done in quick succession.
+constexpr int kQuickBeatChangeDeltaMillis = 800;
 
 } // anonymous namespace
 
@@ -96,6 +100,7 @@ Track::Track(
                 << "->"
                 << numberOfInstancesBefore + 1;
     }
+    m_beatChangeTimer.start();
 }
 
 Track::~Track() {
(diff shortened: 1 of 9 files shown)


ronso0:
> I think this can easily be extended to provide N undo steps, but I'm not sure how heavy Beats objects are.

Yet another coffee and it's done :tada: 
* 1o steps are stored
* initial null beats are not stored
* analyzer set beats twice, so the first result is added to the stack :man_shrugging:

ronso0:
Manually everything works as expected, just the tests fail. Need to wrap my head around setting beats and stuff.

src/track/track.h
```diff
@@ -561,6 +566,8 @@ class Track : public QObject {
 
     // Storage for the track's beats
     mixxx::BeatsPointer m_pBeats;
+    QList<mixxx::BeatsPointer> m_pBeatsUndoStack;
```
daschuer, line 569: Here we may use QStack with push() and pop()

daschuer (changes requested):
Thank you. Works good. 

There is a usability issue when you have used the shift actions form the waveform, like "adjuste earliere" ten times. 
You can do it fast via the GUI buttons, but undoing to the original state = 10 x undo is hard via the menu. 

Is there a way to remove intermediate adjustments form the undo stack? I think Gimp has also such a strategy to not undo a brush stroke pixel by pixel.

ronso0:
> There is a usability issue when you have used the shift actions form the waveform, like "adjuste earliere" ten times.
You can do it fast via the GUI buttons, but undoing to the original state = 10 x undo is hard via the menu.
>
> Is there a way to remove intermediate adjustments form the undo stack? I think Gimp has also such a strategy to not undo a brush stroke pixel by pixel.

Sure, but I didn't look into it, yet.
So we want to "group" actions that we consider were done without thinking, so how long does thinking take? .5, .8 seconds?
We may simply store the last change timestamp and pop+push if the diff is smaller than X millis, so we essentially only keep the last action of a "batch".

Maybe such a timeout also helps with the initial useless beats setting.

daschuer:
This is on any case a feature, not blocking this PR.

src/track/track.h
ronso0: Good to learn about QStack. So the benefit is to use push() and pop() instead of append() and takeLast()/removeLast(). Dunno if that is worth the extra include, but that's tiny, so I'll use it.

ronso0:
> This is on any case a feature, not blocking this PR.

Was rather easy, let's check if the tests also work as desired.

The last commit adds an Undo button to the beatgrid edit controls in LateNight.
I made that commit a fixup! to block merging.

ronso0:
> Was rather easy, let's check if the tests also work as desired.

Nope. Maybe the timing is different in the tests?

src/track/track.cpp
```diff
@@ -431,9 +438,52 @@ bool Track::trySetAndLockBeats(mixxx::BeatsPointer pBeats) {
...
+        qWarning() << "       (currently undoing, don't rec)";
+    }
+    if (m_pBeats == nullptr) {
+        qWarning() << "       (currently null beats, don't rec)";
+    }
+
+    // Don't add null beats to the undo stack. Happens when beats are deserialized,
+    // e.g. when opening the track menu.
+    // Don't add beats to stack which we're about to undo.
+    mixxx::Duration time = mixxx::Time::elapsed();
```
daschuer, line 460: I stumbled here for a short moment about "elapsed() since what?".
Maybe we should rename it to `mixxx::Time::sinceStart()` or such. 

In addition we may want to use a "PerformanceTimer" object instead of implementing it inline. 
Using `start()` `restart()` and `elapsed()`

mxmilkiib:
I'm not sure one can bind x2.

Also, a strange sort order;

image

ronso0:
Wrong PR?

mxmilkiib:
Yeeeep

ronso0:
Manually this works just fine, but I need help with the test (or merge without it).

src/track/track.cpp
daschuer: Did you consider this?

daschuer:
What is your plan here? Remove form 2.5-beta milestone?

ronso0:
I need help with the tests, then this is ready.
Or we add the tests later on so we can already get feedback during beta.

src/track/track.cpp
ronso0: I'm considering this, yes.
Pro argument for the current solution is that we don't create an extra timer object, we simply use the Mixxx time.
A performance timer would be running all the time **for each track we touch**, even if the user doesn't change the beats (apart from initial load/analyse). A regular one-shot timer (auto-stop) might also work (delta is 800 ms).
Hence I don't really see the necessity to change this.

daschuer: First of all, I have to a strong demand here. We can keep it as it is. 

But let's give some insights: 
The PerformanceTimer object not "running" in terms of CPU. It is just a thin wrapper around a single member variable that stores the start time. See: https://github.com/mixxxdj/mixxx/blob/7111b70e3fa346b67bab3f275b8cce037ba6274b/src/util/performancetimer.h#L37
So basically your current solution wrapped by an API.

ronso0: Oh, sorry, should have looked closer.
I'll try that!

ronso0:
I've discovered some issue with the quick actions.
Don't want to fix this in a hurry (might take some time to do it right anyways), so let's remove it from 2.5-beta.

ronso0:
> I've discovered some issue with the quick actions. (that might also be the reason why the test fails...)

Jup, that was it, the test passes.

Ready to remove the Undo button and the trace output.

daschuer (approved):
LGTM, ready for clean up. Thank you.

ronso0:
The manual update is here https://github.com/mixxxdj/manual/pull/654

daschuer:
Thank you

m0dB:
@ronso0 Did you consider using QUndoStack, and if so, why did you prefer to implement your own stack?

I used QUndoStack for cue add/remove: https://github.com/mixxxdj/mixxx/pull/13469 . I have only looked briefly at your implementation, but I think it will be trivial to use the QUndoStack instead of the custom stack for the beat changes, but maybe I am missing something.

ronso0:
No, I was not aware of QUndoStack. I'll comment in #13469 re the details.

PR #12956, opened 2024-03-12
src/library: sort files into sub-directories

ronso0:
(xeruf-style PR :stuck_out_tongue: of an old branch of mine)

... just to make finding files in that huge /library a bit easier, scrolling in a file tree for example.


JoergAtGithub:
LGTM! Local build works as before! Thank you!

PR #12957, opened 2024-03-12
Download cleanup

daschuer:
This adds a manual workflow that deletes files from downloads.mixxx.org 
This deletes only a few files for testing. In case this is successful. I will add another PR that removes all obsolete files to free up space.

### DIFF
--- a/.github/workflows/download_cleanup.yml
+++ b/.github/workflows/download_cleanup.yml
@@ -1,0 +1,34 @@
+name: Clean up downloads
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: "Set up SSH Agent"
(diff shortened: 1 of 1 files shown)


ronso0:
FWIW http://downloads.mixxx.org/snapshots/1.12/debug/ and http://downloads.mixxx.org/snapshots/2.0/debug/ _seem_ identical. Is one of them just a symlink?
Deleting one of them (1.12?) would already free ~3.5 GB

daschuer:
Probably is is a symlink. We need to be careful to not delete files in both we want to keep in one. 
But do we need 2.0 snapshots? Probably we can remove altogether. 

Let's first check if the manual workflow does what it should.

ronso0:
Shall we merge this now?

daschuer:
Yes, if you don't spot and obvious issue.

ronso0:
Well, I don't because I'm not familiar with the runners syntax.
"LGTM" since it matches the syntax in the Build runner :man_shrugging:

PR #12961, opened 2024-03-13
Looping/Beatjump: use seconds if track has no beats

### LINKED ISSUE #11124: make Beatjump use seconds if no beatgrid is detected
### Feature Description

Make Beatjump controls switch from beats to seconds if no beatgrid is detected so these controls are of some use at least.
This is useful for beat-less tracks as well as jumping in long tracks while they're being analyzed.

* the beatjump size spinbox could show the `s` suffix
* I guess making the loop engine deal with seconds is too much hazzle, so the beatloop size spinbox and the loop activate button (with no saved loop) should probably be greyed out

ronso0:
Finally allows looping and beatjumping for tracks with no beats.

If the track has no beats, simply load fake beats: 60 BPM, first beat at track start.
1 'beat' = 1 second
Looping and beatjumping works as desired :tada: 

**edit:** Also sets a `m_trueTrackBeats` flag so loops **don't** unexpectedly snap to the (invisible) fake beats.

For completeness we'd need to add one string to adjust all related tooltips, and one for all items in controlpickermenu.cpp.
If this string freeze viloation is considered not worth for having this nice fix in 2.4.1, or if you think this is definitely a new feature, I'll rebase onto main.

### TODO
- [x] update tooltips
- [x] update controlpickermenu
- [ ] update manual https://github.com/mixxxdj/manual/pull/630

Fixes #11124


ronso0:
> let's see if the tests pass.

Ah, of course not, they assume no-ops for no beats.
I'll take a look.

ronso0:
I expected that Looping tests would fail now, but why on earth do they pass on macOS arm64??

ronso0:
All tests pass :tada: 

So it was just a missing quantize had to be disabled where it is not explicitely expected/required :roll_eyes:

daschuer:
Thank you for taking care. 

An alternative solution would be a default of 120 bpm. 
That will likely be closer to the usual distance in other tracks. But destroys the easy to explain "1 'beat' = 1 second" assumption. 

I have no strong opinion here, I just want to make sure we fix the issue optimal. 

What are the use-cases? Do we have one, or is it just to do "something" that the buttons not feel broken? 

I can think of vokal tracks, ambient tracks. Currently Mixxx applies a beat grid to all of them l, so the user has artificial removed and locked the bpm for some reason.

ronso0:
IMO "1 'beat' = 1 second" makes it predictable.

Yes, ambient tracks. Tbh me personally ran into this situation only a few times in live situations when attempting to mix in  ambient tracks. The UX is currently like this:
* play a mix with beats tracks
* load an ambient track (I explicitely clear the beats for those when tagging new tracks)
* yeah, I want to loop this!
* hit a beatloop button
* nothing happens
* why?
* oh, no beats..
* (either just drop the lopping idea because those few crucial seconds passed until I realized what happens, and a loop wouldn't make sense anymore, or)
* jump back, set loop manually

I use a encoder, so I usually press (4 beats) and instantly turn it to adjust the loop as desired (visually, to either constrain the loop to beatless regions, or the way around, avoid a beatless tail).
Using seconds right away would be a huge step forwrads in terms of UX.

ronso0:
FWIW I didn't test with quantize yet.
I can't really judge if the new behaviour somehow conflicts with users' expectations if quantize is still enabled, because "no (visible) beats" would actually mean "no quantize".

In order to clarify that situation and **not** snap seconds loops to invisible beats, we may consider to also set a `m_fakeBeats` flag when applying the fake beats and use that to override 'quantize enabled'.
Then cue/hotcue behaviour and looping behaviour would be consistent whit quantize and no beats.

src/engine/controls/loopingcontrol.cpp
```diff
@@ -1228,15 +1228,18 @@ void LoopingControl::trackLoaded(TrackPointer pNewTrack) {
...
-                    loopInfo.startPosition, loopInfo.endPosition);
-            if (loaded_loop_size != -1) {
-                m_pCOBeatLoopSize->setAndConfirm(loaded_loop_size);
-            }
+    if (pBeats) {
+        m_pBeats = pBeats;
+    } else {
+        m_pBeats = getFake60BpmBeats();
+    }
+    // TODO All "if (m_pBeats)" checks are now obsolete actually...
```
daschuer, line 1236: Only if we also initialize it in the constructor. This would be the first use case of a not-null smart pointer. A fun project out of the original scope here.

daschuer:
> m_fakeBeats

This flag sounds reasonable.

ronso0:
Added the flag, now waiting curiously how the tests like that...

ronso0:
Cooool, it works and tests pass.

Bugfix or feature?
2.4.1 or main?

I'd argue for bugfix, but I don't mind moving this to 2.5 in order to increase the chance of getting completed translations.

JoergAtGithub:
IMHO, this is a feature with a risk of side effects and should go into Main therefore.

daschuer:
What type of side effects do you expect? 
* Before: buttons have no function
* Now: Buttons work

JoergAtGithub:
I'm not aware of any particular problem, but information about loop length is used at many different places of our code (mappings, skins, waveforms, vsynthread), therefore the probability to break something is high.

ronso0:
I don't see what could be broken:
* the fake beats added here are used only in these looping/beatjump classes
* the loops created by them are treated like any other beatloop
* other classes don't care about the loop/beat relation, no effect on other behaviour
* all looping and beatjump tests pass

Note I don't mind to move this to 2.5, but only if there are valid concerns.

JoergAtGithub:
> 
>     * other classes don't care about the loop/beat relation, no effect on other behaviour

At least clockcontrol takes the beat/loop relation into account too.

ronso0:
Sorry, my phrasing was unclear.

Yes, ClockControl uses the **track** beats and the LoopInfo COs.

What I mean is that no other class gets the beats from LoopingControl, the fake beats are entirely internal for that class. Other classes (EngineBuffer) only get the LoopInfo and that does not contain any beat information.

acolombier:
> An alternative solution would be a default of 120 bpm.

I do agree with this too. Perhaps we could introduce a user preference for this? This way we can default to 60 BPM but also allow user to have other default depending of the BPM range they usually works with?

ronso0:
Yup, we can certainly do that.
Though I'd like to get this merged first.

ronso0:
Moved this to 2.5-beta

daschuer:
Do we have a conclusion for the merge target here? @JoergAtGithub?

JoergAtGithub:
This is an enhancement and not fixing a bug. My opinion is still, that this is too risky for a bugfix release!

src/engine/controls/loopingcontrol.cpp
```diff
@@ -1228,15 +1231,20 @@ void LoopingControl::trackLoaded(TrackPointer pNewTrack) {
...
-                m_pCOBeatLoopSize->setAndConfirm(loaded_loop_size);
-            }
+    if (pBeats) {
+        m_pBeats = pBeats;
+        m_trueTrackBeats = true;
+    } else {
+        m_pBeats = getFake60BpmBeats();
+        m_trueTrackBeats = false;
+    }
+    // TODO All "if (m_pBeats)" checks are now obsolete actually...
```
JoergAtGithub, line 1241: Than these `if (m_pBeats)` need to be removed from the code

src/test/hotcuecontrol_test.cpp
```diff
@@ -1092,28 +1092,6 @@ TEST_F(HotcueControlTest, CueLoopWithSavedLoopToggles) {
...
-    m_pHotcue1CueLoop->set(0);
-
-    EXPECT_DOUBLE_EQ(static_cast<double>(HotcueControl::Status::Set), m_pHotcue1Enabled->get());
-    EXPECT_FRAMEPOS_EQ_CONTROL(mixxx::audio::kStartFramePos, m_pHotcue1Position);
-    EXPECT_FALSE(mixxx::audio::FramePos::fromEngineSamplePosMaybeInvalid(
-            m_pHotcue1EndPosition->get())
-                         .isValid());
-    EXPECT_FALSE(m_pLoopEnabled->toBool());
-}
-
```
JoergAtGithub, line 1116: Why did you remove these unit tests? Now we have more cases to test and not less.

JoergAtGithub (changes requested):

src/engine/controls/loopingcontrol.cpp
ronso0: I have the impression that it is used as `if (m_pTrack)`.
I'll take a look.

src/test/hotcuecontrol_test.cpp
ronso0: This tests that there is no (cue) loop created if the track has no beats. This situation will not occur anymore (with a track loaded).

We don't need to test "no beats" cases --> less tests.
What new cases are you referring to?

ronso0:
> This is an enhancement and not fixing a bug. My opinion is still, that this is too risky for a bugfix release!

I already assigned it to the 2.5-beta milestone, just need to change the base branch.

I'm still curious where you see the risk, please elaborate.

src/engine/controls/loopingcontrol.cpp
ronso0: If refined updateBeats() so we now have a Beats nullptr again if no track is loaded, and all `if (m_pBeats)` checks can stay.

daschuer (approved):
LGTM. We have not yet a formal beta so it should be OK to have the string changes included.

JoergAtGithub:
Thank you!

PR #12965, opened 2024-03-13
(fix) DlgTrackInfo: prevent wiping metadata when applying twice quickly

### LINKED ISSUE #12963: applying track properties quickly twice will delete track metadata
### Bug Description

applying track properties **quickly twice** will delete track metadata.

## Expected Behavior

none

## Current Behavior

track metadata will be deleted

## Steps to Reproduce

1. Open track properties
2. Apply and apply quickly

### log and screenrecord

[mixxx.log](https://github.com/mixxxdj/mixxx/files/14591344/mixxx.log)

https://github.com/mixxxdj/mixxx/assets/1793763/6478957f-dc59-4722-97b9-a449024ca313

### file in filesystem information

The affected file and the backup in filesystem are still the same.

### sqlite information

I didn't check all table, but the value in the `library` table became NULL in mixxx.sqlite .

Here is the records of the affected file(1) and the backup(2).

```sql
SELECT t.* FROM library AS t
-- I tested new DB so I didnt need any WHERE clause
;
```

id | artist | title | album | year | genre | tracknumber | location | comment | url | duration | bitrate | samplerate | cuepoint | bpm | wavesummaryhex | channels | datetime_added | mixxx_deleted | played | header_parsed | filetype | replaygain | timesplayed | rating | key | beats | beats_version | composer | bpm_lock | beats_sub_version | keys | keys_version | keys_sub_version | key_id | grouping | album_artist | coverart_source | coverart_type | coverart_location | coverart_hash | replaygain_peak | tracktotal | color | coverart_color | coverart_digest | last_played_at | source_synchronized_ms
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 |   |   |   |   |   |   | 1 |   |   | 136.176 | 128 | 44100 | 0 | 0.0 |   | 2 | 2024-03-13T17:13:10.438Z | 0 | 0 | 1 |   | 0.305140602914104 | 0 | 0 |   |   |   |   | 0 |   |   | KeyMap-1.0 |   | 0 |   |   | 1 | 1 |   | 61045 | 1.91240704059601 |   |   | 15000290 |   |   |  
2 | Nobuo Uematsu | Spinach Rag | FINAL FANTASY VI Original Soundtrack | 2004 | Other | 6 | 2 |   |   | 136.176 | 128 | 44100 | 0 | 0.0 |   | 2 | 2024-03-13T17:13:10.452Z | 0 | 0 | 1 | mp3 | 0.305140602914104 | 0 | 0 |   |   |   |   | 0 |   |   | KeyMap-1.0 |   | 0 |   | Nobuo Uematsu | 1 | 1 |   | 61045 | 1.91240704059601 | 20 |   | 15000290 |   |   | 1630260428000

## Others

This forum thread may be relevant. https://mixxx.discourse.group/t/mixxx-2-4-metadata-bug-properties/29083 

### Version

2.4.0

### OS

Windows 11

ronso0:
Not sure what exactly is the root cause, but preventing what should be a [no-op in `Track`](https://github.com/mixxxdj/mixxx/blob/24b415c4ff7e3bfd5e418ad31e03c4c627367c3e/src/track/track.cpp#L298) already  in `DlgTrackInfo` seems to suffice. Can't reproduce anymore.

With my sparse basic c++ knowledge, I suspected the cause is that the track record has been moved to Track (is null in DlgTrackRecord?) when applying again before the `Track::changed` signal triggered DlgTrackInfo to fetch the new track record.
But tracing that turned out this is not it since now the changed signal is received _before_ trying to save again (maybe it's just delays caused by the new m_pLoadedTrack->getRecord(), idk :man_shrugging: )

With symptoms fixed this good enough for 2.4.1 I guess.

#### Update:
As [clarified by @uklotzde](90cdb1fa3fd7b89e39beb69e364d8b5ae9e10f4c) the issue was indeed passing the record and the beats to track via `std::move`.

Fixes #12963

### DIFF
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -614,8 +614,13 @@ void DlgTrackInfo::saveTrack() {
     static_cast<void>(updateKeyText()); // discard result
 
     // Update the cached track
-    // The dialog is updated and repopulated by the Track::changed() signal.
-    m_pLoadedTrack->replaceRecord(std::move(m_trackRecord), std::move(m_pBeatsClone));
+    //
+    // If replaceRecord() returns true then both m_trackRecord and m_pBeatsClone
+    // will be updated by the subsequent Track::changed() signal to keep them
+    // synchronized with the track. Otherwise the track has not been modified and
+    // both members must remain valid. Do not use std::move() for passing arguments!
+    // Else triggering apply twice in quick succession might clear the metadata.
+    m_pLoadedTrack->replaceRecord(m_trackRecord, m_pBeatsClone);
 }
 
 void DlgTrackInfo::clear() {
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -452,8 +452,8 @@ void DlgTrackInfo::reloadTrackBeats(const Track& track) {
     updateSpinBpmFromBeats();
     m_trackHasBeatMap = m_pBeatsClone && !m_pBeatsClone->hasConstantTempo();
     bpmConst->setChecked(!m_trackHasBeatMap);
-    bpmConst->setEnabled(m_trackHasBeatMap); // We cannot make turn a BeatGrid to a BeatMap
-    spinBpm->setEnabled(!m_trackHasBeatMap); // We cannot change bpm continuously or tab them
+    bpmConst->setEnabled(m_trackHasBeatMap); // We cannot turn a BeatGrid to a BeatMap
+    spinBpm->setEnabled(!m_trackHasBeatMap); // We cannot change bpm continuously or tap them
     bpmTap->setEnabled(!m_trackHasBeatMap);  // when we have a beatmap
 
     if (track.isBpmLocked()) {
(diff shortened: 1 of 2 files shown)


ronso0:
@zfhrp6 Please test this. Either build from source yoursel or pick a CI build (info is here https://github.com/mixxxdj/mixxx/wiki/Testing).

zfhrp6:
@ronso0 I've tested the CI build (actions/runs/8273014776). The steps to repro in issue can't reproduce, and repeating Alt+A also can't. 👍

src/library/dlgtrackinfo.cpp
```diff
@@ -613,6 +613,13 @@ void DlgTrackInfo::saveTrack() {
...
+    // Don't update if nothing changed. This prevents wiping the record
+    // when calling this function twice quickly.
+    // See https://github.com/mixxxdj/mixxx/issues/12963
+    if (m_trackRecord == m_pLoadedTrack->getRecord()) {
+        return;
+    }
+
     // Update the cached track
     // The dialog is updated and repopulated by the Track::changed() signal.
     m_pLoadedTrack->replaceRecord(std::move(m_trackRecord), std::move(m_pBeatsClone));
```
uklotzde, line 625: Sorry, my bad.

```suggestion
    // Update the cached track
    //
    // If replaceRecord() returns true then both m_trackRecord and m_pBeatsClone
    // will be updated by the subsequent Track::changed() signal to keep them
    // synchronized with the track. Otherwise the track has not been modified and
    // both members must remain valid. Do not use std::move() for passing arguments!
    m_pLoadedTrack->replaceRecord(m_trackRecord, m_pBeatsClone);
```

Without warranty. Consider refactoring the API to prevent such misunderstandings in the future.

ronso0: Thanks for jumping in!

Okay, so this is the actual fix.
IIUC the no-op test I added above just hides the issue because it's unlikely the record (in DlgTrackInfo) has changed in between two quick save() calls. Correct?

ronso0: Also, IIUC this
https://github.com/mixxxdj/mixxx/blob/e2d41999e7c017c6a353f87967152d3bd2102354/src/track/track.cpp#L308-L310 would already `move` the beats even if applying them fails and replaceRecord() returns true?

src/library/dlgtrackinfo.cpp
```diff
@@ -613,9 +613,20 @@ void DlgTrackInfo::saveTrack() {
     slotSpinBpmValueChanged(spinBpm->value());
     static_cast<void>(updateKeyText()); // discard result
 
+    // Don't update if nothing changed.
+    if (m_trackRecord == m_pLoadedTrack->getRecord() &&
+            m_pBeatsClone == m_pLoadedTrack->getBeats()) {
```
ronso0, line 618: In case metadata or beats have been changed elsewhere we'd already have updated here as well, so this scheck is safe IMO (i.e. does **not** undo those changes by re-applying this dialog's (unchanged) record and beats).

Admittedly, since the update is triggered by signals there is a slight chance users revert their own manual changes, but I'd consider that a "user error" because working on metadata in multiple places at once (at inhuman high rate) may inevitably create incosistencies.

Any thoughts?

src/library/dlgtrackinfo.cpp
uklotzde: It only moves out of the argument which is passed by value. No side effects on the outer context.

ronso0:
Has someone time to review?

daschuer (changes requested):

src/library/dlgtrackinfo.cpp
```diff
@@ -613,9 +613,20 @@ void DlgTrackInfo::saveTrack() {
...
+
     // Update the cached track
-    // The dialog is updated and repopulated by the Track::changed() signal.
-    m_pLoadedTrack->replaceRecord(std::move(m_trackRecord), std::move(m_pBeatsClone));
+    //
+    // If replaceRecord() returns true then both m_trackRecord and m_pBeatsClone
+    // will be updated by the subsequent Track::changed() signal to keep them
+    // synchronized with the track. Otherwise the track has not been modified and
+    // both members must remain valid. Do not use std::move() for passing arguments!
+    // See https://github.com/mixxxdj/mixxx/issues/12963
```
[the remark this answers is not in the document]
ronso0: Well, it describes the symptoms, but I can do that here as well.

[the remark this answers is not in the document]
daschuer: I am already fine without it. Then let's keep it. 
No need for extra work.

src/library/dlgtrackinfo.cpp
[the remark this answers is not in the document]
ronso0: I'll revert that commit. I don't want to complicate DlgTrackInfo with a dirty flag, this is just a compact, effecient bugfix after all.

The reason I considered this check in DlgTrackInfo is that `Track::replaceRecord()` does [`std::move(optionalBeats)` too early](https://github.com/mixxxdj/mixxx/blob/7b4f3619e03dae7f88d6ca636029b38dc0092f7c/src/track/track.cpp#L308) (before it's clarified if beats can be applied), betas are moved later on in [`Track::setBeatsWhileLocked()`](https://github.com/mixxxdj/mixxx/blob/7b4f3619e03dae7f88d6ca636029b38dc0092f7c/src/track/track.cpp#L433C6-L433C33-L440) only if applicable.
So beats might be moved even though they're not applied and `Track::updated` would not be emitted and we might get the same situation with the beats which this PR fixes for the other metadata.

daschuer (approved):
Thank you. LGTM

PR #12973, opened 2024-03-17
Fix hid addOutput

JoergAtGithub:
One of the issues fixed in https://github.com/mixxxdj/mixxx/pull/12876 for the Traktor F1 mapping was, that `HIDPacket.addControl` was used to define Outputs. The correct command is `HIDPacket.addOutput` instead.
I found two other mappings containing the same bug, which I fixed in this PR.

In case of the PioneerDJ CDJ mapping, I doubt that this code worked before (at least for 2.3, I can't see how it could work), because the byte offset starts with 0, which points to the byte with the ReportID. Furthermore, two output reports with the same ReportID 0x01 are defined in this mapping. I didn't changed this, because without hardware, I can't verify, what are the correct numbers.


daschuer:
Thank you for taking care.

eslint is complaining
`62:9  error  'packet' is not defined  no-undef`
A false positive?

Swiftb0y (commented):
can you push the pre-commit patch? lgtm otherwise.

JoergAtGithub:
Can we merge this?

daschuer (approved):
All green now. Thank you.

PR #12977, opened 2024-03-17
Removed old examples HID keyboard and HID trackpad

JoergAtGithub:
These examples are only empty stubs of a mapping. A much better example for a HID mapping, is the mapping for the simple Trakor Kontrol Z1.

And while keyboard, mouse and touchpad are HID devices, most operating systems restricted direct access to them via the user space HID APIs nowadays. This is for security reasons, and should prevent password sniffers running with normal user rights.


daschuer (approved):
LGTM, Thank you

PR #12978, opened 2024-03-18
(fix) Track widgets: set `show_track_menu` only for main decks

ronso0:
only for those BaseTrackPlayer created the ControlPushButtons.

Currently a DEBUG_ASSERT is hit if e.g. a sampler's track label receives a contextMenuEvent.
(apparently no one uses that track menu... I just discovered this by accident X)

I'd actually have pushed this minimal fix directly but a second pair of eyes would be good.


daschuer (approved):
LGTM, Thank you.

PR #12990, opened 2024-03-22
animate long press latching

m0dB:
Animate long press latching (e.g. sync button) by capture the off-state in a pixmap and during the long press latch time, gradually draw less of it.

https://github.com/mixxxdj/mixxx/assets/79429057/788ee1f5-bb5c-4e2a-9ba1-884d1610accc

### DIFF
--- a/src/widget/wpushbutton.cpp
+++ b/src/widget/wpushbutton.cpp
@@ -434,8 +434,8 @@ void WPushButton::mousePressEvent(QMouseEvent * e) {
                 m_clickTimer.start(ControlPushButtonBehavior::kLongPressLatchingTimeMillis);
                 if (oldValue == 0.0) {
                     // Capture pixmap with the off state ...
-                    m_preLongPressPixmap = QPixmap(width() * devicePixelRatio(),
-                            height() * devicePixelRatio());
+                    m_preLongPressPixmap = QPixmap(static_cast<int>(width() * devicePixelRatio()),
+                            static_cast<int>(height() * devicePixelRatio()));
                     m_preLongPressPixmap.setDevicePixelRatio(devicePixelRatio());
                     paintOnDevice(&m_preLongPressPixmap);
                     // ... and start the long press latching animation
--- a/src/widget/wpushbutton.cpp
+++ b/src/widget/wpushbutton.cpp
@@ -414,7 +414,7 @@ void WPushButton::mousePressEvent(QMouseEvent * e) {
             if (m_leftButtonMode == ControlPushButton::LONGPRESSLATCHING) {
                 m_clickTimer.setSingleShot(true);
                 m_clickTimer.start(ControlPushButtonBehavior::kLongPressLatchingTimeMillis);
-                if (oldValue == 0.0) {
+                if (oldValue == 0.0 && m_pLongPressLatching) {
                     m_pLongPressLatching->start();
                 }
             }
--- a/src/widget/wpushbutton.cpp
+++ b/src/widget/wpushbutton.cpp
@@ -228,6 +228,9 @@ void WPushButton::setup(const QDomNode& node, const SkinContext& context) {
     }
 
     setFocusPolicy(Qt::NoFocus);
+
+    // To animate the long press latching
+    connect(&m_animTimer, &QTimer::timeout, this, &WPushButton::onAnimTimer);
 }
 
 void WPushButton::setStates(int iStates) {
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
I just tested this on Windows11, and the buttons appear fluidly animated and make the long press behavior really easier to understand.
And even for me as a experienced user, who is aware about the long press sync button functionality, it simplifies the use, because now I see how long I need to hold the button.

m0dB:
Great! Happy you like it, and thanks for validating!

m0dB:
@daschuer @Swiftb0y @ronso0 could you test on linux?

Any objections against this approach?

acolombier (approved):
Looking nice and smooth on Linux (PopOs 22.04 Wayland)


https://github.com/mixxxdj/mixxx/assets/7086688/52163df6-7ce8-4e9e-9869-0cc05f9d74ff

m0dB:
Thanks, Antoine!

ywwg (changes requested):
my main note is that this gives every single pushbutton in the UI the latching logic and support objects even if they are never used.  Would it be possible to make this a child object of a pushbutton (WPushButtonLatching) that inherits WPushButton and just adds on the new stuff?  I think that would also be cleaner from a skin design perspective.

src/widget/wpushbutton.cpp
```diff
@@ -395,12 +425,22 @@ void WPushButton::mousePressEvent(QMouseEvent * e) {
...
                 m_clickTimer.setSingleShot(true);
                 m_clickTimer.start(ControlPushButtonBehavior::kLongPressLatchingTimeMillis);
+                if (oldValue == 0.0) {
+                    // Capture pixmap with the off state ...
+                    m_preLongPressPixmap = QPixmap(static_cast<int>(width() * devicePixelRatio()),
+                            static_cast<int>(height() * devicePixelRatio()));
+                    m_preLongPressPixmap.setDevicePixelRatio(devicePixelRatio());
+                    paintOnDevice(&m_preLongPressPixmap);
+                    // ... and start the long press latching animation
+                    m_animTimer.start(1000 / 60);
```
ywwg, line 442: if I understand correctly, this creates an independent 60fps timer to update this animation.  Would it be possible / feasible to connect instead to the global UI update timer?  or does it not really matter because it's just updating the position data, and the drawing is done elsewhere?

m0dB: I doesn't matter. The overhead of the timer is minimal and it doesn't do the actual drawing, it just triggers the drawing when Qt sees fit. This is the recommended way.

ronso0:
> my main note is that this gives every single pushbutton in the UI the latching logic and support objects even if they are never used.

These objects / connections can be created only if `m_leftButtonMode == ControlPushButton::LONGPRESSLATCHING` in `setup()`?

m0dB:
> my main note is that this gives every single pushbutton in the UI the latching logic and support objects even if they are never used. Would it be possible to make this a child object of a pushbutton (WPushButtonLatching) that inherits WPushButton and just adds on the new stuff? I think that would also be cleaner from a skin design perspective.

It would be possible but IMO this would lead to added complexity in the code rather than the opposite. Also, WPushButton already implements several modes / behaviours as it is, so why only isolating this one? The added overhead is minimal.

Also, as far as I remember, the skin parser is not instantiating different classes based on properties anywhere, so it would be strange to add that only for this particular case.

But I see your objection that this does add code to WPushButton that is only used for a very specific use case. What I could do is place the QTimer, the QPixmap, and the latching related code in a class that is a member of WPushButton, reducing the overhead to an std::unique_ptr and some function calls.

m0dB:
> > my main note is that this gives every single pushbutton in the UI the latching logic and support objects even if they are never used.
> 
> These objects / connections can be created only if `m_leftButtonMode == ControlPushButton::LONGPRESSLATCHING` in `setup()`?

Yes.

I have moved the code and objects to nested class `WPushButton::LongPressLatching`, which is a instantiated as member `std::unique<LongPressLatching> m_pLongPressLatching` only if `m_leftButtonMode == ControlPushButton::LONGPRESSLATCHING`.

@fwcd let me know if this addresses your concerns.

fwcd:
I think you wanted to ping @ywwg? 😄

m0dB:
Indeed! All those 4 letter usernames! 😅

src/widget/wpushbutton.h
```diff
@@ -106,12 +106,24 @@ class WPushButton : public WWidget {
...
-    QPixmap m_preLongPressPixmap;
-
     // short click toggle button long click push button
     ControlPushButton::ButtonMode m_leftButtonMode;
     ControlPushButton::ButtonMode m_rightButtonMode;
     QTimer m_clickTimer;
-    QTimer m_animTimer; // To animate long press latching
     QVector<int> m_align;
+
+    class LongPressLatching {
```
ywwg, line 115: please add an explanatory comment for this new class that summarizes the feature

ywwg (approved):
I like this refactor! I agree since wpushbutton already does a lot of things it doesn't make sense to create a subclass. thanks. 

One more question, is the animation triggered when someone long-presses on a controller? I am guessing right now it only works when initiated with the UI and a mouse. It seems out of scope for this PR to implement this for controllers but we probably want to do that too

src/widget/wpushbutton.h
m0dB: Will do.

m0dB: Comment added. If you find it's sufficient, could you merge this? Thanks!

daschuer:
```
/home/daniel/workspace/mixxx2/src/widget/wpushbutton.h:134:10: error: ‘unique_ptr’ in namespace ‘std’ does not name a template type
  134 |     std::unique_ptr<LongPressLatching> m_pLongPressLatching;
      |          ^~~~~~~~~~
```

`#include <memory> ` is missing.

daschuer (changes requested):
I can confirm it is working nice after adding the include.

src/widget/wpushbutton.h
ywwg: @daschuer is the last outstanding reviewer

ronso0:
@daschuer Okay to merge?

daschuer:
After adding the missing include, yes.

PR #12994, opened 2024-03-23
show beats and time until next marker

m0dB:
Show the number of beat and/or time until the next marker, inside the waveform.

https://github.com/mixxxdj/mixxx/assets/79429057/925f5bfa-fa9c-4623-9c71-7bf9f80a212d

The digits are rendered using a texture (generated on the fly), for two reasons:
- adding a dark blurred outline so the text is readable even when displayed on top of the waveform (similar to tv subtitles). 
- efficiency

The display can be configured in the settings:

E.g:
Screenshot 2024-04-07 at 18 52 02

Looks like this:
Screenshot 2024-04-07 at 18 52 25


m0dB:
I pushed a wrong commit. correct now.

fwcd:
Neat, will definitely have to play around with this PR once I find some time. Just curious, how does this handle markers that are slightly off, i.e. not sample-perfectly aligned to the grid? (My library has quite a few of them since I tend to have quantization off) Are the beats rounded or floored?

m0dB:
That's a good question! I only have my markers on the beat 😄 

It uses ```Beats::ConstIterator Beats::iteratorFrom(audio::FramePos position)```, which does ```std::lower_bound```, so floor. I think this makes sense also conceptually for markers that are not on the beat. After all, it's until the beat that contains the marker.

ywwg:
I have been thinking about exactly this sort of UX addition, thank you for working on it.  In my own head, I was thinking about putting this information in the overview rather than the waveform itself. Right now if the mouse is over the next marker, I get a bunch of useful information, including time left until that marker. What if we always showed that information for the next marker in the track?  And then the user could choose between beat counts and time counts.  That would keep the big waveform clear of numbers.

As written, from a UX perspective, I would prefer the numbers to be at the top of the play marker, not right in the middle where it visually crowds the waveform display.

fwcd:
> I think this makes sense also conceptually for markers that are not on the beat. After all, it's until the beat that contains the marker.

Maybe we could count full beats relative to the next marker? That would take care of the issue, regardless of how misaligned they are. Though that might become challenging with variable grids.

By the way, on an unrelated note, being able to turn quantization on only for setting markers (but not start/stop or jumping to them) would be really awesome.

ywwg:
I think we could round the beat count if the marker is not exactly on the beat.

src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include "waveform/renderers/allshader/vertexdata.h"
+
+// Render digits using a texture (generated) with digits with blurred dark outline
+
+namespace {
+int charToIndex(char ch) {
+    if (ch >= '0' && ch <= '9') {
+        return static_cast<int>(ch - '0');
+    }
+    if (ch == ':') {
```
daschuer, line 23: This and the condition below require a source code comment.

daschuer: Maybe you should also add a indexToChar() to make the code understandable.

src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -0,0 +1,176 @@
...
+    const int n = s.length();
+
+    VertexData posVertices;
+    VertexData texVertices;
+
+    posVertices.reserve(n * 6); // two triangles per character
+    texVertices.reserve(n * 6);
+
+    for (int i = 0; i < n; i++) {
+        int index = charToIndex(s[i].cell());
```
daschuer, line 137: This code is hard to understand. Maybe you should pass a char array or a QByteArray instead?

daschuer (changes requested):
Thank you. Some comment.

Holzhaus:
That looks really nice, thanks. I also love that you seem to have the habit of adding Screencast to your PRs so that it's easy to see the result w/o checking out and building the changes.

One easy improvement I can think of when looking at the recordings is to not center the information vertically. It's a bit hard to read with the signal in the background, so I'd suggest to align them them at the top where the bg is black (at the bottom it would conflict with hotcue labels).

m0dB:
I have added options in the settings dialog to configure what is displayed (none, beats, time, beats + time, beats + time (multiline)), where it is displayed (top, center, bottom), and the font size.

m0dB:
> do you have more you want to do before marking this as ready for review? I think in general people like the feature idea!

no, I think it is ready for review, once I resolve the conflicts with main.

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: done!

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: improved.

m0dB:
This is now ready for review.

JoergAtGithub (changes requested):
The preferences option appear for all waveform types, but the feature works only for the allshaders GLSL waveforms. Either disable the settings for the unsupported waveform types, or add support for all waveform types.

JoergAtGithub:
Is it the intended behavior, that loop end is considered as marker?

m0dB:
> The preferences option appear for all waveform types, but the feature works only for the allshaders GLSL waveforms. Either disable the settings for the unsupported waveform types, or add support for all waveform types.

I am definitely not going to add support for all waveform types, but I will disable them and add a "Requires GLSL waveform type" label. Thanks for spotting this, I was planning to do this but forgot about it.

m0dB:
> Is it the intended behavior, that loop end is considered as marker?

It is "intended" in the sense that I went for the most straightforward implementation, which was "until any marker". I think it works well, though I do find it annoying when you have a normal (not hot cue) but deactivated loop.

But maybe "until next hot cue" would be better? Or maybe this should be an option?

JoergAtGithub:
> But maybe "until next hot cue" would be better? Or maybe this should be an option?

My personal opinion is that it should count to the following markers:

- Last beat before track end
- Hotcues
- Start of saved loops
- Intro/Outro markers

I don't see a need for a further option.

m0dB:
Yes, I agree that makes sense. Unfortunately it was not immediately obvious to me how to access this information, so any help with that would be welcome.

JoergAtGithub:
I just ran into a debug assert testing this PR (in this moment I just clicked on tracks in the library, while a deck was playing):
DEBUG ASSERT: "isValid()" in function double __cdecl mixxx::audio::FramePos::value(void) const at D:/mixxx/src/audio/frame.h:100" 
```

Qt6Core.dll!qAbort() Line 150
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\global\qglobal.cpp(150)
[Inline Frame] Qt6Core.dll!qt_message_fatal(QtMsgType) Line 2059
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\global\qlogging.cpp(2059)
Qt6Core.dll!QMessageLogger::fatal(const char * msg=0x00007ff7d04d0a48, ...) Line 940
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\global\qlogging.cpp(940)
mixxx.exe!mixxx::`anonymous namespace'::handleMessage(QtMsgType type=QtCriticalMsg, const QMessageLogContext & context={...}, const QString & input={...}) Line 358
	at D:\mixxx\src\util\logging.cpp(358)
Qt6Core.dll!qt_message_print(QtMsgType msgType=QtCriticalMsg, const QMessageLogContext & context={...}, const QString & message={...}) Line 2017
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\global\qlogging.cpp(2017)
Qt6Core.dll!qt_message(QtMsgType msgType=QtCriticalMsg, const QMessageLogContext & context={...}, const char * msg, char * ap=0x0000009f42efaf40) Line 375
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\global\qlogging.cpp(375)
Qt6Core.dll!QMessageLogger::critical(const char * msg=0x00007ff7d04d0428, ...) Line 768
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\global\qlogging.cpp(768)
[Inline Frame] mixxx.exe!mixxx_debug_assert(const char *) Line 9
	at D:\mixxx\src\util\assert.h(9)
[Inline Frame] mixxx.exe!mixxx_debug_assert_return_true(const char *) Line 18
	at D:\mixxx\src\util\assert.h(18)
[Inline Frame] mixxx.exe!mixxx::audio::FramePos::value() Line 100
	at D:\mixxx\src\audio\frame.h(100)
mixxx.exe!mixxx::audio::operator-(mixxx::audio::FramePos framePos1, mixxx::audio::FramePos framePos2) Line 172
	at D:\mixxx\src\audio\frame.h(172)
mixxx.exe!mixxx::Beats::iteratorFrom(mixxx::audio::FramePos position) Line 435
	at D:\mixxx\src\track\beats.cpp(435)
mixxx.exe!allshader::WaveformRenderMark::updateUntilMark(double playPosition, double nextMarkPosition) Line 465
	at D:\mixxx\src\waveform\renderers\allshader\waveformrendermark.cpp(465)
mixxx.exe!allshader::WaveformRenderMark::paintGL() Line 285
	at D:\mixxx\src\waveform\renderers\allshader\waveformrendermark.cpp(285)
mixxx.exe!allshader::WaveformWidget::paintGL() Line 42
	at D:\mixxx\src\waveform\widgets\allshader\waveformwidget.cpp(42)
mixxx.exe!allshader::WaveformWidget::render() Line 28
	at D:\mixxx\src\waveform\widgets\allshader\waveformwidget.cpp(28)
mixxx.exe!WaveformWidgetFactory::renderSelf() Line 743
	at D:\mixxx\src\waveform\waveformwidgetfactory.cpp(743)
mixxx.exe!WaveformWidgetFactory::render() Line 779
	at D:\mixxx\src\waveform\waveformwidgetfactory.cpp(779)
[Inline Frame] Qt6Core.dll!QtPrivate::QSlotObjectBase::call(QObject *) Line 374
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qobjectdefs_impl.h(374)
Qt6Core.dll!QMetaCallEvent::placeMetaCall(QObject * object=0x00000234a1449970) Line 657
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qobject.cpp(657)
Qt6Core.dll!QObject::event(QEvent * e=0x00000234a9ab3540) Line 1438
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qobject.cpp(1438)
Qt6Widgets.dll!QApplicationPrivate::notify_helper(QObject * receiver=0x00000234a1449970, QEvent * e=0x00000234a9ab3540) Line 3292
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\widgets\kernel\qapplication.cpp(3292)
Qt6Widgets.dll!QApplication::notify(QObject * receiver=0x00000234a1449970, QEvent * e=0x00000234a9ab3540) Line 3244
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\widgets\kernel\qapplication.cpp(3244)
Qt6Core.dll!QCoreApplication::notifyInternal2(QObject * receiver=0x00000234a1449970, QEvent * event=0x00000234a9ab3540) Line 1118
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qcoreapplication.cpp(1118)
[Inline Frame] Qt6Core.dll!QCoreApplication::sendEvent(QObject *) Line 1536
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qcoreapplication.cpp(1536)
Qt6Core.dll!QCoreApplicationPrivate::sendPostedEvents(QObject * receiver=0x0000000000000000, int event_type, QThreadData * data=0x0000023482477c00) Line 1898
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qcoreapplication.cpp(1898)
Qt6Gui.dll!QWindowsGuiEventDispatcher::sendPostedEvents() Line 44
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\gui\platform\windows\qwindowsguieventdispatcher.cpp(44)
Qt6Core.dll!QEventDispatcherWin32::processEvents(QFlags<enum QEventLoop::ProcessEventsFlag> flags={...}) Line 464
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qeventdispatcher_win.cpp(464)
Qt6Gui.dll!QWindowsGuiEventDispatcher::processEvents(QFlags<enum QEventLoop::ProcessEventsFlag> flags) Line 37
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\gui\platform\windows\qwindowsguieventdispatcher.cpp(37)
[Inline Frame] Qt6Core.dll!QEventLoop::processEvents(QFlags<enum QEventLoop::ProcessEventsFlag>) Line 100
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qeventloop.cpp(100)
Qt6Core.dll!QEventLoop::exec(QFlags<enum QEventLoop::ProcessEventsFlag> flags={...}) Line 182
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qeventloop.cpp(182)
Qt6Core.dll!QCoreApplication::exec() Line 1439
	at C:\mixxx-vcpkg\buildtrees\qtbase\src\here-src-6-7373e57753.clean\src\corelib\kernel\qcoreapplication.cpp(1439)
mixxx.exe!`anonymous namespace'::runMixxx(MixxxApplication * pApp=0x0000009f42eff4b8, const CmdlineArgs & args) Line 108
	at D:\mixxx\src\main.cpp(108)
mixxx.exe!main(int argc=1, char * * argv=0x0000023482496d60) Line 243
	at D:\mixxx\src\main.cpp(243)
[Inline Frame] mixxx.exe!invoke_main() Line 78
	at D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl(78)
mixxx.exe!__scrt_common_main_seh() Line 288
	at D:\a\_work\1\s\src\vctools\crt\vcstartup\src\startup\exe_common.inl(288)
kernel32.dll!BaseThreadInitThunk()
ntdll.dll!RtlUserThreadStart()
```


grafik

JoergAtGithub (approved):

m0dB:
If this happens again, could you inspect the value of nextMarkPosition?

m0dB:
I added an extra check, hopefully this avoids the assert.

JoergAtGithub:
I could reproduce the assert with a debug build of [6caa23e](https://github.com/mixxxdj/mixxx/pull/12994/commits/6caa23e56ff6b9380b5a4db2dcc1b942fd8fb8b2) (without the extra check, you introduced later).
It seems to happen when loading tracks from library by double click - I clicked a lot until it happend.

grafik
grafik
grafik

m0dB:
This time it happened with playPosition though, which has a perfectly valid value...
 
IIRC we had other bugs with this assert failing randomly...

m0dB:
But I see that trackBeats->m_markers.size == 0 , that is suspicious, or in any case enough reason not to call updateUntilMark.

m0dB:
I believe I found the problem, I'll create a separate PR with a fix.

m0dB:
Fixed in https://github.com/mixxxdj/mixxx/pull/13150

JoergAtGithub:
> Fixed in #13150

Than the assert is unrelated to this PR.

@daschuer Do you confirm, that your change request above is properly addressed?

daschuer:
The Preferences can receive some love IMHO: 
grafik
"Until next marker" is probably a translator trap. The alignment feels cluttered. 

How about this idea (just an idea):
|                            |                     | | | |
|------------------------|------------------|------------------------|------------------| -----------------|
|Play marker hints | [ ] Beats until next marker  |  | | |
|| [ ] Time until next marker  |  | | |
|| Size  |  [ 24 px]  | Placement | [bottom] | 

If both is selected we may put them always in two rows.

daschuer (changes requested):

src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    VertexData texVertices;
+
+    posVertices.reserve(n * 6); // two triangles per character
+    texVertices.reserve(n * 6);
+
+    const QByteArray byteArray = s.toUtf8();
+
+    for (qsizetype i = 0; i < byteArray.size(); i++) {
+        // add vertices for the rectangles in the texture corresponding with each char in QString s
+        int index = charToIndex(byteArray[i]);
```
Swiftb0y, line 157: why not index directly? The cast to UTF-8 along with indexing only works if you assume the string only contains ascii anyways (which is true yeah, but this is not a very safe way of expressing that).

```suggestion
    for (QChar c : s) {
        int index = charToIndex(c);
```

src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -0,0 +1,196 @@
...
+    QFont font;
+    font.setFamily("Open Sans");
+    font.setPixelSize(fontPixelSize);
+
+    QFontMetricsF metrics{font};
+
+    qreal totalTextWidth = 0;
+    qreal maxTextHeight = 0;
+    for (int i = 0; i < NUM_CHARS; i++) {
+        assert(charToIndex(indexToChar(i)) == i);
```
Swiftb0y, line 67: wdyt about making both of these constexpr and then assert it for the entire valid set of characters in a static_assert instead?

src/waveform/renderers/allshader/waveformrendermark.cpp
```diff
@@ -43,21 +45,51 @@ class TextureGraphics : public WaveformMark::Graphics {
...
+QString timeToString(double t) {
+    int hundreds = std::lround(t * 100.0);
+    int seconds = hundreds / 100;
+    hundreds -= seconds * 100;
+    int minutes = seconds / 60;
+    seconds -= minutes * 60;
+
+    return QString::number(minutes) + (seconds < 10 ? ":0" : ":") +
+            QString::number(seconds) + (hundreds < 10 ? ".0" : ".") +
+            QString::number(hundreds);
```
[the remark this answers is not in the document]
daschuer: I had also a look into it. Conclusion: It does not work for our special case.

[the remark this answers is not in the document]
Swiftb0y: I don't expect you to write down you entire research, but can you summarize to my why it wouldn't work?

src/waveform/renderers/waveformmark.cpp
```diff
@@ -67,6 +67,17 @@ float overlappingMarkerIncrement(const float labelRectHeight, const float breadt
     return std::max(minIncrement, fullIncrement - std::max(0.f, threshold - breadth));
 }
 
+bool isShowUntilNextPositionControl(const QString& positionControl) {
+    // To identify which markers are included in the beat/time until next marker
+    // display, in addition to the hotcues
+    static const QStringList list{{"cue_point",
```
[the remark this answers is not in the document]
Swiftb0y: > Unfortunately this is still allocating, see:

I mean yeah, of course the suboptimal but it wouldn't only do it on initialization, which should only happen once (per thread) due to `list` being `const static`. Of course I would prefer a `std::array` along with a `std::any_of` as well, just thought the simplicity of `QStringList::contains` might be nice.

src/waveform/renderers/waveformmark.cpp
```diff
@@ -67,6 +67,17 @@ float overlappingMarkerIncrement(const float labelRectHeight, const float breadt
...
+            "intro_start_position",
+            "intro_end_position",
+            "outro_start_position",
+            "outro_end_position"}};
+    return list.contains(positionControl);
```
daschuer, line 78: Something like that (untested) 
```suggestion
    constexpr std::array<QLatin1String, 5>list{{"cue_point",
            "intro_start_position",
            "intro_end_position",
            "outro_start_position",
            "outro_end_position"}};
    auto it = std::find(list.cbegin(), list.cend() ,positionControl);
    return it != arr.cend();
```

Swiftb0y: using namespace Qt::Literals::StringLiterals;
constexpr std::array list = {"cue_point"_L1,
            "intro_start_position"_L1,
            "intro_end_position"_L1,
            "outro_start_position"_L1,
            "outro_end_position"_L1};
    return std::any_of(list.cbegin(), list.cend(), [](auto& view){
        view == positionControl;
}};

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: Can I resolve this, @daschuer  ?

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: Can I resolve this, @daschuer ?

m0dB:
@daschuer wrote:
> The Preferences can receive some love IMHO

I like your proposal. Do you think the two radio buttons could be on one row to save some vertical space?

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -111,6 +111,9 @@ WaveformWidgetFactory::WaveformWidgetFactory()
           m_defaultZoom(WaveformWidgetRenderer::s_waveformDefaultZoom),
           m_zoomSync(true),
           m_overviewNormalized(false),
+          m_untilNextMarkerShow(UntilNextMarkerShow::None),
+          m_untilNextMarkerAlign(Qt::AlignVCenter),
+          m_untilNextMarkerSize(24),
```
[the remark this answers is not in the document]
m0dB: yeah, text, in pixels. I'll rename the member for clarity.

src/waveform/renderers/waveformmark.cpp
m0dB: nice!

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: not sure I understand what you mean

m0dB:
@daschuer I have made changes to the preferences, but I am not sure how you envisioned the row with the alignment combobox and the font size spinbox. Could you have a look and tweak it to your liking?

src/waveform/waveformwidgetfactory.cpp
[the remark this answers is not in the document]
m0dB: done

m0dB:
@daschuer @Swiftb0y I have addressed all reviews. Please re-review and resolve the conversations and let me know if there is anything else you would like to see changed. I think this is good to go now.

daschuer:
The preferences looks good now, thank you. 

There is an issue with the minimum text size though. 
grafik
I think it is reasonable to allow values slightly below the font sizes using in the decks. 

As a feature enhancement, we may avoid such cut of situation by using a fraction of the waveform high or the preferences value depending on what is smaller. We may also automatically pick the single row presentation.

daschuer:
In the library preferences we can select the font-size in pt. I think it is reasonable to us this unit here as well. to have both values comparable.

src/waveform/renderers/allshader/digitsrenderer.cpp
daschuer: something like this: 
```
constexpr bool checkCharToIndex() {
    for (int i = 0; i < NUM_CHARS; i++) {
        if (charToIndex(indexToChar(i)) != i) {
        	return false;
        }
    }
    return true;
}
static_assert(checkCharToIndex());
```

charToIndex() and indexToChar() needs to become constexpr as well see my inline comments.

src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+// space around chars for blurred dark outline
+constexpr qreal SPACE = 3;
+// extra kerning when rendering chars
+constexpr qreal KERNING = 1;
+
+char indexToChar(int index) {
+    const char* str = "0123456789:.";
+    return str[index];
+}
```
daschuer, line 30:
```suggestion
constexpr char indexToChar(int index) {
	constexpr char str[] = "0123456789:.";
    return str[index];
}
```

src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+    if (ch == ':') {
+        return 10;
+    }
+    if (ch == '.') {
+        return 11;
+    }
+    assert(false);
+    return 11; // fallback to dot
+}
```
daschuer, line 44: digitValue() is unfortunatly not a constexpr, DEBUG_ASSERT will not crash release builds. 

```suggestion
constexpr int charToIndex(QChar ch) {
    int value = ch.toLatin1() - '0';
    if (value >= 0 && value <= 9) {
        return value;
    }
    if (ch == ':') {
        return 10;
    }
    if (ch == '.') {
        return 11;
    }
    DEBUG_ASSERT(false);
    return 11; // fallback to dot
}
```

m0dB:
@daschuer the size is now in points, and now keep the text height below a 3rd of the viewer height (using single line instead of multiline, and autoscaling the font)

JoergAtGithub:
Could you please fix pre-commit

m0dB:
> Could you please fix pre-commit

done. it's unfortunate that committing suggestions from reviews doesn't go through pre-commit.

daschuer (approved):
Thank you for this nice feature. And thank you for the extra efforts polishing the code.

src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -0,0 +1,226 @@
...
+    do {
+        font.setPointSizeF(fontPointSize);
+
+        metrics = QFontMetricsF{font};
+
+        totalTextWidth = 0;
+        maxTextHeight = 0;
+
+        for (int i = 0; i < NUM_CHARS; i++) {
+            assert(charToIndex(indexToChar(i)) == i);
```
daschuer, line 90: We avoid plain asserts, because they may crash in release builds. 

You may put this into the anonymous namespace instead: 
```
constexpr bool checkCharToIndex() {
    for (int i = 0; i < NUM_CHARS; i++) {
        if (charToIndex(indexToChar(i)) != i) {
        	return false;
        }
    }
    return true;
}
static_assert(checkCharToIndex());
```

daschuer (changes requested):
Oh, it looks like one comment was lost in the GitHub folding. I have added it again.

daschuer:
As of my knowledge, it is not possible to apply our checks on GitHub suggestions before committing.

m0dB:
> As of my knowledge, it is not possible to apply our checks on GitHub suggestions before committing.

It's no big deal, just something to remember.

ywwg:
> As of my knowledge, it is not possible to apply our checks on GitHub suggestions before committing.

not before committing, no, but as part of the CI pipeline.  This doesn't have to replace the pre-commit hooks, but can be a redundant second check.  If a github suggestion would break the linter, we'd catch it in the CI stage before merging.

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: done

m0dB:
rebased on main (had to resolve some conflicts with @acolombier 's slipmode code) and added @daschuer static_assert suggestion.

m0dB:
@daschuer wrote
> Thank you for this nice feature. And thank you for the extra efforts polishing the code.

You are welcome. And the suggestions, both for code quality and for feature improvements, were much appreciated.

daschuer (approved):
OK, than lets integrate it!

PR #11300, opened 2023-02-23
feat(Controller): add support for settings

acolombier:
This PR addresses the feature request #8135

A few capabilities have been added from the original plan:
- Ability to use `group`  with a `label` attribute
- Ability to use `row` which will organize settings horizontally if the settings group box is wider than 960px, with support for horizontal (default) and and vertical label/widget disposition

Outstanding work:

- [X] Add more setting type (real numbers, boolean, enum)
- [X] Add state persistence
- [X] Add HID support
- ~Add MIDI support~ _This will require refactoring the [load function](https://github.com/mixxxdj/mixxx/blob/1e1a2b7c72b5cfbe474f038e3bf1c21e666ec46f/src/controllers/midi/legacymidicontrollermappingfilehandler.cpp#L11) and IMHO it is wiser to do so in a dedicated PR_
- [X] Add test for setting spec
- [x] Add test for setting edition (from Pref dialog and cfg)
- [x] Add test for the parser

An update of the Wiki is currently in progress.

### Simple

image

<details>
  <summary>XML Config</summary>

```xml
<?xml version='1.0' encoding='utf-8'?>
<MixxxControllerPreset mixxxVersion="2.4.0" schemaVersion="1">
    <info>
        <name>Dummy Device Test Mapping 1</name>
        <author>acolombier</author>
        <description>HID Mapping virtual test device</description>
        <manual>test_dumy_hid</manual>
        <devices>
            <product protocol="hid" vendor_id="0x1234" product_id="0x5678" />
        </devices>
    </info>
    <settings>
        <option
            variable="myFactor"
            type="integer"
            min="1"
            max="1000"
            step="10"
            default="100"
            label="My dummy factor"
            label_de="Mein Dummy-Faktor"
            label_fr="Mon facteur factice" />
        <option
            variable="foo"
            type="string"
            default="Bar"
            label="My unsupport setting" />
        <column></column>
        <option
            variable="myOtherFactor"
            type="real"
            precision="2"
            min="0"
            max="1"
            step="0.05"
            default="0.15"
            label="My other factor">
            <description>
                This factor is useful for one thing that you definitely need.
            </description>
            <description lang="de">
                Dieser Faktor ist nützlich für eine Sache, die Sie unbedingt brauchen.
            </description>
        </option>
        <option
            variable="myToggle"
            type="boolean"
            default="false"
            label="My beta feature">
            <description lang="de">
                It include many things, you should try it.
            </description>
        </option>
        <option
            variable="myOptionList"
            type="enum"
[...]

PR #12995, opened 2024-03-23
feat: add setting definition for Traktor S4 MK3

acolombier:
This PR adds the setting definition for Traktor S4 MK3.

_Manual PR to come soon_

Depends on #11300 

image

### DIFF
--- a/res/controllers/Traktor Kontrol S4 MK3.hid.xml
+++ b/res/controllers/Traktor Kontrol S4 MK3.hid.xml
@@ -9,6 +9,620 @@
             <product protocol="hid" vendor_id="0x17cc" product_id="0x1720" usage_page="0xff01" usage="0x1" interface_number="0x4" />
         </devices>
     </info>
+    <settings>
+        <group label="Deck Lighting">
+            <row orientation="vertical">
+                <option
+                    variable="deckA"
+                    type="enum"
+                    label="Deck A">
+                    <value label="Aqua">aqua</value>
+                    <value label="Azalea">azalea</value>
+                    <value label="Blue">blue</value>
+                    <value label="Celeste">celeste</value>
+                    <value label="Fuscia">fuscia</value>
+                    <value label="Green">green</value>
+                    <value label="Lime">lime</value>
+                    <value label="Orange">orange</value>
+                    <value label="Purple">purple</value>
+                    <value label="Red" default="true">red</value>
+                    <value label="Salmon">salmon</value>
+                    <value label="Sky">sky</value>
+                    <value label="White">white</value>
+                    <value label="Yellow">yellow</value>
+                </option>
+                <option
+                    variable="deckB"
+                    type="enum"
+                    label="Deck B">
+                    <value label="Aqua">aqua</value>
+                    <value label="Azalea">azalea</value>
+                    <value label="Blue" default="true">blue</value>
+                    <value label="Celeste">celeste</value>
+                    <value label="Fuscia">fuscia</value>
+                    <value label="Green">green</value>
+                    <value label="Lime">lime</value>
+                    <value label="Orange">orange</value>
+                    <value label="Purple">purple</value>
+                    <value label="Red">red</value>
+                    <value label="Salmon">salmon</value>
+                    <value label="Sky">sky</value>
+                    <value label="White">white</value>
+                    <value label="Yellow">yellow</value>
+                </option>
+                <option
+                    variable="deckC"
+                    type="enum"
+                    label="Deck C">
+                    <value label="Aqua">aqua</value>
+                    <value label="Azalea">azalea</value>
+                    <value label="Blue" default="true">blue</value>
+                    <value label="Celeste">celeste</value>
+                    <value label="Fuscia">fuscia</value>
+                    <value label="Green">green</value>
+                    <value label="Lime">lime</value>
+                    <value label="Orange">orange</value>
+                    <value label="Purple">purple</value>
+                    <value label="Red">red</value>
+                    <value label="Salmon">salmon</value>
+                    <value label="Sky">sky</value>
+                    <value label="White">white</value>
+                    <value label="Yellow" default="true">yellow</value>
+                </option>
+                <option
+                    variable="deckD"
+                    type="enum"
+                    label="Deck D">
+                    <value label="Aqua">aqua</value>
+                    <value label="Azalea">azalea</value>
+                    <value label="Blue">blue</value>
+                    <value label="Celeste">celeste</value>
+                    <value label="Fuscia">fuscia</value>
+                    <value label="Green">green</value>
+                    <value label="Lime">lime</value>
+                    <value label="Orange">orange</value>
+                    <value label="Purple" default="true">purple</value>
+                    <value label="Red">red</value>
+                    <value label="Salmon">salmon</value>
+                    <value label="Sky">sky</value>
+                    <value label="White">white</value>
+                    <value label="Yellow">yellow</value>
+                </option>
+            </row>
+            <option
+                variable="inactiveLightsAlwaysBacklit"
+                type="boolean"
+                default="true"
+                label="Keep one-color LED dimmed instead of off when inactive">
+                <description>
+                    This will consistently ensure that every button is backlit, which can be helpful in low-light environments, but it might alter the primary color scheme of the deck.
+                </description>
+            </option>
+            <option
+                variable="gridButtonBlinkOverBeat"
+                type="boolean"
+                default="false"
+                label="Make the &lt;tt style='background-color: #111111; color: #d9d9d9;padding: 2px;'&gt;GRID&lt;/tt&gt; button blink over the beat">
+                <description>
+                    This will make the &lt;tt style="background-color: #111111; color: #d9d9d9;padding: 2px;"&gt;GRID&lt;/tt&gt; button blink when the track is going over a detected beat. This can help adjusting the beat grid or BPM.
+                </description>
+            </option>
+            <option
+                variable="deckSelectAlwaysBacklit"
+                type="boolean"
+                default="true"
+                label="Keep both deck select buttons backlit">
+                <description>
+                    Keep both deck select buttons backlit and do not fully turn off the inactive deck button. If enabled, will keep the unselected deck dimmed, otherwise it will fully turn it off.
+                </description>
+            </option>
+            <option
+                variable="wheelLedBlinkOnTrackEnd"
+                type="boolean"
+                default="true"
+                label="Wheel LED blinking when reaching the end of the track">
+                <description>
+                    The wheel LED will blink when reaching the end of the track, as per the settings defined under "Waveforms" > "End of Track Warning".
+                </description>
+            </option>
+            <option
+                variable="tempoFaderSoftTakeoverColorLow"
+                type="enum"
+                label="Tempo fader low soft takeover color">
+                <value label="Aqua">aqua</value>
+                <value label="Azalea">azalea</value>
+                <value label="Blue" default="true">blue</value>
+                <value label="Celeste">celeste</value>
+                <value label="Fuscia">fuscia</value>
+                <value label="Green">green</value>
+                <value label="Lime">lime</value>
+                <value label="Orange">orange</value>
+                <value label="Purple">purple</value>
+                <value label="Red">red</value>
+                <value label="Salmon">salmon</value>
+                <value label="Sky">sky</value>
+                <value label="White" default="true">white</value>
+                <value label="Yellow">yellow</value>
+                <description>
+                    This color defines how the LED next to the tempo fader will be lit when the actual BPM is higher than the fader value.
+                </description>
+            </option>
+            <option
+                variable="tempoFaderSoftTakeoverColorHigh"
+                type="enum"
+                label="Tempo fader high soft takeover color">
+                <value label="Aqua">aqua</value>
+                <value label="Azalea">azalea</value>
+                <value label="Blue">blue</value>
+                <value label="Celeste">celeste</value>
+                <value label="Fuscia">fuscia</value>
+                <value label="Green" default="true">green</value>
+                <value label="Lime">lime</value>
+                <value label="Orange">orange</value>
+                <value label="Purple">purple</value>
+                <value label="Red">red</value>
+                <value label="Salmon">salmon</value>
+                <value label="Sky">sky</value>
+                <value label="White">white</value>
+                <value label="Yellow">yellow</value>
+                <description>
+                    This color defines how the LED next to the tempo fader will be lit when the actual BPM is lower than the fader value.
+                </description>
+            </option>
+        </group>
+
+        <group label="Alternative Mapping">
+            <option
+                variable="useKeylockOnMaster"
+                type="boolean"
+                default="false"
+                label="Keylock on &lt;tt style='background-color: #111111; color: #d9d9d9;padding: 2px;'&gt;SHIFT&lt;/tt&gt;+&lt;tt style='background-color: #111111; color: #d9d9d9;padding: 2px;'&gt;MASTER&lt;/tt&gt; instead of &lt;tt style='background-color: #111111; color: #d9d9d9;padding: 2px;'&gt;SHIFT&lt;/tt&gt;+&lt;tt style='background-color: #111111; color: #d9d9d9;padding: 2px;'&gt;SYNC&lt;/tt&gt;">
+                <description>
+                    This setting toggles the keylock on push instead of release, as there is a long-press action on &lt;tt style="background-color: #111111; color: #d9d9d9;padding: 2px;"&gt;SHIFT&lt;/tt&gt;+&lt;tt style="background-color: #111111; color: #d9d9d9;padding: 2px;"&gt;SYNC&lt;/tt&gt; (copy key).
+                </description>
+            </option>
+            <option
+                variable="useBeatloopRollInsteadOfSampler"
+                type="boolean"
+                default="false"
+                label="Use the Sampler tab as a Beatloop Roll tab instead">
+                <description>
+                    This option is useful if you don't use samplers and would like to access a set of predefined beatloop rolls. Beat size can be customized individually.
+                </description>
+            </option>
+            <option
+                variable="mixerControlsMixAuxOnShift"
+                type="boolean"
+                default="false"
+                label="Shift to mix microphones or auxiliary lines">
+                <description>
+                    When shifting decks, the mixer controls microphones or auxiliary lines. If both a microphone and an auxiliary are configured on the same channel, the mixer will prioritize controlling the auxiliary.
+                </description>
+            </option>
+        </group>
+
+        <group label="Beatloop Roll Size (Only if Enabled Instead of the Sampler Tab)">
+            <row orientation="vertical">
+                <option
+                    variable="beatLoopRollsSize1"
+                    type="enum"
+                    label="First Pad">
+                    <value label="1/32">0.03125</value>
+                    <value label="1/16" default="true">0.0625</value>
+                    <value label="1/8">0.125</value>
+                    <value label="1/4">0.25</value>
+                    <value label="1/2">0.5</value>
+                    <value label="1">1</value>
+                    <value label="2">2</value>
+                    <value label="4">4</value>
+                    <value label="8">8</value>
+                    <value label="Halve beatloop size">half</value>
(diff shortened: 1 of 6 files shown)


acolombier:
Added a second commit to provide backward compatibility with 2.4 as an option. Happy to remove or rebase to target `2.4` instead

res/controllers/Traktor-Kontrol-S4-MK3.js
```diff
@@ -38,97 +38,124 @@ const KeyboardColors = [
...
-    script.LIBRARY_COLUMNS.KEY,
-    script.LIBRARY_COLUMNS.DATETIME_ADDED,
-];
+    getSetting("librarySortableColumns1Value", script.LIBRARY_COLUMNS.ARTIST.toString()),
+    getSetting("librarySortableColumns2Value", script.LIBRARY_COLUMNS.TITLE.toString()),
+    getSetting("librarySortableColumns3Value", script.LIBRARY_COLUMNS.BPM.toString()),
+    getSetting("librarySortableColumns4Value", script.LIBRARY_COLUMNS.KEY.toString()),
+    getSetting("librarySortableColumns5Value", script.LIBRARY_COLUMNS.DATETIME_ADDED.toString()),
+    getSetting("librarySortableColumns6Value", "0"),
+].map(c => parseInt(c)).filter(c => c); // Filter '0' column, equivalent to '---' value in the UI or disabled
```
acolombier, line 80: You would have expected `.map(parseInt)` to work, write?


Probably one I should add [here](https://github.com/denysdovhan/wtfjs?tab=readme-ov-file#parseint-is-a-bad-guy)

ywwg:
"chore" is for low-level repository tasks.  This is a "feat"

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                    <value label="Purple" default="true">purple</value>
+                    <value label="Red">red</value>
+                    <value label="Salmon">salmon</value>
+                    <value label="Sky">sky</value>
+                    <value label="White">white</value>
+                    <value label="Yellow">yellow</value>
+                </option>
+            </row>
+            <option
+                variable="keepLEDWithOneColorDimedWhenInactive"
```
ywwg, line 93: typo  ("dimed") but similar to below, let's say:
```suggestion
                variable="inactiveLightsAlwaysBacklit"
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                type="boolean"
+                default="true"
+                label="Keep one-color LED dimmed instead of off when inactive">
+                <description>
+                    This will always ensure every button has some lighting which may help in dark environment, but may put off the lead color of the deck.
```
ywwg, line 98: not sure what you mean by "put off the lead color"? That doesn't quite scan for my own grammar.

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                </description>
+            </option>
+            <option
+                variable="gridButtonBlinkOverBeat"
+                type="boolean"
+                default="false"
+                label="Make the GRID button blink over beat">
+                <description>
+                    This will Make the GRID button blink when the track is going over a detected beat. This can help adjusting the beat grid or BPM.
```
ywwg, line 107:
```suggestion
                    This will make the GRID button blink when the track is going over a detected beat. This can help adjusting the beat grid or BPM.
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                </description>
+            </option>
+            <option
+                variable="keepDeckSelectDimmed"
```
ywwg, line 111: using the term "dimmed" is a little confusing because it could mean that the brightest value is dimmer.  Instead let's say:
```suggestion
                variable="deckSelectAlwaysBacklit"
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                type="boolean"
+                default="true"
+                label="Keep both deck select buttons backlit">
+                <description>
+                    Keep both deck select buttons backlit and do not fully turn off the inactive deck button. If enabled, will keep the unseclected deck dimmed, otherwise it will fully turn it off.
```
ywwg, line 116: please run a spell check on this PR, there are a lot of words :)

```suggestion
                    Keep both deck select buttons backlit and do not fully turn off the inactive deck button. If enabled, will keep the unselected deck dimmed, otherwise it will fully turn it off.
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                variable="useBeatloopRollInsteadOfSampler"
+                type="boolean"
+                default="false"
+                label="Use the Sampler tab as a beatloop roll tab instead">
+                <description>
+                    This is useful if you don't use samplers and would like to have access to a set of predefined beatloop rolls. Beat size can be customized individually.
+                </description>
+            </option>
+            <option
+                variable="mixerControlsMixAuxOnShift"
```
ywwg, line 189: do we need this option? Why not make it always active?

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                    <value label="Halve beatloop size">half</value>
+                    <value label="Double beatloop size" default="true">double</value>
+                    <description>The bottom right pad button.</description>
+                </option>
+            </row>
+        </group>
+        <group label="Library">
+            <row orientation="vertical">
+                <option
+                    variable="librarySortableColumns1Value"
```
ywwg, line 344: the last item was 0-indexed, this is 1-indexed.  Let's be consistent

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                    <value label="Rating">19</value>
+                    <value label="Key">20</value>
+                    <description>Fifth column when iterating over sortable library columns using VIEW+Library Encoder press</description>
+                </option>
+                <option
+                    variable="librarySortableColumns6Value"
+                    type="enum"
+                    label="Sixth column">
+                    <value label="---" default="true">0</value>
+                    <value label="Artist">1</value>
```
ywwg, line 482: will we have to keep these values in sync with code? Can we add a test that breaks if those change?

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                    <value label="Key">20</value>
+                    <description>Sixth column when iterating over sortable library columns using VIEW+Library Encoder press</description>
+                </option>
+            </row>
+        </group>
+        <group label="Jog wheel">
+            <option
+                variable="loopWheelMoveFactor"
+                type="integer"
+                min="20"
```
ywwg, line 510: make sure the min and max are bigger than is reasonable, just in case someone wants to do something wild :)

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+            </option>
+        </group>
+        <group label="Haptic drive feature (BETA)">
+            <option
+                variable="useMotors"
+                type="boolean"
+                default="false"
+                label="Use beta Haptic Drive features (BETA)">
+                <description>
+                    This will defines whether or not Mixxx is allowed to interact with the motors. &lt;br /&gt;&lt;b&gt;WARNING&lt;/b&gt;: while interacting with motors should be harmless for your device, it isn't possible to give a full guaranty that this will not accelerate wearing off.
```
ywwg, line 574:
```suggestion
                    This will defines whether or not Mixxx is allowed to interact with the motors. &lt;br /&gt;&lt;b&gt;WARNING&lt;/b&gt;: while interacting with motors should be harmless for your device, it could cause excess wear and tear.
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+            <option
+                variable="turnTableSpeedSample"
+                type="integer"
+                min="2"
+                max="80"
+                step="2"
+                default="40"
+                label="Turntable speed sample">
+                <description>
+                    Define how many wheel moves are sampled to compute the speed when using the motor. This is helpful to mitigate delay that occurs in communication as well as Mixxx limitation to 20ms latency. The more you have, the more the speed is accurate, but the less responsive it gets in Mixxx
```
ywwg, line 586:
```suggestion
                    Define how many wheel ticks are sampled to compute the speed when using the motor. This is helpful to mitigate the delay that occurs in communication as well as Mixxx's built-in 20ms latency. Higher values will give smoother speeds at the expense of responsiveness.
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                    variable="tightnessFactor"
+                    type="real"
+                    precision="2"
+                    min="0.05"
+                    max="0.95"
+                    step="0.05"
+                    default="0.5"
+                    label="Jogwheel tension">
+                    <description>
+                        Define how much the jogwheel will resist. It is a similar setting that the Grid+Wheel in Tracktor. 0 is very tight, 1 is very loose.
```
ywwg, line 601:
```suggestion
                        Jogwheel resistance. It is a similar setting that the Grid+Wheel in Tracktor. 0 is very tight, 1 is very loose.
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                        &lt;br /&gt;&lt;b&gt;WARNING&lt;/b&gt;: A loose value can sometimes leads to ghost moves!
```
ywwg, line 602:
```suggestion
                        &lt;br /&gt;&lt;b&gt;WARNING&lt;/b&gt;: A loose value can sometimes lead to unintended motion!
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                    Define how many wheel moves are sampled to compute the speed. The more you have, the more the speed is accurate, but the less responsive it gets in Mixxx.
+                </description>
+            </option>
+            <option
+                variable="baseRevolutionsPerMinute"
+                type="enum"
+                label="Turntable speed">
+                <value label="33.3 RPM" default="true">33.33</value>
+                <value label="45 RPM">45</value>
+                <description>Define how fast the jogwheel would rotate with the default play rate. This will impact scratch sensitivity, motor speed and ring LED pace.</description>
```
ywwg, line 564:
```suggestion
                <description>Define how fast the jogwheel rotates at the default play rate. This will impact scratch sensitivity, motor speed and ring LED pace.</description>
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+            <option
+                variable="wheelSpeedSample"
+                type="integer"
+                min="2"
+                max="80"
+                step="2"
+                default="5"
+                label="Speed sample">
+                <description>
+                    Define how many wheel moves are sampled to compute the speed. The more you have, the more the speed is accurate, but the less responsive it gets in Mixxx.
```
ywwg, line 555: same edits as below

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                <option
+                    variable="loopEncoderShiftMoveFactor"
+                    type="integer"
+                    min="500"
+                    max="12500"
+                    step="20"
+                    default="2500"
+                    label="Loop encoder sensitivity in loop mode (shift)">
+                    <description>
+                        Define the sensitivity of shift+the loop encoder when moving loop position in the Loop Mode.
```
ywwg, line 540:
```suggestion
                        Define the sensitivity of shift+loop encoder when moving loop position in the Loop Mode.
```

res/controllers/Traktor Kontrol S4 MK3.hid.xml
```diff
@@ -9,6 +9,615 @@
...
+                    <value label="Comment">12</value>
+                    <value label="Duration">13</value>
+                    <value label="Bitrate">14</value>
+                    <value label="BPM">15</value>
+                    <value label="ReplayGain">16</value>
+                    <value label="Datetime Added">17</value>
+                    <value label="Times Played">18</value>
+                    <value label="Rating">19</value>
+                    <value label="Key">20</value>
+                    <description>Second column when iterating over sortable library columns using VIEW+Library Encoder press</description>
```
ywwg, line 394: we don't capitalize any other buttons, should we?

res/controllers/Traktor-Kontrol-S4-MK3.js
```diff
@@ -38,97 +38,124 @@ const KeyboardColors = [
...
 
-const LoopWheelMoveFactor = 50;
-const LoopEncoderMoveFactor = 500;
-const LoopEncoderShiftmoveFactor = 2500;
+const LoopWheelMoveFactor = getSetting("loopWheelMoveFactor", 50);
```
ywwg, line 82: the defaults are already defined in the xml, why are they duplicated here?  That is a recipe for unexpected behavior

res/controllers/Traktor-Kontrol-S4-MK3.js
```diff
@@ -38,97 +38,124 @@ const KeyboardColors = [
...
 ];
 
+
+/**
+ * This helper function is used to ensure backward compatibility between 2.4 and later versions
+ * @param {string} key the setting key as per defined in the manifest
+ * @param {any} defaultVal the value to use if not controller setting is available
+ * @returns {any} the value to use for this setting
+ */
+var getSetting = (key, defaultVal) => {
```
ywwg, line 48: ahhh I see what is happening here.  Let's call this  something else so it's clearly a fallback when deck settings aren't available.  OR, we just commit this to main and drop this.
```suggestion
var getSettingForV24 = (key, defaultVal) => {
```

ywwg (changes requested):
Thank you for doing this! It's very fun to see these get used right away.  I mostly made some language edits and notes, as well as some naming nits.

res/controllers/Traktor-Kontrol-S4-MK3.js
acolombier: This is because of 2.4 compatibility (the second commit), I will revert that one and make this a 2.5 mapping only.

acolombier (commented):
> It's very fun to see these get used right away

That branch has been sitting on my local for almost a year, so it is a relief to finally seeing being useful to other :) 
Will do the suggested changed changes in a bit. How hard would it be to help me find all grammar/spelling mistakes? If too hard, I can try an see if ChatGPT can help correcting this in place :rofl:

acolombier:
I have removed 2.4 backward compatibility and tried and rework phrasing and spelling (with a limited success - sorry)

I have also experienced with custom style around button names, similar to how it is done on the manual. What are your thoughts? 

image

res/controllers/Traktor Kontrol S4 MK3.hid.xml
ywwg: oh is that a typo?
```suggestion
                variable="mixerControlsMicAuxOnShift"
```

ywwg: I mean, why not make this true by default and not be an option?

res/controllers/Traktor Kontrol S4 MK3.hid.xml
ywwg: It could be a gross gtest in c++ that loads the xml and compares it to the C++ code.  or maybe a better option is we could make this file somehow templated (qtxml?) and generated as part of the build process.  

The library changes not very often, but often enough. Without an automated way of detecting the change I'm afraid we will never know that we broke this config.

res/controllers/Traktor Kontrol S4 MK3.hid.xml
ywwg: yeah probably a bigger max.  I have seen some videos of djs cranking the CDJ wheel sensitivity way way up and getting some really cool effects.    Sure let's try 10x!

res/controllers/Traktor-Kontrol-S4-MK3.js
```diff
@@ -40,95 +40,102 @@ const KeyboardColors = [
 
 /*
  * USER CONFIGURABLE SETTINGS
- * Adjust these to your liking
+ * Change settings in the preferences (Mixxx >= 2.5)
```
ywwg, line 43: no need to specify version since by definition the user will be on a valid version

res/controllers/Traktor-Kontrol-S4-MK3.js
ywwg: yay  javascript lol sob

src/controllers/legacycontrollersettings.cpp
```diff
@@ -118,7 +120,21 @@ QWidget* LegacyControllerBooleanSetting::buildInputWidget(QWidget* pParent) {
         emit changed();
     });
 
-    return pCheckBox;
+    auto pLabelWidget = make_parented<QLabel>(pWidget);
```
ywwg, line 123: what is this change for?

ywwg (commented):
looking good! just a few more questions and then I think we're good to go

src/controllers/legacycontrollersettings.cpp
acolombier: This change was requried as the `QCheckBox` built-in label doesn't support rich text. Happy to move that to an other PR is we are happy about the way it looks with rich text (that is the emphasis on button )
image

res/controllers/Traktor Kontrol S4 MK3.hid.xml
acolombier: I went with Option 1 - let me know if that's good enough.

src/test/controllers/controller_s4mk3_settings_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    void TearDown() override {
+        mixxx::Time::setTestMode(false);
+    }
+
+    static QHash<QString, TrackModel::SortColumnId> COLUMN_MAPPING;
+};
+
+QHash<QString, TrackModel::SortColumnId> ControllerS4MK3SettingTest::COLUMN_MAPPING = {
+        {"Artist", TrackModel::SortColumnId::Artist},
```
ywwg, line 27: It is unfortunate that we don't have enum-to-string mapping in the actual code. https://github.com/mixxxdj/mixxx/blob/main/src/library/browse/browsetablemodel.cpp#L60

There is an issue with internationalization, we will need a solution for that too eventually.  But for now this is good enough, thank you!

ywwg: Maybe add a comment to this file that if other controllers want to add the same type of feature they can also use this test?  Maybe rename the test to columnid_regression_test.cpp or something like that?

ywwg (approved):
thank you for the regression test!  This looks good to me

src/test/controllers/controller_s4mk3_settings_test.cpp
acolombier: Done!

ywwg (approved):

ywwg:
As is tradition, can you prepare the manual PR? we have to insist on having both ready to enforce that documentation stays up to date <3

acolombier:
> As is tradition, can you prepare the manual PR? we have to insist on having both ready to enforce that documentation stays up to date <3

Already [done](https://github.com/mixxxdj/manual/pull/634) and waiting for your review :D

ywwg:
merging, and then we'll make sure to merge the manual PR in the right place.

PR #13002, opened 2024-03-25
SlipMode waveform visual for RGB GLSL

acolombier:
This is an attempt to address the missing slip mode visual feature as requested in #8136

Other renderer will be addressed in different PRs.

image

TODO:
- [x] Render marker on the slip position part of the wave form upside down
- [x] Render the Post/pre roll


JoergAtGithub:
Did you consider to draw the whole waveform twice, half height stacked on each other?
I would use the indicator colors of the spinnies (white and orange) also on the waveform.

m0dB:
Nice that you are addressing this!

I had a hard time understand what was happening looking at the screen capture, and in the end I realised why: the start of track triangles are not part of the slip mode (which beat positions is). Is that intentional?

How about another approach: draw the slipping waveform on top, using waveformrenderersimple, so it is a single color, which 50% transparency.

Some comments (note, without looking at the code, so I don't know your current approach, it might very well be along the lines):

Assuming the rendering with the normal play position and the slip position is done in the same waveform widget (and I think it should), there is no need to do any synchronisation.

In ```waveformwidgetrenderer.cpp```:
```
    double truePlayPos = m_visualPlayPosition->getAtNextVSync(vsyncThread);
```
you can call 
```
    m_pVisualPlayPos->getPlaySlipAtNextVSync(vSyncThread, &truePlayPos, &slipPlayPos);
```
and you will get both positions synchronised.

The widget has a stack of renderers, and you can easily add more renderers to this stack to render the slip mode, and tell these renderers to use the slipPlayPos instead of the truePlayPos. When not in slip mode, you just skip all those renderers. You can then decide whether you want to only render the waveform itself for the slip mode, or (part of the) additional layers as well, or whether you want them to overlay with transparency or draw them tiled. In other words, it would equally allow to do what I am seeing in the screen capture, what Joerg is proposing and what I am proposing.

Apart from the synchronisation, using a single waveform widget with additional renderers makes sense because play mode and slip mode share a lot of information that is part of the widget (the waveform!), but also there is overhead in using a multiple widgets, as each has it's own QOpenGLWindow, so I would avoid that. (I have been thinking of using a single QOpenGLWindow for multiple WGLWidgets when possible: When multiple WGLWidgets combined occupy a single rectangle)

Let me know if this explanation is clear enough.

acolombier:
> I had a hard time understand what was happening looking at the screen capture

That's my fault. I've tested with that dummy audio file I use to test waveform correctness (monotonous tone from 20Hhz to 20Khz).

> the start of track triangles are not part of the slip mode

Yep I hadn't had implemented the pre/postroll triangle just yet. I just gave it a stab now (still incomplete with part of track missing on the slip waveform) 

https://github.com/mixxxdj/mixxx/assets/7086688/301596d8-2f98-4068-9625-46372a5d4611

> How about another approach: draw the slipping waveform on top, using waveformrenderersimple, so it is a single color, which 50% transparency.

Just to check I understood correctly, do you mean having it on top of the play waveform with transparency? I can think of two problem with this:
- I might be hard to "recognise" the track with the section of track covering each other
- (currently missing for the screen record but in my plans) The cues which will be displayed on the slip waveform are likely to also be hard to distinguish. We could make them look differently too (color, transparency) but things like label may also make the visual bloated 

I guess it could be great to have renderer using different look for the slip visuals? Since it looks like renderer should implement this, I guess there is no technical limitation to make it look different in various waveform, so users may also gain more value out of using different renderers 

> you can call
> 
> ```
>     m_pVisualPlayPos->getPlaySlipAtNextVSync(vSyncThread, &truePlayPos, &slipPlayPos);
> ```

Cool, that's exactly what I'm using :) 

> The widget has a stack of renderers, and you can easily add more renderers to this stack to render the slip mode, and tell these renderers to use the slipPlayPos instead of the truePlayPos. When not in slip mode, you just skip all those renderers. You can then decide whether you want to only render the waveform itself for the slip mode, or (part of the) additional layers as well, or whether you want them to overlay with transparency or draw them tiled. In other words, it would equally allow to do what I am seeing in the screen capture, what Joerg is proposing and what I am proposing.

Sounds good. I will look at how easy to remove the currently hardcoded references to `playPos` and instead use an arbitrary `pos` for all the renderer

> Let me know if this explanation is clear enough.

Thanks, that was very clear.

acolombier:
I just rewrote everything using your feedback and I think it looks much better, simpler and easier! Let me know your thoughts (visual available on [Zulip](https://mixxx.zulipchat.com/#narrow/stream/247619-UI-.26-UX-design/topic/Playback.20cursor.20on.20waveform.20in.20slip.20mode/near/429730348))

m0dB:
Yes, this is the approach I had in mind! Nice!

acolombier:
> I would use the indicator colors of the spinnies (white and orange) also on the waveform.

I have set the same orange as the ghost spinny cursor for LateNight PaleMoon and red-ish for Classic, other themes default to a light grey, but happy to put some custom color if you have suggestion!

image

ronso0:
> Let me know your thoughts (visual available on [Zulip](https://mixxx.zulipchat.com/#narrow/stream/247619-UI-.26-UX-design/topic/Playback.20cursor.20on.20waveform.20in.20slip.20mode/near/429730348))

This looks really good!

src/shaders/slipmodeshader.cpp
```diff
@@ -0,0 +1,60 @@
...
+uniform highp vec2 borders;
+uniform highp vec2 dimension;
+
+in highp vec4 vposition;
+
+void main()
+{
+
+    float xBorder = abs(dimension.x*vposition.x);
+    float yBorder = dimension.y*vposition.y;
```
daschuer, line 31: Not sure why clang-format is not complaining:
```suggestion
    float xBorder = abs(dimension.x * vposition.x);
    float yBorder = dimension.y * vposition.y;
```

src/waveform/renderers/allshader/waveformrendererslipmode.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+#include "control/controlproxy.h"
+#include "waveform/renderers/waveformwidgetrenderer.h"
+#include "waveform/waveformwidgetfactory.h"
+#include "widget/wskincolor.h"
+
+namespace {
+
+constexpr int kBlinkingPeriodMillis = 1600;
+constexpr float positionArray[] = {-1.f, -1.f, 1.f, -1.f, -1.f, 1.f, 1.f, 1.f};
```
daschuer, line 14: This anonymous array is hard to understand. Can we improve on that, at least with a comment?

src/waveform/renderers/allshader/waveformrendererslipmode.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+} // anonymous namespace
+
+namespace allshader {
+
+WaveformRendererSlipMode::WaveformRendererSlipMode(
+        WaveformWidgetRenderer* waveformWidget)
+        : WaveformRenderer(waveformWidget),
+          m_pSlipMode(nullptr),
```
daschuer, line 23: redundant

src/waveform/renderers/allshader/waveformrendererslipmode.cpp
```diff
@@ -0,0 +1,107 @@
...
+          m_slipBorderTopOutlineSize(10.f),
+          m_slipBorderBottomOutlineSize(10.f) {
+}
+
+bool WaveformRendererSlipMode::init() {
+    m_timer.restart();
+
+    m_pSlipMode.reset(new ControlProxy(
+            m_waveformRenderer->getGroup(), "slip_enabled"));
```
daschuer, line 32:
```suggestion
    m_pSlipMode = std::make_unique<ControlProxy>(
            m_waveformRenderer->getGroup(), "slip_enabled");
```

src/waveform/renderers/allshader/waveformrendererslipmode.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    return true;
+}
+
+void WaveformRendererSlipMode::setup(const QDomNode& node, const SkinContext& context) {
+    m_color = QColor(224, 224, 224);
```
daschuer, line 38: Magic number, what does this gray?

src/waveform/renderers/allshader/waveformrendererslipmode.cpp
```diff
@@ -0,0 +1,107 @@
...
+    const QString slipModeOutlineColorName = context.selectString(node, "SlipBorderOutlineColor");
+    if (!slipModeOutlineColorName.isNull()) {
+        m_color = QColor(slipModeOutlineColorName);
+        m_color = WSkinColor::getCorrectColor(m_color);
```
daschuer, line 42: Writing the member twice is confusing here. How about: 
```suggestion
        m_color = WSkinColor::getCorrectColor(QColor(slipModeOutlineColorName));
```
Or introduce another temporary.

src/waveform/renderers/allshader/waveformrendererslipmode.cpp
daschuer: Can we move that to the else branch?

src/waveform/renderers/allshader/waveformrendererslipmode.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+void WaveformRendererSlipMode::paintGL() {
+    if (!m_pSlipMode->toBool() || !m_waveformRenderer->isSlipActive()) {
+        return;
+    }
+
+    const int elapsed = m_timer.elapsed().toIntegerMillis() % kBlinkingPeriodMillis;
+
+    const double blinkIntensity = (double)(2 * abs(elapsed - kBlinkingPeriodMillis / 2)) /
```
daschuer, line 68:
```suggestion
    const double blinkIntensity = static_cast<double>(2 * abs(elapsed - kBlinkingPeriodMillis / 2)) /
```

src/waveform/renderers/allshader/waveformrendererslipmode.cpp
```diff
@@ -0,0 +1,107 @@
...
+        m_shader.setUniformValue(borderLocation,
+                m_slipBorderTopOutlineSize,
+                m_slipBorderBottomOutlineSize);
+
+        m_shader.setAttributeArray(
+                positionLocation, GL_FLOAT, positionArray, 2);
+
+        m_shader.setUniformValue(gradientLocation,
+                (float)m_waveformRenderer->getLength() / 2,
+                (float)m_waveformRenderer->getBreadth() / 2);
```
daschuer, line 98:
```suggestion
                static_cast<float>(m_waveformRenderer->getLength()) / 2,
                static_cast<float>(m_waveformRenderer->getBreadth()) / 2);
```

src/waveform/renderers/allshader/waveformrendererslipmode.h
```diff
@@ -0,0 +1,42 @@
...
+#include "util/class.h"
+#include "util/performancetimer.h"
+#include "waveform/renderers/allshader/waveformrenderer.h"
+
+class ControlProxy;
+class QDomNode;
+class SkinContext;
+
+namespace allshader {
+class WaveformRendererSlipMode;
```
daschuer, line 16: Is this forward declaration redundant?

daschuer (changes requested):
Thank you for this nice new feature. I have added some comments for improving readability.

src/shaders/slipmodeshader.cpp
acolombier: I was wondering that too since, I have copied this formatting in another waveform file. Perhaps it is somehow considered valid?

src/waveform/renderers/allshader/waveformrenderbeat.cpp
```diff
@@ -10,8 +10,10 @@
 
 namespace allshader {
 
-WaveformRenderBeat::WaveformRenderBeat(WaveformWidgetRenderer* waveformWidget)
-        : WaveformRenderer(waveformWidget) {
+WaveformRenderBeat::WaveformRenderBeat(WaveformWidgetRenderer* waveformWidget,
+        ::WaveformRendererAbstract::Type type)
```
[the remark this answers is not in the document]
acolombier: Yes it - `WaveformRendererAbstract` is defined both in `src/waveform/renderers/waveformrendererabstract.h` and `src/waveform/renderers/allshader/waveformrendererabstract.h`. Perhaps worth addressing the name conflict in another PR?

src/waveform/renderers/waveformwidgetrenderer.cpp
```diff
@@ -17,6 +17,7 @@ const double WaveformWidgetRenderer::s_defaultPlayMarkerPosition = 0.5;
 
 namespace {
 constexpr int kDefaultDimBrightThreshold = 127;
+constexpr int kWaveformRendererAbstractTypeEnd = ::WaveformRendererAbstract::Slip + 1;
```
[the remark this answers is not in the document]
acolombier: It is! I just removed it entirely and changed the loop condition

src/shaders/slipmodeshader.cpp
acolombier: I think this because it's part of the GLSL fragment shader so not parsed by clang-format

src/waveform/renderers/allshader/waveformrendererslipmode.h
acolombier: It looks like it is - the class cannot be declared otherwise. I tried declaring in the namespace block but then the base class cannot resolve.

daschuer:
Unfortunately a conflict has been developed.

daschuer (approved):
LGTM, Thank you.

acolombier:
Conflict solved now

daschuer:
Thank you for this nice eye candy.

fwcd:
Love this feature! The only oddity I've noticed is that the markers also "shift" to the bottom half when using a waveform that doesn't support this split view in slip mode:

https://github.com/user-attachments/assets/37395c8f-efde-48ba-b207-f4651f66c437

acolombier:
This is due the work we had started with @m0dB to mutualise all the allshader waveform - I will try to finish implementing the effect on the other waveform type now that it should be easier!

acolombier:
Created #13499 to keep track!

PR #13007, opened 2024-03-27
(fix) Waveform scratch crossing loop boundaries

ronso0:
Previously, scratching with the waveform and crossing loop boundaries at slow rates caused an infinite wrap-around 'loop' even after the mouse was stopped (while still pressed).
Probably affects Spinny scratching, too, but I didn't test it.

PositionScratchController::process() now knows if the play position was wrapped around and calculates the true sample distance travelled.
_I notice this would also be helpful for and probably simplify #12951 hotcue_X_indicator control_

TL;DR
PositionScratchController didn't know the play position was wrapped around, so it
interpreted the raw (huge) sample delta as play pos move and calculated a huge rate, which was
then applied by the engine, resulting in a (true) long sample distance in the next run,
which caused another loop wrap-around, and so on...

### DIFF
--- a/src/engine/controls/ratecontrol.cpp
+++ b/src/engine/controls/ratecontrol.cpp
@@ -34,6 +34,9 @@ RateControl::RateControl(const QString& group,
         UserSettingsPointer pConfig)
         : EngineControl(group, pConfig),
           m_pBpmControl(nullptr),
+          m_wrapAroundCount(0),
+          m_jumpPos(mixxx::audio::FramePos()),
+          m_targetPos(mixxx::audio::FramePos()),
           m_bTempStarted(false),
           m_tempRateRatio(0.0),
           m_dRateTempRampChange(0.0) {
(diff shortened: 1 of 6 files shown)


ronso0:
This is now cleaned up.
You may pick d4a0daa179 to add some debug output.

src/engine/controls/ratecontrol.cpp
```diff
@@ -603,3 +616,18 @@ bool RateControl::isReverseButtonPressed() {
     }
     return false;
 }
+
+void RateControl::notifyWrapAround(mixxx::audio::FramePos triggerPos,
+        mixxx::audio::FramePos targetPos) {
+    // TODO FramePos only for easy validation.
+    // Check how to get kLegacyInvalidEnginePosition in order to use sample pos
+    // for the entire
```
JoergAtGithub, line 624: Why do you want to switch from frame position to sample position?

JoergAtGithub (commented):
I tested this on Windows and everything works as expected!

Swiftb0y:
I remember mentioning this in an issue or zulip post sometime ago, but I can't find it anymore

ronso0:
yeah, we touched this topic sometime, but I can't find it anymore either.

src/engine/controls/ratecontrol.cpp
ronso0: it's passed only to PositionScratchController and uses only sample positions for processing anyway, so I thought we'd better convert it early. No issue though, and getting kLegacyInvalidEnginePosition here requires more includes, so I can as well remove the TODO.

ronso0:
> I tested this on Windows and everything works as expected!

An important test case, which failed with my previous attempts, is a fast throwback with a very short loop (< 1 beat), then hold the mouse button until scratching (inertia) stopped. This should perform exactly like a throwback with no loop.
And scracthing should also **sound** right, not just look reasonable on the waveforms.

JoergAtGithub:
> > I tested this on Windows and everything works as expected!
> 
> An important test case, which failed with my previous attempts, is a fast throwback with a very short loop (< 1 beat), then hold the mouse button until scratching (inertia) stopped. This should perform exactly like a throwback with no loop. And scracthing should also **sound** right, not just look reasonable on the waveforms.

I tested this, and it worked correct!

ronso0:
Great. So I remove the TODO and this is ready?

ronso0:
Done.

JoergAtGithub:
LGTM! Thank you!

PR #13008, opened 2024-03-27
Support for bulk devices on Windows and Mac

acolombier:
Allow bulk devices to be opened on Windows when using `WinUSB` and Mac.

### DIFF
--- a/src/controllers/bulk/bulkcontroller.cpp
+++ b/src/controllers/bulk/bulkcontroller.cpp
@@ -70,13 +70,13 @@ BulkController::BulkController(libusb_context* context,
                   get_string(handle, desc->iSerialNumber))),
           m_context(context),
           m_phandle(handle),
-          in_epaddr(0),
-          out_epaddr(0) {
-    vendor_id = desc->idVendor;
-    product_id = desc->idProduct;
+          m_inEndpointAddr(0),
+          m_outEndpointAddr(0) {
+    m_vendorId = desc->idVendor;
+    m_productId = desc->idProduct;
 
-    manufacturer = get_string(handle, desc->iManufacturer);
-    product = get_string(handle, desc->iProduct);
+    m_manufacturer = get_string(handle, desc->iManufacturer);
+    m_product = get_string(handle, desc->iProduct);
     m_sUID = get_string(handle, desc->iSerialNumber);
 
     setDeviceCategory(tr("USB Controller"));
@@ -122,13 +122,20 @@ bool BulkController::matchProductInfo(const ProductInfo& product) {
     bool ok;
     // Product and vendor match is always required
     value = product.vendor_id.toInt(&ok, 16);
-    if (!ok || vendor_id != value) {
+    if (!ok || m_vendorId != value) {
         return false;
     }
     value = product.product_id.toInt(&ok, 16);
-    if (!ok || product_id != value) {
+    if (!ok || m_productId != value) {
         return false;
     }
+
+#if defined(__WINDOWS__) || defined(__APPLE__)
+    value = product.interface_number.toInt(&ok, 16);
+    if (!ok || m_interfaceNumber != static_cast<unsigned int>(value)) {
+        return false;
+    }
+#endif
 
     // Match found
     return true;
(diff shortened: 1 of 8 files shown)


src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -130,6 +130,13 @@ bool BulkController::matchProductInfo(const ProductInfo& product) {
         return false;
     }
 
+#ifdef _WIN32
```
JoergAtGithub, line 133:
```suggestion
#ifdef __WINDOWS__
```
At other places, we use our own definition __WINDOWS__ instead of _WIN32, which is correct of cause too.

src/controllers/bulk/bulkcontroller.h
```diff
@@ -76,6 +76,9 @@ class BulkController : public Controller {
     unsigned short product_id;
     unsigned char in_epaddr;
     unsigned char out_epaddr;
+#ifdef _WIN32
+    unsigned int interface_number;
+#endif
     QString manufacturer;
     QString product;
```
JoergAtGithub, line 83:
```suggestion
    unsigned short mProductId;
    unsigned char mInEndpointAddr;
    unsigned char mOutEndpointAddr;
#ifdef _WIN32
    unsigned int mInterfaceNumber;
#endif
    QString mManufacturer;
    QString mProduct;
```

src/controllers/controllermappinginfo.h
```diff
@@ -13,15 +13,13 @@ struct ProductInfo {
...
-    QString out_epaddr;
-
-    // Bulk-specific
     QString usage_page;
     QString usage;
     QString interface_number;
+
+    // Bulk-specific
+    QString in_epaddr;
+    QString out_epaddr;
```
JoergAtGithub, line 22:
```suggestion
    QString interface_number;
    
    // HID-specific
    QString usage_page;
    QString usage;

    // Bulk-specific
    QString in_epaddr;
    QString out_epaddr;
```
Every USB device has interface numbers, but only HID works with Uage/UsagePage

JoergAtGithub (commented):
Some general suggestions. I've no USB Bulk device to test this.

src/controllers/bulk/bulkcontroller.h
acolombier: I've renamed the fields with `m_<lower>` as that's what I used in other places

acolombier:
I took the initiative on expending the scope of this PR to also include Mac support as it seems it needs the same changes.

This was tested with S4 Mk3 screen

src/controllers/bulk/bulkcontroller.cpp
JoergAtGithub: This got lost in the last in the last push. Better use indivitual commuts, to leep track.

JoergAtGithub:
You need also to enabled BULK support in our Windows build setups:
https://github.com/mixxxdj/mixxx/blob/99406cb7eb3a90c1048ceee4920b22ddd001e79a/tools/windows_buildenv.bat#L216

https://github.com/mixxxdj/mixxx/blob/99406cb7eb3a90c1048ceee4920b22ddd001e79a/.github/workflows/build.yml#L96

acolombier:
Just a heads up here - there seems to be reports of issues on Zulip about Mac and that feature. Let's hold on merging till we have further details and potentially fix the usecase first.

I did some testing on MacOS M1 (Sonoma I believe) and it worked but there might be edge case other arch/OS version

JoergAtGithub:
I wonder what the kernel driver is, that needs to be detached, a generic one, that comes with the OS like WinUSB, or a Native Instruments driver for the S4 Mk3.
Maybe the detach of a Kernel driver is not necessary in case of Linux, just because this OS is not supported by Native Instruments?

acolombier:
> Maybe the detach of a Kernel driver is not necessary in case of Linux, just because this OS is not supported by Native Instruments?

Yes, I think your theory is correct. 
Do you think this means we should potentially rework the Linux part to make it compatible with the fact that NI may (very unlikely) release NI kernel drivers too (and so require a similar libusb opening procedure?

JoergAtGithub:
I'm not sure, but it would be good to test this without the NI driver package as well, to get a better understanding.
If the NI drivers are installed, we also need stop the service "NIHardwareService" on Windows, because it otherwise communicates via HID in parallel to Mixxx.

acolombier:
It's worth to mention I had to change the kernel driver from the NI to WinUSB one, so I guess this usecase isn't possible, or is it? I'm not sure how we could allow both kernels (And services) to be toggled on demand, and since I don't have much experience on Windows (nor the setup, and the interest to be honest), it's hard for me to work on a robust comprehensive solution here.

acolombier:
Could we consider merging this as-is as it is improving the overall bulk support on non-Linux platform (especially Mac, allowing bulk to work out of the box on MacOS 14).
Happy to create issues for the installer/further work to do on the Windows side to get that also working out of the box. In the meantime, I'd be happy to create a manual section to help installing WinUSB using some utility like [Zadig](https://zadig.akeo.ie/)

JoergAtGithub (approved):
The code is fine now, and I will merge it therefore!
But this functionality is not enough, that a normal Windows user can use USB Bulk devices as on Linux. This is because the WinUSB driver registration for the device must be done manually by an Administrator using a tool like Zadig.

JoergAtGithub:
Thank you, for addressing this long pending limitation!

PR #13010, opened 2024-03-27
Reduce log noise with HID device

acolombier:
Currently, when running in `--log-level debug`, logs can be unusable due to the amount of messages related to received HID packets. 

This PR makes these logging messages only if `--controller-debug` is set


JoergAtGithub:
Tested this locally, and works as expected. Code LGTM! Thank you!

PR #13011, opened 2024-03-27
fix: allow missing COs on QML component

acolombier:
This PR allows `Mixxx.ControlProxy` to reference CO that don't exist without crashing. This is to ensure backward compatibility the controller JS engine which allows to connect to in-existing COs and let the caller handle invalid (see `src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp`/`ControllerScriptInterfaceLegacy::makeConnectionInternal`.

This is a dependency for #11407

### DIFF
--- a/src/qml/qmlcontrolproxy.cpp
+++ b/src/qml/qmlcontrolproxy.cpp
@@ -1,6 +1,7 @@
 #include "qml/qmlcontrolproxy.h"
 
 #include "moc_qmlcontrolproxy.cpp"
+#include "util/cmdlineargs.h"
 
 namespace mixxx {
 namespace qml {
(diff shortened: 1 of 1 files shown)


Holzhaus:
Have you considered to you make this depend on a boolean Q_PROPERTY that is off by default? So that we need to be explicit about it that you 2ant that behavior?

Or, if that is too strict, maybe we could check if the `--developer` flag was passed on the cmdline and only use `AllowInvalidOrMissing` if it's *not* passed?

Holzhaus:
So that we do get a debug assert when doing QML skin development?

acolombier:
> Or, if that is too strict, maybe we could check if the --developer flag was passed on the cmdline and only use AllowInvalidOrMissing if it's not passed?

Yes, I was thinking the same if we wanted to keep the asserting. Will update that!

acolombier:
@Holzhaus are you happy with the look of this fix now?

Holzhaus (approved):

PR #13012, opened 2024-03-27
iTunes/Serato/Traktor/Rhythmbox: Print error if library file could not be opened

fwcd:
Just a minor improvement of the error message to aid debugging.


fwcd:
Thinking about it, this would probably also be useful in the other library integrations (Traktor, Serato etc)

daschuer (approved):
LGTM, Thank you.

PR #13013, opened 2024-03-27
iTunes: Obtain FileAccess before accessing iTunes XML

fwcd:
This fixes an issue where opening an iTunes XML outside sandboxed directories would error in App Sandbox-enabled macOS builds, triggering the following dialog:

Screenshot 2024-03-27 at 23 33 34

#13012 reveals that the error in this case is "Operation not permitted". Obtaining a `mixxx::FileAccess` fixes the issue.

### DIFF
--- a/src/library/itunes/itunesxmlimporter.cpp
+++ b/src/library/itunes/itunesxmlimporter.cpp
@@ -11,6 +11,8 @@
 #include "library/itunes/itunesimporter.h"
 #include "library/itunes/ituneslocalhosttoken.h"
 #include "library/treeitem.h"
+#include "util/fileaccess.h"
+#include "util/fileinfo.h"
 #include "util/lcs.h"
 
 namespace {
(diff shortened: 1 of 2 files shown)


src/library/itunes/itunesxmlimporter.cpp
```diff
@@ -65,6 +67,9 @@ ITunesImport ITunesXMLImporter::importLibrary() {
     ITunesImport iTunesImport;
     bool isMusicFolderLocatedAfterTracks = false;
 
+    // In sandboxed builds we have to obtain an access token
+    auto access = mixxx::FileAccess(mixxx::FileInfo(m_xmlFilePath));
```
fwcd, line 71: Two notes on this:

- It would be nice if someone could verify that this still works properly on Windows/Linux (though I'd be very surprised if there were any issues given that the browse feature obtains access tokens in the same manner).
- We should probably also do this for the Serato/Traktor/... integrations

daschuer:
The error message looks confusing for this case. Can you adjust it as well? 
It is issued here, where we not even go a root item: https://github.com/mixxxdj/mixxx/blob/5ee50d77d794eeb267591c44d76db4066c540052/src/library/itunes/itunesfeature.cpp#L369

daschuer (approved):
LGTM, thank you. 

Can you file an issue for investigating the problem with other third party libraries?

PR #13016, opened 2024-03-28
(fix) Library: respect keepHiddenTracks flag when exporting a History playlist

### LINKED ISSUE #13000: Exporting a playlist removes history entries
### Bug Description

Trying to export a playlist from the history pane when tracks in the library are marked as deleted (field `mixxx_deleted` in the `library` table equals to 1) will wipe their entries from the history. 
Pretty unpleasant issue since this happens silently and you can easily lose a lot of important entries.

### Version

2.4.0

### OS

WIndows 10

ronso0:
Prevents removing hidden tracks permanently from a History playlist when it is exported.

Fixes #13000


daschuer (approved):
LGTM, Thank you.

PR #13022, opened 2024-03-29
Remove optimization mode "legacy" for 2.5

JoergAtGithub:
With the switch to Qt6, we raise the system requirment a lot: https://doc.qt.io/qt-6.2/supported-platforms.html#desktop-platforms Support for i386 legacy instruction set makes no sense anymore and on Windows it did nothing than printing a message anyway.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -307,8 +307,8 @@ endif()
 # Optimizations
 #
 
-set(OPTIMIZE "portable" CACHE STRING "Optimization and Tuning (set to off, portable, native, legacy)")
-set_property(CACHE OPTIMIZE PROPERTY STRINGS "off" "portable" "native" "legacy")
+set(OPTIMIZE "portable" CACHE STRING "Optimization and Tuning (set to off, portable, native)")
+set_property(CACHE OPTIMIZE PROPERTY STRINGS "off" "portable" "native")
 string(TOLOWER "${OPTIMIZE}" OPTIMIZE)
 message(STATUS "Optimization level: ${OPTIMIZE}")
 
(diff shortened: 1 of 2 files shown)


daschuer:
In general I am OK with it. 

Just some thoughts about AVX support: 

It would be nice to distribute Mixxx with AVX support. Qt has for instance the flag `QT_HAVE_AVX` and some of our sample handling code will run twice as fast.

AVX was introduced 2011 with Intel's the Sandy Bridge processors and AMD's Bulldozers. 
This will drop support for instance Core 2 Duo Support. 

Is it time for it now or soon? 

In that case we will probably revive "legacy"? Or will we rename it? 

Can we predict the upcoming optimization names and incorporate them here?

JoergAtGithub:
I suggested using AVX for 2.5 already before:

- There is not Apple computer supported by `macOS 11` (Qt6.5 requirement) anymore, that has no `AVX`
- `Windows 10 (1809 or later)` (Qt6.5 requirement):
  - All supported AMD processors have `AVX`
  - Most supported Intel processors have at least `SSE4.2`
  - The absolute minimum requirement for Windows 10 64Bit is: `PAE, NX-Bit, SSE2, CMPXCHG16b, PrefetchW, LAHF/SAHF`

But I don't think, that the current legacy switch is the right thing anyway, because the biggest impact of SSE4.2/AVX would be on the DSP heavy libraries in the VCPKG buildenv like Rubberband.

daschuer:
In such a case the switch should also pick the vcpkg triplet with the equivalent flags. 

Can you draft than names you have in mind?

JoergAtGithub:
ASAIK triplets are not intended for this. I think we need to set something like this:
            vcpkg_c_flags: /arch:avx
            vcpkg_cxx_flags: /arch:avx

JoergAtGithub:
Do we have enough diskspace and CPU time for yet another buildenv?

daschuer:
Disk space is largely wasted with unused legacy. 
I am in the progress to clean that up. 
The CPU is an issue because of the CO² footprint.
Maybe we can switch to AVX duing development and build a non AVX build only after a release? 

But a usually we need a benchmark to check if the difference is significant enough to or if we want AVX512 or such.

JoergAtGithub:
> The CPU is an issue because of the CO² footprint.

If Mixxx runs more efficient on many thousands of end users computers, we reduce more CO², than the VCPKG build runner consumes.

daschuer:
Yes of cause. The question is the extra CPU for the non AVX build.

daschuer:
The question relevant to this PR is: will we have again legacy, after we move portable to AVX(2) ?

Swiftb0y:
ossia score [tried going above sse2](https://github.com/ossia/score/issues/1013) a couple years ago and reverted. Not sure if they have tried since.
I think instead of raising the bar for the entire application it would make sense to look into [cpu dispatching](https://johnnysswlab.com/cpu-dispatching-make-your-code-both-portable-and-fast/) for the hot parts of the code.

PR #13023, opened 2024-03-29
prohibit un-replace when deck is playing

### LINKED ISSUE #12906: Un-eject / un-replace does not respect "Loading a track when deck is playing" setting
### Bug Description

- Dubbelclick on "Load/EjectSwitch" of a Deck does not clone Deck, even when the option is selected  in settings
- Switching 2 - 4 Decks and back (multiple times) and between the switching clicking on "Load/EjectSwitch" of a playing Deck  causes "Ejecting" of the playing track, 
-  I suppose the option "disable load a track to a playing deck"  means also disable the possibility to eject a track while playing (as it is in 2.3.6)
-> I'm testing the main from 29/2/2024

### Version

2.5-alpha-255-g2d2d5f304b

### OS

Win10 / Win 11

ronso0:
Fixes #12906

### DIFF
--- a/src/mixer/basetrackplayer.cpp
+++ b/src/mixer/basetrackplayer.cpp
@@ -327,6 +327,12 @@ void BaseTrackPlayerImpl::slotEjectTrack(double v) {
         return;
     }
 
+    // Don't allow eject while playing a track. We don't need to lock to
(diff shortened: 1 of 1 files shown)


daschuer (approved):
LGTM, thank you.

PR #13027, opened 2024-03-30
Renamed symbols (Parameter -> NormalizedValue) and added aliases for controller script backward compatibility

JoergAtGithub:
In the COs and the controller script API, the meaningless term "Parameter" was systematically used instead of the obviously correct term "Normalized Value".
This is particularly problematic as it is a symbol that is exposed to the user through our controller scripting API. A user has no chance to understand, that "Parameter" indicates a value with normalized range - unless he looks up in the documentation for it.

I renamed the affected methods therefore, and added aliases to the controller scripting API for backward compatibility.

Additionally I expanded the term Midi in some CO methods to Midi7Bit, as nowadays Midi allows larger ranges than 0...127. But this range is meant here. 

Last change was extending the controllerscriptenginelegacy_test.cpp to cover the previously uncovered methods which were renamed, and a check for the aliases itself.

### DIFF
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -61,6 +61,15 @@ declare namespace engine {
      * @param name Name of the control e.g. "play_indicator"
      * @returns Value of the control normalized to range of 0..1
      */
+    function getNormalizedValue(group: string, name: string): number;
+
+    /**
+     * Gets the control value normalized to a range of 0..1
+     * @param group Group of the control e.g. "[Channel1]"
+     * @param name Name of the control e.g. "play_indicator"
+     * @returns Value of the control normalized to range of 0..1
+     * @deprecated Use {@link getNormalizedValue} instead
+     */
     function getParameter(group: string, name: string): number;
 
     /**
@@ -69,6 +78,15 @@ declare namespace engine {
      * @param group Group of the control e.g. "[Channel1]"
      * @param name Name of the control e.g. "play_indicator"
      * @param newValue Value to be set, normalized to a range of 0..1
+     */
+    function setNormalizedValue(group: string, name: string, newValue: number): void;
+
+    /**
+     * Sets the control value specified with normalized range of 0..1
+     * @param group Group of the control e.g. "[Channel1]"
+     * @param name Name of the control e.g. "play_indicator"
+     * @param newValue Value to be set, normalized to a range of 0..1
+     * @deprecated Use {@link setNormalizedValue} instead
      */
     function setParameter(group: string, name: string, newValue: number): void;
 
@@ -81,6 +99,18 @@ declare namespace engine {
      * @param value Value with the controls range according Mixxx Controls manual page:
      *              https://manual.mixxx.org/latest/chapters/appendix/mixxx_controls.html
      * @returns Value normalized to range of 0..1
+     */
+    function getNormalizedValueForValue(group: string, name: string, value: number): number;
+
+    /**
+     * Normalizes a specified value using the range of the given control,
+     * to the range of 0..1
+     * @param group Group of the control e.g. "[Channel1]"
+     * @param name Name of the control e.g. "play_indicator"
+     * @param value Value with the controls range according Mixxx Controls manual page:
+     *              https://manual.mixxx.org/latest/chapters/appendix/mixxx_controls.html
+     * @returns Value normalized to range of 0..1
+     * @deprecated Use {@link getNormalizedValueForValue} instead
      */
     function getParameterForValue(group: string, name: string, value: number): number;
 
@@ -108,6 +138,15 @@ declare namespace engine {
      * @param group Group of the control e.g. "[Channel1]"
      * @param name Name of the control e.g. "play_indicator"
      * @returns Default value of the specified control normalized to range of 0..1
+     */
+    function getNormalizedDefaultValue(group: string, name: string): number;
+
+    /**
+     * Returns the default value of a control, normalized to a range of 0..1
+     * @param group Group of the control e.g. "[Channel1]"
+     * @param name Name of the control e.g. "play_indicator"
+     * @returns Default value of the specified control normalized to range of 0..1
+     * @deprecated Use {@link getNormalizedDefaultValue} instead
      */
     function getDefaultParameter(group: string, name: string): number;
 
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -58,7 +58,7 @@ declare namespace engine {
      * Gets the control value normalized to a range of 0..1
      *
      * @param group Group of the control e.g. "[Channel1]"
-     * @param name Name of the control e.g. "play_indicator"
+     * @param name Name of the control e.g. "rate" or "pregain"
      * @returns Value of the control normalized to range of 0..1
      */
     function getNormalizedValue(group: string, name: string): number;
@@ -66,7 +66,7 @@ declare namespace engine {
     /**
      * Gets the control value normalized to a range of 0..1
      * @param group Group of the control e.g. "[Channel1]"
-     * @param name Name of the control e.g. "play_indicator"
+     * @param name Name of the control e.g. "rate" or "pregain"
      * @returns Value of the control normalized to range of 0..1
      * @deprecated Use {@link getNormalizedValue} instead
      */
@@ -76,7 +76,7 @@ declare namespace engine {
      * Sets the control value specified with normalized range of 0..1
      *
      * @param group Group of the control e.g. "[Channel1]"
-     * @param name Name of the control e.g. "play_indicator"
+     * @param name Name of the control e.g.  "rate" or "pregain"
      * @param newValue Value to be set, normalized to a range of 0..1
      */
     function setNormalizedValue(group: string, name: string, newValue: number): void;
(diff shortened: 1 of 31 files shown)


daschuer:

daschuer:
Unfortunately It is not always the normalized value. In case of knobs and sliders, it reflects the potentiometer  position normalized to the range of 0 ... 1. For all other COs it is equal to value. 

I am a bit torn if the term "NormilizedValue" just shifts the confusion. Maybe we can find a name the emphasize that this is only relevant for sliders and knobs. Something like PotentiometerPos, PotPosition, controlPosition, or just position.? 

What do you think?

res/controllers/engine-api.d.ts
```diff
@@ -61,6 +61,15 @@ declare namespace engine {
      * @param name Name of the control e.g. "play_indicator"
      * @returns Value of the control normalized to range of 0..1
      */
+    function getNormalizedValue(group: string, name: string): number;
+
+    /**
+     * Gets the control value normalized to a range of 0..1
```
daschuer, line 67:
```suggestion
     * Returns the knob / slider position in the range of 0..1 or the value if there  is no position associated.
```

res/controllers/engine-api.d.ts
```diff
@@ -61,6 +61,15 @@ declare namespace engine {
...
+     * @param group Group of the control e.g. "[Channel1]"
+     * @param name Name of the control e.g. "play_indicator"
```
daschuer, line 69: It is better to use an slider as example
```suggestion
     * @param name Name of the control e.g. "rate"
```

daschuer: The same for the description above.

res/controllers/engine-api.d.ts
```diff
@@ -70,6 +79,15 @@ declare namespace engine {
      * @param name Name of the control e.g. "play_indicator"
      * @param newValue Value to be set, normalized to a range of 0..1
      */
+    function setNormalizedValue(group: string, name: string, newValue: number): void;
+
+    /**
+     * Sets the control value specified with normalized range of 0..1
```
daschuer, line 85:
```suggestion
     * Sets the control value via the position on the knob / slider scale in the range of 0..1 or the value if there  is no position associated.
```

res/controllers/engine-api.d.ts
```diff
@@ -82,6 +100,18 @@ declare namespace engine {
      *              https://manual.mixxx.org/latest/chapters/appendix/mixxx_controls.html
      * @returns Value normalized to range of 0..1
      */
+    function getNormalizedValueForValue(group: string, name: string, value: number): number;
+
+    /**
+     * Normalizes a specified value using the range of the given control,
+     * to the range of 0..1
```
daschuer, line 107:
```suggestion
     * Returns the knob / slider position in the range of 0..1 or the unchanged value if no 
     * no position associated.
```

res/controllers/engine-api.d.ts
```diff
@@ -82,6 +100,18 @@ declare namespace engine {
...
+     * @param group Group of the control e.g. "[Channel1]"
+     * @param name Name of the control e.g. "play_indicator"
```
daschuer, line 109:
```suggestion
     * @param name Name of the control e.g. "rate"
```

JoergAtGithub:
> For all other COs it is equal to value.

In these cases I would suggest to print a warning that set/getNormalizedValues is not supported for non-Pot-Meter-COs, and that set/getValue should be used for this CO instead.

res/controllers/engine-api.d.ts
```diff
@@ -109,6 +139,15 @@ declare namespace engine {
      * @param name Name of the control e.g. "play_indicator"
      * @returns Default value of the specified control normalized to range of 0..1
      */
+    function getNormalizedDefaultValue(group: string, name: string): number;
+
+    /**
+     * Returns the default value of a control, normalized to a range of 0..1
```
daschuer, line 145:
```suggestion
     * Returns the default knob / slider position range of 0..1, or the default value if no no position associated.
```

res/controllers/engine-api.d.ts
```diff
@@ -109,6 +139,15 @@ declare namespace engine {
...
+     * @param group Group of the control e.g. "[Channel1]"
+     * @param name Name of the control e.g. "play_indicator"
```
daschuer, line 147:
```suggestion
     * @param name Name of the control e.g. "rate"
```

daschuer (changes requested):

daschuer:
>>  For all other COs it is equal to value.

> In these cases I would suggest to print a warning that set/getNormalizedValues is not supported for non-Pot-Meter-COs, and that set/getValue should be used for this CO instead.

Good idea. In these case the "parameter" value should not be used.

JoergAtGithub:
> > >
> 
> > In these cases I would suggest to print a warning that set/getNormalizedValues is not supported for non-Pot-Meter-COs, and that set/getValue should be used for this CO instead.
> 
> Good idea. In these case the "parameter" value should not be used.

Sure, backward compatibilty is prio no. 1 for everything controller mapping related. I set the PR to draft, until I've addressed this.

daschuer:
Did you consider to adjust the name?

JoergAtGithub:
> Did you consider to adjust the name?

To be honest, neither Potentiometer nor Position is related to what the function does, and to have this selfdescriptive name is the sole reason I work on this.

daschuer:
>> Did you consider to adjust the name?

> To be honest, neither Potentiometer nor Position is related to what the function does, and to have this selfdescriptive name is the sole reason I work on this.

Normalization to 0..1 alone is IMHO not descriptive enough. For my understanding it is always scaled to the linear position of the slider / knob scale or otherwise defaults to value. This is at least true for rate, pregain and volume. Please give examples where neither Potentiometer nor Position is related.

JoergAtGithub:
An example is the normalized value returned by engine.getParameter is useful for VU-Meter, just multiply it by the number of LEDs in your controler and you get what you need to light them up in the right brightness.

The typical use case for the normalization done by engine.getParameterForValue is range checking in an assert like code.

The problem I'm addressing with this PR is that, as in any good API, the other function names of the controller scripting API describe what they do/implement, but not the "Parameter" ones.
Since terms like "location" or "potentiometer" don't do this either, this would contradict the intent of this PR.

daschuer:
The 'vu_meter' CO is a plain ContolObect it has no potentiometer position assisted and using get/setPatameter may issue the discussed warning. 

> The typical use case for the normalization done by engine.getParameterForValue is range checking in an assert like code.

This does not work for plain ControlObejects and may issue a warning if dome anyway. 

I agree tin general, hat parameter was a name without any meaning so let's find a name that solves the issue. 
NormalizedValue just shifts the problem. 

The relationship is defined here: 
https://github.com/mixxxdj/mixxx/blob/2145f5f7dec746ab314fa136ca2707e15a3a620b/src/control/controlbehavior.cpp We may go through it one by one and check a possible name.

JoergAtGithub:
Since it automatically selects the normalization mode depending on the CO type, what about "AutoNormalizedValue" ?

daschuer:
There term "normalization" does not fit well IMHO.

We need a name for this

* Null (Plain VonzrolObject): value
* ControlNumericBehavior: value 
* ControlPotmeterBehavior: pot position 
* ControlLogPotmeterBehavior: pot position 
* ControlLogInvPotmeterBehevior: pot position
* ControlLinInvPotmeterBehavior: pot position
* ControlAudioTaperPotBehavior: pot position
* ControlTTRotaryBehavior: ?
* ControlLinSteppedIntPotBehavior: pot position
* ControlPushButtonBehavior: value

JoergAtGithub:
Can you describe what the function does in other terms? I never heard another term than normalization for this operation.

daschuer:
The parameter/value pair has been introduced to allow simple mappings of the controller hardware to the values in Mixxx. 
The abstract model of such a control is something like this: 
`Min ----- Center ----- Max`
where min in represented as 0.0 center as 0.5 and max as 1.0. 
The control behaviour is used to "map" this model to the value of the business logic, like the "rate" multiplier in the range of -1, 0 and +1 or in case of volume the logarithmic 0, 2.4 and 1. 

The advantage is that this logic is part of Mixxx and there is no need to implement it in mappings. This makes sure that controller and GUI sliders behave the same. 

The mapping author can override it by changing the "value" directly.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -158,24 +158,25 @@ void ControllerScriptInterfaceLegacy::setValue(
                 coScript->getKey(), ControlFlag::AllowMissingOrInvalid);
         if (pControl &&
                 !m_st.ignore(
-                        pControl, coScript->getParameterForValue(newValue))) {
+                        pControl, coScript->getNormalizedValueForValue(newValue))) {
```
daschuer, line 161: Soft takeover deals with the position of the slider, not with the business logic value in Mixxx. That's we we need to calculate here the position on the slider form the value.

src/qml/qmlcontrolproxy.cpp
```diff
@@ -175,7 +175,7 @@ void QmlControlProxy::reinitializeFromKey() {
 
 void QmlControlProxy::slotControlProxyValueChanged(double newValue) {
     emit valueChanged(newValue);
-    emit parameterChanged(m_pControlProxy->getParameter());
+    emit normalizedValueChanged(m_pControlProxy->getNormalizedValue());
```
daschuer, line 178: Why we do we emit two signals here?

PR #13028, opened 2024-03-30
fix: Replace deprecated `qAsConst` with `std::as_const`

Holzhaus:
This fixes build warnings/errors like this:

    In file included from src/controllers/legacycontrollersettings.h:5,
                     from src/controllers/legacycontrollersettingslayout.cpp:13:
    src/controllers/legacycontrollersettingsfactory.h: In static member function ‘static AbstractLegacyControllerSetting* LegacyControllerSettingBuilder::build(const QDomElement&)’:
    src/controllers/legacycontrollersettingsfactory.h:46:48: error: ‘constexpr typename std::add_const<_Tp>::type& qAsConst(T&) [with T = QList<LegacyControllerSettingBuilder::SupportedSetting>; typename std::add_const<_Tp>::type = const QList<LegacyControllerSettingBuilder::SupportedSetting>]’ is deprecated: Use std::as_const() instead. [-Werror=deprecated-declarations]
       46 |         for (const auto& settingType : qAsConst(instance()->m_supportedSettings)) {
          |                                        ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### DIFF
--- a/src/controllers/dlgprefcontroller.cpp
+++ b/src/controllers/dlgprefcontroller.cpp
@@ -853,7 +853,7 @@ void DlgPrefController::slotShowMapping(std::shared_ptr<LegacyControllerMapping>
         if (pLayout != nullptr && !settings.isEmpty()) {
             m_ui.groupBoxSettings->layout()->addWidget(pLayout->build(m_ui.groupBoxSettings));
 
-            for (const auto& setting : qAsConst(settings)) {
+            for (const auto& setting : std::as_const(settings)) {
                 connect(setting.get(),
                         &AbstractLegacyControllerSetting::changed,
                         this,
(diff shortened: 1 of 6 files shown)


Holzhaus:
By the way, we should check out CI settings, because I wonder how this has passed CI checks.

daschuer (approved):
LGTM, thank you.

daschuer:
qsConst() is deprecated since Qt 6.6 our CI is using Qt 6.5 LTS. That's why it was not detected earlier.

Swiftb0y:
I think the fact that Qt is deprecating features that have been replaced by solutions from the `stdlib` is a good argument to favor `std::` solutions of Qt solutions in general when possible...

PR #13032, opened 2024-03-31
(fix) turn missing effects assert into warning

daschuer:
This fixes the handling of no longer available effects. 
Before Mixxx was aborted by an assertion. Now the missing effect is handle gracefully, and just not displayed.
It is still a warning logged.


ronso0 (approved):
Thank you, LGTM

FWIW now I think we may drop the --- in this part of the preferences, but it's unrelated to this fix.

ronso0:
I'll do a test soon.

ronso0:
Works as desired. Thanks!

PR #13040, opened 2024-04-02
Add test button(logs midi values)

tech-bash:
Fix: #10176 
Added a `test` button on the main controller page for now, setting it to the left of 'help' and then keeping it active(if any MIDIs are found) will be done further. Is the placement ok?


ronso0:
Tbh, since even this minimal commit doesn't even build, I doubt you really want to solve this.
Please avoid wasting developer time with half-baked attempts.

If you want to discuss the GUI / UX I suggest getting familiar with the use cases (try with a controller) and use GUI mockups instead.

Issue #13038, opened 2024-04-02
BPM/tempo lock via MIDI?


mxmilkiib:
### Feature Description

See also #9978 #13031

FrankwaP:
Hey :-)  I closed [my recent PR](https://github.com/mixxxdj/mixxx/pull/13025) as it was a duplicate of yours… but as I said creating controls for locking/unlocking the tempo was my second objective, and here there are: https://github.com/FrankwaP/mixxx/tree/tempo_locking_controls

I've created shortcuts and they work. I've also checked that they were displayed in the MIDI wizard and they are.

One annoying thing I've noticed however: when calling these when the channel is empty a segmentation fault! And this happens quite fast when one thinks these controls apply to the selected track in the library and forget to load it.  
I don't know if it's possible to do something smooth like the actual behavior of the right-click menu, which works on both the library and the channels…

FrankwaP:
I've modified my code so it's closer to yours, but my controls still crash Mixxx when the Channel is empty.

Isn't it why we have these lines in the slot functions ?
```cpp
    if (v <= 0) {
        return;
    }
```

ronso0:
> crash Mixxx when the Channel is empty

Sounds like you want to set beats on a track while the classes TrackPointer is null.
Just add null check like everywhere else in functions that operate on track objects.

FrankwaP:
[Aaaaaaand done!](https://github.com/mixxxdj/mixxx/commit/5c37d2d2ad95bfea1cb4a209541d2249f07cd578) Than you for the pointer, @ronso0 :-)

Do I make a PR ?


BTW I have fetched/merged @mxmilkiib 's work locally and made a specific/minimalist keyboard mapping for tracks preparation:
```cfg
LoadSelectedTrack Shift+Left

cue_default f
play d

beats_set_halve F1
beats_set_twothirds F2
beats_set_threehalves F3
beats_set_double F4

beats_translate_earlier a
beats_translate_later z
unlock_tempo e
lock_tempo r

beatjump_backward q
beatjump_1_backward Shift+q
beatjump_forward s
beatjump_1_forward Shift+s

hotcue_1_activate w
hotcue_2_activate x
hotcue_3_activate c
hotcue_4_activate v
hotcue_5_activate Ctrl+w
hotcue_6_activate Ctrl+x
hotcue_7_activate Ctrl+c
hotcue_8_activate Ctrl+v
hotcue_1_clear Shift+w
[...]

PR #13041, opened 2024-04-03
Tempo locking controls

FrankwaP:
This PR adds controls for tempo locking (`tempo_lock`) and unlocking (`locking`) so they can be used through keyboard shortcuts or MIDI inputs,a s suggested in [13038](https://github.com/mixxxdj/mixxx/issues/13038).

### DIFF
--- a/src/controllers/controlpickermenu.cpp
+++ b/src/controllers/controlpickermenu.cpp
@@ -272,6 +272,10 @@ ControlPickerMenu::ControlPickerMenu(QWidget* pParent)
             tr("Adjust Beatgrid - Match Alignment"),
             tr("Adjust beatgrid to match another playing deck."),
             pBpmMenu);
+    addDeckAndSamplerControl("bpm_toggle_lock",
+            tr("Toggle the BPM/beatgrid lock"),
+            tr("Toggle the BPM/beatgrid lock"),
+            pBpmMenu);
     pBpmMenu->addSeparator();
     addDeckAndSamplerControl("quantize", tr("Quantize Mode"), tr("Toggle quantize mode"), pBpmMenu);
 
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -130,6 +130,15 @@ BpmControl::BpmControl(const QString& group,
             this, &BpmControl::slotTapFilter,
             Qt::DirectConnection);
 
+    m_pToggleBpmLock = std::make_unique<ControlPushButton>(
+            ConfigKey(group, "bpm_toggle_lock"), false);
+    m_pToggleBpmLock->setKbdRepeatable(true);
+    connect(m_pToggleBpmLock.get(),
+            &ControlObject::valueChanged,
+            this,
+            &BpmControl::slotToggleBpmLock,
+            Qt::DirectConnection);
+
     // Measures distance from last beat in percentage: 0.5 = half-beat away.
     m_pThisBeatDistance = new ControlProxy(group, "beat_distance", this);
     m_pSyncMode = new ControlProxy(group, "sync_mode", this);
@@ -1019,6 +1028,18 @@ void BpmControl::slotBeatsTranslateMatchAlignment(double v) {
             pTrack->trySetBeats(*translatedBeats);
         }
     }
+}
+
+void BpmControl::slotToggleBpmLock(double v) {
+    if (v <= 0) {
+        return;
+    }
+    const TrackPointer pTrack = getEngineBuffer()->getLoadedTrack();
+    if (!pTrack) {
+        return;
+    }
+    bool locked = pTrack->isBpmLocked();
+    pTrack->setBpmLocked(!locked);
 }
 
 mixxx::Bpm BpmControl::updateLocalBpm() {
--- a/src/engine/controls/bpmcontrol.h
+++ b/src/engine/controls/bpmcontrol.h
@@ -108,6 +108,7 @@ class BpmControl : public EngineControl {
     void slotUpdateEngineBpm(double v = 0.0);
     void slotBeatsTranslate(double);
     void slotBeatsTranslateMatchAlignment(double);
+    void slotToggleBpmLock(double);
 
   private:
     SyncMode getSyncMode() const {
@@ -144,6 +145,8 @@ class BpmControl : public EngineControl {
     ControlPushButton* m_pTranslateBeatsEarlier;
     ControlPushButton* m_pTranslateBeatsLater;
     ControlEncoder* m_pTranslateBeatsMove;
+
+    std::unique_ptr<ControlPushButton> m_pToggleBpmLock;
 
     // The current effective BPM of the engine
     ControlLinPotmeter* m_pEngineBpm;


ronso0:
Thanks for this PR!

Since this is a deck control we should eep it simple and have just one control:
`bpm_toggle_lock`
"bpm" because "tempo" refers to the playback speed.

In case this shall be adopted for the library later on we will also make do with just one control, even if multiple tracks with locked/unlocked are selected:
* all the same lock state: toggle lock
* different lock state: lock all, trigger again to unlock all

ronso0:

Please comment here when done.
Thanks!

ronso0:
> Since this is a deck control we should eep it simple and have just one control:
bpm_toggle_lock

In case you adopt this you may very well create a fixup commit, squash all (the 4) commits and force-push.

src/controllers/controlpickermenu.cpp
```diff
@@ -272,14 +272,10 @@ ControlPickerMenu::ControlPickerMenu(QWidget* pParent)
             tr("Adjust Beatgrid - Match Alignment"),
             tr("Adjust beatgrid to match another playing deck."),
             pBpmMenu);
-    addDeckAndSamplerControl("lock_tempo",
+    addDeckAndSamplerControl("toggle_lock_bpm",
```
ronso0, line 275: Please use `bpm_toggle_lock` so we have 'hierarchical' naming structure

FrankwaP: Yeah that's a "git moment", as it's actually good on my computer and I have nothing to commit nor push :-) 

I'll see that in 3h.

FrankwaP: (Done)

src/controllers/controlpickermenu.cpp
```diff
@@ -272,6 +272,10 @@ ControlPickerMenu::ControlPickerMenu(QWidget* pParent)
             tr("Adjust Beatgrid - Match Alignment"),
             tr("Adjust beatgrid to match another playing deck."),
             pBpmMenu);
+    addDeckAndSamplerControl("bpm_toggle_lock",
+            tr("Lock the tempo/beatgrid"),
+            tr("Lock the tempo/beatgrid"),
```
ronso0, line 277: It should now be
"Toggle the BPM/beatgrid lock"

src/engine/controls/bpmcontrol.cpp
```diff
@@ -1021,6 +1030,17 @@ void BpmControl::slotBeatsTranslateMatchAlignment(double v) {
...
 
+void BpmControl::slotToggleBpmLock(double v) {
+    if (v <= 0) {
+        return;
+    }
+    const TrackPointer pTrack = getEngineBuffer()->getLoadedTrack();
+    if (!pTrack) {
+        return;
+    }
+    pTrack->setBpmLocked(not pTrack->isBpmLocked());
```
ronso0, line 1041: We don't use `not`, simply negate with `!`.
```suggestion
    pTrack->setBpmLocked(!pTrack->isBpmLocked());
```
This would make it even easier to read
```suggestion
    bool locked = pTrack->isBpmLocked();
    pTrack->setBpmLocked(!locked);
```

src/engine/controls/bpmcontrol.h
```diff
@@ -145,6 +146,9 @@ class BpmControl : public EngineControl {
     ControlPushButton* m_pTranslateBeatsLater;
     ControlEncoder* m_pTranslateBeatsMove;
 
+    // Locking/Unlocking BPM
+    std::unique_ptr<ControlPushButton> m_pToggleBpmLock;
```
ronso0, line 150: IMHO we need comments only if var names are not selfexplanatory, but I don't care if you keep it.

ronso0 (changes requested):
Thank, this is much cleaner now.

Some comments though

src/engine/controls/bpmcontrol.h
FrankwaP: Good points, thank you!

ronso0:
Thanks, I'll try this next days.

ronso0:
(feel free to ping me in case this slips off my table ; )

ronso0 (approved):
perfecto, LGTM

ronso0:
FYI review conversations are resolved by reviewers only (and when the contributor picks a code suggestion :| ). Otherwise it's tedious for reviewers to check whether all comments have been addressed.

ronso0:
Thank you!

Can you please take care of the documentation in https://manual.mixxx.org/2.5/chapters/appendix/mixxx_controls.html ?

ronso0:
> In case this shall be adopted for the library later on we will also make do with just one control, even if multiple tracks with locked/unlocked are selected:
> 
>     * all the same lock state: toggle lock
> 
>     * different lock state: lock all, trigger again to unlock all

If you want to work on that, too, note that the control flow is a bit more elaborate:
LibraryControl -> (active view) -> WTrackTableView -> toggleBpmLock
you may also take a loot at the helper [WTrackMenu::getTrackBpmLockStates()](https://github.com/mixxxdj/mixxx/blob/8ccdb98dfe6d7a7acbe082e82ff323a3fe5987f6/src/widget/wtrackmenu.cpp#L715)

FrankwaP:
> Thank you!
> 
> Can you please take care of the documentation in https://manual.mixxx.org/2.5/chapters/appendix/mixxx_controls.html ?

Please see: https://github.com/mixxxdj/manual/pull/637

FrankwaP:
>  In case this shall be adopted for the library later on we will also make do with just one control…

I'm not sure I understand: how is it supposed to be different from the command which toggles lock/unlock when clicking the lock icon in the track field?

BTW I've just discover the "auto unlock" when one selects a locked track with no BPM. This feels super weird, like it's a bug… why not preventing from locking in the first place?

ronso0:
> I'm not sure I understand: how is it supposed to be different from the command which toggles lock/unlock when clicking the lock icon in the track field?

It would toggle lock for all selected tracks at once.
That control probably doesn't make much sense for the 'controller only' workflow (selecting multiple tracks still requires pressing Shift on the keyboard), but --considering the track management controls that were requested/implemented recently-- it might make sense for 'mouse + keyboard' and 'controller + keyboard' workflows: Tracks knobs usually send key Up/Down controls to the focused widget and holding Shift on the keyboard meanwhile is possible.
(theoretically, need to test this)

src/engine/controls/bpmcontrol.cpp
```diff
@@ -130,6 +130,15 @@ BpmControl::BpmControl(const QString& group,
             this, &BpmControl::slotTapFilter,
             Qt::DirectConnection);
 
+    m_pToggleBpmLock = std::make_unique<ControlPushButton>(
+            ConfigKey(group, "bpm_toggle_lock"), false);
+    m_pToggleBpmLock->setKbdRepeatable(true);
```
ronso0, line 135: Meeh, overlooked this one, too.
It's a toggle, makes no sense to allow repeat, might even cause issues if the mapped key (keyboard) is held down accidentally.

Will push a fix directly to main.

daschuer:
@FrankwaP Thank you for your contributions. I would like to add you to the contributor list at the Mixxx about box. Shall I use your full name or a nick name?

FrankwaP:
Thank you for taking time to explain the logic for the multi-track toggle :-) 
I will look into it next week. Have a great day!

Issue #12580, opened 2024-01-15
Skin; crash on adding tracks while using a custom parallel_waveforms.xml


mxmilkiib:
### Bug Description

I have a custom Deere [parallel_waveforms.xml](https://gist.github.com/mxmilkiib/69f8c03ae358cd7a1938f243210f569a) that hides the waveforms of empty decks.

When I drag a track to deck 1, and then drag a track to deck 3 whilst the waveform for deck 1 is still being rendered out, it crashes.

trace;
```
[New Thread 0x7fff11ffb6c0 (LWP 195053)]
[New Thread 0x7fff117fa6c0 (LWP 195054)]
warning [Main] QPainter::begin: Paint device returned engine == 0, type: 3
warning [Main] QPainter::setRenderHint: Painter must be active to set rendering hints
warning [Main] QPainter::setMatrixEnabled: Painter not active
warning [Main] QPainter::setPen: Painter not active
warning [Main] QPainter::setOpacity: Painter not active
warning [Main] QPainter::setOpacity: Painter not active
warning [Main] QPainter::setPen: Painter not active
warning [Main] QPainter::fillPath: Painter not active
warning [Main] QPainter::setPen: Painter not active
warning [Main] QPainter::setOpacity: Painter not active
warning [Main] QPainter::setPen: Painter not active
warning [Main] QPainter::fillPath: Painter not active
warning [Main] QPainter::end: Painter not active, aborted
warning [Main] QPainter::begin: Paint device returned engine == 0, type: 3
warning [Main] QPainter::setMatrixEnabled: Painter not active
warning [Main] QPainter::setPen: Painter not active
warning [Main] QPainter::setRenderHint: Painter must be active to set rendering hints
warning [Main] QPainter::fillPath: Painter not active
warning [Main] QPainter::drawPath: Painter not active
warning [Main] QPainter::end: Painter not active, aborted

Thread 1 "mixxx" received signal SIGSEGV, Segmentation fault.
QOpenGLTexture::width (this=0x0) at /usr/src/debug/qt6-base/qtbase-everywhere-src-6.6.1/src/opengl/qopengltexture.cpp:2958
Downloading source file /usr/src/debug/qt6-base/qtbase-everywhere-src-6.6.1/src/opengl/qopengltexture.cpp
```

backtrace;
```
(gdb) backtrace
[...]

PR #13043, opened 2024-04-03
Fix null pointer access and avoid nested allocations with QOpenGLTexture

daschuer:
This adds a missing null check in https://github.com/mixxxdj/mixxx/commit/ec4bf17c3b0cdf6167c55851ef7f4c54b1283f11 to fix https://github.com/mixxxdj/mixxx/issues/12580

During testing I have noticed that QOpenGLTexture itself is only a smart pointer to the texture storage. This can also potentially be null. So I refactored the code to avoid a second allocation to store a pointer to QOpenGLTexture.

The last commit is only refactoring. Can this also be integrated in 2.4 or should I split it out for main?


m0dB:
I don't see the problem with using a unique_ptr, but I do not oppose this change. It can be integrated in both 2.4 and main IMO. Thanks for taking care of this while I was away!

m0dB (approved):

src/waveform/renderers/allshader/waveformrendererpreroll.cpp
```diff
@@ -6,13 +6,13 @@
...
 
 #include "skin/legacy/skincontext.h"
-#include "util/texture.h"
 #include "waveform/renderers/allshader/matrixforwidgetgeometry.h"
 #include "waveform/renderers/waveformwidgetrenderer.h"
 #include "widget/wskincolor.h"
 
 namespace {
-std::unique_ptr<QOpenGLTexture> generateTexture(float markerLength,
+void generateTexture(OpenGLTexture2D* pTexture,
```
Swiftb0y, line 14: I really don't see how the tradeoff in code quality makes sense. There is no reason to introduce an out-parameter here IMO. If the extra allocation bothers you, why not return it by value directly?

daschuer: This is this no out parameter which may be considered as bad code quality these days. It just updates the texture via pointer. I kept just to avoid moving code around. I will now make it a normal member function as usual and rename it to avoid future confusion.

daschuer: Refactored to become drawPrerollImage() returning the temporary QImage.

JoergAtGithub:
I tested this on Windows 11 and it works stable. The review comment by Swiftb0y is addressed. Code LGTM! Thank you for taking care!

PR #13044, opened 2024-04-03
Add simple support for STEM files

acolombier:
This feature allows loading STEM file, extract the stem data and ensure the correctness of the file.

The SoundSource will provide the main mix (the first stream by convention) to the audio engine and the stem info data to the track object.

This PR depends on #13042


daschuer:
Can you rebase this on main, to get rid of the first commit?

src/sources/soundsource.cpp
```diff
@@ -32,7 +32,7 @@ QString SoundSource::getTypeFromUrl(const QUrl& url) {
 
 //static
 QString SoundSource::getTypeFromFile(const QFileInfo& fileInfo) {
-    const QString fileSuffix = fileInfo.suffix().toLower().trimmed();
+    const QString fileSuffix = fileInfo.completeSuffix().toLower().trimmed();
```
JoergAtGithub, line 35: This would change the file type recognition for files names like `artistname.songtitle.flac`
I'm not sure if this case is relevant, but `fileinfo.filename().toLower().trimmed().endsWith(".stem.mp4")` below would be the safe way to implement this.

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,139 @@
...
+/// @param reader The IODevice to search MP4 atom in
+/// @param path The list of atom to traverse in the tree, from top to bottom.
+/// Must be null terminated
+/// @param box_size The size of the box currently under the cursor to focus the
+/// search in
+/// @param pathIdx The level of the tree currently search (index of path)
+/// @return the size of the box data if found, -1 otherwise
+uint32_t seekTillAtom(QIODevice& reader,
+        const uint32_t path[],
+        uint32_t box_size = 0,
```
JoergAtGithub, line 41:
```suggestion
        uint32_t boxSize = 0,
```

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,139 @@
...
+        reader.skip(ATOM_TYPE(buffer) - kAtomHeaderSize);
+    }
+    return -1;
+}
+
+} // anonymous namespace
+
+// static
+bool StemInfoImporter::isStemFile(
+        const QString& aFileName) {
```
JoergAtGithub, line 64:
```suggestion
        const QString& fileName) {
```

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,139 @@
...
+        if (!color.isValid() || name.isEmpty()) {
+            kLogger.debug() << "Unexpected or missing stem name or attribute in STEM manifest";
+            return {};
+        }
+        stemsList.emplace_back(name, color);
+    }
+
+    // Extract DSP information
+    // TODO(XXX) DSP only contains limit and a compressor, which aren't
+    // supported by Mixxx yet. parse and implement when supported
```
JoergAtGithub, line 132:
```suggestion
    // used by Mixxx yet. Parse and implement when supported
```

src/track/steminfoimporter.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "track/steminfo.h"
+
+namespace mixxx {
+
+/// Importer class for StemInfo objects that can correct timing offsets when the
+/// signal info (channel number, sample rate, bitrate) is known.
```
JoergAtGithub, line 8: This comment seems not to fit here.

src/sources/soundsourcestem.h
```diff
@@ -0,0 +1,69 @@
...
+#include "sources/soundsourceprovider.h"
+#include "util/samplebuffer.h"
+
+namespace mixxx {
+
+/// @brief Handle a single stem stem embedded in a stem file
+class SoundSourceSingleSTEM : public SoundSourceFFmpeg {
+  public:
+    // streamIdx is the FFmpeg stream id, which may different than stemIdx + 1
+    // because STEm may contain other non audio stream
```
JoergAtGithub, line 13:
```suggestion
    // because STEM may contain other non audio stream
```

JoergAtGithub (commented):
Added some minor remarks

src/track/steminfoimporter.cpp
acolombier: I'm afraid this won't mean what I tried to say here. I have reworded it differently now, let me know if that makes more sense.

JoergAtGithub: Comment is clear now!

CMakeLists.txt
```diff
@@ -3470,6 +3444,51 @@ if(FFMPEG)
...
+  target_compile_definitions(mixxx-lib PUBLIC __STEM__)
+  target_compile_definitions(mixxx-test PUBLIC __STEM__)
+  target_sources(mixxx-test PUBLIC
+    src/test/stemtest.cpp
+    src/test/steminfotest.cpp
+  )
+  target_sources(mixxx-lib PRIVATE
+    src/sources/soundsourcestem.cpp
+    src/track/steminfoimporter.cpp
+    src/track/steminfo.cpp # TODO precompiled header?
```
JoergAtGithub, line 3462: Yes, please add track/steminfo.h to the list of precompiled headers. In the follow-up PRs it's included at various places an makes therefore sense.

src/sources/soundsourcestem.cpp
```diff
@@ -0,0 +1,468 @@
...
+        // TODO(XXX): currently, stem samples are interleaved and packed next to each other as such:
+        //    1L1R1L1R1L1R...2L2R2L2R2L2R2L2R......3L3R3L3R3L3R3L3R......4L4R4L4R4L4R4L4R....
+        //    Can FFmpeg decode as without having to use a decoder per channel?
+        //    1LLLLLLLLLLLLLL....1RRRRRRRRR...2LLLLLLL...?
+
+        // Change the sample layout to interleave all channels together
+        for (SINT i = 0; i < stemSampleLength / 2; i++) {
+            pBuffer[2 * stemCount * i + 2 * streamIdx] = m_buffer[2 * i];
+            pBuffer[2 * stemCount * i + 2 * streamIdx + 1] = m_buffer[2 * i + 1];
+        }
```
JoergAtGithub, line 462: I assume that this memory structure is to be optimized across all modules in follow-up PR?

acolombier: I guess by optimisation, you mean have FFmpeg take care of the layout when decoding? I don't have any immediate plan. I had a quick look and I didn't find any simple solutions for this as each stream is technically independent, handled by a dedicated FFmpeg context. Do you have any suggestion or tips on how to do so? 
FYI, I took some stats to see if that for-loop/re-layout was a bottleneck, and it is about 1/100th of the time used by RubberBand so probably not a problem yet.

acolombier:
I have added the support for `.stem.m4a` extension as well. Note that it isn't strictly mentioned in the spec (or maybe I missed it?) but it appears that Traktor supports it as well, and so project like [stemgen](https://github.com/axeldelafosse/stemgen) seems to generate this extension by default.

src/sources/soundsourcestem.cpp
JoergAtGithub: The rubberband performance is a huge problem - was already before, but stems multiplies this. On my rather fast Windows laptop, I can only use a singles stems deck with Rubberband R2 and R3 becomes unusable with stems.
I could imagine, that one major reason for this is, that the data are not alligned in the memory, to allow the compiler to use efficient SIMD instructions. Maybe this can only be changed in Rubberband itself, but maybe already the memory alignment in our engine might influence this.
Therefore I think, that analysis and overall optimization of the engines memory structures might be necessary.

JoergAtGithub (approved):

src/sources/soundsourcestem.cpp
acolombier: > but maybe already the memory alignment in our engine might influence this

This part is already aligned, but I assume you me the [channel buffer](https://github.com/acolombier/mixxx/blob/b8c19a0c2ba6b77971d2e15cf84faa258610e018/src/engine/bufferscalers/enginebufferscalerubberband.cpp#L166)

Are we sure RubberBand would use SIMD? If yes, I can try to scrape the individual `SampleBuffer` and use a a contiguous mem alloc, instead of individual ones

acolombier: (Note that I had plenty of `ScopedTimer` all across which allow me to confirm that 99% of the CPU time is due to RubberBand::retrieve, which I removed now)

acolombier: I did the experiment with a contiguous global buffer and there was no changes.
Just FYI, I've put back the `ScopedTimer`  and here is the result
```
Stat("EngineBufferScaleRubberBand::scaleBuffer","count=462,sum=2.82884e+09ns,average=6.12302e+06ns,min=2.77938e+06ns,max=1.97878e+07ns,variance=3.76359e+12ns^2,stddev=1.94e+06ns")
Stat("EngineDeck::processStem","count=535,sum=2.88642e+09ns,average=5.39517e+06ns,min=89170ns,max=2.01299e+07ns,variance=7.62824e+12ns^2,stddev=2.76193e+06ns")
Stat("RubberBand::process","count=1839,sum=2.81529e+09ns,average=1.53088e+06ns,min=7860ns,max=4.10194e+06ns,variance=8.39907e+10ns^2,stddev=289812ns")
Stat("RubberBand::retrieve","count=2299,sum=2.65024e+06ns,average=1152.78ns,min=105ns,max=16231ns,variance=1.78589e+06ns^2,stddev=1336.37ns")
```
(Just the sum)
```
EngineDeck::processStem                  2.88642e+09ns
EngineBufferScaleRubberBand::scaleBuffer 2.82884e+09ns
RubberBand::process                      2.81529e+09ns
RubberBand::retrieve                     2.65024e+06ns
```
It was `RubberBand::process` btw, not `RubberBand::retrieve`

I guess another experiment would be to aligned the buffer size with the [sample count](https://github.com/mixxxdj/mixxx/blob/b8c19a0c2ba6b77971d2e15cf84faa258610e018/src/engine/bufferscalers/enginebufferscalerubberband.cpp#L261), I'll try that tomorrow.

JoergAtGithub: Great, that you did the analysis already! It would be even greater, to store such results in a document.
A helpul tool to understand such performance issues is callgrind. I used it once (for a much smaller single thread project), and could clearly see, where the CPU cycles were burned. And I could see, where the compiler picked SIMD implementations of std library functions, and where not.

src/sources/soundsource.cpp
```diff
@@ -52,6 +53,18 @@ QString SoundSource::getTypeFromFile(const QFileInfo& fileInfo) {
...
+        // meant to fall back to MP4 decoder, in case a player doesn't have the
+        // capability to extract the different stem track. However, we need to
+        // detect that early as the file natively looks like a traditional MP4
+        // with multiple tracks (audio and video)
+        return QLatin1String("stem.mp4");
+    }
+    if (normalisedFilename.endsWith(QLatin1String(".stem.m4a"))) {
+        // STEM suffix may also extends the ".m4a" suffix for the same reason as above
+        return QLatin1String("stem.m4a");
+    }
```
daschuer, line 67: We normally prioritize the mime type over the file extension. I think mp4 and m4a both have "audio/mp4".
So I think we need to look up the mime type first and than check if the file name is "*stem.*". This will finally allow any random file extension which has a stem file name. Or we may be more strict and limit it to "stem.m4a" and "stem.mp4".
Optimal would be a magic number to look up. But probably not worth the effort.

src/sources/soundsourceproxy.cpp
```diff
@@ -177,6 +180,13 @@ QList<QMimeType> mimeTypesForFileType(const QString& fileType) {
             }
         }
     }
+    if (fileType == "stem.mp4" || fileType == "stem.m4a") {
+        // *.stem.mp4 and *.stem.m4a suffix do not have a specific MIME type associated
+        // with them, and simply fall back to MP4. To prevent conflicts with the
+        // MP4 decoder able to decode arbitrary MP4 file (such as video,
+        // extracting just the audio track), we don't return any MIME here
```
daschuer, line 187: I think they have. At least libmagic (via Qt) will return one, I guess

src/sources/soundsourcestem.cpp
```diff
@@ -0,0 +1,468 @@
...
+constexpr int kRequiredStreamCount = kNumStreams - 1; // Stem count doesn't include the main mix
+
+const Logger kLogger("SoundSourceSTEM");
+
+} // anonymous namespace
+
+const QString SoundSourceProviderSTEM::kDisplayName = QStringLiteral("STEM with FFmpeg");
+
+QStringList SoundSourceProviderSTEM::getSupportedFileTypes() const {
+    return {"stem.mp4", "stem.m4a"};
```
daschuer, line 37: Is one with video? Or can we normalize them to only one.

src/sources/soundsourcestem.cpp
```diff
@@ -0,0 +1,468 @@
...
+
+QString SoundSourceProviderSTEM::getVersionString() const {
+    return QString::fromUtf8(av_version_info());
+}
+
+SoundSourceSingleSTEM::SoundSourceSingleSTEM(const QUrl& url, unsigned int streamIdx)
+        : SoundSourceFFmpeg(url), m_streamIdx(streamIdx) {
+}
+
+SoundSource::OpenResult SoundSourceSingleSTEM::tryOpen(
```
daschuer, line 54: Is there an early check if this is really a stem file? If not, it should fail in the first attempt. Maybe also later and the normal m4a decoder can take over.

daschuer (changes requested):
This is the first batch of comments.

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+        }
+        reader.skip(ATOM_TYPE(buffer) - kAtomHeaderSize);
+    }
+    return -1;
+}
+
+} // anonymous namespace
+
+// static
+bool StemInfoImporter::isStemFile(
```
daschuer, line 63: Here we need apply the mime type logic. And probably the checking for easy "stem" properties in addition.
Alternative  this may look for these properties only and we can call the function isMp4AStem() or such.  
This raises the general question if this Class can be only used with MP4s or if we need to protect every invocation to be called with other files.

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+/// @brief Seek the reader the atom requested.
+/// @param reader The IODevice to search MP4 atom in
+/// @param path The list of atom to traverse in the tree, from top to bottom.
+/// Must be null terminated
+/// @param boxSize The size of the box currently under the cursor to focus the
+/// search in
+/// @param pathIdx The level of the tree currently search (index of path)
+/// @return the size of the box data if found, -1 otherwise
+uint32_t seekTillAtom(QIODevice& reader,
```
daschuer, line 39: We avoid non const references in Mixxx. Please pass the reader as a pointer.

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+        kLogger.warning()
+                << "Failed to extract the manifest data"
+                << filePath;
+        return {};
+    };
+
+    auto json = jsonData.object();
+
+    // Check the manifest version
+    auto version = json.value("version").toInteger(-1);
```
daschuer, line 103: Can you add here a typical JSON as a source code comment as a reference? 
Maybe also a link to the specification.

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+        if (!stemRef.isObject()) {
+            kLogger.debug() << "Unexpected or missing stems value in STEM manifest";
+            return {};
+        }
+        auto stem = stemRef.toObject();
+        auto color = QColor(stem.value("color").toString());
+        auto name = stem.value("name").toString();
+        if (!color.isValid() || name.isEmpty()) {
+            kLogger.debug() << "Unexpected or missing stem name or attribute in STEM manifest";
+            return {};
```
daschuer, line 130: I think we should be less critical and pick reasonable defaults. A stem file should be play-able, with a minimum of information that is required to play it.

src/sources/soundsource.cpp
daschuer: At which offset are the atoms located? Is the lookup fast? Maybe we can do this.

daschuer (changes requested):

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+const mixxx::Logger kLogger("StemInfoImporter");
+constexpr int kSupportedStemVersion = 1;
+const QStringList kStemFileExtensions = {".stem.mp4", ".stem.m4a"};
+
+const uint32_t kAtomHeaderSize = 8; // 4 bytes (unsigned integer) + 4 char
+const uint32_t kStemManifestAtomPath[] = {
+        ATOM_TYPE("moov"),
+        ATOM_TYPE("udta"),
+        ATOM_TYPE("stem"),
+        0 // This indicate end of the path
```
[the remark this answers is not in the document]
daschuer: What happens if moov and udta are missing? I think it would be more robust to just allow any atom that is before stem.

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    uint32_t byteRead = 0;
+    char buffer[kAtomHeaderSize];
+    while (!reader.atEnd() || (boxSize && byteRead >= boxSize)) {
+        reader.read(buffer, kAtomHeaderSize);
+        byteRead += boxSize;
+        if (ATOM_TYPE(buffer + 4) == path[pathIdx]) {
+            return seekTillAtom(reader, path, ATOM_TYPE(buffer) - kAtomHeaderSize, pathIdx + 1);
+        }
+        reader.skip(ATOM_TYPE(buffer) - kAtomHeaderSize);
```
daschuer, line 55: ATOM_TYPE should not be used to also return the size. 
It would be more easy to read to cast to a struct that contains type and size.

src/track/taglib/trackmetadata_file.cpp
```diff
@@ -48,12 +48,16 @@ namespace taglib {
 // Deduce the TagLib file type from the file name
 FileType getFileTypeFromFileName(
         const QString& fileName) {
-    DEBUG_ASSERT(!fileName.isEmpty());
-    const QString fileSuffix(fileName.section(QChar('.'), -1).toLower().trimmed());
+    QString normalisedFileName = fileName.toLower().trimmed();
```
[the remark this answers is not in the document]
JoergAtGithub: There is no dedicated mime-type assigned for the STEM format: https://www.iana.org/assignments/media-types/media-types.xhtml#audio
There are only two options, either using the file extension, or reading the content of the file and analysing it (which is time consumig).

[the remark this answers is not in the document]
daschuer: In this case this is a taglib function only used to check for mp3. There should be no need to deal with stems here.

src/track/steminfoimporter.cpp
acolombier: > Here we need apply the mime type logic

AFAIK, mime type isn't available in the Track object. Reading the file (to detect the MIME using `QMimeDatabase::mimeTypeForFile` will likely have effect on performance. of every single file being imported, which is likely not negligible when importing a large library. 

> This raises the general question if this Class can be only used with MP4s

I'm not sure why we would want to support another file type. The STEM atom is part of the NI standard, meaning it needs to be in a mp4 file. If we want to implement other STEM standards, there is no guarantee that there will use a STEM atom, as part of the MPEG standard. My suggestion would be to keep the implantation to what we know and support currently (NI standard, mp4) and cross the bridge to support other standards when we decide to do so.

src/track/steminfoimporter.cpp
acolombier: "Stem #<n>" and Red/Green/Blue/Yellow?

src/track/taglib/trackmetadata_file.cpp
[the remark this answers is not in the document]
acolombier: > this is a taglib function only used to check for mp3

I'm not sure this is. This function is used by TagLib to decide which taglib reader shoud be used to extract metadata.

src/track/taglib/trackmetadata_file.cpp
```diff
@@ -48,12 +48,16 @@ namespace taglib {
...
+    DEBUG_ASSERT(!normalisedFileName.isEmpty());
+    const QString fileSuffix(normalisedFileName.section(QChar('.'), -1));
     if (QStringLiteral("mp3") == fileSuffix) {
         return FileType::MP3;
     }
-    if ((QStringLiteral("m4a") == fileSuffix) || (QStringLiteral("m4v") == fileSuffix)) {
+    if ((QStringLiteral("m4a") == fileSuffix) ||
+            (QStringLiteral("m4v") == fileSuffix) ||
+            normalisedFileName.endsWith("stem.mp4") ||
+            normalisedFileName.endsWith("stem.m4a")) {
```
acolombier, line 60:
```suggestion
    if ((QStringLiteral("m4a") == fileSuffix) ||
            (QStringLiteral("m4v") == fileSuffix) ||
            normalisedFileName.endsWith("stem.mp4")) {
```

Or do we want to support `mp4` entirely?

src/track/track.h
```diff
@@ -495,6 +511,17 @@ class Track : public QObject {
     /// caller guards this a lock.
     bool importPendingCueInfosWhileLocked();
 
+#ifdef __STEM__
+    /// Sets stem info and returns a boolean to indicate if stems were updated.
+    /// Only supposed to be called while the caller guards this a lock.
+    bool setStemPointsWhileLocked(const QList<StemInfo>& stemInfo);
```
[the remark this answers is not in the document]
acolombier: Probably a mistake when fixing typos, will rename to `StemInfo`

src/sources/soundsource.cpp
daschuer: Cool, than let's do it. 
Is it correct to do it for all audio/mp4 mime types?

src/sources/soundsourcestem.cpp
acolombier: NI suggest using `stem.mp4` and provide stem samples with the `stem.mp4`, however a few tools online (such as stemgen) generate `stem.m4a` files. Traktor supports both `mp4` and `m4a` extension.

src/sources/soundsourcestem.cpp
acolombier: There is a check that ensure the audio tracks have matching spec. Is there other checks you would like to see?

src/sources/soundsourceproxy.cpp
daschuer: Ok, so please adjust the comments accordingly. 
Maybe this will receive some refactoring when implementing the stem lookup. 

Another (crazy) thought:
I can imagine to put two or more aac streams into an MP4. How to deal with them, play only the first stream like we do now or treat them like stem files?

src/track/steminfoimporter.cpp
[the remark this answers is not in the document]
daschuer: Ah, so I have probably misread the code. 
Do the atoms build a folder structure? 
In that, case everything is fine.

src/track/taglib/trackmetadata_file.cpp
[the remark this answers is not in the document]
daschuer: Ok, so we have found an issue with our mp3 with wav case, or for files without extension?

src/track/steminfoimporter.cpp
dszakallas: I think that that detection based on `stem.mp4` extension is overly restrictive. At the extreme any iso media container could be feature tested for `moov.udta.stem`, but at the least we should be able to detect stems in arbitrary `.mp4` containers.

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+        return boxSize;
+    }
+
+    uint32_t byteRead = 0;
+    char buffer[kAtomHeaderSize];
+    while (!reader.atEnd() || (boxSize && byteRead >= boxSize)) {
+        reader.read(buffer, kAtomHeaderSize);
+        byteRead += boxSize;
+        if (ATOM_TYPE(buffer + 4) == path[pathIdx]) {
+            return seekTillAtom(reader, path, ATOM_TYPE(buffer) - kAtomHeaderSize, pathIdx + 1);
```
dszakallas, line 53: There's also the possibility of 64 bit box; in case `atomsize==1` there's an additional [8 bytes encoding the size](https://github.com/DigiDNA/ISOBMFF/blob/main/ISOBMFF/source/ContainerBox.cpp#L101).

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+                << "Failed to extract the manifest data"
+                << filePath;
+        return {};
+    };
+
+    auto json = jsonData.object();
+
+    // Check the manifest version
+    auto version = json.value("version").toInteger(-1);
+    if (version <= 0) {
```
dszakallas, line 104: I'd rather simplify all this by checking for `version==1` and bailing out on anything else.

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,144 @@
...
+        if (!color.isValid() || name.isEmpty()) {
+            kLogger.debug() << "Unexpected or missing stem name or attribute in STEM manifest";
+            return {};
+        }
+        stemsList.emplace_back(name, color);
+    }
+
+    // Extract DSP information
+    // TODO(XXX) DSP only contains a limiter and a compressor effect, which
+    // Mixxx doesn't have yet. Parse and implement when supported
```
dszakallas, line 137: nit: maybe worth printing a warning if we detected enabled compressor/ limiter the fact that it will be ignored.

src/sources/soundsourcestem.cpp
daschuer: OK

src/sources/soundsourceproxy.cpp
daschuer: OK, sounds reasonable.

src/sources/soundsourcestem.cpp
daschuer: I have no specific demand. Just keep in mind that this tryOpen() is called according to the priority of a certain format.
In other sound sources there is a Strict and a Permissive mode. I have no idea if this here reasonable as well.

src/track/steminfoimporter.cpp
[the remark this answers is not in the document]
acolombier: Yes, they are organise as a tree structure, with multiple root (`moov` being one of them)

src/track/steminfoimporter.cpp
acolombier: Again, I was simply implementing the specification (which I believe @dszakallas also had access to). Happy to go beyond what it defines.

src/track/steminfoimporter.cpp
acolombier: Not entirely pure, but some red, green and blue are used by default for waveform as mentioned in my [3-band waveform PR](https://github.com/mixxxdj/mixxx/pull/13075)

src/track/steminfoimporter.cpp
acolombier: Do you mean a modal popup? If you mean log messages, I'm not sure what would be the value of it, but happy to put a debug a message to reduce noise.

src/track/taglib/trackmetadata_file.cpp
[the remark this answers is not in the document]
acolombier: Yes, but this should only be impacting metadata tho, you should still be able to load and read the file.

src/track/steminfoimporter.cpp
JoergAtGithub: We should not expand this PR by general MP4 support. Scope of this PR is the STEMs format and nothing beyond!
Traktor Pro seems to have another detection logic than just checking for .stems.mp4 extension, because it seems to work also with the extension .stems.m4a (or they just check for one of both file extensions).
Performance for file type detection is important in my opinion. What would it cost to probe for moov.udta.stem ?

src/track/steminfoimporter.cpp
JoergAtGithub: Let's use the ones from stemgen, as the practical use case is stem splitting with stemgen and than use the generated files in Mixxx.

src/track/steminfoimporter.cpp
JoergAtGithub: This information would not be helpful for a normal user. A warning in the log should be enough.

acolombier: I'll put a debug message

CMakeLists.txt
```diff
@@ -3577,8 +3596,8 @@ if(OPUS)
     src/encoder/encoderopussettings.cpp
   )
   target_compile_definitions(mixxx-lib PUBLIC __OPUS__)
-  target_link_libraries(mixxx-lib PRIVATE OpusFile::OpusFile)
-  target_link_libraries(mixxx-lib PRIVATE Opus::Opus)
+  target_link_libraries(mixxx-lib PRIVATE OpusFile::OpusFile Opus::Opus)
```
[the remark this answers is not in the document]
acolombier: It only mean that the tests won't include Opus library, and test case will be omitted (eg soundsourceproxy). So while it impacts 2.4, it probably also have no impact. What's your suggestion?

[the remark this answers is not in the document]
daschuer: Ok, I understand. So let's keep it as it is.

src/track/steminfoimporter.cpp
dszakallas: We already support mp4 some way, as I was able to drop mp4 tracks into mixxx and play them. And the specification says that the container should be an ISOBMFF with `.mp4` as extension. `.stem.mp4` is "preferred", not required. So, if we want to follow the spec we should try to detect stems in any mp4.

src/sources/soundsourcestem.cpp
dszakallas: wouldn't it be possible to try to create a stem soundsource for any mp4, and if we don't find the metadata, or we face other validation error, continue with the generic mp4 reader?

src/sources/soundsource.cpp
acolombier: Sounds good to me.

src/sources/soundsourcestem.cpp
acolombier: We could do that but that feels wrong and biased to assume that MP4 == STEM IMO. What is the the team opinion?

src/track/steminfoimporter.cpp
acolombier: > What would it cost to probe for moov.udta.stem ?

That's hard to assess as the atom offset isn't static and require to traverse boxes in the file. 

> extension. .stem.mp4 is "preferred", not required

That's true but in the meantime, Mixxx aims to be vendor-agonistic IMO, so the flexibility that Tracktor Pro gives to arrange NI shouldn't drive Mixxx implementation. Anyway, last commit adds support for `.mp4`, but happy to remove depending where the consensus goes.

src/track/steminfoimporter.cpp
acolombier: I'm confused. In other comments, we aim to make the implementation be permissive and allow flexibility. 
Happy to change the debug at `108` to warning if we want it to be visible with the default log level, but it goes against the other efforts to hard fail here for no reason.

src/track/steminfoimporter.cpp
JoergAtGithub: Maybe we could add the JSON from src/test/stems/test.stem.mp4 as .json file next to it?

src/track/steminfoimporter.cpp
```diff
@@ -1,31 +1,57 @@
...
-    (uint32_t)(((uint8_t)(value)[0] << 24) | ((uint8_t)(value)[1] << 16) | \
-            ((uint8_t)(value)[2] << 8) | (uint8_t)(value)[3])
-
 namespace mixxx {
 
 namespace {
 
 const mixxx::Logger kLogger("StemInfoImporter");
 constexpr int kSupportedStemVersion = 1;
 const QStringList kStemFileExtensions = {".stem.mp4", ".stem.m4a"};
```
JoergAtGithub, line 21: As far as I learned, there are three cases:

- Anything that ends with `.mp4` is allowed
- `.stem.mp4` is recommended
- `.stem.m4a` is not specified, but is used in practice (Stemgen writes it and Traktor Pro reads it)

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,191 @@
...
+const QStringList kStemMimes = {"audio/mp4", "audio/m4a", "audio/x-m4a", "video/mp4"};
+
+struct MP4BoxHeader {
+    quint32 size;
+    char type[4];
+};
+
+constexpr uint32_t kAtomHeaderSize = sizeof(MP4BoxHeader);
+
+/// @brief Deserialise a MP$ atom/box header
```
JoergAtGithub, line 31:
```suggestion
/// @brief Deserialise a MP4 atom/box header
```

src/sources/soundsource.cpp
```diff
@@ -52,9 +53,18 @@ QString SoundSource::getTypeFromFile(const QFileInfo& fileInfo) {
         // https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/mimetype.20sometimes.20wrong
         return fileSuffix;
     }
-
     QMimeType mimeType = QMimeDatabase().mimeTypeForFile(
             fileInfo, QMimeDatabase::MatchContent);
+    if (
+            // STEM files will be detected as normal MP4 se we check if the file
```
JoergAtGithub, line 59:
```suggestion
            // STEM files will be detected as normal MP4, so we check if the file
```

JoergAtGithub (changes requested):

daschuer:
Just did a test: 
* Remove suffix form an mp3 file. 
* Library scanner ignores it 
* Can't be dropped to decks. 

* Changed suffix form mp3 to wav.
* Library scanner finds the file type is correctly shown as mp3 
* Metadata is not imported 
* Track can be played 
* Track can be dropped to the deck

src/track/steminfoimporter.cpp
acolombier: Yes, that's correct and is already how it work. STEM extension are only defines here as fallback here in case the MIME detection fails from file content. Should I put a comment to explicitly mention that these extension are only used as fallback?

JoergAtGithub: Yes, a commet would help here!

acolombier:
Thanks for these tests @daschuer 

I can confirm that this is now the same behaviour with STEM files:

With `wav`

- Changed suffix form stem.mp4 to wav.
- Library scanner finds the file type is correctly shown as stem.mp4
- Metadata is not imported
- Track can be played
- Track can be dropped to the deck

With `m4a`

- Changed suffix form stem.mp4 to m4a.
- Library scanner finds the file type is correctly shown as stem.mp4
- Metadata **is** imported
- Track can be played
- Track can be dropped to the deck

CMakeLists.txt
JoergAtGithub: Please add the header to the precompiled header list

acolombier:
Note that, as faired, I have noticed a performance impact on probing the STEM box in each MP4 file, especially when running the tests. This is slightly worrying as I have a pretty fast CPU and NVME, so I assume that performance impact will be noticeable.

JoergAtGithub:
> Note that, as faired, I have noticed a performance impact on probing the STEM box in each MP4 file, especially when running the tests. This is slightly worrying as I have a pretty fast CPU and NVME, so I assume that performance impact will be noticeable.

Does it only consumes that much time, if it is a STEM file, or also for normal MP4 audio files?

src/track/steminfoimporter.cpp
dszakallas: this is mostly theoretical as there was no update to the this standard since released nearly 10 years ago, but since this is a single number, an increment likely means a breaking change. so I think it would be cleaner to outright reject something that we have no guarantee of parsing correctly

CMakeLists.txt
acolombier: I haven't figured out how to make that conditionally yet and will add once I have a solution. Happy to do right away if you have a recommended way.

src/track/steminfoimporter.cpp
acolombier: > I think it would be cleaner to outright reject something that we have no guarantee of parsing correctly

That statement can be applied on many of the other requested changes. Again, I think it's best to keep and trying best effort on continuing loading. There are more tests coming later to validate the data (channel layout, matching sample rate, ...) so if this is indeed a breaking change, then it will abort later one, when it gets indeed stuck. At that stage, it's best to carry on IMO. 
As a middle ground, I will update the subsequent `debug` log to a `warning` so it gets displayed by default.

acolombier:
Hard to say - I suspect something is wrong in the way the file type is being detected, leading to the `SoundSourceProxy` test to timeout. I will do further investigation, and come back with further info.

acolombier:
After running the `SoundSourceProxyTest` through callgrind, it looks like the probing of the stem atom was too greedy and could stop recursion early on. There is still a slight performance impact but that should be acceptable now.

src/track/taglib/trackmetadata_file.cpp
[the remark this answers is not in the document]
acolombier: @daschuer do you me to create an issue some taglib also relies on file MIME instead of extension?

src/track/track.cpp
```diff
@@ -1296,6 +1296,21 @@ bool Track::importPendingCueInfosWhileLocked() {
     return setCuePointsWhileLocked(cuePoints);
 }
 
+#ifdef __STEM__
+bool Track::setStemPointsWhileLocked(const QList<StemInfo>& stemInfos) {
```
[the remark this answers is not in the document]
acolombier: I have updated the setter, but noticed that some others in this file also relies on copy by ref (eg. setCuePointsWhileLocked). Should I updated them?

src/track/steminfoimporter.cpp
acolombier: Done

CMakeLists.txt
acolombier: Looks like `target_precompile_headers` adds headers and don't override the list as I thought. Added now

CMakeLists.txt
```diff
@@ -3470,6 +3444,54 @@ if(FFMPEG)
...
+  endif()
+  target_compile_definitions(mixxx-lib PUBLIC __STEM__)
+  target_compile_definitions(mixxx-test PUBLIC __STEM__)
+  target_sources(mixxx-test PUBLIC
+    src/test/stemtest.cpp
+    src/test/steminfotest.cpp
+  )
+  target_precompile_headers(mixxx-lib PUBLIC
+    src/track/steminfo.h
+  )
```
JoergAtGithub, line 3461: I guess only one `target_precompile_headers` statement per target is possible. Try to add the file as entry after line 1334 encapsulated in `if (STEM)` `endif()`

acolombier: Looks like Windows is still not happy. Not sure why only `soundsource.cpp` has unresolved symbol. Or does MSVC compiler only display the first object missing symbols?

JoergAtGithub (approved):
I compiled this locally on Windows11 and it compiles fast and tests pass. All my change requests are addressed!

src/sources/soundsourcestem.h
```diff
@@ -0,0 +1,69 @@
...
+    OpenResult tryOpen(
+            OpenMode mode,
+            const OpenParams& params) override;
+
+  private:
+    unsigned int m_streamIdx;
+};
+
+/// @brief Handle a stem file, composed of multiple audio channel. Can open in
+/// stereo or in stem (4 x stereo)
```
daschuer, line 26: Can the the user of the sound source decide? Probably not. Can you refine the comment?

We have the issue that pure mp4 files are opened with libfaad which is not offset compatible. This means we will have the issue that before merging this PR *.stem.mp4 have been decoded with a different offset and the analysis is no longer consistent. We need to deal with it somehow. Suggestions?

src/sources/soundsourcestem.cpp
daschuer: We have a probing facility already. 
If the file is a stem.mp4 it is probed here first. 
In case tryOpen() fails, libfaad should take over.

src/sources/soundsourcestem.h
JoergAtGithub: Since we can (in future) load a STEM file as well into a deck (all 4 stems seperated) as into a sampler/preview deck (either mixed stereo OR a selected single stem stereo track), STEM files need always to be analyzed with the same offset, independend if we use the mixed stereo, or the stem tracks.
While non-STEM .mp4 are completly independent.

I think we are safe, if we can safely distinguish STEM files from non-STEM .mp4 files and read STEM files always using FFMPEG.

acolombier: @JoergAtGithub is correct - when you open a SoundSource, you pass the `OpenParams` which allow to request a maximum number of supported channel (since Mixxx uses the StereoProxy in case the channel count is less than requested). This allows decks which must not use STEM for performance or usability reason to use the soundsource as a strict stereo (samples, preview, but also analyser currently, till the PR which introduce STEM analysis)

src/sources/soundsourcestem.h
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include "sources/soundsourceffmpeg.h"
+#include "sources/soundsourceprovider.h"
+#include "util/samplebuffer.h"
+
+namespace mixxx {
+
+/// @brief Handle a single stem stem embedded in a stem file
```
dszakallas, line 9:
```suggestion
/// @brief Handle a single stem embedded in a stem file
```

acolombier:
@daschuer friendly ping in case you missed the notification

src/track/steminfoimporter.cpp
```diff
@@ -0,0 +1,215 @@
...
+    auto masteringDsp = json.value("mastering_dsp").toObject();
+    auto compressor = masteringDsp.value("compressor").toObject();
+    auto limiter = masteringDsp.value("limiter").toObject();
+
+    if (compressor.value("enabled").toBool() || limiter.value("enabled").toBool()) {
+        kLogger.debug() << "The STEM manifest is requesting DSP effects but "
+                           "this isn't supported by Mixxx yet.";
+    }
+
+    file.close();
```
dszakallas, line 210: code style: isn't it a problem that this is not called on earlier exit paths of this func?

acolombier: [QFileDevice](https://doc.qt.io/qt-6/qfiledevice.html#dtor.QFileDevice) (which QFile inherit of) will ensure it closes it if necessary on destruction

dszakallas (approved):

daschuer (approved):
LGTM, Thank you. 

Sorry for the delay. I have underestimated the time for branching out 2.5 compared to my free time this week.

PR #13045, opened 2024-04-03
feat: enable controller setting on MIDI

acolombier:
The original plan was to also support settings when parsing the XML mapping definition, thus why a follow up PR was dedicated for this feature.

As the latter is likely not happening, this PR exposes the settings in the JS engine only.

I don't have a MIDI controller available for testing yet, can anyone give it a go please?

### DIFF
--- a/src/controllers/midi/legacymidicontrollermappingfilehandler.cpp
+++ b/src/controllers/midi/legacymidicontrollermappingfilehandler.cpp
@@ -11,7 +11,6 @@ std::shared_ptr<LegacyControllerMapping>
 LegacyMidiControllerMappingFileHandler::load(const QDomElement& root,
         const QString& filePath,
         const QDir& systemMappingsPath) {
-    // TODO (XXX): support for controller settings
     if (root.isNull()) {
         return nullptr;
     }
@@ -26,6 +25,7 @@ LegacyMidiControllerMappingFileHandler::load(const QDomElement& root,
 
     // Superclass handles parsing <info> tag and script files
     parseMappingInfo(root, pMapping);
+    parseMappingSettings(root, pMapping.get());
     addScriptFilesToMapping(controller, pMapping, systemMappingsPath);
 
     QDomElement control = controller.firstChildElement("controls").firstChildElement("control");


ronso0:
How exactly can we test?

acolombier:
It can be tested by adding settings definition and querying them from the engine like for the [S4 MK3](https://github.com/mixxxdj/mixxx/pull/12995#event-12304815386)

ronso0:
So they **do** go to the xml? (to create the settings GUI)

acolombier:
Yes, that's right. The setting definition still goes in XML. The thing that was dropped is conditional mapping definition in XML

ronso0:
Haha, easy one : )
Will give it a go soonish.

ronso0:
I added the S4 Mk3 snip to a VirMIDI mapping and there seems to be an issue with the checkbox labal and rich text:
the inline style is not recognized, thus the string is looooooong and makes the settings page expand that much I don't get a full view with FHD.

Seems this is not based on main?

ronso0:
Yeah, putting this commit on top of main fixes the string issue.

But the checkbox wrapper has a margin now.
before (broken, pure checkbox):
image


on top of main:
image

ronso0:
Btw I think we should add a 'button' style for the label strings, so that the mapping only contains
`This will light up the :btn:SYNC:/btn:` (or whatever syntax fits)
and c++ creates the style depending on the GUI's dark/light flag (Qt theme).

Any other style templates that would simplify the xml strings?
italic?
bold?

ronso0:
I.e. I consider this ready to go if rebased onto main for easy testing.

Swiftb0y (approved):
Code LGTM ;) 
Didn't do any testing. I may find some time to test this out with some of my mappings.

ronso0:
> Btw I think we should add a 'button' style for the label strings, so that the mapping only contains `This will light up the :btn:SYNC:/btn:` (or whatever syntax fits) 

:hwbtn:`LABEL`?
#13046

acolombier:
Merge?

ronso0:
Sorry, I only checked the GUI so far, not the actual settings (I expect that to work, though someone should check it).

JoergAtGithub:
I will check this on Monday. I need to set up a MIDI device first.

JoergAtGithub (changes requested):
I just tested this with a MIDI device. The settings functionality itself works as expected, but not the Restore Defaults button. It unloads the whole mapping instead of resetting the setting values.

acolombier:
image

Could you share a video of the behaviour? It sound like the expected behaviour if there is [no exact mapping match](https://github.com/acolombier/mixxx/blob/c5e34806eb79d9a49151fc514d429351905e2f02/src/controllers/dlgprefcontroller.cpp#L433)

JoergAtGithub:
https://github.com/mixxxdj/mixxx/assets/64457745/2b6cb6c7-a6cc-4d95-b688-d1270e1cf8c3

acolombier:
This looks like expected behaviour when the device doesn't match 100% the mapping definition. Could you confirm the mapping combobox reset to `No mapping` too?

The custom settings of the current mapping are staged for reset, but when you re-select it in the combobox, you undo the reset (as per intended). If you want to reset the setting, you should:
- Click the `Restore Defaults`
- Click `Ok`/`Apply`

Then reselecting the mapping will use the default values

JoergAtGithub:
> This looks like expected behaviour when the device doesn't match 100% the mapping definition. Could you confirm the mapping combobox reset to `No mapping` too?

Yes, can confirm it's exactly the behavior you described for a device that doesn't match 100% the mapping definition.

But matching can't be true for a MIDI device due to:
https://github.com/mixxxdj/mixxx/blob/435da25fa8f361f94c8abb24aa770857a3dce7e9/src/controllers/midi/midicontroller.cpp#L73-L78

Therefor this PR is correct and the controller ettings code will work as soon as MidiController::matchMapping get implemented. The behavior without MidiController::matchMapping implemented is safe and just a UI glitch.

JoergAtGithub (approved):

JoergAtGithub:
Thank you!

PR #13046, opened 2024-04-04
Controller preferences: add :hwbtn: markup parser

ronso0:
Allows using :hwbtn:`LABEL` in xml and moves the styling to c++
image

It's just bold now (dropped the `tt` monospace font tag).

The only way I figured to get horizontal padding was to wrap the label in `&nbsp;`
<sub>_Padding is not applied with `padding: 2px;` alone. Might be one of those qss quirks where we need to set an unrelated property, though with `span` neither `display: inline-block;` nor `box-sizing: border-box;` or `border` work_</sub>

@acolombier Mind taking a look?

### DIFF
--- a/src/controllers/legacycontrollersettings.cpp
+++ b/src/controllers/legacycontrollersettings.cpp
@@ -11,6 +11,32 @@
 
 #include "moc_legacycontrollersettings.cpp"
 
+namespace {
+
+// Regex that allows to use :hwbtn:`LABELSTRING` in xml controller settings xml
+const QRegularExpression kHwbtnRe(QStringLiteral(":hwbtn:`([^`]*)`"));
+
+const QString kHwbtnStyleWrapper = QStringLiteral(
+        "<span style='"
+        "background: #343131;"
+        "color: #d9d9d9;"
+        "font-size: 70%;"
+        "font-weight: 600;"
+        "line-height: 120%;"
+        "text-transform: uppercase;"
+        // Padding does only work with block and table-cell elements, not
+        // inline span. Hence, wrapping the string in `&nbsp;` is the only way
+        // to get the desired look here. See the Qt documentation for details:
+        // https://doc.qt.io/qt-6/richtext-html-subset.html#css-properties
+        // \\1 is the RegEx match group 1.
+        "'>&nbsp;\\1&nbsp;</span>");
+
+QString replaceMarkupStyleStr(QString str) {
+    return str.replace(kHwbtnRe, kHwbtnStyleWrapper);
+}
+
+} // namespace
+
 LegacyControllerSettingBuilder* LegacyControllerSettingBuilder::instance() {
     static LegacyControllerSettingBuilder* s_self = nullptr;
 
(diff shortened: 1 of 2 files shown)


src/controllers/legacycontrollersettings.cpp
```diff
@@ -11,6 +11,28 @@
 
 #include "moc_legacycontrollersettings.cpp"
 
+namespace {
+
+// Regex that allows to use :hwbtn:`LABELSTRING` in xml controller settings xml
+const QRegularExpression kHwbtnRe(QStringLiteral(":hwbtn:`([^`]*)`"));
+
+const QString kHwbtnStyleWrapper() {
+    return QStringLiteral(
```
Swiftb0y, line 20: could this just be a 
```suggestion
const QString kHwbtnStyleWrapper = QStringLiteral(
```
?

src/controllers/legacycontrollersettings.cpp
```diff
@@ -11,6 +11,28 @@
...
+            "padding: 2px; "
+            "font-weight: 700;'>"
+            "\\1</span>");
+}
+
+QString replaceMarkupStyleStr(const QString& rawStr) {
+    QString str = rawStr;
+    str.replace(kHwbtnRe, kHwbtnStyleWrapper());
+    return str;
+}
```
Swiftb0y, line 32: at that point, just take by value. also returning directly looks nicer.
```suggestion
QString replaceMarkupStyleStr(QString str) {
    return str.replace(kHwbtnRe, kHwbtnStyleWrapper());
}
```

Swiftb0y (commented):
small code nitpick

src/controllers/legacycontrollersettings.cpp
ronso0: Yup.

src/controllers/legacycontrollersettings.cpp
ronso0: sure, I vaguely remembered there was a rule to avoid that

Holzhaus:
> _Padding is not applied with `padding: 2px;` alone. Might be one of those qss quirks where we need to set an unrelated property, though with `span` neither `display: inline-block;` nor `box-sizing: border-box;` or `border` work_

It works, but only for block elements or table cells. Unfortunately, using table cells also introduces a linebreak, so we can't use it here:
Screenshot from 2024-04-04 20-37-52

Holzhaus:
https://github.com/ronso0/mixxx/pull/66

ronso0:
Thanks for looking into it @Holzhaus 

I think we're good with the &nbsp; solution. I expect it to look odd only if users have a monospace font configured e.g. with Qt6ct.

ronso0:
oh, you already have fix ...

Holzhaus:
> oh, you already have fix ...

No, just some minor improvement unfortunately 😢

src/controllers/legacycontrollersettings.cpp
Swiftb0y: oh, right. well I'm not sure actually. It returns a reference, so it might be dangling, but then that API wouldn't make sense at all... I'm pretty sure its fine because we're returning a value.

Swiftb0y (approved):
LGTM in terms of code. Can't comment much on the formatting because I don't know HTML&CSS well enough.

JoergAtGithub:
LGTM! Thank you!

ronso0:
Thanks everyone for the quick response!

acolombier:
Sorry I didn't have time test this - thanks for looking into this @ronso0 !

PR #13054, opened 2024-04-05
feat: HERCULES INPULSE 500 mappings

jssotomdz:
included colored hotcues, slicer and beatmatch guides


JoergAtGithub:
grafik
unzip it, and apply it using:
`git apply pre-commit.patch`

res/controllers/Hercules-DJControl-Inpulse-500-scripts.js
```diff
@@ -0,0 +1,641 @@
...
+    16557783: 6,
+    15921919: 7,
+    16744448: 8,
+};
+
+//the color conversion for pads is weird so using array to store values
+const colors = [
+    [64, 16, 18, 80, 2, 66, 78, 82, 76],
+    [96, 28, 31, 124, 3, 99, 123, 127, 116],
+];
```
Holzhaus, line 76: This will not work. The colors are arbitrary, as you can define any custom color scheme. Use the `ColorMapper` class instead, that will automatically map the RGB color sent by Mixxx to the closed predefined color available on the controller. 

See https://mixxx.org/news/2020-08-25-new-in-2-3-hotcue-colors/ for a high-level explanation. An example would look like this:

```js
let colors = new ColorMapper({
    0xFF0000: 1, // Red
    0x0000FF: 2, // Blue
    0xFFFF00: 3, // Yellow
    0x00FF00: 4, // Green
    // ...
});

// Get the color of hotcue 1. Lets assume that
// it's some shade of blue (similar but not equal
// to 0x0000FF.
let hotcueColor = engine.getValue("[Channel1]", "hotcue_1_color");
// Get the MIDI value to send to the controller
// that best matches the color.
let value = colors.getValueForNearestColor(hotcueColor);
// Variable `value` now equals 2
```

Holzhaus (commented):
Thanks for contributing.

res/controllers/Hercules-DJControl-Inpulse-500-scripts.js
jssotomdz: actually this controller allows (almost) any rgb color to be displayed. Rewrote the logic a little bit to allow any color to be sent

alephlm:
Something on this files are not working for me, for example when I change songs in the deck, the hot cues LEDs are not changing for the new hotcues. Also when started, the volume of the deck is not in sync with UI until I start moving the slider. And slicer is just red color all LEDs but I do not get anything when pressed.
I am using the latest version of Mixxx just cloned and compiled it (Mixxx 2.6.0-alpha).

I would like to try to contribute, but I get lost on the sea of documentation and cant find what I really need. As a web developer I would expect the XML act as HTML and JS control its behaving.

But for example, in this code, inside `init()`. Where `midi` is coming from?
`midi.sendShortMsg(0x91, 0x03, 0x7F);`
what `sendShortMsg` does? Where can I check what other methods are available?

same questions for engine and its methods.
`engine.makeConnection("[Channel1]", "play", channelLEDsCallback);`


Thank you

kalsifer-742:
image

Does this means that i have to try the mapping myself and review it positively for it to be approved?
I own this controller, so I can do this if it's needed.

(I also wanted to contribute the mapping for this controller, so if I can help let me know)

ronso0:
The Inpulse 500 mapping has been added recently: for main #14316, 2.5 backport is #14491

So I think we can close this.
I didn't look into the details, but it seems the merged mapping has the functions mentioned in the OP (colored hotcues, slicer and beatmatch guides)
If this mapping here has additional functions, I suggest someone picks it up and tries to move them into the official mapping.

kalsifer-742:
@ronso0 thank you for the update!

I undesrstand that this mapping is now considered official and so it's included in Mixxx, correct?
Going forward if I want to improve the mapping, I have to create another pull request right?

(Also the wiki should be updated listing the controller as community supported and not in development)

ronso0:
> I understand that this mapping is now included in Mixxx, correct?

Yes, see https://github.com/mixxxdj/mixxx/tree/2.5/res/controllers

> Going forward if I want to improve the mapping, I have to create another pull request right?

Exactly! Please create isolated PRs for bugfixes (if there are any) and new features.

PR #13057, opened 2024-04-06
fix: segfault when restoring default controller settings

acolombier:
Fixes a crash when pressing `Restore Defaults` on controller preferences when using enumeration and boolean settings.

This crash was missed on initial testing as the mapping only had numerals (QSpinBox).


JoergAtGithub:
Thank you for the quick fix!

PR #13058, opened 2024-04-06
fix: incorrect reboot required notification on preference updates

acolombier:
This fixes the pop up message suggesting to reboot Mixxx after setting updates that shouldn't require a reboot


JoergAtGithub:
LGTM! Thank you!

PR #13064, opened 2024-04-06
Traktor S3: Load controller defaults after loading the controller report

### LINKED ISSUE #13063: keylock always defaulting on
### Bug Description

Keylock is always on by default now, even though my controller config is set to set it off, and mixxx.cfg has the value preset to zero.

### Version

main

### OS

Ubuntu 22.04

ywwg:
Fixes https://github.com/mixxxdj/mixxx/issues/13063

### DIFF
--- a/res/controllers/Traktor-Kontrol-S3-hid-scripts.js
+++ b/res/controllers/Traktor-Kontrol-S3-hid-scripts.js
@@ -297,6 +297,11 @@ TraktorS3.Controller = class {
             TraktorS3.incomingData([inputReportIdx, ...reportData]);
         }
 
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
The code looks correct, but are defualts are the best user experience here? I guess it would be better to uses the states/values set in Mixxx before:
1.) make connections for all COs that should be set to the controller hardware state (esp. knob and fader states)
2.) read the controller state by getInputReport
3.) make remaining connections for COs that should keep it's states/values

ywwg:
> 3.) make remaining connections for COs that should keep it's states/values

yeah I think this might be the correct approach -- it will be annoying to pick apart the two types of COs but it can be done

PR #13067, opened 2024-04-07
(fix) Library: BPM display uses decimal separator of selected locale

### LINKED ISSUE #13051: Library: make BPM display and editing local consistent
### Bug Description

BPM display:
`126.789` (dot)

BPM editor:
`126,789000` (comma)

When I open the editor and type `126.` (dot) any further num input is rejected.
The editor would only accept comma as separator.

#### Settings
* system locale is `en_GB`, in Mixxx I selected **System**.
* 'Reginal format' (numbers, dates, curency) is German, so my numpad separator is `,`

### Fix:
Use the same separator for both.

### Ideal:
Show same separtor in table and editor, allow both in editor.

### Version

_No response_

### OS

_No response_

### LINKED ISSUE #13065: BPM delegate spinbox is limited to 999 but BPM can be higher
### Bug Description

You can set the track BPM higher than 1000 (double existing BPM).
But the BpmDelegate is limited to 3 digits

This may only impact a few users (Speedcore and faster?) but it's an unnecessary limitation nonetheless.

### Effects:
1. Can't edit BPM higher than 1000
* make BPM higher than 1000
* open BPM editor in the library (enable selected click)
* the lineedit will clamp BPM to 1000
* (blindly type 1234 or whatever _because you know what you're doing ;)_ )
* (Return wipes the previous value)
2. Can't set BPM higher than 1000

### Version

2.4

### OS

_No response_

ronso0:
fixes #13051 

[`QString::number(double, format, precision)` is locale-independent](https://github.com/qt/qtbase/blob/d12490f6614ca0ccf43a73fd938144aa5f1b4e8b/src/corelib/text/qlocale_tools.cpp#L769-L771).

IMO this raises the question how we want to display BPM / beats in general:
* Consistent according to selected locale?
Need to fix `WNumber`, too.
Maybe also dialogs like ~~Track Properties~~ :heavy_check_mark:  uses locale
~~And `WBeatSpinBox`~~ :heavy_check_mark:  uses locale
* "International"? (decimal separator is `.` everywhere)
Need to adjust library's BPM delegate (it's editor), since that currently correctly uses the selected locale (`,` or `.`)

The second commit fixes #13065 

The max BPM limit may be moved to some prefs.h file and made available for BPMDelegate, WTrackProperties etc.

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -726,8 +726,9 @@ QVariant BaseTrackTableModel::roleValue(
                 if (role == Qt::ToolTipRole || role == kDataExportRole) {
                     return QString::number(bpm.value(), 'f', 4);
                 } else {
-                    // custom precision, set in DlgPrefLibrary
-                    return QString::number(bpm.value(), 'f', s_bpmColumnPrecision);
+                    // Use the locale here to make the display and editor consistent.
+                    // Custom precision, set in DlgPrefLibrary.
+                    return QLocale().toString(bpm.value(), 'f', s_bpmColumnPrecision);
                 }
             } else {
                 return QChar('-');
--- a/src/library/bpmdelegate.cpp
+++ b/src/library/bpmdelegate.cpp
@@ -21,7 +21,7 @@ class BpmEditorCreator : public QItemEditorCreatorBase {
         QDoubleSpinBox* pBpmSpinbox = new QDoubleSpinBox(parent);
         pBpmSpinbox->setFrame(false);
         pBpmSpinbox->setMinimum(0);
-        pBpmSpinbox->setMaximum(1000);
+        pBpmSpinbox->setMaximum(9999);
         pBpmSpinbox->setSingleStep(1e-3);
         pBpmSpinbox->setDecimals(8);
         pBpmSpinbox->setObjectName("LibraryBPMSpinBox");


daschuer:
Since we translate Mixxx in various languages, also the decimal separator should be translated. 

I can remember a discussion where the decimal comma has been considered as ugly in the bpm display in the skin.

daschuer:
I would vote for changing the library to localized  decimal separator, and everywhere else. We have already the inconsistency between tooltips and grid element. 

The number display in the skin could be still an issue, because the comma uses space below the bottom line of the numbers. For that reason I would vote keeping the decimal dot for all.

ronso0:
> I can remember a discussion where the decimal comma has been considered as ugly in the bpm display in the skin.

Yep. And I bet I was part of the comma group ; )

The widget commmit (WNumber etc.) was a test. I'm not sure, and I think I'll revert it for now (after others have taken a look).
But I think display and editor in the library it should be consistent. I can move the first commit (library) to a new PR or the second (widgets).

daschuer:
> Yep. And I bet I was part of the comma group ; )

That would be definitely "correct". Do you have screenshots at hand, maybe the issue is no longer visible or we may default to native separator and allow overriding it by the skin.

daschuer:
The library commit can be merged as it is.

ronso0:
I removed f7ed7306c7f5414cec9b187d093bb319796a1ad0 and added a comment.
Ready to roll.

ronso0:
I think #13068 is a good idea nonetheless to make the editor more permissive.

daschuer (approved):
LGTM, Thank you.

Issue #12942, opened 2024-03-08
Doubling the bpm again and again will make Mixxx unresponsive


mxmilkiib:
### Bug Description

Use of `beats_set_double` has revealed that repeated doubling of the tempo will make Mixxx unresponsive.

### Version

2.5-alpha

ronso0:
Would be good to know where the threshold is, and where to limit the BPM.
What's a reasonable value? 999 BPM? What would the benefit of being able to increase it further?

ronso0:
I doubled until the beat bars got so dense they filled the waveform, but obviously you pushed it further :laughing:

mxmilkiib:
One subgenre of speedcore, extratone, starts at 1000bpm.

https://en.wikipedia.org/wiki/Speedcore#Extratone

https://daily.bandcamp.com/features/an-introduction-to-extratone-the-worlds-fastest-music-genre

At a minimum for a max, 2k; max max imho would be 15k, maybe 20k. I'm not a speedcore aficionado myself.

ronso0:
So the waveform becomes unusable at around 50.000 BPM, read: the beat markers are so dense they cover the waveform at lowest zoom level.

The GUI becomes overstrained[1] with BPM > 3.000.000

[1] LateNight, one deck playing, core i7-8650U @ 1.90GHz, with a Intel® UHD Graphics 620

I'd say a limit of 20k is safe. I'll prepare a PR.

ronso0:
> I'd say a limit of 20k is safe. I'll prepare a PR.

Seems tedious and has no prio for me, won't happen soon.

PR #13068, opened 2024-04-07
add WDoubleSpinbox: allow . and , in library BPM editor

### LINKED ISSUE #13051: Library: make BPM display and editing local consistent

ronso0:
currently only used in BPMDelegate.
Fixes #13051 for me.

I guess it's debatable whether we 'need' this, so this is a draft until we agreed on yes or no.

I'll probably look into making WBeatSpinBox equally permissive.

### DIFF
--- a/src/library/bpmdelegate.cpp
+++ b/src/library/bpmdelegate.cpp
@@ -8,6 +8,7 @@
 #include <QTableView>
 
 #include "moc_bpmdelegate.cpp"
+#include "widget/wdoublespinbox.h"
 
 // We override the typical QDoubleSpinBox editor by registering this class with
 // a QItemEditorFactory for the BPMDelegate.
@@ -18,7 +19,7 @@ class BpmEditorCreator : public QItemEditorCreatorBase {
     }
 
     QWidget* createWidget(QWidget* parent) const override {
-        QDoubleSpinBox* pBpmSpinbox = new QDoubleSpinBox(parent);
+        WDoubleSpinBox* pBpmSpinbox = new WDoubleSpinBox(parent);
         pBpmSpinbox->setFrame(false);
         pBpmSpinbox->setMinimum(0);
         pBpmSpinbox->setMaximum(1000);
--- a/src/library/bpmdelegate.cpp
+++ b/src/library/bpmdelegate.cpp
@@ -22,7 +22,7 @@ class BpmEditorCreator : public QItemEditorCreatorBase {
         WDoubleSpinBox* pBpmSpinbox = new WDoubleSpinBox(parent);
         pBpmSpinbox->setFrame(false);
         pBpmSpinbox->setMinimum(0);
-        pBpmSpinbox->setMaximum(1000);
+        pBpmSpinbox->setMaximum(99999);
         pBpmSpinbox->setSingleStep(1e-3);
         pBpmSpinbox->setDecimals(8);
         pBpmSpinbox->setObjectName("LibraryBPMSpinBox");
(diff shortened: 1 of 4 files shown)


src/library/bpmdelegate.cpp
```diff
@@ -18,10 +19,10 @@ class BpmEditorCreator : public QItemEditorCreatorBase {
     }
 
     QWidget* createWidget(QWidget* parent) const override {
-        QDoubleSpinBox* pBpmSpinbox = new QDoubleSpinBox(parent);
+        WDoubleSpinBox* pBpmSpinbox = new WDoubleSpinBox(parent);
         pBpmSpinbox->setFrame(false);
         pBpmSpinbox->setMinimum(0);
-        pBpmSpinbox->setMaximum(1000);
+        pBpmSpinbox->setMaximum(99999);
```
acolombier, line 25: Any reason for allowing this much BPM?

src/library/bpmdelegate.cpp
```diff
@@ -18,10 +19,10 @@ class BpmEditorCreator : public QItemEditorCreatorBase {
...
         pBpmSpinbox->setSingleStep(1e-3);
         pBpmSpinbox->setDecimals(8);
```
acolombier, line 27: Not related to this PR but these settings feel overly precise, would you know if that something that was requested at some point or just arbitrary value that could be reviewed for a better UX?

src/widget/wdoublespinbox.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "moc_wdoublespinbox.cpp"
+
+WDoubleSpinBox::WDoubleSpinBox(QWidget* pParent)
+        : QDoubleSpinBox(pParent),
+          m_decSep(QLocale().decimalPoint()) {
+    lineEdit()->setValidator(new QRegExpValidator(
+            QRegExp("[0-9]{1,5}[." +
+                    // add locale decimal point if it's not a dot
+                    (m_decSep != '.' ? m_decSep : QChar()) +
```
acolombier, line 14: Could it be worth to simply allow both `,` and `.`? I know my current local would be `.` but I would still use `,` as I grew up using `,`.

I guess the risk is that some users may type `1,000` for `1000`, but in the meantime, how often do you set the BPM value to > 999? :sweat_smile:

src/library/bpmdelegate.cpp
ronso0: Obviously sky is the limit ; )
https://github.com/mixxxdj/mixxx/issues/12942#issuecomment-1986566470

src/library/bpmdelegate.cpp
ronso0: Yeah, 8 decimals look scary. I think there are two reason:
* for some tracks it seems it's not possible to match the BPM with `beats_adjust_slower/_faster` (`kBpmAdjustStep` is 0.01), the spinbox allows finer adjusments
* have a chance to discover that the BPM display is rounded when there are more than let's say 4 decimals

src/widget/wdoublespinbox.cpp
ronso0: same as https://github.com/mixxxdj/mixxx/pull/13068#discussion_r1624661452

ronso0:
Thanks for your review.
I guess I'll revisit this some time, but ornow this is low prio.

PR #13069, opened 2024-04-07
Tools: Add `rpm_buildenv.sh` for building on Fedora

fwcd:
This adds a new buildenv script for RPM-based distributions, specifically Fedora, to simplify the workflow of building the `main` branch. The corresponding section in the README is also updated to reflect this and now includes a link to the wiki for other distributions (e.g. Arch).

[Here](https://github.com/fwcd/docker-mixxx/blob/02bf0ac75d48c312453c311706a9d4c3c93be6b5/fedora.Dockerfile) is a Dockerfile demonstrating the use of this new script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,18 @@ source code and navigate to it:
     $ git clone https://github.com/mixxxdj/mixxx.git
     $ cd mixxx
 
-Fetch the required dependencies and set up the build environment (on Windows,
-macOS and Debian/Ubuntu, you can do that by running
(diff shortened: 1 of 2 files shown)


fwcd:
I have removed the CI again as per discussion on Zulip, the corresponding commits can still be found in my [`fedora-ci`](https://github.com/fwcd/mixxx/tree/fedora-ci) branch, should we want to reuse it (or some portion thereof) in the future.

daschuer (approved):
LGTM, Thank you.

PR #13070, opened 2024-04-07
Add support for stem in the engine

acolombier:
This PR refactors the engine to prepare for stem mixing.

Currently, all stems are mixed at the deck level, and continue into the mixer and affect manager as a stereo channel.

Future work will allow the deck to embedded its own effect manager so it can be applied to stem individually.

Depends on #13044


daschuer:
Does it make sense to allocate new memory if a STEM track is loaded and discard it if another track is loaded and than allocate it again if a new STEM track is loaded?

acolombier:
I guess it depends.
In the one hand, I guess we could always allocate memory assuming that maximum number of supported channel, but in the other hand, it feels like a waste to always allocate this max, while many user will probably not use STEM in the early days of the new feature.

Perhaps a good middle ground would be to allocate stereo, them in a STEM gets loaded, re-allocate memory to support stem and keep it as is going forward?

daschuer:
Yes, you are probably right. Than let's stick with the dynamic allocation. This is beneficial if we think to 64 Sampler skins.

acolombier:
Finally, after a lot of headache to try and get the worker to play nice with chunk reallocation, I decided to follow your original though @daschuer and stick to static allocation. 

I'm investigating an issues with the `EngineBufferE2ETest` failing due to mismatching sounds. Here is an example of `SeekExact` failure (orange is expected, blue is actual)
image
My gut feeling point me to the change in the linear scaler, but I cannot see what is wrong in there.... Any helps or tip would be appreciated!

src/engine/bufferscalers/enginebufferscalelinear.cpp
```diff
@@ -248,27 +256,29 @@ double EngineBufferScaleLinear::do_scale(CSAMPLE* buf, SINT buf_size) {
...
             // so take the cached sample, this happens on slow rates
-            floor_sample[0] = m_floorSampleOld[0];
-            floor_sample[1] = m_floorSampleOld[1];
-            ceil_sample[0] = m_bufferInt[0];
-            ceil_sample[1] = m_bufferInt[1];
-        } else if (getOutputSignal().frames2samples(currentFrameFloor) + 3 < m_bufferIntSize) {
+            for (int c = 0; c < chCount; c++) {
+                floor_sample[c] = m_floorSampleOld[c];
+                ceil_sample[c] = m_bufferInt[c];
+            }
```
JoergAtGithub, line 267:
```suggestion
            for (int chIdx = 0; chIdx < chCount; chIdx++) {
                floor_sample[chIdx] = m_floorSampleOld[chIdx];
                ceil_sample[chIdx] = m_bufferInt[chIdx];
            }
```

src/engine/bufferscalers/enginebufferscalelinear.cpp
```diff
@@ -219,13 +228,12 @@ double EngineBufferScaleLinear::do_scale(CSAMPLE* buf, SINT buf_size) {
     SINT unscaled_frames_needed = static_cast<SINT>(frames +
             m_dNextFrame - floor(m_dNextFrame));
 
-    CSAMPLE floor_sample[2];
-    CSAMPLE ceil_sample[2];
+    int chCount = getOutputSignal().getChannelCount();
+    std::vector<CSAMPLE> floor_sample(chCount);
+    std::vector<CSAMPLE> ceil_sample(chCount);
```
JoergAtGithub, line 233:
```suggestion
    std::vector<CSAMPLE> floorSample(chCount);
    std::vector<CSAMPLE> ceilSample(chCount);
```

src/engine/bufferscalers/enginebufferscalerubberband.cpp
```diff
@@ -153,24 +178,79 @@ SINT EngineBufferScaleRubberBand::retrieveAndDeinterleave(
...
+                frames);
+        break;
+    default: {
+        int chCount = getOutputSignal().getChannelCount();
+        for (SINT i = 0; i < frames; ++i) {
+            for (int channel = 0; channel < chCount; channel++) {
+                m_buffers[channel].data()[i] =
+                        pBuffer[i * chCount + channel];
+            }
+        }
```
JoergAtGithub, line 251:
```suggestion
        for (SINT frameIdx = 0; frameIdx < frames; ++frameIdx) {
            for (int chIdx = 0; chIdx < chCount; chIdx++) {
                m_buffers[chIdx].data()[frameIdx] =
                        pBuffer[frameIdx * chCount + chIdx];
            }
        }
```

src/engine/bufferscalers/enginebufferscalest.cpp
```diff
@@ -24,22 +24,20 @@ constexpr SINT kSeekOffsetFramesV20101 = 429;
 // TODO() Compensate that. This is probably cause by the delayed adoption of pitch changes due
 // to the SoundTouch chunk size.
 
-constexpr SINT kBackBufferSize = 1024;
+constexpr SINT kBackBufferFrameSize = 512;
```
JoergAtGithub, line 27: Why this change?

acolombier: This used to be premultiplied by 2 (line 34, used for back buffer allocation). Now it gets multiplied by the effective number of channel at runtime (line 93). In practice, the buffer remains of the same size for stereo channels, but will be extended for stem channels

JoergAtGithub: Make sense!

acolombier:
> Yes, you are probably right. Than let's stick with the dynamic allocation. This is beneficial if we think to 64 Sampler skins.

I finally decided to go with a single allocation, but adjust the buffer size depending of the EngineBuffer. This way, preview or sample channels still use the same buffer size, as they remain stereo, while decks will use stem-ready buffers, even if they play only stereo. This way, the memory footprint remains minimum (~+15MB per deck) 

Doing fully dynamic allocation turned out to be a nightmare due to concurrency with the scheduler, and I don't think the refactor is worth the effort.

src/engine/enginebuffer.cpp
```diff
@@ -1232,8 +1276,8 @@ void EngineBuffer::processSlip(int iBufferSize) {
         // TODO: Check if we can replace `iBufferSize` with the number of
         // frames per buffer in most engine method signatures to avoid this
         // back and forth calculations.
-        const int bufferFrameCount = iBufferSize / mixxx::kEngineChannelCount;
-        DEBUG_ASSERT(bufferFrameCount * mixxx::kEngineChannelCount == iBufferSize);
+        const int bufferFrameCount = iBufferSize / m_channelCount;
+        DEBUG_ASSERT(bufferFrameCount * m_channelCount == iBufferSize);
```
JoergAtGithub, line 1280: Please adjust the comment above. It's no longer "a multiple of 2" but a "a multiple of channel count".

src/engine/enginebuffer.h
```diff
@@ -380,6 +385,7 @@ class EngineBuffer : public EngineObject {
 
     SlipModeState m_slipModeState;
 
+    // Track samples are always given assuming a stereo track
     ControlObject* m_pTrackSamples;
     ControlObject* m_pTrackSampleRate;
```
JoergAtGithub, line 390: I think we should either return the correct sample rate (considering the stem channel count), or deprecate the COs `track_samples` and `track_samplerate` and add new COs `track_frames` and `track_framerate`.

JoergAtGithub (changes requested):

src/engine/enginebuffer.h
acolombier: I'd be in favour of deprecating it and introducing frame count only. There is many places which relies on this CO and device by two to compute frame count, so I don't think we can return the correct sample rate without having to introduce a second CO proxy which will have a performance cost. What would be the value to return the "correct" sample rate tho?

JoergAtGithub: I think we should discuss this on Zulip - to get a common understanding of the terms.

acolombier: Sure. Also just FYI on the current behaviour just so that isn't a blocker for this PR, mono tracks would get a "fake" sample count as if they were stereo track, so this comment is only documenting a current behaviour

acolombier: Created the issue #13168 to address in the future

daschuer:
We have a conflict now. Sorry.

src/util/sample.cpp
```diff
@@ -492,6 +492,30 @@ void SampleUtil::interleaveBuffer(CSAMPLE* M_RESTRICT pDest,
...
+        const CSAMPLE* M_RESTRICT pSrc1,
+        const CSAMPLE* M_RESTRICT pSrc2,
+        const CSAMPLE* M_RESTRICT pSrc3,
+        const CSAMPLE* M_RESTRICT pSrc4,
+        const CSAMPLE* M_RESTRICT pSrc5,
+        const CSAMPLE* M_RESTRICT pSrc6,
+        const CSAMPLE* M_RESTRICT pSrc7,
+        const CSAMPLE* M_RESTRICT pSrc8,
+        SINT numFrames) {
+    // note: LOOP VECTORIZED.
```
daschuer, line 506: Did you test it, like described at the beginning of the file?

acolombier: I believe I did, but I will double check that just to be sure.

daschuer: No need to put extra effort in it. I just wanted to rule out a copy and paste issue without testing.

src/engine/readaheadmanager.cpp
```diff
@@ -306,9 +332,11 @@ double ReadAheadManager::getFilePlaypositionFromLog(
...
-        mixxx::audio::FrameDiff_t numConsumedFrames) {
+        mixxx::audio::FrameDiff_t numConsumedFrames,
+        mixxx::audio::ChannelCount channelCount) {
     const double positionSamples =
-            getFilePlaypositionFromLog(currentPosition.toEngineSamplePos(),
-                    numConsumedFrames * mixxx::kEngineChannelCount);
-    return mixxx::audio::FramePos::fromEngineSamplePos(positionSamples);
+            getFilePlaypositionFromLog(currentPosition.toEngineSamplePos(channelCount),
+                    numConsumedFrames * channelCount,
+                    channelCount);
```
daschuer, line 340: The term Playposition, is somehow the lagacy stereo assumption form the COs. 
It would be nice to use mixxx::audio::FramePos only in this function.

daschuer (changes requested):
Here are my fist batch of comments. 
Sorry your hit the rats nest around the stereo assumption in our CO interface.

src/util/sample.cpp
```diff
@@ -660,11 +860,44 @@ void SampleUtil::reverse(CSAMPLE* pBuffer, SINT numSamples) {
 }
 
 // static
-void SampleUtil::copyReverse(CSAMPLE* M_RESTRICT pDest,
-        const CSAMPLE* M_RESTRICT pSrc, SINT numSamples) {
+void SampleUtil::copyReverseStereo(CSAMPLE* M_RESTRICT pDest,
+        const CSAMPLE* M_RESTRICT pSrc,
+        SINT numSamples) {
     for (SINT j = 0; j < numSamples / 2; ++j) {
```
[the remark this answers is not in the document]
JoergAtGithub: I wonder if it's neccessary to calculate endpos every iteration again. Wouldn't something like 
```
// static
void SampleUtil::copyReverseStereo(CSAMPLE* M_RESTRICT pDest,
        const CSAMPLE* M_RESTRICT pSrc,
        SINT numSamples) {
    pSrc += numSamples - 2;
    for (SINT j = 0; j < numSamples; j += 2) {
        pDest[j] = *pSrc--;
        pDest[j + 1] = *pSrc--;
    }
}
```
do the same?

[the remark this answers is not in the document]
JoergAtGithub: The loop unrolling by the fixed values of 2 or 8 helps the compiler with vectorization, because odd channel numbers do not need to be supported and it can pick the optimal register size. Don't know if this constraints are enough for vectorization, but I think it will help.

src/util/sample.cpp
```diff
@@ -660,11 +860,44 @@ void SampleUtil::reverse(CSAMPLE* pBuffer, SINT numSamples) {
...
+// static
+void SampleUtil::copyReverseUnary(CSAMPLE* M_RESTRICT pDest,
+        const CSAMPLE* M_RESTRICT pSrc,
+        SINT numSamples,
+        int channelCount) {
+    DEBUG_ASSERT(numSamples % channelCount == 0);
+    for (SINT frameIdx = 0; frameIdx < numSamples / channelCount; ++frameIdx) {
+        const int endpos = (numSamples - 1) - frameIdx * channelCount;
+        for (int chIdx = 0; chIdx < channelCount; chIdx++) {
+            pDest[frameIdx * channelCount + chIdx] = pSrc[endpos - channelCount - chIdx + 1];
```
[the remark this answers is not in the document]
daschuer: It is! So the copyReverseStem() and copyReverseStereo() versions are redundant. Another reason to keep the specializations as private implementation detail.

src/util/sample.cpp
```diff
@@ -547,6 +671,82 @@ void SampleUtil::linearCrossfadeBuffersIn(
...
 
+// static
+void SampleUtil::linearCrossfadeStemBuffersIn(
+        CSAMPLE* M_RESTRICT pDestSrcFadeIn,
+        const CSAMPLE* M_RESTRICT pSrcFadeOut,
+        SINT numSamples) {
+    // M_RESTRICT unoptimizes the function for some reason.
+    const CSAMPLE_GAIN cross_inc = CSAMPLE_GAIN_ONE / CSAMPLE_GAIN(numSamples / 8);
+    /// note: LOOP VECTORIZED only with "int i" (not SINT i)
+    for (int i = 0; i < numSamples / 8; ++i) {
```
daschuer, line 682: These loops are not vectorized with gcc 9.4.0

src/util/sample.cpp
```diff
@@ -527,7 +575,83 @@ void SampleUtil::linearCrossfadeBuffersOut(
...
 
 // static
-void SampleUtil::linearCrossfadeBuffersIn(
+void SampleUtil::linearCrossfadeStemBuffersOut(
+        CSAMPLE* M_RESTRICT pDestSrcFadeOut,
+        const CSAMPLE* M_RESTRICT pSrcFadeIn,
+        SINT numSamples) {
+    // M_RESTRICT unoptimizes the function for some reason.
+    const CSAMPLE_GAIN cross_inc = CSAMPLE_GAIN_ONE / CSAMPLE_GAIN(numSamples / 8);
+    // note: LOOP VECTORIZED only with "int i" (not SINT i)
```
daschuer, line 584: This is not true with gcc 9.4.0. 
I think the i * 8 pattern is not recognized.

src/util/sample.cpp
daschuer: I can confirm it is vectorized with with gcc 9.4.0.

daschuer:
Here is the complete compiler output: 
```
src/util/sample.cpp:125:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:146:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:154:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:125:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:187:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:125:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:187:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:223:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:229:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:207:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:213:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:240:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:254:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:272:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:279:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:254:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:254:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:299:23: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:322:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:341:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:370:27: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:377:28: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:465:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:777:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:777:24: optimized:  loop versioned for vectorization because of possible aliasing
src/util/sample.cpp:396:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:409:23: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:425:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:451:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:451:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:465:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:478:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:489:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:507:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:507:24: optimized: basic block part vectorized using 16 byte vectors
src/util/sample.cpp:507:24: optimized: basic block part vectorized using 16 byte vectors
src/util/sample.cpp:507:5: optimized: basic block part vectorized using 16 byte vectors
src/util/sample.cpp:525:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:543:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:570:23: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:564:23: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:667:23: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:661:23: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:757:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:767:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:777:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:777:24: optimized:  loop versioned for vectorization because of possible aliasing
src/util/sample.cpp:791:14: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:791:14: optimized:  loop versioned for vectorization because of possible aliasing
src/util/sample.cpp:803:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:815:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:829:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:829:24: optimized:  loop versioned for vectorization because of possible aliasing
src/util/sample.cpp:843:24: optimized: loop vectorized using 16 byte vectors
src/util/sample.cpp:878:38: optimized: basic block part vectorized using 16 byte vectors
src/util/sample.cpp:899:35: optimized: loop vectorized using 16 byte vectors
```

src/audio/frame.h
```diff
@@ -77,11 +83,13 @@ class FramePos final {
     /// for compatibility with our control objects and legacy parts of the code
     /// base. Using a different code path based on the output of `isValid()` is
     /// preferable.
-    double toEngineSamplePosMaybeInvalid() const {
+    double toEngineSamplePosMaybeInvalid(
+            mixxx::audio::ChannelCount channelCount =
+                    mixxx::kEngineChannelOutputCount) const {
```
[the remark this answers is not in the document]
acolombier: I have introduced `toSample` and `fromSample` functions. I have taken the freedom to use `SINT` for return with a `isFractionnal` assert + some doc as to why (TL;DR: doesn't make sense to work with half a sample in engine buffers) . Alternative would be about ~50 static_cast on the engine code...

src/engine/readaheadmanager.cpp
acolombier: I have done so now (commit is incoming) but that this adding a significant amount of change  (and risk?) to the PR. Happy to move on it's own if you think it makes sense, like for the rubberband multi threading

src/engine/readaheadmanager.cpp
```diff
@@ -285,7 +311,7 @@ double ReadAheadManager::getFilePlaypositionFromLog(
             if (m_pRateControl) {
                 const auto seekPosition =
                         mixxx::audio::FramePos::fromEngineSamplePos(
-                                entry.virtualPlaypositionStart);
+                                entry.virtualPlaypositionStart, channelCount);
```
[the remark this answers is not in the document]
acolombier: Went with the `FramePos`

src/engine/readaheadmanager.cpp
daschuer: Thank you! Separate PRs are always welcome. Decide yourself if it worth the efforts.

acolombier: I actually could get all the test to pass with my commit, so I will move it to another PR.

acolombier:
@daschuer I have updated the vectorization info using a small script and caught some other incorrect in [905bce1588](https://github.com/acolombier/mixxx/commit/905bce1588). 
With your approval, I would like to keep the refactoring of samples to frames  in ReadAheadManager/CachingReader as a follow up PR

JoergAtGithub (approved):

acolombier:
FYI, next PR in order would be the analyzer support in #13106

JoergAtGithub:
@daschuer Are your concerns addressed properly?

daschuer:
Review is in progress.

JoergAtGithub:
There are conflicts now, after merging the Multithreaded Rubberband PR #13143.

I wonder which options the user needs for multithreading with stems. I guess stem support without any multithreading is not of much practical use. But should we offer proccessing of each mono channel in an own thread for stems decks too? Or is it better to process the 4 stereo channels of a stems decks always in 4 threads.

src/util/sample.cpp
```diff
@@ -183,7 +183,6 @@ void SampleUtil::applyAlternatingGain(CSAMPLE* pBuffer, CSAMPLE gain1,
         return;
     }
 
-    // note: LOOP VECTORIZED.
```
daschuer, line 186: Interesting. Maybe we need to add different comments for different compilers? 
The note was mainly because the loop shall not be touched to not disallow vectorization. 
I will investigate the issue.

daschuer (changes requested):
Added some comments. Thank you for all the heavy lifting to N-Channel engine.

acolombier:
> I wonder which options the user needs for multithreading with stems. I guess stem support without any multithreading is not of much practical use. But should we offer proccessing of each mono channel in an own thread for stems decks too? Or is it better to process the 4 stereo channels of a stems decks always in 4 threads.

Yeah, my plan was to use multi threading for stem unconditionally. I still need to do some benchmark if there is value and treating stem as 8 individual mono or is it best to treat them as 4 stereo channels. My gut feeling is, it will depend of the number of cores, but I'll see and report back.

daschuer:
> ... 8 individual mono or is it best to treat them as 4 stereo channels.

That is not a question IMHO, because mono Rubberband destroys the stereo effect and makes no compatibility.

acolombier:
> That is not a question IMHO, because mono Rubberband destroys the stereo effect and makes no compatibility.

Isn't it a problem we accepted to go with anyway in #13143 and with the user prompt about it? My thoughts was, if the benchmark shows better result, have mono processing for both stereo and stem track when "multi threaded RB" is on, only stereo processing on stem otherwise.

daschuer:
> Isn't it a problem we accepted to go with anyway in https://github.com/mixxxdj/mixxx/pull/13143 and with the user prompt about it? 

Yes. 

> My thoughts was, if the benchmark shows better result, 

I am pretty sure benchmarks will show better results, especially on high number of cores, but since the quality is worse and the user expects the best quality when using Rubberband, the default should be stereo processing with four threads in case of stems. The rest can be optional, along with explanations.

src/engine/bufferscalers/enginebufferscalelinear.cpp
```diff
@@ -219,13 +228,12 @@ double EngineBufferScaleLinear::do_scale(CSAMPLE* buf, SINT buf_size) {
...
+    int chCount = getOutputSignal().getChannelCount();
+    std::vector<CSAMPLE> floorSample(chCount);
+    std::vector<CSAMPLE> ceilSample(chCount);
 
-    floor_sample[0] = 0;
-    floor_sample[1] = 0;
-    ceil_sample[0] = 0;
-    ceil_sample[1] = 0;
+    std::fill(floorSample.begin(), floorSample.end(), 0.0);
+    std::fill(ceilSample.begin(), ceilSample.end(), 0.0);
```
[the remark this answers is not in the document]
acolombier: Yes, @daschuer already raised that above. I'm going to make some refactor hand use the `onChange` hook to perform allocation channel count or sampler changes.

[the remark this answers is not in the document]
Swiftb0y: whoops. sorry for the double comment then. I made the comment a couple days ago and just forgot to submit the review until now.

src/engine/enginebuffer.cpp
```diff
@@ -538,7 +541,24 @@ void EngineBuffer::slotTrackLoaded(TrackPointer pTrack,
...
+    if (m_channelCount > mixxx::audio::ChannelCount::stereo()) {
+        // The sample count is indicated downmix. This means that for stem
+        // track, we only consider the track in stereo, as it is perceived by
+        // the user on deck output
+        VERIFY_OR_DEBUG_ASSERT(m_channelCount % mixxx::audio::ChannelCount::stereo() == 0) {
+            // Make it stereo for the frame calculation
+            kLogger.warning() << "Odd number of channel in the track is not supported";
+        };
+    } else {
+        // The EngineBuffer only works with stereo channels. If the track is
```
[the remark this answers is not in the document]
acolombier: This cannot be asserted at that level as the `AudioSourceStereoProxy` will be instantiated by `CachingReaderChunk` and won't return a channel count to the `EngineBuffer`. Do you have a suggestion on how this could be asserted otherwise?

src/engine/channels/enginedeck.cpp
```diff
@@ -44,6 +49,29 @@ EngineDeck::~EngineDeck() {
...
+    int stereoChannelCount = m_pBuffer->getChannelCount() / mixxx::kEngineChannelOutputCount;
+    auto allChannelBufferSize = iBufferSize * stereoChannelCount;
+    if (m_stemBuffer.size() < allChannelBufferSize) {
+        m_stemBuffer = mixxx::SampleBuffer(allChannelBufferSize);
+    }
+    m_pBuffer->process(m_stemBuffer.data(), allChannelBufferSize);
+
+    // TODO(XXX): process effects per stems
+
+    SampleUtil::clear(pOut, iBufferSize);
```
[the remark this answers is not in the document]
acolombier: This is used to set the buffer to zero, before mixing (summing) all the stereo channel in there. Note that my first version used to copy the first stereo channel into pOut, then the merge would only happens for the next channels. However, I simplified this because of what is in the next PR with stem gain, so I thought I will keep the structure as it needs to be for the future.

Not sure if that make sense, but to summarize, I will need this for loop later, so I thought I will use it, even if it could be simplified at that stage.

src/engine/channels/enginedeck.cpp
```diff
@@ -44,6 +49,29 @@ EngineDeck::~EngineDeck() {
...
+    const CSAMPLE* pIn = m_stemBuffer.data();
+    for (int i = 0; i < iBufferSize; i += mixxx::kEngineChannelOutputCount) {
+        for (int chIdx = 0; chIdx < m_pBuffer->getChannelCount();
```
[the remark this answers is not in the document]
acolombier: This is not a copy -  all stereo channel get mixed (summed) into one. See my previous comment

src/engine/cachingreader/cachingreaderchunk.cpp
```diff
@@ -49,34 +48,51 @@ mixxx::IndexRange CachingReaderChunk::bufferSampleFrames(
...
+                    mixxx::audio::ChannelCount::stereo() !=
+            0) {
+        // This happens if the audio source only contain a mono channel, or an
+        // uneven number of channel
+        mixxx::AudioSourceStereoProxy audioSourceProxy(
+                pAudioSource,
+                tempOutputBuffer);
+        DEBUG_ASSERT(
+                audioSourceProxy.getSignalInfo().getChannelCount() ==
+                mixxx::audio::ChannelCount::stereo());
```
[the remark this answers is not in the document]
acolombier: > I am concerned that this will leads to out of bounds access later. But probably this for no reason? 

Sorry - which part? That code should be strictly equivalent to what it used to be.

> Where is the code that handles to throw away or zero out channels?

I have added a VERIFY assert to disqualify zero channel in the parent owning pAudioSource which hopefully should be enough. 
AFAIK, the expected behaviour of `AudioSource::readSampleFrames` is to return the range of available sample, which should be zero in this case, so adding further assert here feels a little late and redundant

src/engine/bufferscalers/enginebufferscalerubberband.cpp
```diff
@@ -153,24 +178,79 @@ SINT EngineBufferScaleRubberBand::retrieveAndDeinterleave(
...
+                m_buffers[5].data(frame_offset),
+                m_buffers[6].data(frame_offset),
+                m_buffers[7].data(frame_offset),
+                received_frames);
+        break;
+    default: {
+        int chCount = getOutputSignal().getChannelCount();
+        for (SINT frameIdx = 0; frameIdx < frames; ++frameIdx) {
+            for (int channel = 0; channel < chCount; channel++) {
+                m_buffers[channel].data()[frameIdx] =
```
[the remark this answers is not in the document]
acolombier: I cannot use copy here as I need to interleave the samples. I have added a comment as to why, let me know if that makes sense.

src/engine/bufferscalers/enginebufferscalerubberband.cpp
```diff
@@ -153,24 +178,79 @@ SINT EngineBufferScaleRubberBand::retrieveAndDeinterleave(
...
+                m_buffers[6].data(),
+                m_buffers[7].data(),
+                pBuffer,
+                frames);
+        break;
+    default: {
+        int chCount = getOutputSignal().getChannelCount();
+        for (SINT i = 0; i < frames; ++i) {
+            for (int channel = 0; channel < chCount; channel++) {
+                m_buffers[channel].data()[i] =
```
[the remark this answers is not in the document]
acolombier: Did the same than [here](https://github.com/mixxxdj/mixxx/pull/13070/files#r1632312289), let me know if it makes sense

src/engine/bufferscalers/enginebufferscale.cpp
```diff
@@ -16,12 +17,22 @@ EngineBufferScale::EngineBufferScale()
     DEBUG_ASSERT(!m_outputSignal.isValid());
 }
 
-void EngineBufferScale::setSampleRate(
-        mixxx::audio::SampleRate sampleRate) {
+void EngineBufferScale::setOutputSignal(
```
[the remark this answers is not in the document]
acolombier: Yes,  I think this is no very well named. There is no input or output signal here, input is always equal to output, thus this work no make the scaler channel agnostic (as there were already sample rate agnostic). Will rename to just `setSignal`

src/engine/cachingreader/cachingreaderchunk.cpp
```diff
@@ -49,34 +48,51 @@ mixxx::IndexRange CachingReaderChunk::bufferSampleFrames(
...
-            kChannels);
-    m_bufferedSampleFrames =
-            audioSourceProxy.readSampleFrames(
-                    mixxx::WritableSampleFrames(
-                            sourceFrameIndexRange,
-                            mixxx::SampleBuffer::WritableSlice(m_sampleBuffer)));
+
+    if (pAudioSource->getSignalInfo().getChannelCount() %
+                    mixxx::audio::ChannelCount::stereo() !=
+            0) {
```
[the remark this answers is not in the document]
acolombier: Whatever I try, it looks like clang format revert it back to that. Are we okay to just keep it as it? I guess I could put the `0 !=` at the beginning otherwise.

src/engine/cachingreader/cachingreaderworker.cpp
```diff
@@ -202,6 +204,19 @@ void CachingReaderWorker::loadTrack(const TrackPointer& pTrack) {
...
 
+    // It is critical that the audio source doesn't contain more channels than
+    // requested as this could lead to overflow when reading chunks
+    VERIFY_OR_DEBUG_ASSERT(m_pAudioSource->getSignalInfo().getChannelCount() <=
+            m_maxSupportedChannel) {
+        m_pAudioSource.reset(); // Close open file handles
+        const auto update = ReaderStatusUpdate::trackUnloaded();
+        m_pReaderStatusFIFO->writeBlocking(&update, 1);
+        emit trackLoadFailed(pTrack,
+                tr("The file '%1' could not be loaded.")
```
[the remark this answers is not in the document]
acolombier: Is that better?

src/engine/bufferscalers/enginebufferscalerubberband.h
```diff
@@ -67,10 +67,10 @@ class EngineBufferScaleRubberBand final : public EngineBufferScale {
     /// The audio buffers samples used to send audio to Rubber Band and to
     /// receive processed audio from Rubber Band. This is needed because Mixxx
     /// uses interleaved buffers in most other places.
-    std::array<mixxx::SampleBuffer, 2> m_buffers;
+    std::vector<mixxx::SampleBuffer> m_buffers;
     /// These point to the buffers in `m_buffers`. They can be defined here
     /// since this object cannot be moved or copied.
-    std::array<float*, 2> m_bufferPtrs;
+    std::vector<float*> m_bufferPtrs;
```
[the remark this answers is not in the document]
acolombier: These are already `mixxx::SampleBuffer`. `m_bufferPtrs` is meant to generate the `float**` which RB required, however, the `float*` are actually pointer to `mixxx::SampleBuffer`, held by the `m_buffers` defined just above. This is what the existing comment was trying to explain but perhaps a reword is necessary. Any suggestion if so?

src/engine/bufferscalers/enginebufferscalelinear.cpp
```diff
@@ -17,15 +17,19 @@ EngineBufferScaleLinear::EngineBufferScaleLinear(ReadAheadManager *pReadAheadMan
       m_dOldRate(1.0),
       m_dCurrentFrame(0.0),
       m_dNextFrame(0.0) {
-    m_floorSampleOld[0] = 0.0;
-    m_floorSampleOld[1] = 0.0;
+    onOutputSignalChanged();
     SampleUtil::clear(m_bufferInt, kiLinearScaleReadAheadLength);
```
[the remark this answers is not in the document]
acolombier: I have removed the clearing of `m_floorSampleOld`

acolombier:
I have added the multi threaded rubberband logic:
- If `Multi threaded` is off, stereo channel will be treated together 
- If on, stereo and stem decks will be treated in mono

I have also added more logic to try and optimise the channel per worker depending of the core, as no every setup would have a 8 cores or more

src/engine/bufferscalers/enginebufferscalerubberband.h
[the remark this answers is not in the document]
daschuer: Oh, right. I was confused when writing the comment.

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -11,9 +11,46 @@ using RubberBand::RubberBandStretcher;
...
+///  | NbOfCore | Stereo | Stem |
+///  |----------|--------|------|
+///  | 1        | 2      | 8    |
+///  | 2        | 1      | 4    |
+///  | 3        | 1      | 2    |
+///  | 4        | 1      | 2    |
+///  | 5        | 1      | 2    |
+///  | 6        | 1      | 2    |
+///  | 7        | 1      | 2    |
+///  | 8        | 1      | 1    |
```
daschuer, line 29: We should always use two channels (stereo) threads, unless the user insists to use single channels which is not recommended.

src/engine/cachingreader/cachingreaderchunk.cpp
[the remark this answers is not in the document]
daschuer: Ok, I though it was easy.

src/engine/enginebuffer.cpp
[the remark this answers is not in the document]
daschuer: OK, than keep it as it is.

src/engine/bufferscalers/rubberbandwrapper.cpp
acolombier: Yes, that's already the case. This table is only valid if the user has requested multi threading (and confirm the warning prompt)

src/engine/channels/enginedeck.cpp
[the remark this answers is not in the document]
acolombier: `SampleUtil::add` assumes a constant index, which isn't the case here. 

I would really like not to waste time on getting `EngineDeck::processStem` perfect as it gets pretty much rewritten again in #13106 and #13123, so I would rather getting something that works for now, and focus on perfection in the last iteration, which is #13123.

src/engine/bufferscalers/rubberbandwrapper.cpp
JoergAtGithub: I think we should sharpen the terminology here. Instead of just "multi threading" we should call it "process stereo channels in seperated threads" or similar.

acolombier: Fair enough, I will rework the tooltip in the preference window

src/engine/channels/enginedeck.cpp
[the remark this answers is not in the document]
JoergAtGithub: > I would really like not to waste time on getting `EngineDeck::processStem` perfect as it gets pretty much rewritten again in #13106 and #13123, so I would rather getting something that works for now, and focus on perfection in the last iteration, which is #13123.

@daschuer Do you agree, that we do not optimize the code of processStem in this PR yet, and instead do a in-depth review in #13123?

src/util/sample.cpp
```diff
@@ -527,33 +568,202 @@ void SampleUtil::linearCrossfadeBuffersOut(
...
+            const CSAMPLE_GAIN cross_inc =
+                    CSAMPLE_GAIN_ONE / CSAMPLE_GAIN(numSamples / channelCount);
+            for (int chIdx = 0; chIdx < channelCount; chIdx++) {
+                for (int i = 0; i < numFrame; ++i) {
+                    const CSAMPLE_GAIN cross_mix = cross_inc * i;
+                    pDestSrcFadeOut[i * channelCount + chIdx] *= (CSAMPLE_GAIN_ONE - cross_mix);
+                    pDestSrcFadeOut[i * channelCount + chIdx] +=
+                            pSrcFadeIn[i * channelCount + chIdx] * cross_mix;
+                }
+            }
```
daschuer, line 650:
```suggestion
            for (int i = 0; i < numFrame; ++i) {
                const CSAMPLE_GAIN cross_mix = cross_inc * i;
                // note: LOOP VECTORIZED
                for (int chIdx = 0; chIdx < channelCount; chIdx++) {
                    pDestSrcFadeOut[i * channelCount + chIdx] *= (CSAMPLE_GAIN_ONE - cross_mix);
                    pDestSrcFadeOut[i * channelCount + chIdx] +=
                            pSrcFadeIn[i * channelCount + chIdx] * cross_mix;
                }
            }
```

src/preferences/dialog/dlgprefsounddlg.ui
```diff
@@ -231,9 +231,6 @@
            <verstretch>0</verstretch>
           </sizepolicy>
          </property>
-         <property name="toolTip">
-          <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Warning!&lt;/span&gt;&lt;/p&gt;&lt;p&gt;Using multi threading may result in pitch and tone imperfaction depending of the platform, leading to mono-incompatibiltiy, due to third party limitations. &lt;/p&gt;&lt;p&gt;&lt;br/&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
-         </property>
          <property name="text">
           <string>Multi-threaded</string>
```
daschuer, line 235: This is highly misleading in the stem context, because stems are always processed multi threaded. 
How about "Dual-threaded Stereo"?

src/preferences/dialog/dlgprefsounddlg.ui
```diff
@@ -231,9 +231,6 @@
            <verstretch>0</verstretch>
           </sizepolicy>
          </property>
-         <property name="toolTip">
-          <string>&lt;html&gt;&lt;head/&gt;&lt;body&gt;&lt;p&gt;&lt;span style=&quot; font-weight:600;&quot;&gt;Warning!&lt;/span&gt;&lt;/p&gt;&lt;p&gt;Using multi threading may result in pitch and tone imperfaction depending of the platform, leading to mono-incompatibiltiy, due to third party limitations. &lt;/p&gt;&lt;p&gt;&lt;br/&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
```
daschuer, line 235: The tooltip was not to bad. Since we have now the message box we can sharpen it a bit: 
"Using dual-threaded stereo may improve CPU utilization, but results to mono-incompatibility (Therefore not recommended)."
Without any formatting.

daschuer (changes requested):
Here some comments

src/util/sample.cpp
```diff
@@ -527,33 +568,202 @@ void SampleUtil::linearCrossfadeBuffersOut(
...
+                    CSAMPLE_GAIN_ONE / CSAMPLE_GAIN(numSamples / channelCount);
+            for (int chIdx = 0; chIdx < channelCount; chIdx++) {
+                for (int i = 0; i < numFrame; ++i) {
+                    const CSAMPLE_GAIN cross_mix = cross_inc * i;
+                    pDestSrcFadeIn[i * channelCount + chIdx] *= cross_mix;
+                    pDestSrcFadeIn[i * channelCount + chIdx] +=
+                            pSrcFadeOut[i * channelCount + chIdx] *
+                            (CSAMPLE_GAIN_ONE - cross_mix);
+                }
+            }
```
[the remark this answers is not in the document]
acolombier: Looking at `SampleUtil::linearCrossfadeStemBuffersIn`, it was containing 8 loops instead of a single one. I believe the specialisation should now be feaster due to vectorization. Could you test it again, so we get consistent hardware to benchmark on?

src/preferences/dialog/dlgprefsounddlg.ui
acolombier: The tooltip is injected dynamically in the UI, this was already the case, and this hardcoded tooltip was redundant. See `DlgPrefSound`

src/preferences/dialog/dlgprefsounddlg.ui
acolombier: I like it, that sounds fancy too!

src/util/sample.cpp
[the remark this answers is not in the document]
daschuer: Wow, It is blazing fats now: 
BM_LinearCrossfadeBuffersOut/64           16,6 ns         16,6 ns     41604160
BM_LinearCrossfadeBuffersOut/512           121 ns          121 ns      5735540
BM_LinearCrossfadeBuffersOut/4096          978 ns          978 ns       747782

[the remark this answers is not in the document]
daschuer: The same trick fails totally for stereo. Can you apply the same trick for 'SampleUtil::linearCrossfadeStemBuffersOut' ?

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -37,22 +37,29 @@ bool soundItemAlreadyExists(const AudioPath& output, const QWidget& widget) {
...
-                "Using multi "
-                "threading may result in pitch and tone imperfection, and this "
+                "Processing stereo signal as mono channel "
+                "may result in pitch and tone imperfection, and this "
                 "is "
                 "mono-incompatible, due to third party limitations.") +
+        QStringLiteral("</p><p>") +
+        QObject::tr(
+                "Note that stem tracks will always be processed in parallel, "
+                "as individual stereo channels.") +
```
daschuer, line 54: This sentence is confusing for me. Delete?

src/util/sample.cpp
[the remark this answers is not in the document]
acolombier: Done!

daschuer:
Thank you. Just left some final comments. We have also a conflict.

daschuer (approved):
LGTM, thank you.

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -303,6 +306,7 @@ DlgPrefSound::DlgPrefSound(QWidget* pParent,
 void DlgPrefSound::slotUpdate() {
     m_bSkipConfigClear = true;
     loadSettings();
+    settingChanged();
```
ronso0, line 309: @acolombier Why was this change necessary?
Thing is, with #12194, Apply will reapply the soundconfig unexpectedly for no obvious reason which causes a playback pause.

acolombier: This was necessary to update the dual threading components, but I guess it could have been in `loadSettings`

PR #13071, opened 2024-04-07
README: Recommend running buildenvs over sourcing them on Linux

fwcd:
A small addendum to #13069, I've noticed that the wiki recommends running the `debian_buildenv.sh` script (in a subshell) instead of sourcing it. Since neither the Debian nor the Fedora/RPM script export any variables, I think we can safely recommend that in the README too, this makes them a bit more robust against users running other shells too, since that way the scripts will always run in Bash. For macOS, we'll still need to source, but the script already guards against that.


daschuer (approved):
Thank you for noticing. LGTM

PR #13072, opened 2024-04-07
ScreenSaver: Add inhibit portal support on Linux

djantti:
This PR implements modern portal based idle inhibit support on Linux and also fixes non-functional idle inhibit on Flatpak Mixxx.

The inhibit portal is supported by both Gnome and KDE and the original idea was to use it as a default. Unfortunately testing revealed undocumented compatibility issues, so portal inhibit is now attempted as a fallback and only if Mixxx is running in a Flatpak sandbox.

I tried to keep the original code logic unchanged and improve things where it seemed appropriate. Let me know if there's anything that should be further improved. Thanks!

### DIFF
--- a/src/util/screensaver.cpp
+++ b/src/util/screensaver.cpp
@@ -1,11 +1,16 @@
 /**
 Documentation:
-OSX: https://developer.apple.com/reference/iokit/1557134-iopmassertioncreatewithname
-Windows: https://msdn.microsoft.com/en-us/library/windows/desktop/aa373208(v=vs.85).aspx
-Freedesktop: https://people.freedesktop.org/~hadess/idle-inhibition-spec/re01.html
+OSX:
+https://developer.apple.com/reference/iokit/1557134-iopmassertioncreatewithname
+Windows:
+https://msdn.microsoft.com/en-us/library/windows/desktop/aa373208(v=vs.85).aspx
+Freedesktop:
+https://people.freedesktop.org/~hadess/idle-inhibition-spec/re01.html
 XScreenSaver: https://linux.die.net/man/3/xscreensaversuspend
-GTK: https://developer.gnome.org/gtk3/stable/GtkApplication.html#gtk-application-inhibit
-Portal: https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.Inhibit
+GTK:
+https://developer.gnome.org/gtk3/stable/GtkApplication.html#gtk-application-inhibit
+Portal:
+https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.Inhibit
 
 With the help of the following source codes:
 
@@ -205,8 +210,7 @@ void ScreenSaverHelper::triggerUserActivity()
     return;
 }
 
-void ScreenSaverHelper::inhibitInternal()
-{
+void ScreenSaverHelper::inhibitInternal() {
     if (!QDBusConnection::sessionBus().isConnected()) {
         qWarning() << "Cannot connect to the D-Bus session bus";
         return;
@@ -273,8 +277,7 @@ void ScreenSaverHelper::inhibitInternal()
     }
 }
 
-void ScreenSaverHelper::uninhibitInternal()
-{
+void ScreenSaverHelper::uninhibitInternal() {
     if (!inhibitor) {
         qDebug() << "Cannot uninhibit without a selected interface";
         return;
--- a/src/util/screensaver.cpp
+++ b/src/util/screensaver.cpp
@@ -166,34 +166,34 @@ const Inhibitor kFreeDesktopPortal = {
         .interface = QStringLiteral("org.freedesktop.portal.Inhibit")};
 
 // Standard D-Bus idle inhibitors
-const QVector<Inhibitor> kDBusInhibitors = {
-        {.service = QStringLiteral("org.freedesktop.ScreenSaver"),
-                .object_path = QStringLiteral("/org/freedesktop/ScreenSaver"),
-                .interface = QStringLiteral("org.freedesktop.ScreenSaver")},
-        {.service = QStringLiteral("org.gnome.ScreenSaver"),
-                .object_path = QStringLiteral("/org/gnome/ScreenSaver"),
-                .interface = QStringLiteral("org.gnome.ScreenSaver")},
-        {.service = QStringLiteral("org.kde.ScreenSaver"),
-                .object_path = QStringLiteral("/org/kde/ScreenSaver"),
-                .interface = QStringLiteral("org.kde.ScreenSaver")},
-        {.service = QStringLiteral("org.cinnamon.ScreenSaver"),
-                .object_path = QStringLiteral("/org/cinnamon/ScreenSaver"),
-                .interface = QStringLiteral("org.cinnamon.ScreenSaver")},
-        {.service = QStringLiteral("org.mate.ScreenSaver"),
-                .object_path = QStringLiteral("/org/mate/ScreenSaver"),
-                .interface = QStringLiteral("org.mate.ScreenSaver")},
-        {.service = QStringLiteral("org.xfce.screenSaver"),
-                .object_path = QStringLiteral("/org/xfce/ScreenSaver"),
-                .interface = QStringLiteral("org.xfce.ScreenSaver")}};
+const std::array<Inhibitor, 6> kDBusInhibitors = {
+        {{.service = QStringLiteral("org.freedesktop.ScreenSaver"),
+                 .object_path = QStringLiteral("/org/freedesktop/ScreenSaver"),
+                 .interface = QStringLiteral("org.freedesktop.ScreenSaver")},
+                {.service = QStringLiteral("org.gnome.ScreenSaver"),
+                        .object_path = QStringLiteral("/org/gnome/ScreenSaver"),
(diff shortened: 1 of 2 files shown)


src/util/screensaver.cpp
```diff
@@ -145,29 +152,48 @@ void ScreenSaverHelper::uninhibitInternal()
...
+                .interface = QStringLiteral("org.cinnamon.ScreenSaver")},
+        {.service = QStringLiteral("org.mate.ScreenSaver"),
+                .object_path = QStringLiteral("/org/mate/ScreenSaver"),
+                .interface = QStringLiteral("org.mate.ScreenSaver")},
+        {.service = QStringLiteral("org.xfce.screenSaver"),
+                .object_path = QStringLiteral("/org/xfce/ScreenSaver"),
+                .interface = QStringLiteral("org.xfce.ScreenSaver")}};
+
+// Pointer to selected idle inhibitor
+const Inhibitor* inhibitor = nullptr;
```
Swiftb0y, line 190: this should also be static and non-const I think.

src/util/screensaver.cpp
```diff
@@ -145,29 +152,48 @@ void ScreenSaverHelper::uninhibitInternal()
...
+
+// Inhibit portal handle
+QString ScreenSaverHelper::session_handle = "";
+
+// Standard inhibit cookie
+uint32_t ScreenSaverHelper::inhibit_cookie = 0;
```
Swiftb0y, line 196: we still have the rule to prefix static variables with `s_` iirc.

src/util/screensaver.cpp
```diff
@@ -183,100 +209,119 @@ void ScreenSaverHelper::triggerUserActivity()
...
+            return;
+        } else {
+            qWarning() << "Inhibit for" << kFreeDesktopPortal.interface << "failed: "
+                       << reply.error().message();
+        }
+    } else {
+        qDebug() << "Interface" << kFreeDesktopPortal.interface << "is not valid";
+    }
+
+    for (int i = 0; i < kDBusInhibitors.size(); ++i) {
```
Swiftb0y, line 249: use ranged based for-loop.
```suggestion
    for (const Inhibitor& inhibitor : kDBusInhibitors) {
```

src/util/screensaver.cpp
```diff
@@ -183,100 +209,119 @@ void ScreenSaverHelper::triggerUserActivity()
...
+        // Portal unavailable, try to use standard D-Bus idle inhibit interfaces
+        QDBusInterface iface(kDBusInhibitors[i].service,
+                kDBusInhibitors[i].object_path,
```
Swiftb0y, line 252: for use with ranged-based for-loop
```suggestion
                inhibitor.object_path,
```

src/util/screensaver.cpp
```diff
@@ -183,100 +209,119 @@ void ScreenSaverHelper::triggerUserActivity()
...
+            s_enabled = true;
+            qDebug() << "Idle inhibitor" << kFreeDesktopPortal.interface << "enabled";
+            return;
+        } else {
+            qWarning() << "Inhibit for" << kFreeDesktopPortal.interface << "failed: "
+                       << reply.error().message();
+        }
+    } else {
+        qDebug() << "Interface" << kFreeDesktopPortal.interface << "is not valid";
+    }
```
Swiftb0y, line 247: the exact same code is duplicated below (semantically). I think its nice if it was factored out into a utility function. that just took the `Inhibitor&` and the replay as a parameter.

src/util/screensaver.cpp
```diff
@@ -183,100 +209,119 @@ void ScreenSaverHelper::triggerUserActivity()
...
     }
-    if (s_cookie > 0) {
-        uninhibit();
-    }
-    s_cookie = 0;
-    for (int i=0; SCREENSAVERS[i][0] != nullptr; i++ ) {
-        QDBusInterface iface(SCREENSAVERS[i][0], SCREENSAVERS[i][1], SCREENSAVERS[i][2], 
-            QDBusConnection::sessionBus());
+
+    if (inhibitor->service == "org.freedesktop.portal.Desktop" && !session_handle.isEmpty()) {
```
Swiftb0y, line 286: wouldn't this also work?
```suggestion
    if (inhibitor == &kFreeDesktopPortal && !session_handle.isEmpty()) {
```

Swiftb0y (commented):
In general, the code could be nicer if it was making use of some polymorphism, but I don't want to force you to rewrite your suggestion entirely, so here is some surface-level feedback.

src/util/screensaver.cpp
Swiftb0y: maybe. I haven't looked too closely.

Swiftb0y:
Great. In terms of polymorphism, I think it would make sense to go for standard inheritance here. Eg. setup an abstract screensaver helper interface here and then inherit from that to build the platform dependent implementations. Mac, IOS, and windows would obviously have only one implementation to choose from, but but on Linux, there would be multiple and the appropriate one would be instanced by a factory function at runtime.

src/util/screensaver.cpp
```diff
@@ -145,29 +152,48 @@ void ScreenSaverHelper::uninhibitInternal()
...
-int ScreenSaverHelper::s_saverindex = -1;
-bool ScreenSaverHelper::s_sendActivity = true;
+// Portal idle ihibitor
+const Inhibitor kFreeDesktopPortal = {
+        .service = QStringLiteral("org.freedesktop.portal.Desktop"),
+        .object_path = QStringLiteral("/org/freedesktop/portal/desktop"),
+        .interface = QStringLiteral("org.freedesktop.portal.Inhibit")};
+
+// Standard D-Bus idle inhibitors
+const QVector<Inhibitor> kDBusInhibitors = {
```
Swiftb0y, line 169: No need for a QVector, have you tried a `std::array` instead?

src/util/screensaver.cpp
Swiftb0y: Ah right I see, you're relying on the fact that const only applies to the object pointed to, not the pointer itself (since otherwise you couldn't even set it anywhere). Then static const is fine, but maybe emphasize that it only applies to the underlying object. 

Anyways, this discussion is not too important anyways because this code would be replaced as part of the larger polymorphic rewrite I suggested.

Swiftb0y:
No worries. Yes feel free to open a zulip thread. I'll try to assist whenever possible.

src/util/screensaver.cpp
Swiftb0y: well that info is specific to the uninhibit method. So if every method becomes its own class, the classes encapsulate that. Alternatively you can look into `std::variant`. But I agree, the code duplication might get out of hand.

Swiftb0y:
I think we can postpone the polymorphic refactoring. If you just apply what I suggested so far, we can start to get this cleaned up enough to merge slowly.

src/util/screensaver.cpp
```diff
@@ -183,100 +209,122 @@ void ScreenSaverHelper::triggerUserActivity()
...
+        }
     }
-    if (!s_sendActivity) {
-        qWarning() << "Could not send activity using the registered DBus methods. " 
-        << "Errors were: " << errors << 
-        "\nWill try to use the Xlib XResetScreensaver instead.";
+
+    // No standard D-Bus inhibitors found, so check if we're running inside a
+    // Flatpak sandbox and try to use the portal instead
+    if (std::getenv("container") && std::string(std::getenv("container")) == "flatpak") {
```
Swiftb0y, line 252: don't 
```suggestion
    char* container = std::getenv("container");
    if (container != nullptr && strcmp(container, "flatpak") == 0) {
```

src/util/screensaver.cpp
```diff
@@ -183,100 +209,122 @@ void ScreenSaverHelper::triggerUserActivity()
...
-                    + reply.error().message();
+                qWarning() << "Inhibit for"
+                           << inhibitor.interface << "failed: "
+                           << reply.error().message();
             }
-        } 
+        } else {
+            qDebug() << "Interface" << inhibitor.interface << "is not valid";
+        }
     }
```
Swiftb0y, line 248: this may cause logspam. It'd be nice if this was only tested once on startup in a follow up PR.

Swiftb0y (commented):
slight codestyle nitpicks. In a future iteration it would be nice to see the if-elsing replaced by an inheritance-based approach. Nevertheless, this is already a big improvement over the previous code. Thank you.

Swiftb0y:
I'm sorry for forgetting about this. Feel free to ping me in the future

src/util/screensaver.cpp
cr7pt0gr4ph7: I'd go one step further: Why not encapsulate that into an `isRunningAsFlatpak()` utility method for clarity? Either in this file, or mabe even as `mixxx::EnvironmentHelper::isRunningAsFlatpak()` in a new `src/util/environmenthelper.h` file.

djantti: Thank you, that is an excellent idea! A simple bool function should be a good fit for this and will make the code easier to read.

src/util/screensaver.cpp
djantti: I agree, that last part (lines 245 - 247) can be especially spammy and it's not super useful either. Let's just remove it. There's already a message if inhibit was successful / unsuccessful or if no supported inhibitors were found.

PR #13073, opened 2024-04-07
LV Mix EQ: Fix pops when enabling

### LINKED ISSUE #13055: pop when enabling Bessel8 ISO in effect unit
### Bug Description

* mix mode: Dry/Wet
* fully Wet
* no effect before or after

Occurs only when toggling ON.
Notable especially if this exact same iso is used as EQ for the playing deck and High and Mid are fully killed (with knob) both in the mixer and the standard effect unit.

No buffer underruns.
Doesn't matter which keylock engine is used.

Loudness seems to vary. When I tried to reproduce after start, the first pop was rather quiet, on further toggling pops where significantly louder but still with some variation in loudness.

FWIW pops also occur when changing the keylock engine.

personal branch based on main 2.5-alpha, no patches related to effects.

### Version

2.5-alpha

### OS

_No response_

daschuer:
This fixes #13055

### DIFF
--- a/src/effects/backends/builtin/lvmixeqbase.h
+++ b/src/effects/backends/builtin/lvmixeqbase.h
@@ -22,8 +22,8 @@ class LVMixEQEffectGroupState : public EffectState {
   public:
     explicit LVMixEQEffectGroupState(const mixxx::EngineParameters& engineParameters)
             : EffectState(engineParameters),
-              m_oldLow(1.0),
-              m_oldMid(1.0),
+              m_oldLow(0.0),
+              m_oldMid(0.0),
               m_oldHigh(1.0),
               m_rampHoldOff(LVMixEQEffectGroupStateConstants::kRampDone),
               m_oldSampleRate(engineParameters.sampleRate()),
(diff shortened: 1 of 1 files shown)


ronso0:
I have almost no clue what's going on here.
Who of @mixxxdj/developers can review this?
Also, some comments would be helpful.

Can this huge commented code block be removed?

daschuer:
Ok, I can confirm that the pop can only be heared by a chance. I happens only after a toggle cycle of enabling and disabling the EQ as effect. 
After the patch the seep slope shown in the bug with mid kill enabled iwhile enabling the effect is ramped.

src/effects/backends/builtin/lvmixeqbase.h
```diff
@@ -22,8 +22,8 @@ class LVMixEQEffectGroupState : public EffectState {
   public:
     explicit LVMixEQEffectGroupState(const mixxx::EngineParameters& engineParameters)
             : EffectState(engineParameters),
-              m_oldLow(1.0),
-              m_oldMid(1.0),
+              m_oldLow(0.0),
+              m_oldMid(0.0),
```
daschuer, line 26: In our LV mix, the high part is carrying the dry stream. 
So this is an obvious fix to ramp from dry after initializing.

ywwg (approved):

ronso0:
I assume it's correct that this is assigned to 2.4.1, please rebase so we can merge it.

daschuer:
Done.

ronso0:
Thanks for the quick fix!

PR #13074, opened 2024-04-08
Tooltips: immediatly show/hide keyboard hints when option is toggled

ronso0:
Fixes #9814

Keyboard shortcuts are removed / added to tooltips when shortcuts are toggled, which previously required a restart
<sub>(kbd mapping was read by CoreServices once on initialization, if shortcuts disabled = no mapping = LegacySkinParser didn't append the shortcuts to the tooltip).</sub>

For this the entire shortcuts strings is stored by the widget and the base tooltip is reconstructed each time keyboard is toggled.
I think this refactoring might also speed up skin loading a little since the shortcuts are sent to each widget as a whole, not line by line (incl. setTooltip() each time).

This is a very low-level fix IMO, but it creates conflicts for the huge, yet unreviewed Keyboard controller PR #1746.
(I took a brief look and noticed I'm not the right one to do that)
In case someone starts reviewing that I'll help fixing conflicts, of course.

### DIFF
--- a/src/controllers/keyboard/keyboardeventfilter.cpp
+++ b/src/controllers/keyboard/keyboardeventfilter.cpp
@@ -6,12 +6,14 @@
 
 #include "moc_keyboardeventfilter.cpp"
 #include "util/cmdlineargs.h"
+#include "widget/wbasewidget.h"
 
 KeyboardEventFilter::KeyboardEventFilter(ConfigObject<ConfigValueKbd>* pKbdConfigObject,
         QObject* parent,
         const char* name)
         : QObject(parent),
-          m_pKbdConfigObject(nullptr) {
+          m_pKbdConfigObject(nullptr),
+          m_enabled(false) {
     setObjectName(name);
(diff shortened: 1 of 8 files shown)


daschuer:
I think https://github.com/mixxxdj/mixxx/pull/1746 is a dead end anyway. There are some functions we can reuse, but I would prefer to reimplement it just like a midi controller with a xml and a js file.

daschuer:
The code looks good, but needs some testing.

ronso0:
(btw I notice I pushed an intermediate state of an experimantal branch here, hence the strange branch name)

ronso0:
Oh, shortcuts are still processed even if Off.
Will fix it.

ronso0:
and some cleanup is required, too

ronso0:
Done.
Might as well go to 2.4, but should not hold up 2.4.1

ronso0:
#13082 is the complete feature, so I put this in hold.

Issue #11915, opened 2023-09-03
Waveform combobox looks cluttered 


daschuer:
### Bug Description

See: 
![grafik](https://github.com/mixxxdj/mixxx/assets/1777442/fafefdf7-18fa-487a-8ca8-04474fdd8952)

There are to many words in braces 

The non accelerated types are IMHO not legacy. 

How is the state of discussion to drop legacy GLSL types? 
Is "deprecated" a better word for the types that will be removed in the next version?

### Version

2.4-beta

ronso0:
We could group the types by (meta)type, and add plain separators, or non-selectable labels (e.g. "Deprecated:", "CPU-accelerated:")

Holzhaus:
Can't we only losz the different waveforms (RGB, HSV, Filtered,...) and determine the proper rendering variant automatically depending on what is available? Most users have no idea what to select anyway.

daschuer:
For me, a minimal 2.4.0 solution is to get rid of the double (..) works:
RGB (GLSL)
RGB 
RGB (legacy GL)
RGB (legacy GLSL)

The grouping idea sounds appealing. I would call CPU "accelerated" though. We may relpace GLSL by accelerated instead.

Swiftb0y:
We're discussing polishing something that will be unnecessary by 2.5 anyways. Lets just keep as is to get the release out. This issue will resolve itself with the removal of the GL and GLSL types soon anyways.

Swiftb0y:
sure...

ronso0:
As noted by others before, there are still way too many choices for beginners, and even advanced users I'd say. 🤪
A '?' hyperlink to a manual chapter with some (basic) explanation may be helpful (like the one for the Cue mode in Pref -> Decks).

Not sure if it's worth the effort for 2.4.x if we drop legacy waveforms in 2.5 anyway.

daschuer:
The mandatory part of this issue is now fixed. Do we want to close this or just remove or change the milestone?

ronso0:
IMO polishing the list can happen anytime, so let's move it to 2.4.1 in case someone wants to take care of it.
(just don't want to dump the ideas I posted)

acolombier:
> RGB (GLSL)
> RGB
> RGB (legacy GL)
> RGB (legacy GLSL)

Are we okay to only display the "best" rendering backend available so we can remove the `(..)` entirely? Perhaps list them all if `--developer` is passed, like for the `VSyncTest`?

[...]

Issue #6428, opened 2022-08-22
Reduce number of waveform options


mixxxbot:
Reported by: **[ywwg](https://github.com/ywwg)**
Date: 2012-05-11T15:22:52Z
Status: Confirmed
Importance: Low
Launchpad Issue: [lp998096](https://bugs.launchpad.net/bugs/998096)
Tags: usability, waveform

---

I love the new waveforms, but the preferences window for them is ugly and over-engineered.  The number of bugs for various individual waveform implementations is also proliferating so reducing the options here will also reduce some of those issues.  

I understand that many of these options allow users to fine-tune the waveform for their individual machine's capability, but I think we need to better balance the interests of debugging and the interest in not scaring our users.  Most DJs don't know what OpenGL is, or why they'd want to use shaders or not... So the basic waveform settings have to be easily-grasped.  Further complexity, for debugging, should be hidden or reduced.

In my ideal world there would be a single waveform dropdown, and it would contain:

* Off
* Simple
* Complex
* Complex (Hardware accelerated)

(where "hardware accelerated" means opengl)

Simple would be our simplest waveform running at a low fps, like 15fps.
Complex would be a pure-QT filtered waveform running at 30fps. This is our "compatibility mode" for fast systems that have graphics card issues with GL.
Complex (hardware accelerated) would our GL implementation, also running at 30.

As for the options, some of them are useful for debugging and development, but some seem unnecessary to me.

Debugging/Advanced elements:

Having the actual fps is ok for development but unnecessary for most users.  If it's running slow they will drop down to Simple.  I've been told some people like to boost this to 60 for ultra-smoothness.

99% of the time I don't care what my opengl status is. 

Superfluous elements:

I think "visual gain" is unnecessary, and might have mostly been a workaround for some bugs in the analyzing code.  I suggest we drop it completely.  Perhaps we could replace it with an option to either fill the height with the waveform, or leave the height as a function of the gain.

We should just pick a default zoom.  If the user changes the zoom, save that setting as the new default.

I'm not sure why waveform zooms wouldn't be synchronized.

RJ  suggested an "advanced" tab... maybe we could have something like:

[ Waveform Dropdown ]
[+] Advanced

So the Advanced options are in a twirl-down section, not a tab, but they are hidden by default.

acolombier:
Following suggestion from Uwe here and in #11615, could we reduce the `Waveform type` to only provide the type in the combobox, that is
- Empty
- Filtered
- HSV
- RGB
[...]

PR #13075, opened 2024-04-08
3-band waveform with moving average

acolombier:
This adds a new type of waveform, using 3-band EQ with overlapped. It also includes a moving average feature to allow adjusting the smoothness of the Waveform.

Here is a preview with different genre of music (Hardcore, Funk and Symphonic):
(Zoomed waveforms)
image
("Normal" look)
image
(Using a custom coloring)
image

The custom coloring used is the following
```xml
      <SetVariable name="SignalHighColor">#aff1eed8</SetVariable>
      <SetVariable name="SignalMidColor">#cfb26606</SetVariable>
      <SetVariable name="SignalLowColor">#ff2154d7</SetVariable>
```

Look with **Deere** skin
image
Look with **Shade** skin
image
Look with **Tango** skin
image

### DIFF
--- a/res/skins/LateNight/skin.xml
+++ b/res/skins/LateNight/skin.xml
@@ -154,8 +154,6 @@
       <SetVariable name="SignalColor_12">#d9b28c</SetVariable>
       <SetVariable name="SignalColor_34">#7bc6c3</SetVariable>
 
-      <!-- Only named colors are accepted, see w3 color chart
-          https://www.w3.org/wiki/CSS/Properties/color/keywords -->
       <SetVariable name="SignalHighColor">mediumblue</SetVariable><!-- #0000cd -->
       <SetVariable name="SignalMidColor">darkgreen</SetVariable><!-- #006400 -->
       <SetVariable name="SignalLowColor">orangered</SetVariable><!-- #ff4500 -->
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1517,6 +1517,7 @@ if(QOPENGL)
     src/waveform/renderers/allshader/waveformrendererfiltered.cpp
     src/waveform/renderers/allshader/waveformrendererhsv.cpp
     src/waveform/renderers/allshader/waveformrendererlrrgb.cpp
+    src/waveform/renderers/allshader/waveformrenderer3band.cpp
     src/waveform/renderers/allshader/waveformrendererpreroll.cpp
     src/waveform/renderers/allshader/waveformrendererrgb.cpp
     src/waveform/renderers/allshader/waveformrenderersignalbase.cpp
@@ -1526,6 +1527,7 @@ if(QOPENGL)
     src/waveform/widgets/allshader/filteredwaveformwidget.cpp
     src/waveform/widgets/allshader/hsvwaveformwidget.cpp
     src/waveform/widgets/allshader/lrrgbwaveformwidget.cpp
+    src/waveform/widgets/allshader/3bandwaveformwidget.cpp
     src/waveform/widgets/allshader/rgbwaveformwidget.cpp
     src/waveform/widgets/allshader/simplewaveformwidget.cpp
     src/waveform/widgets/allshader/waveformwidget.cpp
--- a/res/skins/LateNight/waveform.xml
+++ b/res/skins/LateNight/waveform.xml
@@ -19,12 +19,21 @@
             <BgColor><Variable name="BgColorWaveform"/></BgColor>
             <SignalColor><Variable name="SignalColor"/></SignalColor>
             <!-- Default colors are used for high/mid/low: blue/green/red -->
-            <SignalHighColor></SignalHighColor>
-            <SignalMidColor></SignalMidColor>
-            <SignalLowColor></SignalLowColor>
+            <SignalHighColor><Variable name="SignalHighColor"/></SignalHighColor>
+            <SignalMidColor><Variable name="SignalMidColor"/></SignalMidColor>
+            <SignalLowColor><Variable name="SignalLowColor"/></SignalLowColor>
             <SignalRGBHighColor></SignalRGBHighColor>
             <SignalRGBMidColor></SignalRGBMidColor>
             <SignalRGBLowColor></SignalRGBLowColor>
+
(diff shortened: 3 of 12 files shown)


res/skins/LateNight/skin.xml
```diff
@@ -154,11 +154,9 @@
       <SetVariable name="SignalColor_12">#d9b28c</SetVariable>
       <SetVariable name="SignalColor_34">#7bc6c3</SetVariable>
 
-      <!-- Only named colors are accepted, see w3 color chart
-          https://www.w3.org/wiki/CSS/Properties/color/keywords -->
```
acolombier, line 158: That doesn't seem to be true anymore

aqw42:
There is already 13 (13 !) waveform visualization type.

As a user I couldn't figure out what were the usage of these

As a developer I ask myself "what need does this new one (which looks great) covers ?"

ronso0:
> There is already 13 (13 !) waveform visualization type.
> 
> As a user I couldn't figure out what were the usage of these

True. We're aware of this, see #11915 and #6428

acolombier:
> As a user I couldn't figure out what were the usage of these

The usage is that you can have a lot of flexibility in how the waveform looks. I belive many of the "Legacy" are meant to be removed.

> As a developer I ask myself "what need does this new one (which looks great) covers ?"

This new one provides the 3-band waveform look, similar to how it looks on other mixing software.

aqw42:
> This new one provides the 3-band waveform look, similar to how it looks on other mixing software.
Okay, thanks for the answer !
So I understood well, it's an alternative to the RGB color addition look, with three superposed waves instead ?

@ronso0 Do you know if someone is working on this ?
If not, I can take look to this issue and think about a more user-oriented naming scheme

acolombier:
> So I understood well, it's an alternative to the RGB colour addition look, with three superposed waves instead ?

Kinda. The RGB one will draw a single waveform and colour each "sample" with a gradient of three colours (one per EQ gain). 
This one will draw three overlapping waveforms for each EQ (with some gain adjustment to reduce the overlap) with their one static colour. The advantage are the following:
- Easier to see a specific EQ level over time (low, mid, hid, as in the traditional RGB, you get a blend of the 3 colour)
- Colouring is predictable and constant - If your skins is using a limited colour set (like Deere or Tango), this ensures that the waveform only use 3 colours and not the whole spectrum within the colour space defined by these 3 RGB values

daschuer:
This one looks nice. 

> As a user I couldn't figure out what were the usage of these

In Mixxx we try to balance the first time user experience with our slogan "DJ your way" which allows users to modify every detail to be a perfect match for the personal use case. 

@acolombier how does this PR fit to the latest discussion around waveform preferences?

daschuer:
Conflicts have developed. I can Imagine that this replaces the filtered types?

acolombier:
> @acolombier how does this PR fit to the latest discussion around waveform preferences?

Very good question... I'm actually not quite sure. I think it is indeed "extending" the filtered or stacked type, via a waveform option, so probably need some more work, but not till #13220 is merged.
Ideally, I think this should extend the stacked type and I would suggest a rename to `3-Band` which I think is a m,ore common name for this type of visual. 
I'll bring the topic on [tonight's catchup](https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/Future.20of.20Waveform/near/440566548).

acolombier:
Putting as draft till the blockers have been merged

Swiftb0y:
> Ideally, I think this should extend the stacked type and I would suggest a rename to 3-Band which I think is a m,ore common name for this type of visual.

I think a better solution would be the introduction of some sort of waveform-data preprocessing, allowing the user to choose from a number of algorithms that distort the data before its sent to each renderer, that way the feature is independent from each renderer (= automatically applies to all of them) and is generic enough to easily implement similar features such as the non-linear rekordbox scaling (at least approximately; don't know how RB actually generates them).

acolombier:
The problem is that this would go against the current effort of simplifying all these many options which are just confusing the user. 
Currently, the waveform data gathered is meant to be agnostic to the visual used. I'm not saying that this is the best way to do things, but changing that would have a significant amount of risk and require a huge effort to manage efficiently every aspect. 
AFAIK I can see from the original design of the waveform data, it was meant to gather both peak and RMS value, but I believe it only gathers peak, so perhaps that could be changed? 

> that way the feature is independent from each renderer (= automatically applies to all of them)

I think this simpler said that done. At least from the work @m0dB and I did, it looks like this wouldn't be this simple, but happy to demonstrated wrong ;)

Swiftb0y:
I'll try to keep this in the back of my mind and whip up a PoC soon ;)

TheButlah:
This is so awesome. Hopefully a stacked 3-band waveform lands sometime soon. As a user coming from other proprietary mixing software, all the RGB options are , at least for me, hard to understand and basically don't communicate anything effectively. Having the highs mids and lows stacked on top makes the information easier to understand

bytenugget:
With some small changes in one of the skin files, you can already have a 3 band waveform. Even without this pull.

I've been trying to use the 3band waveform of this pull, but I went back to using "Filtered (GLSL)", but with the different colors. I would be really happy to have a 3band waveform like I had it in rekordbox. But this one isn't really that, apart from the coloring. I'm pretty sure the rekordbox waveform does more than just applying a rolling average. I think it is using additional analyzer data to pronounce kicks and snares specifically.

Would be great if someone could figure out how that is done, and recreate it. It makes beat-gridding soo much easier.

acolombier:
Thanks for your feedback!

> hard to understand and basically don't communicate anything effectively

This was also the same problem I have with the current waveform design

> I'm pretty sure the rekordbox waveform does more than just applying a rolling average. I think it is using additional analyzer data to pronounce kicks and snares specifically.

That's right - AFAIU, RekordBox/Serato use RMS instead of MAX for the waveform sample. This is something Mixxx kinda support, but the decision was made to only keep MAX into the analyser file. There is multiple reason for that:
- Performance - Max is faster than RMS
- Space - Adding RMS information would effectively double the analyzes cache size

I might try and submit a setting to use RMS instead of MAX

PR #13076, opened 2024-04-09
add `hotcue_X_indicator` control, is 1.0 if playposition is at hotcue -- alternative to #12951

ronso0:
alternative to #12951, see https://github.com/mixxxdj/mixxx/pull/12951#issuecomment-2045022685

Just a little helper that allows having callbacks for when the playposition crosses a hotcue / loopcue.
For example:
* always trigger the fog machine on hotcue5 via MIDI_4_Lights
* auto-censor certain spots in _all_ tracks by making hotcue17 trigger a certain sample / engage revere-roll for 1 sec
* flash controller hotcue buttons if deck is stopped at a hotcue
* ...

### Description
Basically, it's something like `beat_active` and `cue_indicator`: it is activated (1.0) if
* the play position is on the hotcue
(when stopped or when the buffer's new position is coincidentally at the hotcue position)
* the play position crossed the hotcue since the last update
(while playing forward or backward and while scratching, **not** when seeking over a hotcue)

_`beat_active` is 1.0 if the playposition is in a certain **range** around a beat marker_
_`cue_indicator`_ is 1.0 (maybe flashing) if the play position is **exactly** at the Cue (unlikely when playing)_

### Improvements?
Integrating it into `hotcue_X_status` would allow GUI feedback via WHotcueButton, it'd be just two more states that could be styled rather easily. #9760 
For now this is separate because integration/extension seems a bit cumbersome.

### Nice to have
- [ ] tests that covers looping (hotcue slight before and at loop_out, slightly after loop_in)

### Proposal for testing
* load this mapping to the MIDI Through controller https://gist.github.com/ronso0/2dd322622ddac2590020ce9b2af5d695
* load a track to deck1, set hotcues 1-2-3 with a distance of ~10s
* load a track to deck2
* both decks stoped
* Test1:
   * play from start
   * jump to hotcue 1
   * jump to hotcue 3
   * == deck2 should **not** start
* Test2:
   * play from start
   * jump to hotcue 1
   * jump to hotcue 2
   * == deck2 should start
* Test1:
   * stopped
   * jump to hotcue 3
   * scratch backwards over hotcue 2
   * == deck2 should start

### DIFF
--- a/src/engine/controls/cuecontrol.cpp
+++ b/src/engine/controls/cuecontrol.cpp
@@ -500,6 +500,7 @@ void CueControl::trackLoaded(TrackPointer pNewTrack) {
         setHotcueFocusIndex(Cue::kNoHotCue);
         m_pLoadedTrack.reset();
         m_usedSeekOnLoadPosition.setValue(mixxx::audio::kStartFramePos);
+        m_prevPosition.setValue(mixxx::audio::kStartFramePos);
     }
 
     if (!pNewTrack) {
@@ -2178,11 +2179,44 @@ void CueControl::updateIndicators() {
             m_pCueIndicator->setBlinkValue(ControlIndicator::ON);
         }
     }
+
+    // Update hotcue indicators: activate the indicator if we are at the cue pos
+    // or if we passed it since the last update, else deactivate.
+    const auto prevPos = m_prevPosition.getValue();
+    const auto currPos = frameInfo().currentPosition;
+    VERIFY_OR_DEBUG_ASSERT(prevPos.isValid()) {
+        m_prevPosition.setValue(currPos);
+        return;
+    }
+
+    for (const auto& pControl : std::as_const(m_hotcueControls)) {
+        const auto cuePos = pControl->getPosition();
+        if (!cuePos.isValid()) {
+            continue;
+        }
+        if ((cuePos >= prevPos && cuePos <= currPos) ||     // forward
+                (cuePos <= prevPos && cuePos >= currPos)) { // reverse
+            pControl->setIndicator(1.0);
+        } else {
+            pControl->setIndicator(0.0);
+        }
+    }
+    m_prevPosition.setValue(currPos);
 }
 
 void CueControl::resetIndicators() {
     m_pCueIndicator->setBlinkValue(ControlIndicator::OFF);
     m_pPlayIndicator->setBlinkValue(ControlIndicator::OFF);
+    for (const auto& pControl : std::as_const(m_hotcueControls)) {
+        pControl->setIndicator(0.0);
+    }
+}
+
+/// Store the position so we can do a correct range check when updating the
+/// hotcue indicators. Otherwise m_prevPosition would be the position from before
+/// the seek and we'd falsely activate indicators.
+void CueControl::notifySeek(mixxx::audio::FramePos position) {
+    m_prevPosition.setValue(position);
 }
 
 CueControl::TrackAt CueControl::getTrackAt() const {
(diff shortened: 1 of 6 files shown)


src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2172,72 @@ void CueControl::updateIndicators() {
...
+        m_prevPosition.setValue(currPos);
+        return;
+    }
+
+    // These are used to check whether we passed the cue while the play position
+    // was wrapped around (loop of full-track repeat).
+    // Note: notifyWrapAround() guarantees that in/out are valid if count > 0.
+    auto loopOut = m_jumpPos;
+    auto loopIn = m_targetPos;
+    bool jump = m_wrapAroundCount == 1;
```
daschuer, line 2190: I do not understand the "jump" Issue. Why is that only exact 1 and not 2?

ronso0: By "jump" I mean one regular seek loop out -> in. Usually the buffered samples don't cover the entire loop. In case of a very short loop the current position _might_ be after the previous position (= overlap of [in->curr] and [prev-out]), but we don't know for sure (unless we don another comparison, curr > prev¹). 
With count 2 we know for sure that the entire loop range was read and therefore it suffices to check whether the cue is inside the loop.

¹I'll check if that makes the situation even easier.

ronso0: > ¹I'll check if that makes the situation even easier.

Yes, another check to detect `fullLoop` (one jump + overlap case) avoids the two comparisons for that case _per cue_.
But the code needs more explanation then (IMO).

Please check with the last commit if that makes sense.

src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2172,92 @@ void CueControl::updateIndicators() {
...
+        // This reduces the jump/fullLoop comparison cases to one
+        std::swap(loopOut, loopIn);
+        reverse = true;
+    }
+    // We jumped once from out to in (or vice versa).
+    bool jump = m_wrapAroundCount == 1;
+    // That means all loop samples were read and we just check if the cue is
+    // inside the loop.
+    bool fullLoop = m_wrapAroundCount > 1 ||
+            (jump && (prevPos == currPos || reverse == (prevPos < currPos)));
```
daschuer, line 2216: We have two cases:
```
P>>|>>>>>>>|
    >>>C           
```
reverse = false 
prevPos < currPos = true 

```
|<<<<<<|<<<<P|
   C<<<              
```
reverse = true
prevPos < currPos = false

src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2172,92 @@ void CueControl::updateIndicators() {
...
+    auto loopIn = m_targetPos;
+    bool reverse = false;
+    if (m_wrapAroundCount > 0 && loopOut < loopIn) {
+        // Reverse loop: swap the two so loopIn/Out are always in/out positions.
+        // This reduces the jump/fullLoop comparison cases to one
+        std::swap(loopOut, loopIn);
+        reverse = true;
+    }
+    // We jumped once from out to in (or vice versa).
+    bool jump = m_wrapAroundCount == 1;
```
daschuer, line 2212: how about
```suggestion
    bool jumped_once = m_wrapAroundCount == 1;
```

src/engine/controls/cuecontrol.cpp
daschuer, line 2216: is it: 
```suggestion
            (jump && (prevPos == currPos || reverse == (prevPos > currPos)));
```

src/engine/controls/cuecontrol.cpp
```diff
@@ -2168,11 +2172,92 @@ void CueControl::updateIndicators() {
...
+        if (fullLoop) { // is cue inside loop?
+            activate = cuePos >= loopIn && cuePos < loopOut;
+        } else if (jump) { // is cue in the range before or after the jump?
+            activate = (cuePos >= prevPos && cuePos < loopOut) ||
+                    // (cuePos >= currPos && cuePos < loopOut) reverse
+                    (cuePos >= loopIn && cuePos <= currPos);
+            //      (cuePos >= loopIn && cuePos <= prevPos) reverse
+        } else { // is cue between previous and current position?
+            activate = cuePos >= prevPos && cuePos <= currPos;
+        }
```
daschuer, line 2241: You did not cover the range where you have entered the loop.

daschuer (changes requested):

daschuer:
I am afraid you reverse transposing does not work well because >= issue at the boundaries. 

I think it is easier to maintain if you check each case independent using a helper that handles reverse hasPassed() or such. 
We have these case: 
Wrap = 0 hasPassed(P,C,cues)
Wrap= 1 hasPassed(P,Trigger,cues) || hasPassed(Target,C,cues) 
Wrap > 1 hasPassed(P,C,cues) || hasPassed(Trigger,Target,cues)

PR #13066, opened 2024-04-07
Track colors controls update

aqw42:
Implements #12905 
Follows #2541 

Tested on debian Bookworm.
Tested for Channel1, Channel2, with "track_color_next" & "track_color_prev".

When the button mapped to `[Channel1], track_color_next` is pressed, the color of the loaded track switches to the next color in the palette. The change is visible in the library view. 

_Hi there !
First contribution, I look forward to do more.
Please tell me if anything's wrong / to improve.
I couldn't find how to add these new config keys in the documentation, is it automatic, or handled by core developers only ?_

PR #2541, opened 2020-03-08
Allow changing track colors from controller

Holzhaus:
Similar to what #2384 did for hotcue colors, this allows changing the current track color from a controller script. Since most controllers lack dedicated controls for this, I mapped this to "press and hold shift + press and hold browse know + turn browse knob" on my Roland DJ-505.

If someone has a better idea how to make this accessible on controllers, let me know ;-)

PR #13077, opened 2024-04-09
Library controls color selector

aqw42:
Add `[Library], track_color_selector`

Follows #13066 and #2541 

Tested on Debian :
-> Using developer mode, messing with the controls values
-> Using a MIDI controller, mapping each of the controls to a push button or a rotary encoder
Everything worked fine.

New controls list (note to myself for the future manual addition)
- [ChannelN], track_color_prev - push button
- [ChannelN], track_color_next - push button
- [ChannelN], track_color_selector - encoder
- [Library], track_color_prev - push button
- [Library], track_color_next - push button
- [Library], track_color_selector - encoder <-- new


src/library/librarycontrol.cpp
```diff
@@ -311,14 +311,29 @@ LibraryControl::LibraryControl(Library* pLibrary)
...
+    m_pTrackColorSelector = std::make_unique<ControlEncoder>(
+            ConfigKey("[Library]", "track_color_selector"), false);
     connect(m_pTrackColorPrev.get(),
             &ControlPushButton::valueChanged,
             this,
-            &LibraryControl::slotTrackColorPrev);
+            [this](double value) {
+                if (value > 0)
+                    LibraryControl::slotTrackColorSelector(-1);
+            });
```
ronso0, line 322: Oh, I didn't notice the clang-tidy warnign in #13066 
> :warning:  statement should be inside braces [readability-braces-around-statements]

Please use braces here and also fix it for the per-track implementation.

ronso0 (changes requested):

src/library/librarycontrol.cpp
aqw42: done

src/library/librarycontrol.cpp
```diff
@@ -1059,30 +1076,23 @@ void LibraryControl::slotDecrementFontSize(double v) {
...
-        if (!pActiveView) {
-            return;
+    while (steps != 0) {
+        if (steps > 0) {
+            pActiveView->assignNextTrackColor();
+            steps--;
+        } else {
+            pActiveView->assignPreviousTrackColor();
+            steps++;
         }
```
Swiftb0y, line 1096: Is this possible without the while? It's not very efficient and could cause visual artifacts.

aqw42: Thanks for your review.
colorPalette.nextColor() only handles color codes, the actual QT event is thrown with track->setColor().
So I think no visual artifacts or performance issue can be caused by this.
Another code using this have been push last week (see https://github.com/mixxxdj/mixxx/pull/13066)

ronso0: I think @Swiftb0y is referring to the Track::colorUpdated(color) signal that is fired for each color change whic causes repaint calls everywhere that track (or many tracks) is loaded (tracks view, decks), and with the `selector` color changes can indeed occur at high rates.
Didn't think of that during my previous review :+1:

ronso0: So ideally the color selector would be moved to the track itself, find the desired color and assign it right away.
IIUC that only refers to the situation when we have select steps buffered, e.g. from MIDI, and LibraryControl is going to do one by one. Because 'quick' color switch may also happen when triggering color_next/_prev quickly.

ronso0: > causes repaint calls everywhere that track (or many tracks)

okay, `WTrackTableView::assignNextTrackColor` acts only on the first selected track.

Swiftb0y: > causes repaint calls everywhere

Yes. thats what I was referring to. 

> IIUC that only refers to the situation when we have select steps buffered,

That seems like an ugly workaround. Ideally the API would just be changed so you can just set n'th next color directly. 

> from MIDI, and LibraryControl is going to do one by one.

If the user quickly cycles through colors by pressing the next/prev button quickly, thats fine. Especially since there is quite a lot of delay between the individual presses (which is not true for a programmed loop).

aqw42: Okay, I was looking at the wrong patch when I first replied (the one in basetrackplayer)
So I updated this one following your advice, what do you think ?

src/widget/wtracktableview.cpp
```diff
@@ -421,28 +421,17 @@ void WTrackTableView::assignPreviousTrackColor() {
...
-        pTrack->setColor(colorPalette.nextColor(color));
+        while (steps != 0) {
+            if (steps > 0) {
+                color = colorPalette.nextColor(color);
+                steps--;
+            } else {
+                color = colorPalette.previousColor(color);
+                steps++;
+            }
+        }
```
Swiftb0y, line 433: this solution still does a lot of searching. we can go one step further 
```suggestion
        const int currentColorIdx = colorPalette.indexOf(color);
        const int newColorIndex = rem_euclid(currentColorIdx + steps, colorPalette.size());
        color = colorPalette.at(newColorIndex);
```
`rem_euclid` is a euclidean remainder/modulo which you'll have to implement in util/math.h`.

You might also want to consider to implement this as `nextNthColor` in `ColorPalette` directly.

aqw42: Good suggestion !
But i'm not sure the modulo efficiency is relevant here

Swiftb0y: Here probably not, no, but if we add it as a general function to math, we should still make sure its within reasonable limits.

ronso0: What about simply moving this to colorpalette so it is available elsewhere?
`ColorPalette::getNthColor()`

aqw42: I simply moved the while loop the getNthColor, as suggested by @ronso0 because there are too many individual cases that are already handled in nextColor and previousColor (for example index -1 is no color, and you need to cycle through it too).

I felt that trying to be overefficient here would damage readability and not gain execution time given the nature of the color change (only manual changes of a few steps can happen).

I called this function in the per deck track_color_selector too.

aqw42: Test results:
The steps maximum value is 99, min is -99. (is it hardcoded in the control definition ?)
It takes 0ms to execute and there are no visual artifacts.
I made sure we really cycle through the palette, even with big values.
Works for Library and Channel.

Swiftb0y: yup, the visual artifact problem should be solved, 0ms doesn't mean its very efficient though. At least add a TODO at the loop location please.

src/library/librarycontrol.cpp
```diff
@@ -1059,30 +1076,17 @@ void LibraryControl::slotDecrementFontSize(double v) {
...
-            return;
-        }
-        pActiveView->assignPreviousTrackColor();
-    }
-}
-
-void LibraryControl::slotTrackColorNext(double v) {
-    if (!m_pLibraryWidget) {
+    LibraryView* pActiveView = m_pLibraryWidget->getActiveView();
+    if (!pActiveView) {
```
ronso0, line 1085: FWIW I have a WIP branch that makes use of [`WLibrary::getCurrentTrackTableView()`](https://github.com/mixxxdj/mixxx/blob/9b33757a4bc1dd38a3c2ab837c20c57bb771875d/src/widget/wlibrary.cpp#L105). This allows to remove many virtual methods from libraryview.h and library features (DlgMissing etc.) which essentially just redirect calls to their own WTrackTableView (which you didn't do here, which means the color controls don't work in AutoDJ, Missing and Hidden?).

aqw42: I'm sorry, my knowledge of the codebase is clearly insufficient to understand this comment.

I cannot find the `getCurrentTrackTableView` method anywhere.
Why would AutoDJ need these features ?
How does he access them ?

Can you clarify or point me to the relevant files ?

ronso0: sorry, I was knee-deep into it when I wrote this.

Like the color select slots, many control slots in LibraryControl call methods of the 'active view'. This can be anything that inherits LibraryView: a pure WTrackTableView, a html root view (top-level item of Crates and Playlists), or some controls row + table view (Auto DJ, Recording, Analysis).

So, until now, a virtual function in libraryview.h needs to be overriden not just in WTrackTableView but in every view/feature where a certain control should be availble. "Until now" because I opened #13335 earlier today which makes use of [WLibrary::getCurrentTrackTableView()](https://github.com/mixxxdj/mixxx/blob/9b33757a4bc1dd38a3c2ab837c20c57bb771875d/src/widget/wlibrary.cpp#L105) to get rid of all the virtuals and overrides (see what I removed there).
That's why AutoDJ doesn't **need** it, but you probably want the color controls to work in the AutoDJ and Recording table, too.

> I cannot find the `getCurrentTrackTableView` method anywhere.

Probably because your branch is based on an older version of main that doesn't have it, yet, not sure.

ronso0: FWIW I've been using the Github code search a ot lately, very handy:
https://github.com/search?q=repo%3Amixxxdj%2Fmixxx%20%22searchterm%22&type=code
I have a search accelerator (shortcut) and I wrap the term in quotes " (%22) so I use search terms with whitespaces.

ronso0:
#13335 has been merged, please fix the conflicts (switch to getCurrentTrackTableView())

aqw42:
> @aqw42 do you intend to finish this, or should someone else jump in?
> 
> If yes:
> 
> > #13335 has been merged, please fix the conflicts (switch to getCurrentTrackTableView())
> 
> Thanks!

I'm on it, thanks for the reminder

src/library/librarycontrol.cpp
```diff
@@ -1109,24 +1126,17 @@ void LibraryControl::slotDecrementFontSize(double v) {
...
 
-    WTrackTableView* pTrackTableView = m_pLibraryWidget->getCurrentTrackTableView();
-    if (pTrackTableView) {
-        pTrackTableView->assignPreviousTrackColor();
-    }
-}
-
-void LibraryControl::slotTrackColorNext(double v) {
-    if (!m_pLibraryWidget || v <= 0) {
+    LibraryView* pActiveView = m_pLibraryWidget->getCurrentTrackTableView();
```
ronso0, line 1134: Please change to WTrackTableView
```suggestion
    WTrackTableView* pTrackTableView = m_pLibraryWidget->getCurrentTrackTableView();
```
and use `pTrackTableView` below.

ronso0 (changes requested):

JoergAtGithub:
Please resolve the merge conflict, otherwise the CI builds will not start.

src/widget/wtracktableview.cpp
```diff
@@ -423,22 +423,23 @@ TrackModel::SortColumnId WTrackTableView::getColumnIdFromCurrentIndex() {
...
     QModelIndex index = indices.at(0);
-    TrackPointer pTrack = trackModel->getTrack(index);
-    if (!pTrack) {
-        return;
+    TrackPointer pTrack = pTrackModel->getTrack(index);
+    if (pTrack) {
+        ColorPaletteSettings colorPaletteSettings(m_pConfig);
+        ColorPalette colorPalette = colorPaletteSettings.getTrackColorPalette();
+        mixxx::RgbColor::optional_t color = pTrack->getColor();
+        ppTrack->setColor(colorPalette.getNthColor(color, steps));
```
ronso0, line 441:
```suggestion
        pTrack->setColor(colorPalette.getNthColor(color, steps));
```

ronso0:
@aqw42 Please ammend the last fixup commit in to the 'merge branch main' commits.
All commits should build, so we can use git bisect for debugging.

So please also check all the previous commits build.
Thank you!

src/widget/wtracktableview.h
```diff
@@ -39,8 +39,7 @@ class WTrackTableView : public WLibraryTableView {
     void resizeEvent(QResizeEvent* event) override;
     void activateSelectedTrack();
     void loadSelectedTrackToGroup(const QString& group, bool play);
-    void assignNextTrackColor() override;
-    void assignPreviousTrackColor() override;
+    void selectTrackColor(int) override;
```
Swiftb0y, line 42: please give it a name so the we can easily see the purpose of the int.
```suggestion
    void selectTrackColor(int steps) override;
```

src/library/libraryview.h
```diff
@@ -38,8 +38,7 @@ class LibraryView {
     }
     /// If applicable, requests that the LibraryView changes the track color of
     /// the selected track. Does nothing otherwise.
-    virtual void assignPreviousTrackColor() {
-    }
-    virtual void assignNextTrackColor() {
+    virtual void selectTrackColor(int steps) {
+        Q_UNUSED(steps)
     }
```
Swiftb0y, line 43: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-unused
```suggestion
    virtual void selectTrackColor(int) {
    }
```

src/library/librarycontrol.cpp
```diff
@@ -337,14 +337,31 @@ LibraryControl::LibraryControl(Library* pLibrary)
...
+    m_pTrackColorSelector = std::make_unique<ControlEncoder>(
+            ConfigKey("[Library]", "track_color_selector"), false);
     connect(m_pTrackColorPrev.get(),
             &ControlPushButton::valueChanged,
             this,
-            &LibraryControl::slotTrackColorPrev);
+            [this](double value) {
+                if (value > 0) {
+                    LibraryControl::slotTrackColorSelector(-1);
+                }
```
Swiftb0y, line 348: you're calling member functions of the same class. Just do this. Otherwise it looks like you're calling a parent or static member function. 
```suggestion
                if (value > 0) {
                    slotTrackColorSelector(-1);
                }
```

src/library/librarycontrol.cpp
```diff
@@ -1115,24 +1132,17 @@ void LibraryControl::slotDecrementFontSize(double v) {
...
+    if (!pTrackTableView) {
         return;
     }
 
-    WTrackTableView* pTrackTableView = m_pLibraryWidget->getCurrentTrackTableView();
-    if (pTrackTableView) {
-        pTrackTableView->assignNextTrackColor();
+    if (steps != 0) {
+        pTrackTableView->selectTrackColor(steps);
     }
```
Swiftb0y, line 1147: you already checked above, no need to check twice.
```suggestion
    pTrackTableView->selectTrackColor(steps);
```

src/util/color/colorpalette.cpp
```diff
@@ -38,6 +38,21 @@ mixxx::RgbColor::optional_t ColorPalette::previousColor(mixxx::RgbColor::optiona
     return at(size() - 1);
 }
 
+mixxx::RgbColor::optional_t ColorPalette::getNthColor(
+        mixxx::RgbColor::optional_t color, int steps) const {
+    // TODO : Use rem_euclid modulo function instead of a loop
+    while (steps) {
```
Swiftb0y, line 44: we should address this in this PR IMO. The repeated calls to `nextColor`/`previousColor` which in turn call `indexOf` are highly suboptimal (in terms of wasted computations).

aqw42:
@Swiftb0y these comments are mostly nicklepicketing or details that we already discussion 4 month ago.
I feel that you are making me lose my time, please resolve them yourself, as you already did in your suggestions.

ronso0:
@aqw42 I don't think this is the appropriate tone, considering the delay of 4 months stems from the fact that you missed my reminder in June. (and none of the contributors, me included, gave this high prio considering more important 2.4/2.5 fixes that had to be done).

Let's fix/polish this and get it merged soonish.

Swiftb0y:
> these comments are mostly unimportant or details that we already discussed 4 month ago.

sure these comments are certainly nitpicks to a degree, but I feel like they're important enough to raise. The fact that I brought them up even though we've already discussed them should emphasize their importance as well as the fact that they have not yet been addressed. Trust me, I want to do anything but to waste your time, but largescale software development requires careful consideration of each change, otherwise the result will become a large mess that nobody wants to work on anymore.

So please bear with me when I ask for a little bit of polish. It helps you to improve your code and us all to improve mixxx. Thank you.

ronso0 (changes requested):

ronso0:
Just some cleanup..

ronso0:
@Swiftb0y Have your concerns all been addressed?

acolombier:
Friendly ping @Swiftb0y :) 

If it's good on your end, there is just a minor conflict and this can go in. Note that this could be a great pilot for our new `/softfix` command as well?

Swiftb0y:
@daschuer what did I do wrong?

Swiftb0y:
Ah, now. Took longer than expected.

Swiftb0y:
great. Seems I fucked up this one (however thats even possible....)

Swiftb0y:
So I can't fix this alone and need your help @aqw42. Make sure you haven't `git pull`ed the commits I messed up. I can't find the original commit on the remote anymore. Please check if you still have them. Do a `git log 7a5e9abd1ee44f62fc70c758b517aa8a7d1b2f2f` (thats the hash of  the commit before I messed it up). If git can show you that commit, then we can proceed. `git checkout library-controls-color-selector` and then `git reset --hard 7a5e9abd1ee44f62fc70c758b517aa8a7d1b2f2f` (notice that any unsaved or uncommitted files will be deleted). After that you can `git push --force` which should restore this PR to the state from before I my fuckup. Could you do that for me? Thank you.

Swiftb0y:
Alternatively, I can reapply the change manually. https://github.com/mixxxdj/mixxx/compare/6f6247c...7a5e9abd1ee44f62fc70c758b517aa8a7d1b2f2f

ronso0:
Yes, that diff looks good.

PR #13078, opened 2024-04-10
FindRubberband: Make sure sleefdft is present before linking it

fwcd:
I have encountered configuration errors when Sleef was built without SleefDFT (`FindSleef.cmake` seems to handle this correctly). To avoid this, this PR makes sure to only link Sleef if the DFT library is present.

In my case this happened because Homebrew Sleef (which apparently is not built with DFT) slipped into the build. While this is not ideal, we should at least make sure that we build against a working version of Sleef.

### DIFF
--- a/cmake/modules/FindSleef.cmake
+++ b/cmake/modules/FindSleef.cmake
@@ -63,7 +63,7 @@ mark_as_advanced(Sleef_LIBRARY)
 
 foreach(_component ${Sleef_FIND_COMPONENTS})
   find_library(Sleef_${_component}_LIBRARY
-    NAMES ${_component}
+    NAMES sleef${_component}
     PATHS ${PC_Sleef_LIBRARY_DIRS})
   mark_as_advanced(Sleef_${_component}_LIBRARY)
   set(ADDITIONAL_REQUIRED_VARS ${ADDITIONAL_REQUIRED_VARS} Sleef_${_component}_LIBRARY)
@@ -125,16 +125,16 @@ if(Sleef_FOUND)
   set(Sleef_INCLUDE_DIRS "${Sleef_INCLUDE_DIR}")
   set(Sleef_DEFINITIONS ${PC_Sleef_CFLAGS_OTHER})
 
-  if(Sleef_sleefdft_FOUND)
-    is_static_library(Sleef_sleefdft_IS_STATIC Sleef::sleefdft)
-    if(Sleef_sleefdft_IS_STATIC)
-      set_property(TARGET Sleef::sleefdft APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+  if(Sleef_dft_FOUND)
+    is_static_library(Sleef_dft_IS_STATIC Sleef::dft)
+    if(Sleef_dft_IS_STATIC)
+      set_property(TARGET Sleef::dft APPEND PROPERTY INTERFACE_LINK_LIBRARIES
         Sleef::sleef
       )
 
       find_package(OpenMP)
       if(OpenMP_CXX_FOUND)
-        set_property(TARGET Sleef::sleefdft APPEND PROPERTY INTERFACE_LINK_LIBRARIES
+        set_property(TARGET Sleef::dft APPEND PROPERTY INTERFACE_LINK_LIBRARIES
           OpenMP::OpenMP_CXX
         )
       endif()
--- a/cmake/modules/FindSleef.cmake
+++ b/cmake/modules/FindSleef.cmake
@@ -61,10 +61,28 @@ find_library(Sleef_LIBRARY
   PATHS ${PC_Sleef_LIBRARY_DIRS})
 mark_as_advanced(Sleef_LIBRARY)
 
-find_library(SleefDFT_LIBRARY
-  NAMES sleefdft
-  PATHS ${PC_Sleef_LIBRARY_DIRS})
-mark_as_advanced(SleefDFT_LIBRARY)
+foreach(_component ${Sleef_FIND_COMPONENTS})
+  find_library(Sleef_${_component}_LIBRARY
+    NAMES ${_component}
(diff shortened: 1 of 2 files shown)


daschuer (changes requested):
The status message is now confusing, because sleep is reporetef as found. I think we can use find_packge to explicit look for sleefdft
If this is not possible please add a status message

fwcd:
IIUC we cannot use `find_package` for that since we locate SleefDFT together with Sleef in `FindSleef.cmake` (we would need a separate `FindSleefDFT.cmake` or similar). Maybe adding a warning to `FindSleef.cmake` if Sleef was found without DFT would be appropriate?

fwcd:
I have added a warning, which should hopefully highlight the problem if someone is wondering why we don't link their Sleef:

```
CMake Warning at cmake/modules/FindSleef.cmake:145 (message):
  Sleef found at /opt/homebrew/lib/libsleef.a does not have sleefdft, this
  means static builds will not link it with Rubberband! Perhaps you meant to
  use some other version of Sleef?
```

cmake/modules/FindSleef.cmake
```diff
@@ -140,4 +140,8 @@ if(Sleef_FOUND)
         INTERFACE_INCLUDE_DIRECTORIES "${Sleef_INCLUDE_DIR}"
     )
   endif()
+
+  if(NOT TARGET Sleef::sleefdft)
+    message(WARNING "Sleef found at ${Sleef_LIBRARY} does not have sleefdft, this means static builds will not link it with Rubberband! Perhaps you meant to use some other version of Sleef?")
```
fwcd, line 145: Should I move the message to the `find_package` invocation in `FindRubberband` or keep it here? I'm a bit torn... (in practice it probably doesn't matter much since we only use Sleef for Rubberband)

daschuer:
I think the best way would be to make 
`find_package(Sleef COMPONENTS sleefdft)`

This can be done by adding the following fo FindSleef.cmake

``` 
foreach (_component ${SLEEF_FIND_COMPONENTS})
    if (${_component} STREQUAL "sleefdft")
        find_library(SleefDFT_LIBRARY
          NAMES sleefdft
          PATHS ${PC_Sleef_LIBRARY_DIRS})
        mark_as_advanced(SleefDFT_LIBRARY)
        set(ADDITIONAL_REQUIRED_VARS ${ADDITIONAL_REQUIRED_VARS} SleefDFT_LIBRARY)
    endif() 
endforeach ()

....

find_package_handle_standard_args(
  Sleef
  REQUIRED_VARS Sleef_INCLUDE_DIR Sleef_LIBRARY ${ADDITIONAL_REQUIRED_VARS}
  VERSION_VAR Sleef_VERSION)
```

fwcd:
Great suggestion, thanks! I have make the search for Sleef components a bit more generic even and dropped the 'sleef' prefix (since at least according to the project's CMakeLists [they all follow the same naming scheme](https://github.com/shibatch/sleef/blob/60e76d2bce17d278b439d9da17177c8f957a9e9b/CMakeLists.txt#L10-L14)), so `Sleef::sleefdft` becomes `Sleef::dft` (and `find_package(Sleef COMPONENTS dft)`, `Sleef_dft_FOUND` etc.)

cmake/modules/FindSleef.cmake
```diff
@@ -61,10 +61,28 @@ find_library(Sleef_LIBRARY
...
-find_library(SleefDFT_LIBRARY
-  NAMES sleefdft
-  PATHS ${PC_Sleef_LIBRARY_DIRS})
-mark_as_advanced(SleefDFT_LIBRARY)
+foreach(_component ${Sleef_FIND_COMPONENTS})
+  find_library(Sleef_${_component}_LIBRARY
+    NAMES sleef${_component}
+    PATHS ${PC_Sleef_LIBRARY_DIRS})
+  mark_as_advanced(Sleef_${_component}_LIBRARY)
+  set(ADDITIONAL_REQUIRED_VARS ${ADDITIONAL_REQUIRED_VARS} Sleef_${_component}_LIBRARY)
```
daschuer, line 69: ><PackageName>_FIND_COMPONENTS
>    List of specified components (required and optional)

so I think we need to add this: 

```suggestion
  if((Sleef_FIND_REQUIRED_${_component})
    set(ADDITIONAL_REQUIRED_VARS ${ADDITIONAL_REQUIRED_VARS} Sleef_${_component}_LIBRARY)
  endif()
```

cmake/modules/FindSleef.cmake
```diff
@@ -61,10 +61,28 @@ find_library(Sleef_LIBRARY
...
+      add_library(Sleef::${_component} UNKNOWN IMPORTED)
+      set_target_properties(Sleef::${_component}
+        PROPERTIES
+          IMPORTED_LOCATION "${Sleef_${_component}_LIBRARY}"
+          INTERFACE_COMPILE_OPTIONS "${PC_Sleef_CFLAGS_OTHER}"
+          INTERFACE_INCLUDE_DIRECTORIES "${Sleef_INCLUDE_DIR}"
+      )
+    endif()
+  else()
+    set(Sleef_${_component}_FOUND FALSE)
```
daschuer, line 83: Not sure what this means: 
> The package configuration file may set <PackageName>_FOUND to false to tell find_package that component requirements are not satisfied.

Do we also need the guard here for optional components, need to check.

daschuer (changes requested):

acolombier:
Do you still have interest in completing this @fwcd? I'll mark this as draft for now, but feel free to mark it active when this should get another look!

fwcd:
Ugh, a lot of conflicts here. Yeah, I believe it would be useful to have to avoid accidentally linking the wrong Sleef, but IIRC this also only came up in one of my more esoteric build setups where I somehow overlayed vcpkg libs with Homebrew-installed libs... I'll take a look when I find some time

PR #13080, opened 2024-04-10
CMakeLists: Always prefer OpenGL framework on macOS

fwcd:
This patch ensures that we always link the system-provided OpenGL on macOS to avoid accidentally picking the wrong implementation (e.g. X11 OpenGL if the user has [XQuartz](https://www.xquartz.org/) installed.


daschuer (approved):
LGTM, thank you.

PR #13081, opened 2024-04-10
Pioneer HID unlock PoC

Swiftb0y:
This is a quick and dirty (untested) implementation of https://swiftb0y.github.io/CDJHidProtocol/hid-analysis/handshake.html

It'd be nice if @JoergAtGithub or @acolombier could have a quick look at the HID related parts as I've never worked with HID in mixxx before and didn't have the time to properly familiarize myself with it for this.


res/controllers/Pioneer-HID-unlock-scripts.js
```diff
@@ -0,0 +1,286 @@
...
+};
+
+PioneerUnlock.seedA = [0x00, 0x00, 0xFF, 0xFF, 0x00, 0x00, 0xFF, 0xFF];
+
+PioneerUnlock.keepAliveTimer = 0;
+
+PioneerUnlock.incomingData = function(data, _length) {
+
+    const manufacturer = "NativeInstruments";
+    // TODO identify correct device properly
```
JoergAtGithub, line 199: We don't have vendor ID and product ID accessible in JS domain.

Swiftb0y: Not necessary I think. I can read that from the handshake with a little extra effort. I was just too lazy to implement it right now, I don't want the extra complexity until this has been confirmed working with a device just hardcoded.

res/controllers/Pioneer-HID-unlock.hid.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<MixxxControllerPreset schemaVersion="1" mixxxVersion="1.11+">
+    <info>
+        <name>Pioneer Hardware Unlock</name>
+        <author>Swiftb0y</author>
+        <description>"Unlocks" some functions pioneer DJ locks behind licensing, supported (but untested) devices: CDJ-2000NXS2, DJM-900NXS2, CDJ-TOUR1, DJM-TOUR1, XDJ-1000MK2, DJM-450, DJM-250MK2, DJM-750MK2, DJM-V10, XDJ-XZ </description>
+        <manual></manual>
+        <devices>
+            <product protocol="hid" vendor_id="0x1157" product_id="0x300" usage_page="0xffa0" usage="0x1" interface_number="0x3" />
```
Swiftb0y, line 9: this is just taken from some other mapping. So this definitely needs changing in order to properly work, I think.

res/controllers/Pioneer-HID-unlock-scripts.js
```diff
@@ -0,0 +1,286 @@
...
+    } else if (supertlv.type === 0x15) {
+        console.assert(view.getInt16(2) === 0x0002);
+        console.assert(supertlv.length === 2);
+        console.info("Hardware authenticated successfully");
+
+        if (PioneerUnlock.keepAliveTimer === 0) {
+            engine.stopTimer(PioneerUnlock.keepAliveTimer);
+        }
+        PioneerUnlock.keepAliveTimer = engine.beginTimer(400, () => {
+            controller.send(Array(64).fill(0));
```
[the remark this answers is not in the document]
Swiftb0y: >  Can you please provide the HID ReportDescriptor, to verify, that this is really an OutputReport

I don't have access to the hardware. When I did the reverse-engineering work a couple years ago, I just looked at the wireshark dump and worked of the raw bytes without a thorough understanding of their semantic meaning. I still have a couple dumps from that time so I can try to look it up through that. 

The protocol pioneer invented on top of HID has no concept of reports and usage pages at all IIRC (they just leave the reportID 0, until the hacked in multi-deck support for the XDJ-XZ and used the reportID for that).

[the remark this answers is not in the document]
Swiftb0y: Those captures where done through USBmon, so its probably not visible in the captures either.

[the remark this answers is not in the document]
JoergAtGithub: A HID device without HID Report Descriptor wouldn't be accepted by the HID subsystems of macOS and Windows - I'm not sure if Linux hidraw does.
Therefore there must be at at least a descriptor with one entry for each report with type and the overall number of bits/bytes of this report. If the type or number bytes does not match with the sent data, the HID subsystems of macOS and Windows would reject the send operation too.

ReportIDs are optional for an HID device, but if the device uses no ReportIDs, than the hidapi library requires 0x00 as placeholder (vaild ReportIDs must be in range of 1...255).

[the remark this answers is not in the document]
Swiftb0y: yeah. I vaguely remember the reportID being `0x00` (or I guess absent) with a length of 64 bytes. And thats it. Here's a copy of the device descriptors as captured by wireshark if you want to have a look yourself.
[Pioneer-CDJ-2000NXS2-Devicedescriptors.zip](https://github.com/mixxxdj/mixxx/files/14933217/Pioneer-CDJ-2000NXS2-Devicedescriptors.zip)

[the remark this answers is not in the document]
JoergAtGithub: DeviceDescriptors and ReportDescriptors are different entities. Here is an example how to read the report descriptor of an HID device under Linux: https://docs.kernel.org/hid/hidintro.html#parsing-hid-report-descriptors
If you have the hexdump, you can put it in this webtool to parse it: http://eleccelerator.com/usbdescreqparser/

acolombier:
Just out of curiosity, is there any legal "risk" with having these mappings as part of Mixxx distributed mapping, while they appears to bypass some licensed features?

I assume these licenses are something you unlock in RekordBox or Serato, but I'm just wondering if you had a chance to confirm there was no legal implication on walking around it on the hardware

JoergAtGithub:
Hardware is something you own if you purchase it. You can do whatever you like with it, including re-engineering and modifications. A similar case are car engines, which you can legally chip-tune to get more horsepower. The only thing the original equipment manufacturers can do, is frightening you, that you get no waranty anymore.

BTW: Similar unlocking code for PioneerDJ devices is part of the Linux kernel.

acolombier:
> Hardware is something you own if you purchase it. You can do whatever you like with it, including re-engineering and modifications. A similar case are car engines, which you can legally chip-tune to get more horsepower. 

That's not completely true. There are modern cars nowadays that are sold with hardware which you don't have access to till you pay extra licenses. Yes, you can "hack" these feature, but the manufacturer may also "brick" you (e.g on the next OTA). 

> The only thing the original equipment manufacturers can do, is frightening you, that you get no warranty any more.

I guess that's my point - would we want to offer mapping out of the box that may void the warranty on user devices without some sort of warning/acknowledgement ? 

Note that I'm only playing devil's advocate here. I'm all in to support hardware fully!

Swiftb0y:
I appreciate your concern, but I think this is pretty low risk. The worst that could happen is that an update bricks our integration again or I get sued for reverse-engineering their protocol. That's a risk I'm personally willing to take.

res/controllers/libpcon.js
```diff
@@ -0,0 +1,378 @@
...
+    }
+};
+
+pcon.DeviceSysexCode = {
+    DJM250MK2: [0x00, 0x00, 0x00, 0x17, 0x00],
+    DJM750MK2: [0x00, 0x00, 0x01, 0x0B, 0x00],
+};
+
+pcon.seedA = [0x00, 0x00, 0xFF, 0xFF, 0x00, 0x00, 0xFF, 0xFF];
+// pcon.seedA = [0, 0, 0x2c, 0xe, 0, 0, 9, 0x46]
```
acolombier, line 179: Is this still needed?

res/controllers/libpcon.js
```diff
@@ -0,0 +1,378 @@
...
+    }
+    const type = data[0];
+    const length = data[1];
+
+    console.assert(length <= data.length);
+    const value = data.slice(2, length);
+
+    console.debug(type.toString(16));
+    console.debug(length.toString(16));
+    console.debug(value);
```
acolombier, line 84: It would be great to either add some message prefix on `console.debug` so we don't get logs with a single word spam which is hard to link back to where it comes from. Alternative, we could the `controllerDebug` bool to hide this if not enabled.

res/controllers/libpcon.js
```diff
@@ -0,0 +1,378 @@
...
+        }
+    };
+
+    const manufacturer = "NativeInstruments";
+    const softProd = "Traktor Pro 2";
+
+    console.debug(pcon.debug.hexDump(data));
+    const payload = (protocol === pcon.protocol.SYSEX) ? pcon.parseHeader.sysex(data).inner : pcon.parseHeader.hid(data).inner;
+    // console.assert(view.getInt16(0) === 0x00f0);
+    // console.assert(view.getInt16(1) === 0x0001);
```
acolombier, line 279: comment cruft?

res/controllers/libpcon.js
```diff
@@ -0,0 +1,378 @@
...
+ */
+
+pcon.handleAuth = function(data, protocol) {
+
+    const send = (stage, inner) => {
+        // console.debug(`outgoing:\n${pcon.debug.hexDump(inner)}`);
+        if (protocol === pcon.protocol.HID) {
+            pcon.send.hidextended(0, 0xf0, 1, stage, inner);
+        } else {
+            // TODO don't hardcode this
```
acolombier, line 268: Is this TODO part of this PR?

acolombier (commented):
Sorry for the delay in the review.
It's looking good, just a Q on all the `console.assert`, what is expected if an assert occur? Does that abort the controller?

Swiftb0y:
> what is expected if an assert occur? Does that abort the controller?

from [the qt docs on the console API](https://doc.qt.io/qt-6/qtquick-debugging.html#console-api) it will only print a stack trace to the console. It should continue running. If not thats not a problem either because the handshake procedure will simply restart.

Swiftb0y:
This PR is very much not ready (the HID part is broken afaict) and I have no hardware to test. I'll leave it here for now, but until I get some spare time and an opportunity for an affordable CDJ or DJM, I'm not particularly comfortable with merging. It shouldn't crash mixxx, but its almost certainly not going to work. We can merge it as experimental and see if some users with the hardware respond and are willing to help.

Swiftb0y:
Thank you for the review though

deveus:
I have an XDJ-1000MK2 that I could use to test on either windows or Ubuntu 24.04 if that would help.

PR #13082, opened 2024-04-10
Keyboard: automatically reload mapping file and update tooltips + menubar accelerators

ronso0:
Yeeiij, finally!
When keyboard shortcuts are toggled On/Off, the widget tooltips are adjusted accordingly.
When the keyboard mapping file has been changed (or removed), tooltips and the mapping are rebuilt.
Same applies to **menubar** shortcuts.

As before, we look for the user's `Custom.kbd.cfg` first, then use the built-in mapping based on the selected locale.
____
Also adds a new tooltip option: Keyboard shortcuts only
____

Fixes #9814 
Update for the Manual is https://github.com/mixxxdj/manual/pull/735

### DIFF
--- a/src/controllers/keyboard/keyboardeventfilter.cpp
+++ b/src/controllers/keyboard/keyboardeventfilter.cpp
@@ -6,6 +6,8 @@
 
 #include "moc_keyboardeventfilter.cpp"
 #include "util/cmdlineargs.h"
+#include "widget/wbasewidget.h"
+#include "widget/wsearchlineedit.h"
 
 namespace {
 const QString mappingFilePath(const QString& dir, const QString& fileName) {
@@ -22,7 +24,8 @@ KeyboardEventFilter::KeyboardEventFilter(UserSettingsPointer pConfig,
 #endif
           m_pConfig(pConfig),
           m_locale(locale),
-          m_enabled(false) {
+          m_enabled(false),
+          m_autoReloader(RuntimeLoggingCategory(QStringLiteral("kbd_auto_reload"))) {
     // Get the enabled state.
     // Note: use the same default value in WMainMenuBar so that the action and
     // our bool are in sync.
@@ -31,6 +34,12 @@ KeyboardEventFilter::KeyboardEventFilter(UserSettingsPointer pConfig,
             true);
 
     createKeyboardConfig();
+
+    // For watching the currently loaded mapping file
+    connect(&m_autoReloader,
+            &AutoFileReloader::fileChanged,
+            this,
+            &KeyboardEventFilter::reloadKeyboardConfig);
 }
 
 KeyboardEventFilter::~KeyboardEventFilter() {
@@ -223,18 +232,125 @@ void KeyboardEventFilter::setEnabled(bool enabled) {
     }
     m_enabled = enabled;
     m_pConfig->setValue(ConfigKey("[Keyboard]", "Enabled"), enabled);
+    // Shortcuts may be toggled off and on again to make Mixxx discover a new
+    // Custom.kbd.cfg, so reload now.
+    // Note: the other way around (removing a loaded Custom.kbd.cfg) is covered
+    // by the auto-reloader and we'll try to load a built-in mapping then.
+    if (enabled) {
+        reloadKeyboardConfig();
+    }
+    // Update widget tooltips
+    emit shortcutsEnabled(enabled);
+}
+
+void KeyboardEventFilter::registerShortcutWidget(WBaseWidget* pWidget) {
+    m_widgets.append(pWidget);
+
+    // Tell widgets to reconstruct tooltips when option is toggled.
+    // WBaseWidget is not a QObject, need to use a lambda
+    connect(this,
+            &KeyboardEventFilter::shortcutsEnabled,
+            this,
+            [pWidget](bool enabled) {
+                pWidget->toggleKeyboardShortcutHints(enabled);
+            });
+}
+
+void KeyboardEventFilter::updateWidgetShortcuts() {
+    qDebug() << "Update widget shortcuts";
+    QStringList shortcutHints;
+    for (auto* pWidget : std::as_const(m_widgets)) {
+        QString keyString;
+        shortcutHints.clear();
+        const QList<std::pair<ConfigKey, QString>> controlsCommands =
+                pWidget->getShortcutControlsAndCommands();
+        for (const auto& [control, command] : controlsCommands) {
+            keyString = m_pKbdConfig->getValueString(control);
+            if (!keyString.isEmpty()) {
+                shortcutHints.append(buildShortcutString(keyString, command));
+            }
+        }
+        // might be empty to clear the previous tooltip
+        pWidget->setShortcutTooltip(shortcutHints.join(QStringLiteral("\n")));
+    }
+    // Update widget tooltips (WBaseWidget handles no-ops).
+    emit shortcutsEnabled(m_enabled);
+}
+
+void KeyboardEventFilter::clearWidgets() {
+    disconnect();
+    m_widgets.clear();
+    m_pSearchBar = nullptr;
+}
+
+void KeyboardEventFilter::registerSearchBar(WSearchLineEdit* pSearchBar) {
+    m_pSearchBar = pSearchBar;
+}
+
+void KeyboardEventFilter::updateSearchBarShortcuts() {
+    if (m_pSearchBar == nullptr) {
+        return;
+    }
+
+    // Translate shortcuts to native text
+    QString searchInCurrentViewShortcut =
+            localizeShortcutKeys(m_pKbdConfig->getValue(
+                    ConfigKey("[KeyboardShortcuts]",
+                            "LibraryMenu_SearchInCurrentView"),
+                    "Ctrl+f"));
+    QString searchInAllTracksShortcut =
+            localizeShortcutKeys(m_pKbdConfig->getValue(
+                    ConfigKey("[KeyboardShortcuts]",
+                            "LibraryMenu_SearchInAllTracks"),
+                    "Ctrl+Shift+F"));
+
+    m_pSearchBar->setupToolTip(searchInCurrentViewShortcut, searchInAllTracksShortcut);
+}
+
+/// Translate shortcut to native text
+QString KeyboardEventFilter::localizeShortcutKeys(const QString& shortcut) const {
+    return QKeySequence(shortcut, QKeySequence::PortableText)
+            .toString(QKeySequence::NativeText);
+}
+
+QString KeyboardEventFilter::buildShortcutString(
+        const QString& shortcut, const QString& cmd) const {
+    VERIFY_OR_DEBUG_ASSERT(!shortcut.isEmpty()) {
+        return QString();
+    }
+
+    QString shortcutTooltip(tr("Shortcut"));
+    if (!cmd.isEmpty()) {
+        shortcutTooltip += " ";
+        shortcutTooltip += cmd;
+    }
+    shortcutTooltip += ": ";
+    shortcutTooltip += localizeShortcutKeys(shortcut);
+    return shortcutTooltip;
+}
+
+void KeyboardEventFilter::reloadKeyboardConfig() {
+    createKeyboardConfig();
+    updateWidgetShortcuts();
+    updateSearchBarShortcuts();
 }
 
 void KeyboardEventFilter::createKeyboardConfig() {
-    // Read keyboard configuration and set kdbConfig object in WWidget
+    // Remove the previously watched file.
+    // Could be the user mapping has been removed and we'll need to switch
+    // to the built-in default mapping.
+    m_autoReloader.clear();
+
     // Check first in user's Mixxx directory
     QString keyboardFile = mappingFilePath(m_pConfig->getSettingsPath(), QStringLiteral("Custom"));
     if (QFile::exists(keyboardFile)) {
         qDebug() << "Found and will use custom keyboard mapping" << keyboardFile;
     } else {
         // check if a default keyboard exists
-        const QString resourcePath = m_pConfig->getResourcePath() + QStringLiteral("keyboard/");
-        keyboardFile = mappingFilePath(resourcePath, m_locale.name());
+        const QString resourcePath = m_pConfig->getResourcePath();
+        keyboardFile = QString(resourcePath).append("keyboard/");
+        keyboardFile += m_locale.name();
+        keyboardFile += ".kbd.cfg";
         if (QFile::exists(keyboardFile)) {
             qDebug() << "Found and will use default keyboard mapping" << keyboardFile;
         } else {
@@ -245,6 +361,10 @@ void KeyboardEventFilter::createKeyboardConfig() {
                 keyboardFile = "";
             }
         }
+    }
+    if (!keyboardFile.isEmpty()) {
+        // Watch the loaded file for changes.
+        m_autoReloader.addPath(keyboardFile);
     }
 
     // Read the keyboard configuration file and set m_pKbdConfig.
--- a/src/controllers/keyboard/keyboardeventfilter.cpp
+++ b/src/controllers/keyboard/keyboardeventfilter.cpp
@@ -7,16 +7,24 @@
 #include "moc_keyboardeventfilter.cpp"
 #include "util/cmdlineargs.h"
 
-KeyboardEventFilter::KeyboardEventFilter(ConfigObject<ConfigValueKbd>* pKbdConfigObject,
-        QObject* parent,
-        const char* name)
-        : QObject(parent),
+KeyboardEventFilter::KeyboardEventFilter(UserSettingsPointer pConfig,
+        const QLocale& locale,
+        QObject* pParent)
+        : QObject(pParent),
 #ifndef __APPLE__
           m_altPressedWithoutKey(false),
 #endif
-          m_pKbdConfigObject(nullptr) {
-    setObjectName(name);
-    setKeyboardConfig(pKbdConfigObject);
+          m_pConfig(pConfig),
+          m_locale(locale),
+          m_enabled(false) {
+    // Get the enabled state.
+    // Note: use the same default value in WMainMenuBar so that the action and
+    // our bool are in sync.
+    m_enabled = m_pConfig->getValue(ConfigKey(QStringLiteral("[Keyboard]"),
+                                            QStringLiteral("Enabled")),
+            true);
+
+    createKeyboardConfig();
 }
 
 KeyboardEventFilter::~KeyboardEventFilter() {
@@ -28,26 +36,31 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
         // because we might not get Key Release events.
         m_qActiveKeyList.clear();
     } else if (e->type() == QEvent::KeyPress) {
-        QKeyEvent* ke = (QKeyEvent *)e;
+        QKeyEvent* pKE = static_cast<QKeyEvent*>(e);
 
 #ifdef __APPLE__
         // On Mac OSX the nativeScanCode is empty (const 1) http://doc.qt.nokia.com/4.7/qkeyevent.html#nativeScanCode
         // We may loose the release event if a the shift key is pressed later
         // and there is character shift like "1" -> "!"
-        int keyId = ke->key();
+        int keyId = pKE->key();
 #else
-        int keyId = ke->nativeScanCode();
+        int keyId = pKE->nativeScanCode();
 #endif
 
         if (shouldSkipHeldKey(keyId)) {
             return true;
         }
 
-        QKeySequence ks = getKeySeq(ke);
+        QKeySequence ks = getKeySeq(pKE);
         if (!ks.isEmpty()) {
 #ifndef __APPLE__
             m_altPressedWithoutKey = false;
 #endif
+            // If inactive, return after logging the key event in getKeySeq()
+            if (!isEnabled()) {
+                return false;
+            }
+
             ConfigValueKbd ksv(ks);
             // Check if a shortcut is defined
             bool result = false;
@@ -55,22 +68,27 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
             for (auto it = m_keySequenceToControlHash.constFind(ksv);
                  it != m_keySequenceToControlHash.constEnd() && it.key() == ksv; ++it) {
                 const ConfigKey& configKey = it.value();
-                if (configKey.group != "[KeyboardShortcuts]") {
-                    ControlObject* control = ControlObject::getControl(configKey);
-                    if (control) {
-                        //qDebug() << configKey << "MidiOpCode::NoteOn" << 1;
-                        // Add key to active key list
-                        m_qActiveKeyList.append(KeyDownInformation(
-                            keyId, ke->modifiers(), control));
-                        // Since setting the value might cause us to go down
-                        // a route that would eventually clear the active
-                        // key list, do that last.
-                        control->setValueFromMidi(MidiOpCode::NoteOn, 1);
-                        result = true;
-                    } else {
-                        qDebug() << "Warning: Keyboard key is configured for nonexistent control:"
-                                 << configKey.group << configKey.item;
-                    }
+
+                if (configKey.group == QStringLiteral("[KeyboardShortcuts]")) {
+                    // We don't handle menubar shortcuts here
+                    continue;
+                }
+
+                ControlObject* pControl = ControlObject::getControl(configKey);
+                if (pControl) {
+                    // qDebug() << configKey << "MidiOpCode::NoteOn" << 1;
+                    // Add key to active key list
+                    m_qActiveKeyList.append(KeyDownInformation(
+                            keyId, pKE->modifiers(), pControl));
+                    // Since setting the value might cause us to go down
+                    // a route that would eventually clear the active
+                    // key list, do that last.
+                    pControl->setValueFromMidi(MidiOpCode::NoteOn, 1);
+                    result = true;
+                } else {
+                    qWarning() << "Key" << keyId
+                               << "is configured for nonexistent control:"
+                               << configKey.group << configKey.item;
                 }
             }
             return result;
@@ -79,14 +97,14 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
             // getKeySeq() returns empty string if the press was a modifier only
             // On most system Alt sends Alt + Qt::Key_Alt, but with Qt 6.9 (on Linux)
             // this changed apparently so that it's just Qt::Key_Alt
-            if (((ke->modifiers() & Qt::AltModifier) || ke->key() == Qt::Key_Alt) &&
+            if (((pKE->modifiers() & Qt::AltModifier) || pKE->key() == Qt::Key_Alt) &&
                     !m_altPressedWithoutKey) {
                 m_altPressedWithoutKey = true;
             }
 #endif
         }
     } else if (e->type() == QEvent::KeyRelease) {
-        QKeyEvent* ke = (QKeyEvent*)e;
+        QKeyEvent* pKE = static_cast<QKeyEvent*>(e);
 
 #ifndef __APPLE__
         // QAction hotkeys are consumed by the object that created them, e.g.
@@ -94,7 +112,7 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
         // However, it may happen that we receive a RELEASE event for an Alt+key
         // combo for which no KEYPRESS was registered.
         // So react only to Alt-only releases.
-        if (m_altPressedWithoutKey && ke->key() == Qt::Key_Alt) {
+        if (m_altPressedWithoutKey && pKE->key() == Qt::Key_Alt) {
             emit altPressedWithoutKeys();
         }
         m_altPressedWithoutKey = false;
@@ -102,11 +120,10 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
 
 #ifdef __APPLE__
         // On Mac OSX the nativeScanCode is empty
-        int keyId = ke->key();
+        int keyId = pKE->key();
 #else
-        int keyId = ke->nativeScanCode();
+        int keyId = pKE->nativeScanCode();
 #endif
-        bool autoRepeat = ke->isAutoRepeat();
 
         //qDebug() << "KeyRelease event =" << ke->key() << "AutoRepeat =" << autoRepeat << "KeyId =" << keyId;
 
@@ -115,13 +132,15 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
         // OS X apparently doesn't deliver KeyRelease events when you are
         // holding Ctrl. So release all key-presses that were triggered with
         // Ctrl.
-        if (ke->key() == Qt::Key_Control) {
+        if (pKE->key() == Qt::Key_Control) {
             clearModifiers = Qt::ControlModifier;
         }
 #endif
 
+        bool autoRepeat = pKE->isAutoRepeat();
         bool matched = false;
-        // Run through list of active keys to see if the released key is active
+        // Run through list of active keys to see if the released key is active.
+        // Start from end because we may remove the current item.
         for (int i = m_qActiveKeyList.size() - 1; i >= 0; i--) {
             const KeyDownInformation& keyDownInfo = m_qActiveKeyList[i];
             ControlObject* pControl = keyDownInfo.pControl;
@@ -141,7 +160,8 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
         return matched;
     } else if (e->type() == QEvent::KeyboardLayoutChange) {
         // This event is not fired on ubunty natty, why?
-        // TODO(XXX): find a way to support KeyboardLayoutChange Bug #997811
+        // TODO: find a way to support KeyboardLayoutChange
+        // https://github.com/mixxxdj/mixxx/issues/6424
         //qDebug() << "QEvent::KeyboardLayoutChange";
     }
     return false;
@@ -189,15 +209,45 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
     return k;
 }
 
-void KeyboardEventFilter::setKeyboardConfig(ConfigObject<ConfigValueKbd>* pKbdConfigObject) {
-    // Keyboard configs are a surjection from ConfigKey to key sequence. We
-    // invert the mapping to create an injection from key sequence to
-    // ConfigKey. This allows a key sequence to trigger multiple controls in
-    // Mixxx.
-    m_keySequenceToControlHash = pKbdConfigObject->transpose();
-    m_pKbdConfigObject = pKbdConfigObject;
+void KeyboardEventFilter::setEnabled(bool enabled) {
+    if (enabled) {
+        qDebug() << "Enable keyboard shortcuts/mappings";
+    } else {
+        qDebug() << "Disable keyboard shortcuts/mappings";
+    }
+    m_enabled = enabled;
+    m_pConfig->setValue(ConfigKey("[Keyboard]", "Enabled"), enabled);
 }
 
-ConfigObject<ConfigValueKbd>* KeyboardEventFilter::getKeyboardConfig() {
-    return m_pKbdConfigObject;
+void KeyboardEventFilter::createKeyboardConfig() {
+    // Read keyboard configuration and set kdbConfig object in WWidget
+    // Check first in user's Mixxx directory
+    QString keyboardFile = QDir(m_pConfig->getSettingsPath()).filePath("Custom.kbd.cfg");
+    if (QFile::exists(keyboardFile)) {
+        qDebug() << "Found and will use custom keyboard mapping" << keyboardFile;
+    } else {
+        // check if a default keyboard exists
+        QString resourcePath = m_pConfig->getResourcePath();
+        keyboardFile = QString(resourcePath).append("keyboard/");
+        keyboardFile += m_locale.name();
+        keyboardFile += ".kbd.cfg";
+        if (!QFile::exists(keyboardFile)) {
+            qDebug() << keyboardFile << " not found, using en_US.kbd.cfg";
+            keyboardFile = QString(resourcePath).append("keyboard/").append("en_US.kbd.cfg");
+            if (!QFile::exists(keyboardFile)) {
+                qDebug() << keyboardFile << " not found, starting without shortcuts";
+                keyboardFile = "";
+            }
+        } else {
+            qDebug() << "Found and will use default keyboard mapping" << keyboardFile;
+        }
+    }
+
+    // Read the keyboard configuration file and set m_pKbdConfig.
+    // Keyboard configs are a surjection from ConfigKey to key sequence.
+    // We invert the mapping to create an injection from key sequence to
+    // ConfigKey. This allows a key sequence to trigger multiple controls in
+    // Mixxx.
+    m_pKbdConfig = std::make_shared<ConfigObject<ConfigValueKbd>>(keyboardFile);
+    m_keySequenceToControlHash = m_pKbdConfig->transpose();
 }
--- a/src/controllers/keyboard/keyboardeventfilter.cpp
+++ b/src/controllers/keyboard/keyboardeventfilter.cpp
@@ -110,7 +110,7 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
 
         //qDebug() << "KeyRelease event =" << ke->key() << "AutoRepeat =" << autoRepeat << "KeyId =" << keyId;
 
-        int clearModifiers = 0;
+        Qt::KeyboardModifiers clearModifiers = Qt::NoModifier;
 #ifdef __APPLE__
         // OS X apparently doesn't deliver KeyRelease events when you are
         // holding Ctrl. So release all key-presses that were triggered with
@@ -126,7 +126,8 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
             const KeyDownInformation& keyDownInfo = m_qActiveKeyList[i];
             ControlObject* pControl = keyDownInfo.pControl;
             if (keyDownInfo.keyId == keyId ||
-                    (clearModifiers > 0 && keyDownInfo.modifiers == clearModifiers)) {
+                    (clearModifiers != Qt::NoModifier &&
+                            keyDownInfo.modifiers == clearModifiers)) {
                 if (!autoRepeat) {
                     //qDebug() << pControl->getKey() << "MidiOpCode::NoteOff" << 0;
                     pControl->setValueFromMidi(MidiOpCode::NoteOff, 0);
--- a/src/controllers/keyboard/keyboardeventfilter.cpp
+++ b/src/controllers/keyboard/keyboardeventfilter.cpp
@@ -7,6 +7,12 @@
 #include "moc_keyboardeventfilter.cpp"
 #include "util/cmdlineargs.h"
 
+namespace {
+const QString mappingFilePath(const QString& dir, const QString& fileName) {
+    return QDir(dir).filePath(fileName + QStringLiteral(".kbd.cfg"));
+}
+} // anonymous namespace
+
 KeyboardEventFilter::KeyboardEventFilter(UserSettingsPointer pConfig,
         const QLocale& locale,
         QObject* pParent)
@@ -222,24 +228,22 @@ void KeyboardEventFilter::setEnabled(bool enabled) {
 void KeyboardEventFilter::createKeyboardConfig() {
     // Read keyboard configuration and set kdbConfig object in WWidget
     // Check first in user's Mixxx directory
-    QString keyboardFile = QDir(m_pConfig->getSettingsPath()).filePath("Custom.kbd.cfg");
+    QString keyboardFile = mappingFilePath(m_pConfig->getSettingsPath(), QStringLiteral("Custom"));
     if (QFile::exists(keyboardFile)) {
         qDebug() << "Found and will use custom keyboard mapping" << keyboardFile;
     } else {
         // check if a default keyboard exists
-        QString resourcePath = m_pConfig->getResourcePath();
-        keyboardFile = QString(resourcePath).append("keyboard/");
-        keyboardFile += m_locale.name();
-        keyboardFile += ".kbd.cfg";
-        if (!QFile::exists(keyboardFile)) {
-            qDebug() << keyboardFile << " not found, using en_US.kbd.cfg";
-            keyboardFile = QString(resourcePath).append("keyboard/").append("en_US.kbd.cfg");
+        const QString resourcePath = m_pConfig->getResourcePath() + QStringLiteral("keyboard/");
+        keyboardFile = mappingFilePath(resourcePath, m_locale.name());
+        if (QFile::exists(keyboardFile)) {
+            qDebug() << "Found and will use default keyboard mapping" << keyboardFile;
+        } else {
+            qDebug() << keyboardFile << " not found, try to use en_US.kbd.cfg";
+            keyboardFile = mappingFilePath(resourcePath, QStringLiteral("en_US"));
             if (!QFile::exists(keyboardFile)) {
                 qDebug() << keyboardFile << " not found, starting without shortcuts";
                 keyboardFile = "";
             }
-        } else {
-            qDebug() << "Found and will use default keyboard mapping" << keyboardFile;
         }
     }
 
--- a/src/controllers/keyboard/keyboardeventfilter.cpp
+++ b/src/controllers/keyboard/keyboardeventfilter.cpp
@@ -6,6 +6,7 @@
 #include <QtDebug>
 
 #include "moc_keyboardeventfilter.cpp"
+#include "preferences/constants.h"
 #include "util/cmdlineargs.h"
 #include "util/logger.h"
 #include "widget/wbasewidget.h"
@@ -59,6 +60,13 @@ KeyboardEventFilter::KeyboardEventFilter(UserSettingsPointer pConfig,
     m_enabled = m_pConfig->getValue(ConfigKey(QStringLiteral("[Keyboard]"),
                                             QStringLiteral("Enabled")),
             true);
+
+    mixxx::preferences::Tooltips cfgOnlyKbdShortcts =
+            pConfig->getValue(
+                    ConfigKey("[Controls]", "Tooltips"),
+                    mixxx::preferences::Tooltips::OnlyKbdShortcuts);
+    m_showOnlyKeyboardShortcuts =
+            cfgOnlyKbdShortcts == mixxx::preferences::Tooltips::OnlyKbdShortcuts;
 
     createKeyboardConfig();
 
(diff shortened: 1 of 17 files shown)


src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -6,14 +6,31 @@
...
+    }
+    m_enabled = pConfig->getValue<bool>(ConfigKey("[Keyboard]", "Enabled"));
+
+    createKeyboardConfig();
+
+    // For watching the currently loaded mapping file
+    connect(&m_fileWatcher,
+            &QFileSystemWatcher::fileChanged,
+            this,
+            &KeyboardEventFilter::reloadKeyboardConfig);
```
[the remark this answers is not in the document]
ronso0: Thank you for this hint!
Do you mind filing a bug report so we can track that?

For now I'd stick with the implementation we also use for controller scripts.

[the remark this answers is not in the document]
acolombier: Will do! 
Note that I fixed it in controllers in my [PR for screen controller](https://github.com/mixxxdj/mixxx/pull/11407)

ronso0:
oh, need to consider skin reload...

ronso0:
> Extending this to WMainMenuBar, as done in https://github.com/mixxxdj/mixxx/pull/1746, is certainly possibly.

Done.

ronso0:
We could add dummy shortcuts (empty) to all currently unmapped menubar actions so users can add shortcuts themselves.
(i.e. add ConfigKeys to menubar and all *.kbd.cfg files)

Nothing that holds up this PR though.

cr7pt0gr4ph7 (commented):
Not too important, but the upper/lowercasing of `KeyboardEventFIlter` got a bit mixed up the commit message [`Keyboard: handle mapping file and config setting in KeyboardEventFIlter`](https://github.com/mixxxdj/mixxx/pull/13082/commits/64647c886177aa444f0804d0e2ea444ea4f6d14f)`.

ronso0:
Sure, assigned it to 2.5 for now because I think this is a bugfix, see #9814 
It can either slip in during the beta period (it's already working without issues), or later.

Set to draft because I've just rebased this and am about to tweak it considering @cr7pt0gr4ph7's suggestions.

cr7pt0gr4ph7 (commented):
Small nitpick: Spelling of "chnanged" in the commit message. LGTM otherwise.

ronso0:
Thanks for taking the time to review!

> Spelling of "chnanged" in the commit message.

Yeah, that happens when I work in bright sunlight...

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -156,15 +180,139 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+    connect(this,
+            &KeyboardEventFilter::shortcutsEnabled,
+            this,
+            [pWidget](bool enabled) {
+                pWidget->toggleKeyboardShortcutHints(enabled);
+            });
+}
+
+void KeyboardEventFilter::updateWidgetShortcuts() {
+    for (auto pWidget : m_widgets) {
```
acolombier, line 203: Looks like there is a clang-tidy here
```suggestion
    for (auto* pWidget: m_widgets) {
```

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -1,19 +1,37 @@
...
+          m_locale(locale),
+          m_enabled(false) {
     setObjectName(name);
-    setKeyboardConfig(pKbdConfigObject);
+
+    // get enabled state
+    if (pConfig->getValueString(ConfigKey("[Keyboard]", "Enabled")).length() == 0) {
+        pConfig->set(ConfigKey("[Keyboard]", "Enabled"), ConfigValue(1));
+    }
+    m_enabled = pConfig->getValue<bool>(ConfigKey("[Keyboard]", "Enabled"));
```
acolombier, line 26: Do we not want this to be enabled by default? I understand it used to be always on previously (`src/coreservices.cpp:459`).
```suggestion
    m_enabled = pConfig->getValue(ConfigKey("[Keyboard]", "Enabled"), true);
```

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -156,15 +180,139 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+                shortcutHints.append(buildShortcutString(keyString, key.second));
+            }
+        }
+        // might be empty
+        pWidget->setShortcutTooltip(shortcutHints, m_enabled);
+    }
+}
+
+void KeyboardEventFilter::clearWidgets() {
+    disconnect(this, nullptr, nullptr, nullptr);
```
acolombier, line 219: I think this is equivalent of the sing arg disconnect
 ```suggestion
     disconnect(this);
```

File: src/controllers/keyboard/keyboardeventfilter.cpp [hunk 5]
```diff
@@ -156,15 +180,139 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+    if (shortcut.isEmpty()) {
+        return QString();
+    }
+
+    // translate shortcut to native text
+    QString nativeShortcut = QKeySequence(shortcut, QKeySequence::PortableText)
+                                     .toString(QKeySequence::NativeText);
+
+    QString shortcutTooltip;
+    shortcutTooltip += QObject::tr("Shortcut");
```
acolombier, line 234: `KeyboardEventFilter` already inherit `QObject`
```suggestion
    shortcutTooltip += tr("Shortcut");
```

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -156,15 +180,139 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+    }
+    shortcutTooltip += ": ";
+    shortcutTooltip += nativeShortcut;
+    return shortcutTooltip;
+}
+
+void KeyboardEventFilter::registerMenuBarActionSetShortcut(QAction* pAction,
+        const ConfigKey& command,
+        const QString& defaultShortcut) {
+    VERIFY_OR_DEBUG_ASSERT(true /* reminder */) {
```
acolombier, line 247: Reminder? :)

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -156,15 +180,139 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+void KeyboardEventFilter::clearMenuBarActions() {
+    m_menuBarActions.clear();
+}
+
+void KeyboardEventFilter::updateMenuBarActionShortcuts() {
+    QHashIterator<QAction*, std::pair<ConfigKey, QString>> it(m_menuBarActions);
+    while (it.hasNext()) {
+        it.next();
+        const QString keyStr = m_pKbdConfig->getValue(it.value().first, it.value().second);
+        auto pAction = it.key();
```
acolombier, line 264: clang-tidy here

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -156,15 +180,139 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
 #endif
 }
 
-void KeyboardEventFilter::setKeyboardConfig(ConfigObject<ConfigValueKbd>* pKbdConfigObject) {
+void KeyboardEventFilter::setEnabled(bool enabled) {
+    m_enabled = enabled;
+    m_pConfig->set(ConfigKey("[Keyboard]", "Enabled"), ConfigValue(enabled));
```
acolombier, line 185: nit here, but perhaps use the templated version since you already use `getValue` later?
```suggestion
    m_pConfig->setValue(ConfigKey("[Keyboard]", "Enabled"), enabled);
```

src/coreservices.cpp
```diff
@@ -453,64 +451,23 @@ void CoreServices::initialize(QApplication* pApp) {
...
+    return m_pKeyboardEventFilter->getKeyboardConfig();
 }
 
 void CoreServices::slotOptionsKeyboard(bool toggle) {
     UserSettingsPointer pConfig = m_pSettingsManager->settings();
     if (toggle) {
-        //qDebug() << "Enable keyboard shortcuts/mappings";
-        m_pKeyboardEventFilter->setKeyboardConfig(m_pKbdConfig.get());
-        pConfig->set(ConfigKey("[Keyboard]", "Enabled"), ConfigValue(1));
+        // qDebug() << "Enable keyboard shortcuts/mappings";
```
acolombier, line 466: Probably worth keeping those two uncommented?

src/widget/wmainmenubar.cpp
```diff
@@ -174,11 +174,11 @@ void WMainMenuBar::initialize() {
     QString createCrateTitle = tr("Create New &Crate");
     QString createCrateText = tr("Create a new crate");
     auto* pLibraryCreateCrate = new QAction(createCrateTitle, this);
-    pLibraryCreateCrate->setShortcut(
-        QKeySequence(m_pKbdConfig->getValue(ConfigKey("[KeyboardShortcuts]",
-                                                  "LibraryMenu_NewCrate"),
-                                                  tr("Ctrl+Shift+N"))));
-    pLibraryCreateCrate->setShortcutContext(Qt::ApplicationShortcut);
+    // TODO update shortcut in Crates feature ?
```
acolombier, line 177: Does this `TODO` need addressing now? Maybe create an issue for tracking otherwise?

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: Right. This is equivalent to `this->disconnect();` so just `disconnect();` will do.

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: Yup. Leftover from when this was a namespace function.

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: ah, thanks for reminding me ; )

I was thinking about validating `defaultShortcut`, like forbid modifier-only key seq.
(they're filtered by eventFilter() anyway)
But that should happen after querying the config.

src/coreservices.cpp
ronso0: Yeah, that's probably better. At least it doens't hurt.

src/widget/wmainmenubar.cpp
ronso0: We also briefly discussed library hotkeys in #13148 so I don't think it's necessary and will remove the TODO.
(and actually there's no shortcut in Crates root, just a push button)

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: Thanks, done.

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: done.

src/widget/wmainmenubar.cpp
```diff
@@ -162,9 +169,10 @@ void WMainMenuBar::initialize() {
     QString createPlaylistText = tr("Create a new playlist");
     auto* pLibraryCreatePlaylist = new QAction(createPlaylistTitle, this);
     pLibraryCreatePlaylist->setShortcut(
```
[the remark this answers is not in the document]
ronso0: Thanks, done.

src/controllers/keyboard/keyboardeventfilter.cpp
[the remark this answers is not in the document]
ronso0: I switched to `AutoFileReloader` now that #13607 has been merged.

ronso0:
Oh sorry, I did a rebase instead of merging main.

src/widget/wmainmenubar.cpp
```diff
@@ -458,16 +450,15 @@ void WMainMenuBar::initialize() {
...
+    m_pKeyboard->registerMenuBarActionSetShortcut(
+            pOptionsKeyboard,
+            ConfigKey("[KeyboardShortcuts]", "OptionsMenu_EnableShortcuts"),
+            tr("Ctrl+`"));
     pOptionsKeyboard->setCheckable(true);
     pOptionsKeyboard->setChecked(keyboardShortcutsEnabled);
     pOptionsKeyboard->setStatusTip(keyboardShortcutText);
     pOptionsKeyboard->setWhatsThis(buildWhatsThis(keyboardShortcutTitle, keyboardShortcutText));
-    connect(pOptionsKeyboard, &QAction::triggered, this, &WMainMenuBar::toggleKeyboardShortcuts);
+    connect(pOptionsKeyboard, &QAction::triggered, m_pKeyboard, &KeyboardEventFilter::setEnabled);
```
Swiftb0y, line 461:
```suggestion
    connect(pOptionsKeyboard, &QAction::triggered, m_pKeyboard.get(), &KeyboardEventFilter::setEnabled);
```

src/widget/wmainmenubar.cpp
```diff
@@ -74,16 +75,19 @@ QUrl documentationUrl(
...
 
-WMainMenuBar::WMainMenuBar(QWidget* pParent, UserSettingsPointer pConfig,
-                           ConfigObject<ConfigValueKbd>* pKbdConfig)
+WMainMenuBar::WMainMenuBar(QWidget* pParent,
+        UserSettingsPointer pConfig,
+        std::shared_ptr<KeyboardEventFilter> pKbd)
         : QMenuBar(pParent),
           m_pConfig(pConfig),
-          m_pKbdConfig(pKbdConfig) {
+          m_pKeyboard(pKbd) {
```
Swiftb0y, line 83: to avoid unnecessary ref count changes.
```suggestion
          m_pKeyboard(std::move(pKbd)) {
```

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -1,22 +1,46 @@
...
+          m_enabled(false),
+          m_autoReloader(RuntimeLoggingCategory(QStringLiteral("kbd_auto_reload"))) {
     setObjectName(name);
-    setKeyboardConfig(pKbdConfigObject);
+
+    // get enabled state
+    if (pConfig->getValueString(ConfigKey("[Keyboard]", "Enabled")).length() == 0) {
+        pConfig->set(ConfigKey("[Keyboard]", "Enabled"), ConfigValue(1));
+    }
+    m_enabled = pConfig->getValue<bool>(ConfigKey("[Keyboard]", "Enabled"), true);
```
Swiftb0y, line 35: somewhere else:
```
const ConfigKey kKeyboardEnabled = ConfigKey(QStringLiteral("[Keyboard]"), QStringLiteral("Enabled"));
```
```suggestion
    if (pConfig->getValueString(kKeyboardEnabled).isEmpty()) {
        pConfig->setValue<bool>(kKeyboardEnabled, true);
    }
    m_enabled = pConfig->getValue<bool>(kKeyboardEnabled, true);
```

src/controllers/keyboard/keyboardeventfilter.cpp
Swiftb0y, line 1: this is mostly copy pasted, right? Do you want to improve it or leave it?

Swiftb0y (commented):
couple codestyle complaints

src/widget/wbasewidget.h
```diff
@@ -95,7 +126,13 @@ class WBaseWidget {
 
   private:
     QWidget* m_pWidget;
+
     QString m_baseTooltip;
+    QString m_shortcutTooltip;
+    QString m_baseTooltipOptShortcuts;
+    QList<std::pair<ConfigKey, QString>> m_shortcutKeys;
```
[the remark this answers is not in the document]
ronso0: For knobs and sliders there'll be 5 controls (control, _up, _down, _up_small, _down_small).
For other widgets it's 1-3 (left- and right-click, maybe a display conection).

src/coreservices.h
```diff
@@ -43,9 +43,7 @@ class CoreServices : public QObject {
         return m_pKeyboardEventFilter;
     }
 
-    std::shared_ptr<ConfigObject<ConfigValueKbd>> getKeyboardConfig() const {
-        return m_pKbdConfig;
-    }
+    std::shared_ptr<ConfigObject<ConfigValueKbd>> getKeyboardConfig() const;
```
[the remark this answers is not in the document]
ronso0: In KeyboardEventFilter yes.
Here it was small before, when CoreServices created the ConfigValueKbd.
Now KeyboardEventFilter creates it and we _may_ not have created that yet, so I think I'll add another test, so it'll be slightly larger. Like this
```c++
    std::shared_ptr<ConfigObject<ConfigValueKbd>> getKeyboardConfig() const {
        VERIFY_OR_DEBUG_ASSERT(m_pKeyboardEventFilter) {
            return std::make_shared<ConfigObject<ConfigValueKbd>>();
        }
        return m_pKeyboardEventFilter->getKeyboardConfig();
    }
```
Wdyt?

src/widget/wbasewidget.h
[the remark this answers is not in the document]
Swiftb0y: thanks. makes sense. This is not duplicated between widgets then?

src/coreservices.h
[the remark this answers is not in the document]
Swiftb0y: makes sense. Though you may want to consider returning a nullptr instead.

src/widget/wbasewidget.h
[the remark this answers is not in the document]
ronso0: No. It's set by LegacySkinParser and fetched by KeyboardEventFilter which looks up the keys per control, composes the tooltip and pushes that to the widget.

src/coreservices.h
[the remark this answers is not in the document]
ronso0: Hmm, actually we don't need that anymore.
The only place where keys are requested is
https://github.com/mixxxdj/mixxx/blob/34dde6fc7a0128ecd92861b19aa063a6e7af6507/src/library/autodj/autodjfeature.cpp#L180
and that gets a KeyboardEventFilter pointer from Library which has a CoreServices pointer.
And KeyboardEventFilter is initialized in the CoreServices ctor so this seems safe.

Will remove.

src/widget/wbasewidget.h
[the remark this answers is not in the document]
ronso0: Btw this QList with a pair feels a bit clunky, so I switched to a QMap<ConfigKey, QString> (sorted by key) with convenient cons_key_value_iterator.
This also has the benefit that shortcuts are now always sorted by control/sub-control's ConfigKey¹ (same group, so sorted by item) instead of order they've been added (QList) or sorted by the command's _tr_ string (std::map<QString, ConfigKey>.
The slight performance regression shouldn't matter much as we add 4 items max.

¹WPushButton: _activate, _toggle
WKnob/WSlider: _down, _down_small, _up, _up_small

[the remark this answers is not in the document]
ronso0: Oh, just noticed that we don't add the other ControlPotmeter controls (_set_minus_one, _set_default etc.)
When I add those for sake of completeness, it'd be 10 sub-controls per knob/slider, and we have a lot of knobs (fx parameters, mixer, maybe 64 samplers, ...)

Will revert to QList<std::pair<QString, ConfigKey)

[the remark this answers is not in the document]
ronso0: and we're also missing the HotcueButton controls (activate, clear + set, goto, .., gotoloop) which would add another 12 controls per button : \

Note that both these extra ControlPotmeter and WHotcueButton controls are not used in built-in mappings, except `rate_perm` and `rate_temp` of deck 1/2 (which are mapped to dedicated buttons in official skins anyway).

Not sure how to proceed, ie. idk whether there's a noteworthy performance impact (when toggling) keyboard shortcuts) if we add all those sub-controls which are likely not used.
What do yo think?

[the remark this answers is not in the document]
Swiftb0y: I'm primarily concerned about the memory overhead from many large QLists in many widgets. But I can't make a good judgement call here either.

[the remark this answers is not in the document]
Swiftb0y: Right, but I was considering memory overhead (likely also not easily measurable because we're wasting so much already in other places), not runtime performance. Moreover 80ms is quite a lot...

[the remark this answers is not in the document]
ronso0: > Moreover 80ms is quite a lot...

ah okay, without the qdebug lines it's down to ~4ms for ~1600 widgets (LateNight).

I'll push the latest version soonish.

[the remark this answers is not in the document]
ronso0: ...and now (without the debug stuff) it's obvious that the extra controls do have a noticeable impact:
with: 4-5ms
without: ~2.5ms

ronso0:
some thoughts on the general tooltip update approach used here and performance:

With the current implementatoin kbd mapppings are (re)loaded
* on startup
* when kbd shortcuts are toggled on, and
* when the mapping file has been edited

so I think a tiny performance hit (GUI thread) in these non-critical situations is an acceptable tradeoff for having the GUI reflect the complete kbd mapping (all possible poti sub-controls) immediatly, even if it's just for double-checking that the mapping has been applied as expected.

### The current method is:
* for each control widget (button, knob, ..) the skin parser collects ConfigKeys for the from-widget connections, optionally creates the sub-control strings to be displayed in the tooltip
('up', 'down', 'up small', 'down small' for knobs and sliders, 'activate', 'clear' etc. for Hotcue buttons)
* this CfgKey/string list is stored in each WBaseWidget
* each widget is registered in KeyboardEventFilter (~1600 for LateNight 16 Samplers)
* KeyboardEventFilter loads the kbd mapping file
* iterate over all registered widgets, check if the controls are in the kbd mapping
   * yes: (re)create kbd tooltip strings
   * no: clear kbd tooltip string
   * push string to widget, (re)construct tooltip
* similar method for menubar actions

The downside:
all widgets are updated even if just one shortcut has been changed.

So I tried another approach, hoping that it would it be faster (in total, ie. load mapping file, check shortcuts, update widgets):
* reload kbd mapping
* compare last/new mapping, collect changed/added/removed shortcuts
* collect affected widgets
* update affected widgets

I had to implement the compare method and store more data in KeyboardEventFilter, and it turned out that this did **not** shorten the total time to the extent that it justifies the additional complexity (and potential errors/inconsistency), given that shortcuts are evaluated rarely.
(current method: ~6-7ms in total, alternative: ~5-6ms, measured when changing one shortcut)
I can provide the commit if someone is interested.

ronso0:
some thoughts on the general tooltip update approach used here and performance:

With the current implementatoin kbd mapppings are (re)loaded
* on startup
* when kbd shortcuts are toggled on, and
* when the mapping file has been edited

so I think a tiny performance hit (GUI thread) in these non-critical situations is an acceptable tradeoff for having the GUI reflect the complete kbd mapping (all possible poti sub-controls) immediatly, even if it's just for double-checking that the mapping has been applied as expected.

### The current method is:
* for each control widget (button, knob, ..) the skin parser collects ConfigKeys for the from-widget connections, optionally creates the sub-control strings to be displayed in the tooltip
('up', 'down', 'up small', 'down small' for knobs and sliders, 'activate', 'clear' etc. for Hotcue buttons)
* this CfgKey/string list is stored in each WBaseWidget
* each widget is registered in KeyboardEventFilter (~1600 for LateNight 16 Samplers)
* KeyboardEventFilter loads the kbd mapping file
* iterate over all registered widgets, check if the controls are in the kbd mapping
   * yes: (re)create kbd tooltip strings
   * no: clear kbd tooltip string
   * push string to widget, (re)construct tooltip
* similar method for menubar actions

The downside:
all widgets are updated even if just one shortcut has been changed.

So I tried another approach, hoping that it would it be faster (in total, ie. load mapping file, check shortcuts, update widgets):
* reload kbd mapping
* compare last/new mapping, collect changed/added/removed shortcuts
* collect affected widgets
* update affected widgets

I had to implement the compare method and store more data in KeyboardEventFilter, and it turned out that this did **not** shorten the total time to the extent that it justifies the additional complexity (and potential errors/inconsistency), given that shortcuts are evaluated rarely.
(current method: ~6-7ms in total, alternative: ~5-6ms, measured when changing one shortcut)
I can provide the commit if someone is interested.

ronso0:
IMO this is now ready for a final review.
@Swiftb0y suggestions for improving the copied code (eventFilter()) are welcome!

acolombier:
Some conflicts have developing unfortunately. 
Friendly ping @Swiftb0y in case you can finish your review, happy to help otherwise!

ronso0:
Conflicts are resolved.
And it seems I screwed up the shortcut tooltip auto-update on config reload. Fixed.

Swiftb0y:
No time for a proper review right now. Can you do a sanity check @acolombier?

ronso0:
Just did a final test: all working as expected!
Except a Qt/platform issue (that exists in main, too) that won't show umlauts (with en_GB locale) and space as hotkey (though hotkeys work).

ronso0:
In general I agree that we can rename `KeyboardEventFilter` to `KeyBoardController` since it does way more than filtering the incoming events, but maybe that raises expectations that the keyboard has the same capabilities as MIDI/HID controllers :man_shrugging:

cr7pt0gr4ph7:
> In general I agree that we can rename `KeyboardEventFilter` to `KeyBoardController` since it does way more than filtering the incoming events, but maybe that raises expectations that the keyboard has the same capabilities as MIDI/HID controllers 🤷‍♂️

I'd stay with `KeyboardEventFilter`, and reserve `KeyboardController` (please... don't use `KeyBoard`) for when we get proper scripting support and integrate into the `Controller`/`MidiController`/`HidController` hierarchy. `Controller` has a pretty specific meaning in the code.

Alternatively, may I suggest the following alternatives?
- `KeyboardShortcutManager`
- `KeyboardMappingHandler`
- `KeyboardEventProcessor`

src/widget/wmainmenubar.cpp
```diff
@@ -844,14 +841,14 @@ void WMainMenuBar::hideMenuBar() {
...
-    if (m_pConfig->getValue<bool>(ConfigKey("[Config]", "hide_menubar"), false)) {
+    if (m_pConfig->getValue<bool>(kHideMenuCfgKey, false)) {
         // don't use setHidden(true) because Alt hotkeys wouldn't work anymore
         setFixedHeight(0);
     }
 }
 
 void WMainMenuBar::slotAutoHideMenuBarToggled(bool autoHide) {
-    m_pConfig->setValue(ConfigKey("[Config]", "hide_menubar"), autoHide ? 1 : 0);
+    m_pConfig->setValue(kHideMenuCfgKey, autoHide ? 1 : 0);
```
acolombier, line 851: Any reason why this is not a boolean?

src/widget/wmainmenubar.cpp
```diff
@@ -883,8 +880,8 @@ void WMainMenuBar::slotDeveloperStatsExperiment(bool enable) {
 }
 
 void WMainMenuBar::slotDeveloperDebugger(bool toggle) {
-    m_pConfig->set(ConfigKey("[ScriptDebugger]","Enabled"),
-                   ConfigValue(toggle ? 1 : 0));
+    m_pConfig->setValue(ConfigKey(QStringLiteral("[ScriptDebugger]"), QStringLiteral("Enabled")),
```
acolombier, line 883: Same here - since `setValue<bool>` relies on integer for serialisation, this should be safe to update

acolombier (commented):
Looking good - some comments, mostly on code style

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -1,22 +1,51 @@
...
+#include "util/logger.h"
+#include "util/timer.h"
+#include "widget/wbasewidget.h"
 
-KeyboardEventFilter::KeyboardEventFilter(ConfigObject<ConfigValueKbd>* pKbdConfigObject,
-        QObject* parent,
-        const char* name)
+namespace {
+const ConfigKey kKbdEnabledCfgKey =
+        ConfigKey(QStringLiteral("[Keyboard]"), QStringLiteral("Enabled"));
```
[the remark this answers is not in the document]
ronso0: No, I just added a constant for it.

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -28,7 +57,7 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
         // because we might not get Key Release events.
         m_qActiveKeyList.clear();
     } else if (e->type() == QEvent::KeyPress) {
-        QKeyEvent* ke = (QKeyEvent *)e;
+        QKeyEvent* ke = static_cast<QKeyEvent*>(e);
```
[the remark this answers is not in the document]
ronso0: yes, `pKE`

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -1,22 +1,51 @@
...
-    setKeyboardConfig(pKbdConfigObject);
+          m_pConfig(pConfig),
+          m_locale(locale),
+          m_enabled(false),
+          m_autoReloader(RuntimeLoggingCategory(QStringLiteral("kbd_auto_reload"))) {
+    // Get the enabled state.
+    // Set the default if the key/value doesn't exist.
+    if (pConfig->getValueString(kKbdEnabledCfgKey).isEmpty()) {
+        pConfig->setValue(kKbdEnabledCfgKey, true);
+    }
```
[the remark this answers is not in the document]
ronso0: I blindly copied this from previous implementation in CoreServices.

I think this was installed to make sure the Shortcuts action in the menubar is also checked by default, ie. work around the fact that in wmainmenubar.cpp we didn't pass a default when querying the value for shortcuts enabled (which happens after CoreServices initialized KeyboardEventFilter).

I think the place to do this is okay, but instead of `getValueString()` we should use `exists()`.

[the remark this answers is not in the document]
ronso0: and I'll add a comment here explaining the relation to WMainMenuBar

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -1,22 +1,51 @@
...
+    m_enabled = pConfig->getValue(kKbdEnabledCfgKey, true);
```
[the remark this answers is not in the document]
ronso0: I don't think so, these are the only two lines that use it, and they're quite close together.

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -142,7 +182,7 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
     } else if (e->type() == QEvent::KeyboardLayoutChange) {
         // This event is not fired on ubunty natty, why?
         // TODO(XXX): find a way to support KeyboardLayoutChange Bug #997811
-        //qDebug() << "QEvent::KeyboardLayoutChange";
+        // kLogger.debug() << "QEvent::KeyboardLayoutChange";
```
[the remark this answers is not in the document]
ronso0: :+1: comment removed.
But I'd rather keep the TODO. Much better here IMO than in some bug report that no one will discover while reading or working on this code.

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -180,24 +220,163 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+void KeyboardEventFilter::clearWidgets() {
+    disconnect();
+    m_widgets.clear();
+}
+
+const QString KeyboardEventFilter::buildShortcutString(
+        const QString& shortcut, const QString& cmd) const {
+    if (shortcut.isEmpty()) {
+        return QString();
+    }
```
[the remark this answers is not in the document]
ronso0: No, not to be expected. Atm this is called only from updateWidgetShortcuts() above.
I added a DEBUG_ASSERT just in case it'll be used elsewhere.

But should I rather make it `inline` in the namespace?

src/widget/wmainmenubar.cpp
ronso0: copypasta here and above.
will fix it

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -107,11 +145,10 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
...
         int keyId = ke->nativeScanCode();
 #endif
-        bool autoRepeat = ke->isAutoRepeat();
-
-        //qDebug() << "KeyRelease event =" << ke->key() << "AutoRepeat =" << autoRepeat << "KeyId =" << keyId;
+        // kLogger.debug() << "KeyRelease event =" << ke->key()
+        // << "AutoRepeat=" << autoRepeat << "KeyId =" << keyId;
 
-        int clearModifiers = 0;
+        int clearModifiers = Qt::NoModifier;
```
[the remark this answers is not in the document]
ronso0: :+1:

ronso0:
Just a heads-up: the `block-fixup` job succeeded even though there are a few !fixups :man_shrugging:

ronso0:
Unfortunately with my habit to do fixup commits and this PR sitting for quite some time, which required merging main, I have created kind of a commit mess here, which is not trivial to resolve without rebasing.

So I went ahead and squashed & rebased. Commits are compact now IMO and, despite the big changes, should be easy/okay to review.

ronso0:
I was very grateful this works as desired when my controllers Play buttons ceased to work during my last live session.
I quickly assigned `1` and `2` to `play`, saved the mapping and could continue my set.

daschuer:
@acolombier and @Swiftb0y can you have a final look here? Are your comments sufficiently fixed?

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -1,22 +1,57 @@
...
+          m_locale(locale),
+          m_enabled(false),
+          m_autoReloader(RuntimeLoggingCategory(QStringLiteral("kbd_auto_reload"))) {
+    // Get the enabled state.
+    // Set the default if the key/value doesn't exist so that the shortcuts action
+    // in WMainMenuBar is also enabled by default.
+    if (!pConfig->exists(ConfigKey("[Keyboard]", "Enabled"))) {
+        pConfig->setValue(ConfigKey("[Keyboard]", "Enabled"), true);
+    }
+    m_enabled = pConfig->getValue<bool>(ConfigKey("[Keyboard]", "Enabled"), true);
```
acolombier: Is this redundant? What is the reason to write the default value in the setting rather than simply use default if unset? Also, not that the template is redundant and would be inferred from the default value type AFAIK

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -28,49 +63,58 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
         // because we might not get Key Release events.
         m_qActiveKeyList.clear();
     } else if (e->type() == QEvent::KeyPress) {
-        QKeyEvent* ke = (QKeyEvent *)e;
+        QKeyEvent* pKE = static_cast<QKeyEvent*>(e);
```
acolombier, line 89: Should this not be a `dynamic_cast` instead? Maybe @Swiftb0y can comfirm?

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -28,49 +63,58 @@ bool KeyboardEventFilter::eventFilter(QObject*, QEvent* e) {
...
+                    // Add key to active key list
+                    m_qActiveKeyList.append(KeyDownInformation(
+                            keyId, pKE->modifiers(), pControl));
+                    // Since setting the value might cause us to go down
+                    // a route that would eventually clear the active
+                    // key list, do that last.
+                    pControl->setValueFromMidi(MidiOpCode::NoteOn, 1);
+                    result = true;
+                } else {
+                    qWarning() << "Key" << keyId
```
acolombier: nit - use the local `klogger` for categorisation

```suggestion
                    kLogger.warning() << "Key" << keyId
```

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -179,24 +223,222 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
-void KeyboardEventFilter::setKeyboardConfig(ConfigObject<ConfigValueKbd>* pKbdConfigObject) {
-    // Keyboard configs are a surjection from ConfigKey to key sequence. We
-    // invert the mapping to create an injection from key sequence to
-    // ConfigKey. This allows a key sequence to trigger multiple controls in
-    // Mixxx.
-    m_keySequenceToControlHash = pKbdConfigObject->transpose();
-    m_pKbdConfigObject = pKbdConfigObject;
+void KeyboardEventFilter::setEnabled(bool enabled) {
+    if (enabled) {
+        qDebug() << "Enable keyboard shortcuts/mappings";
```
acolombier: nit - use `klogger`

```suggestion
        kLogger.debug() << "Enable keyboard shortcuts/mappings";
```

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -179,24 +223,222 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+    } else {
+        qDebug() << "Disable keyboard shortcuts/mappings";
```
acolombier:
```suggestion
        kLogger.debug() << "Disable keyboard shortcuts/mappings";
```

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -179,24 +223,222 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+    m_widgets.append(pWidget);
+
+    // Tell widgets to reconstruct tooltips when option is toggled.
+    // WBaseWidget is not a QObject, need to use a lambda
+    connect(this,
+            &KeyboardEventFilter::shortcutsEnabled,
+            this,
+            [pWidget](bool enabled) {
+                pWidget->toggleKeyboardShortcutHints(enabled);
+            });
```
acolombier, line 295: Is there no risk of use-after-free here? From my understanding, this will capture the pointer inside `pWidget` and use it for connection, but this connection will remain active (and trigger) after `WBaseWidget` has been freed.

I _ believe_ this could be fixed by using `WBaseWidget` as connection receiver, so the connection would automatically be deleted upon release


```suggestion
    connect(this,
            &KeyboardEventFilter::shortcutsEnabled,
            pWidget,
            [pWidget](bool enabled) {
                pWidget->toggleKeyboardShortcutHints(enabled);
            });
```

Perhaps @Swiftb0y can confirm my analyse.

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -179,24 +223,222 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+
+    // If toggled in Preferences -> Interface, we tell the widgets to show
+    // only the keyboard shortcuts tooltip.
+    connect(this,
+            &KeyboardEventFilter::showOnlyKbdShortcuts,
+            this,
```
acolombier, line 305: ditto use-after-freer risk?

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -179,24 +223,222 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+                pWidget->toggleShowOnlyKeyboardShortcuts(enabled);
+            });
+}
+
+void KeyboardEventFilter::updateWidgetShortcuts() {
+    kLogger.debug() << "updateWidgetShortcuts";
+    QStringList shortcutHints;
+    for (auto* pWidget : std::as_const(m_widgets)) {
+        QString keyString;
+        shortcutHints.clear();
```
acolombier, line 316: Nit - move `shortcutHints` declaration to this scope instead?

src/widget/wmainmenubar.h
```diff
@@ -112,7 +114,8 @@ class WMainMenuBar : public QMenuBar {
 
     UserSettingsPointer m_pConfig;
     QAction* m_pViewKeywheel;
-    ConfigObject<ConfigValueKbd>* m_pKbdConfig;
+    // TODO KeyboardEventFilterPointer ??
```
acolombier, line 117: Is this still `TODO`? If yes, could you elaborate it a bit further?

acolombier (commented):
Looking good, haven't retested again! Just a small Q around connections with lambda capture, but all 👍 otherwise.

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: afaik `static_cast` ahs been used for key and mouse events in many places :man_shrugging: 
It's used in the Qt docs btw and considered safe -- after checking QEvent::type() !

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: Unfortunately WBaseWidget is not a QObject.
IIUC the only way pWidget is cleared is when
a) Mixxx is shut down, or
b) the/a skin is re/loaded

In case of b) LegacySkinParser calls `KeyboardEventFilter::clearWidgets()` which also clears all of KeyboardEventFilter's connections.
BUT there is the theoretical possibility that this is called after MixxxMainWindow's `centralWidget()` (the skin wrapper widget) has been destroyed, but LegacySkinParser did not yet have the chance to call `clearWidgets()`.
-- took care of that in `MixxxMainWindow::~MixxxMainWindow()`
Will add it to `MixxxMainWindow::rebootMixxxView()` as well.

Apart from that idk how else to prevent use-after-free.

src/widget/wmainmenubar.h
ronso0: Not necessarily. Will remove it.
I just noticed that I typed `std::shared_ptr<KeyboardEventFilter>` a few times and wondered if we could add a wrapper/typedef `KeyboardEventFilterPointer` like we have for the settings
(`typedef QWeakPointer<UserSettings> UserSettingsWeakPointer;`)

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: Indeed, this could be just
`m_enabled = pConfig->getValue(ConfigKey("[Keyboard]", "Enabled"), true);`

ronso0: ... if I also change the action setup in WMainMenuBar (see comment above this)

ronso0:
Thanks @acolombier for your review, I think I fixed everything.

If all's good I'd like to squash the small fixups -- not 100% sure how well that goes with the intermediate merge commits. Can as well rebase onto main, just to have a clean history.

src/widget/wmainmenubar.h
acolombier: I think that makes sense indeed!

src/controllers/keyboard/keyboardeventfilter.cpp
acolombier: Ah I forgot this this isn't based of QObject! Too many QML lately :sweat_smile: 

I'm happy with the current state of play then.

src/controllers/keyboard/keyboardeventfilter.cpp
acolombier: Makes sense - thanks

src/controllers/keyboard/keyboardeventfilter.cpp
acolombier: Hmm I feel like it would be best to also do a `getValue(ConfigKey("[Keyboard]", "Enabled"), KeyboardEventFilter::kEnabledDefault)` there, but since this code has already be sailing for a while, let's leave it as is.

acolombier (approved):
Looking good! 

IIRC, you should be able to rebase this to the latest head you merged in (`9e947aec194aeb9a0f74330d1ad16b27e8bc1b83` by the looks of it?) and it _should_ drop the merge commit from the history

```
git rebase --autosquash -i 9e947aec194aeb9a0f74330d1ad16b27e8bc1b83
```

Edit: looks like CI is failing on Windows

ronso0:
yeah, all builds are failing because of a conflict, and there is a mismatch between local and this, idk where that comes from.
I'll backup and try to rebase locally, first without the fixups. Let's see how that goes..

src/widget/wmainmenubar.h
ronso0: I tried it, but I think it's not worth it because (like i did it) the forward-declarations won't suffice anymore and it requires extra header includes. Here's a test commit 06bf4cdb632344988590e034424e92db55e7d2c8

ronso0:
I pushed a fixup for the new tooltip option "Only keyboard shortcuts".
Works well now.

**edit** Oh, for that I also need to install the use-after-free protection

**edit** Ha, all good! All additionally connected widgets (with the fixup) are covered by the current protection (disconnect() on KeyboardEventFilter)

daschuer:
A conflict has developed.

ronso0:
I'll move the QKeySequence helpers added in #16219 to KeyboardEventFilter, that's the only place where we need them now.

res/skins/LateNight/style_palemoon.qss
```diff
@@ -155,7 +155,7 @@ QAbstractScrollArea::corner {
 #WaveformsFrame,
 #OverviewBox,
 #OverviewBoxMini,
-#KeyText,
+#Deck #KeyText,
```
ronso0: TODO remove, slipped in by accident

daschuer:
pre commit fails unrelated because of "re-using"

daschuer:
This works, but is not that robust. 

I have tested a typo with a double [[ this causes an assertion, when pressing the short cut. 

```
warning [Main] ControlDoublePrivate::getControl returning NULL for ( "[[Channel1]" , "play" )
critical [Main] DEBUG ASSERT: "flags.testFlag(ControlFlag::NoAssertIfMissing)" in function static QSharedPointer<ControlDoublePrivate> ControlDoublePrivate::getControl(const ConfigKey&, ControlFlags, ControlObject*, bool, bool, bool, double) at ./src/control/control.cpp:196
```
probably not an issue form this PR though. 

It is just that the "on the fly changing" use pattern might require an extra safety net and shall not crash "delayed". 
Can we do a sanity check when reading the file?

ronso0:
> warning [Main] ControlDoublePrivate::getControl returning NULL for ( "[[Channel1]" , "play" )

That's indeed not an issue from this PR.
We can add a regex for square brackets from ConfigObject parser and ignore the entire group (and log a warning).
I'll' take a look.

ronso0:
> It is just that the "on the fly changing" use pattern might require an extra safety net and shall not crash "delayed".

Note that it does not crash in release builds without DEBUG_ASSERTIONS_FATAL. When it hits a debug assertion in [`ControlDoublePrivate::getControl`](https://github.com/mixxxdj/mixxx/blob/eadced77534da358065aec66e816707431ec4eb5/src/control/control.cpp#L128) it 'just' logs a critical warning.

A regex filter is here #16400

daschuer:
Thank you for the nice checker.

Asserts are to verify assumptions that can never fail.
Since the keyboard mapping is kind of user data it can fail. So an assert is semanticsl not correct. 

How Is the same situation handled in skins?  I have not checked. I think we should treat them similar. 

Do we still have as log message when removing the assert?

ronso0:
Well, the assertion is in control.cpp, not the ConfigObject class.
Yes, we still get a qwarning as well a critical logged (due to the DEBUG_ASSERT).
To clarify, ControlDoublePrivate::getControl returns a nullptr and KeyboardEventFilter does nothing in that case.
Plus a "Warning: Keyboard key is configured for nonexistent control:"

In case of skins WPushButton connections use ControlProxies and in case of invalid controls it uses ControlDoublePrivate::getDefaultControl().
So also no danger there.

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -179,24 +246,228 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+    kLogger.debug() << "updateMenuBarActionShortcuts";
+    QHashIterator<QAction*, std::pair<ConfigKey, QString>> it(m_menuBarActions);
+    while (it.hasNext()) {
+        it.next();
+        auto* pAction = it.key();
+        VERIFY_OR_DEBUG_ASSERT(pAction) {
+            kLogger.warning() << "Could not find a valid action for" << it.value() << "- Ignoring";
+            continue;
+        }
+        const QString keyStr = m_pKbdConfig->getValue(it.value().first, it.value().second);
```
daschuer: This code is unclear, because of the anonymous pair. The best way would be to introduce a private struct instead. The quick hack solution is this:  
```suggestion
        const auto& [cfgKey, defaultShortcut] = it.value();
        const QString keyStr = m_pKbdConfig->getValue(cfgKey, defaultShortcut);
```
better: 
```
    struct DefaultShortcut {
        ConfigKey cfgKey;
        QString defaultShortcut;
    };
``

File: src/controllers/keyboard/keyboardeventfilter.cpp [hunk 37]
```diff
@@ -179,24 +246,228 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+        const QKeySequence ks = safeKeySequence(keyStr);
+        pAction->setShortcut(ks);
+    }
+}
+
+void KeyboardEventFilter::reloadKeyboardConfig() {
+    kLogger.debug() << "reloadKeyboardConfig, enabled:" << m_enabled;
```
daschuer, line 430: Do we want to silence (comment out) some of these?

src/controllers/keyboard/keyboardeventfilter.cpp
```diff
@@ -179,24 +246,228 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
...
+        keyboardFile = QString(resourcePath).append("keyboard/");
+        keyboardFile += m_locale.name();
+        keyboardFile += ".kbd.cfg";
+        if (QFile::exists(keyboardFile)) {
+            kLogger.debug() << "Found and will use default keyboard mapping" << keyboardFile;
+        } else {
+            kLogger.debug() << keyboardFile << " not found, using en_US.kbd.cfg";
+            keyboardFile = mappingFilePath(resourcePath, QStringLiteral("en_US"));
+            if (!QFile::exists(keyboardFile)) {
+                kLogger.debug() << keyboardFile << " not found, starting without shortcuts";
```
daschuer: is this actually a warning?

daschuer (changes requested):
OK so at leats the findings are no regression. Lets bring this in. 

I have left comets for a final polish.

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: Indeed. I also refined some other debug/warning messages

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: Tbh I'd rather keep them for debugging purpose.
If we notice they become annoying even though they aren't logged that often, we can still remove them.

src/controllers/keyboard/keyboardeventfilter.cpp
ronso0: Yes, good idea. Works just fine!

daschuer (approved):
LGTM, thank you.

ronso0:
:octopus: thanks for the review!

PR #13086, opened 2024-04-10
Add stem controls

acolombier:
Depends on #13106 

This adds basic COs and UI for stem volume control. Manuel PR is [here](https://github.com/mixxxdj/manual/pull/644)

LateNight:

https://github.com/mixxxdj/mixxx/assets/7086688/fff5b4a9-c73b-4ee8-be44-4d1bbe38ba2e

QML: 

https://github.com/mixxxdj/mixxx/assets/7086688/bc4fa40b-d807-426b-8e6d-c47824950d22


JoergAtGithub:
Looks really cool!

I think we need a table with all planed stems controls and all existing stems hardware to ensure, that we can cover all

- Here you see the use of the Traktor S8 https://www.youtube.com/watch?v=MKJZO_0zQcM&t=1s
  - For the S8 a filter per stem and a filter enable control is additional needed
- Here the Rane Four https://www.youtube.com/watch?v=Pf3GQdkWByk&t=1060s
  - Has small text displays, that show the name of the stem and the effect-chain assigne
d to the stem
  - Has stem-split button which connects one deck with multiple mixer channels

acolombier:
> I think we need a table with all planed stems controls and all existing stems hardware to ensure, that we can cover all

One of the reasons why I added the stem feature flag is so we can iterate on features and controls, definitely worth having them layed out for reference tho! Shall we create an epic like we did for S4 screens? Could you help with that since I believe it requires special permission?

>   * Has small text displays, that show the name of the stem and the effect-chain assigned to the stem

That's the same with S4, which is part if the reason why I wanted to have the QML proxy in place. As it stands, you can already get the label for each stem.

>   * Has stem-split button which connects one deck with multiple mixer channels

I'll have to look how that works and see if that something we can support.

daschuer:
I wonder where unity is on the stem knobs. My original expectation is to put it on center like EQs.
This allows to accentuate a stem, compared to others and allows switching between EQ and Stem knobs without changing scale. 

Do we want to keep this switching in the mapping domain, or do we want to add four mixer meta knob that can be mapped and will switch between EQ and Stem. 

By the way we had a similar discussion with the fourth EQ band, gain and filter knob.

acolombier:
> > * Has stem-split button which connects one deck with multiple mixer channels
> 
> I'll have to look how that works and see if that something we can support.

I just watched the review - that piece of kit looks very good btw - the split button seems to perform a deck clone, and just mute the vocal, then mute everything else on the newly clone deck. IIRC, the NI stem standard suggests that the drum should always be first and the vocal always be last, so I believe this would be pretty trivial to do! We might need new CO to target deck when cloning tho (I believe at the moment, it pick a candidate out the last focus deck)

vantaboard:
> That's a good design, but perhaps this should come a second iteration, when the waveform also supports multiple stem layer? I can imagine the selection/hover of the stem button will "highlight" the stem signal on the waveform, which may be very handy to identify section of track with a specific feature on, say no vocals or instrumental solo
Yeah, I agree. Definitely giving the ability to visualize the layered stems would be great, especially if we could have the waveform color in alignment with the stem color.

> Scratching requires transport/playback control which would go against the logic that was implemented here, so definitely won't be possible. Here again tho, you can clone the deck and scratch just a stem (or set of them) on the individual deck, similar to the [demo](https://www.youtube.com/watch?v=Pf3GQdkWByk&t=1060s) @JoergAtGithub shared above on Serato (?)
Same with scratching and EQ, I agree. I guess I just find cloning a bit cumbersome at the moment. Perhaps in another iteration we can cue stems for cloning, so that a clone from deck A to deck B with a cue on vocals would mute the vocals on deck A and solo them onto B with seamless playback. This would be ideal because there are some transitions where I want to continue some portion of the music into a new track, but cloning feels clunky at the moment.

vantaboard:
I think I have a tendency to want to avoid creating a similar design to what @JoergAtGithub posted  https://youtu.be/Pf3GQdkWByk?t=1555 because the doubling up can be really annoying and it isn't optional even with cloning in the current state.

acolombier:
> Yeah, I agree. Definitely giving the ability to visualize the layered stems would be great, especially if we could have the waveform color in alignment with the stem color.

That will come in future PR tho! I've been trying to keep PR shot to help ease and speed up the review. 

> I guess I just find cloning a bit cumbersome at the moment. Perhaps in another iteration we can cue stems for cloning, so that a clone from deck A to deck B with a cue on vocals would mute the vocals on deck A and solo them onto B with seamless playback. 

Yes, that's the plan, very similar to the "split stem" feature from that demo. This way you easily "dive" in the track stem and have complete freedom on your mixing, without convoluting the deck option and control

> the doubling up can be really annoying and it isn't optional even with cloning in the current state.

Agreed. When we would look into this, I think we will make sure that no doubling up gets hearable - decks get cloned but stems are exclusive to avoid that gain issue (say A plays 0 1 2, B plays 3).

vantaboard:
> > EQ would apply to only the selected stems.
> 
> Not sure this is something we will be able to do due to performance impact. EQ will likely remain applicable to the deck as whole. This means you can still spare a second deck for the voice and adjust the EQ there

Is EQ for the whole deck still accessible somewhere in your design? I'm wondering if it even makes sense to be able to control volume for each stem in that case. Maybe my design would be mute/unmute for the different stems rather than selection.

mixxx stem idea

vantaboard:
Is the skin you're using available anywhere? I have built the branch, but I don't have any options to load a stem.

JoergAtGithub:
What the Rane Four (and other Serato stem controllers) do is bassically a mapping between stem channels playing in any deck and mixer channels. Similar, to what we do with effect units. But exposing this in the UI  similar to the FX1,FX2 buttons might be overwhelming.
Another approach would be to visually merge 2 decks if the user presses the Stem Split button on the controller. If there is only one deck, there can't be a conflict of two decks outputs connected to the same mixer channel.

daschuer:
Does the Rane Four use transport controls of deck one for deck three in split mode? That sounds kind of reasonable if you want to scratch the whole Track. Maybe we can consider two split modes like coupled and Independent. That is almost like our sync scratch mode we have abandoned for Mixxx 2.4

acolombier:
> Is the skin you're using available anywhere? I have built the branch, but I don't have any options to load a stem.

This is the new UI which isn't yet production ready and is lacking some basic features. You can see see by building with `QT6` and `QML`, then run mixxx with `--qml`

JoergAtGithub:
> Does the Rane Four use transport controls of deck one for deck three in split mode?

In Sertao DJ this is not the case, in the Rane Four video, I posted already above, you can see scratching of the deck with the vocal stem, while the deck with the instrumental stems is playing at normal speed: https://www.youtube.com/watch?v=Pf3GQdkWByk&t=1060s

> Maybe we can consider two split modes like coupled and Independent.

Might make sense, we should write down the use cases first.

acolombier:
So I guess my suggestion on creating a variant of clone deck might make sense, right? I assume that the split stem feature only work if the sibling deck is free, so we could do something similar.

> Might make sense, we should write down the use cases first.

That's a fair point. Could you put this into an issue maybe so we can gather all different usecase to expect around stem? I've been trying to keep the stem PRs fairly atomic - one step at a time - so I won't be doing any "advanced" controls in this PR. That "split stem" is definitely something I will look into as that definitely would work well on my S4 Mk3

JoergAtGithub:
We should clarify in general, where to put the documentation of the GSoC projects.

Maybe a STEM-Controls proposal at https://github.com/mixxxdj/proposals might be also suitable here?

acolombier:
> We should clarify in general, where to put the documentation of the GSoC projects.

Not sure I get your point here. The documentation of the STEM feature will go in https://github.com/mixxxdj/manual, like any other features, same with the additional controls.
If you mean more like a set of requirement to make a GSoC project successful, I think this is neither place to discuss this nor the time:  project plan was already written down and submitted two weeks ago. That's not to say it is set in stone and nothing can be added. But again, I would appreciate to iterate on this and not make the same mistakes again as per with some other of my contributions in Mixxx.

> Maybe a STEM-Controls proposal at https://github.com/mixxxdj/proposals might be also suitable here?

Sure, either way works. I guess the proposal project was more here for large piece of disruptive feature, while STEM is more like an additional feature with a  minimal amout of refactoring. That being said, any ways to list and describe devices and routines suggestion around STEM feature would work.

acolombier:
Added some controls on the legacy UI

https://github.com/mixxxdj/mixxx/assets/7086688/fff5b4a9-c73b-4ee8-be44-4d1bbe38ba2e

I'd be curious to hear your thought on it!

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,164 @@
...
+        setAutoFillBackground(true);
+        setPalette(pal);
+    });
+}
+
+void WStemControl::setStem(const StemInfo& stemInfo) {
+    QPalette pal = palette(); // we have to copy out the palette to edit it
+                              // since it's const (probably for threadsafety)
+    QColor backgroundColor = stemInfo.getColor();
+    backgroundColor.setAlphaF(m_mutedCo->get() == 1 ? 0.5 : 1.0);
```
JoergAtGithub, line 137:
```suggestion
    backgroundColor.setAlphaF(m_mutedCo->get() == 1 ? 0.5 : 1.0);
    pal.setColor(backgroundRole(), backgroundColor);
```
Without this, the stem controls appear white.

JoergAtGithub (changes requested):
The Windows build fails, because:

1. FFMPEG is disabled in the CMake settings and therefore STEMS
2. There are `#ifdef __STEM__` statements missing at various places in the code

acolombier:
Thanks for the pre-review @JoergAtGithub!
Will look into this.

src/widget/wstemcontrol.cpp
acolombier: Ah thanks, failed copy pasta :D

JoergAtGithub:
The floating stem widgets behave not completly as the should. Have a look on the following videos:


https://github.com/mixxxdj/mixxx/assets/64457745/2cca4dc5-2f9b-4fdc-8589-a2778f53cf0b



https://github.com/mixxxdj/mixxx/assets/64457745/387ba032-38a2-4d7e-92ca-57e5e69d41a5

JoergAtGithub:
Yes, it fixes the issues shown in the videos! Only a minor glitch remains: When I resize the window, than the floating widgets appear at the new position, a moment before the window is resized.

JoergAtGithub:
It does not compile anymore, if __STEM__ is not set.

acolombier:
> Yes, it fixes the issues shown in the videos! Only a minor glitch remains: When I resize the window, than the floating widgets appear at the new position, a moment before the window is resized.

Ah that's annoying. I'm not sure how we could fix that as I assume it is due to the delay of the main windows to redraw. Technically, the stem control is an independent window. Any suggestion on mitigation? I guess we could hide it till the main window has redrawn with the new size but that might look glitchy too. 

Is the stem control displaying as a window in Windows? Do you see it is listed in the toolbar? On Wayland with X11 bridge, it isn't listed, but still shows on certain operation such as `ALT+SUPER+TAB` (cycling between application windows) which I don't know how to workaround yet.

> It does not compile anymore, if `__STEM__` is not set.

Yep, I didn't make the fix backward compatible just yet, I wanted to hear some feedback to see if I was going on right track. Is there any objection on making the main window a singleton btw? (needed for the stem control to install the event filter)

JoergAtGithub:
I would call it annoying, but it's noticeable. The widget window does not appear in the Windows task bar.
When this PR reaches ReadyForReview status, I recommend isolating the floating widgets in a separate PR. Everything else in this PR, including the waveforms, seems to be a no-brainer to me.

acolombier:
> I recommend isolating the floating widgets in a separate PR.

Quick STEM FX depends and extend that floating widget. Would you suggest I also take it off that PR and just make "floating widget" a PR that come after both and which adds Quick FX control directly? 
I guess I can keep floating widget in for now and remove last minute if people want to test the feature without having to create custom mapping.

JoergAtGithub:
Yes keep it in for now! But I recommend to split it, when the other PRs are merged and this will be ReadyForReview.

acolombier:
Please not that AllShader textured are currently broken because they don't take the new waveformdata structure yet.

JoergAtGithub:
This PR contains several different functionalities (waveforms, controls, QML and classic UI). For a faster review, I recommend to split it into smaller parts.

acolombier:
> For a faster review, I recommend to split it into smaller parts.

As an intermediary. I have rewritten the history into 4 commits: the two first one should go as part of this PR IMO (COs+Stem waveform renderer). The two last one should probably be discarded/moved to a draft PR once everyone had a chance to test the feature.

JoergAtGithub:
That makes sense! Thank you!

src/engine/channels/enginedeck.cpp
```diff
@@ -39,39 +46,110 @@ EngineDeck::EngineDeck(
...
+                stemIdx++) {
+            if (m_stemMute[stemIdx]->toBool()) {
+                continue;
+            }
+            float gain = static_cast<float>(m_stemGain[stemIdx]->get());
+            pOut[i] += pIn[stemCount * i + mixxx::kEngineChannelOutputCount * stemIdx] * gain;
+            pOut[i + 1] +=
+                    pIn[stemCount * i +
+                            mixxx::kEngineChannelOutputCount * stemIdx + 1] *
+                    gain;
```
acolombier, line 136: I expect suggestion here about factoring this code to help vectorization and move it into `SampleUtil`. Please note this is already done as part of the [next iteration](https://github.com/acolombier/mixxx/blob/feat/add-quick-fx-on-stem/src/engine/channels/enginedeck.cpp#L144)

Eve00000:
Antoine, how can I get this controls-pr on my pc to build a test version?

src/waveform/renderers/allshader/waveformrendererstem.cpp
```diff
@@ -0,0 +1,202 @@
...
+            halfBreadth - 0.5f * devicePixelRatio,
+            static_cast<float>(length),
+            m_isSlipRenderer ? halfBreadth : halfBreadth + 0.5f * devicePixelRatio);
+    m_colors.addForRectangle(0.f, 0.f, 0.f, 0.f);
+
+    const double maxSamplingRange = visualIncrementPerPixel / 2.0;
+
+    for (int pos = 0; pos < length; ++pos) {
+        for (int s = 0; s < 4; s++) {
+            for (int z = 0; z < 2; z++) {
```
JoergAtGithub, line 125:
```suggestion
    for (int pixelPos = 0; pixelPos < length; ++pixelPos) {
        for (int stemIdx = 0; stemIdx < kMaxSupportedStem; stemIdx++) {
            for (int chIdx = 0; chIdx < mixxx::audio::ChannelCount::stereo(); chIdx++) {
```
Please use self descriptive names for the indices - I'm still unsure if I got the mening right.

JoergAtGithub (commented):
First batch of comments

Eve00000:
Some study material: [https://www.native-instruments.com/en/products/traktor/dj-software/traktor-pro-4/](url) , Traktor 4 anouncement in my mailbox.

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include "widget/wknobcomposed.h"
+#include "widget/wlabel.h"
+
+namespace {
+// FIXME(XXX) this is a workaround to ensure that that knob keep the aspect
+// ratio as defined in its maxSize. This needs to be moved an implemented by the
+// component directly
+void ensureWidgetRatio(QLayout* layout) {
+    for (int i = 0; i < layout->count(); i++) {
+        auto w = layout->itemAt(i)->widget();
```
daschuer, line 31:
```suggestion
        auto* pWidget = layout->itemAt(i)->widget();
```

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,166 @@
...
+            continue;
+        }
+        float ratio = static_cast<float>(w->maximumWidth()) /
+                static_cast<float>(w->maximumHeight());
+        int maxSize = qMin(w->width(), w->height());
+        QSize currentSize = w->size();
+        QSize newSize;
+
+        if (ratio > 1) {
+            newSize = QSize(maxSize, static_cast<int>(static_cast<float>(maxSize) / ratio));
```
daschuer, line 42: The cast is redundant 
```suggestion
            newSize = QSize(maxSize, static_cast<int>(maxSize / ratio));
```

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+            newSize = QSize(static_cast<int>(static_cast<float>(maxSize) * ratio), maxSize);
```
daschuer, line 44:
```suggestion
            newSize = QSize(static_cast<int>(maxSize * ratio), maxSize);
```

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,166 @@
...
+// component directly
+void ensureWidgetRatio(QLayout* layout) {
+    for (int i = 0; i < layout->count(); i++) {
+        auto w = layout->itemAt(i)->widget();
+        if (!qobject_cast<WKnobComposed*>(w) && !qobject_cast<WKnob*>(w)) {
+            continue;
+        }
+        float ratio = static_cast<float>(w->maximumWidth()) /
+                static_cast<float>(w->maximumHeight());
+        int maxSize = qMin(w->width(), w->height());
```
daschuer, line 37: I do not understand this. Can you add a source code comment?

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,166 @@
...
+        w->resize(newSize);
+        layout->itemAt(i)->invalidate();
+    }
+}
+} // namespace
+
+WStemControlBox::WStemControlBox(
+        const QString& group, QWidget* parent)
+        : WWidgetGroup(parent), m_group(group), m_hasStem(false), m_displayed(true) {
+    auto pLayout = new QVBoxLayout(this);
```
daschuer, line 60:
```suggestion
    auto* pLayout = make_parented<QVBoxLayout>(this);
```

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    int stemCount = qMin(m_stemControl.size(),
+            static_cast<std::size_t>(stemInfo.size()));
+    for (int stemIdx = 0; stemIdx < stemCount; stemIdx++) {
+        m_stemControl[stemIdx]->setStem(stemInfo[stemIdx]);
+    }
+}
+
+void WStemControlBox::addControl(QWidget* control) {
+    auto pWidget = std::make_unique<WStemControl>(control, this, m_group, m_stemControl.size());
```
daschuer, line 107:
```suggestion
    auto* pWidget = std::make_unique<WStemControl>(control, this, m_group, m_stemControl.size());
```

src/main.cpp
```diff
@@ -65,9 +65,9 @@ int runMixxx(MixxxApplication* pApp, const CmdlineArgs& args) {
         // This scope ensures that `MixxxMainWindow` is destroyed *before*
         // CoreServices is shut down. Otherwise a debug assertion complaining about
         // leaked COs may be triggered.
-        MixxxMainWindow mainWindow(pCoreServices);
+        MixxxMainWindow* mainWindow = MixxxMainWindow::createInstance(pCoreServices);
```
daschuer, line 68:
```suggestion
        MixxxMainWindow* pMainWindow = MixxxMainWindow::createInstance(pCoreServices);
```

src/qml/qmlstemsmodel.cpp
```diff
@@ -0,0 +1,67 @@
...
+}
+
+QmlStemsModel::QmlStemsModel(
+        QObject* pParent)
+        : QAbstractListModel(pParent) {
+}
+
+void QmlStemsModel::setStems(QList<StemInfo> stems) {
+    beginResetModel();
+    m_stems = QList<StemInfo>(std::move(stems));
```
daschuer, line 23:
```suggestion
    m_stems = std::move(stems);
```

src/skin/legacy/legacyskinparser.cpp
```diff
@@ -1021,6 +1030,24 @@ QWidget* LegacySkinParser::parseVisual(const QDomElement& node) {
...
+
+    QDomNode child = node.firstChildElement("StemControl");
+    if (!child.isNull()) {
+        setupSize(child, viewer->stemControlWidget());
+        setupConnections(child, viewer->stemControlWidget());
+        QDomElement stem = child.firstChildElement("Stem");
+        for (int i = 0; i < kMaxSupportedStem; i++) {
+            m_pContext->setVariable("StemGroup", group);
+            m_pContext->setVariable("StemIdx", QString::number(i + 1));
+            auto widget = parseWidgetGroup(stem);
```
daschuer, line 1044:
```suggestion
            auto* pWidget = parseWidgetGroup(stem);
```

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,166 @@
...
+    layout()->addWidget(pWidget.get());
+    m_stemControl.push_back(std::move(pWidget));
+}
+
+WStemControl::WStemControl(QWidget* widgetGroup, QWidget* parent, const QString& group, int stemIdx)
+        : WWidget(parent),
+          m_widget(widgetGroup),
+          m_mutedCo(std::make_unique<ControlProxy>(
+                  group, QString("stem_%1_mute").arg(stemIdx + 1))) {
+    auto pLayout = new QHBoxLayout(this);
```
daschuer, line 117:
```suggestion
    auto* pLayout = nmake_parented<QHBoxLayout>(this);
```

src/qml/qmlplayerproxy.h
```diff
@@ -140,6 +156,9 @@ class QmlPlayerProxy : public QObject {
     TrackPointer m_pCurrentTrack;
     QmlBeatsModel* m_pBeatsModel;
     QmlCuesModel* m_pHotcuesModel;
```
daschuer, line 158:
```suggestion
    parented_ptr<QmlCuesModel> m_pHotcuesModel;
```

daschuer (changes requested):
Thank you I have left some comments

src/engine/channels/enginedeck.cpp
```diff
@@ -39,39 +46,110 @@ EngineDeck::EngineDeck(
...
+
+#else
+            mixxx::audio::ChannelCount::stereo()
+#endif
+    );
+
+#ifdef __STEM__
+    if (!primaryDeck) {
+        return;
+    }
```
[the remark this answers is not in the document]
acolombier: No, this is an early return as we don't want to add stem controls on non-primary deck (sampler and preview), since they are not stem-capable due to performance reason. As part of #13268, non-primary deck will be able to load individual channels.

src/mixer/basetrackplayer.cpp
```diff
@@ -278,6 +279,20 @@ BaseTrackPlayerImpl::BaseTrackPlayerImpl(
...
+    if (!primaryDeck) {
+        return;
+    }
+
+#ifdef __STEM__
+    m_pStemColors.reserve(kMaxSupportedStem);
+    for (int i = 0; i < kMaxSupportedStem; i++) {
+        m_pStemColors.emplace_back(std::make_unique<ControlObject>(
+                ConfigKey(getGroup(), QStringLiteral("stem_%1_color").arg(i + 1))));
+        m_pStemColors.back()->setReadOnly();
```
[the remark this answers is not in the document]
acolombier: That sounds like a nice to have indeed - I will add that to the epic.
Changing the colour should be pretty straight forward as it doesn't requires to recompute the offset in the mp4 container. Changing label is quite a headache tho. It's not impossible but it requires to handle that offset update.

src/waveform/renderers/allshader/waveformrendererstem.cpp
acolombier: `z` stands for `z-index` - signal is drawn twice to help visualising the maximum signal when the stem volume isn't 100% or muted.
I added a comment.

src/widget/wstemcontrol.cpp
acolombier: This is a nasty hack to try and keep the knob ratio constant to 1, in this case `maxSize` takes the value of the lower dimension.
I'm planing to remove `wstemcontrol.cpp` once this PR is ready as there is a few issue with the UI, so this can be moved in a subsequent PR/dropped entirely.

src/engine/channels/enginedeck.cpp
```diff
@@ -7,8 +7,15 @@
 #include "engine/enginepregain.h"
 #include "engine/enginevumeter.h"
 #include "moc_enginedeck.cpp"
+#include "track/track.h"
 #include "util/sample.h"
 
+#ifdef __STEM__
+namespace {
+constexpr int kMaxSupportedStem = 4;
```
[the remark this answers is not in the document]
acolombier: `kMaxSupportedStem` is used everywhere else in the code. I used to have an instance of `kMaxStemCount` but I replaced it because i thought `kMaxSupportedStem` had a better meaning. Happy to replace them all if you thing `kMaxStemCount` sounds best.

src/analyzer/analyzerwaveform.cpp
```diff
@@ -142,6 +145,17 @@ bool AnalyzerWaveform::shouldAnalyze(TrackPointer tio) const {
         }
     }
 
+#ifdef __STEM__
+    // If the waveform was generated without stem information, we need to regenerate
+    if (!missingWaveform &&
+            ((!pTrackWaveform.isNull() &&
+                     pTrackWaveform->hasStem() != bIsStemTrack) ||
```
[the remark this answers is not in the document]
acolombier: The opposite is also relevant. During my testing phase, as I was testing [my stemgen](https://github.com/acolombier/stemgen) by the same occasion, I had case where this last one was generating the `stem` atom correctly, and so the track library should be updated for `false != true` and `true != false` only.

src/qml/qmlplayerproxy.h
acolombier: This will require further code change. I will update it in another PR.

JoergAtGithub: This line is not added in this PR - and not related to stems.
This is prior QML work, where we decided that we will allow to get less elaborated PRs merged, to accelerate the overall QML development. Thanks Antoine, for your offer to improve this code in a future PR.

src/widget/wstemcontrol.cpp
```diff
@@ -0,0 +1,166 @@
...
+#include "widget/wknobcomposed.h"
+#include "widget/wlabel.h"
+
+namespace {
+// FIXME(XXX) this is a workaround to ensure that that knob keep the aspect
+// ratio as defined in its maxSize. This needs to be moved an implemented by the
+// component directly
+void ensureWidgetRatio(QLayout* layout) {
+    for (int i = 0; i < layout->count(); i++) {
+        auto pWidget = layout->itemAt(i)->widget();
```
JoergAtGithub, line 31: Please fix the clang-tidy warning here

src/engine/channels/enginedeck.cpp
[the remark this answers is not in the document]
daschuer: The main reason I have stumbled here was the missing plural s. If you add that, I am fine.

acolombier:
@daschuer @JoergAtGithub I took the freedom to remove the legacy UI wiget as it looks like we are getting closer to merge this PR. If you need to do more testing, the widget is available [here](https://github.com/mixxxdj/mixxx/pull/13515).

I have kept the QML part as this one is pretty functional. Arguably, we may want to change the layout in the future, but the fact that nothing happens in QMl for the last 6 months makes me sad so I'd like to have stem in there, if that's okay! :)

JoergAtGithub:
The code in this PR is LGTM now, but I miss test cases for the new COs. could you please add some basic tests for them.

acolombier:
@JoergAtGithub I added some tests now that should cover all new stem COs

daschuer (approved):

daschuer:
Thank you for the next great step. 

Do we have plans regarding a widget based skin? 
I am afraid without it, it will take a while until stem support is usable for all users.

acolombier:
We've been discussing that in the draft PR with [the widget based skin feature](https://github.com/mixxxdj/mixxx/pull/13515). The TL;DR is no easy way to support all themes correctly, but we could release the stem control under some kind of "opt-in" alpha and address some of the common issues with it

daschuer:
Ah Ok. Yes and even a single Stem skin would be reasonable. Maybe it attracts other skin designers to help with the rest.

src/test/stemcontrolobjecttest.cpp
```diff
@@ -0,0 +1,236 @@
...
+        m_pStem4Volume = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_4_volume");
+        m_pStem1Mute = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_1_mute");
+        m_pStem2Mute = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_2_mute");
+        m_pStem3Mute = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_3_mute");
+        m_pStem4Mute = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_4_mute");
+        m_pStem1Color = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_1_color");
+        m_pStem2Color = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_2_color");
+        m_pStem3Color = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_3_color");
+        m_pStem4Color = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_4_color");
+        m_pStemCount = std::make_unique<PollingControlProxy>(m_sGroup1, "stem_count");
```
JoergAtGithub, line 35: Before I merge this PR, which introduces new COs, I would like to hear your opinion about the naming scheme of the stem COs ( the full list can be found in the description of https://github.com/mixxxdj/mixxx/issues/13116 ).
Do you prefer the style:
`[Channel1],stem_1_volume` or `[Channel1Stem1],volume`
The first has the benefit, that the group name is similar to other COs in the mappings,
the second would have the benefit, that we seperate channel and function clearer and we could add it as just another group syntax to the documentation of some existing COs.
The first syntax would be in line with `[Channel1],hotcue_X_color`, the second with `[QuickEffectRack1_[ChannelXStemY]],super1`.
If we go with the first syntax, than I think `[QuickEffectRack1_[Channe1Stem1]],super1` should become `[QuickEffectRack1_[Channel1]],stem1_super1` for consistency.
What's your opinion?

acolombier: Honestly, I think using `[ChannelXStemY]` globally would make a lot of sense, but I remember stumbling across something and settling for `[ChannelX],stem_Y_...`. 
Now that my Mixxx knowledge has grew, I will retry and see if I can get `ChannelXStemY` working for consistency.

acolombier: Ok, that was surprisingly easy finally, I do not know how come I didn't do this before :sweat_smile: 
Also updated [the manual](https://github.com/mixxxdj/manual/pull/644) and will update the epic right away.

src/mixer/basetrackplayer.cpp
```diff
@@ -286,9 +286,13 @@ BaseTrackPlayerImpl::BaseTrackPlayerImpl(
...
 #ifdef __STEM__
     m_pStemColors.reserve(kMaxSupportedStems);
+    QString group = getGroup();
     for (int stemIdx = 1; stemIdx <= kMaxSupportedStems; stemIdx++) {
+        QString stemGroup = QStringLiteral("%1Stem%2]")
+                                    .arg(group.left(group.size() - 1),
+                                            QString::number(stemIdx));
         m_pStemColors.emplace_back(std::make_unique<ControlObject>(
-                ConfigKey(getGroup(), QStringLiteral("stem_%1_color").arg(stemIdx))));
+                ConfigKey(stemGroup, QStringLiteral("track_color"))));
```
JoergAtGithub, line 295: To my understanding, the existing `track_color` CO does something different. It colors the track in the library. Therefore just `color` or something more distinguishable like `stem_color` or `waveform_color` seems better suited here.

acolombier: I went with`color`!

JoergAtGithub (approved):

JoergAtGithub:
Thank you very much!

acolombier:
Thanks for your review! Next stop is either #13268 or #13123

src/test/stemcontrolobjecttest.cpp
```diff
@@ -0,0 +1,254 @@
...
+    m_pStem3Volume->set(0.2);
+    m_pStem4Volume->set(0.4);
+
+    m_pEngineMixer->process(kMaxEngineChannels * kMaxEngineFrames);
+    assertBufferMatchesReference(m_pEngineMixer->getMainBuffer(),
+            kProcessBufferSize,
+            "StemVolumeControlFull");
+}
+
+TEST_F(StemControlTest, VolumeResetOnLoad) {
```
ronso0, line 174: For some reason this test failed here
https://github.com/mixxxdj/mixxx/actions/runs/10981245295/job/30488031679?pr=13077#step:11:1823
(in case the link doesn't work, the PR is #13077)

any idea why @acolombier ?

acolombier: No idea. It looks like it is failing with an unrelated assertion, not quite sure why this would happen

ronso0: oh, I see it's loading one of the dummy tracks. sorry for the noise then

PR #13088, opened 2024-04-11
(fix) Slider tooltip: consider orientation for up/down shortcut tooltips + add support for WKnobComposed

ronso0:
Horizontal sliders were using the keyboard shortcut hints of the default (vertical) slider.
This was broken by 5f86ed50007f7a21a21bd93469bdc5aff7be253e

### Fix
Parse the 'Horizontal' property separately so it's available in `LegacySkinParser::setupConnections()` already (which does the shortcut setup), before WSliderComposed::setup() is called. setup() parses again, but this time with SkinContext (variable support, for whoever needs variable orientation in a slider template...)
Done with a stripped conglomerate of SkinContext member functions (or we add a dedicated static?)
(I have no patience to check whether 5f86ed50007f7a21a21bd93469bdc5aff7be253e can partially be reverted without breaking something else)
There's an alternative which requires passing a SkinContext pointer (unneeded for 99% of the widgets) which I may outline if desired.

### By-catch:
Add shortcut tooltip support for WKnobComposed. (I though we had this before at some point? git history says No)

### DIFF
--- a/src/skin/legacy/legacyskinparser.cpp
+++ b/src/skin/legacy/legacyskinparser.cpp
@@ -2404,7 +2404,7 @@ void LegacySkinParser::setupConnections(const QDomNode& node, WBaseWidget* pWidg
                     ConfigKey subkey;
                     QString shortcut;
 
-                    if (pSlider->isHorizontal()) {
+                    if (pSlider->tryParseHorizontal(node)) {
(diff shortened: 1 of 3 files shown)


daschuer (approved):
LGTM Thank you.

PR #13089, opened 2024-04-11
Don't override XML mapping when using `makeInputHandler`

christophehenry:
Follow up of #12781. This PR is designed to prevent overriting an existing XML mapping when when using `makeInputHandler`.

### DIFF
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -625,14 +625,29 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
         auto mStatusError = QStringLiteral(
                 "Invalid status or midino passed to midi.makeInputHandler. "
                 "Please pass a strictly positive integer. status=%1,midino=%2")
-                                    .arg(status, midino);
+                                    .arg(status)
+                                    .arg(midino);
 
         getScriptEngine()->throwJSError(mStatusError);
         return QJSValue();
     }
 
+    const auto midiKey = MidiKey(status, midino);
+
+    auto it = m_pMapping->getInputMappings().constFind(midiKey.key);
+    if (it != m_pMapping->getInputMappings().constEnd() &&
+            it.value().options.testFlag(MidiOption::Script) &&
+            std::holds_alternative<ConfigKey>(it.value().control)) {
+        qCWarning(m_logBase) << QStringLiteral(
+                "Ignoring anonymous JS function for status=%1,midino=%2 "
+                "because a previous XML binding exists")
+                                        .arg(status)
+                                        .arg(midino);
+        return QJSValue();
+    }
+
     MidiInputMapping inputMapping(
-            MidiKey(status, midino),
+            midiKey,
             MidiOption::Script,
             std::make_shared<QJSValue>(scriptCode));
 


src/controllers/midi/midicontroller.cpp
```diff
@@ -631,6 +631,12 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
         return QJSValue();
     }
 
+    if (m_pMapping->getInputMappings().contains(status)) {
```
christophehenry, line 634: This just check if an existing mapping exists. Should it perform a more thourough search to check if existing binding is an XML type?

src/controllers/midi/midicontroller.cpp
```diff
@@ -631,6 +631,12 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
...
+        qDebug(m_logBase).nospace() << "Ignoring anonymous JS function for status "
+                                    << status << " because a previous binding exists";
```
christophehenry, line 636: Here, I just log a message. I was thinking maybe also log the script line. `getScriptEngine()->throwJSError()` indicates a line in the script. Is there a similar method to just log a warning?

christophehenry:
I didn't write any tests here. Should I?

src/controllers/midi/midicontroller.cpp
daschuer: Before, all bindings are from XML right? So the question is regarding to allow a second js binding. Is that your point? Do we have a use case for it? I think in such cases you could just put both features onto one callback.

src/controllers/midi/midicontroller.cpp
daschuer: My guess is: 
`getScriptEngine()->currentStackFrame->lineNumber()`

src/controllers/midi/midicontroller.cpp
daschuer: I wonder why `midino` is not relevant here?

daschuer (changes requested):
Thank you. Here some comments.

Swiftb0y (commented):
sorry, forgot to submit my review earlier

src/controllers/midi/midicontroller.cpp
Swiftb0y: that doesn't exists / is private, or at least it seems to be undocumented. where is this from?

daschuer: I have skimmed through the Qt source.

christophehenry: I can't find `getScriptEngine()->currentStackFrame`.

christophehenry:
> I wonder why `midino` is not relevant here?

You're right. That's fixed.

src/controllers/midi/midicontroller.cpp
```diff
@@ -625,14 +625,29 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
         auto mStatusError = QStringLiteral(
                 "Invalid status or midino passed to midi.makeInputHandler. "
                 "Please pass a strictly positive integer. status=%1,midino=%2")
-                                    .arg(status, midino);
+                                    .arg(status)
+                                    .arg(midino);
```
Swiftb0y, line 629: I think the previous version is more efficient. why did you change it?

src/controllers/midi/midicontroller.cpp
```diff
@@ -625,14 +625,29 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
...
+
+    auto it = m_pMapping->getInputMappings().constFind(midiKey.key);
+    if (it != m_pMapping->getInputMappings().constEnd() &&
+            it.value().options.testFlag(MidiOption::Script) &&
+            std::holds_alternative<ConfigKey>(it.value().control)) {
+        qCWarning(m_logBase) << QString(
+                "Ignoring anonymous JS function for status=%1,midino=%2 "
+                "because a previous XML binding exists")
+                                        .arg(status)
+                                        .arg(midino);
```
Swiftb0y, line 645:
```suggestion
        qCWarning(m_logBase) << QStringLiteral(
                "Ignoring anonymous JS function for status=%1,midino=%2 "
                "because a previous XML binding exists")
                                        .arg(status, midino);
```

Swiftb0y (commented):
lgtm otherwise

src/controllers/midi/midicontroller.cpp
christophehenry: Erf, no, [that works only for objects that can be casted to QString](https://doc.qt.io/qt-6/qstring.html#arg-14). When used with `(status, midino)`, C++ select [this one](https://doc.qt.io/qt-6/qstring.html#arg-5).

christophehenry: I can do:

```suggestion
        qCWarning(m_logBase) << QString(
                "Ignoring anonymous JS function for status=%1,midino=%2 "
                "because a previous XML binding exists")
                                    .arg(QString::number(status), QString::number(midino));
```

Swiftb0y: whoops, good point. yeah, then the double `arg` is probably fine. suboptimal Qt Api. Still, please wrap the string literal in a `QStringLiteral`

src/controllers/midi/midicontroller.cpp
christophehenry: Oh that's a mistake. Fixed.

src/controllers/midi/midicontroller.cpp
christophehenry: Done.

Swiftb0y (approved):
LGTM. wdyt @daschuer

daschuer (approved):
LGTM, Thank you.

PR #13092, opened 2024-04-11
Playlists: move tracks with Alt + Up/Down/PageUp/PageDown/Home/End

### LINKED ISSUE #10826: Simplify moving tracks in long playlists
> I have a long playlist and I would like to rearrange the order of the songs. I can do it by dragging and dropping but when the new position is pages up or down it gets a bit cumbersome. It would be nice to have some features like cut and paste.

https://mixxx.discourse.group/t/helper-for-moving-files-in-long-playlists/25574

Ideas:
* move selected tracks with Alt+Up/Down/PageUp/PageDown with `takeItem()` and `insertItem()`
* emulate QDragEnter/Move/LeaveEvent ? discouraged in Qt documentation..

ronso0:
Finally. Fixes #10826 
~~This was rather easy only due the new Cut, Copy & Paste feature #12020. Thanks @m0dB !!~~
**edit:** Redone entirely by using the mechanics of table-internal drag'n'drop.

Also fixes an issue that dropping tracks at the bottom of a playlist wouldn't select the dropped tracks nor give a current index for navigation.

Also includes a small fix for pasteTracks() (same: current index for navi)

Applicable only in (unlocked) playlists.
(sort by # otherwise it's rather unpredictable)

* continuous selections are moved right away
* ~~non-continuous selections are first consolidated at first/last row, then moved on next keypress~~ may not be desired
* non-continuous selections can be consolidated at the top/bottom with Up/PageUp and Down/PageDown respectively
* wrap-around is suppressed to avoid undesirable overshoot with longpress (keyrepeat)

### DIFF
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -884,13 +884,14 @@ void WTrackTableView::pasteTracks(const QModelIndex& index) {
         return;
     }
 
+    const auto prevIdx = currentIndex();
+
     const QList<int> rows = trackModel->pasteTracks(index);
     if (rows.empty()) {
         return;
     }
 
     updateGeometries();
-
     const auto lastVisibleRow = rowAt(height());
 
     // Use selectRow to scroll to the first or last pasted row. We would use
@@ -900,6 +901,13 @@ void WTrackTableView::pasteTracks(const QModelIndex& index) {
         selectRow(rows.back());
     } else {
         selectRow(rows.front());
+    }
+
+    const auto idx = prevIdx.siblingAtRow(rows.back());
+    QItemSelectionModel* pSelectionModel = selectionModel();
+    if (pSelectionModel && idx.isValid()) {
+        pSelectionModel->setCurrentIndex(idx,
+                QItemSelectionModel::SelectCurrent | QItemSelectionModel::Select);
     }
 
     // Select all the rows that we pasted
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -639,15 +639,15 @@ void WTrackTableView::dragMoveEvent(QDragMoveEvent * event) {
 
 // Drag-and-drop "drop" event. Occurs when something is dropped onto the track table view
 void WTrackTableView::dropEvent(QDropEvent * event) {
-    TrackModel* trackModel = getTrackModel();
+    TrackModel* pTrackModel = getTrackModel();
 
     // We only do things to the TrackModel in this method so if we don't have
     // one we should just bail.
-    if (!trackModel) {
+    if (!pTrackModel) {
         return;
     }
 
-    if (!event->mimeData()->hasUrls() || trackModel->isLocked()) {
+    if (!event->mimeData()->hasUrls() || pTrackModel->isLocked()) {
         event->ignore();
         return;
     }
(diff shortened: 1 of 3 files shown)


ronso0:
FWIW I think the `event->source() == this` part of `WTrackTableView::dropEvent()` can be replaced with cut/paste.
@m0dB wdyt?

ronso0:
Works, but draft again because I'll look into redoing moveTrack() -> moveTracks(QModelIndexList indices, int insertPos), moving all tracks at once.

nlw0:
That's great, I've tried out and it seems to work. I noticed that 'del' seems to have stopped working for me, though. Could it be related to this patch?

What would be a good way to implement library commands for this now, using `emitKeyEvent`?

ronso0:
> ...but I'll check if a wrapper could work, i.e. a function that can be called by dragDropEvent() and moveTracks()

Yup, that works!
Will pick that to get around the (presumably) costly cut/paste ops

ronso0:
> > ...but I'll check if a wrapper could work, i.e. a function that can be called by dragDropEvent() and moveTracks()
> 
> Yup, that works! Will pick that to get around the (presumably) costly cut/paste ops

Done. Works like a charm, espcially consolidating non-continuous selections.

ronso0:
And it avoids running into bug #13100 (or limitation once that is fixed) with tracks whose files are currently missing.

ronso0:
In case someone experiences issues, I can make a debu commit from the trace output I have in my stash.

src/widget/wtracktableview.cpp
```diff
@@ -902,13 +843,193 @@ void WTrackTableView::pasteTracks(const QModelIndex& index) {
...
+                QItemSelectionModel::Select | QItemSelectionModel::Rows);
+    }
+}
+
+void WTrackTableView::moveSelectedTracks(QKeyEvent* event) {
+    QList<int> selectedRows = getSelectedRowNumbers();
+    if (selectedRows.isEmpty()) {
+        return;
+    }
+    std::sort(selectedRows.begin(), selectedRows.end());
```
daschuer, line 981: Can we move this into getSelectedRowNumbers()? It looks like it is required in all cases.

src/widget/wtracktableview.cpp
```diff
@@ -837,6 +760,23 @@ QModelIndexList WTrackTableView::getSelectedRows() const {
...
+QList<int> WTrackTableView::getSelectedRowNumbers() const {
+    QItemSelectionModel* pSelectionModel = selectionModel();
+    VERIFY_OR_DEBUG_ASSERT(pSelectionModel != nullptr) {
+        qWarning() << "No selection model available";
+        return {};
+    }
+    const QModelIndexList indices = pSelectionModel->selectedRows();
+    if (indices.isEmpty()) {
+        return {};
+    }
```
daschuer, line 772: This is redundant, but you may keep it if you like. 
```suggestion
```

src/widget/wtracktableview.cpp
```diff
@@ -837,6 +760,23 @@ QModelIndexList WTrackTableView::getSelectedRows() const {
     return pSelectionModel->selectedRows();
 }
 
+QList<int> WTrackTableView::getSelectedRowNumbers() const {
+    QItemSelectionModel* pSelectionModel = selectionModel();
+    VERIFY_OR_DEBUG_ASSERT(pSelectionModel != nullptr) {
+        qWarning() << "No selection model available";
+        return {};
+    }
+    const QModelIndexList indices = pSelectionModel->selectedRows();
```
daschuer, line 769:
```suggestion
    const QModelIndexList indices = getSelectedRows();
```

daschuer (changes requested):
Some finding. A manual test is pending.

src/widget/wtracktableview.cpp
ronso0: of course ...

daschuer (approved):
Works like a charm. You may rebase it now. Thank you.

daschuer:
All green. Thank you.

Issue #11285, opened 2023-02-19
Unable to move library cursor from a controller when the Mixxx window loses the focus


acolombier:
## Summary 

When the Mixxx window loses focus, it is impossible to control the selected item in the library from a controller

## Actual behavior

Nothing happens when trigger control from the `[Library]` section (e.g: `MoveVertical`, `focused_widget`)

## Expected behavior

It should still be possible to inetract with the currently focus widget and change the selected item using `[Library]` controls

## How to reproduce

- Map a library focus action on a controller. Example: `[Library] MoveVertical 1`
- Ensure it works while Mixxx window has focus
- Focus an other window
- Try the mapped control again

## Initial exploration 

It looks like the problem could come from the file `src/library/librarycontrol.cpp`. My understanding is that the control focus is closely linked with the UI widget focus. The action of moving the cursor seems to emulate a [keyboard input](https://github.com/mixxxdj/mixxx/blob/1f4088fa6aa23e70eb07c05dfdee2387f31ee08d/src/library/librarycontrol.cpp#L679), which only works while the window has the focus.

I'm happy to explore this further or try to fix it, if you guys have an idea or how I could do that with a minimal disruption. 

### Version

2.4.0~alpha2~1403~gcce89d09f3-1~jammy (Nightlies PPA), main

### OS

PopOS 22.04

ronso0:
Yes, the Move... controls are supposed to be universal, i.e. they can be used in any widget that has keyboard focus incl. dialogs.
So if no Mixxx window is focused these controls are not applicable. That's what the dedicated widget controls are for, e.g. the [[Playlist] controls](https://manual.mixxx.org/2.3/en_gb/chapters/appendix/mixxx_controls.html#the-playlist-controls) and (in Mixxx 2.4) the [searchbox controls](https://manual.mixxx.org/2.4/en_gb/chapters/appendix/mixxx_controls.html#control-[Library]-search_history_next)

> which only works while the window has the focus

No, unfoused widgets can also receive keystrokes, for example run Mixxx from the command line with `--deveoper` and set a value for `[Playlist], SelectPlaylist` in the Developer Tools window (Ctrl+Shift+t).

However, IIRC it has already been discussed how to make the Move... controls work while _no_ Mixxx window has focus. It's questionable whether this is worth the effort implementing this for the legacy skins instead of working on the new QML skin system. Though, if it's doable without too many hacks, that could IMHO be the last UI tweak before switching to QML.

[...]

PR #13098, opened 2024-04-14
Add track move commands to controls library

nlw0:
Add track move to controls library by emulating alt+up/down. Depends on #13092

### DIFF
--- a/src/library/librarycontrol.cpp
+++ b/src/library/librarycontrol.cpp
@@ -167,6 +167,30 @@ LibraryControl::LibraryControl(Library* pLibrary)
                 &LibraryControl::slotMoveFocus);
     }
 
+    // Controls to move tracks on playlist
+    // Track move controls (emulate Alt+up/down button press)
+    m_pMoveTrackUp = std::make_unique<ControlPushButton>(ConfigKey("[Library]", "MoveTrackUp"));
+    m_pMoveTrackDown = std::make_unique<ControlPushButton>(
+            ConfigKey("[Library]", "MoveTrackDown"));
+    m_pMoveTrack = std::make_unique<ControlEncoder>(ConfigKey("[Library]", "MoveTrack"), false);
+#ifdef MIXXX_USE_QML
+    if (!CmdlineArgs::Instance().isQml())
+#endif
+    {
+        connect(m_pMoveTrackUp.get(),
+                &ControlPushButton::valueChanged,
+                this,
+                &LibraryControl::slotMoveTrackUp);
+        connect(m_pMoveTrackDown.get(),
+                &ControlPushButton::valueChanged,
+                this,
+                &LibraryControl::slotMoveTrackDown);
+        connect(m_pMoveTrack.get(),
+                &ControlEncoder::valueChanged,
+                this,
+                &LibraryControl::slotMoveTrack);
+    }
+
     // Direct focus control, read/write
     m_pFocusedWidgetCO = std::make_unique<ControlPushButton>(
             ConfigKey("[Library]", "focused_widget"));
@@ -219,7 +243,7 @@ LibraryControl::LibraryControl(Library* pLibrary)
     }
 
     // Auto DJ controls
-    m_pAutoDjAddTop = std::make_unique<ControlPushButton>(ConfigKey("[Library]","AutoDjAddTop"));
+    m_pAutoDjAddTop = std::make_unique<ControlPushButton>(ConfigKey("[Library]", "AutoDjAddTop"));
     m_pAutoDjAddTop->addAlias(ConfigKey(
             QStringLiteral("[Playlist]"), QStringLiteral("AutoDjAddTop")));
 #ifdef MIXXX_USE_QML
@@ -232,7 +256,8 @@ LibraryControl::LibraryControl(Library* pLibrary)
                 &LibraryControl::slotAutoDjAddTop);
     }
 
-    m_pAutoDjAddBottom = std::make_unique<ControlPushButton>(ConfigKey("[Library]","AutoDjAddBottom"));
+    m_pAutoDjAddBottom = std::make_unique<ControlPushButton>(
+            ConfigKey("[Library]", "AutoDjAddBottom"));
     m_pAutoDjAddBottom->addAlias(ConfigKey(
             QStringLiteral("[Playlist]"), QStringLiteral("AutoDjAddBottom")));
 #ifdef MIXXX_USE_QML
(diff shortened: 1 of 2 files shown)


src/library/librarycontrol.cpp
```diff
@@ -793,6 +816,25 @@ void LibraryControl::slotMoveFocus(double v) {
...
+    if (v > 0) {
+        slotMoveTrack(-1);
+    }
+}
+
+void LibraryControl::slotMoveTrack(double v) {
+    const auto key = (v < 0) ? Qt::Key_Up : Qt::Key_Down;
+    const auto times = static_cast<unsigned short>(std::abs(v));
+    emitKeyEvent(QKeyEvent{
+            QEvent::KeyPress, key, Qt::AltModifier, QString(), false, times});
```
ronso0, line 835: unfortunately emitKeyEvent requires the library to be focused, which would limit the use cases (same with the current navigation controls, see #9294, #11285 and others).

Let's send the key event to `m_pLibraryWidget` directly as this is the only widget that would accept this event (check if is not null)

ronso0 (changes requested):
Thanks for this PR!

Code LGTM, just a comment

src/library/librarycontrol.cpp
nlw0: All right, I couldn't test yet, does the code look right?

nlw0:
Sure. I believe I signed it already, though.

src/library/librarycontrol.cpp
```diff
@@ -166,6 +166,29 @@ LibraryControl::LibraryControl(Library* pLibrary)
                 &LibraryControl::slotMoveFocus);
     }
 
+    // Controls to move tracks on playlist
+    // Track move controls (emulate Alt+up/down button press)
+    m_pMoveTrackForward = std::make_unique<ControlPushButton>(ConfigKey("[Library]", "MoveTrackForward"));
+    m_pMoveTrackBackward = std::make_unique<ControlPushButton>(ConfigKey("[Library]", "MoveTrackBackward"));
+    m_pMoveTrack = std::make_unique<ControlEncoder>(ConfigKey("[Library]", "MoveTrack"), false);
```
ronso0, line 173: I suggest we make it
MoveTracks | ~Up | ~Down

just FYI: we aim to use snake_case for new controls.
Here PascalCase is okay as it it's consistent with the closely related controls.

src/library/librarycontrol.cpp
```diff
@@ -793,6 +816,31 @@ void LibraryControl::slotMoveFocus(double v) {
...
+    }
+}
+
+void LibraryControl::slotMoveTrackBackward(double v) {
+    if (v > 0) {
+        slotMoveTrack(-1);
+    }
+}
+
+void LibraryControl::slotMoveTrack(double v) {
```
ronso0, line 831: Please add a  /// comment here explaining that these moves equal Alt + Up/Down.

ronso0 (changes requested):
LGTM, just the names need to be changed.

src/library/librarycontrol.cpp
nlw0: done

src/library/librarycontrol.cpp
nlw0: done

src/library/librarycontrol.cpp
ronso0: > MoveTracks | ~Up | ~Down

Ha, sorry that was not clear, I meant:
MoveTrack**s**
MoveTrack**s**Up
MoveTrack**s**Down

ronso0:
pre-commit is complaining (didn't take a closer look)

ronso0:
Sorry, my mistake:
`m_pLibraryWidget` is the stacked widget that holds all WTrackTableViews, it won't handle the keypress.

I'll prepare a commit that allows to send the keypress to the current view (if it's a WTrackTableView).

src/library/librarycontrol.cpp
```diff
@@ -166,6 +166,30 @@ LibraryControl::LibraryControl(Library* pLibrary)
                 &LibraryControl::slotMoveFocus);
     }
 
+    // Controls to move tracks on playlist
+    // Track move controls (emulate Alt+up/down button press)
+    m_pMoveTrackUp = std::make_unique<ControlPushButton>(ConfigKey("[Library]", "MoveTrack-Up"));
+    m_pMoveTrackDown = std::make_unique<ControlPushButton>(
+            ConfigKey("[Library]", "MoveTrack-Down"));
```
ronso0, line 173: Let's rename so the controls match the scheme of `MoveUp`, `ScrollUp` etc.
```suggestion
    m_pMoveTrackUp = std::make_unique<ControlPushButton>(ConfigKey("[Library]", "MoveTrackUp"));
    m_pMoveTrackDown = std::make_unique<ControlPushButton>(
            ConfigKey("[Library]", "MoveTrackDown"));
```

ronso0 (changes requested):

ronso0:
If you adopt my code suggestion, then rebase (and squash commits) onto main then there is `WLibrary::getCurrentTrackTableView()` available, and you can pick (and also squash) this commit
ab8c1d21ae
Works nicely!

ronso0:
@nlw0 Do you have time to rebase, cherry-pick and squash as I suggested above?

Let us know if you don't, then I'd take over soonish in order to get this into 2.5 along with the keyboard improvement.

nlw0:
Hi, I can look into it again, but I wasn't able to use that commit, how do I check it out? On GitHub it says `This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.`

nlw0:
The SlotMoveTrack commands don't seem to be on the 2.5 branch yet, should I wait for another branch to be merged firsts?

ronso0:
yes, the Alt moves are available in 2.5 https://github.com/mixxxdj/mixxx/blob/2.5/src/widget/wtracktableview.cpp#L867

I think you need to `git fetch [mixxx upstream]`, you'll see that 2.5 has been updated since your last commit.
Just squash your commits so we have only one essential commit with extra the CI fixup commits, then rebase. Since this is a PR commit that should work flawlessly.

nlw0:
All right I haven''t tested yet but here it is. Hope I didn't mess up anything.

ronso0:
Yep, looks good, thank you!

ronso0:
Could you also take care of documenting the new controls in the manual? : )
https://manual.mixxx.org/2.5/en/chapters/appendix/mixxx_controls.html#the-library-controls

PR #13105, opened 2024-04-15
Delete leftover files in nested dubdirectories

daschuer:
This should fix 
"cannot delete non-empty directory: 2.0/release" 
and similar form here: 
https://github.com/mixxxdj/mixxx/actions/runs/8694490233/job/23843510593

### DIFF
--- a/.github/workflows/download_cleanup.yml
+++ b/.github/workflows/download_cleanup.yml
@@ -25,41 +25,31 @@ jobs:
         if: env.SSH_AUTH_SOCK != null
         run: |
           mkdir empty_folder
-          echo 1.12/ > include_file.txt
-          echo 1.12/* >> include_file.txt
-          echo 1.12/*/* >> include_file.txt
-          echo 2.0/ >> include_file.txt
-          echo 2.0/* >> include_file.txt
-          echo 2.0/*/* >> include_file.txt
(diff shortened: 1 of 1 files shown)


ronso0:
LGTM, we'll see if the list is complete.

.github/workflows/download_cleanup.yml
```diff
@@ -25,41 +25,31 @@ jobs:
...
+          2.0/release/* >> include_file.txt
+          2.0/release/*/* >> include_file.txt
+          2.0/manual >> include_file.txt
+          2.0/manual/* >> include_file.txt
+          2.0/manual/*/pdf >> include_file.txt
+          2.0/manual/*/pdf/* >> include_file.txt
+          2.0/manual/*/html >> include_file.txt
+          2.0/manual/*/html/build >> include_file.txt
+          2.0/manual/*/html/build/html >> include_file.txt
+          2.0/manual/*/html/build/html/* >> include_file.txt
```
ronso0, line 51: Hmm looks like we need to go down one more level.
https://github.com/mixxxdj/mixxx/actions/runs/8703427569/job/23869574192#step:3:47
Can you check if it works with wildcrads only?
http://downloads.mixxx.org/snapshots/1.12/manual/*/*/*/*

PR #13106, opened 2024-04-15
Add analyser support for stem

acolombier:
Make analysers compatible with multi channel and use the best channel for specific features.

As per the NI spec, the first stem is always the drum or rhythmic channel.

This means that:

- Beat and BPM matching will always be using the first stem
- Key will use all other channels

All other analyser perform a stereo mix, except if multichannel is supported internally.

Finally, the waveform analyser will gather EQ samples for the mixed track and per-stem samples.

Depends on #13070


acolombier:
> ## Limitation
> ... the main mix of a stem file is already pre-mastered, the replay gain won't be appropriate ...

An alternative would be not read the main mix pre-mix all the 4 stem together at the SoundSource level. It's pretty trivial, and might be the right way forward, to support other stem format such as the Serato one which only include stem tracks. Should I look into this?

JoergAtGithub:
I think there's no need to support the Serato format, because it's only used for Seratos internal caching. We don't know if they store additional metadata in their library.
Calculating the replay gain from the 4 stems mixed together makes sense to me. Unless we've a compressor/limiter stage, this seems to be the replay gain, that matters.

src/analyzer/analyzersilence.cpp
```diff
@@ -85,7 +87,26 @@ bool AnalyzerSilence::verifyFirstSound(
...
 bool AnalyzerSilence::processSamples(const CSAMPLE* pIn, SINT count) {
-    std::span<const CSAMPLE> samples = mixxx::spanutil::spanFromPtrLen(pIn, count);
+    SINT numFrames = count / m_channelCount;
+
+    const CSAMPLE* pSilenceInput = pIn;
+    CSAMPLE* pMixedChannel = nullptr;
+
+    if (m_channelCount > mixxx::kAnalysisChannels) {
+        // If we have multi channel file (a stem file), we mix all the stems
+        // together in a stereo channel
```
JoergAtGithub, line 97: Can't we used the 5th channel of the STEM file with the mixed signal here, instead of doing the mix operation in the analyzer?

src/analyzer/analyzergain.cpp
```diff
@@ -39,18 +41,39 @@ void AnalyzerGain::cleanup() {
...
 
-    SINT numFrames = count / mixxx::kAnalysisChannels;
+    SINT numFrames = count / m_channelCount;
+
+    const CSAMPLE* pGainInput = pIn;
+    CSAMPLE* pMixedChannel = nullptr;
+
+    if (m_channelCount > mixxx::kAnalysisChannels) {
+        // If we have multi channel file (a stem file), we mix all the stems
+        // together in a stereo channel
```
JoergAtGithub, line 51: If we load the track in a STEM deck, this gain should be correct. But what if we load the mixed stereo signal into a sampler deck? Do we use load 5th mixed channel of the STEM file (which might have a different gain), or do we mix the 4 stem channels at load time?

JoergAtGithub (changes requested):

src/analyzer/analyzersilence.cpp
acolombier: No we cannot because the 5th channel is premixed with the compressor and limiter which we don't support. This shouldn't change anything for silence detection, but it will for other analyzers such as gain.

src/analyzer/analyzerkey.cpp
```diff
@@ -155,12 +157,39 @@ bool AnalyzerKey::processSamples(const CSAMPLE* pIn, SINT count) {
...
     }
 
-    return m_pPlugin->processSamples(pIn, count);
+    const CSAMPLE* pKeyInput = pIn;
+    CSAMPLE* pHarmonicMixedChannel = nullptr;
+
+    if (m_channelCount > mixxx::kAnalysisChannels) {
+        // If we have multi channel file (a stem file), we mix all the stems
+        // together except the first one which contains drums or beats by
+        // convention
```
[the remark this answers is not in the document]
acolombier: Ideally I'd like to keep this PR to strictly the stem change, but if you have suggestion to make it compatible for a future work on Dolby support, happy to update!

src/analyzer/analyzergain.cpp
acolombier: we mix the 4 stem channels at load time, as part of the SoundSource. This done automatically depending of the request max channel count when opening the file.

src/waveform/waveform.cpp
```diff
@@ -199,6 +219,17 @@ void Waveform::readByteArray(const QByteArray& data) {
...
+    m_stemCount = waveform.signal_stems_size();
+    for (int s = 0; s < m_stemCount; ++s) {
+        const io::Waveform::Signal& stem = waveform.signal_stems(s);
+        bool stem_valid = stem.units() == io::Waveform::RMS;
+        for (int i = 0; i < dataSize; ++i) {
+            m_data[i].stems[s] = stem_valid && i < stem.value_size()
+                    ? static_cast<unsigned char>(stem.value(i))
+                    : 0;
+        }
+    }
```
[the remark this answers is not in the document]
acolombier: I cannot use `fill` here as this will override the all struct and set other `stemIdx` to `0` while they technically may have valid data.
I've reworked your suggestion with a for loop

daschuer:
Can you rebase that onto main?

src/analyzer/analyzerbeats.cpp
```diff
@@ -199,12 +201,34 @@ bool AnalyzerBeats::processSamples(const CSAMPLE* pIn, SINT count) {
...
     }
 
-    m_currentFrame += count / mixxx::kAnalysisChannels;
+    SINT numFrames = count / m_channelCount;
+    const CSAMPLE* pBeatInput = pIn;
+    CSAMPLE* pDrumChannel = nullptr;
+
+    if (m_channelCount > mixxx::kAnalysisChannels) {
+        // If we have multi channel file (a stem file), we extract the first
+        // stereo channel, as it is the drum stem by convention
```
daschuer, line 210: I am hesitating to us this assumption unconditionally without a fall back. 
How about check for `StemInfo::getLabel() == "Drums"` and use otherwise the Stereo track?

src/analyzer/analyzerwaveform.cpp
```diff
@@ -221,6 +247,15 @@ bool AnalyzerWaveform::processSamples(const CSAMPLE* buffer, SINT count) {
         storeIfGreater(&m_stride.m_filteredData[Left][High], chigh[Left]);
         storeIfGreater(&m_stride.m_filteredData[Right][High], chigh[Right]);
 
+        for (int s = 0; s < stemCount; s++) {
+            CSAMPLE cstem[2] = {
+                    fabs(pIn[i * stemCount + s * mixxx::kAnalysisChannels]),
+                    fabs(pIn[i * stemCount + s * mixxx::kAnalysisChannels +
+                            1])};
+            storeIfGreater(&m_stride.m_stemData[Left][s], cstem[Left]);
+            storeIfGreater(&m_stride.m_stemData[Right][s], cstem[Right]);
```
daschuer, line 256: This should become a fee function in the anonymous namespace. The whole loop should be copied to sample.cpp near:  `SampleUtil::sumAbsPerChannel` with a vectorization comment.

src/analyzer/analyzerwaveform.h
```diff
@@ -63,10 +57,9 @@ struct WaveformStride {
         for (int i = 0; i < ChannelCount; ++i) {
             m_overallData[i] = 0.0f;
             m_averageOverallData[i] = 0.0f;
-            for (int f = 0; f < FilterCount; ++f) {
-                m_filteredData[i][f] = 0.0f;
-                m_averageFilteredData[i][f] = 0.0f;
-            }
+            std::fill_n(&m_filteredData[i][0], FilterCount, 0.0f);
+            std::fill_n(&m_averageFilteredData[i][0], FilterCount, 0.0f);
+            std::fill_n(&m_stemData[i][0], m_stemCount, 0.0f);
```
daschuer, line 62: Please use SampleUtil::clear()

src/analyzer/analyzerwaveform.h
```diff
@@ -81,15 +74,23 @@ struct WaveformStride {
                     m_postScaleConversion * scaleSignal(m_filteredData[i][Mid], Mid) + 0.5));
             datum.filtered.high = static_cast<unsigned char>(math_min(255.0,
                     m_postScaleConversion * scaleSignal(m_filteredData[i][High], High) + 0.5));
+            for (int stemIdx = 0; stemIdx < m_stemCount; stemIdx++) {
+                datum.stems[stemIdx] = static_cast<unsigned char>(math_min(255.0,
+                        m_postScaleConversion * scaleSignal(m_stemData[i][stemIdx]) + 0.5));
```
daschuer, line 79: scaleSignal() does nothing here and is not prepared allow it later. Remove?

daschuer (changes requested):
I have left some comments:

src/analyzer/analyzerbeats.cpp
JoergAtGithub: AFAIK this assumption is true for all NI STEMs files, and will be most likely true for DEMUCs based real time stem separation as in stemgen. But for other stem and multichannel formats it might be different.
The label is just a human readable information here, wich is not suitable for compare here, because any string is allowed here.

acolombier: Yeah we cannot compare the strings as it may have an arbitrary value. Reading the spec, the first stem must always be the rhythmic stem, but I guess I could add a setting in the preferences to allow using all stem instead, if anyone was using some non compliant stem file, wdyt?

src/analyzer/analyzerwaveform.h
acolombier: I must say I'm not entirely up to speed of what scaleSignal is supposed to do, only reused the same logic as above. Will remove and we can see if that changes anything 
@ywwg I remember during one of our catch up, you mentioned some legacy hardcoded scaling on the waveform signal samples that should be removed, is that it?

src/analyzer/analyzerbeats.cpp
daschuer: If we are convinced that this assumption is always true. The code is just fine. 
A preference option will not really help if there is a single track breaking this assumption.

Maybe a Todi summarizing our discussion is just fine.

src/analyzer/analyzerwaveform.h
daschuer: It currently just scales the bass stream. We can't change it, because the scale factor goes to the stored waveform data.

JoergAtGithub: Than scaleSignal is only relevant for the compatibility to existing waveform cache files, but for stem waveforms we can do it right from the begining and don't apply any scaling in the analyzer.

src/analyzer/analyzerbeats.cpp
daschuer: I did a  quick search at GitHub and found that almost all implementations call the first stem "Drums" following the NI speck closely. 

But I have also other implementations 

"Bass":
https://github.com/stemrollerapp/stemroller/blob/9a397d21e66d28b3adfc5f13b31598ac3a7ff171/main-src/processQueue.cjs#L265

"Melodia":
https://github.com/underson14/criador-de-stems/blob/master/StemMetadata.json

"Vocals"
https://github.com/jim79/stemgen/blob/dfd451910caf87084cf907351202b754d7c59ff0/4stems-22kHz.json#L6

Not sure if that all affects the NI Stem format. But in these cases it is better to use the full down-mix. 

Since the full down Mixxx already produces reasonable results. We should use it to a to avoid blaming our beat or key detector, just because a user has creative ideas how to sort the stem files.  

What do you think? 


By the way, I have found this testing dataset in NI stem format:
https://sigsep.github.io/datasets/musdb.html

JoergAtGithub: This is a label, a text which is only there to be printed in the UI. It is not intended to be machine readable! Each STEM file uses different texts here - and this is intended by the specification.
The machine readable information is the number of the stem channel. According to the specification, the 1st must be the stem with the rythm, and the 4th the vocal stem (if the track has vocals otherwise it should be used for a melodic stem).

We shouldn't try to make assumptions how somebody labels his stem channels. Instead we should strictly follow the specification. And this means we pass the stem labels to the UI only and use the stem channel number for the analyzer.

src/analyzer/analyzerwaveform.h
ywwg: I think we should take the opportunity to rip this out, once and for all.  If we are bumping the analysis version number then we can remove this for all analysis and fix the waveform widgets appropriately

ywwg: (the original reason I wrote the scaling was because it was hard to see the bass signal in the waveforms)

acolombier: Great! Yes, I also think it should be ripped off because of the bump off. I will put it to my list and get it done in a follow up PR tho just to separate it from the stem, since I assume it will also need some logic change in the renderer, where there is is that `x2` logic.

daschuer: For this PR, please just remove the scaleSignal() call from the new code. 

I am a bit in doubt whether the version bump is worth the effort to get rid of this. We need also consider that this makes it harder to go back an forth between release and beta. We just need to make sure to have it on the to-do lost if we need to make a version bump anyway.

src/analyzer/analyzersilence.cpp
JoergAtGithub: Isn't mixMultichannelToStereo a time consuming step?

acolombier: Reading the silence analyser again, it looks like it should support multi channel indeed.

src/analyzer/analyzerkey.cpp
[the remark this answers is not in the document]
acolombier: For now, I have added a comment.
Alternatively, I could extend the `Analyser` abstract class to store whether the file is a NI Stem file (eg. add bool to `initialise`) but at the moment, this feels redundant. WDYT?

src/analyzer/analyzerwaveform.h
acolombier: > I am a bit in doubt whether the version bump is worth the effort to get rid of this.

We need to as currently, the render applies a gain on sample due to this, so if we fix this, old waveform will look unusual due to missing gain. 
Since the stem feature will require a version bump, this is why I suggested making the signal clean up as well, so we can do a "2 birds one stone".

acolombier: > I think we should take the opportunity to rip this out, once and for all. If we are bumping the analysis version number then we can remove this for all analysis and fix the waveform widgets appropriately

Going to try an prepare a PR today

src/analyzer/analyzerwaveform.cpp
acolombier: Are you talking about the for-loop at line `222` of the one at `250`?

src/analyzer/analyzerkey.cpp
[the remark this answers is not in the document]
JoergAtGithub: A boolean if the source is a NI STEM file would be an option, but maybe a struct that contains parameters (like isRythmChannel) for each channel would be more future proof here.
But that's maybe overcomplicating this. Let's add a ToDo comment at all these channelCount conditions for now.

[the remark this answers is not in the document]
acolombier: Are you happy with the comment I already added at line 174? Any suggestion otherwise?

src/analyzer/analyzerwaveform.h
acolombier: #13416

src/analyzer/analyzerwaveform.h
```diff
@@ -131,13 +132,15 @@ struct WaveformStride {
...
     int m_position;
+    int m_stemCount;
     double m_length;
     double m_averageLength;
     int m_averagePosition;
     int m_averageDivisor;
 
     float m_overallData[ChannelCount];
     float m_filteredData[ChannelCount][FilterCount];
+    float m_stemData[ChannelCount][4];
```
JoergAtGithub, line 143: `ChannelCount` and `4` are not self explaining here. Can you add a constexpr kStemCount = 4 in waveform.h

src/util/sample.cpp
```diff
@@ -767,6 +767,32 @@ void SampleUtil::mixMultichannelToMono(CSAMPLE* pDest, const CSAMPLE* pSrc, SINT
...
+                continue;
+            }
+            pDest[mixxx::audio::ChannelCount::stereo() * i] +=
+                    pSrc[numChannels * i +
+                            stemIdx * mixxx::audio::ChannelCount::stereo()];
+            pDest[mixxx::audio::ChannelCount::stereo() * i + 1] +=
+                    pSrc[numChannels * i +
+                            stemIdx * mixxx::audio::ChannelCount::stereo() + 1];
+        }
+    }
```
JoergAtGithub, line 793:
```suggestion
 for (int stemIdx = 0; stemIdx < stereoChCount; ++stemIdx) {
     if ((excludeChannelMask >> stemIdx) & 0b1) {
         continue;
     }
     for (int i = 0; i < numFrames; ++i) {
         const int srcIdx = i * numChannels + stemIdx * mixxx::audio::ChannelCount::stereo();
         const int destIdx = i * mixxx::audio::ChannelCount::stereo();
         pDest[destIdx] += pSrc[srcIdx];
         pDest[destIdx + 1] += pSrc[srcIdx + 1];
     }
 }
Lets put the simple loop without condition inside.
```

src/analyzer/analyzerkey.cpp
```diff
@@ -155,12 +157,42 @@ bool AnalyzerKey::processSamples(const CSAMPLE* pIn, SINT count) {
...
+    const CSAMPLE* pKeyInput = pIn;
+    CSAMPLE* pHarmonicMixedChannel = nullptr;
+
+    if (m_channelCount > mixxx::audio::ChannelCount::stereo()) {
+        // If we have multi channel file (a stem file), we mix all the stems
+        // together except the first one which contains drums or beats by
+        // convention
+        // Note that if Mixxx adds support for other multi channel sound sources
+        // (e.g 5:1 dolby), this condition will have to be reworked to assert if
+        // the sound source is a NI stem
```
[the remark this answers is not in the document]
acolombier: > This would also fix also the 6ch Dolby 5:1 case

It doesn't since some of those analysers relies on stereo signals. 
Since Dolby 5:1 handling seems to have become a requirement for that feature, I have added an assert and an early return.

src/analyzer/analyzerwaveform.h
JoergAtGithub: Please move the constexpr to waveform.h, where it can also be used: https://github.com/mixxxdj/mixxx/blob/a0ab0c3216e0c1353ff87aca2ac27776fd9ca018/src/waveform/waveform.h#L24

src/analyzer/analyzerbeats.cpp
daschuer: > This is a label, a text which is only there to be printed in the UI.

I can unfortunately not confirm this from the standard. _whenever possible_ a label from the given list shall be used. Since this is not a translated, it does not well suite for the GUI. I think we should make them translatable, this way, additional user Stem labels do not look misplaced. 

And than we have the Arrangement paragraph which is also a weak one compared to the bold must and must not in other parts. _file layout should be consistent_

I assume we will face these tracks:
1. Correct Layout and correct labels -> Use "Drums" 
2. Incorrect Layout and Labels -> Don't use fist for Drums
3. Correct Layout and custom labels -> Don't use fist for Drums
4. Wrong Layout and correct labels -> Use "Drums"

Is a pity withb 3., but we cannot distinguish it from 2. So not using Drums is a save route. 

We can clearly blame the track 4. As broken. User has to fix it and not lable a track "Drums" that don't contain them.


> We shouldn't try to make assumptions how somebody labels his stem channels. 

I agree. But in the same way we should not make assumptions how the track is arranged. Only if the names are correct and we find "Drums" we should assume that it is actually Drums. 

Please also keep in mind that Mixxx is able to analyze tracks with s full stem mix. It has hard times to analyze a bpm of a track where kick drums have been removed. So we must not apply the trick for a better detection when when are not certain that it will help.

acolombier: > we should not make assumptions how the track is arranged

We don't - the spec does.

 > Only if the names are correct and we find "Drums" we should assume that it is actually Drums

Here is the [list of recommended label](https://github.com/acolombier/stemgen?tab=readme-ov-file#note-on-stem-customisation) as per the NI spec - It's not just "drums" that can be rhythmic stem. Also, what about internationalisation? Would "Tambours" or "Schlagzeug" valid as well? How would we handle a Stem metadata with German labels but is loaded on a French interface?  Where would we draw the line?

It really sounds like we are going toward a dead end here. As I said above, I'd be happy to put a user setting to allow ignoring the stem analyse optimisation, so if we ever have a user not respecting the spec, they can still analyse the audio like it would for a traditional stereo for any stem files they load.

JoergAtGithub: > I'd be happy to put a user setting to allow ignoring the stem analyse optimisation

I fully support this, using labels for anything ese the GUI output will fail!
And at the same time the DJ media praise the competition for the huge step forward due to AI/STEM base beatgridding. This feature could be the foundation for this.

daschuer: Let's not block this PR over this discussion. 
A logical step for me would be to merge this without the optimization and use always the full down mix. The sups for the optimization can remain. What do you think?

daschuer: > Would "Tambours" or "Schlagzeug" valid as well?

I would argue that translated texts doe not follow the recommendation from the speck. English language is not explicit mentioned, but when you use on one hand one of the preferred terms and than add a German term for a term that is missing in the list it will look misplaced. Using "Tambours" or "Schlagzeug" obviously not allowed because "Drums" the equivalent on the list of recommended terms.  

IMHO the remaining question is how to distinguish case 2. from 3. (from https://github.com/mixxxdj/mixxx/pull/13106#discussion_r1665645429) 
Do we have a real live example for 3.? I have not yet found some. 

>  using labels for anything else the GUI output will fail!

@JoergAtGithub I agree to this statement stand alone. But we have also the evidence that relying on the arrangement also fails which is IMHO worse. 

The degree of failure is different: 
* When the arrangement is wrong with optimization applied the whole analysis fails. This happens for all 2. tracks. 
* When we do not apply analysis because of custom labels the analyse still works, but might be by a chance less accurate. This happens for 3. tracks. 

My conclusion is that we should only apply optimization if we have a sanity check in place to sort out case 2. 

As a workaround we may inform users whether optimizations has been auto applied and allow to force a method using the track context menu. As a foundation we need to store the method in the analysis metadata. 

What do you think?

JoergAtGithub: We can not check if the arrangement is correct, but we could say, if the first channel is labled different than expected, that we fall back to mixxed stems.
We currently have already different analyzer options exposed to the user for fixed and variable beatgrid.
Regarding the case that there is a section, where the drum is not playing, I wonder if should only use this stem, or if we maybe should mix the stems with different gains. We could amplify the 1st stem with gain 2 and the 4th with gain 0.5 for example.

acolombier: > the first channel is labled different than expected, that we fall back to mixxed stems.

Do you mean if the label doesn't contain [a recommended label](https://github.com/acolombier/stemgen?tab=readme-ov-file#note-on-stem-customisation)? I guess it could be a valid trade-off, although I would reiterate on having that configurable with a user setting (something in the Analyser like `Optimise analyse for stem track` and 3 options like `Enforced` (always on, like currently), `Automatic` (if the stem is labelled with recommended one - not just `Drums` is other label can be used to indicate rhythmic channel) and `Disabled` (only used mixed stems)

> I wonder if should only use this stem, or if we maybe should mix the stems with different gains

Indeed, there could be some macros optimisation to do here, but this will require further benchmarking to find the ideal parameters. I think the current state of this PR doesn't prevent adding this in a future iteration, and I would rather do it like this as I cannot guarantee I will have the time to do that benchmarking and correct tweaking.

src/analyzer/analyzerwaveform.cpp
acolombier: @daschuer friendly ping in case you missed my reply here

src/analyzer/analyzerbeats.cpp
daschuer: > Do you mean if the label doesn't contain [a recommended label](https://github.com/acolombier/stemgen?tab=readme-ov-file#note-on-stem-customisation)? 

Yes. 


>  that configurable with a user setting (something in the Analyser like Optimise analyse for stem track and 3 options like Enforced (always on, like currently), Automatic (if the stem is labelled with recommended one - not just Drums is other label can be used to indicate rhythmic channel) and Disabled (only used mixed stems)

That works for me. Do you agree to have "Disabled" in this PR and do the configuration in a follow up? 

The proposed options are IMHO however to difficult to use. How about using only Automatic but keep an extensible list of labels that will enable the optimization? 

In addition we should always offer a reanalyze context menu, because there are likely some tracks that follow the stem specs but have worse results with optimization enabled.

src/analyzer/analyzerwaveform.cpp
daschuer: I original ment 250. However I have no strong opinion here. There are two overlapping principals, my comment is sourced from:
* Private functions not using members shall be moved to the anonymous namespace 
* All sample loops where vectorization matters shall go sample.cpp

acolombier: I'm afraid there may some confusion here:
- That for-loop iterates over the 4 stems, there is nothing to do about sample directly, and vectorization doesn't matter.
- `storeIfGreater` is a very specific to analyzer and IMHO it makes no sense to move it the sample helper function as there is no usecase to use it anywhere currently.
Happy to remove the for-loop and duplicate the the logic for each channel, like it is done for each frequency space above similar to this
```
            CSAMPLE cstem[2] = {
                    fabs(pIn[i * stemCount]),
                    fabs(pIn[i * stemCount + 1])};
            storeIfGreater(&m_stride.m_stemData[Left][0], cstem[Left]);
            storeIfGreater(&m_stride.m_stemData[Right][0], cstem[Right]);
            cstem[2] = {
                    fabs(pIn[i * stemCount + mixxx::kAnalysisChannels]),
                    fabs(pIn[i * stemCount + mixxx::kAnalysisChannels + 1])};
            storeIfGreater(&m_stride.m_stemData[Left][1], cstem[Left]);
            storeIfGreater(&m_stride.m_stemData[Right][1], cstem[Right]);
            cstem[2] = {
                    fabs(pIn[i * stemCount + 2 * mixxx::kAnalysisChannels]),
                    fabs(pIn[i * stemCount + 2 * mixxx::kAnalysisChannels +
                            1])};
            storeIfGreater(&m_stride.m_stemData[Left][2], cstem[Left]);
            storeIfGreater(&m_stride.m_stemData[Right][2], cstem[Right]);
            cstem[2] = {
                    fabs(pIn[i * stemCount + 3 * mixxx::kAnalysisChannels]),
                    fabs(pIn[i * stemCount + 3 * mixxx::kAnalysisChannels +
                            1])};
            storeIfGreater(&m_stride.m_stemData[Left][3], cstem[Left]);
            storeIfGreater(&m_stride.m_stemData[Right][3], cstem[Right]);
```

src/analyzer/analyzerbeats.cpp
acolombier: > Do you agree to have "Disabled" in this PR and do the configuration in a follow up?

I have added a basic configuration, which will be `Disabled` by default, and support `Enforced`

> The proposed options are IMHO however to difficult to use. How about using only Automatic but keep an extensible list of labels that will enable the optimization?

That sounds like a potential plan for the future. I have left `Automatic` unimplemented for now so we can revisit and see if there is a value in implementing such a feature 

> In addition we should always offer a reanalyze context menu, because there are likely some tracks that follow the stem specs but have worse results with optimization enabled.

This can be implemented in the future, if we really see a need for it. Currently, one can always:
- Change the analyser strategy in setting
- Re analyse

Agreed this is 6 clicks vs 2 clicks, but I'm not sure the ROI is worth focusing on.

src/analyzer/analyzerwaveform.cpp
daschuer: Ok, than let's keep it like it is and move on.

daschuer (commented):
The first eatDetectionSettings::StemStrategy settimg alters the analyzer results and shall therefore also be part of the analyzer metadata so that we can later deal with existing analysis of alpha testers. To not put extra work on you for this intermediate solution, it would be also OK to disable or hide the feature to now.

How are existing analysis treated by the way? They have been probably done from the stereo downmix. Do we need to deal with it?

acolombier:
> The first eatDetectionSettings::StemStrategy settimg alters the analyzer results and shall therefore also be part of the analyzer metadata

Is that already supported for the others beats and key settings such as the algorithm used for either? If yes, happy to add this as part of that PR too as well.

daschuer:
For beats, it is implemented here: 
https://github.com/mixxxdj/mixxx/blob/5ced2dee3c6192754b62c21cfad0388b751a89ee/src/analyzer/analyzerbeats.cpp#L181

I am not sure if we have similar for keys.

acolombier:
> For beats, it is implemented here:
> 
> https://github.com/mixxxdj/mixxx/blob/5ced2dee3c6192754b62c21cfad0388b751a89ee/src/analyzer/analyzerbeats.cpp#L181
> 
> I am not sure if we have similar for keys.

AFAICS, this is only taken protobuf serialisation version into account, this doesn't invalidate older analyses. Or am I missing something?

On a side note, I'm just wondering, do we want to invalidate older analyses upon analyser changes? I'm thinking it actually sounds wrong since you may want to change your settings temporarily (e.g with our use case, to analyse a non compliant stem) without invalidate other valid analyses. In case you wanted to redo the analyses you always get the option to do so in the context menu. Wdyt?

daschuer:
We already have a preference option for this: 
https://github.com/mixxxdj/mixxx/blob/d9d23dc238f9cfb8a9e997615fcf576d7971851a/src/preferences/dialog/dlgprefbeatsdlg.ui#L78

But anyway, can we postpone all this and merge this with the unoptimized Analysis? 

In case of beats we have the const and the non const detection and version numbers. We must not automatically reanalyze a track, because beat markers may shift away from the cue points, which may creates surprising effects when mixing. 

I think we should optimize this workflow: 
* Use reasonable defaults in the preferences 
* Track settings in the analysis metadata to be able to destingush/filter tracks using a certain setting.
* Allow to reanalyze single tracks with a different setting to fix special issues with a couple of tracks. 
* Allow to reanalyze all tracks in case of improved  analyzers on demand.

I think we should not recommend to alter the preference options to fix single tracks.

acolombier:
> can we postpone all this and merge this with the unoptimized Analysis?

Yeah that's fine - I'll hide the option in the setting, so I can still use it via config file, if that's okay with you.

> Use reasonable defaults in the preferences

Do you think that `Disabled` is not the best option here?

> I think we should not recommend to alter the preference options to fix single tracks

Agreed - but this is currently how you would deal with tracks needing different options (e.g key plugin). It is suboptimal, but reworking the whole flow is an entire feature so I guess that workaround remain the only way, and thus we should still take it into account.

daschuer:
> Yeah that's fine - I'll hide the option in the setting, so I can still use it via config file, if that's okay with you.

Yes that's fine. The preferences pane needs anyway some love. 

> Do you think that Disabled is not the best option here?

For the vast majority of the tracks the optimized analysis will be the best option. Unfortunately we need to deal with the exceptions. 
Without that in place let's keep it disabled as a safe default known good form stereo tracks. 

> It is suboptimal, but reworking the whole flow is an entire feature. 

Yes. Since it should work without optimization, postponing is a good idea. 

I see the following solution/iterations:
1. Add a stem mode to the analysis metadata to distinguish it later. (We need it anyway for further improvements of the analyzer) 
2. Only optimize when a the first stem is reasonable labeled.
3. Add a context menu option to the analyze menu to reanalyze problematic tracks differently.  
4. Allow to add custom Labels that enable optimizations 

Does that work for you? (I don't want to put the work on you, just plan possible steps)

src/analyzer/analyzerbeats.cpp
acolombier: Captured as part of #13466

src/util/sample.h
```diff
@@ -311,6 +311,19 @@ class SampleUtil {
...
+    // channel buffer input is composed of stereo pair. Note that function
+    // cannot be optimised using loop vectorization and so shouldn't be used for
+    // real-time use case. The exclude channel mask can bne used to exclude a
+    // stereo pair (two consecutive channel) out of the mix. The LSB is the
+    // first stereo channel
+    static void mixMultichannelToStereo(CSAMPLE* pDest,
+            const CSAMPLE* pSrc,
+            SINT numFrames,
+            mixxx::audio::ChannelCount numChannels,
+            int excludeChannelMask = 0);
```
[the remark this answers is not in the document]
acolombier: I have created a full downmix overload and added the vectorization comment - could you please confirm they are correct?

daschuer (approved):
Thank you. LGTM. Please give a hint when I can press merge

daschuer:
@JoergAtGithub is there something open form your review?

acolombier:
All good from my end @daschuer

JoergAtGithub:
Yes all is fine now! Thank you very much!

JoergAtGithub (approved):

PR #13109, opened 2024-04-16
feat: Pioneer DDJ-REV1 mappings 

LauraRozier:
Controller mappings for the Pioneer DDJ-REV1

Skipped features (don't exist within Mixxx):
- TrackingScratch Mode Pads
- Trans Mode Pads
- ScratchBank Mode Pads

Still a draft while I test things and work out some quirks / docs


JoergAtGithub:
Welcome at Mixxx!

JoergAtGithub:
grafik
unzip it, and apply it using:
`git apply pre-commit.patch`

LauraRozier:
Yep, that's why it's in DRAFT mode. ^_^ Work in progress, so builds will fail.
Did the 'ol contributor agreement thing just now as well.

As for documenting the whole shebang for the manual.. That might take some time.. ^_^"
I'm fast with things I enjoy working on, but hella slow with docs.

Currently I'm experiencing some troubles getting the FX buttons to light up properly, but this looks like an issue with the controller itself, rather than Mixxx's mapping. When pressing the FX button it lights up, then immediately goes off again, even though the effect stays on in Mixxx, from that point forwards the button acts inverted. It's weird..
I'm talking about buttons 3,4,5in the image below
image

Issue #13114, opened 2024-04-16
"MIDI for light" mapping tries to kill a non-existent timer


fwcd:
### Bug Description

I've had some issues running this mapping lately. I've been getting errors of the form

```
Uncaught exception: file:///Applications/Mixxx.app/Contents/Resources/controllers/Midi_for_light-scripts.js:105: TypeError: Cannot read property '0' of undefined
Backtrace: @file:///Applications/Mixxx.app/Contents/Resources/controllers/Midi_for_light-scripts.js:105
```

and a bunch of log messages

```
warning [Controller] "Tried to kill Timer \"0\" that does not exists"
warning [Controller] "Tried to kill Timer \"0\" that does not exists"
warning [Controller] "Tried to kill Timer \"0\" that does not exists"
warning [Controller] "Tried to kill Timer \"0\" that does not exists"
```

The MIDI device I am trying to run it with is a virtual loopback port ([midilink](github.com/fwcd/midilink)). Not sure what went wrong here, since I have already been successfully running this configuration on another machine, will have to dig into this...

### Version

2.5-alpha-329-g59a4374a44 (HEAD)

(though this also seems to happen in 2.4 builds)

### OS

macOS 14.3.1

fwcd:
I have noticed that the MIDI for light mapping only seems to work in isolation, but not if another physical controller (e.g. the MC7000) has an enabled mapping. The MC7000 works fine, but the "MIDI for light" MIDI port doesn't receive any messages on the other end...

ekordas:
Can confirm, script crashes on Mac and Win10 in 2.4.1

Swiftb0y:
The crash is weird. Can you test #13117 and see if it that fixes it for you? If not, could you [create a backtrace](https://github.com/mixxxdj/mixxx/wiki/Creating%20Backtraces) of the crash?

ekordas:
Just inserted Midi_for_light-scripts.js from https://github.com/mixxxdj/mixxx/pull/13117 into a Mixx 2.4.1 Install on Windows and it seems to work.

ekordas:
Same error as written above in the initial decription. (only the script crashes not the Program)

Swiftb0y:
Does that error occur with #13117 or without? Sorry for asking, its just not entirely clear to me.

ekordas:
[...]

PR #13117, opened 2024-04-16
MIDI for light: Fix unsound timer handling

fwcd:
This is a first attempt at fixing #13114.

Maybe someone more familiar with the legacy controller scripting system can comment on this, the timer handling in the script did not look sound, so I tried to ensure that timers are only stopped if they are running and that they are unassigned properly if stopped.

### DIFF
--- a/res/controllers/Midi_for_light-scripts.js
+++ b/res/controllers/Midi_for_light-scripts.js
@@ -102,10 +102,11 @@ midi_for_light.init = function(id) { // called when the MIDI device is opened & 
 };
 
 midi_for_light.shutdown = function(id) { // called when the MIDI device is closed
-    engine.stopTimer(deck_beat_watchdog_timer[0]);
-    engine.stopTimer(deck_beat_watchdog_timer[1]);
-    engine.stopTimer(deck_beat_watchdog_timer[2]);
-    engine.stopTimer(deck_beat_watchdog_timer[3]);
+    for (let i = 0; i <= 3; i++) {
+        if (deck_beat_watchdog_timer[i]) {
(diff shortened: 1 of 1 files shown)


res/controllers/Midi_for_light-scripts.js
```diff
@@ -102,26 +102,30 @@ midi_for_light.init = function(id) { // called when the MIDI device is opened &
 };
 
 midi_for_light.shutdown = function(id) { // called when the MIDI device is closed
-    engine.stopTimer(midi_for_light.deck_beat_watchdog_timer[0]);
-    engine.stopTimer(midi_for_light.deck_beat_watchdog_timer[1]);
-    engine.stopTimer(midi_for_light.deck_beat_watchdog_timer[2]);
-    engine.stopTimer(midi_for_light.deck_beat_watchdog_timer[3]);
```
fwcd, line 108: Note that this likely didn't work in the first place. `deck_beat_watchdog_timer` is a free variable, not a member of `midi_for_light`. It was probably never noticed being part of the rarely invoked shutdown handler.

fwcd: The error can be reproduced by selecting the MIDI for light mapping and then switching to another mapping (which errors)

Swiftb0y (approved):
The entire script is a mess, so LGTM I guess :man_shrugging:

fwcd:
Should this be merged then?

Swiftb0y:
Yes

fwcd:
Ah this still targeted 2.4, so we might another merge to 2.5

Issue #13111, opened 2024-04-16
DEBUG_ASSERT when deleting track file that's not in the watched library directories


ronso0:
### Bug Description

I did 
* take a track file that's not in the watched library directories
* drag the file into a playlist
* load it from the playlist into a deck
* open the deck's track menu -> Move track file to Trash, confirm

(via track menu only because Move to Trash is not enabaled in Playlists)
(if enabled there behaviour is the same)

```
DEBUG ASSERT: "!pTrack" in function void BaseTrackCache::replaceRecentTrack(TrackId, TrackPointer) const at ./src/library/basetrackcache.cpp:180
```

Works flawlessly if I move this new file into one of the watched directories and rescan.
Also works when removing track file in Crate view.

Noticed with main / 2.5-alpha

Reminds me of #11909

ronso0:
'regression' because I couldn't reproduce with 2.4
Maybe that assertion was added in 2.5, and library side effects not noticed in 2.4?

PR #13120, opened 2024-04-17
fix typo in WTrackProperty

ronso0:
**edit** after merging 2.4 with the other bugfix that uses the same mechanics, this is now just a typo fix.
_____

one more fix for good track menu / tracks view UX.

* load a track from current view
* deck track menu (via CO / controler) -> move track file to trash -> Yes
* (deck is stopped, track ejected)
* track view is repopulated (and eventually resorted)
* **currently:** current index is reset, next move jumps to top left index
* **this PR:** prev. selection and current index is restored (selection off by one track)

### Note:
~~When testing with a playlist view you'll hit #13111.~~ (can now reproduce only in `main`/2.5-alpha :face_with_spiral_eyes: )


src/widget/wtrackproperty.cpp
```diff
@@ -121,4 +121,15 @@ void WTrackProperty::ensureTrackMenuIsCreated() {
         m_pTrackMenu = make_parented<WTrackMenu>(
                 this, m_pConfig, m_pLibrary, WTrackMenu::kDeckTrackMenuFeatures);
     }
+    // Before and after the loaded tracks file has been removed from disk,
+    // instruct the library to save and restore the current inde for
```
m0dB, line 125: typo, index

m0dB:
LGTM

Note that this PR targets 2.4. Shouldn't that be main?

ronso0:
2.4 is correct.
I mentioned deck menu via CO, but I noticed that's irrelevant. The the view is also not updated when a track is purged via the deck menu.

daschuer (commented):
Now this PR is melted down to be only a typo fix. 
Is that correct? Can you adjust the title then?

daschuer (approved):
Thank you LGTM

PR #13123, opened 2024-04-17
Add quick effect support on stem

acolombier:
https://github.com/mixxxdj/mixxx/assets/7086688/1c51ca82-d110-443b-8232-f1ecf0b356cc

Depends on #13086


JoergAtGithub:
Two minor remarks regarding the floating widgets:

- It's sometimes difficult to distinguish the colored widget from the waveform which has the same colors. A border in a different color might help.
grafik
- The effect toggle button is very small an difficult to hit with the mouse

acolombier:
Yeah, that widget definitely would need some refactoring, but I think the deal was to remove it before we merge that feature, and is only here to facilitate testing, thus I didn't want to spend too much time with it.

JoergAtGithub:
I suggested to move the floating widgets in an own PR, to not block merging this PR and the PRs before.
I'm not against these widgets, I just expect a longer review discussion than for the other code.

acolombier:
Ah I had misunderstood the plan for that. Sounds good then, I try and see if I can use the effect state indicator used in the effect bar instead of the mixer one.

acolombier:
Please note that UI changes (1st and 3rd commits) will be moved to https://github.com/mixxxdj/mixxx/pull/13515 once this PR is ready to be merged!

res/qml/StemKnob.qml
```diff
@@ -6,62 +6,139 @@ import "Theme"
...
     required property string group
     required property string label
     required property color stemColor
     required property int index
 
-    width: 56
-    height: 56
-    radius: 5
-    color: stemColor
+    readonly property string fxGroup: `[QuickEffectRack1_${group.substr(0, group.length-1)}Stem${index + 1}]]`
```
acolombier, line 16: Please note that `qml_formatter` is currently broken. 
I have made [a fix](https://github.com/qarmin/qml_formatter/pull/6) for it but the repo hasn't been looked at for 2 years, so not sure this is still being maintained. Happy to setup my instead, or maybe we could even adopt that project in the MixxxDJ organisation?

acolombier: Ignore me, this was merged already! Will update and fix formatting later.

JoergAtGithub: Will you create a PR to update our CI?

src/effects/effectsmanager.cpp
```diff
@@ -307,8 +354,18 @@ void EffectsManager::saveEffectsXml() {
         quickEffectChainPresets.insert(qeIt.key(), pPreset);
     }
 
+    QHash<QString, EffectChainPresetPointer> quickStemEffectChainPresets;
+    quickStemEffectChainPresets.reserve(m_quickStemEffectChains.size());
+    QHashIterator<QString, QuickEffectChainPointer> sqeIt(m_quickStemEffectChains);
```
JoergAtGithub, line 359:
```suggestion
    QHashIterator<QString, QuickEffectChainPointer> qseIt(m_quickStemEffectChains);
```
I guess the meaning should be QuickStemEffectIterator?

src/effects/presets/effectchainpresetmanager.cpp
```diff
@@ -857,8 +858,30 @@ EffectsXmlData EffectChainPresetManager::readEffectsXml(
...
-    return EffectsXmlData{
-            eqEffectManifests, quickEffectPresets, standardEffectChainPresets, mainEqPreset};
+    // Read names of presets that were loaded into stem QuickEffects on last shutdown
+    QDomElement quickStemEffectPresetsElement =
+            XmlParse::selectElement(root, EffectXml::kQuickStemEffectList);
+    const QDomNodeList quickStemEffectNodeList =
+            quickStemEffectPresetsElement.elementsByTagName(
+                    EffectXml::kChainPresetName);
+    for (int i = 0; i < quickStemEffectNodeList.count(); ++i) {
+        QDomElement presetNameElement = quickStemEffectNodeList.at(i).toElement();
```
JoergAtGithub, line 868:
```suggestion
    quickStemEffectPresets.reserve(quickStemEffectNodeList.count());
    for (const auto& qseNode : quickStemEffectNodeList) {
        QDomElement presetNameElement = qseNode.toElement();
```

src/effects/presets/effectchainpresetmanager.cpp
```diff
@@ -917,6 +940,37 @@ EffectXmlDataSingleDeck EffectChainPresetManager::readEffectsXmlSingleDeck(
...
+    auto pQuickEffectChainPreset = getDefaultQuickEffectPreset();
+
+    // Read name of last loaded QuickEffect preset
+    QDomElement quickEffectPresetsElement =
+            XmlParse::selectElement(root, EffectXml::kQuickStemEffectList);
+    const QDomNodeList quickEffectNodeList =
+            quickEffectPresetsElement.elementsByTagName(
+                    EffectXml::kChainPresetName);
+    for (int i = 0; i < quickEffectNodeList.count(); ++i) {
+        QDomElement presetNameElement = quickEffectNodeList.at(i).toElement();
```
JoergAtGithub, line 957: Same here

JoergAtGithub (changes requested):

src/mixer/playermanager.h
```diff
@@ -154,6 +154,13 @@ class PlayerManager : public QObject, public PlayerManagerInterface {
         return QStringLiteral("[Channel") + QString::number(i + 1) + ']';
     }
 
+    // Returns the group for the ith deck and jth stem where i and j is zero indexed
```
acolombier, line 157: `#ifdef __STEM__`

src/engine/channels/enginedeck.cpp
```diff
@@ -118,31 +120,66 @@ EngineDeck::~EngineDeck() {
...
-            if (m_stemMute[stemIdx]->toBool()) {
-                continue;
-            }
-            float gain = static_cast<float>(m_stemGain[stemIdx]->get());
-            pOut[i] += pIn[stemCount * i + mixxx::kEngineChannelOutputCount * stemIdx] * gain;
-            pOut[i + 1] +=
-                    pIn[stemCount * i +
-                            mixxx::kEngineChannelOutputCount * stemIdx + 1] *
-                    gain;
+            int chOffset = stemIdx * mixxx::audio::ChannelCount::stereo();
```
acolombier, line 153: This is now the final version of `EngineDeck::processStem`!

res/qml/StemKnob.qml
acolombier: Created #13530 with the fix + some other updates for pre-commit

Eve00000:
Remark after Mixxxing a lot (and having lots of fun)
As I feel (and the way I DJ) it would prefer to have the fx standard disabled when a track is loaded, when I switch a lot loading tracks, I forget an effect is enabled and pushing/turning one stemcontrol activates all the others.
So analogous to the [ChannelXStemY],mute (that is disabled by default) ([QuickEffectRackX_[ChannelXStemY]],enabled) also on 0.
Or is there a certain idea to put these on enabled buh default?

src/engine/channels/enginedeck.cpp
daschuer: I am struggling to undertand the code here with all the diffrent Layouts and in and out buffers. 
An output to one function is an input to another. 
Can you go through it and try to rename the variables more verbose?
You may also codider alias pointers that have the correct name for the bufer with the current content. 
Also // LRL .. viualization would be helpfull.

src/util/sample.h
```diff
@@ -370,6 +370,17 @@ class SampleUtil {
             mixxx::audio::ChannelCount numChannels,
             int sourceChannel = 0);
 
+    // Copies and strips interleaved stereo sample data in pSrc with
+    // down to multi-channel samples into pDest. Samples will be written at the
+    // channel pointed by channelOffset. Samples from all other channels will be
+    // ignored. pDst must contain (numFrames * numChannels) samples (numFrames *
+    // 2) samples will be written into pDest
+    static void copyStereoToMulti(CSAMPLE* pDest,
```
daschuer, line 378: In adidition an example in the comment will be also helpfull. 

Does the term "insert" fit here? 

The layout is now like: 
1L1R2L2R1L1R2L2R1L1R2L2R1L1R2L2R
right? 

Does 
1L1R1L1R1L1R1L1R2L2R2L2R2L2R2L2R
Have any benefit? 

Do we have better name these layouts so we can use consistenly throut the code base?

daschuer (changes requested):
The code is probably good. But it is not instantly understadable. I have left some comments.

src/util/sample.h
acolombier: > Does the term "insert" fit here?

Sounds much better indeed. I have added an example as well.

> Does
1L1R1L1R1L1R1L1R2L2R2L2R2L2R2L2R
Have any benefit?

AFAIS, I mostly see issues and limitations. The main one being the inability to read frame sequentially and requiring to consume the entire buffer in order to reconstruct a single frame. With the current layout, you do not have to take into consideration the entire buffer size.
The only advantage of using the following layout is to drop the need of `SampleUtil::copyOneStereoFromMulti`, but since this helper is trivial, it only takes couple of ns to extract a single stereo track after vectorization.

src/effects/presets/effectchainpresetmanager.cpp
acolombier: Unfortunately, `QDomNodeList` isn't iterable. ([doc](https://doc.qt.io/qt-6/qdomnodelist.html))

src/effects/presets/effectchainpresetmanager.cpp
acolombier: same problem

src/engine/channels/enginedeck.cpp
acolombier: Let me know if it easier to understand now! Feel free to suggest better phrasing for the comments.

acolombier:
I have removed the UI widget, so this PR should be ready for merge. If you wish to test it with the UI widget, you can use the branch `feat/add-quick-fx-on-stem-with-ui`

acolombier:
> fx standard disabled when a track is loaded

I will look into this. There is already a setting for mute/volume, so it makes sense to extend it to FX IMO.

JoergAtGithub:
> > fx standard disabled when a track is loaded
> 
> I will look into this. There is already a setting for mute/volume, so it makes sense to extend it to FX IMO.

Before you implement this, have a look on the discussion in PR https://github.com/mixxxdj/mixxx/pull/11198

src/effects/presets/effectchainpresetmanager.cpp
```diff
@@ -857,8 +858,31 @@ EffectsXmlData EffectChainPresetManager::readEffectsXml(
...
-    return EffectsXmlData{
-            eqEffectManifests, quickEffectPresets, standardEffectChainPresets, mainEqPreset};
+    // Read names of presets that were loaded into stem QuickEffects on last shutdown
+    QDomElement quickStemEffectPresetsElement =
+            XmlParse::selectElement(root, EffectXml::kQuickStemEffectList);
+    const QDomNodeList quickStemEffectNodeList =
+            quickStemEffectPresetsElement.elementsByTagName(
+                    EffectXml::kChainPresetName);
+    quickStemEffectPresets.reserve(quickStemEffectNodeList.count());
+    for (const auto& qseNode : quickStemEffectNodeList) {
```
daschuer, line 868: I also suffer this compiler error. qseNode is not used.
```suggestion
    for (for (int i = 0; i < quickStemEffectNodeList.size(); ++i) {
```

src/engine/channels/enginedeck.cpp
daschuer: Perfect. Thank you.

daschuer:
How is the state of the QML GUI inegration? I see the controls, but they are reset immoderately. 
What is the best option to test it without?

acolombier:
QML GUI integration should be functional and I intend to have it fully usable there, so if it's not, it's a bug that I should fix!

daschuer:
Maybe this relates to my old Qt version 6.2. 
Is there a minimum requirement that we need to met?

daschuer:
@JoergAtGithub 
How does the testing succeeds on your side? 
Since we don't officially ship QML anyway and my Qt is outdated it shall not block the merge.

acolombier:
@daschuer I am also using Qt 6.2 - it turns out that something was broken on QML. The control should now work correctly.  Also turned out that I had applied the iterator change and committed by accident.

> Before you implement this, have a look on the discussion in PR #11198

Ah I have a pending review on that PR - I guess we could add the feature suggested by @Eve00000 as part of that change or once this PR is merged so we can use the same setting for it?

daschuer (approved):
Works now like a charm. Thank you.

JoergAtGithub:
> @JoergAtGithub How does the testing succeeds on your side? Since we don't officially ship QML anyway and my Qt is outdated it shall not block the merge.

My last tests before removing the non-QML GUI code was flawless. But I can't start QML anymore, neither with main nor this PR (#13600). As this is unrelated, and Daniel tested the QML changes successful on his system, this shouldn't block the merge of this PR.

JoergAtGithub (approved):

JoergAtGithub:
Thank you!

PR #13125, opened 2024-04-18
Reduce log noise with HID device

acolombier:
This is a follow up on #13010. 
Not sure how come I didn't notice these log output previously as there are quite noisy too. I suspect my dev (The S4 Mk3 screen) branch which I use all the time must already have the fix


JoergAtGithub:
Works as expected. Code LGTM! Thank you!

PR #13127, opened 2024-04-18
Playlists: improve table update after deleting (purging) track files

### LINKED ISSUE #13111: DEBUG_ASSERT when deleting track file that's not in the watched library directories

### LINKED ISSUE #13112: allow 'Move track file(s) to Trash' in PLaylists
### Bug Description

need to add `Capability::RemoveFromDisk` to getCapabilities in PlaylistTableModel.

### Version

2.4

### OS

_No response_

ronso0:
Fixes #13111 which I can actually only reproduce with main (couldn't find the change causing it).
But it doesn't hurt to get rid of the double select and have the now separate PlaylistDAO signals for
a) updating the sidebar and
b) reloading the current tracks in 2.4

Also enables _Delete Track Files_ in playlists #13112 

_Though, no matter if main or 2.4, this feels like a bandaid anyway. PlaylistDAO looks unnecessarily complex, a better solution would probably be PlaylistStorage / PlaylistSummary (Crate equivalent) like proposed by @ninomp in #4846_

### DIFF
--- a/src/library/dao/playlistdao.cpp
+++ b/src/library/dao/playlistdao.cpp
@@ -398,7 +398,8 @@ bool PlaylistDAO::removeTracksFromPlaylist(int playlistId, int startIndex) {
         return false;
     }
     transaction.commit();
-    emit tracksChanged(QSet<int>{playlistId});
+    emit playlistContentChanged(QSet<int>{playlistId});
+    emit tracksRemoved(QSet<int>{playlistId});
     return true;
 }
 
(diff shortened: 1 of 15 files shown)


daschuer (approved):
LGTM and still works as desired. Thank you.

PR #13130, opened 2024-04-19
Delete all experimental build environments

daschuer:
This is the final PR feeing up the major portion of disk space on our download server. 
I have kept all environment that are referenced from the mixxx repository any other intermediate environments will be removed.

This is a requirement to continue with the release process.


JoergAtGithub:
Just for reference, if we delete these buildenvs from the download server, than we can still download the identical file from GitHub artifacts?

daschuer:
The artefacts have a timeout after they will be deleted.
We may recreate them by rerunning the CI run. 
But they are all intermedita steps not further referenced.

JoergAtGithub:
I checked seeral hashes and none occured in a code search in mixxxdj/mixxx. LGTM! Thank you!

daschuer:
:-( did not work

PR #13139, opened 2024-04-20
Labeler: Use more fine-grained controller labels

fwcd:
This should make it easier to determine whether PRs update mappings or change the underlying engine code.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -22,7 +22,12 @@ code quality:
           - .pre-commit-config.yaml
           - pyproject.toml
 
-controllers:
+controller backend:
+  - changed-files:
+      - any-glob-to-any-file:
+          - src/controllers/**
+
+controller mappings:
   - changed-files:
       - any-glob-to-any-file:
           - res/controllers/**


.github/labeler.yml
```diff
@@ -22,7 +22,12 @@ code quality:
           - .pre-commit-config.yaml
           - pyproject.toml
 
-controllers:
+controller backend:
+  - changed-files:
+      - any-glob-to-any-file:
+          - src/controllers/**
```
JoergAtGithub, line 28:
```suggestion
      - any-glob-to-any-file:
          - src/controllers/**.ts
          - src/controllers/common**
          - src/controllers/lodash.mixxx.js
```

Or check if the first character of the file name is upper case

fwcd: You're thinking of res/controllers, not src/controllers. If not every res/controllers file is a mapping, maybe we should make that less specific instead?

JoergAtGithub: You're right - I mixed src and res. I looked for files in res/controllers, which are not specific for a particular mapping.

ronso0:
This is ready for merge, isn't it @JoergAtGithub

JoergAtGithub:
Yes, let's merge this!

JoergAtGithub:
Thank you!

Issue #11361, opened 2023-03-12
New Rubberband R3 breaks mono compatibility, phase issues


atskler:
### Bug Description

I just observed that engaging the Rubberband R3 engine it creates serious phase issues and breaks mono compatibility. This can be really serious problem for broadcasting and for partially or totally mono sound systems. For me the issue is more apparent with speakers than with headphones.

Here is a sound sample:

https://on.soundcloud.com/YwQmG

keylock on
0:0.01 default sound
0:8.00 mono mode on
0:13.00 adjusting speed +0.13, 0, -0.13
0:36.00 switching back to stereo
0:40.00 adjusting speed +0.13, 0, -0.13

### Version

2.4-alpha-1419-g0243efdaf8 (main)

### OS

OpenSUSE Tumbleweed

atskler:
I tried the command line version with:
`-R -t 1.013 -3`
and it sounds like the implementation in Mixxx

The following additional option can somewhat improve the situation:
`--centre-focus` Preserve focus of centre material in stereo (at a cost in width and individual channel quality)

Either way it sounds terrible especially with poor mono compatibility and it is really really far from any kind of near-hi-fi quality (as far as a bad 96k mp3 is), so describing it this way in preferences is quite misleading. _Near-total-cpu-consumption-for-pointless-sonic-destruction_ could be a better description of it. Seriously.

Anyone who would play through a mono system will sound really bad.

This library maybe could be good for individual solo instruments, but not for complex final mixes.

atskler:
I experimented with it a little. For R3 probably the Mid/Side (M/S) processing could be the best compromise (and stamping it "experimental" in preferences). Although I tried it with one track. (In M/S processing the usage of `--centre-focus` is forbidden.)

Downmix to mono would just create another problems, you would need 6dB headroom for it and probably it will result in rightful user complaints, and...

The last question is how it will sound in Mixxx when it engages just above or below 0% of tempo change, as at 0% it does no processing. Probably 0% should be not allowed, it should be skipped, the 0% should be +0.001% or -0.001% to circumvent the sound difference at the point of engage.

Again this is not an ideal solution. It will disturb the original M/S ratio or balance (it translates as modified stereo wideness), but if Rubberband R3 is needed this is probably the best compromise. Until someone suggest a better one.

atskler:
(It seems that M/S processing produces somewhat more pleasing results with R2 also. Still R2 M/S and R3 M/S is far from near hi-fi sound.)

[...]

PR #4898, opened 2022-08-16
PitchShiftEffect: decrease and report latency

davidchocholaty:
This PR improves the current implementation of the Pitch shift effect. The implementation approach is still the "push" model as known. The new version decreases the amount of latency that the effect produces. The implementation is based on the public ladspa-lv2 example implementation by RubberBand. The main difference is, that the maximum number of samples, that will be provided into RubberBand is not set, but instead of it, the minimum required number of samples is provided and processed. Based on the processing of smaller chunks of the input buffer the latency is recognizably decreased.

To avoid the dropouts which are possible in the previous implementation, the specific amount of zero samples is prefilled into the RubberBand. Based on that, the sample dropouts may occur only for the extreme settings changes. The implementation is possible without prefilling the zero samples and the latency decreased even more, anyway, a greater amount of sample dropouts occur.

The known issue is, that the sound quality for the new implementation, primarily for the +12 semitones settings looks a little bit worse to me, but IMO it could be probably solved by the RubberBand option settings.

In this PR the latency is propagated into the delay handling of the effect chain too. For now, the initial delay based on the prefilled samples is provided. Based on that, the latency reporting works properly for the initial setting, however, after dynamic pitch changes, the propagated amount of delay is not valid. This issue should be fixed in this PR too.

Please share your thoughts, comments and proposed changes to the new implementation. Thank you.

PR #13143, opened 2024-04-20
Multithreaded rubberband

acolombier:
<details>

<summary>Slightly outdated info </summary>

This is a naive attempt to parallelise rubberband processing as per suggested [here](https://github.com/breakfastquay/rubberband/issues/89).

(Test using 44100Hz/11.6ms output settings)

So far the result are very promising! previously, audio drop would occur as soon as start playing a second stem deck, when using `powersave` governor, and third stem deck when using `performance`, with both R2 and R3.

Now I can play 4 stem decks in R3 in `powersave`, with near no audio drop, but it increase when I increase pitches on decks. No issue in `performance`, although the indicator is very high

In R2 there is no audio drop at all, and CPU indicator remains in the green, even in `powersave`!

```
$ nproc --all
20
$ cat /proc/cpuinfo
...
model name	: 12th Gen Intel(R) Core(TM) i7-12700H
cpu MHz (max): 2200 in powersave, 4100 in performance
```

</details>

Adds a worker pool for RubberBand. This pool should be used a single threaded environment (engine thread) and will have to be reviewed if the engine is moved to a multi thread design. This decision was made since such a refactor will likely invalidate the need for such a task specific pool.


JoergAtGithub:
This looks indeed very promissing! I can now run 2 stem decks in parallel and it worked stable.

But the thread utilization not optimal yet. Here you see 3 stem decks running in parallel (green is calculation time, red is syncronisation betwee threads (waiting for a mutex etc.)):
grafik

Here you see one deck playing a normal stereo track, and two stem decks (main engine thread is on top):
grafik

acolombier:
I assume you would like to see the deck computing in parallel, right?

I'm afraid that won't only require a refactor of the RubberBand, but rather a refactor of the audio engine which processes the channel sequentially. 

Also, note that simple deck will not use the threaded version and will perform RubberBand in the calling thread 

Or did you mean something else?

acolombier:
Note that this is only a PoC, and production ready solution could include to make a global pool of thread, instead of owned thread per channel (meaning a max of 4 thread, rather than the current 4(deck)x4(stem))

JoergAtGithub:
> I assume you would like to see the deck computing in parallel, right?

Everything in parallel is too much, at least for my laptop. A thread stealing pool with e.g. 4 workers seems more appropriated. 

> I'm afraid that won't only require a refactor of the RubberBand, but rather a refactor of the audio engine which processes the channel sequentially.

This would be of cause beyond the scope of the stems project, but from the pure technical viewpoint, I think this is the way to go. There can be many running decks, if we think about the use of samplers.

acolombier:
With the amount of in-flight PR for the STEM feature, I will pause that one for now. I guess that proves that we have some quick wins to action in order to make STEM more usable. 
Once we have some of these out of the way (at least till #13086 as this is based on it), I will look in implementing this properly if that's okay with you.

acolombier:
I think this feature should be ready for review and testing. I believe the thread usage should now be optimal. I have tested parameters and the current one seems to be the best, slightly better than my initial commit.

JoergAtGithub:
I can't load STEMs files anymore. I fear something got broken in #13044 :
grafik

daschuer:
The last commit contains a lot of useful refactoring. 
Unfortunately it makes it hard to review the essence of multi threading. 

Can you please create a separate PR with these changes. I expect that we can merge them quickly.

acolombier:
> No provider(s) registered for file extension "stem.mp4"

That's very odd! Just to double check, did you use `STEM=ON`?

> Can you please create a separate PR with these changes. I expect that we can merge them quickly.

Yes, no problem, I will remove the STEM changes, only dependency is the stem channel count constant, so I will hardcore it for now and put a `TODO`

daschuer:
This solution unfortunately involves a priority inversion. 
The critical high prio real time thread waits for the lower prio workers, which the OS can freely reassign to other tasks. The whole engine is on the risk. 

In the best case you gain a better CPU utilisation of otherwise idling cores.  This comes with the cost of lost deterministic behavior which is more important IMHO. 

A better working solution would be a buffer swap algorithm, that just swaps the prepared buffers from the engine thread and does the preparation in the low priority workers. The workers will have a whole cycle to finish the task which may be finished even in critical time situations, if not, only one stem is effected. 

This may doubles the latency, but likely allows to compensate that by halving the minimum possible latency.

daschuer:
Did you consider the poor man's solution of mixing only two buffers of temps before doing sound stretching?

acolombier:
> This solution unfortunately involves a priority inversion.

Could that be solve by spawning the stretcher threads with the same priority than the engine?

> The workers will have a whole cycle to finish the task

Does it mean it would need a larger pool so it can complete all decks (4 decks * 4 stems)? If yes, this means more chance for a thread not to complete in time. I guess we could add another scheduler thread we a fifo job queue to help with priority, but this would make the solution significantly more complex.

> Did you consider the poor man's solution of mixing only two buffers of temps before doing sound stretching?

Do you mean mixing the stems before doing the stretching?

JoergAtGithub:
> 
> > Did you consider the poor man's solution of mixing only two buffers of temps before doing sound stretching?
> 
> Do you mean mixing the stems before doing the stretching?

Please not! This, this would break the whole effects architecture of the STEMs feature!

acolombier:
> Please not! This, this would break the whole effects architecture of the STEMs feature!

Yes, that was my next point! :) 
But I think I misunderstood what @daschuer is suggesting. Could you elaborate on that poor man's solution?

JoergAtGithub:
> This solution unfortunately involves a priority inversion. The critical high prio real time thread waits for the lower prio workers

This is not the case here, all involved threads have the same priority. I analyzed the behavior already above, the threading works absolute reliable on Windows, and I guess for Linux as well, because Antoine developed it on Linux.

JoergAtGithub:
> > I guess for Linux as well, because Antoine developed it on Linux
> 
> I guess one risk could be machines with a small amount of cores, since I tested on a beafy CPU. I might be able to test that on a RPI 3 or 4, but can't guarantee

The common approach in low latency audio processing is to define the number of worker threads depending on the number of physical CPU cores (e.g. half the number of physical CPU cores is used for the thread stealing audio processing pool)

acolombier:
> number of worker threads depending on the number of physical CPU cores

Currently, it will use 4 cores (3 worker thread + the engine thread). Should I make that value dynamic (keep as is for 4 and more CPUs, 2 CPUs -> 1 worker thread + the engine thread, 1 CPUs -> no worker thread + the engine thread)?

acolombier:
I've rebased to remove the STEM PR dependecies. Use commit faba575615eba049f8371ea2946a7f078425ee7b if you want to test this with STEM.

Did some test on stereo, and one worker per channel seem to be even more efficient, at least on stereo deck only. @JoergAtGithub could you see if you see better performance with multiple stereo decks?

daschuer:
> Does it mean it would need a larger pool so it can complete all decks (4 decks * 4 stems)?

Not sure, if you trust the scheduler, you may issue one thread per channel. Otherwise issue one thread per CPU core and manage scheduling yourself. However there is no guarantee you really receive all the required CPU cores.

daschuer:
>>> Did you consider the poor man's solution of mixing only two buffers of temps before doing sound stretching?

>> Do you mean mixing the stems before doing the stretching?

> Please not! This, this would break the whole effects architecture of the STEMs feature!

We need effects with stems, that clear. The ideal case would be a fully n-channel engine, like discussed in the Dolby Atmos thread. This is a huge project that will be much bigger than this GSoC project. So we need to find something reliable that is available with a reasonable effort. 

My original idea was to creates the desired stem downmixes before the scaler. If we only want to mute vocals, we have still only one stereo channel. 

If you want to put reverb in the vocals,  we need a dry (rest) and a wet mix (vocals) = two channels. 

Since Rubberband needs some preprocessing, switching it on during playback is probably not a good idea. That's why I have proposed to always do the dry and wet mix and then stretch. 
This will may carry us until we have a n-channel effect region.

JoergAtGithub:
> > No provider(s) registered for file extension "stem.mp4"
> 
> That's very odd! Just to double check, did you use `STEM=ON`?

Indeed, I just missed to specify this CMake setting.

acolombier:
> If you want to put reverb in the vocals, we need a dry (rest) and a wet mix (vocals) = two channels.

I guess the problem with assumption is that it is quite limiting. For example, you may want to put a low pass filter on drum, and reduce bass volume by half. This means that you can never had a single wet/dry setup

> This will may carry us until we have a n-channel effect region.

I guess the alternative could be to have the `Engine` to perform some of the channel processing in parallel - technically, only missing needs to be done sequentially/in single thread, things like stretching or reader management could be distributed.

JoergAtGithub:
> > If you want to put reverb in the vocals, we need a dry (rest) and a wet mix (vocals) = two channels.
> 
> I guess the problem with assumption is that it is quite limiting. 

Indeed, this would be a massive limitation. I strongly recommend to stay with the current 4 stem architecture!

daschuer:
>> If you want to put reverb in the vocals, we need a dry (rest) and a wet mix (vocals) = two channels.

> I guess the problem with assumption is that it is quite limiting. For example, you may want to put a low pass filter on drum, and reduce bass volume by half. This means that you can never had a single wet/dry setup

Yes you are right. Please treat my post as brainstorming for ideas. Not a personal demand or such.

daschuer:
Not sure if this is helpful: 
http://openavproductions.com/audio-engine-design/
Maybe it gives us some Ideas.

We cannot implement it right away, because the engine currently calculates the current tempo and uses it right away. 
For such a buffering attempt we need to calculate the tempo for the next buffer, instruct Rubberband to scale and then use the buffer requested during the previous callback.

acolombier:
Would this not double latency as well? 
Perhaps to refocus slightly the discussion on this PR, do you still think it is fundamentally flawed or do you reckon with some minimal tweaking, we could get that merged?

JoergAtGithub:
The original stems version ( https://github.com/mixxxdj/mixxx/commit/faba575615eba049f8371ea2946a7f078425ee7b ) works for me in general, but the head version of this PR, with with removed stems, plays all tracks very disturbed - somenthing seems to got broken during the refacturing.

daschuer:
> Would this not double latency as well?

Yes, unfortunately. 

> do you still think it is fundamentally flawed or do you reckon with some minimal tweaking, we could get that merged?

Unfortunately yes, at least when reading literature ...

But from the pragmatic side, it improves the situation at least for high core numbers significantly. We just need to take care that this does not introduce a regression for the stereo case. 

How about cut out a refactoring PR first without multi threading. Than add the multi-threading for the stem case only. 
retrieve() and process()

Maybe we can manage to put the tempo calculation at the beginning of the callback for all channels and the buffer retrieval at the end. That we have on loop more over the channels.

acolombier:
> somenthing seems to got broken during the refacturing.

I have changed this version to perform multi threading on each mono channel, as it was showing improvement on my end. I will revert that as I guess it might not bring improvement on all setup.

> add the multi-threading for the stem case only.

Yes, that is already the case: the wrapper logic was to only use the worker pool if the number of channel is more than stereo

> Maybe we can manage to put the tempo calculation at the beginning of the callback for all channels and the buffer retrieval at the end. That we have on loop more over the channels.

I guess this is only shifting the doubled latency issue, isn't it? E.g if a seek occurs between the tempo calculation (the second loop run - and stretching job trigger I assume you meant), what has been processed by the stretcher job is now invalid and the stretcher itself usually needs a flush after seeks too, leading to additional thread sync (not needed currently since the fan-in and -out is part of the same cycle)

> Unfortunately yes, at least when reading literature ...

I'd be curious to hear more on that. That's now has a few years old since, but I remember learning that a thread join will always wake up the thread being joined, since the waiting thread will get its kthread context switched to the joined thread. This means that in worst case scenario, one a CPU/core starving machine, the computation will run the 4 threads sequentially (kthread will run each thread one by one, so 2 channels per 2 channels), which is equivalent to what we have today. Only exception is the risk of context switch which may happen on the wait syscall, but that's probably negligible since most (all?) platforms have preemptive kernels anyway.

JoergAtGithub:
> Only exception is the risk of context switch which may happen on the wait syscall, but that's probably negligible since most (all?) platforms have preemptive kernels anyway.

Yes, Linux, macOS and Windows have preemptive kernels nowadays. Only very old versions like macOS 9 had cooperative kernels.

JoergAtGithub:
Did you consider to use QThreadPool and QRunnable instead of implementing RubberBandWorkerPool?
https://doc.qt.io/qt-6/threads-technologies.html#qthreadpool-and-qrunnable-reusing-threads 
I could imagine to have a dedicated QThreadPool of engine threads, which could be used in future for other tasks than Rubberband too.

daschuer:
>>  add the multi-threading for the stem case only.

> Yes, that is already the case: the wrapper logic was to only use the worker pool if the number of channel is more than stereo

Thank you. Let's really split this PR up to make changes traceable reviewable. Some of them are obvious and can be merged immediately. 

>> Maybe we can manage to put the tempo calculation at the beginning of the callback for all channels and the buffer retrieval at the end. That we have on loop more over the channels.

> I guess this is only shifting the doubled latency issue, isn't it? 

It just shifts the moment when the tempo is consumed. As you see in @JoergAtGithub CPU/thread utilization graph.
You see that the engine thread has dedicated wait periods for each deck. This is due to the loop in https://github.com/mixxxdj/mixxx/blob/8f647908af460e53fe8a860b6ce0964a93a55112/src/engine/enginemixer.cpp#L367

> A solution would be to to split up pChannel->process() at submit(). We already have a split up in two postProcess() functions. 
That's now has a few years old since, but I remember learning that a thread join will always wake up the thread being joined, since the waiting thread will get its kthread context switched to the joined thread.
 
You cannot avoid these joins entirely. But every single instance may bite us in overload situations. We need to meet the timing requirements of the DAC clock. If we fail it, we have currently the issue that the old buffer is repeated. This creates a click noise putting speakers on the risk and a tempo shift that might be notable on the dance floor. 

With the purposed solution of the link, we use only the DAC clock sync via buffering, which involves latency. This means that if a single buffer or all are not ready, we may introduce silence, but the transport is maintained. The other benefit is that rubber band has much more time, to perform the task and less risk to fail the deadline.

My takeaways for our situations are these: 
* Maximize the time for Rubberband, by submit() early and retrieve() late, in a single callback to not double up latency.
* Use only one join in EngineMixer 

Not sure how to effectively handle overload situations. A join with a timeout would be nice, but how to determine the timeout, detect stalled workers .. because there is a lot of processing following which needs also time.

acolombier:
> Did you consider to use QThreadPool and QRunnable instead of implementing RubberBandWorkerPool?

I did not because I wanted to keep the pool lock-free and rely only on atomic counters, since the pool always gets interaction from a single thread (the engine thread).

> I could imagine to have a dedicated QThreadPool of engine threads, which could be used in future for other tasks than Rubberband too.

Yeah, I think that makes sense at Engine level, but I tried to keep that feature as small as possible and localised to RubberBand itself. 

> * Maximize the time for Rubberband, by submit() early and retrieve() late, in a single callback to not double up latency.

I'm still confused on the submit early; I assume your suggesting that submit could be invoked on the N-1 buffer cycle, so it can be retrieve on the cycle N. This still mean that if some event occurred between N-1 and N (e.g seek, loop, rate change, ...), the submit is now invalid and either you delay the event handling at N+1 (and double the latency) or perform a just in time submit, leading to low determinism on performance. Or am I completely missing the point?

Overall, I am not arguing to redesign the engine to allow a wider flexibility on sequential vs parallel processing, I think we should definitely start to put something together and I'd be happy to help with the larger effort, but I just would like to remember that this PR is meant to be a minimal change to make RubberBand usable with STEM. I will make sure that stereo processing doesn't change, but I think we should agree that STEM handling will be suboptimal as well and it's fine, because it is significantly better that what it is today (that is unusable)

Swiftb0y:
> Not sure if this is helpful:
http://openavproductions.com/audio-engine-design/
Maybe it gives us some Ideas.

fyi there was also some discussion on Push vs Pull model in the EngineEffectsDelay-related PRs which were also discussing strategies on dealing with rubberband-induced latency: https://github.com/mixxxdj/mixxx/pulls?q=is%3Apr+rubberband+author%3Adavidchocholaty especially #4898

daschuer:
> I assume your suggesting that submit could be invoked on the N-1 buffer cycle, so it can be retrieve on the cycle N.

No, that is the solution form the linked page. a fully lock free solution. But since it required a major redesign it is probably not appropriated here. By the way, I  consider the double latency not an issue, because we currently need to increase the buffer size when using rubberband. So we may reduce the buffer size again to compensate the doubled latency effect. 

In my last post I have suggested something different: Just to do the submit() in a separate loop but during the same callback and using a single lock in engine mixer. Here the possible iterations: 
```
T = Calculate Transport 
P = Process Scaler 
_ = Idle 
> = Play buffer 

Current main: 
CH1: TP_________>
CH2: __TP_______>   

Stems without threading 
CH1: TPPPP______>
CH2: _____TPPPP_>  

Stems with this PR (lock per channel)  
CH1: T__________>
   : _P_________>
   : _P_________>  
   : _P_________>
   : _P_________>
CH2: __T________>
   : ___P_______>
   : ___P_______>
   : ___P_______>
   : ___P_______>

Stems as suggested (single lock in engine mixer)
CH1: T__________>
   : __P________>
   : __P________> 
   : __P________>
   : __P________>
CH2: _T_________>
   : __P________>
   : __P________>
   : __P________>
   : __P________>

Stereo as suggested (single lock in engine mixer)
CH1: T__________>
   : __P________>
CH2: _T_________>
   : __P________>
```
So you see, that we have a better thread utilization and a longer idle time for the case we have not enough free cores. And even the non stereo case will benefit. This comes with no extra latency and only one lock in the engine thread. 

Luckily this PR is somehow on the route to this, so I agree that this could be an intermediated step. However it should be clear that it is not that stable in overload situation, because

acolombier:
I have reverted the change so stereo channel will proceed as previously

daschuer:
Thank you for the good discussion. I am convinced now this is a good step forward because it is on a good route. I am unfortunately not that this is one commit. Can you move the refactoring of the stereo route into a separate PR or at least commit with a reasonable commit message?

acolombier:
> Can you move the refactoring of the stereo route into a separate PR or at least commit with a reasonable commit message?

I have removed all impact on the stereo channel, so they will proceed in the same way than previously, and [scale directly](https://github.com/mixxxdj/mixxx/blob/4bf8b86f686256f4d88a26fcec34a5e3729319e6/src/engine/bufferscalers/enginebufferscalerubberbandthread.cpp#L158) in the engine thread. Happy to split in multiple commit, but not sure I can see a clear cut in the PR, could you highlights the bits that should live as a dedicated commit?

daschuer:
My idea was to have the introduction of the RubberbandWrapper in a first commit and the introduction of the threading in an other. 
Does that make sense? 

It would be also nice to have enginebufferscalerubberbandthread.h split in three files one per class. 

m_pRubberBand needs to become m_rubberBand because it s no longer just a pointer.

acolombier:
@daschuer I have addressed your request

@JoergAtGithub I found out why the mono split wasn't working. I know you mentioned that had some difficulties having 4 decks on R3 with your setup. If you can, could you try this version and edit [this const](https://github.com/mixxxdj/mixxx/blob/8cad156351af6997b233ce05e48273fa36d13151/src/engine/bufferscalers/rubberbandworker.h#L19) and set to `mono`? If you see improvement, I'd be happy to slightly extend this PR to add an option in preference to distribute stereo scaling in two mono jobs 
image

I think last thing we are missing is the use of `QThread::idealThreadCount` to adjust the pool size to the best size depending of the hardware. 
Also, are we happy with not introducing Runnable QThreadPool at that stage yet?

JoergAtGithub:
> you mentioned that had some difficulties having 4 decks on R3 with your setup. If you can, could you try this version and edit [this const](https://github.com/mixxxdj/mixxx/blob/8cad156351af6997b233ce05e48273fa36d13151/src/engine/bufferscalers/rubberbandworker.h#L19) and set to `mono`?

I tested this with `mono()` and `stereo()` and `mono()` is significiantly better! With `stereo()` the 4 decks with Rubberband R3 sound awful due to the large amount of buffer underruns, with `mono()` I had only 3 buffer underruns at the beginning after loading the tracks.

JoergAtGithub:
> I think last thing we are missing is the use of `QThread::idealThreadCount` to adjust the pool size to the best size depending of the hardware.

I recommend to keep this out of this PR, because this function returns just the number of logical processors, and we have to consider also other threads and processes. This is a topic on it's own.

JoergAtGithub:
> Also, are we happy with not introducing Runnable QThreadPool at that stage yet?

@daschuer Do you aggree with that?

acolombier:
> I tested this with `mono()` and `stereo()` and `mono()` is significiantly better! With `stereo()` the 4 decks with Rubberband R3 sound awful due to the large amount of buffer underruns, with `mono()` I had only 3 buffer underruns at the beginning after loading the tracks.

Glad to hear! @daschuer what do you think of making this accessible in the options? I know we said we don't want to change the rubberband for stereo by default, but it looks like this could help on some setups

daschuer:
Yes sure. I am however unsure what this means to stereo compatibility. I have just skimmed through the Rubberband Documentation but I have not found anything yet. On the other and I remember a discussion. 
Maybe we can just compare the output of both options. If they are bit perfect equal mono can become our new default.

daschuer:
It is the ability that the stereo signal can be later still mixed down to a mono singnal without that frequencies are killed out. 

Just found this in the 3.2 CHANGELOG: 
[ * Fix inability of the R3 engine to produce mono-compatible output](https://github.com/breakfastquay/rubberband/blob/5d296019ff0fd6085fea0838155b0449a4606397/CHANGELOG#L31)

Swiftb0y:
See #11361

JoergAtGithub:
I fear that the problem with the mono downmix functionality could catch us up, when we introduce stem mixing in the engine. After all, we're doing nothing other than mixing 4 independent rubberband channels together. But the OptionChannelsTogether option only seems to work for a single stereo channel pair.

acolombier:
I don't this would be a problem for stem: In principle, stem is only like playing different track on different decks, unlike channels of a stereo set.

> If they are bit perfect equal mono can become our new default.

By the sound of it, there is too many risk going with this as default. Should I still put the option in the setting, with perhaps a warning that references #11361?

JoergAtGithub:
An option would be great, as only some users will be affected by the Mono downmixing issue. But I wonder how to label it, to make the behavior clear to the user.

acolombier:
Perhaps I could put a ⚠️ next to it with a tooltip explaining the risk?

JoergAtGithub:
I tested this locally on Windows 11 and the new UI setting works great!

JoergAtGithub:
There are CI fails now!

acolombier:
Yep, the tests are broken. Just waiting to see what @daschuer and @Swiftb0y think of it! :) 
Will revert otherwise

Swiftb0y (commented):
This comment was pending for a long time. Idk when I'll find time/energy to properly review this.

src/engine/bufferscalers/enginebufferscalerubberband.cpp
```diff
@@ -404,7 +413,10 @@ void EngineBufferScaleRubberBand::reset() {
     }
     while (remaining_padding > 0) {
         const size_t pad_samples = std::min<size_t>(remaining_padding, block_size);
-        m_pRubberBand->process(m_bufferPtrs.data(), pad_samples, false);
+        {
+            ScopedTimer t(u"RubberBand::process");
```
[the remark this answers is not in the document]
Swiftb0y: also in light of #13258, use `QStringLiteral`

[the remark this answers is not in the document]
acolombier: > scoped timer is not very realtime save afai can tell.

Does that matter? I think this is mostly to give a broad overview of the where most of the CPU time goes. What else should I use otherwise?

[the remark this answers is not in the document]
Swiftb0y: > Does that matter? 

It does in realtime code where non-realtime-safe operations (such as (de-)allocating memory) can cause the audio code to run for too long (resulting in buffer underruns). In #13236 I tried to reduce that a little, but the underlying `StatManager` member functions are not realtime safe (in fact they lock, though I don't know how contended that lock is). 

> What else should I use otherwise?

I don't really know. I think the other solutions we had were fairly ugly/low level, so there is no good way of tracking audio code right now. I think you can keep it, I'm just saying that it could result in buffer-underruns with `--developer`. Since its limited to `--developer`, its probably okay. In the longterm, we should consider refactoring StatsManager to use a lock-free queue instead.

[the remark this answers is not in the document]
acolombier: Ah good point. 
Do you think it's best to remove entirely or since it limited to `--developer`, it is fine to keep?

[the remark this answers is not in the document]
Swiftb0y: I'd say its fine to keep. If you really can't tolerate underruns in `--developer` you can max out the buffersize, that should probably work around that. Though in the mid-term future, we should get `StatsManager` lock-free (ideally wait-free too).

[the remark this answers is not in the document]
Swiftb0y: but I'd like to get @daschuer's opinion on this.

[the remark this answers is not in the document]
daschuer: We use the ScopedTimer in other places of the engine already. So it is fine here as well IMHO.

JoergAtGithub:
Please solve the conflict in src/test/playermanagertest.cpp

src/engine/bufferscalers/rubberbandworker.h
```diff
@@ -0,0 +1,49 @@
...
+    QMutex m_waitLock;
+    QWaitCondition m_waitCondition;
+    // Whether or not the scheduled job as completed
+    bool m_completed;
+
+    // Used by RubberBandWorkerPool to manage thread availability.
+    // RubberBandWorker only clears the flag, once the job is completed
+    std::atomic_flag m_assigned;
+
+    friend RubberBandWorkerPool;
```
Swiftb0y, line 48: I know that the legacy code is generally littered with `friend`s, but usually they're discouraged (even for tests). So this is a codesmell.

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+void RubberBandWrapper::setPitchScale(double scale) {
+    for (auto& stretcher : m_pInstances) {
+        stretcher->setPitchScale(scale);
+    }
+}
+
+bool RubberBandWrapper::isValid() const {
+    return m_pInstances.size();
```
Swiftb0y, line 171:
```suggestion
    return !m_pInstances.empty();
```

src/engine/bufferscalers/rubberbandwrapper.h
```diff
@@ -0,0 +1,35 @@
...
+
+    // The following method are helper function and do not wrap any RubberBand calls
+    void clear();
+    void setup(mixxx::audio::SampleRate sampleRate,
+            mixxx::audio::ChannelCount chCount,
+            const RubberBand::RubberBandStretcher::Options& opt);
+    bool isValid() const;
+
+  private:
+    std::vector<std::unique_ptr<RubberBand::RubberBandStretcher>> m_pInstances;
```
Swiftb0y, line 34: why the unique_ptr?
```suggestion
    std::vector<RubberBand::RubberBandStretcher> m_rbInstances;
```

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+    return require;
+}
+int RubberBandWrapper::available() const {
+    int available = std::numeric_limits<int>::max();
+    for (auto& stretcher : m_pInstances) {
+        available = qMin(available, stretcher->available());
+    }
+    return available;
+}
```
Swiftb0y, line 48: this will return int max when the vector is empty. That doesn't seem right. Should be 0 imo.

Swiftb0y (commented):
preliminary nitpick comment. In general I really feel this could be simplified by use of some higher-level threading abstraction instead of implementing the worker threads and the pool ourselves.

src/engine/bufferscalers/rubberbandworker.h
```diff
@@ -0,0 +1,49 @@
...
+            bool final);
+    void stop();
+
+  private:
+    struct Job {
+        RubberBand::RubberBandStretcher* instance;
+        const float* const* input;
+        size_t samples;
+        bool final;
+    };
```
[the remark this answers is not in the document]
Swiftb0y: Interesting that its not an issue in the RubberbandStretcher API though

acolombier:
> In general I really feel this could be simplified by use of some higher-level threading abstraction instead of implementing the worker threads and the pool ourselves

Do you know what guarantee with have around the realtime readiness of QThreadLoop? Please also not that the reason for implementing it the worker as well was to have a lock-free solution, which can easily be single threaded if needed.

Swiftb0y:
> Do you know what guarantee with have around the realtime readiness of QThreadLoop?

I'd be surprised if it was worse than the current solution. At least I wouldn't choose this lower-level solution until we have confirmed it is necessary. Otherwise this smells like premature-optimization.

> Please also not that the reason for implementing it the worker as well was to have a lock-free solution

Can you elaborate? The RubberBandWorker is anything but lock-free IMO.

>  which can easily be single threaded if needed.

That should also be possible using higher-level primitives.

acolombier:
> I'd be surprised if it was worse than the current solution. At least I wouldn't choose this lower-level solution until we have confirmed it is necessary. Otherwise this smells like premature-optimization.

Fine, I will but QThreadPool/QRunnable everywhere.

> Can you elaborate? The RubberBandWorker is anything but lock-free IMO.

Sorry, I forgot I had reverted the atomic-only version.

src/test/signalpathtest.h
```diff
@@ -142,6 +145,18 @@ class BaseSignalPathTest : public MixxxTest, SoundSourceProviderRegistration {
...
+    void SetUp() override {
+        RubberBandWorkerPool::createInstance();
+    }
+#endif
+
+#ifdef __RUBBERBAND__
+    void TearDown() override {
+        RubberBandWorkerPool::destroy();
+    }
+#endif
```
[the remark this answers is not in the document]
acolombier: I don't understand what you want me to do here. Could you elaborate?

acolombier:
> I'd be surprised if it was worse than the current solution. At least I wouldn't choose this lower-level solution until we have confirmed it is necessary. Otherwise this smells like premature-optimization.

Fine, I will but QThreadPool/QRunnable everywhere.

> Can you elaborate? The RubberBandWorker is anything but lock-free IMO.

Sorry, I forgot I had reverted the atomic-only version.

JoergAtGithub:
> > I'd be surprised if it was worse than the current solution. At least I wouldn't choose this lower-level solution until we have confirmed it is necessary. Otherwise this smells like premature-optimization.
> 
> Fine, I will but QThreadPool/QRunnable everywhere.

@daschuer Do you aggree to this approach?

src/test/signalpathtest.h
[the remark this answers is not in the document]
Swiftb0y: otherwise rubberband-less builds fail because otherwise the code further below will try to call the nonexistent methods.
```suggestion
    void SetUp() override {
#ifdef __RUBBERBAND__
        RubberBandWorkerPool::createInstance();
#endif
    }

    void TearDown() override {
#ifdef __RUBBERBAND__
        RubberBandWorkerPool::destroy();
#endif
    }
```

src/engine/bufferscalers/rubberbandwrapper.h
acolombier: copy constructor of 'RubberBandTask' is implicitly deleted

Swiftb0y: Ah yeah, no Copy nor move, difficult for vector to relocate then. Deserves a comment IMO.

acolombier:
I have added a version with QThreadPool/QRunnable. I don't have much experience with this so I'm sure it could be done better.

JoergAtGithub:
The new version asserts at first track loading for me:
grafik

src/engine/bufferscalers/rubberbandworkerpool.cpp
```diff
@@ -0,0 +1,33 @@
...
+RubberBandWorkerPool::RubberBandWorkerPool(UserSettingsPointer pConfig)
+        : QThreadPool() {
+    bool multiThreadedOnStereo = pConfig &&
+            pConfig->getValue(ConfigKey(QStringLiteral("[App]"),
+                                      QStringLiteral("keylock_multithreading")),
+                    false);
+    m_channelPerWorker = multiThreadedOnStereo
+            ? mixxx::audio::ChannelCount::mono()
+            : mixxx::audio::ChannelCount::stereo();
+    DEBUG_ASSERT(0 == mixxx::kEngineChannelCount % m_channelPerWorker);
```
Swiftb0y, line 17: NIT: [yoda](https://en.wikipedia.org/wiki/Yoda_conditions)
```suggestion
    DEBUG_ASSERT(mixxx::kEngineChannelCount % m_channelPerWorker == 0);
```

src/engine/bufferscalers/rubberbandworkerpool.cpp
```diff
@@ -0,0 +1,33 @@
...
+    // waiting all workers to complete. During performance testing, this ahas
+    // show better results
+    for (int w = 0; w < maxThreadCount(); w++) {
+        reserveThread();
+    }
+}
+
+RubberBandWorkerPool::~RubberBandWorkerPool() {
+    waitForDone();
+}
```
Swiftb0y, line 33: NIT: unnecessary. QThreadPool already does that

Swiftb0y:
> If any, please let me know once all nits/code-smell comments are in so I can solve them all at once and not build up on my frustration.

This is difficult to do during review, but I can label the respective comments as nits and you can postpone them until we've decided on a more concrete course. Does that sound like a reasonable compromise?

src/engine/bufferscalers/rubberbandtask.h
```diff
@@ -0,0 +1,45 @@
...
+    RubberBandTask(size_t sampleRate,
+            size_t channels,
+            Options options = DefaultOptions);
+
+    /// @brief Submit a new stretching task
+    /// @param stretcher the RubberBand::RubberBandStretcher instance to use.
+    /// Must remain valid till waitReady() as returned
+    /// @param input The samples buffer
+    /// @param samples the samples count
+    /// @param final whether or not this is the final buffer
```
Swiftb0y, line 25: NIT: oudated comment

Swiftb0y (commented):
only nits so far. its too late in the day to wrap my head around this

acolombier:
> The new version asserts at first clicking on Play button for me

Some last minute untested change... I believe it should be fixed now

JoergAtGithub:
Works now again:
grafik

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -0,0 +1,168 @@
...
+        return m_pInstances[0]->retrieve(output, samples);
+    } else {
+        size_t ret = 0;
+        for (const auto& stretcher : m_pInstances) {
+            size_t thisRet = stretcher->retrieve(output, samples);
+            DEBUG_ASSERT(!ret || thisRet == ret);
+            ret = qMax(thisRet, ret);
+            output += getChannelPerWorker();
+        }
+        return ret;
```
Swiftb0y, line 59: This will probably cause audible artifacts when one stretcher returns less data than the other ones. This is usually not a problem with rubberband because we know how much data has actually been returned, but in this case we take the max of all stretchers which will cause mixxx to read the unwritten part of the buffer. 
```
                    v max of all retrieved data
C1|#################|
C2|#########--------| <-- less data returned
            ^^^^^^^^ unwritten data (zero or leftover from previous call?)
```

Simply truncating is not an option either as the remaining data from the stretcher will then be lost (unless we manually buffer it (complicated to get right)).
```
            vvvvvvvv all the data that has been lost because its out of the stretcher, but not consumed by the engine.
C1|#################|
C2|#########--------| <-- less data returned
           ^ min of all retrieved data
```

So the only remaining solution is to query the minimum beforehand, and then only retrieve that minimum available. 

```suggestion
        // ensure we don't fetch more samples than we really have available.
        samples = std::min{available(), samples};
        for (const auto& stretcher : m_pInstances) {
            size_t numSamplesRequested = stretcher->retrieve(output, samples);
            // there is something very wrong if we got less samples
            // than were available.
            DEBUG_ASSERT(numSamplesRequested == samples);
            output += getChannelPerWorker();
        }
        return samples;
```

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -0,0 +1,168 @@
...
+        RubberBandWorkerPool* pPool = RubberBandWorkerPool::instance();
+        for (auto& pInstance : m_pInstances) {
+            pInstance->set(input, samples, isFinal);
+            if (!pPool->tryStart(pInstance.get())) {
+                pInstance->run();
+            }
+            input += pPool->channelPerWorker();
+        }
+        for (auto& pInstance : m_pInstances) {
+            pInstance->waitReady();
```
Swiftb0y, line 125: Using `QtConcurrent::blockedMap`  and then running that on subsequences would allow you to remove the explicit synchronization in the RubberbandTask, but the resulting code is also not super nice because of Qt Api restrictions...

src/engine/bufferscalers/rubberbandwrapper.cpp
acolombier: This function should always be called with the value returned by `available`, which is adjusted to the minimum available. For safety, I could call it again in this method to make sure that the user isn't using an arbitrary value, but we assume that callers know what they are doing and will use a value between 0..<ReturnedAvailable> 
I think we need to remember we are not making a public library here, but a wrapper only mean to supersede the bare RubberBand instance used in the scaler, which already doesn't something along the line of `std::min{available(), samples}`

acolombier: https://github.com/acolombier/mixxx/blob/05df96d3ca385dade90caf88187355178870ce71/src/engine/bufferscalers/enginebufferscalerubberband.cpp#L143

Swiftb0y: > This function should always be called with the value returned by available, which is adjusted to the minimum available. For safety, I could call it again in this method to make sure that the user isn't using an arbitrary value, but we assume that callers know what they are doing and will use a value between 0..

Well this is essentially a wrapper, so adhering to the liskov substition principle, the behavior needs to match. The rubberband API doesn't assume you only call it with the `available` samples so this wrapper API that looks the exact same. In the case of rubberband, its only a bug if you set `samples` to some value thats larger than the retrieval buffer, in the current iteration it would be bug to set it to something larger than available because of the previous two issues I was describing. Lets adhere to the rubberband API and only assume that `samples` is representing the buffer size. 

> I think we need to remember we are not making a public library here, but a wrapper only mean to supersede the bare RubberBand instance used in the scaler, which already doesn't something along the line of std::min{available(), samples}

That might be true in this specific case, but in a codebase as large as mixxx, every interface is essentially a library, so in order for the consumers of that interface to stay sane, it still needs to be reasonably well designed.

acolombier: I'll add a `DEBUG_ASSERT`

Swiftb0y (commented):
fyi LGTM if you address the nits

Swiftb0y (approved):
LGTM. merge @daschuer?

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -298,6 +320,21 @@ void DlgPrefSound::slotApply() {
...
+                keylockMultithreadedComboBox->isChecked() &&
+                        keylockMultithreadedComboBox->isEnabled());
+        if (keylockMultithreading !=
+                (keylockMultithreadedComboBox->isChecked() &&
+                        keylockMultithreadedComboBox->isEnabled())) {
+            QMessageBox::information(this,
+                    tr("Information"),
+                    tr("Mixxx must be restarted before the multi-threaded "
+                       "RubberBand settings change will take effect."));
+        }
```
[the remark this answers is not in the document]
Swiftb0y: Are you sure it needs that more work? We support changing the stretcher without restarting too, this doesn't seem that much more complicated.

src/engine/bufferscalers/rubberbandtask.h
```diff
@@ -0,0 +1,45 @@
...
+
+    // Wait for the current task to complete.
+    void waitReady();
+
+    void run();
+
+  private:
+    // Used to schedule the thread
+    QMutex m_waitLock;
+    QWaitCondition m_waitCondition;
```
daschuer, line 38: I think we can Remove the QWaitCondition because we have only a two threads that needs to be synched. 
I think a better solution would be to make m_completed a QSemsphore. It's cooky jar model looks probably nicer.

daschuer (changes requested):
Im finally done with the review. 
Thank you for this good solution.

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -0,0 +1,174 @@
...
+#endif
+}
+void RubberBandWrapper::process(const float* const* input, size_t samples, bool isFinal) {
+    if (m_pInstances.size() == 1) {
+        return m_pInstances[0]->process(input, samples, isFinal);
+    } else {
+        RubberBandWorkerPool* pPool = RubberBandWorkerPool::instance();
+        for (auto& pInstance : m_pInstances) {
+            pInstance->set(input, samples, isFinal);
+            if (!pPool->tryStart(pInstance.get())) {
```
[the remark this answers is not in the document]
acolombier: No being able to start is part of the valid expectations. This happens if there is no more worker available and indicate that the caller should take care of performing the stretching itself (This is why on the follow line, we `run` the task directly on the following line). This is to reduce the chance of main thread to wait, since it will perform active stretching on the last instance (the pool is always sized with at most N-1 for that reason), so chances are that it will be the last one terminating the process, and it won't have to wait on the condition in `waitReady`

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -0,0 +1,174 @@
...
+                pInstance->run();
+            }
+            input += pPool->channelPerWorker();
+        }
+        for (auto& pInstance : m_pInstances) {
+            pInstance->waitReady();
```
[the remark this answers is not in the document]
acolombier: See my comment above - if `tryStart` fails, the task is executed in the main thread, so the task is always ready.

Will document that function better

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -0,0 +1,174 @@
...
+        const RubberBandStretcher::Options& opt) {
+    // The instance should have been cleared, or not set before
+    VERIFY_OR_DEBUG_ASSERT(m_pInstances.size() == 0) {
+        m_pInstances.clear();
+    };
+
+    auto channelPerWorker = getChannelPerWorker();
+    qDebug() << "RubberBandWrapper::setup" << channelPerWorker;
+    VERIFY_OR_DEBUG_ASSERT(0 == chCount % channelPerWorker) {
+        m_pInstances.emplace_back(
```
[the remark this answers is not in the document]
acolombier: The strategy is - since we can not divide the stretching in an evenly distributed number of channels, we fallback to the single instance / not multithreaded implementation. TLDR stretgy is (audio) quality over performance (my understanding is that RB with different channel is likely to produce different result)

src/engine/bufferscalers/rubberbandtask.h
acolombier: I believe we can remove the lock too in that case (`set` and and `wait` are called from the main thread).

src/engine/bufferscalers/rubberbandwrapper.cpp
```diff
@@ -0,0 +1,174 @@
...
+        available = qMin(available, stretcher->available());
+    }
+    return available == std::numeric_limits<int>::max() ? 0 : available;
+}
+size_t RubberBandWrapper::retrieve(float* const* output, size_t samples) const {
+    if (m_pInstances.size() == 1) {
+        return m_pInstances[0]->retrieve(output, samples);
+    } else {
+        // ensure we don't fetch more samples than we really have available.
+        samples = std::min(static_cast<size_t>(available()), samples);
```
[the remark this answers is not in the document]
acolombier: Added an argument so we know the channel buffer size

src/engine/bufferscalers/rubberbandwrapper.cpp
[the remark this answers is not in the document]
daschuer: Ah, I understand. In the unlikely case something got messed up we only create a stereo instance.

daschuer:
Did some brief test and it works good. 
The preferences could be improved though. 
The Multithreaded Check Box should be unchecked when using SoundTouch. 
Does it really require a Restart when changing it?

src/preferences/dialog/dlgprefsounddlg.ui
```diff
@@ -221,6 +218,29 @@
        </property>
       </widget>
      </item>
+     <item row="5" column="1">
+      <layout class="QHBoxLayout" name="horizontalLayout">
+       <item>
+        <widget class="QComboBox" name="keylockComboBox"/>
+       </item>
+       <item>
+        <widget class="QCheckBox" name="keylockMultithreadedComboBox">
```
daschuer, line 227:
```suggestion
        <widget class="QCheckBox" name="keylockMultithreadedCheckBox">
```

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -298,6 +329,21 @@ void DlgPrefSound::slotApply() {
...
+        m_pSettings->setValue(ConfigKey(kAppGroup, "keylock_multithreading"),
+                keylockMultithreadedComboBox->isChecked() &&
+                        keylockMultithreadedComboBox->isEnabled());
+        if (keylockMultithreading !=
+                (keylockMultithreadedComboBox->isChecked() &&
+                        keylockMultithreadedComboBox->isEnabled())) {
+            QMessageBox::information(this,
+                    tr("Information"),
+                    tr("Mixxx must be restarted before the multi-threaded "
+                       "RubberBand settings change will take effect."));
```
daschuer, line 344:
```suggestion
                       "RubberBand setting change will take effect."));
```

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -682,6 +735,41 @@ void DlgPrefSound::settingChanged() {
...
+void DlgPrefSound::updateKeylockMultithreading(bool enabled) {
+    m_settingsModified = true;
+    if (!enabled) {
+        return;
+    }
+    QMessageBox msg;
+    msg.setIcon(QMessageBox::Warning);
+    msg.setWindowTitle(tr("Are you sure?"));
+    msg.setText(
+            tr("<p>Using multi threading may result in <span "
```
daschuer, line 761: The html tags required a developer comment. I would just remove them instead.

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -682,6 +735,41 @@ void DlgPrefSound::settingChanged() {
...
+               "style=\"font-weight:600;\">pitch and tone imperfection</span> "
+               "depending of the platform, leading to mono-incompatibility, "
+               "due to third party limitations. </p><p>Are you sure you wish "
+               "to proceed?</p>"));
```
daschuer, line 765: Is this true? I am not aware of a pitch and tone imperfection in the single channel case. I have just found this: 
https://breakfastquay.com/rubberband/code-doc/classRubberBand_1_1RubberBandStretcher.html

> OptionChannelsApart - Channels are handled for maximum individual fidelity, at the expense of synchronisation. In the R3 engine, this means frequency-bin synchronisation is maintained more closely for lower-frequency content than higher. In R2, it means the stereo channels are processed individually and only synchronised at transients. In both engines this gives the highest quality for the individual channels but a more diffuse stereo image, an unnatural increase in "width", and generally a loss of mono compatibility (i.e. mono mixes from stereo can sound phasy). This option is the default.

> OptionChannelsTogether - Channels are handled for higher synchronisation at some expense of individual fidelity. In particular, a stretcher processing two channels will treat its input as a stereo pair and aim to maximise clarity at the centre and preserve mono compatibility. This gives relatively less stereo space and width than the default, as well as slightly lower fidelity for individual channel content, but the results may be more appropriate for many situations making use of stereo mixes.

So how about this warning:
```
Using multi threading result in a loss of mono compatibility and a diffuse stereo image. It is not recommended during broadcasting or recording. 
```

daschuer (changes requested):

src/preferences/dialog/dlgprefsound.cpp
acolombier: That is much better, thank you.

acolombier:
> Does it really require a Restart when changing it?

Yes, as discussed [here](https://github.com/mixxxdj/mixxx/pull/13143#discussion_r1624529293), from my first attempt, adding dynamic reload was at least doubling the size of this PR if not more. Now, with the QThreadPool refactor, it might not be as bad, but I still would like this to be merged first as @JoergAtGithub and @Swiftb0y have already reviewed the existing code. I still have some time free so I'm happy to look into this right away if we me quick enough on this one.

daschuer (approved):
LGTM, thank you.

PR #13148, opened 2024-04-21
Usability/Keyboard: Add a rebindable keyboard shortcut for editing items as a replacement for F2

cr7pt0gr4ph7:
## Issue Statement

Right now, one cannot easily use a keyboard shortcut to begin editing a track table cell when keyboard shortcuts are **enabled**, because the usual F2 key is bound to something else by default the default `.kbd.cfg` files provided.

## Use Case

I am currently setting up my music library, and it is quite annoying that I can't use keyboard shortcuts to both play a track in the preview deck using `p` (which is only possible when keyboard shortcuts are enabled) and then enter edit mode using the keyboard too (the F2 key is only bound to this function when keyboard shortcuts are disabled).

## Proposed Solution (as implemented)

Add a keyboard-triggerable `EditItem` action (name is up for debate, but matches the existing `GoToItem` action) that can be remapped using `Custom.kbd.cfg`. The default keyboard mappings bind `EditItem` to the `r` key,
basically because it was the only easily-reachable key not bound yet, but due to being able to rebind it myself, I am open for other suggestions. What I considered so far:

1. **Not providing a default mapping**: Kinda bad for discoverability of the feature, but would be fine by me
2. **Binding `EditItem` to `Ctrl+Shift+Return`**: Makes sense from the viewpoint of consistency (because `Ctrl+Enter` is already bound to opening the track menu, and `Enter` is interpreted as "edit this shell" e.g. in Excel and  Google Sheets.
3. **Binding `EditItem` to `Ctrl+e`**: Would work, is easily reachable, but incurs the risk of mistiming the `Ctrl` press and invoking `loop_double` (which is mapped to `e`) instead when editing a lot of tracks.
4. **Binding `EditItem` to `r`**: The key is easily reachable, not mapped to anything important yet, and has the mnemonic of **R**ename to remember it by.

## Further Work

- Adapt the keyboard shortcut table at [mixxxdj/manual](https://github.com/mixxxdj/manual) once the default key binding has been agreed upon.

- Making the "Star Rating" column editable by keyboard: I'm already working on this, and have a PR nearly ready, but due to the "special" behavior of the star rating column that allows you to edit it simply by hovering over & clicking a star without focusing the cell first, care has to be taken so that the feature works as expected in all circumstances.

- Open a context menu providing the option to clear the playing count when attempting to edit the playing count column.

- Possibly adding a toggle option (or separate command) that keeps the "editor mode" active not only when navigating horizontally using `Tab`/`Shift+Tab`, but also when navigating vertically using the `Up` and `Down` arrow keys.

### DIFF
--- a/src/library/trackset/baseplaylistfeature.cpp
+++ b/src/library/trackset/baseplaylistfeature.cpp
@@ -76,7 +76,6 @@ void BasePlaylistFeature::initActions() {
             &BasePlaylistFeature::slotCreatePlaylist);
 
     m_pRenamePlaylistAction = new QAction(tr("Rename"), this);
-    m_pRenamePlaylistAction->setShortcut(kRenameSidebarItemShortcutKey);
     connect(m_pRenamePlaylistAction,
             &QAction::triggered,
             this,
--- a/src/library/trackset/crate/cratefeature.cpp
+++ b/src/library/trackset/crate/cratefeature.cpp
@@ -69,7 +69,6 @@ void CrateFeature::initActions() {
             &CrateFeature::slotCreateCrate);
 
     m_pRenameCrateAction = make_parented<QAction>(tr("Rename"), this);
-    m_pRenameCrateAction->setShortcut(kRenameSidebarItemShortcutKey);
     connect(m_pRenameCrateAction.get(),
             &QAction::triggered,
             this,
--- a/src/widget/wlibrarysidebar.cpp
+++ b/src/widget/wlibrarysidebar.cpp
@@ -178,6 +178,20 @@ void WLibrarySidebar::dropEvent(QDropEvent * event) {
     }
 }
 
+void WLibrarySidebar::editSelectedItem() {
+    // Rename crate or playlist (internal, external, history)
+    QModelIndex selIndex = selectedIndex();
+    if (!selIndex.isValid()) {
+        return;
+    }
+    if (isExpanded(selIndex)) {
+        // Root views / knots can not be renamed
+        return;
+    }
+    emit renameItem(selIndex);
+    return;
+}
+
 void WLibrarySidebar::toggleSelectedItem() {
     QModelIndex index = selectedIndex();
     if (index.isValid()) {
@@ -307,16 +321,7 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
         emit setLibraryFocus(FocusWidget::TracksTable);
         return;
     case kRenameSidebarItemShortcutKey: { // F2
-        // Rename crate or playlist (internal, external, history)
-        QModelIndex selIndex = selectedIndex();
-        if (!selIndex.isValid()) {
-            return;
-        }
-        if (isExpanded(selIndex)) {
-            // Root views / knots can not be renamed
-            return;
-        }
-        emit renameItem(selIndex);
+        editSelectedItem();
         return;
     }
     case kHideRemoveShortcutKey: { // Del (macOS: Cmd+Backspace)
--- a/src/widget/wlibrarysidebar.cpp
+++ b/src/widget/wlibrarysidebar.cpp
@@ -184,10 +184,6 @@ void WLibrarySidebar::renameSelectedItem() {
     if (!selIndex.isValid()) {
         return;
     }
-    if (isExpanded(selIndex)) {
-        // Root views / knots can not be renamed
-        return;
-    }
     emit renameItem(selIndex);
     return;
 }
--- a/src/widget/wlibrarysidebar.cpp
+++ b/src/widget/wlibrarysidebar.cpp
@@ -178,7 +178,7 @@ void WLibrarySidebar::dropEvent(QDropEvent * event) {
     }
 }
 
-void WLibrarySidebar::editSelectedItem() {
+void WLibrarySidebar::renameSelectedItem() {
     // Rename crate or playlist (internal, external, history)
     QModelIndex selIndex = selectedIndex();
     if (!selIndex.isValid()) {
@@ -321,7 +321,7 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
         emit setLibraryFocus(FocusWidget::TracksTable);
         return;
     case kRenameSidebarItemShortcutKey: { // F2
-        editSelectedItem();
+        renameSelectedItem();
         return;
     }
     case kHideRemoveShortcutKey: { // Del (macOS: Cmd+Backspace)
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -1297,6 +1297,14 @@ void WTrackTableView::hideOrRemoveSelectedTracks() {
     restoreCurrentIndex();
 }
 
+/// If applicable, requests that the selected field/item be edited
+/// Does nothing otherwise.
+void WTrackTableView::editSelectedItem() {
+    if (state() != EditingState) {
+        edit(currentIndex(), EditKeyPressed, nullptr);
+    }
+}
+
 void WTrackTableView::activateSelectedTrack() {
     const QModelIndexList indices = getSelectedRows();
     if (indices.isEmpty()) {
--- a/res/keyboard/cs_CZ.kbd.cfg
+++ b/res/keyboard/cs_CZ.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+\
+
+[Library]
+EditItem r
--- a/res/keyboard/da_DK.kbd.cfg
+++ b/res/keyboard/da_DK.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+<
+
+[Library]
+EditItem r
--- a/res/keyboard/de_CH.kbd.cfg
+++ b/res/keyboard/de_CH.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+<
+
+[Library]
+EditItem r
--- a/res/keyboard/de_DE.kbd.cfg
+++ b/res/keyboard/de_DE.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+<
+
+[Library]
+EditItem r
--- a/res/keyboard/el_GR.kbd.cfg
+++ b/res/keyboard/el_GR.kbd.cfg
@@ -168,3 +168,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+`
+
+[Library]
+EditItem r
--- a/res/keyboard/en_US.kbd.cfg
+++ b/res/keyboard/en_US.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+`
+
+[Library]
+EditItem r
--- a/res/keyboard/es_ES.kbd.cfg
+++ b/res/keyboard/es_ES.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+<
+
+[Library]
+EditItem r
--- a/res/keyboard/fi_FI.kbd.cfg
+++ b/res/keyboard/fi_FI.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+<
+
+[Library]
+EditItem r
--- a/res/keyboard/fr_CH.kbd.cfg
+++ b/res/keyboard/fr_CH.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+<
+
+[Library]
+EditItem r
--- a/res/keyboard/fr_FR.kbd.cfg
+++ b/res/keyboard/fr_FR.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+`
+
+[Library]
+EditItem r
--- a/res/keyboard/it_IT.kbd.cfg
+++ b/res/keyboard/it_IT.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+<
+
+[Library]
+EditItem r
--- a/res/keyboard/ru_RU.kbd.cfg
+++ b/res/keyboard/ru_RU.kbd.cfg
@@ -163,3 +163,6 @@ OptionsMenu_DeveloperStatsExperiment Ctrl+Shift+E
 OptionsMenu_DeveloperStatsBase Ctrl+Shift+B
 DeveloperMenu_EnableDebugger Ctrl+Shift+D
 OptionsMenu_EnableShortcuts Ctrl+|
+
+[Library]
+EditItem r
--- a/src/library/librarycontrol.cpp
+++ b/src/library/librarycontrol.cpp
@@ -257,6 +257,18 @@ LibraryControl::LibraryControl(Library* pLibrary)
                 &LibraryControl::refocusPrevLibraryWidget);
     }
 
+    // Control to "edit" the currently selected item/field in focused widget (context dependent)
+    m_pEditItem = std::make_unique<ControlPushButton>(ConfigKey("[Library]", "EditItem"));
+#ifdef MIXXX_USE_QML
+    if (!CmdlineArgs::Instance().isQml())
+#endif
+    {
+        connect(m_pEditItem.get(),
+                &ControlPushButton::valueChanged,
+                this,
+                &LibraryControl::slotEditItem);
+    }
+
     // Control to "goto" the currently selected item in focused widget (context dependent)
     m_pGoToItem = std::make_unique<ControlPushButton>(ConfigKey("[Library]", "GoToItem"));
 #ifdef MIXXX_USE_QML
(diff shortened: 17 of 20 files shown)


JoergAtGithub:
Welcome at Mixxx!

cr7pt0gr4ph7:

Done :+1:

src/widget/wtracktableview.cpp
```diff
@@ -1211,6 +1211,13 @@ void WTrackTableView::hideOrRemoveSelectedTracks() {
     restoreCurrentIndex();
 }
 
+void WTrackTableView::editSelectedItem() {
+    if (state() != EditingState) {
+        QKeyEvent keyEvent(QEvent::Type::KeyPress, Qt::Key_F2, Qt::NoModifier);
+        edit(currentIndex(), EditKeyPressed, &keyEvent);
```
ronso0, line 1217: Why not use the simple method [`edit(index)`](https://doc.qt.io/qt-5/qabstractitemview.html#edit)?

src/proto/httpplugin.proto
```diff
@@ -0,0 +1,35 @@
...
+    // If not specified or empty, retrieve all available tracks instead.
+    optional string playlist_id = 1;
+
+    // A search query that is used to filter the tracks.
+    // If not specified or empty: Retrieve all tracks without applying a filter.
+    optional string search_text = 2;
+}
+
+message GetTracksResponse {
+}
```
ronso0, line 35: Leftover from another branch?
Please remove.

ronso0 (commented):
Thank you for working on this!

I left some notes.
Please also check the CI remarks.

src/proto/httpplugin.proto
ronso0: (rebase)

cr7pt0gr4ph7: Oops, good catch. I was experimenting with creating a HTTP API for making plugging in external music library backends easier, and was accidentally left over during cleaning up and rebasing while preparing the PR branch. Will remove.

src/widget/wtracktableview.cpp
cr7pt0gr4ph7: Thank you for the suggestion. I replaced the `keyEvent` with `nullptr` in the new (rebased) version of this PR (because I ended up not needing for the future follow-up PR related to `StarEditor`). The `EditKeyPressed` flag on the other hand is specified for two reasons:
* The edit is triggered because of the user pressing the platform edit key (resp. a replacement for that key), so it makes sense to pass that knowledge along using the API provided for this.
* I need the flag in the follow-up PR to distinguish certain user-triggered invocations of `edit()` from those that are triggered internally by `QAbstractItemView`.

cr7pt0gr4ph7:
I integrated the review suggestions and fixed the build failures (and have setup the CI pipeline on my fork, so the latter shouldn't happen anymore for future PRs). Thanks for the quick reaction!

src/widget/wtracktableview.cpp
ronso0: IIUC you don't need it for this use case. The simple implementation with just the index calls the base function with `AllEditTriggers`
https://github.com/qt/qtbase/blob/e7362764d4931f255d2377462df8ac7a0d4e7c84/src/widgets/itemviews/qabstractitemview.cpp#L1233-L1240
and that starts editing except the trigger is DoubleClicked or SelectedClicked.
https://github.com/qt/qtbase/blob/e7362764d4931f255d2377462df8ac7a0d4e7c84/src/widgets/itemviews/qabstractitemview.cpp#L2749

The EditKey is already enabled here btw, and IIUC that is captured in keyPressEvent and is not relevant for this new slot `editSelectedItem`.
https://github.com/mixxxdj/mixxx/blob/67a41d9dcdb06b37f57be8e88978756140d05ff2/src/widget/wlibrarytableview.cpp#L31

Tl;dr
IMO we don't need the trigger _here_ and I'm curious about your follow-up ; )
Do you mind sharing a link to your WIP branch?

ronso0: Apart from that this PR looks good to me btw.

daschuer (approved):
The code looks good. I have not tested yet. 
@ronso0 does that fit into our keyboard layout? 
In window explorer renaming is on F2

ronso0:
Yep, we only have Ctrl+R for Recording.

The idea is to have an alternative for F2 because it's mapped to rate_temp_up.

ronso0:
IIUC we now need to query the keyboard config via KeyboardEventFilter in order to show the correct shortcut in the sidebar context menu. I.e. show either F2 (keyboard disabled) or R/custom key (keyboard enabled).

This happens here for playlists
https://github.com/mixxxdj/mixxx/blob/67a41d9dcdb06b37f57be8e88978756140d05ff2/src/library/trackset/baseplaylistfeature.cpp#L77
and here for crates
https://github.com/mixxxdj/mixxx/blob/67a41d9dcdb06b37f57be8e88978756140d05ff2/src/library/trackset/crate/cratefeature.cpp#L71

An example is in [WMainMenubar](https://github.com/mixxxdj/mixxx/blob/67a41d9dcdb06b37f57be8e88978756140d05ff2/src/widget/wmainmenubar.cpp#L398-L401) or [LegacySkinParser](https://github.com/mixxxdj/mixxx/blob/67a41d9dcdb06b37f57be8e88978756140d05ff2/src/skin/legacy/legacyskinparser.cpp#L2404)

Updating the shortcut when toggling the keyboard is a bit more tricky.
Though, I think both is required for this PR : |

Please also check #13082 (not sure if it conflicts)

cr7pt0gr4ph7:
@ronso0 Thanks for bringing that to my attention! I'll have a look at the displaying the correct keyboard shortcuts, and will try a local merge with #13082. Based on a quick glance at the latter, this doesn't conflict with my  pull request as of now, but might interact with the reload logic.

ronso0:
Just FYI #13082 hasn't been reviewed, yet, so it's unclear if it'll be merged as it is now and when that'll happen of course.
I.e. I don't recommend to make this PR depend on #13082

ronso0:
I think, with current main, we need to pass a CoreServices (or KeyboardEventFilter) pointer to the library features that need to access the keyboard cfg, and maybe update the sidebar action shortcuts as required in every call of `...Feature::onRightClickChild()`.

Connecting the `WMainMenuBar::toggleKeyboardShortcuts(bool enabled)` signal to the library features is also possible but is certainly  even more work.

cr7pt0gr4ph7:
@ronso0 I've written a quick-and-dirty implementation based on your code, which I will share shortly (as soon as it has built on my machine).

I think the most clean way is to "register" the actions that have to be updated with the `KeyboardEventFilter` (possibly wrapped behind an abstract `KeyboardShortcutManager` interface). The library features already receive the `KeyboardEventFilter*` when `LibraryFeature::bindLibraryWidget(WLibrary*, KeyboardEventFilter*)` is called,so we can just register the actions with the `KeyboardEventFilter` then and there.

The code looks quite clean and natural this way. Additions like using a proxy to control lifetimes etc. are also easily possible, but probably not needed anyway. It should probably suffice to use `QPointer<QAction>` instead of `QAction*` for storing the references to the actions in `KeyboardEventFilter`, and to periodically remove actions that have been deleted.

cr7pt0gr4ph7:
@ronso0 Update: The code can be found here: https://github.com/cr7pt0gr4ph7/mixxx/compare/44a12a8b5eaa9fdedfdbf8e0bb7aabd876b8890a...wip-fix-star-rating-merge
I'll have a look on how to port this to the main branch, if it's forseeable your PR won't be merged anytime soon.

By the way, I am building Mixxx on my Linux machine inside a VS Code DevContainer, and its sloooooooooow as soon as multiple files are affected. Do you maybe have any tips for speeding the builds up? I am still waiting on the rebuild to finish...

cr7pt0gr4ph7:
I have looked into the issue of trying to display the correct shortcuts on context menu items etc. Here's my current conclusion:

* There are a lot of places in the UI where keyboard shortcuts are available/configurable (e.g. for `[Library] AutoDjAddTop`) but not displayed in the UI. It would probably be sensible to implement all of those places in the same way. The implementation for this will most likely look very similar to what @ronso0 has done at #13082. I'd be happy to help on that effort and/or provide a suitable pull request myself.

* This PR doesn't really break things any further, i.e. Mixxx already displays a shortcut for "Rename" that doesn't work when keyboard shortcuts are enabled & F2 is bound to something else.

* Implementing the shortcut display logic correctly & cleanly requires passing around parameters to a lot of places and would benefit from a separate discussion, while  those modifications are very much orthogonal to the contents of this pull request.

I would therefore argue for keeping this PR self-contained and getting it merged, and deferring the issue of displaying the shortcuts to a separate PR.

ronso0:
Thank you, that sounds reasonable.

ronso0:
> I am building Mixxx on my Linux machine inside a VS Code DevContainer, and its sloooooooooow as soon as multiple files are affected. Do you maybe have any tips for speeding the builds up?

I have no idea. I'm using Eclipse on Linux with cmake and ccache. With that only changed files (and those units including them) are rebuilt.

ronso0:
> I think the most clean way is to "register" the actions that have to be updated with the `KeyboardEventFilter`

Yup, like it's done for the menu bar actions in #13082

cr7pt0gr4ph7:
> Yup, like it's done for the menu bar actions in #13082 And that shoudl be straigh forward since we only need to store the KeyboardEventFilter pointer passed around via `bindLibraryWidget()`

See one of my [previous comments](https://github.com/mixxxdj/mixxx/pull/13148#issuecomment-2079400078) for a prototype of that :wink: A bit of care has to be taken to drop references to `QAction` instances that are created temporarily and then destroyed (e.g. when creating context menus) as well as on reloads, if and when we have such code. But besides that, the code is actually pretty simple, provided that #13082 is merged. I'll take a look at reviewing the latter.

cr7pt0gr4ph7:
@ronso0 Is there anything I can do to help getting this merged? I would love to have this in the official release builds, because it's annoying me in me day-to-day use of Mixxx quite a lot...

ronso0:
I'll try this soonish.

Regardless when this will be merged, you can run a personal branch with all PRs merged that you need.

cr7pt0gr4ph7:
> I'll try this soonish.

Thank you! I'm still figuring out how this project "ticks", and which way of contributing (small, frequent PRs vs. longer, larger ones) works best. Thanks for the feedback!

> Regardless when this will be merged, you can run a personal branch with all PRs merged that you need.

Had a bit of trouble setting it up, because I'm running my production version of Mixxx as a Flatpak application (on an atomic Fedora Silverblue/Bluefin desktop) and the [upstream for that](https://github.com/flathub/org.mixxx.Mixxx) only supports 2.4.x and Qt 5 right now... but I've got a [CI Flatpak build](https://github.com/cr7pt0gr4ph7/mixxx-flatpak) of my [personal `staging` branch](https://github.com/cr7pt0gr4ph7/mixxx/tree/staging) up and running now, if anybody else is interested in copying that setup.
I'm running that build as my main Mixxx instance now, so all contributions should be thoroughly tested on Linux.

I have a few more changesets lined up related to improving keyboard navigation in different areas, quality-of-life improvements for the Auto DJ feature, sound device management and additional hotcue buttons (and probably more stuff in the future), so feel free to tell me in which way you would prefer to receive those.

ronso0:
> I'm still figuring out how this project "ticks", and which way of contributing (small, frequent PRs vs. longer, larger ones) works best.

Yeah, there's no general receipt, developer time is.. fluctuating. Sometimes quick/small PRs get merged rather quickly, sometimes the sit for ages like big/hard-to-review PRs. And there may also been more relevant/pressing WIP that require attention. (like I need to delay this PR for now and try finish my own work first)

cr7pt0gr4ph7:
Ping @ronso0: Any chance of getting this merged before more merge conflicts creep up? I've been running this change in my own build for a while now, and so far it has worked like a charm.

AFAIK there are no remaining objections, as the issue of dynamically updating the shortcuts in the context menus [has been deferred](https://github.com/mixxxdj/mixxx/pull/13148#issuecomment-2079666164) to #13082 to avoid a half-baked attempt here.

cr7pt0gr4ph7:
**Update No. 2**: Build errors and issues related to #13335 should be fixed now.

ronso0:
Yes, IMO this is ready for merge.
The only aspect I feel a little uncomfortable with is that we now have two different edit hotkeys:
* F2: applies if shortcuts are disabled or F2 is mapped to edit in the user kbd mapping
* R: applies if kbd shortcuts are enabled and the new default mapping is loaded

What do you think about adding `R` to the keypress event switch in the sidebar and to the respective menu action so the new shortcut is discoverable?

cr7pt0gr4ph7:
> What do you think about adding `R` to the keypress event switch in the sidebar and to the respective menu action so the new shortcut is discoverable?

You mean making it so `R` works in addition to `F2` when when keyboard shortcuts are disabled? That actually sounds like a reasonable idea.

In addition to that, I'd still argue for making the edit key rebindable through the keyboard mapping when keyboard shortcuts are enabled. This leaves the issue of dynamically updating the displayed shortcuts, but as the code for a proper implementation of that would be 80% identical to #13082, I'd still defer that to that PR, or a similar PR extracted from it.

ronso0:
> You mean making it so R works in addition to F2 when when keyboard shortcuts are disabled?

Yup.

> In addition to that, I'd still argue for making the edit key rebindable through the keyboard mapping when keyboard shortcuts are enabled.

Sure, I'm just saying that it would be helpful / consistent to have the permanent R shortcut.

cr7pt0gr4ph7:
Can I do anything to move this PR forward / get it merged? Thanks! :)

daschuer:
I can confirm the issue. F2 is the edit shortcut if shortcuts are disabled. What is the reason not to use F2 for the edit feature when shortcuts are enabled?

ronso0:
Because that's mapped to "rate_perm_up" or similar for ages?

daschuer:
I think the underlying issue is now the move of the expected hotkey for Edit. 
Cumming form other applications it is F2. The same in Mixxx unless shortcuts are enabled.
Just noticing. If this is desired, we can merge. We can change that at any time later if it tuns out to be an issue.

src/widget/wlibrarysidebar.cpp
```diff
@@ -311,17 +321,9 @@ void WLibrarySidebar::keyPressEvent(QKeyEvent* event) {
...
-        if (!selIndex.isValid()) {
-            return;
-        }
-        if (isExpanded(selIndex)) {
-            // Root views / knots can not be renamed
-            return;
-        }
-        emit renameItem(selIndex);
+    case kRenameSidebarItemShortcutKey:              // F2
+    case kRenameSidebarItemAlternativeShortcutKey: { // R
```
daschuer, line 325:
```suggestion
    case kRenameSidebarItemAlternativeShortcutKey: { // R
```

daschuer, line 325:
```suggestion
    case kRenameSidebarItemShortcutKey: { // F2
```

daschuer: R already works without this line in case shortcuts are enabled.

src/library/trackset/baseplaylistfeature.cpp
```diff
@@ -74,7 +75,10 @@ void BasePlaylistFeature::initActions() {
             &BasePlaylistFeature::slotCreatePlaylist);
 
     m_pRenamePlaylistAction = new QAction(tr("Rename"), this);
-    m_pRenamePlaylistAction->setShortcut(kRenameSidebarItemShortcutKey);
+    m_pRenamePlaylistAction->setShortcuts(QList<QKeySequence>{
+            QKeySequence(kRenameSidebarItemShortcutKey),
+            QKeySequence(kRenameSidebarItemAlternativeShortcutKey)});
```
daschuer, line 80:
```suggestion
```
If we want to show the short cut, we need to change it form F2 to the value which is picked in kbd.cfg.

src/library/trackset/crate/cratefeature.cpp
```diff
@@ -68,7 +69,10 @@ void CrateFeature::initActions() {
             &CrateFeature::slotCreateCrate);
 
     m_pRenameCrateAction = make_parented<QAction>(tr("Rename"), this);
-    m_pRenameCrateAction->setShortcut(kRenameSidebarItemShortcutKey);
+    m_pRenameCrateAction->setShortcuts(QList<QKeySequence>{
+            QKeySequence(kRenameSidebarItemShortcutKey),
+            QKeySequence(kRenameSidebarItemAlternativeShortcutKey)});
```
daschuer, line 74:
```suggestion
```
same here. Either show the correct value or nothing.

src/util/defs.h
```diff
@@ -41,3 +41,4 @@ constexpr Qt::Key kHideRemoveShortcutKey = Qt::Key_Delete;
 #endif
 
 constexpr Qt::Key kRenameSidebarItemShortcutKey = Qt::Key_F2;
+constexpr Qt::Key kRenameSidebarItemAlternativeShortcutKey = Qt::Key_R;
```
daschuer, line 44:
```suggestion
```
This is not a constexpr it is set by kbd.cfg

daschuer (changes requested):
Thank you. Left some comments.

src/widget/wlibrarysidebar.cpp
cr7pt0gr4ph7: > I can confirm the issue. `F2` is the edit shortcut if shortcuts are disabled. What is the reason not to use `F2` for the edit feature when shortcuts are enabled?

As [previously discussed by you and @ronso](https://github.com/mixxxdj/mixxx/pull/13148#issuecomment-2078872028), `F1` through `F8` are bound to actions relating to pitch bending/bpm changes, so an alternative was needed.

The reasoning for why `r` was chosen is explained [in the PR description](https://github.com/mixxxdj/mixxx/pull/13148#issue-2255012006). I've been using it for a while now, and can confirm the usability turns out pretty well, especially when editing a number of tracks.

> For my understanding `R` is the adjustable shortcut alternative via kbd.cfg. We should not hard code it here. `R` already works without this line in case shortcuts are enabled.

Yes, you are correct. The newly introduced adjustable keyboard shortcut defaults to `R`.

Hardcoding `R` in addition to `F2` so `R` always works, even when keyboard shortcuts are disabled, was [an explicit request by @ronso0](https://github.com/mixxxdj/mixxx/pull/13148#issuecomment-2185330134). I can see the underlying reasoning, but don't feel strongly either way.

src/util/defs.h
cr7pt0gr4ph7: See [my previous response](https://github.com/mixxxdj/mixxx/pull/13148/files#r1938341361) for why `r` was hardcoded.

src/library/trackset/baseplaylistfeature.cpp
cr7pt0gr4ph7: Updated the code to display the correct shortcut in the display menu (though it does not dynamically react to toggling keyboard shortcuts on/off at runtime, as that requires the infrastructure implemented in #13082).

daschuer:
Can you squash the last commits into one? 
We try to keep the commit history build-able.

src/library/trackset/baseplaylistfeature.cpp
```diff
@@ -695,6 +695,19 @@ void BasePlaylistFeature::bindLibraryWidget(WLibrary* pLibraryWidget,
             &BasePlaylistFeature::htmlLinkClicked);
     m_pLibraryWidget = QPointer(pLibraryWidget);
     m_pLibraryWidget->registerView(m_rootViewName, pEdit);
+
+    // Update shortcuts displayed in the context menu
+    if (pKeyboard->getKeyboardConfig()->exists(ConfigKey("[Library]", "EditItem"))) {
```
daschuer, line 700: This doers not work, because we can have the situation that there is no "Edit item"  configured. 

You need either to show the correct active short cuts or nothing. Showing a not working short cut is a bug IMHO

src/widget/wlibrarysidebar.cpp
daschuer: I don't see this request in the linked post. 
At least it is wrong to maintain a hard coded R here, while it can be actually any key on users choice. 
We consider the kind.cfg files a user configuration.

daschuer (changes requested):

ronso0:
> At least it is wrong to maintain a hard coded R here, while it can be actually any key on users choice.

I think it would be good to have static shortcuts, ie. at least one shortcut that works regardless (user) shortcuts being enabled.

daschuer:
I can confirm that. I am also not particularly happy about the changing meaning of F2. 

A straight forward solution would be to not use F2 in kbd.cfg. Then everything is consistent. F2 is anyway cumbersome for any perfomace hot key, because at least in case of Laptop Keyboard the key kan be permanently mapped to FN (Volume down) in my case. 

If we want to srick with "R" which is clearly in the region of other keyboard mappings, and not a system wide key I will not oppose. 
However the system schall behave consistent in case a user Remaps "R" to diffrent feature.

They may for example map F2 to "[Library], EditItem" to make it alwas work the same. In this case a static mapping to "R" in potions of the code is missleading.

cr7pt0gr4ph7:
Given that most arguments have already been made, I propose the following - provided you (@ronso0 and @daschuer) both agree:
- **Displaying shortcuts:** Display neither `F2` nor `R`, and defer dynamically displaying the correct shortcuts to #13082 (which implements a wholistic approach).

- **`EditItem = R` vs. `EditItem = F2`:**
  - Only hardcode `F2` to keep the current behavior for `"keyboard shortcuts = off"` and honor the users assumption that `"normal" keys (A..Z, 0..9 etc.) by themselves don't do anything when "keyboard shortcuts = off"`.
  - Leave `F1 - F8` bound to their current functions, as some users (me included) heavily use those.
  - Bind `EditItem` to `R` by default for `"keyboard shortcuts = on"`, as it's easily reachable and, based on my personal experience, actually pretty ergnonomic to use, even when mass-editing tracks.

@daschuer
> Can you squash the last commits into one?
> We try to keep the commit history build-able.

Sorry about the noise... I'll clean up the commit history as soon as a decision has been made regarding the points above, as that might lead to the removal of some other commits.

ronso0:
Yup, I agree as well.

cr7pt0gr4ph7:
Okay, I adapted the PR as proposed & agreed above. It seems everything has come full circle, as this is exactly the original contents of the PR (with the only addition being the removal of the shortcuts from the context menus) :D

cr7pt0gr4ph7:
**re: The conflict between hardcoded `switch` statements and the keymap** _(and a hypothetical suggestion for these might be avoided in the future)_:

Instead of handling `shortcuts = on` and `shortcuts = off` through fully separate (and partially duplicated) code paths, it should be possible to route everything through `KeyboardEventFilter` instead, and just swap out the active keymap between:
- the user-defined keymap for `shortcuts = on`
- a restricted, non-editable keymap for `shortcuts = off`

I can think of at least three distinct advantages this would have:
- One degree of freedom / one whole set of code paths less to be tested
- Once #13082 (or a reduced version) is merged, always displaying the correct shortcuts for the correct state becomes a no-brainer
- `diff`ing the "full" vs. the "reduced" keymap becomes easily possible

One could even do away completely with the `shortcuts enabled` setting, and replace it with a combobox that provides a set of different built-in keymaps.

_(This is obviously out-of-scope for this PR, but hopefully useful as a jumping-off point for future discussions)_

daschuer:
Keyboard support is definitely a sore point in Mixxx. How nice would it be for example to pug an old keyboard to a laptop and use it as poor-mans DJ controller ... just an idea. 

One main issue of using the regular keyboard driver are different layouts and dead keys. A solution would be to treat the keyboard like any other controller and react on the key codes and supports the mapping wizard. 

We had a GSoC project that never made it into upstream. https://github.com/mixxxdj/mixxx/pull/1746  @ronso0 has adopted some ideas into https://github.com/mixxxdj/mixxx/pull/13082

daschuer (approved):

Issue #12641, opened 2024-01-24
New GLSL waveforms are less detailed than legacy waveforms


fwcd:
### Bug Description

Moving https://github.com/mixxxdj/mixxx/issues/12448#issuecomment-1874799386 to a new issue here. Comparing the current GLSL waveforms:

![Screenshot 2024-01-24 at 18 04 29](https://github.com/mixxxdj/mixxx/assets/30873659/ffe40b73-7bf5-45dc-8766-bc099f6411ad)

to the legacy GLSL waveforms:

![Screenshot 2024-01-24 at 18 04 38](https://github.com/mixxxdj/mixxx/assets/30873659/b2c58614-d6b7-4275-947c-9aea598151d9)

...there is a fair bit of detail missing in the new GLSL waveforms, especially the bass-heavy parts display more structure. 

### Version

2.5-alpha-207-g7ab184441d (HEAD)

### OS

macOS 14.3

mxmilkiib:
Looks mono, not stereo

fwcd:
Here's another example of GLSL

![Screenshot 2024-02-01 at 16 29 36](https://github.com/mixxxdj/mixxx/assets/30873659/8337d2c1-f94c-47db-8362-f76bc8a4c28a)

vs. GLSL (legacy)

![Screenshot 2024-02-01 at 16 29 44](https://github.com/mixxxdj/mixxx/assets/30873659/0fe76e1a-5e33-4fa1-baa7-78cdadc7aa7b)

I've found myself switching back to the legacy waveforms for live use, the bass peaks are just much more recognizable.

ronso0:
Related https://github.com/mixxxdj/mixxx/issues/11833#issuecomment-1905165836

How shall we proceed here?
Regardless the huge performance improvement, is this a regression? Is it severe enough to be fixed before the release?

daschuer:
Since the legacy waformsvate still there, this should not block the release.

fwcd:
Both screenshots were taken with the (then) most recent 2.5/main builds

acolombier:
It looks like this impacting all waveform with various degree of difference. Here is a video which might help to illustrate better these difference. 

https://github.com/mixxxdj/mixxx/assets/7086688/45c2a2b0-97c4-4504-aa9a-0cbcc2c45e12

ywwg:
> degree of difference

have you tested with the texture PR?

acolombier:
[...]

PR #13151, opened 2024-04-21
allshader RGB,Filtered and Stacked using textures for waveform data

m0dB:
This adds the waveforms for RGB, Filtered and Stacked that use textures to transfer the waveform data to the shader.

This uses the code from the legacy GLSL waveforms, but in combination with the rest of the allshader waveform layers.

(There is code duplication, but since we are planning on removing the legacy ones, I don't care about this)

I added an indicator "waveform texture data" (in addition to the existing indicators GLSL and legacy) to show in the waveform combobox which waveforms use textures to transfer the waveform data to the shader.


JoergAtGithub:
Filtered GLSL waveform appears in RGB colors now. Otherwise it works as intended.

m0dB:
> Filtered GLSL waveform appears in RGB colors now.

Yes, but that is not due to this PR, it was already the case.

ywwg:
this looks fantastic to me -- a good step towards solving https://github.com/mixxxdj/mixxx/issues/12641 though we still have the curve-at-analyze-time problem to work on

src/waveform/renderers/allshader/waveformrenderertextured.cpp
```diff
@@ -0,0 +1,515 @@
...
+    float lowGain(1.0), midGain(1.0), highGain(1.0), allGain(1.0);
+    getGains(&allGain, true, &lowGain, &midGain, &highGain);
+
+    const auto firstVisualIndex = static_cast<GLfloat>(
+            m_waveformRenderer->getFirstDisplayedPosition() * trackSamples /
+            audioVisualRatio / 2.0);
+    const auto lastVisualIndex = static_cast<GLfloat>(
+            m_waveformRenderer->getLastDisplayedPosition() * trackSamples / audioVisualRatio / 2.0);
+
+    // const int firstIndex = int(firstVisualIndex+0.5);
```
ywwg, line 313: remove comment cruft?

ywwg (approved):

ywwg:
> curve-at-analyze-time problem

oh sounds like this may be already solved

acolombier:
It feels like this PR is going against the effort to reduce the number of "same" waveform type: from a user perspective, they don't care whether "RGB" is achieved with texture of GLSL and adding another type (even if we eventually retire the legacy one) will still keep the confusion. Could we replace the other `allshader` instead of adding 3 more type to the existing 13?

m0dB:
I think this would better be discussed in Zulip. I personally prefer the non-textured ones over the textured ones. I don’t think there is a problem with having multiple options, it’s about how to present those options in a useful way.

ywwg:
I posted a comparison in Zulip showing why I prefer textured.  I do think we can reduce some of the options, but I would *strongly* prefer to find a way to keep the extra data that the textured approach shows.  (I will maintain my own custom fork if I have to, it makes that much of a difference in my ability to perform)

src/waveform/renderers/allshader/waveformrenderertextured.cpp
m0dB: I just copied this as-is from the legacy code.

m0dB:
I have deprecated the fixed function pipeline OpenGL widgets and the waveform data texture based widgets that combined with QPainter (replaced with waveform data texture + allshader layers). I put the code in folders waveform/widgets/deprecated and waveform/renderers/deprecated and don't list them in the UI, leaving the preferences dropdown like this:

Screenshot 2024-05-12 at 01 58 34

@ywwg @acolombier maybe we can merge this first, and then merge Antoine's software fallback waveforms, and finally do the waveform preferences reorganisation?

acolombier:
Ideally, I would prefer merging #13220 first so the AllShaderFilter can implement the Option enum directly, and reduce the number of independent waveform widget

ywwg:
I agree with @m0dB to merge this first and clean up the preferences later.  Either way there will be a little pain but it feels right to get the backend stuff (waveform algorithms) figured out before cleaning up the UX.

ywwg:
> deprecated

are we planning to delete them in the future? (Are these the legacy waveforms?)  If we are going to remove them, I don't see a big point in moving them to a "deprecated" folder in the meantime. It will make the git history messier, and calling things "legacy" (like the skin parser) and "deprecated" is tempting fate that they will never actually be removed :)

m0dB:
The mean idea behind leaving them for now was for the sake of comparison. But I guess one might as well run 2.4 and 2.5 side by side for that.

I am okay with removing them, but I didn’t want to make that decision.

m0dB:
@acolombier the advantage of merging this first is that it removes all the renderers and widgets that will not be used anymore, so that might make things a bit easier to comprehend.

ywwg:
I am ok with keeping them for now, but I don't think it's worth the trouble to move the files around unless that does something I haven't realized.  If we mark them in the UI as deprecated in the next version (2.6, since 2.5 is already cut), and then delete them in 2.7.

acolombier:
I guess it doesn't really matter. As I mentioned on Zulip, integrating these changes in #13220 will require removing a significant amount of changes added by this PR so it felt more natural the other way around. 

> Either way there will be a little pain but it feels right to get the backend stuff (waveform algorithms) figured out before cleaning up the UX.

The problem is #13220 makes significant amount of changes on the backend itself (factory, allshader renderer) 

FWIW, I'm keen to get either merge ASAP so we can move to the next step, as I will soon not have much time to spend on Mixxx anymore.

m0dB:
Do you think anything is missing from this PR, or can we merge it?

ywwg:
> Do you think anything is missing from this PR, or can we merge it?

we're going to talk about this at the meeting in a few minutes :)

ywwg (approved):
I've been running this patch and it works great.  LGTM!

src/waveform/renderers/waveformwidgetrenderer.h
```diff
@@ -170,6 +170,13 @@ class WaveformWidgetRenderer {
         return renderer;
     }
 
+    template<class T_Renderer, typename T_Arg>
+    inline T_Renderer* addRenderer(T_Arg arg) {
+        T_Renderer* renderer = new T_Renderer(this, arg);
+        m_rendererStack.push_back(renderer);
+        return renderer;
+    }
+
```
Swiftb0y, line 179: whats the difference between this template and the one above? This shouldn't even compile IMO because its ambiguous which should be chose in the one parameter case.

src/waveform/renderers/allshader/waveformrenderertextured.h
```diff
@@ -0,0 +1,66 @@
...
+    TrackPointer m_loadedTrack;
+    int m_textureRenderedWaveformCompletion;
+
+    // Frame buffer for two pass rendering.
+    std::unique_ptr<QOpenGLFramebufferObject> m_framebuffer;
+
+    // shaders
+    bool m_shadersValid;
+    Type m_type;
+    const QString m_pFragShader;
```
Swiftb0y, line 64: not a pointer
```suggestion
    const QString m_fragShader;
```

src/waveform/renderers/allshader/waveformrenderertextured.cpp
```diff
@@ -0,0 +1,515 @@
...
+    case Type::Filtered:
+        return QLatin1String(":/shaders/filteredsignal.frag");
+    case Type::RGB:
+        return QLatin1String(":/shaders/rgbsignal.frag");
+    case Type::Stacked:
+        return QLatin1String(":/shaders/stackedsignal.frag");
+    default:
+        break;
+    }
+    assert(false);
```
Swiftb0y, line 26: avoid segfault during set. Use QString if ur returning a QString, otherwise it'll unnecessary do a utf8 to 16 conversion and allocation.
```suggestion
    DEBUG_ASSERT(false);
    return QStringLiteral(":/shaders/rgbsignal.frag");
```

src/waveform/renderers/allshader/waveformrenderertextured.cpp
```diff
@@ -0,0 +1,515 @@
...
+    int dataSize = 0;
+    const WaveformData* data = nullptr;
+
+    ConstWaveformPointer pWaveform = m_waveformRenderer->getWaveform();
+    if (pWaveform) {
+        dataSize = pWaveform->getDataSize();
+        if (dataSize > 1) {
+            data = pWaveform->data();
+        }
+    }
```
Swiftb0y, line 102: declare variables at point of use, prefer early returns when it doesn't make sense to continue (this is not the linux kernel ;) )
```suggestion
    ConstWaveformPointer pWaveform = m_waveformRenderer->getWaveform();
    if (!pWaveform) {
        return false;
    }
    int dataSize = pWaveform->getDataSize();
    // is this condition correct? should it actually be just be dataSize == 0?
    if (dataSize <= 1) {
        return false;
    }
    const WaveformData* data = pWaveform->data();
```

src/waveform/renderers/allshader/waveformrenderertextured.cpp
```diff
@@ -0,0 +1,515 @@
...
+        glEnd();
+    }
+    glEndList();
+}
+
+void WaveformRendererTextured::createFrameBuffers() {
+    const float devicePixelRatio = m_waveformRenderer->getDevicePixelRatio();
+    // We create a frame buffer that is 4x the size of the renderer itself to
+    // "oversample" the texture relative to the surface we're drawing on.
+    constexpr int oversamplingFactor = 4;
```
Swiftb0y, line 192: how is that different from multisampling?

m0dB: I don't know, I just took the legacy GLSL renderer code as-is and adapted the minimum so it could be used together with the rest of the allshader layers.

src/waveform/renderers/allshader/waveformrenderertextured.cpp
m0dB: This is not my code.

m0dB:
@Swiftb0y I see you are commenting on the waveformrenderertextured.cpp code. This is not my code. I just took the legacy GLSL renderer code and made the minimum changes in order to use it in combination with the allshader renderer layers.

I am all for cleaning up this code, and I am willing to do so in future PRs, but I don't think that just moving existing code should come with the obligation to clean it up. The improvement here is using this GLSL based renderer in combination with all the other GLSL based allshader layers instead of in combination with QPainter based layers.

src/waveform/renderers/allshader/waveformrenderertextured.h
m0dB: Not my code.

src/waveform/renderers/waveformwidgetrenderer.h
m0dB: You are absolutely right, I somehow missed that the above allowed for arguments.

Swiftb0y:
> @Swiftb0y I see you are commenting on the waveformrenderertextured.cpp code. This is not my code. 

I'm sorry, that wasn't clear so I was just treating it as new code. Disregard my comments for now then, but please leave them open so they're easy to revisit later.

Swiftb0y (approved):
LGTM.

Swiftb0y:
Merging now to keep the velocity. Thank you all for your patience. I'm looking forward to the future cleanup PRs.

PR #13158, opened 2024-04-22
[Test] NowPlaying with settings, problem: no usersettingspointer in MixxxMainWindow

Eve00000:
For Ronso to check 
Mixxx NowPlaying with settings,
my problem is : I can't get the usersettingspointer in MixxxMainWindow
I tried adding m_pConfig to .h etc 
does not work, help please


src/mixxxmainwindow.cpp
```diff
@@ -691,11 +691,26 @@ void MixxxMainWindow::slotUpdateWindowTitle(TrackPointer pTrack) {
         QString trackInfo = pTrack->getInfo();
         if (!trackInfo.isEmpty()) {
             appTitle = QString("%1 | %2").arg(trackInfo, appTitle);
+            //          writing the artist & title of the playing track not only to the windowtitle but also to a file
+            //          location and name for nowplayingfile
+			if (m_pConfig->getValueString(ConfigKey("[Controls]", "CreateNowPlayingFile")) != "0") {
```
ronso0, line 696: `m_pCoreServices->getSettings()` like you do below.
I suggest to get the settings pointer only once in this function.

ronso0:
FYI you can simply open a PR in your Mixxx fork.
(assuming you leave the `main` branch alone (or have some other branch cloned from mixxx/main) and work in feature branches).

src/mixxxmainwindow.cpp
Eve00000: Thank you,
I found it: 
if (m_pCoreServices->getSettings()->getValueString(ConfigKey("[Controls]", "CreateNowPlayingFile")) != "0") 
I thought the coreservices were settings like commandarguments, environment etc .
My apologies.

What do you mean with 'only once'? Getting the configKey and the settingspath in one line?
Or using a var to put the coreservice-settings in and then get the configkey and settingspath out of that var?
Can you explain a little more please.

ronso0: Yes, I mean store the `UserSettingsPointer` that you get from `m_pCoreServices->getSettings()` in a variable.
That's not required though, just reduces the code.

```
UserSettingsPointer pConfig = m_pCoreServices->getSettings();
// then use it to get/set values
QString xyz = pConfig->getValueString(ConfigKey("[Controls]", "CreateNowPlayingFile"));
```

Eve00000: Thank you, I've learned something. I will use it.

But ... euh .... I counted the occurences of m_pCoreServices in MixxxMainWindow : 68 times

I wonder: the m_pCoreServices are stored in the memory, when creating an extra userpointer that refers to the settings in the memory doesn't it fill the memory with extra vars? What is the fastest: directly read the settings or first the var ...
I know the on machines nowadays it won't make a lot of difference but in the days with only 640kb basic memorywe had to be very careful when storing extra vars in the mem. And I've read that some people want to execute Mixxx on a mini-pc like RPI (funny thing, I like it too).
I don't know if the actual cpu/mem speed interest you but what would be the fastest? Although c (and c++) are languages that have a  very fast interaction with the machine and the complete code gets loaded into the memory as well, it's a lot of memory-swapping.

ronso0: > I wonder: the m_pCoreServices are stored in the memory, when creating an extra userpointer that refers to the settings in the memory doesn't it fill the memory with extra vars? What is the fastest: directly read the settings or first the var ...

No worries, this is just a pointer, i.e. a 'link' to the actual object.

People **are** running Mixxx on the RPi and probably other SoC like PandaBoard etc.

acolombier:
Hi @Eve00000 , is this PR superseding #13157 and #12993? If yes, can we close them and keep further work within this branch to limit PR proliferation?

ronso0:
Yes, let's close it. This was opened (accidentally against mixxxdj/mixxx) just for checking the config pointer issue.
@Eve00000 as I wrote earlier: if you want to open PRs only for demonstration purpose or for discussing some aspect but don't have intention to get it merged -> please open PRs against your Mixxx fork.
This will help us clean up the PR list and lower the noise (notifications etc.).
Thank you!

Eve00000:
I've learned a lot meanwhile, also git handling ;-)

PR #13165, opened 2024-04-22
Remove autdated ASSERT fixing #13164

### LINKED ISSUE #13164: DEBUG_ASSERT in basetracktablemodel.cpp on header splitter click
### Bug Description

```
critical [Main] DEBUG ASSERT: "siblingColumn != index.column()" in function virtual QVariant BaseTrackTableModel::rawSiblingValue(const QModelIndex&, ColumnCache::Column) const at ./src/library/basetracktablemodel.cpp:492
```

### Version

2.4.0

### OS

Ubuntu Focal

daschuer:
fixes #13164

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -462,18 +462,6 @@ QVariant BaseTrackTableModel::data(
     return roleValue(index, rawValue(index), role);
 }
 
-QVariant BaseTrackTableModel::rawValue(
-        const QModelIndex& index) const {
-    VERIFY_OR_DEBUG_ASSERT(index.isValid()) {
-        return QVariant();
-    }
-    const auto field = mapColumn(index.column());
-    if (field == ColumnCache::COLUMN_LIBRARYTABLE_INVALID) {
-        return QVariant();
-    }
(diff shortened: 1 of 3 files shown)


daschuer:
cleaning up the 2.4.1 milestone from this hard to reproduce issue

src/library/basetracktablemodel.cpp
```diff
@@ -489,10 +489,6 @@ QVariant BaseTrackTableModel::rawSiblingValue(
         // FIXME: This should never happen but it does. But why??
         return QVariant();
     }
-    VERIFY_OR_DEBUG_ASSERT(siblingColumn != index.column()) {
```
acolombier, line 492: Should this be replaced with a `if` instead of removing the condition check entirely like above? Reading the comment, since it seems the early return was here to prevent infinite recursion. Or is that comment incorrect/out of date?

daschuer:
I have made rawValue() pure virtual to show that we have no infinitive recursive call issue.

Swiftb0y (approved):
LGTM :shrug:

daschuer:
@Swiftb0y merge?

Swiftb0y:
I guess :shrug:

PR #4528, opened 2021-11-17
Macros Backend

xeruf:
This is forked off from https://github.com/mixxxdj/mixxx/pull/2989 with the GUI parts removed, to be merged this week :)
@daschuer @Holzhaus @Be-ing

PR #4529, opened 2021-11-18
Macros Skin Integration

xeruf:
A preliminary Skin integration for the Mixxx Macros feature rebased on #4528, drawn out of the original #2989.

PR #13172, opened 2024-04-23
Macros Backend respin

JoergAtGithub:
This is https://github.com/mixxxdj/mixxx/pull/4528 with fixed CI and solved merge conflicts.

It implements the following COs without GUI: https://github.com/mixxxdj/mixxx/wiki/Macros

This PR is the first step to implement the Macro feature. The Mixxx Macros feature is designed functional compatible to Serato Flip. If we later add a Serato Flip importer in a Follow-Up PR, Mixxx DJs can use all files prepared with Flips provided by the common DJ pools.
A good (maybe a bit flippy ;-) ) introduction into the use cases of Serato Flip can be found here: https://www.youtube.com/watch?v=sHRWfdxkI7E&t=186s


src/engine/enginebuffer.cpp
```diff
@@ -227,6 +227,18 @@ EngineBuffer::EngineBuffer(const QString& group,
     m_pCueControl = new CueControl(group, pConfig);
     addControl(m_pCueControl);
 
+    for (int i = 1; i <= kMacrosPerChannel; ++i) {
+        auto* pMacroControl = new MacroControl(group, pConfig, i);
```
acolombier, line 231: Looks like this might currently be leaked. Perhaps should be replaced by a unique_ptr

acolombier (commented):
I tried the feature but could only get a single hotcue to work. It could be that I'm not using properly tho. Do you know if there is any wiki or doc on how to use this feature beside the GSoC wiki page?

JoergAtGithub:
> I tried the feature but could only get a single hotcue to work. It could be that I'm not using properly tho. Do you know if there is any wiki or doc on how to use this feature beside the GSoC wiki page?

I think there are only the three GSoC pages in our wiki, and the original PRs of cause. One aim of this feature was to achieve compatibility to Serato Flip, to allow data import from Serato. Therefore I guess, that the Serato Flip feature is similar to use: https://support.serato.com/hc/en-us/articles/203157550-Serato-Flip-User-Guide

src/track/macroaction.h
```diff
@@ -0,0 +1,47 @@
...
+/// A MacroAction is the building block of a Macro.
+/// It contains a position as well as the action to be taken at that position.
+///
+/// Note that currently only jumps to a target position are available,
+/// but that is subject to change.
+class MacroAction {
+  public:
+    enum class Type : uint32_t {
+        Jump = 0
+    };
```
acolombier, line 17: Should this be registered with ` Q_ENUM` to ensure support within Qt metatype?

src/track/macroaction.h
```diff
@@ -0,0 +1,47 @@
...
+    mixxx::audio::FramePos getTargetPosition() const {
+        return target;
+    }
+
+    proto::Macro_Action serialize() const;
+
+  private:
+    mixxx::audio::FramePos source;
+    mixxx::audio::FramePos target;
+    Type type;
```
acolombier, line 39:
```suggestion
    mixxx::audio::FramePos m_source;
    mixxx::audio::FramePos m_target;
    Type m_type;
```

src/track/track.cpp
```diff
@@ -1309,6 +1309,42 @@ void Track::importPendingCueInfosMarkDirtyAndUnlock(
...
+}
+
+bool Track::isDirty() {
+    const auto locked = lockMutex(&m_qMutex);
+    if (m_bDirty) {
+        return true;
+    }
+    // FIXME(xeruf) this should be incorporated in the dirty state of the track itself
+    // but that is non-trivial, since, unlike other track properties, Macros are mutable
+    for (const MacroPointer& pMacro : qAsConst(m_macros)) {
```
acolombier, line 1340:
```suggestion
    for (const MacroPointer& pMacro : std::as_const(m_macros)) {
```

src/util/db/dbid.h
```diff
@@ -22,21 +22,27 @@
...
 // only needed for type-safety.
 class DbId {
-  protected:
   public:
     constexpr DbId()
             : m_value(kInvalidValue) {
     }
+    explicit DbId(const int value)
+            : m_value(value) {
+        DEBUG_ASSERT(isValid() || (kInvalidValue == m_value));
```
acolombier, line 31: This assert looks weird and feels like it will always pass. As I understand, it should fail for `m_value < -1`. Maybe we can put that explicitly to help with readability?

acolombier (dismissed):
Thanks for bringing this feature back to life. Appreciate you didn't write this code but I think there quite a few things that don't line up with our coding guidelines. 
Haven't had a chance to test it yet!

src/engine/controls/macrocontrol.cpp
```diff
@@ -0,0 +1,339 @@
...
+    // "recording:" << isRecording();
+    VERIFY_OR_DEBUG_ASSERT(isRecording()) {
+        return false;
+    }
+    bool actionsRecorded = false;
+    while (MacroAction* action = m_recordedActions.front()) {
+        m_pMacro->addAction(*action);
+        m_recordedActions.pop();
+        actionsRecorded = true;
+    }
```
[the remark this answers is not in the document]
JoergAtGithub: I think this is used in the correct way here, and pop should be a very fast operation: https://github.com/rigtorp/SPSCQueue#benchmarks

src/engine/controls/macrocontrol.cpp
```diff
@@ -0,0 +1,339 @@
...
+    if (getStatus() == Status::Armed) {
+        setStatus(Status::Recording);
+    }
+    if (getStatus() != Status::Recording) {
+        return;
+    }
+    // Account for quantization so the jump stays in-phase but snaps exactly to the original target
+    VERIFY_OR_DEBUG_ASSERT(m_recordedActions.try_emplace(
+            frameInfo().currentPosition + (queuedTarget - position),
+            queuedTarget)){};
```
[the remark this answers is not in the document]
JoergAtGithub: This evaluates the condition somehow different and let the build fail.

src/track/macroaction.h
JoergAtGithub: This would require a lot of overhead, as this class is not QObject based and there is no moc_ file generation needed otherwise. I think this should therefore not done, until somebody has a real need for it.

Holzhaus:
Awesome that you picked this up. I'll try to have a look this weekend (but as always, I trust your judgment so don't worry about waiting for my okay 😀)

JoergAtGithub:
> I'll try to have a look this weekend (but as always, I trust your judgment so don't worry about waiting for my okay 😀)

I just fixed code issues - a test of someone who knows the functionality better is highly appreciated!

Holzhaus:
For reference, here's a small demo how it works in serato: https://youtu.be/WWTdAACN7v0

ronso0:
FYI I wanted to create a basic GUI for testing and thought I could cherry-pick (and fix) commits from #4529, but that's a PITA since WTrackProperty has become complex with the inline editors.
Anyone attempting to create a GUI: let's create new widgets for it, eg. WMacroLabel etc.

JoergAtGithub:
The GUI were just a few buttons - very simple. Something like this would be enough:
grafik

ronso0:
I thought macros are per track? (= per deck, not one macro manager.

JoergAtGithub:
The controls are global, you click the record button and it records the deck were the DJ does the next action.

I also thought, that this might be difficult to understand and had the idea of a colored border around the deck - e.g. when armed, all decks get a orange border, and once a deck is selected, only this one get a red border.

ronso0:
But (as I understood it back then) macros are per track, so I can start a macro on deck1, load a track in deck2 and start its macro#5 to do prepared mashups and stuff.
Thats only possible if we have a per-deck GUI.

I don't want to move this off-topic, just saying...

JoergAtGithub:
Yes, lets discuss the GUI on Zulip. The COs in this PRs should be flexible enough to allow per deck controls as well.

daschuer:
Great you picked this up. 

What is the difference between the Status::Playing and Status::Enabled? I have not tested it ... 

Am I right that Macros do not necessarily include transport actions? Will the track start playing if pressing macro play? This depends probably on the macro itself?

JoergAtGithub:
> What is the difference between the Status::Playing and Status::Enabled? I have not tested it ...

See https://github.com/mixxxdj/mixxx/wiki/Macros
This is the unchanged functionality of the GSoC project, which was jointly agreed on. This PR only addresses the open code style and CI issues for which the original GSoC PR was not merged.

Holzhaus:
> What is the difference between the Status::Playing and Status::Enabled? I have not tested it ...

There is no `Status::Enabled`.

https://github.com/mixxxdj/mixxx/blob/3ea8edddb8ec3f613bc69c88662ca074731080f8/src/engine/controls/macrocontrol.h#L24-L31

src/engine/controls/enginecontrol.h
```diff
@@ -48,17 +48,17 @@ class EngineControl : public QObject {
...
 
-    // Called by EngineBuffer::process every latency period. See the above
-    // comments for information about guarantees that hold during this call. An
-    // EngineControl can perform any upkeep operations that are necessary during
-    // this call.
+    /// Called by EngineBuffer::process every latency period.
+    /// See the above comments for information about guarantees that hold during this call.
+    /// An EngineControl can perform any upkeep operations necessary here.
+    /// @param dRate current playback rate in audio frames per second
     virtual void process(const double dRate,
```
Holzhaus, line 55:
```suggestion
    virtual void process(const mixxx::audio::SampleRate rate,
```
(also in the implementation)

src/engine/controls/enginecontrol.h
```diff
@@ -48,17 +48,17 @@ class EngineControl : public QObject {
...
             mixxx::audio::FramePos currentPosition,
             const int iBufferSize);
```
Holzhaus, line 57: Not sure if there is more fitting type (did we already switch to `size_t` elsewhere?). Anyway, at least we don't use the type prefixes anymore, so this should be named `bufferSize`.

src/engine/controls/macrocontrol.cpp
```diff
@@ -0,0 +1,333 @@
...
+ConfigKey MacroControl::getConfigKey(const QString& name) {
+    return ConfigKey(m_group, QStringLiteral("macro_%1_%2").arg(m_slot).arg(name));
+}
+
+MacroControl::MacroControl(const QString& group, UserSettingsPointer pConfig, int slot)
+        : EngineControl(group, pConfig),
+          m_slot(slot),
+          m_queuedJumpTarget(-1),
+          m_recordedActions(kRecordingQueueSize),
+          m_iNextAction(0),
```
Holzhaus, line 26:
```suggestion
          m_nextActionIndex(0),
```
(also in the corresponding code)

src/engine/controls/macrocontrol.cpp
```diff
@@ -0,0 +1,333 @@
...
+void MacroControl::play() {
+    DEBUG_ASSERT(m_pMacro && !m_pMacro->getActions().isEmpty() && !isRecording());
+    m_iNextAction = m_pMacro->size() > 1 ? 1 : 0;
+    setStatus(Status::Playing);
+    m_pMacro->setState(Macro::StateFlag::Enabled);
+}
+
+void MacroControl::stop() {
+    DEBUG_ASSERT(m_pMacro && !m_pMacro->getActions().isEmpty() && !isRecording());
+    m_iNextAction = INT_MAX;
```
Holzhaus, line 189: This looks like a hack, should we use `std::optional` here?

src/engine/controls/macrocontrol.h
```diff
@@ -0,0 +1,82 @@
...
+    bool isRecording() const;
+
+    enum class Status : int {
+        NoTrack = -1,
+        Empty = 0,
+        Armed = 1,
+        Recording = 2,
+        Recorded = 3,
+        Playing = 4,
+    };
```
Holzhaus, line 31:
```suggestion
    enum class Status : int {
        /// No track is loaded.
        NoTrack = -1,
        /// There is no recoded macro present.
        Empty = 0,
        /// Recording is armed. The next supported action (e.g., a hotcue jump will start the recording).
        Armed = 1,
        /// Recording has been started and supported actions will be part of the resulting macro.
        Recording = 2,
        /// The is a recorded macro present, but it's not enabled for playback.
        Recorded = 3,
        /// The macro is currently enabled for playback.
        Playing = 4,
    };
```

We could rename `Armed` to `RecordingArmed` to make this clearer?

src/engine/controls/enginecontrol.h
JoergAtGithub: `mixxx::audio::SampleRate` is an `unsigned int`, but this rate can become negative, if we play reverse.

daschuer:
> There is no Status::Enabled.

Oh I see, I got confused with the `StateFlag::Enabled`

I have still not fully understand how that interacts. 

Is the a description of the state transitions somewhere?
Maybe there is a chance to simplify it a bit.

For my understanding a macro is comparable to Auto DJ, but only for a single track. Is that correct?

It is possible to record a jump between two cue points. 
After recording the user has to press "play" on the macro button for the deck to enable the macro. When the deck is then played in terms of transport, for example by Auto DJ, the jump is automatically performed.

Is that correct?

JoergAtGithub:
I don't see a similarty to AutoDJ, but yes, it's per track.
And please let not put the feature in question at all. It's all defined and lengthy discussed in GSoC. This is just to fix the final code style and and unit test issues, which prevented merging of the original PR.

src/engine/controls/enginecontrol.h
```diff
@@ -48,17 +48,17 @@ class EngineControl : public QObject {
...
     ~EngineControl() override;
 
-    // Called by EngineBuffer::process every latency period. See the above
-    // comments for information about guarantees that hold during this call. An
-    // EngineControl can perform any upkeep operations that are necessary during
-    // this call.
+    /// Called by EngineBuffer::process every latency period.
+    /// See the above comments for information about guarantees that hold during this call.
+    /// An EngineControl can perform any upkeep operations necessary here.
+    /// @param dRate current playback rate in audio frames per second
```
daschuer, line 54: This is a relative value. It indicates how many samples of the track are consumed compared to the output range of Mixxx. The engine can make use of some renamings to avoid this confusion.

JoergAtGithub: What about:

- playRate
- playbackRate
- relativePlayrate

daschuer: In the CO interface we have "rate" as the rate slider, a scalable relative adjustment with 0 at unity. And 
https://github.com/mixxxdj/mixxx/blob/d61749c79c2ed2fd8e85cc86add5e3f58b600925/src/engine/controls/ratecontrol.cpp#L45

"rate_ratio" the fixed ratio relative to the track original rate with 1 at unity. 

Here it is the same relative to the engine rate.

einigeRateRatio  or such. However just "rate" is also ok. I was more concerned to the namings in EngineBuffer leading to this confusion.

src/engine/controls/enginecontrol.h
JoergAtGithub: I created #13819 for the not Macro related changes in this PR. Please follow-up there.

src/engine/enginebuffer.cpp
JoergAtGithub: I remove the macro specific QList. Now it handled as all the other controls in EngineBuffer and released by qDeleteAll(m_engineControls); in the deconstructor.

JoergAtGithub: It's deleted by qDeleteAll(m_engineControls); in the deconstructor, as all other enginebuffer controls.

src/engine/controls/enginecontrol.h
JoergAtGithub: I moved the topic about renaming "rate" to a dedicated issue #13833 and hide this discussion here

src/proto/macro.proto
```diff
@@ -0,0 +1,14 @@
...
+
+package mixxx.track.io;
+
+option optimize_for = LITE_RUNTIME;
+
+message Macro {
+  message Action {
+    optional uint32 type = 1;
+    optional uint64 source_frame = 2;
+    optional uint64 target_frame = 3;
```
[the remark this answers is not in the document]
Holzhaus: Hmm, maybe these should be required. Not sure. My understanding of protobuf is that this does not affect the serialized output, just the deserialization code. So we should probably make these required and "downgrade" them to optional only if the becomes necessary.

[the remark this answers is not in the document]
JoergAtGithub: @daschuer @Swiftb0y Do you know Protobuf well enough to answer this question with certainty?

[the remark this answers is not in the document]
daschuer: Not with certainty, but until now my rule of thump was that everything should be optional just in case. In protobuf 3 the concept of required fields has been removed altogether: 
https://stackoverflow.com/questions/31801257/why-required-and-optional-is-removed-in-protocol-buffers-3

src/engine/controls/macrocontrol.cpp
```diff
@@ -0,0 +1,336 @@
...
+          m_recordedActions(kRecordingQueueSize),
+          m_nextActionIndex(0),
+          m_COStatus([this]() { return getConfigKey("status"); }()),
+          m_CORecord([this]() { return getConfigKey("record"); }()),
+          m_COPlay([this]() { return getConfigKey("play"); }()),
+          m_COEnable([this]() { return getConfigKey("enable"); }()),
+          m_COLoop([this]() { return getConfigKey("loop"); }()),
+          m_activate([this]() { return getConfigKey("activate"); }()),
+          m_toggle([this]() { return getConfigKey("toggle"); }()),
+          m_clear([this]() { return getConfigKey("clear"); }()) {
```
Swiftb0y, line 34: this is dubious because it relies on the class already being partially initialized. I think the better option would be to just pass in the `group` along with `getConfigKey` and make it `static`...

src/engine/controls/macrocontrol.cpp
```diff
@@ -0,0 +1,336 @@
...
+    m_COStatus.setReadOnly();
+    setStatus(Status::NoTrack);
+
+    m_updateRecordingTimer.moveToThread(qApp->thread());
```
Swiftb0y, line 38: this needs a comment.

src/engine/controls/macrocontrol.cpp
```diff
@@ -0,0 +1,336 @@
...
+            &MacroControl::slotToggle);
+    m_clear.setButtonMode(mixxx::control::ButtonMode::Toggle);
+    connect(&m_clear,
+            &ControlObject::valueChanged,
+            this,
+            &MacroControl::slotClear);
+}
+
+// FIXME(xeruf) Jumps while paused (e.g. via GotoAndStop) are not properly recorded
+// since this function is not called
```
Swiftb0y, line 83: do you plan to address these FIXMEs? These certainly sound like bugs.

src/engine/controls/macrocontrol.h
```diff
@@ -0,0 +1,91 @@
...
+
+    void setStatus(Status);
+
+    ConfigKey getConfigKey(const QString& name);
+    int m_slot;
+
+    /// The unquantized FramePos of a jump that is yet to be processed
+    mixxx::audio::FramePos m_queuedJumpTarget;
+
+    rigtorp::SPSCQueue<MacroAction> m_recordedActions;
```
Swiftb0y, line 75: since this is designed for multithreading, this suggests that `MacroControl` is also used from multiple threads. Its not clear to me how though so I'd appreciate if you could document which member functions are expected to be accessed concurrently (and thus need to be threadsafe and which dont).

Swiftb0y (commented):
couple comments to try and help get this along.

src/track/macroaction.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include "audio/frame.h"
+#include "engine/engine.h"
+#include "proto/macro.pb.h"
+namespace proto = mixxx::track::io;
```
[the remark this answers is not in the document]
JoergAtGithub: The suggested code change breaks the build

src/engine/controls/macrocontrol.cpp
```diff
@@ -0,0 +1,336 @@
...
+constexpr uint kRecordingTimerInterval = 100;
+constexpr size_t kRecordingQueueSize = kRecordingTimerInterval / 10;
+
+const QString kRecordingSuffix = QStringLiteral(" [Recording]");
+const QString kInterruptedSuffix = QStringLiteral(" [Interrupted]");
+const QRegExp kAutoGeneratedLabelRegExp(QStringLiteral("[0-9. ]+(\\[\\w*\\])?"));
+} // namespace
+
+ConfigKey MacroControl::getConfigKey(const QString& name) {
+    return ConfigKey(m_group, QStringLiteral("macro_%1_%2").arg(m_slot).arg(name));
```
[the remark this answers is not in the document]
JoergAtGithub: Also this suggestion breaks the build

src/track/macroaction.h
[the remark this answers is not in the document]
Swiftb0y: right, you'll obviously have to replace all the occurences of `proto` with the longer name...

src/engine/controls/macrocontrol.cpp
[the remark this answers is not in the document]
Swiftb0y: right, forgot that the multi-arg `.arg()` only works for string-/char-likes. Ignore then.

src/engine/controls/macrocontrol.h
Holzhaus: MacroControl obviously lives in the GUI thread, but IIRC some methods (e.g., `slotJumpQueued`) are invoked via direct connection from the engine thread.

PR #10989, opened 2022-10-22
qopenglwindow-based replacement of qglwidget, full glsl shader based waveforms, vumeters and spinnies

m0dB:
This PR addresses performance issues, particularly on macOS, by replacing the (deprecated) QGLWidget with a new class WGLWidget. When cmake option QOPENGL is ON (the default), WGLWidget is derived from QWidget and uses QOpenGLWindow for drawing. When cmake option QOPENGL is OFF, WGLWidget derives directly from QGLWidget, effectively leaving everything as before.

Additionally, this PR implements waveform, spinny and vu-meter rendering using GLSL shaders for drawing for better performance. The original implementations were combining QPainter-based drawing with openGL. Using QPainter with QOpenGLWindow as a paint device causes performance issues.

The original WSpinny and WVuMeter classes are still available as fall back for when GLSL is not available or in case we encounter issues. By default the new WSpinnyGLSL and WVuMeterGLSL are used.

All original waveform types are also still available, but the new (all-shader) type waveforms should perform better on all platforms.

Notes:
- The QOPENGL cmake option sets preprocessor define MIXXX_USE_QOPENGL, which is used to select the QOpenGLWIndow based code instead of the QGLWidget based code.
- When WGLWidget derives from QGLWidget, it adds some methods, this is merely to have a common API for both implementations
- I also tried an alternative approach: replacing QGLWidget with the newer QOpenGLWidget, but that required too many changes to get to work
- I have tried to address all the comments in the reviews have been addressed, with the notable exception of a requested API improvement of the shader and vertex code, but this can be done later and will mainly be aesthetics.

Known issues:
- alpha transparency of the beat lines in the legacy GL waveforms causes a graphical artifact and had to be forced to 1.0
- when changing the layout / resizing, the QOpenGLWindows are moved/resized ahead (?) of the rest of the widgets.

PR #13173, opened 2024-04-24
(fix) Waveform viewer: reject focus on click, allow global keyboard shortcuts

### LINKED ISSUE #13160: Play/Pause using keyboards shortcuts fails after loading a track due  to a focus change introduced in v2.4.0
### Bug Description

using the keyboard shortcuts to play/pause a track start a search instead of playing/pausing the track.

To replicate the issue, load a track into a deck by drag'n'drop to either the waveform display or waveform overview, then use the shortcut keys to play/pause. instead of the track playing/pausing, the characters are added to the search box.

I have to click into the waveform display for the shortcuts to work and for some reason every time I alt tab back into mixxx  the focus returns into the search box.

After loading a track into a deck I expect the shortcuts to play/pause and not search for another track. This was the behaviour in previous mixxx version.

### Version

2.4.0

### OS

Linux (manjaro)

ronso0:
Fixes #13160 for waveforms and spinnies, already noticed in the waveforms PR https://github.com/mixxxdj/mixxx/pull/10989#issuecomment-1526191537

Focus is still removed from the current focus widget, but cleared right away so keyboard shortcuts still work.

Obviously, it'd be better not to move focus in the first place, but I didn't manage to prevent that.
Previously I tried with eventFilter in OpenGLWindow but that didn't work for some reason.
Also, only rejecting the FocusIn event is not sufficient, even though the child widget has the Qt::NoFocus policy, so it shouldn't accept focus in the first place :man_shrugging: 

**edit** I'm experiencing inconsitent behaviour: doing another test (since I'm not happy with this hack) it turns out that rejecting the FocusIn **is** sufficient.
However, I'd still prefer to not focus the QOpenGLWindow in the first place.

@m0dB What do you think?

### DIFF
--- a/src/widget/openglwindow.cpp
+++ b/src/widget/openglwindow.cpp
@@ -93,6 +93,11 @@ bool OpenGLWindow::event(QEvent* pEv) {
             return result;
         }
 
+        if (t == QEvent::FocusIn) {
+            pEv->ignore();
+            return false; // clazy:exclude=base-class-event
+        }
+
         // Send all remaining events to the widget that owns the window
         // container widget that contains this QOpenGLWindow. With this mouse
         // events, keyboard events, etc all arrive as intended, including the


ronso0:
now I can't even reproduce the keyboard issue... "verry, verry weird!"
sorry for the noise.

However, now that I'm at it I'll check if the window focus change can be prevented to not steal focus from the main window's widgets.

PR #13174, opened 2024-04-25
(fix) waveform / spinnies: don't take keyboard focus on click

ronso0:
_Currently keyboard shortcuts still work when waveforms/spinnies are click (OpenGLWindow is activated), but focus is removed from the library widgets for example which is not optimal since waveforms/spinnies don't have any special keyboard functions._

_This PR just prevents the focus shift._

Simply set the `ToolTip` flag to reject focus.
Couldn't spot any regressions, waveform/spinny interaction works flawlessly, tooltips are shown correctly.

Tested on Linux with Qt 5.12.8

Note: this trick does apparently not work with Qt 6 :laughing: (Qt 6.2.3)
____

_Remaining issue:
on window activation, and when no widget has focus, the searchbar gets focus, which also prevents keyboard shortcuts to work right away._

### DIFF
--- a/src/widget/openglwindow.cpp
+++ b/src/widget/openglwindow.cpp
@@ -12,6 +12,9 @@
 OpenGLWindow::OpenGLWindow(WGLWidget* pWidget)
         : m_pWidget(pWidget) {
     setFormat(WaveformWidgetFactory::getSurfaceFormat());
+    // Set the tooltip flag to prevent this window/widget from getting
+    // keyboard focus on click.
+    setFlag(Qt::ToolTip);
 }
 
 OpenGLWindow::~OpenGLWindow() {


ronso0:
I was curious what's happening in general with the embedded OpenGL window so I added some trace output to LibraryControl's widget focus slots (plain 2.4 branch):

* start Mixxx, click either a waveform or a spinny
* = OpenGLWindow is QApp's 'active window'
* press Tab key:
* QApp's `focusWidget()` changes multiple times, to be precise it seems to cycle through these widgets **3 times** before finally stopping in the searchbar:
WSearchLineEdit --> WLibrarySidebar --> WTrackTableView

___
With this branch OpenGL has no focus, doesn't receive key events and therefore the focus jumps to the searchbar right away.

m0dB:
I found another solution that works with Qt 6. I'll create PRs for 2.4 and main.

m0dB:
Actually, I think it's better to do it in this PR, with a conditional.

src/widget/openglwindow.cpp
```diff
@@ -12,6 +12,9 @@
 OpenGLWindow::OpenGLWindow(WGLWidget* pWidget)
         : m_pWidget(pWidget) {
     setFormat(WaveformWidgetFactory::getSurfaceFormat());
+    // Set the tooltip flag to prevent this window/widget from getting
+    // keyboard focus on click.
+    setFlag(Qt::ToolTip);
```
m0dB, line 17:
```suggestion
#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
    setFlag(Qt::ToolTip);
#else
    setFlag(Qt::Tool);
#endif
```

m0dB:
If you commit my proposed change, I will approve and merge the PR.

src/widget/openglwindow.cpp
ronso0: Unfortunately this doesn't work either in Qt6.
Also tried _SplashScreen_ and _SubWindow_

I suggest to stick with this for 2.4.x and take a closer look at the qt sources, i.e. the Qt5 -> Qt6 changes.

m0dB: Strange. Adding setFlag(Qt::Tool) definitely works for me with main on macOS...

ronso0: Ha, obvious `Qt::WindowDoesNotAcceptFocus` flag does the trick on Linux wit Qt 6.2.3

Will open a fixup.

PR #13176, opened 2024-04-27
Skins: Add a <NumberDuration> control for skins

cr7pt0gr4ph7:
This PR adds a `WNumberDuration`control for displaying a control value given in seconds as a formatted time (e.g. `hh:mm:ss`).
Although the `WNumberPos` has a similar purpose, it is tailored strictly for displaying the play position of a deck. The `WTime` control is also similar, but only displays the current time.

The goal is to be able to display e.g. the remaining duration of the Auto DJ queue in the UI in a skin-controllable way, like so:
Example screenshot

(Note: The above screenshot is from one of my development builds. Automatically calculating the remaining duration of the Auto DJ queue has its own gotchas, and so will be provided in a separate PR)

### DIFF
--- a/src/widget/wnumberduration.cpp
+++ b/src/widget/wnumberduration.cpp
@@ -1,5 +1,7 @@
 #include "widget/wnumberduration.h"
 
+#include <QTime>
+
 #include "control/controlproxy.h"
 #include "moc_wnumberduration.cpp"
 #include "util/duration.h"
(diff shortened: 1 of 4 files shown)


src/widget/wnumberduration.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+
+#include "preferences/dialog/dlgprefdeck.h"
+#include "skin/legacy/skincontext.h"
+#include "wnumber.h"
+
+class ControlProxy;
+
+class WNumberDuration : public WNumber {
```
acolombier, line 9: Could you please add some doc-string to document what this component is used for?

src/widget/wnumberduration.cpp
```diff
@@ -0,0 +1,71 @@
...
+        m_displayFormat = TrackTime::DisplayFormat::KILO_SECONDS;
+    } else if (formatString == "seconds_long") {
+        m_displayFormat = TrackTime::DisplayFormat::SECONDS_LONG;
+    } else if (formatString == "seconds") {
+        m_displayFormat = TrackTime::DisplayFormat::SECONDS;
+    } else if (formatString == "traditional") {
+        m_displayFormat = TrackTime::DisplayFormat::TRADITIONAL;
+    } else if (formatString == "traditional_coarse") {
+        m_displayFormat = TrackTime::DisplayFormat::TRADITIONAL_COARSE;
+    } else {
```
acolombier, line 28: You might want to use `SKIN_WARNING` macro to alert of a potentially unsupported format. I understand that `TimeFormat` may be an arbitrary string passed to `QTime::toString`. Perhaps you could use two attributes (e.g `TimeFormat` is an enum, including `custom` and `CustomFormat` allow you to define the arbitrary string, stored in `m_displayFormatString`. This way, it will help debugging cases where user accidentally mistype a pre-constructed enum.

src/widget/wnumberduration.cpp
```diff
@@ -0,0 +1,71 @@
...
+        // pass it on to QTime for formatting
+
+        // NOTE: QTime() constructs a 'null' object,
+        // but we need 'zero' here.
+        QTime t = QTime(0, 0).addMSecs(static_cast<int>(
+                dSeconds * mixxx::Duration::kMillisPerSecond));
+
+        fmt = t.toString(m_displayFormatString);
+    } else if (m_displayFormat == TrackTime::DisplayFormat::KILO_SECONDS) {
+        fmt = mixxx::Duration::formatKiloSeconds(dSecondsAbs);
```
acolombier, line 49: Perhaps using a `switch` for the `m_displayFormat` might help with readability, although strictly equivalent?

acolombier (commented):
That's looking good. Perhaps it could be nice to add this component in one skin?

src/widget/wnumberduration.cpp
cr7pt0gr4ph7: How about two properties `PredefinedFormat` (or `StandardFormat`/`DefaultFormat`) and `CustomFormat`)? And just use the one that is specified, and emit a warning if both are specified (and fallback to `CustomFormat`).

ronso0:
Tbh I'm reluctant adding a widget that is not used in official skins.
which duration control values can this be hhooked up to currently?

src/widget/wnumberduration.cpp
acolombier: Yes, that sounds perfect!

cr7pt0gr4ph7:
> Tbh I'm reluctant adding a widget that is not used in official skins. which duration control values can this be hhooked up to currently?

It's purpose would be hooking it up to `[AutoDJ],time_remaining`, as provided by #13183. The purpose still stands, although a different to-be-implemented UI control that could render multiple control values into a single localized template string with reorderable placeholders (e.g. `Auto DJ: %1 tracks in queue, %2 remaining` with `1: [AutoDJ]tracks_remaining` and `2: [AutoDJ]time_remaining`) might fit that purpose better, and would also enable other use cases.

PR #13182, opened 2024-04-28
(fix) WCoverArt: popup on click release

### LINKED ISSUE #13181: Cover art widget doesn't open popup on click
Probably a regression from #13135 
Affects main and 2.4.1 with the backport

I'll take a look.

ronso0:
Fixes #13181 
Forgot to backport #13035 

Though, I'm confused why this can break so easily.

FWIW, this is what various button tests return for trackpad click release with Qt 5.12.8:
**FALSE** pEvent->buttons() & Qt::LeftButton
**FALSE** pEvent->buttons() == Qt::LeftButton
**FALSE** pEvent->buttons().testFlag(Qt::LeftButton)
**TRUE** pEvent->button() == Qt::LeftButton (fix)

* various widgets' mouse**Press**Event() succesfully use 
`pEvent->buttons() == Qt::LeftButton`
* in release events however we use
`pEvent->button()`

### DIFF
--- a/src/widget/wcoverart.cpp
+++ b/src/widget/wcoverart.cpp
@@ -259,7 +259,8 @@ void WCoverArt::mouseReleaseEvent(QMouseEvent* pEvent) {
         return;
     }
 
-    if (pEvent->buttons() == Qt::LeftButton && m_loadedTrack &&
+    if (pEvent->button() == Qt::LeftButton &&
+            m_loadedTrack &&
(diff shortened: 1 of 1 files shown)


ronso0:
FWIW I discovered this by chance when trying to view a cover in a live session.
We should set up a smoke test check list. Or,if such an official list exists, add some cover test cases.

daschuer (approved):
I can confirm the issue is fixed. Thank you.

PR #13188, opened 2024-04-30
(fix) Controller IO table: fix display text for Action/control delegate

ronso0:
The Action column displayed only "No control chosen".


Swiftb0y (approved):
good catch. LGTM. Thank you

ronso0:
> good catch.

well, this was like an hour of debugging to understand that the current implmentation converted the control to QString (control description), which is actually what we want displayText() to return... but that couldn't handle it and returned "No control ..."

Issue #13030, opened 2024-03-31
remove debug output for window state change


ronso0:
remove the output added in #11566
(as soon as it is confirmed to work flawlessly on all targets)

PR #13189, opened 2024-04-30
Revert "DEBUG output for fullscreen changes"

daschuer:
This reverts commit 5a4fe74ca9bd90814ef0eadf3730a073fd01783f.
fixing #13030

### DIFF
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -167,7 +167,6 @@ void MixxxMainWindow::initializeQOpenGL() {
 #endif
 
 void MixxxMainWindow::initialize() {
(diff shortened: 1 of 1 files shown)


ronso0:
I'm not sure 2.5 has been tested sufficiently already. I'd like to keep this during beta and finally remove it when it's clear there are no regressions (only worried about macOS actually).

daschuer:
Sounds good. I was unsure because the milestone.

daschuer:
We are in the process of the 2.5.o release now. Can this be merged?

ronso0:
Yup, there were no complaints afaik.
In case we need debug info we can simply revert this.

PR #13199, opened 2024-05-03
Track/BPM: add `bpmlock` control

ronso0:
This replaces recently introduced `bpm_toggle_lock` with `bpmlock`.
Benefit is that we can now bind GUI controls also for feedback, e.g. have a BPM lock toggle and block/disable beatgrid controls if BPM are locked.

The pushbutton only triggers the un/locking in Track, CO is updated by track signal.

### DIFF
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -130,10 +130,10 @@ BpmControl::BpmControl(const QString& group,
             this, &BpmControl::slotTapFilter,
             Qt::DirectConnection);
 
-    m_pToggleBpmLock = std::make_unique<ControlPushButton>(
-            ConfigKey(group, "bpm_toggle_lock"), false);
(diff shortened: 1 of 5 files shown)


ronso0:
I think I / we missed the opportunity to make full use of this new control in #13041, so in order to not have to break mappings later on I highly recommend to add this to 2.5-beta.

daschuer (approved):
Yes, that looks nicer. Thank you.

ronso0:
Thanks for the quick review!

Nite to myself:
document CO

ronso0:
I forgot to change the ControlPickerMenu, too.
Just pushed the fix directly d59eda694b7292782569195547c46e76bf01e91d

ronso0:
@FrankwaP I hope you noticed this fixup?

FrankwaP:
@ronso0 Thank you for the ping as  I might have missed this update!

PR #13209, opened 2024-05-05
Allow to drop files to decks with unsupported or no file extensions

### LINKED ISSUE #13204: Allow droping any file extntion to a deck and play it accoring to MIME type
### Feature Description

This is already possible with the wrong otherwise supported extension, but not with any.

daschuer:
This fixes #13204

### DIFF
--- a/src/sources/soundsourceproxy.cpp
+++ b/src/sources/soundsourceproxy.cpp
@@ -284,7 +284,10 @@ bool SoundSourceProxy::isUrlSupported(const QUrl& url) {
 
 // static
 bool SoundSourceProxy::isFileSupported(const mixxx::FileInfo& fileInfo) {
-    return isFileNameSupported(fileInfo.fileName());
+    const QString fileType =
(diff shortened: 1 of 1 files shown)


acolombier (approved):
Once this is merged, the last usage of `isFileNameSupported` would be [here](https://github.com/mixxxdj/mixxx/blob/main/src/coreservices.cpp#L445). Could it be worth replacing with file introspection instead of file name extension guessing, this way we can remove this function entirely? Or is there a reason for keeping this in the future?

Swiftb0y (approved):

Swiftb0y:
Thank you.

Swiftb0y:
> Once this is merged, the last usage of isFileNameSupported would be [here](https://github.com/mixxxdj/mixxx/blob/main/src/coreservices.cpp#L445). Could it be worth replacing with file introspection instead of file name extension guessing, this way we can remove this function entirely?

`isFileNameSupported` is brittle, so lets remove it entirely.

PR #13211, opened 2024-05-05
(fixup) OpenGLWindow / waveform / spinnies: don't get keyboard focus on click

ronso0:
fixup for #13174 that also work with Qt6 (Qt 6.2.3 on Linux).

@m0dB Please take a look / test on macOS.


daschuer:
It looks like the flag is fixed for macOS in Qt 5.3 
https://codereview.qt-project.org/c/qt/qtbase/+/88654

LGTM, thank you.

PR #13214, opened 2024-05-06
Track menu, Remove from disk: stop and eject all affected decks

ronso0:
also stop decks and eject tracks if Move to Trash is invoked for track(s) from the tracks view.


daschuer (approved):
LGTM, Thank you.

Issue #13161, opened 2024-04-22
Saved beatjump


acolombier:
### Feature Description

Similar to saved loop, it would be great to be able to add saved beatjump. This would allow the track to automatically jump from a preset point to another preset point. This usecase would be be to skip sections of the track with low interest (silence for example).

Currently, the workaround is to set to cues with various colour or label (eg the first one red with labelled `silence`, the second one green labelled `chorus`) and use the first red cue as visual marker, helping timing the second cue jump. This is suboptimal as:
- It uses two cues
- It is hard to perfectly time and usually leads to audio little glitch 

Like for saved loop, it would be great to be able to enable or disable the jump

acolombier:
Indeed, I think that feature would have been covered by macros. Do I understand correctly that the macros feature was abandoned?
Could we consider saved beatjump without the whole macros framework?

ronso0:
> Do I understand correctly that the macros feature was abandoned?

Yes, sort of. The idea is still good, even though IIUC #4528 was canceled because a) there were _some_ unedited reviews and b) Xeruf lost interest and moved on to other projects.

> Could we consider saved beatjump without the whole macros framework?

I'd say yes.
It requires a special cue type, basically like a loop cue with start and end position (where start may be after end to allow one-time backward jumps?) and do the In -> Out seek like a reverse loop.
As you described, activation would be identical to 'catching loops', just that it 'loops' once.

Setup could be pretty much the same like creating a manual loop (set In pos, seek, set Out pos).
I think as a starter, we could adopt the looping controls In / Out and add a "Jump".
Adjusting LoopingControl will be a lot of Fun :tm: :laughing: 

___
A related feature I was thinking of recently:
simplify the workflow of turing a loop cu into a regular hotcue which currently requires deleting the loop cue and create a hotcue in the same position.

JoergAtGithub:
I just pushed a PR with an updated version of #4528 : https://github.com/mixxxdj/mixxx/pull/13172
Just the backend, no GUI.

PR #13215, opened 2024-05-06
feat: add type toggle to cue popup

acolombier:
This adds the ability to changes the cue type without having to reset it. It also allow updating the saved loop end using the beatloop size.

I would like to also allow right-click on the button to set the end loop to the current play position to allow more granular control, and prepare the field for saved beatjump. Any component I can use or should I make a custom QPushButton?

https://github.com/mixxxdj/mixxx/assets/7086688/91899c40-b968-4fc6-8d44-8673b72708d8

This is a dependency for #13161


daschuer (changes requested):
Just a minor comment. The rest looks good.

src/widget/wcuemenupopup.h
```diff
@@ -52,10 +55,13 @@ class WCueMenuPopup : public QWidget {
   private slots:
     void slotEditLabel();
     void slotDeleteCue();
+    void slotUpdate();
+    void slotSavedLoopCue();
     void slotChangeCueColor(mixxx::RgbColor::optional_t color);
 
   private:
     ColorPaletteSettings m_colorPaletteSettings;
```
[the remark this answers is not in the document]
acolombier: Sounds good. Any suggestion for the right-click requirement I suggested in description? Should I create a small class inheriting PushButton or is there something I could already use?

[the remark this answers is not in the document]
acolombier: I created a custom PushButton. Let me know if there is something already existing.

src/widget/wcuemenupopup.cpp
```diff
@@ -135,6 +186,63 @@ void WCueMenuPopup::slotDeleteCue() {
...
+    VERIFY_OR_DEBUG_ASSERT(m_pCue != nullptr) {
+        return;
+    }
+    VERIFY_OR_DEBUG_ASSERT(m_pTrack != nullptr) {
+        return;
+    }
+    VERIFY_OR_DEBUG_ASSERT(m_pBeatLoopSize.valid()) {
+        return;
+    }
+    if (m_pCue->getType() == mixxx::CueType::Loop) {
```
ronso0, line 225: I've expected that right-click turns it into a loop cue right away.
I.e. make righ-click use the current position, regardless the current cue type.

ronso0 (commented):
Thanks for working on this, much appreciated!

I've lef some comments.

src/widget/wcolorpicker.cpp
```diff
@@ -46,9 +46,6 @@ WColorPicker::WColorPicker(Options options, const ColorPalette& palette, QWidget
     QGridLayout* pLayout = new QGridLayout();
     pLayout->setContentsMargins(0, 0, 0, 0);
 
-    pLayout->setSizeConstraint(QLayout::SetFixedSize);
-    setSizePolicy(QSizePolicy());
```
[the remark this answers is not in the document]
ronso0: Okay, got it:
the popup needs to grow when the cue is turned into a loop cue (start - end are displayed next to the title).

ronso0:
Please also add some styling for Deere.

src/widget/wcolorpicker.cpp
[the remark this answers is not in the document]
ronso0: > use a fixed size for the time display

Well, that depends on the track length, ie. we'd need to calculate the max. length on show.

Additionally, we could remove the "Hotcue" prefix and display on ly the number. I mean, it's pretty clear this is a hotcue.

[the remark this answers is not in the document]
acolombier: > Well, that depends on the track length, ie. we'd need to calculate the max. length on show.

That sounds sketchy. What if the font size/family is changed by style-sheet? Do you think we could keep the size policy change as-is not to over complicate that change?

> Additionally, we could remove the "Hotcue" prefix and display on ly the number

Agreed. Or we could display the type explicitly (Hotcue, Saved loop, Saved jump, ...)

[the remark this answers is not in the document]
ronso0: > That sounds sketchy. What if the font size/family is changed by style-sheet? Do you think we could keep the size policy change as-is not to over complicate that change?

The font can only be changed when the skin is reloaded.
I meant to calculate the label length via FontMetrics when the popup is shown. Like, get the max. possible end position of a loop cue.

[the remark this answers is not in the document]
ronso0: > Agreed. Or we could display the type explicitly (Hotcue, Saved loop, Saved jump, ...)

Yes, I'd prefer icons then.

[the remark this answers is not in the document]
acolombier: Yeah I got that, but QFontMetrics requires you to know the font attributes which are used. I assume you meant to use a `QStyleOption` to capture the style property at painting and invalidate the widget with a new min/fixed size. That's what I meant by "can we not go the over complicated way".

acolombier:
> Please also add some styling for Deere.

Done

> When I just toggle a loop cue's type accidentally, the loop size changes, right? Seems like sub optimal UX.

Done

src/widget/wcolorpicker.cpp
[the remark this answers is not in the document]
ronso0: I thought we could get the metrics via QWidget::fontMetrics() in `setTrackAndCue()`, but indeed we need the style from the painter :+1:

daschuer (approved):

daschuer:
@ronso0 and @Swiftb0y, It this ready to merge now?

Swiftb0y (commented):
I don't have the time to test this myself right now. Can you double check that the loopsize still doesn't change when toggling the type and the loop is not aligned to a beat. For example when a user sets up a loop with the in-out buttons (while having quantize disabled) and then assigns a hotcue to that loop.

src/widget/wcuemenupopup.cpp
```diff
@@ -135,6 +187,65 @@ void WCueMenuPopup::slotDeleteCue() {
...
+    VERIFY_OR_DEBUG_ASSERT(m_pTrack != nullptr) {
+        return;
+    }
+    VERIFY_OR_DEBUG_ASSERT(m_pBeatLoopSize.valid()) {
+        return;
+    }
+    const mixxx::BeatsPointer pBeats = m_pTrack->getBeats();
+    auto position = mixxx::audio::FramePos::fromEngineSamplePos(
+            m_pPlayPos.get() * m_pTrackSample.get());
+    m_pCue->setType(mixxx::CueType::Loop);
```
ronso0, line 232: Shouldn't this better be set after the position has been set?
Otherwise, in case the end could not be set, we'd have a loop cue but no  loop.

src/widget/wcuemenupopup.cpp
```diff
@@ -135,6 +187,65 @@ void WCueMenuPopup::slotDeleteCue() {
...
+    if (!m_pQuantizeEnabled.toBool() || !pBeats) {
+        m_pCue->setEndPosition(position);
```
ronso0, line 234: We need to reject if we're at the cue position, otherwise we'll hit
https://github.com/mixxxdj/mixxx/blob/4164341f5063b58fd9251e75620d48ed9ad18a6d/src/engine/controls/loopingcontrol.cpp#L655 when activating the cue.

ronso0 (changes requested):
Some remarks.

Shall we also change the cue color when changing the type?
if (type == loop && color == defaultLoopCueColor) { cue->setColor(defaultHotcueColor }
and vice versa for hotcue -> loopcue.

ronso0:
I also have a commit (almost) ready to fix the 'active' icon. For some reason `qproperty-icon: url(..);` does not style the checked state, but `image: ..` works (though doesn't allow size adjustments like with the qproperty).
I'll share that as soon as I find time next days.

src/widget/wcuemenupopup.cpp
```diff
@@ -40,6 +53,24 @@ WCueMenuPopup::WCueMenuPopup(UserSettingsPointer pConfig, QWidget* parent)
...
+    m_pSavedLoopCue = new CueTypePushButton(this);
+    m_pSavedLoopCue->setToolTip(
+            tr("Toggle this cue type between normal cue and saved loop, using "
+               "the current beatloop size or the current play position") +
+            "\n\n" +
+            tr("Left-click: Toggle between normal cue and saved loop, using "
+               "the current beatloop size as the loop size") +
+            "\n" +
+            tr("Right-click: Set the current play position as the loop end and "
+               "make the cue a saved loop if not"));
```
ronso0, line 65:
```suggestion
            tr("Right-click: If the current play position is after the cue, set that position as loop end and "
               "make the cue a saved loop"));
```

ronso0 (changes requested):

ronso0:
This is the button style commit
2af95dfef9

acolombier:
> Shall we also change the cue color when changing the type?

Good question. I would suggest we don't, but I guess we could change it if the current color matches the current cue type, before change? What do you think?

ronso0:
> > Shall we also change the cue color when changing the type?
> 
> Good question. I would suggest we don't, but I guess we could change it if the current color matches the current cue type as you suggested, before change? What do you think?

Yup, that's what I meant.
Related code we can adopt is here https://github.com/mixxxdj/mixxx/blob/4164341f5063b58fd9251e75620d48ed9ad18a6d/src/engine/controls/cuecontrol.cpp#L936-L950

src/widget/wcuemenupopup.h
```diff
@@ -52,10 +72,17 @@ class WCueMenuPopup : public QWidget {
   private slots:
     void slotEditLabel();
     void slotDeleteCue();
+    void slotUpdate();
+    void slotSavedLoopCue();
+    void slotAdjustSavedLoopCue();
```
ronso0, line 77: Could we call the slots
```suggestion
    void slotSavedLoopCueAuto();
    void slotSavedLoopCueManual();
```
to better reflect the difference?

ronso0:
Can we remove the obsolete cue end position after switching from loop -> hotcue _when closing the menu_?

It just feel weird to be left with hotcues that have an end position.

ronso0:
> Can you double check that the loopsize still doesn't change when toggling the type and the loop is not aligned to a beat. For example when a user sets up a loop with the in-out buttons (while having quantize disabled) and then assigns a hotcue to that loop.

This is taken care of.
The only way to screw up a loop is
* have a loop cue
* **left**-click loop cue button
* _Whoopsy!_
* **right**-click loop cue button

src/widget/wcuemenupopup.cpp
```diff
@@ -65,11 +98,31 @@ WCueMenuPopup::WCueMenuPopup(UserSettingsPointer pConfig, QWidget* parent)
...
 }
 
-void WCueMenuPopup::setTrackAndCue(TrackPointer pTrack, const CuePointer& pCue) {
-    if (pTrack && pCue) {
-        m_pTrack = pTrack;
-        m_pCue = pCue;
+void WCueMenuPopup::setTrackCueGroup(
+        TrackPointer pTrack, const CuePointer& pCue, const QString& group) {
+    m_pTrack = pTrack;
+    m_pCue = pCue;
```
ronso0, line 104: Please restore the null check, with early return.

ronso0 (changes requested):

src/widget/wcuemenupopup.h
acolombier: Not sure this makes it more clear, to me this feels even more confusing. How about I add a docstring to make it clearer?

acolombier:
> Can we remove the obsolete cue end position after switching from loop -> hotcue _when closing the menu_?

Yes, that was fixed in the follow up PR (the saved jump). Going to fix that one instead.



> > Can you double check that the loopsize still doesn't change when toggling the type and the loop is not aligned to a beat. For example when a user sets up a loop with the in-out buttons (while having quantize disabled) and then assigns a hotcue to that loop.
> 
> This is taken care of. The only way to screw up a loop is
> 
>     * have a loop cue
> 
>     * **left**-click loop cue button
> 
>     * _Whoopsy!_
> 
>     * **right**-click loop cue button

This would require to implement an history of cue modification and I'm not going to do this here. If this a deal breaker, happy to close the PR.

ronso0:
> This would require to implement an history of cue modification and I'm not going to do this here. If this a deal breaker, happy to close the PR.

I wanted to point out that this is unlikely to happen by accident, and definitely not blocking this.

src/widget/wcuemenupopup.h
ronso0: For me it's just that `slotAdjustSavedLoopCue` sounds like we adjust an existing loop cue, but the only difference to `slotSavedLoopCue` is that is uses the position instead of an auto-length.

acolombier: I see. With the docstring, I feel like those slots behaviour are well defined to developer, so happy to rename the symbol to anything we think is better fit.

acolombier:
Ah, I see. I thought you meant there was a missing edge case - I think we slightly lost in translation :)

I think all comments should have been addressed now.

src/widget/wcuemenupopup.h
ronso0: Thank you!

ronso0 (approved):
LGTM nox, thank you!

ronso0:
@daschuer wanna take a look?

acolombier:
Looks like @daschuer already approved - merge?

src/widget/wcuemenupopup.cpp
```diff
@@ -40,6 +53,28 @@ WCueMenuPopup::WCueMenuPopup(UserSettingsPointer pConfig, QWidget* parent)
...
+    m_pSavedLoopCue->setToolTip(
+            tr("Toggle this cue type between normal cue and saved loop, using "
+               "the current beatloop size or the current play position") +
+            "\n\n" +
+            tr("Left-click: Toggle between normal cue and saved loop, using"
+               "the current beatloop size as the loop size") +
+            "\n" +
+            tr("Right-click: If the current play position is after the cue,"
+               "set that position as loop end and make the cue a saved loop "
+               "if not"));
```
daschuer, line 66:
```suggestion
            tr("Toggle this cue type between normal cue and saved loop")
            "\n\n" +
            tr("Left-click: Use the old size or the current beatloop size as the loop size") +
            "\n" +
            tr("Right-click: Use the current play position as loop end if it is after the cue);
```

daschuer: I am not sure if keeping the old loop size is the intended behaviour, because it is invisible, when the cue is an normal cue. We loose the "feature" to adjust the beatloops size of a saved loop by this.

daschuer (changes requested):
During testing I had the situation that the cue type toggle button was lit, but the cue button in the skin did not change the state to loop. Unfortunately I was not able to reproduce the issue. 

At least we should make sure the toogle button state is corrected when changing the cue type fails for some reason.

src/widget/wcuemenupopup.h
```diff
@@ -2,14 +2,34 @@
...
 #include <QLabel>
 #include <QLineEdit>
+#include <QMouseEvent>
 #include <QPushButton>
 
+#include "control/pollingcontrolproxy.h"
 #include "preferences/colorpalettesettings.h"
 #include "track/cue.h"
 #include "track/track_decl.h"
+#include "util/parented_ptr.h"
```
ronso0, line 12: This is not used.
Did you intend to use it for `m_pSavedLoopCue`?
Could as well use it for the othe rmembers `m_pEditLabel`, `m_pColorPicker`, `m_pDeleteCue`.

ronso0 (changes requested):

ronso0:
> Can we remove the obsolete cue end position after switching from loop -> hotcue _when closing the menu_?
> 
> It just feel weird to be left with hotcues that have an end position.

49c052db12

src/widget/wcuemenupopup.cpp
acolombier: This was [requested](https://github.com/mixxxdj/mixxx/pull/13215#issuecomment-2098297092) by @Swiftb0y
Since you can change the loop size with right-click, I guess it is okay.
Perhaps I could also make double-click to change the loop size with the beatloops size value?

ronso0: > Perhaps I could also make double-click

I suggest to keep it simple.
The purpose of this button is to quickly change the type, with two loop options.
Adjusting saved loops should better be done with the looping controls IMO.

Adjusting the right-click behaviour might work:
* if before cue: reject
* if at cue: use beatloop size **NEW**
* if after cue: use play pos

src/widget/wcuemenupopup.h
acolombier: I meant to use `unique_ptr` - updated now

src/widget/wcuemenupopup.cpp
acolombier: That should be the behaviour. Also with your commit, the end pos is now removed after popup close so the left button may use the beatloop size again on next attempt

Swiftb0y: > I suggest to keep it simple.

I also thinks that important. if the functionality for retaining the previous size results in code that is too complicated, we should ensure the code remains understandable. In the worst case, the user accidentally clicks the wrong button (that is not easy to access in the first place) and they learn that it changes the length the hard way.

I'm sorry for being so indecisive on this.

acolombier: No problem. Could you please have a look with the current solution? It will allow to quickly revert the saved loop, and will "clean" the dangling end position when the popup is closed, thanks to @ronso0's work. I think the code remain quite simple, so hopefully we have a good trade-off between usability and code simplicity.

daschuer (approved):
Works good now. Thank you.

acolombier:
@ronso0 friendly ping, in case you missed the notification!

ronso0 (approved):

res/skins/LateNight/style_palemoon.qss
```diff
@@ -2498,20 +2504,32 @@ QPushButton#pushButtonRepeatPlaylist:!checked {
...
 }
+WCueMenuPopup #CueDeleteButton {
+  /* Note: we can't use qproperty-icon here because it's evauated only once
+     and therefore won't style the checked state */
+  image: url(skins:LateNight/palemoon/buttons/btn__delete.svg);
+}
+  WCueMenuPopup #CueDeleteButton:pressed {
+    background-color: #b24c12;
+    outline: none;
+    image: url(skins:LateNight/palemoon/buttons/btn__delete_active.svg);
```
fwcd, line 2520: Randomly stumbled upon this while building for iOS, this file seems to be missing, which results in a bunch of warnings of the form

```
qt.svg: warning Main Cannot open file '/private/var/containers/Bundle/Application/D577115B-62DB-4FFE-B072-4133259002FE/Mixxx.app/skins/LateNight/palemoon/buttons/btn__delete_active.svg', because: No such file or directory
```

(I know this PR didn't introduce this styling, but this is where the code was last touched)

acolombier: Yeah, I can also see this warning on Linux. Happy to create an issue? Should be an easy one to address.

fwcd: Done: #13492

PR #13218, opened 2024-05-06
Fix reading metadata for some file extensions

### LINKED ISSUE #13205: Metadata is not read from files with the wrong extension
### Bug Description

Mixxx uses already the mime type when encoding the file. This does however not work for metadata extraction and should be fixed. 

### Version

_No response_

### OS

_No response_

daschuer:
This replaces all file type guessing code form the file extension, by using the normalized file type stored in the Track object. 
This makes sure all parts are using the same file type, even if the extension is wrong. 

This fixes #13205

### DIFF
--- a/src/sources/metadatasourcetaglib.h
+++ b/src/sources/metadatasourcetaglib.h
@@ -7,16 +7,11 @@ namespace mixxx {
 // Universal default implementation of IMetadataSource using TagLib.
 class MetadataSourceTagLib : public MetadataSource {
   public:
-    explicit MetadataSourceTagLib(
-            const QString& fileName)
-            : m_fileName(fileName),
-              m_fileType(taglib::getFileTypeFromFileName(fileName)) {
-    }
     MetadataSourceTagLib(
             const QString& fileName,
-            taglib::FileType fileType)
+            const QString& fileType)
             : m_fileName(fileName),
-              m_fileType(fileType) {
+              m_fileType(taglib::stringToEnumFileType(fileType)) {
     }
 
     std::pair<ImportResult, QDateTime> importTrackMetadataAndCoverImage(
--- a/src/test/soundproxy_test.cpp
+++ b/src/test/soundproxy_test.cpp
@@ -7,6 +7,7 @@
 #include "sources/soundsourceproxy.h"
 #include "test/mixxxtest.h"
 #include "test/soundsourceproviderregistration.h"
+#include "track/taglib/trackmetadata_file.h"
 #include "track/track.h"
 #include "track/trackmetadata.h"
 #include "util/samplebuffer.h"
@@ -1084,5 +1085,17 @@ TEST_F(SoundSourceProxyTest, freeModeGarbage) {
                 providerRegistration.getProvider());
         ASSERT_TRUE(pContReadSource != nullptr);
         break;
+    }
+}
+
+TEST_F(SoundSourceProxyTest, taglibStringToEnumFileType) {
+    const QStringList fileTypes = SoundSourceProxy::getSupportedFileTypes();
+    for (const auto& fileType : fileTypes) {
+        qDebug() << fileType;
+        if (fileType != "okt" &&     // Oktalyzer
+                fileType != "stm") { // "Scream Tracker";
+            ASSERT_NE(mixxx::taglib::stringToEnumFileType(fileType),
+                    mixxx::taglib::FileType::Unknown);
+        }
     }
 }
--- a/src/track/serato/tags.cpp
+++ b/src/track/serato/tags.cpp
@@ -73,6 +73,7 @@ namespace mixxx {
 
 double SeratoTags::guessTimingOffsetMillis(
         const QString& filePath,
+        const QString& fileType,
         const audio::SignalInfo& signalInfo) {
     // The following code accounts for timing offsets required to
     // correctly align timing information (e.g. cue points) exported from
@@ -82,7 +83,7 @@ double SeratoTags::guessTimingOffsetMillis(
     // PR for more detailed information:
     // https://github.com/mixxxdj/mixxx/pull/2119
     double timingOffset = 0;
-    if (taglib::getFileTypeFromFileName(filePath) == taglib::FileType::MP3) {
+    if (fileType == "mp3") {
         const QString primaryDecoderName =
                 getPrimaryDecoderNameForFilePath(filePath);
         // There should always be an MP3 decoder available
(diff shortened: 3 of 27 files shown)


src/track/taglib/trackmetadata_file.cpp
```diff
@@ -45,36 +45,70 @@ void readAudioProperties(
...
+            TypePair{"module"_L1, FileType::Mod},
+            TypePair{"nst"_L1, FileType::Mod},
+            TypePair{"wow"_L1, FileType::Mod},
+            TypePair{"s3m"_L1, FileType::S3M},
+            TypePair{"xm"_L1, FileType::XM},
+            TypePair{"dsf"_L1, FileType::DSF},
+            TypePair{"dff"_L1, FileType::DSDIFF},
+            TypePair{"dsdiff"_L1, FileType::DSDIFF}};
+
+    auto it = std::find_if(lookupTable.cbegin(),
```
acolombier, line 108: Looks like this could be more specific
```suggestion
    const auto *it = std::find_if(lookupTable.cbegin(),
```

acolombier (commented):
Could it be worth adding one or few test case to ensure the behaviour and prevent potential regression?

src/test/soundproxy_test.cpp
```diff
@@ -1086,3 +1087,14 @@ TEST_F(SoundSourceProxyTest, freeModeGarbage) {
         break;
     }
 }
+
+TEST_F(SoundSourceProxyTest, taglibStringToEnumFileType) {
+    for (const auto& fileType : SoundSourceProxy::getSupportedFileTypes()) {
```
acolombier, line 1092: clazy here

src/track/taglib/trackmetadata_file.cpp
```diff
@@ -45,36 +45,70 @@ void readAudioProperties(
...
+    DEBUG_ASSERT(!fileType.isEmpty());
+
+    struct TypePair {
+        QLatin1String strType;
+        FileType eType;
+    };
+
+    // This table is aligned with detectByExtension() in fileref.cpp
+    constexpr std::array lookupTable = {
+            TypePair{"mp3"_L1, FileType::MPEG},
```
Swiftb0y, line 68: one day we should think about introducing a "minimal perfect hash function"-based approach for LUTs like these.

Swiftb0y: you'll also need to make it static, otherwise the compiler is more likely to choose to create it at runtime. 
```suggestion
    // This table is aligned with detectByExtension() in fileref.cpp
    constexpr static std::array lookupTable = {
            TypePair{"mp3"_L1, FileType::MPEG},
```

src/sources/metadatasourcetaglib.h
```diff
@@ -7,16 +7,11 @@ namespace mixxx {
...
-            : m_fileName(fileName),
-              m_fileType(taglib::getFileTypeFromFileName(fileName)) {
-    }
     MetadataSourceTagLib(
             const QString& fileName,
-            taglib::FileType fileType)
+            const QString& fileType)
             : m_fileName(fileName),
-              m_fileType(fileType) {
+              m_fileType(taglib::stringToEnumFileType(fileType)) {
```
Swiftb0y, line 14: whats the advantage of taking a string here instead of shifting the conversion duty to the caller? This smells [stringly-typed](https://wiki.c2.com/?StringlyTyped)

Swiftb0y: fyi, two of the other three comments were redundant if this would avoid strings better.

src/track/serato/tags.cpp
```diff
@@ -82,7 +83,7 @@ double SeratoTags::guessTimingOffsetMillis(
     // PR for more detailed information:
     // https://github.com/mixxxdj/mixxx/pull/2119
     double timingOffset = 0;
-    if (taglib::getFileTypeFromFileName(filePath) == taglib::FileType::MP3) {
+    if (fileType == "mp3") {
```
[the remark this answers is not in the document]
daschuer: Qt has explicit an operator for it and it is faster than using a QStringLiteral:
https://github.com/qt/qtbase/blob/5686af295c98a5ed4b0793ad0098cef76f39a2d7/src/corelib/text/qstring.h#L1376

src/sources/metadatasourcetaglib.h
daschuer: We have the duality of a string typed fileType used in Soundsource classes outside of Taglib and the enum inside Taglib. We still need the "fileType" as string, because it holds the file extension in case the type has not been normalized.  This is a left over from the legacy when Mixxx uses just the file extension.

I have moved the conversion here, to hide all implementation details of MetadataSourceTagLib and keep the taglib includes in its cpp file. 
Else it would be here: 
https://github.com/mixxxdj/mixxx/blob/d1c09b5c5006ef674cb4ab8eaa4fb617200c79d7/src/sources/soundsource.cpp#L84-L88

daschuer:
Done

src/track/taglib/trackmetadata_file.cpp
```diff
@@ -45,36 +45,70 @@ void readAudioProperties(
...
+            TypePair{"module"_L1, FileType::Mod},
+            TypePair{"nst"_L1, FileType::Mod},
+            TypePair{"wow"_L1, FileType::Mod},
+            TypePair{"s3m"_L1, FileType::S3M},
+            TypePair{"xm"_L1, FileType::XM},
+            TypePair{"dsf"_L1, FileType::DSF},
+            TypePair{"dff"_L1, FileType::DSDIFF},
+            TypePair{"dsdiff"_L1, FileType::DSDIFF}};
+
+    const auto it = std::find_if(lookupTable.cbegin(),
```
acolombier, line 108: clang-tidy

```suggestion
    const auto *const it = std::find_if(lookupTable.cbegin(),
```

src/track/serato/tags.cpp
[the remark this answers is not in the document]
Swiftb0y: good catch

src/track/taglib/trackmetadata_file.cpp
daschuer: This seems to be a false positive. I will try the explicit type.

daschuer: I give up, following clang-tidy with the pointer "cast"

daschuer:
Done

daschuer:
Did not work with MSVC so I have now decided to just silence clang_tidy

daschuer:
@acolombier Is this now ready to merge?

acolombier (approved):
LGTM!

PR #13220, opened 2024-05-07
feat: simplify waveform combobox

acolombier:
This addresses ~#6428~ #13226 

This current uses `RGB` only and the `RGB L/R` is unused.

By default, it prioritises the `GL` version if `AllShader` isn't available, is that the best approach?

https://github.com/mixxxdj/mixxx/assets/7086688/45c2a2b0-97c4-4504-aa9a-0cbcc2c45e12


JoergAtGithub:
The idea is to merge https://github.com/mixxxdj/mixxx/pull/13151 and than retire all classic GL and GLSL waveforms. 

Since https://github.com/mixxxdj/mixxx/pull/12994 we have also first waveform type specific sub options, I could imagine, that we use such an aub option for L/R support as well.

acolombier:
Are we planning to remove `MIXXX_USE_QOPENGL` in this case? 
I guess that doesn't not invalidate this PR, since this is already decoupling the rendering backend complexity away from the waveform type.
When it comes to `RGB L/R`, is it really something that people use? On LateNight default colour scheme, it just make the waveform even more rainbow looking and hard to follow. We could add a check box when `RGB` is selected, but since this task is to simplify the UX, I'm wondering if this is really necessary

acolombier:
(not sure why you edited my message?)
It doesn't quite make sense as of why there will be two separate waveform renderer tho. Arguably, if this is used, the same option should be done for the HSV waveform.
Anyway, I'm going to add another checkbox.

acolombier:
I have merged the `RGB L/R` and `RGB` in a single renderer, and introduced rendering option, which can be reused with HSV in the future, giving slip mode renderer as a free bonus. Happy to move [this commit](https://github.com/mixxxdj/mixxx/pull/13220/commits/0ace13eb51c62046f6e49b37e7f32f3e295752c4) in a dedicated PR. Would be great if @m0dB could have a quick look on that section!

acolombier:
Right, I think this is pretty much ready. The only outstanding question I have is, should I hide the `Acceleration` checkbox is not available/only available? Since I do the following for the L/R option, it might keep it consistent

JoergAtGithub:
> (not sure why you edited my message?)

I did not edit it intentional

acolombier:
Right, that should be ready for review now. 
Once #13151 (and #12641) are in, it should be straight forward to retire the legacy waveforms.

ywwg:
this is a great quality-of-life fix! thank you

src/waveform/renderers/allshader/waveformrendererrgb.cpp
```diff
@@ -123,79 +129,103 @@ void WaveformRendererRGB::paintGL() {
...
-            green = 0.f;
-            blue = 0.f;
-        } else {
-            const float normFactor = 1.f / maxComponent;
-            red *= normFactor;
-            green *= normFactor;
-            blue *= normFactor;
-        }
+        for (int chn = 0;
+                // In case we don't render individual color per channel, all the
```
m0dB, line 154: putting this comment inside the for statement doesn't help readablilty. I propose moving it outside and using 
```
int nchns = (splitLeftRight && !m_isSlipRenderer ? 2 : 1;
for (int chn = 0; chn < nchns; chn++) {
```

m0dB:
This is nice.

acolombier:
I've updated the UI and behaviour according to what #13226 mentioned. @m0dB I have already added the `High details` option so hopefully will make the integration with #13151 seamless.  

Here is the demo:

https://github.com/mixxxdj/mixxx/assets/7086688/f4cc8728-7744-4099-a152-cacaf3716b24

src/preferences/dialog/dlgprefwaveform.h
```diff
@@ -26,6 +30,12 @@ class DlgPrefWaveform : public DlgPreferencePage, public Ui::DlgPrefWaveformDlg
   private slots:
     void slotSetFrameRate(int frameRate);
     void slotSetWaveformType(int index);
+    void slotSetWaveformEnabled(bool checked);
+    void slotSetWaveformAcceleration(bool checked);
+#ifdef MIXXX_USE_QOPENGL
+    DECL_SLOT_WAVEFORM_OPTION(SplitStereoSignal);
+    DECL_SLOT_WAVEFORM_OPTION(HighDetails);
+#endif
```
Swiftb0y, line 38: I don't think it makes sense to generate these here. It would be better to have a single 
```cpp
slotSetWaveformOptions(allshader::WaveformRendererSignalBase opts, checked) {
    // pseudocode
    if (checked) {
        m_options |= opts;
    } else {
        m_options &= !opts;
    }
}
```

Then use the functor `connect` calls to partially bind parameters in the lambda ([avoid `std::bind`](https://abseil.io/tips/108))

```cpp
connect(..., [this](bool checked){slotSetWaveformOptions(SplitStereoSignal, checked)}
```

src/waveform/renderers/allshader/waveformrenderersignalbase.h
```diff
@@ -15,6 +15,12 @@ class WaveformRendererSignalBase;
 class allshader::WaveformRendererSignalBase : public ::WaveformRendererSignalBase,
                                               public allshader::WaveformRendererAbstract {
   public:
+    enum Options {
+        None = 0,
+        SplitStereoSignal = 0x1,
+        HighDetails = 0x2,
+    };
```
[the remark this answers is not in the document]
acolombier: I also prefer binary notation, but I know it usually doesn't make consensus. Is it an establish way to make flags in Mixxx?
Will make it a `QFlags`

src/preferences/dialog/dlgprefwaveform.h
acolombier: What is the benefit from that approach? I tend to dislike the use of lambda-connect when possible as they usually make things more opaque IMHO. For me, atomic slots are easier to read than generic one with a lambda in between.

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -27,9 +51,12 @@ DlgPrefWaveform::DlgPrefWaveform(
...
     // update the combobox.
-    QVector<WaveformWidgetAbstractHandle> handles = factory->getAvailableTypes();
-    for (int i = 0; i < handles.size(); ++i) {
-        waveformTypeComboBox->addItem(handles[i].getDisplayName(), i);
+    QVector<WaveformWidgetAbstractHandle> types = factory->getAvailableTypes();
+    for (int i = 0; i < types.size(); ++i) {
+        if (types[i].getType() == WaveformWidgetType::Empty) {
+            continue;
+        }
+        waveformTypeComboBox->addItem(types[i].getDisplayName(), types[i].getType());
```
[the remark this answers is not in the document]
acolombier: I'll be honest with you, I have no idea what those handle are here for; my gathering is that this initially, `QVector` contains a list of all `WaveformWidgetType` with additional metadata about it. I'm not quite sure why this was ever done, vs. using a hash or even perform capability discovery just in time as it appears that this metadata is only used once or a twice anyway, so no point on caching it.

acolombier (commented):
Thanks for the thorough review @Swiftb0y !

I'd like to bring your attention on the fact that I tried to limit the amount of changes done in the factory. I know a few of your comment are about modernising the codebase, but I would appreciate to keep this PR focused on the (IMO already risky) UI and UX change. 
As we mentioned in the call earlier, we are going for the long game anyway on making waveform better, so definitely happy to take some notes in issues and do some follow up refactoring on that old nasty code if that's okay with you!

src/waveform/renderers/allshader/waveformrendererrgb.cpp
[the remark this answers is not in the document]
m0dB: Yes, I already did. The conditional drawing of split L/R and subsequent reduction of code duplication looks good to me. It will be marginally less efficient due to the additional branching inside the for loop, but I feel that the reduction of code duplication justifies this.

src/waveform/renderers/allshader/waveformrenderersignalbase.h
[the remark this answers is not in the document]
Swiftb0y: no consensus afaik.

src/preferences/dialog/dlgprefwaveform.h
Swiftb0y: it avoids the macro and it makes the code more readable because I can grep for the function if I need to. Otherwise I'd need to rely on my editors "jump to definition" functionality. ["Official Guidance"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es30-dont-use-macros-for-program-text-manipulation) also discourages this style because its a bad tradeoff. 

Can you elaborate a bit on what you mean with atomic vs generic slots?

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -162,18 +207,39 @@ DlgPrefWaveform::~DlgPrefWaveform() {
...
     WaveformWidgetFactory* factory = WaveformWidgetFactory::instance();
 
+    bool isAccelerationEnabled = false;
     if (factory->isOpenGlAvailable() || factory->isOpenGlesAvailable()) {
         openGlStatusData->setText(factory->getOpenGLVersion());
+        useAccelerationCheckBox->setEnabled(true);
+        isAccelerationEnabled = m_pConfig->getValue(
+                                        ConfigKey("[Waveform]", "use_hardware_acceleration"),
+                                        factory->preferredBackend()) !=
+                WaveformWidgetBackend::None;
```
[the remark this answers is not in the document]
acolombier: I think you are misinterpreting the code slightly :) 
Here a comment about `use_hardware_acceleration`
```
    // On the UI, hardware acceleration is a boolean (0 => software rendering, 1
    // => hardware acceleration), but in the setting, we keep the granularity so
    // in case of issue when we release, we can communicate workaround on
    // editing the INI file to target a specific rendering backend. If no
    // complains come back, we can convert this safely to a backend eventually.
```
This means that `m_pConfig->getValue()` will return an integer, which is expected to be a `WaveformWidgetBackend`. The idea I had in mind when I did this was to still keep the field core behaviour as a toggle/boolean, but also allow anyone to change the INI to actually force another rendering backend, not available via the UI (which only all Allshader - accelerated - or software/Qt - fallback). This is here as a safety measure so if someone report an issue during the beta we can easily compare behaviour and looks but also instruct them with a workaround till the fix becomes available.

Swiftb0y:
> I'd like to bring your attention on the fact that I tried to limit the amount of changes done in the factory. I know a few of your comment are about modernising the codebase, but I would appreciate to keep this PR focused on the (IMO already risky) UI and UX change.

I fully agree. I was only asking you to roughly scope the modernization to the lines you're touching already (such as the new functions you created). If the scope is not clear enough, I can also send you a PR with the changes I'm suggesting.

src/preferences/dialog/dlgprefwaveform.cpp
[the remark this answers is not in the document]
Swiftb0y: yeah seems weird. `WaveformWidgetAbstractHandle` was also last touched 12 years ago...

acolombier:
> I was only asking you to roughly scope the modernization to the lines you're touching already (such as the new functions you created)

It's unfortunately not that simple everywhere. Example with `unique_ptr` on new methods, the allocated widgets get passed around in a way which I can't guarantee who is in the fact owner of this data anymore. (as it currently with `WaveformWidgetFactory::createWaveformWidget`) 
I agree this is a bad state to be in, but I guess as @ywwg suggested, perhaps we could accept that the code will get worse before it gets better?

src/preferences/dialog/dlgprefwaveform.h
acolombier: What I meant by atomic is the fact that Qt will connect the signal and the slot with intermediary transformation. 
```
connect(checkbox, QCheckBox::toggled, mywaveform, &Pref::slotSetWaveformAcceleration);
```
This way you don't need to look at the function itself as is performing a single atomic operation, which is flip a bit.

> ["Official Guidance"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es30-dont-use-macros-for-program-text-manipulation) also discourages this style because its a bad tradeoff.

I'm slightly confused about that since:
1. Qt uses macros everywhere.
2. We've introduces macros with some of the latest QML feature that were merged.

src/preferences/dialog/dlgprefwaveform.cpp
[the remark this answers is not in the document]
acolombier: Added in #13245

src/preferences/upgrade.cpp
```diff
@@ -514,15 +496,21 @@ UserSettingsPointer Upgrade::versionUpgrade(const QString& settingsPath) {
...
+            upgradeToAllShaders(
+                    (WaveformWidgetType::Type*)&waveformType,
+                    (WaveformWidgetBackend::Backend*)&waveformBackend);
             config->set(ConfigKey("[Waveform]", "WaveformType"),
-                    ConfigValue(upgradeToAllShaders(
-                            static_cast<WaveformWidgetType::Type>(
-                                    waveformType))));
+                    ConfigValue(waveformType));
+            config->set(ConfigKey("[Waveform]", "use_hardware_acceleration"),
+                    ConfigValue(waveformBackend));
```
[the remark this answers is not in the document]
acolombier: Added in #13245

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -863,283 +862,318 @@ void WaveformWidgetFactory::slotFrameSwapped() {
...
+#ifdef MIXXX_USE_QOPENGL
+    return new GLVSyncTestWidget(viewer->getGroup(), viewer);
+#else
+    return new QtVSyncTest(viewer->getGroup(), viewer);
+#endif
+}
+
 WaveformWidgetAbstract* WaveformWidgetFactory::createWaveformWidget(
         WaveformWidgetType::Type type, WWaveformViewer* viewer) {
     WaveformWidgetAbstract* widget = nullptr;
```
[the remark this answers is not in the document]
acolombier: Added in #13245

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -1205,58 +1239,57 @@ void WaveformWidgetFactory::getAvailableVSyncTypes(QList<QPair<int, QString>>* p
...
-    int index = -1;
     for (int i = 0; i < m_waveformWidgetHandles.size(); i++) {
         const WaveformWidgetAbstractHandle& handle = m_waveformWidgetHandles[i];
         if (handle.m_type == type) {
-            index = i;
+            return i;
         }
     }
-    return index;
+    return -1;
```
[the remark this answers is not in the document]
acolombier: Added in #13245

src/waveform/waveformwidgetfactory.h
```diff
@@ -231,6 +275,13 @@ class WaveformWidgetFactory : public QObject, public Singleton<WaveformWidgetFac
     GuiTick* m_pGuiTick;  // not owned
     VisualsManager* m_pVisualsManager;  // not owned
 
+    WaveformWidgetAbstract* createFilteredWaveformWidget(WWaveformViewer* viewer);
+    WaveformWidgetAbstract* createHSVWaveformWidget(WWaveformViewer* viewer);
+    WaveformWidgetAbstract* createRGBWaveformWidget(WWaveformViewer* viewer);
+    WaveformWidgetAbstract* createStackedWaveformWidget(WWaveformViewer* viewer);
+    WaveformWidgetAbstract* createSimpleWaveformWidget(WWaveformViewer* viewer);
+    WaveformWidgetAbstract* createVSyncTestWaveformWidget(WWaveformViewer* viewer);
```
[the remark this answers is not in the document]
acolombier: Added in #13245

src/waveform/widgets/waveformwidgettype.h
```diff
@@ -5,28 +5,26 @@ class WaveformWidgetType {
...
-        Count_WaveformwidgetType   //    Also used as invalid value
+        Empty = 0,
+        Simple,                  // 5  Simple GL
+        Filtered,                // 7  Filtered GLSL
+        HSV,                     // 8  HSV
+        VSyncTest,               // 9  VSync GL
+        RGB,                     // 12 RGB GLSL
+        Stacked,                 // 16 RGB Stacked
+        Count_WaveformWidgetType //    Also used as invalid value
+    };
```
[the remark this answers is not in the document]
acolombier: It appears I cannot use `Q_ENUM` as `WaveformWidgetType` is not a `QObject`. Adding to #13245 to refactors all these enum with a similar style than `constants` NS in #12798

src/preferences/dialog/dlgprefwaveform.h
Swiftb0y: > This way you don't need to look at the function itself as is performing a single atomic operation, which is flip a bit.

I don't think thats apparent or given from the name. It also applies to a hypothetical

```cpp
connect(checkbox, QCheckBox::toggled, mywaveform, [](bool checked) {
    slotSetWaveformAcceleration(checked);
});
```
Anyways. If you don't like lambdas used here, wdyt about the following compromise: Still implement a single `setWaveformOptions` that essentially contains the code thats currently in the macro. Introduce `slotSetWaveformSplitStereoSignal` as a member function, that just does `slotSetWaveformOptions(SplitStereoSignal, checked)`? The `QObject::connect` call stays the same, in the readable style that you prefer and we still avoid the macro. After all we've had two distinct discussions until now, `connect` with lambda and avoiding the macro, which are basically orthogonal.

> Qt uses macros everywhere.

Yeah, unfortunately, because Qt is Qt and bound to the APIs they came up 25 years ago. 

> We've introduces macros with some of the latest QML feature that were merged.

Yes, in that case the tradeoff was clearly worth it since it avoided boilerplate that was otherwise not possible to avoid. In this case there is almost no extra boilerplate, so the tradeoff tips to the other side.

src/preferences/dialog/dlgprefwaveform.cpp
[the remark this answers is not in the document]
Swiftb0y: Yeah I only saw that comment later and forgot to delete this one again. Sorry for the noise :)

src/waveform/widgets/waveformwidgettype.h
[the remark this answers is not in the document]
Swiftb0y: Yeah, you'll need to wrap it in a Q_NAMESPACE. postponing is fine.

src/preferences/dialog/dlgprefwaveform.h
acolombier: Fair enough, I'll make the suggested change

ywwg:
A good way to manage incoming mess is to file issues and link them in code.  so like

`// TODO([link to issue]): This should be cleaned up by...`

acolombier:
https://github.com/mixxxdj/mixxx/assets/7086688/ed71f320-e448-41cf-aed8-980fc37e80bb

Note that I have added support for stereo split in RGB high details and slip animation so we have some consistency on the waveform type. I'll add the slip animation in follow up for other types

res/shaders/rgbsignal.frag
```diff
@@ -5,6 +5,7 @@ uniform vec4 axesColor;
 uniform vec4 lowColor;
 uniform vec4 midColor;
 uniform vec4 highColor;
+uniform bool splitStereoSignal;
```
acolombier, line 8: @m0dB do you what happens if the shader program doesn't bind the uniform? Is there a consistent default value? Just wondering what to do in the deprecated renderer

acolombier:
> A good way to manage incoming mess is to file issues and link them in code. 

I have added `TODO` on new code, the rest feels counter intuitive to put a TODO on something I didn't write in a first place

---

I think this is pretty much ready for review. As you might have noticed, that PR sized snowballed to 3k line changes, so please try to keep in mind the iterations we discussed about today, and lets address legacy code quality in a follow up PRs asap :)

ywwg:
> I have added `TODO` on new code, the rest feels counter intuitive to put a TODO on something I didn't write in a first place

yeah I hear you! We do want to avoid "you touched the file, now you are responsible for it" but I think it's worthwhile to think about "we are looking at this file for the first time in a long time, let's not forget about it right away."

src/waveform/renderers/allshader/waveformrenderersignalbase.h
```diff
@@ -15,6 +16,13 @@ class WaveformRendererSignalBase;
 class allshader::WaveformRendererSignalBase : public ::WaveformRendererSignalBase,
                                               public allshader::WaveformRendererAbstract {
   public:
+    enum Option {
+        None = 0b0,
+        SplitStereoSignal = 0b1,
+        HighDetails = 0b10,
```
ywwg, line 22: In American English, this should just be "High Detail", not plural Details.  (High modifies "the level of detail" as an adjective. It is ok to say "More detail" and also "more details", but with "high", only the singular form sounds right.)

ywwg (changes requested):
Played around with this, and it works as expected.  A few usage notes:

I was set up to use RGB L/R High Detail, but this change defaulted me to plain RGB.  We should try to preserve the users' existing choices where possible. Though, I am not sure if I had some other patch applied that messed up that migration.

I think vsynctest should not be available in Release builds, or maybe it should be its own special checkbox at the bottom of the screen. It could be very weird and distracting for regular users.

acolombier:
>  this change defaulted me to plain RGB. We should try to preserve the users' existing choices where possible. Though, I am not sure if I had some other patch applied that messed up that migration.

Could you please share the value you had in mixxx.cfg before running that branch? I'll had some tests to make sure it handles migration correctly.

> I think vsynctest should not be available in Release builds

AFAIK vsynctest is included if you run with --developer. I'll make sure that remains the case

m0dB:
I have created a PR on top of this PR (I hope that works, git wise) to get rid of all the code duplication in the allshader waveformwidget, reducing it to a single one + some cleanup of waveformwidgetfactory.

https://github.com/acolombier/mixxx/pull/2

src/test/waveform_upgrade_test.cpp
```diff
@@ -0,0 +1,173 @@
...
+TEST_F(UpgradeTest, useCorrectWaveformType) {
+    struct test_case {
+        QString waveformName;
+        int oldTypeId;
+        WaveformWidgetType::Type expectedType;
+        WaveformWidgetBackend expectedBackend;
+        allshader::WaveformRendererSignalBase::Options expectedOptions;
+    };
+
+    QList<test_case> testCases = {
```
acolombier, line 28: Spec available [here](https://github.com/mixxxdj/mixxx/issues/13226#issuecomment-2124611841)

ywwg:
> I hope that works, git wise

that should work, yes

acolombier:
I have hidden the acceleration checkbox on Mac. Note that it will still be show if running with `--developer` or if the acceleration backend isn't the recommended one (all-shader, happens if you override the setting in INI)

@m0dB could you confirm it hides properly?

acolombier:
> Not entirely, the combobox is a bit smaller than the other full-width comboboxes.

Which backend were you using? All-shader? Otherwise, it is expected it shows it.

> we might as well enable them?

As our Mac developer, I'll follow your call the matter! :) 
We mentioned in our sync Monday that macs have a fairly unified set of hardware, which all support the all-shader one, so I thought it might make sense to default to all shader and simplify the UI, but I also like the unified UI across platform so happy either way.

> We don't seem to have stacked and simple software waveform types though. I guess they should be disabled in the combobox when non-accelerated is selected?

I went the other way around and made the combobox leading the decision with I think is a better UX: if you selected `Simple` or `Stacked` it will enable the checkbox (enable hardware) and disable the option. This way, a user can understand no software fallback is available but also don't have to toggle the checkbox first before selecting the new waveform type.
I guess I could add a tooltip too?

acolombier:
@Mixxx Core Team could we look at merging that PR ASAP, and potentially to a thorough retroactive review? I agree to address comments and concerns in one or more follow up PRs, but this would help @m0dB to stop that PR to grow out of proportion

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -895,323 +851,295 @@ void WaveformWidgetFactory::slotFrameSwapped() {
...
+template<typename WaveformT>
+WaveformWidgetVars waveformWidgetVars() {
+    WaveformWidgetVars result;
+    result.m_useGL = WaveformT::useOpenGl();
+    result.m_useGLES = WaveformT::useOpenGles();
+    result.m_useGLSL = WaveformT::useOpenGLShaders();
+    result.m_category = WaveformT::category();
+
+    return result;
+}
```
Swiftb0y, line 915: could be constexpr

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -895,323 +851,295 @@ void WaveformWidgetFactory::slotFrameSwapped() {
...
 #else
-            setWaveformVarsByType.operator()<allshader::LRRGBWaveformWidget>();
-            break;
+    QHashIterator<WaveformWidgetType::Type,
+            QList<WaveformWidgetBackend>>
+            handleIter(collectedHandles);
+    while (handleIter.hasNext()) {
+        handleIter.next();
+        const auto& type = handleIter.key();
+        const auto& backends = handleIter.value();
```
Swiftb0y, line 979: thinking forward-compatible, I like it :+1: :sunglasses:

src/waveform/widgets/waveformwidgettype.h
```diff
@@ -1,36 +1,41 @@
...
+            WaveformWidgetType::Simple,
+            WaveformWidgetType::Filtered,
+            WaveformWidgetType::HSV,
+            WaveformWidgetType::VSyncTest,
+            WaveformWidgetType::RGB,
+            WaveformWidgetType::Stacked,
     };
 };
+
+enum class WaveformWidgetBackend : int {
```
Swiftb0y, line 34: I think we have the rule somewhere that you shouldn't specify the underlying type unless you have to.

Swiftb0y (commented):
Thank you, mostly nitpicks to be addressed some other time, only thing that needs work is the upgrade code which currently looks like a bug to me.

src/preferences/dialog/dlgprefwaveform.h
```diff
@@ -5,6 +5,14 @@
...
 #include "preferences/dialog/ui_dlgprefwaveformdlg.h"
 #include "preferences/usersettings.h"
+#include "waveform/widgets/waveformwidgettype.h"
+#ifdef MIXXX_USE_QOPENGL
+#include "waveform/renderers/allshader/waveformrenderersignalbase.h"
+#define DECL_SLOT_WAVEFORM_OPTION(opt)                                               \
+    void slotSetWaveformOption##opt(bool checked) {                                  \
+        slotSetWaveformOptions(allshader::WaveformRendererSignalBase::opt, checked); \
+    }
+#endif
```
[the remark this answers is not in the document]
ywwg: I also dislike macros, they obfuscate and make it harder to see what's happening.  sometimes boilerplate is boilerplate.  (But I will also not block on this)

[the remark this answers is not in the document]
ywwg: also this is only used a few times so it doesn't seem that helpful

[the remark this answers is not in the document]
acolombier: Yeah I don't mind just transforming to code. I just thought, as we discussed on the sync, we might want to add more option in the near future, and wanted to introduce some kinda establish way to handle option, like [done here for metadata](https://github.com/mixxxdj/mixxx/blob/6f3aea012ab6dee632b3fa1311f522e3b9a5bcb6/src/track/trackrecord.h#L36)

> I won't insist. Just sharing my opinion

Honestly don't mind either way :) I was just confused because sometime, that seem to be the preferred way, including on some recently merged code. 
I guess it would be nice to update the [wiki guideline](https://github.com/mixxxdj/mixxx/wiki/Coding-Guidelines) about macros, so it is clearer to me when and when to use it.

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -159,21 +187,73 @@ DlgPrefWaveform::DlgPrefWaveform(
...
+                    ConfigKey("[Waveform]", "use_hardware_acceleration"),
+                    factory->preferredBackend()) ==
+                    factory->preferredBackend()) {
+        useAccelerationCheckBox->hide();
+        waveformSpacer->changeSize(0, 20);
+    } else {
+        useAccelerationCheckBox->show();
+        waveformSpacer->changeSize(20, 20);
     }
+#endif
```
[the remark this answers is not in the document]
acolombier: No, this was to address the requirement in [that comment](https://github.com/mixxxdj/mixxx/issues/13226#issuecomment-2124611841)
> There is no software fallback on Mac, and no need for it. Option should be hidden

But I think @m0dB has just restored software renderer on Mac so I will remove this

src/waveform/waveformwidgetfactory.cpp
acolombier: Adding to #13245

src/waveform/widgets/waveformwidgettype.h
acolombier: I had to for the static_cast in upgrade, but with your suggested approach there, I will see if I can remove it.

src/preferences/dialog/dlgprefwaveform.h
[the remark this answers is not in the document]
ywwg: good idea, and I see what you mean about it being used elsewhere.  A couple things I would say:
1. mixxx is super inconsistent in style, despite our best efforts.  Between the evolution of c++ over the past 20 years and the large number of disparate contributors, things are quite messy. Sorry!
2. This macro to me falls into YAGNI ("you ain't gonna need it") territory.  If we get to the point where there are a ton of options funcs and it gets overwhelming, then we can refactor and introduce the macro.  But I don't believe it makes sense to introduce it now with the assumption that we *might* or even *probably* will want it, because we may not :). Only write it when you actually do need it.

This comes down to "code smell" and personal preference, which is why comments like this are non-blocking. but imho, macros are very rarely worth it.

src/waveform/widgets/waveformwidgettype.h
Swiftb0y: yeah. I think you can see how specifying the enums underlying type is a code smell because it allows you to do ugly stuff elsewhere.

src/preferences/dialog/dlgprefwaveform.cpp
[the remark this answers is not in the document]
Swiftb0y: > No, this was to address the requirement in https://github.com/mixxxdj/mixxx/issues/13226#issuecomment-2124611841

A right, in that case the ifdef would've been okay IMO. but yeah, should be removed now that the fallback has been restored.

acolombier:
I think we got everything in order now

src/preferences/upgrade.cpp
```diff
@@ -33,78 +35,107 @@ Upgrade::~Upgrade() {
...
+        WaveformWidgetBackend,
+        allshader::WaveformRendererSignalBase::Options>
+upgradeToAllShaders(int unsafeWaveformType,
+        int unsafeWaveformBackend,
+        int unsafeWaveformOption) {
     // TODO: convert `WaveformWidgetType::Type` to an enum class then shorten more `using enum ...`
     using WWT = WaveformWidgetType;
 
+    if (static_cast<WaveformWidgetBackend>(unsafeWaveformBackend) ==
+            WaveformWidgetBackend::AllShader) {
```
Swiftb0y, line 48: technically, this needs to be the other way around (convert an int, not an enum)...

```suggestion
    if (static_cast<int>(WaveformWidgetBackend::AllShader) == unsafeWaveformBackend) {
            
```

src/preferences/upgrade.cpp
```diff
@@ -33,78 +35,107 @@ Upgrade::~Upgrade() {
...
+        allshader::WaveformRendererSignalBase::Options waveformOption =
+                static_cast<allshader::WaveformRendererSignalBase::Options>(
+                        unsafeWaveformOption) &
+                allshader::WaveformRendererSignalBase::All;
```
Swiftb0y, line 52: I assume is `allshader::WaveformRendererSignalBase::All` means all options, right? This should be reflected in the naming (use `enum class`) in a future PR.

src/preferences/upgrade.cpp
```diff
@@ -33,78 +35,107 @@ Upgrade::~Upgrade() {
...
+        switch (unsafeWaveformType) {
+        case WWT::Simple:
```
Swiftb0y, line 54: this'll also break for enum class, but lets postpone that to #13245

src/preferences/upgrade.cpp
```diff
@@ -555,34 +590,24 @@ UserSettingsPointer Upgrade::versionUpgrade(const QString& settingsPath) {
...
+                correctedWaveformBacked,
+                correctedWaveformOption] =
+                upgradeToAllShaders(
+                        waveformType, waveformBackend, waveformOption);
+        config->setValue(ConfigKey("[Waveform]", "WaveformType"),
+                correctedWaveformType);
+        config->setValue(ConfigKey("[Waveform]", "use_hardware_acceleration"),
+                correctedWaveformBacked);
+        config->setValue<int>(ConfigKey("[Waveform]", "waveform_options"),
+                correctedWaveformOption);
```
Swiftb0y, line 610: very clean IMO :heart_eyes: 
A little redundant, but it doesn't get much more elegant in C++ I think.

Swiftb0y (approved):
LGTM. Any1 else wanting to take a look (just in case I missed any grave errors, not for trivial stuff like I nitpicked in this round)?

src/preferences/upgrade.cpp
acolombier: It's new code so I'm fixing it now

acolombier: Updated now - is `AllOptionsCombined` good enough?

Swiftb0y: Yeah.

ronso0:
I didn't look into the ui file in detail, just want to share a few thoughts on the UI/UX:
* I think it'd look better if Acceleration, Stereo and High Details are in one row
* the label "This function requires ..." should only be shown if waveforms are **enabled** and acceleration is **disabled**
* for consistency, all waveform-related options should be greyed out (end-of-track, play position, ...)
* while working on #13273 I noticed we should move the overview options into their own section, because if we do disable all waveform options, the overview options are in between those disabled options which is a bit confusing IMO
Maybe move the Overview options to the top so they're more discoverable than below the Waveform option?
Maybe use QGroupBoxes for waveform and overview?

I'll check if I can provide some demo commits tomorrow.

acolombier:
Thanks for your input @ronso0 
Are you okay with this being improved in a follow up PR?

ronso0:
Sure.

src/preferences/upgrade.cpp
```diff
@@ -515,36 +577,46 @@ UserSettingsPointer Upgrade::versionUpgrade(const QString& settingsPath) {
...
-        config->set(ConfigKey("[Waveform]", "FrameRate"), ConfigValue(60));
-
+        int waveformType =
+                config->getValue<int>(ConfigKey("[Waveform]", "WaveformType"));
+        // values might be out of range for the enum, avoid undefined
+        // behavior by not casting to the enum type just yet.
+        int waveformBackend = config->getValue<int>(
+                ConfigKey("[Waveform]", "use_hardware_acceleration"));
+        int waveformOption = config->getValue<int>(ConfigKey("[Waveform]", "waveform_options"));
+        auto [correctedWaveformType,
```
Swiftb0y, line 599: I think this needs more work now that this has slipped to 2.6. 

I think we need to keep the old 2.4 code and then add this for 2.6. Whats your opinion @daschuer?

ywwg: yes I do think we bumped this to 2.6

Swiftb0y:
@acolombier do you have time to update the `upgrade.cpp` code? Then we can merge.

Swiftb0y:
also @ywwg please double check [your review](https://github.com/mixxxdj/mixxx/pull/13220#pullrequestreview-2067188011) and approve/dismiss so we can get this cleared formally.

acolombier:
I should get the time sometime today or tomorrow hopefully.

ronso0:
This has really finally been bumped to 2.6?
I though there was the goal to clean this up for 2.5

acolombier:
> This has really finally been bumped to 2.6? I though there was the goal to clean this up for 2.5

This is only the first part of the clean up, more is to come. I'm afraid it might not be ready in time and we probably don't want to delay 2.5 for that

ywwg (approved):
lgtm!

ywwg:
(I am not sure if we have to explicitly wait for @m0dB and @ronso0 since they did not make must-address comments? I am still unsure how we handle multi-reviewer PRs.  It sounded like Ronso0 was lgtm)

acolombier:
Ideally, I just would like confirmation that [this](https://github.com/mixxxdj/mixxx/pull/13220/files#r1607224190) is fine from @m0dB. Alternatively, I think it is small enough to be fixed in later PRs if it isn't.

Swiftb0y (approved):

Swiftb0y (changes requested):
Yeah. so all the additional refactorings here leave this in a difficult state. The software fallback waveforms of course work, but the hardware waveforms sometimes crash my gnome-shell completely. That has already been the case previously (with the legacy GLSL waveforms IIRC), but now it has become more difficult to tell when this happens. also the `Stacked` and the `Simple` waveforms don't work at all (though not producing any crashes).

This is on fedora 39.

acolombier:
Could you provide more information? How to reproduce? What is the crash? Message log? Are you able to reproduce this on main using the same allshader waveform?

Swiftb0y:
Whoops. I'm sorry for the false alarm. I did some more digging and I'm pretty sure I suffered from https://gitlab.gnome.org/GNOME/mutter/-/issues/3462. After carefully ensuring I've upgraded to gnome 46.2 all the waveforms work again (at least without crashes and other regressions).

Swiftb0y (approved):
Did some tests, now that mutter fixed the issue, it works well

Swiftb0y:
Thank you for your continued efforts btw @acolombier

PR #13229, opened 2024-05-12
fix: add missing member in copy ctor

acolombier:


daschuer (approved):
LGTM, Thank you for noticing.

PR #12067, opened 2023-10-06
Update to lates VCPKG environment.

daschuer:
This uses the new environment created here: 
https://github.com/mixxxdj/vcpkg/pull/88

Among others it updates to Qt 6.5.3 and FFmpeg 6.0. See the origin PR for a complete list.

PR #13233, opened 2024-05-13
Upgrade to libdjinterop 0.21.0

mr-smidge:


mr-smidge:
The build errors confuse me.  From the macOS build:

```
Undefined symbols for architecture x86_64:
  "djinterop::crate::create_sub_crate(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >)", referenced from:
      mixxx::EnginePrimeExportJob::run() in libmixxx-lib.a(engineprimeexportjob.cpp.o)
  "djinterop::database::create_root_crate(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >) const", referenced from:
      mixxx::EnginePrimeExportJob::run() in libmixxx-lib.a(engineprimeexportjob.cpp.o)
```

These two functions changed their signature from accepting `std::string` by value to const ref instead.  The header and source files are internally consistent within libdjinterop in the [0.21.0 tag](https://github.com/xsco/libdjinterop/commit/f34a8fb6323236d73c4f03539ef941dc58763aff).

Is there something cached in these builds that would somehow pick up the old signature?

daschuer:
> Is there something cached in these builds that would somehow pick up the old signature?

Yes, the vcpkg build environment has probably the headers of the old version. We need to update this anyway, because the windows build fails with the downloaded source for some reasons.

mr-smidge:
> Yes, the vcpkg build environment has probably the headers of the old version. We need to update this anyway, because the windows build fails with the downloaded source for some reasons.

Ok, I'm not sure how to resolve this - @daschuer are you ok to make (or show me how to make) the necessary changes to get this through?

mr-smidge:
PR to for libdjinterop 0.21.0 in vcpkg: https://github.com/microsoft/vcpkg/pull/38888

daschuer:
Great, thank you. Now we need to cherry pick it to our own vcpkg repro. To the 2.4 and 2.5 branch.

mr-smidge:
Here we go - hopefully I've done this right:

* 2.4 branch - https://github.com/mixxxdj/vcpkg/pull/144
* 2.5 branch - https://github.com/mixxxdj/vcpkg/pull/145

acolombier:
I think you will have to rebase your branch against `main` since the attributed milestone for that PR is `2.5`. This means you shouldn't need a VPKG update for 2.4- @daschuer do you confirm?

daschuer:
The 2.4 target is correct, because 2.4 and main builds are broken.

mr-smidge:
@daschuer Now that the vcpkg PRs are merged, please can you re-run the failed actions?  I don't appear to have permission to do this.  Thanks!

JoergAtGithub:
I doubt that the rerun will be enough, because we need to adjust the buildenv hashes before.

mr-smidge:
> I doubt that the rerun will be enough, because we need to adjust the buildenv hashes before.

I'm not familiar with what this means, but the Windows build is already failing with a similar error to before, so I suspect you're right.  What steps need to be taken now?

JoergAtGithub:
You need to do similar changes as in #12067 , otherwise the Mixxx build will not download the new buildenv zip files.

daschuer:
I have cherry-picked this to https://github.com/mixxxdj/mixxx/pull/13288 https://github.com/mixxxdj/mixxx/pull/13289

daschuer:
Thank you for your help.

PR #13236, opened 2024-05-14
Refactor/modernize shrink scopedtimer

Swiftb0y:
first commit just modernizes it a bit, and doesn't touch the API, the second commit goes a step further and leverages `QString::arg(Args&&...)` to provide arbitrary formastring support. let me know what you think.

### DIFF
--- a/src/util/qstringformat.h
+++ b/src/util/qstringformat.h
@@ -1,0 +1,42 @@
+#pragma once
+
+#include <QString>
+#include <type_traits>
+#include <utility>
+
+namespace {
+
+// taken from Qt
+template<typename T>
+static constexpr bool is_convertible_to_view_or_qstring_v =
+        std::is_convertible_v<T, QString> ||
+        std::is_convertible_v<T, QStringView> ||
+        std::is_convertible_v<T, QLatin1String>;
+
+// check if we can call QString::number(T) with T
+template<typename T>
+static constexpr bool is_number_compatible_v =
+        std::is_invocable_v<decltype(QString::number(std::declval<T>()))(T), T>;
+
+// always false, used for static_assert and workaround for compilers without
+// https://cplusplus.github.io/CWG/issues/2518.html
+template<typename T>
+static constexpr bool always_false_v = false;
+} // namespace
+
+// Try to convert T to a type that would be accepted by QString::args(Args&&...)
+template<typename T>
+auto convertToQStringConvertible(T&& arg) {
+    if constexpr (is_convertible_to_view_or_qstring_v<T>) {
+        // no need to do anything, just return verbatim
+        return std::forward<T>(arg);
+    } else if constexpr (is_number_compatible_v<T>) {
+        return QString::number(std::forward<T>(arg));
+    } else {
+        static_assert(always_false_v<T>, "Unsupported type for QString::arg");
+        // unreachable, but returning a QString results in a better error message
+        // because the log won't be spammed with all the QString::arg overloads
+        // it couldn't match with `void`.
+        return QString();
+    }
+}
--- a/src/util/timer.h
+++ b/src/util/timer.h
@@ -3,6 +3,7 @@
 #include <QObject>
 #include <QString>
 #include <optional>
+#include <utility>
 
 #include "control/controlproxy.h"
 #include "util/cmdlineargs.h"
@@ -54,43 +55,39 @@ class SuspendableTimer : public Timer {
 
 class ScopedTimer {
   public:
-    ScopedTimer(const char* key, int i, Stat::ComputeFlags compute = kDefaultComputeFlags)
+    // Allows the timer to contain a format string which is only assembled
+    // when we're not in `--developer` mode.
+    /// @param compute Flags to use for the Stat::ComputeFlags (can be omitted)
+    /// @param key The format string for the key
+    /// @param args The arguments to pass to the format string
+    template<typename... Ts>
+    ScopedTimer(Stat::ComputeFlags compute, const QString& key, Ts&&... args)
             : m_timer(std::nullopt) {
         if (CmdlineArgs::Instance().getDeveloper()) {
-            initialize(QString(key), QString::number(i), compute);
+            QString assembledKey = key;
+            if constexpr (sizeof...(args) > 0) {
+                // only try to call QString::arg when we've been given parameters
+                assembledKey = key.arg(std::forward<Ts>(args)...);
+            }
+            m_timer = std::make_optional<Timer>(assembledKey, compute);
+            m_timer->start();
         }
     }
-
-    ScopedTimer(const char* key,
-            const char* arg = nullptr,
-            Stat::ComputeFlags compute = kDefaultComputeFlags)
-            : m_timer(std::nullopt) {
-        if (CmdlineArgs::Instance().getDeveloper()) {
-            initialize(QString(key), arg != nullptr ? QString(arg) : QString(), compute);
-        }
+    template<typename... Ts>
+    ScopedTimer(const QString& key, Ts&&... args)
+            : ScopedTimer(kDefaultComputeFlags, key, std::forward<Ts>(args)...) {
     }
-
-    ScopedTimer(const char* key,
-            const QString& arg,
-            Stat::ComputeFlags compute = kDefaultComputeFlags)
-            : m_timer(std::nullopt) {
-        if (CmdlineArgs::Instance().getDeveloper()) {
-            initialize(QString(key), arg, compute);
-        }
+    // for lazy users that don't wrap the key in a QStringLiteral(...)
+    template<typename... Ts>
+    ScopedTimer(const char* key, Ts&&... args)
+            : ScopedTimer(CmdlineArgs::Instance().getDeveloper() ? QString(key) : QString(),
+                      std::forward<Ts>(args)...) {
     }
 
     ~ScopedTimer() {
         if (m_timer.has_value()) {
             m_timer->elapsed(true);
         }
-    }
-
-    void initialize(const QString& key,
-            const QString& arg,
-            Stat::ComputeFlags compute = kDefaultComputeFlags) {
-        QString strKey = arg.isEmpty() ? key : key.arg(arg);
-        m_timer = std::make_optional<Timer>(strKey, compute);
-        m_timer->start();
     }
 
     void cancel() {
--- a/src/util/timer.h
+++ b/src/util/timer.h
@@ -65,7 +65,7 @@ class ScopedTimer {
     template<typename T, typename... Ts>
     ScopedTimer(Stat::ComputeFlags compute, T&& key, Ts&&... args)
             : m_timer(std::nullopt) {
-        static_assert(!std::is_same_v<std::remove_cvref_t<std::decay_t<T>>, char*>,
+        static_assert(!std::is_same_v<std::decay_t<T>, char const*>,
                 "string type likely not UTF-16, please pass QStringLiteral by "
                 "means of u\"key text\"_s");
         if (!CmdlineArgs::Instance().getDeveloper()) {
@@ -74,6 +74,8 @@ class ScopedTimer {
         // key is explicitly `auto` to allow passing non-QString types such as `const char*`
         // its conversion is delayed until after we've checked if we're in developer mode
         auto assembledKey = QString(std::forward<T>(key));
+        // ensure the string was indeed a literal an not unnecessarily heap-allocated
+        DEBUG_ASSERT(assembledKey.capacity() == 0);
         if constexpr (sizeof...(args) > 0) {
             // only try to call QString::arg when we've been given parameters
             assembledKey = assembledKey.arg(convertToQStringConvertible(std::forward<Ts>(args))...);
--- a/src/util/timer.h
+++ b/src/util/timer.h
@@ -3,6 +3,7 @@
 #include <QObject>
 #include <QString>
 #include <optional>
+#include <type_traits>
 #include <utility>
 
 #include "control/controlproxy.h"
@@ -64,6 +65,9 @@ class ScopedTimer {
     template<typename T, typename... Ts>
     ScopedTimer(Stat::ComputeFlags compute, T&& key, Ts&&... args)
             : m_timer(std::nullopt) {
+        static_assert(char_type_size<T>() == sizeof(QString::value_type),
+                "string type likely not UTF-16, please pass QStringLiteral by "
+                "means of u\"key text\"_s");
         if (!CmdlineArgs::Instance().getDeveloper()) {
             return; // leave timer in cancelled state
         }
@@ -92,6 +96,19 @@ class ScopedTimer {
         m_timer.reset();
     }
   private:
+    // utility function
+    template<typename T>
+    constexpr static std::size_t char_type_size() {
+        if constexpr (std::is_array_v<T>) {
+            return sizeof(std::remove_all_extents_t<T>);
+        } else if constexpr (std::is_pointer_v<T>) {
+            // assume this is some point to char array then.
+            return sizeof(std::remove_pointer_t<T>);
+        } else {
+            // assume this is some QString or std::string type
+            return sizeof(typename T::value_type);
+        }
+    }
     // nullopt also counts as cancelled
     std::optional<Timer> m_timer;
 };
--- a/src/util/timer.h
+++ b/src/util/timer.h
@@ -65,7 +65,7 @@ class ScopedTimer {
     template<typename T, typename... Ts>
     ScopedTimer(Stat::ComputeFlags compute, T&& key, Ts&&... args)
             : m_timer(std::nullopt) {
-        static_assert(char_type_size<T>() == sizeof(QString::value_type),
+        static_assert(!std::is_same_v<std::remove_cvref_t<std::decay_t<T>>, char*>,
                 "string type likely not UTF-16, please pass QStringLiteral by "
                 "means of u\"key text\"_s");
         if (!CmdlineArgs::Instance().getDeveloper()) {
@@ -96,19 +96,6 @@ class ScopedTimer {
         m_timer.reset();
     }
   private:
-    // utility function
-    template<typename T>
-    constexpr static std::size_t char_type_size() {
-        if constexpr (std::is_array_v<T>) {
-            return sizeof(std::remove_all_extents_t<T>);
-        } else if constexpr (std::is_pointer_v<T>) {
-            // assume this is some point to char array then.
-            return sizeof(std::remove_pointer_t<T>);
-        } else {
-            // assume this is some QString or std::string type
-            return sizeof(typename T::value_type);
-        }
-    }
     // nullopt also counts as cancelled
     std::optional<Timer> m_timer;
 };
--- a/src/util/timer.h
+++ b/src/util/timer.h
@@ -1,6 +1,8 @@
 #pragma once
 
 #include <QObject>
+#include <QString>
+#include <optional>
 
 #include "control/controlproxy.h"
 #include "util/cmdlineargs.h"
@@ -52,61 +54,51 @@ class SuspendableTimer : public Timer {
 
 class ScopedTimer {
   public:
-    ScopedTimer(const char* key, int i,
-                Stat::ComputeFlags compute = kDefaultComputeFlags)
-            : m_pTimer(NULL),
-              m_cancel(false) {
+    ScopedTimer(const char* key, int i, Stat::ComputeFlags compute = kDefaultComputeFlags)
+            : m_timer(std::nullopt) {
         if (CmdlineArgs::Instance().getDeveloper()) {
             initialize(QString(key), QString::number(i), compute);
         }
     }
 
-    ScopedTimer(const char* key, const char *arg = NULL,
-                Stat::ComputeFlags compute = kDefaultComputeFlags)
-            : m_pTimer(NULL),
-              m_cancel(false) {
+    ScopedTimer(const char* key,
+            const char* arg = nullptr,
+            Stat::ComputeFlags compute = kDefaultComputeFlags)
+            : m_timer(std::nullopt) {
         if (CmdlineArgs::Instance().getDeveloper()) {
-            initialize(QString(key), arg ? QString(arg) : QString(), compute);
+            initialize(QString(key), arg != nullptr ? QString(arg) : QString(), compute);
         }
     }
 
-    ScopedTimer(const char* key, const QString& arg,
-                Stat::ComputeFlags compute = kDefaultComputeFlags)
-            : m_pTimer(NULL),
-              m_cancel(false) {
+    ScopedTimer(const char* key,
+            const QString& arg,
+            Stat::ComputeFlags compute = kDefaultComputeFlags)
+            : m_timer(std::nullopt) {
         if (CmdlineArgs::Instance().getDeveloper()) {
             initialize(QString(key), arg, compute);
         }
     }
 
-    virtual ~ScopedTimer() {
-        if (m_pTimer) {
-            if (!m_cancel) {
-                m_pTimer->elapsed(true);
-            }
-            m_pTimer->~Timer();
+    ~ScopedTimer() {
+        if (m_timer.has_value()) {
+            m_timer->elapsed(true);
         }
     }
 
-    inline void initialize(const QString& key, const QString& arg,
-                Stat::ComputeFlags compute = kDefaultComputeFlags) {
-        QString strKey;
-        if (arg.isEmpty()) {
-            strKey = key;
-        } else {
-            strKey = key.arg(arg);
-        }
-        m_pTimer = new(m_timerMem) Timer(strKey, compute);
-        m_pTimer->start();
+    void initialize(const QString& key,
+            const QString& arg,
+            Stat::ComputeFlags compute = kDefaultComputeFlags) {
+        QString strKey = arg.isEmpty() ? key : key.arg(arg);
+        m_timer = std::make_optional<Timer>(strKey, compute);
+        m_timer->start();
     }
 
     void cancel() {
-        m_cancel = true;
+        m_timer.reset();
     }
   private:
-    Timer* m_pTimer;
-    char m_timerMem[sizeof(Timer)];
-    bool m_cancel;
+    // nullopt also counts as cancelled
+    std::optional<Timer> m_timer;
 };
 
 // A timer that provides a similar API to QTimer but uses render events from the
--- a/src/util/timer.h
+++ b/src/util/timer.h
@@ -10,6 +10,7 @@
 #include "util/duration.h"
 #include "util/parented_ptr.h"
 #include "util/performancetimer.h"
+#include "util/qstringformat.h"
 #include "util/stat.h"
 
 const Stat::ComputeFlags kDefaultComputeFlags = Stat::COUNT | Stat::SUM | Stat::AVERAGE |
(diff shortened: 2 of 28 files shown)


src/waveform/waveformwidgetfactory.cpp
```diff
@@ -688,8 +688,8 @@ void WaveformWidgetFactory::notifyZoomChange(WWaveformViewer* viewer) {
 }
 
 void WaveformWidgetFactory::renderSelf() {
-    ScopedTimer t("WaveformWidgetFactory::render() %1waveforms",
-            static_cast<int>(m_waveformWidgetHolders.size()));
+    ScopedTimer t("WaveformWidgetFactory::render() %1 waveforms",
+            QString::number(m_waveformWidgetHolders.size()));
```
daschuer, line 692: The idea was that all ScopeTimer statements have almost zero CPU overhead, when developer mode is disabled. 
QString does allocate memory even if the scoped timer is disabled so this is not the best idea. 

We may look for a solution that is equal fast in normal mode and faster in developer mode. This can be archived with the constexpr QLatin1String constructor. I have proposed here a Qt < 6.4 solution:   https://github.com/mixxxdj/mixxx/blob/751a104f4b21a2a74c1a9d0c1cd713ecd4605990/src/track/taglib/trackmetadata_file.cpp#L48-L54

daschuer (changes requested):
Thank you for this nice improvement.

src/waveform/waveformwidgetfactory.cpp
Swiftb0y: > The idea was that all ScopeTimer statements have almost zero CPU overhead, when developer mode is disabled.

Yes. I got that from the design and I tried to achieve the same here. This is the one case where the newer API does not exactly match the old one, but I figured this was a small-enough prize to pay in this case. There is also small-string-optimization, which should lessen the cost of the construction considerably. 

> We may look for a solution that is equal fast in normal mode and faster in developer mode. This can be archived with the constexpr QLatin1String constructor.

I'm afraid I can't quite follow you here. How does that help in the `QString::number` case?

daschuer: Just don't use QString::number. It is a heap allocating function. Does passing an int not work?

Swiftb0y: No because the `QString::arg(Args&&...)` version only accepts objects implicitly convertible to int...

I can try creating a wrapper that does also accept numeric types and calls QString::number on them, though I don't know how complex the metaprogramming for that would be.

Swiftb0y:
build all green. I think its easier to review if I squash all the fixups first. Or do you already have a review pending? @daschuer

daschuer:
Do we have a use case where we need to pass a QString as `key` or `args`? We should not allow this IMHO. The current implementation and the related comments will be tempting preferably use a QString.
I think the fastest solution is to use a `constexpr QLatin1String`

```
template<typename... Ts>
ScopedTimer(const QLatin1String& key, Ts&&... args) {
    static_assert(std::is_constexpr_v<decltype(key)>);
```

Alternatinve that does not require `""_L1` but also avoids searching \0 at runtime: 
```
template<std::size_t N, typename... Ts>
ScopedTimer(const char key[N], Ts&&... args) {
    QString k(key, N-1) 
```

(Both untested)

src/util/qstringformat.h
```diff
@@ -0,0 +1,42 @@
...
+// Try to convert T to a type that would be accepted by QString::args(Args&&...)
+template<typename T>
+auto convertToQStringConvertible(T&& arg) {
+    if constexpr (is_convertible_to_view_or_qstring_v<T>) {
+        // no need to do anything, just return verbatim
+        return std::forward<T>(arg);
+    } else if constexpr (is_number_compatible_v<T>) {
+        return QString::number(std::forward<T>(arg));
+    } else {
+        static_assert(always_false_v<T>, "Unsupported type for QString::arg");
```
ronso0, line 36: unfortunately this throws an error when picked for main (#12548), dunno if that is to be expected, let alone how to fix it (except using `static_assert(false, ...)` )
https://github.com/mixxxdj/mixxx/actions/runs/9103424016/job/25025246943?pr=12548
```
/src/util/stringformat.h: In function ‘auto convertToQStringConvertible(T&&)’:
Error: /src/util/stringformat.h:36:23: error: reference to ‘always_false_v’ is ambiguous
   36 |         static_assert(always_false_v<T>, "Unsupported type for QString::arg");
      |                       ^~~~~~~~~~~~~~
/src/util/stringformat.h:24:23: note: candidates are: ‘template<class T> constexpr const bool {anonymous}::always_false_v<T>’
   24 | static constexpr bool always_false_v = false;
      |                       ^~~~~~~~~~~~~~
```

Swiftb0y:
> Do we have a use case where we need to pass a QString as key or args? We should not allow this IMHO. The current implementation and the related comments will be tempting preferably use a QString.

Well, currently both "QLatin1String" and `const char[N]` passing works, but because we end up passing the `QString key` to the `Timer`, it sooner or later needs to become a proper `QString` anyways (or at least an owning string type).

src/util/qstringformat.h
Swiftb0y: I think its getting confused because `always_false_v` is also already defined in other headers (likely from #12781). So you should be able to fix it by shuffling some code around. Potentially just putting a single definition in a single header and then using that across both files.

daschuer:
We need to make sure that the QString creation is always delayed until the developer flag is checked to not malloc memory or issue any other locking system call during the normal run.

src/util/qstringformat.h
ronso0: Ah, okay, in `main` this seems to clash with `always_false_v` declared in midimessage.h (where it is `inline constexpr` bool)
https://github.com/mixxxdj/mixxx/blob/8f647908af460e53fe8a860b6ce0964a93a55112/src/controllers/midi/midimessage.h#L183

ronso0: (comment race condition : )

ronso0: > Potentially just putting a single definition in a single header and then using that across both files could be enough

What place would you pick?

Swiftb0y:
Yes, I'm aware. The only way that could happen is if the caller explicitly constructs a the string the wrong way for example by `QString("some literal")`. Using a literal directly works correctly (and delays construction as intended). I don't want to blacklist passing a QString because that would prevent passing a `QStringLiteral`.

daschuer:
Ah, QStringLiteral. This way we get around the UTF-16 conversion. 
You may check if it is a string literal by assert that the capacity() == 0 but size() > 0 

Maybe we create our own macro that tuns existing calls into a QStringLiteral call?  This way we can add a private function that accepts "const QString&" only.

Swiftb0y:
> You may check if it is a string literal by assert that the capacity() == 0 but size() > 0

Why would I need to check that? I'd need to do that at compile time for it to be useful.

> Maybe we create our own macro that tuns existing calls into a QStringLiteral call? 

Something like this?
```cpp
#define BENCH(key, ...) \
{ScopedTimer(QStringLiteral(key), __VA_ARGS__}
```

Swiftb0y:
tbh, I'd prefer to avoid the macro and instead construct the literal directly:
```cpp
// assumes previous `using namespace Qt::StringLiterals`
ScopedTimer t(u"key"_s);
```
I think the ergonomics are decent enough. The stringliteral is also constructed at compile time once you enable `-O1`. [Demo](https://compiler-explorer.com/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:21,endLineNumber:4,positionColumn:21,positionLineNumber:4,selectionStartColumn:21,selectionStartLineNumber:4,startColumn:21,startLineNumber:4),source:'%23include+%3CQtCore/QString%3E%0A%0AQString+foo()+%7B%0A++++using+namespace+Qt::StringLiterals%3B%0A++++return+u%22key%22_s%3B%0A%7D'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:gsnapshot,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:qt,ver:'660'),(name:pcre2,ver:'1042')),options:'--std%3Dc%2B%2B20+-fPIE+-O1+-Wall+-Wpedantic',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+(trunk)+(Editor+%231)',t:'0')),header:(),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+(trunk)',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+(trunk)+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

daschuer:
With the proposed `ScopedTimer t(u"key"_s);` call, it is IMHO way to easy to use it like `ScopedTimer t("key");` which allocates memory at runtime. 

>> You may check if it is a string literal by assert that the capacity() == 0 but size() > 0

> Why would I need to check that? I'd need to do that at compile time for it to be useful.

To avoid that users are starting to compose the QString outside the function. This would be DEBUG_ASSERT at runtime. 
Or is there a way to check it at compile time? 

One option would be to hide the `ScopedTimer(const QString& key)` somehow. 
That can be done with a macro. 

Like 
```
#define ScopedTimer(key, ...) \
ScopedTimerPrivate(QStringLiteral(key), __VA_ARGS__)
```
Which has the benefit the we need to change nothing in the calling code. 

> tbh, I'd prefer to avoid the macro and instead construct the literal directly:

I don't see a way around it, because QStringLiteral is also a macro.

ronso0:
Infamous "Close with comment"?

daschuer:
My idea was to put it after the developer mode check. If they introduce the timer, I am sure they will also use it. 

Does the `ScopedTimer t("key");` version count until \0? 
Do we want to prevent that because of the UTF-16 conversion? 
A macro version as discussed above will always use the fastest way and we can disallow all other ways.

Swiftb0y:
> Does the ScopedTimer t("key"); version count until \0?

yes. but only after the check. It probably wouldn't with [`QT_RESTRICTED_CAST_FROM_ASCII`](https://doc.qt.io/qt-6/qstring.html#QT_RESTRICTED_CAST_FROM_ASCII) defined. But that might require a bigger change. Also `strlen` is cheap compared to the required allocation and especially from `fromUtf8` conversion. But as I said, we can simply disable construction from `const char*` if you think thats an issue.

> A macro version as discussed above will always use the fastest way and we can disallow all other ways.

We can achieve the same if not more without a macro. Can you restate your exact requirements? I'm sure I can find a macro-less solution.

daschuer:
OK here is the list:
1. The current solution prevents allocating when not in developer mode by 100%. I think a new solution should do the same to be not considered as regression
2. Optional: Fix implicit use of strlen() at runtime when it loops through the whole string
3. Optional: Fix UTF-16 conversion at runtime

Swiftb0y:
> The current solution prevents allocating when not in developer mode by 100%. I think a new solution should do the same to be not considered as regression

This is already the case.

> Fix implicit use of strlen() at runtime when it loops through the whole string.

I assume with "fix" you mean "avoid"? I don't think thats really feasible. The only API that could do that would either be `QString::fromUtf8/16` but that would require special case handling which is unnecessarily clunky to deal with.

> Optional: Fix UTF-16 conversion at runtime

Not entirely sure how to do that reliably, but I can add a heuristic static_assert that tries to verify that the char type is 16-bit wide. Wdyt?

src/util/timer.h
```diff
@@ -52,61 +57,60 @@ class SuspendableTimer : public Timer {
...
+        if constexpr (std::is_array_v<T>) {
+            return sizeof(std::remove_all_extents_t<T>);
+        } else if constexpr (std::is_pointer_v<T>) {
+            // assume this is some point to char array then.
+            return sizeof(std::remove_pointer_t<T>);
+        } else {
+            // assume this is some QString or std::string type
+            return sizeof(typename T::value_type);
+        }
+    }
```
Swiftb0y, line 111: I don't like how brittle this is tbh. I'd much rather just forbid the specific `const char*` so people don't accidentally do the inefficient `ScopedTimer t("key")` case and be done with it.

src/util/timer.h
```diff
@@ -52,61 +57,47 @@ class SuspendableTimer : public Timer {
...
+    // when we're not in `--developer` mode.
+    /// @param compute Flags to use for the Stat::ComputeFlags (can be omitted)
+    /// @param key The format string for the key
+    /// @param args The arguments to pass to the format string
+    template<typename T, typename... Ts>
+    ScopedTimer(Stat::ComputeFlags compute, T&& key, Ts&&... args)
+            : m_timer(std::nullopt) {
+        static_assert(!std::is_same_v<std::remove_cvref_t<std::decay_t<T>>, char*>,
+                "string type likely not UTF-16, please pass QStringLiteral by "
+                "means of u\"key text\"_s");
```
Swiftb0y, line 70: this is much simpler and also forbids `ScopedTimer t("key")`.

src/util/timer.h
Swiftb0y: It also doesn't work properly on the current CI. I could probably work around it, but I don't think its worth it.

daschuer:
I can unfortunately not confirm that it fails after: 
```
-    ScopedTimer t(QStringLiteral("AnalyzerEbur128::processSamples()"));
+    ScopedTimer t("AnalyzerEbur128::processSamples()");
```

```
-    ScopedTimer t(QStringLiteral("AnalyzerEbur128::processSamples()"));
+    ScopedTimer t(QString("AnalyzerEbur128::processSamples()"));
```

daschuer:
It works as desired, when we replace the first template parameter with const QString&, than delete the const char* variant. The not QStringLiteral can be sorted out by DEBUG_ASSERT(key.capacity() == 0).
You may also consider to add an even faster overload without Ts&&

Swiftb0y:
> It works as desired, when we replace the first template parameter with const QString&, than delete the const char* 

Yes, thats also an option, but I thought the static_assert results in nicer debug output when violated compared to what the compiler reduces when it matches the deleted ctor.

> You may also consider to add an even faster overload without Ts&&

I don't see how that would be faster in terms of runtime. It only adds unnecessary repetition which this PR was supposed to reduce.

daschuer:
> I thought the static_assert results in nicer

Probably yes, but I cannot confirm it is working in the current state. The error output with the deleted function is good enough IMHO because it is no template type. 

```
src/analyzer/analyzerebur128.cpp:53:55: error: use of deleted function ‘ScopedTimer::ScopedTimer(char*, Ts&& ...) [with Ts = {}]’
   53 |     ScopedTimer t1("AnalyzerEbur128::processSamples()");
      |                                                       ^
In file included from src/analyzer/analyzerebur128.cpp:9:
src/util/timer.h:62:5: note: declared here
   62 |     ScopedTimer(char*, Ts&&... args) = delete;
      |     ^~~~~~~~~~~
ninja: build stopped: subcommand failed.
```

> I don't see how that would be faster in terms of runtime.

Oh yes, I missed the fact that `if constexpr` is optimized away.

Swiftb0y:
> I can unfortunately not confirm that it fails after:
> 
> ```
> -    ScopedTimer t(QStringLiteral("AnalyzerEbur128::processSamples()"));
> +    ScopedTimer t("AnalyzerEbur128::processSamples()");
> ```

whoops, yeah, forgot `const char*` is not `char const*`. fixed in fixup

Swiftb0y:
> Oh yes, I missed the fact that `if constexpr` is optimized away.

Yeah, fortunately not only optimized away but guaranteed to not be called because otherwise trying to call `QString::arg` without any parameters would be a compile-error (as already pointed out by the comment.

Swiftb0y:
once this lgty I'll rebase out the unnecessary intermediate commits and squash the fixups

daschuer:
Now I get this error message. Unfortunately it is like a wall of text not pointing directly to the point of the error.  
```
In file included from src/analyzer/analyzerebur128.cpp:9:
/home/sperry/workspace/mixxx2/src/util/timer.h: In instantiation of ‘ScopedTimer::ScopedTimer(Stat::ComputeFlags, T&&, Ts&& ...) [with T = const char (&)[34]; Ts = {}; Stat::ComputeFlags = int]’:
src/util/timer.h:88:96:   required from ‘ScopedTimer::ScopedTimer(T&&, Ts&& ...) [with T = const char (&)[34]; Ts = {}]’
src/analyzer/analyzerebur128.cpp:54:55:   required from here
src/util/timer.h:68:23: error: static assertion failed: string type likely not UTF-16, please pass QStringLiteral by means of u"key text"_s
   68 |         static_assert(!std::is_same_v<std::decay_t<T>, char const*>,
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[7/38] Building CXX object CMakeFiles/...lib.dir/src/library/dao/trackdao.cpp.o
ninja: build stopped: subcommand failed.
```
Do you really think this text is easier to understand than just the one demonstrated above: "use of deleted function" with the faulty line of code shown?

src/util/timer.h
```diff
@@ -52,61 +57,49 @@ class SuspendableTimer : public Timer {
...
-
-    ScopedTimer(const char* key, const char *arg = NULL,
-                Stat::ComputeFlags compute = kDefaultComputeFlags)
-            : m_pTimer(NULL),
-              m_cancel(false) {
-        if (CmdlineArgs::Instance().getDeveloper()) {
-            initialize(QString(key), arg ? QString(arg) : QString(), compute);
+        // key is explicitly `auto` to allow passing non-QString types such as `const char*`
+        // its conversion is delayed until after we've checked if we're in developer mode
+        auto assembledKey = QString(std::forward<T>(key));
```
daschuer, line 76: This is obsolete now, we don't want something else that a QString generated form a QStringLiteral.

daschuer (changes requested):

daschuer:
When one uses the `ScopedTimer t3(u"AnalyzerEbur128::processSamples()");` version introduced in main https://github.com/mixxxdj/mixxx/blob/8f647908af460e53fe8a860b6ce0964a93a55112/src/analyzer/analyzerebur128.cpp#L52 the error message becomes awful: 

```
In file included from src/analyzer/analyzerebur128.cpp:9:
src/util/timer.h: In instantiation of ‘ScopedTimer::ScopedTimer(Stat::ComputeFlags, T&&, Ts&& ...) [with T = const char16_t (&)[34]; Ts = {}; Stat::ComputeFlags = int]’:
src/util/timer.h:88:96:   required from ‘ScopedTimer::ScopedTimer(T&&, Ts&& ...) [with T = const char16_t (&)[34]; Ts = {}]’
src/analyzer/analyzerebur128.cpp:55:56:   required from here
src/util/timer.h:76:29: error: conversion from ‘const char16_t [34]’ to ‘QChar’ is ambiguous
   76 |         auto assembledKey = QString(std::forward<T>(key));
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:48,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qhashfunctions.h:44,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:47,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/QList:1,
                 from src/track/track_decl.h:5,
                 from src/analyzer/analyzertrack.h:5,
                 from src/analyzer/analyzer.h:3,
                 from src/analyzer/analyzerebur128.h:5,
                 from src/analyzer/analyzerebur128.cpp:1:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:90:22: note: candidate: ‘constexpr QChar::QChar(char16_t)’ <near match>
   90 |     Q_DECL_CONSTEXPR QChar(char16_t ch) Q_DECL_NOTHROW : ucs(ushort(ch)) {} // implicit
      |                      ^~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:90:22: note:   conversion of argument 1 would be ill-formed:
In file included from src/analyzer/analyzerebur128.cpp:9:
src/util/timer.h:76:29: error: invalid conversion from ‘const char16_t*’ to ‘char16_t’ [-fpermissive]
   76 |         auto assembledKey = QString(std::forward<T>(key));
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                             |
      |                             const char16_t*
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:48,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qhashfunctions.h:44,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:47,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/QList:1,
                 from src/track/track_decl.h:5,
                 from src/analyzer/analyzertrack.h:5,
                 from src/analyzer/analyzer.h:3,
                 from src/analyzer/analyzerebur128.h:5,
                 from src/analyzer/analyzerebur128.cpp:1:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:86:22: note: candidate: ‘constexpr QChar::QChar(int)’ <near match>
   86 |     Q_DECL_CONSTEXPR QChar(int rc) Q_DECL_NOTHROW : ucs(ushort(rc & 0xffff)) {}
      |                      ^~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:86:22: note:   conversion of argument 1 would be ill-formed:
In file included from src/analyzer/analyzerebur128.cpp:9:
src/util/timer.h:76:29: error: invalid conversion from ‘const char16_t*’ to ‘int’ [-fpermissive]
   76 |         auto assembledKey = QString(std::forward<T>(key));
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                             |
      |                             const char16_t*
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:48,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qhashfunctions.h:44,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:47,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/QList:1,
                 from src/track/track_decl.h:5,
                 from src/analyzer/analyzertrack.h:5,
                 from src/analyzer/analyzer.h:3,
                 from src/analyzer/analyzerebur128.h:5,
                 from src/analyzer/analyzerebur128.cpp:1:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:85:22: note: candidate: ‘constexpr QChar::QChar(uint)’ <near match>
   85 |     Q_DECL_CONSTEXPR QChar(uint rc) Q_DECL_NOTHROW : ucs(ushort(rc & 0xffff)) {}
      |                      ^~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:85:22: note:   conversion of argument 1 would be ill-formed:
In file included from src/analyzer/analyzerebur128.cpp:9:
src/util/timer.h:76:29: error: invalid conversion from ‘const char16_t*’ to ‘uint’ {aka ‘unsigned int’} [-fpermissive]
   76 |         auto assembledKey = QString(std::forward<T>(key));
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                             |
      |                             const char16_t*
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:48,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qhashfunctions.h:44,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:47,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/QList:1,
                 from src/track/track_decl.h:5,
                 from src/analyzer/analyzertrack.h:5,
                 from src/analyzer/analyzer.h:3,
                 from src/analyzer/analyzerebur128.h:5,
                 from src/analyzer/analyzerebur128.cpp:1:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:84:22: note: candidate: ‘constexpr QChar::QChar(short int)’ <near match>
   84 |     Q_DECL_CONSTEXPR QChar(short rc) Q_DECL_NOTHROW : ucs(ushort(rc)) {} // implicit
      |                      ^~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:84:22: note:   conversion of argument 1 would be ill-formed:
In file included from src/analyzer/analyzerebur128.cpp:9:
src/util/timer.h:76:29: error: invalid conversion from ‘const char16_t*’ to ‘short int’ [-fpermissive]
   76 |         auto assembledKey = QString(std::forward<T>(key));
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                             |
      |                             const char16_t*
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:48,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qhashfunctions.h:44,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:47,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/QList:1,
                 from src/track/track_decl.h:5,
                 from src/analyzer/analyzertrack.h:5,
                 from src/analyzer/analyzer.h:3,
                 from src/analyzer/analyzerebur128.h:5,
                 from src/analyzer/analyzerebur128.cpp:1:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:82:22: note: candidate: ‘constexpr QChar::QChar(ushort)’ <near match>
   82 |     Q_DECL_CONSTEXPR QChar(ushort rc) Q_DECL_NOTHROW : ucs(rc) {} // implicit
      |                      ^~~~~
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:82:22: note:   conversion of argument 1 would be ill-formed:
In file included from src/analyzer/analyzerebur128.cpp:9:
src/util/timer.h:76:29: error: invalid conversion from ‘const char16_t*’ to ‘ushort’ {aka ‘short unsigned int’} [-fpermissive]
   76 |         auto assembledKey = QString(std::forward<T>(key));
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                             |
      |                             const char16_t*
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/qhashfunctions.h:44,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/qlist.h:47,
                 from /usr/include/x86_64-linux-gnu/qt5/QtCore/QList:1,
                 from src/track/track_decl.h:5,
                 from src/analyzer/analyzertrack.h:5,
                 from src/analyzer/analyzer.h:3,
                 from src/analyzer/analyzerebur128.h:5,
                 from src/analyzer/analyzerebur128.cpp:1:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:225:19: note:   initializing argument 1 of ‘QString::QString(QChar)’
  225 |     QString(QChar c);
      |             ~~~~~~^
[7/34] Building CXX object CMakeFiles/...src/skin/legacy/legacyskinparser.cpp.o
ninja: build stopped: subcommand failed.
```

Swiftb0y:
> Do you really think this text is easier to understand than just the one demonstrated above: "use of deleted function" with the faulty line of code shown?

I'd say so yes because the solution with the deleted overload just says "use of deleted function" with no clear guidance on how to solve the issue. In the static_assert text we can provide an explanation and guidance on how to fix it. Yeah its a little more to read because the error is not on the first line for some reason, but IMO its still easy enough to spot where the error is when you look for the "error: [...]" line.

> When one uses the ScopedTimer t3(u"AnalyzerEbur128::processSamples()"); version introduced in main 

Ah, I forgot about that PR. the changes in `main` are very much incompatible with the changes made here anyways. I'll rebase on `main` then.

daschuer:
I would prefer the deleted function with an explaining source code comment, but I will not insists.

src/util/timer.h
```diff
@@ -52,61 +57,49 @@ class SuspendableTimer : public Timer {
...
+    // when we're not in `--developer` mode.
+    /// @param compute Flags to use for the Stat::ComputeFlags (can be omitted)
+    /// @param key The format string for the key
+    /// @param args The arguments to pass to the format string
+    template<typename T, typename... Ts>
+    ScopedTimer(Stat::ComputeFlags compute, T&& key, Ts&&... args)
+            : m_timer(std::nullopt) {
+        static_assert(!std::is_same_v<std::decay_t<T>, char const*>,
+                "string type likely not UTF-16, please pass QStringLiteral by "
+                "means of u\"key text\"_s");
```
daschuer, line 70:
```suggestion
                "Passing char[] is not supported. Wrap it via QStringLiteral() "
                "to avoid runtime UTF-16 conversion.");
```

daschuer (changes requested):

src/util/timer.h
Swiftb0y: Performance wise, we can rely on CoW here IMO.

Swiftb0y: Moreover, since we're taking an universal reference `T&&` the type will be `QString&&` anyways, resulting simply in a move which should be even cheaper.

Swiftb0y:
Superseeded by #13258

PR #13240, opened 2024-05-17
Refactor/gh13222 remove referenceholder

Swiftb0y:
Simplify messagePipe, rebased out from #13223


Swiftb0y:
ready for review

src/util/messagepipe.h
```diff
@@ -1,109 +1,83 @@
...
   public:
-    MessagePipe(rigtorp::SPSCQueue<SenderMessageType>& receiver_messages,
-                rigtorp::SPSCQueue<ReceiverMessageType>& sender_messages,
-                BaseReferenceHolder* pTwoWayMessagePipeReference)
+    MessagePipe(std::shared_ptr<rigtorp::SPSCQueue<SenderMessageType>> receiver_messages,
+            std::shared_ptr<rigtorp::SPSCQueue<ReceiverMessageType>> sender_messages)
             : m_receiver_messages(receiver_messages),
-              m_sender_messages(sender_messages),
-              m_pTwoWayMessagePipeReference(pTwoWayMessagePipeReference) {
+              m_sender_messages(sender_messages) {
```
acolombier, line 20: Would `std::move` here help with not having to copy `sender_messages` and `receiver_messages`

src/util/messagepipe.h
```diff
@@ -1,109 +1,83 @@
...
     }
 
     // Returns the number of ReceiverMessageType messages waiting to be read by
     // the receiver. Non-blocking.
     int messageCount() const {
-        return m_sender_messages.size();
+        return m_sender_messages->size();
```
acolombier, line 26: Might be worth adding a `DEBUG_ASSERT` either here or in the ctor to make sure `MessagePipe` doesn't get instantiated with null ptr

src/util/messagepipe.h
```diff
@@ -1,109 +1,83 @@
...
+                    receiver_fifo_size);
 
-    DISALLOW_COPY_AND_ASSIGN(TwoWayMessagePipe);
+    return {
+            std::make_unique<MessagePipe<SenderMessageType, ReceiverMessageType>>(
+                    receiver_pipe,
+                    sender_pipe),
+            std::make_unique<MessagePipe<ReceiverMessageType, SenderMessageType>>(
+                    sender_pipe,
+                    receiver_pipe)};
```
acolombier, line 82: Not entirely sure, but I think you may be able to use `std::move` here

acolombier (changes requested):

src/util/messagepipe.h
Swiftb0y: How exactly? Can you share you suggestion?

acolombier: I've seen place in the code where you would see something like

```suggestion
            std::make_unique<MessagePipe<SenderMessageType, ReceiverMessageType>>(
                    std:move(receiver_pipe),
                    std:move(sender_pipe)),
            std::make_unique<MessagePipe<ReceiverMessageType, SenderMessageType>>(
                    std:move(sender_pipe),
                    std:move(receiver_pipe))};
```
but perhaps this is redundant?

Swiftb0y: now that would actually be a bug afaict. `std::move`ing here will result in the `shared_ptr( shared_ptr&& r )` constructor being called. From cppreference:
> 10) Move-constructs a shared_ptr from r. After the construction, *this contains a copy of the previous state of r, r is empty and its stored pointer is null. 

So using either `sender_pipe`/`receiver_pipe` after moving from it would just leave it null. We can only ever move out of a variable once. 

We could do something like

```suggestion
            std::make_unique<MessagePipe<SenderMessageType, ReceiverMessageType>>(
                    receiver_pipe,
                    sender_pipe),
            std::make_unique<MessagePipe<ReceiverMessageType, SenderMessageType>>(
                    std::move(sender_pipe),
                    std::move(receiver_pipe))};
```

Which would result in only one move, but its undefined which argument here is evaluated first (so the value could be moved from before the actually desired copy took place). 

Something like this should work though:
```suggestion
        auto rec_to_sender = std::make_unique<MessagePipe<SenderMessageType, ReceiverMessageType>>(
                    receiver_pipe,
                    sender_pipe);
        auto sender_to_rec = std::make_unique<MessagePipe<ReceiverMessageType, SenderMessageType>>(
                    sender_pipe,
                    receiver_pipe)};
        return {std::move(rec_to_sender), std::move(sender_to_rec)};
```

src/util/messagepipe.h
Swiftb0y: opted for `gsl::not_null`

src/engine/effects/engineeffect.h
```diff
@@ -26,6 +26,7 @@ class EngineEffect final : public EffectsRequestHandler {
             const QSet<ChannelHandleAndGroup>& registeredInputChannels,
             const QSet<ChannelHandleAndGroup>& registeredOutputChannels);
     /// Called in main thread by EffectSlot
+    // Doesn't deal with ownership; only for conditional debug output
```
acolombier, line 29: Perhaps not for that PR, but I'm wondering if this could be done with a build option instead. 
Since `kEffectDebugOutput` needs to be flipped manually when one wants to debug effects, it could as well be a build option.

src/util/messagepipe.h
```diff
@@ -1,109 +1,94 @@
...
+makeTwoWayMessagePipe(
+        int sender_fifo_size,
+        int receiver_fifo_size) {
+    auto sender_pipe = gsl::make_not_null(
+            std::make_shared<rigtorp::SPSCQueue<ReceiverMessageType>>(
+                    sender_fifo_size));
+    auto receiver_pipe =
+            gsl::make_not_null(std::make_shared<rigtorp::SPSCQueue<SenderMessageType>>(
+                    receiver_fifo_size));
+    // TODO: remove unnecessary unique_ptr indirection
```
acolombier, line 86: I assume this `TODO` will need to remain for now, could it be updated with the GH issue it relates to?

acolombier (approved):
LGTM - only some mild nits that can be ignored.

src/effects/backends/builtin/balanceeffect.h
```diff
@@ -27,8 +27,13 @@ class BalanceGroupState : public EffectState {
 
 class BalanceEffect : public EffectProcessorImpl<BalanceGroupState> {
   public:
+    // TODO re-evaluate default/delete here. Is adhering to rule of zero possible?
```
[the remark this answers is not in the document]
Swiftb0y: Not as part of this PR, I just stumbled over this class because the build broke here (transitive header dependency) so this is the best bandaid fix I found.

src/engine/effects/engineeffect.h
Swiftb0y: Usually a constant is preffered because unlike with a preprocessor definition, the code that will get optimized out is still built and checked for correctness. That avoids the issue where code silently breaks because it has not been built under CI. Ideally we could instead compile-out the debug destructor by use of prospective destructors. I didn't want to go into that complexity yet though.

src/util/messagepipe.h
Swiftb0y: Just made the changes here instead. see newest commit.

Swiftb0y:
please rereview :)

acolombier (approved):
LGTM!

src/effects/backends/builtin/balanceeffect.h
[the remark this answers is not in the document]
Swiftb0y: https://github.com/mixxxdj/mixxx/issues/13409

Swiftb0y:
Thanks for the review. can a maintainer also take a look?

Swiftb0y:
friendly ping @mixxxdj/developers

Swiftb0y:
friendly ping ;)

ywwg:
looking

src/engine/effects/engineeffect.h
```diff
@@ -71,5 +72,4 @@ class EngineEffect final : public EffectsRequestHandler {
     QVector<EngineEffectParameterPointer> m_parameters;
     QMap<QString, EngineEffectParameterPointer> m_parametersById;
 
-    DISALLOW_COPY_AND_ASSIGN(EngineEffect);
```
ywwg, line 74: is the point of this PR to make it possible to copy this class? I want to make sure this is ok to remove

Swiftb0y: No, its just already not copy-able because of the `std::unique_ptr` member. So this adheres to the rule of zero. https://en.cppreference.com/w/cpp/language/rule_of_three#Rule_of_zero

Swiftb0y:
Thanks.

PR #13241, opened 2024-05-19
(fix) style Loop Anchor button in LateNight

ronso0:
follow up for #12745 

I desaturated the lock icons, because IMO they stood out too much in the loop section.

Looks like this now
image
image


daschuer (approved):
Thank you. Nice.

PR #13242, opened 2024-05-19
Tango/Qt6: fix rate range label position

ronso0:
currently:
image

this PR:
image


daschuer (approved):
Looks good with Qt 5 and 6

PR #13255, opened 2024-05-21
Numark PartyMix: fix EQ (script binding) display name

ronso0:
Noticed while testing the mapping IO table with a built-in mapping.

Group `[EqualizerRack1_[ChannelN]_Effect1]` will not be translated in the mapping IO GUI, and it's not required since it's already set via components. Simply use `[ChannelN]` to get the translated string '_Deck N_' in the GUI.

### DIFF
--- a/res/controllers/Numark-Party-Mix.midi.xml
+++ b/res/controllers/Numark-Party-Mix.midi.xml
@@ -111,7 +111,7 @@
       </control>
       <!-- EQ -->
       <control>
-        <group>[EqualizerRack1_[Channel1]_Effect1]</group>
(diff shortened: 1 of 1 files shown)


Swiftb0y (approved):
I doubt there is much value to having this translated in a mapping thats not made to be customized through the input-mapping table, but sure.

ronso0:
It's not about the translation itself but about the sorting. With the previous group `[EqualizerRack1_..` the items would appear at the end/top because of leading `[`, not near `Effects..`. Even with the original tr string (Deck N) all deck controls will now be grouped.

PR #13256, opened 2024-05-22
fix slipmode shader for macOS

m0dB:
I had to make some changes for the slipmode shader to compile on macOS.


acolombier (approved):
Thanks for the fix!

Swiftb0y (approved):
Thank you.

PR #13247, opened 2024-05-20
schema: Use only SQL single quotes for strings

saper:
New versions of sqlite no longer accept double
quotes for string literals.

On FreeBSD using sqlite 3.45.1 mixxx is unable
to start, even with an empty database, because
the migrations cannot be performed.

PR #13257, opened 2024-05-22
Fix/gh12894 single quote sqlite string literals

Swiftb0y:
Alternative to #13247 along with some manual follow up trying to replace these literals in our query-strings as well. Also based on 2.4 instead of `main` (as #13247 currently is). I'd appreciate if someone could double check if I've missed any string.

### DIFF
--- a/src/library/banshee/bansheedbconnection.cpp
+++ b/src/library/banshee/bansheedbconnection.cpp
@@ -41,7 +41,7 @@ bool BansheeDbConnection::open(const QString& databaseFile) {
 
 int BansheeDbConnection::getSchemaVersion() {
     QSqlQuery query(m_database);
-    query.prepare("SELECT Value FROM CoreConfiguration WHERE Key = \"DatabaseVersion\"");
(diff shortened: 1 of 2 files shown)


saper:
I've found only those to so far, but I haven't reviewed everything yet.

daschuer:
@Swiftb0y Can you confirm that https://github.com/mixxxdj/mixxx/pull/13247 is now the base of this PR? 
In that case I would prefer to merge both and rebase this one onto the original. This way we get around of the formal author/reviewer swap in this PR. Is this OK?

Swiftb0y:
> @Swiftb0y Can you confirm that https://github.com/mixxxdj/mixxx/pull/13247 is now the base of this PR?

Yes.

> In that case I would prefer to merge both and rebase this one onto the original. This way we get around of the formal author/reviewer swap in this PR. Is this OK?

Yes.

Swiftb0y:
rebased. please review.

JoergAtGithub:
LGTM! Thank you!

Swiftb0y:
Thank you

PR #13258, opened 2024-05-22
Refactor/shrink modernize scopedtimer (Alternative to #13236)

Swiftb0y:
#13236 but squashed and on top of main

### DIFF
--- a/src/util/timer.h
+++ b/src/util/timer.h
@@ -8,6 +8,7 @@
 #include "util/duration.h"
 #include "util/performancetimer.h"
 #include "util/stat.h"
+#include "util/stringformat.h"
 
 static constexpr Stat::ComputeFlags kDefaultComputeFlags = Stat::COUNT | Stat::SUM | Stat::AVERAGE |
         Stat::MAX | Stat::MIN | Stat::SAMPLE_VARIANCE;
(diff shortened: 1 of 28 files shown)


src/util/timer.h
```diff
@@ -39,40 +40,45 @@ class Timer {
...
+            return; // leave timer in cancelled state
         }
-#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
-        QString strKey = arg.isEmpty() ? key.toString() : key.arg(arg);
-#else
-        QString strKey = arg.isEmpty() ? key.toString() : key.toString().arg(arg);
-#endif
-        m_maybeTimer = std::make_optional<Timer>(std::move(strKey), compute);
+        //
+        auto assembledKey = QString(std::forward<T>(key));
```
daschuer, line 59: This one is unfortunately not for free, because the compiler decides to create a local String copy: 
https://godbolt.org/z/67M4MTc6j
Here is the `const QString&``version:  
https://godbolt.org/z/9WbMedoej

daschuer (changes requested):

src/util/timer.h
Swiftb0y: But in both cases, it still holds that `capacity() == 0`, thus the copy is not heap-allocated, so I don't really see a problem. Anyways, wdyt of the fixup. It avoids the construction of the temporary by use of an IILE.

Swiftb0y:
Ready for review @daschuer. Thank you.

daschuer (approved):
LGTM, Thank you.

Swiftb0y:
Thank you.

Issue #13206, opened 2024-05-05
Waveforms do not reject drops with unsupported files


daschuer:
### Bug Description

Instead, the cursor shall turn into a rejected symbol like it does for other widgets. 

### Version

2.4.0

### OS

Ubuntu Focal

daschuer:
A fix is here: https://github.com/mixxxdj/mixxx/pull/13208

m0dB:
fix merged into main

PR #13208, opened 2024-05-05
Handle not supported files when dragging to waveforms and spinnies

daschuer:
This fixes #13206

PR #13271, opened 2024-05-23
Revert "Handle not supported files when dragging to waveforms and spinnies"

ywwg:
Reverts mixxxdj/mixxx#13208

Tested manually, this fixes waveform drag breakage

### DIFF
--- a/src/util/dnd.cpp
+++ b/src/util/dnd.cpp
@@ -256,38 +256,38 @@ QDrag* DragAndDropHelper::dragTrackLocations(
 
 //static
 void DragAndDropHelper::handleTrackDragEnterEvent(
-        QDragEnterEvent* pEvent,
+        QDragEnterEvent* event,
         const QString& group,
         UserSettingsPointer pConfig) {
     if (allowLoadToPlayer(group, pConfig) &&
-            dragEnterAccept(*pEvent->mimeData(), group, true, false)) {
-        pEvent->acceptProposedAction();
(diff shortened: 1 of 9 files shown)


acolombier:
Looks like merging this would un-fix #13206. 
Looking at the code change, the bug appears to be fairly trivial to fix. Shall we consider fixing it instead of reverting entirely?

ywwg:
My preference when there is a breakage is to revert first and then work on a corrected fix, rather than try to "fix forward" 

(example reasoning: https://www.linkedin.com/pulse/service-recovery-rolling-back-vs-forward-fixing-mohamed-el-geish)

ywwg:
too often, what seems to be a simple fix on top of a breakage ends up making the situation worse, and then you have to fix on top of that.... unless the cost of rolling back is huge, it's best to do that first.

daschuer:
@acolombier If you have a fix under your thumb, we may test it and decide. Both are easy to test IMHO.

acolombier:
I don't yet, but might be able to get one together later today. Happy with the revert so this isn't a blocker

ywwg:
can someone approve so we can fix the 2.4 branch then?

acolombier (approved):

Swiftb0y:
fyi @ywwg it seems like you're pushing your personal branches upstream. Check your get config please.

daschuer:
The fix is here: https://github.com/mixxxdj/mixxx/pull/13275

PR #13273, opened 2024-05-23
allow changing the waveform overview type without reloading the skin

ronso0:
... by moving the RGB/LMH/HSV pixmap drawing into the base class WOveriew.
Type change is caught by a ControlProxy.

Switching works super smooth :sunglasses: 

Partially fixes #9745 
This is a bugfix IMO and can slip in 2.5 anytime.

**Nice to have:** enum for overview type

_Hopefully not to many conflicts with the current waveform pref work_


ronso0:
I think this super low-risk and can go to 2.5

FWIW the only change I made here was to move the draw functions to the base class, no further adjustments.

src/widget/woverview.cpp
```diff
@@ -1211,6 +1237,285 @@ void WOverview::drawPassthroughOverlay(QPainter* pPainter) {
...
+    }
+
+    m_actualCompletion = nextCompletion;
+}
+
+void WOverview::drawNextPixmapPartLMH(QPainter* pPainter,
+        ConstWaveformPointer pWaveform,
+        const int nextCompletion) {
+    DEBUG_ASSERT(!m_waveformSourceImage.isNull());
+    ScopedTimer t(u"WOverview::drawNextPixmapPartLMH");
```
daschuer, line 1385:
```suggestion
    ScopedTimer t(QStringLiteral("WOverview::drawNextPixmapPartLMH"));
```

src/widget/woverview.cpp
```diff
@@ -1211,6 +1237,285 @@ void WOverview::drawPassthroughOverlay(QPainter* pPainter) {
...
+    }
+
+    m_actualCompletion = nextCompletion;
+}
+
+void WOverview::drawNextPixmapPartRGB(QPainter* pPainter,
+        ConstWaveformPointer pWaveform,
+        const int nextCompletion) {
+    DEBUG_ASSERT(!m_waveformSourceImage.isNull());
+    ScopedTimer t(u"WOverview::drawNextPixmapPartRGB");
```
daschuer, line 1447:
```suggestion
    ScopedTimer t(QStringLiteral("WOverview::drawNextPixmapPartRGB"));
```

daschuer (changes requested):
It works like a charm, Thank you. Somme comments:

src/widget/woverview.h
```diff
@@ -130,20 +101,41 @@ class WOverview : public WWidget, public TrackDropTarget {
...
+        }
+    }
+
+    // Hold the last visual sample processed to generate the pixmap
+
     const QString m_group;
     UserSettingsPointer m_pConfig;
-    parented_ptr<ControlProxy> m_endOfTrackControl;
+
+    int m_type;
```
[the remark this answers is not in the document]
ronso0: Sure, we can that's why I mentioned it as nice-to-have in the description, and I already took a stab at it.
But I gave up since it took to long and isn't mandatory IMO, even feels like this makes the preferences more complicated (or I simply messed something up there)

I can take another look, but actually I'd appreciate if we could this merged soonish and refine later on (before the 2.5 release)

daschuer (approved):
LGTM thank you.

PR #13278, opened 2024-05-23
GitHub CI: Skip the manifest update job on forks

cr7pt0gr4ph7:
The steps of manifest update job are already skipped on forks because they do not have the necessary secrets configured, so we skip the job altogether for any repository that is not the main `mixxxdj/mixxx` repository to avoid unnecessary noise in the build log (as well as some issues related to canceling build jobs).

It would be preferable to directly check for the presence of the secrets in `jobs.<job_id>.if`, but that is not possible, as per actions/runner#520.


Swiftb0y (approved):
LGTM. any objections?

daschuer (approved):
LGTM

PR #13279, opened 2024-05-24
fix(Roland DJ_505): generate proper warning if audio inputs are not setup for passthrough

### LINKED ISSUE #13252: Allow to permanently disable warning message about missing input device
### Feature Description

I don't want to use an input device at the moment but this message appears every time I open mixxx.

It's quite annoying

image

Please add a "don't show again" checkbox to the dialog

Swiftb0y:
Fixes #13252
Untested as I don't have the controller.
CC @Holzhaus @spotlesscoder

### DIFF
--- a/res/controllers/Roland_DJ-505-scripts.js
+++ b/res/controllers/Roland_DJ-505-scripts.js
@@ -377,6 +377,9 @@ DJ505.setChannelInput = function(channel, control, value, _status, _group) {
         break;
     case 0x01:  // LINE
     case 0x02:  // PHONO
+        if (engine.getValue(channelgroup, "input_configured") === 0) {
+            throw "Configuration Error: a PC/Line/Phono switch has been set to Line or Phono. In order to use these modes audio inputs to be configured accordingly! Please refer to https://manual.mixxx.org/latest/hardware/controllers/roland_dj_505#audio-setup";
+        }
         engine.setValue(channelgroup, "passthrough", 1);
         break;
     }
--- a/res/controllers/Roland_DJ-505.midi.xml
+++ b/res/controllers/Roland_DJ-505.midi.xml
@@ -3,7 +3,7 @@
     <info>
         <name>Roland DJ-505</name>
         <author>Jan Holthuis</author>
-        <description>4-deck mapping for Roland DJ-505 controller</description>
+        <description>4-deck mapping for Roland DJ-505 controller. Make sure you setup audio outputs and inputs correctly (see manual)</description>
         <forums>https://mixxx.discourse.group/t/roland-dj-505/17916</forums>
         <manual>roland_dj_505</manual>
     </info>


ronso0 (dismissed):

daschuer:
Is the mapping unusable without this setting? I am asking, because after merging this it is artificialy unusable.

Swiftb0y:
Well, it depends. Its not unusable because it'll only throw when receiving input from those specific controls (so in an input handler; which doesn't break the mapping). If you look at https://github.com/mixxxdj/mixxx/issues/13252, the user is very annoyed with the fact that the mapping tries to enable passthrough (because of them not RTFM). This PR provides the user with explicit feedback what they need to do to fix it, alternatively they can also use the "ignore this error" in the feedback. Either way, without them setting up the inputs, these buttons are broken anyways. So I wouldn't say this causes further breakage, but it would be nice if someone with access to the controller could confirm.

spotlesscoder:
I can test the change if you give me a hint how to do it. However, I still would have struggled to fix the problem - even with the new message. I would change the text to something like "Configuration Error: Mapping expects audio inputs to be configured! Did you change one or both of the channel inputs to either Line or Phono input unintentionally using the switches on the front panel of the controller? Please refer to https://manual.mixxx.org/latest/hardware/controllers/roland_dj_505#audio-setup"

ronso0:
> I can test the change if you give me a hint how to do it.

You can either check out the modified mapping via git/Github, or you download the two files manually and put them into your [user mapping directory](https://manual.mixxx.org/2.4/chapters/appendix/settings_directory#the-mixxx-settings-directory). (`../controllers`, need to create it if it doesn't exist)

You'll find the files by scrolling to the top of this page, find the [Files changed](https://github.com/mixxxdj/mixxx/pull/13279/files) tab, and get the RAW files.

spotlesscoder:
OK - so what are your opinions about the message I suggested? 
If you agree with it I would love to test it with that message once the PR is updated so we don't have to test it one more time

Swiftb0y:
Changing the text is no big deal, I just want to know whether the rest of the mapping works regardless of the text. The text you suggested is also misleading IMO, the problem is not that you set the switches incorrectly, its that you didn't configure mixxx properly to use the controllers audio inputs IMO.

spotlesscoder:
I would say whether the text is misleading or not is dependent on what you want to achieve. I only started using an external vinyl player this week as an experiment and forgot to change it back. This was causing me not being able to using mixxx as I was used to

Swiftb0y:
Well, thats completely fine, the mapping still expects you to properly configure the inputs, even if you're not sending anything through them (as I assume you're doing with the external vinyl player). In the case where you are using an external player, mixxx will set the corresponding deck to passthrough (to make it apparent that interacting with that deck on the controller will not control mixxx). But passthrough still only works when the input for that deck is actually configured. So no matter whether you use an external player and send the signal to mixxx (for example for vinyl control) or you're bypassing mixxx entirely for improved latency, the audio setup for the controller should still be the exact same (with inputs configured).

Swiftb0y:
friendly ping ;)

res/controllers/Roland_DJ-505-scripts.js
```diff
@@ -371,6 +371,9 @@ DJ505.crossfader.setReverse = function(channel, control, value, _status, _group)
 DJ505.setChannelInput = function(channel, control, value, _status, _group) {
     const number = (channel === 0x00) ? 0 : 1;
     const channelgroup = "[Channel" + (number + 1) + "]";
+    if (engine.getValue(channelgroup, "input_configured") === 0) {
+        throw "Configuration Error: Mapping expects audio inputs to be configured! please refer to https://manual.mixxx.org/latest/hardware/controllers/roland_dj_505#audio-setup";
+    }
```
ronso0, line 376: IIUC this should only be thrown if the input value is 0x01 or 0x02, right?
I.e. when the 505 reports the controls states on startup, and when the user flicks the switch.

In that regard the manual is a bit misleading, users may be confused by
> You should assign the Vinyl Control input channels even if you do not intend to use timecode vinyl. 

because if they don't intend to ever use VC, or use VC with another sound card or whatever, there is no **need** to configure the inputs.

Bottom line: the current message is okay because it matches the manual, but IMHO the manual should be changed.

We could still rephrase it a bit:
```suggestion
    if (engine.getValue(channelgroup, "input_configured") === 0) {
        throw "Configuration Error: a PC/Line/Phono switch has been set to Line or Phono. In order to use these modes audio inputs to be configured accordingly! Please refer to https://manual.mixxx.org/latest/hardware/controllers/roland_dj_505#audio-setup";
    }
```

Holzhaus: > In that regard the manual is a bit misleading, users may be confused by
>
>> You should assign the Vinyl Control input channels even if you do not intend to use timecode vinyl.
>
> because if they don't intend to ever use VC, or use VC with another sound card or whatever, there is no need to configure the inputs.

At the time of writing this, I was not using Vinyl Control (= DVS) at all, but I did include some actual vinyl records into my mixes. So even though I did not intend to use vinyl control, I still needed to configure these inputs for passing though non-timecode records. Maybe we should rename the input channels to "Phono" in the preferences ;)

spotlesscoder: I like the rephrased version a lot more - and also the renaming in the preferences would be good

Swiftb0y:
Done ;)

spotlesscoder:
I tested it but now the message is gone completely. Also, when I switch to phone, the UI does not display "Passthrough" on the deck

ronso0:
> I tested it but now the message is gone completely. Also, when I switch to phone, the UI does not display "Passthrough" on the deck

The switches are only for the routing on the conteoller. There is no auto-switch to Passthrough, why should it? Phono maybe be used for VC or passthrough.

ronso0:
LGTM thanks. @holzhaus wdyt?

spotlesscoder:
> > I tested it but now the message is gone completely. Also, when I switch to phone, the UI does not display "Passthrough" on the deck
> 
> The switches are only for the routing on the conteoller. There is no auto-switch to Passthrough, why should it? Phono maybe be used for VC or passthrough.



> > I tested it but now the message is gone completely. Also, when I switch to phone, the UI does not display "Passthrough" on the deck
> 
> The switches are only for the routing on the conteoller. There is no auto-switch to Passthrough, why should it? Phono maybe be used for VC or passthrough.

Hmmm. Seems like I still don't fully understand how all these things are meant to be used

Swiftb0y:
Actually the DJ-505 seems to do something weird (compared to controllers by Numark or Native Instruments). The tri-state switch on the back only really controls where the audio input is coming from (eg PC or directly (optionally with a phono-preamp)). Then when its set to the PC setting, you would manually need to specify phono/line signal from within serato (see screenshots).  Other controllers usually just have line/phono switches and separate "direct-out"/PC switches. So the most recent change (moving the check to the Line/phono case is actually incorrect). Does that make sense @ronso0? Also further reverse-engineering would need to be done on how specifying controlling the input pre-amp from the host sofware works. 

505 Block Diagramm

DVS Vinyl Setup

ronso0:
> the tri-state switch on the back

... front ; )

> Does that make sense @ronso0?

Nope :laughing: 
To me it looks like the block diagram is missing the PC input, so I wouldn't rely on that.
Actually @Holzhaus and/or @spotlesscoder have to verify what is strictly required to play sounds from a Mixxx deck.

ronso0:
Re: Searato DVS
> If you use DVS, it is not possible to combine a turntable with a DJ player. In this case, you must use either
use two turntables or two DJ players.

> Also further reverse-engineering would need to be done on how specifying controlling the input pre-amp from the host sofware works.

:+1: 
So the switches can be set internally, but not independently :grimacing:

spotlesscoder:
So what exactly can I do to help here?

ronso0:
IIUC this is the situation (please edit this post if I'm wrong):
For vinyl control you need to set the PC/Line/Phono switches to **PC** and to configure the VC inputs in Mixxx.
Use **Phono** or **Line** if you want to use that channel in standalone mode, i.e. use the input without Mixxx.
As the official DJ-505 manual states, for VC in Serato, you need to select the input sources (phono or line level) there (Serato can only set it for both inputs).

Now, the **easy** part is to find out which input level the Phono/Line inputs are set to by default:
* switch one input to PC
* select that input in Mixxx for VC deck1 in Preferences -> Sound Hardware
* in Preferences -> Vinyl Control, unset the VC type for deck1, disable phono pre-amp
* attach a player
* enable Passthrough in deck1
* check the input level/sound in Mixxx
* if it's way too loud while the player is set to a moderate level the DJ-505 expects phono level

The reverse-engineering part is figuring out how the level (phono pre-amp On/Off) is set via software.
For that we need someone with the DJ-505, Serato installed and a USB sniffer (wireshark?) in order to check the message sent to the DJ-505 when Serato is setting it up for either turntable or CDJ.

Swiftb0y:
>  To me it looks like the block diagram is missing the PC input, so I wouldn't rely on that.

If that switch is in PC, it'll likely just disable the internal analog mixing and just rely on the fact that the deck sound is coming through the "Master Input" (thats also how it works on virtually any other controller I own).

> As the official DJ-505 manual states, for VC in Serato, you need to select the input sources (phono or line level) there (Serato can only set it for both inputs).

The screenshot I posted above explicitly state that for VC the input needs to be set to PC IMO. otherwise it would just mix the VC signal internally as if it was some external music input.

ronso0:
> If that switch is in PC, it'll likely just disable the internal analog mixing and just rely on the fact that the deck sound is coming through the "Master Input" (thats also how it works on virtually any other controller I own).

:+1: 

> The screenshot I posted above explicitly state that for VC the input needs to be set to PC IMO. otherwise it would just mix the VC signal internally as if it was some external music input.

Jup, that's what I was trying to say. PC on the controller, Ch1+2 or 3+4 in Mixxx.

spotlesscoder:
If you need me to test something with the device, I need a bit of assistance - you can contact me and we can make a screen share call

spotlesscoder:
any testing still needed?

Holzhaus (approved):
Changes look sensible to me, I can try to actually test this tomorrow. But judging from the diff, I see no reason why there should be an issue with this PR.

res/controllers/Roland_DJ-505-scripts.js
```diff
@@ -377,6 +377,9 @@ DJ505.setChannelInput = function(channel, control, value, _status, _group) {
         break;
     case 0x01:  // LINE
     case 0x02:  // PHONO
+        if (engine.getValue(channelgroup, "input_configured") === 0) {
+            throw "Configuration Error: a PC/Line/Phono switch has been set to Line or Phono. In order to use these modes audio inputs to be configured accordingly! Please refer to https://manual.mixxx.org/latest/hardware/controllers/roland_dj_505#audio-setup";
```
ronso0, line 381:
```suggestion
            throw "Configuration Error: a PC/Line/Phono switch has been set to Line or Phono. In order to use these modes audio inputs need to be configured accordingly! Please refer to https://manual.mixxx.org/latest/hardware/controllers/roland_dj_505#audio-setup";
```

ronso0: Then we can merge this, right? @Holzhaus @Swiftb0y

spotlesscoder:
Bump

Swiftb0y:
Thanks for the ping. In order to implement this properly, I need you capture some midi messages between serato and 505, can you do that for me @spotlesscoder?

spotlesscoder:
If you give me detailed instructions, I can do that

Swiftb0y:
Does this help? https://github.com/mixxxdj/mixxx/wiki/Reverse%20engineering#observing-midihid-messages-with-other-dj-software

PR #13281, opened 2024-05-25
Parametric google-test for mapping validation

JoergAtGithub:
For debugging https://github.com/mixxxdj/mixxx/pull/11625#issuecomment-2131243612 I modified the googletest setup for mapping validation to use a parametrized list of mappings.
While the results look nice in my IDE, where I can easily debug a single mapping validation test now:
grafik

I was disappointed to see that CTest combines all the single GTest instances:

```
        Start 121: */LegacyControllerMappingValidationTest.ValidateMappingXML/*
121/840 Test #121: */LegacyControllerMappingValidationTest.ValidateMappingXML/* ...............................   Passed    0.37 sec
        Start 122: */LegacyControllerMappingValidationTest.LoadMapping/*
122/840 Test #122: */LegacyControllerMappingValidationTest.LoadMapping/* ......................................***Failed    5.44 sec
```

I could solve that by using gtest_discover_tests in the CMakelist.txt, now the CTest output looks like this:
```
          Start 1039: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_CDJ_350_CH2#GetParam()=<MappingInfo>
1039/1107 Test #1039: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_CDJ_350_CH2#GetParam()=<MappingInfo> .................................   Passed    0.45 sec
          Start 1040: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_CDJ_850#GetParam()=<MappingInfo>
1040/1107 Test #1040: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_CDJ_850#GetParam()=<MappingInfo> .....................................   Passed    0.45 sec
          Start 1041: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_DDJ_200#GetParam()=<MappingInfo>
1041/1107 Test #1041: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_DDJ_200#GetParam()=<MappingInfo> .....................................   Passed    0.53 sec
          Start 1042: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_DDJ_400#GetParam()=<MappingInfo>
1042/1107 Test #1042: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_DDJ_400#GetParam()=<MappingInfo> .....................................   Passed    0.45 sec
          Start 1043: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_DDJ_FLX4#GetParam()=<MappingInfo>
1043/1107 Test #1043: MidiMappingsValid/LegacyControllerMappingValidationTest.LoadMapping/Pioneer_DDJ_FLX4#GetParam()=<MappingInfo> ....................................   Passed    0.47 sec
```


src/test/controller_mapping_validation_test.h
```diff
@@ -136,12 +136,40 @@ class FakeController : public Controller {
...
+    QScopedPointer<MappingInfoEnumerator> m_pEnumerator;
+};
+
+class LegacyControllerMappingList {
+  public:
+    LegacyControllerMappingList() {
+        QList<QString> possiblePaths = {
+                QDir::currentPath() + "/res/controllers",
+                QDir::currentPath() + "/../res/controllers",
+                QDir::currentPath() + "/../../res/controllers"};
```
Swiftb0y, line 152: don't we have `getResourcePath` for that instead of having to guess?

src/test/controller_mapping_validation_test.cpp
```diff
@@ -163,35 +159,69 @@ bool lintMappingInfo(const MappingInfo& mapping) {
...
+    template<class ParamType>
+    std::string operator()(const ::testing::TestParamInfo<ParamType>& info) const {
+        // ParamType is MappingInfo
+        QString nameCopy = info.param.getName();
+        static const QRegularExpression regex("[^\\w\\s]");
+        nameCopy.replace(regex, "_");
+        std::string testName = nameCopy.replace(" ", "_").toStdString();
+        return testName;
     }
+};
```
Swiftb0y, line 187: this seems overcomplicated. why not rely on `PrintToStringParamName`?

src/test/controller_mapping_validation_test.cpp
```diff
@@ -163,35 +159,69 @@ bool lintMappingInfo(const MappingInfo& mapping) {
...
+
+auto pEnumerator = createMappingEnumerator();
```
Swiftb0y, line 189: this relies on running before main. I don't know how safe that is considering the heavy Qt classes involved. It would be better if this was run in `main()`, but before the gtest init code. Is that feasible?

acolombier:
I think we should get this improvement in! I have taken some time to rebase this branch against `2.6` and add some minor improvement, related to @Swiftb0y suggestions

https://github.com/mixxxdj/mixxx/commit/9f99d0158f

Let me know if you want me to submit it in a PR directly.

JoergAtGithub:
The problem is that this PR slows down the test significiantly, because creating the list of mappings takes a lot of time. A hardcoded list shouldn't have this issue.
Otherwise thi PR works.

Swiftb0y:
what magnitude of slowdown are we talking about here? seconds?

JoergAtGithub:
The whole testsuite needs about 2x time as before, because the list of mappings is calculate before every test (not only the test that depend on this parameter)

Swiftb0y:
oh yeah, then that is a significant problem.

acolombier:
I have [pushed a version](https://github.com/mixxxdj/mixxx/commit/13611c3b06#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a) with the mapping list computed directly in CMake. Only did MIDI as example, but keen to hear if we would want to go ahead with it. There is undoubtedly an overhead with running multiple tests, perhaps we could make it optional, so only the `coverage` job would run the test in parametrised mode and the slowdown wouldn't impact the build job?

JoergAtGithub:
GTest must always work standalone without CTest, as GTest is what is used in IDEs. One of the challenges to achieve this ist to find the ./res/controller directory in all environments.

acolombier:
Yes, this is what my solution does. It only finds the mapping at compile time to build the parameter list. Once built, `mixxx-test` can be used independently to CTest, and should work fine in the IDE as it will find the `res/controller` directory in the same way it predictably find `src`

JoergAtGithub:
Finally ready, thanks to the suggestion by @acolombier above!

res/controllers/midi-components-0.0.js
```diff
@@ -130,14 +130,18 @@
         disconnect: function() {
             if (this.connections[0] !== undefined) {
                 this.connections.forEach(function(conn) {
-                    conn.disconnect();
+                    if (conn !== undefined) {
```
acolombier, line 133: Is this an expected usecase? It does makes sense in the case of testing due to missing CO, but I assume that `connections` should only contain valid connections, else it would indicate a problem. 
Should we add a `console.warn` to flag potentially failed connection?

src/test/controller_mapping_validation_test.cpp
```diff
@@ -2,7 +2,10 @@
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
+#include <qapplication.h>
+#include <qglobal.h>
```
acolombier: Sorry - I added those imports...
Nit would be to use the alias


```suggestion
#include <QApplication>
#include <Qt>
```

src/test/controller_mapping_validation_test.cpp
```diff
@@ -239,39 +250,47 @@ bool lintMappingInfo(const MappingInfo& mapping) {
...
-        std::string errorDescription = "Error while validating " + mapping.getPath().toStdString();
-        EXPECT_TRUE(mapping.isValid()) << errorDescription;
-        EXPECT_TRUE(lintMappingInfo(mapping)) << errorDescription;
-        EXPECT_TRUE(testLoadMapping(mapping)) << errorDescription;
-    }
+// Inhibit the output of the mapping info to avoid spamming the console
+// std::ostream& operator<<(std::ostream& os, const QString& MappingPath) {
+//     return os << MappingPath.toStdString();
+// }
+
```
acolombier: Remove?
```suggestion
```

src/test/controller_mapping_validation_test.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <qglobal.h>
```
acolombier:
```suggestion
#include <Qt>
```

CMakeLists.txt
```diff
@@ -2848,6 +2852,30 @@ if(BUILD_TESTING)
...
+  list(JOIN CONTROLLER_HID_MAPPING_LIST "\",\"" CONTROLLER_HID_MAPPINGS)
+  list(JOIN CONTROLLER_MIDI_MAPPING_LIST "\",\"" CONTROLLER_MIDI_MAPPINGS)
+
+  target_compile_definitions(
+    mixxx-test
+    PUBLIC CONTROLLER_BULK_MAPPINGS="${CONTROLLER_BULK_MAPPINGS}"
+    PUBLIC CONTROLLER_HID_MAPPINGS="${CONTROLLER_HID_MAPPINGS}"
+    PUBLIC CONTROLLER_MIDI_MAPPINGS="${CONTROLLER_MIDI_MAPPINGS}"
+  )
+
```
acolombier, line 2878: DO we need to make the above conditional to `HID`, and `BULK`?

acolombier (commented):
Looking good, thanks for pushing this to the end! I haven't tested it yet

res/controllers/midi-components-0.0.js
JoergAtGithub: Detailed warnings are already implemented in engine.makeConnection: https://github.com/mixxxdj/mixxx/blob/c10f342ccf7aa12b13348788aae6f109e01387b0/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp#L276-L290
But after the warning engine.makeConnection returns `undefined` and the script continues.

CMakeLists.txt
JoergAtGithub: I think the simplicity of the code is of more value than skiping 1 or 2 regular expressions during CMake configuration.

acolombier (approved):
LGTM! This seems to slow down the test run by about 30 seconds (for Ubuntu). That being said, the added value is worth the delay IMO. 

Giving some time to let other member a chance to review this as well.

JoergAtGithub:
Merge?

PR #13282, opened 2024-05-25
adjust FocusWidget: GoToItem emulates Space key in WSearchRelatedTracksMenu

ronso0:
Requested here https://github.com/mixxxdj/mixxx/pull/12213#issuecomment-2128069547

This allows to toggle the search actions and click the Search button with `[Library],GoToItem`, which would previously activate the actions (trigger search and close the menu).

Needs some testing with controllers !

### DIFF
--- a/src/library/librarycontrol.cpp
+++ b/src/library/librarycontrol.cpp
@@ -1,6 +1,7 @@
 #include "library/librarycontrol.h"
 
 #include <QApplication>
+#include <QCheckBox>
 #include <QKeyEvent>
 #include <QModelIndex>
 #include <QWindow>
@@ -17,6 +18,7 @@
 #include "widget/wlibrary.h"
 #include "widget/wlibrarysidebar.h"
 #include "widget/wsearchlineedit.h"
+#include "widget/wsearchrelatedtracksmenu.h"
 #include "widget/wtracktableview.h"
 
 namespace {
@@ -692,7 +694,8 @@ void LibraryControl::slotMoveVertical(double v) {
                 QEvent::KeyPress, key, Qt::NoModifier, QString(), false, times});
         return;
     }
-    case FocusWidget::ContextMenu: {
+    case FocusWidget::ContextMenu:
+    case FocusWidget::SearchRelatedMenu: {
         // To navigate menus (and activate menus that were just opened) send the
         // keyEvent to focusWindow() (not focusWidget() like emitKeyEvent() does)
         const auto key = (v < 0) ? Qt::Key_Up : Qt::Key_Down;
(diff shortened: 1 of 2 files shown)


acolombier:
I'll give it a go ASAP!

acolombier (commented):
Unfortunatelly, it doesn't work with my controller (S4Mk3, tested with the mapping available on that same branch)

Here is the logic snippet used for testing:

```js
const currentlyFocusWidget = engine.getValue("[Library]", "focused_widget");
// 3 == Tracks table or root views of library features
if (this.shifted && currentlyFocusWidget === 0) {
    script.triggerControl("[Playlist]", "ToggleSelectedSidebarItem");
} else if (currentlyFocusWidget === 3 || currentlyFocusWidget === 0) {
    script.triggerControl(this.group, "LoadSelectedTrack");
} else {
    script.triggerControl("[Library]", "GoToItem");
}
```

However, instead of selecting the checkbox, it just trigger the search with that unique criteria. That seems expected reading the [GoToItem](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[Library]-GoToItem) doc. Perhaps we should add another value to `GoToItem`? (1 == Enter, 2 == Space)

Note that `focused_widget` value was always `4` so something might be wrong with the newly introduced `FocusWidget::SearchRelatedMenu`.

ronso0:
Admittedly this was just a quick shot, I didn't even test it :grimacing: 

Turns out only the **main** popup is considered, not its active child WSearchRelatedTracksMenu.
And when cycling through the search actions, the reported focused widget is first the WSearchRelatedTracksMenu, then the QCheckBox.

ronso0:
@acolombier This should work now, at least `LibraryControl::getFocusedWidget()` now properly detects the search menu.

ronso0:
NOW it works (tested with a MIDI Through script -- will test with my S4 soonish)

src/library/librarycontrol.cpp
```diff
@@ -866,6 +811,9 @@ FocusWidget LibraryControl::getFocusedWidget() {
         return FocusWidget::None;
     }
 
+    qWarning() << "     .";
+    qWarning() << "     focWin:" << focusWindow;
+    qWarning() << "       name:" << focusWindow->objectName();
```
daschuer: This seems to be debug leftover.

src/library/librarycontrol.cpp
```diff
@@ -876,7 +824,18 @@ FocusWidget LibraryControl::getFocusedWidget() {
         // qt_edit_menuWindow    = QLineEdit/QCombobox context menu
         // QComboBoxPrivateContainerClassWindow
         //    = QComboBoxListView of WEffectSelector, WSearchLineEdit, ...
-        return FocusWidget::ContextMenu;
+        auto* focWid = QApplication::focusWidget();
```
daschuer:
```suggestion
        auto* pFocWid = QApplication::focusWidget();
```

daschuer (commented):
The code looks good, except of the bug leftovers.

ronso0:
Added the p prefix and removed the trace commit.
Will do a test with my controller tomorrow and report back (simulation with mapping for MIDI Through already worked)

ronso0:
It works as desired with my mapping.

_____
Though I need to change the keyboard shortcut for `[Skin],show_maximized_library` to something else that Space, else I run into #15997-- but that's apparently related to my system/Qt version

daschuer (approved):
LGTM, Thank you.

PR #4723, opened 2022-04-12
[Don't merge] libshout idjc 2.4.6

daschuer:
This is an experimental branch to update to the recent libshout idjc. 
Currently we have a patched libshout idjc 2.4.1 in our lib folder which works without known issues. 
Once the libshout idjc 2.4.6 will land to Linux distros our build system will link against that and we can later remove our vendored version. This PR can used to do some preliminary checks. 
If we need features form the new 2.4.6 version before, we may consider to merge this.

PR #13285, opened 2024-05-25
feat: add encryption option for broadcasting

acolombier:
This attempts to address #8301 and #8706

Couple of notes:
- Behaviour is pretty random and unstable with the current libshout version, but far more stable and expectable with #4723
- Tested with an Icecast server 2.4.4
- `SHOUT_TLS` seems hardcoded to `1` so I was tempted to remove the flag entirely 
- This touches on section of the code quite off our current C++ standard, but I've decided to stay as aligned as possible with the rest of the file style when possible


JoergAtGithub:
There is a user confirmation, that this PR works: https://github.com/mixxxdj/mixxx/issues/8301#issuecomment-2182140643

src/engine/sidechain/shoutconnection.cpp
```diff
@@ -241,6 +231,27 @@ void ShoutConnection::updateFromPreferences() {
...
+    case BroadcastProfile::EncryptionMode::Preferred:
+        result = shout_set_tls(m_pShout, SHOUT_TLS_AUTO);
+        break;
+    case BroadcastProfile::EncryptionMode::Disabled:
+        result = shout_set_tls(m_pShout, SHOUT_TLS_DISABLED);
+        break;
+    default: // Required
+        result = shout_set_tls(m_pShout, SHOUT_TLS_AUTO_NO_PLAIN);
+        break;
+    }
```
daschuer, line 248: We have these values: 
```
#define SHOUT_TLS_DISABLED          (  0) /* Do not use TLS at all */
#define SHOUT_TLS_AUTO              (  1) /* Autodetect which TLS mode to use if any */
#define SHOUT_TLS_AUTO_NO_PLAIN     (  2) /* Like SHOUT_TLS_AUTO_NO_PLAIN but does not allow plain connections */
#define SHOUT_TLS_RFC2818           ( 11) /* Use TLS for transport layer like HTTPS [RFC2818] does. */
#define SHOUT_TLS_RFC2817           ( 12) /* Use TLS via HTTP Upgrade:-header [RFC2817]. */
```
Since `SHOUT_TLS_AUTO` seems to be broken, we should not offer it. How is the situation with the explicit mode selection, does it work? How does "Auto" work?

src/engine/sidechain/shoutconnection.cpp
```diff
@@ -92,16 +92,6 @@ ShoutConnection::ShoutConnection(BroadcastProfilePtr profile,
         errorDialog(tr("Error setting non-blocking mode:"),
                 shout_get_error(m_pShout));
     }
-
-#ifdef SHOUT_TLS
-    // Libshout defaults to SHOUT_TLS_AUTO if build with SHOUT_TLS
-    // Sometimes autodetection fails, resulting into no metadata send
-    // https://github.com/mixxxdj/mixxx/issues/9599
```
daschuer, line 99: We should keep the reference to the bug.

src/preferences/dialog/dlgprefbroadcast.cpp
```diff
@@ -145,6 +145,14 @@ DlgPrefBroadcast::DlgPrefBroadcast(QWidget *parent,
      comboBoxEncodingChannels->addItem(tr("Stereo"),
              static_cast<int>(EncoderSettings::ChannelMode::STEREO));
 
+     // Encryption mode combobox
+     encryptionModeComboBox->addItem(tr("Required"),
+             static_cast<int>(BroadcastProfile::EncryptionMode::Required));
+     encryptionModeComboBox->addItem(tr("Preferred"),
+             static_cast<int>(BroadcastProfile::EncryptionMode::Preferred));
+     encryptionModeComboBox->addItem(tr("Disabled"),
+             static_cast<int>(BroadcastProfile::EncryptionMode::Disabled));
```
daschuer, line 154: I consider "Preferred" as a security risk. In case encryption fails, it should never fall back to plain. 

We may add a mode like "Try encryption" in case it works, it should be stored as "Required" and fail later if the server is no longer the verified pair.

src/preferences/broadcastprofile.cpp
```diff
@@ -294,6 +297,16 @@ bool BroadcastProfile::loadValues(const QString& filename) {
...
     m_port = XmlParse::selectNodeInt(doc, kPort);
     m_serverType = XmlParse::selectNodeQString(doc, kServertype);
+    switch (static_cast<EncryptionMode>(XmlParse::selectNodeInt(doc, kEncryptionMode))) {
+    case EncryptionMode::Preferred:
+        m_encryptionMode = EncryptionMode::Preferred;
+        break;
+    case EncryptionMode::Disabled:
+        m_encryptionMode = EncryptionMode::Disabled;
+        break;
+    default:
```
daschuer, line 307: What is the migration path? I think we should keep "Disabled" after upgrade. We may consider to keep the "not set" state that we have the chance to ask the user in pop up box in a later PR (or in this if you like)

daschuer (changes requested):

src/engine/sidechain/shoutconnection.cpp
acolombier: Is it still broken? It seems to be working fine on 2.4.6.

Explicit mode works well as well, but it require the user to know which RFC/implementation the server is using (startTLS vs full TLS), which depends of how the broadcast server is configured. Reading https://github.com/mixxxdj/mixxx/issues/8301, I don't think we want the user to have to know/care about this level of details

During my test, I tried edge/ingress level TLS (RFC2818), and container level (RFC2817), explicit option worked well as long as you picked the right one. Auto was able to detect the right setup nicely.

src/preferences/dialog/dlgprefbroadcast.cpp
acolombier: Yeah, that's a good point. I'm not sure how feasible this is tho, without requirement too much refactor. I might disable `Preferred` for now to allow us moving forward, wdyt?

daschuer: That works for me.

src/preferences/broadcastprofile.cpp
daschuer: I don't mind, my idea was to just check if this new option exists or not.

src/engine/sidechain/shoutconnection.cpp
JoergAtGithub: @daschuer Are you sure, that this is still broken in 2.4.6 ?

src/preferences/dialog/dlgprefbroadcast.cpp
JoergAtGithub: Yes, please go ahead!

acolombier: Removed now - I think the idea of "Try encryption" is great, we could pair that with the DNS check as well, which was mentioned during one of the trouble shouting issue we had around broadcasting.

src/preferences/broadcastprofile.cpp
acolombier: Updated - the migration path will now default to disable, which is the current behaviour as we don't support TLS in 2.6

src/engine/sidechain/shoutconnection.cpp
daschuer: This is still pending.

daschuer:
> Behaviour is pretty random and unstable with the current libshout version, but far more stable and expectable with https://github.com/mixxxdj/mixxx/pull/4723

It sound like we should enable the code only for libshout-idjc 2.4.6 and above, right?

acolombier:
Yes, I consider this PR depending on libshout-idjc 2.4.6, and on the TLS RFC being correctly detected on this version

daschuer:
OK, so you need to check `shout_version()` ...

daschuer:
Did you consider to add the version check to make this mergeable dependently?

acolombier:
Yes, and I consider this a dead end.
As you right fully pointed in [here](https://github.com/mixxxdj/mixxx/pull/4723#issuecomment-3904204829), there is likely a small minority that would be able to benefit this. I myself use purposefully a virtual environment on 24.04 to ensure we support our minimum requirements (tho I run a much newer system, on why I also build Mixxx for my leisure). This means the CI tests (and my development setup) wouldn't be able to test and guarantee the behaviour. 
As you also rightfully point in the comment linkled above, this feature would be made available to 2.7, which will bump support to 26.04. (2.7 will be released at least 6 months from 2.6, which is still not GA)

daschuer:
I have not understand what you propose. 

Why not add the version guard and have this PR from our to do list. Main user with the latest libshout-idjc version can benefit from it.

acolombier:
> Main user with the latest libshout-idjc version can benefit from it.

Please tell me how to do that on Ubuntu 24.04.

daschuer:
With the version guard an Ubuntu 24.04 user can still build main. In short Ubuntu 25.04 LTS will be released and users will automatically benefit from this feature here. 

By the way, can you explain the issue you had: 
> Behaviour is pretty random and unstable with the current libshout version. 

Which version did you test? What was the faulty behaviour?

daschuer:
It is probably this: 
```
libshout 2.4.2 (20190429)
* Rewrote TLS handling code and statemachine (#2244, #2298, #2301, #2303).
* Support OPTIONS for RFC 2817 TLS mode.
* Allow manual inspection of server's TLS certificate.
```

acolombier:
> In short Ubuntu 25.04 LTS will be released and users will automatically benefit from this feature here.

This is exactly my point. `main` a.k.a 2.7 will be released with the minimum requirement of 26.04. SO instead of releasing a disable feature, let's make sure it is enabled during the alpha (and eventually the beta) to gather as much feedback as possible.

> Which version did you test? What was the faulty behaviour?

This was 18 months ago, but I believe this would have been 2.4.1. `Rewrote TLS handling code and statemachine` sounds familiar to what would have been the issue, but I don't recall exactly what was the symptoms.

daschuer:
Ah ok. So we have a common agreement to get this one merged as soon as possible.

If you add the version guard now, I consider this ready for merge. A final test is missing if cause.

src/engine/sidechain/shoutconnection.cpp
daschuer: I don't have tested this. We can merge this PR once we have version guards for 2.4.6.
I am really concerned for regressions form this PR so let's move forward with this safety net. 
Once a version guard `shout_version()` is in place we can merge this.

src/preferences/dialog/dlgprefbroadcastdlg.ui
```diff
@@ -598,9 +607,9 @@
               <string notr="true">&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
 &lt;html&gt;&lt;head&gt;&lt;meta name=&quot;qrichtext&quot; content=&quot;1&quot; /&gt;&lt;style type=&quot;text/css&quot;&gt;
 p, li { white-space: pre-wrap; }
-&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'.SF NS Text'; font-size:13pt; font-weight:400; font-style:normal;&quot;&gt;
-&lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;span style=&quot; font-family:'Ubuntu'; font-size:10pt;&quot;&gt;  &lt;/span&gt;&lt;/p&gt;
-&lt;p style=&quot;-qt-paragraph-type:empty; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px; font-family:'Lucida Grande';&quot;&gt;&lt;br /&gt;&lt;/p&gt;&lt;/body&gt;&lt;/html&gt;</string>
+&lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'Fira Sans Semi-Light'; font-size:10pt; font-weight:400; font-style:normal;&quot;&gt;
```
daschuer: This seems to be an unrelated change.

daschuer (changes requested):
Is there a reason to not add a version guard? I think this will remove the deadlock with this PR.

acolombier:
> Is there a reason to not add a version guard? 

Yes - as per our policy, we will drop support for Ubuntu 24.04 in less than two months, so this extra work, as well as extra branching risk which is expected to even be removed before the beta starts.

> I think this will remove the deadlock with this PR.

IMO, removing the deadlock with this solution, means keeping the deadlock in #4723 so this will leave a lot of user waiting for #8301 behind (many user will still be on 24.04 and won't migrate till summer Point Release or even more, some are still on 22.04 - e.g PopOS pre-cosmic). 
While we won't be supporting them officially, we will likely provide a backport like we did for 22.04 and leaving them without TLS connection in 2026 is not serious.

daschuer:
All this discussion is also extra work. Isn't it only a runtime call in addition to `#ifdef SHOUT_TLS` we have already? 

> While we won't be supporting them officially, we will likely provide a backport like we did for 22.04 and leaving them without TLS connection in 2026 is not serious.

Not TLS due to the version guard? 

With runtime version guard: 
* We have reliable broadcasting independent from the users libshout-idjc version 
* We can collect real world experiences experiences with rolling Linux distros. 
* We can merge this independent of the decision we take in https://github.com/mixxxdj/mixxx/pull/4723 

Without the version guard: 
* We need to wait. 
* Users on old unsupported distros may suffer regressions

acolombier:
> All this discussion is also extra work. 

Agreed. The best would be to merge #4723 then this PR. 

> Not TLS due to the version guard?

Since we refuse to update our already vendored deps, TLS won't be available. This also means we would need to keep the branch in place in our code.

>  We have reliable broadcasting independent from the users libshout-idjc version

Calling unencrypted traffic - including broadcaster credentials - a "reliable broadcasting" experience is quite controversial. I certainly do not register myself, and wouldn't want to broadcast with Mixxx. 
Perhaps we could ask users on the issue if they would prefer to see their broadcasting credential compromised or if they prefer to take some risk during an apha, then beta period, which will likely last month and where we will be able to fix forward.

> Without the version guard:
>     * We need to wait.

With this guard, we give another reason not to update this version, which has been live for years and hasn't show any regression. 

Here is my proposal:

- Either we keep the deadlock till 26.04 is out and the min version libshout version is 2.4.6
- We ask the community if they would prefer to have an alpha version with TLS - and thus a safe way to broadcast - but the risk of an alpha, which is instability risk

Happy to quickly get a post ready on Discourse (maybe we even have [survey](https://github.com/discourse/discourse-surveys) enabled?) to remove the deadlock.
Curious to hear @mixxxdj/developers 's take here as this feature seems to be well expected?

daschuer:
Ah got it. You think our vendored version has no TLS. This is not true. It has.  That's why the compile time version guard is not sufficient. 

The compile time version guard only detects which headers have been used during compiling. This may not match the *.so the user has. That's why we need the runtime version guard as well, in any case, independent from the rest of the update strategy.

acolombier:
> You think our vendored version has no TLS. This is not true. It has.

It does, but it suffers from [an issue](https://github.com/acolombier/mixxx/blob/04c69c8525623cb4889db680def9e18f4c9df1fc/src/engine/sidechain/shoutconnection.cpp#L92) that have lead us to disable it. 
As I said in the top of this PR, I have tested this feature with the currently shipped version (2.4.1), and it does behave randomly, particularly fails after a few connecti/disconnect cycle, which I assume are the original symptoms leading to force disable TLS.

> This may not match the *.so the user has.

Appreciate I often get confused with our CMakeFile, but in this case, I believe it is clear that, in some case that we have decided due to two issues, [we ship our vendored fork of the lib
](https://github.com/acolombier/mixxx/blob/04c69c8525623cb4889db680def9e18f4c9df1fc/CMakeLists.txt#L4893). That is both the headers, which impact compilation, and the forked library (not sure if it is dynamically linked into a *.so or statically linked FWIW)

The whole reason I am pushing for vendoring 2.4.6 is that we can simplify that to:

- User have a version greater to 2.4.6, great then can use that (no need for version guard)
- User have version less than 2.4.1, we give them our vendored version

This means that our theoretical 24.04 backport will be shipped with the vendor lib, but the 26.04 won't, tho will have the version constrain in the .deb manifest.

daschuer:
I think your latest post is unrelated to the runtime version guard. Let me explain. 

We have here: 
```
#ifdef SHOUT_TLS
```
This does check if the header used at compile time has  SHOUT_TLS defined. Unfortunately this is not enough for a working encryption support, because we need in addition a certain version. Something similar to this pseudo code.  
```
if (shout_version() >= 2.4.6) 
```

This is my only demand here, because is behaves rock solid under all conditions with only a minimal change. 
Why not just add this her and merge?

acolombier:
Thanks, I understood what you ask and I know how to implement it. But this does not address my concern.

Again. doing this means that this feature would result in a no-op, except for user building Mixxx themselve with libshout > 2.4.6.
Non technically confident user would be yet again left behind, and wouldn't even be able to install the `.deb` build artifact since it is built with Ubuntu 24.04, which would use [our vendored version](https://github.com/mixxxdj/mixxx/actions/runs/22544301176/job/65304391566#step:19:194).

Let's please stop discussing about bringing any guards to this PR.

daschuer:
Your concerns are a compete different topic. 

We need the guard in any case if we want to release before all < 2.4.6 have disappeared. This is because distros sometime rip of libraries in our lib folder. It is just a safety guard that died not hard IMHO her. The no opt behaviour is desired for libshout >= 2.4.6 If users are left behind or not is a business decision. We should be first technically correct here.

src/engine/sidechain/shoutconnection.cpp
acolombier: This comment is now irrelevant since we've agreed not to use `SHOUT_TLS_AUTO`

src/engine/sidechain/shoutconnection.cpp
acolombier: > I am really concerned for regressions form this PR so let's move forward with this safety net.

There is no regression risk in this PR. 
Before, TLS was explicitly disabled. 
Now it can be explicitly enabled or disabled.

Only risk is for a user to experience regression when enabling TLS, but they have the option to get back to TLS being explicitly disabled.

Are you asking to force user with <2.4.6 to stay stuck with TLS disabled?

acolombier: Also https://github.com/mixxxdj/mixxx/pull/13285#discussion_r2173511461

Issue #11373, opened 2023-03-15
Rekordbox Library Tracks Disappear


rosswhyman:
### Bug Description

Exported RekordBox library to USB for use on CDJs etc.

When accessing this library in Mixxx (from the USB), large numbers of tracks are missing from playlists. After a few days of regular use it gradually gets worse until half the playlists are empty.

A PC or Mixxx restart does not fix the issue. Only solution I have found is to re-export the entire USB and then *most* of the tracks will return, but over time will gradually disappear again.

"Check for attached USBs" link doesn't seem to refresh it either.

Exported from RekordBox 6. Using Hercules DJ Controller Steel as MIDI controller and XONE43C as mixer.

Note that if I access the tracks directly within Mixxx (via the drive letter under Rekordbox in the Mixxx track browser), the tracks are there and play just fine. It's just the RB playlists within that drive that are missing tracks which is highly inconvenient!

Only similar issue report that I've found is number #10672, however changing the sorting doesn't seem to affect anything in my case.

Here are the RB export PDB files:
[rekordbox.zip](https://github.com/mixxxdj/mixxx/files/10981493/rekordbox.zip)

Thanks in advance, let me know if you need more information. Cheers

### Version

2.3.4

### OS

Windows 10

rosswhyman:
Small update: it seems to not lose so much of the data until I update the USB with new tracks or playlists. Once I've added to a USB that mixxx has already seen, there will suddenly be big gaps in playlists (some empty entirely). I do have tracks duplicated across several playlists, if that helps diagnose the issue.

Additionally I've tried several different USBs with copies of the same RB library and they seem to lose the information in unison, so I don't think it's the USBs directly, more likely something to do with how the RB library interacts with Mixxx.

Swiftb0y:
may be fixed by #13293 in 2.4.2 (not released yet). if not feel free to reopen.

Swiftb0y:
This may be fixed by #15745

Issue #10923, opened 2022-09-28
Randomly missing playlists in Rekordbox formated drives


emmetiray:
When i go to my Rekordbox usb devices in Mixxx there are always some playlists missing. 
![rekordbox01](https://user-images.githubusercontent.com/114584321/192774780-3dc78bcc-0763-4c30-94fe-c75c0aa0486a.JPG)

Don't know if this is a known issue.

Cheers Jakob

Holzhaus:
Thanks for reporting, but some information is missing in this bug report. For example, what OS and what Mixxx version do you use? (GitHub should ask for this information when you choose "Bug Report" when opening a new issue).

In this case, it would also be helpful if you could attach a `mixxx.log` logfile. Ideally, we'd also appreciate if you could attach the PDB file from the USB drive in case it's some issue with our parser.

emmetiray:
thx for the infos. and great work by the way. at the moment i'm using mixxx under win10 but my plan is to install
it on the steam deck. so it will be steamos/arch linux soon.


![mixxx_version](https://user-images.githubusercontent.com/114584321/192844757-e1b3e45c-f153-40c7-b3e5-a6efec57f5d1.JPG)
[pdb.zip](https://github.com/mixxxdj/mixxx/files/9667399/pdb.zip)
[mixxx.log](https://github.com/mixxxdj/mixxx/files/9667419/mixxx.log)

emmetiray:
2.4 windows alpha same problem.

emmetiray:
its working fine on rekordbox and my xdj700 btw.

Swiftb0y:
@emmetiray can you elaborate what playlists are missing exactly? The database is too big to find irregularities by hand. I'd need to know what to look for. Can you tell me what info is missing exactly? Is it just the playlist name or also its content? Does it have anymore identifying data (how many tracks did it contain, what tracks for example, etc). Are these playlists missing completely or just empty like in your screenshot?

Holzhaus:
Maybe this is due to non-ascii chars? For example, is the playlist `BEAT PVP - BEAT PVP 1ª EDIÇÃO`  one of the missing ones?

emmetiray:
@Swiftb0y @Holzhaus 
[...]

PR #13293, opened 2024-05-26
chore: update rekordbox kaitai definitions

### LINKED ISSUE #13291: Unicode DeviceSQL strings in PDB are UTF-16-BE, not UTF-16-LE
### Bug Description

Mixxx uses kaitai struct to extract data from rekordbox pdb files.
rekordbox data binary was reversed by another project called crate-digger.
Mixxx uses an old copy of crate-pdb.ksy file for parsing pdb files, which needs to be updated.

https://github.com/Deep-Symmetry/crate-digger/issues/14

image
As shown in the image, Japanese characters are not properly displayed.

### Version

_No response_

### OS

_No response_

Swiftb0y:
Should also fix #13291.

Lets see if this builds without updating any additional changes.


Swiftb0y:
Apparently not, seems like we need an updated kaitaistruct runtime. It also doesn't seem appropriate to vendor that anymore. Would `ExternalProject_Add` instead be an option? Can somebody help me with the required CMake legwork for that?

JoergAtGithub:
We have it in VCPKG

Swiftb0y:
yeah, but thats about it, its not packaged on any linux distro. https://repology.org/project/kaitai-struct-cpp-stl-runtime/versions

daschuer:
We cannot download external sources on Launchpad. So let's just update our vendored version.

lib/kaitai/kaitai/kaitaistream.h
```diff
@@ -162,7 +167,11 @@ class kstream {
 
     static std::string bytes_strip_right(std::string src, char pad_byte);
     static std::string bytes_terminate(std::string src, char term, bool include);
-    static std::string bytes_to_str(std::string src, std::string src_enc);
+    static std::string bytes_to_str(const std::string src, const char *src_enc);
+    // workaround for https://github.com/kaitai-io/kaitai_struct_cpp_stl_runtime/issues/67
+    static std::string bytes_to_str(const std::string src, const std::string& enc) {
+      return bytes_to_str(src, enc.c_str());                       
+    }
```
Swiftb0y, line 174: only changed I made to the kaitai files except copy-pasting in the newest commit and moving them to match the includes the kaitai-compiler generates.

src/library/rekordbox/kaitaistructs/rekordbox_anlz.ksy
```diff
@@ -393,77 +411,151 @@ types:
...
+      c:
+        value: len_entries
+      # mask:
+      #   value: |
+      #     [
+      #       (0xCB+c).as<u1>, (0xE1+c).as<u1>, (0xEE+c).as<u1>, (0xFA+c).as<s1>, (0xE5+c).as<s1>, (0xEE+c).as<s1>, (0xAD+c).as<s1>, (0xEE+c).as<s1>,
+      #       (0xE9+c).as<u1>, (0xD2+c).as<u1>, (0xE9+c).as<u1>, (0xEB+c).as<s1>, (0xE1+c).as<s1>, (0xE9+c).as<s1>, (0xF3+c).as<s1>, (0xE8+c).as<s1>,
+      #       (0xE9+c).as<u1>, (0xF4+c).as<u1>, (0xE1+c).as<u1>
+      #     ].as<bytes>
+      raw_mood:
```
Swiftb0y, line 433: see comment

src/library/rekordbox/rekordboxfeature.cpp
```diff
@@ -523,82 +524,57 @@ QString parseDeviceDB(mixxx::DbConnectionPoolPtr dbConnectionPool, TreeItem* dev
...
-                        reckordboxDB.tables()->begin();
-                table != reckordboxDB.tables()->end();
-                ++table) {
-            if ((*table)->type() == tableOrder[tableOrderIndex]) {
-                uint16_t lastIndex = (*table)->last_page()->index();
-                rekordbox_pdb_t::page_ref_t* currentRef = (*table)->first_page();
+        for (const auto& table : *rekordboxDB.tables()) {
+            if (table->type() == tableOrder[tableOrderIndex]) {
+                uint16_t lastIndex = table->last_page()->index();
+                rekordbox_pdb_t::page_ref_t* currentRef = table->first_page();
```
Swiftb0y, line 530: this didn't build anymore because the `table_t` changed to a unique_ptr. I took the liberty and modernized the loop in the same step (same below).

Swiftb0y:
@0xsuk can you [test this PR](https://github.com/mixxxdj/mixxx/wiki/Testing#github-pull-requests) to see whether this doesn't break anything and fixes your string-encoding issue? Thank you.

0xsuk:
Everything looks good after I changed UTF-16BE to UTF-16LE in src/library/rekordbox/rekordboxfeature.cpp line 259.

0xsuk:
image
Thank you :)

Swiftb0y:
whoops. good catch, I assumed kaitai would take care of the conversion automatically. Fixed

Swiftb0y:
Ready for review. I suggest to go through on a commit by commit basis. The PR looks bigger as it is because most of the changes are either auto-generated or artifacts because I moved files around.

Swiftb0y:
friendly ping, would be nice if this made it into 2.4.2. I've highlighted the significant changes above.

CMakeLists.txt
```diff
@@ -2508,12 +2486,20 @@ target_link_libraries(mixxx-lib PRIVATE FpClassify)
...
+  lib/rekordbox-metadata/rekordbox_anlz.cpp
+)
+target_include_directories(rekordbox_metadata SYSTEM PUBLIC lib/rekordbox-metadata)
+target_link_libraries(mixxx-lib PRIVATE rekordbox_metadata)
+
 # Kaitai for reading Rekordbox libraries
 add_library(Kaitai STATIC EXCLUDE_FROM_ALL
-  lib/kaitai/kaitaistream.cpp
+  lib/kaitai/kaitai/kaitaistream.cpp
 )
```
Swiftb0y, line 2499: I got a lot of code quality complaints on the output of the kaitai compiler, so I moved the generated code to `lib/` (first and foremost because the linters already ignore that and second because it also makes sense to treat autogenerated code that way).

daschuer (approved):
The code looks good and I rely on the tests done by @0xsuk
Thank you.

daschuer:
How important is the fix? Does this rectify a release?

Swiftb0y:
Thank you. 

> How important is the fix? Does this rectify a release?

Well this PR contains two fixes, the string decoding issue reported in #13291. The second one concerns history playlists created by rekordbox. I don't even think we ingest those.

I don't think an extra release is warranted as only the string decoding issue is affected, judging by the delay with which this was reported, I wouldn't say its critical enough to justify the effort.

brunchboy:
It’s not just history playlists. Any table which has had deletions over time will likely be missing a few rows at the end without the structure fix.

brunchboy:
And not just deletions, rows are sometimes not present in row groups for inscrutable reasons. The fact that the problem was noticed in the history playlist table shows it could happen anywhere. We need to scan the entire last row group for tracks marked as present, not assume we can stop scanning based on the value of `num_rows`.

brunchboy:
Which is not to say that it necessarily merits an urgent release, I have no idea how many people rely on scanning rekordbox USBs in the first place. 😄

Swiftb0y:
Ah, thank you. In that case it would probably justify a release. Considering there is also #13309 #11373 and #10923 which all could be caused by that issue.

brunchboy:
Hmm, those all could be impacted but they sound worse than this! Please feel free to ping me if this doesn’t fix those other issues and perhaps I can help investigate the problematic export files.

Swiftb0y:
That would be wonderful. We can close these issues once we have a release with the fix and ask the respective issue authors to check again. In case it still happens, I would refer them to you (where exactly? crate-digger repo / deep-symmetry zulip) if thats alright for you.

brunchboy:
Let’s start on the Zulip instance and then I can create issues as appropriate.

PR #13294, opened 2024-05-26
Fix clicks at loop-out when looping into lead-in

daschuer:
This fixes a regression from f0870ff3c19f10bacaa519787c5ddeb840450676


ronso0:
Can you please provide steps to reproduce/check the fix?

Here's a commit with a comment, and another small tweak
7570cdaa39

daschuer:
Here is the receipt to reproduce. 
* Use a monotone test track like a 440 Hz sine tone. 
* Create a loop starting at lead in 
-> hear the click sound at loop out. You may also record it and look at it using audacity 

With this PR applied the click is smoothed out.

ronso0:
I don't hear the clicks tbh, but I see the fading to zero.
LGTM, thanks!

Issue #13248, opened 2024-05-20
Replace macos-11 runner image on GitHub Actions


Holzhaus:
The following announcement was sent out by GitHub via email:

> The macOS 11 runner image will be removed by 6/28/24. To raise awareness of the upcoming removal, jobs using macOS 11 will temporarily fail during scheduled time periods defined below:
>
> - June 17 2024, 8:00 AM - 2:00 PM EST
> - June 19 2024, 12:00 PM - 6:00 PM EST
> - June 24 2024, 3:00 AM - 9:00 PM EST
> - June 26 2024, 8:00 AM - 2:00 PM EST
>
> #### What you need to do
>
> Jobs using the macos-11 YAML workflow label should be updated to macos-14, macos-13, macos-12, or macos-latest. You can always get up-to-date information on our tools by reading about the software in the [runner images repository](https://github.com/actions/runner-images). Please contact GitHub [Support](https://support.github.com/contact?form%5Bsubject%5D=Re%3A+GitHub+Actions&tags=dotcom-contact-params) if you run into any problems or need help.

We actually use the `macos-11` runner (e. g., in the Build workflow for macOS  ARM64) and need come up with a plan how to replace it.

Swiftb0y:
fixed in #13296

PR #13296, opened 2024-05-27
Move to GitHub worflow runner macos-12

daschuer:
This fixes: https://github.com/mixxxdj/mixxx/issues/13248
Deadline is 2024-06-28

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,8 +60,8 @@ jobs:
             artifacts_path: build/*.deb
             artifacts_slug: ubuntu-jammy
             qt_qpa_platform: offscreen
-          - name: macOS 11 x64
-            os: macos-11
+          - name: macOS 12 x64
+            os: macos-12
             cmake_args: >-
               -DBULK=ON
               -DCOREAUDIO=ON
@@ -83,8 +83,8 @@ jobs:
             artifacts_path: build/*.dmg
             artifacts_slug: macos-macosintel
             qt_qpa_platform: offscreen
-          - name: macOS 11 arm64
-            os: macos-11
+          - name: macOS 12 arm64
+            os: macos-12
             cmake_args: >-
               -DBULK=ON
               -DCOREAUDIO=ON
--- a/src/test/globaltrackcache_test.cpp
+++ b/src/test/globaltrackcache_test.cpp
@@ -153,7 +153,15 @@ TEST_F(GlobalTrackCacheTest, concurrentDelete) {
     // NOTE(2019-12-14, uklotzde): On Travis and macOS executing 10_000
     // iterations takes ~1 sec. In order to safely finish this test within
     // the timeout limit of 30 sec. we use 20 * 10_000 = 200_000 iterations.
-    for (int i = 0; i < 200000; ++i) {
+    //
+    // NOTE(2024-06-03, daschuer): Reduced to 100000 because the we hit a
+    // timeout on the macos-12 GitHub workflow runner
+    // With 200000 we had:
+    // ubuntu-22.04 0.73 sec
+    // windows-2019 9.86 sec
+    // macos-11 5.81 sec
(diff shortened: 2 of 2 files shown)


daschuer:
The test fails rather strange. 
``` 
360/812 Test #360: GlobalTrackCacheTest.concurrentDelete ......................................................***Timeout  45.02 sec
dlopen error : dlopen(libjack.0.dylib, 0x0001): tried: 'libjack.0.dylib' (no such file), '/usr/local/lib/libjack.0.dylib' (no such file), '/usr/lib/libjack.0.dylib' (no such file), '/Users/runner/work/mixxx/mixxx/libjack.0.dylib' (no such file) 
dlopen error : dlopen(/usr/local/lib/libjack.0.dylib, 0x0001): tried: '/usr/local/lib/libjack.0.dylib' (no such file) 
Note: Google Test filter = GlobalTrackCacheTest.concurrentDelete
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from GlobalTrackCacheTest
[ RUN      ] GlobalTrackCacheTest.concurrentDelete
info [0x6000037f02b0] GlobalTrackCache - Creating instance
```
For my understanding libjack is linked statically, so there should be no reason for loading the dylib at runtime. It is also interesting why the arm build is not failing.

daschuer:
https://github.com/jackaudio/jack2/issues/862 might be related. 
@fwcd Do you have an idea?

fwcd:
The linked issue looks unrelated, library validation is different from not finding the library. That shouldn't be needed in the first place though, since we're almost exclusively using static libraries from vcpkg, right?

You could check the configure logs, perhaps CMake decided to pick a system library over vcpkg for JACK, in that case I would have expected link errors at an earlier stage though.

Swiftb0y:
Yes, I'm pretty sure the dlopen issue is completely unrelated, I've seen that numerous times on other macOS CI failures which had nothing to do with jack. I'm pretty sure the `concurrentDelete` test has some threading-related issues which surface now.

fwcd:
I remember seeing library loading errors with JACK when launching Mixxx normally on macOS too, even though we don't even use JACK on macOS. So probably something to figure out.

But yeah, the test failure is probably completely unrelated.

JoergAtGithub:
> You could check the configure logs, perhaps CMake decided to pick a system library over vcpkg for JACK, in that case I would have expected link errors at an earlier stage though.

Our vcpkg.json doesn't contain jack2

daschuer:
The `dlopen error : dlopen(libjack.0.dylib, 0x0001):``is originated from here: 
https://github.com/jackaudio/jack2/blob/c46c1b16e0eabbcf55ef69b0ffb96dfe16521cfa/common/JackWeakAPI.c#L55

daschuer:
We do build jack2 with vcpkg. Int comes in as dependency. I am currently struggling to find out how, because Portaudio does not have this dependency. 
```
2024-05-27T05:28:03.9724760Z -- Found JACK: /Users/runner/buildenv/mixxx-deps-2.4-x64-osx-min1012-release-07b9859/installed/x64-osx-min1012-release/lib/libjack.a (found version "1.9") 
```

Swiftb0y:
Let me emphasize again that the jack dlopen error has nothing to do with the failing test!

daschuer:
> Let me emphasize again that the jack dlopen error has nothing to do with the failing test!

Confirmed. That is not even an error. Portaudio Probes the JACK API like all others. Unfortunately Jack has decided to print this error unconditionally. I think I will file a bug against Jack.

daschuer:
This is the upstream report: https://github.com/jackaudio/jack2/issues/977

daschuer:
This is now working but required a good amount of testing. 24 Days left.

Swiftb0y (approved):
LGTM. Thank you for the testing.

PR #13301, opened 2024-05-29
show 'hide menubar?' dialog on Linux only when the menubar will be in the window

ronso0:
i.e. only if global menu is not supported or if it is and we're when going fullscreen

fixes a regression introduced when fixing conflicts before merging #11526


daschuer (approved):
I can confirm it is fixed with Chinamon and Unity on Ubuntu Focal. 
Thank you.

PR #13302, opened 2024-05-29
Preferences: apply only in slotApply()

ronso0:
one more step to fix #7440 

Is there a specific reason to instantly apply ReplayGain settings (Normalization)?


daschuer:
> Is there a specific reason to instantly apply ReplayGain settings (Normalization)?

Yes. You may play a track with either ReplayGain Applied or without and adjust the Gain values watching the VU Meters.
If we Apply() with an extra button changes are applied without a gentle ramp. There is a Risk to blow your speakers because of insane gains. Maybe it is a good idea to place a note at the slider box that output is immediately affected.
The the jump problem already applies to the checkbox. This should probably be covered in the note.

ronso0:
That makes sense. I would rather have a checkbox “Apply settings immediately”, but I don't feel like refactoring this it at the moment.
I'll revert and only keep the groupbox / tabstop tweaks.

daschuer (approved):
LGTM, Thank you.

PR #13304, opened 2024-05-30
Recording: with empty config, save default split size immediately

ronso0:
reported in the forum
https://mixxx.discourse.group/t/recording-cut-into-parts/29631

### DIFF
--- a/src/preferences/dialog/dlgprefrecord.cpp
+++ b/src/preferences/dialog/dlgprefrecord.cpp
@@ -92,10 +92,10 @@ DlgPrefRecord::DlgPrefRecord(QWidget* parent, UserSettingsPointer pConfig)
     if (index >= 0) {
         // Set file split size
         comboBoxSplitting->setCurrentIndex(index);
-    }
-    else {
+    } else {
         //Use max RIFF size (4GB) as default index, since usually people don't want to split.
         comboBoxSplitting->setCurrentIndex(4);
+        saveSplitSize();
     }
 
     setScrollSafeGuard(comboBoxSplitting);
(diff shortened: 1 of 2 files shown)


daschuer (changes requested):
Thank you for the fix. Can you please also adjust this: 
https://github.com/mixxxdj/mixxx/blob/18a3e4a75f776969323b860828df1437e1dca534/src/recording/recordingmanager.cpp#L328

A common class for all the recording defaults would be also nice but probably out of the scope of this bug-fix PR:
"preferences/replaygainsettings.cpp"

ronso0:
Oh, right.
Done.

src/preferences/dialog/dlgprefrecord.cpp
```diff
@@ -92,10 +92,12 @@ DlgPrefRecord::DlgPrefRecord(QWidget* parent, UserSettingsPointer pConfig)
     if (index >= 0) {
         // Set file split size
         comboBoxSplitting->setCurrentIndex(index);
-    }
-    else {
+    } else {
         //Use max RIFF size (4GB) as default index, since usually people don't want to split.
         comboBoxSplitting->setCurrentIndex(4);
+        // Save the value immediately, otherwise RecordingManager would fall back
```
daschuer, line 98: The comment is now outdated.

ronso0: Yeah, well, I overlooked that we actually only need to use the same default in both unit.
Saving to the config doesn't hurt though.

daschuer (approved):
LGTM, Thank you.

PR #13306, opened 2024-05-30
Move download, SHA256 check and unzip of Windows buildenv from windows_buildenv.bat to CMakelists.txt

JoergAtGithub:
This move the VCPKG buildenv download, SHA256 verification and unzipping into CMakelists.txt. This removes the need for third party tools like 7zip and Powershell (see #11524 for reference).

For the developer everything should behave as before. No change in the build instructions is neccessary - just the tasks mentioned above are executed later.

Out of scope of this PR is:

- macOS
- Replacing windows_buildenv.bat and CMakeSettings.json by CMake Presets


CMakeLists.txt
```diff
@@ -47,6 +47,67 @@ if(POLICY CMP0135)
   cmake_policy(SET CMP0135 NEW)
 endif()
 
+
+if(WIN32)
+    if(DEFINED ENV{BUILDENV_BASEPATH})
+        set(BUILDENV_BASEPATH "$ENV{BUILDENV_BASEPATH}\\")
```
daschuer, line 53: In other places we ignore environment variables if they are explicit passed to cmake via -D. 
I think we should follow the same priority here and below with BUILDENV_URL

CMakeLists.txt
```diff
@@ -47,6 +47,67 @@ if(POLICY CMP0135)
...
+    else(NOT DEFINED BUILDENV_BASEPATH)
+        set(BUILDENV_BASEPATH "${CMAKE_SOURCE_DIR}/buildenv/")
```
daschuer, line 55: I think this is redundant with MIXXX_VCPKG_ROOT.

CMakeLists.txt
```diff
@@ -47,6 +47,67 @@ if(POLICY CMP0135)
...
+    if(DEFINED ENV{BUILDENV_URL})
+        set(BUILDENV_URL "$ENV{BUILDENV_URL}")
+    elseif(NOT DEFINED BUILDENV_URL)
+        message(FATAL_ERROR "BUILDENV_URL not specified")
+    endif()
+
+    # Extract the filename from the URL
+    get_filename_component(FILENAME_WITH_EXTENSION "${BUILDENV_URL}" NAME)
+
+    # Remove the extension from the filename (note that our buildenv names contain two dots)
```
daschuer, line 67: Adding an example to the comment would be helpful.

CMakeLists.txt
daschuer: I wonder how it is done in case of git worktrees. I can imagine to have only one buildenv folder that is accessed by all worktrees. Did you consider that? That will safe a lot HD and bandwidth.

daschuer (changes requested):
Thank for adopting this.

CMakeLists.txt
JoergAtGithub: Neither the variable nor the directory structure is changed, in this PR. This is therefore out of scope.
But I think it should help in this usecase, that the variable is now an environment variable, instead of a batch file internal variable. This allows the user to overwrite it after executing the script.

JoergAtGithub: I will replace MIXXX_VCPKG_ROOT with BUILDENV_BASEPATH, than the variable naming scheme becomes consistent.

daschuer: It depends. I have no special demands here, just note that the VCPKG_ROOT env variable is the system wide default. We have created our own derived from that name to be independent from it.

JoergAtGithub: Looking at this again, it's `BUILDENV_PATH` which is redundant to `MIXXX_VCPKG_ROOT` and not `BUILDENV_BASEPATH`.

JoergAtGithub:
@daschuer Friendly Ping

JoergAtGithub:
@daschuer Can we merge this?

CMakeLists.txt
```diff
@@ -47,6 +47,68 @@ if(POLICY CMP0135)
   cmake_policy(SET CMP0135 NEW)
 endif()
 
+
+if(WIN32)
```
fwcd, line 51: I know you mentioned it as out-of-scope, but I agree that sharing this logic with macOS would be great, given that the vcpkg envs are structured practically identically as far as I understand.

JoergAtGithub: It would be great if you could create a follow-up PR for macOS.

JoergAtGithub:
@daschuer Can we merge this now?

daschuer (changes requested):
This only works if we make sure   
"MIXXX_VCPKG_ROOT = BUILDENV_BASEPATH/BUILDENV_NAME"

What are the use cases if this is not the case? 

If we tray to use an alternative build environment "MIXXX_VCPKG_ROOT" needs to be set. 
If that exists, I think we can skip the whole evaluation of "BUILDENV_BASEPATH/BUILDENV_NAME"

In case it does not exists, I think there is no point in downloading to a diffrent folder than "MIXXX_VCPKG_ROOT". 
So we may fail fatal in that case. However if we are in a shell where "BUILDENV_BASEPATH/BUILDENV_NAME" is set it will allways fail. 

Conclusion we may only use "MIXXX_VCPKG_ROOT" and may dispose "BUILDENV_BASEPATH/BUILDENV_NAME"

What do you think?

acolombier:
/softfix:squash
```
feat: move buildenv setup to CMakeFile
```

acolombier:
/softfix
```
feat: move buildenv setup to CMakeFile
```

acolombier:
@daschuer were your change requests addressed?  Merge?

CMakeLists.txt
```diff
@@ -48,6 +48,84 @@ if(POLICY CMP0135)
   cmake_policy(SET CMP0135 NEW)
 endif()
 
+if(WIN32)
+  if(DEFINED ENV{BUILDENV_BASEPATH})
+    set(BUILDENV_BASEPATH "$ENV{BUILDENV_BASEPATH}\\")
+  else(NOT DEFINED BUILDENV_BASEPATH)
+    set(BUILDENV_BASEPATH "${CMAKE_SOURCE_DIR}/buildenv/")
+  endif()
```
daschuer, line 56: In other places we ignore environment variables if they are explicit passed to cmake via -D.
I think we should follow the same priority here and below with BUILDENV_URL

```suggestion
  if(NOT DEFINED BUILDENV_BASEPATH)
    if(DEFINED ENV{BUILDENV_BASEPATH})
      set(BUILDENV_BASEPATH "$ENV{BUILDENV_BASEPATH}\\")
    else()
      set(BUILDENV_BASEPATH "${CMAKE_SOURCE_DIR}/buildenv/")
    endif()
  endif()
```
We may even put this repeating pattern into a helper function.

daschuer: It turns out that there is even a CMake bug in my mininal test. CMake forgets the variable when testing for the same variable on the environment.   

```
message(STATUS "-D ${BUILDENV_BASEPATH} ENV $ENV{BUILDENV_BASEPATH}")

if(NOT DEFINED BUILDENV_BASEPATH)
  message(status "1. BUILDENV_BASEPATH is not defined.")
endif()

if(DEFINED ENV{BUILDENV_BASEPATH})
  set(BUILDENV_BASEPATH "$ENV{BUILDENV_BASEPATH}\\")
else(NOT DEFINED BUILDENV_BASEPATH)
  message(status "2. BUILDENV_BASEPATH is not defined.")
  set(BUILDENV_BASEPATH "${CMAKE_SOURCE_DIR}/buildenv/")
endif()
  
message(FATAL_ERROR "3. ${BUILDENV_BASEPATH}")
```

```
$ cmake ../.. -DBUILDENV_BASEPATH="Hallo"
-- CMAKE_VERSION: 3.22.1
-- -D Hallo ENV 
status2. BUILDENV_BASEPATH is not defined.
CMake Error at CMakeLists.txt:64 (message):
  3.  /home/daniel/workspace/mixxx2/buildenv/
```

daschuer (changes requested):

acolombier:
Sorry for the early fixup @JoergAtGithub , I missred that PR and thought it was ready to go

JoergAtGithub:
Restored original commits

JoergAtGithub:
All done!

Eve00000:
> All done!

Hi Jörg,
So if I understand it correct, you moved the downloading and extraction of the dependencies from the bat to cmake.
While working on these, can you please make the error-messages more clrear?
Before a lot of errors where covered under the message 'did you execute the windows_buildenv.bat' without mentioning the right error. 
Can you also add a note at the beginning of the cmake to explain whitch lines need to be adapted if the dependencies are downloaded manual, and maybe some line to enter the name and sha of the latest dependencies that were downloaded manual.
Thank you.

JoergAtGithub:
@Eve00000 The code with the message 'did you execute the windows_buildenv.bat' is not touched by this PR. This is therefore out of scope. The final goal is of cause to get rid of this batch file at all, but this is up to future PRs.

There is no need to modify the CMakelists.txt code when you use a zip file from other sources (a GitHub PR build artifact). Just manual download the ZIP file in the directory specified as BUILDENV_BASEPATH, set BUILDENV_NAME to the name of these manual downloaded file(without the file extension .zip).
For a command line build you could specify a manual downloaded buildenv zip by:
`cmake ../ -DBUILDENV_BASEPATH="D:\mixxx\buildenv" -DBUILDENV_NAME="mixxx-deps-2.5-x64-windows-c15790e"`
using an VisualStudio, you will find the same CMake settings in the GUI, where you can overwrite them.

Eve00000:
Cool!
I will give it a go this w-e.

daschuer:
@Eve00000 while testing, can you keep an eye in the documentation? I can imagine that some wiki docks needs to be adjusted, error messages rephrased.

Eve00000:
> eye in the documentation? 

I will do.

Eve00000:
Testing PR 13306  Move download, SHA256 check and unzip of Windows buildenv from windows_buildenv.bat to CMakelists.txt

- installed github cli
- on windows in gitbash all instructions need winpty prefix

- set mixxxdj/mixxx as default
-> winpty gh repo set-default 
-> choose mixxxdj/mixxx

-> winpty gh pr checkout 13306
gh downloads pr and activates it

deleted complete buildenv
-> downloaded latest dependencies zip from https://downloads.mixxx.org/dependencies/2.6/Windows/
mixxx-deps-2.6-x64-windows-e6ce218.zip
correction: executed tools\windows_buildenv.bat
opened visual studio
cmake starts -> downloads the latest 2.5 deps -> unzips and deletes the zip


I had an error error
CMake Error at D:\mixxx-git\mixxx\CMakeLists.txt:64 (message):
  BUILDENV_URL not specified
-> because I did not execute the tools\windows_buildenv.bat
-> this stays as before

JoergAtGithub:
@Eve00000 Thank you very much for testing! I improved the error messages and also improved the support for the case that you change the BUILDENV_NAME after running windows_buildenv.bat.
Regarding your questions Q1...Q3, these are not related to this PR. Therefore I will not answer them here. Please ask these on Zulip or open Issues for them.

Eve00000:
> I improved ....

cool! tell me what you want me to do?
test all over with making the same mistake not running the bat?

JoergAtGithub:
It would be great, if you could test the same cases as before, and tell me if the messages are now better understandable for you.

Eve00000:
> It would be great, if you could test the same cases as before, and tell me if the messages are now better understandable for you.

with pleasure

Eve00000:
Extra: 
Download through cmake is much faster than weit bat! -> SUPER

errors (before bat):
CMake Error at D:\mixxx-git\mixxx\CMakeLists.txt:461 (include):
  include could not find requested file:

    DefaultOption
CMake Error at D:\mixxx-git\mixxx\CMakeLists.txt:462 (include):
  include could not find requested file:

    IsStaticLibrary
CMake Error at D:\mixxx-git\mixxx\CMakeLists.txt:928 (default_option):
  Unknown CMake command "default_option".
  
  in output: (big list)
  1> [CMake] CMake Error at D:\mixxx-git\mixxx\CMakeLists.txt:928 (default_option):
1> [CMake]   Unknown CMake command "default_option".
1> [CMake] 
1> [CMake] 
1> [CMake] -- Configuring incomplete, errors occurred!
1> 'C:\WINDOWS\system32\cmd.exe' '/c "%SYSTEMROOT%\System32\chcp.com 65001 >NUL && "C:\PROGRAM FILES\MICROSOFT VISUAL STUDIO\2022\COMMUNITY\COMMON7\IDE\COMMONEXTENSIONS\MICROSOFT\CMAKE\CMake\bin\cmake.exe"  -G "Ninja"  -DCMAKE_BUILD_TYPE:STRING="Debug" -DCMAKE_INSTALL_PREFIX:PATH="D:\mixxx-git\mixxx\install\x64__off" -DCMAKE_TOOLCHAIN_FILE:FILEPATH="D:\mixxx-git\mixxx\buildenv\mixxx-deps-2.5-x64-windows-c15790e\scripts\buildsystems\vcpkg.cmake" -DMIXXX_VCPKG_ROOT:STRING="D:\mixxx-git\mixxx\buildenv\mixxx-deps-2.5-x64-windows-c15790e" -DBUILDENV_BASEPATH:STRING="D:\mixxx-git\mixxx\buildenv" -DBUILDENV_URL:STRING="https://downloads.mixxx.org/dependencies/2.5/Windows/mixxx-deps-2.5-x64-windows-c15790e.zip" -DBUILDENV_SHA256:STRING="138e4685ec73c6a6a509f71f8573be581403b091e4ecea2314df2cc79f9720b9" -DBATTERY:BOOL="True" -DBROADCAST:BOOL="True" -DBULK:BOOL="True" -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL="True" -DDEBUG_ASSERTIONS_FATAL:BOOL="True" -DFFMPEG:BOOL="False" -DHID:BOOL="True" -DHSS1394:BOOL="True" -DKEYFINDER:BOOL="False" -DLOCALECOMPARE:BOOL="True" -DLILV:BOOL="True" -DMAD:BOOL="True" -DMEDIAFOUNDATION:BOOL="True" -DMODPLUG:BOOL="True" -DOPUS:BOOL="True" -DOPTIMIZE:STRING="off" -DQT6:BOOL="True" -DQTKEYCHAIN:BOOL="True" -DSTATIC_DEPS:BOOL="False" -DVCPKG_TARGET_TRIPLET:STRING="x64-windows" -DVINYLCONTROL:BOOL="True" -DWAVPACK:BOOL="True" -DCMAKE_C_COMPILER:FILEPATH="C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.43.34808/bin/Hostx64/x64/cl.exe" -DCMAKE_CXX_COMPILER:FILEPATH="C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.43.34808/bin/Hostx64/x64/cl.exe"   -DCMAKE_MAKE_PROGRAM="C:\PROGRAM FILES\MICROSOFT VISUAL STUDIO\2022\COMMUNITY\COMMON7\IDE\COMMONEXTENSIONS\MICROSOFT\CMAKE\Ninja\ninja.exe" "D:\mixxx-git\mixxx" 2>&1"' execution failed with error: ''C:\WINDOWS\system32\cmd.exe' '/c "%SYSTEMROOT%\System32\chcp.com 65001 >NUL && "C:\PROGRAM FILES\MICROSOFT VISUAL STUDIO\2022\COMMUNITY\COMMON7\IDE\COMMONEXTENSIONS\MICROSOFT\CMAKE\CMake\bin\cmake.exe"  -G "Ninja"  -DCMAKE_BUILD_TYPE:STRING="Debug" -DCMAKE_INSTALL_PREFIX:PATH="D:\mixxx-git\mixxx\install\x64__off" -DCMAKE_TOOLCHAIN_FILE:FILEPATH="D:\mixxx-git\mixxx\buildenv\mixxx-deps-2.5-x64-windows-c15790e\scripts\buildsystems\vcpkg.cmake" -DMIXXX_VCPKG_ROOT:STRING="D:\mixxx-git\mixxx\buildenv\mixxx-deps-2.5-x64-windows-c15790e" -DBUILDENV_BASEPATH:STRING="D:\mixxx-git\mixxx\buildenv" -DBUILDENV_URL:STRING="https://downloads.mixxx.org/dependencies/2.5/Windows/mixxx-deps-2.5-x64-windows-c15790e.zip" -DBUILDENV_SHA256:STRING="138e4685ec73c6a6a509f71f8573be581403b091e4ecea2314df2cc79f9720b9" -DBATTERY:BOOL="True" -DBROADCAST:BOOL="True" -DBULK:BOOL="True" -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL="True" -DDEBUG_ASSERTIONS_FATAL:BOOL="True" -DFFMPEG:BOOL="False" -DHID:BOOL="True" -DHSS1394:BOOL="True" -DKEYFINDER:BOOL="False" -DLOCALECOMPARE:BOOL="True" -DLILV:BOOL="True" -DMAD:BOOL="True" -DMEDIAFOUNDATION:BOOL="True" -DMODPLUG:BOOL="True" -DOPUS:BOOL="True" -DOPTIMIZE:STRING="off" -DQT6:BOOL="True" -DQTKEYCHAIN:BOOL="True" -DSTATIC_DEPS:BOOL="False" -DVCPKG_TARGET_TRIPLET:STRING="x64-windows" -DVINYLCONTROL:BOOL="True" -DWAVPACK:BOOL="True" -DCMAKE_C_COMPILER:FILEPATH="C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.43.34808/bin/Hostx64/x64/cl.exe" -DCMAKE_CXX_COMPILER:FILEPATH="C:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.43.34808/bin/Hostx64/x64/cl.exe"   -DCMAKE_MAKE_PROGRAM="C:\PROGRAM FILES\MICROSOFT VISUAL STUDIO\2022\COMMUNITY\COMMON7\IDE\COMMONEXTENSIONS\MICROSOFT\CMAKE\Ninja\ninja.exe" "D:\mixxx-git\mixxx" 2>&1"' returned with exit code: 1'.
-> closed VS -> windows_buildenv.bat -> reopend vs
-> same errors
-> configured cache
-> errors:  (switchet to x64__portable)
CMake Error at D:\mixxx-git\mixxx\CMakeLists.txt:461 (include): ...
CMake Error at D:\mixxx-git\mixxx\CMakeLists.txt:462 (include): ...
CMake Error at D:\mixxx-git\mixxx\CMakeLists.txt:928 (default_option): ...
-> cannot build :-(

JoergAtGithub:
@Eve00000 Please delete the build directory and ensure that there is no CMakeSettings.json from previous runs of windows_buildenv.bat These errors occur long after the buildenv download and occur most likely because windows_buildenv.bat did not set other settings correct.

Eve00000:
I deleted .vs / build / build / buildenv before starting 2nd test

JoergAtGithub:
How did you started the build, and what messages does the log contain about the buildenv download at the beginning?

Eve00000:
-> chechkout main
-> git branch -> cmake_download_buildenv_zip
-> deleted subdirs .vs / build / buildenv
-> deleted tools\windows_buildenv.bat

-> winpty gh pr checkout 13306 -> completes download & checkout
-> open VS -> after some seconds CMalkeLists starts
-> erros as copied in previous post

JoergAtGithub:
Can you please send me the full CMake output.

Eve00000:
full output
[20250315-PR13306-CMAKE-Output.txt](https://github.com/user-attachments/files/19262326/20250315-PR13306-CMAKE-Output.txt)

JoergAtGithub:
@Eve00000 In your log everything looks like in mine, until the failure. Especially download and extraction of the buildenv is successful. Could you ensure manually, that the buildenv is is complete.

Eve00000:
OK, doing it all again (still without bat to see error message)
zip -> downloaded & unpacked
buildenv -> created
build -> created

now everything looks ok, but I did not execute the bat
-> no error for that
[20250315-PR13306-CMAKE-Output 2.txt](https://github.com/user-attachments/files/19262728/20250315-PR13306-CMAKE-Output.2.txt)

could it have been the proble I downloaded the pr before CI was ready?

JoergAtGithub:
If you did not executed `windows_buildenv.bat` and it configures without error, than the settings from the last run persist, most likely in `CMakeSettings.json`. Did you deleted it?

Eve00000:
yup yup, that's why I find it strange.
Build is ready, I'll redo everything once again to make sure

JoergAtGithub:
Than it's doing what it should in this case!

JoergAtGithub:
I'm already enjoying the sun😊 Hurry up to not miss it! Thanks for testing!

CMakeLists.txt
```diff
@@ -48,19 +48,134 @@ if(POLICY CMP0135)
...
+      message(
+        FATAL_ERROR
+        "BUILDENV_URL not specified!\nDid you run `${CMAKE_SOURCE_DIR}/tools/windows_buildenv.bat`?"
+      )
+    endif()
+    if(NOT DEFINED BUILDENV_SHA256)
+      if(DEFINED ENV{BUILDENV_SHA256})
+        set(BUILDENV_SHA256 "$ENV{BUILDENV_SHA256}")
+      else()
+        message(STATUS "BUILDENV_SHA256 not specified")
```
daschuer, line 115: What might have happen for this message? Can we stop here fatal?

CMakeLists.txt
```diff
@@ -48,19 +48,134 @@ if(POLICY CMP0135)
...
+      )
+      file(SHA256 "${BUILDENV_BASEPATH}/${BUILDENV_NAME}.zip" actual_sha256)
+
+      if(NOT actual_sha256 STREQUAL ${BUILDENV_SHA256})
+        message(
+          FATAL_ERROR
+          "SHA256 checksum mismatch:\nexpected: ${BUILDENV_SHA256}\n     got: ${actual_sha256}"
+        )
+      else()
+        message(STATUS "SHA256 ${BUILDENV_SHA256} is correct!")
```
daschuer, line 144: Did you consider to join the initial message with this one: 
      message(
        STATUS
        "SHA256 checksum verified for downloaded file ${BUILDENV_BASEPATH}/${BUILDENV_NAME}.zip"
      )

and 
      message(
          FATAL_ERROR
          "SHA256 checksum mismatch for downloaded file ${BUILDENV_BASEPATH}/${BUILDENV_NAME}.zip:\nexpected: ${BUILDENV_SHA256}\n     got: ${actual_sha256}\nDownload may be corrupted. Please try again."
        )

CMakeLists.txt
```diff
@@ -48,19 +48,134 @@ if(POLICY CMP0135)
...
+
 # Use this function to throw an error because the build environment is not set
 # up correctly.
 function(fatal_error_missing_env)
   if(WIN32)
     if(CMAKE_BUILD_TYPE MATCHES "Debug")
       message(
         FATAL_ERROR
-        "Did you download the Mixxx build environment using `${CMAKE_SOURCE_DIR}/tools/windows_buildenv.bat`?"
+        "Did you run `${CMAKE_SOURCE_DIR}/tools/windows_buildenv.bat`?"
```
daschuer, line 173: The information that a GB download is started by this is missing. That can be an issue if you are out of disk space or using a slow or expensive internet connection. 
Can we add this info to the windows_buildenv.bat 

Some think like this: 

```
echo Note: 
echo This will download the Mixxx build environment which is > 1 GB.
echo Make sure you have enough disk space and a fast connection with an unlimited plan.
echo.
choice /C YN /N /M "Press Y to continue or N to cancel: "
if errorlevel 2 (
    exit /b
)
```

daschuer (changes requested):
Looks already good. I have left some comments for improvement. 

One concern is what will happen if a user switches between branches with this PR merged and without. Will it run flawlessly or is the user always prompted with errors?

CMakeLists.txt
```diff
@@ -48,19 +48,134 @@ if(POLICY CMP0135)
...
+
+  message(STATUS "BUILDENV_NAME is: ${BUILDENV_NAME}")
+
+  if(
+    EXISTS "${BUILDENV_BASEPATH}/${BUILDENV_NAME}"
+    AND IS_DIRECTORY "${BUILDENV_BASEPATH}/${BUILDENV_NAME}"
+  )
+    message(
+      STATUS
+      "BUILDENV directory ${BUILDENV_BASEPATH}/${BUILDENV_NAME} exists!"
```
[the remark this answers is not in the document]
daschuer: Just realized even this is redundant. We have now: 
```
-- BUILDENV_NAME is: mixxx-deps-2.5-x64-windows-release-40c29ff
-- BUILDENV directory C:\buildenv/mixxx-deps-2.5-x64-windows-release-40c29ff exists!
-- Using MIXXX_VCPKG_ROOT: C:\buildenv\mixxx-deps-2.5-x64-windows-release-40c29ff
```
The last should be enough.

CMakeLists.txt
```diff
@@ -48,19 +48,134 @@ if(POLICY CMP0135)
...
+      endif()
+    endif()
+  endif()
+
+  message(STATUS "BUILDENV_NAME is: ${BUILDENV_NAME}")
+
+  if(
+    EXISTS "${BUILDENV_BASEPATH}/${BUILDENV_NAME}"
+    AND IS_DIRECTORY "${BUILDENV_BASEPATH}/${BUILDENV_NAME}"
+  )
```
daschuer, line 99: This implies EXISTS
```suggestion
  if(IS_DIRECTORY "${BUILDENV_BASEPATH}/${BUILDENV_NAME}")
```

CMakeLists.txt
JoergAtGithub: The batch file does not download anymore, it just writes a settings file and sets some environment variables.
The main goal of this PR is to get rid of application logic in this nasty batch file. The last thing we should do is to add new functionality to this batch file.

CMakeLists.txt
```diff
@@ -48,19 +48,134 @@ if(POLICY CMP0135)
   cmake_policy(SET CMP0135 NEW)
 endif()
 
+if(WIN32)
```
daschuer, line 51: I think we need here to check also for MIXXX_VCPKG_ROOT if that exists, we are likely in an already configured build directory and can skip the whole block. 
```suggestion
if(WIN32 AND IS_DIRECTORY MIXXX_VCPKG_ROOT)
```

JoergAtGithub: This would no longer detect if someone changed the buildenv to use by specifing -DBUILDENV_NAME The typical use case when you download a buildenv ZIP file from a VCPKG PR.

CMakeLists.txt
JoergAtGithub: I would like to leave it as it is, as the one message tells what check is started, and another with result after. I think this is easier to understand than all in one line, and the user is always aware, what happens - this is a check that need 2-3 seconds.

CMakeLists.txt
JoergAtGithub: This is intentional, similar to the checksums of VCPKG ports. You can run it with out checksum specified. It will of cause fail than later, but with the actual checksum printed, that you can copy paste than.

CMakeLists.txt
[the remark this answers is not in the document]
JoergAtGithub: As this comes from different settings, I could imagine failures where both information helps to understand a problem.
It should be the same, but the information is two seperate things:
1.) The directory really exist on the harddisk
2.) How MIXXX_VCPKG_ROOT is specified
These are all existing variables, that were not introduced by this PR.

JoergAtGithub:
> One concern is what will happen if a user switches between branches with this PR merged and without. Will it run flawlessly or is the user always prompted with errors?

If all settings inside `windows_buidenv.bat` and the buildenv remain the same, the user has not to rerun it. It will work as now. If the buildenv or settings differ the user has to rerun it - this is also the same behavior as now.

CMakeLists.txt
daschuer: Ah I see. However it looks like id fails now when BUILDENV_NAME  and BUILDENV_URL is not even though if MIXXX_VCPKG_ROOT is valid. I consider BUILDENV_NAME and BUILDENV_URL as internal, used for communication form the batch file. Does that make sense?

JoergAtGithub: You mean the case that somebody locally builds VCPKG and just set MIXXX_VCPKG_ROOT ? In all other cases BUILDENV_NAME  or BUILDENV_URL are the leading vars, where MIXXX_VCPKG_ROOT could be derived from. And if windows_buildenv.bat is used, all vars are in sync anyway.

CMakeLists.txt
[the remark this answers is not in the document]
daschuer: We are looking now at the "good" case where we don't have a problem. Maybe there is an issue with passing redundant information? What is the expected behavior if MIXXX_VCPKG_ROOT and  BUILDENV_NAME are not matching? MIXXX_VCPKG_ROOT is probably the cached variable indicating the old sate before a vcpkg update.  

Does BUILDENV_NAME also co to CMakeCache.txt? 

How about this approach: 
* plain cmake call (without any ENV vars set) :
  * use MIXXX_VCPKG_ROOT if in cache 
  * else fail and prompt to run the batch file 
* run cmake after running the batch: 
  * Update buid environment. 

What shall happen if one changes directory to a different working directory with a different environment? 
It will be probably also updated undesired. Can we prevent that, be resetting the env variables after using?

CMakeLists.txt
daschuer: It looks like the check is skipped below in this case, is it? Maybe it is cleaner to move this check closer to usage, to make it easier to trace the control flow. Do we need a bypass feature?
VCPKG has handled a missing checksum like this: 
```
  vcpkg_download_distfile requires a SHA512 argument.
  If you do not know the SHA512, add it as 'SHA512 0' and retry.
```
do we want to copy that?

CMakeLists.txt
daschuer: Ah OK that makes sense.

CMakeLists.txt
JoergAtGithub: In the VCPKG case you need to write something into the file (works with any value btw. not only 0). But here this would be an additional step without benefit. currently it prints the information you need to specify BUILDENV_SHA256 without such an extra step. I think better developer experience is not possible here.

CMakeLists.txt
daschuer: This happens now "delayed". I think a typical run is like that: 
* Use cmake without the environment set (no download)  
* Read this prompt 
* Run the batch  
* Run cmake again 
* Being surprised that a GB download is started.

Before the situation was not ideal, but sightly better because only the dedicated batch call did the download, which is documented in the wiki. 

I have stumbled here because you have removed the info that a download happens. It still happens after running the script, but during the next cmake call. This can happen more or less nice by the IDE. 

Do you have an idea  how to improve the situation?

JoergAtGithub: I will rephrase the messages, to provide the removed information in a way that matchs to the new flow.

CMakeLists.txt
daschuer: This is the new situation now after this PR. If one uses a branch from before only MIXXX_VCPKG_ROOT is used by cmake. "VCPKG_ROOT" is the official upstream way to point to the build environment. We have introduced MIXXX_VCPKG_ROOT for the same purpose because "VCPKG_ROOT" is already set to a different directory on the GitHub workflow runners. 

That's why I think it is more intuitive to keep "MIXX_VCPKG_ROOT" for the case you want to switch between two environments without issuing a download. I consider BUILDENV_NAME and BUILDENV_URL as programmatic script helper the user should not set manually.

daschuer:
Did you test it? For my understanding building an old build directory after applying this patch will fail because BUILDENV_NAME and BUILDENV_URL are not set.

CMakeLists.txt
JoergAtGithub: No it's the other way around, MIXXX_VCPKG_ROOT is just a a concatenation from BUILDENV_BASEPATH and BUILDENV_NAME inside windows_buildenv.bat. I will move this concatenation from windows_buildenv.bat to CMakeLists.txt than we have a straight forward data flow without duplicated env variables.
The basepath should be overwriteable for it's own, to allow a central storage location used by all worktrees.

CMakeLists.txt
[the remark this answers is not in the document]
JoergAtGithub: I moved the concatenation of BUILDENV_BASEPATH + BUILDENV_NAME to MIXXX_VCPKG_ROOT from windows_buildenv.bat to CMakeLists.txt. Now we have a straightforward data flow that eliminates the possibility of contradictory information by design.

Eve00000:
If you want me to do some more test ... let me know

JoergAtGithub:
> If you want me to do some more test ... let me know

Oh yes, now it's a good time to test! As all is addressed now!

CMakeLists.txt
JoergAtGithub: I changed the messages and now the information about the download is back.

JoergAtGithub:
> Did you test it? For my understanding building an old build directory after applying this patch will fail because BUILDENV_NAME and BUILDENV_URL are not set.

It tested now with switching from this PR to main. Deleted the CMake cache and configured without calling windows_buildenv.bat again. As the buildenv was still there and the toolchain file remains the same everything worked. I just got these warnings at the end:
```
1> [CMake] -- Configuring done (22.8s)
1> [CMake] -- Generating done (1.2s)
1> [CMake] CMake Warning:
1> [CMake]   Manually-specified variables were not used by the project:
1> [CMake] 
1> [CMake]     BUILDENV_BASEPATH
1> [CMake]     BUILDENV_NAME
1> [CMake]     BUILDENV_SHA256
1> [CMake]     BUILDENV_URL
1> [CMake] 
1> [CMake] 
1> [CMake] -- Build files have been written to: D:/mixxx-2.5-worktree/build/x64__legacy
```

Going the other direction it fails now with the following clear message:
```
1> [CMake] -- CMAKE_VERSION: 3.30.5-msvc23
1> [CMake] -- BUILDENV_URL not specified
1> [CMake] CMake Error at D:\mixxx-2.5-worktree\CMakeLists.txt:86 (message):
1> [CMake]   Neither BUILDENV_URL nor BUILDENV_NAME specified!
1> [CMake] 
1> [CMake]   Did you run `D:/mixxx-2.5-worktree/tools/windows_buildenv.bat`?
1> [CMake] 
1> [CMake] 
1> [CMake] -- Configuring incomplete, errors occurred!
```

Eve00000:
I did the test
-> removed all dep - linked items + .vs . json + ...
-> started VS -> cmake -> error 'tools\windows_buildenv.bat'
-> closed VS
-> executed windows_buildenv.bat
-> started VS
-> download + extract
CMake generation finished.
All good, nice! Congrats.

daschuer:
> Going the other direction it fails now with the following clear message.

But isn't this unnecessary? To make it work was my idea if using MIXXX_VCPKG_ROOT primary. 
What is the benefit of the current situation that forces all contributors to run the batch again?

daschuer:
I can confirm that disposing MIXXX_VCPKG_ROOT is a good idea. It has never been documented as a user variable. 

However in a context of an IDE this PR still introduces an issue. `cmakeToolchain` goes as `CMAKE_TOOLCHAIN_FILE` to the CMakeCache.txt this is a one time task when configuring the project  

> Unless they are experts like you and me, users should run it any time.

I have not tested this because I am not on Windows, but the for my understanding the batch file is not re-run if you use an IDE like QtCreator. In case of Visual studio, the settings ares stored in CMakeSettings.json. Other IDEAs have other ways to configure the CMakeCache.txt. 

In Eclipse (I use on Linux) for example I can imagine to configure the batch file one time, to configure the project form the command line and than use only Eclipse without touching the command line ever again. 

So I think instead of MIXXX_VCPKG_ROOT, CMAKE_TOOLCHAIN_FILE is the real source of truth. This is also widely known because it is a CMake standard variable. Can we check this to skip the prompt for rerunning the batch? 

I will propose the change inline.

CMakeLists.txt
```diff
@@ -48,19 +48,132 @@ if(POLICY CMP0135)
...
```
daschuer, line 51: This should avoid reconfiguration when the CMAKE_TOOLCHAIN_FILE still exists 
```suggestion
if(WIN32 AND NOT EXISTS "${CMAKE_TOOLCHAIN_FILE}")
```

daschuer, line 51:
```suggestion
if(WIN32 AND (NOT EXISTS "${CMAKE_TOOLCHAIN_FILE}" OR NOT "${CMAKE_TOOLCHAIN_FILE}" MATCHES ".*${BUILDENV_NAME}.*"))
```

daschuer:
Since MIXXX_VCPKG_ROOT is now internal, we need to adjust the error messages mentions that.

daschuer:
I have just noticed that there is still the logic of configuring  VCPKG_OVERLAY_PORTS and VCPKG_OVERLAY_TRIPLETS. 
are they part of the CMakeCache? If not we need to take care that they are always configured.

CMakeLists.txt
JoergAtGithub: `${CMAKE_TOOLCHAIN_FILE}` will never match `${BUILDENV_BASEPATH}/${BUILDENV_NAME}`, as the one is a file and the other a directory.
In my case `D:\mixxx\buildenv\mixxx-deps-2.5-x64-windows-c15790e\scripts\buildsystems\vcpkg.cmake` vs `D:\mixxx\buildenv\mixxx-deps-2.5-x64-windows-c15790e\`

What do you want to achieve with this condition?

JoergAtGithub:
> Since MIXXX_VCPKG_ROOT is now internal, we need to adjust the error messages mentions that.

I could, but this variable is still generated by `macos_buildenv.sh`, where it's also generated by concatenating BUILDENV_BASEPATH and BUILDENV_NAME.
Do you agree to modify this script in the same way as `windows_buildenv.bat` ? As requested above by @fwcd https://github.com/mixxxdj/mixxx/pull/13306#discussion_r1720464623

> I have just noticed that there is still the logic of configuring VCPKG_OVERLAY_PORTS and VCPKG_OVERLAY_TRIPLETS. are they part of the CMakeCache? If not we need to take care that they are always configured.

To my understanding these are not of any use with the prebuilt buildenv. These apply only if you link to a local working copy of mixxxdj/VCPKG where you build the dependencies locally.

daschuer:
Nikhil Bisht did some extra for us on windows and discovers some details / requirements: 

* Once the CMakeCache.txt is populated, the project can be build without running the batch. 
* It requires however the VS command prompt to find comctl32.lib 
* Significant cached variables are: 
  * CMAKE_TOOLCHAIN_FILE
  * Z_VCPKG_ROOT_DIR
  * VCPKG_INSTALLED_DIR
  * VCPKG_TARGET_TRIPLET
* All libraries path are cached absolute 
* All other script variables are temporary 

He struggles to reconfigure a configured build directory. 
After starting over with an empty one it works. I have not verified it, because I am in Linux but I can also remembered to suffer from these issues with the cmake cache when installing home build libraries. 
So I consider updating the vcpkg environment without cleaning the CMakeCache.txt as supported. Can you confirm? 

From this we can generate these requirements: 
* Download vcpkg if not found 
* Not change the vcpkg environment if already configured 
* Build without running windows_buildenv.bat if already configured
* Fail if running without windows_buildenv.bat and not configured 
* Warn if running with windows_buildenv.bat and a different vcpkg is configured. 
* Allow to configure any previous downloaded vcpkg environment without script. Currently done via MIXXX_VCPKG_ROOT

Does that make sense?

daschuer:
> Yes, this is because we set CMAKE_TOOLCHAIN_FILE as CACHE_STRING:

Ah, so we mirror (probably unnecessarily) the default bahviour of CMake without VCPKG. See:  
https://cmake.org/cmake/help/latest/envvar/CMAKE_TOOLCHAIN_FILE.html#envvar:CMAKE_TOOLCHAIN_FILE

At least "Z_VCPKG_ROOT_DIR" is reliably used and updated without our extra adjustments. 

> Yes, but CMake detects these cases and tells you very clear, when you have to delete the cache first. But it doesn't do this automatic.

I Have just tested to change the CMAKE_TOOLCHAIN_FILE. I receive this unclear warning: 
```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_TOOLCHAIN_FILE
```

When I try to use a diffrent generator I get such a waring: 
```
CMake Error: Error: generator : Unix Makefiles
Does not match the generator used previously: Ninja
Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory.
```
But this is not our case if we update to a new environment. What did you do to get a better warning? 

> But the most important thing is, that a user never gets the wrong vcpkg environment unless he explicitly overides the one set in windows_buildenv.bat . Otherwise switching between branches would no longer work.

I am not sure if I get you right. Please verify. Let's asume we have updated the main branch to our new 2.6 environment: 
* After updateing to recent master the windows_buildenv.bat pount to the new environment. 
* The CMakeCache is still old. 
* Normal Rebuilding will work and still use the 2.5 environment 
* Update to 2.5 requires to delete CMakeCache.txt, run the batch and reconfigure cmake. 
* Switching to the 2.5 branch will now use the 2.6 environment listed in the CMakeCache.txt
This would be OK for me, like the same situaten when updating the Linux distro. 

I use a 2.5 and a "main" branch to avaoid to much rebuilds. This is also a good receipt to use the right environment. 
Sometimes I accedentally merge a 2.6 PR into the 2.5 working copy. The penalty is a longer rebuild. 
In the windows situation it will likely fail because some required 2.6 libraries are missing. Here we should fail eraly with a descriptive message. 

For convenience we should always keep te 2.5 branch building with the 2.6 environment. (We need probably anyway because of Linux) 

Do you agree or do I miss something? 

>> * Warn if running with windows_buildenv.bat and a different vcpkg is configured.

> This warning makes no sense to me, because if the user runs windows_buildenv.bat, he explicitly set the new buildenv by this action. This is the main purpose of this batch file.

Ok, Then we should fail. Works for me as well.

daschuer:
It looks like with your latest changes we are already close. Thank you. I have not tested, but I think the only missng piece is the question around manual setting the environment. And outdated commenst around `MIXXX_VCPKG_ROOT`
Use case: 
* User has a working copy with the default vcpkg environent.
* Wants to create a second working copy without reusing the original environment. 

Thay may want to do this (or equivalent, I don't mind) 
```
git worktree add ..\experiment main 
cd ..\experiment
mkdir build 
cd build 
cmake .. -G "Ninja" -DMIXXX_VCPKG_ROOT=..\buildenv\mixxx-deps-2.5-x64-windows-c15790e
```
Officially it is: 
```
cmake ../my/project -DCMAKE_TOOLCHAIN_FILE=<vcpkg-root>/scripts/buildsystems/vcpkg.cmake
```
https://learn.microsoft.com/en-us/vcpkg/users/buildsystems/cmake-integration#cmake_toolchain_file

For my underanding both should skip a download if configured as well in the enviroment variables.

daschuer:
The CMAKE_TOOLCHAIN_FILE ends as 
CMAKE_TOOLCHAIN_FILE:FILEPATH=/home/daniel/workspace/vcpkg/scripts/buildsystems/vcpkg.cmake
even if it is specified relative to the source directory

Internally a similar command like this is used: 
```
file(REAL_PATH "${CMAKE_TOOLCHAIN_FILE}" CLEAN_CMAKE_TOOLCHAIN_FILE BASE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}" EXPAND_TILDE)
```
Since we cannot reconfigure cmake safely, we need to fail if the BUILD_ENV variables are set but don't match. 
I will propose changes inline.

daschuer:
If the project is configured and the BUILD_ENV, we can't safely reconfigure. That's why the only chance we have that the buildenv variables can win is to ask for deleting the cache.

JoergAtGithub:
Yes, that is the behavior we currently have and works very well for me. I will not change this in this PR!
Changenging the general behavior is out of scope of this PR!

CMakeLists.txt
```diff
@@ -48,19 +48,153 @@ if(POLICY CMP0135)
...
+  string(REPLACE "\\" "/" BUILDENV_BASEPATH "${BUILDENV_BASEPATH}")
+  string(REPLACE "\\" "/" CMAKE_TOOLCHAIN_FILE "${CMAKE_TOOLCHAIN_FILE}")
+endif()
```
daschuer, line 54: This mirrors the CMAKE_TOOLCHAIN_FILE path handling, but uses a new variable to not change any cached value. 
```suggestion
file(
  REAL_PATH
  "${CMAKE_TOOLCHAIN_FILE}"
  CLEAN_CMAKE_TOOLCHAIN_FILE
  BASE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
  EXPAND_TILDE
)
if(DEFINED BUILDENV_NAME)
  file(TO_CMAKE_PATH ${BUILDENV_BASEPATH} BUILDENV_BASEPATH)
  set(MIXXX_VCPKG_ROOT "${BUILDENV_BASEPATH}/${BUILDENV_NAME}")
endif()
```

CMakeLists.txt
```diff
@@ -48,19 +48,153 @@ if(POLICY CMP0135)
...
+          DEFINED BUILDENV_BASEPATH
+          AND DEFINED BUILDENV_NAME
+          AND
+            NOT
+              CMAKE_TOOLCHAIN_FILE
+                MATCHES
+                "${BUILDENV_BASEPATH}/${BUILDENV_NAME}/scripts/buildsystems/vcpkg.cmake"
+        )
+    )
+)
```
daschuer, line 72: if(
  DEFINED CMAKE_TOOLCHAIN_FILE
  AND DEFINED MIXXX_VCPKG_ROOT
  AND
    NOT
      CLEAN_CMAKE_TOOLCHAIN_FILE
        MATCHES
        "${MIXXX_VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
)
  message(
    FATAL_ERROR
    "Build environment: '${MIXXX_VCPKG_ROOT}' does not match the environment used previously: '${CMAKE_TOOLCHAIN_FILE}'. Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory."
  )
endif()

if(WIN32 AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)

daschuer (changes requested):
This works good on Linux as far as I can tell. I am pretty sure that clanging the environment without clearing the Cache does not work. But I can't prove it yet.

CMakeLists.txt
daschuer, line 72:
```suggestion
if(
  DEFINED CMAKE_TOOLCHAIN_FILE
  AND DEFINED MIXXX_VCPKG_ROOT
  AND
    NOT
      CLEAN_CMAKE_TOOLCHAIN_FILE
        MATCHES
        "${MIXXX_VCPKG_ROOT}/scripts/buildsystems/vcpkg.cmake"
)
  message(
    FATAL_ERROR
    "Build environment: '${MIXXX_VCPKG_ROOT}' does not match the environment used previously: '${CMAKE_TOOLCHAIN_FILE}'. Either remove the CMakeCache.txt file and CMakeFiles directory or choose a different binary directory."
  )
endif()

if(WIN32 AND NOT DEFINED CMAKE_TOOLCHAIN_FILE)
  if(NOT DEFINED BUILDENV_BASEPATH)
```

daschuer:
Here is a CI run that demonstrates the issue 
https://github.com/daschuer/mixxx/actions/runs/13958279397/job/39074483674
It configures the build directory as usual. Than It move the vcpkg package directory and re-configures the project with the new location and fails. 

It will work, if the original environment still exists and you download the new in addition, However this leads to a dubious situation where some parts of the old and some parts of the new environment are used. I consider this as maximal confusing and therefore the early error message with a clear instruction is helpful.

CMakeLists.txt
JoergAtGithub: The suggested condition did not work, as it disabled the buildenv download in the normal workflow (because windows_buildenv.bat sets CMAKE_TOOLCHAIN_FILE). I adjusted the condition for the error message accordingly.

JoergAtGithub:
The requested fata error message is added, this is ready now!

JoergAtGithub:
@daschuer A friendling ping, that this is ready to merge!

CMakeLists.txt
daschuer: It turns out my suggestion was wrong. It should have been: 

```suggestion
if(WIN32 AND NOT IS_DIRECTORY MIXXX_VCPKG_ROOT)
```

The original use case for MIXXX_VCPKG_ROOT is to be a 1:1 VCPKG_ROOT replacement which shall be obvious form the name. This shall be the common source of truth for all targets and shall be usable to override the targets maintained by us. 

See: https://github.com/daschuer/mixxx/blob/eb38c67c0b17d6b0677f9f9050b4f46c94b83184/CMakeLists.txt#L104

And https://github.com/mixxxdj/mixxx/issues/15016

Only if the directory dos not exists, the download and unpacking shall start. We need also make sure that it only happens with a fresh build directory. Changing the build environment with an already configured other environment fails by luck or produces an unwanted mixed result. We need to prevent that early. Probably by caching the MIXXX_VCPKG_ROOT after the download is and unpacking is done. 

We can't use VCPKG_ROOT itself, because the system VCPKG_ROOT might point to a different tree and we don't wan to mess with that. https://learn.microsoft.com/en-us/vcpkg/users/config-environment#vcpkg_root

My suggestion is IMHO required to maintain this override nature and skip our own scripting. 

What is your use case that is broken by that?

daschuer (changes requested):

CMakeLists.txt
daschuer: This comment is also pending

daschuer (changes requested):

CMakeLists.txt
```diff
@@ -48,6 +48,85 @@ if(POLICY CMP0135)
...
+if(WIN32 AND NOT IS_DIRECTORY MIXXX_VCPKG_ROOT)
+  if(NOT DEFINED BUILDENV_BASEPATH)
+    if(DEFINED ENV{BUILDENV_BASEPATH})
+      set(BUILDENV_BASEPATH "$ENV{BUILDENV_BASEPATH}\\")
+    else()
+      set(BUILDENV_BASEPATH "${CMAKE_SOURCE_DIR}/buildenv/")
+    endif()
+  endif()
+
+  if(DEFINED ENV{BUILDENV_URL})
```
daschuer: The same here. ENV shall be used if not manually passed.

CMakeLists.txt
```diff
@@ -48,6 +48,85 @@ if(POLICY CMP0135)
...
+
+  if(
+    NOT EXISTS "${BUILDENV_BASEPATH}${BUILDENV_NAME}"
+    OR NOT IS_DIRECTORY "${BUILDENV_BASEPATH}${BUILDENV_NAME}"
+  )
+    if(DEFINED ENV{BUILDENV_SHA256})
+      set(BUILDENV_SHA256 "$ENV{BUILDENV_SHA256}")
+    elseif(NOT DEFINED BUILDENV_SHA256)
+      message(STATUS "BUILDENV_SHA256 not specified")
+    endif()
```
daschuer, line 92: and here ...

daschuer (changes requested):

daschuer (approved):
LGTM, Thank you.

PR #13311, opened 2024-06-01
Improved readability and scaling of digits and avoid some artefacts

m0dB:
Based on user feedback, improved the readability of the digits and fixed the spacing and the scaling, used for the beats/time until mark counter.

BEFORE:
Screenshot 2024-06-02 at 02 41 52

AFTER:
Screenshot 2024-06-02 at 02 43 06

### DIFF
--- a/src/waveform/renderers/allshader/digitsrenderer.cpp
+++ b/src/waveform/renderers/allshader/digitsrenderer.cpp
@@ -21,9 +21,9 @@ namespace {
 constexpr int NUM_CHARS = 12;
 
 // space around chars for blurred dark outline
-constexpr qreal SPACE = 3;
-// extra kerning when rendering chars
-constexpr qreal KERNING = 1;
+constexpr int OUTLINE_SIZE = 4;
+// alpha of the blurred dark outline
+constexpr int OUTLINE_ALPHA = 224;
 
 constexpr char indexToChar(int index) {
     constexpr char str[] = "0123456789:.";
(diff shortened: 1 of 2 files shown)


daschuer (approved):
This PR looks good. 

Now that I have tested the center mode a bit more, I noticed that the artificial font size limitations kicks in a bit early. 
This as the smallest waveform I consider as reasonable, but the additional beat cont is only barely readable, the font size is ignored. 
grafik
I will file an issue.

m0dB:
That is strange. Before the font scaling kicks in, singleline instead of multiline should kick in first. That that doesn’t happen is the bug. It certainly does in macOS. What resolution is this? And hdpi?

m0dB:
I can't reproduce this, @daschuer . Are you sure you are running this branch?

This auto-scaling is uses
```
float allshader::WaveformRenderMark::getMaxHeightForText() const {
    return m_waveformRenderer->getBreadth() / 3;
} 
```
and if with the selected font size the text would be higher than this max height, it scaled the font size down so it fits. It will only use multiline if the twice the text height still fits in this max height.

So, if you only show beats or time, you should still get the too small font size, on a single line. Is that the case?

What happens if you return a very large number in ```getMaxHeightForText()```?

I am wondering if maybe I am making a mistake with the device pixel ratio. Can you tell me which device pixel ratio you have?

daschuer:
My ratio is just one.

m0dB:
And what happens in the same situation with only beat count?

m0dB:
Never mind, I have been able to repro the issue and wil add a solution to this PR.

m0dB:
The minimum font point size is now limited to 10 pt. 

I also added a 0.5 pixel margin around the text to avoid artefacts that I didn't see on HDPI, but noticed at 1:1 pixel ratio.

For reference: it is possible to run at 1:1 pixel ratio on macOS by building the bundle (cmake -DMACOS_BUNDLE=ON) and editing the Info.plist in the bundle with:
```
    <key>NSHighResolutionCapable</key>
    <string>False</string>
```

Will you do the honours and preset the merge button, @daschuer ? :-)

daschuer:
Oh sorry forgot to merge. Now the CI is failing.

m0dB:
let's see now.

there is a minor issue: depending on the font size, the blurred black outline is sometimes slightly displaced from the white foreground. not a biggie though, but now that I have seen it, I can't unseen it!

daschuer:
I can confirm the minimum font size fixes the issue https://github.com/mixxxdj/mixxx/issues/13314
Are you working on a fix for last minor issue on to of this or shall I press merge?

m0dB:
Let’s wait until tomorrow. I will make a final attempt.

m0dB:
What I tried didn't work, so I'd say merge. If I find a solution I will create a new PR.

daschuer:
Ok, thanks.

Issue #13238, opened 2024-05-16
EQ shelf controls leaked when no sound IO or broadcast is configured


ronso0:
### Bug Description

```
warning [Main] The following 2 controls were leaked:
warning [Main] "[Mixer Profile]" "HiEQFrequency" QObject(0x0)
warning [Main] "[Mixer Profile]" "LoEQFrequency" QObject(0x0)
```

I noticed this repeatedly, now I managed to reproduce it:
* get warning that previous sound device for Main is n/a
* click Reconfigure
* (no selection for Main), click Okay (or Apply + Okay)
* close Mixxx

Otoh, if I select a valid device, Apply/Okay, then clear the Main output, controls are not leaked :shrug: 

FWIW making the controls `std::unique_ptr<ControlPotmeter>` has no effect.
https://github.com/mixxxdj/mixxx/blob/8f647908af460e53fe8a860b6ce0964a93a55112/src/effects/effectsmanager.h#L115-L116

### Version

at least 2.4 and later

ronso0:
It seems the leaked controls are a side effect.
After this
> * click Reconfigure
> * (no selection for Main), click Okay (or Apply + Okay)

we should actually see a the next dialog "Mixxx was configured without any output sound devices."

### Conclusion:
**Apply** doesn't apply what we see Sound preferences, or rather the internal state is not displayed in the GUI, like missing but still 'configured' devices are not displayed in the comboboxes.

Might be related to #6077

ronso0:
(still no clue how that relates to the EQ controls)

ronso0:
> It seems the leaked controls are a side effect. After this
> 
> > * click Reconfigure
> > * (no selection for Main), click Okay (or Apply + Okay)

FYI this also seems to be the reason why sometimes tracks don't load (overview progress stuck).

daschuer:
#13312 and #13333 are merged. Can we consider this one as fixed?

ronso0:
#13333 is unrelated, and #13312 only avoids the problematic situation.

I prefer to leave this open so someone may take a look and maybe finds the root cause.
I suspect it's some missing nullptr check or something related to the main output and/or the output effect chain in `EngineMixer`.

PR #13312, opened 2024-06-01
(fix) Sound config: fix UX when applying config with missing/busy devices

ronso0:
Fixes one cause of #13238 leaking EQ shelf controls / stuck overview on track load.

### The scenario:
* start Mixxx without a previously configured sound I/O
* get warning that previous sound device for Main is n/a
   > Mixxx was unable to open all the configured sound devices.
* click Reconfigure
* (no selection for Main), click **Okay** (or Apply + Okay)
* Mixxx proceeds (should actually throw the "No output devices" warning)

### The issue:
There is a 'modified' flag in DlgPrefSound set on user changes, if false this will prevent re-applying the current config and the resulting sound pause when querying and re-setting the devices.
However, when the sound config is loaded in DlgPrefSound and devices are not found, the respective combobox will display 'None' even though that device is still set in the config. I.e. the visual state diverges from the config.
Clicking Apply would then **not** apply this new (visual) state since the 'modified' flag has not been set.

### The fix:
Set the 'modified' flag if configured devices are not found.
Apply will then apply as desired.
(and, in case of missing Main output, Mixxx will throw the "No output devices" warning)

Leaking EQ shelf controls and stuck overview on track load is another story...

### DIFF
--- a/src/preferences/dialog/dlgpreferences.cpp
+++ b/src/preferences/dialog/dlgpreferences.cpp
@@ -428,6 +428,9 @@ void DlgPreferences::slotButtonPressed(QAbstractButton* pButton) {
     case QDialogButtonBox::AcceptRole:
         // Same as Apply but close the dialog
         emit applyPreferences();
+        // TODO Unfortunately this will accept() even if DlgPrefSound threw a warning
+        // due to inaccessible device(s) or inapplicable samplerate.
+        // https://github.com/mixxxdj/mixxx/issues/6077
         accept();
         break;
     case QDialogButtonBox::RejectRole:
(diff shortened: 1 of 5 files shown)


ronso0:
Assigning this to 2.5-beta even though it's a 2.4 fix so we have it on the radar for the release.

daschuer:
We have the 2.4.2 milestone and will do the last 2.4 release short before the 2.5.0 release. So let's use the 2.4.2 milestone for that.

daschuer (approved):
Thank you for the fix and the good documentation.

PR #13319, opened 2024-06-02
Fix Sony SixxAxis controller not working

marios1861:
Minor fix in mapping file of sony sixxaxis controller. A wrong `HIDPacket` initialization was setting the length to 0, thus packets couldn't be parsed.

### DIFF
--- a/res/controllers/Sony-SixxAxis.js
+++ b/res/controllers/Sony-SixxAxis.js
@@ -10,7 +10,8 @@ function SonySixxAxisController() {
     this.controller.activeDeck = 1;
 
     this.registerInputPackets = function() {
-        packet = new HIDPacket("control",[],49);
+        const packet = new HIDPacket("control", 0);
+        packet.length = 49;
 
         // Toggle buttons
         packet.addControl("hid","select",2,"B",0x1);


JoergAtGithub (approved):
LGTM!

Swiftb0y (commented):
> My thinking was to first get the original mapping back to functional with minimal changes (achieved through this PR) and create an additional mapping without the bugs of this mapping in a separate PR. If you believe these two steps should be combined into a single PR and this dysfunctional mapping should be wholly overwritten, I will keep this PR as a draft and add the revised mapping.

Please keep them separate. In order to not break our CI pipeline, we need ensure that all the checks pass here (which in this case amounts to applying my suggestion below). Please do that and rebase the commits on the `2.4` branch so people receive the updated mapping faster. Thank you.

Swiftb0y (approved):
LGTM. Thanks for improving this! Please sign the [Mixxx Contributor Agreement](https://docs.google.com/forms/d/e/1FAIpQLScC9QG327sjLL0eWftmfGUasxFFLxg6LCXJ2xHDYRzFIRqyiw/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ) and comment here when you have done so. It gives us permission to distribute your contribution under the GPL v2 or later license and the Apple Mac App Store. It is also helpful for us to have contact information for contributors in case we may need it in the future.

marios1861:
@Swiftb0y all done!

PR #13320, opened 2024-06-02
LateNight: deactivate beatgrid edit controls if BPM is locked

ronso0:
Beatgrid edit controls in the waveform container and the CurPos trigger in the deck settings next to the overview.
Also adds the tooltip for `bpmlock`.
image

image

PR #13323, opened 2024-06-03
Deere: deactivate beatgrid edit controls if BPM is locked + LateNight fix

ronso0:
Same as #13320 for Deere, + a fixup for that (hotcue shift buttons were screwed due to wrong variable name)


ronso0:
Same for Tango will come soon.

daschuer (approved):
LGTM, thank you.

PR #13326, opened 2024-06-04
(fix) Effects: show newly added effects, read/write HiddenEffects

### LINKED ISSUE #11343: New effects are hidden after upgrade 2.4
### Bug Description

After a new installation all build in effects are visible. This is not the case if a use upgrades from a previous version of Mixxx and the new version brings new effects. In this case these new effects are hidden and there is a good chance that they are never discovered/used by the user. 

### Version

2.4

### OS

all

ronso0:
Fixes #11343

Add `HiddenEffects` to effects.xml

Small one-time issue after upgrade:
(new) effects that users have set visible and then set hidden again will now be set visible again :man_shrugging:

### DIFF
--- a/src/effects/effectsmanager.cpp
+++ b/src/effects/effectsmanager.cpp
@@ -328,7 +328,8 @@ void EffectsManager::saveEffectsXml() {
                     quickEffectChainPresets,
                     standardEffectChainPresets,
                     outputChainPreset});
-    m_pVisibleEffectsList->saveEffectsXml(&doc);
+
+    m_pVisibleEffectsList->saveEffectsXml(&doc, m_pBackendManager);
 
     QDir settingsPath(m_pConfig->getSettingsPath());
     if (!settingsPath.exists()) {
--- a/src/effects/visibleeffectslist.cpp
+++ b/src/effects/visibleeffectslist.cpp
@@ -48,9 +48,37 @@ const EffectManifestPointer VisibleEffectsList::previous(
 }
 
 void VisibleEffectsList::readEffectsXml(
-        const QDomDocument& doc, EffectsBackendManagerPointer pBackendManager) {
+        const QDomDocument& doc,
+        EffectsBackendManagerPointer pBackendManager) {
+    QList<EffectManifestPointer> visibleEffects = readEffectsList(
+            doc,
+            pBackendManager,
+            EffectXml::kVisibleEffects);
+    const QList<EffectManifestPointer> hiddenEffects = readEffectsList(
+            doc,
+            pBackendManager,
+            EffectXml::kHiddenEffects);
+
+    // New effects will remain hidden since they are neither in the VisibleEffects
+    // nor in the newly introduced HiddenEffects list.
+    // Unhide all effects that are not in either list.
+    const auto manifests = pBackendManager->getManifestsForBackend(EffectBackendType::BuiltIn);
+    for (const EffectManifestPointer& pManifest : std::as_const(manifests)) {
+        if (!visibleEffects.contains(pManifest) &&
+                !hiddenEffects.contains(pManifest)) {
+            // pre-pend so un-hidden effects are discoverable
+            visibleEffects.prepend(pManifest);
+        }
+    }
+    setList(visibleEffects);
+}
+
+QList<EffectManifestPointer> VisibleEffectsList::readEffectsList(
+        const QDomDocument& doc,
+        EffectsBackendManagerPointer pBackendManager,
+        const QString& xmlElementName) {
     QDomElement root = doc.documentElement();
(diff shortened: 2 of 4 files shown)


ronso0:
About the inline TODO:
Currently manifests are split into visible/hidden by DlgPrefEffects only ([here](https://github.com/mixxxdj/mixxx/blob/ac37870c428cc50b0586293844ccbc2b2e1516d1/src/preferences/dialog/dlgprefeffects.cpp#L110-L114)). And this PR adds similiar actions to `EffectsManager::saveEffectsXml()`.
We may keep it that way, or refactor/split VisibleEffectsList in order to have member pointers to both lists in `EffectsManager` and simply call `getList()` when saving to xml.
Wdyt?

Swiftb0y:
> refactor/split VisibleEffectsList in order to have member pointers to both lists in EffectsManager and simply call getList() when saving to xml.


What do you think about refactoring the `VisibleEffectsList`, so its just `EffectsList` which then manages the hidden vs visible effects internally (enforcing the constraint that the visible and hidden effectslists are disjunct). That would allow keeping EffectsManager simpler. Alternatively, having `VisibleEffectsList` and `HiddenEffectsList` would also be an option, though I'm skeptical that that would introduce a lot of boilerplate.

ronso0:
Yeah that could work but I can't promise to work on that soonish.

> Alternatively, having `VisibleEffectsList` and `HiddenEffectsList` would also be an option, though I'm skeptical that that would introduce a lot of boilerplate.

 Well, the difference is only the xml node name for read/write and one else branch when reading. So both could be derived from a common EffectsList.
Though, we actually need HiddenEffectsList only for read/write, no unit is accessing that list otherwise (via at(), next() etc.)

Swiftb0y:
Yeah, idk. I can't decide on the best solution either.

ronso0:
After all, this is a _TO DO_ ; p
I'll add some more..

As long as the current solution works we should merge it to make the new effects visible.
I'll definitely take another look during beta.

src/effects/effectsmanager.cpp
```diff
@@ -244,6 +244,23 @@ void EffectsManager::readEffectsXml() {
     }
 
     m_pVisibleEffectsList->readEffectsXml(doc, m_pBackendManager);
+    // New effects will remain hidden since they are neither in the VisibleEffects
+    // nor in the newly introduced HiddenEffects list.
+    // Unhide all effects that are not in either list.
+    auto visibleEffects = m_pVisibleEffectsList->getList();
+    VisibleEffectsList pHiddenEffectsList;
```
daschuer, line 251:
```suggestion
    VisibleEffectsList hiddenEffectsList;
```

src/effects/effectsmanager.cpp
```diff
@@ -328,8 +345,24 @@ void EffectsManager::saveEffectsXml() {
...
+    // Build the Hidden list from all effects that are not in the visible list.
+    // TODO Currently manifests are split into visible/hidden by DlgPrefEffects
+    // only. Keep it that way, or refactor/split VisibleEffectsList in order to
+    // have member pointers to both lists here and simply getList() when saving
+    // to xml?
+    auto hiddenEffects = m_pBackendManager->getManifests();
+    const auto visibleEffects = m_pVisibleEffectsList->getList();
+    for (const auto& pManifest : std::as_const(visibleEffects)) {
+        hiddenEffects.removeAll(pManifest);
+    }
```
daschuer, line 360: This can be also, integrated into VisibleEffectsList

src/effects/visibleeffectslist.cpp
```diff
@@ -68,16 +72,19 @@ void VisibleEffectsList::readEffectsXml(
         }
     }
 
-    if (!list.isEmpty()) {
-        setList(list);
-    } else {
+    if (list.isEmpty() && readVisible) {
+        // Show all built-in effects by default.
         setList(pBackendManager->getManifestsForBackend(EffectBackendType::BuiltIn));
```
daschuer, line 77: Is this condition still relevant? We will add  new effects anyway in the outside code.

daschuer (changes requested):
The solution looks good. I have added some suggestions for readability.

ronso0:
Thanks for your review @daschuer.
Your comments pretty much point to the refactoring we discussed above.

I'll see what I can do with minimal effort.

daschuer (approved):
LGTM, please sqash the fixup and merge. Thank you.

daschuer:
Since this is a bug fix it can go to the 2.5 branch at any time. I move that to the 2.5.0 milestone and proceed with the 2.5-beta release.

ronso0:
Well, it's ready. Merge whenever it fits.

JoergAtGithub:
Since 2.5 beta is out now, lets merge this into 2.5.

PR #13329, opened 2024-06-05
Preferences Effects: add Hide/Unhide (move) buttons to Effects tab

ronso0:
image

Fixes #10580 

#### ToDo
- [ ] static button positions, currently the move depending on the length of the effect description 

Up/Down buttons would also be nice.

### DIFF
--- a/src/preferences/dialog/dlgprefeffects.cpp
+++ b/src/preferences/dialog/dlgprefeffects.cpp
@@ -1,6 +1,7 @@
 #include "preferences/dialog/dlgprefeffects.h"
 
 #include <QFocusEvent>
+#include <QMimeData>
 
 #include "effects/backends/effectmanifest.h"
 #include "effects/backends/effectsbackend.h"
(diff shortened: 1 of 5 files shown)


ronso0:
We need to check whether the characters used for the buttons are in every font, or at least there are adequate fallbacks.
Unfortunately I can't really check that myself. I don't have `qt6ct` installed, and whatever exotic font I set in my OS font manager, Qt falls back to some sans serif (which either contains ❮and ❯ or gets them from some other fallback).

daschuer (approved):
LGTM and works good. However it does not fix the keyboard navigation bug.

ronso0:
> However it does not fix the keyboard navigation bug.

Haha, true, while this was the reason I started this.
Should be easy with eventFilter().

daschuer:
> Should be easy with eventFilter().

Mmm ... There is also an issue with the Tab navigation. 
Once the outer window is focused you can go back.

ronso0:
> Mmm ... There is also an issue with the Tab navigation.
Once the outer window is focused you can go back.

Yes, this seems to be a Qt6 regression and affects all pref pages.

Issue #11571, opened 2023-05-19
Re-add Crossfader re-center when disabling auto DJ


tlibra10:
### Feature Description

I would like to request for the crossfader auto re-set to be re-instated to the next update of Mixxx.

Thank You.

daschuer:
This feature has been removed to fix bug: 
https://github.com/mixxxdj/mixxx/issues/10683
here:
https://github.com/mixxxdj/mixxx/pull/4714

Please explain why you want to have  that feature back. Maybe you have also an idea how it could be brought back without re-introducing the fixed bug.

tlibra10:
I have been away from the decks a few times. Sometimes at home sometimes live. And on both occasions I have forgot to re-center the crossfader after disabling the auto dj. Started the next track and "Nothing" then realized that I did not re-center the crossfader. Therefore I decided to go back to version 2.3.2 as that one re-centers automatically.
I feel that this is a really useful feature. 
It also happened at a party and I felt so embarrassed. So for now I will stay with 2.3.2 as that is quite a good version anyway.  
Thank you.

mxmilkiib:
Proposal; the fader to move over the course of one second from the AutoDJ position to the centre on disabling the AutoDJ, a slewed transition, so the audio won't pop. PRs welcome. Could that be the situation?

ronso0:
IMHO no, volume will still drop even though there is no actual track-to-track transition, just slower than the instant reset.

I think the best idea that came up was the "Re-center crossfader when disabling AutoDJ" option in the preferences (default OFF). If users don't like the default I assume they'd look in the AutoDJ preferences first.

mxmilkiib:
Ok.

Sorry, I actually misread what the problem was, on one of the other dupes.

Yeah, that option you suggest is the easiest fix.

I mean, I guess ultimately the problem should be stated as: "AutoDJ doesn't know how to deal with 'the end of the mix' phase of a mix"?

To make it consistent, the exact time the user turns off AutoDJ should not be the exact time AutoDJ does it's last action; it should play the current track it into end, then finish.

But I've seen it said that adding more automation is extraneous magic. But finishing a mix is an integral part of a mix. 

So the underlying tension here is that there's more than one reason a user wants to use AutoDJ: Some want it to create nice mixes, say for a hospital radio, etc. and others just want to be able to immediately stop it for reference reason.

[...]

PR #13332, opened 2024-06-06
AutoDJProcessor: Add option to reset the crossfader to neutral

cr7pt0gr4ph7:
**Update:** Closed, see #13303 for a different solution to the same underlying problem (#11571).

# What does it do?
Add a simple configuration option for the Auto DJ that always resets the crossfader to neutral (`0.0`) while only one track is playing. This improves the use case of switching back and force between manually loading tracks and using the Auto DJ.

## Is it tested?
I've been actively using a build that includes these changes for a few weeks now, and so far everything seems to work exactly as expected.

## Are there any remaining open questions?
- Suggestions for improving the label and tooltip texts for the settings are more than welcome.
- The same goes for the name of the config option (`[Auto DJ] ResetFaderToNeutralOnIdle`)

The setting's default value is currently `false`, so the non-reset behavior is the default.
- For new users that just use Mixxx as a player with tempo control, a default value of `true` would be best to avoid the "surprise" that I had.
- For advanced users, the only use case that would be affected by a default value of `true` is:
  - (1) Disable the Auto DJ while only one track is playing.
    or (2) Auto DJ is enabled and only one track is playing.
  - Crossfader is not touched before they...
  - ...manually start playing a second deck, and expect it to not be hearable until the fader is moved manually.
  
  I don't know how common this use case is in practice - maybe one of you has more insight into this?

# What Problem Does This Solve?
> Use a laptop with a small screen, and no external DJ controller. Play two tracks using Auto DJ, then turn it off. Load another track manually, press play, and.... wonder if you have broken Mixxx because now, only one of two decks works. The UI looks the same, so you have no idea why, and just assume it's a bug.[^1]

[^1]: This story happened to me a few times (a part of those live...) before I had figured it out:
    - I had (inadvertenly, due to screen size) configured the skin to not show the crossfader.
    - I expected that turning the Auto DJ on and then later off again would not have any lingering side effects.
    - I just happened to lose on a 50/50 coin toss on which deck was active last.

So, my expectations as a user are:
- **For Novice Users:** I can use Auto DJ to play a few tracks in full, turn it off, and play a few tracks manually without having to care about anything else.

  Which implies: I expect both decks to be playable from (aka. the crossfader to be neutral) when there's no reason (at this moment) for it not to be neutral, and no other visible indicator telling me otherwise.

- **For Advanced Users:** Turning the Auto DJ off while two decks are playing and/or I want to manually control the crossfader does not cause the crossfader to jump.

Three solutions come to my mind, the last of which was implemented here:
1. Make sure the crossfader is always visible (but a user might still overlook it)
2. Reset the crossfader when the Auto DJ is being turned off (but avoid doing so when the user has actively changed the crossfader's value, which could get complicated).
3. Reset the crossfader as soon as the Auto DJ is not actively using it to perform a crossfade.

### DIFF
--- a/src/library/autodj/autodjprocessor.cpp
+++ b/src/library/autodj/autodjprocessor.cpp
@@ -14,11 +14,16 @@
 namespace {
 const char* kTransitionPreferenceName = "Transition";
 const char* kTransitionModePreferenceName = "TransitionMode";
+const char* kResetFaderPreferenceName = "ResetFaderToNeutralOnIdle";
 constexpr double kTransitionPreferenceDefault = 10.0;
 constexpr double kKeepPosition = -1.0;
 
 // A track needs to be longer than two callbacks to not stop AutoDJ
 constexpr double kMinimumTrackDurationSec = 0.2;
+
+constexpr double kCrossfaderLeftOnly = -1.0;
+constexpr double kCrossfaderNeutral = 0.0;
+constexpr double kCrossfaderRightOnly = 1.0;
 
 constexpr bool sDebug = false;
 } // anonymous namespace
@@ -207,6 +212,26 @@ void AutoDJProcessor::setCrossfader(double value) {
         value *= -1.0;
     }
     m_pCOCrossfader->set(value);
+}
+
+void AutoDJProcessor::setCrossfaderToIdle(double value) {
+    DEBUG_ASSERT(value == kCrossfaderLeftOnly || value == kCrossfaderRightOnly);
+
+    // Depending on the user's preferences, the idle position
+    // of the crossfader is either fully to the left/right,
+    // or in the middle.
+    const bool resetFaderToNeutralOnIdle = m_pConfig->getValue<bool>(
+            ConfigKey(kConfigKey, kResetFaderPreferenceName),
+            false);
+
+    if (resetFaderToNeutralOnIdle) {
+        // Move crossfader to neutral. Crossfader will be moved
+        // to the left/right just before starting a crossfade.
+        setCrossfader(kCrossfaderNeutral);
+    } else {
+        // Move crossfader fully to the left/right
+        setCrossfader(value);
+    }
 }
 
 AutoDJProcessor::AutoDJError AutoDJProcessor::shufflePlaylist(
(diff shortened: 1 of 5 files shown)


cr7pt0gr4ph7:
> Did you notice #13303? FWIW that PR also cleans up the AutoDJ preferences page, so there are some conflicts to be expected (if that is merged before this PR, which is very likely since #133030 addresses a long standing feature request and targets 2.5 while this PR targets main)

Oh, well, not until your comment :sweat_smile: Both this PR and #13303 solve the same underlying issue #11571 [^1], so only one of these should be merged anyway. The code in this PR was written one month before #13303 appeared (so the latter didn't turn up when I looked for related PRs), but I intentionally waited to see if my changes were stable. You just beat me to the punch.

Therefore closing this PR for now, feel free to reopen.

[^1]: Albeit in two slightly different ways - resetting to 0 when Auto DJ is disabled vs. resetting to 0 when a crossfade has finished & only one track is playing. Both ways have their pros and cons.

ronso0:
> You just beat me to the punch

Well, sorry, actually it was @Eve00000 and I adopted their PR ; ) Anyway, sorry you invested time for this.

In general, especially with significant UX changes like this (even if they'optional) it is best to discuss them first. Either comment on the bug report, open a separate feature request or join the devs on Zulip.

Happy coding!

cr7pt0gr4ph7:
> Well, sorry, actually it was @Eve00000 and I adopted their PR ; ) Anyway, sorry you invested time for this.

No harm done :) I had an immediate, personal pain point which I wanted to solve, so whipping up a bit of code was teh quickest way to do it.

> In general, especially with significant UX changes like this (even if they'optional) it is best to discuss them first. Either comment on the bug report, open a separate feature request or join the devs on Zulip.

Will do :) If often experiment for myself before going into a discussion, though, so that I'm aware of possible solutions and alternatives - sometimes the experiment actually turns out to be enough of a solution. I'll try to take care not to unnecessarily take up your time with duplicate work in the future. (On another note, I didn't actually consider this UX change to be this significant, though, when implemented in an unobtrusive way).

PR #13333, opened 2024-06-06
(fix) Pref Mixer: don't update EQs/QuickEffects while applying

ronso0:
I just noticed a bug in the Mixer preferences:
* check "Same EQ effect for all decks"
* change EQ in deck 1 (selected for all decks)
* Apply
* = only deck**1** EQ is applied, others are reset
* uncheck + check "Same EQ effect for all decks", apply
* = deck**2** EQ is applied
* uncheck + check "Same EQ effect for all decks", apply
* = deck**3** EQ is applied
* ...

Reason was that the `effectChanged` callback resets the pref UI to the current engine state, i.e. wipes pending changes.
Fixed with yet another bool flag.

I adopted this scheme for the similiar callback for QuickEffects and removed the extra slot (which didn't set `m_quickEffectIndiciesOnUpdate` anyway, i.e. would needlessly reload the QuickEffect and potentially reset the Super knob).

### DIFF
--- a/src/preferences/dialog/dlgprefmixer.cpp
+++ b/src/preferences/dialog/dlgprefmixer.cpp
@@ -93,7 +93,9 @@ DlgPrefMixer::DlgPrefMixer(
           m_gainAutoReset(false),
           m_eqBypass(false),
           m_initializing(true),
-          m_updatingMainEQ(false) {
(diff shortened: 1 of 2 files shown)


daschuer (approved):
LGTM, Thank you

PR #13335, opened 2024-06-06
(opti/cleanup) make use of WLibrary::getCurrentTrackTableView()

ronso0:
and move WLibraryTableView::moveSelection to WTracktableView (simplifies LibraryControl etc.)

This allows to call many methods directly from LibraryControl and remove virtuals from libraryview.h and the respective overrides from various library features.

### DIFF
--- a/src/widget/wlibrary.cpp
+++ b/src/widget/wlibrary.cpp
@@ -106,12 +106,11 @@ WTrackTableView* WLibrary::getCurrentTrackTableView() const {
     QWidget* pCurrent = currentWidget();
     WTrackTableView* pTracksView = qobject_cast<WTrackTableView*>(pCurrent);
     if (!pTracksView) {
-        // This view is no tracks view, but maybe a special tracks view with a
-        // controls row (DlgAutoDJ, DlgRecording)?
-        // qDebug() << "   view is no tracks view. look for tracks view child";
+        // This view is not a tracks view, but possibly a special library view
+        // with a controls row and a track view (DlgAutoDJ, DlgRecording etc.)?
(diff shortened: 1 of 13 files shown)


src/widget/wlibrary.cpp
```diff
@@ -106,8 +106,8 @@ WTrackTableView* WLibrary::getCurrentTrackTableView() const {
     QWidget* pCurrent = currentWidget();
     WTrackTableView* pTracksView = qobject_cast<WTrackTableView*>(pCurrent);
     if (!pTracksView) {
-        // This view is no tracks view, but maybe a special tracks view with a
-        // controls row (DlgAutoDJ, DlgRecording)?
+        // This view is no tracks view, but maybe a special library view with a
+        // controls row and a track view (DlgAutoDJ, DlgRecording)?
```
cr7pt0gr4ph7, line 110: Were already cleaning up this comment, so it would make sense to fix all spelling & grammatical errors:
```
// This view is not a tracks view, but possibly a special library view
// with a controls row and a tracks view (DlgAutoDJ, DlgRecording,
// DlgHidden, DlgMissing, ...).
```

cr7pt0gr4ph7 (changes requested):

cr7pt0gr4ph7:
Looks good (apart from the comment spelling nitpick).

src/widget/wlibrary.cpp
ronso0: Thanks!

FYI if you mark lines in the file diff tab you can click the +- button for change suggestions. This makes it easier to spot the difference.

JoergAtGithub:
LGTM! Thank you!

PR #13200, opened 2024-05-03
Usability/Keyboard: Add Ctrl+Shift+F shortcut to search the main tracks database

cr7pt0gr4ph7:
The new shortcut `Ctrl+Shift+F` (`Search globally`[^1]) complements the existing shortcut `Ctrl+F` (`Search in current location`).

The use case for this is pretty simple (and motivated by me teaching dance classes) and probably pretty common elsewhere, too:

> After digging around in my playlists & crates, leaving the computer to teach, and then coming back to it, I want to be able to quickly search for a specific song without having to first pick up the mouse or fiddle around with the trackpad.

[^1]: Ok, it is only "globally" as long as all tracks are present in the main Mixxx database, and not contained in external libraries like iTunes or Rekordbox.[^2]

[^2]: As a later improvement, one could extend `Ctrl+Shift+F` to, either:
a. Actually initiate a global search over all available libraries, or:
b. Make default library to which `Ctrl+Shift+F` jumps configurable in the preferences, or:
c. Provide different shortcuts for the different external libraries.

Issue #12719, opened 2024-02-04
Rename colours?


mxmilkiib:
### Feature Description

It would be neat to be able to rename colors, allowing for their use to be more transparent.

In Rekordbox; https://youtu.be/kM6s6wSBLHo?t=444

Issue #11107, opened 2022-12-03
Sort by column


conducator:
### Bug Description

Hi, while I can sort songs by clicking any column in the Playlists visualization (duration, key, artist, album, or any other), I can't do the same in the Autodj window, so I can't mix matching the key, 

### Version

2.3.3

### OS

Windows 11

ronso0:
In AutoDJ tracks are always sorted by # so the tracks should always be in the order they'll be played (and removed / requeued) to aoid any confusion.

The AutoDJ view is not meant for manual mixing, if you want to do so simply copy all tracks to a separate playlist.

conducator:
Well this feature would be nice to have in Autodj (sort columns like in Playlist view) not for manual mixing (if it's called Autodj it's not for manual mixing...). You could use Autodj for harmonic mixing too and not only using always the same # tracks

ronso0:
Hmm for mixing by key that makes sense, probably also for in- or decreasing the BPM during an AutoDJ mix.

However, if sorting was enabled, and the position # could be adjusted when sorting, if you had the tracks carefully ordered before, and didn't intend to change the play order, there is no way to restore the previous order.
How could the UI be adjusted to avoid mistakes like that?

daschuer:
In Playlists, you can sort by any column, but the position # of the track in the playlist is kept. So when you sort by # everything is back in consecutive order. 
If we do the same at Auto DJ, you loose track of which track is played next, because it is no longer the topmost. That's why this type of sorting is disabled. 

Playing the tracks by key is another type of sorting. It is more like Shuffle, which we already have. Schuffle with additional rules. 
 
I can imagine to implement that, but that is nothing we should do by a column click.

daschuer:
This duplicates: 
https://github.com/mixxxdj/mixxx/issues/6509
https://github.com/mixxxdj/mixxx/issues/6335

The same for searching:
https://github.com/mixxxdj/mixxx/issues/6408

Workaround: 
https://github.com/mixxxdj/mixxx/issues/6598

conducator:
[...]

PR #13337, opened 2024-06-06
Auto DJ: (Re-)enable search in the Auto DJ view

cr7pt0gr4ph7:
The search feature for the Auto DJ view  was originally disabled by 54f71ea7b8e5d6e87ad9132bcf7fdfe2a7c8372a over 12 years ago.

We now have some new tools available, which means we can implement a workaround instead of disabling search altogether. Searching within normal playlists is already possible and works correctly, and the Auto DJ queue is basically just another playlist, so being able to search there matches user expectations.

The implementation here reuses the `ProxyTrackModel` code already used by the file browser view for the actual filtering behavior (4970e30b0fd67ff3d2094bb080ec2ff749ef3701). Most of the changes here are for making `BaseSqlTableModel` play nicely with `ProxyTrackModel` (75ce4d0a48d672249100bbaa9da7c00959575f09), for making cut, copy & paste work (62c3a0f0435121e0ebbdc800ca1f6c657fa72a39) and behave correctly w.r.t. keyboard focus (35569a49f71955256cb8f514be2ef25f9de94b4f).

### DIFF
--- a/src/library/autodj/autodjfeature.cpp
+++ b/src/library/autodj/autodjfeature.cpp
@@ -192,7 +192,9 @@ TreeItemModel* AutoDJFeature::sidebarModel() const {
 void AutoDJFeature::activate() {
     //qDebug() << "AutoDJFeature::activate()";
     emit switchToView(m_viewName);
-    emit disableSearch();
+    if (m_pAutoDJView) {
+        emit restoreSearch(m_pAutoDJView->currentSearch());
+    }
     emit enableCoverArtDisplay(true);
 }
 
--- a/src/library/autodj/autodjfeature.cpp
+++ b/src/library/autodj/autodjfeature.cpp
@@ -25,18 +25,18 @@
 namespace {
 constexpr int kMaxRetrieveAttempts = 3;
 
-    int findOrCrateAutoDjPlaylistId(PlaylistDAO& playlistDAO) {
-        int playlistId = playlistDAO.getPlaylistIdFromName(AUTODJ_TABLE);
-        // If the AutoDJ playlist does not exist yet then create it.
-        if (playlistId < 0) {
-            playlistId = playlistDAO.createPlaylist(
-                    AUTODJ_TABLE, PlaylistDAO::PLHT_AUTO_DJ);
-            VERIFY_OR_DEBUG_ASSERT(playlistId >= 0) {
-                qWarning() << "Failed to create Auto DJ playlist!";
-            }
+int findOrCrateAutoDjPlaylistId(PlaylistDAO& playlistDAO) {
+    int playlistId = playlistDAO.getPlaylistIdFromName(AUTODJ_TABLE);
+    // If the AutoDJ playlist does not exist yet then create it.
+    if (playlistId < 0) {
+        playlistId = playlistDAO.createPlaylist(
+                AUTODJ_TABLE, PlaylistDAO::PLHT_AUTO_DJ);
+        VERIFY_OR_DEBUG_ASSERT(playlistId >= 0) {
+            qWarning() << "Failed to create Auto DJ playlist!";
         }
-        return playlistId;
     }
+    return playlistId;
+}
 } // anonymous namespace
 
 AutoDJFeature::AutoDJFeature(Library* pLibrary,
--- a/src/library/autodj/dlgautodj.cpp
+++ b/src/library/autodj/dlgautodj.cpp
@@ -7,6 +7,7 @@
 #include "controllers/keyboard/keyboardeventfilter.h"
 #include "library/library.h"
 #include "library/playlisttablemodel.h"
+#include "library/proxytrackmodel.h"
 #include "moc_dlgautodj.cpp"
 #include "track/track.h"
 #include "util/assert.h"
@@ -33,6 +34,7 @@ DlgAutoDJ::DlgAutoDJ(WLibrary* parent,
                   pLibrary,
                   parent->getTrackTableBackgroundColorOpacity())),
           m_bShowButtonText(parent->getShowButtonText()),
+          m_pProxyTableModel(nullptr),
           m_pAutoDJTableModel(nullptr) {
     setupUi(this);
 
@@ -78,7 +80,15 @@ DlgAutoDJ::DlgAutoDJ(WLibrary* parent,
 
     // We do _NOT_ take ownership of this from AutoDJProcessor.
     m_pAutoDJTableModel = m_pAutoDJProcessor->getTableModel();
-    m_pTrackTableView->loadTrackModel(m_pAutoDJTableModel);
+
+    // Wrap the table model so users can search it without
+    // messing up the table view for the AutoDJProcessor.
+    m_pProxyTableModel = new ProxyTrackModel(
+            m_pAutoDJTableModel,
+            /* handle searches*/ true);
+
+    // Use the proxy model for the displayed table & searches
+    m_pTrackTableView->loadTrackModel(m_pProxyTableModel);
 
     // Do not set this because it disables auto-scrolling
     //m_pTrackTableView->setDragDropMode(QAbstractItemView::InternalMove);
@@ -232,6 +242,7 @@ DlgAutoDJ::~DlgAutoDJ() {
     // Delete m_pTrackTableView before the table model. This is because the
     // table view saves the header state using the model.
     delete m_pTrackTableView;
+    delete m_pProxyTableModel;
 }
 
 void DlgAutoDJ::setupActionButton(QPushButton* pButton,
@@ -245,12 +256,15 @@ void DlgAutoDJ::setupActionButton(QPushButton* pButton,
 
 void DlgAutoDJ::onShow() {
     m_pAutoDJTableModel->select();
+    m_pProxyTableModel->select();
 }
 
 void DlgAutoDJ::onSearch(const QString& text) {
-    // Do not allow filtering the Auto DJ playlist, because
-    // Auto DJ will work from the filtered table
-    Q_UNUSED(text);
+    m_pTrackTableView->onSearch(text);
+}
+
+const QString DlgAutoDJ::currentSearch() const {
+    return m_pProxyTableModel->currentSearch();
 }
 
 void DlgAutoDJ::shufflePlaylistButton(bool) {
--- a/src/library/autodj/dlgautodj.h
+++ b/src/library/autodj/dlgautodj.h
@@ -10,6 +10,7 @@
 #include "track/track_decl.h"
 
 class PlaylistTableModel;
+class ProxyTrackModel;
 class WLibrary;
 class WTrackTableView;
 class Library;
@@ -32,6 +33,7 @@ class DlgAutoDJ : public QWidget, public Ui::DlgAutoDJ, public LibraryView {
     void onSearch(const QString& text) override;
     void saveCurrentViewState() override;
     bool restoreCurrentViewState() override;
+    const QString currentSearch() const;
 
   public slots:
     void shufflePlaylistButton(bool buttonChecked);
@@ -71,6 +73,7 @@ class DlgAutoDJ : public QWidget, public Ui::DlgAutoDJ, public LibraryView {
     WTrackTableView* const m_pTrackTableView;
     const bool m_bShowButtonText;
 
+    ProxyTrackModel* m_pProxyTableModel;
     PlaylistTableModel* m_pAutoDJTableModel;
 
     QString m_enableBtnTooltip;
--- a/src/library/basesqltablemodel.cpp
+++ b/src/library/basesqltablemodel.cpp
@@ -36,6 +36,19 @@ const QString COLUMNS_SORTING = QStringLiteral("ColumnsSorting");
 
 const QString kModelName = "table:";
 
+QList<int> columnNamesToIndices(
+        const QStringList& columnNames, const BaseSqlTableModel& tableModel) {
+    QList<int> indices;
+    indices.reserve(columnNames.size());
+    for (const QString& columnName : columnNames) {
+        auto index = tableModel.fieldIndex(columnName);
+        if (index < 0) {
+            continue;
+        }
+        indices.append(index);
+    }
+    return indices;
+}
 } // anonymous namespace
 
 BaseSqlTableModel::BaseSqlTableModel(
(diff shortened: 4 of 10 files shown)


cr7pt0gr4ph7:
Huh, seems like I found a slight bug with the code formatting rules when confronted with changes in files that contain old, wrongly indented code.

ronso0:
Just a hint: IIRC from a related discussion, one of the reasons to keep the search disabled in Auto DJ was that it is not obvious that the view is filtered. A possible solution, also relevant here IMHO, would be inserting narrow, empty dummy lines where the invisible tracks would be. Not sure how that could work though.

cr7pt0gr4ph7:
> Just a hint: IIRC from a related discussion, one of the reasons to keep the search disabled in Auto DJ was that it is not obvious that the view is filtered. A possible solution, also relevant here IMHO, would be inserting narrow, empty dummy lines where the invisible tracks would be. Not sure how that could work though.

Okay, I understand the concerns. I tried to look up the discussion you where mentioning, but couldn't find it. After thinking a bit about it, let me explain my reasoning why I think this is should not be a problem in practice.

cr7pt0gr4ph7:
These are just my two cents, though. I appreciate feedback whether I missed something!

As an aside, here are possibly related issues from the past that I could find:
- Search
  - #7805
  - #6408
  - #6090
- Sorting (only tangentially related w.r.t. the implementation in this PR, but related to the user's mental model of filtering and sorting playlists)
  - #11107
  - #6335

cr7pt0gr4ph7:
My personal, real-world use case that motivated this PR is this:

I have prepared a 3 hour/100 track playlist in the Auto DJ queue for a dance party, and want to remove a specific track from the queue because I have added another copy of it near the top of the queue during the course of the party.

Without neither search nor sorting, this is an arduous task - that's why we have search. Inserting narrow placeholder rows or greying out filtered out tracks instead of fully hiding them would be slightly better, but would still force me to scroll through a large list of tracks.

It seems to me that there are two camps of users:
- Those that want to interact with the Auto DJ just like with a playlist, including sorting & filtering, and whose mental model is that Mixxx will play the underlying playlist, and not their filtered view of it. (This is the solution proposed here, and the one employed by other music players like Spotify)

- Those that expect the visible list in the Auto DJ view to be what will be played, including sorting (and probably filtering, for the sake of consistency). There are some consistency problems with implementing this IMO, but it would be possible. (This is the solution proposed in #6335 AFAIU)

- Those that fall in the middle of not particularly needing to filter a large Auto DJ queue, and either do not expect the filters/sorting to affect the play order, or do not care because they don't use (or know about) it anyway. The main concern is to prevent those users from accidentally shooting themselves in the foot through unintended filtering. (This correlates with the compromise of greying out tracks instead of hiding them, as proposed before).

- And the fourth solution: Disabling search to postpone a decision on the behavior.

I think that the second solution conflicts with how playlists (that is, the specific objects called "playlists" in the UI, not the general concept which also includes in the Auto DJ queue) behave currently in Mixxx, and thus is undesirable in my opinion, but could be solved using a config option or separate feature. The third and fourth solutions are compromises that give noone what they want, except for the people that do not need or use the feature anyway.

ronso0:
Thanks for elaborating!

> Visual Indicators: How do we indicate that the playlist is currently being filtered?
> * The same way as in all other library views: By there being text in the search field (and maybe an additional visual indicator, see below).

tracked by #7922

ronso0:
> How do we avoid users from being confused because they can't see any tracks (due to their search terms filtering out all available tracks)?

I think a status bar / info bar could also solve this. (stale PR #3406).

> an unselectable "phantom row" at the end of the tables

That help much IMO. If the filtered tracks fill the view such a bottom row would be visible.

daschuer (changes requested):
I can confirm your use case. 

However, I am strictly against filtering and sorting in the Auto-DJ playlist. The reliable view what will happen next is, IMHO much more important than any desire for sorting and filtering.

I often search for a successor and forget to go to the Track view first. If that happens with Auto DJ late at night, even a warning bar would be not enough, considering other visual clutter we have already. 

I like however @ronso0  idea to highlight and probably select and auto scroll to the filter matches. That is IMHO quiicker as a normal filtering for your tracks.
You just need to press del and the tracks are gone. 
What do you think?

cr7pt0gr4ph7:
> I can confirm your use case.
> 
> However, I am strictly against filtering and sorting in the Auto-DJ playlist. The reliable view what will happen next is, IMHO much more important than any desire for sorting and filtering.

After giving it some more thought, I can't really come up with a valid use case for sorting an already running Auto DJ playlist, either, so we can restrict the discussion to filtering only, and see whether we can find a good solution for that.

> I often search for a successor and forget to go to the Track view first. If that happens with Auto DJ late at night, even a warning bar would be not enough, considering other visual clutter we have already.

Yeah, I'd be kind of opposed to a warning bar / status bar at the bottom of the screen, too, due to its distraction potential. The search bar is basically _the_ main entry point for starting to interact with tracks/playlists (see #13200), so any visual indicators or such, if any, should be in close visual proximity to it or the track list, because that are the places users will look at.

I think that your use case is very valid. Just to understand correctly, what are your main concerns/issues you want to avoid?
- Accidentally modifying the Auto DJ playlist
- Taking too long to realize that you're in the Auto DJ view, and won't find the tracks you are looking for there
- Taking a glance at the Auto DJ playlist to see what will be played next, and not noticing that there is a filter active

Also, for clarification, are you using an external DJ controller or a Mixxx-only setup? (For context, I'm running Mixxx on a laptop and do not have any experience whatsoever with external controllers).

> I like however @ronso0 idea to highlight and probably select and auto scroll to the filter matches. That is IMHO quiicker as a normal filtering for your tracks. You just need to press del and the tracks are gone. What do you think?

I don't know about introducing a whole different behavior of the search control just for one library view, though. That has a great potential for confusion, and feels a bit like overkill. I like the idea of scrolling, but that only works well for a single selected track. As soon as multiple tracks are selected, you have to find a way for the user to see that there are offscreen track which are selected, and for the user to cycle through those tracks, ...

cr7pt0gr4ph7:
The best solutions I think of so far (based on the suggestions here and elsewhere) are:
1. Providing a config option to enable/disable searching the Auto DJ view (but that feels like cheap pseudo-solution because it's not really discoverable for new users)
2. Visually indicating that there are rows which have been filtered out in the Auto DJ track list itself:

| # | Title | Artist | Album |
| -- | -- | -- | -- |
| 1 | ... Filtered out ... | | |
| 2 | Some track | A | B |
| 3 | Some other track | A | B |
| 4 - 8 | ... Filtered out ... | | |
| 9 | Another track | C | D |
| 10 - 50 | ... Filtered out ... | | |

ronso0:
> I like however @ronso0 idea to highlight and probably select and auto scroll to the filter matches. That is IMHO quiicker as a normal filtering for your tracks.

For the record, I didn't propose that (here ; ) Maybe in some ancient combo. I agree it's bad UX)

Let's take a step back and ignore the imlipcit constraints we set when thinking of the AutoDJ as a regular (playlist) view, e.g. when sorting all rows are sorted, when filtering the filter is strict, no exceptions.

There are basically two situations where filtering and/or sorting can be applied, AutoDJ ON or OFF, but IMO there are different constraints / requirements / expectations:
* AutoDJ on:
  * I agree the next 1, 2, 3 queued tracks should always be visible
  * sorting like in other views can be very confusing, especially in the heat of the night
* AutoDJ off: 
  * no restrictions required, it's just a playlist

In both situations filtering / sorting may be desired, mostly like in playlists:
* clean up the view, remove duplicates
* find a good next track (BPM, key, genre, artist ...) and pull that to the top (drag or with Alt keys, Alt+Home/PageUp)
  * here we have to views: sorted/filtered with a selection and unfiltered, sorted by #
  * it is absolutely necessary that the **selection from the filtered/sorted view is maintained** when switching back to the # view in order to move it to the desired position

So if we now have a somewhat independent table model (at least we can tweak the filtered view to our liking) could it be possible to present the filtered view differently?
1. Like I mentioned above: when filtering, indicate that tracks (spans) are hidden by replacing the hidden tracks with some sort of placeholder? an empty row that is only half the height of a normal row, perhaps with filled with a gradient or other type of infill that distinguishes it from a regular track row?
2. enforce the # column, always left most column, paint it in a special way if the queue is **not** sorted by #12719 
Maybe turn it into  a vertical header so it's always visible?

What do you think?

daschuer:
> I agree the next 1, 2, 3 queued tracks should always be visible

I can confirm.  They can be grayed out or such. That's still ok for me. 

For the rest, the view proposed in https://github.com/mixxxdj/mixxx/pull/13337#issuecomment-2156151728
works for me. However I would prefer to see all of them grayed out. This should be more easy to implement. 
From there we can than interate after some brief testing.

daschuer:
Let's not allow sorting. 

What is the reason you need to see all filter matches on a single page? 

If we pin the first four, and than replace the hidden rows with a single line as you demonstrate above that could work. Maybe with a + icon to uncollapse these rows and remove the filter.

cr7pt0gr4ph7:
Sorry, this PR is still active, just didn't find time to work on it. I'll try to implement the requested features some time in the next months.

ronso0:
Coming back to this after I read https://mixxx.discourse.group/t/search-function-for-auto-dj/32934

Recently I noticed that Betterbird (a ThunderBird fork) changes the background of the email list when a Quick Filter is active. It's pattern of diagonal lines, clearly distinguishing the regular, full view from the filtered one.
I think it's dark grey lines for 'filtered' and drak red lines for 'filtered, no matches'.

So, while I like some of the ideas we discussed here (especially keeping next 1-3 tracks visible), I'd prefer to implement this kind of "filtered" indicator and only keep tracks  1,2,3.

I'll publish a quick POC when it's ready.

daschuer:
Nice, here is a screenshot for reference. 
https://share.google/ZisdK9DNbGhHFDgL0

cr7pt0gr4ph7:
Sorry, I haven't been active for the last months due to other things requiring my time.

Just for reference, I've been running this patch[^1] in production for 1,5 years now (including at dance parties and weddings) and can offer my experience so far:

[^1]: Which enables filtering but still disallows sorting.

- **Filtering the Auto DJ view** is _really useful_ for changing play order on the fly based on the mood, especially when you (1) have prepared a large playlist, (2) use the shuffle feature, or (3) want to see whether you already have tracks of this specific artist in your setlist.

  What I'm mostly missing is just a context menu entry for **`Move to Top/Bottom (Auto DJ Queue)`**, which would nicely mirror the `Add to Auto DJ Queue (top/bottom/replace)` entries[^2]. _My current workaround is using `Ctrl+X` -> `Ctrl+V`._
  Also a `Play Next (Auto DJ)` / `Load as Next (Auto DJ)` / `Move to Top (Next Track)` entry would be nice if I want to replace an already loaded next track. My current workaround is _"Disable Auto DJ (`Shift+F12`)" -> "Add to Top" -> "Enable Auto DJ (`Shift+F12`)"_, but that's annoying if time is of the essence.

   Thinking about it, it could also make the most sense if we have the following set of context entries: `Move to Top (Play Next)`, `Move to Top+1`, `Move to Top+2`, `Move to Top+3`, `Move to Bottom`, as that allows for the common use case of planning out the next few tracks.

[^2]: And should probably be mapped to the same keyboard shortcuts.

- The issue of _**"Why the f* don't I see this track..?!?... ooooh, I have a filter active"**_ isn't exclusive to the Auto DJ view, but applies to all views equally.

  I've learned to subconciously check whether the search field has content, but a **visual indication** (like a small orange border around the search field or the table, or a colored background for the search field, or a striped background like Betterbird) would be nice[^3].

[^3]: Sidenote 2: We should probably delegate the decision of the specific visuals to the skin, that is: (1) Whether to highlight the search field, the tracks table, or both, and/or display an icon somewhere (2) Which highlighting color to use (3) Whether to use a solid or dashed border, a solid background, a striped background...

- Based on my experience, I don't feel like we need to insert **placeholder rows**, for the following reasons:

  1. The index column already tells the same information.
  2. The immediately next track is always visible in the deck.
  3. Users are most likely to learn quickly if we offer a _single, uniform visual indication_ for filtering that's consistent across all other views as well.
  4. Vertical screen space is limited, especially when using 4 decks on a laptop.

  The _much_ bigger Auto DJ-related footgun is that "Move to Auto DJ Queue (replace)" doesn't ask for confirmation before accidentally clearing out your 3 hour playlist (though that may have been fixed in newer versions, as I'm currently running an older version due to my extensive patches), and there's no "Undo" function.

- Also, a **placeholder (or different visual indication) when your search filters out all tracks** would be useful. Betterbird uses a red striped background instead of gray striped background, but I would tend to propose just a line of text _`Your current search "<search text>" filters out all tracks (Click here to clear filter)`_ (or _`No tracks found... (click here to clear filter)`_) in italics.

  The placeholder should be selectable, so that a pressable navigation knob on a DJ controller can be easily used to clear the filter.

cr7pt0gr4ph7:
I'm also a user of Betterbird, so I can say that:
- The striped background when searching is a bit jarring at first (_"You type something, and suddenly a lot of red appears on the screen"_ makes you jump a bit and makes you intuitively think that an error has happened), so a slightly more subtle visual indication would be better for the purpose of not scaring users.

- But it works well for the purpose of preventing you from accidentally having a search active when you didn't intend to.

ronso0:
@cr7pt0gr4ph7 Thanks for your feedback!

> What I'm mostly missing is just a context menu entry for Move to Top/Bottom (Auto DJ Queue)

Since we introduced the `Alt` move keys we have `Alt`+`Home`for this purpose (or `Pos1` or whatever it is called on var. keyboards).
I'm hesitant to add more actions to the context menu, after all we also support Drag'n'drop in playlist (which admittedly may not be practical when you have to move a track from the bottom up to the top, and also there seems to be Qt regression with very slow DnD auto-scrolling

> Thinking about it, it could also make the most sense if we have the following set of context entries: Move to Top (Play Next), Move to Top+1, Move to Top+2, Move to Top+3, Move to Bottom

So "move to top+2" would be `Alt`+`Home` and `Alt`+`Down`

ronso0:
> The much bigger Auto DJ-related footgun is that "Move to Auto DJ Queue (replace)" doesn't ask for confirmation before accidentally clearing out your 3 hour playlist

Indeed, that would be consistent with the existing warning for manual track removal in playlists and crates.
I _thought_ we have some protection against that :thinking: but we don't..

_edit_
Ah we have the popup for the sidebar action "Clear AutoDJ queue"

ronso0:
I managed to get the background warning, but it's not polished, yet. The `alternate-background-color` interferes with the `background-image` and it looks a bit odd.
And yes, this needs to be styled on skin level. In qss we can style it for _all_ WTrackTableViews or individually for all that are in some feature `Dlg..` (AutoDJ, Hidden, Missing, Analyze, Recording).

I'll open a PR for testing, then we can decide whether we want this everywhere (in every library view) or only in Auto DJ.

PR #13354, opened 2024-06-09
Optimize control code

JoergAtGithub:
This PR collects several low level optimizations for our control code. See the single commit messages.

We have currently 36816 ControlDoublePrivate instances in Mixxx the memory realignment reduces the size from 464 byte to 448 byte, this translates to 0.5 MegaByte heap memory reduction.
Furthermore the first fields are aligned to fit into a 128 byte SSE register. But I guess there are very few operation, which can benefit from an SSE copy etc. (in theory we could further optimize for SSE, by reducing ControlValueAtomic<double> from 136byte to 128byte).

Old memory alignment:
ControlDoublePrivateBefore

New memory alignment:
ControlDoublePrivateThisPR


daschuer:
The reason that ControlDoublePrivate is so big is a nasty regression. It should be only for bites. The fix is here:  https://github.com/mixxxdj/mixxx/pull/13355

src/control/control.h
```diff
@@ -183,8 +183,29 @@ class ControlDoublePrivate : public QObject {
...
+    // User-visible, i18n name for what the control is.
+    QString m_name;
+
+    // User-visible, i18n description for what the control does.
+    QString m_description;
+
+    // The control value.
+    ControlValueAtomic<double> m_value;
+    // The default control value.
+    ControlValueAtomic<double> m_defaultValue;
```
daschuer, line 197: How do we protect that for get cluttered again? Maybe define regions via comments?

src/control/controlobject.cpp
```diff
@@ -14,7 +14,7 @@ ControlObject::ControlObject(const ConfigKey& key,
         bool bTrack,
         bool bPersist,
         double defaultValue)
-        : m_key(key) {
+        : m_key(std::move(key)) {
```
daschuer, line 17: We cannot move form a const &. I guess this std::move which boils down to a cast is ignored. 
In this case, we need to pass the key by value to make use of the full benefit of avoiding reference counting.

src/control/controlobject.cpp
```diff
@@ -103,7 +103,7 @@ void ControlObject::setParameterFrom(double v, QObject* pSender) {
 
 // static
 void ControlObject::set(const ConfigKey& key, const double& value) {
-    QSharedPointer<ControlDoublePrivate> pCop = ControlDoublePrivate::getControl(key);
+    auto pCop = ControlDoublePrivate::getControl(key);
```
daschuer, line 106: The function name getControl() does not reveal that it returns a `QSharedPointer<ControlDoublePrivate>`  so we need to keep the full type of rename the function.

src/control/controlpushbutton.cpp
```diff
@@ -10,12 +10,7 @@ ControlPushButton::ControlPushButton(const ConfigKey& key, bool bPersist, double
         : ControlObject(key, false, false, bPersist, defaultValue),
           m_buttonMode(PUSH),
           m_iNoStates(2) {
-    if (m_pControl) {
-        m_pControl->setBehavior(
-                new ControlPushButtonBehavior(
-                        static_cast<ControlPushButtonBehavior::ButtonMode>(m_buttonMode),
-                        m_iNoStates));
-    }
+    setBehavior();
```
daschuer, line 13: A setter usually takes a parameter. Maybe `updateBehaviour()`?  

A setBehaviour() function taking num_states and mode would be also beneficial for places like here, where the first behaviour is allocated and destroyed right away: 
https://github.com/mixxxdj/mixxx/blob/8a00cc00e8e5fad1dfaf5d16c1a4bfcd98c60f3e/src/engine/controls/vinylcontrolcontrol.cpp#L42-L43

src/control/pollingcontrolproxy.h
```diff
@@ -19,7 +19,7 @@ class PollingControlProxy {
     }
 
     PollingControlProxy(const ConfigKey& key, ControlFlags flags = ControlFlag::None) {
-        m_pControl = ControlDoublePrivate::getControl(key, flags);
+        m_pControl = ControlDoublePrivate::getControl(std::move(key), flags);
```
daschuer, line 22: getControl does not always keep the key, so moving has no benefit here. Conditionally copy form a const & should work best.

daschuer (changes requested):

src/control/control.h
Swiftb0y: We could try `-Wpadded` https://developers.redhat.com/blog/2016/06/01/how-to-avoid-wasting-megabytes-of-memory-a-few-bytes-at-a-time#tools

JoergAtGithub:
I merged in the changes from #13355, but I noticed, that the memory consumption of ControlDoublePrivate is much bigger in this Main branch, than in the 2.4 branch.

2.4:
grafik

Main:
grafik

This PR:
grafik

daschuer:
I assume you build main with Qt6 and 2.4 with Qt5. Maybe the `sizeof(QString)` has changed due to the Small String Optimization?

JoergAtGithub:
> I assume you build main with Qt6 and 2.4 with Qt5. 

Yes, I did!

daschuer:
Qt6 has no SSO. The number in increased due to the use of a common QArrayDataPointer with holding two pointers one the header for implicit sharing, one to the actual data and one to the size which is now also 64 bit. This allows faster read only access and allows to the store strings longer than 2.147.483.648 characters (more than 480 Bibles)

src/control/control.h
```diff
@@ -183,8 +183,29 @@ class ControlDoublePrivate : public QObject {
...
+
     QAtomicPointer<ControlObject> m_pCreatorCO;
 
+    QString m_trackKey;
+
```
daschuer, line 202:
```suggestion

// Note: keep the order of the members below to not introduce gaps due to memory alignment in this often used class. 
```

daschuer (changes requested):

src/control/control.h
JoergAtGithub: I tried -Wpadded on CI, but it does not only warn about padding inbetween the data structure, but also if the overall size is not matching the byte alignment size, e.g. any class with 6 bytes would need to filled up with 2 explicit padding bytes.

I found also a currently unused macro in https://github.com/mixxxdj/mixxx/blob/15b4444584b9cd2514946d6a645746cae3f5a315/src/util/platform.h#L15

Maybe someone on Linux could try: https://linux.die.net/man/1/pahole

Swiftb0y: Okay that makes sense I guess when the compiler expects you to nest these packed data structures. The only other potential solution I found was to generate two versions of the type, one packed (potentially with suboptimal alignment and then `static_assert(sizeof(T) == sizeof(_T_packed)) (potentially with a bit of rounding to manually account for the overall size). I wouldn't know how to accomplish that without copy pasting or a macro (doing the copy-pasting automatically) though.

JoergAtGithub:
@daschuer Friendly Ping

src/control/controlpushbutton.h
```diff
@@ -27,42 +28,36 @@
...
     virtual ~ControlPushButton();
 
-    inline ButtonMode getButtonMode() const {
+    inline ControlButtonMode getButtonMode() const {
         return m_buttonMode;
     }
-    void setButtonMode(enum ButtonMode mode);
+    void setButtonMode(enum ControlButtonMode mode);
     void setStates(int num_states);
+    void setBehavior(enum ControlButtonMode mode, int num_states);
```
Swiftb0y, line 56: I'm pretty sure prefixing the type with `enum` is only really there for C-compatibility.
```suggestion
    void setButtonMode(ControlButtonMode mode);
    void setStates(int num_states);
    void setBehavior(ControlButtonMode mode, int num_states);
```

src/control/controlpushbutton.cpp
```diff
@@ -8,39 +8,50 @@
...
+    }
+}
 
+// private
+void ControlPushButton::updateBehavior() {
     if (m_pControl) {
         m_pControl->setBehavior(
                 new ControlPushButtonBehavior(
-                        static_cast<ControlPushButtonBehavior::ButtonMode>(m_buttonMode),
+                        static_cast<ControlButtonMode>(m_buttonMode),
```
Swiftb0y, line 54: unnecessary cast now that the enum type got unified.

src/control/controlbuttonmode.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+enum class ControlButtonMode {
+    PUSH = 0,
+    TOGGLE,
+    POWERWINDOW,
+    LONGPRESSLATCHING,
+    TRIGGER
+};
```
Swiftb0y, line 9: since we're already touching every spot anyways. lets adhere to the core guidelines:
[Enum.5: Don’t use ALL_CAPS for enumerators](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-caps) and [Enum.8: Specify enumerator values only when necessary](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-value)
```suggestion
enum class ControlButtonMode {
    Push,
    Toggle,
    PowerWindow,
    LongPressLatching,
    Trigger
};
```

src/control/controlpushbutton.h
```diff
@@ -27,42 +28,36 @@
...
+        case ControlButtonMode::TOGGLE:
+            return "TOGGLE";
+        case ControlButtonMode::POWERWINDOW:
+            return "POWERWINDOW";
+        case ControlButtonMode::LONGPRESSLATCHING:
+            return "LONGPRESSLATCHING";
+        case ControlButtonMode::TRIGGER:
+            return "TRIGGER";
+        default:
+            return "UNKNOWN";
```
Swiftb0y, line 44: do these strings require stability? Can we change them? lets remove the default case because it should not be possible to pass an invalid value and when the enum gets changed, we get notified thanks to `-Wswitch`. Also making `ControlButtonMode` a Q_ENUM would make this redundant, wdyt?
```suggestion
        case ControlButtonMode::Push:
            return QStringLiteral("Push");
        case ControlButtonMode::Toggle:
            return QStringLiteral("Toggle");
        case ControlButtonMode::PowerWindow:
            return QStringLiteral("PowerWindow");
        case ControlButtonMode::LongPressLatching:
            return QStringLiteral("LongPressLatching");
        case ControlButtonMode::Trigger:
            return QStringLiteral("Trigger");
```

Swiftb0y (commented):
a couple comments, thanks for the cleanup.

src/control/controlpushbutton.h
JoergAtGithub: Q_ENUM did not work here, because this is a static member function and Q_ENUM works only for membar variables.

src/control/control.h
daschuer: Let's not over engineer it. Just a comment state informs there reader that the oder of the variables has been chosen to avoid memory gaps due to alignment is probably sufficient.

src/control/controlpushbutton.h
```diff
@@ -1,69 +1,42 @@
...
+            return QStringLiteral("PowerWindow");
+        case ControlButtonMode::LongPressLatching:
+            return QStringLiteral("LongPressLatching");
+        case ControlButtonMode::Trigger:
+            return QStringLiteral("Trigger");
+        default:
+            // Without the default branch we get compiler warnings
+            DEBUG_ASSERT(false);
+            return "Unknown";
         }
```
Swiftb0y, line 25: I assume you are referring to "control reaches end of non-void function [-Wreturn-type]"? If you just do this: we'll still get `-Wswitch` (which we want), but not `-Wreturn-type`.
```suggestion
        }
        DEBUG_ASSERT(false);
        return QStringLiteral("Unknown");
```

src/control/control.h
JoergAtGithub: It's already added: https://github.com/mixxxdj/mixxx/blob/6728c9156caaa904698da5cef3f723279a5129ed/src/control/control.h#L203-L205

Swiftb0y (approved):
LGTM. Thank you.

Swiftb0y:
@daschuer please resolve your review

daschuer (approved):

daschuer:
Thank you!

PR #13358, opened 2024-06-10
Optimize BPMDelegate::paintItem

JoergAtGithub:
Execute expensive setStleSheet only, if textcolor of BPM checkbox(padlock) changed.
This change shows a significiant performance gain in the profiler during library scrolling - but that's unfortunately not enough to solve #11603 alone.

### DIFF
--- a/src/library/tabledelegates/bpmdelegate.cpp
+++ b/src/library/tabledelegates/bpmdelegate.cpp
@@ -108,7 +108,9 @@ void BPMDelegate::paintItem(QPainter* painter,const QStyleOptionViewItem &option
             textColor = option.palette.color(QPalette::Normal, QPalette::Text);
         }
     }
-    if (textColor.isValid()) {
+
+    if (textColor.isValid() && textColor != m_cachedTextColor) {
+        m_cachedTextColor = textColor;
         m_pCheckBox->setStyleSheet(QStringLiteral(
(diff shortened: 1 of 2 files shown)


src/library/tabledelegates/bpmdelegate.cpp
```diff
@@ -50,6 +50,7 @@ BPMDelegate::BPMDelegate(QTableView* pTableView)
 #else
     m_pFactory->registerEditor(QVariant::Double, new BpmEditorCreator());
 #endif
+    m_cachedTextColor = QColor();
```
daschuer, line 53: Is this redundant? If not, please add a comment why it is reset here.

daschuer (changes requested):

src/library/tabledelegates/bpmdelegate.cpp
JoergAtGithub: Removed!

daschuer:
This is also a regression fix from 
https://github.com/mixxxdj/mixxx/commit/3d8054c3f224c4a2e6e19b356810c1325a1aacd3
Part of the 2.5 branch. Please rebase it as well.

JoergAtGithub:
Rebased!

daschuer (approved):
LGTM, Thank you

PR #13359, opened 2024-06-10
localDateTimeFromUtc: Make argument a const reference and initialize QDateTime at construction

JoergAtGithub:
This shows a small but measureable improvement in the profiler.

### DIFF
--- a/src/util/datetime.h
+++ b/src/util/datetime.h
@@ -12,9 +12,8 @@ namespace mixxx {
 
 /// Obtain the local date time from an UTC date time.
 inline QDateTime localDateTimeFromUtc(
-        QDateTime dt) {
-    dt.setTimeSpec(Qt::UTC);
-    return dt.toLocalTime();
+        const QDateTime& dt) {
+    return QDateTime(dt.date(), dt.time(), Qt::UTC).toLocalTime();
 }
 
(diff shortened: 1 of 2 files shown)


src/util/datetime.h
```diff
@@ -12,9 +12,8 @@ namespace mixxx {
 
 /// Obtain the local date time from an UTC date time.
 inline QDateTime localDateTimeFromUtc(
-        QDateTime dt) {
-    dt.setTimeSpec(Qt::UTC);
-    return dt.toLocalTime();
+        const QDateTime& dt) {
+    return QDateTime(dt.date(), dt.time(), Qt::UTC).toLocalTime();
```
daschuer, line 16: The code is was kind of broken before.  It indicates that we pass around QDateTime with a wrong time zone set. 
I can confirm that the original function is slow in Qt6, because it uses the deprecated Qt5 wrapper. 

Do you have interest to set the correct time zone initially and dispose this function completely?

Otherwise please add a Todo.

daschuer (changes requested):

JoergAtGithub:
Rebased!

daschuer (approved):
LGTM, Thank you.

PR #13364, opened 2024-06-12
(fix) Library: repair Del hotkey in sidebar, add Clear Queue action to AutoDJ, add hotkeys

### LINKED ISSUE #13361: Del key in sidebar doesn't trigger crate/playlist delete dialog anymore
### Bug Description

Due to
https://github.com/mixxxdj/mixxx/blame/abd9d11bf7ee09d259e568590a3f669547b53374/src/widget/wlibrarysidebar.cpp#L246-L248
the Delete code can't be reached anymore.

Probably a leftover from #12020 
And i think we overlooked -or I forgot it- that this does clear the AutoDJ queue without confirmation!
(which I consider a blocker btw. Unlikely it happens, but if it does you might screw your carefully curated AutoDJ playlist)

Removing this branch (and related methods like clear()) fixes the original issue.
But what about clearing AutoDJ queue from the sidebar?
1. rename `AutoDJFeature::clear()` -> `AutoDJFeature::deleteItem` and add a confirmation dialog?
2. add a respective sidebar menu action to make the feature discoverable?
3. remove it?

I'd vote for 1. + 2. Though, if we'd be strict with the string freeze we'd need to drop it and reintroduce it in 2.6 with the proper dialog and sidebar menu action.

Additionally, we may allow `deleteItem` for the AutoDJ crates.

### Version

2.5-alpha

### OS

_No response_

ronso0:
Fixes #13361 

1. restore Del hotkey (delete playlist / crate)
2. to not drop the 'clear' function I added a 'Clear Queue' menu action to the AutoDJ root view incl. a confirmatio dialog
3. show Del hotkey in AutoDJ -> Clear and 'Remove crate/playlist' actions

Alternatively, we pick only the fix for 2.5 and the new action goes to 2.6

### DIFF
--- a/src/library/autodj/autodjfeature.cpp
+++ b/src/library/autodj/autodjfeature.cpp
@@ -15,6 +15,7 @@
 #include "sources/soundsourceproxy.h"
 #include "track/track.h"
 #include "util/clipboard.h"
+#include "util/defs.h"
 #include "util/dnd.h"
 #include "widget/wlibrary.h"
 #include "widget/wlibrarysidebar.h"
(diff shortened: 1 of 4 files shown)


ronso0:
_ding dong_
Is really no one else is bothered by that regression?

We're slowly approaching the 2.5 release so wdyt about this?
> Alternatively, we pick only the fix for 2.5 and the new action goes to 2.6

daschuer (changes requested):
Oh, just realized that I had a pending minor review comment. Here is it.

m0dB:
Sorry for not being very responsive, I am travelling. LGTM but not merging as there is a pre-commit issue.

ronso0:
Thanks.

> LGTM but not merging as there is a pre-commit issue.

I guess that was from the tr code suggestions, fixed by merging that commit as fixup.
Ready to roll!

PR #13367, opened 2024-06-13
Changed WImageStore to use managed pointer types

JoergAtGithub:


daschuer (commented):
What is the issue you are trying to solve? 

Before WImageStore is a global weak reference holder whith a the single purpose to avoid duplicated images in the heap memory. 

If a image is no longer used it is evicted and it needs to be rebuilt when it's needed again. 

Now all images drawed during the whole lifetime of Mixxx are collected, in a kind of leaking memory. 

I suggest to keep WImage store as it is and seek for a better solution for your problem.

PR #13369, opened 2024-06-14
WPixmapStore: Change getPixmapNoCache to std::unique_ptr and further optimizations

JoergAtGithub:


src/skin/legacy/legacyskinparser.cpp
```diff
@@ -901,12 +901,12 @@ QWidget* LegacySkinParser::parseBackground(const QDomElement& node,
...
     QString filename = m_pContext->selectString(node, "Path");
-    QPixmap* background = WPixmapStore::getPixmapNoCache(
-        m_pContext->makeSkinPath(filename), m_pContext->getScaleFactor());
+    auto background = WPixmapStore::getPixmapNoCache(
+            m_pContext->makeSkinPath(filename), m_pContext->getScaleFactor());
 
     bg->move(0, 0);
     if (background != nullptr && !background->isNull()) {
-        bg->setPixmap(*background);
+        bg->setPixmap(*background.get());
```
Swiftb0y, line 909:
```suggestion
        // setPixmap makes a copy of the background
        // which can be disposed of afterwards.
        bg->setPixmap(*background);
```

src/widget/paintable.cpp
```diff
@@ -53,7 +53,12 @@ Paintable::Paintable(const PixmapSource& source, DrawMode mode, double scaleFact
         : m_drawMode(mode),
           m_source(source) {
     if (!source.isSVG()) {
-        m_pPixmap.reset(WPixmapStore::getPixmapNoCache(source.getPath(), scaleFactor));
+            auto pPixmap = WPixmapStore::getPixmapNoCache(source.getPath(), scaleFactor);
+            if (!pPixmap) {
+                qWarning() << "Failed to load pixmap from path:" << source.getPath();
+                return;
+            }
+            m_pPixmap.reset(pPixmap.release());
```
Swiftb0y, line 61: no need to overcomplicate things. `unique_ptr` does the right thing (potentially even faster than the manual reset+release).
```suggestion
            m_pPixmap = pPixmap;
```

src/widget/wpixmapstore.cpp
```diff
@@ -14,32 +14,30 @@ PaintablePointer WPixmapStore::getPaintable(const PixmapSource& source,
     if (source.isEmpty()) {
         return PaintablePointer();
     }
-    QString key = source.getId() + QString::number(mode) + QString::number(scaleFactor);
+    QString key = QString("%1%2%3").arg(source.getId(),
```
Swiftb0y, line 17:
```suggestion
    QString key = QStringLiteral("%1%2%3").arg(source.getId(),
```

Swiftb0y (commented):
couple nits. Thank you for the cleanup

src/widget/wpixmapstore.cpp
```diff
@@ -14,32 +14,30 @@ PaintablePointer WPixmapStore::getPaintable(const PixmapSource& source,
...
+std::unique_ptr<QPixmap> WPixmapStore::getPixmapNoCache(
         const QString& fileName,
         double scaleFactor) {
-    QPixmap* pPixmap = nullptr;
-    QImage* img = m_loader->getImage(fileName, scaleFactor);
-    pPixmap = new QPixmap();
-    pPixmap->convertFromImage(*img);
-    delete img;
-    return pPixmap;
+    auto pImage = m_loader->getImage(fileName, scaleFactor);
```
daschuer, line 36:
```suggestion
    auto* pImage = m_loader->getImage(fileName, scaleFactor);
```

src/widget/paintable.cpp
Swiftb0y: whoops. I didn't notice that m_pPixmap is Qts unique ptr and  `pPixmap` is a `std::unique_ptr`... I'm sorry for the bogus suggestion then. Feel free to revert/drop this suggestion or make them both `std::unique_ptr`, whatever you prefer.

src/widget/wpixmapstore.cpp
```diff
@@ -14,32 +14,30 @@ PaintablePointer WPixmapStore::getPaintable(const PixmapSource& source,
...
-    QImage* img = m_loader->getImage(fileName, scaleFactor);
-    pPixmap = new QPixmap();
-    pPixmap->convertFromImage(*img);
-    delete img;
-    return pPixmap;
+    auto* pImage = m_loader->getImage(fileName, scaleFactor);
+    if (!pImage || pImage->isNull()) {
+        return nullptr;
+    }
+    return std::make_unique<QPixmap>(QPixmap::fromImage(*pImage));
```
Swiftb0y, line 40:
```suggestion
    return std::unique_ptr(&QPixmap::fromImage(*pImage));
```
Otherwise `make_unique` will call the copy ctor with the temporary.

src/widget/wpixmapstore.cpp
```diff
@@ -14,32 +14,30 @@ PaintablePointer WPixmapStore::getPaintable(const PixmapSource& source,
...
-    if (pPaintable) {
-        return pPaintable;
+    auto it = m_paintableCache.find(key);
+    if (it != m_paintableCache.end()) {
+        return it.value();
     }
 
-    pPaintable = PaintablePointer(new Paintable(source, mode, scaleFactor));
-
+    PaintablePointer pPaintable(new Paintable(source, mode, scaleFactor));
```
Swiftb0y, line 27: this is a `QSharedPointer<Paintable>`, right?
```suggestion
    auto pPaintable = PaintablePointer::create(source, mode, scaleFactor);
```

Swiftb0y (commented):
last nits. thank you.

src/widget/wpixmapstore.cpp
Swiftb0y: whoops, blunder v2... 
Unless we eliminate the `unique_ptr` entirely, we need the copy to move it from the stack to the heap... We can think about removing it in a follow up PR (possibly replacing it with a std::optional instead).

JoergAtGithub: Reverted

src/util/opengltexture2d.cpp
```diff
@@ -20,7 +20,7 @@ void OpenGLTexture2D::setData(const QPixmap& pixmap) {
     setData(pixmap.toImage());
 };
 
-void OpenGLTexture2D::setData(const QSharedPointer<Paintable>& pPaintable) {
+void OpenGLTexture2D::setData(const std::shared_ptr<Paintable>& pPaintable) {
     if (pPaintable) {
         setData(pPaintable->toImage());
     }
```
Swiftb0y, line 26: this is not a suggestion for this PR, just a coding guideline in the future. If you don't need to transfer ownership, just take a plain ptr (if you need nullability) or even just a ref to the underlying type directly and leave the rest to the caller. Then this method works regardless how the pointer/type is managed. Does that make sense?
```suggestion
void OpenGLTexture2D::setData(const Paintable* pPaintable) {
    if (pPaintable) {
        setData(pPaintable->toImage());
    }
```

src/widget/wpixmapstore.h
```diff
@@ -1,35 +1,35 @@
...
 
 #include "skin/legacy/imgsource.h"
 #include "skin/legacy/pixmapsource.h"
 #include "widget/paintable.h"
 
-
-typedef QSharedPointer<Paintable> PaintablePointer;
-typedef QWeakPointer<Paintable> WeakPaintablePointer;
+typedef std::shared_ptr<Paintable> PaintablePointer;
+typedef std::weak_ptr<Paintable> WeakPaintablePointer;
```
Swiftb0y, line 18: nitpick, https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using
```suggestion
using PaintablePointer = std::shared_ptr<Paintable> ;
using WeakPaintablePointer = std::weak_ptr<Paintable> ;
```

Swiftb0y (approved):
LGTM. Thank you

src/util/opengltexture2d.cpp
Swiftb0y: In the pointer type it results in the caller doing a semi-pretty `unique_ptr::get()`, but in the reference case, its really ergonomic since you just need to deref, the ptr.

Swiftb0y (approved):
Thank you. Waiting for CI just in case.

JoergAtGithub:
@Swiftb0y This should be ready now!

Swiftb0y:
Yup, thanks for the extensive cleanup and the reminder.

PR #13370, opened 2024-06-14
use enum class for waveform overview type

ronso0:
Follow-up for #13273

### DIFF
--- a/src/preferences/dialog/dlgprefwaveform.cpp
+++ b/src/preferences/dialog/dlgprefwaveform.cpp
@@ -1,6 +1,8 @@
 #include "preferences/dialog/dlgprefwaveform.h"
 
-#include "control/controlobject.h"
+#include <QMetaEnum>
+
+#include "control/controlpushbutton.h"
 #include "library/dao/analysisdao.h"
 #include "library/library.h"
 #include "moc_dlgprefwaveform.cpp"
@@ -8,6 +10,12 @@
 #include "util/db/dbconnectionpooled.h"
 #include "waveform/renderers/waveformwidgetrenderer.h"
 #include "waveform/waveformwidgetfactory.h"
+#include "widget/woverview.h"
+
+namespace {
+const ConfigKey kOverviewTypeCfgKey(QStringLiteral("[Waveform]"),
+        QStringLiteral("WaveformOverviewType"));
+} // namespace
 
 DlgPrefWaveform::DlgPrefWaveform(
         QWidget* pParent,
@@ -19,13 +27,27 @@ DlgPrefWaveform::DlgPrefWaveform(
     setupUi(this);
 
     // Waveform overview init
-    waveformOverviewComboBox->addItem(tr("Filtered")); // "0"
-    waveformOverviewComboBox->addItem(tr("HSV")); // "1"
-    waveformOverviewComboBox->addItem(tr("RGB")); // "2"
-    m_pTypeControl = std::make_unique<ControlObject>(
-            ConfigKey(QStringLiteral("[Waveform]"),
-                    QStringLiteral("WaveformOverviewType")));
+    waveformOverviewComboBox->addItem(
+            tr("Filtered"), QVariant::fromValue(WOverview::Type::Filtered));
+    waveformOverviewComboBox->addItem(tr("HSV"), QVariant::fromValue(WOverview::Type::HSV));
+    waveformOverviewComboBox->addItem(tr("RGB"), QVariant::fromValue(WOverview::Type::RGB));
+    m_pTypeControl = std::make_unique<ControlPushButton>(kOverviewTypeCfgKey);
+    m_pTypeControl->setStates(QMetaEnum::fromType<WOverview::Type>().keyCount());
     m_pTypeControl->setReadOnly();
+    // Update the control with the config value
+    WOverview::Type overviewType =
+            m_pConfig->getValue<WOverview::Type>(kOverviewTypeCfgKey, WOverview::Type::RGB);
+    int cfgTypeIndex = waveformOverviewComboBox->findData(QVariant::fromValue(overviewType));
+    if (cfgTypeIndex == -1) {
+        // Invalid config value, set default type RGB and write it to config
+        waveformOverviewComboBox->setCurrentIndex(
+                waveformOverviewComboBox->findData(QVariant::fromValue(WOverview::Type::RGB)));
+        m_pConfig->setValue(kOverviewTypeCfgKey, cfgTypeIndex);
+    } else {
+        waveformOverviewComboBox->setCurrentIndex(cfgTypeIndex);
+    }
+    // Set the control used by WOverview
+    m_pTypeControl->forceSet(cfgTypeIndex);
 
     // Populate waveform options.
     WaveformWidgetFactory* factory = WaveformWidgetFactory::instance();
@@ -204,13 +226,14 @@ void DlgPrefWaveform::slotUpdate() {
                     factory->getUntilMarkAlign()));
     untilMarkTextPointSizeSpinBox->setValue(factory->getUntilMarkTextPointSize());
 
-    // By default we set RGB woverview = "2"
-    int overviewType = m_pConfig->getValue(
-            ConfigKey("[Waveform]","WaveformOverviewType"), 2);
-    if (overviewType != waveformOverviewComboBox->currentIndex()) {
-        waveformOverviewComboBox->setCurrentIndex(overviewType);
+    WOverview::Type cfgOverviewType =
+            m_pConfig->getValue<WOverview::Type>(kOverviewTypeCfgKey, WOverview::Type::RGB);
+    // Assumes the combobox index is in sync with the ControlPushButton
+    if (cfgOverviewType != waveformOverviewComboBox->currentData().value<WOverview::Type>()) {
+        int cfgOverviewTypeIndex =
+                waveformOverviewComboBox->findData(QVariant::fromValue(cfgOverviewType));
+        waveformOverviewComboBox->setCurrentIndex(cfgOverviewTypeIndex);
     }
-    slotSetWaveformOverviewType(overviewType);
 
     WaveformSettings waveformSettings(m_pConfig);
     enableWaveformCaching->setChecked(waveformSettings.waveformCachingEnabled());
@@ -220,10 +243,6 @@ void DlgPrefWaveform::slotUpdate() {
 }
 
 void DlgPrefWaveform::slotApply() {
-    ConfigValue overviewtype = ConfigValue(waveformOverviewComboBox->currentIndex());
-    if (overviewtype != m_pConfig->get(ConfigKey("[Waveform]", "WaveformOverviewType"))) {
-        m_pConfig->set(ConfigKey("[Waveform]", "WaveformOverviewType"), overviewtype);
-    }
     WaveformSettings waveformSettings(m_pConfig);
     waveformSettings.setWaveformCachingEnabled(enableWaveformCaching->isChecked());
     waveformSettings.setWaveformGenerationWithAnalysisEnabled(
@@ -252,7 +271,8 @@ void DlgPrefWaveform::slotResetToDefaults() {
     synchronizeZoomCheckBox->setChecked(true);
 
     // RGB overview.
-    waveformOverviewComboBox->setCurrentIndex(2);
+    waveformOverviewComboBox->setCurrentIndex(
+            waveformOverviewComboBox->findData(QVariant::fromValue(WOverview::Type::RGB)));
 
     // Don't normalize overview.
     normalizeOverviewCheckBox->setChecked(false);
@@ -303,9 +323,13 @@ void DlgPrefWaveform::updateEnableUntilMark() {
     requiresGLSLLabel->setVisible(!enabled);
 }
 
-void DlgPrefWaveform::slotSetWaveformOverviewType(int index) {
-    m_pConfig->set(ConfigKey("[Waveform]","WaveformOverviewType"), ConfigValue(index));
-    m_pTypeControl->forceSet(index);
+void DlgPrefWaveform::slotSetWaveformOverviewType() {
+    // Apply immediately
+    QVariant comboboxData = waveformOverviewComboBox->currentData();
+    DEBUG_ASSERT(comboboxData.canConvert<WOverview::Type>());
+    auto type = comboboxData.value<WOverview::Type>();
+    m_pConfig->setValue(kOverviewTypeCfgKey, type);
+    m_pTypeControl->forceSet(static_cast<double>(type));
 }
 
 void DlgPrefWaveform::slotSetDefaultZoom(int index) {
@@ -351,6 +375,8 @@ void DlgPrefWaveform::slotClearCachedWaveforms() {
 }
 
 void DlgPrefWaveform::slotSetBeatGridAlpha(int alpha) {
+    // TODO(xxx) For consistency set this in WaveformWidgetFactory like
+    // the other waveform controls.
     m_pConfig->setValue(ConfigKey("[Waveform]", "beatGridAlpha"), alpha);
     WaveformWidgetFactory::instance()->setDisplayBeatGridAlpha(alpha);
 }
--- a/src/widget/woverview.cpp
+++ b/src/widget/woverview.cpp
@@ -56,7 +56,16 @@ WOverview::WOverview(
           m_analyzerProgress(kAnalyzerProgressUnknown),
           m_trackLoaded(false),
           m_pHoveredMark(nullptr),
-          m_scaleFactor(1.0) {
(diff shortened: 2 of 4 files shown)


src/widget/woverview.cpp
```diff
@@ -80,8 +81,8 @@ WOverview::WOverview(
             QStringLiteral("[Waveform]"),
```
ronso0, line 81: I really don't understand those `ControlFlag::NoAssertIfMissing` flags above. Some controls are read when painting and there is no null check. Even if default controls return 0 there will a division by 0 at one place.
Should we remove those flags?

src/widget/woverview.h
```diff
@@ -20,6 +20,13 @@ class PlayerManager;
 class QDomNode;
 class SkinContext;
 
+enum class OverviewType {
+    Filtered,
+    HSV,
+    RGB,
+    Count,
+};
```
Swiftb0y, line 28: can you move this into WOverview and make it a `Q_ENUM`?

Then you can remove the `Count` member (anti-pattern) and instead use `QMetaEnum::keyCount`. It will also be simpler in the QComboBox because the enum can be implicitly converted to QVariant.

src/widget/woverview.cpp
```diff
@@ -400,11 +401,10 @@ void WOverview::onPassthroughChange(double v) {
...
 }
 
-void WOverview::slotTypeChanged(double v) {
-    int type = static_cast<int>(v);
-    VERIFY_OR_DEBUG_ASSERT(type >= 0 && type <= 2) {
-        type = 2;
-    }
+void WOverview::slotTypeControlChanged(double v) {
+    // No need to verify that v is in enum range to prevent UB because
+    // the pushbutton has [Overviewtype::Count] states.
```
Swiftb0y, line 406: lets still add a `DEBUG_ASSERT` just in case.

src/widget/woverview.cpp
ronso0: yeah, so I'll remove those.

src/widget/woverview.h
ronso0: Oh, nice! Thanks.

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -303,9 +320,11 @@ void DlgPrefWaveform::updateEnableUntilMark() {
...
 }
 
-void DlgPrefWaveform::slotSetWaveformOverviewType(int index) {
-    m_pConfig->set(ConfigKey("[Waveform]","WaveformOverviewType"), ConfigValue(index));
-    m_pTypeControl->forceSet(index);
+void DlgPrefWaveform::slotSetWaveformOverviewType() {
+    // Apply immediately
+    int type = waveformOverviewComboBox->currentData().toInt();
+    m_pConfig->setValue(kOverviewTypeCfgKey, type);
+    m_pTypeControl->forceSet(type);
```
Swiftb0y, line 327:
```suggestion
    DEBUG_ASSERT(waveformOverviewComboBox->currentData().canConvert<WOverview::Type>());
    auto type = waveformOverviewComboBox->currentData().value<WOverview::Type>();
    m_pConfig->setValue(kOverviewTypeCfgKey, type);
    m_pTypeControl->forceSet(static_cast<double>(type));
```

src/widget/woverview.cpp
```diff
@@ -400,11 +401,10 @@ void WOverview::onPassthroughChange(double v) {
...
 
-void WOverview::slotTypeChanged(double v) {
-    int type = static_cast<int>(v);
-    VERIFY_OR_DEBUG_ASSERT(type >= 0 && type <= 2) {
-        type = 2;
-    }
+void WOverview::slotTypeControlChanged(double v) {
+    // Assert that v is in enum range to prevent UB.
+    DEBUG_ASSERT(v >= 0 && v < QMetaEnum::fromType<Type>().keyCount());
+    Type type = static_cast<Type>(static_cast<int>(v));
```
Swiftb0y, line 407: is the double cast really required? A single `static_cast<Type>(v)` seemed sufficient from my testing.

src/widget/woverview.h
```diff
@@ -32,6 +32,13 @@ class WOverview : public WWidget, public TrackDropTarget {
     void setup(const QDomNode& node, const SkinContext& context);
     virtual void initWithTrack(TrackPointer pTrack);
 
+    enum Type {
```
Swiftb0y, line 35: pretty sure Q_ENUMs work on `enum class`es. Sorry, should've specified that we should still keep the enum class.

Swiftb0y (commented):
thank you, couple more nits

src/widget/woverview.cpp
ronso0: yes, the simple cast works, just did that out of anticipatory obedience to clang-tidy. will revert it.

Swiftb0y: don't think clang tidy will complain, its tuned with very few rules. I don't even there is a rule (nor a justification) for requiring two casts here.

src/widget/woverview.cpp
```diff
@@ -58,30 +58,31 @@ WOverview::WOverview(
...
-            m_group, QStringLiteral("track_samples"), this);
-    m_playpositionControl = make_parented<ControlProxy>(
-            m_group, QStringLiteral("playposition"), this, ControlFlag::NoAssertIfMissing);
+    // TODO(xxx) Use PollingControlProxy where possible
+    m_trackSampleRateControl = PollingControlProxy(
+            m_group, QStringLiteral("track_samplerate"));
+    m_trackSamplesControl = PollingControlProxy(
+            m_group, QStringLiteral("track_samples"));
+    m_playpositionControl = PollingControlProxy(
+            m_group, QStringLiteral("playposition"));
```
ronso0, line 74: Constructing PollingControlProxy like this throws a lot of
```
critical [Main] DEBUG ASSERT: "flags & ControlFlag::AllowMissingOrInvalid" in function PollingControlProxy::PollingControlProxy(const ConfigKey&, ControlFlags) at ./src/control/pollingcontrolproxy.h:24
warning [Main] ControlDoublePrivate::getControl returning nullptr for invalid ConfigKey "" , ""
critical [Main] DEBUG ASSERT: "!"Unexpected invalid key"" in function static QSharedPointer<ControlDoublePrivate> ControlDoublePrivate::getControl(const ConfigKey&, ControlFlags, ControlObject*, bool, bool, bool, double) at ./src/control/control.cpp:147
```
and I have no clue why.

src/widget/woverview.cpp
ronso0: I just found this note, not sure if it still applies:
https://github.com/mixxxdj/mixxx/blob/796d056a584b6667278355d8a3ba840e9198e78e/src/engine/controls/cuecontrol.cpp#L2734-L2736

Swiftb0y: just checked, it does (compiler explorer with msvc v19.0 (Wine)), even though [it shouldn't](https://developercommunity.visualstudio.com/t/the-compiler-emits-c2440-when-trying-to-static-cas/793475).

Swiftb0y: Sorry for the bogus comment then.

src/widget/woverview.h
ronso0: That would be an `enum class` inside a  `class`?
Tried that but pparently it's not accessible elsewhere even tough it's public.

ronso0: _ping_ I need some assistance here.
What's the benefit of having the enum (class) inside the WOverview class? We need it at least in one other unit (prefs), and according to my test that doesn't work unfortunately.

Or, the other way around: what would be the downside of it being a enum class declared in woverview.h but **not** in the WOverview class? With the distinct name `OverviewType` there's no danger of global name clashes IMO, if that's what the `WOverview` prefix  is supposed to prevent.

Swiftb0y: > ping I need some assistance here.

Sorry. I overlooked your previous comment.

> What's the benefit of having the enum (class) inside the WOverview class? 

unscoped enums (plain `enum`) always implicitly convert to their underlying type. In the majority of cases (including here IMO), that is undesired. 

> We need it at least in one other unit (prefs), and according to my test that doesn't work unfortunately.

[I can't reproduce](
https://compiler-explorer.com/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:106,endLineNumber:9,positionColumn:106,positionLineNumber:9,selectionStartColumn:106,selectionStartLineNumber:9,startColumn:106,startLineNumber:9),source:'class+A+%7B%0A++++public:%0A++++//+declare+in+public+section%0A++++enum+class+Enum+%7B%0A++++++++member1,%0A++++++++member2%0A++++%7D%3B%0A++++//+using+enum+to+%22inline%22+enum+members+into+class%0A++++//+needs+C%2B%2B20%3B+only+benefit+is+having+to+type+a+little+less+because+you+can+omit+%60::Type%60+(see+below)%0A++++using+enum+Enum%3B%0A%7D%3B%0A%0Aclass+B+%7B%0A++++//+use+value%0A++++B()+:+member1(A::Enum::member1)+%7B%7D%0A++++//+use+type%0A++++A::Enum+member1%3B%0A++++//+use+%22inlined%22+member%0A++++A::Enum+member2+%3D+A::member2%3B%0A%7D%3B'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g141,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-Wall+-Wextra+--std%3Dc%2B%2B20',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+14.1+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+14.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+14.1+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4). What code did you test? 

> what would be the downside of it being a enum class declared in woverview.h but not in the WOverview class?

you would have to put it into a namespace that then gets annotated with `Q_NAMESPACE` and make it a Q_ENUM_NS (works, but a little more complicated than just putting it right into the QObject when its already there).

> With the distinct name OverviewType there's no danger of global name clashes IMO, if that's what the WOverview prefix is supposed to prevent.

I'm not too worried about nameclashes. Its more a technical limitation because in order to make an enum a Q_ENUM, Qt's MOC needs to know about it. So it either needs to be inclosed in a Q_NAMESPACE or in a QObject. Since Q_NAMESPACEs work slightly differently from real c++ namespaces, leveraging the QObject is probably less confusing.

Swiftb0y: Ah yeah. I guess thats an explicitly choice from Qt for some reason (mby legacy, idk). The fromValue is fine then IMO. Sorry for dragging this discussion out again. Should've done more research beforehand.

Swiftb0y (approved):
LGTM. There is still more cleanup possible in future PRs, but this is already an improvement. Thank you.

src/widget/woverview.h
ronso0: Never mind..

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -19,13 +27,26 @@ DlgPrefWaveform::DlgPrefWaveform(
...
 
     // Waveform overview init
-    waveformOverviewComboBox->addItem(tr("Filtered")); // "0"
-    waveformOverviewComboBox->addItem(tr("HSV")); // "1"
-    waveformOverviewComboBox->addItem(tr("RGB")); // "2"
-    m_pTypeControl = std::make_unique<ControlObject>(
-            ConfigKey(QStringLiteral("[Waveform]"),
-                    QStringLiteral("WaveformOverviewType")));
+    waveformOverviewComboBox->addItem(
+            tr("Filtered"), QVariant::fromValue(WOverview::Type::Filtered));
```
acolombier, line 31: Is `QVariant::fromValue` needed here? I'm not sure, but the `Q_ENUM` declaration of `WOverview::Type` might declare it as a meta type already. Otherwise, perhaps you can use `Q_DECLARE_METATYPE` explicitly?

Swiftb0y:
@daschuer do you want to take a look?
also @ronso0 do you have nerves to test out https://github.com/mixxxdj/mixxx/pull/13370/files#r1641522940?

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -19,13 +27,26 @@ DlgPrefWaveform::DlgPrefWaveform(
...
+    m_pTypeControl = std::make_unique<ControlPushButton>(kOverviewTypeCfgKey);
+    m_pTypeControl->setStates(QMetaEnum::fromType<WOverview::Type>().keyCount());
     m_pTypeControl->setReadOnly();
+    // Update the control with the config value
+    WOverview::Type overviewType =
+            m_pConfig->getValue<WOverview::Type>(kOverviewTypeCfgKey, WOverview::Type::RGB);
+    int cfgTypeIndex = waveformOverviewComboBox->findData(QVariant::fromValue(overviewType));
+    if (cfgTypeIndex == -1) {
+        // Invalid config value, set default type RGB and write it to config
+        waveformOverviewComboBox->setCurrentIndex(2);
```
acolombier, line 43: Could be worth using `WOverview::Type::RGB` instead of `2` not to have it hardcoded here

acolombier (approved):

src/preferences/dialog/dlgprefwaveform.cpp
ronso0: Unfortunately, neither Q_ENUM nor Q_DECLARE_METATYPE do the trick here after moving the enum class inside the WOverview class, see the convo above https://github.com/mixxxdj/mixxx/pull/13370#discussion_r1641529210 :man_shrugging:

src/preferences/dialog/dlgprefwaveform.cpp
ronso0: Oh, indeed, I overlooked that

src/preferences/dialog/dlgprefwaveform.cpp
acolombier: Oops my bad, I didn't see that conversation!

src/preferences/dialog/dlgprefwaveform.cpp
Swiftb0y: I actually don't agree with this change. This conflates the value of the enum members with the order in which they are inserted. Since the hardcoded index in here is very close to the order in which they are inserted, its fine like this IMO.

Swiftb0y:
yes, only unresolved comment is https://github.com/mixxxdj/mixxx/pull/13370#discussion_r1641522940

src/preferences/dialog/dlgprefwaveform.cpp
ronso0: I already switched to the enum, but apparently that got lost somehow. Will fix again

ronso0: Oh, I confused it with resetToDefaults
fixed.

Swiftb0y:
> yes, the last remaining unresolved comment is [#13370 (comment)](https://github.com/mixxxdj/mixxx/pull/13370#discussion_r1641522940)

@ronso0 friendly ping ;)

ronso0:
Isn't that fixed by the last fixup?
https://github.com/mixxxdj/mixxx/compare/91d9f6c724edf0540d8d595516f9cb8247ec1652..e8eea9fe25e505f4ddbefaa93741711d1224efda

Swiftb0y:
I guess the link anchor doesn't work. I'm referring to this request :sweat_smile: :
image

ronso0:
Oh-oh-kay, got it now :grimacing:

Swiftb0y (approved):
Thank you. Waiting for CI.

JoergAtGithub:
Lets merge this! Thank you!

Swiftb0y:
sorry, forgot, thank you.

PR #13381, opened 2024-06-16
Waveforms: Disable textured waveforms when using OpenGL ES

### LINKED ISSUE #13380: Compile errors in `WaveformRendererTextured` when building for iOS (OpenGL ES 3.0)
### Bug Description

Parts of the new waveform renderer unfortunately no longer build for iOS (and possibly other platforms using OpenGL ES):

Screenshot 2024-06-16 at 16 45 51

We might have to investigate which parts to cond-compile out when using GL ES (e.g. testing for `QT_OPENGL_ES_2`).

### Version

https://github.com/fwcd/mixxx/commit/ea5e98073c15efb4c60f611fb67af457bb5be8ac (based on c368119262913411f94766a3e59735678bd72832)

### OS

macOS 14.5

fwcd:
Fixes #13380 (potential future refactoring moved to #13385)

The textured waveform renderer seems to use some OpenGL APIs that are unavailable in GL ES (e.g. on iOS), therefore this PR cond-compiles out `WaveformRendererTextured` when `QT_OPENGL_ES_2` is set.

### DIFF
--- a/src/waveform/renderers/allshader/waveformrenderertextured.cpp
+++ b/src/waveform/renderers/allshader/waveformrenderertextured.cpp
@@ -1,5 +1,7 @@
 #include "waveform/renderers/allshader/waveformrenderertextured.h"
 
+#ifndef QT_OPENGL_ES_2
+
 #include <QOpenGLFramebufferObject>
 #include <QOpenGLShaderProgram>
 
@@ -531,3 +533,5 @@ void WaveformRendererTextured::paintGL() {
 }
 
 } // namespace allshader
+
+#endif // QT_OPENGL_ES_2
--- a/src/waveform/renderers/allshader/waveformrenderertextured.h
+++ b/src/waveform/renderers/allshader/waveformrenderertextured.h
@@ -1,5 +1,7 @@
 #pragma once
 
+#ifndef QT_OPENGL_ES_2
+
 #include "shaders/rgbshader.h"
 #include "track/track_decl.h"
 #include "util/class.h"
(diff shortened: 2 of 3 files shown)


fwcd:
Perhaps it would make sense to disable the "High Details" checkbox when using GL ES? Given that it will have no effect

IMG_0904

Thoughts welcome.

daschuer:
Disabling or even hiding the box is a good idea as a band aid. But I would like to keep the original bug open until we have figured out the root cause. I have edited the bug PR description accordingly.

daschuer:
Please rebase to 2.5, since this is fixing 2.5-beta code.

fwcd:
I'll rebase it once I'm home later.

> But I would like to keep the original bug open until we have figured out the root cause. 

The root cause is that the textured waveforms are implemented in terms of OpenGL APIs that are unavailable in OpenGL ES. So if we want to keep the original issue open, we might want to rephrase it to something along the lines of "Reimplement textured waveforms in OpenGL ES subset" or similar, since this specific issue is fixed.

fwcd:
I don't think it makes sense to target 2.5 here currently, since it is based on the refactorings introduced in #13220 (`addWaveformSignalRenderer` etc.) that are only in `main`. It might be possible to backport, but in doing so we'll get merge conflicts and I'd base iOS support on `main` anyway (given that it's highly experimental).

acolombier (approved):
LGTM! Thanks for the fix

daschuer (approved):
LGTM, thank you.

Issue #12565, opened 2024-01-14
Library: visual glitch on removing item from Crate, cover art stays, rows move up but ~16 above


mxmilkiib:
### Bug Description

I'm seeing this thing that on removing an item , there's a kind of glitch, the cover art for that row doesn't disappear, it steys the same one is removed like 16 items up, and other tracks' text all jump up one row, but the cover art stays the same height.

### Version

main

### OS

Arch

ronso0:
Could you please provide a short screencast demonstrating the issue?

mxmilkiib:
Took a few attempts.. finally got a clean one. 6s.
[mixxx_library_glitch-2024-01-15_01.59.19.webm](https://github.com/mixxxdj/mixxx/assets/108225/ccf82d1b-9ac7-41e5-bae9-5b64ecb65010)

mxmilkiib:
Another. Accidendally amended it to the first, 7 secs on, ffmpeg editing keeps cutting it wrong and I'm tired so full clip
[mixxx_library_glitch2-2024-01-15_02.09.07.webm](https://github.com/mixxxdj/mixxx/assets/108225/d3cf6330-6686-4271-bbe0-aa9d1c022626)

ronso0:
It's very hard to spot it.
By which column are the track sorted when you remove a track?

Can you reproduce when sort, then remove  a track without doing anything in between that might affect the track metadata?
(e.g. playing a track chnages the play count and the 'last played' value)

mxmilkiib:
Don't play the videos, scrub them so you control what is being shown. Open on mpv and use . and , to move by frame.

I don't know how to reproduce yet. I'm focusing on curating and editing metadata, but I'll see if I can remember. I clicked on the side panel then tab into the library, that might count for something to help reproduced. Probably because I'm on the current Qt.

ronso0:
First clip:
* two rows are removed: the one selected + the one above
* though play counter is only decreased by one
= something happens with the sorting

Same in the second clip:
* one row removed play counter decreased by one)
* tracks in another region are resorted
* covers < > tracks still identical

Currenty you have tracks sorted by key.
Mixxx supports multi-column sorting, so obviously the previous sort columns is relevant here.
A possible explanantion is you previously sorted by Preview (= shuffle, hidden feature).

Please verify sort order is not touched when you sort by two regular properties (columns), then remove a track.

mxmilkiib:
[...]

PR #13390, opened 2024-06-19
feat: static color coding for key column

### LINKED ISSUE #7810: Color code key column in library
Reported by: **[naught101](https://github.com/naught101)**
Date: 2015-01-16T01:35:05Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1411476](https://bugs.launchpad.net/bugs/1411476)
Tags: library

---

It would be great if Mixxx colour coded the key information, according to the Camelot Wheel/circle of fifths (http://www.mixedinkey.com/Book/How-to-Use-Harmonic-Mixing-2), or some similar notation, where similar colours indicate that keys are harmonious.

danferns:
Hi! I just got the core of this feature working and I'm seeking feedback.

A small rectangle is drawn next to the Key label in the Library. The color of the rectangle is based on the Key of the track.

image

---

I'm concerned about the method used for finding the track's key. Right now, I'm using `KeyUtils::guessKeyFromText` on the Key string, but I don't think that'll work if the user sets custom Key Notation. 

I want to use the value in [`ColumnCache::COLUMN_LIBRARYTABLE_KEY_ID`](https://github.com/danferns/mixxx/blob/7931a0756cd8fb00422806d685c7029530864f14/src/library/basetracktablemodel.cpp#L888). But I haven't been able to figure out how to access this column from the `QModelIndex& index` that is passed into `KeyDelegate::paintItem`. I'm not sure if the column is even available.

Is there a place where I can see how the index is structured?

---

Once complete, this will resolve #7810.

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -904,10 +904,6 @@ QVariant BaseTrackTableModel::roleValue(
                 // Otherwise, just use the column value as is
                 return std::move(rawValue);
             }
-            // Convert or clear invalid values
-            VERIFY_OR_DEBUG_ASSERT(keyCodeValue.canConvert<int>()) {
-                return QVariant();
-            }
             bool ok;
             const auto keyCode = keyCodeValue.toInt(&ok);
             VERIFY_OR_DEBUG_ASSERT(ok) {
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -886,8 +886,7 @@ QVariant BaseTrackTableModel::roleValue(
         case ColumnCache::COLUMN_LIBRARYTABLE_KEY: {
             // If we know the semantic key via the LIBRARYTABLE_KEY_ID
             // column (as opposed to the string representation of the key
-            // currently stored in the DB) then lookup the key and render it
-            // using the user's selected notation.
+            // currently stored in the DB) then lookup the key and return it
             const QVariant keyCodeValue = rawSiblingValue(
                     index,
                     ColumnCache::COLUMN_LIBRARYTABLE_KEY_ID);
@@ -908,8 +907,7 @@ QVariant BaseTrackTableModel::roleValue(
             if (key == mixxx::track::io::key::INVALID) {
                 return QVariant();
             }
-            // Render the key with the user-provided notation
-            return KeyUtils::keyToString(key);
+            return QVariant::fromValue(key);
         }
         case ColumnCache::COLUMN_LIBRARYTABLE_REPLAYGAIN: {
             if (rawValue.isNull()) {
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -20,6 +20,7 @@
 #include "mixer/playerinfo.h"
 #include "mixer/playermanager.h"
 #include "moc_basetracktablemodel.cpp"
+#include "track/keyutils.h"
 #include "track/track.h"
 #include "util/assert.h"
 #include "util/clipboard.h"
@@ -100,9 +101,11 @@ QSqlDatabase cloneDatabase(
 constexpr int BaseTrackTableModel::kBpmColumnPrecisionDefault;
 constexpr int BaseTrackTableModel::kBpmColumnPrecisionMinimum;
 constexpr int BaseTrackTableModel::kBpmColumnPrecisionMaximum;
+constexpr bool BaseTrackTableModel::kKeyColorsEnabledDefault;
 
 int BaseTrackTableModel::s_bpmColumnPrecision =
         kBpmColumnPrecisionDefault;
+bool BaseTrackTableModel::s_keyColorsEnabled = kKeyColorsEnabledDefault;
 
 // static
 void BaseTrackTableModel::setBpmColumnPrecision(int precision) {
@@ -113,6 +116,12 @@ void BaseTrackTableModel::setBpmColumnPrecision(int precision) {
         precision = BaseTrackTableModel::kBpmColumnPrecisionMaximum;
     }
     s_bpmColumnPrecision = precision;
+}
+
+// static
+void BaseTrackTableModel::setKeyColorsEnabled(bool keyColorsEnabled) {
+    // todo: need to refresh the key column when this setting changes
+    s_keyColorsEnabled = keyColorsEnabled;
 }
 
 bool BaseTrackTableModel::s_bApplyPlayedTrackColor =
@@ -579,7 +588,8 @@ QVariant BaseTrackTableModel::data(
             role != Qt::CheckStateRole &&
             role != Qt::ToolTipRole &&
             role != kDataExportRole &&
-            role != Qt::TextAlignmentRole) {
+            role != Qt::TextAlignmentRole &&
+            role != Qt::UserRole) {
         return QVariant();
     }
 
@@ -1013,6 +1023,18 @@ QVariant BaseTrackTableModel::roleValue(
         }
         default:
             return QVariant(); // default AlignLeft for all other columns
+        }
+    }
+    case Qt::UserRole: {
+        switch (field) {
+        case ColumnCache::COLUMN_LIBRARYTABLE_KEY: {
+            // return whether or not to display the color rectangle
+            return QVariant::fromValue(s_keyColorsEnabled);
+        }
+        default: {
+            DEBUG_ASSERT(!"unexpected field for UserRole");
+            break;
+        }
         }
     }
     default:
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -9,6 +9,7 @@
 #include "library/tabledelegates/bpmdelegate.h"
 #include "library/tabledelegates/colordelegate.h"
 #include "library/tabledelegates/coverartdelegate.h"
+#include "library/tabledelegates/keydelegate.h"
 #include "library/tabledelegates/locationdelegate.h"
 #include "library/tabledelegates/multilineeditdelegate.h"
 #include "library/tabledelegates/playcountdelegate.h"
@@ -121,7 +122,7 @@ void BaseTrackTableModel::setApplyPlayedTrackColor(bool apply) {
     s_bApplyPlayedTrackColor = apply;
 }
 
-//static
+// static
 QStringList BaseTrackTableModel::defaultTableColumns() {
     return kDefaultTableColumns;
 }
@@ -506,6 +507,8 @@ QAbstractItemDelegate* BaseTrackTableModel::delegateForColumn(
                 this,
                 &BaseTrackTableModel::slotRefreshCoverRows);
         return pCoverArtDelegate;
+    } else if (index == fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_KEY)) {
+        return new KeyDelegate(pTableView);
     }
     return nullptr;
 }
--- a/src/library/tabledelegates/keydelegate.cpp
+++ b/src/library/tabledelegates/keydelegate.cpp
@@ -15,47 +15,45 @@ void KeyDelegate::paintItem(
 
     mixxx::track::io::key::ChromaticKey key =
             index.data().value<mixxx::track::io::key::ChromaticKey>();
-    // Get the key text with the user-provided notation
-    bool keyColorsEnabled = index.data(Qt::UserRole).value<bool>();
-    int rectWidth = 0; // only leave space for rect if it is rendered
-    if (keyColorsEnabled) {
-        const QColor keyColor = KeyUtils::keyToColor(key);
-        rectWidth = 8; // 4px width + 4px right padding
-        if (keyColor.isValid()) {
-            // Draw the colored rectangle next to the key label
-            painter->fillRect(
-                    option.rect.x(),
-                    option.rect.y() + 2,
-                    4, // width
-                    option.rect.height() - 4,
-                    keyColor);
+    if (key != mixxx::track::io::key::INVALID) {
+        // Display the key colors if enabled
+        bool keyColorsEnabled = index.data(Qt::UserRole).value<bool>();
+        if (keyColorsEnabled) {
+            const QColor keyColor = KeyUtils::keyToColor(key);
+            if (keyColor.isValid()) {
+                // Draw the colored rectangle next to the key label
+                painter->fillRect(
+                        option.rect.x(),
+                        option.rect.y() + 2,
+                        4, // width
+                        option.rect.height() - 4,
+                        keyColor);
+            }
         }
+
+        // Display the key text with the user-provided notation
+        int rectWidth = keyColorsEnabled ? 8 : 0; //  4px width + 4px right padding
+        const QString keyText = KeyUtils::keyToString(key);
+        QString elidedText = option.fontMetrics.elidedText(
+                keyText,
+                Qt::ElideLeft,
+                columnWidth(index) - rectWidth);
+
+        if (option.state & QStyle::State_Selected) {
+            // This uses selection-color from stylesheet for the text pen:
+            // #LibraryContainer QTableView {
+            //   selection-color: #fff;
+            // }
+            painter->setPen(QPen(option.palette.highlightedText().color()));
+        }
+
+        painter->drawText(option.rect.x() + rectWidth,
+                option.rect.y(),
+                option.rect.width() - rectWidth,
+                option.rect.height(),
+                Qt::AlignVCenter,
+                elidedText);
     }
-
-    QString keyText = KeyUtils::keyToString(key);
-    if (QString("INVALID").compare(keyText) == 0) { // if the key is invalid, display empty string
-        keyText.clear();
-    }
-
-    QString elidedText = option.fontMetrics.elidedText(
-            keyText,
-            Qt::ElideLeft,
-            columnWidth(index) - rectWidth);
-
-    if (option.state & QStyle::State_Selected) {
-        // This uses selection-color from stylesheet for the text pen:
-        // #LibraryContainer QTableView {
-        //   selection-color: #fff;
-        // }
-        painter->setPen(QPen(option.palette.highlightedText().color()));
-    }
-
-    painter->drawText(option.rect.x() + rectWidth,
-            option.rect.y(),
-            option.rect.width() - rectWidth,
-            option.rect.height(),
-            Qt::AlignVCenter,
-            elidedText);
 
     // Draw a border if the key cell has focus
     if (option.state & QStyle::State_HasFocus) {
--- a/src/library/tabledelegates/keydelegate.cpp
+++ b/src/library/tabledelegates/keydelegate.cpp
@@ -19,31 +19,41 @@ void KeyDelegate::paintItem(
         QPainter* painter,
         const QStyleOptionViewItem& option,
         const QModelIndex& index) const {
+    paintItemBackground(painter, option, index);
+
     QString keyText = index.data().toString();
     mixxx::track::io::key::ChromaticKey key = KeyUtils::guessKeyFromText(keyText);
     int openKeyNumber = KeyUtils::keyToOpenKeyNumber(key);
 
     if (openKeyNumber != 0) {
-        // Open Key numbers start from 1
-        const auto color =
+        const auto rgbColor =
                 mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette.at(
-                        openKeyNumber - 1);
+                        openKeyNumber - 1); // Open Key numbers start from 1
+        const auto qcolor = mixxx::RgbColor::toQColor(rgbColor);
 
-        // this code is from the ColorDelegate, using it temporarily for testing.
-        if (color) {
-            painter->fillRect(option.rect, mixxx::RgbColor::toQColor(color));
-        } else {
-            // Filter out track color that is hidden
-            if (option.state & QStyle::State_Selected) {
-                painter->fillRect(option.rect, option.palette.highlight());
-            }
-        }
-
-        // Draw a border if the color cell has focus
-        if (option.state & QStyle::State_HasFocus) {
-            drawBorder(painter, m_focusBorderColor, option.rect);
-        }
+        // Draw the colored rectangle next to the key label
+        painter->fillRect(
+                option.rect.x() + 2,
+                option.rect.y() + 2,
+                4, // width
+                option.rect.height() - 4,
+                qcolor);
     }
 
-    // QStyledItemDelegate::paint(painter, option, index);
+    QString elidedText = option.fontMetrics.elidedText(
+            index.data().toString(),
+            Qt::ElideLeft,
+            columnWidth(index) - 8); // Subtracting total width of the rectangle
+
+    painter->drawText(option.rect.x() + 8,
+            option.rect.y(),
+            option.rect.width() - 8,
+            option.rect.height(),
+            Qt::AlignVCenter,
+            elidedText);
+
+    // Draw a border if the key cell has focus
+    if (option.state & QStyle::State_HasFocus) {
+        drawBorder(painter, m_focusBorderColor, option.rect);
+    }
 }
--- a/src/library/tabledelegates/keydelegate.cpp
+++ b/src/library/tabledelegates/keydelegate.cpp
@@ -1,7 +1,5 @@
 #include "library/tabledelegates/keydelegate.h"
 
-#include <qnamespace.h>
-
 #include <QPainter>
 #include <QStyle>
 #include <QTableView>
@@ -10,10 +8,6 @@
 #include "track/keyutils.h"
 #include "util/color/predefinedcolorpalettes.h"
 #include "util/color/rgbcolor.h"
-
-KeyDelegate::KeyDelegate(QTableView* pTableView)
-        : TableItemDelegate(pTableView) {
-}
 
 void KeyDelegate::paintItem(
         QPainter* painter,
--- a/src/library/tabledelegates/keydelegate.cpp
+++ b/src/library/tabledelegates/keydelegate.cpp
@@ -41,6 +41,14 @@ void KeyDelegate::paintItem(
             Qt::ElideLeft,
             columnWidth(index) - rectWidth);
 
+    if (option.state & QStyle::State_Selected) {
+        // This uses selection-color from stylesheet for the text pen:
+        // #LibraryContainer QTableView {
+        //   selection-color: #fff;
+        // }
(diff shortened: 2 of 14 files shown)


src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,61 @@
+#include "library/tabledelegates/keydelegate.h"
+
+#include <qnamespace.h>
```
Swiftb0y, line 3: weird include, remove?
```suggestion
```

src/library/tabledelegates/keydelegate.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "library/tabledelegates/tableitemdelegate.h"
+
```
Swiftb0y, line 4: also `#include <QObject>`.

Swiftb0y:
> Once complete, this will resolve https://github.com/mixxxdj/mixxx/issues/7810.

I don't think so. That issue requests to color the key depending on how compatible it is to the currently playing track.

src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <QTableView>
+
+#include "moc_keydelegate.cpp"
+#include "track/keyutils.h"
+#include "util/color/predefinedcolorpalettes.h"
+#include "util/color/rgbcolor.h"
+
+KeyDelegate::KeyDelegate(QTableView* pTableView)
+        : TableItemDelegate(pTableView) {
+}
```
Swiftb0y, line 16: move into header please.

src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,61 @@
...
+    paintItemBackground(painter, option, index);
+
+    QString keyText = index.data().toString();
+    mixxx::track::io::key::ChromaticKey key = KeyUtils::guessKeyFromText(keyText);
+    int openKeyNumber = KeyUtils::keyToOpenKeyNumber(key);
+
+    if (openKeyNumber != 0) {
+        const auto rgbColor =
+                mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette.at(
+                        openKeyNumber - 1); // Open Key numbers start from 1
```
Swiftb0y, line 31: check bounds as well.
```suggestion
        const auto& palette = mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette;
        DEBUG_ASSERT(openKeyNumber <= palette.length());
        const auto rgbColor =
                palette.at(
                        openKeyNumber - 1); // Open Key numbers start from 1
```

src/util/color/predefinedcolorpalettes.cpp
```diff
@@ -144,6 +144,21 @@ constexpr mixxx::RgbColor kVirtualDJTrackColorBlue(0x0000FF);
...
+constexpr mixxx::RgbColor kMixxxKeyColor4(0xFED600);
+constexpr mixxx::RgbColor kMixxxKeyColor5(0x99FE00);
+constexpr mixxx::RgbColor kMixxxKeyColor6(0x42FE3E);
+constexpr mixxx::RgbColor kMixxxKeyColor7(0x0AD58F);
+constexpr mixxx::RgbColor kMixxxKeyColor8(0x0AE7E7);
+constexpr mixxx::RgbColor kMixxxKeyColor9(0x04C9FE);
+constexpr mixxx::RgbColor kMixxxKeyColor10(0x3D8AFD);
+constexpr mixxx::RgbColor kMixxxKeyColor11(0xAC64FE);
+constexpr mixxx::RgbColor kMixxxKeyColor12(0xFD3FEA);
+
```
Swiftb0y, line 161: as a follow up, it would be nice if the palette was customizable via the `Color` Preferences. similar to track colors.

Swiftb0y:
> I'm concerned about the method used for finding the track's key. Right now, I'm using KeyUtils::guessKeyFromText on the Key string, but I don't think that'll work if the user sets custom Key Notation.

Probably not. You should store the underlying `mixxx::track::io::key::ChromaticKey` in the cell when its text gets set. You can use a custom [item role](https://doc.qt.io/qt-6/model-view-programming.html#item-roles) for that (I think, I'm not much into Qt MV programming tbh).

danferns:
Thank you so much @Swiftb0y for the feedback! Preferences for key color palettes is definitely in the plans. I will apply the requested changes and try the item role method.

src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,61 @@
...
+                option.rect.y() + 2,
+                4, // width
+                option.rect.height() - 4,
+                qcolor);
+    }
+
+    const int rectWidth = 10; // 2px left padding + 4px width + 4px right padding
+
+    QString elidedText = option.fontMetrics.elidedText(
+            index.data().toString(),
```
ronso0, line 46:
```suggestion
            keyText,
```

src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    QString keyText = index.data().toString();
+    mixxx::track::io::key::ChromaticKey key = KeyUtils::guessKeyFromText(keyText);
+    int openKeyNumber = KeyUtils::keyToOpenKeyNumber(key);
+
+    if (openKeyNumber != 0) {
+        const auto rgbColor =
+                mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette.at(
+                        openKeyNumber - 1); // Open Key numbers start from 1
+        const auto qcolor = mixxx::RgbColor::toQColor(rgbColor);
```
ronso0, line 32: Just a note: we could move the QString -> QColor translation to KeyUtils so it's available in WKey, too (or some other widget).

src/util/color/predefinedcolorpalettes.cpp
ronso0: Super nice to have, follow up²: set the colors in the notation table in Key preferences

src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,61 @@
...
+                qcolor);
+    }
+
+    const int rectWidth = 10; // 2px left padding + 4px width + 4px right padding
+
+    QString elidedText = option.fontMetrics.elidedText(
+            index.data().toString(),
+            Qt::ElideLeft,
+            columnWidth(index) - rectWidth);
+
```
ronso0, line 49: Like in LocationDelegate we need to get the highlight color from the palette in order to paint the selected key text correctly:
https://github.com/mixxxdj/mixxx/blob/a48562ad96021bd9571c8fbc8d331291bc30c1fc/src/library/tabledelegates/locationdelegate.cpp#L11-L23

ronso0 (dismissed):
Thank you, this looks nice!

src/library/basetracktablemodel.cpp
```diff
@@ -121,7 +122,7 @@ void BaseTrackTableModel::setApplyPlayedTrackColor(bool apply) {
     s_bApplyPlayedTrackColor = apply;
 }
 
-//static
+// static
```
daschuer, line 125: No real issue here in this case. In general we try to avoid reformating untouched code to keep the change set minimal. Make sure your IDE does not automatically reformat whole files.

src/library/tabledelegates/keydelegate.cpp
daschuer: What's the reason for this request?

src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,61 @@
...
+        : TableItemDelegate(pTableView) {
+}
+
+void KeyDelegate::paintItem(
+        QPainter* painter,
+        const QStyleOptionViewItem& option,
+        const QModelIndex& index) const {
+    paintItemBackground(painter, option, index);
+
+    QString keyText = index.data().toString();
```
daschuer, line 24: Here you should request the color directly from the model using a role, the get around the guessing below.

daschuer (changes requested):

src/library/tabledelegates/keydelegate.cpp
Swiftb0y: potentially performance. constructors this trivial will likely always get inlined, but the compiler needs it in the header for that.

src/library/basetracktablemodel.cpp
danferns: Thank you for noting that, I've now set it to only format changes.

src/library/tabledelegates/keydelegate.cpp
danferns: Thank you! Just fixed this.

src/track/keyutils.cpp
```diff
@@ -462,6 +464,20 @@ double KeyUtils::keyToNumericValue(ChromaticKey key) {
...
+QColor KeyUtils::keyToColor(ChromaticKey key) {
+    int openKeyNumber = KeyUtils::keyToOpenKeyNumber(key);
+
+    if (openKeyNumber != 0) {
+        const auto& palette = mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette;
+        DEBUG_ASSERT(openKeyNumber <= palette.length());
+        const auto rgbColor = palette.at(openKeyNumber - 1); // Open Key numbers start from 1
+        return mixxx::RgbColor::toQColor(rgbColor);
+    } else {
+        return QColor::QColor(); // return invalid color
```
daschuer, line 477:
```suggestion
        return {}; // return invalid color
```

src/track/keyutils.cpp
```diff
@@ -462,6 +464,20 @@ double KeyUtils::keyToNumericValue(ChromaticKey key) {
     return key;
 }
 
+// static
+QColor KeyUtils::keyToColor(ChromaticKey key) {
+    int openKeyNumber = KeyUtils::keyToOpenKeyNumber(key);
+
+    if (openKeyNumber != 0) {
+        const auto& palette = mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette;
+        DEBUG_ASSERT(openKeyNumber <= palette.length());
```
daschuer, line 473:
```suggestion
        DEBUG_ASSERT(openKeyNumber <= palette.size() && openKeyNumber >= 1);
```

src/util/color/predefinedcolorpalettes.cpp
Swiftb0y: Yeah, Though I guess we'll have to decide on whether its better to put in the Color or the Key settings. The former would be more consistent IMO.

src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,56 @@
...
+    mixxx::track::io::key::ChromaticKey key = KeyUtils::guessKeyFromText(keyText);
+    const auto keyColor = KeyUtils::keyToColor(key);
+
+    if (keyColor.isValid()) {
+        // Draw the colored rectangle next to the key label
+        painter->fillRect(
+                option.rect.x() + 2,
```
ronso0, line 23: I suggest to remove the left padding in order to allow narrow column configurations.
Would look like this
image

src/library/tabledelegates/keydelegate.cpp
Swiftb0y: yes, guessing the key from the text is inefficient and slow.

src/library/tabledelegates/keydelegate.cpp
danferns: Done. Thank you for this suggestion!

src/library/tabledelegates/keydelegate.cpp
ronso0: Ah, okay. So just get the raw value (key code) via index->data(), cast to int and get the color from `KeyUtils::colorFromNumericValue(int)`?

ronso0: = don't get the **color** from the model, as requested, just the value

Swiftb0y: yeah, though I wouldn't even go over the `int` index and just store and retrieve the `mixxx::track::io::key::ChromaticKey` directly. otherwise we'll have to deal with out-of-bound values and so forth again.

src/util/color/predefinedcolorpalettes.cpp
danferns: Here's a little sketch:

image

Swiftb0y: I'd somehow try to put it in the same column. but I'm not a UI designer. Wdyt @ronso0?

danferns: I think it looks more compact:

image

Swiftb0y: maybe swap the order of the text and the check box so its all still right aligned? idk, IIANA Designer.

daschuer:
Can you update the screenshot in the initial description of this PR? This way it is more significant if one is later following links from the Changelog.

daschuer:
Is the draft state still correct or is it ready for final review and tests?

src/util/color/predefinedcolorpalettes.cpp
```diff
@@ -144,6 +144,21 @@ constexpr mixxx::RgbColor kVirtualDJTrackColorBlue(0x0000FF);
 constexpr mixxx::RgbColor kVirtualDJTrackColorFuchsia(0xFF00FF);
 constexpr mixxx::RgbColor kVirtualDJTrackColorWhite(0xFFFFFF);
 
+// Default Mixxx Key Color Palette
```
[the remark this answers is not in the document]
daschuer: https://www-datylon-com.cdn.ampproject.org/v/s/www.datylon.com/blog/data-visualization-for-colorblind-readers?amp_gsa=1&amp_js_v=a9&hs_amp=true&usqp=mq331AQIUAKwASCAAgM%3D#amp_tf=Von%20%251%24s&aoh=17198172132733&csi=1&referrer=https%3A%2F%2Fwww.google.com&ampshare=https%3A%2F%2Fwww.datylon.com%2Fblog%2Fdata-visualization-for-colorblind-readers

Interesting are the hue curves of different color blind people. So I guess in a follow up PR, we may offer the full hue curve and reduced versions for color blind people that is only using half of the hue range.

daschuer (approved):
Just a minor comment. Rest LGTM, thank you.

danferns:
> Is the draft state still correct or is it ready for final review and tests?

Yes, I am not done yet. I still have to read the Key Colors setting and only render the rectangle if it is enabled. 

Also I'd like to squash a couple of commits before review, at least the one which was failing to build.

danferns:
I want to read the setting using `ConfigKey("[Config]", "KeyColorsEnabled")` inside `KeyUtils::keyToColor` (if key colors are disabled, return an invalid color). But to do that, I would need access to the [`pConfig`](https://github.com/mixxxdj/mixxx/wiki/developer_guide_config#getting-and-setting-preference-options) variable inside keyutils.cpp. 

I see that the `pConfig` is passed around in class constructors. Is there a way to access it directly?

Should I go about this in some other way entirely?

Swiftb0y:
I don't think accessing the config in the KeyUtils is a good idea. Can you share a little more insight on what you're trying to accomplish? Then we can brainstorm together.

Swiftb0y:
Thanks for the explanation.
I think that check should stay inside `KeyDelegate`. The static data in `KeyUtils` is ugly and its usefulness is questionable IMO.

src/util/color/predefinedcolorpalettes.cpp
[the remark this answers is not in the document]
Swiftb0y: We could provide preset palettes optimized for the different types of colorblindness. I think gradients for those already exist online. This would be fairly straightforward to implement once the requusted custom keypalettes are in place.

danferns:
Got it, thank you!

daschuer:
In other cases like that we have added a ControlObject 
With the same name. Than you can connect to the value change signal inside the library view and refresh the whole view. I think this is later also required when changing the color schema. So you may consider to add a "None" shema for disable colors.

daschuer:
Like here:
https://github.com/mixxxdj/mixxx/blob/7b018fbe588f9c43270e63393179f5cc5cce9e6c/src/preferences/dialog/dlgprefkey.cpp#L48

ronso0:
> I was planning on calling `slotRefreshAllRows()` in the new [`BaseTrackTableModel::setKeyColorsEnabled`](https://github.com/mixxxdj/mixxx/blob/c0657253e34f141f17a37740ef20e4e8b2790b65/src/library/basetracktablemodel.cpp#L122) method, but then realized that this method needs to be static, while `slotRefreshAllRows` is not static.

I'd recommend to avoid that method. That refreshes the entire view and resorts, i.e. like purging tracks this has the undesired resorting side effect, see #12565 

I see that both the new color state and s_bpmColumnPrecision are applied in the next paint event, which happens for me when I close the preferences and see the table being redrawn. Don't know if that happens when the main window is raised again or only if the preferences covered the tracks table.

ronso0:
Btw the key color default is ON, but the checkbox is not checked when opening the Color pref page.

src/preferences/dialog/dlgprefcolors.cpp
```diff
@@ -77,6 +84,7 @@ DlgPrefColors::~DlgPrefColors() {
 void DlgPrefColors::slotUpdate() {
     comboBoxHotcueColors->clear();
     comboBoxTrackColors->clear();
+    bKeyColorsEnabled->setChecked(m_bKeyColorsEnabled);
```
ronso0, line 87: slotUpdate() is supposed to read the config values and update the view accordingly.
Here, the previous (GUI) value is used (that's why the box is still unchecked even though the setting is ON), and it's not obvious that `m_colorPaletteSettings.get..` queries the config.
Please read the config with `m_pConfig->getValue([color setting])`.

ronso0 (changes requested):

ronso0:
FWIW I expected the key color setting in the Library preferences since that is what handle the settings and appearance of the tracks table.

src/preferences/dialog/dlgprefcolors.cpp
```diff
@@ -345,6 +358,13 @@ void DlgPrefColors::slotEditHotcuePaletteClicked() {
     openColorPaletteEditor(hotcueColorPaletteName, true);
 }
 
+void DlgPrefColors::slotKeyColorsEnabled(int i) {
+    m_bKeyColorsEnabled = static_cast<bool>(i);
+    BaseTrackTableModel::setKeyColorsEnabled(m_bKeyColorsEnabled);
+    m_pConfig->setValue(kKeyColorsEnabledConfigKey, bKeyColorsEnabled->checkState());
+    slotUpdate();
```
ronso0, line 365: This slot is called when the checkbox is clicked (changed), so the checkbox state and the config state are now identical.
No need to update the entire page now.

ronso0: Sorry for the fragmented review, I didn't take a closer look earlier.

src/preferences/dialog/dlgprefcolorsdlg.ui
```diff
@@ -100,7 +100,19 @@
...
+    <widget class="QCheckBox" name="bKeyColorsEnabled">
+     <property name="toolTip">
+      <string>When key colors are enabled, Mixxx will display a color hint
+associated with each key.</string>
+     </property>
+     <property name="text">
+      <string>Enable Key Colors</string>
+     </property>
+    </widget>
+   </item>
```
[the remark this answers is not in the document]
danferns: Fixed, thank you.

src/preferences/dialog/dlgprefcolors.cpp
danferns: I did not quite get you. Did you mean that we don't need the `slotUpdate()` call because it gets called some other way when the checkbox is clicked?

ronso0: Yes, we don't need it because the checkbox state and the config state are identical after writing to the config and `setKeyColorsEnabled()`.

ronso0:
> FWIW I expected the key color setting in the Library preferences since that is what handle the settings and appearance of the tracks table.

What do you think?
Should we have the enable switch in the Library page, and the future palette config in Colors?

danferns:
> What do you think? Should we have the enable switch in the Library page, and the future palette config in Colors?

If we add key color support to the WKey widget in the future, then it feels more natural to me to have this setting under the Color preferences. Because then this toggle is not only changing the library view but also the deck view.

danferns:
Thank you @daschuer, @ronso0 and @Swiftb0y for your feedback so far! I think I'm ready for a final review.

I was thinking of cleaning up the git log a little, but I think I'll do it at the end before merging (would that be fine?). There might still be more changes to be made from now, and I'm also not sure how much to clean up.

Swiftb0y:
Yes, don't worry with cleaning up the log for now. I'm sure there will be a bunch more comments.

danferns:
I want to work on adding the key color palettes. I'm a bit unsure which branch to do it on. Should I branch off from this PR's branch, and then create a separate PR for the palettes? Or would it be better to keep working on this branch only?

ronso0:
I suggest to make "small" steps, i.e. get this merged first, then improve / tweak / extend this new feature in separate branches.

danferns:
Got it, thank you.

ronso0 (commented):
Alrighty, this LGTM except a few bits.

ronso0:
> maybe swap the order of the text and the check box so its all still right aligned? idk, IIANA Designer.

IMHO it's okay like this.
IIRC it's not possible to swap checkbox and label, that would need to be two separate widgets (QLabel + QCheckBox without label).

src/library/tabledelegates/keydelegate.cpp
```diff
@@ -0,0 +1,64 @@
...
+                    option.rect.x(),
+                    option.rect.y() + 2,
+                    4, // width
+                    option.rect.height() - 4,
+                    keyColor);
+        }
+    }
+
+    QString keyText = KeyUtils::keyToString(key);
+    if (QString("INVALID").compare(keyText) == 0) { // if the key is invalid, display empty string
```
ronso0, line 36: not the other way around, since we already have a QString and don't need to construct another one?
(based what I think I understood fo QString stuff)
```suggestion
    if (keyText.compare("INVALID") == 0) { // if the key is invalid, display empty string
```

Swiftb0y: good suggestion, but better to avoid string comparison altogether. https://github.com/mixxxdj/mixxx/pull/13390#discussion_r1672347393

danferns:
Thank you for the review. I have used the enum comparison now, and placed all the text/color rendering code inside this condition. 

Let me know when it is the right time for rebasing.

ronso0 (approved):
Yep, LGTM : )

src/preferences/colorpalettesettings.h
```diff
@@ -3,6 +3,8 @@
 #include "preferences/usersettings.h"
 #include "util/color/colorpalette.h"
 
+#define KEY_COLORS_ENABLED "KeyColorsEnabled"
```
Swiftb0y, line 6: can you remove the `#define`? just copy paste it into its only use below.

Swiftb0y (commented):
not perfect, but good enough considering the status quo. almost LGTM.

danferns:
> not perfect, but good enough considering the status quo.

I appreciate the honest feedback. Can you please tell me in what aspects I can improve?

Swiftb0y:
> I appreciate the honest feedback. 

Don't get me wrong, this PR is quite great already ;). There is just always room to improve. 

> Can you please tell me in what aspects I can improve?

The primary issue I have is that you followed some suboptimal patterns that are already in the existing code. I'm not expecting you to clean those up, though I'm not particularly fond of the fact that this PR is adding more suboptimal code. Refactoring the code in question is probably outside your skillset and GSoC project scope but there are a bunch of refactorings that are in order IMO. Ideally, all the static global variables (the ones prefixed `s_`) should be eliminated (not sure how, haven't looked too deep into it). `src/util/color/predefinedcolorpalettes.cpp` could use a little bit of a cleanup. `KeyDelegate::paintItem` could use a little more reordering to simplify its controlflow (prefer early returns) and separate the individual steps better (currently a lot of the low-level math of the drawing code is interspersed with the high-level drawing code) (see [SLAP](http://principles-wiki.net/principles:single_level_of_abstraction)). Passing `s_keyColorsEnabled` as part of the `BaseTrackTableModel` is weird (especially via the `roleValue()`) because the value doesn't change depending on the index.

Having said all that, I'm not expecting you to address these concerns. Instead discuss with your mentor (@daschuer IIRC?) on how to proceed.

Swiftb0y (approved):
LGTM. Thank you. Who wants to press the button? Did you want to rebase first?

danferns:
Thank you so much, this is very helpful. I'll discuss with Daniel for the next steps, and I'll keep these points in mind for future PRs. I am growing a lot thanks to this project. :) 

I do want to rebase!

daschuer:
Thank you for the review. Yes, there is room for improvements. On the other hand we have tight plan for the remaining project. From this point of view I would rather see a beautiful Heat column instead of beautiful code. 

We have already the green check marks so I will merge once @danferns gives a go. 

If you feeling lucky to address some of @Swiftb0y points, go ahead and propose it in a separate PR.

danferns:
I'm done with the rebase! I squashed only a build error, and two commits involving indentation / comments. If anything else is needed then let me know, but otherwise I am ready for this PR to be merged!

Swiftb0y:
> On the other hand we have tight plan for the remaining project. From this point of view I would rather see a beautiful Heat column instead of beautiful code.

I agree in this case.

daschuer:
Thank you very much! This looks really nice and works good.

acolombier:
After synching my dev branch, I just seen that eye candy appeared. It looks very good! Thanks for that feature!

Eve00000:
I tested this today, some remarks:
- I don't have a key palette under 'Enable Key Colors' in Preferences -> Colors
-Pushing 'Apply' after changing 'Enable Key Colors' does not work, changes only happen after push on 'OK'
- Nice color bar in 'Key'-column but Key text is cut of on the left side (normally text is cut of on the right side) while my most important part is on the left side.
afbeelding
afbeelding

ronso0:
> * I don't have a key palette under 'Enable Key Colors' in Preferences -> Colors

There is only one palette atm.

Repaint with/without colors only happens when the table is repainted after the preferences are closed.

PR #13393, opened 2024-06-22
Kontrol S4 Mk3: rework jogwheel speed compute and motorized platter

acolombier:
- Rework how the speed is computed when scratching in order to make it more accurate and sharp
- Reword motorized jog wheel to be more accurate
- Add main cue haptic feedback

Targeting `2.5` as it is a bugfix, reported multiple times.

### DIFF
--- a/res/controllers/Traktor Kontrol S4 MK3.hid.xml
+++ b/res/controllers/Traktor Kontrol S4 MK3.hid.xml
@@ -589,19 +589,6 @@
                 </option>
 
             </row>
-
-            <option
-                variable="wheelSpeedSample"
-                type="integer"
-                min="2"
-                max="80"
-                step="2"
-                default="5"
-                label="Speed sample">
-                <description>
-                    Define how many wheel ticks are sampled to compute the speed. Higher values will give smoother speeds at the expense of responsiveness.
-                </description>
-            </option>
             <option
                 variable="baseRevolutionsPerMinute"
                 type="enum"
@@ -616,21 +603,19 @@
                 variable="useMotors"
                 type="boolean"
                 default="false"
-                label="Use beta Haptic Drive features (BETA)">
+                label="Allow Mixxx to interact with jog wheel motors (BETA)">
                 <description>
                     This will defines whether or not Mixxx is allowed to interact with the motors. &lt;br /&gt;&lt;b&gt;WARNING&lt;/b&gt;: while interacting with motors should be harmless for your device, it could cause excess wear and tear.
                 </description>
             </option>
+
             <option
-                variable="turnTableSpeedSample"
-                type="integer"
-                min="2"
-                max="80"
-                step="2"
-                default="40"
-                label="Turntable speed sample">
+                variable="cueHapticFeedback"
+                type="boolean"
+                default="true"
+                label="Emit haptic feedback when the platter goes over the main cue">
                 <description>
-                    Define how many wheel ticks are sampled to compute the speed when using the motor. This is helpful to mitigate the delay that occurs in communication as well as Mixxx's built-in 20ms latency. Higher values will give smoother speeds at the expense of responsiveness.
+                    This will emit a haptic feedback on the platter while scratching over the main cue point.
                 </description>
             </option>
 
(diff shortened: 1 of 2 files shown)


acolombier:
Rebased on `2.5` as it is a bugfix, reported multiple times.

acolombier:
@ywwg if you have a chance, could you give that a go? :pray:

ywwg:
testing these changes, I didn't notice anything haptic  about the main cue. What is it supposed to do?

Kasper0s:
Built branch [traktor-s4-mk3-bleedingedge](https://github.com/acolombier/mixxx/tree/traktor-s4-mk3-bleedingedge) and I can report the haptic cue function is there. =)

For the motorised platter, _scratching_ feels nice and the tempo no longer "swims".
One issue though is that small adjustments of phase aren't possible by braking or speeding up the platter. The platter isn't responsive to adjustments up until a threshold.

acolombier:
@ywwg haptic feedback strength is controlled by the `Jogwheel torque` setting. Traktor uses a value of up to `48000` (no idea the unit here), while we default to a max of `25000` (customizable to `32000` max, although you can override to higher value if you want to). 
Two reasons I reduced the max compared to Traktor: the speed the platter would rotate at is seriously worrying, but also you can hear a mechanic noise that feels like it shouldn't be used for more than couple of μs (which is what Traktor does, as it has a polling interval going to a handful of μs, allow a very fast reactivity, while Mixxx has a static 20ms interval) 

Thanks for the feedback @Kasper0s !

> One issue though is that small adjustments of phase aren't possible by braking or speeding up the platter

Sadly, this is going be a either/or fix: that feature would require fast reactivity of the controller in Mixxx, but for performance reasons, the polling rate interval is limited to 20ms as mentioned above, which isn't enough to handle the platter movement.

ywwg:
for what it's worth, the amount of flutter on a real turntable is *large*, around +/- 2 bpm per rotation of the platter. How much were we seeing without the smoothing?  Without the ability to adjust pitch delicately by touching the platter, this function is not very useful.

acolombier:
I tend to agree with you, but there was a few report on this being a problem for user. I guess the issue comes if you play to tracks, as the desync isn't consistent around the decks. Happy to make that new behaviour configurable?

ywwg:
does the wheel have touch detection? we could use high smoothing when there's no touch, and then immediately go to instant feedback when the wheel is touched.

Kasper0s:
@acolombier I'm available for testing. =)

> does the wheel have touch detection? we could use high smoothing when there's no touch, and then immediately go to instant feedback when the wheel is touched.

This looks like a sound idea. Could do 2 finger braking (or pinch brake heh, nvm), for users accustomed to using the sides for brakes.
In case unclear, the top has touch sens, while the side has a TPU'ish coating with ribs.

Another idea: since rotary encoders move&loop have touch sens, incorporating this for the braking .. it's good to have good support while making adjustments but it sounds like a big learning curve until used to it.
This would although still leave enabled a way to use it as a trigger for another function .. _winkwink_ (likely it only senses touch though, not pressure?)

ywwg:
We can merge what we have and then experiment with different smoothing amounts / touch detection.  

I would also love to know how Traktor does it -- the flutter of the platter is physical, so they are adapting for it somehow

acolombier:
> We can merge what we have and then experiment with different smoothing amounts / touch detection.

I think it'd be great. I agree that more could be done, but also it sin't on my priority for the S4 MK3 (I have restarted putting full steam on the screens support)

@ywwg  I let press the button if you are happy with this :)

Swiftb0y:
Yes, this is a fix in the 2.5.1 release. However, this PR and the linked article don't relate to each other at all. `JogWheelBasic` is not used in this S4 mapping and `JogWheelBasic` has no notion of of motorized platters or anything of that sort. It's simply a convenient wrapper around the `engine.scratch*` functions.

Michael-Z-Freeman:
> Yes, this is a fix in the 2.5.1 release. However, this PR and the linked article don't relate to each other at all. `JogWheelBasic` is not used in this S4 mapping and `JogWheelBasic` has no notion of of motorized platters or anything of that sort. It's simply a convenient wrapper around the `engine.scratch*` functions.

Hi. Right, thanks. What should I be looking at for motorized platters ?

ywwg:
the javascript implementation for this controller has code for the motorized platters

PR #13394, opened 2024-06-23
feat: allow swapping hotcues

ronso0:
This allows to quickly reorder hotcues by position in the track, eg. if you've set a hotcue between existing hotcues (adjacent index and position).
_Previously this was a tedious procedure with either an empty hotcue slot or the main cue as temp storage._

mixxx-hotcue-swap

- [x] add drag'n'drop functionality to WHotcueButton
- [x] start drag with Shift that does not preview/play, ie. does not interrupt performance
- [x] add control `[player],hotcue_X_swap`, expect 1-based index as parameter
- [ ] add test?

**Follow-up:**
* add swap selector to CueMenuPopup
* allow dragging a previewing hotcue to Play button to latch 'play' (is currenty only possible with mouse + keyboard/controller)

_Note: a strange issue occurs (only seen in Tango): dragging a hotcue over the deck labels and then to another hotcue would make the 'hover' style show up for the lables after dropping :shrug:
This doesn't occur when dragging tracks (hover labels on deck1, then drop onto deck2)_


src/track/track.cpp
```diff
@@ -1103,6 +1103,44 @@ void Track::removeCuesOfType(mixxx::CueType type) {
...
+    if (pCueA) {
+        qWarning() << "     >> move" << from << "to" << to;
+        pCueA->setHotCue(to);
+    }
+    if (pCueB) {
+        qWarning() << "     >> move" << to << "to" << from;
+        pCueB->setHotCue(from);
+    }
+    emit cuesUpdated();
+}
```
Swiftb0y, line 1142: I removed the cuetype restriction as I didn't understand why it would be necessary. Wdyt?
```suggestion
void Track::swapHotcues(int a, int b) {
    auto locked = lockMutex(&m_qMutex);
    // REVIEW COMMENT: finding a specific cue by its index could also be a utility member function
    auto is_cue_n = [](int n) {
        return [n](const CuePointer& pCue){
            return pCue && pCue->getHotCue() == n;
        }
    };
    auto cueAIt = std::find_if(m_cuePoints.begin(), m_cuePoints.end(), is_cue_n(a));
    if (cueAIt == m_cuePoints.end()) {
        qWarning() << "could not find Cuepoint" << a << "for swapping with" << b;
        return;
    }
    auto cueBIt = std::find_if(m_cuePoints.begin(), m_cuePoints.end(), is_cue_n(b));
    if (cueBIt == m_cuePoints.end()) {
        qWarning() << "could not find Cuepoint" << b << "for swapping with" << a;
    }
    // we already know the pointers are non-null from is_cue_n
    // now we deliberately set cueB to a and cueA to b to perform the
    // actual swap. 
    (*cueBIt)->setHotCue(a);
    (*cueAIt)->setHotCue(b);
```

src/widget/whotcuebutton.cpp
Swiftb0y, line 1: the Qt specific changes in here are not my strength. Can somebody else look at them?

src/track/cue.h
```diff
@@ -104,7 +105,7 @@ class Cue : public QObject {
     mixxx::CueType m_type;
     mixxx::audio::FramePos m_startPosition;
     mixxx::audio::FramePos m_endPosition;
-    const int m_iHotCue;
+    mutable int m_iHotCue;
```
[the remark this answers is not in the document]
ronso0: Thanks for your review. I should have noted that this is still a raw version.
IIRC mutable here was just a quick fix for some build warning or error I didn't investigate when I noticed it because I wanted it to work somehow, and test it.

In general I'm not 100% sure if the naive solution of simply re-assigning the hotcue number is the way to go, but I suspect the only reason the cue number was const before was that no one thought of the reassign/move use case.

src/track/track.cpp
ronso0: Yes, we can remove the type test.
The idea is that we can move a set hotcue to an empty slot, or swap two set hotcues.
So we need one of both cues, i.e. aborting if either of the two is missing is not correct.

Two `std::find_if()` iterations are faster than one for loop that looks for both cues?

src/widget/whotcuebutton.cpp
ronso0: Are you referring to the QDrag implementation?
Mostly copied from [WEffectParameterNameBase](https://github.com/mixxxdj/mixxx/blob/a48562ad96021bd9571c8fbc8d331291bc30c1fc/src/widget/weffectparameternamebase.cpp#L127) btw.

The order of events is
* mouseMoveEvent: initiates a QDrag if the position delta is big enough
* dragEnterEvent: is accepted if the hotcue swap conditions are met (same player, different hotcue index)
* dropEvent: only occurs if that widget has [accepted the most recent dragEnter/dragMove event](https://doc.qt.io/qt-6/qdropevent.html#details) (that's why we don't need the run tests we already ran in dragEnterEvent)

src/track/track.cpp
Swiftb0y: >  So we need only one cue with index a or b, i.e. aborting if either of the two is missing is not correct.

Ah, I see. I'll think about how to express that cleanly.

> Two std::find_if() iterations are faster than one for loop that looks for both cues?

Probably not, depends on the loop body, but its simpler IMO and probably doesn't matter here, especially if you refactor it into a private getHotcueByIndex function.

src/track/cue.h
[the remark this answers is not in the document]
Swiftb0y: Oh yeah, I didn't even notice that. const members are also an anti feature. Simplfy removing the specifier should make it work as expected. Then the only change you need to make is to ensure that you actually get a non-const `Cue` object when look for it in `swapHotcues`.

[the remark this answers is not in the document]
Swiftb0y: > In general I'm not 100% sure if the naive solution of simply re-assigning the hotcue number is the way to go, 

I think it should be fine. Though my main concern is that it makes it easier to accidentally make two cues with the same index. An alternative solution could be to have a 
`Cue::swap(&Cue other)` method that does the swap internally without the need for a public `setHotCue` setter.

src/widget/whotcuebutton.cpp
Swiftb0y: > Are you referring to the QDrag implementation?

Yeah, but also all of the other MIME-related stuff...

src/engine/controls/cuecontrol.cpp
```diff
@@ -1218,6 +1223,26 @@ void CueControl::hotcueClear(HotcueControl* pControl, double value) {
...
+    if (!m_pLoadedTrack) {
+        return;
+    }
+
+    CuePointer pCue = pControl->getCue();
+    if (!pCue) {
+        return;
+    }
+
+    m_pLoadedTrack->swapHotcues(pCue->getHotCue(), newCuenum);
```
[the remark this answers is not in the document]
ronso0: Yup, both are QRecursiveMutex.

src/track/cue.h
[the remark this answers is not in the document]
ronso0: > `Cue::swap(&Cue other)` method that does the swap internally

Sounds good, I'll try that.
Though I'd like to restrict it to hotcues for now, `Cue::swapHotcue(Cue& other)`, I don't see the need for swapping other (potentially inaccesseible, firstSound) cues (intro, outro, temp loop, main cue).

ronso0:
I removed the cue menu GUI controls (I'm not motivated to add styles and polish for all skins) so this can be merged soonish.

This is now contains only the drag'n'drop feature, and the controls, in case someone wants to use it in mappings.

src/track/track.cpp
```diff
@@ -1062,6 +1062,23 @@ CuePointer Track::findCueById(DbId id) const {
...
 }
 
+CuePointer Track::findHotcueByIndex(int idx) const {
+    auto locked = lockMutex(&m_qMutex);
+    auto is_hotcue_n = [](int n) {
+        return [n](const CuePointer& pCue) {
+            return pCue && pCue->getHotCue() == n;
+        };
+    };
+    auto cueIt = std::find_if(m_cuePoints.begin(), m_cuePoints.end(), is_hotcue_n(idx));
```
Swiftb0y, line 1072: any specific reason for the double lambda? It could also have just been one.

src/track/track.cpp
```diff
@@ -1062,6 +1062,23 @@ CuePointer Track::findCueById(DbId id) const {
...
+    if (cueIt != m_cuePoints.end()) {
+        qWarning() << "     > found hotcue" << idx;
+        return *cueIt;
+    } else {
+        qWarning() << "     ! no hotcue" << idx;
```
Swiftb0y, line 1077: this looks like leftover debugging.

src/track/track.cpp
```diff
@@ -1103,6 +1120,33 @@ void Track::removeCuesOfType(mixxx::CueType type) {
...
+    VERIFY_OR_DEBUG_ASSERT(a != b) {
+        qWarning() << "Track::swapHotcues rejected," << a << "==" << b;
+        return;
+    }
+    VERIFY_OR_DEBUG_ASSERT(a > Cue::kNoHotCue || b > Cue::kNoHotCue) {
+        qWarning() << "Track::swapHotcues rejected, both a and b are kNoHotCue";
+        return;
+    }
+    auto locked = lockMutex(&m_qMutex);
+    qWarning() << "     Track::swapHotcues" << a << "and" << b;
```
Swiftb0y, line 1133: same leftover debugging.

src/widget/whotcuebutton.cpp
```diff
@@ -129,16 +145,95 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
...
+    // Note that Track::swapHotcues() allows both directions.
+    if (m_hotcue == Cue::kNoHotCue) {
+        return;
+    }
+
+    if (DragAndDropHelper::mouseMoveInitiatesDrag(pEvent)) {
+        QDrag* pDrag = new QDrag(this);
+        QMimeData* mimeData = new QMimeData;
+        mimeData->setText(mimeTextIdentifier() + kMimeTextDelimiter + QString::number(m_hotcue));
+        pDrag->setMimeData(mimeData);
```
Swiftb0y, line 175: this is slightly safer because it doesn't leak in case something throws before calling `setMimeData`. Also avoids a "naked new".
```suggestion
        auto mimeData = std::make_unique<QMimeData>();
        mimeData->setText(mimeTextIdentifier() + kMimeTextDelimiter + QString::number(m_hotcue));
        pDrag->setMimeData(mimeData.release());
```

src/widget/whotcuebutton.cpp
```diff
@@ -129,16 +145,95 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
...
+        // Use the currently rendered button as dnd cursor.
+        const QPixmap currLook = grab(rect().marginsRemoved(m_dndRectMargins));
+        pDrag->setDragCursor(currLook, Qt::MoveAction);
+        // Release this button after failed or successful drop.
+        // This prevents both the preview and the pressed state from getting stuck.
+        connect(pDrag,
+                &QDrag::destroyed,
+                this,
+                &WHotcueButton::release);
```
Swiftb0y, line 184: this looks a little odd. Is this a pattern recommended by Qt?

src/widget/whotcuebutton.cpp
```diff
@@ -129,16 +145,95 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
...
+    }
+}
+
+void WHotcueButton::dragEnterEvent(QDragEnterEvent* pEvent) {
+    if (pEvent->source() == this) {
+        pEvent->ignore();
+        return;
+    }
+    TrackPointer pTrack = PlayerInfo::instance().getTrackInfo(m_group);
+    if (!pTrack) {
```
Swiftb0y, line 195: should this be a `VERIFY_OR_DEBUG_ASSERT`?

src/widget/whotcuebutton.cpp
```diff
@@ -129,16 +145,95 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
...
+        pEvent->ignore();
+        return;
+    }
+    const QString& mimeText = pEvent->mimeData()->text();
+    QStringList mimeTextLines = mimeText.split(kMimeTextDelimiter);
+    if (mimeTextLines.size() != 2 || mimeTextLines.at(0) != mimeTextIdentifier()) {
```
Swiftb0y, line 201: since we previously essentially serialized the string data and now deserialize it again here, would it not make more sense instead to pack both in a struct and use binary data instead ([`QMimeData::setData`](https://doc.qt.io/qt-6/qmimedata.html#setData))?

src/widget/whotcuebutton.cpp
```diff
@@ -129,16 +145,95 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
...
+    TrackPointer pTrack = PlayerInfo::instance().getTrackInfo(m_group);
+    if (!pTrack) {
+        return;
+    }
+    const QString& mimeText = pEvent->mimeData()->text();
+    QStringList mimeTextLines = mimeText.split(kMimeTextDelimiter);
+    int otherCueNum = mimeTextLines.at(1).toInt();
+    pTrack->swapHotcues(otherCueNum, m_hotcue);
+    // Alternative way to release drag source button:
+    // WHotcueButton* srcBtn = qobject_cast<WHotcueButton*>(pEvent->source());
```
Swiftb0y, line 222: lets avoid the qobject_cast if possible.

src/widget/whotcuebutton.cpp
```diff
@@ -129,16 +145,95 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
...
+    // srcBtn->release();
+}
+
+const QString WHotcueButton::mimeTextIdentifier() const {
+    return QStringLiteral("hotcue_") + m_group;
+}
+
+void WHotcueButton::release() {
+    // Actions from WPushButton::mouseReleaseEvent()
+    // TODO Create a QMouseVent and call mouseReleaseEvent instead of this hack?
```
Swiftb0y, line 233: where can I find some example code on this?

src/track/track.cpp
ronso0: Hehe, I didn't question that when I accepted you're proposal.

I have no clue how to capture the iterator and n with that function.

But this seems to work:
```suggestion
    auto cueIt = std::find_if(
            m_cuePoints.begin(),
            m_cuePoints.end(),
            [&idx](const auto pCue) {
                return pCue && pCue->getHotCue() == idx;
            });
```

Swiftb0y: I did that? Yeah seems like the overcomplicated stuff I like to do :sweat_smile: 

Probably just wanted to give it a name. I should really try to KISS more often.

What you posted works but is suboptimal because 

1. you're capturing the `int idx` by reference (just capture it by value as per the usual guidance on cheap to copy types).
2. you're taking the pCue by value (`auto` never deduces to a reference, See scott meyers "Effective Modern C++" for a very good explanation on type deduction). Since pCue is some sort of reference-counted shared pointer this causes unnecessary changes in the reference count. So instead take a reference to the pointer like in the original lambda.

We can also keep it like this, just wanted to give my two cents on your proposal.

src/widget/whotcuebutton.cpp
ronso0: It's been a while, I don't remember how I came up with this.

It seems this is wrong anyway (at least how events in the target are rejected):
accidentally dragging a hotcue while previewing could stop the preview.

Will fix this.

ronso0:
Hmm, suddenly I experience issues :disappointed: 
* when dragging only a **short distance** (enough to initiate a drag), ie. drag source = drag target
* when **quickly** dragging to a widget other than WHotcueButton
the button is released and hotcue preview stops.
I have no clue why this happens. The QDrag has not been destroyed, yet, and I can still swap cues afterwards.

I didn't see this when last working on this. In the meantime I switched from Qt 6.2.3 to 6.7.2 so this _might_ be the reason (new bug or bug fixed in 6.7.2) :shrug: 
This happens regardless if I override QWidget::dragEnterEvent() in WHotcueButton (but it's required to QEvent::accept() in order to get the hotcue cursor image).

To play it safe, I'll assign hotcue swap to **Shift + drag**.

ronso0:
Okay, it's a QLeaveEvent being captured in WPushButton::event().
Though, I have no clue why that occurs even if the drag cursor did not yet leave the source button. Qt is fun...

src/widget/whotcuebutton.cpp
ronso0: No, this can happen when dragging a from another player while this no track loaded.

src/widget/whotcuebutton.cpp
ronso0: I'll look into that.

Swiftb0y: does it need to be a `QString`?. What real information is encoded in there? From what I saw its just `"hotcue_"+m_group`, right? does it need to be `m_group` or can the same data be interpreted in a more binary-friendly fashion?

ronso0: Okay, we could as well use the TrackId (= DbId = int).

I came up with this c318193a58 (first time using QByteArray / QDataStream, so bare with me..)
But I'm not sure if the complexity is worth it.

Either way, using the TrackId instead of the player group is a good idea!

ronso0:
I fixed the issue with the sporadic Leave events. (FWIW [there is a distinction between QEvent::Leave and QEvent::HoverLeave](https://doc.qt.io/qt-6/qhoverevent.html#details))

This and previewing now work now as desired.

ronso0:
I fixed the issue with the sporadic Leave events. (FWIW [there is a distinction between QEvent::Leave and QEvent::HoverLeave](https://doc.qt.io/qt-6/qhoverevent.html#details))

This and previewing now work now as desired.

src/widget/whotcuebutton.cpp
Swiftb0y: yeah, thatll work. I thinks its the right direction, just needs a little cleanup

src/track/track.cpp
```diff
@@ -1103,6 +1118,30 @@ void Track::removeCuesOfType(mixxx::CueType type) {
     }
 }
 
+void Track::swapHotcues(int a, int b) {
+    VERIFY_OR_DEBUG_ASSERT(a != b) {
+        qWarning() << "Track::swapHotcues rejected," << a << "==" << b;
+        return;
+    }
+    VERIFY_OR_DEBUG_ASSERT(a > Cue::kNoHotCue || b > Cue::kNoHotCue) {
```
Swiftb0y, line 1126: why `>` instead of `!=`?

src/widget/whotcuebutton.cpp
Swiftb0y: feel free to push that commit so i can review it in context.

acolombier:
>     * [ ]  add test?

I think tests would be great for that one :+1:

src/track/track.cpp
ronso0: Because **I know** that hotcue indices are valid only if they are > 
Of course that might change so != is safer.

ronso0:
> > ```
> > * [ ]  add test?
> > ```
> 
> I think tests would be great for that one 👍

Well, I proposed that, but now I'm not sure if the responsible function [Track::swapHotcues()](https://github.com/mixxxdj/mixxx/pull/13394/files#diff-2cdcb9b0644e5717cbb7de0b3dc6990425eadc37fe4649253a2f44f41fc8d963R1121), which is really compact, requires testing tbh. Test would only feed some valid and some invalid pairs of a, b :shrug: 
Or what exactly did you think of?

ronso0:
Okay, wrt the conversation about mimedata https://github.com/mixxxdj/mixxx/pull/13394#discussion_r1780070403 I switched to using the TrackId instead of the player identifier. (and introduced HotcueDragInfo)
(not sure if _that_ de/serialization is more effeicient that the string variant, but it looks more beautiful ; ) )

This prevents the potential issue of undesired swapping when the track is changed while a swap drag is in progress (_might_ happen, eg. when AutoDJ replaces the track).
Extra 'feature' is that one could also drag a hotcue to another deck with the same track loaded :shrug:

src/track/track.cpp
ronso0: done.

src/widget/whotcuebutton.cpp
```diff
@@ -129,16 +146,100 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
...
+        return;
+    }
+
+    if (DragAndDropHelper::mouseMoveInitiatesDrag(pEvent)) {
+        const TrackId id = pTrack->getId();
+        VERIFY_OR_DEBUG_ASSERT(id.isValid()) {
+            return;
+        }
+        QDrag* pDrag = new QDrag(this);
+        HotcueDragInfo dragData(id, m_hotcue);
```
ronso0, line 180: Interesting that this works locally but CI 8rightly) complains...

will fix it

src/widget/whotcuebutton.h
```diff
@@ -2,12 +2,41 @@
...
 #include <QString>
 
+#include "track/trackid.h"
 #include "util/parented_ptr.h"
 #include "widget/wcuemenupopup.h"
 #include "widget/wpushbutton.h"
 
 class WHotcueButton : public WPushButton {
     Q_OBJECT
+
```
ronso0, line 12: could as well be a class, but we need it only here, so :shrug:

src/widget/whotcuebutton.h
```diff
@@ -2,12 +2,41 @@
...
+        HotcueDragInfo(TrackId id, int cue)
+                : trackId(id),
+                  hotcue(cue){};
+
+        HotcueDragInfo(QByteArray& bytes) {
+            QDataStream stream(&bytes, QIODevice::ReadOnly);
+            QVariant idVar;
+            stream >> idVar >> hotcue;
+            trackId = TrackId(idVar);
+        };
```
Swiftb0y, line 23: lets improve this a bit
```suggestion
        static HotcueDragInfo fromByteArray(const QByteArray& bytes) {
            QDataStream stream(bytes);
            QVariant idVar;
            int hotcue;
            stream >> idVar >> hotcue;
            return HotcueDragInfo(TrackId(idVar), hotcue);
        };
```

src/widget/whotcuebutton.h
```diff
@@ -2,12 +2,41 @@
...
+
+        const QByteArray toByteArray() {
```
Swiftb0y, line 25: no need to return a value as const.
```suggestion
        QByteArray toByteArray() {
```

src/widget/whotcuebutton.cpp
```diff
@@ -129,16 +146,104 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
...
+        // Use the currently rendered button as dnd cursor
+        // (incl. hover and pressed style).
+        // Note: for some reason, both grab() and render() render with sharp corners,
+        // ie. qss 'border-radius' is not applied to the drag image.
+        const QPixmap currLook = grab(rect().marginsRemoved(m_dndRectMargins));
+        pDrag->setDragCursor(currLook, Qt::MoveAction);
+
+        m_dragging = true;
+        pDrag->exec();
+        m_dragging = false;
```
Swiftb0y, line 198: `m_dragging` is never read as far as I can tell. Is this necessary or am I missing something?

Swiftb0y:
The only thing I still want to look into is avoiding the QVariant roundtrip during the (de-)serialization. Otherwise this looks pretty good IMO.

src/widget/whotcuebutton.cpp
ronso0: It's read in WPushButton::event() and prevents that a leave event releases the button while dragging.

Swiftb0y: ah of course. I didn't find it because I had already marked that file as viewed... Thanks.

Swiftb0y:
sorry, misclick (scrolled down and clicked into the textbox but the window lagged and the close button scrolled under the cursor while I wasn't looking :sweat_smile:).

Swiftb0y:
so upon examining `TrackId` (or rather `DbId`), it explicitly tries to not leak the internal datatype representation to the outside. So going through a variant and converting that to the underlying int violates that as well. Following that principle from `DbId`, we instead need to add the appropriate serialization primitives to `DbId` itself. Then use those directly (its more maintainable and faster):
```cpp
// serialization
friend QDataStream& operator<<(QDataStream &out, const DbId& dbId) {
    // explicit cast as recommended by Qt docs
    return out << static_cast<quint32>(dbId.m_value);
}

// deserialization
friend QDataStream& operator>>(QDataStream &in, DbId& dbId) {
    quint32 v;
    in >> v;
    dbId.m_value = v;
    return in;
}
```

src/widget/whotcuebutton.cpp
ronso0: yeah this isn't obvious, I'll add a comment.

ronso0:
I feared that I had to mess with the operators ; )

I wonder if we still need the HotcueDragInfo struct then. IIUC it would take any QByteSream starting with to int so it doesn't help really help.
We could simply de/serialize the stream in the mouseMove/dragEnter/drop events, right?

Swiftb0y:
yes, we should keep it because it does provide nicer encapsulation and because its a good place to bundle the serialization code (which is used multiple times).

Swiftb0y (commented):
Haven't tested yet, but code LGTM. no serious complaints anymore.

ronso0:
Thank you very much, I'll squash the commits then.

ronso0:
Ready for merge then?

Swiftb0y (approved):
Sure...

ronso0:
Wooo.. thanks!

PR #13397, opened 2024-06-24
fix(engine-prime): build-failure

Swiftb0y:
Explicitly instantiate default destructor in out-of-line (in .cpp) because destructing a `std::unique_ptr<Waveform>` needs `Waveform` to be complete at the point of destruction.

### DIFF
--- a/src/library/export/engineprimeexportjob.cpp
+++ b/src/library/export/engineprimeexportjob.cpp
@@ -378,6 +378,10 @@ EnginePrimeExportJob::EnginePrimeExportJob(
     }
 }
 
+// out-of-line declaration because we can't generate dtor in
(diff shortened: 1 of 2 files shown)


Swiftb0y:
Interestingly, this only surfaced because of f031f51393c5ae2bb923fc331408330878285617

daschuer (changes requested):
Please add also the override keyword. 
It is a good practice to make the destructor virtual when the class had virtual functions. A virtual function can not be inlined btw.

Swiftb0y:
right, but `~EnginePrimeExportJob` doesn't need to be virtual as the class is not designed to be inherited from technically, but I'll add it anyways.

daschuer (approved):
Thank you. LGTM

PR #13399, opened 2024-06-24
chore(ci): update azure code-signing action to renamed replacement

Swiftb0y:
The github action `azure/azure-code-signing-action` has been renamed to `azure/trusted-signing-action`. The former is being deleted on Jun 30th so this updates to the new name along with bumping the version to latest.

https://github.com/azure/azure-code-signing-action?tab=readme-ov-file

deprecation notice from repo


daschuer (approved):
Ok, let's try

Swiftb0y:
Thanks

Issue #5843, opened 2022-08-22
Show minute markers on overview waveform


mixxxbot:
Reported by: **[raffitea](https://launchpad.net/~raffitea)**
Date: 2011-03-31T13:59:28Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp746456](https://bugs.launchpad.net/bugs/746456)
Tags: easy, weekend
Attachments: [minutemarker.patch](https://bugs.launchpad.net/bugs/746456/+attachment/2726897/+files/minutemarker.patch), [This patch solves the mentioned bug.](https://bugs.launchpad.net/bugs/746456/+attachment/2732176/+files/This patch solves the mentioned bug.), [Minute Markers 2](https://bugs.launchpad.net/bugs/746456/+attachment/5411922/+files/Minute Markers 2), [Minute Markers with parallel waverforsm](https://bugs.launchpad.net/bugs/746456/+attachment/5411923/+files/Minute Markers with parallel waverforsm), [skin-mockup.png](https://bugs.launchpad.net/bugs/746456/+attachment/5435658/+files/skin-mockup.png)

---

Some skins like NightLightBlues have "markers" on the overview waveform but their distances are not 60 seconds long. I think we should introduce minute markers for that purpose.

mxmilkiib:
See also #12994

ChrisTallon:
Mixxx 2.4.1 with minute markers on the overview waveform.

https://github.com/ChrisTallon/mixxx/tree/2.4.1-minutemarkers

To build, follow all the normal instructions but use this git-clone command instead:

git clone -b 2.4.1-minutemarkers https://github.com/ChrisTallon/mixxx.git

This is a quick hack job.... Comments welcome.

Swiftb0y:
Hi @ChrisTallon if you intend to get your changes included in mixxx, please open a pull request. See https://github.com/mixxxdj/mixxx/wiki/Contribution%20Guidelines

Swiftb0y:
This has been implemented in Mixxx 2.6 by @ChrisTallon in #13401. Thank you very much.

PR #13401, opened 2024-06-24
Add minute markers on horizontal waveform overview

ChrisTallon:
Some users would like a visual indication of minutes on the waveform overview.

There is a discourse thread here: https://mixxx.discourse.group/t/do-minute-markers-exist-in-this-software/27477

There is a bug about it #5843.

Without minute markers:

image

With minute markers:

image

An extra setting in the preferences dialog waveform section to control minute markers:

image

It defaults to off to maintain current Mixxx behaviour. There is no support for vertical waveform overviews - I don't know of a skin to use to write and test it with.

### DIFF
--- a/src/preferences/dialog/dlgprefwaveform.cpp
+++ b/src/preferences/dialog/dlgprefwaveform.cpp
@@ -49,6 +49,11 @@ DlgPrefWaveform::DlgPrefWaveform(
         defaultZoomComboBox->addItem(QString::number(100 / static_cast<double>(i), 'f', 1) + " %");
     }
 
+    m_pOverviewMinuteMarkersControl = std::make_unique<ControlObject>(
+            ConfigKey(QStringLiteral("[Waveform]"),
+                    QStringLiteral("DrawOverviewMinuteMarkers")));
+    m_pOverviewMinuteMarkersControl->setReadOnly();
+
     // Populate untilMark options
     untilMarkAlignComboBox->addItem(tr("Top"));
     untilMarkAlignComboBox->addItem(tr("Center"));
@@ -146,6 +151,10 @@ DlgPrefWaveform::DlgPrefWaveform(
             &QCheckBox::toggled,
             this,
             &DlgPrefWaveform::slotSetNormalizeOverview);
+    connect(overviewMinuteMarkersCheckBox,
+            &QCheckBox::toggled,
+            this,
+            &DlgPrefWaveform::slotSetOverviewMinuteMarkers);
     connect(factory,
             &WaveformWidgetFactory::waveformMeasured,
             this,
@@ -271,6 +280,11 @@ void DlgPrefWaveform::slotUpdate() {
     }
     slotSetWaveformOverviewType(overviewType);
 
+    bool drawOverviewMinuteMarkers = m_pConfig->getValue(
+            ConfigKey("[Waveform]", "DrawOverviewMinuteMarkers"), true);
+    overviewMinuteMarkersCheckBox->setChecked(drawOverviewMinuteMarkers);
+    m_pOverviewMinuteMarkersControl->forceSet(drawOverviewMinuteMarkers);
+
     WaveformSettings waveformSettings(m_pConfig);
     enableWaveformCaching->setChecked(waveformSettings.waveformCachingEnabled());
     enableWaveformGenerationWithAnalysis->setChecked(
@@ -328,6 +342,9 @@ void DlgPrefWaveform::slotResetToDefaults() {
 
     // Don't normalize overview.
     normalizeOverviewCheckBox->setChecked(false);
+
+    // Show minute markers.
+    overviewMinuteMarkersCheckBox->setChecked(true);
 
     // 60FPS is the default
     frameRateSlider->setValue(60);
@@ -531,6 +548,11 @@ void DlgPrefWaveform::slotSetVisualGainHigh(double gain) {
 
 void DlgPrefWaveform::slotSetNormalizeOverview(bool normalize) {
     WaveformWidgetFactory::instance()->setOverviewNormalized(normalize);
+}
+
+void DlgPrefWaveform::slotSetOverviewMinuteMarkers(bool draw) {
+    m_pConfig->setValue(ConfigKey("[Waveform]", "DrawOverviewMinuteMarkers"), draw);
+    m_pOverviewMinuteMarkersControl->forceSet(draw);
 }
 
 void DlgPrefWaveform::slotWaveformMeasured(float frameRate, int droppedFrames) {
--- a/src/widget/woverview.cpp
+++ b/src/widget/woverview.cpp
@@ -83,6 +83,13 @@ WOverview::WOverview(
     m_pTypeControl->connectValueChanged(this, &WOverview::slotTypeChanged);
     slotTypeChanged(m_pTypeControl->get());
 
+    m_pMinuteMarkersControl = make_parented<ControlProxy>(
+            QStringLiteral("[Waveform]"),
+            QStringLiteral("DrawOverviewMinuteMarkers"),
+            this);
+    m_pMinuteMarkersControl->connectValueChanged(this, &WOverview::slotMinuteMarkersChanged);
+    slotMinuteMarkersChanged(static_cast<bool>(m_pMinuteMarkersControl->get()));
+
     m_pPassthroughLabel = make_parented<QLabel>(this);
 
     setAcceptDrops(true);
@@ -414,6 +421,10 @@ void WOverview::slotTypeChanged(double v) {
     slotWaveformSummaryUpdated();
 }
 
+void WOverview::slotMinuteMarkersChanged(bool /*unused*/) {
+    update();
+}
+
 void WOverview::updateCues(const QList<CuePointer> &loadedCues) {
     for (const CuePointer& currentCue : loadedCues) {
         const WaveformMarkPointer pMark = m_marks.getHotCueMark(currentCue->getHotCue());
@@ -618,6 +629,7 @@ void WOverview::paintEvent(QPaintEvent* pEvent) {
         drawEndOfTrackBackground(&painter);
         drawAxis(&painter);
         drawWaveformPixmap(&painter);
+        drawMinuteMarkers(&painter);
         drawPlayedOverlay(&painter);
         drawPlayPosition(&painter);
         drawEndOfTrackFrame(&painter);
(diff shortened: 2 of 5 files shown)


ronso0:
Thank you for this PR, this looks good so far!

Just a quick note: WaveformWidgetFactory only deals with the scrolling waveforms. I see it's handy to use that for getting the minute markers bool in every paint event, but you may also use static enableMinuteMarkers(bool enable) in WOverview or create a ControlObject in the preferences and the respective ControlProxy in WOverview.
Will do a more thorough review soonish.

src/widget/woverview.cpp
```diff
@@ -699,6 +700,42 @@ void WOverview::drawWaveformPixmap(QPainter* pPainter) {
     }
 }
 
+void WOverview::drawMinuteMarkers(QPainter* pPainter) {
+    if (!m_trackLoaded)
+        return;
```
Swiftb0y, line 705:
```suggestion
    if (!m_trackLoaded) {
        return;
    }
```

src/widget/woverview.cpp
```diff
@@ -699,6 +700,42 @@ void WOverview::drawWaveformPixmap(QPainter* pPainter) {
...
+    WaveformWidgetFactory* widgetFactory = WaveformWidgetFactory::instance();
+    if (!widgetFactory->isOverviewMinuteMarkers())
+        return;
+
```
Swiftb0y, line 709:
```suggestion
    if (!widgetFactory->isOverviewMinuteMarkers()) {
        return;
    }
```

src/widget/woverview.cpp
```diff
@@ -699,6 +700,42 @@ void WOverview::drawWaveformPixmap(QPainter* pPainter) {
...
+    // Faster than track->getDuration() and already has playback speed ratio compensated for
+    double trackSeconds = samplePositionToSeconds(getTrackSamples());
```
Swiftb0y, line 711:
```suggestion
    // Faster than track->getDuration() and already has playback speed ratio compensated for
    const double trackSeconds = samplePositionToSeconds(getTrackSamples());
```

src/widget/woverview.cpp
```diff
@@ -699,6 +700,42 @@ void WOverview::drawWaveformPixmap(QPainter* pPainter) {
...
+    QLineF line;
+    pPainter->setPen(QPen(m_axesColor, m_scaleFactor));
+    pPainter->setOpacity(1);
+
+    double overviewHeight = height();
+    double markerHeight = overviewHeight * 0.08;
+    double lowerMarkerYPos = overviewHeight * 0.92;
+    double currentMarkerXPos;
+    double currentMarkerSeconds = 60;
+    while (1) {
```
Swiftb0y, line 722:
```suggestion
    const double overviewHeight = height();
    const double markerHeight = overviewHeight * 0.08;
    const double lowerMarkerYPos = overviewHeight * 0.92;
    const double currentMarkerXPos;
    for (double currentMarkerSeconds = 60; currentMarkerSeconds <= trackSeconds; currentMarkerSeconds += 60) {
```

src/widget/woverview.cpp
```diff
@@ -699,6 +700,42 @@ void WOverview::drawWaveformPixmap(QPainter* pPainter) {
...
+            pPainter->drawLine(line);
+            line.setLine(currentMarkerXPos, lowerMarkerYPos, currentMarkerXPos, overviewHeight);
+            pPainter->drawLine(line);
+        } else {
+            // No support for vertical waveform overview
+        }
+        currentMarkerSeconds += 60;
+        if (currentMarkerSeconds > trackSeconds)
+            break;
+    }
```
Swiftb0y, line 736:
```suggestion
```

src/widget/woverview.cpp
```diff
@@ -699,6 +700,42 @@ void WOverview::drawWaveformPixmap(QPainter* pPainter) {
...
+        currentMarkerXPos = currentMarkerSeconds / trackSeconds * width();
+
+        if (m_orientation == Qt::Horizontal) {
+            line.setLine(currentMarkerXPos, 0.0, currentMarkerXPos, markerHeight);
+            pPainter->drawLine(line);
+            line.setLine(currentMarkerXPos, lowerMarkerYPos, currentMarkerXPos, overviewHeight);
+            pPainter->drawLine(line);
+        } else {
+            // No support for vertical waveform overview
+        }
```
Swiftb0y, line 732:
```suggestion
        } else {
            // untested, best effort basis
            line.setLine(0.0, currentMarkerXPos, markerHeight, currentMarkerXPos);
            pPainter->drawLine(line);
            line.setLine(lowerMarkerYPos, currentMarkerXPos, overviewHeight, currentMarkerXPos);
            pPainter->drawLine(line);
        }
```

src/waveform/waveformwidgetfactory.h
```diff
@@ -197,6 +197,11 @@ class WaveformWidgetFactory : public QObject, public Singleton<WaveformWidgetFac
     void setOverviewNormalized(bool normalize);
     int isOverviewNormalized() const { return m_overviewNormalized;}
 
+    void setOverviewMinuteMarkers(bool minuteMarkers);
+    int isOverviewMinuteMarkers() const {
+        return m_overviewMinuteMarkers;
+    }
```
Swiftb0y, line 203:
```suggestion
    void setDrawOverviewMinuteMarkers(bool draw);
    int drawOverviewMinuteMarkers() const {
        return m_drawOverviewMinuteMarkers;
    }
```

src/waveform/waveformwidgetfactory.cpp
```diff
@@ -687,6 +699,14 @@ void WaveformWidgetFactory::setOverviewNormalized(bool normalize) {
     }
 }
 
+void WaveformWidgetFactory::setOverviewMinuteMarkers(bool minuteMarkers) {
+    m_overviewMinuteMarkers = minuteMarkers;
+    if (m_config) {
+        m_config->set(ConfigKey("[Waveform]", "OverviewMinuteMarkers"),
+                ConfigValue(m_overviewMinuteMarkers));
+    }
```
Swiftb0y, line 707:
```suggestion
        m_config->setValue(ConfigKey("[Waveform]", "OverviewMinuteMarkers"),
                m_drawOverviewMinuteMarkers);
    }
```

Swiftb0y (commented):
Thank you. A couple comments.

src/widget/woverview.cpp
ronso0: we don't need a marker at the end if that is 60*N, so it would be
`for (double currentMarkerSeconds = 60; currentMarkerSeconds < trackSeconds; currentMarkerSeconds += 60) {`

Swiftb0y: fixed

ChrisTallon:
Thanks for the attention so far. It feels like this is on track.

@Swiftb0y I believe I have incorporated all your suggestions manually. I didn't use the one-click commit as it would have left an unbuildable commit.

I think there are two things left - I want to investigate @ronso0 's comment about creating ControlObject / ControlProxy objects. I didn't appreciate WaveformWidgetFactory only dealt with the scrolling waveforms - it doesn't make sense for the Overview minute markers code to be in there.

Also I want to see how to trigger a WWaveformOverview GUI paint when clicking apply / OK on the preferences dialog. It seems like an annoying loose end that if nothing is happening the minute marks don't get painted. This could be confusing for someone looking to see what that option changed.

ronso0:
> Also I want to see how to trigger a WWaveformOverview GUI paint when clicking apply / OK on the preferences dialog

This instant update would be one of the benefits of the ControlObject + ControlProxy. You just need to find a way to trigger a repaint if the control value changes. Maybe just `update();`?

ronso0:
> WaveformWidgetFactory only deals with the scrolling waveforms.

Ahmm, I was wrong: it already provides the overview waveform data and handles the `[Waveform], OverviewNormalized` setting.
So actually, the alternative to the ControlProxy would be WaveformWidgetFactory emitting a `minuteMarkersEnabled(bool)` signal, and connect to that in `WOverview::setup`.

ChrisTallon:
So which should I go for... Are you moving towards a ControlProxy architecture and so that is desirable, or do I just do the WaveformWidgetFactory solution?

ronso0:
I think the WaveformWidgetFactory signal solution is easier to implement.

First I preferred to keep the overview stuff out of WaveformWidgetFactory, but actually (and since it already handles overview settings) the opposite might be better: consolidate all waveform-related settings (main and overview) in the WaveformWidgetFactory singleton, which would result in one place to maintain all that (like get rid of the overview type proxy).

@Swiftb0y wdyt?

Swiftb0y:
not sure tbh. WaveformWidgetFactory is one ugly class, idk if we want that to get further out of control. I would prefer to put it somewhere else if there is a better alternative.

ChrisTallon:
Contributor agreement signed.

I have switched the config to ControlObject/Proxy. This PR now doesn't use WaveformWidgetFactory at all.

As far as I know this is ready for a final(?) review.

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -49,6 +49,11 @@ DlgPrefWaveform::DlgPrefWaveform(
         defaultZoomComboBox->addItem(QString::number(100 / static_cast<double>(i), 'f', 1) + " %");
     }
 
+    m_pOverviewMinuteMarkersControl = std::make_unique<ControlObject>(
+            ConfigKey(QStringLiteral("[Waveform]"),
+                    QStringLiteral("OverviewMinuteMarkers")));
```
Swiftb0y, line 54: I'm sorry for the naming nitpicks, but I think we should add a verb here, also making it consistent with the variable identifiers is nicer. 
```suggestion
    m_pOverviewMinuteMarkersControl = std::make_unique<ControlObject>(
            ConfigKey(QStringLiteral("[Waveform]"),
                    QStringLiteral("DrawOverviewMinuteMarkers")));
```

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -334,6 +343,9 @@ void DlgPrefWaveform::slotResetToDefaults() {
     // Don't normalize overview.
     normalizeOverviewCheckBox->setChecked(false);
 
+    // Don't show minute markers.
+    overviewMinuteMarkersCheckBox->setChecked(false);
+
```
Swiftb0y, line 348: I think this is a nice feature we should enable by default. If users don't like it, they can disable it again.

Swiftb0y:
Please [set up pre-commit](https://pre-commit.com/#installation) and make sure to manually apply the required changes using `pre-commit run --from-ref main --to-ref HEAD`. Also look at the other CI failures and address those. Thank you.

ChrisTallon:
Sorry, I set up a new git tree earlier today and forgot to install pre-commit.  I fixed it for this CI run but I'll look out for any more fails.

ChrisTallon:
I can't get my compiler to generate those same errors so I'm fixing in the blind now...

Swiftb0y:
> I can't get my compiler to generate those same errors so I'm fixing in the blind now...

Yeah... that's usually the case when fixing CI issues with foreign compilers. Just try, push and if that didn't work amend and force push until you have a commit that works.

ChrisTallon:
Hi, what happens next?

Swiftb0y:
Hey there, thank you for reminding me of this. Can you do me a favor and squash your review commits down? Then it be good if someone could test this (idk when I'll find time) and then we can merge. I may also do another review round to make sure I didn't miss anything.

Swiftb0y (approved):

ChrisTallon:
Commits are squashed. Thanks!

Swiftb0y:
Thanks. @ronso0 do you have time for a smoketest by any chance?

ronso0:
This looks good, I'm also fine with white as default marker color.
There are some marker offsets with long tracks, probably due to rounding
image

In vertical mode, the right markers are drawn somewhere on the overview, should be fixed by
6ed9a1ae8f
feel free to adjust the var names. maybe `length` for the longitudinal direction and `width` for the amplitude direction?

commit with  vertical waveforms (dirty):
1507f35b27

The relative length for the markers makes sense, but for tall (vertical: wide) overviews the markers get unnecesarily long IMO.
Shall we limit the length? Or would that limit the skin designers freedom for let's say a skin with huge overviews where long markers are desired?

Swiftb0y:
Lets just get it right without adding the customization complexity yet. The vertical markers were untested anyways so I'm not surprised they're wrong. @ChrisTallon can you pull in the vertical waveform commit ronso0 linked?

ChrisTallon:
https://github.com/mixxxdj/mixxx/commit/6ed9a1ae8febde99933fd01a733db73ef38e1df9 is committed.

Swiftb0y (approved):
LGTM. Thank you.

ChrisTallon:
Mixxx is awesome.

JoergAtGithub:
> Mixxx is awesome.

And you just made it a bit more awesome!

src/preferences/dialog/dlgprefwaveform.cpp
```diff
@@ -49,6 +49,11 @@ DlgPrefWaveform::DlgPrefWaveform(
         defaultZoomComboBox->addItem(QString::number(100 / static_cast<double>(i), 'f', 1) + " %");
     }
 
+    m_pOverviewMinuteMarkersControl = std::make_unique<ControlObject>(
+            ConfigKey(QStringLiteral("[Waveform]"),
+                    QStringLiteral("DrawOverviewMinuteMarkers")));
```
acolombier, line 54: I've noticed a bit late that this PR is introducing new settings using `PascalCase` instead of `snake_case`. Is that something that we can fix before the next release?

Swiftb0y: I don't know if we settled on a convention, but if we want choose one consistently, we should (and can) do it before a release. Though we can only changes keys that have not yet ended up in settings of stable releases.

PR #13404, opened 2024-06-25
feat(controllers): add controller settings for the Numark Scratch

alhadebe:
add controller settings for the Numark Scratch , inspired by the [s4mk3](https://github.com/mixxxdj/mixxx/pull/12995#event-12304815386)

image

### DIFF
--- a/res/controllers/Numark-Scratch-scripts.js
+++ b/res/controllers/Numark-Scratch-scripts.js
@@ -3,35 +3,34 @@ var NumarkScratch = {};
 
 /*
  * USER CONFIGURABLE SETTINGS.
+ * Change settings in the preferences
  */
 
 // Defines the Beatloop Roll sizes for the 4 pads, for available values see:
-// https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-beatlooproll_X_activate
+// https://manual.mixxx.org/latest/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-beatlooproll_X_activate
 // Default: [0.25, 0.5, 1, 2]
 NumarkScratch.autoLoopSizes = [
-    "0.25",
-    "0.5",
-    "1",
-    "2",
+    engine.getSetting("beatLoopRollsSize1") || 0.25,
+    engine.getSetting("beatLoopRollsSize2") || 0.5,
+    engine.getSetting("beatLoopRollsSize3") || 1,
+    engine.getSetting("beatLoopRollsSize4") || 2,
 ];
 
 // Defines how the Loop Encoder functions.
 // If 'true' the Encoder scrolls the library/loads track. Shift + Encoder manages looping.
 // If 'false' the Encoder manages looping. Shift + Encoder scrolls the library/loads track (Serato default).
 // Default: false
-NumarkScratch.invertLoopEncoderFunction = false;
+NumarkScratch.invertLoopEncoderFunction = !!engine.getSetting("invertLoopEncoderFunction");
 
-// Defines the bightness of button LEDs when they inactive.
-// '0x00' sets the LEDSs to completely off. '0x01 sets the LEDs to dim.
-// Default: 0x01 (dim)
-NumarkScratch.LOW_LIGHT = 0x01;
+// Define whether or not to keep LEDs dimmed if they are inactive.
+// 0x01 ('true' in UI) will keep them dimmed, 0x00 ('false' in UI) will turn them off. Default: 0x01 ('true')
+NumarkScratch.noLight = 0x00;
+NumarkScratch.dimLight = 0x01;
+components.Button.prototype.off = engine.getSetting("inactiveLightsAlwaysBacklit") ? NumarkScratch.dimLight : NumarkScratch.noLight;
 
 /*
  * CODE
  */
-
-
-components.Button.prototype.off = NumarkScratch.LOW_LIGHT;
 
 NumarkScratch.init = function() {
     // Initialize component containers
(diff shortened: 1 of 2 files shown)


Swiftb0y (approved):
LGTM. I don't have much experience with the settings though. Do you spot anything wrong @acolombier?

res/controllers/Numark-Scratch-scripts.js
```diff
@@ -3,36 +3,33 @@ var NumarkScratch = {};
...
-NumarkScratch.invertLoopEncoderFunction = false;
+NumarkScratch.invertLoopEncoderFunction = !!engine.getSetting("invertLoopEncoderFunction");
 
-// Defines the bightness of button LEDs when they inactive.
-// '0x00' sets the LEDSs to completely off. '0x01 sets the LEDs to dim.
-// Default: 0x01 (dim)
-NumarkScratch.LOW_LIGHT = 0x01;
+// Define whether or not to keep LEDs dimmed if they are inactive.
+// 'true' will keep them dimmed, 'false' will turn them off. Default: true
+components.Button.prototype.off = !!engine.getSetting("inactiveLightsAlwaysBacklit");
```
acolombier, line 27: Is it safe to change the type of `components.Button.prototype.off` here? You might want to consider a ternary otherwise 

```suggestion
components.Button.prototype.off = !!engine.getSetting("inactiveLightsAlwaysBacklit") ? 0x1 : 0x0;
```

(Worth keeping `0x0` and `0x1` as constant, eg `NumarkScratch.NO_LIGHT` and `NumarkScratch.LOW_LIGHT` as well for readability)

res/controllers/Numark-Scratch.midi.xml
```diff
@@ -1,12 +1,102 @@
 <?xml version='1.0' encoding='utf-8'?>
-<MixxxControllerPreset schemaVersion="1" mixxxVersion="2.3.0+">
+<MixxxControllerPreset schemaVersion="1" mixxxVersion="2.4.0+">
```
acolombier, line 2: Should be `2.5.0` (I believe it is wrong on the S4 Mk3)

acolombier (commented):
Glad to see this feature being used. Just added some very minor comments, otherwise LGTM

res/controllers/Numark-Scratch.midi.xml
NotYourAverageAl: Done

res/controllers/Numark-Scratch-scripts.js
NotYourAverageAl: Done

res/controllers/Numark-Scratch.midi.xml
```diff
@@ -1,12 +1,102 @@
...
+              <value label="1/32">0.03125</value>
+              <value label="1/16">0.0625</value>
+              <value label="1/8">0.125</value>
+              <value label="1/4" default="true">0.25</value>
+              <value label="1/2">0.5</value>
+              <value label="1">1</value>
+              <value label="2">2</value>
+              <value label="4">4</value>
+              <value label="8">8</value>
+              <description>The first pad.</description>
```
NotYourAverageAl, line 50: @acolombier I noticed you had Halve & Double beatloop roll sizes if your s4mk3 settings. Have you found this useful in practice?
```
<value label="Halve beatloop size">half</value>
<value label="Double beatloop size">double</value>
```

acolombier: I use it all the time personally! :) 
The size loop encoder is great to adjust loop size widely, but if you want precision of a button, e.g to half the loop for a build up, it is very handy.
I guess having the option is always great to the user?

NotYourAverageAl: Done

NotYourAverageAl:
@acolombier this is ready for merge.

@Swiftb0y  can someone please merge the 2.4 manual into 2.5, so I can update the controller manual.

acolombier (changes requested):
Unfortunately, the beatloop value `half` and `double` cannot be used without [some special handling](https://github.com/mixxxdj/mixxx/pull/12995/files#diff-e9d337c18b761e937737c0f3c43f7a68de72a7e91fd64ed755b7f1144bc4d860R825), since `beatlooproll_half_activate` and `beatlooproll_double_activate` don't exist (and wouldn't make sense)

The handling should be fairly easy to add [here](https://github.com/mixxxdj/mixxx/blob/2f97bfad4580cf1df19b21ededfd0401cb9d73cb/res/controllers/Numark-Scratch-scripts.js#L370)

res/controllers/Numark-Scratch.midi.xml
Swiftb0y: >  I guess having the option is always great to the user?

Fyi this has often been a point of debate, on the one hand I fully agree (especially because Mixxx's motto is "dj your way") on the other hand, lots of options may overwhelm the user. So every option needs some thought, not only whether its needed but also what is a sensible and consistent default if its needed.

Swiftb0y:
> @Swiftb0y can someone please merge the 2.4 manual into 2.5, so I can update the controller manual.

Done ;)

res/controllers/Numark-Scratch.midi.xml
acolombier: > lots of options may overwhelm the user.

Of course, UX should always be considered, which certainly has been lacking over the past years (these, waveforms, ...).

NotYourAverageAl:
@acolombier I reverted my last commit. Thinking about this more the mixer has only 4 pads instead of 8 and using up 2 pads might not be ideal.

Let's merge as is and ill investigate more if it's worth it when I'm back from holidays.

NotYourAverageAl:
Oops clicked the dreaded comment and close by mistake.

Swiftb0y:
Great. Can you do me a favor and rebase out the reverted commit so it doesn't unnecessarily end up in the commit history?

NotYourAverageAl:
@Swiftb0y done and squashed. Manual PR is [681](https://github.com/mixxxdj/manual/pull/681)

Swiftb0y (approved):
Thank you. LGTM.

PR #13411, opened 2024-06-27
Improve use of parented_ptr

daschuer:
This PR fixed some cases where a patented_ptr is constructed without a parent. This creates a pointer that is temporary without any owner. 

I have moved the assertion to the constructor to avoid this situation in future. 

The risk is that we with debug assertion enabled, Mixxx will crash in case I have missed a case. So a double check during review is required. 

This is part of the pending discussion about the future of parented_ptr in: https://github.com/mixxxdj/mixxx/pull/13395

### DIFF
--- a/src/util/parented_ptr.h
+++ b/src/util/parented_ptr.h
@@ -29,17 +29,12 @@ class parented_ptr final {
 
     explicit parented_ptr(T* t) noexcept
             : m_ptr{t} {
-    }
-// Only generate destructor if not empty, otherwise its empty but will
-// cause the parented_ptr to be not trivially destructible even though it could be.
-#ifdef MIXXX_DEBUG_ASSERTIONS_ENABLED
-    ~parented_ptr() noexcept {
         DEBUG_ASSERT(!m_ptr || static_cast<const QObject*>(m_ptr)->parent());
     }
-#else
+
     // explicitly generate trivial destructor (since decltype(m_ptr) is not a class type)
     ~parented_ptr() noexcept = default;
-#endif
+
     // Rule of 5
     parented_ptr(const parented_ptr<T>&) = delete;
     parented_ptr& operator=(const parented_ptr<T>&) = delete;
--- a/src/util/parented_ptr.h
+++ b/src/util/parented_ptr.h
@@ -99,7 +99,11 @@ class parented_ptr final {
     friend class parented_ptr;
 };
 
+template<typename... Args>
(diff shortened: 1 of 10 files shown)


Swiftb0y:
> I have moved the assertion to the constructor to avoid this situation in future.

Wdyt about adding some compile time checks to `make_parented`? This code will fail if parented_ptr without any QObject being called (if concepts don't work in all the compilers, you can also replace it with `constexpr bool` and `static_assert`). I know its a bit liberal in the sense that it can't know whether the passed object is actually the parent, but it can know that the code is definitely wrong if none of the parameters is a QObject.

```
template<typename... Args>
concept AnyQObject = (... || std::is_base_of_v<QObject, Args>);

template<typename T, typename... Args> 
T make_parented(Args...) requires (AnyQObject<Args...>)
```

Swiftb0y:
whoops. sorry its 
```cpp
template<typename... Args>
constexpr bool AnyQObject = (... || std::is_base_of_v<QObject, std::remove_pointer_t<Args>>);
```

src/util/parented_ptr.h
```diff
@@ -100,10 +100,10 @@ class parented_ptr final {
 };
 
 template<typename... Args>
-concept AnyIsQObject = (... || std::is_base_of_v<QObject, Args>);
+concept AnyIsQObject = (... || std::same_as<QObject*, Args>);
```
Swiftb0y, line 103: that won't work almost always because we never pass a literal QObject...
Please try my previous suggestion instead.

src/util/parented_ptr.h
```diff
@@ -104,7 +99,11 @@ class parented_ptr final {
     friend class parented_ptr;
 };
 
+template<typename... Args>
+concept AnyIsQObject = (... || std::is_base_of_v<QObject, std::remove_pointer_t<Args>>);
```
Swiftb0y, line 103: welp. lets try again
```suggestion
concept AnyIsQObject = (... || std::is_base_of_v<QObject, std::remove_pointer_t<std::remove_cvref_t<Args>>>);
```

Swiftb0y (commented):
if you prefer I can also fix it locally and issue a PR to your branch once I got it working.

Swiftb0y:
ah nice, the static check actually caught something

daschuer:
All green finally. Thank you for your perfect idea with the compile time check.

Swiftb0y (approved):
LGTM. thank you.

PR #13419, opened 2024-06-30
Translations source copy check

daschuer:
When a source string is copied to translation it is listed as translated=yes and will be likely not become translated later. 
This is the issue, this PR tries to fix. 

On the other hand there are strings like "1/4" which are the same in all languages or terms that are by luck the same as the target language.  

It is solved by checking all new translations for source == translation. If this is the case an allow list is consulted and than the commit is rejected. This need to be fixed at Transiflex or if it is a valid untranslated string the allow list has to be maintained. 

The last days I have used the script to put Transiflex into a good shape. But that was a really tedious work. Especially because these bogus translations pop up again from the translation memory when not explicit deleted for each language. 
This script will hopefully prevent future faults.


.pre-commit-config.yaml
```diff
@@ -170,3 +171,15 @@ repos:
           - Markdown==3.4.1
         types: [text]
         files: ^(CHANGELOG\.md|res/linux/org\.mixxx\.Mixxx\.metainfo.xml)$
+      - id: ts-source-copy
+        name: ts-source-copy
+        description: "Find translations literaly copied form the source language"
```
ronso0, line 176:
```suggestion
        description: "Find translations literaly copied from the source language"
```

daschuer:
Now I am finished to clean up the translations form false source copies. The resulting source_copy_allow_list.xml is up to date with all allowed source copies as far as I was able to check it.

res/translations/source_copy_allow_list.xml
```diff
@@ -0,0 +1,3755 @@
...
+    <source>16 bits</source>
+    <allow_all_languages>true</allow_all_languages>
+  </message>
+  <message>
+    <source>24 bits</source>
+    <allow_all_languages>true</allow_all_languages>
+  </message>
+  <message>
+    <source>Param EQ</source>
+    <allow_all_languages>true</allow_all_languages>
```
Swiftb0y, line 874: questionable

res/translations/source_copy_allow_list.xml
```diff
@@ -0,0 +1,3755 @@
...
+    <source>BQ EQ</source>
+    <allow_all_languages>true</allow_all_languages>
+  </message>
+  <message>
+    <source>BQ EQ/ISO</source>
+    <allow_all_languages>true</allow_all_languages>
+  </message>
+  <message>
+    <source>Loudness</source>
+    <allow_all_languages>true</allow_all_languages>
```
Swiftb0y, line 862: questionable

res/translations/source_copy_allow_list.xml
Swiftb0y, line 1: I'm not sure I like this allowlist approach. There are many strings which are debatable and having to maintain this giant list is not great either.

res/translations/source_copy_allow_list.xml
daschuer: What do you mean? Any suggestion?

res/translations/source_copy_allow_list.xml
daschuer: For my understanding the reasons why xml is painful to review is if a tool restructures it. This should not happen here, because additional texts are appended. 
I have picked XML, because the ts files are also XML. 
Any suggestions? 

@Swiftb0y what could be the alternative?

daschuer:
Yes, it is a manual think from a user. If they are too lazy and like to gain some more percentage they seems to just copy the source strings.

res/translations/source_copy_allow_list.xml
Holzhaus: If we really want an allow list (not sure if we really want this), let's use a plain text file with one source per line, followed by a tab character and then a comma separated list of fnmatch expressions:

```
Phase allowed for en and German variants    en,de*
This is allowed for all languages.    *
```

That would be much less verbose than a huge XML.

Holzhaus:
In that case I'm questioning if we really want to check it on pull/during committing.

Can this somehow be prevented on transifex? Or maybe a monthly check which opens a github issue if necessary?

.pre-commit-config.yaml
```diff
@@ -64,7 +64,7 @@ repos:
             "\\W(?:m_p*(?=[A-Z])|m_(?=\\w)|pp*(?=[A-Z])|k(?=[A-Z])|s_(?=\\w))",
             --write-changes,
           ]
-        exclude: ^(packaging/wix/LICENSE.rtf.in|src/dialog/dlgabout\.cpp|.*\.(?:pot?|(?<!d\.)ts|wxl|svg))$
+        exclude: ^(packaging/wix/LICENSE.rtf.in|src/dialog/dlgabout\.cpp|.*\.(?:pot?|ts|wxl|svg))$
```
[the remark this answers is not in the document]
daschuer: all ts files are excluded but not "d.ts" files so it should be OK

res/translations/source_copy_allow_list.xml
daschuer: I have not much interest to write a custom parser. Can we decide for an established format. 

> not sure if we really want this

Can you confirm the issue? Is there a alternative to distinguish wanted form unwanted source copies? 
I have checked the ts format and transifex but there is nothing we can use as a flag.

daschuer:
pre-commit takes care that the check is only done if one is committing changes to the ts files. This is the right moment to reject false translations. 

> Because you cannot really fix the commit locally anyway. Instead you need to go to transifex and remove translations and the perform a fresh tx pull, or am I misunderstanding this?

Correct. 

I think the CI of the new automatic created PR will also fail in that case, which is desired, right?

res/translations/source_copy_allow_list.xml
Holzhaus: Custom parser would be straightforward:

```python
def parse_allowlist(path: pathlib.Path) -> typing.Iterable[tuple[str, list[str]]:
with path.open(mode="r") a fp:
    for line in fp:
        source, _, langstr = line.partition("\t")
        langs = langstr.split(",")
        assert langs
        yield source, langs

allowlist = dict(parse_allowlist("path/to/allowlist")

# Check if (current_source, current_lang) is on allowlist
is_allowed = any(fnmatch.fnmatchcase(current_source, pat) for pat in allowlist.get(current_source))
```

(wrote this on my phone, so it's untested) 

Btw, I can't open the allowlist in the GitHub app on my phone because it is already too large.

daschuer: The allow list has currently 3755 lines with ~4 lines per string we will still almost 2000 lines. Still long. 

Any custom format is not well defined, not extensible, without escape rules. While all these isues are solved with xml because this format is also used for source TS files.

Since the script extends this file automatically, there is no need to build it by hand. We need just confirm new entries which are only new lines, not suffreing any review issues.  

I am not convinced to replace XML.

.pre-commit-config.yaml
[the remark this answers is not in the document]
Swiftb0y: Thats not true. It matches both:
image
You need the negative lookbehind
image

res/translations/source_copy_allow_list.xml
Swiftb0y: Well, in other languages, it may be translated differently. I don't think we can make this blanket statement that its "Param EQ" in all languages. This concern applies to many strings here.

res/translations/source_copy_allow_list.xml
Swiftb0y: if the lack of definition and the unwillingness to write a parser is the problem, just use CSV with two columns... The tree nature of XML is overkill.

Holzhaus: > The allow list has currently 3755 lines with ~4 lines per string we will still almost 2000 lines. Still long.

Where does the 4 lines per string come from? With my proposal it would be 1 line per allowed string (you can list multiple allowed languages in the same line)

@Swiftboy actually the file is already CSV with tab delimiter (or TSV). You could probably already use the stdlib `csv` module with `delimiter='\t'` if there are multiline strings instead of `str.partition`.

TOML would also work (see `tomllib` in the python stdlib), but that would not be much shorter than XML (but way more readable).

.pre-commit-config.yaml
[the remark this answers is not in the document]
daschuer: Ah, I see. Thank you. fixed.

daschuer:
Done.

res/translations/source_copy_allow_list.tsv
```diff
@@ -0,0 +1,436 @@
+lang	source
```
Holzhaus, line 1: If you swap the columns, it's much more readable IMHO.

res/translations/source_copy_allow_list.tsv
```diff
@@ -0,0 +1,436 @@
...
+de,es_CR,es_EC,es_PA,es_UY,it,nl,pt_BR,pt_PT,vi,pl	Hotcues %1-%2
+nl,pl	Hotcue index
+de,it,nl,pl	Hotcue #%1
+fr,de,sv	Information
+fr,es,es_419,es_AR,es_CO,es_CR,es_EC,es_ES,es_MX,es_PA,es_UY	Simple
+ca,es,es_419,es_AR,es_CO,es_CR,es_EC,es_ES,es_MX,es_PA,es_UY	Color
+nl	is
+nl	Cover
+nl	Track BPM: 
+nl	Artist + Title
```
Holzhaus, line 226: Isn't "Titel" the durch word?

res/translations/source_copy_allow_list.tsv
```diff
@@ -0,0 +1,436 @@
...
+sn,sv	Sync
+sn	Deck %1 Quick Effect Super Knob
+sn	Quick Effect
+sq_AL	Quick Links
+sv	Full
+sv	Period
+vi	Shuffle
+vi	Relink
+nl	Lossy
+nl	Lossless
```
Holzhaus, line 256: Isn't this "exact omkeerbaar"?

res/translations/source_copy_allow_list.tsv
```diff
@@ -0,0 +1,436 @@
...
+de	Sound API
+de,it,ja,nl,pt,pt_BR,pt_PT,ro	Deck 1
+de,it,ja,nl,pt,pt_BR,pt_PT,ro	Deck 2
+de,it,ja,nl	Deck 3
+de,it,ja,nl	Deck 4
+de,fr,it,nl,sl	 fps
+de,nl	Decks
+de,nl	Track
+de,nl	Tracks
+de	Add Crate as Track Source
```
Holzhaus, line 434: Nope

tools/ts_source_copy_check.py
```diff
@@ -0,0 +1,205 @@
...
+            if lang is None or lang == "":
+                print(
+                    f"Parse Error: lang is empty at{ALLOW_LIST_PATH}:{i + 1}"
+                )
+                continue
+            source = source.encode("utf-8").decode("unicode_escape")
+            if source == source_text:
+                if lang == "*":
+                    return True
+                if language in lang:
```
Holzhaus, line 45: This means that `en` will also match if the string is `de,en_US,fr` although `en` its not in the list.

```suggestion
                if language in lang.split(","):
```

Even better would be to use `fnmatch` as I suggested before, because if you really want to match all English dialects, you could write `de,en*,fr` without having to list each and every one of them.

tools/ts_source_copy_check.py
```diff
@@ -0,0 +1,205 @@
...
+                    return True
+
+    return False
+
+
+def add_to_allow_list(source_text, language):
```
Holzhaus, line 51: This function basically duplicates the parsing logic from above.

tools/ts_source_copy_check.py
```diff
@@ -0,0 +1,205 @@
...
+
+    ret = 0
+    for changed_file in files_with_added_lines:
+        if check_copied_source_on_lines(rootdir, changed_file):
+            ret = 1
+
+    if ret:
+        print(
+            "\n"
+            "All not allowed copied source translations need to be removed"
```
Holzhaus, line 192:
```suggestion
            "All disallowed copied source translations need to be removed"
```

tools/ts_source_copy_check.py
```diff
@@ -0,0 +1,205 @@
...
+    )
+
+    ret = 0
+    for changed_file in files_with_added_lines:
+        if check_copied_source_on_lines(rootdir, changed_file):
+            ret = 1
+
+    if ret:
+        print(
+            "\n"
```
Holzhaus, line 191:
```suggestion
```

res/translations/source_copy_allow_list.tsv
daschuer: Unfortunately not. This was my first version and I have swapped columns to have languages aligned.

res/translations/source_copy_allow_list.tsv
daschuer: Probably yes. I will remove the translation.

res/translations/source_copy_allow_list.tsv
daschuer: fixed

tools/ts_source_copy_check.py
daschuer: This is desired to have a distance to the individual complains.

daschuer:
Done

.pre-commit-config.yaml
```diff
@@ -43,10 +43,10 @@ repos:
       - id: check-yaml
         exclude: ^\.clang-format$
       - id: end-of-file-fixer
-        exclude: ^.*UTF-8-BOM.txt$
+        exclude: ^.*(UTF-8-BOM\.txt|\.ts)$
```
Holzhaus, line 46:
```suggestion
        exclude: ^(res/translations/.*\.ts|.*UTF-8-BOM\.txt)$
```

Otherwise this will also exclude the typescript files in `res/controllers`.

.pre-commit-config.yaml
```diff
@@ -64,7 +64,7 @@ repos:
             "\\W(?:m_p*(?=[A-Z])|m_(?=\\w)|pp*(?=[A-Z])|k(?=[A-Z])|s_(?=\\w))",
             --write-changes,
           ]
-        exclude: ^(packaging/wix/LICENSE.rtf.in|src/dialog/dlgabout\.cpp|.*\.(?:pot?|(?<!d\.)ts|wxl|svg))$
+        exclude: ^(packaging\/wix\/LICENSE.rtf.in|src\/dialog\/dlgabout\.cpp|.*\.(?:pot?|(?<!\.d\.)ts|wxl|svg))$
```
Holzhaus, line 67: why are the slashes escaped? Is this needed for the regex? this is not shown in the example: https://pre-commit.com/#regular-expressions

.pre-commit-config.yaml
```diff
@@ -81,6 +81,7 @@ repos:
           - "@typescript-eslint/eslint-plugin"
           - "@typescript-eslint/parser"
           - eslint-plugin-diff@^2.0.3
+        exclude: ^.*\.(?<!\.d\.)ts$
```
Holzhaus, line 84: i think if you just exclude the ts files in res/translations it would be much clearer

.pre-commit-config.yaml
Holzhaus:
```suggestion
        description: "Find translations literally copied from the source language"
```

.pre-commit-config.yaml
```diff
@@ -170,3 +171,15 @@ repos:
           - Markdown==3.4.1
         types: [text]
         files: ^(CHANGELOG\.md|res/linux/org\.mixxx\.Mixxx\.metainfo.xml)$
+      - id: ts-source-copy
+        name: ts-source-copy
+        description: "Find translations literaly copied form the source language"
+        entry: python tools/ts_source_copy_check.py
+        stages:
+          - commit
+          - manual
```
Holzhaus, line 180: The stage is called `pre-commit` nowadays, but let's just remove it. I see no reason why using it should be forbidden to use as a pre-push hook for example.

```suggestion
```

.pre-commit-config.yaml
```diff
@@ -170,3 +171,15 @@ repos:
...
+        require_serial: true
+        language: python
+        additional_dependencies:
+          - lxml==4.9.3
```
Holzhaus, line 184:
```suggestion
          - lxml==5.3.0
```

res/translations/source_copy_allow_list.tsv
```diff
@@ -0,0 +1,431 @@
...
+hu,et,fi,nl,pl,pt_BR,ro,zh*	Bm
+hu,et,fi,nl,pl,pt_BR,ro,sl,vi,zh,zh_CN,zh_HK	Cm
+hu,et,fi,nl,pl,pt_BR,ro,sl,zh_CN	C#m
+hu,et,fi,nl,pl,pt_BR,ro,sl,zh*	Dm
+hu,et,fi,nl,pl,pt_BR,ro,sl,zh,zh_CN,zh_HK	Ebm
+hu,de,et,fi,nl,pl,pt_BR,ro,sl,vi,zh*	Em
+hu,et,fi,nl,pl,pt_BR,ro,sl,zh_CN	Fm
+hu,et,fi,nl,pl,pt_BR,ro,sl,zh,zh_CN,zh_HK	F#m
+hu,et,fi,nl,pl,pt_BR,ro,sl,zh,zh_CN,zh_HK	Gm
+hu,et,fi,nl,pl,pt_BR,ro,sl,zh,zh_CN,zh_HK	G#m
```
Holzhaus, line 204: Is this really translatable in any language? I though this was universal.

Holzhaus (commented):
Some suggestions. Regarding the script itself, I'm okay with merging it to solve some pressing issues.

In the long run I'd rather move all the hooks out of this repo to some dedicated `mixxxdj/pre-commit-hooks` repo so that we can update them independently.

res/translations/source_copy_allow_list.tsv
daschuer: I have filed a bug: 
https://github.com/mixxxdj/mixxx/issues/13954

daschuer:
Thank you.

PR #13422, opened 2024-06-30
fix: invalid slip render marker

acolombier:
Looks like the `playposition` was missing in the marker render, likely following a conflict resolution when working on the waveform clean up


Swiftb0y:
What was the bug this is fixing?

acolombier:
Waveform markers use to be display relative to the play position and the not the slip position in the slip half

Swiftb0y:
Ah yeah, I see.

Swiftb0y (approved):
LGTM. Thank you

PR #13423, opened 2024-06-30
Removed unused setSVG and hash functionality from pixmapsource

JoergAtGithub:
Instead use the getPath explicit as in key
Improved cache key string concatenation

### DIFF
--- a/src/widget/wimagestore.h
+++ b/src/widget/wimagestore.h
@@ -12,9 +12,7 @@ struct ImageKey {
     QString path;
     double scaleFactor;
 
-    bool operator==(const ImageKey& other) const {
-        return path == other.path && scaleFactor == other.scaleFactor;
-    }
+    bool operator==(const ImageKey& other) const = default;
 };
 
 namespace std {
--- a/src/widget/wimagestore.h
+++ b/src/widget/wimagestore.h
@@ -8,6 +8,24 @@
 class QImage;
 class ImgSource;
 
+struct ImageKey {
+    QString path;
+    double scaleFactor;
+
+    bool operator==(const ImageKey& other) const {
+        return path == other.path && scaleFactor == other.scaleFactor;
+    }
+};
+
+namespace std {
+template<>
+struct hash<ImageKey> {
+    std::size_t operator()(const ImageKey& key) const {
+        return qHash(key.path) ^ std::hash<double>()(key.scaleFactor);
+    }
+};
+} // namespace std
+
 class WImageStore {
   public:
     static std::shared_ptr<QImage> getImage(const QString& fileName, double scaleFactor);
@@ -23,6 +41,6 @@ class WImageStore {
     static void deleteImage(QImage* p);
 
     // Dictionary of Images already instantiated
-    static QHash<QString, std::weak_ptr<QImage>> m_dictionary;
+    static QHash<ImageKey, std::weak_ptr<QImage>> m_dictionary;
     static std::shared_ptr<ImgSource> m_loader;
 };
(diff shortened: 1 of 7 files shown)


src/widget/wpixmapstore.cpp
```diff
@@ -13,16 +14,19 @@ PaintablePointer WPixmapStore::getPaintable(const PixmapSource& source,
...
         return PaintablePointer();
     }
-    QString key = QStringLiteral("%1%2%3").arg(source.getId(),
-            QString::number(mode),
-            QString::number(scaleFactor));
+    // Generate a key like: 'skins:LateNight/palemoon/buttons/btn___active.svg|STRETCH|1'
+    QString key = QString("%1|%2|%3")
+                          .arg(source.getPath(),
+                                  Paintable::DrawModeToString(mode),
+                                  QString::number(scaleFactor, 'g'));
```
Swiftb0y, line 21: not a requirement for this PR, but a future optimization opportunity would be to avoid the QString entirely. Instead create a new struct that contains all the required data in raw binary form and then implement `qHash` (or whatever interface QHash needs for custom types). That would avoid the 4 temporary strings here. Does that make sense?

JoergAtGithub: As the path is already a string and the other components are tiny, I expect that this is already a cheap operation.

Swiftb0y: Qt does no small-string-optimization, so it involves at least an atomic refcount increase, 1-2 string allocations as well as an additional one for the formatstring assembly. Then that string still needs to get hashed. Everything except the hashing could be avoided. If the path is not hot, then it probably doesn't make any difference, you're right in that regard. Still, this feels like a simple-enough optimization.

I will try to benchmark this in the future.

src/widget/wimagestore.h
```diff
@@ -8,6 +8,24 @@
 class QImage;
 class ImgSource;
 
+struct ImageKey {
+    QString path;
+    double scaleFactor;
+
+    bool operator==(const ImageKey& other) const {
+        return path == other.path && scaleFactor == other.scaleFactor;
+    }
```
Swiftb0y, line 17:
```suggestion
    bool operator==(const ImageKey& other) const = default;
```

Swiftb0y (commented):
Thanks for the (likely) improvement.

Swiftb0y:
I haven't read the imagestore code too closely. Does it cache the data persistently across mixxx restarts? If thats the case, it would likely annoy people during development when they adjust and image but its not updated because the old data still lingers in the cache.

JoergAtGithub:
This behavior is not changed. And this is one of the reasons why I clean up this code step by step. Sometimes Mixxx starts to load SVG again and the UI stucks, see #12904

Swiftb0y:
> This behavior is not changed.

Changing the cachekey to be based on the SVG content to just the file path does change when the cache is invalidated. Are you sure that the behavior is unchanged?

JoergAtGithub:
> Changing the cachekey from being based on the SVG content to being based on the file path does change when the cache is invalidated. Are you sure that the behavior is unchanged?

There were 2 ways to generate the key:

1. Using the the file path of the SVG
2. Hashing the SVG content

The code path that hashed the SVG content was not used (only setSVG could write m_svgSourceData and I there was no caller for this function). therefore I deleted the getKey function in this PR:
https://github.com/mixxxdj/mixxx/blob/6d23434f6f1cd7613ed985f1f9d336dc8652f2f2/src/skin/legacy/pixmapsource.cpp#L45-L51

Instead of getKey I use the getPath function, which is the same as the used code path of getKey:
https://github.com/mixxxdj/mixxx/blob/6d23434f6f1cd7613ed985f1f9d336dc8652f2f2/src/skin/legacy/pixmapsource.cpp#L24-L26

Swiftb0y (approved):
I see. Thank you. LGTM

PR #13424, opened 2024-06-30
Made Paintable::DrawMode an enum class

JoergAtGithub:


src/widget/paintable.h
```diff
@@ -16,7 +16,7 @@ class QSvgRenderer;
 // high fidelity.
 class Paintable {
   public:
-    enum DrawMode {
+    enum class DrawMode {
```
Swiftb0y, line 19: if this change results in a touch to every point of use anyways, can you also change the members to TitleCase?

src/widget/paintable.cpp
```diff
@@ -16,35 +16,36 @@
...
 // static
 QString Paintable::DrawModeToString(DrawMode mode) {
     switch (mode) {
-        case FIXED:
-            return "FIXED";
-        case STRETCH:
-            return "STRETCH";
-        case STRETCH_ASPECT:
-            return "STRETCH_ASPECT";
-        case TILE:
```
Swiftb0y, line 43: these could be autogenerated if `DrawMode` was a Q_ENUM. Do you agree? Do we need to provide a stability guarantee for these strings?

src/widget/paintable.h
JoergAtGithub: I noticed, that the upper case strings are used accross the skin XML files. Therefore the string must remain upper case for backward compatibility.
PS: I just added a debug assert, which worked;-)

JoergAtGithub: Now we have to decide, should we switch to Q_ENUM or should we rename to TitelCase?
Since Q_ENUM adds a overhead, I would prefer to rename and implement these two functions otherwise.

Swiftb0y: > Since Q_ENUM adds a overhead

Out of curiosity, in what sense? From what I can tell, it only adds a little bit of compilation time due to the MOC.

JoergAtGithub: Yes, the class must be a QObject and the MOC compilation comes on top.

Swiftb0y: The QObject could be avoided if the enum was moved to a namespace technically. Though I'm fine with leaving it as a plain `enum class` for now. Its unlikely that this enum would need to get exposed to the QML environment anyways (which is where the biggest value of a QEnum lies IMO).

JoergAtGithub: This enum is only used in our legacy widgets. A use with QML is highly unlikely.

Swiftb0y (approved):
LGTM. Thank you.

PR #13425, opened 2024-06-30
Don't return in JogWheelBasic on deck absent in option

christophehenry:
This allows late binding of `JogWheelBasic` using `ComponentContainer.reconnectComponents` and `Deck.setCurrentDeck`.

### DIFF
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -762,16 +762,50 @@
     });
 
     const JogWheelBasic = function(options) {
+        if (options.deck !== undefined && options.group !== undefined) {
+            console.warn(
+                "options.deck and option.group are both set; " +
+                "options.deck will take priority"
+            );
+        }
+
+        const deck = options.deck;
+        const group = script.deckFromGroup(options.group);
+        delete options.deck;
+        delete options.group;
+
         Component.call(this, options);
 
-        if (!Number.isInteger(this.deck)) {
-            console.warn("missing scratch deck");
-            return;
+        this._deck = undefined;
+
+        Object.defineProperties(this, {
+            deck: {
+                get: () => this._deck,
+                set: (value) => {
+                    if (Number.isInteger(value) && value > 0) {
+                        this._deck = value;
+                        this.reset();
+                    }
+                },
+            },
+            group: {
+                get: () => `[Channel${deck}]`,
+                set: value => {
+                    const deck = script.deckFromGroup(value);
+                    if (deck > 0) {
+                        this._deck = deck;
+                        this.reset();
+                    }
+                },
+            }
+        });
+
+        this.deck = deck;
+
+        if (!this.deck) {
+            this.group = group;  // try setting deck from group
         }
-        if (this.deck <= 0) {
-            console.warn("invalid deck number: " + this.deck);
-            return;
-        }
+
         if (!Number.isInteger(this.wheelResolution)) {
             console.warn("missing jogwheel resolution");
             return;
@@ -786,9 +820,7 @@
         if (!Number.isFinite(this.rpm)) {
             this.rpm = 33 + 1/3;
         }
-        if (this.group === undefined) {
-            this.group = "[Channel" + this.deck + "]";
-        }
+
         this.inKey = "jog";
     };
 
@@ -800,6 +832,10 @@
             return value < 0x40 ? value : value - (this.max + 1);
         },
         inputWheel: function(_channel, _control, value, _status, _group) {
+            if (!this.deck) {
+                return;
+            }
+
             value = this.inValueScale(value);
             if (engine.isScratching(this.deck)) {
                 engine.scratchTick(this.deck, value);
@@ -808,6 +844,10 @@
             }
         },
         inputTouch: function(channel, control, value, status, _group) {
+            if (!this.deck) {
+                return;
+            }
+
             if (this.isPress(channel, control, value, status) && this.vinylMode) {
                 engine.scratchEnable(this.deck,
                     this.wheelResolution,
@@ -822,13 +862,7 @@
             throw "Called wrong input handler for " + status + ": " + control + ".\n" +
                 "Please bind jogwheel-related messages to inputWheel and inputTouch!\n";
         },
-        // this is needed for features such as "deck switching" that work
-        // by changing the component group. It is assumed they call `connect`
-        // afterwards.
-        connect: function() {
-            Component.prototype.connect.call(this);
-            this.deck = parseInt(script.channelRegEx.exec(this.group)[1]);
-        }
+        reset() {},
     });
 
     const EffectUnit = function(unitNumbers, allowFocusWhenParametersHidden, colors) {


res/controllers/midi-components-0.0.js
```diff
@@ -765,12 +763,19 @@
...
+            return Number.isInteger(this.deck) && this.deck > 0;
+        },
         isPress: Button.prototype.isPress,
         inValueScale: function(value) {
             // default implementation for converting signed ints
             return value < 0x40 ? value : value - (this.max + 1);
         },
         inputWheel: function(_channel, _control, value, _status, _group) {
+            if (!this.hasDeckNumber) {
+                console.warn("bad deck number; ignoring inputWheel()");
```
Swiftb0y, line 776: now this should actually return now, imo. same below.

res/controllers/midi-components-0.0.js
```diff
@@ -765,12 +763,19 @@
 
     JogWheelBasic.prototype = new Component({
         vinylMode: true,
+        get hasDeckNumber() {
+            return Number.isInteger(this.deck) && this.deck > 0;
+        },
```
Swiftb0y, line 768: we haven't used js getters until now. have you made sure this works? I'm not confident the Object merging logic handles getters and setters correctly.

Swiftb0y (commented):
can you show me some sample code where this late binding is useful/needed?

christophehenry:
> can you show me some sample code where this late binding is useful/needed?

[The current DN-S3700 controller I use](https://github.com/christophehenry/mixxx-Denon-DN-S3700-hybrid/blob/master/Denon-DN-S3700-hybrid-scripts.js#L376-L403) retrieves its deck number from the deck's settings with a sysex message.

Maybe there's another way but this is how it works right now. Plus, no other component requires an `options` parameter during instanciation. [They just check these are set](https://github.com/mixxxdj/mixxx/blob/main/res/controllers/midi-components-0.0.js#L114-L128) so there's an inconsistancy here.

res/controllers/midi-components-0.0.js
```diff
@@ -733,16 +733,52 @@
...
+            },
+            set deck(value) {
+                if (Number.isInteger(value) && value > 0) {
+                    self._deck = value;
+                    self._group = `[Channel${value}]`;
+                }
+            },
+            get group() {
+                return self._deck;
+            },
```
Swiftb0y, line 765:
```suggestion
            get group() {
                return self._group;
            },
```

Swiftb0y:
> [The current DN-S3700 controller I use](https://github.com/christophehenry/mixxx-Denon-DN-S3700-hybrid/blob/master/Denon-DN-S3700-hybrid-scripts.js#L376-L403) retrieves its deck number from the deck's settings with a sysex message.

Thats a very valid argument. Thanks.

res/controllers/midi-components-0.0.js
```diff
@@ -793,13 +841,7 @@
...
                 "Please bind jogwheel-related messages to inputWheel and inputTouch!\n";
         },
-        // this is needed for features such as "deck switching" that work
-        // by changing the component group. It is assumed they call `connect`
-        // afterwards.
-        connect: function() {
-            Component.prototype.connect.call(this);
-            this.deck = parseInt(script.channelRegEx.exec(this.group)[1]);
-        }
+        reset() {},
```
christophehenry, line 844: This method is just there to allow cleaning when the deck has changed (for instance, stop a motorized jog).

christophehenry:
[So `ControllerScriptEngineLegacyTest.commonScriptHasNoErrors`](https://github.com/mixxxdj/mixxx/actions/runs/9828428632/job/27132307604?pr=13425#step:22:3812) has errors but there's a problem here: this should ne segfault. I also met this problem yesterday on Mixxx 2.4.0 while working on my controller. Is it a know bug?

res/controllers/midi-components-0.0.js
christophehenry: You're right. I had to use `Object.defineProperties` instead.

Swiftb0y:
I don't think so. The segfault is definitely weird. Can you [https://github.com/mixxxdj/mixxx/wiki/Creating%20Backtraces](collect a backtrace)?

christophehenry:
Uh. What's going on with the build? I get the same compilation errors here.

Swiftb0y:
faulty merge in d9d23dc238f9cfb8a9e997615fcf576d7971851a. fixed in 9a27ac92757d56bbf566670affa7de090edae8a8. Please rebase and try again.

res/controllers/common-controller-scripts.js
```diff
@@ -248,18 +256,12 @@ script.midiDebug = function(channel, control, value, status, group) {
...
-            deck = group.substring(8,group.length-1);
-        }
-    */
-    return parseInt(deck);
+    try {
+        const deck = Number.parseInt(script.channelRegEx.exec(`${group}`)[1]);
+        return Number.isNaN(deck) ? 0 : deck;
+    } catch (_) {
+        return 0;
+    }
```
Swiftb0y, line 264: Can you elaborate on why you prefer this over the previous implementation? I don't think [its worth the performance penalty](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJBUHRVQmxwajNNTXFPamZHU1ZEVGgiLCJjb2RlIjoiREFUQS5hcnIubWFwKERBVEEuZGVja0Zyb21Hcm91cFNsb3cpIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6InpBUEZLTnhDcEZQSGVUckowVC0xVyIsImNvZGUiOiJEQVRBLmFyci5tYXAoREFUQS5kZWNrRnJvbUdyb3VwRmFzdCkiLCJkZXBlbmRlbmNpZXMiOltdfV0sImNvbmZpZyI6eyJuYW1lIjoiQmFzaWMgZXhhbXBsZSIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6InJldHVybiB7XG4gIGNoYW5uZWxSZWdFeDogL15cXFtDaGFubmVsKFxcZCspXFxdJC8sXG4gIGFycjogQXJyYXkoMTAwMCkuZmlsbChcIltDaGFubmVsMV1cIiksXG4gIGRlY2tGcm9tR3JvdXBTbG93OiAoZ3JvdXApID0-IHtcbiAgdHJ5IHtcbiAgICAgICAgY29uc3QgZGVjayA9IE51bWJlci5wYXJzZUludChzY3JpcHQuY2hhbm5lbFJlZ0V4LmV4ZWMoYCR7Z3JvdXB9YClbMV0pO1xuICAgICAgICByZXR1cm4gTnVtYmVyLmlzTmFOKGRlY2spID8gMCA6IGRlY2s7XG4gICAgfSBjYXRjaCAoXykge1xuICAgICAgICByZXR1cm4gMDtcbiAgICB9XG59LFxuICBkZWNrRnJvbUdyb3VwRmFzdDogZ3JvdXAgPT4ge1xuICBsZXQgZGVjayA9IDA7XG4gICAgaWYgKGdyb3VwLnN1YnN0cmluZygyLCA4KSA9PT0gXCJoYW5uZWxcIikge1xuICAgICAgICAvLyBFeHRyYWN0IGRlY2sgbnVtYmVyIGZyb20gdGhlIGdyb3VwIHRleHRcbiAgICAgICAgZGVjayA9IGdyb3VwLnN1YnN0cmluZyg4LCBncm91cC5sZW5ndGggLSAxKTtcbiAgICB9XG4gIGRlY2sgPSBwYXJzZUludChkZWNrKTtcbiAgcmV0dXJuIE51bWJlci5pc05hTihkZWNrKSA_IDAgOiBkZWNrO1xufVxufVxuIn19). `deckFromGroup` is called far too often in the "simple" JS mappings so I want this to be fast.

res/controllers/midi-components-0.0.js
```diff
@@ -733,16 +733,53 @@
...
     const JogWheelBasic = function(options) {
+        if (options.deck !== undefined && options.group !== undefined) {
+            console.warn(
+                "options.deck and option.group are both set; " +
+                "options.deck will take priority"
+            );
+        }
+
+        const deck = options.deck;
+        const group = script.deckFromGroup(options.group);
```
Swiftb0y, line 744: group is really the deck here now, right?

res/controllers/midi-components-0.0.js
```diff
@@ -733,16 +733,53 @@
...
+                set: function(value) {
+                    if (Number.isInteger(value) && value > 0) {
+                        this._deck = value;
+                        this._group = `[Channel${value}]`;
+                        this.reset();
+                    }
+                }.bind(this),
+            },
+            "group": {
+                get: function() { return this._group; }.bind(this),
```
Swiftb0y, line 765: I think the `.bind(this)` is unnecessary. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#get

Swiftb0y:
whoops, sorry.

res/controllers/common-controller-scripts.js
christophehenry: [The performance penalty is not that big](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJLUmEtSTFDMS1XdkNPcDFWTUpSYXoiLCJjb2RlIjoiREFUQS5hcnIubWFwKERBVEEuZGVja0Zyb21Hcm91cFNsb3cpIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6IkFQdFVCbHBqM01NcU9qZkdTVkRUaCIsImNvZGUiOiJEQVRBLmFyci5tYXAoREFUQS5kZWNrRnJvbUdyb3VwU2xvdzIpIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6InpBUEZLTnhDcEZQSGVUckowVC0xVyIsImNvZGUiOiJEQVRBLmFyci5tYXAoREFUQS5kZWNrRnJvbUdyb3VwRmFzdCkiLCJkZXBlbmRlbmNpZXMiOltdfV0sImNvbmZpZyI6eyJuYW1lIjoiQmFzaWMgZXhhbXBsZSIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6InJldHVybiB7XG4gIGFycjogQXJyYXkoMTAwMCkuZmlsbChcIltDaGFubmVsMV1cIiksXG4gIGRlY2tGcm9tR3JvdXBTbG93OiAoZ3JvdXApID0-IHtcbiAgdHJ5IHtcbiAgICAgICAgY29uc3QgZGVjayA9IE51bWJlci5wYXJzZUludCgvXlxcW0NoYW5uZWwoXFxkKylcXF0kLy5leGVjKGAke2dyb3VwfWApWzFdKTtcbiAgICAgICAgcmV0dXJuIE51bWJlci5pc05hTihkZWNrKSA_IDAgOiBkZWNrO1xuICAgIH0gY2F0Y2ggKF8pIHtcbiAgICAgICAgcmV0dXJuIDA7XG4gICAgfVxufSxcbiAgZGVja0Zyb21Hcm91cFNsb3cyOiAoZ3JvdXApID0-IHtcbiAgICBjb25zdCBleHRyYWN0ZWQgPSAvXlxcW0NoYW5uZWwoXFxkKylcXF0kLy5leGVjKGAke2dyb3VwfWApO1xuICAgIGlmKGV4dHJhY3RlZCA9PT0gbnVsbCkgeyByZXR1cm4gMDsgfVxuICAgICAgICBjb25zdCBkZWNrID0gTnVtYmVyLnBhcnNlSW50KGV4dHJhY3RlZFsxXSk7XG4gICAgICAgIHJldHVybiBOdW1iZXIuaXNOYU4oZGVjaykgPyAwIDogZGVjaztcbn0sXG4gIGRlY2tGcm9tR3JvdXBGYXN0OiBncm91cCA9PiB7XG4gIGxldCBkZWNrID0gMDtcbiAgICBpZiAoZ3JvdXAuc3Vic3RyaW5nKDIsIDgpID09PSBcImhhbm5lbFwiKSB7XG4gICAgICAgIC8vIEV4dHJhY3QgZGVjayBudW1iZXIgZnJvbSB0aGUgZ3JvdXAgdGV4dFxuICAgICAgICBkZWNrID0gZ3JvdXAuc3Vic3RyaW5nKDgsIGdyb3VwLmxlbmd0aCAtIDEpO1xuICAgIH1cbiAgZGVjayA9IHBhcnNlSW50KGRlY2spO1xuICByZXR1cm4gTnVtYmVyLmlzTmFOKGRlY2spID8gMCA6IGRlY2s7XG59XG59XG4ifX0). `deckFromGroupSlow` had a typo (`script` is `undefined`).

res/controllers/midi-components-0.0.js
christophehenry: Yeah, I wasn't really sure since my IDE still warned me about a potential bad usage of `this`. I can remove.

res/controllers/midi-components-0.0.js
christophehenry: WDYM?

res/controllers/midi-components-0.0.js
Swiftb0y: yeah, or maybe just replace with an arrow function outright? That makes it clear where `this` comes from.

res/controllers/common-controller-scripts.js
Swiftb0y: > deckFromGroupSlow had a typo (script is undefined) making the try to always endup in the catch clause.

good catch (pun intendended ;)

Anyways, [catching is still quite slow](https://jsbenchmark.com/#eyJjYXNlcyI6W3siaWQiOiJLUmEtSTFDMS1XdkNPcDFWTUpSYXoiLCJjb2RlIjoiREFUQS5hcnIubWFwKERBVEEuZGVja0Zyb21Hcm91cFNsb3cpIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6IkFQdFVCbHBqM01NcU9qZkdTVkRUaCIsImNvZGUiOiJEQVRBLmFyci5tYXAoREFUQS5kZWNrRnJvbUdyb3VwU2xvdzIpIiwiZGVwZW5kZW5jaWVzIjpbXX0seyJpZCI6InpBUEZLTnhDcEZQSGVUckowVC0xVyIsImNvZGUiOiJEQVRBLmFyci5tYXAoREFUQS5kZWNrRnJvbUdyb3VwRmFzdCkiLCJkZXBlbmRlbmNpZXMiOltdfV0sImNvbmZpZyI6eyJuYW1lIjoiQmFzaWMgZXhhbXBsZSIsInBhcmFsbGVsIjp0cnVlLCJnbG9iYWxUZXN0Q29uZmlnIjp7ImRlcGVuZGVuY2llcyI6W119LCJkYXRhQ29kZSI6ImNvbnN0IHJlZyA9IC9eXFxbQ2hhbm5lbChcXGQrKVxcXSQvO1xucmV0dXJuIHtcbiAgYXJyOiBBcnJheSgxMDAwKS5maWxsKDApLm1hcCgoKSA9PiBNYXRoLnJhbmRvbSgpIDwgMC41ID8gXCJbQ2hhbm5lbDFdXCIgOiBcImdhcmJhZ2VcIiksXG4gIGRlY2tGcm9tR3JvdXBTbG93OiAoZ3JvdXApID0-IHtcbiAgdHJ5IHtcbiAgICAgICAgY29uc3QgZGVjayA9IE51bWJlci5wYXJzZUludChyZWcuZXhlYyhgJHtncm91cH1gKVsxXSk7XG4gICAgICAgIHJldHVybiBOdW1iZXIuaXNOYU4oZGVjaykgPyAwIDogZGVjaztcbiAgICB9IGNhdGNoIChfKSB7XG4gICAgICAgIHJldHVybiAwO1xuICAgIH1cbn0sXG4gIGRlY2tGcm9tR3JvdXBTbG93MjogKGdyb3VwKSA9PiB7XG4gICAgY29uc3QgZXh0cmFjdGVkID0gcmVnLmV4ZWMoYCR7Z3JvdXB9YCk7XG4gICAgaWYoZXh0cmFjdGVkID09PSBudWxsKSB7IHJldHVybiAwOyB9XG4gICAgICAgIGNvbnN0IGRlY2sgPSBOdW1iZXIucGFyc2VJbnQoZXh0cmFjdGVkWzFdKTtcbiAgICAgICAgcmV0dXJuIE51bWJlci5pc05hTihkZWNrKSA_IDAgOiBkZWNrO1xufSxcbiAgZGVja0Zyb21Hcm91cEZhc3Q6IGdyb3VwID0-IHtcbiAgbGV0IGRlY2sgPSAwO1xuICAgIGlmIChncm91cC5zdWJzdHJpbmcoMiwgOCkgPT09IFwiaGFubmVsXCIpIHtcbiAgICAgICAgLy8gRXh0cmFjdCBkZWNrIG51bWJlciBmcm9tIHRoZSBncm91cCB0ZXh0XG4gICAgICAgIGRlY2sgPSBncm91cC5zdWJzdHJpbmcoOCwgZ3JvdXAubGVuZ3RoIC0gMSk7XG4gICAgfVxuICBkZWNrID0gcGFyc2VJbnQoZGVjayk7XG4gIHJldHVybiBOdW1iZXIuaXNOYU4oZGVjaykgPyAwIDogZGVjaztcbn1cbn1cbiJ9fQ). So can we avoid that?

> Anyway, I find this version more resistent to passing stupid that may cause the controller to crash during a live (like undefined or -1, yes I did that developping my DN-S3700 controller).

Can you elaborate? Couldn't we gain the same resilience without having to reach for the regex?

res/controllers/midi-components-0.0.js
Swiftb0y: well, after assigning to `groupp, `deck == group` should hold (assuming the deck and group is set correctly).

res/controllers/common-controller-scripts.js
christophehenry: > So can we avoid that?

I'm ok with this.

> Can you elaborate? Couldn't we gain the same resilience without having to reach for the regex?

I don't remember exactly. I was probably working on something else and I decided to change that code while I was at it. I often do this kind of things. My development style is not structured at all.

Is there a guaranty that the group form will always be *[Channel1]* and not, for instance "[Channel 1]" or "[Channel 10]"? Generally I prefer regexes in this kind of use case since they tend to be more flexible and I'm not sure the overhead will have any significant impact.

But I'm ok, leaving this code as is. Although I'd still cast `group` to a string using `group = `${group}` to avoid stupid things like `deckFromGroup(undefined)` or `deckFromGroup(null)`.

res/controllers/midi-components-0.0.js
christophehenry: Yes. I wasn't sure a deconnection between deck and group was intentional. The previous implementation allowed to set both deck and group with different values and I'm not sure why. Maybe there's a valid use case? Anyway, yes the group could simply be derived from deck, no need to set the variable.

Swiftb0y: Yup, no deck was essentially just supposed to be a cache for `script.deckFromGroup(group)`. If `deck` can be implemented efficiently in terms of `group` (or vice versa), that would be ideal. The previous implementation was just too lazy (by yours truly :sweat_smile:) to care.

res/controllers/common-controller-scripts.js
Swiftb0y: > My development style is mostly chaotic

No worries ;). Same goes for me. 

> Is there a guaranty that the group form will always be [Channel1] and not, for instance [Channel 1] or [Channel 12]?

Yes. That is pretty much set in stone for as long as CO's will exist.

> not sure the overhead will have any significant impact.

I'm not fully certain either, but its an easy and trivial optimization opportunity, judging by how often its called. 

> Although I'd still cast group to a string using group = `${group}` to avoid stupid things like deckFromGroup(undefined) or deckFromGroup(null).

Not sure I fully agree with that either. IMO nonsense code like that is supposed to crash, not just silently fail because `parseInt("undefined"[8])` is `NaN`. Much better to get a stacktrace right then and there. Wdyt?

christophehenry: Alright, I'm fine with this. I reversed the change.

christophehenry:
Oy, oy! Shall we merge this this week?

Swiftb0y (approved):
LGTM. thank you. Lets see how well the setter/getter stuff works.

PR #13426, opened 2024-07-01
(fix) Library: show Properties action in Missing and Hidden

ronso0:
The Properties dialog is helpful also in Hidden and Missing. You can quickly check stored metadata like album, bitrate etc. of removed tracks. Though the entire Metadata doesn't make much sense there IMO.

Previously, `Capability::EditMetadata` was used to show/hide the metadata menu, star rating (and cover art menu?), that's why I decided to separate that and add `Capability::Properties` only for the properties action.

Wdyt?


daschuer (approved):
LGTM, thank you.

Issue #13403, opened 2024-06-25
Auto-Playlist


joesch2:
### Feature Description

It would be helpful if the system could create dynamic playlists based on song tags. The type of dynamic lists should be configurable.

Example: 
I enable dynamic playlists based on tag "genre". The system displays a folder "Genre” in the archive tree under playlists, in which there is a folder for each genre found that contains the titles that contain the genre as a tag. 

It would also be nice to have an entry “Auto Playlist” in the root directory, under which all automatic playlists-folder are displayed.

daschuer:
Mixxx can already populate the Auto DJ playlist from crates. So you may create crates with the desired genres and make them an Auto DJ track source. 

Sure the used magic formula can be improved all help is welcome.

joesch2:
Thanks for the feedback!

In my use case, tracks newly added to the collection should be automatically added to the corresponding record box at the same time based on the Genre tag. For me it's not about using it with the AutoDJ.

Example: Example: There is a “Walz” and a “Jive” record box. I'm adding three tracks to the collection with "Jive" as the genre tag. The titles are not only visible in the track collection (Tracks folder) but are also automatically visible in the “Jive” box.
This is helpful, for example, in dance sports training when a specific dance needs to be practiced.

joesch2:
Maybe it would be possible to evaluate the name of the record box for this function. If this consists of a valid search term, the box will be filled dynamically.
![grafik](https://github.com/mixxxdj/mixxx/assets/18378305/6f124bde-c4ec-4934-8140-ae32a92b8278)

Eve00000:
I think joesch2 wants something like "smart playlists" in iTunes: playlist that are defined as a result of a query that keeps on being updated.
= kind of pinning the result of a query to a new crate.
This way you could have crates  Springsteen (artist like 'Springsteen'), fast (title like 'fast'), Mozart (composer like 'Mozart'), Jive (genre like 'jive').
In iTunes it's like the more smart playlists there are, the slower iTunes gets. Therefore they made an option 'keep updated" that means you can deactivate the continuous checking and trigger the playlist-update manually.
It would be a nice feature. In iTunes I use this to correct tags, or to list the tracks added in the last month, 100 most played, classical tracks, tracks without artwork...
Executing a query is much faster in Mixxx so it's not hard to put the result of a query in a new/ existing crate (execute query -> select tracks - > right click -> crates -> select the crate or choose new crate). 
[...]

Issue #12283, opened 2023-11-10
Library scanner dialog pops up after the scan finished and never closes


ronso0:
### Bug Description

* start Mixxx with a new profile
* when asked for music directories, select one with only a few tracks (3 in my case)
* roughly when the launch image disappears the scanner dialog shows up, while 'shows up' is a bit of an exageration: neither the progress bar nor the actual dialog background are visible, just the framebuffer (task switcher in this case)
![image](https://github.com/mixxxdj/mixxx/assets/5934199/6c4e634b-54d3-423a-8a37-fd6dfa341b91)
* it won't close itself even though scanning those 3 tracks was finished
(I can close it without issues)

Turns out the dialog should already be gone (added some debug to [LibraryScannerDlg::slotScanFinished](https://github.com/mixxxdj/mixxx/blob/6c5e5cd18cd03f342d2099784a3ee6c034cdb8d2/src/library/scanner/libraryscannerdlg.cpp#L78C6-L78C41)), so is this just a window manager issue?

### Version

2.4-beta

### OS

Ubuntu Studio 20.04 / X11 / xfce 4.14 / Qt 5.12.8

ronso0:
Probably related to #14940

Issue #11738, opened 2023-07-11
DEBUG_ASSERT in SettingsDAO::setValue


ronso0:
### Bug Description

occurs when starting with a new profile, and on every subsequent start
```
warning [Main] SettingsDAO - Failed to set "mixxx.db.model.missing.ColumnsSorting" = QVariant(QString, "artist 1 ") QSqlError("5", "Unable to fetch row", "database is locked")
DEBUG ASSERT: "!"Failed query"" in function bool SettingsDAO::setValue(const QString&, const QVariant&) const at /src/library/dao/settingsdao.cpp:76
```

### Version

2.4-beta 10dba8f4ac6c4e00595869f9e38c0f83da692f6f

### OS

Ubuntu Studio 20.04.5 with xfce4

ronso0:
If I disable the DEBUG_ASSERT (too lazy to rebuild as `Release` or with `-DMIXXX_DEBUG_ASSERTIONS_FATAL` off)...
Mixxx starts just fine and mixxxdb `settings`>`mixxx.db.model.missing.ColumnsSorting` is `artist 1` as it should be :shrug:

ronso0:
100% reproducible when I select a huge music folder during first start.
Active library scanner causing `"database is locked"`?

uklotzde:
#10876 ...and probably more.

All these sporadic or spurious database failures are expected. Not fixable without a major rewrite. The architecture is broken.

Btw, 1.6k issues are ridiculous for such a small project. No one will ever look at them. Instead new issues like this are created over and over again. Advice from an expert: https://twitter.com/allenholub/status/1676367019078488066

ronso0:
so duplicate of #10876, and basically #6631

PR #13427, opened 2024-07-01
Library scan: log summary and show popup

ronso0:
This emphasizes the logged scan summary and allows to show a respective popup, both for automatic scan at startup and manual scan.

Fixes #10720 

For me, this is a step towards being able to manually find successors for moved missing tracks, may also be helpful for the _New Tracks_ filter/feature #13526.

```
info [LibraryScanner 1] ----------------------------------------------
info [LibraryScanner 1] Library scan finished after 963 ms
info [LibraryScanner 1]  1235 unchanged directories
info [LibraryScanner 1]  197 changed/added directories
info [LibraryScanner 1]  0 tracks verified from changed/added directories
info [LibraryScanner 1]  0 new tracks
info [LibraryScanner 1]  0 moved tracks
info [LibraryScanner 1]  0 missing tracks
info [LibraryScanner 1] ----------------------------------------------
```
image
if nothing changed it displays only duration and ""No changes were detected."

TODO
~~Find an alternative for `int LibraryScanner::cleanUpScan()` to get the number of relocated tracks.~~ :heavy_check_mark:


ronso0:
> Find an alternative for int LibraryScanner::cleanUpScan() to get the number of relocated tracks.

Simply storing that in an LibraryScanner int member seems fine to me (reset when scan is started, no need to bother scannerGlobal with this).
Wdyt?

JoergAtGithub:
> show a respective popup, both for automatic scan at startup and manual scan

It's a nice addition to get a feedback, when I triggered the scan amnually. But I would no like to get this pop-up, when I start Mixxx.

ronso0:
Funny thing is, I moved the connection to CoreServices so we can _always_ get the popup.

Why do you consider the info popup disturbing / not helpful for the autorun scan during  startup?

src/library/dao/trackdao.cpp
```diff
@@ -1768,27 +1788,27 @@ void TrackDAO::markUnverifiedTracksAsDeleted() {
...
+            // first mismatch
+            break;
         }
-        return matchLength;
+        ++matchLength;
     }
-    } // namespace
+    return matchLength;
+}
+} // namespace
```
ronso0, line 1804: This is only fixing indentation, pre-commit

ronso0:
friendly ping @JoergAtGithub 
> Why do you consider the info popup disturbing / not helpful for the autorun scan during startup?

We _could_ make the report optional when scanning during startup, but I'd rather avoid that sort of complification / extra pref option.

JoergAtGithub:
In my opinion a dialog should only appear, if:

- it's the response to a users action (like clicking on rescan library here)
- an error occurs

but on startup it is always anoying to click dialogs away. Normal information should be visualize unobtrusive in the UI, in a way that does not require user action to continue.

ronso0:
The last commit disables the report/summary dialog for the auto scan run during startup.

ronso0:
I'm a bit puzzled why the "changed/added" directory count is not 0 repeatedly even if the dirs haven't been modified.
(no regression btw)

ronso0:
> I'm a bit puzzled why the "changed/added" directory count is not 0 repeatedly even if the dirs haven't been modified. (no regression btw)

Turns out it's just the confusing naming scanned vs. changed/added.
"scanned directories" are the directories where no previous hash exists (??) or prev hash != new hash (even though there was absolutely nothing even touching the dirs in between the scans :shrug: )

I adjusted the dialog and the logged lines. All good now!

ronso0:
> Is it the intended behavior, that I get negative values, if I: Delete a file -> Rescan -> Restore the file -> Rescan

Thanks for your thourough testing!
This is not expected, but the numerical result of the TrackDAO/Scanner methods we had at our disposal:
```
  previously existing track locations
+ new tracks (= added tracks - relocated tracks)
- existing track locations
```

In order to reliably report rediscovered tracks I'll add a method to TrackDAO to get missing track locations.

ronso0:
The report now looks like this
image

We _could_ calculate **new** missing tracks but I don't think that's worth it.

ronso0:
> It would make sense to print the scan results into the log output at startup, if Mixxx is started in developer mode.

The report is always logged, just the dialog is now suppressed for the scan during startup.

daschuer:
I have not tested this yet and I have scanning during start up disabled so I probably not fully qualified to comment. But here some thoughts:

I think we have a scanning dialog that pops up when the scan takes longer. There is a risk that mixxx is locked when using the library during scanning. So I would argue the user is disturbed anyway by the startup scan. 

On the other hand I can confirm that the scan during startup is annoying and a "useless" dialog would be more annoying. 

How about showing the dialog only if new tracks are found or some are missing in the startup case? 
For my use case it would be also nice to put new tracks in an "Incoming" crate. 

What do you think?

ronso0:
> I think we have a scanning dialog that pops up when the scan takes longer. There is a risk that mixxx is locked when using the library during scanning. So I would argue the user is disturbed anyway by the startup scan.

Exactly, the scan progress dialog pops up if the scan takes longer than 2s.
(fwiw during testing with auto-scan on startup I see the -empty- dialog artefact which never disappears, have to close it #12283)
Yes, the scan may be disturbing if one has new tracks on each start (many of them or a slow computer). And the scan during start with a fresh profile (new music dirs with lots of files) can lead to #11738 (which is why I have a delayed scan fix stashed).

ronso0:
> How about showing the dialog only if new tracks are found or some are missing in the startup case?

Yes, that wold make sense. new tracks, new missing tracks, track rediscovered.

ronso0:
> For my use case it would be also nice to put new tracks in an "Incoming" crate.

The "New tracks" feature / crate would make sense. How / when tracks are removed needs to be discussed. Automatically when a rating is set? And/or when they're sorted into playlists or crates?

I appreciate the idea, though it's unrelated to this UX fix/feature. I remember we briefly discussed this recently, is there already an Issue for tracking it?

ronso0:
I think the report can be optimized. IMO this is sufficient:
> Scan took 1181 ms.
> 
>1 new tracks found
0 moved tracks detected
1 tracks are missing
0 tracks have been rediscovered
>
> 16066 tracks in total

**ToDo**
* clarify new tracks (subtract moved tracks)
* count new missing tracks (change discovered during this scan)

daschuer:
> The "New tracks" feature ...

My current use case would work perfectly if the tracks stick in the "Incoming" crate, just like in my email in box.

Normally I sort them from there and than weeks later I delete them they are no longer considered as "New".

daschuer:
We have this one: 
https://github.com/mixxxdj/mixxx/issues/13403

I will file new one for the scanner part.

daschuer:
https://github.com/mixxxdj/mixxx/issues/13526

ronso0:
The report dialog now only displays relevant information:
new / moved / new missing / rediscovered tracks only if there are any.

I'll work on a fixup to show the report dialog during startup only if changes have been detected.

ronso0:
> I'll work on a fixup to show the report dialog during startup only if changes have been detected.

done.

src/library/scanner/libraryscanner.cpp
```diff
@@ -422,27 +428,63 @@ void LibraryScanner::slotFinishUnhashedScan() {
...
+    const int numNewTracks = m_scannerGlobal->addedTracks().size() - m_numRelocatedTracks;
+
+    const QSet<QString> existingTracks = m_trackDao.getAllExistingTrackLocations();
+    int numRediscoveredTracks = 0;
+    for (const QString& loc : std::as_const(m_previouslyMissingTracks)) {
+        if (existingTracks.contains(loc)) {
+            numRediscoveredTracks++;
+        }
+    }
+    const auto missingTracks = m_trackDao.getAllMissingTrackLocations();
```
daschuer, line 444: I have no idea how this performs and whether optimization is beneficial:
An idea for optimizion would be to move the compare loop into getAllMissingTrackLocations and get around the temporary QSet(). This will require a lot more compares though.

daschuer (commented):
Some comments. Not yet tested.

src/library/dao/trackdao.cpp
```diff
@@ -263,8 +263,51 @@ QList<TrackId> TrackDAO::resolveTrackIds(
...
+    if (!query.exec()) {
+        LOG_FAILED_QUERY(query);
+        DEBUG_ASSERT(!"Failed query");
+    }
+
+    int locationColumn = query.record().indexOf("location");
+    while (query.next()) {
+        locations.insert(query.value(locationColumn).toString());
+    }
+    return locations;
```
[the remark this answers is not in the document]
ronso0: thanks, done.

src/library/dao/trackdao.cpp
```diff
@@ -1905,6 +1940,7 @@ bool TrackDAO::detectMovedTracks(
             const auto nextSuffixMatch =
                     matchStringSuffix(nextTrackLocation, oldTrackLocation);
             DEBUG_ASSERT(nextSuffixMatch >= filename.length());
+            // document this
```
acolombier, line 1943: Any chance you can address this? Or maybe remove this comment otherwise

src/library/scanner/libraryscanner.cpp
```diff
@@ -575,16 +618,20 @@ void LibraryScanner::slotAddNewTrack(const QString& trackPath) {
             false);
     if (pTrack) {
         DEBUG_ASSERT(!pTrack->isDirty());
-        // The track's actual location might differ from the
-        // given trackPath
+        // The track's actual location might differ from the given trackPath.
+        // TODO why?
```
acolombier, line 622: Did you find out? If not do you want to remove this `TODO`?

src/library/scanner/libraryscanner.cpp
```diff
@@ -575,16 +618,20 @@ void LibraryScanner::slotAddNewTrack(const QString& trackPath) {
...
         const QString trackLocation(pTrack->getLocation());
         // Acknowledge successful track addition
         if (m_scannerGlobal) {
+            // TODO emit pTrack, or use a new signal to store the track id
```
acolombier, line 626: Is this still left `TODO`?

src/library/scanner/scannerglobal.h
```diff
@@ -144,6 +145,9 @@ class ScannerGlobal {
     }
     void trackAdded(const QString& trackLocation) {
         m_addedTracks << trackLocation;
+        // TODO Collect TrackIds?
```
acolombier, line 148: Is this still left `TODO`?

src/library/trackcollectionmanager.cpp
```diff
@@ -84,6 +85,11 @@ TrackCollectionManager::TrackCollectionManager(
                 this,
                 &TrackCollectionManager::libraryScanFinished,
                 /*signal-to-signal*/ Qt::DirectConnection);
+        connect(m_pScanner.get(),
+                &LibraryScanner::scanSummary,
+                this,
+                &TrackCollectionManager::libraryScanSummary,
+                /*signal-to-signal*/ Qt::DirectConnection);
```
acolombier, line 92: What's the meaning of `signal-to-signal`?

acolombier (commented):
Code LGTM, just nitpicking on comment cruft. Haven't had a chance to test yet. Could you address the `pre-commit` as well please?

src/library/trackcollectionmanager.cpp
ronso0: I think this pattern has been used to indicate we don't connect to a slot but to a signal (signal forward essentially)

ronso0:
Thanks for taking a look!

Pre-commit should be happy now.
I removed the TODOs you mentioned.

src/library/scanner/libraryscanner.cpp
ronso0: > whether optimization is beneficial

Well, depends on if there are many tracks missing ('many' like 1.000 or 10.000). In case the library is that huge, there will be a (blocking) progress dialog anyway and IMHO +- a few milliseconds don't really make a difference then.

I prefer to have a clean QSet to work with here.
We could also use QSet::intersect(QSet&) and QSet::subtract(QSet&) to get rediscovered and new missing tracks
This would look cleaner but I wonder if the extra QSets are worth it.

src/library/dao/trackdao.cpp
ronso0: I added the documentation.

src/library/scanner/libraryscanner.cpp
ronso0: A quick benchmark with PerfrmanceTimer:
`TrackDAO - Collecting 16026 track locations took 23 ms`

daschuer (approved):

fwcd:
Just updated to the latest snapshot and stumbled upon this dialog. At the risk of feeling a little https://xkcd.com/1172, is there a setting for disabling this? I occasionally add music and rescan during a live set (via the keyboard shortcut) and having a modal popup show up is a bit obtrusive IMO.

PS. If Mixxx had some form of notification system, I'd imagine this being a great use case for that. Or maybe there's either a Qt or some other cross-platform way of sending desktop notifications?

ronso0:
Nope, there' no setting.
Until we have some kind of notification service (we have a tracking issue for that but i just can't find it)
we could add a setting (a drop-down)
**Show Scan Summary**
* always
* not when playing
* never

PR #13429, opened 2024-07-01
Pioneer DDJ-SR2: Initial controller mapping

WaylonR:
Mapping base on - https://github.com/WaylonR/mixxx/tree/feat_Pioneer_DDJ_SR2

Forum thread - https://mixxx.discourse.group/t/pioneer-ddj-sr2-controller-mapping/30359

Manual PR - To come.

Manufacturer's product page - https://www.pioneerdj.com/en/product/controller/ddj-sr2/


res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+
+	      if (value) {
+	          DDJSR2.scratchMode[deck] = !DDJSR2.scratchMode[deck];
+	          midi.sendShortMsg(
+                    midiBaseAddress + offset,
+                    channel,
+                    DDJSR2.scratchMode[deck]
+		  );
+            }
+        }
```
Swiftb0y:
```suggestion
        trigger: function () {
            this.output();
        },
        outValueScale: function () {
            return deck.jogwheel.vinylMode ? this.on : this.off;
        }
        input: function(channel, control, value, status, group) {
            if (!this.isPress(channel, control, value, status, group)) {
                return;
            }
            deck.jogwheel.vinylMode = !deck.jogwheel.vinylMode; // this needs to change to whatever you named the JogWheelBasic instance I proposed above.
            this.trigger();
        },
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+    });
+	  // Defaulting the scratchmodes.
+    midi.sendShortMsg(
+        0x90 + offset,
+        0x17,
+        DDJSR2.scratchMode[offset]
+    );
```
Swiftb0y: don't do explicit `sendShortMsg`. Instead call trigger on the component that controls the LED (the mapping probably already does that automatically anyways if you have implemented `trigger` correctly).

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+    group: "[Master]",
+    inKey: "crossfader",
+    input: function() {
+        // We need a weird max. for the crossfader to make it cut cleanly.
+        // However, components.js resets max. to 0x3fff when the first value is
+        // received. Hence, we need to set max. here instead of within the
+        // constructor.
+        this.max = (0x7f<<7) + 0x70;
+        components.Pot.prototype.input.apply(this, arguments);
+    }
```
Swiftb0y, line 741: please elaborate. Why couldn't you just do:
```suggestion
    max: (0x7f<<7) + 0x70,
```
Also why the exactly the weird value, what values does the controller send when you put the crossfader to the right?

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+});
+
+// Encoder control cluster
+////////////////////////////////
+
+DDJSR2.BrowseEncoder  = new components.Encoder({
```
Swiftb0y, line 747: consider simplifying this by splitting it up into smaller components (eg one encoder for the actual encoder function and a separate knob component for the button functionality).

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+
+DDJSR2.backButton = function(channel, control, value, status) {
+    script.toggleControl("[Library]", "MoveFocusBackward");
+};
+
+DDJSR2.shiftBackButton = function(channel, control, value, status) {
+    if (value) {
+        script.toggleControl("[Master]", "maximize_library");
+    }
+};
```
Swiftb0y, line 869: convert to components please. You can construct them in `DDJ2SR.init`

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+		case DDJSR2.PadMode.SLICERLOOP:
+		    mode = null;
+			break;
+
+		case DDJSR2.PadMode.TRANS:
+		    mode = this.modes.trans;
+			break;
+		*/
+    }
+    engine.log(mode);
```
Swiftb0y: make sure to remove any debug logging leftover before commiting.
```suggestion
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+    }
+};
+
+DDJSR2.PadSection.prototype.clearHotcueButton = function(channel, control, value, status, group) {
+    const index = control - 0x08 + 1;
+    script.toggleControl(group, `hotcue_${  index  }_clear`);
+};
+
+DDJSR2.PadSection.prototype.stopSamplerButton = function(channel, control, value, status, group) {
+    const index = control - 0x38 + 1,
```
Swiftb0y, line 1206: these belong to the corresponding pad mode instances instead.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+            off: DDJSR2.PadColorMap.OFF,
+            colorMapper: DDJSR2.PadColorMap,
+            outConnect: false,
+        });
+    }
+    this.paramMinusButton = new components.Button({
+        midi: [0x90 + offset, 0x24],
+        group: deck.currentDeck,
+        outKey: "hotcue_focus_color_prev",
+        inKey: "hotcue_focus_color_prev",
```
Swiftb0y, line 1243: same below
```suggestion
        key: "hotcue_focus_color_prev",
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+    this.pads = new components.ComponentContainer();
+    for (let i = 0; i <= 7; i++) {
+        this.pads[i] = new components.HotcueButton({
+            midi: [0x97 + offset, 0x0 + i],
+            sendShifted: true,
+            shiftControl: true,
+            shiftOffset: 8,
+            number: i + 1,
+            group: deck.currentDeck,
+            on: this.color,
```
Swiftb0y, line 1233: this probably doesn't do anything if the colormapper takes preference.

Swiftb0y (commented):
plenty of comments for now. Feel free to ask questions.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+        }
+
+        DDJSR2.effectUnit[i].init();
+    }
+
+    // Get initial state of controls:
+    midi.sendShortMsg(0x9B, 0x08, 0x7F);
+
+    DDJSR2.seratoTimer = engine.beginTimer(250, "DDJSR2.doTimer", 0);
+    midi.sendSysexMsg(DDJSR2.seratoPoll, DDJSR2.seratoPoll.length);
```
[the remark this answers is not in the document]
WaylonR: I would assume from the comment, that it instructs the SR2 to dump the current control state. Will confirm once I've changed the rest of the mapping, according to the review.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+DDJSR2.valueVuMeter = {
+    "[Channel1]_current": 0,
+    "[Channel2]_current": 0,
+    "[Channel3]_current": 0,
+    "[Channel4]_current": 0,
+    "[Channel1]_enabled": 1,
+    "[Channel2]_enabled": 1,
+    "[Channel3]_enabled": 1,
+    "[Channel4]_enabled": 1
+};
```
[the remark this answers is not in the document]
WaylonR: also moving these two into a component or two:
DDJSR2.fx1assignLeds = [0x4C, 0x50, 0x70, 0x54];
DDJSR2.fx2assignLeds = [0x4D, 0x51, 0x71, 0x55];

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
Swiftb0y: Please do. I doubt the midi message does that, usually the sysex message does that.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+                            }
+                        }
+                    }
+                    midi.sendShortMsg(
+                        midiBaseAddress + DDJSR2.channelGroups[index],
+                        channel,
+                        midiOut
+                    );
+                }
+            }
```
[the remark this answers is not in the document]
WaylonR: Not sure I understand it myself. Think it's supposed to be blinking when AutoDJ tracks are playing? Think I'll remove it, as I don't think that would be all that useful.

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
Swiftb0y: Yeah move that directly into the component. The pattern is clear (fx2 is just fx1 + 1).

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+DDJSR2.autoDJShuffleAfterSkip = false;
+
+// If true, by releasing rotary selector,
+// track in preview player jumps forward to "jumpPreviewPosition"
+// (default: jumpPreviewEnabled = true, jumpPreviewPosition = 0.3).
+DDJSR2.jumpPreviewEnabled = true;
+DDJSR2.jumpPreviewPosition = 0.3;
+
+// If true, PFL / Cue (headphone) is being activated by loading a track into certain deck (default: true).
+DDJSR2.autoPFL = true;
```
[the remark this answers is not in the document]
JosepMaJAZ: On hercules dj console 4-mx, there is also this option and also set to true. 
The reason was that virtualdj had such behaviour when used with the controller (a lite version could be dowloaded and used with it).
It is handy, except if one has muscle memory to do that manually.

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
JosepMaJAZ: Indeed. Vumeter is better to be directly connected (connect and disconnect appropiately to not need the enabled setting) and send the value directly (except if you are using the value as "previous value" so that t you don't need to send the same value multiple times and so reduce the communication

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
JosepMaJAZ: I think that the seratopoll method might have to do with unlocking the controller, since this model is a serato one, not a generic midi one.
On the hercules dj console 4-mx, there is also a midi message to tell the controller to send the control values, and they are received as if they had just been tweaked.
I am assuming this controller supports that too.

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
Swiftb0y: Thats a good point, but I think its confusing when people go from for example serato (which is what this controller was made for iiuc) to mixxx and see different behavior... I'm not against handy extra behavior, I just don't want the defaults to confuse people that already own the hardware and want to try mixxx.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+    this.tempoFader = new components.Pot({
+	  unshift: function() {
+	      this.inKey = "rate";
+	      this.inSetParameter = function(value) {
+  	          let sliderRate;
+	          sliderRate = .5 - value;
+	          components.Pot.prototype.inSetValue.call(this, sliderRate*2);
+            };
+	  }
+    });
```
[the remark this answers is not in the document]
JosepMaJAZ: The rule is: the tempo fader on the controller has to move to the same direction than the component on the screen. 
The user preference about increase move up or increase move Down is a Mixxx preference in the decks section.

WaylonR:
A lot of progress done, a bit more to do, before a pre-commit check and push, but not much. Maybe a week.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+DDJSR2.getRotaryDelta = function(value) {
+    let delta = 0x40 - Math.abs(0x40 - value),
+        isCounterClockwise = value > 0x40;
+
+    if (isCounterClockwise) {
+        delta *= -1;
+    }
+
+    return delta;
+};
```
Swiftb0y, line 195: this doesn't look right. this function either returns `0x40` or `-0x40` and the bigger value is, the closer the resulting value goes to zero:

image

Whenever you use `getRotaryDelta` you also scale it way down, which also leads me to believe this is not actually doing what you want. In most cases you'll just want the same formula as already used in `getJogWheelDelta` below.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+
+DDJSR2.wheelLedControl = function(channelOffset, color) {
+    const wheelLedBaseChannel = 0xBB;
+    if (channelOffset !== null) {
+        midi.sendShortMsg(
+            wheelLedBaseChannel,
+            0x20+(channelOffset),
+            color 
+        );
+    }
```
Swiftb0y, line 210: you don't need to test for anything that is not expected to happen.
```suggestion
      midi.sendShortMsg(
          wheelLedBaseChannel,
          0x20 + channelOffset,
          color 
      );
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+};
+
+
+DDJSR2.pitchBendFromJog = function(group, movement) {
+    engine.setValue(group, "jog", movement / 5 * DDJSR2.jogwheelSensitivity);
+};
+
+DDJSR2.ringBufferView = class {
```
Swiftb0y, line 218: fyi since this is a class its name is usually written in `TitleCase` instead of `camelCase`

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+        return this.advanceBy(-1);
+    }
+	current() {
+        return this.indexable[this.index];
+    }
+};
+
+DDJSR2.BrowserContainer = function(){
+
+    this.browseKnob  = new components.ComponentContainer({
```
Swiftb0y, line 240: please fix the formatting here (pre-commit should probably take care of it). One level of indentation per opening brace.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+            return;
+        }
+
+        let sortColumn;
+        switch (control) {
+            case 0x58:  // BPM
+                sortColumn = 15;
+                break;
+            case 0x59:  // ARTIST
+                sortColumn = 1;
```
Swiftb0y, line 320: use `script.LIBRARY_COLUMNS` instead.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+    this.cue = new components.CueButton({
+        midi: [0x90 + channelOffset, 0x0C]
+    });
+    this.cuerewind = new components.Button({
+        midi: [0x90 + channelOffset, 0x48],
+          key: "start_stop"
+    });
+    this.jog = new components.JogWheelBasic({ // Jog Wheel is crap on the SR2, takes a big swing around to even get one midi tick. Use Platter instead.
+        deck: deckNumber,
+        wheelResolution: DDJSR2.scratchSettings.wheelResolution, // measurement (1000) wasn't producing accurate results (alt: 1073)
```
Swiftb0y, line 377: comment doesn't match settings

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+                midiOut = 0x00;
+
+              if (value) {
+                  theDeck.jog.vinylMode = !theDeck.jog.vinylMode;
+                  midi.sendShortMsg(
+                    midiBaseAddress + channelOffset,
+                    channel,
+                    theDeck.jog.vinylMode
+                  );
+            }
```
Swiftb0y, line 539:
```suggestion
            if (this.isPress(channel, control, value, status, group)) {
                theDeck.jog.vinylMode = !theDeck.jog.vinylMode;
                this.send(theDeck.jog.vinylMode);
            }
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+        }
+    });
+
+    theDeck.jog.vinylMode = 1; // Set to vinylMode, and the vinyl button is lit..
```
Swiftb0y, line 543: do this while constructing the JogWheelBasic instead. Also please use types properly (`true` instead of `1`), this isn't 1980's C code...

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+    });
+
+    this.keyShiftUp = new components.Button({
+        midi: [0x90 + channelOffset, 0x73],
+          key: "pitch_up"
+    });
+
+    this.keyReset = new components.Button({
+          // midi: [x,x], no light
+          key: "reset_key"
```
Swiftb0y, line 582: add the key back (for documentation purposes) and just specify `inKey` instead.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+    });
+
+    var blinkStatus = 0;
+    if (DDJSR2.wheelColor[deckNumber] === "POSITION") {
+      // Every time the playposition  changes, update the wheel color.
```
Swiftb0y, line 587: this would be much nicer if it was encapsulated into a custom component...

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+        this.channels[i] = new DDJSR2.Channel(i);
+    }
+
+    this.crossfader = new components.Pot({
+        group: "[Master]",
+        inKey: "crossfader",
+    });
+    DDJSR2.panels = [];
+    DDJSR2.panels[0] = false;
+    DDJSR2.panels[1] = false;
```
Swiftb0y, line 645:
```suggestion
    DDJSR2.panels = [false, false];
```

Swiftb0y (changes requested):
a couple more comment, notice that there are also another batch of open comments on the previous review
image
I also prefer to resolve review comments myself because its often the case that a review comment is not understood or a request not properly executed so I want to make sure myself that they are actually resolved.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+    let delta = 0x40 - Math.abs(0x40 - value);
+    const isCounterClockwise = value > 0x40;
+    // BrowserKnob returns 1-30 going down, and 127-98 going up.
+
+    if (isCounterClockwise) {
+        delta *= -1;
+    }
+    return delta;
+};
+
```
Swiftb0y, line 169: right, that can be done way simpler...
```suggestion
DDJSR2.getRotaryDelta = function(value) {
    return value > 0x40 : value - 0x7F : value;
};

```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+        trackColorCycleEnabled: false,
+        trackColorCycleHappened: false,
+        previewSeekEnabled: false,
+        previewSeekHappened: false,
+        turn: new components.Encoder({
+            group: "[Library]",
+            input: function(channel, control, value, _status, _group) {
+                const rotateValue = DDJSR2.getRotaryDelta(value);
+                engine.setValue("[Library]", "MoveVertical", rotateValue);
+            }
```
Swiftb0y, line 222: pretty sure this works just as well. 
```suggestion
            inKey: "MoveVertical",
            inValueScale : DDJSR2.getRotaryDelta
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+
+    this.loadPrepareButtonShifted = new components.Button({
+        group: "[Library]",
+        inKey: "AutoDjAddTop"
+    });
+
+
+};
+DDJSR2.BrowserContainer.prototype = new components.ComponentContainer();
+DDJSR2.ticks = 0;
```
Swiftb0y, line 322:
```suggestion
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+        deck: deckNumber,
+        wheelResolution: DDJSR2.scratchSettings.wheelResolution,
+        alpha: DDJSR2.scratchSettings.alpha,
+        beta: DDJSR2.scratchSettings.beta,
+        vinylMode: false, // Set the default to match the SR2's default.
+        inValueScale: function(value) {
+            if (!this.vinylMode) {
+                const delta = DDJSR2.getJogWheelDelta(value);
+                const newvalue = delta / 5 * DDJSR2.jogwheelSensitivity;
+                return newvalue;
```
Swiftb0y, line 354:
```suggestion
                return (DDJSR2.getJogWheelDelta(value) / 5) * DDJSR2.jogwheelSensitivity;
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+            } else {
+                return value < 0x40 ? value - (this.max + 1) : value;
+            }
+        }
+    });
+    this.tempoFader = new components.Pot({
+        invert: true,
+        key: "rate"
```
Swiftb0y, line 362:
```suggestion
        inKey: "rate"
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+        midi: [0x90 + channelOffset, 0x1A],
+        key: "keylock",
+        type: components.Button.prototype.types.toggle
+    });
+    this.needleSearchStripPositionShifted = new components.Pot({
+        group: theDeck.group,
+        key: "playposition"
+    });
+    this.needleSearchStripPosition = new components.Pot({
+        key: "playposition",
```
Swiftb0y, line 391:
```suggestion
        inKey: "playposition",
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+        }
+    });
+    this.keyLock = new components.Button({
+        midi: [0x90 + channelOffset, 0x1A],
+        key: "keylock",
+        type: components.Button.prototype.types.toggle
+    });
+    this.needleSearchStripPositionShifted = new components.Pot({
+        group: theDeck.group,
+        key: "playposition"
```
Swiftb0y, line 388:
```suggestion
        inKey: "playposition"
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+        key: "sync_enabled",
+        type: components.Button.prototype.types.toggle
+    });
+    this.syncOff = new components.Button({
+        midi: [0x90 + channelOffset, 0x5C],
+        key: "sync_enabled",
+        input: function(_channel, _control, _value, _status, _group) {
+            this.inSetValue(false);
+        }
+    });
```
Swiftb0y, line 411: you'll probably don't want both controlling the same led.
```suggestion
    this.sync = new components.Button({
        midi: [0x90 + channelOffset, 0x58],
        key: "sync_enabled",
        type: components.Button.prototype.types.toggle
    });
    this.syncOff = new components.Button({
        midi: [0x90 + channelOffset, 0x5C],
        inKey: "sync_enabled",
        input: function(_channel, _control, _value, _status, _group) {
            this.inSetValue(false);
        }
    });
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+    this.autoLoop = new components.Button({
+        midi: [0x90 + channelOffset, 0x14],
+        outKey: "loop_enabled",
+        input: function(channel, control, value, status, group) {
+            if (value) {
```
Swiftb0y, line 416:
```suggestion
            if (this.isPress(channel, control, value, status, group) {
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+                if (engine.getValue(group, "loop_enabled")) {
+                    engine.setValue(group, "reloop_toggle", true);
+                    engine.setValue(group, "reloop_toggle", false);
+                } else {
+                    engine.setValue(group, "beatloop_activate", true);
+                    engine.setValue(group, "beatloop_activate", false);
```
Swiftb0y, line 422:
```suggestion
                    script.triggerControl(group, "reloop_toggle");
                } else {
                    script.triggerControl(group, "beatloop_activate");
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+    this.softStartButton = function(channel, control, value, _status, _group) {
+        const activate = value > 0;
+        if (activate) { // act on button press
+            engine.softStart(deckNumber, true);
+        }
+    };
+
+    this.gridSetButton = new components.Button({
+        midi: [0x90 + channelOffset, 0x64],
+        key: "beats_translate_curpos"
```
Swiftb0y, line 519:
```suggestion
        inKey: "beats_translate_curpos"
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+            } else if ((DDJSR2.panels[0] === true) && (DDJSR2.panels[1] === true)) {
+                DDJSR2.panels[1] = false;
+            } else if ((DDJSR2.panels[1] === false) && (DDJSR2.panels[0] === true)) {
+                DDJSR2.panels[0] = false;
+            }
+
+            engine.setValue("[Samplers]", "show_samplers", DDJSR2.panels[0]);
+            engine.setValue("[EffectRack1]", "show", DDJSR2.panels[1]);
+        }
+    };
```
Swiftb0y, line 597: unless JS weird semantics about function calling screw me over again, this should work.
```suggestion
    const panelStates = new DDJSR2.RingBufferView([[false, false], [true, false], [true, true], [false, true]]);
    const makePanelSelectButton = method => function(channel, control, value, _status, _group) { 
            if (value) {
                const [showSamplers, showFX] = method();
    
                engine.setValue("[Samplers]", "show_samplers", showSamplers);
                engine.setValue("[EffectRack1]", "show", showFX);
            }
        };
    this.panelSelectButton = makePanelSelectButton(panelStates.next);
    this.shiftPanelSelectButton = makePanelSelectButton(panelStates.previous);
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+        group: deck,
+        inKey: "pregain",
+    });
+
+    this.eqKnob = [];
+    // EQ High, Mid and Low
+    for (let k = 1; k <= 3; k++) {
+        this.eqKnob[k] = new components.Pot({
+            group: `[EqualizerRack1_${  deck  }_Effect1]`,
+            inKey: `parameter${  k}`,
```
Swiftb0y, line 615:
```suggestion
            group: `[EqualizerRack1_${deck}_Effect1]`,
            inKey: `parameter${k}`,
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1411 @@
...
+
+    let lastVuLevel = 0;
+    this.vuMeter = new components.Component({
+        outKey: "vu_meter",
+        group: deck,
+        midi: [0xB0, 0x02],
+        output: function(value, group, _control) {
+            // Remark: Only deck vu meters can be controlled! Master vu meter is handled by hardware!
+
+            value = parseInt(value * 0x76); //full level indicator: 0x7F
```
Swiftb0y, line 652: the parseInt makes no sense here.
```suggestion
            value = value * 0x76; //full level indicator: 0x7F
```

Swiftb0y (commented):
couple more comments, this is taking shape

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -561,40 +552,21 @@ DDJSR2.MixerContainer = function() {
...
-            engine.setValue("[Samplers]", "show_samplers", DDJSR2.panels[0]);
-            engine.setValue("[EffectRack1]", "show", DDJSR2.panels[1]);
+            engine.setValue("[Samplers]", "show_samplers", showSamplers);
+            engine.setValue("[EffectRack1]", "show", showFX);
         }
     };
+
+    // Pass the bound method reference
+    this.panelSelectButton = makePanelSelectButton(panelStates.next.bind(panelStates));
+    this.shiftPanelSelectButton = makePanelSelectButton(panelStates.previous.bind(panelStates));
```
Swiftb0y, line 569: using bind like this is a little ugly.
```suggestion
    this.panelSelectButton = makePanelSelectButton(() => panelStates.next());
    this.shiftPanelSelectButton = makePanelSelectButton(() => panelStates.previous());
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+                        } else {
+                            script.triggerControl("[PreviewDeck1]", "stop");
+                        }
+                    } else {
+                        this.trackColorCycleHappened = false;
+                    }
+                } else {
+                    this.trackColorCycleEnabled = true;
+                }
+            }.bind(this)
```
Swiftb0y, line 236: deep nesting is usually a code smell (makes the code hard to follow). 
```suggestion
                
                if (this.isPress(channel, control, value, status, group)) {
                    theContainer.browseKnob.trackColorCycleEnabled = true;
                    return;
                }
                // On release, not press. Allows for track color cycling.
                theContainer.browseKnob.trackColorCycleEnabled = false;
                if (theContainer.browseKnob.trackColorCycleHappened) {
                    theContainer.browseKnob.trackColorCycleHappened = false;
                    return;
                }
                if (!engine.getValue("[PreviewDeck1]", "play")) {
                    engine.setValue("[PreviewDeck1]", "LoadSelectedTrackAndPlay", 1);
                } else {
                    script.triggerControl("[PreviewDeck1]", "stop");
                }
            }
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+        }),
+        shiftTurn: new components.Encoder({
+            group: "[PreviewDeck1]",
+            input: function(channel, control, value, _status, _group) {
+                const rotateValue = DDJSR2.getRotaryDelta(value);
+                if (!this.trackColorCycleEnabled) {
+                    const oldPos = engine.getValue("[PreviewDeck1]", "playposition");
+                    const newPos = Math.max(0, oldPos + (0.05 * rotateValue));
+                    engine.setValue("[PreviewDeck1]", "playposition", newPos);
```
Swiftb0y, line 245: alternatively, depending on the encoder direction, just beatjumping 4-8 beats in either direction would also work nicely, wdyt?

Swiftb0y (commented):
Please do me the favor and convert the remaining couple "plain" handlers to components as well.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+            break;
+        case 0x59:  // ARTIST
+            sortColumn = script.LIBRARY_COLUMNS.ARTIST;
+            break;
+        default:
+            // unknown sort column
+            return;
+        }
+        engine.setValue("[Library]", "sort_column", sortColumn);
+    };
```
Swiftb0y, line 287: Obviously the duplicated initial definitions aren't great either and could be trivially solved by simply subclassing, but thats rather tedious unfortunately, so I'm okay with the repetition here.
```suggestion
    this.sortLibraryBpm = new components.Button({
        inKey: "sort_column",
        group: "[Library]",
        type: components.Button.prototype.types.toggle;
        inToggle: function() {
            this.inSetValue(script.LIBRARY_COLUMNS.BPM);
        }
    });
    this.sortLibraryArtist= new components.Button({
        inKey: "sort_column",
        group: "[Library]",
        type: components.Button.prototype.types.toggle;
        inToggle: function() {
            this.inSetValue(script.LIBRARY_COLUMNS.ARTIST);
        }
    });
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+        this.loadButtons[channelOffset] = new components.Button({
+            group: group,
+            key: "LoadSelectedTrack",
+            midi: [0x9B, 0x0+channelOffset]
+        });
+        this.ejectTrackButton = new components.Button({
+            inKey: "eject",
+            outKey: "track_loaded",
+            midi: [0x96, 0x46+channelOffset]
+        });
```
Swiftb0y, line 266: this is unused nor does it make sense to create repeatedly in the loop.
```suggestion
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+        }),
+    });
+    this.loadButtons = [];
+    for (let channelOffset = 0; channelOffset < 4; channelOffset++) {
+        const group = `[Channel${channelOffset+1}]`;
+        this.loadButtons[channelOffset] = new components.Button({
+            group: group,
+            key: "LoadSelectedTrack",
+            midi: [0x9B, 0x0+channelOffset]
+        });
```
Swiftb0y, line 261: alternatively just create them in each Channel instead instead of on the mixer...
```suggestion
    this.loadButtons = [0, 1, 3, 4].map(channelOffset => new components.Button({
        group: `[Channel${channelOffset+1}]`,
        key: "LoadSelectedTrack",
        midi: [0x9B, channelOffset]
    }));
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+                const rotateValue = DDJSR2.getRotaryDelta(value);
+                if (!Browser.browseKnob.trackColorCycleEnabled) {
+                    engine.setValue("[PreviewDeck1]", "beatjump", rotateValue*8);
+                } else {
+                    const key = (rotateValue > 0) ? "track_color_next" : "track_color_prev";
+                    engine.setValue("[Library]", key, 1.0);
+                    Browser.browseKnob.trackColorCycleHappened = true;
+                }
+            }
+        }),
```
Swiftb0y, line 252: this should work just as well.
```suggestion
            input: function(channel, control, value, status, _group) {
                this.isPressed = this.isPress(channel, control, value, status);
                if (this.isPressed) {
                    return;
                }
                if (!engine.getValue("[PreviewDeck1]", "play")) {
                    engine.setValue("[PreviewDeck1]", "LoadSelectedTrackAndPlay", 1);
                } else {
                    script.triggerControl("[PreviewDeck1]", "stop");
                }
            }
        }),
        shiftTurn: new components.Encoder({
            beatJumpSize: 8,
            inValueScale: DDJSR2.getRotaryDelta,
            inSetValue: function(rotateValue) {
                if (Browser.browseKnob.isPressed) {
                    const direction = rotateValue > 0 ? "next" : "prev";
                    script.toggleControl("[Library]", `track_color_${direction}`);
                } else {
                    engine.setValue("[PreviewDeck1]", "beatjump", rotateValue*beatJumpSize;
                }
            }
        }),
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+
+DDJSR2.BrowserContainer = function() {
+    const Browser = this;
+    this.browseKnob  = new components.ComponentContainer({
+        longPressTimer: 0,
+        longPressTimeout: 250,
+        trackColorCycleEnabled: false,
+        trackColorCycleHappened: false,
+        previewSeekEnabled: false,
+        previewSeekHappened: false,
```
Swiftb0y, line 210:
```suggestion
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+		case DDJSR2.PadMode.SLICERLOOP:
+		    mode = null;
+			break;
+
+		case DDJSR2.PadMode.TRANS:
+		    mode = this.modes.trans;
+			break;
+		*/
+    }
+    return mode;
```
Swiftb0y, line 867:
```suggestion
    switch (control) {
    case DDJSR2.PadMode.HOTCUE:
        return this.modes.hotcue;
    case DDJSR2.PadMode.ROLL:
        return this.modes.roll;
    case DDJSR2.PadMode.SAMPLER:
        return this.modes.sampler;
    case DDJSR2.PadMode.SAMPLER2:
        return this.modes.sampler;
    case DDJSR2.PadMode.PITCHPLAY:
        return this.modes.pitchplay;
/*		
    case DDJSR2.PadMode.SLICR:
		return this.modes.slicer;
	case DDJSR2.PadMode.CUELOOP:
    case DDJSR2.PadMode.SAVEDLOOP:
    case DDJSR2.PadMode.SLICERLOOP:
		return null;
    case DDJSR2.PadMode.TRANS:
	    return this.modes.trans;
*/
    }
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+            if (value) {
+                const jumpSize = engine.getValue(this.group, "beatjump_size");
+                if (jumpSize > this.mode.minSize) {
+                    engine.setValue(this.group, "beatjump_size", jumpSize / 2);
+                }
+            }
+        },
+        output: function(value, _group, _control) {
+            this.send((value > this.mode.minSize) ? this.on : this.off);
+        },
```
Swiftb0y, line 1040:
```suggestion
        outValueScale: function(value) {
            return value > this.mode.minSize ? this.on : this.off);
        },
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+    this.color = DDJSR2.PadColor.CELESTE;
+    this.pads = new components.ComponentContainer();
+    this.loopSize = 0.03125;
+    this.minSize = 0.03125;  // 1/32
+    this.maxSize = 32;
+
+    let loopSize;
+    let i;
+    for (i = 0; i <= 3; i++) {
+        loopSize = (this.loopSize * Math.pow(2, i));
```
Swiftb0y, line 994:
```suggestion
    for (let i = 0; i <= 3; i++) {
        const loopSize = (this.loopSize * Math.pow(2, i));
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
Swiftb0y: > where is theContainer defined? Do you mean DDJSR2.BrowserContainer? Should there be a theContainer = this; at the start of the BrowserContainer function, to define theContainer?

Yeah, you need to define it. I just couldn't add a suggestion for that because of github limitations.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+        }
+    });
+    this.ShiftedPerformancePad = function(n) {
+        this.midi = [0x97 + offset, 0x78 + n];
+        this.number = n + 1;
+        this.on = this.color + DDJSR2.PadColor.DIM_MODIFIER;
+        this.colorMapper = DDJSR2.PadColorMap;
+        this.colorKey = `hotcue_${  this.number  }_color`;
+        components.Button.call(this);
+    };
```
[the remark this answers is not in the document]
Swiftb0y: My question was unrelated to pitch play (I just filed the PR at a suboptimal place). If you do the procedure I outlined just on the plain `HotcueMode` pads, does the pad stay pressed? (you can test that by doing it when the track is paused, if the tracks continues to play after you released the button, it didn't register the release).

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1323 @@
...
+        inKey: "playposition"
+    });
+    this.needleSearchStripPosition = new components.Pot({
+        inKey: "playposition",
+        group: theDeck.group,
+        input: function(channel, control, value, _status, _group) {
+            if (!engine.getParameter(this.group, "play")) {
+                const newValue = this.inValueScale(value);
+                this.inSetParameter(newValue);
+            }
```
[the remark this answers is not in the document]
Swiftb0y: Ah didn't see that...

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1323 @@
...
+    });
+
+    this.eqKnob = [];
+    // EQ High, Mid and Low
+    for (let k = 1; k <= 3; k++) {
+        this.eqKnob[k] = new components.Pot({
+            group: `[EqualizerRack1_${deck}_Effect1]`,
+            inKey: `parameter${k}`,
+        });
+    }
```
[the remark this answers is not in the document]
Swiftb0y: yeah, putting in a comment is also completely fine. Can't really make much of the error. It seems the eqKnobs never got filled correctly due to some other error thrown while setting up the mapping. You'll need to look further back in the log.

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
WaylonR: If you hold down hotcue, to play a hotcue on a track that's not already playing (so playing is reliant on the hotcue being held) press shift, it will still play, but if you let go of the hotcue button, it will stop.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1323 @@
...
+        key: "play_stutter"
+    });
+    this.cue = new components.CueButton({
+        midi: [0x90 + channelOffset, 0x0C]
+    });
+    this.cuerewind = new components.Button({
+        midi: [0x90 + channelOffset, 0x48],
+        key: "start_stop"
+    });
+    this.jog = new components.JogWheelBasic({ // Jog Wheel is crap on the SR2, takes a big swing around to even get one midi tick. Use Platter instead.
```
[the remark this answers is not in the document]
Swiftb0y: right, thats probably due to mixxx's weird scaling, but when determining the ticks for the wheel, you should get the same number regardless of whether you're touching the top or the side of the platter.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -1280,104 +1242,83 @@ DDJSR2.PitchPlayMode = function(deck, offset) {
...
-            } else {
-                this.colorCode = colorCode;
-                DDJSR2.wheelLedControl(this.channelOffset, colorCode);
-            }
+            const revolutionsPerSecond = DDJSR2.scratchSettings.vinylSpeed / 60;
+            const maxColorValue = DDJSR2.wheelLedCircle.maxVal;
+            // Determine the current position in the color cycle
+            // The modulo operation ensures the color value stays within the available range
+            // The multiplication by maxColorValue ensures that the position is spread over the entire range of indexed colors.
+            colorIndex = Math.round((revolutionsPerSecond * elapsedTime * maxColorValue) % maxColorValue);
```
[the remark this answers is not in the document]
Swiftb0y: sure... so then lets simplify the code a bit. I'm just making sure that high-frequency color changing behavior is actually what you want.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1305 @@
...
+        if (remainingTime > 0 && remainingTime < 30 && !engine.isScratching(this.deckNumber)) {
+            //increase blinking according time left
+            const blinkInterval = Math.max(remainingTime / 3, 3);
+            if (this.blinkStatus < blinkInterval) {
+                colorIndex = DDJSR2.wheelLedCircle.blink;
+            } else if (this.blinkStatus > (blinkInterval - parseInt(6 / blinkInterval))) {
+                this.blinkStatus = 0;
+                colorIndex = DDJSR2.wheelLedCircle.maxVal;
+            }
+            this.blinkStatus++;
```
Swiftb0y, line 1293: the midi mapping reference mentions an "End Warning" (`midi: [0x9B, 0x04 + channelOffset]`). Why don't you just use that?

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
Swiftb0y: Okay thats good to know. Thanks for the confirmation.

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
WaylonR: nope. the SR2's hardware doesn't send any midi event unless you really put some speed into the wheel movement, if you  are just using the jog, not the platter. It sends different controls/channels too.

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
WaylonR: set the decks in the PR to track mode by default?

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
WaylonR: except, the other method of filling eqknob works fine. Guess I'll have to make a eqknob object, and do new eqKnob(parameterIndex)?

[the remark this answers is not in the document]
WaylonR: Fixed, Had to change the xml, the knobs with the for loop, were 1 based. map is 0 based.

res/controllers/Pioneer-DDJ-SR2-scripts.js
WaylonR: tried it. seems like it does nothing. The jog ring goes blank.

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
Swiftb0y: Ah sorry, I didn't realize that. Its probably better if you don't have empty slots in the array.

res/controllers/Pioneer-DDJ-SR2-scripts.js
Swiftb0y: thats weird... you could try reverse-engineering it by looking at the traffic from/to serato. Have you tried sending all possible values or just specific ones?

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
Swiftb0y: thats dumb... but I guess thats to get the pioneer CDJ feel...

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
Swiftb0y: yes, I think it should be that by default as well.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+        midi: [0x90 + channelOffset, 0x58],
+        key: "sync_enabled",
+        type: components.Button.prototype.types.toggle
+    });
+    this.syncOff = new components.Button({
+        midi: [0x90 + channelOffset, 0x5C],
+        inKey: "sync_enabled",
+        input: function(_channel, _control, _value, _status, _group) {
+            this.inSetValue(false);
+        }
```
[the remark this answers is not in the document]
Swiftb0y: you can either make use of the `sendShifted` facilities in componentsJS or just add `sync_enabled` as an outkey on the shifted button component. I think the latter is simpler.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1383 @@
...
+    outKey: "playposition",
+    output: function(value) {
+        if (this.color === "POSITION") {
+            // Every time the playposition  changes, update the wheel color.
+        // Timing calculation is handled in seconds!i
+            const duration = engine.getValue(this.group, "duration");
+            const elapsedTime = value * duration;
+            const remainingTime = duration - elapsedTime;
+            const revolutionsPerSecond = DDJSR2.scratchSettings.vinylSpeed / 60;
+            const speed = parseInt(revolutionsPerSecond * DDJSR2.wheelLedCircle.maxVal);
```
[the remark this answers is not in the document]
Swiftb0y: sure. do you have a specific curve in mind? maybe just go linearly from 1Hz to 5Hz over the 30 second duration?

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1459 @@
...
+
+// DDJSR2.shiftPressed = false;
+// DDJSR2.shiftPanelSelectPressed = false;
+
+DDJSR2.rateRanges = [0.08, 0.16, 0.32, 0.64, 0.90, 1.00];
+
+DDJSR2.scratchSettings = {
+    "alpha": 1.0 / 8,
+    "beta": 1.0 / 8 / 32,
+    "wheelResolution": 20000,
```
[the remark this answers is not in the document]
Swiftb0y: > Needs calibration. 50000 seems to be correct, where a 360 rotation of the wheel is the same as a 360 rotation in the Spinning Vinyl onscreen. Is this not the intended purpose?

Yes that is usually the intended purpose. A one-to-one mapping between the onscreen vinyl sticker and the physical one (in theory if there was one). I'm just confused because usually the number if ticks per rotation is much lower (~1000 on highend controllers) so either the calculations in the mapping are wrong or pioneer is cheating again because I don't think their jogwheel is actually that high-res.

res/controllers/Pioneer-DDJ-SR2-scripts.js
Swiftb0y: > Removed "on: this.color" no change.. but a bug that occurs occasionally, white hotcues, where there are no hotcues present in the track, still happens.


Does this still happen?

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+   		input: function(channel, control, value, status, group) {
+			    if (value) {
+                if (engine.getValue(group, "loop_enabled")) {
+                    engine.setValue(group, "reloop_toggle", true);
+                    engine.setValue(group, "reloop_toggle", false);
+                } else {
+                    engine.setValue(group, "beatloop_activate", true);
+                    engine.setValue(group, "beatloop_activate", false);
+                }
+            }
```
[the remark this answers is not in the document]
Swiftb0y: thats weird. does it work now?

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1600 @@
...
+	  type: components.Button.prototype.types.toggle
+    });
+
+    this.gridSetButton = new components.Button({
+        midi: [0x90 + offset, 0x64],
+        key: "beats_translate_curpos"
+    });
+
+    // Every time the playposition  changes, update the wheel color.
+    engine.makeConnection(this.currentDeck, "playposition", function(value, group) {
```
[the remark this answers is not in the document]
Swiftb0y: this should be fixed with my latest refactoring, right?

res/controllers/Pioneer-DDJ-SR2-scripts.js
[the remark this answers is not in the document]
WaylonR: sounds good.

res/controllers/Pioneer-DDJ-SR2-scripts.js
WaylonR: the 1hz to 5hz end warning will do fine. On all jog modes, unless it's 'off'?

WaylonR: That's a bit more complicated than I know to do. Can we just use the 1hz to 5hz idea for End Warning?

WaylonR: We forgot about the CO end_of_track ... we don't need the math to create a warning.

WaylonR: But still need math to make the blinking. hmm...

WaylonR: Unless I just have it do a 1hz blinking throughout, abandon the linear ramping. Will go with that, see what it's like.

WaylonR: It's actually not bad at all.

Swiftb0y: great. the simpler the mapping, the better.

Swiftb0y: yeah, try [`[App],indicator_500ms`](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls#control-[App]-indicator_500ms)

WaylonR: nope, needs the context of the wheelring of the deck it's set from. I'll work it out.

WaylonR: Yes, gathered that, but it needs the deck info, and to know what deck jog wheel to blink. It may mean setting four timers by default.. one per deck set to track. Unless we hook into playposition somehow, and track_color at the same time... how do we do that?

WaylonR: Implement the playposition as a new components.pot inside the wheelring prototype?

WaylonR: this idea is good for when color is set to "POSITION":
    endWarning: function(colorIndex) {

        // Check if we're nearing the end of the track and not scratching
        if (engine.getValue(this.group, 'end_of_track') && !engine.isScratching(this.deckNumber)) {
            const framesPerHalfCycle = 30; // 1 second cycle, split into two half-cycles (on/off)

            // Toggle between blink and maxVal states based on the half-cycle frame count
            if (this.blinkStatus < framesPerHalfCycle) {
                colorIndex = DDJSR2.wheelLedCircle.blink; // Blink state for half the cycle
            } else if (this.blinkStatus < framesPerHalfCycle * 2) {
                colorIndex = DDJSR2.wheelLedCircle.maxVal; // MaxVal state for the other half
            } else {
                this.blinkStatus = 0; // Reset blink status at the end of the full cycle
            }
            if(this.color === "TRACK"){
                this.send(colorIndex);
            }
            this.blinkStatus++; // Increment frame counter
        } else {
            // Reset blink state when not at the end of the track
            this.blinkStatus = 0;
        }

        return colorIndex;
}
endWarning is called by sendIfChanged.. however, track_color never gets updated, unless track_color changes, therefore, endWarning is not run when end_of_track happens in TRACK mode.

Swiftb0y: yes, thats the most difficult part here, aggregating all the different events that play here. I will try to come up with a PoC.

WaylonR: What about turning WheelRing into a container, and making components of playposition outKey and track_color?

Swiftb0y: yeah not sure yet. I don't want to create distinct components for each mode because there is quite a bunch of shared logic due to the end of track warning in all modes. I didn't manage to whip up something today but I'll try so tomorrow.

WaylonR: not distinct modes for each component, but have the componentcontainer be  the driver of the thing, and have track_color and playposition be individual components... only playposition runs endWarning... and it skips coloring the jog wheel if it's in track mode. track_color has logic in it, to only trigger if it's in track mode.

Swiftb0y: having multiple separate components controlling the same LED doesn't make sense, they will just step over each other when they can't communicate. And if they can communicate (aka depend on each other), there is little reason to make them separate components in the first place. Before I continue digging into the code here, I need to know whether the endWarning LED message actually works. You previously said that it just turns of the jogwheel LED, which I took as the feature being essentially broken, is that correct?

Swiftb0y: sure, though judging by the buggy behavior you described, its probably more flawed than you think...

WaylonR: so, code works for POSITION mode only. Just changes the color that gets send() if it's end_of_track.

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1317 @@
...
+        const revolutionsPerSecond = DDJSR2.scratchSettings.vinylSpeed / 60;
+        const maxColorValue = DDJSR2.wheelLedCircle.maxVal;
+        // Determine the current position in the color cycle
+        // The modulo operation ensures the color value stays within the available range
+        // The multiplication by maxColorValue ensures that the position is spread over the entire range of indexed colors.
+        colorIndex = Math.round((revolutionsPerSecond * elapsedTime * maxColorValue) % maxColorValue);
+
+        return colorIndex;
+    }
+});
```
Swiftb0y, line 1317:
```suggestion
const WheelRing = function(options) {
    components.Component.call(this, options);
}

WheelRing.prototype = new components.Component({
    send: function(value) {
        if (this.value !== value) {
            this.value = value;
            components.Component.send.call(this, value);
        }
    },
    outputMaybe: function() {
        if (engine.getValue(this.group, "end_of_track") && engine.getValue("[App]", "indicator_500ms")) {
            this.send(this.off);
            return;
        }
        components.Component.output.call(this, this.outGetValue(), this.group, this.outKey);
    },
    trigger: function() {
        this.outputMaybe();
    },
    connect: function() {
        if (typeof this.group !== "string") {
            throw `invalid group: ${this.group}`;
        }
        const COs = [, {group: this.group, key: "end_of_track"}, {group: "[App]", key: "indicator_500ms"}];
        if (typeof this.outKey === "string") {
            COs[0] = {group: this.group, key: this.outKey};
        }
        this.connections = COs.map(co => engine.makeConnection(co.group, co.key, this.outputMaybe.bind(this)));
    }
});
```

res/controllers/Pioneer-DDJ-SR2-scripts.js
```diff
@@ -0,0 +1,1317 @@
...
+
+    this.keyReset = new components.Button({
+        midi: [0x90 + channelOffset, 0x71],
+        inKey: "reset_key"
+    });
+    this.wheelRing = new WheelRing({
+        midi: [0xBB, 0x20 + channelOffset],
+        color: DDJSR2.wheelColor[deckNumber],
+        deckNumber: deckNumber,
+    });
```
Swiftb0y, line 500:
```suggestion
    switch (DDJSR2.wheelColor[deckNumber]) {
    case "TRACK":
        this.wheelRing = new WheelRing({
            midi: [0xBB, 0x20 + channelOffset],
            group: this.group,
            outKey: "track_color",
            output: function(colorRGB) { this.sendIfChanged(this.colorMapper.getValueForNearestColor(colorRGB)); };
        });
        break;
    case "POSITION":
        this.wheelRing = new WheelRing({
            midi: [0xBB, 0x20 + channelOffset],
            group: this.group,
            outKey: "position",
            revolutionsPerSecond: DDJSR2.scratchSettings.vinylSpeed / 60;
            output: function(position) {
                // Every time the playposition  changes, update the wheel color.
                // Timing calculation is handled in seconds!!
                const elapsedTime = position * engine.getValue(this.group, "duration");;
                // Determine the current position in the color cycle
                // The modulo operation ensures the color value stays within the available range
                // The multiplication by maxColorValue ensures that the position is spread over the entire range of indexed colors.
                this.send(Math.round((this.revolutionsPerSecond * elapsedTime * maxColorValue) % maxColorValue));
            }
        });
        break;
    default:
        this.wheelRing = new WheelRing({
            midi: [0xBB, 0x20 + channelOffset],
            group: this.group,
            outValueScale: () => DDJSR2.wheelColor[deckNumber]
        });
    };
```

Swiftb0y (commented):
try this. this overlays the end_of_track warning in a way that doesn't rely on the frequent updates of `position`. That way it should work correctly in the static and the `track_color` mode.

res/controllers/Pioneer-DDJ-SR2-scripts.js
Swiftb0y: yeah, thats what I was afraid of. Try the approach I just posted in the separate review.

res/controllers/Pioneer-DDJ-SR2-scripts.js
WaylonR: Pushed a new update to the wheelRing code, to leverage the 500ms timer, and ensure that the endWarning occurs even in TRACK mode. Also fixes the jog wheel color on load of a new track.

Swiftb0y: can you elaborate on why you chose what you pushed over what I proposed? I think what you pushed is much less clean...

WaylonR: because it didn't work. kept getting messages about .call'ing on undefined, for components.Component.output.call and send.call, couldn't work out why it wasn't working. Knew my code worked, just needed the TRACK mode working, and it synced with 500ms, you provided the glue for that.

Swiftb0y: > components.Component.output.call and send.call,

yeah, should have been `components.Component.prototype.output.call`. my bad.

WaylonR: If it did work, still needed the endWarning blink code in it, going from red to white alternating with each 500ms, which is what this.blink = !this.blink does. .fixcolor is there, as I found a bug, Once endWarning happens, if one changes track, and it happens to be the same colored track, as the engine thinks there is no track color change, it doesn't run the output, so the new track doesn't reset the jog wheel color. using track_loaded is forcing the refresh.

Swiftb0y: The endWarning code was effectively already integrated (for every mode). It just alternated between the wheel being off and the color produced by the mode instead. You can change the color sent during the off-cycle easily. I didn't know you wanted red-white-alternating as the warning pattern.

Swiftb0y: primarily that the concerns are more cleanly separated. The stock wheelring is the component that implements the end_of_track warning logic. If you were to add another mode, you can just create another customized instance of the wheelring and you get the end_of_track logic for free. See the open-closed principle for more. In addition there are also a couple little things:
1. indicator_500ms is used wrong. You should respond to the actual value of the CO (1 is on, 0 is off) instead of tracking the state yourself (the `.blink` property you introduced). Otherwise depending on the exact timing, you may get a wheelrings that blink in phase and other times 180 degrees out-of-phase. 
2. returning some sentinel value `"EOT"` in getPositionColor makes it clear that your code is not separated into clean chunks (see "separation of concerns").
3. the need for `fix_color` is yet another workaround that indicates the your solution is suboptimal. in the solution I posted it should automatically use the correct color (at least in principle; there may be still be bugs that occur when coding blind, I'm happy to fix them if you report those bugs to me).

WaylonR: Your revised code was fixed, and pushed. Made the adjustment to outputMaybe, to output via the output code, rather than the prototype, which bypassed each modes output. adjusted the endwarning code to blink red/white as otherwise, if one had a red track, no warning would happen. Same if it was a white track.

Swiftb0y: Thank you. It seems you forgot to push however.

WaylonR: sorry, will do that now. Got distracted.

WaylonR: Pushed

Swiftb0y (approved):
I think thats a LGTM. Thanks for your continued patience. Did you already open a manual PR? I can't find one.

WaylonR:
> I think thats a LGTM. Thanks for your continued patience. Did you already open a manual PR? I can't find one.
Haven't done a manual PR for it. I need to do that? I also need to set the default color on all decks to "TRACK" before we send this for merging. Also, did you want a credit in it, for usage of the code from your mapping, and your assistance?
What's your byline? Your mapping didn't include one.

Swiftb0y:
> Haven't done a manual PR for it. I need to do that? 

yes: https://github.com/mixxxdj/mixxx/wiki/Contributing%20mappings#documenting-the-mapping

> I also need to set the default color on all decks to "TRACK" before we send this for merging.

Sure. feel free to commit that now. 

> Also, did you want a credit in it, for usage of the code from your mapping, and your assistance?

I appreciate the offer but there is no need ;)

WaylonR:
Push done.

acolombier:
Hey @WaylonR, are you still looking at completing this? I'll make that a draft PR for now, but feel free to mark it ready when you have prepared the manual PR and it should be reconsidered for merging.

JoergAtGithub:
@WaylonR Why did you merge the `main` branch? Unless your mapping uses features, which only exist in the main branch, the target for new mappings is the stable Mixxx branch, which is currently `2.5`

WaylonR:
> @WaylonR Why did you merge the `main` branch? Unless your mapping uses features, which only exist in the main branch, the target for new mappings is the stable Mixxx branch, which is currently `2.5`

Did not realise I merged the main branch, by pressing 'sync', My code was done for 2.4, thought that it was a good idea to sync first, then continue work. My apoligies. There was no reason for merging to the main branch. Good catch, thank you, sorry, reverting the commit.

PR #13435, opened 2024-07-03
Paintable cleanup

JoergAtGithub:
Removed 2 unused methods
Changed code to return early style
Implemented missing SVG branch for Paintable::toImage()
Replaced hardcoded transparency color by Qt::transparent
Use std::move instead reset/release

### DIFF
--- a/src/widget/paintable.cpp
+++ b/src/widget/paintable.cpp
@@ -94,8 +94,7 @@ Paintable::Paintable(const PixmapSource& source, DrawMode mode, double scaleFact
                 m_pSvg->render(&painter);
                 WPixmapStore::correctImageColors(&copy_buffer);
 
-                m_pPixmap.reset(new QPixmap(copy_buffer.size()));
-                m_pPixmap->convertFromImage(copy_buffer);
+                m_pPixmap = std::make_unique<QPixmap>(QPixmap::fromImage(copy_buffer));
         }
     }
 }
--- a/src/widget/paintable.cpp
+++ b/src/widget/paintable.cpp
@@ -62,15 +62,16 @@ Paintable::Paintable(const PixmapSource& source, DrawMode mode, double scaleFact
             m_pPixmap = std::move(pPixmap);
     } else {
         auto pSvg = std::make_unique<QSvgRenderer>();
-        if (!source.getPath().isEmpty()) {
-                if (!pSvg->load(source.getPath())) {
-                    // The above line already logs a warning
-                    return;
-                }
-        } else {
+        if (source.getPath().isEmpty()) {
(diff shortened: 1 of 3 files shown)


src/widget/paintable.cpp
```diff
@@ -103,60 +104,74 @@ bool Paintable::isNull() const {
...
-    //   return m_pPixmap->toImage();
-    // relying on QPixmap returning QImage() when it was null.
-    return !m_pPixmap ? QImage() : m_pPixmap->toImage();
+    if (m_pPixmap) {
+        return m_pPixmap->toImage();
+    }
+
+    if (m_pSvg) {
+        QImage image(m_pSvg->defaultSize(), QImage::Format_ARGB32);
+        image.fill(Qt::transparent);
```
daschuer, line 161: Why is this required? Please add a source code comment.

daschuer (changes requested):
LGTM, only a minor comment. 

Reading the code I notice that we render the SVG over and over again when for example called here:

https://github.com/mixxxdj/mixxx/blob/c0b4f50082b27767305357551c4c8f18b4db95cc/src/widget/wbattery.cpp#L168

Does it make sense to keep a copy of the QImage, and skip all the logic in draw, if the size did not change? I am not sure if this works, because I don't know when the SVG is exactly rendered and if there is already a caching involved.

src/widget/paintable.cpp
```diff
@@ -103,60 +104,74 @@ bool Paintable::isNull() const {
...
-    // This confusion let to the wrong assumption that we could simple
-    //   return m_pPixmap->toImage();
-    // relying on QPixmap returning QImage() when it was null.
-    return !m_pPixmap ? QImage() : m_pPixmap->toImage();
+    if (m_pPixmap) {
+        return m_pPixmap->toImage();
+    }
+
+    if (m_pSvg) {
+        QImage image(m_pSvg->defaultSize(), QImage::Format_ARGB32);
```
Swiftb0y, line 160:
```suggestion
        QImage image(m_pSvg->defaultSize(), QImage::Format_ARGB32_Premultiplied);
```
image

JoergAtGithub:
> Reading the code I notice that we render the SVG over and over again when for example called here:
> 
> https://github.com/mixxxdj/mixxx/blob/c0b4f50082b27767305357551c4c8f18b4db95cc/src/widget/wbattery.cpp#L168

Yes we do, this is addressed in #12904 for 2.4 and not in this PR.

Swiftb0y (approved):
LGTM. waiting for @daschuer

daschuer (approved):
Thank you

PR #4550, opened 2021-12-05
Components ES6 rewrite with improvements

Swiftb0y:
Reopen #4171 with Be-ing#56 on top. 

In the old PR, there is still a review from @Holzhaus which has not been addressed, though I plan to take care when I find the time.

PR #13437, opened 2024-07-03
ES6+ ComponentsJS

christophehenry:
Superseeds #4550

This is an attempt at rewriting `midi-components-0.0.js` with modern JS and correct type hinting fr IDEs.

I'm starting small. Component by component in order to get regular feedbacks, starting with the general `Component`.

Some rationals guiding the development of the new API and design decisions (will be updated as development goes):
1. The API is developped for eas of use first. It priorizes readability and documentation with JSDoc to ease IDEs type hinting.
2. The `Component`'s constructor doesn't not merge all of its parameters into `this`. This is to avoid multiple ways of overriding `Component`'s behavior and easier documenting constructor parameters with JSDoc. If you need a more specific behavior, you should create a new class extending the original component.

### DIFF
--- a/res/controllers/common-controller-scripts.js
+++ b/res/controllers/common-controller-scripts.js
@@ -141,6 +141,77 @@ var script = function() {
 };
 
 /**
+ * Wraps the engine timer API in JS promises
+ *
+ * Use `new TimerPromise()` to call `engine.beginTimer`, use {@link TimerPromise#cancel} to call `engine.stopTimer()`.
+ * The Promise resolves the first time the timer completes and rejects when canceled.
+ */
+class TimerPromise extends Promise {
+    /**
+     * @param {number} interval Time in milliseconds until the function is executed. Intervals below 20ms are ignored
+     * @param {boolean} oneShot If true the function is only once, if false the function is executed repeatedly and the
+     *          promise resolves the first time the timer completes. [default = false]
+     */
+    constructor(interval, oneShot = false) {
+        let resolve = undefined;
+        let reject = undefined;
+        super((res, rej) => {
+            resolve = res;
+            reject = rej;
+        });
+
+        this._resolve = resolve;
+        this._reject = reject;
+
+        this.__timer = 0;
+        if (Number.isInteger(interval) && interval >= 20) {
+            this.__timer = engine.beginTimer(interval, () => {
+                this.__done = true;
+                this.__canceled = false;
+                this._resolve();
+            }, oneShot);
+        }
+
+    }
+
+    get settled() {
+        return this.__done || this.__canceled;
+    }
+
+    get done() {
+        return this.__done;
+    }
+
+    get canceled() {
+        return this.__canceled;
+    }
+
+    cancel() {
+        if (this.__done || this.__canceled) { return; }
+        this.__done = false;
+        this.__canceled = true;
+        if (this.__timer !== 0) {
+            engine.stopTimer(this.__timer);
+        }
+        this._reject();
+    }
+}
+
+TimerPromise.resolve = function resolve(value) {
+    const promise = new TimerPromise(0);
+    promise._resolve(value);
+    return promise;
+};
+
+TimerPromise.cancelled = function cancelled() {
+    const promise = new TimerPromise(0);
+    promise.cancel();
+    return promise;
+};
+
+script.TimerPromise = TimerPromise;
+
+/**
  * Discriminates whether an object was created using the `{}` synthax.
  *
  * Returns true when was an object was created using the `{}` synthax.
--- a/res/controllers/midi-components-0.1.js
+++ b/res/controllers/midi-components-0.1.js
@@ -1,0 +1,219 @@
+/**
+ * Components JS library for Mixxx
+ * Documentation is on the Mixxx wiki at
+ * https://github.com/mixxxdj/mixxx/wiki/Components-JS
+ *
+ * Copyright (C) 2017 Be <be.0@gmx.com>
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+
+(function(global) {
+    /**
+     * @typedef ComponentOpts
+     * @property {number=} midiStatus
+     * @property {number=} midiNo
+     * @property {string=} group
+     * @property {string=} inKey
+     * @property {string=} outKey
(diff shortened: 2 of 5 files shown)


Swiftb0y:
Can you also add a dummy mapping that makes use of the components so the QJSEngine actually runs the code? I don't think it would work in the current form  you're using features (`static`, private properties, spread syntax) not available in QJSEngine atm (~ES7 is the maximum IIRC).

Swiftb0y:
Thank you for picking this up btw.

res/controllers/midi-components-0.1.js
```diff
@@ -0,0 +1,401 @@
...
+                } else { // Unbound action
+                    console.debug(`Unbound action to ${evts}`);
+                }
+            });
+        }
+
+        // eslint-disable-next-line no-unused-vars
+        isPress(channel, control, value, status) {
+            return value > 0;
+        }
```
Swiftb0y, line 330: We prefer the `_` prefix tactic instead. Or do you have some rationale why the `eslint-disable` is better here?
```suggestion
        isPress(_channel, _control, value, _status) {
            return value > 0;
        }
```

res/controllers/midi-components-0.1.js
```diff
@@ -0,0 +1,401 @@
...
+            this.shiftOffset = shiftOffset;
+            this.sendShifted = sendShifted;
+
+            if (this.outConnect) {
+                this.connect();
+                this.outTrigger ? this.trigger() : this.noop();
+            }
+        }
+
+        noop() { /* Does nothing */ }
```
Swiftb0y, line 75: I know this is a nitpick, but a `noop` is not the same as a mathematical identity function. The latter is useful in mathematical contexts, the former is code smell (using a ternary merely for its sideffects is one as well). KISS.
```suggestion
                if (this.outTrigger) {
                    this.trigger();
                }
            }
        }
```

res/controllers/midi-components-0.1.js
```diff
@@ -0,0 +1,401 @@
...
+                this.onLongPress = onLongPress.bind(this);
+            }
+
+            if (onDoublePress instanceof Function) {
+                this.onDoublePress = onDoublePress.bind(this);
+            }
+
+            this.__btnEvts = new InputTimeWindow(this.longPressTimeout, Object.values(Button.Events), evts => {
+                if (evts[Button.Events.PRESS] === 1 && evts[Button.Events.RELEASE] === 1) {
+                    this.onPress();
```
Swiftb0y, line 316: I may be misinterpreting the code, but from what I could tell this abstraction would only call the callback after it has collected all the events for `this.longPressTimeout` milliseconds. That means that even simple button presses where you just want to run code once the button has been pressed (not even waiting for the release) would introduce this huge artificial latency. This is not an option, input latency has to be reduced to a sensible minimum (essentially as much as possible, though I wouldn't sacrifice other aspects such as code quality for it).

res/controllers/midi-components-0.1.js
```diff
@@ -0,0 +1,401 @@
...
+            this.outConnect = outConnect;
+            this.outTrigger = outTrigger;
+            this.max = max;
+            this.shiftOffset = shiftOffset;
+            this.sendShifted = sendShifted;
+
+            if (this.outConnect) {
+                this.connect();
+                this.outTrigger ? this.trigger() : this.noop();
+            }
```
Swiftb0y, line 72: This strictly typed approach looses the feature where the user can add extra properties to the type when instantiating it. I think that is one of the strengths in QML and components-0.0. I fear the usage in mappings will get too clunky if we remove that feature. See [my reply on zulip](https://mixxx.zulipchat.com/#narrow/stream/113295-controller-mapping/topic/Pioneer.20DDJ-SR2/near/448621304) where I listed a bunch of usage examples where I used that pattern to avoid more broadly scoped (global) variables.

res/controllers/common-controller-scripts.js
```diff
@@ -140,6 +140,77 @@ var colorCodeToObject = function(colorCode) {
 var script = function() {
 };
 
+/**
+ * Wraps the engine timer API in JS promises
+ *
+ * Use `new TimerPromise()` to call `engine.beginTimer`, use {@link TimerPromise#cancel} to call `engine.stopTimer()`.
+ * The Promise resolves the first time the timer completes and rejects when canceled.
+ */
+class TimerPromise extends Promise {
```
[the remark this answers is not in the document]
christophehenry: I am not aware of a previous discussion on this matter.

So the idea here is to offer a more modern and ES6+ way of using timers. Promises a particularly suited for this kind of task.

Of course, since Promises only resolve once, it makes more sense for one-shot timers. But the `cancel()` method I added also handle the case where the timer was already cancelled which has the advantage of avoiding the warning when trying to `engine.stopTimer()` a one-shot that was already executed. This warning message really bothers me.

Of course, if this is too controversial, I'll remove this.

res/controllers/midi-components-0.1.js
christophehenry: The problem with the `_` prefix strategy is that it propagates when you override the method using IDEs code gen like Webstorm's. IMO the one `_` strategy is suited for end users of an API but not when designing the API. That a parameter is unused in the base class is not relevant since these methods are designed to be overriden.

res/controllers/midi-components-0.1.js
christophehenry: I don't agree but I won't fight over personal taste just to gain 2 lines of code. I'm ok with changing this 😉

res/controllers/midi-components-0.1.js
christophehenry: Yes, I agree. Which is why [this code don't execute if the `onLongPress` and `onDoublePress` methods are not overrided](https://github.com/mixxxdj/mixxx/pull/13437/files#diff-7e35655817545d6748ef8999d6bfb61a8ae3b1b8c4fdedf9977bbb23e3f9a949R343).

Sadly, as of now, it's not possible to bind several scripts to the same (midiStatus, midiNo) tuple. Which means you must use the same component if you want to use several interaction to the same button. I think it's an acceptable tradeoff.

Or we could modify the engine to execute multiple callbacks for the same midi signal.

res/controllers/midi-components-0.1.js
christophehenry: IMO, using the constructor parameter to override any component method defeats the purpose of using a class system to easier IDEs code completion and type hinting. If a user wants to override a component's methods without explicitly declaring a class, they can write the following:

`js
const comp = new class extends components.Component {
    inValueScale(value) {
        return value === 0x40 ? true : false;
    }
}({midiStatus: 0x70, midiNo: 0x40}
```

res/controllers/midi-components-0.1.js
christophehenry: Ok, reflecting about this, I think I may have a more reactive algorithm for this feature. It goes like this:
- the event bus recieves a push event:
  - if the event is the first detected, wait 50ms
  - after 50 ms,
    - if the bus received a release event, we're either on a single or double push, wait another 50ms,
        - if the bus didn't receive a second push event, we're on a single push, call the dispatch callback signaling a single push, reset the bus, end of cycle, reaction time: 100ms.
        - if the bus received a second push event, we're on a double push, call the , reset the bus, end of cycle, reaction time: 100ms.
    - if the bus didn't receive a release event, wait until the end of `longPressTimeout` and call the dispatch callback signaling a long push.
    - if a release event is received in between, then it's a single push.

Would that algorithm be more acceptable?

christophehenry:
Closing in favor of #13459.

PR #13448, opened 2024-07-07
Library: fix font reset in multiline comment editor

ronso0:
In the multi-line comment editor, with Qt6 and with a library font significantly taller than the default, I noticed this a few times:
* initially the editor font is the library font (size)
* add a line break and the editor font is reset (to app default? library default?)

I have no idea  what exactly was causing this, but at least I can't reproduce anymore when adopting the library font explicitly when constructung the editor.


m0dB:
LGTM

PR #13450, opened 2024-07-07
(fix) Sound preferences: don't set m_settingsModified in update slots

ronso0:
Currently the `m_settingsModified` is set in `DlgPrefSound::slotUpdate()` which will cause a sound pause when pressing **Apply** in any pref page, because `DlgPrefSound::slotApply()` would reapply the current (unchanged) sound config.
https://github.com/mixxxdj/mixxx/pull/13070#discussion_r1667683929

Fixed by moving the keylock dual-threading checkbox update to a dedicated method and call then when necessary.

@acolombier please check if I got everything right.

### DIFF
--- a/src/preferences/dialog/dlgprefsound.cpp
+++ b/src/preferences/dialog/dlgprefsound.cpp
@@ -216,6 +216,10 @@ DlgPrefSound::DlgPrefSound(QWidget* pParent,
             QOverload<int>::of(&QComboBox::currentIndexChanged),
             this,
             &DlgPrefSound::settingChanged);
+    connect(keylockComboBox,
+            QOverload<int>::of(&QComboBox::currentIndexChanged),
+            this,
+            &DlgPrefSound::updateKeylockDualThreadingCheckbox);
 #ifdef __RUBBERBAND__
     connect(keylockDualthreadedCheckBox,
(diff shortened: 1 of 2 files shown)


acolombier (approved):
I haven't had a chance to test it due to time, but reviewed the code, and I can see no missing logic. Thanks for the fix!

daschuer (approved):
LGTM and thank you for the control proxy refactoring

src/preferences/dialog/dlgprefsound.cpp
```diff
@@ -742,6 +751,9 @@ void DlgPrefSound::settingChanged() {
     m_settingsModified = true;
 
 #ifdef __RUBBERBAND__
+}
+
+void DlgPrefSound::updateKeylockDualThreadingCheckbox() {
```
fwcd, line 756: This `#ifdef` looks suspicious, does it work as intended? I'm getting some errors building for iOS (without Rubberband)...

fwcd: Seems to work, but the brace placement makes this confusing. I've fixed this and some related issues in

- https://github.com/mixxxdj/mixxx/pull/13577

Issue #13396, opened 2024-06-23
show decks 3/4 if AutoDJ picks one of them


ronso0:
It may happen that AutoDJ resorts to deck 3 and/or 4 in case deck 1 and/or 2 are not "left" and "right" decks.
If decks 3/4 are not shown in the GUI this can cause quite some confusion.

AutoDJProcessor could simply toggle decks 3/4 i it uses one/both.
The common control is `[Skin],show_4decks`
___
FWIW the crossfader assignment conditions/issue is not covered in https://manual.mixxx.org/2.4/en/chapters/djing_with_mixxx#auto-dj or https://manual.mixxx.org/2.4/en/chapters/library#library-auto-dj

PR #13455, opened 2024-07-09
AutoDJ: force-show decks 3/4 if we are going to use them

ronso0:
Avoids potential of confusion if AutoDJ resorts to deck 3 and/or 4 while deck 1/2 are assigned to crossfader center.

(obviously doesn't cover the case where controller mappings request 4 decks but e.g. Shade skin can't show them)

Fixes: https://github.com/mixxxdj/mixxx/issues/13396


daschuer (approved):
LGTM, Thank you.

Issue #13440, opened 2024-07-04
Create a reactive programming API for controllers


christophehenry:
### Feature Description

This is an issue that I've already risen in the past [proposing to add ability to pass arbitrary keys `engine.setValue` and `engine.setParameter`](https://github.com/mixxxdj/mixxx/issues/12766) but this seem to have met some opposition.

So I'd like to start a broader discussion about adding an API to do reactive programming in controller especially now that [I started working on ES6+ components JS](https://github.com/mixxxdj/mixxx/pull/13437).

What I propose is to reflect on a standard API for Mixxx, setting aside the implementation. This API could mimic [ES `Event` API](https://developer.mozilla.org/fr/docs/Web/API/Event) with methods like `dispatchEvent`, `addEventListener` and `removeEventListener` a be closer to Qt's signals/slots API (also I'm not sure what this would look like…).

This API could be standard to `Component` and `ComponentContainer`. 

Regarding the implementation, I propose a first implementation in pure JS using `EventTarget`, `CustomEvent` and `Proxy` (this seem supported by Mixxx' JS engine, also I need to formally check) that could be transparantly replaced by a C++ implem in the `engine` namespace thanks to the standard API.

Swiftb0y:
I'm slightly confused, is this about a reactive programming paradigm or sharing data between `QJSEngine` instances?

IMO the long term plan for enabling reactive programming in Mixxx would be to programm mappings in a `QQmlEngine` and use native QML reactive bindings. Have you looked into that?

christophehenry:
No I'm talking specifically about cross-component communication between components inside the same controller for complex use cases (like the one I exposed in #12766).

Correct me if I'm wrong but a `QQmlEngine` implementation is not comming any time soon?

Swiftb0y:
Well, actually #11407 did a good job at laying the foundations. Its currently controller screen focused, but I'm sure we can bend it to work for plain controllers as well.

christophehenry:
Ok, so `EventTarget` is not declared by `QJSEngine` so my first proposition of a pure JS implementation is off the table 😔

Swiftb0y:
Yeah... `EventTarget` [is part of the DOM Spec](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget#specifications)... 

Wdyt about the QML implementation?

christophehenry:
[...]

PR #13459, opened 2024-07-10
QML-based components API for controllers

christophehenry:
Ref: #13440

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2778,6 +2778,8 @@ if(QML)
     src/qml/qmlwaveformoverview.cpp
     src/qml/mixxxcontroller.cpp
     src/qml/mixxxcontroller.h
+    src/qml/mixxxscreen.cpp
+    src/qml/mixxxscreen.h
     # The following sources need to be in this target to get QML_ELEMENT properly interpreted
     src/control/controlmodel.cpp
     src/control/controlsortfiltermodel.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2776,6 +2776,8 @@ if(QML)
     src/qml/qmlplayerproxy.cpp
     src/qml/qmlvisibleeffectsmodel.cpp
     src/qml/qmlwaveformoverview.cpp
+    src/qml/mixxxcontroller.cpp
+    src/qml/mixxxcontroller.h
     # The following sources need to be in this target to get QML_ELEMENT properly interpreted
     src/control/controlmodel.cpp
     src/control/controlsortfiltermodel.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2777,9 +2777,7 @@ if(QML)
     src/qml/qmlvisibleeffectsmodel.cpp
     src/qml/qmlwaveformoverview.cpp
     src/qml/mixxxcontroller.cpp
-    src/qml/mixxxcontroller.h
     src/qml/mixxxscreen.cpp
-    src/qml/mixxxscreen.h
     # The following sources need to be in this target to get QML_ELEMENT properly interpreted
     src/control/controlmodel.cpp
     src/control/controlsortfiltermodel.cpp
--- a/res/controllers/Denon-DN-S3700.qml
+++ b/res/controllers/Denon-DN-S3700.qml
@@ -1,6 +1,6 @@
-import QtQml
+import QtQuick
 
-QtObject {
+Item {
     id: controller
 
     property string controllerId: ""
@@ -9,9 +9,10 @@ QtObject {
     function init(controllerId, debugMode) {
         controller.controllerId = controllerId;
         controller.debugMode = debugMode;
-        console.log(controllerId, debugMode);
+        console.error(controllerId, debugMode);
+        console.error(controller.controllerId, controller.debugMode);
     }
     function shutdown() {
-        console.log(`Shutting down ${controller.controllerId} with debug mode ${controller.controllerId}`);
+        console.error(`Shutting down ${controller.controllerId} with debug mode ${controller.controllerId}`);
     }
 }
--- a/res/controllers/Denon-DN-S3700.qml
+++ b/res/controllers/Denon-DN-S3700.qml
@@ -11,4 +11,10 @@ MixxxController {
     function shutdown() {
         console.error(`Shutting down ${controller.controllerId} with debug mode ${controller.debugMode}`);
     }
+
+    MixxxScreen {
+        screenId: "screen 7"
+        splashOff: 5000
+        Component.onCompleted: console.error(`MixxxScreen.identifier=${screenId} ${splashOff}`)
+    }
 }
--- a/res/controllers/Denon-DN-S3700.qml
+++ b/res/controllers/Denon-DN-S3700.qml
@@ -1,0 +1,17 @@
+import QtQml
+
+QtObject {
+    id: controller
+
+    property string controllerId: ""
+    property bool debugMode: false
+
+    function init(controllerId, debugMode) {
+        controller.controllerId = controllerId;
+        controller.debugMode = debugMode;
+        console.log(controllerId, debugMode);
+    }
+    function shutdown() {
+        console.log(`Shutting down ${controller.controllerId} with debug mode ${controller.controllerId}`);
+    }
+}
--- a/res/controllers/Denon-DN-S3700.qml
+++ b/res/controllers/Denon-DN-S3700.qml
@@ -1,18 +1,14 @@
-import QtQuick
+import QtQml
 
-Item {
+import "Mixxx"
+
+MixxxController {
     id: controller
 
-    property string controllerId: ""
-    property bool debugMode: false
-
-    function init(controllerId, debugMode) {
-        controller.controllerId = controllerId;
-        controller.debugMode = debugMode;
-        console.error(controllerId, debugMode);
+    function init() {
         console.error(controller.controllerId, controller.debugMode);
     }
     function shutdown() {
-        console.error(`Shutting down ${controller.controllerId} with debug mode ${controller.controllerId}`);
+        console.error(`Shutting down ${controller.controllerId} with debug mode ${controller.debugMode}`);
     }
 }
--- a/res/controllers/Denon-DN-S3700.qml
+++ b/res/controllers/Denon-DN-S3700.qml
@@ -5,16 +5,12 @@ import "Mixxx"
 MixxxController {
     id: controller
 
-    function init() {
-        console.error(controller.controllerId, controller.debugMode);
-    }
-    function shutdown() {
-        console.error(`Shutting down ${controller.controllerId} with debug mode ${controller.debugMode}`);
-    }
+    onInit: console.error(`Starting controller ${controller.controllerId} with debug mode ${controller.debugMode}`)
+    onShutdown: console.error(`Shutting down ${controller.controllerId} with debug mode ${controller.debugMode}`)
 
     MixxxScreen {
         screenId: "screen 7"
         splashOff: 5000
-        Component.onCompleted: console.error(`MixxxScreen.identifier=${screenId} ${splashOff}`)
+        onInit: console.error(`MixxxScreen.screenId=${screenId}, MixxxScreen.splashOff=${splashOff}`)
     }
 }
--- a/res/controllers/Denon-DN-S3700.qml
+++ b/res/controllers/Denon-DN-S3700.qml
@@ -12,5 +12,9 @@ MixxxController {
         screenId: "screen 7"
         splashOff: 5000
         onInit: console.error(`MixxxScreen.screenId=${screenId}, MixxxScreen.splashOff=${splashOff}`)
+        transformFrame: (frame, timestamp, area) => {
+            console.error(frame)
+            return new ArrayBuffer(0)
+        }
     }
 }
--- a/src/controllers/legacycontrollermappingfilehandler.cpp
+++ b/src/controllers/legacycontrollermappingfilehandler.cpp
@@ -390,7 +390,8 @@ void LegacyControllerMappingFileHandler::addScriptFilesToMapping(
         QFileInfo file = findScriptFile(mapping, filename, systemMappingsPath);
         if (file.suffix() == "qml") {
 #ifdef MIXXX_USE_QML
-            QString identifier = scriptFile.attribute("identifier", "");
+            QString identifier = scriptFile.attribute(
+                    "identifier", scriptFile.attribute("functionprefix", ""));
             mapping->addScriptFile(LegacyControllerMapping::ScriptFileInfo{
                     filename,
                     identifier,
--- a/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
@@ -828,11 +828,10 @@ bool ControllerScriptEngineLegacy::evaluateScriptFile(const QFileInfo& scriptFil
 }
 
 #ifdef MIXXX_USE_QML
-bool ControllerScriptEngineLegacy::
-        : instanciateQMLComponent(
-                  const LegacyControllerMapping::ScriptFileInfo& qmlScript,
-                  QMap<QString, std::shared_ptr<ControllerRenderingEngine>>
-                          availableScreens) {
+bool ControllerScriptEngineLegacy::instanciateQMLComponent(
+        const LegacyControllerMapping::ScriptFileInfo& qmlScript,
+        QMap<QString, std::shared_ptr<ControllerRenderingEngine>>
+                availableScreens) {
     // Like for Javascript, if the script is invalid, it should be watched so the user can fix it
     // without having to restart Mixxx. So, add it to the watcher before
     // evaluating it.
--- a/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
@@ -129,7 +129,7 @@ bool ControllerScriptEngineLegacy::callShutdownFunction() {
 
 #ifdef MIXXX_USE_QML
     for (const auto& controller : m_mixxxController) {
-        controller->shutdown();
+        emit controller->shutdown();
     }
 
     if (!m_bQmlMode) {
@@ -207,7 +207,7 @@ bool ControllerScriptEngineLegacy::callInitFunction() {
         }
 
         for (const auto& controller : m_mixxxController) {
-            controller->init();
+            emit controller->init();
         }
 
         QHashIterator<QString, std::shared_ptr<QQuickItem>> i(m_rootItems);
@@ -261,48 +261,7 @@ bool ControllerScriptEngineLegacy::callInitFunction() {
             // ControllerScriptEngineBase::handleQMLErrors
         }
 
-        QListIterator<std::shared_ptr<mixxx::qml::MixxxController>> controllers(m_mixxxController);
-        bool controllersSuccess = true;
-        while (controllers.hasNext()) {
-            const QMetaObject* metaObject = controllers.next()->metaObject();
-
-            VERIFY_OR_DEBUG_ASSERT(metaObject) {
-                qCWarning(m_logger) << "Invalid meta object for controller";
-                continue;
-            }
-
-            QMetaMethod initFunction;
-            bool typed = false;
-            int methodIdx = metaObject->indexOfMethod(kQmlComponentInitFunctionUntypedSignature);
-
-            if (methodIdx == -1 || !metaObject->method(methodIdx).isValid()) {
-                qCDebug(m_logger) << "QML controller has no valid untyped init method.";
-                methodIdx = metaObject->indexOfMethod(kQmlComponentFunctionTypedSignature);
-                typed = true;
-            }
-
-            initFunction = metaObject->method(methodIdx);
-
-            if (!initFunction.isValid()) {
-                qCDebug(m_logger) << "QML controller has no valid untyped init method; Skipping.";
-                continue;
-            }
-
-            qCDebug(m_logger) << "Executing init on QML controller";
-            if (typed) {
-                success &= initFunction.invoke(i.value().get(),
-                        Qt::DirectConnection,
-                        Q_ARG(QString, controllerName),
-                        Q_ARG(bool, m_logger().isDebugEnabled()));
-            } else {
-                success &= initFunction.invoke(i.value().get(),
-                        Qt::DirectConnection,
-                        Q_ARG(QVariant, controllerName),
-                        Q_ARG(QVariant, m_logger().isDebugEnabled()));
-            }
-        }
-
-        return success && controllersSuccess;
+        return success;
     }
 #endif
 }
--- a/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
@@ -95,6 +95,7 @@ bool ControllerScriptEngineLegacy::callFunctionOnObjects(
 
     const QJSValue global = m_pJSEngine->globalObject();
 
+    // TODO: ICI
     bool success = true;
     for (const QString& prefixName : scriptFunctionPrefixes) {
         QJSValue prefix = global.property(prefixName);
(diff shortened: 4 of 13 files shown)


res/controllers/Denon-DN-S3700.qml
```diff
@@ -0,0 +1,17 @@
+import QtQuick
```
Swiftb0y, line 1: we don't actually want QtQuick (since thats the GUI framework). You instead just want the language basics `QtQml`. You may also add a version (not sure whats appropriate, in the preexisting qml files we have in mixxx we use 2.12, but thats quite old (I think from Qt 5.12 still)).
```suggestion
import QtQml
```

Swiftb0y (commented):
this is roughly going into the direction I'm thinking. Keep it up.

christophehenry:
Ok, so the QML loading logic is really entangled with the screen handling logic. So now I'm thinking: Why not dropt the XML entirely and handle all the metadata in QML? That could be done using a `MixxxController` QML component. Then, when we detect and parse a QML file in `controllers`, if it contains a single `MixxxController`, then it's a controller, execute it!

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -458,6 +458,66 @@ bool ControllerScriptEngineLegacy::initialize() {
...
+                            std::shared_ptr<QQuickItem>(qobject_cast<QQuickItem*>(pRootObject));
+                    if (!rootItem) {
+                        qWarning("Oopsie run: Not a QQuickItem");
+                        delete pRootObject;
+                    }
+
+                    watchFilePath(script.file.absoluteFilePath());
+
+                    m_rootItems.insert("screenIdentifier", rootItem);
+                }
```
christophehenry, line 520: Ugh. No please don't comment. This is not here to stay…

JoergAtGithub:
> So now I'm thinking: Why not drop the XML entirely and handle all the metadata in QML?

That should be the target! One semantic syntax, with use of a programming language, where it is not possible to express it in general semantic language.

Important is, that our MIDI learning wizard can read all necessary information from this semantic information.

Swiftb0y:
The potential problem I have with that is that loading a QML module just to read its metadata may be too slow considering you need to parse >100 mappings on startup. I also don't like that it couples the metadata to a particular execution engine, making it impossible to easily try out new execution engines. The biggest issue is that there is no public QML (de-)serialization library. So manipulating the document as it would be required for a controller mapping wizard would not be possible. All in all, I'm not in favor of dropping the separate metadata manifest for now.

christophehenry:
What if [We create a custom C++ class](https://doc.qt.io/qt-6/qtqml-cppintegration-interactqmlfromcpp.html#accessing-qml-objects-via-well-defined-c-interfaces)? That should be well faster, no? All QML object have a C++ class counterpart, I can't believe Qt would release a new language if its heavy to parse.

Swiftb0y:
> I can't believe Qt would release a new language if its heavy to parse.

Sure, we can't be sure until we measured, but the other two concerns I've voiced still apply. 

>> we'd need to generate type definitions from the C++ APIs, etc
>
> Maybe this can be the occasion to do so. Can someone point me in the correct direction to do it?

Not sure, what you trying to achieve? I think these type definitions don't give us much more than better editor integration.

JoergAtGithub:
The parsing of the mappings at every startup (we do it 2x at for each XML file at each startup btw.) seems to be wrong to me anyway. As we need only very few data from these files at startup.
I think we should do this only at the very first startup, and then store the few needed values in a global file or database. If we have cached this information, we just need to check for modified files at the next startup.

Swiftb0y:
Sure... filehash should be sufficient to determine if something changed. But thats a separate topic IMO.

JoergAtGithub:
Since it would solve all the performance related concerns, it should be addressed first (not in this PR of cause).
Then we can define the new QML mapping syntax, without restrictions from the legacy implementation.

Swiftb0y:
sure... but again, performance wasn't my main concern. It was actually the fact that 

1. we can't make the metadata parsing independent from a QMLEngine and 
2. QML documents are essentially read-only. If we want to manipulate its content persistently (like its necessary for a mapping wizard), we would need to serialize it into a separate file in a format that we can actually manipulate.

daschuer:
It looks like we have conflicting or unclear  requirements here. We need to decide the way forward to not end up with a blocked PR or double work. 
We need to decide which of them are mandatory optional. 

Let's summarize (please correct me): 

User Requirements
* Good performance, responsive scratching (Real-Time) 
* Easy reassign buttons and sliders 
* Easy map a new controller from the scratch

Mapping Developer 
* Everything at one place, one file, folder, pack 
* Easy to share
* Reuse common programming skills 
* High abstraction level
* Complete API documentation 
* Good IDE support .. code completion.  
* On the fly edits, no Mixxx restart.
* Stable: Compatible to future Mixxx versions

Did I miss something? Please complete. 


> we can't make the metadata parsing independent from a QMLEngine 

That is correct if we use only one parser. I can image to workaround that by a stripped Mixxx QML parser. reading a file hash,  is in this context the minimum version of such parser, but we can read also other info. 

> We would need to serialize it into a separate file in a format that we can actually manipulate.

I can imagine QT does exactly that. Maybe we can copy from them and use it for our second parseing. 

The other topic is how the data stream is handled. I like to keep the connection factory and the dispatcher in the C++ domain, controlled by QML XML or JS to share concepts between the mapping types. I can also imagine to read the mapping by XML and write them as QML or such for an easy transition. 

Another point that bugs me is the weak scratching stability and the programming skills required to map the jogg wheel. Both can be targeted by handling the jog in wheel in the C++ domain and configure it from QML XML or JS. 

Does that make sense?

Swiftb0y:
I'm sorry but I don't think it makes sense to discuss this feature further in text form. The problem at hand is simply to complicated and widening the scope by bringing discussing the metadata format and currently irrelevant features such as scratching does not help! I will try to find some time and write down my thoughts in a sorted fashion for https://github.com/mixxxdj/proposals/. 
@christophehenry I would advise you to pause any work on this for now. We can resume once we've resolved the "too many cooks in the kitchen" problem brewing right now.

christophehenry:
It's better if I continue to work on this, which has the benefit to familiarise myself with the code.

Swiftb0y:
Sure... just be prepared that it probably won't get merged until we all agree on a design.

Swiftb0y:
> Or else: as a starting point, I can keep the XML

Yeah. I would keep the XML for now. Changing the metadata format is orthogonal to a potential QML API.

> and just design a rough QML API based on componentsJS.

 I don't think it makes sense to prototype a QML API outside a QML environment since the fact that that reactive bindings are available would change the API a lot. Or am I misunderstanding you here?

christophehenry:
I mean: I will keep the main design concepts of ComponentsJS to design the QML API, changing a few things for more flexibility.

Swiftb0y:
Yes, that would work. I think the main challenge is that with QML, we would need to insert a layer that adapts between the explicit message sending model to the implicit reactive binding model. For instance we would need to differentiate between when a binding was triggered (which can happen pretty much spuriously in a reactive system) for no good reason and when we actually do want to send a message. For that, I would introduce the concept of idempotence (each component should be by default idempotent, meaning sending the same message again would not introduce a visual difference and thus spontanous messages that are the same as the one before can be ignored). In the long-term I also want to experiment with abstracting the MIDI-onlyness out of the components, but thats a slightly different topic, though the idempotence concept would be added to the midi part of that abstraction). In short: in order to properly work with reactive binding and reduce overhead, components should swallow duplicate messages by default. Does that make sense?

daschuer:
I took the momentum and put my post into a proposals document: 
https://github.com/mixxxdj/proposals/pull/6

I think we need a separate documents to define single aspects. The requirements document shall make sure that all changes contributes to a common goal. From this point of view @christophehenry can carry on implementing a POC and also keeps his momentum. 

Does the current list of requirements match your ideas by the way?  

> TLDR; in order to properly work with reactive bindings and reduce overhead, components should swallow duplicate messages by default (inbound and outbound; separably configurable). Does that make sense?

That does not always work, because we do not always receive a release event to track double tabs. I think the CO system already covers all aspects we need to consider with these reactive bindings. 
This is controlled by https://github.com/mixxxdj/mixxx/blob/d61841ce9a16c618219d37d1a63f8f3a5a02dc43/src/control/controlobject.cpp#L13

Swiftb0y:
> I think we need a separate documents to define single aspects. 

I fully agree.

> Does the current list of requirements match your ideas by the way?

Not yet. I will prepare a proposal as well today. 

> That does not always work, because we do not always receive a release event to track double tabs.

Yes, I'm aware that it won't always work, but it should in the majority of cases and it should be easy to turn of on a per-component basis. I also don't think relying on / referring to COs works here because components may not always be bound to components. It also doesn't always work (for example the `track_loaded` CO will get signaled when loading a new track but its value doesn't change, thus there is also no reason to send a midi message if the action on the controller is idempotent (eg just controlling a feedback led)). Anyways, this is going far too deep into unnecessary detail again.

src/qml/mixxxcontroller.h
```diff
@@ -0,0 +1,30 @@
...
+namespace mixxx {
+namespace qml {
+
+class MixxxController : public QObject {
+    Q_OBJECT
+    QML_ELEMENT
+    Q_PROPERTY(QString controllerId READ controllerId WRITE setControllerId)
+    Q_PROPERTY(bool debugMode READ debugMode WRITE setDebugMode)
+  public:
+    QString controllerId();
```
christophehenry, line 17: So I get a:

```
/usr/bin/ld: .../moc_mixxxcontroller.cpp:119:(.text+0x177): undefined reference to `mixxx::qml::MixxxController::controllerId()'
```

on this. It's not clear to me whether the `Q_PROPERTY` macro generates the accessors definition or not. If not, I don't really get the usage of this macro…

Swiftb0y: > It's not clear to me whether the Q_PROPERTY macro generates the accessors definition or not. It does not generate accessors. Its there to wire up the classic C++ setter/getter and signal logic to a single property available to QML. This does usually result in a lot of boilerplate. If you don't need the explicit access control, you can just specify a `MEMBER` and eliminate the need for accessors. 

> So I get a:
the issue is that you _declared_ a `controllerId` member function but never _defined_ (implemented) it.

res/controllers/Denon-DN-S3700.qml
```diff
@@ -0,0 +1,16 @@
+import QtQml
+
+import "Mixxx"
+
+import "." as Skin
```
Swiftb0y, line 5: also no need afaict?

src/qml/mixxxcontroller.h
Swiftb0y: see https://doc.qt.io/qt-6/properties.html

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -863,34 +867,40 @@ bool ControllerScriptEngineLegacy::evaluateScriptFile(const QFileInfo& scriptFil
 }
 
 #ifdef MIXXX_USE_QML
-std::shared_ptr<QQuickItem> ControllerScriptEngineLegacy::loadQMLFile(
+bool ControllerScriptEngineLegacy::instanciateQMLComponent(
```
christophehenry, line 870: So this function here is a bit of a mess right now but I don't see how simplify this, unless I make screen delcaration a part of `MixxxController`, which is doable I think?

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -201,6 +205,11 @@ bool ControllerScriptEngineLegacy::callInitFunction() {
             qCDebug(m_logger) << "Unhandled controller JS error is:"
                               << m_pJSEngine->catchError().toString();
         }
+
+        for (const auto& controller : m_mixxxController) {
+            controller->init();
```
christophehenry, line 210: So this doesn't seem to be overriden from QML, sadly. I need to dig up a bit more…

christophehenry:
So following [r1677677964](https://github.com/mixxxdj/mixxx/pull/13459#discussion_r1677677964), I propose we drop <screens> from the XML in favor of declaring them in `MixxxController`. From a user standpoint, this can be done like this :

```qml
MixxxController {
    screens: {
        screenId: Item {
          …
        } 
    } 
}
```

Or using a dedicated class like this:

```qml
MixxxController {
    screens: [
        Screen {
            screenId: … 
        } 
    ] 
}
```

I'm rather in favor of a dedicated class so that all the screen-related code can be extracted from `controllerscriptenginelegacy.cpp` into this new class.

christophehenry:
So following [r1677677964](https://github.com/mixxxdj/mixxx/pull/13459#discussion_r1677677964), I propose we drop <screens> from the XML in favor of declaring them in `MixxxController`. From a user standpoint, this can be done like this :

```qml
MixxxController {
    screens: {
        screenId: Item {
          …
        } 
    } 
}
```

Or using a dedicated class like this:

```qml
MixxxController {
    screens: [
        Screen {
            screenId: … 
        } 
    ] 
}
```

I'm rather in favor of a dedicated class so that all the screen-related code can be extracted from `controllerscriptenginelegacy.cpp` into this new class.

res/controllers/Denon-DN-S3700.qml
```diff
@@ -0,0 +1,14 @@
...
+
+MixxxController {
+    id: controller
+
+    function init() {
+        console.error(controller.controllerId, controller.debugMode);
+    }
+    function shutdown() {
+        console.error(`Shutting down ${controller.controllerId} with debug mode ${controller.debugMode}`);
+    }
```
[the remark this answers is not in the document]
acolombier: These hooks will be invoked on component instantiation and destruction. Assuming this was inspired from the existing screen rendering, `init` and `shutdown` are called at different time.

- `init` is called once the controller has started. It is the direct equivalent to the same hook in Javascript. Unlike the `constructor` (equivalent to `Component.onCompleted`), it is invoked once the engine is running.
- `shutdown` is called when the engine is about to shutdown. This allows the controller to perform some animation or screen clean up, which is required for some devices (e.g restoring the splash screen, playing a shutdown animation, clearing the screen to a static colour, ...). In case of `Component.onDestruction`, as the name suggest, you don't have a component to render anymore. 

In case of "pure" controller (without UI rendering), I think it makes sense to ditch these function, but if we want to eventually merge the engines together, they will have to remain. 
I believe you might be able to replace them with signals instead tho (e.g `MixxxController::onStart`, `MixxxController::onShutdown`)

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -251,7 +260,49 @@ bool ControllerScriptEngineLegacy::callInitFunction() {
             // connection QQmlEngine::warnings ->
             // ControllerScriptEngineBase::handleQMLErrors
         }
-        return success;
+
+        QListIterator<std::shared_ptr<mixxx::qml::MixxxController>> controllers(m_mixxxController);
+        bool controllersSuccess = true;
+        while (controllers.hasNext()) {
```
acolombier, line 266: Please use a for-loop instead

```suggestion
        bool controllersSuccess = true;
        for (auto controller: m_mixxxController) {
```

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -863,34 +869,40 @@ bool ControllerScriptEngineLegacy::evaluateScriptFile(const QFileInfo& scriptFil
...
+
     QQmlComponent qmlComponent = QQmlComponent(
             std::dynamic_pointer_cast<QQmlEngine>(m_pJSEngine).get());
 
-    QFile scene = QFile(qmlScript.file.absoluteFilePath());
-    if (!scene.exists()) {
-        qCWarning(m_logger) << "Unable to load the QML scene:" << qmlScript.file.absoluteFilePath()
+    QFile file = QFile(qmlScript.file.absoluteFilePath());
+    if (!file.exists()) {
+        qCWarning(m_logger) << "Unable to load the QML file:" << qmlScript.file.absoluteFilePath()
```
acolombier, line 892: I believe `scene` is the right terminology. I would rather we keep it instead of the cryptic `file`

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
acolombier: I'm not sure moving the scene binding logic in this function helps much to be honest. Perhaps you should move the logic from line `460` to a new function, but I would suggest not making  `loadQMLFile` more complex.

src/controllers/legacycontrollermappingfilehandler.cpp
```diff
@@ -390,7 +390,8 @@ void LegacyControllerMappingFileHandler::addScriptFilesToMapping(
         QFileInfo file = findScriptFile(mapping, filename, systemMappingsPath);
         if (file.suffix() == "qml") {
 #ifdef MIXXX_USE_QML
-            QString identifier = scriptFile.attribute("identifier", "");
+            QString identifier = scriptFile.attribute(
+                    "identifier", scriptFile.attribute("functionprefix", ""));
```
acolombier, line 394: In case of the screens, it makes no sense to default to `functionprefix`

acolombier (commented):
I would suggest you use the `Dummy Device Screen.hid.xml` (Doc for [Linux](https://github.com/mixxxdj/mixxx/blob/676cd906dad5788c5faff7a3500d2f277d258296/tools/README#L5)) provided as part of the code to make sure you aren't breaking the UI rendering. I haven't tested it yet, but from a quick code review, I believe there is a few things that won't work.

christophehenry:
> I believe there is a few things that won't work.

I'm pretty sure there is a few things that won't work. This is me right now:



But I believe the screens code can be greatly simplified if separated in a QML component which is what I started to do. There's no logic yet bit [res/controllers/Denon-DN-S3700.qml](https://github.com/mixxxdj/mixxx/pull/13459/files#diff-a5da07ac185d2e1184ccec0649eb4e35b6549872ba6fc9efd23fe7fb1745b6b3) shows how it would look like. I'd like your opinion. @JoergAtGithub and @Swiftb0y too.

acolombier:
I'm not sure this will change much since there is close to no logic with the root item (thus why interpreted as a QQuickItem)
One gain you may get is the hooks to be replaced to signal when asynchronous, and nicer callback for the transform function, which may simplify the whole method discovery. 
I'm not sure there is much to simplify on the scene binding and rendering engine init.
It's worth to point out that, as it stands, the QML scene cannot be used to perform interactions with the controller, other than UI rendering. This is due to the fact the rendering is throttled to a certain (max) rate, but also cannot be done in the controllers thread since this would impact performance, so latency. 

The reason why all of this is like this is... _legacy_ :)

I would suggest to watch the proposals currently getting discussed, as I'm not entirely sure it makes sense to make QML support in the legacy engine, since there is a lot of limitations there, and it is likely not going to yield to great result because of those limitations.

christophehenry:
> This is due to the fact the rendering is throttled to a certain (max) rate, but also cannot be done in the controllers thread since this would impact performance, so latency.

Damn. I was hoping to move that code into the `MixxxScreen` class…

christophehenry:
Uh… Can I have some help on the build error?

src/qml/mixxxscreen.h
```diff
@@ -0,0 +1,77 @@
...
+// Created by augier on 15/07/24.
+//
+
+#ifndef MIXXX_MIXXXSCREEN_H
+#define MIXXX_MIXXXSCREEN_H
+
+#include <QImage>
+#include <QObject>
+#include <QSize>
+#include <QtQmlIntegration>
```
Swiftb0y, line 11: this header doesn't exist. Where did you get this from?

christophehenry: I don't remember. It provides `QML_ELEMENT`. But it appears `<QtQml/qqmlregistration.h>` does the same.

christophehenry: It is described [here](https://doc.qt.io/qt-6/qt-generate-foreign-qml-types.html).

christophehenry:
Ok, so previous compilation error is solved. [Moving on to the one that was causing problem locally](https://github.com/mixxxdj/mixxx/actions/runs/9954472109/job/27500162906?pr=13459#step:20:484)

christophehenry:
Ok, I'm deep into compilation errors, here. I'd appreciate hel.

CMakeLists.txt
```diff
@@ -2776,6 +2776,9 @@ if(QML)
     src/qml/qmlplayerproxy.cpp
     src/qml/qmlvisibleeffectsmodel.cpp
     src/qml/qmlwaveformoverview.cpp
+    src/qml/mixxcomponentlifecycle.cpp
```
christophehenry, line 2779: I'm not sure this here is usefule. `MixxxComponentLifecycle` has no implementation since it's an interface but compilation fails without it.

Swiftb0y: yeah, I think thats due to the MOC. Just keep it.

src/qml/mixxxscreen.cpp
```diff
@@ -3,6 +3,37 @@
...
+    int untypedIdx = meta->indexOfMethod(kScreenTransformFunctionUntypedSignature);
+    if (typedIdx >= 0) {
+        auto transformMethod = meta->method(typedIdx);
+        if (!transformMethod.isValid()) {
+            // TODO
+        } else {
+            transform = [transformMethod, this](const QByteArray input,
+                                const QDateTime& timestamp) -> QVariant {
+                return transform(transformMethod, input, timestamp, true);
+            };
```
christophehenry, line 21: This should be a bit more efficient than previous implementation since transform function is searched only once trhough relfection. But then again, I know very few C++ so I may awefully wrong.

src/qml/mixxxscreen.cpp
```diff
@@ -3,6 +3,37 @@
...
+}
+
+void MixxxScreen::componentComplete() {
+    const auto* const meta = metaObject();
+    int typedIdx = meta->indexOfMethod(kScreenTransformFunctionTypedSignature);
+    int untypedIdx = meta->indexOfMethod(kScreenTransformFunctionUntypedSignature);
+    if (typedIdx >= 0) {
+        auto transformMethod = meta->method(typedIdx);
+        if (!transformMethod.isValid()) {
+            // TODO
```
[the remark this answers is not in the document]
christophehenry: How do I get the engine from a QML element? I can see that [`QQuickView` has an `engine()`](https://doc.qt.io/qt-6.5/qquickview.html#engine) which I suppose is because [it takes the `QQmlEngine` has ctor argument](https://doc.qt.io/qt-6.5/qquickview.html#QQuickView-1) but I don't understand where it gets this argument from. In my code, if I try put any other ctor than `QObject`'s default, element stops being creatable.

src/qml/mixxxscreen.h
```diff
@@ -0,0 +1,105 @@
...
+    bool m_reversedColor = false;
+    // Whether or not the screen is allowed to receive bare data,
+    // not transformed.
+    bool m_rawData = false;
+    // The item to render
+    QQuickItem m_item;
+    // Transform function
+    std::function<QVariant(const QByteArray, const QDateTime&)> transform =
+            [](const QByteArray input, const QDateTime& timestamp) {
+                return QVariant(input);
```
christophehenry, line 87: I'd better use `std::expected` here with an error type enum instead here.

src/qml/mixxxscreen.h
```diff
@@ -0,0 +1,105 @@
...
+    Q_CLASSINFO("DefaultProperty", "item")
+
+  public:
+    enum class ColorEndian {
+        Big = static_cast<int>(std::endian::big),
+        Little = static_cast<int>(std::endian::little),
+    };
+    Q_ENUM(ColorEndian)
+
+    explicit MixxxScreen(MixxxController* parent = nullptr)
```
christophehenry, line 49: So I did the test and `parent` here is `nullptr`. How can I force a QML component to be instanciated inside a `MixxxController`.

christophehenry:
So code in the last commit is not good, but it's a glance of the solution I have in mind to allow `MixxxScreen`s to declare themselves to the `ScriptControllerEngine`. The basic idea is that `ScriptControllerEngine` exposes a small API via an interface to the QML components that they can retrieve via the `QQmlEngine`.

If someone has a more elegant solution, I take it.

christophehenry:
Let's put this on pause for now.

Issue #13457, opened 2024-07-09
Deprecate `lodash.mixxx.js` to be removed in a future version


christophehenry:
### Feature Description

Mixxx has been supporting ES6 features since 2.4.0. Lodash implement features that were missing in ES5 but most of them got resolved with ES6 and ES7. As of now, all the current controllers use is  `_.merge()`, `_.assign()`, which can be replaced by `Object.assign()` in most cases, `_.forEach`, which can be replaced by `Array.prototype.forEach()` and `_.filter()` which can be replaced by `Array.prototype.filter()`.

[`script.deepMerge()`](https://github.com/mixxxdj/mixxx/blob/main/res/controllers/common-controller-scripts.js#L168) could also be deprecated in 2.5.0 in favor of `Object.assign()` since I'm not really sure it really has a broader usage. Currently it has been only used in [`ComponentContainer.applyLayer()`](https://github.com/mixxxdj/mixxx/blob/main/res/controllers/midi-components-0.0.js#L664) since [dropping lodash dependency in ComponentJS](https://github.com/mixxxdj/mixxx/pull/12779).

Anyway, less code is better.

Swiftb0y:
I'm in favor. Though care needs to be taken when doing the replacement as lodash and built-in methods work slightly differently on sparse arrays (for example`Array.prototype.forEach()` and `_.forEach`, for their `map()` equivalents).

christophehenry:
Can you expand on how they work differently?

Swiftb0y:
:shrug: yeah. Idk. just saying that if you're replacing code like `_.map(Array(N), f)` it needs to be something like `Array(N).fill(0).map(f)`

christophehenry:
Maybe we could open a wiki migration guide?

Swiftb0y:
Yeah, unfortunately we had a problem of people abusing wiki pages for distributing malware so we had to lock it down. I don't have the time to write a guide myself but I can offer to upload a page for you once you have written it.

christophehenry:
Good. I'll propose something tonight.

christophehenry:
[Here you go.](https://github.com/user-attachments/files/16147781/lodash-migration.md)

Swiftb0y:
Done, thank you. Made some edits, let me know if you have a problem with it. https://github.com/mixxxdj/mixxx/wiki/Lodash-Migration

christophehenry:
Nice, thank you. I'll open a PR resolving this.

[...]

PR #13460, opened 2024-07-10
Deprecate `lodash.mixxx.js`, and `script.deepMerge`

christophehenry:
As decided in #13457.

### DIFF
--- a/res/controllers/common-controller-scripts.js
+++ b/res/controllers/common-controller-scripts.js
@@ -166,6 +166,8 @@ script.isSimpleObject = isSimpleObject;
  * @param source {object | Array} Object to merge into source
  */
 const deepMerge = function(target, source) {
+    console.warn("script.deepMerge is deprecated; use Object.assign instead");
+
     if (target === source || target === undefined || target === null || source === undefined || source === null) {
         return;
     }
@@ -190,6 +192,7 @@ const deepMerge = function(target, source) {
     }
 };
 
+/** @deprecated Use {@link Object.assign} instead */
 script.deepMerge = deepMerge;
 
 // ----------------- Mapping constants ---------------------
--- a/res/controllers/common-controller-scripts.js
+++ b/res/controllers/common-controller-scripts.js
@@ -158,14 +158,13 @@ const isSimpleObject = function(obj) {
     return obj !== null && typeof obj === "object" && obj.constructor.name === "Object";
 };
 
-script.isSimpleObject = isSimpleObject;
-
 /**
  * Deeply merges 2 objects (Arrays and Objects only, not Map for instance).
  * @param target {object | Array} Object to merge source into
  * @param source {object | Array} Object to merge into source
+ * @deprecated Use {@link Object.assign} instead
  */
-const deepMerge = function(target, source) {
+script.deepMerge = function(target, source) {
     console.warn("script.deepMerge is deprecated; use Object.assign instead");
 
     if (target === source || target === undefined || target === null || source === undefined || source === null) {
@@ -191,9 +190,6 @@ const deepMerge = function(target, source) {
         });
     }
 };
-
-/** @deprecated Use {@link Object.assign} instead */
-script.deepMerge = deepMerge;
 
 // ----------------- Mapping constants ---------------------
 
--- a/res/controllers/lodash.mixxx.js
+++ b/res/controllers/lodash.mixxx.js
@@ -8,6 +8,12 @@
  * Copyright Jeremy Ashkenas, DocumentCloud and Investigative Reporters & Editors
  */
 ;(function() {
+  const DEPRECATION_MSG = (
+      "Lodash is deprecated and will be removed in a future release of Mixxx; refer to the page " +
+      "https://github.com/mixxxdj/mixxx/wiki/Lodash-Migration for migrations instructions"
+  );
+
+  console.warn(DEPRECATION_MSG);
 
   /** Used as a safe reference for `undefined` in pre-ES5 environments. */
   var undefined;
@@ -17068,8 +17074,17 @@
   /*--------------------------------------------------------------------------*/
 
   // Export lodash.
+  /**
+   * @deprecated since 2.5.0
+   * @global
+   */
   var _ = runInContext();
 
   // Export to the global object.
-  root._ = _;
+  Object.defineProperty(root, "_", {
+    get() {
+      console.warn(DEPRECATION_MSG);
+      return _;
+    }
+  })
 }.call(this));
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -651,7 +651,13 @@
             // Unset isShifted for each ComponentContainer recursively
             this.isShifted = false;
         },
+        /**
+         * @param newLayer Layer to apply to this
+         * @param reconnectComponents Whether components should be reconnected or not
+         * @deprecated since 2.5.0. Use @{ComponentContainer#setLayer} instead
+         */
         applyLayer: function(newLayer, reconnectComponents) {
+            console.warn("ComponentContainer.applyLayer is deprecated; use ComponentContainer.setLayer instead");
             if (reconnectComponents !== false) {
                 reconnectComponents = true;
             }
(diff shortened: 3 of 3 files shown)


res/controllers/midi-components-0.0.js
```diff
@@ -670,6 +676,26 @@
...
+            Object.assign(this, newLayer);
+
+            if (reconnectComponents === true) {
+                this.forEachComponent(function(component) {
+                    component.connect();
+                    component.trigger();
+                });
+            }
+
+        },
```
christophehenry, line 698: I decided to create a compy with the new behavior so that it is more explicit that the implementation changes.

christophehenry, line 698: I decided to create a copy with the new behavior so that it is more explicit that the implementation changes.

res/controllers/common-controller-scripts.js
```diff
@@ -190,6 +192,7 @@ const deepMerge = function(target, source) {
     }
 };
 
+/** @deprecated Use {@link Object.assign} instead */
 script.deepMerge = deepMerge;
```
christophehenry, line 196: Damn this file need a big refactoring…

Swiftb0y: can you elaborate?

res/controllers/midi-components-0.0.js
```diff
@@ -651,7 +651,13 @@
             // Unset isShifted for each ComponentContainer recursively
             this.isShifted = false;
         },
+        /**
+         * @param newLayer Layer to apply to this
+         * @param reconnectComponents Whether components should be reconnected or not
+         * @deprecated Use @{Component#setLayer} instead
+         */
         applyLayer: function(newLayer, reconnectComponents) {
+            console.warn("Components.applyLayer is deprecated; use Component.setLayer instead");
```
Swiftb0y, line 660: `applyLayer` is used extensively. Do you think its feasible to transition some code to `setLayer` in certain mappings?

res/controllers/common-controller-scripts.js
christophehenry: Everything is declared using `var` sothat variables endup in the global scope. `script` should be declared as such:

```js
(function(global) {
    const script = Object.freeze({
        LIBRARY_COLUMNS: Object.freeze({ … }),
        debug() { … }
    });
})(this)
```

That would improve readability IMHO.

Maybe even [`globalThis`](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/globalThis) can be used? I need to check if implemented in `QJSEngine`.

christophehenry: Also `print()` is stillused everywhere (could be remove in 2.60 too)./

res/controllers/midi-components-0.0.js
christophehenry: In this PR? Sure! Every mapping or just one or two?

Swiftb0y: not sure. I'd say every one where you can be sure that the change to `Object.assign` does not cause regressions. Ideally any hardware you have on hand to test/verify that it didn't break anything.

res/controllers/common-controller-scripts.js
Swiftb0y: I assume you mean `global.script = ...`? But yeah, employing that pattern would probably be better, even though I don't find it to improve readability particularly.

res/controllers/midi-components-0.0.js
christophehenry: Uh. I only have 2 Denon DN-S3700 here, sorry.

christophehenry: Do you have an idea aboutwhat are the two most used controllers? I propose to start by them and them ask the community to test.

Swiftb0y: not sure. I'm guessing many pioneer controllers. Apart from that I could test on my NS6II and there are several people in the mixxx team that have a traktor S4MK3 (though that uses its own ComponentsJS implementation IIRC).

Swiftb0y: We could just switch it for all mappings before a beta and see if we receive complaints during the beta phase... Thats what betas are for IMO.

christophehenry: So we switch the mappings in a seperate PR ?

res/controllers/lodash.mixxx.js
```diff
@@ -8,6 +8,12 @@
  * Copyright Jeremy Ashkenas, DocumentCloud and Investigative Reporters & Editors
  */
 ;(function() {
+  const DEPRECATION_MSG = (
+      "Lodash is deprecated and will be removed in Mixxx 2.6.0; refer to the page " +
```
JoergAtGithub, line 12:
```suggestion
      "Lodash is deprecated and will be removed in a future release of Mixxx; refer to the page " +
```
I'm in favor of deprecating Lodash and printing a warning, but we should not remove it to early. Third party mappings might break due to this change.

JoergAtGithub (dismissed):

res/controllers/lodash.mixxx.js
christophehenry: Done. Thank you.

Swiftb0y (approved):
LGTM. wdyt @JoergAtGithub?

res/controllers/midi-components-0.0.js
```diff
@@ -651,7 +651,13 @@
             // Unset isShifted for each ComponentContainer recursively
             this.isShifted = false;
         },
+        /**
+         * @param newLayer Layer to apply to this
+         * @param reconnectComponents Whether components should be reconnected or not
+         * @deprecated since 2.5.0. Use @{Component#setLayer} instead
+         */
         applyLayer: function(newLayer, reconnectComponents) {
+            console.warn("Components.applyLayer is deprecated; use Component.setLayer instead");
```
christophehenry, line 660:
```suggestion
            console.warn("Component.applyLayer is deprecated; use Component.setLayer instead");
```

christophehenry:
@Swiftb0y @JoergAtGithub Uh wait. There's one more typo.

res/controllers/midi-components-0.0.js
Swiftb0y: Right?
```suggestion
            console.warn("ComponentContainer.applyLayer is deprecated; use ComponentContainer.setLayer instead");
```

christophehenry: Done.

Swiftb0y (approved):

res/controllers/common-controller-scripts.js
JoergAtGithub: > Damn this file need a big refactoring…

Are you aware of #11625? Help to fix the build failure there is appreciated.

res/controllers/common-controller-scripts.js
```diff
@@ -190,6 +192,7 @@ const deepMerge = function(target, source) {
     }
 };
 
+/** @deprecated Use {@link Object.assign} instead */
```
JoergAtGithub, line 195: This `@deprecated` is not recognized. I think it only works before the function definition.

christophehenry: Isn't it? What IDE are you on? This is correctly interpreted on CLion.

JoergAtGithub:
It neither work with VisualStudio nor VS Code

res/controllers/common-controller-scripts.js
JoergAtGithub: It neither work with VisualStudio nor VS Code

JoergAtGithub: Than it warns in line `script.deepMerge = deepMerge;` that deepMerge is deprecated.

christophehenry: My last commit should fix this. Can you test (I don't know how to tell to VSCode "there are global definitions in these files, plz source")?

res/controllers/common-controller-scripts.js
```diff
@@ -158,14 +158,15 @@ const isSimpleObject = function(obj) {
     return obj !== null && typeof obj === "object" && obj.constructor.name === "Object";
 };
 
-script.isSimpleObject = isSimpleObject;
```
christophehenry, line 161: I don't think this should be exported.

christophehenry:
I think it's ready for a merge, isn't it?

JoergAtGithub:
I adapted the changes in `res/controllers/common-controller-scripts.js` in #11625 in a way that works with IDEs too. This was the only issue I had.

Thank you!

christophehenry:
Oh that was fast! Thank you!

Issue #10890, opened 2022-09-12
Wrong Key format written to the file Tags 


daschuer:
### Bug Description

According to https://id3.org/id3v2.3.0 the content of TKEY is well defined: 

>TKEY
>    The 'Initial key' frame contains the musical key in which the sound starts. It is represented as a string with a maximum length of three characters. The ground keys are represented with "A","B","C","D","E", "F" and "G" and halfkeys represented with "b" and "#". Minor is represented as "m". Example "Cbm". Off key is represented with an "o" only. 

Mixxx currently writes the String visible at the GUI into the TKEY field. This is wrong and can cause issues in third party applications. 

### Version

2.3.3 and main 

### OS

All

daschuer:
Since this bug fix may break user workflows with external programs that are showing the key info literally, I will move it to the 2.4-beta

acolombier:
@daschuer is this completely fixed now? Close?

daschuer:
Yes

PR #11301, opened 2023-02-24
Key refactoring

daschuer:
This is some refactoring taken form https://github.com/mixxxdj/mixxx/pull/11096 rebased from 2.3. to 2.4
I have decided to do this in an extra PR, to have finally a clear view on the functional changes in the follow up PRs

PR #11098, opened 2022-11-28
ID3v2 key notation export 

daschuer:
This makes sure that newly edited keys are written with the ID3v2 format to the database and track metadata. 
Before the current selected key notation from the preferences is used. 
This is likely incompatible to other tools. 

Fixes #10890

This PR is on top of https://github.com/mixxxdj/mixxx/pull/11096

PR #11096, opened 2022-11-27
Keep the metadata key text unchanged, use it as the origin of information 

daschuer:
This  fixes unnecessary writing of file metadata #11095

The issue is demonstrated in a unit-test. 

This PR is required to not loose extra metadata written by Rapid Evolution #11001 #10992

PR #13462, opened 2024-07-10
Fix "Wrong Key format written to the file Tags"

Swiftb0y:
This is my yet untested attempt at fixing #10890

The reading of the tag is there, I have yet do discover all the places where the `TrackInfo` needs the key to be properly set. I have also not yet decided to how to handle the case where the user sets some custom key (guess the appropriate `ChromaticKey` and handle appropriately when we can't guess it). Also yet untested. 

Adoption is welcome.

### DIFF
--- a/src/track/keyutils.cpp
+++ b/src/track/keyutils.cpp
@@ -38,31 +38,66 @@ const QString s_sharpSymbol = QString::fromUtf8("♯");
 //static const QString s_flatSymbol = QString::fromUtf8("♭");
 
 const QString s_traditionalKeyNames[] = {
-        QString::fromUtf8("INVALID"),
-        QString::fromUtf8("C"),
-        QString::fromUtf8("D♭"),
-        QString::fromUtf8("D"),
-        QString::fromUtf8("E♭"),
-        QString::fromUtf8("E"),
-        QString::fromUtf8("F"),
-        QString::fromUtf8("F♯/G♭"),
-        QString::fromUtf8("G"),
-        QString::fromUtf8("A♭"),
-        QString::fromUtf8("A"),
-        QString::fromUtf8("B♭"),
-        QString::fromUtf8("B"),
-        QString::fromUtf8("Cm"),
-        QString::fromUtf8("C♯m"),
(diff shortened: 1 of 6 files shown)


daschuer:
My work has stalled on the discussion in https://github.com/mixxxdj/mixxx/pull/11098 and 
https://github.com/mixxxdj/mixxx/pull/11301
around broken user workflow. 

I just did a brief look, but I have not found the comment. 
It mentions the need for a preferences option. Can you recal 

I think the underlying issue was that we need to make sure we do not alter extended user tags written from third party tools. 

How is the situation with this PR?

Swiftb0y:
> It mentions the need for a preferences option. Can you recal

From what I understand summarizing the discussion, the user wants to keep the buggy behavior and have tool to rewrite from the buggy format to the standard one. That is nice to have, but feels like overcomplication to me. I wouldn't implement it if it requires more than 30 SLOC. 

> I think the underlying issue was that we need to make sure we do not alter extended user tags written from third party tools.

That is reasonable. And probably not too hard.

> How is the situation with this PR?

Neither issues have not been addressed here, but I have an idea on how to do so.

I would introduce a `key` property (similar to how I've done in this PR) that carries the `ChromaticKey` value. On import, we try guessing it from the `keyText` and store the raw key in the `keyText`. If we can't guess it, set the `ChromaticKey` to `INVALID`. On export, if the `ChromaticKey` is invalid, fallback to the `keyText`. if the `ChromaticKey` is valid, check the preferences on whether to re-export the `keyText` verbatim (legacy) or overwrite it with the `standard` notation.

Since we can't really determine whether the key exported is our non-standard text or something custom from the user carrying extra information, we can't really do much more than overwrite the key with standard notation or leave it alone.

Anyways, do you have interested in reviving #11096 so we can get this back on track?

daschuer:
Yes, I will have y look.

daschuer:
This is now integrated in https://github.com/mixxxdj/mixxx/pull/11096

Swiftb0y:
superseeded by #11096

Issue #11814, opened 2023-08-12
Pitch value on startup


gui-lux:
### Bug Description

Hi,
I noticed v2.4 sometimes applies a wrong value on pitch sliders at startup.
This is rather minor because one pitch slider touch totally wipes the trouble and this could be related to my controller as previously discussed. On the other hand, v2.3.x didn't behave like this.

So, when starting v2.4 with my MC4000 from Denon with pitch sliders set on a value that is not zero, not min value, not max value, then mixxx displays a pitch value that looks multiplied by ten and more, and giving strange zoomed-in (in case of positive hardware pitch value) waveforms.
I can reproduce across shutdowns and power cut only, can't reproduce when only restarting mixxx, and not reproducing 100% with a simple reboot.

The following line from an lsusb output that relates my controller :
`Bus 001 Device 005: ID 15e4:8001 Numark MC4000`

From mixxx.log :

```
Debug [Controller] ControllerManager:slotInitialize
Debug [Controller] Extension .midi.xml total 3 mappings
Debug [Controller] Extension .hid.xml total 0 mappings
Debug [Controller] Extension .bulk.xml total 0 mappings
Debug [Main] Set root GL Context widget valid: WInitialGLWidget(0x558a9f3af0) false
Debug [Main] Displaying main window
Debug [Main] Running Mixxx
Debug [Main] OpenGL driver version string "2.1 Mesa 20.3.5", vendor "Broadcom", renderer "V3D 4.2"
Debug [Main] Supported OpenGL version: 2.1
Debug [Main] RLimit Cur  95
Debug [Main] RLimit Max  95
Debug [Main] ControllerManager::getControllerList
Info [Main] Loaded skin "Deere"
Warning [Main] Unsupported locale "cmn"
Debug [Controller] Extension .midi.xml total 129 mappings
Debug [Controller] Extension .hid.xml total 15 mappings
Debug [Controller] Extension .bulk.xml total 1 mappings
Debug [Controller] ControllerManager: Setting up devices
Debug [Controller] Scanning PortMIDI devices:
Debug [Controller]  Found output device # 2 MC4000 MIDI 1
Debug [Controller]  Found input device # 3 MC4000 MIDI 1
Debug [Controller]     Linking to output device # 2 "MC4000 MIDI 1"
Debug [Controller] Scanning USB Bulk devices:
Info [Controller] Scanning USB HID devices
[...]

PR #4495, opened 2021-10-30
midi-components-0.0.js: Pot support for arbitrary maximums in 7-/14-bit handlers

Swiftb0y:
previously, it was not possible for 14-bit controls to have maxima
that are not (2^14)-1. Some controller might only have 3-bits of
resolution in their MSB. This commit allows for arbitrary precision
in MSB and LSB by specifying an appropriate value as the components
max property.

PR #13463, opened 2024-07-11
fix: `components.Pot` producing wrong result on first value

### LINKED ISSUE #12569: Initial tempo after connecting Roland DJ-505 is set to very high BPM
### Bug Description

This happens on both decks after I select the midi mapping for the already connected Roland DJ-505
A track with 138 BPM is then set to about 1264 BPM and thus cannot be played back. After connecting the controller, I always have to move the tempo fader a bit back and forth in the center position and then the BPM are set correctly once the change is picked up

### Version

2.3.6, 2.5 built today

### OS

Debian 12, macOS 14 (both with all the latest updates)

Swiftb0y:
This fixes a regression introduced in https://github.com/mixxxdj/mixxx/pull/4495 and reported as https://github.com/mixxxdj/mixxx/issues/11814

### DIFF
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -481,8 +481,7 @@
                 if (this.max === Component.prototype.max) {
                     this.max = (1 << 14) - 1;
                 }
-                value = (value << 7) + (this._firstLSB ? this._firstLSB : 0);
-                this.input(channel, control, value, status, group);
+                this.input(channel, control, (value << 7) + (this._firstLSB ? this._firstLSB : 0), status, group);
             }
             this.MSB = value;
         },


res/controllers/midi-components-0.0.js
```diff
@@ -481,8 +481,7 @@
                 if (this.max === Component.prototype.max) {
                     this.max = (1 << 14) - 1;
                 }
-                value = (value << 7) + (this._firstLSB ? this._firstLSB : 0);
-                this.input(channel, control, value, status, group);
+                this.input(channel, control, (value << 7) + (this._firstLSB ? this._firstLSB : 0), status, group);
             }
             this.MSB = value;
```
Swiftb0y, line 486: A quick word on the fix:
This only occurred when receiving the first value because this branch is only executed when receiving the first value. `this.MSB` is only supposed to contain the value of the most significant byte, but the branch erroneously set it contain the entire value. That meant if the MSB was received first everything would work out for the first message, however, when the LSB was received afterwards it would take the bogus MSB value set here resulting in huge bogus values. Once the second MSB was received, everything would work out as usual again. The fix now ensures that `this.MSB` gets the actual MSB value instead of the "pre-scaled" one that we use in the branch.

m0dB: LGTM

PR #13468, opened 2024-07-13
20240713 NL Dutch gui translations updated (+/- 300 items)

Eve00000:


daschuer (approved):
Thank you. I will take care to upload it to Transifex.

daschuer:
Transifex is complaining about these terms: 
* Effect Eenheid vs. Effecteenheid
* gain vs versterking
The rest I was able to fix.

Eve00000:
> Effect Eenheid and Effecteenheid : are both correct dutch
> gain vs versterking: gain is the term soundpeople (dj's, sound engineers, Pa people...  and audio equipement use) versterking is in dutch more used for an amplifier, an amplifier versterkt. (amplifier in dutch = versterker)

daschuer:
> Effect Eenheid and Effecteenheid : are both correct dutch

If we can use both interchangable, we should default to one and adjust the Glossary accordingly. Which one do you prefer?

daschuer:
Can you double check the translation of gain? 

I think we have an issue if the source code uses one term for two different things, but the target language has two terms. In case it are two different things, we need to take care that we pick the right translation term when when have texts referencing each other.

I see in other cases that you have used both of these terms with the English version in (). That is probably not the best idea, because of string length constraints. 

If there are any, translator comments in the source would be very helpful.

Eve00000:
Dutch / Nederlands is a strange language, even more because the Netherlands are more English orientated and Flanders, the Dutch speaking part of Belgium is more French orientated. So the ingluences in the language come from the same orientation.
Other than French and German in Dutch foreign words are just copied.
fi Computer:, the french invented a word for it: Ordinateur, in Dutch we did not invent a word for it, just copied Computer. (The South-Africans invented the word rekenaar, kind of calculator).
Dutch has not such a rich vocabulary as French and Italian, a lot of French / Italian / Englisg words have the same translation in Dutch, so people use the English words to be more particular in their meaning.

When you just translate unit in Dutch it would be eenheid.
But the Dutch word eenheid can mean: unity of measurement / unity (eendracht maakt macht = l"union fait la force national slofan) / a parts that form a economical unit (a family, company)  / living room
Effect can be: the result of something / spin / value paper (share in a company) in fact 
Matbe we should prefer "effect eenheid' analogous to the English.

Gain in dutch is verdienen (gain money)  / krijgen (receive) / winnen (score) / verkrijgen (obtain) / verwerven (acquire) / bereiken (reach) / behalen (achieve) / toenemen (increase) / aanwinnen (gain) / aanwinst (asset) / voordeel (advantage) / profijt (profit) / buit (booty) / prooi (prey) / acquest (accession)
So no meaningful meaning for the term used in audio except 'toenemen (increase).
In my opinion gain is an international used term, it's not alike volume for which we use 'versterking (fortification :-) )

Volume is the loudness of the sound, changing volume changes the loudness, not the tone. If you put the amplifier on 100 the soundwave stays the same, only the playback volume is changed.
Gain (pre-fader adjustment) changes the waveform (volume and distortion) so this ain't 'versterking', in Dutch it should be something like 'adjusting the finetuning of the input signal to achieve the best possible quality of input level"

Let's keep it "gain " ;-)

translator comments : the <extra comments> ?

PR #13339, opened 2024-06-07
Skins: add 'beats undo' button

ronso0:
Besides the buttons, this also adds a new read-only control `beats_undo_possible` which is updated if beats change or BPM lock is toggled and allows to deactivate (cover) the Undo button.
(this restores the big CurPos button in Tango & LateNight)

image
image
image
image

PR #13469, opened 2024-07-13
undo/redo cue add/remove, wip

m0dB:
Using a QUndoStack to undo/redo adding and removing (hot)cues.

The undo's/redo's are triggered from the menu bar, edit menu and can also be triggered using the standard shortkeys.

Tested using UI interaction. I am not sure if using a controller goes through the same code path.

The caveat is that when the track is changed, the commands would become invalid. I now use a stop-gap solution and clean the entire stack in the Track destructor. There is unfortunately no way to purge the stack only of the commands that have become invalid.

Maybe we should include loading / unloading tracks in the undo stack? That could solve the issue.

I do feel that using QUndoStack is the correct solution; the amount of code to implement this is minimal. 

I see that Track uses a custom mechanism, m_pBeatsUndoStack. I think it would make sense to also use the QUndoStack  for those changes.

### DIFF
--- a/src/coreservices.cpp
+++ b/src/coreservices.cpp
@@ -55,6 +55,7 @@
 #include "util/statsmanager.h"
 #include "util/time.h"
 #include "util/translations.h"
+#include "util/undostack.h"
 #include "util/versionstore.h"
 #include "vinylcontrol/vinylcontrolmanager.h"
 
@@ -143,6 +144,8 @@ CoreServices::CoreServices(const CmdlineArgs& args, QApplication* pApp)
     mixxx::Translations::initializeTranslations(
             m_pSettingsManager->settings(), pApp, m_cmdlineArgs.getLocale());
     initializeKeyboard();
+
+    UndoStack::createInstance();
 }
 
 CoreServices::~CoreServices() {
@@ -158,6 +161,8 @@ CoreServices::~CoreServices() {
     if (m_cmdlineArgs.getDeveloper()) {
         StatsManager::destroy();
     }
+
+    UndoStack::destroy();
 
     // HACK: Save config again. We saved it once before doing some dangerous
     // stuff. We only really want to save it here, but the first one was just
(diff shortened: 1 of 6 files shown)


ronso0:
Thank you for this next track management improvement!
Some thoughts:

**Do we really need Redo?**
IIUC the most common use case for Undo is having accidentally deleted a cue point.
Undo add can be done much easier IMO by simply deleting a cue via the cue menu.

**Undo/Redo in the menu bar / Ctrl+Z/Y hotkeys**
While putting Undo/Redo into the Edit menu certainly makes sense in terms of discoverabilty and consistency with many other file editing apps, I'm actually glad that currently most track/library opeartions can be done from within GUI (without the menu bar, a button for the preferences would complete this IMHO)

**If/how to integrate the BPM undo into the new Undo stack**
What I like a lot about the current Undo/lock controls is that they are avialbale directly in the skin where they are needed (beatgrid control section). And 'can undo' is viisble at a glance, too, coupled to the beats ando stack (#13339 -> 367e3ba973304833055bef15deeb2a0c57bf302e). I'm not sure how that distinct undo property can be realized with QUndoStack which also deals with cues (and maybe more track props in the future).

ronso0:
Some observations after a brief test:
* works as expected :tada: 
* Undo applies to **all** cue operations, i.e. also to Intro/Outro cues set automatically on (re)analysis by the silence analyser.

m0dB:
Hi @ronso0 , thanks for your feedback. 

I think the main question is: do we want to a single application wide Undo stack, or do we want track or even feature specific undo stacks. You are advocating for the second and your arguments are very valid. On the other hand, a single application wide Undo(/Redo) stack is the common paradigm. But that doesn't mean it's the right approach for Mixxx.

I think this requires a broader discussion. I will take this to zulip.

m0dB:
> Undo applies to all cue operations, i.e. also to Intro/Outro cues set automatically on (re)analysis by the silence analyser.

Well spotted. Automatically set cues should not become part of the Undo/Redo stack. I will see if this can be avoided.

PR #13470, opened 2024-07-14
Scrolling waveform using SceneGraph in QML

acolombier:
This PR is a rebase of @m0dB's `qml_scrolling_waveform` on the latest main

The waveform rendering approach should provided better result in the long run and should integrate with the new `allshader` approach.

Currently, the waveform rendering logic which happen in `QmlWaveformDisplay::updatePaintNode` is duplicated (or rather strongly inspired) from the RGB allshader rendering. A "definition of ready" for this PR would be to have some kind of compatibility layer/integration to reuse the rendering logic define in `allshader` and should support at least one waveform type logic (e.g `RGB`)

Depends on:
- [x] #14185 
- [x] #14186 
- [x] #14187 
- [x] #14188 
- [x] #14189 
- [x] #14190 
- [x] #14191 
- [x] #14192 

TODO:
 - [x] Crash on shutdown
 - [x] Waveform zoom init
 - [X] Something is wrong with the clipping
 - [x] Something is wrong with the stem waveform

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4750,6 +4750,7 @@ if(QML)
   )
   target_link_libraries(mixxx-qml-lib PRIVATE rendergraph_sg)
   target_compile_definitions(mixxx-qml-lib PRIVATE rendergraph=rendergraph_sg)
+  target_compile_definitions(mixxx-qml-lib PRIVATE __SCENEGRAPH__)
   target_compile_definitions(mixxx-qml-lib PRIVATE allshader=allshader_sg)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3372,6 +3372,7 @@ if(QML)
       res/qml/Mixxx/Controls/WaveformOverviewHotcueMarker.qml
       res/qml/Mixxx/Controls/WaveformOverviewMarker.qml
       res/qml/Mixxx/Controls/WaveformOverview.qml
+      res/qml/Mixxx/Controls/WaveformDisplay.qml
   )
   target_link_libraries(mixxx-qml-lib PRIVATE mixxx-qml-mixxxcontrolsplugin)
 
@@ -3397,6 +3398,21 @@ if(QML)
       src/qml/qmlchainpresetmodel.cpp
       src/qml/qmlwaveformoverview.cpp
       src/qml/qmlmixxxcontrollerscreen.cpp
+      src/qml/qmlwaveformdisplay.cpp
+      src/qml/qmlwaveformrenderer.cpp
+      src/waveform/renderers/allshader/digitsrenderer.cpp
+      src/waveform/renderers/allshader/waveformrenderbeat.cpp
+      src/waveform/renderers/allshader/waveformrenderer.cpp
+      src/waveform/renderers/allshader/waveformrendererendoftrack.cpp
+      src/waveform/renderers/allshader/waveformrendererpreroll.cpp
+      src/waveform/renderers/allshader/waveformrendererrgb.cpp
+      src/waveform/renderers/allshader/waveformrenderersignalbase.cpp
+      src/waveform/renderers/allshader/waveformrendermark.cpp
+      src/waveform/renderers/allshader/waveformrendermarkrange.cpp
+      src/waveform/renderers/allshader/waveformrendererslipmode.cpp
+      src/waveform/renderers/allshader/waveformrendererfiltered.cpp
+      src/waveform/renderers/allshader/waveformrendererhsv.cpp
+      src/waveform/renderers/allshader/waveformrenderersimple.cpp
       # The following sources need to be in this target to get QML_ELEMENT properly interpreted
       src/control/controlmodel.cpp
       src/control/controlsortfiltermodel.cpp
@@ -4359,7 +4375,12 @@ if(STEM)
   endif()
   if(QML)
     target_compile_definitions(mixxx-qml-lib PUBLIC __STEM__)
-    target_sources(mixxx-qml-lib PRIVATE src/qml/qmlstemsmodel.cpp)
+    target_sources(
+      mixxx-qml-lib
+      PRIVATE
+        src/waveform/renderers/allshader/waveformrendererstem.cpp
+        src/qml/qmlstemsmodel.cpp
+    )
   endif()
 endif()
 
@@ -4714,8 +4735,23 @@ endif()
 
 # rendergraph
 add_subdirectory(src/rendergraph)
+target_compile_definitions(
+  rendergraph_gl
+  PUBLIC $<$<CONFIG:Debug>:MIXXX_DEBUG_ASSERTIONS_ENABLED>
+)
 target_link_libraries(mixxx-lib PUBLIC rendergraph_gl)
 target_compile_definitions(mixxx-lib PUBLIC rendergraph=rendergraph_gl)
+target_compile_definitions(mixxx-lib PRIVATE allshader=allshader_gl)
+target_compile_definitions(mixxx-test PRIVATE allshader=allshader_gl)
+if(QML)
+  target_compile_definitions(
+    rendergraph_sg
+    PUBLIC $<$<CONFIG:Debug>:MIXXX_DEBUG_ASSERTIONS_ENABLED>
+  )
+  target_link_libraries(mixxx-qml-lib PRIVATE rendergraph_sg)
+  target_compile_definitions(mixxx-qml-lib PRIVATE rendergraph=rendergraph_sg)
+  target_compile_definitions(mixxx-qml-lib PRIVATE allshader=allshader_sg)
+endif()
 
 # WavPack audio file support
 find_package(wavpack)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3613,15 +3613,6 @@ else()
 
       install(
         IMPORTED_RUNTIME_ARTIFACTS
-          Qt${QT_VERSION_MAJOR}::QuickControls2
-          DESTINATION
-          "${MIXXX_INSTALL_DATADIR}"
-          COMPONENT
-          applocal
-      )
-
-      install(
-        IMPORTED_RUNTIME_ARTIFACTS
           Qt${QT_VERSION_MAJOR}::QuickControls2Impl
           DESTINATION
           "${MIXXX_INSTALL_DATADIR}"
@@ -3641,15 +3632,6 @@ else()
       install(
         IMPORTED_RUNTIME_ARTIFACTS
           Qt${QT_VERSION_MAJOR}::QuickShapesPrivate
-          DESTINATION
-          "${MIXXX_INSTALL_DATADIR}"
-          COMPONENT
-          applocal
-      )
-
-      install(
-        IMPORTED_RUNTIME_ARTIFACTS
-          Qt${QT_VERSION_MAJOR}::QuickTemplates2
           DESTINATION
           "${MIXXX_INSTALL_DATADIR}"
           COMPONENT
--- a/res/controllers/DummyDeviceDefaultScreen.qml
+++ b/res/controllers/DummyDeviceDefaultScreen.qml
@@ -170,7 +170,7 @@ Mixxx.ControllerScreen {
                         Layout.fillHeight: true
                         Text {
                             text: qsTr("Group")
-                            font.pixelSize: 24
+                            font.pixelSize: 18
                             font.family: "Noto Sans"
                             font.letterSpacing: -1
                             color: fontColor
@@ -183,7 +183,7 @@ Mixxx.ControllerScreen {
                         Layout.fillHeight: true
                         Text {
                             text: `${root.group}`
-                            font.pixelSize: 24
+                            font.pixelSize: 18
                             font.family: "Noto Sans"
                             font.letterSpacing: -1
                             color: fontColor
@@ -207,7 +207,7 @@ Mixxx.ControllerScreen {
                         Layout.fillHeight: true
                         Text {
                             text: qsTr("Widget")
-                            font.pixelSize: 24
+                            font.pixelSize: 18
                             font.family: "Noto Sans"
                             font.letterSpacing: -1
                             color: fontColor
@@ -258,7 +258,7 @@ Mixxx.ControllerScreen {
                         anchors.bottomMargin: 6
                         Layout.fillWidth: true
                         Layout.fillHeight: true
-                        spacing: 6
+                        spacing: 4
                         required property var modelData
 
                         Mixxx.ControlProxy {
@@ -273,7 +273,7 @@ Mixxx.ControllerScreen {
                             Layout.fillHeight: true
                             Text {
                                 text: qsTr(modelData.title)
-                                font.pixelSize: 24
+                                font.pixelSize: 18
                                 font.family: "Noto Sans"
                                 font.letterSpacing: -1
                                 color: fontColor
@@ -286,10 +286,161 @@ Mixxx.ControllerScreen {
                             Layout.fillHeight: true
                             Text {
                                 text: `${mixxxValue.value}`
-                                font.pixelSize: 24
+                                font.pixelSize: 18
                                 font.family: "Noto Sans"
                                 font.letterSpacing: -1
                                 color: fontColor
+                            }
+                        }
+                    }
+                }
+                RowLayout {
+                    anchors.leftMargin: 6
+                    anchors.rightMargin: 6
+                    anchors.topMargin: 6
+                    anchors.bottomMargin: 6
+
+                    Layout.fillWidth: true
+                    Layout.fillHeight: true
+
+                    Mixxx.ControlProxy {
+                        id: zoomControl
+
+                        group: root.group
+                        key: "waveform_zoom"
+                    }
+
+                    MixxxControls.WaveformDisplay {
+                        group: root.group
+                        x: 0
+                        width: root.width
+                        height: 100
+
+                        zoom: zoomControl.value
+                        backgroundColor: "#36000000"
+
+                        Mixxx.WaveformRendererEndOfTrack {
+                            color: 'blue'
+                            endOfTrackWarningTime: 30
+                        }
+
+                        Mixxx.WaveformRendererPreroll {
+                            color: '#998977'
+                        }
+
+                        Mixxx.WaveformRendererMarkRange {
+                            // Loop
+                            Mixxx.WaveformMarkRange {
+                                startControl: "loop_start_position"
+                                endControl: "loop_end_position"
+                                enabledControl: "loop_enabled"
+                                color: '#00b400'
+                                opacity: 0.7
+                                disabledColor: '#FFFFFF'
+                                disabledOpacity: 0.6
+                            }
+                            // Intro
+                            Mixxx.WaveformMarkRange {
+                                startControl: "intro_start_position"
+                                endControl: "intro_end_position"
+                                color: '#2c5c9a'
+                                opacity: 0.6
+                                durationTextColor: '#ffffff'
+                                durationTextLocation: 'after'
+                            }
+                            // Outro
+                            Mixxx.WaveformMarkRange {
+                                startControl: "outro_start_position"
+                                endControl: "outro_end_position"
+                                color: '#2c5c9a'
+                                opacity: 0.6
+                                durationTextColor: '#ffffff'
+                                durationTextLocation: 'before'
+                            }
+                        }
+
+                        Mixxx.WaveformRendererRGB {
+                            axesColor: '#00ffffff'
+                            lowColor: 'red'
+                            midColor: 'green'
+                            highColor: 'blue'
+
+                            gainAll: 1.0
+                            gainLow: 1.0
+                            gainMid: 1.0
+                            gainHigh: 1.0
+                        }
+
+                        Mixxx.WaveformRendererStem {
+                            gainAll: 1.0
+                        }
+
+                        Mixxx.WaveformRendererBeat {
+                            color: '#cfcfcf'
+                        }
+
+                        Mixxx.WaveformRendererMark {
+                            playMarkerColor: 'cyan'
+                            playMarkerBackground: 'transparent'
+                            defaultMark: Mixxx.WaveformMark {
+                                align: "bottom|center"
+                                color: "#FF0000"
+                                textColor: "#FFFFFF"
+                                text: " %1 "
+                            }
+
+                            untilMark.showTime: false
+                            untilMark.showBeats: false
+                            untilMark.align: Qt.AlignCenter
+                            untilMark.textSize: 14
+
+                            Mixxx.WaveformMark {
+                                control: "cue_point"
+                                text: 'CUE'
+                                align: 'top|right'
+                                color: '#FF8000'
+                                textColor: '#FFFFFF'
+                            }
+                            Mixxx.WaveformMark {
+                                control: "loop_start_position"
+                                text: '↻'
+                                align: 'top|left'
+                                color: 'green'
+                                textColor: '#FFFFFF'
+                            }
+                            Mixxx.WaveformMark {
+                                control: "loop_end_position"
+                                align: 'bottom|right'
+                                color: 'green'
+                                textColor: '#FFFFFF'
+                            }
+                            Mixxx.WaveformMark {
+                                control: "intro_start_position"
+                                text: '◢'
+                                align: 'top|right'
+                                color: 'blue'
+                                textColor: '#FFFFFF'
+                            }
+                            Mixxx.WaveformMark {
+                                control: "intro_end_position"
+                                text: '◢'
+                                align: 'top|left'
+                                color: 'blue'
+                                textColor: '#FFFFFF'
+                            }
+                            Mixxx.WaveformMark {
+                                control: "outro_start_position"
+                                text: '◣'
+                                align: 'top|right'
+                                color: 'blue'
+                                textColor: '#FFFFFF'
+                            }
+                            Mixxx.WaveformMark {
+                                control: "outro_end_position"
+                                text: '◣'
+                                align: 'top|left'
+                                color: 'blue'
+                                textColor: '#FFFFFF'
                             }
                         }
                     }
--- a/res/qml/Deck.qml
+++ b/res/qml/Deck.qml
@@ -327,6 +327,54 @@ Item {
         }
 
         Row {
+            anchors.left: playButton.right
+            anchors.leftMargin: 10
+            anchors.bottom: playButton.bottom
+            anchors.topMargin: 5
+            spacing: -1
+
+            Skin.IntroOutroButton {
+                keyPrefix: "intro_start"
+                group: root.group
+
+                text: "Intro\nStart"
+
+                width: playButton.height * 2 - 1
+                height: playButton.height
+            }
+
+            Skin.IntroOutroButton {
+                keyPrefix: "intro_end"
+                group: root.group
+
+                text: "Intro\nEnd"
+
+                width: playButton.height * 2 - 1
+                height: playButton.height
+            }
+
+            Skin.IntroOutroButton {
+                keyPrefix: "outro_start"
+                group: root.group
+
+                text: "Outro\nStart"
+
+                width: playButton.height * 2 - 1
+                height: playButton.height
+            }
+
+            Skin.IntroOutroButton {
+                keyPrefix: "outro_end"
+                group: root.group
+
+                text: "Outro\nEnd"
+
+                width: playButton.height * 2 - 1
+                height: playButton.height
+            }
+        }
+
+        Row {
             anchors.left: cueButton.right
             anchors.top: parent.top
             anchors.leftMargin: 10
--- a/res/qml/WaveformDisplay.qml
+++ b/res/qml/WaveformDisplay.qml
@@ -8,6 +8,7 @@ Item {
     id: root
 
     required property string group
+    property bool splitStemTracks: false
 
     enum MouseStatus {
         Normal,
@@ -74,7 +75,8 @@ Item {
         }
 
         Mixxx.WaveformRendererStem {
-            gainAll: 1.0
+            gainAll: root.splitStemTracks ? 2.0 : 1.0
+            splitStemTracks: root.splitStemTracks
         }
 
         Mixxx.WaveformRendererBeat {
@@ -188,6 +190,11 @@ Item {
 
         anchors.fill: parent
         acceptedButtons: Qt.LeftButton | Qt.RightButton
+        onDoubleClicked: {
+            if (mouse.button == Qt.RightButton) {
+                root.splitStemTracks = !root.splitStemTracks;
+            }
+        }
         onPressed: {
             mouseAnchor = Qt.point(mouse.x, mouse.y);
             if (mouse.button == Qt.LeftButton) {
--- a/res/qml/WaveformDisplay.qml
+++ b/res/qml/WaveformDisplay.qml
@@ -1,0 +1,247 @@
+import "." as Skin
+import Mixxx 1.0 as Mixxx
+import Mixxx.Controls 1.0 as MixxxControls
+import QtQuick 2.12
+import "Theme"
+
+Item {
+    id: root
+
+    required property string group
+
+    enum MouseStatus {
+        Normal,
+        Bending,
+        Scratching
+    }
+
+    MixxxControls.WaveformDisplay {
+        anchors.fill: parent
+        group: root.group
+        zoom: zoomControl.value
+        backgroundColor: "#5e000000"
+
+        Mixxx.WaveformRendererEndOfTrack {
+            color: '#ff8872'
+            endOfTrackWarningTime: 30
+        }
+
+        Mixxx.WaveformRendererPreroll {
+            color: '#ff8872'
+        }
+
+        Mixxx.WaveformRendererMarkRange {
+            // Loop
+            Mixxx.WaveformMarkRange {
+                startControl: "loop_start_position"
+                endControl: "loop_end_position"
+                enabledControl: "loop_enabled"
+                color: '#00b400'
+                opacity: 0.7
+                disabledColor: '#FFFFFF'
+                disabledOpacity: 0.6
+            }
+            // Intro
+            Mixxx.WaveformMarkRange {
+                startControl: "intro_start_position"
+                endControl: "intro_end_position"
+                color: '#2c5c9a'
+                opacity: 0.6
+                durationTextColor: '#ffffff'
+                durationTextLocation: 'after'
+            }
+            // Outro
+            Mixxx.WaveformMarkRange {
+                startControl: "outro_start_position"
+                endControl: "outro_end_position"
+                color: '#2c5c9a'
+                opacity: 0.6
+                durationTextColor: '#ffffff'
+                durationTextLocation: 'before'
+            }
+        }
+
+        Mixxx.WaveformRendererRGB {
+            axesColor: '#a1a1a1a1'
+            lowColor: '#ff2154d7'
+            midColor: '#cfb26606'
+            highColor: '#e5029c5c'
+
+            gainAll: 1.0
+            gainLow: 1.0
+            gainMid: 1.0
+            gainHigh: 1.0
+        }
+
+        Mixxx.WaveformRendererStem {
+            gainAll: 1.0
+        }
+
+        Mixxx.WaveformRendererBeat {
+            color: '#a1a1a1a1'
+        }
+
+        Mixxx.WaveformRendererMark {
+            playMarkerColor: 'cyan'
+            playMarkerBackground: 'orange'
+            defaultMark: Mixxx.WaveformMark {
+                align: "bottom|right"
+                color: "#00d9ff"
+                textColor: "#1a1a1a"
+                text: " %1 "
+            }
+
+            untilMark.showTime: true
+            untilMark.showBeats: true
+            untilMark.align: Qt.AlignBottom
+            untilMark.textSize: 11
+
+            Mixxx.WaveformMark {
+                control: "cue_point"
+                text: 'CUE'
+                align: 'top|right'
+                color: 'red'
+                textColor: '#1a1a1a'
+            }
+            Mixxx.WaveformMark {
+                control: "loop_start_position"
+                text: '↻'
+                align: 'top|left'
+                color: 'green'
+                textColor: '#FFFFFF'
+            }
+            Mixxx.WaveformMark {
+                control: "loop_end_position"
+                align: 'bottom|right'
+                color: 'green'
+                textColor: '#FFFFFF'
+            }
+            Mixxx.WaveformMark {
+                control: "intro_start_position"
+                text: '◢'
+                align: 'top|right'
+                color: 'blue'
+                textColor: '#FFFFFF'
+            }
+            Mixxx.WaveformMark {
+                control: "intro_end_position"
+                text: '◢'
+                align: 'top|left'
+                color: 'blue'
+                textColor: '#FFFFFF'
+            }
+            Mixxx.WaveformMark {
+                control: "outro_start_position"
+                text: '◣'
+                align: 'top|right'
+                color: 'blue'
+                textColor: '#FFFFFF'
+            }
+            Mixxx.WaveformMark {
+                control: "outro_end_position"
+                text: '◣'
+                align: 'top|left'
+                color: 'blue'
+                textColor: '#FFFFFF'
+            }
+        }
+    }
+
+    Mixxx.ControlProxy {
+        id: scratchPositionEnableControl
+
+        group: root.group
+        key: "scratch_position_enable"
+    }
+
+    Mixxx.ControlProxy {
+        id: scratchPositionControl
+
+        group: root.group
+        key: "scratch_position"
+    }
+
+    Mixxx.ControlProxy {
+        id: wheelControl
+
+        group: root.group
+        key: "wheel"
+    }
+
+    Mixxx.ControlProxy {
+        id: rateRatioControl
+
+        group: root.group
+        key: "rate_ratio"
+    }
+
+    Mixxx.ControlProxy {
+        id: zoomControl
+
+        group: root.group
+        key: "waveform_zoom"
+    }
+
+    MouseArea {
+        property int mouseStatus: WaveformDisplay.MouseStatus.Normal
+        property point mouseAnchor: Qt.point(0, 0)
+
+        anchors.fill: parent
+        acceptedButtons: Qt.LeftButton | Qt.RightButton
+        onPressed: {
+            mouseAnchor = Qt.point(mouse.x, mouse.y);
+            if (mouse.button == Qt.LeftButton) {
+                if (mouseStatus == WaveformDisplay.MouseStatus.Bending)
+                    wheelControl.parameter = 0.5;
+
+                mouseStatus = WaveformDisplay.MouseStatus.Scratching;
+                scratchPositionControl.value = 0;
+                scratchPositionEnableControl.value = 1;
+            } else {
+                if (mouseStatus == WaveformDisplay.MouseStatus.Scratching)
+                    scratchPositionEnableControl.value = 0;
+
+                wheelControl.parameter = 0.5;
+                mouseStatus = WaveformDisplay.MouseStatus.Bending;
+            }
+        }
+        onPositionChanged: {
+            const diff = mouse.x - mouseAnchor.x;
+            switch (mouseStatus) {
+                case WaveformDisplay.MouseStatus.Bending: {
+                    // Start at the middle of [0.0, 1.0], and emit values based on how far
+                    // the mouse has traveled horizontally. Note, for legacy (MIDI) reasons,
+                    // this is tuned to 127.
+                    const v = 0.5 + (diff / root.width);
+                    // clamp to [0.0, 1.0]
+                    wheelControl.parameter = Math.max(Math.min(v, 1), 0);
+                    break;
+                };
+                case WaveformDisplay.MouseStatus.Scratching:
+                // TODO: Calculate position properly
+                    scratchPositionControl.value = -diff * zoomControl.value * 200;
+                    break;
+            }
+        }
+        onReleased: {
+            switch (mouseStatus) {
+                case WaveformDisplay.MouseStatus.Bending:
+                    wheelControl.parameter = 0.5;
+                    break;
+                case WaveformDisplay.MouseStatus.Scratching:
+                    scratchPositionEnableControl.value = 0;
+                    scratchPositionControl.value = 0;
+                    break;
+            }
+            mouseStatus = WaveformDisplay.MouseStatus.Normal;
+        }
+
+        onWheel: {
+            if (wheel.angleDelta.y < 0 && zoomControl.value > 1) {
+                zoomControl.value -= 1;
+            } else if (wheel.angleDelta.y > 0 && zoomControl.value < 10.0) {
+                zoomControl.value += 1;
+            }
+        }
+    }
+}
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -274,7 +274,7 @@ void MixxxMainWindow::initialize() {
 
     WaveformWidgetFactory::createInstance(); // takes a long time
     WaveformWidgetFactory::instance()->setConfig(m_pCoreServices->getSettings());
-    WaveformWidgetFactory::instance()->startVSync(m_pGuiTick, m_pVisualsManager);
+    WaveformWidgetFactory::instance()->startVSync(m_pGuiTick, m_pVisualsManager, false);
 
     connect(this,
             &MixxxMainWindow::skinLoaded,
@@ -514,6 +514,8 @@ MixxxMainWindow::~MixxxMainWindow() {
     delete m_pPrefDlg;
 
     m_pCoreServices->getControlIndicatorTimer()->setLegacyVsyncEnabled(false);
+
+    qDebug() << t.elapsed(false).debugMillisWithUnit() << "deleting ControllerManager";
 
     WaveformWidgetFactory::destroy();
 
--- a/src/qml/qmlwaveformrenderer.cpp
+++ b/src/qml/qmlwaveformrenderer.cpp
@@ -141,6 +141,7 @@ QmlWaveformRendererFactory::Renderer QmlWaveformRendererStem::create(
             waveformWidget, m_position);
 
     pRenderer->setAllChannelVisualGain(m_gainAll);
+    pRenderer->setSplitStemTracks(m_splitStemTracks);
     connect(this,
             &QmlWaveformRendererStem::gainAllChanged,
             pRenderer.get(),
--- a/src/qml/qmlwaveformrenderer.cpp
+++ b/src/qml/qmlwaveformrenderer.cpp
@@ -1,0 +1,233 @@
+#include "qml/qmlwaveformrenderer.h"
+
+#include <memory>
+
+#include "moc_qmlwaveformrenderer.cpp"
+#include "util/assert.h"
+#include "waveform/renderers/allshader/waveformrenderbeat.h"
+#include "waveform/renderers/allshader/waveformrendererendoftrack.h"
+#include "waveform/renderers/allshader/waveformrendererpreroll.h"
+#include "waveform/renderers/allshader/waveformrendererrgb.h"
+#ifdef __STEM__
+#include "waveform/renderers/allshader/waveformrendererstem.h"
+#endif
+#include "waveform/renderers/allshader/waveformrendermark.h"
+#include "waveform/renderers/allshader/waveformrendermarkrange.h"
+
+namespace mixxx {
+namespace qml {
+
+QmlWaveformRendererMark::QmlWaveformRendererMark()
+        : m_defaultMark(nullptr),
+          m_untilMark(std::make_unique<QmlWaveformUntilMark>()) {
+}
+
+QmlWaveformRendererFactory::Renderer QmlWaveformRendererEndOfTrack::create(
+        WaveformWidgetRenderer* waveformWidget) const {
+    auto pRenderer = std::make_unique<allshader::WaveformRendererEndOfTrack>(waveformWidget);
+
+    pRenderer->setColor(m_color);
+    pRenderer->setEndOfTrackWarningTime(m_endOfTrackWarningTime);
+    connect(this,
+            &QmlWaveformRendererEndOfTrack::colorChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererEndOfTrack::setColor);
+    connect(this,
+            &QmlWaveformRendererEndOfTrack::endOfTrackWarningTimeChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererEndOfTrack::setEndOfTrackWarningTime);
+    return QmlWaveformRendererFactory::Renderer{pRenderer.get(), std::move(pRenderer)};
+}
+
+QmlWaveformRendererFactory::Renderer QmlWaveformRendererPreroll::create(
+        WaveformWidgetRenderer* waveformWidget) const {
+    auto pRenderer = std::make_unique<allshader::WaveformRendererPreroll>(
+            waveformWidget, m_position);
+    pRenderer->setColor(m_color);
+    connect(this,
+            &QmlWaveformRendererPreroll::colorChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererPreroll::setColor);
+    return QmlWaveformRendererFactory::Renderer{pRenderer.get(), std::move(pRenderer)};
+}
+
+QmlWaveformRendererFactory::Renderer QmlWaveformRendererRGB::create(
+        WaveformWidgetRenderer* waveformWidget) const {
+    auto pRenderer = std::make_unique<allshader::WaveformRendererRGB>(
+            waveformWidget, m_position, m_options);
+
+    pRenderer->setAxesColor(m_axesColor);
+    pRenderer->setLowColor(m_lowColor);
+    pRenderer->setMidColor(m_midColor);
+    pRenderer->setHighColor(m_highColor);
+    connect(this,
+            &QmlWaveformRendererRGB::axesColorChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererRGB::setAxesColor);
+    connect(this,
+            &QmlWaveformRendererRGB::lowColorChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererRGB::setLowColor);
+    connect(this,
+            &QmlWaveformRendererRGB::midColorChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererRGB::setMidColor);
+    connect(this,
+            &QmlWaveformRendererRGB::highColorChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererRGB::setHighColor);
+
+    pRenderer->setAllChannelVisualGain(m_gainAll);
+    pRenderer->setLowVisualGain(m_gainLow);
+    pRenderer->setMidVisualGain(m_gainMid);
+    pRenderer->setHighVisualGain(m_gainHigh);
+    connect(this,
+            &QmlWaveformRendererRGB::gainAllChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererRGB::setAllChannelVisualGain);
+    connect(this,
+            &QmlWaveformRendererRGB::gainLowChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererRGB::setLowVisualGain);
+    connect(this,
+            &QmlWaveformRendererRGB::gainMidChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererRGB::setMidVisualGain);
+    connect(this,
+            &QmlWaveformRendererRGB::gainHighChanged,
+            pRenderer.get(),
+            &allshader::WaveformRendererRGB::setHighVisualGain);
+    return QmlWaveformRendererFactory::Renderer{pRenderer.get(), std::move(pRenderer)};
+}
+
+QmlWaveformRendererFactory::Renderer QmlWaveformRendererBeat::create(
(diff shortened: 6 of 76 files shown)


acolombier:
Note that this is still being actively worked on with @m0dB and myself!

Holzhaus:
Unfortunately, there are merge conflicts now.

Holzhaus:
FYI, the PR is still broken :(

```
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


endoftrack.frag -> endoftrack.frag.qsb exposed as :/shaders/rendergraph/endoftrack.frag.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


endoftrack.vert -> endoftrack.vert.qsb exposed as :/shaders/rendergraph/endoftrack.vert.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


pattern.frag -> pattern.frag.qsb exposed as :/shaders/rendergraph/pattern.frag.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


pattern.vert -> pattern.vert.qsb exposed as :/shaders/rendergraph/pattern.vert.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


rgb.frag -> rgb.frag.qsb exposed as :/shaders/rendergraph/rgb.frag.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


rgb.vert -> rgb.vert.qsb exposed as :/shaders/rendergraph/rgb.vert.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


rgba.frag -> rgba.frag.qsb exposed as :/shaders/rendergraph/rgba.frag.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


rgba.vert -> rgba.vert.qsb exposed as :/shaders/rendergraph/rgba.vert.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


texture.frag -> texture.frag.qsb exposed as :/shaders/rendergraph/texture.frag.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


texture.vert -> texture.vert.qsb exposed as :/shaders/rendergraph/texture.vert.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)


unicolor.frag -> unicolor.frag.qsb exposed as :/shaders/rendergraph/unicolor.frag.qsb
CMake Error at /usr/lib/cmake/Qt6Core/Qt6CoreMacros.cmake:2127 (get_target_property):
  get_target_property() called with non-existent target "rendergraph_sg".
Call Stack (most recent call first):
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:103 (_qt_internal_expose_source_file_to_ide)
  /usr/lib/cmake/Qt6ShaderToolsTools/Qt6ShaderToolsMacros.cmake:354 (_qt_internal_add_shaders_impl)
  res/shaders/rendergraph/CMakeLists.txt:16 (qt6_add_shaders)
```

acolombier:
@Holzhaus this should now be ready for some very early testing!

JoergAtGithub:
@m0dB @acolombier I wonder if a clearer seperation between QML and QtQuick would be of benefit here, as we could use a QQuickWindow instead of the QOpenGLWindow in the existing QWidget GUI: https://doc.qt.io/qt-6/qtquick-embeddedinwidgets-example.html
Wouldn't this provide us the use of native graphic APIs, instead of OpenGL - for all builds with recent Qt? And we are just in the process to update the VCPKG buildenv for macOS and Windows builds to Qt6.8.1.

Holzhaus:
Hi, configuring works now. I'm getting the following build error now:

```cpp
[ 25%] Building CXX object CMakeFiles/mixxx-qml-lib.dir/src/qml/qmlstemsmodel.cpp.o
In file included from src/rendergraph/scenegraph/../common/rendergraph/assert.h:4,
                 from src/rendergraph/scenegraph/../common/rendergraph/material.h:6,
                 from src/rendergraph/scenegraph/../common/rendergraph/geometrynode.h:5,
                 from src/waveform/renderers/allshader/waveformrendermark.h:5,
                 from src/waveform/renderers/allshader/waveformrendermark.cpp:1:
src/waveform/renderers/allshader/waveformrendermark.cpp: In member function ‘void {anonymous}::WaveformMarkNode::update(float, float, float)’:
src/waveform/renderers/allshader/waveformrendermark.cpp:50:35: error: ‘RoundToPixel’ was not declared in this scope
   50 |         DEBUG_ASSERT(std::abs(x - RoundToPixel(devicePixelRatio)(x)) < epsilon);
      |                                   ^~~~~~~~~~~~
src/rendergraph/scenegraph/../common/rendergraph/../../../util/assert.h:59:32: note: in definition of macro ‘DEBUG_ASSERT’
   59 |         if (!static_cast<bool>(cond)) [[unlikely]] {                        \
      |                                ^~~~
src/waveform/renderers/allshader/waveformrendermark.cpp:51:35: error: ‘RoundToPixel’ was not declared in this scope
   51 |         DEBUG_ASSERT(std::abs(y - RoundToPixel(devicePixelRatio)(y)) < epsilon);
      |                                   ^~~~~~~~~~~~
src/rendergraph/scenegraph/../common/rendergraph/../../../util/assert.h:59:32: note: in definition of macro ‘DEBUG_ASSERT’
   59 |         if (!static_cast<bool>(cond)) [[unlikely]] {                        \
      |                                ^~~~
make[2]: *** [CMakeFiles/mixxx-qml-lib.dir/build.make:666: CMakeFiles/mixxx-qml-lib.dir/src/waveform/renderers/allshader/waveformrendermark.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:803: CMakeFiles/mixxx-qml-lib.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```

m0dB:
I fixed this in the rg-wr-… PR, so if @acolombier merges the latest commits of my branches it will be fixed here too.

acolombier:
@Swiftb0y I assume you are currently looking into this - you reacted with the :eye: on the PR - thank you!

There is still a lot to work on, but I'd been keen to hear your feedback on how to propagate properties for each renderers.

Currently, there is two clashing design, and I will need to unify it into one, but I'm having a hard time to take a decision.

- **Design 1**: The renderer gets initialised (the `setup` method) and holds properties statically. This means that the `LegacySkinParser` will need to extract these property from the XMLs and inject it, in the same way that the QML proxy will do it with its property. Pro: contained, no lifecycle issue, easy to maintain. Con: No dynamic propagation - property update requires to rebuild the renderer stack (slow!)
- **Design 2**: The renderer gets initialised with a reference to the a "provider", which holds properties. This means that the `LegacySkinParser` (or some object in between) will need to extract and hold these properties from the XMLs and will provide them ondemand over few getters, in the same way that the QML proxy will do it with its property. Pro: generic interface, dynamic update, quick feedback. Con: lifecycle (no parenting between QML object and renderer, possible hierarchy on Legacy)

It feels like there is no perfect answers. Also QML is still quite unstable (note the number of changes made in this PR to consolidate it - happy to move some in other PR btw) so I don't think we should overly concern on stability yet with QML, as long as we don't introduce risks on legacy.

Swiftb0y:
> I assume you are currently looking into this - you reacted with the 👁️ on the PR - thank you!

I'm actually not atm, I'm sorry my reaction gave that impression. I'll try to once the preliminary PRs have been merged. 

I have not looked enough into this PR to even understand the problem you're describing, so I can't provide any opinion.

Holzhaus:
Ok, I applied to following patch to work around the build error:

```udiff
diff --git a/src/qml/qmlwaveformrenderer.cpp b/src/qml/qmlwaveformrenderer.cpp
index 3095243945..1d368ebfd4 100644
--- a/src/qml/qmlwaveformrenderer.cpp
+++ b/src/qml/qmlwaveformrenderer.cpp
@@ -89,7 +89,7 @@ QmlWaveformRendererFactory::Renderer QmlWaveformRendererPreroll::create(
 QmlWaveformRendererFactory::Renderer QmlWaveformRendererRGB::create(
         WaveformWidgetRenderer* waveformWidget) const {
     auto* renderer = new WaveformRendererRGB(waveformWidget, m_position, m_options, this);
-    renderer->setup(
+    renderer->setupColors(
             m_axesColor,
             m_lowColor,
             m_midColor,
diff --git a/src/waveform/renderers/allshader/waveformrendererrgb.cpp b/src/waveform/renderers/allshader/waveformrendererrgb.cpp
index 40b92d28e5..409232a66a 100644
--- a/src/waveform/renderers/allshader/waveformrendererrgb.cpp
+++ b/src/waveform/renderers/allshader/waveformrendererrgb.cpp
@@ -29,7 +29,7 @@ WaveformRendererRGB::WaveformRendererRGB(WaveformWidgetRenderer* waveformWidget,
     setUsePreprocess(true);
 }

-void WaveformRendererRGB::setup(const QColor& axesColor,
+void WaveformRendererRGB::setupColors(const QColor& axesColor,
         const QColor& lowColor,
         const QColor& midColor,
         const QColor& highColor) {
diff --git a/src/waveform/renderers/allshader/waveformrendererrgb.h b/src/waveform/renderers/allshader/waveformrendererrgb.h
index acd06ec50f..47833edd0d 100644
--- a/src/waveform/renderers/allshader/waveformrendererrgb.h
+++ b/src/waveform/renderers/allshader/waveformrendererrgb.h
@@ -21,7 +21,7 @@ class allshader::WaveformRendererRGB final
     // Pure virtual from WaveformRendererSignalBase, not used
     void onSetup(const QDomNode& node) override;

-    void setup(const QColor& axesColor,
+    void setupColors(const QColor& axesColor,
             const QColor& lowColor,
             const QColor& midColor,
             const QColor& highColor);
```


Now it crashes with a segfault:

```
Core was generated by `./mixxx --qml'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  ControlDoublePrivate::getKey (this=<optimized out>) at /home/jan/Projects/mixxx/src/control/control.h:143
143	        return m_key;
[Current thread is 1 (Thread 0x7ed96b7fe6c0 (LWP 176578))]
>>> bt
#0  ControlDoublePrivate::getKey (this=<optimized out>) at /home/jan/Projects/mixxx/src/control/control.h:143
#1  ControlProxy::getKey (this=0x0) at /home/jan/Projects/mixxx/src/control/controlproxy.cpp:25
#2  0x000062b631380b8d in allshader_sg::WaveformRendererStem::init (this=0x7ed95c366280)
    at /home/jan/Projects/mixxx/src/waveform/renderers/allshader/waveformrendererstem.cpp:35
#3  0x000062b630e2ec04 in WaveformWidgetRenderer::init (this=this@entry=0x62b65a1149c8)
    at /usr/include/qt6/QtCore/qarraydatapointer.h:119
#4  0x000062b63135c21e in mixxx::qml::QmlWaveformDisplay::updatePaintNode
    (this=0x62b65a1149a0, node=<optimized out>) at /home/jan/Projects/mixxx/src/qml/qmlwaveformdisplay.cpp:132
#5  0x00007edac7a6e984 in QQuickWindowPrivate::updateDirtyNode (this=0x62b661806220, item=0x62b65a1149a0)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/quick/items/qquickwindow.cpp:2271
#6  0x00007edac7a61e1d in QQuickWindowPrivate::updateDirtyNodes (this=this@entry=0x62b661806220)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/quick/items/qquickwindow.cpp:2013
#7  0x00007edac7a629c6 in QQuickWindowPrivate::syncSceneGraph (this=<optimized out>)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/quick/items/qquickwindow.cpp:566
#8  0x00007edac7c465db in QSGRenderThread::sync (this=<optimized out>, inExpose=true)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/quick/scenegraph/qsgthreadedrenderloop.cpp:553
#9  QSGRenderThread::syncAndRender (this=<optimized out>)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/quick/scenegraph/qsgthreadedrenderloop.cpp:716
#10 QSGRenderThread::run (this=0x62b664890830)
    at /usr/src/debug/qt6-declarative/qtdeclarative/src/quick/scenegraph/qsgthreadedrenderloop.cpp:975
#11 0x00007edac22d831b in operator() (__closure=<optimized out>)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qthread_unix.cpp:375
#12 (anonymous namespace)::terminate_on_exception<QThreadPrivate::start(void*)::<lambda()> > (t=<optimized out>)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qthread_unix.cpp:311
#13 QThreadPrivate::start (arg=0x62b664890830)
    at /usr/src/debug/qt6-base/qtbase/src/corelib/thread/qthread_unix.cpp:339
#14 0x00007edac1aa32ce in ??? () at /usr/lib/libc.so.6
#15 0x00007edac1b2829c in ??? () at /usr/lib/libc.so.6
```

Holzhaus:
@acolombier Kudos, looks great! And the annoying flickering that we saw with the previous waveforms is also gone.

Some minor issues I found:
1. Something is wrong with the clipping - if i turn up the Hi/Mids using the EQ, the waveforms overlap the rest of the UI.
2. The waveform marks wiggle around a bit when scratching.


https://github.com/user-attachments/assets/74708116-7b91-4dc1-b57e-607e69b347b9

acolombier:
Thanks for you feedback @Holzhaus! Kudos to @m0dB to make that possible with the rendergraph!
 
> Some minor issues I found:
> 
>     1. Something is wrong with the clipping - if i turn up the Hi/Mids using the EQ, the waveforms overlap the rest of the UI.
I'm actually still not completely sure how to use the scene graph clipping - Any pointer or fix for this @m0dB ?
>     2. The waveform marks wiggle around a bit when scratching.
Indeed, I can also reproduce it locally - it feels like the renderer are not synched for some reason (beats are responding faster than signal, itself responding faster than marker). I must say here again, I have no idea where to start and could definitely use some help @m0dB :smile:

acolombier:
> 2. The waveform marks wiggle around a bit when scratching. 

This seems to be better now, I would appreciate if you could give this another go 

Regarding [this comment](https://github.com/mixxxdj/mixxx/pull/13470#issuecomment-2631091530), I went with a 3rd option, using signal/slots between the QML element and the rendergraph instances

acolombier:
I'm getting `Target "rendergraph_sg" links to target "Qt6::Quick" but the target was not` on Windows - any idea how this can be solved? I believe QtQuick should already be included in VCPKG, as it builds fine on Mac

src/qml/qmlwaveformrenderer.h
```diff
@@ -0,0 +1,586 @@
...
+    void colorChanged(const QColor&);
+
+  private:
+    QColor m_color;
+    ::WaveformRendererAbstract::PositionSource m_position{::WaveformRendererAbstract::Play};
+};
+
+class QmlWaveformMarkRange : public QObject {
+    Q_OBJECT
+    Q_PROPERTY(QColor color READ color WRITE setColor)
```
Swiftb0y, line 237: IIRC properties also need a `NOTIFY` to properly work in Qml. At least if they're supposed to support reactive bindings.

src/qml/qmlwaveformrenderer.h
```diff
@@ -0,0 +1,586 @@
...
+    QmlWaveformRendererMarkRange();
+
+    QQmlListProperty<QmlWaveformMarkRange> ranges() {
+        return {this, &m_ranges};
+    }
+
+    Renderer create(WaveformWidgetRenderer* waveformWidget) const override;
+
+  private:
+    QList<QmlWaveformMarkRange*> m_ranges;
```
Swiftb0y, line 490: what are the lifetime requirements on the `QmlWaveformMarkRange*`? Is it possible to get a use after free if Qml holds on to this list after a new track has been loaded?

Swiftb0y (commented):
couple surface-level things I noticed

src/qml/qmlwaveformrenderer.h
```diff
@@ -0,0 +1,586 @@
...
+    QML_NAMED_ELEMENT(WaveformMarkRange)
+
+  public:
+    QColor color() const {
+        return m_color;
+    }
+
+    void setColor(const QColor& value) {
+        m_color = value;
+    }
```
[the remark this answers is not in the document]
acolombier: Yup, makes sense!

src/qml/qmlwaveformrenderer.h
acolombier: Yeah, it's missing on few since the property/binding was not used previously, but with my yesterday commit, it should be possible to allow runtime update of those so I'll had the missing one

src/qml/qmlwaveformrenderer.h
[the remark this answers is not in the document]
acolombier: Changed and removed unused accessor, reduce file from 587 lines to 430

acolombier:
@Holzhaus the two issues you raised should be fixed now, could you please give it another go?

res/controllers/DummyDeviceDefaultScreen.qml
```diff
@@ -286,14 +286,165 @@ Mixxx.ControllerScreen {
...
+                            color: 'blue'
+                            endOfTrackWarningTime: 30
+                        }
+
+                        Mixxx.WaveformRendererPreroll {
+                            color: '#998977'
+                        }
+
+                        Mixxx.WaveformRendererMarkRange {
+                            // <!-- Loop -->
```
Holzhaus, line 332:
```suggestion
                            // Loop
```
And elsewhere

res/qml/Deck.qml
```diff
@@ -326,6 +326,54 @@ Item {
...
+            anchors.leftMargin: 10
+            anchors.bottom: playButton.bottom
+            anchors.topMargin: 5
+            spacing: -1
+
+            Skin.IntroOutroButton {
+                keyPrefix: "intro_start"
+                group: root.group
+
+                text: "Intro\nStart"
```
Holzhaus, line 340: What do you think about this?

```suggestion
                text: '◢'
```

res/qml/WaveformDisplay.qml
```diff
@@ -0,0 +1,247 @@
...
+            color: '#ff8872'
+            endOfTrackWarningTime: 30
+        }
+
+        Mixxx.WaveformRendererPreroll {
+            color: '#ff8872'
+        }
+
+        Mixxx.WaveformRendererMarkRange {
+            // <!-- Loop -->
```
Holzhaus, line 34:
```suggestion
            // Loop
```
And elsewhere

res/qml/WaveformDisplay.qml
```diff
@@ -0,0 +1,247 @@
...
+                textColor: "#1a1a1a"
+                text: " %1 "
+            }
+
+            untilMark.showTime: true
+            untilMark.showBeats: true
+            untilMark.align: Mixxx.WaveformUntilMark.AlignBottom
+            untilMark.textSize: 11
+
+            Mixxx.WaveformMark {
```
Holzhaus, line 99: Btw, in the long term I'd like to be able to define a mark delegate from QML. That would allow different skins to have a way to define a custom look (not just custom colors) for waveform marks.

Maybe by defining a QML item here, which is then rendered to a pixmap and then copied to the waveform. Nothing for this PR, but this would be a long term goal.

src/qml/qmlwaveformrenderer.cpp
```diff
@@ -0,0 +1,243 @@
...
+    connect(this, &QmlWaveformRendererBeat::colorChanged, renderer, &WaveformRenderBeat::setColor);
+    return QmlWaveformRendererFactory::Renderer{renderer, renderer};
+}
+
+QmlWaveformRendererFactory::Renderer QmlWaveformRendererMarkRange::create(
+        WaveformWidgetRenderer* waveformWidget) const {
+    auto* renderer = new WaveformRenderMarkRange(
+            waveformWidget);
+
+    for (auto* pMark : m_ranges) {
```
Holzhaus, line 136: Shouldn't this be const?

src/qml/qmlwaveformrenderer.cpp
```diff
@@ -0,0 +1,243 @@
...
+QmlWaveformRendererFactory::Renderer QmlWaveformRendererMark::create(
+        WaveformWidgetRenderer* waveformWidget) const {
+    VERIFY_OR_DEBUG_ASSERT(!!m_untilMark) {
+        return QmlWaveformRendererFactory::Renderer{};
+    }
+    auto* renderer = new WaveformRenderMark(waveformWidget,
+            ::WaveformRendererAbstract::Play);
+
+    renderer->setFgPlayColor(m_playMarkerColor);
+    renderer->setBgPlayColor(m_playMarkerBackground);
```
Holzhaus, line 177: I'd suggest to name these consistently and expressive:

```suggestion
    renderer->setPlayMarkerForegroundColor(m_playMarkerForegroundColor);
    renderer->setPlayMarkerBackgroundColor(m_playMarkerBackgroundColor);
```

src/qml/qmlwaveformrenderer.cpp
```diff
@@ -0,0 +1,243 @@
...
+            &QmlWaveformRendererMark::playMarkerColorChanged,
+            renderer,
+            &WaveformRenderMark::setFgPlayColor);
+    connect(this,
+            &QmlWaveformRendererMark::playMarkerBackgroundChanged,
+            renderer,
+            &WaveformRenderMark::setBgPlayColor);
+
+    int priority = 0;
+    for (auto* pMark : m_marks) {
```
Holzhaus, line 212: Const?

src/qml/qmlwaveformrenderer.cpp
```diff
@@ -0,0 +1,243 @@
...
+                waveformWidget->getGroup(),
+                pMark->control(),
+                pMark->visibilityControl(),
+                pMark->textColor(),
+                pMark->align(),
+                pMark->text(),
+                pMark->pixmap(),
+                pMark->icon(),
+                pMark->color(),
+                --priority)));
```
Holzhaus, line 223: Is this really supposed to go 0, -1, -2, ...?

Also, I would be in favor of splitting the increment/decrement step onto a separate line for readability.

src/waveform/renderers/allshader/waveformrendererstem.cpp
```diff
@@ -107,7 +107,7 @@ bool WaveformRendererStem::preprocessInner() {
 
     // Per-band gain from the EQ knobs.
     float allGain(1.0);
-    // applyCompensation = true, as we scale to match filtered.all
+    // applyCompensation = false, as we scale to match filtered.all
```
Holzhaus, line 110: I have no idea about this but why did the value change but the comment text did not?

Holzhaus (commented):
Hey, didn't really review the whole thing, just some minor comments.

src/qml/qmlwaveformrenderer.cpp
acolombier: This was auto-generated from the members name, but I think we can rename it as well

src/qml/qmlwaveformrenderer.cpp
acolombier: This was copied from the renderer setup logic used by the legacy skin parser so I cannot do much on what is expected there. Sure, happy to make that on two lines

res/qml/Deck.qml
acolombier: I started with this and it didn't render so well in the UI. Overall, the whole QML design probably needs a good rework anyway (see Zulip) so I thought I'll go with the more explicit now but happy to keep this char if we think it's better

src/qml/qmlwaveformrenderer.cpp
Swiftb0y: use `std::as_const` too. Otherwise it may still detach the container.

src/qml/qmlwaveformrenderer.cpp
acolombier: Added a comment above to explain where this logic comes from

src/waveform/renderers/allshader/waveformrendererstem.cpp
acolombier: The comment was incorrect (see line above, using `applyCompensation = false`). It was copy from [here](https://github.com/mixxxdj/mixxx/blob/7b75517639fa1d9cfcc1905291ed2bd25122cc6a/src/waveform/renderers/allshader/waveformrendererrgb.cpp#L110) so looks like I must have messed up by accident when I implemented this renderer

res/qml/Deck.qml
acolombier: Hmmm not a big fan
image

Wdyt? Shall I keep with text till we can to properly design this?

Holzhaus: I think the icon looks weird, but better than the text ;-) but I leave it up to you, we need to improve the UI later on anyway (Maybe have a frame around both buttons, put the intro icon as the group label, and then have something like `|-` and `-|` as button labels or something like that).

acolombier: Yeah, for now I've been only focusing on feature, since I'm also working on a whole design project for QML so didn't want to waste time with temporary measure for this

Holzhaus: Yup. I think the waveforms and the library are the two major issues that prevent me from using in in a set. Most other stuff is working from my controller anyway 😁

Holzhaus: Yup. I think the waveforms and the library are the two major issues that prevent me from making qml actually usable in a set. Most other stuff is working from my controller anyway 😁

JoergAtGithub:
I could build this PR locally now, the build passed, but there are obviously errors during the build, which are not detected as build fail:
```
  [22/941] Generating .qsb/unicolor.frag.qsb
  Failed to run spirv-opt -O C:\Users\Joerg.WORLDWARTWEB\AppData\Local\Temp\qsb-NkwHbf\qsb_spv_temp -o C:\Users\Joerg.WORLDWARTWEB\AppData\Local\Temp\qsb-NkwHbf\qsb_spv_temp_out: Process failed to start: Das System kann die angegebene Datei nicht finden.
```

JoergAtGithub:
At the start I get always asserts, because renderer.renderer is NULL:
```
Program: D:\mixxx\build\x64__off\Qt6Cored.dll
Module: 6.5.3
File: D:\mixxx\src\util\logging.cpp
Line: 372

DEBUG ASSERT: "renderer.renderer" in function class QSGNode *__cdecl mixxx::qml::QmlWaveformDisplay::updatePaintNode(class QSGNode *,struct QQuickItem::UpdatePaintNodeData *) at D:\mixxx\src\qml\qmlwaveformdisplay.cpp:123
```
If I ignore these asserts, it starts up and work somehow. With much lower CPU usage than the normal OpenGL version.

acolombier:
> It seems spirv-tools is not specified in our VCPKG buildenv. But the configuration seems also be wrong, as it runs this build step in %temp% instead of the configured build directory.

I'll try to look into this - I've also spotted the problem, but not entirely sure how to approach it

> At the start I get always asserts, because renderer.renderer is NULL:

Ah yes, this is likely because you are running with `STEM=OFF`, leading to a [null renderer](https://github.com/mixxxdj/mixxx/blob/7b6888943a4119054d42dd5007352d24773a7487/src/qml/qmlwaveformrenderer.h#L357). I'll add some kind of `isSupported` logic to silence this correctly.

> Mixxx starts up and waveforms work. With much lower CPU usage than the normal OpenGL version.

That is great to hear!

acolombier:
> > It seems spirv-tools is not specified in our VCPKG buildenv. But the configuration seems also be wrong, as it runs this build step in %temp% instead of the configured build directory.
> 
> I'll try to look into this - I've also spotted the problem, but not entirely sure how to approach it

After investigation, this is coming from [this rendergraph command](https://github.com/mixxxdj/mixxx/blob/84f8966f79389d0f87573f20be23b7bd4f34d9fa/src/rendergraph/shaders/CMakeLists.txt#L35). I will see if I can fix it within a given time, but I'd like this not to block this feature to go in 2.6. Note the byproduct if this failure, is only to lack some shader optimisation, which didn't exist prior rendergraph so no regression there.

acolombier:
> > It seems spirv-tools is not specified in our VCPKG buildenv. But the configuration seems also be wrong, as it runs this build step in %temp% instead of the configured build directory.
> 
> I'll try to look into this - I've also spotted the problem, but not entirely sure how to approach it

Sadly, this is no easy fix, so I have disabled the broken optimisation of the shader. This will need to be looked at separately

acolombier:
@JoergAtGithub all the issues reported should be fixed now - I have squashed my commits so this is ready to go for 2.6!

src/waveform/renderers/allshader/waveformrendererstem.cpp
```diff
@@ -121,7 +135,7 @@ bool WaveformRendererStem::preprocessInner() {
 
     const int numVerticesPerLine = 6; // 2 triangles
 
-    const int reserved = numVerticesPerLine * (8 * pixelLength + 1);
+    const int reserved = numVerticesPerLine * (8 * stripLength + 1);
```
JoergAtGithub, line 138:
```suggestion
    const int reserved = numVerticesPerLine * (mixxx::audio::ChannelCount::stem() * stripLength + 1);
```
Did I guess right, where 8 comes from?

src/waveform/renderers/waveformrenderersignalbase.cpp
```diff
@@ -188,19 +197,16 @@ void WaveformRendererSignalBase::getGains(float* pAllGain,
...
 
-            lowGain *= static_cast<CSAMPLE_GAIN>(
-                    factory->getVisualGain(WaveformWidgetFactory::Low));
-            midGain *= static_cast<CSAMPLE_GAIN>(
-                    factory->getVisualGain(WaveformWidgetFactory::Mid));
-            highGain *= static_cast<CSAMPLE_GAIN>(
-                    factory->getVisualGain(WaveformWidgetFactory::High));
+            lowGain *= m_lowVisualGain;
+            midGain *= m_midVisualGain;
+            highGain *= m_highVisualGain;
```
JoergAtGithub, line 209:
```suggestion
            lowVisualGain *= m_lowVisualGain;
            midVisualGain *= m_midVisualGain;
            highVisualGain *= m_highVisualGain;
```
Left hand side is also a visual gain

JoergAtGithub (changes requested):
Please fix the remaining 10 Clazy warnings: https://github.com/mixxxdj/mixxx/actions/runs/13497007080?pr=13470

src/waveform/waveform.h
```diff
@@ -12,7 +12,11 @@
 #include "util/class.h"
 #include "util/compatibility/qmutex.h"
 
-enum FilterIndex { Low = 0, Mid = 1, High = 2, FilterCount = 3};
+enum FilterIndex { AllChannel = 0,
+    Low = 1,
+    Mid = 2,
+    High = 3,
+    FilterCount = 4 };
```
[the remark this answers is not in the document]
acolombier: Makes sense - since the enum will create a symbol on the global scope, I kept `All` and `Count` prefixed with enum type. This could probably use a cleanup but this is out of scope for this PR.

JoergAtGithub (approved):
Code LGTM! Works flawless on Windows, with and without --qml!

src/rendergraph/scenegraph/backend/baseattributeset.cpp
```diff
@@ -9,6 +11,8 @@ int toQSGGeometryType(const PrimitiveType& t) {
         return QSGGeometry::FloatType;
     case PrimitiveType::UInt:
         return QSGGeometry::UnsignedIntType;
+    default:
+        throw std::runtime_error("not implemented");
```
Swiftb0y, line 15: two nits: 
1. exceptions in Mixxx are a touchy subject. 
2. this essentially silences `-Wswitch` which is usually undesirable. If you need to add a return value or throw (because the compiler complains otherwise), just do it after the switch/outside.

src/waveform/renderers/allshader/waveformrenderbeat.h
```diff
@@ -14,8 +14,10 @@ class WaveformRenderBeat;
...
 
 class allshader::WaveformRenderBeat final
-        : public ::WaveformRendererAbstract,
+        : public QObject,
+          public ::WaveformRendererAbstract,
           public rendergraph::GeometryNode {
+    Q_OBJECT
   public:
     explicit WaveformRenderBeat(WaveformWidgetRenderer* waveformWidget,
             ::WaveformRendererAbstract::PositionSource type =
```
Swiftb0y, line 23: I don't get why this needs to be a Q_OBJECT. You don't need to declare a member function as slot in order to bind it to a signal. Any member function is automatically a slot. Only the object that emits the signal needs to be the QObject afaik.

src/waveform/renderers/allshader/waveformrendererstem.cpp
```diff
@@ -14,6 +14,17 @@
...
+namespace {
+#ifdef __SCENEGRAPH__
+// FIXME this is a workaround an issue with waveform only drawing partially in
+// SG. The workaround is to reduce the the number of vertices, by reducing the
+// precision of waveform strips.
+const float kPixelPerStrip = 2;
+#else
+const float kPixelPerStrip = 1;
+#endif
+} // namespace
```
Swiftb0y, line 26: file an issue after merge so we don't forget?

Swiftb0y (commented):
couple high level thoughts

src/mixxxmainwindow.cpp
```diff
@@ -515,6 +515,8 @@ MixxxMainWindow::~MixxxMainWindow() {
 
     m_pCoreServices->getControlIndicatorTimer()->setLegacyVsyncEnabled(false);
 
+    qDebug() << t.elapsed(false).debugMillisWithUnit() << "deleting ControllerManager";
```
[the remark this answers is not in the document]
acolombier: I've used this to keep it consistent with the other debug message above, but definitely could use a clean up. Happy for this going in follow up work? There is more job to do around timer cleanup and standardisation around `chrono`. I have tried to keep the new code more align with this, but it was hard to draw the line somewhere to not add up an additional batch of changes in this piece of work (see 8ece65b5fcedc93587708fae2cc394c5f352a54d).

src/rendergraph/scenegraph/backend/baseattributeset.cpp
acolombier: Yeah, I'm definitely keen to control the proliferation of pesky exception. I've added this one to align with the code style of the rendergraph (see [here](https://github.com/mixxxdj/mixxx/blob/89bf0a23bfa795cb4b002f5443a8079ea0797de4/src/rendergraph/opengl/backend/basegeometrynode.cpp#L21), [here](https://github.com/mixxxdj/mixxx/blob/89bf0a23bfa795cb4b002f5443a8079ea0797de4/src/rendergraph/scenegraph/geometry.cpp#L17) or [here](https://github.com/mixxxdj/mixxx/blob/89bf0a23bfa795cb4b002f5443a8079ea0797de4/src/rendergraph/scenegraph/geometry.cpp#L28)). I'm almost tempted to replace these with a `VERIFY_OR_DEBUG_ASSERT` but felt like potential cleanup task to keep out of this PR scope.

Happy to "break" the rendergraph style already here tho if you think this is sensible

src/waveform/renderers/allshader/waveformrendererstem.cpp
acolombier: Yep, good call - created as #14407

src/waveform/renderers/allshader/waveformrenderbeat.h
acolombier: It's not a warning, it prevents compiling entirely. Or is this a "warning as error" case?

> is not related to any other changes in the class?

Kinda  - this class was never build with `allshader=allshader_sg` and `rendergraph=rendergraph_sg` before so these are the "major changes" here

Swiftb0y: Ah I see. Yeah makes sense. Thanks for the explanation.

src/rendergraph/scenegraph/backend/baseattributeset.cpp
Swiftb0y: Oh, I'm sorry. I had lost context and thought we were switching on our our own sg `enum`. The discussion is void then.

Holzhaus:
Thanks, looks and works great. 🎉

Once we get this PR in, the only big remaining issue with the QML interface is the library IMHO.

src/waveform/renderers/waveformmarkset.h
```diff
@@ -11,6 +11,17 @@
...
+    struct Seed {
+        QString positionControl;
+        QString visibilityControl;
+        QString textColor;
+        QString markAlign;
+        QString text;
+        QString pixmapPath;
+        QString iconPath;
+        QColor color;
+    };
```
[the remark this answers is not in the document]
acolombier: Sure - the renderer has the concept of "default marker" (<DefaultMarker>` in skin I believe) which is used to put styles on hotcue. QML has a custom object to allow define this, but since it is in the `mixxx-qml-lib`, this couldn't be directly referenced or used. Since there will be soon the concept of "delegate", meaning that it didn't feel worth spending too much effort on refactoring this too much, so I just wanted to add a quick container - does that make sense? It does feel like the name `Seed`, what should I name it? Just `DefaulMarker`?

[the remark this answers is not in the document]
Swiftb0y: Yeah, just rename it to something like `DefaultMarkerStyle`. Potentially add the slated replacement plan/idea as a comment near the type.

acolombier:
Glad you like it! Most of the credit goes to @m0dB tho

Agreed on the library, and the settings, as well as a good revamp of the deck controls, but I'm on it! :D (cf Zulip)

acolombier:
Addressed the last comments and squashed the fixups - ready to go :rocket:

acolombier:
@mixxxdj/developers merge?

src/waveform/renderers/allshader/waveformrendermark.cpp
```diff
@@ -172,10 +173,40 @@ allshader::WaveformRenderMark::WaveformRenderMark(
     }
 }
 
-void allshader::WaveformRenderMark::draw(QPainter*, QPaintEvent*) {
+void allshader::WaveformRenderMark::draw(QPainter* painter, QPaintEvent* event) {
+    Q_UNUSED(painter);
+    Q_UNUSED(event);
     DEBUG_ASSERT(false);
```
Swiftb0y, line 179: codestyle regression technically

https://github.com/mixxxdj/mixxx/wiki/Coding%20Guidelines#unused-parameters

src/qml/qmlwaveformrenderer.cpp
```diff
@@ -0,0 +1,231 @@
...
+namespace qml {
+
+QmlWaveformRendererMark::QmlWaveformRendererMark()
+        : m_defaultMark(nullptr),
+          m_untilMark(std::make_unique<QmlWaveformUntilMark>()) {
+}
+
+QmlWaveformRendererFactory::Renderer QmlWaveformRendererEndOfTrack::create(
+        WaveformWidgetRenderer* waveformWidget) const {
+    auto* pRenderer = new allshader::WaveformRendererEndOfTrack(waveformWidget);
```
Swiftb0y, line 25: `parented_ptr`?

src/waveform/renderers/allshader/waveformrendererendoftrack.cpp
```diff
@@ -79,10 +80,8 @@ bool WaveformRendererEndOfTrack::preprocessInner() {
             kBlinkingPeriodMillis;
 
     const double remainingTime = m_pTimeRemainingControl->get();
-    const double remainingTimeTriggerSeconds =
-            WaveformWidgetFactory::instance()->getEndOfTrackWarningTime();
-    const double criticalIntensity = (remainingTimeTriggerSeconds - remainingTime) /
-            remainingTimeTriggerSeconds;
+    const double criticalIntensity = (m_remainingTimeTriggerSeconds - remainingTime) /
+            m_remainingTimeTriggerSeconds;
```
Swiftb0y, line 84: Isn't this configurable from the user preferences? This would ignore that user preference. Should the skin designer have to wire that up?

src/waveform/renderers/allshader/waveformrendererendoftrack.h
```diff
@@ -34,11 +36,20 @@ class allshader::WaveformRendererEndOfTrack final
...
+    void setEndOfTrackWarningTime(int endOfTrackWarningTime) {
+        m_remainingTimeTriggerSeconds = endOfTrackWarningTime;
+    }
+
   private:
     std::unique_ptr<ControlProxy> m_pEndOfTrackControl;
     std::unique_ptr<ControlProxy> m_pTimeRemainingControl;
 
     QColor m_color;
+    int m_remainingTimeTriggerSeconds;
```
Swiftb0y, line 52: I don't see this getting initialized anywhere. Am I missing something?

src/waveform/renderers/allshader/waveformrendererstem.cpp
```diff
@@ -91,8 +102,10 @@ bool WaveformRendererStem::preprocessInner() {
     const float devicePixelRatio = m_waveformRenderer->getDevicePixelRatio();
     const int length = static_cast<int>(m_waveformRenderer->getLength());
     const int pixelLength = static_cast<int>(m_waveformRenderer->getLength() * devicePixelRatio);
-    const float invDevicePixelRatio = 1.f / devicePixelRatio;
-    const float halfPixelSize = 0.5f / devicePixelRatio;
+    const int stripLength = static_cast<int>(
+            std::floor(static_cast<float>(pixelLength) / kPixelPerStrip));
```
Swiftb0y, line 106: NIT: unnecessary call to `std::floor` afaict. Also [not something the compiler optimizes out](https://compiler-explorer.com/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TA3cAGQImbAA5PwAjbFIQC3IAB3QlYhcmL19/A2TU5yEQsMi2GLiEh2wndJEiFlIiTL8AnntsR3ymGrqiQojo2Pj7WvrG7JalYZ7QvpKBiwBKe3QfUlROLksAZlDUXxwAajNNjwA3SpJSI9wzLQBBG9uT9AJMfcx0AH0ldA4PgHcyABrCChIjzI42O4PJ4vfaJUjoNZKJQfYjYNhKCAYJgTfYTTAgEBnJxkI4eUFXSwANn2xPmhwA7JDbvtWfsWD4SHiAF6HTYAEVplQAdKludgIODNsy2ftBKQIPjCWLsB8iPsCHzBVoIRq%2BR4ebqCLZbPSzEyHrLZe8vj9VQDSMDiWYAKxWY0u/lSmVs838h5%2Bh5cRb0bgu/gBLg6cjobgeE3WPHLVbYQ4WTZ8chEbTBxaAkAurSGbjSfhsAtFyPR2NcfhKEBF7NR4PkOCwFAYHD4YhkSjUOiMVgcbiZwTCMQSTgyOTCZRqTTN8j6BJGExoBN2CpVVwQdyjAIWFrBabFUogBlJFJpIT7%2BItXLXpi9U8DC9bjpdEbeJp31rtaqTM%2B/RxG%2Bky3oeQzdEBswgYs3wrGs3BbDseypmSxIXJSUJ3DCrw2t8vwOsCoLegGOHPK88KItgyKokQ6KYtiuJKkS5ykscFKbNcFg0nSjI%2Bqy8oQByXJomw%2BwgEKqBmhadxWmy%2BF2v8QIgvRbCkXJvoMv6ULaUGIZhhGOYxnGG5JghaHphY/BNjo8yLEg2AsDgcSSsWXCluQ5aFuQVb8DWdYNlmOb2eQ%2BbSC6wpUgAnDw0VaFSWhaC6FiHtFmzuZsRmLgFwXNosbbIGgPyJAwsR9liJVlXEuzGMAPCbFoLQDvRpD1hAUTGVEoR1AAniO/AYGwHDCAA8kw9D9YuOBsHVkjTQQpDnAQZz1ou2DqJUnLrJmoJtMZ9AEFEpB9V4ODGUQpAEOWvAtjQRjAEoABqBDYH8o2JMwA0zqI4iSNOY6KCoGjGfoLSrqYG6GEd9aQIs6CJB0a0ALSjVlf7LTue7fmM5DHkUwE5FeHS3vexPpNBZ7jG0mOdGBOPNBj2501BJ6E%2BM9NZIzEyswTMFSHByaIS0l3YDtLahlw4a%2BcZNbqAAHFSyNUtI%2By1SY%2BwNcKWjCjw%2BwQPG1i2OQ%2BzdhcaYZibXhsKVjCkBbPDzDZIWth2VV2xVQ22wM6v1Y1zUMK17WdYu3WsKQU2ZkNI1EONk3GTNc3rNGhBLVUq3GRtW30d9e2S9Gh3HadWDJzZV03Xwiz3Swj0vW9H1fbdP0Tv9siA3OIOLsuhh1euRvWNDUSw25CNI9wqPo%2B%2B6RuEwngM4Es%2BUwMZN5OkpOXqvBRs/z1P/kIn4NPPu%2B0wfS9xBz3Trzz9RnwLSwWYMoviwZUvZdW3AK0rKtqxDmubNrut9aGxsAPU2hBzZbASPsa23t7aQKdnlOyrsQCEBoDQCqgMW5TjbvIDuC4C5IHrODQhqCaBEF6p9IKpBCGDGoUoUh5DKFaBftLPyJkuD8gIGg/Ydc/ixH2J/ZWqtfZ/wAXrE4SgBGKyET/OqoidaZlsrmcgjlnIDDcvmF00hhSNXllodK0VorSAZAyPRDIqTuU8uWHgSUZY5W4IFBISjQqS2snY9%2BtZEHKLOG1ae0ggA)

src/waveform/renderers/allshader/waveformrendererstem.h
```diff
@@ -33,8 +33,14 @@ class allshader::WaveformRendererStem final
...
     void preprocess() override;
 
+  public slots:
+    void setSplitStemTracks(bool splitStemTracks) {
+        m_splitStemTracks = splitStemTracks;
+    }
+
   private:
     bool m_isSlipRenderer;
+    bool m_splitStemTracks;
```
Swiftb0y, line 43: initialization?

src/waveform/renderers/waveformmark.cpp
```diff
@@ -93,6 +93,75 @@ bool isShowUntilNextPositionControl(const QString& positionControl) {
...
+          m_level{},
+          m_iPriority(priority),
+          m_iHotCue(hotCue),
+          m_showUntilNext{} {
+    QString endPositionControl;
+    QString typeControl;
+    if (hotCue != Cue::kNoHotCue) {
+        positionControl = "hotcue_" + QString::number(hotCue + 1) + "_position";
+        endPositionControl = "hotcue_" + QString::number(hotCue + 1) + "_endposition";
+        typeControl = "hotcue_" + QString::number(hotCue + 1) + "_type";
```
Swiftb0y, line 122: NIT: save the `QString::number` call in a local var. Seems unnecessary to call this often.

src/waveform/renderers/waveformmarkrange.cpp
```diff
@@ -75,6 +75,66 @@ WaveformMarkRange::WaveformMarkRange(
...
+        } else {
+            // Show warning only when there's no EnabledControl,
+            // like for intro & outro ranges.
+            QString rangeSuffix = QStringLiteral("_start_position");
+            QString rangeName = QString(startControl).remove(rangeSuffix);
+            int gray = qGray(m_activeColor.rgb());
+            m_disabledColor = QColor(gray, gray, gray);
+            qDebug() << "Didn't get DisabledColor for mark range" << rangeName
+                     << "- using desaturated Color:" << m_disabledColor;
+        }
```
Swiftb0y, line 134: NIT: expensive string operations just for debug logging

Swiftb0y (commented):
Did another sanity check. Most of the comments are nits, though the missing initialization and the endOfTrackWarning need some discussion.

src/waveform/renderers/waveformmarkrange.cpp
acolombier: This operation is only ran at init time. Also, I always assumed that `qDebug` would no-op the rest of the stream of the operation if it is not enabled, isn't that right?

src/waveform/renderers/allshader/waveformrendererendoftrack.h
acolombier: Nope, it's missing indeed

src/waveform/renderers/allshader/waveformrendererendoftrack.cpp
acolombier: Yes, that's right. The reason for doing so is to allow renderer to be used in some other context (e.g S4 screen rendering), where different setting may be taken into account

src/waveform/renderers/waveformmarkrange.cpp
acolombier: Anyway, I would argue the value of this message is important as it would help indicating user that the specify color was somehow not working (e.g missing, typo, ...). Do you think I should increase the level?

src/qml/qmlwaveformrenderer.cpp
acolombier: Unfortunately, `WaveformWidgetRenderer` doesn't inherit `QObject` so `parented_ptr` cannot be used. Note that the ptr is eventually encapsulated inside a `unique_ptr` [here](https://github.com/mixxxdj/mixxx/blob/5a7e04b80ee6a1c8c54f06ba37193fd975123171/src/qml/qmlwaveformdisplay.cpp#L129) to ensure proper de-allocation.

src/waveform/renderers/waveformmarkrange.cpp
Swiftb0y: > Also, I always assumed that qDebug would no-op the rest of the stream of the operation if it is not enabled, isn't that right?

not really as far as I can tell. `rangeSuffix` and `rangeName` are still created and constructed, no matter how `qDebug` actually behaves. Also even if `qDebug` doesn't do anything, the arguments to the `operator<<` still need to be evaluated. Anyways, as I said, just a nit. I think we can achieve the same without the allocation overhead if you just implement it in terms of `QStringView`

src/qml/qmlwaveformrenderer.cpp
Swiftb0y: Oh, I just assumed it was a QObject and thus managed by the QObject tree. So this is actually a raw owning pointer? In that case we should definitely work a little on this. You can easily convert a `std::unique_ptr<Derived>` to a `std::unique_ptr<Base>` (the types are just not covariant compared to raw pointers). I'm fairly certain you can `std::make_unique` from the start and then also move that into the `QmlWaveformRendererFactory::Renderer`.

src/waveform/renderers/allshader/waveformrendererendoftrack.cpp
Swiftb0y: Right, but that means the user preference is being ignored by default. So this seems like a regression to me.

acolombier: Good catch - I fixed it in bffc6f335389d4131233ca8fba4e29d1393e97f5. I tested that it would dynamically update when editing the settings (since the singleton wouldn't be used in place) and it works well. This allows not needing references to that bloated singleton on the QML code (for which this renderer code is also used)

src/qml/qmlwaveformrenderer.cpp
acolombier: No, unfortunately I can't because I have to pass that pointer twice inside the Render struct. This is because I need to keep track of the two base classes of the renderers (WaveformRenderer and Rendergraph node, unrelated together). I'm sure this is very hacky but unfortunately I cannot think of any other ways to wrap that elegantly, without adding some bloating on allshader and rendergraph.
Happy to try an alternative if you have ideas. Note that I cannot use a shared_ptr since one of its usage (inside rendergraph) will be downcasted to unique_ptr

Swiftb0y: I think its fairly straightforward if adapt something like this:

```cpp

class A {};
class B {};
class C : public A, public B {};

struct Renderer {
    // ensure owning ptr is before the borrowing one
    // to ensure the borrow doesn't outlive the owner.
    std::unique_ptr<A> node{nullptr};
    B* renderer{nullptr};
};

Renderer create() {
    auto c = std::make_unique<C>();
    // local temporary is needed so we don't call `.get` on `c` after it may have been moved from
    B* b = c.get();
    return Renderer{std::move(c), b};
}

```

wdyt?

src/waveform/renderers/waveformmarkrange.cpp
Swiftb0y:
```suggestion
            QStringView rangeName = rangeSuffix; 
            constexpr static QStringView rangeSuffix = "_start_position";
            if (startControl.endsWith(rangeSuffix)) {
                rangeName.chop(rangeSuffix.length);
            }
            int gray = qGray(m_activeColor.rgb());
            m_disabledColor = QColor(gray, gray, gray);
            qDebug() << "Didn't get DisabledColor for mark range" << rangeName
                     << "- using desaturated Color:" << m_disabledColor;
        }
```

src/qml/qmlwaveformrenderer.cpp
acolombier: Yep, works with me. I thought it was a bit strange to use a mixture of both, but happy to do so if you think it's the best tradeoff

Swiftb0y: I think so, yes. Lets try our best to adhere to [`ES.60: Avoid new and delete outside resource management functions`](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es60-avoid-new-and-delete-outside-resource-management-functions)

src/waveform/renderers/waveformmarkrange.cpp
acolombier: Ah sorry, I just realised you weren't talking just about this `qDebug`, but the whole `else` block. I only copied this code and didn't realise that

Swiftb0y (approved):
LGTM. Thank you.

acolombier:
Thanks @Swiftb0y - all squashed and ready to go now!

Holzhaus:
There a still a bunch of fixup commits

acolombier:
Indeed, I didn't notice that the push failed for some unrelated git hooks checks. We should be good to go now!

CMakeLists.txt
```diff
@@ -4714,8 +4717,24 @@ endif()
...
 # rendergraph
 add_subdirectory(src/rendergraph)
+target_compile_definitions(
+  rendergraph_gl
+  PUBLIC $<$<CONFIG:Debug>:MIXXX_DEBUG_ASSERTIONS_ENABLED>
+)
 target_link_libraries(mixxx-lib PUBLIC rendergraph_gl)
 target_compile_definitions(mixxx-lib PUBLIC rendergraph=rendergraph_gl)
+target_compile_definitions(mixxx-lib PRIVATE allshader=allshader_gl)
+target_compile_definitions(mixxx-test PRIVATE allshader=allshader_gl)
```
PPN-SD, line 4727: missing if(BUILD_TESTING)/endif()

```
CMake Error at CMakeLists.txt:4727 (target_compile_definitions):
  Cannot specify compile definitions for target "mixxx-test" which is not
  built by this project.

```

Swiftb0y: Is there a CI run for this or how does this happen?

PPN-SD: checked with live ebuild from gentoo. But I think base/test and SHARED_LIBS (https://github.com/mixxxdj/mixxx/pull/14021#issuecomment-2630601087) should be integrated in CI.

Swiftb0y: @acolombier is it feasible for you to look into this?

PPN-SD: > how can I reproduce this?

BUILD_TESTING=OFF

acolombier: This should be fixed now. Thanks for reporting!

PR #13473, opened 2024-07-14
Let some time consuming table delegates render into cached pixmaps

JoergAtGithub:
The table delegates of the library take a lot time to paint, especially these with complex widgets, but also normal text rendering. This results in #11603.
This PR improves the execution time for BPM, Preview and the PlayCount delegates by painting into cached pixmaps instead of rendering every control again and again at each scroll iteration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -992,6 +992,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/library/tabledelegates/stardelegate.cpp
   src/library/tabledelegates/stareditor.cpp
   src/library/tabledelegates/tableitemdelegate.cpp
+  src/library/tabledelegates/textdelegate.cpp
   src/library/trackcollection.cpp
   src/library/trackcollectioniterator.cpp
   src/library/trackcollectionmanager.cpp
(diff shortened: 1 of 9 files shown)


ronso0:
It feels like this really makes the waveforms smoother when scrolling a huge cached part of the tracks table.

Did you measure how this affects the global QPixmapCache? Like, how much memory is consumed by let's say 1000 lines?
And do we need to increase the cache limit in order to not drop other cached images like cover art?

JoergAtGithub:
> 
> Did you measure how this affects the global QPixmapCache?#

I did not yet. Before I dig in such details, I want see how it performes on the different platform, especially macOS. Could you run on your Linux system and share the results.

ronso0:
Hmm I have other priorities atm. If there are no regressions on any of the platforms and with GUI scaling != 100% I'm fine with this change.
Tbh I don't care about frame drops when I'm focused on the library.

JoergAtGithub:
This is just a Draft for testing and profiling now. 
I personally find frame drops very annoying, but of course this is not a functional issue.

daschuer:
This one is conflicting. I have interest to check it out after #13719 is merged.

Issue #13471, opened 2024-07-14
wayland crash when changing waveform prefs in latest main


ywwg:
### Bug Description

Ubuntu 24.04, QT6/no QML/Open GL On.  when changing waveform preferences, the screen locks and then wayland crashes.  nothing useful in logs

### Version

HEAD

### OS

Ubuntu 24.04

ywwg:
This is a Dell XPS 13, 4.6 (Compatibility Profile) Mesa 24.0.5-1ubuntu1 (Mesa Intel(R) Xe Graphics (TGL GT2))

no issues under xorg

ywwg:
looks like gnome shell crashed:

```
2024-07-14T14:19:12.139947-04:00 glacier gnome-shell[8178]: Errors from xkbcomp are not fatal to the X server
2024-07-14T14:19:13.076200-04:00 glacier gnome-shell[4289]: JS ERROR: TypeError: this._container is undefined#012_setCustomStyle@file:///home/owen/.local/share/gnome-shell/extensions/workspace-switcher-manager@G-dH
.github.com/extension.js:428:32#012display@file:///home/owen/.local/share/gnome-shell/extensions/workspace-switcher-manager@G-dH.github.com/extension.js:343:14#012_showWorkspaceSwitcher@resource:///org/gnome/shell/
ui/windowManager.js:1840:42#012@resource:///org/gnome/shell/ui/init.js:21:20
2024-07-14T14:19:14.776151-04:00 glacier tracker-miner-f[4820]: WARNING: Available disk space (1.292136%) is below configured threshold for acceptable working (5%)
2024-07-14T14:19:14.863498-04:00 glacier gnome-shell[4289]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
2024-07-14T14:19:18.835709-04:00 glacier gnome-shell[4289]: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
2024-07-14T14:19:18.835794-04:00 glacier gnome-shell[4289]: GNOME Shell crashed with signal 11
2024-07-14T14:19:18.835809-04:00 glacier gnome-shell[4289]: == Stack trace for context 0x5ba9ee385240 ==
2024-07-14T14:19:18.835819-04:00 glacier gnome-shell[4289]: #0   5ba9ee451c98 i   resource:///org/gnome/shell/ui/init.js:21 (272755b70bf0 @ 48)
[...]

PR #13474, opened 2024-07-15
fix: support for new WaveformData struct in shaders

### LINKED ISSUE #13472: Waveform issues with "High details" enabled
### Bug Description

RGB high details waveform are showing an offset:
High details, the beat marker is not on the visual beat: 
image
No High details, the beat marker is at the expected position: 
image

Tested with QT 6.2

### Version

2.6~alpha

### OS

Ubuntu Focal

acolombier:
Fixes #13472

As part of the newly introduce `Waveform-6.0` format, the waveform data now allocates space for stem information. This PR fixes the rendering issues with shaders, by offsetting the waveform data.

Going forward, we may consider add some more advanced logic to use the stem data if available, but this will require feeding more attribute to GLSL, such as stem gain.

@daschuer @ywwg would appreciate a test if you can :)

### DIFF
--- a/res/shaders/filteredsignal.frag
+++ b/res/shaders/filteredsignal.frag
@@ -21,6 +21,11 @@ uniform sampler2D waveformDataTexture;
 
 vec4 getWaveformData(float index) {
     vec2 uv_data;
+    // The Waveform data is composed of 8 bytes per sample, (low, mid, high, all,
+    // stem1, stem2, stem3, stem4), see WaveformData struct. Currently, shaders
+    // don't support stem rendering.
+    index = 2 * index;
+
     uv_data.y = floor(index / float(textureStride));
     uv_data.x = floor(index - uv_data.y * float(textureStride));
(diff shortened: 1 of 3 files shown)


ywwg:
this does fix the offset / half-waveform issue, but it does not fix the crash when changing preferences (https://github.com/mixxxdj/mixxx/issues/13471)

ywwg (approved):

acolombier:
Thanks for the review @ywwg - sorry did that between two meetings, will grab the time to fix the comments 

Note that the Wayland issue is not related to stem nor waveform changes.

acolombier:
@ywwg I amended the commit -  feel free to merge when ready :)

ywwg:
hmm it looks like this didn't actually work right, it seems to be off by a factor of 2 by length.  This makes it seem to line up with the beat grid, but if you fast forward to the end of the track, it doesn't

PR #13475, opened 2024-07-15
fix: Key text is elided from left, should be right

danferns:
The text for Key column was being [elided to the left](https://github.com/mixxxdj/mixxx/pull/13390#issuecomment-2227330693) after #13390. This PR changes it back to the right - the default.

Screenshot after the fix:
image


Swiftb0y (approved):
LGTM. thank you for the quick fix

PR #13331, opened 2024-06-06
DEBUG 'LoadSelectedTrack' control, triggered via keyboard

ronso0:
added some warnings to maybe debug #13305 

@DJMaxergy Hopefully this helps finding the root cause of that issue.
As soon as CI builds succeeded, please get the macOS build and test (see Testing & Backup instructions https://github.com/mixxxdj/mixxx/wiki/Testing)
I think it would simplify the process if you start Mixxx from the command line and check the output right after you tried to load via keyboard. https://manual.mixxx.org/2.4/en/chapters/appendix/commandline_dev_tools
The copy the last lines with warnings and paste them here (also attach the entire log file https://github.com/mixxxdj/mixxx/wiki/Finding-the-mixxx.log-file)

PR #13481, opened 2024-07-18
(fix) avoid catching Num key modifier on macOS

### LINKED ISSUE #13305: Keyboard shortcuts for loading track from library into left/right deck not working anymore
### Bug Description

Normally I use custom keyboard shortcuts which worked without problems since 2015. Since the update to 2.4.1 today the shortcuts for loading tracks from the library to the left or right deck are not working anymore (Shift + left/right arrow). Other shortcuts for playing sampler or triggering cue points are working as expected.
I also deleted my custom shortcuts to use the default ones instead, but there these shortcuts are not working, too.
As a last step I started Mixxx from terminal with debug logs, but there were no entries concerning my keyboard shortcuts.

Steps to reproduce the problem:
- Start Mixxx, click into the library window to focus on the library.
- Choose a song by pressing the up/down key of the keyboard.
- Press Shift + left/right arrow to load a song into the left or right deck -> Shift gets ignored. By pressing the left/right arrow I can only change the focus of the song property / column of the song entry in the library window.

CPU: 1,6 GHz Dual-Core Intel Core i5
Graphics: Intel HD Graphics 6000 1536 MB

### Version

2.4.1

### OS

macOS Monterey 12.7.5

ronso0:
fixes #13305

https://github.com/mixxxdj/mixxx/commit/59b54da8c16a26234bfc56f3e16a01e865afa450 introduced a regression on macOS:
debugging in #13331 revealed that on macOS the arrow keys are sent with Num modifier, for which there would be no match in our keyseq/control hash.

Revert to checking the individual modifiers and cosntruct the QKeySequence with QStrings.

### DIFF
--- a/src/controllers/keyboard/keyboardeventfilter.cpp
+++ b/src/controllers/keyboard/keyboardeventfilter.cpp
@@ -125,23 +125,29 @@ QKeySequence KeyboardEventFilter::getKeySeq(QKeyEvent* e) {
         return {};
     }
 
+    // Note: test for individual modifiers, don't use e->modifiers() for composing
+    // the QKeySequence because on macOS arrow key events are sent with the Num
+    // modifier for some reason. This result in a key sequence for which there
+    // would be no match in our keyseq/control hash.
+    // See https://github.com/mixxxdj/mixxx/issues/13305
+    QString modseq;
+    if (e->modifiers() & Qt::ShiftModifier) {
+        modseq += "Shift+";
+    }
+    if (e->modifiers() & Qt::ControlModifier) {
+        modseq += "Ctrl+";
+    }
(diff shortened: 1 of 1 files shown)


ronso0:
Note: alternatively we could also check for
`if (macOS && Num mod && (key == Up || key == Down || key == Left || key == Right))`
and remove the Num mod in these cases.

Not sure if that's worth the effort.
___
And I'm not sure whether we should explicitly check for the `Num` mod in general. This would allow to use the numpad for shortcuts, too. But that's separate from this rather urgent bugfix.
Please someone check this and merge asap. Thanks!

m0dB:
I confirm the bug and the fix (I had no idea that shift left / shift right was supposed to load the track, very convenient indeed!).

PR #13483, opened 2024-07-18
CMakeLists: Generate hierarchical source groups for IDEs

fwcd:
This is a largely cosmetic improvement, but when generating a Visual Studio or Xcode project from Mixxx, the default is to generate a flat list of source files. For editing, however, it would be much nicer to have the same directory structure as we do in the file system. Therefore this PR uses the [`source_group(TREE ...)`](https://cmake.org/cmake/help/latest/command/source_group.html) command[^1] to mimic this layout.

For comparison, this is how it looks in Xcode:

| Before | After |
| - | - |
|  |  |

[^1]: As e.g. suggested in https://stackoverflow.com/a/56496104

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4922,3 +4922,9 @@ if(APPLE AND MACOS_BUNDLE)
   )
   install(SCRIPT "${CMAKE_CURRENT_BINARY_DIR}/BundleInstall.cmake")
 endif()
+
+# Generate hierarchical source groups when using IDE generators such as
+# Xcode or Visual Studio. Has to be done after all sources have been added.
+get_target_property(MIXXX_LIB_SOURCES mixxx-lib SOURCES)
+list(FILTER MIXXX_LIB_SOURCES INCLUDE REGEX "^src/.*") # Ignore files outside src, e.g. lib/apple
+source_group(TREE ${CMAKE_CURRENT_SOURCE_DIR}/src PREFIX "Source Files" FILES ${MIXXX_LIB_SOURCES})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4923,8 +4923,10 @@ if(APPLE AND MACOS_BUNDLE)
   install(SCRIPT "${CMAKE_CURRENT_BINARY_DIR}/BundleInstall.cmake")
 endif()
 
-# Generate hierarchical source groups when using IDE generators such as
-# Xcode or Visual Studio. Has to be done after all sources have been added.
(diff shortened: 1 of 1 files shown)


daschuer (commented):
Sounds good. Do we want to apply it to 2.4 to make it easier to change between branches (one day)?

JoergAtGithub:
In my VisualStudio it appears already structured without this PR:
grafik

fwcd:
2.4 or 2.5? Targeting main was unintentional, we should probably go with 2.5?

fwcd:
@JoergAtGithub Interesting, that's the precompiled header group though. Could you check the contents of "Source Files"?

JoergAtGithub:
> hat's the precompiled header group though

I don't think so, precompiled header is collapsed. Here it's expanded pch tree:
grafik

fwcd:
Oh right, I misread, you're right

fwcd:
Just out of curiosity, does CMake generate an additional "Source Files" group in your VS project?

JoergAtGithub:
I'm not sure, what you mean, but these are all targets listed after CMake configure completes:
grafik

fwcd:
I mean within mixxx-lib. I think VS calls groups "filters"?

fwcd:
I have rebased the PR to 2.5, so it should be ready for re-review.

JoergAtGithub:
Now I see two subtrees, `Source Files` and `src`:
grafik

fwcd:
Interesting, so it looks like either Visual Studio or CMake are implicitly adding the `src` tree. I wonder why `coreservices.cpp` et al. show up under `src` though, in Xcode they are also placed under `Source Files`...

Maybe we should just scope this logic to the Xcode generator for now since the VS project already generates a hierarchy?

daschuer:
Do you know how it relates to visual studio. Does this also improve the situation there? Are files not added to mixxx-lib still displayed reasonable?

fwcd:
Haven't had the chance to test it with Visual Studio, but according to @JoergAtGithub it made the hierarchy a bit more confusing there since VS already seems to group files hierarchically. Hence why I've scoped the logic to Xcode for now, since it definitely improves the situation there.

fwcd:
I've moved this to `main`, hopefully this should also address the new style checks

acolombier:
Any objection to merge that? This looks like a net developer experience for XCode user.

daschuer (approved):
Looks good. Can you re-base that to 2.5 to give a unified development experience on all active branches.

PR #13485, opened 2024-07-19
SlipModeShader: Set precision qualifiers for GL ES support

fwcd:
OpenGL ES requires setting these qualifiers explicitly since there is no default precision, otherwise we get these errors:

```
QOpenGLShader::compile(Fragment): ERROR: 0:12: 'float' : declaration must include a precision qualifier for type
ERROR: 0:13: 'float' : declaration must include a precision qualifier for type
ERROR: 0:14: 'float' : declaration must include a precision qualifier for type
ERROR: 0:15: 'float' : declaration must include a precision qualifier for type
```

We do this everywhere else already, so I don't think there should be any risk in declaring them explicitly here too.

### DIFF
--- a/src/shaders/slipmodeshader.cpp
+++ b/src/shaders/slipmodeshader.cpp
@@ -25,13 +25,13 @@ varying highp vec4 vposition;
 void main()
 {
 
-    float xBorder = abs(dimension.x * vposition.x);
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
LGTM! Thank you!

PR #13486, opened 2024-07-19
CmdLineArgs: Determine whether to color output based on `TERM` variable

fwcd:
To avoid cluttering up the output with ANSI escape sequences in terminals that do not support them (e.g. Xcode), this adds a heuristic that checks the `TERM` environment variable.

### DIFF
--- a/src/util/cmdlineargs.cpp
+++ b/src/util/cmdlineargs.cpp
@@ -38,7 +38,7 @@ bool calcUseColorsAuto() {
 
     // Check if terminal is known to support ANSI colors
     QString term = QProcessEnvironment::systemEnvironment().value("TERM");
-    return term == "ansi" || term == "cygwin" || term == "linux" ||
+    return term == "alacritty" || term == "ansi" || term == "cygwin" || term == "linux" ||
             term.startsWith("screen") || term.startsWith("xterm") ||
             term.startsWith("vt100") || term.startsWith("rxvt") ||
             term.endsWith("color");
--- a/src/util/cmdlineargs.cpp
+++ b/src/util/cmdlineargs.cpp
@@ -24,18 +24,24 @@ bool calcUseColorsAuto() {
     // see https://no-color.org/
     if (QProcessEnvironment::systemEnvironment().contains(QLatin1String("NO_COLOR"))) {
         return false;
-    } else {
-#ifndef __WINDOWS__
-        if (isatty(fileno(stderr))) {
-            return true;
-        }
-#else
-        if (_isatty(_fileno(stderr))) {
-            return true;
-        }
-#endif
     }
-    return false;
+
+#ifndef __WINDOWS__
+    if (!isatty(fileno(stderr))) {
+        return false;
+    }
+#else
+    if (!_isatty(_fileno(stderr))) {
+        return false;
+    }
+#endif
+
+    // Check if terminal is known to support ANSI colors
+    QString term = QProcessEnvironment::systemEnvironment().value("TERM");
+    return term == "ansi" || term == "cygwin" || term == "linux" ||
+            term.startsWith("screen") || term.startsWith("xterm") ||
+            term.startsWith("vt100") || term.startsWith("rxvt") ||
+            term.endsWith("color");
 }
 
 } // namespace


daschuer (commented):
I have the same issue when watching the console output in Eclipse. So I don't think this a compile time issue. Can we detect it at runtime?

daschuer:
Is the CMake output correct colored? 
CMake uses this: 
https://github.com/Kitware/CMake/blob/d146baa5561914f87cec6d27df9163bfa57614d9/Utilities/cmpdcurses/pdcurses/color.c
I have not jet understood to be honest.

fwcd:
CMake logs aren't shown in Xcode (at least I haven't found where). I think it's somewhat of an IDE-specific issue, but since iOS apps are rarely run outside of Apple's tools (Xcode/Console.app), I don't think it makes sense to use ANSI colors if those terminals don't support them.

daschuer:
How about the compiler output? is that colored?

daschuer:
This a also an interesting article: 
https://medium.com/@alasher/colored-c-compiler-output-with-ninja-clang-gcc-10bfe7f2b949

fwcd:
IIRC Xcode hides the raw compiler output in a submenu, so I'd have to check whether that supports ANSI colors. The runtime console is different though.

daschuer:
clang seems also be able to detect color capability: 
```
-f[no-]color-diagnostics

    This option, which defaults to on when a color-capable terminal is detected, controls whether or not Clang prints diagnostics in color.
```

fwcd:
I consider that a separate issue, this is about Mixxx' runtime logs. I agree that we might want to put more effort into detecting the Terminal though. We might be able to detect if we run in Xcode via environment variable.

On the other hand, I think defaulting to non-colored output on iOS is probably safer since apps have little control over how and where the OS aggregates those logs.

daschuer:
clang uses this function: 
https://github.com/llvm/llvm-project/blob/2b371003d101ccc50081c08c7ed5fd7dcbfdef71/llvm/include/llvm/Support/Process.h#L178

Which boils down to: 
https://github.com/llvm/llvm-project/blob/2b371003d101ccc50081c08c7ed5fd7dcbfdef71/llvm/lib/Support/Unix/Process.inc#L344
and 
https://github.com/llvm/llvm-project/blob/2b371003d101ccc50081c08c7ed5fd7dcbfdef71/llvm/lib/Support/Windows/Process.inc#L289

daschuer:
QT has also a private function for it: 
https://github.com/qt/qtquick3d-assimp/blob/5d5e0bad3c4033124f0f227bf6120c6f9b17ede4/contrib/googletest/googletest/src/gtest.cc#L3238

fwcd:
Ah interesting, checking the `TERM` variable seems like a good option on Unix

fwcd:
I have updated the PR to determine whether a terminal supports colors based on the `TERM` variable. The implementation is effectively an adapted version of the LLVM check.

The PR also targets 2.5 now, since this is a rather minor fix that might be useful there too.

fwcd:
The sporadic macOS CI failures (e.g. [this one](https://github.com/mixxxdj/mixxx/actions/runs/10029992787/job/27718835538?pr=13486)) are a known issue btw: https://github.com/actions/runner-images/issues/7522

Re-running the workflow usually fixes it, but of course that isn't ideal. For my m1xxx build I use a workaround which kills the XProtect process that supposedly interferes with `cpack`'s DMG creation and retry the step a few times as a fallback, but of course that is still far from being a clean solution: https://github.com/fwcd/m1xxx/blob/36fd9a2a352a7a1d81435de5863dc34af6f2ac27/.github/workflows/build.yml#L221-L240

src/util/cmdlineargs.cpp
```diff
@@ -24,18 +24,24 @@ bool calcUseColorsAuto() {
...
     }
-    return false;
+#endif
+
+    // Check if terminal is known to support ANSI colors
+    QString term = QProcessEnvironment::systemEnvironment().value("TERM");
+    return term == "ansi" || term == "cygwin" || term == "linux" ||
+            term.startsWith("screen") || term.startsWith("xterm") ||
+            term.startsWith("vt100") || term.startsWith("rxvt") ||
+            term.endsWith("color");
```
Holzhaus, line 44: please also add `alacritty` and `xterm-256color`.

fwcd: `xterm-256color` is already handled by the last condition. Is `alacritty` the full string or is it suffixed with `-...color` too?

Holzhaus: It's the full string:

```
$ echo $TERM
alacritty
$
```

Holzhaus: on second thought, let's invoke `tput colors` to check for color support instead. If the command is not available, we just don't use colors.

fwcd: Yeah, that's probably a cleaner solution indeed. Only downside is that we'd have to spawn a process at every launch.

Holzhaus: I think we can live with that :)

Maybe something like this (untested):

```diff
diff --git a/src/util/cmdlineargs.cpp b/src/util/cmdlineargs.cpp
index f7b16684f0..356e9a0be3 100644
--- a/src/util/cmdlineargs.cpp
+++ b/src/util/cmdlineargs.cpp
@@ -9,6 +9,7 @@

 #include <QCommandLineParser>
 #include <QCoreApplication>
+#include <QProcess>
 #include <QProcessEnvironment>
 #include <QStandardPaths>
 #include <QtGlobal>
@@ -36,12 +37,32 @@ bool calcUseColorsAuto() {
     }
 #endif

-    // Check if terminal is known to support ANSI colors
-    QString term = QProcessEnvironment::systemEnvironment().value("TERM");
-    return term == "ansi" || term == "cygwin" || term == "linux" ||
-            term.startsWith("screen") || term.startsWith("xterm") ||
-            term.startsWith("vt100") || term.startsWith("rxvt") ||
-            term.endsWith("color");
+    // Check if terminal is support ANSI colors according to terminfo database
+    // using tput. If the command fails, don't use colors.
+    constexpr int kTimeoutMillis = 500;
+    QProcess tput;
+    tput.start("tput", {"colors"});
+    if (!tput.waitForStarted(kTimeoutMillis)) {
+        return false;
+    }
+
+    if (!tput.waitForFinished(kTimeoutMillis)) {
+        return false;
+    }
+
+    if (tput.exitStatus() != QProcess::NormalExit) {
+        return false;
+    }
+
+    if (tput.exitCode() != 0) {
+        return false;
+    }
+
+    if (tput.readAllStandardOutput().trimmed().compare("256") != 0) {
+        return false;
+    }
+
+    return true;
 }

 } // namespace
```

fwcd: We should check for platforms where we can't spawn processes (Wasm/iOS) too, but otherwise that looks good I'd say. I'll have a look at integrating it later

fwcd: Also I'm not sure if this would work on Windows (though the `TERM` heuristic probably wouldn't either), @JoergAtGithub do you by chance know if there's `tput` there or if there is another way of checking for colors?

JoergAtGithub: Sorry, I'm unfamilar with this topic

Holzhaus: On Windows I think we need to call [`GetConsoleMode()`](https://learn.microsoft.com/en-us/windows/console/getconsolemode) and check if the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag is set.

Holzhaus: This could be useful:  https://github.com/build2/build2/issues/312#issuecomment-1693263055

daschuer:
I am reluctant of using tput. tput is part of ncurses, and just a thin wrapper around a library call. The process call at the startup is kind of fragile and requires a good portion of error handling. 

Instead let's follow CMake and use PDCurses that is for my understanding a wrapper around ncurses but works also for other targets. https://github.com/wmcbrine/PDCurses

https://github.com/Kitware/CMake/blob/d146baa5561914f87cec6d27df9163bfa57614d9/Utilities/cmpdcurses/pdcurses/color.c

daschuer:
https://pdcurses.org/docs/MANUAL.html#portability-8
has_colors() is available in ncurses.h and curses.h so we may use https://github.com/microsoft/vcpkg/tree/master/ports/pdcurses for windows and ncurses for the rest. 
Like here: 
https://github.com/Coul33t/c-roguelike/commit/f8c5c8ed13816ea1e736f5bab372c7d851afe5f4

fwcd:
Both pdcurses and ncurses require adding a dependency. `tput` requires spawning a process. While the custom `TERM` check is not particularly elegant, I'm not sure if it's worth going through the trouble over something as minor as this to support a tiny fraction of use cases that could easily be amended. The worst case of the current implementation is that we don't use colors on a terminal that might support it, but is that really an issue?

IMO we could look into something in the future, but I'd love to move my iOS branch closer to upstream to avoid having to patch around these minor things.

Holzhaus (approved):
I'm okay with merging this in the current state. I'm also weary of adding new dependencies.

For simplicity, we could consider opting for something like this in the future: http://bixense.com/clicolors/ so that users have a way to force ANSI colors in case the autodetection is not working.

daschuer (approved):
I am also fine with the current state. Thank you.

PR #13489, opened 2024-07-19
Overview waveform: draw minute markers on top of played overlay

ronso0:
I thought "why did the minute markers vanish?" just to realize they're covered by the 'played' overlay.
Let's move them above.


JoergAtGithub:
Looks better now, with the fully saturated minute markers! Thank you!

PR #13493, opened 2024-07-21
Library: Automatically relink paths when iOS sandbox prefix moves

fwcd:
On iOS, apps are generally restricted to their sandbox, which is located in the file system at:

```
/private/var/mobile/Containers/Data/Application/<uuid>
```

Unfortunately, this UUID changes after reinstalling or updating the app, so we cannot rely on this prefix being stable (though everything inside the sandbox is persisted, so relative paths stay intact). The general recommendation is to [never store absolute paths into the sandbox](https://aplus.rs/2013/never-save-absolute-file-paths-in-your-ios-app/), but doing so in Mixxx would be impractical, since especially on desktop we want the flexibility to add arbitrary music directories, for example.

Thus this PR implements a slight workaround that queries the current sandbox prefix on startup and automatically relinks music directories and the recordings directory if it has changed. While we're still dealing with absolute paths, this seems like the best compromise that requires few changes and should "just work" in pretty much all standard cases[^1].

[^1]: The only situation where this logic may break down is in edge cases, e.g. when the user adds some path outside our sandbox directory. This is either forbidden by iOS (unless jailbroken) or not well supported by us (when a music directory points to another app's sandbox. This is possible if such a directory was added via the file picker, such directories would, however, be inaccessible on the next launch anyway, unless we implement proper URL bookmarking). In any case, this should be better than the status quo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1190,6 +1190,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/util/safelywritablefile.cpp
   src/util/sample.cpp
   src/util/sandbox.cpp
+  src/util/sandboxios.cpp
   src/util/screensaver.cpp
   src/util/screensavermanager.cpp
   src/util/semanticversion.cpp
--- a/src/library/dao/directorydao.cpp
+++ b/src/library/dao/directorydao.cpp
@@ -217,8 +217,10 @@ std::pair<DirectoryDAO::RelocateResult, QList<RelocatedTrack>> DirectoryDAO::rel
     // Appending '/' is required to disambiguate files from parent
     // directories, e.g. "a/b.mp3" and "a/b/c.mp3" where "a/b" would
     // match both instead of only files in the parent directory "a/b/".
-    DEBUG_ASSERT(!oldDirectory.endsWith('/'));
-    const QString oldDirectoryPrefix = oldDirectory + '/';
+    QString oldDirectoryPrefix = oldDirectory;
+    if (!oldDirectory.endsWith('/')) {
+        oldDirectoryPrefix += '/';
+    }
     query.prepare(QStringLiteral(
             "SELECT library.id,track_locations.id,track_locations.location "
             "FROM library INNER JOIN track_locations ON "
--- a/src/library/library.cpp
+++ b/src/library/library.cpp
@@ -703,7 +703,6 @@ void Library::requestRelocateIOSSandboxDirs() {
 
     // TODO: Do we need to handle external track collections?
 
-    bool needsRescan = false;
     QStringList rootDirs = m_pTrackCollectionManager->internalCollection()->getRootDirStrings();
 
     for (const QString& dir : rootDirs) {
@@ -722,13 +721,6 @@ void Library::requestRelocateIOSSandboxDirs() {
             qWarning() << "Could not relink music directory after iOS sandbox moved";
             continue;
         }
-
-        needsRescan = true;
-    }
-
-    if (needsRescan) {
-        qInfo() << "Rescanning library since iOS sandbox moved";
-        m_pTrackCollectionManager->startLibraryScan();
     }
 }
 #endif
--- a/src/library/library.cpp
+++ b/src/library/library.cpp
@@ -3,6 +3,7 @@
 #include <QApplication>
 #include <QDir>
 #include <QMessageBox>
+#include <QRegularExpression>
 
 #include "control/controlobject.h"
 #include "controllers/keyboard/keyboardeventfilter.h"
@@ -691,6 +692,64 @@ bool Library::requestRelocateDir(const QString& oldDir, const QString& newDir) {
     }
     return false;
 }
+
+#ifdef Q_OS_IOS
+void Library::requestRelocateiOSSandboxDirs() {
+    // Get current sandbox directory
+    QString newSandboxDir = QDir::homePath();
+
+    static const QRegularExpression sandboxDirRegex = QRegularExpression(
+            QStringLiteral("^/private/var/mobile/Containers/Data/Application/"
+                           "[a-zA-Z0-9\\-]+(/.*)"));
+
+    // TODO: If the user selects a different app sandbox than ours (which is
+    // possible via the file picker) relinking will point those directories to
+    // our sandbox. This is not a supported scenario, however, and we should probably
+    // guard against picking directories outside the Mixxx sandbox when adding
+    // new directories.
+
+    // Patch old sandbox directories
+    bool needsRescan = false;
+    QStringList rootDirs = m_pTrackCollectionManager->internalCollection()->getRootDirStrings();
+
+    for (const QString& dir : rootDirs) {
+        QRegularExpressionMatch match = sandboxDirRegex.match(dir);
+        if (!match.hasMatch()) {
+            qWarning() << "Leaving directory outside the iOS sandbox untouched:"
+                       << dir
+                       << "(This is unusual, since iOS generally does not "
+                          "allow accessing such paths, perhaps the regex in "
+                          "Library::requestRelocateiOSSandboxDirs has to be "
+                          "updated?)";
+            continue;
+        }
+
+        QString relativePath = match.captured(1);
+        QString newDir = newSandboxDir + relativePath;
+
+        if (dir == newDir) {
+            // Sandbox directory did not move
+            continue;
+        }
+
+        qInfo() << "Relinking music directory since iOS sandbox moved:"
+                << dir << "->" << newDir;
+
+        auto result = m_pTrackCollectionManager->relocateDirectory(dir, newDir);
+        if (result != DirectoryDAO::RelocateResult::Ok) {
+            qWarning() << "Could not relink music directory after iOS sandbox moved";
+            continue;
+        }
+
+        needsRescan = true;
+    }
+
+    if (needsRescan) {
+        qInfo() << "Rescanning library since iOS sandbox moved";
+        m_pTrackCollectionManager->startLibraryScan();
+    }
+}
+#endif
 
 void Library::setFont(const QFont& font) {
     QFontMetrics currMetrics(m_trackTableFont);
--- a/src/library/library.cpp
+++ b/src/library/library.cpp
@@ -3,7 +3,6 @@
 #include <QApplication>
 #include <QDir>
 #include <QMessageBox>
-#include <QRegularExpression>
 
 #include "control/controlobject.h"
 #include "controllers/keyboard/keyboardeventfilter.h"
@@ -37,6 +36,7 @@
 #include "util/assert.h"
 #include "util/logger.h"
 #include "util/sandbox.h"
+#include "util/sandboxios.h"
 #include "widget/wlibrary.h"
 #include "widget/wlibrarysidebar.h"
 #include "widget/wsearchlineedit.h"
@@ -695,37 +695,19 @@ bool Library::requestRelocateDir(const QString& oldDir, const QString& newDir) {
 
 #ifdef Q_OS_IOS
 void Library::requestRelocateiOSSandboxDirs() {
-    // Get current sandbox directory
-    QString newSandboxDir = QDir::homePath();
-
-    static const QRegularExpression sandboxDirRegex = QRegularExpression(
-            QStringLiteral("^/private/var/mobile/Containers/Data/Application/"
-                           "[a-zA-Z0-9\\-]+(/.*)"));
-
     // TODO: If the user selects a different app sandbox than ours (which is
     // possible via the file picker) relinking will point those directories to
     // our sandbox. This is not a supported scenario, however, and we should probably
     // guard against picking directories outside the Mixxx sandbox when adding
     // new directories.
 
-    // Patch old sandbox directories
+    // TODO: Do we need to handle external track collections?
+
     bool needsRescan = false;
     QStringList rootDirs = m_pTrackCollectionManager->internalCollection()->getRootDirStrings();
 
     for (const QString& dir : rootDirs) {
-        QRegularExpressionMatch match = sandboxDirRegex.match(dir);
-        if (!match.hasMatch()) {
-            qWarning() << "Leaving directory outside the iOS sandbox untouched:"
-                       << dir
-                       << "(This is unusual, since iOS generally does not "
-                          "allow accessing such paths, perhaps the regex in "
-                          "Library::requestRelocateiOSSandboxDirs has to be "
-                          "updated?)";
-            continue;
-        }
-
-        QString relativePath = match.captured(1);
-        QString newDir = newSandboxDir + relativePath;
+        QString newDir = mixxx::updateiOSSandboxPath(dir);
 
         if (dir == newDir) {
             // Sandbox directory did not move
--- a/src/library/library.cpp
+++ b/src/library/library.cpp
@@ -694,7 +694,7 @@ bool Library::requestRelocateDir(const QString& oldDir, const QString& newDir) {
 }
 
 #ifdef Q_OS_IOS
-void Library::requestRelocateiOSSandboxDirs() {
+void Library::requestRelocateIOSSandboxDirs() {
     // TODO: If the user selects a different app sandbox than ours (which is
     // possible via the file picker) relinking will point those directories to
     // our sandbox. This is not a supported scenario, however, and we should probably
@@ -707,7 +707,7 @@ void Library::requestRelocateiOSSandboxDirs() {
     QStringList rootDirs = m_pTrackCollectionManager->internalCollection()->getRootDirStrings();
 
     for (const QString& dir : rootDirs) {
-        QString newDir = mixxx::updateiOSSandboxPath(dir);
+        QString newDir = mixxx::updateIOSSandboxPath(dir);
 
         if (dir == newDir) {
             // Sandbox directory did not move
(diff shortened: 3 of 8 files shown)


src/library/library.cpp
```diff
@@ -692,6 +693,46 @@ bool Library::requestRelocateDir(const QString& oldDir, const QString& newDir) {
...
+        qInfo() << "Relinking music directory since iOS sandbox moved:"
+                << dir << "->" << newDir;
+
+        auto result = m_pTrackCollectionManager->relocateDirectory(dir, newDir);
+        if (result != DirectoryDAO::RelocateResult::Ok) {
+            qWarning() << "Could not relink music directory after iOS sandbox moved";
+            continue;
+        }
+
+        needsRescan = true;
```
ronso0, line 726: Did you verify a rescan is indeed required after relinking?
Various tests I ran showed it was not.
See also https://github.com/mixxxdj/mixxx/issues/12858#issuecomment-1961193790

fwcd: I think it was indeed required to make it work, if I remember right, but I'll check just to make sure I didn't miss anything (Though rescanning is probably still a good idea, the sandbox doesn't move that often, usually only when the app is updated, so I think we can afford to code a bit defensively here)

fwcd: Just tested it on a device and can confirm that a rescan is indeed needed, without it dragging a track from the library to the decks won't do anything.

src/util/sandboxios.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace mixxx {
+
+#ifdef Q_OS_IOS
+
+static const QRegularExpression sandboxPrefixRegex = QRegularExpression(
+        QStringLiteral("^/private/var/mobile/Containers/Data/Application/"
+                       "[a-zA-Z0-9\\-]+(/.*)"));
+
+QString updateiOSSandboxPath(const QString& path) {
```
daschuer, line 17: It is hard to read when using camel case and trade mark case in one name. 
```suggestion
QString updateIosSandboxPath(const QString& path) {
```

src/library/library.cpp
daschuer: Mm do you know why? Maybe there is just a refresh of the library page required?

fwcd: What do you mean by that? This happens after reinstalling Mixxx, so the whole app is already relaunched.

src/util/sandboxios.cpp
fwcd: Personally I've found the original casing easier to parse than a differently cased version, but I'd be fine either way. If we do that, we should probably do it consistently for all methods in this PR

src/library/library.cpp
ronso0: I think what @daschuer suspects is that the track view still shows old data.
Not sure if/how thats possible, but you may verify by switching to another library feature and back to **Tracks** for example to trigger a select(), then try again to load a track.

Of course, if load is still not possible after a second relaunch, the rescan is indeed required (still don't understand why..)

fwcd: Switching the view unfortunately doesn't change this. I did a quick database dump before and after rescanning (after reinstall + relink) and it looks like the relink doesn't update the track locations:

image

The only entry in the pre-rescan/post-relink db that's updated to the new prefix is the `directories` entry:

```sql
INSERT INTO directories VALUES('/private/var/mobile/Containers/Data/Application/928F499E-48C7-43BB-B3E6-8D585E6FA011/Documents/');
```

Haven't dug into this more deeply, but perhaps there's some step in relinking that my implementation in this PR misses?

fwcd: What makes it more curious is that my branch does contain #12462, so I'm not sure why it's not updating the tracks

src/util/sandboxios.cpp
fwcd: I have updated the casing to `updateIOSSandboxPath` here and in the other methods, to stay consistent with the spelling we use for the iTunes importers (e.g. #12690). Does that work for you?

src/library/library.cpp
ronso0: Very strange. You should see this being logged
`Relocated tracks: [num of tracks]`
otherwise it fails silently earlier?

fwcd: Couldn't find that line in the log (after reinstalling), but found something else that looks suspicious:

```
warning [Main] Library - Skipping access check for missing or invalid directory QFileInfo(/private/var/mobile/Containers/Data/Application/FD476962-4D00-48DE-8F20-ABF7E49791DA/Documents/)
critical [Main] DEBUG ASSERT: "!oldDirectory.endsWith('/')" in function std::pair<DirectoryDAO::RelocateResult, QList<RelocatedTrack>> DirectoryDAO::relocateDirectory(const QString &, const QString &) const at src/library/dao/directorydao.cpp:220
```

Notably this is the old sandbox prefix. Maybe Mixxx doesn't like to relink something that no longer exists in the fs?

ronso0: Yeah, that's probably it! (the debug assert)
> Maybe Mixxx doesn't like to relink something that no longer exists in the fs?

well, that's the entire point of relocating, no ; )

(oh, you build with DEBUG_ASSERTIONS_FATAL=OFF )

ronso0: Does this log line already have the trailing / ?
`Relocating directory in internal track collection: oldDir -> newDir`

src/util/sandboxios.cpp
daschuer: That works nicely as well.

src/library/library.cpp
fwcd: Yes:

```
debug [Main] TrackCollectionManager - Relocating directory in internal track collection: "/private/var/mobile/Containers/Data/Application/3892CA05-7DC8-4DC1-B535-BB6CF05744A3/Documents/" -> "/private/var/mobile/Containers/Data/Application/C5C1DF42-1036-4328-B1AF-0B96889A02F0/Documents/"
```

For reference, here's a full log:

[mixxx-ios.log](https://github.com/user-attachments/files/16652277/mixxx-ios.log)

fwcd: So it looks like we already store the directory with a trailing slash. Patching the check to accept an existing trailing slash fixes the debug assert, but for some reason the track locations are still not updated properly in the db (there are probably other assumptions going on, haven't dug in that deeply):

```diff
diff --git a/src/library/dao/directorydao.cpp b/src/library/dao/directorydao.cpp
index c1022eb47e..e815498960 100644
--- a/src/library/dao/directorydao.cpp
+++ b/src/library/dao/directorydao.cpp
@@ -217,8 +217,10 @@ std::pair<DirectoryDAO::RelocateResult, QList<RelocatedTrac
k>> DirectoryDAO::rel
     // Appending '/' is required to disambiguate files from parent
     // directories, e.g. "a/b.mp3" and "a/b/c.mp3" where "a/b" would
     // match both instead of only files in the parent directory "a/b/".
-    DEBUG_ASSERT(!oldDirectory.endsWith('/'));
-    const QString oldDirectoryPrefix = oldDirectory + '/';
+    QString oldDirectoryPrefix = oldDirectory;
+    if (!oldDirectory.endsWith('/')) {
+        oldDirectoryPrefix += '/';
+    }
     query.prepare(QStringLiteral(
             "SELECT library.id,track_locations.id,track_locations.location "
             "FROM library INNER JOIN track_locations ON "
```

fwcd: I have updated the directory DAO to accept an existing trailing slash (28ed67bcf8) and now relinking seems to work properly without rescanning 🎉

ronso0 (changes requested):

src/library/dao/directorydao.cpp
```diff
@@ -217,8 +217,10 @@ std::pair<DirectoryDAO::RelocateResult, QList<RelocatedTrack>> DirectoryDAO::rel
     // Appending '/' is required to disambiguate files from parent
     // directories, e.g. "a/b.mp3" and "a/b/c.mp3" where "a/b" would
     // match both instead of only files in the parent directory "a/b/".
-    DEBUG_ASSERT(!oldDirectory.endsWith('/'));
-    const QString oldDirectoryPrefix = oldDirectory + '/';
+    QString oldDirectoryPrefix = oldDirectory;
+    if (!oldDirectory.endsWith('/')) {
+        oldDirectoryPrefix += '/';
+    }
```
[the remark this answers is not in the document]
fwcd: I think the trailing slash came up much earlier, the directory stored in the db seems to have it. Relinking seems to be the only place where this came up.

If we don't want trailing slashes, we should probably assert this when adding the directory? (and maybe here too as a sanity check)

Fwiw, double slashes are fine in Unix and simply ignored.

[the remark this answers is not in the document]
fwcd: Off top of my head, I think we must pass the old directory verbatim as it is stored in the db (including any trailing slash), otherwise the query won't update it, so we'd have to update this elsewhere one way or the other.

Not sure how double slashes could come up here, if we check for the trailing slash before adding it, then the only scenario where this would come up is if we would already pass in such a string. The debug assert only prevents this for the trailing case and doesn't for check intermediate //s (though since Unix is fine with them, it probably wouldn't have been an issue either way)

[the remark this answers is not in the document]
ronso0: AFAICT we had no issues with trailing / when relocating (after the latest reloc fixes), which indicates that assert is fine.

IMHO we should rather check where trailing / comes from in the first place (fwiw I don't have them in my `directories` table) and/or make otherwise sure we don't have // 

Do you see trailing slashes also in win/macOS/linux databases / preferences?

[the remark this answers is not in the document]
fwcd: On macOS/Linux I don't have any trailing slashes in my db either. I think it's either the iOS file chooser or Qt's home path (which effectively delegates to the iOS API). Just stripping any trailing slashes when adding the directory might be the best solution? I'd be hesitant to do it in the iOS-specific code, since we'd still bake the assumption that other platforms don't return any trailing slashes (which might hold, but feels a bit fragile).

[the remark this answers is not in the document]
fwcd: Ah, hm, we don't guard against trailing slashes in the new directory either, during relocation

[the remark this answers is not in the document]
ronso0: True.
I think we should fix that in 2.4. For example add something like
`while (newDirString.endsWith('/') { newDirStr.chop(1); }`

FWIW I also just figured that
* `QFileInfo::absoluteFilePath()` may return a path **with** a trailing slash if the string for used for construction had one.
used by [`mixxx::FileInfo::location()`](https://github.com/mixxxdj/mixxx/blob/17edb4e5f0554a2b3a0c5c5952d4468209b1ca87/src/util/fileinfo.h#L115-L117) which is used by `DirectoryDAO::addDirectory()`, and `DirectoryDAO::relocateDirectory()` for verifying the QString newDir
* otoh, `QFileInfo::absolutePath()` returns the directory part only (**no trailing slash**)
used by `mixxx::FileInfo::locationPath()`
(just verified on Linux with Qt 6.2.3)

Another finding:
[QFileInfo::absoluteFilePath](https://doc.qt.io/qt-6/qfileinfo.html#absoluteFilePath) will not have trailing slashes, if... ([Qt 6 documentaion](https://doc.qt.io/qt-6/qfileinfo.html#details)):
> QFileInfo objects referring to directories created by Qt classes will not have a trailing directory separator '/'.

but we just pass the _strings_ from the db around :\

___
About the origin of trailing / in music dirs:
adding a new dir works like this:
1. DlgPrefLibrary::slotAddDir() gets a **QString** from QFileDialog::getExistingDirectory()
2. Library::requestAddDir() creates a **mixxx::FileInfo** from that
3. TrackCollectionManager::addDirectory()  -> TrackCollection::addDirectory
4. DirectoryDAO::addDirectory() turns this into a **QString** with `mixxx::FileInfo::canonicalLocation()`
(which calls `QFileInfo::canonicalFilePath()`)
which does **not** have a trailing / even if I append one manually in DlgPrefLibrary::slotAddDir()
(also tested with Qt 6.2.3 on Linux)

Conclusion: the trailing slash may only stem from relocating?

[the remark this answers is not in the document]
fwcd: Tbh is it really that much of an issue if we store directories which may or may not have a trailing slash? I feel like documenting and upholding this invariant everywhere (across whichever platform APIs we encounter) is probably going to be fragile, one way or another.

Trailing slashes seems to work fine for almost everything else and if we really need a path with (or without) the slash, then adding or removing it should be easy where needed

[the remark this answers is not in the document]
fwcd: But yeah, if it only stems from relocating, maybe we could just call `QFileInfo::absolutePath` or `mixxx::FileInfo::canonicalPath` on the new directory to strip it?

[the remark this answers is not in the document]
ronso0: Well, I'm really glad relocating seems to work now, even cross-OS.
(failed (cross-OS ) relocation means loosing the library, and debugging that (remotely in the forums) is PITA)

Therefore I'd really love to play it safe and not risk any regressions, i.e. prevent any trailing / or //.
(didn't check what might happen if we for example relocate to Windows from iOS which already did sandbox relocation (adds /))
The only place where we need / is for finding the old track(location)s.

So, yes, this should do the trick (IIUC :zany_face: )
> maybe we could just call QFileInfo::absolutePath or mixxx::FileInfo::canonicalPath on the new directory to strip it?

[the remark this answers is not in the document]
fwcd: Hm, unfortunately relocation does not seem to be the only place where this could come up, adding a directory already produces a trailing slash...

https://github.com/user-attachments/assets/e57f3ced-1b37-4f44-be2b-4b78116321ee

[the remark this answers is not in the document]
ronso0: Indeed. Somehow I screwed up the test 
> 1. DlgPrefLibrary::slotAddDir() gets a **QString** from QFileDialog::getExistingDirectory()
> 2. Library::requestAddDir() creates a **mixxx::FileInfo** from that
> 3. TrackCollectionManager::addDirectory()  -> TrackCollection::addDirectory
> 4. DirectoryDAO::addDirectory() turns this into a **QString** with `mixxx::FileInfo::canonicalLocation()`
> (which calls `QFileInfo::canonicalFilePath()`)
> which does **not** have a trailing / even if I append one manually in DlgPrefLibrary::slotAddDir()
> (also tested with Qt 6.2.3 on Linux)

for **comparisson with existing directories**.
But for actually **adding the directory** it uses `mixxx::FileInfo::location()` which **can have** a trailing / 

So, finally, my conclusion:
* add directory: since `QFileDialog::getExistingDirectory()` may return a path with /, let's use `newCanonicalLocation` here
https://github.com/mixxxdj/mixxx/blob/f4785f4720bc244fe51a0fdd1a1e9396d9b1e4c8/src/library/dao/directorydao.cpp#L125-L127
* relocate dir: let's make _somehow_¹ sure that `oldDirectoryPrefix` ends with **one /**
https://github.com/mixxxdj/mixxx/blob/f4785f4720bc244fe51a0fdd1a1e9396d9b1e4c8/src/library/dao/directorydao.cpp#L221
personally I don't mind whether that is a smart regex check/replacement or QString operation.

What do you think?

___
Ideally we'd have no / in the `directories` table.
Shall we add a cleanup method that is called each time someone calls `DirectoryDAO::getRootDirStrings()` or `DirectoryDAO::loadAllDirectories()`?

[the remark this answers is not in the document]
fwcd: What's the status on this? I'd be in favor of keeping the sanity check, that seems less brittle with different file system APIs at play IMO and I don't see how that check would potentially corrupt a library (if there's no trailing slash, that means the assumption was already broken elsewhere and this will just crash Mixxx with debug assertions enabled), but I'd be open to solving this in some other way too, if there's not too much debugging involved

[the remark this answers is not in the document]
ronso0: Umm, it's been a while, have recap first what we figured out.

src/util/sandboxios.cpp
```diff
@@ -0,0 +1,35 @@
...
+                   << path
+                   << "Perhaps the regex in util/sandboxios.cpp needs to be "
+                      "updated?";
+        return path;
+    }
+
+    QString newSandboxPrefix = QDir::homePath();
+    QString relativePath = match.captured(1);
+
+    return newSandboxPrefix + relativePath;
```
acolombier, line 30: Could we use `[QDir::filePath](https://doc.qt.io/qt-6/qdir.html#filePath)` here?
```suggestion
    return QDir::home().filePath(match.captured(1));
```

src/util/sandboxios.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <QString>
+#include <QtGlobal>
+
+namespace mixxx {
+
+#ifdef Q_OS_IOS
```
acolombier, line 8: Should we remove this macro and move it at include level?

CMakeLists.txt
```diff
@@ -1190,6 +1190,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/util/safelywritablefile.cpp
   src/util/sample.cpp
   src/util/sandbox.cpp
+  src/util/sandboxios.cpp
```
acolombier, line 1193: Can we move this src file to conditional list so it is only added to the list on iOS platform?

PR #13494, opened 2024-07-21
BaseTrackTableModel: Disable inline track editing on iOS

fwcd:
On iOS tapping a track in the library triggers the on-screen keyboard, even if the track cannot actually be edited. Until we figure out a better user experience here, we should just disable inline editing there. Of course, tracks can still be edited via the `Cmd+Enter` shortcut or by double tapping title/artist on a loaded deck.

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -2,6 +2,7 @@
 
 #include <QGuiApplication>
 #include <QScreen>
+#include <QtGlobal>
 
 #include "library/coverartcache.h"
 #include "library/dao/trackschema.h"
@@ -1110,6 +1111,12 @@ Qt::ItemFlags BaseTrackTableModel::readWriteFlags(
         // Cells are editable by default
         itemFlags |= Qt::ItemIsEditable;
     }
+#ifdef Q_OS_IOS
+    // Make items non-editable on iOS by default, since tapping any track will
+    // otherwise trigger the on-screen keyboard (even if they cannot actually
+    // be edited).
+    itemFlags &= ~Qt::ItemIsEditable;
+#endif
     return itemFlags;
 }
 
--- a/src/library/browse/browsetablemodel.cpp
+++ b/src/library/browse/browsetablemodel.cpp
@@ -403,6 +403,12 @@ Qt::ItemFlags BrowseTableModel::flags(const QModelIndex& index) const {
 
     int column = index.column();
 
+#ifdef Q_OS_IOS
+    // Make items non-editable on iOS by default, since tapping any track will
+    // otherwise trigger the on-screen keyboard (even if they cannot actually
+    // be edited).
+    return defaultFlags;
+#else
     switch (column) {
     case COLUMN_FILENAME:
     case COLUMN_BITRATE:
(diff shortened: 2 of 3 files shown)


daschuer (approved):
LGTM, thank you

ronso0:
> On iOS tapping a track in the library triggers the on-screen keyboard

IIUC this is the case on any touch (OS) device, like RPi with on-screen keyboard.
Is the issue actually that the keyboard may cover the inline editor?

> even if the track cannot actually be edited.

Which situation are you referring to?
Yes, editing in Browse/Computer is pointless unless the selected track has already been added to the library.

fwcd:
>  Is the issue actually that the keyboard may cover the inline editor?

The UI dodges the keyboard correctly, that works fine.

> Which situation are you referring to?
> Yes, editing in Browse/Computer is pointless unless the selected track has already been added to the library.

Tapping a track in the regular library/tracks view. Pressing keys doesn't actually do anything useful (it doesn't edit the field) and instead jumps around in the library. Additionally, the keyboard covers a large part of the UI, so it gets in the way more often that it's useful.

Here's a quick demo:

https://github.com/user-attachments/assets/674832f0-e5d2-4ed9-9168-869ceb29e074

ronso0:
I understand. For consistency let's also disable the check box in the preferences.
Alternatively, just unset `EditMetadataSelectedClick` on iOS and disable the check box?

ronso0:
The editor isn't opened at all, or is it closed right away when the keyboard pops up??

fwcd:
> The editor isn't opened at all

From what I can tell, the editor doesn't seem to open at all

ronso0:
> From what I can tell, the editor doesn't seem to open at all

You _could_ verify by adding some logging to `MultiLineEditDelegate`

Note that I don't mind, the current state is okay for me (+ disbale the checkbox), it might be helpful for the future long-press implementation.

fwcd:
Interestingly, the checkbox was already disabled. Enabling it causes the editor to show, but the keyboard always pops up, regardless of whether editing is enabled or not.

ronso0:
> Interestingly, the checkbox was already disabled. 

By "disable" I meant not untick it but deactivate it (greyed out) to avoid confusion with an setting that's not applied on iOS.
```
checkBox_edit_metadata_selected_clicked->setEnabled(false);
checkBox_edit_metadata_selected_clicked->setChecked(false);
```
somewhere in the ctor of DlgPrefLibrary.

fwcd:
I've disabled it, I don't think we would need to uncheck it explicitly since the default is already `false`. Should we do that anyway just to be sure? In that case we should probably also make sure not to update it elsewhere (on iOS)?

ronso0:
I don't mind. Would be consistent to untick it, but Idk if it's worth the trouble/work :shrug:

ronso0:
Disabling editing would still allow toggling the BPM and Played checkboxes, right?
I.e. Qt::ItemIsEditable doesn't interfere with Qt::ItemIsUserCheckable?

Btw what about the star rating? That's an editor created on hover.
How/does that work on iOS / touchscreens in general?

ronso0 (approved):

fwcd:
Editing ratings and the BPM checkbox still works as expected (i.e. touching/dragging will change it).

ronso0:
Okay, thanks!

PR #13497, opened 2024-07-22
Add Key Color Palettes

danferns:
This PR adds color palette support for the Key Colors that were added in #13390. 

New third party key palettes:
- Traktor
- Mixed In Key

New accessible key palettes:
- Protanopia / Protanomaly
- Deuteranopia / Deuteranomaly
- Tritanopia / Tritanomaly

### Screenshots

image

(Tritanopia Palette)

image

### DIFF
--- a/src/preferences/dialog/dlgprefcolors.cpp
+++ b/src/preferences/dialog/dlgprefcolors.cpp
@@ -96,19 +96,21 @@ void DlgPrefColors::slotUpdate() {
                     BaseTrackTableModel::kKeyColorsEnabledDefault));
     for (const auto& palette : std::as_const(mixxx::PredefinedColorPalettes::kPalettes)) {
         QString paletteName = palette.getName();
+        QString translatedName = QCoreApplication::translate(
+                "PredefinedColorPalettes", qPrintable(paletteName));
         QIcon paletteIcon = drawPalettePreview(paletteName);
-        comboBoxHotcueColors->addItem(paletteName);
+        comboBoxHotcueColors->addItem(translatedName, paletteName);
         comboBoxHotcueColors->setItemIcon(
                 comboBoxHotcueColors->count() - 1,
                 paletteIcon);
 
-        comboBoxTrackColors->addItem(paletteName);
+        comboBoxTrackColors->addItem(translatedName, paletteName);
         comboBoxTrackColors->setItemIcon(
                 comboBoxTrackColors->count() - 1,
                 paletteIcon);
 
         if (palette.size() == 12) {
-            comboBoxKeyColors->addItem(paletteName);
+            comboBoxKeyColors->addItem(translatedName, paletteName);
             comboBoxKeyColors->setItemIcon(
                     comboBoxKeyColors->count() - 1,
                     paletteIcon);
@@ -117,12 +119,14 @@ void DlgPrefColors::slotUpdate() {
 
     const QSet<QString> colorPaletteNames = m_colorPaletteSettings.getColorPaletteNames();
     for (const auto& paletteName : colorPaletteNames) {
+        QString translatedName = QCoreApplication::translate(
+                "PredefinedColorPalettes", qPrintable(paletteName));
         QIcon paletteIcon = drawPalettePreview(paletteName);
-        comboBoxHotcueColors->addItem(paletteName);
+        comboBoxHotcueColors->addItem(translatedName, paletteName);
         comboBoxHotcueColors->setItemIcon(
                 comboBoxHotcueColors->count() - 1,
                 paletteIcon);
-        comboBoxTrackColors->addItem(paletteName);
+        comboBoxTrackColors->addItem(translatedName, paletteName);
         comboBoxTrackColors->setItemIcon(
                 comboBoxHotcueColors->count() - 1,
                 paletteIcon);
@@ -130,19 +134,19 @@ void DlgPrefColors::slotUpdate() {
 
     const ColorPalette trackPalette =
             m_colorPaletteSettings.getTrackColorPalette();
-    comboBoxTrackColors->setCurrentText(
-            trackPalette.getName());
+    comboBoxTrackColors->setCurrentText(QCoreApplication::translate(
+            "PredefinedColorPalettes", qPrintable(trackPalette.getName())));
 
     const ColorPalette hotcuePalette =
             m_colorPaletteSettings.getHotcueColorPalette();
-    comboBoxHotcueColors->setCurrentText(
-            hotcuePalette.getName());
+    comboBoxHotcueColors->setCurrentText(QCoreApplication::translate(
+            "PredefinedColorPalettes", qPrintable(hotcuePalette.getName())));
     slotHotcuePaletteIndexChanged(comboBoxHotcueColors->currentIndex());
 
     const ColorPalette keyPalette =
             m_colorPaletteSettings.getKeyColorPalette();
-    comboBoxKeyColors->setCurrentText(
-            keyPalette.getName());
+    comboBoxKeyColors->setCurrentText(QCoreApplication::translate(
+            "PredefinedColorPalettes", qPrintable(keyPalette.getName())));
 
     bool autoHotcueColors = m_pConfig->getValue(kAutoHotcueColorsConfigKey, false);
     if (autoHotcueColors) {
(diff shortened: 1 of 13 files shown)


src/util/color/predefinedcolorpalettes.cpp
```diff
@@ -159,6 +159,85 @@ constexpr mixxx::RgbColor kMixxxKeyColor10(0x3D8AFD);
...
+constexpr mixxx::RgbColor kMIKKeyColor5(0x01EAEC);
+constexpr mixxx::RgbColor kMIKKeyColor6(0x00EECB);
+constexpr mixxx::RgbColor kMIKKeyColor7(0x20EF7F);
+constexpr mixxx::RgbColor kMIKKeyColor8(0x7FF448);
+constexpr mixxx::RgbColor kMIKKeyColor9(0xE0CA6D);
+constexpr mixxx::RgbColor kMIKKeyColor10(0xFDA078);
+constexpr mixxx::RgbColor kMIKKeyColor11(0xFF8693);
+constexpr mixxx::RgbColor kMIKKeyColor12(0xFD7EB3);
+
+// Accessible Color Palettes
```
daschuer, line 194: An additional sentence about the trick you have used and the goal you achieved would be nice for later reference

src/util/color/predefinedcolorpalettes.cpp
```diff
@@ -390,6 +469,96 @@ const ColorPalette PredefinedColorPalettes::kMixxxKeyColorPalette =
...
+                        kDeutKeyColor8,
+                        kDeutKeyColor9,
+                        kDeutKeyColor10,
+                        kDeutKeyColor11,
+                        kDeutKeyColor12,
+                });
+
+const ColorPalette PredefinedColorPalettes::kTritKeyColorPalette =
+        ColorPalette(
+                QStringLiteral("Tritanopia / Tritanomaly Key Colors"),
```
daschuer, line 546: It would be nice to have the user visible strings translatable. However make sure that the untranslated string is written to the preferences. 

You can use. QT_TR_NOOP() to use the same string translated and untranslated:  https://doc.qt.io/qt-5/i18n-source-translation.html#using-qt-tr-noop-and-qt-translate-noop-in-c

daschuer (changes requested):

src/util/color/predefinedcolorpalettes.cpp
daschuer: The same applies to all pallets.

danferns:
@daschuer thank you for your feedback! I have made the changes. I replaced `QStringLiteral` with `QT_TR_NOOP`, I hope that is the right way.

daschuer:
This was the first step, now the Palette names appear in the translatable resources. 
The second step would be using it before showing it in the preferences, but not stirimg it in the config file. 

You can access the translated string like:
`tr(palette.getName())`

I suggest to add the translated string to the combo box as well as the untranslated string as item data QVariant for accessing the untranslated name after selection.

danferns:
Done. Thank you for the help. Let me know if there's anything else.

JoergAtGithub:
There is a merge conflict now - this needs to be fixed!

src/preferences/dialog/dlgprefcolors.cpp
```diff
@@ -84,31 +90,39 @@ DlgPrefColors::~DlgPrefColors() {
...
     comboBoxTrackColors->clear();
+    comboBoxKeyColors->clear();
     checkboxKeyColorsEnabled->setChecked(
             m_pConfig->getValue(kKeyColorsEnabledConfigKey,
                     BaseTrackTableModel::kKeyColorsEnabledDefault));
     for (const auto& palette : std::as_const(mixxx::PredefinedColorPalettes::kPalettes)) {
         QString paletteName = palette.getName();
         QIcon paletteIcon = drawPalettePreview(paletteName);
-        comboBoxHotcueColors->addItem(paletteName);
+        comboBoxHotcueColors->addItem(tr(qPrintable(paletteName)), paletteName);
```
daschuer, line 100: Did you consider to make paletteName `const char*` to get around the QString roundtrip?

danferns: Didn't think of that, but sounds like a good idea. I will give that a shot!

danferns: Done, this looks nicer now.

danferns: Wait. I just broke some of the palette names, will have to see why.

image

danferns: Interestingly, the palette names break differently each time I restart. 
image
image

danferns:
> There is a merge conflict now - this needs to be fixed!

Yes! I'm planning to rebase with the main branch after review.

src/preferences/dialog/dlgprefcolors.cpp
```diff
@@ -84,31 +90,39 @@ DlgPrefColors::~DlgPrefColors() {
 void DlgPrefColors::slotUpdate() {
     comboBoxHotcueColors->clear();
     comboBoxTrackColors->clear();
+    comboBoxKeyColors->clear();
     checkboxKeyColorsEnabled->setChecked(
             m_pConfig->getValue(kKeyColorsEnabledConfigKey,
                     BaseTrackTableModel::kKeyColorsEnabledDefault));
     for (const auto& palette : std::as_const(mixxx::PredefinedColorPalettes::kPalettes)) {
-        QString paletteName = palette.getName();
+        const char* paletteName = qPrintable(palette.getName());
```
daschuer, line 98: The returned pointer becomes invalid immediately after the call. If you like to reuse the value, store the translated string in a QString()

daschuer (changes requested):

src/util/color/predefinedcolorpalettes.cpp
```diff
@@ -169,7 +251,7 @@ namespace mixxx {
 
 const ColorPalette PredefinedColorPalettes::kMixxxHotcueColorPalette =
         ColorPalette(
-                QStringLiteral("Mixxx Hotcue Colors"),
+                QT_TR_NOOP("Mixxx Hotcue Colors"),
```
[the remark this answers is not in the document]
danferns: No probs! I understand that `tr()` is just a shorthand for `QCoreApplication::translate()`?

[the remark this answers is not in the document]
daschuer: tr() is a member of QObject that defined also the translation context. With translate() you can use any string as context. 
This needs to match the one in QT_TRANSLATE_NOOP()

daschuer:
Thank you. I have not tested it yet but it looks good.
Can you rebase it on the current main branch and squash out the forth and back commits? 
They where helpful during review!

danferns:
Done. Thank you for the review!

daschuer:
I can confirm that the translation and stroring works as expected.
Thank you. 

Howerver I get DEBUG assert violations: 

```
critical [Main] DEBUG ASSERT: "!"unexpected field for DecorationRole"" in function QVariant BaseTrackTableModel::roleValue(const QModelIndex&, QVariant&&, int) const at ./src/library/basetracktablemodel.cpp:1066
```
Hint: 
Under GDB you may break at these point by running Mixxx with 
```
run --developer --logFlushLevel debug --debugAssertBreak  
```

danferns:
Thank you. I just opened a thread on Zulip to help me figure out my debug build, but in the mean time, I removed all columns from the library except the key column, and these assertions disappeared. So it seems that Qt widgets call the DecorationRole by default. So I think we can either:
1) Remove the assertion to ignore calls from other columns
2) Go back from DecorationRole to UserRole. I had chosen DecorationRole as it was more descriptive of the purpose (used for colors/icons), but if it isn't the right choice, then we can switch to something else like the UserRole (custom purpose).

daschuer:
I have skimmed though the Qt source and seen the DecorationRole is manly used for file like icons in tree and list views. From this perspective it is a good fit for our icon like key indicator. So my vote is to remove the assertion and keep DecorationRole.

danferns:
Thank you! I also feel that keeping DecorationRole is more suitable. I've gone with this option.

daschuer (approved):

daschuer:
Who give this a final test and merge?

src/preferences/dialog/dlgprefcolors.cpp
```diff
@@ -207,6 +244,12 @@ void DlgPrefColors::slotApply() {
                         m_colorPaletteSettings.getTrackColorPalette()));
     }
 
+    if (!bKeyColorPaletteFound) {
+        m_colorPaletteSettings.setKeyColorPalette(
+                m_colorPaletteSettings.getColorPalette(keyColorPaletteName,
+                        m_colorPaletteSettings.getKeyColorPalette()));
+    }
```
Swiftb0y, line 251: I'm not happy with the amount of string roundtrips here....

daschuer: This is a string based lookup, because the user is able to specify custom palettes.
There might be a workaround for build in pallets, but since this is not a hot path I like to avoid special case handling here. Or do you have something different in mind?

Swiftb0y: I see. disregard my comment.

danferns:
Thank you for the feedback! I have made the changes requested.

Swiftb0y (approved):
LGTM I guess, though I haven't tested yet.

Swiftb0y:
tested, works very well, so I'll go ahead and merge. Thank you.
I have three requests for a follow up PR:

1. Make the key colors editable like the other colors
2. if key colors are disabled, grey out the key palette text and dropdown (and edit button)
3. update the key color palette preference when clicking apply, not only after closing the preference window.

daschuer:
Thank you for testing. 
Do you have a special use case for editable key colors? 
Just asking because editing is more difficult. They need to be on a circle of 12 neighbor colors. Maybe it is more easy to add a missing palette statically.

Swiftb0y:
I think it would nice for feature completeness... Though I acknowledge the challenges involved and that its probably overkill.

danferns:
Thank you! I've added your requests to my notes. Will get to them as soon as I can. :)

hofq:
Is this not included in the beta?
Don't see the menu here:
grafik

PR #13498, opened 2024-07-22
CoreServices: Default to `~/Music` as a music directory on Wasm and iOS

fwcd:
Both on the web and on iOS we are running in a sandbox that is generally empty on first launch, so there is not much point in asking the user for a directory. To simplify the UX, this PR proposes to create a music directory at the standard path and use it as a default music directory.

A (potentially undesirable?) side effect of this is that `~/Music` is an ancestor of the default recordings directory at `~/Music/Mixxx/Recordings`, so recordings will show up in the main library.

Maybe just ensuring that `~/Music` exists would already be enough to improve the UX here? On iOS this directory doesn't exist in the sandbox, so the directory picker initially points to the sandbox home instead, which could be confusing for new users.

### DIFF
--- a/src/coreservices.cpp
+++ b/src/coreservices.cpp
@@ -380,6 +380,18 @@ void CoreServices::initialize(QApplication* pApp) {
     bool musicDirAdded = false;
 
     if (m_pTrackCollectionManager->internalCollection()->loadRootDirs().isEmpty()) {
+#if defined(Q_OS_IOS) || defined(Q_OS_WASM)
+        // On the web and iOS, we are running in a sandbox (a virtual file
+        // system on the web). Since we are generally limited to this sandbox,
+        // there is not much point in asking the user, so we just default to
+        // ~/Music within it (creating it since it usually does not exist).
+        // Advanced users can still customize this directory in the settings.
+        QString fd = QStandardPaths::writableLocation(QStandardPaths::MusicLocation);
+        QDir dir = fd;
+        if (!dir.exists()) {
+            dir.mkpath(".");
+        }
+#else
         // TODO(XXX) this needs to be smarter, we can't distinguish between an empty
         // path return value (not sure if this is normally possible, but it is
         // possible with the Windows 7 "Music" library, which is what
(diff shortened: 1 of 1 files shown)


daschuer:
I have looked up the Qt implementation of `StandardPaths::MusicLocation`, but I did not find a specialisation for wasm or iOS. Is the unix implementation used? Do we need som sanity checks? 

In any case it would be nice to have a comment that clarifies which implementation we like to use.

fwcd:
I believe it defaults to `~/Music` where `~` is the sandbox's home directory. In both cases IIRC that directory doesn't exist before launching so it would probably be helpful to create it, even if we don't want to _default_ to it automatically

daschuer:
Did you check it with WASM? There are some confusions with ~ and WASM. Trusting Qt is not a good approach because it just used the Unix default fallback. So the result can be untested "by luck". 

My Music folder id "Musik" by the way. Should this also be translated or not. What does Qt return in the translated case? 

If there is only one correct answer and we want assert it we may hard code it.

fwcd:
> Did you check it with WASM? There are some confusions with ~ and WASM. Trusting Qt is not a good approach because it just used the Unix default fallback. So the result can be untested "by luck".

Yeah, ~ is the sandbox root in Wasm IIRC, but since the Wasm sandbox doesn't have a default music directory, I'd say the Unix fallback should be fine there.

> What does Qt return in the translated case?

I don't know what Windows does, but at least on macOS localized folders are only translated in the GUI, not in the file system. I _think_ it's the case there too, but they localize paths in a lot of placed anyway.

daschuer:
Can you double check that and add a comment?

acolombier:
Friendly ping @fwcd  :)

fwcd:
[The Qt docs state](https://doc.qt.io/qt-6/qstandardpaths.html#StandardLocation-enum) that `MusicLocation` resolves to `~/Documents/Music` on iOS and `~/Music` on Wasm (as part of the fallback) and I've updated the comment to reflect that. Both are reasonable choices, I'd say, and it doesn't really make sense to let the user pick anything since the available view of the file system is limited to the sandbox anyway. Localization should not affect us on either platform since the paths aren't localized in the file system.

acolombier:
Cool, thanks for confirming!

acolombier (approved):
LGTM! 
I'll go ahead and merge this - hope it's fine with you @daschuer

daschuer:
Yes, thank you.

PR #13504, opened 2024-07-23
GCC 14.1.1 fixes

### LINKED ISSUE #13467: Cannot build latest main (`-Werror-stringop-overflow=`)
### Bug Description

```
In file included from /home/jan/Projects/mixxx/src/util/samplebuffer.h:7,
                 from /home/jan/Projects/mixxx/src/engine/channels/enginedeck.h:8,
                 from /home/jan/Projects/mixxx/src/engine/channels/enginedeck.cpp:1:
In static member function ‘static void SampleUtil::clear(CSAMPLE*, SINT)’,
    inlined from ‘void EngineDeck::processStem(CSAMPLE*, int)’ at /home/jan/Projects/mixxx/src/engine/channels/enginedeck.cpp:62:22:
/home/jan/Projects/mixxx/src/util/sample.h:41:15: error: ‘void* memset(void*, int, size_t)’ specified bound between 18446744065119617024 and 18446744073709551612 exceeds maximum object size 9223372036854775807 [-Werror=stringop-overflow=]
   41 |         memset(pBuffer, 0, sizeof(*pBuffer) * numSamples);
      |         ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Version

commit 1b9ebc1b6cf203768f218d9a03962b8d249fa0b8

### OS

Arch Linux (gcc 14.1.1 20240522)

Holzhaus:
Fixes #13467.

### DIFF
--- a/src/util/sample.h
+++ b/src/util/sample.h
@@ -35,17 +35,24 @@ class SampleUtil {
     // Sets every sample in pBuffer to zero
     inline
     static void clear(CSAMPLE* pBuffer, SINT numSamples) {
-        // Special case: This works, because the binary representation
-        // of 0.0f is 0!
-        memset(pBuffer, 0, sizeof(*pBuffer) * numSamples);
-        //fill(pBuffer, CSAMPLE_ZERO, iNumSamples);
+        DEBUG_ASSERT(numSamples >= 0);
+        // We need to cast `numSamples` to an unsigned type to fix a
+        // `-Wstringop-overflow` warning on GCC 14.1.1. Casting to unsigned is
+        // okay, because the `DEBUG_ASSERT` above catches negative values
+        // anyway.
+        const auto sampleCount = static_cast<std::size_t>(numSamples);
+        // Special case: We can use memset here, because the binary representation
+        // of 0.0f is 0! This is much faster without optimizations than using
+        // `fill()`.
+        memset(pBuffer, 0, sizeof(*pBuffer) * sampleCount);
     }
 
     // Sets every sample in pBuffer to value
     inline
     static void fill(CSAMPLE* pBuffer, CSAMPLE value,
             SINT numSamples) {
-        std::fill(pBuffer, pBuffer + numSamples, value);
+        DEBUG_ASSERT(numSamples >= 0);
+        std::fill_n(pBuffer, numSamples, value);
     }
 
     // Copies every sample from pSrc to pDest
--- a/src/waveform/waveformwidgetfactory.h
+++ b/src/waveform/waveformwidgetfactory.h
@@ -122,7 +122,9 @@ class WaveformWidgetFactory : public QObject, public Singleton<WaveformWidgetFac
     void setOverviewNormalized(bool normalize);
     int isOverviewNormalized() const { return m_overviewNormalized;}
 
-    const QVector<WaveformWidgetAbstractHandle> getAvailableTypes() const { return m_waveformWidgetHandles;}
+    const QVector<WaveformWidgetAbstractHandle>& getAvailableTypes() const {
+        return m_waveformWidgetHandles;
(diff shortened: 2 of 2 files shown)


src/util/sample.h
```diff
@@ -35,17 +35,14 @@ class SampleUtil {
     // Sets every sample in pBuffer to zero
     inline
     static void clear(CSAMPLE* pBuffer, SINT numSamples) {
-        DEBUG_ASSERT(numSamples >= 0);
-        // Special case: This works, because the binary representation
-        // of 0.0f is 0!
-        memset(pBuffer, 0, sizeof(*pBuffer) * numSamples);
-        //fill(pBuffer, CSAMPLE_ZERO, iNumSamples);
+        fill(pBuffer, CSAMPLE_ZERO, numSamples);
     }
```
Swiftb0y, line 39: fyi, this is [only equally performant on `-O2` and not on `-O1` on gcc](https://quick-bench.com/q/tYpHlnq8rlRf3aNihqCQC3oetwU). I guess thats fine? idk about msvc but afaikt their optimizer is terrible...

The original bug was saying that the `memset` is operating on wayy too much memory right? Maybe thats a hint that there is actually a bug here somewhere (maybe passing some negative `numSamples` and then underflowing)?

Holzhaus:
I guess it makes sense to backport thesr to 2.4?

src/util/sample.h
Holzhaus: I was also wondering if these was a reason for using memset. The original code already had the `fill` call commented out. Not sure if `fill` was the original implementation and turned out to be too slow, so that it was replaced by `memset`? Unfortunately there is no comment to explain it.

Holzhaus: Fun fact on clang 17 both Implementations are equivalent even with `-O1`.

Holzhaus:
Rebased to 2.4. One commit wasn't applicable to 2.4, moved that to #13505 instead.

src/util/sample.h
JoergAtGithub: https://travisdowns.github.io/blog/2020/01/20/zero.html

daschuer: I can remember that fill was slow with MSVC. With GCC, memset even used AVX registers if it finds some on the target hardware. 

I have just checked with GCC and it turn out that fill also calls memset. 
In addition the fill has two extra checks that skips memset call if numSamples is zero or numSamples * sizeof(*pBuffer) is zero. They are redundant in our case. 

What is the warning you are trying to fix?  

I think we should test the current situation with MSVC before doing this change.

daschuer: I have noticed that 
```
void EngineDeck::processStem(CSAMPLE* pOut, const int iBufferSize) {
``` 
uses int and not SINT. Does that fix the warning?

daschuer: The diff value the compilers is complaining about is 0x1FFFFFFFC. Strange.

daschuer: The whole clear call is redundant at that place. 
Can we just replace it and keep the clear() implementation unchanged?

JoergAtGithub: Yes, let's the clear implementation as it is and silence this false positive. This is proven code and fast under all conditions, while fill is only fast under some conditions, when it internally calls memset.

Holzhaus: > https://travisdowns.github.io/blog/2020/01/20/zero.html

lol, it's actually faster with GCC and `-O1`: https://quick-bench.com/q/H1hlRRCNe_t-6-iRgTplPpVMKQA

daschuer (approved):
LGTM, Thank you.

JoergAtGithub (approved):

PR #13505, opened 2024-07-23
fix(basetracktablemodel): Fix `-Wimplicit-fallthrough` warning on GCC 14.1.1

Holzhaus:
Fixes the following warning:

    src/library/basetracktablemodel.cpp: In member function ‘QVariant BaseTrackTableModel::roleValue(const QModelIndex&, QVariant&&, int) const’:
    src/library/basetracktablemodel.cpp:1031:9: error: this statement may fall through [-Werror=implicit-fallthrough=]
     1031 |         switch (field) {
          |         ^~~~~~
    src/library/basetracktablemodel.cpp:1042:5: note: here
     1042 |     default:
          |     ^~~~~~~
    cc1plus: all warnings being treated as errors

This seems to be a false-positive, but this commit should fix the issue without any changes to the behavior.

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -1033,11 +1033,10 @@ QVariant BaseTrackTableModel::roleValue(
             // return whether or not to display the color rectangle
             return QVariant::fromValue(s_keyColorsEnabled);
         }
-        default: {
+        default:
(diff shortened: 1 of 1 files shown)


Swiftb0y (approved):
LGTM. thank you.

PR #13506, opened 2024-07-23
fix(cmake): Remove target compile defs for non-existing QML target

Holzhaus:
The target `mixxx-qml-lib` does not exist in the 2.4 branch. This is likely some left-over/copy-and-paste error.

If you build with `-DQT6=ON` (and the `GITHUB_ACTIONS` environment variable set), this leads to the CMake error:

    CMake Error at CMakeLists.txt:2675 (target_compile_definitions):
      Cannot specify compile definitions for target "mixxx-qml-lib" which is not
      built by this project.


JoergAtGithub:
LGTM! Thank you!

Note: We must ensure that this get not merged into 2.5 and main.

Holzhaus:
> Note: We must ensure that this get not merged into 2.5 and main.

#13705

PR #13520, opened 2024-07-30
add mapping for Intech TEK2

Greg-Orca:


ronso0:
Just a note: it's advised to name feature branches acordingly, e.g. `tek2-mapping`. I.e. don't commit to the base branch of your fork (2.4 in this case). See https://github.com/mixxxdj/mixxx/wiki/using-git#create-a-new-branch

(I'm mentioning this because just recently we (I) ran into issues when I pushed to a contriutor's branch that was named like the target branch)

PR #13521, opened 2024-07-30
Add mapping for Intech TEK2

Greg-Orca:


res/controllers/Intech TEK2.scripts.js
```diff
@@ -0,0 +1,32 @@
+const TEK2 = {};
```
Swiftb0y, line 1: Have you ensured that this still works with `const`? I'm skeptical that it will work because `const` (and `let`) work a little differently than `var`.

res/controllers/Intech TEK2.scripts.js
```diff
@@ -0,0 +1,32 @@
...
+TEK2.shutdown = function() {};
+
+// The button that enables/disables scratching
+TEK2.wheelTouch = function(channel, control, value, status, group) {
+    const deckNumber = script.deckFromGroup(group);
+    if (value === 0x7f) {
+    // If button downs
+        const alpha = 1.0 / 8;
+        const beta = alpha / 32;
+        engine.scratchEnable(deckNumber, 128, 33 + 1 / 3, alpha, beta);
```
Swiftb0y, line 16: 128 seems quite low-res... Are you sure thats how many ticks there are in a full rotation?

res/controllers/Intech TEK2.scripts.js
Greg-Orca: I checked it, and it's only works with var. but pre-commit is not let me use var. why is const or let not working? "TEK2 is not defined"

res/controllers/Intech TEK2.scripts.js
Greg-Orca: I adjusted the Endless Potmeter in the Grid Editor, and it's worked nicely with this value.

res/controllers/Intech TEK2.scripts.js
Swiftb0y: its a little complicated on why it makes a difference. In short `var` modifies the `this` object of the current scope while `let` and `const` don't. Since we need lookup `TEK2` on the global object when handling midi message, its important to use `var`. So just do 
```suggestion
// eslint-disable-next-line no-var
var TEK2 = {};
```
instead.

res/controllers/Intech TEK2.scripts.js
Swiftb0y: So if with the current value, doing one full rotation of physical encoder matches one full rotation of the vinyl spinny widget?

res/controllers/Intech TEK2.scripts.js
Greg-Orca: thank you!

res/controllers/Intech TEK2.scripts.js
Greg-Orca: Good point. I made some adjustments, and now it matches both Scratch and Simple Turn.

Swiftb0y:
Great. thank you. We also have a policy that each controller mapping needs a manual page. Would it be possible for you to write one for https://github.com/mixxxdj/manual ? It doesn't need to be complicated, just mentioning all the mixxx-specific parts and linking to your own documentation for the rest is fine.

ronso0:
Do I understand correctly that this maps the wheels and two buttons on each side?
What about the other two button? Could they be mapped to Pfl?
I'm asking because it would be nice to provide a 'complete' mapping = as many mapped controls as possible.

Greg-Orca:
Actually, this map includes 4 buttons, so I will add this to the article tomorrow! The other two buttons are not configured; it really depends on what the user prefers. I used them for forward and backward, but it's also nice to use them for loop in/out.

ronso0:
What about the LEDs above the mapped buttons?
Would make sense they reflect the 'playing' and 'at cue' status. Can they be addressed via MIDI with the 'TEK2 DJ setup' config? Do yo intend to map them , too? Do you need help with that?

ronso0:
> Would make sense they reflect the 'playing' and 'at cue' status. Can they be addressed via MIDI with the 'TEK2 DJ setup' config? Do yo intend to map them , too? Do you need help with that?

FYI the outputs are described here https://github.com/mixxxdj/mixxx/wiki/Midi-Controller-Mapping-File-Format#outputs

And the controls are
cue (lit when playpos is at the main cue): [`cue_indicator`](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-cue_indicator)
play (when playing, incl. previewing hotcues): [`play`](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-play)
play (only when play button is latched): [`play_latched`](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-play_latched)
play indicator (may blink): [`play_indicator`](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[ChannelN]-play_indicator)

Greg-Orca:
I currently lack the time for this, and I'm about to going to vacation. Let's proceed this, and I'll enhance it later.

daschuer:
The manual PR is here: 
https://github.com/mixxxdj/manual/pull/686

@Swiftb0y Merge or wait?

Swiftb0y:
merge :shrug:

daschuer:
Hallo @Greg-Orca I have just realized that we missed to ask formally for permission to distribute your code with Mixxx. In order to catch that up, please sign https://docs.google.com/a/mixxx.org/spreadsheet/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ
and comment here when done. Thank you.

Greg-Orca:
Hi @daschuer, I believe I had already done this before, but I did it again just now.

daschuer:
@Greg-Orca Thank you. I like to add you name to the Mixxx about box. Which name should I use?

PR #13430, opened 2024-07-01
Skins: fix time display to allow AM/PM

ronso0:
I removed the fixed hh:mm format which allows AM/PM and adjusted the size policies and the parent WidgetGroups so AM/PM is not cut off.

Please double check each skin.

Fixes #13421

Issue #13421, opened 2024-06-30
Clock display incomplet


RafaelFrance:
### Bug Description

Tango skin

When 11:27 AM, the "M" is not visible.
![Display clock 11 27 PM - incomplet](https://github.com/mixxxdj/mixxx/assets/51053041/d56d0f05-1fce-493f-bda3-d3c36fec07bb)

When 1:34 PM, it is correct.
![Display clock display 1 34 PM - Correct](https://github.com/mixxxdj/mixxx/assets/51053041/8f1810f7-7bf1-4c06-8eb1-69c14ff7e49a)

### Version

2.5-beta-13-gc6d1d5f691

### OS

Windows 11 Professionnal version 23H2

Swiftb0y:
@ronso0 probably scaling related?

ronso0:
Hmm, the screenshots look as they are @100%.
It's just that the widget has afixed size and will clamp AM/PM

Fix is underway

RafaelFrance:
I confirm Interface setting. HiDPI / Retina scaling : 100% 

Laptop screen : 17''

ronso0:
The fix is in #13430, please test.

RafaelFrance:
Tested. Git Version : 2.4.1-36-ga9b6791259 (HEAD).

AM/PM is absent. Only "10:02" in place "10:02 PM".

ronso0:
I can verifiy AM/PM is used when I pick the `en_US` locale in Mixxx.

FWIW my `LC_CTYPE` is `en_GB.UTF-8` but somehow Mixxx detects this as 'C' locale (IIRC there was a related fix, but since I'm not into that so I can't tell if it's a actually a bug), skips translations and gives me hh:mm time format (as expected for en_GB).

PR #13525, opened 2024-07-30
Remove custom clock format from skins

ywwg:
Mixxx should use the locale's preferred clock format rather than arbitrarily overriding it.  For instance on en_US locale, it's customary to use 12 hour time + am/pm rather than 24 hour time. This should have no effect on machines set to locales that use 24hr time.

If people would prefer to have a preference for clock format I could see adding one.


ronso0:
Hmm I already opened #13430 (fixing #13421)

PR #13528, opened 2024-07-31
Update GitHub actions

daschuer:
Hopefully this fixes the following warning 
```
The following actions uses Node.js version which is deprecated and will be forced to run on node20: actions/cache@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
```


daschuer:
Worked

ronso0:
Thanks!

PR #13530, opened 2024-08-01
chore: update pre-commit hook

acolombier:
`qml_formatter` needs to be updated to include the latest fix needed to handle string templating. 

Currently there is no tag release for it so I've pin the commit for safety.

I also took the freedom to batch update all the hooks. Let me know if you would like me to be kept at their current version.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -32,7 +32,7 @@ default_language_version:
   rust: 1.64.0
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.4.0
+    rev: v4.6.0
     hooks:
       - id: fix-byte-order-marker
         exclude: ^.*(\.cbproj|\.groupproj|\.props|\.sln|\.vcxproj|\.vcxproj.filters|UTF-8-BOM.txt)$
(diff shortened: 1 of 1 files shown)


daschuer:
An unrelated test is failing:
`[  FAILED  ] StemControlTest.StemColor`

daschuer:
Do we want to rename this on 2.4?

acolombier:
> An unrelated test is failing:

Ah that's weird, looks like this might be a flaky one. COuld you re-run the job to see if that was a one-off? I will make a fix otherwise

> Do we want to rename this on 2.4?

I assume you meant rebase, happy to do so if you think that would be valuable! Arguably, the core issue this fixes (QML formatting) is only for 2.6+ tho

Swiftb0y (approved):
LGTM. but please rebase on 2.4. Failed stem tests seem to be flaky, would be nice if you could look into those.

Swiftb0y:
friendly ping to rebase on 2.4 @acolombier ;)

acolombier:
Sorry for the delay @Swiftb0y - done!

Swiftb0y:
Thank you.

PR #13537, opened 2024-08-04
Non-floating stem controls for LateNight

JoergAtGithub:
This is an alternative to the UI PR #13515 by @acolombier . It use classic widget groups instead of the floating widget.
grafik

grafik

The functionality is the same, but the widget positioning differs. Both PRs are in prove of concept state and not visually fine tuned.

The benefit of the floating widgets is that the waveform is only hidden by the stem widgets for the tracks that are stem files, while here, the stem widgets are either shown for all tracks or for none.

Please test both PRs #13515 and this one. (Build with FFMPEG and STEM enabled in CMake) and load a .stem file. And review the code of the second commit here. Which approach should we follow?

If you need STEM files for test, you can download the reference files here: https://www.native-instruments.com/fileadmin/ni_media/downloads/stems/free_stems_for_traktor_pro_2.zip

### DIFF
--- a/src/skin/legacy/legacyskinparser.cpp
+++ b/src/skin/legacy/legacyskinparser.cpp
@@ -983,6 +983,13 @@ QWidget* LegacySkinParser::parseStemLabelWidget(const QDomElement& element) {
 
     QString group = lookupNodeGroup(element);
     BaseTrackPlayer* pPlayer = m_pPlayerManager->getPlayer(group);
+    if (!pPlayer) {
+        SKIN_WARNING(element,
+                *m_pContext,
+                QStringLiteral("No player found for group: %1").arg(group));
+        return nullptr;
+    }
+
     connect(pPlayer,
             &BaseTrackPlayer::newTrackLoaded,
             pLabel,
@@ -992,6 +999,13 @@ QWidget* LegacySkinParser::parseStemLabelWidget(const QDomElement& element) {
             &BaseTrackPlayer::trackUnloaded,
             pLabel,
             &WStemLabel::slotTrackUnloaded);
+
+    TrackPointer pTrack = pPlayer->getLoadedTrack();
+    if (pTrack) {
+        // Set the trackpoinnter to the already loaded track,
+        // needed at skin change
+        pLabel->slotTrackLoaded(pTrack);
+    }
 
     return pLabel;
 }
--- a/src/skin/legacy/legacyskinparser.cpp
+++ b/src/skin/legacy/legacyskinparser.cpp
@@ -740,6 +740,11 @@ void LegacySkinParser::parseChildren(
 }
 
 QWidget* LegacySkinParser::parseWidgetGroup(const QDomElement& node) {
+#ifndef __STEM__
+    if (requiresStem(node)) {
+        return nullptr;
+    }
+#endif
     WWidgetGroup* pGroup = new WWidgetGroup(m_pParent);
     setupBaseWidget(node, pGroup);
     setupWidget(node, pGroup->toQWidget());
@@ -2576,4 +2581,9 @@ QString LegacySkinParser::stylesheetAbsIconPaths(QString& style) {
     // skins directory for the launch image style.
     style.replace("url(skins:", "url(" + m_pConfig->getResourcePath() + "skins/");
     return style.replace("url(skin:", "url(" + m_pContext->getSkinBasePath());
+}
+
+bool LegacySkinParser::requiresStem(const QDomElement& node) {
+    QDomElement requiresStemNode = node.firstChildElement("RequiresStem");
+    return !requiresStemNode.isNull() && requiresStemNode.text() == "true";
 }
--- a/src/skin/legacy/legacyskinparser.cpp
+++ b/src/skin/legacy/legacyskinparser.cpp
@@ -979,7 +979,7 @@ void LegacySkinParser::setupLabelWidget(const QDomElement& element, WLabel* pLab
 #ifdef __STEM__
 QWidget* LegacySkinParser::parseStemLabelWidget(const QDomElement& element) {
     WStemLabel* pLabel = new WStemLabel(m_pParent);
-    setupStemLabelWidget(element, pLabel);
+    setupLabelWidget(element, pLabel);
 
     QString group = lookupNodeGroup(element);
     BaseTrackPlayer* pPlayer = m_pPlayerManager->getPlayer(group);
@@ -1008,20 +1008,6 @@ QWidget* LegacySkinParser::parseStemLabelWidget(const QDomElement& element) {
     }
 
     return pLabel;
-}
-
-void LegacySkinParser::setupStemLabelWidget(const QDomElement& element, WStemLabel* pLabel) {
-    // NOTE(rryan): To support color schemes, the WWidget::setup() call must
-    // come first. This is because WLabel derivatives change the palette based
-    // on the node and setupWidget() will set the widget style. If the style is
-    // set before the palette is set then the custom palette will not take
-    // effect which breaks color scheme support.
-    pLabel->setup(element, *m_pContext);
-    commonWidgetSetup(element, pLabel);
-    pLabel->installEventFilter(m_pKeyboard);
-    pLabel->installEventFilter(
-            m_pControllerManager->getControllerLearningEventFilter());
-    pLabel->Init();
 }
 #endif
 
--- a/src/skin/legacy/legacyskinparser.cpp
+++ b/src/skin/legacy/legacyskinparser.cpp
@@ -74,6 +74,9 @@
 #include "widget/wsplitter.h"
 #include "widget/wstarrating.h"
 #include "widget/wstatuslight.h"
+#ifdef __STEM__
+#include "widget/wstemlabel.h"
+#endif
 #include "widget/wtime.h"
 #include "widget/wtrackproperty.h"
 #include "widget/wtrackwidgetgroup.h"
@@ -559,7 +562,13 @@ QList<QWidget*> LegacySkinParser::parseNode(const QDomElement& node) {
         result = wrapWidget(parseLabelWidget<WNumberDb>(node));
     } else if (nodeName == "Label") {
         result = wrapWidget(parseLabelWidget<WLabel>(node));
-    } else if (nodeName == "Knob") {
+    }
+#ifdef __STEM__
+    else if (nodeName == "StemLabel") {
+        result = wrapWidget(parseStemLabelWidget(node));
+    }
+#endif
+    else if (nodeName == "Knob") {
         result = wrapWidget(parseStandardWidget<WKnob>(node));
     } else if (nodeName == "KnobComposed") {
         result = wrapWidget(parseStandardWidget<WKnobComposed>(node));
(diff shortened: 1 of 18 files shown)


acolombier:
That's looking very promising! :rocket:

Swiftb0y (commented):
had a quick look. in pretty good shape. I just wish the `*Manager` classes wouldn't get bloated so much, but there is not much we can do about that because Manager classes are fundamentally flawed...

acolombier:
Thanks for the review @Swiftb0y !

Looks like most of those comments are meant for #13123, on which this PR is based. I will address them, but feel free to continue the review there, in case you have more feedback!

src/skin/legacy/legacyskinparser.cpp
```diff
@@ -1732,7 +1777,6 @@ QDomElement LegacySkinParser::loadTemplate(const QString& path) {
 
     if (!templateFile.open(QIODevice::ReadOnly)) {
         qWarning() << "Could not open template file:" << absolutePath;
-        return QDomElement();
     }
 
     QDomDocument tmpl("template");
```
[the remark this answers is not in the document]
JoergAtGithub: That is a wrong resolved merge conflict

Eve00000:
I really appreciate the work/time your investing in this matter, so I did a thorough test and I wanted to write a profound report.
due to some errors on my behalve I had a rather strange effect that made it on the other side very clear to compair (see image).

I want to start with a +/- list
subject / design 1 / design 2
**  indicating clearly a stem is loaded
-> design 1 only shows stem controls when a stem is loaded, design 2 shows them stem controls fora all decks even when there is no (stem) file loaded. the floating widget seems to be more scalable than the fixed. if you don't need the guicontrols to manipulate the stems they take a lot of space in design 2. Having 4 decks + samplers (even mini view) leaves no space for library.
-> this can / is confusing. A possible solution can be to have little rectangles (same design as Daniel uses in the Keys) in the stemcontrols that show the trackcolor as wel. (maybe the stars can be that color to)

** clarity of the stemcontrols
->the coloring of the tracks in design 1 was/is a little too 'invasive', the colors blinded the readability of the information (loaded effect, volume/fx potmeter level) especially when stems are loaded in all decks. The opacity of the controls when a stem-track is muted is clear. design 2 has the mute buttons as well.
-> design 2 is more 'relaxing" for the eyes, but a stem-track-color would clearify the view from a distance. 

** floating or fixed
-> in the floating design the controls cover all views that are not showing the complete waveform (big library, 4-> 2 decks with stems loaded in 3/4) can be confusing. In the fixed design the controls are always there (unless hidden with the button) witch has an influence on the waverforms window as well. (4 decks = huge)
-> I find the fixed design more relaxing

Resumé
Both designs have + and - , so the optimal design must be somewhere in the middle (the truth is out there),
- stem-track-colored rectangles indicating a stem is loaded, can the opacity be combined?
- can the effectname be smaller?
- can there be a different design for "only info" / "info + manipulate", the controls in the first could be smaller 
- can there be an indication in the deck that a stem is loaded?
- floating design is scalable and leaves place for library , samplers and fx, the fixed design doesn't.

So there should be thought about diffeent user types:
- using the gui as info ->smaller (hidable) stemcontrols
- using the gui to control Mixxx -> bigger clear controls
 

PS: The track in the image remembered me of the early 90's, for a couple of weeks I was overwelmed with "Bells of NY', I got a white label and played (parts of) it maybe 5 times a night. Thanks for remembering me.

afbeelding

daschuer:
Here my 2 ct: 
* The floating version was too colorfull. 
* The uncovered area in the floating type was useless
* The changing size was more anoying than helpfull.
* Requrement: Alignment of deck waveforms 
* The colored texts may not work well / are readable with all Skin Color schemes and stem colors -> extra area or border
* The combined waveforms sometimes hide details in single stems. Idea: Side by side version. 
* Nice: Show/Hide controls per deck. Waform chopped on the left, not scaled to always match non stem decks.
* Force a minimm wavform size that all controls have the normal size
* I have no demand to show and hide the waveforms, but a demand to switch back to stereo mode and let the track flow without a high CPU load. Maybe an extra small widget that is shown in stereo mode and allows to switch back to stem mode showing the full controlls.

ronso0:
Just a quick feedback / idea, let me know what you think:
the play/mute controls could be buttons with a label (not labels + small buttons)
Ie. I'd prefer larger controls as GUI equivalent for pads on (stem-ready) controllers.

I'll provide a commit fixing the indentation soonish, then take a look at the UX and the c++ part.

acolombier (approved):
LGTM!

@ronso0's suggestion sounds great, but also conscious than working on the legacy UI is not very easy/rewarding so happy with the current look!

JoergAtGithub:
Yes, a WStemButton would also make sense, an saves some space. I derived WStemLabel from WLabel, but I guess it should be easy to do the same with WPushButton instead.

Swiftb0y (approved):
can't comment on the qss and xml changes (@ronso0 could you have a look?) Cpp changes LGTM.

ronso0:
Yes, working on it right now.

ronso0:
Fixup is here 3975eddd2e
I fixed the knobs, adjusted the layout a bit, etc.

FWIW the action buttons in LateNight indicate what happens when you push them (not the current state, eg. `| |` icon when a track is playing), so I changed the stem mute icons, too.
image

ronso0:
> Fixup is here https://github.com/mixxxdj/mixxx/commit/3975eddd2e786c18aeabae4d04dedf6ee586fb4c

whoops, this had some residues in the SVG.
here's the clean one 709da44e6c, please pick that instead.

JoergAtGithub:
@ronso0 Is it intended, that the Effect-Name is no longer left alligned?
grafik

JoergAtGithub:
@ronso0 In your patch the changed SVG for LateNight Classic seems to be missing

ronso0:
> @ronso0 Is it intended, that the Effect-Name is no longer left alligned?

Whoops, only tested with long names. Will fix it.

I didn't change the Classic icon. Will take a look.

ronso0:
Done. Polished commit is b08fe0e7a2

JoergAtGithub:
I noticed two visual issues, which are independent of the last commit:
grafik

1.) With very long stem labels, only the  channel section with this label becomes wider. This is an issue because the waveform shrinks and is no longer aligned to the other waveforms. I think we will need an additional invisible container around all channels
2.) The default quick effect can be "---" or ""

Than we have also a build issue. We have the CMake setting STEM to enable the feature, but this does not affect the skin, only the C++ code. AFAIK an #ifdef like command in the skins is not available?

ronso0:
> We have the CMake setting STEM to enable the feature, but this does not affect the skin, only the C++ code. AFAIK an #ifdef like command in the skins is not available?

Some ideas:
1. abuse new ControlObjects to hide the widgets
= cumbersome
2. add a wrapper class for WidgetGroup, eg. StemWidgetGroup, that is only parsed if STEM is enabled
(+) safe IMO
(-) we'd need to implement this for every widget class used for stems that is placed in regular widget groups, eg. some stem control in the mixer where wrapping in an extra StemWidgetGroup would be overkill
3. add a `<RequiresStem>` node to WidgetGroups in question, and only parse the WidgetGroup if STEM is enabled
(+) should be easy to implement in [`LegacySkinParser::parseWidgetGroup`](https://github.com/mixxxdj/mixxx/blob/8790b5fbf2eb555da740a16ae39cf08215868031/src/skin/legacy/legacyskinparser.cpp#L733)
(+) could also be a helper function called by **any widget parser**

3 is my favourite.

src/widget/wstemlabel.cpp
```diff
@@ -0,0 +1,51 @@
...
+        : WLabel(pParent),
+          m_stemInfo(QString(), QColor()),
+          m_stemNo(0) {
+}
+
+void WStemLabel::setup(const QDomNode& node, const SkinContext& context) {
+    m_stemNo = context.selectInt(node, "StemNum");
+}
+
+void WStemLabel::slotTrackUnloaded(TrackPointer track) {
```
ronso0, line 15: here and elsewhere
```suggestion
void WStemLabel::slotTrackUnloaded(TrackPointer pTrack) {
```

src/widget/wstemlabel.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include "moc_wstemlabel.cpp"
+
+WStemLabel::WStemLabel(QWidget* pParent)
+        : WLabel(pParent),
+          m_stemInfo(QString(), QColor()),
+          m_stemNo(0) {
+}
+
+void WStemLabel::setup(const QDomNode& node, const SkinContext& context) {
+    m_stemNo = context.selectInt(node, "StemNum");
```
ronso0, line 12: assert that num is >= 0 and < N (what's the upper limit?)
otherwise we'd get `ASSERT failure in QList::operator[]: "index out of range"`

src/widget/wstemlabel.cpp
```diff
@@ -0,0 +1,51 @@
...
+        return;
+    }
+
+    auto stemInfo = track->getStemInfo();
+
+    if (stemInfo.isEmpty()) {
+        return;
+    }
+
+    m_stemInfo = stemInfo[m_stemNo - 1];
```
ronso0, line 32: assert that m_stemNo <= m_stemInfo.size()
otherwise we'd get `ASSERT failure in QList::operator[]: "index out of range"`

ronso0 (changes requested):
LGTM, except we need to add assertions when accessing the steminfo list.

ronso0:
@JoergAtGithub I'm currently travelling and i'll review as soon as I'm back home.

Eve00000:
@ronso0 : enjoy your holiday, have a good time

src/widget/wstemlabel.cpp
```diff
@@ -0,0 +1,69 @@
...
+          m_stemNo(0) {
+}
+
+void WStemLabel::setup(const QDomNode& node, const SkinContext& context) {
+    m_stemNo = context.selectInt(node, "StemNum");
+
+    VERIFY_OR_DEBUG_ASSERT(m_stemNo >= 1 && m_stemNo <= mixxx::kMaxSupportedStems) {
+        kLogger.warning() << "Skin warning: StemNum is out of range. It should "
+                             "be between 1 and "
+                          << mixxx::kMaxSupportedStems;
```
ronso0, line 20: A mixxx::logger is nice, but here we'd actually use the `SKIN_WARNING` macro IMO

ronso0:
https://github.com/ronso0/mixxx/commit/b9da04dd3f2eb45ae87fccf96303575e2647a299 fixes the tabs in waveform_container.xml

src/skin/legacy/legacyskinparser.cpp
```diff
@@ -967,6 +976,55 @@ void LegacySkinParser::setupLabelWidget(const QDomElement& element, WLabel* pLab
...
+    // on the node and setupWidget() will set the widget style. If the style is
+    // set before the palette is set then the custom palette will not take
+    // effect which breaks color scheme support.
+    pLabel->setup(element, *m_pContext);
+    commonWidgetSetup(element, pLabel);
+    pLabel->installEventFilter(m_pKeyboard);
+    pLabel->installEventFilter(
+            m_pControllerManager->getControllerLearningEventFilter());
+    pLabel->Init();
+}
```
ronso0, line 1025: `WStemLabel` is a `WLabel`, so why not use `LegacySkinParser::setupLabelWidget()`? 

If there's a reason _not_ to use that, why put this into an extra function that's used only for WStemLabel? (vs. put it in `parseStemLabelWidget)`

ronso0: `LegacySkinParser::setupLabelWidget()` works just fine.
74a1c863e1

ronso0:
> Than we have also a build issue. We have the CMake setting STEM to enable the feature, but this does not affect the skin, only the C++ code. AFAIK an #ifdef like command in the skins is not available?

What about not ifdef'ing the stem label parser itself (which gives us a skin warning "Invalid node name in skin: .."), but use the ifdef in `parseStemLabelWidget` to log a note saying that WStemLabel requires building with STEM enabled?

ronso0:
LGTM, but the indentations of the xml files are screwed up again. I think you need to adjust your editor to use spaces instead of tabs like in c++ code. Fixup or append a commit, as you like.

JoergAtGithub:
Done!

ronso0 (approved):
LGTM, thank you very much!

PR #13538, opened 2024-08-05
(fix) use 'played' track color also for played/play count delegate

ronso0:
`PlayCountDelegate` was not picking up the 'played' track color like other delegate(s).
(tick the played checkbox and its label color wouldn't change to grey)

This adds `CheckboxDelegate` base class (and removes now obsolete `PlayCountDelegate`).
It's used for BPMDelegate and the play count delegate, which allows moving redundant paintItem() code to `CheckboxDelegate`.


JoergAtGithub:
Works as expected!

JoergAtGithub:
Thank you!

Issue #13449, opened 2024-07-07
Mixxx incorrectly detects silence at the end of some (few) tracks


Gilles-Poulleau:
### Bug Description

I have experienced Mixxx incorrectly detecting silence at the end of some (few) tracks. 
It's different from #10791 **replacing a track with a better version leads to the risk of the song being cut off abruptly during broadcast**.
For these files, resetting the parameters doesn't change anything, the new analysis gives the same result which leads to the same behaviour in AUTO-DJ. Except if your are in full track mode, the end of the track is abruptly cut, leading to dancers bad experience and it's much worse for the DJ.

By the way, if I detect the problem, removing the mark before the incorrect end, makes the mark disappear from the waveform but the song will still be cut too early.... ?
The only way around is too quickly select "full track" mode.
![mixx incorrectly detecting end of track 2](https://github.com/mixxxdj/mixxx/assets/92171520/d63ebb09-b7c2-4f29-868b-b2c44d836507)
![mixx incorrectly detecting end of track](https://github.com/mixxxdj/mixxx/assets/92171520/21e010a5-af54-410c-95cb-4f07372e677c)
[- Humillación Que Nadie Sepa Mi Sufrir too early cut.zip](https://github.com/user-attachments/files/16119695/-.Humillacion.Que.Nadie.Sepa.Mi.Sufrir.too.early.cut.zip)

### Version

2.4.1

### OS

Windows 11

Swiftb0y:
Can confirm on Fedora 40 as well. filetype/decoder doesn't matter afaict.

ronso0:
> By the way, if I detect the problem, removing the mark before the incorrect end, makes the mark disappear from the waveform but the song will still be cut too early.... ?

Initially, the outro_end marker is positioned at what the analyzer considers the last sound (level falls below -60dB).
However, manally moving or removing the outro_end will unfortunately not re/move the 'last sound' cue.

ronso0:
@Gilles-Poulleau @Swiftb0y Could you test affected with some earlier Mixxx version (2.3.x)?

Gilles-Poulleau:
Sorry for the delay. I have been able to test it with mixxx version 2.3.3.
And it works fine, the end of both songs is correctly detected in this version.

daschuer:
I see. It is a regression form https://github.com/mixxxdj/mixxx/commit/d993c034a553e16063abf6f530fc1acad00e141e when introducing the `std::reverse_iterator`
I will take a look.

[...]

PR #13545, opened 2024-08-08
Fix detection last sound if track does not end with silence.

daschuer:
This can be fixed by deleting the outro cue points and reload the track into a deck. Fixes https://github.com/mixxxdj/mixxx/issues/13449


Swiftb0y (approved):
LGTM. Thanks for investigating. This may have been spotted earlier if we properly encoded the "no silence found" state in the typesystem (`std::optional<SINT>` maybe).

PR #13547, opened 2024-08-08
Merge 2.5

ronso0:


ronso0:
This is with 199a5433d5 from #13549 (which I'll remove before merge) to verify only qsscheck is failing.

PR #13524, opened 2024-07-30
Friendlier error message if Engine export fails for track

mr-smidge:
Intended to help with issues like https://github.com/mixxxdj/mixxx/issues/13480

PR #13564, opened 2024-08-13
Merge 2.4 to 2.5

acolombier:
Merges the following changes to `2.5`
- #13524 
- #13545 
- #13530


acolombier:
I might have gone a bit too quick with this one... @Swiftb0y should I close that PR?

Swiftb0y:
I'm afraid I can't follow you? Whats wrong with which PR?

acolombier:
This is a follow up with #13530, however I took all the changes from 2.4 for 2.5, instead of cherry picking just my fix, so perhaps it would have been better? If we are happy with a full sync, then all good!

Swiftb0y:
Ah, well we also have the 2.5 branch, so usually the workflow is 2.4->2.5->main. I'll take care of merging locally and push directly.

Swiftb0y:
fyi, I'm currently fighting with https://github.com/DavidAnson/markdownlint/issues/1328 So merging to the newer branches will have to wait until we have resolved that (or we downgrade markdownlint again for now).

Swiftb0y:
done. though CI still fails [because of codespell](https://github.com/mixxxdj/mixxx/pull/13567)...

PR #13569, opened 2024-08-13
chore: fix shellcheck SC2319 warning

Swiftb0y:
pretty sure this is the last fix needed to get CI to be happy again. I wonder why these issues were caught in the PR?


Swiftb0y:
Thank you

Issue #11371, opened 2023-03-15
Common Mixer `loaded_chain_preset` CO


Swiftb0y:
### Feature Description

See https://github.com/mixxxdj/mixxx/pull/4027#discussion_r1136877275

It would make sense for to have a single CO that changes the `loaded_chain_preset` for all Channels of the mixer since that workflow is common on most controllers.

PR #13572, opened 2024-08-14
Bundle ControlObject bool members into QFlag

Swiftb0y:
### Motivation:
the following code is getting too overcomplicated and inflexible:
```cpp
ControlDoublePrivate(const ConfigKey& key,
        ControlObject* pCreatorCO,
        bool bIgnoreNops,
        bool bTrack,
        bool bPersist,
        double defaultValue)
```
This PR tries to address that by bundling the enum parameters into some QFlags.
Once completed, this should slightly simplify the CO APIs by reducing the number of parameters. 

This is a prerequisite for a future feature PR that is supposed to address #11371. I was not comfortable adding to this mess before cleaning it up.

Merge conflicts with #13354 should be minimal and I will do my best to take care of them once they arise.   
I submitted this PR early to collect feedback on whether this is going into the right direction before I start propagating this further outwards into the higher level APIs.


daschuer:
Good idea. This probably also had a minimal performance gain. There is just a huge regression risk when introducing it everywhere.

Swiftb0y:
How do you think this would regress exactly?

Swiftb0y:
Fair, but those should be caught during review. I think doing this refactor with minor regression risk is fine since it'll receive a full beta period of testing.

daschuer:
The last commit is out of scope of the PR description.
Let's move it to a separate PR.

daschuer:
And split it up in separate commits, just in case we need to besect this heard of Mixxx

Swiftb0y:
well, its used in a follow up commit, but I can factor it out.

Swiftb0y:
This is ready, though I'm not 100% happy because its often the case that we want to specify a couple parameters during construction, but in order to specify the later parameters, we also need to explicitly mention the former ones. This is ugly and not scalable. Instead we need some kind of builder pattern that lets us specify how to configure a controlobject in arbitrary order. I will have to do some research on how to do that efficiently in C++.

daschuer:
We may introduce a constructor that takes a struct as aggregate with designated initializers.

Swiftb0y:
That does sound like a reasonable solution. I'm just not sure if appleclang is recent enough. I'll look into it soon.

Swiftb0y:
actually, [designed initializers work perfectly](https://compiler-explorer.com/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TEAstcncAGQImbAA5PwAjbFIpAA5yAAd0JWIXJi9ff0Dg9MznIXDImLZ4xJ4Uh2wnbJEiFlIiXL8AngBme2xHEqYmlqIy6LiE5Ptm1vb87qnh0Yqq5IBKe3QfUlROLlQjJSUAagBBI7MAdhstM6PUn1j6AlQQMxujj6OlIlIfJyOAPKpIjHS7XW6fT4RIhHFjnLoAESOWjMXXBkMh0KOsXhSJ4qPRn0uCIJbwhpwgGCY30BwKUlgAbEcIsRVkcQLCICyiAA6FjrbFcpjEHmxNlg4mk953UgEABuLCInDJmOEsKl5KxsQ1krRZLJ%2BxYhyOVnZdweT1Qp3OVxV5sez1e0o%2B31%2B/wRCXl2EwZvuDqtQJBNsJGOZaqtqLxGshupDJspQhpHtlcu9jLDrLNJyFrPIR1QOd5qHFtouJL10tSKcVyudYZh5grJ1j%2BpuZzYLAA1ths1qS3HSNgiFsmBSwaLcdjiasdWXWydTR3u1YhQ3%2B3aPoPh6RRyvxxHEfnp7OSTcyYuu9grBZV7D13WtyP4xKrnzJyxJWWZ03JWeblx1nobgAFZ%2BACXh%2BHQbgPFsWwvk2bZsHOCwuj4cgiG0AD1k7EBgOCICuGkfg2Fw4JwJ0cgoK4fglBAYIMK4HR1jgWAUAwHB8GIMhKGoOhGFYDhuDQwRhDECROBkORhGUNRNEYvRAkMYxTFg6xen6bI3CYTxvA6RSwgiMZKgmLpgLSDIsiEWYAgsczimyJZxkSUz1PqAYhhmXT8lsuoGiEDyRkM5YTLM75hmsxSwtaRzjOc4D1iUBCdm4V0/hhQNQVtOssThSNkWjKE1RxPL8R/OcfxuVL3U9VMfQ5P1LVpIMJWy8NJ1K8EWz/E5DWNM4WtuBrHQ3CkcrzPszRYCB%2BTzWIIDFYMusG6slSdTU1Q/JtVRhbUyvLTqbl645TXqi1nmtAaPiGl4RpXMb62xcbwzZDlsxYWaBQLYtFvKwkq3lGs1u2o89p1NsjiXHtbwWy7PkfHcKWCbFv06uduovZdb2%2B2HNyHJ8VyRng82LE95wxq8bxy%2B9yXh3dprzImjh4FH9TRk4AMMECwMwyjoNU01Eq2HZkK6Cx%2BAYpj1iQbAWBwRIIEA7giPIEi8PIcjIO4Gi6PQzDVmwkBpAuHkkmkYCAE5gK6LQtAsBkeC0c2Lk5rgum5%2BTeeo3X5OYxAIDY9A2FSBgEh4hMg5DxJDRMHgGRtgQGCVUhaPmnnYgiFoAE8hP4dPWFITOAViXQ3Jz8gMDYDhhABJh6Gzj2cFiHxgA8CR6FoiDyBwDsTEkBuCEHBpUw7ijsHUeofFWzvoT6HmnliUgs68HAeZ%2BAgSM71NSFiDJsA9HvgCeEw9YEIxgCUAA1AhsAAdyBZgy5E0RxEkSSn5kjQef0ImjGP8xrFsIYAgsRaKQHWOgYE2QO4AFoARu0olvWUOBQEK1cn5AIEB3ARSJgZcoTkDBFEsjkLynQ7JEJiisImvl3LTDaCQgw1DGi0IoRMKhtDsELGikFfBzMNhCwkorLgoF1Y8yokcdQSQGTQIZNIfMv9gBMwZDyLQyijgQE4iQUgItmbixPv7NAgdg6MAoFQcORiJjR2ALHeOfEk4p1iGnDOBcy55yzkXEuTgy4VyrkQGudceaN2bq3eg7cy7d2Un3CihBB7OGHjzMeE8p5oRngRCi89F4F2XrsCia8N5oS3jvFQ%2B8IkRFAD7U%2BLBz5X1vvfcCwl5BiVfrId%2BKhP4e30LZeRaB%2BZAJAfAcBkDEzcFgfA9AiCCDIP6WggYWkdJ5FIbgoylCyEDA4YQgYLDqjTKYeFehVC%2BhuR2VwvBsUGHsL2ZwwKJzKEJSSpwCwgjhEa09uIyR0jZGWMUco1R6jCCaJFg83R5T9GEBoDQMOT9GkSWafID%2Bck0lIFoj/RFoKaBEEzqkTg%2BFEVjx%2BCwCBMyvakERYpYlShUXosxTrMluLF4Es0tRR57sKJUQRAQMFRxqk3wSK8qRMi5HKS%2BSorQRw5THAkXyj58ihXKKBZLcg0tZYTFQThYCFgeQ8GAtILQ0gGTmydl0JI5sdUu2ViRB2ZFRFa3sIpCWWEXZixER7Kicr7Vb0yK4aQQA) but they need clang 16 to work properly. that is only shipped in xcode 16 which is currently still in beta... :vomiting_face:

Swiftb0y:
I've not gone through and converted all of the sensible parts to QFlags, but actually for designated initializers separate bools will make more sense instead. So I'll probably end up reverting that instead at least in the struct used for initialization.

Swiftb0y:
Closing as I don't think this is a good way forward.

PR #13574, opened 2024-08-16
Modernize `ControlValueAtomic` 

Swiftb0y:
As requested in https://github.com/mixxxdj/mixxx/pull/13572#issuecomment-2291560589

### DIFF
--- a/src/control/controlvalue.h
+++ b/src/control/controlvalue.h
@@ -1,12 +1,8 @@
 #pragma once
 
-#include <QAtomicInt>
-#include <QObject>
 #include <atomic>
 #include <bit>
 #include <limits>
-
-#include "util/compatibility/qatomic.h"
 
 // for lock free access, this value has to be >= the number of value using threads
 // value must be a fraction of an integer
@@ -36,26 +32,27 @@ class ControlRingValue {
     // slot, because the stored value is preserved.
     bool tryGet(T* value) const {
         // Read while consuming one readerSlot
-        if (m_readerSlots.fetchAndAddAcquire(-1) > 0) {
+        if (m_readerSlots.fetch_sub(1, std::memory_order_acquire) > 0) {
             // Reader slot has been acquired, no writer is active
             *value = m_value;
-            m_readerSlots.fetchAndAddRelease(1);
+            m_readerSlots.fetch_add(1, std::memory_order_release);
             // We need the early return here to make the compiler
             // aware that *value is initialised in the true case.
             return true;
         }
-        m_readerSlots.fetchAndAddRelease(1);
+        m_readerSlots.fetch_add(1, std::memory_order_release);
         return false;
     }
 
     bool trySet(const T& value) {
         // try to lock this element entirely for reading
-        if (m_readerSlots.testAndSetAcquire(kMaxReaderSlots, 0)) {
+        int expected = kMaxReaderSlots;
+        if (m_readerSlots.compare_exchange_strong(expected, 0, std::memory_order_acquire)) {
             m_value = value;
             // We need to re-add kMaxReaderSlots instead of storing it
             // to keep the balance if readers have decreased the number
             // of slots in the meantime!
-            m_readerSlots.fetchAndAddRelease(kMaxReaderSlots);
+            m_readerSlots.fetch_add(kMaxReaderSlots, std::memory_order_release);
             return true;
         }
         return false;
@@ -63,7 +60,7 @@ class ControlRingValue {
 
   private:
     T m_value;
-    mutable QAtomicInt m_readerSlots;
+    mutable std::atomic<int> m_readerSlots;
 };
 
 // Ring buffer based implementation for all Types sizeof(T) > sizeof(void*)
@@ -80,7 +77,9 @@ class ControlValueAtomicBase {
   public:
     inline T getValue() const {
         T value;
-        unsigned int index = static_cast<unsigned int>(atomicLoadRelaxed(m_readIndex)) % cRingSize;
+        unsigned int index = static_cast<unsigned int>(m_readIndex.load(
+                                     std::memory_order_relaxed)) %
+                cRingSize;
         while (!m_ring[index].tryGet(&value)) {
             // We are here if
             // 1) there are more then kMaxReaderSlots reader (get) reading the same value or
@@ -100,7 +99,9 @@ class ControlValueAtomicBase {
         // This test is const and will be mad only at compile time
         unsigned int index;
         do {
-            index = static_cast<unsigned int>(m_writeIndex.fetchAndAddAcquire(1)) % cRingSize;
+            index = static_cast<unsigned int>(m_writeIndex.fetch_add(
+                            1, std::memory_order_acquire)) %
+                    cRingSize;
             // This will be repeated if the value is locked
             // 1) by another writer writing at the same time or
             // 2) a delayed reader is still blocking the formerly current value
@@ -117,8 +118,8 @@ class ControlValueAtomicBase {
     // In worst case, each reader can consume a reader slot from a different ring element.
     // In this case there is still one ring element available for writing.
     ControlRingValue<T> m_ring[cRingSize];
-    QAtomicInt m_readIndex;
-    QAtomicInt m_writeIndex;
+    std::atomic<int> m_readIndex;
+    std::atomic<int> m_writeIndex;
 };
 
 // Specialized template for types that are deemed to be atomic on the target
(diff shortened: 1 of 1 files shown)


src/control/controlvalue.h
```diff
@@ -152,13 +153,10 @@ class ControlValueAtomicBase<T, cRingSize, true> {
...
-// sizeof(T) to determine which underlying implementation of
-// ControlValueAtomicBase to use. For types where sizeof(T) <= sizeof(void*),
-// the specialized implementation of ControlValueAtomicBase for types that are
-// atomic on the architecture is used.
-template <typename T, int cRingSize = kDefaultRingSize>
-class ControlValueAtomic : public ControlValueAtomicBase<T, cRingSize, sizeof(T) <= sizeof(void*)> {
+template<typename T, int cRingSize = kDefaultRingSize>
+class ControlValueAtomic : public ControlValueAtomicBase<T,
+                                   cRingSize,
+                                   std::atomic<T>::is_always_lock_free> {
```
daschuer, line 154: We are now depending on the std library implementation. Can you add a static assert that confirms our assumption, just in case? 
`static_assert((std::atomic<T>::is_always_lock_free == true) || (sizeof(T) > sizeof(void*))`

src/control/controlvalue.h
```diff
@@ -118,14 +113,14 @@ class ControlValueAtomicBase {
     // In worst case, each reader can consume a reader slot from a different ring element.
     // In this case there is still one ring element available for writing.
     ControlRingValue<T> m_ring[cRingSize];
-    std::atomic<int> m_readIndex;
-    std::atomic<int> m_writeIndex;
+    std::atomic<std::size_t> m_readIndex;
+    std::atomic<std::size_t> m_writeIndex;
```
daschuer, line 125: This doubles the size of the indexes in case of 64 bit archs and non atomic readable data types for no reason.
How about using `unsigned int`?  Even a `uint8_t` would be sufficient if, we are sure it `is_always_lock_free`
Please static_assert the later.

daschuer (commented):
Thank you, left some comments.

daschuer (changes requested):
.

src/control/controlvalue.h
Swiftb0y: I'm afraid I can't quite follow you. The static assert you proposed doesn't make much sense to me logically because we if its not `always_lock_free`, why would it be when `T` is larger than `void*`?

src/control/controlvalue.h
Swiftb0y: yup, `uint8_t` also seems fine for me because I doubt there there would ever need to be a ring size that large. Using types smaller than native though (that includes `int` on x64_86) [do incur an extra `mov` instruction though](https://compiler-explorer.com/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:12,positionColumn:2,positionLineNumber:12,selectionStartColumn:2,selectionStartLineNumber:12,startColumn:2,startLineNumber:12),source:'%23include+%3Ccstdint%3E%0A%23include+%3Ccstddef%3E%0A%0Aextern+int+arr%5B512%5D%3B%0A%0Aint+getArr(int+i)+%7B%0A++++return+arr%5Bi%5D%3B%0A%7D%0A%0Aint+getArr(unsigned+int+i)+%7B%0A++++return+arr%5Bi%5D%3B%0A%7D%0A%0Aint+getArr(long+int+i)+%7B%0A++++return+arr%5Bi%5D%3B%0A%7D%0A%0Aint+getArr(std::size_t+i)+%7B%0A++++return+arr%5Bi%5D%3B%0A%7D'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g142,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-Wall+-Wconversion+-O3',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+14.2+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+14.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+14.2+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

daschuer: Interesting. OK, than let's keep std::size_t for the sake of the fastest solution.

src/control/controlvalue.h
daschuer: Let's imagine an incomplete library implementation where a certain type is not flagged as "is_always_lock_free" even though it is smaller than void* and a solution should be possible. 
In this case the first part is false and the assertion is violated if the type is smaller than "void*". In such a case the old code would have used the aligned value and the new implementation uses the ring.

src/control/controlvalue.h
daschuer: But please add an assertion for is_always_lock_free

Swiftb0y: well, I've already made the change in a fixup, so we can keep it or drop it. I'm not to worried about a single `mov` instruction between registers (I haven't looked very deeply into it but I can't take more than a cycle). Shrinking the object might be a more significant speedup because it decreases cache pressure... I don't know. I'm personally not too worried about optimizing here because from looking at the rest of the codebase the eventually-consistent specialization is only used very rarely anyways... 
Two potentially interesting godbolts:
1. Due to padding, `std::size_t` vs `std::uint8_t` does barely make a difference. If it all, only for small types for which the ControlRingValue never gets instantiated. https://compiler-explorer.com/z/KcEEf3Ko7  
2. `is_always_lock_free` is also the case for types larger than `void*` on ARMv8 on clang: https://compiler-explorer.com/z/K57d1d5MP

In conclusion, we should keep `std::size_t` and (related to your other comment) rely on `is_always_lock_free`.

src/control/controlvalue.h
Swiftb0y: right, but on that same hypothetical architecture where types smaller or equal than `void*` were not lock free, the assumption that aligned access is always atomic may be just as wrong, in which case we do want to use the ring. Besides, I don't think such an architecture exists right now (at least it does not seem to be the case on x86, arm or RISC-V).

If we want a solution that still performs optimally at runtime depending on the target, we could create a wrapper that switches its type depending on the runtime [`std::atomic<T>::is_lock_free()`](https://en.cppreference.com/w/cpp/atomic/atomic/is_lock_free) function. 

In general I don't think we should mistrust the standard library because the probability that it is incomplete is much lower than the probability that we make a mistake and the additional complexity is not worth it. The established std implementations are used and tested by so many people and companies that have a close eye on its completeness and correctness that its fine to rely on that.

daschuer: > right, but on that same hypothetical architecture where types smaller or equal than void* were not lock free, the assumption that aligned access is always atomic may be just as wrong,

Yes. In such a case the assertion is violated for a good reason and we can fix it by a `reinterpret_cast` to a lock free type or using a memory region where atomic access is possible. 

> If we want a solution that still performs optimally at runtime depending on the target, we could create a wrapper that switches its type depending on the runtime [std::atomic<T>::is_lock_free()](https://en.cppreference.com/w/cpp/atomic/atomic/is_lock_free) function.

Yes, the proposed assertion gives us just the evidence that there is such a problem to solve. 

> In general I don't think we should mistrust the standard library 

Better same than sorry. At least you find some related discussions: 
https://stackoverflow.com/questions/30048533/why-isnt-atomic-double-fully-implemented
https://github.com/WebAssembly/threads/issues/7

We are discussion here about a single compile time check, that is impossible to trigger from our current knowledge.

Swiftb0y: sure I can add that check so we get alerted to such a case, but I don't think we should leverage implicit assumptions to fix them (the reinterpret_cast or manually layout out the memory) and instead use std-library tools. I'll add the assertion.

Swiftb0y: done

daschuer (approved):

daschuer:
Thank you. Looks nice!

Swiftb0y:
Thank you.

PR #13575, opened 2024-08-16
(fix) Playlists: paint sidebar items bold immediateley after dropping tracks

ronso0:
... like it already works for crates.

Currently playlists are updated only after selection in the sidebar changes.


daschuer (approved):
LGTM, Thank you.

PR #13577, opened 2024-08-17
DlgPrefSound: Add missing ifdefs for building without Rubberband

fwcd:
This fixes a small regression that caused builds without Rubberband to fail.


JoergAtGithub:
LGTM! Thank you!

ronso0:
Thanks for the cleanup!

PR #13581, opened 2024-08-19
more `ControlDoublePrivate` optimization

Swiftb0y:
Taking the cleanup parts from #13572 and adding a little extra (saving another 9 bytes).

### DIFF
--- a/src/control/control.h
+++ b/src/control/control.h
@@ -230,14 +230,8 @@ class ControlDoublePrivateConst : public ControlDoublePrivate {
   public:
     ~ControlDoublePrivateConst() override = default;
 
-    void setInner(double value, QObject* pSender) override {
-        Q_UNUSED(value)
-        Q_UNUSED(pSender)
+  private:
+    void setInner(double, QObject*) override {
         DEBUG_ASSERT(!"Trying to modify a default constructed (const) control object");
     };
-
-  protected:
-    ControlDoublePrivateConst() = default;
-
-    friend ControlDoublePrivate;
 };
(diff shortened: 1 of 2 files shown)


src/control/control.h
```diff
@@ -229,14 +230,8 @@ class ControlDoublePrivateConst : public ControlDoublePrivate {
   public:
     ~ControlDoublePrivateConst() override = default;
 
-    void setInner(double value, QObject* pSender) override {
-        Q_UNUSED(value)
-        Q_UNUSED(pSender)
+  private:
+    void setInner(double, QObject*) override {
         DEBUG_ASSERT(!"Trying to modify a default constructed (const) control object");
     };
```
Swiftb0y, line 236: adhering to [cpp core guidelines "F.9: Unused parameters should be unnamed"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-unused). I personally prefer that because it does not have the risk where a parameter is marked with `Q_UNUSED` but still used in the body somewhere later.

src/control/control.h
```diff
@@ -198,14 +201,12 @@ class ControlDoublePrivate : public QObject {
 
     QAtomicPointer<ControlObject> m_pCreatorCO;
 
+    // name of the key to track using stats framework, unless the trackKey isNull().
     QString m_trackKey;
```
JoergAtGithub, line 205:
```suggestion
    // name of the key to track using stats framework, unless the trackKey isNull().
    QString m_trackingKey;
```
I think this will help readability, because it would make it immediately visible, that no music track is meant here.

JoergAtGithub (approved):

src/control/control.h
Swiftb0y: Done

Swiftb0y:
Thanks for the review

Swiftb0y:
merge?

Holzhaus (approved):

Swiftb0y:
Thanks

PR #13582, opened 2024-08-19
Refactor/designated init control object

Swiftb0y:
As suggested in https://github.com/mixxxdj/mixxx/pull/13572#issuecomment-2295308117 This PR explores using a struct along with designated initialization for a form of named parameters. 

This PR is based on #13581. 

One thing I already noticed that using inheritance for easy reuse and minimizing boilerplate does not work perfectly with designated initializers. Ideally there was something like [P2287](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2287r2.html) but for now we'll have to sometimes fall back on plain list initialization.

Let me know what you think.


src/control/control.cpp
```diff
@@ -43,32 +43,25 @@ QWeakPointer<ControlDoublePrivate> s_pDefaultCO;
...
-          m_kbdRepeatable(false) {
-    if (bPersist) {
+        : ControlDoublePrivate({{}, true}){};
+
+ControlDoublePrivate::ControlDoublePrivate(const ParametersWithConfirm& params)
+        : m_key(params.key),
+          m_pBehavior(params.behavior),
+          m_name(params.name),
+          m_description(params.description),
+          m_value(params.defaultValue),
```
daschuer, line 53: This is re-initialized below. Maybe add a free helper function that does it on one go.

src/control/control.h
```diff
@@ -36,6 +36,23 @@ class ControlDoublePrivate : public QObject {
     // TODO: don't expose this implementation detail
     constexpr static double kDefaultValue = 0.0;
 
+    struct Parameters {
+        ConfigKey key;
+        ControlObject* pCreatorCO = nullptr;
+        QString name{};
+        QString description{};
```
daschuer, line 43: {} is redundant. 
```suggestion
        QString name;
        QString description;
```

src/control/control.h
```diff
@@ -36,6 +36,23 @@ class ControlDoublePrivate : public QObject {
...
+        ControlNumericBehavior* behavior = nullptr;
+        double defaultValue = kDefaultValue;
+        bool ignoreNops = true;
+        bool track = false;
+        bool persist = false;
+        bool keyboardRepeatable = false;
+    };
+
+    struct ParametersWithFlags : public Parameters {
```
daschuer, line 52: I have not yet fully understand why we have two inherited types. Maybe we can consider to merge them in a next step? Maybe we can even make some more booleans flags.

src/control/control.h
Swiftb0y: Technically yes, though adding it always simplifies the language IMO since the reader would not have to know whether a typename is a primitive or a class type. The {} initializes both.

src/control/control.h
Swiftb0y: because the downstream APIs want to take more parameters. Taking those parameters as separate function parameters would defeat the purpose again (taking a single struct so they are ordering independent and can act as named arguments). Merging all the downstream options into the upstream (parent/base) struct would make the API confusing because the caller can now pass options which don't make sense in the context of the upstream API. 
Merging these downstream options with the upstream parameters would violate "separation of concerns" and the "open-closed principle". 

Does it help if you interpret the public inheritance simply as a set union operation ( eg, `ParametersWithFlags` $=$`Parameters` $\cup$ `Flags`)?

src/control/control.cpp
Swiftb0y: are you referring to the `m_value.setValue`? What value should we initialize with if the config is null? just `defaultValue`?

src/control/control.h
Swiftb0y: Rereading this and addressing your original question:
we have `ParametersWithConfirm` because that what the private `ControlDoublePrivate` ctor needs and we have `ParametersWithFlags` because thats what `ControlDoublePrivate::getControl` wants. These are separate because `ParametersWithConfirm` doesn't care about the extra flags in `ParametersWithFlags` while the public `ParametersWithFlags` should not expose the internal `confirmRequired` of `ParametersWithConfirm`. Merging these two doesn't make sense because they're supposed to be used in different contexts.

src/control/control.cpp
Swiftb0y: done in fixup

src/control/control.h
daschuer: Following this rule we will clutter the whole Mixxx code with many of these {}. Since now we always remove redundant initializes in code reviews. Let's stick to it for now.

Swiftb0y: I'm not saying we should apply to all of mixxx, but I think it would make sense for aggregate classes. Its also only redundant in the sense that its explicit, it doesn't result in extra instructions (unlike redundant initialization in a ctor body). I would really prefer to keep it at least for aggregates because it clearly says "this member can't ever be uninitialized".

daschuer: I really don't mind in this case. I am just afraid future discussions in other cases. IMHO this does not improve readability, because of the following example:  Let's imagine is an unknown type. Without initialization the value might be uninitialized or default initialized. To see the difference you need to look up the implementation that is out of scope here. So you need to do a lookup in both cases. 

In https://github.com/mixxxdj/mixxx/pull/13586 (Still waiting for merge b.t.w.) we had: 
`mixxx::audio::FramePos startPosition`
vs:
`mixxx::audio::FramePos startPosition = mixxx::audio::kInvalidFramePos;`
the later avoids the lookup. 

using here 
`mixxx::audio::FramePos startPosition{};` 
will no longer inform the reader that the position is initialized as invalid.

Swiftb0y: > Without initialization the value might be uninitialized or default initialized.

Right, using `type member{}` explicitly alleviates that.

> In https://github.com/mixxxdj/mixxx/pull/13586 (Still waiting for merge b.t.w.) we had:

Yes, I actually prefer the latter, I just didn't want to argue on that PR. 

> using here
mixxx::audio::FramePos startPosition{};
will no longer inform the reader that the position is initialized as invalid.


But neither does `mixxx::audio::FramePos startPosition;`...

I think these are two different cases. In the case in this PR I just say "I want this default initialized" explicitly. In the `FramePos` example you brought up, we rather want to say "I want this framepos to be initialized with a known constant value". In the framepos example, incidentally the default ctor sets it to the same value sure, so you can write it in both ways, but one of them is more expressive than the other. 

Either way, we should not introduce rules to always do X (such as "we don't initialize default constructed classes") because that removes the ability for the code author to appropriately express their intent when writing the code.

PR #13583, opened 2024-08-20
(fix/followup) Tracks: apply played/missing text color also to selected tracks

ronso0:
During performance I find it quite confusing that when selecting played tracks they are marked like unplayed tracks.

### Fix
I picked the **DefaultDelegate** from #11644 and tweaked some paint methods.
Works great, also for the star editor.

As a bycatch I managed to simplify some of them (see StarEditor for example, and removed QColorGroup setting).

<sub>Only remaining very minor issue: if the editor is active while the played state changes, the polygon color will not be updated immediately (though that's a confusing situation anyway since changed index (row) data will trigger a paint event which resets the pending (unsaved) rating anyway.
The style of hovered star editors when changing row selection is updated correctly as before.</sub>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -945,6 +945,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/library/dao/settingsdao.cpp
   src/library/dao/trackdao.cpp
   src/library/dao/trackschema.cpp
+  src/library/tabledelegates/defaultdelegate.cpp
   src/library/dlgcoverartfullsize.cpp
   src/library/dlgcoverartfullsize.ui
   src/library/dlgtagfetcher.cpp
--- a/res/skins/Deere/style.qss
+++ b/res/skins/Deere/style.qss
@@ -242,9 +242,22 @@ WTrackTableView::item:selected,
 WLibrarySidebar::item:selected,
 #LibraryBPMButton::item:selected,
 #LibraryPlayedCheckbox::item:selected {
-  color: #D6D6D6;
+  /* disabled since it interferes with selection-color
+   which we use in DefaultDelegate and TableItemDelegate to
+   blend the highlight color with the played/missing color.
+  color: #D6D6D6; */
+  /* This is not required for painting the background,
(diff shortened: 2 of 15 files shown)


ronso0:
Okay, looks as expected in all skins.

I wonder: are the existing 'played' colors okay in all skins?

ronso0:
This is ready IMO.

A general issue remains however:
with bright track colors, the dim 'played' color might be unreadable.
**Possible fixes:**
1. make text brighter until a certain contrast is reached, try to keep it darker than the regular text color?
not sure if this is even possible to get a color that is still distinct enough to recognise the tracks as 'played'...
2. switch to dark color that provides enough contrast
   * use a fixed secondary `trackPlayedColor`
   * determine the color based on the background to provide just enough contrast

daschuer:
Unfortunately you cannot read the text anymore in Deere and Shade: 
grafik
grafik
Tango is at the edge IMHO: 
grafik

daschuer (changes requested):
The colors need to be tweaked.

ronso0:
Right, colors are not suitable. Will try to fix them.
With new `DefaultDelegate` this might be a candidate for 2.6 anyway?

ronso0:
I'll try a different approach:
for played + selected: blend selected font color and background color
That should get us enough contrast for
* selected bg / selected text (unplayed) as well as
* selected text / selected text (played)

ronso0:
TODO:
fix text for 'inactive' palette, i.e. another window is in front (e.g. Track Info)

ronso0:
The played/selected colors are now blended 50/50.
Looks good in all skins IMO
image 
image
image
image
image
We could also allow skins/schemes to set their own custom ratio :shrug:

daschuer (approved):
LGTM and the visibility issue is fixed. Thank you.

ronso0:
Thanks for your review.

I think this still needs some tweaking, but that doesn't have to happen for 2.5.1 to not hold that up.
The font/bg contrasts should be increased
* in Deere: brighter selected text, white as in other skins
* in Shade Classic: dim selected bg a bit

PR #13588, opened 2024-08-21
Denon MC7000: Fix star up/down logic by only handling button down events

### LINKED ISSUE #13587: Star up/down in MC7000 mapping is broken
### Bug Description

The MC7000 mapping applies a star up/down on button down _and_ button up, resulting in two stars being applied or removed per press:

https://github.com/user-attachments/assets/10088abc-7d94-4913-87de-e6adc370d9ae

I suspect this to be a regression from #4729, since that PR last refactored this logic.

### Version

2.5

### OS

macOS 14.5

fwcd:
### Fixes #13587 

This fixes what looks like a typo, making sure that the logic in `StarsUp` and `StarsDown` only executes when a button is pressed down and not when released.


daschuer (approved):
LGTM, thank you.

PR #4729, opened 2022-04-20
Mc7000 23

holopansel:
start and stop with softstart and break function, time configurable with stop time knob,
pitch play,
sampler with up to 32 samplers,
velocity sampler with up to 32 samplers.

PR #4310, opened 2021-09-19
[WIP]: Denon mc7000 parameter buttons

flosse:
**Motivation**:
Currently the parameter buttons modify the stars of a track or (if shifted) it changes the color.
I never used this functionality and if I would, I'd do that at home at the laptop and not at stage
whereas during a live performance I really like to have quick access to buttons for beat jumps.

If somebody really like to keep the old behavior I could introduce a boolean flag in the script,
otherwise I'd suggest do map this as default behavior.

**Note**:
This is based on #4307 but if you like me to rebase it to `2.3` branch just tell me :)

PR #13589, opened 2024-08-22
Denon MC7000: Unify parameter button logic and add customizable modes

fwcd:
This is a rebased version of #4310, which aims to refactor the logic for the customizable parameter buttons on the Denon MC7000. Currently, the buttons change the star rating (track color if shifted), but some users rarely change these and would rather wish to map them differently.

Fortunately, the new controller settings feature lets us have our cake and it eat it too. This PR unifies the parameter button, refactors the code to be more concise/readable and adds three new parameter button modes:

| Mode | Description |
| - | - |
| `Change Star Rating (Shift: Track Color)` | The current and default mode |
| `Perform Beat Jump (Shift: Adjust Jump Size)` | The logic from the previous PR |
| `Set Intro/Outro (Shift: Clear Intro/Outro)` | A new mode that sets/clears intro-end/outro-start |

image

This PR also lets the user customize whether to override the parameter button mode with changing the pitch range. This is the default behavior as of #4729, which might be surprising to users that change the mode, hence the ability to disable it.

### DIFF
--- a/res/controllers/Denon-MC7000-scripts.js
+++ b/res/controllers/Denon-MC7000-scripts.js
@@ -989,6 +989,13 @@ MC7000.parameterButton = function(value, group, {isLeftButton, isShiftPressed}) 
                 script.triggerControl(group, `beatjump_${isLeftButton ? "backward" : "forward"}`);
             }
             break;
+        case "introOutro":
+            {
+                const cue = isLeftButton ? "intro_end" : "outro_start";
+                const action = isShiftPressed ? "clear" : "activate";
+                script.triggerControl(group, `${cue}_${action}`);
+            }
+            break;
         default:
             break;
         }
--- a/res/controllers/Denon-MC7000-scripts.js
+++ b/res/controllers/Denon-MC7000-scripts.js
@@ -110,6 +110,11 @@ MC7000.jogParams = {
     }
 };
 
+// Parameter button mode. See Denon-MC7000.midi.xml for documentation of the available modes.
+MC7000.parameterButtonSettings = {
+    mode: engine.getSetting("parameterButtonMode") || "starsAndColor",
+};
+
 /*/////////////////////////////////
 //      USER VARIABLES END       //
 /////////////////////////////////*/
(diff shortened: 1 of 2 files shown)


res/controllers/Denon-MC7000-scripts.js
```diff
@@ -952,35 +959,87 @@ MC7000.censor = function(channel, control, value, status, group) {
         }
     }
 };
-// Param Button for Pitch Play to decrease pitch, or decrease star rating otherwise
-MC7000.StarsDown = function(channel, control, value, status, group) {
-    const deckNumber = script.deckFromGroup(group);
-    const deckIndex = deckNumber - 1;
-    if (value >= 0x00) {
```
fwcd, line 959: Small implementation note: This will create a conflict on the next 2.4 -> 2.5 merge, assuming #13588 (a small fix to this logic) is merged. This is expected and can safely be discarded, the new logic already handles button down events correctly.

res/controllers/Denon-MC7000-scripts.js
```diff
@@ -952,35 +959,87 @@ MC7000.censor = function(channel, control, value, status, group) {
...
-        if (MC7000.PADMode[deckIndex] === "Pitch") {
-            for (let padIdx = 0; padIdx < 8; padIdx++) {
-                MC7000.halftoneToPadMap[deckIndex][padIdx] = MC7000.halftoneToPadMap[deckIndex][padIdx] - 8; // pitch down
-            }
-        } else {
-            engine.setValue(group, "stars_down", true); // stars down
-        }
+
+// Parameter Buttons
+MC7000.parameterButton = function(value, group, {isLeftButton, isShiftPressed}) {
```
Swiftb0y, line 964: I'm not 100% confident that destructuring in the parameter works in the QJSEngine JS version. I assume you have this controller and tested this?

Swiftb0y (commented):
LGTM otherwise

res/controllers/Denon-MC7000-scripts.js
fwcd: That's a good point, I was a bit unsure about the `??` operator too (which we need to handle a boolean default, because using `||` results in different semantics), which needs ES 2020. At least in Qt 6 it seems to work fine though, I could do a quick test of the CI build too to make sure I didn't accidentally use a newer Qt 6 locally.

Admittedly, I haven't checked Qt 5, do we care about breaking this? Rewriting both the null coalescing and destructuring shouldn't be too hard, just slightly more verbose.

Swiftb0y: > That's a good point, I was a bit unsure about the `??` operator too

Thats fine, IIRC that was implemented as a GSoC project back then (Qt 5.15 IIRC) (optional chaining with Qt 6.2). Pretty sure that QJSEngine did not receive significant changes since then. I have a small qjsengine repl I can use to check this if we wanna be sure.

fwcd: Just tested the CI build, which uses Qt 6.5.3, and it least there it seems to work fine, so I assume we should be fine, but feel free to test it in your REPL too!

fwcd:
The branch is now rebased onto the latest 2.5 and the conflict from #13588 is fixed.

res/controllers/Denon-MC7000.midi.xml
```diff
@@ -2123,23 +2151,23 @@
...
+                <group>[Channel4]</group>
+                <key>MC7000.parameterButtonLeftShifted</key>
                 <description></description>
                 <status>0x97</status>
                 <midino>0x2A</midino>
                 <options>
-                    <normal/>
+                    <script-binding/>
                 </options>
             </control>
```
Swiftb0y, line 2162: modelling a shifted button as a separate component has the issue that if you press the button, press and hold shift and then release the button, the controller usually does only creates a release event for the shifted button. So the unshifted button stays on. ComponentsJS has some code that works around this but for that the `shift` and `unshift` mechanic needs to be used (see `ComponentContainer.shift` implementation for details). Can you test whether thats a problem here?

fwcd: I don't think this should be an issue since we only handle button down events anyway:

```js
if (value === 0) {
    return;
}
```

Swiftb0y: > I don't think this should be an issue since we only handle button down events anyway:

For these controls yes, though it does make a difference for the hotcue pads for example. Thanks for double checking though.

Swiftb0y (approved):
LGTM. thank you. Sorry for having this dragged out this long

PR #13591, opened 2024-08-22
(fix) Playlists: update sidebar when dragging tracks into the playlist table

### LINKED ISSUE #13590: playlist label with count and duration does not update when you add or remove items from the playlist
### Bug Description

In the playlist view, when you add or remove tracks from a given playlist, the label (which displays the name, quantity, and duration) on the left-hand side does not update. It only updates when you click on a different playlist and then back into the one you were working on.

### Version

2.4.1 (HEAD)

### OS

Mac 13.5.2

ronso0:
Fixes #13590


daschuer (approved):
Works good, thank you. LGTM

PR #13594, opened 2024-08-24
Static initalization order fix

daschuer:
I have experienced a crash because the static s_keyColorPalette was initialized by another uninitiated static Palette. 
This PR fixed that.

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -108,8 +108,7 @@ constexpr bool BaseTrackTableModel::kKeyColorsEnabledDefault;
 int BaseTrackTableModel::s_bpmColumnPrecision =
         kBpmColumnPrecisionDefault;
 bool BaseTrackTableModel::s_keyColorsEnabled = kKeyColorsEnabledDefault;
-ColorPalette BaseTrackTableModel::s_keyColorPalette =
-        mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette;
+std::optional<ColorPalette> BaseTrackTableModel::s_keyColorPalette;
 
 // static
 void BaseTrackTableModel::setBpmColumnPrecision(int precision) {
@@ -1055,10 +1054,10 @@ QVariant BaseTrackTableModel::roleValue(
                 return QVariant();
             }
             const auto key = KeyUtils::keyFromNumericValue(keyCode);
-            if (key == mixxx::track::io::key::INVALID) {
+            if (key == mixxx::track::io::key::INVALID || !s_keyColorPalette.has_value()) {
                 return QVariant();
             }
-            return QVariant::fromValue(KeyUtils::keyToColor(key, s_keyColorPalette));
+            return QVariant::fromValue(KeyUtils::keyToColor(key, s_keyColorPalette.value()));
         }
         default:
             return QVariant();
--- a/src/library/basetracktablemodel.h
+++ b/src/library/basetracktablemodel.h
@@ -3,6 +3,7 @@
 #include <QAbstractTableModel>
 #include <QList>
 #include <QPointer>
+#include <optional>
 
 #include "library/columncache.h"
 #include "library/trackmodel.h"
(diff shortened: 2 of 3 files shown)


src/library/basetracktablemodel.h
```diff
@@ -300,7 +300,8 @@ class BaseTrackTableModel : public QAbstractTableModel, public TrackModel {
 
     static int s_bpmColumnPrecision;
     static bool s_keyColorsEnabled;
-    static ColorPalette s_keyColorPalette;
+    // This need to be a pointer to avoid static initialization order issues
+    static std::unique_ptr<ColorPalette> s_pKeyColorPalette;
```
Swiftb0y, line 304: I don't think the problem is static initialization. The only way that can occur if another global static is initialized with `s_pKeyColorPalette` in another translation unit before `s_pKeyColorPalette` is initialized. That can't happen because `s_pKeyColorPalette` is only accessible by `basetracktablemodel.cpp`. The only other way to access it is `BaseTrackTableModel::setKeyColorPalette` and `BaseTrackTableModel::roleValue` both of which do not get called before `main()` and thus `s_pKeyColorPalette` should be initialized at that point. 

I think there is something going on which we don't understand. Can you share the crashing stacktrace? 

Also, even if it was accessed before initialization, changing into to a pointer does not actually fix the associated undefined behavior. The object is still accessed before its initialization, the underlying memory is likely just the same bitpattern as `nullptr` so it doesn't crash in practice, but in theory the compiler could still optimize this into a crash because this relies on UB.

Swiftb0y (changes requested):

Swiftb0y:
Ah, my bad I just now understood the problem. Its not `s_keyColorPalette`, but actually `mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette` (`mixxx::PredefinedColorPalettes` entirely actually...)

src/library/basetracktablemodel.cpp
```diff
@@ -108,8 +108,7 @@ constexpr bool BaseTrackTableModel::kKeyColorsEnabledDefault;
 int BaseTrackTableModel::s_bpmColumnPrecision =
         kBpmColumnPrecisionDefault;
 bool BaseTrackTableModel::s_keyColorsEnabled = kKeyColorsEnabledDefault;
-ColorPalette BaseTrackTableModel::s_keyColorPalette =
-        mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette;
+std::unique_ptr<ColorPalette> BaseTrackTableModel::s_pKeyColorPalette;
```
Swiftb0y, line 111: any reason you chose a pointer over the more value-oriented `std::optional`?

daschuer: I just did not thought about it. A printer type is probably right, because in a static context  the memory allocated in a value type object is leaked. (Only esoteric reasons when using valgrind, because all memory is freed anyway on termination)

Swiftb0y:
another option if course would be to use `static` at block scope and let the compiler do the heavy lifting by initializing it on first use instead.

Swiftb0y:
we may also consider doing the same for PredefinedColorPalettes to make those safer to use.

src/library/basetracktablemodel.cpp
Swiftb0y: I disagree... Why would the memory be leaked? The memory of the inner object has automatic storage duration, no cleanup needed. It can never be leaked if its not never dynamically allocated in the first place.

daschuer: Oh yes you are right. But `std::optional` is for me only a workaround for types without a default constructor. 
I would rather add this instead.

daschuer: Just realized that this will not work, because the colors are always considered as valid. I will use an `std::optional`.

src/library/basetracktablemodel.h
```diff
@@ -300,7 +301,8 @@ class BaseTrackTableModel : public QAbstractTableModel, public TrackModel {
 
     static int s_bpmColumnPrecision;
     static bool s_keyColorsEnabled;
-    static ColorPalette s_keyColorPalette;
+    // This need to be left uninitialized to avoid static initialization order issues
+    static std::optional<ColorPalette> s_keyColorPalette;
```
Swiftb0y, line 305: This comment is misleading IMO. its not uninitialized, it just gets implicitly initialized to its `std::nullopt` state...

Swiftb0y (approved):
LGTM. waiting for pre-commit fix. And do you want to squash the last two commits?

Swiftb0y (approved):
LGTM. Thank you.

PR #13598, opened 2024-08-25
Safe Global Predefined Color Palettes

Swiftb0y:
As suggested (by me) in https://github.com/mixxxdj/mixxx/pull/13594#issuecomment-2308818785 as an alternative to that PR.

Note that in case we manage to make `ColorPalette` constexpr friendly, we can trivially migrate `get` to be constexpr as well without having to change the callers with this approach.

### DIFF
--- a/src/preferences/colorpaletteeditor.cpp
+++ b/src/preferences/colorpaletteeditor.cpp
@@ -146,7 +146,9 @@ void ColorPaletteEditor::initialize(
     m_resetPalette = paletteName;
     QString saveName = paletteName;
 
-    for (const ColorPalette& palette : mixxx::PredefinedColorPalettes::kPalettes) {
+    const auto& kPalettes = mixxx::predefinedcolorpalettes::get();
+
+    for (const ColorPalette& palette : kPalettes.palettes) {
         if (paletteName == palette.getName()) {
             saveName = paletteName + QStringLiteral(" (") + tr("Edited") + QChar(')');
             ColorPaletteSettings colorPaletteSettings(m_pConfig);
@@ -206,7 +208,8 @@ void ColorPaletteEditor::slotRemoveColor() {
 void ColorPaletteEditor::slotPaletteNameChanged(const QString& text) {
     bool bPaletteIsReadOnly = false;
     bool bPaletteExists = false;
-    for (const ColorPalette& palette : mixxx::PredefinedColorPalettes::kPalettes) {
+    const auto& kPalettes = mixxx::predefinedcolorpalettes::get();
+    for (const ColorPalette& palette : kPalettes.palettes) {
         if (text == palette.getName()) {
             bPaletteExists = true;
             bPaletteIsReadOnly = true;
@@ -222,7 +225,7 @@ void ColorPaletteEditor::slotPaletteNameChanged(const QString& text) {
     if (bPaletteExists) {
         if (!m_pModel->isDirty()) {
             bool bPaletteFound = false;
-            for (const ColorPalette& palette : mixxx::PredefinedColorPalettes::kPalettes) {
+            for (const ColorPalette& palette : kPalettes.palettes) {
                 if (text == palette.getName()) {
                     bPaletteFound = true;
                     m_pModel->setColorPalette(palette);
@@ -231,7 +234,7 @@ void ColorPaletteEditor::slotPaletteNameChanged(const QString& text) {
             }
             if (!bPaletteFound) {
                 m_pModel->setColorPalette(colorPaletteSettings.getColorPalette(
-                        text, mixxx::PredefinedColorPalettes::kDefaultHotcueColorPalette));
+                        text, kPalettes.defaultHotcueColorPalette));
             }
         }
     }
@@ -277,7 +280,7 @@ void ColorPaletteEditor::slotResetButtonClicked() {
     ColorPaletteSettings colorPaletteSettings(m_pConfig);
     ColorPalette palette = colorPaletteSettings.getColorPalette(
             m_resetPalette,
-            mixxx::PredefinedColorPalettes::kDefaultHotcueColorPalette);
+            mixxx::predefinedcolorpalettes::get().defaultHotcueColorPalette);
     m_pModel->setColorPalette(palette);
     slotUpdateButtons();
 }
--- a/src/preferences/colorpaletteeditor.cpp
+++ b/src/preferences/colorpaletteeditor.cpp
@@ -12,6 +12,7 @@
 
 #include "moc_colorpaletteeditor.cpp"
 #include "preferences/colorpalettesettings.h"
+#include "util/color/colorpalette.h"
 #include "util/color/predefinedcolorpalettes.h"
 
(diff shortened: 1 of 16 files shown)


src/library/basetracktablemodel.cpp
```diff
@@ -109,7 +109,7 @@ int BaseTrackTableModel::s_bpmColumnPrecision =
         kBpmColumnPrecisionDefault;
 bool BaseTrackTableModel::s_keyColorsEnabled = kKeyColorsEnabledDefault;
 ColorPalette BaseTrackTableModel::s_keyColorPalette =
-        mixxx::PredefinedColorPalettes::kDefaultKeyColorPalette;
+        mixxx::predefinedcolorpalettes::get().defaultKeyColorPalette;
```
daschuer, line 112: In #13594 we have an `std::optional` which makes redundant initialization here obsolete.  
Let's mere #13594 first.

daschuer:
Did you consider to make PredefinedColorPalettes a singleton? This avoids to initalize it over and over again and will avoid the new predefinedcolorpalettes namespace.

Swiftb0y:
It is effectively a singleton. It is only getting initialized once. see [Static block variables](https://en.cppreference.com/w/cpp/language/storage_duration#Static_block_variables)

Swiftb0y:
fyi, an alternative implementation would be to use `const constinit`. That works with qt containers (in theory once they make move operations constexpr). Alternatively we can also make ColorPalettes use `std::vector` and `std::string` internally. Then we can even `constexpr static` construct them. All these solutions avoid the initialization-order-fiasco because the palettes get created at compile time. 

My current favorite is the latter (gutting out the Qt containers in favor of `std` ones) because I don't think the implicit sharing is of much benefit here.

src/preferences/colorpaletteeditor.cpp
```diff
@@ -146,7 +146,9 @@ void ColorPaletteEditor::initialize(
     m_resetPalette = paletteName;
     QString saveName = paletteName;
 
-    for (const ColorPalette& palette : mixxx::PredefinedColorPalettes::kPalettes) {
+    const auto& kPalettes = mixxx::predefinedcolorpalettes::get();
```
daschuer, line 149: A free get() function in a namespace looks weird. This is an indicator that the Namespace is an object.

src/util/color/predefinedcolorpalettes.cpp
```diff
@@ -241,367 +241,333 @@ constexpr mixxx::RgbColor kTritKeyColor10(0x666600);
...
+            kRekordboxHotcueColor10,
+            kRekordboxHotcueColor11,
+            kRekordboxHotcueColor12,
+            kRekordboxHotcueColor13,
+            kRekordboxHotcueColor14,
+            kRekordboxHotcueColor15,
+            kRekordboxHotcueColor16,
+    };
+
+    const static PredefinedColorPalettes kPalettes{
```
daschuer, line 274: Returning a local static is one way ti implement a singleton pattern. How about fully make use of that here? 
as in a `PredefinedColorPalettes::instance()` function.

src/util/color/predefinedcolorpalettes.cpp
```diff
@@ -241,367 +241,333 @@ constexpr mixxx::RgbColor kTritKeyColor10(0x666600);
...
+                            kTritKeyColor5,
+                            kTritKeyColor6,
+                            kTritKeyColor7,
+                            kTritKeyColor8,
+                            kTritKeyColor9,
+                            kTritKeyColor10,
+                            kTritKeyColor11,
+                            kTritKeyColor12,
+                    }},
+            .defaultHotcueColorPalette = kPalettes.mixxxHotcueColorPalette,
```
daschuer, line 546: This is a deep copy. Can we avoid it?

src/util/color/predefinedcolorpalettes.cpp
```diff
@@ -241,367 +241,333 @@ constexpr mixxx::RgbColor kTritKeyColor10(0x666600);
...
+            .defaultTrackColorPalette = kPalettes.mixxxTrackColorPalette,
+            .defaultKeyColorPalette = kPalettes.mixxxKeyColorPalette,
+            .palettes{
+                    // Hotcue Color Palettes
+                    kPalettes.mixxxHotcueColorPalette,
```
daschuer, line 551: This are also deep copies, which should be avoided.

daschuer (changes requested):

src/preferences/colorpaletteeditor.cpp
Swiftb0y: well, its effectively a singleton. So its up to interpretation whether it should be modeled as a class with a static member function or a namespace with a free function. they're basically equivalent, though I find the namespace more elegant.

src/util/color/predefinedcolorpalettes.cpp
Swiftb0y: we can simply store a const ref because we know the object outlives the reference.

src/util/color/predefinedcolorpalettes.cpp
Swiftb0y: The original were also deep copies though IIRC, but it should be easy to change.

Swiftb0y:
summarizing your review. there are two issues:
1. implementation of the singleton.
  I personally prefer the namespace + free function because its simpler than wrapping both into a class. If there is a significant advantage of the class then I'm happy to change it. Can you think of one?
2. deep copy of palettes. 
  That was already the case previously, though it should be easy to change here if desired.

daschuer (changes requested):
Sorry for letting this go stalled. 

I think we have two concurrent ideas here. On one hande the weak reason to avoid singletons https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#i3-avoid-singletons because they are basically complicated global objects in disguise.

On the other hand we have the astablished well known `instance()` function along with a private constructor that enforces an object to be a singleton. 

I did a brief look into the discussions around it and concluded that the main reasond for i3 are around non const sigletons, that prevent to modulize the code later and avoid unittesting. These reason do not apply here, because we "just" want to gurantee a certain initalization order for a constant object. 

The other part is that the proposed implementation puts the global in disguise in the same way as critizised. You kind of trick i3 by using a singleton namespace instead of a singleton class. I consider this even more of a problem, because it also not uses the esablished sigleton pattern which is well known. This makes the readbilty issue worse. 

What is the problem we like to solve: The fragile situation about the static initalization where on static is initalized with another by following i3 and not introducing a singleton. 

I think that can be done by removing class PredefinedColorPalettes intirely, replacing it with free functions that inline some code now done after get() in the usage code. This may still use static objects, but they can be hidden in an anoymous namespace. 

Since this is a bit of work, deliberately violateing i3 and implement an establische singleton class works also form me as well. 

What do you think?

acolombier:
I know you are quite busy lately @Swiftb0y so I'll make it a draft for now, but feel free to revive if you get time and motivation!

PR #13601, opened 2024-08-25
Fix Reloop Beatmix 4 eject button

sblaisot:
Fix Reloop Beatmix 2/4 controller mapping eject button (long press on load deck button) by fixing typo introduced by #12401


ronso0 (approved):
Fix LGTM, thanks!

Please rebase onto 2.4 so it'll be availble in 2.4.2

PR #13606, opened 2024-08-27
chore: upgrade macos runner to macos-13 Ventura and xcode15

Swiftb0y:
MacOS 14 is also already available. Updating from 12 to 13 should give us a nice boost in C++20 compatibility on MacOS.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,8 +38,8 @@ jobs:
             artifacts_path: build/*.deb
             artifacts_slug: ubuntu-jammy
             qt_qpa_platform: offscreen
-          - name: macOS 12 x64
-            os: macos-12
+          - name: macOS 13 x64
+            os: macos-13
             cmake_args: >-
               -DBULK=ON
               -DCOREAUDIO=ON
@@ -63,8 +63,8 @@ jobs:
             artifacts_path: build/*.dmg
             artifacts_slug: macos-macosintel
             qt_qpa_platform: offscreen
-          - name: macOS 12 arm64
-            os: macos-12
+          - name: macOS 13 arm64
+            os: macos-13
             cmake_args: >-
               -DBULK=ON
               -DCOREAUDIO=ON


JoergAtGithub:
Will this increase the runtime requirements as well?

daschuer:
This virtually drops the still supported macOS Monterey (V 12) form 2021-08. 
Only three month ago we have dropped macOS Big Sur (V 11) from 2020-11 because GitHub has dropped the runner support. It has fallen out of Apple's support 09.2023 
https://github.com/mixxxdj/mixxx/pull/13296 
Apple will release a new version once a year and usually maintains two older versions 

Mixxx 2.6 is scheduled earliest at 2025.06  (after 2.5 + 60 day + beta)  https://github.com/mixxxdj/mixxx/wiki/250_release_proposal

Conclusion:
Changing the runner now in main is fine because Mixxx 2.6 will be released after macOS Monterey is dropped.

The other question is the XCode version. macOS 13.0 comes with XCode 14.1 
XCode 15 requires macOS 13.5 with  SDK macosx14.0 and has supports following deployment targets: macOS 10.13-14.2

Here is the readme describing the GitHub runner: 
https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md

So it looks like this will not limit the Mixxx binary support, but may be a barrier for contributors? 
@fwcd what is your experience? 

For now I think we should stick to the default pair macOS 13 + XCode 14.1 
At least this would be an intermediate step and form there we may advance once we have a compelling reason

daschuer:
Since GitHub defaults to the max XCode version. We use now XCode 14.2 after the upgrade we will use XCode 15.2 unless we add:
```
DEVELOPER_DIR: /Applications/Xcode_14.3.1.app/Contents/Developer
/bin/bash -c "sudo xcode-select --switch /Applications/Xcode_14.3.1.app/Contents/Developer"
```
https://github.com/mixxxdj/vcpkg/pull/146/commits/e1076c9e0ac9d50af5d1e6752a6bfe8d705d6b62

fwcd:
Small note: It's capitalized "Xcode", not "XCode" :D

> Will this increase the runtime requirements as well?

Not necessarily. In general, building with a newer SDK/Xcode without bumping the requirements should usually if not always be fine. For the App Store Apple even requires using the newest one IIRC. Due to ABI stability, the binary should still run on whichever macOS version our deployment target is set to, which in practice likely means we're bound to the oldest macOS version supported by Qt.

Building with a newer compiler also has the advantage that we get better C++20/23 support, so the only[^1] reason to build on an older macOS would be making sure that our build wasn't misconfigured somewhere (e.g. some dependency not built against the older deployment target).

My impression is that most devs on macOS are usually on the latest version anyway, so this likely doesn't really impact contributors much.

[^1]: Actually there was another reason, namely that there ist a cpack bug, which occasionally crashes due to an hdiutil race condition on newer macOS versions, see https://github.com/actions/runner-images/issues/7522 and https://gitlab.kitware.com/cmake/cmake/-/issues/25671. Not sure what the status of this is and whether we still need to [patch around](https://github.com/fwcd/m1xxx/blob/479afb5885211f8c59fcfc131050b625d0f1408a/.github/workflows/build.yml#L221) this issue.

fwcd:
I think there is some merit to not using the latest runner, since we'll likely have few contributors using them and thus (if the build is misconfigured), we may unintentionally break backwards compatibility if we forget to set the deployment target somewhere. Maybe a more contrived option, but perhaps we could _build_ on the newest macOS and then run the test suite on the older OS? Not sure if that would impose too much overhead though.

daschuer:
If, we have a runner to check backwards compatibility, we can IMHO use it right away, there is less risk that such a binary will crash after updating. That is the reason why we normally use the oldest runner. The bump here is OK, because the macos-12 runner will likely retired during 2.6-beta and this way we avoid a runner bump during beta. I remember some incidences where a runner update broke our CI.

Our CI is already quite "expensive" and has a certain CO² foodprint, so we should use it deliberately.

daschuer (approved):
Ok, So it looks like an update is safe.

Swiftb0y:
Thanks

fwcd:
This is the bug I mentioned. We'll have to fix it anyway at some point, since the macos-12 runners likely won't be supported for that long either.

fwcd:
The issue seems to be that macOS's antivirus (XProtect) scans the disk image, which causes a race condition in CPack. The two main workarounds are killing the XProtect process (which is still a bit brittle, since it automatically relaunches) and simply rerunning CPack a few times until it succeeds. Neither of them are particularly elegant, but they seem to work "well enough" in practice.

If nothing else works, we could also package Mixxx as zip or tarball, but at least for the releases it would be nice to have a dmg, for consistency and to avoid having to hack around this in packages (e.g. Homebrew).

daschuer:
Both measures sounds like a workaround not a solution. Is there a ignore list or such that we can configure to not do the scan during Cpack? 

In general we need an estimate how long a proper upstream solution will take and how stable our workaround is. 

For now it seems to be more stable to revert main and test possible solutions on the https://github.com/mixxxdj/mixxx/tree/chore/upgrade-macos13-xcode15.2 branch. Is that OK?

fwcd:
Works for me! Admittedly I haven't looked into what upstream has been doing, but I do believe it's still a CPack bug

Swiftb0y:
> For now it seems to be more stable to revert main and test possible solutions on the https://github.com/mixxxdj/mixxx/tree/chore/upgrade-macos13-xcode15.2 branch. Is that OK?

Sure... Though I wonder why this didn't happen before. explicitly pushed this branch upstream so the codesign CI would run... I guess the failure is spontaneous?

daschuer:
Done 1dc2119

fwcd:
> Though I wonder why this didn't happen before. explicitly pushed this branch upstream so the codesign CI would run... I guess the failure is spontaneous?

It's a race condition, so it only happens sometimes. That's why retrying CPack is one of the workarounds.

Swiftb0y:
Ah right, yeah...

fwcd:
This looks like a Homebrew message, not a deprecation from GitHub. It just means, they won't provide prebuilt packages (bottles) anymore, so our CI workflows will likely take longer as they have to build dependencies from scratch. They should still work though.

I agree that we'll have to revisit this, sooner or later, anyway though.

Swiftb0y:
welp, now we do.
* https://github.com/actions/runner-images/issues/10721

Issue #13083, opened 2024-04-10
Double reload when editing controller mapping file


acolombier:
### Bug Description

When editing a controller mapping file using IDE like VSCode, Mixxx triggers two reloads upon file save, instead of one.

There was the same issue for QML and it was fixed [here](https://github.com/mixxxdj/mixxx/pull/12844) ([initial attempt](https://github.com/mixxxdj/mixxx/pull/12795)).

The problem is documented [here](https://github.com/Holzhaus/mixxx/blob/main/src/qml/qmlautoreload.h#L29) ([Qt doc about it](https://doc.qt.io/qt-6/qfilesystemwatcher.html#fileChanged)) and a similar issue will likely impact the [keyboard auto reload](https://github.com/mixxxdj/mixxx/pull/13082).

### Version

main

### OS

All

PR #13607, opened 2024-08-27
refactor: split out `AutoFileReloader` from `QmlAutoReload`

Swiftb0y:
`AutoFileReload` is useful for almost all occasions where hot- reloading is needed. Thus it belongs into a separate class for easier reuse. Plus I added an extra RuntimeLoggingCategory so we don't have a category-less `qDebug` call.

### DIFF
--- a/src/util/autofilereloader.cpp
+++ b/src/util/autofilereloader.cpp
@@ -1,0 +1,23 @@
+#include "util/autofilereloader.h"
+
+#include <QDebug>
+#include <QString>
+
+#include "moc_autofilereloader.cpp"
+#include "util/runtimeloggingcategory.h"
+
+AutoFileReloader::AutoFileReloader(const RuntimeLoggingCategory& loggingCategory)
+        : m_loggingCategory(loggingCategory) {
+    connect(&m_fileWatcher,
+            &QFileSystemWatcher::fileChanged,
+            this,
+            &AutoFileReloader::slotFileChanged);
+}
+
+void AutoFileReloader::slotFileChanged(const QString& changedFile) {
+    qCDebug(m_loggingCategory) << "File" << changedFile << "has been changed.";
+    if (m_pathTriggeringReload.remove(changedFile)) {
+        m_fileWatcher.removePath(changedFile);
+        emit fileChanged(changedFile);
+    }
+}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1146,6 +1146,7 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/track/tracknumbers.cpp
   src/track/trackrecord.cpp
   src/track/trackref.cpp
+  src/util/autofilereloader.cpp
   src/util/battery/battery.cpp
   src/util/cache.cpp
   src/util/clipboard.cpp
--- a/src/qml/qmlautoreload.cpp
+++ b/src/qml/qmlautoreload.cpp
@@ -5,40 +5,28 @@
 #include <QUrl>
 
 #include "moc_qmlautoreload.cpp"
+#include "util/autofilereloader.h"
 
 namespace mixxx {
 
 namespace qml {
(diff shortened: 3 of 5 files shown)


src/util/autofilereloader.cpp
```diff
@@ -0,0 +1,23 @@
...
+        : m_loggingCategory(loggingCategory) {
+    connect(&m_fileWatcher,
+            &QFileSystemWatcher::fileChanged,
+            this,
+            &AutoFileReloader::slotFileChanged);
+}
+
+void AutoFileReloader::slotFileChanged(const QString& changedFile) {
+    qCDebug(m_loggingCategory) << "File" << changedFile << "has been changed.";
+    if (m_pathTriggeringReload.remove(changedFile)) {
```
daschuer, line 19: Why is a changed file removed form the watcher?

src/util/autofilereloader.cpp
```diff
@@ -0,0 +1,23 @@
...
+        m_fileWatcher.removePath(changedFile);
```
daschuer, line 20: removePath returns also true on success. What is the role for m_pathTriggeringReload?

src/util/autofilereloader.cpp
Holzhaus: See the comment in the header file. It is done to prevent double-reload caused by the different editor saving strategies (VSCode vs. vim/sed).

@Swiftb0y I think it should be pointed out in a comment that it's the callers responsibility to re-add the file once the reload was handled (in the QML case, this is done in the intercept call, after the changed file was parsed).

src/util/autofilereloader.cpp
Swiftb0y: to ensure we don't emit a second signal in case of the append/truncate process (see the doc in the header).

acolombier: @daschuer this is to specifically fix #13083.
When I implemented this auto reload, I wanted to make sure not to introduce the same limitation on QML.

acolombier (commented):
The code looks good, thanks for taking the time to do this well needed refactor.
I'll try to test that locally within the next few hours, days max. Please feel free to merge without my :+1:  if I'm taking too long to do so.

Swiftb0y:
Thanks for the testing. I'm not sure my tests tested both cases comprehensively.

src/util/autofilereloader.cpp
acolombier: iIRC, my initial implementation was exclusively relying on the file watcher to keep track of which files are being watched, but I can remember why I ended adding that extra array. I believe the initial PR/commits may have the explanation, I will try to dig it later, when I'm not AFK.

acolombier: So the [original implementation](https://github.com/mixxxdj/mixxx/pull/12795/files#diff-f9cb0d49677f8a2d66f91e60c0d4c802bc51bdb25161a65118bbdf884a8d3d7c) only contained the QFileSystemWatcher, but we [had to add](https://github.com/mixxxdj/mixxx/pull/12844/files#diff-f9cb0d49677f8a2d66f91e60c0d4c802bc51bdb25161a65118bbdf884a8d3d7c) the array to desperately handle acknowledgement as [QFileSystemWatcher::removePath](https://doc.qt.io/qt-6/qfilesystemwatcher.html#removePath)'s returns appears to not be consistent against platforms.

acolombier (approved):
LGTM, thanks!

daschuer (changes requested):

src/util/autofilereloader.cpp
Swiftb0y: Its in the header of `AutoFileReloader` as documentation for the `m_pathTriggeringReload` member.

daschuer (approved):
Ok, thank you. LGTM

Swiftb0y:
Thank you

PR #13608, opened 2024-08-28
Key Colors for the Key text in the decks

danferns:
This PR adds key colors to the UI of the decks.

key

### DIFF
--- a/src/widget/wkey.cpp
+++ b/src/widget/wkey.cpp
@@ -1,28 +1,37 @@
 #include "widget/wkey.h"
 
+#include <QStyleOption>
+#include <QStylePainter>
+
 #include "library/library_prefs.h"
 #include "moc_wkey.cpp"
+#include "preferences/usersettings.h"
 #include "skin/legacy/skincontext.h"
 #include "track/keyutils.h"
 
-WKey::WKey(const QString& group, QWidget* pParent)
+WKey::WKey(const QString& group, UserSettingsPointer pConfig, QWidget* pParent)
         : WLabel(pParent),
-          m_dOldValue(0),
           m_keyNotation(mixxx::library::prefs::kKeyNotationConfigKey, this),
           m_engineKeyDistance(group,
                   "visual_key_distance",
                   this,
-                  ControlFlag::AllowMissingOrInvalid) {
-    setValue(m_dOldValue);
+                  ControlFlag::AllowMissingOrInvalid),
+          m_engineKey(group,
+                  "key",
+                  this,
+                  ControlFlag::AllowMissingOrInvalid),
+          m_colorPaletteSettings(pConfig) {
+    setValue();
     m_keyNotation.connectValueChanged(this, &WKey::keyNotationChanged);
     m_engineKeyDistance.connectValueChanged(this, &WKey::setCents);
 }
 
 void WKey::onConnectedControlChanged(double dParameter, double dValue) {
     Q_UNUSED(dParameter);
+    Q_UNUSED(dValue);
     // Enums are not currently represented using parameter space so it doesn't
     // make sense to use the parameter here yet.
-    setValue(dValue);
+    setValue();
 }
 
 void WKey::setup(const QDomNode& node, const SkinContext& context) {
@@ -31,23 +40,21 @@ void WKey::setup(const QDomNode& node, const SkinContext& context) {
     m_displayKey = context.selectBool(node, "DisplayKey", true);
 }
 
-void WKey::setValue(double dValue) {
-    m_dOldValue = dValue;
-    mixxx::track::io::key::ChromaticKey key =
-            KeyUtils::keyFromNumericValue(dValue);
-    if (key != mixxx::track::io::key::INVALID) {
+void WKey::setValue() {
+    m_key = KeyUtils::keyFromNumericValue(m_engineKey.get());
+    m_diff_cents = m_engineKeyDistance.get();
+    if (m_key != mixxx::track::io::key::INVALID) {
         // Render this key with the user-provided notation.
         QString keyStr = "";
         if (m_displayKey) {
-            keyStr = KeyUtils::keyToString(key);
+            keyStr = KeyUtils::keyToString(m_key);
         }
         if (m_displayCents) {
-            double diff_cents = m_engineKeyDistance.get();
-            int cents_to_display = static_cast<int>(diff_cents * 100);
+            int cents_to_display = static_cast<int>(m_diff_cents * 100);
             char sign = ' ';
-            if (diff_cents < 0) {
+            if (m_diff_cents < 0) {
                 sign = '-';
-            } else if (diff_cents > 0) {
+            } else if (m_diff_cents > 0) {
                 sign = '+';
             }
             keyStr.append(QString(" %1%2c").arg(sign).arg(qAbs(cents_to_display)));
@@ -56,15 +63,73 @@ void WKey::setValue(double dValue) {
     } else {
         setText("");
     }
+    update();
 }
 
 void WKey::setCents() {
-    setValue(m_dOldValue);
+    setValue();
 }
 
 void WKey::keyNotationChanged(double dKeyNotationValue) {
     Q_UNUSED(dKeyNotationValue);
     // NOTE: dKeyNotationValue is the index of the key notation type, NOT the
     // key itself, so we intentionally set the old value again to update the UI.
-    setValue(m_dOldValue);
+    setValue();
+}
+
+void WKey::paintEvent(QPaintEvent* event) {
+    if (m_key == mixxx::track::io::key::INVALID || !m_colorPaletteSettings.getKeyColorsEnabled()) {
+        WLabel::paintEvent(event);
+        return;
+    }
+
+    ColorPalette keyColorPalette = m_colorPaletteSettings.getConfigKeyColorPalette();
+
+    QColor colorTop, colorBottom;
+    double splitPoint = 0; // 'height' of top color
+    if (m_diff_cents < 0) {
+        colorTop = KeyUtils::keyToColor(m_key, keyColorPalette);
+        colorBottom = KeyUtils::keyToColor(KeyUtils::scaleKeySteps(m_key, -1), keyColorPalette);
+        splitPoint = m_diff_cents + 1;
+    } else {
+        colorTop = KeyUtils::keyToColor(KeyUtils::scaleKeySteps(m_key, 1), keyColorPalette);
+        colorBottom = KeyUtils::keyToColor(m_key, keyColorPalette);
+        splitPoint = m_diff_cents;
+    }
+
+    QStyleOption option;
+    option.initFrom(this);
+    QStylePainter painter(this);
+
+    const QStyle* pStyle = style();
+    const QRect contRect = pStyle->subElementRect(QStyle::SE_FrameContents, &option, this);
+
+    const int rectWidth = 4;
+    const int splitHeight = static_cast<int>(contRect.height() * splitPoint);
+
+    painter.fillRect(contRect.left(),
+            contRect.top(),
+            rectWidth,
+            splitHeight,
+            colorTop);
+
+    painter.fillRect(contRect.left(),
+            splitHeight + 1,
+            rectWidth,
+            contRect.height() - splitHeight,
+            colorBottom);
+
+    painter.setPen(option.palette.text().color());
+
+    QString elidedText = option.fontMetrics.elidedText(
+            text(),
+            Qt::ElideRight,
+            width() - rectWidth);
+
+    painter.drawText(rectWidth,
+            contRect.top(),
+            contRect.width() - rectWidth,
+            contRect.height(),
+            Qt::AlignCenter,
+            elidedText);
 }
--- a/src/preferences/colorpalettesettings.h
+++ b/src/preferences/colorpalettesettings.h
@@ -29,7 +29,7 @@ class ColorPaletteSettings {
     void removePalette(const QString& name);
     QSet<QString> getColorPaletteNames() const;
 
-    DEFINE_PREFERENCE_HELPERS(KeyColorsEnabled, bool, "[Config]", "KeyColorsEnabled", true);
+    DEFINE_PREFERENCE_HELPERS(KeyColorsEnabled, bool, "[Config]", "key_colors_enabled", true);
 
   private:
     UserSettingsPointer m_pConfig;
--- a/src/skin/legacy/legacyskinparser.cpp
+++ b/src/skin/legacy/legacyskinparser.cpp
@@ -1338,7 +1338,7 @@ QWidget* LegacySkinParser::parseNumberPos(const QDomElement& node) {
 
 QWidget* LegacySkinParser::parseEngineKey(const QDomElement& node) {
     QString group = lookupNodeGroup(node);
-    WKey* pEngineKey = new WKey(group, m_pParent);
+    WKey* pEngineKey = new WKey(group, m_pConfig, m_pParent);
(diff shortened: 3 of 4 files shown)


src/widget/wkey.cpp
```diff
@@ -1,18 +1,25 @@
 #include "widget/wkey.h"
 
+#include <qboxlayout.h>
+#include <qevent.h>
```
ronso0, line 4: Why not
```suggestion
#include <QBoxlayout>
#include <QEvent>
```
like elsewhere?

ronso0:
Thanks for your continuous work on this!

It doesn't build, did you forget to commit something?
Why a WidgetGroup (that doesn't seem to be set up), doesn't it suffice to set a stylesheet with color and background color?

daschuer:
This is a midair PR, I have requested to wrap up GSoC.
We are unsure what is the best approach here visually.
Maybe we should collect first design mockups and than continue here? 

@ronso0 what do you think. Is coloring the background the right approach? How to deal with off-tune pitches? 
Alternative: a border or a separate box.

One idea was to fade out the color in the range of 10 to 20 ct off and show no color for tracks not on the key wheel.

ronso0:
Ah okay, I wasn't aware of the context, thanks for the clarification.

AFAIK the key-related design process happend on Zulip, so maybe we should stick to that, idk.

I agree with the color bar approach as mentioned in the description. A for consistency with the library's key column, B a distinct spot/area with information, compared to a (dimmed) colored area where the color information maybe affected by a contrasting text. It'd also be consistent with the track color bar in the decks.

What information do I expect from the key widget? _(none actually since I don't use it ...)_
Before mixing tracks the decks' key color should tell me at a glance whether tracks are compatible. So basically: same color = match. The more advanced method would require having a scheme of compatible colors in mind, for example like green matches yellow and bright blue.
I imagine it to be helpful indeed that, if they is off by XX cents after pitch shifting, the color should fade significantly.
Maybe we can incorporate a slider-like marker on the bar to indicate the position related to the current key. (= `visual_key_distance` control like in Shade skin. For feedback, not for control).

vespadj:
Color key is in Library (from v. 2.6), so in the same way should be shown in Deck > Track Key.

danferns:
Hi! Just a ping that I've managed to get the bar design working! Thank you all for your feedback.

image

It's quite rough. And I need to add key palette support. 

I will most likely open a separate PR for this one a little later as I did it in a new branch.

daschuer:
Great, thank you for picking this up again yourself. 
I really need some time to add the tuning thing to the key analysis. 
Did you consider how to show off tune keys?

Eve00000:
> ... get the bar design working

that looks nice!

ronso0:
> Did you consider how to show off tune keys?

What about painting the upper 1/5 of the bar in a contrasting color?

danferns:
> > Did you consider how to show off tune keys?
> 
> What about painting the upper 1/5 of the bar in a contrasting color?

I love this idea! Adjacent keys (eg. C and C#/Db) are already contrasting, so as we're changing the rate slider, we can see the next color slide into the rectangle:

image

daschuer:
Sometimes the solution is to near to see it :-)

danferns:
That worked really nicely.

key

ronso0:
Just a heads-up: in my personal branch I have [3c02a0a LateNight: add key offset indicator](https://github.com/mixxxdj/mixxx/commit/3c02a0af3de7798dea7575952fe9dbe8046d5869) which makes key controls complete for me (I often forgot to reset the key on track load, and there's still a bug with key reset with at least one combination of keylock/unlock + tempo shift)

Maybe I'll try to backport that at some time-

src/widget/wkey.cpp
```diff
@@ -2,18 +2,23 @@
...
                   this,
-                  ControlFlag::AllowMissingOrInvalid) {
-    setValue(m_dOldValue);
+                  ControlFlag::AllowMissingOrInvalid),
+          m_engineKey(group,
+                  "key",
+                  this,
+                  ControlFlag::AllowMissingOrInvalid),
+          m_colorPaletteSettings(pConfig) {
+    setValue();
```
Swiftb0y, line 24: this is drifting into an undesirable direction. You should already have everything you need without the new ControlProxy

src/widget/wkey.cpp
```diff
@@ -56,15 +59,59 @@ void WKey::setValue(double dValue) {
...
     Q_UNUSED(dKeyNotationValue);
     // NOTE: dKeyNotationValue is the index of the key notation type, NOT the
     // key itself, so we intentionally set the old value again to update the UI.
-    setValue(m_dOldValue);
+    // NOTE: This also seems to handle changing the Key Palette?
+    setValue();
+}
+
+void WKey::paintEvent(QPaintEvent* event) {
+    QString keyText = this->text();
```
Swiftb0y: `this->` is not used in mixxx and considered controversial for many C++ users.

src/widget/wkey.cpp
ronso0: and all proxies can be turned into PollingControlProxy btw

ronso0: > You should already have everything you need without the new ControlProxy

@danferns the widget connections are set up by LegacySkinParser: `setupLabelWidget` -> `commonWidgetSetup()` -> `setupConnections()` ..

Swiftb0y: > and all proxies can be turned into PollingControlProxy btw

Can they?

https://github.com/mixxxdj/mixxx/blob/837f0ab69904c3f4a086fab5323029fb0d0f0c5d/src/widget/wkey.cpp#L17-L18

danferns: > @danferns the widget connections are set up by LegacySkinParser: `setupLabelWidget` -> `commonWidgetSetup()` -> `setupConnections()` ..

Thank you, TIL that much of the UI in Mixxx is built dynamically based on XML files! That's so powerful.

I think I'd need to create connections to the engine key and key distance values in `res/skins/latenight/decks/key_controls.xml`. I'll look into how that is done.

Swiftb0y: Well, the purpose of `WKey` is clear. There is no need to shift code around to make the widget harder to use (by having to create the correct controls in the XML). As I said, you already have all the data you need. You can access the `"key"` using the parameter passed to `setValue` and the fine tuning via the `m_engineKeyDistance` CO. Now the only thing you need to do is to properly recombine them again and and then draw the color bar.

danferns: > You can access the `"key"` using the parameter passed to `setValue` and the fine tuning via the `m_engineKeyDistance` CO.

Initially, I was using the `"key"` passed into `setValue`, but I found that it would not get updated in sync with `m_engineKeyDistance`. When I tried logging the key and cents it would look something like this (from memory):

```
key: 7 cents: 0.45
key: 7 cents: 0.48
key: 7 cents: -0.48 <- cents got updated before key
key: 8 cents: -0.48
```

This would appear as a momentary glitch in the colored rectangle when the keys transition. 

I'm not sure why there is this delay, but when I used engineKey, it was perfectly in sync with engineKeyDistance. 

Also, I noticed that `key_controls.xml` creates a connection to `visual_key` instead of `key`.

ronso0: > > and all proxies can be turned into PollingControlProxy btw
> 
> Can they?

Umm, no :grimacing:

ronso0: > ... This would appear as a momentary glitch in the colored rectangle when the keys transition.
> I'm not sure why there is this delay, but when I used engineKey, it was perfectly in sync with engineKeyDistance.
> 
> Also, I noticed that `key_controls.xml` creates a connection to `visual_key` instead of `key`.

And the existing `visual_key_distance` connection doesn't work for this?

danferns: The `visual_key_distance` from `m_engineKeyDistance` is out of sync with the `visual_key` connection in XML (which as I understand is passed into `setValue`). In the XML, I don't see a `visual_key_distance` connection.

But `visual_key_distance` from `m_engineKeyDistance` is in sync with `key` from `m_engineKey`.

ronso0: > Initially, I was using the "key" passed into setValue, but I found that it would not get updated in sync with m_engineKeyDistance. When I tried logging the key and cents it would look something like this (from memory):

It seesm [`visual_key` is identical to `key`](https://github.com/mixxxdj/mixxx/blob/837f0ab69904c3f4a086fab5323029fb0d0f0c5d/src/waveform/visualsmanager.cpp#L52), except that [`visual_key` is updated at the same rate as the waveforms](https://github.com/mixxxdj/mixxx/blob/837f0ab69904c3f4a086fab5323029fb0d0f0c5d/src/waveform/waveformwidgetfactory.cpp#L790)?

And IIUC `visual_key_distance` is updated with the engine's rate (determined by latency/buffer length).

danferns: This does seem to be the case. Should we rename `visual_key_distance -> key_distance`, and create a new `visual_key_distance` that *is* in sync with `visual_key`?

src/widget/wkey.cpp
danferns: Thank you!

src/widget/wkey.cpp
danferns: I've done the renaming for visual key, and switched back to using the dValue that was passed. Thanks to both of you for your feedback!

danferns:
I'm a bit unsure about using `WKey::paintEvent`, because I had to manually draw the text with it, so the font rendered now is not precisely the same as before (for example, I had to specify the color `Qt::white` in code, instead of using the text color from the theme). 

It is possible to call the parent's `WLabel::paintEvent` instead, but the text will be centered relative to the entire WKey widget, not just the space to the right of the colored rectangle.

image

It's also possible to align the text to the left, and add some left padding, to make it look like the cells in the library's Key column. But this would be a design change.

What would be the best way to go about this?

daschuer:
Did you consider to buy both parts into separate sibling witdgets? The original widget becomes a vertical layout that holds the text and the bar widget.

ronso0:
You could create a `QStyleOption` and do all kind of cusotm painting with it.
See `WSliderComposed::paintEvent()` for example.

ronso0:
A separate widget would also work, but requires skin changes (for example in LateNight, WKey is styled like a text edit and we'd need to adjust the key color bar to match that)

danferns:
The `QStyleOption` method worked well!

image

image

Using separate widgets was indeed what I was trying first, but I could not figure out how to go about it so I had resorted to the paintEvent approach. Now I'm quite happy with it!

I'll now fix the build errors, and add a check for whether key colors are enabled.

danferns:
The PR is ready for a final review!

ronso0:
I think you can retrive the rect inside the borders so the color bar is inside, not "attached" to the entire.
This way we could use WKey as before and wouldn't have to adjust skins.

danferns:
Thanks for the suggestion, it does look nicer:

image

src/widget/wkey.cpp
```diff
@@ -66,5 +71,63 @@ void WKey::keyNotationChanged(double dKeyNotationValue) {
...
+
+    const int padding = 2;
+    const int rectWidth = 4;
+    const int rectPaddedWidth = rectWidth + 2 * padding;
+    const int rectHeight = height() - 2 * padding;
+    const int splitHeight = static_cast<int>(rectHeight * splitPoint);
+
+    QStyleOption option;
+    option.initFrom(this);
+    QStylePainter painter(this);
```
ronso0, line 102: I suggest to remove the hard-coded padding and simply place the bar inside the **content rect**.
Then the bar(s) will always be inside the rect defined by custom borders and margins.
You can get it with
```
    const QStyle* pStyle = style();
    const QRect contRect = pStyle->subElementRect(QStyle::SE_FrameContents, &option, this);
```
then use the rectangle's `top()`, `left()` and `height()` to paint the color rects.

danferns: Thank you, much cleaner solution!

ronso0:
Oh, this has been waiting way too long.

@danferns Are you still around and want to finish this?
Taking another look I noticed that f7b9aa1f1b703747f9716387dcfb827dd1e73630 wil break controller scripts and Shade.
Looks like this refactoring isn't strictly required for the new Key widget feature, right?

danferns:
> @danferns Are you still around and want to finish this?

Absolutely! 

> Looks like this refactoring isn't strictly required for the new Key widget feature, right?

Yep, the refactor was mostly for consistent naming. I'll be able to revert and test it tomorrow.

danferns:
I found that reverting the refactor alone makes  `m_key` and `m_diff_cents` go momentarily out of sync again. (It appears as a UI glitch when we're near +/- 50 cents)

I've also reverted [68372a0](https://github.com/mixxxdj/mixxx/commit/68372a0f41c62d62351a7cf8ef8a43727914d776) (this commit was rebased away).

Now Shade's key widget works as expected, and there is also no glitch in the colored rectangle.

ronso0:
Great, thanks!
I think you can squash all commits now.

I'll test later on.

src/widget/wkey.cpp
```diff
@@ -56,15 +63,74 @@ void WKey::setValue(double dValue) {
...
+    const int splitHeight = static_cast<int>(contRect.height() * splitPoint);
+
+    painter.fillRect(contRect.left(),
+            contRect.top(),
+            rectWidth,
+            splitHeight,
+            colorTop);
+
+    painter.fillRect(contRect.left(),
+            splitHeight,
```
ronso0: Nit: currently the bars overlap at `splitHeight`.

Image

```suggestion
            splitHeight + 1,
```
or `- 1` for the upper part 🤷‍♂️

ronso0:
LGTM, just a minor issue.

Unfortunately the key bar is painted immediately when enabling key colors in the preferences, it's updated on next pitch change, but that's a minor issue, as well. Fixing that would require a CO or something, but that's overkill IMO.
**edit** same for key color palette changes.

src/widget/wkey.cpp
```diff
@@ -56,15 +63,74 @@ void WKey::setValue(double dValue) {
...
-    setValue(m_dOldValue);
+    setValue();
 }
 
 void WKey::keyNotationChanged(double dKeyNotationValue) {
     Q_UNUSED(dKeyNotationValue);
     // NOTE: dKeyNotationValue is the index of the key notation type, NOT the
     // key itself, so we intentionally set the old value again to update the UI.
-    setValue(m_dOldValue);
+    // NOTE: This also seems to handle changing the Key Palette?
```
ronso0: nope, it's not.

ronso0:
Great, no squash again please ; ) then it's ready for merge.

src/widget/wkey.cpp
danferns: Yep, just tried it and it does not. I might have not tested it properly when I wrote this. Thanks for spotting!

danferns:
Done, finally! :smile:

ronso0:
:tada:
Now let's wait for CI to finish.

ronso0 (approved):

ronso0:
Shall we also update the  tooltip? (in a follow-up PR)
https://github.com/mixxxdj/mixxx/blob/a28d029a8ea6ebee6ec7c55da1d481bc583a2a5e/src/skin/legacy/tooltips.cpp#L418-L421
(oh, at least LateNight still uses the `track-key` tooltip id. will fix)

Something like
"It also shows a colored bar if Key colors are enabled in the Preferences."
"This will be split vertically if the track's key is in between full keys."
(adjust as you like)

danferns:
Good suggestion! Will open it later today.

PR #13609, opened 2024-08-29
(fix) WStarRating: remove unneeded signal that may cause rating reset via track menu

### LINKED ISSUE #13597: Stars rating is lose when right click on multiple tracks.
### Bug Description

Select multiple tracks at the same time, tracks should have different stars count.
Right click on one track to edit all selected tracks (like if you want to apply a color to all selected tracks).
Notice that stars rating is gone for all selected tracks. 

No problem on Mixxx 2.4.1

### Version

2.5-beta-43-g34aa89c780

### OS

PopOS 22.04 (Ubuntu 22.04), X11, NVIDIA

ronso0:
fixes #13597 

I removed the unneeded signal added in e4ce36207c0a3bc9a1c94936374a5eeefa5d5f8a
Still don't understand why this signal is emitted (received/processed) only if the [repro steps](https://github.com/mixxxdj/mixxx/issues/13597#issuecomment-2314142159) are done in quick succession.

### DIFF
--- a/src/library/dlgtrackinfomulti.cpp
+++ b/src/library/dlgtrackinfomulti.cpp
@@ -313,11 +313,8 @@ void DlgTrackInfoMulti::updateFromTracks() {
         }
     }
     // Update the star widget
-    // Block signals to not set the 'modified' flag.
-    m_pWStarRating->blockSignals(true);
     m_pWStarRating->slotSetRating(commonRating);
     m_starRatingModified = false;
-    m_pWStarRating->blockSignals(false);
 
     // Same procedure for the track color
     mixxx::RgbColor::optional_t commonColor = m_trackRecords.first().getColor();
--- a/src/widget/wstarrating.cpp
+++ b/src/widget/wstarrating.cpp
@@ -39,7 +39,6 @@ void WStarRating::slotSetRating(int starCount) {
     }
     m_starCount = starCount;
     updateVisualRating(starCount);
-    emit ratingChangeRequest(starCount);
 }
 
 void WStarRating::paintEvent(QPaintEvent * /*unused*/) {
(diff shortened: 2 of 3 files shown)


daschuer (changes requested):
Thank you for the fast responds. The change looks good. However it was a bit confusing to review the PR. 

The issue is that the name `WStarRating::slotSetRating()` does not indicate that it does not have effect to the focussed track anymore while ` WTrackMenu::slotSetRating()` has.

I have not a strong option which one should be changed. Maybe something like 
`WStarRating::slotRatingChanged()` and ` WTrackMenu::slotApplyRating()` changing one of them should be also sufficient. 

Related is also this: 
https://github.com/mixxxdj/mixxx/blob/400ae4618078fb07c063fceae3056253c6725d09/src/widget/wstarratingaction.cpp#L29-L31


Can we remove the `blockSignals()` calls here? 
https://github.com/mixxxdj/mixxx/blob/400ae4618078fb07c063fceae3056253c6725d09/src/library/dlgtrackinfomulti.cpp#L317-L320

ronso0:
> The issue is that the name WStarRating::slotSetRating() does not indicate that it does not have effect to the focussed track anymore while  WTrackMenu::slotSetRating() has.

This is no issue: WStarRating is responsible only for the visual -- it doesn't deal with the tracks, it only emits the change signal.
WTrackMenu does handle tracks and the name `slotSetRating` is consistent with the other set slots.

ronso0:
> Can we remove the blockSignals() calls here?

Yes, in the blocks in DlgTrackInfoMulti can be removed now. Thanks.

daschuer (approved):
LGTM, Thank you.

PR #13610, opened 2024-08-29
(fix) Track file export: various fixes

ronso0:
1. The overwrite dialog had confusing/duplicate buttons,and clicking Skip would cancel the dialog
   before:
   export-dlg-before

   now: 
   export-dlg-now

2. Crate: use the right-clicked carte (no the selected crate, which might be null)
3. abort track export if no tracks are selected (empty crate/playlist), previously the progress dialog would be empty and remain open forever

### DIFF
--- a/src/library/export/trackexportdlg.cpp
+++ b/src/library/export/trackexportdlg.cpp
@@ -68,29 +68,30 @@ void TrackExportDlg::slotAskOverwriteMode(
             QMessageBox::Warning,
             tr("Overwrite Existing File?"),
             tr("\"%1\" already exists, overwrite?").arg(filename),
-            QMessageBox::Cancel | QMessageBox::No | QMessageBox::NoToAll
-            | QMessageBox::Yes | QMessageBox::YesToAll);
-    question_box.setDefaultButton(QMessageBox::No);
-    question_box.addButton(tr("&Overwrite"), QMessageBox::YesRole);
-    question_box.addButton(tr("Over&write All"), QMessageBox::YesRole);
-    question_box.addButton(tr("&Skip"), QMessageBox::NoRole);
-    question_box.addButton(tr("Skip &All"), QMessageBox::NoRole);
+            QMessageBox::Cancel);
 
-    switch (question_box.exec()) {
-    case QMessageBox::No:
+    QPushButton* pSkip = question_box.addButton(
+            tr("&Skip"), QMessageBox::NoRole);
+    QPushButton* pSkipAll = question_box.addButton(
+            tr("Skip &All"), QMessageBox::NoRole);
+    QPushButton* pOverwrite = question_box.addButton(
+            tr("&Overwrite"), QMessageBox::YesRole);
+    QPushButton* pOverwriteAll = question_box.addButton(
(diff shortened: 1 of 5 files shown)


daschuer:
In the corresponding Nemo File Manager dialog we have checkbox for all files and only three buttons. 
But that would be not a bugfix PR.

daschuer (changes requested):
After fixing the build it works here. I will file a feature request for the checkbox.

src/library/trackset/crate/cratefeature.cpp
```diff
@@ -833,10 +833,15 @@ void CrateFeature::slotExportPlaylist() {
 }
 
 void CrateFeature::slotExportTrackFiles() {
+    m_lastRightClickedIndex = index;
```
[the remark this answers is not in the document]
ronso0: umm :shrug:
I made the fix in 2.5, build, then rebased onto 2.4 but didn't build again, so my guess is this is a wronlgy resolved conflict.
Interesting that CI can build it.

src/library/export/trackexportdlg.cpp
```diff
@@ -68,29 +68,30 @@ void TrackExportDlg::slotAskOverwriteMode(
...
+    QPushButton* pSkipAll = question_box.addButton(
+            tr("Skip &All"), QMessageBox::NoRole);
+    QPushButton* pOverwrite = question_box.addButton(
+            tr("&Overwrite"), QMessageBox::YesRole);
+    QPushButton* pOverwriteAll = question_box.addButton(
+            tr("Over&write All"), QMessageBox::YesRole);
+    question_box.setDefaultButton(pSkip);
+
+    question_box.exec();
+    auto pBtn = question_box.clickedButton();
```
daschuer, line 84: clang-tidy is suggesting: 
```suggestion
    auto* pBtn = question_box.clickedButton();
```

ronso0: :+1:

src/library/export/trackexportworker.cpp
```diff
@@ -26,10 +26,13 @@ QMap<QString, mixxx::FileInfo> createCopylist(const TrackPointerList& tracks) {
...
     QMap<QString, mixxx::FileInfo> copylist;
     for (const auto& pTrack : tracks) {
+        VERIFY_OR_DEBUG_ASSERT(pTrack != nullptr) {
+            continue;
+        }
         auto fileInfo = pTrack->getFileInfo();
         if (fileInfo.resolveCanonicalLocation().isEmpty()) {
             qWarning()
-                    << "File not found or inaccessible while exporting"
+                    << "TrackExportWorker: File not found or inaccessible while exporting"
```
Swiftb0y, line 35: wbout a QLoggingCategory instead of the prefix?

ronso0: sure. I decided on mixxx::Logger though

Swiftb0y (approved):
LGTM. wdyt @daschuer?

daschuer (approved):
Works and LGTM. Thank you.

PR #13614, opened 2024-08-30
Track File Export: add 'Apply to all' checkbox, remove ".. All" buttons

### LINKED ISSUE #13613: Align track export replace dialog with the file managers
### Feature Description

The track export feature has this dialog: 
   export-dlg-now

The file manager looks like this: 
grafik

We should align the text overwrite/replace and add the checkbox to reduce the number of buttons.

ronso0:
based on #13610, so it's only the last commit 1392b33cae056e817fc3172af6a2cb19de2d9104

Looks like this now:
track-export-checkbox

**new tr strings:**
Replace
Apply to all files

(if you like I can rebase onto 2.5)

Fixes #13613

### DIFF
--- a/src/library/export/trackexportdlg.cpp
+++ b/src/library/export/trackexportdlg.cpp
@@ -1,5 +1,6 @@
 #include "library/export/trackexportdlg.h"
 
+#include <QCheckBox>
 #include <QMessageBox>
 
 #include "moc_trackexportdlg.cpp"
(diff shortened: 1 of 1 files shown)


daschuer:
Uh that was quick. Thank you. The new texts will likely be untranslated in a 2.5 release, so let's keep this for 2.6.

src/library/export/trackexportdlg.cpp
```diff
@@ -66,30 +67,34 @@ void TrackExportDlg::slotAskOverwriteMode(
...
-    QPushButton* pSkipAll = question_box.addButton(
-            tr("Skip &All"), QMessageBox::NoRole);
     QPushButton* pOverwrite = question_box.addButton(
-            tr("&Overwrite"), QMessageBox::YesRole);
-    QPushButton* pOverwriteAll = question_box.addButton(
-            tr("Over&write All"), QMessageBox::YesRole);
+            tr("&Replace"), QMessageBox::YesRole);
     question_box.setDefaultButton(pSkip);
 
+    QCheckBox* pApplyToAll = new QCheckBox(tr("Apply to all files"));
```
acolombier, line 80: You might need to add a parent to `question_box` otherwise it would leak IIRC. Or you could also delete explicitly

src/library/export/trackexportdlg.cpp
```diff
@@ -66,30 +67,34 @@ void TrackExportDlg::slotAskOverwriteMode(
...
+        }
     } else if (pBtn == pOverwrite) {
-        promise->set_value(TrackExportWorker::OverwriteAnswer::OVERWRITE);
-    } else if (pBtn == pOverwriteAll) {
-        promise->set_value(TrackExportWorker::OverwriteAnswer::OVERWRITE_ALL);
+        if (pApplyToAll->isChecked()) {
+            promise->set_value(TrackExportWorker::OverwriteAnswer::OVERWRITE_ALL);
+        } else {
+            promise->set_value(TrackExportWorker::OverwriteAnswer::OVERWRITE);
+        }
```
acolombier, line 97: nitpick: could be great if we could prevent the nested ifs with a ternary

ronso0:
Thanks for your review!
All done I guess.

acolombier (approved):
LGTM! I'm tempted to try the `/softfix` command but scared to make a mess...

ronso0:
Well, let's see...

ronso0:
Looking good! :tada:

ronso0:
Though CI hasn't been triggered (clazy was still running when I use softfix)

acolombier:
> Though CI hasn't been triggered (clazy was still running when I use softfix)

Ah yeah, this is because we are using the default github token (`[github-actions](https://github.com/apps/github-actions)`, which cannot trigger sub-sequent builds). Might want to create a `mixxx-bot` app soon!

PR #13615, opened 2024-08-30
(usability) Waveform prefs: group options, adjust tabstops, reorder ui file

ronso0:
This improves usability in the waveform preferences by grouping the options:
Waveforms / Overview / Caching /OpenGl info

**edit** Additionally I fixed how the Waveform options are enabled/disabled when toggling the 'Enabled' checkbox.
The gain options are enabled if either Waveforms are enabled or 'Normalize overview waveform' is disabled.

pref-waveforms_75

The checkbox
`    [  ] Enabled`
looks a bit weird in the 2nd column. Could as well be
`Enabled [  ]` (in 1st column)

Note:
Unfortunately there is an issue with QGroupBox: it seems to have some minimal height which is not reached with 3 items, which in turn unnecessarily expands the group box. A spacer at the bottom of the page fixes this, downside is the extra space at the bottom (didn't manage to minimize that..)


Swiftb0y (commented):
can't really comment much on the ui file changes but ur in the expert in that anyways so there is not much to argue here

Swiftb0y (approved):
LGTM. thank you. Does anyone want to look at the ui file?

daschuer (commented):
Thank you. This is already an improvement. 

Here some comments: 
* The Group box lable "Wavform" can become "Scrolling Wavewform" or such, to better distinguished it from the "Waveform Overview". 
* The checkboxes around the Waveform comobox are looking spreaded. Can we put the m all in one line? 
* The title column in both group boxes are not aligned. 
* If you disable the waveforms, it feels random what is grayed out and what not. In addition the hint about missing accelerations appears. Do we need to gray out anything at all? In case of a crashing config, the user may want to keep the checkbox disabled, adjust the settings an than enable.  

If you prefer to merge it as it is, just give a hint.

ronso0:
> If you disable the waveforms, it feels random what is grayed out and what not. In addition the hint about missing accelerations appears.

I torn about this.
I think it unused / inapplicable options shoud indeed be grayed out. About the 'Requires acceleration' hint: I implemented that approach for the skin settings in LateNight, but these hints are overlays there and don't change the layout.

Therefore, IMO all (scrolling) waveform options should be grayed out.
Except... the Visual Gain settings, since they're also applied to the overviews if normalization is disabled.
(which I noticed just now, after using Mixxx for almost ten years :laughing: )
I think we should add a hint in the overview group.

ronso0:
> The Group box lable "Wavform" can become "Scrolling Wavewform" or such, to better distinguished it from the "Waveform Overview".

Yep, fixed.

ronso0:
> The checkboxes around the Waveform comobox are looking spreaded. Can we put the m all in one line?

Fixed.

ronso0:
All good now I think.

Additionally I fixed how the Waveform options are enabled/disabled when toggling the 'Enabled' checkbox.
The gain options are enabled if either Waveforms are enabled or 'Normalize overview waveform' is disabled.

ronso0:
Also, what do you think about making the marker hints options layout more compact?
This allows to shrink the preferences window (without horizontal scrollbars) and see more of how the waveforms settings look like in the GUI.

before:
image


new:
image

m0dB:
I like that more compact layout, @ronso0 . Note only more compact, also more clear.

ronso0:
@m0dB  You mean the "before" image?
That's only "more compact" in height.

My intention was to make the preferences require less horizontal screen space. Then you can shrink them more and see more of the Mixxx GUI / don't need horizontal scrollbars (which are a PITA IMHO).
In the old layout it was better that the checkboxes are aligned vertically.

Happy to revert it if it helps getting this merged.

### Vote
restore the old layout for "beats/time until next.." or keep the new one? see https://github.com/mixxxdj/mixxx/pull/13615#issuecomment-2537529510
:+1: keep new // :-1: revert

ronso0:
@mixxxdj/developers anyone still wants to vote above?

Is there anything left to do?

acolombier:
I think we got a consensus here!  Shipping :rocket:

PR #13619, opened 2024-09-01
AutoDj: Option to start the Crossfader centered so the track starts at full volume (default off).

davidlmorris:
Option to set the Crossfader to the center as AutoDj starts a crossfade so the track starts at full volume.

This overcomes the issue of a track fading in, so that it loses the start, especially on cold start vocals or drum fills.  This sounds much more professional, especially for community radio or internet stations.

The setting is in preferences Auto DJ under Crossfader Behaviour (noting that the hint also applies to this setting).
image

When set all of the AutoDj settings are affected, however, despite starting from the middle the crossfade still takes the same time.


src/library/autodj/autodjprocessor.cpp
```diff
@@ -1343,6 +1358,7 @@ void AutoDJProcessor::calculateTransition(DeckAttributes* pFromDeck,
     }
 
     switch (m_transitionMode) {
+//    case TransitionMode::RadioFullIntroOutro:
     case TransitionMode::FullIntroOutro: {
```
Swiftb0y: remove?

src/library/autodj/autodjprocessor.h
```diff
@@ -285,9 +285,11 @@ class AutoDJProcessor : public QObject {
     double m_transitionProgress;
     double m_transitionTime; // the desired value set by the user
     TransitionMode m_transitionMode;
+    bool m_crossfaderStartCenter;
 
     QList<DeckAttributes*> m_decks;
 
+    
     ControlProxy* m_pCOCrossfader;
```
Swiftb0y, line 293: remove

daschuer:
Thank you for this PR. I wonder what exactly is your use case and if this plays nicely with all transition modes and configured intros. I can confirm that throwing a track in at full volume is a good solution when they start with vocals right away. What your use case?

src/library/autodj/autodjprocessor.cpp
```diff
@@ -566,6 +559,8 @@ AutoDJProcessor::AutoDJError AutoDJProcessor::toggleAutoDJ(bool enable) {
                 setCrossfader(1.0);
             }
         }
+        m_crossfaderStartCenter = m_pConfig->getValue<bool>(ConfigKey(kConfigKey,
+                QStringLiteral("center_xfader_when_starting_xfade")));
```
ronso0, line 563: please set the default here
```suggestion
        m_crossfaderStartCenter = m_pConfig->getValue<bool>(ConfigKey(kConfigKey,
                QStringLiteral("center_xfader_when_starting_xfade")),
                false);
```

src/library/autodj/autodjprocessor.cpp
```diff
@@ -859,17 +855,28 @@ void AutoDJProcessor::playerPositionChanged(DeckAttributes* pAttributes,
...
                 // the transition immediately.
-                double remainingCrossfader = crossfaderTarget - currentCrossfader;
-                double adjustment = remainingCrossfader /
-                        (1.0 - m_transitionProgress) * transitionStep;
-                // we move the crossfader linearly with
-                // movements in this track's play position.
-                setCrossfader(currentCrossfader + adjustment);
+                if ((m_crossfaderStartCenter) &&
+                        (transitionProgress <= ((thisDeck->fadeEndPos - thisDeck->fadeBeginPos) / 2))) {
+                    // We want the cross fader to remain in the middle.
```
ronso0, line 870: It'd be good to explain the purpose here.
Something along those lines, feel free to edit.
```suggestion
                    // For the first half of the transition we want the "from" deck
                    // to play at full volume like with a full crossfade transition,
                    // hence we keep the crossfader in the middle.
```

ronso0 (changes requested):

ronso0:
> There are style changes in unrelated regions if the code. Please make sure that your editor does not auto-format the whole file. This keeps the PR small and easy to review.

Some of those changes look liek pre-commit proposals. Do you have that installed?

src/library/autodj/autodjprocessor.cpp
davidlmorris: Yes.  This was an artifact of an earlier experiment.  I thought I had removed all of those.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -859,17 +862,29 @@ void AutoDJProcessor::playerPositionChanged(DeckAttributes* pAttributes,
...
                 // the transitions. If there has been a seek beyond endPos, end
                 // the transition immediately.
-                double remainingCrossfader = crossfaderTarget - currentCrossfader;
-                double adjustment = remainingCrossfader /
-                        (1.0 - m_transitionProgress) * transitionStep;
-                // we move the crossfader linearly with
-                // movements in this track's play position.
-                setCrossfader(currentCrossfader + adjustment);
+                if ((m_crossfaderStartCenter) &&
+                    (transitionProgress <= ((thisDeck->fadeEndPos - thisDeck->fadeBeginPos)/2))) {
```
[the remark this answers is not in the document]
davidlmorris: If you mean to use the standard lib calls instead of the arithmetic, then sure, that makes the intent clearer.  (My C++ usage has been mostly on micro-controllers where one avoids both the speed and space overhead of libraries if possible.)

src/library/autodj/autodjprocessor.h
davidlmorris: I don't think m_pCOCrossfader has anything to do with my code, other than I used it as a template for some of my changes.  I think m_pCOCrossfader refers to https://github.com/mixxxdj/mixxx/pull/13303.

src/library/autodj/autodjprocessor.cpp
davidlmorris: Sure.   Note that ...what happens is the fader moves to the mid, and then the transition from mid to the end position takes the full duration of the change (which is much neater than my intent!).  This code (and comment)  was to make sure that a re-calculation of 'transitionProgress' didn't drop us back to (near) the start.

davidlmorris:
> > There are style changes in unrelated regions if the code. Please make sure that your editor does not auto-format the whole file. This keeps the PR small and easy to review.
> 
@ronso0 
> Some of those changes look liek pre-commit proposals. Do you have that installed?

I did for the second attempt.  I had thought that the clang-format would be tied into the VS settings, and because I hadn't run it this was the cause of the original reject.   [Lots of gratuitous commentary of code formatting redacted here.]

Most of this tool set (including git really) is pretty alien to me, so I am probably making quite a mess of this.  

However, the code changes are simple and do work across all four AutoDj options.

ronso0:
Thanks, I'll take another look soon.

FYI: resoving review conversations should be done by the reviewer. Otherwise (contributor) responses are hidden and it's tedious for the reviewer to expand and double-check previous reviews. Thanks!

daschuer:
Can this one be closed in favour of: https://github.com/mixxxdj/mixxx/pull/13628?

davidlmorris:
Yes.  Change addressed in 13628(https://github.com/mixxxdj/mixxx/pull/13628).

davidlmorris:
Change addressed in 13628(https://github.com/mixxxdj/mixxx/pull/13628).

PR #13622, opened 2024-09-02
Modernize Hid/Bulk Lists

Swiftb0y:


src/controllers/hid/hiddenylist.h
```diff
@@ -1,21 +1,37 @@
...
+// TODO: unify this with the invalid interfacenumber from the bulkenumerator
+constexpr static std::int32_t kInvalidInterfaceNumber = -1;
+constexpr static std::int32_t kBlanketValue = 0x0;
+
+struct hid_denylist_t {
+    std::uint16_t vendor_id;
+    std::uint16_t product_id;
+    std::uint16_t usage_page;
+    std::uint16_t usage;
+    std::int32_t interface_number = kInvalidInterfaceNumber;
```
JoergAtGithub, line 14: Why changing the type here? Isn't is safer to compare int with int than int with std::int32_t ?

Swiftb0y: I assumed it was defined to always be a 32-bit int (similar to how the vid/pid are always 16-bit), though I didn't bother checking. Revert to int?

Swiftb0y: I don't know any architecture we suport that defines `sizeof(int) != 32 bit` so it probably doesn't matter.

JoergAtGithub: Propably it doesn't matter, as common USB devices have only very few interfaces. But why change the type, if we just compare it? The header which defines it is here: https://github.com/libusb/hidapi/blob/c3c79a7d82933fb9642f89046745aacc5454f438/hidapi/hidapi.h#L178

Swiftb0y: Well, my though process was basically that the data here carried is defined by a protocol that is [independent of the data model](https://en.cppreference.com/w/cpp/language/types#Properties). I assumed the data source had similar fixed-width definitions (I didn't look at the hidapi definition previously). I see it makes sense to match the "upstream" definition. I'll revert.

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,34 @@
+#pragma once
+
+#include <cstdint>
+
 // A list of supported USB bulk devices
 
-#pragma once
+static constexpr std::uint32_t kInvalidInterfaceNumber = 0;
```
JoergAtGithub, line 7: Are you sure that 0 is an invalid interface number? AFAIK the USB interface is counted zero based.

Swiftb0y: I didn't look too much into it either, but thats what I gathered from this code
https://github.com/mixxxdj/mixxx/blob/b2ff633773defdf6ede84c3a9349bba87ea4160b/src/controllers/bulk/bulkcontroller.cpp#L180-L203

But thats also defined on windows, even though `libusb_kernel_driver_active` is [not available on windows](https://libusb.sourceforge.io/api-1.0/group__libusb__dev.html#ga1cabd4660a274f715eeb82de112e0779)... so I'm not sure in how good of a shape that code is...

JoergAtGithub: The code for kernel driver unloading was introduced in #13008. According to the documentation linked by you, this is neither functional nor necessary on Windows.
But it's seems to be a validated argument in BulkController::open(). There 0 means the first interface of the USB device and not InvalidInterface number.
While Bulk is a part of the transport layer (OSI Level 4) of USB, the HID protocol is in the application layer (OSI level 7). This is why HID can use other transport layers than USB. And this is the reason why the USB interface number can be invalid for HID, but never for USB Bulk.

Swiftb0y: right, but that doesn't explain why the interface logic is operating-system specific...

JoergAtGithub: Indeed, but my remark was that `kInvalidInterfaceNumber` does not match here.

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -148,20 +150,18 @@ int BulkController::open() {
...
 
     /* Look up endpoint addresses in supported database */
-    int i;
-    for (i = 0; bulk_supported[i].vendor_id; ++i) {
-        if ((bulk_supported[i].vendor_id == m_vendorId) &&
-                (bulk_supported[i].product_id == m_productId)) {
-            m_inEndpointAddr = bulk_supported[i].in_epaddr;
-            m_outEndpointAddr = bulk_supported[i].out_epaddr;
+    const bulk_supported_t* device = std::find(
+            std::begin(bulk_supported), std::end(bulk_supported), bulk_supported_t {
```
daschuer, line 154: Before only the to vendor and product ID have been compared, now this is a deep compare. Is there a reason? I suggest to restore the old faster lookup via a lambda. 

Regarding the code style, I think we should put the { without space to the type to match a constructor call style. Is there a way to put that into clang-format?

Swiftb0y: > Before only the to vendor and product ID have been compared, now this is a deep compare. Is there a reason? I suggest to restore the old faster lookup via a lambda.

whops, I completely misread the code here, of course that makes more sense. 

> Regarding the code style, I think we should put the { without space to the type to match a constructor call style. Is there a way to put that into clang-format?

Yes, there is [SpaceBeforeCpp11BracedList](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#spacebeforecpp11bracedlist)

Swiftb0y: fixed

src/controllers/bulk/bulksupported.h
Swiftb0y: I agree. I removed the named constant.

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -148,23 +150,20 @@ int BulkController::open() {
...
-            break;
-        }
-    }
 
-    if (bulk_supported[i].vendor_id == 0) {
+    const bulk_supported_t* device = std::find_if(
+            std::begin(bulk_supported), std::end(bulk_supported), [this](const auto& dev) {
+                return dev.vendor_id == m_vendorId && dev.product_id == m_productId;
+            });
+    if (!device) {
```
daschuer, line 158: To does that work, I think it his points after the last element and should be:
```suggestion
    if (device == std::end(bulk_supported)) {
```

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -148,23 +150,20 @@ int BulkController::open() {
...
-            m_outEndpointAddr = bulk_supported[i].out_epaddr;
-#if defined(__WINDOWS__) || defined(__APPLE__)
-            m_interfaceNumber = bulk_supported[i].interface_number;
-#endif
-            break;
-        }
-    }
 
-    if (bulk_supported[i].vendor_id == 0) {
+    const bulk_supported_t* device = std::find_if(
```
daschuer, line 154:
```suggestion
    const bulk_supported_t* pDevice = std::find_if(
```

src/controllers/bulk/bulkcontroller.cpp
Swiftb0y: whoops, yeah... just saw that it returned a plain pointer and checked for null out of habit.

Swiftb0y: done

src/controllers/bulk/bulkcontroller.cpp
Swiftb0y: done

Swiftb0y:
anything else? can I squash the fixups?

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,20 @@
...
-        {0x06f8, 0xb107, 0x83, 0x03, 0}, // Hercules Mk4
-        {0x06f8, 0xb100, 0x86, 0x06, 0}, // Hercules Mk2
-        {0x06f8, 0xb120, 0x82, 0x03, 0}, // Hercules MP3 LE / Glow
-        {0, 0, 0, 0, 0}};
+constexpr static bulk_supported_t bulk_supported[] = {
+        {0x06f8, 0xb105, 0x82, 0x03}, // Hercules MP3e2
+        {0x06f8, 0xb107, 0x83, 0x03}, // Hercules Mk4
+        {0x06f8, 0xb100, 0x86, 0x06}, // Hercules Mk2
+        {0x06f8, 0xb120, 0x82, 0x03}, // Hercules MP3 LE / Glow
+};
```
JoergAtGithub, line 20: I think the interface number should be handled like in_epaddr and out_epaddr. This is a lookup table to get these 3 device parameters for the keypair vendor_id/product_Id. It makes the table difficult to read, if the 0 is a default value.
It's just random that Hercules devices use the interface number 0 for bulk and 1 for the audio interface. 0 is no special value here, it can be also enumerated the other way around for devices of another manufacturer.

Swiftb0y: makes sense.

Swiftb0y: done

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,20 @@
...
+#include <cstdint>
+
+// A list of supported USB bulk devices
+
+struct bulk_supported_t {
+    std::uint16_t vendor_id;
+    std::uint16_t product_id;
+    std::uint8_t in_epaddr;
+    std::uint8_t out_epaddr;
+    std::uint32_t interface_number;
```
JoergAtGithub, line 12:
```suggestion
    std::uint8_t interface_number;
```
One byte is enough here: https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/libusb.h#L749

Swiftb0y: done

JoergAtGithub:
LGTM now!

Swiftb0y:
Thanks. waiting for @daschuer

JoergAtGithub (approved):

src/controllers/bulk/bulkenumerator.cpp
```diff
@@ -21,14 +21,14 @@ BulkEnumerator::~BulkEnumerator() {
...
-    }
-    return false;
+static bool is_interesting(const libusb_device_descriptor& desc) {
+    auto vendorId = desc.idVendor;
+    auto productId = desc.idProduct;
+    return std::any_of(std::cbegin(bulk_supported),
+            std::cend(bulk_supported),
+            [=](const auto& dev) {
+                return dev.vendor_id == vendorId && dev.product_id == productId;
+            });
```
daschuer, line 31: This is a bit verbose without giving infos. How about this: 
```suggestion
    return std::any_of(std::cbegin(bulk_supported),
            std::cend(bulk_supported),
            [&](const auto& dev) {
                return dev.vendor_id == desc.vendorId && 
                    dev.product_id == desc.productId;
            });
```

src/controllers/hid/hiddenylist.h
```diff
@@ -1,21 +1,35 @@
...
 /// USB HID device that should not be recognized as controllers
-constexpr hid_denylist_t hid_denylisted[] = {
-        {0x1157, 0x300, 0x1, 0x2, -1},  // EKS Otus mouse pad (OS/X,windows)
-        {0x1157, 0x300, 0x0, 0x0, 0x3}, // EKS Otus mouse pad (linux)
-        {0x04f3, 0x2d26, 0x0, 0x0, -1}, // ELAN2D26:00 Touch screen
-        {0x046d, 0xc539, 0x0, 0x0, -1}, // Logitech G Pro Wireless
+constexpr static hid_denylist_t hid_denylisted[] = {
+        {0x1157, 0x300, 0x1, 0x2},                          // EKS Otus mouse pad (OS/X,windows)
+        {0x1157, 0x300, kBlanketValue, kBlanketValue, 0x3}, // EKS Otus mouse pad (linux)
+        {0x04f3, 0x2d26, kBlanketValue, kBlanketValue},     // ELAN2D26:00 Touch screen
```
daschuer, line 19: allowing to omit the interface_number but not the usage_page and usage is not consistent. 
I think I would prefer always pass all values and revert the default in hid_denylist_t. But we can decide also the other way around.

src/controllers/hid/hidenumerator.cpp
```diff
@@ -33,34 +33,28 @@ bool recognizeDevice(const hid_device_info& device_info) {
...
+        if (denylisted.interface_number != kInvalidInterfaceNumber &&
+                device_info.interface_number != denylisted.interface_number) {
+            continue;
         }
         // Denylist entry based on usage_page and usage (both required)
-        if (denylisted.usage_page != 0 && denylisted.usage != 0) {
+        if (denylisted.usage_page != kBlanketValue && denylisted.usage != kBlanketValue) {
             // Skip matching for devices with no usage_page/usage info.
-            if (device_info.usage_page == 0 && device_info.usage == 0) {
+            if (device_info.usage_page == kBlanketValue && device_info.usage == kBlanketValue) {
```
daschuer, line 57: This condition is also redundant.

daschuer (changes requested):
some more comments.

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,20 @@
...
-    unsigned char in_epaddr;
-    unsigned char out_epaddr;
-    unsigned int interface_number;
-} bulk_supported_t;
+#include <cstdint>
+
+// A list of supported USB bulk devices
+
+struct bulk_supported_t {
+    std::uint16_t vendor_id;
```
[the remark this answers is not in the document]
Swiftb0y: > This "std::" looks untypical because libusb is a c library,

yes, but this on the other hand is C++ code and in C++ code, the fixed width integer types here are in fact in the std namespace.

> We can be sure that the types are matching but they are not the "same"

The have the same bit width and signedness. I would consider them the same.

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,20 @@
...
+    std::uint16_t product_id;
+    std::uint8_t in_epaddr;
+    std::uint8_t out_epaddr;
+    std::uint8_t interface_number;
```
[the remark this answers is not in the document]
Swiftb0y: where do you get that from? its uint8 in the relevant struct. https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/libusb.h#L749

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,20 @@
...
-} bulk_supported_t;
+#include <cstdint>
+
+// A list of supported USB bulk devices
+
+struct bulk_supported_t {
+    std::uint16_t vendor_id;
+    std::uint16_t product_id;
+    std::uint8_t in_epaddr;
+    std::uint8_t out_epaddr;
```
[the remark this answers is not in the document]
Swiftb0y: should be the same...

src/controllers/hid/hiddenylist.h
Swiftb0y: > allowing to omit the interface_number but not the usage_page and usage is not consistent.

why?

Swiftb0y:
whoops. sorry misclick, didn't mean to delete your review comment.

src/controllers/hid/hiddenylist.h
```diff
@@ -1,21 +1,35 @@
 #pragma once
 
-typedef struct hid_denylist {
+// TODO: unify this with the invalid interfacenumber from the bulkenumerator
+constexpr static int kInvalidInterfaceNumber = -1;
+constexpr static unsigned short kBlanketValue = 0x0;
```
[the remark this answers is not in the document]
Swiftb0y: > However there are devices which report the -1

Are you sure? that violates the assumption hidapi makes:
https://libusb.info/hidapi/structhid__device__info.html#a9163d8d5d7db8dc47bddfaf876e17547

src/controllers/hid/hiddenylist.h
```diff
@@ -1,21 +1,35 @@
...
+
+struct hid_denylist_t {
     unsigned short vendor_id;
     unsigned short product_id;
     unsigned short usage_page;
     unsigned short usage;
-    int interface_number;
```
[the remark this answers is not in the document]
Swiftb0y: I don't quite follow you here, these are the hidapi types. What do you want me to do?

src/controllers/hid/hidenumerator.cpp
Swiftb0y: done

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: It is OK to keep it here. 
In general It is a good idea to not change types used in a library. This allows us just to include libusb.h without any additional headers and all types are available without implicit casting. In this case this is only, a marginal difference.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: See my reasoning above.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: strange, everywhere else in libusb.h it is an int.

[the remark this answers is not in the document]
daschuer: You point to bInterfaceNumber which is a byte. Everywhere else it is used as int because of -1 for invalid.

src/controllers/hid/hiddenylist.h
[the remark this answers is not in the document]
daschuer: vendor_id  is always the same 16 bit type. 
One time we picked equivalent c++ types and the other time you adopt the type used in the library. 

It is interesting that libusb is not consistent with itself here. Do we mix hid and libusb structs somewhere?

src/controllers/hid/hiddenylist.h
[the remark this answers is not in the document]
daschuer: This reads as if non HID devices are -1. 
Somewhere else I have read that the interface number is Linux only. I don't understand the issue fully.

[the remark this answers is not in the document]
JoergAtGithub: No, this reads that non-USB devices are -1, because only USB interfaces can have a valid interface number. HID is a high level protocol, that can be used with many other transport protocols, which do not know the concept of enumerated interfaces.
To my knowledge, an USB device can not have any other interface number than 0...255. This is why uint8 is enough for the lookup table of USB bulk interfaces. Invalid is impossible in an enumeration and it's not a filter, where a don't-care would make sense.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
Swiftb0y: I agree. In the end these are just type aliases anyway and if we're worried about implicit conversions we should look into `-Wnarrowing`.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
Swiftb0y: right, so `int16_t` or `std::optional<std::uint8_t>` or `int`? for the `int` you'll have to battle it out with @JoergAtGithub because [he requested it](https://github.com/mixxxdj/mixxx/pull/13622#discussion_r1747537727)

src/controllers/hid/hiddenylist.h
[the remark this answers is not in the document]
Swiftb0y: > It is interesting that libusb is not consistent with itself here. Do we mix hid and libusb structs somewhere?

Pretty sure we don't. 

I agree, this inconsistency is weird, but probably an indicator that we're overthinking this way too much...

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: I am against  `std::optional<std::uint8_t>` because we need to convert that to an `int` before calling libusb functions. 
int16_t suits here when it is not narrowed form int. 

Here it is stored as unsigned int which is even more confusing, and probably wrong. 
```
#if defined(__WINDOWS__) || defined(__APPLE__)
    m_interfaceNumber = pDevice->value.interface_number;
#endif
```
This may reveals some insides. 
https://github.com/libusb/hidapi/commit/bd6be4d83b36d0dcd5401fb335384586b066a2eb
Is the guard excluding Linux still correct?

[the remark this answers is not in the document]
daschuer: This seems to be a bug at least: 
https://github.com/mixxxdj/mixxx/blob/564a31845a8695cc7fcedbca0fe5c705a9c10bf5/src/controllers/bulk/bulkcontroller.cpp#L180C30-L180C57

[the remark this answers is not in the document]
JoergAtGithub: This is code for the USB Bulk devices, but you're referencing hidapi. This is an unrelated library.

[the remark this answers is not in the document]
daschuer: Here an unsigned int is narrowed to a signed int and the invalid -1 value is not considered: 
`if (m_interfaceNumber && libusb_kernel_driver_active(m_phandle, m_interfaceNumber) == 1) {`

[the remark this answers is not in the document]
daschuer: https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/core.c#L1823C50-L1823C67
The interface_numbers 0...31 are valid.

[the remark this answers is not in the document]
JoergAtGithub: I will not insist on uint8, but Invalid can not occur here. The sole source of information is our lookuptable. And you can't access an USB device without valid interface number. Adding -1 to an entry in our lookuptable would be a bug on it's own.

If we change it to int, libusb_kernel_driver_active will typecast it to uint8 internally anyway, so it shouldn't make a difference: https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/core.c#L2084-L2099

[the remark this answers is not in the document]
JoergAtGithub: My point in the original review was that the type for interface_number was changed from `unsigned int` to `std::uint32_t` and hardcoding the fixed size of 32 bit seemed misleading to me.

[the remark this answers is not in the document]
daschuer: I can confirm that. libusb is using uint8_t and int. I think we need to adjust our code to that. 
I have not yet understand Linux situation.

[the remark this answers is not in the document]
daschuer: We have the option to revert the change here and carry on, there was no user visible issue with the old types, was it? 

If we decide for a new type here we may adopt 
int, because the API will convert the given value to int anyway. What does the 0 below actually mean? Reading the code and the comment, it could be an unused value what we may want to indicate with -1 or is it actually interface 0? 

Where is the term "tweak" coming from and why is it referenced as value in the structure and not as tweak?

src/controllers/hid/hiddenylist.h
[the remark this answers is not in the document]
daschuer: The limit seems to be 0...32.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: The interface number along with 0 have been introduced in a bulk action here: 

https://github.com/mixxxdj/mixxx/commit/5e162f8c5ecc866c38385d82ee28ea89448c53c1

What does 0 mean? Why are all 0?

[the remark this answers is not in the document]
Swiftb0y: > We have the option to revert the change here and carry on, there was no user visible issue with the old types, was it?

Not really. I can revert if that is seriously a condition to merge...

> What does the 0 below actually mean? Reading the code and the comment, it could be an unused value what we may want to indicate with -1 or is it actually interface 0?

Its the default. 0 on most devices, but not all (thus the table). 

> Where is the term "tweak" coming from and why is it referenced as value in the structure and not as tweak?

Because those are the nonstandard values we actually want to tweak. I can also call it "quirk", which is the terminology the linux kernel uses for this sort of thing.

> Why are all 0?

Because the only device that has a non-zero interface has not been added yet. 
https://github.com/acolombier/mixxx/blob/881a482f2000bf888fd9599922129981b00ac8aa/src/controllers/bulk/bulksupported.h#L13-L18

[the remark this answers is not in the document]
JoergAtGithub: It's just numbering of the interfaces. We currently just support bulk devices from the manufacturer Hercules. And Hercules always use the first interface for bulk and the second for audio. This order is completely up to the manufacturer.

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,28 @@
...
+    std::uint16_t product_id;
+};
+
 // A list of supported USB bulk devices
 
-#pragma once
+struct bulk_device_tweak {
+    std::uint8_t in_epaddr;
+    std::uint8_t out_epaddr;
+    std::uint8_t interface_number;
```
daschuer, line 15: Conclusion "std::uint8_t" is OK

daschuer (changes requested):
It looks like, we have discovered an issue ..

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,28 @@
...
-        {0x06f8, 0xb105, 0x82, 0x03, 0}, // Hercules MP3e2
-        {0x06f8, 0xb107, 0x83, 0x03, 0}, // Hercules Mk4
-        {0x06f8, 0xb100, 0x86, 0x06, 0}, // Hercules Mk2
-        {0x06f8, 0xb120, 0x82, 0x03, 0}, // Hercules MP3 LE / Glow
-        {0, 0, 0, 0, 0}};
+constexpr static bulk_support_lookup bulk_supported[] = {
+        {{0x06f8, 0xb105}, {0x82, 0x03, 0}}, // Hercules MP3e2
+        {{0x06f8, 0xb107}, {0x83, 0x03, 0}}, // Hercules Mk4
+        {{0x06f8, 0xb100}, {0x86, 0x06, 0}}, // Hercules Mk2
+        {{0x06f8, 0xb120}, {0x82, 0x03, 0}}, // Hercules MP3 LE / Glow
```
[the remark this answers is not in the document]
Swiftb0y: I agree, but I'm doubtful there is much to fix if we don't know what and have no hardware to test.

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,28 @@
...
 
-static bulk_supported_t bulk_supported[] = {
-        {0x06f8, 0xb105, 0x82, 0x03, 0}, // Hercules MP3e2
-        {0x06f8, 0xb107, 0x83, 0x03, 0}, // Hercules Mk4
-        {0x06f8, 0xb100, 0x86, 0x06, 0}, // Hercules Mk2
-        {0x06f8, 0xb120, 0x82, 0x03, 0}, // Hercules MP3 LE / Glow
-        {0, 0, 0, 0, 0}};
+constexpr static bulk_support_lookup bulk_supported[] = {
+        {{0x06f8, 0xb105}, {0x82, 0x03, 0}}, // Hercules MP3e2
+        {{0x06f8, 0xb107}, {0x83, 0x03, 0}}, // Hercules Mk4
```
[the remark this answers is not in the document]
Swiftb0y: no idea. @JoergAtGithub may know.

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,28 @@
...
+
+struct bulk_device_id {
+    std::uint16_t vendor_id;
+    std::uint16_t product_id;
+};
+
 // A list of supported USB bulk devices
 
-#pragma once
+struct bulk_device_tweak {
```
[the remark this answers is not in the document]
Swiftb0y: sure...

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,28 @@
...
-typedef struct bulk_supported {
-    unsigned short vendor_id;
-    unsigned short product_id;
-    unsigned char in_epaddr;
-    unsigned char out_epaddr;
-    unsigned int interface_number;
-} bulk_supported_t;
+struct bulk_support_lookup {
+    bulk_device_id key;
+    bulk_device_tweak value;
```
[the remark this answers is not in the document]
Swiftb0y: done (went for endpoints though).

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
Swiftb0y: done

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: @acolombier Do you have a device to test it? 

My idea is that we need to remove all conditional code and just let libusb do the OS abstarction.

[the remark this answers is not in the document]
Swiftb0y: That would be ideal, yes.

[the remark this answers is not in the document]
daschuer: @acolombier 

Here, you have introduces a special treatment for interface 0. However it turns out taht interface 0 is a valid interface. 
I have dived to the libusb and there is no hint that 0 is special. 

https://github.com/mixxxdj/mixxx/blob/b2ff633773defdf6ede84c3a9349bba87ea4160b/src/controllers/bulk/bulkcontroller.cpp#L180-L203

in libusb an invalid interfaces number is -1. This is used to skip the calls. 

When I connect all thi infos, it looks like we need to introduce a -1 for interface to skip the code. 
We can also remove all conditional calls and instead handle the return values. 
It is: 
LIBUSB_ERROR_INVALID_PARAM = -2, if interface -1
LIBUSB_ERROR_NO_DEVICE = -4, Device is missing 
LIBUSB_ERROR_NOT_SUPPORTED = -12, e.g. on windows-  

This means also that we must use in the table above -1 for retaining the old behaviour that might be broken on windows, until we know the real interface number and enable the device on windows. What do you think?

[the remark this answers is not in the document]
daschuer: We may postpon the fix here, by just using `int` as interface_number and replace all unknown interface numbers by -1 in the table.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: At least untill all numbers are known we need here int(-1) for invalid. see my other comment.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: Workaround: Add a TODO, and use -1 as interfac number.

src/controllers/hid/hiddenylist.h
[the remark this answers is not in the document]
daschuer: I can confirm. Resolved.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
Swiftb0y: > This means also that we must use in the table above -1 for retaining the old behaviour that might be broken on windows, until we know the real interface number and enable the device on windows. What do you think?

Why is the situation on windows special?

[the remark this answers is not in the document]
daschuer: I don't know, it looks like it was working on Linux without specifying the interface when it has been introduced. 

I hope @acolombier can give some insight as he has enabled bulk support for windows here:
https://github.com/mixxxdj/mixxx/commit/5e162f8c5ecc866c38385d82ee28ea89448c53c1

Swiftb0y:
done, though I can't test for regressions easily.

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -175,31 +171,18 @@ int BulkController::open() {
...
-        int ret = libusb_claim_interface(m_phandle, m_interfaceNumber);
-        if (ret < 0) {
-            qCWarning(m_logBase) << "Cannot claim interface for" << getName()
-                                 << ":" << libusb_error_name(ret);
-            libusb_close(m_phandle);
-            return -1;
-        } else {
-            qCDebug(m_logBase) << "Claimed interface for" << getName();
-        }
+    if (int ret = libusb_claim_interface(m_phandle, m_interfaceNumber); ret < 0) {
```
daschuer, line 178: Let's avoid this confusing style with an assignment in an if condition.

src/controllers/bulk/bulkcontroller.h
```diff
@@ -72,13 +72,12 @@ class BulkController : public Controller {
...
-    unsigned char m_inEndpointAddr;
-    unsigned char m_outEndpointAddr;
-#if defined(__WINDOWS__) || defined(__APPLE__)
-    unsigned int m_interfaceNumber;
-#endif
+    std::uint16_t m_vendorId;
+    std::uint16_t m_productId;
+    std::uint8_t m_inEndpointAddr;
+    std::uint8_t m_outEndpointAddr;
+    std::uint8_t m_interfaceNumber;
```
daschuer, line 79: This should become an int, to match the usbapi type and to allow us to retain the 2.5 behaviour for the devices where we don't know the interface number by setting it to -1.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
acolombier: I have the S4 Mk3 to test, which has non-zero interface. I have add the interface number, which was missing in the previous implementation, so I needed a none value to exclude the other device on which I couldn't confirm the value.

>  // This functionality is not available on Windows.

Without `libusb_kernel_driver_active` and `libusb_claim_interface`, I was unable to use my device on Windows. Note I have tested on a VM so perhaps there is some different behaviour to expect?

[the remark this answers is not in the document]
acolombier: Also, note that `libusb_kernel_driver_active`/`libusb_claim_interface` should also be used on Linux. The reason for me excluding it is that the original Linux-only implementation didn't include it, and as I was implementing for Mac & Windows , I changed only the logic on these platform, but it appears to be a global oversight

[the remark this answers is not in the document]
daschuer: `libusb_claim_interface()` is the one that is available on windows and probably the required one in your case.

src/controllers/bulk/bulkcontroller.cpp
Swiftb0y: Why is this confusing? Its just a new syntax which makes sense here since I don't need `ret` outside the if-body
https://en.cppreference.com/w/cpp/language/if#if_statements_with_initializer

src/controllers/bulk/bulkcontroller.h
Swiftb0y: std::optional<std::uint8_t> would be so much more appropriate though.

daschuer: Yes, when starting on a plan surface. 

Libusb has decided to use -1 for invalid, so we should IMHO not introduce another concept. The API will just work with -1 and has internal checks, no need for any has_value() checks, in our code.

Swiftb0y: > Libusb has decided to use -1 for invalid,

I was not able to find that definition. Can you link it to me?

> The API will just work with -1 and has internal checks, no need for any has_value() checks, in our code.

Well, if we want to preserve how it works currently though, we need to check for -1 anyways, so we might as well check for has_value. I'm not sure if passing -1 deliberately is a good choice because it'll only make it more complicated to then handle the error (because we need to then distinguish from an error that occurred because we passed in -1 or we actually did pass in a garbage value we didn't account for).

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
JoergAtGithub: > Interestingly the MK4 is reported here at: https://mixxx.discourse.group/t/hercules-mk4-hid-under-linux-and-1-11/13447/3 {0x83, 0x02, 0} and {0x83, 0x03, 5} Can we have the endpoint address twice? Is that the same end point?

This is not the same USB endpoint, because only the address `0x83` is identical, but not the number of the endpoint (`0x02` vs. `0x03`).

Endpoints and Interfaces are fields on different USB protocol layers:

- Endpoints are raw hardware addresses which the operating systems USB kernel driver uses to prioritize the data transports with guranteed latency (Isochronous or Interupt transport mode) and non real-time transports (Control or Bulk transport mode)
- Interfaces/Pipes are the a protocol level higher and depend not only on the hardware, but also on configuration from the host software (drivers like power-management or user mode applications). You can assign a Endpoint only to one USB Interface, otherwise you could get conflicting access. But an interface can run in different alternative modes (e.g. a camera can run in realtime mode for video using the endpoint Isocronous transfer mode or in high-resolution photo mode using the same endpoint in Bulk transfer mode instead).

See: https://beyondlogic.org/usbnutshell/usb5.shtml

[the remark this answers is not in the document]
daschuer: I am still confused, because we don't use endpoint number in our code only the address.
Just double checked but I don't see the double 0x83 anymore, instead I found a double 0x81 with: 
Interface 1 Endpoint Address 0x81
Interface 3 Endpoint Address 0x81
Interface 4 Endpoint Address 0x81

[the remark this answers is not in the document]
JoergAtGithub: This information is not unambiguous. If you talk about the transport layer need to specify the pair of Endpoint Address and Enpoint Number: 
endpoint

src/controllers/bulk/bulkcontroller.cpp
daschuer: Because it break the flow when reading the code. You expect an if condition like everywhere else but you read the assignment, which is every where else in the line before. This reduces IMHO maintainability and lead to misunderstandings. The benefit of the tight scope is negligible here.

src/controllers/bulk/bulkcontroller.h
daschuer: It is just used literally in libusb like here: 
https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/descriptor.c#L221
and 
https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/os/windows_winusb.c#L2927

The only related macro is USB_MAXINTERFACES

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
daschuer: It is still puzzling form me, because we do not consider the Interface number with Linux at all but it seams to work with the enpoint adress only, even without an Endpoint number.  libusb_claim_interface() is written in a way that we may claim multiple interfaces with one handel. 
https://github.com/mixxxdj/mixxx/blob/4df25086d1aa060dcc3b2793bc5be20c7380e0a1/src/controllers/bulk/bulkcontroller.cpp#L199
libusb_bulk_transfer

Here we use only the enpoint adress 
https://github.com/mixxxdj/mixxx/blob/4df25086d1aa060dcc3b2793bc5be20c7380e0a1/src/controllers/bulk/bulkcontroller.cpp#L42C18-L42C38

The underlying questions is what is the interface number for -1 and will it work like before? 
How does the driver know which endpoint shall be used if there are tw owith the same address. 
Or are they the same listed in two interfaces?

src/controllers/bulk/bulkcontroller.cpp
Swiftb0y: > The benefit of the tight scope is negligible here.

I disagree actually because if the variable does not get its own scope the variables storing the return values either need different names or the one variable needs to be reassigned. The different names is confusing because that leads me to believe that we're actually interested in this value while reassigning looks weird because the added mutability makes the code hard to reason about. 

> You expect an if condition like everywhere else but you read the assignment, which is every where else in the line before.

Well you only expect the condition there because you're not used to there being something else but this is a well designed and documented feature so its reasonable to expect it to be usable. The same applies to other "new" features such as lambdas or trailing return types.

Let me reiterate that this very much distinct from the relying on return value of `operator=` (eg. `if (int var = f()`) because that is mixing an init-statement with a condition while the latter one is clearly defined.

This is just a case of "Was der Bauer nicht kennt, frisst er nicht". 

I won't insist on using this style, but resisting to embrace new and encouraged features and styles is a common pattern in code review. This is neither rewarding nor productive.

src/controllers/bulk/bulkcontroller.h
daschuer: The change has been introduced here: 
https://github.com/libusb/libusb/commit/e9eec3a680cad3b2c9c5213fb7d60148cf6900da

All function taking int interface number and are checked for < 0 and >= USB_MAXINTERFACES so -1 is just one particular invalid interface number. 

There is: 
https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/os/windows_winusb.c#L2998-L3005

In our code is: 
```
constexpr static int kInvalidInterfaceNumber = -1;
```

That are my point voting for -1. If you still like to use an optional `std::optional<std::uint8_t>` and introduce extra checks and conversions to int, go ahead.

JoergAtGithub: No checks or conversions are needed, because the case of an invalid interface number is impossible here. As the hardcoded lookup table in Mixxx is the sole source of this information, and libusb the receiver.

I think you are confusing this with hidapi, which allows you to use transport protocols other than USB. And for these alternative transport protocols, there can't exist a valid USB interface number.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
JoergAtGithub: > I am still confused, because we don't use endpoint number in our code only the address.
> Just double checked but I don't see the double 0x83 anymore, instead I found a double 0x81 with:
> Interface 1 Endpoint Address 0x81
> Interface 3 Endpoint Address 0x81
> Interface 4 Endpoint Address 0x81

I double checked this, and I found out, that the Endpoint-Number is part of the Endpoint-Address:
Each Interfaces can have up to 16 Endpoints and the address is a single byte build as follows:

- Bits 0..3b Endpoint Number.
- Bits 4..6b Reserved. Set to Zero
- Bits 7 Direction 0 = Out, 1 = In (Ignored for Control Endpoints)

intftree

See https://beyondlogic.org/usbnutshell/usb5.shtml for reference

src/controllers/bulk/bulkcontroller.h
Swiftb0y: From what I could tell is that @daschauer wants to introduce a sentinel value (-1 currently) that determines whether we should claim the interface at all. At least the current code only claimed an interface when it was non-zero. Whether that is correct is dubious so I just implemented it to claim it unconditionally.

Swiftb0y: Anyways, I think we should try the state as currently implemented. None of us has a hercules controller anyways so its not like we can easily verify that this breaks something.

JoergAtGithub: It is impossible to transfer data to or from an USB device without specifiing the interface (there might be a default, but it must be specified somehow). This is mandatory by the standard.
But hardcoding the Endpoint Addresses is not in sense of the standard, as these should be read back from the descriptors in the device, taking into account the currently active device configuration and alternate mode of the interface .

daschuer: Claiming interfaces has been introduced recently in this PR https://github.com/mixxxdj/mixxx/pull/13008. 
The code must have been worked before at least on Linux without claiming the interfaces. I have read that the Kernel automatically detaches the driver and frees up the interface for the user space application without explicit claiming.

The libusb implementation is implemented in a way that a handle can claim one interface or more. Transferring data works with the endpoint address only, without specifying the interface. So I assume that the endpoint address alone is sufficient for transferring the data. Claiming the interface seems to be only an exclusive access thing. 

Searching the forum has revealed that the interface number is not always 0. 

> It is impossible to transfer data to or from an USB device without specifying the interface

I am afraid this is wrong. My conclusion is that claiming the interface 0 here for all devices is also wrong, since we don't have the knowledge which interface to claim. As a workaround we may restore the pre https://github.com/mixxxdj/mixxx/pull/13008 by not claiming the interface indicated by `-1` or `std::nullopt` if you wish.
So we may call the variable:  
`kClaimNoInterface = -1;`

JoergAtGithub: I agree, if we don't know the correct interface number, we shouldn't assume that the device use 0.

src/controllers/bulk/bulkcontroller.cpp
daschuer: I have considered this a bit and it turns out that this is a general code style question, we should discuss on Zulip. 

We have a plenty of places where we may consider to tighten the scope of a local variable, because we can. On the other hand we don't care about them, because it will just introduce boiler plate and nesting, interrupting the flow when reading the code. This is an issue when you only skim over the code, and especially when the if condition is broken over two lines of code.

I have no interest to ban this new statement from the Mixxx code entirely, but let's use it deliberately, backed up by a rule in our style guide. 

This is here an edge case because we have two return values that will end up in the same variable of temporary nature otherwise. For now this is a new style, that does not fit to the rest of Mixxx. I am open to decide in any direction, but  let's do it consistently. 

I will not block the PR on this point, but prefer reusing `ret` here without  a scope.

Swiftb0y:
changed it so we only claim the interface if we know it.

src/controllers/bulk/bulkcontroller.cpp
daschuer: https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/C.2B.2B17.20if.20with.20init-statement

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -175,11 +175,14 @@ int BulkController::open() {
...
-        return -1;
+    if (m_interfaceNumber.has_value()) {
+        int error = libusb_claim_interface(m_phandle, *m_interfaceNumber);
+        if (error < 0) {
+            qCWarning(m_logBase) << "Cannot claim interface for" << getName()
+                                 << ":" << libusb_error_name(error);
+            libusb_close(m_phandle);
+            return -1;
+        }
     } else {
```
daschuer, line 186: This needs to be moved to the inner scope

src/controllers/bulk/bulksupported.h
```diff
@@ -12,7 +13,8 @@ struct bulk_device_id {
 struct bulk_device_endpoints {
     std::uint8_t in_epaddr;
     std::uint8_t out_epaddr;
-    std::uint8_t interface_number;
+    // we may not know the interface, in which case we should not try to claim it.
+    std::optional<std::uint8_t> interface_number = std::nullopt;
```
daschuer, line 17: The initializer is redundant. I would  rather make it explicit in the table, that the value is unknown and we need to fill it to make it work on Windows and macOs.  In addition a brief to-do comment in the table is useful. We may also file a bug as a reminder that these devices are not yet usable with windows.

src/controllers/bulk/bulkcontroller.cpp
Swiftb0y: whoops of course. fixed

src/controllers/bulk/bulksupported.h
Swiftb0y: are we sure they only don't work under windows?

Swiftb0y: done

daschuer: I don't know. But it looks like @acolombier as added the claiming to enable the controllers in windows.

Swiftb0y: yeah I saw that. but maybe it also just didn't work under windows because libusb didn't claim the right interface automatically (since from what I could tell an interface needs to be claimed anyways in order to do any operation on it). Do we want to rely on the automatic claiming (if it exists across all backends at all)?

daschuer: It looks like libusb is not able to automatically claim an interface. The auto feature is regarding detaching the kernel driver which is not supported on Windows. 
But we need to test, I have no setup for that.

Swiftb0y: I guessed so because I saw [`auto_claim`](https://github.com/libusb/libusb/blob/467b6a8896daea3d104958bf0887312c5d14d150/libusb/os/windows_winusb.c#L613) in winusb but upon further research its not used anywhere else. I don't have a setup for it either, but I guess we can continue not claiming it until someone with the setup files an issue.

JoergAtGithub: Getting libusb running under Windows requires further changes. It is not functional yet, as we need to register a device driver (e.g. the generic WinUSB driver that comes with Windows) for the particular device type. This could be done using https://github.com/pbatard/libwdi but requires elevated user rights.  Either in the Mixxx installer or at runtime of Mixxx itself with temporary elevated rights.
This is why BULK is default off for Windows.

acolombier:
What's the status of this @Swiftb0y ? Do you need me to do further testing?

Swiftb0y:
I'll look into it. In the meantime, there were a couple review comments we tagged you on @acolombier. I would appreciate if you could take a look at those.

src/controllers/bulk/bulksupported.h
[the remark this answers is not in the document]
acolombier: > @acolombier may be able to share some more info on why detach kernel driver code is used on windows even though it doesn't work and why its not used on linux even though its at least possible.

This turned out to be the least minimal snippet of code to work on both Mac and Windows. 

> We have the option to revert the change here and carry on, there was no user visible issue with the old types, was it?

Not sure why we would revert a non-breaking issue here, but that would only impact Windows and Mac ability to use the S4 Mk3, on which effort of getting it officially supported by Mixxx has faded away, so I guess it doesn't really matter.
It is quite clear that this change was the happy path MVP to get Windows and Mac running, as some user was asking support for Mac on Zulip S4 thread. But I also don't own either and don't have the interest/money/time to do so and work on something comprehensive here.

Swiftb0y:
@acolombier so I went through this again and I think its ready. Though I'd appreciate if you could test with your S4 to ensure I didn't break anything obvious.

acolombier:
Thanks for looking into this. I will give that a test today.

daschuer (approved):
LGTM, Thank you.

acolombier:
After a good 3 hours of fighting if Windows to just work, I'm afraid I won't be able to test it - the issue is with QML and GL support in the VM so not related to any of the changes, but required to prove the correct behaviour. I also couldn't borrow the Mac I was planning to test with. 
I think overall the code looks fine so I'm happy with merging that change now, and test that in the future, maybe get someone from the community with a more realistic setup (aka not a VM) to test it!

daschuer:
Ok, @Swiftb0y can you squash, than we can merge.

JoergAtGithub:
LGTM! Thank you!

Swiftb0y:
Thank you.

PR #13623, opened 2024-09-03
(fix) WTrackTableView: assert track model, p prefix

ronso0:
followup for #13620 
(sorry for the big diff)

I'll take care of merging to 2.5/main

### DIFF
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -124,9 +124,9 @@ void WTrackTableView::slotGuiTick50ms(double /*unused*/) {
             const QModelIndexList indices = getSelectedRows();
             if (indices.size() == 1 && indices.first().isValid()) {
                 // A single track has been selected
-                TrackModel* trackModel = getTrackModel();
-                if (trackModel) {
-                    TrackPointer pTrack = trackModel->getTrack(indices.first());
+                TrackModel* pTrackModel = getTrackModel();
+                if (pTrackModel) {
+                    TrackPointer pTrack = pTrackModel->getTrack(indices.first());
                     if (pTrack) {
                         emit trackSelected(pTrack);
                     }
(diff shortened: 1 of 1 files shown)


src/widget/wtracktableview.cpp
```diff
@@ -590,44 +593,48 @@ void WTrackTableView::mouseMoveEvent(QMouseEvent* pEvent) {
...
     }
 }
 
 // Drag enter event, happens when a dragged item hovers over the track table view
 void WTrackTableView::dragEnterEvent(QDragEnterEvent * event) {
-    auto* trackModel = getTrackModel();
+    auto* pTrackModel = getTrackModel();
+    if (!pTrackModel || !event->mimeData()->hasUrls()) {
+        event->ignore();
+    }
```
Swiftb0y, line 607:
```suggestion
        event->ignore();
        return;
    }
```

ronso0: Thanks! I fixed this also in other places.

src/widget/wtracktableview.cpp
```diff
@@ -505,6 +502,10 @@ void WTrackTableView::slotShowHideTrackMenu(bool show) {
 }
 
 void WTrackTableView::contextMenuEvent(QContextMenuEvent* event) {
+    TrackModel* pTrackModel = getTrackModel();
+    if (!pTrackModel) {
+        return;
+    }
```
Swiftb0y, line 508: trackmodel is unused here. Do we really need to check it then?

src/widget/wtracktableview.cpp
```diff
@@ -837,6 +845,9 @@ void WTrackTableView::dropEvent(QDropEvent * event) {
 }
 
 QModelIndexList WTrackTableView::getSelectedRows() const {
+    if (getTrackModel() == nullptr) {
+        return {};
+    }
```
Swiftb0y, line 850: same here

src/widget/wtracktableview.cpp
ronso0: Hm, not directly. But this is called by `[Library], show_track_menu`, so this is one function a controller could call before we have track model (or just item model).

Tbh I'm not entirely happy with this PR but it's the best / most simple fix I found.
There are two aspects:
1. verify we have a selectionModel()
_could_ happen there is none and we want to throw a warning then, eg. it's theoretically_ possible to `setSelectionModel(nullptr)`
2. verify we have a model and simply return if there's none
that's the controller case I mentioned, and (again _theoretically_) it is possible such functions in other ways before the GUI is ready.
I think we should not throw an assertion then. Admittedly this only affects builds with DEBUG_ASSERTIONS_FATAL, but nevertheless such assertions are hard to debug.

src/widget/wtracktableview.cpp
ronso0: I installed this as some sort of pre-DEBUG_ASSERT exit in case someone uses getSelectedRows() without verifying there's a model. (see my explanation above)

src/widget/wtracktableview.cpp
ronso0: But I can remove it here, since it's in getSelectedRows(), too.

Swiftb0y: yeah, remove it when `getSelectedRows()` already takes care of it.

src/widget/wtracktableview.cpp
Swiftb0y: yeah keep this one then. Though I'm undecided on whether it should be a debug assert as well...

daschuer (approved):
Thank you. LGTM

PR #13627, opened 2024-09-04
modernize ControllerManager

Swiftb0y:
And a little bit of the surrounding code. This is what I have so far, though the multithreading code could use a little larger restructuring. based on (outdated) #13622.

### DIFF
--- a/src/controllers/bulk/bulkenumerator.cpp
+++ b/src/controllers/bulk/bulkenumerator.cpp
@@ -17,9 +17,9 @@ BulkEnumerator::BulkEnumerator(UserSettingsPointer pConfig)
           m_context(nullptr, &libusb_exit),
           m_pConfig(pConfig) {
     // TODO: use C++23 std::out_ptr instead
-    libusb_context* ctx;
-    libusb_init(&ctx);
-    m_context.reset(ctx);
+    libusb_context* pCtx;
+    libusb_init(&pCtx);
+    m_context.reset(pCtx);
 }
 
 BulkEnumerator::~BulkEnumerator() {
@@ -38,28 +38,28 @@ static bool is_interesting(const libusb_device_descriptor& desc) {
 
 QList<Controller*> BulkEnumerator::queryDevices() {
     qDebug() << "Scanning USB Bulk devices:";
-    libusb_device** listPtr; // contiguous array of pointers to libusb_device
-    ssize_t cnt = libusb_get_device_list(m_context.get(), &listPtr);
+    libusb_device** deviceListPointers; // contiguous array of pointers to libusb_device
+    ssize_t cnt = libusb_get_device_list(m_context.get(), &deviceListPointers);
     VERIFY_OR_DEBUG_ASSERT(cnt >= 0) {
         // this assumes that if libusb_get_device_list returned an error, no memory was allocated
         // and listPtr is still uninitialized. Unfortunately the libusb docs don't specify the state
         // of the arguments in the case of errors.
         qWarning() << "Error retrieving devicelist: " << libusb_error_name(cnt);
     }
-    auto list = std::unique_ptr<libusb_device*[],
+    auto pDeviceList = std::unique_ptr<libusb_device*[],
             decltype([](libusb_device** list) {
                 libusb_free_device_list(list, true);
-            })>(listPtr);
+            })>(deviceListPointers);
 
-    for (libusb_device* device : std::span(list.get(), cnt)) {
+    for (libusb_device* pDevice : std::span(pDeviceList.get(), cnt)) {
         struct libusb_device_descriptor desc;
 
-        libusb_get_device_descriptor(device, &desc);
+        libusb_get_device_descriptor(pDevice, &desc);
         if (!is_interesting(desc)) {
             continue;
         }
         struct libusb_device_handle* handle = nullptr;
-        int err = libusb_open(device, &handle);
+        int err = libusb_open(pDevice, &handle);
         if (err) {
             qWarning() << "Error opening a device:" << libusb_error_name(err);
             continue;
--- a/src/controllers/bulk/bulkenumerator.cpp
+++ b/src/controllers/bulk/bulkenumerator.cpp
@@ -2,23 +2,28 @@
 
 #include <libusb.h>
 
+#include <QDebug>
+#include <algorithm>
+#include <memory>
+#include <span>
+
 #include "controllers/bulk/bulkcontroller.h"
 #include "controllers/bulk/bulksupported.h"
 #include "moc_bulkenumerator.cpp"
+#include "util/assert.h"
 
 BulkEnumerator::BulkEnumerator(UserSettingsPointer pConfig)
         : ControllerEnumerator(),
-          m_context(nullptr),
+          m_context(nullptr, &libusb_exit),
           m_pConfig(pConfig) {
-    libusb_init(&m_context);
+    // TODO: use C++23 std::out_ptr instead
+    libusb_context* ctx;
+    libusb_init(&ctx);
+    m_context.reset(ctx);
 }
 
 BulkEnumerator::~BulkEnumerator() {
     qDebug() << "Deleting USB Bulk devices...";
-    while (m_devices.size() > 0) {
-        delete m_devices.takeLast();
-    }
-    libusb_exit(m_context);
 }
 
 static bool is_interesting(const libusb_device_descriptor& desc) {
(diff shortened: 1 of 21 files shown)


src/controllers/hid/hidenumerator.cpp
```diff
@@ -120,10 +118,13 @@ QList<Controller*> HidEnumerator::queryDevices() {
             continue;
         }
 
-        HidController* newDevice = new HidController(std::move(deviceInfo));
-        m_devices.push_back(newDevice);
+        m_devices.push_back(std::make_unique<HidController>(std::move(deviceInfo)));
```
Swiftb0y, line 121: not that independent of my changes, this is currently broken when querying devices multiple times because we don't clear the list from the previous query... Not sure if I should fix that here because its not feasible to test nor do I have enough non-keyboard/-mice HID devices here.

daschuer: I think we should fix it anyway. Main is still in alpha stage.

daschuer (changes requested):
some comments of #13627 apply here as well. Can you adopt them?

Swiftb0y:
> some comments of https://github.com/mixxxdj/mixxx/pull/13627 apply here as well. Can you adopt them?

I assume you mean #13622? Yes, I will rebase on that once everything is happy.

acolombier (commented):
Looks good overall! One request tho, could you make sure that updated pointer variables are prefixed with `p...` to help with readability? 

One mild concerned is around BULK, not directly related to your change: Interface also appears to be relevant on Linux, and not using them leads to `dmesg` warnings, so I'm wondering if would be worth removing the preprocessor around this. If you're keen to give it a try, I'm happy to test it with my devices to confirm it works as expected!

Swiftb0y:
Will look into both. Thanks for taking a look @acolombier

src/controllers/midi/portmidienumerator.cpp
```diff
@@ -1,13 +1,15 @@
 #include "controllers/midi/portmidienumerator.h"
 
 #include <portmidi.h>
+#include <qlatin1stringview.h>
```
acolombier, line 4: I belieb you should use `#include <QLatin1String>` instead here

acolombier, line 4: What is this used for? Looks like it doesn't exist on MacOS

Swiftb0y: whoops, yeah clangd tries its best to add the relevant includes but its gets it wrong and I don't notice it. I'll remove and force push (and look into disabling that for certain headers).

Swiftb0y: done

Swiftb0y: fyi in case any1 else stumbles over this, see https://clangd.llvm.org/guides/include-cleaner#adjusting-behavior and use `IgnoreHeaders: q.*.h` until [QTBUG-124159](https://bugreports.qt.io/browse/QTBUG-124159) is fixed.

daschuer:
Some conflicts have developed.

src/controllers/bulk/bulkenumerator.cpp
```diff
@@ -2,60 +2,73 @@
...
 
-static bool is_interesting(struct libusb_device_descriptor *desc) {
-    for (int i = 0; bulk_supported[i].vendor_id; ++i) {
-        if ((bulk_supported[i].vendor_id == desc->idVendor) &&
-            (bulk_supported[i].product_id == desc->idProduct)) {
-            return true;
-        }
-    }
-    return false;
+static bool is_interesting(const libusb_device_descriptor& desc) {
```
daschuer, line 28: This can go to an anonymous namespace

src/controllers/bulk/bulkenumerator.cpp
```diff
@@ -2,60 +2,73 @@
...
+    VERIFY_OR_DEBUG_ASSERT(cnt >= 0) {
+        // this assumes that if libusb_get_device_list returned an error, no memory was allocated
+        // and listPtr is still uninitialized. Unfortunately the libusb docs don't specify the state
+        // of the arguments in the case of errors.
+        qWarning() << "Error retrieving devicelist: " << libusb_error_name(cnt);
+    }
+    auto pDeviceList = std::unique_ptr<libusb_device*[],
+            decltype([](libusb_device** list) {
+                libusb_free_device_list(list, true);
+            })>(deviceListPointers);
```
daschuer, line 51: This looks bulky. How about creating `deviceListDeleter()` in the anonymous namespace?

src/controllers/bulk/bulkenumerator.cpp
```diff
@@ -2,60 +2,73 @@
...
-        libusb_device *device = list[i];
+    libusb_device** deviceListPointers; // contiguous array of pointers to libusb_device
+    ssize_t cnt = libusb_get_device_list(m_context.get(), &deviceListPointers);
+    VERIFY_OR_DEBUG_ASSERT(cnt >= 0) {
+        // this assumes that if libusb_get_device_list returned an error, no memory was allocated
+        // and listPtr is still uninitialized. Unfortunately the libusb docs don't specify the state
+        // of the arguments in the case of errors.
+        qWarning() << "Error retrieving devicelist: " << libusb_error_name(cnt);
+    }
+    auto pDeviceList = std::unique_ptr<libusb_device*[],
```
daschuer, line 48: Wy using only here `libusb_device*[]` and not above?

src/controllers/bulk/bulkenumerator.cpp
```diff
@@ -2,60 +2,73 @@
...
+        int err = libusb_open(pDevice, &handle);
+        if (err) {
+            qWarning() << "Error opening a device:" << libusb_error_name(err);
+            continue;
+        }
+        m_devices.push_back(std::make_unique<BulkController>(m_context.get(), handle, &desc));
+    }
+    QList<Controller*> devices;
+    devices.reserve(m_devices.size());
+    for (const auto& pController : std::as_const(m_devices)) {
```
daschuer, line 70: Can we get rid of this loop by returning a const refrences to m_devices? Before it was not an issue because of QTs implicit sharing.

src/controllers/bulk/bulksupported.h
```diff
@@ -1,18 +1,34 @@
+#pragma once
+
+#include <cstdint>
+
 // A list of supported USB bulk devices
 
-#pragma once
+static constexpr std::uint32_t kInvalidInterfaceNumber = 0;
+struct bulk_supported_t {
```
daschuer, line 8: Did you consider to make this a class along with the helper functions as members?

daschuer (changes requested):
I have left some more comments.

src/controllers/bulk/bulkenumerator.cpp
Swiftb0y: Well that would blow up the scope because I would need to refactor the the API to use something other than `QList`. That is a long-term goal which I wanted to avoid for now because the container is usually very small (less then two dozen elements). I don't think this stopgap is an issue for now.

src/controllers/bulk/bulksupported.h
Swiftb0y: Well, which helper functions are you referring to exactly? `is_interesting` should belong to the enumerator IMO.

daschuer: Yes. But I have no special demands here.

Swiftb0y: I don't think moving that is an improvement really. So I'd like to keep as is if thats okay for you.

src/controllers/bulk/bulkenumerator.cpp
```diff
@@ -2,60 +2,73 @@
...
 #include "moc_bulkenumerator.cpp"
+#include "util/assert.h"
 
-BulkEnumerator::BulkEnumerator(UserSettingsPointer pConfig)
+BulkEnumerator::BulkEnumerator()
         : ControllerEnumerator(),
-          m_context(nullptr),
-          m_pConfig(pConfig) {
-    libusb_init(&m_context);
+          m_context(nullptr, &libusb_exit) {
```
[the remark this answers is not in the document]
daschuer: The compiler should be smart enough to dispose the function wrapper in the same way as the lambda. Isn't a lambda basically a function without a name but we want a name here? 

Only using the comment works as well if you have performance concerns.

src/controllers/bulk/bulksupported.h
daschuer: Ok

src/controllers/bulk/bulkenumerator.cpp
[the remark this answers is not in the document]
Swiftb0y: well not really performance concerns, more just using the right tool for the job. The solution with the plain function just uses a function pointer (and thus a form of late-dispatch, which we don't need here) while the lambda solution doesn't. [it also results in a size advantage](https://compiler-explorer.com/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:33,endLineNumber:10,positionColumn:33,positionLineNumber:10,selectionStartColumn:33,selectionStartLineNumber:10,startColumn:33,startLineNumber:10),source:'%23include+%3Cmemory%3E%0A%0Avoid+del(int*)%3B%0Aconst+auto+deleterLambda+%3D+%5B%5D(int*+i)+%7Bdel(i)%3B%7D%3B%0A%0Ausing+lambda_ptr+%3D+std::unique_ptr%3Cint,+decltype(deleterLambda)%3E%3B%0Ausing+dynamic_ptr1+%3D+std::unique_ptr%3Cint,+decltype(%26del)%3E%3B%0Ausing+dynamic_ptr2+%3D+std::unique_ptr%3Cint,+void(*)(int*)%3E%3B%0A%0Aint+lambda+%3D+sizeof(lambda_ptr)%3B%0Aint+dptr1+%3D+sizeof(dynamic_ptr1)%3B%0Aint+dptr2+%3D+sizeof(dynamic_ptr2)%3B%0A%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:g142,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+14.2+(Editor+%231)',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+14.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+14.2+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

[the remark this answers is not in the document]
daschuer: Interesting. The function pointer to the final libusb call must be somewhere though in the lambda case as well.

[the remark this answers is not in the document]
Swiftb0y: well not really. Its "just" a static call into another function. The linker fills in the address so its the same as any other function call. No indirect call needed in the lambda case (ignoring any implementation details related to cross-library calls).

acolombier:
I'll make that a draft PR for now, but feel free to mark it ready when you get the time to pick this work up!

Issue #11648, opened 2023-06-11
AutoDJ don't fallback on crossfade time when intro or outro is not set


Bacadam:
### Bug Description

When using AutoDJ in either Full Intro + Outro or Fade at Outro Start, I think the behaviour is not really what it should.

I quote the tooltip :

> Full Intro + Outro:
> Play the full intro and outro. Use the intro or outro length as the crossfade time, whichever is shorter. If no intro or outro are marked, use the selected crossfade time.

> Fade at Outro Start:
> Start crossfading at the outro start. If the outro is longer than the intro, cut off the end of the outro. Use the intro or outro length as the crossfade time, whichever is shorter. If no intro or outro are marked, use the selected crossfade time.

So in both cases, if intro OR outro is not set, AutoDJ should fallback on crossfade time.

But in reality, that's the case only if both outro AND intro are not set.
If intro is set, but not outro, then the intro length will be the crosfade time.
If outro is set, but not intro, then the outro length will be the crosfade time.

That's not good when you have a song with a long outro, like fade-out, and a transition to a song that starts immediately (no intro).
The crossfade time in this scenario is your long outro time, with a long period where both songs and their beats overlap.

I think crossfade setting should represent a maximum transition time when either intro or outro is missing.
See the diagrams below,

```
- is part of a track outside the outro/intro,
o is part of the outro
i is part of the intro
One of the above character represent 1 second

| marks the boundaries of the transition

These examples assume a 5 seconds crossfade setting

Full Intro + Outro, long outro, no intro:
------ooo|ooooo|
         |-----|------

Full Intro + Outro, short outro, no intro:
---------|ooo|
         |---|------

Full Intro + Outro, long intro, no outro:
---------|-----|
         |iiiii|iii---

Full Intro + Outro, short intro, no outro:
---------|---|
         |iii|------

Full Intro + Outro, no outro and no intro:
---------|-----|
         |-----|------

Full Intro + Outro, both outro and intro:
------oo|oooooooo|
        |iiiiiiii|------
         
Fade at Outro Start, long outro, no intro:
---------|ooooo|ooo
         |-----|------
         
Fade at Outro Start, short outro, no intro:
---------|ooo|
         |---|------

Fade at Outro Start, long intro, no outro:
---------|-----|
         |iiiii|iii---

Fade at Outro Start, short intro, no outro:
---------|---|
         |iii|--------
      
Fade at Outro Start, no outro and no intro:
---------|-----|
         |-----|------

[...]

PR #13628, opened 2024-09-04
Auto dj cross fader center

davidlmorris:
[An attempt to fix my understanding of git, by creating a new branch to fix code format 'issues'.]

_edit(ronso0)_ original PR is #13619

_edit ([davidlmorris](https://github.com/davidlmorris))_ This is now a new AutoDj transition option **'Skip Silence Start Center'**, that works exactly like the 'skip silence' transition and takes the same time, except that it now moves the cross-fader to the centre at the start of the transition, so that the next (the 'to' deck) track will start at full volume.

image

Note: this might lead to a sudden reduction in the volume of the playing deck if you have set the Crossfader Curve to Constant power.  See Preferences->Mixer and make sure 'Slow fade/Fast cut (additive)' is set (And move the fader to the position somewhere close to where the 'Logarithmic chart' looks like a triangle).

image

### DIFF
--- a/src/library/autodj/autodjprocessor.cpp
+++ b/src/library/autodj/autodjprocessor.cpp
@@ -120,15 +120,17 @@ AutoDJProcessor::AutoDJProcessor(
           m_eState(ADJ_DISABLED),
           m_transitionProgress(0.0),
           m_transitionTime(kTransitionPreferenceDefault) {
-    m_pAutoDJTableModel = new PlaylistTableModel(this, pTrackCollectionManager,
-                                                 "mixxx.db.model.autodj");
+    m_pAutoDJTableModel = new PlaylistTableModel(
+            this, pTrackCollectionManager, "mixxx.db.model.autodj");
     m_pAutoDJTableModel->selectPlaylist(iAutoDJPlaylistId);
     m_pAutoDJTableModel->select();
 
     m_pShufflePlaylist = new ControlPushButton(
             ConfigKey("[AutoDJ]", "shuffle_playlist"));
-    connect(m_pShufflePlaylist, &ControlPushButton::valueChanged,
-            this, &AutoDJProcessor::controlShuffle);
+    connect(m_pShufflePlaylist,
+            &ControlPushButton::valueChanged,
+            this,
+            &AutoDJProcessor::controlShuffle);
 
     m_pSkipNext = new ControlPushButton(
             ConfigKey("[AutoDJ]", "skip_next"));
@@ -169,6 +171,7 @@ AutoDJProcessor::AutoDJProcessor(
 
     m_pCOCrossfader = new ControlProxy("[Master]", "crossfader");
     m_pCOCrossfaderReverse = new ControlProxy("[Mixer Profile]", "xFaderReverse");
+    m_crossfaderStartCenter = false;
 
     QString str_autoDjTransition = m_pConfig->getValueString(
             ConfigKey(kConfigKey, kTransitionPreferenceName));
@@ -808,12 +811,14 @@ void AutoDJProcessor::playerPositionChanged(DeckAttributes* pAttributes,
                     otherDeck->setPlayPosition(otherDeck->startPos);
                 }
 
-                if (!otherDeckPlaying) {
-                    otherDeck->play();
+                if (m_crossfaderStartCenter) {
+                    setCrossfader(0.0);
+                } else if (thisDeck->fadeBeginPos >= thisDeck->fadeEndPos) {
+                    setCrossfader(thisDeck->isLeft() ? 1.0 : -1.0);
                 }
 
-                if (thisDeck->fadeBeginPos >= thisDeck->fadeEndPos) {
-                    setCrossfader(thisDeck->isLeft() ? 1.0 : -1.0);
+                if (!otherDeckPlaying) {
+                    otherDeck->play();
                 }
 
                 // Now that we have started the other deck playing, remove the track
@@ -832,6 +837,7 @@ void AutoDJProcessor::playerPositionChanged(DeckAttributes* pAttributes,
         double crossfaderTarget;
         if (m_eState == ADJ_LEFT_FADING) {
             crossfaderTarget = 1.0;
+
         } else if (m_eState == ADJ_RIGHT_FADING) {
             crossfaderTarget = -1.0;
         } else {
@@ -1342,6 +1348,7 @@ void AutoDJProcessor::calculateTransition(DeckAttributes* pFromDeck,
                  << "outroLength" << outroLength;
     }
 
+    m_crossfaderStartCenter = false;
     switch (m_transitionMode) {
     case TransitionMode::FullIntroOutro: {
         // Use the outro or intro length for the transition time, whichever is
@@ -1441,6 +1448,10 @@ void AutoDJProcessor::calculateTransition(DeckAttributes* pFromDeck,
             useFixedFadeTime(pFromDeck, pToDeck, fromDeckPosition, outroEnd, toDeckStartSeconds);
         }
     } break;
+    case TransitionMode::FixedStartCenterSkipSilence:
+        m_crossfaderStartCenter = true;
+        // fall through intended!
+        [[fallthrough]];
     case TransitionMode::FixedSkipSilence: {
         double toDeckStartSecond;
         pToDeck->fadeBeginPos = getLastSoundSecond(pToDeck);
--- a/src/library/autodj/autodjprocessor.h
+++ b/src/library/autodj/autodjprocessor.h
@@ -163,7 +163,8 @@ class AutoDJProcessor : public QObject {
         FullIntroOutro,
         FadeAtOutroStart,
         FixedFullTrack,
-        FixedSkipSilence
+        FixedSkipSilence,
+        FixedStartCenterSkipSilence
     };
 
     AutoDJProcessor(QObject* pParent,
@@ -285,6 +286,7 @@ class AutoDJProcessor : public QObject {
     double m_transitionProgress;
     double m_transitionTime; // the desired value set by the user
     TransitionMode m_transitionMode;
+    bool m_crossfaderStartCenter;
 
     QList<DeckAttributes*> m_decks;
 
--- a/src/library/autodj/dlgautodj.cpp
+++ b/src/library/autodj/dlgautodj.cpp
@@ -146,7 +146,11 @@ DlgAutoDJ::DlgAutoDJ(WLibrary* parent,
             "Skip Silence:\n"
             "Play the whole track except for silence at the beginning and end.\n"
             "Begin crossfading from the selected number of seconds before the\n"
-            "last sound.");
+            "last sound.\n"
+            "\n"
+            "Skip Silence:\n"
+            "The same as Skip Silence, but starting transitions with a centered\n"
+            "crossfader, so that the intro starts at full volume.\n");
 
     pushButtonFadeNow->setToolTip(fadeBtnTooltip);
     pushButtonSkipNext->setToolTip(skipBtnTooltip);
--- a/src/library/autodj/dlgautodj.cpp
+++ b/src/library/autodj/dlgautodj.cpp
@@ -179,6 +179,8 @@ DlgAutoDJ::DlgAutoDJ(WLibrary* parent,
             static_cast<int>(AutoDJProcessor::TransitionMode::FixedFullTrack));
     fadeModeCombobox->addItem(tr("Skip Silence"),
             static_cast<int>(AutoDJProcessor::TransitionMode::FixedSkipSilence));
+    fadeModeCombobox->addItem(tr("Skip Silence Start Full Volume"),
+            static_cast<int>(AutoDJProcessor::TransitionMode::FixedStartCenterSkipSilence));
     fadeModeCombobox->setCurrentIndex(
             fadeModeCombobox->findData(static_cast<int>(m_pAutoDJProcessor->getTransitionMode())));
     connect(fadeModeCombobox,


daschuer:
pre-commit is complaining. 
I assume you have installed it locally. 

You may test it by adding a space at the end of the affected lines that are shown following "Details" above. 

If everything works you need to commit twice. 
`git commit -a -m"Fix code style"`
The first commit is rejected by pre-commit on purpose and changes are automatically applied in the second commit.

Swiftb0y:
@davidlmorris took the liberty to fix it for you. make sure you `git pull` before making new changes so your local branch is the same as the upstream one I changed.

ronso0:
... and please "sync" this with #13619 (updates after reviews).

ronso0:
I wonder if it's worth adding this option to the controls in the GUI.
(could be tricky to find an icon for this option)

I mean: is this something you'd want to toggle quickly while mixxxing? without having to go to the preferences

davidlmorris:
> I wonder if it's worth adding this option to the controls in the GUI.

I would say no.  It is what I originally set out to do, but basically, this setting could/would apply to all AutoDj options, so we would end up with 8 rather than 4, which would make that GUI setting too busy and too confusing.

I have an idea though for my 'next change' which would require a much bigger effort to dig through and understand the code:  Add more 'unseen' cue points, being the first and last (and here I'm just guessing - this needs to be thought out) say: -6db (from ref), - 10db, -20db and maybe others.  Then have an AutoDJ option that works like the 'skip silence' option, except that it picks up the timing from, for example, the -6db point (selectable in options), called 'Selectable Fade' or 'Radio Fade' so that you can get much better crossfades (or radio fades) on songs where you don't have to add start and end clue markers.   This way a fade-out would be merged better with a cold start.  So a cold (abrupt) end would transition perfectly (and tightly) into a cold opening.

Naturally which cue marker you used would be selectable in AuduDJ preferences.  (I am assuming that adding cue marks would not be a huge pain or overhead- given they would need to be checked and loaded with each song - and I presume updated in the DB ... lots here I don't know about that I (or someone) would need to check).

davidlmorris:
> took the liberty to fix it for you. make sure you git pull before making new changes so your local branch is the same as the upstream one I changed.

@Swiftb0y Thankyou!  Brilliant!  (I have done the pull.  But until I am confident I won't break things further, will abstain from any other git shinanigans.)

ronso0:
> this setting could/would apply to all AutoDj options, so we would end up with 8 rather than 4, which would make that GUI setting too busy and too confusing.

What I meant was: have this option in the Auto DJ controls row, too. (this does not require duplicating all transistion modes)
It could be off for tracks where outro and intro blend nicely, and on if there's a track that starts with a vocal that you want with full volume.

daschuer:
I am in doubt that this option suits for every transition mode. For instance if you have defined an into of a track you may not want to start it immediatly at full volume. In the other hands there are tracks that start with a significant sound like a signature vocal. that shall not be fadet in, because it is silenced at the beginning. In addition I want not fade in Track started at 0:00, because the usually have their own "intro". Sometimes I fade the previous track out to make the signature sound of the new track even more notable. So It is a case by case decision. 

So we may define these rules: 
* Full Intro + Outro:
  * If intro not defined or zero start new track at full volume. 
* Fade At Outro Start:
  * If  intro not defined or zero start new Track at full volume. 
* Full Track:
   * Here to conditional option may make sense "Full Track, no Fade-In" 
   * Since this is an "unprepared" mode any intro dependency doe not make sense
* Skip Silence
   * Same here. Maybe we can rethink both. Does playing a track with silence ever make sense?
   * Maybe we need a gepless mode instead that just plays the tracks like pressed on the vinyl or live recordings. 
  
What is your use case? How we can distinguish the intro types?

ronso0:
Thanks, that's excactly why I proposed to have the option in the skins, not only in the preferences.

> * Full Intro + Outro:
>   * If intro not defined or zero start new track at full volume. 
> * Fade At Outro Start:
>   * If  intro not defined or zero start new Track at full volume. 

I don't think we should assume the track has no intro/outro just because users didn't configure it.
IIUC the center mode is most helpful for Full Track and Skip Silence. I imagine there are users who don't use intro/outro but run Auto DJ with these modes in "manual" fade, ie. trigger Fade Now when it seems appropriate.

daschuer:
Yes, I can confirm that. On the other hand I like to be able to set the behavior track by track, like we do with the intros. 

My rational was that: If a track has an option intro, it is assumed to faded in. We have no other option yet. 
The track is never faded after intro end. 
We may reuse this assumption. 

In addition I think the default should be no fade in. 
This fits better to most of my tracks, because they have a natural intro, or start with lyrics right away.

davidlmorris:
From a radio perspective, there is almost NO 'use case' for when you would want to fade a track in.  (Ok there is one: when you fade an instrumental in from the middle so you can time a cold ending to the top of the hour where you have time pips like BBC World Service or a syndicated news bulletin... but this isn't an AutoDj thing.)   I've never worked in or seen a radio studio with a 'crossfader'.  It just isn't (wasn't?) a thing.  Every announcer I've seen starts the track at (or near full volume) talking over the intro, or running a sweeper, or doing nothing as required.  On some public and community radio stations I have heard the occasional fade-in, but it sounds unprofessional and messy and was probably a mistake.  Perhaps this is different in other parts of the world?

So from my perspective in my use case, I would have the option checked all the time.  I suspect this would be true for most people using it this way.   So if a song has an intro specified (and the previous has an outro or not) I would still want to start it at full volume.  (In my case when I use the intro/outro I use it as an indicator for when it is safe to voice over music... which still mostly works for AutoDJ)   To be clear this is not a full solution to the problem of having the ability to do a 'full radio mix' (more on that later), but at the risk of a pun, it does take us halfway there.

I have enormous respect for those that are using Mixxx (and the rest) as musical instruments, manually matching beats, and fading bits and pieces in and out to create something new.  That is a skill set I just don't have.  And of course, the requirements there are very different.   In those cases, I can't imagine AutoDJ is used at all, and a cross-fader might be used a lot.  But perhaps Electro and House music lends itself to a fade-in and so for those having that as an option would be useful.  Not my area of expertise (at all!).

I have no objection to it being part of the UI bar for AutoDJ, but right now, my suggestion would be to include it as the option is proposed.  And then as a separate pull request if someone needs it have it as an option of the AutoDJ UI 'bar'.   

I would be keen to have the feature included as soon as possible, rather than stalled out trying to make it perfect.  After all, there is simply no option at the moment that allows a track to start at full volume in AutoDj.

davidlmorris:
I might add, that I use the 'Skip Silence' with the cross-over time set to 5 seconds currently.  This works with the center option as well as you would expect (and a whole lot better than a fade-in).  Even using 'Full Track' and adding silence would work better for me since you are hearing the full track as the creator intended (like an LP or CD) without a fade-in.

daschuer:
> ... there is almost NO 'use case' for when you would want to fade a track in ...

I can confirm. 

> rather than stalled out trying to make it perfect. 

We are in a good progress to make it perfect. I like rather include only your use case as a special transition than increase the complexity of the Auto DJ by making it a separate option.  

So it looks like we already have a conclusion ..  

Skip Silence is the same as the two intro/outro modes in case no intros and outros are defined. Do you have some defined? 

My guess is if a track has defined an outro, you want to use it, right? If not, you like the 5 second fade out as described. 

This is "Fade At Outro Start". Which has currently the IMHO buggy behaviour that it fades a track into lyrics or such if you have a long outro and no intro defined. 

So I think this can be fixed by your mode. In case the is no intro, start with full volume at sound start. 

I don't think there is any use case for an uncontrolled long intro in that case. 
This is probably a minimal solution without any config option. 

Does it work for you?

davidlmorris:
> Skip Silence is the same as the two intro/outro modes in case no intros and outros are defined. Do you have some defined?

I do.  While most work as advertised (as you stated), occasionally I marked some from the last vocal with a very extended outro, mostly so I have a mark for starting an extended talk break over the tail.  Unfortunately, this sometimes makes the crossfade seem a bit messy, making the song seem like it has been cut short - which it has, so I just leave it on skip silence.  

I have tested the change against all or the options, and they all seem to work.  So to be clear this works for every type of AutoDj transition currently, and I would be against limiting it to any particular one.

So I would be keen to have the change (as it is) included as soon as possible.

The only reason that I would not propose making it a default, is that I don't think that changes that 'break expectations' or 'break behavior' should be introduced as a default, even if I suspect most people will end up using it that way.

daschuer:
The issue I have with this extra toggle, that this is clearly a "per track" option. You can either have club tracks with an intro you like to fade in and a track that starts with vocals that should be started with full volume. 

I want a feature that allows unattended mixes as well.
Even the two intro/outro modes are not good in this regards and only a compromise, because we could not decide for one or another approach.

I am not afraid to break a feature in favor of it. We will learn if this was good or bad during the beta period and may change the implementation.

So how about doing both, add another skip silence option with only fade out and alter the "Fade At Outro Start" feature as described?

davidlmorris:
I thought AutDJ was really only for unattended mixes (although I also use it as a 'current' playlist), so a per-track option is not a 'live' option unless you have gone through each song on the playlist and set this up, right?  Or have I misunderstood what you are suggesting?

This doesn't mean I would be against another separate option.  My solution is just a quick fix for the problem of tracks not starting at full volume when using AutoDj, regardless of what setting you are using.  One I would be very keen to see implemented ASAP.

daschuer:
Auto DJ shall be usable for complet unattanded mixing but also allow seamless interaction with a real DJ. 

> So a per-track option is not a 'live' option unless you have gone through each song on the playlist and set this up, right?

Yes, right. For many of my track I have only set the cue point, used as load cue. Some have also a difined intro and outro.
Most club mixes have a reasonable long intro that it works by default using the fade time. Somtimes I set an intro to protect the track from a transition into vocals or skipping unusable embient noise. That it works, but than there are Tracks like "Draft Punk - Technologic" which really shall be used without any fade in. Without an intro set, the probably long outro is used. With short intro the bevious track is cut off unleasant. I need a way to fix it in an unattenden mix.

> My solution is just a quick fix

The issue with the extra option for all transition is that it is more than a quick fix for your particular use case, but it still does not help for my use case. It may limit use for future improvments because some useres will get used to this option and we may hasitate to remove it.

daschuer:
Proposal: 
Only do one step: Add a new "Skip Silence, fade out" transition to fix you use case. 

Later or in addition, we may also add the workaround for my use case.

davidlmorris:
> It may limit use for future improvments because some useres will get used to this option and we may hasitate to remove it.

You say that like it is a bad thing?

> Proposal:
> Only do one step: Add a new "Skip Silence, fade out" transition to fix you use case.
> 

Why limit this?  It is after all a setting in preferences, that doesn't affect extant behavior at all.  If you want less, then don't set the check box.  If you want more... then create a new feature that can be implemented later, right?  What problem are we trying to prevent here?

daschuer:
> You say that like it is a bad thing?

Yes, I mean it actually: https://www.youtube.com/watch?v=glZ1C-Yu5tw&t=206s

> Why limit this?

because it breaks the concept of full auto, for the intro/outro related modes.

davidlmorris:
> because it breaks the concept of full auto, for the intro/outro related modes.

How?  Just don't set the checkbox (like it is now).  

Are we trying to do too many things here?

davidlmorris:
> Yes, I mean it actually: https://www.youtube.com/watch?v=glZ1C-Yu5tw&t=206s

I watched this.  I did find him a tad hyperbolic.  It is an argument of usability and functionality over the complexity of development.  These things are, of course, a trade-off.   In my view, not an argument for an incomplete user experience.

However, this isn't my software - it is yours or your team's.  If you think this is a bad proposal, then I understand that you won't include it, even if I don't understand the reasoning.  Do let me know, since I would like to adapt it for my purposes (since I do think Mixxx is great software), and I would like to fork a more stable version than the nightly build.

daschuer:
Oh sorry, it was not my intention to reject your work. It was probably also a bad idea to post the video, which shows the issue exaggerated. I highlighted appreciate your work and the fader trick you demonstrate here is really a missing feature.

daschuer:
Let's some up:
* You need a new mode like "skip silence" without fade in. And you will use it permanently. 
* I need no fade in options for certain tracks in "Fade At Outro Start" 

I am happy to merge your demand. 
Adding this center mode to all transitions would create features where we have no demand for. It is always difficult to test and discuss edge cases for a "feature" that is not used anyway and only a by-catch. 

That's why I have proposed to go in small steps and only solve your problem, which is for my understanding covered with ""Skip Silence, fade out".

Once this is merged, we can build in that in any direction.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -859,17 +864,31 @@ void AutoDJProcessor::playerPositionChanged(DeckAttributes* pAttributes,
             double transitionProgress = (thisPlayPosition - thisDeck->fadeBeginPos) /
                     (thisDeck->fadeEndPos - thisDeck->fadeBeginPos);
             double transitionStep = transitionProgress - m_transitionProgress;
+
+            if ((transitionProgress == 0.0) &&
```
daschuer, line 868: I am not sure if the == 0.0 state is always reached, because of rounding issues.i think it is more reliable to move this in an else branch of the condition below.

davidlmorris:
Thanks.  (I was trying to reduce complexity, while not messing with other user's expectations.)

For the most part  "Skip Silence, fade out" is at least good enough for now (and certainly better than nothing!).  I can see situations where I might not be skipping silence (dinner party background et al) or using the into and outro, but "Skip Silence" is certainly my most pressing requirement.

I have a further more detailed idea which I hinted at above...  which would require a much bigger effort to dig through and understand the code, the biggest part probably is to add more 'unseen' cue points, being the first and last -6db (from ref), - 10db, -20db and maybe others (I would need to play around and hear the results).

What would you see as the biggest problem with adding hidden cue points like the existing 60db point?  (e.g. Database reformat?)

src/library/autodj/autodjprocessor.cpp
```diff
@@ -859,17 +864,31 @@ void AutoDJProcessor::playerPositionChanged(DeckAttributes* pAttributes,
...
                 // Backward seeks pause the transitions; forward seeks speed up
                 // the transitions. If there has been a seek beyond endPos, end
                 // the transition immediately.
-                double remainingCrossfader = crossfaderTarget - currentCrossfader;
-                double adjustment = remainingCrossfader /
-                        (1.0 - m_transitionProgress) * transitionStep;
-                // we move the crossfader linearly with
-                // movements in this track's play position.
-                setCrossfader(currentCrossfader + adjustment);
+                if ((m_crossfaderStartCenter) &&
```
daschuer, line 877: Auto DJ should allow to adjust the Crossfader during a transition. So the code must be able to continue fading from any position. In this case I can imagine a emergency condition where the user realizes that he wants fade in and moves the xfader rapidly to the beginning.

davidlmorris:
One other thing worth mentioning is the potential unintended consequence for someone who has set (under preferences Mixer) the Crossfader Curve to Constant power.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -859,17 +864,31 @@ void AutoDJProcessor::playerPositionChanged(DeckAttributes* pAttributes,
...
+                        (transitionProgress <=
+                                ((thisDeck->fadeEndPos -
+                                         thisDeck->fadeBeginPos) /
+                                        2))) {
+                    // We want the cross fader to remain in the middle.
```
daschuer, line 882: This will constantly fight with the user. I think it is better to just not touch the crossfadere here. It will feel naturally if we fall back to the default mode when xfader is no longer in the expected center position.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -1343,6 +1362,7 @@ void AutoDJProcessor::calculateTransition(DeckAttributes* pFromDeck,
     }
 
     switch (m_transitionMode) {
+        // case TransitionMode::RadioFullIntroOutro:
```
daschuer, line 1365: What is that?

daschuer (changes requested):
The new mode does not work well with const power mode, which is also usable for radio. Because it avoids loudness peaks during the transition. In this case we have an immediately loudness drop in the outgoing track. Original this double loudness issue was only short, during crossing the center. Now we have it during the full first half. A real DJ will use EQ or filter to avoid blowing the speakers. 

What could be the approach here? 

Fade out the outgoing track and delay the start of the incoming track? Ideas?

daschuer:
> I can see situations where I might not be skipping silence (dinner party background et al)

Crossfading in full track mode is probably another by catch without a use case. We kept this mode, because it was the default before introducing silence detection. 
I am happy to alter it to the real use cases like gapless life recording / symphony mode. 

Do you need the fade out part at dinner? 

> Or using the into and outro ...

Can you confirm my demand in that case?

daschuer:
There is no issue to add more hidden cue points as long at they are auto detected.

src/library/autodj/autodjprocessor.cpp
davidlmorris: As I said elsewhere that is a comment form an earlier experiment, that I neglected to delete.

src/library/autodj/autodjprocessor.cpp
davidlmorris: This should now be:

`                 transitionProgress <= std::midpoint(thisDeck->fadeBeginPos, thisDeck->fadeEndPos) {`

to make it clearer what is going on.  

In my testing, neither constantly fights the user.  If the user moves the fader it goes there and completes the transition.

davidlmorris:
My changes left the existing behavior completely intact unless you had the option check box ticked.  If you move the fader after it starts the crossfade it just goes to where you put it as you would expect (like when the option box is unchecked).

> Fade out the outgoing track and delay the start of the incoming track? Ideas?

My idea originally was the option check box, as a follow-up to another [change](https://github.com/mixxxdj/mixxx/pull/13332) to use the same message as a reminder of the side effect.  See https://github.com/mixxxdj/mixxx/pull/13619 for an illustration.

I think you are making significant changes to what I proposed.   I suspect that some of these are breaking changes, and since you know the software better than me, then I think you ought to be making them rather than me. (That and my lack of confidence in my understanding of Git).   I am still unclear if you are adding another option to the drop-down menu, or changing an existing one.

And, I can't see what you are doing to test it.   Have you tested my changes?

> There is no issue to add more hidden cue points as long at they are auto detected.

That is brilliant news.  These would and should be auto-detected and hidden just like the -60db silence marker.  But that is for another time in the future (I suspect a distant future, now).

daschuer:
> My changes left the existing behavior completely intact unless you had the option check box ticked. If you move the fader after it starts the crossfade it just goes to where you put it as you would expect (like when the option box is unchecked).

I can confirm this. In a new test I just did. 

> I think you are making significant changes to what I proposed.

My demand for this PR is to turn you option in to a new transition mode as discussed. Any other changes can be done in follow up PRs. 

> I think you ought to be making them rather than me. 

I have currently other priorities, like releasing 2.4.2. Sorry. 

> And, I can't see what you are doing to test it. Have you tested my changes?

I have tested your original PR and did a review here without testing. I have probably misread your code.

src/library/autodj/autodjprocessor.cpp
daschuer: Confirmed, this condition is never true and can be removed.

src/library/autodj/autodjprocessor.cpp
daschuer: I can confirm that this readjust the user tweaks in some conditions. It happens because you compare here fractions of the full track length with fractions of the transition. I think the best solution is to move this code to Line 823 as 
```
                if (m_crossfaderStartCenter) { 
                    setCrossfader(0.0);
                } else if (thisDeck->fadeBeginPos >= thisDeck->fadeEndPos) {
                    setCrossfader(thisDeck->isLeft() ? 1.0 : -1.0);
                }
```

davidlmorris: Please show the conditions.  I can't find any in my testing.

src/library/autodj/autodjprocessor.cpp
davidlmorris: This only happens once and the crossfader is now at the new position, so the user can adjust.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -566,6 +568,8 @@ AutoDJProcessor::AutoDJError AutoDJProcessor::toggleAutoDJ(bool enable) {
                 setCrossfader(1.0);
             }
         }
+        m_crossfaderStartCenter = m_pConfig->getValue<bool>(ConfigKey(kConfigKey,
```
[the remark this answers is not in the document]
davidlmorris: As far as I can tell AutoDj is instantiated just once at the program start, so I've made the change even if I think is a tad redundant for completeness.  "Currently we need to restart auto DJ to switch the modes.", is true, like a lot of changes made in options.

src/library/autodj/autodjprocessor.cpp
davidlmorris: After some testing (and an assert), I couldn't find this state being reached either.  So I agree it can be removed.

davidlmorris:
While I have made the changes I mentioned above to my code.  I won't be making any further commits until I understand git better and have understood the implications of other changes, some of which as I said before I am uncomfortable with.  

(I've been playing with it this morning, and it seems to make even cold cuts sound 'ok-ish' [as in so, so much better] with a 4-second crossfade time - so I'll modify version 2.4 for my use in the meantime, and maybe 2.5 when it comes out).

src/library/autodj/autodjprocessor.cpp
davidlmorris: I've had a play around with this.  And I think you are correct.  It seems you only need to set the position of the crossfader to the middle once.  And the best position would be just before the other desk starts playing.  So I would propose this at around line 820:

```
                if (m_crossfaderStartCenter) {
                    setCrossfader(0.0);
                } else if (thisDeck->fadeBeginPos >= thisDeck->fadeEndPos) {
                        setCrossfader(thisDeck->isLeft() ? 1.0 : -1.0);
                }

                if (!otherDeckPlaying) {
                    otherDeck->play();
                }
``` 
The reason is that I noticed that occasionally the audio seems to just miss the very start of a track (I noticed it on a cold vocal start).  This may have been a glitch in my system, so I cannot attest to this, but it seems the best thing to do in any event.

davidlmorris:
Ok, I pushed this to my repository... forgetting that it was connected to this pull...  hopefully, this is not an 'ooops' situation again.  These are changes made or suggested above.

daschuer:
> My demand for this PR is to turn you option in to a new transition mode as discussed. Any other changes can be done in follow up PRs.

Can you also re-introduce the the new transition mode rather than the preferences option? Than this is immediately merge-able IMHO.

davidlmorris:
As "demanded":
Reintroduction of a new transition mode: called "Skip Silence Start Full Volume".
Removal of preferences option.

I've left in the global m_crossfaderStartCenter and supporting code in `autodjprocessor.cpp`.  This can be used to extend other transmission modes, when or if required.

**Not included**:  is a much much **better transition**, which (I assume) I should do as a separate pull request, that creates some hidden cue points, which correspond to the first and last sound in the track at specific volumes, and then uses those to start the fade out, up to a maximum set by the user in the spin box.  This means that cold endings and cold starts can mesh perfectly, and fade-outs happen sensibly.  Other than a few gremlins which I "think" I have sorted out - but will test more, I think it now does a better job "mixing tracks" than I do especially on cold endings!

davidlmorris:
Build Ubuntu 22.04

trips up when I let a case fall through (deliberately) on a switch statement.

```
/home/runner/work/mixxx/mixxx/src/library/autodj/autodjprocessor.cpp: In member function ‘void AutoDJProcessor::calculateTransition(DeckAttributes*, DeckAttributes*, bool)’:
Error: /home/runner/work/mixxx/mixxx/src/library/autodj/autodjprocessor.cpp:1454:33: error: this statement may fall through [-Werror=implicit-fallthrough=]
 1454 |         m_crossfaderStartCenter = true;
      |         ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
/home/runner/work/mixxx/mixxx/src/library/autodj/autodjprocessor.cpp:1455:5: note: here
 1455 |     case TransitionMode::FixedSkipSilence: {
      |     ^~~~
cc1plus: all warnings being treated as errors
```

Is there some warning suppression I can/should apply here?

davidlmorris:
I've added  this (C++17 option) :
```
    case TransitionMode::FixedStartCenterSkipSilence:
        m_crossfaderStartCenter = true;
        // fall through intended!
        [[fallthrough]];
    case TransitionMode::FixedSkipSilence: {
        double toDeckStartSecond;
```
to see if it will pass.

daschuer:
Thank you. 
The code looks good. Are you able to squash the commits to one to remove the forth and back commits? 
Like (on a clean branch):
`git reset --soft abcd88e20a85591b2e74500eb5768a3461892dae && git commit --amend --no-edit`
Else I can take care during manual testing.

davidlmorris:
> The code looks good. Are you able to squash the commits to one to remove the forth and back commits?
Like (on a clean branch):
git reset --soft abcd88e20a85591b2e74500eb5768a3461892dae && git commit --amend --no-edit

I can enter that command, but I confess this will be 'cargo cult' scripting at its worst since I don't know git enough to really understand what you are asking, and won't know what to do if it breaks. (I've been using the GitHub Windows app.)  For that reason, I suggest you do it.

davidlmorris:
[Update auto dj fade mode tooltip](https://github.com/mixxxdj/mixxx/pull/13628/commits/55a893a16eaedb8a9535144f5e9f143f34af4d1f).  Nice catch.  I managed to miss that one.

daschuer (approved):
All green now. Thank you for this nice new mode.

davidlmorris:
> All green now. Thank you for this nice new mode.

My pleasure.  And, thank you.

I'll create another branch (soon... a few things on the go) and upload the Pull Request for the 'feature request' AutoDj: New transmission mode 'Radio Laneway' [#13716]( https://github.com/mixxxdj/mixxx/issues/13716)

daschuer:
I like this new mode a lot. It sounds more "natural".
However it is not realy usable for me, because I use: 
"Fade At Outro Start" 
Does you ongoing work also considering a solution for that mode?  Shall I file an issue for it? 
Currently my tracks have mostly no intro marked. I use the main cue for a good start point (first downbeat) and intro end, to protect vocals from being faded in.

davidlmorris:
I hadn't especially considered it, but it would be trivial to implement.  I can do this if you like.  I'll basically just add (yet another) transition mode, called "Fade At Outro Start, Start Center", using the same code (effectively).  I'll probably do that as a separate pull (for practice), and then do the 'Radio laneway crossfade" one as another pull.

Feel free to 'assign' me to those jobs if you file an issue.

daschuer:
This https://github.com/mixxxdj/mixxx/issues/11648
contains also some interesting use cases.

daschuer:
Ok I will file a new issue. Yet another transition mode does not work for me. It shall auto detect from the intro and intros if starting at center is desired or not.

davidlmorris:
cf [Yet another](https://en.wikipedia.org/wiki/Yet_another) for my oblique humour.

From https://github.com/mixxxdj/mixxx/issues/11648 and "It shall auto detect from the intro and intros if starting at center is desired or not", sounds like a non-trivial change from "Fade At Outro Star" as it stands (I was expecting to just add another case to the switch statement with one line).   I think this might be best left to after you get a chance to see https://github.com/mixxxdj/mixxx/issues/13716 in action since that (just) might solve your issue.

daschuer:
Here is the issue: https://github.com/mixxxdj/mixxx/issues/13742

Looking forward to see #13716. In general I consider all fully atonome modes as good defaults, but the user should be able to anotate tracks to adjust it for edge cases where the auto detection fails.

davidlmorris:
In the tool tips can we change 
line 151 `"Skip Silence:\n"`
 to
line 151 `"Skip Silence Start Full Volumne:\n"`
to make it clear which transition is being referred to?

daschuer:
Fixed in https://github.com/mixxxdj/mixxx/commit/e2abcf9fae6aa59faf7cabf622314d8494bc0ca6

davidlmorris:
> Looking forward to see https://github.com/mixxxdj/mixxx/issues/13716. 

Have you seen https://github.com/mixxxdj/mixxx/pull/13750 yet (the PR for 13716)?

daschuer:
@davidlmorris We need your permission to ship your changes with Mixxxx. Can you pleas sign https://docs.google.com/a/mixxx.org/spreadsheet/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ
and comment here when done? Thank you very much.

davidlmorris:
I filled this out this morning.  And you should have had it an hour or two back.

PR #13631, opened 2024-09-06
(fix) Multi-track 'Track Properties': accept entered text directly (clearing not required)

### LINKED ISSUE #13915: modifying multiple tracks comment tags only works by selecting an existing comment from the list
### Bug Description

I would like to pre-set the comments of a group of tracks.

1. I select all the tracks
2. I open the properties editor
3. I write a new comment in the box
4. I click Apply or OK

what happens: nothing changes

But, if I edit one track to have the comment I want I can do:

1. I select all the tracks
2. I open the properties editor
3. Select the comment I want from the dropdown list
4. I click Apply or OK

then all the tracks get that comment.

### Version

HEAD

### OS

_No response_

ronso0:
Something I overlooked earlier, it seems:
previously we had to explicitly clear a tag before a manually entered text would be accepted as new text.

**Background:**
Editors with multiple values have a placeholder text `<various>`. It is removed when another item is selected (Clear or text item).
On Apply this placeholder is checked to detect whether a multi-value field has been edited.

**Fix:**
Now the placeholder is also cleared when the `editingFinished()` signal is fired (checks for valid text on Return and focus out).
For the comment editor this signal doesn't exist. There is `textChanged()` but this is fired for each change (character add/remove), which is not ideal. Instead we catch the `focusOut` event of the comment editor and check whether the current text is 'new'.

**Small fix:**
Don't use QString::trimmed() to be consistent with the track table's inline editor.
Only trailing whitespaces are removed, which allows to clear tags by entering  a single space.

### DIFF
--- a/src/library/dlgtrackinfomulti.cpp
+++ b/src/library/dlgtrackinfomulti.cpp
@@ -28,6 +28,7 @@ namespace {
 const QString kVariousText = QChar('<') + QObject::tr("various") + QChar('>');
 const char* kOrigValProp = "origVal";
 const QString kClearItem = QStringLiteral("clearItem");
+const QString kClearIconPath = QStringLiteral(":images/library/ic_library_cross_grey.svg");
 
 void setItalic(QWidget* pEditor, bool italic) {
     auto font = pEditor->font();
@@ -47,21 +48,27 @@ void setBold(QWidget* pEditor, bool bold) {
     pEditor->setFont(font);
 }
 
-/// Check if the text has been edited, i.e. is not <various>
+/// Check if the text has been edited
 QString validEditText(QComboBox* pBox) {
-    QString origVal = pBox->property(kOrigValProp).toString();
-    if (pBox->currentIndex() == -1 &&
-            (pBox->lineEdit()->text() == origVal ||
-                    pBox->lineEdit()->placeholderText() == kVariousText)) {
-        // This is either a single-value box and the value changed, or this is a
-        // multi-value box and the placeholder text was removed when clearing it.
+    // For multi-value boxes we check if there is a placeholder (our 'modified'
+    // flag). For single-value boxes we compare with the original value.
+    auto* pLine = pBox->lineEdit();
+    const QString origVal = pBox->property(kOrigValProp).toString();
+    QString currVal = pLine->text();
+    if ((pBox->count() > 0 && !pLine->placeholderText().isNull()) ||
+            (pBox->count() == 0 && currVal == origVal)) {
         return QString();
     }
-    // We have a new text
-    return pBox->currentText().trimmed();
+    // We have a new text.
+    // Remove trailing whitespaces. Keep Leading whitespaces to be consistent
+    // with the track table's inline editor.
+    while (currVal.endsWith(' ')) {
+        currVal.chop(1);
+    }
+    return currVal;
 }
 
-/// Sets the text of a QLabel, either the only value or the 'various' string.
+/// Sets the text of a QLabel, either the only/common value or the 'various' string.
 /// In case of `various`, the text is also set italic.
 /// This is used for bitrate, sample rate and file directories.
 /// Optionally toggle bold (bitrate and sample rate).
@@ -164,9 +171,9 @@ void DlgTrackInfoMulti::init() {
     valueComboBoxes.append(txtGrouping);
 
     for (QComboBox* pBox : valueComboBoxes) {
-        // This will be displayed if there are multiple values
+        // This sets/enables the QLineEdit.
         pBox->setEditable(true);
-        // We allow editing the value but we don't want to add each edit to the item list
+        // We allow editing but we don't want to add each edit to the item list.
         pBox->setInsertPolicy(QComboBox::NoInsert);
         // Avoid showing scrollbars if not needed. The dialog has at least 17
         // QLineEdit/QLabel/QPushButton rows + layout spacing + title bar, so
@@ -178,6 +185,15 @@ void DlgTrackInfoMulti::init() {
                 QOverload<int>::of(&QComboBox::currentIndexChanged),
                 this,
                 &DlgTrackInfoMulti::slotTagBoxIndexChanged);
+
+        auto* pLine = pBox->lineEdit();
+        // editingFinished() is also emitted when the list view is opened.
+        connect(pLine,
+                &QLineEdit::editingFinished,
+                this,
+                [this, pBox, pLine]() {
+                    slotEditingFinished(pBox, pLine);
+                });
     }
     // Note: unlike other tags, comments can be multi-line, though while QComboBox
     // can have multi-line items its Q*Line*Edit is not suitable for editing multi-
@@ -185,7 +201,7 @@ void DlgTrackInfoMulti::init() {
     // regular tags, the two buddies require a special setup:
     // * txtCommentBox is not editable
     // * if an item is selected in txtCommentBox, the text is shown in txtComment
-    // * for multiple values, we show the <various> placeholder also in txtCommentBox
+    // * for multiple values, we show the <various> placeholder in txtComment
     // This also requires some special handling in saveTracks().
     txtCommentBox->setInsertPolicy(QComboBox::NoInsert);
     // We create a view in order to enable word-wrap.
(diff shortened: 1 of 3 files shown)


src/library/dlgtrackinfomulti.cpp
```diff
@@ -774,28 +799,60 @@ void DlgTrackInfoMulti::slotTagBoxIndexChanged() {
...
         txtComment->setPlainText(txtCommentBox->currentText());
     }
     txtCommentBox->blockSignals(false);
+    txtComment->blockSignals(false);
+}
+
+void DlgTrackInfoMulti::commentTextChanged() {
+    if (!txtComment->placeholderText().isNull() &&
+            !txtComment->toPlainText().isEmpty()) {
+        // The comboox has multiple values and has not been cleared yet,
```
cr7pt0gr4ph7, line 840: Small nitpick: `comboox` -> `combobox`

src/library/dlgtrackinfomulti.cpp
```diff
@@ -774,28 +799,60 @@ void DlgTrackInfoMulti::slotTagBoxIndexChanged() {
...
+        // and the user typed/pasted a text (might be a whitespace).
+        // Let's clear the placeholder text so we know this is new text.
+        txtCommentBox->blockSignals(true);
+        txtComment->setPlaceholderText(QString());
+        // The Clear item is not needed anymore, remove.
```
cr7pt0gr4ph7, line 845: Suggestion: `// The Clear item is not needed anymore, so remove it.`

cr7pt0gr4ph7 (changes requested):
LGTM, except for the "trim trailing whitespace" code that should really be in an utility function (and should probably handle other types of (white)space instead of just the `" "`)

src/util/string.h
```diff
@@ -84,6 +84,14 @@ inline QString convertWCStringToQString(
     return QString::fromWCharArray(wcs, static_cast<int>(wcsnlen_s(wcs, maxLen)));
 }
 
+inline QString removeTrailingWhitespaces(const QString& str) {
+    QString out = str;
+    while (out.endsWith(' ')) {
```
cr7pt0gr4ph7, line 89: Maybe it would make sense to handle other kinds of whitespace that one may encounter when pasting something from somewhere else? For example, non-breakable spaces.

Although, after checking the sources and against my expectations, [`QString::trimmed()`](https://github.com/qt/qtbase/blob/55b5e78ac31171be61ceeb932d494d5060d1fa02/src/corelib/text/qstringalgorithms_p.h#L67-L72) also [only](https://github.com/qt/qtbase/blob/55b5e78ac31171be61ceeb932d494d5060d1fa02/src/corelib/text/qstringalgorithms_p.h#L30) [handles](https://github.com/qt/qtbase/blob/96c07655c46afdd3e2da7ef7d339b518d36c83c5/src/corelib/text/qlocale_p.h#L573-L577) the [ASCII whitespace characters](https://github.com/qt/qtbase/blob/96c07655c46afdd3e2da7ef7d339b518d36c83c5/src/corelib/text/qlocale_p.h#L70-L76), and none of the UTF-8/Unicode Whitespace Characters. This is opposed to e.g. .NET [`String::Trim()`](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/String.Manipulation.cs#L2140-L2147) which [supports all Unicode whitespace characters](https://learn.microsoft.com/de-de/dotnet/api/system.char.iswhitespace?).

cr7pt0gr4ph7 (approved):

src/util/string.h
```diff
@@ -84,6 +84,16 @@ inline QString convertWCStringToQString(
...
 
+/// Remove trailing spaces from the specified string.
+/// Currently only handles simple ASCII spaces (" ").
+inline QString removeTrailingWhitespaces(const QString& str) {
+    QString out = str;
+    while (out.endsWith(' ')) {
+        out.chop(1);
+    }
+    return out;
+}
```
daschuer, line 95: Le's fix it here: 

```suggestion
/// Remove trailing spaces from the specified string.
inline QString removeTrailingWhitespaces(QString str) {
    auto it = str.crbegin();
    while (it != str.crend() && it->isSpace()) {
        ++it;
    }
    if (it != str.crbegin()) {
       str.resize(std::distance(it, str.crend()));
    }
    return str;
}
```

daschuer (changes requested):
I can confirm that this fixed the issue. 

However there is an issue that a single space is shown as empty, but becomes `<various>` after applying. I think this should be fixed. 

I did not expect that deleting everything after editing turns back to `<various>` 
Once you have deleted using the delete button the behavior changes but there is no option to go back to `<various>`.

This is OKish, but can be probably improved. 

Idea: use Backspace/Del for toggling between delete and `<various>` 
* Backspace/Del when `<various>` is shown = delete. 
* After editing and beck to an empty field = delete. 
* Backspace/Del when field is empty = `<various>`

Alternative: 
Adding `<various>` or "undo" to the menu.

ronso0:
Thanks for your thourough test!

> However there is an issue that a single space is shown as empty, but becomes <various> after applying.

I can not confirm:
* display text is `<various>`
* enter, type Space (display text is cleared)
* Apply (either Alt+A or click Apply)
* single space is removed, `<various>` is not restored (because this placeholder and the Clear item have been removed)

ronso0:
> I did not expect that deleting everything after editing turns back to <various>

This is true only before Apply.
And IMHO in that situation it's an easy way to get back to the original state (compared to Reload Metadata from files).

daschuer:
We see the same: 

- display text is `<various>`
- enter, type Space (display text is cleared)
- Apply (either Alt+A or click Apply)
- single space is removed, `<various>` is not restored (because this placeholder and the Clear item have been removed)

My original issue with it is that, I was supprised that the empty box becomes "magically" `<various>` after "Apply" or "OK". 
I would prefere to be able to confirm the state before pressing apply. 

But I agrees that this is a minor issue in an edge case. My suggestions where just a brainstorming to make it more consistent discoverable.

ronso0:
Na, it's alright. The reset to various is an error in validEditText(). isNull() should return false for a not-null (just cropped) empty string.

Will check whats going on there.

ronso0:
It turns out it's tricky to get the UX right.

> Idea: use Backspace/Del for toggling between delete and `<various>` 

Yes that indeed feels good and seems to work reliably. It's not intuitive though, so I'll add a _Reset_ item, too.

Btw by now it feels like we could as well use a custom QComboBox, but won't test that in this fix PR.

daschuer:
Yes, or a separate checkbox indicating the "will update" state ... but also a separate PR. 
Since this is a bugfix PR any any quick improvement works as well. Please give a hint when you consider this ready for merge.

ronso0:
I'm still working on this. I've been experimenting with a few variants and now have to clean it up.

ronso0:
I think this works now.
I did a lot of testing, though maybe something slipped through. Please test!

FWIW I noticed a bug with the key field:
after editing (DlgTrackInfo) or selecting on eof the items (DlgTrackInfoMulti) the display doesn't use the "Traditional" not the configured notation.
I have fix but will apply it in a separate PR

JoergAtGithub:
2.5.0, wdyt?

ronso0:
> 2.5.0, wdyt?

Yes.
Just needs some manual testing and a review ; )

src/util/string.h
daschuer: Did you consider my suggestion?

daschuer (approved):
Works like a charm now and the code LGTM. The only open point is the non universal trim function.

src/util/string.h
ronso0: Ah sorry, I was focused on the functional changes in the dialogs.

Looks good, thank you!

daschuer:
Oh sorry for the pre-commit failure .. did it in the Github GUI.

ronso0:
Never mind, fixed.

ronso0:
> Works like a charm now

Yeah, but code became even more complex and I fear I have no clue what's going on when I'll be looking at this in like 6 months :laughing:

PR #13632, opened 2024-09-06
Track Info dialogs: move metatdata buttons below color picker

ronso0:


src/library/dlgtrackinfomulti.ui
```diff
@@ -437,7 +429,7 @@
        </widget>
       </item>
 
-      <item row="0" column="0">
+      <item row="10" column="0">
```
Swiftb0y, line 432: do you know why this was `row="0"` previously?

ronso0: 🤷‍♂️ 
items with the same position are stacked, that's why we didn't notice.

Swiftb0y: so the spacer was overlayed on another object somewhere near the top of the table, effectively not existing?

ronso0: yup. that's what I noticed with QLabels when I messed up column/row values.

Swiftb0y: Great. thanks for the clarification then.

Swiftb0y (approved):
LGTM. Thank you

ronso0:
Thanks for the quick review!

PR #13634, opened 2024-09-06
WOverview: update immediately when the normalize option or Global gain changed

ronso0:
Just another step to get the same immediate update as for the scrolling waveforms.

Initially I thought about adding a control for Normalize to move some stuff out of WaveformWidgetFactory, but actually it's okay with the new signals.


JoergAtGithub:
Works as expected. Thank you!

Issue #13611, opened 2024-08-29
Metronome effect click is offset by one audio buffer size too early


jypma:
### Bug Description

- Add "Metronome" as an effect to deck 1
- That FX unit is only assigned to deck 1
- Using ALSA as audio output, buffer size ~50ms
- Align the beat grid by ear to the metronome click.

Now, visually, the actual beat grid can be seen to be 50ms too late with respect to the actual downbeat:
![Screenshot from 2024-08-29 20-00-21](https://github.com/user-attachments/assets/9d3cae4b-a5c6-41c8-b88b-195a826f2138)
This correlates with the audio buffer size. If I set the buffer size down to 5ms, and again align the grid to the downbeat by ear, it's much closer to the actual beat.

So, somehow the metronome effect is one audio buffer too early (or too late?) with creating its click.

### Version

2.4.1

### OS

Arch Linux

PR #13636, opened 2024-09-08
Fix quantization in the effect engine (metronome effect)

daschuer:
This was originally a fix for #13611 but it turns out that the beat information provided into the effect engine itself was broken. I have fixed this at the root by refactoring GroupFeatureState and the "beat_next" and "beat_prev" COs.
Than I was able to finally fix the metronome, which works now also reliable during scratching.

Not sure if there is a real use case for it, but the metronome was perfectly to test that, and also helps verifying a beat grid by ear. It would be a real fun to play any sample and not only the metronome click. That could become our first "transport" effect. I will file a feature request. 

Open is to check/fix other effects. I want to have that reviewed and merged first before building on that. 
See commits for details.


src/effects/backends/builtin/metronomeeffect.cpp
```diff
@@ -77,9 +77,8 @@ void MetronomeEffect::processChannel(
     }
 
     SINT maxFrames;
-    if (m_pSyncParameter->toBool() && groupFeatures.has_beat_length_sec) {
-        maxFrames = static_cast<SINT>(
-                engineParameters.sampleRate() * groupFeatures.beat_length_sec);
+    if (m_pSyncParameter->toBool() && groupFeatures.has_beat_length_frames) {
+        maxFrames = static_cast<SINT>(groupFeatures.beat_length_frames);
```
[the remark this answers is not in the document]
daschuer: I have replaced it by double precision in a later commit.

daschuer:
Done.

src/effects/backends/builtin/metronomeeffect.cpp
Swiftb0y, line 1: the changes here are really hard to follow because they mix offset and timing calculations with buffer indexing (mixing levels of abstraction) and thus I'm neither confident about the reliability nor safety of the code...

Do you think its feasible to reorganize it a little so its easier to understand?

daschuer: I have tried to fix that by calculate everything in frames. Maybe I was not too successful or comments are missing. 
Can you give some hints what can be improved?

Swiftb0y: I will rework it a little and open a PR to your branch ASAP.

Swiftb0y: Please take a look https://github.com/daschuer/mixxx/pull/97

daschuer:
I have added the first commit of your branch from https://github.com/daschuer/mixxx/pull/97
Can we merge this first, before doing the std::span refactoring? 
I am not convinced that this reduces complexity, because it brings back the sample count where I am very happy to with the frame based counters.

Swiftb0y:
> Can we merge this first, before doing the std::span refactoring?

Sure, but unfortunately I encountered a couple bugs during testing:

1. Sync on the Metronome is off: expected it should continue clicking based on the fixed rate of the knob. Currently Observed: it only does not click at all, only when enabling the effect, and with its transient chopped of.
2. Sync on the Metronome is on but track has no beatgrid. Expected: not quite sure actually whether it should stay silent or just play as if sync was turned off. Observed: it just clicks once when enabling, though this time its the full sample.

Both of these issues are regressions.

daschuer:
The described issues are fixed now. It was caused by a faulty fix to avoid SINT overflow.

Swiftb0y:
I don't understand the fix... There also remains the issue where the first click is with sync disabled is muffled.

daschuer:
> There also remains the issue where the first click is with sync disabled is muffled.

This is a "feature", the fade in of the effect. But we can safely disable this here. 

I have removed a redundant check and added comments to the last summit.

src/effects/backends/builtin/metronomeeffect.cpp
```diff
@@ -57,71 +58,87 @@ void MetronomeEffect::processChannel(
...
+                        beatLength * -1 *
+                        (1 - *groupFeatures.beat_fraction_buffer_end);
+            }
+
+            if (bufferEnd > beatToBufferEnd) {
+                // We have a new beat before the current buffer ends
+                nextClickStart = bufferEnd - beatToBufferEnd;
+            }
+        } else {
+            nextClickStart = *groupFeatures.beat_length_frames;
```
Swiftb0y, line 125: Can this actually happen?

daschuer: No, I have added it only for completeness. We may consider to use both in a optional struct to get around it.

Swiftb0y: do you mean `beat_length_frames` and `beat_fraction_buffer_end` or `beat_fraction_buffer_end` and `scratch_rate`? because according to `BpmControl::collectFeatures` if `rate_ratio` is 0, we can get a `beat_fraction_buffer_end` but not a `beat_length_frames`

daschuer: We always have it all in reality, because we do not allow a rate of 0 but it is possible via writing the CO. 
We should probably treat this case in the same way we treat scratch = 0

Swiftb0y: yes, I agree. I implemented that in a follow up. will push shortly.

daschuer: I am already on it.

daschuer:
Done, please have a look.

src/effects/backends/builtin/metronomeeffect.cpp
Swiftb0y: sure that works, but it doesn't eliminate the original case I asked about (which is when `beat_length_frames.has_value()` but `!beat_fraction_buffer_end.has_value()`).

daschuer: This can not happen. But looking at the API it, can. `beat_fraction_buffer_end.has_value() == false` means the track has no value. In this case `beat_length_frames.has_value()` is always false.

Swiftb0y: right, so that should be reflected in `GroupFeatureState`, do you agree?

Swiftb0y:
fyi, I got https://github.com/daschuer/mixxx/pull/97 into a state I'm happy with. Please take a look.

src/effects/backends/builtin/metronomeeffect.cpp
daschuer: I am not sure, I because we may consider a sound source that has bpm info but no phase info.

daschuer:
I have adopted the full span based approach. But I have not adopted the rest, because of the std::size_t / std::ptrdiff_t / SINT clash with gsl/span. I also must admit that I can not fully understand how your code works.

Swiftb0y:
I think the primary advantage of my span approach was not the fact that it uses span (it just made it easier for myself to do the cleanup I wanted to do). The real improvement was separating the messy sync code from the rest. But I'm okay with the current state as a temporary stopgap.

Swiftb0y (approved):
This LGTM for now. I will retarget the follow up at `main`.

PR #13638, opened 2024-09-08
Library: add overview column

ronso0:
This is based on @ninomp's initial POC commit 9932cfcba93a2931271e9e8a612ead5b1374a32b
https://github.com/ninomp/mixxx/tree/overviewsinlibrary

I removed the disabled parts and added some TODOs/concerns.
(I'll split up some of the commits so it's a bit easier to review)
image

* all overview renderers are now in src/waveform/overviews, i.e. WOverView and OverviewDelegate use the same renderer
(I adopted the RGB simplifications made by @**Nino MP**, and simplified the others, too)
* the renderer selected for the decks is also used in the library (incl. live update when changing it in the preferences)

Peforms nicely, though I bet some parts can be optimized.
For example, why pass a `mixxx::DbConnectionPoolPtr` to `static FutureResult OverviewCache::prepareOverview` to create a AnyalysisDAO for each request, and not a AnyalysisDAO pointer (probably in TrackCollection)?

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -258,7 +258,7 @@ void BaseTrackTableModel::initHeaderProperties() {
             defaultColumnWidth() * 6);
     setHeaderProperties(ColumnCache::COLUMN_LIBRARYTABLE_WAVESUMMARYHEX,
             tr("Overview"),
-            defaultColumnWidth());
+            defaultColumnWidth() * 8);
     setHeaderProperties(
             ColumnCache::COLUMN_LIBRARYTABLE_PREVIEW,
             tr("Preview"),
@@ -1217,17 +1217,23 @@ void BaseTrackTableModel::slotRefreshCoverRows(
     emitDataChangedForMultipleRowsInColumn(rows, column);
 }
 
-void BaseTrackTableModel::slotOverviewReady(int row) {
-    const int column = fieldIndex(LIBRARYTABLE_COVERART);
+void BaseTrackTableModel::slotOverviewReady(TrackId trackId) {
+    const int column = fieldIndex(LIBRARYTABLE_WAVESUMMARYHEX);
     VERIFY_OR_DEBUG_ASSERT(column >= 0) {
         return;
     }
-    QModelIndex idx = index(row, column);
-    emit dataChanged(idx, idx);
+    const QList<int> rows = getTrackRows(trackId);
+    for (int row : rows) {
+        if (row == -1) {
+            continue;
+        }
+        QModelIndex idx = index(row, column);
+        emit dataChanged(idx, idx);
+    }
 }
 
 void BaseTrackTableModel::slotOverviewChanged(TrackId trackId) {
-    const int column = fieldIndex(LIBRARYTABLE_COVERART);
+    const int column = fieldIndex(LIBRARYTABLE_WAVESUMMARYHEX);
     VERIFY_OR_DEBUG_ASSERT(column >= 0) {
         return;
     }
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -14,6 +14,7 @@
 #include "library/tabledelegates/keydelegate.h"
 #include "library/tabledelegates/locationdelegate.h"
 #include "library/tabledelegates/multilineeditdelegate.h"
+#include "library/tabledelegates/overviewdelegate.h"
 #include "library/tabledelegates/previewbuttondelegate.h"
 #include "library/tabledelegates/stardelegate.h"
 #include "library/trackcollection.h"
@@ -69,6 +70,7 @@ const QStringList kDefaultTableColumns = {
         LIBRARYTABLE_TITLE,
         LIBRARYTABLE_TRACKNUMBER,
         LIBRARYTABLE_YEAR,
+        LIBRARYTABLE_WAVESUMMARYHEX,
 };
 
 inline QSqlDatabase cloneDatabase(
@@ -254,6 +256,9 @@ void BaseTrackTableModel::initHeaderProperties() {
             ColumnCache::COLUMN_TRACKLOCATIONSTABLE_LOCATION,
             tr("Location"),
             defaultColumnWidth() * 6);
+    setHeaderProperties(ColumnCache::COLUMN_LIBRARYTABLE_WAVESUMMARYHEX,
+            tr("Overview"),
+            defaultColumnWidth());
     setHeaderProperties(
             ColumnCache::COLUMN_LIBRARYTABLE_PREVIEW,
             tr("Preview"),
@@ -526,6 +531,17 @@ QAbstractItemDelegate* BaseTrackTableModel::delegateForColumn(
         return pCoverArtDelegate;
     } else if (index == fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_KEY)) {
         return new KeyDelegate(pTableView);
+    } else if (index == fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_WAVESUMMARYHEX)) {
+        auto* pOverviewDelegate = new OverviewDelegate(pTableView);
+        connect(pOverviewDelegate,
+                &OverviewDelegate::overviewReady,
+                this,
+                &BaseTrackTableModel::slotOverviewReady);
+        connect(pOverviewDelegate,
+                &OverviewDelegate::overviewChanged,
+                this,
+                &BaseTrackTableModel::slotOverviewChanged);
+        return pOverviewDelegate;
     }
     return nullptr;
 }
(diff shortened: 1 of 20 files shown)


ninomp:
> For example, why pass a mixxx::DbConnectionPoolPtr to static FutureResult OverviewCache::prepareOverview to create a AnyalysisDAO for each request, and not a AnyalysisDAO pointer (probably in TrackCollection)?

IIUC Each thread needs its own database connection, we cannot share/reuse them.

ywwg:
I will def review this when you think it's ready for a look

ronso0:
> I will def review this when you think it's ready for a look

In case you refer to this PR (thanks!), I think it's better to wait for @ninomp's update.
I did this primarily to get it working for myself so I can test it in my live builds. (should probably add some VERIFY_OR_DEBUG_ASSERT here and there, just in case.. :laughing: )

src/library/basetracktablemodel.cpp
```diff
@@ -1201,6 +1217,34 @@ void BaseTrackTableModel::slotRefreshCoverRows(
...
+void BaseTrackTableModel::slotOverviewReady(TrackId trackId) {
+    const int column = fieldIndex(LIBRARYTABLE_WAVESUMMARYHEX);
+    VERIFY_OR_DEBUG_ASSERT(column >= 0) {
+        return;
+    }
+    const QList<int> rows = getTrackRows(trackId);
+    for (int row : rows) {
+        if (row == -1) {
+            continue;
+        }
```
Swiftb0y, line 1229: why is this needed here but not in `slotOverviewChanged`?

src/library/librarytablemodel.cpp
```diff
@@ -68,7 +68,7 @@ bool LibraryTableModel::isColumnInternal(int column) {
     return column == fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_ID) ||
             column == fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_URL) ||
             column == fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_CUEPOINT) ||
-            column == fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_WAVESUMMARYHEX) ||
+            // column == fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_WAVESUMMARYHEX) ||
```
Swiftb0y, line 71: remove completely, right?

src/library/overviewcache.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+mixxx::Logger kLogger("OverviewCache");
+
+QString pixmapCacheKey(TrackId trackId, QSize size, WOverview::Type type) {
+    return QString("Overview_%1_%2_%3_%4")
+            .arg(static_cast<int>(type))
+            .arg(trackId.toString())
+            .arg(size.width())
+            .arg(size.height());
+}
```
Swiftb0y, line 27: using a string as a cache key is not particularly efficient. a struct with a `qHash` implementation would be better.

src/library/overviewcache.cpp
```diff
@@ -0,0 +1,156 @@
...
+
+// The transformation mode when scaling images
+const Qt::TransformationMode kTransformationMode = Qt::SmoothTransformation;
+
+inline QImage resizeImageSize(const QImage& image, QSize size) {
```
Swiftb0y, line 32:
```suggestion
QImage resizeImageSize(const QImage& image, QSize size) {
```

src/waveform/overviews/waveformoverviewrenderer.h
```diff
@@ -0,0 +1,60 @@
...
+#include "util/singleton.h"
+#include "waveform/waveform.h"
+#include "widget/woverview.h"
+
+class QPainter;
+
+class WaveformOverviewRenderer : public Singleton<WaveformOverviewRenderer> {
+  public:
+    QImage render(ConstWaveformPointer, WOverview::Type type);
+    void drawWaveformPartRGB(
```
Swiftb0y, line 14: this doesn't have internal state, nor does it need to be a singleton. Consider making this free functions instead and removing the static data. Pass it instead when calling the functions.

Swiftb0y (commented):
I'm afraid this needs a bunch of work before I trust it...

ronso0:
FYI I'll keep pushing fixups to hopefully make this safer and maybe a bit faster for my own builds.

src/library/overviewcache.cpp
ronso0: (I think most of this is adopted from CoverArtCache)

ronso0: What about this?
https://doc.qt.io/qt-6/qpixmapcache-key.html#details
> Use [QPixmapCache::insert](https://doc.qt.io/qt-6/qpixmapcache.html#insert)() to receive an instance of Key generated by the pixmap cache. You can store the key in your own objects for a very efficient one-to-one object-to-pixmap mapping.

Does such a one-to-one map QString <--> Key map/hash make sense performance wise?

Swiftb0y: Probably not, you're right. I didn't see that this was using QPixmapCache internally. The only improvement then is 
```suggestion
QString pixmapCacheKey(TrackId trackId, QSize size, WOverview::Type type) {
    return QStringLiteral("Overview_%1_%2_%3_%4")
            .arg(static_cast<int>(type),
                trackId.toString(),
                size.width(),
                size.height());
}
```

ninomp:
I just created a PR that contains the work I've done in the past week or so: #13644

ywwg:
> I just created a PR that contains the work I've done in the past week or so: #13644

@ronso0 @ninomp is the intent that we work on that PR instead of this one?  If so let's close this one so it's more clear where the active development is

ronso0:
Jup, I'm only push updates to this one to polish it a bit for my personal branch until #13644 is usable for me.

I can as well close this PR if it creates too much confusion.

ninomp:
Well, personally, I'm not sure how to proceed. I tried implementing so many things, but saw little improvement. Could we have a call/meeting to discuss these things before deciding how to proceed?

ywwg:
yeah I would be willing to be on such a call though I don't have a lot of expertise in this particular area

ronso0:
I'll close this one, moved it to https://github.com/ronso0/mixxx/pull/79

ronso0:
Personally I prefer a written discussion when it comes to technical stuff because I have time to look things up and think about it.

IMO we could as well discuss it on Zulip.
Just some quick feedback:
> * moving all rendering logic away from WOverview so it can be shared/reused

Yes. The OverviewRenderer works great so far (can be improved, see Swiftb0y's review above).
Not sure if it needs a dedicated thread. I didn't notice any GUI performance regression yet.

> * two-level cache for rendered overviews

I didn't fully understand the purpose of the first QImage cache. Is it supposed to speed up loading of  overviews in WOverview when loading a track from the library? Or just save some CPU?
IMO loading overviews of tracks with waveform data is currently (main and this branch) so fast (fast enough) that I don't **see** the benefit tbh.

ywwg:
One model that has worked well for me in the past is writing up a short doc with the relevant issues / references and then a meeting to talk it over. I think a meeting can be more effective to reach a consensus, but I agree that having a specific set of text to discuss helps in doing that.

ninomp:
> Yes, if you refer to only the waveform image rendering. IMO we should not overload the tracks table with any kind of markers

Actually, I was referring to overview normalization.

> I didn't fully understand the purpose of the first QImage cache. Is it supposed to speed up loading of overviews in WOverview when loading a track from the library? Or just save some CPU?

The idea was to copy implementation in WOverview where we have a full sized QImage which is updated as the track is being analyzed so that we can show analysis progress. Yes, I was trying to save a bit of CPU time, but it seems that I was just trading a bit of CPU time for a lot of memory usage and I agree with you that this isn't really worth it.

@ronso0 I'm testing your branch and I can say that you managed to get this working very good. I'm currently thinking/considering closing my PR and going forward with yours.

ronso0:
> @ronso0 I'm testing your branch and I can say that you managed to get this working very good.

Thanks, you provided a great foundation!
I'm very happy with the current state, too, now that clearing and reanalyzing ie. all updates in decks and library works just fine.

ronso0:
I'll try to summarize, these would be nice to have:
* normalization / ReplayGain
yes, more details are good, but I also like to see if it's a quiet or loud track
* analysis progress
yes, thinking about it, at least something like a bar as progress indicator would be cool in the library
* half / mono-mixdown
need to check with a mockup if that makes sense / looks good when it's bottom-aligned
* ?

ronso0:
Ha, all those items are now implemented :tada: 

@ninomp I'd appreciate if you'd take a look at https://github.com/ronso0/mixxx/pull/79 and let me know what you think.

ninomp:
@ronso0 I left a few comments on your PR.

PR #13648, opened 2024-09-13
fix: prevent div by zero and use correct config key

acolombier:
This fixes a regression introduced in #13401 where the UI would freeze if the rate control is reduce to zero.

This also adjust the config key to match the `snake_case`

### DIFF
--- a/src/preferences/dialog/dlgprefwaveform.cpp
+++ b/src/preferences/dialog/dlgprefwaveform.cpp
@@ -73,7 +73,7 @@ DlgPrefWaveform::DlgPrefWaveform(
 
     m_pOverviewMinuteMarkersControl = std::make_unique<ControlObject>(
             ConfigKey(QStringLiteral("[Waveform]"),
-                    QStringLiteral("DrawOverviewMinuteMarkers")));
+                    QStringLiteral("draw_overview_minute_markers")));
     m_pOverviewMinuteMarkersControl->setReadOnly();
 
     // Populate untilMark options
@@ -304,7 +304,7 @@ void DlgPrefWaveform::slotUpdate() {
     }
 
     bool drawOverviewMinuteMarkers = m_pConfig->getValue(
-            ConfigKey("[Waveform]", "DrawOverviewMinuteMarkers"), true);
+            ConfigKey("[Waveform]", "draw_overview_minute_markers"), true);
     overviewMinuteMarkersCheckBox->setChecked(drawOverviewMinuteMarkers);
     m_pOverviewMinuteMarkersControl->forceSet(drawOverviewMinuteMarkers);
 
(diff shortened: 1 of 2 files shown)


acolombier:
>> I've noticed a bit late that this PR is introducing new settings using PascalCase instead of snake_case. Is that something that we can fix before the next release?

> I don't know if we settled on a convention, but if we want choose one consistently, we should (and can) do it before a release. Though we can only changes keys that have not yet ended up in settings of stable releases.

@Swiftb0y indeed I don't think it was ever establish and properly committed. Over my last contribution, I always went with `snake_case` as I thought it actually was the standard, but I'd be happy to update to `PascalCase` or `kebabCase`. 
Just conscious that 2.6 is about to bring a lot of new keys (config and COs) and it would be great to establish a standard on newly introduce keys :)

Swiftb0y:
I agree. Lets discuss this on zulip so everybody in charge sees it. ;)

daschuer:
@ronso0 Knows probably the solution!?

ronso0:
42

ronso0:
sorry, I think I missed the question. ?

Swiftb0y:
@ronso0 what casing should our usersettings configkeys be? `snake_case`, `PascalCase`, etc?

src/widget/woverview.cpp
```diff
@@ -725,6 +725,10 @@ void WOverview::drawMinuteMarkers(QPainter* pPainter) {
         return;
     }
 
+    if (m_pRateRatioControl->get() == 0) {
+        return;
+    }
```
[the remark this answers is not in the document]
acolombier: It happens in `WOverview::samplePositionToSeconds`, however I think it make sense to handle it and exist early

acolombier:
@ronso0 @daschuer I have create a thread in [Zulip](https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/Naming.20convention.20for.20keys.20in.20new.20COs.20and.20settings) to summarise the situation

src/widget/woverview.cpp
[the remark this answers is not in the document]
acolombier: I added an assert just for safety

ronso0:
> @ronso0 what casing should our usersettings configkeys be? `snake_case`, `PascalCase`, etc?

Ah okay, I was confused by the fact that I should know/decide it.

I think we already agreed on snake_case for (skin) controls, didn't we? Ime we haven't enforced it, IMO because it feels weird to break the legacy PascalCase scheme when adding new controls in a certain context/namespace.

so it should be snake_case 🤷🏼‍♂️

src/preferences/dialog/dlgprefwaveform.cpp
Swiftb0y, line 1: should this go into 2.5 or main? I don't remember what branch this landed in.

Swiftb0y: the overview_minute_markers feature I mean.

acolombier: This feature was only merged in main!

Swiftb0y: but after the 2.5 branch off, right?

acolombier: Yes, that's right

Swiftb0y: Great. Thanks for clarifying.

Swiftb0y (approved):
LGTM. Thank you.

PR #13654, opened 2024-09-15
refactor: use higher-level `std::span` based logic

Swiftb0y:
This is https://github.com/daschuer/mixxx/pull/97 rebased on main with the primary goal of making the logic easier to understand.

### DIFF
--- a/src/effects/backends/builtin/metronomeeffect.cpp
+++ b/src/effects/backends/builtin/metronomeeffect.cpp
@@ -62,16 +62,16 @@ std::span<CSAMPLE> syncedClickOutput(double beatFractionBufferEnd,
 
 // static
 QString MetronomeEffect::getId() {
-    return "org.mixxx.effects.metronome";
+    return QStringLiteral("org.mixxx.effects.metronome");
 }
 
 // static
 EffectManifestPointer MetronomeEffect::getManifest() {
-    EffectManifestPointer pManifest(new EffectManifest());
+    auto pManifest = EffectManifestPointer::create();
     pManifest->setId(getId());
     pManifest->setName(QObject::tr("Metronome"));
-    pManifest->setAuthor("The Mixxx Team");
-    pManifest->setVersion("1.0");
+    pManifest->setAuthor(QObject::tr("The Mixxx Team"));
+    pManifest->setVersion(QStringLiteral("1.0"));
     pManifest->setDescription(QObject::tr("Adds a metronome click sound to the stream"));
     pManifest->setEffectRampsFromDry(true);
 
@@ -79,7 +79,7 @@ EffectManifestPointer MetronomeEffect::getManifest() {
     // The maximum is at 128 + 1 allowing 128 as max value and
     // enabling us to pause time when the parameter is above
     EffectManifestParameterPointer period = pManifest->addParameter();
-    period->setId("bpm");
+    period->setId(QStringLiteral("bpm"));
     period->setName(QObject::tr("BPM"));
     period->setDescription(QObject::tr("Set the beats per minute value of the click sound"));
     period->setValueScaler(EffectManifestParameter::ValueScaler::Logarithmic);
@@ -88,7 +88,7 @@ EffectManifestPointer MetronomeEffect::getManifest() {
 
     // Period unit
     EffectManifestParameterPointer periodUnit = pManifest->addParameter();
-    periodUnit->setId("sync");
+    periodUnit->setId(QStringLiteral("sync"));
     periodUnit->setName(QObject::tr("Sync"));
     periodUnit->setDescription(QObject::tr(
             "Synchronizes the BPM with the track if it can be retrieved"));
@@ -101,8 +101,8 @@ EffectManifestPointer MetronomeEffect::getManifest() {
 
 void MetronomeEffect::loadEngineEffectParameters(
         const QMap<QString, EngineEffectParameterPointer>& parameters) {
-    m_pBpmParameter = parameters.value("bpm");
-    m_pSyncParameter = parameters.value("sync");
+    m_pBpmParameter = parameters.value(QStringLiteral("bpm"));
+    m_pSyncParameter = parameters.value(QStringLiteral("sync"));
 }
 
 void MetronomeEffect::processChannel(
--- a/src/effects/backends/builtin/metronomeeffect.cpp
+++ b/src/effects/backends/builtin/metronomeeffect.cpp
@@ -1,6 +1,7 @@
 #include "metronomeeffect.h"
 
 #include <cmath>
+#include <cstddef>
 #include <limits>
 #include <optional>
 #include <span>
@@ -60,6 +61,15 @@ std::span<CSAMPLE> syncedClickOutput(double beatFractionBufferEnd,
         return output.last(beatToBufferEndSamples);
     }
     return {};
+}
+
+std::span<CSAMPLE> unsyncedClickOutput(mixxx::audio::SampleRate framesPerSecond,
+        std::size_t framesSinceLastClick,
+        double bpm,
+        std::span<CSAMPLE> output) {
+    std::size_t offset = framesSinceLastClick %
+            framesPerBeat(framesPerSecond, bpm);
+    return subspan_clamped(output, offset * mixxx::kEngineChannelOutputCount);
 }
 
 } // namespace
@@ -163,20 +173,17 @@ void MetronomeEffect::processChannel(
     playMonoSamplesWithGain(subspan_clamped(click, gs->framesSinceLastClick), output, gain);
     gs->framesSinceLastClick += engineParameters.framesPerBuffer();
 
-    std::span<CSAMPLE> outputBufferOffset = [&] {
-        if (shouldSync && hasBeatInfo) {
-            return syncedClickOutput(*groupFeatures.beat_fraction_buffer_end,
-                    groupFeatures.beat_length,
-                    output);
-        } else {
-            // EngineParameters::sampleRate in reality returns the framerate.
-            mixxx::audio::SampleRate framesPerSecond = engineParameters.sampleRate();
-            std::size_t offset = gs->framesSinceLastClick %
-                    framesPerBeat(framesPerSecond,
-                            m_pBpmParameter->value());
-            return subspan_clamped(output, offset * mixxx::kEngineChannelOutputCount);
-        }
-    }();
+    std::span<CSAMPLE> outputBufferOffset = shouldSync && hasBeatInfo
+            ? syncedClickOutput(*groupFeatures.beat_fraction_buffer_end,
+                      groupFeatures.beat_length,
+                      output)
+            : unsyncedClickOutput(
+                      engineParameters
+                              .sampleRate(), // engineParameters::sampleRate()
+                                             // in reality returns the frameRate
+                      gs->framesSinceLastClick,
+                      m_pBpmParameter->value(),
+                      output);
 
     if (!outputBufferOffset.empty()) {
         gs->framesSinceLastClick = playMonoSamplesWithGain(click, outputBufferOffset, gain);
--- a/src/effects/backends/builtin/metronomeeffect.cpp
+++ b/src/effects/backends/builtin/metronomeeffect.cpp
@@ -1,22 +1,61 @@
 #include "metronomeeffect.h"
 
+#include <cmath>
+#include <limits>
+#include <optional>
+#include <span>
+
+#include "audio/types.h"
 #include "effects/backends/effectmanifest.h"
 #include "engine/effects/engineeffectparameter.h"
+#include "engine/engine.h"
 #include "metronomeclick.h"
-#include "util/math.h"
 #include "util/sample.h"
+#include "util/types.h"
 
 namespace {
 
-void playMonoSamples(std::span<const CSAMPLE> monoSource, std::span<CSAMPLE> output) {
-    const auto outputBufferFrames = output.size() / mixxx::kEngineChannelOutputCount;
-    SINT framesPlayed = std::min(monoSource.size(), outputBufferFrames);
+std::size_t playMonoSamples(std::span<const CSAMPLE> monoSource, std::span<CSAMPLE> output) {
+    const std::size_t outputBufferFrames = output.size() / mixxx::kEngineChannelOutputCount;
+    std::size_t framesPlayed = std::min(monoSource.size(), outputBufferFrames);
     SampleUtil::addMonoToStereo(output.data(), monoSource.data(), framesPlayed);
+    return framesPlayed;
 }
 
-double framesPerBeat(mixxx::audio::SampleRate sampleRate, double bpm) {
-    double framesPerMinute = sampleRate * 60;
-    return framesPerMinute / bpm;
+template<class T>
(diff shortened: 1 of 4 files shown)


Swiftb0y:
noticed recently that the there was no use of the metaparameter within the metronome (and the gain always having the same volume being annoying), so I added a little simple gain knob in the newest commit.

daschuer:
Unfortunately the non synced mode does not produce a stady click. It sounds more like as rhythm.

src/effects/backends/builtin/metronomeeffect.cpp
```diff
@@ -1,46 +1,86 @@
...
+        CSAMPLE_GAIN gain) {
+    const std::size_t outputBufferFrames = output.size() / mixxx::kEngineChannelOutputCount;
+    std::size_t framesPlayed = std::min(monoSource.size(), outputBufferFrames);
+    SampleUtil::addMonoToStereoWithGain(gain, output.data(), monoSource.data(), framesPlayed);
+    return framesPlayed;
+}
+
+template<class T>
+std::span<T> subspan_clamped(std::span<T> in, typename std::span<T>::size_type offset) {
+    // TODO (Swiftb0y): should we instead create a wrapper type that implements
```
daschuer, line 30: I think we should remove this function altogether. 
It does not make sense to replace here an illformated span with an empty one and than check for it in the outer scope. Avoid the empty span as well.

Swiftb0y: I can't follow you. There is no ill-formed span (any operation that would produce one is already UB).

Swiftb0y:
> Unfortunately the non synced mode does not produce a stady click. It sounds more like as rhythm.

Fixed. mistakenly used `kMaxEngineChannelInputCount` instead of `kEngineChannelOutputCount`.

Swiftb0y:
> Unfortunately the non synced mode does not produce a stady click. It sounds more like as rhythm.

fixed. mistakenly used `kMaxEngineChannelInputCount` instead of `kEngineChannelOutputCount`.

src/effects/backends/builtin/metronomeeffect.cpp
```diff
@@ -1,46 +1,86 @@
...
+    // UB-free "clamped" operations?
+    return in.subspan(std::min(offset, in.size()));
 }
 
-double framesPerBeat(mixxx::audio::SampleRate sampleRate, double bpm) {
-    double framesPerMinute = sampleRate * 60;
-    return framesPerMinute / bpm;
+std::size_t samplesPerBeat(mixxx::audio::SampleRate sampleRate, double bpm) {
```
daschuer, line 35: The old name was better, because mixxx::audio::SampleRate is actually frame rate. We used the name sample rate because this is more common for audio and also used in external libraries. In a second step it can be than converted to monoSmaples whith the same value.

src/effects/backends/builtin/metronomeeffect.cpp
```diff
@@ -83,73 +137,45 @@ void MetronomeEffect::processChannel(
...
-            // no transport, continue until the current click has been fully played
-            if (gs->m_framesSinceClickStart < clickSize) {
-                gs->m_framesSinceClickStart += engineParameters.framesPerBuffer();
-            }
-            return;
+            std::size_t samplesSinceLastClick =
+                    gs->framesSinceLastClick * mixxx::kEngineChannelOutputCount;
+            std::size_t offset = samplesSinceLastClick %
+                    samplesPerBeat(engineParameters.sampleRate(),
+                            m_pBpmParameter->value());
```
daschuer, line 173: Here is somewhere a 2x issue.

src/effects/backends/builtin/metronomeeffect.cpp
```diff
@@ -83,73 +137,45 @@ void MetronomeEffect::processChannel(
...
-            }
-
-            if (bufferEnd > beatToBufferEnd) {
-                // We have a new beat before the current buffer ends
-                nextClickStart = bufferEnd - beatToBufferEnd;
-            }
+    playMonoSamplesWithGain(subspan_clamped(click, gs->framesSinceLastClick), output, gain);
+    gs->framesSinceLastClick += engineParameters.framesPerBuffer();
+
+    std::span<CSAMPLE> outputBufferOffset = [&] {
```
daschuer, line 163: Can't we keep the original logic? It is hard to check if all conditions are still considered.

src/effects/backends/builtin/metronomeeffect.cpp
Swiftb0y: Oh, are you sure? I wasn't aware of that.

src/effects/backends/builtin/metronomeeffect.cpp
Swiftb0y: No. That is the point of this PR. Independent of the `std::span` usage, this was the part I wanted to improve because I very much struggled with understanding the logic you came up with here. So I rewrote it in a way I found more intuitive.
I'm confident that all the logic is still there (except #13733 because I couldn't reproduce that here). It just has been disentangled into smaller functions.

We're at an impasse now because we simply disagree about what is more readable. I don't understand your code, you don't understand mine. So I suggest we ask other team members to evaluate what solution they prefer and decide based on that.

src/effects/backends/builtin/metronomeeffect.cpp
Swiftb0y: right, but `offset <  size()` should not be part of the businesslogic. Explicit bounds checks are about as low level you can get so it doesn't make sense to have in high-level business logic. Thats explicitly why I put it into a utility function. From a high level perspective there is now only `span`. Either it has some sample content to play, or its empty and has none. This is a clear and intuitive API separation thats well worth paying [two instructions](https://compiler-explorer.com/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:6,positionColumn:1,positionLineNumber:6,selectionStartColumn:1,selectionStartLineNumber:6,startColumn:1,startLineNumber:6),source:'%23include+%3Calgorithm%3E%0A%0Aint+min(int+a,+int+b)+%7B%0A++++return+std::min(a,+b)%3B%0A%7D%0A%0A//+comparision+to+demonstrate+which+instructions%0A//+can!'t+be+omitted+due+to+the+function+call+mechanism%0Aint+id(int+a)+%7B%0A++++return+a%3B%0A%7D'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g142,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-O1',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+14.2+(Editor+%231)',t:'0')),k:33.333333333333336,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+14.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+14.2+(Compiler+%231)',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) for.

src/effects/backends/builtin/metronomeeffect.cpp
Swiftb0y: fixed in fixup.

src/effects/backends/builtin/metronomeeffect.cpp
Swiftb0y: yup, confirmed. fixed in fixup

src/effects/backends/builtin/metronomeeffect.cpp
JoergAtGithub: The new code looks much cleaner, but I can also not see, if it's functional identical.
May I suggest to implement a unit test for this code first, and than apply the refactured code in a second commit?

Swiftb0y: Sure, do we already test other effect code or do I need to come up with infrastructure for that first?

Swiftb0y: Regardless, It'll probably take some time until I come up with reasonable test code...

Swiftb0y: so I experimented a bit with this, but coming up with tests is incredibly tedious and brittle.

Swiftb0y: yeah, so spinning this a bit further. I think verifying the correctness by inspecting the buffer output is not sensible. At least not more than doing some QA testing by ear or by code review.

Swiftb0y:
Heres my attempt at walking through the diff and explaining it along the way (text is always about the diff above)

image
The check is not needed anymore because subspan_clamped already takes care of it. calling `playMonoSamples` on an empty buffer doesn't do anything as you might expect.
Since we just played some samples, we need to track that in `gs->framesSinceLastClick`. 


image
image
these are equivalent. Only this time the entire function body is encapsulated in its own function. 

image
image
image

Here the preconditions are not satisfied so we can simply return with an empty span, which means nothing needs to be done. The else case handles #13727 which is not necessary because we already advance `gs->framesSinceLastClick` directly after (potentially) playing the buffer.

image
image

These two chunks are mathematically equivalent just rewritten in a way that I found a little easier to understand/meaningful. 

image
image
image

These are also equivalent, only that this time I didn't make the assumption that `gs->framesSinceLastClick` was sufficiently small. This was an issue when playing a slow track and having the metronome synced and then switching to an unsynced click with sufficiently high speed (more than twice the track speed IIRC). 

image

At this point we know that if we need to play new click we know where to play it in the buffer so just play it and record how far into the buffer that was. No need to shift `framesSinceLastClick` around "due to seeking" (honestly I don't understand why this is needed even today). 

In summary:

I highly prefer this solution because it clearly breaks the problem down into subproblems (do we need to play the (remaining) click? where?) that are easily composed back together while hiding many of the uglier details about buffer management in the lower-level functions. I'm fairly confident that this refactor does not introduce regressions while raising maintainability significantly.

ChasonDeshotel (changes requested):
Agree fully with @Swiftb0y. This is a solid improvement, nicely done

>I see UB as anything can happen, but most likely there are no boundary checks to keep subspan() blazing fast, leading to a size_t overflow or such.

This is what subspan_clamped does. It effectively mitigates the risk of undefined behavior by ensuring that the offset passed to subspan() never exceeds the size of the span.

As for the performance impact:

> Cost of subspan_clamped
> 
> The subspan_clamped function adds a single std::min operation, which is a straightforward comparison between two size_t values. This operation:
> 
>     Takes constant time: It’s an O(1) operation, meaning it doesn’t depend on the size of the data.
>     Is negligible in CPU cycles: On modern processors, a comparison and assignment generally take only a few nanoseconds, which is practically negligible, even for audio processing.

Some performance improvements that would be nice to add:

- Make subspan_clamped inline
- Store `engineParameters.samplesPerBuffer()`, `engineParameters.framesPerBuffer()`, and `engineParameters.sampleRate()` as local variables -- this would negate the additional cycles introduced from the extra function calls added with this PR. Not sure how/when metronomeeffect is instantiated or how we handle the case when the audio device changes elsewhere in the codebase

However, I wouldn't consider those a blocker. I imagine there's a lot better stuff to spend our time benchmarking if performance is an issue. Overall, these changes make the code more robust and maintainable without significant performance sacrifices.

Marked 'request changes' for inlining subspan_clamped. On the other, maybe @Swiftb0y has time to dig in or someone knows the answer already, if not, nbd.

Swiftb0y:
Thank you @ChasonDeshotel for chiming in here. As for your suggestions.

> Make subspan_clamped inline

It already is if the compiler decide its worth it (likely is).

> Store engineParameters.samplesPerBuffer(), engineParameters.framesPerBuffer(), and engineParameters.sampleRate() as local variables

Not needed as those accessors are already visible to the compiler, thus also likely inlined. 

Feel free to verify these claims in compiler explorer.

daschuer:
After a bit of testing I can confirm that this change works without regressions. I also confirm that the full use of span and the better separation of tasks improve the readability.  

However there is still an issue with unnecessary playMonoSamplesWithGain() which impairs legibility. I have just proposed https://github.com/Swiftb0y/mixxx/pull/19

This avoids to call playMonoSamplesWithGain() when not necessary and solves some other issue in one go. It makes the code better traceable, because the cases where actually a new click is started can be better identified, it removes useless calls from the flow saving some ns and avoids the creation of empty spans entirely. 

The patch also made the handling of the static bpm and quatized mode  along with calling span::subspan() and span::last() symmetric and better comparable. 

What do you think?

daschuer:
The other issue is that the gain knob does not look right: 
grafik

Did you consider to make it from 0 ... 4 ratio like the isolator gain knobs. This what unity is at center and it allows to go to full zero.

daschuer (commented):
.

Swiftb0y:
> The other issue is that the gain knob does not look right

Thats somewhat on purpose. The center of the knob is on 0db but since the sample is already quite close to 0db absolute I didn't want to allow for much range above 0db either, thats why its so asymmetrical. -24db is basically quiet even though its technically 0, its quiet enough. I figured a db readout was more informative than a unitless ratio. Do you disagree?

> After a bit of testing I can confirm that this change works without regressions. I also confirm that the full use of span and the better separation of tasks improve the readability.

Thank you. Lets continue the code quality discussion on https://github.com/Swiftb0y/mixxx/pull/19

daschuer:
> I figured a db readout was more informative than a unitless ratio. Do you disagree?

I agree to that. That is a general issue with many effects.

Swiftb0y:
great. so then keep the gain knob as is?

daschuer:
If you insist. I still consider that ugly and prefer the solution we did for the idolators, to adopt the same pattern everywhere and than we can fix the missing unit issue in one go in a later PR. 
Using ration is faster as a side effect.

daschuer:
The maximum number is what you like. 3 dB is ~1.4.  
Like here: 
https://github.com/mixxxdj/mixxx/blob/9381a84aa63b791f8bf0f74f49ff44ff9685f153/src/effects/backends/builtin/equalizer_util.h#L43

Swiftb0y:
Right, but that means that the entire $[0, 1]$ range is on the left 50% of the pot and $[1, 1.4]$ on the right. So the pot is not continuous ("stetig" in german) anymore (in the mathematical sense). Thats not pretty either...

daschuer (approved):
Since this is already an improvement let's merge it and continue iterating from here.

PR #13655, opened 2024-09-15
fix: exclude stem samples for QML waveform

acolombier:
Fixes the same issue than #13472 on QML

### DIFF
--- a/src/qml/qmlplayerproxy.cpp
+++ b/src/qml/qmlplayerproxy.cpp
@@ -215,8 +215,26 @@ void QmlPlayerProxy::slotWaveformChanged() {
     }
     const int textureWidth = pWaveform->getTextureStride();
     const int textureHeight = pWaveform->getTextureSize() / pWaveform->getTextureStride();
-    const uchar* data = reinterpret_cast<const uchar*>(pWaveform->data());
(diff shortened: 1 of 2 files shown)


src/qml/qmlplayerproxy.cpp
```diff
@@ -215,8 +215,26 @@ void QmlPlayerProxy::slotWaveformChanged() {
...
+
+    const WaveformData* data = pWaveform->data();
+    // Make a copy of the waveform data, stripping the stems portion. Note that the datasize is
+    // different from the texture size -- we want the full texture size so the upload works. See
+    // m_data in waveform/waveform.h.
+    if (m_waveformData.size() == 0 ||
+            static_cast<int>(m_waveformData.size()) !=
+                    pWaveform->getTextureSize()) {
+        m_waveformData.resize(pWaveform->getTextureSize());
+    }
```
Swiftb0y, line 227: Why not call resize unconditionally? I don't see why this `if` is needed.

acolombier: I just copied the patch that was merge for the legacy UI. Make sense to remove it tho!

Swiftb0y: yeah. I'm not even sure how safe the condition is technically.

Swiftb0y (approved):
LGTM. Thank you

PR #13656, opened 2024-09-15
fix: allow controller setting integer to have a negative range

acolombier:


src/controllers/legacycontrollersettings.h
```diff
@@ -259,11 +259,16 @@ class LegacyControllerNumberSetting
...
     /// @return true if valid
     bool valid() const override {
+        static constexpr bool typeSize = sizeof(long) == sizeof(int);
         return AbstractLegacyControllerSetting::valid() &&
                 m_defaultValue >= m_minValue && m_savedValue >= m_minValue &&
                 m_editedValue >= m_minValue && m_defaultValue <= m_maxValue &&
                 m_savedValue <= m_maxValue && m_editedValue <= m_maxValue &&
-                m_stepValue > 0 && m_stepValue < m_maxValue;
+                m_stepValue > 0 &&
+                (typeSize ||
```
acolombier, line 268: To be confirmed, but I read that some platform use the same byte size for `int` and `long` and since the default min and max uses the int limits, without a larger type, this would lead to an overflow.

Swiftb0y:
What is the usecase for this? If I understand `LegacyControllerNumberSetting` correctly, its just modeling a [closed integer interval](https://en.wikipedia.org/wiki/Interval_(mathematics)#Integer_intervals) $[ a \ldotp \ldotp b ]$ (where $a \leq b$), right? What is the purpose of allowing $a \ge b$ too?

acolombier:
This doesn't allow a `a > b`, it allows the step to be less than the max. My usecase was the following: I have a setting that can be in the range `[-250; 0]` and the step is `1`. Currently, this would fall. With this changes, instead of doing making sure the step is less than the max, it makes sure the the step is less than the range side (250 in my case)

src/controllers/legacycontrollersettings.h
daschuer: In case of an int a step value can't be bigger than the half of the range. If we assume 8 bit, we may have 
-128 ... 127 and the maximum Step that can be expressed is 127. 

So you may check: 
m_minValue <= 0 && m_minValue + m_stepValue <= m_maxValue &&
m_maxValue >= 0 && m_maxValue - m_stepValue >= sm_minValue

daschuer (changes requested):

src/controllers/legacycontrollersettings.h
daschuer: And m_maxValue - m_stepValue < m_maxValue for the unsigned case ...

Swiftb0y:
> This doesn't allow a `a > b`, it allows the step to be less than the max. My usecase was the following: I have a setting that can be in the range `[-250; 0]` and the step is `1`. Currently, this would fall. With this changes, instead of doing making sure the step is less than the max, it makes sure the the step is less than the range side (250 in my case)

ah, yeah. That makes total sense. Thanks for the explanation.

src/controllers/legacycontrollersettings.h
Swiftb0y: and put it in a utility function please. The long condition chain is already not super easy to follow.

acolombier: > And m_maxValue - m_stepValue < m_maxValue for the unsigned case ...

Not sure which usecase you are referring to?

Swiftb0y (changes requested):

src/controllers/legacycontrollersettings.h
```diff
@@ -6,6 +6,23 @@
...
+template<class T>
+bool valid_range(T min, T max, T step) {
+    // Detecting overflow
+    VERIFY_OR_DEBUG_ASSERT(max + step > max) {
+        return false;
+    }
+    // Detecting overflow
+    VERIFY_OR_DEBUG_ASSERT(min - step < min) {
+        return false;
+    }
```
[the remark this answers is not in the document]
acolombier: Would you suggest to simply ignore overflow? Or do you have an alternative suggestion?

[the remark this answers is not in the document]
daschuer: I have not understand the purpose of the assertion. 
An overflow aka undefined behavior can happen in this case, but is not problem. 
INT_MIN - 1 must never be done.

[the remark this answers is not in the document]
Swiftb0y: > Would you suggest to simply ignore overflow? Or do you have an alternative suggestion?

Ensure that over-/underflow can not happen. So as already pointed out ensure there is no `min - x` (where $x > 0$). So with the condition you have written, you may `VERIFY_OR_DEBUG_ASSERT(step >= 0)`.

src/controllers/legacycontrollersettings.h
```diff
@@ -6,6 +6,14 @@
 #include "controllers/legacycontrollersettingslayout.h"
 #include "util/parented_ptr.h"
 
+namespace {
+template<class T>
+bool valid_range(T min, T max, T step) {
+    return (min <= 0 && min + step <= max) ||
+            (max >= 0 && max - step >= min);
```
Swiftb0y, line 13:
```suggestion
    VERIFY_OR_DEBUG_ASSERT(step >= 0) {
        return false;
    }
    return (min <= 0 && min + step <= max) ||
            (max >= 0 && max - step >= min);
```

Swiftb0y (changes requested):
lgtm otherwise.

src/controllers/legacycontrollersettings.h
daschuer: Alternative would be moving the test "step > 0" into this range check as well.

src/controllers/legacycontrollersettings.h
```diff
@@ -263,7 +271,7 @@ class LegacyControllerNumberSetting
                 m_defaultValue >= m_minValue && m_savedValue >= m_minValue &&
                 m_editedValue >= m_minValue && m_defaultValue <= m_maxValue &&
                 m_savedValue <= m_maxValue && m_editedValue <= m_maxValue &&
-                m_stepValue > 0 && m_stepValue < m_maxValue;
+                m_stepValue > 0 && valid_range(m_minValue, m_maxValue, m_stepValue);
```
daschuer, line 274: We may face the condition that the default "default" = 0 in case of a missing attribute is not valid. Do we want to improve that as well for the negative range?

acolombier: Added the usecase at the de-serialisation step.

src/controllers/legacycontrollersettings.h
```diff
@@ -6,6 +6,16 @@
 #include "controllers/legacycontrollersettingslayout.h"
 #include "util/parented_ptr.h"
 
+namespace {
+template<class T>
+bool valid_range(T min, T max, T step) {
+    VERIFY_OR_DEBUG_ASSERT(step >= 0) {
+        return false;
+    }
```
Swiftb0y, line 14: do we actually ensure this somewhere else (and provide appropriate debug feedback)? I'm not particularly familiar with the code.

acolombier: Yes and no. If a setting definition is invalid, you will get a warning log message telling you about it, no actual UI feedback on the matter.

At the time of implementation, the assumption was made that mapping developer would be able to gather log if they needed to

daschuer (approved):
LGTM, thank you.

src/controllers/legacycontrollersettings.h
acolombier: No, it just says it will ignore a definition due to an invalid settings

Swiftb0y: thats unfortunate. we should fix that in the future. Just saying something is invalid gets increasingly frustrating the more complex the definition is. how would I know which of my (hypothetical) 10 parameters are invalid? Out-of-scope here though. Just  suggestion.

acolombier: You would know which settings is invalid, but it wouldn't details what's wrong with it (e.g Default out of range? Min after max?)

The warning for that is [here](https://github.com/mixxxdj/mixxx/blob/cb9a5c20cdd962e095ec409751486443aa05af69/src/controllers/legacycontrollermappingfilehandler.cpp#L299)

Let's say you have 10 setitings (`my setting1`..`mySetting10`), and `mySetting6` is invalid, you would get something like
```
Warning: The parsed setting in file "/home/me/myfile.hid/xml" at line 42 appears to be invalid. It will be ignored.
```

Swiftb0y: yeah, thats not great once a setting has a bunch of parameters (along with non-obvious constraints). Ideally the parser should be able to report validation failures as they happen.

acolombier: I might look into that for the future. Happy to get this bugfix in the meantime?

Swiftb0y: yes.

Swiftb0y (approved):
LGTM. Thank you.

PR #13657, opened 2024-09-15
fix: prevent crash when no app windows have the focus

acolombier:
This crash occurs when using the `GoToItem` while a window is open in Mixxx (eg. settings) and was last to have focus, but the Mixxx application doesn't have the focus anymore


src/library/librarycontrol.cpp
```diff
@@ -1039,8 +1039,11 @@ void LibraryControl::slotGoToItem(double v) {
         // press & release Space (QAbstractButton::clicked() is emitted on release)
         QKeyEvent pressSpace = QKeyEvent{QEvent::KeyPress, Qt::Key_Space, Qt::NoModifier};
         QKeyEvent releaseSpace = QKeyEvent{QEvent::KeyRelease, Qt::Key_Space, Qt::NoModifier};
-        QApplication::sendEvent(QApplication::focusWindow(), &pressSpace);
-        QApplication::sendEvent(QApplication::focusWindow(), &releaseSpace);
+        auto window = QApplication::focusWindow();
```
Swiftb0y, line 1042: same below
```suggestion
        auto* window = QApplication::focusWindow();
```

Swiftb0y (commented):
LGTM otherwise

Swiftb0y (approved):
LGTM. Thank you.

PR #13658, opened 2024-09-15
chore: add note about ConfigKey naming convention

acolombier:
Let me know if you can see anywhere where it would be worth pointing out as well.

### DIFF
--- a/src/preferences/configobject.h
+++ b/src/preferences/configobject.h
@@ -15,6 +15,8 @@
 
 // Class for the key for a specific configuration element. A key consists of a
 // group and an item.
+//
+// NOTE: new ConfigKey's item should use the `snake_case` formatting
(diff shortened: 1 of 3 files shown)


Swiftb0y (commented):
sorry I forgot to submit the review earlier. anyways, not super important just wanted to mention it.

JoergAtGithub:
This way it appears for the class 
grafik
To appear befor the constructor, it must be place immediately before the constructor

Swiftb0y:
yeah thats what I meant. I don't know which usage is more common so its more likely to be seen.

JoergAtGithub:
If you want it in Intellisense, only the place beforr thr constructor is helpful, because this is where you enter the control name.

Swiftb0y:
then lets add it there too?

PR #13669, opened 2024-09-19
feat: add file and color controller setting types

acolombier:
Add new controller setting type, particularly useful for screen mapping.

https://github.com/user-attachments/assets/af1e2df9-e273-4471-ac39-535bfbc4fbdd

TODO:
- [x] Unit test
- [x] Improved UI

### DIFF
--- a/src/controllers/legacycontrollersettings.cpp
+++ b/src/controllers/legacycontrollersettings.cpp
@@ -256,8 +256,8 @@ void LegacyControllerEnumSetting::parse(const QString& in, bool* ok) {
     save();
 
     size_t pos = 0;
-    for (const auto& value : std::as_const(m_options)) {
-        if (std::get<0>(value) == in) {
+    for (const auto& item : std::as_const(m_options)) {
+        if (item.value == in) {
             if (ok != nullptr) {
                 *ok = true;
             }
@@ -272,8 +272,15 @@ void LegacyControllerEnumSetting::parse(const QString& in, bool* ok) {
 QWidget* LegacyControllerEnumSetting::buildInputWidget(QWidget* pParent) {
     auto* pComboBox = new QComboBox(pParent);
 
-    for (const auto& value : std::as_const(m_options)) {
-        pComboBox->addItem(std::get<1>(value));
+    for (const auto& item : std::as_const(m_options)) {
+        if (item.color.isValid()) {
+            QPixmap icon(24, 24);
+            QPainter painter(&icon);
+            painter.fillRect(0, 0, 24, 24, item.color);
+            pComboBox->addItem(QIcon(icon), item.label);
+        } else {
+            pComboBox->addItem(item.label);
+        }
     }
     pComboBox->setCurrentIndex(static_cast<int>(m_editedValue));
 
--- a/src/controllers/legacycontrollersettings.cpp
+++ b/src/controllers/legacycontrollersettings.cpp
@@ -2,12 +2,19 @@
 
 #include <util/assert.h>
 
+#include <QBoxLayout>
 #include <QCheckBox>
+#include <QColorDialog>
 #include <QComboBox>
 #include <QDoubleSpinBox>
+#include <QFileDialog>
 #include <QLabel>
 #include <QLayout>
+#include <QPainter>
+#include <QPixmap>
+#include <QPushButton>
 #include <QSpinBox>
+#include <QStringLiteral>
 
 #include "moc_legacycontrollersettings.cpp"
 
@@ -54,6 +61,8 @@ LegacyControllerSettingBuilder::LegacyControllerSettingBuilder() {
     registerType<LegacyControllerIntegerSetting>();
     registerType<LegacyControllerRealSetting>();
     registerType<LegacyControllerEnumSetting>();
+    registerType<LegacyControllerColorSetting>();
+    registerType<LegacyControllerFileSetting>();
 }
 
 AbstractLegacyControllerSetting::AbstractLegacyControllerSetting(const QDomElement& element) {
@@ -221,7 +230,15 @@ LegacyControllerEnumSetting::LegacyControllerEnumSetting(
             !value.isNull();
             value = value.nextSiblingElement("value")) {
         QString val = value.text();
-        m_options.append(std::tuple<QString, QString>(val, value.attribute("label", val)));
+        QColor color = QColor(value.attribute("color"));
+        // TODO: Remove once we mandate GCC 10/Clang 16
+#if defined(__cpp_aggregate_paren_init) &&       \
+        __cpp_aggregate_paren_init >= 201902L && \
+        !defined(_MSC_VER) // FIXME: Bug in MSVC preventing the use of this feature
+        m_options.emplace_back(val, value.attribute("label", val), color);
+#else
+        m_options.emplace_back(Item{val, value.attribute("label", val), color});
+#endif
         if (value.hasAttribute("default")) {
             m_defaultValue = pos;
         }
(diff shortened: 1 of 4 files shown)


JoergAtGithub:
These setting types are great features!

For the file selector I would expect a widget like this:
grafik
and a XML settings for the allowed file types.

More complicated might be a generic definiton for colors, as:

- we have already certain purpose specific colormaps:
grafik
- but more important here, many controllers have only a small list of supported colors, as this list of named colors: https://github.com/mixxxdj/mixxx/blob/067792b1e7e5cb524a91df6b2e83533f362c5bf1/res/controllers/Traktor-Kontrol-S3-hid-scripts.js#L149-L171
 - our current color mapper API has no link to the XML yet: https://github.com/mixxxdj/mixxx/blob/main/res/controllers/color-mapper-api.d.ts
 - Maybe we need a solution to define a color map from XML instead of JavaScript?

acolombier:
Thanks for the early feedback!

> For the file selector I would expect a widget like this:

Yes, that makes sense. It's in the todo list already :) 

> we have already certain purpose specific colormaps:

The current widget are more tailored for palette AFAIK, not unique colour selection, so not sure how to do this? I was thinking of adding a color indication inside/next to the pick button

> but more important here, many controllers have only a small list of supported colors, as this list of named colors:

Yes, this usecase is already covered with enum, and actually [used by the S4Mk3](https://github.com/mixxxdj/mixxx/blob/main/res/controllers/Traktor%20Kontrol%20S4%20MK3.hid.xml#L15). One bits of improvement could be to add some color indication in the dropdown, I guess we could easily add an attribute on the `option` element if we wanted to improve the UX for those?

Swiftb0y:
I think using the Qt color picker is fine. If we want to we can pre-fill their color with for example the users currently track-color palette (the palette source could also be configured as part of the setting). But lets KISS for now until we have a good reason to couple this to our palettes. A palette selector could be interesting indeed, but that is separate from the color picker.

acolombier:
I have reworked slightly the widget, let me know what you think of it.

image
image
image
image

(Using the vertical orientation and the horizontal one)

acolombier:
As suggested by @JoergAtGithub , I've also added the ability to add a color icon next to a dropdown menu, which may be used for color selection.

image
image

src/controllers/legacycontrollersettings.cpp
```diff
@@ -230,7 +230,8 @@ LegacyControllerEnumSetting::LegacyControllerEnumSetting(
             !value.isNull();
             value = value.nextSiblingElement("value")) {
         QString val = value.text();
-        m_options.append(std::tuple<QString, QString>(val, value.attribute("label", val)));
+        QColor color = QColor(value.attribute("color"));
+        m_options.append(Item{val, value.attribute("label", val), color});
```
Swiftb0y, line 234:
```suggestion
        m_options.emplace_back(val, value.attribute("label", val), color);
```

src/controllers/legacycontrollersettings.cpp
```diff
@@ -274,3 +291,143 @@ QWidget* LegacyControllerEnumSetting::buildInputWidget(QWidget* pParent) {
 
     return pComboBox;
 }
+
+LegacyControllerColorSetting::LegacyControllerColorSetting(
+        const QDomElement& element)
+        : AbstractLegacyControllerSetting(element) {
+    m_defaultValue = QColor(element.attribute("default"));
+    m_savedValue = m_defaultValue;
+    m_editedValue = m_defaultValue;
```
Swiftb0y, line 300: prefer ctor initializer list. they're evaluated in order and you can refer between members. So this works without issues I'm pretty sure.
```suggestion
        const QDomElement& element)
        : AbstractLegacyControllerSetting(element), m_defaultValue(QColor(element.attribute("default")), m_savedValue(m_defaultValue), m_editedValue(m_defaultValue) {
```

src/controllers/legacycontrollersettings.cpp
```diff
@@ -274,3 +291,143 @@ QWidget* LegacyControllerEnumSetting::buildInputWidget(QWidget* pParent) {
...
+    }
+    reset();
+    save();
+
+    m_savedValue = QColor(in);
+    if (!m_editedValue.isValid()) {
+        *ok = false;
+        return;
+    }
+    m_editedValue = m_savedValue;
```
Swiftb0y, line 324: this doesn't work properly. you didn't check for `nullptr` on the second access to `*ok` and you never actually set it `true`, potentially leaving it uninitialized (which should never be the case for an out pointer like this).

src/controllers/legacycontrollersettings.cpp
```diff
@@ -274,3 +291,143 @@ QWidget* LegacyControllerEnumSetting::buildInputWidget(QWidget* pParent) {
...
+    return pPushButton;
+}
+
+LegacyControllerFileSetting::LegacyControllerFileSetting(
+        const QDomElement& element)
+        : AbstractLegacyControllerSetting(element) {
+    m_defaultValue = QFileInfo(element.attribute("default"));
+    m_fileFilter = element.attribute("pattern");
+    m_savedValue = m_defaultValue;
+    m_editedValue = m_defaultValue;
```
Swiftb0y, line 370: same here. prefer ctor init list.

src/controllers/legacycontrollersettings.cpp
```diff
@@ -274,3 +291,143 @@ QWidget* LegacyControllerEnumSetting::buildInputWidget(QWidget* pParent) {
...
+    }
+    reset();
+    save();
+
+    m_editedValue = QFileInfo(in);
+    if (!m_editedValue.exists()) {
+        *ok = false;
+        return;
+    }
+    m_savedValue = m_editedValue;
```
Swiftb0y, line 394: same here.

src/controllers/legacycontrollersettings.cpp
```diff
@@ -274,3 +291,143 @@ QWidget* LegacyControllerEnumSetting::buildInputWidget(QWidget* pParent) {
...
+}
+
+QWidget* LegacyControllerFileSetting::buildInputWidget(QWidget* pParent) {
+    auto* pWidget = new QWidget(pParent);
+    pWidget->setLayout(new QHBoxLayout);
+    auto* pPushButton = new QPushButton(tr("Browse..."), pWidget);
+    auto* pLabel = new QLabel(QStringLiteral("<i>%1</i>").arg(tr("No file selected")), pWidget);
```
Swiftb0y, line 401: wdyt of this to make sure the format string doesn't get out of sync?
```suggestion
    auto* pLabel = new QLabel(QStringLiteral("<i>%1</i>").arg(tr("No file selected")), pWidget);
    auto setLabelText = [pLabel](QString&& text) {
        pLabel->setText(QStringLiteral("<i>%1</i>").arg(text));
    };
```

src/controllers/legacycontrollersettings.h
```diff
@@ -423,6 +431,141 @@ class LegacyControllerEnumSetting
     friend class ControllerS4MK3SettingTest_ensureLibrarySettingValueAndEnumEquals;
 };
 
+class LegacyControllerColorSetting
+        : public LegacyControllerSettingFactory<LegacyControllerColorSetting>,
+          public AbstractLegacyControllerSetting {
+  public:
+    LegacyControllerColorSetting(const QDomElement& element);
+
+    virtual ~LegacyControllerColorSetting() = default;
```
Swiftb0y, line 440: consider putting the `= default;` into the implementation. Since there call needs to be virtual, there is no advantage to it being in the header (because it can't be inlined).
Also use `override` when overriding instead of something that looks like a redeclaration.
```suggestion
    ~LegacyControllerColorSetting() override;
```

src/controllers/legacycontrollersettings.h
```diff
@@ -423,6 +431,141 @@ class LegacyControllerEnumSetting
...
+    virtual QWidget* buildInputWidget(QWidget* parent) override;
+
+  private:
+    QColor m_savedValue;
+    QColor m_defaultValue;
+
+    QColor m_editedValue;
+
+    friend class LegacyControllerMappingSettingsTest_enumSettingEditing_Test;
+    friend class ControllerS4MK3SettingTest_ensureLibrarySettingValueAndEnumEquals;
```
Swiftb0y, line 498: I'd prefer `FRIEND_TEST` from `<gtest/gtest_prod.h>`.

src/controllers/legacycontrollersettings.h
```diff
@@ -423,6 +431,141 @@ class LegacyControllerEnumSetting
...
+    bool valid() const override {
+        return AbstractLegacyControllerSetting::valid() &&
+                m_defaultValue.isValid() &&
+                m_savedValue.isValid();
+    }
+
+    static AbstractLegacyControllerSetting* createFrom(const QDomElement& element) {
+        return new LegacyControllerColorSetting(element);
+    }
+    static inline bool match(const QDomElement& element);
```
Swiftb0y, line 478: `inline`? doesn't make sense afaict. If you want to optimize the call away, you'll have to move the implementation here.

src/controllers/legacycontrollersettings.cpp
acolombier: Looks like I still need to the `stuct` type

acolombier:
All cleaned up now!

src/controllers/legacycontrollersettings.h
```diff
@@ -413,14 +423,158 @@ class LegacyControllerEnumSetting
...
+  protected:
+    LegacyControllerFileSetting(const QDomElement& element,
+            const QFileInfo& currentValue,
+            const QFileInfo& defaultValue)
+            : AbstractLegacyControllerSetting(element),
+              m_defaultValue(defaultValue),
+              m_savedValue(currentValue) {
+    }
+
+    virtual QWidget* buildInputWidget(QWidget* parent) override;
```
Swiftb0y, line 567: already implied by `override`. at least my clangd-based linter complains about this.
```suggestion
    QWidget* buildInputWidget(QWidget* parent) override;
```

src/controllers/legacycontrollersettings.h
```diff
@@ -413,14 +423,158 @@ class LegacyControllerEnumSetting
...
+    /// @return true if valid
+    bool valid() const override {
+        return AbstractLegacyControllerSetting::valid() &&
+                (m_defaultValue == m_savedValue || m_savedValue.exists());
+    }
+
+    static AbstractLegacyControllerSetting* createFrom(const QDomElement& element) {
+        return new LegacyControllerFileSetting(element);
+    }
+    static inline bool match(const QDomElement& element) {
```
Swiftb0y, line 551: definitions in the header are always implicitly `inline` (only in C++ though).
```suggestion
    static bool match(const QDomElement& element) {
```

src/controllers/legacycontrollersettings.h
```diff
@@ -413,14 +423,158 @@ class LegacyControllerEnumSetting
...
+    FRIEND_TEST(LegacyControllerMappingSettingsTest, enumSettingEditing);
+};
+
+class LegacyControllerFileSetting
+        : public LegacyControllerSettingFactory<LegacyControllerFileSetting>,
+          public AbstractLegacyControllerSetting {
+  public:
+    LegacyControllerFileSetting(const QDomElement& element);
+
+    virtual ~LegacyControllerFileSetting() = default;
```
Swiftb0y, line 514: consider moving to `impl`. No perf benefit from the definition being in the header. You already did it for the other class too.
```suggestion
    virtual ~LegacyControllerFileSetting();
```

src/controllers/legacycontrollersettings.h
```diff
@@ -180,7 +184,7 @@ class LegacyControllerBooleanSetting
     bool m_defaultValue;
     bool m_editedValue;
 
-    friend class LegacyControllerMappingSettingsTest_booleanSettingEditing_Test;
+    FRIEND_TEST(LegacyControllerMappingSettingsTest, booleanSettingEditing);
```
Swiftb0y, line 187: :+1:

src/controllers/legacycontrollersettings.h
```diff
@@ -413,14 +423,158 @@ class LegacyControllerEnumSetting
...
+                (m_defaultValue == m_savedValue || m_savedValue.exists());
+    }
+
+    static AbstractLegacyControllerSetting* createFrom(const QDomElement& element) {
+        return new LegacyControllerFileSetting(element);
+    }
+    static inline bool match(const QDomElement& element) {
+        return element.hasAttribute("type") &&
+                QString::compare(element.attribute("type"),
+                        "file",
```
Swiftb0y, line 554: these all take `QString`s. so wrap in `QStringLiteral` please. Same above.

src/controllers/legacycontrollersettings.cpp
```diff
@@ -221,7 +230,8 @@ LegacyControllerEnumSetting::LegacyControllerEnumSetting(
             !value.isNull();
             value = value.nextSiblingElement("value")) {
         QString val = value.text();
-        m_options.append(std::tuple<QString, QString>(val, value.attribute("label", val)));
+        QColor color = QColor(value.attribute("color"));
+        m_options.emplace_back(Item{val, value.attribute("label", val), color});
```
Swiftb0y, line 234: `emplace_back` constructs the target type from the arguments in place, thus pass the args directly:
```suggestion
        QColor color = QColor(value.attribute("color"));
        m_options.emplace_back(val, value.attribute("label", val), color);
```
otherwise you'll construct a temporary and the move from it instead of a single ctor call.

Swiftb0y (commented):
couple more comments. this is getting pretty close to the finishing line though.

Swiftb0y:
The main issue right now is that there is a lot of repetitive boilerplate needed in the widgets. I'd really like to see a future PR that cleans that up instead of having to repeat the same logic over and over again.

Swiftb0y:
well, for example each type essentially has a `m_savedValue`, `m_defaultValue`, `m_editedValue` and reimplements the matching logic `save()`, `reset()`, `isDirty()`, `isDefault()` again and again. This makes it easy to create inconsistent behavior between different widgets, makes it more complicated to implement new widgets and obstructs the parts which actually are unique between these widgets/types. I feel like this could be solved fairly easily by creating a base class template. Then you could have `class FilePickerWidget : public SingleValuedBase<QFileInfo>` and that class gives you the basic members you need. You can still do something more costum by not inheriting from that base class, but at least know you don't need to copy and paste the common boilerplate.

Swiftb0y:
As long as we make sure we don't go overboard with this and create a 5 level deep inheritance hierarchy this should probably a reasonably okay use of inheritance.

acolombier:
Makes sense - I will look into that next!

src/controllers/legacycontrollersettings.cpp
acolombier: This doesn't work as there is no ctor on the struct. Do you want me to add one? Or replace with a `push_back`?

Swiftb0y: [It does on compiler explorer](https://compiler-explorer.com/z/415acfGxe). I guess use push_back then, but I'd still be interested in the error message...

Swiftb0y: What compiler version do you have?
image

image

acolombier: clang shipped with Ubuntu 22.04, which is 15 AFAIK. Is there any preprocessor I should put to support it going forward?

acolombier: Appreciate I could make it work, but I unfortunately also use my machine and toolchain for my full-time job, as well as other projects than Mixxx and I don't think I will take the risk to ~break~update it right now. Appreciate `main` runs on 24.04 as per the consensus, so completely okay with getting code to to compile and run stable there, but if I was to continue to contribute to Mixxx, I would need to get the code to compile on my setup too.

Are you happy to go with something like this?

```cpp
#if ((__GNUC__ && !__clang_major__ && __GNUC__ < 10) || (__clang_major__ && __clang_major__ < 16))
    myvec.emplace_back(Foo{1, 2});
#else
    myvec.emplace_back(1, 2);
#endif
```

Just want to put the emphasis on the fact that I'm not discussing the consensus on C++20 and appreciate other contributions will break Clang 14 eventually (that's the one on Ubuntu 22.04 btw).

acolombier: Oh lovely! Didn't know C++ features had macro directly available.

Swiftb0y: yeah, always worth looking the original paper for a proposal ;)

acolombier: Looks like this might not work with MSVC... any suggestion?

Swiftb0y: > C:\buildenv\mixxx-deps-2.5-x64-windows-release-c616147\installed\x64-windows-release\include\Qt6\QtCore\qarraydataops.h(570) : fatal error C1001: Internal compiler error.
Looks like a compiler error...  I'll have a look.

acolombier:
@Swiftb0y is [that](https://github.com/acolombier/mixxx/compare/feat/controller-setting-file-and-color...acolombier:mixxx:feat/refactor-controller-setting) what you have in mind for the refactor?

Swiftb0y:
basically yes. Wdyt about it?

src/controllers/legacycontrollersettings.cpp
Swiftb0y: Actually, MSVC on compiler explorer [doesn't support external libraries](https://github.com/compiler-explorer/compiler-explorer/issues/6681), so I don't really have an easy way to debug this. I guess we'll have to work around it. Lets just add `&& !defined(_MSC_VER)` so msvc uses the `push_back` path.

acolombier:
Yeah, that looks good to me, since it does remove boilerplate and duplicated logic. Is this still okay to have this as a separate PR?

Swiftb0y:
LGTM as soon as we fixed the MSVC issue

Swiftb0y:
LGTM. Please squash the fixups

Swiftb0y (approved):

PR #13671, opened 2024-09-20
fix: sync rate using the current BPM instead of the file one

### LINKED ISSUE #12738: BPM mutliplier should be using the current deck BPM rather than the file BPM
### Bug Description

# How to reproduce

Here is the usecase:
- Deck A has a track with a BPM of 80
- Deck B has a track with a BPM of 128

- Adjust the BPM of deck A and B to 110 (using tempo fader)
- Enable sync using `sync_enable` or `beatsync_tempo`

## Expected 

The BPM multiplier should be `1`, and both track should keep the exact same BPM value (110 in the example)

## Actual

The non-sync master (of the channel calling `beatsync_tempo`)  BPM of the non deck gets set to 55 (if it's A) or 220 (if it's B)

# Fix options

1. In [synccontrol.cpp](https://github.com/mixxxdj/mixxx/blob/85d3d96f25e112e4b76b71f30d517b87bfe809d2/src/engine/sync/synccontrol.cpp), `SyncControl::slotControlBeatSyncTempo` and `SyncControl::reinitLeaderParams` could use `getBpm()` instead of `fileBpm()` if valid
2. Add new CO to ignore the multiplier, e.g `beatsync_tempo_absolute` and `absolute_sync_enable`

While `1.` feels like a better fix, I guess it may also be a breaking change in some cases so I thought I will mention `2.` although it definitely feels sub-optimal 

### Version

2.4.0

### OS

All

acolombier:
Fixes #12738

It doesn't make much sense to use the original file BPM as it is irrelevant to the user. In case it is, they could always reset the fader to zero before syncing, in which case sync behaviour will have the same effect as today

### DIFF
--- a/src/engine/sync/synccontrol.cpp
+++ b/src/engine/sync/synccontrol.cpp
@@ -278,7 +278,7 @@ void SyncControl::reinitLeaderParams(
         kLogger.trace() << "SyncControl::reinitLeaderParams" << getGroup()
                         << beatDistance << baseBpm << bpm;
     }
-    m_leaderBpmAdjustFactor = determineBpmMultiplier(fileBpm(), baseBpm);
+    m_leaderBpmAdjustFactor = determineBpmMultiplier(mixxx::Bpm(m_pBpm->get()), bpm);
     updateLeaderBpm(bpm);
     updateLeaderBeatDistance(beatDistance);
 }
--- a/src/engine/sync/synccontrol.cpp
+++ b/src/engine/sync/synccontrol.cpp
@@ -284,7 +284,7 @@ void SyncControl::reinitLeaderParams(
 }
 
 double SyncControl::determineBpmMultiplier(mixxx::Bpm myBpm, mixxx::Bpm targetBpm) const {
-    if (!myBpm.isValid() || !targetBpm.isValid()) {
+    if (!myBpm.isValid() || !targetBpm.isValid() || myBpm.isNull() || targetBpm.isNull()) {
         return kBpmUnity;
     }
     double unityRatio = myBpm / targetBpm;
--- a/src/engine/sync/synccontrol.cpp
+++ b/src/engine/sync/synccontrol.cpp
@@ -284,7 +284,7 @@ void SyncControl::reinitLeaderParams(
 }
 
 double SyncControl::determineBpmMultiplier(mixxx::Bpm myBpm, mixxx::Bpm targetBpm) const {
-    if (!myBpm.isValid() || !targetBpm.isValid() || myBpm.isNull() || targetBpm.isNull()) {
+    if (!myBpm.isValid() || !targetBpm.isValid()) {
         return kBpmUnity;
     }
     double unityRatio = myBpm / targetBpm;
--- a/src/track/bpm.h
+++ b/src/track/bpm.h
@@ -45,6 +45,10 @@ public:
         return isValidValue(m_value);
     }
 
+    bool isNull() const {
+        return m_value == kValueMin;
+    }
+
     /// Return the valid value.
     ///
     /// Triggers a debug assertion if the value is invalid
--- a/src/track/bpm.h
+++ b/src/track/bpm.h
@@ -45,10 +45,6 @@ public:
         return isValidValue(m_value);
     }
 
-    bool isNull() const {
-        return m_value == kValueMin;
-    }
-
     /// Return the valid value.
     ///
     /// Triggers a debug assertion if the value is invalid
--- a/src/test/enginesynctest.cpp
+++ b/src/test/enginesynctest.cpp
@@ -3145,3 +3145,37 @@ TEST_F(EngineSyncTest, KeepCorrectFactorUponResync) {
             ControlObject::get(ConfigKey(m_sGroup2, "rate")),
             0.005);
 }
+
+// There is a race condition preventing this usecase to work.
+// https://github.com/mixxxdj/mixxx/issues/????
+TEST_F(EngineSyncTest, DISABLED_KeepCorrectFactorOnLoad) {
+    /* Usecase
+        - load track @ 174bpm in deck 1 and enable sync leader
+        - enable sync follower on the empty second track
+        - load track @ 87 bpm in deck 2
+        - deck 2 should remain at 87 BPM and rate of 1.0
+    */
+    m_pMixerDeck1->loadFakeTrack(false, 174.0);
+    ProcessBuffer();
+
+    EXPECT_DOUBLE_EQ(174.0, ControlObject::get(ConfigKey(m_sGroup1, "bpm")));
+    EXPECT_DOUBLE_EQ(0.0, ControlObject::get(ConfigKey(m_sGroup2, "bpm")));
+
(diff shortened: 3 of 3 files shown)


acolombier:
@Core Team this is potentially a breaking change, although IMHO this improves the existing behaviour. Let me know what you think, I'd be happy to put that behaviour change behind a setting.

ronso0:
I agree this is a fix.

I hardly ever use sync so I can't really tell whether this would break use cases. One could be too going wild and play a few chipmunk'ed tracks in a row (eg. with 90% rate range).

Either way, if we decide this significantly breaks use cases it should go to 2.5, otherwise it should go to 2.4, no?

acolombier:
> Either way, if we decide this significantly breaks use cases it should go to 2.5, otherwise it should go to 2.4, no?

I was even tempted to put it in 2.6 since we are already in the beta process of 2.5. This way, if we do break some user's routines, this give use more lead way to gather the feedback. 

Happy to release in previous version otherwise!

src/engine/sync/synccontrol.cpp
```diff
@@ -278,13 +278,13 @@ void SyncControl::reinitLeaderParams(
...
     }
-    m_leaderBpmAdjustFactor = determineBpmMultiplier(fileBpm(), baseBpm);
+    m_leaderBpmAdjustFactor = determineBpmMultiplier(mixxx::Bpm(m_pBpm->get()), bpm);
     updateLeaderBpm(bpm);
     updateLeaderBeatDistance(beatDistance);
 }
 
 double SyncControl::determineBpmMultiplier(mixxx::Bpm myBpm, mixxx::Bpm targetBpm) const {
-    if (!myBpm.isValid() || !targetBpm.isValid()) {
+    if (!myBpm.isValid() || !targetBpm.isValid() || myBpm.isNull() || targetBpm.isNull()) {
```
ywwg, line 287: is null ever a valid bpm? seems like isValid could check for null value rather than having a separate function.

ywwg (commented):
The intent of the original behavior was to handle cases where drum and bass tracks were detected at 1/2 the correct bpm. (that's what happened to Grimes at her [famously disastrous Coachella set](https://www.reddit.com/r/Techno/comments/1c4lla3/a_few_thoughts_on_the_grimes_coachella_fiasco/)).  In that situation, the file bpms are a better hint as to whether the tracks themselves, at native rate, are 1/2 or 2x from each other.

As long as the beatmatch behavior still works for the 70/140 (dubstep) or 87/174 (drum and bass) scenario at reasonable rate slider settings (+/- 8), then this fix is fine.  Can you do some quick checking to see if that's the case?

acolombier:
> As long as the beatmatch behavior still works for the 70/140 (dubstep) or 87/174 (drum and bass) scenario at reasonable rate slider settings (+/- 8), then this fix is fine. Can you do some quick checking to see if that's the case?

Yes, this was part of my testing already, and it worked fine on my end. But it would be very appreciated if someone could also try, just to make sure I didn't have a "lucky" case

src/engine/sync/synccontrol.cpp
acolombier: As part of my routine, I sometime makes breaks where I slide the tempo fader all the way to zeo (with a 100% range), meaning the final bpm will be `0.00`, on potentially the leader deck. Technically, `0.00` is a valid BPM, thus why the `isNull` introduction

ywwg:
I'll give it a shot

ywwg:
hm, ok try this:

enable sync
1. load track @ 174bpm in deck 1
2. load track @ 87 bpm in deck 2
--> works correctly, tracks are still at native bpm

3. adjust rate slider on deck 1
--> deck 2 bpm suddenly jumps to +92, matching the exact bpm of deck 1 (instead of staying at 1/2 the bpm).

Although, I think it's doing that in HEAD as well!

This is why I occasionally suggest that we should drop the half/double feature... it's too hard to get right without these edge cases.  People just need to make sure their tracks are detected right.

acolombier:
I will try and fix this usecase. Just to confirm, we have
1. load track @ 174bpm in deck 1
2. load track @ 87 bpm in deck 2
3. Sync deck 2 to 1, keep 87 BPM on deck 2
4. adjust rate slider on deck 1 to 184bpm
5. Sync deck 2 to 1, get 92 BPM on deck 2
  --> currently get 184bpm, which is not expected

ywwg:
probably when the rate slider signal is being processed, the sync code is not calling the bpm multiplier code somewhere

JoergAtGithub:
Yeah, testcases are the way to go here!

acolombier:
@ywwg I tried really hard the usecase and more stress test (disable and re-enable sync, adjust leader slider again, then follower, change leader, ...) and I couldn't get it to behave as you suggested.

I have added[ a unit test that capture that reproducing steps](https://github.com/mixxxdj/mixxx/pull/13671/commits/8bbd004254b2e5ddad436588889f8c5b96bb1ed6) but it is passing. Could you confirm I translated it correctly? Is that 100% reproducible btw?

ywwg:
Here, the code does not do the correct multiplier on first load, but it does on the second.  And then, the rate slider adjustment shows the bug as I described


https://github.com/user-attachments/assets/473657b4-7c16-4fa3-8a2a-3a9593a0e41e

acolombier:
Right, it took me some times, but I finally got the source of the bug. There is a race condition when you load a track, where the BPM multiplier is recalculated by the audio thread before the file has loaded (some BPM is invalid -> multiplier get's the value of 1). then later the CachingReader completes the file load, and reinit the multiplier to one.

Honestly, it doesn't look trivial to fix, so since this is another bug that already exists in main, how would you feel with merging this bugfix already? I have added the test case that reproduce this issue so if that's okay with you, I will create an issue to address that separately.

ywwg:
that's ok by me.  Thanks for looking in to it.

src/track/bpm.h
```diff
@@ -45,6 +45,10 @@ class Bpm final {
         return isValidValue(m_value);
     }
 
+    bool isNull() const {
```
ywwg, line 48: This is not testing for null, so at the very least it should be 

```suggestion
    bool isMinValue() const {
```

However we have other places in the code that simply test for bpm == 0.0 (see https://github.com/mixxxdj/mixxx/blob/main/src/engine/controls/bpmcontrol.cpp#L1107), so either we need to update those locations with the new function, or, my preference, just test for == 0.0.

src/engine/sync/synccontrol.cpp
ywwg: isValidValue only returns true if `kValueMin < value;`, so 0.0 is not a valid bpm by that function, so the original check should be fine.

ywwg (changes requested):

src/track/bpm.h
ywwg: but I think this comment is obviated by the one below

src/engine/sync/synccontrol.cpp
acolombier: Ah indeed. I thought I had seen the min value being included in the valid range. 
I removed the above

acolombier:
Created #13689 to address the issue above

ywwg (approved):

PR #13672, opened 2024-09-20
Musicbrainz: Improved messages in case of empty results 

### LINKED ISSUE #13666: "Could not find this track in the MusicBrainz database." in case the lokal track file is not accessible
### Bug Description

This is a misleading message, leading the user in the wrong direction. 

### Version

2.4.1

### OS

Ubuntu Focal

daschuer:
Allow to use different messages in case Musikbrainz lookup ends without results.
This fixes #13666

Since this PR moves a translatable text form on object to another we need an Transifex cycle to not loose the translation in the stable 2.4 branch.

### DIFF
--- a/src/musicbrainz/tagfetcher.cpp
+++ b/src/musicbrainz/tagfetcher.cpp
@@ -116,7 +116,8 @@ void TagFetcher::slotFingerprintReady() {
     if (fingerprint.isEmpty()) {
         emit resultAvailable(
                 m_pTrack,
-                QList<mixxx::musicbrainz::TrackRelease>());
+                {},
+                tr("Reading track for fingerprinting failed."));
         return;
     }
 
@@ -165,7 +166,8 @@ void TagFetcher::slotAcoustIdTaskSucceeded(
 
         emit resultAvailable(
                 std::move(pTrack),
-                QList<mixxx::musicbrainz::TrackRelease>());
+                {},
+                tr("Could not identify track through Acoustid."));
         return;
     }
 
@@ -299,9 +301,15 @@ void TagFetcher::slotMusicBrainzTaskSucceeded(
     auto pTrack = m_pTrack;
     terminate();
 
+    QString whyEmptyMessage;
+    if (guessedTrackReleases.empty()) {
+        whyEmptyMessage = tr("Could not find this track in the MusicBrainz database.");
+    }
+
     emit resultAvailable(
             std::move(pTrack),
-            std::move(guessedTrackReleases));
+            std::move(guessedTrackReleases),
+            whyEmptyMessage);
 }
 
 void TagFetcher::startFetchCoverArtLinks(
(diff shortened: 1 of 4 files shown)


src/musicbrainz/tagfetcher.cpp
```diff
@@ -165,7 +166,8 @@ void TagFetcher::slotAcoustIdTaskSucceeded(
 
         emit resultAvailable(
                 std::move(pTrack),
-                QList<mixxx::musicbrainz::TrackRelease>());
+                {},
+                tr("Could not identify track through Acoustid."));
```
acolombier, line 170: Nit: casing of `AcoustID`

acolombier (approved):
LGTM

Just a small optional nit

src/musicbrainz/tagfetcher.cpp
```diff
@@ -299,9 +301,15 @@ void TagFetcher::slotMusicBrainzTaskSucceeded(
...
+    QString whyEmptyMessage;
+    if (guessedTrackReleases.empty()) {
+        whyEmptyMessage = tr("Could not find this track in the MusicBrainz database.");
+    }
+
     emit resultAvailable(
             std::move(pTrack),
-            std::move(guessedTrackReleases));
+            std::move(guessedTrackReleases),
+            whyEmptyMessage);
```
Swiftb0y, line 312: the check is clunky and unnecessary from what I can tell. Can you just inline like you did with the other calls?

Swiftb0y: so just
```suggestion
    emit resultAvailable(
            std::move(pTrack),
            std::move(guessedTrackReleases),
            tr("Could not find this track in the MusicBrainz database.");
```

Swiftb0y: if unconditionally looking up the translated string presents too much overhead, we can also introduce an enum and switch on it only in the failure case (in `DlgTagFetcher::fetchTagFinished`).

daschuer: This way we add the text even if we have a list of tracks. I think it is clearer to have to dedicated emmit calls.

Swiftb0y: yes, one in case of success and a different one in case of error would also be better.

src/musicbrainz/tagfetcher.cpp
daschuer: I have also fixed another occourenc, and will also fix Translations after merge.

src/musicbrainz/tagfetcher.cpp
Swiftb0y: I rather thought of two different signals (with differing parameters), but the current solution works too ig. not widening the scope too much.

Swiftb0y (approved):
LGTM. Thanks you. Waiting for CI.

daschuer:
clazy fixed

ronso0:
Thank you!

PR #13673, opened 2024-09-20
Musicbrainz: Elide long error messages in the  progress bar

### LINKED ISSUE #13664: Musicbrainz error is too long for progress bar
### Bug Description

grafik

### Version

2.5-beta 52

### OS

Ubuntu Jammy

daschuer:
And put the full text into a tooltip, like we do in the library table. 
This fixes #13664

### DIFF
--- a/src/library/dlgtagfetcher.cpp
+++ b/src/library/dlgtagfetcher.cpp
@@ -272,6 +272,7 @@ void DlgTagFetcher::loadTrack(const TrackPointer& pTrack) {
     statusMessage->setVisible(false);
     loadingProgressBar->setVisible(true);
     loadingProgressBar->setValue(kMinimumValueOfQProgressBar);
+    loadingProgressBar->setToolTip(QString());
     addDivider(tr("Original tags"), tags);
     addTagRow(trackColumnValues(*m_pTrack), kOriginalTrackIndex, tags);
 
@@ -435,6 +436,7 @@ void DlgTagFetcher::retry() {
     checkBoxTags->setDisabled(true);
     checkBoxCover->setDisabled(true);
     loadingProgressBar->setValue(kMinimumValueOfQProgressBar);
+    loadingProgressBar->setToolTip(QString());
     m_tagFetcher.startFetch(m_pTrack);
 }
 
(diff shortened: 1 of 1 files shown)


acolombier (approved):
LGTM

src/library/dlgtagfetcher.cpp
```diff
@@ -535,8 +537,16 @@ void DlgTagFetcher::slotNetworkResult(
         const QString& app,
         const QString& message,
         int code) {
-    QString cantConnect = tr("Can't connect to %1: %2");
-    loadingProgressBar->setFormat(cantConnect.arg(app, message));
+    QString cantConnect = tr("Can't connect to %1: %2").arg(app, message);
+    QFontMetrics metrics(loadingProgressBar->font());
+    QString elidedCantConnect = metrics.elidedText(
```
ronso0, line 542: All are const, right?

ronso0:
I wonder if we need the full error string/url at all.
Wouldn't it suffice to show "Can't connect to MusicBrainz" and log the full string?
Or "Can't connect to MusicBrainz. See log for details."

Besides, I wouldn't expect a tooltip on a progressbar.
(is the tooltip always shown, regardless the tooltip setting in Interface preferences?)

src/library/dlgtagfetcher.cpp
```diff
@@ -535,8 +537,16 @@ void DlgTagFetcher::slotNetworkResult(
...
+            cantConnect,
+            Qt::ElideRight,
+            loadingProgressBar->width() - 4);
+    loadingProgressBar->setFormat(elidedCantConnect);
+    if (cantConnect.size() != elidedCantConnect.size()) {
```
Swiftb0y, line 547: Is there the possibility of the edge case that the elided text has the same length as the non-elided text? Could this happen when the text gets elided but then the added ellipses pads it back out to the same length again? That would cause information to be missing (not much sure, but still a character or two) and no tooltip being shown. 
Unless we know this is a performance problem, I would just always set the tooltip. makes the code simpler and avoids the edge case.

daschuer:
> I wonder if we need the full error string/url at all.
> Wouldn't it suffice to show "Can't connect to MusicBrainz" and log the full string?

We show the translated error string form the Qt backend that is intendend to be displayed to the user. So I consider it a good idea to show it. The errror message is long if it contains a long path auto generated path, so eliding will probably not cut off sigificant data and we may consider to remove the tooltip.

> is the tooltip always shown, regardless the tooltip setting in Interface preferences?

Only if the original message is too long and independent froem the preferendes. 

This was just a quick workaround. If you prefer I will remove the tooltip. Please give a hint.

src/library/dlgtagfetcher.cpp
daschuer: Right, that is likely, but no gurantee, because the elipsis can be either the … (https://www.compart.com/de/unicode/U+2026) or ... (three dots) depending on the font.

Swiftb0y (approved):
LGTM. Thank you.

daschuer:
Merge? @ronso0 Is that OK for now?

ronso0 (approved):
Yup, LGTM.
Thank you!

PR #13678, opened 2024-09-21
Paintable: Fix some type roundtrips

daschuer:
This was a by catch during debugging https://github.com/mixxxdj/mixxx/pull/12904
It can be the base for caching the scaled QPixmap

PR #12407, opened 2023-12-07
svg viewbox fix

m0dB:
Adjusted the Latenight palemoon and classic `btn__*.svg` so they render correctly on retina + some changes by @ronso0 to the qss files.

Almost all files were done automatically (with a perl script), multiplying the width and height by two and inserting viewBox with the non-multiplied width and height. The `btn__fx_selector_down.svg` and `btn__xfader_deck_*.svg` and `btn__xfader_aux_*.svg` were done manually.

This replaces https://github.com/mixxxdj/mixxx/pull/12364 and fixes https://github.com/mixxxdj/mixxx/issues/12361

PR #13679, opened 2024-09-21
Scaled svg cache

daschuer:
This is on top of https://github.com/mixxxdj/mixxx/pull/13678

It is used to demonstrates a scaled SVG cache. 

@JoergAtGithub Can you benchmark that on your side. With manual testing I do not notice an improvement.
There are currently test debug messages, to trace how it works. they need to be removed before benchmarking.

When we have confirmed, I will streamline the whole class to properly consider the device pixel ratio everywhere.

### DIFF
--- a/src/widget/paintable.cpp
+++ b/src/widget/paintable.cpp
@@ -237,8 +237,10 @@ void Paintable::drawInternal(const QRectF& targetRect, QPainter* pPainter,
                 m_pPixmap = std::make_unique<QPixmap>(m_pSvg->defaultSize());
                 m_pPixmap->setDevicePixelRatio(devicePixelRatio);
                 m_pPixmap->fill(Qt::transparent);
-                auto pixmapPainter = QPainter(m_pPixmap.get());
-                m_pSvg->render(&pixmapPainter);
+                { // QPainter Scope
+                    auto pixmapPainter = QPainter(m_pPixmap.get());
+                    m_pSvg->render(&pixmapPainter);
+                }
                 mayCorrectColors();
             }
             // The SVG renderer doesn't directly support tiling, so we render
@@ -253,9 +255,11 @@ void Paintable::drawInternal(const QRectF& targetRect, QPainter* pPainter,
                 m_pPixmap = std::make_unique<QPixmap>(targetRect.size().toSize());
                 m_pPixmap->setDevicePixelRatio(devicePixelRatio);
                 m_pPixmap->fill(Qt::transparent);
-                auto pixmapPainter = QPainter(m_pPixmap.get());
-                m_pSvg->setViewBox(sourceRect);
-                m_pSvg->render(&pixmapPainter);
+                { // QPainter Scope
+                    auto pixmapPainter = QPainter(m_pPixmap.get());
+                    m_pSvg->setViewBox(sourceRect);
+                    m_pSvg->render(&pixmapPainter);
+                }
                 mayCorrectColors();
                 m_lastSourceRect = sourceRect;
             }
--- a/src/widget/paintable.cpp
+++ b/src/widget/paintable.cpp
@@ -100,39 +100,43 @@ bool Paintable::isNull() const {
 }
 
 QSize Paintable::size() const {
+    if (m_pSvg) {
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
I just tried it, and can confirm, that this solves the issue as well!

daschuer:
The initial booting takes longer with caching enabled: 
```
debug [Main] Stat("SkinLoader::loadConfiguredSkin","count=1,sum=2.023e+09ns,average=2.023e+09ns,min=2.023e+09ns,max=2.023e+09ns,variance=0ns^2,stddev=0ns")
debug [Main] Stat("SkinLoader::loadConfiguredSkin","count=1,sum=2.06473e+09ns,average=2.06473e+09ns,min=2.06473e+09ns,max=2.06473e+09ns,variance=0ns^2,stddev=0ns")
```
Tested with LateNight Classic

daschuer:
It turns out that the Apple conditions are no longer relevant in Qt5 after we have removed the custom scaling. 
Ready for review.

JoergAtGithub:
This crashes for me reproduceable, if I switch to Shade `Summer Sunset` or `Dark`. While other Skins including Shade `Classic` work as expected:
Unhandled exception at 0x00007FFC1D0CF6FE (ucrtbase.dll) in mixxx.exe: Fatal program exit requested.
grafik
grafik
This does not happen with 2.4 ( tested after merge of #13678 ), so I guess something with the color correction code of this PR is wrong.

daschuer:
Good catch. A narrowing the painter scope fixes the issue. I have no idea why it does not crash when starting with Shade Sunset or Shade Dark. Maybe by luck?

daschuer:
Ah got it. Without the last commit it crashes here all the time, no luck anymore ... good.

daschuer:
Done.

JoergAtGithub:
Works as expected now! Thank you! Code looking good to me!

JoergAtGithub (approved):

daschuer:
Thank you for review. Merge?

JoergAtGithub:
I think we need a test with a Mac and the high-DPI Retina-Display as well. @fwcd / @m0dB could you give this PR a try?
Expected is a less blurry graphic with Shade skin (please test with all three color schemes).
It would be also interesting how the SVG from pre https://github.com/mixxxdj/mixxx/pull/12407 behave now. I could imagine, that all platforms behave here identical now.

daschuer:
There are only a few SVG buttons in Shade. 
The battery indicator and the Auto DJ buttons are SVGs and may benefit in Hidpi mode.

m0dB:
Note quite sure what I am looking at, but the auto-dj button does not use retina for sure.
Screenshot 2024-09-27 at 22 15 55

m0dB:
And when switching back to LateNight palemoon I get this:
 
Screenshot 2024-09-27 at 22 18 23

m0dB:
Screenshot 2024-09-27 at 22 18 23

JoergAtGithub:
> It turns out that these buttons are rendered via qss and not via the Paintable. @m0dB: Can you confirm if you also see blurry buttons with the 2.4 branch?

There seems to be related changes in Qt 6.8:
grafik

ronso0:
This will unfortunatly not be available in the next relevant Ubuntu version.
If we stick to the plan that the latest Ubuntu LTS defines the minimum Qt version we can rely on, we'll have
Qt 6.2.4 in Mixxx 2.5, and Qt 6.6.2 in Mixxx 2.6.
Maybe 26.04 will use Qt 6.8, this is roughly 2 years from now : \

daschuer:
Ok, this means we can merge this safety? 

With QSS SVGs we have now discovered two issues:
* Repeated disc access when hoovering on all targets 
* Blurry pictures on macOs 

So maybe it is worth to hack around the Qt limitations

m0dB:
I confirm that the qss buttons look blurry with 2.4. This branch does not improve that (but IIRC that is expected?)

I confirm that the scaling issue that was introduced with Latenight is solved.

JoergAtGithub:
Great! Thanks for testing again!

PR #13680, opened 2024-09-21
(fix) AnalyzerEbur128: catch invalid return values inf/-inf

ronso0:
see #13676

### DIFF
--- a/src/analyzer/analyzerebur128.cpp
+++ b/src/analyzer/analyzerebur128.cpp
@@ -69,7 +69,11 @@ void AnalyzerEbur128::storeResults(TrackPointer pTrack) {
         qWarning() << "AnalyzerEbur128::storeResults() failed with" << e;
         return;
     }
-    if (averageLufs == -HUGE_VAL || averageLufs == 0.0) {
+    if (averageLufs == -HUGE_VAL ||
+            averageLufs == HUGE_VAL ||
+            // This catches 0 and abnormal values inf and -inf (that may have
+            // slipped through in libebur for some reason.
+            util_isnormal(averageLufs)) {
         qWarning() << "AnalyzerEbur128::storeResults() averageLufs invalid:"
                    << averageLufs;
         return;
--- a/src/util/versionstore.cpp
+++ b/src/util/versionstore.cpp
@@ -22,6 +22,7 @@
 
 #include <FLAC/format.h>
 #include <chromaprint.h>
+#include <ebur128.h>
 #include <lame/lame.h>
 #include <portaudio.h>
 #include <sndfile.h>
@@ -151,6 +152,12 @@ QStringList VersionStore::dependencyVersions() {
     sf_command(nullptr, SFC_GET_LIB_VERSION, sndfile_version, sizeof(sndfile_version));
     // Null-terminate just in case.
     sndfile_version[sizeof(sndfile_version) - 1] = '\0';
+
+    int eburMaj = 0;
+    int eburMin = 0;
+    int eburP = 0;
+    ebur128_get_version(&eburMaj, &eburMin, &eburP);
+
     // WARNING: may be inaccurate since some come from compile-time header
     // definitions instead of the actual dynamically loaded library).
     QStringList result;
@@ -181,6 +188,10 @@ QStringList VersionStore::dependencyVersions() {
                        .arg(QString::number(CHROMAPRINT_VERSION_MAJOR),
                                QString::number(CHROMAPRINT_VERSION_MINOR),
                                QString::number(CHROMAPRINT_VERSION_PATCH))
+            << QString("libebur128: %1.%2.%3")
+                       .arg(QString::number(eburMaj),
+                               QString::number(eburMin),
+                               QString::number(eburP))
             // Should be accurate.
             << QString("Vorbis: %1").arg(vorbis_version_string())
             // Should be accurate.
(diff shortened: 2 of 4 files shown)


src/util/versionstore.cpp
```diff
@@ -151,6 +152,10 @@ QStringList VersionStore::dependencyVersions() {
     sf_command(nullptr, SFC_GET_LIB_VERSION, sndfile_version, sizeof(sndfile_version));
     // Null-terminate just in case.
     sndfile_version[sizeof(sndfile_version) - 1] = '\0';
+
+    int eburMaj, eburMin, eburP = 0;
```
Swiftb0y, line 156: this leaves the first two variables uninitialized, which is one of the key reason why this style is discouraged.
```suggestion
    int eburMaj = 0;
    int eburMin = 0; 
    int eburP = 0;
```

Swiftb0y (changes requested):

src/util/versionstore.cpp
ronso0: thanks, I pushed a fixup.

Swiftb0y (approved):
LGTM. Thank you. How reproducible is the issue? We should escalate this to upstream libebur128.

ronso0:
I can reproduce it with a silent track (empty Mixxx recording, ~3 sec)
[space.mp3.zip](https://github.com/user-attachments/files/17085084/space.mp3.zip)

Swiftb0y:
right, makes sense that you'd need "infinite" gain to get that to a fixed level. This should be caught but libebur though.

src/analyzer/analyzerebur128.cpp
```diff
@@ -69,7 +69,13 @@ void AnalyzerEbur128::storeResults(TrackPointer pTrack) {
         qWarning() << "AnalyzerEbur128::storeResults() failed with" << e;
         return;
     }
-    if (averageLufs == -HUGE_VAL || averageLufs == 0.0) {
+    if (averageLufs == -HUGE_VAL ||
+            averageLufs == HUGE_VAL ||
+            // This catches inf and -inf that slipped through in libebur for some reason..
+            // TODO Move this to a VERIFY_OR_DEBUG_ASSERT() once the root cause
+            // has been identified.
```
daschuer, line 76:
```suggestion
            // This catches inf and -inf that slipped through in libebur for some reason.
            // TODO: Move this to a VERIFY_OR_DEBUG_ASSERT() once the root cause
            // has been identified.
```

src/analyzer/analyzerebur128.cpp
```diff
@@ -69,7 +69,13 @@ void AnalyzerEbur128::storeResults(TrackPointer pTrack) {
...
+            util_isinf(averageLufs) != 0 ||
```
daschuer, line 77: util_isinf introduces a insymmetry compared the the C89 HUGE_VAL constant. 

How about this:
```suggestion
            util_isnormal(averageLufs) ||
```
We can only really deal with a normal value. 
The alternative would be to check for HUGE_VAL instead. 

There is by the way the issue that util_isnormal() shall return bool, not int.

daschuer (changes requested):

src/analyzer/analyzerebur128.cpp
ronso0: Thanks, I've prepended a commit to fix the return types in fpclassify and use `util_isnormal()` here

ronso0:
> How reproducible is the issue? We should escalate this to upstream libebur128.

Hmm, the last (code) activity was 3 years ago.

src/analyzer/analyzerebur128.cpp
```diff
@@ -69,7 +69,13 @@ void AnalyzerEbur128::storeResults(TrackPointer pTrack) {
...
         return;
     }
-    if (averageLufs == -HUGE_VAL || averageLufs == 0.0) {
+    if (averageLufs == -HUGE_VAL ||
+            averageLufs == HUGE_VAL ||
+            // This catches 0, inf and -inf (that may have slipped through in
+            // libebur for some reason.
+            // TODO Move this to a VERIFY_OR_DEBUG_ASSERT() once the root cause
+            // has been identified, and only do a zero check here.
+            util_isnormal(averageLufs)) {
```
daschuer, line 78: Now the check for 0 is gone. 
My idea was to just

```suggestion
    if (!util_isnormal(averageLufs) || averageLufs == 0.0) {
            // This catches 0, inf (HUGE_VAL), -inf (-HUGE_VAL) and denormals (treated as zero), 
            // that may have slipped through in libebur for some reason.
```

Returning -HUGE_VAL is the documented invalid value. I am in doubt there will be a "fix" for HUGE_VAL and even then we deal with user data, where Bad things are not "impossible". So I would not add the TODO.

daschuer (changes requested):

src/analyzer/analyzerebur128.cpp
ronso0: Zero is covered by std::isnormal
https://en.cppreference.com/w/cpp/numeric/math/isnormal

I'll remove the TODO then.

daschuer: Ups, thank you for the hint.

daschuer (approved):
LGTM, thank you.

PR #13682, opened 2024-09-22
BaseTrackTableModel: Fix column header mapping

uklotzde:
Without these fixes `BaseTrackTableModel` doesn't work as expected when used as a base class, i.e. internal column values are undefined and visible columns are missing.

However `BaseSqlTableModel` doesn't seem to suffer from these deficiencies and inconsistencies in its base class.

### DIFF
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -35,40 +35,6 @@ const mixxx::Logger kLogger("BaseTrackTableModel");
 constexpr double kRelativeHeightOfCoverartToolTip =
         0.165; // Height of the image for the cover art tooltip (Relative to the available screen size)
 
-// Mandatory columns that are either used directly or in delegates.
-const QStringList kDefaultTableColumns = {
-        LIBRARYTABLE_ALBUM,
-        LIBRARYTABLE_ALBUMARTIST,
-        LIBRARYTABLE_ARTIST,
-        LIBRARYTABLE_BPM,
-        LIBRARYTABLE_BPM_LOCK,
-        LIBRARYTABLE_BITRATE,
-        LIBRARYTABLE_CHANNELS,
-        LIBRARYTABLE_COLOR,
-        LIBRARYTABLE_COMMENT,
-        LIBRARYTABLE_COMPOSER,
-        LIBRARYTABLE_COVERART,
-        LIBRARYTABLE_DATETIMEADDED,
-        LIBRARYTABLE_DURATION,
-        LIBRARYTABLE_FILETYPE,
-        LIBRARYTABLE_GENRE,
-        LIBRARYTABLE_GROUPING,
-        LIBRARYTABLE_KEY,
-        LIBRARYTABLE_KEY_ID,
-        TRACKLOCATIONSTABLE_LOCATION,
-        TRACKLOCATIONSTABLE_FSDELETED,
-        LIBRARYTABLE_PLAYED,
-        LIBRARYTABLE_PREVIEW,
-        LIBRARYTABLE_RATING,
-        LIBRARYTABLE_REPLAYGAIN,
-        LIBRARYTABLE_SAMPLERATE,
-        LIBRARYTABLE_TIMESPLAYED,
-        LIBRARYTABLE_LAST_PLAYED_AT,
-        LIBRARYTABLE_TITLE,
-        LIBRARYTABLE_TRACKNUMBER,
-        LIBRARYTABLE_YEAR,
-};
-
 inline QSqlDatabase cloneDatabase(
         const QSqlDatabase& prototype) {
     const auto connectionName =
@@ -122,11 +88,6 @@ bool BaseTrackTableModel::s_bApplyPlayedTrackColor =
 
 void BaseTrackTableModel::setApplyPlayedTrackColor(bool apply) {
     s_bApplyPlayedTrackColor = apply;
-}
-
-//static
-QStringList BaseTrackTableModel::defaultTableColumns() {
-    return kDefaultTableColumns;
 }
 
 BaseTrackTableModel::BaseTrackTableModel(
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -121,23 +121,21 @@ BaseTrackTableModel::BaseTrackTableModel(
 void BaseTrackTableModel::initTableColumnsAndHeaderProperties(
         const QStringList& tableColumns) {
     m_columnCache.setColumns(tableColumns);
-    if (m_columnHeaders.size() < tableColumns.size()) {
-        m_columnHeaders.resize(tableColumns.size());
-        // Init the mapping of all columns, even for internal columns that are
-        // hidden/invisible. Otherwise mapColumn() would not return a valid result
-        // for those columns.
-        for (int columnValue = 0; columnValue < ColumnCache::NUM_COLUMNS; ++columnValue) {
-            const auto column = static_cast<ColumnCache::Column>(columnValue);
-            DEBUG_ASSERT(column != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
-            const int headerIndex = m_columnCache.fieldIndex(column);
-            if (headerIndex < 0) {
-                // Missing table column.
-                continue;
-            }
-            DEBUG_ASSERT(headerIndex < m_columnHeaders.size());
-            m_columnHeaders[headerIndex].column = column;
-            DEBUG_ASSERT(mapColumn(headerIndex) == column);
+    m_columnHeaders.resize(tableColumns.size());
+    // Init the mapping of all columns, even for internal columns that are
+    // hidden/invisible. Otherwise mapColumn() would not return a valid result
+    // for those columns.
+    for (int columnValue = 0; columnValue < ColumnCache::NUM_COLUMNS; ++columnValue) {
+        const auto column = static_cast<ColumnCache::Column>(columnValue);
+        DEBUG_ASSERT(column != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
+        const int headerIndex = m_columnCache.fieldIndex(column);
+        if (headerIndex < 0) {
+            // Missing table column.
+            continue;
         }
+        DEBUG_ASSERT(headerIndex < m_columnHeaders.size());
+        m_columnHeaders[headerIndex].column = column;
+        DEBUG_ASSERT(mapColumn(headerIndex) == column);
     }
     initHeaderProperties();
 }
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -159,8 +159,8 @@ BaseTrackTableModel::BaseTrackTableModel(
 
 void BaseTrackTableModel::initTableColumnsAndHeaderProperties(
         const QStringList& tableColumns) {
+    // This function might be invoked repeatedly and not only once during construction.
     m_columnCache.setColumns(tableColumns);
-    DEBUG_ASSERT(m_columnHeaders.isEmpty());
     m_columnHeaders.resize(tableColumns.size());
     // Init the mapping of all columns, even for internal columns that are
     // hidden/invisible. Otherwise mapColumn() would not return a valid result
@@ -174,7 +174,6 @@ void BaseTrackTableModel::initTableColumnsAndHeaderProperties(
             continue;
         }
         DEBUG_ASSERT(headerIndex < m_columnHeaders.size());
-        DEBUG_ASSERT(mapColumn(headerIndex) == ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
         m_columnHeaders[headerIndex].column = column;
         DEBUG_ASSERT(mapColumn(headerIndex) == column);
     }
--- a/src/library/basetracktablemodel.cpp
+++ b/src/library/basetracktablemodel.cpp
@@ -126,8 +126,14 @@ void BaseTrackTableModel::initTableColumnsAndHeaderProperties(
     m_columnHeaders.clear();
     m_columnHeaders.resize(tableColumns.size());
 
-    // Init the mapping of all columns, even for internal columns that are
-    // hidden/invisible. Otherwise mapColumn() would not return a valid result
+    // Init the header properties of selected/visible columns.
+    // This method might by overridden by derived classes.
+    initHeaderProperties();
+    // The derived class might have added column headers.
+    DEBUG_ASSERT(m_columnHeaders.size() >= tableColumns.size());
+
+    // Init the mapping of all remaining columns, even for internal columns that
+    // are hidden/invisible. Otherwise mapColumn() would not return a valid result
     // for those columns.
     for (int columnValue = 0; columnValue < ColumnCache::NUM_COLUMNS; ++columnValue) {
         const auto column = static_cast<ColumnCache::Column>(columnValue);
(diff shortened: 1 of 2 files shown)


src/library/basetracktablemodel.cpp
```diff
@@ -587,18 +594,23 @@ QVariant BaseTrackTableModel::rawSiblingValue(
...
-        return QVariant();
+        qWarning() << "Unknown sibling column"
+                   << siblingField
+                   << m_columnCache.columnName(siblingField)
+                   << "for model index"
+                   << index;
+        return {};
     }
     const QModelIndex siblingIndex = index.sibling(index.row(), siblingColumn);
+    DEBUG_ASSERT(siblingIndex.isValid());
```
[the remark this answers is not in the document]
uklotzde: I did not change this code. It works as before. I only added more debug assertions to catch unexpected behavior during development.

The last debug assertion is just to verify that the resulting `siblingIndex` is valid, independent of what `rawValue()` does. `rawValue()` could be overridden and we cannot make any assumptions about its actual implementation.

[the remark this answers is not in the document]
uklotzde: https://github.com/mixxxdj/mixxx/pull/13682/commits/a3ddee75c41c72bb736411d622b816add36746d5

I have reverted the unrelated changes, because other table models (e.g. history) still seem to be affected.

src/library/basetracktablemodel.cpp
```diff
@@ -157,8 +160,23 @@ BaseTrackTableModel::BaseTrackTableModel(
...
+    for (auto columnValue = 0; columnValue < ColumnCache::NUM_COLUMNS; ++columnValue) {
+        const auto column = static_cast<ColumnCache::Column>(columnValue);
+        DEBUG_ASSERT(column != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
+        const auto headerIndex = m_columnCache.fieldIndex(column);
+        if (headerIndex < 0) {
+            // Missing table column.
+            continue;
+        }
+        DEBUG_ASSERT(headerIndex < m_columnHeaders.size());
+        DEBUG_ASSERT(mapColumn(headerIndex) == ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
```
Swiftb0y, line 177: this doesn't look right, it shouldn't be invalid, right?
```suggestion
        DEBUG_ASSERT(mapColumn(headerIndex) != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
```

uklotzde: before: invalid, after: valid

uklotzde: Obsolete: https://github.com/mixxxdj/mixxx/pull/13682/commits/bb0495e7f1cfb2fc3fd327b90f8bc8a52ff79159

Coincidentally this just happened while switching the view during a library rescan. Never occurred before.

src/library/basetracktablemodel.cpp
```diff
@@ -35,6 +35,7 @@ const mixxx::Logger kLogger("BaseTrackTableModel");
 constexpr double kRelativeHeightOfCoverartToolTip =
         0.165; // Height of the image for the cover art tooltip (Relative to the available screen size)
 
+// Mandatory columns that are either used directly or in delegates.
 const QStringList kDefaultTableColumns = {
```
Swiftb0y, line 39: unrelated to your changes, just something I noticed in general: The content here doesn't actually matter, the only thing thats ever used by `kDefaultTableColumns` is its size. Thats probably not how this was supposed to be used originally.

uklotzde: It is used if you derive your custom model from `BaseTrackTableModel` and want/need to reuse the implicitly defined and supported default columns.

`BaseSqlTableModel` seems to compose its own list that somehow matches the implicit requirements of its base class.

The whole design is neither composable nor maintainable. I just use it as is.

uklotzde: Good catch: https://github.com/mixxxdj/mixxx/pull/13682/commits/9fdafecad0ad55ce15621703a16564ee00954e06

I'll maintain a local patch and need to keep track of all future changes and extensions myself.

Swiftb0y: Thanks for doing the cleanup (even though I didn't ask you to).

Swiftb0y:
I'm not sure how to proceed here. Who is qualified enough to review this?

src/library/basetracktablemodel.cpp
uklotzde: It was probably me who added this unneeded code when splitting off the base class  :sweat_smile:

src/library/basetracktablemodel.cpp
```diff
@@ -156,9 +120,23 @@ BaseTrackTableModel::BaseTrackTableModel(
 
 void BaseTrackTableModel::initTableColumnsAndHeaderProperties(
         const QStringList& tableColumns) {
+    // This function might be invoked repeatedly and not only once during construction.
```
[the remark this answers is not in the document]
uklotzde: The method is called repeatedly when activating a crate view. The debug assertions that should verify that this method is called at most once triggered:

```
#5  0x00000000006e4556 in mixxx_debug_assert (assertion=<optimized out>, file=<optimized out>, line=<optimized out>, function=<optimized out>)
    at /home/uk/volumes/Build/cpp/mixxx/src/util/assert.h:9
#6  0x0000000000bbb08e in BaseTrackTableModel::initTableColumnsAndHeaderProperties (this=0x302a5eb8, tableColumns=...)
    at /home/uk/volumes/Build/cpp/mixxx/src/library/basetracktablemodel.cpp:142
#7  0x0000000000b883c0 in BaseSqlTableModel::setTable (this=this@entry=0x302a5eb8, tableName=..., idColumn=..., tableColumns=..., trackSource=...)
    at /home/uk/volumes/Build/cpp/mixxx/src/library/basesqltablemodel.cpp:400
#8  0x0000000000e26586 in CrateTableModel::selectCrate (this=this@entry=0x302a5eb8, crateId=..., crateId@entry=...) at /usr/include/qt6/QtCore/qarraydata.h:59
#9  0x0000000000e05e87 in CrateFeature::activateChild (this=0x302a5e40, index=...) at /home/uk/volumes/Build/cpp/mixxx/src/library/trackset/crate/cratefeature.cpp:310
#10 0x00007ffff07fc8f2 in void doActivate<false>(QObject*, int, void**) () at /lib64/libQt6Core.so.6
#11 0x00007ffff5a7f49c in QAbstractItemView::clicked(QModelIndex const&) () at /lib64/libQt6Widgets.so.6
```

See also: https://github.com/mixxxdj/mixxx/commit/bb0495e7f1cfb2fc3fd327b90f8bc8a52ff79159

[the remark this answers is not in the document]
uklotzde: Doesn't cause issues. Just unexpected and inefficient.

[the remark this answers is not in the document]
uklotzde: https://github.com/mixxxdj/mixxx/pull/13682/commits/bf4af50f236d856eb5af45d5dfc8e49508a7901c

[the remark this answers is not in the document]
uklotzde: I have restored the original optimization.

Just want to mention that this is not correct, because `initHeaderProperties()` is a virtual function. An override in a derived class might depend on the actual contents of `ColumnCache` even if the implementation in the base class does not.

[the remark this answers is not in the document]
uklotzde: This commit should restore the original optimization: https://github.com/mixxxdj/mixxx/pull/13682/commits/dc287ebcf6151355e42ac705b456706ad4343ae3

uklotzde:
I recomment to squash this PR into a single commit when merging, using the title as the commit message.

src/library/basetracktablemodel.cpp
```diff
@@ -159,6 +123,21 @@ void BaseTrackTableModel::initTableColumnsAndHeaderProperties(
...
+            DEBUG_ASSERT(column != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
+            const auto headerIndex = m_columnCache.fieldIndex(column);
+            if (headerIndex < 0) {
+                // Missing table column.
+                continue;
+            }
+            DEBUG_ASSERT(headerIndex < m_columnHeaders.size());
+            m_columnHeaders[headerIndex].column = column;
+            DEBUG_ASSERT(mapColumn(headerIndex) == column);
+        }
```
daschuer, line 140:
```suggestion
        for (int columnValue = 0; columnValue < ColumnCache::NUM_COLUMNS; ++columnValue) {
            const auto column = static_cast<ColumnCache::Column>(columnValue);
            DEBUG_ASSERT(column != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
            const int headerIndex = m_columnCache.fieldIndex(column);
```

src/library/basetracktablemodel.cpp
```diff
@@ -159,6 +123,21 @@ void BaseTrackTableModel::initTableColumnsAndHeaderProperties(
     m_columnCache.setColumns(tableColumns);
     if (m_columnHeaders.size() < tableColumns.size()) {
```
daschuer, line 124: We want that the m_columnHeaders is always current. So I think this condition needs to be removed.

daschuer (changes requested):

src/library/basetracktablemodel.cpp
uklotzde: I didn't notice that re-initialized `m_columnCache` is accessed in the loop.

src/library/basetracktablemodel.cpp
```diff
@@ -159,6 +123,21 @@ void BaseTrackTableModel::initTableColumnsAndHeaderProperties(
...
+        for (auto columnValue = 0; columnValue < ColumnCache::NUM_COLUMNS; ++columnValue) {
+            const auto column = static_cast<ColumnCache::Column>(columnValue);
+            DEBUG_ASSERT(column != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
+            const auto headerIndex = m_columnCache.fieldIndex(column);
+            if (headerIndex < 0) {
+                // Missing table column.
+                continue;
+            }
+            DEBUG_ASSERT(headerIndex < m_columnHeaders.size());
+            m_columnHeaders[headerIndex].column = column;
```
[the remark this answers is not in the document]
uklotzde: Out of scope of this PR.

IMHO the negligible performance overhead for initialitzing the mapping of some columns twice is not worth the maintenance burden.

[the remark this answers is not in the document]
daschuer: At least this is redundant now: 
https://github.com/mixxxdj/mixxx/blob/dc8fba5806cc086b628410eae4d0111261e30758/src/library/basetracktablemodel.cpp#L259-L262

src/library/basetracktablemodel.cpp
```diff
@@ -121,23 +121,21 @@ BaseTrackTableModel::BaseTrackTableModel(
...
-            DEBUG_ASSERT(column != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
-            const int headerIndex = m_columnCache.fieldIndex(column);
-            if (headerIndex < 0) {
-                // Missing table column.
-                continue;
-            }
-            DEBUG_ASSERT(headerIndex < m_columnHeaders.size());
-            m_columnHeaders[headerIndex].column = column;
-            DEBUG_ASSERT(mapColumn(headerIndex) == column);
+    m_columnHeaders.resize(tableColumns.size());
```
daschuer, line 124: A clear() before is missing else ther is a rsik to use patially old data.

src/library/basetracktablemodel.cpp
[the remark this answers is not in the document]
uklotzde: No. Because `setHeaderProperties()` is protected and might be called from another context by derived classes. Not only from within `initTableColumnsAndHeaderProperties()`.

I am not able to make a qualified decision here.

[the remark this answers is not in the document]
daschuer: You have decided to introduce the code here which introduces the first go of redundnandeny: 
https://github.com/mixxxdj/mixxx/blob/dc8fba5806cc086b628410eae4d0111261e30758/src/library/basetracktablemodel.cpp#L137
Can this than be removed?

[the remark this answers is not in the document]
daschuer: > Because `setHeaderProperties()` is protected ...

It looks like the protected state is wrong and it can be made private.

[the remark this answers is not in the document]
uklotzde: > You have decided to introduce the code here which introduces the first go of redundnandeny:
> 
> https://github.com/mixxxdj/mixxx/blob/dc8fba5806cc086b628410eae4d0111261e30758/src/library/basetracktablemodel.cpp#L137
> 
> Can this than be removed?

This line of code is the essence of this PR.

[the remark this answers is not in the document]
uklotzde: > > Because `setHeaderProperties()` is protected ...
> 
> It looks like the protected state is wrong and it can be made private.

Its is used by `BaseSqlTableModel::initHeaderProperties()`. Within the same context, but this is could not be guaranteed for other derived classes.

uklotzde:
I appreciate that this base class is still maintained: https://github.com/mixxxdj/mixxx/pull/13682/commits/8983ca861844906589d70baeba4736e5805a9205

Any chances to get these fixes merged?

Swiftb0y (approved):
LGTM. Thank you. Waiting for @daschuer

uklotzde:
If you are not interested in these fixes I will close the PR, stash the commits, and maintain a local patch. Only affects my use case.

Swiftb0y:
friendly ping @daschuer

src/library/basetracktablemodel.cpp
```diff
@@ -157,9 +121,28 @@ BaseTrackTableModel::BaseTrackTableModel(
...
+        DEBUG_ASSERT(column != ColumnCache::COLUMN_LIBRARYTABLE_INVALID);
+        const int headerIndex = m_columnCache.fieldIndex(column);
+        if (headerIndex < 0) {
+            // Missing table column.
+            continue;
+        }
+        DEBUG_ASSERT(headerIndex < m_columnHeaders.size());
+        m_columnHeaders[headerIndex].column = column;
+        DEBUG_ASSERT(mapColumn(headerIndex) == column);
     }
```
daschuer, line 143: It turns out that the whole loop Is redundant, because the header properties are reinitialized below. There is only a dummy entry missing rot the ID header. Something like:

```
    setHeaderProperties(
            ColumnCache::COLUMN_LIBRARYTABLE_ID,
            QString(), // hidden
            0);
```

You may assert below `initHeaderProperties()` that all `m_columnHeaders` have been initialized. 

It would be better to not initalizer the header properties at runtime, because all columns are known at compile time. But it is OK for me to not extend the scope of this PR as suggested in the other comment.

daschuer (changes requested):
We are still interested in this PR, sorry for the delay. 

Can you please squash this PR as suggested after applying the final fix? 

Thank you.

src/library/basetracktablemodel.cpp
uklotzde: That's how an alternative initialization ordering looks like: https://github.com/mixxxdj/mixxx/pull/13682/commits/e0d43748aad2197f7ce6b651db585c30a0627708. IMHO more complicated and cluttered.

I suppose this code either vanishes or needs to be rewritten when migrating to QML. I wouldn't put too much effort in it.

Squashing can be done on merge. The PR title already contains the final commit message.

daschuer: I was probably not clear enough: Please **remove** this complicated and cluttered loop. It is not necessary. 

Instead just this is requied: 
```
 setHeaderProperties(
            ColumnCache::COLUMN_LIBRARYTABLE_ID,
            QString(), // hidden
            0);
```

daschuer (changes requested):

uklotzde:
@daschuer Sorry, but what you are proposing doesn't work. There are more hidden columns which remain uninitialized. Initializing them one by one is brittle and will break as soon as more (hidden) columns are added in the future. Initializing all columns in a loop is the only safe and sustainable option.

I will maintain a patch for myself and leave the rest to you.

uklotzde:
The proposed changes could be found as a single commit in this (permanent) branch: https://github.com/uklotzde/mixxx/tree/aoide.

I decided to keep the branch accessible for everyone. It will be available as long as rebasing and maintaining is manageable.

Thanks for the review. It has considerably helped improving and simplifying the code.

PR #13684, opened 2024-09-22
Limit mixxx.log size

daschuer:
This sets a limit to 100 MB to lint the issue reported here: 
https://github.com/mixxxdj/mixxx/issues/13660

The limit can be adjusted by the command line parameter --log-max-file-size

### DIFF
--- a/src/util/cmdlineargs.h
+++ b/src/util/cmdlineargs.h
@@ -52,6 +52,9 @@ class CmdlineArgs final {
     bool getSettingsPathSet() const { return m_settingsPathSet; }
     mixxx::LogLevel getLogLevel() const { return m_logLevel; }
     mixxx::LogLevel getLogFlushLevel() const { return m_logFlushLevel; }
+    qint64 getLogMaxFileSize() const {
+        return m_logMaxFileSize;
+    }
     bool getTimelineEnabled() const { return !m_timelinePath.isEmpty(); }
     const QString& getLocale() const { return m_locale; }
     const QString& getSettingsPath() const { return m_settingsPath; }
@@ -91,6 +94,7 @@ class CmdlineArgs final {
     bool m_parseForUserFeedbackRequired;
     mixxx::LogLevel m_logLevel; // Level of stderr logging message verbosity
     mixxx::LogLevel m_logFlushLevel; // Level of mixx.log file flushing
+    qint64 m_logMaxFileSize;
     QString m_locale;
     QString m_settingsPath;
     QString m_resourcePath;
--- a/src/util/logging.cpp
+++ b/src/util/logging.cpp
@@ -12,6 +12,7 @@
 #include <QString>
 #include <QTextStream>
 #include <QThread>
+#include <atomic>
 #include <string_view>
 
 #include "util/assert.h"
@@ -29,7 +30,7 @@ QMutex s_mutexStdErr;
 // The file handle for Mixxx's log file.
 QFile s_logfile;
 qint64 s_logMaxFileSize = mixxx::kLogMaxFileSizeDefault;
-bool s_logMaxFileSizeReached = false;
+std::atomic<bool> s_logMaxFileSizeReached = false;
 
 QLoggingCategory::CategoryFilter oldCategoryFilter = nullptr;
 
@@ -151,7 +152,7 @@ inline void writeToFile(
         const QString& message,
         const QString& threadName,
         bool flush) {
-    if (s_logMaxFileSizeReached) {
+    if (s_logMaxFileSizeReached.load(std::memory_order_relaxed)) {
         return;
     }
 
@@ -168,7 +169,7 @@ inline void writeToFile(
             formattedMessage =
                     "Maximum log file size reached. It can be adjusted via: "
                     "--log-max-file-size <bytes>";
-            s_logMaxFileSizeReached = true;
+            s_logMaxFileSizeReached.store(true, std::memory_order_relaxed);
             flush = true;
         }
         const int written = s_logfile.write(formattedMessage);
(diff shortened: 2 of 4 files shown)


acolombier (commented):
What is the expected behaviour with log rotation? Currently, my log files are rotated with up to 10 log archive:

```
-rw-rw-r--  90K  ../.mixxx/mixxx.log
-rw-rw-r--  90K  ../.mixxx/mixxx.log.1
-rw-rw-r-- 128K  ../.mixxx/mixxx.log.10
-rw-rw-r--  80K  ../.mixxx/mixxx.log.2
-rw-rw-r--  80K  ../.mixxx/mixxx.log.3
-rw-rw-r-- 116K  ../.mixxx/mixxx.log.4
-rw-rw-r--  80K  ../.mixxx/mixxx.log.5
-rw-rw-r-- 121K  ../.mixxx/mixxx.log.6
-rw-rw-r-- 254K  ../.mixxx/mixxx.log.7
-rw-rw-r-- 927K  ../.mixxx/mixxx.log.8
-rw-rw-r--  64K  ../.mixxx/mixxx.log.9
```

Is this something that Mixxx handle cross platform? Is that some logrotate Linux specific behaviour? In case linux already ship with some logrotate setup, I would suggest we keep unlimited the default in there.

Also a suggestion, instead of preventing further append to the log file, shall we discard the N first lines, N being the minimum required line to discard to allow appending the new line. Appreciate this will have a performance impact, but this will also make the log usable. 

Another suggestion, we could solve this issue by introducing a duplication limiter like for ffmpeg, where the log file will not write duplicated lines but will instead write the line once, and maintain a repeater counter on the next line. 

```
[HidIoThread Traktor Kontrol S2 9E2E_3] Unable to read buffered HID InputReports from "Traktor Kontrol S2 9E2E_3" : "hid_read_timeout: unexpected poll error (device disconnected)"
[Previous line repeated 1547 times]
```

The limitation is, you may have a repeated pattern, which wouldn't be de-dupped.

daschuer:
The log rotation happens after each start up of Mixxx. It is a Mixxx home made implementation. 
We can improve on this topic a lot. This PR was just meant as a band aid for a real party stopper. 
Intended for the 2.4. stable branch. 

Does that suite?

acolombier:
I think it make sense to quickly issue that mitigation to 2.4, but I would be worried that this becomes a long term fix as well. Shall we merge that in 2.4 but not sync it with next version till we have a proper solution?

src/util/logging.cpp
```diff
@@ -158,6 +164,13 @@ inline void writeToFile(
     // Writing to a closed QFile could cause an infinite recursive loop
     // by logging to qWarning!
     if (s_logfile.isOpen()) {
+        if (s_logMaxFileSize >= 0 && s_logfile.pos() >= s_logMaxFileSize) {
+            formattedMessage =
+                    "Maximum log file size reached. It can be adjusted via: "
+                    "--log-max-file-size <bytes>";
+            s_logMaxFileSizeReached = true;
+            flush = true;
```
ronso0, line 172: shouldn't this be false?

daschuer: This is the last line written to the file and flushing makes sure that it is also immediately written to the HDD.

src/util/logging.h
```diff
@@ -22,6 +22,7 @@ Q_DECLARE_OPERATORS_FOR_FLAGS(LogFlags);
 
 /// Default log level for (console) logs.
 constexpr LogLevel kLogLevelDefault = LogLevel::Warning;
+constexpr qint64 kLogMaxFileSizeDefault = 100000000; // 100 MB
```
Swiftb0y, line 25:
```suggestion
constexpr qint64 kLogMaxFileSizeDefault = 100'000'000; // 100 MB
```

src/util/cmdlineargs.h
```diff
@@ -52,6 +52,9 @@ class CmdlineArgs final {
     bool getSettingsPathSet() const { return m_settingsPathSet; }
     mixxx::LogLevel getLogLevel() const { return m_logLevel; }
     mixxx::LogLevel getLogFlushLevel() const { return m_logFlushLevel; }
+    qint64 getLogMaxFileSite() const {
+        return m_logMaxFileSize;
+    }
```
Swiftb0y, line 57:
```suggestion
    qint64 getLogMaxFileSize() const {
        return m_logMaxFileSize;
    }
```

Swiftb0y (changes requested):

daschuer:
I have amended your suggestions.

daschuer:
merge?

acolombier:
> What is the main disadvantage?

IMO, this could lead to the log file to include data as it will stop outputting if it reaches the limit. Usually, relevant information to help fixing a bug or a crash tends to be closer to the end of the file, which means this could limit the usefulness of the log file.

In the meantime, 100MB is fairly large, but I can imagine DJs using Mixxx to broadcast to radios for example may extend that limit

daschuer:
Do you propose a different size? Different approach? 
On one hand we have the valid request for not using the entire disk for log files to protect against crashes. On the other hand we want to see everything in case of a crash.

daschuer:
@acolombier can we merge this as it is or is there something to do?

acolombier:
I'm happy with merging this as mitigation. I thought @Swiftb0y had change requests tho?

src/util/logging.cpp
```diff
@@ -28,6 +28,8 @@ QMutex s_mutexStdErr;
 
 // The file handle for Mixxx's log file.
 QFile s_logfile;
+qint64 s_logMaxFileSize = mixxx::kLogMaxFileSizeDefault;
+bool s_logMaxFileSizeReached = false;
```
Swiftb0y, line 32: In spirit of continuing the `tsan` discussion, either this needs to be atomic or the early check below must be moved below the mutex acquire, right?

daschuer:
done

Swiftb0y (approved):
Pretty sure this LGTM.

JoergAtGithub:
Lets merge this!

Issue #13222, opened 2024-05-08
Mark Mixxx util function deprecated when they shouldn't be used anymore.


acolombier:
A few Mixxx util function were flagged as deprecated on recent PR reviews and the request was made to use the std lib instead. In order to help with the DevX, it would be great if the Mixxx Core Team could take the time to mark all utils functions which shouldn't be used anymore as deprecated to prevent unnecessary effort, and document which is the preferred replacement for it. (eg `mixxx::Duration` -> `std::chrono::duration`)

acolombier:
@Swiftb0y do you think there is more to do here or shall we close?

Swiftb0y:
Not sure. lost track. I can go through it again. remind me in case I forget.

acolombier:
Will do! :)

PR #13687, opened 2024-09-23
Cleanup and deprecate more `util/` classes

Swiftb0y:
This should conclude #13222 along with a bunch of drive by fixes I couldn't resist not to do.


src/engine/controls/ratecontrol.cpp
```diff
@@ -165,19 +165,18 @@ RateControl::RateControl(const QString& group,
...
+    //     m_eRateRampMode = static_cast<RampMode>(
+    //         getConfig()->getValue(ConfigKey("[Controls]","RateRamp"),
+    //                               static_cast<int>(RampMode::Stepping)));
 
-//     // Set the Sensitivity
-//     m_iRateRampSensitivity =
-//             getConfig()->getValueString(ConfigKey("[Controls]","RateRampSensitivity")).toInt();
+    //     // Set the Sensitivity
+    //     m_iRateRampSensitivity =
+    //             getConfig()->getValueString(ConfigKey("[Controls]","RateRampSensitivity")).toInt();
```
acolombier, line 179: Shall we just kill this comment cruft? Looks like it's been there for 7 years...

src/skin/legacy/legacyskinparser.cpp
```diff
@@ -2306,7 +2307,9 @@ void LegacySkinParser::setupConnections(const QDomNode& node, WBaseWidget* pWidg
         ValueTransformer* pTransformer = nullptr;
         QDomElement transform = m_pContext->selectElement(con, "Transform");
         if (!transform.isNull()) {
-            pTransformer = ValueTransformer::parseFromXml(transform, *m_pContext);
+            pTransformer =
+                    ValueTransformer::parseFromXml(transform, *m_pContext)
+                            .get(); // TODO don't leak here
```
acolombier, line 2312: `TODO`?  Could it be possible to make `pTransformer` a smart pointer?

acolombier (commented):
This a great cleanup! Thanks for looking into this.

The CI seems to be failing, can you look at this?

src/skin/legacy/legacyskinparser.cpp
Swiftb0y: yes, but didn't want to here for the sake of keeping the scope small. Actually this is a bug, I want release here otherwise it'll be a double free...

src/util/valuetransformer.cpp
```diff
@@ -1,35 +1,124 @@
...
+            : m_compareValue(compareValue) {
+    }
+
+    double transform(double argument) const override {
+        return argument == m_compareValue;
+    }
+
+    double transformInverse(double argument) const override {
+        if (argument > 0.0) {
+            return m_compareValue;
```
daschuer, line 75: Is that correct? Maybe we should add a brief comment.

Swiftb0y: didn't bother checking, I just moved this part from the header into the cpp.

Swiftb0y:
actually I just now realized that QQueue is terribly inefficient in terms of memory use, so that'll take a little while to rectify.

src/util/compatibility/qatomic.h
```diff
@@ -9,24 +9,9 @@
...
-inline T* atomicLoadAcquire(const QAtomicPointer<T>& atomicPtr) {
-#if QT_VERSION >= QT_VERSION_CHECK(5, 14, 0)
-    return atomicPtr.loadAcquire();
-#else
-    return atomicPtr.load();
-#endif
-}
+#define DEPRECATED_QATOMIC_OPS /// Deprecated, use the corresponding member function directly
 
+DEPRECATED_QATOMIC_OPS
```
daschuer, line 14: What does this do? 
Did you consider to fix the code instead?

Swiftb0y: It just copy-pastes the message above so it's shown during code completion. I could fix this (since `main` is Qt6 only), but that would be touching another 20 files which I tried to avoid. I can do it though if you prefer.

daschuer: The preprocessor stops at comments. In my case Eclipse shows as macro expansion just an empty field. 
So I think you should remove this and the comment all together. From the function content it is clear that the function is deprecated, because the main branch is on QT6.  

It is OK to keep the current state with this wrapper since there is no user benefit to remove it.

src/engine/controls/ratecontrol.cpp
daschuer: Removing this comments is welcome.

src/util/rotary.h
```diff
@@ -1,40 +1,15 @@
...
-    int m_iFilterPos;
-    // Pointer to rotary filter buffer
-    std::vector<double> m_pFilter;
-    // Calibration value
-    double m_dCalibration;
-    // Last value
-    double m_dLastValue;
-    int m_iCalibrationCount;
+  private:
+    QQueue<double> m_filterHistory;
```
daschuer, line 14: QQueue is not effective for this prupose. I suggest to restore the `std::vector<double>` based solution. 
The `m_pFilter` can be called `m_pRingBuffer` to make the use cas obvious.

src/util/rotary.cpp
```diff
@@ -1,93 +1,24 @@
...
-
-Rotary::Rotary()
-        : m_iFilterLength(kiRotaryFilterMaxLen),
-          m_iFilterPos(0),
-          m_pFilter(m_iFilterLength),
-          m_dCalibration(1.0),
-          m_dLastValue(0.0),
-          m_iCalibrationCount(0) {
-    for (int i = 0; i < m_iFilterLength; ++i) {
-        m_pFilter[i] = 0.;
```
daschuer, line 15: Without that loop, the jog wheel is unfiltered for the first samples. This here assumes that the jog is not moving before the controller is initalized. Is it it a problem? The workarund was the removed fillBuffer() function.

src/util/class.h
```diff
@@ -1,7 +1,9 @@
 #pragma once
 
-// A macro to disallow the copy constructor and operator= functions
-// This should be used in the private: declarations for a class
+// DEPRECATED: Instead rely on "the rule of zero". See
+// https://en.cppreference.com/w/cpp/language/rule_of_three#Rule_of_zero
+// See also: Cpp Core Guidelines C.20, C.21 & C.22 as well as the clang-tidy rule
+// `modernize-use-equals-delete`
```
daschuer, line 6: I don't undestand the comment. The `modernize-use-equals-delete` explict cites this solution as good. 
You are probably refering the old impmentation of the macro. 
```suggestion
// DEPRECATED: Instead delete functions explicit without using this macro. 
```

src/util/rotary.h
Swiftb0y: yes, I did something similar in a follow up with I haven't pushed yet.

src/util/rotary.cpp
Swiftb0y: I'm not sure. If the loop is pre-filled with zeros then the mean is heavily distorted until those zeros have been flushed.

Swiftb0y: Actually, thinking about this more, this could be the reason why I have had the issue for years where moving the jog suddenly would be slow at first and then really fast after a couple split seconds. I will do some testing to confirm, but now I'm fairly certain that filling the buffer is wrong!

src/util/class.h
Swiftb0y: well what I actually want to say that people should reconsider using this macro as per the rule of zero. There are two problems:
1. Its used all over the place (more than is necessary nor good).
2. It violates [CppCoreGL C.21: If you define or =delete any copy, move, or destructor function, define or =delete them all](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all). 

So the best practice IMO is that if a class deals with ownership it should be its sole purpose (I think we both agree on that). If it does that, it should adhere to the rule of zero and do that explicitly (not via this macro or by mixing this macro). Classes that don't deal with ownership should rely on implicitly generated SMF's and not delete them explicitly (for example by using this macro). Considering all that, this macro should not exist in code following best practices and thus I mark it deprecated. 

The actual point however is not do it explicitly everywhere, but rather reconsider refactoring to follow the rule of zero, which is why I added all the references. So I don't agree with your suggested change because it does not link to the proper best practices.

> The modernize-use-equals-delete explict cites this solution as good.

I wanted to mention it so people aren't tempted to do it the pre C++11 way.

daschuer: OK, But than write an explicit advice what to do instead of this macro. A short for of your last comment?

Swiftb0y: Sure, I'll try my best.

src/skin/legacy/legacyskinparser.cpp
Swiftb0y: done

src/util/compatibility/qatomic.h
Swiftb0y: I have removed the comment and `#define`

src/util/rotary.h
Swiftb0y: see 7578cfaf14fcb054f4f8dd210fcabcd2a30f1c7c

src/util/class.h
Swiftb0y: done.

src/util/rotary.h
```diff
@@ -2,39 +2,25 @@
...
-    void calibrateStart();
-    // End calibration measurement
-    double calibrateEnd();
-    // Set calibration
-    void setCalibration(double c);
-    // Get calibration
-    double getCalibration();
+  public:
+    Rotary(qsizetype filterLength)
+            : m_filterHistory(filterLength, 0.0) {
```
Swiftb0y, line 14: note, that this does restore the previous behavior. I have not yet experimented with more dynamic filtering.

daschuer (approved):
LGTM

Swiftb0y:
@daschuer still lgtm with the two new commits?

src/util/versionstore.cpp
```diff
@@ -64,56 +80,58 @@ QString VersionStore::applicationName() {
...
+#define PLATFORM_STR "Wasm32"
 #elif defined(__wasm__)
-    base.append(" Wasm");
+#define PLATFORM_STR "Wasm"
 #endif
 
-    return base;
+QString VersionStore::platform() {
+    return QStringLiteral(OS_VERSION_STR "" PLATFORM_STR);
 }
```
Swiftb0y, line 135: whoops. a space got lost here

Swiftb0y: fixed.

Swiftb0y:
friendly ping

src/util/font.cpp
```diff
@@ -67,15 +67,13 @@ void FontUtils::initializeFonts(const QString& resourcePath) {
...
             QDir::NoDotAndDotDot | QDir::Files | QDir::Readable);
-    for (const QFileInfo& file : files) {
-        const QString& path = file.filePath();
-
+    for (const QString& path : files) {
         // Skip text files (e.g. license files). For all others we let Qt tell
         // us whether the font format is supported since there is no way to
         // check other than adding.
-        if (path.endsWith(".txt", Qt::CaseInsensitive)) {
+        if (path.endsWith(QStringLiteral(".txt"), Qt::CaseInsensitive)) {
```
daschuer, line 76: The original char[] verison is faster. Because the compare happens in the char domain.

daschuer (changes requested):

src/util/font.cpp
Swiftb0y: Potentially yes, but the bare char literal is actually even slower because it gets implicitly converted to a QString at runtime. The proper solution is
```suggestion
        if (path.endsWith(QLatin1StringView(".txt"), Qt::CaseInsensitive)) {
```
(according to compiler explorer](https://compiler-explorer.com/z/9sehWz55K)

Digging through the Qt source, the comparision eventually happens in the correct overload of `ucstricmp`:
https://github.com/qt/qtbase/blob/182604ae9ce5fac6c1c79bb3f21b5d50e6da7258/src/corelib/text/qstring.cpp#L1466-L1494

I simply can't judge which is faster in this case without measuring. I doubt it makes much of a difference. Regardless, the plain `".txt"` literal is worse than `QStringLiteral` and `QLatin1StringView`.

daschuer: Ah I see, I was wrong. Qt has only the "faster" `QLatin1StringView`overload for cases where a Latin1 string is already in memory. This saves the runtime conversion to UTF-16. Since `QStringLiteral`we no longer have such a conversion in the literal case. In this case it saves the  
> expand to UTF-16 via unpacking
https://github.com/qt/qtbase/blob/182604ae9ce5fac6c1c79bb3f21b5d50e6da7258/src/corelib/text/qstring.cpp#L588-596
As well. 

The compiler explorer results with a plain char array are surprising. Why is QString::fromUtf8 used. 
For my understanding plain char literals are ASCII and if we want UTF8 we need to add u8"... "

daschuer: It was: fromAscii in QT5 
https://github.com/qt/qtbase/blob/83f5c3c26a59849197bd8d3a661afc87893866ba/src/corelib/text/qstring.h#L807
now: fromUtf8
https://github.com/qt/qtbase/blob/182604ae9ce5fac6c1c79bb3f21b5d50e6da7258/src/corelib/text/qstring.h#L835

daschuer (approved):
LGTM. This can be merged after fixup. Thank you.

src/util/font.cpp
Swiftb0y: So you agree that the `QStringLiteral` is correct?

daschuer: Yes

daschuer:
Thanks

Swiftb0y:
Thanks for the review.

src/util/versionstore.cpp
```diff
@@ -64,56 +80,58 @@ QString VersionStore::applicationName() {
...
         defined(__IA32__) || defined(_M_IX86) || defined(_X86_) || \
         defined(__X86__) || defined(__I86__)
-    base.append(" x86");
+#define PLATFORM_STR "x86"
 #elif defined(IA64)
-    base.append(" IA64");
+#define PLATFORM_STR "IA64"
 #elif defined(__aarch64__)
-    base.append(" ARM64");
+#define PLATFORM_STR "IA64"
```
fwcd, line 116: Typo: Should be ARM64, see https://github.com/mixxxdj/mixxx/issues/14086

Issue #13633, opened 2024-09-06
Mixxx doesn't send Sysex or MIDI out to Launchpad X (custom script, macOS Sonoma ARM)


frankgrimy:
### Bug Description

Hi.

I've coded a script for a Novation Launchpad X controller on my Windows machine. As of now, it can start the controller's DAW mode (by sending Sysex), and it also sends Midi messages to it, to light some pads (based on the 4 decks play status), just fine.

Now, I'm trying to get that script working on my Macbook Air M1, but neither Sysex nor Midi messages seems to be sent to the device. I thought of this because I started the DAW mode manually with Pocket MIDI, and when pressing the configured pads, they perform the actions I've coded.

I wonder if this is a Mixxx bug, or some macOS permissions, but I've search some info about this and haven't found anything.

If you want to check the code, I'd probably open a repo for it soon, but as I said the script works fine on Windows.

### Version

2.4.1

### OS

macOS Sonoma ARM

Swiftb0y:
Hey there, thank you for filling your issue. In order to get to the bottom of this. We would like to inspect your log file. Please use [our guide to find it](https://github.com/mixxxdj/mixxx/wiki/Finding%20the%20Mixxx.log%20file) and then upload it here and we'll see if we can spot anything in the log. Thank you.

frankgrimy:
Hi, here's the log file.
[mixxx.log](https://github.com/user-attachments/files/16913704/mixxx.log)

frankgrimy:
Line 234 says that the controller is being opened for input
> Info [Controller] PortMidiController: Opening Launchpad X LPX DAW Out index 0 for input

But two lines in the log file on my windows machine say that the device was opened for input and output.
> Info [Controller] PortMidiController: Opening LPX MIDI index 2 for input
> Info [Controller] PortMidiController: Opening LPX MIDI index 6 for output

So I guess the output port isn't being configured properly.

PS: Windows name scheme for the DAW port is different from macOS, that LPX MIDI is the DAW connection.

frankgrimy:
Hi.
This bug might be happening because macOS lists Launchpad X with different names for input and output, while on Linux and Windows they have the same name.
Names for input and output are respectively:

`Launchpad X LPX DAW Out`
`Launchpad X LPX DAW In`

[...]

PR #13691, opened 2024-09-24
Fix Launchpad X detection on macOS

frankgrimy:
As stated on #13633, Mixxx wasn't configuring Novation Launchpad X correctly, because of a name mismatch between its input and output interfaces in macOS.

### DIFF
--- a/src/controllers/midi/portmidienumerator.cpp
+++ b/src/controllers/midi/portmidienumerator.cpp
@@ -88,6 +88,12 @@ bool namesMatchAllowableEdgeCases(const QString& input_name,
     if (input_name == "MIDIIN2 (Ableton Push)" && output_name == "MIDIOUT2 (Ableton Push)") {
         return true;
     }
+
+    // Novation Launchpad X (macOS)
+    if (input_name == "Launchpad X LPX DAW Out" && output_name == "Launchpad X LPX DAW In") {
+        return true;
+    }
+    
     return false;
 }
 


daschuer:
The code looks good, thank you. The is just a stray white space character reported by pre-commit.
This can even go to our stable 2.4 branch. 
Can you rebase this to 2.4? If you have a local working copy, you may:
`git reset upstream/2.4 --hard`
`git cherry-pick 81d1a8406ec`
`git push -f`
Than use the edit feature above to change the merge target. 

I'm addition we need you permission to distribute your changes along with Mixxx. 
Please sign: https://docs.google.com/a/mixxx.org/spreadsheet/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ

And comment here when done.

frankgrimy:
Well, it seems ok now. I've also signed the form.

daschuer:
Cool. Thank you. I would like to add you to the contributor list in the Mixxx about box. Which name should I use?

frankgrimy:
Just go with _Frank Grimy_

Swiftb0y:
I'm a little too late, but an alternative solution would be change the other heuristic code instead of stripping `input` and `output`, we should just strip `in` and `out`.

daschuer:
It is hard to predict property naming conventions anyway. I think a better solution would be to make this part of the mappings. Something like collecting this matching table from all found mapping files.

PR #13692, opened 2024-09-24
HID: Avoid repeated error messages from hid_write()/hid_read()

daschuer:
This is fixed the huge log file issue with #13660 for log spam in any disconnect case, without fixing the root cause in that particular case.

### DIFF
--- a/src/controllers/hid/hidiooutputreport.cpp
+++ b/src/controllers/hid/hidiooutputreport.cpp
@@ -15,6 +15,7 @@ constexpr size_t kMaxHidErrorMessageSize = 512;
 HidIoOutputReport::HidIoOutputReport(
         const quint8& reportId, const unsigned int& reportDataSize)
         : m_reportId(reportId),
+          m_hidWriteErrorLogged(false),
           m_possiblyUnsentDataCached(false),
           m_lastCachedDataSize(0) {
     // First byte must always contain the ReportID - also after swapping, therefore initialize both arrays
@@ -121,10 +122,16 @@ bool HidIoOutputReport::sendCachedData(QMutex* pHidDeviceAndPollMutex,
             reinterpret_cast<const unsigned char*>(m_lastSentData.constData()),
             m_lastSentData.size());
     if (result == -1) {
-        qCWarning(logOutput) << "Unable to send data to device :"
-                             << mixxx::convertWCStringToQString(
-                                        hid_error(pHidDevice),
-                                        kMaxHidErrorMessageSize);
+        if (!m_hidWriteErrorLogged) {
+            qCWarning(logOutput) << "Unable to send data to device :"
+                                 << mixxx::convertWCStringToQString(
+                                            hid_error(pHidDevice),
+                                            kMaxHidErrorMessageSize);
+            // Stop logging error messages if every hid_write() fails to avoid large log files
+            m_hidWriteErrorLogged = true;
+        }
+    } else {
+        m_hidWriteErrorLogged = false;
     }
 
     hidDeviceLock.unlock();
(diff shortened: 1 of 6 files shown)


acolombier (approved):
LGTM! Thanks

Do we think this is a sufficient mitigation for the log file size till we have a proper limiter in place?

src/controllers/hid/hidiooutputreport.cpp
```diff
@@ -121,10 +122,15 @@ bool HidIoOutputReport::sendCachedData(QMutex* pHidDeviceAndPollMutex,
...
-        qCWarning(logOutput) << "Unable to send data to device :"
-                             << mixxx::convertWCStringToQString(
-                                        hid_error(pHidDevice),
-                                        kMaxHidErrorMessageSize);
+        if (!m_hidWriteErrorLogged) {
+            qCWarning(logOutput) << "Unable to send data to device :"
+                                 << mixxx::convertWCStringToQString(
+                                            hid_error(pHidDevice),
+                                            kMaxHidErrorMessageSize);
+            m_hidWriteErrorLogged = true; // Don't log the same error again
```
JoergAtGithub, line 130: This will not wor as the comment describe. To ensure that only identical errors are inhibited, you must compare the error message string.

src/controllers/hid/hidiothread.cpp
```diff
@@ -99,16 +100,22 @@ void HidIoThread::pollBufferedInputReports() {
...
-                                              hid_error(m_pHidDevice),
-                                              kMaxHidErrorMessageSize);
             DEBUG_ASSERT(bytesRead == -1);
+            if (!m_hidReadErrorLogged) {
+                qCWarning(m_logOutput) << "Unable to read buffered HID InputReports from"
+                                       << m_deviceInfo.formatName() << ":"
+                                       << mixxx::convertWCStringToQString(
+                                                  hid_error(m_pHidDevice),
+                                                  kMaxHidErrorMessageSize);
+                m_hidReadErrorLogged = true; // Don't log the same error again
```
JoergAtGithub, line 110: Same here

JoergAtGithub (changes requested):

daschuer:
I have adjusted the error message. We want also stop log spam in the unlikely case that there are altering different error messages, because this would still fill the log file.

daschuer:
> Do we think this is a sufficient mitigation for the log file size till we have a proper limiter in place?

I consider this a permanent solution, because we also want to avoid the performance impact, until the 100 MB log limit is from https://github.com/mixxxdj/mixxx/pull/13684 is reached.

JoergAtGithub (commented):
hid_write is also used here: https://github.com/mixxxdj/mixxx/blob/d6eeb45ceeaa5878b06d1b3b86a9facdad08ac81/src/controllers/hid/hidioglobaloutputreportfifo.cpp#L68-L76

src/controllers/hid/hidiothread.cpp
```diff
@@ -99,16 +100,23 @@ void HidIoThread::pollBufferedInputReports() {
...
-                                   << mixxx::convertWCStringToQString(
-                                              hid_error(m_pHidDevice),
-                                              kMaxHidErrorMessageSize);
             DEBUG_ASSERT(bytesRead == -1);
+            if (!m_hidReadErrorLogged) {
+                qCWarning(m_logOutput) << "Unable to read buffered HID InputReports from"
+                                       << m_deviceInfo.formatName() << ":"
+                                       << mixxx::convertWCStringToQString(
+                                                  hid_error(m_pHidDevice),
+                                                  kMaxHidErrorMessageSize);
```
JoergAtGithub, line 109:
```suggestion
                qCWarning(m_logOutput) << "Unable to read buffered HID InputReports from"
                                       << m_deviceInfo.formatName() << ":"
                                       << mixxx::convertWCStringToQString(
                                                  hid_error(m_pHidDevice),
                                                  kMaxHidErrorMessageSize)
                                       << "Note that, this message is only logged once and may not appear again until all hid_read errors have disappeared.";
```

daschuer:
Done

JoergAtGithub:
Works as expected. Thank you!

PR #13693, opened 2024-09-24
Overview: clear pickup position display when opening cue menu

ronso0:
IMO it only adds noise, all relevant info like the cue position is displayed in the cue menu.

### DIFF
--- a/src/widget/woverview.cpp
+++ b/src/widget/woverview.cpp
@@ -592,6 +592,8 @@ void WOverview::mousePressEvent(QMouseEvent* e) {
                     m_pCurrentTrack->removeCue(pHoveredCue);
                     return;
                 } else {
+                    // Clear the pickup position display, we have all cue info in the menu.
+                    leaveEvent(nullptr);
                     m_pCueMenuPopup->setTrackCueGroup(m_pCurrentTrack, pHoveredCue, m_group);
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
                     m_pCueMenuPopup->popup(e->globalPosition().toPoint());


src/widget/woverview.cpp
```diff
@@ -592,6 +592,10 @@ void WOverview::mousePressEvent(QMouseEvent* e) {
                     m_pCurrentTrack->removeCue(pHoveredCue);
                     return;
                 } else {
+                    // Clear the pickup position display.
+                    // ev could actually be any event type, it's not used anyway.
+                    QEvent* ev = new QEvent(QEvent::HoverLeave);
+                    leaveEvent(ev);
```
daschuer, line 598: ev is leaked. This should do the trick:
```suggestion
                    QEvent ev = QEvent(QEvent::HoverLeave);
                    leaveEvent(&ev);
```

daschuer (changes requested):

src/widget/woverview.cpp
ronso0: Thank you.
I definitly have to do some homework wrt pointers.

ronso0: done.

ywwg: when in doubt, never write "new" :).

src/widget/woverview.cpp
```diff
@@ -592,6 +592,10 @@ void WOverview::mousePressEvent(QMouseEvent* e) {
                     m_pCurrentTrack->removeCue(pHoveredCue);
                     return;
                 } else {
+                    // Clear the pickup position display.
+                    // ev could actually be any event type, it's not used anyway.
```
ywwg, line 596: if it's not used can we pass nullptr?

ywwg (approved):

src/widget/woverview.cpp
ronso0: let's find out..

ronso0: works, ready for merge!

PR #13694, opened 2024-09-25
Tracks: allow searching for track id

ronso0:
Just a little helper for debugging.

Initially this was --developer only, but why restrict it? I think it might be helpful for (remote) debugging.
Wdyt?


daschuer:
There is no drawback for normal users, so I have no objections.

daschuer (approved):
LGTM, thank you.

Issue #13600, opened 2024-08-25
QML GUI is no longer working on Windows


JoergAtGithub:
### Bug Description

I get the followin error and the UI does not come up anymore:
```
warning [Main] file:///D:/mixxx-main-pr-test/install/x64__portable/qml/DeckRow.qml:47:13: QML PropertyChanges: Cannot assign to non-existent property "visible"
warning [Main] qrc:/qt-project.org/imports/Qt5Compat/GraphicalEffects/GaussianBlur.qml:258: TypeError: Property '_rebuildShaders' of object [null] is not a function
warning [Main] qrc:/qt-project.org/imports/Qt5Compat/GraphicalEffects/GaussianBlur.qml:258: TypeError: Property '_rebuildShaders' of object [null] is not a function
warning [Main] qrc:/qt-project.org/imports/Qt5Compat/GraphicalEffects/GaussianBlur.qml:258: TypeError: Property '_rebuildShaders' of object [null] is not a function
warning [Main] qrc:/qt-project.org/imports/Qt5Compat/GraphicalEffects/GaussianBlur.qml:258: TypeError: Property '_rebuildShaders' of object [null] is not a function
warning [Main] file:///D:/mixxx-main-pr-test/install/x64__portable/qml/DeckRow.qml:59: TypeError: Cannot read property 'left' of null
warning [Main] file:///D:/mixxx-main-pr-test/install/x64__portable/qml/DeckRow.qml:54: TypeError: Cannot read property 'height' of null
warning [Main] file:///D:/mixxx-main-pr-test/install/x64__portable/qml/ComboBox.qml:41: TypeError: Cannot read property 'width' of null
warning [Main] file:///D:/mixxx-main-pr-test/install/x64__portable/qml/DeckRow.qml:11: TypeError: Cannot read property 'height' of null
warning [Main] file:///D:/mixxx-main-pr-test/install/x64__portable/qml/DeckInfoBar.qml:209: TypeError: Cannot read property 'color' of null
warning [Main] file:///D:/mixxx-main-pr-test/install/x64__portable/qml/DeckInfoBar.qml:29: TypeError: Cannot read property 'coverArtUrl' of null
warning [Main] file:///D:/mixxx-main-pr-test/install/x64__portable/qml/DeckInfoBar.qml:98: TypeError: Cannot read property 'title' of null
[...]

PR #13696, opened 2024-09-25
set QQuickStyle to "basic"

m0dB:
Set QQuickStyle to "basic". The widget modifications in the QML code is not compatible with the macOS style, which triggers many errors and resulted in a black window on macOS.


JoergAtGithub:
This fixes #13600 ! Thank you!

PR #13698, opened 2024-09-27
(fix) disable Properties shortcut in Computer feature views

ronso0:
In Computer feature, 'Properties' is not added to the track menu, but Track Properties could still be invoked via hotkey.
We should disable that altogether in Computer because whatever is changed in Track Properties¹ is not shown in the (Computer) tracks view.

¹even though the selected track may already be in the library, if not will be added when loading the file to Track Properties


acolombier (approved):
LGTM!

daschuer (approved):
LGTM, thank you.

PR #13711, opened 2024-09-30
Fix writing metadata via symlink 

### LINKED ISSUE #13707: Mixxx destroys symlinks when writing metadata.
### Bug Description

A smylinked track can be loaded with Mixxx. When metadata is written, the symlink is replaced by the actual track. 

### Version

2.4.1

### OS

Ubuntu Focal

daschuer:
and add a test that verified the issue and the fix.

This fixes #13707 

Note: Mixxx does not support following Windows Shortcuts (*.lnk).

### DIFF
--- a/src/test/taglibtest.cpp
+++ b/src/test/taglibtest.cpp
@@ -84,6 +84,7 @@ TEST_F(TagLibTest, WriteID3v2TagViaLink) {
             MixxxTest::getOrInitTestDir().filePath(QStringLiteral("id3-test-data/empty.mp3")),
             tmpFileName);
 
+    // Access the MP3 file indirectly via a symlink when reading & writing the tags
     const QString linkFileName = tempDir.filePath("no_id3v1_mp3_link");
     EXPECT_TRUE(QFile::link(tmpFileName, linkFileName));
 
@@ -103,7 +104,7 @@ TEST_F(TagLibTest, WriteID3v2TagViaLink) {
 
     qDebug() << "Setting track title";
 
-    // Write metadata -> only an ID3v2 tag should be added
+    // Write metadata (via the symlink) -> only an ID3v2 tag should be added
     mixxx::TrackMetadata trackMetadata;
     trackMetadata.refTrackInfo().setTitle(QStringLiteral("title"));
     const auto exported =
@@ -122,9 +123,11 @@ TEST_F(TagLibTest, WriteID3v2TagViaLink) {
         EXPECT_FALSE(mixxx::taglib::hasAPETag(mpegFile));
     }
 
+    // Verify that the symlink still exists and still points to the correct file
     auto linkFileInfoAfter = QFileInfo(linkFileName);
 
     EXPECT_TRUE(linkFileInfoAfter.exists());
     EXPECT_EQ(linkFileInfoAfter.canonicalFilePath().toStdString(), tmpFileName.toStdString());
+    EXPECT_TRUE(linkFileInfoAfter.isSymLink());
 }
 #endif
--- a/src/test/taglibtest.cpp
+++ b/src/test/taglibtest.cpp
@@ -70,3 +70,61 @@ TEST_F(TagLibTest, WriteID3v2Tag) {
         EXPECT_FALSE(mixxx::taglib::hasAPETag(mpegFile));
     }
 }
+
+#ifndef __WINDOWS__ // Note: Following Windows links (*.lnk shortcuts) is not supported by Mixxx yet
+TEST_F(TagLibTest, WriteID3v2TagViaLink) {
+    QTemporaryDir tempDir;
+    ASSERT_TRUE(tempDir.isValid());
+
+    // Generate a file name for the temporary file
+    const QString tmpFileName = tempDir.filePath("no_id3v1_mp3");
+
+    // Create the temporary file by copying an existing file
+    mixxxtest::copyFile(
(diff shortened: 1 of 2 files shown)


src/test/taglibtest.cpp
```diff
@@ -70,3 +70,61 @@ TEST_F(TagLibTest, WriteID3v2Tag) {
...
+
+    const QString linkFileName = tempDir.filePath("no_id3v1_mp3_link");
+    EXPECT_TRUE(QFile::link(tmpFileName, linkFileName));
+
+    auto linkFileInfoBefore = QFileInfo(linkFileName);
+    EXPECT_TRUE(linkFileInfoBefore.exists());
+    EXPECT_TRUE(linkFileInfoBefore.isSymLink());
+    EXPECT_EQ(linkFileInfoBefore.canonicalFilePath().toStdString(), tmpFileName.toStdString());
+
+    // Verify that the file has no tags
```
cr7pt0gr4ph7, line 95: Suggestion: `// Verify that the file (accessed via the symlink) has no tags`

src/test/taglibtest.cpp
```diff
@@ -70,3 +70,61 @@ TEST_F(TagLibTest, WriteID3v2Tag) {
...
+        TagLib::MPEG::File mpegFile(
+                TAGLIB_FILENAME_FROM_QSTRING(linkFileName));
+        EXPECT_FALSE(mixxx::taglib::hasID3v1Tag(mpegFile));
+        EXPECT_FALSE(mixxx::taglib::hasID3v2Tag(mpegFile));
+        EXPECT_FALSE(mixxx::taglib::hasAPETag(mpegFile));
+    }
+
+    qDebug() << "Setting track title";
+
+    // Write metadata -> only an ID3v2 tag should be added
```
cr7pt0gr4ph7, line 106: Suggestion: `// Write metadata (via the symlink) -> ...`

src/test/taglibtest.cpp
```diff
@@ -70,3 +70,61 @@ TEST_F(TagLibTest, WriteID3v2Tag) {
...
+
+    // Generate a file name for the temporary file
+    const QString tmpFileName = tempDir.filePath("no_id3v1_mp3");
+
+    // Create the temporary file by copying an existing file
+    mixxxtest::copyFile(
+            MixxxTest::getOrInitTestDir().filePath(QStringLiteral("id3-test-data/empty.mp3")),
+            tmpFileName);
+
+    const QString linkFileName = tempDir.filePath("no_id3v1_mp3_link");
```
cr7pt0gr4ph7, line 87: Suggestion: `// Access the MP3 file indirectly via a symlink when reading & writing the tags`

src/test/taglibtest.cpp
```diff
@@ -70,3 +70,61 @@ TEST_F(TagLibTest, WriteID3v2Tag) {
...
+    // Check that the file only has an ID3v2 tag after writing metadata
+    {
+        TagLib::MPEG::File mpegFile(
+                TAGLIB_FILENAME_FROM_QSTRING(linkFileName));
+        EXPECT_FALSE(mixxx::taglib::hasID3v1Tag(mpegFile));
+        EXPECT_TRUE(mixxx::taglib::hasID3v2Tag(mpegFile));
+        EXPECT_FALSE(mixxx::taglib::hasAPETag(mpegFile));
+    }
+
+    auto linkFileInfoAfter = QFileInfo(linkFileName);
```
cr7pt0gr4ph7, line 125: Suggestion: `// Verify that the symlink still exists and still points to the correct file`

src/test/taglibtest.cpp
```diff
@@ -70,3 +70,61 @@ TEST_F(TagLibTest, WriteID3v2Tag) {
...
+
+    EXPECT_TRUE(linkFileInfoAfter.exists());
+    EXPECT_EQ(linkFileInfoAfter.canonicalFilePath().toStdString(), tmpFileName.toStdString());
```
cr7pt0gr4ph7, line 128: Shouldn't we also verify for completeness sake that `linkFileName` is, in fact, still a symlink, and has not been overwritten?
```cpp
EXPECT_TRUE(linkFileInfoAfter.isSymLink());
```

cr7pt0gr4ph7 (approved):
A few small nitpicks, otherwise LGTM.

cr7pt0gr4ph7 (approved):

acolombier (approved):
Code LGTM - didn't test but change looked small enough so merging.

Thanks!

doodlebro:
I think there's an incorrect assumption being made in this solution with respect to how many symlinks are followed consecutively, which led to me laughing heartily in ZFS snapshots and turning off metadata sync in Mixxx to fix temporarily.

You can symlink to a symlink. I do this for some mixes which have re-used tracks from old ones. This Mixxx logic handling metadata sync back to symlinks stops after one level on 2.6 beta 6. In my case (Mac accessing files via SMB) this turned my multilevel symlinks into copies of files when metadata was written. Not great, definitely a footgun. When this happens, other file metadata (extended attributes) attached to the original file that the symlink pointed are also lost.

ronso0:
[According to the documenation, `QFileInfo::canonicalPath()`](https://doc.qt.io/qt-6/qfileinfo.html#canonicalPath) should already resolve all levels of symlinks in a given file path.

(compared to [`QFileInfo::filesystemSymLinkTarget()`](https://doc.qt.io/qt-6/qfileinfo.html#filesystemSymLinkTarget) which only resolves the next link level.)

Could this be an issue with your SMB permissions / configuration?
CC @daschuer

PR #13721, opened 2024-10-06
MIDI for light:  add settings ui

ekordas:
Adds Setting definition for UI of the Midi for Light Script

### DIFF
--- a/res/controllers/Midi_for_light-scripts.js
+++ b/res/controllers/Midi_for_light-scripts.js
@@ -16,33 +16,33 @@ function midi_for_light() {}
 ///////////////////////////////////////////////////////////////
 //                       USER OPTIONS                        //
 ///////////////////////////////////////////////////////////////
-var midi_channel = 1; // set midi_channel. Valid range: 1 to 16.
-var enable_beat = true; // set to false if you not need beat
-var enable_bpm = true; // set to false if you not need BPM
-var enable_mtc_timecode = false; // set to false if you not need midi time code
-var enable_vu_mono_current = false; // set to false if you not need VU mono current
-var enable_vu_mono_average_min = false; // set to false if you not need VU mono average min
-var enable_vu_mono_average_mid = false; // set to false if you not need VU mono average mid
-var enable_vu_mono_average_max = false; // set to false if you not need VU mono average max
-var enable_vu_mono_average_fit = true; // set to false if you not need VU mono average fit
-var enable_vu_mono_current_meter = false; // set to false if you not need VU mono current meter
-var enable_vu_mono_average_meter = true; // set to false if you not need VU mono average meter
-var enable_vu_left_current = false; // set to false if you not need VU left current
-var enable_vu_left_average_min = false; // set to false if you not need VU left average min
-var enable_vu_left_average_mid = false; // set to false if you not need VU left average mid
-var enable_vu_left_average_max = false; // set to false if you not need VU left average max
-var enable_vu_left_average_fit = true; // set to false if you not need VU left average fit
-var enable_vu_left_current_meter = false; // set to false if you not need VU left current meter
-var enable_vu_left_average_meter = false; // set to false if you not need VU left average meter
-var enable_vu_right_current = false; // set to false if you not need VU right current
-var enable_vu_right_average_min = false; // set to false if you not need VU right average min
-var enable_vu_right_average_mid = false; // set to false if you not need VU right average mid
-var enable_vu_right_average_max = false; // set to false if you not need VU right average max
-var enable_vu_right_average_fit = true; // set to false if you not need VU right average fit
-var enable_vu_right_current_meter = false; // set to false if you not need VU right current meter
-var enable_vu_right_average_meter = false; // set to false if you not need VU right average meter
-var deck_ending_time = 15; // set a time (in seconds) in which the playing track is considered to be ending
-var deck_ending_priority_factor = 0.9; // decrease the priority of the ending track by this factor
+var midi_channel = engine.getSetting("midi_channel"); // set midi_channel. Valid range: 1 to 16.
+var enable_beat = engine.getSetting("enable_beat"); // set to false if you not need beat
+var enable_bpm = engine.getSetting("enable_bpm"); // set to false if you not need BPM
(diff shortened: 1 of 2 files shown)


ronso0:
Thank you for this PR.

Do you mind attaching a screenshot of the new settings?

ekordas:
I allready signed the agreement.

Here is a screenshot:
midi_for_light-settings

ronso0:
Okay, thank you, I didn't remember you already worked on MIDI for lights.

The VU checkboxes could use more human-readable labels, not just the internal variable names, no?

I didn't work with the controller settings, yet, wasn't there an option/flag that would arange group items horizontally if the preference page allows it?

ekordas:
I didnt work with the controller settings either, If there is a documentation for it i can set other flags. Leaving the Orientation to unset flipps checkboxes to a row while resizing the settings wich is really destroying the layout:
Bildschirmfoto vom 2024-10-07 00-00-49
So i leave it to vertical.
The Labels are changed in the next commit.

JoergAtGithub:
> If there is a documentation for it i can set other flags.

@acolombier Do we have this described in the manual?

The best reference is the mapping for the S4 Mk3: https://github.com/mixxxdj/mixxx/pull/12995

ekordas:
The Mapping xml file is described at https://github.com/mixxxdj/mixxx/wiki/MIDI%20controller%20mapping%20file%20format but not the ui Stuff. I Used the diff mentioned as reference.

ekordas:
The <row orientation="vertical"> seems to do the intended behavior, but if the vertical list of checkboxes is flipped to a row  it does not flip back. And it does not matter how slim you resize the window. 
The section above does change to a row if the Size allows it and back, but it seems there is a Problem with a long list of checkboxes.
Since the UI is more aimed at developers and technical applications I personally would not bother making it pretty to much. It only requires a little scrolling.
Following is a new attempt:
Bildschirmfoto vom 2024-10-07 00-52-49

ekordas:
Is anything missing or is everyone OK with the design?

ronso0:
I'm not familiar with controller settings.
@JoergAtGithub wdyt?

JoergAtGithub:
The settings syntax looks goot to me. But I'm not familar with MIDI for Light.

res/controllers/Midi_for_light-scripts.js
```diff
@@ -16,33 +16,33 @@ function midi_for_light() {}
...
-var enable_vu_right_current = false; // set to false if you not need VU right current
-var enable_vu_right_average_min = false; // set to false if you not need VU right average min
-var enable_vu_right_average_mid = false; // set to false if you not need VU right average mid
-var enable_vu_right_average_max = false; // set to false if you not need VU right average max
-var enable_vu_right_average_fit = true; // set to false if you not need VU right average fit
-var enable_vu_right_current_meter = false; // set to false if you not need VU right current meter
-var enable_vu_right_average_meter = false; // set to false if you not need VU right average meter
-var deck_ending_time = 15; // set a time (in seconds) in which the playing track is considered to be ending
-var deck_ending_priority_factor = 0.9; // decrease the priority of the ending track by this factor
+var midi_channel = engine.getSetting("midi_channel"); // set midi_channel. Valid range: 1 to 16.
```
fwcd, line 19: In previous mappings we've "defensively" added the default values here too (either using the nullish coalescing operator `??` or `||`). I don't think that should be necessary, so this looks good to me, but just to confirm, is this safe @acolombier?

JoergAtGithub: AFAIK, the default values there were only necessarry for backward compatibility of the mapping to 2.4 without the controller settings feature.

fwcd: Ah interesting, so `engine.getSetting` was already available in 2.4, but invocations always returned `undefined`?

JoergAtGithub: It does not exist in 2.4, and is therefore `undefined`.

fwcd: You can't call `undefined` without getting a `TypeError` though.

ekordas: I can confirm the mapping does not work with the proposed fallback. Changing the line to
`var midi_channel = engine.getSetting("midi_channel") || 1;`
causes an error in Version 2.4:
`Uncaught exception: file:///C:/Program Files/Mixxx/controllers/Midi_for_light-scripts.js:19: TypeError: Property 'getSetting' of object ControllerScriptInterfaceLegacy(0x16b9d336520) is not a function`

ekordas: `var midi_channel = engine.getSetting ? engine.getSetting("midi_channel") : 1;` seems to work

fwcd: Depending on how new the JS engine is you might be able to do

```ts
engine.getSetting?.("midi_channel") ?? 1
```

But tbh if we're merging this to 2.5, it shouldn't be needed anyway (and I don't think there'll be many 2.4 releases anymore apart from minor bug fixes)

acolombier:
I have added some [documentation in the wiki](https://github.com/mixxxdj/mixxx/wiki/MIDI-Scripting#settings-api), hopefully this can help refining the UI a bit.

ekordas:
So, should I
a) implement a fallback and re target to 2.4 (previous underlying patch was considered to go into 2.5)
b) leave target to 2.4
and
1) retouch the ui? Personally i am quite happy now...
2) you have some style issues to fix?
3) leave as is?

fwcd:
I'd say targeting 2.5 is fine, along with keeping everything as is. From a quick glance the PR looks good, I'll see if I find some time to test it soon.

fwcd:
I have tested this in QLC+[^1] and it seems to work well. Thanks!

[^1]: Just in case anyone else wants to experiment with this, here's a workspace that maps every MIDI control (it's a bit fiddly to do that manually): [Mixxx MIDI.qxw.zip](https://github.com/user-attachments/files/18055597/Mixxx.MIDI.qxw.zip) This is configured to use a device named "Loop", which is a small virtual MIDI device created by another small tool I wrote: https://github.com/fwcd/midiloop

PR #13724, opened 2024-10-07
(POC) LateNight: allow to show 4 decks in a row

ronso0:
A fun (two-) coffee break hack.
Looks okay, please test and let me know what you think.

image

### TODO
- [x] apply to mini decks, too (library maximized)
- [ ] adjust Classic theme

### DIFF
--- a/res/skins/LateNight/decks/decks_left.xml
+++ b/res/skins/LateNight/decks/decks_left.xml
@@ -25,5 +25,41 @@
       </WidgetGroup>
 
     </Children>
+    <Connection>
+      <ConfigKey>[Skin],show_4decks_row</ConfigKey>
+      <BindProperty>visible</BindProperty>
+      <Transform><Not/></Transform>
+    </Connection>
+  </WidgetGroup>
+
+  <WidgetGroup>
+    <ObjectName>DecksLeft</ObjectName>
+    <Layout>horizontal</Layout>
+    <SizePolicy>me,min</SizePolicy>
+    <Children>
+
+      <WidgetGroup>
+        <Layout>vertical</Layout>
+        <SizePolicy>me,min</SizePolicy>
+        <Children>
+          <SingletonContainer>
+            <ObjectName>Deck3_Src</ObjectName>
+          </SingletonContainer>
+        </Children>
+        <Connection>
+          <ConfigKey>[Skin],show_4decks</ConfigKey>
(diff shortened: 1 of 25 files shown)


Eve00000:
very nice!
only 2 coffees?

ronso0:
Thanks. Nice, but not polished yet.
I'll try to move more components into templates and not add much complexity.

Yes, two flat white with sugar and a glass of water.

ronso0:
FWIW, changing the deck order (3 1 2 4 or 1 2 3 4) can be done with color schemes and deck num variables, ie. requires skin reload to update the deck numbers in each deck/waveform template.

For LateNight we'd then have
* PaleMoon (Decks 3 1 2 4) (default)
* PaleMoon (Decks 1 2 3 4)
* Classic (Decks 3 1 2 4)
* Classic (Decks 1 2 3 4)

with almost exact duplicate themes.

Idk if that's worth it. Is this a desired feature? Idk.
If yes, the color scheme solution is definitely easier than adding a skin option (Interface preferences), with all the safety checks, can it be applied (does skin support 4 decks?) and what not.

daschuer:
The demand for 1234 is only sourced by certain controllers. With "wrong" labeling. right?  Is switching between 3124 and 1234 a mapping feature? We need probably also swap the deck xfader orientation. Do we know which controller are affected?

ronso0:
> The demand for 1234 is only sourced by certain controllers. With "wrong" labeling. right?

Yes, probably. And we don't have deck number labels, but we do have the "LoadToDeck#" controls, and with external mixers one has to map mixer <--> Mixxx numbers twice: for the mapping and for the deck sound I/O configuration. Needs to be done only once, so yeah, maybe it's too much complexity considering the little demand.

daschuer:
> Those people could only use Deck 1 & 2,, Now with a big screen and 1 2 3 4 they can use all 4 of them. They will be grateful.

I was not aware that a certain order prevents people from using deck 3 and 4. In general we should make all users feel at home with Mixxx. Unfortunately the issue is more complex that it may seem. We need to consider switching from:
```
1 2 
3 4 
``` 
To 
````
1234
````
And take 1234 and 3124 labeled controllers into account. I consider a wrong labeling more a mental issue than a surprising order. So how about file an issue for that and solve it independently for all skins?

dmt-hub:
Definitely appreciated. Thank you.

PR #13730, opened 2024-10-08
Refactor: `MovingInterquartileMean`

Swiftb0y:
what was initially just supposed to be a cleanup/refactor, mutated into a full blown optimization. This is primarily intended as a cleanup though, and I only implemented the benchmark to show that this was not causing regressions in performance. There is still a little more to be done, but these are some preliminary numbers.
Before:
```
-----------------------------------------------------------------------------
Benchmark                   Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------
BM_insertion/16         17485 ns        16857 ns            1 items_per_second=3.79664M/s
BM_insertion/64         20922 ns        20926 ns            1 items_per_second=12.2336M/s
BM_insertion/512       774264 ns       770344 ns            1 items_per_second=2.65855M/s
BM_insertion/4096    55161004 ns     54684805 ns            1 items_per_second=299.608k/s
BM_insertion/16384  891138337 ns    886767771 ns            1 items_per_second=73.9044k/s
```
After:
```
-----------------------------------------------------------------------------
Benchmark                   Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------
BM_insertion/16         10458 ns         9738 ns            1 items_per_second=6.57219M/s
BM_insertion/64         16030 ns        15969 ns            1 items_per_second=16.0311M/s
BM_insertion/512       745585 ns       735096 ns            1 items_per_second=2.78603M/s
BM_insertion/4096    40907286 ns     40517411 ns            1 items_per_second=404.369k/s
BM_insertion/16384  644210901 ns    641245320 ns            1 items_per_second=102.201k/s
```

### DIFF
--- a/src/test/movinginterquartilemean_test.cpp
+++ b/src/test/movinginterquartilemean_test.cpp
@@ -1,7 +1,10 @@
-#include <gtest/gtest.h>
-#include <QDebug>
-
 #include "util/movinginterquartilemean.h"
+
+#include <benchmark/benchmark.h>
+#include <gtest/gtest.h>
+
+#include <QDebug>
+#include <random>
 
 namespace {
 
@@ -109,4 +112,28 @@ TEST_F(MovingInterquartileMeanTest, doubles9) {
     }
 }
 
+void BM_MovingIQM_Insertion(benchmark::State& state) {
+    std::mt19937 gen; // explicitly don't seed for reproducibility
+    std::uniform_real_distribution<> dis(0.0, 1.0);
+    // first quarter of the test will be plain insertion
+    // the remaining three quarters will evict the oldest value
+    std::size_t num_iters = state.range(0) * 4;
+    for (auto _ : state) {
+        MovingInterquartileMean iqm(state.range(0));
+        for (double i = 0; i < num_iters; ++i) {
+            benchmark::DoNotOptimize(iqm.insert(dis(gen)));
+        }
+    }
+    state.SetItemsProcessed(state.iterations() * num_iters);
+}
+
+BENCHMARK(BM_MovingIQM_Insertion)
+        ->RangeMultiplier(2)
+        ->Range(1 << 1, 1 << 10)
+        // the benchmark is so slow, it usually only gets a single iteration
+        // each, so manually force a couple more
+        ->Repetitions(100)
+        // don't spam the output with the individual repetition data
+        ->DisplayAggregatesOnly()
+        ->Unit(benchmark::kMicrosecond);
 }  // namespace
--- a/src/util/movinginterquartilemean.cpp
+++ b/src/util/movinginterquartilemean.cpp
@@ -1,46 +1,41 @@
 #include "movinginterquartilemean.h"
 
-MovingInterquartileMean::MovingInterquartileMean(const unsigned int listMaxSize)
-        : m_dMean(0.0),
-          m_iListMaxSize(listMaxSize),
-          m_bChanged(true) {
-}
+#include <algorithm>
+#include <cstddef>
+#include <iterator>
+#include <numeric>
 
-MovingInterquartileMean::~MovingInterquartileMean() {};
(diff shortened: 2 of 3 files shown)


Swiftb0y:
So I added yet another optimization (custom `replace_sorted` algorithm instead of `erase`+`insert`) which adds further speedups, especially with realistic IQM sizes.

MovingIQM Benchmark graph

Please review.

Swiftb0y:
I decided to drop `replace_sorted` as it introduced quite a lot of extra code without much benefit. This should now quite drastically simplify the code while still providing a speedup.

src/util/movinginterquartilemean.cpp
```diff
@@ -1,46 +1,44 @@
...
     return mean();
 }
 
 void MovingInterquartileMean::clear() {
     m_bChanged = true;
-    m_queue.clear();
+    // std::queue has no .clear(), so creating a temporary and std::swap is the
+    // next most elegant solution
+    std::queue<double> empty;
+    std::swap(m_queue, empty);
```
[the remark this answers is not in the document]
Swiftb0y: good catch, yeah that should work.

JoergAtGithub:
LGTM! Thank you!

JoergAtGithub (approved):

Swiftb0y:
just noticed I still had some unpushed commits in my tree, it would be nice if could take a look at those too. Thank you.

src/test/movinginterquartilemean_test.cpp
```diff
@@ -109,4 +112,28 @@ TEST_F(MovingInterquartileMeanTest, doubles9) {
     }
 }
 
+void BM_MovingIQM_Insertion(benchmark::State& state) {
+    std::mt19937 gen; // explicitly don't seed for reproducibility
```
JoergAtGithub, line 116: Why make it not reproducible? This makes future debbuging really hard.
In my opinion a test case shall always use the same data. If you want increase the coverage, you should increase the test data size instead of making it random.

Swiftb0y: I 100% agree. Its explicitly **not** initialized with a seed (usually you would leverage `std::random_device` for that). So its seed is implicitly 0 (which I could make explicit I guess), so it will always generate the same sequence of numbers and is thus reproducible.
Thinking about this `uniform_real_distribution` is not really realistic, instead it `normal_distribution` would likely make more sense at least when thinking about the TapFilter usecase specifically.

JoergAtGithub: Ah, than I just misunderstood the wording

Swiftb0y: See the [doc for the mersenne_twister_engine](https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine/mersenne_twister_engine):

> The default constructor.
> * If the default-constructed engine is of type std::mt19937, the 10000th consecutive invocation of it produces the value 4123659995.

JoergAtGithub:
LGTM! Please squash the fixups!

Swiftb0y:
Done. Thank you.

src/util/movinginterquartilemean.cpp
```diff
@@ -1,95 +1,85 @@
 #include "movinginterquartilemean.h"
 
-MovingInterquartileMean::MovingInterquartileMean(const unsigned int listMaxSize)
-        : m_dMean(0.0),
-          m_iListMaxSize(listMaxSize),
-          m_bChanged(true) {
-}
+#include <qlogging.h>
```
Swiftb0y, line 4: yikes, `clangd` f-ing me over again.

Swiftb0y: done

JoergAtGithub:
Thank you!

Swiftb0y:
Thank you.

Issue #13727, opened 2024-10-07
Metronome effect sometimes indefinitely repeats last buffer when releasing hotcue during at certain timing


Swiftb0y:
### Bug Description

Just filing this right now quickly so we don't forget. I suspects its a regression from #13636.

### Version

2.4.1-73-g8cb86e9449 (local-patches)

### OS

Fedora 41

daschuer:
Fixed here: 
https://github.com/mixxxdj/mixxx/pull/13733

PR #13733, opened 2024-10-08
Metronome: Fix noise when pausing track in in the middle of the click sound. 

daschuer:
Continue to play started metronome click even if track is paused.
This fixes repeating the same buffer over and over again causing loud noise.

https://github.com/mixxxdj/mixxx/issues/13727


Swiftb0y (approved):
LGTM. Thank you. will rebase #13654 shortly (for which I would appreciate a review).

PR #13735, opened 2024-10-08
Prevent deadlock with BULK transfert and reduce log noise

acolombier:
- Set a timeout (currently 5sec) to prevent deadlock (occurred with the S4Mk3, as the device seems to have froze, leading to the Mixxx controller thread to starve.)
- Make success log message only displayed when using `--controller-debug`

### DIFF
--- a/src/controllers/bulk/bulkcontroller.cpp
+++ b/src/controllers/bulk/bulkcontroller.cpp
@@ -299,11 +299,12 @@ void BulkController::sendBytes(const QByteArray& data) {
             (unsigned char*)data.constData(),
             data.size(),
             &transferred,
-            0);
+            5000 // Send timeout in milliseconds
+    );
     if (ret < 0) {
         qCWarning(m_logOutput) << "Unable to send data to" << getName()
                                << "serial #" << m_sUID << "-" << libusb_error_name(ret);
-    } else {
+    } else if (CmdlineArgs::Instance().getControllerDebug()) {
         qCDebug(m_logOutput) << transferred << "bytes sent to" << getName()
                              << "serial #" << m_sUID;
     }


JoergAtGithub:
In #13622 we have a open discussion because of the interface number for bulk already

acolombier:
Ah yeah, completely forgot about that one. I guess we might still want to get that PR for timeout and error detection fix?

JoergAtGithub:
I think this one is not in conflict with #13622

acolombier:
Realised that the signature change makes more sense to move into  #13737 as it used in there to stop upon comm failure.

Repurposing the PR for only timeout and spamming log messages

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -5,9 +5,14 @@
 #include "controllers/bulk/bulksupported.h"
 #include "controllers/defs_controllers.h"
 #include "moc_bulkcontroller.cpp"
+#include "util/cmdlineargs.h"
 #include "util/time.h"
 #include "util/trace.h"
 
+namespace {
+constexpr int kBulkSendTimeout = 5000; // In milliseconds
```
Swiftb0y, line 13: lets leverage `<chrono>`.
```suggestion
constexpr std::seconds kBulkSendTimeout = 5s;
```

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -299,11 +304,12 @@ void BulkController::sendBytes(const QByteArray& data) {
             (unsigned char*)data.constData(),
             data.size(),
             &transferred,
-            0);
+            kBulkSendTimeout);
```
Swiftb0y, line 307:
```suggestion
            std::milliseconds(kBulkSendTimeout).count();
```

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -299,11 +304,12 @@ void BulkController::sendBytes(const QByteArray& data) {
...
     if (ret < 0) {
         qCWarning(m_logOutput) << "Unable to send data to" << getName()
                                << "serial #" << m_sUID << "-" << libusb_error_name(ret);
-    } else {
+        ;
```
Swiftb0y, line 311: stray semicolon.
```suggestion
```

src/controllers/bulk/bulkcontroller.cpp
acolombier: What's the aim to do that? Appreciate this is all constexpr and won't have an impact on CPU cycles, but why would we to effectively do a double cast? (int -> std::milliseconds -> int)
If it's for readability, how about I rename `kBulkSendTimeoutMillsecond` or something like this?

acolombier: Ah no, you changed `kBulkSendTimeout` - find then. This still feels redundant IMO, but I'm happy to aim to use as much `std` as possible.

acolombier:
@Swiftb0y the code suggestion doesn't seem to work. Any suggestion to fix it?

src/controllers/bulk/bulkcontroller.cpp
```diff
@@ -5,9 +5,15 @@
 #include "controllers/bulk/bulksupported.h"
 #include "controllers/defs_controllers.h"
 #include "moc_bulkcontroller.cpp"
+#include "util/cmdlineargs.h"
 #include "util/time.h"
 #include "util/trace.h"
 
+namespace {
+using namespace std::literals;
+constexpr std::seconds kBulkSendTimeout = 5s;
```
daschuer, line 14: I think it is: 
```
    using namespace std::literals::chrono_literals; 
    constexpr std::chrono::seconds kBulkSendTimeout = 5s;
```
However the struggle here, confirms IMHO that using 5s is a bit over engineered here when not using this for chronio functions. So I would release the compiler from some work and use use the libusb type without a cast:
```
constexpr unsigned int kBulkSendTimeoutMs = 5000;
```
The Ms prefix documents the value unit. 
Both is fine for me.

src/controllers/bulk/bulkcontroller.cpp
Swiftb0y: well, its about using semantic types. If there is type to represent the data, then prefer to use that type instead of a primitive. Otherwise its too easy to accidentally end up with garbage calculations like `seconds * donuts + earthRadius` (because the semantic type should prevent you trying to do garbage operations like that). 

Fortunately, do to constexpr, the resulting binary should be same.

Swiftb0y:
whoops, yeah. Its `std::chrono::unit`.
```cpp
#include <chrono>

using namespace std::literals; 
constexpr std::chrono::seconds kBulkSendTimeout = 5s;

int usage() {
    return std::chrono::milliseconds(kBulkSendTimeout).count();
}
```

src/controllers/bulk/bulkcontroller.cpp
Swiftb0y: I disagree honestly. The added compile time is negligible. This is about avoiding [primitive obsession](https://refactoring.guru/smells/primitive-obsession).

src/controllers/bulk/bulkcontroller.cpp
acolombier: `std::milliseconds(kBulkSendTimeout).count() * donuts + earthRadius` would still be possible.
I get the point, but I'm just saying for a value that gets defined in anonymous namespace and is referenced once as input to a function that wants a `uint`, I don't think it is adding much value to semantic or type safety. If that variable was coming for the function input, he would definitely make sense!

src/controllers/bulk/bulkcontroller.cpp
acolombier: The resource you linked make sense, but in our case, we don't own the API and simply must comply to libusb asking for a tasteless `uint`.
To be honest, I'm leaning towards @daschuer's suggestion as I find it more readable - `std::milliseconds(kBulkSendTimeout).count()` feels a bit cryptic to me (had to go to the std lib to see what `count` was doing). Appreciate you are trying to roll out std on time within the Mixxx API tho so can't make up my mind. 
Would a @mixxxdj/developers help us with the tie here?

JoergAtGithub: In general I'm in favor for these literal types, and we should all get common to this std C++ syntax, as libraries like Qt convert their timing types towards this standard.
In this case, the literal type give us no type safety, as it's converted to the raw integer type before use.
Regarding readability, where a const value is only used once in the whole code - inline with the unit as comment is IMHO the most readable style:
```
c_function(5000, // Send timeout in milliseconds
           otherArgument); 
```

acolombier: Thanks @JoergAtGithub , that does sound like a great middle ground. @daschuer @Swiftb0y happy with that option?

Swiftb0y: >  In this case, the literal type give us no type safety, as it's converted to the raw integer type before use.

Yeah thats unfortunate, but the conversion is done at the API boundary (so its like calling `.get()` on a unique_ptr when passing to a C API. I still prefer my original solution (an unnamed constant is not much better), but I won't insist on it. Just my two cents for what is best in the general case. This is only a one-off so as you rightfully pointed out, it doesn't make much difference.

src/controllers/bulk/bulkcontroller.cpp
Swiftb0y: > std::milliseconds(kBulkSendTimeout).count() * donuts + earthRadius would still be possible.

Yes, but its much easier to spot that something fishy is going on here.

Swiftb0y (commented):
not blocking this any further. please fix pre-commit and clang-tidy and then its LGTM.

src/controllers/bulk/bulkcontroller.cpp
daschuer: As I already mentioned, all suggestions are fine for me as well. 

I can confirm the original request of @Swiftb0y to replace a comment by code.

Comment difficult code, but avoid comments by making the code itself expressive when possible.

Swiftb0y (approved):
:shrug:

daschuer (approved):
Thank you.LGTM

PR #13736, opened 2024-10-08
fix: trigger QML waveform slot at init

acolombier:
This fix the bug where the waveform isn't displayed when loading QML with a track already loaded (can be reproduced by loading a track, and trigger a QML reload with a `touch` on any used QMl file, e.g `res/qml/main.qml`)


Swiftb0y (approved):
LGTM. Thank you.

PR #4523, opened 2021-11-15
Controller logging categories (with --controllerDebug support)

Holzhaus:
This is the abandoned PR #4522 with support for the `--controllerDebug` flag added.

PR #13737, opened 2024-10-08
feat: improve screen rendering framework

acolombier:
This is a refactor largely inspired from the work in #13459 to simplify the hook system used for screen rendering.

### DIFF
--- a/src/controllers/controller.h
+++ b/src/controllers/controller.h
@@ -141,8 +141,9 @@ class Controller : public QObject {
 
   public:
     // This must be reimplemented by sub-classes desiring to send raw bytes to a
-    // controller.
-    virtual void sendBytes(const QByteArray& data) = 0;
+    // controller. Return true in case of successful completion, false in case
+    // of partial completion or failure.
+    virtual bool sendBytes(const QByteArray& data) = 0;
 
   private: // but used by ControllerManager
 
--- a/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
@@ -1,5 +1,7 @@
 #include "controllers/scripting/legacy/controllerscriptenginelegacy.h"
 
+#include <memory>
+
 #ifdef MIXXX_USE_QML
 #include <QDirIterator>
 #include <QQmlEngine>
@@ -11,36 +13,17 @@
 
 #include "control/controlobject.h"
 #include "controllers/controller.h"
-#ifdef MIXXX_USE_QML
-#include "controllers/rendering/controllerrenderingengine.h"
-#endif
 #include "controllers/scripting/colormapperjsproxy.h"
 #include "controllers/scripting/legacy/controllerscriptinterfacelegacy.h"
 #include "errordialoghandler.h"
 #include "mixer/playermanager.h"
 #include "moc_controllerscriptenginelegacy.cpp"
 #ifdef MIXXX_USE_QML
+#include "qml/qmlmixxxcontrollerscreen.h"
 #include "util/assert.h"
 #include "util/cmdlineargs.h"
 
 using Clock = std::chrono::steady_clock;
-#endif
-
-#ifdef MIXXX_USE_QML
-namespace {
-const QByteArray kScreenTransformFunctionUntypedSignature =
-        QMetaObject::normalizedSignature(
-                "transformFrame(QVariant,QVariant)");
-const QByteArray kScreenTransformFunctionTypedSignature =
-        QMetaObject::normalizedSignature("transformFrame(QVariant,QDateTime)");
-const QByteArray kScreenInitFunctionUntypedSignature =
-        QMetaObject::normalizedSignature(
-                "init(QVariant,QVariant)");
-const QByteArray kScreenInitFunctionTypedSignature =
-        QMetaObject::normalizedSignature("init(QString,bool)");
-const QByteArray kScreenShutdownFunctionSignature =
-        QMetaObject::normalizedSignature("shutdown()");
-} // anonymous namespace
 #endif
 
 ControllerScriptEngineLegacy::ControllerScriptEngineLegacy(
@@ -133,50 +116,28 @@ bool ControllerScriptEngineLegacy::callShutdownFunction() {
         return callFunctionOnObjects(m_scriptFunctionPrefixes, "shutdown");
 #ifdef MIXXX_USE_QML
     } else {
-        VERIFY_OR_DEBUG_ASSERT(!m_pJSEngine->hasError()) {
-            qCWarning(m_logger) << "Controller JS engine has an unhandled error.";
-            qCDebug(m_logger) << "Unhandled controller JS error is:"
-                              << m_pJSEngine->catchError().toString();
-        }
-        QHashIterator<QString, std::shared_ptr<QQuickItem>> i(m_rootItems);
         bool success = true;
-        while (i.hasNext()) {
-            i.next();
-            const QMetaObject* metaObject = i.value()->metaObject();
-            const QString& screenIdentifier = i.key();
-
-            VERIFY_OR_DEBUG_ASSERT(metaObject) {
-                qCWarning(m_logger)
-                        << "Invalid meta object for screen" << screenIdentifier
-                        << "It may be that an unhandled issue occurred when importing "
-                           "the scene.";
-                continue;
-            }
-
-            QMetaMethod shutdownFunction;
-            int methodIdx = metaObject->indexOfMethod(kScreenShutdownFunctionSignature);
-
-            if (methodIdx == -1 || !metaObject->method(methodIdx).isValid()) {
+        for (const auto& [screenIdentifier, screen] : m_rootItems) {
+            if (!screen->getShutdown().isCallable()) {
                 qCDebug(m_logger) << "QML Scene for screen" << screenIdentifier
                                   << "has no valid shutdown method.";
                 continue;
             }
 
-            shutdownFunction = metaObject->method(methodIdx);
-
             qCDebug(m_logger) << "Executing shutdown on QML Scene " << screenIdentifier;
 
-            VERIFY_OR_DEBUG_ASSERT(!m_pJSEngine->hasError()) {
-                qCWarning(m_logger) << "Controller JS engine has an unhandled error. Discarding.";
-                qCDebug(m_logger) << "Controller JS error is:"
-                                  << m_pJSEngine->catchError().toString();
-            }
+            auto result = screen->getShutdown().call(QJSValueList{});
+            if (result.isError()) {
+                qCWarning(m_logger) << "Could not shutdown the QML scene for screen"
+                                    << screenIdentifier
+                                    << "gracefully";
 
-            success &= shutdownFunction.invoke(i.value().get(),
-                    Qt::DirectConnection);
-            // Error handling is done in ControllerScriptEngineBase, with the
-            // connection QQmlEngine::warnings ->
-            // ControllerScriptEngineBase::handleQMLErrors
+                // We manually stop the screen before we trigger the shutdown procedure
+                // as this last one may continue rendering process in order to perform
+                // screen splash off.
+                showScriptExceptionDialog(result, false);
+                success = false;
+            }
         }
         return success;
     }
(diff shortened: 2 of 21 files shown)


src/controllers/controller.h
```diff
@@ -141,8 +141,8 @@ class Controller : public QObject {
 
   public:
     // This must be reimplemented by sub-classes desiring to send raw bytes to a
-    // controller.
-    virtual void sendBytes(const QByteArray& data) = 0;
+    // controller. Return true in case of successful completion, false in case of failure.
```
Swiftb0y, line 144:
```suggestion
    // controller. Return true in case of successful completion, false in case of partial completion or failure.
```
as seems to be case in the Hss case

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -523,7 +524,7 @@ int ControllerScriptInterfaceLegacy::beginTimer(
     m_timers[timerId] = info;
     if (timerId == 0) {
         m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Script timer could not be created"));
-    } else if (oneShot) {
+    } else if (oneShot && CmdlineArgs::Instance().getControllerDebug()) {
         qCDebug(m_logger) << "Starting one-shot timer:" << timerId;
     } else {
         qCDebug(m_logger) << "Starting timer:" << timerId;
```
Swiftb0y, line 530: this seems half backed now, either check it for both or none. Though shouldn't it be unnecessary to check, because the logging category takes care of it?

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -693,13 +685,8 @@ TEST_F(ControllerScriptEngineLegacyTest, screenWontSentRawDataIfNotConfigured) {
...
             "accept raw data. Aborting screen rendering.");
 
-    transformScreenFrameFunctions().insert(
-            dummyScreen.identifier,
-            newTransformScreenFrameFunction(
-                    QMetaMethod(),
-                    false));
     renderingScreens().insert(dummyScreen.identifier, pDummyRender);
-    rootItems().insert(dummyScreen.identifier, std::make_shared<QQuickItem>());
+    rootItems().insert(dummyScreen.identifier, new mixxx::qml::QmlMixxxController());
```
Swiftb0y, line 689: this leaks. Instead consider inserting the value directly (instead storing a pointer). You can still take the address of the object later.

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -185,73 +149,42 @@ bool ControllerScriptEngineLegacy::callShutdownFunction() {
 bool ControllerScriptEngineLegacy::callInitFunction() {
     // m_pController is nullptr in tests.
     const auto controllerName = m_pController ? m_pController->getName() : QString{};
+    const auto args = QJSValueList{
+            controllerName,
```
Swiftb0y, line 153: implicit copy instead of move.
```suggestion
    const auto args = QJSValueList{
             m_pController ? m_pController->getName() : QString{},
```

src/qml/qmlmixxxcontroller.h
```diff
@@ -0,0 +1,51 @@
...
+
+class QmlMixxxController : public QQuickItem {
+    Q_OBJECT
+    QML_NAMED_ELEMENT(Controller)
+    Q_PROPERTY(QJSValue init READ getInit WRITE setInit
+                    NOTIFY initChanged REQUIRED);
+    Q_PROPERTY(QJSValue shutdown READ getShutdown WRITE setShutdown
+                    NOTIFY shutdownChanged REQUIRED);
+    Q_PROPERTY(QJSValue transformFrame READ getTransform WRITE setTransform
+                    NOTIFY transformChanged REQUIRED);
```
Swiftb0y, line 18: do we want to support changing these at runtime? Have you thought about marking them `CONSTANT`? Does this still work without `NOTIFY` signal?
[the Qt docs aren't clear on this](https://doc.qt.io/qt-6/qtqml-cppintegration-exposecppattributes.html#notes-on-use-of-notify-signals)

src/qml/qmlmixxxcontroller.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include <QQuickItem>
+
+#include "controllers/rendering/controllerrenderingengine.h"
+
+namespace mixxx {
+namespace qml {
+
+class QmlMixxxController : public QQuickItem {
```
Swiftb0y, line 10: can we rename this to `QmlMixxxControllerScreen`? Otherwise I wouldn't know what to call the hypothetical plain (screen-less) controller feature.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
acolombier: Ah indeed, needs more doing.
By category, I assume you mean the debug category? If yes, this is unfortunately not enough about the noisy logging related to controller communication is blindly fed to the debug category everywhere on the code base.

src/qml/qmlmixxxcontroller.h
acolombier: > Have you thought about marking them CONSTANT?

We could - I guess there could be value in swapping these functions.

> Does this still work without NOTIFY signal?

AFAIK, it does - `NOTIFY` is only relevant from the QML property binding, but in our case we wouldn't care as we will use the value help by the property when needed, no reactiveness is required.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: ah, yeah, thats unfortunate. I think we have some code somewhere though that adjusts the default logging category. That should get rid of the controller debug code in theory. Maybe thats broken? Can you look into it?

acolombier: Fair enough, I didn't realise this was supposed to be a feature since #4523

I'm not sure if it doesn't work as expected or if I misunderstood the point of having a dedicated CLI arg - are we expecting controller debug message to show when running with `--log-level debug`?

src/qml/qmlmixxxcontroller.h
Swiftb0y: right, I'm asking because leveraging `CONSTANT` and `MEMBER` would simplify the class drastically.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: > I'm not sure if it doesn't work as expected or if I misunderstood the point of having a dedicated CLI arg 

Well, the dedicated CLI flag was there first, so we tried our best to keep it for backwards compatibility. 

> are we expecting controller debug message to show when running with --log-level debug?

I'm actually not sure. I'd have to go through the PR again to look it up. Either way, it seems broken, right?

src/qml/qmlmixxxcontroller.h
acolombier: Fair enough - I think making `CONSTANT` is fine . Arguably, if there is the need to swap the function at runtime, this can be done with a bid of logic in those function

Swiftb0y: right, and I don't think there is much of a usecase to re-evaluate other code based on the behavior change, so we can skip the `NOTIFY` signal (which we need to for the `CONSTANT`).

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
acolombier: > Either way, it seems broken, right?

My assumption from reading the `--help` output was that it wouldn't log controller related debug message if not explicitly opted-in so with that understanding, yes it's broken. 
But we could also say the `man` is misleading and that `--controller-debug` will enable debug message only for controller, whereas `--log-level debug` enable **all** log messages, including controller ones.

IMO, it should be `1)` - when I implemented the S4 Mk3 mapping, I add to add that [nasty workaround](https://github.com/mixxxdj/mixxx/blob/42a1fb8ba0da9e83a50bbeaabc9042e5b2b59098/src/controllers/hid/hidioglobaloutputreportfifo.cpp#L81), otherwise I would get a 10G+ logfile after an hour session, due to that message logging every single HID packet coming in every 20ms

acolombier: @Swiftb0yare you happy to keep this work around as-is and create a follow up issue + `FIXME` comment?

src/test/controllerscriptenginelegacy_test.cpp
acolombier: I've added a free in the teardown. This pointer is owned by the QML scene. I have added a comment in the header file.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: works for me, yes

src/controllers/scripting/legacy/controllerscriptenginelegacy.h
```diff
@@ -114,9 +120,9 @@ class ControllerScriptEngineLegacy : public ControllerScriptEngineBase {
     QHash<QString, std::shared_ptr<ControllerRenderingEngine>> m_renderingScreens;
     // Contains all the scenes loaded for this mapping. Key is the scene
     // identifier (LegacyControllerMapping::ScreenInfo::identifier), value in
-    // the QML root item
-    QHash<QString, std::shared_ptr<QQuickItem>> m_rootItems;
-    QHash<QString, TransformScreenFrameFunction> m_transformScreenFrameFunctions;
+    // the QML root item. Note that the pointer is owned by the QML scene which
+    // will free them on shutdown (ControllerScriptEngineLegacy::shutdown)
+    QHash<QString, mixxx::qml::QmlMixxxControllerScreen*> m_rootItems;
```
Swiftb0y, line 125: lets avoid introducing manual memory management. You can wrap this in a unique_ptr. If that doesn't work because QHash needs the key to be copy-able, use `std::unordered_map` instead.

Swiftb0y (commented):
only remaining complaint from me tbh

src/controllers/scripting/legacy/controllerscriptenginelegacy.h
acolombier: Any reason why introducing a smart pointer on a Qt-owned reference here? Just in case you've missed the comment, the pointer is managed by the QML scene, we don't have any control over it.

Swiftb0y: Ah, I just spotted it now. Well the problem is that we do call `delete` on it. So its very easy to mistake it as properly owning. I also fear that the pointer can easily outlive the RenderingEngine their tied to. Ideally the pointers would stay entirely in their rendering engine and in order to access them, you need to access the rendering engine.

Swiftb0y: So sometimes you do delete them and sometimes you don't. It would be really nice if you could fix those ownership issues here.

Swiftb0y: ah, yeah. I was interpreting the 
```cpp
rootItem->setParentItem(pScreen->quickWindow()->contentItem());
```
line as transferring ownership, but thats obviously wrong on second look.

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -693,13 +690,8 @@ TEST_F(ControllerScriptEngineLegacyTest, screenWontSentRawDataIfNotConfigured) {
...
             "accept raw data. Aborting screen rendering.");
 
-    transformScreenFrameFunctions().insert(
-            dummyScreen.identifier,
-            newTransformScreenFrameFunction(
-                    QMetaMethod(),
-                    false));
     renderingScreens().insert(dummyScreen.identifier, pDummyRender);
-    rootItems().insert(dummyScreen.identifier, std::make_shared<QQuickItem>());
+    rootItems().emplace(dummyScreen.identifier, std::make_unique<mixxx::qml::QmlMixxxControllerScreen>());
```
Swiftb0y, line 694: same rationale for `push_back` vs `emplace_back` on `std::vector<std::unique_ptr<T>>` here:
```suggestion
    rootItems().insert(dummyScreen.identifier, std::make_unique<mixxx::qml::QmlMixxxControllerScreen>());
```

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -630,6 +511,8 @@ void ControllerScriptEngineLegacy::handleScreenFrame(
         return;
     };
 
+    auto& pScreen = m_rootItems.at(screenInfo.identifier);
```
Swiftb0y, line 514: consider using `find` to handle the case where its not in the map, otherwise it'll throw.

acolombier: Line 509, there is already an assertin for the string to be in the map

src/test/controllerscriptenginelegacy_test.cpp
acolombier: What is the gain here? Also, please not that this is a test file so macro optimisation is unnecessary.

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
Swiftb0y: Ah right. Well the `contains`+`at` pattern is suboptimal because you do two lookups instead of one. It'd be nice if you could switch these around to only use one `find`.

src/test/controllerscriptenginelegacy_test.cpp
acolombier: Ah good to know! I thought `unique_ptr` were pretty safe when it come to leakge protection, but great to learn they can still s**** you :sweat_smile:  
In the context of unittest, a memory allocation failure would fail the test. Since that one-liner suggestion doesn't work, are you happy with keeping this as-is to limit the effort?

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
acolombier: Makes sense. I have rework it to use a single lookup. Hopefully this works as you had it in mind.

Swiftb0y: yup, pretty much exactly. thanks.

src/test/controllerscriptenginelegacy_test.cpp
Swiftb0y: ah yeah, I think my knowledge on `unordered_map` is not quite good enough to make an educated decision here. Keep as is for now.

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -55,6 +58,9 @@ class ControllerScriptEngineLegacyTest : public ControllerScriptEngineLegacy, pu
 
     void TearDown() override {
         mixxx::Time::setTestMode(false);
+#ifdef MIXXX_USE_QML
+        m_rootItems.clear();
+#endif
```
Swiftb0y, line 63: does it work without the explicit `.clear()`? Ideally RAII would take care of this fully by itself.

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -943,7 +813,7 @@ std::shared_ptr<QQuickItem> ControllerScriptEngineLegacy::loadQMLFile(
         rootItem->setHeight(pScreen->quickWindow()->height());
     }
 
-    return rootItem;
+    return std::unique_ptr<mixxx::qml::QmlMixxxControllerScreen>(rootItem);
```
Swiftb0y, line 816: I think in order to truly benefit from resource safety here, the unique_ptr wrapping needs to be done on the return value of `qmlComponent.createWithInitialProperties` (directly in the same statement) (potentially after a qobject_cast). I don't want to force you to do it, but I don't have the time right now. If you're up for it, feel free to do it now and if not, remind me on sunday and I will file a PR against your branch.

src/test/controllerscriptenginelegacy_test.cpp
acolombier: It will probably work, but I preferred to make it explicit so if something was to crash when deallocating the screen, it still happens as part of the test hooks

Swiftb0y: makes sense. It was more of a general question whether this would be needed in production code.

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
Swiftb0y: yeah I'm just worried about forgetting about it. But you're right, lets postpone for now.

Swiftb0y (approved):
LGTM. Thank you for the cleanup. Please rebase.

acolombier:
Merge?

Swiftb0y:
yup, sorry. I've been busy until now.

PR #13740, opened 2024-10-08
fix: prevent null CO access when cloning sampler or preview

acolombier:


Swiftb0y (approved):
LGTM, I guess, though I'll let someone else that reviewed this code be the judge.

src/engine/channels/enginedeck.cpp
```diff
@@ -206,6 +207,18 @@ void EngineDeck::cloneStemState(const EngineDeck* deckToClone) {
     VERIFY_OR_DEBUG_ASSERT(deckToClone) {
         return;
     }
+    // Sampler and preview decks don't have stem controls
+    if ((m_stemGain.empty() && m_stemMute.empty()) ||
+            (deckToClone->m_stemGain.empty() &&
+                    deckToClone->m_stemMute.empty())) {
+        return;
+    }
```
daschuer, line 215: The verify blow includes this check and sticks around at release time. So this check can be removed.

daschuer (changes requested):

src/engine/channels/enginedeck.cpp
acolombier: It does but it's not equivalent. 

Having the two vector empty is an expected case, for which we do an early return. The assert condition adds extra usecase that would crash the for-loop but are not expected usecase on which we should assert.

I agree that this would technically add couple of extra instructions at release time, but since this function is not time critical, I would suggest we keep the clear distinction between the early return on excepted usecase, and protective return on unexpected case.

daschuer: Ah I see. How about replacing the first loop by 
`If (isPrimaryDeck())`?

acolombier: That sounds much cleaner indeed. Updated.

daschuer (approved):
LGTM. Thank you. Please squash the fixup and I will merge.

acolombier:
Done!

JoergAtGithub:
LGTM! Thank you!

PR #13747, opened 2024-10-09
Show wait cursor when re/loading a skin (not during startup)

ronso0:
Often enough when reloading a skin during develpoment I wait and wonder if the Ctrl+Shift+R hotkey triggered a reload.

Now we see a wait cursor while the skin is being loaded, like when applying a sound config.
A dim translucent overlay while reloading would also be nice IMO.


cr7pt0gr4ph7 (approved):
Looks good :+1:

daschuer (approved):
LGTM, Thank you.

Issue #13716, opened 2024-10-02
AutoDj: New transmission mode 'Radio Laneway'


davidlmorris:
### Feature Description

This transmission mode is designed for social functions, especially playing popular music, designed to sound like a commercial radio station. For this purpose, it is vastly superior to existing transmission modes, and for cold endings especially, consistently performing better than most people I suspect (well me at least!).

We will need to add two new Cues, FadeIn and FadeOut.  (cf analyzersilence.cpp  pN60dBSound and kSilenceThreshold.)   The FadeIn range is the time from the Main cue point (or first sound if the  MainCue is not set) to the first time a track reaches kFadeInThreshold which is set to -27 dBV (or 0.0447f).  The distance between these is the fadeInLength. The FadeOut range is from the last time a track was at kFadeOutThreshold which is set to  -12 dBV (or 0.2511f), to the last sound. The distance between these is the fadeOutLength. The values are selected based on "sounding good" while reducing the chance of pathological times for fade-ins or outs.

This transmission mode also separates (on occasions) the start of a crossfade  (fadeBeginPos)      from the start of the track where the next track starts is sometimes earlier than the start of the crossfade (playNextPos).  In all cases the crossfade starts the next track at full volume, meaning that the crossfader is set to the centre position.  This requires the code from TransitionMode::FixedStartCenterSkipSilence ([Auto dj cross fader center #13628](https://github.com/mixxxdj/mixxx/pull/13628)).

This transition mode does the following:

1. If a playing (from) track has reached the FadeOut cue start, then a crossfade starting center is initiated, with the transmission time taking the fadeOutLength, or value from the spin box (fixed value), whichever is smaller.  This fits the majority of cases an provides a very smooth crossfade on tracks with a fadeout, and excellent timing when a track has a sharp or cold ending.

[...]

PR #13750, opened 2024-10-10
Adding RadioLanewayCrossfade transition

davidlmorris:
RadioLanewayCrossfade transition.
See AutoDj: New transition mode 'Radio Laneway'  https://github.com/mixxxdj/mixxx/issues/13716.

This transition mode is designed for social functions, especially playing popular music, and is intended to sound like a commercial radio station. For this purpose, it is vastly superior to existing transmission modes, and for cold endings especially, it consistently performs better than most people I suspect (well, me at least!).

This adds two new Cues, FadeIn and FadeOut. (cf analyzersilence.cpp pN60dBSound and kSilenceThreshold.) The FadeIn range is the time from the Main cue point (or first sound if the MainCue is not set) to the first time a track reaches kFadeInThreshold which is set to -27 dBV (or 0.0447f). The distance between these is the fadeInLength. The FadeOut range is from the last time a track was at kFadeOutThreshold which is set to -12 dBV (or 0.2511f), to the last sound. The distance between these is the fadeOutLength. The values are selected based on "sounding good" while reducing the chance of pathological times for fade-ins or outs.  Note that tracks always start a full volume (cf. TransitionMode::FixedStartCenterSkipSilence https://github.com/mixxxdj/mixxx/pull/13628).

This transition mode does the following:

1. If a playing (from) track has reached the FadeOut cue start, then a crossfade starting center is initiated, with the transmission time taking the fadeOutLength, or value from the spin box (fixed value), whichever is smaller. This fits most cases and provides a smooth crossfade on tracks with a fadeout, and excellent timing when a track has a sharp or cold ending.

2. If the next (to) track has a fadeInLength greater than 0.25 seconds then the next (to) track starts a fadeInLength value (playNextPos) earlier than the current (from) deck fadeOutStart. This covers the case where we have a slow fade-in of a song, for example, when Boston's "More Than A Feeling", comes next after a song with a cold (stab) end, removing the perception of a gap, while preserving the cold ending. (In this example it is about 3 seconds earlier, though the new track is not perceptible during a cold ending). Note that the crossfader holds in the center while the next (to) track starts, and starts moving the fader once fadeOutStart is reached. If the fadeOutLength + playNextPos difference is greater than the spin box, then revert to item 1 above using (fixed value) calculated from fadeOutLength.
 
3. If the next (to) track has a duration of less than 15 seconds then this track is likely a jingle or sample. In this case, the fadeInLength is 0 and the fadeOutLength is 1. This way we have a fast neat transition while maximizing the time for the next track to load.

### DIFF
--- a/src/analyzer/analyzersilence.cpp
+++ b/src/analyzer/analyzersilence.cpp
@@ -115,8 +115,7 @@ SINT AnalyzerSilence::findFirstFadeOutChunk(std::span<const CSAMPLE> samples) {
     // -1 is required, because the distance from the fist sample index (0) to crend() is 1,
     SINT ret = std::distance(
                        first_fade_out_sound(samples.rbegin(), samples.rend()),
-                       samples.rend()) -
-            1;
+                       samples.rend()) - 1;
     if (ret == -1) {
         ret = samples.size();
     }
@@ -126,9 +125,6 @@ SINT AnalyzerSilence::findFirstFadeOutChunk(std::span<const CSAMPLE> samples) {
 // static
 SINT AnalyzerSilence::findLastFadeInChunk(std::span<const CSAMPLE> samples) {
     SINT ret = std::distance(samples.begin(), last_fade_in_sound(samples.begin(), samples.end()));
-    //    if (ret == samples.size()) {
-    //        ret = 0;
-    //    }
     return ret;
 }
 
--- a/src/analyzer/analyzersilence.cpp
+++ b/src/analyzer/analyzersilence.cpp
@@ -115,7 +115,8 @@ SINT AnalyzerSilence::findFirstFadeOutChunk(std::span<const CSAMPLE> samples) {
     // -1 is required, because the distance from the fist sample index (0) to crend() is 1,
     SINT ret = std::distance(
                        first_fade_out_sound(samples.rbegin(), samples.rend()),
-                       samples.rend()) - 1;
+                       samples.rend()) -
+            1;
     if (ret == -1) {
         ret = samples.size();
     }
--- a/src/analyzer/analyzersilence.cpp
+++ b/src/analyzer/analyzersilence.cpp
@@ -119,11 +119,6 @@ SINT AnalyzerSilence::findFirstFadeOutChunk(std::span<const CSAMPLE> samples) {
                                      kFadeOutThreshold),
                        samples.rend()) -
             1;
-    // if we don't find it (track only partially loaded - and/or pathological)
-    // give us the track size.
-    if (ret == -1) {
-        ret = samples.size();
-    }
     return ret;
 }
 
@@ -159,14 +154,14 @@ bool AnalyzerSilence::processSamples(const CSAMPLE* pIn, SINT count) {
     }
     if (m_fadeThresholdFadeInEnd >= 0) {
         const SINT lasttSampleBeforeFadeOut = findFirstFadeOutChunk(samples);
-        if (lasttSampleBeforeFadeOut < (count - 1)) {
+        if (lasttSampleBeforeFadeOut >= 0) {
             m_fadeThresholdFadeOutStart = m_framesProcessed +
                     (lasttSampleBeforeFadeOut / m_channelCount) + 1;
         }
     }
     if (m_fadeThresholdFadeOutStart >= 0) {
         const SINT lastSoundSample = findLastSoundInChunk(samples);
-        if (lastSoundSample < (count - 1)) { // not only sound or silence
+        if (lastSoundSample >= 0) {
             m_signalEnd = m_framesProcessed + (lastSoundSample / m_channelCount) + 1;
         }
     }
--- a/src/analyzer/analyzersilence.cpp
+++ b/src/analyzer/analyzersilence.cpp
@@ -12,12 +12,34 @@ constexpr CSAMPLE kSilenceThreshold = 0.001f; // -60 dB
 // TODO: Change the above line to:
 //constexpr CSAMPLE kSilenceThreshold = db2ratio(-60.0f);
 
+// These are in dBV expressed as Volts RMS (which seems, sensibly,
+// the way Mixxx works).
+// Don't change these to const as they are used only to feed the
+// fade thresholds which are consts below themselves while we work
+// out which one is best, and you'll just be adding to exe size.
+#define N_10DB_FADEOUT_THRESHOLD 0.3162f
+#define N_12DB_FADEOUT_THRESHOLD 0.2511f
+#define N_15DB_FADEOUT_THRESHOLD 0.1778f
+#define N_18DB_FADEOUT_THRESHOLD 0.1259f
+#define N_20DB_FADEOUT_THRESHOLD 0.1f
+#define N_24DB_FADEOUT_THRESHOLD 0.0631f
+#define N_25DB_FADEOUT_THRESHOLD 0.0562f
+#define N_27DB_FADEOUT_THRESHOLD 0.0447f
+#define N_30DB_FADEOUT_THRESHOLD 0.0316f
+#define N_40DB_FADEOUT_THRESHOLD 0.01f
+
+constexpr CSAMPLE kFadeInThreshold = N_27DB_FADEOUT_THRESHOLD;
+constexpr CSAMPLE kFadeOutThreshold = N_12DB_FADEOUT_THRESHOLD;
+
+
 bool shouldAnalyze(TrackPointer pTrack) {
     CuePointer pIntroCue = pTrack->findCueByType(mixxx::CueType::Intro);
     CuePointer pOutroCue = pTrack->findCueByType(mixxx::CueType::Outro);
     CuePointer pN60dBSound = pTrack->findCueByType(mixxx::CueType::N60dBSound);
+    CuePointer pFadeIn = pTrack->findCueByType(mixxx::CueType::FadeIn);
+    CuePointer pFadeOut = pTrack->findCueByType(mixxx::CueType::FadeOut);
 
-    if (!pIntroCue || !pOutroCue || !pN60dBSound || pN60dBSound->getLengthFrames() <= 0) {
+    if (!pFadeIn || !pFadeOut || !pIntroCue || !pOutroCue || !pN60dBSound || pN60dBSound->getLengthFrames() <= 0) {
         return true;
     }
     return false;
@@ -30,13 +52,29 @@ Iterator first_sound(Iterator begin, Iterator end) {
     });
 }
 
+template<typename ForwardIterator>
+ForwardIterator last_fade_in_sound(ForwardIterator begin, ForwardIterator end) {
+    return std::find_if(begin, end, [](const auto elem) {
+        return fabs(elem) >= kFadeInThreshold;
+    });
+}
+
+template<typename Iterator>
+Iterator first_fade_out_sound(Iterator begin, Iterator end) {
+    return std::find_if(begin, end, [](const auto elem) {
+        return fabs(elem) >= kFadeOutThreshold;
+    });
+}
+
 } // anonymous namespace
 
 AnalyzerSilence::AnalyzerSilence(UserSettingsPointer pConfig)
         : m_pConfig(pConfig),
           m_framesProcessed(0),
           m_signalStart(-1),
-          m_signalEnd(-1) {
+          m_signalEnd(-1),
+          m_fadeThresholdFadeInEnd(-1),
+          m_fadeThresholdFadeOutStart(-1) {
 }
 
 bool AnalyzerSilence::initialize(const AnalyzerTrack& track,
@@ -53,6 +91,8 @@ bool AnalyzerSilence::initialize(const AnalyzerTrack& track,
     m_framesProcessed = 0;
     m_signalStart = -1;
     m_signalEnd = -1;
+    m_fadeThresholdFadeInEnd = -1;
+    m_fadeThresholdFadeOutStart = -1;
     m_channelCount = channelCount;
 
     return true;
@@ -67,6 +107,25 @@ SINT AnalyzerSilence::findFirstSoundInChunk(std::span<const CSAMPLE> samples) {
 SINT AnalyzerSilence::findLastSoundInChunk(std::span<const CSAMPLE> samples) {
     // -1 is required, because the distance from the fist sample index (0) to crend() is 1,
     SINT ret = std::distance(first_sound(samples.rbegin(), samples.rend()), samples.rend()) - 1;
+    return ret;
+}
+
+// static
+SINT AnalyzerSilence::findFirstFadeOutChunk(std::span<const CSAMPLE> samples) {
+    // -1 is required, because the distance from the fist sample index (0) to crend() is 1,
+    SINT ret = std::distance(first_fade_out_sound(samples.rbegin(), samples.rend()), samples.rend()) - 1;
+    if (ret == -1) {
+        ret = samples.size();
+    }
+    return ret;
+}
+
+// static
+SINT AnalyzerSilence::findLastFadeInChunk(std::span<const CSAMPLE> samples) {
+    SINT ret = std::distance(samples.begin(), last_fade_in_sound(samples.begin(), samples.end()));
+    //    if (ret == samples.size()) {
+    //        ret = 0;
+    //    }
     return ret;
 }
 
@@ -93,10 +152,23 @@ bool AnalyzerSilence::processSamples(const CSAMPLE* pIn, SINT count) {
             m_signalStart = m_framesProcessed + firstSoundSample / m_channelCount;
         }
     }
-    if (m_signalStart >= 0) {
+
+    if (m_fadeThresholdFadeInEnd < 0) {
+        const SINT lastSampleOfFadeIn = findLastFadeInChunk(samples);
+        if (lastSampleOfFadeIn < count) {
+            m_fadeThresholdFadeInEnd = m_framesProcessed + (lastSampleOfFadeIn / m_channelCount);
+        }
+    }
+    if (m_fadeThresholdFadeInEnd >= 0) {
+        const SINT lasttSampleBeforeFadeOut = findFirstFadeOutChunk(samples);
+        if (lasttSampleBeforeFadeOut < (count - 1)) {
+            m_fadeThresholdFadeOutStart = m_framesProcessed + (lasttSampleBeforeFadeOut / m_channelCount) + 1;
+        }
+    }
+    if (m_fadeThresholdFadeOutStart >= 0) {
         const SINT lastSoundSample = findLastSoundInChunk(samples);
-        if (lastSoundSample >= 0) {
-            m_signalEnd = m_framesProcessed + lastSoundSample / m_channelCount + 1;
+        if (lastSoundSample < (count - 1)) { // not only sound or silence
+            m_signalEnd = m_framesProcessed + (lastSoundSample / m_channelCount) + 1;
         }
     }
 
@@ -136,6 +208,36 @@ void AnalyzerSilence::storeResults(TrackPointer pTrack) {
 
     setupMainAndIntroCue(pTrack.get(), firstSoundPosition, m_pConfig.data());
     setupOutroCue(pTrack.get(), lastSoundPosition);
+
+    if (m_fadeThresholdFadeInEnd < 0) {
+        m_fadeThresholdFadeInEnd = 0;
+    }
+    if (m_fadeThresholdFadeOutStart < 0) {
+        m_fadeThresholdFadeOutStart = m_framesProcessed;
+    }
+    const auto fadeInEndPosition = mixxx::audio::FramePos(m_fadeThresholdFadeInEnd);
+    CuePointer pFadeIn = pTrack->findCueByType(mixxx::CueType::FadeIn);
+    if (pFadeIn == nullptr) {
+        pFadeIn = pTrack->createAndAddCue(
+                mixxx::CueType::FadeIn,
+                Cue::kNoHotCue,
+                firstSoundPosition,
+                fadeInEndPosition);
+    } else {
+        pFadeIn->setStartAndEndPosition(firstSoundPosition, fadeInEndPosition);
+    }
+
+    const auto fadeOutStartPosition = mixxx::audio::FramePos(m_fadeThresholdFadeOutStart);
+    CuePointer pFadeOut = pTrack->findCueByType(mixxx::CueType::FadeOut);
+    if (pFadeOut == nullptr) {
+        pFadeOut = pTrack->createAndAddCue(
+                mixxx::CueType::FadeOut,
+                Cue::kNoHotCue,
+                fadeOutStartPosition,
+                lastSoundPosition);
+    } else {
+        pFadeOut->setStartAndEndPosition(fadeOutStartPosition, lastSoundPosition);
+    }
 }
 
 // static
--- a/src/analyzer/analyzersilence.cpp
+++ b/src/analyzer/analyzersilence.cpp
@@ -12,24 +12,22 @@ constexpr CSAMPLE kSilenceThreshold = 0.001f; // -60 dB
 // TODO: Change the above line to:
 // constexpr CSAMPLE kSilenceThreshold = db2ratio(-60.0f);
 
-// These are in dBV expressed as Volts RMS (which seems, sensibly,
-// the way Mixxx works).
-// Don't change these to const as they are used only to feed the
-// fade thresholds which are consts below themselves while we work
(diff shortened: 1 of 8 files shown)


src/analyzer/analyzersilence.cpp
```diff
@@ -10,14 +10,36 @@ namespace {
...
+#define N_10DB_FADEOUT_THRESHOLD 0.3162f
+#define N_12DB_FADEOUT_THRESHOLD 0.2511f
+#define N_15DB_FADEOUT_THRESHOLD 0.1778f
+#define N_18DB_FADEOUT_THRESHOLD 0.1259f
+#define N_20DB_FADEOUT_THRESHOLD 0.1f
+#define N_24DB_FADEOUT_THRESHOLD 0.0631f
+#define N_25DB_FADEOUT_THRESHOLD 0.0562f
+#define N_27DB_FADEOUT_THRESHOLD 0.0447f
+#define N_30DB_FADEOUT_THRESHOLD 0.0316f
+#define N_40DB_FADEOUT_THRESHOLD 0.01f
```
Swiftb0y, line 29: make them `constexpr`. they have internal linkage and the compiler will strip them out.

src/analyzer/analyzersilence.cpp
```diff
@@ -30,13 +52,29 @@ Iterator first_sound(Iterator begin, Iterator end) {
...
+    });
+}
+
+template<typename Iterator>
+Iterator first_fade_out_sound(Iterator begin, Iterator end) {
+    return std::find_if(begin, end, [](const auto elem) {
+        return fabs(elem) >= kFadeOutThreshold;
+    });
+}
+
```
Swiftb0y, line 68: this is a copy of the above for very little reason. Consider deduplicating by simply taking the Threshold as a parameter.

src/library/autodj/dlgautodj.cpp
```diff
@@ -150,7 +151,13 @@ DlgAutoDJ::DlgAutoDJ(WLibrary* parent,
...
             "Skip Silence Start Full Volume:\n"
             "The same as Skip Silence, but starting transitions with a centered\n"
-            "crossfader, so that the intro starts at full volume.\n");
+            "crossfader, so that the intro starts at full volume.\n"
+            "\n"
+            "Radio Laneway Crossfade:\n"
+            "Starts the next track at full volume.  Starts the crossfade when the\n"
+            "volume last falls below -12Db or at the spin box setting which ever\n"
+            "is lower, and potentially starts the next earlier if it starts below\n"
+            "-27Db.");
```
Swiftb0y, line 160: its very easy to change the threshold in the analyzer but forget to update it here. Do you think its valuable to have the thresholds documented here or can we just omit them?

src/library/autodj/dlgautodj.cpp
```diff
@@ -70,8 +70,9 @@ DlgAutoDJ::DlgAutoDJ(WLibrary* parent,
             &WTrackTableView::setSelectedClick);
 
     QBoxLayout* box = qobject_cast<QBoxLayout*>(layout());
-    VERIFY_OR_DEBUG_ASSERT(box) { //Assumes the form layout is a QVBox/QHBoxLayout!
-    } else {
+    VERIFY_OR_DEBUG_ASSERT(box) { // Assumes the form layout is a QVBox/QHBoxLayout!
+    }
+    else {
```
Swiftb0y, line 75: was this clang-format or you?

Swiftb0y (commented):
didn't bother looking at the autodj for now.

src/analyzer/analyzersilence.cpp
davidlmorris: The second one goes backward, while the first is strictly forward.  I spent a long time playing around with these, trying to determine why I was getting weird results (and so I was keen to keep them separate).  It turns out AutoDj processing can take place before a track is fully analysed, so I am less invested in that now.

src/library/autodj/dlgautodj.cpp
davidlmorris: Was not intentionally me...  But I gave up trying to remove spaces at the ends of lines and I did run a format document at some point, so perhaps then.

src/library/autodj/dlgautodj.cpp
davidlmorris: I nearly made it a more general statement (e.g., 'a certain level'), but I figured someone would want to know the exact values. It may well be worth making it more general for the reason you suggested.

src/analyzer/analyzersilence.cpp
davidlmorris: No. I want the compiler to strip them out.  Only a couple of these are used, and then only as part of a const.  They are there just while we(?) work out the best values. This makes it easier to change. They can be deleted later when we have a fixed value.

src/analyzer/analyzersilence.cpp
Swiftb0y: The functions are identical. In this context `ForwardIterator` and `Iterator` are just names that resolve to the underlying iterator implementation when called. So there is no difference in terms of what function can be called when. A [`std::reverse_iterator`](https://en.cppreference.com/w/cpp/iterator/reverse_iterator) is a ["bidirectional iterator"](https://en.cppreference.com/w/cpp/named_req/BidirectionalIterator) and thus also a ["forward iterator"](https://en.cppreference.com/w/cpp/named_req/ForwardIterator).

src/library/autodj/dlgautodj.cpp
Swiftb0y: can you revert that then?

src/analyzer/analyzersilence.cpp
Swiftb0y: `std::find_if` only requires a forward iterator.

src/analyzer/analyzersilence.cpp
Swiftb0y: > No. I want the compiler to strip them out. Only a couple of these are used, and then only as part of a const. 

Right, [as long as you mark them `const`. the compiler will remove them.](https://compiler-explorer.com/#g:!((g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:9,positionColumn:1,positionLineNumber:9,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:'namespace+%7B%0Aconstexpr+int+internal1+%3D+1%3B%0A%7D%0Aconstexpr+int+internal2+%3D+2%3B%0Aconst+int+internal3+%3D+3%3B%0A%0A//+only+this+shows+up+in+the+binary%0Aint+external+%3D+3%3B%0A'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+14.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+(trunk)+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0'),(g:!((h:compiler,i:(compiler:gsnapshot,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+(trunk)+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

Anyways, not particularly important if these are only meant to be temporary (that was not clear from the source).

src/analyzer/analyzersilence.cpp
davidlmorris: If you like I'll refactor all of those into one, including the db60.

davidlmorris:
> didn't bother looking at the autodj for now.

I suggest you do.  This transition is groundbreaking.

src/analyzer/analyzersilence.cpp
davidlmorris: >>  while we work out which one is best

src/analyzer/analyzersilence.cpp
Swiftb0y: yes, that would be nice. Thank you.

davidlmorris: As I went to do this I've had second thoughts.

Can you look again at this, and consider how much if any savings will be made by a refactor here?  Since are are assigning parameters to the stack in a call for functions have are already present in std lib (so nothing additional) then I don't see a saving, in fact, we are adding to the time to complete.  And, the code is more readable with several functions (note the difference is a const).  Unnecessarily concise code does not equal readability, right?

src/library/autodj/dlgautodj.cpp
davidlmorris: Done.

davidlmorris: Rejected by your code styler.

```
diff --git a/src/library/autodj/dlgautodj.cpp b/src/library/autodj/dlgautodj.cpp
index 539ece414f..9c11598612 100644
--- a/src/library/autodj/dlgautodj.cpp
+++ b/src/library/autodj/dlgautodj.cpp
@@ -71,7 +71,8 @@ DlgAutoDJ::DlgAutoDJ(WLibrary* parent,
 
     QBoxLayout* box = qobject_cast<QBoxLayout*>(layout());
     VERIFY_OR_DEBUG_ASSERT(box) { // Assumes the form layout is a QVBox/QHBoxLayout!
-    } else {
+    }
+    else {
         box->removeWidget(m_pTrackTablePlaceholder);
         m_pTrackTablePlaceholder->hide();
         box->insertWidget(1, m_pTrackTableView);
```

davidlmorris: Reverting the revert.

src/analyzer/analyzersilence.cpp
Swiftb0y: well, the think that trips me up when reading is that the body is repeated several times. Its difficult to spot where the difference is. I don't expect any meaningful performance improvements from this. If you want to give the difference functions a name, you're welcome to do so, but remove the duplicated bodies.
Something this would work for me:

```cpp
template<typename It, typename T>
It find_first_above_threshold(It begin, It end, T threshold) {
    return std::find_if(begin, end, [threshold](const auto& elem) {
        return std::abs(elem) >= threshold;
    });
}

template<typename It>
It last_fade_in_sound(It begin, It end) {
    return find_first_above_threshold(begin, above, kFadeInThreshold);
}

template<typename It>
Iterator first_fade_out_sound(Iterator begin, Iterator end) {
    return find_first_above_threshold(begin, above, kFadeOutThreshold);
}

```

src/library/autodj/dlgautodj.cpp
Swiftb0y: Interesting, then keep it. Sorry for the noise

src/analyzer/analyzersilence.cpp
davidlmorris: I understood what you wanted, and to be honest there is nothing in it really, not from the readability point of view (as I see it).  And while having the const and not pushing as much on the stack, should be slightly faster, but not meaningfully faster... and who knows what the the compiler will actually do, so this isn't a big deal. 

However, given that this is the analysis section... is performance important?

Swiftb0y: Not in terms of what this would a difference, no. I'm purely suggesting this because I find it easier to read.

davidlmorris: How anyone reads lambda functions with closure readable is beyond me anyway, so I have made the changes.  And while I remain slightly concerned about performance, this is not the hill I would choose to die on.

(And besides I find braces on the same line as an 'if' far from readable, as studies into this have shown cf Code Complete 1st ed, so perhaps I am not the best to judge).

davidlmorris:
I noticed https://github.com/mixxxdj/mixxx/pull/13545 which had fixed a regression issue that was more recent than the code I was working from, so I have "re-re-verted" this so the [issue](https://github.com/mixxxdj/mixxx/issues/13449) won't suddenly appear again.  In my testing, though,  I didn't see this problem occur.

src/analyzer/analyzersilence.cpp
Swiftb0y: I'd recommend you to look a bit into functional programming. Learning Haskell is really eye opening ;) I can recommend learnyouahaskell.com

davidlmorris: I did Prolog at Uni many decades ago.  Functional programming has its value in a few small cases for recursive algorithms and for muti-thread CPUs (which were not readily available when I was studying).  But there are reasons why the vast bulk of code is written in a procedural language.  And, one of those is readability. (Another is performance.)

Swiftb0y: Prolog isn't really a functional language though, its declaritive. If you're coming from the performance angle, I can recommend you to learn rust. Its amazing how nice code can look if its functional-ish while still being blazingly fast. The bulk of code is primarily written because its old and there was nothing better back then thats still performant. That is not the case anymore today. C++ is also a multi-paradigmatic language which includes all the necessary tools to clarify as functional as well. Anyways, it's not very productive to argue about this here. I was just trying to encourage you to broaden your horizon a little bit ;)

davidlmorris: You are technically correct. As I remember it at the time Prolog was considered to be like a functional language... and the way it works has similarities at least.   Haskal was a thing. There was an elective for Lambda Calculus and Haskal.  Looked too much like math (and formal logic which I did in philosophy) to attract my interest.  I did Networking and Databases.  This was the late 80's early 90's as a mature-age student.

As for Rust, this is one I have already downloaded and considered for my microcontroller projects, mostly in the wake of Counterstrike.  I've backed off a little since then because it doesn't have the lib support yet (for that area).  (And let's face it it is a tad faddish at the moment - yes I am saying that like it is a bad thing - even if Rust will stick.)  

As for my horizons, you do realise that I am in my 60's and that I have been doing this for longer (probably) than you have been around, having first programmed on a Z-80 back in the very early 80's, right?  When I started PC's weren't even a thing.  The Commodore 64 was a year or two away.  And, I had to build my first computer.   Back then you could understand everything that was happening inside the computer.  I nearly built the [EDUC-8](https://en.wikipedia.org/wiki/EDUC-8) back in 1975 when I was still in high school, but could not afford it, and I also doubled the utility of it (I was right - though it would have been a great learning exercise.  And it was complex (for the time).  Switches and lights... no KB.  Back then I was focused on sound and mixers.  And I was more results-focused.  I wanted to mix and sound like the Radio (strangely that hasn't changed!)  In the next seven years, computers went from unobtainable to commonplace - the Commodore 64 I purchased in 1983 was my favourite.)

There is a time in your life when you are young when the "latest and greatest" (= fashionable) always seems to be the best.  And, old fogeys don't seem to adjust well.   However, the 'latest and greatest' doesn't always work out, even if the latter statement is still probably true.  'Amiga' anyone?

You clearly know your stuff, much more than I did at your age.  (I am making assumptions about your age - please correct me if I am wrong).  Since you have taken the liberty of recommending things to me, how about you take a look at Steve McConnell's "Code Complete"?

Swiftb0y: Thank you for your recommendation. I've studied some of Robert C. Martins books so far but "Code Complete" also seems interesting. I will try to get a hold of a copy.

davidlmorris: Yeah, I've read some Uncle Bob stuff.  I should check out his book(s?).  What I liked about Steve's work was it was so concrete (practical), and backed up by evidence.

So to the matter at hand.  Is there anything else that we need to do to get this merged?

Swiftb0y: I currently don't have the time to give this yet another proper review. Please consult with @daschuer on the next steps.

davidlmorris:
@daschuer ping

daschuer:
I am a bit puzzled how this iteracts with manual set intros and outros. While this works reasonable good by default for radio mixes, it will probably fail for tracks with club intros and outros or track with ambience noise at the beginning or end. 

This is IMHO a perfect solution for the default case where a track has only an IntroStrat and an OutroEnd Cue set. 
Is this new analyzer suitable to set the IntroEnd and OutroSTart Cue? 

Did you consider to priorize the manual set intro and outro, to give the user the power to adjust exceptional tracks?

daschuer (changes requested):

src/analyzer/analyzersilence.cpp
```diff
@@ -10,33 +10,58 @@ namespace {
...
 constexpr CSAMPLE kSilenceThreshold = 0.001f; // -60 dB
 // TODO: Change the above line to:
-//constexpr CSAMPLE kSilenceThreshold = db2ratio(-60.0f);
+// constexpr CSAMPLE kSilenceThreshold = db2ratio(-60.0f);
+
+// These are in dBV expressed as Volts RMS (which seems, sensibly,
+// the way Mixxx works).
+// Don't change these to const as they are used only to feed the
+// fade thresholds which are consts below themselves while we work
+// out which one is best, and you'll just be adding to exe size.
```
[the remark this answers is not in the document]
davidlmorris: The whole thing is a comment in case the chosen values are not the best.  I should just delete the lot. I believe I have good values here.

src/analyzer/analyzersilence.cpp
```diff
@@ -53,20 +78,47 @@ bool AnalyzerSilence::initialize(const AnalyzerTrack& track,
...
-    SINT ret = std::distance(first_sound(samples.rbegin(), samples.rend()), samples.rend()) - 1;
+    SINT ret = std::distance(find_first_above_threshold(samples.rbegin(),
+                                     samples.rend(),
+                                     kSilenceThreshold),
+                       samples.rend()) -
+            1;
+    return ret;
+}
+
+// Find the number of first sound sample where the sound is above kFadeInThreshold (-27db)
```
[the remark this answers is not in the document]
davidlmorris: Agreed.

src/analyzer/analyzersilence.cpp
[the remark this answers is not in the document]
davidlmorris: > I am a bit puzzled how this iteracts with manual set intros and outros.

You asked about this before.  It does not interact or use manual set intros and outros cues at all.  It runs as discussed in [ AutoDj: New transition mode 'Radio Laneway' #13716](https://github.com/mixxxdj/mixxx/issues/13716) 

I've yet to find a track where it isn't 'good enough', and in some cases does consistently better than me (where I might miss 1 in 10 cold endings), so I'm not sure I agree with you about not working with ambient noise at the beginning or end, but since I can't claim any great familiarity with club music I will leave that for you to decide.  

Why not try it and see?

src/track/cueinfo.h
```diff
@@ -19,6 +19,10 @@ enum class CueType {
     Outro = 7,
     N60dBSound = 8, // range that covers beginning and end of audible
                     // sound; not shown to user
+    FadeIn = 9,     // First time sound reaches a certain level
+                    // (cf analyzersilence.cpp);  not shown to user
+    FadeOut = 10    // Last time sound reaches a certain level
+                    // (cf analyzersilence.cpp);  not shown to user
```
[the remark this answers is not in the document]
davidlmorris: >  A perfect solution would be to use a certain Loudness level after applying replay gain. Did you consider to implement that?

I did, but I was struggling to see how the track could have silence markers and not be normalised.   I haven't quite understood the order of operation, and how often the analysis section is called.  And importantly if normalisation is before or after silence detection.  It seems to be resolved several times as far as I can tell... (mostly from the spurious results I was getting while a track was being loaded).  All of which says this probably needs to be considered well before implementation.  Right now, I'm not sure I know enough.  But I am happy to take a look.  Should this be a discussion in chat, since this is a separate topic?

> Please add also a comment about the level itself.
> Is it the last time of full volume, or just an alternative silence threshold?

It is the last or first time the sound reaches a certain level.  Since full volume is subjective, I thought this was more precise.  It is like an alternative silence threshold but is different because there are two different threshold values.  

 (I guess this isn't clear - so perhaps I should add something like: So to be clear, this is the last sample in the track where the volume was above the FadeOut level, and the first sample in the track where the sound volume was above the FadeIn level.)

src/analyzer/analyzersilence.cpp
```diff
@@ -10,33 +10,58 @@ namespace {
...
+#define N_10DB_FADE_THRESHOLD 0.3162f
+#define N_12DB_FADE_THRESHOLD 0.2511f
+#define N_15DB_FADE_THRESHOLD 0.1778f
+#define N_18DB_FADE_THRESHOLD 0.1259f
+#define N_20DB_FADE_THRESHOLD 0.1f
+#define N_24DB_FADE_THRESHOLD 0.0631f
+#define N_25DB_FADE_THRESHOLD 0.0562f
+#define N_27DB_FADE_THRESHOLD 0.0447f
+#define N_30DB_FADE_THRESHOLD 0.0316f
+#define N_40DB_FADE_THRESHOLD 0.01f
```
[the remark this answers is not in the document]
davidlmorris: (My intent with those comments was to have a placeholder for the values while I played around.  Assume those are deleted by the time someone is making a change here.  They really are not necessary now)

But I understand the main point.  It is the same problem as the -60db setting.

There would be a cost if we do a lot of processing every time a track is loaded.   Should we have a Cue value that is simply a version flag associated with every track, which could be checked by a future version to fix any issues?  No cue = 4.x or earlier?

For the user adjustment, I suggest that the next step would be to add some UI elements to make this all visible and allow the user adjust.  We also discussed this.  Happy to work on it as a next step (when I get time).  But I think this might be a future version change, based on any perceived need.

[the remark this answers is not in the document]
davidlmorris: As I sent that I realised 'Cue value' may not make sense.  But at least have a flag with the track data that shows a version number.  Or, alternatively have a cueinfo.version setting for each cue (though that smells of data duplication).   After a very quick look at the code base - noting exceptions for things like "Serato's tag data", I realise even this might be problematic.  So this too is also, worthy of a chat discussion.

src/track/cueinfo.h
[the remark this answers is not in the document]
davidlmorris:
```
    FadeIn = 9,     // Range from the start of the track to the first sample in
                    // the track where the sound volume was above the
                    // kFadeInThreshold (cf analyzersilence.cpp);
                    // not shown to user
    FadeOut = 10    // Range from the last sample in the track where
                    // the sound volume was above the kFadeOutThreshold
                    // to the end of the track (cf analyzersilence.cpp);
                    // not shown to user
```

src/analyzer/analyzersilence.cpp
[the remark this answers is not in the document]
davidlmorris: I don't know what changes that have been requested remain incomplete at this stage.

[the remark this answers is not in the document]
davidlmorris: Looking through the code, I can't see where cueinfo.m_label is being used, apart from a == comparison.  So, perhaps we can squeeze in the value (e.g -27db) and the version, so we can check this in future version updates should the code change for processing -60sb et al?

Have I understood this correctly?

I would then suggest I write another pull request as an update to:
1) Add an option to AutoDj preferences to set the intro and outro levels (in db presumably).
2) A Visual adjustment (like the existing intro and outro).   These should be kept separate from outro and intro cues, since at least I use these already to show safe zones for talking up to vocals and talking over outros.

src/analyzer/analyzersilence.cpp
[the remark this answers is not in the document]
davidlmorris: Since I wrote that, I have found two examples (both where the transitions start early) that are not ideal... and on two BIG songs... 'Stairway to Heaven" and "Bohemian Rhapsody".  This is a good ~~use case~~ example of where manually setting a transition start would be useful.  Alternatively or additionally, a setting in preferences so that each user can set the fade out (or even the fade in) level.

I don't see these as blockers to using this transition, it is after all a huge improvement as is, on the existing traditions.

davidlmorris:
Ping.

PR #13756, opened 2024-10-12
fix: add explicit virtual destructor in order to fix linker issues

### LINKED ISSUE #13755: 'undefined reference' after autofilereloader commit
### Bug Description

Hello,
After 45363b7c3107c41f224b44184972a821345f29b2 builds no longer work against main, failing when linking the mixxx executable with the errors below.
Fresh builds (on Gentoo x86, QT 6.7.2) on prior commits run to completion.
Unfortunately I don't have the Qt skills to attempt a patch.

Many thanks for a great piece of software
Matt
```
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlapplication.cpp.o): in function `mixxx::qml::QmlApplication::~QmlApplication()':
/home/matt/mixxx/src/qml/qmlapplication.cpp:80:(.text+0x11c6): undefined reference to `vtable for AutoFileReloader'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlapplication.cpp.o): in function `AutoFileReloader::~AutoFileReloader()':
/home/matt/mixxx/src/util/autofilereloader.h:9:(.text._ZN16AutoFileReloaderD2Ev[_ZN16AutoFileReloaderD5Ev]+0x1c): undefined reference to `vtable for AutoFileReloader'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlapplication.cpp.o): in function `AutoFileReloader::~AutoFileReloader()':
/home/matt/mixxx/src/util/autofilereloader.h:9:(.text._ZN16AutoFileReloaderD0Ev[_ZN16AutoFileReloaderD5Ev]+0x1c): undefined reference to `vtable for AutoFileReloader'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlautoreload.cpp.o): in function `mixxx::qml::QmlAutoReload::QmlAutoReload()':
/home/matt/mixxx/src/qml/qmlautoreload.cpp:15:(.text+0x234c): undefined reference to `AutoFileReloader::AutoFileReloader(RuntimeLoggingCategory const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: /home/matt/mixxx/src/qml/qmlautoreload.cpp:17:(.text+0x236e): undefined reference to `AutoFileReloader::fileChanged(QString const&)'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlautoreload.cpp.o): in function `QMetaObject::Connection QObject::connect<void (AutoFileReloader::*)(QString const&), void (mixxx::qml::QmlAutoReload::*)()>(QtPrivate::FunctionPointer<void (AutoFileReloader::*)(QString const&)>::Object const*, void (AutoFileReloader::*)(QString const&), QtPrivate::ContextTypeForFunctor<void (mixxx::qml::QmlAutoReload::*)(), void>::ContextType const*, void (mixxx::qml::QmlAutoReload::*&&)(), Qt::ConnectionType)':
/usr/include/qt6/QtCore/qobject.h:263:(.text+0x23d6): undefined reference to `AutoFileReloader::staticMetaObject'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlautoreload.cpp.o): in function `non-virtual thunk to mixxx::qml::QmlAutoReload::~QmlAutoReload()':
/home/matt/mixxx/src/qml/qmlautoreload.h:11:(.text._ZN5mixxx3qml13QmlAutoReloadD2Ev[_ZN5mixxx3qml13QmlAutoReloadD5Ev]+0x2f): undefined reference to `vtable for AutoFileReloader'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlautoreload.cpp.o): in function `mixxx::qml::QmlAutoReload::~QmlAutoReload()':
/home/matt/mixxx/src/qml/qmlautoreload.h:11:(.text._ZN5mixxx3qml13QmlAutoReloadD2Ev[_ZN5mixxx3qml13QmlAutoReloadD5Ev]+0x44f): undefined reference to `vtable for AutoFileReloader'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlautoreload.cpp.o): in function `non-virtual thunk to mixxx::qml::QmlAutoReload::~QmlAutoReload()':
/home/matt/mixxx/src/qml/qmlautoreload.h:11:(.text._ZN5mixxx3qml13QmlAutoReloadD0Ev[_ZN5mixxx3qml13QmlAutoReloadD5Ev]+0x33): undefined reference to `vtable for AutoFileReloader'
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: libmixxx-qml-lib.a(qmlautoreload.cpp.o): in function `mixxx::qml::QmlAutoReload::~QmlAutoReload()':
/home/matt/mixxx/src/qml/qmlautoreload.h:11:(.text._ZN5mixxx3qml13QmlAutoReloadD0Ev[_ZN5mixxx3qml13QmlAutoReloadD5Ev]+0x45f): undefined reference to `vtable for AutoFileReloader'
```

### Version

34dde6fc7a0128ecd92861b19aa063a6e7af6507

### OS

gentoo

Swiftb0y:
May fix #13755

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2792,6 +2792,8 @@ if(QML)
     # The following sources need to be in this target to get QML_ELEMENT properly interpreted
     src/control/controlmodel.cpp
     src/control/controlsortfiltermodel.cpp
+    # needed for qml/qmlautoreload.cpp
+    src/util/autofilereloader.cpp
   )
 
   # qt_finalize_target takes care that the resources :/mixxx.org/imports/Mixxx/
--- a/src/util/autofilereloader.cpp
+++ b/src/util/autofilereloader.cpp
@@ -6,6 +6,8 @@
 #include "moc_autofilereloader.cpp"
 #include "util/runtimeloggingcategory.h"
 
+AutoFileReloader::~AutoFileReloader() = default;
+
 AutoFileReloader::AutoFileReloader(const RuntimeLoggingCategory& loggingCategory)
         : m_loggingCategory(loggingCategory) {
     connect(&m_fileWatcher,
--- a/src/util/autofilereloader.h
+++ b/src/util/autofilereloader.h
@@ -10,6 +10,7 @@ class AutoFileReloader : public QObject {
     Q_OBJECT
   public:
     explicit AutoFileReloader(const RuntimeLoggingCategory& loggingCategory);
+    ~AutoFileReloader() override;
 
     void clear() {
         const auto files = m_fileWatcher.files();


daschuer:
For my understanding, without explicit declaration, the implicit default destructor is not virtual. This is an issue when a class is deleted by its base class. 
There are not much cases where this is desired so overriding the destructor is kind of mandatory. 

Can we add a CI check for this?

daschuer:
I have no explanation for the issue though, the AutoFileReloader is used as a member variable where this issue is not relevant.

daschuer:
Since https://github.com/mixxxdj/mixxx/commit/707a08f90316cafff215024f29b15a5ba8ed2029 fixes the issue, can we close this?

Swiftb0y:
> Can we add a CI check for this?

I don't think so. I would expect the compiler to catch this by itself already.

> And the ~QObject is virtual so this is even more not an issue here.

Yeah I agree. The implicit destructor was a red herring. Nevertheless, its probably better practice to have it declared and defined in the `.cpp` (because it can't benefit from the implicitly generated one in the header).

daschuer:
>  So closing PR doesn't make sense.

Oh yes, I see.  

autofilereloader.cpp is not used in mixxx-lib and can be removed here: 
https://github.com/mixxxdj/mixxx/blob/707a08f90316cafff215024f29b15a5ba8ed2029/CMakeLists.txt#L1150

> Nevertheless, its probably better practice to have it declared and defined in the .cpp (because it can't benefit from the implicitly generated one in the header).

Why? For my understanding it is better to remove the destructor again and allow the Compiler to inline it and finally optimize it away.

CMakeLists.txt
```diff
@@ -2792,6 +2792,8 @@ if(QML)
     # The following sources need to be in this target to get QML_ELEMENT properly interpreted
     src/control/controlmodel.cpp
     src/control/controlsortfiltermodel.cpp
+    # needed for qml/qmlautoreload.cpp
```
daschuer, line 2795: We normally don't have such comments, remove?

daschuer (changes requested):

CMakeLists.txt
ronso0: FYI I'm using autoreloader in #13082

Swiftb0y: Thanks for the headsup. well then it needs to be linked in both anyways.

Swiftb0y:
> Why? For my understanding it is better to remove the destructor again and allow the Compiler to inline it and finally optimize it away.

It can't inline the dtor though because its `virtual`. It may if the class was declared `final`, but even then only in certain cirumstances. Since its a QObject it will also have to call the QObject dtor at some point, which for sure can't be inlined because the PIMPL pattern. 

I would fully agree with your conclusion if this was [_TriviallyDestructible_](https://en.cppreference.com/w/cpp/types/is_destructible), but with QObjects, the very opposite is the case (they always have a virtual destructor). So the compiler will likely declare the function as inline in the header and let the linker deduplicate it. So in the binary, its effectively as if it was always in the translation unit. So the hope that the compiler can optimize much here doesn't apply.

I will try to reproduce in compiler explorer, but I don't have time right now.

daschuer (approved):
Thank you for explaining. LGTM 

Our library layout is still not optimal, but let's merge this now make main build again.

PR #13757, opened 2024-10-12
Fix Clazy v1.12 errors in 2.4

JoergAtGithub:
I ran Clazy v1.12 on our codebase, and got many errors of the following types:

- https://github.com/KDE/clazy/blob/master/docs/checks/README-range-loop-detach.md
- https://github.com/KDE/clazy/blob/master/docs/checks/README-fully-qualified-moc-types.md

This PR fixes them.

### DIFF
--- a/src/broadcast/broadcastmanager.cpp
+++ b/src/broadcast/broadcastmanager.cpp
@@ -42,8 +42,8 @@ BroadcastManager::BroadcastManager(SettingsManager* pSettingsManager,
     shout_init();
 
     // Initialize connections list from the current state of BroadcastSettings
-    QList<BroadcastProfilePtr> profiles = m_pBroadcastSettings->profiles();
-    for (const BroadcastProfilePtr& profile : std::as_const(profiles)) {
+    const QList<BroadcastProfilePtr> profiles = m_pBroadcastSettings->profiles();
+    for (const BroadcastProfilePtr& profile : profiles) {
         addConnection(profile);
     }
 
@@ -117,8 +117,8 @@ void BroadcastManager::slotControlEnabled(double v) {
     } else {
         m_pBroadcastEnabled->set(false);
         m_pStatusCO->forceSet(STATUSCO_UNCONNECTED);
-        QList<BroadcastProfilePtr> profiles = m_pBroadcastSettings->profiles();
-        for (BroadcastProfilePtr profile : std::as_const(profiles)) {
+        const QList<BroadcastProfilePtr> profiles = m_pBroadcastSettings->profiles();
+        for (BroadcastProfilePtr profile : profiles) {
             if (profile->connectionStatus() == BroadcastProfile::STATUS_FAILURE) {
                 profile->setConnectionStatus(BroadcastProfile::STATUS_UNCONNECTED);
             }
@@ -137,8 +137,8 @@ void BroadcastManager::slotProfileRemoved(BroadcastProfilePtr profile) {
 }
 
 void BroadcastManager::slotProfilesChanged() {
-    QVector<NetworkOutputStreamWorkerPtr> workers = m_pNetworkStream->outputWorkers();
-    for (const NetworkOutputStreamWorkerPtr& pWorker : std::as_const(workers)) {
+    const QVector<NetworkOutputStreamWorkerPtr> workers = m_pNetworkStream->outputWorkers();
+    for (const NetworkOutputStreamWorkerPtr& pWorker : workers) {
         ShoutConnectionPtr connection = qSharedPointerCast<ShoutConnection>(pWorker);
         if (connection) {
             BroadcastProfilePtr profile = connection->profile();
@@ -198,8 +198,8 @@ bool BroadcastManager::removeConnection(BroadcastProfilePtr profile) {
 }
 
 ShoutConnectionPtr BroadcastManager::findConnectionForProfile(BroadcastProfilePtr profile) {
-    QVector<NetworkOutputStreamWorkerPtr> workers = m_pNetworkStream->outputWorkers();
-    for (const NetworkOutputStreamWorkerPtr& pWorker : std::as_const(workers)) {
+    const QVector<NetworkOutputStreamWorkerPtr> workers = m_pNetworkStream->outputWorkers();
+    for (const NetworkOutputStreamWorkerPtr& pWorker : workers) {
         ShoutConnectionPtr connection = qSharedPointerCast<ShoutConnection>(pWorker);
         if (connection.isNull()) {
             continue;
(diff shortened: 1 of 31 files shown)


src/effects/presets/effectpreset.cpp
```diff
@@ -53,10 +53,12 @@ EffectPreset::EffectPreset(const EffectSlotPointer pEffectSlot)
     for (int parameterTypeId = 0; parameterTypeId < numTypes; ++parameterTypeId) {
         const EffectParameterType parameterType =
                 static_cast<EffectParameterType>(parameterTypeId);
-        for (const auto& pParameter : pEffectSlot->getLoadedParameters().value(parameterType)) {
+        const auto& loadedParameters = pEffectSlot->getLoadedParameters().value(parameterType);
+        for (const auto& pParameter : loadedParameters) {
```
Swiftb0y, line 57: how does that make a difference?

Swiftb0y (commented):
changes LGTM, though this does violate our "no mass reformatting" rule. Do we want to break it here? I'm slightly in favor.

JoergAtGithub:
> changes LGTM, though this does violate our "no mass reformatting" rule.

I don't think that the rule applies here, as this is fixing real code issues and not style issues.

src/effects/presets/effectpreset.cpp
JoergAtGithub: Have a look at https://stackoverflow.com/questions/35811053/using-c11-range-based-for-loop-correctly-in-qt

Swiftb0y: Ah, I found it in a comment on the solution. you can't call `std::as_const` on an rvalue (which is what I was suggesting). That solves that issue. the `const auto&` is a little weird here tbh (I'd argue just having a `const auto` would make more sense), but it works. I guess.

JoergAtGithub: Yes, all errors and warnings are gone

daschuer:
That is strange. The issue for my understanding applies only in cases where the object is shared. So it looks like we hit a lot of false positives. Has something changed in this topic?

daschuer:
OK, I found a related bug: https://invent.kde.org/sdk/clazy/-/issues/16 that explains the issue. 
Can you check all cases if making the temporary const is possible? That will also improve constant-correctness and remove some visual clutter.

src/broadcast/broadcastmanager.cpp
```diff
@@ -43,7 +43,7 @@ BroadcastManager::BroadcastManager(SettingsManager* pSettingsManager,
 
     // Initialize connections list from the current state of BroadcastSettings
     QList<BroadcastProfilePtr> profiles = m_pBroadcastSettings->profiles();
-    for (const BroadcastProfilePtr& profile : profiles) {
+    for (const BroadcastProfilePtr& profile : std::as_const(profiles)) {
```
daschuer, line 46: For instance here: 

```suggestion
    const QList<BroadcastProfilePtr> profiles = m_pBroadcastSettings->profiles();
    for (const BroadcastProfilePtr& profile : profiles) {
```

daschuer (changes requested):

node_modules/.cache/prettier/.prettier-caches/59798cdac4200f82f2c4ebba069ab1c61d895b67.json
Swiftb0y, line 1: remove?

src/soundio/soundmanagerconfig.cpp
```diff
@@ -513,9 +514,10 @@ void SoundManagerConfig::loadDefaults(SoundManager* soundManager, unsigned int f
     if (flags & SoundManagerConfig::DEVICES) {
         clearOutputs();
         clearInputs();
-        QList<SoundDevicePointer> outputDevices = soundManager->getDeviceList(m_api, true, false);
+        const QList<SoundDevicePointer> outputDevices =
+                soundManager->getDeviceList(m_api, true, false);
```
Swiftb0y, line 518: Have you made sure to double check that these are only applied to instances where you return a `const&`? If they return proper values, making them const may lead to other inefficiencies later (disabling move semantics primarily).

node_modules/.cache/prettier/.prettier-caches/59798cdac4200f82f2c4ebba069ab1c61d895b67.json
JoergAtGithub: Upps

src/soundio/soundmanagerconfig.cpp
JoergAtGithub: It should be only for the QLists, where Clazy recommended std::as_const and the list is never modified.

src/widget/whotcuebutton.cpp
```diff
@@ -108,7 +108,7 @@ void WHotcueButton::mousePressEvent(QMouseEvent* e) {
 
             CuePointer pHotCue;
             QList<CuePointer> cueList = pTrack->getCuePoints();
```
daschuer, line 110: This can also become const

src/broadcast/broadcastmanager.cpp
```diff
@@ -118,10 +118,10 @@ void BroadcastManager::slotControlEnabled(double v) {
         m_pBroadcastEnabled->set(false);
         m_pStatusCO->forceSet(STATUSCO_UNCONNECTED);
         QList<BroadcastProfilePtr> profiles = m_pBroadcastSettings->profiles();
```
daschuer, line 120: this can also be const

src/broadcast/broadcastmanager.cpp
```diff
@@ -138,7 +138,7 @@ void BroadcastManager::slotProfileRemoved(BroadcastProfilePtr profile) {
 
 void BroadcastManager::slotProfilesChanged() {
     QVector<NetworkOutputStreamWorkerPtr> workers = m_pNetworkStream->outputWorkers();
-    for (const NetworkOutputStreamWorkerPtr& pWorker : workers) {
+    for (const NetworkOutputStreamWorkerPtr& pWorker : std::as_const(workers)) {
```
daschuer, line 141: This can also be turned into a const

src/broadcast/broadcastmanager.cpp
```diff
@@ -199,7 +199,7 @@ bool BroadcastManager::removeConnection(BroadcastProfilePtr profile) {
 
 ShoutConnectionPtr BroadcastManager::findConnectionForProfile(BroadcastProfilePtr profile) {
     QVector<NetworkOutputStreamWorkerPtr> workers = m_pNetworkStream->outputWorkers();
-    for (const NetworkOutputStreamWorkerPtr& pWorker : workers) {
+    for (const NetworkOutputStreamWorkerPtr& pWorker : std::as_const(workers)) {
```
daschuer, line 202: This can also be const

src/broadcast/broadcastmanager.cpp
```diff
@@ -220,7 +220,7 @@ void BroadcastManager::slotConnectionStatusChanged(int newState) {
 
     // Collect status info
     QList<BroadcastProfilePtr> profiles = m_pBroadcastSettings->profiles();
-    for (BroadcastProfilePtr profile : profiles) {
+    for (BroadcastProfilePtr profile : std::as_const(profiles)) {
```
daschuer, line 223: const

daschuer (changes requested):
I have found some more cases where  const works, see comments. The rest LGTM

daschuer:
Thank you. LGTM.

PR #13759, opened 2024-10-13
Cues & Loops upgrade for Stems (& Widget)

Eve00000:
upgrade of cue/loop routines for stems

dbase needs to be updated with 4 fields in cues-table -> REAL + snake_case
alter table cues add stem_1_vol REAL DEFAULT 1.00 NOT NULL
alter table cues add stem_2_vol REAL DEFAULT 1.00 NOT NULL
alter table cues add stem_3_vol REAL DEFAULT 1.00 NOT NULL
alter table cues add stem_4_vol REAL DEFAULT 1.00 NOT NULL

when track has stems the mute and volume settings are saved in the hotcue / loop
else the values are all 1 (default)
volume -> 0-100 (= %) in widget, background double 0.0 .... 1.0
mute -> negative volume value
cue widget adapted to be able to change the stemvolumes after the cue is set.

the serato-markers give me grey hair.
20240913 - StemsCuesLoopsWidget


acolombier:
I haven't looked at the code yet as is doesn't seem to be ready to review looking at the CI result. My 2 cents on the overall look is that using integer for a volume that is a double feels redundant, so I would recommend using the same type between backend and persistency layer. 

In terms of feature, I am not sure that fitting this usecase on the existing hotcue is a wise idea. Perhaps we could consider another type of cue, eg. `stemcue`? This could be great as that would allow potentially even more stem customisation (eg FX set/dry-run volume)

Also, please note that to make this feature ready for review, the DB migration will need to be automated.

Eve00000:
>CI result. 
>integer for a volume that is a double

The serato-markers give me grey hair :-( I would like another eye on that. :-)
For a user volume needs to be a clear number, so a % is a clear value. Other double values as position, and length are also converted to integers to save in the cues-table in the dbase. Duration, replaygain are converted and saved as float in the library-table in the dbase.
I didn't want to add the dbase convertions in the pr before dbase names can be agreed.

daschuer:
The database should contain full precision, which is double in terms of COs, but float is precise enough for our 32 bit float engine. The representation in the GUI is unrelated. The proposed 0..100 % is OK we may consider to use 0.0 ... 1.00 or a dB scale.

src/engine/controls/cuecontrol.cpp
```diff
@@ -607,8 +611,8 @@ void CueControl::seekOnLoad(mixxx::audio::FramePos seekOnLoadPosition) {
 }
 
 void CueControl::cueUpdated() {
-    //auto lock = lockMutex(&m_mutex);
-    // We should get a trackCuesUpdated call anyway, so do nothing.
+    // auto lock = lockMutex(&m_mutex);
+    //  We should get a trackCuesUpdated call anyway, so do nothing.
```
daschuer, line 615: Unrelated change? Befor clang format we had the rule to use "//" without space for commented code and "// " for information.

src/engine/controls/cuecontrol.cpp
```diff
@@ -859,6 +863,10 @@ void CueControl::hotcueSet(HotcueControl* pControl, double value, HotcueSetMode
     mixxx::audio::FramePos cueStartPosition;
     mixxx::audio::FramePos cueEndPosition;
     mixxx::CueType cueType = mixxx::CueType::Invalid;
+    int passStem1Vol;
```
daschuer, line 866: What means the pass prefix?

src/engine/controls/cuecontrol.cpp
```diff
@@ -883,6 +891,54 @@ void CueControl::hotcueSet(HotcueControl* pControl, double value, HotcueSetMode
         }
     }
 
+    // EveCue-Loop
+    bool TrackStem = false;
+    if (ControlObject::exists(ConfigKey(getGroup(), "stem_count"))) {
+        PollingControlProxy proxyStem(getGroup(), "stem_count");
```
daschuer, line 897: we should avoid CO look up during Mixxx runtime, because this is a looking operation. Can you move this into the constructor or somewhere else? Checking for exist already creates the CO half way. 
You can instead create the PollingProxy with ControlFlag::AllowMissingOrInvalid flags.

src/engine/controls/cuecontrol.cpp
```diff
@@ -883,6 +891,54 @@ void CueControl::hotcueSet(HotcueControl* pControl, double value, HotcueSetMode
         }
     }
 
+    // EveCue-Loop
+    bool TrackStem = false;
```
daschuer, line 895: bools should be named for building a "sentences" with if. Like `if (hasStems) {` or such. And start with a lower case.

src/engine/controls/cuecontrol.cpp
```diff
@@ -883,6 +891,54 @@ void CueControl::hotcueSet(HotcueControl* pControl, double value, HotcueSetMode
...
+    if (ControlObject::exists(ConfigKey(getGroup(), "stem_count"))) {
+        PollingControlProxy proxyStem(getGroup(), "stem_count");
+        TrackStem = proxyStem.get() > 1;
+    }
+
+    if (TrackStem) {
+        const QString groupBaseName = getGroup().remove("[").remove("]");
```
daschuer, line 902: you remove the [ her and add it below.

daschuer (changes requested):

daschuer:
> Perhaps we could consider another type of cue, eg. stemcue?

What are the pros an cons?  

Without dedicated stem cues all hot-cues of stem will have volume values. 
This means that if you want to use the cue with the current stem layout there is a risk of unwanted volume change. 
I can also imagine to use cues after moving a single stem to a second deck. Recalling a cue there will then mess up the moved state. This are my argument for a dedicated cue type. 

What are the cons? Is there an alternative solution for a workaround?

src/engine/controls/cuecontrol.cpp
Eve00000: pass = pass 2 the 'cuecreation', maybe not clear, sorry

daschuer:
Oh I have jumped into detailed comments before sharing my opinion about this approach.

I consider this a great feature, than you :-) . It is useful to mark "teaser" parts a track where you want only vocals or another specific layout. It also is helpful to start a stem based transition with a predefined layout, where you for instance bring later the whole track in. 

Are there more use cases? 

> The serato-markers give me grey hair :-( I would like another eye on that. :-)

What do you mean by that? Can you explain it or share an external reference?

src/engine/controls/cuecontrol.cpp
Eve00000: this was my best silution to make [ChannelXStemY]  from [ChannelX]

Eve00000:
> Oh I have jumped into detailed comments before sharing my opinion about this approach.
> 
> I consider this a great feature, than you :-) . It is useful to mark "teaser" parts a track where you want only vocals or another specific layout. It also is helpful to start a stem based transition with a predefined layout, where you for instance bring later the whole track in.
> 
> Are there more use cases?
> 
> > The serato-markers give me grey hair :-( I would like another eye on that. :-)
> 
> What do you mean by that? Can you explain it or share an external reference?

Other use case: create reMixxxes, narrowing the line between DJ and producer.
example: I used the Daft Punks 'Da funk' drums  - loop with the piano of 9 to 5 and the piano of Think (Aretha Franklin), brought in the bassline of  another one bites the dust (cue -> loop)
The advantages = 
1. speed in preparing / mixing / switching
2. cpu load: preparing a cue with pfl takes cpu 
3.  creativity boost  using all possibilities of digital DJ-ing -> I have the correct sample / loop under my thumbs
4. fun : you can watch the reactions of the public, not just watching the screen

Eve00000:
> The serato-markers give me grey hair :-( I would like another eye on that. :-)

To let the serato cues match the 'adapted' cues, the need to have the 4 stem volumes, but  compiler says they are unused. If I remove them compiler says missing ....

Eve00000:
> Without dedicated stem cues all hot-cues of stem will have volume values. 
-> but they are only used on Stems

This means that if you want to use the cue with the current stem layout there is a risk of unwanted volume change. 
-> But it can easily be checked with right click.
-> Volume change of stems fappens immediately (so also in headphone check before mixxxing)
-> Maybe another color?

I can also imagine to use cues after moving a single stem to a second deck.
-> In this case there sould be a 'grey out' of cues where the selected stem is muted or volume = 0

src/engine/controls/cuecontrol.cpp
Eve00000: this happens automatically with the pre-commit since the update about 10 days ago (was an update of some minutes)

Eve00000:
> The database should contain full precision, which is double in terms of COs, but float is precise enough for our 32 bit float engine. The representation in the GUI is unrelated. The proposed 0..100 % is OK we may consider to use 0.0 ... 1.00 or a dB scale.

Do you mean you prefer to have 0.00 to 1.00 in the dbase in stead of 0 - 100 ?

acolombier:
> What are the pros an cons?

Another limitation I can think about is the limiting factor of having only stem volume - assuming you want to isolate a vocal sampler, you may also want to add a reverb + 25% dry/run volume.
In the meantime, adding all those parameters may scope creep hotcue usage, but it could also not so I'd be happy with that approach.
My only request would be to use a proto for the hotcue parameter definition, so we can evolve it overtime, and don't need DB migration

daschuer:
The cue table is defined here 
https://github.com/mixxxdj/mixxx/blob/60abf58e20af9e02c6475b0861008d4f074fa6fb/res/schema.xml#L79
and updated here: 
https://github.com/mixxxdj/mixxx/blob/60abf58e20af9e02c6475b0861008d4f074fa6fb/res/schema.xml#L441

In other columns we use sbake_case. so I think "stem_1_vol" does fit better. 

We had a discussion around "Spatial Audio", where we may have n channels and aposition in space. This kind of applies partially as well to instuments stored in a stem.  I have not any demand here but I just want to bring the question up if these four colums are the right choice or if we need something more generic like a meta data blob.

Something like 
* volume cue 
* effect cue 
* panning cue 

What do you think?

src/track/serato/markers.h
```diff
@@ -35,6 +35,10 @@ class SeratoMarkersEntry {
             bool hasEndPosition,
             int endPosition,
             SeratoStoredHotcueColor color,
+            int stem1vol,
```
daschuer, line 38: They are unused here and can be removed. 
Can you propose this change and we con look at the missing warning?

src/track/serato/markers.cpp
```diff
@@ -236,6 +236,10 @@ SeratoMarkersEntryPointer SeratoMarkersEntry::parseID3(const QByteArray& data) {
                     hasEndPosition,
                     endPosition,
                     color,
+                    100,
```
daschuer, line 239: This can be probably removed?

daschuer:
> Do you mean you prefer to have 0.00 to 1.00 in the dbase in stead of 0 - 100 ?

yes as a "real" value.

daschuer:
> My only request would be to use a proto for the hotcue parameter definition, so we can evolve it overtime, and don't need DB migration

Yes that make sense, use a protobuffer to store the volume information and probably a verison number to be backward compatible. 

We use it already to store the keys information in the data base: 
https://github.com/mixxxdj/mixxx/blob/60abf58e20af9e02c6475b0861008d4f074fa6fb/res/schema.xml#L351
https://github.com/mixxxdj/mixxx/blob/60abf58e20af9e02c6475b0861008d4f074fa6fb/src/proto/keys.proto

Eve00000:
> you may also want to add a reverb + 25% dry/run volume. 
-> this was indeed something I was thinking of as well, but I was in dubio because sometimes I want to use the FX-rack super in stead of the one fx selected on the stem. On the other hand to be consequent these settings should be saved as well.
Can I interpret your comment as 'add the fx - parameters also to the saved cue' ?

> scope creep .... proto for the hotcue parameter definition, so we can evolve it overtime, and don't need DB migration
-> I try to redefine your words in my language (I never adapted to programmers-slang): 
With starting to add parameters to the cues we know where we start but we don't know where we'll end. Maybe one day we add more possible settings to the stems that we can't forsee at the moment. To have a way to cope with the future you propose proto (googles memory buffer type, sort of json and xml?) 
With proto's you can define a way the needed information is stored in the memory. 
So at the moment you would store a 'message' id / track_id / ... / color / stem1vol / ... /stem4vol
but in the future this message can become 
id / track_id / ... / color / stem1vol / ... /stem4vol /.... parameterXsettingA / ... / parameterZsettingC/ ....
In order to keep the routines that work on the first parameters working you'll define a new 'message' that uses the new parameters for the new routines and keep the first message for the old routines.
Am I correct? (If so, I agree completely).

I just don't understand why you don't need a dbase migration then.
Is it because you already added a lot of fields at the beginning or is it by reusing fields for other meaning, like a second hotcue record with another 'type' that stores other parameters in the stemxvol fields?

excuse me for my ignorance

Eve00000:
> In other columns we use sbake_case. so I think "stem_1_vol" does fit better.
->Top.

> We had a discussion around "Spatial Audio" 
-> Yes with someone who wanted to create a 'spatial venue'
-> So you would like to add the fields for 'spatial' or use the same fields with another 'proto? I'm still not complete in the proto-system.

I think I asked once (at the beginning of the stems-implementation) to have a pan-knob besides the volume knob. 
But I stopped pushing my luck. Fort the record: 'STEMS ARE SUPER'
I would love the addition of pan-settings to a stem-track (and saving/loading it with the cues)
But spatial audio needs a kind of XY vector for the stem-tracks  fi drums on 40-40 bass 60 -40 melody 60-60 and vocals 40/60,
I don't want to hear it that way (except when it's an own mastered stem with only voices on the 4 tracks)

src/track/serato/markers.h
Eve00000: I'll do,

src/engine/controls/cuecontrol.cpp
Eve00000: fixed

src/engine/controls/cuecontrol.cpp
Eve00000: fixed

Eve00000:
> > Do you mean you prefer to have 0.00 to 1.00 in the dbase in stead of 0 - 100 ?
> Yes, I think it would make more sense.
-->fixed

src/track/serato/markers.cpp
Eve00000: fixed, there was another occurrence that was too much... solved Compiler happy

daschuer:
I gave that a second thought and consider how much we want to open padoras box. 

Hot cues are points in the track where you like to cue it in, (the whole track). We have one track, with one transport. All effects are NOT related naturally. 

We have recently introduced loop cues, used to recall loops. This are not Hot cues in the original terms, but since they are close related to transport it fits somehow mentally. 

We have also the use case of jumping around in a track using cues. This can be used to repeat or skip portions of a track or use the cur buttons like s drum pad. 

Changing the volume has never be part of cues, but I think here it is more used for selecting a certain stem layout, right. Somehow second dimension of curing. 

On the other hand we may treat the volume changes of stems like EQing, which is also not a part of cues.

You see, opening cues for stem volume may also create demands for EQs and than for the whole effect unit and whole Mixxx. 

All this are valid use cases. But I see three different feature: 
* Hot cues 
* Recording (redo) of DJ a performance/setting
* Create predefined samples

To which category dos the stem volume control belongs to. 

To be clear: I do not want to stop the project. I just want to discuss a perfect solution, to make sure that everything we introduce will align with it. Later we can decide if dedicated columns or a protbuf blob is the best fit and if it is am implicit feature if hot-cues or a new layer of controls.

daschuer:
That sounds good. So you actually looking for different thing than an advanced hot cue. It is a "performance cue" or such, right? 

Is it more like a recoded performance, or like something you would normally do in a pre recorded sampler from the sampler button? 

Do we also need to consider looping?  For this new cue type?

Eve00000:
> "advanced hot cue" -> "performance cue" or such, right?
> "recoded performance" -> sampler from the sampler button?
> Do we also need to consider looping? For this new cue type?

You're on track! 😄 ❤️ 
It's a cue, but a special one. In the development stage of stems I explaned my wish for stem-track-cues (not part of the whole track) that would make it possible to use the stem-track in a sampler with the cues of that track. 
In the ideal world, and I'm studying on @ronso0 's  "feat: allow swapping hotcues" #13394 I would like to drag the cue (you can call it performance cue) or the stemtrack (with stem-tracks own cues) to a sampler (as mentioned in https://mixxx.zulipchat.com/#narrow/channel/109122-general/topic/Hotcue.20stop.20mode), also with the length of the hotccue.
When a stem-track with it's own cues (volume for other stem-tracks = 0) and fx-settings is loaded in a sampler = slicing
When 1 'performance cue' is opened in a sampler (with defined length), the sampler becomes part of a 'soundboard' (thanks to OSC😄)
Thinking further: as bassdrum-kick / snare / handclap ... can be a 'performance cue' to, combined with loop and rate in the sampler we get a drummachine -> samplermachine (like fi DJS 1000)

PR #13760, opened 2024-10-13
Add script.deburr to common-controller-scripts to convert UTF-8 string to pure Latin-1

christophehenry:
This is copied and adapted from Lodash's `_.deburr`. Useful to transform UTF-8 strings for display on old devices with screens that can only display Latin-1 letters.

But maybe an implementation using a C++ library may be more efficient?

### DIFF
--- a/res/controllers/common-controller-scripts.js
+++ b/res/controllers/common-controller-scripts.js
@@ -245,6 +245,89 @@ script.midiDebug = function(channel, control, value, status, group) {
         " status: 0x" + status.toString(16) + " group: " + group);
 };
 
+
+
+class StringDeburr {
+    constructor() {
+        this.regexLatin = /[\xc0-\xd6\xd8-\xf6\xf8-\xff\u0100-\u017f]/g;
+
+        this.regexComboMark = /[\u0300-\u036f\ufe20-\ufe2f\u20d0-\u20ff]/g;
+
+        this.deburredLetters = {
+            // Latin-1 Supplement block.
+            "\xc0": "A",  "\xc1": "A", "\xc2": "A", "\xc3": "A", "\xc4": "A", "\xc5": "A",
+            "\xe0": "a",  "\xe1": "a", "\xe2": "a", "\xe3": "a", "\xe4": "a", "\xe5": "a",
+            "\xc7": "C",  "\xe7": "c",
+            "\xd0": "D",  "\xf0": "d",
+            "\xc8": "E",  "\xc9": "E", "\xca": "E", "\xcb": "E",
+            "\xe8": "e",  "\xe9": "e", "\xea": "e", "\xeb": "e",
+            "\xcc": "I",  "\xcd": "I", "\xce": "I", "\xcf": "I",
+            "\xec": "i",  "\xed": "i", "\xee": "i", "\xef": "i",
+            "\xd1": "N",  "\xf1": "n",
+            "\xd2": "O",  "\xd3": "O", "\xd4": "O", "\xd5": "O", "\xd6": "O", "\xd8": "O",
+            "\xf2": "o",  "\xf3": "o", "\xf4": "o", "\xf5": "o", "\xf6": "o", "\xf8": "o",
+            "\xd9": "U",  "\xda": "U", "\xdb": "U", "\xdc": "U",
+            "\xf9": "u",  "\xfa": "u", "\xfb": "u", "\xfc": "u",
+            "\xdd": "Y",  "\xfd": "y", "\xff": "y",
+            "\xc6": "Ae", "\xe6": "ae",
+            "\xde": "Th", "\xfe": "th",
+            "\xdf": "ss",
+            // Latin Extended-A block.
+            "\u0100": "A",  "\u0102": "A", "\u0104": "A",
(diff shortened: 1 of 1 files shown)


res/controllers/common-controller-scripts.js
```diff
@@ -245,6 +245,89 @@ script.midiDebug = function(channel, control, value, status, group) {
...
+class StringDeburr {
+    constructor() {
+        this.regexLatin = /[\xc0-\xd6\xd8-\xf6\xf8-\xff\u0100-\u017f]/g;
+
+        this.regexComboMark = /[\u0300-\u036f\ufe20-\ufe2f\u20d0-\u20ff]/g;
+
+        this.deburredLetters = {
+            // Latin-1 Supplement block.
+            "\xc0": "A",  "\xc1": "A", "\xc2": "A", "\xc3": "A", "\xc4": "A", "\xc5": "A",
+            "\xe0": "a",  "\xe1": "a", "\xe2": "a", "\xe3": "a", "\xe4": "a", "\xe5": "a",
```
Swiftb0y, line 259: have you considered using a `sparse` array instead? Maybe that would result in a speedup because indexing could be cheaper?

Swiftb0y:
I think in order to address the efficiency question, we would need to know how often this is run. I don't know the hardware/usecase for this so I can't really make an educated guess. Can you?

Swiftb0y:
> But maybe an implementation using a C++ library may be more efficient?

Ideally we could just wrap [`QStringConverted`](https://doc.qt.io/qt-6/qstringconverter.html#availableCodecs). Though efficiency would likely need to be benchmarked.

christophehenry:
> Though efficiency would likely need to be benchmarked.

Hmm… How would I do that.

res/controllers/common-controller-scripts.js
christophehenry: How do create a correspondance between UTF-8 chars and Latin-1 with sparse arrays? I don't get it.

Swiftb0y:
good question. Wbout just generating a really long string, running the conversions on that (potentially in a loop as well) and measure the time with `Date.now()`? Its only millisecond precision, which is not great, but if run it on enough data, I'm sure there will be some differences. 

That said, before wasting time on premature optimization, it would make more sense to determine how hot this really is. If you don't plan on calling more often than once per second on small enough data, it'll probably be fine.

christophehenry:
My main use case is displaying track names on DN-S3700 screen which should not happen that often. This is mainly a use case I have but I thought this could be useful for other controllers. However I though  maybe a implementation using Qt would be easier to maintain. Qt should have some unaccent algorithm?

Swiftb0y:
The only thing I found was `QStringConverter`. The direct methods on `QString` like `toLatin1` are just type-casts that result in UB if the underlying string does not conform. I think a wrapper over `QStringConverter` would be just as big as this little class, so I'm not sure which one is better. I mean there is a usecase for `UCS-2` for example (which is what pioneer hardware wants), but thats not guaranteed to be supported by `QStringConverter` either...

res/controllers/common-controller-scripts.js
```diff
@@ -245,6 +245,89 @@ script.midiDebug = function(channel, control, value, status, group) {
         " status: 0x" + status.toString(16) + " group: " + group);
 };
 
+
+
+class StringDeburr {
+    constructor() {
+        this.regexLatin = /[\xc0-\xd6\xd8-\xf6\xf8-\xff\u0100-\u017f]/g;
```
Swiftb0y, line 252: can you explain this regex a little more? From what I can tell, this doesn't match the layout on wikipedia. https://en.wikipedia.org/wiki/ISO/IEC_8859-1#Code_page_layout

christophehenry: Eeeh… No I can't. Most of this code is copied from Lodash's `_.deburr`, which I trust :sweat_smile:

Swiftb0y:
Mhmm. I'm not sure about this now. If I can't explain the code, I can't maintain it. Considering the usecase is quite niche to begin with, I'd prefer to keep this in the mapping for now. Is that okay for you?

christophehenry:
> I mean there is a usecase for UCS-2 for example

[Is it this](https://en.wikipedia.org/wiki/ISO/IEC_8859-15#Codepage_layout)? DN-S3700 [seems to support a superset of this](https://cdn.inmusicbrands.com/denondj/legacy/dns3700/dns3700-midi-command-list.pdf) (look page 8).

It may be useful to have a algorithm to translate UTF-8 into UCS-2 replacing unknow chars to `"?"`. It may even be possibile to make a 1:1 map with offset between UCS-2 code point and controller screen MIDI commands.

Swiftb0y:
> > I mean there is a usecase for UCS-2 for example
> 
> [Is it this](https://en.wikipedia.org/wiki/ISO/IEC_8859-15#Codepage_layout)? 

No UCS-2 is essentially UTF-16 but all codepoints must be 16-bit long. So it is/was a subset of UTF-16. So mostly unrelated to the `deburr` code. However, the DN-S3700 code does indeed seem to match the Latin1 code page layout.

christophehenry:
Ok, I started a discussion on Zulip. I'm closing this for now.

PR #13762, opened 2024-10-13
Fix check for unrelated decks playing when starting Auto DJ

### LINKED ISSUE #13734: Auto DJ: Wrong error message about playing deck 3 and 4 if they are uses as Auto DJ decks
### Bug Description

The message: 
"Decks 3 and 4 must be stopped to enable Auto DJ mode."
is wrong if deck 3 and 4 are used for Auto DJ. 

### Version

2.4.1

### OS

Ubunut Focal

daschuer:
This fixes #13734

Unfortunately this breaks string freeze. So we have the choice for keeping the check 2.4 broken for the uncommon case of using deck 3 or 4 for Auto DJ, but with a translated error or fix the check introducing an English error message.

What do you think. Let's adjust the merge target accordingly.

### DIFF
--- a/src/library/autodj/autodjprocessor.cpp
+++ b/src/library/autodj/autodjprocessor.cpp
@@ -401,6 +401,9 @@ AutoDJProcessor::AutoDJError AutoDJProcessor::toggleAutoDJ(bool enable) {
         // TODO: This is a total bandaid for making Auto DJ work with four decks.
         // We should design a nicer way to handle this.
         for (const auto& pDeck : std::as_const(m_decks)) {
+            VERIFY_OR_DEBUG_ASSERT(pDeck) {
+                continue;
+            }
             if (pDeck == pLeftDeck) {
                 continue;
             }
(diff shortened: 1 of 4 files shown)


src/library/autodj/autodjprocessor.cpp
```diff
@@ -398,14 +398,20 @@ AutoDJProcessor::AutoDJError AutoDJProcessor::toggleAutoDJ(bool enable) {
...
         }
 
-        // TODO: This is a total bandaid for making Auto DJ work with decks 3
-        // and 4.  We should design a nicer way to handle this.
-        for (int i = 2; i < m_decks.length(); ++i) {
-            if (m_decks[i] && m_decks[i]->isPlaying()) {
+        // TODO: This is a total bandaid for making Auto DJ work with four decks.
+        // We should design a nicer way to handle this.
+        for (const auto& pDeck : std::as_const(m_decks)) {
+            if (pDeck == pLeftDeck) {
```
Swiftb0y, line 404: this omits the null check that was previously there. Can we add a `DEBUG_ASSERT` just in case?

Swiftb0y, line 404: this omits the null check that was previously there. Can we add a `DEBUG_ASSERT` just in case?

daschuer: Done

Swiftb0y (approved):
LGTM. Thank you. I think violating string freeze is acceptable here.

daschuer:
Merge? Can one take care in @Swiftb0y's absence?

ronso0:
Merging this now, thanks @daschuer and @Swiftb0y !

PR #13764, opened 2024-10-14
fix: make `hotcue_focus_color_next`/`_prev` COs `ControlPushButton`s

Swiftb0y:
This fixes two issues:
1. The Control is not accessible via the controller mapping XML
2. The Control ignores nops, leading to confusing why `engine.setValue(group, ..., 1)` was only working on the first press (which is a confusion that comes up because this is not the case for other button-like controls, since those are implemented using `ControlPushButton`).

Both issues [were raised on Zulip](https://mixxx.zulipchat.com/#narrow/stream/113295-controller-mapping/topic/Problem.20with.20hotcue_focus_color_next.20and.20_prev).

### DIFF
--- a/src/engine/controls/cuecontrol.cpp
+++ b/src/engine/controls/cuecontrol.cpp
@@ -201,9 +201,9 @@ void CueControl::createControls() {
 
     m_pHotcueFocus = std::make_unique<ControlObject>(ConfigKey(m_group, "hotcue_focus"));
     setHotcueFocusIndex(Cue::kNoHotCue);
-    m_pHotcueFocusColorPrev = std::make_unique<ControlObject>(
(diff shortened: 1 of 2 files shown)


paulusr:
I tested this PR on Ubuntu 22.04 and can confirm it works as expected

Swiftb0y:
Thanks for confirming.

JoergAtGithub:
LGTM, but shouldn't this go into 2.4 ?

Swiftb0y:
yup, of course. Forgot to set the right branch when filing the PR.

JoergAtGithub:
Thank you!

Swiftb0y:
Thank you.

PR #13767, opened 2024-10-14
Automatically open PRs to sync branches

Holzhaus:
Related Zulip discussion: https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/Accidentental.20push.20to.20main.20reverted

Demo: https://github.com/mixxxdj/playground/pull/10

### DIFF
--- a/.github/workflows/sync_branches.yml
+++ b/.github/workflows/sync_branches.yml
@@ -19,7 +19,8 @@ jobs:
     steps:
       - name: Merge ${{ matrix.from_branch }} into ${{ matrix.to_branch }}
         if: github.event_name == 'workflow_dispatch' || github.ref_name == matrix.from_branch
-        uses: tretuna/sync-branches@1.4.0
+        # Commit is pinned due to low popularity of the action repo.
+        uses: tretuna/sync-branches@ea58ab6e406fd3ad016a064b31270bbb41127f41
         with:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
           FROM_BRANCH: ${{ matrix.from_branch }}
--- a/.github/workflows/sync_branches.yml
+++ b/.github/workflows/sync_branches.yml
@@ -16,16 +16,55 @@ jobs:
           - from_branch: "2.5"
             to_branch: "main"
     runs-on: ubuntu-latest
+    env:
+      SYNC_COMMITTER: github-actions[bot]
     steps:
-      - name: Merge ${{ matrix.from_branch }} into ${{ matrix.to_branch }}
-        if: github.event_name == 'workflow_dispatch' || github.ref_name == matrix.from_branch
-        # Commit is pinned due to low popularity of the action repo.
-        uses: tretuna/sync-branches@ea58ab6e406fd3ad016a064b31270bbb41127f41
-        with:
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-          FROM_BRANCH: ${{ matrix.from_branch }}
+      - name: "Check out repository"
+        uses: actions/checkout@v4.1.7
+
+      - name: "Install merge driver"
+        run: |
+          git config --global merge.ours.driver true
+
+      - name: "Check if merge branch already exists"
+        id: check_sync
+        run: |
+          if git fetch origin "${SYNC_BRANCH}"; then
+            echo "Branch ${SYNC_BRANCH} already exists, checking if the branch was modified..."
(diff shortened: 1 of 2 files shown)


.github/workflows/sync_branches.yml
```diff
@@ -0,0 +1,31 @@
+name: Sync Branches
+on:
+  push:
+    branches:
+      - "2.4"
+      - "2.5"
```
acolombier, line 6: Is is possible to use a pattern here, like you have in the labeller or does the API expect a static branch name?

.github/workflows/sync_branches.yml
```diff
@@ -0,0 +1,31 @@
...
+  workflow_dispatch:
+
+jobs:
+  sync-branches:
+    strategy:
+      matrix:
+        include:
+          - from_branch: "2.4"
+            to_branch: "2.5"
+          - from_branch: "2.5"
```
acolombier, line 16: Would we always backport via the `n-1` version? What I mean by that is, if there is a change in `2.4`, would we expect `2.4 -> 2.5` PR to be reviewed and merged, then `2.5 -> main`? Or should we consider `2.4 -> 2.5` and `2.4 ->  main` to be created in parallel? 
I'm thinking, there could be case we the `2.4 -> 2.5` (or any stable versions) needs to be delayed due to issues or further work needed to support the backport, but in the meantime, there wouldn't be a need to hold it for the dev branch (`main`)

acolombier (approved):
LGTM!

Just some small questions, but nothing preventing the merge.

.github/workflows/sync_branches.yml
Holzhaus: I think we should almost always just go to the next branch, not open multiple PRs from the same source. This just leads to more noise and confusion.

If we ever need this, nothing is stopping us from opening a PR manually 😉

.github/workflows/sync_branches.yml
Holzhaus: I think it's fnmatch syntax (glob-like), so we could use `[0123456789].[0123456789]`. But it's a bit pointless since we still need to hardcode the from/to branch mapping below.

Swiftb0y (approved):
LGTM. Thank you. Does it matter which branch this goes into? does still run the 2.4->2.5 job if it only lands in `main`?

Holzhaus:
> Does it matter which branch this goes into? does still run the 2.4->2.5 job if it only lands in main?

IIRC it suffices to have it in the default branch (`main`), but I'll double check on the playground repo.

Holzhaus:
Or I'll just rebase onto 2.4 to be on the safe side.

Swiftb0y:
Great thanks. Lets give everyone 24h to object before we merge.

JoergAtGithub (approved):

Holzhaus:
One sec. I think this is actually lacking the instruction from gitattributes

acolombier:
Is this ready to be merged? I was about to cherry-pick  #13913 into `2.5` and `main` but I think it would be a great test case for this feature!

Holzhaus:
I think this is not usable due to the translations mess. See https://github.com/mixxxdj/mixxx/pull/13767#issuecomment-2415184556.

daschuer:
In the past we the merge commit was always the auto generated message: 
`Merge remote-tracking branch 'upstream/2.5'` Can we keep it? 

In your Demo branch it is now: 

```
Merge pull request https://github.com/mixxxdj/playground/pull/10 from mixxxdj/2.5
Merge changes from `2.5` into `main`
```

It is unfortunately less obvious that this is no functional change.  


How shall it work finally? Collect all merged PRs until one of us presses merge? 
Or merge every single PR of a stable branch automatically? 

Current solution: 
* Wait for a contributor who merges and test locally 
* In case of conflicts, wait for an experienced contributor or 
* Do a preliminary merge as a PR, if this is successful confirmed merge manually

How about just model this: 
* Add a workflow that updates a single PR whenever a PR was merged in a version branch. 
* Add a text command to create a direct merge commit like we have before when all checks are green

Swiftb0y:
We've got 3 LGTM on this and no objections. Merge?

acolombier:
Merging that too late after `main` got broken by a failed merge... Hopefully this prevents it in the future!

daschuer:
It still seem to be broken. The red X on the 2.5 branch is really annoying. 
https://github.com/mixxxdj/mixxx/actions/runs/13553942275/job/37883894095

@Holzhaus will you have time to fix it?

Holzhaus:
See #14415

daschuer:
Did you look for existing actions? A quick look around discovered some however no exact fit. I wonder why, because our idea sms to be not exceptional.

PR #13772, opened 2024-10-15
Expose convertCharset convenience function to controllers

christophehenry:
Allow to convert from UTF-8 to whatever encoding the device supports

### DIFF
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -301,4 +301,33 @@ declare namespace engine {
      *               SoftStart with low factors would take a while until sound is audible. [default = 1.0]
      */
     function softStart(deck: number, activate: boolean, factor?: number): void;
+
+    enum WellKnownCharsets {
+        Latin1,
+        ISO_8859_1,
+        Latin9,
+        ISO_8859_15,
+        UCS2,
+        ISO_10646_UCS_2
+    }
+
+    /**
+     * Converts a string into another charset.
+     *
+     * This function is useful to display text on a device that does not make use of UTF-8.
+     * Available charset names are listed here: http://www.iana.org/assignments/character-sets/character-sets.xhtml.
+     * Characters that are unsupported by target charset will be transformed to null character (0x00).
+     * @param targetCharset The charset to encode the string into.
+     * @param value The string to encode
+     * @returns The converted String as an array of bytes. Will return an empty buffer on conversion error.
+     */
+    function convertCharset(targetCharset: string, value: string): ArrayBuffer
+
+    /**
+     * Version of {@link engine.convertCharset} to use with {@link engine.WellKnownCharsets}.
+     * @param targetCharset The charset to encode the string into.
+     * @param value The string to encode
+     * @returns The converted String as an array of bytes. Will return an empty buffer on conversion error.
+     */
+    function convertCharset(targetCharset: WellKnownCharsets, value: string): ArrayBuffer
 }
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -308,7 +308,116 @@ declare namespace engine {
         Latin9,
         ISO_8859_15,
         UCS2,
-        ISO_10646_UCS_2
+        ISO_10646_UCS_2,
+        UTF_8,
+        UTF_16,
+        UTF_16BE,
+        UTF_16LE,
+        UTF_32,
+        UTF_32BE,
+        UTF_32LE,
+        UTF_16BE_Version_1,
+        UTF_16LE_Version_1,
+        UTF_16_Version_1,
+        UTF_16_Version_2,
+        UTF_7,
+        SCSU,
+        BOCU_1,
+        CESU_8,
+        US_ASCII,
+        GB18030,
+        ISO_8859_2,
+        ISO_8859_3,
+        ISO_8859_4,
+        ISO_8859_5,
+        ISO_8859_6,
+        ISO_8859_7,
+        ISO_8859_8,
+        ISO_8859_9,
+        ISO_8859_10,
+        ISO_8859_13,
+        ISO_8859_14,
+        Shift_JIS,
+        EUC_JP,
+        Big5,
+        Big5_HKSCS,
+        GBK,
+        GB2312,
+        EUC_KR,
+        CP1363,
+        KSC_5601,
+        Windows_874_2000,
+        TIS_620,
+        IBM437, /** OEM United States */
+        IBM775, /** OEM Baltic */
+        IBM850, /** OEM Multilingual Latin 1 */
+        CP851,
+        IBM852, /** OEM Latin 2 */
+        IBM855, /** OEM Cyrillic */
+        IBM857, /** OEM Turkish */
+        IBM00858, /** OEM Multilingual Latin 1 + Euro symbol */
+        IBM860, /** OEM Portuguese */
+        IBM861, /** OEM Icelandic */
+        IBM862, /** OEM Hebrew */
+        IBM863, /** OEM Canadian French */
+        IBM864, /** OEM Arabic */
+        IBM865, /** OEM Nordic */
+        IBM866, /** OEM Russian */
+        IBM868, /** OEM Arabic */
+        IBM869, /** OEM Greek */
+        KOI8_R, /** KOI8-R */
+        KOI8_U, /** KOI8-U */
+        Windows_1250, /** Windows Central Europe */
+        Windows_1251, /** Windows Cyrillic */
+        Windows_1252, /** Windows Latin 1 */
+        Windows_1253, /** Windows Greek */
+        Windows_1254, /** Windows Turkish */
+        Windows_1255, /** Windows Hebrew */
+        Windows_1256, /** Windows Arabic */
+        Windows_1257, /** Windows Baltic */
+        Windows_1258, /** Windows Vietnamese */
+        Macintosh,
+        X_Mac_Greek,
+        X_Mac_Cyrillic,
+        X_Mac_CentralEuroRoman,
+        X_Mac_Turkish,
+        HP_Roman8,
+        Adobe_Standard_Encoding,
+        ISO_2022_JP,
+        ISO_2022_JP_1,
+        ISO_2022_JP_2,
+        ISO_2022_KR,
+        ISO_2022_CN,
+        ISO_2022_CN_EXT,
+        HZ_GB_2312,
+        IBM037, /** IBM EBCDIC US-Canada */
+        IBM273, /** IBM EBCDIC Germany */
+        IBM277, /** IBM EBCDIC Denmark-Norway */
+        IBM278, /** IBM EBCDIC Finland-Sweden */
+        IBM280, /** IBM EBCDIC Italy */
+        IBM284, /** IBM EBCDIC Latin America-Spain */
+        IBM285, /** IBM EBCDIC United Kingdom */
+        IBM290, /** IBM EBCDIC Japanese Katakana Extended */
+        IBM297, /** IBM EBCDIC France */
+        IBM420, /** IBM EBCDIC Arabic */
+        IBM424, /** IBM EBCDIC Hebrew */
+        IBM500, /** IBM EBCDIC International */
+        IBM_Thai,
+        IBM870, /** IBM EBCDIC Multilingual Latin 2 */
+        IBM871, /** IBM EBCDIC Icelandic */
+        IBM918, /** IBM EBCDIC Arabic */
+        IBM1026, /** IBM EBCDIC Turkish Latin 5 */
+        IBM1047, /** IBM EBCDIC Latin 1 Open System */
+        IBM01140, /** IBM EBCDIC US-Canada + Euro symbol */
+        IBM01141, /** IBM EBCDIC Germany + Euro symbol */
+        IBM01142, /** IBM EBCDIC Denmark-Norway + Euro symbol */
+        IBM01143, /** IBM EBCDIC Finland-Sweden + Euro symbol */
+        IBM01144, /** IBM EBCDIC Italy + Euro symbol */
+        IBM01145, /** IBM EBCDIC Latin America-Spain + Euro symbol */
+        IBM01146, /** IBM EBCDIC United Kingdom + Euro symbol */
+        IBM01147, /** IBM EBCDIC France + Euro symbol */
+        IBM01148, /** IBM EBCDIC International + Euro symbol */
+        IBM01149, /** IBM EBCDIC Icelandic + Euro symbol */
     }
 
     /**
@@ -317,14 +426,6 @@ declare namespace engine {
      * This function is useful to display text on a device that does not make use of UTF-8.
      * Available charset names are listed here: http://www.iana.org/assignments/character-sets/character-sets.xhtml.
      * Characters that are unsupported by target charset will be transformed to null character (0x00).
-     * @param targetCharset The charset to encode the string into.
-     * @param value The string to encode
-     * @returns The converted String as an array of bytes. Will return an empty buffer on conversion error.
-     */
-    function convertCharset(targetCharset: string, value: string): ArrayBuffer
-
-    /**
-     * Version of {@link engine.convertCharset} to use with {@link engine.WellKnownCharsets}.
      * @param targetCharset The charset to encode the string into.
      * @param value The string to encode
      * @returns The converted String as an array of bytes. Will return an empty buffer on conversion error.
--- a/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
@@ -7,6 +7,7 @@
 #endif
 #include <gsl/pointers>
 
+#include <QtEndian>
 #include "control/controlobject.h"
 #include "control/controlobjectscript.h"
 #include "control/controlpotmeter.h"
@@ -1064,39 +1065,279 @@ QByteArray ControllerScriptInterfaceLegacy::convertCharset(
     switch (targetCharset) {
     case WellKnownCharsets::Latin1:
     case WellKnownCharsets::ISO_8859_1:
-        return convertCharset(QStringLiteral("ISO-8859-1"), value);
+        return convertCharsetInternal(QStringLiteral("ISO-8859-1"), value);
     case WellKnownCharsets::Latin9:
     case WellKnownCharsets::ISO_8859_15:
-        return convertCharset(QStringLiteral("ISO-8859-15"), value);
+        return convertCharsetInternal(QStringLiteral("ISO-8859-15"), value);
     case WellKnownCharsets::UCS2:
     case WellKnownCharsets::ISO_10646_UCS_2:
-        return convertCharset(QStringLiteral("ISO-10646-UCS-2"), value);
+        return convertCharsetInternal(QStringLiteral("ISO-10646-UCS-2"), value);
+    case WellKnownCharsets::UTF_8:
+        return convertCharsetInternal(QStringLiteral("UTF-8"), value);
+    case WellKnownCharsets::UTF_16:
+        return convertCharsetInternal(QStringLiteral("UTF-16"), value);
+    case WellKnownCharsets::UTF_16BE:
+        return convertCharsetInternal(QStringLiteral("UTF-16BE"), value);
+    case WellKnownCharsets::UTF_16LE:
+        return convertCharsetInternal(QStringLiteral("UTF-16LE"), value);
+    case WellKnownCharsets::UTF_32:
+        return convertCharsetInternal(QStringLiteral("UTF-32"), value);
+    case WellKnownCharsets::UTF_32BE:
+        return convertCharsetInternal(QStringLiteral("UTF-32BE"), value);
+    case WellKnownCharsets::UTF_32LE:
+        return convertCharsetInternal(QStringLiteral("UTF-32LE"), value);
+    case WellKnownCharsets::UTF_7:
+        return convertCharsetInternal(QStringLiteral("UTF-7"), value);
+    case WellKnownCharsets::SCSU:
+        return convertCharsetInternal(QStringLiteral("SCSU"), value);
+    case WellKnownCharsets::BOCU_1:
+        return convertCharsetInternal(QStringLiteral("BOCU-1"), value);
+    case WellKnownCharsets::CESU_8:
+        return convertCharsetInternal(QStringLiteral("CESU-8"), value);
+    case WellKnownCharsets::US_ASCII:
+        return convertCharsetInternal(QStringLiteral("US-ASCII"), value);
+    case WellKnownCharsets::GB18030:
+        return convertCharsetInternal(QStringLiteral("GB18030"), value);
+    case WellKnownCharsets::ISO_8859_2:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-2"), value);
+    case WellKnownCharsets::ISO_8859_3:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-3"), value);
+    case WellKnownCharsets::ISO_8859_4:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-4"), value);
+    case WellKnownCharsets::ISO_8859_5:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-5"), value);
+    case WellKnownCharsets::ISO_8859_6:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-6"), value);
+    case WellKnownCharsets::ISO_8859_7:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-7"), value);
+    case WellKnownCharsets::ISO_8859_8:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-8"), value);
+    case WellKnownCharsets::ISO_8859_9:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-9"), value);
+    case WellKnownCharsets::ISO_8859_10:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-10"), value);
+    case WellKnownCharsets::ISO_8859_13:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-13"), value);
+    case WellKnownCharsets::ISO_8859_14:
+        return convertCharsetInternal(QStringLiteral("ISO-8859-14"), value);
+    case WellKnownCharsets::Shift_JIS:
+        return convertCharsetInternal(QStringLiteral("Shift_JIS"), value);
+    case WellKnownCharsets::EUC_JP:
+        return convertCharsetInternal(QStringLiteral("EUC-JP"), value);
+    case WellKnownCharsets::Big5:
+        return convertCharsetInternal(QStringLiteral("Big5"), value);
+    case WellKnownCharsets::Big5_HKSCS:
+        return convertCharsetInternal(QStringLiteral("Big5-HKSCS"), value);
+    case WellKnownCharsets::GBK:
+        return convertCharsetInternal(QStringLiteral("GBK"), value);
+    case WellKnownCharsets::GB2312:
+        return convertCharsetInternal(QStringLiteral("GB2312"), value);
+      case WellKnownCharsets::EUC_KR:
+        return convertCharsetInternal(QStringLiteral("EUC-KR"), value);
+    case WellKnownCharsets::CP1363:
+        return convertCharsetInternal(QStringLiteral("cp1363"), value);
+    case WellKnownCharsets::KSC_5601:
+        return convertCharsetInternal(QStringLiteral("KSC_5601"), value);
+    case WellKnownCharsets::Windows_874_2000:
+        return convertCharsetInternal(QStringLiteral("windows-874-2000"), value);
+    case WellKnownCharsets::TIS_620:
+        return convertCharsetInternal(QStringLiteral("TIS-620"), value);
+    case WellKnownCharsets::IBM437:
+        return convertCharsetInternal(QStringLiteral("IBM437"), value);
+    case WellKnownCharsets::IBM775:
+        return convertCharsetInternal(QStringLiteral("IBM775"), value);
+    case WellKnownCharsets::IBM850:
+        return convertCharsetInternal(QStringLiteral("IBM850"), value);
+    case WellKnownCharsets::CP851:
+        return convertCharsetInternal(QStringLiteral("cp851"), value);
+    case WellKnownCharsets::IBM852:
+        return convertCharsetInternal(QStringLiteral("IBM852"), value);
+    case WellKnownCharsets::IBM855:
+        return convertCharsetInternal(QStringLiteral("IBM855"), value);
+    case WellKnownCharsets::IBM857:
+        return convertCharsetInternal(QStringLiteral("IBM857"), value);
+    case WellKnownCharsets::IBM00858:
+        return convertCharsetInternal(QStringLiteral("IBM00858"), value);
+    case WellKnownCharsets::IBM860:
+        return convertCharsetInternal(QStringLiteral("IBM860"), value);
+    case WellKnownCharsets::IBM861:
+        return convertCharsetInternal(QStringLiteral("IBM861"), value);
+    case WellKnownCharsets::IBM862:
+        return convertCharsetInternal(QStringLiteral("IBM862"), value);
+    case WellKnownCharsets::IBM863:
+        return convertCharsetInternal(QStringLiteral("IBM863"), value);
+    case WellKnownCharsets::IBM864:
+        return convertCharsetInternal(QStringLiteral("IBM864"), value);
+    case WellKnownCharsets::IBM865:
+        return convertCharsetInternal(QStringLiteral("IBM865"), value);
+    case WellKnownCharsets::IBM866:
+        return convertCharsetInternal(QStringLiteral("IBM866"), value);
+    case WellKnownCharsets::IBM868:
+        return convertCharsetInternal(QStringLiteral("IBM868"), value);
+    case WellKnownCharsets::IBM869:
+        return convertCharsetInternal(QStringLiteral("IBM869"), value);
+    case WellKnownCharsets::KOI8_R:
+        return convertCharsetInternal(QStringLiteral("KOI8-R"), value);
+    case WellKnownCharsets::KOI8_U:
+        return convertCharsetInternal(QStringLiteral("KOI8-U"), value);
+    case WellKnownCharsets::Windows_1250:
+        return convertCharsetInternal(QStringLiteral("windows-1250"), value);
+    case WellKnownCharsets::Windows_1251:
+        return convertCharsetInternal(QStringLiteral("windows-1251"), value);
+    case WellKnownCharsets::Windows_1252:
+        return convertCharsetInternal(QStringLiteral("windows-1252"), value);
+    case WellKnownCharsets::Windows_1253:
+        return convertCharsetInternal(QStringLiteral("windows-1253"), value);
+    case WellKnownCharsets::Windows_1254:
+        return convertCharsetInternal(QStringLiteral("windows-1254"), value);
+    case WellKnownCharsets::Windows_1255:
+        return convertCharsetInternal(QStringLiteral("windows-1255"), value);
+    case WellKnownCharsets::Windows_1256:
+        return convertCharsetInternal(QStringLiteral("windows-1256"), value);
+    case WellKnownCharsets::Windows_1257:
+        return convertCharsetInternal(QStringLiteral("windows-1257"), value);
+    case WellKnownCharsets::Windows_1258:
+        return convertCharsetInternal(QStringLiteral("windows-1258"), value);
+    case WellKnownCharsets::Macintosh:
+        return convertCharsetInternal(QStringLiteral("macintosh"), value);
+    case WellKnownCharsets::X_Mac_Greek:
+        return convertCharsetInternal(QStringLiteral("x-mac-greek"), value);
+    case WellKnownCharsets::X_Mac_Cyrillic:
+        return convertCharsetInternal(QStringLiteral("x-mac-cyrillic"), value);
+    case WellKnownCharsets::X_Mac_CentralEuroRoman:
+        return convertCharsetInternal(QStringLiteral("x-mac-centraleurroman"), value);
+    case WellKnownCharsets::X_Mac_Turkish:
+        return convertCharsetInternal(QStringLiteral("x-mac-turkish"), value);
+    case WellKnownCharsets::HP_Roman8:
+        return convertCharsetInternal(QStringLiteral("hp-roman8"), value);
+    case WellKnownCharsets::Adobe_Standard_Encoding:
+        return convertCharsetInternal(QStringLiteral("Adobe-Standard-Encoding"), value);
+    case WellKnownCharsets::ISO_2022_JP:
+        return convertCharsetInternal(QStringLiteral("ISO-2022-JP"), value);
+    case WellKnownCharsets::ISO_2022_JP_1:
+        return convertCharsetInternal(QStringLiteral("ISO-2022-JP-1"), value);
+    case WellKnownCharsets::ISO_2022_JP_2:
+        return convertCharsetInternal(QStringLiteral("ISO-2022-JP-2"), value);
+    case WellKnownCharsets::ISO_2022_KR:
+        return convertCharsetInternal(QStringLiteral("ISO-2022-KR"), value);
+    case WellKnownCharsets::ISO_2022_CN:
+        return convertCharsetInternal(QStringLiteral("ISO-2022-CN"), value);
+    case WellKnownCharsets::ISO_2022_CN_EXT:
+        return convertCharsetInternal(QStringLiteral("ISO-2022-CN-EXT"), value);
+    case WellKnownCharsets::HZ_GB_2312:
+        return convertCharsetInternal(QStringLiteral("HZ-GB-2312"), value);
+    case WellKnownCharsets::IBM037:
+        return convertCharsetInternal(QStringLiteral("IBM037"), value);
+    case WellKnownCharsets::IBM273:
+        return convertCharsetInternal(QStringLiteral("IBM273"), value);
+    case WellKnownCharsets::IBM277:
+        return convertCharsetInternal(QStringLiteral("IBM277"), value);
+    case WellKnownCharsets::IBM278:
+        return convertCharsetInternal(QStringLiteral("IBM278"), value);
+    case WellKnownCharsets::IBM280:
+        return convertCharsetInternal(QStringLiteral("IBM280"), value);
+    case WellKnownCharsets::IBM284:
+        return convertCharsetInternal(QStringLiteral("IBM284"), value);
+    case WellKnownCharsets::IBM285:
+        return convertCharsetInternal(QStringLiteral("IBM285"), value);
+    case WellKnownCharsets::IBM290:
+        return convertCharsetInternal(QStringLiteral("IBM290"), value);
+    case WellKnownCharsets::IBM297:
+        return convertCharsetInternal(QStringLiteral("IBM297"), value);
+    case WellKnownCharsets::IBM420:
+        return convertCharsetInternal(QStringLiteral("IBM420"), value);
+    case WellKnownCharsets::IBM424:
+        return convertCharsetInternal(QStringLiteral("IBM424"), value);
+    case WellKnownCharsets::IBM500:
+        return convertCharsetInternal(QStringLiteral("IBM500"), value);
+    case WellKnownCharsets::IBM_Thai:
+        return convertCharsetInternal(QStringLiteral("IBM-Thai"), value);
+    case WellKnownCharsets::IBM870:
+        return convertCharsetInternal(QStringLiteral("IBM870"), value);
+    case WellKnownCharsets::IBM871:
+        return convertCharsetInternal(QStringLiteral("IBM871"), value);
+    case WellKnownCharsets::IBM918:
+        return convertCharsetInternal(QStringLiteral("IBM918"), value);
+    case WellKnownCharsets::IBM1026:
+        return convertCharsetInternal(QStringLiteral("IBM1026"), value);
+    case WellKnownCharsets::IBM1047:
+        return convertCharsetInternal(QStringLiteral("IBM1047"), value);
+    case WellKnownCharsets::IBM01140:
+        return convertCharsetInternal(QStringLiteral("IBM01140"), value);
+    case WellKnownCharsets::IBM01141:
+        return convertCharsetInternal(QStringLiteral("IBM01141"), value);
+    case WellKnownCharsets::IBM01142:
+        return convertCharsetInternal(QStringLiteral("IBM01142"), value);
+    case WellKnownCharsets::IBM01143:
+        return convertCharsetInternal(QStringLiteral("IBM01143"), value);
+    case WellKnownCharsets::IBM01144:
+        return convertCharsetInternal(QStringLiteral("IBM01144"), value);
+    case WellKnownCharsets::IBM01145:
+        return convertCharsetInternal(QStringLiteral("IBM01145"), value);
+    case WellKnownCharsets::IBM01146:
+        return convertCharsetInternal(QStringLiteral("IBM01146"), value);
+    case WellKnownCharsets::IBM01147:
+        return convertCharsetInternal(QStringLiteral("IBM01147"), value);
+    case WellKnownCharsets::IBM01148:
+        return convertCharsetInternal(QStringLiteral("IBM01148"), value);
+    case WellKnownCharsets::IBM01149:
+        return convertCharsetInternal(QStringLiteral("IBM01149"), value);
     default:
         m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unknown charset specified"));
         return QByteArray();
     }
 }
 
-QByteArray ControllerScriptInterfaceLegacy::convertCharset(
-        const QString& targetCharset, const QString& value) {
-    QByteArray encoderNameArray = targetCharset.toUtf8();
 #if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+// Add Byte Order Mark (BOM) to the beginning of the data if it is not already present
+static QByteArray addBom(const QByteArray& data) {
+    static const QByteArray bomBE = QByteArray::fromHex("FEFF");
+    static const QByteArray bomLE = QByteArray::fromHex("FFFE");
+
+    // Check system endianness
+    if (QSysInfo::ByteOrder == QSysInfo::BigEndian) {
+        if (!data.startsWith(bomBE)) {
+            return bomBE + data;
+        }
+    } else {
+        if (!data.startsWith(bomLE)) {
+            return bomLE + data;
+        }
+    }
+    return data;
+}
+#endif
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharsetInternal(
+        const QString& targetCharset, const QString& value) {
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+    QByteArray encoderNameArray = targetCharset.toUtf8();
     auto* pCodec = QTextCodec::codecForName(encoderNameArray);
     if (!pCodec) {
-        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
+        m_pScriptEngineLegacy->logOrThrowError(
+                QStringLiteral("Unable to get QTextCodec name for charset: %1").arg(targetCharset));
         return QByteArray();
     }
-    return pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull)->fromUnicode(value);
+    std::unique_ptr<QTextEncoder> encoder(
+            pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull));
+    if (encoderNameArray == "ISO-10646-UCS-2" || encoderNameArray == "UCS2") {
+        // For these encodings QTextCodec does not set, the BOM which QStringEncoder does
+        return addBom(encoder->fromUnicode(value));
+    }
+    return encoder->fromUnicode(value);
+
 #else
 #if QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)
-    QAnyStringView encoderName = QAnyStringView(encoderNameArray);
+    QAnyStringView encoderName = QAnyStringView(targetCharset);
 #else
+    QByteArray encoderNameArray = targetCharset.toUtf8();
     const char* encoderName = encoderNameArray.constData();
 #endif
     QStringEncoder fromUtf16 = QStringEncoder(
             encoderName, QStringEncoder::Flag::ConvertInvalidToNull);
     if (!fromUtf16.isValid()) {
-        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
+        m_pScriptEngineLegacy->logOrThrowError(
+                QStringLiteral("Unable to open encoder for charset: %1").arg(targetCharset));
         return QByteArray();
     }
     return fromUtf16(value);
--- a/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
@@ -1289,19 +1289,42 @@ QByteArray ControllerScriptInterfaceLegacy::convertCharset(
 }
 
 #if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
-// Add Byte Order Mark (BOM) to the beginning of the data if it is not already present
-static QByteArray addBom(const QByteArray& data) {
-    static const QByteArray bomBE = QByteArray::fromHex("FEFF");
-    static const QByteArray bomLE = QByteArray::fromHex("FFFE");
+// Enum to specify the encoding type
+enum class EncodingType {
+    UTF8,
+    UTF16,
+    UTF32,
+    UCS2
+};
 
-    // Check system endianness
-    if (QSysInfo::ByteOrder == QSysInfo::BigEndian) {
-        if (!data.startsWith(bomBE)) {
-            return bomBE + data;
+// Add Byte Order Mark (BOM) to the beginning of the data if it is not already present
+static QByteArray addBom(const QByteArray& data, QSysInfo::Endian byteOrder, EncodingType encodingType) {
+    static const QByteArray bomUTF8 = QByteArrayLiteral("\xEF\xBB\xBF");
+    static const QByteArray bomUTF16BE = QByteArrayLiteral("\xFE\xFF");
+    static const QByteArray bomUTF16LE = QByteArrayLiteral("\xFF\xFE");
+    static const QByteArray bomUTF32BE = QByteArrayLiteral("\x00\x00\xFE\xFF");
+    static const QByteArray bomUTF32LE = QByteArrayLiteral("\xFF\xFE\x00\x00");
+
+    if (data.startsWith(bomUTF8) || data.startsWith(bomUTF16BE) || data.startsWith(bomUTF16LE) ||
+            data.startsWith(bomUTF32BE) || data.startsWith(bomUTF32LE)) {
+        return data;
+    }
+
+    switch (encodingType) {
+    case EncodingType::UTF8:
+        return bomUTF8 + data;
+    case EncodingType::UTF16:
+    case EncodingType::UCS2:
+        if (byteOrder == QSysInfo::BigEndian) {
+            return bomUTF16BE + data;
+        } else {
+            return bomUTF16LE + data;
         }
-    } else {
-        if (!data.startsWith(bomLE)) {
-            return bomLE + data;
+    case EncodingType::UTF32:
+        if (byteOrder == QSysInfo::BigEndian) {
+            return bomUTF32BE + data;
+        } else {
+            return bomUTF32LE + data;
         }
     }
     return data;
@@ -1320,9 +1343,23 @@ QByteArray ControllerScriptInterfaceLegacy::convertCharsetInternal(
     }
     std::unique_ptr<QTextEncoder> encoder(
             pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull));
+    // For these encodings, add the BOM if not already present
     if (encoderNameArray == "ISO-10646-UCS-2" || encoderNameArray == "UCS2") {
-        // For these encodings QTextCodec does not set, the BOM which QStringEncoder does
-        return addBom(encoder->fromUnicode(value));
+        return addBom(encoder->fromUnicode(value), QSysInfo::ByteOrder, EncodingType::UCS2);
+    } else if (encoderNameArray == "UTF-16") {
+        return addBom(encoder->fromUnicode(value), QSysInfo::ByteOrder, EncodingType::UTF16);
+    } else if (encoderNameArray == "UTF-16LE") {
+        return addBom(encoder->fromUnicode(value), QSysInfo::LittleEndian, EncodingType::UTF16);
+    } else if (encoderNameArray == "UTF-16BE") {
+        return addBom(encoder->fromUnicode(value), QSysInfo::BigEndian, EncodingType::UTF16);
+    } else if (encoderNameArray == "UTF-32") {
+        return addBom(encoder->fromUnicode(value), QSysInfo::ByteOrder, EncodingType::UTF32);
+    } else if (encoderNameArray == "UTF-32LE") {
+        return addBom(encoder->fromUnicode(value), QSysInfo::LittleEndian, EncodingType::UTF32);
+    } else if (encoderNameArray == "UTF-32BE") {
+        return addBom(encoder->fromUnicode(value), QSysInfo::BigEndian, EncodingType::UTF32);
+    } else if (encoderNameArray == "UTF-8") {
+        return addBom(encoder->fromUnicode(value), QSysInfo::ByteOrder, EncodingType::UTF8);
     }
     return encoder->fromUnicode(value);
 
@@ -1334,7 +1371,7 @@ QByteArray ControllerScriptInterfaceLegacy::convertCharsetInternal(
     const char* encoderName = encoderNameArray.constData();
 #endif
     QStringEncoder fromUtf16 = QStringEncoder(
-            encoderName, QStringEncoder::Flag::ConvertInvalidToNull);
+            encoderName, QStringEncoder::Flag::ConvertInvalidToNull | QStringEncoder::Flag::WriteBom);
     if (!fromUtf16.isValid()) {
         m_pScriptEngineLegacy->logOrThrowError(
                 QStringLiteral("Unable to open encoder for charset: %1").arg(targetCharset));
--- a/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
@@ -1,5 +1,10 @@
 #include "controllerscriptinterfacelegacy.h"
 
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+#include <QTextCodec>
+#else
+#include <QStringEncoder>
+#endif
 #include <gsl/pointers>
 
 #include "control/controlobject.h"
@@ -1051,4 +1056,49 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double 
 
     // activate the ramping in scratchProcess()
     m_ramp[deck] = true;
+}
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
+        const ControllerScriptInterfaceLegacy::WellKnownCharsets targetCharset,
+        const QString& value) {
+    switch (targetCharset) {
+    case WellKnownCharsets::Latin1:
+    case WellKnownCharsets::ISO_8859_1:
+        return convertCharset(QStringLiteral("ISO-8859-1"), value);
+    case WellKnownCharsets::Latin9:
+    case WellKnownCharsets::ISO_8859_15:
+        return convertCharset(QStringLiteral("ISO-8859-15"), value);
+    case WellKnownCharsets::UCS2:
+    case WellKnownCharsets::ISO_10646_UCS_2:
+        return convertCharset(QStringLiteral("ISO-10646-UCS-2"), value);
+    default:
+        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unknown charset specified"));
+        return QByteArray();
+    }
+}
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
+        const QString& targetCharset, const QString& value) {
+    QByteArray encoderNameArray = targetCharset.toUtf8();
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+    auto* pCodec = QTextCodec::codecForName(encoderNameArray);
+    if (!pCodec) {
+        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
+        return QByteArray();
+    }
+    return pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull)->fromUnicode(value);
+#else
+#if QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)
+    QAnyStringView encoderName = QAnyStringView(encoderNameArray);
+#else
+    const char* encoderName = encoderNameArray.constData();
+#endif
+    QStringEncoder fromUtf16 = QStringEncoder(
+            encoderName, QStringEncoder::Flag::ConvertInvalidToNull);
+    if (!fromUtf16.isValid()) {
+        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
+        return QByteArray();
+    }
+    return fromUtf16(value);
+#endif
 }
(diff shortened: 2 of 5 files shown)


src/controllers/controller.h
```diff
@@ -1,8 +1,12 @@
 #pragma once
 
+#include <unicode/ucnv.h>
+#include <unicode/unistr.h>
```
Swiftb0y, line 4: can't help to notice that this looks like plain C headers (at least judging by the impl below). ICU also has a native C++ API afaict which is likely easier to use.

christophehenry: Ok. Do you know how I can use them?

Swiftb0y: Ah, just found the module list, apparently there is no C++ equivalent. https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/

src/controllers/controller.h
```diff
@@ -200,6 +206,39 @@ class ControllerJSProxy : public QObject {
...
+    // http://www.iana.org/assignments/character-sets/character-sets.xhtml
+    Q_INVOKABLE QJSValue convertEncoding(const QString& targetCharset, QString& value) {
+        std::shared_ptr<QJSEngine> pJsEngine = m_pController->getScriptEngine()->jsEngine();
+        VERIFY_OR_DEBUG_ASSERT(pJsEngine) {
+            return QJSValue::UndefinedValue;
+        }
+
+        icu::UnicodeString unicodeString;
+        UErrorCode errorCode;
+        UConverter* latin9Converter = ucnv_open(targetCharset.toLocal8Bit().data(), &errorCode);
```
Swiftb0y, line 219:
```suggestion
        UConverter* latin9Converter = ucnv_openU(targetCharset.data(), &errorCode);
```

src/controllers/controller.h
```diff
@@ -183,6 +187,8 @@ class Controller : public QObject {
     friend class MidiControllerTest;
 };
 
+class Charsets : public QObject {
+};
```
Swiftb0y, line 191: ?

src/controllers/controller.h
Swiftb0y: also consider using `std::unique_ptr` with a custom deleter.

src/controllers/controller.h
```diff
@@ -200,6 +206,39 @@ class ControllerJSProxy : public QObject {
...
+            return QJSValue::UndefinedValue;
+        }
+
+        char* result = nullptr;
+        ucnv_fromUChars(latin9Converter,
+                result,
+                0,
+                &value.data()->unicode(),
+                value.length(),
+                &errorCode);
```
Swiftb0y, line 232: this is completely wrong judging by the documentation. https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#a4e7ebc80f4ee5a02472b139dbaa99c0f

Swiftb0y:
I think design-wise, it would make most sense to just return a QByteArray, since if you're converting into a different format, you'll likely want to send out the bytes soon after without manipulating them more (at least in terms of characters). Also putting it back into a QString would either convert it back to UTF-16 or create a (likely invalid, UB causing) string.

src/controllers/controller.h
christophehenry: I wanted to expose some common charsets as constants to JS, maybe moving `convertEncoding` to that new object too so that from JS, it would look like: `engine.charsets.convertEncoding(engine.charsets.WELL_KNOWN_CHARSETS.LATIN_9)`, for instance. But I still havent figured out how to do that. I feel like I may be trying to bit off more that I can chew here, provided my C++ skills…

src/controllers/controller.h
christophehenry: It feels like it, yes. `UChar` is an alias to `char16_t` but I can't figure out how to get a `char16_t*` from QString.

christophehenry:
> it would make most sense to just return a QByteArray

Isn't [what I'm doing here](https://github.com/mixxxdj/mixxx/pull/13772/files#diff-3732186daf1acb1c62a232361fe0fd61fbfadfbe9d374c197cae263bd927370fR236)?

> Also putting it back into a QString

I don't understand… Where am I doing that?

src/controllers/controller.h
Swiftb0y: I think making it part of the engine class is unnecessary. you can simply add it to the global object, it requires a little bit of boilerplate but not much. See `ControllerScriptEngineBase::initialize()`.

src/controllers/controller.h
Swiftb0y: I think you want `QString::utf16` for reading. Then write directly into an appropriately [pre-allocated QByteArray](https://doc.qt.io/qt-6/qbytearray.html#QByteArray-4)

Swiftb0y:
> I don't understand… Where am I doing that?

whoops, yeah, that assumption was based on the `ucnv_fromUChars` call that currently has its arguments swapped.

src/controllers/controller.h
christophehenry: Oh you mean have `charsets` exposed in the global namespace? Aye. I could do that.

Swiftb0y: To add an enum, simply add it to the `Charsets` class and make it a `Q_ENUM`. Seems fairly doable IMO, I'm sure you can figure it out. If not, feel free to ask.

src/controllers/controller.h
christophehenry: `ucnv_openU` won't work. `QString::data()` returns `QChar*`.

src/controllers/controller.h
christophehenry: Wait… `QString::utf16()` returns `const ushort *` while `ucnv_fromUChars` expects `UChar*` as input and `char*` as output. How can that work? Can I just `(UChar*)QString::utf16()` and `(char*) myQByteArray`?

Swiftb0y: kind of. I'd add some checks though. like `static_assert(sizeof(ushort) == sizeof(UChar) && std::is_signed_v<ushort> == std::is_signed_v<UChar> && std::is_convertible_v<ushort, UChar>)`. Not sure thats sufficient but good enough. For the output, you can probably do `reinterpret_cast<char *>(QByteArray::data())` (**assuming you have pre-allocated a large enough buffer**).

src/controllers/controller.h
Swiftb0y: right, then try `QString::utf16()`.

src/controllers/controller.h
```diff
@@ -200,6 +206,39 @@ class ControllerJSProxy : public QObject {
...
+        ucnv_close(latin9Converter);
```
daschuer, line 233: Qt has a wrapper for all this. I think it should be: 
```suggestion
        QTextCodec* codec = QTextCodec::codecForName(targetCharset);
        if (!codec) {
            return QJSValue::UndefinedValue;
        }
        QByteArray result = codec->fromUnicode(value);
```
(Not tested)

JoergAtGithub: Unfortunately this nice API was replaced in Qt6 by QStringConverter, which supports much less codecs. The old API is still available in the "Qt 5 Core Compat module" for Qt6, but I guess this is nothing that should be used in new code.

daschuer: Mmm interesting. It would be:

```suggestion
        std::optional<QStringEncoder> encoder = QStringEncoder::forName(targetCharset);
        if (!encoder) {
            return QJSValue::UndefinedValue;
        }
        QByteArray result = encoder->encode(value);
```

daschuer: Our QT vcpkg is build with "icu" 
https://github.com/daschuer/vcpkg/blob/5d58718e04ee9196314dc77d7cf414f36be24b65/ports/qtbase/vcpkg.json#L62

christophehenry: @daschuer thank you for pointing this up. If I can't use `QTextCodec`, there's definitely some code I can copy here.

daschuer: I think you can use Qt6`s QStringEncoder

JoergAtGithub: Yes, but you need to use Qt 6.6 or higher.

daschuer: QStringEncoder Is available from Qt 6.0, Which part requires Qt 6.6?

JoergAtGithub: I looked it up, 6.4 is required, not 6.6. This is the version where support for ICU codecs was added: https://doc.qt.io/qt-6/whatsnew64.html

daschuer: I wonder what the limitations are because from 6.0 the targetCharset name was passed directly as string to ICU. 
Since there is nothing written about the version dependency we may test if it works anyway.

christophehenry: Ok, so Qt6.4. Am I guaranteed that Mixxx will be compiled using Qt6.4 or should I add a default behavior when Mixxx is compiled with Qt<6.4.

daschuer: I think our minimum is still Qt 6.2. The proposed code will work with that, but probably with some issue fixed in Qt 6.4 as @JoergAtGithub pointed out. 

My proposal is this: 

If it turns out your original use case works with 6.2, and QStringEncoderQStringEncoder, fine. Only if not, we may consider using ICU dirrectly or the QTextCodec compatibility layer. 

I wonder if QTextCodec than fixes the issue since it is based in the same implementation as QStringEncoder.

christophehenry: So if I'm not mistaken, this should look like on my last commit. This, however doesn't work as expected. From JS side, `convertEncoding` returns an object that, printed using `console.log`, display the original string. Not a `TypedArray`. I can't figure out what's happening. I'm leaving it ofr tonight.

Swiftb0y: Maybe do some type introspection (`instanceof`) and if it is an `ArrayBuffer`.

src/controllers/controller.h
```diff
@@ -193,13 +196,33 @@ class ControllerJSProxy : public QObject {
...
+        if (!codec) {
+            return nullptr;
+        }
+        return codec->fromUnicode(value);
+#else
+        QStringEncoder fromUtf8 = QStringEncoder(targetCharset.toUtf8().data());
+        if (!fromUtf8.isValid()) {
+            return nullptr;
+        }
+        return fromUtf8(value);
```
Swiftb0y, line 222: maybe an explicit cast to QByteArray will fix it?
```suggestion
        QStringEncoder fromUtf16 = QStringEncoder(targetCharset.toUtf8().data());
        if (!fromUtf16.isValid()) {
            return nullptr;
        }
        return QByteArray(fromUtf16(value));
```

src/controllers/controller.h
christophehenry: Ah you're right. I expected a `TypedArray` and thus a `.forEach()`. So I'd need to do `new Uint8Array(midi.convertEncoding("ISO-8859-15", "Thing to display"))` from JS? Is there a way to directly return a `TypedArray`?

src/controllers/controller.h
christophehenry: `DecodedData`  already has a `operator QByteArray()` and [that's what's documented](https://doc.qt.io/qt-6/qstringencoder.html#details).

christophehenry: No that was me. It correcty returns an `ArrayBuffer`. I just didn't know how to use it.

src/controllers/controller.h
```diff
@@ -193,13 +196,33 @@ class ControllerJSProxy : public QObject {
...
     // The length parameter is here for backwards compatibility for when scripts
     // were required to specify it.
     Q_INVOKABLE virtual void send(const QList<int>& data, unsigned int length = 0) {
         Q_UNUSED(length);
         m_pController->send(data, data.length());
     }
 
+    // Available charsets should be available here:
+    // http://www.iana.org/assignments/character-sets/character-sets.xhtml
+    Q_INVOKABLE QByteArray convertEncoding(const QString& targetCharset, const QString& value) {
```
christophehenry, line 210: So I can expose enums to `QJSEngine` but C++ enum only accepts ints as value. Is there a way that I can use strings as value or modify the signature of this function to accepts either a QString or an enum?

src/controllers/controller.h
```diff
@@ -193,13 +196,33 @@ class ControllerJSProxy : public QObject {
...
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+        auto* codec = QTextCodec::codecForName(targetCharset.toUtf8());
```
christophehenry, line 212: It seems like I can't `delete codec` because `QTextCodec` destrucotr is `protected`. Won't this cause a resource leak problem?

src/controllers/controller.h
```diff
@@ -193,13 +196,33 @@ class ControllerJSProxy : public QObject {
...
+        if (!codec) {
+            return nullptr;
+        }
+        return codec->fromUnicode(value);
+#else
+        QStringEncoder fromUtf8 = QStringEncoder(targetCharset.toUtf8().data());
```
christophehenry, line 218: `QStringEncoder` and `QTextCodec` seem to allocate resources that can be reused. Is it a good idea, performance-wise, to cache them in a `QMap` or something?

src/controllers/controller.h
Swiftb0y: You can technically execute some JS code in C++ that essentially does that and return the resulting QJSValue, but thats quite ugly IMO. I think it should up to the caller how they want to interpret that buffer. If thats too much typing for you, you can always create a utility function in your mapping, but lets not bake that into the API. It will cause API inconsistencies and unnecessary overhead.

src/controllers/controller.h
Swiftb0y: only if the pointer returned is owning. Thats why plain pointers are bad, they could be owning or non-owning and you never know (nor can be sure either is always the case) because that information is not in the type system. In this case, I'm fairly certain its non-owning so you don't need to do anything with the pointer afterwards (it could just as well be a reference being returned instead of a pointer). This is probably why the destructor is protected, to protect the user from calling delete on it.

src/controllers/controller.h
Swiftb0y: I don't think so. From what I can tell (by only looking at the docs, not the source), it supports encoding in chunks instead of all at once, so it needs internal state there. That state should probably not be reused across unrelated strings.

src/controllers/controller.h
Swiftb0y: you can take a QJSValue (which acts as a QVariant) and then switch internally depending on the type of the contained value. I'm not sure if that complexity is worth it though. I would prefer to have an enum exposed. A string would make more sense if the set of encoding is open, but in this case its closed because we're limited by `QStringConverter::Encoding`.

christophehenry:
I can confirm, this code works on DN-S3700.

christophehenry:
Ok, so I can confirm this code works.

src/controllers/controller.h
christophehenry: Oh no, since 6.4, the set is opened with `QStringConverter`. That's why my last commit uses `QTextCodec` prior to 6.4. Virtually any encoding supported by UCI should be available. See my last comment, I used Latin-9 which is not available in `QStringConverter::Encoding`.

Swiftb0y: Actually, there is probably a little more we can do here. Apparently there are more optional codecs available using [`QStringConverter::availableCodecs()`](https://doc.qt.io/qt-6/qstringconverter.html#availableCodecs) which we might want to support. In this case we should probably mirror the `QStringEncoder` api a little and have a function with the enum and a function with the string. I wouldn't try to emulate that via the QJSValue overload stuff though because the manual dispatch is annoying. Instead I would create two different `Q_INVOKEABLE` methods that are implemented in terms of the correct `QTextEncoder` overload. I'd all add a another `Q_VERSION_CHECK`, so we can skip the UTF16 to UTF8 conversion for > Qt 6.8.

Swiftb0y: > Oh no, since 6.4, the set is opened with `QStringConverter`. That's why my last commit uses `QTextCodec` prior to 6.4. Virtually any encoding supported by UCI should be available. See my last comment, I used Latin-9 which is not available in `QStringConverter::Encoding`.

Yeah, you're right. I didn't see that comment when I wrote the last one.

src/controllers/controller.h
christophehenry: There's a `resetState` method that can be called first, though.

Swiftb0y: yeah, but I don't think its worth the complexity. Taking a quick glance at the source, there is no heap allocation or anything else expensive that is worth reusing. We can KISS for now and optimize later.

christophehenry: Ok. this conversation is resolved then.

daschuer (changes requested):
Some nit picks.

src/controllers/controller.h
```diff
@@ -193,13 +196,33 @@ class ControllerJSProxy : public QObject {
             : m_pController(m_pController) {
     }
 
+    QHash<QString, QString> m_cache;
```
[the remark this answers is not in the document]
daschuer: What, is this unused?

JoergAtGithub (changes requested):

src/controllers/controller.h
```diff
@@ -16,7 +18,7 @@ class Controller : public QObject {
     Q_OBJECT
   public:
     explicit Controller(const QString& deviceName);
-    ~Controller() override;  // Subclass should call close() at minimum.
+    ~Controller() override; // Subclass should call close() at minimum.
```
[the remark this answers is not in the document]
christophehenry: Hmm… This is a `pre-commit` fix, isn't it?

src/controllers/controller.h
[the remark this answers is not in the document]
christophehenry: Ah no, sorry. This was left from a previous experimentation. This must be removed.

src/controllers/controller.h
```diff
@@ -193,13 +196,33 @@ class ControllerJSProxy : public QObject {
...
+
     // The length parameter is here for backwards compatibility for when scripts
     // were required to specify it.
     Q_INVOKABLE virtual void send(const QList<int>& data, unsigned int length = 0) {
         Q_UNUSED(length);
         m_pController->send(data, data.length());
     }
 
+    // Available charsets should be available here:
+    // http://www.iana.org/assignments/character-sets/character-sets.xhtml
```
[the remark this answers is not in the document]
Swiftb0y: alternatively we just let the mapping handle it. Not sure whats better.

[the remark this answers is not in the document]
christophehenry: Meh. I'm not in favor of that. The idea is to allow controller developers to use charsets if available without having to open a PR to make it available. I can, however check against `QStringConverter::availableCodecs()` and `QTextCodec::availableCodecs()`.

src/controllers/controller.h
```diff
@@ -16,7 +18,7 @@ class Controller : public QObject {
     Q_OBJECT
   public:
     explicit Controller(const QString& deviceName);
```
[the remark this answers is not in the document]
christophehenry: I think it makes sense since this is computing data that it supposed to be send through midi. Isn't it adding complexity to add yet another interface in the global object; in particular an interface with only one function?

christophehenry:
Should I expose `QStringConverter::availableCodecs()`/`QTextCodec::availableCodecs() to JS too, so that developers can chek what is available on their environment?

src/controllers/controller.h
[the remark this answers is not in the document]
JoergAtGithub: Every Mixxx installation must support each mapping. The mapping API must be stable accross installations. There is no problem, to add a long list of charsets to the positive list.

src/controllers/controller.h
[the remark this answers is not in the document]
JoergAtGithub: The `engine` class is already there, and used for nearly every function in the mapping API other than raw IO. There is no overhead, just a different existing class.

JoergAtGithub:
> Should I expose `QStringConverter::availableCodecs()`/`QTextCodec::availableCodecs()` to JS too, so that developers can chek what is available on their environment?

No, the API must guarantee to support the same charsets o all platforms and builds of Mixxx.

src/controllers/controller.h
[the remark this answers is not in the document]
Swiftb0y: yeah, but pre-commit only touches the lines in the diff.

Swiftb0y:
> No, the API must guarantee to support the same charsets on all platforms and builds of Mixxx.

That'll limit us to a somewhat small common subset though. Is that a price we're willing to pay? Do we want to tell a contributor "no we can not accept your mapping because one niche feature doesn't work on one particular platform"?

src/controllers/controller.h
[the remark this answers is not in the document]
christophehenry: It makes less sense to me but I won't fight on this and move to the `engine` object.

JoergAtGithub:
> That'll limit us to a somewhat small common subset though.

I don't think so, if you use a Qt build with the same ICU version, you will get always the same charsets. We just need to ensure, that the configuration of the Mixxx build fails, if someone uses an inproper Qt build.

christophehenry:
Plus, it's not as if it poses the risk of a crash. At worst, the function would return `undefined` and the mapping can propose an alternative behavior.

res/controllers/engine-api.d.ts
```diff
@@ -301,4 +301,15 @@ declare namespace engine {
...
     function softStart(deck: number, activate: boolean, factor?: number): void;
+
+    /**
+     * Converts a string into another charset.
+     *
+     * This function is useful to display text on a device that does not make use of UTF-8.
+     * Available charset names are listed here: http://www.iana.org/assignments/character-sets/character-sets.xhtml
+     * @param {string} targetCharset The charset to encode the string into.
+     * @param {string} value The string to encode
+     * @returns {ArrayBuffer | undefined}The converted String as an array of bytes or undefined if an error happened when performing conversion
```
christophehenry, line 312: Maybe `jsdoc/require-param-type` and `jsdoc/require-returns-type` should be disabled since we're docmenting tpyes using JS.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1047,39 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+    auto* pCodec = QTextCodec::codecForName(targetCharset.toUtf8());
+    if (!pCodec) {
+        return nullptr;
+    }
+    return pCodec->fromUnicode(value);
+#else
+    QStringEncoder fromUtf8 = QStringEncoder(targetCharset.toUtf8().data());
+    if (!fromUtf8.isValid()) {
+        return nullptr;
```
christophehenry, line 1081: Wait, that doesn't translate into `undefined`. Am I really forced to use QJSValue as a return type?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
```diff
@@ -17,6 +17,17 @@ class ConfigKey;
...
+    enum WellKnownCharsets {
+        LATIN_9,
+        ISO_8859_15,
+        LATIN_1,
+        ISO_8859_1,
+        UCS_2,
+        ISO_10646_UCS_2
+
+    };
+    Q_ENUM(WellKnownCharsets)
```
christophehenry, line 29: Dammit! I can't find how to expose this enum under `engine.WellKnownCharsets`. Can someone help me?

JoergAtGithub: You need to tell the JavaScript-Interpreter (QJSEngine) about it. Something like this:
```C++
  void registerEnums(QJSEngine& engine) {
      qRegisterMetaType<mixxx::preferences::constants::WellKnownCharsets>("WellKnownCharsets");

      const QMetaObject& metaObject = mixxx::preferences::constants::staticMetaObject;
      QMetaEnum metaEnum = metaObject.enumerator(metaObject.indexOfEnumerator("WellKnownCharsets"));

      QJSValue enumObject = engine.newObject();
      for (int enumIdx = 0; enumIdx  < metaEnum.keyCount(); ++enumIdx ) {
          enumObject.setProperty(metaEnum.key(enumIdx ), metaEnum.value(i));
      }

      engine.globalObject().setProperty("WellKnownCharsets", enumObject);
  }
```

src/controllers/controller.h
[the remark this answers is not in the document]
daschuer: How likely is the case that QT is build without ICU? It is a core feature of every Linux distribution and for Window an macOs we provide our own solution. 

In the unlikely case a certain codec is not available we get an empty string. This is mich better than failed mapping. 

So I guess we can be relaxed here and use an easy solution like a simple isEmpt() sanity check or such.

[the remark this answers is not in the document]
christophehenry: > In the unlikely case a certain codec is not available we get an empty string

No. Current implementation returns `undefinedè and I documented that it will return `undefined` if charset is unavailable. This lets mapping developers implement an alternative behavior.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: if you want to return undefined, yes. you can also return an empty array buffer though.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
Swiftb0y: use `Q_ENUM`, its way simpler. https://stackoverflow.com/questions/59728876/how-to-register-an-enum-type-to-qjsengine-to-be-used-from-the-scripting-environm

christophehenry: I tried that. It exposes an empty object. Here is the code I tried:

```diff
diff --git a/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp b/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
index 22978c5f5c..89d9f02538 100644
--- a/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
@@ -415,8 +415,11 @@ bool ControllerScriptEngineLegacy::initialize() {
     ControllerScriptInterfaceLegacy* legacyScriptInterface =
             new ControllerScriptInterfaceLegacy(this, m_logger);
 
-    engineGlobalObject.setProperty(
-            "engine", m_pJSEngine->newQObject(legacyScriptInterface));
+    auto engine = m_pJSEngine->newQObject(legacyScriptInterface);
+    auto meta = m_pJSEngine->newQMetaObject(&ControllerScriptInterfaceLegacy::staticMetaObject);
+
+    engine.setProperty("WellKnownCharsets", meta);
+    engineGlobalObject.setProperty("engine", engine);
 
 #ifdef MIXXX_USE_QML
     if (m_bQmlMode) {
```

```js
console.error(Object.keys(engine.WellKnownCharsets));
```

displays `[]`.

christophehenry:
Should I add tests to this feature?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
Swiftb0y: Okay, shooting a little in the dark here, but have you just tried `WellKnownCharsets.LATIN_9` in the JS? is it still `undefined`?

Apart from that, try deleting the `engine.setProperty("WellKnownCharsets", meta);` line.

christophehenry: Oh, no, wait, that code actually works! I'm so sorry for the noise…

christophehenry:
Ok, so the `QJSValue` implementation don't pass the tests and make Mixxx segfault when using `convertCharset` in script, and the `QByteArray` don't let me return `undefined` on error… I'm a bit lost here, I admit…

Swiftb0y:
I think just returning an empty QByteArray in the error case is fine.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1050,39 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
     m_ramp[deck] = true;
 }
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
+        const ControllerScriptInterfaceLegacy::WellKnownCharsets targetCharset,
+        const QString& value) {
+    switch (targetCharset) {
+    case WellKnownCharsets::LATIN_1:
+    case WellKnownCharsets::ISO_8859_1:
+        return this->convertCharset("ISO-8859-1", value);
```
Swiftb0y, line 1060: QStringLiteral

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1050,39 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+        return nullptr;
+    }
+}
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
+        const QString& targetCharset, const QString& value) {
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+    auto* pCodec = QTextCodec::codecForName(targetCharset.toUtf8());
+    if (!pCodec) {
+        return nullptr;
```
Swiftb0y, line 1078: just default-construct it (which is even nicely `constexpr`). If you return a nullptr, it will call [`QByteArray(char*, size = -1)`](https://doc.qt.io/qt-6/qbytearray.html#QByteArray-3) instead. Same above and below.
```suggestion
        return QByteArray();
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1050,39 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+    }
+    return pCodec->fromUnicode(value)
+#else
+    QStringEncoder fromUtf8 = QStringEncoder(targetCharset.toUtf8().data());
```
Swiftb0y, line 1082: please add another version check so we don't do the `utf8` conversion on >Qt6.8. Also `fromUtf8` is an incorrect name. It should be `fromUtf16` (which is what QStrings are encoded in).

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
```diff
@@ -17,6 +17,17 @@ class ConfigKey;
 class ControllerScriptInterfaceLegacy : public QObject {
     Q_OBJECT
   public:
+    enum WellKnownCharsets {
```
Swiftb0y, line 20: consider using `enum class` and `using enum WellKnownCharsets` to avoid implicit conversions to/from `int`.

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -745,3 +745,29 @@ TEST_F(ControllerScriptEngineLegacyTest, screenWillSentRawDataIfConfigured) {
...
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetCorrectValueWellKnown) {
+    const auto result = evaluate(R"Javascript(
+        Array.from(
+            new Uint8Array(
+                engine.convertCharset(engine.WellKnownCharsets.LATIN_9, "Hello!")))
+        .map(it => it.toString(16))
+        .join(",")
+    )Javascript");
+    ASSERT_QSTRING_EQ(result.toString(), "48,65,6c,6c,6f,21");
+}
```
Swiftb0y, line 762: I'd prefer something like this:
```suggestion
    const auto result = evaluate(R"Javascript(
        engine.convertCharset(engine.WellKnownCharsets.LATIN_9, "Hello!"))
    )Javascript");
    EXPECT_EQ(jsEngine()->toScriptValue<QByteArray>(result), QByteArrayLiteral("Hello!"));
}
```

Swiftb0y (changes requested):

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
christophehenry: I'm not sure I'm satisfied with empty array being the marker for an error. Are you sure there isn't a valid use case for returning an empty array? For instance passing an empty string? I know that sounds like a silly use-case bit IDK…

> which is even nicely `constexpr`

WDYM? That I can `return constexpr QByteArray()` or that the compiler automatically `constexpr` a `return QByteArray()`?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
christophehenry: Wait, where should I use `using enum WellKnownCharsets`?

src/test/controllerscriptenginelegacy_test.cpp
christophehenry: Wait, how can that work? There shouldn't be the same bytes in the two arrays 🤔

christophehenry:
I ended up writing an implementation using `QJSValue` and dumping the data in a JS array. WDYT?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: well, it can knows how to construct the qbytearray inplace while for the other other constructor it needs to jump into the Qt code. See [this example](https://compiler-explorer.com/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:20,endLineNumber:7,positionColumn:20,positionLineNumber:7,selectionStartColumn:20,selectionStartLineNumber:7,startColumn:20,startLineNumber:7),source:'%23include+%3CQByteArray%3E%0A%0AQByteArray+someFunk(bool+cond)+%7B%0A++++if+(cond)+%7B%0A++++++++return+QByteArray()%3B%0A++++%7D%0A++++return+nullptr%3B%0A%7D'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g142,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:qt,ver:'680')),options:'-fPIE+-O2',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+14.2+(Editor+%231)',t:'0')),k:50,l:'4',m:100,n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) where the `return QByteArray()` is fully inlined while the `return nullptr` is not.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
Swiftb0y: At class scope, usually right below the `enum class` declaration. See [this example](https://compiler-explorer.com/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:102,endLineNumber:10,positionColumn:102,positionLineNumber:10,selectionStartColumn:102,selectionStartLineNumber:10,startColumn:102,startLineNumber:10),source:'struct+A+%7B%0A++++enum+class+Enum+%7B%0A++++++++V1,+V2%0A++++%7D%3B%0A++++using+enum+Enum%3B%0A%7D%3B%0A%0Aint+main()+%7B%0A++++A::Enum+v+%3D+A::V1%3B+//+would+have+to+be+A::Enum::V1+without+the+%60using%60+above.%0A++++return+static_cast%3Cint%3E(v)%3B+//+we+still+retain+the+%22no+implicit+conversions+from/to+int%22+feature.%0A%7D'),l:'5',n:'1',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g142,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'--std%3Dc%2B%2B20',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+14.2+(Editor+%231)',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4). I'm unsure however how well this works with `Q_ENUM`. You'll need to experiment a little whether it can pick up on the `using enum` feature.

src/test/controllerscriptenginelegacy_test.cpp
Swiftb0y: well in this case there should be because because were using only characters which are the same in ASCII and LATIN_9, right? For others you could manually create a string with the correct bytes `\x40\x41`. See https://en.cppreference.com/w/cpp/language/escape

Swiftb0y: I'm not sure if that works, but it be nice if you could experiment with it a little.

Swiftb0y:
> I ended up writing an implementation using QJSValue and dumping the data in a JS array. WDYT?

I'm not sure. IMO the same arguments against wrapping it in a `UInt8Array` apply here as well. If you want an array instead of an `ArrayBuffer`, you can always do `Array.from(new UInt8Array(engine.convertChars(...)))`. The conversion to the array also again assumes you want to view the buffer in 8-bit chunks, only with an array its even more annoying to reinterpret that as 16-bit (or others). I really still think `ArrayBuffer` is the simplest, most elegant and efficient solution.

christophehenry:
It's not that I want an array over `ArrayBuffer` , it's just that I really think the function should return `undefined` over empty buffer on encoding error. However if you tell me it's fine to return an empty buffer as long as it's documented, then I'm fine with it.

Swiftb0y:
> However if you tell me it's fine to return an empty buffer as long as it's documented, then I'm fine with it.

I think its fine, but we can also work together on the `undefined` return value. The problem is that I never got to see what code you came up with there so I was never able to make an educated guess why it was crashing. Can you share that code again so I can take a look at it?

Swiftb0y:
One thing to not though is that `new TypedArray(undefined)` still results in an empty array TypedArray, so for the most common usecase, it doesn't make a difference whether its returning an empty `ArrayBuffer` or `undefined`.

christophehenry:
Ok, I reverted my changes and applied your suggestions. Now let's look at these failing build on macOS…

Swiftb0y:
I think you're simply missing the `#include <QStringEncoder>`

christophehenry:
What wait? Why don't I need it here, locally?

Swiftb0y:
Potentially slightly different includes depending on the platform and/or you're using a different Qt version, so the `#if` directives evaluate differently (that is usually one of the reason why its desirable to avoid them if possible).

christophehenry:
Sapristi.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1052,44 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+        return QByteArray();
+    }
+}
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
+        const QString& targetCharset, const QString& value) {
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+    auto* pCodec = QTextCodec::codecForName(targetCharset.toUtf8());
+    if (!pCodec) {
+        return QByteArray();
```
Swiftb0y, line 1080: I think we also want logoutput here and/or throw an error. wdyt?

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -745,3 +745,34 @@ TEST_F(ControllerScriptEngineLegacyTest, screenWillSentRawDataIfConfigured) {
...
 #endif
+
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetUndefinedOnUnknownCharset) {
+    const auto result = evaluate("engine.convertCharset('NULL', 'Hello!')");
+    EXPECT_EQ(qjsvalue_cast<QByteArray>(result), QByteArray());
+}
+
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetCorrectValueWellKnown) {
+    const auto result = evaluate(
+            "`${engine.convertCharset(engine.WellKnownCharsets.LATIN_9, 'Hello!')}`");
```
Swiftb0y, line 756: I think the implicit `toString` cast is wrong I think (and likely also the cause for the failing test).
```suggestion
    const auto result = evaluate(
            "engine.convertCharset(engine.WellKnownCharsets.LATIN_9, 'Hello!'));
```

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -745,3 +745,34 @@ TEST_F(ControllerScriptEngineLegacyTest, screenWillSentRawDataIfConfigured) {
     ASSERT_ALL_EXPECTED_MSG();
 }
 #endif
+
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetUndefinedOnUnknownCharset) {
+    const auto result = evaluate("engine.convertCharset('NULL', 'Hello!')");
+    EXPECT_EQ(qjsvalue_cast<QByteArray>(result), QByteArray());
```
Swiftb0y, line 751: of course `qjsvalue_cast`, very elegant! :+1:

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
```diff
@@ -17,6 +17,16 @@ class ConfigKey;
 class ControllerScriptInterfaceLegacy : public QObject {
     Q_OBJECT
   public:
+    enum class WellKnownCharsets {
+        LATIN_1,
+        ISO_8859_1,
+        LATIN_9,
+        ISO_8859_15,
+        UCS_2,
+        ISO_10646_UCS_2
```
Swiftb0y, line 26: for `enum class` I prefer `CamelCase` (also looks nicer in JS). I don't know how to express the ISO names well, so I guess the current naming is fine.
```suggestion
        Latin1,
        ISO_8859_1,
        Latin9,
        ISO_8859_15,
        UCS2,
        ISO_10646_UCS_2
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1052,44 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
     m_ramp[deck] = true;
 }
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
+        const ControllerScriptInterfaceLegacy::WellKnownCharsets targetCharset,
+        const QString& value) {
+    switch (targetCharset) {
+    case WellKnownCharsets::LATIN_1:
+    case WellKnownCharsets::ISO_8859_1:
+        return this->convertCharset(QStringLiteral("ISO-8859-1"), value);
```
Swiftb0y, line 1062: this is not common in C++ and some devs even frown upon it.
```suggestion
        return convertCharset(QStringLiteral("ISO-8859-1"), value);
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
christophehenry: Like:

```c++
VERIFY_OR_DEBUG_ASSERT(…) {

}
jsEngine()->throwError()
```
?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
christophehenry: Ah the habit of Python's `self` 😌

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1052,45 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+    case WellKnownCharsets::UCS2:
+    case WellKnownCharsets::ISO_10646_UCS_2:
+        return convertCharset(QStringLiteral("ISO-10646-UCS-2"), value);
+    default:
+        m_pScriptEngineLegacy->throwJSError(QStringLiteral("Unknown charset specified"));
+        return QByteArray();
+    }
+}
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
```
christophehenry, line 1075: Hmm… Now I'm thinking… Should it try to remove diacritics for charsets like Latin-1 that don't support accents? Should it be an option?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: How exactly does it fail currently? I'd need to see some code to make a proper judgement.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: Is that an option in `QStringEncoder`? Couldn't spot any feature there. It may make sense to pass `QStringEncoder::Flag::ConvertInvalidToNull` along though.

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -745,3 +745,34 @@ TEST_F(ControllerScriptEngineLegacyTest, screenWillSentRawDataIfConfigured) {
...
+    // aka ISO-8859-15 encoding for 'Hello!'
+    const char expected[] = {72, 101, 108, 108, 111, 33};
+    EXPECT_EQ(qjsvalue_cast<QByteArray>(result), QByteArray(expected, 6));
+}
+
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetCorrectValueStringCharset) {
+    const auto result = evaluate("`${engine.convertCharset('ISO-8859-15', 'Hello!')}`");
+    // int repr for 0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x21
+    // aka ISO-8859-15 encoding for 'Hello!'
+    const char expected[] = {72, 101, 108, 108, 111, 33};
```
Swiftb0y, line 768: just put the hex in there if you prefer ;)
```suggestion
    ISO-8859-15 encoded 'Hello!'
    const char expected[] = {0x48, 0x65, 0x6c, 0x6c, 0x6f, 0x21};
```

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -745,3 +745,34 @@ TEST_F(ControllerScriptEngineLegacyTest, screenWillSentRawDataIfConfigured) {
...
+    EXPECT_EQ(qjsvalue_cast<QByteArray>(result), QByteArray(expected, 6));
+}
+
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetUnsupportedChars) {
+    const auto result = evaluate("`${engine.convertCharset('ISO-8859-15', 'مايأ نامز')}`");
+    // int repr for 0x1A which is ISO-8859-15 for
+    // https://en.wikipedia.org/wiki/Substitute_character
+    const char expected[] = {26, 26, 26, 26, 32, 26, 26, 26, 26};
+    EXPECT_EQ(qjsvalue_cast<QByteArray>(result), QByteArray(expected, 9));
```
Swiftb0y, line 777: hardcoding the length is ugly (and error prone), lets change that. Add a utility (feel free to change the name):
```cpp
template <int N>
QByteArray intByteArray(const char(&array)[N]) {
    return QByteArray(array, N);
}
```
```suggestion
    const QByteArray expected = intByteArray({26, 26, 26, 26, 32, 26, 26, 26, 26});
    EXPECT_EQ(qjsvalue_cast<QByteArray>(result), expected);
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: well, in either way we would need to know exactly how the codec works so we can exploit the conversion process correctly. If the string from the user already contains the replacement character then the post-processing would alter that. You also need to know the byte-layout of the encoding to know how many bytes you would need to alter (8 vs 16 vs 32-bit chars depending on which UTF encoding you use). Pre-processing wouldn't even work, at least I can't think of a solution there. Lets KISS for now.

src/test/controllerscriptenginelegacy_test.cpp
Swiftb0y: I think it deduces that `{26, 26, 26, 26, 32, 26, 26, 26, 26}` is list-initialization of _some_ object type and then looks at the parameter, since `const char(&array)[size]` is an array literal, it can deduce that the elements within need to be `char`s and since its a list it can deduce `size` as well (since the array has not yet [decayed](https://en.cppreference.com/w/cpp/types/decay) into a pointer). How it does it internally? don't know either, black magic. Type deduction just works 95% of the time ;) (See the beginning of Scott Mayers "Effective Modern C++" for a good explanation).

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: yeah, thats why `logMessagesExpected` is quite a hack...

src/test/controllerscriptenginelegacy_test.cpp
christophehenry: I didn't know that C++ could do type conversion that powerful :thinking:

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -745,3 +745,43 @@ TEST_F(ControllerScriptEngineLegacyTest, screenWillSentRawDataIfConfigured) {
     ASSERT_ALL_EXPECTED_MSG();
 }
 #endif
+
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetUndefinedOnUnknownCharset) {
+    logMessagesExpected.append(std::tuple<QtMsgType, QRegularExpression>(
+            QtMsgType::QtWarningMsg, QStringLiteral("Unable to open encoder")));
```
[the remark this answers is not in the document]
christophehenry: Using something like `try { } catch { }` is the JS code?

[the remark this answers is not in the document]
Swiftb0y: Rather just seeing if instead of a QByteArray the QJSValue returns an error using `QJSValue::isError`. Haven't tried that yet, but I'm fairly certain it should work.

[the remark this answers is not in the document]
christophehenry: Ok, so as it turns out, [`SETUP_LOG_CAPTURE`](https://github.com/mixxxdj/mixxx/blob/main/src/test/controllerscriptenginelegacy_test.cpp#L674C5-L674C22) was responsible. What hinted me was that [other earlier tests](https://github.com/mixxxdj/mixxx/blob/main/src/test/controllerscriptenginelegacy_test.cpp#L122-L126), also calling `logOrThrowError` weren't failing. Since there doesn't seem to be a `qUninstallMessageHandler`, I just moved the tests before the first `SETUP_LOG_CAPTURE`.

Tests are passing locally so I suppose this PR is ready for another review.

christophehenry:
Linux tests failing. Conversion of unsupported characters don't produce the same result on Ubuntu than it does on Fedora or macOS? Well that's unfortunate. How should I solve that? I can rewrite the test to replace substitute characters with *?* so that we get a consistant comparison. How about that?

res/controllers/engine-api.d.ts
```diff
@@ -301,4 +301,57 @@ declare namespace engine {
...
+     * charset to either the codepoint for '?' or the substitute character.
+     * @see https://en.wikipedia.org/wiki/Substitute_character
+     *
+     * Example usage:
+     *
+     *         const group = "…"  // device midi group
+     *         const titleLength = engine.getValue(group, "track_title_length")
+     *         let title = ""
+     *         for (let i = 0; i < titleLength; i++) {
+     *             title += String.fromCodePoint(engine.getValue(group,`track_title_${i}`))
```
christophehenry, line 330: I'm not sure `String.fromCodePoint` exists in `QJSEngine`. [MDN states that it is widely available](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/String/fromCodePoint) my IDE, however, didn't know of it.

res/controllers/engine-api.d.ts
```diff
@@ -301,4 +301,57 @@ declare namespace engine {
...
+     * Available charset names are listed here: http://www.iana.org/assignments/character-sets/character-sets.xhtml
+     *
+     * Depending on the system you run on, this function may convert characters that are unsupported by the target
+     * charset to either the codepoint for '?' or the substitute character.
+     * @see https://en.wikipedia.org/wiki/Substitute_character
+     *
+     * Example usage:
+     *
+     *         const group = "…"  // device midi group
+     *         const titleLength = engine.getValue(group, "track_title_length")
```
christophehenry, line 327: Unless I'm wrong, `track_title_length` and `track_title_x` are undocumented. Should I document them here?

res/controllers/engine-api.d.ts
```diff
@@ -301,4 +301,57 @@ declare namespace engine {
...
+     *         const converted = new Uint8Array(engine.convertCharset(engine.WellKnownCharsets.Latin9, title))
+     *         for (const char of converted) {
+     *            // 0x1A is Latin 9 code point for substitute character,
+     *            // 0x3F is Latin 9 code point for '?'
+     *             if(char === 0x1A  ){
+     *                 controller.display(line, col, 0x3F)
+     *             } else {
+     *                 controller.display(line, col, char)
+     *             }
+     *         }
```
christophehenry, line 343: Maybe this gist should belong into the Wiki, idk.

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -659,6 +659,52 @@ TEST_F(ControllerScriptEngineLegacyTest, connectionExecutesWithCorrectThisObject
...
+            evaluate("engine.convertCharset('ISO-8859-15', 'مايأ نامز')"));
+
+    // On some plateforms, conversion replaces unknown chars by '?' rather
+    // than the substitute character (https://en.wikipedia.org/wiki/Substitute_character)
+    // Both are valid results but we need a consistent one for the test.
+    for (int i = 0; i < result.length(); ++i) {
+        if (result[i] == 0x1A) {
+            result[i] = 0x3F;
+        }
+    }
```
christophehenry, line 701: Or maybe I should just use the [`ConvertInvalidToNull`](https://doc.qt.io/qt-6/qstringconverter.html#Flag-enum) flag?

JoergAtGithub: If this is just relevant for <Qt6.4 than use an ifdef in the testcase as well.

christophehenry: No need. `QTextCodecè [supports the `ConvertInvalidToNull` flag](https://doc.qt.io/qt-6/qtextcodec.html#makeEncoder).

christophehenry:
@Swiftb0y @JoergAtGithub @daschuer Can you resolve the conversations that are resolved? It's getting hard for me to follow what comments are still relevant and what are outdated.

res/controllers/engine-api.d.ts
Swiftb0y: Are those even merged? Can you link me to the PR? I highly oppose those COs...

christophehenry:
Looks good for a merge to me.

JoergAtGithub (changes requested):

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
```diff
@@ -72,6 +82,13 @@ class ControllerScriptInterfaceLegacy : public QObject {
             const double rate = -10.0);
     Q_INVOKABLE void softStart(const int deck, bool activate, double factor = 1.0);
 
+    Q_INVOKABLE QByteArray convertCharset(
+            const ControllerScriptInterfaceLegacy::WellKnownCharsets
+                    targetCharset,
+            const QString& value);
+
+    Q_INVOKABLE QByteArray convertCharset(const QString& targetCharset, const QString& value);
```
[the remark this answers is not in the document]
christophehenry: Yes I understand your proposition. I just find it really, really inconvenient. Let's suppose someone wants to add a controller for a device that understands an exotic charset. Let's say GB2312. Then what they have to do is open a PR, wait for a review, merge and new release to be able to use this feature. Then let's say this charset is unsupported on RasPi4. Then we declare this feature unusable for all other platforms? I don't get why. With the current implementation, we have a predictable behavior and a sane warning message. I really don't get why this isn't sufficient.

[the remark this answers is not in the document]
JoergAtGithub: That's not what I suggested - I suggested to extend the list of charsets in this PR. Than the users have not only the code safety, but also the mapping developer gets developer experience benefit of code completion due to the known symbol.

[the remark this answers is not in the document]
christophehenry: There's a lot of charsets. Is there a way to autogenerate the enum et the corresponding `.d.ts`?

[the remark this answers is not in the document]
Swiftb0y: That suggestion conflicts with your previous requirement that the available codecs need to be consistent across the different operating systems though IIUC. The feedback from `QTextCodec::availableCodecs()` depends on runtime conditions, so may even vary across the same operating system.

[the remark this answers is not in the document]
JoergAtGithub: As long as Qt is built with ICQ enabled it should be always the same, at least until ICQ adds support for a new charset.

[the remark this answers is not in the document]
Swiftb0y: Agree, that is definitely an aspect we should discuss as part of the requirements.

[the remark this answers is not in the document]
christophehenry: So @JoergAtGithub, you want to generate the code yourself, as we discussed? Or else, here's what I propose: we leave the code as is but any officially supported controller must use the enum. That way, we ensure official controllers ar working on all platforms but this leaves freedom of implementation for community controllers. I can add a comment to the code specifying that. Is that solution acceptable?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1052,47 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+    if (!pCodec) {
+        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
+        return QByteArray();
+    }
+    return pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull)->fromUnicode(value);
+#else
+#if QT_VERSION > QT_VERSION_CHECK(6, 8, 0)
+    const auto* const encoderName = targetCharset.data();
+#else
+    auto* const encoderName = targetCharset.toUtf8().data();
```
JoergAtGithub, line 1088:
```suggestion
    QByteArray encoderNameArray = targetCharset.toUtf8();
    const char* encoderName = encoderNameArray.constData();
```
This line worked for me only in normal operation, but not if I run it in the debugger. In the debugger .toUtf8() returned only garbage. I think the lifetime of the array wasn't guaranteeed here. After introducing the deep copy into a local array it works also in the debugger.
I only tested this branch (Windows 11 with Qt6.5) but the others look similar.

JoergAtGithub (changes requested):

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
christophehenry: Won't line 1086 have the same problem?

JoergAtGithub: Yes, that is what I meant with the others look similar.

christophehenry: K. Met pe fix that.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1052,47 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+#endif
+    QStringEncoder fromUtf16 = QStringEncoder(
+            encoderName, QStringEncoder::Flag::ConvertInvalidToNull);
```
christophehenry, line 1091:
```suggestion
    QStringEncoder fromUtf16 = QStringEncoder(
            QAnyStringView(targetCharset), QStringEncoder::Flag::ConvertInvalidToNull);
    if (!fromUtf16.isValid()) {
        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
        return QByteArray();
    }
    return fromUtf16(value);
```

@JoergAtGithub I'm not sure what I'm doing here, but would that code work for every Qt>6?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
christophehenry: Wait… Why was it useful to remove `.toUtf8()` for Qt>=6.8.0 again? Can't I just do:

```c++
QStringEncoder(targetCharset.targetCharset.toUtf8().data()
```

Swiftb0y: > Wait… Why was it useful to remove .toUtf8() for Qt>=6.8.0 again? Can't I just do:

Because it does a utf16 to utf8 conversion along with a heap allocation at runtime while on Qt6.8 [QStringEncoder can take a `QAnyStringView`](https://doc.qt.io/qt-6/qstringencoder.html#QStringEncoder-1) which eliminates that. So the removal of the `toUtf8` call was an optimization opportunity thats only possible on new enough Qt versions.

Swiftb0y: Regardless, Jörg is correct. `.toUtf8().data()` creates a dangling reference so we need to store that (unnecessary) copy separately.

christophehenry: Ok, I wrote a slightly different version in my last commit. Is it OK ?

JoergAtGithub: The last commit does not address the issue. You need a deep copy for this string reference - I'm not sure if this is needed for the StringView in Qt6.8 as well, but I guess so.


``` C++
QByteArray ControllerScriptInterfaceLegacy::convertCharset(
        const QString& targetCharset, const QString& value) {
    QByteArray encoderNameArray = targetCharset.toUtf8();
#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
    auto* pCodec = QTextCodec::codecForName(encoderNameArray);
    if (!pCodec) {
        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
        return QByteArray();
    }
    return pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull)->fromUnicode(value);
#else
#if QT_VERSION > QT_VERSION_CHECK(6, 8, 0)
    QAnyStringView encoderName = QAnyStringView::fromArray(encoderNameArray);
#else
    const char* encoderName = encoderNameArray.constData();
#endif
    QStringEncoder fromUtf16 = QStringEncoder(
            encoderName, QStringEncoder::Flag::ConvertInvalidToNull);
    if (!fromUtf16.isValid()) {
        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
        return QByteArray();
    }
    return fromUtf16(value);
#endif
}
```

I only tested this with Qt6.5

christophehenry: Hmm… Ok I won't pretend I understand what you are both saying here. I amended my last commit with @JoergAtGithub's version. I tested running the tests in debug. They are passing here for Qt6.8. I just changed `QT_VERSION > QT_VERSION_CHECK(6, 8, 0)` to `QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)`.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1052,48 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
+        const QString& targetCharset, const QString& value) {
+    QByteArray encoderNameArray = targetCharset.toUtf8();
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+    auto* pCodec = QTextCodec::codecForName(encoderNameArray);
+    if (!pCodec) {
+        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unable to open encoder"));
+        return QByteArray();
+    }
+    return pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull)->fromUnicode(value);
```
Swiftb0y, line 1084: This leaks. 
> Creates a [QTextEncoder](https://doc.qt.io/qt-6/qtextencoder.html) with a specified flags to encode chunks of Unicode data as char * data.
>
> The caller is responsible for deleting the returned object.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1052,48 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+#else
+#if QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)
+    QAnyStringView encoderName = QAnyStringView(encoderNameArray);
+#else
+    const char* encoderName = encoderNameArray.constData();
+#endif
```
Swiftb0y, line 1090: This defeats the purpose of `QAnyStringView`. `QAnyStringView` is polymorphic over the the string encoding. You don't need to convert to another format.
```suggestion
#if QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)
    QAnyStringView encoderName = QAnyStringView(targetCharset);
#else
    QByteArray encoderNameArray = targetCharset.toUtf8();
    const char* encoderName = encoderNameArray.constData();
#endif
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1047,3 +1052,48 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unknown charset specified"));
+        return QByteArray();
+    }
+}
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharset(
+        const QString& targetCharset, const QString& value) {
+    QByteArray encoderNameArray = targetCharset.toUtf8();
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+    auto* pCodec = QTextCodec::codecForName(encoderNameArray);
```
Swiftb0y, line 1079: continuation from previous comment
```suggestion
#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
    QByteArray encoderNameArray = targetCharset.toUtf8();
    auto* pCodec = QTextCodec::codecForName(encoderNameArray);
```

Swiftb0y (changes requested):
addressing @JoergAtGithub now since this is his code.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: > I'm not sure if this is needed for the StringView in Qt6.8 as well, but I guess so.

Its not, which is the primary difference between `QLatin1String` and `QAnyStringView` and why the `QAnyStringView` is more efficient in this case. Its essentially a `variant<string_view, u8string_view, u16string_view>` so you can efficiently construct it from most common string formats.

christophehenry:
@Swiftb0y @JoergAtGithub I'm really sorry, the ongoing discussion is way above my skills. So just tell me what to do when you agree on a correct implementation.

Swiftb0y:
No worries. I'll create a PR to your branch.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: I think the easiest solution here would be
```suggestion
    return std::unique_ptr(pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull))->fromUnicode(value);
```

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -658,6 +660,99 @@ TEST_F(ControllerScriptEngineLegacyTest, connectionExecutesWithCorrectThisObject
...
+    if (!failedCharsetsResult.isEmpty()) {
+        ADD_FAILURE() << "Charsets with result length not equal to expected:\n"
+                      << failedCharsetsResult.toStdString();
+    }
+
+    // ControlObjectScript connections are processed via QueuedConnection. Use
+    // processEvents() to cause Qt to deliver them.
+    processEvents();
+    EXPECT_DOUBLE_EQ(counter->get(), charsetEnumEntry.keyCount());
+}
```
Swiftb0y, line 736: this would be more robust if instead of relying on an out-of-band CO, we would simply evaluate the code and use the expression result. At least this current approach seems very convoluted to me.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1052,3 +1058,288 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+        m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unknown charset specified"));
+        return QByteArray();
+    }
+}
+
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+// Add Byte Order Mark (BOM) to the beginning of the data if it is not already present
+static QByteArray addBom(const QByteArray& data) {
+    static const QByteArray bomBE = QByteArray::fromHex("FEFF");
+    static const QByteArray bomLE = QByteArray::fromHex("FFFE");
```
Swiftb0y, line 1295: constexpr all the things
```suggestion
    static const QByteArray bomBE = QByteArrayLiteral("\xFE\xFF");
    static const QByteArray bomLE = QByteArrayLiteral("\xFF\xFE");
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1052,3 +1058,288 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+        m_pScriptEngineLegacy->logOrThrowError(
+                QStringLiteral("Unable to get QTextCodec name for charset: %1").arg(targetCharset));
+        return QByteArray();
+    }
+    std::unique_ptr<QTextEncoder> encoder(
+            pCodec->makeEncoder(QTextCodec::Flag::ConvertInvalidToNull));
+    if (encoderNameArray == "ISO-10646-UCS-2" || encoderNameArray == "UCS2") {
+        // For these encodings QTextCodec does not set, the BOM which QStringEncoder does
+        return addBom(encoder->fromUnicode(value));
+    }
```
Swiftb0y, line 1326: it would be nice to expose `QStringConverter::Flag::WriteBom` as in I found in practice that the BOM is rarely used in controller protocols.

JoergAtGithub: Here it's very relevant, as we send strings encoded in multi-byte charsets to a controller with endianess known at development time of the mapping script, but from a host computer with endianess unknown at time of mapping development.

Swiftb0y: So the endianes of a string encoded in UCS2 always matches that of the host?

JoergAtGithub: I think so, but don't know for sure. The same applies for most other charsets, but for all the other multi-byte charsets there was no difference between the output of `QTextCodec` and `QStringEncoder`.
On the other hand, the number of users running Mixxx on a BigEndian CPU architecture is presumably negligible.

Swiftb0y: likely yeah, but it would be nice to confirm whats the case. The mapping creator needs to be certain what endianness their returned string is...

Swiftb0y (changes requested):
I agree. This as sprawled into an overengineered mess I'm not comfortable merging. If we want something like this, we need to go back to the drawing board.

JoergAtGithub (approved):

christophehenry:
Closing in favor of #13935

PR #13774, opened 2024-10-17
feat(cmake): Compile translations at build-time

Holzhaus:
Compilation actually doesn't take very long (just a few seconds). This allows removing the binary `*.qm` files from the git source tree and thus reduce of both the size of diffs and the transferred git object deltas on translation changes.

**NOTE:** This currently fails because macOS 12 arm64 buildenv does not contain `lrelease`.


CMakeLists.txt
```diff
@@ -2828,6 +2817,64 @@ elseif(WIN32)
...
 
+# Mixxx translation files
+file(GLOB MIXXX_TS_FILES "${CMAKE_CURRENT_SOURCE_DIR}/res/translations/*.ts")
+message(VERBOSE "Mixxx translation files: ${MIXXX_TS_FILES}")
+
+if(QT6)
+  find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS LinguistTools)
+  set_source_files_properties(
+    ${MIXXX_TS_FILES}
+    PROPERTIES OUTPUT_LOCATION "${CMAKE_CURRENT_SOURCE_DIR}/res/translations"
```
JoergAtGithub, line 2828: Normally, if we compile a source file, we store the output in `${CMAKE_CURRENT_BINARY_DIR}`, Why storing the .qm files in the source directory instead?

tools/debian_buildenv.sh
```diff
@@ -100,6 +100,7 @@ case "$1" in
             qt5keychain-dev \
             qtdeclarative5-dev \
             qtscript5-dev \
+            qttools5-dev \
```
JoergAtGithub, line 103: I think we shouldn't target this to 2.4.

CMakeLists.txt
Holzhaus: > Why storing the .qm files in the source directory instead?

Because otherwise I guess it won't be picked up when compiling Mixxx locally and then running it directly from the build dir. Personally I don't really care (using `LANG=en_US.UTF8` anyway), but I think makes sense to be able to test localization without installing Mixxx.

tools/debian_buildenv.sh
Holzhaus: 2.5? main? IDC, just need to know what to target.

JoergAtGithub: I recommend main, as 2.5 is already close to release.

CMakeLists.txt
daschuer: We have already a hack to find the resource files from a Mixxx source tree.

https://github.com/mixxxdj/mixxx/blob/a716202b0b5bd6923d4d4438d62dba5060a19791/src/preferences/configobject.cpp#L46

We may handele the translations separate to load them from the build tree.

Holzhaus: Sure, sounds like a good follow up to this PR \*wink-wink\*.

tools/debian_buildenv.sh
Holzhaus: rebased.

Holzhaus:
I rebased again because the PR checks didn't run for some reason.

acolombier:
Could you resync your branch? Wondering if the Mac13 runner will fix the following:
> NOTE: This currently fails because macOS 12 arm64 buildenv does not contain lrelease.

Holzhaus:
> Could you resync your branch? Wondering if the Mac13 runner will fix the following:
> > NOTE: This currently fails because macOS 12 arm64 buildenv does not contain lrelease.

Sure, will do tomorrow.

JoergAtGithub:
> 
> Sure, will do tomorrow.

@Holzhaus Gentle reminder that this was 90days ago

PR #13781, opened 2024-10-19
Update Linux-GitHub runner to Ubuntu 24.04.01 LTS

JoergAtGithub:
According our https://github.com/mixxxdj/mixxx/wiki/Minimum-requirements-policy Ubuntu 24.04.01 LTS is the earliest version, that developers need to support in new code.
I'm working on some code that fullfills this requirement, but does not build on CI, because the version of the Linux GitHub runner is lagging behind the minimum requirement. This update ensures that the CI will still pass.

I found out, that clazy version 1.11, that Ubuntu ships with 24.04.01 LTS is incompatible with the included Qt 6.4. Therfore we can't use the clazy package from the Ubuntu distribution. I added the steps to download and build it as described on the clazy homepage.


acolombier:
Overall, I'm not sure this is a good idea to drop support for Ubuntu 22.04/Qt 6.2 LTS. A few reasons for that:
- Ubuntu 24.04 doesn't rely on a LTS version of Qt
- PopOS (which I appear to be using for full disclosure) hasn't released a 24.04 LTS due to some delay on their end
Appreciate that last point is strongly biased, but as PopOS is still fairly popular among the Linux community, so I think it would be wise to support at least, at least till we provide support for the next Qt LTS version on Linux.
Another point is, I know this is the established strategy in our wiki, but is it wise to drop support for Ubuntu LTS after only 2 years? I understand we may not want to follow precisely Canonical's strategy, but many of our Linux community relies on Ubuntu and derivatives and usually pick LTS version to ensure they won't have to break and reinstall their system every year or 2.

TL;DR is, could we support both 22.04 and 24.04, at least till we move to support the next Qt LTS version?

JoergAtGithub:
>     * Ubuntu 24.04 doesn't rely on a LTS version of Qt

That make no difference, because 6.2 LTS and 6.4 are both out of maintenance. 6.5 is the oldes Qt6 release where bugfixes will be developed for: https://endoflife.date/qt

> but is it wise to drop support for Ubuntu LTS after only 2 years? 

We decided to introduce these rules, to not hinder development. Ubuntu has already very old packages, at the time of release. So we have a delay of more than 2 years until we can use upstream features.

acolombier:
> could we support both 22.04 and 24.04, at least till we move to support the next Qt LTS version?

How about this?

JoergAtGithub:
> > could we support both 22.04 and 24.04, at least till we move to support the next Qt LTS version?
> 
> How about this?

This would prevent us to use new functionality.

We introduced this rule based system to give all devs predictability, and not to depend on case by case decisions with long discussions.

daschuer:
We have established our [Minimum-requirements-policy](https://github.com/mixxxdj/mixxx/wiki/Minimum-requirements-policy) after exchanging quite some arguments. One major argument for me was, the desire of contributors to gain experience with bleeding etch features soon after they are available. This is "feel good" issue is important IMHO. So there is no reason to repeat this discussion. 

It does on the other hand not stop us from maintaining a back port branch if there is a need for it. I am for example still at Ubuntu Focal with one of my devices to test the 2.4. branch (and urgently need to update it btw., we are beyond the deadline) and maintain such a branch: https://github.com/daschuer/mixxx/tree/focal_main. 

I will do a similar branch for Ubuntu Jammy as well, that you may use as well. We may consider to make it more accessible and move it to "mixxxdj/mixxx". This is controversial of cause.

.github/workflows/build-checks.yml
```diff
@@ -81,6 +97,7 @@ jobs:
             -DMAD=ON \
             -DMODPLUG=ON \
             -DWAVPACK=ON \
+            -DCMAKE_CXX_FLAGS="-Wno-nan-infinity-disabled" \
```
daschuer, line 100: We should use our fpclassify.cpp instead.

.github/workflows/build-checks.yml
```diff
@@ -21,13 +21,27 @@ jobs:
...
+    runs-on: ubuntu-24.04
     name: ${{ matrix.name }}
     steps:
       - name: Check out repository
         uses: actions/checkout@v4.2.1
       - name: Install build dependencies
         run: tools/debian_buildenv.sh setup
+      - name: Build clazy v1.12
+        if: matrix.name == 'clazy'
+        # Ubuntu2404 comes with v1.11, which is too old for the Qt6.4 shipped with Ubuntu 2404
```
daschuer, line 33: Do we have a reference for this, a Mixxx and a Ubuntu bug? This sounds like an issue we should solve in our https://github.com/daschuer/mixxx/blob/focal_main/tools/debian_buildenv.sh instead.

.github/workflows/build-checks.yml
```diff
@@ -116,19 +133,12 @@ jobs:
...
       - name: Build
         # Do not abort on errors and build/check the whole project
         run: cmake --build . -j $(nproc) -- --keep-going
         working-directory: build
         env:
-          CLAZY_CHECKS: level2,container-inside-loop,heap-allocated-small-trivial-type,isempty-vs-count,qhash-with-char-pointer-key,qproperty-type-mismatch,tr-non-literal,no-rule-of-two-soft,no-non-pod-global-static,no-qproperty-without-notify,no-qstring-allocations,no-function-args-by-value,no-copyable-polymorphic,no-ctor-missing-parent-argument,no-missing-qobject-macro,no-rule-of-three,no-returning-void-expression
+          CLAZY_CHECKS: level2,container-inside-loop,heap-allocated-small-trivial-type,isempty-vs-count,qhash-with-char-pointer-key,qproperty-type-mismatch,tr-non-literal,no-rule-of-two-soft,no-non-pod-global-static,no-qproperty-without-notify,no-qstring-allocations,no-function-args-by-value,no-copyable-polymorphic,no-ctor-missing-parent-argument,no-missing-qobject-macro,no-rule-of-three,no-returning-void-expression,no-no-module-include
```
daschuer, line 141: Isn't this just a matter of alter include directives: 
https://github.com/KDE/clazy/blob/master/docs/checks/README-no-module-include.md
I would prefer to fix the issue instead of disable the check.

daschuer (changes requested):

.github/workflows/build-checks.yml
JoergAtGithub: I could move it to debian_buildenv.sh, but than clazy would be always build and not only for the clazy CI action.

daschuer: The goal should be to have a working clazy locally by default. I looked around to find a PPA for that but not yet found. Maybe we add this to our Launchpad PPA? What is the issue you trying to solve btw.?

JoergAtGithub: It can not compile the Qt6.4 headers, because of unsupported C++ features

.github/workflows/build-checks.yml
daschuer: I wonder if this is a clang issue as well. 

Is this the issue we are fighting: 

https://github.com/JoergAtGithub/mixxx/actions/runs/11297991113/job/31426009871#step:9:1008

JoergAtGithub: Yes, clazy is based on clang. And the old clang version, where clazy 1.11 is based on does not the C++ features used in the Qt6.4 headers. clazy 1.12 is build on base of a recent clang version and can compile this code.

daschuer: Does it build with clang 15? 

I think the GCC headers are not compatible.

https://askubuntu.com/questions/1525087/after-upgrading-ubuntu-to-24-04-lts-clang-does-not-work-properly

Did you try to use the clang libraries with 
`-stdlib=libc++"`

.github/workflows/build-checks.yml
JoergAtGithub: This is not a regression of this PR, as the check hasn't existed before. Therefore it's out of scope of this PR!

.github/workflows/build-checks.yml
JoergAtGithub: Clazy passes now with unmodified ruleset

.github/workflows/build-checks.yml
daschuer: This takes 5 min. A bit long IMHO. 
With a bit of luck the Ocular version will work: https://packages.ubuntu.com/oracular/clazy

.github/workflows/build-checks.yml
daschuer: Yes, but let's keep the warning until the issue is fixed in the other PR.

src/qml/qmlchainpresetmodel.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 #include <QAbstractListModel>
-#include <QtQml>
+#include <QQmlEngine>
```
daschuer, line 3: This is also out of scope of this PR.

.github/workflows/build-checks.yml
JoergAtGithub: But than all CI runs fail?

src/qml/qmlchainpresetmodel.h
JoergAtGithub: This is necessary, that Clazy passes, should I move it to an own PR?

.github/workflows/build-checks.yml
Holzhaus: Opened a new issue, so that we don't we forget about it: https://github.com/mixxxdj/mixxx/issues/13837

Problem solved?

.github/workflows/build-checks.yml
Holzhaus: Can't we just create a docker container with the new clazy? No ideal, but should work okay without delaying every CI build.

daschuer: Using the oracular version does drag a lot of clang dot dependencys in, because they are dependent by an exact version. I consider to push the build to our PPA and make it available for home builds as well.

src/qml/qmlchainpresetmodel.h
daschuer: It is not very important. I actually prefere to have everything fixed either in one go or before doing the switch. This as well as the nan-infinity-disabled issue. We need to do it anyway at one point, let's avoid intermediate workarounds.

.github/workflows/build-checks.yml
daschuer: In the meantime I have filed a bug: https://bugs.launchpad.net/ubuntu/+source/clazy/+bug/2086665

daschuer: using the Oracular build works 
``
          echo "deb http://archive.ubuntu.com/ubuntu oracular main universe" | sudo tee /etc/apt/sources.list.d/oracular-temp.list
          echo -e "Package: *\nPin: release n=oracular\nPin-Priority: 100" | sudo tee /etc/apt/preferences.d/oracular.pref
          sudo apt update
          sudo apt install -t oracular clazy
```
Penalty 46s

Comment (daschuer) [reply .github/workflows/build-checks.yml:33]: The Oracular hack updates a lot of dependency. Nothing I would like to do at home. 

The macOS solution sounds promising. It looks like macOS builds are in general faster so it is probably a win-win.

Comment (JoergAtGithub) [reply .github/workflows/build-checks.yml:33]: macOS is not the way to go, the Homebrew installation of Clazy took 1h 50m 50s because it builds the whole llvm package as dependency.

Comment (daschuer) [reply .github/workflows/build-checks.yml:33]: Oh, that's bad. So we have two remaining options. Push the clazy build to our PPA, or make use of the GitHub cache. I will look in the PPA solution. It should be straight forward using the Oracular rule file. This has the benefit that Ubuntu users can use it at home as well.

Comment (JoergAtGithub) [reply .github/workflows/build-checks.yml:33]: We've also the docker container solution proposed by Jan as third option. But I'm also in favor to trying the PPA solution first, as it covers not only CI.

**Comment by daschuer** [comment]
https://github.com/JoergAtGithub/mixxx/pull/2
Contains the new clazy and the infinity fix. 
It was not the upload no brainer I have first thought because of failing test with the older Qt version. But now everything is in shape. .. hopefully.

**Comment by JoergAtGithub** [comment]
This is ready for merge now:

- @daschuer set up the Clazy 1.12 PPA, which works great!
- The last remaining Clazy error is fixed in this PR
- The new default clang-tidy check for "use of infinity" with fast-math (see https://github.com/mixxxdj/mixxx/issues/13780 for details) remains disabled as before -> This is no regression!

File: .github/workflows/build-checks.yml [hunk 5]
```diff
@@ -81,6 +90,7 @@ jobs:
             -DMAD=ON \
             -DMODPLUG=ON \
             -DWAVPACK=ON \
+            -DCMAKE_CXX_FLAGS="-Wno-nan-infinity-disabled" \
```
Swiftb0y, line 93: could you add a comment at the issue locations so we don't forget to remove this once fixed?

daschuer (approved):
LGTM, Thank you.

daschuer:
Unfortunately we have a conflict now.

JoergAtGithub:
I fixed the merge-conflict and added the comments suggested by Niko.

.pre-commit-config.yaml
```diff
@@ -124,7 +124,7 @@ repos:
       - id: prettier
         types: [yaml]
   - repo: https://github.com/qarmin/qml_formatter.git
-    rev: 37c2513b1b8275a475a160ed2f5b044910335d5f # No release tag yet including #6 fix
+    rev: latest # No release tag yet including #6 fix
```
Swiftb0y, line 127: this should be reverted... pre-commit does that automatically unfortunately when running `autoupdate`.

JoergAtGithub: I opened PR https://github.com/mixxxdj/mixxx/pull/13880, which reverts this unintended change.

daschuer:
It turns out the clazy PPA upload is not required on a default Ubuntu 24.04. 
https://bugs.launchpad.net/ubuntu/+source/clazy/+bug/2086665
The issue happens only if gcc-14 is also installed. 
```
          sudo apt-get remove -y gcc-14
          sudo apt-get autoremove -y
```
https://github.com/daschuer/mixxx/commit/f0fddab26bf8bad764fcb1351d7ebf9c5c81d4e8

works as well.

PR #13782, opened 2024-10-19
BaseTrackTableModel: Fix column header mapping

daschuer:
This finishes the work of @uklotzde in https://github.com/mixxxdj/mixxx/pull/13682 to fix the column header mapping in case using additional columns in his fork, not used in Mixxx. 

I avoid a redundant initialization loop and in addition this improves the maintenance a lot by using a const kColumnPropertiesByEnum[] table to initialize thee header properties. This has the advantage that the compiler (tested with gcc) can check weather the table is complete. 
 
If an entry is missing or not in consecutive order we will get: 
```
src/library/columncache.cpp:168:92: sorry, unimplemented: non-trivial designated initializers not supported
  168 |         [ColumnCache::COLUMN_REKORDBOX_ANALYZE_PATH] = {REKORDBOX_ANALYZE_PATH, nullptr, 0}};
      |                                                                                            ^
src/library/columncache.cpp:169:50: error: static assertion failed
  169 | static_assert(std::size(kColumnPropertiesByEnum) == ColumnCache::NUM_COLUMNS);
      |               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Note: This requires a Transifex roundtrip to move the header translations to the new translation scope, to the new common place for all.


JoergAtGithub:
> Note: This requires a Transifex roundtrip to move the header translations to the new translation scope, to the new common place for all.

We are past string freeze for 2.5 - Shouldn't this go into main instead?

daschuer:
We may decide. I have no particular demand. It is at 2.5 because @uklotzde has picked that target. 
Since we just have moved strings, my hope is that Transifex fixes i automatically. If it fails a manual fix would be trivial.This should be done in main anyway to not loose any already translated strings.

src/library/baseexternaltrackmodel.cpp
```diff
@@ -35,21 +35,50 @@ BaseExternalTrackModel::BaseExternalTrackModel(QObject* parent,
...
-    QString album = index.sibling(index.row(), fieldIndex("album")).data().toString();
-    QString year = index.sibling(index.row(), fieldIndex("year")).data().toString();
-    QString genre = index.sibling(index.row(), fieldIndex("genre")).data().toString();
-    float bpm = index.sibling(index.row(), fieldIndex("bpm")).data().toString().toFloat();
-
-    QString nativeLocation = index.sibling(index.row(), fieldIndex("location")).data().toString();
+    QString artist =
+            index.sibling(index.row(),
+                         fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_ARTIST))
+                    .data()
```
Swiftb0y, line 48: this deserves a utility function IMO.

src/library/baseexternaltrackmodel.cpp
```diff
@@ -35,21 +35,50 @@ BaseExternalTrackModel::BaseExternalTrackModel(QObject* parent,
...
+            index.sibling(index.row(),
+                         fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_GENRE))
+                    .data()
+                    .toString();
+    float bpm =
+            index.sibling(index.row(),
+                         fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_BPM))
+                    .data()
+                    .toString()
+                    .toFloat();
```
Swiftb0y, line 75:
```suggestion
                    .toFloat();
```

src/library/columncache.cpp
```diff
@@ -13,6 +13,56 @@ const QString kSortNoCase = QStringLiteral("lower(%1)");
 const QString kSortNoCaseLex = mixxx::DbConnection::collateLexicographically(
         QStringLiteral("lower(%1)"));
 
+const QString kColumnNameByEnum[] = {
+        [ColumnCache::COLUMN_LIBRARYTABLE_ID] = LIBRARYTABLE_ID,
+        [ColumnCache::COLUMN_LIBRARYTABLE_ARTIST] = LIBRARYTABLE_ARTIST,
```
Swiftb0y: this is unfortunately not valid C++.

daschuer: For my understanding this is the C++ 20 feature "designated initializers".

daschuer: :-( 
```
Error: /home/runner/work/mixxx/mixxx/src/library/columncache.cpp:27:9: error: array designators are a C99 extension [-Werror,-Wc99-designator]
        [ColumnCache::COLUMN_LIBRARYTABLE_ID] = {LIBRARYTABLE_ID,
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
``

Comment (Swiftb0y) [reply src/library/columncache.cpp]: yup. Designated initializers are cool, but they deliberately work differently to allow for for example default arguments. I'm afraid there is no real way of achieving this trick in C++. You'll implicitly have to assume that the order of the list matches the enum order. The better strategy would instead simply be to not rely on the enum order at all though. Instead of having a bunch of globals arrays, consider having a single array of structs that bundles these together.

Comment (daschuer) [reply src/library/columncache.cpp]: > You'll implicitly have to assume that the order of the list matches the enum order. 

A appreciate exactly this behavior a feature. This makes sure the list is complete and consistent. 

I don't understand you proposal. 

I would like to treat this as a compiler feature we can use like `#pragma once`, and just silence clazy. 

What do you think?

File: src/library/columncache.cpp [hunk 4]
```diff
@@ -154,3 +144,8 @@ void ColumnCache::slotSetKeySortOrder(double notationValue) {
     // Replace the existing sort order
     m_columnSortByIndex[keyColumnIndex] = keySortSQL;
 }
+
+const QString& ColumnCache::columnName(Column column) const {
+    DEBUG_ASSERT(static_cast<std::size_t>(column) < std::size(kColumnNameByEnum));
+    return kColumnNameByEnum[column];
+}
```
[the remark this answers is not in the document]
daschuer: Creating a QString form a QStringView is a deep copy. 
kColumnNameByEnum uses LIBRARYTABLE_ID and friends a QStringLiteral which is basically a constexpr, but does not have the keyword because of MSVC limitations.
So using a QStringView here for the sake of becoming a constexpr, will introduce more deep copies later where now implicit sharing is used.

src/library/columncache.cpp
Swiftb0y: > I don't understand you proposal.

Well currently you have
```cpp
auto kColumnPropertiesByEnum[] = {}; 
auto kColumnPropertiesByEnum[] = {}; 

> I would like to treat this as a compiler feature we can use like #pragma once, and just silence clazy.

I'm not sure its worth the trouble trying to opt into this extension and maintain it on every compiler. I'm also not sure what the semantics are when the enum suddenly gets different values (are entries in the array that are now empty unitialized, zero or default initialized?) I don't think we should mix C99 extensions with C++ code. 

Have you instead considered putting this into a utility function with a switch-case? Performance is likely comparable and this time we don't have issues if the enum suddenly gets gaps and we can rely on `-Wswitch` for exhaustiveness (also we don't need `NUM_COLUMNS` anymore, which is always a hack).

Comment (daschuer) [reply src/library/columncache.cpp]: > I'm also not sure what the semantics are when the enum suddenly gets different values 

That is the great thing with this "feature". Compiling fails. So you cannot mess up the initalization you can do if this are single cals or an anonymous initalizer like in MSVC.

Comment (Swiftb0y) [reply src/library/columncache.cpp]: Ok. I won't stress this for now then. Lets focus on the other stuff first.

File: src/library/columncache.cpp (hunk 4 above)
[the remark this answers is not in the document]
Comment (daschuer) [reply src/library/columncache.cpp]: With a pointer it works.

File: src/library/basesqltablemodel.cpp [hunk 5]
```diff
@@ -633,13 +622,18 @@ int BaseSqlTableModel::fieldIndex(const QString& fieldName) const {
...
+            // Subtract one from the fieldIndex() because the id column is in both
             return m_tableColumns.size() + sourceTableIndex - 1;
         }
     }
     return tableIndex;
 }
 
+int BaseSqlTableModel::maxFieldIndex() const {
+    // Subtract one to remove the id column which is in both
+    return m_tableColumns.size() + (m_trackSource ? m_trackSource->maxFieldIndex() : 0);
```
uklotzde, line 634: That doesn't look correct to me.

The brittle _max index_ = _size minus one_ semantics are error prone.

`maxFieldIndex()` should better become `endFieldIndex()` which would be more consistent with how ranges are represented in C++.

daschuer:
Done

uklotzde:
No objections. I didn't review the additional changes in detail because they don't seem to affect my use case.

I'll update my fork after the changes have reached main.

Hint: This could have been achieved by merging smaller, independent PRs from different contributors. One at a time. Would cause less friction and frustration.

daschuer:
This is still waiting for review. Who likes to have a look?

src/library/basetracktablemodel.h
```diff
@@ -21,6 +30,11 @@ class BaseTrackTableModel : public QAbstractTableModel, public TrackModel {
             const char* settingsNamespace);
     ~BaseTrackTableModel() override = default;
 
+    QVariant getFieldVariant(const QModelIndex& index, ColumnCache::Column column) const;
+    QVariant getFieldVariant(const QModelIndex& index, const QString& fieldName) const;
+    QString getFieldString(const QModelIndex& index, ColumnCache::Column column) const;
+    QString getFieldString(const QModelIndex& index, const QString& fieldName) const;
+
```
Swiftb0y: nit: can the string-based overloads be deprecated or do they fulfill a need the ColumnCache ones can't?

src/library/basetracktablemodel.cpp
```diff
@@ -1008,11 +852,10 @@ QVariant BaseTrackTableModel::roleValue(
...
         const QModelIndex& index) const {
-    const auto bpmLockIndex =
-            index.sibling(
-                    index.row(),
-                    fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_BPM_LOCK));
-    return bpmLockIndex.data().toBool();
+    const bool bpmLockIndex =
+            getFieldVariant(index, ColumnCache::COLUMN_LIBRARYTABLE_BPM_LOCK)
+                    .toBool();
+    return bpmLockIndex;
```
Swiftb0y: Nit: variable name doesn't match anymore. can simply be removed.

```suggestion
    return getFieldVariant(index, ColumnCache::COLUMN_LIBRARYTABLE_BPM_LOCK)
                    .toBool();
```

src/library/columncache.cpp
```diff
@@ -13,93 +15,206 @@ const QString kSortNoCase = QStringLiteral("lower(%1)");
...
-    for (int i = 0; i < NUM_COLUMNS; ++i) {
-        m_columnIndexByEnum[i] = -1;
-    }
-    for (auto it = m_columnNameByEnum.constBegin(); it != m_columnNameByEnum.constEnd(); ++it) {
-        DEBUG_ASSERT(it.key() >= 0);
-        DEBUG_ASSERT(it.key() < NUM_COLUMNS);
-        DEBUG_ASSERT(m_columnIndexByEnum[it.key()] == -1);
-        m_columnIndexByEnum[it.key()] = fieldIndex(it.value());
+    static_assert(std::size(kColumnPropertiesByEnum) ==
+            sizeof(m_columnIndexByEnum) / sizeof(m_columnIndexByEnum[0]));
```
Swiftb0y, line 215:
```suggestion
    static_assert(std::size(kColumnPropertiesByEnum) == std::size(m_columnIndexByEnum));
```

src/library/basetrackcache.h
```diff
@@ -64,6 +64,7 @@ class BaseTrackCache : public QObject {
     QString columnNameForFieldIndex(int index) const;
     QString columnSortForFieldIndex(int index) const;
     int fieldIndex(ColumnCache::Column column) const;
+    int endFieldIndex() const;
```
Swiftb0y, line 67: As far as I can tell, the only reason for `endFieldIndex()` to exist is the optimization of calling `resize` with it. I don't think that warrants the complexity.

src/library/columncache.cpp
```diff
@@ -154,3 +269,25 @@ void ColumnCache::slotSetKeySortOrder(double notationValue) {
...
+const QString& ColumnCache::columnName(Column column) const {
+    DEBUG_ASSERT(static_cast<std::size_t>(column) < std::size(kColumnPropertiesByEnum));
+    DEBUG_ASSERT(kColumnPropertiesByEnum[column].pName);
+    return *kColumnPropertiesByEnum[column].pName;
+}
+
+QString ColumnCache::columnTitle(Column column) const {
+    DEBUG_ASSERT(static_cast<std::size_t>(column) < std::size(kColumnPropertiesByEnum));
+    return QCoreApplication::translate(
+            "PredefinedColorPalettes", kColumnPropertiesByEnum[column].pTrTitle);
```
Swiftb0y: this should be `"ColumnProperties"`, right?

Swiftb0y (commented):
mostly LGTM. Couple nits

src/library/basetracktablemodel.h
daschuer: I was able to get rid of getFieldString() getFieldVariant() is still required.

Swiftb0y: No I meant the `getField*(*, const QString& fieldName)` overloads. Having both `getFieldString` and `getFieldVariant` is fine.

src/library/basetrackcache.h
daschuer: There was before a more cumbersome logic for resizing. Removing it will introduce undefined behaviour or repeated allocations. So I think it is better to keep it.

Swiftb0y: well does it really optimize that much? I feel like it would be simpler to just `.clear()`, then call `.reserve()` with a good guess and then `push_back` accordingly. That'll only reallocate if the vector was actually too small. Its also safer than doing the `resize()` and then indexing directly.

src/library/columncache.cpp
daschuer: This does not work in a static assert because m_columnIndexByEnum is no constexpr
```
/home/daniel/workspace/mixxx2/src/library/columncache.cpp:214:54: error: non-constant condition for static assertion
  214 |     static_assert(std::size(kColumnPropertiesByEnum) == std::size(m_columnIndexByEnum));
      |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/daniel/workspace/mixxx2/src/library/columncache.cpp:214:54: error: use of ‘this’ in a constant expression
```

Swiftb0y: Interesting. https://brevzin.github.io/c++/2020/02/05/constexpr-array-size/

Then revert. Sorry I wasn't aware of this edge case.

src/library/basetrackcache.h
daschuer: The issue is that we can't use `push_back()`, because the m_columnHeaders is filled via the enum and not in consecutive order.

src/library/columncache.cpp
daschuer: Oh yes that was a regression. Thank you for the finding.

daschuer:
Done.

src/library/basetracktablemodel.h
```diff
@@ -21,6 +30,10 @@ class BaseTrackTableModel : public QAbstractTableModel, public TrackModel {
             const char* settingsNamespace);
     ~BaseTrackTableModel() override = default;
 
+    QVariant getFieldVariant(const QModelIndex& index, ColumnCache::Column column) const;
+    QVariant getFieldVariant(const QModelIndex& index, const QString& fieldName) const;
```
Swiftb0y, line 34: any chance of removing this overload too?

src/library/basetracktablemodel.cpp
```diff
@@ -1008,11 +852,8 @@ QVariant BaseTrackTableModel::roleValue(
 
 bool BaseTrackTableModel::isBpmLocked(
         const QModelIndex& index) const {
-    const auto bpmLockIndex =
-            index.sibling(
-                    index.row(),
-                    fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_BPM_LOCK));
-    return bpmLockIndex.data().toBool();
+    return getFieldVariant(index, ColumnCache::COLUMN_LIBRARYTABLE_BPM_LOCK)
+                    .toBool();
```
Swiftb0y: pre-commit

src/library/basetrackcache.h
Swiftb0y: fine

src/library/basetracktablemodel.h
daschuer: Unfortunatly not. This is used by https://github.com/daschuer/mixxx/blob/eb38c67c0b17d6b0677f9f9050b4f46c94b83184/src/library/basesqltablemodel.cpp#L787
Which can point to diffrent ColumnCache::Column enums. 
(Of cause it is, but not with a bigger refactoring)

src/library/basetracktablemodel.cpp
daschuer: This is already fixed. GitHub is still processing the updat. I will try to edit the PR to get out of this lock situation.

daschuer: This seems to did the trick.

src/library/basetracktablemodel.h
Swiftb0y: Thanks for the clarification.

Swiftb0y:
Not sure whats wrong with the pre-commit job not running. Its probably fine.

Swiftb0y (approved):

Issue #13783, opened 2024-10-19
Error with Stems in directories with special symbols (Addendum to #13550)


Eve00000:
### Bug Description

I wanted to reorganize my library, put stems at the same place of the original.
Importing the stem is no problem, opening (analyzing or player) makes Mixxx crash.

Deduction of the problem learns:
F:\Music\#0500 - Artists\Anne Clark\Anne Clark - Best Of Anne Clark (1992)\19 Anne Clark - Our Darkness (12_).stem.m4a
plays perfect, analazer, deck, sampler ...
F:\Music\#0500 - Artists\Anne Clark •\Anne Clark - Best Of Anne Clark (1992)\19 Anne Clark - Our Darkness (12_).stem.m4a
makes Mixxx crash. Normal ALAC files in this directory play fine.

The difference betweeb the 2 locations is the " •" after the name in the artist directory (to mark a directory as a Artist Container).

I compaired 
src\sources\soundsourceffmpeg.cpp
src\sources\soundsourceflac.cpp
src\sources\soundsourcemp3.cpp
src\sources\soundsourcestem.cpp
src\sources\soundsourceopus.cpp
src\sources\soundsourcewv.cpp
src\sources\soundsourcem4a.cpp
they all have a different way to translate the location, 
I will highlight the windows lines

M4A: 
m_hFile = MP4Read(getLocalFileName().toUtf8().constData(), 0);
Opus: 
QByteArray qBAFilename = getLocalFileName().toUtf8();
...
in some cases the filename is converted to utf8 and const, on otherr formats it isn't.

I played around with converting the "getLocalFileName()" without success.
I think #13550 suffers of a similar problem.
As I can remember 2.3.4 (or 5) had the same problem with " •" and other symbols for all files but the next release solved that.
I tried to find the pr / solution but couldn't find it.

### Version

2.5 & 2.6

### OS

Win

Eve00000:
Solved in PR #13784

PR #13784, opened 2024-10-19
Solves problem with special characters in path to stems

Eve00000:
This change solves the problem when special charactersare in the name of the directories containing stems. In my case ' •'
See issue #13783.

### DIFF
--- a/src/sources/soundsourceffmpeg.cpp
+++ b/src/sources/soundsourceffmpeg.cpp
@@ -290,7 +290,7 @@ AVFormatContext* SoundSourceFFmpeg::openInputFile(
     // Open input file and allocate/initialize AVFormatContext
     const int avformat_open_input_result =
             avformat_open_input(
-                    &pavInputFormatContext, fileName.toLocal8Bit().constData(), nullptr, nullptr);
+                    &pavInputFormatContext, fileName.toUtf8().constData(), nullptr, nullptr);
     if (avformat_open_input_result != 0) {
         DEBUG_ASSERT(avformat_open_input_result < 0);
         kLogger.warning().noquote()


acolombier:
This code was copied from the FFMPEG sounds source so it is probably impacting this one too. Shall we fix it as well?

Eve00000:
> This code was copied from the FFMPEG sounds source so it is probably impacting this one too. Shall we fix it as well?

I don't know if this solution only fixes the problem on windows or also on other OS's.

acolombier:
Best would probably be to add some unit tests just to be sure!

JoergAtGithub:
@Eve00000 Will you add a unit test as suggested by Antoine, or should we merge it as it is?

Eve00000:
I also wanted to check some other soundsources, 
I wanted to add a test file to  the test routines with a name that contains a special character in a directory with a special character. but I'm not there yet.  (doing something else at the mometn, you'll see)

nickwhaley:
Looked through the other sound sources. None are using toLocal8Bit and are all passing the correct encoding of filenames.

JoergAtGithub:
Let's merge this fix this now!
@Eve00000 Just create a new PR to add the test for this, when it's ready!

daschuer:
Hi @Eve00000, 

it looks like we have not collected the formal agreement form you to distribute you changed with Mixxx. Can you singn:
https://docs.google.com/a/mixxx.org/spreadsheet/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ
And comment here when done. 

Thank you.

Eve00000:
Done

ronso0:
Should this go to 2.6? :thinking:

ronso0:
And is this a general problem with some encoders?
See #15681

JoergAtGithub:
This PR got merged into main one year ago. Shouldn't it be in 2.6 branch already?

ronso0:
I think it should, but it's only in main.
2.6:
https://github.com/mixxxdj/mixxx/blob/5df3bb2d113f897a4b65c044b93b6b5e354d41ae/src/sources/soundsourceffmpeg.cpp#L291-L293

Do you mind pushing the backport? Or shall I?

JoergAtGithub:
Please go ahead!

acolombier:
If you could add some test case as well, this would be awesome, tho definitely not a blocker 😃
I assume the backport should address #15681

Issue #13758, opened 2024-10-13
Replace module wide Qt includes to improve build performance


JoergAtGithub:
### Bug Description

According to this KDAP article, using type specific Qt include files instead of module wide include files can safe about 15% build time: https://www.kdab.com/beware-of-qt-module-wide-includes/

We use module wide Qt includes at many places especially for Qml and QSql.

Clazy v1.12 reports an error for each of them unless we specifiy: -Wno-no-module-include

PR #13788, opened 2024-10-20
Replace module wide Qt includes to make 2.5 compliant with Clazy 1.12

JoergAtGithub:
Closes: #13758


daschuer (approved):
All green, thank you for the likely tedious work. 
LGTM.

PR #13731, opened 2024-10-08
(skins) allow to set LaunchImage style per color scheme

ronso0:
A small, low-impact tweak.

New look for
**LateNight Classic**
Screenshot_2024-10-08_13-19-46

**Shade Dark**
Screenshot_2024-10-08_13-24-39

**Shade SummerSunset**
Screenshot_2024-10-08_13-25-18

PR #13796, opened 2024-10-22
LateNight: use Classic launch image style also for 64 samplers version

ronso0:
follow-up for #13731


JoergAtGithub:
Works as expected! Thank you!

PR #13804, opened 2024-10-28
Update mapping for Stanton-DJC-4: [Library],GoToItem

VJPandaAU:
shift-scratch mapped to GoToItem

### DIFF
--- a/res/controllers/Stanton-DJC-4.midi.xml
+++ b/res/controllers/Stanton-DJC-4.midi.xml
@@ -2579,6 +2579,16 @@
         </options>
       </control>
       <control>
+        <group>[Library]</group>
+        <key>GoToItem</key>
+        <description>GoToItem - contextual behaviour.</description>
+        <status>0x90</status>
+        <midino>0x47</midino>
+        <options>
+          <normal/>
+        </options>
+      </control>
+      <control>
         <group>[Channel1]</group>
         <key>DJC4.deck[0].wheelTouch</key>
         <status>0x90</status>


Swiftb0y:
Welcome and thanks for improving this! Please sign the [Mixxx Contributor Agreement](https://docs.google.com/forms/d/e/1FAIpQLScC9QG327sjLL0eWftmfGUasxFFLxg6LCXJ2xHDYRzFIRqyiw/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ) and comment here when you have done so. It gives us permission to distribute your contribution under the GPL v2 or later license and the Apple Mac App Store. It is also helpful for us to have contact information for contributors in case we may need it in the future.

Swiftb0y:
Since this is a change to a pre-existing controller mapping, I must also ask you to update [the manual](https://github.com/mixxxdj/manual/) accordingly. Thank you.

res/controllers/Stanton-DJC-4.midi.xml
```diff
@@ -2578,6 +2578,16 @@
...
       </control>
+      <control>
+        <group>[Library]</group>
+        <key>GoToItem</key>
+        <description>GoToItem - contextual behaviour.</description>
+        <status>0x90</status>
+        <midino>0x47</midino>
+        <options>
+          <normal/>
+        </options>
```
Swiftb0y, line 2589: does this already cover both decks?

Swiftb0y: well this was more about the fact that from what I can tell, there is one shift button and one vinyl button per deck and this only maps a single message. So this seems to me like you're only mapping one button even though there is one per deck (two per controller).

Swiftb0y: Ideally, GoToItem is mapped to the browse encoder pressed down, but I assume this hardware doesn't have that?

ronso0: We were discussing other options in the AutoDJ thread.
This devices has two rotary/push encoders. @xxxmicrobexxx wanted to map AutoDJ actions to the lower encoder, while I think it would make more sense to have those on the per-deck vinyl buttons and move all library actions to the center section.
tbc

ronso0:
@xxxmicrobexxx Do you plan to continue here?

acolombier:
I'll make that a draft PR for now, but feel free to mark it ready if it should be reconsidered for merging.

Issue #13799, opened 2024-10-26
DEBUG ASSERT: "delay_frames > 0" in EchoEffect::processChannel()


ronso0:
### Bug Description

I hit DEBUG ASSERT: "delay_frames > 0" in  EchoEffect::processChannel() at ./[src/effects/backends/builtin/echoeffect.cpp:151](https://github.com/mixxxdj/mixxx/blob/75a44d66316cbc1679eb8010df247626b4cdf710/src/effects/backends/builtin/echoeffect.cpp#L151)
when repeatedly triggering `hotcue_1_setloop` (with default 4 beats).

Strange thing is that the track was not playing.

### Version

main

SjxSubham:
I think The issue arises when `delay_frames` is calculated to be zero or negative, which happens if the track is not playing and the `beat_length` or period results in a very small value.

#13805

PR #13805, opened 2024-10-28
This adjustment guarantees delay_frames will always be positive, preventing the assertion from triggering when the track is not playing.

SjxSubham:
In the `EchoEffect::processChannel()` function, when the track is not playing, a default value for `delay_frames` should be set to ensure it is greater than zero and to avoid assertion failure. You can set `delay_frames` to a minimal positive value like `1` to handle the case when the track is not playing.

This ensures that `delay_frames` is always set to a positive value, avoiding the assertion failure.

Hope this will work...
Issue #13799


JoergAtGithub:
This would only eliminate the symptom, but does not fix the cause. To eliminate the cause, you must first get an understanding of why these values occur.

PR #13216, opened 2024-05-06
Saved beat jumps

acolombier:
This addresses the feature request #13161

Currently, the activating the saved jump result in enabling or disabling the auto jump (different icon used to provided feedback on jump status)

https://github.com/mixxxdj/mixxx/assets/7086688/cd2084a8-a5d9-4b19-9edc-c52e57378a04

TODO:
- Normal hotcue behaviour if playback ahead
- Other theme?
- Manual update

~~Depends on #13215~~ merged

PR #13808, opened 2024-10-28
add controls to order Hotcues by position

ronso0:
This is a small helper for my hotcue workflow, see below.

**TL;DR:**
Adds two controls to order hotcues by position in the track:
6 4 5 2 ==> 2 4 5 6
(only index is changed, see static hotcue colors in the waveform)
* `sort_hotcues` just reorder (keeps unset cues)
hotcues-sort-75

* `sort_hotcues_remove_offsets` reorder and compress: hotcues start at 1, remove unset cues
hotcues-sort-compress-75

### TODO
- [x] add tests
- [x] include loop cues and jump cues
- [x] move actions to 'Hotcue' menu (track menu)

There's no a 'Hotcues' menu with these two actions:
hotcues-menu

**Background:**
I usually set 1 at the start/first sound, 2 at the first beat (ifi t's not 1 already), 3 left empty for some interesting spot in long intros, 4 first phrase for mixing (main) beat etc.
So sometimes it turns out I don't need 2 & 3 but need some extra cues in between 4 - 8.
Or I need extra cues before 4.
Being able to rearrange cues with #13394 is great, but might be cumbersome, especially when I'm in a hurry during (live) preparation

Draft until #13394 has been merged and there's some agreement how to proceed here.

### DIFF
--- a/src/track/track.cpp
+++ b/src/track/track.cpp
@@ -989,6 +989,67 @@ void Track::shiftCuePositionsMillis(double milliseconds) {
     markDirtyAndUnlock(&locked);
 }
 
+void Track::setHotcueIndicesSortedByPosition(HotcueSortMode sortMode) {
+    auto locked = lockMutex(&m_qMutex);
+
+    // Populate lists of positions and indices
+    QList<int> indices;
+    QList<mixxx::audio::FramePos> positions;
+    indices.reserve(m_cuePoints.size());
+    positions.reserve(m_cuePoints.size());
+    for (const CuePointer& pCue : std::as_const(m_cuePoints)) {
+        if (pCue->getType() != mixxx::CueType::HotCue &&
+                pCue->getType() != mixxx::CueType::Loop &&
+                pCue->getType() != mixxx::CueType::Jump) {
+            continue;
+        }
+        const auto pos = pCue->getPosition();
+        positions.append(pos);
+        if (sortMode == HotcueSortMode::KeepOffsets) {
+            // We shall keep empty hotcues (start offset, gaps), so we need
+            // to store the indices
+            indices.append(pCue->getHotCue());
+        }
+    }
+
+    std::sort(positions.begin(), positions.end());
+    if (sortMode == HotcueSortMode::KeepOffsets) {
+        DEBUG_ASSERT(positions.size() == indices.size());
+        std::sort(indices.begin(), indices.end());
+    }
+
+    // The actual sorting:
+    // re-map hotcue positions to indices in ascending order
+    QHash<mixxx::audio::FramePos, int> posIndexHash;
+    if (sortMode == HotcueSortMode::RemoveOffsets) {
+        // Assign new indices, start with 0
+        int index = mixxx::kFirstHotCueIndex;
+        for (int i = 0; i < positions.size(); i++) {
+            posIndexHash.insert(positions[i], index);
+            index++;
+        }
+    } else { // HotcueSortMode::KeepOffsets
+        // Assign sorted indices
+        for (int i = 0; i < positions.size(); i++) {
+            posIndexHash.insert(positions[i], indices[i]);
+        }
+    }
+
+    // Finally set new indices on hotcues
+    for (CuePointer& pCue : m_cuePoints) {
+        if (pCue->getType() != mixxx::CueType::HotCue &&
+                pCue->getType() != mixxx::CueType::Loop &&
+                pCue->getType() != mixxx::CueType::Jump) {
+            continue;
+        }
+        int newIndex = posIndexHash.take(pCue->getPosition());
+        pCue->setHotCue(newIndex);
+    }
+
+    markDirtyAndUnlock(&locked);
+    emit cuesUpdated();
+}
+
 void Track::analysisFinished() {
     emit analyzed();
 }
--- a/src/widget/wtrackmenu.cpp
+++ b/src/widget/wtrackmenu.cpp
@@ -217,6 +217,9 @@ void WTrackMenu::createMenus() {
         m_pColorMenu->setTitle(tr("Select Color"));
     }
 
+    m_pHotcueMenu = make_parented<QMenu>(this);
+    m_pHotcueMenu->setTitle(tr("Hotcues"));
+
     if (featureIsEnabled(Feature::Reset)) {
         m_pClearMetadataMenu = make_parented<QMenu>(this);
         //: Reset metadata in right click track context menu in library
@@ -470,6 +473,23 @@ void WTrackMenu::createActions() {
 
         m_pClearAllMetadataAction = make_parented<QAction>(tr("All"), m_pClearMetadataMenu);
         connect(m_pClearAllMetadataAction, &QAction::triggered, this, &WTrackMenu::slotClearAllMetadata);
+
+        m_pSortHotcuesByPositionCompressAction = make_parented<QAction>(
+                tr("Sort hotcues by position (remove offsets)"), m_pHotcueMenu);
+        connect(m_pSortHotcuesByPositionCompressAction,
+                &QAction::triggered,
+                this,
+                [this]() {
+                    slotSortHotcuesByPosition(HotcueSortMode::RemoveOffsets);
+                });
+        m_pSortHotcuesByPositionAction = make_parented<QAction>(
+                tr("Sort hotcues by position"), m_pHotcueMenu);
+        connect(m_pSortHotcuesByPositionAction,
+                &QAction::triggered,
+                this,
+                [this]() {
+                    slotSortHotcuesByPosition(HotcueSortMode::KeepOffsets);
+                });
     }
 
     if (featureIsEnabled(Feature::BPM)) {
@@ -570,6 +590,7 @@ void WTrackMenu::createActions() {
 }
 
 void WTrackMenu::setupActions() {
+    addSeparator();
     if (featureIsEnabled(Feature::SearchRelated)) {
         addMenu(m_pSearchRelatedMenu);
     }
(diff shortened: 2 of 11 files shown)


Eve00000:
Thank you @ronso0 , I realy like this! Both thumbs up!
I also have my own numbering of hotcues, quite similar as yours.

src/track/track.cpp
```diff
@@ -989,6 +989,63 @@ void Track::shiftCuePositionsMillis(double milliseconds) {
     markDirtyAndUnlock(&locked);
 }
 
+void Track::setHotcueIndicesSortedByPosition(bool compress) {
+    auto locked = lockMutex(&m_qMutex);
```
Swiftb0y, line 993: This would be a good candidate for a unit test. wdyt?

src/widget/wtrackmenu.cpp
```diff
@@ -463,6 +463,23 @@ void WTrackMenu::createActions() {
...
+        connect(m_pSortHotcuesByPositionCompressAction,
+                &QAction::triggered,
+                this,
+                [this]() {
+                    slotSortHotcuesByPosition();
+                });
+        m_pSortHotcuesByPositionAction = new QAction(
+                tr("Sort hotcues by position (keep offsets)"), m_pClearMetadataMenu);
+        connect(m_pSortHotcuesByPositionAction,
+                &QAction::triggered,
```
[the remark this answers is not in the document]
ronso0: Yeah, just "(remove offsets)" for the first case would be more clear.

src/track/track.cpp
ronso0: :+1:

Swiftb0y: Do you have time to implement one?

ronso0: Sure, done.

ywwg:
oh I would love to merge this in

ywwg:
personally I assign hotcue numbers based on proportional position in the track.  1 is near the beginning of the track, 4 is a little past the middle, 8 is at the end.  This isn't always possible but helps keep them sorted manually

src/track/track.cpp
```diff
@@ -989,6 +989,63 @@ void Track::shiftCuePositionsMillis(double milliseconds) {
     markDirtyAndUnlock(&locked);
 }
 
+void Track::setHotcueIndicesSortedByPosition(bool compress) {
```
ywwg, line 992: Even when there are only two options, I would prefer to make this an enum, not a bool.  This makes it more clear at the call site what the parameter is for.  (Also I would like to add my own scheme, assigning numbers by relative track position, which would be a third option here)

ywwg: that would be a post-merge change

Swiftb0y: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#i4-make-interfaces-precisely-and-strongly-typed

ronso0:
Besides the enum aspect, how do you think we move forward here?
Just merge the COs, then discuss how to make them accessible in the GUI?
* track menu? it's huge aready and these options may not be needed
* two buttons in the cue menu?

Swiftb0y:
> Just merge the COs, then discuss how to make them accessible in the GUI?

I'm not sure. We currently only have GUI language to talk about a single hotcue or the entire track... Adding a button that affects all hotcues to a menu that only discerns a single hotcue does not make sense to me. We could add another entry to the context menu or we could add some sort of visible container of the hotcues in the skins for it. There are plenty of (not yet implemented) actions that apply to all hotcues. Maybe some sort of extra cogwheel button next to the hotcue grid could work?

src/track/track.cpp
ronso0: I made it an enum class and rebased.
Let me know when you want to start reviewing, I'll stop force-pushing then.

ywwg:
Does everything need to be accessible from a controller? This is a very "preparing tracks" operation and does not seem like something that needs to be available at performance time.  Putting it only in the right-click menu seems fine to me

ronso0:
No of course not. Here I have unconsciously started with the CO, then added the track menu action.

I noticed that already in the first session I used it a lot, I think on S4 I'll map it to Shift+Hotcues longpress or something (the padmode button) instead of via track menu. Now the actions are at the top level for quick access (demo) but they should indeed be in a Hotcue submenu.

ronso0:
> We currently only have GUI language to talk about a single hotcue or the entire track... Adding a button that affects all hotcues to a menu that only discerns a single hotcue does not make sense to me. We could add another entry to the context menu or we could add some sort of visible container of the hotcues in the skins for it. There are plenty of (not yet implemented) actions that apply to all hotcues. Maybe some sort of extra cogwheel button next to the hotcue grid could work?

True, a dedicated 'all hotcues' edit section could be helpful.
Though I'd say we're good atm with the Clear action in the track menu, the shift controls next to the waveforms and color settings in the preferences.

Swiftb0y:
sure... though in the long-term collecting those actions in their own menu would definitely make sense.

ronso0:
Well, how to proceed here, what do you think?
Do you agree to move the actions into new 'Hotcue' menu now (track menu)?

Swiftb0y:
yeah, lets do that. anything else is for a follup up PR.

ronso0:
Okay, they're in the menu now.
Most importantly I included LoopCues and JumpCues (still unused but there's #13216) which would otherwise have been overwritten :grimacing: 

I decided to rebase to resolve the conflicts since all review comments have been addressed, hope that's okay?

src/widget/wtrackmenu.cpp
```diff
@@ -2170,6 +2210,28 @@ void WTrackMenu::slotClearHotCues() {
...
+                "Sorting hotcues of %n track(s) by position (remove offsets)",
+                "",
+                getTrackCount());
+        break;
+    case HotcueSortMode::KeepOffsets:
+    default:
+        progressLabelText =
+                tr("Sorting hotcues of %n track(s) by position", "", getTrackCount());
+        break;
+    }
```
Swiftb0y, line 2227: lets avoid the `default`. That we get warnings in case you add another sort mode and don't update the switch-case here.
```suggestion
    case HotcueSortMode::KeepOffsets:
        progressLabelText =
                tr("Sorting hotcues of %n track(s) by position", "", getTrackCount());
        break;
    }
```

Swiftb0y (commented):
last nit

src/widget/wtrackmenu.cpp
ronso0: I guess you mean to separate the `default` branch and print a warning, or even add a DEBUG_ASSERT there?

Swiftb0y: No, just remove it completely. See this [godbolt](https://compiler-explorer.com/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:1,endLineNumber:29,positionColumn:1,positionLineNumber:29,selectionStartColumn:1,selectionStartLineNumber:29,startColumn:1,startLineNumber:29),source:'enum+class+E+%7B%0A++++A,%0A++++B,%0A%7D%3B%0A%0Aint+foo(E+v)+%7B%0A++++switch+(v)+%7B%0A++++//+commenting+out+one+of+the+below+cases+will+result+in+a+compile+error+(as+intended)%0A++++case+E::A:%0A++++++++return+42%3B%0A++++case+E::B:%0A++++++++return+43%3B%0A++++%7D%0A++++//+the+return+afterwards+is+needed+because+the+technical+definition+of+enums+is+weird+(to+allow+for+bitflags)+%0A++++return+0%3B%0A%7D%0A%0A%0Aint+bar(E+v)+%7B%0A++++switch+(v)+%7B%0A++++//+while+the+same+doesn!'t+happen+when+a+%60default%60+case+is+present:%0A++++//+case+E::A:%0A++++//+++++return+42%3B%0A++++default:%0A++++case+E::B:%0A++++++++return+43%3B%0A++++%7D%0A%7D%3B%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:g142,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!(),options:'-Wall+-Werror',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+14.2+(Editor+%231)',t:'0')),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+14.2',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+14.2+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) for a demonstration.

ronso0: Okay, so you say the compiler warning is sufficient/better (as long as they're fatal)

Swiftb0y: yes, much better than no warning and a best guess runtime handling. The more we can do/know at compile-time the better.

Swiftb0y (approved):
LGTM. Thank you. feel free to squash

ywwg:
happy to get this in!

Swiftb0y:
Thanks for the reminder

PR #13816, opened 2024-10-31
Code cleanup in SidebarModel and WLibrarySidebar

cr7pt0gr4ph7:
Just a few small code format cleanups split off from a larger PR series for implementing crate folders.

- SidebarModel: Hide debug logging behind a `constexpr` flag instead of comments & use correct `printf` format for `void*` pointer (to avoid clazy warning/error)
- WLibrarySidebar: Remove `#ifdef`'d code for pre-Qt6 versions, adjust linebreaks in comment for readability, add `const` to declaration of `dragMoveAccept` to match the definition

### DIFF
--- a/src/library/sidebarmodel.cpp
+++ b/src/library/sidebarmodel.cpp
@@ -18,7 +18,7 @@ namespace {
 constexpr int kPressedUntilClickedTimeoutMillis = 300;
 
 /// Enables additional debugging output.
-constexpr bool sDebug = false;
+constexpr bool kDebug = false;
 
 } // anonymous namespace
 
@@ -104,7 +104,7 @@ void SidebarModel::activateDefaultSelection() {
 
 QModelIndex SidebarModel::index(int row, int column,
                                 const QModelIndex& parent) const {
-    if constexpr (sDebug) {
+    if constexpr (kDebug) {
         qDebug() << "SidebarModel::index row=" << row
                  << "column=" << column << "parent=" << parent;
     }
@@ -143,7 +143,7 @@ QModelIndex SidebarModel::index(int row, int column,
 }
 
 QModelIndex SidebarModel::getFeatureRootIndex(LibraryFeature* pFeature) {
-    if constexpr (sDebug) {
+    if constexpr (kDebug) {
         qDebug() << "SidebarModel::getFeatureRootIndex for" << pFeature->title().toString();
     }
     QModelIndex ind;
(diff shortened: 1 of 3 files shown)


src/library/sidebarmodel.cpp
```diff
@@ -17,6 +17,9 @@ namespace {
 /// been chosen as a compromise between usability and responsiveness.
 constexpr int kPressedUntilClickedTimeoutMillis = 300;
 
+/// Enables additional debugging output.
+constexpr bool sDebug = false;
```
JoergAtGithub, line 21:
```suggestion
constexpr bool kDebug = false;
```
Our code style is to use the prefix k for const expressions: https://github.com/mixxxdj/mixxx/wiki/Coding-Guidelines#constexpr

JoergAtGithub (changes requested):

src/library/sidebarmodel.cpp
cr7pt0gr4ph7: Thanks, fixed! I was using the existing code as a guideline, which only ever uses `constexpr bool sDebug` (albeit only in two places).

src/widget/wlibrarysidebar.cpp
```diff
@@ -93,19 +89,17 @@ void WLibrarySidebar::dragMoveEvent(QDragMoveEvent * event) {
             if (sidebarModel) {
                 accepted = false;
                 for (const QUrl& url : urls) {
-#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
```
JoergAtGithub, line 96: We have another policy, that says "backwards compatibility hacks for older dependencies should be removed only if new Mixxx features are incompatible with those old dependencies.": https://github.com/mixxxdj/mixxx/wiki/Minimum-requirements-policy

Even if I doubt, that main can be build with Qt5, we've to restore this, to comply with the policy.

JoergAtGithub (changes requested):

src/widget/wlibrarysidebar.cpp
cr7pt0gr4ph7: Okay, I reverted those specific changes. I also opened #13826 (which I actually did because of a separate review thread, before I was aware of your comment) to discuss whether it might make sense to reevaluate Qt5 vs. Qt6 when 4 years have passed since the last discussion.

JoergAtGithub:
LGTM! Thank you!

Issue #13754, opened 2024-10-11
Gradual Volume Reduction Feature


ckmrt:
### Feature Description

Add Gradual Volume Reduction Button for Non-MIDI Users

This PR introduces a button in Mixxx that allows users to gradually reduce the volume of the currently playing track. This feature is designed for users without MIDI controllers, such as DJs at events like Milonga, where smooth transitions are essential.

Motivation:

As a Tango DJ, I need smooth volume transitions between music sets. Manually handling volume can be disruptive. This feature provides an easy way to reduce volume gradually with a button click, improving the event's flow.

Implementation:

UI/UX: Adds a new button near the volume control.
Functionality: Gradually decreases volume in steps until it reaches a minimum level.

ronso0:
Thanks for this feature requests, it already came up a few times IIRC.

Should apply to volume increase, too, I'd say.
This behavior could be applied to all ControlpotMeter controls IMO, ie. all knobs and sliders: crossfader, picth, EQs, effect mix knobs (wet/dry) etc.
* fade_to_1
* fade_to_center
* fade_to_0

daschuer:
I can confirm the feature request and also the proposed solution. However let's look at the requirements to do it "right".  I can remember a similar request around Auto DJ and curtina tracks. 

@ckmrt can you describe your desired transition with more details, timing and involve tracks with artist an title?  

The proposed CO interface is useful for pad only controllers and wizard map-able. I am just puzzled how to determine the fade speed and how to abort a started fade process. 

Do we have experience with js mapping based solutions yet?

We may also consider to model the requirements as a tango Auto DJ mode as alternative or in addition.

ronso0:
> I am just puzzled how to determine the fade speed and how to abort a started fade process.

I thought we use a default fade speed (set in preferences) so these basic controls can be mapped to keyboard, skin buttons and controllers.

In my mind, the 'fade' controls would only set a flag and each EngineControl::process() would do the fading.
It can be aborted by setting the fade control 0, so when fade controls are toggle buttons, another button press would stop the fade.
Additionally, every manipulation of the underlying control (eg. volume) would abort the fade, too.
That could be clicking the respective slider or knob.

Of course we could provide UI feedback with frames or (responsive) overlays, like we do with the crossfader frame when AutoDJ is active.

cr7pt0gr4ph7:
[...]

PR #13817, opened 2024-10-31
(POC) Add a CO for automatically fading a channel's volume to 0

cr7pt0gr4ph7:
Currently in a quick-and-dirty WIP state. See #13754 for related discussion.


cr7pt0gr4ph7:
(Yes, I know that this probably belongs into an `EngineControl`, the current state was more or less quickly hacked together to test the idea)

PR #13717, opened 2024-10-02
Usability/Keyboard: Make the star ratings in the tracks view editable via keyboard

cr7pt0gr4ph7:
It is now possible to press ~~enter~~ the edit key (`F2`, or, once #13148 is merged, 'R` resp. the confiugured edit key) on a selected "star rating" to then change the rating via the keyboard (either left/right arrow keys or numbers). Changing the rating using the mouse is still possible, and should work & interact with the keyboard edits as expected.

ezgif-6-299426cc3f

### But why is this PR so large?
The code for handling the actual keyboard edits is quite simple. Nearly all of the complexity relates to handling the interaction between editing via mouse hover, editing via keyboard and the assumptions of the Qt `QTableView` framework code.

**In short:** Cells in the `QTableView` can either be in "edit mode" or "display mode". Pressing the edit key should enter edit mode, canceling or committing the edit should leave edit mode. BUT: The cells should also automatically enter edit mode when the mouse hovers over the cell, and exit it when the mouse leaves the cell. The rating should also temporarily adapt to the position of the mouse when hovering over the cell. But what happens e.g. when the user presses the edit key while already hovering over the cell?

All in all, the desired (and hereby implemented) behavior boils down to the following state machine:

image
[...]

PR #13818, opened 2024-10-31
Track Info Dialog: Improve keyboard usability & resize behavior

cr7pt0gr4ph7:
This PR bundles a number of keyboard usability improvements and general polish for the track info dialog.
It combines a number of small changes that are functionally independent and could be split off into their own PRs if necessary, but would likely produce merge conflicts due to most of them affecting the same set of files.

- In short, the track info dialog and all of its controls (except for the cover art image) are now easily accessible and usable via the keyboard, tab order matches visual order and small screens/large font sizes are handled more gracefully.

- The visual look is only changed minimally (the border style of the comments field now matches the other text fields & the "Import metadata" buttons display an ellipsis when collapsed below their minimum text width & a few pixel-level alignments, see below for details).

> _**PS:** If you want to quickly try it out without having to build it yourself, you can use the `io.github.cr7pt0gr4ph7.Mixxx` Flatpak published at [`cr7pt0gr4ph7/mixxx-flatpak`](https://github.com/cr7pt0gr4ph7/mixxx-flatpak) that includes this and a few other changes. Parallel installation is possible, i.e. any existing official or Flatpak installation of Mixxx will not be affected:_
> ```
> sudo flatpak remote-add --if-not-exists mixxx-flatpak https://cr7pt0gr4ph7.github.io/mixxx-flatpak/default.flatpakrepo
> sudo flatpak install io.github.cr7pt0gr4ph7.Mixxx
> ```

(As a sidenote, the changes from #13632 have been integrated).

## Summary

The Git history has been rewritten to group related commits together into the following changesets (presented in the chronological order of their Git history):

### Add "Last Played" field (2 commits)

Because, why not? This was the only piece of metadata that was available as a column but not shown in the track info dialog. There's space in the layout for it, so let's just add it.

Will be shown as `-` when not present in the track's metadata.

### General Cleanup (6 commits)

These should all (hopefully) be uncontentious:

- Fix alignment of the labels for "Comments" and "Grouping" by moving them a few pixels.
- Make text in the "Bitrate" field selectable, like the other read-only fields.
- Just a bit of miscellanceous cleanup:
  - Give the `QGroupBox`es speaking names instead of auto-generated ones
  - Use variables to avoid code duplication in `updateTrackMetatdataFields`
  - Use empty lines to logically group related fields in `dlgtrackinfo.h`.

### Edit star rating via keyboard (5 commits)

The star rating widget was focusable, but it was not possible to use the keyboard to actually change the rating.

- Pressing the `0`, `1`, ..., `9` key now sets the rating. The result will be clamped to the range `0..maxStarCount` (Reason: The existing code internally allows for a `maxStarCount` other than `5`).
- Pressing the `+` or `Right` key increments the rating.
- Pressing the `-` or `Left` key increments the rating.

As a sidenote, editing the star rating cells in the library table view was suffienctly different to be split into the separate PR #13717.

### Improve track list navigation (3 commits)

Navigation within a list of tracks has been improved by adding keyboard shortcuts (`Alt+Left`/`Alt+Right`[^1] [^2]) and conditionally enabling/disabling the prev/next buttons. Focus is kept within the currently focused editor when moving to another track, but any "on focus" behaviours (like selecting all text in a QLineEdit) for the current control are re-triggered.

[^1]: Sidenote: `Alt+Left`/`Right` was choosen because `Ctrl+Left`/`Right` and `Left`/`Right` already have a meaning within the text edit widgets.

[^2]: Sidenote: `Alt+Up`/`Down` can be added if desired.

**Future work:** It is currently not explained (or visually indicated) to the user that navigating to the previous/next track cancels changes made to the current track (like `Escape` or `Cancel` would do).

### Resize behaviour (5 commits)

- Avoid the minimum size of `WMultiLineTextEdit` (the "Comments" field) jumping to a larger size when the dialog is resized by the user.

- Text edit widgets: Show the start of the contained text (e.g. the album title) by default instead of the end when the text's length exceeds the widgets width.

- Allow the `Import metadata from MusicBrainz` and `Re-import metadata from file` to be collapsed to smaller widths by displaying an ellipsis (`...`).

### Keyboard navigation using `Up`/`Down` (6 commits)

Add `Up`/`Down` as an alternative to `Tab`/`Shift+Tab` for moving focus within the track info dialog to make the keyboard experience a bit more seamless.

- **Exception:** For the "Comments" control, focus only moves out of the text editor when the cursor is at the very start or end of the content when you start to hold down the corresponding arrow key, i.e. when you are at the start of the "Comments" field and hold down `Down` to move the cursor to the very end, focus does not leave the "Comments" field until you release the `Down` key and then press it again.

- **Exception:** For the BPM spinner on the second tab, `Up` and `Down` retain their original behavior (Note: haven't found a good way to differentiate between the different user intentions here)

### Edit color picker via keyboard (3 commits)

The color picker popover could be triggered, but not controlled, via the keyboard.

- `Space` or `Enter` on the color picker button triggers the popover and moves focus into the popover.
- `Tab`/`Shift+Tab` and `Up`/`Down`/`Left`/`Right` can be used to navigate within the color picker popover when it is open.
- Pressing `Esc` closes the color picker popover without making a selection.
- Pressing `Enter` or `Space` selects the focused color and closes the color picker popover.

### Tooltip for cover art image (1 commit)

Added a compact tooltip to the cover art image to indicate the available interactions (left click/right click).

### Merge & additional cleanup (2 commits)

- Merge `mixxxdj/mixxx:main` and resolve merge conflicts caused by #13632.
- Fix small misspelling `meatdata_buttons_layout` -> `metadata_buttons_layout`
- Fix code style issues in `wcolorpicker.cpp`

### DIFF
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -281,10 +281,6 @@ void DlgTrackInfo::init() {
             &DlgTrackInfo::slotOpenInFileBrowser);
 
     // Cover art
-    m_pWCoverArtLabel->setToolTip(
-            tr("Left-click to show larger preview") + "\n" +
-            tr("Right-click for more options"));
-
     CoverArtCache* pCache = CoverArtCache::instance();
     if (pCache) {
         connect(pCache,
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -82,6 +82,7 @@ void DlgTrackInfo::init() {
     m_propertyWidgets.insert("key", txtKey);
     m_propertyWidgets.insert("grouping", txtGrouping);
     m_propertyWidgets.insert("comment", txtComment);
+    m_propertyWidgets.insert("color", btnColorPicker);
 
     coverLayout->setAlignment(Qt::AlignRight | Qt::AlignTop);
     coverLayout->setSpacing(0);
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -83,9 +83,6 @@ void DlgTrackInfo::init() {
     m_propertyWidgets.insert("grouping", txtGrouping);
     m_propertyWidgets.insert("comment", txtComment);
 
-    coverLayout->setAlignment(Qt::AlignRight | Qt::AlignTop);
-    coverLayout->setSpacing(0);
-    coverLayout->setContentsMargins(0, 0, 0, 0);
     coverLayout->insertWidget(0, m_pWCoverArtLabel.get());
 
     starsLayout->setAlignment(Qt::AlignRight | Qt::AlignVCenter);
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -21,6 +21,7 @@
 #include "util/datetime.h"
 #include "util/desktophelper.h"
 #include "util/duration.h"
+#include "widget/wcolorpickeractionmenu.h"
 #include "widget/wcoverartlabel.h"
 #include "widget/wcoverartmenu.h"
 #include "widget/wstarrating.h"
@@ -52,7 +53,7 @@ DlgTrackInfo::DlgTrackInfo(
           m_pWCoverArtMenu(make_parented<WCoverArtMenu>(this)),
           m_pWCoverArtLabel(make_parented<WCoverArtLabel>(this, m_pWCoverArtMenu)),
           m_pWStarRating(make_parented<WStarRating>(this)),
-          m_pColorPicker(make_parented<WColorPickerAction>(
+          m_pColorPicker(make_parented<WColorPickerActionMenu>(
                   WColorPicker::Option::AllowNoColor |
                           // TODO(xxx) remove this once the preferences are themed via QSS
                           WColorPicker::Option::NoExtStyleSheet,
@@ -60,6 +61,13 @@ DlgTrackInfo::DlgTrackInfo(
                   this)),
           m_widgetSizesFixed(false) {
     init();
+}
+
+DlgTrackInfo::~DlgTrackInfo() {
+    // ~parented_ptr() needs the definition of the wrapped type
+    // upon deletion! Otherwise the behavior is undefined.
+    // The wrapped types of some parented_ptr members are only
+    // forward declared in the header file.
 }
 
 void DlgTrackInfo::init() {
@@ -308,12 +316,10 @@ void DlgTrackInfo::init() {
             &DlgTrackInfo::slotRatingChanged);
 
     btnColorPicker->setStyle(QStyleFactory::create(QStringLiteral("fusion")));
-    QMenu* pColorPickerMenu = new QMenu(this);
-    pColorPickerMenu->addAction(m_pColorPicker);
-    btnColorPicker->setMenu(pColorPickerMenu);
+    btnColorPicker->setMenu(m_pColorPicker.get());
 
     connect(m_pColorPicker.get(),
-            &WColorPickerAction::colorPicked,
+            &WColorPickerActionMenu::colorPicked,
             this,
             [this](const mixxx::RgbColor::optional_t& newColor) {
                 trackColorDialogSetColor(newColor);
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -341,6 +341,59 @@ void DlgTrackInfo::init() {
                 trackColorDialogSetColor(newColor);
                 m_trackRecord.setColor(newColor);
             });
+
+    tabWidget->installEventFilter(this);
+    txtTrackName->installEventFilter(this);
+    txtArtist->installEventFilter(this);
+    txtAlbum->installEventFilter(this);
+    txtAlbumArtist->installEventFilter(this);
+    txtComposer->installEventFilter(this);
+    txtGenre->installEventFilter(this);
+    txtGrouping->installEventFilter(this);
+    txtYear->installEventFilter(this);
+    txtKey->installEventFilter(this);
+    txtTrackNumber->installEventFilter(this);
+    txtDuration->installEventFilter(this);
+    txtBpm->installEventFilter(this);
+    txtDateAdded->installEventFilter(this);
+    txtDateLastPlayed->installEventFilter(this);
+    txtType->installEventFilter(this);
+    txtBpm->installEventFilter(this);
+    txtBitrate->installEventFilter(this);
+    txtSamplerate->installEventFilter(this);
+    txtReplayGain->installEventFilter(this);
+    txtLocation->installEventFilter(this);
+}
+
+bool DlgTrackInfo::eventFilter(QObject* pObj, QEvent* pEvent) {
+    if (pEvent->type() == QEvent::KeyPress) {
+        auto* pKeyEvent = static_cast<QKeyEvent*>(pEvent);
+
+        const bool noModifiersPressed = !(pKeyEvent->modifiers() &
+                (Qt::ControlModifier | Qt::AltModifier |
+                        Qt::ShiftModifier | Qt::MetaModifier));
+
+        if (!noModifiersPressed) {
+            return false;
+        }
+
+        if (pObj == this || qobject_cast<QLabel*>(pObj) ||
+                qobject_cast<QLineEdit*>(pObj) ||
+                qobject_cast<QTabWidget*>(pObj))
+
+            if (pKeyEvent->key() == Qt::Key_Up) {
+                if (focusPreviousChild()) {
+                    pEvent->accept();
+                    return true;
+                }
+            } else if (pKeyEvent->key() == Qt::Key_Down) {
+                if (focusNextChild()) {
+                    pEvent->accept();
+                    return true;
+                }
+            }
+    }
+    return false;
 }
 
 void DlgTrackInfo::slotApply() {
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -356,9 +356,14 @@ void DlgTrackInfo::slotPrevDlgTagFetcher() {
     loadPrevTrack();
 }
 
+QModelIndex DlgTrackInfo::getPrevNextTrack(bool next) {
+    return m_currentTrackIndex.sibling(
+            m_currentTrackIndex.row() + (next ? 1 : -1),
+            m_currentTrackIndex.column());
+}
+
 void DlgTrackInfo::loadNextTrack() {
-    auto nextRow = m_currentTrackIndex.sibling(
-            m_currentTrackIndex.row() + 1, m_currentTrackIndex.column());
+    auto nextRow = getPrevNextTrack(true);
     if (nextRow.isValid()) {
         loadTrack(nextRow);
         emit next();
@@ -366,8 +371,7 @@ void DlgTrackInfo::loadNextTrack() {
 }
 
 void DlgTrackInfo::loadPrevTrack() {
-    QModelIndex prevRow = m_currentTrackIndex.sibling(
-            m_currentTrackIndex.row() - 1, m_currentTrackIndex.column());
+    auto prevRow = getPrevNextTrack(false);
     if (prevRow.isValid()) {
         loadTrack(prevRow);
         emit previous();
@@ -557,6 +561,9 @@ void DlgTrackInfo::loadTrack(const QModelIndex& index) {
         return;
     }
     m_currentTrackIndex = index;
+    btnPrev->setEnabled(getPrevNextTrack(false).isValid());
+    btnNext->setEnabled(getPrevNextTrack(true).isValid());
+
     loadTrackInternal(pTrack);
     if (m_pDlgTagFetcher && m_pDlgTagFetcher->isVisible()) {
         m_pDlgTagFetcher->loadTrack(m_currentTrackIndex);
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -312,10 +312,6 @@ void DlgTrackInfo::init() {
     pColorPickerMenu->addAction(m_pColorPicker);
     btnColorPicker->setMenu(pColorPickerMenu);
 
-    connect(btnColorPicker,
-            &QPushButton::clicked,
-            this,
-            &DlgTrackInfo::slotColorButtonClicked);
     connect(m_pColorPicker.get(),
             &WColorPickerAction::colorPicked,
             this,
@@ -612,13 +608,6 @@ void DlgTrackInfo::slotOpenInFileBrowser() {
         return;
     }
     mixxx::DesktopHelper::openInFileBrowser(QStringList(m_pLoadedTrack->getLocation()));
-}
-
-void DlgTrackInfo::slotColorButtonClicked() {
-    if (!m_pLoadedTrack) {
-        return;
-    }
-    btnColorPicker->showMenu();
 }
 
 void DlgTrackInfo::trackColorDialogSetColor(const mixxx::RgbColor::optional_t& newColor) {
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -1,5 +1,6 @@
 #include "library/dlgtrackinfo.h"
 
+#include <QShortcut>
 #include <QSignalBlocker>
 #include <QStyleFactory>
 #include <QtDebug>
@@ -104,6 +105,18 @@ void DlgTrackInfo::init() {
                 &QPushButton::clicked,
                 this,
                 &DlgTrackInfo::slotPrevButton);
+
+        QShortcut* nextShortcut = new QShortcut(QKeySequence("Alt+Right"), this);
+        QShortcut* prevShortcut = new QShortcut(QKeySequence("Alt+Left"), this);
+
+        connect(nextShortcut,
+                &QShortcut::activated,
+                btnNext,
+                [this] { btnNext->animateClick(); });
+        connect(prevShortcut,
+                &QShortcut::activated,
+                btnPrev,
+                [this] { btnPrev->animateClick(); });
     } else {
         btnNext->hide();
         btnPrev->hide();
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -95,6 +95,11 @@ void DlgTrackInfo::init() {
     // This is necessary to pass on mouseMove events to WStarRating
     m_pWStarRating->setMouseTracking(true);
 
+    // Workaround: Align the baseline of the "Comments" label
+    // with the baseline of the text inside the comments field
+    const int topMargin = txtComment->frameWidth() + int(txtComment->document()->documentMargin());
+    lblTrackComment->setContentsMargins(0, topMargin, 0, 0);
+
     if (m_pTrackModel) {
         connect(btnNext,
                 &QPushButton::clicked,
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -379,6 +379,7 @@ void DlgTrackInfo::loadNextTrack() {
     auto nextRow = getPrevNextTrack(true);
     if (nextRow.isValid()) {
         loadTrack(nextRow);
+        refocusCurrentWidget();
         emit next();
     }
 }
@@ -387,7 +388,23 @@ void DlgTrackInfo::loadPrevTrack() {
     auto prevRow = getPrevNextTrack(false);
     if (prevRow.isValid()) {
         loadTrack(prevRow);
+        refocusCurrentWidget();
         emit previous();
+    }
+}
+
+/// Simulate moving the focus out of and then back into the currently
+/// focused widget to trigger behaviors like the "Select all on focus"
+/// for QLineEdit.
+///
+/// This is done when moving to the previous/next track, because,
+/// logically, the user has moved to a new dialog, even though we
+/// reuse the current dialog instance internally.
+void DlgTrackInfo::refocusCurrentWidget() {
+    QWidget* focusedWidget = QApplication::focusWidget();
+    if (focusedWidget && isAncestorOf(focusedWidget)) {
+        focusedWidget->clearFocus();
+        focusedWidget->setFocus(Qt::ShortcutFocusReason);
     }
 }
 
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -393,6 +393,19 @@ bool DlgTrackInfo::eventFilter(QObject* pObj, QEvent* pEvent) {
                 }
             }
     }
+    if (pEvent->type() == QEvent::FocusIn) {
+        auto* pFocusEvent = static_cast<QFocusEvent*>(pEvent);
+
+        if (pFocusEvent->reason() == Qt::TabFocusReason ||
+                pFocusEvent->reason() == Qt::BacktabFocusReason ||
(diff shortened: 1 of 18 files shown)


ronso0:
> PS: If you want to quickly try it out without having to build it yourself, you can

... just download the CI builds. Info is here https://github.com/mixxxdj/mixxx/wiki/Testing#github-pull-requests

ronso0:
Can you please provide before/after screenshots?

cr7pt0gr4ph7:
As requested, here are before and after screenshots (note: different looking window decorations may be related to Wayland vs. X11):

| Before | After (at minimum size) |
| -- | -- |
| "Before" screenshot | "After" screenshot, minimum size |

| Before | After (at an expanded size) |
| -- | -- |
| "Before" screenshot | "After" screenshot, expanded size |

| After: Unfocused comments field | After: Focused comments field |
| -- | -- |
| Unfocused comments field | Focused comments field |

| After: Focused star rating |
| -- |
| Focused star rating editor |

| After: Focus rectangle in color picker |
| -- |
| Bildschirmfoto vom 2024-11-01 20-38-29 |

src/widget/wbasicpushbutton.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    QString txtToolTip = toolTip();
+
+    if (!txtToolTip.isEmpty()) {
+        return txtToolTip;
+    }
+
+    QString txtLabel = text();
+    QSize currentSize = size();
+    QSize fullSize = sizeHint();
```
ronso0, line 83:
```suggestion
    const QString txtToolTip = toolTip();
    if (!txtToolTip.isEmpty()) {
        return txtToolTip;
    }

    const QString txtLabel = text();
    const QSize currentSize = size();
    const QSize fullSize = sizeHint();
```

src/widget/wbasicpushbutton.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    if ((currentSize.height() < fullSize.height() ||
+                currentSize.width() < fullSize.width()) &&
+            !txtLabel.isEmpty()) {
```
ronso0, line 87: Why do we need to check the height here?

src/widget/wbasicpushbutton.cpp
```diff
@@ -0,0 +1,115 @@
...
+        return txtLabel;
+    }
+
+    return QString();
+}
+
+bool WBasicPushButton::event(QEvent* e) {
+    switch (e->type()) {
+    case QEvent::ToolTip: {
+        QString toolTipToShow = buildToolTip();
```
ronso0, line 99:
```suggestion
        const QString toolTipToShow = buildToolTip();
```

src/widget/wcolorpicker.cpp
```diff
@@ -29,14 +31,117 @@ inline int idealColumnCount(int numItems) {
...
+        QWidget* parent)
+        : QPushButton(parent),
+          m_color(color),
+          m_row(row),
+          m_column(column) {
+    if (color) {
+        // Set the background color of the button.
+        // This can't be overridden in skin stylesheets.
+        setStyleSheet(
+                QString("QPushButton { background-color: %1; }")
```
ronso0, line 54:
```suggestion
                QStringLiteral("QPushButton { background-color: %1; }")
```

ronso0 (changes requested):
Thanks for this PR!

Just skimmed through the code and did a shallow first round of review.

src/widget/wbasicpushbutton.cpp
cr7pt0gr4ph7: It is possible that there is enough horizontal space but not enough vertical space to display the full button text. The simplest case I can think of is when `QPushButton::text()` contains linebreaks, but the button height is restricted to a single line.

The fallback tooltip should be enabled in all cases where the button text is not fully visible, which applies here. (Sidenote: It is not enabled when the button text is fully visible to avoid "useless"/confusing tooltips that only duplicate the information shown on the button itself.)

src/library/dlgtrackinfo.cpp
```diff
@@ -55,38 +56,52 @@ DlgTrackInfo::DlgTrackInfo(
...
+    m_propertyWidgets.insert("datetime_added", txtDateAdded);
+    m_propertyWidgets.insert("duration", txtDuration);
+    m_propertyWidgets.insert("timesplayed", txtDateLastPlayed);
+    m_propertyWidgets.insert("last_played_at", txtDateLastPlayed);
+    m_propertyWidgets.insert("filetype", txtType);
+    m_propertyWidgets.insert("bitrate", txtBitrate);
+    m_propertyWidgets.insert("samplerate", txtSamplerate);
+    m_propertyWidgets.insert("replaygain", txtReplayGain);
+    m_propertyWidgets.insert("replaygain_peak", txtReplayGain);
+    m_propertyWidgets.insert("track_locations.location", txtLocation);
```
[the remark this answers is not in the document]
cr7pt0gr4ph7: `WTrackProperty` can be used by skins for arbitrary properties, including the non-editable ones. So, basically, just for completeness sake & to provide uniform behavior for those properties, too.

cr7pt0gr4ph7:
> Thanks for this PR!
> 
> Just skimmed through the code and did a shallow first round of review.

Thank your for the quick review, very much appreciated :+1:

src/widget/wbasiclabel.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <QLabel>
+
+/// Subclass of QLabel that allows the Up and Down arrow keys
+/// to be used like the Tab/Backtab keys.
+///
+/// See WLabel when you need to connect to a ControlObject.
+class WBasicLabel : public QLabel {
```
cr7pt0gr4ph7, line 9: @ronso0 I'm thinking whether it make sense to separate the skin widgets and the "basic" widgets into separate folders, e.g `src/widget/basic`. My gut feeling would be to create a more granular folder structure, though the current practice within Mixxx seems to favor bigger directories + prefixes.

As an aside, one could consider moving the skin widgets into `src/widget/skin` - though that's a lot of code churn for probably not a lot of benefit.

src/widget/wbasiclineedit.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <QLineEdit>
+
+/// Subclass of QLineEdit that allows the Up and Down arrow keys
+/// to be used like the Tab/Backtab keys.
+class WBasicLineEdit : public QLineEdit {
```
cr7pt0gr4ph7, line 7: Same applies here

src/widget/wbasicpushbutton.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <QPushButton>
+
+/// Subclass of QPushButton that replaces a portion of the button label
+/// with an ellipsis (according to Qt::TextElideMode) when the text would
+/// otherwise overflow the available space.
+///
+/// See WPushButton when you need to connect to a ControlObject.
+class WBasicPushButton : public QPushButton {
```
cr7pt0gr4ph7, line 10: And here

src/widget/wbasictabwidget.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <QTabWidget>
+
+/// Subclass of QTabWidget that allows the Up and Down arrow keys
+/// to be used like the Tab/Backtab keys.
+class WBasicTabWidget : public QTabWidget {
```
cr7pt0gr4ph7, line 7: And here

src/widget/wmultilinetextedit.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <QPlainTextEdit>
+
+/// An implementation of QPlainTextTedit with a more sensible minimum
+/// size of 2 text lines.
+class WMultiLineTextEdit : public QPlainTextEdit {
```
cr7pt0gr4ph7, line 7: And here

src/widget/wbasiclabel.h
ronso0: Hmm, I'd rather stick to the `widgets` directory tbh.

src/widget/wbasicpushbutton.cpp
ronso0: I think we should implement only the cases we need currently and expand as soon as we need multi-line buttons.
IMHO the context should be clear so that we don't need multi-line buttons in the first place.
Or do you have an exmple where we might need one?

src/library/dlgtrackinfo.cpp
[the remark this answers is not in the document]
ronso0: Hmm AFAIU non-editable fileds (labels) can't be focused anyway, so wouldn't the UI state be somewhat unclear then?
I mean, we wouldn't know the current focus widget/position until we press Tab/Shift+Tab, right?

ronso0:
I think the buttons should stretch to full width as before
image

Previously it looked better with the stars centered below the cover art
image

ronso0:
> I think the buttons should stretch to full width as before

Oh, they already do. (just the screen shot doesn't reflect that)

ronso0:
It's nice the star rating can now be focused, but a focus frame would be helpful to make it clear.

ronso0:
The color button now is two lines tall for me (Qt 6.2.3), both in DlgTrackInfo and DlgTrackInfoMulti.

ronso0:
Just tried this branch and navigation feels nice indeed!
I wonder if the Down/Up navigation will discovered by users since it's rather unusual (at least I didn't see it in other apps, yet). I guess all we can do is document it in the manual.

re Down/Up for moving focus:
There's an inconsistency now between the single and multi-track dialog now since pressing on the multi-track fields (QComboBoxes) chnages the selected item. Not sure if we should override/disable that and let users open the drop-down with Ctrl+Space like in WSearchLineEdit.

ronso0:
I'll take a look at this huge change set soonish.

src/widget/wbasiclabel.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include <QLabel>
+
+/// Subclass of QLabel that allows the Up and Down arrow keys
+/// to be used like the Tab/Backtab keys.
```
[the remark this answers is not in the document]
ronso0: or prefix the class, eg. WDlgLabel?
(applies to all new sub-classes)

ronso0 (changes requested):
Some more findings

src/library/dlgtrackinfo.cpp
```diff
@@ -248,6 +275,10 @@ void DlgTrackInfo::init() {
             &DlgTrackInfo::slotOpenInFileBrowser);
 
     // Cover art
+    m_pWCoverArtLabel->setToolTip(
+            tr("Left-click to show larger preview") + "\n" +
+            tr("Right-click for more options"));
```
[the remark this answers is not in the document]
cr7pt0gr4ph7: I agree :+1: I'll move the code.

src/widget/wbasiclineedit.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <QLineEdit>
+
+/// Subclass of QLineEdit that allows the Up and Down arrow keys
+/// to be used like the Tab/Backtab keys.
```
cr7pt0gr4ph7, line 6:
```suggestion
/// Subclass of QLineEdit for DlgTrackInfo/~Multi and similar dialogs
/// that allows the Up and Down arrow keys to move keyboard focus
/// to the prev/next focusable widget.
```

src/widget/wbasictabwidget.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <QTabWidget>
+
+/// Subclass of QTabWidget that allows the Up and Down arrow keys
+/// to be used like the Tab/Backtab keys.
```
cr7pt0gr4ph7, line 6:
```suggestion
/// Subclass of QTabWidget for DlgTrackInfo/~Multi and similar dialogs
/// that allows the Up and Down arrow keys to move keyboard focus
/// to the prev/next focusable widget.
```

src/widget/wmultilinetextedit.h
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <QPlainTextEdit>
+
+/// An implementation of QPlainTextTedit with a more sensible minimum
+/// size of 2 text lines.
```
cr7pt0gr4ph7, line 6:
```suggestion
/// An implementation of QPlainTextTedit with a more sensible minimum
/// size of 2 text lines.
///
/// Used for DlgTrackInfo/~Multi and similar dialogs.
```

src/library/dlgtrackinfo.cpp
```diff
@@ -248,6 +275,10 @@ void DlgTrackInfo::init() {
...
+
```
cr7pt0gr4ph7, line 281:
```suggestion
    // Cover art
```

cr7pt0gr4ph7:
@ronso0
> Just tried this branch and navigation feels nice indeed! I wonder if the Down/Up navigation will discovered by users since it's rather unusual (at least I didn't see it in other apps, yet). I guess all we can do is document it in the manual.

Spreadsheets applications implement a similar behavior, so IMO this shouldn't confuse users too much. It's also already implemented in Qt for QPushButton, and mostly intended for power users that are used to other data entry applications.
 
> re Down/Up for moving focus: There's an inconsistency now between the single and multi-track dialog now since pressing on the multi-track fields (QComboBoxes) chnages the selected item. Not sure if we should override/disable that and let users open the drop-down with Ctrl+Space like in WSearchLineEdit.

I think that the differing behavior for `QComboBoxes` is acceptable here, as they serve a different purpose and are visually distinct enough from normal text boxes for users to quickly learn the difference.

cr7pt0gr4ph7:
@ronso0
> The color button now is two lines tall for me (Qt 6.2.3), both in DlgTrackInfo and DlgTrackInfoMulti.

I can reproduce that behavior (the color button gets taller as the dialog gets wider) for DlgTrackInfoMulti, but not for DlgTrackInfo (using Qt 6.7.3). I'll investigate, should probably be an easy fix.

ronso0:
I came back to this, and right away I remembered why I left last time:
it's too much for my taste. Not necessarily the changeset of 45 commits, I rather feel that adding a WBasic~ class for pretty much every QWidget used in the dialog is overkill. I mean Up/Down to move focus can be almost entirely implemented in DlgTrackInfo. So basically what you implemented in [`WBasicTabWidget::keyPressEvent`](https://github.com/mixxxdj/mixxx/pull/13818/files#diff-45ad59c6fe043b57ec2febfc605fdcd375d99dc650b7bfb542845b8f01e433dbR11).
A minimal POC is here https://github.com/ronso0/mixxx/commit/f586662f6575ded84b950a3c1b23152df3f3f67c

Yes, WMultiLineTextEdit is very helpful for the sizeHint and Up/Down to move focus.
Yes, a WBasicPushButton helps reduce minimum width and gives us a tooltip, just in case.

But that's about it IMHO
Wdyt?

(didn't look into WStarRating in detail, yet)

ronso0 (commented):
some comments..

src/widget/wcolorpickeractionmenu.cpp
```diff
@@ -0,0 +1,59 @@
...
+    if (event->key() == Qt::Key_F) {
+        m_pColorPickerAction->defaultWidget()->setFocus();
+        event->accept();
+        return;
+    }
+    if (event->key() == Qt::Key_A) {
+        setActiveAction(m_pColorPickerAction);
+        event->accept();
+        return;
+    }
```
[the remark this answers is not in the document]
cr7pt0gr4ph7: Oops, that is actually some debugging code I accidentally left in. Good catch, will be removed!

src/widget/wmultilinetextedit.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+bool WMultiLineTextEdit::event(QEvent* e) {
+    switch (e->type()) {
+    case QEvent::Paint: {
+        QStylePainter p(this);
+        QStyleOptionFrame option;
+        initStyleOption(&option);
+        if (isReadOnly()) {
+            option.state |= QStyle::State_ReadOnly;
+        }
```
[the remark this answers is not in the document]
cr7pt0gr4ph7: `WMultiLineTextEdit` reuses the rendering primitives from `QLineEdit` (instead of `QPlainTextEdit`) to get focus highlighting when focus is inside the control.

The default styles for `QLineEdit` provide a styling for the read-only state (when `QLineEdit::isReadOnly() == true`), and as a user, I would expect `WMultiLineTextEdit::isReadOnly()` to behave similarly.

src/widget/wstarrating.cpp
```diff
@@ -41,23 +44,115 @@ void WStarRating::slotSetRating(int starCount) {
...
+    case Qt::Key_0:
+    case Qt::Key_1:
+    case Qt::Key_2:
+    case Qt::Key_3:
+    case Qt::Key_4:
+    case Qt::Key_5:
+    case Qt::Key_6:
+    case Qt::Key_7:
+    case Qt::Key_8:
+    case Qt::Key_9: {
```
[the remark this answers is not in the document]
cr7pt0gr4ph7: `StarRating` provides a `maxStarCount()` method, implying that its value is not fixed to always be `5`. It's straightforward to handle those cases here to avoid future surprises when `WStarRating` is used in other contexts.

src/widget/wmultilinetextedit.cpp
```diff
@@ -0,0 +1,96 @@
...
+                    Qt::ShiftModifier | Qt::MetaModifier);
+
+    // Holding down a key to scroll all the way to the
+    // top or bottom of the text edit should not cause
+    // the focus to leave the text edit control.
+    const bool isAutoRepeat = event->isAutoRepeat();
+
+    if (!m_upDownChangesFocus || !(isUp || isDown) ||
+            isAutoRepeat || anyModifiersPressed) {
+        QPlainTextEdit::keyPressEvent(event);
```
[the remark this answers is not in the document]
cr7pt0gr4ph7: I think both alternatives are equally readable, so I don't have a specific preference. I'll rewrite the code as proposed 👍

(As a sidenote, performance should not be a deciding factor here, as this is in no way a hot path).

cr7pt0gr4ph7:
> I came back to this, and right away I remembered why I left last time: it's too much for my taste. Not necessarily the changeset of 45 commits, I rather feel that adding a WBasic~ class for pretty much every QWidget used in the dialog is overkill. I mean Up/Down to move focus can be almost entirely implemented in DlgTrackInfo. So basically what you implemented in [`WBasicTabWidget::keyPressEvent`](https://github.com/mixxxdj/mixxx/pull/13818/files#diff-45ad59c6fe043b57ec2febfc605fdcd375d99dc650b7bfb542845b8f01e433dbR11). A minimal POC is here [ronso0@f586662](https://github.com/ronso0/mixxx/commit/f586662f6575ded84b950a3c1b23152df3f3f67c)

IIUC, your concerns are related to `WBasicLabel` and `WBasicLineEdit`. I don't really see an advantage in mixing in the focus handling code with unrelated other code in DlgTrackInfo and DlgTrackInfoMulti, instead of cleanly separating it into the two utility classes.

ronso0:
My point is that my POC is **30 LOC** (withou debug stuff), and it's not unrelated, it's isolated in `DlgTrackInfo::eventFilter()`, theobject which receives the events, ie. easy to maintain. (+ the multi-line edit)
Whereas your proposal is **145 LOC** with focus code scattered across **3 extra classes** we'd need to maintain.
(WBasicLineEdit, WBasicTabWidget, WBasicLabel?)

cr7pt0gr4ph7:
> My point is that my POC is **30 LOC** (withou debug stuff), and it's not unrelated, it's isolated in `DlgTrackInfo::eventFilter()`, theobject which receives the events, ie. easy to maintain. (+ the multi-line edit) Whereas your proposal is **145 LOC** with focus code scattered across **3 extra classes** we'd need to maintain. (WBasicLineEdit, WBasicTabWidget, WBasicLabel?)
> 
> **edit** my new POC [0aa786d](https://github.com/mixxxdj/mixxx/commit/0aa786db60f84418c5963e37f92cf05e1c384c76) also handles the 'no item focused' case, so with that only `WMultiLineTextEdit` is required for Up/Down to change focus.

I see your point. I don't have a very strong opinion on this, so I'll refactor the code as proposed. I'll extend your POC to also handle the labels, though, as these can receive keyboard focus when clicked using the mouse, and for consistency `Up` and `Down` should work on them as well.

(For context, my own usage pattern when I don't see which control currently has the focus is to press `Up`/`Down`/`Tab` multiple times until I see the focus frame on some control in my field of view. I therefore prefer if all controls consistently support all suitable navigation patterns).

src/library/dlgtrackinfo.cpp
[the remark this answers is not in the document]
cr7pt0gr4ph7: The non-editable fields can actually be focused by clicking them with the mouse, which is desirable to be able to copy their contents to the clipboard.

After testing it out locally, I noticed that text fields perform a `selectAll()` when navigating using `Alt+Left`/`Alt+Right`, while the labels do not. I propose performing a `selectAll()` on the labels too, both on `Alt+Left`/`Alt+Right` as well as when `focusField` is called via `WTrackMenu`. This makes the visual state clear and allows easy usage of the clipboard to copy some track information.

ronso0:
Why do the labels need to get keyboard focus?

cr7pt0gr4ph7:
> Why do the labels need to get keyboard focus?

It's the reverse: The labels can get keyboard focus by clicking on them with the mouse (which is necessary to make it possible to copy their contents to the clipboard). They can't currently be focused via the keyboard, but it should be possible to move focus _away_ from them using the keyboard.

src/widget/wmultilinetextedit.cpp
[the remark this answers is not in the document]
ronso0: Okay, so this is indeed
> just for sake of complete reimplemetation

because we don't have read-only multi-lineedits (now).

src/widget/wstarrating.cpp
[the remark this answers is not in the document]
ronso0: Ah, well we had that discussion elsewhere already.
While I still think yagni applies here I won't insist.

ronso0:
> The labels can get keyboard focus by clicking on them with the mouse (which is necessary to make it possible to copy their contents to the clipboard)

Hmm this is supposed to be configured by the TextInteractionFlags. But apparently setting those sets the focus policy to ClickFocus. Interesting is that NoFocus set in ui doesn't make a difference (lable can get focus) but if set in c++ the selected text is not cleared when I select text of another label. Qt.. :man_shrugging: 

Anyhow, the build is broken currently due to some } and ; missing. Please fixup and force-push, we'd love every commit to be buildable (to ease later git bisect).

And Down press now jumps twice because of `installEventFilter(this);`. I removed it and everything still works as expected (even we no widget has focus initially)

Will take a look at the code this weekend I think.

vespadj:
Let me suggest a little request for improvement:
**A checkbox "Save on Next/Previous".**

_(Workaround: **Alt+A**pply , **Alt+N**ext)_

ronso0:
> Will take a look at the code this weekend I think.

... and _snap_ ..  a year passed.
@cr7pt0gr4ph7  Sorry for forgetting about this, will take another look soon.
Are you still motivated to work on this and resolve the conflicts that have emerged?

cr7pt0gr4ph7:
> > Will take a look at the code this weekend I think.
> 
> ... and _snap_ .. a year passed. @cr7pt0gr4ph7 Sorry for forgetting about this, will take another look soon. Are you still motivated to work on this and resolve the conflicts that have emerged?

@ronso0 And yet another year later, I'm finally getting around to resolving the conflicts :sweat_smile:

I've split this PR up into smaller PR, in the hopes of getting those merges. I'll therefore close this PR, but mention it in the new PRs.

PR #13819, opened 2024-11-01
Improved comments in enginecontrol and use of std::size_t for bufferSize accross the codebase

JoergAtGithub:
This is a spin-off of https://github.com/mixxxdj/mixxx/pull/13172#discussion_r1824896186 to keep the changes not related to Macros out of that PR.

### DIFF
--- a/src/encoder/encodermp3.cpp
+++ b/src/encoder/encodermp3.cpp
@@ -81,32 +81,32 @@ void EncoderMp3::setEncoderSettings(const EncoderSettings& settings) {
     }
 }
 
-int EncoderMp3::bufferOutGrow(int size) {
-    if (m_bufferOutSize >= size) {
+int EncoderMp3::bufferOutGrow(std::size_t bufferSize) {
+    if (m_bufferOutSize >= bufferSize) {
         return 0;
     }
 
-    m_bufferOut = (unsigned char *)realloc(m_bufferOut, size);
+    m_bufferOut = (unsigned char*)realloc(m_bufferOut, bufferSize);
     if (m_bufferOut == nullptr) {
         return -1;
     }
 
-    m_bufferOutSize = size;
+    m_bufferOutSize = bufferSize;
     return 0;
 }
 
-int EncoderMp3::bufferInGrow(int size) {
-    if (m_bufferInSize >= size) {
+int EncoderMp3::bufferInGrow(std::size_t bufferSize) {
+    if (m_bufferInSize >= bufferSize) {
         return 0;
     }
 
-    m_bufferIn[0] = (float *)realloc(m_bufferIn[0], size * sizeof(float));
-    m_bufferIn[1] = (float *)realloc(m_bufferIn[1], size * sizeof(float));
+    m_bufferIn[0] = (float*)realloc(m_bufferIn[0], bufferSize * sizeof(float));
+    m_bufferIn[1] = (float*)realloc(m_bufferIn[1], bufferSize * sizeof(float));
     if ((m_bufferIn[0] == nullptr) || (m_bufferIn[1] == nullptr)) {
         return -1;
     }
 
-    m_bufferInSize = size;
+    m_bufferInSize = bufferSize;
     return 0;
 }
 
@@ -117,7 +117,7 @@ void EncoderMp3::flush() {
         return;
     }
     // Flush also writes ID3 tags.
-    int rc = lame_encode_flush(m_lameFlags, m_bufferOut, m_bufferOutSize);
+    int rc = lame_encode_flush(m_lameFlags, m_bufferOut, static_cast<int>(m_bufferOutSize));
     if (rc < 0) {
         return;
     }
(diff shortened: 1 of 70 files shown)


Swiftb0y (commented):
two quick notes:

1. I'm fairly certain `size_t` is the C nomenclature, the proper C++ name/type live in `std`, so `std::size_t`. 
2. While I prefer to use `std::size_t` when doing indexing especially in realtime code for compatibility with `std` containers, it does go against Qt's mantra of using `int` (at least in their external APIs). Previous experiments have also shown that MSVC is apparently worse at autovectorizing buffer indexes that use `size_t`s rather than `int`s, which was the reason to typedef `SINT` to `int` rather than `size_t`.

I'd prefer if you can take care of 1.
2. I just wanted to mention so you're aware even though I don't personally think its a strong argument against `size_t`.

src/encoder/encodersndfileflac.cpp
```diff
@@ -58,10 +58,10 @@ void EncoderSndfileFlac::setEncoderSettings(const EncoderSettings& settings)
     m_compression = static_cast<double>(settings.getCompression()) / 8.0;
 }
 
-void EncoderSndfileFlac::encodeBuffer(const CSAMPLE* pBuffer, const int iBufferSize) {
+void EncoderSndfileFlac::encodeBuffer(const CSAMPLE* pBuffer, const std::size_t bufferSize) {
     if (m_pClampBuffer) {
-        SINT numSamplesLeft = iBufferSize;
-        while (numSamplesLeft > 0) {
+            SINT numSamplesLeft = bufferSize;
+            while (numSamplesLeft > 0) {
```
Swiftb0y, line 64: was this a confused clang-format or where is the bogus indentation from?

src/encoder/encodermp3.cpp
Swiftb0y, line 1: `EncoderMp3::initStream()` still assumes `m_bufferOutSize` is `int`.

src/engine/channelmixer.cpp
```diff
@@ -69,7 +69,7 @@ void ChannelMixer::applyEffectsInPlaceAndMixChannels(
     //    B) Applies effects to the buffer, modifying the original input buffer
     // 4. Mix the channel buffers together to make pOutput, overwriting the pOutput buffer from the last engine callback
     ScopedTimer t(QStringLiteral("EngineMixer::applyEffectsInPlaceAndMixChannels"));
-    SampleUtil::clear(pOutput, iBufferSize);
+    SampleUtil::clear(pOutput, bufferSize);
```
Swiftb0y, line 72: I think this needs a `std::size_t` overload, wdyt?

src/engine/channels/enginedeck.cpp
```diff
@@ -164,7 +164,7 @@ void EngineDeck::processStem(CSAMPLE* pOut, const int iBufferSize) {
     // effect manager so we can also apply the individual stem quick FX
     GroupFeatureState featureState;
     collectFeatures(&featureState);
-    for (int stemIdx = 0; stemIdx < stemCount;
+    for (unsigned int stemIdx = 0; stemIdx < stemCount;
             stemIdx++) {
```
Swiftb0y, line 168: why `unsigned int` over `std::size_t`?

daschuer:
https://github.com/mixxxdj/mixxx/blob/edae0ca3b2fc36c51d9f310f341fb2e4bbea7d26/src/util/types.h#L13

And Qt used int in Qt5 an ssize_t form Qt6
https://doc.qt.io/qt-6/qttypes.html#qsizetype-typedef

We have also a pending discussion to pass all buffers as a std::span, wich includes std::size_t a size. 

This however goes against our alternative attempt to pass buffers without the stereo assumption. This issue is also visible in the https://github.com/mixxxdj/mixxx/pull/13654

So maybe we want an own span type with a frame size?

src/engine/filters/enginefilterdelay.h
```diff
@@ -45,8 +45,7 @@ class EngineFilterDelay : public EngineObjectConstIn {
         m_delaySamples = delaySamples;
     }
 
-    virtual void process(const CSAMPLE* pIn, CSAMPLE* pOutput,
-                         const int iBufferSize) {
+    virtual void process(const CSAMPLE* pIn, CSAMPLE* pOutput, const size_t bufferSize) {
```
Swiftb0y, line 48:
```suggestion
    virtual void process(const CSAMPLE* pIn, CSAMPLE* pOutput, const std::size_t bufferSize) {
```

Swiftb0y (commented):
couple more comments. LGTM otherwise though

Swiftb0y:
> So maybe we want an own span type with a frame size?

In the long-term, yes. As I pointed out previously `std::mdspan` would likely be the good solution here, we just need to wait for GCC to implement it ([or use `Kokkos::mdspan`](https://github.com/mixxxdj/mixxx/pull/11533)).

Swiftb0y:
I think transitioning to `std::size_t` and `std::span` in the meantime is a good enough solution. In actual audio code, we may want to leverage `std::simd` anyways ([which is currently being proposed for standardization in C++26](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p1928r12.pdf)).

src/encoder/encodersndfileflac.cpp
JoergAtGithub: I think this is not clang-format itself. I've clang-format support enabled in VisualStudio and it formats it like this:
grafik
but if I ran pre-commit, it looks like this:
grafik

I will not address this in this PR!

src/engine/channelmixer.cpp
JoergAtGithub: An overload is ambiguous, we would need two function names and adjust all the callers accordingly. This seems to much work to me.

src/encoder/encodersndfileflac.cpp
Swiftb0y: > I will not address this in this PR!

Not forcing you to, though I wonder whether it makes sense to just enforce yourself and skip clang-format...

src/engine/channelmixer.cpp
Swiftb0y: well, its only ambiguous if an implicit conversion was required. That shouldn't be in many places. I will experiment in a follow up.

src/encoder/encodersndfileflac.cpp
JoergAtGithub: This I can do, but not fixing the script

JoergAtGithub: Done! CI fails now as expected.

Swiftb0y: yeah thats fine for me. Since it only runs on the diff we shouldn't be bothered by it again after the merge, right?

src/engine/channels/enginedeck.cpp
```diff
@@ -164,7 +164,7 @@ void EngineDeck::processStem(CSAMPLE* pOut, const int iBufferSize) {
     // effect manager so we can also apply the individual stem quick FX
     GroupFeatureState featureState;
     collectFeatures(&featureState);
-    for (int stemIdx = 0; stemIdx < stemCount;
+    for (size_t stemIdx = 0; stemIdx < stemCount;
```
Swiftb0y, line 167:
```suggestion
    for (std::size_t stemIdx = 0; stemIdx < stemCount;
```

Swiftb0y (approved):
LGTM otherwise. Thank you.

Swiftb0y (approved):
Thank you. LGTM.

PR #13830, opened 2024-11-03
Update 2.5 buildenv to latest version

JoergAtGithub:
Including latest libdjinterop


daschuer (approved):
Ok, all green. Thank you for ting care. I was dragged away from it.

PR #13841, opened 2024-11-05
Tracks: focus the editing field in the track properties that corresponds to the focused column

ronso0:
Just a small helper for the editing workflow.
(it's just the equivalent to the flow with the track labels in decks, see
https://github.com/mixxxdj/mixxx/blob/938cc311cd1a5f8e691691f6bbe044d8392548f6/src/widget/wtrackproperty.cpp#L187-L195)
* in the tracks table the 'Comment' column is focused
* hit Ctrl+Enter to open Track Properties (one or more tracks)
* Track Properties opens and keyboard focus is on the Comment field

### DIFF
--- a/src/library/dlgtrackinfo.ui
+++ b/src/library/dlgtrackinfo.ui
@@ -440,19 +440,6 @@
               </property>
              </widget>
             </item>
-            <item row="0" column="0">
-             <spacer name="horizontalSpacer_5">
-              <property name="orientation">
-               <enum>Qt::Horizontal</enum>
-              </property>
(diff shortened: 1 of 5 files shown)


ronso0:
completely wrong commit..
will check if I find it again.

ronso0:
here we go!

2.5 or 2.5.1, I don't mind

daschuer (approved):
Works and Looks good. 

Do we want have the same for the track context menu?

ronso0:
:thinking:

ronso0:
:+1:

PR #13842, opened 2024-11-06
(POC) add skin styles for Preferences & Track Properties, LateNight PaleMoon only

ronso0:
One of my infamous 'coffee break projects'
Reason is that I don't have / can't built Qt6ct to get a proper dark theme, yet, ie. I get the rather bright default Qt theme which is annoying when playing in dark places.
So I styled Track Properties and the preferences manually. PaleMoon only!

Still WIP but is already usable and a huge improvement for me.

**Update:**
After doing some live tests I decided to style ALL dialogs.
This required some not-so-nice fixes / hacks in c++, see (A). If someone has ideas for alternative solutions feel free to post them.

### Disclaimer:
1) For now, my intention is only to share the progress, not to get this into a mergable state, but I'll continue polishing it a bit.
2) The c++ changes probably screw up Track Properties for skins/themes other than LateNight PaleMoon because the dialogs now have a parent (WTrackMenu) so they inherit styles, either from WTrackProperty or WTrackTableView. Maybe this can be avoided by using explicit (library) object selectors in the stylesheets.
3) In order to style QGroupBoxes in the preferences I had to remove to the workaround for supporting legacy skins (< Mixxx 1.12) from LegacySkinparser. Ie. in stylesheets that use the `QGroupBox` selector that will not anymore be transformed to `WWidgetGroup`. I could add a warning so this change can be noticed by users but this has no prio at all.

(screenshots are a bit outdated but you get the idea..)
Screenshot_2024-11-06_18-16-00
Screenshot_2024-11-06_18-21-15

(A) The most important "fix" is to use the main "Skin" widget as parent for all dialogs created by non-GUI classes, accomplished with a widgethelper function that finds the `centralWidget()` "Skin" amongst the `QApplication::topLevelWindgets()`.
While it works, it introduces the tiny risk that the Skin widget is being destroyed / recreated while a dialog is open. This is rather unlikely (who would reload a skin while something like Track Export is active?) and would only kill such dialogs.
Alternative (neither evaluated nor tested):
use the mainwindow as parent, catch 'ChildAdded' events, add the centralWidget()'s stylesheet to new children.

### TODO
- [ ] check when/why style of early warnings is broken (eg. sound device error before skin load)
already called `loadConfiguredSkin()` so :man_shrugging:


ronso0:
After doing some live tests I decided to style ALL dialogs.
This required some not-so-nice fixes / hacks in c++ and makes skin load take a little longer but everything looks good now.

Eve00000:
> After doing some live tests I decided to style ALL dialogs. This required some not-so-nice fixes / hacks in c++ and makes skin load take a little longer but everything looks good now.

Is it an option to get a day/night switch in the preferences in the future?

ronso0:
A daylight color scheme would do the trick.

ronso0:
Alright, I'm now happy with this.

Looks like it's easy to keep the preferences theme intact for skins/color themes other than LateNigh PaleMoon, will finish this.

Also, I was wrong about the slowdown: skin re/load is as fast as before (besides those extra ~200 ms for styling the preferences, of course).

ronso0:
Let me know if there's interest in polishing this for getting it in main.
Otherwise I'll close this PR and move it to my repo to not spam the Mixxx repo/notifications anymore with updates.

ronso0:
> > Let me know if there's interest ...in getting it in main.
> 
> YES!

Yeah I'm sure there are users who appreciate this, though my question was rather whether core devs a) see a necessity for this and b) are okay with the c++ hacks. (and I myself am not 100% convinced that this is appropriate for the official version)

daschuer:
a) 
I am personally fine with the current state because the style aligns with the rest of the OS. But I am not opposing against this. 
The downside might be a bit more hassle later if we need to consider all skins when altering something in the preferences. 
But if you like to do the effort, go ahead. 

b) 
It does not look too scary .. 

Is there much work left? 

If you ask me what else is important regarding skins: 
* Fix the foreign title bar in: https://github.com/mixxxdj/mixxx/issues/13854
* Daylight sin: https://github.com/mixxxdj/mixxx/pull/11748

ronso0:
Thanks for your feedback!

I'm aware of the other helpful tweaks (just commented in #13854, not possible in the QWidget skins, and I might take a stab at the (LateNight) daylight theme during winter).

> It does not look too scary ..

Great!

> Is there much work left?

Yes. Keeping the other skins/themes intact is a hazzle because of how style sheets are inherited.  In particular QTooltips and WColorPicker's bg & button style in Track Properties.
I might be able to fix it with dedicated object names.

> The downside might be a bit more hassle later if we need to consider all skins when altering something in the preferences.
But if you like to do the effort, go ahead.

I don't expect trouble with that because, except for some `QLine`s in the preferences, the preferences style selectors are general, ie. addressing the QWidgets by type not by name.

ronso0:
While this works well for me when using only LateNight PaleMoon, it still needs some minor polishing (or rather porting my custom styles to this branch) and there are still some small issues to be ironed out, eg. not append (?) links twice in preferences.
Though, given that there is little interest in this besides Daniel's review, and with Mixxx QML evolving, I think it'd better to close it and spent my time on preparing the migration to a new OS in order to participate in the QML development.

Any opinions on this?

ronso0:
Will close this now for two reasons: little feedback/interest and growing maintenance effort.
In order to get a complete implementation also for dialogs like the Scanner progress dialog I had to some non-trivial refactoring (which I don't plan to backport to this branch) and I don't want to care about conflicts both here and in my prod branch.

If anyone wants to pick this up I still have related commits in my prod branch https://github.com/ronso0/mixxx/tree/rop3210

daschuer:
I am sorry about this, but OK.

ronso0:
No need to be sorry. Getting this in with the patch to make it complete would require review effort, and currently styles exist only for PaleMoon which are also not polished IMO as I didn't backport the tweaks I made in my production branch.

So, a lot of work for one theme, plus maintenance after merge. All this with the QML GUI in sight and existing, offcial customization available via Qt6Ctrl, makes me think we should spend our dev time for more important features and fixes.

The branch is still there, and I can provide the Scanner dialog commit, so if anyone wants to pick it up, it's there -- and will be for some time I guess.

PR #13843, opened 2024-11-06
chore: Fix pre-commit on F41 and update remaining dependencies

Swiftb0y:
Edit: this also updates the internal dependencies to be reasonably up-to-date enough to even successfully install. otherwise `lxml` doesn't compile on Fedora 41. Essentially without this PR, developing on Fedora is broken and thus we should merge this reasonably quickly. 

newer Pre-commit versions issue warnings about our config using outdated stage names. This fixes to be complaint to pre-commit v4.

> [WARNING] hook id `eslint` uses deprecated stage names (commit) which will be removed in a future version.  run: `pre-commit migrate-config` to automatically fix this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -73,7 +73,7 @@ repos:
         files: \.m?[jt]s$
         types: [file]
         stages:
-          - commit
+          - pre-commit
           - manual
         additional_dependencies:
           - eslint@^8.48.0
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
There is still one of these warnings left, but I don't see why.

Swiftb0y:
I think thats because we're currently targeting and old version of https://github.com/pre-commit/pre-commit-hooks that also does not yet have its config migrated. a `pre-commit autoupdate` would likely fix that. Should I include that here?

JoergAtGithub:
LGTM! Thank you!

JoergAtGithub:
I needed to call `pip install pre-commit --upgrade` to get pre-commit working again. As my installed version did not know the command `autoupdate` and failed therefore.

PR #13844, opened 2024-11-06
Fix talkover ducking bug

### LINKED ISSUE #12451: Can’t use microphone talkover mode with compressor effect

fonsargo:
This is a small fix for microphone talkover ducking bug. The idea is just to apply ducking after applying effects in order to avoid ducking neutralization by some of the effects (for example compressor).
Fixes #12451

### DIFF
--- a/src/engine/enginemixer.cpp
+++ b/src/engine/enginemixer.cpp
@@ -40,6 +40,7 @@ EngineMixer::EngineMixer(
           m_boothGainOld(0.0),
           m_headphoneMainGainOld(0.0),
           m_headphoneGainOld(1.0),
+          m_duckingGainOld(1.0),
           m_balleftOld(1.0),
           m_balrightOld(1.0),
           m_numMicsConfigured(0),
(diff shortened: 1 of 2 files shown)


fonsargo:
@daschuer Hi! Just wanted to leave a friendly reminder about my PR :)

fonsargo:
@daschuer is it possible to check this PR? It's been waiting for review since November :)

daschuer (approved):
LGTM, Thank you. Sorry for the delayed review.

PR #13847, opened 2024-11-07
(fix) skin reload: store new color scheme

ronso0:
It wasn't stored, so Apply/Okay did not reload the skin with the new scheme.

I thought we already fixed that earlier, but it simply slipped through in #11853


Swiftb0y (approved):
LGTM. Thank you.

PR #13851, opened 2024-11-08
Update ic_library_drag_and_drop.svg

### LINKED ISSUE #13836: Drag n drop from file manager to deck doesn't show drag cursor like when dragging from deck to deck
### Bug Description

Resulting in me not understanding if I've dragged the file to an appropriate place to drop it

### Version

2.5

### OS

_No response_

OfficialMunu:
[BUG] Resolves
#13836 

setAcceptDrops(true): This enables the widget to accept drag and drop events.
dragEnterEvent: This event is triggered when a drag operation enters the widget. If the dragged data (like a file URL) is valid, we accept the drop and set the cursor to Qt::CopyCursor, which indicates that the object can be copied (or dropped).
dragMoveEvent: This event happens as the dragged object moves inside the widget area. It’s important to update the cursor here as well.
dragLeaveEvent: When the dragged object leaves the widget area without dropping, you should revert the cursor back to its original state.
dropEvent: Handles the final drop operation, where you process the dragged data (e.g., load the file onto the deck).

### DIFF
--- a/res/images/library/ic_library_drag_and_drop.svg
+++ b/res/images/library/ic_library_drag_and_drop.svg
@@ -1,238 +1,45 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<svg
-   xmlns:osb="http://www.openswatchbook.org/uri/2009/osb"
-   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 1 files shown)


Swiftb0y:
you copy-pasted example code into a vector graphics file?

PR #13815, opened 2024-10-30
Remove tap beat mapping for DDJ-400 and DDJ-FLX4

Aquassaut:
The key combination was used for both toggle channel quantize and ajust channel BPM to match tapped BPM. After discussion with @jusko, beat tap being less useful, it's the one that should go.

This fixes #13813 

No documentation change is needed as the toggle quantize behavior that remains after this PR is the one that is already documented [here](https://github.com/mixxxdj/manual/blob/2.4/source/hardware/controllers/pioneer_ddj_400.rst?plain=1#L273) and [here](https://github.com/mixxxdj/manual/blob/2.4/source/hardware/controllers/pioneer_ddj_flx4.rst?plain=1#L264)

Issue #13813, opened 2024-10-29
Pioneer FLX4 mapping dupicate use of shift+ channel cue


Aquassaut:
### Bug Description

Hi,
First off, thank you for this amazing piece of software and all the good work.

Pressing shift + channel cue on the flx4 triggers both [channel beat tap](https://github.com/mixxxdj/mixxx/blob/main/res/controllers/Pioneer-DDJ-FLX4.midi.xml#L903) and [channel quantize toggle](https://github.com/mixxxdj/mixxx/blob/main/res/controllers/Pioneer-DDJ-FLX4.midi.xml#L697), which is not super great as it makes it easy to mess up the bpm of a song.
I can't verify this as I don't have the device handy but seeing the config file the same issue seems to be present in the very similar DDJ-400.

I think quantize toggle is more useful but beat tap is the intended use of shift + channel cue [according to the manual](https://docs.pioneerdj.com/manuals/DDJ_FLX4_DRI1804A_manual/?page=26).

This issue doesn't block me (I simply use an edited mapping for myself) but it might be confusing for new users.

ronso0:
Thanks for your report!

Do you have interest to provide a PR with a fix?

Aquassaut:
Sure can, however I'd like some guidance, maybe by the original mapping authors, as to which effect to keep.

Aquassaut:
It seems that @jusko added the quantize toggle behavior under the impression that the mapping wasn't used [in this commit](https://github.com/mixxxdj/mixxx/commit/3120764eb5e3b4151601e79408d1a3524ab5306b) so he might have some useful insight into it

jusko:
Hi @Aquassaut, good spot on this. It's been a _long_ time since that commit... but can say for sure that beat tap was never something I used so it was probably an oversight. Weird that the bug never got picked up until now. I guess quantize toggle is rarely used and even less often "tapped".

I would think to just remove beat tap as it's not really a useful feature. The mixxx hardware manual documents for quantize toggle, which as you say is more useful.

ronso0:
Note that in Mixxx 2.5 we added `tempo_tap` which allows manual syncing BPM to external sources.

PR #13857, opened 2024-11-08
Fix shift + play midi data for ddj-flx4 controller mapping

Aquassaut:
Hi,

After the issue #13813 and the resolving PR #13815, I found another controller mapping issue for the DDJ-flx4, whereby the mapping expects midino 0x47 for shift + play, while the device sends 0x0E.

I think this stems from the fact that the very similar DDJ-400, from which the mapping was modified to create the flx4 mapping, does send 0x47. I think it fell through the cracks because the feature doesn't seem to be used all that often, I myself discovered the issue because I was trying to remap it to another behavior.

For reference, [here's the DDJ-400 midi messages document](https://www.pioneerdj.com/-/media/pioneerdj/software-info/controller/ddj-400/ddj-400_midi_message_list_e1.pdf) and [here the DDJ-FLX4 midi messages document](https://www.pioneerdj.com/-/media/pioneerdj/software-info/controller/ddj-flx4/ddj-flx4_midi_message_list_e1.pdf). More conveniently here are the screenshots outlining the difference:

DDJ-FLX4
ddj-flx4-shift-play
DDJ-400
ddj-400-shift-play

Since last time, you guys asked me to base the PR off of branch 2.4, I'm doing the same thing here, let me know if I should rebase it off of another one or if anything else should be done.

### DIFF
--- a/res/controllers/Pioneer-DDJ-FLX4.midi.xml
+++ b/res/controllers/Pioneer-DDJ-FLX4.midi.xml
@@ -118,7 +118,7 @@
                 <group>[Channel1]</group>
                 <key>reverseroll</key>
                 <status>0x90</status>
-                <midino>0x47</midino>
+                <midino>0x0E</midino>
                 <options>
                     <Normal/>
                 </options>
@@ -138,7 +138,7 @@
                 <group>[Channel2]</group>
                 <key>reverseroll</key>
                 <status>0x91</status>
-                <midino>0x47</midino>
+                <midino>0x0E</midino>
                 <options>
                     <Normal/>
                 </options>


Swiftb0y (approved):
Thank you very much for catching this. You're totally right with your conclusion that this was copied from the DDJ-400 and slipped through the cracks. LGTM.

Swiftb0y:
fyi, in the future I recommend you create a separate [feature branch](https://www.atlassian.com/git/tutorials/comparing-workflows/feature-branch-workflow). Without you're bound to run into trouble if you want to sync your own 2.4 branch with our upstream one.

Aquassaut:
Thanks I'll make sure I do for anything more consequent. In this case I was not really that worried about the hard reset/force push on my end, but you're right.

PR #13872, opened 2024-11-10
tsan fix browsethread

### LINKED ISSUE #13861: thread sanitizer issue 01
### Bug Description

[tsan01.txt](https://github.com/user-attachments/files/17691451/tsan01.txt)

### Version

2.5

### OS

_No response_

m0dB:
Fixes #13861 and improves the code quality of BrowseThread a bit.

### DIFF
--- a/src/library/browse/browsethread.cpp
+++ b/src/library/browse/browsethread.cpp
@@ -14,80 +14,109 @@ namespace {
 constexpr int kRowBatchSize = 10;
 } // namespace
 
-QWeakPointer<BrowseThread> BrowseThread::m_weakInstanceRef;
-static QMutex s_Mutex;
-
 /*
  * This class is a singleton and represents a thread
  * that is used to read ID3 metadata
  * from a particular folder.
  *
  * The BrowseTableModel uses this class.
- * Note: Don't call getInstance() from places
+ * Note: Don't call getInstanceRef() from places
  * other than the GUI thread. BrowseThreads emit
  * signals to BrowseModel objects. It does not
  * make sense to use this class in non-GUI threads
  */
-BrowseThread::BrowseThread(QObject *parent)
-        : QThread(parent) {
-    m_bStopThread = false;
-    m_model_observer = nullptr;
-    //start Thread
+BrowseThread::BrowseThread(QObject* parent)
+        : QThread{parent}, m_model_observer{}, m_bRun{} {
+    // Start thread
     start(QThread::LowPriority);
-
+    qDebug() << "Wait to start browser background thread";
+    // Wait until the thread is running
+    m_mutex.lock();
+    while (!m_bRun) {
+        m_condition.wait(&m_mutex);
+    }
+    m_mutex.unlock();
+    qDebug() << "Browser background thread started";
 }
 
 BrowseThread::~BrowseThread() {
-    qDebug() << "Wait to finish browser background thread";
-    m_bStopThread = true;
-    //wake up thread since it might wait for user input
-    m_locationUpdated.wakeAll();
-    //Wait until thread terminated
-    //terminate();
+    qDebug() << "Wait to terminate browser background thread";
+    // Inform the thread we want it to terminate
+    m_mutex.lock();
+    m_bRun = false;
+    m_condition.wakeOne();
+    m_mutex.unlock();
+    // Wait until thread terminated
     wait();
-    qDebug() << "Browser background thread terminated!";
+    qDebug() << "Browser background thread terminated";
 }
 
 // static
 BrowseThreadPointer BrowseThread::getInstanceRef() {
-    BrowseThreadPointer strong = m_weakInstanceRef.toStrongRef();
+    // We create a single BrowseThread instead which is used by multiple
+    // BrowseTableModel instances. We store a weakpointer so when all
+    // BrowseTableModel have been destroyed, so it the BrowseThread.
+
+    static QWeakPointer<BrowseThread> s_weakInstanceRef;
+    static QMutex s_mutex;
+
+    s_mutex.lock();
+    BrowseThreadPointer strong = s_weakInstanceRef.toStrongRef();
     if (!strong) {
-        s_Mutex.lock();
-        strong = m_weakInstanceRef.toStrongRef();
-        if (!strong) {
-            strong = BrowseThreadPointer(new BrowseThread());
-            m_weakInstanceRef = strong.toWeakRef();
-        }
-        s_Mutex.unlock();
+        strong = BrowseThreadPointer(new BrowseThread());
+        s_weakInstanceRef = strong.toWeakRef();
     }
+    s_mutex.unlock();
+
     return strong;
 }
 
-void BrowseThread::executePopulation(mixxx::FileAccess path, BrowseTableModel* client) {
-    m_path_mutex.lock();
+void BrowseThread::requestPopulateModel(mixxx::FileAccess path, BrowseTableModel* client) {
+    // Inform the thread to populate a new path.
+    // Note: if another request is currently being processed, this will replace it.
+
+    qDebug() << "Request populate model" << path.info() << client;
+    m_mutex.lock();
     m_path = std::move(path);
     m_model_observer = client;
-    m_path_mutex.unlock();
-    m_locationUpdated.wakeAll();
+    m_condition.wakeOne();
+    m_mutex.unlock();
 }
 
 void BrowseThread::run() {
     QThread::currentThread()->setObjectName("BrowseThread");
-    m_mutex.lock();
 
-    while (!m_bStopThread) {
-        //Wait until the user has selected a folder
-        m_locationUpdated.wait(&m_mutex);
+    // Inform the constructor the thread started
+    m_mutex.lock();
+    m_bRun = true;
+    m_condition.wakeOne();
+    m_mutex.unlock();
+
+    while (true) {
+        // Wait for a new population request, or for a termination request
+        qDebug() << "Wait for a new population request";
+        m_mutex.lock();
+        while (!m_path.isSet() && m_bRun) {
+            m_condition.wait(&m_mutex);
+        }
+        auto path = std::move(m_path);
+        auto modelObserver = m_model_observer;
+        auto bRun = m_bRun;
+        m_path = mixxx::FileAccess();
+        m_mutex.unlock();
+
         Trace trace("BrowseThread");
 
-        //Terminate thread if Mixxx closes
-        if(m_bStopThread) {
-            break;
+        if (!bRun) {
+            qDebug() << "Termination request";
+            // Terminate thread if Mixxx closes
+            return;
+        } else {
+            qDebug() << "New population request";
+            // Populate the model
+            populateModel(path, modelObserver);
         }
-        // Populate the model
-        populateModel();
     }
-    m_mutex.unlock();
 }
 
 namespace {
@@ -112,29 +141,27 @@ public:
 
 } // namespace
 
-void BrowseThread::populateModel() {
-    m_path_mutex.lock();
-    auto thisPath = m_path;
-    BrowseTableModel* thisModelObserver = m_model_observer;
-    m_path_mutex.unlock();
+void BrowseThread::populateModel(mixxx::FileAccess path, BrowseTableModel* modelObserver) {
+    // Executed by the thread run for incoming populate model requests
 
-    if (!thisPath.info().hasLocation()) {
+    if (!path.info().hasLocation()) {
         // Abort if the location is inaccessible or does not exist
-        qWarning() << "Skipping" << thisPath.info();
+        qWarning() << "Skipping" << path.info();
         return;
     }
+    qDebug() << "populateModel" << path.info() << modelObserver;
 
     // Refresh the name filters in case we loaded new SoundSource plugins.
     const QStringList nameFilters = SoundSourceProxy::getSupportedFileNamePatterns();
 
-    QDirIterator fileIt(thisPath.info().location(),
+    QDirIterator fileIt(path.info().location(),
             nameFilters,
             QDir::Files | QDir::NoDotAndDotDot);
 
     // remove all rows
     // This is a blocking operation
     // see signal/slot connection in BrowseTableModel
-    emit clearModel(thisModelObserver);
+    emit clearModel(modelObserver);
 
     QList<QList<QStandardItem*>> rows;
     rows.reserve(kRowBatchSize);
@@ -144,15 +171,19 @@ void BrowseThread::populateModel() {
     int row = 0;
     // Iterate over the files
     while (fileIt.hasNext()) {
-        // If a user quickly jumps through the folders
-        // the current task becomes "dirty"
-        m_path_mutex.lock();
-        auto newPath = m_path;
-        m_path_mutex.unlock();
-
-        if (thisPath.info() != newPath.info()) {
-            qDebug() << "Abort populateModel()";
-            populateModel();
+        // If while processing a new population request or a termination request
+        // was received, we abort the current request. This happens if a user
+        // quickly jumps through the folders and the current task becomes
+        // "dirty".
+        m_mutex.lock();
+        const bool abortProcess = !m_bRun ||
+                (m_path.isSet() &&
+                        (m_path.info() != path.info() ||
+                                modelObserver != m_modelObserver));
+        m_mutex.unlock();
+        if (abortProcess) {
+            qDebug() << "Abort populateModel";
+            // The run function will take care of the new request.
             return;
         }
 
@@ -162,7 +193,7 @@ void BrowseThread::populateModel() {
 
         const auto fileAccess = mixxx::FileAccess(
                 mixxx::FileInfo(fileIt.next()),
-                thisPath.token());
+                path.token());
         {
             mixxx::TrackMetadata trackMetadata;
             // Both resetMissingTagMetadata = false/true have the same effect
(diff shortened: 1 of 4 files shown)


src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
-    m_bStopThread = false;
-    m_model_observer = nullptr;
-    //start Thread
+BrowseThread::BrowseThread(QObject* parent)
+        : QThread{parent}, m_model_observer{}, m_bRun{} {
+    // Start thread
     start(QThread::LowPriority);
-
+    qDebug() << "Wait to start browser background thread";
+    // Wait until the thread is running
```
m0dB, line 33: this fixes the data race

src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
+    s_mutex.unlock();
+
     return strong;
 }
 
-void BrowseThread::executePopulation(mixxx::FileAccess path, BrowseTableModel* client) {
-    m_path_mutex.lock();
+void BrowseThread::requestPopulateModel(mixxx::FileAccess path, BrowseTableModel* client) {
+    // Inform the thread to populate a new path.
+    // Note: if another request is currently being processed, this will replace it.
```
m0dB, line 76: and that seems weird to me. it makes sense when the client is the same, but what if it's a different client? anyway, I didn't want to change the behaviour.

src/library/browse/browsethread.cpp
```diff
@@ -144,15 +171,19 @@ void BrowseThread::populateModel() {
...
     while (fileIt.hasNext()) {
-        // If a user quickly jumps through the folders
-        // the current task becomes "dirty"
-        m_path_mutex.lock();
-        auto newPath = m_path;
-        m_path_mutex.unlock();
-
-        if (thisPath.info() != newPath.info()) {
-            qDebug() << "Abort populateModel()";
-            populateModel();
```
m0dB, line 155: I didn't like this recursion, better give the control back to the run() method to handle the new request.

src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
  * signals to BrowseModel objects. It does not
  * make sense to use this class in non-GUI threads
  */
-BrowseThread::BrowseThread(QObject *parent)
-        : QThread(parent) {
-    m_bStopThread = false;
-    m_model_observer = nullptr;
-    //start Thread
+BrowseThread::BrowseThread(QObject* parent)
+        : QThread{parent}, m_modelObserver{}, m_bRun{} {
```
daschuer, line 29: Not your fault: We use a p prefix for pointers and normally stack the initializer list. Would be nice if you adjust this as well.

daschuer (commented):
Some comments. I have not yet fully understand the issue.

src/library/browse/browsethread.cpp
m0dB: will do

m0dB:
> Some comments. I have not yet fully understand the issue.

To explain the issue: tsan reported a data race in browsetablemodel.cpp:163 when connecting the table model with the result from BrowseThread::getInstanceRef(). IIRC the fix is waiting until the thread is started before returning from getInstanceRef (in the BrowseThread constructor) but I also changed getInstanceRef itself because the placing of the mutex didn’t seem correct to me. 

The other changes are mainly esthetics / code clarity.

I do have my doubts about this code, it seems strange to me that a new populate request for any browsetablemodel aborts the running request. But I didn’t want to change to behavior.

src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
+    // Start thread
     start(QThread::LowPriority);
-
+    qDebug() << "Wait to start browser background thread";
+    // Wait until the thread is running
+    m_mutex.lock();
+    while (!m_bRun) {
+        m_condition.wait(&m_mutex);
+    }
```
daschuer, line 37: I have not understood the loop. Isn't there only one event/codition we are waiting for? m_bRun shall become a semaphore? On the other hand syncronizing the thread start by halting the main thread feels wrong. It would be better to protect the shared sources explicit. What are the shared resources in this case we have tsan complains?

src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
+
+    while (true) {
+        // Wait for a new population request, or for a termination request
+        qDebug() << "Wait for a new population request";
+        m_mutex.lock();
+        while (!m_path.isSet() && m_bRun) {
+            m_condition.wait(&m_mutex);
+        }
+        auto path = std::move(m_path);
+        auto modelObserver = m_modelObserver;
```
daschuer, line 103: Do we have a gurantee that the m_modelObserver outlives the thread?

src/library/browse/browsethread.cpp
```diff
@@ -112,29 +141,27 @@ class YearItem: public QStandardItem {
 
 } // namespace
 
-void BrowseThread::populateModel() {
-    m_path_mutex.lock();
-    auto thisPath = m_path;
-    BrowseTableModel* thisModelObserver = m_model_observer;
-    m_path_mutex.unlock();
+void BrowseThread::populateModel(mixxx::FileAccess path, BrowseTableModel* modelObserver) {
```
daschuer, line 144: p prefix would be nice: pModelObserver

src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
+    m_mutex.unlock();
 }
 
 void BrowseThread::run() {
     QThread::currentThread()->setObjectName("BrowseThread");
+
+    // Inform the constructor the thread started
     m_mutex.lock();
+    m_bRun = true;
+    m_condition.wakeOne();
```
daschuer, line 92: wakeOne() and wait() below is confusing, Let's use a second sync method.

src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
+    // Note: if another request is currently being processed, this will replace it.
+
+    qDebug() << "Request populate model" << path.info() << modelObserver;
+    m_mutex.lock();
     m_path = std::move(path);
-    m_model_observer = client;
-    m_path_mutex.unlock();
-    m_locationUpdated.wakeAll();
+    m_modelObserver = modelObserver;
+    m_condition.wakeOne();
```
daschuer, line 82: What can happen if the thread is not yet waiting? In this case no one is waked up. 
But the `m_modelObserver` and `m_path` have a value. I think the thread can first check if m_modelObserver is not null and pick that up before waiting. Probably the waiting in the constructor is obsolete than?

daschuer (changes requested):

src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
 BrowseThreadPointer BrowseThread::getInstanceRef() {
-    BrowseThreadPointer strong = m_weakInstanceRef.toStrongRef();
+    // We create a single BrowseThread instead which is used by multiple
+    // BrowseTableModel instances. We store a weakpointer so when all
+    // BrowseTableModel have been destroyed, so it the BrowseThread.
+
+    static QWeakPointer<BrowseThread> s_weakInstanceRef;
+    static QMutex s_mutex;
+
+    s_mutex.lock();
```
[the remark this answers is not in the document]
m0dB: Ok, I will check again where the issue comes from, here or the thread being used before being started.

src/library/browse/browsethread.cpp
```diff
@@ -14,80 +14,109 @@ namespace {
...
+        s_weakInstanceRef = strong.toWeakRef();
     }
+    s_mutex.unlock();
+
     return strong;
 }
 
-void BrowseThread::executePopulation(mixxx::FileAccess path, BrowseTableModel* client) {
-    m_path_mutex.lock();
+void BrowseThread::requestPopulateModel(mixxx::FileAccess path, BrowseTableModel* modelObserver) {
```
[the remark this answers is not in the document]
m0dB: will do

src/library/browse/browsethread.h
```diff
@@ -41,16 +41,13 @@ class BrowseThread : public QThread {
   private:
     BrowseThread(QObject *parent = 0);
 
-    void populateModel();
+    void populateModel(mixxx::FileAccess path, BrowseTableModel* modelObserver);
 
     QMutex m_mutex;
-    QWaitCondition m_locationUpdated;
-    volatile bool m_bStopThread;
+    QWaitCondition m_condition;
```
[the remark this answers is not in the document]
m0dB: the thing is, I am using it for 2 things: waiting for the thread to be started and waiting for a new populate request.

src/library/browse/browsethread.cpp
m0dB: if m_path has a value, the thread will not enter the wait state.

src/library/browse/browsethread.cpp
m0dB: yes, because it's holds a shared pointer to it.

src/library/browse/browsethread.cpp
m0dB: Yes, we are waiting for m_bRun to be set to true, as set by the thread::run method. A semaphore would work as well, but since we already have the condition, I take advantage of it.

daschuer: > A semaphore would work as well, but since we already have the condition, I take advantage of it.

This reusing makees the code harder to read. So I would prefer two primitived for two tasks. 

The ThreadSanitizer warning is really confusing, what does it check for? Whether a variable is written/read from two threads? It must be a control structure in Qt. However waiting for the thread to start should not change the issue for my undertsanding. 

But if it is the solution i don't mind.

src/library/browse/browsethread.cpp
daschuer: m_modelObserver holds a share pointer to the Browsthread? Maybe you can add a comment to the code?

src/library/browse/browsethread.cpp
daschuer: How does the solution work? I am afraid instead of a solution, we trick ThreadSanitizer.
Can, it happen that it treated the unrelated mutex as a proper guard?
Qt allows to send signals and reconnect them concurrently. So it should not be a problem. 
Maybe only a missing atomic annotation in Qt? 

Is there a way to find out what 0x0001466d5f28 is and which mutexted I are considered? 

Or is it s QT bug? Is 2.4 with Qt 5 affected as well.

daschuer: Got it, I n this thread they discussed exactly our problem. Solution is to build Qt wir sanitizer flag: 
https://lists.qt-project.org/pipermail/interest/2022-November/038561.html

Conclusion here: We can safely ignore the issue.

src/library/browse/browsethread.cpp
m0dB: Confirmed. This is the fix for the data race. Connecting the signals/slots while the thread is not yet running is the issue.

src/library/browse/browsethread.cpp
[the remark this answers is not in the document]
m0dB: You are right. The original code was correct. I have reverted it, except for making 
```
   static QWeakPointer<BrowseThread> s_weakInstanceRef;
   static QMutex s_mutex;
```
local variables in the function instead of globals.

Please resolve if you are okay with the latest version.

src/library/browse/browsethread.h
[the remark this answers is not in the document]
m0dB: I have renamed the member vars for clarity. Please review and resolve.

src/library/browse/browsethread.cpp
m0dB: I have build Qt with the sanitizer flag, though I used the normal way (passing the -fsanitize=thread flag), so unless this configure option does something else (I doubt it), the problem is still there.

I don't see you problem with using one and the same mutex and condition.

src/library/browse/browsethread.cpp
m0dB: There is a single, refcounted, BrowseThread. This is what is returned by BrowseThread::getInstanceRef();

In BrowseTableModel.cpp it is stored: m_pBrowseThread = BrowseThread::getInstanceRef();

So, when all BrowseTableModel instances are gone, the ref count is 0, and the BrowseThread instance is destroyed.

But I am not sure why I have to explain this or add more comments. This is not my code! I only want to solve the thread sanitizer warnings. If you feel the (existing!) code needs more explanation, go ahead.

src/library/browse/browsethread.cpp
m0dB: I put the request in a separate method.

src/library/browse/browsethread.cpp
m0dB: Anyway, I will see if I can build qt in my vcpkg with this configure option, and if that makes a difference. But I would appreciate it if you can point me to where I need to change this, given my inexperience with vcpkg.

src/library/browse/browsethread.cpp
m0dB: That makes sense.

src/library/browse/browsethread.cpp
daschuer: I think you can add it here: 
https://github.com/mixxxdj/vcpkg/blob/74ff6c7ceb657e9f099626b5834884184209bf50/ports/qt5-base/portfile.cmake#L94

m0dB: Ok, I will give that a try.

m0dB: I am building against Qt6 though

m0dB: which seems to use a different approach... Any idea how to do it there?

m0dB: Anyway, I am now building Qt5. But it is on purpose that the 2.5 branch of mixxx vcpkg has Qt6 as the default? I am confused now as to what we are supposed to use for Mixxx 2.5.

m0dB: I am running into lots of problems trying to build against Qt5, so it would be much appreciated if you could help me building Qt6 with vcpkg with this Qt sanitize thread option.

daschuer: Mixxx 2.4.2 will be our last Qt5 release. So there is IMHO no need to deal with QT5 in this case. We just keep the Qt5 support in our 2.5 branch just in case we want to test regresions. 

I suggest to issue a PR against our vcpkg branch and use the build package downloaded from GitHub. For my undersanding you jsut need to add the sanitizer flag at the position I pointed out above.

m0dB: Yes, I prefer to use Qt 6, but you pointed it out in the qt5 portfile.cmake. The qt6 portfile.cmake is completely different, and I don't see where to add this option.

m0dB: It looks like this works. I'd prefer to set this from my triplets file arm64-osx-tsan.cmake, but for now this have to do.

diff --git a/ports/qtbase/portfile.cmake b/ports/qtbase/portfile.cmake
index 8b9e3b0d3..4a96412d7 100644
--- a/ports/qtbase/portfile.cmake
+++ b/ports/qtbase/portfile.cmake
@@ -92,7 +92,7 @@ INVERTED_FEATURES
     )
 
 list(APPEND FEATURE_OPTIONS -DCMAKE_DISABLE_FIND_PACKAGE_Libudev:BOOL=ON)
-list(APPEND FEATURE_OPTIONS -DFEATURE_xml:BOOL=ON)
+list(APPEND FEATURE_OPTIONS -DFEATURE_xml:BOOL=ON -DFEATURE_sanitize_thread:BOOL=ON)

m0dB:
I have checked with Qt 6 with thread sanitizer activated through the Qt feature flags, and it still has to same problem, so I would like to ask to merge this PR. I feel I addressed all comments.

m0dB:
Is there is a reason not to merge this? I confirm that, including with a thread sanitizer enabled Qt, I get TSAN warnings unless I wait for the thread to have started.

daschuer:
I did a bit of a research and discovered a related QT issues. 

https://bugreports.qt.io/browse/QTBUG-100336
.. looks quite similar. My guess is that the different backtrace happens due to not optimized code. 

Can you please use the attached the testing code and confirm that this is actually our issue? 

This bug is fixed in: Qt v6.8.0 and planned to 6.5 the stable branch we use for Mixxx 2.5 
https://github.com/qt/qtbase/commit/75d82afa0d3aad9b4f9857e439535fc49c4616bc

I was able to trace it down to https://github.com/qt/qtbase/commit/34fe9232dbf6a9fe58ebc4c7680bb67d2f642c40
> receiver->d_func()->connections.loadRelaxed()->currentSender = previous
Where it has been portet from QAtomic in v5.14.0
But the issue exists way longer with the old API. 

The issue seems affect a bunch of atomic variables and whenever a connection is astabisched between threads. Most likely this fix fixes only one symptome of the issue and the issue will happen again with the other signals. 
Luckily this is currently "only" a linter issue and we do not have a real user issue with that, I would prefer to revert the quite bulky and run-time consuming solution here and just wait for the upstream fix. The other refactorings in this thread are welcome. 

An alternative solution for all signals would be to connect the the connections via "m_model_observer" from the worker thread itself, this way the affected variables are only touched by one thread. I consider this a nice refactoring that can remain even if the upstrem issue is fixed. But I don't want to put that work on you. You may also directky "QMetaObject::invokeMethod" without connecting.

Issue #13868, opened 2024-11-10
thread sanitizer issue 07


m0dB:
### Bug Description

PaUtil_WriteRingBuffer / PaUtil_ReadRingBuffer (duplicate of #13863?)

<details><summary>Details</summary>
<p>

```cpp


==================
WARNING: ThreadSanitizer: data race (pid=48840)
  Write of size 4 at 0x000111b25c44 by thread T71 (mutexes: write M0, write M1, write M2):
    #0 PaUtil_WriteRingBuffer <null>:230508612 (mixxx:arm64+0x102110a24)
    #1 FIFO<CachingReaderChunkReadRequest>::write(CachingReaderChunkReadRequest const*, int) fifo.h:34 (mixxx:arm64+0x1008d2c9c)
    #2 CachingReader::hintAndMaybeWake(QVarLengthArray<Hint, 512ll> const&) cachingreader.cpp:582 (mixxx:arm64+0x1008d2970)
    #3 EngineBuffer::hintReader(double) enginebuffer.cpp:1518 (mixxx:arm64+0x10099410c)
    #4 EngineBuffer::processTrackLocked(float*, int, mixxx::audio::SampleRate) enginebuffer.cpp:1151 (mixxx:arm64+0x100992d74)
    #5 EngineBuffer::process(float*, int) enginebuffer.cpp:1183 (mixxx:arm64+0x1009943e4)
    #6 EngineDeck::process(float*, int) enginedeck.cpp:64 (mixxx:arm64+0x1008f5230)
    #7 EngineMixer::processChannels(int) enginemixer.cpp:367 (mixxx:arm64+0x1009a7dc8)
    #8 EngineMixer::process(int) enginemixer.cpp:426 (mixxx:arm64+0x1009a8560)
    #9 SoundManager::onDeviceOutputCallback(long) soundmanager.cpp:599 (mixxx:arm64+0x10117ac44)
    #10 SoundDevicePortAudio::callbackProcessClkRef(long, float*, float const*, PaStreamCallbackTimeInfo const*, unsigned long) sounddeviceportaudio.cpp:998 (mixxx:arm64+0x101172984)
    #11 (anonymous namespace)::paV19CallbackClkRef(void const*, void*, unsigned long, PaStreamCallbackTimeInfo const*, unsigned long, void*) sounddeviceportaudio.cpp:69 (mixxx:arm64+0x10116fe5c)
[...]

Issue #13863, opened 2024-11-10
thread sanitizer issue 03


m0dB:
### Bug Description

[tsan03.txt](https://github.com/user-attachments/files/17691457/tsan03.txt)

### Version

2.5

daschuer:
These findings are probably false positive, because TSAN does not understand PaUtil_FullMemoryBarrier(). However it is poosible to anotate the code with `__tsan_acquire`, `__tsan_release` and firends to make it work. 

Do you have interest to try it out? If it turns out that these are real issues. Mixxx will suffer them via Portausdio as well.

PR #13877, opened 2024-11-10
Reimplement fifo as a proper lock free SCSP ring buffer 

### LINKED ISSUE #13864: thread sanitizer issue 04
### Bug Description

PaUtil_WriteRingBuffer / PaUtil_ReadRingBuffer (duplicate of #13863?)

<details><summary>Details</summary>
<p>

```cpp

==================
WARNING: ThreadSanitizer: data race (pid=48840)
  Write of size 4 at 0x000112915ef4 by thread T71 (mutexes: write M0, write M1):
    #0 PaUtil_WriteRingBuffer <null>:230508612 (mixxx:arm64+0x102110a24)
    #1 FIFO<float>::write(float const*, int) fifo.h:34 (mixxx:arm64+0x1009dd6dc)
    #2 EngineSideChain::writeSamples(float const*, int) enginesidechain.cpp:81 (mixxx:arm64+0x1009dd440)
    #3 EngineMixer::process(int) enginemixer.cpp:757 (mixxx:arm64+0x1009a9740)
    #4 SoundManager::onDeviceOutputCallback(long) soundmanager.cpp:599 (mixxx:arm64+0x10117ac44)
    #5 SoundDevicePortAudio::callbackProcessClkRef(long, float*, float const*, PaStreamCallbackTimeInfo const*, unsigned long) sounddeviceportaudio.cpp:998 (mixxx:arm64+0x101172984)
    #6 (anonymous namespace)::paV19CallbackClkRef(void const*, void*, unsigned long, PaStreamCallbackTimeInfo const*, unsigned long, void*) sounddeviceportaudio.cpp:69 (mixxx:arm64+0x10116fe5c)
    #7 AdaptingOutputOnlyProcess <null>:230508612 (mixxx:arm64+0x10210fed4)
    #8 PaUtil_EndBufferProcessing <null>:230508612 (mixxx:arm64+0x10210e09c)
    #9 AudioIOProc <null>:230508612 (mixxx:arm64+0x1021185a8)
    #10 <null> <null>:230508612 (CoreAudio:arm64e+0xa8d8)

  Previous read of size 4 at 0x000112915ef4 by thread T16:
    #0 PaUtil_ReadRingBuffer <null>:230508612 (mixxx:arm64+0x102110a88)
    #1 FIFO<float>::read(float*, int) fifo.h:31 (mixxx:arm64+0x1009d2580)
    #2 EngineSideChain::run() enginesidechain.cpp:111 (mixxx:arm64+0x1009dd9a8)
    #3 QThreadPrivate::start(void*) <null>:230508612 (mixxx:arm64+0x100248f60)

  Location is heap block of size 176 at 0x000112915ea0 allocated by main thread:
    #0 operator new(unsigned long) <null>:230508612 (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x83de0)
    #1 EngineMixer::EngineMixer(QSharedPointer<ConfigObject<ConfigValue>>, QString const&, EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>, bool) enginemixer.cpp:168 (mixxx:arm64+0x1009a45bc)
    #2 EngineMixer::EngineMixer(QSharedPointer<ConfigObject<ConfigValue>>, QString const&, EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>, bool) enginemixer.cpp:52 (mixxx:arm64+0x1009a0018)
    #3 EngineMixer* std::__1::construct_at[abi:v160006]<EngineMixer, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool, EngineMixer*>(EngineMixer*, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) construct_at.h:38 (mixxx:arm64+0x100647ae0)
    #4 void std::__1::allocator_traits<std::__1::allocator<EngineMixer>>::construct[abi:v160006]<EngineMixer, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool, void, void>(std::__1::allocator<EngineMixer>&, EngineMixer*, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) allocator_traits.h:304 (mixxx:arm64+0x100647754)
    #5 std::__1::__shared_ptr_emplace<EngineMixer, std::__1::allocator<EngineMixer>>::__shared_ptr_emplace[abi:v160006]<QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool>(std::__1::allocator<EngineMixer>, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) shared_ptr.h:284 (mixxx:arm64+0x1006475c0)
    #6 std::__1::__shared_ptr_emplace<EngineMixer, std::__1::allocator<EngineMixer>>::__shared_ptr_emplace[abi:v160006]<QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool>(std::__1::allocator<EngineMixer>, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) shared_ptr.h:276 (mixxx:arm64+0x100646fd8)
    #7 std::__1::shared_ptr<EngineMixer> std::__1::allocate_shared[abi:v160006]<EngineMixer, std::__1::allocator<EngineMixer>, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool, void>(std::__1::allocator<EngineMixer> const&, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) shared_ptr.h:995 (mixxx:arm64+0x100646dbc)
    #8 std::__1::shared_ptr<EngineMixer> std::__1::make_shared[abi:v160006]<EngineMixer, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool, void>(QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) shared_ptr.h:1004 (mixxx:arm64+0x100638fcc)
    #9 mixxx::CoreServices::initialize(QApplication*) coreservices.cpp:264 (mixxx:arm64+0x10063735c)
    #10 (anonymous namespace)::runMixxx(MixxxApplication*, CmdlineArgs const&) main.cpp:87 (mixxx:arm64+0x100010fcc)
    #11 main main.cpp:243 (mixxx:arm64+0x10000ff08)

  Mutex M0 (0x000132118e18) created at:
    #0 pthread_mutex_init <null>:230508612 (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x30e30)
    #1 HALB_Mutex::HALB_Mutex() <null>:230508612 (CoreAudio:arm64e+0x582f70)
    #2 Pa_Initialize <null>:230508612 (mixxx:arm64+0x10210b61c)
    #3 SoundManager::queryDevicesPortaudio() soundmanager.cpp:261 (mixxx:arm64+0x101176d00)
    #4 SoundManager::queryDevices() soundmanager.cpp:239 (mixxx:arm64+0x1011750c4)
    #5 SoundManager::SoundManager(QSharedPointer<ConfigObject<ConfigValue>>, EngineMixer*) soundmanager.cpp:77 (mixxx:arm64+0x101174b18)
    #6 SoundManager::SoundManager(QSharedPointer<ConfigObject<ConfigValue>>, EngineMixer*) soundmanager.cpp:59 (mixxx:arm64+0x101175324)
    #7 SoundManager* std::__1::construct_at[abi:v160006]<SoundManager, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*, SoundManager*>(SoundManager*, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*&&) construct_at.h:38 (mixxx:arm64+0x100648e68)
    #8 void std::__1::allocator_traits<std::__1::allocator<SoundManager>>::construct[abi:v160006]<SoundManager, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*, void, void>(std::__1::allocator<SoundManager>&, SoundManager*, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*&&) allocator_traits.h:304 (mixxx:arm64+0x100648b48)
    #9 std::__1::__shared_ptr_emplace<SoundManager, std::__1::allocator<SoundManager>>::__shared_ptr_emplace[abi:v160006]<QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*>(std::__1::allocator<SoundManager>, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*&&) shared_ptr.h:284 (mixxx:arm64+0x1006489e4)
    #10 std::__1::__shared_ptr_emplace<SoundManager, std::__1::allocator<SoundManager>>::__shared_ptr_emplace[abi:v160006]<QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*>(std::__1::allocator<SoundManager>, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*&&) shared_ptr.h:276 (mixxx:arm64+0x10064842c)
    #11 std::__1::shared_ptr<SoundManager> std::__1::allocate_shared[abi:v160006]<SoundManager, std::__1::allocator<SoundManager>, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*, void>(std::__1::allocator<SoundManager> const&, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*&&) shared_ptr.h:995 (mixxx:arm64+0x100648240)
    #12 std::__1::shared_ptr<SoundManager> std::__1::make_shared[abi:v160006]<SoundManager, QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*, void>(QSharedPointer<ConfigObject<ConfigValue>>&, EngineMixer*&&) shared_ptr.h:1004 (mixxx:arm64+0x100639118)
    #13 mixxx::CoreServices::initialize(QApplication*) coreservices.cpp:274 (mixxx:arm64+0x1006373ec)
    #14 (anonymous namespace)::runMixxx(MixxxApplication*, CmdlineArgs const&) main.cpp:87 (mixxx:arm64+0x100010fcc)
    #15 main main.cpp:243 (mixxx:arm64+0x10000ff08)

  Mutex M1 (0x000113b29b60) created at:
    #0 pthread_mutex_init <null>:230508612 (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x30e30)
    #1 CADeprecated::CAMutex::CAMutex(char const*) <null>:230508612 (libAudioToolboxUtility.dylib:arm64e+0x2920c)
    #2 OpenStream <null>:230508612 (mixxx:arm64+0x102115abc)
    #3 Pa_OpenStream <null>:230508612 (mixxx:arm64+0x10210c704)
    #4 SoundDevicePortAudio::open(bool, int) sounddeviceportaudio.cpp:338 (mixxx:arm64+0x10116f380)
    #5 SoundManager::setupDevices() soundmanager.cpp:475 (mixxx:arm64+0x101178660)
    #6 MixxxMainWindow::initialize() mixxxmainwindow.cpp:346 (mixxx:arm64+0x100f321e4)
    #7 QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (MixxxMainWindow::*)()>::call(void (MixxxMainWindow::*)(), MixxxMainWindow*, void**) qobjectdefs_impl.h:137 (mixxx:arm64+0x100f46a14)
    #8 void QtPrivate::FunctionPointer<void (MixxxMainWindow::*)()>::call<QtPrivate::List<>, void>(void (MixxxMainWindow::*)(), MixxxMainWindow*, void**) qobjectdefs_impl.h:174 (mixxx:arm64+0x100f468e0)
    #9 QtPrivate::QSlotObject<void (MixxxMainWindow::*)(), QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) qobjectdefs_impl.h:432 (mixxx:arm64+0x100f46744)
    #10 void doActivate<false>(QObject*, int, void**) <null>:230508612 (mixxx:arm64+0x10011ef08)
    #11 QMetaObject::activate(QObject*, QMetaObject const*, int, void**) <null>:230508612 (mixxx:arm64+0x10011db88)
    #12 WInitialGLWidget::onInitialized() moc_winitialglwidget.cpp:160 (mixxx:arm64+0x1015c2d78)
    #13 WInitialGLWidget::initializeGL() winitialglwidget.cpp:17 (mixxx:arm64+0x1015c3294)
    #14 OpenGLWindow::initializeGL() openglwindow.cpp:31 (mixxx:arm64+0x1015c00b8)
    #15 QOpenGLWindowPrivate::initialize() <null>:230508612 (mixxx:arm64+0x1044fcd88)
    #16 QOpenGLWindow::resizeEvent(QResizeEvent*) <null>:230508612 (mixxx:arm64+0x1044fdd78)
    #17 QWindow::event(QEvent*) <null>:230508612 (mixxx:arm64+0x104cdbd1c)
    #18 QPaintDeviceWindow::event(QEvent*) <null>:230508612 (mixxx:arm64+0x104ca4814)
    #19 OpenGLWindow::event(QEvent*) openglwindow.cpp:70 (mixxx:arm64+0x1015c03fc)
    #20 QApplicationPrivate::notify_helper(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fabdc)
    #21 QApplication::notify(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fc184)
    #22 MixxxApplication::notify(QObject*, QEvent*) mixxxapplication.cpp:200 (mixxx:arm64+0x100ef48dc)
    #23 QCoreApplication::notifyInternal2(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b3dec)
    #24 QCoreApplication::sendSpontaneousEvent(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b4f08)
    #25 QGuiApplicationPrivate::processGeometryChangeEvent(QWindowSystemInterfacePrivate::GeometryChangeEvent*) <null>:230508612 (mixxx:arm64+0x104c8a1b4)
    #26 QGuiApplicationPrivate::processWindowSystemEvent(QWindowSystemInterfacePrivate::WindowSystemEvent*) <null>:230508612 (mixxx:arm64+0x104c86310)
    #27 QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) <null>:230508612 (mixxx:arm64+0x104ce3884)
    #28 QWindowSystemInterface::flushWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) <null>:230508612 (mixxx:arm64+0x104ce3304)
    #29 QCocoaWindow::handleGeometryChange() <null>:230508612 (mixxx:arm64+0x101c7da5c)
    #30 QWindow::show() <null>:230508612 (mixxx:arm64+0x104cd2c24)
    #31 QWindowContainer::event(QEvent*) <null>:230508612 (mixxx:arm64+0x102299710)
    #32 QApplicationPrivate::notify_helper(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fabdc)
    #33 QApplication::notify(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fd364)
    #34 MixxxApplication::notify(QObject*, QEvent*) mixxxapplication.cpp:200 (mixxx:arm64+0x100ef48dc)
    #35 QCoreApplication::notifyInternal2(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b3dec)
    #36 QCoreApplication::sendEvent(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b4ea8)
    #37 QWidgetPrivate::show_helper() <null>:230508612 (mixxx:arm64+0x1022741b0)
    #38 QWidgetPrivate::setVisible(bool) <null>:230508612 (mixxx:arm64+0x102275c10)
    #39 QWidget::setVisible(bool) <null>:230508612 (mixxx:arm64+0x1022753f0)
    #40 QWidget::show() <null>:230508612 (mixxx:arm64+0x102273fc8)
    #41 WGLWidget::showEvent(QShowEvent*) wglwidgetqopengl.cpp:44 (mixxx:arm64+0x1015c2698)
    #42 QWidget::event(QEvent*) <null>:230508612 (mixxx:arm64+0x102276240)
    #43 QApplicationPrivate::notify_helper(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fabdc)
    #44 QApplication::notify(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fd364)
    #45 MixxxApplication::notify(QObject*, QEvent*) mixxxapplication.cpp:200 (mixxx:arm64+0x100ef48dc)
    #46 QCoreApplication::notifyInternal2(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b3dec)
    #47 QCoreApplication::sendEvent(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b4ea8)
    #48 QWidgetPrivate::show_helper() <null>:230508612 (mixxx:arm64+0x1022741b0)
    #49 QWidgetPrivate::setVisible(bool) <null>:230508612 (mixxx:arm64+0x102275c10)
    #50 QWidget::setVisible(bool) <null>:230508612 (mixxx:arm64+0x1022753f0)
    #51 QWidget::show() <null>:230508612 (mixxx:arm64+0x102273fc8)
    #52 MixxxMainWindow::initializeQOpenGL() mixxxmainwindow.cpp:161 (mixxx:arm64+0x100f30e64)
    #53 (anonymous namespace)::runMixxx(MixxxApplication*, CmdlineArgs const&) main.cpp:92 (mixxx:arm64+0x100010fd8)
    #54 main main.cpp:243 (mixxx:arm64+0x10000ff08)

  Thread T71 (tid=8891322, running) created by main thread at:
    #0 pthread_create <null>:230508612 (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x2ffec)
    #1 HALB_IOThread::DispatchPThread(void* (*)(void*), void*) <null>:230508612 (CoreAudio:arm64e+0x332910)
    #2 Pa_StartStream <null>:230508612 (mixxx:arm64+0x10210c97c)
    #3 SoundDevicePortAudio::open(bool, int) sounddeviceportaudio.cpp:363 (mixxx:arm64+0x10116f51c)
    #4 SoundManager::setupDevices() soundmanager.cpp:475 (mixxx:arm64+0x101178660)
    #5 MixxxMainWindow::initialize() mixxxmainwindow.cpp:346 (mixxx:arm64+0x100f321e4)
    #6 QtPrivate::FunctorCall<QtPrivate::IndexesList<>, QtPrivate::List<>, void, void (MixxxMainWindow::*)()>::call(void (MixxxMainWindow::*)(), MixxxMainWindow*, void**) qobjectdefs_impl.h:137 (mixxx:arm64+0x100f46a14)
    #7 void QtPrivate::FunctionPointer<void (MixxxMainWindow::*)()>::call<QtPrivate::List<>, void>(void (MixxxMainWindow::*)(), MixxxMainWindow*, void**) qobjectdefs_impl.h:174 (mixxx:arm64+0x100f468e0)
    #8 QtPrivate::QSlotObject<void (MixxxMainWindow::*)(), QtPrivate::List<>, void>::impl(int, QtPrivate::QSlotObjectBase*, QObject*, void**, bool*) qobjectdefs_impl.h:432 (mixxx:arm64+0x100f46744)
    #9 void doActivate<false>(QObject*, int, void**) <null>:230508612 (mixxx:arm64+0x10011ef08)
    #10 QMetaObject::activate(QObject*, QMetaObject const*, int, void**) <null>:230508612 (mixxx:arm64+0x10011db88)
    #11 WInitialGLWidget::onInitialized() moc_winitialglwidget.cpp:160 (mixxx:arm64+0x1015c2d78)
    #12 WInitialGLWidget::initializeGL() winitialglwidget.cpp:17 (mixxx:arm64+0x1015c3294)
    #13 OpenGLWindow::initializeGL() openglwindow.cpp:31 (mixxx:arm64+0x1015c00b8)
    #14 QOpenGLWindowPrivate::initialize() <null>:230508612 (mixxx:arm64+0x1044fcd88)
    #15 QOpenGLWindow::resizeEvent(QResizeEvent*) <null>:230508612 (mixxx:arm64+0x1044fdd78)
    #16 QWindow::event(QEvent*) <null>:230508612 (mixxx:arm64+0x104cdbd1c)
    #17 QPaintDeviceWindow::event(QEvent*) <null>:230508612 (mixxx:arm64+0x104ca4814)
    #18 OpenGLWindow::event(QEvent*) openglwindow.cpp:70 (mixxx:arm64+0x1015c03fc)
    #19 QApplicationPrivate::notify_helper(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fabdc)
    #20 QApplication::notify(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fc184)
    #21 MixxxApplication::notify(QObject*, QEvent*) mixxxapplication.cpp:200 (mixxx:arm64+0x100ef48dc)
    #22 QCoreApplication::notifyInternal2(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b3dec)
    #23 QCoreApplication::sendSpontaneousEvent(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b4f08)
    #24 QGuiApplicationPrivate::processGeometryChangeEvent(QWindowSystemInterfacePrivate::GeometryChangeEvent*) <null>:230508612 (mixxx:arm64+0x104c8a1b4)
    #25 QGuiApplicationPrivate::processWindowSystemEvent(QWindowSystemInterfacePrivate::WindowSystemEvent*) <null>:230508612 (mixxx:arm64+0x104c86310)
    #26 QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) <null>:230508612 (mixxx:arm64+0x104ce3884)
    #27 QWindowSystemInterface::flushWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) <null>:230508612 (mixxx:arm64+0x104ce3304)
    #28 QCocoaWindow::handleGeometryChange() <null>:230508612 (mixxx:arm64+0x101c7da5c)
    #29 QWindow::show() <null>:230508612 (mixxx:arm64+0x104cd2c24)
    #30 QWindowContainer::event(QEvent*) <null>:230508612 (mixxx:arm64+0x102299710)
    #31 QApplicationPrivate::notify_helper(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fabdc)
    #32 QApplication::notify(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fd364)
    #33 MixxxApplication::notify(QObject*, QEvent*) mixxxapplication.cpp:200 (mixxx:arm64+0x100ef48dc)
    #34 QCoreApplication::notifyInternal2(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b3dec)
    #35 QCoreApplication::sendEvent(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b4ea8)
    #36 QWidgetPrivate::show_helper() <null>:230508612 (mixxx:arm64+0x1022741b0)
    #37 QWidgetPrivate::setVisible(bool) <null>:230508612 (mixxx:arm64+0x102275c10)
    #38 QWidget::setVisible(bool) <null>:230508612 (mixxx:arm64+0x1022753f0)
    #39 QWidget::show() <null>:230508612 (mixxx:arm64+0x102273fc8)
    #40 WGLWidget::showEvent(QShowEvent*) wglwidgetqopengl.cpp:44 (mixxx:arm64+0x1015c2698)
    #41 QWidget::event(QEvent*) <null>:230508612 (mixxx:arm64+0x102276240)
    #42 QApplicationPrivate::notify_helper(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fabdc)
    #43 QApplication::notify(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1021fd364)
    #44 MixxxApplication::notify(QObject*, QEvent*) mixxxapplication.cpp:200 (mixxx:arm64+0x100ef48dc)
    #45 QCoreApplication::notifyInternal2(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b3dec)
    #46 QCoreApplication::sendEvent(QObject*, QEvent*) <null>:230508612 (mixxx:arm64+0x1000b4ea8)
    #47 QWidgetPrivate::show_helper() <null>:230508612 (mixxx:arm64+0x1022741b0)
    #48 QWidgetPrivate::setVisible(bool) <null>:230508612 (mixxx:arm64+0x102275c10)
    #49 QWidget::setVisible(bool) <null>:230508612 (mixxx:arm64+0x1022753f0)
    #50 QWidget::show() <null>:230508612 (mixxx:arm64+0x102273fc8)
    #51 MixxxMainWindow::initializeQOpenGL() mixxxmainwindow.cpp:161 (mixxx:arm64+0x100f30e64)
    #52 (anonymous namespace)::runMixxx(MixxxApplication*, CmdlineArgs const&) main.cpp:92 (mixxx:arm64+0x100010fd8)
    #53 main main.cpp:243 (mixxx:arm64+0x10000ff08)

  Thread T16 (tid=8890871, running) created by main thread at:
    #0 pthread_create <null>:230508612 (libclang_rt.tsan_osx_dynamic.dylib:arm64e+0x2ffec)
    #1 QThread::start(QThread::Priority) <null>:230508612 (mixxx:arm64+0x100249b9c)
    #2 EngineSideChain::EngineSideChain(QSharedPointer<ConfigObject<ConfigValue>>, float*) enginesidechain.cpp:37 (mixxx:arm64+0x1009dc904)
    #3 EngineSideChain::EngineSideChain(QSharedPointer<ConfigObject<ConfigValue>>, float*) enginesidechain.cpp:30 (mixxx:arm64+0x1009dcb38)
    #4 EngineMixer::EngineMixer(QSharedPointer<ConfigObject<ConfigValue>>, QString const&, EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>, bool) enginemixer.cpp:168 (mixxx:arm64+0x1009a4630)
    #5 EngineMixer::EngineMixer(QSharedPointer<ConfigObject<ConfigValue>>, QString const&, EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>, bool) enginemixer.cpp:52 (mixxx:arm64+0x1009a0018)
    #6 EngineMixer* std::__1::construct_at[abi:v160006]<EngineMixer, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool, EngineMixer*>(EngineMixer*, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) construct_at.h:38 (mixxx:arm64+0x100647ae0)
    #7 void std::__1::allocator_traits<std::__1::allocator<EngineMixer>>::construct[abi:v160006]<EngineMixer, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool, void, void>(std::__1::allocator<EngineMixer>&, EngineMixer*, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) allocator_traits.h:304 (mixxx:arm64+0x100647754)
    #8 std::__1::__shared_ptr_emplace<EngineMixer, std::__1::allocator<EngineMixer>>::__shared_ptr_emplace[abi:v160006]<QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool>(std::__1::allocator<EngineMixer>, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) shared_ptr.h:284 (mixxx:arm64+0x1006475c0)
    #9 std::__1::__shared_ptr_emplace<EngineMixer, std::__1::allocator<EngineMixer>>::__shared_ptr_emplace[abi:v160006]<QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool>(std::__1::allocator<EngineMixer>, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) shared_ptr.h:276 (mixxx:arm64+0x100646fd8)
    #10 std::__1::shared_ptr<EngineMixer> std::__1::allocate_shared[abi:v160006]<EngineMixer, std::__1::allocator<EngineMixer>, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool, void>(std::__1::allocator<EngineMixer> const&, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) shared_ptr.h:995 (mixxx:arm64+0x100646dbc)
    #11 std::__1::shared_ptr<EngineMixer> std::__1::make_shared[abi:v160006]<EngineMixer, QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*, std::__1::shared_ptr<ChannelHandleFactory>&, bool, void>(QSharedPointer<ConfigObject<ConfigValue>>&, char const (&) [9], EffectsManager*&&, std::__1::shared_ptr<ChannelHandleFactory>&, bool&&) shared_ptr.h:1004 (mixxx:arm64+0x100638fcc)
    #12 mixxx::CoreServices::initialize(QApplication*) coreservices.cpp:264 (mixxx:arm64+0x10063735c)
    #13 (anonymous namespace)::runMixxx(MixxxApplication*, CmdlineArgs const&) main.cpp:87 (mixxx:arm64+0x100010fcc)
    #14 main main.cpp:243 (mixxx:arm64+0x10000ff08)

SUMMARY: ThreadSanitizer: data race (mixxx:arm64+0x102110a24) in PaUtil_WriteRingBuffer+0x190

```

</p>
</details>

### Version

2.5

### OS

_No response_

m0dB:
Reimplement fifo as a proper lock free SCSP ring buffer with atomics for thread-safety instead of using PaUtilRingBuffer as a backend.

Added unit tests.

One might consider that the tsan warnings for PaUtilRingBuffer or not "serious" but they sure are ugly and spam my tsan logs.

Fixes #13864 #13863  #13866 #13868

### DIFF
--- a/src/test/fifotest.cpp
+++ b/src/test/fifotest.cpp
@@ -75,10 +75,19 @@ TEST_P(FifoTest, flushReadTest) {
 
     ASSERT_EQ(0, fifo.readAvailable());
     ASSERT_EQ(100, fifo.write(data.data(), 100));
-    ASSERT_EQ((param.offset + 50) % param.expectedBufferSize,
-            fifo.flushReadData(50));
-    ASSERT_EQ((param.offset + 100) % param.expectedBufferSize,
-            fifo.flushReadData(1000000));
+
+    int expected;
+    expected = (param.offset + 50) % param.expectedBufferSize;
+    if (expected < 0) {
+        expected += param.expectedBufferSize;
+    }
+    ASSERT_EQ(expected, fifo.flushReadData(50));
+
+    expected = (param.offset + 100) % param.expectedBufferSize;
+    if (expected < 0) {
+        expected += param.expectedBufferSize;
+    }
+    ASSERT_EQ(expected, fifo.flushReadData(1000000));
     ASSERT_EQ(param.expectedBufferSize, fifo.write(data.data(), 1000000));
     ASSERT_EQ(param.expectedBufferSize, fifo.readAvailable());
 }
--- a/src/test/fifotest.cpp
+++ b/src/test/fifotest.cpp
@@ -8,8 +8,8 @@
 namespace {
 
 struct Param {
-    std::size_t requestedBufferSize;
-    std::size_t expectedBufferSize;
+    int requestedBufferSize;
+    int expectedBufferSize;
     int offset;
 };
 
@@ -56,8 +56,8 @@ TEST_P(FifoTest, readAvailableTest) {
     ASSERT_EQ(100, fifo.readAvailable());
     ASSERT_EQ(50, fifo.read(data.data(), 50));
     ASSERT_EQ(50, fifo.readAvailable());
-    ASSERT_EQ(static_cast<int>(param.expectedBufferSize - 50), fifo.write(data.data(), 1000000));
-    ASSERT_EQ(static_cast<int>(param.expectedBufferSize), fifo.readAvailable());
+    ASSERT_EQ(param.expectedBufferSize - 50, fifo.write(data.data(), 1000000));
+    ASSERT_EQ(param.expectedBufferSize, fifo.readAvailable());
 }
 
 TEST_P(FifoTest, flushReadTest) {
@@ -75,12 +75,12 @@ TEST_P(FifoTest, flushReadTest) {
 
     ASSERT_EQ(0, fifo.readAvailable());
     ASSERT_EQ(100, fifo.write(data.data(), 100));
-    ASSERT_EQ(static_cast<int>((param.offset + 50) % param.expectedBufferSize),
+    ASSERT_EQ((param.offset + 50) % param.expectedBufferSize,
             fifo.flushReadData(50));
-    ASSERT_EQ(static_cast<int>((param.offset + 100) % param.expectedBufferSize),
+    ASSERT_EQ((param.offset + 100) % param.expectedBufferSize,
             fifo.flushReadData(1000000));
-    ASSERT_EQ(static_cast<int>(param.expectedBufferSize), fifo.write(data.data(), 1000000));
-    ASSERT_EQ(static_cast<int>(param.expectedBufferSize), fifo.readAvailable());
+    ASSERT_EQ(param.expectedBufferSize, fifo.write(data.data(), 1000000));
+    ASSERT_EQ(param.expectedBufferSize, fifo.readAvailable());
 }
 
 TEST_P(FifoTest, readWriteStressTest) {
@@ -104,7 +104,7 @@ TEST_P(FifoTest, readWriteStressTest) {
     std::random_device rd;
     std::mt19937 mt(rd());
     std::uniform_int_distribution<int> dist(
-            0, static_cast<int>(param.expectedBufferSize + param.expectedBufferSize / 10));
+            0, param.expectedBufferSize + param.expectedBufferSize / 10);
 
     while (k < 1000000) {
         int n = dist(mt);
@@ -112,10 +112,10 @@ TEST_P(FifoTest, readWriteStressTest) {
         for (int i = 0; i < m; i++) {
             wdata[i] = k++;
         }
-        ASSERT_EQ(static_cast<int>(m), fifo.write(wdata.data(), n));
+        ASSERT_EQ(m, fifo.write(wdata.data(), n));
         n = dist(mt);
         m = std::min(n, fifo.readAvailable());
-        ASSERT_EQ(static_cast<int>(m), fifo.read(rdata.data(), n));
+        ASSERT_EQ(m, fifo.read(rdata.data(), n));
         for (int i = 0; i < m; i++) {
             ASSERT_EQ(j++, rdata[i]);
         }
@@ -152,8 +152,8 @@ TEST_P(FifoTest, readWriteStressTestRegions) {
         ring_buffer_size_t size1;
         uint32_t* ptr2;
         ring_buffer_size_t size2;
-        ASSERT_EQ(static_cast<int>(m), fifo.aquireWriteRegions(n, &ptr1, &size1, &ptr2, &size2));
-        ASSERT_EQ(static_cast<int>(m), size1 + size2);
+        ASSERT_EQ(m, fifo.aquireWriteRegions(n, &ptr1, &size1, &ptr2, &size2));
+        ASSERT_EQ(m, size1 + size2);
         for (int i = 0; i < size1; i++) {
             ptr1[i] = k++;
         }
@@ -163,8 +163,8 @@ TEST_P(FifoTest, readWriteStressTestRegions) {
         fifo.releaseWriteRegions(m);
         n = dist(mt);
         m = std::min(n, fifo.readAvailable());
-        ASSERT_EQ(static_cast<int>(m), fifo.aquireReadRegions(n, &ptr1, &size1, &ptr2, &size2));
-        ASSERT_EQ(static_cast<int>(m), size1 + size2);
+        ASSERT_EQ(m, fifo.aquireReadRegions(n, &ptr1, &size1, &ptr2, &size2));
+        ASSERT_EQ(m, size1 + size2);
         for (int i = 0; i < size1; i++) {
             ASSERT_EQ(j++, ptr1[i]);
         }
--- a/src/test/fifotest.cpp
+++ b/src/test/fifotest.cpp
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <random>
+#include <thread>
 
 #include "util/fifo.h"
 
@@ -198,3 +199,191 @@ INSTANTIATE_TEST_SUITE_P(FifoTestSuite,
                 Param{1234, 2048, 1200}));
 
 } // namespace
+
+constexpr int rwtotal = 500000000;
+
+template<class T_FIFO>
+class MultiThreadRW {
+    T_FIFO m_fifo;
+    bool m_ok;
+    const int m_bufferSize;
+    const int m_total;
+    const bool m_wait;
+
+  public:
+    MultiThreadRW(int ringBufferSize, int bufferSize, int total, bool wait)
+            : m_fifo(ringBufferSize),
+              m_ok{},
+              m_bufferSize(bufferSize),
+              m_total(total),
+              m_wait(wait) {
+    }
+
+    void write() {
+        int k = 0;
+        std::vector<int> buffer(m_bufferSize);
+        while (k != m_total) {
+            int n = std::min(m_bufferSize, m_total - k);
+            if (m_wait) {
+                while (m_fifo.writeAvailable() < n) {
+                }
+            }
+            n = std::min(n, m_fifo.writeAvailable());
+            for (int j = 0; j < n; j++) {
+                buffer[j] = k++;
+            }
+            m_fifo.write(buffer.data(), n);
+        }
+    }
+
+    void read() {
+        m_ok = true;
+        int k = 0;
+        std::vector<int> buffer(m_bufferSize);
+        while (k != m_total) {
+            int n = std::min(m_bufferSize, m_total - k);
+            if (m_wait) {
+                while (m_fifo.readAvailable() < n) {
+                }
+            }
+            n = m_fifo.read(buffer.data(), n);
+            for (int j = 0; j < n; j++) {
+                m_ok &= (buffer[j] == k++);
+            }
+        }
+    }
+
+    bool run() {
+        m_ok = true;
+        std::thread th1(&MultiThreadRW<T_FIFO>::write, this);
+        std::thread th2(&MultiThreadRW<T_FIFO>::read, this);
+        th1.join();
+        th2.join();
+        return m_ok;
+    }
+};
+
+TEST(FifoTest, MultiThreadRW) {
+    MultiThreadRW<FIFO<int>> io(65536, 1024, rwtotal, false);
+    bool ok = io.run();
+    ASSERT_TRUE(ok);
+}
+
+TEST(FifoTest, MultiThreadRW_PA) {
+    MultiThreadRW<PA::FIFO<int>> io(65536, 1024, rwtotal, false);
+    bool ok = io.run();
+    ASSERT_TRUE(ok);
+}
+
+TEST(FifoTest, MultiThreadRW_Wait) {
+    MultiThreadRW<FIFO<int>> io(65536, 1024, rwtotal, true);
+    bool ok = io.run();
+    ASSERT_TRUE(ok);
+}
+
+TEST(FifoTest, MultiThreadRW_PA_Wait) {
+    MultiThreadRW<PA::FIFO<int>> io(65536, 1024, rwtotal, true);
+    bool ok = io.run();
+    ASSERT_TRUE(ok);
+}
+
+template<class T_FIFO>
+class MultiThreadRegionRW {
+    T_FIFO m_fifo;
+    bool m_ok;
+    const int m_bufferSize;
+    const int m_total;
+    const bool m_wait;
+
+  public:
+    MultiThreadRegionRW(
+            int ringBufferSize, int bufferSize, int total, bool wait)
+            : m_fifo(ringBufferSize),
+              m_ok{},
+              m_bufferSize(bufferSize),
+              m_total(total),
+              m_wait(wait) {
+    }
+
+    void write() {
+        int k = 0;
+        std::vector<int> buffer(m_bufferSize);
+        while (k != m_total) {
+            int n = std::min(m_bufferSize, m_total - k);
+            int* ptr1;
+            int* ptr2;
+            ring_buffer_size_t size1;
+            ring_buffer_size_t size2;
+            if (m_wait) {
+                while (m_fifo.writeAvailable() < n) {
+                }
+            }
+            n = m_fifo.aquireWriteRegions(n, &ptr1, &size1, &ptr2, &size2);
+            for (int j = 0; j < size1; j++) {
+                ptr1[j] = k++;
+            }
+            for (int j = 0; j < size2; j++) {
+                ptr2[j] = k++;
+            }
+            m_fifo.releaseWriteRegions(n);
+        }
+    }
+
+    void read() {
+        m_ok = true;
+        int k = 0;
+        std::vector<int> buffer(m_bufferSize);
+        while (k != m_total) {
+            int n = std::min(m_bufferSize, m_total - k);
+            int* ptr1;
+            int* ptr2;
+            ring_buffer_size_t size1;
+            ring_buffer_size_t size2;
+            if (m_wait) {
+                while (m_fifo.readAvailable() < n) {
+                }
+            }
+            n = m_fifo.aquireReadRegions(n, &ptr1, &size1, &ptr2, &size2);
+            for (int j = 0; j < size1; j++) {
+                m_ok &= (ptr1[j] == k++);
+            }
+            for (int j = 0; j < size2; j++) {
+                m_ok &= (ptr2[j] == k++);
+            }
+            m_fifo.releaseReadRegions(n);
+        }
+    }
+
+    bool run() {
+        m_ok = true;
+        std::thread th1(&MultiThreadRegionRW<T_FIFO>::write, this);
+        std::thread th2(&MultiThreadRegionRW<T_FIFO>::read, this);
+        th1.join();
+        th2.join();
+        return m_ok;
+    }
+};
+
+TEST(FifoTest, MultiThreadRegionRW) {
+    MultiThreadRegionRW<FIFO<int>> io(65536, 256, rwtotal, false);
+    bool ok = io.run();
+    ASSERT_TRUE(ok);
+}
+
+TEST(FifoTest, MultiThreadRegionRW_PA) {
+    MultiThreadRegionRW<PA::FIFO<int>> io(65536, 256, rwtotal, false);
+    bool ok = io.run();
+    ASSERT_TRUE(ok);
+}
+
+TEST(FifoTest, MultiThreadRegionRW_Wait) {
+    MultiThreadRegionRW<FIFO<int>> io(65536, 256, rwtotal, true);
+    bool ok = io.run();
+    ASSERT_TRUE(ok);
+}
+
+TEST(FifoTest, MultiThreadRegionRW_PA_Wait) {
+    MultiThreadRegionRW<PA::FIFO<int>> io(65536, 256, rwtotal, true);
+    bool ok = io.run();
+    ASSERT_TRUE(ok);
+}
--- a/src/test/fifotest.cpp
+++ b/src/test/fifotest.cpp
@@ -1,0 +1,191 @@
+#include <gtest/gtest.h>
+
+#include <algorithm>
+#include <random>
+
+#include "util/fifo.h"
+
+namespace {
+
+struct Param {
+    std::size_t requestedBufferSize;
+    std::size_t expectedBufferSize;
+    int offset;
+};
+
+class FifoTest : public testing::TestWithParam<Param> {
+};
+
+TEST_P(FifoTest, writeAvailableTest) {
+    const auto param = FifoTest::GetParam();
+    std::vector<float> data(param.requestedBufferSize);
+    FIFO<float> fifo(param.requestedBufferSize);
+    fifo.releaseReadRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+    fifo.releaseWriteRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+
+    ASSERT_EQ(param.expectedBufferSize, fifo.writeAvailable());
+    ASSERT_EQ(100, fifo.write(data.data(), 100));
+    ASSERT_EQ(param.expectedBufferSize - 100, fifo.writeAvailable());
+    ASSERT_EQ(50, fifo.read(data.data(), 50));
+    ASSERT_EQ(param.expectedBufferSize - 50, fifo.writeAvailable());
+    ASSERT_EQ(param.expectedBufferSize - 50, fifo.write(data.data(), 1000000));
+    ASSERT_EQ(0, fifo.writeAvailable());
+}
+
+TEST_P(FifoTest, readAvailableTest) {
+    const auto param = FifoTest::GetParam();
+    std::vector<float> data(param.requestedBufferSize);
+    FIFO<float> fifo(param.requestedBufferSize);
+    fifo.releaseReadRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+    fifo.releaseWriteRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+
+    ASSERT_EQ(0, fifo.readAvailable());
+    ASSERT_EQ(100, fifo.write(data.data(), 100));
+    ASSERT_EQ(100, fifo.readAvailable());
+    ASSERT_EQ(50, fifo.read(data.data(), 50));
+    ASSERT_EQ(50, fifo.readAvailable());
+    ASSERT_EQ(static_cast<int>(param.expectedBufferSize - 50), fifo.write(data.data(), 1000000));
+    ASSERT_EQ(static_cast<int>(param.expectedBufferSize), fifo.readAvailable());
+}
+
+TEST_P(FifoTest, flushReadTest) {
+    const auto param = FifoTest::GetParam();
+    std::vector<float> data(param.requestedBufferSize);
+    FIFO<float> fifo(param.requestedBufferSize);
+    fifo.releaseReadRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+    fifo.releaseWriteRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+
+    ASSERT_EQ(0, fifo.readAvailable());
+    ASSERT_EQ(100, fifo.write(data.data(), 100));
+    ASSERT_EQ(static_cast<int>((param.offset + 50) % param.expectedBufferSize),
+            fifo.flushReadData(50));
+    ASSERT_EQ(static_cast<int>((param.offset + 100) % param.expectedBufferSize),
+            fifo.flushReadData(1000000));
+    ASSERT_EQ(static_cast<int>(param.expectedBufferSize), fifo.write(data.data(), 1000000));
+    ASSERT_EQ(static_cast<int>(param.expectedBufferSize), fifo.readAvailable());
+}
+
+TEST_P(FifoTest, readWriteStressTest) {
+    const auto param = FifoTest::GetParam();
+    std::vector<uint32_t> data(param.expectedBufferSize);
+    FIFO<uint32_t> fifo(param.requestedBufferSize);
+    fifo.releaseReadRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+    fifo.releaseWriteRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+
+    std::vector<uint32_t> wdata(param.expectedBufferSize + param.expectedBufferSize / 10);
+    std::vector<uint32_t> rdata(param.expectedBufferSize + param.expectedBufferSize / 10);
+    uint32_t k = 0;
+    uint32_t j = 0;
+
+    std::random_device rd;
+    std::mt19937 mt(rd());
+    std::uniform_int_distribution<int> dist(
+            0, static_cast<int>(param.expectedBufferSize + param.expectedBufferSize / 10));
+
+    while (k < 1000000) {
+        int n = dist(mt);
+        int m = std::min(n, fifo.writeAvailable());
+        for (int i = 0; i < m; i++) {
+            wdata[i] = k++;
+        }
+        ASSERT_EQ(static_cast<int>(m), fifo.write(wdata.data(), n));
+        n = dist(mt);
+        m = std::min(n, fifo.readAvailable());
+        ASSERT_EQ(static_cast<int>(m), fifo.read(rdata.data(), n));
+        for (int i = 0; i < m; i++) {
+            ASSERT_EQ(j++, rdata[i]);
+        }
+    }
+}
+
+TEST_P(FifoTest, readWriteStressTestRegions) {
+    const auto param = FifoTest::GetParam();
+    std::vector<uint32_t> data(param.expectedBufferSize);
+    FIFO<uint32_t> fifo(param.requestedBufferSize);
+    fifo.releaseReadRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+    fifo.releaseWriteRegions(param.offset >= 0
+                    ? param.offset
+                    : std::numeric_limits<std::size_t>::max() + param.offset +
+                            1);
+
+    std::vector<uint32_t> wdata(param.expectedBufferSize + param.expectedBufferSize / 10);
+    std::vector<uint32_t> rdata(param.expectedBufferSize + param.expectedBufferSize / 10);
+    uint32_t k = 0;
+    uint32_t j = 0;
+
+    std::random_device rd;
+    std::mt19937 mt(rd());
+    std::uniform_int_distribution<int> dist(
+            0, param.expectedBufferSize + param.expectedBufferSize / 10);
+
+    while (k < 1000000) {
+        int n = dist(mt);
+        int m = std::min(n, fifo.writeAvailable());
+        uint32_t* ptr1;
+        ring_buffer_size_t size1;
+        uint32_t* ptr2;
+        ring_buffer_size_t size2;
+        ASSERT_EQ(static_cast<int>(m), fifo.aquireWriteRegions(n, &ptr1, &size1, &ptr2, &size2));
+        ASSERT_EQ(static_cast<int>(m), size1 + size2);
+        for (int i = 0; i < size1; i++) {
+            ptr1[i] = k++;
+        }
+        for (int i = 0; i < size2; i++) {
+            ptr2[i] = k++;
+        }
+        fifo.releaseWriteRegions(m);
+        n = dist(mt);
+        m = std::min(n, fifo.readAvailable());
+        ASSERT_EQ(static_cast<int>(m), fifo.aquireReadRegions(n, &ptr1, &size1, &ptr2, &size2));
+        ASSERT_EQ(static_cast<int>(m), size1 + size2);
+        for (int i = 0; i < size1; i++) {
+            ASSERT_EQ(j++, ptr1[i]);
+        }
+        for (int i = 0; i < size2; i++) {
+            ASSERT_EQ(j++, ptr2[i]);
+        }
+        fifo.releaseReadRegions(m);
+    }
+}
+
+INSTANTIATE_TEST_SUITE_P(FifoTestSuite,
+        FifoTest,
+        testing::Values(
+                Param{1024, 1024, 0},
+                Param{1024, 1024, 1200},
+                Param{1024, 1024, -1200},
+                Param{65536, 65536, 0},
+                Param{65536, 65536, 1200},
+                Param{65536, 65536, -1200},
+                Param{1234, 2048, 0},
+                Param{1234, 2048, -1200},
+                Param{1234, 2048, 1200}));
+
+} // namespace
--- a/src/util/fifo.h
+++ b/src/util/fifo.h
@@ -1,69 +1,164 @@
 #pragma once
 
-#include "pa_ringbuffer.h"
+#include <cstddef>
 
 #include "util/class.h"
 #include "util/math.h"
 
+// Fast, trivial type only single producer single consumer ring buffer, lock and wait free.
+// Internal buffer size will be rounded up to a power of two.
+
+// Internally we use std::size_t, but the original API used int for the return values.
+// It would be better to use std::size_t also for the return values, but this breaks
+// Windows compilation and should be done in a follow-up PR.
+using ring_buffer_size_t = int;
+
 template <class DataType>
 class FIFO {
   public:
-    explicit FIFO(int size)
-            : m_data(roundUpToPowerOf2(size)) {
-        // If we can't represent the next higher power of 2 then bail.
-        if (m_data.size() == 0) {
-            return;
-        }
-        PaUtil_InitializeRingBuffer(&m_ringBuffer,
-                static_cast<ring_buffer_size_t>(sizeof(DataType)),
-                static_cast<ring_buffer_size_t>(m_data.size()),
-                m_data.data());
+    using size_type = std::size_t;
+
+    explicit FIFO(size_type size)
+            : m_size{roundUpToPowerOf2(static_cast<unsigned int>(size))},
+              m_mask{m_size - 1},
+              m_data(m_size),
+              m_writeIndex{0},
+              m_readIndex{0} {
     }
-    virtual ~FIFO() {
-    }
+
+    // Returns the number of values in the ringbuffer available for read
     int readAvailable() const {
-        return PaUtil_GetRingBufferReadAvailable(&m_ringBuffer);
+        const size_type readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_acquire);
+        return static_cast<int>(writeIndex - readIndex);
     }
+    // Returns the space in the ringbuffer available for write
     int writeAvailable() const {
-        return PaUtil_GetRingBufferWriteAvailable(&m_ringBuffer);
+        const size_type readIndex = m_readIndex.load(std::memory_order_acquire);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_relaxed);
+        return static_cast<int>(m_size - (writeIndex - readIndex));
     }
-    int read(DataType* pData, int count) {
-        return PaUtil_ReadRingBuffer(&m_ringBuffer, pData, count);
+    // Read count values to the ring buffer. If less then count values are
+    // available, only read the available amount. The return value is the
+    // actual number of values read. If the read index reached the end of the
+    // ringbuffer, the remainder is read from the start.
+    int read(DataType* pData, size_type count) {
+        size_type readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_acquire);
+        const size_type available = writeIndex - readIndex;
+        count = std::min(available, count);
+        readIndex = readIndex & m_mask;
+        const size_type n = std::min(m_size - readIndex, count);
+        std::copy(m_data.data() + readIndex, m_data.data() + readIndex + n, pData);
+        std::copy(m_data.data(), m_data.data() + count - n, pData + n);
+        m_readIndex.fetch_add(count, std::memory_order_release);
+        return static_cast<int>(count);
     }
-    int write(const DataType* pData, int count) {
-        return PaUtil_WriteRingBuffer(&m_ringBuffer, pData, count);
+    // Write count samples to the ring buffer. If space available is less then
+    // count, only write the available space amount. The return value is the
+    // actual number of values written. If the write index reached the end of the
+    // ringbuffer, the remainder is written to the start.
+    int write(const DataType* pData, size_type count) {
+        const size_type readIndex = m_readIndex.load(std::memory_order_acquire);
+        size_type writeIndex = m_writeIndex.load(std::memory_order_relaxed);
+        const size_type available = m_size - (writeIndex - readIndex);
+        count = std::min(available, count);
+        writeIndex = writeIndex & m_mask;
+        const size_type n = std::min(m_size - writeIndex, count);
+        std::copy(pData, pData + n, m_data.data() + writeIndex);
+        std::copy(pData + n, pData + count, m_data.data());
+        m_writeIndex.fetch_add(count, std::memory_order_release);
+        return static_cast<int>(count);
     }
-    void writeBlocking(const DataType* pData, int count) {
-        int written = 0;
+    void writeBlocking(const DataType* pData, size_type count) {
+        size_type written = 0;
         while (written < count) {
             written += write(pData + written, count - written);
         }
     }
-    int aquireWriteRegions(int count,
-            DataType** dataPtr1, ring_buffer_size_t* sizePtr1,
-            DataType** dataPtr2, ring_buffer_size_t* sizePtr2) {
-        return PaUtil_GetRingBufferWriteRegions(&m_ringBuffer, count,
-                (void**)dataPtr1, sizePtr1, (void**)dataPtr2, sizePtr2);
+    // Give direct access to the ring buffer at the read index
+    // to a region of count values. If less than count values
+    // are available, limit to the available amount.
+    // If the region surpasses the end of the ring buffer,
+    // the first region will be until the end, the second
+    // region will be the remainder from the start, otherwise
+    // the second region will be nullptr and size 0.
+    int aquireReadRegions(size_type count,
+            DataType** dataPtr1,
+            int* sizePtr1,
+            DataType** dataPtr2,
+            int* sizePtr2) {
+        size_type readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_acquire);
+        const size_type available = writeIndex - readIndex;
+        count = std::min(available, count);
+        readIndex = readIndex & m_mask;
+        *sizePtr1 = static_cast<int>(std::min(m_size - readIndex, count));
+        *sizePtr2 = static_cast<int>(count) - *sizePtr1;
+        *dataPtr1 = m_data.data() + readIndex;
+        *dataPtr2 = *sizePtr2 == 0 ? nullptr : m_data.data();
+
+        return count;
     }
-    int releaseWriteRegions(int count) {
-        return PaUtil_AdvanceRingBufferWriteIndex(&m_ringBuffer, count);
+    // Advance the read index after aquireReadRegions. Count should
+    // be the return value of aquireReadRegions, which is the total
+    // region size acquired.
+    int releaseReadRegions(size_type count) {
+        const size_type readIndex = (m_readIndex.load(std::memory_order_relaxed) + count);
+        m_readIndex.store(readIndex, std::memory_order_release);
+        return static_cast<int>(readIndex & m_mask);
     }
-    int aquireReadRegions(int count,
-            DataType** dataPtr1, ring_buffer_size_t* sizePtr1,
-            DataType** dataPtr2, ring_buffer_size_t* sizePtr2) {
-        return PaUtil_GetRingBufferReadRegions(&m_ringBuffer, count,
-                (void**)dataPtr1, sizePtr1, (void**)dataPtr2, sizePtr2);
+    // Same as aquireReadRegions, for write operations
+    int aquireWriteRegions(size_type count,
+            DataType** dataPtr1,
+            int* sizePtr1,
+            DataType** dataPtr2,
+            int* sizePtr2) {
+        const size_type readIndex = m_readIndex.load(std::memory_order_acquire);
+        size_type writeIndex = m_writeIndex.load(std::memory_order_relaxed);
+        const size_type available = m_size - (writeIndex - readIndex);
+        count = std::min(available, count);
+        writeIndex = writeIndex & m_mask;
+        *sizePtr1 = static_cast<int>(std::min(m_size - writeIndex, count));
+        *sizePtr2 = static_cast<int>(count) - *sizePtr1;
+        *dataPtr1 = m_data.data() + writeIndex;
+        *dataPtr2 = *sizePtr2 == 0 ? nullptr : m_data.data();
+
+        return static_cast<int>(count);
     }
-    int releaseReadRegions(int count) {
-        return PaUtil_AdvanceRingBufferReadIndex(&m_ringBuffer, count);
+    // Same as releaseReadRegions, for write operations
+    int releaseWriteRegions(size_type count) {
+        const size_type writeIndex = (m_writeIndex.load(std::memory_order_relaxed) + count);
+        m_writeIndex.store(writeIndex, std::memory_order_release);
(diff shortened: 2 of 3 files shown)


Swiftb0y:
There are loads of ringbuffers in mixxx already and we already vendor https://github.com/rigtorp/SPSCQueue which has been proven to be correct.

Swiftb0y:
Sorry for not making mention of this earlier. I leave the decision to you whether you want to keep this implementation or use `rigtorps` implementation.

m0dB:
Oh, it is unfortunate I didn't know that. Also unfortunate that FIFO wasn't using that. Can we bring all these ring buffers back to a single implementations?

m0dB:
rigtorp/SPSCQueue.h is not comparable though. It using queue semantics, to push and pop single items. The ringbuffer I implemented (and what PaUtilRingBuffer does) is to read / write buffers (typically of samples), either by copy or by accessing regions in the ringbuffer directly.

Swiftb0y:
Thanks. I'll try to give this a review. Though I unfortunately don't have the most experience with multithreaded programming.

src/test/fifotest.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include <random>
+
+#include "util/fifo.h"
+
+namespace {
+
+class FifoTest : public testing::Test {
+};
+
+TEST_F(FifoTest, writeAvailableTest) {
```
Swiftb0y, line 13: KISS?

```suggestion
TEST(FifoTest, writeAvailableTest) {
```

src/test/fifotest.cpp
```diff
@@ -0,0 +1,182 @@
...
+    std::vector<float> data(1024);
+    FIFO<float> fifo(1024);
+    ASSERT_EQ(1024, fifo.writeAvailable());
+    ASSERT_EQ(100, fifo.write(data.data(), 100));
+    ASSERT_EQ(924, fifo.writeAvailable());
```
Swiftb0y, line 18: would an `EXPECT_EQ` fit more or do only asserts make sense here?

src/test/fifotest.cpp
```diff
@@ -0,0 +1,182 @@
...
+    ASSERT_EQ(50, fifo.read(data.data(), 50));
+    ASSERT_EQ(974, fifo.writeAvailable());
+    ASSERT_EQ(974, fifo.write(data.data(), 10000));
+    ASSERT_EQ(0, fifo.writeAvailable());
+}
+
+TEST_F(FifoTest, readAvailableTest) {
+    std::vector<float> data(1024);
+    FIFO<float> fifo(1024);
```
Swiftb0y, line 27: a (potentially Value-Parameterized) test that tests a couple more (non power-of-two) queue sizes would make sense, wdyt?

src/util/fifo.h
```diff
@@ -1,69 +1,116 @@
...
-    int writeAvailable() const {
-        return PaUtil_GetRingBufferWriteAvailable(&m_ringBuffer);
+    uint32_t read(DataType* pData, uint32_t count) {
+        const uint32_t readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const uint32_t writeIndex = m_writeIndex.load(std::memory_order_acquire);
+        const uint32_t available = writeIndex - readIndex;
+        count = std::min(available, count);
+        const uint32_t n = std::min(m_size - (readIndex & m_mask), count);
+        std::memcpy(pData, m_data.data() + (readIndex & m_mask), n * sizeof(DataType));
+        std::memcpy(pData + n, m_data.data(), (count - n) * sizeof(DataType));
```
Swiftb0y, line 37: Pretty sure this requires https://en.cppreference.com/w/cpp/types/is_trivially_copyable, right?

src/util/fifo.h
```diff
@@ -1,69 +1,116 @@
 #pragma once
 
-#include "pa_ringbuffer.h"
+#include <cstdint>
 
 #include "util/class.h"
 #include "util/math.h"
 
+using ring_buffer_size_t = uint32_t;
```
Swiftb0y, line 8: can you share some insight why not just `std::size_t`?

src/util/fifo.h
```diff
@@ -1,69 +1,116 @@
...
     }
 
   private:
+    const uint32_t m_size;
+    const uint32_t m_mask;
     std::vector<DataType> m_data;
-    PaUtilRingBuffer m_ringBuffer;
+    std::atomic<uint32_t> m_writeIndex;
+    std::atomic<uint32_t> m_readIndex;
     DISALLOW_COPY_AND_ASSIGN(FIFO);
```
Swiftb0y, line 115: can you implement move semantics and adhere to the rule of 5?

Swiftb0y (commented):
A couple surface-level comments. I'll follow up with a proper audit of the thread safety aspects later.

src/test/fifotest.cpp
m0dB: I agree

src/test/fifotest.cpp
m0dB: I didn’t check the style of the other tests but I am used to using ASSERT_xxx in cases like this. The condicion really has to hold, continuing is useless, things are very wrong if any of these conditions fail and the remainder of the test wont make sense.

src/test/fifotest.cpp
m0dB: Yes, also the offset near the numeric limit of the indices should be parametrized

src/util/fifo.h
m0dB: It does.

src/util/fifo.h
m0dB: Also works for me.  Or uint_fast32_t. I gladly adapt to what you think fits the mixxx codebase best.

src/util/fifo.h
m0dB: I’m just following the original FIFO here, apparently this was sufficient for the current use.

src/test/fifotest.cpp
Swiftb0y: okay, then it fully makes sense to have asserts. Thanks for the clarification.

src/util/fifo.h
Swiftb0y: right, so then we should either consider adding a `static_assert` for that or use the appropriate more generic algorithm (`std::copy`?).

src/util/fifo.h
Swiftb0y: I think for indexing into `std::vector` it always converts to a `std::size_t` anyways, so if we want to avoid that one sign extension instruction we should use whats most compatible with the underlying container (and avoid exposing it if possible).

src/util/fifo.h
Swiftb0y: Haven't looked at the diff, but I have a suspicion that the original implementation was written before move semantics were a thing (`C++11`), so we might as well go the extra step. We then may also look for unnecessarily heap-allocated `FIFO`s that workaround the lack of move semantics and inline those (the eliminated pointer dereference may result in a nice runtime speedup).

src/util/fifo.h
m0dB: Sure. I’ll change it to size_t

src/util/fifo.h
m0dB: Ok, but let’s do that in a follow up PR. The goal of these PRs is to get rid of all the tsan warnings.

src/util/fifo.h
m0dB: Yes, I’ll change it to std::copy.  Old habits die hard and I often notice that you are not plagued by the baggage of over 30 years of hardcore C and primitive C++ 😅

src/test/fifotest.cpp
m0dB: Note though that internally the queue sizes are always rounded up to the next power of two. Still, good to add parametrized tests.

src/util/fifo.h
```diff
@@ -1,69 +1,116 @@
...
+              m_mask{m_size - 1},
+              m_data(m_size),
+              m_writeIndex{0},
+              m_readIndex{0} {
     }
-    virtual ~FIFO() {
+    uint32_t readAvailable() const {
+        const uint32_t readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const uint32_t writeIndex = m_writeIndex.load(std::memory_order_acquire);
+        return writeIndex - readIndex;
```
daschuer, line 23: Can you add a comment about the case if a read or write happens between the wo atomic assesses?

daschuer: Cam you please also explain the use of the memory barriers? Is it correct?

src/util/fifo.h
daschuer: std::copy is slow:
https://github.com/mixxxdj/mixxx/blob/2397016ea899185207e03912773ec43553d0c6f2/src/util/sample.h#L62

We want here a replacement for the original implementation, that only works with trivial types. For C++ objects, we have already an alternative.

src/util/fifo.h
```diff
@@ -1,69 +1,116 @@
 #pragma once
 
-#include "pa_ringbuffer.h"
+#include <cstdint>
 
 #include "util/class.h"
 #include "util/math.h"
 
+using ring_buffer_size_t = uint32_t;
+
```
daschuer, line 9: This requires some comments about the purpose and limitations 

... Fast, trivial type only single reader and writer ...

daschuer (changes requested):

src/util/fifo.h
daschuer: The heap allocation happens twice. One time by the std::vector and probably a second time for the control structure. 
We may consider to use a std::array() based version where the size becomes a template parameter. 
This way we get also around of the unneccessary default initalization of the vector.

Swiftb0y: If its feasible to shift the size calculation to compile-time, I'm all for it.

src/util/fifo.h
Swiftb0y: The speed here depends entirely on the stdlib implementation. Those benchmarks have been made 9 years ago, I doubt they accurately reflect the current state of things. I think a re-evaluation is in order before making a decision based on these numbers.

src/test/fifotest.cpp
Swiftb0y: right, but if the size and the rounded up size are always the same and the implementation mistakenly uses the wrong values, the edge case is not caught. So even though it may not matter with the current implementation we should add a unit test to ensure this mistake is not made accidentally in the future.

src/util/fifo.h
Swiftb0y: fyi, A [quick and dirty quickbench](https://quick-bench.com/q/yS9ve_ZtfA3UUV0CaVPuiJY9Ack) reveals that there is zero difference (apart from the zero-sized case, in which case `std::copy` wins).

daschuer: Nice. Thank you.

src/test/fifotest.cpp
m0dB: Sure, sure, and once parametrized, we can add whatever size.

src/util/fifo.h
m0dB: Yes, I agree with Niko, my tests show the same. I find the for loop in [mixxx/src/util/sample.h](https://github.com/mixxxdj/mixxx/blob/2397016ea899185207e03912773ec43553d0c6f2/src/util/sample.h#L62) more dubious!

src/util/fifo.h
m0dB: I haven't looked at the actual use of this, so I don't know if using size as a template argument is even an option. Anyway, let's keep additional improvements beyond fixing the tsan issue for later PRs.

src/util/fifo.h
m0dB: Yes, I can write some comments about the atomics and the memory barriers.

Swiftb0y:
Agreed. This is probably a good time to post [CppCon 2014: Herb Sutter "Lock-Free Programming (or, Juggling Razor Blades)](https://www.youtube.com/watch?v=c1gO9aB9nbs)

uklotzde:
This implementation supports bulk operations and should cover most use cases: https://github.com/cameron314/concurrentqueue

With a single producer/consumer it should behave like an ordinary SPSC queue.

Instead if vendoring a second library it might be worth migrating the [`rigtorp/SPSCQueue`](https://github.com/rigtorp/SPSCQueue) code.

daschuer:
In that case let's stick with the established Portaudio version. It is exactly designed for our use case of fast bulk transfer of samples and is just missing sanitizer animations.

src/test/fifotest.cpp
m0dB: done. please check and resolve.

src/util/fifo.h
m0dB: using std::copy, please check and resolve.

src/util/fifo.h
m0dB: using std::size_t, please check and resolve.

src/util/fifo.h
m0dB: comments added. please review and resolve.

m0dB:
> In that case let's stick with the established Portaudio version. It is exactly designed for our use case of fast bulk transfer of samples and is just missing sanitizer animations.

No, please! Thread sanitizer flags it up! Sure, it will work, but it is not correct. Also, seeing that Portaudio has not been thread sanitised, I would not be surprised if the hanger I experienced comes from Portaudio itself. But I consider code that does not execute cleanly with thread sanitizer broken!

I know the pitfalls of lock-free programming very well. I am confident this implementation is correct. I am not against using other options, but I needed something that would easily adapt to the FIFO class API. None of the mentioned options fit this purpose. They are about reading / writing single items, not about adding blocks of memory and accessing regions in the ringbuffer.

m0dB:
I am going to use revert the FIFO API to use int. I am having to touch way to many files to fixes the windows compilation.

daschuer:
With force push, we need to review the whole PR over and over again. It is better to add commit that we can track the changes. If you like to have finally only one commit, feel free to squash everything before merge into the target branch.  
See also: https://github.com/mixxxdj/mixxx/blob/main/CONTRIBUTING.md

src/util/fifo.h
daschuer: I did not see comment about the memory barriers and what happens if read and write happens in between reading the two atomics. Is it not yet pushed?

m0dB:
> With force push, we need to review the whole PR over and over again. It is better to add commit that we can track the changes. If you like to have finally only one commit, feel free to squash everything before merge into the target branch. See also: https://github.com/mixxxdj/mixxx/blob/main/CONTRIBUTING.md

Yeah, sorry about that. I had all these casts added allover the code to make windows accept size_t instead of int as the return type, and I checkout just have checked out those files from 2.5 instead of rebase.

Anyway, the only things to review are fifo.h itself and the fifotest.cpp, nothing else has changed now (in other words, it really is a drop-in replacement).

src/util/fifo.h
m0dB: It is the last comment, after the declaration of the two atomics.

daschuer:
Thank you for the coment about the atomics. This puts a certein requirement on readAvailable() and writeAvailable() can you document that at the function itself? Both read two atomics. Can you also described how it is deal with the situation a read and write happens betwwen the both accesses? 

I am a bit concerned, bcause the original implementation PaUtil_FullMemoryBarrier() diffrently. 
Every memory barierre flushes the prefetched CPU cache which has perfomace implications. On the other hand a missing memory barriere will introduce broken functionallity. 

Is PaUtil_RingBuffer() broken because of this? I am in doubt, because it is established code we use without any issues for years. 
The TSAN finding are IMHO only false positive bacaus of the missing instrmentation with  __tsan_acquire` and `__tsan_release`.

On the other hand we may introduce here extra barieres that may slow down the queue unnecessarily. 
https://github.com/PortAudio/portaudio/commits/57aa393109ec996799d3a5846c9ecb0a65b64644/src/common/pa_ringbuffer.c .. has some commits regarding barrieres.

However if we are convinced it is broken actually broken, we will suffer the same issue when using it via portaudio and we should contribut a fix back. Thats the reason why need to full undertand the whole topic.

Swiftb0y:
> The TSAN finding are IMHO only false positive bacaus of the missing instrmentation with `__tsan_acquire` and `__tsan_release`.

I think the fact that these are missing is by itself an argument against using it since that indicates IMO that the code is not well maintained (also evident from the commit history). Moreover, does it also make sense to make ourselves less dependent on PortAudio if we ever want to switch to another library that does care about features such as hotplug. 

> However if we are convinced it is broken actually broken, we will suffer the same issue when using it via portaudio and we should contribut a fix back. Thats the reason why need to full undertand the whole topic.

This is a C++ implementation though and the PA code is pure C, a port wouldn't be trivial if even feasible.

m0dB:
For me the fact that thread sanitizer flags this is up is sufficient. I treat this the same as compiler warnings. One might think that a warning is harmless, and if so, even use pragmas to disable a warning. I much prefer to treat all compiler warnings as errors, and actually fix them.

I don't feel like going down the rabbit hole of investigating the PortAudio source code. I am sure this implementation is correct and it allows me to continue my thread sanitizer investigation. If you don't want to use it and prefer to stick with the original PortAudio ringbuffer, I would find that disappointing. But since it is a drop-in replacement, I will use it during my thread sanitizer investigation and that's it. And I will use it for my local builds.

As for efficiency, there is one thing missing: the atomics should be aligned with the cache size, as documented here:
https://en.cppreference.com/w/cpp/thread/hardware_destructive_interference_size , but I didn't want to add more noise to this already noisy conversation.

daschuer:
It looks like we have mixed up things in the discussion. Let's clarify this: 

1. We have discussed if the this is the right way forward to implement and probably maintain our own SCSP as proposed in this PR. I have no opinion here, but I agree that it is not an easy task. Therefor we need be extra cautiousness to make it right. The unit test are great, but they are final running on a specific metal, not on the C++ programming model that covers all our user targets. 

Regarding this topic I have some concerns whether that the memory barriers are working but not unnecessarily introduce a performance penalty. Can you please describe it as source code comment? 

2. From your posts, I got the impression that you consider the port audio implementation broken. If this is the case we have a way bigger issue, because we are rely on Portaudio a lot. Therefore we need to describe the issue to Portaudio upstream that the issue can be fixed. We can do this in a separate bug report. This also helps in 1. to show that the barriers are just right and not as wrong as in PA. 

> I treat this the same as compiler warnings. One might think that a 
warning is harmless, and if so, even use pragmas to disable a warning. I much prefer to treat all compiler warnings as errors, and actually fix them.

I am exactly the same opinion. 

> I don't feel like going down the rabbit hole of investigating the PortAudio source code. 

That's OK for me, but the one who is stepping in needs your probably your help. Is that OK? 

I have the following ideas: 
1. Add instrumentation to the PA cue like described her to check if it is really broken: https://github.com/google/sanitizers/wiki/ThreadSanitizerAtomicOperations
2. Put the new and the PA implementation into the benchmark used here: https://github.com/cameron314/concurrentqueue for my understanding both are equal fast and also "Knock-your-socks-off" 

I think 1. is easy. Just uses these defines: 
``` 
#define PaUtil_FullMemoryBarrier()  __tsan_atomic_thread_fence(memory_order_seq_cst)
#define PaUtil_ReadMemoryBarrier()  __tsan_atomic_thread_fence(memory_order_acquire)
#define PaUtil_WriteMemoryBarrier() __tsan_atomic_thread_fence(memory_order_release)
```

Can you confirm this? Do you have a CI run or something we can use for this test?  

2. Can be complex, not sure if we can drop our implementation into that test or if we need something else. Do you have ideas?

m0dB:
I added some multithread R/W tests. This is running without thread sanitizer. The variations are, for both my implementation and the PA ring buffer:

- RW : reading and writing as fast as possible (as soon as space / data is available)
- RW_Wait:  reading and writing with a fixed buffer size (waiting until buffer size is available)
- RegionRW: like RW, but instead of copying, access the ring buffer memory directly
- RegionRW_Wait: like RW_Wait, idem

The test duration serves as a simple benchmark. As you can see, the differences are negligible.

[----------] 8 tests from FifoTest
[ RUN      ] FifoTest.MultiThreadRW
[       OK ] FifoTest.MultiThreadRW (545 ms)
[ RUN      ] FifoTest.MultiThreadRW_PA
[       OK ] FifoTest.MultiThreadRW_PA (597 ms)
[ RUN      ] FifoTest.MultiThreadRW_Wait
[       OK ] FifoTest.MultiThreadRW_Wait (530 ms)
[ RUN      ] FifoTest.MultiThreadRW_PA_Wait
[       OK ] FifoTest.MultiThreadRW_PA_Wait (527 ms)
[ RUN      ] FifoTest.MultiThreadRegionRW
[       OK ] FifoTest.MultiThreadRegionRW (527 ms)
[ RUN      ] FifoTest.MultiThreadRegionRW_PA
[       OK ] FifoTest.MultiThreadRegionRW_PA (549 ms)
[ RUN      ] FifoTest.MultiThreadRegionRW_Wait
[       OK ] FifoTest.MultiThreadRegionRW_Wait (513 ms)
[ RUN      ] FifoTest.MultiThreadRegionRW_PA_Wait
[       OK ] FifoTest.MultiThreadRegionRW_PA_Wait (540 ms)
[----------] 8 tests from FifoTest (4333 ms total)


As there is no winner performance wise, I see the following ways forwards:

1. Stick with the PA ringbuffer of FIFO and
  a. Either patch it so tsan doesn't trip over it
  b. Or I use my implementation when I am running tsan (I could put if behind an ifdef which is set by cmake when building for tsan).
2. Use my implementation of FIFO anyway.

The advantage of 2 is that we don't depend on PA in case we ever want to migrate to something else (I might consider using CoreAudio directly on macOS, if I would address the hot plug issues) and that we don't have to go through the hassle of patching PA.

The advantage of 1 is that the PA ringbuffer has seen more milage than my implementation, but I am fully confident it is correct (and the tests and tsan show this).

daschuer (changes requested):

src/util/fifo.h
```diff
@@ -1,69 +1,164 @@
...
+              m_data(m_size),
+              m_writeIndex{0},
+              m_readIndex{0} {
     }
+
+    // Returns the number of values in the ringbuffer available for read
     int readAvailable() const {
-        return PaUtil_GetRingBufferReadAvailable(&m_ringBuffer);
+        const size_type readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_acquire);
```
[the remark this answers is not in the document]
m0dB: No issue. This function will only be called in the read thread, the only thread modifying the readIndex is the read thread, so its correct value is guaranteed. I will add a comment to this extend.

src/util/fifo.h
```diff
@@ -1,69 +1,164 @@
...
     }
-    int read(DataType* pData, int count) {
-        return PaUtil_ReadRingBuffer(&m_ringBuffer, pData, count);
+    // Read count values to the ring buffer. If less then count values are
+    // available, only read the available amount. The return value is the
+    // actual number of values read. If the read index reached the end of the
+    // ringbuffer, the remainder is read from the start.
+    int read(DataType* pData, size_type count) {
+        size_type readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_acquire);
```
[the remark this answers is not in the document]
m0dB: No issue, same as above.

m0dB:
> From your posts, I got the impression that you consider the port audio implementation broken. I

I consider it broken in the sense that it hasn't been updated to run correctly with thread sanitizer.

daschuer:
Did you consider to add a single option to our CMakeList.txt that adds all the required options for tsan? That would make future work with it more easy?

src/util/fifo.h
```diff
@@ -67,3 +82,162 @@ class FIFO {
...
+    // Returns the number of values in the ringbuffer available for read
+    int readAvailable() const {
+        const size_type readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_acquire);
+        return static_cast<int>(writeIndex - readIndex);
+    }
+    // Returns the space in the ringbuffer available for write
+    int writeAvailable() const {
+        const size_type readIndex = m_readIndex.load(std::memory_order_acquire);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_relaxed);
```
daschuer, line 117: Following the logic below should we swap both loads?

src/util/fifo.h
```diff
@@ -67,3 +82,162 @@ class FIFO {
...
+        return static_cast<int>(readIndex & m_mask);
+    }
+    // Same as aquireReadRegions, for write operations
+    int aquireWriteRegions(size_type count,
+            DataType** dataPtr1,
+            ring_buffer_size_t* sizePtr1,
+            DataType** dataPtr2,
+            ring_buffer_size_t* sizePtr2) {
+        const size_type readIndex = m_readIndex.load(std::memory_order_acquire);
+        size_type writeIndex = m_writeIndex.load(std::memory_order_relaxed);
```
daschuer, line 197: The same here, swap statements?

src/util/fifo.h
```diff
@@ -67,3 +82,162 @@ class FIFO {
...
+    }
+    // Same as releaseReadRegions, for write operations
+    int releaseWriteRegions(size_type count) {
+        const size_type writeIndex = (m_writeIndex.load(std::memory_order_relaxed) + count);
+        m_writeIndex.store(writeIndex, std::memory_order_release);
+        return static_cast<int>(writeIndex & m_mask);
+    }
+    // Advance the read index with count values, or maximum until the write index.
+    // Returns the new read index (wrapped inside the buffer size)
+    int flushReadData(size_type count) {
```
daschuer, line 216: Is this function allowed from reader, writer or both? Do we need a second one for the other counterpart?

src/util/fifo.h
m0dB: No, it doesn't matter. The important thing here is that the m_readIndex is loaded with std::memory_order_acquire, as it is changed in the other thread, and we want to make sure we see changed value here. m_writeIndex will not be modified in any other thread than the current thread, so it's correct value will always be guaranteed in this thread. Swapping the two loads would make no difference.

src/util/fifo.h
m0dB: Only allowed from reader, but apparently it is only used once (in ./engine/sidechain/shoutconnection.cpp). We do not need a second one for the counter part and in fact we might as well remove this one, and simply call releaseReadRegions there, which would amount to the same.

And releaseReadRegions and releaseWriteRegions would be more aptly named advanceReadIndex, advanceWriteIndex. But all of that is IMO beyond the PR.

src/util/fifo.h
```diff
@@ -67,3 +82,162 @@ class FIFO {
...
+    int writeAvailable() const {
+        const size_type readIndex = m_readIndex.load(std::memory_order_acquire);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_relaxed);
+        return static_cast<int>(m_size - (writeIndex - readIndex));
+    }
+    // Read count values to the ring buffer. If less then count values are
+    // available, only read the available amount. The return value is the
+    // actual number of values read. If the read index reached the end of the
+    // ringbuffer, the remainder is read from the start.
+    int read(DataType* pData, size_type count) {
```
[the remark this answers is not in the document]
m0dB: Sure, but that is a refactoring of this class, independent of the underlying implementation, be it PA ring buffer, or my implementation.

m0dB:
Note: As mentioned with respect to the memory ordering: the read functions should only be called in the consumer thread, and the write functions only in the producer thread.

But there is indeed nothing in the API that enforces this. But likewise, there is also nothing that protects the FIFO from being used by more than a single consumer and single producer. I can think of mechanisms to enforce this (at a cost), but I think it should be enough to simply document it. TSAN will detect such abuse anyway :-)

src/util/fifo.h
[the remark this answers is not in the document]
Swiftb0y: yeah, you're right. Lets concentrate on getting the lock-free stuff right first.

src/util/fifo.h
daschuer: Yes, so documenting this as a source comment is sufficient.

daschuer:
Here is the pa_ringbuffer with tsan functions: 
https://github.com/daschuer/mixxx/tree/pa_tsan
Would be nice if you can give this a try.

m0dB:
> Here is the pa_ringbuffer with tsan functions: https://github.com/daschuer/mixxx/tree/pa_tsan Would be nice if you can give this a try.

Thanks, I will when I have a moment!

m0dB:
#0 PaMixxx_GetRingBufferReadAvailable pa_ringbuffer.c:82 (mixxx-test:arm64+0x10192ac14)
    #1 PaMixxx_GetRingBufferReadRegions pa_ringbuffer.c:159 (mixxx-test:arm64+0x10192b114)
    #2 PaMixxx_ReadRingBuffer pa_ringbuffer.c:224 (mixxx-test:arm64+0x10192b72c)
    #3 PA::FIFO<int>::read(int*, int) fifo.h:34 (mixxx-test:arm64+0x1003947f8)
    #4 MultiThreadRW<PA::FIFO<int>>::read() fifotest.cpp:249 (mixxx-test:arm64+0x1003936b0)

  Previous write of size 4 at 0x00016b70b02c by thread T6:
    #0 PaMixxx_AdvanceRingBufferWriteIndex pa_ringbuffer.c:145 (mixxx-test:arm64+0x10192b084)
    #1 PaMixxx_WriteRingBuffer pa_ringbuffer.c:214 (mixxx-test:arm64+0x10192b6a8)
    #2 PA::FIFO<int>::write(int const*, int) fifo.h:37 (mixxx-test:arm64+0x100393860)
    #3 MultiThreadRW<PA::FIFO<int>>::write() fifotest.cpp:235 (mixxx-test:arm64+0x100393464)

WARNING: ThreadSanitizer: data race (pid=26154)
  Write of size 4 at 0x00016b70b030 by thread T7:
    #0 PaMixxx_AdvanceRingBufferReadIndex pa_ringbuffer.c:193 (mixxx-test:arm64+0x10192b444)
    #1 PaMixxx_ReadRingBuffer pa_ringbuffer.c:235 (mixxx-test:arm64+0x10192b8d0)
    #2 PA::FIFO<int>::read(int*, int) fifo.h:34 (mixxx-test:arm64+0x1003947f8)
    #3 MultiThreadRW<PA::FIFO<int>>::read() fifotest.cpp:249 (mixxx-test:arm64+0x1003936b0)

  Previous read of size 4 at 0x00016b70b030 by thread T6:
    #0 PaMixxx_GetRingBufferReadAvailable pa_ringbuffer.c:82 (mixxx-test:arm64+0x10192ac30)
    #1 PaMixxx_GetRingBufferWriteAvailable pa_ringbuffer.c:88 (mixxx-test:arm64+0x10192acc4)
    #2 PA::FIFO<int>::writeAvailable() const fifo.h:31 (mixxx-test:arm64+0x1003937e0)
    #3 MultiThreadRW<PA::FIFO<int>>::write() fifotest.cpp:231 (mixxx-test:arm64+0x100393378)

src/util/fifo.h
```diff
@@ -67,3 +82,162 @@ class FIFO {
...
+    // ringbuffer, the remainder is written to the start.
+    int write(const DataType* pData, size_type count) {
+        const size_type readIndex = m_readIndex.load(std::memory_order_acquire);
+        size_type writeIndex = m_writeIndex.load(std::memory_order_relaxed);
+        const size_type available = m_size - (writeIndex - readIndex);
+        count = std::min(available, count);
+        writeIndex = writeIndex & m_mask;
+        const size_type n = std::min(m_size - writeIndex, count);
+        std::copy(pData, pData + n, m_data.data() + writeIndex);
+        std::copy(pData + n, pData + count, m_data.data());
```
daschuer, line 148: This saves some µs 
```suggestion
        if ((count - n) > 0) {
            std::copy(pData + n, pData + count, m_data.data());
        }
```

```
Testes with Intel Core Ultra 5 125U
Before 
debug [Main] Stat("read 1","count=2714,sum=594029ns,average=218.876ns,min=29ns,max=6259ns,variance=61965ns^2,stddev=248.928ns")
debug [Main] Stat("read 65536","count=68,sum=1.38058e+06ns,average=20302.6ns,min=61ns,max=187712ns,variance=7.63838e+08ns^2,stddev=27637.6ns")
debug [Main] Stat("write 1","count=187,sum=55599ns,average=297.321ns,min=40ns,max=866ns,variance=12077.8ns^2,stddev=109.899ns")
debug [Main] Stat("write 2048","count=876,sum=1.54796e+06ns,average=1767.07ns,min=712ns,max=14696ns,variance=393266ns^2,stddev=627.109ns")

After 
debug [Main] Stat("read 1","count=2726,sum=570707ns,average=209.357ns,min=28ns,max=3766ns,variance=41299.8ns^2,stddev=203.224ns")
debug [Main] Stat("read 65536","count=78,sum=1.51157e+06ns,average=19379.1ns,min=73ns,max=184084ns,variance=7.06956e+08ns^2,stddev=26588.6ns")
debug [Main] Stat("write 1","count=183,sum=44592ns,average=243.672ns,min=70ns,max=619ns,variance=10338.9ns^2,stddev=101.681ns")
debug [Main] Stat("write 2048","count=994,sum=1.70078e+06ns,average=1711.05ns,min=625ns,max=15022ns,variance=464792ns^2,stddev=681.757ns")
``

File: src/util/fifo.h [hunk 16]
```diff
@@ -67,3 +82,162 @@ class FIFO {
...
+    // ringbuffer, the remainder is read from the start.
+    int read(DataType* pData, size_type count) {
+        size_type readIndex = m_readIndex.load(std::memory_order_relaxed);
+        const size_type writeIndex = m_writeIndex.load(std::memory_order_acquire);
+        const size_type available = writeIndex - readIndex;
+        count = std::min(available, count);
+        readIndex = readIndex & m_mask;
+        const size_type n = std::min(m_size - readIndex, count);
+        std::copy(m_data.data() + readIndex, m_data.data() + readIndex + n, pData);
+        std::copy(m_data.data(), m_data.data() + count - n, pData + n);
```
daschuer, line 132:
```suggestion
        if ((count - n) > 0) {
            std::copy(m_data.data(), m_data.data() + count - n, pData + n);
        }
```

daschuer (changes requested):

Issue #13780, opened 2024-10-18
Use of infinity() in code compiled with -ffast-math ( durationutiltest.cpp & frametest.cpp )


JoergAtGithub:
### Bug Description

If I compile with clang-tidy 18, I get the following errors, because we use infinity() in code that is compiled with option `-ffast-math`. And `-ffast-math` sets `-ffinite-math-only`.

```
 Error: /home/runner/work/mixxx/mixxx/src/test/durationutiltest.cpp:118:21: error: use of infinity is undefined behavior due to the currently enabled floating-point options [-Werror,-Wnan-infinity-disabled]
  118 |     formatTime("?", std::numeric_limits<double>::infinity());
```

```
Error: /home/runner/work/mixxx/mixxx/src/test/frametest.cpp:21:41: error: use of infinity is undefined behavior due to the currently enabled floating-point options [-Werror,-Wnan-infinity-disabled]
   21 |     EXPECT_FALSE(mixxx::audio::FramePos(std::numeric_limits<double>::infinity()).isValid());
```

```
 Error: /home/runner/work/mixxx/mixxx/src/test/frametest.cpp:57:36: error: use of infinity is undefined behavior due to the currently enabled floating-point options [-Werror,-Wnan-infinity-disabled]
   57 |             mixxx::audio::FramePos(std::numeric_limits<
      |                                    ^~~~~~~~~~~~~~~~~~~~
   58 |                     mixxx::audio::FramePos::value_t>::infinity()));
```

```
Error: /home/runner/work/mixxx/mixxx/src/test/frametest.cpp:61:22: error: use of infinity is undefined behavior due to the currently enabled floating-point options [-Werror,-Wnan-infinity-disabled]
   61 |                     -std::numeric_limits<
      |                      ^~~~~~~~~~~~~~~~~~~~
   62 |                             mixxx::audio::FramePos::value_t>::infinity()));
```

```
 Error: /home/runner/work/mixxx/mixxx/src/test/frametest.cpp:65:36: error: use of infinity is undefined behavior due to the currently enabled floating-point options [-Werror,-Wnan-infinity-disabled]
   65 |             mixxx::audio::FramePos(std::numeric_limits<
      |                                    ^~~~~~~~~~~~~~~~~~~~
   66 |                     mixxx::audio::FramePos::value_t>::infinity()));
```

### Version

main

### OS

Clang 18

Swiftb0y:
[...]

PR #13884, opened 2024-11-14
Fix undefined behaviour of infinity()

daschuer:
This fixes the undefined behviour of of `std::numeric_limits<double>::infinity())` in `-ffastmath (-ffinite-math-only)` code. By moving it to fclassify.cpp which is compiled without this flags leading to defined behavior. It makes sure that the IEEE infinity is returned which is `0x7FF0000000000000`. 

Clang 18 warns about this as `use of infinity is undefined behavior due to the currently enabled floating-point options`. This warning may indicate a real issue in case of aggressive optimization. 

This solution not only mutes this warning as `-Wnan-infinity-disabled`. We have argued that we shall not use infinity at all, but we can't entirely prevent it because all UB does not mean infinity does not happen especially when we interact with non `-ffastmath` libraries.    

Fixes: #13780

The discussion around using inf() as invalid value in this is orthogonal. We have test in place that verify that it is working for now. I consider a refactoring of it as unnecessary regression risk, but will not block an attempt. Just let's merge this as a fix for UB.

### DIFF
--- a/src/test/mathutiltest.cpp
+++ b/src/test/mathutiltest.cpp
@@ -1,6 +1,7 @@
 #include <gtest/gtest.h>
 
 #include <QtDebug>
+#include <cstring>
 
 #include "util/denormalsarezero.h"
 #include "util/fpclassify.h"
@@ -91,15 +92,14 @@ TEST_F(MathUtilTest, Denormal) {
 
 TEST_F(MathUtilTest, DoubleValues) {
     // This verifies that the infinity value can be copied into -ffastmath code
-    if (std::numeric_limits<double>::is_iec559) {
-        union {
-            long long int_value;
-            double double_value;
-        } inf_union;
-        inf_union.double_value = util_double_infinity();
-        // IEC 559 (IEEE 754) Infinity
-        EXPECT_EQ(inf_union.int_value, 0x7FF0000000000000);
-    }
+
+    // All supported targets are using IEC559 floats
+    static_assert(std::numeric_limits<double>::is_iec559);
+    long long int_value;
+    double double_value = util_double_infinity();
+    std::memcpy(&int_value, &double_value, sizeof(double_value));
+    // IEC 559 (IEEE 754) Infinity
+    EXPECT_EQ(int_value, 0x7FF0000000000000);
 }
 
 }  // namespace
--- a/src/test/mathutiltest.cpp
+++ b/src/test/mathutiltest.cpp
@@ -1,9 +1,9 @@
 #include <gtest/gtest.h>
 
 #include <QtDebug>
-#include <limits>
 
 #include "util/denormalsarezero.h"
+#include "util/fpclassify.h"
 #include "util/math.h"
 
 namespace {
@@ -51,11 +51,11 @@ TEST_F(MathUtilTest, IsNaN) {
 TEST_F(MathUtilTest, IsInf) {
     // Test floats can be recognized as infinity.
     EXPECT_FALSE(util_isinf(0.0f));
-    EXPECT_TRUE(util_isinf(std::numeric_limits<float>::infinity()));
+    EXPECT_TRUE(util_isinf(util_float_infinity()));
 
     // Test doubles can be recognized as infinity.
     EXPECT_FALSE(util_isinf(0.0f));
-    EXPECT_TRUE(util_isinf(std::numeric_limits<double>::infinity()));
+    EXPECT_TRUE(util_isinf(util_double_infinity()));
 }
 
 TEST_F(MathUtilTest, Denormal) {
--- a/src/test/mathutiltest.cpp
+++ b/src/test/mathutiltest.cpp
@@ -89,5 +89,17 @@ TEST_F(MathUtilTest, Denormal) {
 #endif
 }
 
+TEST_F(MathUtilTest, DoubleValues) {
+    // This verifies that the infinity value can be copied into -ffastmath code
+    if (std::numeric_limits<double>::is_iec559) {
+        union {
+            long long int_value;
+            double double_value;
+        } inf_union;
+        inf_union.double_value = util_double_infinity();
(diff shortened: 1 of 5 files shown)


Swiftb0y (commented):
I don't think this is a good solution (nor does it fix undefined behavior, it just silences the warnings) but I couldn't find any good resources on what is UB and what isn't in `-ffinite-math-only` to make a well founded argument against it here.

The only suggestion I would like to make here is that instead of introducing  `util_double_infinity` and making it available in source code and tests, we should consider simply removing `-ffast-math` from the tests so we can use the `<limits>` facilities as originally assumed.

daschuer:
Let me explain my view a bit more: 
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-ffinite-math-only
>Allow optimizations for floating-point arithmetic that assume that arguments and results are not NaNs or +-Infs. 

For my understanding that means, that all results from floating-point arithmetic where NaNs or +-Infs or ifs are involved are undefined behaviour. 

The compiler warns that using `std::numeric_limits<double>::infinity())` is also undefined behaviour. The movement to non ffastmath code fixes this issue effectifely without any doubts. 

Copying float values around is not floating-point arithmetic, in addition there is no gain for optimization. So we can be sure that the result is well defined even in ffastMath code. It is just copying `0x7FF0000000000000` like any other data type. 

The unittest helps to confirm that. If we follow you suggestion to not set the -ffastMath flag, the unittest doe no longer represents the Mixxx code wher the fclassify functions are used in that way. 

Changing the test falgs only affets the tests, where we see immeditaly any issues in our CI. So I would not worry that much.

Considering, this we may add a reinterprete cast test, checking the IEEE infinity raw value.

Swiftb0y:
Right, I see that. My primary problem is that adding `util_*_infinity` to produce `Inf` in `finite-math-only` code makes it too easy to produce these values in contexts where they're not expected and thus produce UB. I agree that copying these values is probably safe, but its too easy to accidentally do something unsafe.

IMO the proper solution would be to create a separate opaque type, compiled without `finite-math-only` and if you want to use infinity, you can only do it through that type. Conversion from that type to `double` would then do that check. I would offer to create this type if you agree that this is a better solution.

daschuer:
> Right, I see that. My primary problem is that adding util_*_infinity to produce Inf in finite-math-only code makes it too easy to produce these values in contexts where they're not expected and thus produce UB.

I can confirm you concerns. This shall however not block this PR, because it just fixes an UB risk that already exists. Without the newly introduced clang-18 warning it is even more easy to do it but I can not really imagine how anyone wants to do math with infinity. 
The only case I can imagine is:
```
    double a; 
    double b;
    double c = a / b;
   if (c == infinity()) {
   ...
   }
```
However in case b is 0, the value of c is already UB  before the comparison.

Swiftb0y:
> I can not really imagine how anyone wants to do math with infinity.

Right, it doesn't make sense when looking at it from a microscale, but once `infinity` passes function boundaries (which it does) its not as clear anymore. In the context of `finite-math-only` code, `infinity` needs to be treated just as careful as `nullptr` when doing operations on `pointers`. Only this time, checking for it is even more expensive...

Swiftb0y:
in practice instead of messing with this hack, we can simply replace the usage of `infinity` as a sentinel value in `FramePos` with a different (safe) sentinel value and we're good. `mixxx::Duration` is basically deprecated and from what I can tell, there are no other uses anyway...

daschuer:
We are here here in a stable branch. The work can be done in the main branch if you wish. 

However I cannot follow your reasoning. Because division by zero is always there in ffastmath it is UB and without infinity. This patch does not relate to your concerns. 

> In the context of finite-math-only code, infinity needs to be treated just as careful as nullptr when doing operations on pointers. Only this time, checking for it is even more expensive...

That is not true. Independent from finite-math-only we must not do calculations that result in infinity, because it is a dead end. It is reasonable to check devisors for zero and similar before doing the calculation and go the safe branch. 

The difference is that the compiler does not check for it in such code. It assumes inf does not exist and the result is whatever the CPU does with it. 

You see. There is no issue here and all expectations are verified by unittest.

JoergAtGithub:
> `util_*_infinity` is a prime example of what is commonly referred to as a "footgun" because its very hard to not do something stupid with it.

As we call it only from the unit test cases, why not use 0x7FF0000000000000 as expected value there directly? Than we don't need the dangling "footgun" in the code of the application.

Swiftb0y:
just document that its only supposed to be used in test code. Then I'm fine with it.

Swiftb0y (approved):
LGTM. Waiting for CI.

daschuer:
The CI Failure is fixed here: 
https://github.com/mixxxdj/mixxx/pull/13904

Swiftb0y:
yeah, should be good regardless.

src/test/mathutiltest.cpp
```diff
@@ -89,5 +89,17 @@ TEST_F(MathUtilTest, Denormal) {
...
+    if (std::numeric_limits<double>::is_iec559) {
+        union {
+            long long int_value;
+            double double_value;
+        } inf_union;
+        inf_union.double_value = util_double_infinity();
+        // IEC 559 (IEEE 754) Infinity
+        EXPECT_EQ(inf_union.int_value, 0x7FF0000000000000);
+    }
+}
```
Swiftb0y, line 103: `is_iec559` should be a static assert IMO. otherwise the test just passes when the double is not a proper float. Moreover type punning using a union like this is also highly undefined behavior! You'll need `std::bit_cast` of `std::memcpy`, those are the only ways to do type punning safely.

Swiftb0y (commented):
took another look and saw that I forgot to look at that test earlier...

Please also don't forget to remove the silencing of the clang-warning this is fixing.

Swiftb0y (approved):
LGTM. For now.

PR #13887, opened 2024-11-15
AudioUnitBackend: Initialize parameters concurrently and load music effects too

fwcd:
### Based on #13945 

This addresses a todo note in the `AudioUnitBackend` implementation:

https://github.com/mixxxdj/mixxx/blob/028e3ff4ee0ca464c7c47faa89a1c12a3e99b4e9/src/effects/backends/audiounit/audiounitbackend.mm#L81-L82

With this PR, Audio Units plugins are now instantiated concurrently and out-of-process during the initial load. This has a few advantages over the current approach of instantiating everything synchronously:

- It's faster. Since all effect manifests are to be loaded synchronously while Mixxx is starting up, doing this concurrently for all plugins potentially improves startup time.
- It's more robust. Currently a bad plugin implementation could crash Mixxx during startup. By doing this out-of-process and adding a timeout, an unstable plugin shouldn't prevent Mixxx from starting up properly.

Additionally, we now load plugins of the category `kAudioUnitType_MusicEffect` (in addition to `kAudioUnitType_Effect`), which are MIDI controllable, but otherwise function just like regular effects. This is useful, since it lets Mixxx discover plugins like ShaperBox, which are marked as music effects.

### DIFF
--- a/src/effects/backends/audiounit/audiounitbackend.mm
+++ b/src/effects/backends/audiounit/audiounitbackend.mm
@@ -18,7 +18,7 @@
 /// An effects backend for Audio Unit (AU) plugins. macOS-only.
 class AudioUnitBackend : public EffectsBackend {
   public:
-    AudioUnitBackend() : m_componentsById([[NSDictionary alloc] init]) {
+    AudioUnitBackend() : m_componentsById([[NSMutableDictionary alloc] init]) {
         loadAudioUnits();
     }
 
@@ -59,18 +59,23 @@ class AudioUnitBackend : public EffectsBackend {
     }
 
   private:
-    NSDictionary<NSString*, AVAudioUnitComponent*>* m_componentsById;
+    NSMutableDictionary<NSString*, AVAudioUnitComponent*>* m_componentsById;
     QHash<QString, EffectManifestPointer> m_manifestsById;
 
     void loadAudioUnits() {
         qDebug() << "Loading audio units...";
 
+        loadAudioUnitsOfType(kAudioUnitType_Effect);
+        loadAudioUnitsOfType(kAudioUnitType_MusicEffect);
+    }
+
+    void loadAudioUnitsOfType(OSType componentType) {
         // See
         // https://developer.apple.com/documentation/audiotoolbox/audio_unit_v3_plug-ins/incorporating_audio_effects_and_instruments?language=objc
 
         // Create a query for audio components
         AudioComponentDescription description = {
-                .componentType = kAudioUnitType_Effect,
+                .componentType = componentType,
                 .componentSubType = 0,
                 .componentManufacturer = 0,
                 .componentFlags = 0,
@@ -85,10 +90,6 @@ class AudioUnitBackend : public EffectsBackend {
         auto components = [manager componentsMatchingDescription:description];
 
         // Assign ids to the components
-        NSMutableDictionary<NSString*, AVAudioUnitComponent*>* componentsById =
-                [[NSMutableDictionary alloc] init];
-        QHash<QString, EffectManifestPointer> manifestsById;
-
         for (AVAudioUnitComponent* component in components) {
             qDebug() << "Found audio unit" << [component name];
 
@@ -97,13 +98,10 @@ class AudioUnitBackend : public EffectsBackend {
                               [component manufacturerName],
                               [component name],
                               [component versionString]]);
-            componentsById[effectId.toNSString()] = component;
-            manifestsById[effectId] = EffectManifestPointer(
+            m_componentsById[effectId.toNSString()] = component;
+            m_manifestsById[effectId] = EffectManifestPointer(
                     new AudioUnitManifest(effectId, component));
         }
-
-        m_componentsById = componentsById;
-        m_manifestsById = manifestsById;
     }
 };
 
--- a/src/effects/backends/audiounit/audiounitbackend.mm
+++ b/src/effects/backends/audiounit/audiounitbackend.mm
@@ -126,11 +126,11 @@ class AudioUnitBackend : public EffectsBackend {
             });
         }
 
-        int64_t timeoutMs = 10000;
+        const int64_t TIMEOUT_MS = 6000;
 
         qDebug() << "Waiting for audio unit manifests to be loaded...";
         if (dispatch_group_wait(group,
-                    dispatch_time(DISPATCH_TIME_NOW, timeoutMs * 1000000)) ==
+                    dispatch_time(DISPATCH_TIME_NOW, TIMEOUT_MS * 1000000)) ==
                 0) {
             qDebug() << "Successfully loaded audio unit manifests";
         } else {
--- a/src/effects/backends/audiounit/audiounitbackend.mm
+++ b/src/effects/backends/audiounit/audiounitbackend.mm
@@ -3,10 +3,12 @@
 #import <AVFAudio/AVFAudio.h>
 #import <AudioToolbox/AudioToolbox.h>
 #import <Foundation/Foundation.h>
+#import <dispatch/dispatch.h>
 
 #include <QDebug>
 #include <QHash>
 #include <QList>
+#include <QMutex>
 #include <QString>
 #include <memory>
 
@@ -14,6 +16,7 @@
 #include "effects/backends/audiounit/audiouniteffectprocessor.h"
 #include "effects/backends/audiounit/audiounitmanifest.h"
 #include "effects/defs.h"
+#include "util/compatibility/qmutex.h"
 
 /// An effects backend for Audio Unit (AU) plugins. macOS-only.
 class AudioUnitBackend : public EffectsBackend {
@@ -61,6 +64,7 @@ class AudioUnitBackend : public EffectsBackend {
   private:
     NSMutableDictionary<NSString*, AVAudioUnitComponent*>* m_componentsById;
     QHash<QString, EffectManifestPointer> m_manifestsById;
+    QMutex m_mutex;
 
     void loadAudioUnits() {
         qDebug() << "Loading audio units...";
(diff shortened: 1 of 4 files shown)


Swiftb0y:
I'm unsure who can properly review this. @m0dB has some apple-specific experience, right?

m0dB:
Yes, I am very familiar with AU and I will review this when I find the time.

fwcd:
I've rebased this to 2.5 since I believe the improvements in stability would be worth having there too. The GUI stuff (#13888) can then go to 2.6.

Swiftb0y:
Not sure if its worth the regression risk. I'll leave that decision to @m0dB

fwcd:
I ran into occasional crashes with ShaperBox at startup, the only reason why it didn't affect the release build was that "luckily" that plugin was marked as a music effect. Since it could affect other (regular effect) plugins though, I think instantiating everything out-of-process would be worth it 😅

src/effects/backends/audiounit/audiounitbackend.mm
```diff
@@ -59,51 +62,80 @@ bool canInstantiateEffect(const QString& effectId) const override {
...
+
+                // Register manifest
+                auto locker = lockMutex(&m_mutex);
+                m_manifestsById[effectId] = manifest;
+            });
         }
 
-        m_componentsById = componentsById;
-        m_manifestsById = manifestsById;
+        int64_t timeoutMs = 10000;
```
m0dB, line 129: This seems quite long for a timeout. Have you found that loading the manifests takes this long?

src/effects/backends/audiounit/audiounitmanifest.mm
```diff
@@ -18,9 +20,25 @@
     setDescription(QString::fromNSString([component typeName]));
     setAuthor(QString::fromNSString([component manufacturerName]));
 
-    // Try instantiating the unit in-process to fetch its properties quickly
+    // Instantiate audio unit (out-of-process) to load parameters
+    AudioUnitManager manager{component};
+
+    const int TIMEOUT_MS = 5000;
```
m0dB, line 26: Same question here.

m0dB:
Apart from my objection against the use of a sleep in a while loop instead of a wait condition, this LGTM.

src/effects/backends/audiounit/audiounitbackend.mm
fwcd: Not usually, I figured being a bit conservative here might be useful for slower machines, but I'd definitely be open to lower this.

I would, however, still make this timeout (which is for all plugins) a bit higher than the other one, in case GCD for some reason exhausted its thread pool, e.g. if the user has a lot (> 64) of plugins installed.

fwcd:
The [test failures](https://github.com/mixxxdj/mixxx/actions/runs/11881516836/job/33105946479?pr=13887) confuse me:

image

Did I miss some race condition somewhere? Is this caused by this PR or did these happen elsewhere already?

fwcd:
I don't quite understand why we get sporadic segfaults with the `QWaitCondition`-based implementation, unless I misunderstand how the wait condition is supposed to be used, the only explanation I see is that the wait condition doesn't like being invoked from some possibly GCD-managed background thread where our callback may be invoked, but that would kind of defeat the purpose of the wait condition and its associated mutex in the first place.

Any ideas? The sleep loop seemed to work more reliably, maybe it would be less risky to stick to that for now?

m0dB:
Hm, let me quickly run this in the debugger if I can see what goes wrong. But yeah, let's stick with the while sleep loop otherwise.

m0dB:
Please feel free to revert the QWaitCondition to the while sleep loop; I have a local branch with the version that crashes so I can debug this.

m0dB:
Hm, mixxx-test freezes on me at:

```
[ RUN      ] AutoDJProcessorTest.FullIntroOutro_LongerIntro

GMOCK WARNING:
Uninteresting mock function call - returning directly.
    Function call: emitAutoDJStateChanged(0)
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/main/docs/gmock_cook_book.md#knowing-when-to-expect-useoncall for details.

GMOCK WARNING:
Uninteresting mock function call - returning directly.
    Function call: emitAutoDJStateChanged(1)
NOTE: You can safely ignore the above warning unless this call should not happen.  Do not suppress it by blindly adding an EXPECT_CALL() if you don't mean to enforce the call.  See https://github.com/google/googletest/blob/main/docs/gmock_cook_book.md#knowing-when-to-expect-useoncall for details.
```

Have to force quit.

m0dB:
Some happens with your branch... If I skip the hanging test, it hangs on
```
[ RUN      ] ControllerScriptEngineLegacyTest.trigger
 test:warning [Main] "Script tried to connect to ([Test], co) using `connectControl` which is deprecated. Use `makeConnection` instead!"
```

I am going to try with 2.5, maybe something is messed up with my build environment.

Swiftb0y:
> the only explanation I see is that the wait condition doesn't like being invoked from some possibly GCD-managed background thread where our callback may be invoked,

Slightly related. I once hit a very nasty bug when calling `pthread_cond_signal` from a unix signal handler. Someone else fortunately figured out that that is apparently undefined behavior (though only resulted in a deadlock in 1/1000 cases). We fixed it by spinning on an atomic shared variable instead (which I'm **not** recommending here). I don't kow how the GCD stuff is implemented, but something similar may be the case here.

src/effects/backends/audiounit/audiounitbackend.mm
```diff
@@ -59,51 +62,80 @@ bool canInstantiateEffect(const QString& effectId) const override {
...
+            });
         }
 
-        m_componentsById = componentsById;
-        m_manifestsById = manifestsById;
+        const int64_t TIMEOUT_MS = 6000;
+
+        qDebug() << "Waiting for audio unit manifests to be loaded...";
+        if (dispatch_group_wait(group,
+                    dispatch_time(DISPATCH_TIME_NOW, TIMEOUT_MS * 1000000)) ==
```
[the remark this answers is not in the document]
Swiftb0y: oops, my bad. I thought this was the place where the earlier comments where referring to, not the `QThread::sleep` in `AudioUnitManifest`.

src/effects/backends/audiounit/audiounitmanager.mm
```diff
@@ -17,6 +19,8 @@
         return;
     }
 
+    dispatch_group_enter(m_instantiationGroup);
+
```
Swiftb0y, line 23: the doc says `dispatch_group_enter` must balanced with a `dispatch_group_leave` call. This is only the case in the `instantiateAudioUnitSync` case, not in the `instantiateAudioUnitAsync` case. Is this a bug or is there a reason for that?

src/effects/backends/audiounit/audiounitmanager.mm
```diff
@@ -114,4 +118,10 @@
 
     m_audioUnit = audioUnit;
     m_isInstantiated.store(true);
+    dispatch_group_leave(m_instantiationGroup);
```
fwcd, line 121: Both code paths end here

src/effects/backends/audiounit/audiounitmanager.mm
fwcd: In both cases the `instantiateWith` method is called, which includes the leave call (see below). In the async case `initializeWith` is called from the completion callback.

fwcd:
I suppose there must be a logic error somewhere. The [dispatch group-based implementation](https://github.com/fwcd/mixxx/tree/au-music-effects-dispatchgroup) seems to have [the same issue](https://github.com/mixxxdj/mixxx/actions/runs/11918594956/job/33216383461?pr=13887) as the `QWaitCondition`-based implementation.

I have reset the branch back to the sleep loop implementation (for some reason GitHub is still processing the push, not sure why the updated commits don't show up).

fwcd:
Interestingly, I've noticed that this fixes a few plugins that don't seem to like being instantiated in-process (e.g. Endless Smile, which doesn't provide any parameters when used in-process).

Unfortunately, testing the GitHub Actions build from this branch I also ran into occasional crashes on startup that were clearly plugin-related, e.g. this one: [au-crash.txt](https://github.com/user-attachments/files/17922295/au-crash.txt)

Maybe we should indeed defer this PR to `main`, just to give it a bit more time for ironing out the bugs.

src/effects/backends/audiounit/audiounitmanifest.mm
```diff
@@ -18,9 +20,18 @@
...
+    AudioUnitManager manager{component};
+
+    const int TIMEOUT_MS = 2000;
+    if (!manager.waitForAudioUnit(TIMEOUT_MS)) {
+        qWarning() << name() << "took more than" << TIMEOUT_MS
+                   << "ms to initialize, skipping manifest initialization "
+                      "for this effect. This means this effect will not "
+                      "display any parameters and likely not be useful!";
+        return;
+    }
```
fwcd, line 33: I just had another idea what the issue could be... if we bail out here (due to a timeout), we deinitialize the `AudioUnitManager` instance early, which may cause the async initialization callback to get a dangling (captured) `this` pointer

fwcd: I can reproduce the crash relatively reliably by lowering the timeout, so could indeed be it...

fwcd: Fixed in 10140c9

fwcd:
Always wrapping `AudioUnitManager` instances in a shared pointer and then passing that to the Audio Unit initialization callback seems to fix the dangling `this` issues during async initialization as described in https://github.com/mixxxdj/mixxx/pull/13887#discussion_r1859128871. The dispatch group-based implementation of waiting now also seems to work fine, so I replaced the sleep loop with it.

Note that this shared pointer refactoring (10140c9) is effectively a _bug fix_ making async initialization of Audio Units safe. On the current 2.5/main branch we instantiate Audio Units synchronously for manifests (where this is not an issue, since there is no callback), but asynchronously for processing, so there may be a minor race condition if the user happens to switch effects very very quickly. Since this PR makes _all_ initializations (including the manifest ones for parsing out the parameters) asynchronous, the bug/race condition simply became more likely to occur.

In principle the PR should be ready for re-review. I wonder if it would make sense to split out the shared pointer refactoring to a PR targeting 2.5 (while leaving all of the sync -> async migration stuff here). @m0dB What do you think?

fwcd:
I have split out the shared pointer-refactoring into #13945, which IMO is an important bug fix for 2.5, and rebased the other PRs, so we can focus on the async initialization here and on GUIs in #13888, respectively.

m0dB:
Thanks for splitting out the shared pointer refactoring, I agree it makes sense to include in 2.5.

As for this PR, it looks safe to me, and I don't see a reason not to include it in 2.5, except for the limited testing exposure this has had...

JoergAtGithub:
@fwcd Please rebase to main and fix the merge conflicts

src/effects/backends/audiounit/audiounitbackend.mm
JoergAtGithub, line 133:
```suggestion
                    dispatch_time(DISPATCH_TIME_NOW, TIMEOUT_MS * 1000)) ==
```
I guess MS means milli-seconds and not micro-seconds.

fwcd: Yes, it means milliseconds, but `dispatch_time` takes nanoseconds: https://developer.apple.com/documentation/dispatch/1420519-dispatch_time, hence the factor 1 000 000

JoergAtGithub:
LGTM! Thank you!

PR #13888, opened 2024-11-15
Effects: Add support for GUIs and reference implementation for Audio Unit

### LINKED ISSUE #13147: Wishlist: Support Audio Unit GUIs
### Feature Description

#12112 implements basic support for Audio Unit effects on macOS, but (similar to #10695) it would be very nice if Mixxx could host the plugin's GUI too, e.g. in a pop up window like the keywheel.

fwcd:
### Based on #13887, fixes #13147

This PR adds a generic interface for effect backends for exposing GUIs in the form of `QDialog`s, along with a reference implementation for Audio Unit effects on macOS[^1]:

image

Effect UIs are toggled via a new button, implemented in LateNight/Palemoon (the only skin for now) and shown for effects that provide a UI:

image

image

These can be hidden via the skin settings to reduce UI clutter, if not needed:

image

### To do

- Generic implementation
  - [x] Add button for toggling effect UIs
  - [x] Add `DlgEffect` abstraction that sets up a generic effect UI (a floating dialog)
  - [x] Update the `ui_shown` control object when the window is closed manually
  - [x] Make sure that effect UIs are also closed upon closing the main window
  - [x] Figure out if we can gray out/hide the button for toggling the UI if the effect doesn't support UIs. `EffectManifest` now exposes a `hasUI` property, which may be helpful for that.
- Audio Unit implementation
  - [x] Host plugin views by loading them from the corresponding bundle
  - [x] Fall back to generic view if the plugin does not provide a custom UI
  - [x] Automatically size the dialog window correctly

### Potentially out-of-scope

- Fix sampling rate mismatch between Audio Unit plugins and Mixxx
  - See https://github.com/mixxxdj/mixxx/pull/13888#issuecomment-2477938796
- Update the parameter knobs on the Mixxx side after changing them from the GUI (the other direction already works and the effect also updates as intended from the GUI, this seems to be purely cosmetic)
- Sync tempo/beat with AU effects
  - Currently plugins that operate primarily in beat domain (such as ShaperBox) are not super useful, since they can't use the correct tempo (and their default tempo is almost always misaligned)
  - See e.g. https://forums.developer.apple.com/forums/thread/46085

Feedback welcome!

[^1]: Support for UIs of other effect backends like LV2 (or VST plugins once we implement #9019) could be added in future PRs as needed.

### DIFF
--- a/src/effects/backends/audiounit/audiouniteffectprocessor.h
+++ b/src/effects/backends/audiounit/audiouniteffectprocessor.h
@@ -4,8 +4,10 @@
 #import <AudioToolbox/AudioToolbox.h>
 #import <CoreAudioTypes/CoreAudioTypes.h>
 
+#include <QDialog>
 #include <QList>
 #include <atomic>
+#include <memory>
 
 #include "audio/types.h"
 #include "effects/backends/audiounit/audiounitmanager.h"
@@ -53,6 +55,8 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
             const mixxx::EngineParameters& engineParameters,
             const EffectEnableState enableState,
             const GroupFeatureState& groupFeatures) override;
+
+    std::unique_ptr<QDialog> createUI() override;
 
   private:
     AudioUnitManagerPointer m_pManager;
--- a/src/effects/backends/audiounit/audiouniteffectprocessor.h
+++ b/src/effects/backends/audiounit/audiouniteffectprocessor.h
@@ -70,5 +70,5 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
     void syncParameters();
     void syncStreamFormat(const mixxx::EngineParameters& engineParameters);
 
-    NSView* _Nullable createNativeUI(CGSize size);
+    NSView* _Nullable createNativeUI(CGSize size, QString& outError);
 };
--- a/src/effects/backends/audiounit/audiouniteffectprocessor.h
+++ b/src/effects/backends/audiounit/audiouniteffectprocessor.h
@@ -69,6 +69,4 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
 
     void syncParameters();
     void syncStreamFormat(const mixxx::EngineParameters& engineParameters);
-
-    NSView* _Nullable createNativeUI(CGSize size, QString& outError);
 };
--- a/src/effects/backends/audiounit/audiouniteffectprocessor.h
+++ b/src/effects/backends/audiounit/audiouniteffectprocessor.h
@@ -5,7 +5,6 @@
 #import <AudioToolbox/AudioToolbox.h>
 #import <CoreAudioTypes/CoreAudioTypes.h>
 
-#include <QDialog>
 #include <QList>
 #include <atomic>
 #include <memory>
@@ -13,6 +12,7 @@
 #include "audio/types.h"
 #include "effects/backends/audiounit/audiounitmanager.h"
 #include "effects/backends/effectprocessor.h"
+#include "effects/dlgeffect.h"
 #include "engine/engine.h"
 
 class AudioUnitEffectGroupState final : public EffectState {
@@ -57,7 +57,7 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
             const EffectEnableState enableState,
             const GroupFeatureState& groupFeatures) override;
 
-    std::unique_ptr<QDialog> createUI() override;
+    std::unique_ptr<DlgEffect> createUI() override;
 
   private:
     AudioUnitManagerPointer m_pManager;
--- a/src/effects/backends/audiounit/audiouniteffectprocessor.h
+++ b/src/effects/backends/audiounit/audiouniteffectprocessor.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #import <AVFAudio/AVFAudio.h>
+#import <AppKit/AppKit.h>
 #import <AudioToolbox/AudioToolbox.h>
 #import <CoreAudioTypes/CoreAudioTypes.h>
 
@@ -68,4 +69,6 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
 
     void syncParameters();
     void syncStreamFormat(const mixxx::EngineParameters& engineParameters);
+
+    NSView* _Nullable createNativeUI(CGSize size);
 };
--- a/src/effects/backends/audiounit/dlgaudiounit.mm
+++ b/src/effects/backends/audiounit/dlgaudiounit.mm
@@ -131,8 +131,8 @@ DlgAudioUnit::DlgAudioUnit(AudioUnitManagerPointer pManager) {
     // Style effect UI as a floating, but non-modal, HUD window
     NSWindow* dialogWindow = [dialogView window];
     [dialogWindow setStyleMask:NSWindowStyleMaskTitled |
-            NSWindowStyleMaskClosable | NSWindowStyleMaskResizable |
-            NSWindowStyleMaskUtilityWindow | NSWindowStyleMaskHUDWindow];
+            NSWindowStyleMaskClosable | NSWindowStyleMaskUtilityWindow |
+            NSWindowStyleMaskHUDWindow];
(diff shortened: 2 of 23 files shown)


fwcd:
I wonder why there a bunch of frequencies around 10k are missing according to spectrum analyzers in plugins. For comparison, here's some white noise in both Logic and Mixxx, viewed with the TDR Prism spectrum analyzer:

| Logic | Mixxx |
| - | - |
|  | ! |

Notably, this does not show up when I open the recorded mix in Logic:



I wonder if there's something wrong in the effect implementation (passing the audio buffer in some strange way) or whether there's something elsewhere in the Mixxx engine that could be causing this?

src/effects/backends/effectprocessor.h
```diff
@@ -148,6 +155,20 @@ class EffectProcessorImpl : public EffectProcessor {
...
+
+        QVBoxLayout* layout = new QVBoxLayout();
+        layout->addWidget(label);
+
+        std::unique_ptr<QDialog> dialog = std::make_unique<QDialog>();
+        dialog->setWindowTitle("No UI available");
+        dialog->setLayout(layout);
+
+        return dialog;
+    }
```
Swiftb0y, line 170: is this really needed? This should only be shown when a plugin has `hasUI` is true even though it shouldn't be right?

src/engine/effects/engineeffect.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <QDialog>
```
Swiftb0y, line 3: forward decl suffices actually

src/effects/backends/audiounit/audiouniteffectprocessor.h
```diff
@@ -70,5 +70,5 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
     void syncParameters();
     void syncStreamFormat(const mixxx::EngineParameters& engineParameters);
 
-    NSView* _Nullable createNativeUI(CGSize size);
+    NSView* _Nullable createNativeUI(CGSize size, QString& outError);
```
Swiftb0y, line 73: consider returning a `std::variant<NSView* _Nullable, QString>` (as in for example Rusts `std::Result` type). This would nicely map to `std::expected`.

Swiftb0y (commented):
couple surface level comments. We should consider creating more C++like friendly wrappers for the underlying native C calls...

src/effects/backends/audiounit/audiouniteffectprocessor.h
fwcd: I wasn't sure how well that meshes with the ObjC type, since there's reference counting and stuff involved, but I'll give that a try

src/effects/backends/effectprocessor.h
fwcd: Exactly, I haven't gotten around to do that check yet and decided that this was nicer than an empty dialog 😄 

I've also considered dropping the `hasUI` and just returning an optional dialog here, but that wouldn't give the UI a chance to hide/style the UI button beforehand.

src/effects/backends/audiounit/audiouniteffectprocessor.h
Swiftb0y: right... yeah idk enough about that to evaluate how feasible it is. I just assumed C++ semantics hold. If you see more value in a different approach then use that.

src/effects/backends/effectprocessor.h
Swiftb0y: yeah... I think hiding the button if not available would be definitely good to have. I'd prefer to keep that if we can.

src/engine/effects/engineeffect.h
fwcd: So just `class QDialog;`? Are forward decls generally preferred for headers?

src/effects/backends/audiounit/dlgaudiounit.mm
```diff
@@ -0,0 +1,159 @@
...
+
+        // Automatically resize the dialog to fit the view after layout
+        [audioUnitView setPostsFrameChangedNotifications:YES];
+        [[NSNotificationCenter defaultCenter]
+                addObserverForName:NSViewFrameDidChangeNotification
+                            object:audioUnitView
+                             queue:[NSOperationQueue mainQueue]
+                        usingBlock:^(NSNotification* notification) {
+                            NSView* audioUnitView =
+                                    (NSView*)notification.object;
```
fwcd, line 141: Small implementation note: We deliberately extract the view from the notification again to avoid creating a potential retain cycle by capturing the view again

src/effects/backends/effectprocessor.h
fwcd: One consequence is that we keep a dialog open when the user switches to an effect that doesn't provide a UI. On the one hand I think this could be nice (e.g. the user might arrange their effect UIs on the screen in a particular way and it might be nice to keep them open until explicitly closed). On the other hand, the blanket implementation doesn't really do anything useful, so we might as well also just close the UI, in which case the next effect with a UI will automatically open the dialog since the control object enabling the UI is still toggled on.

Don't really have a preference what the best/most intuitive behavior is here. Any thoughts?

src/engine/effects/engineeffect.h
Swiftb0y: its a tradeoff IMO. If the header is big (usually the case with Qt headers) and the header where its included is used in lots of other headers and you only heap-allocate the class anyways (with a unique_ptr for example), then I'd forward-declare them. In this case it would make sense because you really only need to name it in order to return from a single member function. So pulling in an entire header does come with quite some bloat IMO.

src/effects/backends/effectprocessor.h
Swiftb0y: I think we should follow the approach DAWs usually take. You can open as many effect UIs as there are effects and if you replace one effect with one that doesn't have a window or you remove the effect, the window just closes. Now you usually don't replace effects in DAWs directly (you insert a new one and remove the old one) but in mixxx we can replace an effect directly since we have slots. In that case we can consider replacing the old UI with the new UI directly (if the old effect had its UI open). Wdyt?

fwcd: > Now you usually don't replace effects in DAWs directly

Depends on the DAW, at least in Logic the workflow is similar to Mixxx, where effects are essentially also slots with a dropdown.

> in mixxx we can replace an effect directly since we have slots. In that case we can consider replacing the old UI with the new UI directly (if the old effect had its UI open). Wdyt?

Sounds reasonable. The way the API is designed currently, we'd still instantiate a new dialog (the new effect UI might want to set its own title/resizability/etc.), but I was thinking of maybe just carrying over the position manually.

Swiftb0y: > Depends on the DAW, at least in Logic the workflow is similar to Mixxx, where effects are essentially also slots with a dropdown.

ah yeah. My primary experience lies with bitwig and ableton... 

> but I was thinking of maybe just carrying over the position manually.

That may not be feasible on all platforms (not particularly relevant for MacOS I guess, but I don't think we could implement the same on wayland for example).

fwcd:
Marking this one as a draft until we figure out

- #13887

JoergAtGithub:
If you consider "Update the parameter knobs on the Mixxx side after changing them from the GUI" out of scope of this PR, than please create an issue with target `2.6-beta` for this!
Please consider also, what this means for controller mappings with soft-takeover for the parameter knobs.

src/effects/backends/effectprocessor.h
```diff
@@ -148,6 +155,19 @@ class EffectProcessorImpl : public EffectProcessor {
         return 0;
     }
 
+    std::unique_ptr<DlgEffect> createUI() override {
```
JoergAtGithub, line 158: This seems wrong, as you declare this as pure virtual in this same file line 119: `virtual std::unique_ptr<DlgEffect> createUI() = 0;`

I would expect the GUI button disabled instead of showing a dialog that this effect has no GUI.

fwcd: > This seems wrong, as you declare this as pure virtual in this same file line

The pure virtual signature applies to `EffectProcessor`, which serves as the interface, this method is in `EffectProcessorImpl`

> I would expect the GUI button disabled instead of showing a dialog that this effect has no GUI.

We do that too, but there are other advantages to keeping the dialog available, see this discussion for rationale: https://github.com/mixxxdj/mixxx/pull/13888#discussion_r1843617674

Issue #13885, opened 2024-11-14
2.5 and main crash on shutdown


ronso0:
### Bug Description

I'm getting this crash on 2.5 and main 359c40eb82
* `--developer`
* with Qt 6.2.3 and
* the legacy skin (backtrace below)

Did git bisect for some hours and it seems that it's caused by https://github.com/mixxxdj/mixxx/commit/4ba370b5cabf9a7e75bb45d107a3ed54871c80cb  (or rather that/2.5/2.4 being merged to main?).
https://github.com/mixxxdj/mixxx/blob/028e3ff4ee0ca464c7c47faa89a1c12a3e99b4e9/src/mixxxapplication.cpp#L202-L212

If I disable the logging everything shuts down cleanly.

The crash happens when trying to with `target->metaObject()` for
`MixxxMainWindow(0x561df6faa0d0, ...) QChildEvent(ChildRemoved, QWidget(0x5605e0e18530) name="Mixxx")`
= when the skin widget has been deleted.

With some debug logging added in this method (which expectedly slows everything down significantly) I get the crash even for
`MixxxApplication::notify: LaunchImage(0x5605e1028700) QChildEvent(ChildRemoved, QObject(0x5605e0e18530))`
Later on, with simple logging only in the 'long time' branch, I get it for
`QWidgetWindow(0x561df6faa0d0, name = "MixxxMainWindowClassWindow") QCloseEvent(Close, 0x7ffd1966c8a0)`

### Conclusion?
The target has been deleted after `QApplication::notify(target, event);`
Hence even the additional condition for logging
`target->metaObject() != nullptr`
won't work.
Could it be that the logging itself causes the delay which in turn causes the crash??

<details><summary>Backtrace of thread #1</summary>
debug [Main] ~DlgAutoDJ()
Thread 1 "mixxx" received signal SIGSEGV, Segmentation fault.
0x0000046200000000 in ?? ()

Thread 1 (Thread 0x7fffeade0100 (LWP 1002332) "mixxx"):
#0  0x0000046200000000 in  ()
#1  0x00005555560c4e9b in MixxxApplication::notify(QObject*, QEvent*) (this=<optimized out>, target=0x5555797d84d0, event=<optimized out>) at /src/mixxxapplication.cpp:206
        time = {m_startTime = std::chrono::_V2::steady_clock time_point = { 304985613821353ns }, static kStoppedTimerValue = {__d = std::chrono::duration = { -9223372036854775808ns }}}
[...]

Issue #8356, opened 2022-08-22
Display MIDI Through device in non-developer mode


mixxxbot:
Reported by: **[Pegasus-RPG](https://github.com/Pegasus-RPG)**
Date: 2015-12-03T17:28:19Z
Status: In Progress
Importance: Low
Launchpad Issue: [lp1522505](https://bugs.launchpad.net/bugs/1522505)
Tags: midi

---

There is a use case for regular users having access to the MIDI Through devices: controlling lighting software. There needs to be a way to enable them without engaging full developer mode.

PR #13889, opened 2024-11-15
Fix crash due to dangling pointer access in developer mode.

daschuer:
This fixes a regression #13885 since https://github.com/mixxxdj/mixxx/pull/12094

### DIFF
--- a/src/mixxxapplication.cpp
+++ b/src/mixxxapplication.cpp
@@ -140,7 +140,7 @@ void MixxxApplication::registerMetaTypes() {
     qRegisterMetaType<mixxx::FileInfo>("mixxx::FileInfo");
 }
 
-bool MixxxApplication::notify(QObject* target, QEvent* event) {
+bool MixxxApplication::notify(QObject* pTarget, QEvent* event) {
 #if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
     // All touch events are translated into two simultaneous events: one for
     // the target QWidgetWindow and one for the target QWidget.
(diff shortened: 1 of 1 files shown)


src/mixxxapplication.cpp
```diff
@@ -140,7 +140,7 @@ void MixxxApplication::registerMetaTypes() {
     qRegisterMetaType<mixxx::FileInfo>("mixxx::FileInfo");
 }
 
-bool MixxxApplication::notify(QObject* target, QEvent* event) {
+bool MixxxApplication::notify(QObject* pTarget, QEvent* event) {
```
ronso0, line 143: what about pEvent, if you already touch this?

daschuer:
Done

Swiftb0y (approved):
LGTM. Thank you.

JoergAtGithub:
@daschuer Thank you for jumping in! The fix LGTM and I just tested it on Win11 -> Works as expected!

ronso0:
Thanks @daschuer for quick fix!
Can confirm it doesn't crash anymore.

However, I want to note that I saw the crash for other event types, too, eg. `QChildEvent(ChildRemoved, ..)`.
So, presuming 
* #12094 is for diagnosis only
* we can't be sure it happens only for `QEvent::DeferredDelete`
* a crash can cause losing data from the current session
* users may run Mixxx with --developer in regular session to have MIDI Through for MIDI4Light

shall we tie to another developer option, eg. a certain log level?

JoergAtGithub:
> However, I want to note that I saw the crash for other event types, too, eg. `QChildEvent(ChildRemoved, ..)`.

Than I see only two options:

1. Cache the object informantion before calling notify, which will affect the general performance in developer mode
2. Remove the printout of the target object information and just print the information about the event type

ronso0:
3. implementing #8356 would "fix" it IMO since this eliminates the only reason I know to run Mixxx with --developer in regular sessions

PR #13896, opened 2024-11-17
Make extended controller information available for device selection

JoergAtGithub:
This is the first PR, where I make use of new hidapi capabilities. It provides all available information that is useful for device selection to the user/mapping developer. In follow-ups I will use this foundation to dig deeper in the provided Plug&Play information about collections and report data structures.

This PR:
- Show the physical transport protocol if available USB, Bluetooth, Firewire, ...
- Shows different icons for MIDI, HID and USB Bulk-Controllers
- Clearly seperates device properties and mapping poperties in the controller preferences
- Replaces the hardcoded constant for HID Usages by the official JSON reference file provided on usb.org
- Hides all properties, where no data are available (e.g. USB Interface-No. for a Portmidi-Controller).
- All Properties have human readable labels now, instead of the old concatenated device category string
- Devendored the outdated hidapi 0.11.1 copy in the lib folder - which is much older than required by now.

Note:

- The number of lines changed lines looks much bigger as it is, due to devendoring hidapi and adding the Usage JSON file from usb.org.
- I grouped the changes in seperated commits for easier review

grafik
grafik
grafik


src/controllers/hid/hidusagetables.h
Swiftb0y, line 1: I'm not a fan of having this big json in the tree but then loading and parsing it at runtime all the time, I'd prefer to move this to compile-time (ideally even before). Have you measured the `HidUsageTables()` ctor and `getUsagePageDescription()` already by any chance? That would also eliminate some performance-overhead from the Qt classes and lifetime issues since all the data could be made static. Do you agree that would make sense?

Swiftb0y (commented):
thank you very much for this valuable PR. I only have one high-level complaint. Codestyle nitpicks follow in a separate review.

src/controllers/controller.h
```diff
@@ -9,6 +9,21 @@
...
 class ControllerScriptEngineLegacy;
 
+enum class PhysicalTransportProtocol : uint8_t {
+    UNKNOWN,
+    USB,
+    BlueTooth,
+    I2C,
+    SPI,
+    FireWire // IEEE 1394
+};
```
Swiftb0y, line 19: in the later commits, you're often resolving this to strings via a bunch of different techniques (switch-case or temporary map). May suggest you unify these different occurences?

src/controllers/dlgprefcontrollerdlg.ui
```diff
@@ -53,6 +53,425 @@
         </widget>
        </item>
        <item row="10" column="0" colspan="2">
+        <layout class="QGridLayout" name="gridLayout_8" rowstretch="0">
+         <item row="0" column="0">
+          <widget class="QGroupBox" name="groupBoxDeviceInfo">
+           <property name="sizePolicy">
+            <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+             <horstretch>0</horstretch>
```
Swiftb0y, line 61: are these device infos visible all the time? Are the relevant to all users all the time? Would it make sense to hide it behind `--developer` or `--controller-debug`?

Swiftb0y (commented):
couple nitpicks

src/controllers/dlgprefcontrollerdlg.ui
ronso0: --controllerDebug and later an always visible, collapsible region?

JoergAtGithub: This informaton is also intended for the end-user! An end-user needs it to identify the correct device out of many with similar names. It should be visible for the end-user by default therefore.
But once the end-user identified the correct device, and the mapping is successful loaded and working, he no longer needs these details.

src/controllers/dlgprefcontrollers.cpp
```diff
@@ -203,10 +203,24 @@ void DlgPrefControllers::setupControllerWidgets() {
 
         QTreeWidgetItem* pControllerTreeItem = new QTreeWidgetItem(
                 QTreeWidgetItem::Type);
+
+        const QString treeImage = [protocol = pController->getDataRepresentationProtocol()] {
+            switch (protocol) {
+            case DataRepresentationProtocol::USB_BULK_TRANSFER:
+                return QStringLiteral("ic_preferences_bulk.svg");
```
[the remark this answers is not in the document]
JoergAtGithub: We use this filename as argument for addPageWidget. This function need a filename only and does not work with paths.

src/controllers/dlgprefcontrollerdlg.ui
Swiftb0y: right, but how user friendly is that? In the long-term we probably need a different mechanism for identifying devices...

JoergAtGithub: Yes, the long term target is Plug&Play. That is why I make the device data accessible in this first step.

src/controllers/dlgprefcontrollers.cpp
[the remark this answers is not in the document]
Swiftb0y: Ah. Thats a future refactoring opportunity IMO.

src/controllers/dlgprefcontrollerdlg.ui
Swiftb0y: nice. So leave like this for now and `--developer` once plug'n'play is achieved?

src/controllers/hid/hidusagetables.h
Swiftb0y: If the script and/or the header describes where to find the original json, then lets leave the json out of the tree.

src/controllers/controller.cpp
```diff
@@ -37,6 +37,18 @@ ControllerJSProxy* Controller::jsProxy() {
...
+QString Controller::physicalTransport2String(PhysicalTransportProtocol protocol) {
+    static const QMap<PhysicalTransportProtocol, QString> protocolMap = {
+            {PhysicalTransportProtocol::USB, QStringLiteral("USB")},
+            {PhysicalTransportProtocol::BlueTooth, QStringLiteral("Bluetooth")},
+            {PhysicalTransportProtocol::I2C, QStringLiteral("I2C")},
+            {PhysicalTransportProtocol::SPI, QStringLiteral("SPI")},
+            {PhysicalTransportProtocol::FireWire, QStringLiteral("Firewire - IEEE 1394")},
+            {PhysicalTransportProtocol::UNKNOWN, tr("Unknown")}};
+
+    return protocolMap.value(protocol, tr("Unknown"));
```
Swiftb0y, line 49: why the map over a `switch-case`?

src/controllers/controller.h
```diff
@@ -9,6 +9,21 @@
 class ControllerJSProxy;
 class ControllerScriptEngineLegacy;
 
+enum class PhysicalTransportProtocol : uint8_t {
```
Swiftb0y, line 12: is the explicit type necessary for ABI compatibility?

src/controllers/hid/hiddevice.h
```diff
@@ -59,25 +64,46 @@ class DeviceInfo final {
...
     }
 
-    QString formatVID() const;
-    QString formatPID() const;
-    QString formatReleaseNumber() const;
-    QString formatInterface() const;
-    QString formatUsage() const;
+    bool isValid() const {
+        return !getProductString().isNull() && !getSerialNumber().isNull();
+    }
```
Swiftb0y, line 106: are you sure all hid devices have a productstring and serial number they report? Why do exactly these attributes make the device "valid"? What does valid mean in this context? Usually objects that are not valid result in UB if you call most member functions on them, but thats not the case here, afaict.

src/controllers/hid/hidusagetables.h
```diff
@@ -0,0 +1,18 @@
...
+namespace mixxx {
+
+namespace hid {
+
+class HidUsageTables {
+  public:
+    HidUsageTables() = default;
+    static QString getUsagePageDescription(uint16_t usagePage);
+    static QString getUsageDescription(uint16_t usagePage, uint16_t usage);
+};
```
Swiftb0y, line 14: convert to namespace instead? There is no member data.

tools/hid_usage_tables.h
Swiftb0y, line 1: empty file?

src/controllers/controller.cpp
JoergAtGithub: It doesn't matter here, but in general, a map scales better in lookup performance. So it's just use of the common pattern.

src/controllers/hid/hiddevice.h
JoergAtGithub: I didn't touch this line in this PR - and do not plan to.

But I can tell you, that only USB HID devices can store a product string. But the numeric VID & PID exist for all HID devices. This is because the HID descriptors are not specified with fields for strings. They just contain the numerical index of a string in the USB String Descriptor. And non-USB devices simply have no USB String Descriptor to reference.
But it seems the fields are always filled by the OS, in case of my Lenovo Thinkpad with an I2C Touchpad from Elan, Windows reports "Microsoft" "HIDI2C Device" with S/N 9999.

src/controllers/controller.cpp
Swiftb0y: > It doesn't matter here, but in general, a map scales better in lookup performance. So it's just use of the common pattern.

I actually highly doubt that when they keys are effectively `int`s.

src/controllers/hid/hiddevice.h
Swiftb0y: whoops yeah sorry about the complaint. The diff split weirdly so I didn't notice that this was already here previously. Then disregard the comment.

src/controllers/controller.cpp
```diff
@@ -37,6 +37,24 @@ ControllerJSProxy* Controller::jsProxy() {
...
+    case PhysicalTransportProtocol::I2C:
+        return QStringLiteral("I2C");
+    case PhysicalTransportProtocol::SPI:
+        return QStringLiteral("SPI");
+    case PhysicalTransportProtocol::FireWire:
+        return QStringLiteral("Firewire - IEEE 1394");
+    case PhysicalTransportProtocol::UNKNOWN:
+    default:
+        return tr("Unknown");
+    }
```
Swiftb0y, line 55: nit: remove `default` case and benefit from `-Wswitch`.
```suggestion
    case PhysicalTransportProtocol::UNKNOWN:
        break; // effectively fallthrough
    }
    return tr("Unknown");
```

Swiftb0y (commented):
only nit apart from this and the namespace request from the previous review. LGTM afterwards.

JoergAtGithub:
All done!

Swiftb0y (approved):
Yup. Thanks for the reminder.

Issue #13870, opened 2024-11-10
thread sanitizer issue 09


m0dB:
### Bug Description

[tsan09.txt](https://github.com/user-attachments/files/17691471/tsan09.txt)

### Version

2.5

PR #13899, opened 2024-11-17
avoid data race on m_pStream

m0dB:
This fixes several tsan warnings related with stopping / closing portaudio streams, including #13870.

Apart from access to a non-atomic variable m_pStream, the real problem here was that 
readProcess() writeProcess() would take a copy of m_pStream:

```
PaStream* pStream = m_pStream;
```

and close() would then call Pa_StopStream(m_pStream); Apparently stopping a stream and still accessing it in the callback leads to data races, at least on macOS.

Now, it the portaudio docs I read:

> [Pa_StopStream()](https://www.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#af18dd60220251286c337631a855e38a0) is designed to make sure that the buffers you've processed in your callback are all played, which may cause some delay. Alternatively, you could call [Pa_AbortStream()](https://www.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#a138e57abde4e833c457b64895f638a25). On some platforms, aborting the stream is much faster and may cause some data processed by your callback not to be played.
> Another way to stop the stream is to return either paComplete, or paAbort from your callback. paComplete ensures that the last buffer is played whereas paAbort stops the stream as soon as possible. If you stop the stream using this technique, you will need to call [Pa_StopStream()](https://www.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#af18dd60220251286c337631a855e38a0) before starting the stream again.

So what I do: in close() I change the (atomic) return value for readProcess and writeProcess from paContinue to paAbort, which will stop the stream. Pa_StopStream can then safely be called and m_pStream set to null.

As a side effect this also addresses a comment in the code about a failed intent to use Pa_AbortStream instead of PaStopStream.

This works well on macOS and I have not seen any portaudio related tsan warnings anymore. But since the platform specific aspects of Portaudio it is important to ensure this works as expected on Windows and Linux.

### DIFF
--- a/src/soundio/sounddeviceportaudio.cpp
+++ b/src/soundio/sounddeviceportaudio.cpp
@@ -94,7 +94,8 @@ SoundDevicePortAudio::SoundDevicePortAudio(UserSettingsPointer config,
           m_framesSinceAudioLatencyUsageUpdate(0),
           m_syncBuffers(2),
           m_invalidTimeInfoCount(0),
-          m_lastCallbackEntrytoDacSecs(0) {
+          m_lastCallbackEntrytoDacSecs(0),
+          m_streamState(StreamState::STOP) {
     // Setting parent class members:
     m_hostAPI = Pa_GetHostApiInfo(deviceInfo->hostApi)->name;
     m_sampleRate = mixxx::audio::SampleRate::fromDouble(deviceInfo->defaultSampleRate);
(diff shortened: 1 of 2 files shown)


daschuer:
m_pStream is meant to be an atomic. I think we can use the null state of this pointer instead of the additional atomic 
https://github.com/mixxxdj/mixxx/blob/4a35def01a54fd5a13b871a1c8e08ed8100f1d8e/src/soundio/sounddeviceportaudio.h#L62

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -820,7 +856,9 @@ int SoundDevicePortAudio::callbackProcessDrift(
             //qDebug() << "callbackProcess read:" << (float)readAvailable / outChunkSize << "Buffer empty";
         }
     }
-    return paContinue;
+    return m_streamState.load(std::memory_order_acquire) == StreamState::STOP
```
daschuer, line 859: This is not used for syncing. Does relax also work? We may avoid the condition to store the return value directly.

m0dB: Yeah, I initially stored the return value in an atomic, but IIRC I needed the extra state when starting the streams. But I don't remember so I will revise this.

m0dB:
I have made m_pStream atomic and now store the callback return value as atomic directly. Definitely simpler, and it fixes the issues just as well! Thanks for suggesting this, @daschuer

m0dB:
It would also be nice to know if this solves the bugs on linux and windows related with opening and closing device streams, e.g when waking up from sleep and unplugging devices.

I am okay with postponing other tsan fixes to post 2.5, but I believe this one should go in.

daschuer (approved):
LGTM, Thank you. Yes this is a bug fix that belongs to 2.5.

PR #13900, opened 2024-11-17
Fix crash in case of QEvent::ChildRemoved

daschuer:
It turns out the QEvent::ChildRemoved also creates a dangling pointer. 
Added the band-aid fix for this as well. 

Discussed in: 
https://github.com/mixxxdj/mixxx/pull/13889

Reduplicated the code a bit as well.


ronso0 (approved):

ronso0:
Thank you!

PR #13902, opened 2024-11-18
(fix) Library > Analysis: enable Stop button when anaylzing crate/playlist

ronso0:
I just noticed that when you select [crate] -> Analyse entire crate, the analysis can't be stopped because the Stop Analysis button is deactivated (until the table selection changes or Select All is clicked).


Swiftb0y (approved):
LGTM. Thank you.

PR #13906, opened 2024-11-18
(fix) adjust some skin controls, allow point-and-click mapping

ronso0:
Now these can also be mapped by click in MIDI Learning Wizard


Swiftb0y (approved):
LGTM. Thank you.

PR #13907, opened 2024-11-18
(fix) Controller: clear mapping pointer when unloading a mapping

ronso0:
With #13842 I noticed that the controller's pMapping still points to the previously loaded mapping when loading 'No mapping'.
Is there any reason for that?

### DIFF
--- a/src/controllers/controller.h
+++ b/src/controllers/controller.h
@@ -81,6 +81,10 @@ class Controller : public QObject {
     template<typename SpecificMappingType>
     std::shared_ptr<SpecificMappingType> downcastAndTakeOwnership(
             std::shared_ptr<LegacyControllerMapping>&& pMapping) {
+        // When unsetting a mapping (select 'No mapping') we receive a nullptr
+        if (pMapping == nullptr) {
+            return nullptr;
+        }
         // Controller cannot take ownership if pMapping is referenced elsewhere because
         // the controller polling thread needs exclusive accesses to the non-thread safe
         // LegacyControllerMapping.
--- a/src/controllers/controllermanager.cpp
+++ b/src/controllers/controllermanager.cpp
@@ -423,6 +423,7 @@ void ControllerManager::slotApplyMapping(Controller* pController,
     if (!pMapping) {
         closeController(pController);
         // Unset the controller mapping for this controller
+        pController->setMapping(nullptr);
         m_pConfig->remove(key);
         emit mappingApplied(false);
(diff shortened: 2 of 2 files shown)


src/controllers/bulk/bulkcontroller.h
```diff
@@ -43,6 +43,9 @@ class BulkController : public Controller {
 
     virtual std::shared_ptr<LegacyControllerMapping> cloneMapping() override;
     void setMapping(std::shared_ptr<LegacyControllerMapping> pMapping) override;
+    void clearMapping() override {
+        m_pMapping = nullptr;
+    }
```
Swiftb0y, line 48: just call `setMapping` with a `nullptr` directly.

ronso0: That was the first thing I tried.
Gives me this when selecting No mapping
```
DEBUG ASSERT: "pMapping.use_count() == 1" in function std::shared_ptr<_Tp> Controller::downcastAndTakeOwnership(std::shared_ptr<LegacyControllerMapping>&&) [with SpecificMappingType = LegacyMidiControllerMapping] at ./src/controllers/controller.h:89
```

ronso0: https://github.com/mixxxdj/mixxx/blob/815219d8a048165afd60c5283874da71f6f3e342/src/controllers/controller.h#L82-L96

Swiftb0y: right, that function does not take that usecase into account. Simply add a `if (pMapping == nullptr) { return nullptr; }` so it supports that usecase. Alternatively you can add the appropriate handling to the `setMapping` implementations.

ronso0: Okay. I only took a brief look at this and didn't want to touch it.
But yeah, that's a straight forward solution.
Thanks!

Swiftb0y (approved):
conceptually LGTM. Thank you. Waiting for CI.

ronso0:
Ready for merge, right?

daschuer (approved):
Thank you. Maybe this also fixes https://github.com/mixxxdj/mixxx/issues/13940? 
https://github.com/mixxxdj/mixxx/pull/13980

ronso0:
Hmm I dont see how this could be related.

daschuer:
Yes, probably, not but at least it touches the same code.

PR #13909, opened 2024-11-19
Controllers: allow to enable MIDI Through Port in non-developer sessions

ronso0:
Fixes #8356
Replaces #4148 

Enabling requires restart.
Setting is stored in config as `[Controller] midi_through_enabled`

The description is just a proposal.
In a followup we may replace it with brief description and a link to to-be-written manual chapter.

image

PR #13910, opened 2024-11-19
MixxxApplication: Only log opaque pointer after notify

fwcd:
As an addendum to #13900, `QEvent::Timer` seems to be another event type where a dangling pointer could come up (the screenshot includes some more logging which I've removed again):

Screenshot 2024-11-19 at 18 46 45

Since the target pointer could be dangling for any event point, this PR takes the safer road by only logging the opaque pointer after the notify call.

### DIFF
--- a/src/mixxxapplication.cpp
+++ b/src/mixxxapplication.cpp
@@ -201,21 +201,12 @@ bool MixxxApplication::notify(QObject* pTarget, QEvent* pEvent) {
 
     if (m_isDeveloper &&
             time.elapsed() > kEventNotifyExecTimeWarningThreshold) {
-        QDebug debug = qDebug();
-        debug << "Processing"
-              << pEvent->type()
-              << "for object";
-        if (pEvent->type() == QEvent::DeferredDelete ||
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
I think that this could happen with any event type - but QEvent::DeferredDelete has the speciality, that it always happen (if we hit the time limit).
As I expressed already in https://github.com/mixxxdj/mixxx/pull/13889#issuecomment-2481005443 I se only two proper fixes:

1.     Cache the object informantion before calling notify, which will affect the general performance in developer mode
2.     Remove the printout of the target object information and just print the information about the event type

fwcd:
IMO not printing the target (or only the hex pointer) would be a safer option for now if we don't know whether the pointer might be dangling.

JoergAtGithub:
@ronso0 has already opened a related PR #13909, which removes the only known case to use `--developer` for non-development.

fwcd:
Yeah, I believe we should still fix this though, since non-deterministic crashes shouldn't happen, even in developer mode.

PR #13917, opened 2024-11-22
Fix read before m_bufferInt during scratching

daschuer:
This should be a last minute 2.4 fix, because it may use an insane high value of the buffer allocated before. 
When this is at the end of a memory segment it will even crash Mixxx.


Swiftb0y (approved):
LGTM.

PR #13923, opened 2024-11-23
fix: waveform EQ High&Mid visualization broken

### LINKED ISSUE #13922: "Kill Mid" and "High" affect each other in the Waveform Display
### Bug Description

I won't go into depth describing the bevavior. Because either I am completely missing something and there's no bug or it is apperent that something is wrong and must be looked into further anyway.

**BUG:**
"Kill Mid" button and "High" knob affect each other in the Waveform Display when using "Filtered", "RGB" or "Stacked" Waveform Type. This does only affect the Waveform Display as the audio output is as expected.

Image shows two cases (there are more) where the display is incorrect.
mixxx 2.4.1

edit: Just noticed the "Kill High" button does not affect the Waveform Display.

### Version

2.4.1, 2.6-alpha-136-g894e96f9b2 (main)

### OS

Linux 6.11.9-arch1-1

Swiftb0y:
~~(Likely)~~ fixes #13922

We should also get this into 2.4.2.


JoergAtGithub:
I can confirm, that this PR fixes the issue! Thank you!

Swiftb0y:
Thanks. Also just confirmed it myself. Had some trouble getting it to build because the laptop kept overheating :sweat_smile:

PR #13930, opened 2024-11-24
(fix) Track property editor: don't close when opening context menu

ronso0:
This slipped through in #11755
Currently the editor is closed on _every_ FocusOut event, but that's wrong if the QLineEdit's menu pops up.
That one seems to be the only event that has the reason `Qt::PopupFocusReason` so we can filter easily.


daschuer (approved):
I can confirm the issue and the fix. 
LGTM

PR #13268, opened 2024-05-23
feat: add advanced stem loading COs

acolombier:
Currently, this goes with a mix of extending the existing load signal/slot and creating a second one, called subsequently. Is there any preference? Should I change the implementation to use only one? (only extending existing signal or only subsequently emit)

Here is documentation, as per in the EPIC (#13116)

## Pre-mixed stereo load

`[ChannelX],load_selected_track_stems` can be used as followed:~

- `1` (`0b0001`) means only the first stem is selected for stereo pre-mixing.
- `13` (`0b1101`) means all stem but the second are selected for stereo pre-mixing.
- `15` (`0b1111`) means the entire track as stereo. Since it is impossible to load that track without invalidating the replaygain due to the missing DSP support, all the four stem are currently mixed upstream. In the future, the pre-mastered track will be used.

Note that using `0` is not possible as this is the "idle state" of the CO. `0` also means `default` (load as stem deck on primary decks, or load the pre-mastered track on sampler/preview), which is equivalent to `[ChannelX],LoadSelectedTrack`

Individual bit mask will be reflected in the `stem_Y_selected` CO.

PR #13934, opened 2024-11-24
[POC] stem selection menu

daschuer:
grafik

This is a prove of concept of a stem selection menu. 

It opens if you press "->" (cursor right) when a stem file is selected in the library. This is useful to use the library encoder to select a stem for loading to samplers. 
* Turn Library Encoder to select a Track 
* Shift + Turn, to open the selection menu. 
* Turn and push to select a stem 
* click the sampler button to load. 

Alternative you may also focus one of the buttons to load the stem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -780,6 +780,8 @@ add_library(mixxx-lib STATIC EXCLUDE_FROM_ALL
   src/dialog/dlgkeywheel.ui
   src/dialog/dlgreplacecuecolor.cpp
   src/dialog/dlgreplacecuecolordlg.ui
+  src/dialog/dlgstemselect.cpp
+  src/dialog/dlgstemselect.ui
   src/effects/backends/builtin/autopaneffect.cpp
   src/effects/backends/builtin/balanceeffect.cpp
   src/effects/backends/builtin/bessel4lvmixeqeffect.cpp
--- a/src/dialog/dlgstemselect.cpp
+++ b/src/dialog/dlgstemselect.cpp
@@ -1,0 +1,173 @@
+#include "dialog/dlgstemselect.h"
+
+#include <QDir>
+#include <QKeyEvent>
+#include <QMouseEvent>
+#include <QSvgRenderer>
+
+#include "control/controlobject.h"
+#include "library/coverartcache.h"
+#include "library/library_prefs.h"
+#include "moc_dlgstemselect.cpp"
+#include "track/track.h"
+#include "widget/wcoverartlabel.h"
+
+namespace {
+
+} // namespace
+
+DlgStemSelect::DlgStemSelect(QWidget* parent)
+        : QDialog(parent) {
+    setupUi(this);
+
+    // setWindowFlags(Qt::Popup);
+    // setAttribute(Qt::WA_StyledBackground);
+
+    m_pWCoverArtLabel = make_parented<WCoverArtLabel>(this);
+
+    // Cover art
+    CoverArtCache* pCache = CoverArtCache::instance();
+    if (pCache) {
+        connect(pCache,
+                &CoverArtCache::coverFound,
+                this,
+                &DlgStemSelect::slotCoverFound);
+    }
+
+    coverLayout->setAlignment(Qt::AlignRight | Qt::AlignTop);
+    coverLayout->setSpacing(0);
+    coverLayout->setContentsMargins(0, 0, 0, 0);
+    coverLayout->insertWidget(0, m_pWCoverArtLabel.get());
+
+    installEventFilter(this);
+
+    connect(buttonAbort,
+            &QAbstractButton::clicked,
+            this,
+            &QDialog::accept);
+
+    connect(radioButtonStem,
+            &QRadioButton::toggled,
+            this,
+            &DlgStemSelect::slotStemMixToggled);
+    connect(radioButtonStereo,
+            &QRadioButton::toggled,
+            this,
+            &DlgStemSelect::slotStemMixToggled);
+
+    connect(checkBoxStem1,
+            &QCheckBox::stateChanged,
+            this,
+            &DlgStemSelect::slotStemChecked);
+    connect(checkBoxStem2,
+            &QCheckBox::stateChanged,
+            this,
+            &DlgStemSelect::slotStemChecked);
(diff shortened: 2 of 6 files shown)


daschuer:
This is the alternative of loading a sampler via the context menu: 
grafik
It is not easy to use because you can't see clearly which sampler in the skin corresponds to which sampler in the menu: 
The user will probably use drag and drop. 

In case a stem file is dropped to a sampler, we can also pop up this menu to select which stem selection shall be loaded. 

What do you think?

Eve00000:
> What do you think?
I understand wat you're trying to do but IMHO you're making it to complicated.
there are 3 usercases
1 -> controller -> co's
2 -> keyboard -> load to -> sampler -> load the 'drums' stems (your 2nd image)
3 -> mouse -> drag and drop -> 2 possibilities 
    A: pop up  after drop : what do you want to load  Master  / Stem (4 tracks) / Stem 1 / Stem 2 / Stem 3 / Stem 4
    B: drag the wanted stem track: [M] | [S] | [1] [2] [3] [4] (see image, example in comment column) with icons or ...
afbeelding

daschuer:
> 1 -> controller -> co's

Which CO do you like to have to control visual feedback, if you consider that the Stem controlls are on the a second TouchOSC device? This POC shall demonstrate how to use the LibraryEncoder and also Keyboard. My idea is to use the already mapped 
"[Library]", "MoveHorizontal" and "[Library]", "MoveVertical". will this work for you? 

> 2 -> keyboard -> load to -> sampler -> load the 'drums' stems (your 2nd image)

I consider this as not usable, because of the long list of samplers. Do we really have this keyboard only use case? Does anyone use this? 

> 3 ->  mouse -> drag and drop -> 2 possibilities

> A: pop up after drop : what do you want to load Master / Stem (4 tracks) / Stem 1 / Stem 2 / Stem 3 / Stem 4

The menu form my first picture can ne also used as a pop up after dropping to Samplers. Does int work for you? 

> B: drag the wanted stem track: [M] | [S] | [1] [2] [3] [4] (see image, example in comment column) with icons or ...

I have also considered this, especially the second seems to be usefull. 

However showing allways all stem rows feels a bit cluttered. We need a control to show this. 
My idea was to use the "->" key / shift + turn to show this view. Will this work? 

The current menu pop up can be used as an itermediate solution until the embedded list is in place. The implemntation of the embeedded list is a bit more work and has more dependencies, to ther libary actions.

Eve00000:
> > 1 -> controller -> co's
> 
> Which CO do you like to have to control visual feedback, if you consider that the Stem controlls are on the a second TouchOSC device? This POC shall demonstrate how to use the LibraryEncoder and also Keyboard. My idea is to use the already mapped "[Library]", "MoveHorizontal" and "[Library]", "MoveVertical". will this work for you?

I think that's a good option, with a check for the focus to be correct
The logic tells me (as in analog DJ-ing): select the track (vinyl/cd/digital)->, put it on a player -> play -> this is ok for 45rpm/CDS/CDM/digital,
But to be correct it's more: select the track in your mind -> search and take the carrier (Vinyl Album/CD Album/Digital) -> put it on the player ->select the track -> play
So the step to select the stem before dragging or after dragging is logical.
-> I am lucky to have some F1's on which I select the player / sampler (Quant or Capture -> shift-potmeter) ->  select the track with potmeter -> select the stem 1 - 4 (top row pads) or the complete stem potmeter-button.
On OSC I think the logic is 6 buttons, first select the track with the code you developed with Antoine: Master / full stem / stem tracks 1 - 4 -> after selection press load

> > 2 -> keyboard -> load to -> sampler -> load the 'drums' stems (your 2nd image)
> 
> I consider this as not usable, because of the long list of samplers. Do we really have this keyboard only use case? Does anyone use this?
>  I don't think somebody would use it because it's to long, to many actions, but you'll never now ;-)

> > 3 ->  mouse -> drag and drop -> 2 possibilities
> 
> > A: pop up after drop : what do you want to load Master / Stem (4 tracks) / Stem 1 / Stem 2 / Stem 3 / Stem 4
> 
> The menu form my first picture can ne also used as a pop up after dropping to Samplers. Does int work for you?
> 
> > B: drag the wanted stem track: [M] | [S] | [1] [2] [3] [4] (see image, example in comment column) with icons or ...
> 
> I have also considered this, especially the second seems to be usefull.
> 
> However showing allways all stem rows feels a bit cluttered. We need a control to show this. My idea was to use the "->" key / shift + turn to show this view. Will this work?
>  When Antoine started the whole stem-adventure I thought that having every stemtrack as an individual entry in the dbase would be the correct solution, that was when I was thinking about hotcues per stem, but that problem is already solved ;-)
In my head is still playing the possibility to have 'linked tracks' (remember our discussion->conversation about AI giving 'next track advice'? 
Linked (virtual) tracks could be:
- tracks you often play after the selected track
- tracks that match perfect Key/BPM/Energy
- tracks that are included (STEMS)
in the bigger picture: 
-> a tracklist pop-in with 'virtual' tracks

> The current menu pop up can be used as an itermediate solution until the embedded list is in place. The implemntation of the embeedded list is a bit more work and has more dependencies, to ther libary actions.
TOP

daschuer:
In the OSC case, I can follow you, that the slection menu similar to the one presented in this PR will be shown with 6 buttons, right? Is GUI feedback in the skin in addition desired in that case?  

>  -> after selection press load 

Where exactly? On the smapler buttons on your second controller? In this case we neet to communicate the selected stems into the engine, right?

Eve00000:
> In the OSC case, I can follow you, that the slection menu similar to the one presented in this PR will be shown with 6 buttons, right? Is GUI feedback in the skin in addition desired in that case?
there will be users  that want to experiment, even before installing the OSC on a 2nd device, and possibly on less powerfull machines, these users will be happy with a gui solution.
> > -> after selection press load
> 
> Where exactly? On the smapler buttons on your second controller? In this case we neet to communicate the selected stems into the engine, right?
the after selection (OSC or Gui) or / load selected (controller) is just one extra handle but can make it more reliable, sometimes you have a last second sample idea and don't have the time to prelisten in headphones, just load & play. So the right stemtrack needs to be loaded. 
#13268 was in the right direction with [LibraryStemX] and [stem_Y_selected]
just drop the switching from Master to  Stem or StemtrackX

PR #13935, opened 2024-11-25
Salvage #13772

Swiftb0y:
Simplified alternative of #13772 which implements the original scope I proposed.

CC @christophehenry @JoergAtGithub

### DIFF
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -303,16 +303,31 @@ declare namespace engine {
     function softStart(deck: number, activate: boolean, factor?: number): void;
 
     enum WellKnownCharsets {
-        US_ASCII,
-        Latin1,
-        ISO_8859_1,
-        Latin9,
-        ISO_8859_15,
-        UCS2, // with prepended Byte-Order-Mark
-        ISO_10646_UCS_2, // with prepended Byte-Order-Mark
-        UTF_8,
-        UTF_16BE,
-        UTF_16LE,
+        ASCII,          // American Standard Code for Information Interchange (7-Bit)
+        UTF_8,          // Unicode Transformation Format (8-Bit)
+        UTF_16LE,       // UTF-16 for Little-Endian devices (ARM, x86)
+        UTF_16BE,       // UTF-16 for Big-Endian devices (MIPS, PPC)
+        UTF_32LE,       // UTF-32 for Little-Endian devices (ARM, x86)
+        UTF_32BE,       // UTF-32 for Big-Endian devices (MIPS, PPC)
+        CentralEurope,  // Windows_1250 which includes all characters of ISO_8859_2
+        Cyrillic,       // Windows_1251 which includes all characters of ISO_8859_5
+        Latin1,         // Windows_1252 which includes all characters of ISO_8859_1
+        Greek,          // Windows_1253 which includes all characters of ISO_8859_7
+        Turkish,        // Windows_1254 which includes all characters of ISO_8859_9
+        Hebrew,         // Windows_1255 which includes all characters of ISO_8859_8
+        Arabic,         // Windows_1256 which includes all characters of ISO_8859_6
+        Baltic,         // Windows_1257 which includes all characters of ISO_8859_13
+        Vietnamese,     // Windows_1258 which includes all characters of ISO_8859_14
+        Latin9,         // ISO_8859_15
+        Shift_JIS,      // Japanese Industrial Standard (JIS X 0208)
+        EUC_JP,         // Extended Unix Code for Japanese
+        EUC_KR,         // Extended Unix Code for Korean
+        Big5_HKSCS,     // Includes all characters of Big5 and the Hong Kong Supplementary Character Set (HKSCS)
+        KOI8_U,         // Includes all characters of KOI8_R for Russian language and adds Ukrainian language characters
+        UCS2,           // Universal Character Set (2-Byte) ISO_10646
+        SCSU,           // Standard Compression Scheme for Unicode
+        BOCU_1,         // Binary Ordered Compression for Unicode
+        CESU_8          // Compatibility Encoding Scheme for UTF-16 (8-Bit)
     }
 
     /**
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -301,4 +301,33 @@ declare namespace engine {
      *               SoftStart with low factors would take a while until sound is audible. [default = 1.0]
      */
     function softStart(deck: number, activate: boolean, factor?: number): void;
+
+    enum WellKnownCharsets {
+        Latin1,
+        ISO_8859_1,
+        Latin9,
+        ISO_8859_15,
+        UCS2,
+        ISO_10646_UCS_2
+    }
+
+    /**
+     * Converts a string into another charset.
+     *
+     * This function is useful to display text on a device that does not make use of UTF-8.
+     * Available charset names are listed here: http://www.iana.org/assignments/character-sets/character-sets.xhtml.
+     * Characters that are unsupported by target charset will be transformed to null character (0x00).
+     * @param targetCharset The charset to encode the string into.
+     * @param value The string to encode
+     * @returns The converted String as an array of bytes. Will return an empty buffer on conversion error.
+     */
+    function convertCharset(targetCharset: string, value: string): ArrayBuffer
+
+    /**
+     * Version of {@link engine.convertCharset} to use with {@link engine.WellKnownCharsets}.
+     * @param targetCharset The charset to encode the string into.
+     * @param value The string to encode
+     * @returns The converted String as an array of bytes. Will return an empty buffer on conversion error.
+     */
+    function convertCharset(targetCharset: WellKnownCharsets, value: string): ArrayBuffer
 }
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -303,12 +303,16 @@ declare namespace engine {
     function softStart(deck: number, activate: boolean, factor?: number): void;
 
     enum WellKnownCharsets {
+        US_ASCII,
         Latin1,
         ISO_8859_1,
         Latin9,
         ISO_8859_15,
-        UCS2,
-        ISO_10646_UCS_2
+        UCS2, // with prepended Byte-Order-Mark
+        ISO_10646_UCS_2, // with prepended Byte-Order-Mark
+        UTF_8,
+        UTF_16BE,
+        UTF_16LE,
     }
 
     /**
--- a/res/controllers/engine-api.d.ts
+++ b/res/controllers/engine-api.d.ts
@@ -302,7 +302,7 @@ declare namespace engine {
      */
     function softStart(deck: number, activate: boolean, factor?: number): void;
 
-    enum WellKnownCharsets {
+    enum Charset {
         ASCII,          // American Standard Code for Information Interchange (7-Bit)
         UTF_8,          // Unicode Transformation Format (8-Bit)
         UTF_16LE,       // UTF-16 for Little-Endian devices (ARM, x86)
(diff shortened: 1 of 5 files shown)


src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -658,6 +661,78 @@ TEST_F(ControllerScriptEngineLegacyTest, connectionExecutesWithCorrectThisObject
...
+    case ISO_8859_15:
+        return 32;
+    case Latin1:
+    case ISO_8859_1:
+        return 33;
+    case UCS2:
+    case ISO_10646_UCS_2:
+        return 68;
+    }
+    // unreachable (TODO assert false?)
```
acolombier, line 712: That assert would make sense IMO

acolombier (approved):
LGTM!

src/test/controllerscriptenginelegacy_test.cpp
Swiftb0y: yeah, still TODO

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.h
```diff
@@ -72,6 +86,13 @@ class ControllerScriptInterfaceLegacy : public QObject {
             const double rate = -10.0);
     Q_INVOKABLE void softStart(const int deck, bool activate, double factor = 1.0);
 
+    Q_INVOKABLE QByteArray convertCharset(
+            const ControllerScriptInterfaceLegacy::WellKnownCharsets
+                    targetCharset,
+            const QString& value);
+
+    Q_INVOKABLE QByteArray convertCharset(const QString& targetCharset, const QString& value);
```
JoergAtGithub, line 94:
```suggestion
```
This API is unsafe and error prone! The number of supportet charsets must mit finite and tesable. This change was already requested in the orignial PR # 13772 and applies here as well!

JoergAtGithub (dismissed):

res/controllers/engine-api.d.ts
```diff
@@ -301,4 +301,35 @@ declare namespace engine {
...
+     * Converts a string into another charset.
+     *
+     * This function is useful to display text on a device that does not make use of UTF-8.
+     * Available charset names are listed here: http://www.iana.org/assignments/character-sets/character-sets.xhtml.
+     * Characters that are unsupported by target charset will be transformed to null character (0x00).
+     * @param targetCharset The charset to encode the string into.
+     * @param value The string to encode
+     * @returns The converted String as an array of bytes. Will return an empty buffer on conversion error or unavailable charset.
+     */
+    function convertCharset(targetCharset: string, value: string): ArrayBuffer
```
christophehenry, line 328: Wait this is still documented even though it's now an internal API?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1052,3 +1057,45 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+QByteArray ControllerScriptInterfaceLegacy::convertCharsetInternal(
+        const QString& targetCharset, const QString& value) {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)
+    QAnyStringView encoderName = QAnyStringView(targetCharset);
+#else
+    QByteArray encoderNameArray = targetCharset.toUtf8();
+    const char* encoderName = encoderNameArray.constData();
+#endif
+    QStringEncoder fromUtf16 = QStringEncoder(encoderName);
+    if (!fromUtf16.isValid()) {
```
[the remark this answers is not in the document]
Swiftb0y: see the commit message, writing the replacement char is better than replacing with null bytes IMO.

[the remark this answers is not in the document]
christophehenry: This may be a problem. During the tests I noticed that replacement char varies between Ubuntu and Fedora (maybe between Qt versions). Replace invalid chars with `\0x00` is the most predictable option.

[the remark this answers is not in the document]
Swiftb0y: Are you sure? I think it depends on the codec. The Qt docs say they use [`QChar::ReplacementCharacter`](https://doc.qt.io/qt-6/qchar.html#SpecialCharacter-enum) [or a question mark](https://doc.qt.io/qt-6/qstringconverter.html#Flag-enum).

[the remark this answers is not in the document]
christophehenry: [This is what I noticed here](https://github.com/mixxxdj/mixxx/pull/13772#issuecomment-2437566659). The relevant commit is probably [this one](https://github.com/mixxxdj/mixxx/commit/a12e5e3b1f0de67f0950dfca248da0d911132b2d).

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -658,6 +661,84 @@ TEST_F(ControllerScriptEngineLegacyTest, connectionExecutesWithCorrectThisObject
...
+}
+
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetUnsupportedChars) {
+    auto result = qjsvalue_cast<QByteArray>(
+            evaluate("engine.convertCharset('ISO-8859-15', 'مايأ نامز')"));
+    char sub = '\x1A'; // ASCII/Latin9 SUB character
+    EXPECT_EQ(result,
+            QByteArrayView::fromArray(
+                    {sub, sub, sub, sub, '\x20', sub, sub, sub, sub}));
+}
```
christophehenry, line 694: Those tests aren't relevant anymore since it's an internal API now, are they?

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -658,6 +661,84 @@ TEST_F(ControllerScriptEngineLegacyTest, connectionExecutesWithCorrectThisObject
     EXPECT_DOUBLE_EQ(1.0, pass->get());
 }
 
+TEST_F(ControllerScriptEngineLegacyTest, convertCharsetUndefinedOnUnknownCharset) {
+    const auto result = evaluate("engine.convertCharset('NULL', 'Hello!')");
+
+    EXPECT_EQ(qjsvalue_cast<QByteArray>(result), QByteArrayView(""));
+}
```
christophehenry, line 668: [Ditto](https://github.com/mixxxdj/mixxx/pull/13935/files#r1862222749).

src/test/controllerscriptenginelegacy_test.cpp
Swiftb0y: mostly yes. I should translate them to their `enum` equivalent so we can at least be reasonably sure about the output data.

src/test/controllerscriptenginelegacy_test.cpp
Swiftb0y: yeah... Not quite sure what to do here honestly. It seems that there are implicit conversions to enums happening, which isn't great if it just silently succeeds. I wonder what value plain `undefined` results in? I'd guess `undefined`->`0`->`US_ASCII`?

res/controllers/engine-api.d.ts
```diff
@@ -301,4 +301,50 @@ declare namespace engine {
...
```
christophehenry, line 343: Don't forget to remove this version.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1052,3 +1057,79 @@ void ControllerScriptInterfaceLegacy::softStart(int deck, bool activate, double
...
+    case CESU_8:
+        return convertCharsetInternal(QStringLiteral("CESU-8"), value);
+    }
+    m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unknown charset specified"));
+    return QByteArray();
+}
+
+QByteArray ControllerScriptInterfaceLegacy::convertCharsetInternal(
+        const QString& targetCharset, const QString& value) {
+#if QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)
```
christophehenry, line 1123: The 6.4 version has disappeared?

Swiftb0y: yeah, in hindsight I there was no reason to use `QTextEncoder` when QStringEncoder is also available, so we dropped it. QStringEncoder is also available in versions older than 6.4, so we can simply use that there too.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1,5 +1,10 @@
 #include "controllerscriptinterfacelegacy.h"
 
+#if QT_VERSION < QT_VERSION_CHECK(6, 4, 0)
+#include <QTextCodec>
+#else
+#include <QStringEncoder>
+#endif
```
christophehenry, line 7:
```suggestion
#include <QStringEncoder>
```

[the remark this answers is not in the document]
Swiftb0y: done

src/test/controllerscriptenginelegacy_test.cpp
Swiftb0y: I changed them. please take a look

src/test/controllerscriptenginelegacy_test.cpp
Swiftb0y: I just removed it for now. I don't really know how what the proper way to handle implicit enum conversions would be. We could take a QJSValue instead and implement our own checks, but ideally Qt would warn by itself.

res/controllers/engine-api.d.ts
```diff
@@ -301,4 +301,41 @@ declare namespace engine {
      *               SoftStart with low factors would take a while until sound is audible. [default = 1.0]
      */
     function softStart(deck: number, activate: boolean, factor?: number): void;
+
+    enum WellKnownCharsets {
```
JoergAtGithub, line 305:
```suggestion
    enum Charsets {
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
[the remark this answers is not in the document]
christophehenry: I'll try to reproduce later this week if I have time but I don't think this should be a blocker. We'll have time to change it later or introduce a new API if this proves unstable.

Swiftb0y:
@JoergAtGithub Anything left do to here? Pre-commit failure is unrelated. The commit history is garbage so I'd like to squash everything into a single commit.

res/controllers/engine-api.d.ts
JoergAtGithub: WellKnownCharsets still appears here

Swiftb0y: whoops yeah, I only `grep` in `src/` and forgot to double check. Fixed and amended.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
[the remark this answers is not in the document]
JoergAtGithub: @christophehenry The flags here are the last open point, did you find to check this out?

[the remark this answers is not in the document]
christophehenry: Oh sorry I didn't know you were waiting for me. I couldn't reproduce this. Feel free to merge as is.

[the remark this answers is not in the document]
JoergAtGithub: Ok, than lets merge this!

PR #13936, opened 2024-11-25
test: improve flaky stem color test

### LINKED ISSUE #13924: Flaky `StemControlTest.StemColor`
Unfortunately `StemControlTest.StemColor` often fails on CI.
```
The following tests FAILED:
	871 - StemControlTest.StemColor (Failed)
```
<details><summary>Relevant log output</summary>
<p>

```
./src/test/stemcontrolobjecttest.cpp:164: Failure
Expected equality of these values:
  m_pStem1Color->get()
    Which is: -1
  0xfd << 16 | 0x4a << 8 | 0x4a
    Which is: 16599626

./src/test/stemcontrolobjecttest.cpp:165: Failure
Expected equality of these values:
  m_pStem2Color->get()
    Which is: -1
  0xff << 16 | 0xff << 8 | 0x00
    Which is: 16776960

./src/test/stemcontrolobjecttest.cpp:166: Failure
Expected equality of these values:
  m_pStem3Color->get()
    Which is: -1
  0x00 << 16 | 0xe8 << 8 | 0xe8
    Which is: 59624

./src/test/stemcontrolobjecttest.cpp:167: Failure
Expected equality of these values:
  m_pStem4Color->get()
    Which is: -1
  0xad << 16 | 0x65 << 8 | 0xff
    Which is: 11363839
```

</p>
</details> 

@acolombier could you do me a favor and look into it?

acolombier:
This will improve the stability of the `StemControlTest.StemColor`, but slighlty slow the test run by adding a sleep. The alternative would be to add some extensive watch to the `CachingReader`, which is not worth the effort.

Previous benchmark to get some failure (*10 runs, sequentially*)

```sh
seq 10 | xargs -n 1 ctest --timeout 45 -R 'StemControlTest.StemColor'
```

Current benchmark without a single failure (*500 runs, on 20 workers, with 20 threads*)

```sh
seq 500 | xargs -P 20 -n 1 ctest --timeout 45 -R 'StemControlTest.StemColor'
```

Fixes #13924

### DIFF
--- a/src/engine/enginebuffer.cpp
+++ b/src/engine/enginebuffer.cpp
@@ -541,7 +541,6 @@ void EngineBuffer::slotTrackLoaded(TrackPointer pTrack,
 
     m_visualPlayPos->setInvalid();
     m_playPos = kInitialPlayPosition; // for execute seeks to 0.0
-    m_pCurrentTrack = pTrack;
 
     m_channelCount = trackChannelCount;
     if (m_channelCount > mixxx::audio::ChannelCount::stereo()) {
@@ -587,6 +586,7 @@ void EngineBuffer::slotTrackLoaded(TrackPointer pTrack,
         m_iSeekPhaseQueued = 0;
     }
 
+    m_pCurrentTrack = pTrack;
     // Start buffer processing after all EngineContols are up to date
     // with the current track e.g track is seeked to Cue
     m_iTrackLoading = 0;


acolombier:
Let me know what you think @Swiftb0y 
It's not perfect but it improves the stability enough with a minimum effort hopefully

Swiftb0y:
Well, if the delay works for `loadTrack()` then it might as well work for `ejectTrack()`. Doesnt change the fact that the trackloading logic is too flawed to be testable though. But thats not your fault.

acolombier:
> Well, if the delay works for `loadTrack()` then it might as well work for `ejectTrack()`. 

That's right but in order to reduce the waste of time, I've added it only when necessary to get my 500 run benchmarks to stay green. Happy to put it on if you think it's better for consistency.

> Doesn't change the fact that the trackloading logic is too flawed to be testable though. But that's not your fault.

Agreed - slightly off-topic, but I'm wondering if we could consider some E2E test setup. I've noticed this is something that can be done [in QML](https://doc.qt.io/qt-6/qml-qttest-testcase.html#simulating-keyboard-and-mouse-events). No idea how that looks like in practice, but this could be a great way to test this kind of complex scenario which aren't great fix for unit testing.

daschuer:
A certain blindly called sleep is not sufficient, because we don't know nothing about CI runner performance for certain Linux distros on Raspberry and friend. 

Can you please add a check of sufficient time has passed?

acolombier:
> A certain blindly called sleep is not sufficient

It definitely isn't! 

> we don't know nothing about CI runner performance for certain Linux distros on Raspberry and friend.

That's correct, this is the reason why I used 20 runners in parallel on my benchmark to "stress" the performance, but I agree it's still isn't perfect.

> Can you please add a check of sufficient time has passed?

Unfortunately, this require a way of catching or polling state from other threads, and AFAIS this would need a significant refactor to allow a correct design. 
Please note that this PR's ambition was only to improve the stability and reduce the false negative count. Would you prefer if I disable the test?

src/test/stemcontrolobjecttest.cpp
```diff
@@ -103,6 +103,7 @@ class StemControlTest : public BaseSignalPathTest {
         // CachingReaderWorker::trackLoaded (CachingReader) ->
         // EngineBuffer::loaded (CachingReader, direct) ->
         // BaseTrackPlayerImpl::slotTrackLoaded  (main)
+        QTest::qSleep(20); // sleep 20 ms for waiting event from CachingReaderWorker
```
daschuer, line 106: We already have a loop here that waits for
pEngineDeck->getEngineBuffer()->isTrackLoaded()
https://github.com/mixxxdj/mixxx/blob/437db42f5fc36ba8a1b6b0a6e84162d7525ba7d2/src/test/signalpathtest.h#L180

This means the event trackLoaded() should be already queued. 
https://github.com/mixxxdj/mixxx/blob/437db42f5fc36ba8a1b6b0a6e84162d7525ba7d2/src/engine/enginebuffer.cpp#L690

If this is not the case, something else seems to be wrong. 

At least the comments are missleading and should be adjusted.

daschuer (changes requested):

acolombier:
I have fixed the race condition between `EngineBuffer`'s thread and the main one.

src/engine/enginebuffer.cpp
```diff
@@ -587,6 +586,7 @@ void EngineBuffer::slotTrackLoaded(TrackPointer pTrack,
         m_iSeekPhaseQueued = 0;
     }
 
+    m_pCurrentTrack = pTrack;
```
daschuer, line 589: This change in EngineBuffer is a bit brittle. Now `EngineBuffer::isTrackLoaded()` will return false after `notifyTrackLoaded(pTrack, pOldTrack);` and the "track_loaded" CO is set to 1. 

Can you explain the race condition a bit more? Maybe we find a solution closer to the failing test.

daschuer (changes requested):

src/engine/enginebuffer.cpp
acolombier: `EngineBuffer::isTrackLoaded()` will return `true` but the color won't yet have been parsed till `notifyTrackLoaded` has emitted the signal to get the `BasePlayer` to proceed the stem color.
Since `notifyTrackLoaded` doesn't use `m_pCurrentTrack` directly, I thought it was fine to change.

So basically the tests probe for `engine` load instead of `player` load.

daschuer:
Here is a PR that implements waiting for the deck signal: 
https://github.com/mixxxdj/mixxx/pull/13960
Please have a look. 

Wen I remember correct, you had a commit for the eject case as well. Do you have interest to revive it?

acolombier:
> you had a commit for the eject case as well. Do you have interest to revive it?

Do you think this is necessary? AFAIK, this doesn't make any difference, does it? Happy to close in favour of #13960 anyway, anything that fixes the flaky test is good IMO!

daschuer:
> Do you think this is necessary?

Don't know. That's why I am asking.

acolombier:
Closing in favour of https://github.com/mixxxdj/mixxx/pull/13960

PR #13938, opened 2024-11-26
AudioUnitManager: Avoid undefined behavior by initializing `m_isInstantiated` to `false`

fwcd:
Turns out creating a `std::atomic<bool>` without initializing it is very likely undefined behavior, so we ought to initialize it.


daschuer (approved):
LGTM, Thank you. 

The behavior has changed since C++20 which we use in the 2.5 branch. 
https://github.com/mixxxdj/mixxx/blob/6f94fd0705ba7e01d660d9608c7718a06605796b/CMakeLists.txt#L268

But it can't hurt to explicit initialize a bool here.

PR #13945, opened 2024-11-29
AudioUnitManager: Wrap every instance in a shared pointer

fwcd:
This is a small patch I extracted from #13887 (and rebased to 2.5), which fixes a dangling pointer issue when Audio Units are instantiated asynchronously.

The issue is that the async initialization callback captures the `AudioUnitManager` instance, which is unsafe if the instance doesn't outlive the callback. The solution is to simply wrap the `AudioUnitManager` in a `QSharedPointer`, which ensures that the callback keeps the manager instance alive for as long as necessary. See https://github.com/mixxxdj/mixxx/pull/13887#issuecomment-2501814755 for further rationale.

I consider it relatively safe to merge this for 2.5 since it doesn't change anything about the instantiation logic itself (Audio Units are still instantiated synchronously to read the parameters and asynchronously once activated). The migration to make everything asynchronous will then be deferred to 2.6 via #13887.

### DIFF
--- a/src/effects/backends/audiounit/audiouniteffectprocessor.h
+++ b/src/effects/backends/audiounit/audiouniteffectprocessor.h
@@ -55,7 +55,7 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
             const GroupFeatureState& groupFeatures) override;
 
   private:
-    AudioUnitManagerPointer m_manager;
+    AudioUnitManagerPointer m_pManager;
 
     QList<EngineEffectParameterPointer> m_parameters;
     QList<AudioUnitParameterValue> m_lastValues;
--- a/src/effects/backends/audiounit/audiouniteffectprocessor.h
+++ b/src/effects/backends/audiounit/audiouniteffectprocessor.h
@@ -55,7 +55,7 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
             const GroupFeatureState& groupFeatures) override;
 
   private:
-    AudioUnitManager m_manager;
+    AudioUnitManagerPointer m_manager;
 
     QList<EngineEffectParameterPointer> m_parameters;
     QList<AudioUnitParameterValue> m_lastValues;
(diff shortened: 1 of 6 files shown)


src/effects/backends/audiounit/audiouniteffectprocessor.h
```diff
@@ -55,7 +55,7 @@ class AudioUnitEffectProcessor final : public EffectProcessorImpl<AudioUnitEffec
             const GroupFeatureState& groupFeatures) override;
 
   private:
-    AudioUnitManager m_manager;
+    AudioUnitManagerPointer m_manager;
```
m0dB, line 58: per https://github.com/mixxxdj/mixxx/wiki/Coding-Guidelines pointers should be prefixed with p, so m_pManager.

src/effects/backends/audiounit/audiounitmanager.mm
```diff
@@ -6,28 +6,36 @@
 
 #include "effects/backends/audiounit/audiounitmanager.h"
 
-AudioUnitManager::AudioUnitManager(AVAudioUnitComponent* _Nullable component,
-        AudioUnitInstantiationType instantiationType)
-        : m_name(QString::fromNSString([component name])),
+AudioUnitManager::AudioUnitManager(AVAudioUnitComponent* _Nullable component)
+        : m_name(component != nil ? QString::fromNSString([component name])
+                                  : "Unknown"),
           m_isInstantiated(false) {
```
m0dB, line 12: if this occurs often, it might make sense to append a counter for if you have multiple nameless components. but I guess this is more a safety net than anything else, right?

m0dB:
LGTM except for the missing p prefix.

src/effects/backends/audiounit/audiounitmanager.mm
fwcd: Yes, I don't expect this to happen usually, so this is just a fallback

fwcd:
Can this be merged?

m0dB:
It LGTM, but I find it hard to assess the risk so close to 2.5 release. Then again, this is all very isolated. If @daschuer doesn't object, I will merge this.

JoergAtGithub:
> It LGTM, but I find it hard to assess the risk so close to 2.5 release.

At least we have one user who reports, that this PR fixes an issue: https://github.com/mixxxdj/mixxx/issues/13967#issuecomment-2517191945

m0dB:
LGTM

PR #13950, opened 2024-11-30
Denon MC7000: Add sampler options to mapping settings

fwcd:
Scratching another (minor) itch, this exposes a few sampler-related options from the Denon MC7000 mapping to the user:

Screenshot 2024-11-30 at 01 36 15

I ran into this while trying to figure out why playing a sample would automatically stop all other playing samples and found that it wasn't a Mixxx issue, but something from the mapping. Since changing the script is inconvenient, here's a PR to fix it.

### DIFF
--- a/res/controllers/Denon-MC7000.midi.xml
+++ b/res/controllers/Denon-MC7000.midi.xml
@@ -92,6 +92,33 @@
                 </option>
             </group>
         </group>
+        <group label="Samplers">
+            <option
+                variable="prevSamplerStop"
+                label="Auto-stop previous samplers when starting new sampler"
+                type="boolean"
+                default="true">
+                <description>
+                    Whether the previous sampler shall stop before a new sampler starts.
+                    When true, a running sampler will stop before the new sampler starts.
+                    When false, all triggered samplers will play simultaneously.
+                </description>
+            </option>
+            <option
+                variable="samplerQty"
+                type="enum"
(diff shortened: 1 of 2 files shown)


res/controllers/Denon-MC7000.midi.xml
```diff
@@ -92,6 +92,33 @@
                 </option>
             </group>
         </group>
+        <group label="Samplers">
+            <option
+                variable="prevSamplerStop"
+                label="Auto-stop previous samplers when starting new sampler"
+                type="boolean"
+                default="true">
```
JoergAtGithub, line 100: Does true as default state make sense here? As it alters how Mixxx normaly behaves.

acolombier (approved):

res/controllers/Denon-MC7000.midi.xml
fwcd: This was the previous default, so I figured it would be less surprising to not automatically change it

JoergAtGithub: Maybe we can change the default after the next major release, than every existing user has stored the setting and new user get the proper default.

fwcd: Sure!

JoergAtGithub:
LGTM! Thank you!

PR #13952, opened 2024-11-30
Optimized tooltip generation in WBaseWidget

JoergAtGithub:
Code cleanup. Main change is that empty lines at the end of the tooltips are no longer printed.

### DIFF
--- a/src/widget/wbasewidget.cpp
+++ b/src/widget/wbasewidget.cpp
@@ -139,20 +139,16 @@ void WBaseWidget::setControlParameterRightUp(double v) {
 }
 
 void WBaseWidget::updateTooltip() {
-    // Remove leading and trailing whitespace/line breaks from the tooltip.
-    QString base = baseTooltip().trimmed();
-
     // If we are in developer mode, expand the tooltip.
     if (CmdlineArgs::Instance().getDeveloper()) {
         QStringList debug;
         fillDebugTooltip(&debug);
 
+        QString base = baseTooltip();
         if (!base.isEmpty()) {
             debug.append(QStringLiteral("Tooltip: \"%1\"").arg(base));
         }
         m_pWidget->setToolTip(debug.join(QChar('\n')));
-    } else {
-        m_pWidget->setToolTip(base);
     }
 }
 
--- a/src/widget/wbasewidget.cpp
+++ b/src/widget/wbasewidget.cpp
@@ -150,7 +150,7 @@ void WBaseWidget::updateTooltip() {
         if (!base.isEmpty()) {
             debug.append(QStringLiteral("Tooltip: \"%1\"").arg(base));
         }
-        m_pWidget->setToolTip(debug.join(QStringLiteral("\n")));
+        m_pWidget->setToolTip(debug.join(QChar('\n')));
     } else {
         m_pWidget->setToolTip(base);
     }
--- a/src/widget/wbasewidget.cpp
+++ b/src/widget/wbasewidget.cpp
@@ -154,24 +154,7 @@ void WBaseWidget::updateTooltip() {
 
 template<>
 QString toDebugString(const QSizePolicy::Policy& policy) {
-    switch (policy) {
-    case QSizePolicy::Fixed:
-        return QStringLiteral("Fixed");
-    case QSizePolicy::Minimum:
-        return QStringLiteral("Minimum");
-    case QSizePolicy::Maximum:
-        return QStringLiteral("Maximum");
-    case QSizePolicy::Preferred:
-        return QStringLiteral("Preferred");
-    case QSizePolicy::Expanding:
-        return QStringLiteral("Expanding");
-    case QSizePolicy::MinimumExpanding:
-        return QStringLiteral("MinimumExpanding");
-    case QSizePolicy::Ignored:
-        return QStringLiteral("Ignored");
-    default:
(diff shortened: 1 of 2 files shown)


src/widget/wbasewidget.cpp
```diff
@@ -139,63 +139,72 @@ void WBaseWidget::setControlParameterRightUp(double v) {
...
         QStringList debug;
         fillDebugTooltip(&debug);
 
-        QString base = baseTooltip();
         if (!base.isEmpty()) {
-            debug.append(QString("Tooltip: \"%1\"").arg(base));
+            debug.append(QStringLiteral("Tooltip: \"%1\"").arg(base));
         }
-        m_pWidget->setToolTip(debug.join("\n"));
+        m_pWidget->setToolTip(debug.join(QStringLiteral("\n")));
```
daschuer, line 153:
```suggestion
        m_pWidget->setToolTip(debug.join(QChar('\n')));
```

src/widget/wbasewidget.cpp
```diff
@@ -139,63 +139,72 @@ void WBaseWidget::setControlParameterRightUp(double v) {
...
-
-    for (const auto& pControlConnection : std::as_const(m_leftConnections)) {
-        *debug << QString("LeftConnection: %1").arg(pControlConnection->toDebugString());
-    }
-    for (const auto& pControlConnection : std::as_const(m_rightConnections)) {
-        *debug << QString("RightConnection: %1").arg(pControlConnection->toDebugString());
-    }
-    for (const auto& pControlConnection : std::as_const(m_connections)) {
-        *debug << QString("Connection: %1").arg(pControlConnection->toDebugString());
+    *debug << QStringLiteral("ClassName: %1")
```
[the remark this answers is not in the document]
daschuer: The QStringBuilder does it implicit for us, if we replace the %1 placeholder by normal string concat using +. 
However is it worth the efforts in this PR?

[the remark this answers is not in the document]
Swiftb0y: `debug` is a `QList<QString>`. I don't see how QStringBuilder is relevant here.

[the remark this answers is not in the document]
daschuer: I mean 
`QStringLiteral("ObjectName: %1").arg(m_pWidget->objectName())` 
-> 
`QStringLiteral("ObjectName: ") + m_pWidget->objectName()`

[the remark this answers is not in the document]
Swiftb0y: I don't think that makes a meaningful difference for a single argument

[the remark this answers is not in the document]
JoergAtGithub: I will not touch this, because it's seldom executed code only active in developer mode - and there is obviously no consensus on the correct solution either.

Swiftb0y (approved):
Loads of more stuff that could be done, but this suffices for now. LGTM. Thank you.

src/widget/wbasewidget.cpp
```diff
@@ -139,63 +139,72 @@ void WBaseWidget::setControlParameterRightUp(double v) {
...
+        return QStringLiteral("Preferred");
+    case QSizePolicy::Expanding:
+        return QStringLiteral("Expanding");
+    case QSizePolicy::MinimumExpanding:
+        return QStringLiteral("MinimumExpanding");
+    case QSizePolicy::Ignored:
+        return QStringLiteral("Ignored");
+    default:
+        return QString::number(static_cast<int>(policy));
+    }
```
ronso0, line 178: this does the trick, just without the boilerplate
`return QVariant::fromValue(policy).toString();`

src/widget/wbasewidget.cpp
```diff
@@ -139,63 +139,72 @@ void WBaseWidget::setControlParameterRightUp(double v) {
...
 
-        QString base = baseTooltip();
         if (!base.isEmpty()) {
-            debug.append(QString("Tooltip: \"%1\"").arg(base));
+            debug.append(QStringLiteral("Tooltip: \"%1\"").arg(base));
         }
-        m_pWidget->setToolTip(debug.join("\n"));
+        m_pWidget->setToolTip(debug.join(QChar('\n')));
+    } else {
+        m_pWidget->setToolTip(base);
```
ronso0, line 155: This means QWidget::setTooltip() is now called on every tooltip event which is redundant.
Please move it back into the CmdLine..dev parantheses

ronso0 (changes requested):

src/widget/wbasewidget.cpp
ronso0: naa,  meant as replacement for the entire switch.

ronso0: I did test it earlier, works beautifully

JoergAtGithub: Done!

ronso0:
Thank you, LGTM!

ronso0:
Whoopsy, this appends the shortcut tooltips without the leading linebreak.
[QString::trimmed() removes both leading and trailing whitespace, this includes the ASCII characters '\t', '\n', '\v', '\f', '\r'](https://doc.qt.io/qt-6/qstring.html#trimmed)

We should use `mixxx::removeTrailingWhitespaces()`
Will provide a PR

ronso0:
#14757

PR #13955, opened 2024-11-30
Add some ready-to-use Effect chain presets

esbrandt:
Hello, this PR aims to add some effect chain presets.

Sorting is kinda confusing if the effect chain list gets longer, so i opted to add ``FX=`` and ``EQ=`` prefixes. 
This works well if the effects are sorted by dnd in  
 ``Preferencess > Effect > Effect Chain presets``
effect-chain

The names of the effect chains pretty much sum up, what the chain will do. The easiest way to use the FX`s is by turning the superknobs. Just make sure all effect are enabled in the rack, and the signal is wet. Not all FX work equally well with every genre of music. 

Thanks for testing.


daschuer (changes requested):
Thank you for this nice addition.

Unfortunately they are not picked up after merging. 
What is required that new upstream settings are shown? Maybe we need to do something in the code? Else no one will see them after update. 

It there an issue with =? at least this is not easy to handle across different OSs. 

The existing presets do not follow the new naming schema. Maybe we can drop FX and prefix the EQs with EQ without =? 

What is the use case for Flat? 

I consider devices with EQ settings for a certain genre as toys. 
Can we find better names that reflect real use cases? 

In my experience I do usually the opposite. Remove Base form modern Rap tracks with a big sub base foundation. I have also hard times to play a 70th track after such a track, regarding EQ. Maybe we can reflect these issues in the name. 

Maybe users with DAW experience can recommend good names?

ronso0:
Thanks for setting up this preset!

However, these are too many to included IMHO, especially the graphic EQ presets.
Wasn't one idea for the new effects preset system that users can easily import, export and share presets?
So, shall we include just a few, like a few good starting points/templates, and publish the others in the forum (new section Effect Presets? + put a link in the effects preferences page)?

ronso0:
FYI I've created the Effect Presets forum a few days ago
https://mixxx.discourse.group/c/effect-presets/17

I intend to create some tutorials for that, eg. how to create dedicated LPF / HPF presets.

ronso0:
How do we want ot proceed here?
I still think we should publish them in [Forum -> Effect Presets](https://mixxx.discourse.group/c/effect-presets/17) and add a link to **Preferences -> Effects**. Would be cool if each preset has a short description.

PR #13966, opened 2024-12-03
EngineRecord: Fix recorded duration at start of recording

### LINKED ISSUE #13964: First timestamp of CUE sheet is bogus after restarting a recording
### Bug Description

To reproduce:

- Play a track
- Start a recording
- Let it record a moment
- Stop the recording
- Start a new recording
- Stop the new recording

In the first recording the first track starts at 00:00:00 (as expected), e.g.:

```
FILE "2024-12-03_00h28m23s.wav" WAVE
  TRACK 01 AUDIO
    TITLE "Harder, Better, Faster, Stronger"
    PERFORMER "Daft Punk"
    INDEX 01 00:00:00
```

In the second recording, however, the first timestamp is not 00:00:00 despite the track even though the track already played when pressing the record button:

```
FILE "2024-12-03_00h28m31s.wav" WAVE
  TRACK 01 AUDIO
    TITLE "Harder, Better, Faster, Stronger"
    PERFORMER "Daft Punk"
    INDEX 01 00:05:00
```

Further investigation showed that this timestamp (00:05:00) is actually the _length of the first recording_, which shouldn't have any effect on further recordings. All subsequent timestamps are fine, however.

This seems to be a really old bug, I can reproduce it all the way back to Mixxx 2.3.0 (haven't checked older versions).

### Version

2.6-alpha-138-g0eccb1b2db (HEAD)

### OS

macOS 15.1.1

fwcd:
### Fixes #13964 

This fixes the first timestamp of CUE sheets upon restarting the recording, see https://github.com/mixxxdj/mixxx/issues/13964#issuecomment-2513231487 for details.

In the medium-term this logic should probably be refactored to contain less duplication, but I'd consider it more important to get this bug fixed for 2.5 before looking at (potentially riskier) refactorings, which should probably also target `main` to get more testing.

(Hope it's okay that I added this to the 2.5.0 milestone, since it's a really trivial fix)

### DIFF
--- a/src/engine/sidechain/enginerecord.cpp
+++ b/src/engine/sidechain/enginerecord.cpp
@@ -146,6 +146,7 @@ void EngineRecord::process(const CSAMPLE* pBuffer, const int iBufferSize) {
             // clean frames counting and get current sample rate.
             m_frames = 0;
             m_sampleRate = mixxx::audio::SampleRate::fromDouble(m_sampleRateControl.get());
+            m_recordedDuration = 0;
 
             if (m_bCueIsEnabled) {
                 openCueFile();


Swiftb0y (approved):
LGTM. Thank you. Definitely agree with the assessment that this requires a little refactoring, but not in 2.5

PR #13968, opened 2024-12-03
fix: Rotary Jog filter ring buffer indexing

### LINKED ISSUE #13965: `jog` CO behaviour regression
### Bug Description

Starting from 0df68b7c0fb7f57aef2c10e80233394cc92fe894 or 64d9f000a5356ef5e089dff064a3759b93de95ce, the behaviour of `[ChannelX],jog` has changed and doesn't seem to react to the same value any more.

This has broken the the S4 MK3 mapping, but I suspect this would have broken other mapping as well. 

## Expected/previous behaviour

Nudge the jog wheel crown slows down or speeds up the track

## Current behaviour

Nothing noticeable

### Version

main

### OS

Ubuntu 22.04/Qt 6.2

Swiftb0y:
Fixes #13965. Please confirm.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2131,6 +2131,7 @@ add_executable(mixxx-test
   src/test/replaygaintest.cpp
   src/test/rescalertest.cpp
   src/test/rgbcolor_test.cpp
+  src/test/rotary_test.cpp
   src/test/ringdelaybuffer_test.cpp
   src/test/samplebuffertest.cpp
   src/test/sampleutiltest.cpp
--- a/src/test/rotary_test.cpp
+++ b/src/test/rotary_test.cpp
@@ -1,0 +1,28 @@
+#include "util/rotary.h"
+
+#include <gtest/gtest.h>
+
+TEST(Rotary, noOverFlow) {
+    Rotary rot(3);
+    EXPECT_EQ(rot.filter(3), 1);
+    EXPECT_EQ(rot.filter(3), 2);
+    EXPECT_EQ(rot.filter(3), 3);
+}
+
+TEST(Rotary, basicOverflow) {
+    Rotary rot(2);
+    EXPECT_EQ(rot.filter(2), 1);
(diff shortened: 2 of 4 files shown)


acolombier:
Will test that ASAP. Any chance to had some test coverage on this?

Swiftb0y:
like 65060ee9cba0e7935ccf0642cff6160d83a294b5?

acolombier:
Yeah, look like a good start! Any test is better than none ;) 
I think anything that we can run on the previous version (before the 2 commits mentioned in https://github.com/mixxxdj/mixxx/issues/13965) and on the new version would be a good way to ensure no regression there and in the future. I appreciate this is no easy task so happy with anything you feel comfortable doing there.

Swiftb0y:
Its just a simple moving average filter, I lets not overcomplicate things. The public interface of the type is deliberately small and I think these tests cover basically all of it. So I doing anymore is just bloat IMO.

Swiftb0y:
Merge?

acolombier:
If it's okay with you, I'd like to test it first. 
The issue I have IMO is that we do not have any test using the signal path in the reference buffer to ensure behaviour consistency, so it's hard to make sure that this small fix indeed restore the initial behaviour
I'll make sure to test that ASAP!

Swiftb0y:
>  The issue I have IMO is that we do not have any test using the signal path in the reference buffer to ensure behaviour consistency, so it's hard to make sure that this small fix indeed restore the initial behaviour

Yes, I agree. Though the tests you asked me to add explicitly test the previously broken code. I can rearrange the commits and then you would be able to see the test fail before the fix and be fixed after. Otherwise we can of course wait with merging until you tested it. I just wanted to make sure this didn't get lost and the issue is fixed asap.

acolombier (approved):
LGTM!

Tested and work as expected. Thank you for that quick fix and the added test cases!

Swiftb0y:
Thanks the testing.

Issue #5634, opened 2022-08-22
add multi level crate support


mixxxbot:
Reported by: **[alouw](https://launchpad.net/~alouw)**
Date: 2010-11-05T21:02:15Z
Status: In Progress
Importance: Wishlist
Launchpad Issue: [lp671632](https://bugs.launchpad.net/bugs/671632)

---

    request for multi level crate support to build similar to;
    Oldies -> 60's
    Oldies -> 70's 
    Oldies -> 80's 

    Or maybe even;
    Rock -> Metal -> Speed
    Rock -> Metal -> Trash

and so on

Russe:
As #1304 is closed, does it mean that nested crates would not be implemented?
I would also like nested crates, and I'm not sure if the custom tags feature is easy to use like nested crates would.

cr7pt0gr4ph7:
> Any news about multi level creates or/and playlists? I see this feature is still missing in Mixxx

I've actually got a working implementation of folders for creates (which can be found at [cr7pt0gr4ph7/mixxx:ready-for-merge/crate-folders](https://github.com/cr7pt0gr4ph7/mixxx/tree/ready-for-merge/crate-folders):

![image](https://github.com/user-attachments/assets/f093793f-8cf5-43f5-8ebf-c5b01329cefa)

I'm currently working on getting it into a mergeable state and resolving remaining usability issues:
- Avoiding unexpected/unwanted changes to the 
- **"How to avoid the `Crates' submenu of `WTrackMenu` filling the whole screen", i.e. how to prevent this**
  ![image](https://github.com/user-attachments/assets/4fdd5a1d-bf31-4b57-9ac6-495b0ba7f36f)
  The solution is probably will probably be very similar to what e.g. Spotify is doing:
  ![Spotify screenshot](https://github.com/user-attachments/assets/e1b81ada-8283-49bc-8b20-baf1208b7369)

cr7pt0gr4ph7:
I have decided to make folders a separate entity type from crates, so that tracks can only be added to and removed from crates (which are by definition the leaves of the tree) to avoid certain logical inconsistencies (and also because it was quicker to implement).

Namely, there is no good way to handle modifications to a merged list of crates:
1. Suppose that the you have the following structure, which I'm going to use as a minimal example:
   ```
   Folder/
     Crate A (containing tracks 1, 2)
     Crate B (containing tracks 2, 3)
     Crate C (containing track 4)
   ```
[...]

PR #13978, opened 2024-12-04
Grouped crates

Eve00000:
This is the [solution I proposed](https://github.com/mixxxdj/mixxx/issues/5634#issuecomment-2509744962) in #5634, as it continued spinning in my head since sunday, I had to create it, I couldn't concentrate on anything else. A sideproject. 
I hope I can make somebody happy with it.

In this PR I made an extra library feature by analogy to 'history' playlists.
In history playlists, playlists are 'grouped' on the year in creationdate.

For grouped crates you can select  in options->preferences->library->grouped crates 
to group crates based on a fixed length or a variable length with a delimiter (eg /////)

Grouped crates will make a 'group' for all crates starting with the same defined string.

This is NOT nested crates or multilevel crates, This adds 1 level, a grouping level to the cratesfeature.
For me this 1 level is more than enough but I can imagine that other people need a multilevel structure.

----->>>>>>
Multilevel is added + parent group shows all tracks in member crates and member groups
Selecting a group shows all tracks contained in membercrates.
-----
Multilevel when a variable length (mask) is selected and defined.

eg: if you have crates like
Pop - 60's
Pop - 70's
Pop - 80's
Pop - 90's
you can choose to group on 6 characters, the result will be:
group: Pop - 
crates: 60's / 70's / 80's / 90's

or you can choose to use variable length with delimiter ' - ' the result will be:
group: Pop
crates : 60's / 70's / 80's / 90's

the advantage with variable length with delimiter is that the crates 
Rock - 60's
Rock - 70's
Rock - 80's
Rock - 90's
New-Wave - 60's
New-Wave - 70's
New-Wave - 80's
New-Wave - 90's
will be processed in the same handling, the result will be 
group: Rock
crates: 60's / 70's / 80's / 90's

group: New-Wave
crates: 60's / 70's / 80's / 90's

all crate functions stay the same
groups with only 1 crate (all unique names with no similarity in the beginning of the name) are NOT displayed as a group, but as the crate itself, so you get a mixxx of crates and groups.

[20241112 - Eve Mixxx ](https://youtu.be/doS1Q99Q0sY)
20241212 Grouped crates - ungrouped
20241212 Grouped crates - grouped

-> with variable length (mask) groups crates can be multileveled grouped,  (sub)groups are only created if group has more then 1 crate, tracks in crates are displayed in all parent (sub)groups, group decoration when track is in one of the (sub)groups member crates.


acolombier:
I haven't had a chance to look at the code but I love the idea! This is basically why I haven't used crate as much as I wanted to. I particularly like the fact that this can be based off the name and doesn't need some complex parenting mecanism. Just one early feedback, I think it would be grate to have that as part of the main `Crate` feature and enable on demand, like you've done, instead of duplicating the library feature. Also, it would be amazing to allow multiple level

Eve00000:
> Making it opt-in to avoid surprises with any particular delimiters is a good idea.
You'd prefer a selection between possible delimiters? Can you list some?

> My library heavily relies on this for subgenres
Can you write / screenshot some examples?

> Do they show the union of all crates below (IMO that would be most useful)?
I was wandering about this to, iTunes makes the Union. But this means recalculating the groups elements each time, even when it's just passing by with up/down ... could be a lot for slower machines and it would make some users unhappy when their machine gets slow.

fwcd:
> You'd prefer a selection between possible delimiters? Can you list some?

No, I think a text box is just fine. What I meant is that we should leave this setting off by default to avoid surprising users that e.g. already use "/" or "-" as part of an intentionally flat crate hierarchy.

> Can you write / screenshot some examples?

```
Genres
  House
    Big Room House
    Deep House
    Tech House
    ...
```

For ballroom dances I do something similar:

```
Dancing
  Standard
    Slow Waltz
    Quickstep
    Tango
    ...
  Latin
    Cha-Cha-Cha
    Rumba
    ...
```

> I was wandering about this to, iTunes makes the Union. But this means recalculating the groups elements each time, even when it's just passing by with up/down ... could be a lot for slower machines and it would make some users unhappy when their machine gets slow.

I think it might be worth benchmarking that. If we can compute the union in a single SQL query, this doesn't necessarily have to be slow, SQLite is pretty good at churning through huge datasets. And given that we're probably "just" dealing with libraries on the order of tens of thousands of tracks (in the common case), I don't see why this should be much more of a performance issue than in other programs like iTunes.

Eve00000:
> No, I think a text box is just fine. What I meant is that we should leave this setting off by default to avoid surprising users that e.g. already use "/" or "-" as part of an intentionally flat crate hierarchy.
 
Top

>     Rumba ...

I understand, I would have merged some group titles but I do know that it's important that people find themselves  'at ease' .
How do you use the library at the moment?
Crates Like 
Dancing // Standard // Slow Waltz
Dancing // Latin // Cha-Cha-Cha
How many levels would you like, I see you use the root level + 2 sublevels.
I propose using the same delimiter (to make it not to complicated) and  use it only with the variable length with delimiter.

This wil result in an extra sqlsearch like ' select .... where id = 'idOf1stGroupmember' or ...  id = 'idOfLastGroupmember'.
I"m already using this in [smarties](https://github.com/Eve00000/mixxx/pull/13)  for some crates / playlists (also history), the difference: in smarties I gave the locked the meaning of cached. 
Let's try it :-)

Eve00000:
While using pr #14130 I came to the conclusion 'just replacing' crates with 'grouped crates' is not working for me.
There are moments I want to see both crates & grouped crates, to get the naming correct, to have a complete overview.

I closed pr #14130 and made an update for this "grouped crates" pr #13978 with a extra selection in preferences 'replace crates with grouped crates'  so that users can select to see only crates / only grouped crates or both (like me).

TheTerwiz:
Preferences section for crate groups is a bit garbled at least on my screen:
image

It would be nice to have ability to drag and drop crates from group to another. At the moment moving a crate from group to another requires renaming the crate, which is a bit non-intuitive.
Maybe this could be implemented with separate crate group object instead of the "crate-name-based-hierarchy?"
It's great that user can opt-in/out this feature.
It would be nice to have this kind of grouping feature also for playlists.

I haven't looked at the code, so take my suggestions with a grain of salt. I have no idea how hard it is to implement what I have suggested.

IMHO this is definitely step into right direction, and even as-is it would be a useful feature for me, thank you.

Eve00000:
> Preferences section for crate groups is a bit garbled at least on my screen: !
> It would be nice to have ability to drag and drop crates from group to another. At the moment moving a crate from group to another requires renaming the crate, which is a bit non-intuitive. Maybe this could be implemented with separate crate group object instead of the "crate-name-based-hierarchy?" It's great that user can opt-in/out this feature. It would be nice to have this kind of grouping feature also for playlists.

Hi, @TheTerwiz ,
Thank you for the feedback.
1. I'll take a look at the preferences window, may I ask your screen's resolution setting? I tested it with 1080.  It looks like the window is to small to show the whole line.
Did you try to stretch the preferences window?

2. As I mentioned in the name and the description, this pr is about GROUPED crates and not about NESTED crates.
To be able to drag groups & crates from 1 group to another the whole library needs to be rewritten for NESTED containers.
As I'm trying to solve some old issues which I see as 'basic' I wanted to create a way to 'orden' crates / playlists / smarties (smartplaylists = WIP coming up). 
As I don't have the ambition / motivation to rewrite the whole library I thought about a 'simple' solution as the 1st step was (only 1 level) fast. With the questions for multilevel + combining all tracks from child-containers in the parent container it got rather complicated as well 😃 
Nested crates could cause a lot of errors & mistakes, dragging wrong, mouse / touch problem... 
Grouped crates are much more stable IMO as you need to rename them on purpose. 

So please excuse me but this is as far as I want to go : grouped crates (grouped playlist & smarties are comming up if I got enough positive feedback 😃 ) 
IMO it's very useful to orden your lists and get a much smaller / cleaner default list, if you use crates & grouped crates at the same time you can see what's happening. Renaming is only pressing F2 :-) 

Extra advice: dont use spaces around your delimiter, in your case use group1#subgroup2#subsubgroup3#crate1 and maybe use some more signs for the delimiter like ### 
Please enjoy it, and if you're curious take a look at the code 😄

Eve00000:
@acolombier @fwcd 
May I invite you to review this pr?

Eve00000:
@TheTerwiz , you were keen on having the 'grouped' feature for playlists to, right?
Can you test it? You can find it in my [pr #19](https://github.com/Eve00000/mixxx/pull/19), I want to expand the history - grouping a little before putting it in public.
But you can test.

laex333:
Unfortunately I cannot test it for real right now; but I want to thank and congratulate you for this very smart and straightforward implementation! It looks like exactly the behaviour I was hoping for, meeting my needs perfectly!

What is shown in your Youtube-video is already sufficient for me. I will be very happy with it and I am looking forward to getting this greatly appreciated feature soon on my new machine that is in the making...

Eve00000:
Thank you for the feedback!

> I personally wouldn't want the groups to expand all the time automatically
--> you have to expand the manually 
> number of contained tracks and total play time in a collapsed parent crate...
--> groups arte no objects, just temporary 'labels', to give the a summary the calculations need to be done on every selection.
      and the difference between a crate / group would be more difficult. 
> In this case it should still only appear once in the parent crate.)
--> Tracks appear only once

PR #13980, opened 2024-12-04
Fix crash due to concurrent access in MidiController

### LINKED ISSUE #13940: qhash.h:790 ASSERT on multiple MIDI messages in quick succession
### Bug Description

When my script sends multiple MIDI messages in quick succession, it often makes Mixxx trigger an assert;
```
 [Controller] fatal /usr/include/qt6/QtCore/qhash.h:790 ASSERT: "!isUnused()" in file /usr/include/qt6/QtCore/qhash.h, line 790
```

### Version

_No response_

### OS

_No response_

daschuer:
This fixes #13940 
The fix is implemented by locks around m_pMapping and reference counting. 

The locks are either a mutex or a spin lock depending on the compiler support. 
I have applied them everywhere, because it is not obvious which threads are readers and writers. 

There is potential for further refactoring, because the Mapping is finally no longer shared when used. This is however out of scope of this band aid PR.

### DIFF
--- a/src/controllers/bulk/bulkcontroller.cpp
+++ b/src/controllers/bulk/bulkcontroller.cpp
@@ -283,8 +283,8 @@ void BulkController::send(const QList<int>& data, unsigned int length) {
 
 void BulkController::sendBytes(const QByteArray& data) {
     VERIFY_OR_DEBUG_ASSERT(!m_pMapping ||
-            m_pMapping->getDeviceDirection() &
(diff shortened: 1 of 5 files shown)


JoergAtGithub:
Such threading issues require a longer testing period, than we have until 2.5.0.

daschuer:
Closing this in favor of https://github.com/mixxxdj/mixxx/pull/14159

PR #13983, opened 2024-12-05
chore: upgrade macos runner to macos-13 Ventura and xcode15

daschuer:
This applies again failing changes form #13606. 
Since macos-12 is no longer available we have no other chance. 

I apply this to 2.4 to keep the CI allive. until 2.5.0 is finally released.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,8 +60,8 @@ jobs:
             artifacts_path: build/*.deb
             artifacts_slug: ubuntu-jammy
             qt_qpa_platform: offscreen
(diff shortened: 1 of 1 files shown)


Swiftb0y:
Since its almost always packaging that fails, should we consider temporarily disabling it on macos?

Swiftb0y:
yeah, that would be better than a failing job or no packages (or failing CI forcing us to restart it, unnecessarily running through the entire build and test suite again). I would highly appreciate if you could open a PR that implements this for us.

fwcd:
I've opened #13991

PR #13988, opened 2024-12-05
refactor: remove samplew_autogen.h

Swiftb0y:
As it doesn't have much use in practice. Currently only the variants for up to 3 buffers are used. This removes all unused implementations along with the script that generated them. It also un-inlines their definitions as there likely isn't much parameter passing overhead (compared to other factors such as increased instruction-cache pressure created from inlining long, highly vectorized and loop-unrolled code).

Alternative to #13928, simply removing the unneeded variants instead of replacing them with a block of template metaprogramming.


daschuer (approved):
LGTM, Thank you.

JoergAtGithub:
This PR breaks the build on Windows!

PR #13990, opened 2024-12-05
(fix/UX) set scrollSafeGuard also for controller settings widgets

ronso0:
... like for all other input widgets in the preferences.

Didn't work right away because the [current way to 'safe' all child input widgets](https://github.com/mixxxdj/mixxx/blob/d18764a9258f1400f7d9922df6ba7210e4c5d18d/src/preferences/dialog/dlgpreferencepage.cpp#L24-L27) via `QObject::children()` + `qobject_cast<Type*>(QObject*)` requires Q_OBJECT for all setting widget sub-classes (not just the base class).
However "Template classes not supported by Q_OBJECT"...

I'm pedantic but I'm also lazy so I didn't try to fix the Q_OBJECT stuff (no clue where to start anyway..) but found another way (which is also much faster¹ : ) -> first commit.

Works beautifully.
Fixing the scroll guard for Qt6 en passant.

¹<sub>for example Library preferences: ~200 μs vs. ~700 μs</sub>

### DIFF
--- a/src/preferences/dialog/dlgpreferencepage.cpp
+++ b/src/preferences/dialog/dlgpreferencepage.cpp
@@ -21,39 +21,16 @@ QUrl DlgPreferencePage::helpUrl() const {
     return QUrl();
 }
 
-void DlgPreferencePage::setScrollSafeGuardForAllInputWidgets(QObject* obj) {
+void DlgPreferencePage::setScrollSafeGuardForAllInputWidgets(QObject* pObj) {
     // Set the focus policy to for scrollable input widgets and connect them
-    // to the custom event filter
-    for (auto* ch : obj->children()) {
-        // children() does not descend into QGroupBox,
-        // so we need to do it manually
-        QGroupBox* gBox = qobject_cast<QGroupBox*>(ch);
-        if (gBox) {
-            setScrollSafeGuardForAllInputWidgets(gBox);
-            continue;
-        }
-
-        QComboBox* combo = qobject_cast<QComboBox*>(ch);
-        if (combo) {
-            setScrollSafeGuard(combo);
-            continue;
-        }
-        QSpinBox* spin = qobject_cast<QSpinBox*>(ch);
-        if (spin) {
-            setScrollSafeGuard(spin);
-            continue;
-        }
-        QDoubleSpinBox* spinDouble = qobject_cast<QDoubleSpinBox*>(ch);
-        if (spinDouble) {
-            setScrollSafeGuard(spinDouble);
-            continue;
-        }
-        QSlider* slider = qobject_cast<QSlider*>(ch);
-        if (slider) {
-            setScrollSafeGuard(slider);
-            continue;
-        }
-    }
+    // to the custom event filter.
+    // Note: finding all relevant widgets with pObj->findchildren<Type*>
+    // is much faster than with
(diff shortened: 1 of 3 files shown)


acolombier (approved):
Code looks good! I would like to give that a go before merging

ronso0:
Of course.
FWIW I tested with MIDI Through + Numark Scratch mapping.
With the S4 Mk3 mapping I notice that scrolling is not propagated to the page when the cursor hovers a combobox while scrolling. Ie. scroling stops.
I wrote a TODO earlier, will check if the fix works.

ronso0:
Done.

src/preferences/dialog/dlgpreferencepage.cpp
```diff
@@ -21,38 +21,30 @@ QUrl DlgPreferencePage::helpUrl() const {
...
-            continue;
-        }
+    // to the custom event filter.
+    // Note: finding all relevant widgets with pObj->findchildren<Type*>
+    // is much faster than with
+    // for (auto* ch : pObj->children()) { qobject_cast<Type*<(ch); }
+    auto pCombos = pObj->findChildren<QComboBox*>();
+    for (auto* pCombo : pCombos) {
+        setScrollSafeGuard(pCombo);
+    }
```
Swiftb0y, line 33: nit: add another function that shortens this to one function template call. A private template function would work. You can declare it in the header and then only implement it in the `.cpp` using 
```cpp
template<typename T>
void DlgPreferencePage::setScrollSafeGuardForChildrenOfType(QObject* pObj) {
    QList<T*> children = pObj->findChildren<T*>();
    for (T* pChild : children) {
        setScrollSafeGuard(pChild);
    }
}

File: src/preferences/dialog/dlgpreferencepage.cpp [hunk 2]
```diff
@@ -21,38 +21,30 @@ QUrl DlgPreferencePage::helpUrl() const {
...
-        // so we need to do it manually
-        QGroupBox* gBox = qobject_cast<QGroupBox*>(ch);
-        if (gBox) {
-            setScrollSafeGuardForAllInputWidgets(gBox);
-            continue;
-        }
+    // to the custom event filter.
+    // Note: finding all relevant widgets with pObj->findchildren<Type*>
+    // is much faster than with
+    // for (auto* ch : pObj->children()) { qobject_cast<Type*<(ch); }
```
Swiftb0y, line 29: nit:
```suggestion
    // for (auto* ch : pObj->children()) { qobject_cast<Type*>(ch); }
```

src/preferences/dialog/dlgpreferencepage.cpp
ronso0: Perfect, thanks!
Took me a while to figure I need to call it like this
```cpp
setScrollSafeGuardForChildrenOfType<QComboBox>(pObj);
```

Swiftb0y: yeah. sorry for not providing an example.

Swiftb0y (approved):
Thank you. LGTM. please rebase.

src/preferences/dialog/dlgpreferencepage.cpp
ronso0: Never mind, I didn't mean it like that actually. It's just been late and I didn't notice I'm using typename functions all the time, like here, qobject_cast.

PR #13653, opened 2024-09-15
S4 Mk3 screen support

acolombier:
This PR include support for the S4 Mk3 screen support. It is made to work without #12199 although there is a setting to enable communication between mapping. This way, I can maintain a single mapping definition, without needing to get an API commitment for communication between mapping.

It includes two themes:
- Traktor Pro stock inspired theme
image
- Joe Easton's inspired theme
image

Depends on:
- [x] #13669 
- [x] #13737
- [x] #13470
- [x]  #14606

Here is the list of improvement made on the core mapping
- Beatjump tab
- Stem tab
- Support for beatloop roll inside a loop
- Screen feedback (optional, require #12199 - temporary merge available on `feat/s4-mk3-screen-support-with-shared-data`)
- Add library page scroll
- Ability to preview the active deck hotcue
- Ability change the hotcue color

PR #13997, opened 2024-12-06
(fix) Traktor Kontrol S4 Mk3: revert QuickEffect preset offset

ronso0:
0 is now the empty item ---
1 is the first item etc.

To clarify:
previously [1] would select the 3rd item = preset 2
now [1] selects preset 1

### DIFF
--- a/res/controllers/Traktor-Kontrol-S4-MK3.js
+++ b/res/controllers/Traktor-Kontrol-S4-MK3.js
@@ -1262,7 +1262,7 @@ class FXSelect extends Button {
         if (this.mixer.firstPressedFxSelector !== null) {
             for (const deck of [1, 2, 3, 4]) {
                 const presetNumber = this.mixer.calculatePresetNumber();
-                engine.setValue(`[QuickEffectRack1_[Channel${deck}]]`, "loaded_chain_preset", presetNumber + 1);
+                engine.setValue(`[QuickEffectRack1_[Channel${deck}]]`, "loaded_chain_preset", presetNumber);
             }
         }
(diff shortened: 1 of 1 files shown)


acolombier (changes requested):
The reason why `0` was removed is because, as per [the doc](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-[EffectRack1_EffectUnitN]-loaded_chain_preset), `0` will unselect the quick FX, so when assigning `FX 1` to a deck (or a stem starting from #13653), which made a very poor user experience. 
If you think this should still be possible, please make that a configurable setting.

ronso0:
Yeah, I thinks it's fixed now.
@acolombier please take a look.

ronso0:
> The reason why `0` was removed is because, as per [the doc](https://manual.mixxx.org/2.4/en/chapters/appendix/mixxx_controls.html#control-%5BEffectRack1_EffectUnitN%5D-loaded_chain_preset), `0` will unselect the quick FX, so when assigning `FX 1` to a deck (or a stem starting from #13653), which made a very poor user experience. If you think this should still be possible, please make that a configurable setting.

Umm, without this fix [1] would select the 3rd item (preset 2).
I just tested it again: [1] now selects the 1st preset, 0 can't be selected.

acolombier:
Right, looking at `FXSelect`, it looks like the computation always return 1 indeed. I'm AFK now, but I will test that later. I cannot remember why I added that offset, but it could also have been the initial Be's mapping definition.

acolombier (approved):
Tested and work as expected! Thank you

PR #13999, opened 2024-12-07
Merge 2.4 into 2.5

Swiftb0y:
Lets see what CI says about this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -359,7 +359,9 @@ jobs:
           timeout_minutes: 30
           max_attempts: 12
           retry_wait_seconds: 1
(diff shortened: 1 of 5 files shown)


JoergAtGithub:
LGTM! CI is pass! Thank you!

ronso0:
Great!
I tried to merge 2.5 into main the other day, but there are conflicts I don't know how to deal with, especially the translations.
Could someone take a look please?

Holzhaus:
@Swiftb0y we cannot do merges like this if the comment in our `.gitattributes` is correct. Because these kind of GUI merges do not honor it, at least as far I remember.

Swiftb0y:
right, but no translation changes were made in this case judging by the diff. So in this specific case I'm fairly certain we didn't break anything.

PR #14003, opened 2024-12-07
Fix for loading MP4 files with Unicode filenames on windows.

nickwhaley:
Attempt to load a MP4 format file when on Windows that has Unicode files names will result in (specifically Tiësto - The Business):

Exception thrown at 0x00007FF87292EF09 (avformat-60.dll) in mixxx.exe: 0xC0000005: Access violation reading location 0x0000000000000000.

The function call here expects UTF8 encoding. Using toLocal8Bit will make UTF8 on linux but not on Windows. Use toUtf8 instead.


Swiftb0y:
Thanks for the PR. This is identical to #13784 which never got merged. @Eve00000 how do you want to proceed?

Swiftb0y:

nickwhaley:
Closing as duplicate.

Issue #14000, opened 2024-12-07
Traktor Kontrol S4 Mk3: issue with slow scratching


ronso0:
### Bug Description

I noticed that when scratching very slowly in one direction (also when trying to hold the wheel steady in order to set a hotcue), the play position makes tiny jumps forwards and backwards resulting in unexpectedly loud sound artifacts.

However, scratching sounds pretty smooth when doing similarly slow/small moves with the waveform or with my old, rather low-res wheel on the Reloop TerminalMix 2/4 (the mapping does no filtering/averaging).
_Yes, there are artifacts, much louder, higher pitch than comparable moves with vinyl of course, but they are still more quiet than with the S4 mapping._

I saw the wheel also sends _absolute position data_ which is not used for certain reasons.
Could that be used when the wheel moves slowly?

I'll compare with the mappings in 2.4 or 2.5

https://github.com/mixxxdj/mixxx/blob/8354c8e0f57a635acb7f4b3cc16b9745dc83312c/res/controllers/Traktor-Kontrol-S4-MK3.js#L2439-L2474

ronso0:
Furthermore if I rotate the wheel veeeery slowly in one direction I can move it quite a distance before anything happens.

I tried the old way in the S4 mapping an it's much smoother:
* enable scratching:
   ```js
  if (touched) {
      engine.scratchEnable(this.deck.currentDeckNumber,
        2150, // matches LED position @33 1/3
        33+1/3,
        1.0/8,
        1.0/256);
  } else {
      engine.scratchDisable(this.deck.currentDeckNumber);
  }
   ```
* scratch: simply use `diff` for `engine.scratchTick()`
   ```js
    case wheelModes.vinyl:
        if (this.deck.wheelTouch.touched || engine.getValue(this.group, "scratch2") !== 0) {
            engine.scratchTick(this.deck.currentDeckNumber, diff);
        } else {
            engine.setValue(this.group, "jog", this.speed);
        }
        break;
   ```

This is my temporary fix until someone more familiar with this will take a look :shrug:

ronso0:
This is the Zulip thread https://mixxx.zulipchat.com/#narrow/channel/113295-controller-mapping/topic/engine.2EscratchTick.28.29.20or.20setValue.28group.2C.20.22scratch2.22.2C.20speed.29.20.3F

and my fix #14004

PR #14004, opened 2024-12-08
(DEBUG) S4 Mk3: compare scratch functions `engine.scratch..` vs `engine.setValue(group, "scratch2..")`

ronso0:
Helper to fix/debug #14000 

This adds two options for scratching:
* 3 scratch modes:
   curr: `engine.setValue(group, "scratch2..)`
   alt1: `engine.scratchEnable()`/`engine.scratchTick()` 
   alt2: `scratch_position_enable` / `scratch_position` (currently used for waveform scratching)
* sync scratch speed with wheel LED speed (~2x as fast as the current scratch speed)

With the alternative methods (scratchTick, wheelPos) scratching sounds pretty smooth **at low speeds**, while with the current implementation it's rather just noise (no buffer underruns, really just how wheel ticks are filtered & emitted)

**Another notable difference:**
with the `scratchTick()` method the track is very slowly drifting backwards when doing many (40+) fast + short back & forth scratches¹.

### My results:
**curr: `scratch2`:** 
* noisy at low speeds
* small drift for fast back'n'forth scratches¹
* no notable drift for one-direction scratches
(some revs forth, then same backwards, should land at the same position)

**alt1: `scratchTick`:**
* smooth sound at all speeds
* drifts backwards  for fast back'n'forth scratches¹
* no notable drift for one-direction scratches

**alt2: `scratchPos`:**
* bit noisy at very low speeds
* **no drift** whatsoever!

### Bottom line:
So, to get the best of both methods, could we use the alt. method for very slow scratches?
~~I'll test that soon.~~
**edit** No, I won't. I'm using `scratchPos` and it suits my needs. Sounds good a mid/high speeds, no drift.

Or fix the filtering for the current method?
I have no clue how..

### DIFF
--- a/res/controllers/Traktor-Kontrol-S4-MK3.js
+++ b/res/controllers/Traktor-Kontrol-S4-MK3.js
@@ -170,6 +170,23 @@ const BeatJumps = [
 // the motor if enable. Recommended value are 33 + 1/3 or 45.
 // Default: 33 + 1/3
 const BaseRevolutionsPerMinute = engine.getSetting("baseRevolutionsPerMinute") || 33 + 1/3;
+const ScratchSyncToWheelLED = !!engine.getSetting("scratchSyncToWheelLED");
+
+const scratchModes = {
+    scratch2: 0,
+    scratchTick: 1,
+    wheelPos: 2,
+    wheelPosNew: 3,
+};
+const ScratchMode = scratchModes[engine.getSetting("scratchMode")] || scratchModes.scratch2;
+// Parameters for scratchTick mode
+// alpha and beta are the recommended start values from
+// https://github.com/mixxxdj/mixxx/wiki/midi%20scripting#scratching-and-jog-wheels
+const ScratchAlpha = 1.0/8;
+const ScratchBeta = ScratchAlpha/32;
+const ScratchTicksPerRev = ScratchSyncToWheelLED ? 2150 : 6750;
+const ScratchSpeedFactor = ScratchSyncToWheelLED ? 2.14 : 1;
+const HotcuePressStopsScratching = !!engine.getSetting("hotcuePressStopsScratching");
 
 // Define whether or not to use motors.
 // This is a BETA feature! Please use at your own risk. Setting this off means that below settings are inactive
@@ -634,7 +651,7 @@ class Deck extends ComponentContainer {
 
         this.selectedStem.fill(false);
 
-        engine.setValue(this.group, "scratch2_enable", false);
+        this.deck.wheelTouch.toggleScratching(false);
         this.group = newGroup;
         this.color = this.groupsToColors[newGroup];
 
@@ -644,7 +661,7 @@ class Deck extends ComponentContainer {
 
             if (this.wheelMode === wheelModes.motor) {
                 engine.beginTimer(MotorWindUpMilliseconds, () => {
-                    engine.setValue(newGroup, "scratch2_enable", true);
+                    this.deck.wheelTouch.toggleScratching(true);
                 }, true);
             }
         }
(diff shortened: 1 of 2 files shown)


res/controllers/Traktor-Kontrol-S4-MK3.js
```diff
@@ -2413,8 +2423,16 @@ class S4Mk3Deck extends Deck {
...
+                        if (useEngineScratch) {
+                            engine.scratchEnable(this.deck.currentDeckNumber,
+                                ScratchIntervalsPerRev,
+                                BaseRevolutionsPerMinute,
+                                ScratchAlpha,
+                                ScratchBeta); // ramoing true by default
+                        } else {
+                            engine.setValue(this.group, "scratch2_enable", true);
+                        }
+                    } else { // release
```
acolombier, line 2435: This looks wrong - seems like `engine.setValue(this.group, "scratch2_enable", true);` is always run both in `if` and `else` if `useEngineScratch` isn't used

acolombier (commented):
I haven't had a chance to test yet, but I'll try to do that ASAP

res/controllers/Traktor-Kontrol-S4-MK3.js
ronso0: I don't understand, sorry. The diff looks good, I just added an alternative path for every enable/disable branch.
But indeed, with `engine.scratchEnable()` `scratch2_enable` is also set true (in c++).

Just give a spin (haha!) when you find some time and let me know how it works for you.

acolombier (changes requested):
The new scratching experience feels very good and precision change doesn't feel as bad as I feared. I'm happy to get those two settings as part of the core mapping. :+1: 

Only "regression" from my perceptive is the hotcue not "cutting" the scratch anymore. 
My routine being, I would scratch over, say a voice sampler and would be able to resume playing instantly by activating a hotcue. How would you feel about restoring that behaviour via an alternative setting? (e.g `Stop scratch when activating hotcue` and slight refactor of L2458-2470)

Another bug is, if you set a jogwheel tension higher than `0.80`, the jog will speed up and never stop! I guess you might be able to adjust the range.

Let me know if you would want me to suggest patches

ronso0:
> Only "regression" from my perceptive is the hotcue not "cutting" the scratch anymore.
How would you feel about restoring that behaviour via an alternative setting? (e.g Stop scratch when activating hotcue and slight refactor of L2458-2470)

Sure, I don't plan to remove features, though I'd make the cutting opt-in.

ronso0:
> The new scratching experience feels very good and precision change doesn't feel as bad as I feared. I'm happy to get those two settings as part of the core mapping. 👍

Great to hear!
So we'd have 2 new scratching options then, both opt-in:
* make scratch speed match wheel LED speed
* seeking to hotcue stops scratching

Or where you saying the engine.scratch.. functions should be optional, too?
Given that both this and scratch2 show sticker drift and engine.scratch.. sounds much smoother at low speeds I'd actually drop scratch2.
Wdyt?

ronso0:
Bit weird: I've been using PositionScratchController lately and it works smoothly, but it seems I get let's say variable results. Fast scratching, eg. when drop-cueing a track, sometimes sounds rather harsh. Not sure where that stems from.
Will revert to scratchTick() until I figured what's wrong. Bit of a pity since PositionScratchController sounds much better at slow rates.

ywwg:
Is this something we should look at merging officially for making the MK3 feel better?  Even as a selectable option

ronso0:
Yup, IMO that would make sense.
I'll test again, but as far as I remember the scratch feeling was better. Especially as it eliminates the drift entirely.

Did you test it @ywwg ?

Dunno when I have the nerves to rebase and test, but the conflict is probably only because of the new "stop scratch on hotcue press" option.

PR #14005, opened 2024-12-09
Fix broken Windows build

JoergAtGithub:
Removed src/util/sample_autogen.h from CMakeLists.txt to fix broken PCH build.
This is a regression introduced by #13988.


Holzhaus:
Just saying that branch protection would have prevented this 😛

wimstefan:
This happened also on Linux ...
Gentoo in my case. 
Commenting out the line in Cmakelist.txt fixed it for me.

PR #14006, opened 2024-12-09
Reorganize content into tabs of controller preferences

JoergAtGithub:
- Convert groupBoxSettings/groupBoxScreens into tabs
- Show only the tabs, which have valid content in the current context
- Sort tabs that settingsTab is shown by default, if controller settings are defined in mapping
- Moved controller tabs below info and warning section
- Made hyperlink warning label clickable and warning text select- and copy-able
- Made warning label collapsing in height
- Hide 'Mapping Info' fields if there is no content

This is a preparatory GUI cleanup, prior to the addition of further properties in additional tabs.

Before:
grafik

This PR:
grafik

### DIFF
--- a/src/controllers/dlgprefcontroller.cpp
+++ b/src/controllers/dlgprefcontroller.cpp
@@ -80,7 +80,9 @@ DlgPrefController::DlgPrefController(
           m_pOutputTableModel(nullptr),
           m_pOutputProxyModel(nullptr),
           m_GuiInitialized(false),
-          m_bDirty(false) {
+          m_bDirty(false),
+          m_inputMappingsTabIndex(-1),
+          m_outputMappingsTabIndex(-1) {
     m_ui.setupUi(this);
     // Create text color for the file and wiki links
     createLinkColor();
@@ -191,6 +193,10 @@ DlgPrefController::DlgPrefController(
     }
 
     m_ui.groupBoxWarning->hide();
+    m_ui.labelWarning->setTextFormat(Qt::RichText);
+    m_ui.labelWarning->setTextInteractionFlags(Qt::TextBrowserInteraction |
+            Qt::LinksAccessibleByMouse | Qt::TextSelectableByMouse);
+    m_ui.labelWarning->setOpenExternalLinks(true);
     m_ui.labelWarning->setText(tr(
             "<font color='#BB0000'><b>If you use this mapping your controller "
             "may not work correctly. "
@@ -205,6 +211,7 @@ DlgPrefController::DlgPrefController(
             "href='https://mixxx.org/wiki/doku.php/"
             "controller_engine_versions'>Controller Engine Versions</a>.")
                                        .arg("2", "1"));
+
     QIcon icon = style()->standardIcon(QStyle::SP_MessageBoxWarning);
     m_ui.labelWarningIcon->setPixmap(icon.pixmap(50));
 
@@ -242,7 +249,7 @@ DlgPrefController::DlgPrefController(
 #endif
 
     // Open script file links
-    connect(m_ui.labelLoadedMappingScriptFileLinks,
+    connect(m_ui.labelMappingScriptFileLinks,
             &QLabel::linkActivated,
             [](const QString& path) {
                 mixxx::DesktopHelper::openUrl(QUrl::fromLocalFile(path));
@@ -296,6 +303,10 @@ DlgPrefController::DlgPrefController(
             &QPushButton::clicked,
             this,
             &DlgPrefController::slotOutputControlSearch);
+
+    // Store the index of the input and output mappings tabs
+    m_inputMappingsTabIndex = m_ui.controllerTabs->indexOf(m_ui.inputMappingsTab);
+    m_outputMappingsTabIndex = m_ui.controllerTabs->indexOf(m_ui.outputMappingsTab);
 }
 
 DlgPrefController::~DlgPrefController() {
@@ -429,38 +440,6 @@ QString DlgPrefController::mappingShortName(
     return mappingName;
 }
 
-QString DlgPrefController::mappingName(
-        const std::shared_ptr<LegacyControllerMapping> pMapping) const {
-    if (pMapping) {
-        QString name = pMapping->name();
-        if (name.length() > 0) {
-            return name;
-        }
-    }
-    return tr("No Name");
-}
-
-QString DlgPrefController::mappingDescription(
-        const std::shared_ptr<LegacyControllerMapping> pMapping) const {
-    if (pMapping) {
-        QString description = pMapping->description();
-        if (description.length() > 0) {
-            return description;
-        }
-    }
-    return tr("No Description");
-}
-
-QString DlgPrefController::mappingAuthor(
-        const std::shared_ptr<LegacyControllerMapping> pMapping) const {
-    if (pMapping) {
-        QString author = pMapping->author();
-        if (author.length() > 0) {
-            return author;
-        }
-    }
-    return tr("No Author");
-}
 
 QString DlgPrefController::mappingSupportLinks(
         const std::shared_ptr<LegacyControllerMapping> pMapping) const {
@@ -587,10 +566,8 @@ void DlgPrefController::enumerateMappings(const QString& selectedMappingPath) {
     QString newMappingFilePath = mappingFilePathFromIndex(index);
     if (index == -1) {
         m_ui.chkEnabledDevice->setEnabled(false);
-        m_ui.groupBoxSettings->setVisible(false);
     } else {
         m_ui.comboBoxMapping->setCurrentIndex(index);
-        m_ui.chkEnabledDevice->setEnabled(true);
     }
     m_ui.comboBoxMapping->blockSignals(false);
     if (newMappingFilePath != currentMappingFilePath) {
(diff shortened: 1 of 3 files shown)


src/controllers/dlgprefcontroller.h
```diff
@@ -140,4 +137,6 @@ class DlgPrefController : public DlgPreferencePage {
     ControllerMappingTableProxyModel* m_pOutputProxyModel;
     bool m_GuiInitialized;
     bool m_bDirty;
+    int m_inputMappingsTabIndex;  // Index of the input mappings tab
+    int m_outputMappingsTabIndex; // Index of the output mappings tab
```
Swiftb0y, line 141:
```suggestion
    int m_inputMappingsTabIndex;  // Index of the input mappings tab in m_ui.controllerTabs
    int m_outputMappingsTabIndex; // Index of the output mappings tab in m_ui.controllerTabs
```

Swiftb0y (commented):
LGTM otherwise. Thank you.

Swiftb0y (approved):
LGTM. Waiting for CI. @ronso0 do you have interest to take a quick look at the `.ui` changes?

ronso0:
Umm, looking at `.ui` diffs is pointless here. Before and after grid/layout items are in random order (that's why I usually reorder them so GUI order matches .ui order).
And I can't built main currently because I don't have hidapi 0.14.0

So if the screenshots reflect the current state of this PR and UX (correct tabstops etc.) is the same this LGTM :+1: 
Just the grids/2-column-layouts in Device Info / Mapping Info could be optimised so that eg. Mapping -> Description can use the available space.

JoergAtGithub:
I reordered the .ui file in GUI order.
I added a tabstop section.
But I could not find a way to use all the available vertical space in the grids/2-column-layouts in Device Info / Mapping Info. I tried for hours, and the only partially working solution was to set the spacer height from C++, which I consider more ugly than a little uneccessary whitespace.

ronso0:
I'll take a look asap, can build again with a libhidapi backport.

src/controllers/dlgprefcontroller.cpp
```diff
@@ -992,25 +988,51 @@ void DlgPrefController::slotShowPreviewScreens(
...
+    m_ui.labelMappingAuthor->setVisible(!author.isEmpty());
+    m_ui.labelMappingAuthorValue->setVisible(!author.isEmpty());
+    m_ui.labelMappingSupport->setVisible(!supportLinks.isEmpty());
+    m_ui.labelMappingScriptFiles->setVisible(!scriptFileLinks.isEmpty());
+    // We do not hide the support links label and
+    // the script files label if they are empty,
+    // because without them the layout would be collapse.
+
+    // Remove all tabs first
+    m_ui.controllerTabs->clear();
```
ronso0, line 1021: Why remove and insert tabs?
There is also [`QTabWidget::setTabVisible(int index, bool visible)`](https://doc.qt.io/qt-6/qtabwidget.html#setTabVisible)
IIUC tab indices are unchanged (unlike visual position in ui) so we'd only need to store them in the ctor.

This way we'd also get around double tab tr strings (here and in ui file)

ronso0: same applies to the screens and settings tab, of course.

src/controllers/dlgprefcontroller.cpp
```diff
@@ -1019,13 +1041,15 @@ void DlgPrefController::slotShowMapping(std::shared_ptr<LegacyControllerMapping>
...
 
-        if (settings.isEmpty()) {
-            m_ui.groupBoxSettings->setVisible(false);
-        } else {
-            m_ui.groupBoxSettings->setVisible(true);
-            setScrollSafeGuardForAllInputWidgets(m_ui.groupBoxSettings);
+    // Show or hide the settings tab based on the presence of settings
+    if (pMapping && !pMapping->getSettings().isEmpty()) {
+        if (m_ui.controllerTabs->indexOf(m_ui.settingsTab) == -1) {
+            m_ui.controllerTabs->addTab(m_ui.settingsTab, tr("Mapping Settings"));
```
ronso0, line 1049: This is obsolete with `settabVisible()`, so just FYI to avoid duplicate tr strings here and n ui file, we could use `QTabwidget::tabText(QTabwidget::indexOf(QWidget*))`

ronso0 (changes requested):
my first review, will take another look soonish

ronso0:
A regression I noticed is that now the Input/Output tabs can't use the entire height of the dialog as before, mapping and device info will always occupy space at the top. Also 

One solution would be to allow collapsing groupboxes.
1. subclass QGroupBox, add collapse/expand buttons, which I'd appreciate also for mappings with many options where I'm interested only in one or two settings catgeories.
Not trivial though..
2. add an expand/collapse toggle above Device / Mapping Info.
Much easier, but the toggle itself would occupy a row

ronso0:
Btw
QCheckBox "chkEnabledDevice" and QLabel "labelLoadMapping" are both in row="1" column="0".
This will probably never cause issues (overlap) because the checkbox is left-aligned (default) and the label is right-aligned, only if the entire dialog is very narrow **and** all device info fields are empty **and** the big, bold device name is very short **and** the mapping description is not empty.
Just saying..

JoergAtGithub:
> Btw QCheckBox "chkEnabledDevice" and QLabel "labelLoadMapping" are both in row="1" column="0".

I also noticed this, but this is not new in this PR.

ronso0 (commented):
Thanks, I'll do a test with some HID and MIDI gear... next year ; )

src/controllers/dlgprefcontroller.cpp
```diff
@@ -796,6 +776,39 @@ void DlgPrefController::slotMappingSelected(int chosenIndex) {
...
+    // These tabs are only usable for MIDI controllers
+    if (m_pController->getDataRepresentationProtocol() == DataRepresentationProtocol::MIDI &&
+            !mappingFilePath.isEmpty()) {
+        m_ui.controllerTabs->setTabVisible(m_inputMappingsTabIndex, true);
+        m_ui.controllerTabs->setTabVisible(m_outputMappingsTabIndex, true);
+
+    } else {
+        m_ui.controllerTabs->setTabVisible(m_inputMappingsTabIndex, false);
+        m_ui.controllerTabs->setTabVisible(m_outputMappingsTabIndex, false);
+    }
```
ronso0, line 789: just an idea..
```suggestion
    bool showIOtabs = m_pController->getDataRepresentationProtocol() == DataRepresentationProtocol::MIDI &&
            !mappingFilePath.isEmpty();
    m_ui.controllerTabs->setTabVisible(m_inputMappingsTabIndex, showIOtabs);
    m_ui.controllerTabs->setTabVisible(m_outputMappingsTabIndex, showIOtabs);
```

JoergAtGithub:
> Thanks, I'll do a test with some HID and MIDI gear... next year ; )

@ronso0 Did you already found the time to test this?

ronso0:
Sorry, this got shifted down in my TODO list.
Will try to test tonight.

src/controllers/dlgprefcontroller.cpp
```diff
@@ -796,6 +776,35 @@ void DlgPrefController::slotMappingSelected(int chosenIndex) {
...
+    unsigned int visibleTabsCount = 0;
+    for (int tabIdx = 0; tabIdx < m_ui.controllerTabs->count(); ++tabIdx) {
+        if (m_ui.controllerTabs->isTabVisible(tabIdx)) {
+            ++visibleTabsCount;
+        }
+    }
+    return visibleTabsCount;
+}
+
+int DlgPrefController::getIndexOfFirstVisibleTabs() {
```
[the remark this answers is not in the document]
ronso0: fixed in https://github.com/JoergAtGithub/mixxx/pull/3

src/controllers/dlgprefcontroller.cpp
```diff
@@ -296,6 +305,12 @@ DlgPrefController::DlgPrefController(
             &QPushButton::clicked,
             this,
             &DlgPrefController::slotOutputControlSearch);
+
+    // Store the index of the input and output mappings tabs
+    m_inputMappingsTabIndex = m_ui.controllerTabs->indexOf(m_ui.inputMappingsTab);
+    m_outputMappingsTabIndex = m_ui.controllerTabs->indexOf(m_ui.outputMappingsTab);
+    m_settingsTabIndex = m_ui.controllerTabs->indexOf(m_ui.settingsTab);
+    m_screensTabIndex = m_ui.controllerTabs->indexOf(m_ui.screensTab);
```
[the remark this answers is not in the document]
ronso0: I don't see any side effects, just saying..

ronso0 (commented):
LGTM otherwise!

I opened https://github.com/JoergAtGithub/mixxx/pull/3 with some fixes:
* Screens tab was always visible when built with QML off
* finally managed to fix the random layout changes / v-stretch of the device / mapping info groupboxes
(this took hours, literally... sometimes Qt widget/layout props are a PITA)
* nits

Tested with MIDI Through and Traktor S4mk3 (HID).
All looking good now with the fixup PR.
image

src/controllers/dlgprefcontroller.cpp
```diff
@@ -1107,6 +1136,15 @@ void DlgPrefController::slotShowMapping(std::shared_ptr<LegacyControllerMapping>
...
     slotOutputControlSearch();
+
+    // Hide the entire QTabWidget if all tabs are removed
+    m_ui.controllerTabs->setVisible(getNumberOfVisibleTabs() > 0);
+
+    // Set the first visible tab as the current tab
+    int firstVisibleTabIndex = getIndexOfFirstVisibleTabs();
+    if (firstVisibleTabIndex >= 0) {
+        m_ui.controllerTabs->setCurrentIndex(firstVisibleTabIndex);
+    }
```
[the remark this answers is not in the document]
JoergAtGithub: I see your use case. But I think another use case is more important:
If a (new) user steps/scrolls through the list of mappings. Than it's likely that there was a mapping without controller settings beneth them. This means Midi In becomes the default - an expert setting, not intendend for average users. For them we introduced the controller settings, which should appear as default therefore.

ronso0:
Okay, let's merge it then.
Thanks!

PR #14007, opened 2024-12-09
rendergraph: add rendergraph library

m0dB:
Part of the rendergraph PR split into multiple PRs.

See https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/rendergraph.20PR.20split/near/487124048


src/rendergraph/common/rendergraph/attributeinit.h
```diff
@@ -0,0 +1,18 @@
+#pragma once
+
+#include "rendergraph/types.h"
+
+namespace rendergraph {
+struct AttributeInit;
+}
+
+// helper to create an AttributeSet using an initializer_list
```
Holzhaus, line 9:
```suggestion
/// Helper to create an AttributeSet using an initializer_list. 
```

src/rendergraph/common/rendergraph/material.h
```diff
@@ -0,0 +1,71 @@
...
+class rendergraph::Material : public rendergraph::BaseMaterial {
+  public:
+    Material(const UniformSet& uniformSet);
+    virtual ~Material();
+
+    // see QSGMaterial::compare.
+    // TODO decide if this should be virtual. QSGMaterial::compare is virtual
+    // to concrete Material can implement a compare function, but in rendergraph
+    // we can compare the uniforms cache and texture already here, which seems
+    // sufficient.
```
Holzhaus, line 26: Should this use triple slash?

src/rendergraph/common/rendergraph/assert.h
```diff
@@ -0,0 +1,3 @@
+// rendergraph is a module but we want to include util/assert.h from Mixxx,
+// without adding the entire src/ dir to the include paths.
+#include "../../../util/assert.h"
```
Holzhaus, line 3: Not a fan of the relative import from the parent dir. We don't do that anywhere else. Not sure if there is a better way.

src/rendergraph/opengl/CMakeLists.txt
```diff
@@ -0,0 +1,82 @@
...
+  ../common/rendergraph/node.h
+  ../common/rendergraph/opacitynode.h
+  ../common/rendergraph/texture.h
+  ../common/rendergraph/types.h
+  ../common/rendergraph/uniform.h
+  ../common/rendergraph/uniformscache.cpp
+  ../common/rendergraph/uniformscache.h
+  ../common/rendergraph/uniformset.cpp
+  ../common/rendergraph/uniformset.h
+  ../common/types.cpp
```
Holzhaus, line 31: This is repeated multiple times. Can't we define a RENDERGRAPH_COMMON_HEADERS variable in the parent dir and then reference this instead of copy/paste?

src/rendergraph/opengl/backend/basenode.cpp
```diff
@@ -0,0 +1,62 @@
...
+        auto pChild = m_pFirstChild;
+        removeChildNode(pChild);
+        delete pChild;
+    }
+}
+
+// This mimics QSGNode::appendChildNode.
+// Use NodeInterface<T>::appendChildNode(std::unique_ptr<BaseNode> pNode)
+// for a more clear transfer of ownership. pChild is considered owned by
+// this at this point.
```
Holzhaus, line 22:
```suggestion
/// This mimics QSGNode::appendChildNode.
/// Use NodeInterface<T>::appendChildNode(std::unique_ptr<BaseNode> pNode)
/// for a more clear transfer of ownership. pChild is considered owned by
/// this at this point.
```

And elsewhere.

src/rendergraph/opengl/rendergraph/openglnode.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include "backend/baseopenglnode.h"
+
+namespace rendergraph {
+using OpenGLNode = BaseOpenGLNode;
+} // namespace rendergraph
```
Holzhaus, line 7: There are many files with less than 10 lines. Are you going to extend these files in the subsequent PRs? Because this makes the code harder to follows and reeks of YAGNI.

src/rendergraph/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+find_package(QT NAMES Qt6 REQUIRED COMPONENTS Core)
+message(STATUS "Qt version ${QT_VERSION_MAJOR}.${QT_VERSION_MINOR}")
+
+if(QT_VERSION_MINOR GREATER_EQUAL 6)
+  set(USE_QSHADER_FOR_GL ON)
+endif()
+
+add_subdirectory(opengl)
+add_subdirectory(scenegraph)
+add_subdirectory(../../res/shaders/rendergraph shaders)
```
Holzhaus, line 10: Isn't using a parent directory kind of the opposite of what `add_subdirectory` is supposed to do?

Holzhaus (commented):
Thanks for the PR. I started reviewing (see my comments) but then figured I'm likely not the right person to review this because I wasn't involved in the prior discussion and don't want to reopen discussion that you potentially already agreed upon. So it doesn't really make sense for me to review this further.

My general impression is that this code is quite complex and I'm a bit worried that it's *more complex than it needs to be*. But again, if you think this is needed that is okay for me because I wasn't involved in prior discussion and also don't know what the follow up PRs look like (might also be that you create the needed foundation for them).

src/rendergraph/common/rendergraph/material.h
m0dB: I should probably not leave TODOs!

src/rendergraph/common/rendergraph/assert.h
m0dB: It’s a bit ugly, but I didn’t want to add the entire src/ dir to the include paths.

src/rendergraph/opengl/CMakeLists.txt
m0dB: will do

src/rendergraph/opengl/rendergraph/openglnode.h
m0dB: This is a consequence of the design of having two backend implementations, and a thin common API to interface with either. I don’t see away around this with the current design.

src/rendergraph/CMakeLists.txt
m0dB: The alternative is to move the rendergraph shaders to src/rendergraph/shaders. I am not against that.

m0dB:
Thanks for the review, Jan @Holzhaus. You are right that there has been a lot of prior discussion and reviews and the current state is the end result after many iterations. Still, your comments are welcome.

As for the complexity, I don’t think it is that complex. What makes it seem more complex than it is, is having two implementations of the same API, one using OpenGL and one using Qt scenegraph. But if you look past that: the scenegraph layer is extremely thin, and the OpenGL classes are pretty straightforward, especially if you are familiar with the Qt scenegraph design.

Anyway, any complexity here is compensated in the waveform renders code, which become much cleaner by having a very simple and standard way of doing things: You render a tree of nodes, typically GeometryNode, each with a Material of choice and a Geometry that you populate. Completely hiding all the OpenGL boilerplate, shaders, bindings, etc.

I think that when you review the waveformrenderer code you will appreciate this. And of course the fact that the same code will run within QML!

src/rendergraph/README
acolombier, line 1: Should this be a `README.md`? Arguably it doesn't have much formatting, but that might play nicer into some IDE or on Windows which uses the extension to guess the filename

src/rendergraph/common/rendergraph/assert.h
acolombier: I'm wondering if we could get CMake to copy `assert.h` into the rendergraph include? CMake noob so I have no idea if it could work

src/rendergraph/common/rendergraph/material.h
acolombier: Is this TODO solved somewhere in the following planned PRs? Otherwise, perhaps worth creating an issue and reference it there so we keep track and context of it?

src/rendergraph/common/rendergraph/material.h
```diff
@@ -0,0 +1,71 @@
...
+    int compare(const Material* pOther) const {
+        DEBUG_ASSERT(type() == pOther->type());
+        int cacheCompareResult = std::memcmp(m_uniformsCache.data(),
+                pOther->m_uniformsCache.data(),
+                m_uniformsCache.size());
+        if (cacheCompareResult != 0) {
+            return cacheCompareResult < 0 ? -1 : 1;
+        }
+        // TODO multiple textures
```
acolombier, line 35: Same here

src/rendergraph/common/rendergraph/material.h
```diff
@@ -0,0 +1,71 @@
...
+    Material(const UniformSet& uniformSet);
+    virtual ~Material();
+
+    // see QSGMaterial::compare.
+    // TODO decide if this should be virtual. QSGMaterial::compare is virtual
+    // to concrete Material can implement a compare function, but in rendergraph
+    // we can compare the uniforms cache and texture already here, which seems
+    // sufficient.
+    int compare(const Material* pOther) const {
+        DEBUG_ASSERT(type() == pOther->type());
```
acolombier, line 28: Do we want to early return here in case this is not the same type? I guess we might want to do that if `m_uniformsCache` and `pOther->m_uniformsCache` have different size which could lead to out of bound access, right?

src/rendergraph/common/rendergraph/material/rgbamaterial.cpp
```diff
@@ -0,0 +1,34 @@
...
+        : Material(uniforms()) {
+}
+
+// static
+const AttributeSet& RGBAMaterial::attributes() {
+    static AttributeSet set = makeAttributeSet<QVector2D, QVector4D>({"position", "color"});
+    return set;
+}
+
+/* static */ const UniformSet& RGBAMaterial::uniforms() {
```
acolombier, line 21: super nit -  just for consistency with the above
```suggestion
// static
const UniformSet& RGBAMaterial::uniforms() {
```

src/rendergraph/common/rendergraph/uniformset.cpp
```diff
@@ -0,0 +1,20 @@
+#include "rendergraph/uniformset.h"
+
+using namespace rendergraph;
+
+UniformSet::UniformSet(std::initializer_list<Uniform> list, const std::vector<QString>& names) {
+    int i = 0;
+    for (auto item : list) {
```
acolombier, line 7: Clazy here

src/rendergraph/common/types.cpp
```diff
@@ -0,0 +1,111 @@
...
+    case Type::Vector2D:
+        return sizeof(float) * 2;
+    case Type::Vector3D:
+        return sizeof(float) * 3;
+    case Type::Vector4D:
+        return sizeof(float) * 4;
+    case Type::Matrix4x4:
+        return sizeof(float) * 4 * 4;
+    }
+    return 0;
```
acolombier, line 36: Ditto assert?

src/rendergraph/common/types.cpp
```diff
@@ -0,0 +1,111 @@
...
+using namespace rendergraph;
+
+int rendergraph::sizeOf(PrimitiveType type) {
+    switch (type) {
+    case PrimitiveType::UInt:
+        return sizeof(uint32_t);
+    case PrimitiveType::Float:
+        return sizeof(float);
+    }
+    return 0;
```
acolombier, line 18: Should we add a `DEBUG_ASSERT` here to detect potentially unsupported new renderers?

src/rendergraph/opengl/backend/baseattributeset.cpp
```diff
@@ -0,0 +1,18 @@
...
+    DEBUG_ASSERT(list.size() == names.size());
+    int i = 0;
+    int offset = 0;
+    m_attributes.reserve(list.size());
+    for (auto item : list) {
+        m_attributes.push_back(BaseGeometry::Attribute{
+                offset, item.m_tupleSize, item.m_primitiveType, names[i++]});
+        offset += item.m_tupleSize * sizeOf(item.m_primitiveType);
+    }
+    m_sizeOfVertex = offset;
```
acolombier, line 17: Should we use `m_sizeOfVertex` directly? You could also remove `i` and use `m_attributes.count()` I believe

```suggestion
        const std::vector<QString>& names): m_sizeOfVertex(0)  {
    DEBUG_ASSERT(list.size() == names.size());
    int i = 0;
    m_attributes.reserve(list.size());
    for (auto item : list) {
        m_attributes.push_back(BaseGeometry::Attribute{
                offset, item.m_tupleSize, item.m_primitiveType, names[i++]});
        m_sizeOfVertex += item.m_tupleSize * sizeOf(item.m_primitiveType);
    }
```

src/rendergraph/opengl/backend/basegeometry.cpp
```diff
@@ -0,0 +1,21 @@
...
+} // namespace
+
+BaseGeometry::BaseGeometry(
+        const BaseAttributeSet& attributeSet, int vertexCount)
+        : m_pAttributes(attributeSet.attributes().data()),
+          m_attributeCount(attributeSet.attributes().size()),
+          m_sizeOfVertex(attributeSet.sizeOfVertex()),
+          m_drawingMode(defaultDrawingMode),
+          m_vertexCount(vertexCount),
+          m_vertexData(m_vertexCount * m_sizeOfVertex / sizeof(float)) {
```
acolombier, line 20: nit: might want to use `vertexCount` instead of `m_vertexCount` in case BaseGeometry's properties were somehow reorder in the feature which could lead to init-ing with a zero value.

src/rendergraph/opengl/backend/basegeometrynode.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+namespace {
+GLenum toGlDrawingMode(DrawingMode mode) {
+    switch (mode) {
+    case DrawingMode::Triangles:
+        return GL_TRIANGLES;
+    case DrawingMode::TriangleStrip:
+        return GL_TRIANGLE_STRIP;
+    default:
+        throw std::runtime_error("not implemented");
```
acolombier, line 21: What's our take with exception? AFAIK, there are pretty rare in the code base. IMHO, it wouldn't be a bad idea to start using them, but we should catch them. Otherwise, we could use the `DEBUG_ASSERT` pattern.
@Swiftb0y wdyt?

src/rendergraph/opengl/backend/basegeometrynode.cpp
```diff
@@ -0,0 +1,115 @@
...
+                shader.setUniformValue(location, cache.get<QVector4D>(i));
+                break;
+            case Type::Matrix4x4:
+                shader.setUniformValue(location, cache.get<QMatrix4x4>(i));
+                break;
+            }
+        }
+    }
+
+    // TODO this code assumes all vertices are floats
```
acolombier, line 78: Do we want to assert that?

src/rendergraph/opengl/backend/basegeometrynode.cpp
```diff
@@ -0,0 +1,115 @@
...
+        int location = material.attributeLocation(i);
+        shader.enableAttributeArray(location);
+        shader.setAttributeArray(location,
+                geometry.vertexDataAs<float>() + vertexOffset,
+                attribute.m_tupleSize,
+                geometry.sizeOfVertex());
+        vertexOffset += attribute.m_tupleSize;
+    }
+
+    // TODO multiple textures
```
acolombier, line 91: Ditto assertion?

src/rendergraph/opengl/backend/basenode.h
```diff
@@ -0,0 +1,70 @@
...
+    void setEngine(Engine* engine) {
+        m_pEngine = engine;
+    }
+    Engine* engine() const {
+        return m_pEngine;
+    }
+
+    // Prefer using NodeInterface<T>::appendChildNode(std::unique_ptr<BaseNode> pNode);
+    void appendChildNode(BaseNode* pChild);
+    // Prefer using std::unique_ptr<BaseNode> NodeInterface<T>::detachChildNode(BaseNode* pNode);
```
acolombier, line 41: Not sure to understand the doc-strings here - is there a reason for keeping `appendChildNode` and `removeChildNode` public?

src/rendergraph/opengl/engine.cpp
```diff
@@ -0,0 +1,98 @@
+#include "rendergraph/engine.h"
+
+#include <QDebug>
+#include <cassert>
```
acolombier, line 4: Shall we replace those with `DEBUG_ASSERT`/`VERIFY_OR_DEBUG_ASSERT`?

src/rendergraph/opengl/texture.cpp
```diff
@@ -0,0 +1,58 @@
...
+    if (image.format() == QImage::Format_RGBA8888_Premultiplied) {
+        return QImage(image.bits(), image.width(), image.height(), QImage::Format_RGBA8888).copy();
+    }
+    return QImage(
+            image.convertToFormat(QImage::Format_RGBA8888_Premultiplied)
+                    .bits(),
+            image.width(),
+            image.height(),
+            QImage::Format_RGBA8888).copy();
+    */
```
acolombier, line 46: Looking at the code and the doc, I think this should work! Do you want to just get in and we will fix forward if turns out not to?

src/rendergraph/scenegraph/backend/basematerialshader.cpp
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// override for QSGMaterialShader; this function is called by the Qt scene graph to prepare use of
+// sampled images in the shader, typically in the form of combined image samplers.
+void BaseMaterialShader::updateSampledImage(RenderState& state,
+        int binding,
+        QSGTexture** texture,
+        QSGMaterial* newMaterial,
+        QSGMaterial* oldMaterial) {
+    if (!newMaterial || !static_cast<Material*>(newMaterial)->texture(binding)) {
```
acolombier, line 31: Shall we add a `VERIFY_OR_DEBUG_ASSERT` to prevent the nullptr exception?

src/rendergraph/scenegraph/geometry.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+namespace {
+QSGGeometry::DrawingMode toSgDrawingMode(DrawingMode mode) {
+    switch (mode) {
+    case DrawingMode::Triangles:
+        return QSGGeometry::DrawTriangles;
+    case DrawingMode::TriangleStrip:
+        return QSGGeometry::DrawTriangleStrip;
+    default:
+        throw std::runtime_error("not implemented");
```
acolombier, line 17: Ditto C++ exception

src/rendergraph/scenegraph/geometry.cpp
```diff
@@ -0,0 +1,70 @@
...
+            vertexCount() * vertexStride) {
+        return;
+    }
+
+    float* to = static_cast<float*>(QSGGeometry::vertexData());
+    to += vertexOffset;
+    while (numTuples--) {
+        int k = tupleSize;
+        while (k--) {
+            *to++ = *from++;
```
acolombier, line 58: Ditto nullptr exception

acolombier (commented):
This is looking good, no major issue, only bunch of nits. Let me know if you would like me to approve anyway.

m0dB:
Thanks. I prefer to merge as is, and when all rendergraph code is merged do a follow up PR to address the things spotted here. When you merge this I will create the next PR. 

(For the next two PRs goes the same: unless you spot something particularly bad or broken, let's merge anyway and keep polishing / nitpicking for the end. After that, there are the PRs for the waveform render nodes, these can be done in parallel, so any changes don't require me to rebase the rest)

src/rendergraph/common/rendergraph/assert.h
m0dB: I really wouldn't go down that path. This is isolated and commented.

src/rendergraph/common/rendergraph/material.h
m0dB: I edited the comments.

src/rendergraph/opengl/texture.cpp
m0dB: No, I prefer to leave it like this and check once all PRs are in and you can rebase your qml code.

m0dB:
I improved some of the comments and removed the common source/headers list duplication. I prefer to do the rest of @acolombier 's (valid) comments to be done in a follow up PR, once all rendergraph PRs are in.

acolombier (approved):
LGTM!

Let's get the render-graph train started :bullettrain_side:

m0dB:
Thanks! I have created the next PR https://github.com/mixxxdj/mixxx/pull/14021

src/rendergraph/opengl/backend/basegeometrynode.cpp
Swiftb0y: I think the consensus was that using exceptions currently will result in brittle code (because you never know what throws and when) so the consensus was to avoid exceptions currently and instead use `std::expected`. In this case a dummy value or a `std::optional` could work.

Swiftb0y (commented):
review I started by never finished. Better late than never, feel free to address in a follow up whenever you find time.

Issue #13686, opened 2024-09-22
Sort by Key broken (happens intermittently)


m0dB:
### Bug Description

This has happened to me several times now. During a long session, at some point "sort by Key" stopped working correctly.

Steps: sort by BPM (click in column header), select some row, sort by Key (click in column header).

The sorting itself worked correctly, except for the selected row, which would not be in the  part of the table with the same Key.

E.g. I would select a row with key Fm, and after sorting the rows above and below the selected row would be with key A, and subsequently sorted as expected. But the selected row was not in the right position.

In addition to that, normally the rows are ordered by tempo as the secondary sort key (as in all tracks with the same key are grouped, and order by tempo). When this situation occurs the tracks with the same key are ordered seemingly random (or probably by some key, but not tempo).

This behaviour continued but occasionally the sorting would work correctly. After restarting the app the problem disappeared.

I am not sure if this happens at random, or if it is some specific action or circumstance that triggers this (I am even wondering if it is by design and I haven't figured out the rationale). I haven't seen this in the past though.

### Version

2.5

### OS

macOS

m0dB:
I am seeing this quite often. I will try to debug this.

m0dB:
So far I have only seen this with 2.5, not with 2.4, but that might be coincidence.

Swiftb0y:
You may want to look into these PRs: https://github.com/mixxxdj/mixxx/pulls?q=is%3Apr+is%3Aclosed+author%3Adanferns

m0dB:
At these PRs or just at the code, since there is no guarantee this was introduced by any of these PRs... But It will be difficult to figure this out as it only happens occasionally... I will ask in zulip for people to watch out for a specific sequence of events to trigger this bug.

Swiftb0y:
I rather meant just as a starting point where to poke at when you want to do a `git bisect`. Buf of course thats hard with a problem thats hard to reproduce...

fwcd:
I ran into this a couple of times too, see also the Zulip thread: https://mixxx.zulipchat.com/#narrow/channel/109122-general/topic/Key.20sorting.20bug.20on.202.2E6.2Fmain

Changing the sort order causes it to sort correctly, so it definitely seems like a bug:

https://github.com/user-attachments/assets/0ba6ed87-d632-4f1f-9f6c-8cd23bdb6dc7

fwcd:
[...]

PR #14008, opened 2024-12-10
determine the key string from either key id or key in the basetrackcache

m0dB:
Fixes https://github.com/mixxxdj/mixxx/issues/13686

Under certain circumstances, when sorting tracks, `BaseTrackCache::findSortInsertionPoint` would be called and in turn call `BaseTrackCache::compareColumnValues`. When comparing the Key column, it would compare the string value (KEY column) which could be empty, instead of taking the KEY_ID column into account. As a result, findSortInsertionPoint would return a wrong value, resulting in https://github.com/mixxxdj/mixxx/issues/13686 

This fix uses the code that was originally in `BaseTrackTableModel::roleValue`, now moved to KeyUtils, to determine the Key string value based on both columns.

Note that this is not the most efficient: we potentially go from numeric key code to string, only to go back to numeric key code in `BaseTrackCache::compareColumnValues`. But I found that avoiding that would have been much more invasive.


fwcd:
Seems to work for me, though I still can't reproduce the bug reliably using your steps on main/2.5 builds. I'm afraid that I don't have the time to dig into this in detail, so ideally someone more familiar with the key sorting algorithm should review this

src/library/basetrackcache.cpp
```diff
@@ -381,6 +381,8 @@ void BaseTrackCache::getTrackValueForColumn(TrackPointer pTrack,
         trackValue.setValue(pTrack->getRating());
     } else if (fieldIndex(ColumnCache::COLUMN_LIBRARYTABLE_KEY) == column) {
         trackValue.setValue(pTrack->getKeyText());
+        QVariant keyCodeValue(static_cast<int>(pTrack->getKey()));
+        trackValue = KeyUtils::keyFromColumns(std::move(trackValue), std::move(keyCodeValue));
```
ywwg, line 385: I do not think this use of std::move and rvalue reference in the called function do what you think it will do, and it adds noise to the code.  Almost certainly, given the function signatures, the compiler will do the right thing and avoid extra copies by itself.  The function should just take const& and it will work fine
```suggestion
        trackValue = KeyUtils::keyFromColumns(trackValue, keyCodeValue);
```

src/track/keyutils.cpp
```diff
@@ -763,3 +767,34 @@ int KeyUtils::keyToCircleOfFifthsOrder(mixxx::track::io::key::ChromaticKey key,
...
+    // to determine the Key string from either the LIBRARYTABLE_KEY or the
+    // LIBRARYTABLE_KEY_ID field.
+    //
+    // If we know the semantic key via the LIBRARYTABLE_KEY_ID
+    // column (as opposed to the string representation of the key
+    // currently stored in the DB) then lookup the key and render it
+    // using the user's selected notation.
+    if (keyCodeValue.isNull()) {
+        // Otherwise, just use the column value as is
+        return std::move(rawValue);
```
ywwg, line 783:
```suggestion
QVariant KeyUtils::keyFromColumns(const QVariant& rawValue, const QVariant& keyCodeValue) {
    // Helper function used by basetrackcache.cpp and basetracktablemodel.cpp
    // to determine the Key string from either the LIBRARYTABLE_KEY or the
    // LIBRARYTABLE_KEY_ID field.
    //
    // If we know the semantic key via the LIBRARYTABLE_KEY_ID
    // column (as opposed to the string representation of the key
    // currently stored in the DB) then lookup the key and render it
    // using the user's selected notation.
    if (keyCodeValue.isNull()) {
        // Otherwise, just use the column value as is
        return rawValue;
```

src/track/keyutils.h
```diff
@@ -131,6 +132,8 @@ class KeyUtils {
     static int keyToCircleOfFifthsOrder(mixxx::track::io::key::ChromaticKey key,
                                         KeyNotation notation);
 
+    static QVariant keyFromColumns(QVariant&& rawValue, QVariant&& keyCodeValue);
```
ywwg, line 135:
```suggestion
    static QVariant keyFromColumns(const QVariant& rawValue, const QVariant& keyCodeValue);
```

ywwg (changes requested):
https://devblogs.microsoft.com/oldnewthing/20231124-00/?p=109059

src/library/basetrackcache.cpp
m0dB: Yeah, I got carried a way a bit here 😄

m0dB:
Thanks for the suggestions. @ywwg . I applied them and did some additional cleanup. Please re-review.

Thanks for testing @fwcd . Yes, reproducing the bug, even with the steps provided, is not very clear, and neither is the code path that eventually leads to the code with the bug. The whole library code is really hard to follow and seems needlessly complex.

ywwg (commented):
is there a unit test we can write for this?

m0dB:
> is there a unit test we can write for this?

If you mean a test for the sorting of the library by key: I am not very familiar with what tests we have, but I get the impression we don't have any tests that populate a testing mixxxdb.sqlite. So I would say that is outside the scope of this bug fix.

If you mean a unit test for the KeyUtils::keyFromColumns function that I isolated from BaseTrackTableModel::roleValue, sure, that could be done easily.

daschuer:
I can confirm the bug. I have also found a couple of tracks in my library where the key text is an empty string (not NULL) but a valid key ID. I have other tracks where the key column contain not recognized strings but a "0" key ID. 

I have no idea how this happens, 

I have checked with the unrecognized track and mixxx 2.6-alpha and the it is imported with an empty string and key ID 0 which is reasonable and does not lead to the bug. 

We need to keep an eye of it id the root cause for the library inconstancy is still there. This reminds me by the we to this pending PR I need to finish. https://github.com/mixxxdj/mixxx/pull/11098

daschuer (changes requested):

src/track/keyutils.cpp
```diff
@@ -763,3 +767,34 @@ int KeyUtils::keyToCircleOfFifthsOrder(mixxx::track::io::key::ChromaticKey key,
...
+    // to determine the Key string from either the LIBRARYTABLE_KEY or the
+    // LIBRARYTABLE_KEY_ID field.
+    //
+    // If we know the semantic key via the LIBRARYTABLE_KEY_ID
+    // column (as opposed to the string representation of the key
+    // currently stored in the DB) then lookup the key and render it
+    // using the user's selected notation.
+    if (keyCodeValue.isNull()) {
+        // Otherwise, just use the column value as is
+        return rawValue;
```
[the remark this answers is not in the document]
m0dB: > Since this is only the exceptional path (right?)

Correct.

src/track/keyutils.cpp
```diff
@@ -763,3 +767,34 @@ int KeyUtils::keyToCircleOfFifthsOrder(mixxx::track::io::key::ChromaticKey key,
         return s_sortKeysCircleOfFifthsLancelot[static_cast<int>(key)];
     }
 }
+
+// static
+QVariant KeyUtils::keyFromColumns(const QVariant& rawValue, const QVariant& keyCodeValue) {
```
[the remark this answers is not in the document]
m0dB: renamed

m0dB:
done, please re-review and resolve.

daschuer (approved):
LGTM, Thank you.

daschuer:
@ywwg It looks like your change request is also included. Merge?

ywwg (approved):

PR #14011, opened 2024-12-12
(fix) DlgTrackInfo/~Multi: use configured notation for key display

ronso0:
Previously, Track Properties did show the key text stored in the track, not the corresponding key of the notation selected in Key preferences. Furthermore, when selecting such a key literal text of  a different notation in DlgTrackInfoMulti, that string was always converted to 'Traditional' notation (select 5A, get Cm).

This fixes Track Properties and uses the new helper function in Key preferences.

### DIFF
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -744,7 +744,7 @@ mixxx::UpdateResult DlgTrackInfo::updateKeyText() {
 }
 
 void DlgTrackInfo::displayKeyText() {
-    const QString keyText = m_trackRecord.getMetadata().getTrackInfo().getKeyText();
+    const QString keyText = KeyUtils::keyToString(m_trackRecord.getKeys().getGlobalKey());
     txtKey->setText(keyText);
 }
 
--- a/src/track/keyutils.cpp
+++ b/src/track/keyutils.cpp
@@ -5,6 +5,7 @@
 #include <QRegularExpression>
 #include <QtDebug>
 
+#include "preferences/keydetectionsettings.h"
 #include "util/compatibility/qmutex.h"
 
 using mixxx::track::io::key::ChromaticKey;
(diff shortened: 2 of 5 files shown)


src/library/dlgtrackinfo.cpp
```diff
@@ -744,7 +748,10 @@ mixxx::UpdateResult DlgTrackInfo::updateKeyText() {
 }
 
 void DlgTrackInfo::displayKeyText() {
-    const QString keyText = m_trackRecord.getMetadata().getTrackInfo().getKeyText();
+    const QString keyText =
+            KeyUtils::tryConvertKeyString(
+                    m_trackRecord.getMetadata().getTrackInfo().getKeyText(),
+                    m_notation);
```
daschuer, line 754: The issue is probably caused of hard to used code. We have here a hint:
https://github.com/mixxxdj/mixxx/blob/9817e5eab23ce907f8e49bfdb0050fef9549257d/src/track/track.cpp#L1453
but not here: 
https://github.com/mixxxdj/mixxx/blob/24a84b9d64607bb2b2cdcfc270bf9cb885074f3a/src/track/trackinfo.h#L35

The later seems to ba a container fr the key text stored in track metadata. Do we want to rename (again) it or at least add a comment? 

I think here we want exactly what found in the metadata, right?  
https://github.com/mixxxdj/mixxx/blob/9817e5eab23ce907f8e49bfdb0050fef9549257d/src/library/browse/browsethread.cpp#L239

daschuer, line 754: I think it is better to use the same string the track object will return for display purpose: 
```suggestion
    const QString keyText = KeyUtils::keyToString(m_record.getKeys().getGlobalKey());
```
this should work without `m_notation` which can be probably be removed, (not tested)

src/track/keyutils.cpp
```diff
@@ -538,6 +544,24 @@ KeyUtils::KeyNotation KeyUtils::keyNotationFromNumericValue(double value) {
     return static_cast<KeyNotation>(value_floored);
 }
 
+KeyUtils::KeyNotation KeyUtils::keyNotationFromString(const QString& notationName) {
```
daschuer, line 547: Please keep this nice refactoring.

daschuer (changes requested):
I can confirm the issue. The fix can be improved.

ronso0:
You're right, that code ishard to use : P
I overlooked `KeyUtils::setNotation()` which makes `keyToString()` use the selected notation.

Here are my observations:
Throwing Snow - Iridis
| | |
|---|---|
|Tracks  | 1A |
Deck |    1A
Track Properties dialog |     G#m
-> Re-import from file |1A
| | | 
database |  G#m
file metadat|     G#m

ronso0:
Yes, it does work without `m_notation`

ronso0:
I have no energy to look into the [inconsistency described above](https://github.com/mixxxdj/mixxx/pull/14011#issuecomment-2542653733), though we should fix that IMO.

(btw I don' consider this here a blocker, it's probably been like this for quite some time, 2.4 and earlier I think)

daschuer (approved):
LGTM, thank you.

ronso0:
Thanks @daschuer for your hints!

PR #14012, opened 2024-12-12
(fix) also exclude QEvent::Timer in MixxxApplication::notify()

ronso0:
With #13653 I noticed that QEvent::Timer is also affected.


daschuer (approved):
LGTM, thank you.

Issue #13959, opened 2024-12-01
UI: Missing Icons in Deere with 150% Scaling


uklotzde:
### 100%

![Screenshot From 2024-12-01 23-32-24](https://github.com/user-attachments/assets/2a3d881c-c386-42ad-83c9-62cd4ff0f488)

### 150%

![Screenshot From 2024-12-01 23-31-18](https://github.com/user-attachments/assets/1e8ad461-8b95-45df-afcb-4f4c1d4cdfcb)

### Version

main

### OS

Fedora 41

daschuer:
I can not exactly confirm the issue with Ubuntu Jammy, but there is at least an issue: 
![grafik](https://github.com/user-attachments/assets/2773f782-ea3f-4ff6-9bc9-238f8979fd25)

The 2.5 branch is also affected. 

2.4.2 works as expected but has blurry icons.

uklotzde:
Might be unrelated:

```
warning [Main]      $ connectMenuBar
warning [Main] Failed to load pixmap from path: "skin:/buttons/btn_"
```

uklotzde:
The artifacts in the upper left corners disappear after hovering over the almost empty buttons:

![Screenshot From 2024-12-02 02-33-51](https://github.com/user-attachments/assets/7d4f94c2-0b7f-4a76-a554-7f833d886085)

daschuer:
The 2.5 qt5 build is not affected.

uklotzde:
More glitches.

![Screenshot From 2024-12-02 19-22-30](https://github.com/user-attachments/assets/24ea1770-c06c-411e-92a2-aef5c109e46e)

ronso0:
The culprit seems to be that the `loop` icon has width and height set, while `loop_in` doesn't.
Both have the viewbox property set `viewBox="0 0 48 48"`
https://github.com/mixxxdj/mixxx/blob/main/res/skins/Deere/icon/ic_loop_in_48px.svg?short_path=2502654
https://github.com/mixxxdj/mixxx/blob/main/res/skins/Deere/icon/ic_loop_48px.svg?short_path=a969991

@uklotzde Can you try adding `width="48" height="48"` to the `<svg>` node of loop_in and check if that fixes it?

uklotzde:
Seems to work with explicit `height`/`width`:

![image](https://github.com/user-attachments/assets/a004b91d-dc9d-40b2-b21d-4d2fb8e99ad5)

[...]

PR #14013, opened 2024-12-12
insert width and height in svgs where missing

m0dB:
This adds width="W" and height="H" to the Deere svg files that were missing these attributes in the svg tag. The values are taken from the viewBox="X Y W H" attribute.

This aims to fix https://github.com/mixxxdj/mixxx/issues/13959

I am not familiar with Deere, so @ronso0 please confirm that this corrects the issue, or if there is anything else that needs to be changed in the svgs.


ronso0:
I tried to repro with Mixxx' internal Retina scaling, and it's kinda funny:
some icons get rendered smaller each time they're used, see the play icon here
image

ronso0:
LGTM Thanks!

PR #14014, opened 2024-12-12
(fix) Tango icons: use 'image' instead of 'background-image' to fix upscaling

ronso0:
The up/down arrows of the expand toggles should look sharp now with any scaling.
loop size sections, fx units, sampler rows
image


daschuer (approved):
LGTM, Thank you.

PR #14016, opened 2024-12-13
(fix) Traktor Kontrol S4 Mk3: correct wheel timestamp wrap-around

ronso0:
just a copy/paste error I guess


acolombier (approved):
Indeed, looks like a copy pasta problem.
Thank you!

PR #14018, opened 2024-12-14
Devcontainer support

### LINKED ISSUE #13914: Provide devcontainer config
### Feature Description

I want to save time setting up the development environment of mixxx and also keep my system clean not installing compilers and libraries to my PC

The devcontainer specification allows to install dependencies etc. in a Docker Container and more and more IDEs are supporting to develop the code in such a container while preserving all IDE features for debugging / code analysis, language server etc.

 See this example I built for Rust: https://github.com/spotlesscoder/rust-template
Just clone in VSCode, install the devcontainer extension in VS Code, run the "Reopen in Container" Action and everything installs automatically and you're ready to go
Once done trying out you can delete the docker container and the cloned git folder and your system is as clean as before

I want to provide something similar for the C++ stack that mixxx needs and make a PR for .devcontainer/ folder with the necessary config files in the repo root. 
Would you be accepting such a PR?

acolombier:
<details>
 <summary>Some context from the author</summary>
 
As I have started trying new distro to replace my unsupported PopOS22.04, I have stopped on Fedora 41 SilverBlue. While I love the concept of immutable system, and everything is a container, it does bring a chjalenge of how to develop Mixxx.

This is my attempt to provide a compressive solution. Note that I have been using a patched version of devpod to get this to work locally, but I haven't yet shared my patch for it.

</details>

This fixes  #13914

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,46 @@
+FROM --platform=$BUILDPLATFORM ubuntu:24.04
+ARG TARGETARCH
+
+RUN DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get update && \
+    apt-get install -y \
+    alsa-utils \
+    # Includes basic building tools
+    build-essential \
+    ninja-build \
+    # Used by most OpenSource LSPs extensions
+    clangd \
+    # Debugger
+    gdb \
+    # VCS
+    git \
+    # Audio integration
+    pipewire \
+    pipewire-alsa \
+    pulseaudio-utils \
+    # Required for pre-commit
+    python3-pip \
+    # Default Qt QPA
+    qt6-wayland \
+    # Allow privileges escalation in container (e.g to install custom deps)
+    sudo \
+    # Allow udev rules debug and test
+    udev \
+    # Convenience for git interactive commands
+    vim && \
+    pip3 install pre-commit --break-system-packages
+
+COPY tools/debian_buildenv.sh /tmp/
+
+RUN /tmp/debian_buildenv.sh setup
+
+RUN userdel -f -r ubuntu && \
+    mkdir /home/developer && echo "developer ALL=(root) NOPASSWD:ALL" > /etc/sudoers.d/developer && \
+    mkdir -p \
+    /home/developer/.cache/pre-commit \
+    /home/developer/.devcontainer \
+    /home/developer/.mixxx \
+    /tmp/ccache && \
+    touch /home/developer/.devcontainer/.bash_history && \
+    chown -R 1000:1000 /home/developer/ /tmp/ccache && \
+    echo ".cache\n.vscode\n.node_modules" >> /home/developer/.gitignore && \
+    echo ". /etc/bash_completion\nexport PROMPT_COMMAND='history -a'\nexport HISTFILE='/home/developer/.devcontainer/.bash_history'" >> /home/developer/.bashrc
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,65 @@
+{
+  "name": "mixxx",
+  "build": {
+    "cacheFrom": "ghcr.io/mixxxdj/mixxx/devcontainer:latest",
+    "dockerfile": "Dockerfile",
+    "context": ".."
+  },
+  "runArgs": [
+    "--network=host" // This is needed for native X.org support
+  ],
+  "features": {
+    "ghcr.io/devcontainers/features/common-utils:2": {
+      "installZsh": "false",
+      "username": "developer",
+      "userUid": "1000",
+      "userGid": "1000",
+      "upgradePackages": "false"
+    }
+  },
+  "initializeCommand": "touch ${localEnv:HOME}${localEnv:USERPROFILE}/.gitignore",
+  "customizations": {
+    "vscode": {
+      "extensions": [
+        "llvm-vs-code-extensions.vscode-clangd",
+        "twxs.cmake",
+        "ms-vscode.cmake-tools",
+        "streetsidesoftware.code-spell-checker",
+        "vadimcn.vscode-lldb",
+        "GitHub.vscode-pull-request-github",
+        "eamodio.gitlens"
+      ]
+    }
+  },
+  "remoteEnv": {
+    "SSH_AUTH_SOCK": "${localEnv:SSH_AUTH_SOCK}",
+    "PATH": "/usr/lib/qt6/bin/:${containerEnv:PATH}",
+    "DISPLAY": "${localEnv:DISPLAY::0}",
+    "XDG_SESSION_DESKTOP": "${localEnv:XDG_SESSION_DESKTOP:gnome}",
+    "XDG_SESSION_TYPE": "${localEnv:XDG_SESSION_TYPE:wayland}",
+    "XAUTHORITY": "${localEnv:XAUTHORITY}",
+    "WAYLAND_DISPLAY": "${localEnv:WAYLAND_DISPLAY:wayland-0}",
+    "XDG_SESSION_CLASS": "${localEnv:XDG_SESSION_CLASS:user}",
+    "XDG_RUNTIME_DIR": "${localEnv:XDG_RUNTIME_DIR:/run/user/1000}",
+    "QT_QPA_PLATFORM": "${localEnv:QT_QPA_PLATFORM:wayland}", // Default to Wayland session, can also use 'xcb' to use X.org instead
+    "CLAZY_CHECKS": "level2,container-inside-loop,heap-allocated-small-trivial-type,isempty-vs-count,qhash-with-char-pointer-key,qproperty-type-mismatch,tr-non-literal,no-rule-of-two-soft,no-non-pod-global-static,no-qproperty-without-notify,no-qstring-allocations,no-function-args-by-value,no-copyable-polymorphic,no-ctor-missing-parent-argument,no-missing-qobject-macro,no-rule-of-three,no-returning-void-expression",
+    "CLAZY_IGNORE_DIRS": "lib/.*"
+  },
+  "remoteUser": "developer",
+  "postCreateCommand": "git config --global core.excludesFile '/home/developer/.gitignore' && git config --global --add safe.directory ${containerWorkspaceFolder} && pre-commit install --overwrite --install-hooks",
+  "mounts": [
+    "type=bind,source=${localEnv:HOME},target=${localEnv:HOME},rslave",
+    "type=bind,source=${localEnv:XDG_RUNTIME_DIR},target=${localEnv:XDG_RUNTIME_DIR}",
+    "type=bind,source=/var/run/dbus/system_bus_socket,target=/var/run/dbus/system_bus_socket",
+    "type=bind,source=/dev,target=/dev,rslave",
+    "type=bind,source=/media,target=/media,rslave",
+    "type=bind,source=/mnt,target=/mnt,rslave",
+    "type=bind,source=/run/media,target=/run/media,rslave",
+    // Dev volumes used for persistence between containers
+    "type=volume,source=mixxx-bashistory,target=/home/developer/.devcontainer",
+    "type=volume,source=mixxx-precommit,target=/home/developer/.cache/pre-commit",
+    "type=volume,source=mixxx-config,target=/home/developer/.mixxx",
+    "type=volume,source=mixxx-ccache,target=/tmp/ccache",
+    "type=volume,source=mixxx-build,target=/workspaces/mixxx/build"
+  ]
+}
--- a/.github/workflows/devcontainer.yml
+++ b/.github/workflows/devcontainer.yml
@@ -1,0 +1,40 @@
+name: Docker
+
+on:
+  push:
+    branches: ["main"]
+
+  pull_request:
+    branches: [main]
+    types:
+      - opened
+      - synchronize
+      - labeled
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  devcontainer:
+    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && github.repository == 'mixxxdj/mixxx' || contains(github.event.pull_request.labels.*.name, 'dev container') }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v6
+        with:
+          file: .devcontainer/Dockerfile
+          platforms: linux/arm64,linux/amd64
+          pull: true
+          push: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' && github.repository == 'mixxxdj/mixxx' }}
+          tags: ghcr.io/mixxxdj/mixxx/devcontainer:latest
--- a/README.md
+++ b/README.md
@@ -60,6 +60,33 @@ run. Alternatively, can generate a package using `cpack`.
 
 Detailed build instructions for each target OS can be found [on the wiki](https://github.com/mixxxdj/mixxx/wiki#compile-mixxx-from-source-code)
 
+### Using Dev Container
+
+> [!NOTE]
+> Dev container has recently been introduced and it is likely incomplete! Currently, it's been tested with `devpod` on Zed and Codium exclusively
+
+We provide a Dev Container definition for Mixxx, based on Ubuntu 24.04 to ensure a close similarity with the CI.
+
+> [!TIP]
+> New to [Development Containers](https://containers.dev/)? If you use VS Code, the IDE should offer you to use Dev Container when you open the Mixxx working copy folder. Otherwise,
+you may consider using [devpod](https://devpod.sh/docs/developing-in-workspaces/create-a-workspace#create-a-workspace)
+
+You can build and run Mixxx using the following command:
+
+```bash
+cd build
+cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON # Needed for clangd
+cmake --build . -j $(nproc)
+./mixxx
+```
+
+#### Using your device in Dev Container
+
+Similar to [Fedora Toolbox's](https://github.com/containers/toolbox/blob/4f4c3c9d19d1027537f69920de46b9cf09c799b9/src/cmd/create.go#L460), the default configuration provide a comprehensive definition of mounts which will allow you to use Mixxx seamlessly in the container. (UI, audio, ...)
+If you are not comfortable with this for security concern, you may want to remove some or all of the binding before starting the container.
+
+Devices should be supported by default, thanks to the provided binds. Note that if you are encountering permission issue, check that you are not using SELinux, as this is currently not well supported in Dev Container
+
 ## Documentation
 
 For help using Mixxx, there are a variety of options:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -6,6 +6,12 @@ build:
           - build/**
           - cmake/**
 
+dev container:
+  - changed-files:
+      - any-glob-to-any-file:
+          - .devcontainer/**
+          - tools/debian_buildenv.sh
+
 cmake:
   - changed-files:
       - any-glob-to-any-file:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -38,6 +38,7 @@ repos:
         exclude: ^.*(\.cbproj|\.groupproj|\.props|\.sln|\.vcxproj|\.vcxproj.filters|UTF-8-BOM.txt)$
       - id: check-case-conflict
       - id: check-json
+        exclude: ^(.devcontainer/devcontainer.json)$ # Contains JSON comment, which isn't supported by the hook
       - id: check-merge-conflict
       - id: check-xml
       - id: check-yaml


acolombier:
It's not perfect, but I'd like to get this initial version merged and iterate on it

tools/debian_buildenv.sh
```diff
@@ -59,6 +59,7 @@ case "$1" in
             g++ \
             lcov \
             libbenchmark-dev \
+            libboost-dev \
```
daschuer, line 62: this is not required

tools/debian_buildenv.sh
```diff
@@ -104,6 +105,7 @@ case "$1" in
             qt6-declarative-private-dev \
             qt6-base-private-dev \
             qt6-qpa-plugins \
+            qt6-wayland \
```
daschuer, line 108: This pulls Wayland in. We don't want this. Mixxx does not depend on wayland.

daschuer (commented):
In general there is nothing particular against this, except the maintenance task. So let's consider who will used this. 

If a user on an older machine wants to develop a new Mixxx version, but don't want to used it later? That sounds as an exception. 

A way more streamlined approach is to use appImage/snap/flatpak this has an equal maintenance effort, but is also usable for others and I consider this a better benefit for our users. 

Do I miss something? So I vote for not merging this.

tools/debian_buildenv.sh
acolombier: This is required if you are running with `QT_QPA_PLATFORM=wayland`. Since our minimal system requirement is Wayland by default (Ubuntu 24.04), I think it would make sense to have it, but happy to move that in the Dockerfile otherwise

acolombier:
Thanks for your review @daschuer , I think you are missing the point of what [DevContainers](https://containers.dev/) are about. DevContainer (!= container) are stricly used in develpment to ensure:
- System immutability (a must if you are developing on an immutable system or a system without superuser access)
- Reproducibility - it allows to always have a prisitine copy of the dev environment, without any external factors, which can be very useful to align with our CI env
- Cloud/Ephemeral - it is the foundation of CDE (Cloud Dev Env) such as Codespaces or DevPod in the OpenSource
image

DevContainer are completely unrelated to installable user artifact such as Snap, AppImage or Flatpak and are not meant in any ways to replace them. 

Regarding maintainability, this is similar to our [NixOS dev setup](https://github.com/mixxxdj/mixxx/blob/main/shell.nix) - it may fall unmaintained, but we are expecting developer using it to take it of it themselves

Finally, regarding your vote against, I find it puzzling since you voted for it in the [related issue](https://github.com/mixxxdj/mixxx/issues/13914) when I asked if I should bring this in 2 months ago. With the amount of time I spent on this, I'm going to assume your statement was made in confusion.
image

jarey:
I personally consider this really useful, speciallly for people like me, that are considering to co tribute to some issues. It would minimise friction a lot and will make it way easier to new contributors to just start checking the code and do things on it without necessarily setting up any local dev environment.

Great job on this, by the way @acolombier .

tools/debian_buildenv.sh
acolombier: Moved the Dockerfile

acolombier:
Thanks for your feedback @jarey, greatly appreciated! 

Since no other active contributors use devconatiners to my knowledge, would you be interested in giving the setup a go? No need to do too much in there, just running the devcontainer and try [the suggested command to get Mixxx running](https://github.com/mixxxdj/mixxx/blob/b55319f5b1663b7fecbf366c517fb7a5e807efbb/README.md#using-dev-container). I mainly expect issues with X.org/Wayland connectivity and it would be great if I could at least document them

JoergAtGithub:
> Finally, regarding your vote against, I find it puzzling since you voted for it in the [related issue](https://github.com/mixxxdj/mixxx/issues/13914) when I asked if I should bring this in 2 months ago. With the amount of time I spent on this, I'm going to assume your statement was made in confusion.

We need a certain level of commitment when a core-dev makes such a vote. If no unexpected aspects arise during implementation, the contributor must assume that this work is wanted by the person who made the vote, and that this person will not oppose a merging, if the implementation is clean.

daschuer:
Sorry for the confusion. I should have made more clear that I am not blocking this PR. I have just reconsidered the pro and cons with a different result. I was under the expression that this no longer a need for it after @acolombier has moved to Fedora.

acolombier:
I have moved to Fedore Silverblue, which is immutable. I can build and run Mixxx from [toolbox](https://docs.fedoraproject.org/en-US/fedora-silverblue/toolbox/), but it is not possible to run an IDE insided these since apps are all flatpak based. This means devcontainer is the only viable approach 

That being said, I think allowing anybody to work on the project without having to install random dependencies is very valuable, so beside my case and immutable OSes, I think we should still allow such a setup to allow easy plug-n-dev

acolombier (commented):
I have squashed the fixups and removed SELinux support as it is quite unstable. With the latest version, it possible to get a true plug-n-dev or plug-n-play experience with your own controller or audio card without tinkering. 

It'd be great if we could look at merging this as it is and fix forward if we find better optimisation.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    git \
+    libadwaitaqt1 \
+    pipewire \
+    pipewire-alsa \
+    pulseaudio-utils \
+    python3-pip \
+    qt6-wayland \
+    sudo \
+    udev \
+    vim && \
```
JoergAtGithub, line 19: Are all these packages developer tools, or do some belong to debian_buildenv.sh instead?

acolombier: Good question - most of these packages are expected to already be installed on the machine (git, udev for example), some others provide an opinionated out of the box experience (vim, clangd, gdb).

I think it makes sense to keep it separated but open to move some of these packages in the script.

acolombier:
Friendly ping @jarey :)
Have you had a chance to give this a spin?

daschuer:
For my understanding the devcontainer.json file is missing: https://containers.dev/implementors/json_reference/

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    clangd \
+    gdb \
+    git \
+    libadwaitaqt1 \
+    pipewire \
+    pipewire-alsa \
+    pulseaudio-utils \
+    python3-pip \
+    qt6-wayland \
+    sudo \
```
daschuer, line 18: Do we need sudo? For my understanding you are already superuser in a devcontainer. 
I am afraid there are some more command line tools missing. Isn't there not a different base container of a devcontainer feature that comes with all this?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    adwaita-qt \
+    alsa-utils \
+    build-essential \
+    clangd \
+    gdb \
+    git \
+    libadwaitaqt1 \
+    pipewire \
+    pipewire-alsa \
+    pulseaudio-utils \
```
daschuer, line 15: Do we need all this? How is it supposed to work?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    python3-pip \
+    qt6-wayland \
+    sudo \
+    udev \
+    vim && \
+    pip3 install pre-commit --break-system-packages
+
+COPY tools/debian_buildenv.sh /tmp/
```
daschuer, line 23: Why we need a copy?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN /tmp/debian_buildenv.sh setup
+
+RUN userdel -f -r ubuntu && \
```
daschuer, line 27: What is the reason to for this change of user name? Is there an issue with the defaults?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    mkdir /home/developer && echo "developer ALL=(root) NOPASSWD:ALL" > /etc/sudoers.d/developer && \
```
daschuer, line 28: Why && all commands?

README.md
```diff
@@ -58,6 +58,29 @@ To build Mixxx, run
 There should now be a `mixxx` executable in the current directory that you can
 run. Alternatively, can generate a package using `cpack`.
 
+### Using Dev Container
+
+> [!NOTE]
+> Dev container has recently been introduced and it is likely incomplete! Currently, it's been tested with `devpod` on Zed and Codium exclusively
```
daschuer, line 64: This will be outdated soon. Please change it that it is valid for long.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,62 @@
...
+    "SSH_AUTH_SOCK": "${localEnv:SSH_AUTH_SOCK}",
+    "PATH": "/usr/lib/qt6/bin/:${containerEnv:PATH}",
+    "DISPLAY": "${localEnv:DISPLAY::0}",
+    "XDG_SESSION_DESKTOP": "${localEnv:XDG_SESSION_DESKTOP:gnome}",
+    "XDG_SESSION_TYPE": "${localEnv:XDG_SESSION_TYPE:wayland}",
+    "XAUTHORITY": "${localEnv:XAUTHORITY}",
+    "WAYLAND_DISPLAY": "${localEnv:WAYLAND_DISPLAY:wayland-0}",
+    "XDG_SESSION_CLASS": "${localEnv:XDG_SESSION_CLASS:user}",
+    "XDG_RUNTIME_DIR": "${localEnv:XDG_RUNTIME_DIR:/run/user/1000}",
+    "QT_QPA_PLATFORM": "${localEnv:QT_QPA_PLATFORM:wayland}" // Default to Wayland session, can also use 'xcb' to use X.org instead
```
daschuer, line 43: Is this required at all?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,62 @@
...
+        "streetsidesoftware.code-spell-checker",
+        "vadimcn.vscode-lldb",
+        "GitHub.vscode-pull-request-github",
+        "eamodio.gitlens"
+      ]
+    }
+  },
+  "remoteEnv": {
+    "SSH_AUTH_SOCK": "${localEnv:SSH_AUTH_SOCK}",
+    "PATH": "/usr/lib/qt6/bin/:${containerEnv:PATH}",
```
daschuer, line 35: Is this required?

.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,40 @@
...
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push Docker image
```
daschuer, line 33: How does this work?

README.md
```diff
@@ -58,6 +58,29 @@ To build Mixxx, run
...
+
+We provide a DevContainer definition for Mixxx. The container base uses Ubuntu 24.04 to ensure a close similarity with the CI.
+Similar to Fedora Toolbox's, the default configuration provide a comprehensive definition of mounts which will allow use to use Mixxx seamlessly in the container. (UI, audio, ...)
+If you are not comfortable with this, you may want to remove some or all of the binding before starting the container.
+
+You can build and run Mixxx using the following command:
+
+```bash
+mkdir build
```
daschuer, line 73: Following this, we build locally, right?

daschuer (changes requested):
I have added some comments an questions.

README.md
```diff
@@ -58,6 +58,29 @@ To build Mixxx, run
 There should now be a `mixxx` executable in the current directory that you can
 run. Alternatively, can generate a package using `cpack`.
 
+### Using Dev Container
+
+> [!NOTE]
+> Dev container has recently been introduced and it is likely incomplete! Currently, it's been tested with `devpod` on Zed and Codium exclusively
+
+We provide a DevContainer definition for Mixxx. The container base uses Ubuntu 24.04 to ensure a close similarity with the CI.
+Similar to Fedora Toolbox's, the default configuration provide a comprehensive definition of mounts which will allow use to use Mixxx seamlessly in the container. (UI, audio, ...)
```
daschuer, line 67: Why reference  "Fedora Toolbox's"

acolombier: Because this is the other big initiative around dev containerisation. It's less mature, supported and popular (yet), but I know we discuss about it as well so I thought it was worth to make the parallel for user that my have heard only of Fedora Toolbox (e.g Silverblue user) since it literally equivalent and work as a drop-in replacement

acolombier (commented):
Thanks for looking into this @daschuer !

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    build-essential \
+    clangd \
+    gdb \
+    git \
+    libadwaitaqt1 \
+    pipewire \
+    pipewire-alsa \
+    pulseaudio-utils \
+    python3-pip \
+    qt6-wayland \
```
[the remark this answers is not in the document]
acolombier: It is needed - without this, Mixxx cannot start
```
qt.qpa.plugin: Could not find the Qt platform plugin "wayland" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: minimalegl, offscreen, eglfs, xcb, vkkhrdisplay, minimal, linuxfb, vnc.
```

daschuer:
Thank you for changes. Can you squash the fixup. 
The readme is missing a step by step guide how to use the container. I think that is required.

acolombier:
Thanks for the follow up - I have added a small tip here, let me known if something more hands on should be provided there.

README.md
```diff
@@ -58,6 +58,32 @@ To build Mixxx, run
...
+If you are not comfortable with this, you may want to remove some or all of the binding before starting the container.
+
+> [!TIP]
+> New to devcontainer? If you use VSCode, the IDE should offer you to use devcontainer when you open the Mixxx project. If you are not using VSCode,
```
daschuer, line 71: Spelled "VS Code" 
```suggestion
> If you use VS Code, the IDE should offer you to use Dev Container when you open the mixxx working copy folder. Otherwise, 
```

README.md
daschuer:
```suggestion
The default configuration provide a comprehensive definition of mounts which will allow you to use Mixxx seamlessly in the container. (UI, audio, ...)
```

README.md
```diff
@@ -58,6 +58,32 @@ To build Mixxx, run
...
+You can build and run Mixxx using the following command:
+
+```bash
+cd build
+cmake .. -DCMAKE_EXPORT_COMPILE_COMMANDS=ON # Needed for clangd
+cmake --build . -j $(nproc)
+./mixxx
+```
+
+#### Using your device in DevContainer
```
daschuer, line 83:
```suggestion
#### Using your device in Dev Container
```

README.md
```diff
@@ -58,6 +58,32 @@ To build Mixxx, run
...
+
+Devices be supported by default, thanks to the provided binds. Note that if you are encountering permission issue, check that you are not using SELinux, as this is currently not well supported in DevContainer
```
daschuer, line 85:
```suggestion
Devices be supported by default, thanks to the provided binds. Note that if you are encountering permission issue, check that you are not using SELinux, as this is currently not well supported in Dev Container
```

daschuer: Or remove the paragraph and move the note after "the container. (UI, audio, ...)" above.

daschuer (changes requested):
I have suggested some text tweaks

README.md
daschuer: OK, in that case I would add an additional sentences that explains it the rationals and implications ... or link external resources.

README.md
```diff
@@ -58,6 +58,32 @@ To build Mixxx, run
...
 run. Alternatively, can generate a package using `cpack`.
 
+### Using Dev Container
+
+> [!NOTE]
+> Dev container has recently been introduced and it is likely incomplete! Currently, it's been tested with `devpod` on Zed and Codium exclusively
+
+We provide a DevContainer definition for Mixxx. The container base uses Ubuntu 24.04 to ensure a close similarity with the CI.
+Similar to Fedora Toolbox's, the default configuration provide a comprehensive definition of mounts which will allow use to use Mixxx seamlessly in the container. (UI, audio, ...)
+If you are not comfortable with this, you may want to remove some or all of the binding before starting the container.
```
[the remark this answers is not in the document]
daschuer: Ok, than put it in the additional sentences about security concerns.

README.md
```diff
@@ -58,6 +58,32 @@ To build Mixxx, run
...
+
+> [!NOTE]
+> Dev container has recently been introduced and it is likely incomplete! Currently, it's been tested with `devpod` on Zed and Codium exclusively
+
+We provide a Dev Container definition for Mixxx, based on Ubuntu 24.04 to ensure a close similarity with the CI.
+Similar to [Fedora Toolbox's](https://github.com/containers/toolbox/blob/4f4c3c9d19d1027537f69920de46b9cf09c799b9/src/cmd/create.go#L460), the default configuration provide a comprehensive definition of mounts which will allow you to use Mixxx seamlessly in the container. (UI, audio, ...)
+If you are not comfortable with this, you may want to remove some or all of the binding before starting the container.
+
+> [!TIP]
+> New to [dev container](https://containers.dev/)? If you use VS Code, the IDE should offer you to use Dev Container when you open the Mixxx working copy folder. Otherwise,
```
daschuer, line 71: OK, but use the right spelling: 
```suggestion
> New to [Development Containers](https://containers.dev/)? If you use VS Code, the IDE should offer you to use Dev Container when you open the Mixxx working copy folder. Otherwise,
```

README.md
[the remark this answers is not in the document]
daschuer: Maybe a paragraph unter "Using your device in Dev Container" ? 

I have been original stumbled over it, because we start with concerns. That apply in the same way to the alternative of native install anyway. Instead it would be better to start with the benefits compared to that. 

The point which is missing is the additional disk space it needs compare to a native build environment. Can you give an estimate?

README.md
daschuer: Did you gain more experience with this meanwhile? 
This note is very prominent compared to the rest of the Readme. Can we remove it?

daschuer:
How about moving the whole Dev Container paragraph out of the main repository readme to the wiki?

daschuer (changes requested):
After merging, every VS-Code user is urged to use UbuntuDev containers instead of native builds which is not recommended. I think it is required to make it not the defaul for VS-Code.

I consider this a blocker for merging. Can we move the .devcontainer folder to the tools subfolder for VS-Code?

acolombier:
> How about moving the whole Dev Container paragraph out of the main repository readme to the wiki?

What is the point of this? I can only see drawbacks:
1. Non core member won't be able to contribute (comment, update accordingly)
2. Information will only be harder to reach

> every VS-Code user is urged to use UbuntuDev containers instead of native builds

Sorry, but that's not true. You are only "suggested" to open in Dev Container if:
1. You have the dev container extension installed (meaning you use dev container)
2. You have't disabled that tip (clicked the `Don't Show Again` button, or set the setting key accordingly)
 
(I assuming the urge you are referring to is this popup)
image

> which is not recommended

What is the rational for that statement? This sounds like an opinion rather than objective fact? Could you please elaborate?

> Can we move the .devcontainer folder to the tools subfolder for VS-Code?

No you cannot, please read [the spec](https://containers.dev/implementors/templates-distribution/#packaging). This is that same as moving the `.github` or the `README.md` out of the project root, this will simply break the feature.

daschuer:
>> How about moving the whole Dev Container paragraph out of the main repository readme to the wiki?

> What is the point of this?

This gives the chance to give more explanations, without clutter the main Readme.md. But I don't mind which way you prefer.

daschuer:
>> every VS-Code user is urged to use UbuntuDev containers instead of native builds

> Sorry, but that's not true. You are only "suggested" to open in Dev Container if:
>  You have the dev container extension installed (meaning you use dev container)
> You have't disabled that tip (clicked the Don't Show Again button, or set the setting key accordingly)

I cannot confirm. In my case VS-Code installed automatically all requirements when you except all proposals. That is what new contributors do. 
This must not happen.

daschuer:
>> which is not recommended

> What is the rational for that statement? This sounds like an opinion rather than objective fact? Could you please elaborate?

Ins't it obvious that windows users want a windows build? Please describe why you want that windows and macOS users should use Mixxx in a Ubuntu docker cintainer?

daschuer:
>> Can we move the .devcontainer folder to the tools subfolder for VS-Code?

> No you cannot, please read the spec. This is that same as moving the .github or the README.md out of the project root, this will simply

Sorry that I am not expert in dev containers. I think it is up to you to propose a solution. 

For me it works flawlessly to open a subfolder with a `. devcontainer` folder. What issue do you expect?

acolombier:
> This gives the chance to give more explanations, without clutter the main Readme.md. But I don't mind which way you prefer.

Fair enough. As a middle ground, I could trim the content from the `README.md` and move additional details (e.g bindings etc) inside a `docs/devcontainer.md` file. Wdyt? 

> Ins't it obvious that windows users want a windows build? Please describe why you want that windows and macOS users should use Mixxx in a Ubuntu docker cintainer?

My experience with developing Mixxx on Windows and Mac (disclaimer - in VM, but I think my point stands) is that working with VCPKG is quite complicated, compare to the embedded build environment. Of course, depending of what you want to do, building with VPCKG may be required, but since we ship build artifacts as part of[ the pipeline ](https://github.com/mixxxdj/mixxx/actions/runs/13687241124), users needing to build Mixxx locally to test new feature has gone down, meaning new contributor could benefit from having a fully-fledge dev environment not specific to their OS (also, note that having that `build` volume allow user to actually do both) and would help with for example, pre-commits, which is still something we have with many new contributors, or less experimented developers.) 

> For me it works flawlessly to open a subfolder with a `.devcontainer` folder. What issue do you expect?

The problem comes with the workspace - say you move `.devcontainer` inside `tools`, Dev Container would bind the workspace as `tools` instead of the root of the repo, making the devcontainer spec unusable. There is some tinkering you can do with your VSCode extenssion or `devpod up` setup, but this requires a significant amount of tinkering

README.md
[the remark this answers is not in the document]
acolombier: Fair, I moved it there.

> Can you give an estimate?

The [base image](https://hub.docker.com/_/ubuntu/tags?name=24.04) is about 30M, the dev container apt deps appears to require about 1.3 GB of additional disk space (as mentioned via APT, haven't confirmed the measure). It is hard to compare with native build environment tho because these are dependencies you would have to install on your native machine anyway (build-essential or ninja-build for example). For full disclosure, here is the current footprint of my dev container environment, including volumes (so not part of the container or image), in case you want to compare it with your native env:

```
localhost/vsc-mixxx-0637a   2.485GB
mixxx-bashistory            22.09kB
mixxx-ccache                0B
mixxx-config                255.5MB
mixxx-precommit             1.361GB
mixxx-build                 3.739GB
```

JoergAtGithub:
@daschuer This does not affect the build of Mixxx itself. This PR can therefore be targeted to a stable release without any risk! Please set the tzarget to 2.6 therefore!

daschuer:
Our stable branch is the 2.5 branch. If you like we can rebase it to that. Since this is independent from any release. I will remove the milestone for now. Set it again If you like.

acolombier:
Rebased against `2.5` and fixed conflict. @daschuer could you point me to what is remaining to address?

daschuer:
I have double checked again, this comment still applies: 
https://github.com/mixxxdj/mixxx/pull/14018#pullrequestreview-2669342570

After merging, every VS-Code user is urged to use UbuntuDev containers instead of native builds which is not recommended. 

No MacOS or Window user should ever build in a Linux Dev container. Also Linux user should not build in a dev container unless they have a strong reason for it. It is a high value for our project to have a high testing coverage of a the systems out there. After merging this it will become more unified which is not desired. The build done in a dev container is not the same than the final build we release.  

So please move the `.devcontainer` folder out of the root. Users who want to use it can use a symlink instead.

JoergAtGithub:
What is the problem with using a symlink here? Isn't that a one-time operation on the local file system that remains in place after a git checkout/switch?

acolombier:
No problem with creating a symlink. 

The problem that this is supposed to solve however doesn't exist and I am not interested in supporting a non-compliant hacky dev-container setup, dictated by members that don't even use them.

ywwg:
> After merging, every VS-Code user is urged to use UbuntuDev containers instead of native builds which is not recommended.

@daschuer have you confirmed that this is actually the case? You have been asserting that there would be a specific non-desirable workflow for mac or windows developers but I do not see evidence to confirm it.  Have you tested on windows or mac and confirm that this situation is true? If I understand, specifically you're saying that mac or windows developers would be encouraged to create ubuntu devcontainer builds instead of native builds.  Without confirmation of this problem, there is no basis for your requests for changes.

acolombier:
Just to clarify @ywwg, it is correct that opening a project with a devcontainer configuration will ask you if you want to open it into a devcontainer (popup at bottom right)

image

I'm arguing this isn't an urge to use it, simply an option.

Furthermore, this implies that users have installed [the addon](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers). IMO, this suggests that they would be informed of what DevContainer are, and wouldn't be trapped into using a Linux emulation against their will, when running another OS

ywwg:
This seems like a bug with VSCode, not Mixxx.

ywwg:
could we provide a little cmake command like `cmake devcontainer` (or whatever) that creates the .devcontainer folder if people want it?  Then it would not be a hacky symlink and would not exist unless someone follows specific directions to initialize it.

daschuer:
@ywwg 

> @daschuer have you confirmed that this is actually the case? You have been asserting that there would be a specific non-desirable workflow for mac or windows developers but I do not see evidence to confirm it. 

Yes. I have tested this on Ubuntu and a Windows 10 PC. https://github.com/mixxxdj/mixxx/pull/14018#issuecomment-2708808933

ywwg:
> In my case VS-Code installed automatically all requirements when you except all proposals.

I do not feel Mixxx is responsible for the user selecting an option presented to them by a third party software.  We could at the very least provide information on the wiki about whether a windows/mac developer would want this.

Regardless, what do you think of my suggestion to only create the devcontainer infrastructure if the user follows specific instructions? That would seem to address both your desire not to expose this option to users and @acolombier 's desire to have devcontainer usable without hacks.

acolombier:
> We could at the very least provide information on the wiki about whether a windows/mac developer would want this.

Yeah, I think this is entirely reasonable. It's also fair to say that IMO, the fact a user have the option to run Mixxx on a devcontainer on Windows without having to go through the painful step of installing the deps (Visual Studio and co)  feels more like a feature than a bug, but appreciate I do not dev or use Windows (The same doesn't apply on Mac where a user would need to install [xquartz](https://www.xquartz.org/) to enable graphical apps - built-in on WSL)

> what do you think of my suggestion to only create the devcontainer infrastructure if the user follows specific instructions?

The reasons why I am reluctant to go with this option is that it would require user to install `cmake`, which would prevent the use case of immutable distros  (which the reason I adopted devcontainer on Mixxx) and/or usage within integrated IDE installed via Snap or Flatpak. 
Though arguably a `ln -s` could work, but I am not sure of the implication of the bind mounting there (which is why I feel like this non-compliant setup may turn out to be a burden I don't want to carry)

ywwg:
even a shell script could accomplish the task of creating the proper directory / putting files in it, etc.  doesn't have to be cmake

acolombier:
Sure, I would even say a single one-liner command like `cp -r ./tools/devcontainer .devcontainer` would get that likely working.

However, this measure is being put in place because we assume our average contributor just blindly click CTAs when they pop in non-blocking notification. 

Again, this is another case of solutionising and in this case by breaking away from an established spec. 

Instead we could have got that in 6 months ago, and if indeed we had user raising issues, we could have moved that in another folder, or even better, we could have engaged on the [spec repo](https://github.com/devcontainers/spec), with actual prove on this limit, and get Mixxx to help the spec to progress.

PR #14020, opened 2024-12-14
(fix) Track Info: show key text as in tracks table

### LINKED ISSUE #14019: key empy in track Info, but displayed in tracks table
### Bug Description

quoting from https://github.com/mixxxdj/mixxx/pull/14011#issuecomment-2542653733
**Max Graef - Jazz 104**, FLAC

 | | |
 |---|---|
  Tracks | 8B
  Deck | 8B
 Track Properties dialog|   ---
  -> Re-imp| 8B
 Track Properties dialog |    ---
 -> Re-import from file |8B
 | | |  
 database |  ---
 file metadata|     8B

So the track models query the file metadata for missing properties?
Note that I didn't have "Syncronize metadata from/to file tags" enabled while testing.

Didn't check if this is related to and maybe fixed by #14008, will do.

### Version

2.5

### OS

_No response_

ronso0:
Fixes #14019 butit feels like this is just a workaround.
(I didn't look into the Track / TrackRecord details)

Seems like TrackRecord doesn't hold the key text from the database, whereas `BaseTrackCache::getTrackValueForColumn()` takes the Track's key text.

quoting my findings from https://github.com/mixxxdj/mixxx/pull/14011#issuecomment-2542653733 again:
**Max Graef - Jazz 104**, FLAC

 | | |
 |---|---|
 database |  ---
 file metadata|     8B
| | |
  Tracks | 8B
  Deck | 8B
 Track Properties dialog |    ---
 -> Re-import from file |8B

Based on #14011, so it's only the last commit.

### DIFF
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -744,7 +744,12 @@ mixxx::UpdateResult DlgTrackInfo::updateKeyText() {
 }
 
 void DlgTrackInfo::displayKeyText() {
-    const QString keyText = KeyUtils::keyToString(m_trackRecord.getKeys().getGlobalKey());
+    QString keyText;
+    if (m_pLoadedTrack) {
+        keyText = KeyUtils::keyFromKeyTextAndIdValues(
+                m_pLoadedTrack->getKeyText(),
+                m_pLoadedTrack->getKey());
+    }
     txtKey->setText(keyText);
 }
 
--- a/src/library/dlgtrackinfomulti.cpp
+++ b/src/library/dlgtrackinfomulti.cpp
@@ -386,7 +386,11 @@ void DlgTrackInfoMulti::updateTrackMetadataFields() {
         composers.insert(rec.getMetadata().getTrackInfo().getComposer());
         grouping.insert(rec.getMetadata().getTrackInfo().getGrouping());
         years.insert(rec.getMetadata().getTrackInfo().getYear());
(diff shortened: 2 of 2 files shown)


src/library/dlgtrackinfo.cpp
```diff
@@ -744,7 +744,7 @@ mixxx::UpdateResult DlgTrackInfo::updateKeyText() {
 }
 
 void DlgTrackInfo::displayKeyText() {
-    const QString keyText = KeyUtils::keyToString(m_trackRecord.getKeys().getGlobalKey());
+    const QString keyText = m_pLoadedTrack ? m_pLoadedTrack->getKeyText() : QStringLiteral("");
```
m0dB, line 747: Hm, I wonder if this is enough. In #14008 I found that getKeyText() could be empty and I had to use getKey().

ronso0: yeah, might also be `getKeys().getGlobalKey()`.

I'm still digging a bit, trying to understand how the metadata flow is supposed to work.

ronso0:
@m0dB so with #14008 being merged the preferred call for tracks table / track info consistency would be
`KeyUtils::keyFromKeyTextAndIdValues(pTrack->getKeyText(), pTrack->getKey())`
right?

m0dB:
Yes!

ronso0:
Alright, done.

This will create conflicts for #14022, though since that is not urgent @daschuer assigned it to 2.5.**1**

daschuer (approved):
LGTM

PR #14022, opened 2024-12-15
Key metadata track properties edit fix

daschuer:
This fixed reloading and deleting of key value in the Track properties dialog. 

These are the leftover commits form the key metadata handling PRs for #10890

### DIFF
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -732,13 +732,9 @@ void DlgTrackInfo::slotSpinBpmValueChanged(double value) {
 
 void DlgTrackInfo::updateKeyText() {
     const auto keyText = txtKey->text();
-    const auto updateResult =
-            m_trackRecord.updateGlobalKeyNormalizeText(
-                    keyText,
-                    mixxx::track::io::key::USER);
-    if (updateResult == mixxx::UpdateResult::Rejected) {
-        // Restore the current key text
-    }
+    m_trackRecord.updateGlobalKeyNormalizeText(
+            keyText,
+            mixxx::track::io::key::USER);
     displayKeyText();
 }
 
--- a/src/library/dlgtrackinfo.cpp
+++ b/src/library/dlgtrackinfo.cpp
@@ -610,7 +610,7 @@ void DlgTrackInfo::saveTrack() {
     // handlers manually to capture any changes. If the bpm or key was unchanged
     // or invalid then the change will be ignored/rejected.
     slotSpinBpmValueChanged(spinBpm->value());
-    static_cast<void>(updateKeyText()); // discard result
+    updateKeyText();
 
     // Update the cached track
     //
@@ -730,17 +730,16 @@ void DlgTrackInfo::slotSpinBpmValueChanged(double value) {
     updateSpinBpmFromBeats();
 }
 
-mixxx::UpdateResult DlgTrackInfo::updateKeyText() {
-    const auto keyText = txtKey->text().trimmed();
+void DlgTrackInfo::updateKeyText() {
+    const auto keyText = txtKey->text();
     const auto updateResult =
             m_trackRecord.updateGlobalKeyNormalizeText(
                     keyText,
                     mixxx::track::io::key::USER);
     if (updateResult == mixxx::UpdateResult::Rejected) {
         // Restore the current key text
-        displayKeyText();
     }
-    return updateResult;
+    displayKeyText();
 }
 
 void DlgTrackInfo::displayKeyText() {
@@ -749,10 +748,7 @@ void DlgTrackInfo::displayKeyText() {
 }
 
 void DlgTrackInfo::slotKeyTextChanged() {
-    if (updateKeyText() != mixxx::UpdateResult::Unchanged) {
-        // Ensure that the text field always reflects the actual value
-        displayKeyText();
-    }
+    updateKeyText();
 }
 
 void DlgTrackInfo::slotRatingChanged(int rating) {
--- a/src/track/trackrecord.h
+++ b/src/track/trackrecord.h
@@ -16,17 +16,17 @@ class TrackDAO;
 namespace mixxx {
 
 /// Effect of updating a property with a new value.
-enum class [[nodiscard]] UpdateResult{
-        /// The value has been updated and changed.
-        Updated,
+enum class UpdateResult {
+    /// The value has been updated and changed.
+    Updated,
 
-        /// The value didn't change and has not been updated.
-        Unchanged,
+    /// The value didn't change and has not been updated.
+    Unchanged,
 
-        /// The provided value is invalid or insonsistent with
-        /// any existing value(s) and has been rejected, i.e.
-        /// the current value didn't change either.
-        Rejected,
+    /// The provided value is invalid or insonsistent with
+    /// any existing value(s) and has been rejected, i.e.
+    /// the current value didn't change either.
+    Rejected,
 };
 
 // Properties of tracks that are stored in the database.
--- a/src/library/dlgtrackinfo.h
+++ b/src/library/dlgtrackinfo.h
@@ -90,7 +90,7 @@ class DlgTrackInfo : public QDialog, public Ui::DlgTrackInfo {
     void clear();
     void init();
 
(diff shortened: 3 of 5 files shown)


src/library/dlgtrackinfo.cpp
```diff
@@ -730,17 +730,16 @@ void DlgTrackInfo::slotSpinBpmValueChanged(double value) {
...
+void DlgTrackInfo::updateKeyText() {
+    const auto keyText = txtKey->text();
     const auto updateResult =
             m_trackRecord.updateGlobalKeyNormalizeText(
                     keyText,
                     mixxx::track::io::key::USER);
     if (updateResult == mixxx::UpdateResult::Rejected) {
         // Restore the current key text
-        displayKeyText();
     }
```
Swiftb0y, line 741: empty branch?

daschuer: It is now removed.

daschuer:
Rebased  to resolve conflicts.

src/track/trackrecord.h
```diff
@@ -16,17 +16,17 @@ class TrackDAO;
 namespace mixxx {
 
 /// Effect of updating a property with a new value.
-enum class [[nodiscard]] UpdateResult{
-        /// The value has been updated and changed.
-        Updated,
+enum class UpdateResult {
+    /// The value has been updated and changed.
+    Updated,
```
Swiftb0y, line 22: This tells me that the enum was deliberately made for error handling. Specifically `Rejected` should be checked because it almost always would require handling from the caller. This removes that feature, making the error checking optional (which I hope we can agree is very bad practice). 

I need more explanation why its appropriate to weaken the guarantees this enum was supposed to provide.

Swiftb0y: Concretely: What is your rationale for removing the checks you removed in this PR?

daschuer: We have discussed this earlier and conclude that `[[nodiscard]]` is usefull if discarding is an error. In this case it is only an enum. Discarding can't be an error. It might be an error depending on the returning function. In our case it is `updateGlobalKeyNormalizeText()` wich works well without error checking. So the change allows us to remove `static_cast<void>()` on these calls. See at [dlgtrackinfomulti.cpp](https://github.com/mixxxdj/mixxx/pull/14022/files#diff-39b5603d1dbe5d82fe8852e827f9291cbb8d3d6cbe297992922ec54620e971cc)

The counter example are places like this where `[[nodiscard]]` heps to avois programming errors: https://github.com/mixxxdj/mixxx/blob/a4e61f9d245079d0f00aabf9e2876c1e9285679f/src/util/compatibility/qmutex.h#L32

src/library/dlgtrackinfo.cpp
```diff
@@ -730,17 +730,12 @@ void DlgTrackInfo::slotSpinBpmValueChanged(double value) {
...
-        // Restore the current key text
-        displayKeyText();
-    }
-    return updateResult;
+void DlgTrackInfo::updateKeyText() {
+    const auto keyText = txtKey->text();
+    m_trackRecord.updateGlobalKeyNormalizeText(
+            keyText,
+            mixxx::track::io::key::USER);
+    displayKeyText();
```
daschuer, line 738: Here we need to call `displayKeyText()` independent of the result of `updateGlobalKeyNormalizeText()` because we either want to see the normalized value or the old.

src/track/trackrecord.h
Swiftb0y: > In this case it is only an enum. Discarding can't be an error.

Is it though? It seems that `Rejected` and `Unchanged` did need some special handling in some cases (which you have removed). I'm asking specifically why its fine to remove the handling for those cases here.

daschuer: I think so. I have left an inline comment: 
https://github.com/mixxxdj/mixxx/pull/14022#discussion_r1888507649

ronso0:
Is there an agreement wrt the `nodiscard` issue?
If yes, I'd also take a look / test.

Swiftb0y:
No agreement afaict. The type is made for error handling and error values should not be ignored. In the few cases they should, they can be documented explicitly using program logic (either explicitly ignore or check for them). If the check is unnecessary, the compiler can optimize them away.

daschuer:
It looks like I have not made my point clear enough regarding [[nodiscard]].

We have this rule: 

A single reasonable  void cast on a no discard function IMHO  that does not introduce a programming error rectifies that our rule is violated. There was already such a void cast in the 2.5 branch. This PR would have introduce another. 

In addition it feels flawed to put a nodiscard on a type declaration, because it limits it use, it would belong to a function. 

I am happy to reintroduce error checks if you can draft  them along with a note which error it will fix.

Swiftb0y:
> We have this rule:

That rule has been written without considering this usecase. referring back to it unconditionally right now makes no sense.

`UpdateResult` is essentially an error code. If the return value can be easily ignored, the error handling is broken. Thats why `[[nodiscard]]` on this type is the right thing to do (and keep). See the bottom of this cppcoregl that supports this argument: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e25-if-you-cant-throw-exceptions-simulate-raii-for-resource-management 

In the few instances where it makes sense to ignore the return value, use `std::ignore` or just simply check the value anyway.

daschuer:
For my understanding the use case has been considered. We have set this rule to not clutter our code with `[[nodiscard]]` and `static_cast<void>()` pairs. It is in addtion a bit cumbersome to add [[nodiscard]] to a return type, because this way it is out of sight when using a function returning the value. 

I can confirm the view form the link. The discussion is around throwing an error, which will force error handling or breaking the program and the alterntive of forcing error handling by `[[nodiscard]]`. 

See also: 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts

> Never cast to (void) to ignore a [[nodiscard]] return value. If you deliberately want to discard such a result, first think hard about whether that is really a good idea (there is usually a good reason the author of the function or of the return type used [[nodiscard]] in the first place).

In our case we have no reason to force error handling because there is nothing reasonable to handle IMHO. We do not not introduce a bug or anything so we don't need hard thinkining and hint later readers that we have ignored here something deliberal in an exeptional way. 

The fucntion in question is here: https://github.com/mixxxdj/mixxx/blob/ae2dc103bc991f1c3fd182dbcc2fa8bf535e8f77/src/track/trackrecord.cpp#L30 If you have and idea how the enforced error handling should look like, I am happy to add it to this PR, else let's merge it as it is.

cr7pt0gr4ph7:
`UpdateResult` is (was) used as an optimization to skip updating the text shown in the UI when the underlying key hasn't changed.

Due to normalization, there are actually 4 different cases (given `QString oldKey, newKey;`):

1. `oldKey == newKey && oldKey == normalize(newKey) && isValid(newKey)` -> `UpdateResult::Unchanged`, no display text update required
2. `oldKey != newKey && oldKey == normalize(newKey) && isValid(newKey)` -> still `UpdateResult::Unchanged`, **but display text update required**
3. `oldKey != normalize(newKey) && isValid(newKey)` -> `UpdateResult::Updated`, display text update required
4. `!isValid(newKey)` -> `UpdateResult::Rejected`, display text update required

So it's not really about error handling, but about optimizing the synchronization `UI <-> Data Model`. The current code fails to handle handle case 2 properly, treating it as case 1 instead. Instead of distinguishing cases 1 and 2, this PR chooses to simply always update the display text, which is correct in all cases.

I think that removing `[[nodiscard]]` is the correct way forward here, as it's not really about error handling or resource leakage, and the more complex logic required for state synchronization (exceeding a simple "check the return code") requires a detailed review anyway.

daschuer:
@Swiftb0y Is it OK to merge now?

Swiftb0y:
Sorry, I have missed the ping previously. Thank you @cr7pt0gr4ph7 for the detailed explanation. It has motivated to me to indeed take another deep look at the surrounding code and can see how `[nodiscard]` was likely used without very good reason in this particular instance. Nevertheless, weakening the guarantees a type provides always needs careful consideration (to warrant my previous firm stance).

Swiftb0y (approved):
LGTM. Thank you.

PR #14023, opened 2024-12-15
fix running mixxx-test directly

daschuer:
Some test where failing if you run mixxx-test directly instead of using ctest. This PR fixes this.


JoergAtGithub:
Code LGTM! Only test affected -> No risk for 2.5 release. Thank you!

PR #14024, opened 2024-12-16
(fix) allow seeking (eg. to hotcue) while scratching with the waveform

ronso0:
1. scratch with the mouse on waveform, eg. spinback (mouse button still pressed)
2. press hotcue in order to seek to it
= player seeks to hotcue and stays there

Previously this was not possible because the seek position was not propagated, and the player would forward/rewind back to the position where we stopped the track with insane speed.
(well, it worked if PositionScratchController didn't register a movement, ie. if the mouse has been pressed but not moved)

### DIFF
--- a/src/engine/controls/ratecontrol.cpp
+++ b/src/engine/controls/ratecontrol.cpp
@@ -628,3 +628,7 @@ void RateControl::notifyWrapAround(mixxx::audio::FramePos triggerPos,
     m_jumpPos = triggerPos;
     m_targetPos = targetPos;
 }
+
+void RateControl::notifySeek(mixxx::audio::FramePos position) {
+    m_pScratchController->notifySeek(position);
+}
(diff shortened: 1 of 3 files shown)


ronso0:
> (well, it worked if PositionScratchController didn't register a movement, ie. if the mouse has been pressed but not moved)

So the issue seems to that the filter(s) aren't reset properly when the mouse comes to halt?

ronso0:
FWIW the insane fast seek/scratch still happens (rarely) when doing this the first time in a session:
* click waveform, move mouse to the left (scratch backwards)
* press a hotcue that's behind (earlier) than the current position
* play pos settles somewhere between previous play pos and hotcue, but nowhere near we'd expect it with slow scratching

daschuer (approved):
LGTM, thank you. 
PositionScratchController::notifySeek() has not been called at all before. Must have been lost.

PR #14028, opened 2024-12-16
S4MK3: stop wheel led blinking when track is over/stopped

### LINKED ISSUE #13995: Traktor Kontrol S4 Mk3: stop `end_of_track` flashing when deck is stopped
### Bug Description

... both when deck is stopped manually and when it reaches the end and stops.
https://github.com/mixxxdj/mixxx/blob/8354c8e0f57a635acb7f4b3cc16b9745dc83312c/res/controllers/Traktor-Kontrol-S4-MK3.js#L2571

I fixed it for auto-stop with the additonal condition `fractionOfTrack !== 1.0`, but for when the deck is stopped manually I didn't manage to install a callback for `end_of track` itself and set/read a deck var appropriately (still a noob in js/components..)

@acolombier is that something that you already fixed in WIP/private versions of the script?

### Version

2.4++

### OS

_No response_

acolombier:
Fixes #13995

Arguably, the code could be slightly cleaner, but I wanted to submit a fix for this before the 2.5 release.

@ronso0 could you please have a look?

### DIFF
--- a/res/controllers/Traktor-Kontrol-S4-MK3.js
+++ b/res/controllers/Traktor-Kontrol-S4-MK3.js
@@ -2523,8 +2523,33 @@ class S4Mk3Deck extends Deck {
         this.wheelLED = new Component({
             deck: this,
             lastPos: 0,
-            outKey: "playposition",
-            output: function(fractionOfTrack) {
+            lastMode: null,
+            outConnect: function() {
+                if (this.group !== undefined) {
+                    const connection0 = engine.makeConnection(this.group, "playposition", (position) => this.output.bind(this)(position, true, true));
+                    // This is useful for case where effect would have been fully disabled in Mixxx. This appears to be the case during unit tests.
+                    if (connection0) {
+                        this.outConnections[0] = connection0;
+                    } else {
+                        console.warn(`Unable to connect ${this.group}.playposition' to the controller output. The control appears to be unavailable.`);
+                    }
+                    const connection1 = engine.makeConnection(this.group, "play", (play) => this.output.bind(this)(engine.getValue(this.group, "playposition"), play, play || engine.getValue(this.group, "track_loaded")));
+                    // This is useful for case where effect would have been fully disabled in Mixxx. This appears to be the case during unit tests.
+                    if (connection1) {
+                        this.outConnections[1] = connection1;
+                    } else {
+                        console.warn(`Unable to connect ${this.group}.play' to the controller output. The control appears to be unavailable.`);
+                    }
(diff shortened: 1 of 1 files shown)


res/controllers/Traktor-Kontrol-S4-MK3.js
```diff
@@ -2523,8 +2523,25 @@ class S4Mk3Deck extends Deck {
...
+                    // This is useful for case where effect would have been fully disabled in Mixxx. This appears to be the case during unit tests.
+                    if (connection0) {
+                        this.outConnections[0] = connection0;
+                    } else {
+                        console.warn(`Unable to connect ${this.group}.playposition' to the controller output. The control appears to be unavailable.`);
+                    }
+                    const connection1 = engine.makeConnection(this.group, "play", (play) => this.output.bind(this)(engine.getValue(this.group, "playposition"), play));
+                    // This is useful for case where effect would have been fully disabled in Mixxx. This appears to be the case during unit tests.
+                    if (connection1) {
+                        this.outConnections[0] = connection1;
```
ronso0, line 2538: Do I understand this correctly:
this will replace `connection0`?

ronso0:
There's now a small glitch:
when I stop the deck in eot range, the LED spot will briefly light up at the previous position.
* play deck
* seek into eot range
* stop deck (note LED spot position = pos1)
* play (flash)
* stop: spot briefly at pos1, then at correct new position

res/controllers/Traktor-Kontrol-S4-MK3.js
acolombier: Ah that's a bug :sweat_smile:

acolombier:
I will look into this when I can

ronso0:
Sure.
It's working as desired btw : )

acolombier:
> There's now a small glitch

Looks like this is a hardware limitation. I've debug the HID messages, and what we are saying is good. 
It looks like there is a built-in animation management of the controller, which is managed individual for each wheel lighting mode. It seems this this animation is enlightening the delta between two pos in order create a transition animation. 

I have added a nasty workaround that basically makes the backlit off during the animation and then restore the colour. This mean that LED stays off for about ~200ms, which seems to be the animation length. Or we could keep the limitation as is?

I have also added another commit which display the LED spot again if the wheel is touched inside the EOT range, and also keep the wheel off is no track is loaded.

ronso0:
Thank you looking into this glitch and finishing this!

Works nicely now!

ronso0:
Squash please, then this is ready to roll!

acolombier:
Done! Thanks for looking into this @ronso0

PR #14035, opened 2024-12-18
(fix) skins: center rate range labels (Qt6 regression)

ronso0:
Fixes a Qt6 regression:
the numbers were left-aligned, in LateNight they overlap the slider background.
It's not possible anymore to correct this with a stylesheet like this
```css
#RateRangeText {
  qproperty-alignment: AlignRight;
  text-align: right;
}
```

Fix was necessary in LateNight and Tango.

Fixed 'center' is okay for official skins. If we need more customization we can add a parser for an `Align` node.

| 2.4 | 2.5 | this PR |
|-|-|-|
image | image | image

### DIFF
--- a/res/skins/Tango/decks/rate_pitch_key.xml
+++ b/res/skins/Tango/decks/rate_pitch_key.xml
@@ -180,6 +180,7 @@ Variables:
 
               <!-- Pitch range indicators: prefix / spacer / range number -->
               <WidgetGroup>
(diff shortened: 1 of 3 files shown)


JoergAtGithub:
Could you please share screenshots, that show the issue.

ronso0:
Sure, added them to the description.

daschuer (approved):
I can confirm the issue and the fix. Thank you.

PR #14036, opened 2024-12-19
Tracks: use clicked column to focus property editor in Track Info dialog

ronso0:
follow-up for #13841 
and a small tweak for DlgTrackInfo (single track)

### DIFF
--- a/src/widget/wtrackmenu.cpp
+++ b/src/widget/wtrackmenu.cpp
@@ -2546,6 +2546,7 @@ void WTrackMenu::slotShowDlgTrackInfo() {
                 [this]() {
                     if (m_pDlgTrackInfoMulti.get() == sender()) {
                         m_pDlgTrackInfoMulti.release()->deleteLater();
+                        setTrackPropertyName();
                     }
                 });
         QList<TrackPointer> tracks;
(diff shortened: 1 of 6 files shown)


src/widget/wtrackmenu.cpp
```diff
@@ -2567,6 +2569,7 @@ void WTrackMenu::slotShowDlgTrackInfo() {
                 [this]() {
                     if (m_pDlgTrackInfo.get() == sender()) {
                         m_pDlgTrackInfo.release()->deleteLater();
+                        setTrackPropertyName();
```
daschuer, line 2572: This one was not obvious for me. How about `clearTrackPropertyName();`? 

It is a bit of an odd logic that the user of m_trackProperty needs to clean up. Did you consider to pass the property along with `loadTrack()`? This way every caller has to descide about the property to focus. 
I have not checked how big this change would be, so it is just a suggestion.

daschuer (changes requested):
That was quick. Thank you. I have found one readibility issue.

src/widget/wtrackmenu.cpp
ronso0: I just didn't want to introduce yet another method.
We could also simply reset it directly:
`m_trackproperty = QString();`

ronso0: Do you think the cleanup is an issue if it's done close to where we call `setFocus()`?

> Did you consider to pass the property along with loadTrack()?

Then we'd need to override QDialog::show() and call `setFocus()` there.

ronso0: We could also **not** reset it, so it'll be reused in case no property is specified?
Might be unexpected if it was "bpm" before (BPM tab would be selected).

daschuer: `m_trackproperty = QString();` is just fine for a quick merge. If my other idea is not a short thing, just drop it.

daschuer (approved):
LGTM, thank you.

PR #14044, opened 2024-12-20
Update skin preview images

daschuer:


ronso0:
unfortunately we can't use these screens: colors are not correct, library font is custom
See https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/2.2E5.2E0.20release.20deadline/near/490199318

ronso0:
These are still the images with the screen calibration by @Eve00000 , ie. wrong colors.
See this diff for example
image

ronso0:
Actually, the only difference 2.4 --> 2.5 I can spot at a glance in the **previews** are
* the new effect unit menu button
* no menubar anymore

Layout / features shown in the old screens is much better than the new ones.
After all the previews shoud give users an idea of the layout and colors, this is not a feature showcase like the press screenshots on the website.

= I'd rather keep the old previews than waiting for new polished ones.

(formally "requesting changes" so we don't merge the current state)

ronso0 (changes requested):
.

daschuer:
OK, works for me. I have just issued the version bump PR. https://github.com/mixxxdj/mixxx/pull/14049

daschuer:
We need to start over if we wish. closing this.

PR #14058, opened 2024-12-25
RateControl/Position ScratchController: use std::unique_ptr, PollingControlProxy etc.

ronso0:
Just some preparation for #14059
Targeting 2.5 because #14059 (or an alternative) may become an appropriate bug fix.

### DIFF
--- a/src/engine/controls/ratecontrol.cpp
+++ b/src/engine/controls/ratecontrol.cpp
@@ -30,142 +30,172 @@ RateControl::RampMode RateControl::m_eRateRampMode;
 const double RateControl::kWheelMultiplier = 40.0;
 const double RateControl::kPausedJogMultiplier = 18.0;
 
-RateControl::RateControl(const QString& group,
-        UserSettingsPointer pConfig)
+RateControl::RateControl(const QString& group, UserSettingsPointer pConfig)
         : EngineControl(group, pConfig),
           m_pBpmControl(nullptr),
+          m_pSampleRate(QStringLiteral("[App]"), QStringLiteral("samplerate")),
+          m_pRateRatio(std::make_unique<ControlObject>(
+                  ConfigKey(group, QStringLiteral("rate_ratio")),
+                  true,
+                  false,
+                  false,
+                  1.0)),
+          m_pRateDir(std::make_unique<ControlObject>(
+                  ConfigKey(group, QStringLiteral("rate_dir")))),
+          m_pRateRange(std::make_unique<ControlPotmeter>(
+                  ConfigKey(group, QStringLiteral("rateRange")), 0.01, 4.00)),
+          // We need the sample rate so we can guesstimate something close
+          // what latency is.
+          // Allow rate slider to go out of bounds so that sync lock rate
+          // adjustments are not capped.
+          m_pRateSlider(std::make_unique<ControlPotmeter>(
+                  ConfigKey(group, QStringLiteral("rate")), -1.0, 1.0, true)),
+          // Search rate. Rate used when searching in sound. This overrules the
+          // playback rate
+          m_pRateSearch(std::make_unique<ControlPotmeter>(
+                  ConfigKey(group, QStringLiteral("rateSearch")), -300., 300.)),
+          // Temporary rate-change buttons
+          m_pButtonRateTempDown(std::make_unique<ControlPushButton>(
+                  ConfigKey(group, QStringLiteral("rate_temp_down")))),
+          m_pButtonRateTempDownSmall(std::make_unique<ControlPushButton>(
+                  ConfigKey(group, QStringLiteral("rate_temp_down_small")))),
+          m_pButtonRateTempUp(std::make_unique<ControlPushButton>(
+                  ConfigKey(group, QStringLiteral("rate_temp_up")))),
+          m_pButtonRateTempUpSmall(std::make_unique<ControlPushButton>(
+                  ConfigKey(group, QStringLiteral("rate_temp_up_small")))),
+          // Permanent rate-change buttons
+          m_pButtonRatePermDown(std::make_unique<ControlPushButton>(
+                  ConfigKey(group, QStringLiteral("rate_perm_down")))),
+          m_pButtonRatePermDownSmall(std::make_unique<ControlPushButton>(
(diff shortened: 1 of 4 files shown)


src/engine/controls/ratecontrol.cpp
```diff
@@ -34,138 +34,167 @@ RateControl::RateControl(const QString& group,
...
+          // what latency is.
+          m_pSampleRate(QStringLiteral("[App]"), QStringLiteral("samplerate")),
           m_wrapAroundCount(0),
           m_jumpPos(mixxx::audio::FramePos()),
           m_targetPos(mixxx::audio::FramePos()),
           m_bTempStarted(false),
           m_tempRateRatio(0.0),
           m_dRateTempRampChange(0.0) {
-    m_pScratchController = new PositionScratchController(group);
+    m_pScratchController = std::make_unique<PositionScratchController>(group);
```
Swiftb0y, line 48: lets move this to the constructor initializer list.

src/engine/controls/ratecontrol.cpp
```diff
@@ -34,138 +34,167 @@ RateControl::RateControl(const QString& group,
...
     // The track original rate ratio is 1.
-    m_pRateRatio = new ControlObject(ConfigKey(group, "rate_ratio"),
-                  true, false, false, 1.0);
-    connect(m_pRateRatio, &ControlObject::valueChanged,
-            this, &RateControl::slotRateRatioChanged,
+    m_pRateRatio = std::make_unique<ControlObject>(ConfigKey(group, QStringLiteral("rate_ratio")),
+            true,
+            false,
+            false,
+            1.0);
```
Swiftb0y, line 56: same here and if you're up for it all the other late initializations as well.

src/engine/controls/ratecontrol.cpp
```diff
@@ -34,138 +34,167 @@ RateControl::RateControl(const QString& group,
...
+            ConfigKey(group, QStringLiteral("scratch2_indicates_scratching")));
     // Enable by default, because it was always scratching before introducing
     // this control.
     m_pScratch2Scratching->set(1.0);
 
-
-    m_pJog = new ControlObject(ConfigKey(group, "jog"));
-    m_pJogFilter = new Rotary();
+    m_pJog = std::make_unique<ControlObject>(ConfigKey(group, QStringLiteral("jog")));
+    m_pJogFilter = std::make_unique<Rotary>();
```
Swiftb0y, line 197: this is a refactoring, right? Should we rebase on `main` instead? I know I changed this line in `main` in #13853 so there are bound to be conflicts...

src/engine/positionscratchcontroller.h
```diff
@@ -6,18 +6,42 @@
...
+    RateIIFilter();
+
+    void setFactor(double factor);
+    void reset(double last_rate);
+    double filter(double rate);
+
+  private:
+    double m_factor;
+    double m_last_rate;
+};
```
Swiftb0y, line 35: I know you just moved this here into the header, but unless you intend to use these two classes outside of `PositionScratchController`, they don't make much sense in the header.

Swiftb0y (commented):
couple nits, address as you wish.

src/engine/controls/ratecontrol.cpp
ronso0: > this is a refactoring, right? Should we rebase on main instead?

Yes, and no: as metioned above, these are just preparational steps for the scratch fix #14059 which should go to 2.5 IMO

src/engine/positionscratchcontroller.h
ronso0: I wasn't able to use std::unique_ptr with the definition in ratecontrol.cpp, foward declaration didn't work out.
Any hints how to make this possible are welcome.

And I had the idea to clone `PositionScratchController` --> `WheelScratchCotroller`, so we have one for the waveforms and another one for controller wheels (maybe with some more customization, eg. controls for the max. throw/spinback time, filter parameters and maybe sample interval)

Swiftb0y: > I wasn't able to use std::unique_ptr with the definition in ratecontrol.cpp, foward declaration didn't work out.

I'm gonna have to make an educated blind guess here without a compiler log, but the issue is probably that the ctor and/or dtor got auto-instantiated in the class that used it in the header. In that case, you'll need to declare the respective special member function (ctor/dtor) (in `PositionScratchController`) and then implement it in the `.cpp` (preferably just using an explicit `= default`).

ronso0: I reverted that change and got
`/usr/include/c++/13/bits/unique_ptr.h:97:23: error: invalid application of ‘sizeof’ to incomplete type ‘VelocityController’`

So yes, adding the dtor that fixed it, thanks!
(no I understand the note in [BaseTrackCache::~BaseTrackCache()](https://github.com/mixxxdj/mixxx/blob/f0ce199b4383595ec46c92f974996cdefd198d52/src/library/basetrackcache.cpp#L37-L41))

ronso0: `= default;` didn't suffice btw

src/engine/controls/ratecontrol.h
```diff
@@ -106,51 +107,53 @@ public slots:
...
   ControlObject* m_pVCMode;
-  ControlObject* m_pScratch2Scratching;
-  Rotary* m_pJogFilter;
-
-  ControlObject* m_pSampleRate;
 
-  // For Sync Lock
-  BpmControl* m_pBpmControl;
+  PollingControlProxy m_pSyncMode;
+  PollingControlProxy m_pSlipEnabled;
```
Swiftb0y, line 150: these technically need renaming now that they're not ptrs anymore.

Swiftb0y (commented):
LGTM otherwise.

src/engine/positionscratchcontroller.h
Swiftb0y: It should've if you've done it in the `.cpp` (not in the header!)

Anyways, not important in this case

Swiftb0y (commented):
did another round and found another couple nits. Pls address and rebase squash afterwards so we can merge.

Swiftb0y (approved):

daschuer:
All Green. Thank you.

Issue #13981, opened 2024-12-05
pressing hotcue while holding waveform or spinny should seek to hotcue (and stay there)


ronso0:
### Bug Description

* play a track
* click and hold waveform or spinny, playback stops
* press hotcue (keyboard or controller)

**Current behaviour:**
* play pos seeks to hotcue and 'scratches' back to previous play position (where the waveform/spinny stopped)

**Expected behaviour:**
* seek to hotcue and stay there (or rather: continue the scratch move as if nothing happened)

### Version

2.5

PR #14059, opened 2024-12-25
(fix) Waveform scratching: fix seeking to hotcue

ronso0:
This is a first (still dirty) fix for #13981 

**TL;DR**
it's now possible to seek to hotcues while doing wave scratches.
Scratching at/across  loop boundaries (active) still works.

Recently I [compared the available methods for scratching with controllers](https://github.com/mixxxdj/mixxx/pull/14004), and while PositionScratchController (developed for waveform scratching) turned out to be the best (for me, no drift, smooth at low rates, nice spinback) I couldn't really use it as I want to because seeking to hotcues while scratching (holding the wheel, or turning very slowly) produced insanely high rates, and didn't stop at the cue pos.

Even though the bug is fixed I still consider this 'dirty' because I don't fully understand what was going wrong with the filters after seeking.

### DIFF
--- a/src/engine/readaheadmanager.cpp
+++ b/src/engine/readaheadmanager.cpp
@@ -283,22 +283,8 @@ double ReadAheadManager::getFilePlaypositionFromLog(
     }
 
     double filePlayposition = 0;
-    bool shouldNotifySeek = false;
     while (m_readAheadLog.size() > 0 && numConsumedSamples > 0) {
         ReadLogEntry& entry = m_readAheadLog.front();
-
-        // Notify EngineControls that we have taken a seek.
-        // Every new entry start with a seek
-        // (Not looping control)
-        if (shouldNotifySeek) {
-            if (m_pRateControl) {
-                const auto seekPosition =
-                        mixxx::audio::FramePos::fromEngineSamplePos(
-                                entry.virtualPlaypositionStart);
-                m_pRateControl->notifySeek(seekPosition);
-            }
-        }
-
         // Advance our idea of the current virtual playposition to this
         // ReadLogEntry's start position.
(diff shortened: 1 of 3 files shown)


src/engine/readaheadmanager.cpp
```diff
@@ -283,22 +283,8 @@ double ReadAheadManager::getFilePlaypositionFromLog(
...
-        // (Not looping control)
-        if (shouldNotifySeek) {
-            if (m_pRateControl) {
-                const auto seekPosition =
-                        mixxx::audio::FramePos::fromEngineSamplePos(
-                                entry.virtualPlaypositionStart);
-                m_pRateControl->notifySeek(seekPosition);
-            }
-        }
-
```
ronso0, line 301: Until now this was a no-op in RateControl and with the fix it's causing trouble so I removed it.

ronso0:
Still not working as desired. I'm a bit lost here since I don't fully understand how the filters work here:
Either seeking works _mostly_ and loop-scratching works _mostly_, or seeking works perfectly and loop-scratching is broken sometimes (now I'm using the latter for my mapping since seeking is more importat to me). 
Note that I've been testing this lately with a controller implementation, ie. `scratch_position` is set with my jog wheel position. (which works flawlessly btw)

So, I really don't get how to fix both as I'm not 100% sure loop-scratching works 100% in main.
Maybe the sampling intervals are not workign well  for controller mapping which sends updates much faster that the audio engine @5ms latency, so maybe we can ditch the interval stuff and simply process at ever call from RateControl.
That said, it might make sense to sub-class / clone PositionScratchController for controller use.

Any help is appreciated!

daschuer:
This is now conflicting.

PR #14060, opened 2024-12-26
ci(dependabot): Open PRs against 2.5 branch instead of main

Holzhaus:


JoergAtGithub:
LGTM! Thank you!

daschuer:
This change has enabled the dependant on my fork, which is highly undesired. Workflow upgrades shall be manager on upstream only. 
Normally forks need to explicit enable dependabot. In my case it is still disabled.

PR #14078, opened 2024-12-30
Tooltips: always show if Ctrl key is pressed

ronso0:
I don't have tooltips enabled because I usually don't need them.
Though, sometimes I _do_ need them to check hotkeys but else I don't want them to get in the way.

So with this, if you have tooltips disabled you can press `Ctrl` and hover a widget (or some library item) they'll be shown..

### DIFF
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -1250,8 +1250,9 @@ void MixxxMainWindow::tryParseAndSetDefaultStyleSheet() {
 
 /// Catch ToolTip and WindowStateChange events
 bool MixxxMainWindow::eventFilter(QObject* obj, QEvent* event) {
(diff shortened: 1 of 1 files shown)


acolombier (commented):
Tested and this LGTM!

This is a very nice little improvement - I also struggle a lot with tooltips always on, but sometime do need a reminder of what right click does :)

acolombier (approved):
Tested and this LGTM!

This is a very nice little improvement - I also struggle a lot with tooltips always on, but sometime do need a reminder of what right click does :)

ronso0:
oh, did I just hide an easter egg?

We should probably mention it somewhere. The Interface preferences are compact, how about adding a line there?
Or just an :information_source: tag with a tootip?
What do you think?

PR #14079, opened 2024-12-30
Controller preferences: don't break support link texts

ronso0:
Before, links could be broken due to wordwrap="true" which is confusing, just weird and not necessary:
[Mixxx Forums](..) [Mixxx](.)
[Manual](.) [Troublsehooting](.)

Now it looks like this:
[Mixxx Forums](..) [Mixxx Manual](.)
[Troublsehooting](.)

### DIFF
--- a/src/controllers/dlgprefcontroller.cpp
+++ b/src/controllers/dlgprefcontroller.cpp
@@ -343,7 +343,7 @@ QString DlgPrefController::mappingSupportLinks(
     if (!forumLink.isEmpty()) {
         linkList << coloredLinkString(
                 m_pLinkColor,
-                "Mixxx Forums",
+                "Mixxx&nbsp;Forums",
                 forumLink);
     }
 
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
I cannot reproduce this, because there is never a word wrap and always the horizontal scrollbar apears when the window width is reduced too much. This is independent if I youse 2.5 or this PR:
grafik

ronso0:
Okay, 2.5 is the wrong target.
I think I noticed this with your PR #14006

Btw in main both device and mapping info is just cropped, there are no horizontal scrollbars:
image

ronso0:
Reopening this since it's (still) happening with main: Mixxx / Manual
image

JoergAtGithub:
I just tried this with Main and reuced the dialog size to minimum width. I see no line break:
grafik

Swiftb0y:
these are different OSes? Maybe there are different platform specific defaults that cause this?

ronso0:
Please try with the **AKAI MPD24** mapping.
If other parts of the mapping description make the box expand it might look okay.

JoergAtGithub:
It works with 2.5 as with this PR. Code LGTM! Let's merge it therefore!

PR #14087, opened 2024-12-30
VersionStore: Fix ARM64 platform string

### LINKED ISSUE #14086: Regression: macOS ARM builds display incorrect architecture in About dialog
### Bug Description

A minor regression I've noticed in recent builds is that the "About" dialog reports the wrong architecture for arm64 macOS builds, this should be "ARM64":

Screenshot 2024-12-30 at 22 33 31

A quick bisect shows that the issue was introduced between 98012f3c9a (good) and 9f43c3e798 (bad) (i.e. in [this diff](https://github.com/mixxxdj/mixxx/compare/98012f3c9aa4b12d098a50c0314d5ca467cba95d...9f43c3e798bb05f835ac9ee85403317586723bb1)), so very likely in this PR:

- https://github.com/mixxxdj/mixxx/pull/13687

cc @Swiftb0y 

### Version

2.6-alpha-144-g8354c8e0f5

### OS

macOS 15.1.1

fwcd:
### Fixes #14086 

This fixes a small regression introduced in #13687, which inadvertently changed the version string to "IA64" for arm64.

### DIFF
--- a/src/util/versionstore.cpp
+++ b/src/util/versionstore.cpp
@@ -113,7 +113,7 @@ QString VersionStore::applicationName() {
 #elif defined(IA64)
 #define PLATFORM_STR "IA64"
 #elif defined(__aarch64__)
-#define PLATFORM_STR "IA64"
+#define PLATFORM_STR "ARM64"
 #elif defined(__arm__) || defined(__thumb__) || defined(_ARM) || \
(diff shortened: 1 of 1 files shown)


Swiftb0y (approved):
Whoops. Copy paste issue. Sorry for the trouble. LGTM. Thank you.

PR #14091, opened 2024-12-31
Playlists: add 'Unlock all' and 'Delete all unlocked' menu actions

### LINKED ISSUE #8960: Delete multiple playlists at once
Reported by: **[christiegrinham](https://launchpad.net/~christiegrinham)**
Date: 2017-10-30T18:45:06Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1728686](https://bugs.launchpad.net/bugs/1728686)

---

It would be useful to be able to delete multiple playlists at once.
For example, I have hundreds and hundreds of playlists in my history. I want to tidy them up so that only ones from actual club nights are left, but to do this I must delete many playlists one at a time.

Ideas to accomplish this:
* Enable Shift - Click selection of playlists in the side bar
* When you click on "History", the main library panel shows an information panel. List the playlists here and enable shift-click selection

ronso0:
Another coffee break feature I found in my git stash.
Kind of fixes #8960 

This add two actions to the **Playlists** context menu (sidebar) like the one we already have in the History feature:
* Delete all playlists
* Delete all unlocked playlists

Both need to be confirmed twice.

~~Will do the same for crates soonish.~~

### DIFF
--- a/src/library/dao/playlistdao.cpp
+++ b/src/library/dao/playlistdao.cpp
@@ -253,6 +253,34 @@ bool PlaylistDAO::deletePlaylists(const QStringList& idStringList) {
     return true;
 }
 
+bool PlaylistDAO::deletePlaylistsByType(
+        PlaylistDAO::HiddenType type,
+        bool unlockedOnly) {
+    QString queryStr =
+            "SELECT id FROM Playlists "
+            "WHERE hidden = :type";
+    if (unlockedOnly) {
+        queryStr.append(" AND locked = 0");
+    }
+    QSqlQuery query(m_database);
+    query.prepare(queryStr);
+    query.bindValue(":type", static_cast<int>(type));
+    if (!query.exec()) {
+        LOG_FAILED_QUERY(query);
+        return false;
+    }
+
+    QStringList idStringList;
+    while (query.next()) {
+        idStringList.append(query.value(0).toString());
+    }
+    qInfo() << "Prepared deletion of" << idStringList.size()
+            << QString(unlockedOnly ? "unlocked" : QString())
+            << "playlists of type" << type;
+
+    return deletePlaylists(idStringList);
+}
+
 bool PlaylistDAO::deleteUnlockedPlaylists(QStringList&& idStringList) {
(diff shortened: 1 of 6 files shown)


daschuer (commented):
The Code Looks good. The workflow can be improved. 

I don't like the two popups in a row. It feels as if we don't take user decisions serious. 

I would also expect to "allways" respect the Lock.

Let's repeat the pattern of single delete: 
* Only delete unlocked playlists with "one" confirmation.
* Add a feature to unlock all playlists. If we add a confirm dialog as well. We have still a double safety net. 

If you are still concerned about unwanted mass delete. We may add a text input where users need to type "y"  for deletion. We may even make it a math challenge "type the result of 4+2"

ronso0:
In History we have two dialogs as well to avoid accidenst. IIRC think you proposed the second one to be a 'summary' so it doesn't come across as "we don't take user decisions serious", which works for me.

image
image

ronso0:
> I would also expect to "allways" respect the Lock.
Add a feature to unlock all playlists.

Hmmyeah, that would work for me.
I'll add the 'Unlock all' action and add a "Also delete locked playlists" checkbox.

ronso0:
Otoh, with an 'Unlock all" we can keep "Delete all unlocked playlists" and don't need the "Also delete locked" checkbox.

daschuer:
This has a conflict.

daschuer (changes requested):

daschuer (commented):
The fix looks good. You may squash it. Than I can merge after a final test.

daschuer:
You can squash. 

However I have found another issue. Deleting does not happen when only one Playlist is deleted. 
It does not matter if there are sibling locked playlists or not.

ronso0:
Pff, stupid typo.
all..working..now

ronso0:
..and sqashed

daschuer (approved):
Works like a charm now. LGTM

ronso0:
Cool. Thanks for taking the time to test it!

Issue #13260, opened 2024-05-22
Effects Routing: Echo cuts off when Unassigning Effect Unit from Deck


ZenKyma:
### Bug description
When playing a single deck, I often build up tension with the Echo Effect before a drop and let the feedback fade out from the buffer after the drop. My first impulse was to use the deck's _**Assign Effect Unit**_ `FX1`/`FX2` button to cut the signal input to the Echo buffer, but then the Echo Effect immediately stops processing.

### Current workaround
One could achieve the desired fade out by turning down the `Send` knob of the Echo, but the problem is that this knob is not always immediately available when using meta-knobs mainly. Furthermore the `Send` knob does not differentiate between input sources, while they can be easily controllable with the `FX1`/`FX2`  buttons. This could allow letting one track fade out while keeping input steady for a second track.

### Expected behaviour
The expected behaviour of the _**Assign Effect Unit**_ `FX1`/`FX2` buttons is to cut the input signal to the effect processors, but not to interfere with effect signal output. The _**Enable Effect**_ buttons provide this functionality already, if one wishes to disable the effect.

### Motivation
Allowing unassigned effect units to continue processing and outputting buffered signals would allow for a more versatile usage of effects and correspond better to the expectations of DJs starting off with Mixxx.

Sidenote: The behaviour described in #11385 (possibly related) could not be confirmed, even with a second dummy track the effect output suddenly cuts off. 

### Version

2.4.1

### OS

Linux x86_64

manyfew:
Traktor user here looking to switch, I'm testing MIXXX again after a few years and it's come a long way, so much so that it's almost as good as Traktor and in some areas even better. However, the lack of reverb and echo tails for me is a deal-breaker. In Traktor you can disable the effect, which kills the tail, and you can also disable the send to the effect, which effectively cuts the input to the effect but the effect continues to process.

Issue #11385, opened 2023-03-17
Allow effect to continue process if the buffer isn't empty


acolombier:
### Feature Description

_(Initial discussion on [Zulip](https://mixxx.zulipchat.com/#narrow/stream/109122-general/topic/Allow.20effect.20to.20process.20without.20channel.20attached.20to.20them))_

When effects stops being used (e.g all channel de-assigned, or stopped), there should be a way to let the effect continue processing even if the buffer is empty.

Note that the current behavior will always lead to abrupt cut.

Here is a suggestion on how to achieve this:
- Allow effects to control whether or not they should process (internal API change)
- Add a button parameter to Echo and Reverb to allow the effect to continue processing if there is still audio in the buffer

It's worth to say that this can currently be achieved by assigning an dummy silently playing track. When de-assigning the track on which the audio effect is being applied (not the dummy one), the effect will continue process and smothly fade out,

PR #14099, opened 2025-01-01
Allow effect to continue processing if its buffer isn't empty (builtin reverb/delay)

mrnicegyu11:
### Summary

Hey guys, this imperfection of mixxx's default effects has frustrated me personally when using mixxx in the past, and now I have set out to potentially fix it ;) . This functionality feels much better than the current "hard-cutoff" for me. It enhances the perception of turning on/off reverb or delay effects (that have a buffer) by playing the effect until the buffer is empty. This is how the reverb/echo effects of e.g. Pioneer Digital DJ-Mixers behave. 

Works smoothly in my local builds, all tests are passing for this code. I would be very happy about a review and comments on what would be needed to potentially get this in :--)

Cheers, all the best to you.

### Related Issues

- https://github.com/mixxxdj/mixxx/issues/11385
- https://github.com/mixxxdj/mixxx/issues/13260

### Notable Code changes

- The `EffectEnableState::Disabling` state would previously be used only for one buffer-processing iteration, as far as I can see mainly to ramp volume. With this new code, the `EffectEnableState::Disabling` can be prolonged by effects responding to the call to `isReadyForDisable()` with `false`, and the effect can continue playing sound.
- The `EffectProcessor` base-class gets the new function `virtual bool isReadyForDisable(void)` that defaults to true (current behaviour, hard cutoff of sound when effect is switched off). By overwriting this function effects can control weather they still have sound to play, even after the "shutoff" button has been pressed, or not.

### When will effects then be disabled?
- Reverb: One sample-buffer of the reverb's combined dry-wet output is identical to the input. More precisely, the *energy* (summed abs values) of one sample-buffer of the combined dry-wet output is identical to the input buffer's energy, up to a small threshold.
- Echo (Feedback < 1): When the delay's ringbuffer is is approx. zero everywhere, all echos are faded out.
- Echo (Feedback == 1): Turning the effect off would now never actually turn it off, since the delayline would never empty and the tail that continues playing is infinitely long. The echo would perpetually stay in `EffectEnableState::Disabling`. I find this not ideal, although one seldomly sets feedback==1 in practice I guess. Any ideas?
- All other effects: Behavior is as before, volume is ramped down over one buffer-processing length and practically speaking effects are switched off immediately.

### Ideas
- At first I wanted to add a button to the echo&reverb effects that control the shut-off behavior, but after pondering this idea for a few days I convinced myself that a reverb or a delay emptying their buffers is the natural and expected user experience, and the commercial tools also implement the effects like this usually. An extra button I guess is not warranted, no such "shutoff-behaviour-control-button" would e.g. be found on hardware delay/reverb units. But this is I guess opinionated and maybe debatable :--)


JoergAtGithub:
Happy New Year and Welcome at Mixxx!

mrnicegyu11:
Hi @JoergAtGithub , happy new year as well to you and thanks for answering! I have just signed the `Mixxx Contributor Agreement` :muscle:

src/effects/backends/effectprocessor.h
```diff
@@ -110,6 +110,9 @@ class EffectProcessor {
     /// the dry signal is delayed to overlap with the output wet signal
     /// after processing all effects in the effects chain.
     virtual SINT getGroupDelayFrames() = 0;
+    virtual bool isReadyForDisable(void) {
+        return true;
+    };
```
Swiftb0y, line 115: I'm unsure if a virtual member is the best solution. Maybe a better solution would be to return a struct from the `process` member function instead.

src/effects/backends/builtin/reverbeffect.cpp
```diff
@@ -109,13 +112,22 @@ void ReverbEffect::processChannel(
...
+                ++i) {
+            differenceSum += fabsf(pOutput[i] - pInput[i]);
+        }
+        // Calculate average of the differences
+        const float averageDifference = differenceSum / tailCheckLength;
+        constexpr float threshold = 0.002f;
+        const bool reverbTailFullyFaded = averageDifference < threshold;
+        if (reverbTailFullyFaded) {
+            m_isReadyForDisable = true;
+        }
```
Swiftb0y, line 131: Then lets make it a function

src/effects/backends/builtin/echoeffect.cpp
```diff
@@ -236,16 +242,26 @@ void EchoEffect::processChannel(
...
+            differenceSum += fabsf(pGroupState->delay_buf[i]);
+        }
+        // Calculate average of the differences
+        constexpr float threshold = 0.00001f;
+        const bool echoTailFullyFaded = differenceSum < threshold;
+        if (echoTailFullyFaded) {
+            m_isReadyForDisable = true;
+
+            pGroupState->delay_buf.clear();
+        }
```
Swiftb0y, line 264: same here. lets just actually make a function.

Swiftb0y (changes requested):
Thanks for your feature. Lets iterate on this a bit.

src/effects/backends/effectprocessor.h
mrnicegyu11: Thanks for the feedback let me answer a bit longer:

I went for the virtual members since this was I guess the most atomic way to get this working w.r.t. code-lines and logic that would have to be touched. It works quite naturally by letting the "Disabled" state of the effect stay longer, during the tail. All that was needed to be changed in the `src/engine` files was the one line change to "ask the effect if it is quiet yet", because the engine will not call the `process` function at all if the effect is in the state `Disabled`.

So from how I see it, without further code changes we can't let the effect be set to "Disabled" while it has its tail playing, states "Disabling" or "Enabled" are fine (but the power button was pressed so "Enabled" is also not appropriate...).

I guess what you propose is to return a `bool` ( inside of a struct) from the `process` function in `engineffect.h/.cpp` that indicates that the effect is still requesting to play its tail. This would then be accounted for a few lines down to a agument the conditional  `if (effectOnChannelState == EffectEnableState::Disabling) ` and not set the effect to `Disabled` if it requests to still stay active playing its tail. Are your objections from a performance point-of-view? Then for sure there might be better ways, I do not understand the while interconnectedness of mixxx at this point. In this case I'll of course change the code according to your suggestion, if you can confirm that this is what you had in mind. Otherwise it seems like a cosmetic change to me, but I am still happy to change it so it might comply better with how you guys code mixxx usually :--)

let me know, cheers

Swiftb0y: Actually I just looked the the surrounding code, ignore the suggestion for now. I thought it'd make sense style-wise but it doesn't. Eliminating unnecessary virtual function calls is a topic for a separate PR. Lets focus on other areas for now.

Swiftb0y: Sorry for the bogus suggestion.

Swiftb0y: The primary motivation was that in the effects `m_isReadyForDisable` only gets changed in the `process` function, so there is quite a lot of boilerplate to manage that state in the class. Producing that as a return value of the `process` function would simplify that a bit.

mrnicegyu11: Alright no worries at all :--)

src/effects/backends/builtin/reverbeffect.cpp
mrnicegyu11: I made it into a function :--)

src/effects/backends/builtin/echoeffect.cpp
mrnicegyu11: I made it into a function :--)

src/effects/backends/builtin/echoeffect.cpp
```diff
@@ -118,6 +118,17 @@ void EchoEffect::loadEngineEffectParameters(
     m_pTripletParameter = parameters.value("triplet");
 }
 
+float averageSampleEnergy(SINT delayBufferSize, mixxx::SampleBuffer const& delay_buffer) {
+    float differenceSum = 0.0f;
+    for (SINT i = 0; i < delayBufferSize;
+            i = i + 16) { // We don't need to consider *all* samples,
```
Swiftb0y, line 124: Are you sure? This just seems like downsampling that would introduce artifacts and miss high frequency signals. Is this just an optimization?

mrnicegyu11:
@Swiftb0y thanks for the thorough review, definitely some valid points. January is a bit hectic for me, I will address them and ask for re-review when I have some quiet time :--)

acolombier:
Thanks for looking into this feature @mrnicegyu11! I am definitely interested in getting this in 2.6 beta, planned for the end of the month. Do you need some help to complete this PR?

mrnicegyu11:
@acolombier Thanks for your kind words, I will try to address the comments in the PR this weekend. If I dont manage till then please feel free to jump in. 

Just as a disclaimer: I've worked with `juce` and such in the past but I don't know the details of the `mixxx` codebase, so my codestyle is probably off. Nevertheless, to the best of my judgement, my codechanges should just work. Actually I use my local build for DJ'ing since I made the changes, and it works like a charm on my archlinux laptop :--)

mrnicegyu11:
Looking again I agree with the reviewer fully though that I have some ambiguous naming and some premature optimizations in there, probably best to clean that up once more

src/effects/backends/builtin/reverbeffect.h
```diff
@@ -54,10 +57,17 @@ class ReverbEffect : public EffectProcessorImpl<ReverbGroupState> {
...
     EngineEffectParameterPointer m_pDampingParameter;
     EngineEffectParameterPointer m_pSendParameter;
 
     DISALLOW_COPY_AND_ASSIGN(ReverbEffect);
 };
+
+float averageSampleDifferenceEnergy(const SINT samplesPerBuffer,
+        const CSAMPLE* buffer_in,
+        const CSAMPLE* buffer_out,
+        const SINT tailCheckLength);
```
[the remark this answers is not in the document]
mrnicegyu11: true thanks!

src/effects/backends/builtin/echoeffect.cpp
mrnicegyu11: good point, let's start with the simple thing considering all sample. Was a premature optimization-attempt

mrnicegyu11: reverted to `i = i + 1`

src/effects/backends/builtin/echoeffect.cpp
```diff
@@ -118,6 +118,17 @@ void EchoEffect::loadEngineEffectParameters(
...
+                          // assuming 44.1kHz sample-rate this sums every
+                          // ~1/3ms
+        differenceSum += fabsf(delay_buffer[i]);
+    }
+    return differenceSum;
```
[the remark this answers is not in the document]
mrnicegyu11: Correct, I made it return an average while the actual cutoff overall stays the same

src/effects/backends/builtin/echoeffect.cpp
```diff
@@ -118,6 +118,17 @@ void EchoEffect::loadEngineEffectParameters(
     m_pTripletParameter = parameters.value("triplet");
 }
 
+float averageSampleEnergy(SINT delayBufferSize, mixxx::SampleBuffer const& delay_buffer) {
```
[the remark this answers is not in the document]
mrnicegyu11: done but I feel a bit insecure about this, please have a look if this is as you intended ;)

mrnicegyu11:
Alright, change requests are addressed :mechanical_arm: hope this is ok, let me know, thanks for the very friendly communication to you guys, very welcoming :--)

src/effects/backends/builtin/echoeffect.cpp
```diff
@@ -118,6 +118,14 @@ void EchoEffect::loadEngineEffectParameters(
     m_pTripletParameter = parameters.value("triplet");
 }
 
+float averageSampleEnergy(mixxx::SampleBuffer const& delay_buffer) {
```
Swiftb0y, line 121: don't take a reference to an owning type if you can take the non-owning counterpart directly.
```suggestion
float averageSampleEnergy(std::span<const CSAMPLE> delay_buffer) {
```

src/effects/backends/builtin/echoeffect.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <QMap>
+#include <span>
```
Swiftb0y, line 5: move this to the `.cpp`. You don't need it in the header.

src/effects/backends/builtin/echoeffect.cpp
```diff
@@ -236,16 +250,17 @@ void EchoEffect::processChannel(
...
-        SampleUtil::applyRampingGain(pOutput, 1.0, 0.0, engineParameters.samplesPerBuffer());
-        pGroupState->delay_buf.clear();
         pGroupState->prev_send = 0;
-    } else {
-        pGroupState->prev_send = send_current;
+        const SINT delayBufferSize = pGroupState->delay_buf.size();
+        // Calculate if the delayline-buffer is approx. zero/empty.
+        const float avgSampleEnergy = averageSampleEnergy(pGroupState->delay_buf);
+        // If echo tail fully faded
+        if (avgSampleEnergy < (0.00001f / delayBufferSize)) {
```
Swiftb0y, line 260: why does the energy depend on the buffer size? The average shouldn't be influenced by the sample size of the dataset. 

If the concern is that the fading tail influences the current average, simply don't sample it. Only sample as much as you need:

src/effects/backends/builtin/echoeffect.cpp
```diff
@@ -236,16 +250,17 @@ void EchoEffect::processChannel(
...
+    pGroupState->prev_send = send_current;
     if (enableState == EffectEnableState::Disabling) {
-        SampleUtil::applyRampingGain(pOutput, 1.0, 0.0, engineParameters.samplesPerBuffer());
-        pGroupState->delay_buf.clear();
         pGroupState->prev_send = 0;
-    } else {
-        pGroupState->prev_send = send_current;
+        const SINT delayBufferSize = pGroupState->delay_buf.size();
+        // Calculate if the delayline-buffer is approx. zero/empty.
+        const float avgSampleEnergy = averageSampleEnergy(pGroupState->delay_buf);
```
Swiftb0y, line 258:
```suggestion
        const float avgSampleEnergy = averageSampleEnergy(pGroupState->delay_buf.span().first(engineParameters.sampleRate()));
```

Swiftb0y (commented):
couple more nits

src/effects/backends/builtin/echoeffect.cpp
mrnicegyu11: The average sample energy depends on the buffer-size since it is computed from the buffer (average over buffer length, the average over the next buffer as well, or only half the buffer, would be numerically different). Thus, `avgSampleEnergy` implicitly depends on the buffer length.

Before your previous review, the `averageSampleEnergy` would not return an average but the summed Sample Energy over the buffer, this was simply bad code and is now fixed. To retain the exactly same threshold which was test-driven by me as before, the `0.00001f` magic number now needs to be deleted by the `delayBufferSize`. Of course I can also shift the `0.00001f`  a bit by making it 100x smaller or so and call it a day :--) let me know

src/effects/backends/builtin/echoeffect.h
mrnicegyu11: done :--)

mrnicegyu11:
@Swiftb0y please check again :--)

src/effects/backends/builtin/echoeffect.cpp
Swiftb0y: > Thus, avgSampleEnergy implicitly depends on the buffer length.

Well, now that we're only looking at the samples at the beginning of the delay buffer (exactly the samples that feedback into the effect to be exact) the size shouldn't anymore, right? Am I making sense here?

Besides I think `.first(engineParameters.sampleRate())` is probably wrong, my mistake. I meant `.first(<how big the current playback buffer is>)`...

mrnicegyu11: Mhm I am not sure I follow sorry, I think the average is calculated over the whole delay buffer. But we can in any case remove the division by the delay buffer length in the treshhold check. This is just numerics:--)

src/effects/backends/builtin/echoeffect.cpp
```diff
@@ -118,6 +120,14 @@ void EchoEffect::loadEngineEffectParameters(
     m_pTripletParameter = parameters.value("triplet");
 }
 
+float averageSampleEnergy(std::span<const CSAMPLE> delay_buffer) {
```
daschuer, line 123: Energy is voltage * current. Here you use only voltage. 
```suggestion
float averageSampleLevel(std::span<const CSAMPLE> delay_buffer) {
```

daschuer: We collect here such sample functions that are reusable together with a hint that indicates if the function is auto vectorized.

This is probably already the case here, but worth to double check. 

https://github.com/mixxxdj/mixxx/blob/ae2dc103bc991f1c3fd182dbcc2fa8bf535e8f77/src/util/sample.cpp#L463

daschuer (changes requested):
Thank you for this nice PR and sorry for juming in late. 
I have left some comments.

daschuer:
CI is failing, back to draft for now.

acolombier:
Hi @mrnicegyu11, are you still interested in completing this? Feel free to ask in case you need any help with the CI issues.

mrnicegyu11:
@acolombier hey, due to my main job and various family things I am very pressed for time right now, and the laptop I have made the changes on is defunct. So currently I dont have a dev-env setup. While I might pick this up eventually when muse or time hits me, please yourself (or anyone else) feel free to pick this up. As said my changes are effectively functional, but I guess I made some bad choices w.r.t. signal processing details and such. Might be a quick thing for someone who is more routinely doing cpp signal processing. 

So, if anyone reads this and wants the feature, please feel free to jump in :--) Sorry and thanks for understanding

[ CC  fyi @daschuer potentially ]

mrnicegyu11:
Finally, I have spent some more time and I think I have addresses all review comments, and merged the upstream changes is, feel free to consider it again @daschuer @acolombier :--)

Sorry for the very long delay

acolombier:
@daschuer @Swiftb0y could you please have a look again and see if your change request has been addressed?

PR #14106, opened 2025-01-02
fix: handle `script.deckFromGroup(undefined)` gracefully

Swiftb0y:
This fixes an exception occuring since #13425 where constructing a `components.JogWheelBasic` with only a `deck` number with a matching `group` would result an exception (since `script.deckFromGroup` would call `substring` on `undefined`).

CC @christophehenry


daschuer (approved):
LGTM, Thank you.

Swiftb0y:
Thank you.

PR #14107, opened 2025-01-02
Rotary: Fix accidental quantization

### LINKED ISSUE #14095: Regression: Strangely slow jog wheel scrubbing (with Denon MC7000, but potentially others too?)
### Bug Description

In recent 2.6/main builds scrubbing the jog wheel has become inconsistently slow, to the point where the jogs don't move until spinned very quickly. Kind of hard to demo, but here's the bad version (watch the LEDs):

https://github.com/user-attachments/assets/87742c56-dedc-4b74-a22b-01f3a78d9664

Here's the good version:

https://github.com/user-attachments/assets/dc382d97-fea8-4c9c-8201-8c88b368fb46

The track moves accordingly. In practice it feels like the sensitivity of the jog wheels is way lower than its supposed to be.

A quick bisect pinned the issue down to 98012f3c9a (last good) to 9f43c3e798 (first bad) ([diff](https://github.com/mixxxdj/mixxx/compare/98012f3c9aa4b12d098a50c0314d5ca467cba95d...9f43c3e798bb05f835ac9ee85403317586723bb1)), so very likely introduced here:

- https://github.com/mixxxdj/mixxx/pull/13687

We should probably get better testing infrastructure in place to avoid regressing on things like this after larger refactorings.

cc @Swiftb0y 

### Version

2.6 (main)

### OS

macOS 15.2

fwcd:
### Fixes #14095 

This fixes the issue where `std::accumulate` would accidentally get instantiated to `int` (and thus quantize the entire sum to integers), causing the rotary filter to get insensitive to small changes, as they'd just be zeroed out. I have added a unit test to verify this.

I presume the lesson learned is that we should be careful about blindly replacing loops with `std::accumulate` or the like.

### DIFF
--- a/src/test/rotary_test.cpp
+++ b/src/test/rotary_test.cpp
@@ -17,6 +17,13 @@ TEST(Rotary, basicOverflow) {
     EXPECT_EQ(rot.filter(-2), -2);
 }
 
+TEST(Rotary, noQuantization) {
+    Rotary rot(2);
+    EXPECT_EQ(rot.filter(0.25), 0.125);
+    EXPECT_EQ(rot.filter(0.0), 0.125);
+    EXPECT_EQ(rot.filter(0.0), 0.0);
+}
+
 TEST(Rotary, hugeFilterSequential) {
     static constexpr int size = 200;
     Rotary rot(size);
--- a/src/test/rotary_test.cpp
+++ b/src/test/rotary_test.cpp
@@ -19,9 +19,9 @@ TEST(Rotary, basicOverflow) {
 
 TEST(Rotary, noQuantization) {
     Rotary rot(2);
-    EXPECT_EQ(rot.filter(0.25), 0.125);
-    EXPECT_EQ(rot.filter(0.0), 0.125);
-    EXPECT_EQ(rot.filter(0.0), 0.0);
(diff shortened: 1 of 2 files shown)


src/test/rotary_test.cpp
```diff
@@ -17,6 +17,13 @@ TEST(Rotary, basicOverflow) {
     EXPECT_EQ(rot.filter(-2), -2);
 }
 
+TEST(Rotary, noQuantization) {
+    Rotary rot(2);
+    EXPECT_EQ(rot.filter(0.25), 0.125);
+    EXPECT_EQ(rot.filter(0.0), 0.125);
+    EXPECT_EQ(rot.filter(0.0), 0.0);
```
fwcd, line 24: Small note: Yes, we do precise comparisons on floats here, which is generally discouraged, but powers of two are well-representable in IEEE 754, so we should be fine.

Swiftb0y: or just use `EXPECT_DOUBLE_EQ` just to be sure. The ringbuffer is not large enough to suffer from large inaccuracies.

Swiftb0y (commented):
Thank you. I was just about to do this as well.

Swiftb0y (approved):

PR #14108, opened 2025-01-02
ControllerScriptInterfaceLegacy: Windows 1252 fix

daschuer:
Latin1 is not Windows Western Europe charset. The later contains ™ and some other characters. This PR introduces both, "windows-1252" and "ISO-8859-1"

In addition this has some refactoring regarding string handling and makes the test pass on Qt 6.2 (I have not updated my dev machine to maintain the 2.5 branch)

### DIFF
--- a/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
@@ -1075,7 +1075,7 @@ QByteArray ControllerScriptInterfaceLegacy::convertCharset(
         return convertCharsetInternal(QLatin1String("windows-1250"), value);
     case Cyrillic:
         return convertCharsetInternal(QLatin1String("windows-1251"), value);
-    case Latin1:
+    case WesternEurope:
         return convertCharsetInternal(QLatin1String("windows-1252"), value);
     case Greek:
         return convertCharsetInternal(QLatin1String("windows-1253"), value);
@@ -1109,7 +1109,10 @@ QByteArray ControllerScriptInterfaceLegacy::convertCharset(
         return convertCharsetInternal(QLatin1String("BOCU-1"), value);
     case CESU_8:
         return convertCharsetInternal(QLatin1String("CESU-8"), value);
+    case Latin1:
+        return convertCharsetInternal(QLatin1String("ISO-8859-1"), value);
     }
+
     m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unknown charset specified"));
     return QByteArray();
 }
--- a/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
@@ -1060,67 +1060,66 @@ QByteArray ControllerScriptInterfaceLegacy::convertCharset(
     using enum Charset;
     switch (targetCharset) {
     case ASCII:
-        return convertCharsetInternal(QStringLiteral("US-ASCII"), value);
+        return convertCharsetInternal(QLatin1String("US-ASCII"), value);
     case UTF_8:
-        return convertCharsetInternal(QStringLiteral("UTF-8"), value);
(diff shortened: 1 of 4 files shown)


JoergAtGithub:
This was intentional, we grouped all charsets that are full subsets of other charsets to the superset. This is described in the documentation, which you didn't change in this PR: https://github.com/mixxxdj/mixxx/blob/436aa3cf4566de12adc8c41555054da0a6464847/res/controllers/engine-api.d.ts#L317

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -1060,67 +1060,66 @@ QByteArray ControllerScriptInterfaceLegacy::convertCharset(
...
-        return convertCharsetInternal(QStringLiteral("CESU-8"), value);
+        return convertCharsetInternal(QLatin1String("CESU-8"), value);
     }
     m_pScriptEngineLegacy->logOrThrowError(QStringLiteral("Unknown charset specified"));
     return QByteArray();
 }
 
 QByteArray ControllerScriptInterfaceLegacy::convertCharsetInternal(
-        const QString& targetCharset, const QString& value) {
+        const QLatin1String& targetCharset, const QString& value) {
```
Swiftb0y, line 1121: This is supposed to be called from Javascript where `String` is implemented directly in terms of `QString` (which I'm quite certain about). So you're just forcing the Qt runtime to always do the conversion. The conversion is also already avoided in the >=Qt6.8 path. 
Moreover, there is no reason to take a reference type like `QLatin1String` by const ref. So even if we were to introduce it here, we should take it by value.

Swiftb0y (changes requested):

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: I propose to drop this commit entirely.

daschuer: This is the "private" function convertCharsetInternal(), so we can benefit form no conversion in the < Qt6.8 case as well.

daschuer:
> we grouped all charsets that are full subsets of other charsets to the superset

Ah, I understand. Than the Enum name was only misleading. I stumbled over this because "ASCII" and ""windows-1252" is not available without ICU or QT < 6.4.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: Ah I see. You're correct.

daschuer:
Done

JoergAtGithub (dismissed):
If you change the enum, you need to change the typescript declaration as well: [mixxx/res/controllers/engine-api.d.ts](https://github.com/mixxxdj/mixxx/blob/436aa3cf4566de12adc8c41555054da0a6464847/res/controllers/engine-api.d.ts#L317)

daschuer:
Done

daschuer:
@Swiftb0y Can we merge this now?

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -763,6 +812,7 @@ TEST_F(ControllerScriptEngineLegacyTest, convertCharsetAllCharset) {
                 "'" COMPLICATEDSTRINGLITERAL "')")
                                  .arg(key);
         auto result = qjsvalue_cast<QByteArray>(evaluate(source));
+        qDebug() << result;
         EXPECT_EQ(result.size(), convertedCharsetForString(enumValue))
```
Swiftb0y, line 816: debugging leftover?

Swiftb0y (approved):
LGTM. Otherwise

daschuer:
Done and sqashed fixup. Reday for merge.

Swiftb0y (approved):
LGTM. Thank you.

PR #14112, opened 2025-01-03
Num deck streamline

daschuer:
This mainly fixes a lifetime issue with the non owning 
```
static QAtomicPointer<ControlProxy> PlayerManager::m_pCOPNumDecks;
```

Since it is non owning it can become dangling. I have fixed it by moving it as a PollingControlProxy to the singleton PlayerInfo class


src/test/playermanagertest.cpp
```diff
@@ -129,7 +129,7 @@ class PlayerManagerTest : public MixxxDbTest, SoundSourceProviderRegistration {
 
 TEST_F(PlayerManagerTest, UnEjectTest) {
     // Ejecting an empty deck with no previously-recorded ejected track has no effect.
-    auto deck1 = m_pPlayerManager->getDeck(1);
+    auto deck1 = m_pPlayerManager->getDeck(0);
```
Swiftb0y, line 132: our `[ChannelN]` COs are 1-based while this is now 0-based. Is that really an improvement? I'm concerned this will cause confusion and off-by-one errors later.

daschuer: You point out the downside of the change.

The benefit is that it now matches `PlayerManager::groupForDeck(i)` and pevents cases where we have to + 1 before the call and - 1 after it. 

Like here: (I will also place an inline comment) 
```
for (int i = 0; i < m_pPlayerManager->numberOfSamplers(); ++i) {
```

In addition there was a naming clash with Index vs. Number. 

In general we should stick to the rule to use 0 based counting inside the code base and 1 base in all user visible cases. 

Please have a second look. If you still think we should keep the 1 based index here I will fix the issue in that direction.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -155,9 +155,8 @@ AutoDJProcessor::AutoDJProcessor(
 
     // TODO(rryan) listen to signals from PlayerManager and add/remove as decks
     // are created.
-    for (unsigned int i = 0; i < pPlayerManager->numberOfDecks(); ++i) {
-        QString group = PlayerManager::groupForDeck(i);
-        BaseTrackPlayer* pPlayer = pPlayerManager->getPlayer(group);
+    for (int i = 0; i < pPlayerManager->numberOfDecks(); ++i) {
+        BaseTrackPlayer* pPlayer = pPlayerManager->getDeckBase(i);
```
daschuer, line 159: Here we benefit form the 0 base.

src/mixer/samplerbank.cpp
```diff
@@ -107,8 +107,8 @@ bool SamplerBank::saveSamplerBankToPath(const QString& samplerBankPath) {
     QDomElement root = doc.createElement(QStringLiteral("samplerbank"));
     doc.appendChild(root);
 
-    for (unsigned int i = 0; i < m_pPlayerManager->numSamplers(); ++i) {
-        Sampler* pSampler = m_pPlayerManager->getSampler(i + 1);
+    for (int i = 0; i < m_pPlayerManager->numberOfSamplers(); ++i) {
+        Sampler* pSampler = m_pPlayerManager->getSampler(i);
```
daschuer, line 111: Here we benefit form the 0 based index

src/test/playermanagertest.cpp
Swiftb0y: Thank you for the explanation. I see the tension there. I too prefer 0-based indexing. I will take a second look.

src/coreservices.cpp
```diff
@@ -441,7 +441,9 @@ void CoreServices::initialize(QApplication* pApp) {
     // Load tracks in args.qlMusicFiles (command line arguments) into player
     // 1 and 2:
     const QList<QString>& musicFiles = m_cmdlineArgs.getMusicFiles();
-    for (int i = 0; i < (int)m_pPlayerManager->numDecks() && i < musicFiles.count(); ++i) {
+    for (int i = 0; i < m_pPlayerManager->numberOfDecks() &&
+            i < musicFiles.count();
```
Swiftb0y, line 445: NIT
```suggestion
    const int numTracks = std::min(m_pPlayerManager->numberOfDecks(), musicFiles.count());
    for (int i = 0; i < numTracks
```

src/mixer/playermanager.h
```diff
@@ -36,21 +36,20 @@ class PlayerManagerInterface {
     virtual BaseTrackPlayer* getPlayer(const QString& group) const = 0;
     virtual BaseTrackPlayer* getPlayer(const ChannelHandle& channelHandle) const = 0;
 
-    // Get the deck by its deck number. Decks are numbered starting with 1.
-    virtual Deck* getDeck(unsigned int player) const = 0;
+    // Get the deck by its deck index
+    virtual BaseTrackPlayer* getDeckBase(int deckIndex) const = 0;
```
Swiftb0y, line 40: Why this replacement?

daschuer: This interface allows the unittest to return a mocked Deck instead of the real one. I will add a comment. Before `getPlayer()` has been used, which requires a QString group for lookup. I will add a comment.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -155,9 +155,8 @@ AutoDJProcessor::AutoDJProcessor(
 
     // TODO(rryan) listen to signals from PlayerManager and add/remove as decks
```
acolombier, line 156: I know it's not related to your change, but anything we can do about this 10yo+ TODO?

src/coreservices.cpp
```diff
@@ -441,7 +441,9 @@ void CoreServices::initialize(QApplication* pApp) {
     // Load tracks in args.qlMusicFiles (command line arguments) into player
     // 1 and 2:
     const QList<QString>& musicFiles = m_cmdlineArgs.getMusicFiles();
-    for (int i = 0; i < (int)m_pPlayerManager->numDecks() && i < musicFiles.count(); ++i) {
+    const int numTracks = std::min(m_pPlayerManager->numberOfDecks(),
+            static_cast<int>(musicFiles.count()));
+    for (int i = 0; i < numTracks; ++i) {
```
acolombier, line 446: What's the benefit of preincr vs postincr? Is this preferred?

src/mixer/playerinfo.cpp
```diff
@@ -14,10 +14,16 @@ constexpr int kPlayingDeckUpdateIntervalMillis = 2000;
 
 PlayerInfo* s_pPlayerInfo = nullptr;
 
+const QString kAppGroup = QStringLiteral("[App]");
+const QString kMasterGroup = QStringLiteral("[Master]");
```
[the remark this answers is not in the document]
daschuer: That's nothing we can do in a stable branch. We need to introduce an alias. I don't want to do it in this PR. Let's keep it separate.

src/coreservices.cpp
daschuer: `i++` was the original one preferred in C and the origin of the name for C++. In C++ itself `++i` is preferred if possible, because it returns the incremented value and not the original one which is considered to be faster in case of complex types, because the original one does not need to be kept in memory. In this particular case the optimized machine code will most likely be the same anyway because the return value is not used.

daschuer:
This should be now ready

src/coreservices.cpp
acolombier: Thanks for the detailed explanation. I knew about the usecase where using the value, but I was wondering if there is reason why you would use this form when the value is irrelevant. Is that something we should [unify](https://github.com/mixxxdj/mixxx/wiki/Coding-Guidelines)? I know we have a few for-loop which use the two form, beside no apparent reason for switching between the two forms

daschuer: I don't really mind. Both is correct and readable in this case IMHO.

src/library/autodj/autodjprocessor.cpp
```diff
@@ -195,6 +188,17 @@ AutoDJProcessor::~AutoDJProcessor() {
     delete m_pAutoDJTableModel;
 }
 
+void AutoDJProcessor::slotNumberOfDecksChanged(int decks) {
+    for (int i = m_decks.size(); i < decks; ++i) {
+        BaseTrackPlayer* pPlayer = m_pPlayerManager->getDeckBase(i);
+        // Shouldn't be possible.
+        VERIFY_OR_DEBUG_ASSERT(pPlayer) {
+            continue;
```
acolombier: You should insert a null entry in `m_decks`, otherwise you may have some offset issues in `PlayerManager::getDeck`

src/library/autodj/autodjprocessor.cpp
```diff
@@ -195,6 +188,17 @@ AutoDJProcessor::~AutoDJProcessor() {
...
+        }
+        m_decks.append(new DeckAttributes(i, pPlayer));
```
acolombier: Could you use a `unique_ptr` here?

daschuer:
I went ahead and removed all plain new()

src/library/autodj/autodjprocessor.cpp
```diff
@@ -173,44 +160,30 @@ AutoDJProcessor::AutoDJProcessor(
...
-    delete m_pShufflePlaylist;
-    delete m_pEnabledAutoDJ;
-    delete m_pFadeNow;
-
-    delete m_pAutoDJTableModel;
-}
-
 void AutoDJProcessor::slotNumberOfDecksChanged(int decks) {
+    m_decks.reserve(decks);
     for (int i = m_decks.size(); i < decks; ++i) {
```
acolombier: nit
```suggestion
    for (size_t i = m_decks.size(); i < decks; ++i) {
```

src/library/autodj/autodjprocessor.cpp
```diff
@@ -180,33 +160,30 @@ AutoDJProcessor::AutoDJProcessor(
...
-    delete m_pFadeNow;
-
-    delete m_pAutoDJTableModel;
+void AutoDJProcessor::slotNumberOfDecksChanged(int decks) {
+    m_decks.reserve(decks);
+    for (int i = m_decks.size(); i < decks; ++i) {
+        BaseTrackPlayer* pPlayer = m_pPlayerManager->getDeckBase(i);
+        // Shouldn't be possible.
+        VERIFY_OR_DEBUG_ASSERT(pPlayer) {
+            return;
```
acolombier, line 171: This will `m_decks` in an incomplete state. How do you feel about inserting null values?

src/library/autodj/autodjprocessor.cpp
```diff
@@ -180,33 +160,30 @@ AutoDJProcessor::AutoDJProcessor(
...
-
-    delete m_pSkipNext;
-    delete m_pAddRandomTrack;
-    delete m_pShufflePlaylist;
-    delete m_pEnabledAutoDJ;
-    delete m_pFadeNow;
-
-    delete m_pAutoDJTableModel;
+void AutoDJProcessor::slotNumberOfDecksChanged(int decks) {
+    m_decks.reserve(decks);
```
acolombier, line 164: If I understand correctly, this slots could be called every time the number of decks changes, so you will needs to either reset the `m_decks` vector, or only add (or remove) the missing decks based on the new requested number.

Or is this slot not used like so?

acolombier (commented):
Some more minor comments

src/library/autodj/autodjprocessor.cpp
daschuer: If this happens Mixxx is broken anyway. A Null value likely causes crashes somewhere else.

src/library/autodj/autodjprocessor.cpp
daschuer: Mixxx does not support deleting decks. I will dorp a comment.

daschuer:
Done

daschuer:
@acolombier Can you have a final look here? Thank you.

daschuer:
Friendly ping @acolombier

src/library/autodj/autodjprocessor.cpp
acolombier: Could we add a verify/assert there in case the deck count is being reduced? The CO can be decreased without any warnings or preventions (tested on Mixxx for Android as I am AFK)

acolombier:
I am AFK, but I will try to have a look when I'm back at my desk

daschuer:
@acolombier, this is a friendly ping.

acolombier:
Sorry for letting this drag...

acolombier (approved):
There are some changes I am not entirely sure why we need them, but this overall seems to well improve the state of the code, and remove many fat pointers, so let's get that in.

daschuer:
Thank you.

acolombier:
Just realised there was a fixup commit (9c24d1a64c1959a4bc24bbe43265bf2953811086) but somehow it wasn't caught by the CI... too bad!

PR #14114, opened 2025-01-03
(fix) Tracks: copy fix Ctrl+C

### LINKED ISSUE #14065: ctrl-c / copy in libary doesn't copy the text of the selected field
### Bug Description

In previous versions of mixxx, when selecting a field in the library (not editing it) was copy the selected field text.
No it copies the text to the "path to the file" in every selected column. 
library2

Workaround: It can now only be copied, when editing the field.

### Version

2.5.0

### OS

ubuntu 24.04

ronso0:
fixes #14065 
* **Ctrl+Shift+C** copies the content of the selected cell(s)
* (Ctrl+C copies the selected track(s) as before)

small fixup:
allow Copy Track also in locked track models

### DIFF
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -1151,9 +1151,14 @@ void WTrackTableView::keyPressEvent(QKeyEvent* event) {
             moveSelectedTracks(event);
             return;
         }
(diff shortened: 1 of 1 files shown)


daschuer (approved):
Works like a charm and LGTM

PR #14123, opened 2025-01-06
(fix) Traktor Kontrol S3: use pitch absolute mode as described in the manual

### LINKED ISSUE #13927: Traktor Kontrol S3 mapping issue in v2.4.1 (PitchSlider = FALSE)
### Bug Description

> Great! Thanks for testing! I guess this is because the tempo-sliders are mapped to work relative and not absolute - can be configured at the top of the mapping file btw.:
> 
> https://github.com/mixxxdj/mixxx/blob/d7a769ba4a0753861a3cadab1c592ca953326b60/res/controllers/Traktor-Kontrol-S3-hid-scripts.js#L54-L67

Hello, I just realized that this modification (the TRUE command must be set to FALSE) of the config file "Traktor-Kontrol-S3-hid-scripts.js" of the Native Instrument Traktor S3 controller has not been taken into account and integrated directly into the latest version v2.4.1 of Mixxx. Best regards.

### Version

2.4.1

### OS

Windows 10, Macos BigSur, Linux Mint 22

ronso0:
Fixes #13927 
See https://manual.mixxx.org/2.5/en/hardware/controllers/native_instruments_traktor_kontrol_s3#rate-keylock

### DIFF
--- a/res/controllers/Traktor-Kontrol-S3-hid-scripts.js
+++ b/res/controllers/Traktor-Kontrol-S3-hid-scripts.js
@@ -64,7 +64,7 @@ TraktorS3.QuickEffectModeChannelColors = false;
 // * Hold shift to move the pitch slider without adjusting the rate
 // * Hold keylock and move the pitch slider to adjust musical pitch
 // * keylock will still toggle on, but on release, not press.
-TraktorS3.PitchSliderRelativeMode = true;
+TraktorS3.PitchSliderRelativeMode = false;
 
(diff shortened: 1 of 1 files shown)


ronso0:
@zebebop Please check if that works as desired now.
Thanks!

acolombier (approved):
LGTM!

Change looks simple enough to skip testing so I'm merging and we can fix forward in case it doesn't have the expected impact.

PR #1713, opened 2018-06-16
borrowable_ptr, a threadsave callback solution

daschuer:
A wrapper class to protect pointers owned by another entity, allowing
safe access to the referenced object across threads. This prevents the
object from being deleted by its owner while other threads are executing
functions on the referenced object, implemented mostly lock-free, except during
reset(), if the object is still in use.

Use cases are:
 - Implementing callbacks to objects with shorter lifetimes.
 - A safe replacement for Qt's direct connections, which must not be used across
   threads in such cases.

Unlike `std::shared_ptr`, the object is not deleted when the last
`borrowed_ptr` falls out of scope. Ownership and deletion are always managed
by the original owner (via `borrowable_ptr`).

If the `borrowable_ptr` falls out of scope while there are active `borrowed_ptr`
instances, the thread is suspended until all `borrowed_ptr` instances are deleted
and the reference count reaches zero.

Thread safety:
 - It is thread-safe to replace the object in `borrowable_ptr`.
 - The transition is not atomic: `borrowed_ptr` instances become null before pointing
   to the new object.

Usage:
```
	int i = 5;
	{
	    // Make i borrowable, owned by the stack
            auto borrowable = borrowable_ptr(&i);
	    {
	        // Borrow i, this can be borrowd across threads
                // borrowed_ptr is a strong reference that guranties the borrowed object is
                // valid unit the end of the scope
                borrowed_ptr borrowed1 = borrowable.borrow();
	        borrowed_ptr borrowed2 = borrowable.borrow();
	    } // borrowed objects are returned
	} // borrowable_ptr falls out of scope, suspended if a borrowed pointer is not returned
```

PR #14124, opened 2025-01-06
Preferences: show 'real' xfader configuration

ronso0:
This makes the Mixer preferences load the xfader configuration
* from config on first load (if all xfader controls are still at their default values¹)
* from controls each following update

This means touching the usual 3-state switch or curve knob on a controller will update the xfader graphic immedtiately.

¹Since mappings are loaded before DlgPreferences is instantiated this is supposed to detect if mappings have already set the xfader config in their init(), either explicitly or received updates via inputReport request.

Will do another test soon with the S4 Mk3 mapping which requests a status report on startup to sync all controls.

If someone wants to test it without controller, here's a mapping for Vir-MIDI or MIDI Through
https://gist.github.com/ronso0/ae7a6f3fc2b28e43f16bf50bfb84e62e

### DIFF
--- a/src/preferences/dialog/dlgprefmixer.cpp
+++ b/src/preferences/dialog/dlgprefmixer.cpp
@@ -41,14 +41,12 @@ const ConfigKey kLowEqFreqKey = ConfigKey(kMixerProfile, kLowEqFrequency);
 const ConfigKey kLowEqFreqPreciseKey =
         ConfigKey(kMixerProfile, QStringLiteral("LoEQFrequencyPrecise"));
 
-const QString kXfaderCalibration = QStringLiteral("xFaderCalibration");
-
 const ConfigKey kXfaderModeKey = ConfigKey(EngineXfader::kXfaderConfigKey,
         QStringLiteral("xFaderMode"));
 const ConfigKey kXfaderCurveKey = ConfigKey(EngineXfader::kXfaderConfigKey,
         QStringLiteral("xFaderCurve"));
 const ConfigKey kXfaderCalibrationKey = ConfigKey(EngineXfader::kXfaderConfigKey,
-        kXfaderCalibration);
+        QStringLiteral("xFaderCalibration"));
 const ConfigKey kXfaderReverseKey = ConfigKey(EngineXfader::kXfaderConfigKey,
         QStringLiteral("xFaderReverse"));
 
@@ -74,13 +72,12 @@ DlgPrefMixer::DlgPrefMixer(
         : DlgPreferencePage(pParent),
           m_pConfig(pConfig),
           m_xFaderMode(MIXXX_XFADER_ADDITIVE),
-          m_xFaderTransform(EngineXfader::kTransformDefault),
+          m_xFaderCurve(EngineXfader::kTransformDefault),
           m_xFaderCal(0.0),
           m_xfModeCO(make_parented<ControlProxy>(kXfaderModeKey, this)),
           m_xfCurveCO(make_parented<ControlProxy>(kXfaderCurveKey, this)),
           m_xfReverseCO(make_parented<ControlProxy>(kXfaderReverseKey, this)),
-          m_xfCalibrationCO(
-                  ConfigKey(EngineXfader::kXfaderConfigKey, kXfaderCalibration)),
+          m_xfCalibrationCO(make_parented<ControlProxy>(kXfaderCalibrationKey, this)),
           m_crossfader(QStringLiteral("[Master]"), QStringLiteral("crossfader")),
           m_xFaderReverse(false),
           m_COLoFreq(kLowEqFreqKey),
@@ -113,14 +110,30 @@ DlgPrefMixer::DlgPrefMixer(
     connect(SliderXFader,
             QOverload<int>::of(&QSlider::valueChanged),
             this,
-            &DlgPrefMixer::slotUpdateXFader);
-    connect(SliderXFader, &QSlider::sliderMoved, this, &DlgPrefMixer::slotUpdateXFader);
-    connect(SliderXFader, &QSlider::sliderReleased, this, &DlgPrefMixer::slotUpdateXFader);
-    connect(radioButtonAdditive, &QRadioButton::clicked, this, &DlgPrefMixer::slotUpdateXFader);
-    connect(radioButtonConstantPower,
-            &QRadioButton::clicked,
+            &DlgPrefMixer::slotXFaderSliderChanged);
+    connect(SliderXFader, &QSlider::sliderMoved, this, &DlgPrefMixer::slotXFaderSliderChanged);
+    connect(SliderXFader, &QSlider::sliderReleased, this, &DlgPrefMixer::slotXFaderSliderChanged);
+    connect(buttonGroupCrossfaderModes,
+            &QButtonGroup::buttonClicked,
             this,
-            &DlgPrefMixer::slotUpdateXFader);
+            &DlgPrefMixer::slotXFaderModeBoxToggled);
+    connect(checkBoxReverse,
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+            &QCheckBox::checkStateChanged,
+#else
+            &QCheckBox::stateChanged,
+#endif
+            this,
+            &DlgPrefMixer::slotXFaderReverseBoxToggled);
+
+    m_xfModeCO->connectValueChanged(
+            this, &DlgPrefMixer::slotXFaderModeControlChanged);
+    m_xfCurveCO->connectValueChanged(
+            this, &DlgPrefMixer::slotXFaderCurveControlChanged);
+    m_xfCalibrationCO->connectValueChanged(
+            this, &DlgPrefMixer::slotXFaderCalibrationControlChanged);
+    m_xfReverseCO->connectValueChanged(
+            this, &DlgPrefMixer::slotXFaderReverseControlChanged);
 
     // Don't allow the xfader graph getting keyboard focus
     graphicsViewXfader->setFocusPolicy(Qt::NoFocus);
@@ -135,33 +148,57 @@ DlgPrefMixer::DlgPrefMixer(
     connect(SliderLoEQ, &QSlider::sliderReleased, this, &DlgPrefMixer::slotLoEqSliderChanged);
 
     connect(CheckBoxEqAutoReset,
-            &QCheckBox::toggled,
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+            &QCheckBox::checkStateChanged,
+#else
+            &QCheckBox::stateChanged,
+#endif
             this,
             &DlgPrefMixer::slotEqAutoResetToggled);
     connect(CheckBoxGainAutoReset,
-            &QCheckBox::toggled,
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+            &QCheckBox::checkStateChanged,
+#else
+            &QCheckBox::stateChanged,
+#endif
             this,
             &DlgPrefMixer::slotGainAutoResetToggled);
 #ifdef __STEM__
     connect(CheckBoxStemAutoReset,
-            &QCheckBox::toggled,
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+            &QCheckBox::checkStateChanged,
+#else
+            &QCheckBox::stateChanged,
+#endif
             this,
             &DlgPrefMixer::slotStemAutoResetToggled);
 #else
     CheckBoxStemAutoReset->hide();
 #endif
     connect(CheckBoxBypass,
-            &QCheckBox::toggled,
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+            &QCheckBox::checkStateChanged,
+#else
+            &QCheckBox::stateChanged,
+#endif
             this,
             &DlgPrefMixer::slotBypassEqToggled);
 
     connect(CheckBoxEqOnly,
-            &QCheckBox::toggled,
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+            &QCheckBox::checkStateChanged,
+#else
+            &QCheckBox::stateChanged,
+#endif
             this,
             &DlgPrefMixer::slotEqOnlyToggled);
 
     connect(CheckBoxSingleEqEffect,
-            &QCheckBox::toggled,
+#if QT_VERSION >= QT_VERSION_CHECK(6, 7, 0)
+            &QCheckBox::checkStateChanged,
+#else
+            &QCheckBox::stateChanged,
+#endif
             this,
             &DlgPrefMixer::slotSingleEqToggled);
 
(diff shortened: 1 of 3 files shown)


Swiftb0y (changes requested):

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -41,14 +42,14 @@ const ConfigKey kLowEqFreqKey = ConfigKey(kMixerProfile, kLowEqFrequency);
...
-const ConfigKey kXfaderReverseKey = ConfigKey(EngineXfader::kXfaderConfigKey,
-        QStringLiteral("xFaderReverse"));
+const ConfigKey kXfaderModeKey = ConfigKey(EngineXfader::kXfaderGroup,
+        EngineXfader::kXfaderModeKey);
+const ConfigKey kXfaderCurveKey = ConfigKey(EngineXfader::kXfaderGroup,
+        EngineXfader::kXfaderCurveKey);
+const ConfigKey kXfaderCalibrationKey = ConfigKey(EngineXfader::kXfaderGroup,
+        EngineXfader::kXfaderCalibrationKey);
+const ConfigKey kXfaderReverseKey = ConfigKey(EngineXfader::kXfaderGroup,
+        EngineXfader::kXfaderReverseKey);
```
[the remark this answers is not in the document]
ronso0: Hmm, so IIUC this is not specific to the diff here but also applies to the state before?
Like `EngineXfader::kXfaderConfigKey` being defined and initialized in the namespace here?
And moving those to the initializer list would avoid it, right?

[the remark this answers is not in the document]
ronso0: Or create a few `getXYZCfgKey()` helpers in dlgprefmixer.h?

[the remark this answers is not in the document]
ronso0: Or simply define them in enginexfader.h like this:
`static const QString kXfaderGroup = QStringLiteral("[Mixer Profile]");`
?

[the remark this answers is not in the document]
Swiftb0y: The problem is not that its defined in the namespace (please do and keep that), the problem is that you expose the symbol via the header and then other translation units initialize their static data based on the static data of another translation unit. The solution in both cases is either some "construct on first use" idiom (there are multiple ways to achieve that in C++) or just don't depend on static data from other translation units. So duplicate the QString across both translation units. Its ugly but thats what we get for using `QString`s for everything...

src/engine/enginexfader.h
```diff
@@ -17,7 +19,11 @@ class EngineXfader {
             CSAMPLE_GAIN* gain1,
             CSAMPLE_GAIN* gain2);
 
-    static const char* kXfaderConfigKey;
+    static const QString kXfaderGroup;
+    static const QString kXfaderModeKey;
+    static const QString kXfaderCurveKey;
+    static const QString kXfaderCalibrationKey;
+    static const QString kXfaderReverseKey;
```
Swiftb0y, line 26: These are the offenders. If you can come up with a solution that removes this static data, you should be good. Again, I recommend you simply duplicate the required `QString`s/`ConfigKey`s across both files.

ronso0: Thanks for the explanation, though I'm at a loss: I read about Construct on first Use but explanations and tutorials only refer to obejcts/pointers and classes with a constructor, which is not the case for EngineXfader.
I'll revert the commit and stick with the duplicate QStrings.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -110,14 +114,40 @@ DlgPrefMixer::DlgPrefMixer(
...
     connect(radioButtonConstantPower,
             &QRadioButton::clicked,
             this,
-            &DlgPrefMixer::slotUpdateXFader);
+            &DlgPrefMixer::slotXFaderWidgetsChanged);
+    connect(checkBoxReverse,
+            // TODO deprecated after Qt 6.9, replace with checkStateChanged(Qt::CheckState)
+            &QCheckBox::stateChanged,
+            this,
+            &DlgPrefMixer::slotXFaderReverseBoxToggled);
```
Swiftb0y, line 132: lets add a `QT_VERSION_CHECK` for this now.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -110,14 +114,40 @@ DlgPrefMixer::DlgPrefMixer(
...
+    connect(&m_xfaderControlUpdateTimer,
+            &QTimer::timeout,
+            this,
+            [this]() {
+                slotUpdateXFaderFromConfigOrControls();
+                // store new settings in config
+                applyXFader();
+            });
+    m_xfaderControlUpdateTimer.setSingleShot(true);
+    m_xfaderControlUpdateTimer.setInterval(200);
```
Swiftb0y, line 150: I doubt this is a super good solution. Do we actually get truly concurrent calls (in different threads) or just rapid calls in quick succession? Which part is actually so slow that it warrants the debouncing?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -717,6 +734,22 @@ void DlgPrefMixer::slotApply() {
...
+    m_xfCalibrationCO.set(m_xFaderCal);
+    if (m_xFaderReverse != (m_xfReverseCO->get() > 0)) {
+        double position = m_crossfader.get();
+        m_crossfader.set(0.0 - position);
+    }
+    m_xfReverseCO->set(m_xFaderReverse ? 1.0 : 0.0);
+
+    m_pConfig->set(kXfaderModeKey, ConfigValue(m_xFaderMode));
+    m_pConfig->set(kXfaderCurveKey, ConfigValue(QString::number(m_xFaderTransform)));
+    m_pConfig->set(kXfaderReverseKey, ConfigValue(m_xFaderReverse ? 1 : 0));
```
Swiftb0y, line 750: NIT:
```suggestion
    m_pConfig->setValue(kXfaderModeKey, m_xFaderMode);
    m_pConfig->setValue(kXfaderCurveKey, m_xFaderTransform));
    m_pConfig->setValue(kXfaderReverseKey, m_xFaderReverse);
```

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -72,13 +75,14 @@ DlgPrefMixer::DlgPrefMixer(
...
           m_xFaderMode(MIXXX_XFADER_ADDITIVE),
-          m_transform(EngineXfader::kTransformDefault),
-          m_cal(0.0),
-          m_mode(kXfaderModeKey),
-          m_curve(kXfaderCurveKey),
-          m_calibration(kXfaderCalibrationKey),
-          m_reverse(kXfaderReverseKey),
-          m_crossfader("[Master]", "crossfader"),
+          m_xFaderTransform(EngineXfader::kTransformDefault),
+          m_xFaderCal(0.0),
```
Swiftb0y, line 79: proposal: Create a `struct CrossfaderParameter` in the class to bundle the calibration, transform, reverse and mode together instead of prefixing all of those with `xFader`?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -815,8 +826,46 @@ void DlgPrefMixer::slotUpdate() {
...
+    // (or skin attributes) have changed the xfader controls.
+    // Else and on later calls, always read the current state from controls.
+    if (m_initializing &&
+            m_xfCurveCO->get() == m_xfCurveCO->getDefault() &&
+            m_xfCalibrationCO.get() == m_xfCalibrationCO.getDefault() &&
+            m_xfModeCO->get() == m_xfModeCO->getDefault() &&
+            m_xfReverseCO->get() == m_xfReverseCO->getDefault()) {
+        m_xFaderTransform = m_pConfig->getValue(kXfaderCurveKey, EngineXfader::kTransformDefault);
+        m_xFaderMode = m_pConfig->getValueString(kXfaderModeKey).toInt();
+        m_xFaderReverse = m_pConfig->getValueString(kXfaderReverseKey).toInt() == 1;
```
Swiftb0y, line 841:
```suggestion
        m_xFaderMode = m_pConfig->getValue<int>(kXfaderModeKey);
        m_xFaderReverse = m_pConfig->getValue<bool>(kXfaderReverseKey);
```

src/preferences/dialog/dlgprefmixer.cpp
Swiftb0y: If you know what attributes exactly influence the graphic, you can compare the old and new state and reject calls that result in the same or similar values. That seems like a reasonable way to do this without having to introduce a hacky timer.

Swiftb0y (commented):
couple thoughts

src/preferences/dialog/dlgprefmixer.cpp
ronso0: Yeah, "concurrent" is the wrong term, it's indeed rapid calls in quick succession.
I saw that I switched the 3-state xfader toggle on the S4mk3 (0 -> 1) and it caused some update misses / reverted the mode state.
But I think is might have been happening in an earlier stage only.

I'll need to rework this further anyway, like more separation of updates from control changes / widget changes. (maybe some blockSignals() will already fix it)

ronso0:
Sorry what? Why is CI failing on that one now?
> Error: /home/runner/work/mixxx/mixxx/src/test/borrowabletest.cpp:4:10: error: Module QtConcurrent should not be included directly [-Wclazy-no-module-include]
    4 | #include <QtConcurrent>

Swiftb0y:
not your fault. ignore clazy for now. https://github.com/mixxxdj/mixxx/pull/1713#issuecomment-2639295718

src/preferences/dialog/dlgprefmixer.cpp
ronso0: You mean just for holding the parameters in the preferences?
Sure, we can do that. But tbh I'd like to get the mechanics right first before we start optimizing it.

ronso0:
Alright, all works as desired now!

Swiftb0y:
great. what next steps would you prefer? Remove the debugging statements, address my previous comments, etcc?

ronso0:
Testing! :smile:

ronso0:
then with :+1: remove the debug stuff

ronso0:
all comments addressed, except the struct

ronso0:
> Alright, all works as desired now!

whoops, not quite, widget update after controls change was missing.

ronso0:
This has proven to be very handy recently:
already late into a session when I wanted to cut, I noticed that the S4's xfader switch must have been flipped when I've put it into its case, and I could quickly adjust and very the desired curve.
(some sort of popup would have been super nice in this case, so i don't have to open the preferences and navigate to Mixer anymore :thinking: bit like the key wheel window with a auto-hide timer)

Swiftb0y:
I agree. Do you intend to finish this then? There are still open review comments and the code is also  still full of debug printing.

src/preferences/dialog/dlgprefmixer.cpp
ronso0: Done, also for the other checkboxes.

src/preferences/dialog/dlgprefmixer.cpp
ronso0: Yup, let's postpone it.
Wanna file a request for it?

ronso0:
I think I addressed all comments.

I'll squash and remove the debug commit once this has received some testing.

src/preferences/dialog/dlgprefmixer.cpp
Swiftb0y: nah, its no need to file an issue over code improvement. someone else will clean it up when they look at it again.

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -908,28 +1007,91 @@ void DlgPrefMixer::drawXfaderDisplay() {
...
-    m_cal = EngineXfader::getPowerCalibration(m_transform);
+    m_xFaderCurve = std::round(curve * 10000) / 10000;
+    // If the curve has been changed in the GUI we fetch the engine value for
+    // calibration which gives us a smooth curve.
+    // This wipes any previous value set by controller mappings for example.
+    m_xFaderCal = EngineXfader::getPowerCalibration(m_xFaderCurve);
+    m_xFaderMode = radioButtonConstantPower->isChecked()
+            ? MIXXX_XFADER_CONSTPWR
+            : MIXXX_XFADER_ADDITIVE;
+
```
Swiftb0y, line 1038: does this need to be in the slider slot?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -815,9 +836,87 @@ void DlgPrefMixer::slotUpdate() {
...
+        m_xFaderCurve = m_pConfig->getValue(kXfaderCurveKey, EngineXfader::kTransformDefault);
+        // "xFaderCalibration" is not stored in the config and it's not expsoed
+        // with a slider here. Each time the slider is touched it's calculated
+        // to get us a smooth curve for ConstPower mode. And hos no effect for
+        // Additive mode.
+        // TODO This also means custom values set by controller mappings are
+        // wiped on shutdown.
+        m_xFaderCal = EngineXfader::getPowerCalibration(m_xFaderCurve);
+        m_xFaderMode = m_pConfig->getValue<int>(kXfaderModeKey);
+        m_xFaderReverse = m_pConfig->getValue<int>(kXfaderReverseKey) == 1;
```
Swiftb0y, line 862:
```suggestion
        m_xFaderReverse = m_pConfig->getValue<bool>(kXfaderReverseKey);
```
or is there a reason otherwise?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -717,6 +744,22 @@ void DlgPrefMixer::slotApply() {
...
 }
 
+void DlgPrefMixer::applyXFader() {
+    m_xfModeCO->set(m_xFaderMode);
+    m_xfCurveCO->set(m_xFaderCurve);
+    m_xfReverseCO->set(m_xFaderReverse ? 1.0 : 0.0);
+    m_xfCalibrationCO->set(m_xFaderCal);
+    if (m_xFaderReverse != (m_xfReverseCO->get() > 0)) {
+        double position = m_crossfader.get();
+        m_crossfader.set(0.0 - position);
```
Swiftb0y, line 754: negative? Should've this been `1.0 - position`?

src/preferences/dialog/dlgprefmixer.cpp
ronso0: No, good catch :+1:

src/preferences/dialog/dlgprefmixer.cpp
ronso0: `[Master],crossfader` range is -1..0..1, so 0 - position mirrors the value.
Or am I missing something?
(didn't change this btw)

But I made a mistake here, the CO is set and than compared to maybe flip the xfader, then set again.
Fixed.

daschuer:
@Swiftb0y Are all review findings solved?

ronso0:
If added the testing mapping I was using for#14361
> If someone wants to test it without controller, here's a mapping for Vir-MIDI or MIDI Through
https://gist.github.com/ronso0/ae7a6f3fc2b28e43f16bf50bfb84e62e

src/preferences/dialog/dlgprefmixer.cpp
Swiftb0y: > range is -1..0..1, so 0 - position mirrors the value.
Or am I missing something?

Ah, no I assumed this set the parameter and thus was in the normalized $[0,1]$ range

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -815,9 +835,87 @@ void DlgPrefMixer::slotUpdate() {
...
+    qWarning().noquote() << "         reverse:    "
+                         << QString(m_xFaderReverse ? "Yes" : "normal");
+
+    updateXFaderWidgets();
+}
+
+void DlgPrefMixer::updateXFaderWidgets() {
+    SliderXFader->blockSignals(true);
+    buttonGroupCrossfaderModes->blockSignals(true);
+    checkBoxReverse->blockSignals(true);
```
Swiftb0y, line 886: lets use a `QSignalBlocker` instead. Otherwise an exception or an inconsiderate future early return could cause the widgets to stay blocked.

Swiftb0y (commented):
please remove the `qWarning`s. I will test ASAP.

Swiftb0y (approved):
LGTM. Thank you.

Swiftb0y (approved):

PR #14130, opened 2025-01-07
Grouped crates replacing crates

Eve00000:
following the suggestions by @acolombier and @fwcd:

This pr is equal to [#13978](https://github.com/mixxxdj/mixxx/pull/13978) with the difference that the 'grouped crates' are not an extra library feature but replacing the cratefeature as requested.

Selecting a group shows all tracks contained in membercrates.

Multilevel when a variable length (mask) is selected and defined.
a4e983ec-1588-4515-a76d-63a742f5bbc0


Eve00000:
While using this I came to the conclusion 'just replacing' crates with 'grouped crates' is not working for me.
There are moments I want to see both crates & grouped crates, to get the naming correct, to have a complete overview.

This PR will be closed, I made an update for "grouped crates" pr #13978 whit a extra selection in preferences 'replace crates with grouped crates'  so that users can select to see only crates / only grouped crates or both (like me).

PR #14136, opened 2025-01-07
[WIP] hide Tracks: make removal from playlists and crates optional

ronso0:
I propose to make hiding from playlists and crates optional.

As described in #14135 the UX is currently inconsistent and not suitable for my use case:
I have many short recordings & live sets I don't want to see in the Tracks view, but rather have them organized in a 'Samples/rec' crate and also occassionally add them to playlists.

Making hiding tracks from Playlists optional (they always remain in History) is easy, however, when selecting a playlist it calls [`PlaylistDAO::removeHiddenTracks(playlistId)`](https://github.com/mixxxdj/mixxx/blob/c04076ad5cc385928f214fda49f1d2c17d80ba24/src/library/dao/playlistdao.cpp#L559) which removes them : |
IIUC this function is simply there to have consistent UX with older databases (see quoted comment below).

So according to this comment, `removeHiddenTracks()` actually needs to be run on **all playlists** (except History?) but **only** when upgrading from Mixxx 2.1, because in later versions tracks are removed immediately.
https://github.com/mixxxdj/mixxx/blob/c04076ad5cc385928f214fda49f1d2c17d80ba24/src/library/playlisttablemodel.cpp#L150-L155

What do you think?

Furthermore, there's an inconsistency Playlists vs. Crates when hiding tracks:
**edit** figured that out, see first comment

Related bugs:
#10711 
#14135 

### TODO
- [x] make hiding tracks from playlists optional (they always remain in History)
- [ ] make hiding from crates optional (currently they remain in the crates table)
- [ ] install cleanup when upgrading from Mixxx < 2.1


ronso0:
> Furthermore, there's an inconsistency Playlists vs. Crates when hiding tracks:
> * tracks are removed from playlists immediately
> * tracks remain in crates (go to Hidden and select a track that was in a crate -> crate label is made bold)

Ah, okay, this is explained here
https://github.com/mixxxdj/mixxx/blob/1803e0269a8013d380eedc8d1cc0ea3001ca7cc4/src/library/trackset/crate/cratetablemodel.cpp#L53-L56

Playlists otoh are ordered ...
https://github.com/mixxxdj/mixxx/blob/c04076ad5cc385928f214fda49f1d2c17d80ba24/src/library/playlisttablemodel.cpp#L151-L154

which means handling hidden tracks in playlists is complicated, eg. when calculating the # position when displaying playlists, when moving tracks etc.

This means, for my personal use case (have a crate with samples which I'd like to hide from the Tracks view) there's no quick fix in sight.

Not so quick fixes:
1. have a per-track flag 'showHidden' in the `crate_tracks` db table
would allow adding  tracks from Hidden to a crate without automatically un-hiding it globally
Feasible, in conjunction with a "Hide selected track(s) from these crates" dialog it could work.
2. have a crate property 'showHidden'
Easy IMO. Assuming my use case (see above) is a niche use case, this might be the way to go.
Though that also requires to rework the auto-unhide when adding hidden tracks to a crate : |

PR #14146, opened 2025-01-08
Fix wrong access to ENV var MIXXX_VCPKG_ROOT instead of CMake setting MIXXX_VCPKG_ROOT

JoergAtGithub:
I ran into the problem, that I couldn't build any Mixxx branch anymore. I found out, that the reason was, that I created a directory D:\ports for other reasons. I have no environment variable MIXXX_VCPKG_ROOT set and therefore $ENV{MIXXX_VCPKG_ROOT}/ports pointed to D:\ports.


daschuer (approved):
Oh yes, that was wrong. Thank you for the fix. LGTM

PR #14147, opened 2025-01-08
Debian: recommend qt6-translations-l10n

### LINKED ISSUE #14137: Dialog buttons not translated in "Mapping has been edited" Dialog
### Bug Description

grafik

In the log I find the warning: 
warning [Main] Failed to load "qt" translations for locale "de_DE" from "/home/daniel/workspace/mixxx/res/translations/"

This is caused by the missing 
`qt6-translations-l10n`
package on Ubuntu Jammy. After installing it,  buttons are translated. 

Interestingly 
`qttranslations5-l10n`
is installed. but it has no dependencies to other packages. 

Does anyone know how the policy with this package is? 

### Version

2.5.0

### OS

Ubuntu Jammy

daschuer:
This shall fix missing Qt translations. Fixes #14137


daschuer:
@ronso0 Can you have a look here? 

I think because of the Deere Waveform crasher #14170  we should do a 2.5.1 bug fix version soon.

Swiftb0y (approved):

PR #14140, opened 2025-01-08
Library: add overview column

ronso0:
continuation of #13638
This is now based on @ninomp's initial POC commit 80d8b5f6b6a4c0843ed64722588073717395ec28
from https://github.com/ninomp/mixxx/commits/overviewsinlibrary-rebased/
(incl. the fix from #14150 )
I addressed @Swiftb0y's review and a lot evolved anyway.

image

* the library shows a pseudo-mono view (hardcoded in `OverviewCache::prepareOverview`), though with a bit of refactoring this _could_ be made optional
* overviews are rendered as soon as track analysis finished (loading tracks not required)
* overviews (decks and library) are updated when the library row height, overview type or the Normalize/visual gain options are changed, and the cache is cleared
* the signal colors can be set in the `<Library>` node exactly like in the waveform templates `(<Visual>` widget)
* WOverView and OverviewDelegate use the same renderer for all types which I extracted from WOverview to src/waveform/renderers/waveformoverviewrenderer
(I adopted the RGB simplifications made by @**Nino MP**, and simplified the others, too)
* there's an OverviewCache which holds the images sized for the overview column

### Ideas / Nice to have
* show analyzer progress / partial painting like in decks
* add preferences option for mono-mixdown / stereo
* click Overview column header to sort by 'has overview'
* cleanup / optimize draw functions in WaveformOverviewRenderer
* some sort of analysis progress indicator in the library? (though I _think_ partial rendering would be distracting)
* implement preview interface: click to select, click position to preview
* show cue markers?

PR #14150, opened 2025-01-09
(fix) WOverview: also render analysis progress when triggered by WTrackMenu/AnalysisFeature

ronso0:
Currently, the analysis progress is rendered in the overview widgets only for analysis triggered when loading a track.
Now it's also rendered when the loaded track is re-analyzed (after clearing the wavefom).

1. load a new track to a deck (or take an analyzed track all clear waveform, BPM, ReplayGain ..)
= analysis starts, in overview the analysis progress bar is drawn and waveform is rendered as data is being created
2. track menu -> Clear -> Waveform
3. re-analyze (via deck track menu, library track menu or find track in Analyze feature)
= no progress bar in overview, no feedback <-- **fixed**, same feedback as with 1.

by-catch of #14140


daschuer:
I can confirm the issue and the fix. LGTM, Thank you.

Issue #13262, opened 2024-05-22
DEBUG ASSERT: "it == cbegin() || it == cend() || *it >= position" in function class mixxx::Beats::ConstIterator __cdecl mixxx::Beats::iteratorFrom(class mixxx::audio::FramePos) const at src\track\beats.cpp:470


JoergAtGithub:
### Bug Description

Today I go with main and some other PR based on it the following assert, after very short time of playing with 4 synced decks:

```
 	[Inline Frame] mixxx.exe!mixxx_debug_assert(const char *) Line 9	C++
 	mixxx.exe!mixxx::Beats::iteratorFrom(mixxx::audio::FramePos position) Line 471	C++
 	mixxx.exe!mixxx::Beats::findNBeatsFromPosition(mixxx::audio::FramePos position, double beats) Line 501	C++
 	mixxx.exe!LoopingControl::hintReader(gsl::not_null<QVarLengthArray<Hint,512> *> pHintList) Line 591	C++
 	[Inline Frame] mixxx.exe!EngineBuffer::hintReader(const double) Line 1510	C++
 	mixxx.exe!EngineBuffer::processTrackLocked(float * pOutput=0xc0a53125b32dee8a, const int iBufferSize, mixxx::audio::SampleRate sampleRate) Line 1146	C++
 	mixxx.exe!EngineBuffer::process(float * pOutput=0x000002121549a0b0, const int iBufferSize) Line 1180	C++
 	mixxx.exe!EngineDeck::process(float * pOut=0x000002121549a0b0, const int iBufferSize) Line 65	C++
 	mixxx.exe!EngineMixer::processChannels(int iBufferSize) Line 370	C++
 	mixxx.exe!EngineMixer::process(const int iBufferSize) Line 430	C++
 	mixxx.exe!SoundDevicePortAudio::callbackProcessClkRef(const __int64 framesPerBuffer=1024, float * out=0x0000021232932160, const float * in=0x0000000000000000, const PaStreamCallbackTimeInfo * timeInfo=0x0000002b309ff7a8, unsigned long statusFlags=0) Line 998	C++
 	mixxx.exe!`anonymous namespace'::paV19CallbackClkRef(const void * inputBuffer, void * outputBuffer, unsigned long framesPerBuffer, const PaStreamCallbackTimeInfo * timeInfo, unsigned long statusFlags=0, void * soundDevice=0x00000212114788f0) Line 72	C++
[...]

PR #14159, opened 2025-01-10
Fix crash due to concurrent access in MidiController (Alternative)

### LINKED ISSUE #13940: qhash.h:790 ASSERT on multiple MIDI messages in quick succession

daschuer:
This is an alternative to fix #13940

I have now "hopfully" managed to wrap my head around the issue. The issue was that the mapping has been cloned from the GUI thread while the Controller thread has applied JS created mappings. This creates a broken copy of the mapping that crashes Mixxx after it was applied back to the Controller thread.  

The solution here keeps now a clean mutable copy of the mapping for the GUI thread. The GUI thread waits until the controller thread has cloned the mapping during "Apply" this way we have sorted out all concurrent access to the mappings. 

This PR is unfortunately a big bigger than expected, because some related code and comments have not been maintained over the years and was misleading making the issue hard to understand.

### DIFF
--- a/src/controllers/dlgprefcontroller.cpp
+++ b/src/controllers/dlgprefcontroller.cpp
@@ -62,7 +62,7 @@ DlgPrefController::DlgPrefController(
     initTableView(m_ui.m_pOutputMappingTableView);
 
     std::shared_ptr<LegacyControllerMapping> pMapping = m_pController->getMapping();
-    slotShowMapping(pMapping);
+    showMapping(pMapping);
 
     m_ui.labelDeviceName->setText(m_pController->getName());
     QString category = m_pController->getCategory();
@@ -200,7 +200,7 @@ void DlgPrefController::showLearningWizard() {
         m_pMapping = std::make_shared<LegacyMidiControllerMapping>();
         emit applyMapping(m_pController, m_pMapping, true);
         // shortcut for creating and assigning required I/O table models
-        slotShowMapping(m_pMapping);
+        showMapping(m_pMapping);
     }
 
     // Note that DlgControllerLearning is set to delete itself on close using
@@ -689,7 +689,7 @@ void DlgPrefController::slotMappingSelected(int chosenIndex) {
         // the preset combobox.
         enumerateMappings(mappingFilePath);
     }
-    slotShowMapping(pMapping);
+    showMapping(pMapping);
 }
 
 bool DlgPrefController::saveMapping() {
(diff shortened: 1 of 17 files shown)


daschuer:
@mxmilkiib I have added mutex with debug Information. 
This will hopefully finally fix the crash. However It needs some polishing.
Please Test and provide the mixxx.log. (or the backtrace if it is still crashing)

daschuer:
@mxmilkiib Does this still crash?

mxmilkiib:
yes 

maaaybe this is config file related, removing files manually but Mixxx not forgetting

can't remember the state from before though

daschuer:
Can you please share the mixxx.log of the crash? 
A backtrace is also helpful.

mxmilkiib:
ah this is the older issue

I'll see if I can replicate now, not had any recent ones for this issue though, building from this branch that is

daschuer:
Cool, please also share the mixxx.log from a short test run without crash. This will help me polishing.

mxmilkiib:
https://gist.github.com/mxmilkiib/b223874ec20f85b07840b7e0d2a5ba55

[I managed to make it crash](https://gist.github.com/mxmilkiib/b223874ec20f85b07840b7e0d2a5ba55), on clicking play though.

`DEBUG ASSERT: "it == cbegin() || it == cend() || *it >= position" in function mixxx::Beats::ConstIterator mixxx::Beats::iteratorFrom(mixxx::audio::FramePos) const at ./src/track/beats.cpp:470`

daschuer:
The crasher is unrelated and tracked here: 
https://github.com/mixxxdj/mixxx/issues/13262

daschuer:
@mxmilkiib Please test if the crasher is fixed now.

mxmilkiib:
I haven't had any further issues so I guess draw a line under it for now.

acolombier (approved):
Code's LGTM, but no Midi mapping to test it with. Looks reasonably safe to merge without it, thanks!

src/controllers/dlgprefcontroller.cpp
```diff
@@ -103,7 +103,11 @@ DlgPrefController::DlgPrefController(
     connect(this,
             &DlgPrefController::applyMapping,
             m_pControllerManager.get(),
-            &ControllerManager::slotApplyMapping);
+            &ControllerManager::slotApplyMapping,
+            Qt::BlockingQueuedConnection);
```
acolombier, line 107: This has introduced a deadlock when using with screen controller, which rely on the main thread (GUI) availability ([see doc](https://doc.qt.io/qt-6/qoffscreensurface.html#create))

PR #14164, opened 2025-01-12
Add local time sync ref (for AbletonLink etc.) selection to SoundManager

JoergAtGithub:
This PR implements the algorithm for determining a synchronization sound-device, that is explicitly selected for time synchronization with functions outside of audio processing in the engine. The synchronization itself, will be added in a follow-up PR.
Everything the DJ hears and sees must be in sync with the local PortAudio sound device, taking into account the full output latency (including the latency contributed by the particular sound API). Even in cases where we use the network clock as reference for the interval of the audio buffer in the engine.

This PR:
- Adds local time sync ref selection to SoundManager
- Moved pNewMainClockRef selection out of main loop and added proper handling of setups where only headphones or booth are configured. 
- Simplified fallback mechanisms and removed redundant code. 
- Added unit tests for new function selectLocalTimeSyncRef


src/soundio/soundmanager.cpp
```diff
@@ -363,7 +393,77 @@ SoundDeviceStatus SoundManager::setupDevices() {
...
+                    0);
+            deviceOutputs[pDevice].append(out);
+        }
+    }
+
+    // Select pNewLocalTimeSyncRef
+    // The local time sync reference is the device that is used for the
+    // synchronization of processes outside the audio processing engine
+    // to the DAC timing of the local sounddevice the DJ hears.
+    // This sync reference shall be used for:
```
daschuer, line 415: This comment is missleading. Because for my undertsanding we only have one clock everything is synced too. 
pNewLocalTimeSyncRef is only a temporary, right?

daschuer:
> Everything the DJ hears and sees must be in sync with the local PortAudio sound device, taking into account the full output latency (including the latency contributed by the particular sound API). Even in cases where we use the network clock as reference for the interval of the audio buffer in the engine.

I have probably not yet understand the issue. All of thy syncing is only amatter win which stream padding samples are added. Ony the Master is "clean" and does not need padding samples. 

The case that could be problematic is the case if the network clock is used and we have passing samples in the Master output. What does AbletonLink do in that case? Which clock is used for syncing AbletonLink devices?

daschuer (changes requested):

src/soundio/soundmanager.cpp
JoergAtGithub: This PR is about selecting the sounddevice that is to be used as reference for external syncronisation. This is the sounddevice with the DAC that generates the sound the DJ hears.
Time syncronization has nothing to do with the clock and this comment section does not contain this term therefore.
Output Latency is specific for each sounddevice and contains various buffers inside and outside of Mixxx. All DAC channels of the same sound-device have the same output-latency, but different sound-devices may have different output-latency. Therefore we need to determine the sound-device that is the reference, before we can gather the correct output-latency.
Logically, the DJ always hears the sound via one of the local PortAudio devices and never via the stream broadcasted over the network. Accordingly, a soundDeviceNetwork can never be useful as a reference for synchronization.

src/soundio/soundmanager.cpp
```diff
@@ -363,7 +388,77 @@ SoundDeviceStatus SoundManager::setupDevices() {
 
     // pair is isInput, isOutput
     QVector<DeviceMode> toOpen;
-    bool haveOutput = false;
+
+    // Get all outputs for each device
+    QHash<SoundDevicePointer, QList<AudioOutput>> deviceOutputs;
+    for (const auto& pDevice : std::as_const(m_devices)) {
+        deviceOutputs[pDevice] = m_config.getOutputs().values(pDevice->getDeviceId());
```
daschuer, line 395: This avoid the hash lookup before writing:
```suggestion
        deviceOutputs.insert(pDevice, m_config.getOutputs().values(pDevice->getDeviceId()));
```

src/soundio/soundmanager.cpp
daschuer: The Mixxx engine picks one DAC crystal for synchronization or the network clock. 
Let's imagine we have a buffer size of 1024 frames and the network clock is 0,1% faster than the DAC (unlikely, but good for this example. 

For synchronization, the DAC is feed with only 1023 frames. Broadcasting and recording still receive all 1024 frames. 
This means that from DAC perspective a 120 BPM track is played with 119,9 BPM, however Mixxx reports 120 BPM.
This means if the DAC becomes now the pNewLocalTimeSyncRef it will be off compared to the Mixxx data.

daschuer (changes requested):

src/soundio/soundmanager.cpp
```diff
@@ -321,6 +321,35 @@ void SoundManager::queryDevicesMixxx() {
...
+            AudioPathType::Deck,
+            AudioPathType::Bus,
+            AudioPathType::Headphones,
+            AudioPathType::Booth};
+
+    for (const auto& pDevice : devices) {
+        if (pDevice->getDeviceId().name == kNetworkDeviceInternalName) {
+            continue;
+        }
+        QList<AudioOutput> outputs = deviceOutputs[pDevice];
```
daschuer: This avoids a copy with implicit sharing and a risk for a deep copy later: 
```suggestion
        const QList<AudioOutput>& outputs = deviceOutputs[pDevice];
```

src/soundio/soundmanager.cpp
```diff
@@ -321,6 +321,35 @@ void SoundManager::queryDevicesMixxx() {
     m_devices.append(currentDevice);
 }
 
+SoundDevicePointer SoundManager::selectLocalTimeSyncRef(
+        const QHash<SoundDevicePointer, QList<AudioOutput>>& deviceOutputs,
+        const QList<SoundDevicePointer>& devices) {
+    const std::array<AudioPathType, 5> priorityOrder = {
```
daschuer:
```suggestion
    constexpr std::array<AudioPathType, 5> priorityOrder = {
```
to assert it is not populated at run time (should normally already be the case)

src/soundio/soundmanager.cpp
daschuer: After having a second look, I think the confusion is sourced by the fact that we have here now 
- pNewLocalTimeSyncRef
- pNewMainClockRef

pNewLocalTimeSyncRef is redundant when we move the `if (m_config.getForceNetworkClock() && !jackApiUsed())` block up and call selectLocalTimeSyncRef() in the else branch.

src/soundio/soundmanager.cpp
```diff
@@ -363,7 +388,77 @@ SoundDeviceStatus SoundManager::setupDevices() {
...
+            }
+        }
+    } else {
+        pNewMainClockRef = pNewLocalTimeSyncRef;
+    }
+
+    // Fallback to keep waveforms running if no local SoundDevice is configured
+    // If pNewLocalTimeSyncRef or pNewMainClockRef is still nullptr,
+    // set it to the first network clock device.
+    if (!pNewLocalTimeSyncRef || !pNewMainClockRef) {
```
daschuer: This whole block can be integrated in selectLocalTimeSyncRef.

daschuer (changes requested):

src/soundio/soundmanager.cpp
JoergAtGithub: No, not at all. This ist the only feature this PR implemets. Without this the whole PR is for nuts.

JoergAtGithub:
> The Mixxx engine picks one DAC crystal for synchronization or the network clock.
> Let's imagine we have a buffer size of 1024 frames and the network clock is 0,1% faster than the DAC (unlikely, but good for this example.
> 
> For synchronization, the DAC is feed with only 1023 frames. Broadcasting and recording still receive all 1024 frames.
> This means that from DAC perspective a 120 BPM track is played with 119,9 BPM, however Mixxx reports 120 BPM.
> This means if the DAC becomes now the pNewLocalTimeSyncRef it will be off compared to the Mixxx data.

This is completly off topic and has nothing to do with the pNewLocalTimeSyncRef added in this PR.

JoergAtGithub:
I addressed the code review comments, but the only aim of the PR is to determine pNewLocalTimeSyncRef and to return it:
- This is, as the symbol name says, about **time** !!!
- It is not about clocks, crystal or drifting clock rates
- It is out of scope of soundmanager, and also doesn't matter, which clock is used elsewhere to measure the time

src/soundio/soundmanager.cpp
daschuer: This one is still open.

src/soundio/soundmanager.cpp
daschuer: pNewLocalTimeSyncRef is only used here: 
https://github.com/JoergAtGithub/mixxx/blob/782c430d22f6ee4d01fe1ae63bde6ce97866d4f5/src/soundio/soundmanager.cpp#L634

What can you explain how you want to use it later? Why does this use case require to bypass the ForceNetworkClock flag? 

> It is out of scope of soundmanager, and also doesn't matter, which clock is used elsewhere to measure the time

I don't understand this sentences. In terms of a computer, time is measured number via clocks there is no "time" we can grap. In Mixxx we have three relevant clocks: 
* DAC
* Wallclock synced Network clock
* PerformanceTimer CPU clock 

For my understanding, Ableton link can be used to play two tracks in sync using two devices connected via Network by sharing beat information. Can you confirm that?

daschuer (changes requested):

JoergAtGithub:
The feature added in this PR is to find the local time reference - nothing else.

This has feature is completly unrelated to any clock. Therefore none of the three mentioned clocks has any relation to this PR.

The feature added in this PR selects a time reference that is local (that what the DJ hears - not what the listeners of a radio station hear in Tokio or New York at the same time).

JoergAtGithub:
@daschuer Could you please merge this now, to prevent conflicts with #16707

daschuer:
This is unfortunately not yet in a merge-able, because of the open questions around pNewLocalTimeSyncRef.
This PR seems to me to implement a change that is maybe or maybe not required depending on the outcome of our discussion in [AbletonLink](https://github.com/mixxxdj/mixxx/pull/10999)
@pri-yan-shu will check the related Pipwire behaviour in hins GSoC project maybe he can also shade some light to the issue. 
I you wish to get the uncritical refactoring of this PR anyway, you may spit them out to another PR.

JoergAtGithub:
I answered all the questions several times, and added all the verbose comments on your request. I will request a formal vote now!

daschuer:
The pending issue is that pNewLocalTimeSyncRef is unused. There is IMHO no reason to add unused code. You have also not responded to my investigations in #10999. 
Can we prostpone these questions after we have a 2.6 release? I am afraid this will drag a lot of energy and leads to more delay for this important releasee already overdue by 10 month.

acolombier:
As part of the vote submitted to Zulip, I have tried to review the state and the conversation the followed. Please bear in mind I might have missed some details somewhere.

From my understanding, there is two issues:
1. The unused/redundant code
2. The inline annotation wording 

# 1. The unused/redundant code

From my understanding, this issue is that `pNewLocalTimeSyncRef` will always be computed, even if it won't be used in some cases (that is network clock is forced and JACK API is not used). Alternatively, when `pNewLocalTimeSyncRef` gets used, it is only referenced one, to assign `pNewMainClockRef`.
Though I agreed it would be best to remove `pNewLocalTimeSyncRef` and compute it directly to `pNewMainClockRef`, I consider this a minor code smell, which could justify a nit comment, not a hard change request.

# 2. The inline wording 

The current wording make sense to me but since we are all not native speaker, I have used an external judge and ask it to explain it back to me (ChatGPT). 
image
This looks correct to me and seems to confirm the current wording is acceptable.

Unless there are other elements I've missed, the blocking change request appears to be unjustified and should be dismissed to allow other member to approve and merge this PR.

daschuer:
@acolombier thank you for jumping in. 

lt;dr;
This is the summary of my blocking claim: 
The proposed functional change in this PR is redundant, because Mixxx already supports both hardware audio clock synchronization and network-based time synchronization, making the new pNewLocalTimeSyncRef unnecessary. While the PR attempts to address synchronization issues between sound cards, the core problem is already handled. For merging I request to shot the issue that shall be solved and that it is solved. 

Details: 

Your analysis about the PR itself is correct and normally there would be nothing against merging it. The cited comment are minor and not worth such a clash we have. 

The blocking issue is however that this PR presents a the solution for an non existing problem. I have probably failed to describe it here in an understandable way. I also could be wrong that's why an unbiased pair of eyes like yours is very helpful. 
I did however already a detailed analysis in https://github.com/mixxxdj/mixxx/pull/10999 that also build up a wrong assumption how Mixxx works, likely adopted form other tools where the described problem exists. 

So let's start from the beginning. 

This PR tries to be the base for the synchronization issue between two sound cards, either connected to the same computer or on a different computer or electronical instrument.  

The issue is well described here: 
https://www.tonstudio-forum.de/blog/entry/108-clock-synchronization-background-rme-features-scenarios-en-de/

We have two options,
1. The audio stream itself using one of the soundcard clock as reference.   
2. Word Clock (WC), a clock synchronized via NTP, GNSS or DCF, driven by an atomic reference clock e.g. located in Braunschweig (DE) 

@JoergAtGithub is correct with his requirement for 2. 
> Everything the DJ hears and sees must be in sync with the local PortAudio sound device,

He however missed the fact that we already have implemented the two options. (All blame on me that I have written the code probably not so obvious. We need to improve on that. 

For synchronization with two clocks, we need to make sure that the track transport is done with the same clock. 
Let's assume we two soundcards, each with a 1% crystal. One running at +1% the other at -1%. 
Compared to WC, the a 120 BPM track will be played with 121,2 BPM the other with 118,8 causing an awful shifting double beat even both are played back at unity speed of 120 BPM. 

For solution 1 the + 1% clock may become the leader (Im Mixxx when running the main output). This will cause the Mixxx engine running with +1% and the track running with 121,2 BPM measured with the world clock. The engine itself will however still indicate 120 BPM. The effect is that too many samples are produced for the second soundcard at -1 %. The poor man's solution in Mixxx just drops 2 %  of samples which is not hearable. In recent Pipewire a resampler is used to reduce the number of samples. 

While solution 1 works fine and allows a bit perfect output at  main output and for recording (the primary product of a DJ). 
It does not work for broadcasting. The network buffers filling values are not accessible at the receiver device and that causes a stacking up of 1 % of the samples causing buffer overflow in regular intervals. 

This is solved with the "Netwok Clock" in Mixxx running the transport engine. It is the NTP atomic clock synced, usually smoothly adjusted via a PLL filter.In this solution, the broadcasting output provides World clock exact sample rate and BPM. The downside is that the main clock in our scenario is 1% to fast, using copied padding samples to catch up. 

If we assume the DJ has also a absolute WC reference in his head he would hear the track exactly at 120 BPM 

The solution presented in https://github.com/mixxxdj/mixxx/pull/10999 is not 1 and not pure 2. It uses the performance timer driven by the CPU crystal, luckily way more exact than the DAC clock, usually < 0.01%. But still not synced across devices as our Network clock is. Ablaton Live mitigate this by a decentralized peer-to-peer adjustment loop. A kind of WC solution without the Need for NTP. 

In Mixxx we now have the option to run the tracks bit perfect with the DAC clock and communicate 121,2 BPM via Ableton Live
(solution 1) or run the engine with the NTP sycned network clock, which is the CPU clock on absence of NTP. 

My conclusion is, that we have already our two clocks and there is no need to introduce a third one. We may add a NTP preferences option check mark for our network clock to it exactly like  presented in the Ableton live reference implementation. 

That's why I think pNewLocalTimeSyncRef is redundant. I might be wrong, but it is hard to guess without code that actually makes use of pNewLocalTimeSyncRef. 

What I know is that it does not work for 

    // 1.) VSync of the waveforms
    // 2.) Sync of external audio device (e.g. drum machine) with feedback
    //     to an auxiliary input of the mixer in a external mixing setup
    // 3.) Sync of lighting (DMX) or video (VJ)

Because the selected device is driven with dropped and duplicated samples at rather random moments depending on the CPU scheduling and the resulting buffer load, this clock is not reliable as any reference.  

I hope you understand now my concerns. I might be however wrong. 

------

But what this PR needs to be able to prove or disprove is is this: 

I need to see code that uses pNewLocalTimeSyncRef.  How is this used. How are is the jitter by the dropped frames compensated. 

Why does the network clock not work. Maybe you can show a graph that shows how unreliable it is. Maybe it is on Windows? 

I actually can imagine that it is unreliable if the NTP sync is implemented sloppy. The solution would be to add a preference option to run Mixxx with the performance timer like expected by Ableton Live. 

I can also imagine that the engine implementation based on the network clock is faulty. The short hand solution would be not to use the Network clock for Live. 

----

Here some references: 
https://github.com/mixxxdj/mixxx/blob/337154fdeb47e56c0ddbec01d7adcce89f9f00c7/src/soundio/sounddeviceportaudio.cpp#L286-L287
https://github.com/mixxxdj/mixxx/blob/337154fdeb47e56c0ddbec01d7adcce89f9f00c7/src/soundio/sounddeviceportaudio.cpp#L787
https://github.com/mixxxdj/mixxx/blob/337154fdeb47e56c0ddbec01d7adcce89f9f00c7/src/soundio/sounddeviceportaudio.cpp#L815-L829

I am happy to discuss this in a conference call after you have read this. We really need a solution that works. It is not bearable that Ableton live feature is rotting for so long.

daschuer:
After the phone call with @JoergAtGithub I know now what this PR should lay the ground for. It is the air to air or likewise analog to analog synchronization of two Ableton Link devices. 

Ableton Link uses a common timeline for synchronization, based on the CPU crystal and a filter that measures the tempo offset in case of two devices with sepate CPUs. This timeline has in addition to the pure tempo also position information. To have that in sync, Mixxx need to shift the timeline to the position when the sound reaches the air. In Ableton Live and also on Mixxx there is a delay parameter that allows to shift this time line. 

Some audio APIs deliver precise timing information of the delay until a buffer reaches the analog domain. 
We use it to adjust or waveform properly the the beats the user hears, and not the beats Mixxx is processing. 

This is a welcome addition to Mixxx. This PR need however some polishing to make that understandable and maintainabel. 
We need third clock option for Ableton Live. 
And we need to consider three use cases
* Analog Mixxer 
* Independent Mixxer App (e.g. pipewire) 
* Aux Mixing in Mixxx (for broadcasting) 

@JoergAtGithub did I summarize everything correctly?

JoergAtGithub:
> 
> @JoergAtGithub did I summarize everything correctly?

Mostly, but I don't understand what you mean with "We need third clock option for Ableton Live." This sentence makes no sense to me.

JoergAtGithub:
I still don't now what you want me to change in this PR. From my point of view, the selection logic is correct and this PR should be merged as it is.

daschuer:
> We need third clock option for Ableton Live.

Ableton Link used the CPU clock for synchronizing. To have big rate conversion factor it would be useful to let Mixxx run with the CPU clock. Currently we can use the soundcard clock or the NTP syncs Network clock.

JoergAtGithub:
Ableton Link does not use a clock interval at all. Therefore there is no rate conversion involved. This is not how it works!

daschuer:
How does it work without rate conversion? Please explain some details.

daschuer:
According to our new assumptions the naming 'isClkRefDevice' and 'pNewMainClockRef' is weak. 
In pipewite such a device is called the "driver of a graph". It controlles the timing and transport. 

So how about renaming it to 'isTransportDriver` and `pTransportDriver` 

pNewLocalTimeSyncRef is also weakly named. It is used to adust the offset of the Ableton Link time line. The API can probably report the time until the signal reaches the analog domain. 
So this can become the pTimeToAnalogRef. 

Does that make sense? If yes, you should change the code and comment to the new nomenclature. 
I will than do final review with this new knowledge.

JoergAtGithub:
> According to our new assumptions the naming 'isClkRefDevice' and 'pNewMainClockRef' is weak. In pipewite such a device is called the "driver of a graph". It controlles the timing and transport.
> 
> So how about renaming it to 'isTransportDriver`and`pTransportDriver`

The symbol pNewMainClockRef is not added by this PR, and isClkRefDevice is not existing in soundmanager. I think the naming still fits as I did not change it's behavior. The code is just refactured and returns in any case the same as the old implementation.
But I wouldn't mind, if you rename this one an own PR, but this is clearly out of scope of this PR.

> pNewLocalTimeSyncRef is also weakly named. It is used to adust the offset of the Ableton Link time line. The API can probably report the time until the signal reaches the analog domain. So this can become the pTimeToAnalogRef.
> 
> Does that make sense?

No, pTimeToAnalogRef fits not at all. It is missing every single characteristic of it, which are:

- It's local (where the DJ is located)
- It's a time synchronisation reference output
- The "to" is also incorect, as the syncronisation reference is the time "at" this output.

JoergAtGithub:
> From my understanding, this issue is that `pNewLocalTimeSyncRef` will always be computed, even if it won't be used in some cases (that is network clock is forced and JACK API is not used).

@acolombier The new introduced method SoundManager::selectLocalTimeSyncRef does not contain an exception for JACK.
A key feature is that a local audio device is also returned as a reference for local time synchronization, in cases where the master clock is not taken from another source.

daschuer:
This PR introduces a new type of Ref where I have first struggled completely to understand the purpose. 
One reason is IMHO the similar names for two different timing references: 
* pNewLocalTimeSyncRef
* pNewMainClockRef

One part is the bad legacy name and one the bad new name. This was not a big issue before. Now in this PR the diffrence becomes crucial to understand the purpose of this PR. That is why this is a blocking issue for this PR. 

The proposed names would help me and probably to understand the diffrence flowing the establish pipewire nomenclature.  
* pTransportDriver
* pTimeToAnalogRef.

Feel free to propose other names if you like.

JoergAtGithub:
The proposed names would help me and probably to understand the diffrence flowing the establish pipewire nomenclature.

    pTransportDriver
    pTimeToAnalogRef

The naming `pTransportDriver` is fine to me.
But, as explained above, `pTimeToAnalogRef` does not fit in meaning. What about:

- pLocalTimeSyncRef
- pRefSndDevForAbsoluteTime
- pAbsTimeRefSndDev
- pTimeSyncSndDev 
- pTimeRefSndDev 
- pSoundDeviceUsedAsTimeSyncRef
- pSoundDeviceUsedAsReferenceForExternalTimeSyncing
- pLocalOutputDeviceUsedForTimeSyncing
- pRefOutputDeviceForTimeSync
- pRefSndDevForTimeSync
- pLocalSyncTimeRef
- pLatencyRefDevice

daschuer:
pLatencyRefDevice fits the best IMHO. 

"AbsTime" and time "TimeSync" names do not work, because all these ad similar name do not reflect that the device is not used for rate syncing. It is only used for automatically determining the offset parameter in the sound preferences also used for he Ableton Link common time line.

acolombier:
Thanks @daschuer for the detailed explanation. Indeed I was missing the larger picture and it now make more sense why this clock appears as redundant at first.

I am very unfamiliar with the synchronisation concept but I think the suggested naming convention does clarify to intend behind these clocks

JoergAtGithub:
I renamed pNewLocalTimeSyncRef to pLatencyRefDevice.
I renamed ClkRef to TransportDriver, which affected many other files than soundmanager.cpp too.
Please confirm that this is what you wanted.

PR #14170, opened 2025-01-14
Fix crash when hiding waveforms in Deere

### LINKED ISSUE #14157: Crash when minimising waveforms section
### Bug Description

**Steps to reproduce**
1. Load any track(s) in any deck(s). **edit(@ronso0)**: a marker must be visible in the waveform (eg. a hotcue, with Cue alone it won't  crash)
2. Grab the bottom edge of parallel waveforms section with mouse click, drag to top to minimise it.
3. Mixxx instantly hang and then crash. Every time

I can enlarge the section and then make it smaller back with no issue. But dragging the edge up to the point it would disappear cause the bug.

This was first spotted on **2.4.1.** Updated to **2.4.2** to still get the bug. Updated to **2.5.0** and the bug is still there.

**Related settings:**
Skin: Deere, 4 decks, Waveform type: Filtered (GLSL)

### Version

2.5.0

### OS

Windows 10 Pro 22H2

daschuer:
This fixes #14157

### DIFF
--- a/src/waveform/renderers/allshader/waveformrendermark.cpp
+++ b/src/waveform/renderers/allshader/waveformrendermark.cpp
@@ -97,7 +97,7 @@ void allshader::WaveformRenderMark::initializeGL() {
 }
 
 void allshader::WaveformRenderMark::drawTexture(
-        const QMatrix4x4& matrix, float x, float y, QOpenGLTexture* texture) {
+        const QMatrix4x4& matrix, float x, float y, QOpenGLTexture* pTexture) {
     const float devicePixelRatio = m_waveformRenderer->getDevicePixelRatio();
     const float texx1 = 0.f;
     const float texy1 = 0.f;
@@ -105,9 +105,9 @@ void allshader::WaveformRenderMark::drawTexture(
     const float texy2 = 1.f;
 
     const float posx1 = x;
-    const float posx2 = x + static_cast<float>(texture->width() / devicePixelRatio);
+    const float posx2 = x + static_cast<float>(pTexture->width() / devicePixelRatio);
     const float posy1 = y;
-    const float posy2 = y + static_cast<float>(texture->height() / devicePixelRatio);
+    const float posy2 = y + static_cast<float>(pTexture->height() / devicePixelRatio);
 
     const float posarray[] = {posx1, posy1, posx2, posy1, posx1, posy2, posx2, posy2};
     const float texarray[] = {texx1, texy1, texx2, texy1, texx1, texy2, texx2, texy2};
@@ -130,11 +130,11 @@ void allshader::WaveformRenderMark::drawTexture(
 
     m_textureShader.setUniformValue(textureLocation, 0);
 
-    texture->bind();
+    pTexture->bind();
 
     glDrawArrays(GL_TRIANGLE_STRIP, 0, 4);
 
-    texture->release();
+    pTexture->release();
 
     m_textureShader.disableAttributeArray(positionLocation);
     m_textureShader.disableAttributeArray(texcoordLocation);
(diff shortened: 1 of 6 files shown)


src/waveform/renderers/waveformmark.cpp
```diff
@@ -335,7 +335,7 @@ class MarkerGeometry {
 };
 
 QImage WaveformMark::generateImage(float devicePixelRatio) {
-    assert(needsImageUpdate());
+    DEBUG_ASSERT(needsImageUpdate());
```
ronso0, line 338: VERIFY_OR_DEBUG_ASSERT ?
IIUC all below would be a no-op otherwise, right?

src/waveform/renderers/allshader/waveformrendermark.cpp
```diff
@@ -383,20 +388,22 @@ void allshader::WaveformRenderMark::drawUntilMark(const QMatrix4x4& matrix, floa
 // Note that in the legacy waveform widgets this is drawn directly
 // in the WaveformWidgetRenderer itself. Doing it here is cleaner.
 void allshader::WaveformRenderMark::updatePlayPosMarkTexture() {
-    float imgwidth;
-    float imgheight;
+    const float imgheight = m_waveformRenderer->getBreadth();
+    const float imgwidth = 11.f;
```
ronso0, line 392:
```suggestion
    const float imgHeight = m_waveformRenderer->getBreadth();
    const float imgWidth = 11.f;
```
and where used, of course.

src/waveform/renderers/allshader/waveformrendermark.cpp
```diff
@@ -383,20 +388,22 @@ void allshader::WaveformRenderMark::drawUntilMark(const QMatrix4x4& matrix, floa
...
 
-    imgwidth = 11.f;
-    imgheight = height;
-
     QImage image(static_cast<int>(imgwidth * devicePixelRatio),
             static_cast<int>(imgheight * devicePixelRatio),
             QImage::Format_ARGB32_Premultiplied);
+    if (image.isNull()) {
+        return;
+    }
```
ronso0, line 406: imgWidth is < 0 and there is already a check for `imgHeight != 0` above, so this can only be NULL if devicePixelRatio is 0.
Can this ever happen?

src/waveform/renderers/waveformmark.cpp
```diff
@@ -355,11 +377,14 @@ QImage WaveformMark::generateImage(float devicePixelRatio) {
...
 
-    m_label.setAreaRect(markerGeometry.m_labelRect);
+    m_label.setAreaRect(markerGeometry.labelRect());
 
     // Create the image
     QImage image{markerGeometry.getImageSize(devicePixelRatio),
             QImage::Format_ARGB32_Premultiplied};
+    if (image.isNull()) {
+        return image;
+    }
```
ronso0, line 387: how can that happen?
There is already the `m_breadth == 0.0f` above.

ronso0 (changes requested):

src/waveform/renderers/allshader/waveformrendermark.cpp
daschuer: When looking at the Qt source, there are other reasons like out of memory or such. All can't probably not happen but this saves Mixxx from crash if it happen anyway. I think a 
VERIFY_OR_DEBUG_ASSERT is a good choice.

ronso0:
FWIW since this happens only with Deere (which is the only skin that uses waveform Singletons), I assume this is an issue with WSplitter/QSplitter and Singleton, like when the waveform part of the splitter is collapsed the waveform is not hidden or something?

src/waveform/renderers/waveformmark.cpp
daschuer: Yes, but we assert that it never happens and when it happens, it does no harm.

daschuer:
> like when the waveform part of the splitter is collapsed the waveform is not hidden or something?

Yes, the widget is sized to 1 before it is hidden due to the border the images have 0 height, which has caused the crash.

daschuer:
Done

ronso0:
Ah, now I could reproduce the crash: you need to have one or more hotcue markers visible in waveform.

Can confirm the fix :+1:

ronso0 (approved):
LGTM, thanks for this fix!

PR #14175, opened 2025-01-15
Expose track info to JS controllers

christophehenry:
Following #13935, my proposed solution to expose track infos to JS mappings based on @Swiftb0y's idea.

### DIFF
--- a/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
@@ -31,7 +31,8 @@ constexpr double kBrakeRampToRate = 0.01;
 ControllerScriptInterfaceLegacy::ControllerScriptInterfaceLegacy(
         ControllerScriptEngineLegacy* m_pEngine, const RuntimeLoggingCategory& logger)
         : m_pScriptEngineLegacy(m_pEngine),
-          m_logger(logger) {
+          m_logger(logger)
+{
     // Pre-allocate arrays for average number of virtual decks
     m_intervalAccumulator.resize(kDecks);
     m_lastMovement.resize(kDecks);
@@ -134,6 +135,13 @@ QJSValue ControllerScriptInterfaceLegacy::getSetting(const QString& name) {
                         .arg(name));
         return QJSValue::UndefinedValue;
     }
+}
+
+QObject* ControllerScriptInterfaceLegacy::getPlayer(const QString& deck) {
+#ifndef MIXXX_USE_QML
+    return nullptr;
+#endif
+    return m_pScriptEngineLegacy->getPlayer(deck);
 }
 
 double ControllerScriptInterfaceLegacy::getValue(const QString& group, const QString& name) {
--- a/src/qml/qmlplayermanagerproxy.cpp
+++ b/src/qml/qmlplayermanagerproxy.cpp
@@ -60,7 +60,7 @@ void QmlPlayerManagerProxy::loadLocationToPlayer(
 }
 
 // static
-QmlPlayerManagerProxy* QmlPlayerManagerProxy::create(QQmlEngine* pQmlEngine, QJSEngine* pJsEngine) {
+QmlPlayerManagerProxy* QmlPlayerManagerProxy::create(QJSEngine* pJsEngine, QObject* parent) {
     // The implementation of this method is mostly taken from the code example
     // that shows the replacement for `qmlRegisterSingletonInstance()` when
     // using `QML_SINGLETON`.
@@ -71,7 +71,7 @@ QmlPlayerManagerProxy* QmlPlayerManagerProxy::create(QQmlEngine* pQmlEngine, QJS
         qWarning() << "PlayerManager hasn't been registered yet";
         return nullptr;
     }
-    return new QmlPlayerManagerProxy(s_pPlayerManager, pQmlEngine);
+    return new QmlPlayerManagerProxy(s_pPlayerManager, parent);
 }
 
 } // namespace qml
--- a/src/controllers/scripting/controllerscriptenginebase.cpp
+++ b/src/controllers/scripting/controllerscriptenginebase.cpp
@@ -23,6 +23,7 @@ ControllerScriptEngineBase::ControllerScriptEngineBase(
           m_logger(logger),
           m_bAbortOnWarning(false),
 #ifdef MIXXX_USE_QML
+          m_pPlayerManager(s_pPlayerManager),
           m_bQmlMode(false),
 #endif
           m_bTesting(false) {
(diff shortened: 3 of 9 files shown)


res/controllers/engine-api.d.ts
```diff
@@ -31,6 +31,24 @@ declare interface ScriptConnection {
     readonly isConnected: boolean;
 }
 
+/** QmlPlayerProxy */
+declare interface PlayerManager {
+    isLoaded(): boolean;
+    getTrack(): string;
+    getTitle(): string;
```
Swiftb0y, line 38: I don't think these are accessible via JS because those functions are not `Q_INVOKABLE`. You should however be able to directly access the declared `Q_PROPERTY`s. So just `console.log(player.title)` for example. 

Also slight misnaming, this is not the playermanager:
```suggestion
declare interface PlayerInfo {
    isLoaded(): boolean;
    getTrack(): string;
    getTitle(): string;
```

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -31,7 +31,9 @@ constexpr double kBrakeRampToRate = 0.01;
 ControllerScriptInterfaceLegacy::ControllerScriptInterfaceLegacy(
         ControllerScriptEngineLegacy* m_pEngine, const RuntimeLoggingCategory& logger)
         : m_pScriptEngineLegacy(m_pEngine),
-          m_logger(logger) {
+          m_logger(logger),
+          m_pPlayerManagerProxy(mixxx::qml::QmlPlayerManagerProxy::create(m_pEngine->jsEngine().get(), m_pEngine))
```
christophehenry, line 35: Still struggling with C++'s pointers. Not sure about `m_pEngine->jsEngine().get()` here. I feel like ideally I should be using `.get()` which seem to defeat `std::shared_ptr`. I though I could use C++ smart pointers (`shared_ptr` and `unique_ptr`) as if they were raw pointers but no.

src/qml/qmlplayermanagerproxy.cpp
```diff
@@ -60,7 +60,7 @@ void QmlPlayerManagerProxy::loadLocationToPlayer(
 }
 
 // static
-QmlPlayerManagerProxy* QmlPlayerManagerProxy::create(QQmlEngine* pQmlEngine, QJSEngine* pJsEngine) {
+QmlPlayerManagerProxy* QmlPlayerManagerProxy::create(QJSEngine* pJsEngine, QObject* parent) {
```
christophehenry, line 63: Changing the order of arguments here to closer match the signature of `QmlPlayerManagerProxy`.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
Swiftb0y: yeah, I see how thats confusing. In this case its fine though because `create` does not take the pointer for storing (in fact, its unused, we should double check that). So reaching into the shared_ptr and taking the raw one is completely fine in this case...

src/qml/qmlplayermanagerproxy.cpp
Swiftb0y: Also note that the ordering was likely chosen deliberately because it matches the Qt example here: https://doc.qt.io/qt-6/qqmlengine.html#QML_SINGLETON

res/controllers/engine-api.d.ts
```diff
@@ -31,6 +31,24 @@ declare interface ScriptConnection {
...
+    album(): string;
+    albumArtist(): string;
+    genre(): string;
+    composer(): string;
+    grouping(): string;
+    year(): string;
+    trackNumber(): string;
+    trackTotal(): string;
+    comment(): string;
+    keyText(): string;
```
JoergAtGithub, line 49: Please add unit tests that:
-Test each function
-Ensure that the test case fails if somebody add a field to PlayerInfo. This is to ensure that nobody expose unintentional objects without implementing test and documentation

res/controllers/engine-api.d.ts
Swiftb0y: No!
afaict this would remove the `.connect(handler)`! That is a deal breaker. Moreover, we're now unnecessarily duplicating APIs again!

JoergAtGithub: We need proper abstraction of the stable API and the experimental QML code. How to ensure this without a proper wrapper?
The other issue is that we need to prevent writing to internal strings in Mixxx.

Swiftb0y: > How to ensure this without a proper wrapper?

`QmlPlayerProxy` is the wrapper.

>  The other issue is that we need to prevent writing to internal strings in Mixxx.

Again, what is "internal"? Everything exposed via `QmlPlayerProxy` should be freely readable and writable. There is no need for a second `QmlPlayerProxy` copy IMO.

Swiftb0y: > We need proper abstraction of the stable API from the experimental QML code.

Well, the code has been around long enough. Lets make it stable then... The QML controller screen feature already assumes its stable.

res/controllers/engine-api.d.ts
Swiftb0y: Thats quite a lot of work and repeated code for not a lot of gain IMO...

src/qml/qmlplayermanagerproxy.cpp
christophehenry: Hmm… Wait, it's supposed to be a singleton instance but it doesn't implement the singleton pattern. The `create` static method doesn't check the existance of a static field like it's done on https://doc.qt.io/qt-6/qqmlengine.html#QML_SINGLETON.

res/controllers/engine-api.d.ts
```diff
@@ -31,6 +31,24 @@ declare interface ScriptConnection {
     readonly isConnected: boolean;
 }
 
+/** QmlPlayerProxy */
+declare interface PlayerInfo {
+    isLoaded(): boolean;
+    track(): string;
```
daschuer, line 37: What does the string contain returened here? track() is too common to be self explaining.

res/controllers/engine-api.d.ts
```diff
@@ -31,6 +31,24 @@ declare interface ScriptConnection {
...
+    title(): string;
+    artist(): string;
+    album(): string;
+    albumArtist(): string;
+    genre(): string;
+    composer(): string;
+    grouping(): string;
+    year(): string;
+    trackNumber(): string;
+    trackTotal(): string;
```
daschuer, line 47: trackTotal() is also too common. Is it number of Tracks in the Album?

daschuer (commented):
Ups.

PR #2460, opened 2020-01-21
Prevent pushbuttons with 'blocked' controls getting stuck pressed

ronso0:
When a disallowed click is performed, like enabling Passthrough, VinylControl or TalkOver with no respective device set up, the button release is not registered because modal message boxes block interaction with Mixxx main window, and the button is 'pressed' only after the message is closed and remains so. Noticed with Shade (colored button) and LateNight (3D effect).
* click Passthrough button
* `group,passthrough` is set `1`
* message pops up, blocks release event, `group,passthrough` is forced `0`
* close popup
* button is set 'pressed', no further control update

Solution:
* set the 'pressed' property before setting any control (which could enforce a popup)
* allow `onConnectedControlChanged` to update the 'pressed' property also for pushbutton with more than one state 

I didn't discover a regression, yet, so I'm curious why this limitation was installed in the first place.?

Maybe we could also use a special pushbutton widget or a different control connection for those buttons that checks if setting the control is allowed to be toggled **before** actually doing it, which would also prevent disallowed Passthrough stopping playing decks.
But IMO the fix proposed here is the easiest way to resolve this issue.

PR #2307, opened 2019-09-30
keep the library focused when any deck controls or the overview are clicked

ronso0:
trying to fix https://bugs.launchpad.net/mixxx/+bug/1259040
"Clicking on deck controls affects library keyboard focus"

I added 'setFocusPolicy(Qt::NoFocus);' to the setup slot of most interactive skin widgets, like it worked for the library Preview button in #2264 .
This allows to keep the library focused for keyboard interaction when any deck controls or the overview are clicked.

I didn't apply the fix to the effect selector or spinboxes because I don't overlook the consequences and IMO there's no point to keep the library focused when interacting with those widgets.

This is a quite naive attempt but it works for buttons, sliders, knobs ...
It's so simple that I just can't believe there are no side effects. What can could wrong?
Is there a better place to set the focus policy?
Can it safely be set in wbasewidget ?

**New behaviour:**
* click buttons, knobs, sliders, overview: library table/sidebar remains focused
* effect selector: library is re-focused when any effect wass selected
* beat size spinboxes accept focus as before

PR #14177, opened 2025-01-16
(cleanup) remove obsolete focusOutEvent() from WPushButton

ronso0:
[Focus is disabled in setup()](https://github.com/mixxxdj/mixxx/blob/278806dfa0c77572c853e887838d5067be6d097c/src/widget/wpushbutton.cpp#L236) anyway and there is no WPushButton-derived WWidget that sets a different focus policy, so this is never called.
['pressed' and 'hovered' states are reset in event()](https://github.com/mixxxdj/mixxx/blob/278806dfa0c77572c853e887838d5067be6d097c/src/widget/wpushbutton.cpp#L433), eg. when switching to another window.
(I traced the events to verify)

So this is basically a fixup/cleanup for #2307 / #2460

### DIFF
--- a/src/widget/wpushbutton.cpp
+++ b/src/widget/wpushbutton.cpp
@@ -460,18 +460,6 @@ bool WPushButton::event(QEvent* e) {
     return WWidget::event(e);
 }
 
-void WPushButton::focusOutEvent(QFocusEvent* e) {
-    qDebug() << "focusOutEvent" << e->reason();
-    if (m_bPressed && e->reason() != Qt::MouseFocusReason) {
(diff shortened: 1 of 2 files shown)


Swiftb0y (commented):
not 100% familiar with all the implicit assumptions around Qt here. I'd be more comfortable merging this if it was targetted at `main`...

ronso0:
Sure, I don't mind.
I updated the description, the snippet I linked was the wrong one.

Swiftb0y (approved):

Swiftb0y:
Thank you.

PR #14178, opened 2025-01-16
(fix/UI) HotcueButton: show drag cursor also when hovering drag source button

ronso0:
i.e. don't show the 'drop not allowed' cursor.
Drop is still rejected.


Swiftb0y (approved):
Sure. Thank you.

PR #14179, opened 2025-01-16
drop Hotcue onto Play button to latch `play`

ronso0:
Finally implementing #9570

* (deck is stopped)
* click and hold Hotcue to preview / stutter-play
* "Oh, sounds good, I want to continue!"
* move cursor to initiate drag
* drop onto Play button
* = 'play' latched, playback continues on mouse release

Previously this was possible only if two buttons could be pressed simultaneously, ie. with controllers, with mouse and keyboard mapping or multitouch (screens).

### Note:
Only works with the new `PlayButton` (inherits WPushButton) which requires a valid `Group` set in xml (like WHotcueButton, `<Group>[Channel1]</Group>`) and can currently be tested with **LateNight** only (in decks that have hotcue buttons, so normal decks and samplers).

### TODO
- [ ] update all skins

### DIFF
--- a/src/widget/hotcuedrag.h
+++ b/src/widget/hotcuedrag.h
@@ -1,0 +1,65 @@
+#pragma once
+
+#include <QBuffer>
+#include <QMimeData>
+#include <QString>
+
+#include "track/cue.h"
+#include "track/trackid.h"
+
+const QString kDragMimeType = QStringLiteral("hotcueDragInfo");
+
+class TrackId;
+class QDragEnterEvent;
+class QDropEvent;
+
+namespace mixxx {
+
+namespace hotcuedrag {
+
+struct HotcueDragInfo {
+    constexpr HotcueDragInfo() {};
+    constexpr HotcueDragInfo(TrackId id, int cue)
+            : trackId(id),
+              hotcue(cue) {};
+
+    static HotcueDragInfo fromByteArray(const QByteArray& bytes) {
+        QDataStream stream(bytes);
+        TrackId trackId;
+        int hotcue;
+        stream >> trackId >> hotcue;
+        return HotcueDragInfo(trackId, hotcue);
+    };
+
+    QByteArray toByteArray() const {
+        QByteArray bytes;
+        QDataStream dataStream(&bytes, QIODevice::WriteOnly);
+        dataStream << trackId << hotcue;
+        return bytes;
+    };
+
+    bool isValid() const {
+        return trackId.isValid() && hotcue != Cue::kNoHotCue;
+    }
+
+    TrackId trackId = TrackId();
+    int hotcue = Cue::kNoHotCue;
+};
+
+/// These check if the event is a valid hotcue drag or drop event.
+/// Event must be a QDragEnterEvent or a QDropEvent.
+/// In case of QDropEvent, the HotcueDragInfo is extracted from the QDrag and
+/// assigned to the passed pointer. It's used by the caller (WHotcueButton)
+/// to swap hotuces.
+bool isValidHotcueDragEvent(QDragEnterEvent* pEvent,
+        const QString& group,
+        int ignoreHotcueIndex = Cue::kNoHotCue);
+bool isValidHotcueDropEvent(QDropEvent* pEvent,
+        const QString& group,
+        QObject* pTarget,
+        int ignoreHotcueIndex = Cue::kNoHotCue,
+        HotcueDragInfo* pDragData = nullptr);
+
+} // namespace hotcuedrag
+
+} // namespace mixxx
--- a/src/widget/whotcuebutton.cpp
+++ b/src/widget/whotcuebutton.cpp
@@ -1,5 +1,7 @@
 #include "widget/whotcuebutton.h"
 
+#include <widget/hotcuedrag.h>
+
 #include <QApplication>
 #include <QDrag>
 #include <QDragEnterEvent>
@@ -18,10 +20,11 @@
 
 namespace {
 constexpr int kDefaultDimBrightThreshold = 127;
-const QString kDragDataType = QStringLiteral("hotcueDragInfo");
 } // anonymous namespace
 
-WHotcueButton::WHotcueButton(const QString& group, QWidget* pParent)
+using namespace mixxx::hotcuedrag;
+
+WHotcueButton::WHotcueButton(QWidget* pParent, const QString& group)
         : WPushButton(pParent),
           m_group(group),
           m_hotcue(Cue::kNoHotCue),
@@ -185,13 +188,14 @@ void WHotcueButton::mouseMoveEvent(QMouseEvent* pEvent) {
         QDrag* pDrag = new QDrag(this);
         HotcueDragInfo dragData(id, m_hotcue);
         auto mimeData = std::make_unique<QMimeData>();
-        mimeData->setData(kDragDataType, dragData.toByteArray());
+        mimeData->setData(kDragMimeType, dragData.toByteArray());
         pDrag->setMimeData(mimeData.release());
 
         // Use the currently rendered button as dnd cursor
         // (incl. hover and pressed style).
-        // Note: for some reason, both grab() and render() render with sharp corners,
-        // ie. qss 'border-radius' is not applied to the drag image.
+        // Note: both grab() and render() use the pure rect(),
+        // i.e. these render with sharp corners and qss 'border-radius'
+        // is not visible in the drag image.
         const QPixmap currLook = grab(rect().marginsRemoved(m_dndRectMargins));
         pDrag->setDragCursor(currLook, Qt::MoveAction);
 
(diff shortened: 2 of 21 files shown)


src/widget/wpushbutton.cpp
```diff
@@ -142,10 +150,16 @@ void WPushButton::setup(const QDomNode& node, const SkinContext& context) {
     }
 
     if (leftConnection) {
+        const ConfigKey& configKey = leftConnection->getKey();
+        if (configKey.item == "play" ||
+                configKey.item == "play_stutter" ||
+                configKey.item == "start_play") {
+            m_isPlayButton = true;
+            m_group = configKey.group;
+        }
```
Swiftb0y, line 159: doing this is the class is really suboptimal. I would prefer to see an inheritance-based version that does not touch WPushButton this intrusively.

ronso0: I knew you'd say that : )

ronso0: I looked into this and the two ways to have a dedicated WPlayButton (I can think of) don't seem very elegant:
1) use `WPlayButton` in offical skins:
custom skins won't provide the drop feature
2) detect the 'play' left-click in the skin parser and instantiate a `WPlayButton`:
the WPushButton/WBaseWidget is instantiated _before_ its connections are parsed https://github.com/mixxxdj/mixxx/blob/d2e37ace7df609ed3518286123e3f96972683bbb/src/skin/legacy/legacyskinparser.cpp#L963
Even if possible, I don't think this feature justifies the (risky) refactoring the skin parser.

Do you have other ideas how to do it?

Swiftb0y: I'd definitely prefer the first one. 3rd Party skin designers may not even like this feature, so we shouldn't implement it behind their backs. If they want to opt into it, the needed changes are likely trivial.

ronso0: Alright then, I'll move this to WPlayButton :+1:

src/widget/hotcuedrag.h
```diff
@@ -0,0 +1,60 @@
...
+#include <QString>
+
+#include "track/cue.h"
+#include "track/trackid.h"
+
+const QString kDragDataType = QStringLiteral("hotcueDragInfo");
+
+class TrackId;
+
+struct HotcueDragInfo {
```
ronso0, line 14: should I move this and kDragDataType to mixxx::hotcuedrag, too?
and the implementation to cpp?

ronso0:
Note: After the refactoring this can currently only be tested with **LateNight** (in decks that have hotcue buttons, so normal decks and samplers).

src/widget/hotcuedrag.h
ronso0: Just did it.
and with `using namespace mixxx::hotcuedrag` it's all quite nice now.

src/widget/hotcuedrag.h
```diff
@@ -0,0 +1,60 @@
...
+namespace mixxx {
+
+namespace hotcuedrag {
+
+struct HotcueDragInfo {
+    HotcueDragInfo() {};
+    HotcueDragInfo(TrackId id, int cue)
+            : trackId(id),
+              hotcue(cue) {};
+
```
Swiftb0y, line 23:
```suggestion
    constexpr HotcueDragInfo() {};
    constexpr HotcueDragInfo(TrackId id, int cue)
            : trackId(id),
              hotcue(cue) {};

```

src/widget/hotcuedrag.h
```diff
@@ -0,0 +1,60 @@
...
+    static HotcueDragInfo fromByteArray(const QByteArray& bytes) {
+        QDataStream stream(bytes);
+        TrackId trackId;
+        int hotcue;
+        stream >> trackId >> hotcue;
+        return HotcueDragInfo(trackId, hotcue);
+    };
+
+    QByteArray toByteArray() {
```
Swiftb0y, line 32:
```suggestion
    QByteArray toByteArray() const {
```

src/widget/wpushbutton.h
```diff
@@ -75,7 +75,6 @@ class WPushButton : public WWidget {
     void focusOutEvent(QFocusEvent* e) override;
     void fillDebugTooltip(QStringList* debug) override;
 
-  protected:
     virtual void restyleAndRepaint();
```
Swiftb0y, line 78: revert?

Swiftb0y (commented):
couple nits, LGTM otherwise. Thank you.

src/widget/wpushbutton.h
ronso0: no, it's redundant because `protected` is already [in line 70](https://github.com/mixxxdj/mixxx/pull/14179/files#diff-8f2c55e3eeb24c3d39d5a596a249a13bd555e6877488901b74033c04614c7706R70)

src/widget/whotcuebutton.cpp
```diff
@@ -207,39 +211,30 @@ void WHotcueButton::mouseMoveEvent(QMouseEvent* pEvent) {
...
-    }
-    QByteArray mimeDataBytes = pEvent->mimeData()->data(kDragDataType);
-    if (mimeDataBytes.isEmpty()) {
-        return;
-    }
-    HotcueDragInfo dragData = HotcueDragInfo::fromByteArray(mimeDataBytes);
-    if (dragData.isValid() &&
-            dragData.trackId == pTrack->getId()) {
+    if (isValidHotcueDragOrDropEvent<QDragEnterEvent>(
+                pEvent,
```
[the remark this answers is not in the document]
ronso0: Thanks.

Interestingly this was required at a state where all worked _without_ the exlicit instantation (template function in wpushbutton.cpp). Read: I could build and test it.
Then, after a small unrelated change, it would build not anymore. Even without the change. This made my scratch my head,. and I learned about exlicit template instantation.
I was definitely not haluznating or something, so I assume this was a weird compiler/caching issue.. :shrug:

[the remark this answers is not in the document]
Swiftb0y: Yeah, I mean explicit template instantiation is quite an advanced tool. In this particular case, I'm unsure if templates are even the best solution. The problem is that the compiler assumes it can use any `T`, so I could even call `isValidHotcueDragOrDropEvent<int>(...)` and the compiler would not complain directly. Only when the linker then can't find the instantiation, the build fails. It would be better if the compiler already knew what `T`s are acceptable and which aren't. So I'd like to suggest the following additional change:

Instead of declaring the template in the header, just declared the functions with an overload. Then keep the template entirely within the `.cpp` and just pass along the arguments (and delete the explicit instantiations). The generated binary will be almost the same, but the compiler can warn you early on why `isValidHotcueDragOrDropEvent<int>` didn't work and even more importantly, what types it actually expects (the linker couldn't tell you that, you would have to manually look up which versions got explicitly instantiated). Does that make sense?

Swiftb0y:
There's still the one codestyle proposal above.

src/widget/whotcuebutton.cpp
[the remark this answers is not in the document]
ronso0: Sure, makes sense. Thought the linker warnings are enough to protect it.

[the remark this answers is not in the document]
Swiftb0y: They are to a degree, but they don't tell you _whats_ wrong, only that _something_ is wrong...

ronso0:
> There's still the one codestyle proposal above.

Done.
This also allowed some small simplification (while also causing some headaches ; )
Anyhow, it works. (and it doesn't work where it's not supposed to)

src/widget/hotcuedrag.h
```diff
@@ -0,0 +1,71 @@
...
+/// Check if the event is a valid hotcue drag or drop event.
+/// Event must be a QDragEnterEvent or a QDropEvent.
+/// In case of QDropEvent, the HotcueDragInfo is extracted and the pointer
+/// data is used by the caller (WHotcueButton) to swap hotuces.
+template<typename T>
+bool isValidHotcueDragOrDropEvent(T* pEvent,
+        const QString& group,
+        QObject* pTarget = nullptr,
+        int ignoreHotcueIndex = Cue::kNoHotCue,
+        HotcueDragInfo* pDragData = nullptr);
```
Swiftb0y, line 58: this (fortunately) can't be called anymore, so remove it from the header. 
```suggestion
```

ronso0: Okay, so I didn't get completely how namespaces work.
It's not a class so it can be defined in cpp only. Nice! Now everything makes sense.

Swiftb0y: well, its not really whether its a class or not, its about whether its accessible from other translation units. Since only the wrappers are accessible, only they need to be declared in the header.

Swiftb0y (approved):
Didn't test myself, but code LGTM. Thank you for your patience.

src/widget/hotcuedrag.h
ronso0: > its not really whether its a class or not,  its about whether its accessible from other translation units

yeah, I meant to say member function (not inline/namespace) which needs a declaration.
Thanks!

ronso0:
Sure, thanks for your help!

Will squash now.

ronso0:
Ready for merge?

Eve00000:
@ronso0 
I love this!
Can I push my luck? Have you got an eye on the work needed to make it possible to drop a hotcue on a sampler?
Told you: I pushed my luck.

ronso0:
Dropping onto another player involves track cloning.
Probably not too much work but still..

PR #14181, opened 2025-01-17
Osc in mixxx arm64 2.6 with TrackInfo (Title&Artist) -- Special OSCleton version

Eve00000:
Osc in mixxx arm64 2.6 with TrackInfo (Title&Artist) -- Special OSCleton version
Extra in this version : GetT# function to request TrackArtist & TrackTitle


JosepMaJAZ:
This is not a review, I just suggest you to do a small source reorganization.

I see that in /src/osc/ip and /src/osc/osc , there is code from the library.  This code should probably be relocated into /lib/osc
Also, /src/osc/oscreceiver.cpp/.h and /src/osc/oscfunctions.h which I assume is your code, should probably go in /src/controllers/osc

you can do that with git (i don't know if you use the commandline or some graphical application) and it will keep the history.

src/osc/oscfunctions.h
```diff
@@ -0,0 +1,246 @@
...
+// function to convert and carry special characters in TrackArtist & TrackTitle to ASCII
+QString escapeStringToJsonUnicode(const QString& input) {
+    QString escaped;
+    for (QChar c : input) {
+        if (c.isPrint() && c.unicode() < 128) {
+            // Keep printable ASCII characters as is
+            escaped += c;
+        } else {
+            // Escape non-ASCII and special characters
+            escaped += QString("\\u%1").arg(c.unicode(), 4, 16, QChar('0')).toUpper();
```
daschuer, line 36: Here memory extra alocation is involved. To be honest, I have no better idea. It would be great to know the final string size forehead and print everything into it in one go. However this requires some not too easy code. Probably not worse the effort.

daschuer (changes requested):

src/osc/oscfunctions.h
Eve00000: This pr is a special one for Arthur concerning his testing with the OSCleton, please keep the review of OSC in [pr #13835](https://github.com/mixxxdj/mixxx/pull/13835)

I will copy your remarks to #13835 and answer them there

Eve00000:
To keep the overview and avoid confusion, this PR is closed.
The maintained OSC in Mixxx PR = https://github.com/mixxxdj/mixxx/pull/14388 OSC T/B/L LIBLO

PR #14183, opened 2025-01-18
build: add missing deps for RPM systems

acolombier:
Found couple of missing dependencies when installing on Fedora 41 using a Toolbox

### DIFF
--- a/tools/rpm_buildenv.sh
+++ b/tools/rpm_buildenv.sh
@@ -53,6 +53,7 @@ case "$1" in
             lilv-devel \
             mesa-libGL-devel \
             mesa-libGLU-devel \
+            mold \
             ninja-build \
             opus-devel \
             opusfile-devel \


tools/rpm_buildenv.sh
```diff
@@ -26,6 +26,7 @@ case "$1" in
         # builddep mixxx`).
         sudo dnf install -y \
             appstream \
+            boost-devel \
```
daschuer, line 29: What is our boost dependency on Fedora? I am not aware that boost is required on Debian based distros.

daschuer (changes requested):

tools/rpm_buildenv.sh
acolombier: Turns out it is actually needed on Debian-based system too, as patched in my [Devcontainer PR](https://github.com/mixxxdj/mixxx/pull/14018/files#diff-0d882499d865cf0356240933b767b1b26ca308530f71572059f0cfd18d574c25)
Boost is a dependency for `qm-dsp`, `rigtorp` and `libdjinterop`

acolombier: I never had to install it in the past, so I suspect Ubuntu-based OSes ship it by default?

daschuer: They don't depend on boost. Only the tests of libdjinterop do. But we don't deal with the tests. We use the prebuild binaries.

daschuer (changes requested):

tools/rpm_buildenv.sh
acolombier: Tried with a fresh box on `main` and can't reproduce so this might not be needed anymore

daschuer (approved):
LGTM. Please rebase fixup

acolombier:
@daschuer Done!

PR #14188, opened 2025-01-18
mark waveformrenderer using rendergraph

m0dB:
Ports the mark waveformrenderer to a rendergraph node. This can be reviewed and merged independently from the PRs for the other waveformrenderers.


src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -5,18 +5,21 @@
...
+#include "rendergraph/material/texturematerial.h"
+#include "rendergraph/vertexupdaters/texturedvertexupdater.h"
 
 // Render digits using a texture (generated) with digits with blurred dark outline
 
+using namespace rendergraph;
+
 namespace {
 
 // The texture will contain 12 characters: 10 digits, colon and dot
```
Holzhaus, line 25:
```suggestion
 /// The texture will contain 12 characters: 10 digits, colon and dot
```
And elsewhere, so that it gets picked up by doxygen and IDEs.

src/waveform/renderers/allshader/digitsrenderer.cpp
```diff
@@ -5,18 +5,21 @@
...
 #include <cmath>
 
 #include "util/assert.h"
 #include "util/roundtopixel.h"
-#include "waveform/renderers/allshader/matrixforwidgetgeometry.h"
-#include "waveform/renderers/allshader/vertexdata.h"
+#include "rendergraph/context.h"
+#include "rendergraph/geometry.h"
+#include "rendergraph/material/texturematerial.h"
+#include "rendergraph/vertexupdaters/texturedvertexupdater.h"
```
Holzhaus, line 17: These should be before the util includes.

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: Is there a clear policy on this? IMO this is just a comment to help understand the code, not really API doc.

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: done

src/waveform/renderers/allshader/digitsrenderer.cpp
Holzhaus: I mean we don't really do API docs because we don't give any stability guarantees.

We don't really have a policy but some time ago we said that we try to use triple slash where it makes sense (i.e. if we add comments to classes, functions or constants). That makes the comment accessible to IDEs/LSP and people can see it when hovering over a method/constant, so that they have the explanation at hand instead of having to scroll to the definition.

Just to clarify: we said that we don't want the full-blown `@param` syntax (also hard to maintain), just that we add an additional slash so that we get IDE support for free.

acolombier (approved):
LGTM! 

@Holzhaus happy to merge?

src/waveform/renderers/allshader/digitsrenderer.cpp
m0dB: Ok. I prefer to do a follow-up PR to review all allshader code and change to /// where appropriate.

acolombier:
A conflict has developed following the merge of the markrange support

m0dB:
> A conflict has developed following the merge of the markrange support

Yup, this one was expected :-) Fixed! I think this can be merged.

acolombier (approved):
LGTM! Thanks :rocket: !

PR #14189, opened 2025-01-18
markrange waveformrenderer using rendergraph

m0dB:
Ports the markrange waveformrenderer to a rendergraph node. This can be reviewed and merged independently from the PRs for the other waveformrenderers.

### DIFF
--- a/src/waveform/renderers/allshader/waveformrendermarkrange.cpp
+++ b/src/waveform/renderers/allshader/waveformrendermarkrange.cpp
@@ -1,70 +1,46 @@
 #include "waveform/renderers/allshader/waveformrendermarkrange.h"
 
+#include "rendergraph/geometry.h"
+#include "rendergraph/geometrynode.h"
+#include "rendergraph/material/unicolormaterial.h"
+#include "rendergraph/vertexupdaters/vertexupdater.h"
 #include "skin/legacy/skincontext.h"
-#include "waveform/renderers/allshader/matrixforwidgetgeometry.h"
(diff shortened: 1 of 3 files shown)


acolombier (changes requested):
Looks like there is an issue with range. For example with an active loop here, the play position is going beyond the end marker 

image

m0dB:
I can't reproduce this, @acolombier , but it also don't see how this could be caused by this PR.

If you set loop points, the range is drawn exactly where you set it, right? And if you close the track and re-open it, the range is where you expect it?

If this is the case, the range code is working. The fact that your play position goes beyond. the end marker must be something else. Is it only visual or also audible?

JoergAtGithub:
It could also be that the visual playposition is wrong. The related code to determine it is here: [VisualPlayPosition::determinePlayPosInLoopBoundries](https://github.com/mixxxdj/mixxx/blob/675d4273ed372f0866da0c04d5117c449872e5b4/src/waveform/visualplayposition.cpp#L122)

m0dB:
Ping @acolombier . Can you provide more info?

acolombier:
> If you set loop points, the range is drawn exactly where you set it, right? And if you close the track and re-open it, the range is where you expect it?

No, it seems to be slightly offseted 

> Is it only visual or also audible?

 Only visual.
I'll try to provide a recording asap to help with debugging!

acolombier:
I tried again and I couldn't reproduce neither on legacy nor QML UI! No idea why this was a problem previously.

acolombier (approved):
LGTM, Thanks!

PR #14192, opened 2025-01-18
stem waveformrenderer using rendergraph

m0dB:
Ports the stem waveformrenderer to a rendergraph node. This can be reviewed and merged independently from the PRs for the other waveformrenderers.

### DIFF
--- a/src/waveform/renderers/allshader/waveformrendererstem.cpp
+++ b/src/waveform/renderers/allshader/waveformrendererstem.cpp
@@ -6,12 +6,15 @@
 
 #include "engine/channels/enginedeck.h"
 #include "engine/engine.h"
+#include "rendergraph/material/rgbamaterial.h"
+#include "rendergraph/vertexupdaters/rgbavertexupdater.h"
 #include "track/track.h"
+#include "util/assert.h"
 #include "util/math.h"
-#include "waveform/renderers/allshader/matrixforwidgetgeometry.h"
-#include "waveform/renderers/allshader/rgbdata.h"
 #include "waveform/renderers/waveformwidgetrenderer.h"
 #include "waveform/waveform.h"
+
+using namespace rendergraph;
 
 namespace allshader {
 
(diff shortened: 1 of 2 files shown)


src/waveform/renderers/allshader/waveformrendererstem.cpp
```diff
@@ -5,79 +5,103 @@
...
 #include "util/math.h"
-#include "waveform/renderers/allshader/matrixforwidgetgeometry.h"
-#include "waveform/renderers/allshader/rgbdata.h"
 #include "waveform/renderers/waveformwidgetrenderer.h"
 #include "waveform/waveform.h"
 
+using namespace rendergraph;
+
+namespace {
+constexpr int kMaxSupportedStems = 4;
```
acolombier, line 19: This constant is already duplicated twice in `engine/channels/enginedeck.cpp` and `analyzer/constants.h` (my fault). How would you feel about using either of them? analyzer probably makes the most sense here

acolombier (approved):
LGTM! Just a small nit
Just need to do testing

Edit: tested and just one nit!

src/waveform/renderers/allshader/waveformrendererstem.h
```diff
@@ -18,30 +15,31 @@ class WaveformRendererStem;
...
-    void paintGL() override;
+    bool init() override;
 
-  private:
-    mixxx::RGBAShader m_shader;
-    mixxx::TextureShader m_textureShader;
-    VertexData m_vertices;
-    RGBAData m_colors;
+    bool supportsSlip() const override {
+        return true;
```
acolombier, line 30: After testing, looks like the stem waveform doesn't support the slip animation

src/waveform/renderers/allshader/waveformrendererstem.cpp
m0dB: sure, but twice is still one too many :-)

but I don;t see the constant you are referring to in analyser

src/waveform/renderers/allshader/waveformrendererstem.h
m0dB: not sure what you mean by this. are you saying this should this return false? or that there is a bug?

acolombier: Actually I found an unrelated bug! If you have `High Details` enabled, it the slip animation doesn't work! Not introduced by your changes tho!

src/waveform/renderers/allshader/waveformrendererstem.cpp
acolombier: Actually, you are right, I have removed the duplicate in analyser! You should be able to use the one in [here](https://github.com/mixxxdj/mixxx/blob/a54051a30ad9c09381e8659b54fca2d828f79f3b/src/engine/engine.h#L14)

src/waveform/renderers/allshader/waveformrendererstem.h
m0dB: Ok, so I can ignore this? Please resolve the conversation if so.

src/waveform/renderers/allshader/waveformrendererstem.cpp
m0dB: Thanks, done!

acolombier:
Some conflict have developed - Happy to merge once resolved!

m0dB:
I have resolved this conflict, but I have one doubt: in main we use ControlProxy, and here we use PollingControlProxy. One of the conflicts was around this code. I think the change to PollingControlProxy comes from you, right?

acolombier:
Yes, that's right - a `ControlProxy` is not needed here as we don't need any signal/connection so a `PollingControlProxy` a sufficient

acolombier (approved):
LGTM, thanks!

PR #14195, opened 2025-01-19
Improvement libdjinterop : Change menu option name (more generic / less confusing)

audministrator:
Feature Description
It seems that the menu to export to Engine now says Engine Prime
This might be confusing for users who are using a Denon / Numarks controller.

image

Maybe a good idea in the next release to make is more generic.
All Denon / Numark standalone gear these day are running Engine OS embedded software.

`Engine Prime` to **Engine DJ OS**


ronso0:
I couldn't spot a commit by @audministrator here or in https://github.com/audministrator/mixxx/tree/main
Setting this to draft until it's rebased onto 2.5 and there is something to review.

PR #14196, opened 2025-01-19
Properties window position is hidden behind the taskbar in Windows

audministrator:
The properties window position is hidden behind the taskbar on my laptop when opening?
Action buttons like SAVE, are not vissible aymore ?

image

**2 possible solutions :** 
- resize the comment field to a much smaller size
- add a windows parameter to alsways be at the top of the screen instead of the bottom if in windows.
  Linux might behave differently ?

Thanks


ronso0:
Why is this a PR? (did you mean to file a bug report?)
If this is intentional, which commit is the fix?

ronso0:
I don't find a change in src/library/dlgtrackinfo.ui that might be causing this.
But we can of course experiment with the SizePolicies of `txtComment`.
Or try to fix window positioning. Like, parse screen size, subtract height of (potential top and) bottom panel (taskbar) and shrink DlgTrackInfo as much as necessary/possible so it fits in between.

PR #13602, opened 2024-08-26
(fix/UX) Track menu: Left key in Search related & Crates menus closes them again

ronso0:
If a QWidgetAction in either Search related & Crates menu is focused, Left/Right keys are translated to Up/Down which prevents closing the submenus with Left key like in other submenus.
(in the Crates menu left works as soon as the New Crate button is reached)

### Fix:
~~Translate Left to Esc to close this submenu.~~
Call hide() of the respective sumenu.

For fixing the crates menu this should actually go to 2.4 ("Search related Tracks" has QWidgetActions only in 2.5), but since we're going to release that shortly before 2.5 I think it's okay to merge the double fix to 2.5-beta

PR #14199, opened 2025-01-19
Fix memory leak and use parented_ptr in WTrackMenu

daschuer:
The widgets of the delete dialog where not properly parented and causing a memory leak. 
I have now used partented_ptr for all QObject pointers. 

This was a by-catch searching for the crasher reported in: 
https://github.com/mixxxdj/mixxx/pull/13602


Swiftb0y (approved):
LGTM. Thank you.

PR #14205, opened 2025-01-20
Preferences Effects: left/right key in effect lists trigger hide/unhide

ronso0:
Followup for #13329
Fixes #10580 

Or should this rather be Alt + Left/Right, like the track move hotkeys?

### DIFF
--- a/src/preferences/dialog/dlgprefeffects.cpp
+++ b/src/preferences/dialog/dlgprefeffects.cpp
@@ -269,6 +269,7 @@ void DlgPrefEffects::slotHideUnhideEffect() {
         return;
     }
 
+    // We emulate a drag drop to move the effect
     QMimeData* mimeData = new QMimeData;
     mimeData->setText(pManifest->uniqueId());
     // Append the selected effect to the target list
@@ -282,7 +283,11 @@ void DlgPrefEffects::slotHideUnhideEffect() {
     if (!pSourceModel->removeRows(selIdx.row(), 1, selIdx.parent())) {
         // If removing failed, undo add to target list
         pTargetModel->removeRows(pMovedEffect.row(), 1, pMovedEffect.parent());
+        return;
     }
+    // Make sure the Type column is stretched to accommodate "Built-in" in case
+    // the view was last updated with LV2 effects only.
+    pTargetList->resizeColumnToContents(0);
 }
 
 void DlgPrefEffects::slotChainPresetSelectionChanged(const QItemSelection& selected) {
--- a/src/preferences/dialog/dlgprefeffects.cpp
+++ b/src/preferences/dialog/dlgprefeffects.cpp
@@ -402,8 +402,8 @@ void DlgPrefEffects::slotDeletePreset() {
     }
 }
 
-bool DlgPrefEffects::eventFilter(QObject* object, QEvent* event) {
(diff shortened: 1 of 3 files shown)


ronso0:
TODO
mention the keys in the description above the list views. This will invalidate the existing translations though..
As a workaround I may grab the tr string from the ui file and append the new hint in the cpp.
When merging to main we may then either update the string in the ui file or move it to the cpp (would allow to painlessly append more info in the future / point releases).

daschuer:
It works, however it does not feel right, that the opposite direction works as well. Using the plain cursor keys works for me without Alt. 
What do you think?

ronso0:
Ah okay, so that Right works only in the left list, and vice versa.

I'll add that constraint.

ronso0:
I just thought that it might be expected Left Right move focus, not items. Though focus is usually moved with Tab key, but Left/Right seem to work, too, in button rows at least.

And we have the Alt+key combos to move tracks in playlists (like moving lines in text editors), but there we don't have an equivalent for Left/Right.

So to fully fix #10580 we probably need all Alt combos, Up, down, left, right, home, end. Should be rather easy for single-selection.

daschuer:
Initial expectation was to change to the right window with tab and back with shif+ tab. 
Selecting and moving with > key works greate, even as a mouse user. It was just surprising for me that the opposite direction works as well. It feels more like UnDo. Hoverver I have no idea how to implement undo reasonable. Because for bulk action, moving the focus to the next item in the same box is just right. (This is no feature request)

ronso0:
> Initial expectation was to change to the right window with tab and back with shif+ tab.

Ah okay, so we simply change the tab order:
Visible -> Hidden -> Hide -> Unhide

ronso0:
About Undo:
impIementing real Undo (view-indpendent hotkey like Ctrl+Z)  is a bit over the top here of course.
But maybe we can auto-select the last moved item when switching to the view it has been moved to (with Tab). This would auto-scroll but that would be an okayish tradeoff IMO.

daschuer:
If you like to go that route and it feels natural, fine. 
I just don't want to impose extra work in you.

ronso0:
I'll prevent Left/Right to right & left view and try the other tab order, and yes, if it feels good I'll do it here, otherwise.. "another time".

ronso0:
Done, now only Right in the left view and Left in the right view work, matching the enabled state of the GUI buttons.

ronso0:
> When removing the last effect form the right table, the focus disappears. I can't use shift-Tab to go back to the left table.

I cannot reproduce this.
Last effect moved, Shift+Tab, back in Visible Effects.
Qt 6.2.3

ronso0:
> The default Type column with is to small when opening the dialog without an effect.

Can confirm it's to small if there are only `LV2` effects and a `Built-in` effect is added.
That's strange though, we already set `setSectionResizeMode(0, QHeaderView::ResizeToContents)` for the Hidden and Visible table.
I'll test some fixes.

ronso0:
Fixed, ready to roll!

daschuer:
> Last effect moved, Shift+Tab, back in Visible Effects.

It is still an issue for me with Qt 6.2.4. But I don't mind. 
Thank you for taking care.

PR #14208, opened 2025-01-20
wait for the portaudio finished callback to avoid sporadic deadlocks

### LINKED ISSUE #14055: hanging on quit
### Bug Description

I have mixxx hanging on quit. I managed to generate a report (attached). I didn't have the time to look into it yet in detail but a quick glance points to shutting down the audio device.

[report.txt](https://github.com/user-attachments/files/18246918/report.txt)

### Version

2.5

### OS

macOS

m0dB:
This fixes #14055 . I describe the issue in detail in the bug report.

The fix consists of using [Pa_SetStreamFinishedCallback](https://www.portaudio.com/docs/v19-doxydocs/portaudio_8h.html#aa11e7b06b2cde8621551f5d527965838) to set a callback to wait for the stream to finish before continuing the main thread.

I have tested this by starting and stopping the devices programmatically in a loop. After 5000 iterations I still haven't had a deadlock, so I am confident the issue is fixed (without the fix I would get a deadlock every ~50 iterations).

Btw, there was this warning in the code (I have removed it):
```
        // BIG FAT WARNING: Pa_AbortStream() will kill threads while they're
        // waiting on a mutex, which will leave the mutex in an screwy
        // state. Don't use it!
```

This might be the same problem as described in #14055, and it is possible that with this fix we could also use Pa_AbortStream, instead of using the paAbort return value from the callback. But I didn't want to change the code more than necessary, and only add the wait condition, which is simple and safe.

### DIFF
--- a/src/soundio/sounddeviceportaudio.cpp
+++ b/src/soundio/sounddeviceportaudio.cpp
@@ -476,7 +476,7 @@ SoundDeviceStatus SoundDevicePortAudio::close() {
             // stream has become inactive.
             err = Pa_StopStream(pStream);
             if (err != paNoError) {
-                qWarning() << "PortAudio: Stop stream error:"
+                qWarning() << "PortAudio: Error while attempting to stop stream:"
                            << Pa_GetErrorText(err) << m_deviceId;
             }
         }
--- a/src/soundio/sounddeviceportaudio.cpp
+++ b/src/soundio/sounddeviceportaudio.cpp
@@ -71,6 +71,10 @@ int paV19CallbackClkRef(const void *inputBuffer, void *outputBuffer,
             (const CSAMPLE*) inputBuffer, timeInfo, statusFlags);
 }
 
+void paFinishedCallback(void* soundDevice) {
+    return ((SoundDevicePortAudio*)soundDevice)->finishedCallback();
+}
+
 const QRegularExpression kAlsaHwDeviceRegex("(.*) \\((plug)?(hw:(\\d)+(,(\\d)+))?\\)");
 
 const QString kAppGroup = QStringLiteral("[App]");
@@ -360,12 +364,14 @@ SoundDeviceStatus SoundDevicePortAudio::open(bool isClkRefDevice, int syncBuffer
     }
 #endif
 
-    // Start stream
+    // Set the finished callback, used when stopping the stream
+    m_bFinished = false;
+    Pa_SetStreamFinishedCallback(pStream, paFinishedCallback);
 
-    // Tell the callback that we are starting but not ready yet.
-    // The callback will return paContinue but not access m_pStream
-    // yet.
+    // Tell the callback to return paContinue, once running.
     m_callbackResult.store(paContinue, std::memory_order_release);
+
+    // Start stream
     err = Pa_StartStream(pStream);
     if (err != paNoError) {
         qWarning() << "PortAudio: Start stream error:" << Pa_GetErrorText(err);
@@ -407,12 +413,28 @@ bool SoundDevicePortAudio::isOpen() const {
     return m_pStream.load() != nullptr;
 }
 
+void SoundDevicePortAudio::finishedCallback() {
+    // This callback is called by PortAudio when the process callback returns
+    // paAbort to indicate the stream has to finish. This is triggered in the
+    // SoundDevicePortAudio::close() method.
+    std::unique_lock lock(m_finishedMutex);
+    m_bFinished = true;
+    m_finishedCV.notify_one();
+}
+
+void SoundDevicePortAudio::waitForFinished() {
+    // Wait until finishedCallback has been called. The timeout should never be
+    // reached, because when the process callback return paAbort, the stream
+    // will finish as soon as possible.
+    std::unique_lock lock(m_finishedMutex);
+    if (!m_finishedCV.wait_for(lock, std::chrono::seconds(1), [&] { return m_bFinished; })) {
+        qWarning() << "Timeout reached when waiting for PortAudio stream to finish";
+    }
+}
+
 SoundDeviceStatus SoundDevicePortAudio::close() {
     //qDebug() << "SoundDevicePortAudio::close()" << m_deviceId;
 
-    // Tell the callback to return paAbort. This stops the stream as soon as possible.
-    // After stopping the stream using this technique, Pa_StopStream() still has to be called.
-    m_callbackResult.store(paAbort, std::memory_order_release);
     PaStream* pStream = m_pStream.load(std::memory_order_relaxed);
     if (pStream) {
         // Make sure the stream is not stopped before we try stopping it.
@@ -430,26 +452,23 @@ SoundDeviceStatus SoundDevicePortAudio::close() {
             return SoundDeviceStatus::Error;
         }
 
-        // Stop the stream. Since the start of this function we are returning
-        // paAbort from the callback. paAbort stops the stream as soon as possible.
-        // When stopping the stream using this technique, we still need to call
-        // Pa_StopStream() before starting the stream again.
+        // Tell the process callback to return paAbort, which will cause
+        // PortAudio to stop the stream as soon as possible. This has the
+        // advantage over calling Pa_StopStream that it will not wait until all
+        // the buffers have been flushed, which can take a few (annoying)
+        // seconds when you're doing soundcard input.
+        m_callbackResult.store(paAbort, std::memory_order_release);
 
+        // We wait until the callback set with Pa_SetStreamFinishedCallback has
+        // been called. Without this, there is the risk of a deadlock when we
+        // call Pa_StopStream and Pa_CloseStream below.
+        waitForFinished();
+
+        // Call Pa_StopStream() in case the above failed.
         err = Pa_StopStream(pStream);
+
         // We should now be able to safely set m_pStream to null
         m_pStream.store(nullptr, std::memory_order_release);
-
-        // Note: With the use of return value paAbort as described above,
-        // the comment below is not relevant anymore, but I am leaving it
-        // because of the BIG FAT WARNING, just in case.
-        //
-        // Trying Pa_AbortStream instead, because StopStream seems to wait
-        // until all the buffers have been flushed, which can take a
-        // few (annoying) seconds when you're doing soundcard input.
-        // (it flushes the input buffer, and then some, or something)
-        // BIG FAT WARNING: Pa_AbortStream() will kill threads while they're
-        // waiting on a mutex, which will leave the mutex in an screwy
-        // state. Don't use it!
 
         if (err != paNoError) {
             qWarning() << "PortAudio: Stop stream error:"
--- a/src/soundio/sounddeviceportaudio.cpp
+++ b/src/soundio/sounddeviceportaudio.cpp
@@ -71,14 +71,12 @@ int paV19CallbackClkRef(const void *inputBuffer, void *outputBuffer,
             (const CSAMPLE*) inputBuffer, timeInfo, statusFlags);
 }
 
-void paFinishedCallback(void* soundDevice) {
-    return ((SoundDevicePortAudio*)soundDevice)->finishedCallback();
-}
-
 const QRegularExpression kAlsaHwDeviceRegex("(.*) \\((plug)?(hw:(\\d)+(,(\\d)+))?\\)");
 
 const QString kAppGroup = QStringLiteral("[App]");
 } // anonymous namespace
+
+void paFinishedCallback(void* soundDevice);
 
 SoundDevicePortAudio::SoundDevicePortAudio(UserSettingsPointer config,
         SoundManager* sm,
@@ -364,7 +362,7 @@ SoundDeviceStatus SoundDevicePortAudio::open(bool isClkRefDevice, int syncBuffer
     }
 #endif
 
-    // Set the finished callback, used when stopping the stream
+    // Set the finished callback. See makeStreamInactiveAndWait()
     m_bFinished = false;
     Pa_SetStreamFinishedCallback(pStream, paFinishedCallback);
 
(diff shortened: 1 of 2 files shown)


daschuer (changes requested):
Thank you for this important fix. 

The logic is hard to understand form the code and needs some improvements. 

The callback is passed to CoreAudio here: https://github.com/PortAudio/portaudio/blob/aa1cfb046f93b18db4d12986ddbff84cbaa952cb/src/hostapi/coreaudio/pa_mac_core.c#L1329
In `OpenAndSetupOneAudioUnit()`; called by Pa_StartStream()

Did you consider to lock the directly after here:
https://github.com/mixxxdj/mixxx/blob/ec3c95b795b2aa6f2e0fa48f5bdb742b86352d32/src/soundio/sounddeviceportaudio.cpp#L387

I have also not understand why a single Mutex is not sufficent? 
Idea: 
Just lock it when the stream is started and release it in the callback. Than lock it again with timeout in waitForFinished. 
After Timeout try to lock again after manually stopping. So that Pa_CloseStream is always mutex guarded. 
Release the Mutex at the very end that it is allowed to start another stream. 

If there is a reason for the Conditional variable, please explain as source code comment.  

I helps probably also to undertsand the code if you put 
m_callbackResult.store(paAbort, std::memory_order_release);
And the content form waitForFinished();
err = Pa_StopStream(pStream);
wait again ... 
into dedicated function.

m0dB:
> Thank you for this important fix.
> 
> The logic is hard to understand form the code and needs some improvements.

Can you explain what it is that is hard to understand? I think it is very straightforward:

This problem is that with:

1. we set the return variable of the callback which tells portaudio to stop processing
2. we call Pa_StopStream

-> a deadlock can occur (deep inside coreaudio) between the audio thread and the main thread because both try to terminate the stream.

The solution is inserting between 1 and 2:
- we wait for portaudio to call it's stream finished callback
 
> The callback is passed to CoreAudio here: https://github.com/PortAudio/portaudio/blob/aa1cfb046f93b18db4d12986ddbff84cbaa952cb/src/hostapi/coreaudio/pa_mac_core.c#L1329 In `OpenAndSetupOneAudioUnit()`; called by Pa_StartStream()
> 
> Did you consider to lock the directly after here:
> 
> https://github.com/mixxxdj/mixxx/blob/ec3c95b795b2aa6f2e0fa48f5bdb742b86352d32/src/soundio/sounddeviceportaudio.cpp#L387
> 
> I have also not understand why a single Mutex is not sufficent? Idea: Just lock it when the stream is started and release it in the callback. Than lock it again with timeout in waitForFinished.

Do I understand correctly that you are suggestion locking a mutex in one place and unlocking it elsewhere? I am very much against this. Mutexes should be scoped and shorted-lived. Everything else is too risky.

> After Timeout try to lock again after manually stopping. So that Pa_CloseStream is always mutex guarded. Release the Mutex at the very end that it is allowed to start another stream.

I really think that is a much more convoluted solution.

> If there is a reason for the Conditional variable, please explain as source code comment.

Yes: literally, we need to wait for a condition! The condition being: the stream finished callback has to be called in order to proceed. A conditional variable is the obvious solution.
 
> I helps probably also to undertsand the code if you put m_callbackResult.store(paAbort, std::memory_order_release); And the content form waitForFinished(); err = Pa_StopStream(pStream); wait again ... into dedicated function.

I am happy to put things in functions and add a comment, but I don't want to change this solution. It is clean and simple and well tested: I have been running stress tests, starting and stopping the engine in a loop running for hours.

Swiftb0y:
I'll have to jump in here to give my two cents. Do I understand the problem correctly that both PortAudio (via the bool in the callback) and we (via Pa_StopStream) want to close the stream concurrently, which leads to a deadlock? The proposed solution is then to serialize the attempts by first letting PortAudio do it before we try to do it as well via `Pa_StopStream`? 

If thats the correct interpretation, then the solution is to serialize the execution order of operations. While this could be done via a Mutex, its too primitive of a tool. Instead the correct tool for the job is a `barrier`. This has effectively been implemented using the condition variable.

If the goal is to improve the quality of the code, the next step would either be to use better fitting abstractions or to encapsulate the logic better. Better fitting abstractions would either be `std::binary_semaphore` or `std::barrier`. Unfortunately the semaphore has no way to protect against spurious wakeups and the barrier does not have a timeout feature. So the next logical step would be to simplify the complexity by encapsulating the barrier semantics in a small utility class. That would still provide the correct solution as implemented by @m0dB while improving the logic as @daschuer requested. Removing the condition variable and manually locking and unlocking the mutex at a bunch of different locations only increases the complexity!

daschuer:
The hard to understand part is probably, that the CV solution is not consistently used as normally, because the solution is attached to the existing port-audio solution. It is unusual to have here two variables m_callbackResult and m_bFinished for synchronization. 

My original expectation to put  a locking region around `Pa_CloseStream(pStream);` 

I have now understand how it works.  

Please also keep in mind that you fix a macOS specific issue in code that is also used for Windows and Linux with a lot of different APIs. There is risk that a fix of regressions or other problems will break your solution here. 

I will try to proposed improved comments.

daschuer:
Ah got it. If you replace m_bFinished by `Pa_IsStreamStopped()` as a CV predicate, everything becomes more robust and understandable. I will inline my comments.

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -360,12 +364,14 @@ SoundDeviceStatus SoundDevicePortAudio::open(bool isClkRefDevice, int syncBuffer
     }
 #endif
 
-    // Start stream
+    // Set the finished callback, used when stopping the stream
```
daschuer, line 367:
```suggestion
    // Set the finished callback, informs us via m_bFinishedCV when Pa_IsStreamStopped() becomes true
    // and we can safely call Pa_CloseStream() 
```

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -407,12 +413,28 @@ bool SoundDevicePortAudio::isOpen() const {
...
+void SoundDevicePortAudio::finishedCallback() {
+    // This callback is called by PortAudio when the process callback returns
+    // paAbort to indicate the stream has to finish. This is triggered in the
+    // SoundDevicePortAudio::close() method.
+    std::unique_lock lock(m_finishedMutex);
+    m_bFinished = true;
+    m_finishedCV.notify_one();
+}
+
+void SoundDevicePortAudio::waitForFinished() {
```
daschuer, line 425: This function should be inclined for a better readability.

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -407,12 +413,28 @@ bool SoundDevicePortAudio::isOpen() const {
...
+    // Wait until finishedCallback has been called. The timeout should never be
+    // reached, because when the process callback return paAbort, the stream
+    // will finish as soon as possible.
```
daschuer, line 428: // Wait until finishedCallback has been called. 
// The timeout is here as a last resort if the callback has stalled for some reasons. 

Correct?

m0dB: Yes, that is correct! (Never happened though during my stress test)

src/soundio/sounddeviceportaudio.cpp
m0dB: Sorry, what do you mean by that?

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -430,27 +452,24 @@ SoundDeviceStatus SoundDevicePortAudio::close() {
...
+        // seconds when you're doing soundcard input.
+        m_callbackResult.store(paAbort, std::memory_order_release);
 
+        // We wait until the callback set with Pa_SetStreamFinishedCallback has
+        // been called. Without this, there is the risk of a deadlock when we
+        // call Pa_StopStream and Pa_CloseStream below.
+        waitForFinished();
+
+        // Call Pa_StopStream() in case the above failed.
         err = Pa_StopStream(pStream);
```
daschuer, line 468: This should be moved into the m_finishedCV timeout branch.

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -430,27 +452,24 @@ SoundDeviceStatus SoundDevicePortAudio::close() {
...
+
         // We should now be able to safely set m_pStream to null
```
daschuer, line 470: Before, DEBUG_ASSERT(Pa_IsStreamStopped())

src/soundio/sounddeviceportaudio.cpp
m0dB: Did you confirm by looking into the portaudio code that Pa_IsStreamStopped() coincides with the finished callback? (It's what I  understand from the documentation, but confirming this would be good, if we add this in the comments)

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -430,27 +452,24 @@ SoundDeviceStatus SoundDevicePortAudio::close() {
...
-        //
-        // Trying Pa_AbortStream instead, because StopStream seems to wait
-        // until all the buffers have been flushed, which can take a
-        // few (annoying) seconds when you're doing soundcard input.
-        // (it flushes the input buffer, and then some, or something)
-        // BIG FAT WARNING: Pa_AbortStream() will kill threads while they're
-        // waiting on a mutex, which will leave the mutex in an screwy
-        // state. Don't use it!
-
         if (err != paNoError) {
```
daschuer, line 473: This condition is totally unclear for me? What is the reasoning?

src/soundio/sounddeviceportaudio.cpp
m0dB: I agree.

src/soundio/sounddeviceportaudio.cpp
daschuer: The result from     resetBlioRingBuffers( &stream->blio );

m0dB: The `if (err != paNoError)` condition? To log it if stop stream fails. But this was already here. Or are you referring to something else?

src/soundio/sounddeviceportaudio.cpp
m0dB: Yeah, I actually think calling Pa_StopStream is not needed anymore now, but I didn't want to remove it just in case.

src/soundio/sounddeviceportaudio.cpp
daschuer: Shouldn't we call  CloseStream() anyway Else we create a memory leak. At least it is called from every error in OpenStream();

src/soundio/sounddeviceportaudio.cpp
daschuer: Oh typo. I mean inlined. The function body should be closer to the other calls or the other calls can be moved here and the function renamed a bit. This will help to trace the control flow a bit better.

src/soundio/sounddeviceportaudio.cpp
daschuer: So we may do: 
```
VERIFY_OR_DEBUG_ASSERT(Pa_IsStreamStopped()) {
    err = Pa_StopStream(pStream);
}
``

Comment (m0dB) [reply src/soundio/sounddeviceportaudio.cpp]: I have come to the conclusion that this is not the case. Pa_IsStreamStopped() is not true after the finished callback has been called, at least not on macOS, and also according to the documentation, after return paAbort from the callback, the finished callback will be called when the stream becomes _inactive_

File: src/soundio/sounddeviceportaudio.cpp (hunk 6 above)
Comment (m0dB) [reply src/soundio/sounddeviceportaudio.cpp]: Ok.

**Comment by m0dB** [comment]
@daschuer I have improved the comments based on your suggestions and moved the code to set the callback return value and wait for the subsequent finished callback to `makeStreamInactiveAndWait()`

I also removed the early returns so we always end up in Pa_CloseStream even if errors occur.

I have been reading through the portaudio code and it is really hard to say if the documentation is correct and if all platforms implement the API in the same way, but I would say that our code follows the documentation and is still robust for platform differences.

**Comment by daschuer** [comment]
Thank you. Did you consider to dispose `m_bFinished` in favor of `Pa_IsStreamStopped()`? for my understanding this makes the code more easy to read and it is closer to the truth, because it is a query for the internal `m_bFinished` equivalent.

File: src/soundio/sounddeviceportaudio.cpp [hunk 7]
```diff
@@ -407,54 +411,71 @@ bool SoundDevicePortAudio::isOpen() const {
...
-        // 1 means the stream is stopped. 0 means active.
-        if (err == 1) {
-            //qDebug() << "PortAudio: Stream already stopped, but no error.";
-            return SoundDeviceStatus::Ok;
-        }
-        // Real PaErrors are always negative.
+        // 1 means the stream is stopped. 0 means active, negative means error
         if (err < 0) {
-            qWarning() << "PortAudio: Stream already stopped:"
+            qWarning() << "PortAudio: Pa_IsStreamStopped returned error:"
```
daschuer, line 462: This happens only if `PaUtil_ValidateStreamPointer()` which happens only if there is a programming error. Passing a 0 or wrong pointer type.

So we can probably use `VERIFY_OR_DEBUG_ASSERT`for it. In case of a failure every subsequent call using pStream
will fail es well. So the early return here was OK.

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -407,54 +411,71 @@ bool SoundDevicePortAudio::isOpen() const {
...
+        if (err == 1) {
+            qWarning() << "PortAudio: Stream already stopped";
+        } else {
+            // Tell the process callback to return paAbort and wait for the
+            // callback set with Pa_SetStreamFinishedCallback to be called.
+            makeStreamInactiveAndWait();
+
+            // We can now safely call Pa_StopStream, which should not block as the
+            // stream has become inactive.
+            err = Pa_StopStream(pStream);
```
daschuer, line 474: this Pa_StopStream call here is redundant in case the timeout in makeStreamInactiveAndWait() has not happened. 
So please over it into the `if (!m_finishedCV.wait_for...` branch.

src/soundio/sounddeviceportaudio.cpp
daschuer: I can confirm.

daschuer:
I have deleted some wrong comments sorry about the noise. Only one comment is left. Than we can merge.

src/soundio/sounddeviceportaudio.cpp
m0dB: Ok, thanks for going to the bottom of that. I'll make the change as suggested.

m0dB: done

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -407,54 +411,74 @@ bool SoundDevicePortAudio::isOpen() const {
...
+    // finishedCallback - as set with Pa_SetStreamFinishedCallback - has
+    // been called, using m_bFinished and conditional variable m_finishedCV.
+    // Without this, we can have a deadlock when we call Pa_StopStream
+    // later on.
+    {
+        std::unique_lock lock(m_finishedMutex);
+        if (!m_finishedCV.wait_for(lock, std::chrono::seconds(1), [&] { return m_bFinished; })) {
+            // The timeout should not be reached, because when the process
+            // callback returns paAbort, the stream will finish as soon as possible.
+            // We have it as a last result in case inactivating the stream stalls.
```
cr7pt0gr4ph7, line 447:
```suggestion
            // We have it as a last resort in case inactivating the stream stalls.
```

src/soundio/sounddeviceportaudio.cpp
```diff
@@ -407,54 +411,74 @@ bool SoundDevicePortAudio::isOpen() const {
...
+        } else {
+            // Tell the process callback to return paAbort and wait for the
+            // callback set with Pa_SetStreamFinishedCallback to be called.
+            makeStreamInactiveAndWait();
+
+            // We can now safely call Pa_StopStream, which should not block as the
+            // stream has become inactive.
+            err = Pa_StopStream(pStream);
+            if (err != paNoError) {
+                qWarning() << "PortAudio: Stop stream error:"
```
cr7pt0gr4ph7, line 479:
```suggestion
                qWarning() << "PortAudio: Error while attempting to stop stream:"
```

cr7pt0gr4ph7 (approved):
Left two small wording nitpicks, looks good to me otherwise. Haven't tested the code, though.

daschuer:
In my Ubuntu Jammy 22.04 I see ALSA and OSS as APIs 
After using ALSA I have switched to OSS and got 
```
warning [Main] PortAudio: Timeout reached when waiting for PortAudio finish callback
```
In other cases it works. 

So this kind of confirms that the workaround is required and working.  

Thank you very much.

daschuer (approved):

daschuer:
Can you commit (or squash fix if you prefer) the suggested changes of @cr7pt0gr4ph7 and than we can merge. 

This is one important 2.5 fix.  
There are still 14 pending PRs 
https://github.com/mixxxdj/mixxx/pulls?q=is%3Aopen+is%3Apr+milestone%3A2.5.1
where we need to decide of move or include. 
Any help is welcome.

m0dB:
I have committed @cr7pt0gr4ph7 ’s suggested changes (thanks!) so this can be merged. I am surprised about the timeout you get with OSS (I am in fact surprised that OSS is still around!) and I expect that that might be a bug in the portaudio OSS layer, but I don’t really worry about that, and it’s good to know the fallback with the timeout works. Do you think that before merging we should ask a windows user to do a smoketest? @JoergAtGithub maybe?

daschuer:
Oh yes, that would be helpful. Especisally 2.5 is the stable branch that is released without a beta period.

daschuer:
I have fired up my Win 10 virtual machine and it works like a flaw. Thank you very much.

PR #14209, opened 2025-01-21
Show translator file path in debug message 

daschuer:
This helps to pick the correct file for translation in case a fallback is picked.


ronso0 (approved):
LGTM
So basically this allows to check if the intended file has been loaded or if QTranslator auto-picked another 'matching' file, right?

daschuer:
Yes, right. I like to see the actual loaded file.

PR #14228, opened 2025-01-25
Qt6.8: Ensure Mixxx uses "windowsvista" Qt style on Windows

JoergAtGithub:
- Ensure Mixxx uses "windowsvista" Qt style on Windows, as our skinning looks inconsistent with Qt's "windows" or "windows11" styles
- Add logging of the selected style on all platforms

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -3,6 +3,7 @@
 #include <QPixmapCache>
 #include <QString>
 #include <QStringList>
+#include <QStyle>
 #include <QTextCodec>
 #include <QThread>
 #include <QtDebug>
@@ -142,6 +143,26 @@ void adjustScaleFactor(CmdlineArgs* pArgs) {
     }
 }
 
+void applyStyleOverride(CmdlineArgs* pArgs) {
+    if (!pArgs->getStyle().isEmpty()) {
+        qDebug() << "Default style is overwritten by command line argument "
+                    "-style"
+                 << pArgs->getStyle();
+        QApplication::setStyle(pArgs->getStyle());
+        return;
+    }
+    if (qEnvironmentVariableIsSet("QT_STYLE_OVERRIDE")) {
+        QString styleOverride = QString::fromLocal8Bit(qgetenv("QT_STYLE_OVERRIDE"));
+        if (!styleOverride.isEmpty()) {
+            // The environment variable overrides the command line option
+            qDebug() << "Default style is overwritten by env variable "
+                        "QT_STYLE_OVERRIDE"
+                     << styleOverride;
+            QApplication::setStyle(styleOverride);
+        }
+    }
+}
+
 } // anonymous namespace
 
 int main(int argc, char * argv[]) {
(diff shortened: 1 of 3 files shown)


daschuer:
Does QT_STYLE_OVERRIDE environment variable still work? I think it should if user needs to fix a corner case. 

Do we want to make the 2.5 branch also work with the 2.6 environment? I can imagine that this can be convenient for us developers.

JoergAtGithub:
No it does not work - and I do not plan to spend work for this env variable, which we never documented.
If you prefer, I retarget this to 2.5.

daschuer:
Yes. Please Target 2.5.

daschuer:
This is working for all Qt applications by default: 
https://doc.qt.io/qt-6/qapplication.html#QApplication

This code breaks it for windows. 

The main issue is that Qt prints also a warning message that a certain style is used for override. This message is wrong if we change it afterwards and should be avoided. 

It looks like `QApplication::setStyle("windowsvista");` can be set outside QApplication, before MixxxApplication(). 
This way QT_STYLE_OVERRIDE will still work and everything is fine.

src/main.cpp
```diff
@@ -142,6 +143,26 @@ void adjustScaleFactor(CmdlineArgs* pArgs) {
     }
 }
 
+void getStyleOverride(CmdlineArgs* pArgs) {
```
daschuer, line 146: This dos not rellay "get" anything. Do we have a better name?

src/main.cpp
```diff
@@ -142,6 +143,26 @@ void adjustScaleFactor(CmdlineArgs* pArgs) {
...
+    if (!pArgs->getStyle().isEmpty()) {
+        qDebug() << "Default style is overwritten by command line argument "
+                    "-style"
+                 << pArgs->getStyle();
+        QApplication::setStyle(pArgs->getStyle());
```
daschuer, line 151: Isn't this redundant?

daschuer (changes requested):

src/main.cpp
```diff
@@ -208,6 +229,21 @@ int main(int argc, char * argv[]) {
 
     MixxxApplication app(argc, argv);
 
+#if defined(Q_OS_WIN)
+    // The Mixxx style is based on Qt's WindowsVista style
+    QApplication::setStyle("windows");
```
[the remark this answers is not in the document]
JoergAtGithub: Thanks for catching `"windows"` that was my debug version accidental commited. 

Setting the style before does not work, its always overwritten by the constructor of MixxxApplication.

src/main.cpp
JoergAtGithub: No it's not. Without the branch would be out of any effect.

src/main.cpp
JoergAtGithub: Renamed to `applyStyleOverride`

src/main.cpp
```diff
@@ -208,6 +229,21 @@ int main(int argc, char * argv[]) {
 
     MixxxApplication app(argc, argv);
 
+#if defined(Q_OS_WIN)
+    // The Mixxx style is based on Qt's WindowsVista style
+    QApplication::setStyle("windowsvista");
+#endif
```
daschuer, line 235: I have just double checked on Linux using QT 6.2.4. and confirmed that you can move this block above `MixxxApplication app(argc, argv);` and remove `applyStyleOverride(&args);`

src/util/cmdlineargs.cpp
```diff
@@ -350,6 +350,15 @@ bool CmdlineArgs::parse(const QStringList& arguments, CmdlineArgs::ParseMode mod
     parser.addOption(debugAssertBreak);
     parser.addOption(debugAssertBreakDeprecated);
 
+    const QCommandLineOption styleOption(QStringLiteral("style"),
+            forUserFeedback
+                    ? QCoreApplication::translate("CmdlineArgs",
+                              "Sets the application GUI style. Possible values "
+                              "depend on your system configuration.")
```
daschuer, line 357:
```suggestion
                              "Sets the application GUI style. Possible values: %1")
                              .arg(QStyleFactory::keys().join(QStringLiteral(", ")))
```
And 
```
#include <QStyleFactory>
```
above

daschuer (changes requested):

src/main.cpp
JoergAtGithub: I tested this, it does not work on my system, MixxxApplication app(argc, argv) overrides the style every time. It works neither with Qt6.5 nor with Qt6.8.

daschuer (approved):
LGTM Thank you.

PR #14229, opened 2025-01-25
[POC] OSCQuery

daschuer:
This PR implements the a bit of the OSCQuery protocol into Mixxx. It finally allows to subscribe CO updates form remote. Somehow a mapping mechanism. See https://github.com/Vidvox/OSCQueryProposal

@Eve00000 and I have discussed this and came to the conclusion that we need two root namespaces: 
cop = Returns the parameter represenatation of a ControlObject (Controller Domain) 
cov = Returns the value presnetation of a ControlObject (Engine Domain) 

Both are duplicated as /get/cop and /get/cov to query the values from clients which do not have implemented the 
OSCQuery protocol 

This results for a compelex example to this OSC path: 
"/cop/EffectRack1/EffectUnit1/group/(Channel1Stem1)/enable"

This results in a nice tree of COs, however we put OSCQuery to the edge, because the resulting formatted JSON file is 233 MB. 

I have created a button in the developer menu to store the file in the Mixxx config folder as oscquerry.json. 

You can Open it with Firefox (after copy it to your download folder) to nicely collapse all and navigate. 
This is a good preview how it will work in client tools when selecting a Mixxx control.

### DIFF
--- a/src/dialog/dlgdevelopertools.h
+++ b/src/dialog/dlgdevelopertools.h
@@ -21,6 +21,7 @@ class DlgDeveloperTools : public QDialog, public Ui::DlgDeveloperTools {
     void slotControlSearch(const QString& search);
     void slotLogSearch();
     void slotControlDump();
+    void slotOscQuerryDump();
 
   private:
     UserSettingsPointer m_pConfig;
--- a/src/osc/query/oscquerydescription.cpp
+++ b/src/osc/query/oscquerydescription.cpp
@@ -176,10 +176,10 @@ bool OscQueryDescription::saveToFile(const QString& filePath) const {
 
 void OscQueryDescription::insertControlKey(const ConfigKey& key) {
     QString address = toOscAddress(key);
-    addAddress("/cop/" + address, "f", "3", "");
-    addAddress("/get/cop/" + address, "f", "3", "");
-    addAddress("/cov/" + address, "f", "3", "");
-    addAddress("/get/cov/" + address, "f", "3", "");
+    addAddress("/cop/" + address, "d", "3", "");
+    addAddress("/get/cop/" + address, "d", "3", "");
+    addAddress("/cov/" + address, "d", "3", "");
+    addAddress("/get/cov/" + address, "d", "3", "");
 }
 
 void OscQueryDescription::removeControlKey(const ConfigKey& key) {
--- a/src/osc/query/oscquerydescription.cpp
+++ b/src/osc/query/oscquerydescription.cpp
@@ -5,6 +5,8 @@
 #include <QJsonDocument>
 #include <QJsonValue>
 
+#include "preferences/configobject.h"
+
 namespace {
 
 template<typename T>
@@ -12,6 +14,15 @@ T take_back(std::vector<T>* pVec) {
     T last_element = std::move(pVec->back());
     pVec->pop_back();
     return last_element;
+}
+
+QString toOscAddress(const ConfigKey& key) {
+    QString groupWithoutBrackets = key.group.mid(1, key.group.size() - 2);
+    QString oscAddress = groupWithoutBrackets + QChar('/') + key.item;
+    oscAddress.replace(QChar('['), QChar('('));
+    oscAddress.replace(QChar(']'), QChar(')'));
+    oscAddress.replace(QChar('_'), QChar('/'));
+    return oscAddress;
 }
 
 } // namespace
@@ -161,4 +172,20 @@ bool OscQueryDescription::saveToFile(const QString& filePath) const {
     file.write(toJsonString().toUtf8());
     file.close();
     return true;
+}
+
+void OscQueryDescription::insertControlKey(const ConfigKey& key) {
+    QString address = toOscAddress(key);
+    addAddress("/cop/" + address, "f", "3", "");
+    addAddress("/get/cop/" + address, "f", "3", "");
+    addAddress("/cov/" + address, "f", "3", "");
+    addAddress("/get/cov/" + address, "f", "3", "");
+}
+
+void OscQueryDescription::removeControlKey(const ConfigKey& key) {
+    QString address = toOscAddress(key);
+    removeAddress("/cop/" + address);
+    removeAddress("/get/cop/" + address);
+    removeAddress("/cov/" + address);
+    removeAddress("/get/cov/" + address);
 }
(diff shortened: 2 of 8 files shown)


daschuer:
Mixxx has currenty 118899 COs tendency rising.

Eve00000:
Someone had been working a lot to, that makes me happy because I've been corking on new MixxxInTouchOSCCovCop version  all day (and will be working on it tomorrow to)

src/osc/query/oscquerydescription.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+void OscQueryDescription::insertControlKey(const ConfigKey& key) {
+    QString address = toOscAddress(key);
+    addAddress("/cop/" + address, "f", "3", "");
+    addAddress("/get/cop/" + address, "f", "3", "");
+    addAddress("/cov/" + address, "f", "3", "");
+    addAddress("/get/cov/" + address, "f", "3", "");
+}
+
```
Eve00000, line 184: what is the "f", "3", "" for?
f = float?

daschuer: All is described here, but we should make the code readable on its own. 
https://github.com/Vidvox/OSCQueryProposal
"f" = float 
"3" = rw = Read and write.

src/dialog/dlgdevelopertools.h
```diff
@@ -21,6 +21,7 @@ class DlgDeveloperTools : public QDialog, public Ui::DlgDeveloperTools {
     void slotControlSearch(const QString& search);
     void slotLogSearch();
     void slotControlDump();
+    void slotOscQuerryDump();
```
Swiftb0y, line 24: typo is repeated multiple times, please `grep` for the other occurrences
```suggestion
    void slotOscQueryDump();
```

src/osc/query/oscquerydescription.cpp
```diff
@@ -0,0 +1,190 @@
...
+    objects.reserve(pathParts.size());
+
+    // Create a list of QJsonObjects.
+    // Note: QJsonObject is only temporary helper class for writing a QJsonValue
+    for (const auto& pathPart : std::as_const(pathParts)) {
+        if (pathPart.isEmpty()) {
+            continue;
+        }
+        fullPath += "/" + pathPart;
+        if (objects.size()) {
```
Swiftb0y, line 54: implicit narrowing cast. what does this imply? this code is only executed on first iteration?

src/osc/query/oscquerydescription.h
```diff
@@ -0,0 +1,30 @@
...
+  public:
+    OscQueryDescription();
+
+    void addAddress(
+            const QString& address,
+            const QString& type,
+            const QString& access,
+            const QString& description);
+
+    void removeAddress(const QString& address);
```
Swiftb0y, line 19: the implementation of both of these is really hard to understand IMO... Can you try to simplify it?

src/osc/query/oscquerydescription.cpp
daschuer: Yes.

src/osc/query/oscquerydescription.h
daschuer: Not yet, it is still premature. Or do you see a low hanging fruit?

Swiftb0y: I suspect the algorithm would be easier to implement if addAddress and removeAddress would not mutate the json object inplace. Instead it should save it in its own internal datastructure (that is easier to understand than QJsonObject) and only saveToFile serializes that datastructure to a QJsonDocument in one go. I think doing it this way may result in simpler code. wdyt?

daschuer: Yes, that is probably true. I have started with that but than I realized that the main issue remains. You need to start from n-leaves to built the tree. That's why I started with that here. 

We finally need auch a tree to allow wildcard addressing and storing our CO objects. 
It could be interesting how this performed compared to QHash using the full address and not support wild card addressing..

@Eve00000 has implemented a code to restore the original ini name of our COs and use the established QHash. 

We neet to put all in a pot, steer and have a look. 

What do you think?

Swiftb0y: > you need to start from n-leaves to built the tree. 

I don't understand, can you elaborate?

Swiftb0y: >  can we, instead of making a hash of the complete collection of CO's implement a sort of 'favourites', I can imagine the user won't use all CO's in /tree/branch/leaf form but only a 'selection as needed'.

Not sure its worth the complexity. Certainly won't improve performance IIU your proposal correctly.

daschuer: >> you need to start from n-leaves to built the tree.
> I don't understand, can you elaborate?

Qt holds a the string like QJsonValue in memory. The QJsonObject is only an editor interface. 
If we want  a root QJsonValue with {"a":{"b":{"leaf":{}}}} you need to create the leaf first, create an empty b, insert the leaf create an empty a insert be and insert a to the root.  

If you want to add something into the leaf you cannot do via the root, because the leaf is read only. You can copy the leaf, alter it and put it back to b. however b is also read only and you need to copy it ... and so forth. 
The Root QJsonObject is only one dimensional array [1] with only a, a QJsonValue with all the rest as a string. You need to create QJsonObject for a, then get b, create the QJsonObject for leaf.

Swiftb0y: Sure... but thats orthogonal to the design I'm questioning. Why does the class manage the json tree directly instead of keeping its own (easier to manipulate) datastructure and then only build the json tree when actually writing the file?

daschuer: I had have this also in mind. We will need our own data structure anyway when we decide to allow wild card lockups. However, we always need a similar single branch up ad down crawling when writing the Json File. Since the structure is constant, it is reasonable to keep it in memory and not do it for all 118883 leaves again on every OSCQuery call. Writing the file is only a debug feature for the developers. If we have that code anyway, id does not matter in which function it is.

Swiftb0y: Ah I see, I misunderstood then. Yes in case this is called often, it does make sense to do it in place. I thought writing the file was the primary purpose of the class.

PR #14203, opened 2025-01-20
Quick fix for #14197 (`TypeError` in `midi-components-0.0.js`)

git-developer:
This PR contains a quick fix for #14197 which was introduced by #12779.

Issue #14197, opened 2025-01-19
`TypeError` since 2.5 when using MIDI controller, caused by `deepMerge()` in `applyLayer()`


git-developer:
### Problem

The following error is raised since Mixxx 2.5 when using a certain controller mapping:
```
Uncaught exception: file:///home/christian/.mixxx/controllers/common-controller-scripts.js:187:
TypeError: Cannot assign to read-only property "id"

Backtrace: @file:///home/christian/.mixxx/controllers/common-controller-scripts.js:187
deepMerge@file:///home/christian/.mixxx/controllers/common-controller-scripts.js:180
@file:///home/christian/.mixxx/controllers/common-controller-scripts.js:185
deepMerge@file:///home/christian/.mixxx/controllers/common-controller-scripts.js:180
deepMerge@file:///home/christian/.mixxx/controllers/common-controller-scripts.js:176
@file:///home/christian/.mixxx/controllers/common-controller-scripts.js:185
deepMerge@file:///home/christian/.mixxx/controllers/common-controller-scripts.js:180
@file:///home/christian/.mixxx/controllers/common-controller-scripts.js:185
deepMerge@file:///home/christian/.mixxx/controllers/common-controller-scripts.js:180
applyLayer@file:///home/christian/.mixxx/controllers/midi-components-0.0.js:664
shift@file:///home/christian/.mixxx/controllers/Behringer-Extension-scripts.js:1115
@file:///home/christian/.mixxx/controllers/midi-components-0.0.js:614
applyOperationTo@file:///home/christian/.mixxx/controllers/midi-components-0.0.js:536
forEachComponent@file:///home/christian/.mixxx/controllers/midi-components-0.0.js:548
shift@file:///home/christian/.mixxx/controllers/midi-components-0.0.js:593
inSetValue@file:///home/christian/.mixxx/controllers/Behringer-Extension-scripts.js:154
input@file:///home/christian/.mixxx/controllers/midi-components-0.0.js:184
input@file:///home/christian/.mixxx/controllers/Behringer-Extension-scripts.js:1143
input@file:///home/christian/.mixxx/controllers/Behringer-Extension-scripts.js:1310
@:1
```

### Analysis
The problem
- does not occur in 2.4.
- does no longer occur in 2.5 if `midi-components-0.0.js` and `common-controller-scripts.js` are replaced by the revisions of 2.4.
[...]

PR #14232, opened 2025-01-26
Update controller mappings for Behringer DDM4000 & BCR2000 to 2.5

git-developer:
This PR updates controller mappings for Behringer DDM4000 and BCR2000 to 2.5:

* Removal of references to lodash
* ~Removal of a quirk that was required before Qt6~
* Replacement of `var` with `const` / `let`

Lodash's `_.merge()` function performing a deep merge has been replaced. Most locations could be refactored to use a shallow merge (`Object.assign()`). The only actual requirement for a deep merge is in `Behringer-Extension-scripts.js` (for component container definitions). It is replaced by a custom implementation.

The code was tested against 2.5 using Behringer controllers DDM4000 and BCR2000.

### DIFF
--- a/res/controllers/Behringer-Extension-scripts.js
+++ b/res/controllers/Behringer-Extension-scripts.js
@@ -10,6 +10,49 @@
     var engine = global.engine;
 
     /**
+     * Determines the merge strategy for a value when merging a component container definition.
+     *
+     * @param value anything
+     * @returns {boolean} merge strategy for the value: true for 'assign', false for 'deep merge'
+     * @private
+     */
+    const doAssign = value => {
+        return typeof value !== "object"
+            || value === null
+            || value instanceof components.ComponentContainer
+            || value instanceof components.Component;
+    };
+
+    /**
+     * Merges a list of component container definitions into a target object.
+     *
+     * This is not a generic deep merge algorithm for JS objects. It does not handle circular references.
+     * If a definition contains a reference to an object instance of a component or component container
+     * (e.g. `ShiftButton.target`), the instance is taken over by reference, no deep copy is made.
+     *
+     * @param {object} targetObject Target object for the merged definitions
+     * @param {Array} sources Source definitions
+     * @returns {object} The target object
+     * @see `GenericMidiController` on component container definition
+     * @private
+     */
+    const mergeDefinitions = (targetObject, ...sources) => sources.reduce((target, source) => {
+        for (const [key, value] of Object.entries(source || {})) {
+            if (doAssign(value)) {
+                target[key] = value;
+            } else if (value !== undefined) {
+                if (Array.isArray(value) && !Array.isArray(target[key])) {
+                    target[key] = [];
+                } else if (typeof target[key] !== "object" || target[key] === null) {
+                    target[key] = {};
+                }
+                mergeDefinitions(target[key], value);
+            }
+        };
+        return target;
+    }, targetObject);
+
+    /**
      * Contains functions to print a message to the log.
      * `debug` output is suppressed unless the caller owns a truthy property `debug`.
      *
@@ -34,7 +77,7 @@
      * Determine an ID from a component's MIDI address
      *
      * @param {Array<number>} midiAddress MIDI address consisting of two integers
-     * @return {string} ID for the MIDI address; `undefined` on error
+     * @returns {string} ID for the MIDI address; `undefined` on error
      * @private
      */
     var findComponentId = function(midiAddress) {
@@ -51,7 +94,7 @@
      * Create a human-readable string to identify a component.
      *
      * @param {components.Component} component A component
-     * @return {string} A short string that describes the component; `undefined` on error
+     * @returns {string} A short string that describes the component; `undefined` on error
      * @private
      */
     var stringifyComponent = function(component) {
@@ -92,11 +135,11 @@
      *
      * @param {object} parent Constructor of parent whose prototype is used as base
      * @param {object} members Own members that are not inherited
-     * @return {object} A new prototype based on parent with the given members
+     * @returns {object} A new prototype based on parent with the given members
      * @private
      */
     var deriveFrom = function(parent, members) {
-        return _.merge(Object.create(parent.prototype), members || {});
+        return Object.assign(Object.create(parent.prototype), members);
     };
 
     /**
@@ -231,7 +274,7 @@
      * @see https://github.com/mixxxdj/mixxx/wiki/Script-Timers
      */
     var Timer = function(options) {
-        _.assign(this, options);
+        Object.assign(this, options);
         this.disable();
     };
     Timer.prototype = {
@@ -274,7 +317,7 @@
     var Throttler = function(options) {
         options = options || {};
         options.delay = options.delay || 0;
-        _.assign(this, options);
+        Object.assign(this, options);
         this.locked = false;
         this.jobs = [];
         this.unlockTimer = new Timer(
@@ -810,7 +853,7 @@
          * Create a new ComponentContainer within this registry.
          *
          * @param {string} name Name of the ComponentContainer
-         * @return {components.ComponentContainer} The ComponentContainer; `undefined` on error
+         * @returns {components.ComponentContainer} The ComponentContainer; `undefined` on error
          * @public
          */
         createContainer: function(name) {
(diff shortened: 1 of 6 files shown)


JoergAtGithub (dismissed):
Replacing the code, that calls code marked as deprecated, is a good thing. But we must not remove the deprecated code for compatibility reasons.
Backward compatibility to existing mapping scripts (this includes non-official mappings by our users) must be guranteed. Therefore please restore the deleted code in common-controller-scripts.js and midi-components-0.0.js.

git-developer:
Thanks for looking into this PR so quickly.

I agree that in general, removal of code without deprecation (warning) is a no-go because it could break existing code. But I think the situation is different here.

The functions I remove here were introduced recently in 2.5. `deepMerge()` throws runtime errors as soon as it is called (because of a missing missing `script.` prefix on recursive call). That means it is actually never called. `setLayer()` was introduced in 2.5 as variant of `applyLayer()` without dependency on lodash. But `applyLayer()` was refactored to no longer rely on lodash, so it's no longer necessary.

I'd appreciate if @Swiftb0y and @christophehenry could add their opinion on keeping or removing these two functions. They know about this topic from #14197. I will then change the code accordingly.

res/controllers/common-controller-scripts.js
```diff
@@ -158,39 +158,6 @@ const isSimpleObject = function(obj) {
...
-        return;
-    }
-
-    if (Array.isArray(target) && Array.isArray(source)) {
-        const objTarget = target.reduce((acc, val, idx) => Object.assign(acc, {[idx]: val}), {});
-        const objSource = source.reduce((acc, val, idx) => Object.assign(acc, {[idx]: val}), {});
-        deepMerge(objTarget, objSource);
-        target.length = 0;
-        target.push(...Object.values(objTarget));
-    } else if (isSimpleObject(target) && isSimpleObject(source)) {
```
christophehenry, line 180: You can remove the `isSimpleObject` function above.

christophehenry (changes requested):

Swiftb0y:
@JoergAtGithub, I agree with the assessment made by @git-developer. `setLayer()` has only been introduced recently, so the chance that mappings rely on them is very slim. `deepMerge()` seems to have been broken from the start and we have not received a report on this until now, I think that this is evidence that the number of affected mappings is negligible. Wdyt?

JoergAtGithub:
Yes, if we just added it in 2.5.0 and it's not working as intended, then I think it's best to remove it immediately in 2.5.1.

res/controllers/common-controller-scripts.js
git-developer: Thanks! Done.

git-developer:
> I think it's best to remove it immediately in 2.5.1

I agree, it's just that this PR is currently targeted at `main`. `setLayer()` does not exist in `2.5`. I can open a second PR against `2.5` and cherry-pick the appliable changes there. Would that be OK?

res/controllers/Behringer-Extension-scripts.js
```diff
@@ -4,10 +4,53 @@
...
+     * If a definition contains a reference to an object instance of a component or component container
+     * (e.g. `ShiftButton.target`), the instance is taken over by reference, no deep copy is made.
+     *
+     * @param {object} targetObject Target object for the merged definitions
+     * @param {Array} sources Source definitions
+     * @returns {object} The target object
+     * @see `GenericMidiController` on component container definition
+     * @private
+     */
+    const mergeDefinitions = (targetObject, ...sources) => sources.reduce((target, source) => {
```
christophehenry, line 39: I don't understand… You need a deep merging algorithm after all?

git-developer: Yes, 1 of 4 uses of lodash's `_.merge()` requires a deep merge, the other 3 uses are working with a shallow merge. The scope is within `Behringer-Extension-scripts.js` only. It is used to merge mapping definitions of the user with defaults from the user and defaults from the mapping. Most mapping definitions simply contain data, but the `ShiftButton` may also reference a JS object (the target that is shifted by the button). This may cause a circular reference in the mapping description that must be considered in the merge algorithm.

This is the only case I found requiring a deep merge. No commonly used location (like `applyLayer()`) is affected anymore. Due to the limited scope, verifying the behavior was feasible.

git-developer:
I removed lodash from the XML file. Now the build [fails](https://github.com/mixxxdj/mixxx/actions/runs/12982709740/job/36202891851?pr=14232#step:22:722):
```
warning [0x56179af74560] ControllerScriptHandlerBase:
"Uncaught exception: file:///home/runner/work/mixxx/mixxx/res/controllers/Behringer-DDM4000-scripts.js:6:
ReferenceError: behringer is not defined

Backtrace: %entry@file:///home/runner/work/mixxx/mixxx/res/controllers/Behringer-DDM4000-scripts.js:6"
```

The mapping works in 2.5. Does the test setup require special handling when `scriptfiles` are included?

Swiftb0y:
not really. I suspect its because of the change from `var` to `const`. Try changing line 6 back to `var` in the DDM4000 script. Add a `// eslint-next-line-disable no-var` above in case eslint complains.

christophehenry:
I would advise that you keep your changes to the Behringer scripts in a separate PR so that this one is kept atomic.

git-developer:
This PR is now focused on updating the Behringer mappings to 2.5. Removal of `deepMerge()` and `setLayer()` has been moved to #14203 (2.5) and * #14251 (`main`)

git-developer:
May I ask a related question to @christophehenry: I noticed that [Registering MIDI Input Handlers From Javascript](https://github.com/mixxxdj/mixxx/wiki/registering-midi-input-handlers-from-javascript) was implemented in #12781. Do you know an example where a `<script-binding />` element was refactored to Javascript?

The Behringer mappings heavily rely on ComponentsJS and contain hundred of lines of XML ([DDM4000](https://github.com/mixxxdj/mixxx/blob/b28a314fd5bc84063fa2f6c4c70aeb9fdccab851/res/controllers/Behringer%20DDM4000.midi.xml#L15) / [BCR2000](https://github.com/mixxxdj/mixxx/blob/b28a314fd5bc84063fa2f6c4c70aeb9fdccab851/res/controllers/Behringer%20BCR2000.midi.xml#L16)) just to delegate to a single input dispatcher function. It would be great if this could be simplified using JS.

res/controllers/Behringer-Extension-scripts.js
```diff
@@ -230,21 +273,21 @@
...
         start: function() {
             this.reset();
-            this.id = engine.beginTimer(this.timeout, function() {
+            this.id = engine.beginTimer(this.timeout, () => {
                 if (this.oneShot) {
                     this.disable();
                 }
                 this.action.call(this.owner);
-            }.bind(this), this.oneShot); // .bind(this) is required instead of arrow function for Qt < 6.2.4 due to QTBUG-95677
+            }, this.oneShot);
```
Swiftb0y, line 290: 2.5 is still technically supposed to be Qt5 compatible, I think we can't remove this yet.

Swiftb0y (commented):
Almost LGTM. I assume you have done a smoke test with the DDM or BCR?

christophehenry:
> Do you know an example where a <script-binding /> element was refactored to Javascript?

Not yet. I am working on a Denon DN-S3700 controller that will use it at some point. When I'm done, I'll propose for official inclusion.

res/controllers/Behringer-Extension-scripts.js
git-developer: OK, reverted.

git-developer:
> I assume you have done a smoke test with the DDM or BCR?

Thanks for the reminder. I did, but not after that latest changes. All global objects require `var`. I changed that accordingly and made a test _after_ the latest change, using both DDM4000 and BCR2000.

Swiftb0y (approved):
LGTM. Thank you raising the original issue and adressing it. Thank you also @christophehenry for being so responsive and brainstorming possible solutions as well. Thank you both.

res/controllers/Behringer-Extension-scripts.js
git-developer: Would it be OK to apply this removal to 2.6 in a separate PR, @Swiftb0y?

Swiftb0y: Yeah, `main`/`2.6` is supposed to be Qt6, so you can remove that in that.

PR #14237, opened 2025-01-26
make `passthrough` / `vinylcontrol_enabled`  exclusive

### LINKED ISSUE #14236: VinylControl: avoid logspam when passthrough is enabled
### Bug Description

When passthrough is enabled, there is no point in checking or visualizing the VC signal quality.

I use one input for one or two decks' passthrough, no control vinyl.
The warning is logged here:
https://github.com/mixxxdj/mixxx/blob/675d4273ed372f0866da0c04d5117c449872e5b4/src/vinylcontrol/vinylcontrolprocessor.cpp#L118
I didn't yet figure when this happens (noticed during/after performance where I couldn't debug ; )
____
Furthermore, why can `passthrough_enabled` and `vinylcontrol_enabled` be ON at the same time?
Passthrough stops the deck and VC
Both should be exclsuive and switch off the other one.

(have a feeling that I reported this years ago already, but can't find the report)

### Version

2.4++

### OS

_No response_

ronso0:
turn VC on -> passthrough off
turn passthrough on -> VC off

Passthrough stops the deck and then VC has nothing to do anyway.
And with passthrough it makes no sense to show the VC signal quality in the spinnies.
So basically this is just a small usability improvement with no harm. Or did I overlook something?

(one step to fix #14236)

### DIFF
--- a/src/engine/channels/enginechannel.cpp
+++ b/src/engine/channels/enginechannel.cpp
@@ -17,42 +17,49 @@ EngineChannel::EngineChannel(const ChannelHandleAndGroup& handleGroup,
           m_sampleBuffer(nullptr),
           m_bIsPrimaryDeck(isPrimaryDeck),
           m_active(false),
+          m_pMainMix(std::make_unique<ControlPushButton>(
+                  ConfigKey(getGroup(), QStringLiteral("main_mix")))),
+          m_pPFL(std::make_unique<ControlPushButton>(
+                  ConfigKey(getGroup(), QStringLiteral("pfl")))),
+          // crossfader assignment is persistent
+          m_pOrientation(std::make_unique<ControlPushButton>(
+                  ConfigKey(getGroup(), QStringLiteral("orientation")),
+                  true,
+                  defaultOrientation)),
+          m_pOrientationLeft(std::make_unique<ControlPushButton>(
+                  ConfigKey(getGroup(), QStringLiteral("orientation_left")))),
+          m_pOrientationRight(std::make_unique<ControlPushButton>(
+                  ConfigKey(getGroup(), QStringLiteral("orientation_right")))),
+          m_pOrientationCenter(std::make_unique<ControlPushButton>(
+                  ConfigKey(getGroup(), QStringLiteral("orientation_center")))),
+          m_pTalkover(std::make_unique<ControlPushButton>(
+                  ConfigKey(getGroup(), QStringLiteral("talkover")))),
           m_bIsTalkoverChannel(isTalkoverChannel),
           m_channelIndex(-1) {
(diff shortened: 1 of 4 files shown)


Holzhaus:
Does the current state introduce any issues? I'm wondering because maybe some people use this behavior to use a vinyl track in their set and then continue with VC without having to re-enable it again.

I think this change, because it makes sense, but I wonder if we should remember the state before enabling pass-through and the switch VC back on automatically?

ronso0:
> Does the current state introduce any issues?

Nope, except the noise of VC signal renderer.
I had only GUI users in mind, and for them it's still just one click: VC on -> pass on -> VC on ...

However, if users
* have only passthrough mapped on a controller and
* enable VC in the GUI/with hotkey once at the start of the session

they'd now have to re-enable it again. (there are VC hotkeys Ctrl+T/Y/U/I though)

The other way around, if they have VC enable mapped they now don't have to click passthrough in the GUI anymore when switching back to VC : )

Holzhaus:
@ronso0 The Roland DJ-505 for example has a 3-way PC/Line/Phono switch where the latter two are mapped to passthrough. There is no way to re-enable VC from the controller.

@Swiftb0y Can you elaborate? I'm not entirely sure how this would solve the issue at hand, but maybe I'm missing something.

ronso0:
> The Roland DJ-505 for example has a 3-way PC/Line/Phono switch where the latter two are mapped to passthrough. There is no way to re-enable VC from the controller.

Okay, so this PR would probably break more than it'd help. I'm fine with dropping it.

daschuer:
We have also discussed to have a BPM detector running on the vinyl input. Maybe related.

daschuer:
Wait, there's another selector the loaded track. 
We have anyway the visual distraction when a track is loaded, but a vinyl is playing in passthrough mode.

How about introducing an auto_passthrough preferences option that requires an empty deck for passthrough and automatically enabled it. 

There is a risk of playing time code garbage when ejecting a track VC was off, but Vinyl is still playing. 

Maybe we can use the play or cue button. To finally enable passthrough.

ronso0:
Closing this since the change is a regression.
Shall we pick the first commit, the pointer change?

ronso0:
I think @Holzhaus was referring to the flip switch on the Roland. (and with PC at one end there is no risk btw)

This has become a feature brainstorming, move to a separate topic if you like so nothing gets lost.

Swiftb0y:
> @Swiftb0y Can you elaborate? I'm not entirely sure how this would solve the issue at hand, but maybe I'm missing something.

It was rather orthogonal that in addition to making the separate CO's exclusive, we also properly encode that nature in a new CO, but that would do little to solve the usability issues discussed.

daschuer:
https://github.com/mixxxdj/mixxx/issues/14257

PR #14239, opened 2025-01-26
move Key Wheel action to View menu

ronso0:
Just to have them all in one menu.
The shortcut is `[KeyboardShortcuts], ViewMenu_ShowKeywheel` already, so .. :shrug: 

image

Should we change the hotkey to Ctrl+7 so all View items are Ctrl hotkeys? (in main)

<sub>yes, i don't see `Space` for maximize library anymore for some reason..</sub>

### DIFF
--- a/src/widget/wmainmenubar.cpp
+++ b/src/widget/wmainmenubar.cpp
@@ -315,6 +315,22 @@ void WMainMenuBar::initialize() {
             ConfigKey(kSkinGroup, QStringLiteral("show_library_coverart")));
     pViewMenu->addAction(pViewShowCoverArt);
 
+    //: menu title
+    QString keywheelTitle = tr("Show Keywheel");
+    //: tooltip text
+    QString keywheelText = tr("Show keywheel");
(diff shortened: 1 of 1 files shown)


daschuer:
Thank you. Does it fit to a the stable branch?

daschuer:
> Should we change the hotkey to Ctrl+7 so all View items are Ctrl hotkeys? (in main)

Yes.

Swiftb0y (approved):

ronso0:
> Thank you. Does it fit to a the stable branch?

IMHO it's a UX fix, so: yes.
Same for changing the hotkey: updates are done explicitly and the info should be in the changelog.
If some is used to F12 they found that in the menu. So if F12 doesn't work anymore they'll likely search the menus for Keywheel and find it in View :shrug: 

If I overlooked something, or a new hotkey is too much of a breaking change, we may as well update it in main. I don't really care.

daschuer (approved):
LGTM, OK than let's just merge.

PR #14241, opened 2025-01-26
Fix Clazy warning in main

daschuer:
See Commits.

### DIFF
--- a/src/rendergraph/common/rendergraph/uniformset.cpp
+++ b/src/rendergraph/common/rendergraph/uniformset.cpp
@@ -3,9 +3,8 @@
 using namespace rendergraph;
 
 UniformSet::UniformSet(std::initializer_list<Uniform> list, const std::vector<QString>& names) {
-    int i = 0;
-    for (auto item : list) {
-        add(Uniform{item.m_type, names[i++]});
+    for (std::size_t i = 0; i < list.size() && i < names.size(); ++i) {
+        add(Uniform{list.begin()[i].m_type, names[i]});
     }
 }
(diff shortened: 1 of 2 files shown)


src/rendergraph/common/rendergraph/uniformset.cpp
```diff
@@ -3,9 +3,8 @@
 using namespace rendergraph;
 
 UniformSet::UniformSet(std::initializer_list<Uniform> list, const std::vector<QString>& names) {
-    int i = 0;
-    for (auto item : list) {
-        add(Uniform{item.m_type, names[i++]});
+    for (std::size_t i = 0; i < list.size() && i < names.size(); ++i) {
```
Holzhaus, line 6:
```suggestion
    const auto num_items = std::min(list.size(), names.size());
    for (std::size_t i = 0; i < num_items; ++i) {
```

src/rendergraph/common/rendergraph/uniformset.cpp
```diff
@@ -3,9 +3,8 @@
...
+        add(Uniform{list.begin()[i].m_type, names[i]});
```
Holzhaus, line 7: Is `begin()` really needed here?

daschuer: Yes, see: https://en.cppreference.com/w/cpp/utility/initializer_list#Example

daschuer:
Done.

Swiftb0y (approved):
LGTM

PR #14249, opened 2025-01-27
(fix) Traktor S4mk3: set 4 decks, avoid CO warnings for decks 3/4, eg. VU meter

ronso0:
same as for the samplers: ensure all deck COs we connect to exist


daschuer (approved):
LGTM, thank you

PR #14250, opened 2025-01-27
(fix) Library scanner: update cached 'missing' flag when file is redicovered

ronso0:
Otherwise the table view may still paint the track with 'track missing' color even though it has been re-discovered.


daschuer (approved):
LGTM, thank you

Issue #14094, opened 2025-01-01
The latest mixxx-2.5dont show menu, 


GusStudio:
### Bug Description

i download the latest for win 11 and the window menu like preferences etc don't show , not able to access the preferences for the software , only the skin settings on right is working,  on the 2.4 all is fine

### Version

mixxx-2.5-beta-152-g70d7999549-win64.msi

### OS

windows 11

DjagbleyEmmanuel:
You must've inadvertently toggled it off or something. Try launching the software again, then use the Ctrl+P shortcut to open Preferences. Click on 'Interface' and scroll down to 'Miscellaneous.' You'll see the 'Menu Bar' option, where you can toggle it On or Off or enable toggling with the Alt key. Hope this helps!

Eve00000:
To make something clear: the ALT-key (normally on the left of the spacebar) is NOT the ALT-GR-key (normally on the right of the spacebar)

The menubar can't be made visible with the ALT-GR-key, only with the ALT-key.
Are you sure, you're using the ALT-key?

DjagbleyEmmanuel:
Thanks for pointing that out! You’re absolutely right—the Alt key (on the left of the spacebar) isn’t the same as the AltGr key (on the right), and I appreciate you bringing that up.
Aside from the Alt key part, the rest of the steps I mentioned should still be accurate. That said, if the Alt key isn’t working, it’s possible it’s been assigned to another function, which could cause a conflict. Unless that’s changed, pressing the Alt key might not toggle the menu bar as expected.
Thanks again for the clarification it’s super helpful!

ronso0:
Could someone with a Windows box please try to reproduce with a clean profile (empty/no mixxx.cfg)?
Is the menubar there initially?

If not we need to call `MixxxMainWindow::alwaysHideMenuBarDlg()` in the ctor, too.

Eve00000:
I tried it yesterday (to test new signed download) and just tested it again (I forgot if I tested it without Mixxx settings and cfg so I did it all again), in a clean Win10 profile. It works as a 'tierelier' (dutch word special for @ronso0  meaning: it works nice, as expected).

1 remark:
![afbeelding](https://github.com/user-attachments/assets/14c3b770-209f-43f4-85de-d6f84de1d750)
[...]

PR #14254, opened 2025-01-28
(fix) menubar: don't show 'hide' dialog when switching skins

ronso0:
don't show the dialog when switching skins while fullscreen.
furthermore, the hide dialog would block the preferences (wait cursor).

reported in https://github.com/mixxxdj/mixxx/issues/14094#issuecomment-2567461150


daschuer (approved):
LGTM, Thank you.

PR #14256, opened 2025-01-29
DlgKeywheel: Make dialog a floating tool window

fwcd:
#14239 got me thinking whether we could make the keywheel a floating dialog to make it a bit more useful, given that the keywheel is usually needed e.g. while browsing tracks. This is essentially the same style that the effect GUI dialogs proposed in #13888 use.

I've also added the tool modifier which gives it a slightly thinner window bar, at least on macOS, where this nicely follows the convention laid out by the OS color picker:

| Before | After | OS color picker (for comparison) |
| --- | --- | --- |
|  |  |  |

### DIFF
--- a/src/dialog/dlgkeywheel.ui
+++ b/src/dialog/dlgkeywheel.ui
@@ -19,6 +19,9 @@
   <property name="windowTitle">
    <string>Keywheel</string>
   </property>
+  <property name="windowFlags">
+   <set>Qt::Tool|Qt::WindowStaysOnTopHint</set>
+  </property>
   <property name="layoutDirection">
    <enum>Qt::LeftToRight</enum>
   </property>


daschuer (approved):
LGTM, Thank you.

ronso0:
Beside the different style I don't see a difference, Keywheel is always on top ofr me (Linux, Qt 6.2.3)
Was that not the case on macOS?

fwcd:
Nope, wasn't the case before (just like none of of the other dialogs float, including things like preferences etc.). IIRC preferences don't float on Linux either, so that surprises me

ronso0:
Hmmkay, for me all dialogs float, i.e. stay on top no matter which window I activate :man_shrugging: 
this is Ubuntu with xfce desktop

PR #14260, opened 2025-01-30
(fix) DlgControllerLearning: make control box completer usable

### LINKED ISSUE #14258: Make MIDI Wizard text search fuzzy
### Feature Description

Currently, to get a result using the text search method, one needs to type out the control name from it's initial letter, matching it character by character.

It would be more practical if this was a fuzzy search interface instead, as in [fzf](https://github.com/junegunn/fzf) .

ronso0:
The implementation was complete but not very usable because some defaults were used.
* use `MatchContains` (default is `Qt::MatchStartsWith`
* set case-insensitive (default is `Qt::CaseSensitive`)

:tada:


Swiftb0y (approved):
LGTM. Thank you.

daschuer (commented):
LGTM, I can confirm the issue and the fix. Thank you.

daschuer (approved):

PR #14264, opened 2025-01-31
Building without tests-tools

PPN-SD:
Today, we can't build mixxx without links to gtest/gperftools/benchmark

This PR allows building only mixxx and mixxx-lib targets without tests-tools.

Two common cmake-definitions are added :
* BUILD_TESTING
* BUILD_BENCH

And one compile definition for test.main.cpp :
* USE_BENCH

This is helpful for packaging (gentoo here).

The negative effect is sorting of tests source files as soon as they are called by mixxx-benchmark target.

### DIFF
--- a/res/qml/EqColumn.qml
+++ b/res/qml/EqColumn.qml
@@ -19,7 +19,7 @@ Column {
     }
 
     function stemGroup(group, index) {
-        return `${group.substr(0, group.length-1)}Stem${index + 1}]`
+        return `${group.substr(0, group.length-1)}_Stem${index + 1}]`
     }
 
     Row {
--- a/res/skins/LateNight/stem_channel.xml
+++ b/res/skins/LateNight/stem_channel.xml
@@ -1,5 +1,5 @@
 <Template>
-  <SetVariable name="StemGroup">[Channel<Variable name="ChanNum"/>Stem<Variable name="StemNum"/>]</SetVariable>
+  <SetVariable name="StemGroup">[Channel<Variable name="ChanNum"/>_Stem<Variable name="StemNum"/>]</SetVariable>
 
 	<WidgetGroup>
 		<ObjectName>StemChannel_ControlContainer</ObjectName>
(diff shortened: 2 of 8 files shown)


CMakeLists.txt
```diff
@@ -4248,6 +4267,13 @@ if(NOT WIN32)
   )
 endif()
 
+if(BUILD_BENCH)
+  if(NOT BUILD_TESTING)
```
daschuer, line 4271: For a bit less headaches: (need to be moved before all other checks.
```suggestion
  if(NOT DEFINED BUILD_TESTING)
    set(BUILD_TESTING ON)
  endif()
  if(NOT BUILD_TESTING)   
```

daschuer (changes requested):
Thank you. 

I can confirm with VCPKG: 
- BUILD_TEST 16
- BUILD_TESTS 222
- BUILD_TESTING 218

BUILD_TESTING is fine. 

The indentation of the file is not correct. I wonder why pre-commit is not complaining.

daschuer:
This changes the default behavior form BUILD_TESTING=ON to BUILD_TESTING=OFF which will required manual work to continue as before. 
Instead I think it could be reasonable to auto set BUILD_TESTING when Google Test is found, using `cmake_dependent_option()`

PPN-SD:
> This changes the default behavior form BUILD_TESTING=ON to BUILD_TESTING=OFF which will required manual work to continue as before. Instead I think it could be reasonable to auto set BUILD_TESTING when Google Test is found, using `cmake_dependent_option()`

Do you want BENCH by default too ?

daschuer:
For the sake of a quick merge, yes. We may consider to change the default at one point, because I think benchmarking does not make sense without explicit looking into the results.

If the default behavior does not change this can even go to the 2.5 branch for the upcoming 2.5.1 release. 
Does it help In you case?

PPN-SD:
> If the default behavior does not change this can even go to the 2.5 branch for the upcoming 2.5.1 release. Does it help In you case?

Ok. I'll rebase the patch as soon as you're agree with the master one.

PPN-SD:
Damn, I forgot to create a branch. Create a new PR.

Issue #14265, opened 2025-01-31
`MidiController::makeInputHandler` rejects `midino` 0


git-developer:
### Bug Description

Some MIDI controllers send commands having `midino == 0`, e.g. [Behringer DDM4000](https://github.com/mixxxdj/mixxx/blob/b28a314fd5bc84063fa2f6c4c70aeb9fdccab851/res/controllers/Behringer%20DDM4000.midi.xml#L21).

#### Observed behavior
These commands are rejected by an explicit check in [`MidiController::makeInputHandler`](https://github.com/mixxxdj/mixxx/blob/531ad2dbf405865327baa498d30f755a7dd03943/src/controllers/midi/midicontroller.cpp#L625).

#### Expected behavior
These commands are accepted by `MidiController::makeInputHandler`.

### Version

2.5.0

### OS

Ubuntu 24.04

PR #14266, opened 2025-01-31
Fix: Allow `midino` 0 in `MidiController::makeInputHandler`

git-developer:
Fix for #14265

### DIFF
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -607,7 +607,9 @@ void MidiController::processInputMapping(const MidiInputMapping& mapping,
                          << MidiUtils::formatSysexMessage(getName(), data, timestamp);
 }
 
-QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue& scriptCode) {
+QJSValue MidiController::makeInputHandler(unsigned char status,
+        unsigned char control,
+        const QJSValue& scriptCode) {
     auto pJsEngine = getScriptEngine()->jsEngine();
     VERIFY_OR_DEBUG_ASSERT(pJsEngine) {
         return QJSValue();
@@ -622,28 +624,28 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
         return QJSValue();
     }
 
-    if (status <= 0 || midino < 0) {
+    if (status < 0x80 || control > 0x7F) {
         auto mStatusError = QStringLiteral(
-                "Invalid status or midino passed to midi.makeInputHandler. "
-                "Please pass a strictly positive integer. status=%1,midino=%2")
+                "Invalid status or control passed to midi.makeInputHandler. "
+                "Please pass status >= 0x80 and control <= 0x7F. status=%1,control=%2")
                                     .arg(status)
-                                    .arg(midino);
+                                    .arg(control);
 
         getScriptEngine()->throwJSError(mStatusError);
         return QJSValue();
     }
 
-    const auto midiKey = MidiKey(status, midino);
+    const auto midiKey = MidiKey(status, control);
 
     auto it = m_pMapping->getInputMappings().constFind(midiKey.key);
     if (it != m_pMapping->getInputMappings().constEnd() &&
             it.value().options.testFlag(MidiOption::Script) &&
             std::holds_alternative<ConfigKey>(it.value().control)) {
         qCWarning(m_logBase) << QStringLiteral(
-                "Ignoring anonymous JS function for status=%1,midino=%2 "
+                "Ignoring anonymous JS function for status=%1,control=%2 "
                 "because a previous XML binding exists")
                                         .arg(status)
-                                        .arg(midino);
+                                        .arg(control);
         return QJSValue();
     }
 
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -622,7 +622,7 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
         return QJSValue();
     }
 
-    if (status <= 0 || midino <= 0) {
+    if (status <= 0 || midino < 0) {
         auto mStatusError = QStringLiteral(
                 "Invalid status or midino passed to midi.makeInputHandler. "
                 "Please pass a strictly positive integer. status=%1,midino=%2")
--- a/src/test/midicontrollertest.cpp
+++ b/src/test/midicontrollertest.cpp
@@ -635,10 +635,17 @@ TEST_F(MidiControllerTest, JSInputHandler_ControllerShutdownSlot) {
     m_pController->setMapping(m_pMapping->clone());
     EXPECT_EQ(getControllerMapping()->getInputMappings().count(), 0);
     evaluateAndAssert(
-            "midi.makeInputHandler(0x90, 0x43, (channel, control, value, status) => {"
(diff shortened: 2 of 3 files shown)


src/controllers/midi/midicontroller.cpp
```diff
@@ -622,7 +622,7 @@ QJSValue MidiController::makeInputHandler(int status, int midino, const QJSValue
         return QJSValue();
     }
 
-    if (status <= 0 || midino <= 0) {
+    if (status <= 0 || midino < 0) {
```
daschuer, line 625: Both can only be a unsigned char. I wonder why the function even allows int. What would happen if one passes -1 form the js code if this would be unsigned char? 
Can't status be 0? I think from the Mixxx perspective it is allowed and we are not the midi police.

git-developer: I can't comment on that due to my limited knowledge. If you're right, the check would be obsolete. Maybe @christophehenry can help out.

src/test/midicontrollertest.cpp
```diff
@@ -635,10 +635,17 @@ TEST_F(MidiControllerTest, JSInputHandler_ControllerShutdownSlot) {
...
     shutdownController();
     EXPECT_EQ(getControllerMapping()->getInputMappings().count(), 0);
 }
+
+TEST_F(MidiControllerTest, JSInputHandler_ErrorWhenMidinoIsNegative) {
+    m_pController->setMapping(m_pMapping->clone());
+    EXPECT_EQ(getControllerMapping()->getInputMappings().count(), 0);
+    bool isError = evaluateAndAssert(
+            "midi.makeInputHandler(0x90, -1, (channel, control, value, status) => {})");
+    ASSERT_TRUE(isError);
```
Swiftb0y, line 649:
```suggestion
    bool exception = evaluateAndAssert(
            "midi.makeInputHandler(0x90, -1, (channel, control, value, status) => {})");
    ASSERT_TRUE(exception.isError());
```

Swiftb0y (approved):
Thank you. Pesky off-by-one errors.

src/controllers/midi/midicontroller.cpp
christophehenry: Hmm… I don't remember. I think it was because passing -1 to unsigned char doesn't raise an exception in JS engine but crashes the app which is not something we want.

christophehenry: I can try again. But I remember something like a hard crash of Mixxx when trying to pass a bad value to a C++ with C++ types. Maybe it was when working on #13772. I don't remember exactly.

git-developer: Would it be possible to use `QJSValue::`[toUInt](https://doc.qt.io/qt-5/qjsvalue.html#toUInt), letting Qt take care of negative values?

git-developer: As I stated before, my knowledge in this area is quite limited, and I'd like to make sure that I understand your discussion.

My view is: the conversion between JS and C++ values is done within the QJSEngine. In JS, `midino` might be anything: an integer or fractional number, or even a string, an object or a function. In CPP, the type is declared as `int`, therefore Qt tries to convert _whatever_ into an `int`. This conversion is part of Qt, not Mixxx. So when the value is inconvertible, it's Qt's job to handle this some how. Might be an exception or a default value, I don't know. For example [QScript::ToUInt16](https://github.com/qt/qt/blob/0a2f2382541424726168804be2c90b91381608c6/src/script/api/qscriptengine.cpp#L395) (which might be irrelevant here) never raises an explicit error.

I suppose that a "strictly typed API" would mean that we replace `int` by `unsigned char` - the CPP type that matches the domain best (perfect would be a type that supports 0..127 only). We delegate the conversion to Qt in this case and don't need to care about negative values ourselves.

Assuming that all this is correct, I did not understand Swiftb0y's explanation about the downsides of this approach. Why should we use an `int` and do the checks manually here? Is Qt's conversion behavior not appropriate in error cases?

Swiftb0y: Okay, so basically if we wanted to do the validation ourselves, we could just take a QJSValue, that has three primary downsides:
1. Unnecessary overhead when invoking these APIs directly. You can also access a `Q_PROPERTY` or a `Q_INVOKABLE` directly from within C++ (we don't do this currently, but we could). In this case, Qt conversion doesn't kick in (its just a normal C++ function call). Going through some type-erased intermediate (like QJSValue) increases overhead and complexity. 
1. The function signatures get very unspecific. Its much harder to understand the API by just looking at the header (the header doesn't document what value is actually expected to be contained in the `QJSValue`)
1. The functions themselves get very long because they need to do lots of tedious and repetitive validation when casting from the generic type-erased (QJSValue) to the value they actually expect.

> I suppose that a "strictly typed API" would mean that we replace int by unsigned char - the CPP type that matches the domain best (perfect would be a type that supports 0..127 only). We delegate the conversion to Qt in this case and don't need to care about negative values ourselves.

Yes thats basically what I mean.

> Why should we use an int and do the checks manually here? Is Qt's conversion behavior not appropriate in error cases?

Yes, IMO Qt's conversion is not appropriate in the error case. The problem in this particular case is that its very easy for users to accidentally pass bogus values (outside `char` value range) and because Qts conversion automatically overflows we have no way to warn them (or detect) that they just passed us an invalid value. This issue primarily applies to number-like types, I think for Qt does actually emit warnings itself for more complex type-mismatches.

daschuer: The internal "number" type of JS is double. 
A float would not able to express all 32 bit int numbers. 

I don't think that we have here a situation that we need to catch particular scripting issues, other invocables don't have. Sure the user can pass a wrong value. It can be wrong inside the int 32 range or the double range, or just the wrong 8 bit number. The check is anyway not checking the 0x7F limit or such. 

There is also no crash for a crasher. Result the binding does not work. 

So let's be informative (and correct) with the data types in the function signature and use unsigned char.

Swiftb0y: I don't particularly care, though if you really wanna go that route, `std::byte` may be more appropriate since we don't want to do arithmatic on the `midino`. Regardless, we still need a check for `< 128`. That may also be able to catch some primitive overflows.

git-developer: The code is updated.

I used `std::byte` first, but then saw that all other occurrences of `status` and `control` are typed as `unsigned char` in `midicontroller.{h,cpp}`, and thus decided for unity. I will change to `std::byte` if you insist.
Validation was updated, too, please review, I'm not sure if that's "midi police" now.

git-developer: (According to the [MIDI spec](https://midi.org/expanded-midi-1-0-messages-list): a MIDI status byte requires the leading bit to be `1`, and in data bytes the leading bit must be `0`.)

Swiftb0y: > I used std::byte first, but then saw that all other occurrences of status and control are typed as unsigned char in midicontroller.{h,cpp}, and thus decided for unity. I will change to std::byte if you insist.

Don't really care, though `std::byte` is the newer type. The only reason the others use `unsigned char` is because `std::byte` didn't exist when that code was written.

Swiftb0y (approved):
LGTM. Thank you.

daschuer (approved):
LGTM, Thank you.

PR #14269, opened 2025-01-31
Building without tests-tools 2.5

PPN-SD:
Today, we can't build mixxx without links to gtest/gperftools/benchmark

This PR allows building only mixxx and mixxx-lib targets without tests-tools.

Two common cmake-definitions are added : 
* BUILD_TESTING
* BUILD_BENCH

And one compile definition for test.main.cpp :
* USE_BENCH

This is helpful for packaging (gentoo here).

The negative effect is sorting of tests source files as soon as they are called by mixxx-benchmark target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2061,156 +2061,194 @@ endif()
 # Tests
 #
 
-add_executable(mixxx-test
-  src/test/analyserwaveformtest.cpp
-  src/test/analyzersilence_test.cpp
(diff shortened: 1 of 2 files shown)


daschuer:
Ups, I have merged the "main" version. We regularly merge 2.5 to main. So the main version should have been rebase onto 2.5. But never mind, I merge this now and resolve conflicts if there are any.

Issue #10811, opened 2022-08-23
Move beat grid half beat


mixxxbot:
Reported by: **[leontheoriginal](https://launchpad.net/~leontheoriginal)**
Date: 2022-08-16T00:29:05Z
Status: Confirmed
Importance: Wishlist
Launchpad Issue: [lp1986596](https://bugs.launchpad.net/bugs/1986596)
Tags: beatgrid, easy

---

Sometimes, the beat is detected exactly a half bar off. To correct this (while another track is playing), at the moment you have to disable sync lock (if enabled), beatshift a half beat and right click the beatgrid adjust button.
This also won't work when a loop is active as the beatjump control will move the loop boundaries instead, so you have to exit the loop before, making the procedure even more complicated.
There is already an option to move the grid itself, but this is not helpful as it is only intended for small corrections, no big, exact offsets.
It would be very helpful to have an option to offset the grid with a single click/tap (similar to doubling/halving BPM or using beatjump).

MetriSomesh:
I wanted to work on this issue, but I don't understand how to test that the beat is a half bar off

ronso0:
The request is about adding a ControlObject (for push buttons) to trigger the action manually, not about improving the (external) beat detection algorithms.

This requires some discussion how and where to expose this control in the GUI.

MetriSomesh:
I meant I wanted to test the issue to understand it more efficiently. I know that he has mentioned to add a ControlObject.

ronso0:
Just analyze a track and manually set the detected beat grid 1/2 beat off?

MetriSomesh:
yeah I understood thank you very much.

ronso0:
This is the place.
Do you have an IDE set up? If yes, simply look into how the existing beatgrid actions are working, `beats_translate_later` `beats_translate_earlier`, and if it's possible to get the (local?¹) BPM / beat distance.
Thinking about it, this CO probably only makes sense for (constant) beat grids.

VARUN-009:
I wanna work on this issue. Please assign this to me and since I'm new to this repo so I request you to please guide me through.

daschuer:
We do not use the issue assign feature. What do you like to know?

neopostmodern:
[...]

PR #14279, opened 2025-02-01
slotTranslateBeatsHalf: Move beatgrid half beat

jstolberg:
Function to translate beat grid half a beat.
Implementing this: [Move beat grid half beat #10811](https://github.com/mixxxdj/mixxx/issues/10811)

Invoked by right click on Adjust Beats buttons.

slotTranslateBeatsMove adjusted for precision and coherence:
Removed rounding of input, as it invoked imprecision when
moving beat grid half a bar.
Removed the division by number of channels of the frames,
as the framerate is independent of number of channels.

### DIFF
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -110,6 +110,13 @@ BpmControl::BpmControl(const QString& group,
             this,
             &BpmControl::slotTranslateBeatsLater,
             Qt::DirectConnection);
+    m_pTranslateBeatsHalf = std::make_unique<ControlPushButton>(
+            ConfigKey(group, "beats_translate_half"), false);
+    connect(m_pTranslateBeatsHalf.get(),
+            &ControlObject::valueChanged,
+            this,
+            &BpmControl::slotTranslateBeatsHalf,
+            Qt::DirectConnection);
     m_pTranslateBeatsMove = std::make_unique<ControlEncoder>(
             ConfigKey(group, "beats_translate_move"), false);
     connect(m_pTranslateBeatsMove.get(),
@@ -315,8 +322,20 @@ void BpmControl::slotTranslateBeatsLater(double v) {
     slotTranslateBeatsMove(1);
 }
 
+void BpmControl::slotTranslateBeatsHalf(double v) {
+    if (v <= 0) {
+        return;
+    }
+    const TrackPointer pTrack = getEngineBuffer()->getLoadedTrack();
+    if (!pTrack) {
+        return;
+    }
+    const double bpm = pTrack->getBpm();
+    const double halfBeat = 3000.0 / bpm;
+    slotTranslateBeatsMove(halfBeat);
+}
+
 void BpmControl::slotTranslateBeatsMove(double v) {
-    v = std::round(v);
     if (v == 0) {
         return;
     }
@@ -327,9 +346,7 @@ void BpmControl::slotTranslateBeatsMove(double v) {
     const mixxx::BeatsPointer pBeats = pTrack->getBeats();
     if (pBeats) {
         // TODO(rryan): Track::frameInfo is possibly inaccurate!
-        const double sampleOffset = frameInfo().sampleRate * v * 0.01;
-        const mixxx::audio::FrameDiff_t frameOffset =
-                sampleOffset / mixxx::kEngineChannelOutputCount;
+        const double frameOffset = frameInfo().sampleRate * v * 0.01;
         const auto translatedBeats = pBeats->tryTranslate(frameOffset);
         if (translatedBeats) {
             pTrack->trySetBeats(*translatedBeats);
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -331,12 +331,14 @@ void BpmControl::slotTranslateBeatsHalf(double v) {
         return;
     }
     const mixxx::BeatsPointer pBeats = pTrack->getBeats();
-    if (pBeats) {
-        const auto translatedBeats = pBeats->tryTranslateBeats(0.5);
-        if (translatedBeats) {
-            pTrack->trySetBeats(*translatedBeats);
-        }
+    if (!pBeats) {
+        return;
     }
+    const auto translatedBeats = pBeats->tryTranslateBeats(0.5);
+    if (!translatedBeats) {
+        return;
+    }
+    pTrack->trySetBeats(*translatedBeats);
 }
 
 void BpmControl::slotTranslateBeatsMove(double v) {
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -322,6 +322,7 @@ void BpmControl::slotTranslateBeatsLater(double v) {
     slotTranslateBeatsMove(1);
 }
 
+// slotTranslateBeatsHalf works only with constant BPM tracks
 void BpmControl::slotTranslateBeatsHalf(double v) {
     if (v <= 0) {
         return;
--- a/src/engine/controls/bpmcontrol.cpp
+++ b/src/engine/controls/bpmcontrol.cpp
@@ -330,12 +330,17 @@ void BpmControl::slotTranslateBeatsHalf(double v) {
     if (!pTrack) {
         return;
     }
-    const double bpm = pTrack->getBpm();
-    const double halfBeat = 3000.0 / bpm;
-    slotTranslateBeatsMove(halfBeat);
+    const mixxx::BeatsPointer pBeats = pTrack->getBeats();
+    if (pBeats) {
+        const auto translatedBeats = pBeats->tryTranslateBeats(0.5);
+        if (translatedBeats) {
+            pTrack->trySetBeats(*translatedBeats);
+        }
+    }
 }
 
 void BpmControl::slotTranslateBeatsMove(double v) {
+    v = std::round(v);
     if (v == 0) {
         return;
     }
@@ -346,7 +351,9 @@ void BpmControl::slotTranslateBeatsMove(double v) {
     const mixxx::BeatsPointer pBeats = pTrack->getBeats();
     if (pBeats) {
         // TODO(rryan): Track::frameInfo is possibly inaccurate!
-        const double frameOffset = frameInfo().sampleRate * v * 0.01;
+        const double sampleOffset = frameInfo().sampleRate * v * 0.01;
+        const mixxx::audio::FrameDiff_t frameOffset =
+                sampleOffset / mixxx::kEngineChannelOutputCount;
         const auto translatedBeats = pBeats->tryTranslate(frameOffset);
         if (translatedBeats) {
             pTrack->trySetBeats(*translatedBeats);
--- a/src/track/beats.cpp
+++ b/src/track/beats.cpp
@@ -1,7 +1,9 @@
 #include "track/beats.h"
 
 #include <cmath>
+#include <cstddef>
 #include <iterator>
+#include <optional>
 #include <unordered_map>
 #include <vector>
 
@@ -24,6 +26,7 @@ constexpr double kEpsilon = 0.01;
 } // namespace
 
 namespace mixxx {
+using std::transform;
 
 mixxx::audio::FrameDiff_t Beats::ConstIterator::beatLengthFrames() const {
     if (m_it == m_beats->m_markers.cend()) {
@@ -628,6 +631,20 @@ std::optional<BeatsPointer> Beats::tryTranslate(audio::FrameDiff_t offsetFrames)
 
     const auto lastMarkerPosition = m_lastMarkerPosition + offsetFrames;
     return BeatsPointer(new Beats(markers,
+            lastMarkerPosition.toLowerFrameBoundary(),
+            m_lastMarkerBpm,
+            m_sampleRate,
+            m_subVersion));
+}
+
+std::optional<BeatsPointer> Beats::tryTranslateBeats(double xBeats) const {
+    if (!m_markers.empty()) {
+        return std::nullopt;
+    }
+    const mixxx::audio::FrameDiff_t lastOffsetFrames =
+            xBeats * m_sampleRate.value() * 60.0 / m_lastMarkerBpm.value();
+    const auto lastMarkerPosition = m_lastMarkerPosition + lastOffsetFrames;
+    return BeatsPointer(new Beats(m_markers,
             lastMarkerPosition.toLowerFrameBoundary(),
             m_lastMarkerBpm,
             m_sampleRate,
--- a/src/track/beats.cpp
+++ b/src/track/beats.cpp
@@ -1,9 +1,7 @@
 #include "track/beats.h"
 
 #include <cmath>
-#include <cstddef>
 #include <iterator>
-#include <optional>
 #include <unordered_map>
 #include <vector>
 
@@ -26,7 +24,6 @@ constexpr double kEpsilon = 0.01;
 } // namespace
 
 namespace mixxx {
-using std::transform;
 
 mixxx::audio::FrameDiff_t Beats::ConstIterator::beatLengthFrames() const {
     if (m_it == m_beats->m_markers.cend()) {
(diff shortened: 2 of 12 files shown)


src/engine/controls/bpmcontrol.cpp
```diff
@@ -327,9 +346,7 @@ void BpmControl::slotTranslateBeatsMove(double v) {
     const mixxx::BeatsPointer pBeats = pTrack->getBeats();
     if (pBeats) {
         // TODO(rryan): Track::frameInfo is possibly inaccurate!
-        const double sampleOffset = frameInfo().sampleRate * v * 0.01;
-        const mixxx::audio::FrameDiff_t frameOffset =
-                sampleOffset / mixxx::kEngineChannelOutputCount;
+        const double frameOffset = frameInfo().sampleRate * v * 0.01;
```
Swiftb0y, line 349: Please elaborate. Fyi the term "samplerate" in mixxx also takes into account the number of channels. So in the case of stereo, one second of stereo PCM DVD Audio (also referred to as 16-bit/48kHz) actually has a 92kHz samplerate (in mixxx terms). (someone correct me if I'm wrong please). What is otherwise commonly referred to as the samplerate (independent of the amount of channels) is referred as the "frame rate". So thats why the divide was originally there (we want to transition away from our incorrect definition of samplerate). 

So, I propose an alternative solution: To preserve encapsulation and to simplify the API, you should consider implementing a `Beats::tryTranslateBeats` member function. Then leverage that directly within `slotTranslateBeatsHalf`. Afterwards, revert any changes you made in `slotTranslateBeatsMove` as I don't think they're correct (or at least API compatible with the established behavior).

Swiftb0y (commented):
lets iterate on this a bit

Swiftb0y:
Please also sign the [Mixxx Contributor Agreement](https://docs.google.com/forms/d/e/1FAIpQLScC9QG327sjLL0eWftmfGUasxFFLxg6LCXJ2xHDYRzFIRqyiw/viewform?formkey=dEpYN2NkVEFnWWQzbkFfM0ZYYUZ5X2c6MQ) and comment here when you have done so. It gives us permission to distribute your contribution under the GPL v2 or later license and the Apple Mac App Store. It is also helpful for us to have contact information for contributors in case we may need it in the future.

src/engine/controls/bpmcontrol.cpp
ronso0: btw pTrack->getBpm() gets the track's global (average/dominant?) BPM
so, new func `Beats::tryTranslateBeats` could also handle non-const beats.

src/engine/controls/bpmcontrol.cpp
```diff
@@ -315,8 +322,20 @@ void BpmControl::slotTranslateBeatsLater(double v) {
...
+void BpmControl::slotTranslateBeatsHalf(double v) {
+    if (v <= 0) {
+        return;
+    }
+    const TrackPointer pTrack = getEngineBuffer()->getLoadedTrack();
+    if (!pTrack) {
+        return;
+    }
+    const double bpm = pTrack->getBpm();
+    const double halfBeat = 3000.0 / bpm;
```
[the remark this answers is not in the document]
jstolberg: Thanks for reviewing!
1 beat is 1minute/bpm = 60seconds/bpm. So half a beat is 30seconds/bpm. slotTranslateBeatsMove take its input in hundreds of a second (interpreted from the 0.01 factor that the input is multiplied with to convert it back into seconds, before multiplying with samplerate, which is in hz), and so we get 3000/bpm is the appropriate input for slotTranslateBeatsMove, if we wish to translate the grid half a beat.

src/engine/controls/bpmcontrol.cpp
jstolberg: Hmmm, okay. I assumed dividing by number of output channels was a mistake, because `slotTranslateBeatsMove `only translate by 0.5 a hundredth of a second, not 1, which I thought was because of wrongfully dividing by number of channels (2). I tested this by taking a song of 120bpm, and tapping Adjust Beatgrid Earlier, until the grid had translated by one full beat. This took 100 taps, but should take 50 for a song of 120bpm. 

Still, I like the idea of making a `Beats::tryTranslateBeats` member function, I can try to look into that. Still I wonder if `slotTranslateBeatsMove` is working the way it is intended (or if I'm misunderstanding what it is intended to do). I was also baffled why input was rounded at first in this function. I could only find `slotTranslateEarlier` and `slotTranslateLater` using the function and these call with the inputs 1 and -1.

src/engine/controls/bpmcontrol.cpp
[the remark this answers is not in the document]
Swiftb0y: Thanks for the explanation, we should document that in code. You can leverage `constexpr static` variables for documenting the intermediate steps.

src/track/beats.cpp
```diff
@@ -24,6 +26,7 @@ constexpr double kEpsilon = 0.01;
 } // namespace
 
 namespace mixxx {
+using std::transform;
```
Swiftb0y, line 29: lets avoid `using` at namespace scope. Especially with a namespace this large. I don't see how this is necessary anyways.

src/engine/controls/bpmcontrol.cpp
```diff
@@ -330,12 +330,17 @@ void BpmControl::slotTranslateBeatsHalf(double v) {
...
-    const double bpm = pTrack->getBpm();
-    const double halfBeat = 3000.0 / bpm;
-    slotTranslateBeatsMove(halfBeat);
+    const mixxx::BeatsPointer pBeats = pTrack->getBeats();
+    if (pBeats) {
+        const auto translatedBeats = pBeats->tryTranslateBeats(0.5);
+        if (translatedBeats) {
+            pTrack->trySetBeats(*translatedBeats);
+        }
+    }
```
Swiftb0y, line 339: Stylewise, prefer early returns. It makes the code easier to read because of the reduced nesting.

```suggestion
    const mixxx::BeatsPointer pBeats = pTrack->getBeats();
    if (!pBeats) {
        return;
    }
    const auto translatedBeats = pBeats->tryTranslateBeats(0.5);
    if (!translatedBeats) {
        return;
    }
    pTrack->trySetBeats(*translatedBeats);
```

src/track/beats.cpp
```diff
@@ -634,6 +637,20 @@ std::optional<BeatsPointer> Beats::tryTranslate(audio::FrameDiff_t offsetFrames)
             m_subVersion));
 }
 
+std::optional<BeatsPointer> Beats::tryTranslateBeats(double xBeats) const {
+    if (!m_markers.empty()) {
+        return std::nullopt;
+    }
+    const mixxx::audio::FrameDiff_t lastOffsetFrames =
+            xBeats * m_sampleRate.value() * 60.0 / m_lastMarkerBpm.value();
```
Swiftb0y, line 645: The changes here seem very sensible, it would be nice if they were cemented in a unit test though. The test for the `Beats` class live in `src/test/beatgridtest.cpp`.

Swiftb0y:
>  Either way, if you agree, I suggest adding the functionality for constant bpms. If the same functionality turns out to be necessary/sensible for non-constant bpms, this could possibly be suggested as a separate issue?

Yeah, if you're not interested in doing it in this PR you can raise a separate issue. I can think of a couple ways to implement this sensibly for non-constant beatgrids.

Swiftb0y:
Another feature-request from me would be to also have this functionality available in the Track BPM context menu:
image

That would be implemented in `src/widget/wtrackmenu.cpp`. Look for occurrences of `m_pBPMMenu`.

jstolberg:
> Another feature-request from me would be to also have this functionality available in the Track BPM context menu

I initially thought of this, but decided against it because you cannot see the beat grid on unloaded tracks, so you would  see no result of clicking this (and also wouldn't know it was needed, unless track is loaded). I think it makes sense to have the BPM adjustments available in the library view, since the resulting change can also be monitored there. But I don't feel strongly about it, that was just my reasoning.

Swiftb0y:
Good point indeed. So I assume you propose to let the skins implement a button instead?

jstolberg:
So far I have implemented it as the right click option on the Adjust Beatgrid Earlier/Later buttons (before there was none). I think this is a fairly intuitive solution, which then also doesn't require any interface change in any of the skins. Creating a separate button for this also seems excessive, as it is just an occasional one-shot use.

ronso0:
> I initially thought of this, but decided against it because you cannot see the beat grid of unloaded tracks in the library view, so you would see no result of clicking this

FYI it can be added to deck track menus only, like the "update replaygain from pregain" action.

ronso0:
I think a menu action would make it discoverable, I mean it migt make users look for it in the beatgrid section, too.

src/widget/wtrackmenu.cpp
```diff
@@ -1408,6 +1425,24 @@ void WTrackMenu::slotUpdateReplayGainFromPregain() {
     m_pTrack->adjustReplayGainFromPregain(gain);
 }
 
+void WTrackMenu::slotTranslateBeatsHalf() {
+    VERIFY_OR_DEBUG_ASSERT(m_pTrack) {
+        return;
+    }
+    VERIFY_OR_DEBUG_ASSERT(!m_deckGroup.isEmpty()) {
+        return;
+    }
```
Swiftb0y, line 1434: you're not using `m_deckGroup`, no need to check it IMO. checking whether `m_pTrack` exists is sufficient. But I'm not an expert in this part of the code. Wdyt @ronso0?

Swiftb0y (commented):
Thank you. Last NIT

src/widget/wtrackmenu.cpp
ronso0: Yes, it's not needed. The action is enabled only if it's not empty, ie. if this menu is parented to a WTrackProperty or alike.

ronso0:
Where exactly is the check that this can only performed on a const-BPM track?

src/widget/wtrackmenu.cpp
jstolberg: Does that mean it is also unnecessary in `WTrackMenu::slotUpdateReplayGainFromPregain`? If so, I can remove both for next commit, otherwise I will just remove it in `slotTranslateBeatsHalf`.

src/track/beats.cpp
```diff
@@ -634,6 +634,20 @@ std::optional<BeatsPointer> Beats::tryTranslate(audio::FrameDiff_t offsetFrames)
             m_subVersion));
 }
 
+std::optional<BeatsPointer> Beats::tryTranslateBeats(double xBeats) const {
+    if (!m_markers.empty()) {
```
jstolberg, line 638: This statement checks that the track is with constant BPM. Tracks with non-constant BPM has track markers, while tracks with constant BPM only has the "last marker", which is not contained in `m_markers`, but separately.

src/track/beats.cpp
```diff
@@ -634,6 +634,20 @@ std::optional<BeatsPointer> Beats::tryTranslate(audio::FrameDiff_t offsetFrames)
...
+        return std::nullopt;
+    }
```
jstolberg, line 640: This statement checks that the track is with constant BPM. Tracks with non-constant BPM has track markers, while tracks with constant BPM only has the "last marker", which is not contained in `m_markers`, but separately.

src/widget/wtrackmenu.cpp
ronso0: Indeed, it's the same there.
Let's just deal with the BPM action in this PR
(btw I'd appreciate to squash all commits once this is finished and get's LGTM)

src/track/beats.cpp
ronso0: Ah okay. I'm not too familiar with the Beats class so a comment in BpmControl would be helpful I think.
And a hint in controlpickermenu.cpp

And we need to add new left-right-click combo tooltips:
`beats_translate_earlier_move_half-beat`
`beats_translate_later_move_half-beat`

it's here
https://github.com/mixxxdj/mixxx/blob/d6abcab13affd047ba16640f64e0d0997d1647b2/src/skin/legacy/tooltips.cpp#L438

Hmm, I'll provide a patch.

ronso0: [bpm-half-tooltip.txt](https://github.com/user-attachments/files/18722331/bpm-half-tooltip.txt)
Then please use 
`beats_translate_later_move_half_beat`
`beats_translate_earlier_move_half_beat`
for the left-right buttons you changed.

Swiftb0y: > This statement checks that the track is with constant BPM. Tracks with non-constant BPM has track markers, while tracks with constant BPM only has the "last marker", which is not contained in m_markers, but separately.


Additionally, you might want to capture this feature in a (`private`) utility method. That way its easier to spot in case that assumption might change.

jstolberg: > I'm not too familiar with the Beats class so a comment in BpmControl would be helpful I think.
And a hint in controlpickermenu.cpp

I'm sorry, but I am not quite sure what should be commented in bpmcontrol. That it only works for constant bpm beats? I also don't understand what you mean with hint in controlpickermenu.cpp. Also stating the fact it is for constant BPM tracks only in the description?

> And we need to add new left-right-click combo tooltips:

This is actually part of this PR already. But maybe the way I did this not right?

jstolberg: > Additionally, you might want to capture this feature in a (`private`) utility method. 

Turns out this is already implemented in beats.h. However the suggested TODO in the comments does not align with how `tryTranslateBeats` work (this specifically needs no beat markers, as it does not deal with them if there are any). This means we can use the existing `hasConstantTempo` as long as it will not change, if it does, `tryTranslateBeats` will have to be corrected too. What do you think is best to do in this case?

Swiftb0y: I think its fine to use `hasConstantTempo()` directly. If `hasConstantTempo` needs to change, so does `tryTranslateBeats`. Moreover, the `hasConstantTempo` check is only needed because we didn't bother to implement shifting for arbitrary beats. Once we do the check is no longer needed.

ronso0:
> what should be commented in bpmcontrol. That it only works for constant bpm beats?

I would find that helpful, but if the implementation is changed (ie. made to work with non-const BPM someday) the comment may be out sight. Not _that_ important.

> I also don't understand what you mean with hint in controlpickermenu.cpp

That and the tooltip are the user-facing strings (no one will look it up in the changelog or controls documentation) so I'd like to hint users that it's gonna work only for const BPM.
Also the new tooltip assumes that we'll have only combo-buttons (left: translate small, right: translate half beat) and for separate buttons we need separate tooltips (like we have for all play/cue combos, for bpm_tap, tempo_tap and both).

Swiftb0y:
> I would find that helpful, but if the implementation is changed (ie. made to work with non-const BPM someday) the comment may be out sight. Not that important.

I agree, but I think we can file an issue for that and not drag out this PR.

jstolberg:
I think all comments have now been considered @Swiftb0y and @ronso0.

Swiftb0y (approved):
Yup. Thank you for the reminder. LGTM.

Swiftb0y:
Are you interested in implementing the variable tempo grid in a follow up?

jstolberg:
I am, but I need to understand better how these beat markers are being used, in particular how the detection algorithm sets them. I tried to implement a version of it, which assigned new position to using this approach:
`newpos = thismarker.position() + 0.5 * (nextmarker.position() - thismarker.position()) / thismarker.beatsTillNextMarker()`
The right hand side of the plus calculates the amount of frames in half a beat, and then adds that to the old position. Thought it was a good approach, until realizing repeated use of the button made Mixxx crash entirely. My worry is that this approach may unintentionally make some value diverge or tend to zero on repeated use, but not sure, especially cause I might be misinterpreting the use of the beat markers.

Swiftb0y:
>  We could discuss how to approach this in Zulip?

Yes. thats a good idea.

Swiftb0y:
Sorry for the delay. I forgot about this.

ronso0:
Please also document the new control in https://manual.mixxx.org/2.5/en/chapters/appendix/mixxx_controls.html
The manual PR should target the `main` branch (refers to next major release 2.6)

Thank you!

Issue #14280, opened 2025-02-02
Confirmation popup when entering preferences caused by `MidiController::makeInputHandler`


git-developer:
#### Description
* _Given_ that an active MIDI controller mapping calls `MidiController::makeInputHandler` on initialization,
* _when_ the user selects _Options / Preferences_ for the first time,
* _then_ a confirmation dialog pops up: 
![Image](https://github.com/user-attachments/assets/99b0254f-e920-4b69-a5f1-f1afb7ce9bd8)

#### How to reproduce
This minimal [testcase](https://github.com/user-attachments/files/18631219/testcase.zip) reproduces the problem:
* `Demo-scripts.js`
    ```js
    var Demo = {
        init: () => midi.makeInputHandler(0x90, 0x01, (channel, control, value, status) => {})
    };
    ```    
* `Demo.midi.xml`:
    ```xml
    <?xml version="1.0" encoding="UTF-8"?>
    <MixxxControllerPreset schemaVersion="1" mixxxVersion="2.5+">
      <info>
        <name>Demo Controller</name>
        <description>Demo Controller</description>
      </info>
      <controller id="DemoController">
        <scriptfiles>
          <file filename="Demo-scripts.js" functionprefix="Demo" />
        </scriptfiles>
       </controller>
    </MixxxControllerPreset>
    ```
    

### Version

2.5.0

### OS

Ubuntu 24.04, Windows

git-developer:
[`DlgPrefController::isDirty()`](https://github.com/mixxxdj/mixxx/blob/32907f86ef50e3f523c14a7a73de994520a9ce8c/src/controllers/dlgprefcontroller.cpp#L656) seems to return `true` even if the controller mapping is unchanged.

daschuer:
A fix is here: 
https://github.com/mixxxdj/mixxx/pull/14281
Please test and comment at the PR.

mxmilkiib:
dupe of #14261, though you do gooder word than I, fix in #14234 that was merged into 2.5

PR #14281, opened 2025-02-02
Fix wrong mapping change confirmation

daschuer:
This fixes https://github.com/mixxxdj/mixxx/pull/14159

This also includes some other confirmed refactoring form https://github.com/mixxxdj/mixxx/pull/14159
On the way to fix https://github.com/mixxxdj/mixxx/issues/13940

I have forked that out because the original crasher fix requires some polishing.

### DIFF
--- a/src/controllers/midi/legacymidicontrollermapping.cpp
+++ b/src/controllers/midi/legacymidicontrollermapping.cpp
@@ -20,8 +20,13 @@ void LegacyMidiControllerMapping::addInputMapping(uint16_t key, const MidiInputM
 }
 
 void LegacyMidiControllerMapping::removeInputMapping(uint16_t key) {
+    for (auto [it, end] = m_inputMappings.equal_range(key); it != end; ++it) {
+        const MidiInputMapping& mapping = it.value();
+        if (!std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
+            setDirty(true);
+        }
+    }
     m_inputMappings.remove(key);
-    setDirty(true);
 }
 
 bool LegacyMidiControllerMapping::removeInputMapping(
--- a/src/controllers/midi/legacymidicontrollermapping.cpp
+++ b/src/controllers/midi/legacymidicontrollermapping.cpp
@@ -13,7 +13,10 @@ bool LegacyMidiControllerMapping::isMappable() const {
 
 void LegacyMidiControllerMapping::addInputMapping(uint16_t key, const MidiInputMapping& mapping) {
     m_inputMappings.insert(key, mapping);
-    setDirty(true);
+    if (!std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
+        // Note: JS handler are not saved to the XML file
+        setDirty(true);
+    }
 }
 
 void LegacyMidiControllerMapping::removeInputMapping(uint16_t key) {
--- a/src/controllers/midi/midicontroller.cpp
+++ b/src/controllers/midi/midicontroller.cpp
@@ -286,8 +286,10 @@ void MidiController::receivedShortMessage(unsigned char status,
         }
     }
 
-    auto it = m_pMapping->getInputMappings().constFind(mappingKey.key);
-    for (; it != m_pMapping->getInputMappings().constEnd() && it.key() == mappingKey.key; ++it) {
+    for (auto [it, end] =
+                    m_pMapping->getInputMappings().equal_range(mappingKey.key);
+            it != end;
+            ++it) {
         processInputMapping(it.value(), status, control, value, timestamp);
     }
 }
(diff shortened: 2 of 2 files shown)


git-developer:
This fixes #14280. Thank your very much for your instant support!

src/controllers/controller.h
```diff
@@ -79,24 +79,16 @@ class Controller : public QObject {
 
   protected:
     template<typename SpecificMappingType>
-    std::shared_ptr<SpecificMappingType> downcastAndTakeOwnership(
-            std::shared_ptr<LegacyControllerMapping>&& pMapping) {
+    std::unique_ptr<SpecificMappingType> downcastAndClone(const LegacyControllerMapping* pMapping) {
         // When unsetting a mapping (select 'No mapping') we receive a nullptr
-        if (pMapping == nullptr) {
+        auto* pSpecifiedMapping = dynamic_cast<const SpecificMappingType*>(pMapping);
```
Swiftb0y, line 84:
```suggestion
        const auto* pSpecifiedMapping = dynamic_cast<const SpecificMappingType*>(pMapping);
```
I think this results in a linter warning down the road otherwise.

src/controllers/midi/midicontroller.cpp
```diff
@@ -655,5 +656,13 @@ QJSValue MidiController::makeInputHandler(unsigned char status,
             std::make_shared<QJSValue>(scriptCode));
 
     m_pMapping->addInputMapping(inputMapping.key.key, inputMapping);
-    return pJsEngine->newQObject(new MidiInputHandleJSProxy(m_pMapping, inputMapping));
+    // The returned object can be used for disconnecting like this:
+    // var connection = midi.makeInputHandler();
```
Swiftb0y, line 660: NIT
```suggestion
    // const connection = midi.makeInputHandler();
```

src/controllers/controller.h
```diff
@@ -79,24 +79,16 @@ class Controller : public QObject {
...
-        VERIFY_OR_DEBUG_ASSERT(pMapping.use_count() == 1) {
-            return nullptr;
-        }
-        auto pDowncastedMapping = std::dynamic_pointer_cast<SpecificMappingType>(pMapping);
-        VERIFY_OR_DEBUG_ASSERT(pDowncastedMapping) {
-            return nullptr;
-        }
-        return pDowncastedMapping;
+        // LegacyControllerMapping. So we do a deep copy here.
+        return std::make_unique<SpecificMappingType>(*pSpecifiedMapping);
```
Swiftb0y, line 91: unless you know that the `SpecificMappingType` is `final` this will result in sliced classes. Or said differently, this can only work because `LegacyControllerMapping` has not deleted its copy and move operations (even though it should). 
See: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-copy-virtual
And: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rh-copy

Swiftb0y (changes requested):
The ownership related changes in this PR are unfortunately flawed and need thorough discussion.

Swiftb0y:
Let me clarify: I wasn't asking you to leverage metaprogramming to ensure slicing doesn't happen, I was asking you to consider splitting the ownership-related changes from the actual bugfix. And if you don't think thats possible I'd like to know why.

daschuer:
This would be possible. I have just picked the commits I had here, as described in the initial post. 

If you prefer to merge them in two commits, I can split. 
Is there anything left with the pointed out issue?

Swiftb0y:
Well, the ownership stuff in the mappings is flawed, I 100% agree with that, but lets not conflate changes to that with #14280.

daschuer:
Other commit removed.

git-developer:
I can confirm this still fixes the issue.

Swiftb0y (approved):
Thank you. LGTM.

Swiftb0y:
fyi wouldn't a similar thing also need to apply to `removeInputMapping`?

daschuer:
Yes, good catch. Done!

src/controllers/midi/legacymidicontrollermapping.cpp
```diff
@@ -20,8 +20,14 @@ void LegacyMidiControllerMapping::addInputMapping(uint16_t key, const MidiInputM
...
-    m_inputMappings.remove(key);
-    setDirty(true);
+    auto it = m_inputMappings.find(key);
+    if (it != m_inputMappings.end()) {
+        const MidiInputMapping& mapping = it.value();
+        if (!std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
+            setDirty(true);
+        }
+        m_inputMappings.erase(it);
+    }
```
Swiftb0y, line 30: since `m_inputMappings` is a `QMultiHash` I assume multiple controls could be stored under the same key? `QMultiHash::remove` removes all mappings under that key while this only seems to remove the first one. I think you'll need something like:

```suggestion
    for (auto [it, end] = m_inputMappings.equal_range(key); it != end; ++it) {
        const MidiInputMapping& mapping = it.value();
        if (!std::holds_alternative<std::shared_ptr<QJSValue>>(mapping.control)) {
            setDirty(true);
        }
        m_inputMappings.erase(it);
    }
```

I'm not 100% sure how safe this is in regards to iterator invalidation because I wasn't able to find any docs on `QMultiHash::erase` (in Qt6 where `QMultiHash` doesn't publicly inherit from `QHash` anymore).

Swiftb0y (approved):

daschuer:
CI is done. Thank you for review.

Swiftb0y:
sorry, for missing it.

PR #14285, opened 2025-02-02
Remove XML input declarations from Behringer DDM4000 & BCR2000 mapping

git-developer:
The mappings for Behringer [DDM4000](https://github.com/mixxxdj/mixxx/blob/2.4/res/controllers/Behringer%20DDM4000.midi.xml) and [BCR2000](https://github.com/mixxxdj/mixxx/blob/2.4/res/controllers/Behringer%20BCR2000.midi.xml) require a total of ~1700 lines of XML. All MIDI input is sent to a single `input` function that dispatches to the appropriate Mixxx Component.

#12781 makes it possible to register MIDI input handlers from JS.

This PR replaces the XML input declarations with a few lines of JS code, creating an input connection when a component is registered and disconnecting when it's unregistered. The code was tested against current `2.5` branch (`mixxx-2.5.0-51-gcd101f1d23.deb` from [CI](https://github.com/mixxxdj/mixxx/actions/runs/13097502050)) containing several additions related to these mappings (#14203, #14232, #14266 & #14281). Behavior of both controllers is unchanged.


res/controllers/Behringer-Extension-scripts.js
```diff
@@ -1101,6 +1102,13 @@
          */
         register: function(component, shift) {
             this.onRegistry(this.componentRegistry.register, component, shift);
+            if (component.midi) {
+                const id = findComponentId(component.midi);
+                if (!Object.prototype.hasOwnProperty.call(this.inputConnections, id)) {
```
Swiftb0y, line 1107: whats the difference between this and 
```suggestion
                if (this.inputConnections[id] !== undefined) {
```


I mean I see there is a difference in general, but does it make a concrete difference in this case to simplify like this?

git-developer: It was meant be a bit more `null`- & `undefined`-safe. But it looks like it just crashes _differently_ when `inputConnections` is missing.

Swiftb0y: I think its reasonable to assume you `inputConnections` is a valid object, after all you set it up in the constructor. Also testing only for `undefined` explicitly is fine as long as you ensure you never put `null`s into that object.

Swiftb0y (commented):
This does make use of new feature (which is also not fully fledged out as you fortunately noticed). I would prefer to retarget this at `main` instead. wdyt?

res/controllers/Behringer-Extension-scripts.js
git-developer: Just for the record: it needs to be `=== undefined`, not `!== undefined`.

git-developer:
I agree, in favor of stability of the 2.5 line.

Swiftb0y (approved):
LGTM. Thank you. Waiting for CI

res/controllers/Behringer-Extension-scripts.js
```diff
@@ -1101,6 +1102,13 @@
          */
```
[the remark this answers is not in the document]
git-developer: Not really, what makes you think so?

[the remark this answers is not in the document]
daschuer: The function doe not only "Add a component to a layer." it also greisteres the Input Handle.

daschuer:
While you have removed the XML part of the code, the commponent looup is still there. I think you can remove this whole looup call by reistering directly this function: 
https://github.com/mixxxdj/mixxx/blob/a6d3250729be6b9f46809d532113266d496676e5/res/controllers/Behringer-Extension-scripts.js#L1195

git-developer:
I don't think that this lookup can be omitted.

When a MIDI message arrives, its address might be associated with multiple components: one on the Default Layer and another one on the Shift Layer. The LayerManager calls `findComponent` to determine the component on the currently active layer and delegates only to this component. Your suggestion would skip the LayerManager and delegate to both components.

```mermaid
sequenceDiagram
XML ->> GenericMidiController: input()
GenericMidiController ->> LayerManager: input()
LayerManager ->> LayerManager: findComponent()
LayerManager ->> Active Component: input()
```

What we could do instead is remove the `input()` function from the `GenericMidiController`, because XML is no longer required. But I didn't remove that (yet) for compatibility and because I don't know if there might be valid use cases to mix XML and JS.

res/controllers/Behringer-Extension-scripts.js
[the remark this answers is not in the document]
git-developer: Ah, that's true. I extended the docs.

daschuer:
OK than lets merge it. Thank you.

git-developer:
Thanks for your review, and thanks to @christophehenry for making this possible by implementing `makeInputHandler()`.

PR #14289, opened 2025-02-03
2.5 vcpkg update and libdjinteropt include fix

daschuer:
This updates to the new vcpkg environment including libdjinteropt 0.24.3

Normally it is intended to download the correct version If a different version is provided. However this was broken, because the system headers where used with priority. This is fixed here as well.


daschuer:
Without the fix we see this: 
https://github.com/daschuer/mixxx/actions/runs/13098980787

mr-smidge (approved):

JoergAtGithub:
Thank you!

PR #14291, opened 2025-02-03
fix: use generic way to get the group on stem waveform renderer

acolombier:
Not sure how this one slipped through  - since the new stem rendergraph support, `WaveformRendererStem::init` may run while `m_pEQEnabled` is still null, leading to a crash. Using the correct way to get the group in renderer fixes the issue

### DIFF
--- a/src/waveform/renderers/allshader/waveformrendererstem.cpp
+++ b/src/waveform/renderers/allshader/waveformrendererstem.cpp
@@ -31,9 +31,8 @@ void WaveformRendererStem::onSetup(const QDomNode&) {
 }
 
 bool WaveformRendererStem::init() {
-    auto group = m_pEQEnabled->getKey().group;
     for (int stemIdx = 0; stemIdx < mixxx::kMaxSupportedStems; stemIdx++) {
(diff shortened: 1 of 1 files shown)


Swiftb0y (approved):
:shrug:

Holzhaus:
Ah, I had the same issue when testing #13470. My fix was to invoke the parent class `WaveformRendererSignalBase::init()` from `WaveformRendererStem::init()`. See https://github.com/mixxxdj/mixxx/pull/13470#issuecomment-2632131601

I have no idea what approach is better or more correct, just wanted to let you know.

Holzhaus:
Looking at the code here, this one is much less hacky. However, It's still weird that we have controlproxies in the base class that are not initialized or used in the subclass. Maybe we should move them to another class and let waveformrendererstem directly inherit from the base without it?

Holzhaus:
anyway, this could happen in a follow-up maybe, no need to block this bug-fix

acolombier:
> Maybe we should move them to another class and let waveformrendererstem directly inherit from the base without it?

Yeah, I think this could definitely use a cleanup! I'll see if I get the time soon, but anyone feel free to go ahead on this one! :)

PR #14292, opened 2025-02-03
Refactor passing of LegacyControllerMapping

daschuer:
This is the groundwork of fixing https://github.com/mixxxdj/mixxx/issues/13940

It contains the confirmed refactoring form https://github.com/mixxxdj/mixxx/pull/14159
I have forked that out because the original crasher fix requires some more polishing. 

It basically enforces that m_pMapping is unique by using a unique pointer instead of using an assertion on a shared pointer, which is not 100 % reliable in a multi-threading context.

### DIFF
--- a/src/controllers/controller.h
+++ b/src/controllers/controller.h
@@ -79,6 +79,7 @@ class Controller : public QObject {
 
   protected:
     template<typename SpecificMappingType>
+        requires(std::is_final_v<SpecificMappingType> == true)
     std::unique_ptr<SpecificMappingType> downcastAndClone(const LegacyControllerMapping* pMapping) {
         // When unsetting a mapping (select 'No mapping') we receive a nullptr
         auto* pSpecifiedMapping = dynamic_cast<const SpecificMappingType*>(pMapping);
--- a/src/controllers/controller.h
+++ b/src/controllers/controller.h
@@ -82,6 +82,9 @@ class Controller : public QObject {
         requires(std::is_final_v<SpecificMappingType> == true)
     std::unique_ptr<SpecificMappingType> downcastAndClone(const LegacyControllerMapping* pMapping) {
         // When unsetting a mapping (select 'No mapping') we receive a nullptr
+        if (!pMapping) {
+            return nullptr;
+        }
         auto* pSpecifiedMapping = dynamic_cast<const SpecificMappingType*>(pMapping);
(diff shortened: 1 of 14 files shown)


src/controllers/legacycontrollermapping.h
```diff
@@ -38,10 +38,10 @@ class LegacyControllerMapping {
               m_scripts(other.m_scripts),
               m_deviceDirection(other.m_deviceDirection) {
     }
+    LegacyControllerMapping& operator=(const LegacyControllerMapping&) = delete;
```
Swiftb0y, line 41: rule of 5. lets also define the move operations.

src/controllers/controller.h
```diff
@@ -79,24 +79,17 @@ class Controller : public QObject {
...
   protected:
     template<typename SpecificMappingType>
-    std::shared_ptr<SpecificMappingType> downcastAndTakeOwnership(
-            std::shared_ptr<LegacyControllerMapping>&& pMapping) {
+        requires(std::is_final_v<SpecificMappingType> == true)
+    std::unique_ptr<SpecificMappingType> downcastAndClone(const LegacyControllerMapping* pMapping) {
         // When unsetting a mapping (select 'No mapping') we receive a nullptr
-        if (pMapping == nullptr) {
+        auto* pSpecifiedMapping = dynamic_cast<const SpecificMappingType*>(pMapping);
+        VERIFY_OR_DEBUG_ASSERT(pSpecifiedMapping) {
```
Swiftb0y, line 86: the comment indicates why `nullptr` is valid and thus we should not `DEBUG_ASSERT` here.

Swiftb0y (commented):
while the current solution does not really rectify the bad pre-established design, the changes seem reasonable.

src/controllers/controller.h
daschuer: Oh yes. done.

src/controllers/legacycontrollermapping.h
Swiftb0y: Right, but the rule of five primarily says that you should explicit about it instead of rely on the complicated rules on whether they're autogenerated or not. 
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c21-if-you-define-or-delete-any-copy-move-or-destructor-function-define-or-delete-them-all

Swiftb0y (approved):

daschuer:
CI is done. Thank you for review.

PR #14309, opened 2025-02-07
Overview: scale by ReplayGain

ronso0:
For all those who use the 'Global' gain scaling also used by the waveforms, ie. for consistency with the waveforms.

Might as well go to 2.5 if you think this is a bugfix.

### DIFF
--- a/src/widget/woverview.cpp
+++ b/src/widget/woverview.cpp
@@ -111,6 +111,9 @@ WOverview::WOverview(
             &WaveformWidgetFactory::overallVisualGainChanged,
             this,
             &WOverview::slotNormalizeOrVisualGainChanged);
+    // Also listen to ReplayGain changes to scale the waveform
+    m_pReplayGain = make_parented<ControlProxy>(m_group, "replaygain", this);
+    m_pReplayGain->connectValueChanged(this, &WOverview::slotNormalizeOrVisualGainChanged);
 
     m_pPassthroughLabel = make_parented<QLabel>(this);
 
@@ -730,15 +733,16 @@ void WOverview::drawAxis(QPainter* pPainter) {
 
 void WOverview::drawWaveformPixmap(QPainter* pPainter) {
     if (!m_waveformSourceImage.isNull()) {
-        PainterScope painterScope(pPainter);
         WaveformWidgetFactory* pWidgetFactory = WaveformWidgetFactory::instance();
         float diffGain;
         bool normalize = pWidgetFactory->isOverviewNormalized();
         if (normalize && m_pixmapDone && m_waveformPeak > 1) {
             diffGain = 255 - m_waveformPeak - 1;
         } else {
+            DEBUG_ASSERT(m_pCurrentTrack);
             const auto visualGain = static_cast<float>(
-                    pWidgetFactory->getVisualGain(WaveformWidgetFactory::All));
+                    pWidgetFactory->getVisualGain(WaveformWidgetFactory::All) *
+                    m_pCurrentTrack->getReplayGain().getRatio());
             diffGain = 255.0f - (255.0f / visualGain);
         }
 
--- a/src/widget/woverview.cpp
+++ b/src/widget/woverview.cpp
@@ -102,12 +102,12 @@ WOverview::WOverview(
 
     // Update immediately when the normalize option or the visual gain have been
     // changed in the preferences.
-    WaveformWidgetFactory* widgetFactory = WaveformWidgetFactory::instance();
-    connect(widgetFactory,
+    WaveformWidgetFactory* pWidgetFactory = WaveformWidgetFactory::instance();
+    connect(pWidgetFactory,
             &WaveformWidgetFactory::overviewNormalizeChanged,
             this,
             &WOverview::slotNormalizeOrVisualGainChanged);
-    connect(widgetFactory,
+    connect(pWidgetFactory,
             &WaveformWidgetFactory::overallVisualGainChanged,
             this,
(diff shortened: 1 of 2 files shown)


Swiftb0y (approved):
Code LGTM. Haven't tested though. I'm neutral to whether this should go into `main` or 2.5

src/widget/woverview.cpp
```diff
@@ -737,8 +739,10 @@ void WOverview::drawWaveformPixmap(QPainter* pPainter) {
         if (normalize && m_pixmapDone && m_waveformPeak > 1) {
             diffGain = 255 - m_waveformPeak - 1;
         } else {
+            DEBUG_ASSERT(m_pCurrentTrack);
```
acolombier, line 742: How about using `VERIFY_OR_DEBUG_ASSERT` so you can prevent the segfault at line 745?

ronso0: `m_waveformSourceImage` ic cleared when a track is loaded (null or track) so we shouldn't get here with null track in the first place. So I think we're good with a DEBUG_ASSERT like in the other places.

ronso0:
I had to rebase: I accdidentally removed an unrelated line, and prepended a commit that adds p prefix for WaveformWidgetFactory.

Swiftb0y (approved):
LGTM. Thank you.

src/widget/woverview.cpp
acolombier: In this case, how about removing this `DEBUG_ASSERT`? Any reasons why you added it in a first place?

ronso0: My understanding is that DEBUG_ASSERT should be used where we are sure a certain precondition is met at the time of writing, but surrounding code / callers may be changed later on and programming errors should be catched in debug builds.

Otoh VERIFY_OR_DEBUG_ASSERT is for when we are not 100% sure.

This targets main so there's plenty of time to discover situations where the assumption is wrong.
Does that make sense?

The only dev discussion about this that I found is https://mixxx.zulipchat.com/#narrow/stream/109171-development/topic/Bpm.3A.3Avalue.28.29.20without.20checking.20validity

acolombier: It does make sense, thanks for clarifying!

daschuer:
Sorry for jumping in after merge. 

Now the overview waveform disappears for tracks without replay gain. 

Did you consider to also reflect the replay gain preferences?  The main waveform does it. With the the default ReplayGain of -18 LUFS the overview waveform has a lot of head room. So we may also want to follow the Target Loudness? 
(Unsure)

Swiftb0y:
whoops, good point. I guess a replaygain of `0` needs to default to `1` then...

fwcd:
Hm not sure how I feel about this, I would have filed a regression for this... the large overview waveforms were very helpful for viewing the track at a glance:

Screenshot 2025-02-21 at 01 39 25

Now everything is scaled down, making it much harder to identify sections of the song quickly:

Screenshot 2025-02-21 at 01 38 49

For the main waveforms this makes sense due to markers etc., but for the smaller ones I found it very useful that they were scaled to the maximum height.

fwcd:
Regarding the configuration, it's a bit unintuitive that the visual gain is listed under "Scrolling Waveforms" while affecting the overview waveforms. Also this only seems to apply to the global gain, not low/mid/high gain:

image

Maybe this preference page could be restructured to have a section for settings that affect both waveforms. Or maybe a separate, user-configurable visual gain for overview waveforms?

ronso0:
Of course I can revert this if it breaks use cases I didn't consider.
> tbh I would have expected the overview to just display the file's gain as-is, that's what it previously did I suppose?

Previously it used the file gain. So, if you didn't have Normalize checked, but are using ReplayGain, the overview amplitude would differ from that of the main waveforms. That was the reasoning behind this (I noticed it with the library overview column where for some tracks I was missing 'energy' information I see in the man waveforms.

ronso0:
> Regarding the configuration, it's a bit unintuitive that the visual gain is listed under "Scrolling Waveforms" while affecting the overview waveforms. Also this only seems to apply to the global gain, not low/mid/high gain:

Yes, true. During the redesing of that page I thought about it but didn't want to expand the scope of that PR even further.
> Maybe this preference page could be restructured to have a section for settings that affect both waveforms. Or maybe a separate, user-configurable visual gain for overview waveforms?

Yeah, I think separate overview options are the way to go, for example radio buttons for
- [ ] Normalize
- [ ] use 'All' gain of scrolling waveform incl. ReplayGain
- [ ] adjust gain: [QDoubleSpinBox]

ronso0:
I started working on this and think we should provide these scale options:
- File gain (what we had before this PR with 'Normalize' unchecked)
- Normalize
- waveform "All" gain + ReplayGain
- custum gain [spinbox]

They can be represented either as combobox or radiobutton list.
I prefer radiobutton because you can see all options. Of course a combobox would  more compact but I thunk we can afford the extra space in the waveforms page.

For the modes I'd introduce an `enum class OverviewScaleMode`, either in `WaveformWidgetFactory` or standalone like the [`OverviewType`](https://github.com/mixxxdj/mixxx/pull/14140/files#diff-733ba5eeef642a6e77c22b82b7c59ff78bda5440e81db5978e09a56e752585d1) introduced in #14140.
The "custom gain" would be a separate member of `WaveformWidgetFactory`.

What do you think?

ronso0:
Uh, just noticed my comment hasn't been posted until now.
PR is #14463 (still draft to reolve conflicts with #13470 )

PR #14311, opened 2025-02-08
Make QmlPlayerProxy read-only

christophehenry:
Per discussion on Zulip, this PR disable setters in `QmlPlayerProxy` for now.


Swiftb0y (changes requested):
Blocking for now as the motivation (setters being broken) has not been confirmed to be true.

Swiftb0y:
Thanks for the fast PR @christophehenry but I think this needs more conversation for now.

PR #14314, opened 2025-02-09
Create JavascriptPlayerProxy

christophehenry:
Continuation of #14175.


cr7pt0gr4ph7 (changes requested):
**Update:** This comment is outdated, please refer to the [Zulip discussion](https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/Expose.20track.20informations.20to.20controllers) referenced by @ronso0 below instead!

Thank you for the PR! As far as I can tell, the ultimate purpose of this PR is to allow JS controller scripts to access information about the tracks loaded into each deck, so it can be displayed on more advanced DJ controllers that have integrated displays.

As it stands, this seems to only be a first draft, and missing the actual integration. Here are a few notes, based on a quick initial review:
- In the end, the goal of this PR is to extend the API offered to controller scripts. This has implications about backward compatibility in the future, as once those interfaces are released, changing them runs the risk of breaking existing controller scripts.
  - For this reason, it would be advisable to only expose strictly necessary information, like title, artist etc. The waveform data seems to be the odd one out, as there a various different formats possible. I would therefore advise on excluding the waveform data for a first approach.
  - Also, it would be best to get consensus on the desired API offered to controller scripts before homing in on a concrete implementation. Could you give an example on what usage would look like in an actual controller script?

The final API might look totally different. For example, based on the current design pattern of the [Engine JS API](https://github.com/mixxxdj/mixxx/wiki/MIDI-Scripting#interact-with-mixxx), the following design could make sense:

```js
engine.getString(string group, string key); // returns string

engine.getString("[Channel1]", "track_artist")
engine.getString("[Channel1]", "track_title")
```

(As a sidenote, it is possible to mark PRs such as this as drafts until the implementation is finished).

ronso0:
FYI the corresponding Zulip convo is
https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/Expose.20track.20informations.20to.20controllers

cr7pt0gr4ph7:
> FYI the corresponding Zulip convo is https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/Expose.20track.20informations.20to.20controllers

Thanks, I missed that! Still getting used to how the communication is spread across the different channels.

cr7pt0gr4ph7:
> Thanks for chirping in @cr7pt0gr4ph7. I appreciate your suggestion but IMO it has also been discussed previously and rejected (See [the corresponding zulip thread](https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/Expose.20track.20informations.20to.20controllers)).

I wasn't aware of the Zulip discussion when I wrote that comment. I've hidden my comment and added a note at the top to avoid future confusion, sorry for the noise. I'll continue the discussion w.r.t. API design over at Zulip.

Additionally, I'd also like to make the proposal of preactively linking Zulip discussions in GitHub issues and PRs for future cross-reference when these contain highly relevant discussion, such as here ;)

src/qml/qmlplayerproxy.cpp
```diff
@@ -67,124 +44,40 @@ void QmlPlayerProxy::loadTrackFromLocationUrl(const QUrl& trackLocationUrl, bool
...
-                &QmlPlayerProxy::slotHotcuesChanged);
-#ifdef __STEM__
-        connect(pTrack.get(),
-                &Track::stemsUpdated,
-                this,
-                &QmlPlayerProxy::slotStemsChanged);
-#endif
-        slotBeatsChanged();
-        slotHotcuesChanged();
+void QmlPlayerProxy::subclassesSlotTrackLoaded() {
```
acolombier, line 47: I believe this is a bit of an anti pattern. I think it would be best to let sub classes register a connection directly, in case they need custom actions

acolombier (commented):
This is looking pretty much good to go IMO!
Only thing is that it would be worth to combining our effort with  #14516, which break the dependencies between the player and a track (quick TL;DR on why this is needed - QML needs a way to interact with track that aren't loaded on player). Note that this change also removes all the waveform left over bits, that are now irrelevant thanks to the rendergraph waveform library, and this should help simplifying your PR.

Also, it would be great if you could put typing instructions in `res/controllers/engine-api.d.ts` and also provide a small example on how this is intended to be used. Also, feel free to implement a basic usage in a controller mapping you feel relevant just to help with understanding and testing (just make sure to use a separate commit for it for better change control)

christophehenry:
@acolombier Since you're moving most of `QmlPlayerProxy`'s code into `QmlTrackProxy`in  #14516, would you agree to create the read-only `JavascriptTrackProxy`?

JoergAtGithub:
The Deck/Player might be better parent object here than the track, as it could provide additional information that depend on playposition or rate, like the formated actual key string, which is often used in jogwheel displays.

christophehenry:
Ok so we're back to defining a wrapper around `QmlTrackProxy`+`QmlPlayerProxy` then?

acolombier:
> The Deck/Player might be better parent object here than the track, as it could provide additional information that depend on playposition or rate

The idea with the split is to allow both. Player (which can be a deck, sampler or preview) will have playing information (rate, position, ...) as well current loaded track, which itself will have its own information, unrelated to player.

JoergAtGithub:
> Ok so we're back to defining a wrapper around `QmlTrackProxy`+`QmlPlayerProxy` then?

This code is fine, it just needs to be adjusted as described by @acolombier above!

christophehenry:
Well @acolombier's PR splits the code in two classes so I don't really understand how this code fits.

christophehenry:
I rebased this PR and added the implementation.

src/controllers/scripting/controllerscriptenginebase.cpp
```diff
@@ -194,6 +199,22 @@ void ControllerScriptEngineBase::showQMLExceptionDialog(
...
+QObject* ControllerScriptEngineBase::getPlayer(const QString& deck) {
+    if (!s_pPlayerManager) {
+        qWarning() << "PlayerManagerProxy failed to find player for group" << deck;
+        return nullptr;
+    }
+
+    // Don't set a parent here, so that the QML engine deletes the object when
+    // the corresponding JS object is garbage collected.
+    JavascriptPlayerProxy* pPlayerProxy = new JavascriptPlayerProxy(
+            s_pPlayerManager->getPlayer(deck), nullptr);
```
acolombier, line 212: `PlayerManager::getPlayer` will return a null ptr if `deck` doesn't exist, could you please handle that gracefully and return a `QJSEngine::newErrorObject`?

src/controllers/scripting/controllerscriptenginebase.h
```diff
@@ -10,6 +10,8 @@
 #include "util/runtimeloggingcategory.h"
 #ifdef MIXXX_USE_QML
 #include "controllers/controllerenginethreadcontrol.h"
+#include "javascriptplayerproxy.h"
+#include "mixer/playermanager.h"
```
acolombier, line 14: Do we want to make this feature depending of `QML`? (MIXXX_USE_QML) I don't have an objection on this since we are now investing effort in getting QML out, so it will likely be enabled on 2.7.

src/controllers/scripting/javascriptplayerproxy.h
```diff
@@ -0,0 +1,68 @@
...
+    Q_PROPERTY(QString artist READ getArtist NOTIFY artistChanged)
+    Q_PROPERTY(QString title READ getTitle NOTIFY titleChanged)
+    Q_PROPERTY(QString album READ getAlbum NOTIFY albumChanged)
+    Q_PROPERTY(QString albumArtist READ getAlbumArtist NOTIFY albumArtistChanged)
+    Q_PROPERTY(QString genre READ getGenre STORED false NOTIFY genreChanged)
+    Q_PROPERTY(QString composer READ getComposer NOTIFY composerChanged)
+    Q_PROPERTY(QString grouping READ getGrouping NOTIFY groupingChanged)
+    Q_PROPERTY(QString year READ getYear NOTIFY yearChanged)
+    Q_PROPERTY(QString trackNumber READ getTrackNumber NOTIFY trackNumberChanged)
+    Q_PROPERTY(QString trackTotal READ getTrackTotal NOTIFY trackTotalChanged)
```
acolombier, line 27: I'm wondering if it is necessary to implement `NOTIFY` here. Since this code won't be used by QML (as it has its own implementation), there is no need to provide this to support property binding. This will allow you to remove most of the `signals` and reduce the code duplication with `src/qml/qmltrackproxy.h`

src/controllers/scripting/controllerscriptenginebase.cpp
```diff
@@ -36,6 +36,11 @@ void ControllerScriptEngineBase::registerTrackCollectionManager(
     s_pTrackCollectionManager = std::move(pTrackCollectionManager);
 }
 
+void ControllerScriptEngineBase::registerPlayerManager(
```
acolombier, line 39: This function is never called. Will this be expected to be whenever a controller starts? From ControllerManager?

In this case, how about changing the singleton pattern to simple dependency injection, when you [start the engine](https://github.com/mixxxdj/mixxx/blob/0d9fe85e3b923c69cba6a968ce82a2f35b2bd25c/src/controllers/controller.cpp#L58)?

src/controllers/scripting/javascriptplayerproxy.cpp
```diff
@@ -0,0 +1,125 @@
...
+        emit trackLoading();
+        return;
+    }
+
+    const TrackPointer pTrack = m_pCurrentTrack;
+    if (pTrack != nullptr) {
+        disconnect(pTrack.get(), nullptr, this, nullptr);
+    }
+    m_pCurrentTrack.reset();
+    m_pCurrentTrack = pNewTrack;
```
acolombier, line 94: `pTrack` looks redundant

```suggestion
    if (m_pCurrentTrack != nullptr) {
        disconnect(m_pCurrentTrack.get(), nullptr, this, nullptr);
    }
    m_pCurrentTrack = pNewTrack;
```

src/controllers/scripting/javascriptplayerproxy.cpp
```diff
@@ -0,0 +1,125 @@
...
+    emit genreChanged();
+    emit composerChanged();
+    emit groupingChanged();
+    emit yearChanged();
+    emit trackNumberChanged();
+    emit trackTotalChanged();
+}
+
+bool JavascriptPlayerProxy::isLoaded() const {
+    return m_pCurrentTrack != nullptr;
```
acolombier, line 113: I'm wondering if it would be best to check the internal state instead of the locally cached value, since this is a proxy. This could help with race condition, but also no strong opinion there.

```suggestion
    return m_pTrackPlayer->getLoadedTrack() != nullptr;
```

src/controllers/scripting/javascriptplayerproxy.cpp
```diff
@@ -0,0 +1,125 @@
...
+                this,
+                &JavascriptPlayerProxy::trackTotalChanged);
+    }
+    emit trackChanged();
+    emit trackLoaded();
+}
+
+void JavascriptPlayerProxy::slotLoadingTrack(TrackPointer pNewTrack, TrackPointer pOldTrack) {
+    VERIFY_OR_DEBUG_ASSERT(pOldTrack == m_pCurrentTrack) {
+        qWarning() << "QML Player proxy was expected to contain "
```
acolombier, line 79:
```suggestion
        qWarning() << "Javascript Player proxy was expected to contain "
```

src/controllers/scripting/controllerscriptenginebase.h
christophehenry: Yeah, probably not. This residual from when `QmlPlayerProxy` and `JavascriptPlayerProxy` where related.

src/controllers/scripting/javascriptplayerproxy.h
Swiftb0y: Yup, we should def also use signals from JS. Otherwise we'll need yet another `makeConnection` like facility or people will abuse (likely only tangentially related) COs instead.

christophehenry, line 27: Wait, thoses `NOTIFY` don't work as I expected. From JS side, the new value isn't passed to the callback as I thought. You need to write something like:

```js
const player = engine.getPlayer("[Channel1]");
player.trackChanged.connect(() => console.log(`New track: ${player.title}`));
```

Or did I miss something?

src/controllers/scripting/controllerscriptenginebase.cpp
acolombier: Happy with the middle ground of returning `null` if we think this is best. I think we shouldn't return an invalid `JavascriptPlayerProxy` tho

src/controllers/scripting/javascriptplayerproxy.h
acolombier: Ah sorry, I didn't see that this will actively be supported in the script. This is looking good then. The duplication remains a bit of a bummer IMO, but don't want to block this further so if this isn't a problem for other member, happy to ignore!

src/coreservices.cpp
```diff
@@ -495,6 +495,8 @@ void CoreServices::initialize(QApplication* pApp) {
 
     m_isInitialized = true;
 
+    ControllerScriptEngineBase::registerPlayerManager(getPlayerManager());
```
acolombier, line 498: What do you think about not using a singleton here since you don't have the QML requirement to do so? ?

src/controllers/scripting/controllerscriptenginebase.cpp
christophehenry: Yes I changed that in the last commit.

src/controllers/scripting/javascriptplayerproxy.h
christophehenry: Ok. I'd modify the code to pass the value but I'll keep it like this if other oppose.

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -136,6 +136,13 @@ QJSValue ControllerScriptInterfaceLegacy::getSetting(const QString& name) {
     }
 }
 
+QObject* ControllerScriptInterfaceLegacy::getPlayer(const QString& deck) {
+#ifndef MIXXX_USE_QML
+    return nullptr;
+#endif
+    return m_pScriptEngineLegacy->getPlayer(deck);
```
christophehenry, line 143:
```suggestion
    return m_pScriptEngineLegacy->getPlayer(deck);
```

src/coreservices.cpp
christophehenry: Wait, how would I do that? How can I retrieve the `PlayerManager` from `JavascriptPlayerProxy` ?

src/controllers/scripting/javascriptplayerproxy.h
```diff
@@ -0,0 +1,70 @@
...
+        const TrackPointer pTrack = m_pCurrentTrack;        \
+        if (pTrack == nullptr) {                            \
+            return TYPE();                                  \
+        }                                                   \
+        return pTrack->GETTER();                            \
+    }
+
+#define STATE_EMPTY "empty"
+#define STATE_LOADING "loading"
+#define STATE_LOADED "loaded"
```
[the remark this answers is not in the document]
Swiftb0y: yup. an enum would be the better tool. `#define`s for constants are ugly and string comparisons are slow (compared to enums). If you want to expose it to JS, make it a `Q_ENUM`.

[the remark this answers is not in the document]
christophehenry: OK. I didn't want to expose yet another object but if you think it's cleaner, I'll go for it.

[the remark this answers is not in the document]
Swiftb0y: yes, better than some magic string constants. Though we should properly think whether the newly introduced state logic is really necessary. I currently don't really see the usecase. Can you share some of your rationale for the recent change?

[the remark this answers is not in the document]
JoergAtGithub: Q_ENUM only makes sense if you want to expose the enumeration, but the enum class itself makes sense, regardless of whether you intend to expose it or not.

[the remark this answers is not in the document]
christophehenry: If I don't expose it, I'd have to write something like:

```js
engine.getPlayer("[Channel1]").stateChanged.connect(state => {
    if(state === 1) { /* … */ }
});
```

It's not ideal, is it? I can"t find how to expose the enum so I can use it like so:

```js
const player = engine.getPlayer("[Channel1]");
stateChanged.connect(state => {
    if(state === player.State.LOADED) { /* … */}
});
```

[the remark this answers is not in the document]
christophehenry: > Whats the purpose for stateChanged in the first place?

So you can display something when the text is loading. IDK, maybe it's not useful. The loading stage seems fast enough so that the turntable won't have the time to display anything anyway. If it's not, just tell me, I'm happy to throw it away. Less code to maintain and test is better.

src/controllers/scripting/controllerscriptenginebase.cpp
christophehenry: Can I resolve this so it's clearer for me what's left?

src/controllers/scripting/javascriptplayerproxy.h
christophehenry: Can I resolve this so it's clearer for me what's left?

src/controllers/scripting/legacy/controllerscriptinterfacelegacy.cpp
```diff
@@ -136,6 +136,15 @@ QJSValue ControllerScriptInterfaceLegacy::getSetting(const QString& name) {
     }
 }
 
+QObject* ControllerScriptInterfaceLegacy::getPlayer(const QString& deck) {
+    auto* const player = m_pScriptEngineLegacy->getPlayer(deck);
+    auto* const jsEngine = qjsEngine(this);
+    auto stateEnum = jsEngine->newQMetaObject(
+            QMetaType::fromType<JavascriptPlayerProxy::State>().metaObject());
+    jsEngine->newQObject(player).setProperty("State", stateEnum);
+    return player;
```
christophehenry, line 145: Should I return the result of `jsEngine->newQObject` here? I'm unsure but I suppose `Q_INVOKABLE` transforms the `QObject` to a `QJSValue` so since that's what I'm already doing here, better directly return it?

src/controllers/scripting/javascriptplayerproxy.h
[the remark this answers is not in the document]
Swiftb0y: yeah, lets only add it once we sure we need it.

christophehenry:
Should this be tested? If so, how can I test?

res/controllers/engine-api.d.ts
```diff
@@ -31,10 +34,38 @@ declare interface ScriptConnection {
...
+    artistChanged: QtSlot<(newArtist: string) => void>
+    titleChanged: QtSlot<(newTitle: string) => void>
+    albumChanged: QtSlot<(newAlbum: string) => void>
+    albumArtistChanged: QtSlot<(newAlbumArtist: string) => void>
+    genreChanged: QtSlot<(newGenre: string) => void>
+    composerChanged: QtSlot<(newComposer: string) => void>
+    groupingChanged: QtSlot<(newGrouping: string) => void>
+    yearChanged: QtSlot<(newYear: string) => void>
+    trackNumberChanged: QtSlot<(newTrackNumber: string) => void>
+    trackTotalChanged: QtSlot<(newTrackTotal: string) => void>
```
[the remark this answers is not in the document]
Swiftb0y: No. the `.connect` in QJS works like QML and `QObject::connect`. You usually initialize the value from the property directly and afterwards subscribe to updates. Alternatively you can also subscribe first and then call the handler directoy by calling the signal property. Maybe you can show me some code so I can better understand the usecase or ergonomics you're trying to achieve?

Swiftb0y:
Well you can create a test that enumerates all the properties to ensure API stability...

res/controllers/engine-api.d.ts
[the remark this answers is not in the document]
Swiftb0y: yeah, in this case you would have to call `engine.getPlayer("[Channel1]").titleChanged(engine.getPlayer("[Channel1]").title)`. But the same also applies to the `engine.makeConnection` API. In that case you do get a handler back and can call `trigger` on it. I don't think there is a `engine.getPlayer("[Channel1]").titleChanged.trigger()` on Qt signals. 
From QML, IIRC its also common for the signal to actually emit the value but instead query it explicitly from the handler (at least thats the pattern in Qt Quick iirc). Afaict thats explicitly to make it easier to emit the signal. It also avoids overhead when the connected handler only wants the signal but doesn't actually needs the value. 
This page gives a good overview on whats possible with signals (just ignore the QML specific parts): https://doc.qt.io/qt-6/qtqml-syntax-signals.html#connecting-signals-to-methods-and-signals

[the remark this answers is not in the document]
christophehenry: Ok. I'm happy to leave it like this and we handle the burden in ComponentJS.

Swiftb0y:
No worries. I wouldn't even go as far as that. At that point you're really just testing Qt's library functions (which we should just trust their tests to verify). In this case you can just create a list of properties and then verify whether they're available on `QMetaObject` of the `JavascriptPlayerProxy`. I personally think even thats overkill, but @JoergAtGithub wants some way to ensure API stability and this would be a good way to do that. The other way would be to create a `QJSengine` instance and then `evaluate` some code within that, but IMO thats on the hand overkill while still not being fully adequate...

src/controllers/scripting/controllerscriptenginebase.cpp
```diff
@@ -30,6 +30,11 @@ ControllerScriptEngineBase::ControllerScriptEngineBase(
     qRegisterMetaType<QMessageBox::StandardButton>("QMessageBox::StandardButton");
 }
 
+void ControllerScriptEngineBase::registerPlayerManager(
+        std::shared_ptr<PlayerManager> pPlayerManager) {
+    ControllerScriptEngineBase::s_pPlayerManager = pPlayerManager;
+}
```
Swiftb0y, line 36: we should try to eliminate the static here. Would be be feasible to pass the PlayerManager in the `ControllerScriptEngineBase` constructor?

src/controllers/scripting/javascriptplayerproxy.cpp
```diff
@@ -0,0 +1,112 @@
...
+    emit artistChanged(m_pCurrentTrack->getArtist());
+    emit titleChanged(m_pCurrentTrack->getTitle());
+    emit albumChanged(m_pCurrentTrack->getAlbum());
+    emit albumArtistChanged(m_pCurrentTrack->getAlbumArtist());
+    emit genreChanged(m_pCurrentTrack->getGenre());
+    emit composerChanged(m_pCurrentTrack->getComposer());
+    emit groupingChanged(m_pCurrentTrack->getGrouping());
+    emit yearChanged(m_pCurrentTrack->getYear());
+    emit trackNumberChanged(m_pCurrentTrack->getTrackNumber());
+    emit trackTotalChanged(m_pCurrentTrack->getTrackTotal());
```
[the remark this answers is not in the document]
christophehenry: > The track may change but the artist hasn't.

Those are emitted when the track is loaded to the deck so yes.

> JS won't receive the news until the track has been reloaded

I tested that case and yes, the signal is emitted when you changed the ID3 tag.

christophehenry:
> In this case you can just create a list of properties and then verify whether they're available on QMetaObject of the JavascriptPlayerProxy.

I don't really see the point in testing the presence of properties on `QMetaObject`. What I had in mind was more to test that all the signals are correctly triggered with the correct values when a track is loaded. But I don't know how to spy a signal with gMock or load a track within a test. Or event where to put the test code.

src/controllers/scripting/javascriptplayerproxy.cpp
```diff
@@ -0,0 +1,112 @@
...
+                   << pOldTrack.get() << "as active track but got"
+                   << m_pCurrentTrack.get();
+    }
+
+    if (pNewTrack == m_pCurrentTrack) {
+        return;
+    }
+
+    if (m_pCurrentTrack != nullptr) {
+        disconnect(m_pCurrentTrack.get(), nullptr, this, nullptr);
```
Swiftb0y, line 98: NIT: easier to read
```suggestion
        m_pCurrentTrack->disconnect(this);
```

Swiftb0y:
Welp, I have nothing to complain which I've not already been overruled on. @JoergAtGithub please give your LGTM too.

christophehenry:
Good. Lemme squash the commits.

JoergAtGithub (changes requested):
The C++ code is LGTM! TS declaration needs verbose documention as all other APIs.
Any new API needs at least a basic test case. Please implement one in src\test\controllerscriptenginelegacy_test.cpp

christophehenry:
>  Any new API needs at least a basic test case. Please implement one in src\test\controllerscriptenginelegacy_test.cpp

Ok then can you help me implement them? I know how to test the getters in theory but I don't know how to load a track to the player and test the slots.

JoergAtGithub:
> > Any new API needs at least a basic test case. Please implement one in src\test\controllerscriptenginelegacy_test.cpp
> 
> Ok then can you help me implement them? I know how to test the getters in theory but I don't know how to load a track to the player and test the slots.

Yes, I can look into this! But I might need some time.

res/controllers/engine-api.d.ts
```diff
@@ -31,10 +34,50 @@ declare interface ScriptConnection {
...
```
christophehenry, line 67: I wrote a global documentation for those I don't think it's necessary to document them individually?

JoergAtGithub: It only appear in the editor/IDE if you document it per item. This file is not intended for human readers that read it from top to down.
This way the user sees it as 
grafik
but only for artistChanged.

Please add documentation for each item therefore, as in the rest of the typescript declaration files.

res/controllers/engine-api.d.ts
```diff
@@ -31,10 +34,50 @@ declare interface ScriptConnection {
...
+    readonly artist: string
+    readonly title: string
+    readonly album: string
+    readonly albumArtist: string
+    readonly genre: string
+    readonly composer: string
+    readonly grouping: string
+    readonly year: string
+    readonly trackNumber: string
+    readonly trackTotal: string
```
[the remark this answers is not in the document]
JoergAtGithub: I doubt that terms like trackTotal or trackNumber are self explaining for an average mapping developer. Better document too much than too less.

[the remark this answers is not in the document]
christophehenry: Ok, done.

Swiftb0y:
friendly ping @christophehenry

christophehenry:
@Swiftb0y Did you mean to ping @JoergAtGithub? I'm waiting for their input on how to load a track to the player and test the slots.

Swiftb0y:
Ah, I see. I didn't know where we were. Then this is indeed a friendly reminder @JoergAtGithub

christophehenry:
Ah I see @JoergAtGithub opened christophehenry/mixxx#3. I don't why I didn't get notified… I may be able to progress this week.

christophehenry:
Alright! Tests done!

*Edit:* ah… I should add a test for the `trackUnloaded` slot too. I'll do tonight. In the meantime, don't hesitate to leave your comments!

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -818,11 +918,61 @@ TEST_F(ControllerScriptEngineLegacyTest, convertCharsetAllCharset) {
...
+            false);
+    m_pEngine->process(1024);
+    while (!deck->getEngineDeck()->getEngineBuffer()->isTrackLoaded()) {
+        QTest::qSleep(100);
+    }
+    processEvents();
+    for (auto [key, value] : expectedValues.asKeyValueRange()) {
+        EXPECT_QSTRING_EQ(value, qjsvalue_cast<QString>(evaluate(QString("player.%1").arg(key))));
+        EXPECT_QSTRING_EQ(value, qjsvalue_cast<QString>(evaluate(QString("result.%1").arg(key))));
+    }
```
christophehenry: Is there a GoogleTest equivalent to [Python's `unittest.TestCase.subTest`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.subTest)? Couldn't find one.

Swiftb0y: Not really. The decoding test has some code that tries to accomplish something similar, but I don't want you to reuse that. Looking at your test code however, I think you'll want to write a [Value-Parameterized Test](https://google.github.io/googletest/advanced.html#value-parameterized-tests).

christophehenry: I think I'll just go with a custom failure message.

Swiftb0y: thats fine too, though a parametrized test wouldn't really be that much complicated IMO.

src/test/controllerscriptenginelegacy_test.cpp
```diff
@@ -57,13 +79,104 @@ class ControllerScriptEngineLegacyTest : public ControllerScriptEngineLegacy, pu
...
+        }
+        processEvents();
+    }
+
+    void unloadTrackSync() {
+        auto* deck = m_pPlayerManager->getDeck(1);
+        deck->slotEjectTrack(1.0);
+        while (deck->getEngineDeck()->getEngineBuffer()->isTrackLoaded()) {
+            QTest::qSleep(100);
+        }
```
christophehenry: Sadly, this doesn't work. The test just hangs. If I execute this in the debugger with breakpoints, `deck->slotEjectTrack(1.0)` correctly ejects the track and the test passes. So I'm not correctly waiting here. There's a concurrency problem. I'd accept any help on how to solve this.

JoergAtGithub: Did you try to eject from JavaScript using `engine.setValue('[Channel1]', 'eject', 1);` ?

christophehenry: I admit I didn't have that idea. Lemme try.
*Edit:* @JoergAtGithub hmm… This solution seems to not even eject the track. Even in the debugger, the event aren't fired :thinking:

JoergAtGithub: I would be fine if the testcase doens't check the eject case.

christophehenry: Alright, deal.

Swiftb0y: maybe also `processEvents()` in the busy-wait loop?

christophehenry: Sadly no, that still hangs or not even eject, depending on the solution. I tried a number of combinations around this.

JoergAtGithub: You probably need to call it twice, as in other tests: https://github.com/mixxxdj/mixxx/blob/14c19a55b9ab0c4bde94d7f61033206081dd698d/src/test/controlobjectscripttest.cpp#L85-L91

christophehenry: I called the [`processEvents`](https://github.com/christophehenry/mixxx/blob/JavascriptPlayerProxy/src%2Ftest%2Fcontrollerscriptenginelegacy_test.cpp#L185-L194) of the test class which already does that. But it doesn't make a difference.

JoergAtGithub (approved):
LGTM! Thank you for your contribution and your patience!

daschuer:
After merging this Mixxx crashed during shutdown: See https://github.com/mixxxdj/mixxx/issues/14982

Eve00000:
Congratulations for the work and perseverance.
(Too bad for the crash on exit now)

christophehenry:
Awesome! It's the last piece I needed to propose an official DN-S3700 controller ❤️

PR #11001, opened 2022-10-24
Don't reject file metadata Keys values with non Minor/Major scales.  

daschuer:
This fixes https://github.com/mixxxdj/mixxx/issues/10995

It allows to read keys entries created by the Rapid evolution "detailed key mode" 
"Ab ionian" or "04I"

In Mixxx they are displayed as a compatible Minor/Major key.

PR #14318, opened 2025-02-10
Add case-insensitive parsing for Lancelot key notation

cr7pt0gr4ph7:
When a user types in `key:5a` into the search field, the intent is most likely to search for `key:5A` instead of doing a text-only search.

The only case where case-insensitive parsing is ambigous is for the `d/m` (OpenKey) and `D/M` (Lancelot) suffixes, where lowercase will imply OpenKey, and uppercase will imply Lancelot (due to the order in which `guessKeyFromText` checks for matches with the different notations).

### DIFF
--- a/src/track/keyutils.cpp
+++ b/src/track/keyutils.cpp
@@ -27,7 +27,7 @@ const QRegularExpression s_openKeyRegex(QStringLiteral(
 // Lancelot notation, the numbers 1-12 followed by A (minor) or B(I) (major).
 // or "I", "L", "M", "D", "P", "C" for the advanced modes
 const QRegularExpression s_lancelotKeyRegex(
-        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPCabilmdpc])\\s*$)\\z"));
+        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPC])\\s*$)\\z"), QRegularExpression::CaseInsensitiveOption);
 constexpr std::string_view s_lancelotMajorModes = "BILMbilm";
 
 // a-g followed by any number of sharps or flats, optionally followed by
--- a/src/track/keyutils.cpp
+++ b/src/track/keyutils.cpp
@@ -27,8 +27,8 @@ const QRegularExpression s_openKeyRegex(QStringLiteral(
 // Lancelot notation, the numbers 1-12 followed by A (minor) or B(I) (major).
 // or "I", "L", "M", "D", "P", "C" for the advanced modes
 const QRegularExpression s_lancelotKeyRegex(
-        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPC])\\s*$)\\z"));
-constexpr std::string_view s_lancelotMajorModes = "BILM";
+        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPCabilmdpc])\\s*$)\\z"));
+constexpr std::string_view s_lancelotMajorModes = "BILMbilm";
 
 // a-g followed by any number of sharps or flats, optionally followed by
 // a scale mode spec (m = minor, min, maj ..)
--- a/src/track/keyutils.cpp
+++ b/src/track/keyutils.cpp
@@ -27,8 +27,9 @@ const QRegularExpression s_openKeyRegex(QStringLiteral(
 // Lancelot notation, the numbers 1-12 followed by A (minor) or B(I) (major).
 // or "I", "L", "M", "D", "P", "C" for the advanced modes
 const QRegularExpression s_lancelotKeyRegex(
-        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPC])\\s*$)\\z"), QRegularExpression::CaseInsensitiveOption);
-constexpr std::string_view s_lancelotMajorModes = "BILMbilm";
+        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPC])\\s*$)\\z"),
+        QRegularExpression::CaseInsensitiveOption);
+constexpr std::string_view s_lancelotMajorModes = "BILM";
 
 // a-g followed by any number of sharps or flats, optionally followed by
 // a scale mode spec (m = minor, min, maj ..)
@@ -460,8 +461,9 @@ ChromaticKey KeyUtils::guessKeyFromText(const QString& text) {
         int openKeyNumber = lancelotNumberToOpenKeyNumber(lancelotNumber);
 
         const QChar lancelotScaleMode = lancelotMatch.captured(2).at(0);
-        bool major = (s_lancelotMajorModes.find(lancelotScaleMode.toLatin1()) != std::string::npos);
-
+        bool major = (s_lancelotMajorModes.find(
+                              lancelotScaleMode.toUpper().toLatin1()) !=
+                std::string::npos);
         return openKeyNumberToKey(openKeyNumber, major);
     }
 
--- a/src/test/keyutilstest.cpp
+++ b/src/test/keyutilstest.cpp
@@ -32,6 +32,12 @@ TEST_F(KeyUtilsTest, LancelotNotation) {
     EXPECT_EQ(mixxx::track::io::key::C_SHARP_MINOR,
               KeyUtils::guessKeyFromText("12A"));
 
+    // Allow lower-case to be more flexible when parsing search queries
+    EXPECT_EQ(mixxx::track::io::key::C_MAJOR,
+            KeyUtils::guessKeyFromText("8b"));
+    EXPECT_EQ(mixxx::track::io::key::C_SHARP_MINOR,
+            KeyUtils::guessKeyFromText("12a"));
+
     // whitespace is ok
     EXPECT_EQ(mixxx::track::io::key::B_MINOR,
             KeyUtils::guessKeyFromText("\t10A  "));


cr7pt0gr4ph7:
I know that case-insensitivity was [intentionally removed](https://github.com/mixxxdj/mixxx/pull/11001/commits/6a88dba9ada253ecd51b84aa72acf046fc1becb9#r1286658757) when adding the advanced Lancelot notations "ILMDPC" (though I haven't yet found a document describing which scales these are associated with), but this has caused a sub-par user experience:

Typing `key:2a` into the search field will fail to parse the provided key notation, and fall back to a case-insensitive text search in the `key` column instead, which will then return results that are _nearly_ correct:
- :heavy_check_mark:  Searching for e.g. `key:5a` will only turn up results with `key == 5A`.
- :x: Searching for `key:2a` will turn up results with `key == 2A || key == 12A`
- :x: Fuzzy notation `~key:5a` will not work, while `~key:5A`.

We can avoid the ambiguity issue for `d/m` (OpenKey) vs. `D/M` (Lancelot) by depending on the lowercase/uppercase distinction there, while still providing case-insensitivity for the more common `A/B` (/ `a/b`) (Lancelot) suffix.

src/track/keyutils.cpp
```diff
@@ -27,8 +27,8 @@ const QRegularExpression s_openKeyRegex(QStringLiteral(
 // Lancelot notation, the numbers 1-12 followed by A (minor) or B(I) (major).
 // or "I", "L", "M", "D", "P", "C" for the advanced modes
 const QRegularExpression s_lancelotKeyRegex(
-        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPC])\\s*$)\\z"));
-constexpr std::string_view s_lancelotMajorModes = "BILM";
+        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPCabilmdpc])\\s*$)\\z"));
```
acolombier, line 30: You should be able to use the the `CaseInsensitiveOption` [regex option](https://doc.qt.io/qt-6/qregularexpression.html#PatternOption-enum). This will help keeping the regex readable and may also help with processing performance 
```suggestion
        QStringLiteral("\\A(?:^\\s*0*(1[0-2]|[1-9])([ABILMDPC])\\s*$)\\z"), QRegularExpression::CaseInsensitiveOption);```

File: src/track/keyutils.cpp [hunk 2]
```diff
@@ -27,8 +27,8 @@ const QRegularExpression s_openKeyRegex(QStringLiteral(
...
+constexpr std::string_view s_lancelotMajorModes = "BILMbilm";
```
acolombier, line 31: It might be easier to read to us `toLower()` on line 462. `toLower` will effectively make a single subtraction, while doubling the size might significantly decrease the `find` performance. Likely invisible on our modern CPU, but we need to remain mindful of the slimer specs such as old RPi :smile:

acolombier (changes requested):
Thanks for this nice improvement! Some minor change request but otherwise good to go

cr7pt0gr4ph7:
@acolombier
> It might be easier to read to us `toLower()` on line 462. `toLower` will effectively make a single subtraction, while doubling the size might significantly decrease the `find` performance. Likely invisible on our modern CPU, but we need to remain mindful of the slimer specs such as old RPi 😄

We could also use a binary tree hash for more efficient matching, or even a bloom filter. Okay, I'm just joking now 😄 Converted the code to use `toUpper()` instead.

> Thanks for this nice improvement! Some minor change request but otherwise good to go.

I originally wanted to implement `~key:5a` matching - just to find out after one year of using Mixxx that it's actually already present, and it just wasn't parsing `5a` as expected...

cr7pt0gr4ph7:
On a more serious note, after thinking about parsing of the `d/D` and `m/M` prefix: As a user typing in a search query, I would probably still expect `5d` to be parsed using Lancelot notation when I have selected `KeyUtils::KeyNotation::Lancelot` or `KeyUtils::KeyNotationLancelotAndTraditional` as the default key notation in the settings.

What do you think?

daschuer:
The additionals modes have been introduced by Rapid Ecolution. See: 
https://www.myqnap.org/product/rapid-evolution-3/
https://github.com/djqualia/RapidEvolution3/blob/6ef912c1843bf5201068973434237bb378c5e02f/src/com/mixshare/rapid_evolution/music/key/Key.java#L56

acolombier:
> I originally wanted to implement `~key:5a` matching - just to find out after one year of using Mixxx that it's actually already present, and it just wasn't parsing `5a` as expected...

I was looking at this some time ago and realise the search feature was actually quite complete! I think our UI doesn't give it good usability, but this is something as part of the new Mixxx design :) 

> I would probably still expect `5d` to be parsed using Lancelot notation when I have selected `KeyUtils::KeyNotation::Lancelot` or `KeyUtils::KeyNotationLancelotAndTraditional` as the default key notation in the settings.

I agree - I feel like search should allow you to use any key annotation. independently of the display preference. Searching for `4d` wile having Lancelot display selected, should still work, even if the results display `11b` in the `Key` column. Ideally, the column should provide context about the search input and display both the preferred annotation + the input one (e.g `11b (4d)`), but I know working with our library model is quite challenging, so I'm not sure this small improvement is worth the effort.

Alternatively, we could display a warning message and CTA to update the selected key notation in one click.

Appreciate this may become a fairly large change, so happy to see this coming as part of a follow up PR if you would prefer!

daschuer:
> And, vice versa, 5D to be parsed as 5d when KeyUtils::KeyNotation::OpenKey is selected

I think we can very much assume that the usere searches for the notation he also sees in the library. We just to need to be carefull to not mess with metadata read form the file tags. Here we can't use this assumption.

cr7pt0gr4ph7:
@daschuer
> The additionals modes have been introduced by Rapid Ecolution. See: https://www.myqnap.org/product/rapid-evolution-3/ https://github.com/djqualia/RapidEvolution3/blob/6ef912c1843bf5201068973434237bb378c5e02f/src/com/mixshare/rapid_evolution/music/key/Key.java#L56

Thanks! Googling for [`lancelot key notation`](https://www.google.com/search?q=lancelot+key+notation) did not turn up any relevant results at all.

For future reference, [this page](https://www.myqnap.org/product/rapid-evolution-3/) contains a reference on what the "advanced" Lancelot modes actually mean:

> Basic key notation
> - A - minor
> - B - major
>
> Advanced key notation
> – A – Aeolian (natural minor)
> – I – Ionian (normal major, normally B)
> – D – Dorian
> – P – Phrygian
> – L – Lydian
> – M – Mixolydian
> – C – Locrian

cr7pt0gr4ph7:
@acolombier
> Appreciate this may become a fairly large change, so happy to see this coming as part of a follow up PR if you would prefer!

Yes, I think that this PR is good to go as-is (provided you also agree) as it's a 80% solution. As it turns out, though [there is infrastructure in place](https://github.com/mixxxdj/mixxx/blob/799ff78ae19a17a0e4fdc844f2d13fdf60e25e0c/src/track/keyutils.cpp#L201) for parsing different scale modes like `Dorian`, `Phrygian` etc., it is not currently utilized for Lancelot notation... which might be considered a different bug.

@daschuer
> I think we can very much assume that the user searches for the notation he also sees in the library.

The other table columns all use case-insensitive text matching of what the user can visually see, so its very reasonable for `key` to adhere to that as close as possible.

> We just to need to be carefull to not mess with metadata read form the file tags. Here we can't use this assumption.

I agree. My thought was that there should probably be two parsing modes:
- A strict one were the parsing mode is always deterministic and non-configurable[^1] (i.e. the current `guessKeyFromText`).
- A fuzzy one where an expected key notation can be passed in, to be used in places were we coerce user input into our internal representation (but not for parsing track metadata).
  Custom key notations might require some more consideration, as these should IMO probably be parsed in a case-insensitive as well, except for case is required to distinguish two different keys.
  
The necessary infrastructure already seems to be in place for this, as e.g. [`TrackRecord::updateGlobalKeyNormalizeText`](https://github.com/mixxxdj/mixxx/blob/799ff78ae19a17a0e4fdc844f2d13fdf60e25e0c/src/track/trackrecord.cpp#L30) already receives a `source` parameter that can have the values `mixxx::track::io::key::Source::ANALYZER`, `FILE_METADATA` and `USER`, and we could just pass that on to `guessKeyFromText`.

Eve00000:
@cr7pt0gr4ph7 
while you're taking a look at the searchLineEdit, can you also take a look at src/library/trackset/smarties/smartiesfeaturehelper.cpp in [#14182](https://github.com/mixxxdj/mixxx/pull/14182/files#diff-6671f413cdae54b3d70cdbfc67186c930bbf26e5c43b8816a40ac6b3d912405f) ,
in this file the conversion from searchLineEdit to smarties (smartcrates) is done: create the conditions to save the search.
So I would realy appreciate it if you could enter the updated search for keys there to.

acolombier (approved):
Code LGTM, thanks you!

cr7pt0gr4ph7:
@Eve00000 
> @cr7pt0gr4ph7 while you're taking a look at the searchLineEdit, can you also take a look at src/library/trackset/smarties/smartiesfeaturehelper.cpp in [#14182](https://github.com/mixxxdj/mixxx/pull/14182/files#diff-6671f413cdae54b3d70cdbfc67186c930bbf26e5c43b8816a40ac6b3d912405f) , in this file the conversion from searchLineEdit to smarties (smartcrates) is done: create the conditions to save the search. So I would realy appreciate it if you could enter the updated search for keys there to.

I had a look at #14182, and it seems to me like `SmartiesFeatureHelper::createEmptySmartiesFromSearch` basically duplicates `SearchQueryParser::parseQuery` (but not completely, some features like e.g. `~key:5A` which should match all compatible keys, i.e. `4A`, `5A`, `5B` and `6A`, are missing).

I can't think of a good reason for why the search syntax should differ between Smarties/SmartCrates vs. the searchLineEdit (and for why that would not be really confusing for users). Given that assumption, I don't see why it shouldn't be possible possible to use `SearchQueryParser::parseQuery` to parse the queries to a `QueryNode` AST, and then turn that AST into the format required for the smart crates.

PR #14321, opened 2025-02-11
Fix and improve missing env error message 

daschuer:
The detection that we are on Linux was broken.
This is fixed now. In addition I added a hint for Apple arm64 builds and detection of rat head and Debian based distros.


cr7pt0gr4ph7 (approved):
LGTM :+1: I also don't expect this to break anything, as it's only used when the build breaks anyway.

acolombier (approved):

PR #14323, opened 2025-02-12
Fix: Avoid timer warning on button release

git-developer:
* _Given_ a `Button` with `type == Button.types.powerWindow` is bound to a MIDI controller and no _press_ command (`0x7F`) was sent yet
* _When_ the controller sends a _release_ command (`0x00`)
* _Then_ a warning is printed to the log: `Warning [Controller] "Tried to kill Timer \"0\" that does not exist"`

Cause: When the button is initialized, `this.longPressTimer` is `undefined`. Comparing `!== 0` returns true, so a non-existing timer is stopped.

### DIFF
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -27,6 +27,7 @@
  */
 
 (function(global) {
+    const NO_TIMER = 0;
     const Component = function(options) {
         if (Array.isArray(options) && typeof options[0] === "number") {
             this.midi = options;
@@ -176,6 +177,8 @@
         // in any Buttons that act differently with short and long presses
         // to keep the timeouts uniform.
         longPressTimeout: 275,
+        isLongPressed: false,
+        longPressTimer: NO_TIMER,
         isPress: function(channel, control, value, _status) {
             return value > 0;
         },
@@ -192,15 +195,15 @@
                     this.isLongPressed = false;
                     this.longPressTimer = engine.beginTimer(this.longPressTimeout, () => {
                         this.isLongPressed = true;
-                        this.longPressTimer = 0;
+                        this.longPressTimer = NO_TIMER;
                     }, true);
                 } else {
                     if (this.isLongPressed) {
                         this.inToggle();
                     }
-                    if (this.longPressTimer !== 0) {
+                    if (this.longPressTimer !== NO_TIMER) {
                         engine.stopTimer(this.longPressTimer);
-                        this.longPressTimer = 0;
+                        this.longPressTimer = NO_TIMER;
                     }
                     this.isLongPressed = false;
                 }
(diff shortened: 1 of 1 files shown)


res/controllers/midi-components-0.0.js
```diff
@@ -198,7 +198,7 @@
                     if (this.isLongPressed) {
                         this.inToggle();
                     }
-                    if (this.longPressTimer !== 0) {
+                    if (this.longPressTimer !== undefined && this.longPressTimer !== 0) {
```
acolombier, line 201: Do I dare suggesting checking for a falsy value?

```suggestion
                    if (this.longPressTimer) {
```

acolombier: I'm not sure if there could be any side effect here

Swiftb0y: generally not a fan of this pattern because JS implicit conversions to bool are somewhat of a footgun.

acolombier: Yeah I was expecting this feedback. My main frustration is the case it wouldn't handle something like `null`, meaning that once the timer as run, you are expected to set it to `0` or `undefined` which can feel a bit weird.

Anyway, I'm nitpicking here, happy to go with the current solution!

Swiftb0y (changes requested):
Thanks. This is a good point. I'd like to suggest an alternative solution however: Since the code assumes `0` is the sentinel value for a stopped timer, instead of introducing a second sentinel (`undefined`) lets just ensure the `Button` is initialized with the correct sentinel in the first place. That would be a oneline change, simply add `longPressTimer: 0` somewhere in the config object (for example above `isPress`). For additional brownie points, add a static constant and replace the occurences of `0` with references to that constant (refactor out the magic constant). Makes sense?

res/controllers/midi-components-0.0.js
Swiftb0y: yeah, `null` is also weird. Ideally we'd introduce a named constant (see request below).

git-developer:
Done, together with some unifications in derived buttons. Timers are handled differently on each place they occur. The changes have been tested, but please review nonetheless, it's a central location.

BTW this kind of duplication is the reason for a dedicated [Timer](https://github.com/mixxxdj/mixxx/blob/9d2f6f67d457389c4d3b9db11789939e6f809c5f/res/controllers/Behringer-Extension-scripts.js#L233) object in the Behringer extensions.

res/controllers/midi-components-0.0.js
```diff
@@ -1134,8 +1137,6 @@
 
         this.effectFocusButton = new Button({
             group: this.group,
-            longPressed: false,
```
git-developer, line 1137: This should have been `isLongPressed` I guess, but isn't used anyway.

git-developer:
Build is broken, but I guess this is due to an unrelated error in `Reloop-Digital-Jockey-2-IE.scripts.js` (looks like required deck number is not available on prototype initialization).

Swiftb0y (approved):
Yup. Failure is unrelated. LGTM. Thank you.

acolombier:
Could you please merge 2.5 in your branch so we can get all CI checks to pass before merging?

git-developer:
Should I squash or can that be done by the maintainer on merge?

git-developer:
OK, so: maintainer approves -> committer squashes -> maintainer merges?

git-developer:
Is there a schedule to to merge `2.5` into `main`?

Swiftb0y:
> Is there a schedule to to merge 2.5 into main?

Not really, we just do it whenever someone thinks about it. Thanks for the reminder.

> And is [squashing for pull requests](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests) disabled on purpose?

Yes. They don't create a merge commit which the changelog generator script relies upon.

PR #14324, opened 2025-02-12
add WCollapsibleGroupBox, use for controller mapping info boxes and settings

ronso0:
This adds a QGroupBox-derived `WCollapsibleGroupBox` that can be collapsed/expanded by clicking its title.
Screenshot_2025-02-12_18-16-08

IMHO this fixes the regression with the new tab'ed controller page where the device and mapping info would push the setting and I/O tabs down.
It also allows to focus on a certain settings group when adjusting the mapping.
* Down/Right arrow like on treeview branches
* state of Device / Mapping info boxes is persistent per session (per controller)
* state of settings boxes is restored / persists as long as the mapping is selected
(they are expanded every time the controller page is shown because the controller settings are rebuild each time)

`WCollapsibleGroupBox` abuses [`QGroupBox::setChecked()`](https://doc.qt.io/qt-6/qgroupbox.html#checked-prop) mechanism. Originally this is a checkbox in the title bar and toggling it enables/disables all children. 
Adopting this saves us from reimplementing the checked state tracking, signal etc.
This is not touched but also irrelevant since the content is now hidden anyway when it's collapsed.
Collapsing is achieved by simply setting the maximum height to height of the title row.
Inspired by [ctkCollapsibleGroupBox](https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkCollapsibleGroupBox.cpp), just without the complexity of hiding/showing child widgets (which requires some small, low-risk hacks to get the appearance right under all circumstances).

I imagine this groupbox can be helpful in other preferences pages, too.
(as quick fix for long pages, compared to restructuring into tab or otherwise splitting pages)

### DIFF
--- a/src/controllers/dlgprefcontroller.cpp
+++ b/src/controllers/dlgprefcontroller.cpp
@@ -804,6 +804,9 @@ void DlgPrefController::slotMappingSelected(int chosenIndex) {
             !mappingFilePath.isEmpty();
     m_ui.controllerTabs->setTabVisible(m_inputMappingsTabIndex, showMidiTabs);
     m_ui.controllerTabs->setTabVisible(m_outputMappingsTabIndex, showMidiTabs);
+    // Also disable accordingly so hidden tabs can not get keyboard focus
+    m_ui.controllerTabs->setTabEnabled(m_inputMappingsTabIndex, showMidiTabs);
+    m_ui.controllerTabs->setTabEnabled(m_outputMappingsTabIndex, showMidiTabs);
 
     // Hide the entire QTabWidget if all tabs are removed
     m_ui.controllerTabs->setVisible(getNumberOfVisibleTabs() > 0);
@@ -1115,8 +1118,9 @@ void DlgPrefController::slotShowMapping(std::shared_ptr<LegacyControllerMapping>
     }
 
     // Show or hide the settings tab based on the presence of settings
-    m_ui.controllerTabs->setTabVisible(
-            m_settingsTabIndex, pMapping && !pMapping->getSettings().isEmpty());
+    bool showSettings = pMapping && !pMapping->getSettings().isEmpty();
+    m_ui.controllerTabs->setTabVisible(m_settingsTabIndex, showSettings);
+    m_ui.controllerTabs->setTabEnabled(m_settingsTabIndex, showSettings);
 
     // If there is still settings that may be saved and no new mapping selected
     // (e.g restored default), we keep the the dirty mapping live so it can be
@@ -1135,11 +1139,13 @@ void DlgPrefController::slotShowMapping(std::shared_ptr<LegacyControllerMapping>
         auto screens = pMapping->getInfoScreens();
         bool hasScreens = !screens.isEmpty();
         m_ui.controllerTabs->setTabVisible(m_screensTabIndex, hasScreens);
+        m_ui.controllerTabs->setTabEnabled(m_screensTabIndex, hasScreens);
         if (hasScreens) {
             slotShowPreviewScreens(m_pController->getScriptEngine().get());
         }
     } else {
         m_ui.controllerTabs->setTabVisible(m_screensTabIndex, false);
+        m_ui.controllerTabs->setTabEnabled(m_screensTabIndex, false);
     }
 #endif
 
--- a/src/controllers/dlgprefcontroller.cpp
+++ b/src/controllers/dlgprefcontroller.cpp
@@ -29,6 +29,7 @@
 #include "util/desktophelper.h"
 #include "util/parented_ptr.h"
 #include "util/string.h"
+#include "widget/wcollapsiblegroupbox.h"
 
 namespace {
 const QString kMappingExt(".midi.xml");
@@ -761,6 +762,7 @@ void DlgPrefController::slotMappingSelected(int chosenIndex) {
         if (m_pControllerManager->getConfiguredMappingFileForDevice(
                     m_pController->getName()) != mappingFilePath) {
             setDirty(true);
+            m_settingsCollapsedStates.clear();
         } else if (m_pMapping && m_pMapping->isDirty()) {
             // We have pending changes, don't reload the mapping from file!
             // This is called by show()/slotUpdate() after MIDI learning ended
@@ -1068,7 +1070,40 @@ void DlgPrefController::slotShowMapping(std::shared_ptr<LegacyControllerMapping>
         }
 
         if (pLayout != nullptr && !settings.isEmpty()) {
-            m_ui.settingsTab->layout()->addWidget(pLayout->build(m_ui.settingsTab));
+            QWidget* pSettingsWidget = pLayout->build(m_ui.settingsTab);
+            m_ui.settingsTab->layout()->addWidget(pSettingsWidget);
+
+            // Add an expanding spacer so that when we collapse all groups,
+            // they are pushed to the top.
+            m_ui.settingsTab->layout()->addItem(new QSpacerItem(
+                    1, 1, QSizePolicy::Minimum, QSizePolicy::Expanding));
+
+            // Make all top-level groupboxes checkable so we get the
+            // collapse/expand functionality. Qt::FindDirectChildrenOnly
+            // ensures we only iterate over the top-level groupboxes.
+            const QList<WCollapsibleGroupBox*> boxes =
+                    pSettingsWidget->findChildren<WCollapsibleGroupBox*>(
+                            QString() /* match any ObjectName */,
+                            Qt::FindDirectChildrenOnly);
+            for (auto* pBox : std::as_const(boxes)) {
+                const QString title = pBox->title();
+                pBox->setCheckable(true);
+                // The collapsed state is saved/restored via the groupbox' title.
+                // Note: If multiple top-levle groups happen to have the same title
+                // (which should not normally happen in well-behaved controller mappings,
+                // but is not strictly prohibited), the last one to be expanded/
+                // collapsed determines the state that will be restored.
+                if (m_settingsCollapsedStates.contains(title)) {
+                    pBox->setChecked(m_settingsCollapsedStates.value(title));
+                }
+
+                connect(pBox,
+                        &WCollapsibleGroupBox::toggled,
+                        this,
+                        [this, title](bool checked) {
+                            m_settingsCollapsedStates.insert(title, checked);
+                        });
+            }
 
             for (const auto& setting : std::as_const(settings)) {
                 connect(setting.get(),
--- a/src/controllers/dlgprefcontrollerdlg.ui
+++ b/src/controllers/dlgprefcontrollerdlg.ui
@@ -1152,6 +1152,8 @@
   <tabstop>comboBoxMapping</tabstop>
   <tabstop>groupBoxDeviceInfo</tabstop>
   <tabstop>groupBoxMappingInfo</tabstop>
+  <tabstop>labelMappingSupportLinks</tabstop>
+  <tabstop>labelMappingScriptFileLinks</tabstop>
   <tabstop>controllerTabs</tabstop>
  </tabstops>
  <resources/>
--- a/src/controllers/dlgprefcontrollerdlg.ui
+++ b/src/controllers/dlgprefcontrollerdlg.ui
@@ -100,384 +100,422 @@
      </property>
     </widget>
    </item>
+
    <item row="2" column="0">
-    <widget class="QGroupBox" name="groupBoxDeviceInfo">
+    <widget class="QWidget" name="deviceInfoGroupLayoutWidget">
      <property name="sizePolicy">
-      <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
+      <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
        <horstretch>0</horstretch>
        <verstretch>0</verstretch>
       </sizepolicy>
      </property>
-     <property name="title">
-      <string>Device Info</string>
-     </property>
-     <property name="alignment">
-      <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignVCenter</set>
-     </property>
-     <layout class="QGridLayout" name="gridLayoutDeviceInfo" rowminimumheight="1,1,1,1,1,1,1,1,1,1,0">
-      <property name="sizeConstraint">
-       <enum>QLayout::SetMaximumSize</enum>
-      </property>
-      <item row="0" column="0">
-       <widget class="QLabel" name="labelPhysicalInterface">
+     <layout class="QVBoxLayout" name="devInfoGroupLayout">
+      <item>
+       <widget class="WCollapsibleGroupBox" name="groupBoxDeviceInfo">
         <property name="sizePolicy">
-         <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
+         <sizepolicy hsizetype="Preferred" vsizetype="Minimum">
           <horstretch>0</horstretch>
           <verstretch>0</verstretch>
          </sizepolicy>
         </property>
-        <property name="text">
-         <string>Physical Interface:</string>
+        <property name="title">
+         <string>Device Info</string>
+        </property>
+        <property name="checkable">
+         <bool>true</bool>
         </property>
         <property name="alignment">
-         <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
+         <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignVCenter</set>
         </property>
+        <layout class="QGridLayout" name="gridLayoutDeviceInfo" rowminimumheight="1,1,1,1,1,1,1,1,1,1,0">
+         <property name="sizeConstraint">
+          <enum>QLayout::SetMaximumSize</enum>
+         </property>
+         <item row="0" column="0">
+          <widget class="QLabel" name="labelPhysicalInterface">
+           <property name="sizePolicy">
+            <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
+             <horstretch>0</horstretch>
+             <verstretch>0</verstretch>
+            </sizepolicy>
+           </property>
+           <property name="text">
+            <string>Physical Interface:</string>
+           </property>
+           <property name="alignment">
+            <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
+           </property>
+          </widget>
+         </item>
+         <item row="0" column="1">
+          <widget class="QLabel" name="labelPhysicalInterfaceValue">
+           <property name="sizePolicy">
+            <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
+             <horstretch>0</horstretch>
+             <verstretch>0</verstretch>
+            </sizepolicy>
+           </property>
+           <property name="minimumSize">
+            <size>
+             <width>0</width>
+             <height>0</height>
+            </size>
+           </property>
+           <property name="text">
+            <string notr="true">(USB|Bluetooth|...)</string>
+           </property>
+          </widget>
+         </item>
+         <item row="1" column="0">
+          <widget class="QLabel" name="labelDataHandlingProtocol">
+           <property name="sizePolicy">
+            <sizepolicy hsizetype="Minimum" vsizetype="Minimum">
+             <horstretch>0</horstretch>
+             <verstretch>0</verstretch>
+            </sizepolicy>
+           </property>
+           <property name="text">
+            <string>Data protocol:</string>
+           </property>
+           <property name="alignment">
+            <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
+           </property>
+          </widget>
+         </item>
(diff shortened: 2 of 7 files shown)


src/controllers/dlgprefcontroller.cpp
```diff
@@ -1068,7 +1070,35 @@ void DlgPrefController::slotShowMapping(std::shared_ptr<LegacyControllerMapping>
...
+                    1, 1, QSizePolicy::Minimum, QSizePolicy::Expanding));
+
+            // Set all groupboxes checkable so we get the collapse/expand
+            // functionality. We want to make only the top-level groupboxes
+            // collapsible, so find direct children only. QString() is required
+            // to pass second arg Qt::FindChildOption. This is fine since our
+            // groupboxes don't have ObjectNames.
+            const QList<WCollapsibleGroupBox*> boxes =
+                    pSettingsWidget->findChildren<WCollapsibleGroupBox*>(
+                            QString(), Qt::FindDirectChildrenOnly);
```
cr7pt0gr4ph7, line 1087:
```suggestion
            // Make all top-level groupboxes checkable so we get the
            // collapse/expand functionality. Qt::FindDirectChildrenOnly
            // ensures we only iterate over the top-level groupboxes.
            const QList<WCollapsibleGroupBox*> boxes =
                    pSettingsWidget->findChildren<WCollapsibleGroupBox*>(
                            QString() /* match any ObjectName */,
                            Qt::FindDirectChildrenOnly);
```

src/widget/wcollapsiblegroupbox.cpp
```diff
@@ -0,0 +1,107 @@
...
+                m_maxH = maximumHeight();
+            }
+        }
+    }
+    return QGroupBox::event(pEvent);
+}
+
+void WCollapsibleGroupBox::slotToggled(bool checked) {
+    // React to the 'toggled' signal. By now, QGroupBox would already have
+    // enable/disable its children but that doesn't matter for us since we
```
cr7pt0gr4ph7, line 98: Small spelling nitpick.
```suggestion
    // enabled/disabled its children but that doesn't matter for us since we
```

cr7pt0gr4ph7 (approved):
LGTM :+1: Just a few small comment-related nitpicks. No major objections.

ronso0:
Thanks for your review!

I pushed the fixup and will squash before merge.

src/widget/wcollapsiblegroupbox.cpp
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    QProxyStyle::drawPrimitive(element, pOption, pPainter, pWidget);
+}
+
+WCollapsibleGroupBox::WCollapsibleGroupBox(QWidget* pParent)
+        : QGroupBox(pParent),
+          m_minH(-1),
+          m_maxH(10000) {
```
Swiftb0y, line 39: negative height? shouldn't this be `0` instead?

src/controllers/dlgprefcontroller.cpp
```diff
@@ -1068,7 +1070,35 @@ void DlgPrefController::slotShowMapping(std::shared_ptr<LegacyControllerMapping>
...
+                if (m_settingsCollapsedStates.contains(title)) {
+                    pBox->setChecked(m_settingsCollapsedStates.value(title));
+                }
+
+                connect(pBox,
+                        &WCollapsibleGroupBox::toggled,
+                        this,
+                        [this, title](bool checked) {
+                            m_settingsCollapsedStates.insert(title, checked);
+                        });
```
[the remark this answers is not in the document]
Swiftb0y: how about we just store whether the groupbox is collapsed in the groupbox itself?

src/controllers/legacycontrollersettingslayout.cpp
```diff
@@ -53,13 +54,18 @@ QWidget* LegacyControllerSettingsLayoutContainer::build(QWidget* pParent) {
 }
 
 QWidget* LegacyControllerSettingsGroup::build(QWidget* pParent) {
-    auto pContainer = make_parented<QGroupBox>(m_label, pParent);
+    auto pContainer = make_parented<WCollapsibleGroupBox>(m_label, pParent);
```
Swiftb0y, line 57: wdyt of making this configurable in the settings XML? Shouldn't be hard IMO.

Swiftb0y (commented):
couple thoughts

src/controllers/dlgprefcontroller.cpp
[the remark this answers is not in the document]
cr7pt0gr4ph7: IIUC, the MIDI learning wizard can cause the mapping to be reloaded while the preferences dialog is open, and this is an attempt to restore the expansion state in that case (because reloading also means that the mapping-defined GUI widgets are destroyed and recreated).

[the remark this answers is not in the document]
ronso0: Yes, that, too. The primary reason we need (want) to restore the state in the first place is that the settings widgets are rebuild every time the Preferences are shown. Same when pressing Cancel.

We may re-show the mapping only if it has been changed (I/O mapping, setting(s)). Eg. we could compare each setting's value with its saved value (when a setting is toggled on and of again) and somehow link that to mapping->setDirty().
But that seems a lot of work compared to the QHash implementation.

src/controllers/legacycontrollersettingslayout.cpp
ronso0: Do you mean with an extra xml node?
If yes, I 'd rather not put that work on mapping developers. I mean collapsing top-level groups is desirable, and on lower levels it only makes the GUI more complex.

But maybe we can check if the groupbox' parent is a WCollapsibleGroupBox, too, and auto-set checkable if it's not.
Will check..

ronso0: Naa, we already query the groups in DlgPrefController so that's the place to set the checkable IMO.

Swiftb0y: > Do you mean with an extra xml node?

Either that or add an attribute to the node

> If yes, I 'd rather not put that work on mapping developers. 

Right. On the flipside (and why I was suggesting to make it configurable), what if devs really want some settings to be easily visible and hide others (more advances ones for instance)?

ronso0: Ah okay, so you're proposing to allow setting the default `collapsed` state in xml (not the collapsible feature in general).

Swiftb0y: yeah, that works too.

acolombier: I like the idea of the `collapsed` default state! I was looking at adding some kin d of `depends` or `condition` to help with the large amount of settings in #13653, but this looks like anm elegant solution to address this with a minimum effort.

src/widget/wcollapsiblegroupbox.cpp
acolombier, line 39: Could we use a constant here? Or maybe some `std::numeric_limits`?

src/widget/wcollapsiblegroupbox.cpp
```diff
@@ -0,0 +1,107 @@
...
+    setFocusPolicy(Qt::NoFocus);
+}
+
+WCollapsibleGroupBox::WCollapsibleGroupBox(const QString& title, QWidget* pParent)
+        : WCollapsibleGroupBox(pParent) {
+    setTitle(title);
+}
+
+bool WCollapsibleGroupBox::event(QEvent* pEvent) {
+    if (isCheckable()) {
```
acolombier, line 55: Could you please refactor some of these nested condition to flatten it and use early return instead? It is a bit hard to follow at the moment

src/widget/wcollapsiblegroupbox.h
```diff
@@ -0,0 +1,38 @@
...
+
+#include <QGroupBox>
+#include <QProxyStyle>
+
+/// Custom style for QGroupBox that renders the 'enabled' checkbox as ArrowDown/ArrowRight
+/// icon and skips drawing the groupbox frame when the box is collapsed.
+class CollapsibleGroupBoxStyle : public QProxyStyle {
+    Q_OBJECT
+  public:
+    explicit CollapsibleGroupBoxStyle(QStyle* pStyle = 0)
```
acolombier, line 11: Should this be `nullptr` instead of `0`?

src/widget/wcollapsiblegroupbox.h
```diff
@@ -0,0 +1,38 @@
...
+
+  protected:
+    bool event(QEvent* pEvent) override;
+
+  private slots:
+    void slotToggled(bool checked);
+
+  private:
+    int m_minH;
+    int m_maxH;
```
acolombier, line 37: Does `H` stands for "height" here? How do you feel about making it explicit?

acolombier (commented):
Thank you for adding this, this is looking very good! Just some minor comments and feedbacks

cr7pt0gr4ph7:
I already have an additional use case for `WCollapsibleGroupBox` as soon as its merged: Making the `File Info` section of the track info dialog collapsible, as the latter has the tendency to exceed the screen height on wide but "height-challenged"[^1] laptop screens when combined with even slightly larger font sizes.

Also, could you make sure/test that the `WCollapsibleGroupBox` is well-behaved when using keyboard navigation?
- The header should be reachable using the keyboard, both from outside the groupbox, and from the inside (when expanded), and should have suitable visual feedback when focused via the keyboard.
- Clicking on the header should move the focus to it.
- `Spacebar` should expand/collapse the groupbox when the header is focused.
- Controls inside the box should be reachable via `Tab` when expanded, and skipped over when collapsed.
- (I'm not 100% sure what the correct behavior should be when the groupbox is programmatically collapsed while keyboard focus is in it. Should the focus stay on the previously focused but now invisible control, or should focus move to the groupbox header? Disabling the controls inside collapsed groupbox seems to actually be a good thing because it prevents making accidental edits to those controls.)

[^1]: A question for the native English speakers: What's the correct adjective for the opposite of tall when you specifically mean a small height instead of a small overall size? It's `wide` vs. `narrow`, and `tall` vs. `...`? I know there is `short`, but a `short screen` does not really sound right to me.

src/widget/wcollapsiblegroupbox.h
ronso0: Sure, no prob: m_minHeight, m_maxHeight then

src/widget/wcollapsiblegroupbox.h
ronso0: Umm, right.
Seems I mixed up the original Qt implementation with that of [ctkCollapsibleGroupBox](https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkCollapsibleGroupBox.cpp)

src/widget/wcollapsiblegroupbox.cpp
ronso0: Sure, I'll try.

src/widget/wcollapsiblegroupbox.cpp
ronso0: I tried `std::numeric_limits<int>::quiet_NaN()` earlier but it didn't work as expected, and I was lazy / it was late ; )
Will check again..

ronso0: Ah, that's only available for floating-point types.
so `std::numeric_limits<T>::min` seems to be the only applicable member?

src/widget/wcollapsiblegroupbox.cpp
cr7pt0gr4ph7:
```suggestion
    if (!isCheckable()) {
        QGroupBox::event(pEvent);
        return;
    }
```

src/widget/wcollapsiblegroupbox.h
cr7pt0gr4ph7:
```suggestion
    explicit CollapsibleGroupBoxStyle(QStyle* pStyle = nullptr)
```

src/widget/wcollapsiblegroupbox.cpp
```diff
@@ -0,0 +1,107 @@
...
+            // If this has been toggled 'collapsed' before it has been shown we
+            // collapsed it now that we have a valid min height.
+            if (isChecked()) {
+                m_maxH = maximumHeight();
+            } else {
+                setMaximumHeight(m_minH);
+            }
+        } else if (pEvent->type() == QEvent::LayoutRequest ||
+                pEvent->type() == QEvent::ContentsRectChange ||
+                pEvent->type() == QEvent::Resize) {
```
cr7pt0gr4ph7, line 81: It might make sense to replace this `if/else` blocks with a `switch`.

src/widget/wcollapsiblegroupbox.cpp
ronso0: Yes, that's the only one.

src/widget/wcollapsiblegroupbox.cpp
ronso0: Hmm, while it works with `std::numeric_limits<int>::min()`, I'd rather stick with `-1` here:
just in case one of the `style()->subControlRect()` queries returns a QRect with an invalid height (-1) we have the chance to set it on the next call.

Wdyt?

ronso0:
> * The header should be reachable using the keyboard, both from outside the groupbox, and from the inside (when expanded), and should have suitable visual feedback when focused via the keyboard.
> * Clicking on the header should move the focus to it.
> * Spacebar should expand/collapse the groupbox when the header is focused.
> * Controls inside the box should be reachable via Tab when expanded, and skipped over when collapsed.

:heavy_check_mark: since we just hook up to the 'checked' state this is all handled correctly by QGroupBox
(now that I removed the wrong `setFocusPolicy(Qt::NoFocus);` ; )

> (I'm not 100% sure what the correct behavior should be when the groupbox is programmatically collapsed while keyboard focus is in it. Should the focus stay on the previously focused but now invisible control, or should focus move to the groupbox header? Disabling the controls inside collapsed groupbox seems to actually be a good thing because it prevents making accidental edits to those controls.)

The same here: QGroupBox handles this. The original behavior of setChecked() is to enable/disable all children and we simply expect that to work.

src/widget/wcollapsiblegroupbox.cpp
Swiftb0y: imo fine as long as you refactor it into a `static constexpr` constant. no magic numbers and all that u know.

ronso0:
Btw I fixed soe other keyboard focus issue in the controller page (settings checkbox, more tabstops, don't focus hidden tabs ..)

src/controllers/legacycontrollersettingslayout.cpp
ronso0: Thanks for the feedback!

Though I'd love to merge the basic implementation first.

src/controllers/dlgprefcontrollerdlg.ui
```diff
@@ -100,114 +100,422 @@
...
-        <property name="alignment">
-         <set>Qt::AlignRight|Qt::AlignTrailing|Qt::AlignVCenter</set>
-        </property>
-       </widget>
-      </item>
-      <item row="1" column="1">
-       <widget class="QLabel" name="labelDataHandlingProtocolValue">
+     <layout class="QVBoxLayout" name="deviceInfoGroupLayout">
+      <item>
+       <widget class="WCollapsibleGroupBox" name="groupBoxDeviceInfo">
```
JoergAtGithub, line 114: Please ensure that groupBoxDeviceInfo and groupBoxMappingInfoare always expanded, if no mapping is loaded to a device, as these boxes contain the information needed by the end user to select the correct mapping for the right device.

JoergAtGithub (changes requested):

src/controllers/dlgprefcontrollerdlg.ui
ronso0: [By default these groups are expanded and remain so until the user explicitly collapses them.](https://www.deepl.com/translator)
And the most important info is the device name which is the big label in the top left corner.
What other info do you refer to?

JoergAtGithub: The text in the big label is not unique, this is why Device Info was implemented for 2.6. On my system I have 3 devices with exactly the same string:
grafik
This is normal because this string depends on assumptions that are only sometimes true. This was addressed by adding the device info box. If this information is hidden, the user has no chance to distinguish the devices.
There is no use case for collapsing these 2 boxes of valuable information needed for the mapping selection, until a mapping is loaded. This is because the tabs below are hidden at this time and enough space is available to show them.

src/widget/wcollapsiblegroupbox.cpp
```diff
@@ -35,15 +35,14 @@ void CollapsibleGroupBoxStyle::drawPrimitive(QStyle::PrimitiveElement element,
 
 WCollapsibleGroupBox::WCollapsibleGroupBox(QWidget* pParent)
         : QGroupBox(pParent),
-          m_minH(-1),
-          m_maxH(10000) {
+          m_minHeight(-1),
+          m_maxHeight(QWIDGETSIZE_MAX) {
```
acolombier, line 39: Nice :100: ! I didn't know about this macro

acolombier (approved):
Code LGTM! Will take that for a spin in a bit

acolombier (changes requested):
There is a weird glitch when expending a collapsed group. tested on Fedora 41 (Wayland)

https://github.com/user-attachments/assets/6f29073b-6331-4ac8-b837-e83b1383ff81

ronso0:
I'm seeing this, too, if I look for it.
Will check..

ronso0:
Hmm, I traced the layout / resize events and it seems there is no easy fix:
when a group is expanded (and also sometimes when it's collapsed??) the parent layout is resized, too, but apparently doesn't keep up with the resize events of its children. The groups above are shrunk and regrown temporarily in 1-2 steps until all is setlled and the desired layout is established.

Maybe sequential hiding/showing of the group children fixes it but I fear this also adds more complexity / more stuff to care about. Bit like it's done in [ctkCollapsibleGroupBox](https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkCollapsibleGroupBox.cpp) which also resizes via `setMaximumHeight()`, but I didn't build with that (Qt-derived?) framework.

ronso0:
> Maybe sequential hiding/showing of the group children fixes it

No, it doesn't.
I suspect it's an resize issue with the parent layout, or someQSizePolicy glitch QLayout/QTabWidget.
> the parent layout is resized, too, but apparently doesn't keep up with the resize events of its children.

The weird thing is: it also happens when a group is collapsed.. so in that case the layout seems to shrink before the box is actually collapsed, squeezing the groups above.

These are extemely hard to debug/trace and I wasted countless hours already fixing (or trying to) some Mixxx dialogs.
I'll give it another go (looking at the settings layout) but tbh I don't expect we can do much about it :man_shrugging:

ronso0:
Yeah, I'll stop working on this (fix).
It's the order of the events (resize, layou update) and the glitch appears randomly (but is easier to reproduce if the (actually static) group above holds QCheckBoxes).

Btw when collapsing a group, the parent QScrollArea might toggle the vertical scrollbars which is another resize reason.

I think we either have to live with this Qt glitch, or delay this PR until someone finds a fix.

acolombier (approved):

acolombier:
Tbh, I'm happy living with this bug. It's a small discomfort, but doesn't fundamentally prevent or break anything. Thanks for trying anyway!

ronso0:
Btw I think it's not related to the settings layout, I also see the glitch in when I use WCollapsibleGroupBox in Deck preferences **if the expanded groups don't fit in the page, ie. scrollbars are toggled**.

FWIW I think we should fix the abnormally large layout padding in the controller settings:
Library
Screenshot_2025-02-18_16-23-16
Controller
Screenshot_2025-02-18_16-23-23

ronso0:
Alright then, thanks for your review!
Squashed and main merged, ready to roll.

@JoergAtGithub Could I dispel your concerns re the Device/Mapping groups?

acolombier:
Friendly ping @JoergAtGithub in case you missed that one :)

src/controllers/dlgprefcontrollerdlg.ui
JoergAtGithub: The use case I'm thinking about is the beginner or first time user who needs this information to distiguish devices with same or similar names. Once it's collapsed, he might have not recognise that this information is hidden here.
the collapsing makes only sense if there is content to show instead. It makes the life harder for the beginner user with no benefit for any other use case.
But as changing the default can be done at any time later let's merge this know and look later how to restore the current user experience afterwards.
An alternative to changing the collapse state itself might be a second state, which disables the collapse mode and let it behave like a normal GroubBox.

acolombier:
@ronso0 would you mind updating [the documentation](https://github.com/mixxxdj/mixxx/wiki/MIDI-Scripting#settings-api) with that new feature?

ronso0:
Umm, for users this is only a 'tiny' new front end feature, literally just the expand toggles.
So I'm not sure what to add to the doc. What were you thinking of?

PR #13979, opened 2024-12-04
Controller settings: have clickable checkbox label (like QCheckBox)

ronso0:
From the preferences and basically any QCheckBox in Mixxx' dialogs I'm used to being able to click on the label to toggle the checkbox (big label area, small checkbox).

However, in the controller settings we have separate QCheckBoxes (empty label) and QLabels because QCheckBox doesn't support html/richtext for formatting its label, and label click doesn't toggle the box.

This PR fixes this.
Wrt implementation we could as well use a custom QCheckBox and override its `mousePressEvent()`, the solution with eventFilter() was just much easier/quicker.

PR #14327, opened 2025-02-13
Fix 14326

### LINKED ISSUE #14326: many warnings [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
### Bug Description

On startup I get the following. Is anybody else seeing this?

warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.
warning [Main] QObject::installEventFilter(): Cannot filter events for objects in a different thread.

### Version

2.5

### OS

macOS

m0dB:
Fixes #14326

installEventFilter was using an object created in a different thread than the main thread. this is illegal. Fixed by using a helper class with the eventFilter method, which is instantiated in the main thread.

### DIFF
--- a/src/controllers/legacycontrollersettings.cpp
+++ b/src/controllers/legacycontrollersettings.cpp
@@ -155,7 +155,8 @@ QWidget* LegacyControllerBooleanSetting::buildInputWidget(QWidget* pParent) {
     // We want to format the checkbox label with html styles. This is not possible
     // so we attach a separate QLabel and, in order to get a clickable label like
     // with QCheckBox, we associate the label with the checkbox (buddy).
-    // When the label is clicked we toggle the checkbox in eventFilter().
+    // When the label is clicked we toggle the checkbox in the function
+    // eventFilter() of helper class ToggleCheckboxEventFilter.
     auto pLabelWidget = make_parented<QLabel>(pWidget);
     pLabelWidget->setSizePolicy(QSizePolicy::Preferred, QSizePolicy::Maximum);
     pLabelWidget->setText(label());
@@ -187,6 +188,11 @@ bool LegacyControllerBooleanSetting::match(const QDomElement& element) {
 
 bool LegacyControllerBooleanSetting::ToggleCheckboxEventFilter::eventFilter(
         QObject* pObj, QEvent* pEvent) {
+    // eventFilter was a method of LegacyControllerBooleanSetting, but this
+    // doesn't work, as LegacyControllerBooleanSetting is created from a
+    // different thread, and Qt cannot filter events for objects in a
+    // different thread.
+
     QLabel* pLabel = qobject_cast<QLabel*>(pObj);
     if (pLabel && pEvent->type() == QEvent::MouseButtonPress) {
         QCheckBox* pCheckBox = qobject_cast<QCheckBox*>(pLabel->buddy());
--- a/src/controllers/legacycontrollersettings.cpp
+++ b/src/controllers/legacycontrollersettings.cpp
@@ -115,7 +115,6 @@ QWidget* AbstractLegacyControllerSetting::buildWidget(QWidget* pParent,
 LegacyControllerBooleanSetting::LegacyControllerBooleanSetting(
         const QDomElement& element)
         : AbstractLegacyControllerSetting(element) {
-    m_pToggleCheckboxEventFilter = nullptr;
     m_defaultValue = parseValue(element.attribute("default"));
     m_savedValue = m_defaultValue;
     m_editedValue = m_defaultValue;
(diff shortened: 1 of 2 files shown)


daschuer:
Is there a need to allocate the event filter separately on the heap. I think we can just dispose the whole class and override the eventFilter like described here: 
https://doc.qt.io/qt-6/qobject.html#eventFilter

Else, please use a parented_ptr

daschuer (changes requested):
Ups, I missed the code and deleted the comment. 

Please use a parented_ptr. Than it is ready to go..

Swiftb0y:
does the same issue apply to the other setting types?

m0dB:
> does the same issue apply to the other setting types?

I don’t know. I am not familiar at all with this part of the code base. I just spotted these warnings and decided to address them because they looked problematic. I have not seen the warning for other use of installEventFilter but of course there might be code paths that are not commonly executed, just like this one.

We could run with code coverage, identify all installEventFilter calls that are not executed and see what needs to be done so they are executed. Not a task I am interested in. 

Another option is to add a debug assert next to each installEventFilter call to check if the object was created in the main thread, so at least this warning doesn’t go unnoticed. Or intercept the warning and debug assert there.

m0dB:
> Ups, I missed the code and deleted the comment.
> 
> Please use a parented_ptr. Than it is ready to go..

Great. I will do this tonight. I was not familiar to parented_ptr, very good to know about this!

Swiftb0y:
> I don’t know. I am not familiar at all with this part of the code base. I just spotted these warnings and decided to address them because they looked problematic. I have not seen the warning for other use of installEventFilter but of course there might be code paths that are not commonly executed, just like this one.

Okay, I just double checked and it seems its only there to implement the convience feature of toggling the checkbox by clicking on the label. That behavior is does not apply / is not implemented for any other data type. So its only an issue here.

Swiftb0y (approved):
LGTM apart from daniels review. Thank you.

daschuer (approved):
LGTM, Thank you

m0dB:
Hm, the change to the parented_ptr actually causes a crash...

m0dB:
Right, I set the wrong parent... I did a quick fixed and merged it https://github.com/mixxxdj/mixxx/pull/14333

m0dB:
And now correctly https://github.com/mixxxdj/mixxx/pull/14338

m0dB:
BTW, I would actually be in favour of removing the entire "feature" altogether. Sure, it is nice you can click on the label instead of having to click exactly at in the checkbox, but is it really necessary?

Swiftb0y:
not necessary, but I personally do find it very convenient especially on small screens. I'm not a fan of removing it unless there is very good reason to do so.

ronso0:
Hmm, so with this and #14338 I can set the Denon MC7000 mapping on my Vir-MIDI.

But if I start with that mapping configured I get
```
warning [Main] QObject: Cannot create children for a parent that is in a different thread.
(Parent is AbstractLegacyControllerSetting(0x7ff6f0046b00), parent's thread is QThread(0x7ff7080089a0), current thread is QThread(0x559d62d939d0)
DEBUG ASSERT: "!m_ptr || static_cast<const QObject*>(m_ptr)->parent()" in function parented_ptr<T>::parented_ptr(T*) [with T = LegacyControllerBooleanSetting::ToggleCheckboxEventFilter] at ./src/util/parented_ptr.h:32
```

ronso0:
> BTW, I would actually be in favour of removing the entire "feature" altogether. Sure, it is nice you can click on the label instead of having to click exactly at in the checkbox, but is it really necessary?

The checkbox + label is a hack in the first place, and I added the "click feature" to bring it (back) ne par with the default QCheckBox usability. Please don't remove it.

In #13979 I proposed to use a custom QCheckBox and override the mousePressEvent().
I just did it :tm: now #14355

PR #4614, opened 2022-01-09
Reloop Digital Jockey 2 IE controller mapping

dj-ak:
Controller mapping for Reloop Digital Jockey 2 Interface Edition.

In Mixxx, there is already a mapping for a "Controller Edition" version of Reloop Digital Jockey 2, although it is not usable. Compared to the "Controller Edition", the "Interface Edition" is additionally equipped with an USB audio interface and touch-sensitive jog wheels.

Maybe the existing CE mapping was working with older versions of Mixxx, however I had to write my own mapping from scratch for Mixxx v2.1 as the existing mapping lacks lot of functionality and does not work properly anyway. Especially the jog wheels are not working properly and the user cannot do e.g. pitch bending. My mapping solves these issues. The commits are old but I rebased them lately onto 2.3 branch, built mixxx from source and tested under Kubuntu 20.04.

More information in the forum thread:
https://mixxx.discourse.group/t/reloop-digital-jockey-2-mapping-by-dj-ak/23971

PR #14328, opened 2025-02-13
fix: skip message send if sysex code is undefined

acolombier:
Fixes test regression introduced in #4614

CC @daschuer

### DIFF
--- a/res/controllers/Reloop-Digital-Jockey-2-IE.scripts.js
+++ b/res/controllers/Reloop-Digital-Jockey-2-IE.scripts.js
@@ -446,7 +446,7 @@ RDJ2.JogModeSelector.prototype = new components.Component({
         var scratchValue = this.jogMode === RDJ2.JOGMODES.vinyl ? RDJ2.MIDI_ON : RDJ2.MIDI_OFF;
         var fxDryWetValue = this.jogMode === RDJ2.JOGMODES.fxdrywet ? RDJ2.MIDI_ON : RDJ2.MIDI_OFF;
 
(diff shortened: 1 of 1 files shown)


Swiftb0y:
Auditing the surrounding code, I don't understand how any of these properties could be `undefined`. So I'm unsure if this is really just hiding symptom of a bigger issue.

git-developer:
Looks to me as if `this.searchMidiCtrl` is undefined, probably because the deck number is required but not available; cause might be that the object is created [as part of a prototype declaration](https://github.com/mixxxdj/mixxx/blob/8022489ed07512a4b5894dc455a12fa4d3befd5e/res/controllers/Reloop-Digital-Jockey-2-IE.scripts.js#L142).

acolombier:
I'm only making this fix so the test passes. I don't have the device so I don't want to make major changes blindly, and cannot run proper debug.

Alternative is to revert #4614, but the consensus we had about mapping PRs during the last monthly catchup was to merge mapping changes as is, as long as they roughly match our standards.

The TL;DR was; it's better to have something potentially broken, than nothing

Swiftb0y:
yeah agreed. Still making @dj-ak aware. maybe they can look into it.

git-developer:
I agree on the quick fix.

On the long run I'd prefer to avoid calling functions from [`unshift()`](https://github.com/mixxxdj/mixxx/blob/8022489ed07512a4b5894dc455a12fa4d3befd5e/res/controllers/Reloop-Digital-Jockey-2-IE.scripts.js#L532) while a prototype is declared.

Issue #14334, opened 2025-02-14
Allow feedback on every release of a `powerWindow` button


git-developer:
### Feature Description

This feature request proposes a solution for a problem that occurs on some MIDI controllers: the LED of a button on the controller does not resemble the state of the corresponding `Button` component when the `Button`'s type is `powerWindow` and the hardware button is released.

### Problem
Buttons on MIDI controllers often show the behavior of a switch: they send `0x7F` on button press and `0x00` on button release. There might also be an LED that shows the current state. On some controllers, _the LED is automatically unlit when the button is released_. This is fine for `push` and `toggle` buttons, but causes an inconsistent LED state when the corresponding `Button` component im Mixxx has type `powerWindow`. The reason is that an incoming release message (`0x00`) does not change the `Button`'s state in this case. In the end, the `Button` component is enabled but the controller's LED shows disabled state.

At least the Behringer BCR2000 controller is affected of this problem. It is not possible to change the controller's behavior by configuration.

### Proposal
The problem can be solved by sending the current state of a `Button` with type `powerWindow` to the controller on release even if the `Button` state did not change.

Swiftb0y:
So do I understand this correctly: The controller is automatically turning off the LED when releasing the button (behavior backed into firmware)?

Swiftb0y:
That would also cause issues in `toggle` mode, not only powerWindow, so I'm not sure I understand the problem correctly.

git-developer:
> So do I understand this correctly: The controller is automatically turning off the LED when releasing the button (behavior backed into firmware)?

Yes, exactly.

> That would also cause issues in toggle mode

I didn't experience any issue for `toggle`. Controller state and software state are in sync. On press, the controller enables the LED, and Mixxx sends out a feedback of `on` in addition. On release, the controller disables the LED and Mixxx sends out a feedback of `off` in addition.

git-developer:
I think I might have mixed up the behavior of `push` and `toggle`.

Toggle behavior can be configured in the firmware, see [manual](https://manual.mixxx.org/2.5/en/hardware/controllers/behringer_bcr2000.html#button-behaviour).

Swiftb0y:
Yes. It seems you have. The configuration via the firmware makes sense.

git-developer:
> That would also cause issues in toggle mode

I see. When the controller wouldn't be configurable, this would be an issue. Do you think I should I add the optional feedback for `toggle`, too? For the sake of consistency?

[...]

PR #14335, opened 2025-02-14
Allow feedback on every release of a `powerWindow` button

git-developer:
This PR addresses #14334 by adding an option to send the current state of a `Button` with type `powerWindow` to the controller on release even if the `Button` state did not change. The new behavior is hidden behind a new feature flag to keep backwards compatibility. Only when the options of a `Button` contain the new property `triggerOnRelease`, the current state is sent to the controller instead of doing nothing. Only buttons of type `powerWindow` are affected.

The implementation has been tested successfully on a Behringer BCR2000 controller.

The `Behringer-Extension-scripts.js` have been extended to apply the new behavior to effect units, quick effect units and equalizers if their configuration contains the newly introduced option `feedbackOnRelease`. These component containers contain buttons of type `powerWindow`.

### DIFF
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -176,6 +176,7 @@
         // in any Buttons that act differently with short and long presses
         // to keep the timeouts uniform.
         longPressTimeout: 275,
+        triggerOnRelease: false,
         isPress: function(channel, control, value, _status) {
             return value > 0;
         },
@@ -197,6 +198,8 @@
                 } else {
                     if (this.isLongPressed) {
                         this.inToggle();
+                    } else if (this.triggerOnRelease === true) {
+                        this.trigger();
                     }
                     if (this.longPressTimer !== 0) {
                         engine.stopTimer(this.longPressTimer);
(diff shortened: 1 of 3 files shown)


res/controllers/midi-components-0.0.js
```diff
@@ -197,6 +198,8 @@
                 } else {
                     if (this.isLongPressed) {
                         this.inToggle();
+                    } else if (this.triggerOnRelease === true) {
```
acolombier, line 201: @Swiftb0y do you think truthy values assertion would be sufficient here? Happy to keep it if we think this is necessary, it's just that in general, checking if a boolean is true in a condition makes me itchy :rofl:

cr7pt0gr4ph7 (approved):
LGTM. Except for the `<boolean value> == true`, everything seems fine.

res/controllers/midi-components-0.0.js
Swiftb0y: nah, just this is fine:
```suggestion
                    } else if (this.triggerOnRelease) {
```

git-developer: Done.

Swiftb0y (approved):
LGTM. Thank you.

Issue #13827, opened 2024-11-02
Add links to development policies & guidelines to CONTRIBUTING.md


cr7pt0gr4ph7:
I have read [CONTRIBUTING.md](https://github.com/mixxxdj/mixxx/blob/main/CONTRIBUTING.md) when I started developing with Mixxx, but have since learned through various (albeit very friendly :)) comments in issues and reviews that there are a few important documents, like:
- [Minimum requirements policy](https://github.com/mixxxdj/mixxx/wiki/Minimum-requirements-policy)
- [Coding guidelines](https://github.com/mixxxdj/mixxx/wiki/Coding-Guidelines)

that are not linked there, but which new contributors-to-be should probably be made aware of as soon as possible.

@JoergAtGithub @daschuer @ronso0 @Holzhaus @Swiftb0y @fwcd @acolombier @uklotzde (basically everyone who was active in the last few months)
Are there any other important documents or pieces of information that should be linked there?

Thanks in advance for any helpful pointers!

daschuer:
Maybe: 
https://github.com/mixxxdj/mixxx/wiki/Internationalization
and 
https://github.com/mixxxdj/mixxx/wiki/Release-Checklist-2.4.0

cr7pt0gr4ph7:
(Just as an aside: Would it make sense to have a dedicated GitHub label for this and similar issues related to development & contributing, e.g. `developer-experience` (Description: `Issues, bugs and PRs related to the development process, development environment & developer documentation of Mixxx itself.`)?)

daschuer:
Done

PR #14343, opened 2025-02-14
Labeler: Add `developer experience` issue label

cr7pt0gr4ph7:
The `developer experience` issue label created by request in [#13827](https://github.com/mixxxdj/mixxx/issues/13827#issuecomment-2453422945) should also be handled by the automatic issue labeler.

---

I also have two related requests:
- [x] Can we rename **`developer-experience`** to **`developer experience`** to match the naming convention of the other issue labels?
- [x] Can someone fix the currently truncated description of the `developer experience` label to read: `Issues, bugs and PRs related to the development process, development environment & developer docs.`?

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -32,6 +32,14 @@ controller mappings:
       - any-glob-to-any-file:
           - res/controllers/**
 
+developer experience:
+  - changed-files:
+      - any-glob-to-any-file:
+          - CONTRIBUTING.md
+          - README.md
+          - .github/labeler.yml
+          - .pre-commit-config.yaml
+
 analyzer:
   - changed-files:
       - any-glob-to-any-file:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -35,10 +35,13 @@ controller mappings:
 developer experience:
   - changed-files:
       - any-glob-to-any-file:
+          - .github/ISSUE_TEMPLATE/**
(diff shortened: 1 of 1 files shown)


Swiftb0y (approved):

.github/labeler.yml
```diff
@@ -32,6 +32,17 @@ controller mappings:
       - any-glob-to-any-file:
           - res/controllers/**
 
+developer experience:
+  - changed-files:
+      - any-glob-to-any-file:
+          - .github/ISSUE_TEMPLATE/**
+          - .github/labeler.yml
+          - tools/**
```
acolombier, line 40: Not sure if `tools` is relevant for this tags - there is a lot of tools in here which aren't related to developer experience IMO.
I would also argue that labeler is not related to developer experience either

cr7pt0gr4ph7: I'd say that labeler is related to `development processes`, and thus fits the `developer experience` tag.

re: `tools/`: AFAICT the labeler is only applied initially to new pull requests, and the labels added using these heuristics can be removed manually if desired, so a bit of over- or underlabeling is better than none at all.

But I'm also fine with removing both `.github/labeler.yml` and `tools/`, every way of labeling will be some kind of "best effort" anyway.

acolombier: Let's keep it like this for now, we can always remove it if we get too many false positive

PR #14345, opened 2025-02-14
PitchBendControl: Extract PitchBendControl from RateControl.

cr7pt0gr4ph7:
As part of an effort to make maintenance of the `EngineControl` code easier, this PR splits off the **pitch bending logic & controls** into its own class. This was easily possible because the pitch bending code did not have any dependencies on the other code inside `RateControl` anyway.

Apart from the changed classname and the slightly changed signature of `processTempRate`, the `PitchBendControl` code is a one-to-one copy of the corresponding existing code in `RateControl`.

### DIFF
--- a/src/engine/controls/pitchbendcontrol.cpp
+++ b/src/engine/controls/pitchbendcontrol.cpp
@@ -1,0 +1,199 @@
+#include "engine/controls/pitchbendcontrol.h"
+
+#include <QtDebug>
+
+#include "control/controlobject.h"
+#include "control/controlpushbutton.h"
+#include "engine/controls/enginecontrol.h"
+#include "moc_pitchbendcontrol.cpp"
+
+namespace {
+constexpr int kRateSensitivityMin = 100;
+constexpr int kRateSensitivityMax = 2500;
+} // namespace
+
+// Static default values for rate buttons (percents)
+ControlValueAtomic<double> PitchBendControl::m_dTemporaryRateChangeCoarse;
+ControlValueAtomic<double> PitchBendControl::m_dTemporaryRateChangeFine;
+int PitchBendControl::m_iRateRampSensitivity;
+PitchBendControl::RampMode PitchBendControl::m_eRateRampMode;
+
+PitchBendControl::PitchBendControl(const QString& group,
+        UserSettingsPointer pConfig)
+        : EngineControl(group, pConfig),
+          m_bTempStarted(false),
+          m_tempRateRatio(0.0),
+          m_dRateTempRampChange(0.0) {
+    // We need the sample rate so we can guesstimate something close
+    // what latency is.
+    m_pSampleRate = ControlObject::getControl(
+            ConfigKey(QStringLiteral("[App]"), QStringLiteral("samplerate")));
+
+    // Temporary rate-change buttons
+    m_pButtonRateTempDown =
+            new ControlPushButton(ConfigKey(group, "rate_temp_down"));
+    m_pButtonRateTempDownSmall =
+            new ControlPushButton(ConfigKey(group, "rate_temp_down_small"));
+    m_pButtonRateTempUp =
+            new ControlPushButton(ConfigKey(group, "rate_temp_up"));
+    m_pButtonRateTempUpSmall =
+            new ControlPushButton(ConfigKey(group, "rate_temp_up_small"));
+
+    //     // Update Internal Settings
+    //     // Set Pitchbend Mode
+    //     m_eRateRampMode = static_cast<RampMode>(
+    //         getConfig()->getValue(ConfigKey("[Controls]","RateRamp"),
+    //                               static_cast<int>(RampMode::Stepping)));
+
+    //     // Set the Sensitivity
+    //     m_iRateRampSensitivity =
+    //             getConfig()->getValueString(ConfigKey("[Controls]","RateRampSensitivity")).toInt();
+}
+
+PitchBendControl::~PitchBendControl() {
+    delete m_pButtonRateTempDown;
+    delete m_pButtonRateTempDownSmall;
+    delete m_pButtonRateTempUp;
+    delete m_pButtonRateTempUpSmall;
+}
+
+// static
+void PitchBendControl::setRateRampMode(RampMode mode) {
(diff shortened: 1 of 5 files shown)


ronso0:
CI fails to build.

src/engine/controls/pitchbendcontrol.cpp
```diff
@@ -0,0 +1,199 @@
...
+    //     m_iRateRampSensitivity =
+    //             getConfig()->getValueString(ConfigKey("[Controls]","RateRampSensitivity")).toInt();
+}
+
+PitchBendControl::~PitchBendControl() {
+    delete m_pButtonRateTempDown;
+    delete m_pButtonRateTempDownSmall;
+    delete m_pButtonRateTempUp;
+    delete m_pButtonRateTempUpSmall;
+}
```
Swiftb0y, line 58: its not easy to see what code is just moved out from a class and what code is newly introduced. As such I usually treat code like this as new and would ask for a thorough clean up at the same time (akin to the "boyscout rule"). Is that okay for you?

ronso0: I guess you mainly refer to parented_ptr for all the controls/proxies?

cr7pt0gr4ph7: "Leave it better than you found it" is a good motto to live by. I'd just like to avoid making this initial changeset bigger than necessary to avoid

My process is usually to extract the code, get it to compile then run `git diff <oldfile> <newfile>` to see if I only changed things I should. Sometimes (as in this case) this is then accompanied by a reordering of the source code.

The output of `git diff upstream/main:src/engine/controls/ratecontrol.cpp ready-for-merge/engine-controls/pitch-bend-control:src/engine/controls/pitchbendcontrol.cpp --function-context` can be [found here](https://gist.github.com/cr7pt0gr4ph7/8265c87cc435bdf42dfe673b37892bdc).

cr7pt0gr4ph7: > I guess you mainly refer to parented_ptr for all the controls/proxies?

The current code of the different `EngineControl`s (and also other parts of Mixx) uses a mixture of raw pointers, `std::unique_ptr` and `parented_ptr` when referring to `ControlObject` and `ControlProxy`. At least most files are self-consistent in which variant they use.

The _only_ combination that does _not_ seem to occur is using a `parented_ptr<ControlObject>`, though I don't see a reason why it shouldn't be used. As `ControlObject` **is** a `QObject`, it might even be considered to be the most appropriate.

ronso0: here's some losely related Zuip convo (about ControlProxy pointer types):
https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/ControlProxy.20preferred.20pointer.20type

cr7pt0gr4ph7: Thank you for the link, that is very much related. To summarize my understanding:

While we care about the `QObject` tree and thus proper parent-child relationships for widgets due to how it interacts with event handling, we don't need it for `ControlObject` and `ControlProxy`.

Not using `parented_ptr` when not necessary avoids both some threading-related validation code in the `QObject` constructor as well as the initialization of the `QObject::children()` array, so it is likely faster to use `std::unique_ptr`. This is relevant because of the _huge_ number of of controls inside of Mixxx.

There does not appear to be a downside to using `std::unique_ptr`, as the `deleteLater` processing does not depend on the `QObject` tree per se, but only on whether `QObject::thread()` refers to the thread that actually owns the object.

cr7pt0gr4ph7:
Sorry for the build failures, I'm currently fighting with my local build setup and with porting over my old changesets from ~4 months ago to the current state of `main`.

src/engine/controls/pitchbendcontrol.cpp
cr7pt0gr4ph7: Wdyt about pulling all of the following forward declarations into their own header file `src/control/control_decl.h`?

```
class ControlAudioTaperPot;
class ControlEffectKnob;
class ControlEncoder;
class ControlIndicator;
class ControlIndicatorTimer;
class ControlLinPotmeter;
class ControlLogPotmeter;
class ControlObject;
class ControlPushButton;
class ControlTTRotary;

class ControlProxy
```

Swiftb0y: Better to only forward declare what you're actually using IMO. Moreover, if you don't need polymorphism or you don't specifically care about decoupling dependencies, you can always just use the class directly as a member (no need for heap allocation at all).

cr7pt0gr4ph7: > Better to only forward declare what you're actually using IMO.

The advantage of a dedicated `control_decl.h` would be a better indication of where the corresponding full declarations for a given forward declaration are, but I see your point.

> Moreover, if you don't need polymorphism or you don't specifically care about decoupling dependencies, you can always just use the class directly as a member (no need for heap allocation at all).

True. But given that, is there a reason why we're almost exlusively using (both smart and raw) pointers for owning `ControlObject` instances?

Swiftb0y: not really. I mostly suspect that most of it is actually very old code and the mixxx devs back then just `new` because they thought C++ with Qt was Java. Then when it came around to refactoring, changing from a pointer type a value to type was too tedious so they just wrapped it in a `unique_ptr` to maintain the syntax.

cr7pt0gr4ph7: I'm happy to replace these occurences with values instead of pointers, but this sounds like something that we should maybe do a benchmark on, just in case.

Swiftb0y: I don't really see how though. At least in terms of a microbenchmark. Though if we decrease the number of heap allocations, its IMO quite likely that we decrease memory usage in general (remember every pointer is 8 bytes of overhead) and potentially also startup/shutdown times. Any malloc/free is an improvement if there are enough of them. 

Nevertheless, this isn't really about performance but rather choosing the right tool for the job.

PR #14346, opened 2025-02-15
open Sound preferences with sprecific I/O tab selected

ronso0:
Just a little helper that saves one click when being redirected from skin to Sound preferences after agreeing to set up a Mic, Aux or VC / Passthrough device.

### DIFF
--- a/src/preferences/constants.h
+++ b/src/preferences/constants.h
@@ -39,6 +39,12 @@ enum class MultiSamplingMode {
 };
 Q_ENUM_NS(MultiSamplingMode);
 
+enum class SoundHardwareTab {
+    Output,
+    Input
+};
+Q_ENUM_NS(SoundHardwareTab);
+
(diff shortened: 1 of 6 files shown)


src/preferences/constants.h
```diff
@@ -39,6 +39,13 @@ enum class MultiSamplingMode {
 };
 Q_ENUM_NS(MultiSamplingMode);
 
+enum class SoundHardwareTab {
+    Output,
+    Input,
+    Invalid
```
Swiftb0y, line 45: do we really need `Invalid`?

ronso0: I added it to have a default `tab` argument for `showSoundHardwarePage`.

Thinking about it, that could also be std::optional.
Fixed.

daschuer (approved):
LGTM, Thank you.

Issue #11473, opened 2023-04-11
Problem with midi-components-0.0.js


elgithubo:
### Bug Description

The following ``git diff`` is required for the Behringer DDM4000 in order to work and might be required for all midi devices on FreeBSD 13.2-RC3.
```
diff --git a/res/controllers/midi-components-0.0.js b/res/controllers/midi-components-0.0.js
index 6bed8a8bf2..8631be97a9 100644
--- a/res/controllers/midi-components-0.0.js
+++ b/res/controllers/midi-components-0.0.js
@@ -123,7 +123,7 @@
                 undefined !== this.outKey &&
                 undefined !== this.output &&
                 typeof this.output === "function") {
-                this.connections[0] = engine.makeConnection(this.group, this.outKey, this.output.bind(this));
+                //this.connections[0] = engine.makeConnection(this.group, this.outKey, this.output.bind(this));
             }
         },
         disconnect: function() {
```

Swiftb0y:
As I outlined in your previous PR, this is vital line for componentsjs to work correctly. Please post the actual error message your receive when you use your DDM4000 so we can work on fixing the problem. Repeatedly posting a patch that will not work won't get this fixed. Thank you!

elgithubo:
Sorry I was just trying to refer to the topic, I understand that coding is required,
When I start mixxx with the original coding I receive the following error:

```
Uncaught exception: file:///usr/local/share/mixxx/controllers/midi-components-0.0.js:126: Error: script tried to connect to ControlObject ([Channel1], ) which is non-existent.
Backtrace: connect@file:///usr/local/share/mixxx/controllers/midi-components-0.0.js:126
Component@file:///usr/local/share/mixxx/controllers/midi-components-0.0.js:50
Button@file:///usr/local/share/mixxx/controllers/midi-components-0.0.js:163
createComponent@file:///usr/local/share/mixxx/controllers/Behringer-Extension-scripts.js:1568
@file:///usr/local/share/mixxx/controllers/Behringer-Extension-scripts.js:1394
createDeck@file:///usr/local/share/mixxx/controllers/Behringer-Extension-scripts.js:1391
@file:///usr/local/share/mixxx/controllers/Behringer-Extension-scripts.js:1366
notify@file:///usr/local/share/mixxx/controllers/Behringer-Extension-scripts.js:293
schedule@file:///usr/local/share/mixxx/controllers/Behringer-Extension-scripts.js:287
[...]

Issue #13826, opened 2024-11-02
Drop Qt5 support completely for 2.5 & main


cr7pt0gr4ph7:
As it [came up](https://github.com/mixxxdj/mixxx/pull/13807#issuecomment-2453000013) in a PR discussion, it seems that the issue of whether to support both Qt5+6, or only Qt6, has not been entirely decided yet.

- While versions up to and including `2.4` are built with Qt5,
- the upcoming `2.5` version as well as the current `main` are built with Qt6 instead.

To clarify, this proposal only applies to `main`, and not to the older branches.

In my opinion, there is no sense in keeping the remnants of Qt5 support around if we are not actively making sure that Mixxx can actually still be built (and run) with Qt5 using appropriate CI jobs. Otherwise, the code base will silently develop into a state where it either will not build on Qt5 anymore, or (IMO worse), will have buggy code that is only active when building & running on Qt6.

So, with the aim of cleaner code, less bugs and improved readability in mind, I propose we decide for one of three options:
1. **Keep Qt5 support around.** Some work would still need to be done:
  - Clarify the coding guidelines to reflect that
  - Make sure that PRs are reviewed for Qt5 support
  - Establish a Qt5 build job that checks for compatibility & runs unit tests.
2. **Drop Qt6 support in a concerted effort.** So, create a big PR that removes all `#if QT_VERSION ...` code from all files.
  - Pro: Everything is clean after that, so new people reading the code won't be confused.
  - Con: I anticipate a lot of merge conflicts related to long-running PRs.
  - Con: I think it's easier to mess up but harder to review and test if this is done in one big PR.
3. **Gradually remove Qt6 support if you're touching a file anyway.** (which is what I would prefer).
  - Pro: Merge conflicts are limited to files that are would probably have conflicts anyway.
  - Pro: If you're touching some functionality, you are hopefully testing it anyway.
  - Pro: Reviews are easier if you're not changing too much code at once.
  - Con: No big, clean cut - people might need to read the documentation to see our guidance on Qt versions.
  - We can also do focused cleanups for certain areas of functionality in separate PRs if desired, to speed up the removal process.

As an additional datapoint, a [quick GitHub search](https://github.com/search?q=repo%3Amixxxdj%2Fmixxx+%23if+QT_VERSION&type=code) uncovered 77 files with some variation of `#if QT_VERSION`, some with multiple occurences per file (the most I found was 19 occurences in a single file).

Questions:
[...]

PR #14347, opened 2025-02-15
Remove Qt5 workaround from controller mapping

git-developer:
This PR removes a workaround for a bug in Qt5. See https://github.com/mixxxdj/mixxx/pull/14232#discussion_r1940949552 and #11473

Tested successfully with nightly release.

### DIFF
--- a/res/controllers/Behringer-Extension-scripts.js
+++ b/res/controllers/Behringer-Extension-scripts.js
@@ -282,12 +282,12 @@
         isEnabled: function() { return this.id !== 0; },
         start: function() {
             this.reset();
-            this.id = engine.beginTimer(this.timeout, function() {
+            this.id = engine.beginTimer(this.timeout, () => {
                 if (this.oneShot) {
                     this.disable();
                 }
                 this.action.call(this.owner);
-            }.bind(this), this.oneShot); // .bind(this) is required instead of arrow function for Qt < 6.2.4 due to QTBUG-95677
+            }, this.oneShot);
         },
         reset: function() {
             if (this.isEnabled()) {


ronso0:
Is there a strong reason for this?

The comment mentions Qt < 6.2.4 and we do not strictly require a certain Qt6 version (I'm building with 6.2.3 for example), so this change bares the risk of regressions -- and fixup support for users is tedious.

git-developer:
No, this is by no means required, just maintenance. Should I close?

Swiftb0y:
lets wait with the Qt5 removal until 2.6 has been released.

cr7pt0gr4ph7:
I'm less concerned with this PR itself (both merging or closing would be fine by me), and more with the implications that a partially-broken JS engine has on controller scripts...

According to [Minimum Requirements Policy](https://github.com/mixxxdj/mixxx/wiki/Minimum-requirements-policy) the baseline is defined by the latest Ubuntu LTS which has at least one point release. At the time of this writing [this is Ubuntu 24.04.01 LTS Noble Numbat](https://wiki.ubuntu.com/Releases), which _would_ allow us to bump up our minimum required version [to Qt 6.4.2](https://launchpad.net/ubuntu/noble/+package/libqt6qml6).

Given that [QTBUG-95677: Using `this` within arrow function expressions within class methods can point to wrong `this`](https://bugreports.qt.io/browse/QTBUG-95677) is quite a serious one, it _could_ make sense to bump up the required minimum version of Qt.

---

_Though it has turned out that is PR is not only about Qt5, here's the link to the discussion and decision about dropping Qt5 support from `main`:_
* #13826

cr7pt0gr4ph7:
> The primary reason why we're lagging behind the minimum requirements policy is because some devs wait even longer (or can't follow because their distro lacks behind Ubuntu (Debian or PopOS for instance)). It wouldn't be super nice to break their setup just because our policy allows it.

I agree, unnecessarily breaking setups does not make sense.

PR #14354, opened 2025-02-16
(fix) Controller preferences: make Enabled checkbox clickable again

ronso0:
Since #14006 both the Enabled checkbox and the 'Load mapping:' label are in the same cell of the QGridLayout.
The issue is: click events don't reach the checkbox anymore.

This can be fixed by putting the checkbox **after** the label in the ui file (ie. on top of the label, see first commit), but the better fix is to put both in a **QHBoxLayout** (fixup commit) since that also prevents checkbox / label overlap.

Is this actually an issue for anyone else? I'm building with Qt 6.2.3
If yes, does the first fix work you? (just curious, the QHBoxLayout should be merged)

### DIFF
--- a/src/controllers/dlgprefcontrollerdlg.ui
+++ b/src/controllers/dlgprefcontrollerdlg.ui
@@ -65,30 +65,34 @@
     </widget>
    </item>
    <item row="1" column="0">
-    <widget class="QCheckBox" name="chkEnabledDevice">
-     <property name="text">
-      <string>Enabled</string>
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
Tested this and it works as expected! Thank you for the fix, I didn't noticed it until now.

JoergAtGithub (approved):

ronso0:
Squashed, ready for merge.

PR #14357, opened 2025-02-17
(fix) Waveform scratching: fix seeking to hotcue (reduced, alt. for #14059)

### LINKED ISSUE #13981: pressing hotcue while holding waveform or spinny should seek to hotcue (and stay there)

ronso0:
This is #14059 without the member/slot refactoring, i.e. just the fixes + switch to parented_ptr etc.
I hope this lowers the barrier to get the fix into 2.5.1

Original description:
___
This is a first (still dirty) fix for #13981 

**TL;DR**
it's now possible to seek to hotcues while doing wave scratches.
Scratching at/across  loop boundaries (active) still works.

Recently I [compared the available methods for scratching with controllers](https://github.com/mixxxdj/mixxx/pull/14004), and while PositionScratchController (developed for waveform scratching) turned out to be the best (for me, no drift, smooth at low rates, nice spinback) I couldn't really use it as I want to because seeking to hotcues while scratching (holding the wheel, or turning very slowly) produced insanely high rates, and didn't stop at the cue pos.

Even though the bug is fixed I still consider this 'dirty' because I don't fully understand what was going wrong with the filters after seeking.

### DIFF
--- a/src/engine/positionscratchcontroller.cpp
+++ b/src/engine/positionscratchcontroller.cpp
@@ -95,6 +95,9 @@ PositionScratchController::PositionScratchController(const QString& group)
           m_isScratching(false),
           m_inertiaEnabled(false),
           m_prevSamplePos(0),
+          // TODO we might as well use FramePos in order to use more convenient
+          // mixxx::audio::kInvalidFramePos, then convert to sample pos on the fly
+          m_seekSamplePos(std::numeric_limits<double>::quiet_NaN()),
           m_samplePosDeltaSum(0),
           m_scratchTargetDelta(0),
           m_scratchStartPos(0),
@@ -165,6 +168,15 @@ void PositionScratchController::process(double currentSamplePos,
     if (m_mouseSampleTime >= kDefaultSampleInterval || !m_isScratching) {
         scratchPosition = m_pScratchPos->get();
         m_mouseSampleTime = 0;
+    }
+
+    bool adoptSeekPos = false;
+    if (!util_isnan(m_seekSamplePos)) {
+        // If we were notified about a seek, adopt the new position immediately.
+        m_prevSamplePos = m_seekSamplePos;
+        m_seekSamplePos = std::numeric_limits<double>::quiet_NaN();
+
+        adoptSeekPos = true;
     }
 
     if (m_isScratching) {
@@ -269,7 +281,11 @@ void PositionScratchController::process(double currentSamplePos,
                     m_scratchTargetDelta = scratchTargetDelta;
                 }
 
-                if (calcRate) {
+                // If we just adopted the seek position we need to avoid false
+                // high rate and simply report the previous rate.
+                // It'll adapt to the scratch speed in the next run.
+                // Setting rate to 0 has the same effect apparently.
+                if (calcRate && !adoptSeekPos) {
                     double ctrlError = m_pRateIIFilter->filter(
                             scratchTargetDelta - m_samplePosDeltaSum);
                     m_rate = m_pVelocityController->observation(ctrlError);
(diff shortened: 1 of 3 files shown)


src/engine/positionscratchcontroller.cpp
```diff
@@ -187,7 +203,7 @@ void PositionScratchController::process(double currentSamplePos,
                 bool reverse = triggerPos < targetPos;
                 double loopLength = reverse ? -1 * (targetPos - triggerPos)
                                             : triggerPos - targetPos;
-                if (wrappedAround > 2) {
+                if (wrappedAround > 1) {
```
daschuer, line 206: I think I can confirm that this fixes an issue, however I think it still does not behave right in case of reverse. 
This is because the reverse case is considered loopLength but not for `sampleDelta +=` below. 
It can become via math: 
`sampleDelta +=  (triggerPos - targetPos ) + (currentSamplePos - m_prevSamplePos);` 

Can we refactor the whole block into a forward and a backward branch? 

Something like (untested): 

```
            double sampleDelta = currentSamplePos - m_prevSamplePos;
            if (wrappedAround > 0) {
                // If we wrapped around, add  loop length for each wrap-aound 
                // (necessary if the buffer is longer than the loop, e.g. when looping 
                // at high rates / with short loops. 
                double triggerPos = trigger.toEngineSamplePos();
                double targetPos = target.toEngineSamplePos();
                double loopLength = triggerPos - targetPos;
                if (loopLength < 0.0) {
                    // reverse 
                    loopLength *= -1; 
                }
                sampleDelta += wrappedAround * loopLength;
            }
``

File: src/engine/positionscratchcontroller.cpp [hunk 2]
```diff
@@ -129,6 +132,15 @@ void PositionScratchController::process(double currentSamplePos,
     m_pVelocityController->setPD(p, d);
     m_pRateIIFilter->setFactor(f);
 
+    bool adoptSeekPos = false;
+    if (!util_isnan(m_seekSamplePos)) {
+        // If we were notified about a seek, adopt the new position immediately.
+        m_prevSamplePos = m_seekSamplePos;
```
daschuer, line 138: How does the m_seekSamplePos relates to currentSamplePos? 
Can you add a comment? Are they equal for one call?

src/engine/positionscratchcontroller.cpp
ronso0: Thanks, we can try that.
Because this is actually not related to the scratch/seeking issue I've removed the commit from this branch and will create a separate one.

ronso0:
In order to keep you review I have decided to close #14059, rebased this one (fix conflicts) and adopt the only different commit from #14059.
Also reworded a commit because the already merged parented_ptr commit already contains the PositionScratchController refactoring.

Will address your review soon.

src/engine/positionscratchcontroller.cpp
ronso0: took me a while with my non-math brain...
Yup, this looks good!

PR is #14379

src/engine/positionscratchcontroller.cpp
daschuer: I have tested a bit an it turns out that the new position is applied second `process()` call so you need to skip one or put the adoption of the seek position at the bottom of the process() call. The `adoptSeekPos` does also not work right. I think we need to reset the `m_samplePosDeltaSum` instead. I have tried this but it fails as well. At one version it seeks to the cue point and the immediately back to the original position. So we probably need to tweak the value and just subs tact the seek distance or such. (not tested). 

Testing with the maximum audio buffer size was by the way the best to trace what actually happens.

ronso0: > I have tested a bit an it turns out that the new position is applied second process()

FWIW I could never observe that **exact** the seek position is used for `PositionScratchController::process()`, there was always a diff (usable minimum was ~100 samples). I figured that with a version where I waited until the currPos and stored seekPos were close and then adopted the seekPos as prevPos https://github.com/mixxxdj/mixxx/commit/6927fdc727422a8491f35215732946291c6ac257

> I think we need to reset the m_samplePosDeltaSum instead. I have tried this but it fails as well.

Thought so, too, and tried it as well as some other combinations, but none worked.
Except this one.

> At one version it seeks to the cue point and the immediately back to the original position.

Yes, IIUC that is the original issue that causes the high rate.

> So we probably need to tweak the value and just subs tact the seek distance or such. (not tested).

If you like to debug further (I don't) and you figure something I'm happy to test it!

ronso0:
> there's always a visual stop and a dip in rate

I think that occured simply because I used hotkeys for seeking.
If I use a virt. MIDI script and let that do the timed seeking all is good, the waveform sticks to the pointer even while seeking :+1:

daschuer:
Here is a final bit: https://github.com/ronso0/mixxx/pull/91

daschuer (approved):
LGTM, thank you.

Issue #14297, opened 2025-02-05
Reduce the CI footprint


acolombier:
Currently, all our workflows run unconditionality from the type of change, this is leading to running the whole build and checks process on changes that may not require it (e.g doc, or os-specific update). 
Mapping changes are also a big topic, as they arguably don't need to run the build process to be tested (or maybe just once, before we merge, to ensure the manifest correctness)

![Image](https://github.com/user-attachments/assets/26dad673-9719-430d-a687-3c339dba787d)

**Build and checks constitute 98% of our CI footprint**. Looking at the recently merged 

Here are some recently merged (when this issue is being reported) PR which could have partially or entirely skip the build process

- https://github.com/mixxxdj/mixxx/pull/14183
- https://github.com/mixxxdj/mixxx/pull/14246
- https://github.com/mixxxdj/mixxx/pull/14289

Filtering workflow can be tricky, especially when using the `on.<event>.path` since they may be complex usecase where you want to run the workflow anyway. 

This can however easily be addressed by using tag-based trigger, as per this [Labeller example](https://github.com/actions/labeler?tab=readme-ov-file#outputs). One limitation tho, is that the `github-bot` (a.k.a generated `GITHUB_TOKEN`) doesn't trigger subsequent build (which you need to have `on.pull_request.type.labelled` working). An easy workaround is to create a app (e.g `mixxx-bot`) on the repo and use its credentials to run labeller. It becomes then just a matter of keeping `.github/labeler.yml` up to date.

My proposal is to:
- [ ] Create a Github App for the labeller (To be done by an admin or maintainer)
- [ ] Update the workflow to trigger upon certain labels

No change would be expected for `main`

Swiftb0y:
I'm too unfamiliar with github actions, but sometimes jobs just fail because a test or some external tool (code signing) is flaky/broken. Having to rerun all the expensive CI steps (building+testing) just so a subsequent step can try again at succeeding seems super wasteful to me. It would be great if there was a way to make this more granular. I think that may also save quite some CI hours. 

Apart from that, we should verify whether our build-artifact caching works correctly. It would save a lot of time and resources if PRs would only need to rebuild the few translation units that we're actually modified in the PR. I doubt this is currently the case.

acolombier:
Yeah I agree - I think caching a breaking down step could be quite helpful. Honestly, actions could be a rabbit hole tho, so I feel it might be worth for me to focuse on conditional build, before we introduce caching and steps with artifacts.

[...]

Issue #12266, opened 2023-11-05
CI doesn't fail when QML GUI is not loadable


JoergAtGithub:
### Bug Description

Currently the QML user interface is not loadable on Windows (and most likely on macOS too), but our CI builds doesn't detect this.

### Version

Main

### OS

Windows 11

daschuer:
QML work is continued in 2.6

acolombier:
Note to anyone willing to fix QML on Windows (and Mac?) - In order to make it run in #14360, I had to use [`windeployqt6.exe`](https://doc.qt.io/qt-6/windows-deployment.html) to add missing Qt components to the working directory (where `mixxx.exe` will be ran) as following

```ps
&"${env:MIXXX_VCPKG_ROOT}\installed\x64-windows-release\tools\Qt6\bin\windeployqt6.exe" ./mixxx.exe --dir . --libdir bin --plugindir plugins --qml-deploy-dir qml --qmldir ../res/qml/
```

JoergAtGithub:
I just tried `mixxx.exe --qml` with latest `main` and the QML GUI came up.

acolombier:
Weird, it doesn't work out of the box for me on my Win 11 VM. Are you using Ninja as generator on CMake?

JoergAtGithub:
Yes I use Ninja, everything is default, no settings modified. Runs from both, build and install directory.

acolombier:
Right, I think that's my problem - I'm using Visual Studio generator

JoergAtGithub:
Just run `windows_buildenv.bat`, it should set everything as needed.

JoergAtGithub:
Do you build with IDE or from command line?

acolombier:
Command line - Just retry on a fresh machine - It compiles with `Visual Studio 16 2019` generator by default, which stores the target in subfolder named after the build type (`RelWithDebInfo`, `Debug`, ...) which I think was always the reason for these problem I was having on windows.

`ninja` is installed an visible in PATH so not sure why it default to VS

JoergAtGithub:
Do you use the correct shell `x64 Native Tools Command Prompt for VS 2019` as described in the Readme.md ?

acolombier:
I use the PowerShell version. Could that be the problem? I thought there would be interchangeable?

JoergAtGithub:
No, only this particular shell works (or the equivalent with suffix 2022 if you use VS 2022).

acolombier:
[...]

PR #14360, opened 2025-02-17
Add UI tests for QML

acolombier:
This is a naive attempt to address #12266

Note that two commits were also added to address low hanging fruits of #14297 and reduce amount of compute used by our our CI

Note: I was unable to get it running on Mac - I'll see if I can get something but this will likely have to be skipped at first since I don't own a Mac and have limited interest in supporting Apple ecosystem

Depends on https://github.com/mixxxdj/vcpkg/pull/171

Disabling on Windows 11 arm as the runner VM appears not to be correctly setup.
image

UI results will be visible in the job summary. I couldnt get it to work in the comment for now!
image


Eve00000:
>     * Record/publish video?

Yes, and we can test on Mac VM.

Eve00000:
> Do you mean manual testing?

we need to start somewhere

JoergAtGithub:
Did you have a look at: https://github.com/faaxm/spix
I've no experience with it, but it seems to me pretty much what is needed here.

tools/windows_buildenv.bat
```diff
@@ -29,8 +29,8 @@ IF DEFINED BUILDENV_RELEASE (
 ) ELSE (
     SET BUILDENV_BRANCH=2.6
     SET VCPKG_TARGET_TRIPLET=x64-windows
-    SET BUILDENV_NAME=mixxx-deps-2.6-x64-windows-12239ed
-    SET BUILDENV_SHA256=5c60b2c61d6448a99979d7cc997e92f2fd5b4b65f65e2439abfca3fa6fd30f8d
+    SET BUILDENV_NAME=mixxx-deps-2.6-x64-windows-7cc6152
+    SET BUILDENV_SHA256=44b0fc493959ffec09dc48af3b4400d3e550aed970187222f95558942a42e066
```
JoergAtGithub, line 33: In CI we use the release only version of the buildenv. The full version with additional debug binaries is for local development. This change will not enable ffmpeg therfore.

acolombier: Yes, I've commented that part out in the CI so I can start trying and see if I can get ffmpeg to work before submitting a PR to make a new vcpkg rel version

acolombier:
@JoergAtGithub I have resync'ed this branch and updated it with the latest module refactor from Spix - good news is, it is working great... and detected the bug, fixed by #15361.

Video results can be found in the [Artifacts sections](https://github.com/mixxxdj/mixxx/actions/runs/18935253747?pr=14360) - I'll spend some time to add a nice auto-comments with links to video recording, but in the meantime, it would be good to merge this other PR so we can fix the tests

JoergAtGithub:
CI is failing

acolombier:
See my message above - it is failing, because QML is broken. Agree this is cryptic and I will add a nice message, either in build summary or in PR comment, but it will keep failing till we actually fixed QML

acolombier:
Finally ready!

acolombier:
A bit of back and forth, but we should now have something stable. I had to do some changes on Android setup, so not sure if you want me to split the commits?

JoergAtGithub:
There seems to be quite a lot of changes in CI setup. Would it make sense to move them in another PR?

acolombier:
> Would it make sense to move them in another PR?

All the changes are quite untangled so this is not an easy task. I have split the different changes in specific commits tho so hopefully this helps with review and history. Is that an okay trade off?

acolombier:
Friendly ping @JoergAtGithub

src/coreservices.cpp
```diff
@@ -615,7 +615,10 @@ void CoreServices::initialize(QApplication* pApp) {
 
     bool musicDirAdded = false;
 
-    if (m_pTrackCollectionManager->internalCollection()->loadRootDirs().isEmpty()) {
+    if (!m_cmdlineArgs.getDeveloper() &&
```
JoergAtGithub, line 618: Why does the music directory depend on developer mode?

.github/workflows/build.yml
```diff
@@ -71,17 +66,18 @@ jobs:
             artifacts_name: Ubuntu 24.04 Qt6 DEB
             artifacts_path: build/*.deb
             artifacts_slug: ubuntu-jammy
+            coverage: ON
+            qml: ON
+            warning_fatal: OFF
+            build_type: Debug
+            optimize: off
```
JoergAtGithub: This seems to change our defaults.

.github/workflows/build.yml
```diff
@@ -446,6 +448,118 @@ jobs:
...
 
+      - name: "[Ubuntu] Setup UI dependencies for QML UI Test"
+        if: runner.os == 'Linux' && matrix.qml == 'ON' && matrix.crosscompile != true
+        run: |
+          sudo apt-get update && sudo apt-get install -y xvfb ffmpeg
+          echo "QT_QPA_PLATFORM=xcb" >> "$GITHUB_ENV"
+        working-directory: build
+
+      - name: "[Windows] Setup UI dependencies for QML UI tests"
+        if: runner.os == 'Windows' && matrix.qml == 'ON' && matrix.crosscompile != true && matrix.os != 'windows-11-arm' # Disabling Windows 11 ARM as the VM doesn't appears to be correctly provisioned
```
JoergAtGithub, line 453: What is happening on Windows 11 ARM? Please create an issue with the details for this, and refer it in a comment here.

JoergAtGithub (changes requested):
- Please move the changes for the Coverage build to an own PR, as it seems unrelated to the QML UI tests. Combining unrelated changes into a single PR, is a practice that often leads to delays in approval.
- Please check the description of the PR. I think what is written there is not correct/up-to-date.

src/coreservices.cpp
acolombier: This is to circumvent the current behaviour, where at first launch, there is dialog asking to select a music directory. The UI test does not support QWidget element and the QML feature which will replace that is still to be done.
This allows getting the main windows immediately, when launching with `--developer`

.github/workflows/build.yml
acolombier: These are the coverage default

Image

This allows to save a build, by removing a run and performing coverage here. FWIW, I am not sure why we need `Debug`/`optimise=OFF` tho.

.github/workflows/build.yml
acolombier: Please see the PR description, where I have detailed the problem.

src/coreservices.cpp
JoergAtGithub: But this impacts also users who starts Mixxx in developer mode.

.github/workflows/build.yml
JoergAtGithub: Until now, the Ubuntu 24.04 artifacts are build with -DWARNINGS_FATAL=ON, -DCMAKE_BUILD_TYPE=RelWithDebInfo and generation of coverage files off. Please revert this unrelated change!

src/coreservices.cpp
acolombier: Yes. The rational is that a user starting Mixxx for the first time AND using the `--developer` is unlikely. In the case this happens, they can still add music folder in the preferences.

So the only risk would be for user new to Mixxx, to launch Mixxx for the first time AND use the `--developer`, which sounds very unlikely.

If you feel strongly about it, happy to add `--qml` to the condition.

.github/workflows/build.yml
acolombier: Will move this commit to a separated PR.

acolombier: Removed 3927cef84b98434f13ef2aa17adff28d930b3c80

JoergAtGithub:
There is now a conflict

acolombier:
I am planning to look into some BDD+RPC still improvement here, to see how we could easily generate test cases

PR #14361, opened 2025-02-17
(feat) show popup with Xfader curve when curve parameters changed (eg. by script)

ronso0:
Just a little helper to verify the xfader curve when setting it with a controller's switch
xfader-popup

* reuses the graphic from Mixer preferences
* shown centered on screen, can be dragged elsewhere (position is restored on next show)
* hide after 3 sec (after show or last dnd click) (make it 1.5 - 2 sec)
* close on click

If someone wants to test it without controller, here's a mapping for Vir-MIDI or MIDI Through
https://gist.github.com/ronso0/ae7a6f3fc2b28e43f16bf50bfb84e62e

### TODO
- [ ] make this optional

### TODO (custom)
- [ ] show when GUI is ready, no auto-hide then? so it's not missed?

### DIFF
--- a/src/preferences/dialog/dlgpreferences.cpp
+++ b/src/preferences/dialog/dlgpreferences.cpp
@@ -180,11 +180,16 @@ DlgPreferences::DlgPreferences(
             tr("Decks"),
             "ic_preferences_decks.svg");
 
+    auto* pDlgPrefMixer = new DlgPrefMixer(this, pEffectsManager, m_pConfig);
     addPageWidget(PreferencesPage(
-                          new DlgPrefMixer(this, pEffectsManager, m_pConfig),
+                          pDlgPrefMixer,
                           new QTreeWidgetItem(contentsTreeWidget, QTreeWidgetItem::Type)),
             tr("Mixer"),
             "ic_preferences_crossfader.svg");
+    connect(this,
+            &DlgPreferences::showXfaderPopupPersist,
+            pDlgPrefMixer,
+            &DlgPrefMixer::slotShowXfaderPopupPersist);
 
     addPageWidget(PreferencesPage(
                           new DlgPrefEffects(this, m_pConfig, pEffectsManager),
--- a/src/preferences/dialog/dlgprefmixer.cpp
+++ b/src/preferences/dialog/dlgprefmixer.cpp
@@ -102,14 +102,16 @@ void DlgXfaderCurve::setScene(QGraphicsScene* pScene) {
     adjustSize();
 }
 
-void DlgXfaderCurve::show() {
+void DlgXfaderCurve::show(bool autoHide) {
     if (graphicsViewXfader->scene() == nullptr) {
         return;
     }
     // This saves us from resetting pressed on close.
     m_mousePressed = false;
     QDialog::show();
-    m_hideTimer.start();
+    if (autoHide) {
+        m_hideTimer.start();
+    }
 
     if (m_moved) {
         return;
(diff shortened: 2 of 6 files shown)


JoergAtGithub:
I like the idea of instantly visualizing the fader curves, but the pop up dialog seems to be to much visual noise to me.
Did you consider to display this information silent in the fader widget itself, like:
grafik

ronso0:
Btw this has now become the "I'm alive!" sign for my mapping, especially after debugging.
Like, I know it managed to create all components and reached init() : )

src/preferences/dialog/dlgpreferences.cpp
```diff
@@ -180,11 +180,16 @@ DlgPreferences::DlgPreferences(
             tr("Decks"),
             "ic_preferences_decks.svg");
 
+    auto* pDlgPrefMixer = new DlgPrefMixer(this, pEffectsManager, m_pConfig);
```
acolombier, line 183: How about using a `parented_ptr` here instead of a fat pointer?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -65,6 +69,128 @@ bool isMainEQ(EffectManifest* pManifest) {
...
+    QWidget* centerOverWidget = parentWidget();
+    VERIFY_OR_DEBUG_ASSERT(centerOverWidget) {
+        qWarning() << "DlgXfaderCurve does not have a parent widget";
+        centerOverWidget = this;
+    }
+
+    const QScreen* const pScreen = mixxx::widgethelper::getScreen(*centerOverWidget);
+    QRect screenGeometry;
+    VERIFY_OR_DEBUG_ASSERT(pScreen) {
+        qWarning() << "Assuming screen size of 800x600px.";
```
acolombier: Since, this is a warning, how about clarifying the issue?

```suggestion
        qWarning() << "No screen could be found. Assuming screen size of 800x600px.";
```

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -65,6 +69,128 @@ bool isMainEQ(EffectManifest* pManifest) {
...
+        screenGeometry = QRect(0, 0, 800, 600);
+    }
+    else {
```
acolombier, line 135: This is an interesting "abuse" of `VERIFY_OR_DEBUG_ASSERT`! Tho, I suspect, it may fail if you run with debug assertion and continue with the debugger when you hit one?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -65,6 +69,128 @@ bool isMainEQ(EffectManifest* pManifest) {
...
+        }
+
+        m_mousePressed = true;
+        m_clickTimer.start();
+        m_hideTimer.start();
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+        const QPoint eventPosition = pME->globalPosition().toPoint();
+#else
+        const QPoint eventPosition = pME->globalPos();
+#endif
```
acolombier, line 162: Do we want to keep supporting Qt < 6.x?

src/preferences/dialog/dlgprefmixer.cpp
```diff
@@ -893,14 +1022,16 @@ void DlgPrefMixer::updateXFaderWidgets() {
...
     if (m_pxfScene) {
         m_pxfScene->clear();
     } else {
         m_pxfScene = make_parented<QGraphicsScene>(this);
         // The size of the QGraphicsView doesn't change so we need to do this only once
         graphicsViewXfader->setLineWidth(1); // frame width
-        int sizeX = graphicsViewXfader->width() - 2;
-        int sizeY = graphicsViewXfader->height() - 2;
+        // int sizeX = graphicsViewXfader->width() - 2;
+        // int sizeY = graphicsViewXfader->height() - 2;
```
acolombier: Looks like this was moved above - delete?

src/preferences/dialog/dlgpreferences.cpp
ronso0: I was following the existing pref page pattern here, but of course we can migrate all pages to `parented_ptr`

src/preferences/dialog/dlgprefmixer.cpp
ronso0: Well, that's the purpose VERIFY_OR_DEBUG_ASSERT, no?
If one runs into it we can debug when it occurs and maybe even resolve it?

src/preferences/dialog/dlgprefmixer.cpp
ronso0: As of now, besides my #15251 (which I opened much later), there have been no efforts to finally remove Qt5, so I just followed the compat pattern.

ronso0:
I pushed a small fixup, should be fine now.

I'm tempted to remove the "show on startup" commit. I think it's just not necessary as this entire tool is only for a niche anyway. What do you think?

src/preferences/dialog/dlgprefmixer.cpp
JoergAtGithub: @acolombier Could you please explain the problem you see here?

PR #14367, opened 2025-02-17
(fix) Hotcue press sets drag start pos to use correct distance threshold

### LINKED ISSUE #14365: regression: can't use hotkeys while dragging hotcue button
### Bug Description

Unfortunately dnd can happen by accident and then one can e.g. not press the 'play' hotkey.
Noticed with Ubuntu + xfce, not sure how other desktops behave or if it's Qt behavior.

Maybe we can increase the dnd threshold for hotcue buttons, eg. half width?

### Version

2.6-alpha

### OS

_No response_

ronso0:
Previously, drag'n'drop could be initiated unintentionally very easily because `DragAndDropHelper` didn't have the correct drag start point, ie. starts a drag for even the shortest cursor move and thereby prohibits use of hotkeys
(behaviour may depend on desktop environment).

The result is that hotkeys could not be applied during unintentinal DnD.

Fixes #14365

### DIFF
--- a/src/widget/whotcuebutton.cpp
+++ b/src/widget/whotcuebutton.cpp
@@ -155,6 +155,7 @@ void WHotcueButton::mousePressEvent(QMouseEvent* pEvent) {
     // starting the preview.
     if (!pEvent->modifiers().testFlag(Qt::ShiftModifier)) {
         WPushButton::mousePressEvent(pEvent);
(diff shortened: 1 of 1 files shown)


ronso0:
Ups, draft state was not intentional, this is ready for merge.

acolombier (approved):
Tested and LGTM, thanks!

PR #14372, opened 2025-02-18
fix: use correct detected channel count on CoreAudio

### LINKED ISSUE #14089: Regression: Mixxx 2.6 reanalyzes tracks on every load
### Bug Description

See the associated Zulip thread: https://mixxx.zulipchat.com/#narrow/channel/109122-general/topic/Mixxx.20reanalyzes.20tracks.20every.20time

Mixxx 2.6 seems to reanalyze tracks every time they are loaded to a deck instead of using the cached data:

https://github.com/user-attachments/assets/25b00322-11d0-4d39-a6cf-ab916401dbb1

According to a bisect 7a3c5151a3 was the last good, 9b31059fba the first bad one ([diff](https://github.com/mixxxdj/mixxx/compare/7a3c5151a3a3eb286bbc6c30f48b649dc41ce567...9b31059fbacdd2d66bcf397b2354bb32c7ad1872)), so the issue was very likely introduced here:

- https://github.com/mixxxdj/mixxx/pull/13086

cc @acolombier 

### Version

2.6 (latest main)

### OS

macOS 15.1.1

acolombier:
Fixes #14089

Would be great if you give this a test @fwcd @m0dB

### DIFF
--- a/src/sources/soundsourcecoreaudio.cpp
+++ b/src/sources/soundsourcecoreaudio.cpp
@@ -3,6 +3,7 @@
 #include "engine/engine.h"
 #include "sources/mp3decoding.h"
 #include "util/appleosversion.h"
+#include "util/assert.h"
 #include "util/logger.h"
 #include "util/math.h"
 
@@ -114,13 +115,21 @@ SoundSource::OpenResult SoundSourceCoreAudio::tryOpen(
     }
     m_bFileIsMp3 = m_inputFormat.mFormatID == kAudioFormatMPEGLayer3;
 
+    VERIFY_OR_DEBUG_ASSERT(m_inputFormat.mChannelsPerFrame != 0) {
+        kLogger.warning()
+                << "File"
+                << fileName
+                << "appears to have no audio channels";
+        return OpenResult::Failed;
+    }
+
     // create the output format
-    const UInt32 numChannels =
+    const UInt32 maxChannels =
             params.getSignalInfo().getChannelCount().isValid()
             ? params.getSignalInfo().getChannelCount()
-            : mixxx::kEngineChannelOutputCount;
+            : mixxx::kMaxEngineChannelInputCount;
     m_outputFormat = CAStreamBasicDescription(m_inputFormat.mSampleRate,
-            numChannels,
+            std::min(m_inputFormat.mChannelsPerFrame, maxChannels),
             CAStreamBasicDescription::kPCMFormatFloat32,
             true);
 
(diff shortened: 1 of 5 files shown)


src/sources/soundsourcecoreaudio.cpp
```diff
@@ -115,9 +115,8 @@ SoundSource::OpenResult SoundSourceCoreAudio::tryOpen(
     m_bFileIsMp3 = m_inputFormat.mFormatID == kAudioFormatMPEGLayer3;
 
     // create the output format
-    const UInt32 numChannels =
-            params.getSignalInfo().getChannelCount().isValid()
-            ? params.getSignalInfo().getChannelCount()
+    const UInt32 numChannels = m_inputFormat.mChannelsPerFrame != 0
+            ? m_inputFormat.mChannelsPerFrame
```
fwcd, line 119: Do `params` contain bogus information here, should we fix that?

fwcd (commented):
Thanks for looking into this so quickly! CI currently seems to fail because `params` is unused, but I'll have a look whether I can test this locally

src/sources/soundsourcecoreaudio.cpp
acolombier: `params` contains the maximum number of  supported channels. This used to always be `2` before STEMs were introduced.

fwcd: Ah hm, do the other sound sources not parse the channel count from `params`?

acolombier: No they should detect it from the file they are loading. I remember updating it (and forgot CoreAudio since it's on Mac only) but I'll give it a sanity check

acolombier:
Interesting! I worked off a fresh Mixxx install so I couldn't catch this. It's worth to say the analysis is likely corrupted as the track (and waveform) will be know to have 8 channels, but my understanding was that it should only reanalyse it once. I'll see if I can make sense of this

acolombier:
Other SoundSource like the MediaFoundation seem to suffer from the same problem. Most of the source seems to ignore the open params, so I cannot tell if `params.channelCount` is meant to be the maximum channel to open or the expected one, since soundsource have the ability to explicitly account for how many channel they will effectively use (using `initChannelCountOnce`)

I've decide to interpret as "up to `params.channelCount`" and provided a fix accordingly.

I also simplified the invalidation condition and remove the case that also invalidate if the waveform had stem data but the track is not a stem track anymore as this makes no sense and likely add more room for bug. Would appreciate if you can test that again :)

fwcd (approved):
Seems to fix it for both mp3 and m4a on macOS, very nice!

fwcd:
> I've decide to interpret as "up to params.channelCount" and provided a fix accordingly.

Sounds reasonable! The code looks good to me, thanks!

PR #14375, opened 2025-02-19
CMakeLists: Disable `QTlsBackendOpenSSLPlugin` on iOS

fwcd:
The current iOS Qt 6 build (as of https://github.com/fwcd/vcpkg/commit/57ce3aded1b447dbb618f76a0cfc9bfe1c52775e) doesn't seem to support this plugin:

```
CMake Error at CMakeLists.txt:3464 (target_link_libraries):
  Target "mixxx-lib" links to:

    Qt6::QTlsBackendOpenSSLPlugin

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



-- Generating done (0.9s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

Hence this PR disables it when targeting iOS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3468,9 +3468,16 @@ if(Qt_IS_STATIC)
       Qt${QT_VERSION_MAJOR}::QSvgPlugin
       # sqldrivers
       Qt${QT_VERSION_MAJOR}::QSQLiteDriverPlugin
-      # network plugins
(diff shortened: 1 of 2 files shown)


acolombier (approved):
LGTM, thanks!

PR #14376, opened 2025-02-19
DlgPrefController: Cond-compile out HID settings when building without HID

fwcd:
This fixes a regression in the iOS build, which doesn't support HID and therefore relies on everything HID-related being gated behind `__HID__`.


acolombier (approved):
LGTM, thanks!

PR #14379, opened 2025-02-20
(fix) PositionScratchController: fix loop wrap-around case

ronso0:


src/engine/positionscratchcontroller.cpp
```diff
@@ -199,27 +199,18 @@ void PositionScratchController::process(double currentSamplePos,
...
                 double triggerPos = trigger.toEngineSamplePos();
                 double targetPos = target.toEngineSamplePos();
-                bool reverse = triggerPos < targetPos;
-                double loopLength = reverse ? -1 * (targetPos - triggerPos)
-                                            : triggerPos - targetPos;
-                if (wrappedAround > 2) {
-                    sampleDelta = (wrappedAround - 1) * loopLength;
+                double loopLength = triggerPos - targetPos;
+                if (loopLength < 0) {
+                    loopLength *= -1;
```
ronso0, line 211: This is not necessary, the trigger / target returned by [`LoopingControl::nextTrigger`](https://github.com/mixxxdj/mixxx/blob/4c1b8813fba2029047dc06656af25d6e58ea6d52/src/engine/controls/loopingcontrol.cpp#L410) are already swapped in case of reverse.

daschuer (approved):
Removed code is good code :-) confirmed that it works like a charm. Thank you.

ronso0:
Great, thanks for your review!

PR #14381, opened 2025-02-20
(fix) fix file dialog extension handler, playlist export

### LINKED ISSUE #14362: MIXXX SOFTWARE 2.5.0
### Bug Description

Hello, I have noticed that this version does not allow export a play list , in iny format: M3U, M3U8, PLS to Itunes, simultaneously I cannot drag form MIXXX to Itunes
thanks for taking notice
best regards

### Version

2.5.0

### OS

WIndows 11 64bit

ronso0:
Yet another workaround for QFileDialog not appending the selected extension if the filename doesn't end with it.
(follow-up for #11332)

Steps to reproduce:
* select playlist, open menu, clcik Export
* insert file name without extension
* select "m3u8" (default is m3u)
* Ok
* dialog returns /home/user/Music/name
* handler should turn it into /home/user/Music/name.m3u8
* 2.5 bug: /home/user/Musi.m3u8

Btw this util is also used by the sampler bank export, so should be reproducible with that, too.

Probably fixes #14362

### DIFF
--- a/src/util/file.cpp
+++ b/src/util/file.cpp
@@ -4,6 +4,8 @@
 #include <QRegExp> // required for 'indexIn(QString &str, int pos)
 #include <QRegularExpression>
 
+#include "util/assert.h"
+
 namespace {
 
 const QRegularExpression kExtractExtensionRegex(R"(\(\*\.(.*)\)$)");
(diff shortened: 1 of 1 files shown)


daschuer (changes requested):
I can confirm the issue is fixed. Two code improvements.

acolombier:
@daschuer May I ask you don't submit "Change requests" reviews on code smell and nitpicking as this shouldn't not block the merge. It is good to point them out and most of the time, the effort will be made to address them, but we shouldn't abuse the "veto power" for these.

daschuer:
I don't think we have a problem here. A review is either successful and can be merged right away. If not it is failed and the reviewer is now in the loops to verify that the failure is fixed. The contributor can argue that it is not worth the effort in cases of this. Than the reviewer can review again and and merge or propose an alternative solution. 
Without a failed review another contributor can review and merge. This is now permitted which is intended by me.

ronso0:
> Without a failed review another contributor can review and merge.

This is also how I perceive "Changes requested": it simply reduces the risk that another user merges a PR after just skimmin through and noticing **comments** reviews with code improvement proposals.
Btw I agree with the proposals made here, thanks.

acolombier:
Since we expect other contributor to merge the PR, I assume that they wouldn't not blindly merge a PR which have unacked comments. At least this is how I feel this is working with some other core members. Anyway, if @ronso0 and you are happy working this way, then I don't want to disrupt your workflow.

ronso0:
This is not necessarily "my workflow", it's just that I'm fine with both change requests and omments with change requests.

ronso0:
I adopted the proposals, ready to roll!

daschuer (approved):
Thank you. LGTM

PR #14385, opened 2025-02-20
Tracks: `show_track_menu` at index position

ronso0:
... like with a regular mouse context menu event.

Just a little helper.
My use case: On the Traktor S4 one button toggles the track menu for a deck or, with Shift, for the library.
I want to be sure which one it is when I select "Move To Trash", just in case I released the Shift button too early / or still had it pressed for whatever reason.

### DIFF
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -501,42 +501,85 @@ void WTrackTableView::slotUnhide() {
 
 void WTrackTableView::slotShowHideTrackMenu(bool show) {
     VERIFY_OR_DEBUG_ASSERT(m_pTrackMenu.get()) {
-        return;
+        initTrackMenu();
     }
     if (show == m_pTrackMenu->isVisible()) {
         emit trackMenuVisible(show);
         return;
     }
     if (show) {
-        QContextMenuEvent event(QContextMenuEvent::Mouse,
-                mapFromGlobal(QCursor::pos()),
-                QCursor::pos());
-        contextMenuEvent(&event);
+        const auto selectedIndices = selectionModel()->selectedIndexes();
+        if (selectedIndices.isEmpty()) {
+            // If selection is empty, contextMenuEvent() won't work anyway.
+            return;
+        }
+
+        // Show at current index if it's valid. When using only a controller for
+        // for track selection, there's only on row selected.
+        // If it's not part of the selection, like when Ctrl+click was used to
+        // deselect a row, we use the first selected row and the column of the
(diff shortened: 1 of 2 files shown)


src/widget/wtracktableview.cpp
```diff
@@ -508,9 +508,20 @@ void WTrackTableView::slotShowHideTrackMenu(bool show) {
...
+        // Show at current index if it's valid, else at cursor position
+        QPoint evPos;
+        auto currIdx = currentIndex();
+        if (currIdx.isValid()) {
+            evPos = mapToGlobal(visualRect(currIdx).center());
+            // The viewport start below the header, but mapToGlobal() apparently
+            // uses the rect() as reference. Add header height to y and we're good.
+            evPos += QPoint(0, horizontalHeader()->height());
+        } else {
+            evPos = QCursor::pos();
```
ronso0, line 520: might as well use `getSelectedRows().first()` here (if it's not empty!)

ronso0: naah, use first selected index (that's what contextMenuEvent() uses to load the track menu), then shift to column of current index if that is valid.

acolombier:
This is a great improvement! I use that feature all the time and always gets frustrated when the menu spawns on the other side of the screen, where the unused cursor is.
Thanks for looking into this!

ronso0:
I tweaked it a bit to cover more of the possible situations:
1. current index is (valid and) part of the selection:
use index position
2. current index is (valid and) NOT part of the selection:
use first selected row and index column
3. current index is invalid (not sure if that can happen, maybe after hide/remove/purge?) and there is a selection:
use cursor position as before.
(might use top left index of selection but since it's not clear whether we can get this situation at all, I didn't implement it)

BUT when the menu is taller than the tracks view, its position doesn't really clarify where it belongs, so a visual hint might be more helpful. Eg. another border color. Will certainly try that, but won't propose to merge this considering showing the menu via controller is already a niche case.
(hmm, might as well set a custom property on WTrackMenu when it's invoked via slotShowHideTrackMenu..? and add a style to the common scheme stylesheet? might work)
Or wdyt @acolombier ?

ronso0:
Okay, I think this is now working very well.

I can revert to https://github.com/mixxxdj/mixxx/commit/9d9128f85af3ea402734fb1cc0b40c3989ef7537 if you think the changes are too much for 2.5

daschuer (approved):
LGTM and still working. Thank you.

ronso0:
Thank you for testing!

PR #14390, opened 2025-02-22
Merge 2.5

ronso0:
There were many conflicts in the ts files.
I did keep the man version, but when I checked one the line references didn't match.
Guess we just need to `lupdate`?


src/engine/positionscratchcontroller.h
```diff
@@ -42,4 +53,13 @@ class PositionScratchController : public QObject {
     double m_rate;
     double m_moveDelay;
     double m_mouseSampleTime;
+
+    int m_bufferSize;
```
JoergAtGithub, line 57:
```suggestion
    std::size_t m_bufferSize;
```

src/engine/positionscratchcontroller.cpp
```diff
@@ -105,33 +150,18 @@ void PositionScratchController::process(double currentSamplePos,
         return;
     }
 
-    // The latency or time difference between process calls.
-    const double dt = static_cast<double>(bufferSize) / m_pMainSampleRate->get() / 2;
+    if (static_cast<int>(bufferSize) != m_bufferSize) {
```
JoergAtGithub, line 153:
```suggestion
    if (bufferSize != m_bufferSize) {
```

ronso0:
Thanks, correct. I resolved those conflicts like 10 times already, my main fix onto 2.5, my performance branch onto main :zany_face: 
Should be okay now, finally.

daschuer:
I had also a merge branch in progress and squashed both and pushed. so we can close this.

PR #14401, opened 2025-02-24
(fix) History: don't allow joining with locked previous playlist

### LINKED ISSUE #14399: Join playlist with previous below option works on locked playlists
### Bug Description

Add some items to a playlist and lock it. Then create another playlist above it, right click, and choose "Join with playlist below". The option completes even though the playlist below it was locked and should not be editable.

### Version

2.5

### OS

Arch Linux

ronso0:
Fixes #14399


daschuer (approved):
LGTM, Thank you.

PR #14402, opened 2025-02-24
(fix) track menu: reset `eject` after moving track file to trash

ronso0:
I was wondering why eject remained pressed after moving a track file to thrash, and I suspected the S4 mapping because I just noticed it with that.

A simple reset does the trick.


ronso0:
Btw I don't understand `ControlPushButton::TRIGGER`.
What is that supposed to do? I expected auto-reset to 0, no matter if connecting to `valueChanged()` or `valueChangeRequest()`.

ronso0:
Ah okay, it's used in WPushButton, implicitly for left-click it seems.

FWIW I don't want to mess with ControlObject behaviour here, just fix the misuse of eject.

daschuer (approved):
LGTM, Thank you.

PR #14341, opened 2025-02-14
Add softfix.yml

daschuer:
This adds a softfix comment feature to squash a branch. Currently it squashes all commits to one. Maybe we can also add a mode to squash fixup commits.

Find the manual here: 
https://github.com/daschuer/softfix

With this in place we can disable squash merges again and are still able to squash from remote.

PR #14405, opened 2025-02-24
CI Hardening

Holzhaus:
Let's improve the security of our CI workflows. See commit messages for details.


Swiftb0y (approved):
LGTM. Thank leaving the others a chance to comment too

.github/workflows/build.yml
```diff
@@ -354,11 +354,11 @@ jobs:
         # No need to do the PPA build for both Ubuntu versions
         if: matrix.name == 'Ubuntu 22.04'
         run: |
-          if [[ "${{ github.ref }}" == "refs/heads/main" ]] && [[ "${{ github.repository }}" == "mixxxdj/mixxx" ]]; then
+          if [[ "${GITHUB_REF}" == "refs/heads/main" ]] && [[ "${GITHUB_REPOSITORY}" == "mixxxdj/mixxx" ]]; then
```
acolombier, line 357: I'm quite interest with the rational here in term of security hardening? My understanding is that `${{ github.ref }}` will be interpolated by Github before it executes the actions, whereas `${GITHUB_REPOSITORY}` is an envvar that could be corrupted by previous steps/actions if they were to overwrite it inside the the `GITHUB_ENV` file? Am I missing something here?

.pre-commit-config.yaml
```diff
@@ -129,6 +129,14 @@ repos:
     rev: 37c2513b1b8275a475a160ed2f5b044910335d5f # No release tag yet including #6 fix
     hooks:
       - id: qml_formatter
+  - repo: https://github.com/gitleaks/gitleaks
+    rev: v8.24.0
+    hooks:
+      - id: gitleaks
```
acolombier, line 135: What would be the case for running Gitleaks in the case of a desktop application like Mixxx? As a rule of thumbs, I tend to reduce as much as possible software I would run from remote sources (which is what effectively pre-commit are), so I'd be interest to hear the pros for introducing this as mandatory hooks?

.pre-commit-config.yaml
```diff
@@ -129,6 +129,14 @@ repos:
...
+  - repo: https://github.com/woodruffw/zizmor-pre-commit
```
acolombier, line 136: This repo has a very low popularity, so I'm really not keen to have this running on our machines without protection. Could we consider introducing it using commit sha instead of mutable tags?

acolombier (changes requested):

.github/workflows/build.yml
Holzhaus: Sure, env vars could be modified by previous steps if they are persisted (i.e., written to `$GITHUB_ENV`). However, we control the workflow, and an attacker cannot easily change it without write access to our repo, tricking some contributor into merging a malicious PR without proper review or taking over some third-party action we use.

But here the issue is that these GitHub Expressions are not subject to normal shell quoting rules. They are always inserted verbatim. If - for any reason - one of these expressions like `github.ref` or `github.repository` contain something like `x" == "x" ]];  then cat /etc/passwd; fi #`, then the entire line would evaluate to:
```
if [[ "x" == "x" ]];  then cat /etc/passwd; fi #" == "refs/heads/main" ]] && [[ "some-user/some-repo" == "mixxxdj/mixxx" ]]; then
```
This kind of code injection cannot happen with env vars. If the env var contained the aforementioned string, the shell would not allow breaking out of the quotes.

To be clear: I do not suspect these expressions (`github.ref` or `github.repository`) to be actually easy to manipulate for an attacker. However, some expression are (e.g., `github.event.pull_request.title`).

.pre-commit-config.yaml
Holzhaus: Some time ago I wrote a helper script for the Launchpad-to-GitHub-Issues migration. I could have accidently committed the API credentials to the repo for example. If something like that is detected before committing, the creds are not actually exposed and leaked.

.pre-commit-config.yaml
Holzhaus: This is a mirror specifically for pre-commit. I suppose most people who use that tool actually starred the main repo that contains the actual linter source: https://github.com/woodruffw/zizmor
That one has 2K stars.

acolombier: Sure, but the code that you would run on your machine still comes from that repo with a very low amount of community surveillance. If you want to keep it as it, could we please make it an optional hook as I'm personally not keen with running this on my machine?

.pre-commit-config.yaml
acolombier: Hmm fair enough. I'm still ensure of the net worth value, especially that we do have consensus to reduce resources footprint as part of #14297. Adding more hooks means more bandwidth and CPU time (both CI and on contributor machines) and I'm wondering if this justify these rares usecase. 
Please don't get me wrong, there is a lot of value in adding hooks as it means potentially one less push (and thus CI trigger), so I'm not again pre-commit hooks like the CMake one or the clazy one being introduced in #13642, but I feel like it is worth considering this. How about making it optional?

.github/workflows/build.yml
acolombier: Thanks for clarifying. I feel like the risk with tempering with remote CI actions (either by direct repo take over -unlikely in case of `actions/checkout` - or cache poisoning) feels like a higher risk than being able to inject code directly inside the Github action engine (apart from loose one as you mentioned, such as `github.event.pull_request.title` or even `github.event.comment.body`)

Holzhaus: The risk impact is vastly different. Even if you somehow manipulate the env variable, you cannot achieve command injection as you could with the expressions.

.pre-commit-config.yaml
Holzhaus: Sure, just run `export SKIP=gitleaks` to disable it.

acolombier: this will disable it for other project on which `gitleaks` is relevant. Could we please adopt an opt-in policy instead, like our XML hook?

.github/workflows/build.yml
acolombier: I agree with you in theory, in practice tho, this could allow an attacker to deploy a rogue `.deb` to our PPA, which is IMO as bad. Anyway, this comment was only curiosity on the considered attack vector, no change request here!

.pre-commit-config.yaml
Holzhaus: Just to make clear: It's mirror of the tool by the same author, packaged as a python package to make it easier to sort up as a pre commit hook. 

We can add our own custom local hook that references the original repo (from the same github user), but that means every contributor needs to install the entire rust toolchain and rebuild the crate and all its dependencies every time you update the hook.

Can you elaborate a bit on your threat model?

Are you worried that the author is malicious? The author is the same author as the actual tool.

Or are you worried some attacker manages to sneak malicious code into that repo and nobody notices?

I can ask him if he has 2FA set up if this eases your mind (I strongly suspect he has).

For the record, if we are that cautious about projects without low community oversight, then we probably shouldn't give *write* permissions on the repo and its pull requests to Actions that have basically zero community oversight (and also 0 stars on GitHub, cf. 
https://github.com/mixxxdj/mixxx/pull/14341).

acolombier: > Are you worried that the author is malicious? The author is the same author as the actual tool.

> Or are you worried some attacker manages to sneak malicious code into that repo and nobody notices?

Both - as general rule, I tend not to trust anybody by default. Furthermore, as we can see from [the commit history](https://github.com/woodruffw/zizmor-pre-commit/commits/main/), this is not just about trusting one person, but every maintainers who may have write access on this repo at some point

> For the record, if we are that cautious about projects without low community oversight, then we probably shouldn't give write permissions on the repo and its pull requests to Actions that have basically zero community oversight (and also 0 stars on GitHub, cf.
https://github.com/mixxxdj/mixxx/pull/14341).

This repo is owned by a single maintainer, who I (and the Mixxx organisation) trust. In this case, Daniel has full admin access anyway so the case doesn't really stand, but it's worth detailing why I'm not considering any threat here, because we also have the e.V protection, which I assume would mean that Daniel would face the law in case he decided to do something rogue against the organisation. Of course, we could expect legal actions if somebody from gitleaks was to cause some harm, but this will likely be much harder and longer.
This trust is vital between core member, as I do trust you for example and trust you wouldn't do anything wrong [here](https://github.com/mixxxdj/mixxx/blob/a1db057fea8f3e61e422b16c3bff1c6e3e0a39d7/.github/workflows/pre-commit.yml#L18) (even with less visibility btw)

Holzhaus: Don't get me wrong, I didn't want to imply that Daniel shouldn't be trusted. But the point still stands that if we are worried about a a silent repo takeover by an attacker (e. g. because someone stole credentials) which is not immediately noticed due to lack of community oversight then this also affects core members like Daniel or myself.

I'm not saying that this isn't a possibility, but we have a lot of attack surface in that regard already. And we need to weigh the risk of adding this pre-commit hook against the risk of not having it and then doing insecure stuff on our CI. And I think the reduction of the probability of the latter (i.e. the risk of someone finding and exploiting a security fla in our CI) outweighs the risk associated with adding the hook.

acolombier: The point I'm trying to make here, is that in case of Daniel's GH action fork (or your Docker image), the attack vector dimension is known and defined. Sure, both of you could still be compromised, and this will likely allow an attacker to take over the whole Mixxx organisation anyway.
In case of this repo, the dimension is both unknown and undefined: we don’t know how many people have write access to it (likes to be ~4 looking at commits history, but it could be more... or less) and we have no idea about security policy that the owner has enforce to other contributor (the unknown). Furthermore, due to the lack of visibility of the repo, we don't know how this is going to involve in the future, meaning that accepting the risk now, could be an exponential risk in a few weeks, months, years... who knows. (same than liblzma for instance)

Just to illustrate my point here with Daniel's or your repo+Docker image:
- The write access is restricted to only you both (thus only your security is at stake). (the known)
- These repos are solely used for Mixxx purpose so the security policy is implied to be the same as Mixxx organisation (the "trust" part, e.g I'm pretty sure you guys wouldn't let some unrelated to Mixxx contributors or if you would, you would also have a solid trust to them)

> we have a lot of attack surface in that regard already

IMO, this makes it an argument against enlarging that surface. 

> the risk of someone finding and exploiting a security fla in our CI

I'm not too concerned on our CI, I'm mostly concerned on the security of our contributors (and mine selfishly). It means that every time I'll make a commit, I'll have to trust that hooks hasn't been tempered with and "pray" nobody wants me harm.

As per my change request, I would appreciate make this hooks manual so anybody who wants to enable them can do so. As a tradeoff, I'll settle for SHA pinning

Holzhaus: > I'm not too concerned on our CI, I'm mostly concerned on the security of our contributors (and mine selfishly)

The distinction is fuzzy tbh. When a malicious action commits to our repo and you pull it, the can execute stuff on your machine (for example via a manipulated `CMakeLists.txt`).

I agree with you that pinning commit hashes is desirable. I did it as part of this PR, both for pre-commit and GH Actions.

acolombier (approved):
I feel like pinning also Github official's `actions/*` was maybe one level to high, but appreciate you looked into the other one, less looked at. Thanks

PR #14412, opened 2025-02-26
Add M-Vave SMK-25 II piano keyboard mapping

SamWhited:
This patch adds support for the M-Vave SMK-25 II keyboard. I confirm that I have signed the CLA.

The corresponding manual PR is here: https://github.com/mixxxdj/manual/pull/728

### DIFF
--- a/res/controllers/MVave-SMK-25-II-scripts.js
+++ b/res/controllers/MVave-SMK-25-II-scripts.js
@@ -206,42 +206,43 @@ var SMK25II;
         switch (data[3]) {
         case 0x60: {
             const knob = SMK25II.controller.activeDeck.gainKnob;
+            knob.input(undefined, undefined, data[5], undefined, undefined);
             knob.inSetParameter(knob.inValueScale(data[5]));
             break;
         }
         case 0x61: {
             const knob = SMK25II.controller.activeDeck.highKnob;
-            knob.inSetParameter(knob.inValueScale(data[5]));
+            knob.input(undefined, undefined, data[5], undefined, undefined);
             break;
         }
         case 0x62: {
             const knob = SMK25II.controller.activeDeck.midKnob;
-            knob.inSetParameter(knob.inValueScale(data[5]));
+            knob.input(undefined, undefined, data[5], undefined, undefined);
             break;
         }
         case 0x63: {
             const knob = SMK25II.controller.activeDeck.lowKnob;
-            knob.inSetParameter(knob.inValueScale(data[5]));
+            knob.input(undefined, undefined, data[5], undefined, undefined);
             break;
         }
         case 0x64: {
             const knob = SMK25II.controller.activeDeck.effectKnob;
-            knob.inSetParameter(knob.inValueScale(data[5]));
+            knob.input(undefined, undefined, data[5], undefined, undefined);
             break;
         }
         case 0x65: {
             const knob = SMK25II.controller.xfadeKnob;
-            knob.inSetParameter(knob.inValueScale(data[5]));
+            knob.input(undefined, undefined, data[5], undefined, undefined);
             break;
         }
         case 0x66: {
             const knob = SMK25II.controller.headGainKnob;
-            knob.inSetParameter(knob.inValueScale(data[5]));
+            knob.input(undefined, undefined, data[5], undefined, undefined);
             break;
         }
         case 0x67: {
             const knob = SMK25II.controller.gainKnob;
-            knob.inSetParameter(knob.inValueScale(data[5]));
+            knob.input(undefined, undefined, data[5], undefined, undefined);
             break;
         }
         }
--- a/res/controllers/MVave-SMK-25-II-scripts.js
+++ b/res/controllers/MVave-SMK-25-II-scripts.js
@@ -138,12 +138,7 @@ var SMK25II;
         }
     }
 
-    /**
-     *
-     * @param _id The controller ID
-     * @param _debugging Whether we've started in debugging mode.
-     */
-    SMK25II.init = function(_id, _debugging) {
+    SMK25II.init = function() {
         SMK25II.controller = new Controller();
     };
     SMK25II.shutdown = function() {
@@ -152,7 +147,7 @@ var SMK25II;
     const MMCHeader = [0xF0, 0x35, 0x59];
     /**
      *
-     * @param data
+     * @param data the full sysex message data.
      */
     function decodeButton(data) {
         // Hotcues
@@ -250,8 +245,8 @@ var SMK25II;
     }
     /**
      *
-     * @param data
-     * @param length
+     * @param data the sysex message data
+     * @param length the length of the data (no longer used)
      */
     function incomingData(data, length) {
         if (length < 6) {
--- a/res/controllers/MVave-SMK-25-II-scripts.js
+++ b/res/controllers/MVave-SMK-25-II-scripts.js
@@ -116,15 +116,19 @@ var SMK25II;
             this.deckLeftButton = new components.Button({
                 group: "[Channel1]",
                 midi: [0x90, 0x2E], // << Channel Left Button
-                inToggle: function() {
-                    SMK25II.controller.activeDeck.setCurrentDeck("[Channel1]");
+                input: function(_channel, _control, value, _status, _group) {
+                    if (value === 0x00) {
+                        SMK25II.controller.activeDeck.setCurrentDeck("[Channel1]");
+                    }
                 },
             });
             this.deckRightButton = new components.Button({
                 group: "[Channel2]",
                 midi: [0x90, 0x2F], // >> Channel Right button
-                inToggle: function() {
-                    SMK25II.controller.activeDeck.setCurrentDeck("[Channel2]");
+                input: function(_channel, _control, value, _status, _group) {
+                    if (value === 0x00) {
+                        SMK25II.controller.activeDeck.setCurrentDeck("[Channel2]");
+                    }
                 },
             });
             // Drum Pads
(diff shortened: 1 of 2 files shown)


ronso0:
Thanks for this mapping!

Controller mappings are added to the current stable branch, unless they require features of newer versions.
Can you please rebase 2.5?

ronso0:
The touch strips aren't mapped, yet, right?

In case you need inspiration, there are other controllers that have this kind of strips
https://github.com/search?q=repo%3Amixxxdj%2Fmixxx+%22strip%22+path%3A%2F%5Eres%5C%2Fcontrollers%5C%2F%2F&type=code

SamWhited:
>  Can you please rebase 2.5?

Oops, done (and for the other mapping I just submitted), thank you!

> The touch strips aren't mapped, yet, right?

They are, the touch strips are mapped to tempo for the right one (this one is a bit useless as it resets to 0 when you let go), and volume for the right one. I'll take a look at the other controllers you mentioned though, and am happy to move them around if you think they should be different.

There's also the option of having all the encoders do different things on each layer, but I thought the EQ was the most important thing and wanted to be able to access it at all times, so I just mapped both layers the same way. I'm open to suggestions if you think it makes more sense to get other stuff on a rotary encoder though.

SamWhited:
I pushed up a few changes recommended by eslint. It's unclear to me why some of the things it complained about in the other mapping it's not complaining about with this one, and some of the things in CI on the other mapping aren't being complained about if I run it locally. Maybe the same version of eslint isn't being pinned and downloaded by pre-commit? Anyways, unsure if this will run cleanly or not when CI gets run next.

ronso0:
> They are, the touch strips are mapped to ...

Okay, great, so ignore my comment.

> and am happy to move them around if you think they should be different.

I don't own that keyboard so I rely on your experience entirely.
Not sure if you're aware that it's now (2.5) possible to add controller settings, that allows to implement tweaks and alternative mappings / behaviour in general.

SamWhited:
>  Not sure if you're aware that it's now (2.5) possible to add controller settings, that allows to implement tweaks and alternative mappings / behaviour in general.

I am, thanks! I'm new to Mixxx (2.5 is the first version I've used) so I didn't realize this was a new feature. I've used it pretty heavily in a mapping for another device I have and am a big fan, well done with that! If we think of other things that it could be in the future we can always add some settings. For now though I think it works pretty well as-is.

SamWhited:
Per @Swiftb0y's comment [here](https://github.com/mixxxdj/mixxx/pull/14411#issuecomment-2688597226) I added the hardware IDs back to the XML file and documented that they're not unique in a comment.

res/controllers/MVave-SMK-25-II-scripts.js
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    SMK25II.init = function() {
+        SMK25II.controller = new Controller();
+    };
+    SMK25II.shutdown = function() {
+        SMK25II.controller.shutdown();
+    };
+
+    const MMCHeader = [0xF0, 0x35, 0x59];
```
Swiftb0y, line 148: as far as I can tell, this is just sysex (MMC messages start with `0xF0, 0x7F`), so you may want to change the name

res/controllers/MVave-SMK-25-II-scripts.js
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     *
+     * @param data the full sysex message data.
+     */
+    const decodeButton = function(data) {
+        // Hotcues
+        if (data[4] !== undefined && data[4] >= 0x00 && data[4] <= 0x07) {
+            SMK25II.controller.activeDeck.hotcues[data[4]].inToggle();
```
Swiftb0y, line 157: the standard way would be to map `input`, not `inToggle` directly.

Swiftb0y: Also, does it really not send a release message?

res/controllers/MVave-SMK-25-II-scripts.js
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     *
+     * @param data the full sysex message data.
+     */
+    const decodePots = function(data) {
+        switch (data[3]) {
+        case 0x60: {
+            const knob = SMK25II.controller.activeDeck.gainKnob;
+            knob.inSetParameter(knob.inValueScale(data[5]));
```
Swiftb0y, line 209: also wouldn't recommend this. both functions are supposed to be private. Just dispatch to `input` and pass `undefined` to the non-applicable parameters.

res/controllers/MVave-SMK-25-II-scripts.js
```diff
@@ -0,0 +1,277 @@
...
+        }
+        for (let n = 0; n < MMCHeader.length; n++) {
+            if (data[n] !== MMCHeader[n]) {
+                console.log("unknown sysex packet");
+                return;
+            }
+        }
+        if (data[length - 1] !== 0xF7) {
+            console.log("sysex packet missing trailer");
+            return;
```
Swiftb0y, line 267: I actually wouldn't rely on length. Its a relic from when the devs thought JS was like C. Just take the `data.length` as that should always be correct.

Swiftb0y (commented):
couple thoughts

res/controllers/MVave-SMK-25-II-scripts.js
SamWhited: Oops, yup, fixed!

res/controllers/MVave-SMK-25-II-scripts.js
SamWhited: Fixed; thanks!

res/controllers/MVave-SMK-25-II-scripts.js
SamWhited: Done; thank you!

res/controllers/MVave-SMK-25-II-scripts.js
SamWhited: It does send a release message, this is probably toggling on both now that I look at it again, I'll swap it to "input" and check the value.

SamWhited: Actually, that's not true, they appear to send the release message immediately even if you're still pressing it. Oh well, a quirk of the firmware I suppose. I'll leave a comment.

SamWhited: If I change this to "input" it will break the deck left and deck right buttons because I overwrote inToggle for those to handle swapping the active deck. I thought this is what I did in the other mapping too, but they stop working if I call input. Should I be overriding and calling something else?

SamWhited: Actually, that makes perfect sense, inToggle only worked in the other plugin because it was a power-window button. Fixed!

SamWhited:
Everything looks green here! I went ahead and rebased the manual on the 2.5 branch and applied a patch that it wanted to optimize the SVGs. Manual page is now passing too (with the exception of linkcheck on another package, but that's just standard internet flakiness because linuxmusicians.com is returning a 502 at the moment).

Swiftb0y (commented):
last nit. LGTM otherwise. Thank you.

res/controllers/MVave-SMK-25-II-scripts.js
```diff
@@ -0,0 +1,291 @@
...
+    /**
+     *
+     * @param data the full sysex message data.
+     */
+    const decodePots = function(data) {
+        switch (data[3]) {
+        case 0x60: {
+            const knob = SMK25II.controller.activeDeck.gainKnob;
+            knob.input(undefined, undefined, data[5], undefined, undefined);
+            knob.inSetParameter(knob.inValueScale(data[5]));
```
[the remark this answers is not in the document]
SamWhited: oops, looks like a mistake when I was updating these to use input; good catch! Fixed.

SamWhited:
Just came back to double check CI, looks like there's another problem with the pre-commit stuff, but it doesn't appear to be related to this patch. Maybe someone could restart it?

Swiftb0y (approved):
LGTM. Thank you. Lets finish the manual now.

res/controllers/MVave-SMK-25-II-scripts.js
```diff
@@ -0,0 +1,290 @@
...
+            // Hotcues
+            this.hotcues = [];
+            for (let i = 0; i < 8; i++) {
+                const hotcueButton = new components.HotcueButton({
+                    number: i + 1,
+                    group: "[Channel1]",
+                    midi: [0x90, i],
+                    type: components.Button.prototype.types.push,
+                });
+                this.hotcues[i] = hotcueButton;
```
Swiftb0y, line 49: Small thing I noticed. why the `type: push`? Also: can you influence the color of the pads by the midi value? It would be great if the color of the pad matched the color of the hotcue.

SamWhited: This is the default so it's not strictly needed, just being explicit.
Unfortunately the color can't be set, this is a very cheap controller :)

Swiftb0y: So how does it work then? Because the pads are RGB, right? Does the firmware do all the mode switching? Have you tried to send different velocity values to the pads and see if that changes anything?

SamWhited: Yes, it's all handled by the firmware. The color depends on which layer you're on, and you can turn some of them on or off depending on the layer but you can't affect the color

Swiftb0y: Okay thats good to know. Maybe add a comment for that in the manual and the code so people aren't tempted to file issues for it (or waste their time trying to implement it).

Swiftb0y: Its funny to see how cheap controllers just get it wrong all the time. I guarantee it would've been cheaper not to build this functionality into the controller and just let the host handle the lighting, but manufacturers just choose to get it wrong.

SamWhited: Done; comment added; thanks!

SamWhited:
Green again! Pre-commit is the aforementioned issue and the recent commit just adds a comment so it shouldn't affect anything.

SamWhited:
I see that the change-log for 2.5.1 is being constructed with a date in the near future; is it possible that we could sneak this one in so both the SMK controllers are supported? Anything else missing for merge?

Swiftb0y:
not really, just giving ronso0 a chance to resolve their manual review. Also don't worry about the changelog and its date, we update it frequently and afaict don't plan on shipping it right now.

PR #14413, opened 2025-02-26
Use std::shared_ptr in  controller settings to fix memory leak

daschuer:
There are remaining objects not deleted in `controller_mapping_settings_test.cpp`

I have fixed it by using std::shared_ptr from the very beginning by std::make_shared. In Mixxx itself the objects are later managed by a std::shared_ptr anyway. 
In addition I have added two drive by fixes regarding move semantic and uninitialized variables.

### DIFF
--- a/src/controllers/legacycontrollersettings.h
+++ b/src/controllers/legacycontrollersettings.h
@@ -176,7 +176,8 @@ class LegacyControllerBooleanSetting
             bool defaultValue)
             : AbstractLegacyControllerSetting(element),
               m_savedValue(currentValue),
-              m_defaultValue(defaultValue) {
+              m_defaultValue(defaultValue),
+              m_editedValue(currentValue) {
     }
 
     bool parseValue(const QString& in) {
@@ -421,7 +422,8 @@ class LegacyControllerEnumSetting
             : AbstractLegacyControllerSetting(element),
               m_options(options),
               m_savedValue(currentValue),
-              m_defaultValue(defaultValue) {
+              m_defaultValue(defaultValue),
+              m_editedValue(currentValue) {
     }
 
     virtual QWidget* buildInputWidget(QWidget* parent) override;
--- a/src/controllers/legacycontrollersettings.h
+++ b/src/controllers/legacycontrollersettings.h
@@ -165,8 +165,8 @@ class LegacyControllerBooleanSetting
         emit valueReset();
     }
 
-    static AbstractLegacyControllerSetting* createFrom(const QDomElement& element) {
-        return new LegacyControllerBooleanSetting(element);
+    static std::shared_ptr<LegacyControllerBooleanSetting> createFrom(const QDomElement& element) {
+        return std::make_shared<LegacyControllerBooleanSetting>(element);
     }
     static bool match(const QDomElement& element);
 
@@ -281,8 +281,8 @@ class LegacyControllerNumberSetting
                 valid_range(m_minValue, m_maxValue, m_stepValue);
     }
 
-    static AbstractLegacyControllerSetting* createFrom(const QDomElement& element) {
-        return new LegacyControllerNumberSetting(element);
+    static std::shared_ptr<LegacyControllerNumberSetting> createFrom(const QDomElement& element) {
+        return std::make_shared<LegacyControllerNumberSetting>(element);
     }
     static bool match(const QDomElement& element);
 
@@ -352,8 +352,8 @@ class LegacyControllerRealSetting : public LegacyControllerNumberSetting<double,
         }
     }
 
-    static AbstractLegacyControllerSetting* createFrom(const QDomElement& element) {
-        return new LegacyControllerRealSetting(element);
+    static std::shared_ptr<LegacyControllerRealSetting> createFrom(const QDomElement& element) {
+        return std::make_shared<LegacyControllerRealSetting>(element);
     }
 
     QWidget* buildInputWidget(QWidget* parent) override;
@@ -408,8 +408,8 @@ class LegacyControllerEnumSetting
                 static_cast<int>(m_editedValue) < m_options.size();
     }
 
-    static AbstractLegacyControllerSetting* createFrom(const QDomElement& element) {
-        return new LegacyControllerEnumSetting(element);
+    static std::shared_ptr<LegacyControllerEnumSetting> createFrom(const QDomElement& element) {
+        return std::make_shared<LegacyControllerEnumSetting>(element);
     }
     static inline bool match(const QDomElement& element);
 
--- a/src/controllers/legacycontrollersettingsfactory.h
+++ b/src/controllers/legacycontrollersettingsfactory.h
@@ -42,7 +42,7 @@ class LegacyControllerSettingBuilder {
     /// @param element The XML element to parse to build the new setting
     /// @return an instance if a a supported setting has been found, null
     /// otherwise
-    static AbstractLegacyControllerSetting* build(const QDomElement& element) {
+    static std::shared_ptr<AbstractLegacyControllerSetting> build(const QDomElement& element) {
         for (const auto& settingType : std::as_const(instance()->m_supportedSettings)) {
             if (settingType.matcher(element)) {
                 return settingType.builder(element);
@@ -54,8 +54,8 @@ class LegacyControllerSettingBuilder {
 
   private:
     struct SupportedSetting {
-        bool (*matcher)(const QDomElement&);
-        AbstractLegacyControllerSetting* (*builder)(const QDomElement&);
+        std::function<bool(const QDomElement&)> matcher;
+        std::function<std::shared_ptr<AbstractLegacyControllerSetting>(const QDomElement&)> builder;
     };
 
     LegacyControllerSettingBuilder();
--- a/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
+++ b/src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
@@ -91,12 +91,12 @@ QJSValue ControllerScriptEngineLegacy::wrapFunctionCode(
 }
 
 void ControllerScriptEngineLegacy::setScriptFiles(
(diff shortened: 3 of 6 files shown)


src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
```diff
@@ -91,12 +91,12 @@ QJSValue ControllerScriptEngineLegacy::wrapFunctionCode(
 }
 
 void ControllerScriptEngineLegacy::setScriptFiles(
-        const QList<LegacyControllerMapping::ScriptFileInfo>& scripts) {
+        QList<LegacyControllerMapping::ScriptFileInfo> scripts) {
```
Swiftb0y, line 94: I think the actually generated a clazy warning because QList is too large?

src/controllers/legacycontrollersettings.h
```diff
@@ -165,8 +165,8 @@ class LegacyControllerBooleanSetting
         emit valueReset();
     }
 
-    static AbstractLegacyControllerSetting* createFrom(const QDomElement& element) {
-        return new LegacyControllerBooleanSetting(element);
+    static std::shared_ptr<LegacyControllerBooleanSetting> createFrom(const QDomElement& element) {
+        return std::make_shared<LegacyControllerBooleanSetting>(element);
```
Swiftb0y, line 169: slight nitpick: creating a `shared_ptr` from a `unique_ptr` is easy, the other way around is hard. So unless we are sure that all consumers will always need shared ownership, prefer that first... I'll leave it up to you to make that evaluation.

src/controllers/legacycontrollersettingsfactory.h
```diff
@@ -54,8 +54,8 @@ class LegacyControllerSettingBuilder {
 
   private:
     struct SupportedSetting {
-        bool (*matcher)(const QDomElement&);
-        AbstractLegacyControllerSetting* (*builder)(const QDomElement&);
+        std::function<bool(const QDomElement&)> matcher;
+        std::function<std::shared_ptr<AbstractLegacyControllerSetting>(const QDomElement&)> builder;
```
Swiftb0y, line 58: Can you explain your rationale here? This replaces an non-owning pointer type with an owning one and I don't see why that would need to be the case. The only real usecase where `std::function` is needed is when you need to pass a lambda with owning captures afaik.

Swiftb0y (commented):
good catch, couple thoughts.

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
daschuer: I don't think so, because it is kept inside the object. So a copy-by value is here the preferred signature.

src/controllers/legacycontrollersettings.h
daschuer: In Mixxx we always need a shared_ptr. We don't have shared pointers in tests. 
I like to benefit here form `std::make_shared` which allocates the object and the control structure for the shared pointer in one go.

src/controllers/legacycontrollersettingsfactory.h
daschuer: I have used `std::function` here, because it allows to take functions that are returning sub classes of AbstractLegacyControllerSetting. 
Before I had only luck using an ugly reinterpret_cast or an extra wrapper function. 
Is there a downside of using std::function compared to a wrapper for type conversion?

src/controllers/legacycontrollersettings.h
Swiftb0y: I know, questionable whether thats really needed and whether that assumption will really be true in the future too, but I won't insist.

src/controllers/legacycontrollersettingsfactory.h
Swiftb0y: Okay, with some careful double checking `std::function` does seem to be a good idea here (just be aware that it needs some RTTI). I just don't understand how the conversion mechanism works exactly. Do you have any insight here?

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
daschuer: clazy is not complaining. Due to the std::move() passing a const&  is no longer possible. I assume clazy also consideres that.

src/controllers/legacycontrollersettingsfactory.h
daschuer: I don't know. I have just poked around a bit and found out that this was working. Is ther actually a runtime overhead? 
If so, we should probably go for tha wrapper call with no overhead.

src/controllers/scripting/legacy/controllerscriptenginelegacy.cpp
Swiftb0y: Okay. Lets just merge and see then. Was just a speculation on my part, neither do I necessarily agree with clazy either.

src/controllers/legacycontrollersettingsfactory.h
Swiftb0y: Well, there is an unlikely conditional branch originating from the fact that a `std::function(nullptr)` results in an exception while the same for a function pointer would just segfault do to the `nullptr` deref. `std::function` also has a larger memory footprint. [godbolt](https://compiler-explorer.com/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyahUAUgBMAIUtXyKxqiIEh1ZpgDC6egFc2TEAtydwAZAiZsADk/ACNsUikecgAHdCViFyYvX39AlLSMoTCI6LY4hJ4kh2wnTJEiFlIibL8AoOraoXrGomKo2PjE%2Bwamltz2kd7w/rLByoBKe3QfUlROLksAZnDUXxwAajNNjxofJk7WeiPcMy0AQVu7n3STfYAxQ82AEX2lIkwQCBTudnEIjh5wkQIPNrkcbPdHpDfp8fukAF7YdA0CBvGGbeEPe5I7F4gmI4T7bDqGo4/Y0JTzQ4AdgJ%2BzZ%2B1I2CIKyYdKU0LhjzMTK%2Bjy4i3o3AArPwAlwdOR0NwPLZbL9lqtsIcLJs%2BOQiNpxYsANYgKVaQzcaT8Nhmi1yhVKrj8JQgC0G%2BXi8hwWAoDA4fDEMiUah0RisDjcPWCYRiCScGRyYTKNSaT3kfRBIwmNCq6z2bCOUEBCDuMYBJKhaalcoGVLpYvlusFYt9GtzAtFuqTJtVQs1YvdJptgYVYY9Xvj4fV0dSRZKDVrbjPcLAd4o/aQiAAKnmAvxQqJFKUG/RmOxuMFCKPRDp%2B7JN8p1NQtPpjOFrPZnO5pF59PvQoimKErSrKhqKsqeZWOqKxrNqmwWPwHo6PMixINgLA4Ak0KWlw1rkLa5rkA6/BOi6br6oaqHkKa0hSgAdAAbAAnDwzFaIxWhaFKFgWGxmy4ZsYHpmRlGeosPrIGg6BsMkDDxCGEAYLJ8kJLsxjADwmxaEkYZEPEroQDE4ExOEjQAJ5RvwpmsKQ5kAPIxLoA5WeQykcMI9lMPQlnpjgMQ%2BMAHgSPQrq8PwOBsBpkh%2BQQnK1AAbtgYUKlSNQ%2BPprmQoW4H0AQMSkBZXg4OBRCkAQtrheQSWkDEaTYF82BRSYeU5uJAhGMASgAGoENgADu9nJMwrkxqI4iSImY0pho4H6Ek2amFBhj5a6kCLOgyTFmFAC09mIYqNXlTga04R0xZuEwnjeK0BhViUs5JPWhRZDduRPS2mQjrMY7nd2E5vRWnYDv904PT9Bh/ADORA1DYMzLWPDzouCYgVwMrEeBTr7OoAAcjE7Yx0j7OprxafRWj0Tw%2BwQIGJCkPBSNIVR3p%2BjJcmMBQVBKezqloItWk6QIDD6aQhnGemNkWa5Ut2Y5zlOK57nMEQXk%2BeB/mBcF9Cha5kXResCqEPFzhJSl/BpagGXrHq2WSumeUFUVWCG0h5WVXqNV1SojXNcArWgO1NCdT1fWDcNcrRvIcaTbI00qLN6aZoYGm5tYtgrTEp0bVtmS7ftpFHQQJ3wPO/YXJd10w3dV3fYj%2BQNpkk7Pa2M4Q32XZdD2gOQ%2BXg6THXHZw80PdVAPbeI8jsGcBYaMYyREFcDj%2BOE8TpNruTlPU7ThD0/Bs/M%2B1EAoIQNA0IpY0xwmcfyDNaYKvQSCugtT%2BnzQRDmcNFGPyo6hlSwm0LrOnIKQJ%2BeRQFKDfh/L%2BFoIFUn/oAzI3AvT23nljbgXwCBn32L1Aa8Rl4EyJiTRa%2BxN5U32HtCw%2BwEonjxoQteJCyHU0oYfFCaEMJYWoCaM00h6LaVxloZimxmLMWkEyJkAimSMVwvhW0PAuKYxEsg%2BweRkJGlwgdBeok1HURqukVw0ggA)

Nevertheless. This is not hot code so lets not overthink it and go for the simplest solution which is `std::function`.

src/controllers/legacycontrollersettings.h
daschuer: I did a closer look. We require a std::share pointer because the AbstractLegacyControllerSetting is part of the mapping which is copied to the preferences (main thread) altered and and than moved back to the Controller thread. 

Copying all the control structures along with it feels wrong, however changing this is another refactoring, at least out of scope of this PR.

daschuer:
I think all topis are resolved. Merge?

Swiftb0y (approved):
Yup. Thank you.

PR #14415, opened 2025-02-26
Fix sync branches

Holzhaus:
Follow up to #13767


daschuer (approved):
Thank you. Let's see if this fixes the issue.

PR #14417, opened 2025-02-26
Mixxxbot in CI and build count reduction

### LINKED ISSUE #14297: Reduce the CI footprint

acolombier:
Attempt to address #14297


.github/workflows/labeler.yml
```diff
@@ -11,18 +11,24 @@ name: "Pull Request Labeler"
...
+      #   id: generate-token-key-pair
+      #   uses: actions/create-github-app-token@v1
+      #   with:
+      #     private-key: ${{ secrets.MIXXX_BOT_APP_PRIVATE_KEY }}
+      #     app-id: ${{ vars.MIXXX_BOT_APP_ID }}
+      - uses: actions/labeler@v5
         with:
-          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          # PAT setup with content:read and pull_request:write
+          repo-token: ${{ secrets.MIXXX_LABELLER_PAT }}
```
acolombier, line 33: PAT to be created for @mixxxbot and stored as secret

.github/workflows/softfix.yml
```diff
@@ -29,8 +25,22 @@ jobs:
...
+      #   with:
+      #     private-key: ${{ secrets.MIXXX_BOT_APP_PRIVATE_KEY }}
+      #     app-id: ${{ vars.MIXXX_BOT_APP_ID }}
       - name: Softfix
         if: steps.check-maintainer.outputs.result == 'true'
         uses: daschuer/softfix@v3
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # PAT setup with content:write and pull_request:write
+          repo-token: ${{ secrets.MIXXX_SOFTFIX_PAT }}
```
acolombier, line 46: PAT to be created for @mixxxbot and stored as secret

.github/workflows/pr-command.yml
```diff
@@ -0,0 +1,90 @@
...
+      #     app-id: ${{ vars.MIXXX_BOT_APP_ID }}
+      - name: Add or remove the skip-ci label
+        run: >-
+          gh pr edit
+            ${{ startsWith(github.event.comment.body, '/disable-ci') && '--add-label' || '--remove-label' }} 'skip-ci'
+            ${{ github.event.issue.pull_request }}
+        if: steps.check-maintainer.outputs.result == 'true'
+        env:
+          # PAT setup with pull_request:write
+          GH_TOKEN: ${{ secrets.MIXXX_PR_COMMAND_PAT }}
```
acolombier, line 90: PAT to be created for @mixxxbot and stored as secret

Swiftb0y: That would be `github_pat_anVzdCBhIHByYW5rIGJybw==` ;)

daschuer:
I have no access to the @mixxxbot user. I think @Swiftb0y has.

daschuer:
The PR changes a lot and is really hard to track. Can you spit it up into small PRs that are traceable. 

To be honest I am a bit in doubt of the complexity rectifies the benefit. #14297

Do we have an overview which measure saves which CI run? I am still convinced that the biggest benefit is to safe the double CI runs in my private repository vs. upstream.

Swiftb0y:
> I think @Swiftb0y has.

Antoine does too, I PM'd him the creds.

acolombier:
> The PR changes a lot and is really hard to track. Can you spit it up into small PRs that are traceable.

It's not that easy unfortunately. However, note that there is not as much changes as git highlights: the `build.yml` file was moved to `build-action.yml` with near to know change (except the input argument definition on top)

I can move the `pr-command` (later referenced as `3.`) and branch filter (`1.`)

> To be honest I am a bit in doubt of the complexity rectifies the benefit. #14297

We can always revert after sometime if we've seen no changes at all?

> Do we have an overview which measure saves which CI run? 

Sure:
1. Filtering the `push` rules to limit to production branches: (rational: even when doing experiment, contributor largely create draft PR. If one decides to make experiments without a PR, they can comment out the filtering line)
2. Label-based trigger (Rational: no need to build Mixxx 6 times per push and many cases, but we may still want to request further changes on some cases)
3. "Turn CI off" feature (rational: some PRs are quite near readiness or are meant to be ever be merged. This feature allow anybody to turn things off)

> I am still convinced that the biggest benefit is to safe the double CI runs in my private repository vs. upstream.

This is included in this PR, as part of the push branch filtering. However the net reduction of the Mixxx CI usage is zero, since this is counted on your own fork. Obviously, this is still very important ([20k for you](https://github.com/daschuer/mixxx/actions/metrics/usage) on the last month, [39k for me](https://github.com/acolombier/mixxx/actions/metrics/usage) :fearful: ). This should take down this number to a handful of minutes/months.

acolombier:
> The PR changes a lot and is really hard to track. Can you spit it up into small PRs that are traceable.

I have tried to help with the review here. Since this cannot be slitted into smaller PRs, I have added another intermediary commit which should help with reviewing the actually change to the `build.yml` (which become `build-action.yml`). Hopefully this should make that better

.github/workflows/build-action.yml
```diff
@@ -1,145 +1,90 @@
-# yaml-language-server: $schema=https://json.schemastore.org/github-workflow
-
 name: Build
 
 on:
-  push:
-  pull_request:
+  workflow_call:
+    inputs:
```
acolombier, line 5: The following argument are a carbon copy of the `strategy` parameters previously used

.github/workflows/build-action.yml
```diff
@@ -1,145 +1,90 @@
...
+  # macOS codesigning
+  MACOS_CODESIGN_CERTIFICATE_P12_BASE64: ${{ secrets.macos_codesign_certificate_p12_base64 }}
+  MACOS_CODESIGN_CERTIFICATE_PASSWORD: ${{ secrets.macos_codesign_certificate_password }}
+
 jobs:
   build:
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
```
acolombier: All of the following are now parameters of the build action

acolombier:
> > The PR changes a lot and is really hard to track. Can you spit it up into small PRs that are traceable.
> 
> I have tried to help with the review here. Since this cannot be slitted into smaller PRs, I have added another intermediary commit which should help with reviewing the actually change to the `build.yml` (which become `build-action.yml`). [The first one](https://github.com/mixxxdj/mixxx/pull/14417/commits/f6ce914d378cb73198e95cc5b91662190abf7692) move the file into a reusable action, [the second one](https://github.com/mixxxdj/mixxx/pull/14417/commits/7b9ab48be4186979cc9862b35040eab30156aab6) bring the necesseryy change to make it an action. Hopefully this should make that better

Friendly ping @daschuer :)

daschuer:
The new labels will clutter the list of PRs, because they are almost always applied. How about just introducing skip labels instead? 
skip-ci / skip-ci-linux / skip-ci-macos / skip-ci-windows

I am afraid the simple glob patterns are not enough to select which target should be build. Can we be more conservative and skip CI only when we know for sure. We need also consider the unit-tests for mappings. I am anyway in doubt that the automatic will sort out many runs. It is more likely a manual feature.

.github/workflows/build-checks.yml
```diff
@@ -5,12 +5,18 @@ name: build-checks
...
+      - synchronize
+      - labeled
+      - unlabeled
 
 permissions:
   contents: read #  to fetch code (actions/checkout)
 
 jobs:
   build-checks:
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip-ci') && (contains(github.event.pull_request.labels.*.name, 'core') || contains(github.event.pull_request.labels.*.name, 'platform/linux') || github.event_name == 'push') }}
```
daschuer, line 22: If one has disabled Linux CI because testing something with windows. Is it a good idea to skip this? We need probably an extra label to skip build-checks?

daschuer:
How is made sure that the labels are applied before the CI is run? 

The most CI runs are happening in the private repository anyway without a PR in the feature branch. How can we control that?

daschuer (changes requested):

acolombier:
> The new labels will clutter the list of PRs, because they are almost always applied

The only label that will often be added is `core`. `platform` labels will only be added upon specific platform changes (see `labeller.yml`). `skip-ci` is only meant to be on PR whilst they are being worked on.  

> How about just introducing skip labels instead?

From experience, managing an inverse list can be harder overtime for the `platform` labels. With the c`core` being problematic, I could replacing it with the following:

```
skip-ci:
  - changed-files:
    - all-globs-to-all-files: '!src/**/*'
    - all-globs-to-all-files: '!src/**/*'
``

This could also turn out hard to manage overtime, as [labeller rules](https://github.com/actions/labeler?tab=readme-ov-file#match-object) are limited to an "all-or-nothing" model. Furthermore, false positive with the `skip-ci` means that every pushes to this PR will add the label back, requiring a core ember to come and remove this back everytime, whereas the other don't apply since we only perform label additions.

I would recommend we start with the explicit `core` label and if this turns out to make too much noise, we can pivot to your suggestion.

> I am afraid the simple glob patterns are not enough to select which target should be build. Can we be more conservative and skip CI only when we know for sure. 

The added value of this solution is that you can still add labels yourself to trigger part of the CI that didn't trigger. Say we have a PR that only triggered the `macos` build, but it turns out this should have also triggered `windows`, you just have to tag `platform/windows`. We can also refines the rules as we go, as we find missing cases, but this will never be a blocker as reviewers can add a label.

> How is made sure that the labels are applied before the CI is run?

It doesn't matter - the CI will now only trigger on "sync" (new pushes to the PR) or "labelling" (when somebody or something  - the mixxxbot - adds a label to that PR). This means than on opening, the CI will not run, then labeller will run and add the right set of labels, which will trigger the CI. If a reviewer wants more confidence, they will also be able to add more labels and get ta new build kicking in. 



> The most CI runs are happening in the private repository anyway without a PR in the feature branch. How can we control that?

.github/workflows/build-checks.yml
acolombier: The `platform/windows` or `platform/macos` only applies for platform specific changes (e.g buildenv, packing) so these tests would be irrelevant to run since they only applied to the source code. If we decide to add [E2E/UI testing](https://github.com/mixxxdj/mixxx/pull/14360), we might want to reconsider this decision but since we are not ready to do so, it's probably best to be conservative. In case there was a time one wanted to run the `build-checks` anyway, they could always "abuse" the `core` label as an immediate workaround.

daschuer (commented):
Is this now obsolete?

acolombier:
The Mixxxbot part yes, the rest is still relevant as a measure to reduce the number of runs.

PR #14418, opened 2025-02-27
(fix) Track info dialog: fixed cover label (max) size

ronso0:
I noticed some flickering (quick dialog reisze) when switching tracks in the single-track dialog.
Looking closer, the current fixed size of the cover label is larger than the adjeacent three tag rows.

### Fix / TL;DR
* max cover **height** is now height of the three top rows
* max cover **width** is now width of the two rightmost columns
* the former won't change, and the latter is adjusted when the dialog is resized

Now all covers fit nicely, regardless their aspect ratio.
Noticeable especially with wide covers.
____
**Issue**: Originally, the cover art was granted 100x100 px in the dialog and each cover is scaled so it fits in, regardless its aspect ratio. The fixed height of 100 px was picked as it was <= the height of the 3 adjacent tag rows (with Qt5). But this depends on the OS theme / Qt theme picked by the OS, so it likely fits for many configurations, but not for all

Now WCoverArtLabel is scaled & resized on dialog resize (after show() and manual resize) and **only if necessary**.
Other dialogs using it (DlgTagFetcher) are not adjusted hence the current default size is still used.
(there are some glitches there, too, but I won't fix them here)

This prevents
* flickering of single-track dialog when switching tracks
parent cover widget is temporarily very large and the entire dialog is resized (twice: big, then adjusted to preferred size again)
* unintentional vertical stretching of grid rows
with my system's Qt theme this is just 3px and not really noticable, but with themes that haver smaller fonts, or other styles in general, it may look odd

It's a bit odd that the resize function needs to be in both DlgTrackInfo and ~Multi but I don't see an easy way to move this to a base class due to the UI dependency.

### DIFF
--- a/src/library/dlgtagfetcher.cpp
+++ b/src/library/dlgtagfetcher.cpp
@@ -255,7 +255,7 @@ void DlgTagFetcher::loadTrack(const TrackPointer& pTrack) {
                 &DlgTagFetcher::slotTrackChanged);
     }
 
-    m_pWFetchedCoverArtLabel->setCoverArt(CoverInfo{}, QPixmap{});
+    m_pWFetchedCoverArtLabel->setCoverInfoAndPixmap(CoverInfo{}, QPixmap{});
 
     m_coverCache.clear();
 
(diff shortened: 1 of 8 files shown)


daschuer (changes requested):
There seems to be a naming leftover. 
```
AutoUic: /home/daniel/workspace/mixxx2/src/library/dlgtrackinfo.ui: Warning: Tab-stop assignment: 'txtTrackName' is not a valid widget.
```

daschuer:
I can confirm the PR fixes the issue. The comment box however still changes the size. The content is only one line.

ronso0:
> There seems to be a naming leftover.
> 
> ```
> AutoUic: /home/daniel/workspace/mixxx2/src/library/dlgtrackinfo.ui: Warning: Tab-stop assignment: 'txtTrackName' is not a valid widget.
> ```

Fixed, was just the old `tabstop`.

ronso0:
> The comment box however still changes the size. The content is only one line.

Yeah, that's weird.
Maybe already fixed en-passant by 13818? I'll test next time I'll take a look at that PR.

daschuer (approved):
Still works. Thank you.

PR #14419, opened 2025-02-27
MIDI learn control menu: add more cue controls for samplers

ronso0:
Just noticed that not all cue controls are in the MIDI Learn control menu


daschuer (approved):
LGTM, Thank you.

PR #14423, opened 2025-02-28
(fix) Hotcue sort: collect only cues with valid hotcue index

ronso0:
Fixup for #13808

Also check for valid hotcue index when collecting hotcues.
These create issues:
* `Loop` and index -1 is the temporary loop
* and there may be orphaned `Hotcue`s with invalid index

____
I noticed that sorting hotcues would magically create new hotcues and loopcues, some even at the same pos / close to existing (valid) hotcues. Checking the database, I have about 3000 of such cues.
Seems they have accumulated over the past ~10 years :shrug:

### DIFF
--- a/src/track/track.cpp
+++ b/src/track/track.cpp
@@ -998,9 +998,14 @@ void Track::setHotcueIndicesSortedByPosition(HotcueSortMode sortMode) {
     indices.reserve(m_cuePoints.size());
     positions.reserve(m_cuePoints.size());
     for (const CuePointer& pCue : std::as_const(m_cuePoints)) {
-        if (pCue->getType() != mixxx::CueType::HotCue &&
-                pCue->getType() != mixxx::CueType::Loop &&
-                pCue->getType() != mixxx::CueType::Jump) {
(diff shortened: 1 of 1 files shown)


ronso0:
We may also add a cleanup function,  to be called for example when reading track cues?

Swiftb0y (approved):
Thank you.

ronso0:
I added the same check when setting the new indices :roll_eyes:

ronso0:
> We may also add a cleanup function, to be called for example when reading track cues?

Wdyt?
Or should this rather be part of https://github.com/mixxxdj/mixxx/blob/main/tools/mixxxdb_cleanup.sql, or a separate script?

Swiftb0y:
If you're talking about the orphaned hotcues, then it should be part of the cleanup script IMO.

ronso0:
My thought was that regular users are not aware of the script.
I'll file an issue.

ronso0:
Ready for merge, right?

Swiftb0y:
Sure, since nobody else is reviewing.

PR #14429, opened 2025-03-02
Continue after the maximum learning time is over

daschuer:
This is a band aid for a controller that repeatedly send the same message. I have implemented it for debugging the race condition in #14159 
In such a case the midi learn dialog was stuck. Now it continues after the progress bar is at maximum. 
Useful learning is however still not possible with such a hypothetical controller, because Mixxx will always pick the repeated message for learning. At least the GUI behaves as expected.


ronso0 (approved):
LGTM, thanks!

PR #14431, opened 2025-03-03
Tracks: add hotkey Shift+Del to purge and move track files to trash

ronso0:
Shortcut for quick cleanup. Faster than opening the menu with right-click, select Purge, hit Enter.
Shift + Del is also the "direct delete" shortcut in some file managers.

### TODO
- [ ] document hotkey in the manual https://manual.mixxx.org/2.5/en/chapters/appendix/keyboard_mapping_table#appendix-keyboard

### DIFF
--- a/src/widget/wtracktableview.cpp
+++ b/src/widget/wtracktableview.cpp
@@ -1112,6 +1112,8 @@ void WTrackTableView::keyPressEvent(QKeyEvent* event) {
     case kHideRemoveShortcutKey: {
         if (event->modifiers() == kHideRemoveShortcutModifier) {
             hideOrRemoveSelectedTracks();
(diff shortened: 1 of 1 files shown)


ronso0:
FWIW this has proven useful for cleaning up my library.

Right now my workflow is to preview tracks, set a rating >=3 (else I won't play them anyway..), then go to History, select all played tracks with rating <3, purge.

Actually, since I already added hotkeys for editing the comment of deck1/2, I'm thinking about deck Purge hotkeys.
This would allow to purge tracks right away (with the keyboard this would be faster than doing so via the track menu, less hazzle with the touchpad)

daschuer (approved):
LGTM, Thank you.

ronso0:
Cool, thanks for reviewing!

PR #14434, opened 2025-03-04
add 'LoadTrackFromPreviewDeck' control

ronso0:
to make the `LoadTrackFrom..` controls complete.

My use case is storing tracks in samplers while the track may not be visible in the library anymore but is still loaded in preview deck. Been missing this feature in my recent sessions with the Traktor S4 Mk3.

### DIFF
--- a/src/mixer/basetrackplayer.cpp
+++ b/src/mixer/basetrackplayer.cpp
@@ -191,6 +191,13 @@ BaseTrackPlayerImpl::BaseTrackPlayerImpl(
             &ControlObject::valueChanged,
             this,
             &BaseTrackPlayerImpl::slotLoadTrackFromSampler);
+    m_pLoadTrackFromPreviewDeck = std::make_unique<ControlObject>(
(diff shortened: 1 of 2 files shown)


daschuer (approved):
LGTM, thank you. 

Can you also add this to the manual?

ronso0:
Sure, will add it.

I think this is something that can be backported to 2.6, right?

daschuer:
If you consider this as a bug fix.

ronso0:
I think this makes the Load/Clone controls feature complete.
(for 100% completion we'd also need CloneFromPreviewDeck, but let's do this only if there is a demand, which I doubt).
And it's low-/no-risk :shrug:

daschuer:
Ok, go ahead.

ronso0:
Cool, here it is https://github.com/mixxxdj/mixxx/commit/fc2b71882076c2ee4f49383aa773681f609255a1

PR #14435, opened 2025-03-04
[VinylControl] Reduce sticker drift

ArthurVimond:
# Goal
This PR aims at reducing the sticker drift, especially in relative mode.

# Approach
In relative mode, we keep track of the absolute vinyl position and compute the drift amount delta to progressively change the VC Rate speed via the existing `dDriftControl` property, until the drift is back to 0 (or close to 0 at least).
This `m_relativedDriftAmtMem` variable is reset in case of a needle drop, a file position change (by using cue points, loops, mouse click etc.), if it's playing in reverse, if passthrough is enabled or scratch position is enabled (via waveform scratch for instance).

# How to test
- Using Serato vinyls in Relative mode, try different scratching patterns and ensure the sticker drift is not increasing.

# Related issues
- https://github.com/mixxxdj/mixxx/issues/8944
- https://github.com/mixxxdj/mixxx/issues/8927

Feedback is very welcome!

### DIFF
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -41,8 +41,7 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
     m_passthroughEnabled = make_parented<PollingControlProxy>(group, "passthrough");
-    scratchPositionEnabled = make_parented<ControlProxy>(
-            group, QStringLiteral("scratch_position_enable"), this);
+    scratchPositionEnabled = make_parented<ControlProxy>(group, QStringLiteral("scratch_position_enable"), this);
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
     m_bIsEnabled = wantenabled->get() > 0.0;
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -40,10 +40,8 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
     signalenabled = new ControlProxy(
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
-    m_passthroughEnabled = make_parented<ControlProxy>(
-            group, QStringLiteral("passthrough"), this);
-    m_scratchPositionEnabled = make_parented<ControlProxy>(
-            group, QStringLiteral("scratch_position_enable"), this);
+    m_passthroughEnabled = PollingControlProxy(group, QStringLiteral("passthrough"));
+    m_scratchPositionEnabled = PollingControlProxy(group, QStringLiteral("scratch_position_enable"));
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
     m_bIsEnabled = wantenabled->get() > 0.0;
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -40,8 +40,9 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
     signalenabled = new ControlProxy(
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
-    m_passthroughEnabled = make_parented<ControlProxy>(group, "passthrough", this);
-    scratchPositionEnabled = make_parented<ControlProxy>(
+    m_passthroughEnabled = make_parented<ControlProxy>(
+            group, QStringLiteral("passthrough"), this);
+    m_scratchPositionEnabled = make_parented<ControlProxy>(
             group, QStringLiteral("scratch_position_enable"), this);
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -40,8 +40,9 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
     signalenabled = new ControlProxy(
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
-    m_passthroughEnabled = make_parented<PollingControlProxy>(group, "passthrough");
-    scratchPositionEnabled = make_parented<ControlProxy>(group, QStringLiteral("scratch_position_enable"), this);
+    m_passthroughEnabled = make_parented<ControlProxy>(group, "passthrough", this);
+    scratchPositionEnabled = make_parented<ControlProxy>(
+            group, QStringLiteral("scratch_position_enable"), this);
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
     m_bIsEnabled = wantenabled->get() > 0.0;
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -41,7 +41,8 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
     m_passthroughEnabled = make_parented<PollingControlProxy>(group, "passthrough");
-    scratchPositionEnabled = make_parented<ControlProxy>(group, QStringLiteral("scratch_position_enable"), this);
+    scratchPositionEnabled = make_parented<ControlProxy>(
+            group, QStringLiteral("scratch_position_enable"), this);
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
     m_bIsEnabled = wantenabled->get() > 0.0;
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -40,9 +40,8 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
     signalenabled = new ControlProxy(
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
-    passthroughEnabled = make_parented<ControlProxy>(group, QStringLiteral("passthrough"), this);
-    scratchPositionEnabled = make_parented<ControlProxy>(
-            group, QStringLiteral("scratch_position_enable"), this);
+    m_passthroughEnabled = make_parented<PollingControlProxy>(group, "passthrough");
+    scratchPositionEnabled = make_parented<ControlProxy>(group, QStringLiteral("scratch_position_enable"), this);
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
     m_bIsEnabled = wantenabled->get() > 0.0;
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -41,7 +41,8 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
     m_passthroughEnabled = PollingControlProxy(group, QStringLiteral("passthrough"));
-    m_scratchPositionEnabled = PollingControlProxy(group, QStringLiteral("scratch_position_enable"));
+    m_scratchPositionEnabled = PollingControlProxy(
+            group, QStringLiteral("scratch_position_enable"));
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
     m_bIsEnabled = wantenabled->get() > 0.0;
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -40,7 +40,7 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
     signalenabled = new ControlProxy(
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
-    passthroughEnabled(group, QStringLiteral("passthrough"));
+    passthroughEnabled = make_parented<ControlProxy>(group, QStringLiteral("passthrough"));
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
     m_bIsEnabled = wantenabled->get() > 0.0;
--- a/src/vinylcontrol/vinylcontrol.cpp
+++ b/src/vinylcontrol/vinylcontrol.cpp
@@ -40,7 +40,7 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
     signalenabled = new ControlProxy(
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
-    passthroughEnabled = new ControlProxy(group, "passthrough", this);
+    passthroughEnabled(group, QStringLiteral("passthrough"));
 
     //Enabled or not -- load from saved value in case vinyl control is restarting
     m_bIsEnabled = wantenabled->get() > 0.0;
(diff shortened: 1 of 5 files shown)


src/vinylcontrol/vinylcontrolxwax.cpp
```diff
@@ -650,6 +665,27 @@ void VinylControlXwax::analyzeSamples(CSAMPLE* pSamples, size_t nFrames) {
...
+            }
+        if (!m_relativedDriftAmtSet) {
+            m_relativedDriftAmtMem = m_dDriftAmt;
+            m_relativedDriftAmtSet = true;
+        }
+
+        m_deltaRelativeDriftAmount = m_dDriftAmt - m_relativedDriftAmtMem;
+    }
+}
+
```
ywwg, line 688: Let's rework this function to keep it focused: It should be a pure function that takes parameters or reads member variables and returns a value and does not otherwise update member variables.

1. Move the mode conditional out of this function. That should be the responsibility of the caller to call in the correct case.
2. delta file position should not be calculated here, that should be outside the function.
3. I think we can make this function simply return the new value of m_deltaRelativeDriftAmount and nothing else.
4. m_relativedDriftAmtSet is only used locally and is immediately unset if it's false... that logic is quite tortured. 

I think we can pare it down to mostly just the thresholds and conditionals, as well as memorizing the original drift amount

```suggestion
// returns the delta between the current drift amount and the last reset value of the drift amount.
double VinylControlXwax::calcDeltaRelativeDriftAmount(double deltaFilePos) {
     // Reset m_relativeDriftAmtMem in case of needle drop, file position change (hotcue, loop etc.),
        // when passthrough is enabled or is playing in reverse
        if (fabs(m_deltaRelativeDriftAmount) > 1.5 || fabs(deltaFilePos) > 0.03 ||  // TODO: thresholds to adjust probably
            passthroughEnabled->toBool() || reverseButton->toBool()) {
            m_relativedDriftAmtMem = m_dDriftAmt;
        }
        return m_dDriftAmt - m_relativedDriftAmtMem;
    }
}

```

src/vinylcontrol/vinylcontrolxwax.cpp
```diff
@@ -466,6 +470,8 @@ void VinylControlXwax::analyzeSamples(CSAMPLE* pSamples, size_t nFrames) {
             //save the absolute amount of drift for when we need to estimate vinyl position
             m_dDriftAmt = m_dVinylPosition - filePosition;
 
+            checkDeltaRelativeDriftAmount(filePosition);
```
ywwg, line 473:
```suggestion
     if (m_iVCMode == MIXXX_VCMODE_RELATIVE) {
             m_deltaRelativeDriftAmount = calcDeltaRelativeDriftAmount(filePosition - m_dOldFilePos);
     }
```

ywwg (changes requested):
I know the vinylcontrol code is quite bad and functions have a lot of side effects, but let's try to make new functions as pure as possible as we improve the code.  I am not 100% sure if my suggestion will work but I think it is along the right lines

src/vinylcontrol/vinylcontrol.cpp
```diff
@@ -40,6 +40,7 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
     signalenabled = new ControlProxy(
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
+    passthroughEnabled = new ControlProxy(group, "passthrough", this);
```
daschuer, line 43: This can become: 
```suggestion
    passthroughEnabled = make_parented<ControlProxy>(group, "passthrough", this);
```

or even better: 
PollingControlProxy m_passthroughEnabled;

because it has no connection 

(Sorry, you have touched old code that is implements our recent style guide we like to introduce for changes.

ArthurVimond: @daschuer With `PollingControlProxy`, this will be instantiated as the following, right?
```
passthroughEnabled(group, QStringLiteral("passthrough"));
```

ywwg:
I'll try to test this tonight.

JoergAtGithub:
Welcome at Mixxx!

ArthurVimond:
Hi @JoergAtGithub, I signed the Mixxx Contributor Agreement.

Swiftb0y:
I have just built this branch and I unfortunately can't really confirm that this results in any sticker drift improvements. I haven't double checked whether the code touched in this PR is really being executed though.

Swiftb0y:
> It doesn't sound like your comment is meant to block merging, is that correct?

No not at all. I just wanted to say that I personally wasn't able to confirm any improvement and was hoping on my detailed instructions on why that may be the case.

ArthurVimond:
Hi @ywwg, just to be sure, is there any needed action on my side? Would it be helpful in validating the PR if I record a comparaison video like before and after those commits, where we can see the sticker drift appearing (in the before part) and compensating (in the after part)? Thanks.

src/vinylcontrol/vinylcontrol.cpp
ronso0: Yeah, let's update to `parented_ptr`, switch to `PollingControlProxy` and add `m_p`/`m_` prefix when all functional changes have been implemented and this PR got approved.
Wdyt?

ArthurVimond: Hi @ronso0!
Got it, sorry, I completely missed that part. I will do it very soon.

ronso0: No hurry..

ywwg:
hi, thanks for the ping.  Let me do one more round of reviewing and then we'll merge in. @jclsn are you interested in testing these changes?

jclsn:
Sure

jclsn:
So I just had a look at the code. Seems like some logic to estimate and correct the drift. I just don't understand why this is happening in the first place. The time between two LFSR states is 1000Hz/23bits = 43.5ms, which should be below human perception. Apart from that the whole pitch detection is done in the pitch.h using an alpha-beta filter. Even if the pitch should drift from the timecode, which could be 43ms max, this should not get fixed somewhere else than there. Instead of adding some logic to correct a drift, I would advocate to remove that part in Mixxx that causes it entirely.

ywwg:
I wonder if the correct thing to do is work in main instead of 2.5 and burn down all of the "correction" logic in the mixxx code and start over.  Remove all the smoothing, all the pitch compensation, etc. Make it worse before we make it better :). @ArthurVimond what's your preference here? I think if this patch makes your experience better I'm ok to merge it but I think jclsn is right that we need to attack the cause and not treat the symptoms.

jclsn:
I just tried it and I haven't noticed any sticker drift, but I haven't noticed one in the master branch either. I think the scratching behavior is still sluggish and doesn't feel very direct. I am aware that this patch is not intending to correct this, but I think the root cause of the sticker drift is in this smoothing filter somewhere. You are basically filtering the output of a filter with another filter. This just makes the system less reliable, adds more delay and complexity. 

In xwax all of this works well and it feels like true vinyl. I would just try to default back to this. I am planning to rework pitch detection at some point by using a Kalman filter instead of an alpha beta filter. A Kalman filter would dynamically adjust the currently fixed alpha and beta, which would let the filter correct itself. Kalman filters are actually the gold standard here. 

I still need to complete the whole Traktor MK2 stuff anyway, which might still take some time. So if you feel like this improves something, feel free to merge it. Just know that if I should ever work on the pitch detection here, I would rip out all of this anyway :P

ArthurVimond:
@ywwg @jclsn I agree with both of you ;) I mean, this drift correction patch does improve my scratching experience, but I understand that the real problem may be deeper.

I tried to change the alpha and beta values you suggested @jclsn (and disabling my drift correction code), but it didn't solve the original sticker drift and it was definitely accumulating more than 43.5 ms after a few minutes of fast scratch.

On the short run, I would prefer to include this PR in a coming release as a "quick fix" then work on the foundation to completely remove the drift, hopefully.

ArthurVimond:
And most importantly, the first thing I did was to remove the pitch value smoothing as suggested by @ywwg. The drift correction was added in the next commits.

jclsn:
@ArthurVimond If you want to dive into how the pitch detection actually works, I recommend you to read [this Jupyter Notebook](https://github.com/rlabbe/Kalman-and-Bayesian-Filters-in-Python). The first chaper "The g-h-filter" describes how the current pitch detection works. 

What Mark is basically doing is doing is

```c

if (tc->primary != 0 || tc->secondary != 0)
    filter_input = 0;
else
   filter input = 1.0 / carrier_frequency / 4;
``` 

So it uses four values (quadrants) per cycle. Basically every 15 minutes on a clock, it registers pitch movement. This is quite clever because it easily gives you an even pitch with a simple filter implementation. I am not 100% sure, but I think there is a better way. In the book it says to never throw away any data, but this is exactly what the current pitch detection is doing. 

I would really like to improve this as well, but the reverse engineering of the timecode is really time-consuming and summer is around the corner.

Swiftb0y:
Since I can reproduce sticker drift reliably (with and without this PR) could it be that its related to the type of timecode signal being used (and thus to xwax (or at least how we use xwax))? I'm using the serato scratch live vinyls.

ArthurVimond:
FYI, this drift correction will only be effective when we receive the vinyl timecode position. If we are only fast scratching without never letting the vinyl play normally, the drift will not be compensated (since we can't know the absolute vinyl position, and compute the offset).

jclsn:
@Swiftb0y I don't think so. The timecode is fixed on the vinyl. It could be that either [this](https://github.com/mixxxdj/mixxx/blob/0b72f45f0a21c4448a568c196678febbddeefe6e/lib/xwax/timecoder.c#L638) or the pitch smoothing is causing it.

ywwg:
> Since I can reproduce sticker drift reliably (with and without this PR) could it be that its related to the type of timecode signal being used (and thus to xwax (or at least how we use xwax))? I'm using the serato scratch live vinyls.

no, we are doing something wrong in Mixxx that is causing excessive drift. It will take some time to figure out what that is.

@ArthurVimond do you have a preference of merging this PR with this compensation code vs removing it and just keeping the change to remove the smoothing?

ArthurVimond:
@ywwg I would prefer to include both.

ywwg:
@jclsn is it ok if we merge this as-is and then keep working on reducing the actual drift?

ywwg (approved):

ArthurVimond:
@ywwg I still have some small refactoring to do (use parented_ptr, PollingControlProxy and m_ prefix) and it will be ready for merging. Will do that this evening. Thanks.

src/vinylcontrol/vinylcontrol.cpp
ArthurVimond: @ywwg I made the change. The PR is ready.

src/vinylcontrol/vinylcontrol.cpp
```diff
@@ -40,6 +40,9 @@ VinylControl::VinylControl(UserSettingsPointer pConfig, const QString& group)
     signalenabled = new ControlProxy(
             group, "vinylcontrol_signal_enabled", this);
     reverseButton = new ControlProxy(group, "reverse", this);
+    m_passthroughEnabled = make_parented<ControlProxy>(group, "passthrough", this);
```
Swiftb0y, line 43: NIT:
```suggestion
    m_passthroughEnabled = make_parented<ControlProxy>(group, QStringLiteral("passthrough"), this);
```

src/vinylcontrol/vinylcontrol.h
```diff
@@ -65,6 +67,8 @@ class VinylControl : public QObject {
     ControlProxy* signalenabled;
     // When the user has pressed the "reverse" button.
     ControlProxy* reverseButton;
+    parented_ptr<ControlProxy> m_passthroughEnabled;
+    parented_ptr<ControlProxy> scratchPositionEnabled;
```
Swiftb0y, line 71: NIT:
```suggestion
    parented_ptr<ControlProxy> m_scratchPositionEnabled;
```

src/vinylcontrol/vinylcontrolxwax.cpp
```diff
@@ -516,10 +524,19 @@ void VinylControlXwax::analyzeSamples(CSAMPLE* pSamples, size_t nFrames) {
...
-                    fabs(m_dDriftAmt) > 0.1 && fabs(m_dDriftAmt) < 5.0) {
+                    fabs(m_dDriftAmt) > 0.03 && fabs(m_dDriftAmt) < 5.0) {
                 dDriftControl = m_dDriftAmt * .01;
             } else {
-                dDriftControl = 0.0;
+                // Apply relative drift control
+                if (m_iVCMode == MIXXX_VCMODE_RELATIVE) {
+                    if (fabs(m_deltaRelativeDriftAmount) > 0.006 &&
+                            fabs(m_deltaRelativeDriftAmount) < 1.0 &&
+                            fabs(m_deltaFilePos) < 0.03) {
```
Swiftb0y, line 534: NIT (same above and below)
```suggestion
                // Apply relative drift control
                if (m_iVCMode == MIXXX_VCMODE_RELATIVE) {
                    if (std::fabs(m_deltaRelativeDriftAmount) > 0.006 &&
                            std::fabs(m_deltaRelativeDriftAmount) < 1.0 &&
                            std::fabs(m_deltaFilePos) < 0.03) {
```

src/vinylcontrol/vinylcontrolxwax.h
```diff
@@ -135,6 +138,8 @@ class VinylControlXwax : public VinylControl {
     // The drift between the vinyl position and the file position from the most
     // recent run of analyzeSamples.
     double m_dDriftAmt;
+    double m_relativedDriftAmtMem;
```
Swiftb0y, line 141: `Mem` is not a common postfix. What does it mean?

ArthurVimond: @Swiftb0y In my head, `Mem` means `Memory`. It is to keep track of the drift amount at some point and only update it with some specific condition (needle drop, passthrough mode etc.). I didn't find a better name.

src/vinylcontrol/vinylcontrol.cpp
ArthurVimond: @Swiftb0y For my understanding, how the `QStringLiteral` is important compared to a simple String? Thanks.

src/vinylcontrol/vinylcontrolxwax.h
ArthurVimond: What about `m_initialRelativeDriftAmt`?

src/vinylcontrol/vinylcontrol.cpp
Swiftb0y: QString's are utf-16 internally, while (plain) string literals are plain ascii (or depending on compiler details sometimes implicitly utf-8) encoded. If you just write a string literal, Qt's implicit conversions do an expensive conversion from ascii to utf-16 at runtime and then store that string in heap memory. If you wrap the literal in a `QStringLiteral`, the compiler does the conversion at compile time. This saves the runtime conversion and a heap allocation. Moreover, the compiler can then also deduplicate different instances of that string (calling plain `QString("some str")` results in a different heap allocation on every instance, while multiple `QStringLiteral("some str")` essentially just fall back to being a pointer and the underlying string being pointed to is always the same). 

I'm not saying this makes a performance difference in this case, but it does when it adds up and I want to encourage using `QStringLiteral` wherever possible.

Swiftb0y: Also look at the Qt documentation in this case: https://doc.qt.io/qt-6/qstring.html#QStringLiteral

ArthurVimond:
@ronso0
I had the following errors:
- `static_cast from 'PollingControlProxy *' to 'const QObject *', which are not related by inheritance, is not allowed`
- `invalid ‘static_cast’ from type ‘PollingControlProxy*’ to type ‘const QObject*’`
cf. https://github.com/ArthurVimond/mixxx/actions/runs/14252530792
Not sure what it means and how to solve them.

src/vinylcontrol/vinylcontrol.cpp
ArthurVimond: Thanks for the detailed explanation!

src/vinylcontrol/vinylcontrolxwax.h
Swiftb0y: sure...

Swiftb0y:
You're likely trying to construct a `parented_ptr<PollingControlProxy>`. Since `PollingControlProxy` is not a QObject (and thus a little lighter) wrapping it in a `parented_ptr` is neither required (nor supported, which is why you get the issues (`parented_ptr<T>` assumes `T` inherits from `QObject`, since thats not the case for `PollingControlProxy`, you get the compile error)).
The solutions here is to simply add `PollingControlProxy` members directly (since those are designed to do that).

daschuer (approved):
LGTM, Thank you.

PR #14445, opened 2025-03-05
Check for deck undefineds in Traktor S4 Mk2 HID

grahamhall123:
Fixes first interaction with loop size encoder not registering on the number display.

### DIFF
--- a/res/controllers/Traktor-Kontrol-S4-MK2-hid-scripts.js
+++ b/res/controllers/Traktor-Kontrol-S4-MK2-hid-scripts.js
@@ -1640,6 +1640,9 @@ TraktorS4MK2.onLoopEnabledChanged = function(value, group, key) {
 
 TraktorS4MK2.onLoopSizeChanged = function(value, group, key) {
     var deck = TraktorS4MK2.resolveDeckIfActive(group);
+    if (deck === undefined) {
+        return;
+    }
     //deal with single digit values
     if (value.toString().length === 1) {
         TraktorS4MK2.sendLoopSizeMessage(deck, "", value, false, false);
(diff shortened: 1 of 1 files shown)


JoergAtGithub:
Welcome at Mixxx!

grahamhall123:
Done 🙂

JoergAtGithub:
grafik
unzip it, and apply it using:
`git apply pre-commit.patch`

grahamhall123:
Ah excellent - thanks for your help!

SamWhited (changes requested):
It may be better to handle this at lines 1657 and 1767 below the `resolveDeckIfActive` call. This way you're catching it for any possible uses of deck in the call stack below that, even if those uses change in the future (note that I'm not a reviewer with +1 powers though, this is just where I'd handle this).

acolombier:
Jumping on this late - I think @SamWhited's feedback makes sense, are you happy to look into this @grahamhall123?

Otherwise, I'm happy merging this as is, as this seems to fix an annoying bug.

JoergAtGithub:
The code looks good. But the target for mapping PRs should be the `2.5` branch not `main`. Could you please retarget/rebase this PR as described here: https://github.com/mixxxdj/mixxx/wiki/Using-Git#targeting-another-base-branch

grahamhall123 (commented):
Switched target branch to 2.5

JoergAtGithub:
You only changed the target in the GitHub PR. But this is only the half of the process. You also need to rebase the git branch as described in https://github.com/mixxxdj/mixxx/wiki/Using-Git#targeting-another-base-branch

grahamhall123:
Reset branch to 2.5 and re-commited the changes. Hopefully should be up to date now!

JoergAtGithub:
LGTM! Thank you!

JoergAtGithub (approved):

PR #14447, opened 2025-03-06
Fix pitch bend buttons and rate/volume knob on Hercules DJ Console Mk1

SamWhited:
Previously the pitch bend buttons were mapped the same way as the jog wheels, however this is identical to the jog wheels right below them and felt a bit useless. This patch maps the buttons to their original purpose, bumping the pitch up or down, and adds a setting that allows reverting to the previous behavior for backwards compatibility.

I only have one of these temporarily and won't be able to do long term maintenance on this controller, but if there are any other changes that need to be made or tested I have one for the next day or two.

### DIFF
--- a/res/controllers/Hercules DJ Console Mk1.hid.xml
+++ b/res/controllers/Hercules DJ Console Mk1.hid.xml
@@ -7,6 +7,23 @@
         <forums>http://www.mixxx.org/forums/viewtopic.php?f=7&amp;t=4081</forums>
         <manual>hercules_dj_console_mk1</manual>
     </info>
+    <settings>
+        <group label="Backwards Compatibility">
+            <row orientation="vertical">
+                <option
+                    variable="pitchBendAsJogWheel"
+                    type="boolean"
+                    default="false"
+                    label="Use the pitch bend buttons as jog wheels">
+                    <description>
+                      This will change the pitch bend buttons to behave like the
+                      jog wheels, scratching when the track is paused or seeking
+                      when the track is playing.
+                    </description>
+                </option>
+            </row>
+        </group>
+    </settings>
     <controller id="Hercules DJ Console Mk1 HID">
         <scriptfiles>
             <file filename="Hercules-DJ-Console-Mk1-hid-scripts.js" functionprefix="HerculesMk1Hid"/>
--- a/res/controllers/Hercules DJ Console Mk1.hid.xml
+++ b/res/controllers/Hercules DJ Console Mk1.hid.xml
@@ -21,6 +21,16 @@
                       when the track is playing.
                     </description>
                 </option>
+                <option
+                    variable="gainAsTempo"
+                    type="boolean"
+                    default="false"
+                    label="Use the gain encoder for setting the tempo">
+                    <description>
+                      This will change the gain encoder to modify the tempo
+                      instead.
+                    </description>
+                </option>
             </row>
         </group>
     </settings>
--- a/res/controllers/Hercules-DJ-Console-Mk1-hid-scripts.js
+++ b/res/controllers/Hercules-DJ-Console-Mk1-hid-scripts.js
@@ -84,20 +84,26 @@ HerculesMk1Hid.init = function() {
     //
 
     c.capture("pitchbend_down", "all", function(g, e, v) {
-        if (engine.getValue(g, "play") == 0) {
-            engine.setValue(g, "back", v > 0 ? 1 : 0);
-        }
-        else if (v > 0) {
-            engine.setValue(g, "jog", -3);
+        if (engine.getSetting("pitchBendAsJogWheel")) {
+            if (!engine.getValue(g, "play")) {
+                engine.setValue(g, "back", v > 0 ? 1 : 0);
+            } else if (v > 0) {
+                engine.setValue(g, "jog", -3);
+            }
+        } else {
+            engine.setParameter(g, "pitch_down", v > 0);
         }
     });
 
     c.capture("pitchbend_up", "all", function(g, e, v) {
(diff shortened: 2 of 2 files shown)


res/controllers/Hercules-DJ-Console-Mk1-hid-scripts.js
```diff
@@ -84,20 +84,28 @@ HerculesMk1Hid.init = function() {
...
-            engine.setValue(g, "jog", -3);
+        if (engine.getSetting("pitchBendAsJogWheel")) {
+            if (engine.getValue(g, "play") == 0) {
+                engine.setValue(g, "back", v > 0 ? 1 : 0);
+            }
+            else if (v > 0) {
+                engine.setValue(g, "jog", -3);
+            }
+        } else {
+            script.toggleControl(g, "pitch_down");
```
Swiftb0y, line 94: using `toggleControl` has the problem that in case a message gets lost, the assumed state of the controller and the device would become disconnected. Please do `engine.setParameter(g, "pitch_down", v > 0);` instead.

SamWhited: Oh right, that makes perfect sense, I was forgetting that this would obviously hit the "on" and "off" states as well and probably happen twice. Done!

res/controllers/Hercules DJ Console Mk1.hid.xml
```diff
@@ -7,6 +7,23 @@
         <forums>http://www.mixxx.org/forums/viewtopic.php?f=7&amp;t=4081</forums>
         <manual>hercules_dj_console_mk1</manual>
     </info>
+    <settings>
+        <group label="Backwards Compatibility">
+            <row orientation="vertical">
+                <option
+                    variable="pitchBendAsJogWheel"
```
Swiftb0y, line 14: please also document this setting (as well as the new behavior in 2.5.1) in the manual.

Swiftb0y (approved):
LGTM. Thank you.

res/controllers/Hercules DJ Console Mk1.hid.xml
SamWhited: Currently there is no manual page for this controller. I may or may not be able to find the time to make one since that's a much bigger change, but I'd like to do a few patches first since I only have this controller for ~12 more hours while I do a repair on it.

Swiftb0y:
Since you already have the controller, does it still work as described on 2.5? The last time someone changed this mapping and actually had access to the hardware was likely when the mapping got added in 2013...

SamWhited:
I have a few more small patches like this that I'd like to make; would it be better to do them all at once in this PR, or make separate small PRs for ease of review?

Thanks for taking a look at this so quickly, I only have the controller for a short amount of time, so this was great timing!

Swiftb0y:
Nah, just do it in one PR and split into appropriate commits. Consider using fixup commits for incorporating review changes.

SamWhited:
> Since you already have the controller, does it still work as described on 2.5? The last time someone changed this mapping and actually had access to the hardware was likely when the mapping got added in 2013...

As far as I can tell everything works as expected except the LEDs (which as far as I can tell from the discussion and the mapping have never worked correctly). However, there are a few quirks of the mapping like the one fixed in this PR.

SamWhited:
> Nah, just do it in one PR and split into appropriate commits

In that case, I've added another commit swapping the "volume" rotary encoder, which was previously mapped to the tempo, to be the gain instead. I know gain *technically* also isn't volume, which is how the control is labeled, but I suspect that's what it was originally used for since actual volume is normally a linear encoder, and I think this will best meet user expectations (also actual volume is currently on the slider, so no need for having two controls mapped to it).

SamWhited:
I think I'm done for now actually; there's a ton more that could be done to this controller (the "1", "2", and "3" buttons across the top don't make a lot of sense IMO, there are a few buttons in the middle that are mappable but appear unused, etc.), but this at least makes it somewhat usable and I don't know if it's really worth it to go to deep on a controller that most people probably aren't using.

Swiftb0y (approved):
honestly. LGTM for now. Anything else you wanna do to it?

SamWhited:
> Anything else you wanna do to it?

I may come back to it later if I get bored, but for now I think this at least makes it somewhat usable again and gets rid of the duplicate controls; good to merge.

Swiftb0y (changes requested):

res/controllers/Hercules-DJ-Console-Mk1-hid-scripts.js
```diff
@@ -84,20 +84,26 @@ HerculesMk1Hid.init = function() {
...
-        else if (v > 0) {
-            engine.setValue(g, "jog", -3);
+        if (engine.getSetting("pitchBendAsJogWheel")) {
+            if (!engine.getValue(g, "play")) {
+                engine.setValue(g, "back", v > 0 ? 1 : 0);
+            } else if (v > 0) {
+                engine.setValue(g, "jog", -3);
+            }
+        } else {
+            engine.setParameter(g, "pitch_down", v > 0);
```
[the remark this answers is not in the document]
SamWhited: Done! Assuming they actually follow the ECMAScript standard this should work either way, but also this is one of the reasons I deeply dislike JavaScript and all its variants and derivatives, so being explicit always sounds good t me.

Swiftb0y (approved):
Thank you.

SamWhited:
Anything left on this one while we're merging stuff? It has an approval and all checks are green. Thanks!

Swiftb0y:
Nope. Was just waiting to give others a chance to have a look in case they'd like to. Thanks for the ping.

PR #14448, opened 2025-03-06
Fix LED on M-Vave SMC-Mixer

SamWhited:
Apparently we somehow broke the LED code on the M-Vave SMC-Mixer during the review the other day and I didn't notice.

The `component.Pot` that we override has itself overridden the default connect method to remove connecting the output. This seems like it might be a bug itself that should be fixed, as pots may also have LED ring lights around them or what not to indicate their position, but it's unclear to me if this is deliberate or not so I filed this patch instead of a bug report against the midi components library. I can also fix it there if you think that's best.

### DIFF
--- a/res/controllers/MVave-SMC-Mixer-scripts.js
+++ b/res/controllers/MVave-SMC-Mixer-scripts.js
@@ -96,7 +96,7 @@ var SMCMixer;
             // anything less than the tolerance window, we consider them the
             // same. This way we're not constantly blinking the soft takeover
             // indicator because we didn't get the control matched up exactly.
-            this.toleranceWindow = 0.001;
+            this.toleranceWindow = 0.03;
         }
         input(_channel, _control, value, _status, _group) {
             const receivingFirstValue = this.hardwarePos === undefined;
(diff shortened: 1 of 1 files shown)


Swiftb0y:
So how exactly does this work here? I see the Pots are used for the faders, so I assume the LEDs you're referring to are the ones between each fader and pot?

SamWhited:
Yes, so before these were working but apparently we changed from an encoder to a pot or something and broke this and I didn't notice. We keep track of the value seen by input (ie. whent he hardware slider changes), and we keep track of the value seen by the connected output (ie. when the virtual slider changes in Mixxx) and if they're different we turn the LED on until they match up again.

This was part of the origijnal PR and was working fine, but apparently sometime during code review we broke it, I'm not sure exactly when.

Swiftb0y:
> and we keep track of the value seen by the connected output (ie. when the virtual slider changes in Mixxx) and if they're different we turn the LED on until they match up again.

And thats implemented in firmware?

Swiftb0y:
Ah right, I didn't reread enough of the code just now.

Swiftb0y (approved):
LGTM. Thank you.

SamWhited:
Thanks! I would still be curious if this is a bug in the Pot code itself? I don't see why this would be the one component that doesn't have a call to connect?

Swiftb0y:
Good question. My assumption would be that there is no standard way a Pot would respond to feedback messages, so this may have been stubbed out like that to force the user to implement the behavior if they really need it.

Swiftb0y:
In this case, its very much satisfied.

PR #14449, opened 2025-03-06
Add Browse FileSytem ChildItems Limit

Eve00000:
This PR adds the possibility to define a max number of childitems that will be added in the Folder-TreeView.
Number can be set between 100 & 50.000. Standard not enabled.

Special for slow PC's or PC's connected to a slow external drive / nas, or users making a mistake causing Mixxx to 'hang"

-> usecase PR #14441 "Opening a directory with a large number of files in the sidebar freezes Mixxx "

### DIFF
--- a/src/library/browse/browsefeature.cpp
+++ b/src/library/browse/browsefeature.cpp
@@ -18,8 +18,12 @@
 #include "widget/wlibrarysidebar.h"
 #include "widget/wlibrarytextbrowser.h"
 
-bool s_browseLimitChildItemsEnabled;
-int s_browseLimitChildItemsNumber;
+const ConfigKey kChildItemLimitNumberKey =
+        ConfigKey(QStringLiteral("[Library]"),
+                QStringLiteral("BrowseFilesystemLimitChildItemsNumber"));
+const ConfigKey kChildItemsLimitEnabledKey =
+        ConfigKey(QStringLiteral("[Library]"),
+                QStringLiteral("BrowseFilesystemLimitChildItemsEnabled"));
 
 namespace {
 
@@ -150,13 +154,6 @@ BrowseFeature::BrowseFeature(
 
     // initialize the model
     m_pSidebarModel->setRootItem(std::move(pRootItem));
-    // Vars to limit the created childitems in tree,
-    // for slow pc's or external sources withslow connection
-    s_browseLimitChildItemsEnabled = pConfig->getValue<bool>(
-            ConfigKey("[Library]",
-                    "BrowseFilesystemLimitChildItemsEnabled"));
-    s_browseLimitChildItemsNumber = pConfig->getValue<int>(
-            ConfigKey("[Library]", "BrowseFilesystemLimitChildItemsNumber"));
 }
 
 BrowseFeature::~BrowseFeature() {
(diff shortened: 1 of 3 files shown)


src/preferences/dialog/dlgpreflibrary.cpp
```diff
@@ -340,6 +341,11 @@ void DlgPrefLibrary::slotUpdate() {
         break;
     }
 
+    checkBox_browsefilesystemlimitchilditemsenabled->setChecked(m_pConfig->getValue(
+            ConfigKey("[Library]", "BrowseFilesystemLimitChildItemsEnabled"), false));
+    spinbox_browsefilesystemlimitchilditemsnumber->setValue(m_pConfig->getValue<int>(
+            ConfigKey("[Library]", "BrowseFilesystemLimitChildItemsNumber")));
```
Swiftb0y, line 347: consider adding globals for the `ConfigKey`s

src/library/browse/browsefeature.cpp
```diff
@@ -469,8 +469,20 @@ std::vector<std::unique_ptr<TreeItem>> BrowseFeature::getChildDirectoryItems(
...
     // loop through all the item and construct the children
     foreach (QFileInfo one, all) {
+        if (m_pConfig->getValue(
+                    ConfigKey("[Library]",
+                            "BrowseFilesystemLimitChildItemsEnabled"),
+                    true)) {
+            if (count >= m_pConfig->getValue<int>(ConfigKey("[Library]",
+                                 "BrowseFilesystemLimitChildItemsNumber"))) {
+                // Limit -> Stop adding items
+                break;
```
Swiftb0y, line 483: I'm unsure about the performance of `getValue()`. Lets query it outside the loop and then compare to the local variable inside the loop. There is no advantage to querying in every iteration.

Also, feel free to replace the `foreach` with a standard ranged-ranged-for-loop while you're at it.

src/library/browse/browsefeature.cpp
```diff
@@ -458,7 +458,7 @@ void BrowseFeature::onLazyChildExpandation(const QModelIndex& index) {
 std::vector<std::unique_ptr<TreeItem>> BrowseFeature::getChildDirectoryItems(
         const QString& path) const {
     std::vector<std::unique_ptr<TreeItem>> items;
-
+    // line to start CI
```
Swiftb0y, line 461: ?

src/preferences/dialog/dlgpreflibrary.cpp
Eve00000: OK, I will do that

src/library/browse/browsefeature.cpp
Eve00000: will do

src/library/browse/browsefeature.cpp
Eve00000: CI had html problems, to restart it when fixed I added this line

Swiftb0y: I see. well you can remove that again now I guess.

src/preferences/dialog/dlgpreflibrary.cpp
ronso0: btw new config item and controls should use snake_case

for example
`browse_child_item_limit_enabled` and
`browse_child_item_limit`
?

src/library/browse/browsefeature.cpp
```diff
@@ -18,6 +18,9 @@
 #include "widget/wlibrarysidebar.h"
 #include "widget/wlibrarytextbrowser.h"
 
+bool s_browseLimitChildItemsEnabled;
+int s_browseLimitChildItemsNumber;
```
Swiftb0y, line 22: This is not what I meant. I'm sorry I've been unclear with my previous suggestion. I was rather suggesting you make these two local variables and initialize them just before the loop. With `ConfigKey` globals I was suggesting to just create them as constants (and also employ `QStringLiteral`s):

```suggestion
const ConfigKey kChildItemLimitKey = ConfigKey(QStringLiteral("[Library]"), QStringLiteral("BrowseFilesystemLimitChildItemsNumber")));
const ConfigKey kChildItemsLimitEnabledKey = ConfigKey(QStringLiteral("[Library]"), QStringLiteral("BrowseFilesystemLimitChildItemsEnabled"))
```

In general, avoid writeable globals whenever possible, they make it easy to write race conditions and don't work well in an OOP environment.

src/library/browse/browsefeature.cpp
```diff
@@ -469,8 +478,18 @@ std::vector<std::unique_ptr<TreeItem>> BrowseFeature::getChildDirectoryItems(
     QFileInfoList all = dirAccess.info().toQDir().entryInfoList(
             QDir::Dirs | QDir::NoDotAndDotDot);
 
+    int count = 0;
+
     // loop through all the item and construct the children
-    foreach (QFileInfo one, all) {
+    // foreach (QFileInfo one, all) {
+
+    for (const QFileInfo& one : all) {
```
Swiftb0y, line 486: to fix the clazy warning:
```suggestion
    for (const QFileInfo& one : std::as_const(all)) {
```

src/library/browse/browsefeature.cpp
Eve00000: >globals

No problem, I thought it was a little bit odd, but you asked for globals ;-)

>What about enabling this by default?
... will be disappointed by Mixxx ... something like 100, ... "Load more" ... bottleneck ... 200

I can't  find any arguments against your idea. I guess this is how a coffee-project becomes a big dinnerparty project 😄 
-> default 100 -> popup continue/stop with reference to setting. we don't want users to be disappointed.
-> bottleneck= filesystem; chrono= opening/displaying a folder with 100 items takes the same time as opening a folder with 3200 items: +/- 59secs

OwKee

ronso0: > -> bottleneck= filesystem; chrono= opening/displaying a folder with 100 items takes the same time as opening a folder with 3200 items: +/- 59secs

Okay, just to be sure: you measured that with the Browse feature, right?

Eve00000: Yes, to simulate the issue: Browse via wired LAN to NAS. (local drive is faster)
Why?

ronso0: Just wanted to make sure there are no false assumptions due comparing Browse with Crates or something, that's all.

Eve00000:
I forgot this PR, Was there some else requested?

Issue #14450, opened 2025-03-07
Crossfader doesn't cut sound on Traktor Kontrol Z1 with scratching curve


ArthurVimond:
### Bug Description

# Steps to reproduce
- Plug the Traktor Kontrol Z1 to the computer
- Open Mixxx
- In the Controllers settings, Set the Traktor Kontrol Z1 HID mapping
- In the Mixer settings, set the crossfader curve to scratching (slider completely on the right)
- Load and play two songs on decks 1 and 2 respectively
- On the Traktor Kontrol Z1, put the physical crossfader completely on the left
- Notice the song on deck 2 can still be heard
- On the Traktor Kontrol Z1, put the physical crossfader completely on the right
- Notice the song on deck 1 can still be heard

# Cause
In theory, the crossfader value range is from 0 to 4095, but in practice, it never reaches the boundaries. On my unit, the minimum safe value is 36 (32 at the lowest) and the maximum safe value is 4083 (4087 at the highest).

# Solution
When setting the crossfader parameter value, we should apply those safe "margins" to ensure that the crossfader will always cut the sound.

ArthurVimond:
The corresponding PR is here: https://github.com/mixxxdj/mixxx/pull/14451

SamWhited:
I've seen this same problem on the Hercules DJ Console Mk1. The console appears to be sending `0xFF` in this case, but for some reason Mixxx doesn't fully bring the crossfader to the right and you have to bump it over a tiny bit more in the UI to actually cut the audio from the other channel completely out. It may be worth fixing this in a more general way in Mixxx itself.

PR #14451, opened 2025-03-07
fix(controllers): fix crossfader cut on Traktor Kontrol Z1

ArthurVimond:
# Problem
When the crossfader curve is set to scratching in the settings, the physical crossfader of the Traktor Kontrol Z1 doesn't cut the sound of the corresponding deck.

# Cause
In theory, the crossfader value range is from 0 to 4095, but in practice, it never reaches the boundaries. On my unit, the minimum safe value is 36 (32 at the lowest) and the maximum safe value is 4083 (4087 at the highest).

# Approach
When setting the crossfader parameter value, we should apply those safe "margins" to ensure that the crossfader will always cut the sound.

Update: We now get the crossfader calibration values using `controller.getFeatureReport()` and apply them to the value to set. We also apply extra safe "margins" (5 on a total range of 4096) to compensate the value returned by the hardware, which can fluctuate a little bit on some crossfader movement (it doesn't always reach the calibrated value everytime).

# Steps to reproduce
- Plug the Traktor Kontrol Z1 to the computer
- Open Mixxx
- In the Controllers settings, set the existing Traktor Kontrol Z1 HID mapping
- In the Mixer settings, set the crossfader curve to scratching (slider completely on the right)
- Load and play two songs on decks 1 and 2 respectively
- On the Traktor Kontrol Z1, put the physical crossfader completely on the left
- Notice the song on deck 2 can still be heard
- On the Traktor Kontrol Z1, put the physical crossfader completely on the right
- Notice the song on deck 1 can still be heard

# Linked issue
[https://github.com/mixxxdj/mixxx/issues/14450](https://github.com/mixxxdj/mixxx/issues/14450)

### DIFF
--- a/res/controllers/Traktor-Kontrol-Z1-scripts.js
+++ b/res/controllers/Traktor-Kontrol-Z1-scripts.js
@@ -16,11 +16,15 @@ class TraktorZ1Class {
         this.vuLeftConnection = {};
         this.vuRightConnection = {};
         this.vuMeterThresholds = {"vu-30": (1 / 7), "vu-15": (2 / 7), "vu-6": (3 / 7), "vu-3": (4 / 7), "vu0": (5 / 7), "vu3": (6 / 7), "vu6": (7 / 7)};
+
+        this.rawCalibration = {};
+        this.calibration = null;
     }
 
     init(_id) {
         this.id = _id;
 
+        this.calibrate();
         this.registerInputPackets();
         this.registerOutputPackets();
         this.readCurrentPosition();
@@ -116,6 +120,32 @@ class TraktorZ1Class {
         this.lightDeck(false);
     }
 
+    calibrate() {
+        this.rawCalibration.faders = new Uint8Array(0x20 * 3);
+        this.rawCalibration.faders.set(new Uint8Array(controller.getFeatureReport(0xD1)), 0x00);
+        this.rawCalibration.faders.set(new Uint8Array(controller.getFeatureReport(0xD2)), 0x20);
+        this.rawCalibration.faders.set(new Uint8Array(controller.getFeatureReport(0xD3)), 0x40);
+        this.calibration = this.parseRawCalibration();
+    }
+
+    parseRawCalibration() {
+        return {
+            crossfader: this.parseCrossfaderCalibration(0x3C),
+        };
+    }
+
+    parseCrossfaderCalibration(index) {
+        const data = this.rawCalibration.faders;
+        return {
+            min: this.parseUint16Le(data, index),
+            max: this.parseUint16Le(data, index+2),
+        };
+    }
+
+    parseUint16Le(data, index) {
+        return data[index] + (data[index+1]<<8);
+    }
+
     readCurrentPosition() {
         // Sync on-screen controls with controller knob positions
         const report0x01 = new Uint8Array(controller.getInputReport(0x01));
@@ -207,10 +237,13 @@ class TraktorZ1Class {
 
     parameterHandler(field) {
         let value = field.value / 4095;
-        // Crossfader value don't reach boundaries and need safe margins
-        // min: 36 - max: 4083
+        // Crossfader value don't reach boundaries and need to use calibration values
+        // Also apply extra safe margins
+        const safeMargins = 5;
         if (field.name === "crossfader") {
-            value = (field.value - 36) / 4047;
+            const min = this.calibration.crossfader.min;
+            const max = this.calibration.crossfader.max;
+            value = script.absoluteLin(field.value, 0, 1, min + safeMargins, max - safeMargins);
         }
         engine.setParameter(field.group, field.name, value);
     }
--- a/res/controllers/Traktor-Kontrol-Z1-scripts.js
+++ b/res/controllers/Traktor-Kontrol-Z1-scripts.js
@@ -5,6 +5,8 @@
 // Author: djantti
 //
 
+// eslint definitions
+/* global controller, HIDController, HIDPacket */
 class TraktorZ1Class {
     constructor() {
         this.controller = new HIDController();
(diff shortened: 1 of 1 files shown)


Eve00000:
Hi Arthur,
I 'm just wondering if a calibration-tool would not be better?
Maybe the crossfader-value-differences can be higher/smaller on other controllers (of the same model and other models).

ArthurVimond:
Hi Eve,
You are right, I thought about it and that's what Traktor do in the Kontrol Z1 settings, there is a calibration button for the sliders (volumes and crossfader).

I think @djantti developed the original HID mapping. Would it be possible for you (@djantti) to test the values received from the crossfader?

ronso0:
re calibration tool:
a first step could be spinboxes for the upper/lower end in the controller settings, by default using the hardware min/max values.

ArthurVimond:
@ronso0 If I understand correctly, it is actually possible to create per-controller settings GUI elements directly from the controller mapping JS file? I looked for that in the controllers code but didn't find clear examples. Is it related to the Component JS API? If it's possible, I will clearly do that!

ronso0:
Controller settings are available for MIDI and HID, so unrelated to components. One example mapping you could take a look at would be that for the Traktor S4 Mk3.
Options are created and grouped into layouts in xml and read in js.

ArthurVimond:
Oh yes! I was not aware of this! I definitely love Mixxx!!!

JoergAtGithub:
The Z1 stores calibration values for the end and center positions of all the faders and knobs in a device internal memory, accessed via HID FeatureReports. The current Z1 mapping does not read these values and therefore every Z1 will have slightly different fader behavior. This means that the hard-coded values you suggest are suitable for the current wear condition of your controller, but others may require different values.
Currently we've only one mapping, that reads these calibration values from the device and applies them. This is the mapping for the Traktor S2 Mk1.

djantti:
On my Z1 I get crossfader values from 4 to 4087, so there's definitely variation between devices. 😅

I did a fair bit of NI Traktor USB sniffing back in the day and only briefly experimented with reading the Z1 calibration data. But calibration data import with spinboxes for manual override would be a great modern solution for this!

ronso0:
@ArthurVimond Do you intend to finish this by implementing the calibration method used in the Traktor S2 Mk1 mapping?

ArthurVimond:
Hi @ronso0,
Sorry, yes, I do but didn't find the time to work on it yet.

ArthurVimond:
Hi @djantti!
If you have some time and still have your Kontrol Z1, would you mind testing this PR again, please? I now use the crossfader calibration values (min and max) to set the corresponding value in Mixxx.

@ronso0 If @djantti confirms it's working fine with his device, could you review it again, please?
Many thanks!!

djantti:
Sure thing! I'll test the PR tomorrow and will report back here. It's awesome that you're working on this feature again! :star_struck:

res/controllers/Traktor-Kontrol-Z1-scripts.js
djantti:
```suggestion
        this.registerInputScaler(InputReport0x01, "[Master]", "crossfader", 0x1B, 0xFFFF, this.crossfaderHandler.bind(this));
```
The crossfader is a special case now, so we could add a separate handler for it.

res/controllers/Traktor-Kontrol-Z1-scripts.js
```diff
@@ -16,11 +18,15 @@ class TraktorZ1Class {
         this.vuLeftConnection = {};
         this.vuRightConnection = {};
         this.vuMeterThresholds = {"vu-30": (1 / 7), "vu-15": (2 / 7), "vu-6": (3 / 7), "vu-3": (4 / 7), "vu0": (5 / 7), "vu3": (6 / 7), "vu6": (7 / 7)};
+
```
djantti, line 21:
```suggestion

// Calibration data
```
There can never be enough code comments. 😄

res/controllers/Traktor-Kontrol-Z1-scripts.js
djantti:
```suggestion
        this.registerInputScaler(InputReport0x01, "[Master]", "crossfader", 0x1B, 0xFFFF, this.crossfaderHandler.bind(this));
```
The crossfader is a special case now, so we could use a separate handler for it.

res/controllers/Traktor-Kontrol-Z1-scripts.js
```diff
@@ -206,7 +238,16 @@ class TraktorZ1Class {
     }
 
     parameterHandler(field) {
-        engine.setParameter(field.group, field.name, field.value / 4095);
+        let value = field.value / 4095;
+        // Crossfader value don't reach boundaries and need to use calibration values
+        // Also apply extra safe margins
+        const safeMargins = 5;
+        if (field.name === "crossfader") {
```
djantti: The if statement works fine, but we could also keep parameterHandler() as it originally is and add a separate crossfaderHandler() with all the new calibration stuff. This might lead to better code structure.

djantti (commented):
Testing done and the PR works great!

Looks like the crossfader on my Z1 was calibrated to 4 - 4079. I do see it go up 4087, but that's close enough for factory calibration. Or maybe I've just punched it to the sides too much. 😅

I also left some small code suggestions.

res/controllers/Traktor-Kontrol-Z1-scripts.js
ArthurVimond: I totally agree. I was not confident enough to make that change myself at first but it was actually easy. Thank you!

res/controllers/Traktor-Kontrol-Z1-scripts.js
ArthurVimond: Agreed too!

ArthurVimond:
Hi @ronso0, this PR is ready for review. Thanks!

ronso0:
@JoergAtGithub do you mind taking a look?
I'm not familiar with calbration stuff.

djantti (approved):
Nice work, the code is better structured now. I also did another test with my Z1 and the calibration is still functioning as expected. Thank you for improving this mapping. 🙏

JoergAtGithub (approved):
LGTM! Thank you!

JoergAtGithub:
@ArthurVimond @djantti Did you test this mapping change with Mixxx 2.5 or with a newer version of Mixxx?
The release check for 2.5.5 failed with this mapping.

djantti:
This was a while ago, but I'm pretty sure I tested it with 2.5 branch. And I've been testing 2.5 branch Flatpak builds ever since and haven't had any issues with my Z1.

PR #14456, opened 2025-03-09
Remove accidentally comitted "backup" file

SamWhited:
I'm assuming this was accidentally committed? If not, sorry for the noise.


acolombier:
I'm pretty sure we already had this file removed :D

It must have made its way back when we synched branches. Thanks for cleaning this up!

acolombier (approved):
I'm pretty sure we already had this file removed :D

It must have made its way back when we synched branches. Thanks for cleaning this up!

PR #14458, opened 2025-03-09
Make sure all LEDs on SMC-Mixer go off on shutdown

SamWhited:
I noticed a moment ago that all of the LEDs on the SMC-Mixer don't go off when you shutdown Mixxx. The problem appears to be that the `EQRack` class is neither a component nor a component container, so the `forEachComponents` call that shuts down everything cleanly skips over it and the buttons it contains never get turned off. Changing it to a `ComponentContainer` fixes this, but also triggers a bug (I think) in QJSEngine where `this` is not defined in the callback which also had to be worked around.

### DIFF
--- a/res/controllers/MVave-SMC-Mixer-scripts.js
+++ b/res/controllers/MVave-SMC-Mixer-scripts.js
@@ -130,8 +130,9 @@ var SMCMixer;
         }
     }
 
-    class EqRack {
+    class EqRack extends components.ComponentContainer {
         constructor(index) {
+            super({});
             const channel = mapIndexToChannel(index);
             this.knob = new Encoder({
                 group: `[Channel${channel}]`,
@@ -139,13 +140,20 @@ var SMCMixer;
                 inKey: "pregain",
             });
 
+            const _this = this;
             const btnInToggle = () => {
-                const knob = this.knob;
-                const origGroup = this.knob.group;
-                const origInKey = this.knob.inKey;
+                // "this" is undefined here due to a bug in QJSEngine, so we
+                // redefine it as "_this" above and close over it.
+                // In this context it refers to the EqRack.
+                const knob = _this.knob;
+                const origGroup = knob.group;
+                const origInKey = knob.inKey;
                 return function() {
+                    // "this" in this context refers to the button itself, and
+                    // will be defined after this returned callback is set on
+                    // the button as "inToggle" on each of the buttons below.
                     if (this.isLongPressed) {
-                        if (knob.inKey === this.inKey.replace("button_", "") || (this.inKey === "enabled" && this.knob.inKey === "super1")) {
+                        if (knob.inKey === this.inKey.replace("button_", "") || (this.inKey === "enabled" && knob.inKey === "super1")) {
                             knob.group = origGroup;
                             knob.inKey = origInKey;
                         } else {


res/controllers/MVave-SMC-Mixer-scripts.js
```diff
@@ -130,40 +130,42 @@ var SMCMixer;
         }
     }
 
-    class EqRack {
+    class EqRack extends components.ComponentContainer {
         constructor(index) {
+            super(undefined);
```
Swiftb0y, line 135:
```suggestion
            super({});
```

res/controllers/MVave-SMC-Mixer-scripts.js
```diff
@@ -130,40 +130,42 @@ var SMCMixer;
...
-                const origGroup = this.knob.group;
-                const origInKey = this.knob.inKey;
+                const knob = _this.knob;
+                const origGroup = _this.knob.group;
+                const origInKey = _this.knob.inKey;
                 return function() {
-                    if (this.isLongPressed) {
-                        if (knob.inKey === this.inKey.replace("button_", "") || (this.inKey === "enabled" && this.knob.inKey === "super1")) {
+                    if (_this.isLongPressed) {
+                        if (knob.inKey === _this.inKey.replace("button_", "") || (_this.inKey === "enabled" && _this.knob.inKey === "super1")) {
```
Swiftb0y, line 150: wait, shouldn't it have been `knob` instead of `this` in the first place?

SamWhited: oh yah, I guess it could have been. I forgot I redeclared it at the top (though also I have no idea why I did that, writing "this.knob" a few times isn't that much longer or harder). Fixed anyways. Thanks!

SamWhited:
Hmm, something else is still going wrong with this. It appears QJSEngine has some weird scope issues, I'm getting random errors now about stuff that definitely exists not existing. Working on it. One more commit incoming.

SamWhited:
My find/replace was a little too broad, I forgot the inner function has a different this. Have I complained yet about how much I hate javascript? *sigh*. Fixed and rebased, everything appears to be working now.

res/controllers/MVave-SMC-Mixer-scripts.js
```diff
@@ -130,22 +130,30 @@ var SMCMixer;
...
                 midi: [0xB0, 0x10 + index],
                 inKey: "pregain",
             });
 
+            const _this = this;
             const btnInToggle = () => {
-                const knob = this.knob;
-                const origGroup = this.knob.group;
-                const origInKey = this.knob.inKey;
+                // "this" is undefined here due to a bug in QJSEngine, so we
```
Swiftb0y, line 145: `this` also always refers to the EqRack because this is an arrow function. You'll need to declare it as an anonymous function (so `const btnInToggle = function() {`), so the `this` can be appropriately set when this function is getting called.

Swiftb0y: Then `this` should also work correctly (and always refer to the correct button instance instead of the `EqRack` instance).

SamWhited: You'd be right except for the bug in QJSEngine I mentioned, if I call super above, this stops being defined entirely and using "this" results in an error. It definitely shouldn't be this way, but it is.

Swiftb0y: Oh, no sorry for the bogus comment. I totally missed where we already return a function in the `btnInToggle` case. Yeah in that case the local alias is probably best.

Swiftb0y (approved):
LGTM. Thank you.

PR #14459, opened 2025-03-09
Add fractional beat_active style controls

### LINKED ISSUE #14391: control like beat_active for scaled tempo beats?
### Feature Description

which can be connected to, to flash pads with, to avoid timer drift by doing it manually

mxmilkiib:
Resolves #14391

### DIFF
--- a/src/engine/controls/clockcontrol.cpp
+++ b/src/engine/controls/clockcontrol.cpp
@@ -1,5 +1,7 @@
 #include "engine/controls/clockcontrol.h"
 
+#include <cmath>
+
 #include "control/controlobject.h"
 #include "control/controlproxy.h"
 #include "engine/controls/enginecontrol.h"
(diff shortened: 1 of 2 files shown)


acolombier:
I'm not sure we would want to create additional COs for this purpose, IMO this should be computed in JS, by the controller. If certain variables are missing to perform that computation, we could add such COs, but maintain a subjective list feels wrong here. The rational here is that the number of fractions could grow infinity (e.g `0.25`, `1.66`, `1.75`, and this is only considering the 2 beats range, but we could imagine a 3, 4 16-beat range).

Did you encounter any blocker when computing this directly on your controller?

mxmilkiib:
Mixxx only has these six ways to manipulate the tempo already, because they are the most frequently misanalysed tempo ratio differences, so saying anyone would say there should be infinity is a bit ab absurdo

I mentioned on Zulip that doing this by timer in JS doesn't keep with the actual beat, not that I'm a great coder

mxmilkiib:
I can imagine calculating at what sample position every flash should be starting on, and listening for the start of playback etc, but how to actually fire the flashes at these points? Wouldn't that gap either need timers or a constant loop to checking the sample position?

mxmilkiib:
But how do I trigger the flashes?

acolombier:
Yes, basically `makeConnection` -> your callback function which computes status like `updateFractionalTempoIndicator` and set the LEDs on accordingly and set a timer to turn it off again -> timer finished and the LED that were set on can now go off

daschuer:
The flashing in fractional beats is interesting. What is your use case? Find out which is the right multiplication? 

Maybe we can add this feature to the metronome? 

Or add an "visual effect" in the effect unit using the effect pad LEDs.

mxmilkiib:
to visually determine which tempo correction is correct

https://youtube.com/shorts/aVJCytU06LM

I ended up with a 1/12 step system to allow multiple scale fractions

https://github.com/mxmilkiib/mixxx-launchpadpromk3/blob/main/LaunchpadProMK3.js#L1701

it's working now but still needs finishing

it wouldn't really work with like detailed movements like grabbing the platter n fucking about

I've set matching colours to denote which ratios can be switched back and forth

PR #14463, opened 2025-03-10
Overview scale options

### LINKED ISSUE #14708: Small overview in 2.6-alpha
### Bug Description

Mixxx 2.6 alpha 301 - Linux Mint 21.3
In installed today the 2.6 alpha version and the waveform is very small. Is that new, a bug, or can that increased anywhere?

Image

### Version

2.6 alpha 301

### OS

Linux Mint 21.3

ronso0:
Fixup for #14309 as the new ReplayGain scaling was considered a regression for certain use case(s).
https://github.com/mixxxdj/mixxx/pull/14309#issuecomment-2673055066

This replaces the overview 'Normalize' checkbox with two options
* Normalize to peak
* Use Waveform "Global" gain and ReplayGain (if enabled)

The second option uses the same gains/factors as the scrolling waveforms, meaning it uses settings in the Waveforms and Normalization pages.
Making 'not normalize' an explicit option hopefully clarifies the situation.

The overviews in the library are now always normalized.
_____
<sub>Initial concept so the conversation here can be understood</sub>
~~This adds 3 overview gain modes~~
* ~~File gain (what we had before this PR with 'Normalize' unchecked) + factor spinbox~~
* ~~Normalize to peak~~
* ~~waveform "All" gain + ReplayGain~~

### DIFF
--- a/src/library/overviewcache.cpp
+++ b/src/library/overviewcache.cpp
@@ -37,41 +37,7 @@ inline QImage resizeImageSize(const QImage& image, QSize size) {
 OverviewCache::OverviewCache(UserSettingsPointer pConfig,
         mixxx::DbConnectionPoolPtr pDbConnectionPool)
         : m_pConfig(pConfig),
-          m_pDbConnectionPool(std::move(pDbConnectionPool)),
-          m_clearingCache(false),
-          m_stopClearing(false) {
-}
-
-void OverviewCache::onNormalizeOrVisualGainChanged() {
-    // Clear the cache and emit changed signal so OverviewDelegate requests
-    // new pixmaps.
-    // Prevent interferences of repeated calls when Normalize or VisualGainAll
-    // are changed in quick succession.
-    if (m_clearingCache) {
-        m_stopClearing = true;
-    }
-    m_clearingCache = true;
-
-    QMultiHashIterator<TrackId, QString> it(m_cacheKeysByTrackId);
-    QSet<TrackId> ids;
-    while (it.hasNext()) {
-        if (m_stopClearing) {
-            m_stopClearing = false;
-            return;
-        }
-        it.next();
-        TrackId id = it.key();
-        ids.insert(id);
-        const auto cacheKey = it.value();
-        QPixmapCache::remove(cacheKey);
-    }
-
-    m_cacheKeysByTrackId.clear();
-    m_clearingCache = false;
-
-    for (const auto trackId : ids) {
-        emit overviewChanged(trackId);
-    }
+          m_pDbConnectionPool(std::move(pDbConnectionPool)) {
 }
 
 void OverviewCache::onTrackAnalysisProgress(TrackId trackId, AnalyzerProgress analyzerProgress) {
--- a/src/library/overviewcache.h
+++ b/src/library/overviewcache.h
@@ -41,7 +41,6 @@ class OverviewCache : public QObject, public Singleton<OverviewCache> {
     };
 
   public slots:
-    void onNormalizeOrVisualGainChanged();
     void overviewPrepared();
     void onTrackAnalysisProgress(TrackId trackId, AnalyzerProgress analyzerProgress);
 
@@ -75,6 +74,4 @@ class OverviewCache : public QObject, public Singleton<OverviewCache> {
     QSet<TrackId> m_currentlyLoading;
     QSet<TrackId> m_tracksWithoutOverview;
     QMultiHash<TrackId, QString> m_cacheKeysByTrackId;
-    bool m_clearingCache;
-    bool m_stopClearing;
 };
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -301,7 +301,7 @@ void MixxxMainWindow::initialize() {
             pOverviewCache,
             &OverviewCache::onNormalizeOrVisualGainChanged);
     connect(WaveformWidgetFactory::instance(),
-            &WaveformWidgetFactory::overviewNormalizeChanged,
+            &WaveformWidgetFactory::overviewScalingChanged,
             pOverviewCache,
             &OverviewCache::onNormalizeOrVisualGainChanged);
 
--- a/src/mixxxmainwindow.cpp
+++ b/src/mixxxmainwindow.cpp
@@ -40,7 +40,6 @@
 #include "library/export/libraryexporter.h"
 #endif
 #include "library/library_prefs.h"
-#include "library/overviewcache.h"
 #include "library/trackcollectionmanager.h"
 #include "mixer/playerinfo.h"
 #include "mixer/playermanager.h"
@@ -292,18 +291,6 @@ void MixxxMainWindow::initialize() {
     WaveformWidgetFactory::createInstance(); // takes a long time
     WaveformWidgetFactory::instance()->setConfig(m_pCoreServices->getSettings());
     WaveformWidgetFactory::instance()->startVSync(m_pGuiTick, m_pVisualsManager, false);
-    // Connect OverviewCache so we can clear and re-render overviews in the library
-    // when "OverviewNormalized" or "VisualGain_0" (all) have been changed in the
-    // preferences.
-    auto* pOverviewCache = OverviewCache::instance();
-    connect(WaveformWidgetFactory::instance(),
-            &WaveformWidgetFactory::visualGainChanged,
-            pOverviewCache,
-            &OverviewCache::onNormalizeOrVisualGainChanged);
-    connect(WaveformWidgetFactory::instance(),
-            &WaveformWidgetFactory::overviewScalingChanged,
-            pOverviewCache,
-            &OverviewCache::onNormalizeOrVisualGainChanged);
 
     connect(this,
             &MixxxMainWindow::skinLoaded,
--- a/src/preferences/dialog/dlgprefwaveform.cpp
+++ b/src/preferences/dialog/dlgprefwaveform.cpp
@@ -22,6 +22,9 @@ const ConfigKey kHardwareAccelerationKey(kWaveformGroup,
         QStringLiteral("use_hardware_acceleration"));
 } // namespace
 
+// for OverviewType
+using namespace mixxx;
+
 DlgPrefWaveform::DlgPrefWaveform(
         QWidget* pParent,
         UserSettingsPointer pConfig,
@@ -33,20 +36,20 @@ DlgPrefWaveform::DlgPrefWaveform(
 
     // Waveform overview init
     waveformOverviewComboBox->addItem(
-            tr("Filtered"), QVariant::fromValue(mixxx::OverviewType::Filtered));
-    waveformOverviewComboBox->addItem(tr("HSV"), QVariant::fromValue(mixxx::OverviewType::HSV));
-    waveformOverviewComboBox->addItem(tr("RGB"), QVariant::fromValue(mixxx::OverviewType::RGB));
+            tr("Filtered"), QVariant::fromValue(OverviewType::Filtered));
+    waveformOverviewComboBox->addItem(tr("HSV"), QVariant::fromValue(OverviewType::HSV));
+    waveformOverviewComboBox->addItem(tr("RGB"), QVariant::fromValue(OverviewType::RGB));
     m_pTypeControl = std::make_unique<ControlPushButton>(kOverviewTypeCfgKey);
-    m_pTypeControl->setStates(QMetaEnum::fromType<mixxx::OverviewType>().keyCount());
+    m_pTypeControl->setStates(QMetaEnum::fromType<OverviewType>().keyCount());
     m_pTypeControl->setReadOnly();
     // Update the control with the config value
-    mixxx::OverviewType overviewType =
-            m_pConfig->getValue<mixxx::OverviewType>(kOverviewTypeCfgKey, mixxx::OverviewType::RGB);
+    OverviewType overviewType =
+            m_pConfig->getValue<OverviewType>(kOverviewTypeCfgKey, OverviewType::RGB);
     int cfgTypeIndex = waveformOverviewComboBox->findData(QVariant::fromValue(overviewType));
     if (cfgTypeIndex == -1) {
         // Invalid config value, set default type RGB and write it to config
         waveformOverviewComboBox->setCurrentIndex(
-                waveformOverviewComboBox->findData(QVariant::fromValue(mixxx::OverviewType::RGB)));
+                waveformOverviewComboBox->findData(QVariant::fromValue(OverviewType::RGB)));
         m_pConfig->setValue(kOverviewTypeCfgKey, cfgTypeIndex);
     } else {
         waveformOverviewComboBox->setCurrentIndex(cfgTypeIndex);
@@ -177,14 +180,16 @@ DlgPrefWaveform::DlgPrefWaveform(
             QOverload<double>::of(&QDoubleSpinBox::valueChanged),
             this,
             &DlgPrefWaveform::slotSetVisualGainHigh);
-    connect(normalizeOverviewCheckBox,
-            &QCheckBox::toggled,
+
+    connect(overview_scale_options,
+            QOverload<QAbstractButton*>::of(&QButtonGroup::buttonClicked),
             this,
-            &DlgPrefWaveform::slotSetNormalizeOverview);
+            &DlgPrefWaveform::slotSetOverviewScaling);
     connect(overviewMinuteMarkersCheckBox,
             &QCheckBox::toggled,
             this,
             &DlgPrefWaveform::slotSetOverviewMinuteMarkers);
+
     connect(factory,
             &WaveformWidgetFactory::waveformMeasured,
             this,
@@ -305,7 +310,6 @@ void DlgPrefWaveform::slotUpdate() {
     lowVisualGain->setValue(factory->getVisualGain(BandIndex::Low));
     midVisualGain->setValue(factory->getVisualGain(BandIndex::Mid));
     highVisualGain->setValue(factory->getVisualGain(BandIndex::High));
-    normalizeOverviewCheckBox->setChecked(factory->isOverviewNormalized());
     // Round zoom to int to get a default zoom index.
     defaultZoomComboBox->setCurrentIndex(static_cast<int>(factory->getDefaultZoom()) - 1);
     playMarkerPositionSlider->setValue(static_cast<int>(factory->getPlayMarkerPosition() * 100));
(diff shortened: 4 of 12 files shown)


ronso0:
Can someone take a look?
As mentioned in the description, it looks like the previous change (use ReaplyGain) is considered a bug by some users.

ronso0:
Well, I tried to stick with the terms we already use in the preferences, which are "ReplayGain" in Normalization preferences and "Normalize", the previous overview checkbox.

But we may switch to
* Original File Gain
* Peak Normalized
* Waveform "All" Gain incl. ReplayGain
* Fixed Gain [spinbox]

Or, appending the custom factor to file gain:
* Original File Gain, custom Factor [spinbox]
* Peak Normalized
* Waveform "All" Gain incl. ReplayGain

> Fixed gain [spinbox] (do we want to have that in DB?) for 0 at unity? Both works for me.

I'm not sure what you are suggesting, can you explain please?
IMO this is a user setting that should not interfere with the db.

daschuer:
Currently the spin box cintains a factor with 1 at unity. 
Normally gain values are expressed as dB with 0 at unity.

daschuer:
I vote for three rows, because than it is clear that the custom gain is applied to the file samples. 

"Original File Gain" there is nothing like that. We have "flies samples" in the file without a gain and apply now one. So it may be "file level" or "original level", "recorded waveform" or such

The idea with "Audio gain" was to emphasize that the visuals follow the audio. The overview act like a vu-meter.

I have no string demands here. Do please consider my points and pick what works best for you.

ronso0:
> I vote for three rows, because than it is clear that the custom gain is applied to the file samples.

True, I've changed it accordingly (updated pic in description)

You're right about gain vs. factor/level, I also changed that everywhere.

daschuer (changes requested):
Looks good now. Thank you. Unfortunately the "All" gain mode is broken. (Tested this PR merged to main)

ronso0:
Hmm, I can not confirm. All modes work as expected, both this PR and with main merged.
overview-scale-modes

daschuer:
In the recording you use the "Global Gain" in the "Scrolling waveform" region. We however adjust the "Overview Waveforms". I did not expect this. 

My expectation was the the last option makes waveform follow the gain in the Mixer and the replay gain inkl. target loudness and boost value just like the scrolling one. This does not happen. 

Looking at the RGB scrolling waveforms. I am able  to shift the colours with the there "Visual Gains". The RGB overview does not follow, which creates a different visual. I think this shall not happen. How about applying the visual gain to both waveforms and in all thee gain settings?

ronso0:
> In the recording you use the "Global Gain" in the "Scrolling waveform" region. We however adjust the "Overview Waveforms". I did not expect this.

Indeed slightly "unintuitive", but that's how it was since long. Run 2.5, uncheck "Normalize overview waveforms", adjust the Global ~~gain~~ scale factor (the tooltips says Gain btw).
It's just that we separated the overview / waveform options, and in that PR #13615 we already discussed the not obvious effect of Global gain on waveforms. The temp fix/hint was if waveforms are Empty, all their settings are greyed out but the Global gain. Not optimal, I agree.

"All" [gain] should be "Global" to clarify the situation (my mistake)

I'll push a fixup.

ronso0:
> How about applying the visual gain to both waveforms and in all thee gain settings?

Sure. Requires to rework the individual`waveformOverviewRenderer` and would thereby also apply to the overview column.
But out of scope for this PR.

ronso0:
Btw we also had and still have the issue that the "Global" factor produces different results for each waveform type.
Let's consider this one step forward and improve it in follow-ups. Maybe ReplayGain * custom factor would be the best option.

**And: I think this should go to 2.6 as this is a fixup for #14309** 
Wdyt?

ronso0:
Oh, `<comment>` does not work as translator hint in ui files' `text` elements.
Will revert.

Any other way to add tr hints there? Or some combination with `notr`?

daschuer:
Ok, to move on if it is to much wor,   but I am puzzled with one thing: 

The scrolling and overview facing the same scaling issues. I assume that the third option is designed to make them equally behave (in case EQs and individual visual gains are at unity). However this does not work, because the two replay gain setting sliders are ignored for the overview. 

I normally use -16 LUFS instead of the -18 default. Other user may use others. This leads to the situation that one of the waveform is to small while their other exceeds the visual range. Conclusion: we need to apply all replay-gain related gains to make it usefull.

ronso0:
> Conclusion: we need to apply all replay-gain related gains to make it usefull.

If you mean to include the deck's current `pregain`, I disagree. Pre-gain is not necessarily constant during playback. This is because some tracks are very dynamic and I may have corrected ReplayGain to give a proper level for the most important parts of the track while being aware of potential clipping (-> adjust pregain as desired, eg. I raise pregain in quiet regions, or lower it it if necessary).

So the desired level is always the ReplayGain IMO.
However this will not gives us the same visual gain as mentioned above. Eg. for 

My track gain / visual gain adjust procedure is this:
1. load a track
2. check the overview to find the loudest part, create a short loop
3. play the loop, adjust pregain to make the VU stay in the yellow region (= loud but no clipping)
4. in Waveform preferences, select the desired type, and adjust Global visual gain to get the same 'just not clipping' level
5. overview should match that

So 5. is the hardest part because 4. all waveform renderer types have a specific visual gain.
I guess if this is fixed, 'ReplayGain + All gain' should be the default.

Personally I think I'll stick to 'Normalize'.
'File Level' with optional factor probably only makes sense if you have ReplayGain disabled (all files already have the desired level?).

daschuer:
> If you mean to include the deck's current pregain, I disagree.

No, pregain is unrelated. I mean the "Target loudness" and "Initial boost without ReplayGain data".  

> So the desired level is always the ReplayGain IMO.

Yes, but the replaygain target is adjustable this adjustments need to be applied. 

The waveform does not use the pregain or replaygain, it uses the total_gain 

https://github.com/mixxxdj/mixxx/blob/21a3f865615897e899423e4ec56b77bb37729817/src/waveform/renderers/waveformrenderersignalbase.cpp#L176
https://github.com/mixxxdj/mixxx/blob/21a3f865615897e899423e4ec56b77bb37729817/src/waveform/renderers/waveformwidgetrenderer.cpp#L113

Which is 
`total_gain = pregain * replaygain * ReplayGainBoost`
or 
`total_gain = pregain * DefaultBoost`
or 
`total_gain = pregain`  if replaygain is disabled 

What you need here is probably 
overview_gain = `total_gain/pregain` * all_gain (Global)` 

With this in palace we have always a waveform related to the VU-meter, but without following pregain and EQs like the scrolling waveform. This is true independent form the Replay gain preferences. 

I have initially not understand that "global" is for all waveform and the GUI with the extra gain for the file level is confusing. 

Do we really have a use case for file level if replaygain is used? This will produce tiny waveforms or overshoots dependening on the track. I would propose to just drop it. 

Than we can probably replace everything with one checkbox: 
[  ] Normalize to peak

Tooltip: 
"Uncheck to scale with the scrolling waveform but without gain and EQs."

ronso0:
You're right about ReplayGainBoost, but let's leave pregain out of the equation for the overviews: unlike waveforms, the overview **should not change** when mixer controls are touched.
The purpose of ReplayGain (* Boost) is that you can leave pregain untouched, and as I explained above, it is of no use for the overviews IMO.

I'll try to implement ReplayGainBoost.

daschuer:
Please look closely to my equations. The proposed overview_gain does **not** follow pregain. Using the CO interface for this is straight forward. the only issue you need to remove "pregain" from "total_gain" by a division. The rest of the logic is already part of "total_gain". So you don't need to repeat any replay gain preferences business logic

ronso0:
Okay, so listening to `total_gain` and querying `pregain` should suffice.
And the overview column just has Normalize or file level for now.

ronso0:
Hmm no, `total_gain` is not usable IMO. It's technically correct, though it's updated every time `pregain` is updated, means we scale the overview every time which is unnecessary and leads to flickering. This also can't cover the case when `total_gain` is 0 (due to `pregain` being 0), the visual_gain would increase it that case.

daschuer:
Ah yes there is a division by zero issue whenever pregain is zereo. 
So, you may extract a reusable function from the code where total_gain is calculated and use it for the overview gain without adding pregain. Than we make sure it follow the same business logic like total_gain used for the scrolling waveforms (and audio).

ronso0:
Yup, that's what I just did, and it works as desired :tada: 

Though, if I repeat my procedure
> 1. load a track
> 2. check the overview to find the loudest part, create a short loop
> 3. play the loop, adjust pregain to make the VU stay in the yellow region (= loud but no clipping)
> 4. in Waveform preferences, select the desired type, and adjust Global visual gain to get the same 'just not clipping' level
> 5. overview should match that

the overview is still taller when Normalized :shrug: 
Is it just that the spikes (peaks) are not rendered in the overview as detailed as they are in the waveforms?

ronso0:
Wookay, update pushed.

I'd leave the FileLevel mode in for now.

ronso0:
and squash and rebase onto 2.6 as mentioned earlier.
Wdyt?

daschuer:
I seen a warning: 
```
/home/daniel/workspace/mixxx/src/widget/woverview.cpp:783:40: warning: conversion from ‘double’ to ‘CSAMPLE_GAIN’ {aka ‘float’} may change value [-Wfloat-conversion]
  783 |                             replayGain * static_cast<CSAMPLE_GAIN>(m_pReplayGainBoost->get());
```

daschuer:
I can confirm. It works 🎉

Would you use personally  "file level with factor"? What I cannot think of a reason. 

There is a scaling issue with the scrolling Filtered and Stacked type. They still not match to the overview. When I remove the factor 2 here: https://github.com/mixxxdj/mixxx/blob/5320334da5f3f33783871d89f59cdc862152b7a0/src/waveform/renderers/waveformwidgetrenderer.h#L148
it works. 

Regading defaults: 
Mixxx commes up with -18 LUFS replay Gain enabled. This is the setting which also allows high dynamic classic tracks. With this the waveforms look really small. 

From my experience with normal modern compressed tracks. -16 LUFS (also recommended) is fine to be always in the Green VU-meter area, but will still unlikely clip during transitions. 

In that case a Global Gain of 2.0 works perfect with all overview types. I need no headroom in the overview vissual area. Maybe this was the internal gain factor? I think we should expose it to show the truth. 


> Though, if I repeat my procedure
> 1. load a track
> 2. check the overview to find the loudest part, create a short loop
> 3. play the loop, adjust pregain to make the VU stay in the yellow region (= loud but no clipping)
> 4. in Waveform preferences, select the desired type, and adjust Global visual gain to get the same 'just not clipping' level
> 5. overview should match that

That does not work, because the pregain does only affect the scolling waveforms. You may use the Target Loudness instead. 

I can confirm it work for RGB. For filtered it is difficult, because the if the track is clipping, the sum of all bands is clipping. Not an individual peak in the waveform.

ronso0:
> > Though, if I repeat my procedure
> > 
> > 1. load a track
> > 2. check the overview to find the loudest part, create a short loop
> > 3. play the loop, adjust pregain to make the VU stay in the yellow region (= loud but no clipping)
> > 4. in Waveform preferences, select the desired type, and adjust Global visual gain to get the same 'just not clipping' level
> > 5. overview should match that
> 
> That does not work, because the pregain does only affect the scolling waveforms. You may use the Target Loudness instead.

Ups, I forgot to mention step 3.1: **adjust ReplayGain from pregain**
The overview is adjusted as expected.

ronso0:
> Would you use personally "file level with factor"? What I cannot think of a reason.

I don't use it either, but as I wrote I actually want to put this atop 2.6 in order to fixup #14309 which is a regression for some users. And I'm hesitant to remove this legacy 'file level' option in beta. We may do that in 2.7

ronso0:
> There is a scaling issue with the scrolling Filtered and Stacked type. They still not match to the overview. When I remove the factor 2 here:
> 
> https://github.com/mixxxdj/mixxx/blob/5320334da5f3f33783871d89f59cdc862152b7a0/src/waveform/renderers/waveformwidgetrenderer.h#L148
> 
> it works.

I'm not familiar with the renderers, so this is related to
#12448 (fixed?)
#12115
and I think another one which I don't seem to find atm.

daschuer:
> .. fixup https://github.com/mixxxdj/mixxx/pull/14309 which is a regression for some users.

For my understanding "file level" option is already restored implicit in case users disable ReplayGain in Normalization preferences. Other users don't care about file level, fully invisible for them.   

We have also this comment:  
https://github.com/mixxxdj/mixxx/pull/14309#issuecomment-2673055066

@jwcd Maybe you can jump in and can confirm my assumption. 

The first impression of the change felt like a regression was "overwiew is now too small". This happens because he has mostly modern compressed, almost clipping tracks tracks. They are significant reduced by ReplayGain. A 70th or classic track with a ReplayGain of 0 dB would have been also that small. That's why I think the file gain is useless as universal solution. 

Now the users can use "Global" gain to 2.0 or such to make use of the visual headroom. 
Classic tracks may have clipping regions when the Trombones take over, but OK.    

--- 

https://github.com/mixxxdj/mixxx/issues/12448 Was about clipping in an earlier stage and has been fixed. 

https://github.com/mixxxdj/mixxx/issues/12115 is somehow related. My guess is that it is fixed in the data structures, but @m0dB was not brave enough to also remove this comment:

https://github.com/mixxxdj/mixxx/blob/5320334da5f3f33783871d89f59cdc862152b7a0/src/waveform/renderers/waveformwidgetrenderer.h#L137

My interpretation is like that: 
At the time we made the waveform depending on the audio gain, we noticed that they where too small. Almost the same we did here. To compensate that we added the 2.0 internally. Because of https://github.com/mixxxdj/mixxx/issues/12115, it was not suitable for all waveform and we made it conditionally via "applyCompensation". Now the issue is fixed which had the side effect of making the RGB and HSV waveforms with `applyCompensation = false` too small. 

Solution: Remove the "applyCompensation" hack and make "Global" waveform gain 2.0 by default. 

People using Mixxx with default ReplayGain = On have a nice scaled waveform by default. 
People with a higher Loudness setting or Replay Gain Off, needs to adjust the "Global" waveform gain to their needs. 

Does that make sense?

ronso0:
Hmm, I have to take another look how it was in 2.5
We can of course remove the File level option if it wasn't therer previously.

daschuer:
The Replay Gain is now always applied. I think it shall fall back to file gain if replay gain is disabled. For users with disabled replaygain, the replaygain is not interesting. They want the file gain, to be able to estimate the manual gain the need to apply to the track.  

In 2.5 we don't have a preferences options at all for the overview waveform. It uses the file level + Global gain. This looks good for modern compressed tracks but not for high dynamic tracks. 

image

In this case the Global gain of 2.0 and -18 LUFS gives a good results for the moving waveform. It let's however the compressed track overview clip. This can be considered as bug. 

The following screenshot is from this branch with the ReplayGain option and the same settings like above. 

image

We have a bit of a headroom in the Cock Sparrer track, which is the penalty for the heavy compressing during mastering. Else it looks perfect. User with only modern tracks can adjust the LUFS to -16 or more to make use of the headroom. 

Conclusion: File gain overview waveform is useless when user uses ReplayGain and wise versa. We can remove the file gain option in waveform preferences and follow the "ReplayGain enable" checkbox.

ronso0:
> The Replay Gain is now always applied.

Huäh?? (german for "WTF mate?")

It worked perfectly as desired last time I built it.
Now when I rebuild, I see the wrong behavior, too. Stuck ~~in ReplayGain mode.~~ with ReplayGain when I change the Normalization settings (toggle Apply ReplayGain).
Interestingly, it works again when I add some qWarning() to trace `WOverview::drawWaveformPixmap`.
All signal/slot connections are correct.

I have absolutely no clue why that happens.

ronso0:
Ouuuh..

`if (m_pReplayGainEnabled->toBool()) {`
not
`if (m_pReplayGainEnabled) {` (ControlProxy)

ronso0:
Okay, I removed the File Level option, which leaves us with
- [ ] Normalized
- [ ] waveform "Gobal" gain + ReplayGain (if enabled)

I.e. we can handle this with the existing `[Waveform] OverviewNormalized` config option.

Thanks for your thourough review, with it I could shrink this to the essentials :+1: 

Only think left is: how to scale the overviews in the library?
(and update them when a track's ReplayGain changes)
1. always normalize to peak
2. duplicate the ReplayGain calculation in overviewRenderer (ideally move to a static helper with `ControlObject::get()`)

Currently it does normalize or use the file level.
If left a TODO for now.

ronso0:
> Only think left is: how to scale the overviews in the library?

Notifying the overview chache of track gain changes is doable, it could then be passed to overview renderer, or that queries the ReplayGain from GlobalTrackCacheLocker.
But updating when ReplayGain settings changes is currently pretty elaborate (I think of DlgPrefReplayGain -> ReplayGainSettings -> MixxxMainWindow -> OverviewCache). Something like a `ReplayGainSettings` singleton might simplify this a lot.

daschuer:
Usingen the normalized version would be fine for me. This probably saves us from complex code.
I don't think there is much use for a gain estimation in a tiny library row.

ronso0:
Are you suggesting to show **only Normalized** overviews in the library?
Would make sense because now they're changed when selecting ReplayGain but don't show the truth.

daschuer:
> Are you suggesting to show only Normalized overviews in the library?

Yes

daschuer:
Sorry, the famous close with comment button ...

ronso0:
Alright, this is ready for a code review.

daschuer:
Code looks good, Thank you. I need to do a final manual test. 

How do we want to proceed with the different sizes in case of filtered view. The filtered view in the overview look has the same size than the RGB view (subject of this PR) . This is however not the case for the moving waveform. Here it is twice as big due to the `applyCompensation` which is now done elsewhere and it is actually an overcompensation. 

Proposal: Remove `applyCompensation` and use a 2.0 default for Global.

ronso0:
This PR is only about the overviews, so please file an Issue where we can discuss the waveform gain issue.

**Edit:**
In general I'd like the waveform gain match the VU meter level, ie.
maximize ReplayGain so that the track just never clips so that waveforms amplitude matches the VU meter range (incl. yellow).
Currently Stacke and Filtered are to tall, and RGB (accelerated, stereo, high details) is a bit too small.

ronso0:
> Sorry, the famous close with comment button ...

In Firefox I've been using this snippet for years and it works beautifully
https://gist.github.com/ronso0/a30d1a11f4248df98cccdd3b4ccbd330
image

daschuer:
I have noticed that this was targeting main, however it is fixing an 2.6 issue. 
Can you rebase it to 2.6? 
(I have already changed the target branch)

ronso0:
Alright, rebased.

@daschuer please don't change the target branch.
I prefer to do this myself so that during `rebase -i` I can see which commits are actually from this branch
(I reset target to `main` and could spot them, so no big deal)

daschuer:
> @daschuer please don't change the target branch.

OK next time. This is now ready to go. Thank you

daschuer (approved):
Still works. LGTM

PR #14472, opened 2025-03-13
(fix) use canonical path to load font file

ronso0:
For some reason I get these warnings with main, but not with 2.5
```
warning [Main] Failed to add font: "OpenSans-Bold.ttf"
warning [Main] Failed to add font: "OpenSans-Regular.ttf"
warning [Main] Failed to add font: "OpenSans-Semibold.ttf"
warning [Main] Failed to add font: "Ubuntu-B.ttf"
warning [Main] Failed to add font: "Ubuntu-R.ttf"
```

Fixed by using the files' canonical paths.

### DIFF
--- a/src/util/font.cpp
+++ b/src/util/font.cpp
@@ -67,16 +67,16 @@ void FontUtils::initializeFonts(const QString& resourcePath) {
         return;
     }
 
-    const QList<QString> files = fontsDir.entryList(
+    const QFileInfoList files = fontsDir.entryInfoList(
             QDir::NoDotAndDotDot | QDir::Files | QDir::Readable);
-    for (const QString& path : files) {
+    for (const QFileInfo& file : files) {
         // Skip text files (e.g. license files). For all others we let Qt tell
         // us whether the font format is supported since there is no way to
         // check other than adding.
-        if (path.endsWith(QStringLiteral(".txt"), Qt::CaseInsensitive)) {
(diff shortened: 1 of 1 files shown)


daschuer:
I can confirm that this is only an issue with main, not with 2.5 and 2.6

daschuer:
The API mentions to use the "fileName" https://doc.qt.io/qt-6/qfontdatabase.html#addApplicationFont not the path. 
So maybe there are implications with our other targets. We need either test on those or even better to fin the root cause of this regression.

ronso0:
My understanding is that QFontDatabase is populated with fonts from system font directories.
Here we have additional fonts in our own _fonts_ directory which may not have been indexed.

Or (assumption, didn't find any documentation, only hints in Qt forums):
The fonst are also added to the font database if they're in (the QApplication's) [resource-dir]/fonts.
Maybe there were some changes to how the resource dir is marked?
Unfortunately I have no idea where to start looking.. Do you?

Anyhow, adding fonts by path works.

ronso0:
Okay, so it's confirmed that this fixes the failing font import on Windows and Ubuntu.
[#testing > group: windows_testers @ 💬](https://mixxx.zulipchat.com/#narrow/channel/267968-testing/topic/group.3A.20windows_testers/near/544301285)

Let's merge?

daschuer (approved):
Top. Thank you.

PR #14475, opened 2025-03-13
github labeler: add Dev Tools to `developer experience`

ronso0:
Btw can we backport the labeler patterns to 2.5?


acolombier (approved):
backport sounds good to me!

ronso0:
Alright, can do.
But what I meant was actually: do we have the same github action/versions on all branches, or are there any other reasons why the patterns are way more detailed in the main branch?

Issue #14363, opened 2025-02-17
make Left/Right key in beatjump spinbox jump/loopmove


ronso0:
### Feature Description

Would allow quick jump/loopmove while changing the size without having to switch to mouse / controller / keyboard shortcuts.

ronso0:
Maybe even do beatjump/loopmove when in the loopsize spinbox?

icezuk:
Hello @ronso0 ,

Regarding the issue [#14363](https://github.com/mixxxdj/mixxx/issues/14363) I've found a solution that I will test further as soon as the [build error](https://mixxx.zulipchat.com/#narrow/channel/109171-development/topic/build.20error.20in.20main) is resolved. So far it works perfectly but until then I have two questions:

1) Do you want the beatjump/loopmove in the loopsize spinbox to behave the same as in the beatjump spinbox? I am asking this as a somewhat intuitive functionality would be for the left and right buttons to match the arrows underneath the spinbox(which in this case would be the start and stop of loop). 

2) This is a somewhat minor thing, but it might result in code review return, do you prefer the beatjump to be done via beatjump_forward/backward for the corresponding jump or rather beatjump with positive and negative values?

Thank you!

ronso0:
Thanks @icezuk for working on this!

> Do you want the beatjump/loopmove in the loopsize spinbox to behave the same as in the beatjump spinbox?

Good question. I think it's unepxected that Left/Right keys do anything except moving the cursor. I have WIP where I worked with the cursor position to not break the default QLineEdit behavior.
L/R in the loopsize spinbox should maybe not do anything if there's no active loop?
If there is one, beatjump is switched to loopmove.
Additional functions could be
* Ctrl + L/R moves the loop by one beat
* Alt + L/R moves the loop by 1/8 beat (I use this often to adjust manually set loops)

The current workflow to do both of these adjustements is a bit slow / cumbersome as it requires adjusting the beatjump size first, so this could be a nice addition.
Wdyt?


Another tricky part IMO is how to detect in an elegant way what the spinbox is controlling.

Looking forward to see you implementation!

icezuk:
Hello @ronso0 ,

I did a draft pull request with the current changes if you have the time to check them out: https://github.com/mixxxdj/mixxx/pull/14476

There are two known issues:
----One is the bug where the beatjumpForward/Backward arrows light up with beatjumpSize one
----The other is the hard-coded values for QSize

[...]

PR #14476, opened 2025-03-13
WBeatspinbox: added jump feature with L/R button in beatjump and loopmove

icezuk:
Adding the functionality to jump using the L/R buttons while in Beatjump/Loopmove or Loopsize spinbox. 
Related to [14363](https://github.com/mixxxdj/mixxx/issues/14363). 
Additions:
1. Ctrl + L/R moves the loop by one beat
2. Alt + L/R moves the loop by 1/8 beat
Not ready to be merged, because of known issue: After using the L/R buttons the BeatjumpBackward/Forward arrows get marked if the value of BeatjumpSize is 1 and unmarked if it is anything else. 

Also would love to know any other mechanism of detecting which spinbox is currently selected other than the hardcoded QSize.


src/widget/wbeatspinbox.cpp
```diff
@@ -21,6 +21,18 @@ WBeatSpinBox::WBeatSpinBox(QWidget* parent,
...
+          beatJumpSize(configKey.group, "beatjump_size"),
+          beatJumpOneForward(configKey.group, "beatjump_1_forward"),
+          beatJumpOneBackward(configKey.group, "beatjump_1_backward"),
+          beatJumpEightForward(configKey.group, "beatjump_0.125_forward"),
+          beatJumpEightBackward(configKey.group, "beatjump_0.125_backward"),
+          beatLoopEnabled(configKey.group, "loop_enabled"),
+          beatLoopSize(configKey.group, "beatloop_size"),
+          // Hardcoded values for spinbox and loopbox for differentiation between the two
+          sizeSpinBox(60, 26),
+          sizeLoopBox(78, 26),
```
ronso0, line 35: The size can not be used as identifier, it's different in each skin that uses this widget. A skin template sets a fixed width, height or both, or a minimum size and some QSizePolicy. Please take a look at the skins, search for `<BeatSpinBox>`.

But you could use `configKey`. That is a `ConfigKey` consisting of a group and a key. Currently used keys are `beatloop_size` and `beatjump_size`.

src/widget/wbeatspinbox.cpp
```diff
@@ -295,6 +307,36 @@ void WBeatSpinBox::keyPressEvent(QKeyEvent* pEvent) {
         ControlObject::set(ConfigKey("[Library]", "refocus_prev_widget"), 1);
         return;
     }
+    if (this->size() == sizeSpinBox) {
+        if (pEvent->key() == Qt::Key_Right && pEvent->modifiers() == Qt::ControlModifier) {
+            beatJumpOneForward.set(beatJumpSize.get());
```
ronso0, line 312: Please check what `beatJumpSize.get()` returns.
And which values make sense for [`beatjump_N_forward`](https://github.com/mixxxdj/mixxx/blob/b1967eaeabd8012ec7d51bce7e8c59e18011d39a/src/engine/controls/loopingcontrol.cpp#L1945) and `beatjump_forward/_backward`.

ronso0: `this->` is redundant btw.

src/widget/wbeatspinbox.h
```diff
@@ -31,6 +31,19 @@ class WBeatSpinBox : public QDoubleSpinBox, public WBaseWidget {
...
+    ControlProxy beatJumpBackward;
+    ControlProxy beatJumpSize;
+    ControlProxy beatJumpOneForward;
+    ControlProxy beatJumpOneBackward;
+    ControlProxy beatJumpEightForward;
+    ControlProxy beatJumpEightBackward;
+    ControlProxy beatLoopEnabled;
+    ControlProxy beatLoopSize;
+    QSize sizeSpinBox;
+    QSize sizeLoopBox;
```
ronso0, line 45: For class members we use `m_` prefix.
I suggest you take a look at https://github.com/mixxxdj/mixxx/wiki/Coding-Guidelines

ronso0 (changes requested):
Thanks for this PR!
I left some notes.

src/widget/wbeatspinbox.cpp
ronso0: Re Zulip direct message:
I refer to `configkey` in the `WBeatSpinBox` constructor. You can compare the `configkey.key` (QString) with "beatjump_size" and "beatloop_size". Depending on the match you know which spinbox you're deaing with.

ronso0:
> Not ready to be merged, because of known issue: After using the L/R buttons the BeatjumpBackward/Forward arrows get marked if the value of BeatjumpSize is 1 and unmarked if it is anything else.

That is to be expected.
In skins there are  `WPushButton` widgets, connected to corresponding ControlPushButtons.
They are styled with qss (Qt stylesheets), and stylesheets usually only contain selectors for the expected states.
`beatjump_forward` is a trigger button so it has only 2 states styled: 0 and 1.
(you can start Mixxx with `--developer` arg to get more info in the widget tooltips, eg. state of connected controls, size etc.)

To trigger the beatjump via `BeatJump::slotJumpForward` you only need value > 0. So if you set it to `beatJumpSize.get()` (which can be  anything between 1/32 and 512) there is simply no stylesheet selector for this ControlObject state.
If you use value 1, the GUI buttons are lit (which is fine since it provides feedback for what is happeing).
So to reset the GUI button (and the ControlPushButton) you'd do this:

`beatJumpOneForward.set(1);` CO is 1, button is lit ([style](https://github.com/mixxxdj/mixxx/blob/2091886c53c3194130dd6a64cf71b76c665dd7e4/res/skins/LateNight/style_palemoon.qss#L1709)), beatjump slot is called, beatjump is performed by engine
`beatJumpOneForward.set(0);` CO is 0, button unlit ([style](https://github.com/mixxxdj/mixxx/blob/2091886c53c3194130dd6a64cf71b76c665dd7e4/res/skins/LateNight/style_palemoon.qss#L1616)), nothing else happens

ronso0:
I think I already answered your questions above.

icezuk:
Hi @ronso0 ,

Changes done. Appreciate if you could take a look.

Thanks

ronso0:
I'm sorry I pushed this back on my list.
There are a few more adjustements necessary, will review soonish.

src/widget/wbeatspinbox.cpp
```diff
@@ -295,6 +302,47 @@ void WBeatSpinBox::keyPressEvent(QKeyEvent* pEvent) {
...
+        } else if (pEvent->key() == Qt::Key_Left && pEvent->modifiers() == Qt::AltModifier) {
+            m_beatJumpEightBackward.set(1);
+            m_beatJumpEightBackward.set(0);
+        } else if (pEvent->key() == Qt::Key_Right) {
+            m_beatJumpForward.set(1);
+            m_beatJumpForward.set(0);
+        } else if (pEvent->key() == Qt::Key_Left) {
+            m_beatJumpBackward.set(1);
+            m_beatJumpBackward.set(0);
+        }
```
ronso0, line 324: I think `Alt` and `Alt`+`Shift` are the only combinations we can safely use without breaking default QLineEdit behavior:
* `Ctrl`+`Left/Right` normally jumps to prev/next word (boundary)
* `Ctrl`+`Shift`+`Left/Right` normally extends/shrinks selection up to prev/next word boundary
* pure `Left/Right` jumps back/forth by character

So we actually have only two jump sizes available. One should be the current `beatjump_size`.
The other one I don't know.

Let's try evaluate which jump sizes would be useful useful.
The [default keyboard mapping](https://manual.mixxx.org/2.5/en/chapters/controlling_mixxx#using-a-keyboard) already provides shortcuts for 1 beat back/forth for deck 1 & 2.
One scenario I can imagine is that you just activated a loop and go to the beatjump size box to shift it.

In case the loop is positioned a bit too early or too late, 1/8 beat would be handy (without having to adjust the jump size).
And even though we have GUI buttons for +- current `beatjump_size`, doing these jumps from within the size box seems reasonable (ie. without having to use the touchpad/mouse for the GUI buttons).

What do you think?
What's youruse case?

src/widget/wbeatspinbox.h
```diff
@@ -31,6 +31,15 @@ class WBeatSpinBox : public QDoubleSpinBox, public WBaseWidget {
...
     ControlProxy m_valueControl;
 
+    // for L/R button BeatJump/LoopMove
+    ControlProxy m_beatJumpForward;
+    ControlProxy m_beatJumpBackward;
+    ControlProxy m_beatJumpOneForward;
+    ControlProxy m_beatJumpOneBackward;
+    ControlProxy m_beatJumpEightForward;
+    ControlProxy m_beatJumpEightBackward;
+    ControlProxy m_beatLoopEnabled;
```
ronso0, line 41: We usually use pointers, specifically `parented_ptr` (src/util/parented_ptr.h). You'll find many examples in the codebase for how to use ad create those.

ronso0 (changes requested):

src/widget/wbeatspinbox.cpp
icezuk: I think that the use cases that you described are what I've been using these for. I am using this for quickly jumping when I'm inside the the beatjump size box. That's also what's the description of the issue is. I personally don't use the 1/8 that much but you've said in the issue to this PR that you like it, so I don't see why leave it.

About the buttons we can use safely without breaking QLineEdit, do you think that just switching the modifiers from alt and ctrl to alt and alt+shift is reasonable and applicable here? The more natural (to me at least) distribution is alt for 1 beat and alt+shift for 1/8. What do you think about that?

The changes here are simple but we need to come to some consensus on how to do it exactly.

ronso0: > do you think that just switching the modifiers from **alt** and **ctrl** to **alt** and **alt+shift** is reasonable and applicable here?

That's what I was trying to say.

ronso0: > The [default keyboard mapping](https://manual.mixxx.org/2.5/en/chapters/controlling_mixxx#using-a-keyboard) already provides shortcuts for 1 beat back/forth for deck 1 & 2.

I just noticed that, unfortunately, the regular shortcuts (a-z + modifiers) don't work in the boxes (when they have focus). But solving this is not necessary to implement the jump shortcuts at least partially.

For the loop box I suggest the quick fix shortcuts:
* Alt: jump +-1 beat
* Alt+Shift: jump 1/8 beat

For the jumpsize box we may use
* Alt for the selected jump size
* Alt+Shift for 1/8 beat

What do you think?
What do others think? @mixxxdj/developers

icezuk: Hi @ronso0 

I implemented the changes and will push them here soon but I have a question. In your suggestion you say that the Alt modifier should be used in jumpsize box for selected jump size, but that is already done via no modifiers(just left/right). Did you mean Alt for +- 1 beat, same as loop box?

ronso0: > that is already done via no modifiers

pure Left/Right are used to move the cursor, so performing any special actions for these would be surprising.
Let's stick to the modofier combos.

icezuk: Ok so Shift + Alt is not it, haha

I changed the modifiers to be Shift + Alt. I used both the code in this PR(with pEvent->modifiers() == Qt::$some_modifier) and pEvent->modifiers().testFlag(Qt::$some_modifier). Both result in broken behavior, which is I believe what you refer to when talking about broken QLineEdit.

Here is a video showcasing the aforementioned broken behavior. The first part is only using Alt for selected beatjump and then it's Alt + Shift for 1/8 first only forward and then both forward and backward(the first break happens with backward).

https://github.com/user-attachments/assets/3dec8712-9261-4dac-a25a-f920f074c135

If this is not a mistake on my front, I suggest using Ctrl + Alt or even Shift + Ctrl as this does not happen with these combinations. Shift + Alt seems to break behavior even when it is not being detected for any beatjump.

ronso0: Uuuh, I totally overlooked that `Alt`+`L/R` are the default shortcuts for loading a track to deck 1/2, and `Alt`+`Shift`+L/R does **eject** deck 1/2.
These default global shortcuts should always work, regardless which widget has focus, they are caught in the main window before they are passed down the widget tree until they are maybe processed by individual widgets.
So all that's left that would work without (Mixxx internal) conflicts would be `Ctrl`+`Alt`, but that may be used for (system) global shortcuts (I have it mapped to move left/right between virtual desktops).

I'm sorry, I think this is a dead end.

icezuk: What about substituting Left/Right with A/D? The pairing with Alt and Alt+Shift could be the same as [your previous comment](https://github.com/mixxxdj/mixxx/pull/14476#discussion_r2038212668). I tested it and it works properly. But do you think I got ahead of myself and maybe this solution is not practical for users? I personally feel as if it is comfortable, but I have to factor in my own bias, haha. Anyway do share what you think!

PR #14477, opened 2025-03-14
Move VCPKG to GH Release

acolombier:


tools/macos_buildenv.sh
```diff
@@ -15,28 +15,82 @@ realpath() {
...
+        find . -name "mixxx-deps-${1}-${3}-${2}.*" -delete
+        exit 1
+    fi
+}
+
 # some hackery is required to be compatible with both bash and zsh
 THIS_SCRIPT_NAME=${BASH_SOURCE[0]}
 [ -z "$THIS_SCRIPT_NAME" ] && THIS_SCRIPT_NAME=$0
 
+VCPKG_COMMIT_SHA="d3b6db3"
```
daschuer, line 76: I think we had this at one point, but is was a hell of a nightmare to maintain. It is much better to just copy the content of the sha256 file literally. It would even better if we can just use the whole string as it is in these files and add here a parse for it. This way it will no loner happen that you get confused with the right checksum at the right branch and so on.

tools/macos_buildenv.sh
```diff
@@ -15,28 +15,82 @@ realpath() {
...
+BUILDENV_MINOR_VERSION="2.6"
+
 if [ -n "${BUILDENV_ARM64}" ]; then
     VCPKG_TARGET_TRIPLET="arm64-osx-min1100-release"
-    BUILDENV_BRANCH="2.5-rel"
-    BUILDENV_NAME="mixxx-deps-2.5-arm64-osx-min1100-release-40c29ff"
```
daschuer, line 25: In https://github.com/mixxxdj/vcpkg/pull/157 I have proposed to drop min1100 from the file name and add cross in case it is a cross compiled environment. I would prefer to merge my PR first and get around of a second PR that introduces the renaming here. This keep the naming schema on GitHub consistent from the start.

daschuer (changes requested):

tools/macos_buildenv.sh
```diff
@@ -56,22 +110,14 @@ case "$1" in
...
-                if [ "$http_code" -ne 200 ]; then
-                    echo "Downloading  failed with HTTP status code: $http_code"
-                    exit 1
-                fi
-                curl "https://downloads.mixxx.org/dependencies/${BUILDENV_BRANCH}/macOS/${BUILDENV_NAME}.zip" -o "${BUILDENV_PATH}.zip"
-                OBSERVED_SHA256=$(shasum -a 256 "${BUILDENV_PATH}.zip"|cut -f 1 -d' ')
-                if [[ "$OBSERVED_SHA256" == "$BUILDENV_SHA256" ]]; then
-                    echo "Download matched expected SHA256 sum $BUILDENV_SHA256"
+                if [ "$BUILDENV_SRC" = "gh" ]; then
+                    echo "Build environment $BUILDENV_NAME not found in mixxx repository, downloading from Mixxx's VCPKG repo"
```
daschuer, line 114:
```suggestion
                    echo "Build environment $BUILDENV_NAME not found at ${BUILDENV_BASEPATH}, downloading from https://github.com/mixxxdj/vcpkg/releases"
```
Showing the URLs for a bit more confidence.

tools/macos_buildenv.sh
```diff
@@ -56,22 +110,14 @@ case "$1" in
...
+                    buildenv_from_gh_release "${BUILDENV_MINOR_VERSION}" "${VCPKG_COMMIT_SHA}" "${VCPKG_TARGET_TRIPLET}" "${BUILDENV_PATH}.zip"
                 else
-                    echo "ERROR: Download did not match expected SHA256 checksum!"
-                    echo "Expected $BUILDENV_SHA256"
-                    echo "Got $OBSERVED_SHA256"
-                    exit 1
+                    echo "Build environment $BUILDENV_NAME not found in mixxx repository, downloading from Mixxx download server"
```
daschuer, line 117:
```suggestion
                    echo "Build environment $BUILDENV_NAME not found at ${BUILDENV_BASEPATH}, downloading from https://downloads.mixxx.org/dependencies"
```

daschuer (changes requested):

JoergAtGithub:
@acolombier I fear you missed it, but we just moved the download process into CMake: https://github.com/mixxxdj/mixxx/pull/13306
The batch an shell files for Windows and macOS are now just settings files without program logic, that should be replaced by platform independent CMakePresets in the next step. All program logic should be implemented in CMakelists.txt or sub-files therefore.

acolombier:
Ah thanks, I though I had seen something like this but couldn't find it on `main`.

Shame's on me, could have saved me bunch of hours try to get CMD to work :sweat: 

I will migrate that to CMakeFile, which is a far more elegant way.

JoergAtGithub:
> Shame's on me, could have saved me bunch of hours try to get CMD to work 😓

Sorry for that - I always try to prevent batch and bash scripts on Windows - sooner or later, they make always trouble.

acolombier:
No, it's great to hear! The more I was fiddling with the script, the less confident I was getting about maintaining it, so seeing we are consolidating into something a bit more robust and common across platform is great.

Looking at the scripts now, they seems to only set variables and store active version for given platform, is the long term plan to entirely migrate them to CMake and drop them?

JoergAtGithub:
> Looking at the scripts now, they seems to only set variables and store active version for given platform, is the long term plan to entirely migrate them to CMake and drop them?

Yes, that's my plan. It comes on top, that we currently generate the VisualStudio specific CMakeSettings.txt file in this batch file. But this syntax is nowadays standardized for all IDEs and platform independent as CMakePresets. This would also unify the way how IDE users, and users that build on shell, specify CMake cache variables.
 I'm still playing around with CMakePresets, but I think it should be able to replace everything that's left in the batch/shell scripts.

PR #14478, opened 2025-03-14
Detect arch and fail early if not supported when installing buildenv

daschuer:
In case the host architecture is not supported a link to 
https://github.com/mixxxdj/mixxx/wiki/Compiling-dependencies-for-macOS-arm64
is displayed that allows to build the environment at home.

### DIFF
--- a/tools/macos_buildenv.sh
+++ b/tools/macos_buildenv.sh
@@ -19,23 +19,32 @@ realpath() {
 THIS_SCRIPT_NAME=${BASH_SOURCE[0]}
 [ -z "$THIS_SCRIPT_NAME" ] && THIS_SCRIPT_NAME=$0
 
-if [ -n "${BUILDENV_ARM64}" ]; then
-    VCPKG_TARGET_TRIPLET="arm64-osx-min1100-release"
-    BUILDENV_BRANCH="2.5-rel"
-    BUILDENV_NAME="mixxx-deps-2.5-arm64-osx-min1100-release-40c29ff"
-    BUILDENV_SHA256="b76685e77f681baf8fdc5037297b0f16d323a405d09ce276d8844304530278e1"
-else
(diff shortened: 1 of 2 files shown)


JoergAtGithub:
I would prefer to not add this additional functionality to these batch scripts, especially to windows_buildenv.bat. Let's us first merge https://github.com/mixxxdj/mixxx/pull/13306 and than use the CMAKE_SYSTEM_PROCESSOR / CMAKE_HOST_SYSTEM_PROCESSOR instead.

daschuer:
The windows packages are not renamed. It looks like the GitHub Div view has issues to detect that this is just an indentation change, because we only install the environments on x64 windows. 

The windows change should catch x86 and arm64 hosts. They are rare but I have added it for completeness. 

For my understanding even after https://github.com/mixxxdj/mixxx/pull/13306 the script can be run stand alone and detecting the host architecture is beneficial. Is that correct?

JoergAtGithub:
> For my understanding even after #13306 the script can be run stand alone and detecting the host architecture is beneficial. Is that correct?

With #13306 the windows_buildenv.bat will only write/set static content and all application logic is moved intro CMakeLists.txt. There we can also check for the host architecture by CMAKE_HOST_SYSTEM_PROCESSOR.

daschuer:
With this patch the script is able to pick the right environment according to the host. This is especially relevant on macOS where we have x63 and arm64 devices. A check inside CMakeLists.txt cant do this.

daschuer:
I have back ported the windows_buildenv.bat solution form the 2.6 branch: https://github.com/mixxxdj/mixxx/pull/14815

daschuer:
This is ready for merge.

JoergAtGithub (approved):
LGTM! Thank you!

